PR #1, opened 2015-04-16
Readme update

DougBurke:
I didn't think it made sense to have the source build instructions separated by download type, so I have amalgamated the two.

Several links have been added, and a little bit of explanation.

I think something like the History section is useful, but can be taken out if need be (which is why it's in a separate commit).

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,3 +350,14 @@ These options include:
 The `setup.cfg` file in the Sherpa source distribution contains more information
 about these options.
 
+History
+=======
+
+Sherpa is developed by the
+[Chandra X-ray Observatory](http://chandra.harvard.edu/) to provide
+fitting and modelling capabilities to the
+[CIAO](http://cxc.harvard.edu/ciao/) analysis package. It has been
+released onto
+[GitHub](https://github.com/sherpa/sherpa) under the
+GNU General Public Licence, version 3 for users to extend (whether
+to other areas of Astronomy or in other domains).
--- a/README.md
+++ b/README.md
@@ -1,9 +1,13 @@
 
 Sherpa
 ======
+
 Sherpa is a modeling and fitting application for Python. It contains a powerful language for combining simple models
 into complex expressions that can be fit to the data using a variety of statistics and optimization methods.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,49 +184,85 @@ terminal or deactivate the Sherpa environment.
...
-    $ cd sherpa-<version>
-    $ python setup.py install
-
-Prerequisites:
+The prerequisites for building from source are:
 
  - Python: `setuptools`, `numpy`
  - System: `gcc`, `g++`, `gfortran`, `make`, `flex`, `bison`
 
+At present it is suggested that Sherpa be built with Python 2.7.
```
olaurino: I don't think Sherpa builds with Python 3.x, and last time I tried (by mistake) to build Sherpa with Python 2.6 it failed. Sherpa is not supported with these versions. So, shouldn't this statement about the Python version somewhat stronger?

README.md
```diff
@@ -181,49 +184,85 @@ terminal or deactivate the Sherpa environment.
...
+    $ python setup.py install
+
+### 2d. Testing the build
+
+To test that your installation of Sherpa is working, type:
+
+    $ sherpa_test
+
+which will run a small test suite (the location of this script depends
+on where the installation step chose). Note that the test may report
```
olaurino: "the location of this script depends on where the installation step chose" -> It sounds like something is missing.

README.md
```diff
@@ -181,49 +184,85 @@ terminal or deactivate the Sherpa environment.
...
+several `WARNING` lines and failed attempts to load modules. These 
+messages are expected - as these are from tests of some of the optional
+packages within Sherpa - and what is important is that the test ends
+with an
+
+    OK
+
+message.
```
olaurino: May I suggest:

Note that the test may report several `WARNING` lines and failed attempts to load modules. These messages are expected - as some of the tests require optional packages to be installed alongside Sherpa. These warnings may be ignored, as long as the test ends with an OK message.

DougBurke:
Hopefully this addresses your points.

olaurino:
Perfect, thanks!

PR #2, opened 2015-04-17
Fix for -Werror=format-security build error

DougBurke:
On some systems, building Sherpa can end in the following error:

```
c++: sherpa/optmethods/tests/_tstoptfct.cc
sherpa/optmethods/tests/_tstoptfct.cc: In function ‘PyObject*
init_optfcn(PyObject*, PyObject*)’:
sherpa/optmethods/tests/_tstoptfct.cc:1482:40: error: format not a
string literal and no format arguments [-Werror=format-security]
     static_cast<const char*>( errmsg ) );
```

This patch provides one fix. It is also possible to override the compile-time flags, as shown below,
to avoid needing to apply a patch. Run `python-config` to figure out the `CFLAGS` that are going to be used to build C/C++ extensions:

```
$ python-config --cflags
    -I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7  -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes
```

Then call `python setup.py install` with the CFLAGS output from the above command, but *without* `-Werror=format-security`. For example (using bash):

```
$ OPT="" BASECFLAGS="" CFLAGS="-I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7  -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat  -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes" python setup.py install
```

### DIFF
--- a/sherpa/optmethods/tests/_tstoptfct.cc
+++ b/sherpa/optmethods/tests/_tstoptfct.cc
@@ -1474,12 +1474,12 @@ static PyObject *init_optfcn( PyObject *self, PyObject *args ) {
 
   if ( xpar.get_size() != lo.get_size() ||
        xpar.get_size() != hi.get_size() ) {
-    char errmsg[128];
-    sprintf( errmsg, "init_optfcn: Incompatible array sizes "
-	     "xpar=%d, lo=%d, hi=%d\n",
-	     (int) xpar.get_size(), (int) lo.get_size(), (int) hi.get_size());
     PyErr_Format( PyExc_ValueError,
-		  static_cast<const char*>( errmsg ) );
+		  "init_optfcn: Incompatible array sizes "
+		  "xpar=%d, lo=%d, hi=%d\n",
+		  (int) xpar.get_size(), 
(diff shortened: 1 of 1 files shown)


olaurino:
We are already in Release Mode, and merging the PR would require to re-build and re-test. The PR itself does not fix an issue with the released binaries, but a source build issue on some platforms with a specific version of Python built with certain flags.

We suggest that you re-issue the PR against the master branch after the release. Then we can document the issue and point users to the commit that fixes it, so that they can build Sherpa on the affected platforms as an additional workaround.

Does this sound OK?

Thanks for the patch!

DougBurke:
I wanted to make sure there was some documentation for this/it didn't get
forgotten about as I'm away next week.

Doug

PR #6, opened 2015-04-23
Default to use astropy.io.fits instead of pyfits

taldcroft:
This make `astropy.io.fits` the default FITS file reader but falls back to `pyfits` if it is not available.

It also adds a test of reading a simple two-column FITS file, which I have tested locally with `astropy` installed but not `pyfits.


olaurino:
This PR has been replaced by #29

PR #7, opened 2015-04-23
Update .gitignore to exclude inline outputs of build process

taldcroft:
Currently `git status` is dominated by build process outputs.  Not sure if this list is not too encompassing, but it seems reasonable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,30 @@
 # C extensions
 *.so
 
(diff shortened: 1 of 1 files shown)


olaurino:
We are still reviewing our practices for code reviews on PRs, but this is trivial enough I am going to take it. Worst case: we will need to tweak gitignore again.

PR #10, opened 2015-04-23
Skip unit tests instead of pretending to pass

taldcroft:
Currently the unit tests just pretend to pass tests that cannot be run.  This changes so that they are explicitly skipped and show up as such when running tests.

This is based off of PR #6 (io-fits branch) because I needed some functionality from that.  The first commit here is therefore a repeat that will go away once PR #6 is merged.


olaurino:
Tom, thanks for doing this... it has been on my wishlist for a while :)
I am currently working on the improvement of the tests infrastructure, and this PR helps a lot.

Some issues:

  * there is an error on this line:
https://github.com/sherpa/sherpa/pull/10/files#diff-7f4e3ad74bd90eae5be1dea0d84d9ca3R57
It should be:
```
im = sherpa.image.ds9_backend.DS9.DS9Win(sherpa.image.ds9_backend.DS9._DefTemplate,
```

  * The copyright year should be updated (and I'll need to add this to the CONTRIBUTING page I am drafting).

  * Moreover, if you `python setup.py develop` then the xspec tests are not skipped (if xspec is missing) and the tests result in an error.

I have fixed these issues in a branch on my local repo, and I'll issue a PR against your branch as soon as I am done testing (at this point I guess it will be tomorrow).

taldcroft:
@olaurino - sounds good!

olaurino:
@taldcroft I would like to remove the dependency of this PR from astropy, so I am going to create two new PRs, one that has the functionality introduced by #10 (plus the latest commits I added) and one with the additions in #6. Basically I am refactoring the code changes so to invert the dependency of these two PRs. This will also allow me to base the new test infrastructure on the new version of #10 without having to also test for the astropy inclusion.

I am going to maintain the authorship of your original commits.

Unfortunately, changes to the skip conditions need to be made in many places, manually or by relying on `sed` (or similar tools), even to just change the skip messages. At some point I'll try to refactor things so that one only has to add a simple decorator, and changes can be done in a single place, i.e. in the definition of the decorator.

olaurino:
This PR has been replaced by #28

PR #11, opened 2015-04-24
Add some "standard" classifiers to the setup

DougBurke:
Based on a #pyastro15 comment at https://twitter.com/kylebarbary/status/591272058560077824 I've taken a subset of the astropy classifiers. Two questions

  1. does Sherpa actually use Cython?

  2. it wasn't clear whether the license is GPL 3 or GPL3 or later, so I opted for the former.

I referenced the list of topics at https://pypi.python.org/pypi?%3Aaction=list_classifiers


olaurino:
It seems like this PR introduces a syntax error, as `classifiers` is introduced before the end of the `entry_points` dictionary. I am putting the PR on hold so it does not get integrated in the next snapshot build.

DougBurke:
I've actually built Sherpa with the new setup.py this time! Can you access these strings from Python, or is it only for built packages? Also, the following fields need to be reviewed to see if they are correct (are we using "GPLv3" or" GPLv3 or later" and does it use Cython):

    'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
    'Programming Language :: Cython',

olaurino:
@DougBurke Sherpa does **not** use Cython.

Regarding the License, I'll need to check with the rest of the team whether we want to add the language "or any later version", which is currently missing (see https://github.com/sherpa/sherpa/blob/master/COPYRIGHT#L18).
For the record CIAO does have an "or any later version" clause.

Also, as far as I know the classifiers are only used in the Python Package Index.

olaurino:
@DougBurke I can confirm that Sherpa is released under the terms of the GPLv3+ (v3 or at the user's option any other later version). I am creating a different PR to clarify this in the COPYRIGHT file.

DougBurke:
This is ready for review now.

olaurino:
:+1:

olaurino:
I am rebasing this one before I merge it in order to reduce the clutter in the commits history. The original commit was d898dc4.

PR #17, opened 2015-04-29
docstring updates (ahelp ingest)

DougBurke:
# Aim

Update the Python docstrings with information from the CIAO documentation system. This is not intended to change any code, just documentation. I have spent a lot of time merging and fixing up the existing documentation, but there are definitely still places where things could be improved. However, I would much prefer to get this merged and then fix things up in a later pass.

# Status

Finished and ready for review.

Functions in `sherpa.astro.utils` without a doc string (excluding those beginning with a _ or are not a top-level function): `notice2d_image`, `ignore2d_image`, `save_session`. Surprisingly enough we don't seem to have ahelp files for the first two. I do not intend to update their documentation in this PR, this can be done later. All top-level user routines in `sherpa.utils` now have docstrings (several did not have ahelp files). 

Documentation has been added to other parts of the system where appropriate, but this is not extensive, or complete. This is for "helper" routines and classes which should be documented but for which we do not have an ahelp file.

The PR contains 437 functions labelled with the `Ahelp ingest` comment I used for tracking changes. I am leaving these lines in to aid review, but I expect to delete them after review. Note that this was only used for functions which had ahelp pages; changes made to other routines were not necessarily labelled in this manner.

## Description

The Sherpa user-level documentation was provided as part of the CIAO help system (for various technical and social reasons). While there are docstrings for functions, they have not all been kept up to date, and some functions have no docstrings. I aim to consolidate the existing information (from the docstrings and CIAO help system - e.g.  the information at http://cxc.harvard.edu/sherpa/ahelp/) and add this, using the [astropy guidelines](https://astropy.readthedocs.org/en/stable/development/docrules.html).

This is mainly going to address the routines in `sherpa.ui.utils` and `sherpa.astro.ui.utils` but there will eventually be changes throughout the code base (but not for all functions).

I have added a number of comments, indicating that a command has been updated (`### ahelp ingest: ...`), as well as todo's (`### DOC-TODO: ...`) and some notes (`### DOC-NOTE: ...`). These will be cleared up/turned into tickets/... after the PR has been merged.

## Not Done

The model descriptions need to be added. This will be done in a later PR, and requires some changes to Sherpa to be truly effective (so that a user can say `help(powlaw1d)` and get it to display the documentation rather than for the `ModelWrapper` class, as discussed below).

Two "top-level" modules need documentation review: [datastack](http://cxc.harvard.edu/sherpa/ahelp/datastack.html)  and [pyblocxs](http://cxc.harvard.edu/sherpa/ahelp/pyblocxs.html). The individual pyblocxs commands have been documented - e.g. `get_draws` - but the module-level documentation hasn't been addressed in this patch. I decided to leave the `datastack` out of this round of changes as I am still not sure how to integrate it, documentation wise.

I have not reviewed the post-CIAO 4.7 changes made to the ahelp files (this conversion was done using the ahelp files distributed with CIAO 4.7). Any changes can be made in a separate PR.

## Testing plan

As this is meant to be a documentation-only PR, my testing is limited to running the sherpa_test executable and spot tests that the information I've added has been picked up by a `help(...)` call.

# Notes added

This is based on things I've noticed whilst looking through the code and trying some of the corner cases out. Bugs can be filed later if necessary.

## Open issues

Some of the CIAO documentation contains information that makes more sense to be in general documentation - e.g. created by sphinx - rather than inserted as docstrings. I don't plan to work on this in this PR.

I plan to move the model descriptions to the relevant classes. However, due to the way that the model wrapping works, it is not yet clear whether a user who writes

    >>> help(ui.powlaw1d)

will get the information, but they should for a model instance - e.g.

    >>> ui.powlaw1d.mdl
    >>> help(mdl)

## Bugs

 - `save_quality` writes out a column called 'GROUP' when it should be called 'QUALITY'

 - `list_priors` should return a dictionary, not `str(dict(...))`!

- `pack_pha` fails with recent PyFITS version (`CardList` deprecation, `append` method); I
   haven't looked to see if this has been addressed by #6 

- I can run `int_proj` with various params - e.g. `min=6.8, max=11.5, delv=0.2` but if I then try `get_int_proj` with these parameters and add the necessary `recalc=True`, then I get various errors.

- write_table() of a PHA data set using crates raises a ValueError: `Unable to create column HEADER: unsupported datatype ('numpy.object_').`

- for a data set for which `sample_photon_flux` runs, I get a TypeError from `plot_photon_fluxes`: `TypeError: scales option must be iterable of length 2`. Also, I expected this routine to accept the scales parameter but it isn't directly listed.

## Strong suggestions for fixes

 - it looks like the `xlabel` parameter of `plot_trace` does nothing and should be removed

 - do `load_grouping`, `load_quality`, `load_staterror`, ... need the `ncols` parameter, as they seem to only read in one column of data.

- `save_error` will run on an image data set, but the output is not very useful (two columns, presumably X0 and ERR values, so missing X1). This should either be fixed or an error thrown. I'm guessing it holds for other `save_xxx` commands in `sherpa.astro.utils`.

- it looks like `get_int_proj` should support the `fast` parameter.

## Things to look at

- `get_method_name` and related could be changed to use the `.name` field rather than converting the class name.

-  `delete_data` does not remove any model expression created by `set_model`. Is this the indented behavior (it may well be)?

- The `.cache` value of models appear to be 5 and not 1, which I expected. Does the documentaion need updating or do the values have any special meaning?

- The `group_xxx` set of routines could have some of their argument names changed: e.g. `group_width` to use `val` rather than `num`, `group_adapt_snr` to use `snr` rather than `min` (to match `group_snr`).

- ~~Does `unpack_rmf` work with AstroPy (a simple test with PyFITS threw an error)?~~ User Error!

- can `unpack_table` and `unpack_ascii` be combined?

- `sherpa.utils.set_iter_method` raises a `TypeError` rather than something derived from `sherpa.utils.err.SherpaErr` (I forget know why we have our own error hierarchy; it may be worth reviewing this, but at the moment we may want to remain consistent, or at least note when we are not).

- there doesn't appear to be any validation of the gridsearch method option (I set it to 'bob' and `fit`seemed to run okay). 

- `set_model_autoassign_func()` fails if called with the default of `func=None`: `AttributeError: 'Session' object attribute '_model_autoassign_func' cannot be replaced with a non-callable attribute`. It either should not be the default value or `None` should be handled.

- does `calc_model_sum2d` work when a PSF is applied via `load_psf`/`set_psf`?

- `sherpa.ui.utile.get_indep()` does not support a `filter` parameter when it looks like it should (the `sherpa.astro.ui.utils.get_indep()` variant does, as does `sherpa.data.Data.get_indep()`

- should there be a `sherpa.astro.utils.get_filter` version with a `bkg_id` parameter?

- does the `newframe` option make sense for `image_fit`?

- `get_source_image` throws an AttributeError about `_check_shape` being missing; I expected, given the other commands, that it would raise `sherpa.utils.err.DataErr` instead.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -8930,6 +8930,7 @@ class Session(sherpa.ui.utils.Session):
         --------
         load_conv : Load a 1D convolution model.
         load_psf : Create a PSF model
+        load_template_model : Load a set of templates and use it as a model component.
         set_model : Set the source model expression for a data set.
         set_full_model : Define the convolved model expression for a data set.
 
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -3498,7 +3498,7 @@ class Session(sherpa.ui.utils.Session):
         See Also
         --------
         save_data : Save the data to a file.
-        save_image :
+        save_image : Save the pixel values of a 2D data set to a file.
         save_table :
 
         Examples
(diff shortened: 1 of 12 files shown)


cdeil:
Wonderful, thank you!

Do you have a Sphinx build already for `sherpa` and if yes, could you post the HTML version of the API docs somewhere online? (We usually do this in Astropy for extensive docstring changes to make it a bit easier to review)

DougBurke:
I have several "explorations" of the sphinx output lying around, but it's
all a bit of a mess. I wanted to get this conversion done first before
getting back into the overall docs.

Doug

taldcroft:
:clap:

PR #18, opened 2015-05-05
improve .gitignore and remove generated region files

olaurino:
**WIP - This branch may be rebased**

Several files are created by `make` in the region library and need to be ignored. Some of them may have been included in the source distribution by mistake and should probably be `git rm`-ed and then `.gitignore`-d.

Other useful additions:
 * python `.eggs`
 * vi swap and backup files
 * files created by `f2py`

I personally have some of the above ignored at the global level, so I am trying to build Sherpa from the `git` repo on vanilla virtual machines without any globally defined ignore files in order to figure out what to include.


DougBurke:
@olaurino - did you make @cd94302 ? I'm wondering why there's no associated user.

olaurino:
I did. Sometimes I make changes directly from the virtual workers, but git usually asks me to set user name and email even if I forget to do so myself, so I am not sure why that one is anonymous.

DougBurke:
Looks good to me - after `python setup.py build / install` I get a clean `git status` report rather than the long list of files that you currently get.

DougBurke:
As a note, I checked the existing repository against the contents of `.gitignore` and got the following:

```
% git clone https://github.com/sherpa/sherpa sherpa-test
% cd sherpa-test
% find . -type f | grep -v '\.git' > ../lis.files
% git check-ignore --no-index --stdin < ../lis.files
./extern/wcssubs/wcssubs-3.8.7/Makefile
./extern/stklib-4.6/python/Makefile
```

You get the same results with the current PR:

```
% git checkout gitignore-region
% find . -type f | grep -v '\.git' > ../lis2.files
% git check-ignore --no-index --stdin < ../lis2.files
./extern/wcssubs/wcssubs-3.8.7/Makefile
./extern/stklib-4.6/python/Makefile
```

Is this expected? Note that it's not an issue for this PR since (as shown) it's got nothing to do with the proposed changes.

olaurino:
@DougBurke thanks for checking. Those files are there for different reasons:
  * `wcssubs` is wrapped in an `autoconf` configuration for consistency with the other packages in `extern`, and because Sherpa was already expecting an external library rather than a more pythonic inclusion in the build of the extension. So, the original `Makefile` from the package is there, but it is ignored.
  * stklib has not yet been ported to autoconf in the original CIAO versioning system, and there may be some leftovers from the temporary migration in Sherpa, like that `Makefile`. Again, this file is ignored.

olaurino:
@DougBurke do you think we can merge this?

DougBurke:
Yes :shipit:

PR #21, opened 2015-05-11
Upgrade versioneer

### LINKED ISSUE #13: GIT not defined in sherpa/_version.py
I get an error that GIT is not defined on this line when running `python setup.py build_ext --inplace`:
https://github.com/sherpa/sherpa/blame/master/sherpa/_version.py#L129

Can someone reproduce the issue?

(My git is at `/opt/local/bin/git` in case it matters.)

olaurino:
Fix #13 by upgrading versioneer code and relative files.

The `versioneer` allows `sherpa.__version__` to pinpoint the exact `git` commit (or tag, if available) that was used to build Sherpa, including whether the working tree was in a *dirty* state. It uses this project: https://github.com/warner/python-versioneer and produces PEP-440 compliant version strings.

During the latest update to version 0.13dev (which introduced PEP-440 compliance), the `_version.py` file was not properly upgraded, and this might have resulted into issue #13.

However, the changes in this PR would be required independently of issue #13.


DougBurke:
I've labelled this pr:merge after testing alongside PR #23.

Issue #22, opened 2015-05-20
datastack module and optional requirements


olaurino:
As mentioned in the [4.7 release notes](https://github.com/sherpa/sherpa/releases/tag/4.7), the new `datastack` module fails to load when `matplotlib` is not installed.

This is a side effect of the decision to have the IO and plotting backends as soft dependencies, while in CIAO there is always a back-end installed.

Moreover, if `matplotlib` is installed, but `pyfits` is not, the `datastack` module will indeed be loaded, but the tests will fail because `pyfits` is not available.

The `datastack` should check for the presence of the backends and act accordingly.

Installing `matplotlib` and `pyfits` works around the issue.

DougBurke:
On a possibly-related note, in tests I've been doing to set up some sphinx documentation, I've had problems with the `autosummary` extension (in particular, the `astropy_helpers.sphinx.ext.astropyautosummary` version) because the `sherpa.image.DS9` module raises an error when imported and DS9 is not available. I haven't had time to track down what's going here, but thought I'd raise it here (as it's related to how Sherpa deals with optional dependencies, even if it turns out to be a different problem).

DougBurke:
@olaurino - from your original comment, do you know what "act accordingly" means, or does this need to be worked out? I assume that if a backend is missing then either:

a) there's a warning message saying something like "datastack support is not available" and the module is not loaded

b) the module will load, but will do nothing when a backend is missing.

I guess option a fits in better with the current approach of Sherpa (but I haven't looked at the tests/code to see which might make more sense functionally).

olaurino:
@DougBurke 

option a) is what happens now. The module fails to load and the user cannot use it.

Some of the module's features do not have anything to do with plotting, and they should be available even if plotting backends are not installed.

So, I believe we should make the module work like in option b), with some of the functionality always available, and some other functionality available only if a plotting backend is installed.

Also, we should refactor the parts of the code that directly probe for the installed backend. First, we should define common interfaces that can be factored out so that the main code gets the implementation from a factory. Then we should evaluate whether we want to fold this new abstraction layer of the datastack module into the main sherpa backend implementations themselves, but at that point this should just be a matter of moving things around.

We should also make sure we track the issue you were having with DS9 and sphinx, and in general this could be part of the larger review of the backends design.

To summarize, here is what I believe we should do:
[...]

PR #23, opened 2015-05-20
make python setup.py develop transparent to user

### LINKED ISSUE #14: python setup.py develop not transparent to the user
When I work around issue #13, then `python2.7 setup.py build_ext --inplace` succeeds, but on import I get this error:
```
$ python2.7 -c 'import sherpa.all'
  File "sherpa/all.py", line 35, in <module>
    from sherpa.data import *
  File "sherpa/data.py", line 29, in <module>
    from sherpa.utils.err import DataErr, NotImplementedErr
  File "sherpa/utils/__init__.py", line 37, in <module>
    from sherpa.utils._psf import extract_kernel, normalize, set_origin, \
ImportError: No module named _psf
```
Most extension modules are created OK, but for some reason `_psf.so` is not:
```
$ find . -name '*.so'
./sherpa/astro/models/_modelfcts.so
./sherpa/astro/utils/_pileup.so
./sherpa/astro/utils/_utils.so
./sherpa/estmethods/_est_funcs.so
./sherpa/models/_modelfcts.so
./sherpa/optmethods/_minim.so
./sherpa/optmethods/_minpack.so
./sherpa/optmethods/_saoopt.so
./sherpa/optmethods/_tstoptfct.so
./sherpa/stats/_statfcts.so
./sherpa/utils/_utils.so
./sherpa/utils/integration.so
```
Maybe there's an issue how the extensions to build are defined here?
https://github.com/sherpa/sherpa/blob/master/helpers/extensions/__init__.py

olaurino:
This PR enables transparent `python setup.py {build_ext, test, develop}` commands, by making sure that Sherpa-specific commands are run before the standard ones.

It introduces several aliases in `setup.cfg` to ensure that `sherpa_config` and `xspec_config` commands are always run when `build_ext`, `develop`, and `test` commands are given to `setup.py`, while still allowing the user to have control over which commands are run.

Moreover, it introduces a custom `develop` command that makes sure `stk` and `group` extensions are installed as well.

To test the PR, run `python setup.py develop` and check that `sherpa_test` runs and without failures. Also, test that `python -c "import group"` and `python -c "import stk"` don't fail.

Moreover, `python setup.py build_ext --inplace` should build the `_psf.so` and `_wcs.so` should be built (see #14).

To test this PR one also needs PR #21 and both `pyfits` and `matplotlib` to be installed (see issue #22).

This PR fixes #14 and supports the new testing infrastructure that is under development.

### DIFF
--- a/helpers/__init__.py
+++ b/helpers/__init__.py
@@ -1,5 +1,5 @@
 # 
-#  Copyright (C) 2014  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2014, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
--- a/helpers/develop.py
+++ b/helpers/develop.py
@@ -1,5 +1,5 @@
 # 
-#  Copyright (C) 2014  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
--- a/helpers/develop.py
+++ b/helpers/develop.py
@@ -17,21 +17,16 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+import shutil
+import os
+from numpy.distutils.command.develop import develop as _develop
 
-from build import build
-from clean import clean
-from develop import develop
-from install import install
-from sdist import sdist
-from sherpa_config import sherpa_config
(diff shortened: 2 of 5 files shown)


olaurino:
@DougBurke could you please review the latest changes?

I made the few changes you asked for. Plus, I made a change to avoid that the `configure` script in `extern` is run every time, as long as `extern/built` is present.

If you are okay with the changes, you can label the PR to merge.

DougBurke:
I've just tried a `python setup.py develop` with your latest changes and things still work.

PR #8, opened 2015-05-22
Set up travis-ci

cdeil:
First of all, it's great that Sherpa is on Github now, thank you!

How about setting up https://travis-ci.org/ for continuous testing of Sherpa?
I've done this for a few Python packages, so I could make a pull request adding the `.travis.yml` file.
But an admin for the Sherpa Github repo has to activate builds on travis-ci as described [here](http://docs.travis-ci.com/user/getting-started/#Step-one%3A-Sign-in) ... it just takes a minute.

### DIFF
--- a/.continuous-integration/README.md
+++ b/.continuous-integration/README.md
@@ -1,0 +1,15 @@
+About
+=====
+
+This directory contains a set of scripts that are used by the 
+``.travis.yml`` and ``appveyor.yml`` files for the 
+[Travis](http://travis-ci.org) and [AppVeyor](http://www.appveyor.com/) 
+services respectively.
+
+The scripts include:
+
+* ``appveyor/install-miniconda.ps1`` - set up conda on Windows
+* ``appveyor/windows_sdk.cmd`` - set up the compiler environment on Windows
+* ``travis/setup_dependencies_common.sh`` - set up conda packages on Linux and MacOS X
+* ``travis/setup_environment_linux.sh`` - set up conda and non-Python dependencies on Linux
+* ``travis/setup_environment_osx.sh`` - set up conda and non-Python dependencies on MacOS X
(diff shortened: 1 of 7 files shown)


olaurino:
@cdeil thanks for volunteering, but keep in mind that I am already working on the travis-ci configuration, since I already have experience with both travis and setting an environment for testing sherpa on virtual workers. After I set up a basic configuration we can complicate it at will (e.g. by testing with different versions of numpy).

We do not support windows. I tried building Sherpa for windows on a weekend at some point, and it should not be terribly difficult to support this platform, but that's a project of its own and we never got this request, so that's unlikely to happen (at least on our side). If somebody wanted to try, I can share by notes on the windows build.

On the other hand, I will see if I can enable OSX testing, and your experience with setting it up for Astropy will be beneficial.

As for the Python 3 support, it's a big project. Not only the pure Python part of Sherpa needs to be updated, but also all the C and Fortran extensions need to be ported. I am not saying it's not going to happen, just that it can't happen overnight.

cdeil:
@olaurino – I didn't know you are setting up a travis-ci test. Great! We can just close this PR once that's done.

olaurino:
@cdeil thanks for the commits, they certainly help. May I ask if you tested them?

cdeil:
I just copied the files from the Astropy repo in the first commit and made some obvious edits in the second commit. I didn't do any testing and as you can see, the travis-ci build fails. Now if I were to do it, I'd just repeatedly edit `.travis.yml` until all builds pass, or, if I run into weird errors I don't understand, try to reproduce the error locally (see notes [here](https://github.com/gammapy/gammapy/blob/master/dev/docker/commands.sh)).

Let me know if there's anything I can help with here (although I'm offline now for a few hours).

olaurino:
sorry, I made a stupid question... I should have gone to lunch before commenting ;)

olaurino:
This PR is replaced by PR #31. Closing. Thanks @cdeil for the contributions! I hope to get your feedback on #31.

PR #28, opened 2015-05-27
Skip tests instead of pretending to pass

olaurino:
Tests are properly skipped when some conditions are not met, e.g. when there is no `FITS` backend available, or the external data/threads data folder is missing.

This is largely based on @taldcroft's original PR #10, but it removes the dependency on astropy (PR #6), it includes more checks, and covers the CIAO use cases, i.e. it also checks for the presence of `pycrates`.

To test:

```
python setup.py develop
sherpa_test
```
tests should be run with and without pyfits or pycrates, with and without xspec, with and without the test data directory (i.e. `sherpa_test -d /data/scialg/testdata`).

When running the full regression test suite (i.e. pyfits+testdata, with or without pyfits) some tests are expected to fail because of unmet dependencies (CIAO data model, pycrates), and they will be fixed in the PR that introduces the new testing infrastructure.

Below is a log of the tests that are expected to fail:
```
$ sherpa_test -d /data/scialg/testdata
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available
Failed importing sherpa.astro.xspec: No module named _xspec
  Found 7/7 tests for sherpa.astro.tests.test_datastack
  Found 1/1 tests for sherpa.astro.models.tests.test_models
  Found 25/25 tests for sherpa.astro.tests.test_optical
  Found 2/2 tests for sherpa.astro.sim.tests.test_sim
  Found 26/26 tests for sherpa.astro.tests.test_astro
  Found 6/6 tests for sherpa.astro.tests.test_data
  Found 1/1 tests for sherpa.astro.tests.test_plot
  Found 1/1 tests for sherpa.astro.ui.tests.test_nan
  Found 15/15 tests for sherpa.astro.ui.tests.test_ui
  Found 1/1 tests for sherpa.astro.utils.tests.test_utils
  Found 4/4 tests for sherpa.estmethods.tests.test_estmethods
  Found 0/0 tests for sherpa.image.tests.test_image
  Found 1/1 tests for sherpa.models.tests.test_basic
  Found 14/14 tests for sherpa.models.tests.test_model
  Found 12/12 tests for sherpa.models.tests.test_parameter
  Found 6/6 tests for sherpa.models.tests.test_template
  Found 26/26 tests for sherpa.optmethods.tests.test_optmethods
  Found 18/18 tests for sherpa.plot.tests.test_plot
  Found 18/18 tests for sherpa.sim.tests.test_sim
  Found 5/5 tests for sherpa.tests.test_sherpa
[...]

PR #29, opened 2015-05-27
default to astropy.io.fits instead of pyfits

olaurino:
Sherpa tries to import `astropy.io.fits` when the backend configuration is set to `pyfits`. If `astropy` is not present, Sherpa tries to fall back to using `pyfits`, if available.

This PR is largely based on #6, but it has been rebased to depend on #10 (now #28) instead of the other way around.

Since this PR is based off #28, when #28 is merged this PR will result in a single commit introducing `astropy`, a simple test, and an updated `has_fits_support` check.

PR #25, opened 2015-05-23
Doc updates

DougBurke:
Status
------

Finished. There's always more doc-work to be done, but I think this is a reasonable chunk to merge, and I have to focus on some other things on the short term, so would rather get these changes installed before there's too many changes made to the master branch.

Aim
----

These changes

 - fix some typos and cut-and-paste issues from #17

 - update several functions to use the NumPy/AstroPy style (the main user-facing ones being `notice2d_image` and `ignore2d_image`, which don't have ahelp files so were missed in #17)

 - fix up warnings generated by Sphinx and changes to make the documentation look better in sphinx-generated HTML (note that there's no sphinx documentation included in this PR as that is still being worked on).

PR #24, opened 2015-05-21
Make __all__ a tuple

DougBurke:
Sphinx complains that `__all__` isn't a tuple.

I've also added 2015 to the copyright.

PR #31, opened 2015-05-29
Feature - New build and test infrastructure

olaurino:
### Release Note
The GitHub repo is now integrated with [Travis CI](https://travis-ci.org), so that builds will be automatically started every time there a commit is pushed to GitHub, or a Pull Request is issued. The full regression tests suite will be run after the build.
Developers using `git` will now be able to `init` and `update` a `git` submodule with external data and additional regression tests. A simple `python setup.py test` will run the tests suite. Package [`py.test`](http://pytest.org/latest/) is required for running the tests suite in this fashion. For compatibility with CIAO, `python -c "import sherpa; sherpa.test()"` is left untouched, and can be run without any additional dependencies. The `sherpa_test` console script has been updated to use the new framework and now requires `py.test`. The tests now try and find the external data directory at runtime, so no command line arguments are required and discovery of tests is automatic. However, for compatibility, `sherpa.test()` still needs the `datadir` argument to properly run: if `datadir` is not provided, only the embedded tests will be run (*smoke* test).

### PR Description
This PR introduces a new testing infrastructure for running the whole Sherpa unit and regression tests suite from the `git` repo, a source distribution, or a binary package. This includes migrating the tests that would run only on the CXC systems (see [the `sherpa-test-data` repo](https://github.com/sherpa/sherpa-test-data)) as a `git` submodule. Moreover, it integrates the GitHub repository with [Travis CI](https://travis-ci.org), so that builds will be automatically started every time there a commit is pushed to GitHub, or a Pull Request is issued, and tests are run after the build.

This PR depends on PR #28. 

### Requirements
  * we want to keep `python -c "import sherpa; sherpa.test()"` as is for compatibility with CIAO: the data directory path can be passed from the command line in this case.
  * we want the tests to find out at runtime where the data directory is.
  * the data directory needs to be external to the sherpa codebase, i.e. not in the same `git` repository.
  * the data directory needs to be versioned.
  * since the data directory and sherpa are versioned independently, we need to synchronize the Sherpa revision with the `sherpa-test-data` revision.
  * ideally, `python setup.py test` should be able to discover the test data directory independently of whether `git` is being used or not, i.e. a regular user without `git` should be able to run the tests in a transparent way.

### Design
One natural choice to meet most of the above requirements is to create a Python project containing the test data, and to include it as a `git` submodule of the main Sherpa repository. The project could be either a Python module (i.e. with a `__init__.py`), or its own installable Python package. 

In the first case, we should rely on the relative position of the `sherpatest` module so that it can be imported. 

However, the first case would work fine assuming `git` is installed. Having the test data directory in its own Python project would allow it to be installed without `git`: it may be listed as a regular package in `test_requirements.txt` and installed with a `pip install -r test_requirements` command, or it might have its own `conda` package. In this case, however, it is not clear how one would synchronize the versions: this could be done for releases, but not during development.

So, making `git` unnecessary does not seem workable if not for releases, and having the `git` submodule as a standalone Python package seems to be more flexible.

In summary the design can be summarized as follows:
  * create a standalone python package for holding the test data
  * add it as a `git` submodule to the main Sherpa repo
  * at release time, the test data package might be released alongside Sherpa with the same version, so that users may be able to run the full test suite simply by installing the test data package.
  * in CIAO one may rely on the folder of the `git` submodule as the new `testdata` directory.
  * `sherpa_test` will be used as a handle to run the tests in non-development mode. `sherpa.test()` can be used for testing in CIAO.

#### Summary of the UI
  * During development, one runs `python setup.py develop` and tests with `python setup.py test`. The `git` submodule must be initialized and updated for the full regression tests to run, as described in the [`git-submodule` documentation](http://git-scm.com/docs/git-submodule).
  * When installing from sources one runs `python setup.py install` and tests with `sherpa_test`.
    * If using `git`, one can initialize and update the submodule for access to the full regression tests suite. 
    * Otherwise one can install the `sherpatest` package and the tests will figure out the location of the external data directory.
  * When installing from binaries, after the installation one can run `sherpa_test`. For running the whole regression suite one can install `sherpatest` and run `sherpa_test`.

### Implementation
The implementation relies on the cleanup of the `testdata` directory that @jbudynk performed. The cleaned-up folder was imported into [a new `git` repo](https://github.com/sherpa/sherpa-test-data) and transformed into an installable Python package. The repo was also included as a `git` submodule into the main Sherpa repo, so that changes in the revisions of the repos can be tracked, and one can always couple the Sherpa revision with the `sherpa-test-data` matching revision.

Some of the old Sherpa tests required CIAO components and would fail when run in standalone mode. Such tests were modified accordingly. In particular:
  * `ciao4.3/lev3fft/bar.py`: This script requires the CIAO Data Model support to read a region from a `FITS` region file. The test was changed to employ the region string equivalent to the one defined in the `FITS` file.
  * `sherpa/astro/ui/tests/test_astro_ui.py:245`: This test failed because the `pyfits_backend` did not return a proper response matrix from a non-Chandra file representing a diagonal matrix (`threads/chi2/SWIFTJ0840.1+2946.pha.gz` in the old CXC `testdata` dir). The tests works fine when `PyCRATES` (the CIAO backend) is used. Since the test file did not have anything special to do with the code being tested, it was replaced by a Chandra file.

One more issue was triggered by using `py.test` rather than the plain `unittest`, as apparently they have different settings of the system path when tests are run: the failing test was `sherpa/astro/tests/test_astro.py:59`, which raised a `ImportError: No module named acis_bkg_model`. Making the test `eval` the model script is enough to make the test more portable and run both with and without `py.test`.

### Tests
[Build 10 on Travis CI](https://travis-ci.org/olaurino/sherpa/builds/64483010) shows several possible testing scenarios described above. In particular, builds are created with the following assumptions:
  - `pyfits` is not installed, the user builds Sherpa with `python setup.py develop`, and tests it using the `git` submodule.
  - same as above, but `pyfits` is installed.
  - `pyfits` is not installed, the user builds Sherpa with `python setup.py install`, and tests it by installing the `sherpatest` package and by using `sherpa_test`.
  - same as above, but `pyfits` is installed.
  - *smoke test* mode: the user installs Sherpa with `python setup.py install`, `pyfits` is installed, `sherpatest` is not installed, and `sherpa_test` is used to test the installation. By *smoke test* we mean a partial test that probes for obvious issues that would prevent the installation from working, as described in the [CIAO user documentation](http://cxc.harvard.edu/ciao/threads/ciao_install/index.html#smoke)

Build 10 Travis configuration is too heavyweight for being the default configuration, so some of the environment combinations in the Travis configuration were commented out for the time being.

[Build 13 on Travis](https://travis-ci.org/olaurino/sherpa/builds/64579829) shows the single scenario that has been left uncommented.

I also tested on a CXC system that the full tests suite runs when XSPEC is installed.

### Pass additional arguments to py.test
In order to run `py.test` with custom arguments, one can call `py.test` directly, or pass the arguments to `python setup.py test`.

For instance, the commands below show how to execute a single test case:
```
python setup.py test --pytest-args="python setup.py test \
  --pytest-args="sherpa/estmethods/tests/test_estmethods.py::test_estmethods"
```

or a single test:
```
python setup.py test --pytest-args="python setup.py test \
  --pytest-args="sherpa/estmethods/tests/test_estmethods.py::test_estmethods::test_covar"
```

### Future improvements
  * we may want to add xspec, but this requires figuring out the most straightforward way of installing xspec and making it available to Travis workers.
  * clean up `sherpa-test-data` some more: in particular, get rid of the `ciao-4.3` folder, and move files around to get a more rational folder structure. This requires changing the tests embedded in Sherpa, so it was left for a later stage.
  * enable coverage reports (e.g. through `coveralls`). The configuration should already be there, but has been commented out for the time being.
  * Improve test coverage by adding more tests.
  * improve the `sherpa-test-data python package metadata.
  * investigate issue with the pyfits backend and non-Chandra response matrices.
  * test against different versions of Numpy in Travis.
  * test conda binaries creation and upload development artifacts to binstar.


olaurino:
Added some improvements following @DougBurke feedback:
  * removed `__init__.py` file from a test folder. Having `__init__.py` would have been better for `unittest` self-discovery, but it confuses `py.test` and Sherpa did not make use of `unittest` self-discovery anyway.
  * `pytest` was erroneously listed as a dependency in `setup.py`
  * the `pytest` and `setup.py` helpers configuration was tweaked to fix support of `pytest` specific command line arguments, in particular for making it easier to run single tests.

I am going to update the PR description with information about how to pass arguments to `pytest`, in particular for running individual tests.

DougBurke:
@olaurino - do you need to add a rule to `gitignore` for the coverage files, since I'm now seeing the following (although having them listed does remind me to clean them out).

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.coverage.brobdingnag.13051.304912
	.coverage.brobdingnag.13052.998985
	.coverage.brobdingnag.13053.600528
	.coverage.brobdingnag.13054.738715
	.coverage.brobdingnag.13055.447196
	.coverage.brobdingnag.13058.51082
```

olaurino:
Ah, you are correct. I re-enabled my global gitignore too soon! :)

olaurino:
I spoke too soon, as I remembered adding such files on a system, but they were produced by an IDE and that is why I added them to my global gitignore.

On my system I do not see those files, only a bunch of `.coverage` files that are properly `gitignore`-d.

.travis.yml
```diff
@@ -0,0 +1,47 @@
...
+    fi
+  - if [ ${TEST} == smoke ];
+     then git submodule deinit -f .;
+    fi
+
+install:
+  - python setup.py $INSTALL_TYPE
+
+script:
+  - if [ ${PYFITS} == true ]; then conda install --yes pyfits; fi
```
cdeil: Put this line in the `before_install` section where the other dependencies are installed?

I think pyfits is already deprecated or it will be in the coming years, so better to install astropy and test against `astropy.io.fits`? (Maybe in addition to `pyfits` as a separate build if that's important to CIAO?)

olaurino:
@cdeil thanks for your comments. You are right that it would make more sense to move the `conda install pyfits` line in the `before_install` section. I moved stuff around while complicating the script but forgot to move that one. Maybe we can move it when reviewing the file. Which segues into your next comment: the code in this PR does not include the astropy support, so installing astropy instead of pyfits would just result in all the FITS tests to be skipped. When adding astropy support (PR #29) we should make sure that tests pass with both pyfits and `astropy.io.fits`, and update `travis.yml` accordingly. CIAO uses yet another FITS I/O package, i.e. [PyCRATES](http://cxc.harvard.edu/ciao/ahelp/crates.html), with support for Chandra-specific Data Model features.

cdeil:
@olaurino – I read the description of this PR: https://github.com/sherpa/sherpa/pull/31#issue-82468248 and only have two small comments:

It would be nice to have some of that info how to run the tests in the https://github.com/sherpa/sherpa/blob/master/README.md#2d-testing-the-build or a separate `DEVELOP.md`. Or maybe you want to add that later to the Sphinx docs?

> Build 10 Travis configuration is too heavyweight for being the default configuration, so some of the environment combinations in the Travis configuration were commented out for the time being.

Why? It's 5 builds that each ran for 5 minutes. Astropy has more builds and a much higher rate of triggered builds because of the many commits to many PRs.
I think ~ 10 or even 20 builds for Sherpa on travis-ci would be OK.

cdeil:
One more thing ... if you do add some docs on the tests for future co-developers, please explain what smoke- and non-smoke tests are. It's not obvious to me (but I'll have a look tomorrow).

olaurino:
@cdeil 

> It would be nice to have some of that info how to run the tests in the https://github.com/sherpa/sherpa/blob/master/README.md#2d-testing-the-build or a separate DEVELOP.md. Or maybe you want to add that later to the Sphinx docs?

Agreed. @DougBurke how do we coordinate documentation additions to PRs? We probably didn't discuss this, did we? I am happy to update the README.md, although I was not planning to do that just yet, as key aspects of this PR might still change.

>> Build 10 Travis configuration is too heavyweight for being the default configuration, so some of the environment combinations in the Travis configuration were commented out for the time being.

> Why? It's 5 builds that each ran for 5 minutes. Astropy has more builds and a much higher rate of triggered builds because of the many commits to many PRs.
> I think ~ 10 or even 20 builds for Sherpa on travis-ci would be OK.

I agree. However, I was testing different ways of installing and testing Sherpa in order to test the test infrastructure while I was tweaking it. At full speed, I don't think it's necessary to do this kind of meta-testing, while more builds may be necessary for "better" reasons. When there will be tests for different numpy versions, for pyfits vs astropy and maybe xspec vs no-xspec (not to mention python2 vs python3, linux vs mac), this kind of builds would just be noise. I can be convinced otherwise, though. So please let me know if you think that at least a subset of those original tests should be kept. Maybe we should keep the pyfits/no-pyfits tests and ditch the others?

> One more thing ... if you do add some docs on the tests for future co-developers, please explain what smoke- and non-smoke tests are. It's not obvious to me (but I'll have a look tomorrow).

Good point. A generic description of smoke testing is [on Wikipedia](http://en.wikipedia.org/wiki/Smoke_testing_(software)), but in CIAO we explicitly refer to smoke tests in the [user documentation](http://cxc.harvard.edu/ciao/threads/ciao_install/index.html#smoke): since the goal of this PR is to serve both the standalone development while keeping compatibility with the CIAO testing infrastructure, I referred to the *smoke test* in the CIAO lingo. I'll update the description of the PR to better explain the term. I am not sure this term will end up in the end-user documentation, but I'll keep your comment in mind!

DougBurke:
@olaurino - documentation wise, I thin we just have a `DEVELOPER.md` - or some other file - listing things for now. We don't yet have a set of sphinx docs - I'm waiting for the build/infrastructure improvements and things like #24 and #25 to land before getting the sphinx stuff up and running.

cdeil:
> So please let me know if you think that at least a subset of those original tests should be kept. Maybe we should keep the pyfits/no-pyfits tests and ditch the others?

@olaurino – I don't know which builds would be most useful to keep, i.e. where new issues typically creep in and are only noticed later. So do whatever you think is useful. My point was only that you don't need to worry about a few more travis-ci builds for a repo like this one which doesn't trigger loads of builds per day.

DougBurke:
@olaurino - I got two errors - shown below - when I had no I/O module loaded (I have astropy but not pyfits) and with the `sherpa-test-data` submodule installed. It appears to be because I went straight to `python setup.py test` rather than first calling `... develop`.

I'm using 960f2ea

```
% python -c 'impor sherpa; print(sherpa.__version__)'
4.7+416.g960f2ea
```

Details
-------

```
% git clone https://github.com/sherpa/sherpa sherpa-pr31
% cd sherpa-pr31
% git submodule init
% git submodule update
% conda create -n sherpa-pr31 matplotlib ipython astropy
% source activate sherpa-pr31
% python setup.py test
...
sherpa/utils/tests/test_utils.py::test_utils::test_print_fields PASSED

==================================== ERRORS ====================================
_______ ERROR collecting sherpa/astro/datastack/tests/test_datastack.py ________
sherpa/astro/datastack/tests/test_datastack.py:26: in <module>
    from sherpa.astro.datastack import *
sherpa/astro/datastack/__init__.py:46: in <module>
    import stk
E   ImportError: No module named stk
------------------------------- Captured stdout --------------------------------
WARNING: the group module (from the CIAO tools package) is not installed.
Dynamic grouping functions will not be available.
WARNING: failed to import sherpa.astro.io; FITS I/O routines will not be available
_________________ ERROR collecting sherpa/tests/test_stack.py __________________
sherpa/tests/test_stack.py:22: in <module>
    import stk
E   ImportError: No module named stk
--------------- coverage: platform linux2, python 2.7.10-final-0 ---------------
Name                                          Stmts   Miss Branch BrMiss  Cover
...
-------------------------------------------------------------------------------
TOTAL                                         25884  17043   7309   6049    30%
=========================== short test summary info ============================
SKIP [6] /home/naridge/data/my-sherpa/sherpa-pr31/.eggs/pytest-2.7.1-py2.7.egg/_pytest/unittest.py:114: reqiured package sherpa.image.ds9_backend not available
SKIP [38] /home/naridge/data/my-sherpa/sherpa-pr31/.eggs/pytest-2.7.1-py2.7.egg/_pytest/unittest.py:114: need pycrates, pyfits
SKIP [7] /home/naridge/data/my-sherpa/sherpa-pr31/.eggs/pytest-2.7.1-py2.7.egg/_pytest/unittest.py:114: required sherpa.astro.xspec module missing
SKIP [1] /home/naridge/data/my-sherpa/sherpa-pr31/.eggs/pytest-2.7.1-py2.7.egg/_pytest/unittest.py:114: TODO: failing test used to have a different name and was never executed
=============== 218 passed, 52 skipped, 2 error in 40.54 seconds ===============
```

I then set up the developer build and re-ran the tests, and they now passed (and a few more were run and skipped):

```
% python setup.py develop
...
% python setup.py test
...
=================== 222 passed, 56 skipped in 45.44 seconds ====================
```

If I try a clean install and run `python setup.py develop` and then `python setup.py test` the tests all pass (222 pass, 56 skipped).

olaurino:
@DougBurke thanks for catching this. I redefined the `test` alias definition to include `develop`. I checked in the [setuptools documentation](https://pythonhosted.org/setuptools/setuptools.html#test) that this definition is semantically and technically consistent with the standard one, i.e. `test` builds everything locally by running `build_ext -i` and `egg_info`, which is also what `develop` does. We should be careful and make sure that these redefinitions of the standard setuptools commands are consistent and do not produce unexpected side effects or surprising behaviors.

DougBurke:
@olaurino - I can confirm that with e59b455 I can run `python setup.py test` in a clean download and it works (and with pyfits installed I now get 252 passed 26 skipped, but that's beside the point here).

valkenar:
I got this warning error on a fresh build cloned yesterday.

WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be
available

But got
----------------------------------------------------------------------
Ran 253 tests in 12.551s

OK (skipped=65)

Note that I had to run sudo python setup.py install because I encountered
this error when running as a regular user
running install_lib
creating /usr/local/lib/python2.7/dist-packages/sherpa
error: could not create '/usr/local/lib/python2.7/dist-packages/sherpa':
Permission denied

olaurino:
@valkenar The warnings are benign, and just inform the user that some of the Sherpa components will not be available because some of the requirements are missing. Installing [ds9](http://ds9.si.edu/) should get rid of the first warning. If you want to add support for `xspec` models you should follow [these instructions](https://github.com/sherpa/sherpa#xspec).

A permission denied error is also expected if you try to install Sherpa system-wide without root privileges. You may want to look at [virtualenv](https://virtualenv.pypa.io/en/latest/) for an alternative to installing python packages in system folders and keep different python environments isolated from each other.

I hope this helps.

valkenar:
Yup, that's how I interpreted the errors, the only reason I mentioned the
permission denied issue is that the instructions include the line "$ python
setup.py install" and I wanted to go ahead and run through the instructions
as-written to experience the process as a naive user. Since the XSpec
section came afterwards I chose not to configure it on the first pass and
see what would happen. For me, personally, none of it was a real problem.

The platform-specific dependency management details are probably outside
the scope of Sherpa's documentation, but if you're interested I saved the
commands I needed to install all the required and suggested dependencies on
a relatively clean Mint install. It's not tremendously useful as a script
since they're easy enough to figure out if you know what you're doing (just
a few basic apt-get and pip commands), but some of the libs (like
matplotlib) did initially fail on their own missing dependencies. Depending
on the level of detail you're going for and the intended audience it might
be worth including this extra information.
<https://github.com/sherpa/sherpa#xspec>

PR #32, opened 2015-05-29
workaround new crates API: crates now returns var-length arrays

olaurino:
#### Release Note
Fix segfault from CRATES update in 4.8b1

In 4.8b1 CRATES returns variable length arrays by default, rather than the zero-padded fixed length ones it used to return. Sherpa manipulated the arrays so to remove the zero-padding and obtaining variable length arrays. The change in the CRATES API resulted in Sherpa segfaulting when trying to manipulate the data coming from CRATES.

In the patch, we use a new API offered by CRATES to get the old-style fixed-length arrays instead of the new default ones.

In the future, we may want to update the Sherpa code to deal with the new arrays directly. Also, this patch uncovered some code duplication.

#### Tests
The patch was tested in the CIAO 4.8b1 builds where the segmentation fault occurred, and new CIAO builds have been produced with this change.

Additional tests (after commit 683f2d4) have been performed on CIAOD from June 25, 2015.

### DIFF
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -236,7 +236,7 @@ def _try_col(crate, colname, make_copy=False, fix_type=False, dtype=SherpaFloat)
     if col is None:
         return None
 
-    if col.is_varlen():
+    if hasattr(col, 'is_varlen') and col.is_varlen():
         values = col.get_fixed_length_array()
     else:
         values = col.values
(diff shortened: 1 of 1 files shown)


olaurino:
@DougBurke I thought you had a commit to add to this PR... or should I implement the changes you suggested?

DougBurke:
I wasn't planning on adding to this patch as I'm running out of time before my vacation ;-), so can you add the change I suggested above (or some variant of it) @olaurino ?

olaurino:
no problem.

olaurino:
I am rebasing this branch for convenience. Original commit (with comments from @DougBurke) is  41bd323.

olaurino:
@DougBurke if it looks good to you, as soon as Travis gives me the green light (although I would be surprised if it didn't, since it is not going to exercise the changes) I'll merge this PR.

I tested my latest commit in today's CIAOD.

DougBurke:
Looks good to me to merge (assuming Travis gives us its blessings).

PR #34, opened 2015-06-12
Bugfix: Improving the X-Spec build

DougBurke:
# Aim

Improve the build for the X-Spec module

# Issue

I have build X-Spec 12.8.2 and then used this as the libraries against which to build the X-Spec module.
Building `master` with the `xspec_config` section of `setup.cfg` set to (expanding out the HEADAS environment variable, but I'm keeping it in here to make things a bit more obvious what's going on):

```
with-xspec=True
xspec_lib_dirs=$HEADAS/lib
xspec_libraries = XSFunctions XSModel XSUtil XS wcs-4.20
cfitsio_libraries=cfitsio_3.37
ccfits_libraries=CCfits_2.4
```

results in the following failure when trying to import `sherpa.astro.xspec`:

```
ImportError: sherpa/astro/xspec/_xspec.so: undefined symbol: C_superExpCutoff
```

As a "work around", I can get the build to work if I also set the following values in `setup.cfg` (i.e. explicltly list all the directories):

```
cfitsio_lib_dirs=$HEADAS/lib
ccfits_lib_dirs=$HEADAS/lib
gfortran_lib_dirs=$HEADAS/lib
```

# Details

There are four main changes:

1. The use of `ld3` rather than `l3` when constructing `ld`. This is obviously a typo. Fixing this is only needed for those set ups where the CCfits library (which is what the `*3` settings refer to) is in a different location to any of the other libraries.

2. The removal of empty strings from the `ld`, `l`, and `inc` variables before sending to `self.distribution.ext_modules.append()`.  I believe this is what causes the problems for me above, since I did not set `cfitsio_lib_dirs`, `ccfits_lib_dirs, and `gfortran_lib_dirs`, so they were set to `''`, which resulted in a link line which included `-L -L -L`. Somewhere this lead to link problems. To fix this, either remove the empty elements (as the patch does) or require all the `*_lib_dirs` settings to be set (which I don't like). It's not clear to me whether this also needs to be done to the `l` and `inc` variables (library name and include paths), but it looks like a sensible thing to do (and I think I've seen `-I -I ...` in the build line, which this removes, but could be mistaken)

3. The addition of `wcslib_lib_dirs` and `wcslib_libraries` options. The library is now needed to build the XSPEC module, so it makes sense to follow the cfitsio and CCfits settings here. Note that in CIAO 4.7, the XSPEC code was built in such a way that a static WCSLIB library was used, so this setting is not needed when building against CIAO's version of the library.

4. pep8 fixes.

# Example

Without these changes, using the `setup.cfg` setting given above, the link line on a Linux (Ubuntu) machine was (I've replaced paths by env var names for simplicity, $anaconda is my conda installation path):

```
g++ -pthread -shared -L$anaconda/lib -Wl,-rpath$anaconda/lib,--no-as-needed \
  build/temp.linux-x86_64-2.7/sherpa/astro/xspec/src/_xspec.o \
  -L$HEADAS/lib -L -LCCfits_2.4 -L \
  -L$anaconda/lib -Wl,-R$HEADAS/lib \
  -Wl,-R -Wl,-RCCfits_2.4 -Wl,-R \
  -lXSFunctions -lXSModel -lXSUtil -lXS \
  -lwcs-4.20 -lcfitsio_3.37 -lCCfits_2.4 -lgfortran \
  -lpython2.7 -o sherpa/astro/xspec/_xspec.so
```

You can see that there is `-L$HEADAS_/lib -L -LCCfits_2.4 -L ...`, which comes out because empty elements are not removed from the link line (the `-L ` entries) and `-LCCfits_2.4` because of the type (`l3` instead of `ld3`). You can see that this is repeated in the `-Wl,-R` sections. Something in this setup causes the resulting library (_xspec.so) to not find the `C_superExpCutoff` symbol which is defined (in `libXSFunctions.so`).

With the changes (this was written before I added the `wcslib_lib_dirs/wcslib_libraries` option, so I had just added `wcs-4.20` to the end of the `xspec_lib_dirs` options), the link line becomes

```
g++ -pthread -shared -L$anaconda/lib -Wl,-rpath=$anaconda/lib,--no-as-needed \
  build/temp.linux-x86_64-2.7/sherpa/astro/xspec/src/_xspec.o \
  -L$HEADAS/lib -L$anaconda/lib \
  -Wl,-R$HEADAS/lib \
  -lXSFunctions -lXSModel -lXSUtil -lXS \
  -lwcs-4.20 -lcfitsio_3.37 -lCCfits_2.4 -lgfortran \
  -lpython2.7 -o sherpa/astro/xspec/_xspec.so
```

# Notes

If merged into  `master` and `python setup.py test` run after `git submodule init ; git submodule update`, then one test will fail due to a missing file. This is a known issue with the test suite and not to do with this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,32 +327,58 @@ Sherpa does not support
 default. However, it is possible to instruct Sherpa to build its
 `XSPEC` extension module by changing the build configuration options.
 
-There are three ways to build the extension:
-
- 1. build the full XSpec system;
-
- 2. use the model-only build of XSpec, which will also require
-    building the cfitsio, CCfits, and WCS libraries;
- 
- 3. or point to the X-Spec libraries provided by
-    [CIAO](http://cxc.harvard.edu/ciao/).
-
-In all cases, the same version of `gfortran` should be used to
-build Sherpa and X-Spec, to avoid possible incompatabilities.
-
 The `xspec_config` section of the `setup.cfg` file will need
 changing to point to the libraries, and to turn on the extension.
-An example, for the case when a full XSpec build was made, is:
+In all cases, set
 
     with-xspec=True
-    xspec_lib_dirs=$HEADAS/lib
-    xspec_libraries=XSFunctions XSModel XSUtil XS wcs-4.20
-    cfitsio_libraries=cfitsio_3.37
-    ccfits_libraries=CCfits_2.4
 
-The environment variable `$HEADAS` should be expanded out, and the
-version numbers of the `wcs`, `cfitsio`, and `CCfits` libraries
-may need to be changed.
+The remaining settings depend on how the XSPEC libraries have
+been built (in the examples below, environment variables are
+used, but the full path should be in your own copy of the file):
+
+ 1. If the full XSPEC system has been built, then use
+
+        xspec_lib_dirs=$HEADAS/lib
+        xspec_libraries=XSFunctions XSModel XSUtil XS wcs-4.20
+        cfitsio_libraries=cfitsio_3.37
+        ccfits_libraries=CCfits_2.4
+
+    The environment variable `$HEADAS` should be expanded out, and the
+    version numbers of the `wcs`, `cfitsio`, and `CCfits` libraries
+    may need to be changed.
+
+ 2. Use the model-only build of XSPEC, which will also require
+    building the
+    [cfitsio](http://heasarc.gsfc.nasa.gov/docs/software/fitsio/fitsio.html),
+    [CCfits](http://heasarc.gsfc.nasa.gov/docs/software/fitsio/ccfits/),
+    and
+    [WCSLIB](http://www.atnf.csiro.au/people/mcalabre/WCS/wcslib/)
+    libraries (it is not clear if version 5 is supported, since
+    XSPEC 12.8.2 uses version 4.20). If all the libraries are installed
+    into the same location ($HEADAS/lib), then a similar set up to the
+    full XSPEC build is used
+
+        xspec_lib_dirs=$HEADAS/lib
+        xspec_libraries=XSFunctions XSModel XSUtil XS wcs
+
+    except that the library names (`cfitiso`, `CCfits`, and
+    `wcs`) do not need version numbers. If placed in different
+    directories then the `cfitsio_lib_dirs`, `ccfits_lib_dirs`,
+    and (possibly) `gfortran_lib_dirs` values should be set
+    appropriately.
+
+ 3. or point to the XSPEC libraries provided by
+    [CIAO](http://cxc.harvard.edu/ciao/). In this case the
+    `wcs` library does not need to be specified because of
+    the way the XSPEC models-only version was built with
+    CIAO 4.7.
+
+        xspec_lib_dirs=$ASCDS_INSTALL/ots/lib
+        xspec_libraries=XSFunctions XSModel XSUtil XS
+
+In all cases, the same version of `gfortran` should be used to
+build Sherpa and XSPEC, to avoid possible incompatabilities.
 
 If there are problems building, or using, the module, then the other
 options may need to be set - in particular the `gfortran_lib_dirs` and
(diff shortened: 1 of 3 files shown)


olaurino:
@DougBurke I can only do limited testing by making the build point to the CIAO X-Spec libraries. For convenience, I suggest that we rebase this PR, because a) this branch does not contain the new test infrastructure, and b) the new sherpa-test-data fixes the missing file issue you mentioned in your notes.

After I am done testing, and as long as mighty Travis gives its blessings, I'll merge this PR.

We need to be able to reliably build X-Spec in order to include it in the Travis build. Could you please share the commands you use to build it? I would like to create a reproducible provisioning script for virtual workers that would work on systems similar than the ones used by Travis (I am assuming you built X-Spec on Ubuntu, that's already close enough), so that I can then include the X-Spec build in the Travis configuration.

DougBurke:
There are three ways to build the XSpec module

a) against a full XSpec build
b) using the models-only XSpec distribution
c) against the CIAO binaries

I am going to concentrate on a here as I had problems with both b and c.

## a) Full Xspec build

Use the [HEASOFT download page](http://heasarc.gsfc.nasa.gov/lheasoft/download.html), where you select 'Source Code' and then XSpec (under the Xanadu) link [you can unselect the  HEAGen package that gets selected] - and apply the latest set of patches (as of today it is 12.8.2q) from the [issues page](https://heasarc.gsfc.nasa.gov/xanadu/xspec/issues/issues.html).

I have not tried the binary downloads (and as we will want to apply patches I think the source code version is the one to use).

The build is basically (going from memory here, so check the install docs, in particular for what directory to be in when you run these comands):

    ./configure --prefix=/code/xspec
    make
    make install

which will place the libraries we are interested in in `/code/xspec/<some long name>/lib/`, and we need to set the `HEADAS` environment variable to `/code/xspec/<some long name>` for the Xspec module to be useful (but not needed during the build).

The `setup.cfg` file then gets set to, in the `xspec_config` section,

    with-xspec=True
    xspec_lib_dirs=/code/xspec/<some long name>/lib
    cfitsio_libraries=cfitsio_3.37
    ccfits_libraries=CCfits_2.4
    wcs_libraries=wcs-4.20

I suggest leaving `gfortran_lib_dirs` and `gfortran_libraries` blank, and then set them if needed. Note that you should check in `$HEADAS/lib` to see if the version numbers on cfitsio/CCfits/wcs need changing or not.

## b) models-only build

I have not worked out how to apply the patches to this build. It's also more work since you have to manually download `cfitsio`, `CCfits`, and `WCSLIB` (which is now up to version 5.x; this does not seem to be a problem for XSpec but I have done limited testing).

The source download is from the [HEASOFT download page](http://heasarc.gsfc.nasa.gov/lheasoft/download.html) but this time look for the "Xspec models-only source code distribution" link in the "XSpec notes" section below the main download area. The build appears to be the same as for XSpec proper.

When I did this I installed cfitsio, CCfits, and WCSLIB after installing the xspec model-only build, so that I could install them into the same directory as the XSpec models (ie `<whatever your prefix>/<long dir name>/lib/`).

The `setup.cfg` for Sherpa is not that much different, except that you should only have to give the

    with-xspec=True
    xspec_lib_dirs=/code/xspec/<some long name>/lib

options, since you will have presumably installed cfitsio/CCfits/WCSLIB using their default names

## c) CIAO build

From a and b you can probably guess that for a build-against-CIAO build I suggest setting `xspec_lib_dirs` to `$ASCDS_INSTALL/ots/lib`, but this time you probably want to set `gfortran_lib_dirs` to this too. I then hit version mis-match errors in my on-disk gfortran vs the one used for CIAO 4.7 so didn't really explore any more.

# Set up for importing the module

Before starting the python session (can actually do this wth an `os.environ` call before importing any sherpa model) make sure you

    export HEADAS=/code/xspec/<some long name>

and then you should be able to either

    from sherpa.astro import ui

or

    from sherpa.astro import xspec
    print(xspec.get_xsversion())
    mdl = xspec.XSapec()
    print(mdl([0.1,0.2,0.3,0.4])

which exercises the module a bit. It appears that you *do not* need to set up the full HEADAS environment (i.e. `source $HEADAS/headas-init.sh`) before using the module, but I cannot guarantee that there isn't something in the full setup that is needed (I think unlikely, but wanted to provide a warning).

# What files are needed?

I would imagine you will have to copy everything that gets installed into `/code/xspec/` (or wherever you install Xspec) and make some tarfile/other system to copy these files over to the travis instance. It's not just the lib files but also the files in the `spectral/` directory, and I htink it's just safest to take everything. This is why I suggest removing the `HEAGen` option when selecting the source code to download, to reduce the amount of irrelevant stuff in the install directory.

We do *not* want to be building XSpec along with Sherpa every time there's a travis build kicked off!

PR #36, opened 2015-06-14
Typo: reqiured

DougBurke:
This only appears when a test is skipped, so has no functional impact.


anetasie:
I just reviewed the diffs and this looks good. It is ready to merge.

PR #44, opened 2015-06-16
BugFix: save_quality output column name

DougBurke:
#### Release Note
save_quality now correctly outputs 'QUALITY' as the column name, instead of 'GROUPS'

#### Description
This fixes what looks to be a cut-and-paste error from `save_grouping` (the column name was 'GROUPS' when it should be 'QUALITY'). It was noted in PR #17.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -4417,8 +4417,7 @@ class Session(sherpa.ui.utils.Session):
         respectively. The remaining parameters are expected to be
         given as named arguments.
 
-        The quality column is labelled as ``GROUPS`` rather than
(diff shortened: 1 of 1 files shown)


DougBurke:
Note that the fact that the Travis CI checks passed here show that we don't have a test for the column names created by `save_quality` (or, presumably, related functions). I don't want this particular patch slowed down by the need to add in a full test suite for this function.

olaurino:
@dtnguyen2 do you think I can merge this PR?

dtnguyen2:
I checked it, and you can merge the PR

dtnguyen2:
Yes

PR #45, opened 2015-06-16
BugFix: list_priors to return a dict

DougBurke:
Returning a string copy of the dictionary is not useful, and I claim it to be a bug. This was noted in PR #17 .

### DIFF
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -1,5 +1,5 @@
 # 
-#  Copyright (C) 2011  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2011, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -113,16 +113,15 @@ class MCMC(NoNewAttributesAfterInit):
         self.__dict__.update(state)
 
 
-    ### DOC-TODO: include examples once this returns something useful
-    ### TODO: this should really return a dict, not a string
+    ### DOC-TODO: include examples
     def list_priors(self):
         """Return the priors set for model parameters, if any.
 
         Returns
         -------
-        priors : string
-           A string representation of the dictionary mapping between
-           parameters (keys) and priot functions (values).
+        priors : dict
+           The dictionary of mappings between parameters (keys)
+           and prior functions (values) created by `set_prior`.
 
         See Also
         --------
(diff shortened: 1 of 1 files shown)


olaurino:
@dtnguyen2, @anetasie Do you think there was a reason for returning a string instead of a dictionary?

Test coverage information shows that the list_priors method is exercised, however the value is only read and nothing is done with the value returned, see:
https://github.com/sherpa/sherpa/blob/0daa0d2cf4f126e043c0f9832642e4ec2e7347f1/sherpa/sim/tests/test_sim.py#L200

This seems to be an API function that just provides a convenient accessor. Maybe this is the reason why it returned a string? You would access the dictionary with the pythonic `mcmc.priors` and would get a string with `mcmc.list_priors()`?

@DougBurke am I missing something?

DougBurke:
I have no idea why it returned a string, as it is not a convenient way to
access this information (since it's a dictionary)!

olaurino:
Convenient in the sense that you can get the actual dictionary by usual python means (`mcmc.priors`) but you can use `mcmc.list_priors` for getting a "pretty" listing (except that `str(mcmc.priors)` is not that pretty or convenient). With your change you now have two ways for getting the same dictionary. Maybe the method was meant to return a string, and it should just return something "prettier" than `str()`?

DougBurke:
I believe this is a "high-level" function to match the other ones - i.e.
something that sits at the `sherpa.astro.ui` layer. Yes you can get at this
information going to the object API layer, but that's what these
convenience functions in the "ui" layer are. I have actually used this
function, and there is no way that I believe a string representation is
what people want.

olaurino:
Fair enough... I would like @dtnguyen2 and @anetasie to comment, then I'll be happy to merge it.

anetasie:
The function is to get a listing and the pretty output is nice for the visual listing.
The original function was only listing the priors and was not used in any other part of the code.
One can think about ways to actually use the output when setting up the priors for complex models.

olaurino:
@anetasie does it mean that you would close this PR without merging it?

anetasie:
I do not see the difference in the screen output with this change. I think we should merge, so we have consistency with the ui functions, as @DougBurke said.

olaurino:
@anetasie you don't see any difference because the current implementation of `list_priors` just returns whatever the implementation of `dict.__str__` is, which is what a `print` call would do.

However, in the current implementation you cannot manipulate what comes back from `list_priors` (it's a string) while in Doug's patch the method returns an actual dictionary.

I am getting contradictory inputs (the `list_priors` function is supposed to be returning a "pretty" representation of the string, but you are also fine with it returning the raw dictionary itself).

I'll merge it anyway as the method does not seem to be used anywhere and it is easy to go back.

anetasie:
@olaurino It would be good to have a nice printout, but that's a different issue.
Having a useful dictionary makes sense.

PR #48, opened 2015-06-17
improvements to plot labels

DougBurke:
# Release Note

Fix up several issues seen in plot labels - titles and Y-axis labels - for commands such as `sherpa.ui.plot_data`, `sherpa.ui.plot_fit_resid`, and `sherpa.ui.plot_chisqr`:

1. Plot titles when the `name` attribute of the data set is empty
2. Y axis label for residual plots (non PHA data sets)
3. `plot_chisqr` title and Y axis label takes advantage of LaTeX support using the matplotlib backend.

# Details

There are three changes:

1. Plot titles when the `name` attribute of the data set is empty
2. Y axis label for residual plots (non PHA data sets)
3. `plot_chisqr` title and Y axis label takes advantage of LaTeX support using the matplotlib backend.

##  1. Plot titles when the `name` attribute of the data set is empty

For a dataset created with `sherpa.ui.load_arrays`, it is likely that the `name` attribute of the data structure will be empty. This led to plot titles like 'Residuals of - Model', since the code expected the `name` to not be empty. I have changed the title code so that:

  - the logic is in the `_make_title` routine, and use the `.format` method of strings rather than
    the old-style string interpolation (although there's a lot of that left in the rest of the code)
  - the `name` attribute is only added to the title if it is not empty
  - the titles are more consistent in appearance, so that it's now "\<title> for \<name>" rather than some 
    saying "\<title> of \<name>" or having the name in the middle of the string.

This code is independent of the plotting back end.

## 2. Y axis label for residual plots (non PHA data sets)

The old behaviour for the Y axis label of residual plots (those created by `sherpa.ui.plot_resid`) was to use `y` as the label. I find this confusing, particularly when using `sherpa.ui.plot_fit_resid()`, so switched it to `Data - Model`. This better matches the existing behaviour of `sherpa.ui.plot_ratio()`, which uses the label 'Data / Model'.

This code is independent of the plotting back end.

**Note:** for PHA datasets (i.e. `DataPHA` objects), the y label is changed to be the units displayed for that axis (e.g. `photon/cm^2/s/keV`). I have left this in, rather than change it to `Data - Model`. This should be reviewed.

## 3. `plot_chisqr` title and Y axis label takes advantage of LaTeX support using the matplotlib backend.

ChIPS and matplotlib use different approaches to displaying LaTeX: ChIPS does it automatically whereas matplotlib requires the LaTeX be surrounded by '$'. The only plot where this matters is `sherpa.ui.plot_chisqr()`, which uses a title and Y axis label of `'\chi^2'`. I have a simple wrapper which
converts this to matplotlib format if this backend is in use.

**Note:** there are other places where this could be relevant or important (e.g. the `photom/cm^2/s/keV` label mentioned in point 2 above), but it's not clear how to handle this given the current design of the plots. I would consider this a later RFE for potential improvements to the plotting (along with fixes to simplify the note from point 2), and not be a part of this PR.

# Example

The following code can be used to compare the behaviour before and after the PR (run from the top-level of the sherpa distribution). The comments, when talking about comparing plots, refers to the plots with the pre- and post-PR changes in them.

```
from sherpa.astro import ui

ui.load_arrays(1, [11,22,33], [4,6,5])
ui.set_source(ui.const1d.m1)

# compare the title and the Y axis label
ui.plot_resid()

# compare the Y axis labels
ui.plot_fit_resid()

# compare the titles
ui.plot_delchi()
ui.plot_ratio()

# with the matplotlib back end, the title and Y axis now use the LaTeX support
ui.plot_chisqr()


ui.load_pha(1, 'sherpa/astro/datastack/tests/data/3c273.pi')
ui.set_source(ui.powlaw1d.m1)

# compare y axes (they should be the same)
ui.plot_resid()
ui.plot_fit_resid()
ui.plot_delchi()
```

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -58,6 +58,7 @@ except:
     warning('failed to import sherpa.plot.%s;' % plot_opt +
             ' plotting routines will not be available')
     import dummy_backend as backend
+    plot_opt = 'dummy_backend'
 
 backend.init()
 
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -1202,6 +1202,14 @@ class ResidPlot(ModelPlot):
         else:
             self.yerr = data.get_yerr(True,stat.calc_staterror)
 
+        # Some data sets (e.g. DataPHA, which shows the units) have a y
+        # label that could (should?) be displayed (or added to the label).
+        # To avoid a change in behavior, the label is only changed if
+        # the "generic" Y axis label is used. To be reviewed.
+        #
+        if self.ylabel == 'y':
+            self.ylabel = 'Data - Model'
+
         self.title = _make_title('Residuals', data.name)
 
     def plot(self, overplot=False, clearwindow=True):
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -78,6 +78,7 @@ _stats_noerr = ('cash', 'cstat', 'leastsq')
 begin = backend.begin
 end = backend.end
 exceptions = backend.exceptions
+get_latex_for_string = backend.get_latex_for_string
 
 def _make_title(title, name=''):
     """Return the plot title to use.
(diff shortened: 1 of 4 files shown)


anetasie:
I have reviewed and tested this PR. It looks good. 

There was a typo in your write-up example, but I guess we are not going to use that as a test script, so this is just for the record.

This is a correct spelling:
ui.set_source(ui.powlawd1.m1)

olaurino:
I'll try to look at this today so that one DS developer can give the thumbs up for merging the PR.

olaurino:
@DougBurke I would try and avoid having conditionals that probe the backend in use: if we added a new backend we should go after all the conditionals that probe the backend in use, and add a new branch. Same would be true if the backends implementations changed their behavior.

Instead, I would add methods to the backends themselves, and have different implementations for each of them. In this case, I would consider adding a method like `get_latex_for_string(txt)` in the backends.

While a change to the backends API should be reviewed more carefully by the team, I believe it would make things more future-proof.

Does it sound reasonable?

DougBurke:
I thought about several things for the latex support - e.g. a class that could be used to create a string, sub-classed by the back-end as appropriate - and thought this was the easiest. I do not object to the logic being moved into the back-end (I wouldn't want to spend too much time on engineering an approach as it's one of those things that really depends on the backends in use). I'll try and re-work the patch.

DougBurke:
@olaurino - was this what you were thinking of?

olaurino:
@DougBurke It is! Although as I mentioned, we need to be more careful with changes to the backends API, so I'll ask other team members to review this before merging it, if you don't mind. I don't want it to take too long, either.

anetasie:
@olaurino I'm ok with this change.

PR #51, opened 2015-06-23
add missing acis model file in sherpa-test-data

olaurino:
This PR simply updates the `sherpa-test-data` git submodule to a new state that includes a test file that was missing.

The actual change is located in sherpa/sherpa-test-data@7b0e89debb38195c8a6a68ca10708d0e5581ff39


DougBurke:
I've tested this and see the fixed test (only visible when the XSPEC module is built). Ready for merge.

We should update the `README.md` file to mention the "new" way of building/testing - i.e.

    python setup.py develop
    python setup.py test

and that more tests can be run with 

    git submodule init
    git submodule update

and that `sherpa_test` is still available, but this is really a different PR.

PR #57, opened 2015-06-30
Fix #9 autopep8 --select E101,W191 --in-place --recursive .

### LINKED ISSUE #9: Don't use tabs for indentation
Apologies for bringing up a trivial and annoying issue.

In my editor (PyCharm) some Sherpa code looks mis-indented:
screen shot 2015-04-23 at 16 54 44

This is how it should look:
https://github.com/sherpa/sherpa/blob/c9f412a3c3d0a193ba5a65027f7630f058fe41a5/sherpa/models/parameter.py#L243

The PEP8 recommendation is to use 4 spaces and never tabs.
Are you willing to change this?

Would you prefer I make a pull request or do this yourself?
(It's easy to do in batch mode e.g. with the `autopep8` tool or a Python editor and then only review the diff.)

olaurino:
This PR fixes the following two PEP8 issues:
  - W191	indentation contains tabs
  - E101	indentation contains mixed spaces and tabs

I also made a pass with E111 (indentation is not a multiple of four) but that did not result in any changes.


olaurino:
@taldcroft @cdeil please let me know if you think this is enough or you would like me to apply other pep8 fixes, but I think I am happy with fixing the indentation issues.

I don't think the line ending is a problem (other than the one issue we already fixed), since git is rather good at dealing with them seamlessly. I am currently developing on all the three main platforms (Windows, Linux, and OSX) and I never had any issues, as long as git is properly configured. See for instance the quote below, taken from [The git pro book](https://git-scm.com/book/tr/v2/Customizing-Git-Git-Configuration).

If we really wanted to normalize the line endings I would rather use native git features rather than some shell tricks. I am going to create a new PR for the line ending normalization, just in case.

>core.autocrlf
>If you’re programming on Windows and working with people who are not (or vice-versa), you’ll probably run into line-ending issues at some point. This is because Windows uses both a carriage-return character and a linefeed character for newlines in its files, whereas Mac and Linux systems use only the linefeed character. This is a subtle but incredibly annoying fact of cross-platform work; many editors on Windows silently replace existing LF-style line endings with CRLF, or insert both line-ending characters when the user hits the enter key.

>Git can handle this by auto-converting CRLF line endings into LF when you add a file to the index, and vice versa when it checks out code onto your filesystem. You can turn on this functionality with the core.autocrlf setting. If you’re on a Windows machine, set it to true – this converts LF endings into CRLF when you check out code:

>    $ git config --global core.autocrlf true
>If you’re on a Linux or Mac system that uses LF line endings, then you don’t want Git to automatically convert them when you check out files; however, if a file with CRLF endings accidentally gets introduced, then you may want Git to fix it. You can tell Git to convert CRLF to LF on commit but not the other way around by setting core.autocrlf to input:

>    $ git config --global core.autocrlf input
This setup should leave you with CRLF endings in Windows checkouts, but LF endings on Mac and Linux systems and in the repository.

>If you’re a Windows programmer doing a Windows-only project, then you can turn off this functionality, recording the carriage returns in the repository by setting the config value to false:

>$ git config --global core.autocrlf false
~~~~

cdeil:
From my side getting rid of the tabs and what looks like mis-indented code in some editors is the most important thing.

Concerning the line ending ... we just started talking about that because the diffs on Github looked incorrect. I did mention this to Github at the time (~ 1 month ago) and they agreed that it's a bug in the software they use to create the diffs and will look into a fix. I haven't heard something since or checked again, but it's probably something that can be ignored and left as-is in the Sherpa codebase.

@olaurino – Thanks!

DougBurke:
I've reviewed this and also tried to test it  - with astropy, matplotlib, Xspec, and ds9 available, so that 278 tests are run, 1 skipped - and the tests pass. I say merge it.

PR #58, opened 2015-07-01
Line ending normalization

olaurino:
This PR normalizes the line-ending characters on the few files that would require it.

For convenience, I based the commit off the autopep8 changes.

I followed the [git documentation for `.gitattributes`](http://git-scm.com/docs/gitattributes):

~~~~
$ echo "* text=auto" >>.gitattributes
$ rm .git/index     # Remove the index to force Git to
$ git reset         # re-scan the working directory
$ git status        # Show files that will be normalized
$ git add -u
$ git add .gitattributes
$ git commit -m "Introduce end-of-line normalization"
~~~~

I am not sure whether it makes sense to actually merge this one, but it spawned from the conversation on issue #9.


olaurino:
I don't think it makes sense to merge this one.

the only two files affected are a file generated by ipython notebook/jupyter, and the other one was incorporated from an external source.

Plus, the original issue that originated this change was fixed by a previous commit.

Issue #30, opened 2015-05-29
Warning message about numpy.sqrt seen when running sherpa_test


DougBurke:
It appears that, if you have numpy 1.9.2 installed, `sherpa_test` will produce the following on-screen warning (originally discussed as part of PR #29 but has been pulled out into a separate issue). I don't think anything is wrong here, but

 - it's visually ugly
 - I can't explain the change in behavior, which makes me nervous

```
.................................ssssssssssssssssssssssssssss.......sss..ssssssssssss./home/naridge/data/my-sherpa/sherpa-omar/sherpa/estmethods/__init__.py:369: RuntimeWarning: invalid value encountered in sqrt
  diag = (sigma * numpy.sqrt(inv_info)).diagonal()
....sssssssssss..........................................................ssssss...........................sssss.sssssss.......................................................
----------------------------------------------------------------------
```

This does not happen if numpy 1.8.1 is installed.

Running `sherpa_test -v 3` shows that the message is from `test_covar`

```
test_covar (sherpa.estmethods.tests.test_estmethods.test_estmethods)/home/naridge/data/my-sherpa/sherpa-omar/sherpa/estmethods/__init__.py:369: RuntimeWarning: invalid value encountered in sqrt
  diag = (sigma * numpy.sqrt(inv_info)).diagonal()
 ... ok
```

However, I get this warning from both versions of numpy when running `sqrt` directly, as shown below, so it's not clear to me what is causing this message to appear. A quick look at `test_estmethods.py` doesn't suggest that the warning is being "turned off" in any way - I haven't looked to see if `numpy.seterr` allows this - but maybe there's some change in behavior in one of the earlier tests that means that with numpy 1.9 the message hasn't already been created, whereas it is with numpy 1.8 (in a place where the message is "hidden"); this is all speculation.

```
% python
Python 2.7.9 |Continuum Analytics, Inc.| (default, Apr 14 2015, 12:54:25) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
>>> import numpy as np
>>> np.version.version
'1.9.2'
>>> np.sqrt([1,2,3])
array([ 1.        ,  1.41421356,  1.73205081])
>>> np.sqrt([1,-2,3])
__main__:1: RuntimeWarning: invalid value encountered in sqrt
[...]

Issue #38, opened 2015-06-15
Importing Bugs #13545 - grouping twice gives an IndexError exception


anetasie:
This is a bug imported from an older bug tracking system.

The bug #13545 shows up in the current version in both CIAO and standalone.
Grouping twice gives an IndexError.

Here is the current output in my local version.

~~~~ python
(sherpa-test)13545 $ ipython
Python 2.7.10 |Continuum Analytics, Inc.| (default, May 28 2015, 17:04:42) 
Type "copyright", "credits" or "license" for more information.

IPython 3.1.0 -- An enhanced Interactive Python.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: from sherpa.astro.ui import *
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available

In [2]: load_data("xte.pi")
read ARF file xte.warf
read RMF file xte.wrmf
read ARF (background) file xte_bkg.warf
read RMF (background) file xte_bkg.wrmf
read background file xte_bkg.pi

In [3]: notice(0.3,2)

In [4]: show_data()
Data Set: 1
Filter: 0.2993-1.9929 Energy (keV)
Bkg Scale: 0.332619
Noticed Channels: 21-137
name           = xte.pi
channel        = Float64[1024]
counts         = Float64[1024]
staterror      = None
syserror       = None
bin_lo         = None
bin_hi         = None
grouping       = None
quality        = None
exposure       = 18513.2782715
backscal       = 2.18391968294e-07
areascal       = 1.0
grouped        = False
subtracted     = False
units          = energy
rate           = True
plot_fac       = 0
response_ids   = [1]
background_ids = [1]

RMF Data Set: 1:1
name     = xte.wrmf
detchans = 1024
energ_lo = Float64[1070]
energ_hi = Float64[1070]
n_grp    = UInt64[1070]
f_chan   = UInt64[1354]
n_chan   = UInt64[1354]
matrix   = Float64[382802]
offset   = 1
e_min    = Float64[1024]
e_max    = Float64[1024]

[...]

PR #60, opened 2015-07-08
clean up logging level change

olaurino:
This PR simply makes sure that one test case I stumbled upon while investigating issue #38 cleans up the logging level after changing it to start the tests. Might this be the origin of some logging weirdness @DougBurke experienced, e.g. issue #30?


DougBurke:
Looks good to me: I say merge it.

It's hard to review if this fixes #30 since this is no longer triggered with the new test suite, but the proposed change makes sense.

PR #61, opened 2015-07-08
add astropy build

olaurino:
This PR changes the automatic build settings to include two builds: one with `pyfits` and one with `astropy`. After Travis CI successfully completed the tests on the push build, I visually inspected the build log to double-check that the correct dependencies were being installed, and that the tests involving FITS files were correctly run and not skipped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ python:
   - "2.7"
 
 env:
-#  - PYFITS=false INSTALL_TYPE=develop TEST=submodule
-  - PYFITS=true INSTALL_TYPE=develop TEST=submodule
-#  - PYFITS=false INSTALL_TYPE=install TEST=package
-#  - PYFITS=true INSTALL_TYPE=install TEST=package
-#  - PYFITS=true INSTALL_TYPE=install TEST=smoke
+  - FITS="astropy" INSTALL_TYPE=develop TEST=submodule
+  - FITS="pyfits" INSTALL_TYPE=develop TEST=submodule
+#  - INSTALL_TYPE=install TEST=package
+#  - FITS="astropy" INSTALL_TYPE=install TEST=package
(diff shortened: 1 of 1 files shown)


DougBurke:
@olaurino - would it be possible to move the conda install of $FITS to the main conda install line: i.e. so that it's something like

```
conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy matplotlib $FITS
```

rather than install it separately later?

The reason for this is to make sure that the dependencies match; for instance, to make sure that we pick up the matplotlib built against numpy 1.8 (as used by pyfits) which I think is an older version than the one we get with numpy 1.9. I believe, from trying things out, that it's not actually a problem at the moment (i.e. the different versions work well together), but it's nice to get a "free" test of different package versions.

olaurino:
@DougBurke I am fine with the change you are suggesting, but I would like to make sure we understand the reason why we are doing it, as in principle there should be no difference (in practice we know there is).

Conda could do a much better job at automatically managing the versions, but it does provide user tools for such management.

If the idea is having control over the versions, we could have a build matrix where we control each individual version to whatever combinations we decide to explicitly support. I would still like to have one or two builds where a) we use the latest defaults (as you are suggesting) and b) we try to break things by forcing builds like the one introduced by adebaec: installing pyfits/astropy separately may potentially break the build, but it is also something the users might do, right?

olaurino:
@DougBurke we left this up in the air. As I made more changes to the travis configuration, what if I merge this one as is, and then we review the builds configuration when we add the actual package versions?

I am also concerned that we are ending up writing the build matrix by hand, so a review is necessary at some point anyway.

DougBurke:
I was waiting to discuss this at out telecon, as I kept on meaning to
discuss this, but I'm happy to merge this now and tweak things later.

Issue #56, opened 2015-06-26
Importing Bugs #13924 - SegFault with fake_pha


anetasie:
SegFault with fake_pha on some systems was reported.
The original bug and the data are located here:
/data/lenin2/Bugs/Sherpa/Nustar/HD_15934 

and the script 'test2.sph' fails on MAC.

This bug is specific to X-ray spectral analysis. 
In CIAO version the screen output when running on the command line shows the following:

~~~~
% sherpa 
----------------------------------------------------- 
Welcome to Sherpa: CXC's Modeling and Fitting Package 
----------------------------------------------------- 
CIAO 4.6 Sherpa version 2 Wednesday, January 22, 2014 
sherpa-1> set_source("faked", "xsapec.A") 
sherpa-2> fake_pha("faked", arf="tvcol_A.arf", rmf="tvcol_A.rmf", exposure=1.e8, 
grouped=False) 
/soft/ciao-4.6/binexe/sherpa: line 199: 28184 Segmentation fault (core 
dumped) ipython --profile sherpa 
~~~~

olaurino:
Might this be related to issue #55, in particular [this comment](https://github.com/sherpa/sherpa/issues/55#issuecomment-119374222) from @DougBurke?

DougBurke:
The data files (tvcol_A.arf and tvcol_A.rmf) plus an example script are available at https://gist.github.com/DougBurke/04a41831a70e7a59d3e4

DougBurke:
So, it appears that this is related to #62. The problem here is that the ARF grid is not contiguous, that is the `energ_hi` value for row `i` is not always the same as the `energ_lo` column for row `i+1`. It appears that this makes Sherpa send in a non-contiguous grid of lo/hi values to the models. Since the apec model is in use here, it triggers #62.

Using the data from the gist at https://gist.github.com/DougBurke/04a41831a70e7a59d3e4 and the latest master branch (so using astropy for file I/O, but the same result is seen in CIAO 4.7 with crates as the I/O back end):

```
In [1]: import numpy as np

In [2]: from sherpa.astro import ui
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'

In [3]: arf = ui.unpack_arf('tvcol_A.arf')

In [4]: elo = arf.energ_lo

In [5]: ehi = arf.energ_hi

In [6]: elo.size
Out[6]: 4096
[...]

PR #63, opened 2015-07-09
Fix and clean up of the interface to XSpec models

### LINKED ISSUE #62: XSpec model evaluation for non-continuous grids (Xspec models that call sumdem, such as apec, can crash)
# Summary

XSpec models that call the SUMDEM routine - e.g. `xsapec` but many others - can cause a crash if called with only one bin, as shown below ("XSpec model crash" section). This has been reported to the XSpec team, but is unlikely to be a major problem for Xspec. It is a *serious* issue for Sherpa than XSpec due to the way that the wrapper code that interfaces to the XSpec model library is written; in particular, how it handles being given a non-continuous grid ("Sherpa model crash" and "What causes the crash (in Sherpa)?" sections).

It is not clear to me how many of the existing issues are affected by this: #55 and #56 may be but I'm not convinced that #42 is, because the test code isn't written to test out this behavior (and so doesn't show it).

# Plan

I am going to update the xspec test code to make sure it compares the results from  model comparisons run with a single grid (contiguous) and xlo,xhi grids (made to be non-contiguos) and compare the overlap. I've already done this in off-to-the-side work and it shows the crash behavior for some models, as well as other differences which I believe to be down to the fact that many models involve interpolating between grids, so if you run with different grids you can get different results.

Once the test suite has been improved I plan to change the wrapper code so that it deals with non-contiguous grids by only calling the XSpec function once, with extra bins added to make it contiguous, and then remove them on output. Unless someone wants to do this before me ;-)

# Details

## XSpec model crash

```
% uname -a
Linux madeupname 3.19.0-21-generic #21-Ubuntu SMP Sun Jun 14 18:31:11 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
% xspec

		XSPEC version: 12.8.2q
	Build Date/Time: Wed Jun 24 12:42:59 2015

XSPEC12>energies 0.1 0.2 1

Models will now use energy array created from:
   0.1 - 0.2   1 linear bins

XSPEC12>mo apec

              1       0.01(      0.01)      0.008      0.008         64         64
1:apec:kT>/*
Segmentation fault (core dumped)
```

## Sherpa model crash

```
% python
Python 2.7.10 |Continuum Analytics, Inc.| (default, May 28 2015, 17:02:03) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://binstar.org
>>> import sherpa
>>> sherpa.__version__
'4.7+475.g0ee48e5.dirty'
>>> xspec.get_xsversion()
'12.8.2q'
>>> mdl = xspec.XSapec()
>>> mdl([0.1,0.2,0.3,0.4,0.5])
array([ 2.01070237,  0.31376797,  0.21693133,  0.12136491,  0.        ], dtype=float32)
>>> mdl([0.1,0.2,0.3,0.4], [0.2,0.3,0.4,0.5])
array([ 2.01070237,  0.31376797,  0.21693133,  0.12136491], dtype=float32)
>>> mdl([0.1,0.3,0.4], [0.2,0.4,0.5])
Segmentation fault (core dumped)
```

So, when given `xlo` and `xhi` grids which are *not* continuous, we see a crash. 

# What causes the crash (in Sherpa)?

The Sherpa behavior is down to `sherpa/include/sherpa/astro/xspec_extension.hh`, which has templates that include some variant of

```
		// Suppose the data were filtered, such that there is a gap
		// in the middle of the energy array.  In that case *only*,
		// we will find that xlo[i+1] != xhi[i].  However, XSPEC models
		// expect that xlo[i+1] == xhi[i].
		//
		// So, if we pass in filtered data and xlo[i+1] != xhi[i],
		// then at energy bin i we will end up calculating an energy
		// flux that is far too great.  We will correct that by gathering
		// information to allow us to recalculate individual bins, with
		// boundaries xlo[i], xhi[i], to correct for cases where
		// boundaries xlo[i], xlo[i+1] results in a bin that is too big.
		//
		// We will gather the locations of the gaps here, and calculate
		// actual widths based on xhi[i] - xlo[i] downstream.
		//
		// If we are working in wavelength space we will also correct for that.
		// SMD 11/21/12.

		for (int i = 0; i < nelem-1; i++) {
			double cmp;
			if ( is_wave ) {
				cmp = sao_fcmp(xlo[i], xhi[i+1], DBL_EPSILON);
			} else {
				cmp = sao_fcmp(xhi[i], xlo[i+1], DBL_EPSILON);
			}
			if (0 != cmp) {
				gaps.push_back(i);
				double width = fabs(xhi[i] - xlo[i]);
				if( is_wave ) {
					width = hc / width;
				}
				gap_widths.push_back(width);
			}
		}
```

and then, after evaluating the model

```

		// If there were gaps in the energy array, because of locations
		// where xlo[i+1] != xhi[i], then this is place where we recalculate
		// energy fluxes for those bins *only*.
		//
		// For each such location in the energy grid, construct a new
		// 2-bin energy array, such that the 2-bin array is [xlo[i],
		// xhi[i]].  This is accomplished by:
		//
		// ear2[0] = ear[location of gap]
		// ear2[1] = ear[location of gap] + abs(xhi[location of gap] -
		//                                      xlo[location of gap])
		// The locations of the gaps, and the actual widths of the energy
		// bins at those locations, were calculated above.  So use the
		// gaps and gap_widths vectors here to recalculate energy fluxes
		// at affected bins *only*. SMD 11/21/12

		while(!gaps.empty()) {
			std::vector<FloatArrayType> ear2(2);
			int bin_number = gaps.back();
			ear2[0] = ear[bin_number];
			ear2[1] = ear2[0] + gap_widths.back();
			int ear2_nelem = 1;
			XSpecFunc( &ear2[0], &ear2_nelem, &pars[0], &ifl, &result[bin_number],
					&error[bin_number]);

			gaps.pop_back();
			gap_widths.pop_back();
		}
```

where you can see that there are calls to evaluate the model for grids with 1 bin, which triggers the crash.

DougBurke:
This PR is being left open for now, but it has been superceded by PRs:  #80, #81, #82, #84, and #101

# Release Note

In XSpec 12.8.2q (and possibly earlier versions), XSpec models that call the routine SUMDEM can crash if evaluated on a single bin, i.e. the XSpec model is called with an energy array of size 2 and a flux array of size 1. This is now fixed by only calling the XSpec model once per grid, by adding in elements into the grid to handle non-contiguous input grids, and removing them once the function has been called.

If an XSpec model was sent a non-contiguous grid in Angstroms, the result for bins where gaps occurred was wrong. This is now fixed.

The XSpec extension code was reviewed and cleaned up.

# Details

The following sections (Fixes, Changes, and Open Questions) provide an overview of the changes. The text following this (the Discussion section) repeats this information, providing more details).

## Fixes

The fixes are (this only discusses the final solution; see the details below for any intermediate solutions):

### XSpec models that call the SUMDEM routine (e.g. apec/nei) could crash - issue #62

In XSpec 12.8.2q (and earlier, but I have not checked how far back), XSpec models that call the (XSpec) routine SUMDEM can crash if evaluated on a single bin (that is, the XSpec model is called with an energy array of size 2 and a flux array of size 1). This crash has been reported to HEASARC and is apparently fixed in the forthcoming release.

The explanation for the crash in the existing code is given in #62. It is solved here by only calling the XSpec model once per grid, by adding in elements into the grid to handle non-contiguous input grids, and removing them once the function has been called. This is applied to all the templates except for `xspecmodelfct_con`, since it is not obvious how to handle non-contiguous grids for these models (and this interface is experimental, so there's no Python code currently calling it). An alternative approach would have been to call the XSpec model once per contiguous grid (but then there is still a chance of a contiguous section only containing one element). I also prefer (theoretically and aesthetically) the idea of calling the XSpec model only once per model evaluation.

The test suite has been updated to have a test where the XSpec models are evaluated on a non-contiguous grid. The results are compared to those from evaluating the same model on a
contiguous grid, after removing the "extra" bins. This check is done in two parts:

- the non-edge bins, for which a small tolerance is used
- the edge bins - i.e. those by where the non-contiguous section is - which use a larger tolerance

This was done for purely observational reasons, since some XSpec models (presumably those that are doing something that requires an internal energy grid to be constructed and mapped onto the output grid in some way) were found to require it. It is not particularly easy to follow and is likely fragile, as these tolerances were derived from running on a single machine and so may well require tweaking on other systems.

### Table models could crash if given a non-contiguous energy grid.

Reviewing the table model template (`xspectablemodel` template) I found that its support for non-contiguous grids lead to the code trying to use NULL as an array pointer. A crash is shown in the details below. It was explicitly fixed, but this code is no-longer used, as it was re-written to handle issue #62.

A test case was added to test table models (since there was no existing test of the functionality). It requires an update to the version of `sherpa-test-data` since a new data file is needed. The test is basic (i.e. it checks that the result of evaluating the model is not NaN or Inf and doesn't crash).

### Non-contiguous grids given in Angstroms

If an XSpec model was sent a non-contiguous grid in Angstroms, the result for bins where gaps occurred was wrong (see below for details). This was fixed in the changes for #62 (I have not investigated the logic error in the old code since the new version is different).

I have added tests for handling wavelength grids (both contiguous and non-contiguous) since there was no testing of this functionality. It is based on the existing tests, and just checks that the results for a grid evaluated in Angstroms is similar-enough to the results from when the grid is in keV.

## Changes

These are the most-obvious changes not explicitly mentioned above in the fixes section/

### Normalise the templates

The three templates existing all very similar, but it was hard to see this. I have re-written the templates to bring out the similarities (and differences). I have also, as discussed in the next point, added a template.

### Separate out the convolution interface

Support for XSpec convolution-style models was shoe-horned into the `xspecmodelfct_C` template so that we could experiment with these models. However, it doesn't really make sense, so I pulled out this code into a separate template `xspecmodelfct_con`, and changed the API slightly to make more sense (and be more Python-like).

Note that the convolution-style models are only provided via this function interface. There is currently no set of Python classes to take advantage of them. I have code to do so - e.g. https://github.com/DougBurke/sherpa/tree/feature/xspec-convolution-models - but this is out of band for this PR. All I want to do is to make sure that this convolution interface is not removed.

*EDIT* added note to issue #68 which is a request to add the necessary Python classes to support these models.

### 32-byte handling

The handling of those XSpec models that use the FORTRAN 4-byte real interface (i.e. float32 rather than float64) was changed so that the input grid (or grids) was converted to 32-byte real by the argument parser (`PyArg_ParseTuple`) rather than the original approach which cast them to 64-byte reals and then converted to 32-byte real just before calling the model. The new approach is going to produce slightly-different results (since the wavelength to energy calculation is now done in float rather than double).

### Input argument checking and non-contiguous grid

I added checks to enforce invariants that were assumed by the code - e.g. that if xlo and xhi are given, then the sizes must match. I also added checks to ensure that the grid sent to XSpec is monotonically increasing. XSpec (the program) can enforce such a check outside of the model library, but I think it may be useful here. There is an issue here in that what happens with "bad input", as we get in real-world data - e.g. see https://github.com/sherpa/sherpa/issues/56#issuecomment-121622535 - which should probably be handled in a similar way to XSpec (the program); that is, the Sherpa ARF/RMF objects can ensure that a grid is sensible, and deal with cleaning up any problematic energy arrays. The checks would remain to catch any users calling the models directly. There are also other approaches which can be taken, such as changing the tolerance for defining the edge of bins, or having some way for a user to "fix up" an ARF or RMF that causes problems.

### Clearing out the flux array

The XSpec model API provides no general way for a caller to know if a model failed (e.g. because a data file was missing or some internal constraint is not met, such as the energy grid being out of bounds for a model). The original code just allocated a random section of memory for the output array, sent it to the XSpec model, and then returned it. If the model failed the values would be random, since there is no constraint that XSpec models change the flux array they are sent. I've seen cases where repeated calls to `plot_source` would produce wildly-different plots because of this. The code now ensures that the flux array is set to 0 before calling the model (and is an RFE I've made a while ago but I couldn't find a Sherpa ticket for it in the CXC system).

Note that there is no guarantee that a model doesn't error out part-way through changing the flux array, but I think this change is a worthwhile change.

### Changing ifl/spectrumNumber to 1

The XSpec models provide an ifl/spectrumNumber parameter which we have been setting to 0. This PR changes it to 1 because I have seen one (user model) that assumed that the value is >= 1 (for caching data it had read in). The XSpec documentation does not provide guidance on what the range of this should be. I do not believe there should be any change in behavior due to this change; it is somewhat an aesthetic change.

We currently do not support those models that require this to be set to a meaningful value (that is, the smaug additive model and the mixing models).

### _XOPEN_SOURCE

I re-arranged the order of imports to stop getting compiler warnings about `_XOPEN_SOURCE` being
redefined.

## Open questions

Some of these are discussed above, but called out here:

In the code re-writes, I have generally tried to avoid optimising the code, going for clarity. I have not tried any regression tests to check whether the new code is slower or faster than the old code (my guess is that any difference is going to be small  compared to the run time of all but the simplest of the XSpec models themselves).

How to handle edge-of-bin validation failures (in some way this is a bigger issue than this patch set, but as written the code would lead to user surprise if other changes aren't made).

Should we just use the 64-byte C-style interface for all models, and take advantage of the type-conversions that XSpec provides (i.e. it provides wrapper functions for the other "styles" of interface, whatever the model actually uses)?

The test script has been updated to add tests for the functionality changed here - e.g. table model, wavelength grids, non-contiguous grids - but it's not necessarily robust, or the best approach for general testing purposes.

Should there be some logging added to the interface code, to create debug-level messages, either using the Sherpa logging infrastructure or (easier) the XSpec chatter level?

I would like to be able to support users building XSpec user models against our library/interface. I have yet to look into whether this is possible with the current code. It is out-of-band for this PR, but I would like to make sure that we are aware of this possibility.

-----

# Discussion

The rest of this note provides context for the release notes above (and repeats much of the same information).

# Aim

Fix #62 (handling of non-contiguous energy grids when calling XSpec models) and a general clean up of the interface code. It also fixes a crash in the table-model handling, and fixes up a different problem with the handling on non-contiguous grids (this time when given in wavelength); details for these are below.

# Current Status

Ready for review. I see this as a starting-point in the discussion of how to update the interface to the XSpec models, rather than necessarily being the final patch set.

# Problems with the current code

See #62 for a discussion of non-contiguous grids in energy units (this info should probably be summarized here).

## Crash with table models

If a table model is used with a non-contiguous energy grid, then the code will use a null pointer; an example of this is shown using CIAO 4.7 below (the same behaviour is also seen using the current master branch). The table model in use `RCS.mod` is from https://heasarc.gsfc.nasa.gov/xanadu/xspec/models/rcs.html but has also been added to the sherpa-test-data repository (as one of the commits below adds this test to `test_xspec.py`)

```
% sherpa
CIAO 4.7 Sherpa version 1 Thursday, December 4, 2014

sherpa-1> load_table_model("tmod", "RCS.mod")
sherpa-2> tmod([0.1,0.2,0.3,0.4])
          array([ 2.94548774,  2.69927979,  1.91469491,  0.        ], dtype=float32)
sherpa-3> tmod([0.1,0.2,0.3],[0.2,0.3,0.4])
          array([ 2.94548774,  2.69927979,  1.91469491], dtype=float32)
sherpa-4> tmod([0.1,0.3],[0.2,0.4])
/home/dburke/local/ciao-4.7/binexe/sherpa: line 199:  9838 Segmentation fault      (core dumped) ipython --profile sherpa
```

This has been addressed in the commits below.

## Wavelength and non-contiguous grids

The gist at https://gist.github.com/DougBurke/b70485a9280f1b52a83e shows how the XSpec and Sherpa power-law models do not calculate the same values when the input energy grid is in Angstroms and contains a gap - the output from CIAO 4.7 and the master branch is

```
wy1: [ 0.69314718  1.38629436  0.18232156  0.15415068]
wy2: [ 0.69314718  0.40546511  0.18232156  0.15415068]
```

where `wy1` shows the output from the XSpec power-law model and `wy2` is from the Sherpa model. The second bin of `wy1` (1.386) is wrong.

This has been addressed in the commits below; the output is now

```
wy1: [ 0.69314718  0.40546511  0.18232156  0.15415068]
wy2: [ 0.69314718  0.40546511  0.18232156  0.15415068]
```

# Changes made in this PR

Some of the commits in the PR end up being re-written in later commits, as a different approach (or more clean up) invalidates the previous commit. The following list the major changes, but there are some small changes that are not explicitly listed here.

### Fix table model code crash (non-contiguous grid)

The original version of the `xspectablemodel` template had a null pointer exception when a non-contiguous grid was used. This was fixed in @709a365, but the code is re-written in a later commit. A test case is added for this problem, and required an addition to the `sherpa-test-data` repository in sherpa/sherpa-test-data#1

I also updated the code to go back to using the `std::vector` approach that the other templates used, rather than manually managing memory - see @d7201a4 - as it's not obvious why it was changed.

### Fix for non-contiguous grids: energy

If a non-contiguous xlo/xhi grid is sent in - e.g. `xlo=[0.1,0.2,1.0]` and `xhi=[0.2,0.3,1.1]` - then the original code evaluated the model using `[0.1,0.2,1.0,1.1]` and then went back and calculated the model on the grid `[0.2,0.3]`, using this value to replace the erroneous value (which was evaluated for 0.2-1.0). This caused problems since some XSpec models - which internally called XSpec's SUMDEM routine - would crash if called with only one bin (i.e. when sent a 2-element `ear` grid), which is the cause of #62. This crash has been reported to Keith Arnaud and is reportedly fixed in the next XSpec release, but that doesn't help here.

The first fix - in @114fbe9 - addressed this by ensuring that the secondary calls to the XSpec function used a three-element `ear` grid, rather than two. This is most-definitely a work around fix, and some strange behaviour is still seen (e.g. including the `xsnsagrav` model in the test suite would cause a crash). It wouldn't surprise me if other XSpec models assumed some minimum size on the input energy grid. This fix was made to all three templates.

I don't like this approach, so have several commits that change the method so that the XSpec model is only called once. This is done by identifying gaps and inserting extra elements into the `ear` grid at these gaps to create extra bins that are not in the user grid. After running the model, the results are copied over to remove these bins. So, with the example above of `xlo=[0.1,0.2,1.0]` and `xhi=[0.2,0.3,1.1]`, an `ear` grid of `[0.1, 0.2, 0.3, 1.0, 1.1]` is created - effectively adding the bin `0.3-1.0` - and then, after calling the model, the data moved around and re-sized before being returned to the user. This requires adding a `resize1d` method to the `Array` class. This change was made in several changesets: @c335068 and @c431a42 apply to the `xspecmodelfct` template (the first check in missed out a rather crucial change), @6f1884e add it to `xspecmodelfct_C`, and @5d8a712 adds it to `xspectablemodel`. The `xspecmodelfct_con` template has not been updated since it is likely that non-contiguous grids are not going to be supported for these models.

These re-writes fix the problem with non-contiguous wavelength grids (discussed next).

An alternative approach to handling non-contiguous grids would be to to take the gap-handling code and to evaluate the model once for each contiguous grid, filling in the output grid as required. This would not be an extensive change to make to the code. I have not tried this as I don't like the idea of running the model multiple times with a different grid, but it is an option.

### Fix for non-contiguous grids: wavelength

There was some error in the code that meant the "fix-up" step was calculating the wrong value if the non-contiguous grid was in Angstrom. This has been fixed as part of the non-contiguous grid change.

### Separate out the convolution-model code into a separate template

The current experimental support for XSpec convolution-style models was shoe-horned into the template for C++ style models. I have separated them out as they are different: @237c904.

Change @71c9a73 changes how the convolution models are called from Python (the argument order). I feel safe making these changes since this is an experimental interface, only provided as a low-level, so anyone relying on this (which I believe is only some of my code in the CIAO contributed package) will have to manually update things.

Change @b7a0faf updates the code to "match" the C++ template (as much as it can, as their is no support for gaps in the grid).

The Python support needed to use these models from `sherpa.astro.ui` is not part of this patch set.

### Clean up of the templates

I have tried to make the templates more consistent (same structure) so it is more obvious where the commonalities are. I have also re-organised the code to be more readable (at least to me) and hopefully to avoid some of the un-needed duplication when the input grid is in wavelength units (Angstrom) and so has to be converted to keV. The code may not be quite as efficient, since some checks could be moved into a single loop, but I'd prefer to understand the code and then worry about optimisation (I also haven't checked to see if there is actually much run-time difference; I would expect any differences to be small compared to the evaluation time of a model).

### Use appropriate types

The template for the FORTRAN-API models - e.g. `xspecmodelfct` - treated the input grid array (or arrays) as `double`, with a down-cast to `float` just before calling the XSpec function. I have changed this so that they are cast to `float` by the argument parser (`PyArg_ParseTuple`) as there seems little point in the previous behavior. This is definitely up for debate (e.g. see the discussion of Added validation checks of the input). See @081eb5d.

### Added validation checks of the input

The code assumed that, when given two grids (i.e. `xlo` and `xhi`) that they were the same size. This is now enforced. Similarly, there are now checks that the grid is monotonic (e.g. grid boundaries do not overlap). This can lead to surprising behavior for the user - e.g. see https://github.com/sherpa/sherpa/issues/56#issuecomment-121622535 - but I think this is more an issue for the ARF/RMF classes (in that they can either error out or sanitize the input so that a sensible grid is sent to the model), leaving these checks for users who manually call the models. Alternatively, the tolerance used to determine whether a bin edge is aligned (which was `DBL_EPSILON` before this PR and is now either `DBL_EPSILON` or `FLT_EPSILON` in the new code) may need to be adjusted.

Commits include: @ed127be and @171d182

### On error return 0 for all flux elements

The return array is now initalized to 0 for each element, rather than being left uninitialized - see @66c93e9. This is to catch the case of a model error-ing out and not changing the flux values. Prior to this PR this would result in the array having random values in, whereas now they are all 0. For an example of the problems the old behavior can cause, see CXC ticket #13163 (although that focuses more on the error reporting rather than this change). Note that due to the lack of error handling in the XSpec model API, there's no guaranteed way to be able to tell a model has failed or not, so this is likely as well as we can do. For instance, we can do nothing if a model starts editing the flux output array and then decides to error out (no idea if this ever happens in the code base, and not something I am worrying too much about).

###  Use an ifl/spectrum number of 1

The XSpec model interface has an ifl/spectrumNumber parameter which was set to 0. I have changed it to use 1 (@e04c433) . This value is not used by most models, and there's no official guidance on what the value should be, but I have seen at least one user model assume that its value is >= 1 (I don't think it actually resulted in different numerical results, but was used for caching data). This is really a stylistic change as I have no evidence that using 0 is problematic.

There is an existing CXC ticket (#13974) which is related to being able to set this parameter so that certain models (`smaug` and the mixture models) could be supported. This is outside the scope of this PR.

### Avoid a warning about re-defining `_XOPEN_SOURCE`

The order of imports was changed so that `Python.h` is included first, to stop seeing warning messages about `_XOPEN_SOURCE` being redefined. See @5e84ac0.

### Testing tweak

Removed reporting an error using the Sherpa logger, as this isn't helpful with the current testing set up. See the first commit @d75220. More substantive changes are discussed in the Testing section below.

## Notes

Some thoughts (it may repeat information presented above):

 - How should overlapping bins be handled? The present approach is to error out, but other ones
   are possible.

 - Should FORTRAN style float models work entirely with floats (as they do now) or only convert
   the energy grid to float from double at the last moment (as it was before)?

   Alternatively, we could just use the C++ interface for all the models, and let XSpec deal with
   converting the data types if necessary.

 - Should there be some logging of work done by the templates (e.g. for bug tracking)? Messages
   could be written either to the Sherpa logging infrastructure or just piggy back on the XSpec
   logging infrastructure.

 - Should the template code instead be in routines (e.g. that accept a pointer to the function) and
   then the model code would call them with the appropriate function? I am thinking about how best
   to support compiling XSpec user models for use by Sherpa. This is really out of band of this PR,
   but wanted to mention it in case I forgot it.

 - Testing needs thinking about (and hence the Testing section below)

## Testing

The test suite - `sherpa/astro/xspec/tests/test_xspec.py` - has been updated to add tests for existing functionality - namely: handling of non-contiguous grids, grid in angstroms rather than keV, table model, convolution model. I also changed it so that it could be run from the command line and take in the location of the test data suite, rather than hard-coding it to a path that is not valid for most people.

As we don't have any external dataset against which to compare models (other than those models for which we have a model in Sherpa, which is only going to be simple geometrical models like a power law, gaussian, constant), the tests include checking that > 0 and finite (so not NaN/Inf) and (added in this PR) that the results agree with each other. That is, that using a single array matches two arrays (when the grids are the same), that using a non-contiguous grid (which requires two input arrays) matches the single array case (once the "missing" bins are removed), and that the same results are returned when the input array(s) are in Angstroms rather than keV. In some cases exact numerical equality is expected, but in many we expect there to be some numerical differences (see the note on XSpec model evaluation and non-contiguous grids below). Since the tests loop through all the models, this ends up with a rather ugly set of tolerances, as some models require different tolerances. These tolerances have been tuned to my particular machine, so it is expected that they would need tweaking on other machines (and this has been seen in some internal builds here at the CXC). The tests are useful as is for this development (checking that changes don't make a large difference to the output), but  they may not make sense for general testing once this PR has been merged (or the results used as a basis for a different PR).

# A note on XSpec model evaluation and non-contiguous grids

The XSpec models are given a single, contiguous grid (the `ear` grid). Some models calculate the spectrum on an internal grid, and then do some form of interpolation onto the `ear` grid. This is problematic for supporting non-contiguous grids;e.g. see https://gist.github.com/DougBurke/d9a0074489b6d1de108e

It's not obvious to me that the convolution-style models can be sensibly used with a non-contiguous grid.


DougBurke:
I now have an initial fix for #62, but it's not without its own problems, so this is still not ready for review.

DougBurke:
Using the latest commit - @6f1884e - the example at https://gist.github.com/DougBurke/b70485a9280f1b52a83e calculates the same numbers for the Sherpa `powlaw1d` and XSpec `powerlaw` models when evaluated with a wavelength grid (at least to the precision used to display to the screen); e.g.

```
wy1: [ 0.69314718  0.40546511  0.18232156  0.15415068]
wy2: [ 0.69314718  0.40546511  0.18232156  0.15415068]
```

Before this, the result was (with `wy1` being the XSpec model using the buggy interface code):

```
wy1: [ 0.69314718  1.38629436  0.18232156  0.15415068]
wy2: [ 0.69314718  0.40546511  0.18232156  0.15415068]
```

DougBurke:
As discussed in https://github.com/sherpa/sherpa/issues/56#issuecomment-121622535 - @6f1884e seems to "fix" #56 at the cost of refusing to evaluate the model (for some models). I think it is okay to do this, as long as the ARF and RMF "instrument model" classes are made to "fix up", or report, or both, cases where the grids are incompatible or invalid (there's arguments to be made for "fixing up" vs error-ing out).

anetasie:
I had an issue with the non-continuous grids a long time ago. When you plot such models the model curve had a huge spike at the beginning of ignored section of the spectrum. The fix for this was supposed to be sending a full continuous data array to XSPEC for model evaluation. This works ok in the ciao4.7 code and passes my test. Should we include this case in testing your PR?

olaurino:
@DougBurke before I make any comments, probably now it would be a good time to change the PR's title and to add a Release Note section that captures the main changes and that we can curate as the PR evolves.

DougBurke:
I've tried to summarize the contents of the changes in a "Release Note" section at the top, but have left the existing text below it to provide details and context.

olaurino:
The Release Note section should make the task of collecting the Release Notes for both CIAO and the Standalone release easier. I reviewed the Release Note section so that we can simply copy and paste it in the actual release notes. We will further review it if the PR changes and if necessary. @DougBurke please let me know if my changes make sense. Thanks.

dtnguyen2:
1) I have looked over the code and the question I have is:  As I have
mentioned in my Xspec report which was sent about a month ago, the original
author (Stephen, I believe) wrote the interfaces to the Xspec Fortran and
C/C++  as templates functions (Xspec Fortran and C++ models are written in
float and double, respectively) and yet he hard wired the code with either
float or double as a result Doug had to duplicate the following sections 4
times (to be accurate, there is a slight variation for the convolution
model).  So if there is an error in logic in the aforementioned section
then one has to modify it in four different places.  I think that it should
take a competent programer about 2-3 days to re-write this as a real
template C++ functions.

        if ( nelem < 2 ) {
          std::ostringstream err;
          err << "input array must have at least 2 elements, found " <<
nelem;
          PyErr_SetString( PyExc_TypeError, err.str().c_str() );
          return NULL;
        }

        if( xhi && (nelem != int(xhi.get_size())) ) {
          std::ostringstream err;
          err << "input arrays are not the same size: " << nelem
              << " and " << int( xhi.get_size() );
          PyErr_SetString( PyExc_TypeError, err.str().c_str() );
          return NULL;
        }

    int ifl = 1;

    bool is_wave = (xlo[0] > xlo[nelem-1]) ? true : false;
        FloatArray *x1 = &xlo;
        FloatArray *x2 = &xhi;
        if (is_wave && xhi) {
            x1 = &xhi;
            x2 = &xlo;
        }

        // Are there any non-contiguous bins?
        std::vector<int> gaps_index;
        std::vector<FloatArrayType> gaps_edges;
        if (xhi) {
          const int gap_found = is_wave ? 1 : -1;
          for (int i = 0; i < nelem-1; i++) {
            int cmp = sao_fcmp((*x2)[i], (*x1)[i+1], FLT_EPSILON);
            if (cmp == gap_found) {
              gaps_index.push_back(i);
              gaps_edges.push_back((*x2)[i]);
            } else if (cmp != 0) {
              std::ostringstream err;
              // not convinced this is understandable to users, particularly
              // if the grid is in Angstrom. It is also possible that the
              // format used isn't sufficient to show the problem, but I
              // do not want to tweak the format here just yet.
              err << "Grid cells overlap: cell " << i
                  << " (" << (*x1)[i] << " to " << (*x2)[i] << ")"
                  << " and cell " << (i+1)
                  << " (" << (*x1)[i+1] << " to " << (*x2)[i+1] << ")";
              PyErr_SetString( PyExc_ValueError, err.str().c_str() );
              return NULL;
            }
          }
        }
....


2) I have to admit I have not run the sherpa test but I honestly don't
understand how the code can  work:  The Xspec C/C++ interface starts on
line 328 as:

template <npy_intp NumPars, bool HasNorm,
void (*XSpecFunc)( const double* energy, int nFlux,
        const double* params, int spectrumNumber,
        double* flux, double* fluxError,
        const char* initStr )>
PyObject* xspecmodelfct_C( PyObject* self, PyObject* args )

Note that the first and 6th arguments to the Xspec C/C++ model functions
are 'const double*'.  However, on line 392, 424-425 and 500-503

std::vector<SherpaFloat> gaps_edges;

        // XSpec traditionally refers to the input energy grid as ear.
        std::vector<SherpaFloat> ear(ngrid);

and

// Since the flux error is discarded, it does not need to be a
        // NumPy array. Should be set to zeros for safety.
        std::vector<SherpaFloat> error(dims[0]);

respectively.  That is the ear and error arrays are defined as float
arrays.  Yet, the call to XspecFunc is called with ear and error as the
first and 6th arguments which are supposed to be a 'const double *'

          XSpecFunc( &ear[0], npts, &pars[0], ifl,
                     &result[0], &error[0], NULL );


I honestly don't know how this could ever work!  I know that I should run
the test but ....

3) Replace: FLT_EPSILON and DBL_EPSILON with the appropiate calls to C++
std::numeric_limits<T>::epsilon once we properly C++ templatize it.

not to mention

         float hc = (float) (sherpa::constants::c_ang<SherpaFloat>() *
                              sherpa::constants::h_kev<SherpaFloat>());

etc...

dtnguyen2:
Just realized something, SherpaFloat is defined as double.  Very cryptic so I guess #2 in my mail is a non-issue

DougBurke:
@dtnguyen2 I had thought about cleaning up the float/double/SherpaFloat changes, but I wasn't sure what the semantics of SherpaFloat was, and then forgot about it by the time it came to finalising the code for sbumission a a PR.

DougBurke:
This builds against the recently-released version 12.9.0 of Xspec. The tests fail, but this appears to be own to tolerance settings when comparing the `agauss` and `zagauss` models (e.g. removing these models make the tests pass). Although not explicitly mentioned in the changelog, there was a change to these models in this release.

DougBurke:
I have reported an issue with the agauss/zagauss models in Xspec 12.9.0 to the Xspec developers.

anetasie:
Waiting for XSPEC Travis build  before we can merge this

olaurino:
I am labeling this `pr:hold`, as @DougBurke is rolling out atomic PRs that will eventually replace this one. This way this PR does not get included in the `snapshot` branch and builds.

DougBurke:
I am closing this as the changes in it are either already merged (in other patches) or in #101

PR #69, opened 2015-07-21
Remove S-Lang files.

DougBurke:
As S-Lang is no-longer supported in Sherpa, these files can be removed.

This closes out CXC ticket 13968 (the other part was #24 which has been merged).


olaurino:
Merging. Thanks.

PR #59, opened 2015-07-08
fix #38 - grouping twice gives an IndexError exception

olaurino:
### Release Note
Fix bug #38 (grouping twice gives an IndexError exception). An un-handled corner case in one of the Sherpa internal methods (`utils.create_expr`) was triggering an `IndexError` when two `group_counts` operations were performed back to back. The fix handles the case so that applying `group_counts` twice does not result in an Exception

### Tests
The test exercises a simple case where the grouping is applied more than once in contiguous calls. The test does not check that the grouping is done correctly, and it looks like no tests are actually checking that grouping operations provide meaningful results. We need to decide whether we want to add meaningful tests at this iteration (before merging this PR) or not.

PR #70, opened 2015-07-23
Fixes and PEP 8/pyflakes updates for utils

DougBurke:
These commits fix a potential error, in that `OutOfBoundErr` exceptions were not being caught (although it's not actually clear that this is ever thrown). There were several places where screen output used either `print` or `sys.stderr.write`: these have been changed to use the `debug` and `warning` loggers instead (respectively). Several routines which use `min` and `max` as variable names have been changed. There are very-minor documentation changes for Sphinx (mainly change `:` to `::` when introducing some maths).

It also addresses a number of pyflakes/PEP-8 violations: more Pythonic use of boolean variables (e.g. `not xxx` rather than `False == xxx`); add spaces between operators; remove spaces before/after various brackets; line length; two blank lines between top-level declarations; explicit module export; remove blank characters at the ends of lines.

The commits have been broken up into separate sections to try and make reviewing the changes somewhat easier.

The reason for these changes is that I want to make some changes to the test class code, but found a number of issues that were being flagged up by

```
pyflakes "$1"
pep8 --ignore=E221,E701,E202 --repeat "$1"
```

so decided to clean these up first.


DougBurke:
Note that these changes are included in #71 - perhaps due to my lack of git mumbo jumbo.

olaurino:
@DougBurke Thanks for noting it. In general, we should avoid dependencies between PRs, but sometimes that's just the way to go, as long as we know that one PR has to be merged before the other.

In this case, I wonder if we shouldn't close #70 and work with #71 as a single clean-up PR. Otherwise I would suggest we rebase #71 on top of master and keep them separate, since they do not overlap.

DougBurke:
I think that some form of rebasing is going to be needed once #59 is
merged. One reason for keeping them separate is that it might be a bit
cleaner to review the changes broken up this way (i.e. two PRs), but I
don't have a strong opinion on this.

olaurino:
This PR has been replaced by #78.

PR #71, opened 2015-07-23
Test cleanup

DougBurke:
The aim was to add the `SherpaTestCase.make_path()` routine for adding paths to the data directory, and use it in the code. A secondary change was to improve support for running an individual file - e.g. `python sherpa/astro/datastack/tests/test_datastack.py` - with or without the Sherpa test directory.

Two tests which were previously skipped - due to name clashes - have been re-named: `sherpa/models/tests/test_template.py::test_grid_search_with_discrete_template_parvals` and `sherpa/ui/tests/test_ui.py::test_psf_model1d``.

For those tests that changed the logging level in their `setUp` method, the level has been restored in `tearDown` (not checked for all tests, just those looked at here). Similarly, some directory changes (in those tests with a `run_thread` method) were changed to ensure that the original directory was restored.

There was also some minor code cleanup (e.g. removal of unused imports and adding/deleting spaces).

I have not changed the XSpec module tests since I know there's changes (e.g. PR #63) so not worth working on here.

These changes build on the test-suite cleanup in PR #70. I am not convinced I "git-ted" things correctly to link things together properly.

### DIFF
--- a/sherpa/astro/datastack/tests/test_datastack.py
+++ b/sherpa/astro/datastack/tests/test_datastack.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2014  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2014, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


olaurino:
Replaced by PR #79

PR #72, opened 2015-07-28
add xspec build

olaurino:
This PR adds Xspec to the Travis CI builds. Xspec and its dependencies (cfitsio and CCfits) are installed as conda binaries from a non-public channel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ addons:
       - bison
 
 matrix:
+  fast_finish: true
+  allow_failures:
+    - env: XSPEC="true" FITS="astropy" INSTALL_TYPE=develop TEST=submodule
   include:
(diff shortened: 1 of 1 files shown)


DougBurke:
There is a new version of Xspec out - 12.9.0, currently at patch b - which may improve the stability of the Xspec run. Otherwise, we really need to understand what is causing the problem.

olaurino:
I discussed this with @wmclaugh. He is OK with merging this PR, as it is better to merge it than not to, at least for Xspec-related changes. Non-Xspec builds are unaffected. We all agree we should understand where the problem is, as there may be a serious underlying issue. For that, I will try and open a ticket on the Travis CI repo, as they can look at the logs and give us more information.

Issue #64, opened 2015-07-09
Comparison to None in Numpy 1.9


olaurino:
Since the Numpy 1.9 upgrade, users have been reporting warning messages like the following:

> FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.

Generally speaking, it is good practice to do None checks with `if var is None` rather than `if var == None` because the equality check can be (at least in principle) overridden. Which is what some of the Numpy structures do or are going to do in the future.

The issue is usually easy to fix, but getting a list of these warnings is not easy. We did fix some of them in the past (see 61fbfcdc) but more have been reported, in particular:

~~~~
sherpa/astro/io/crates_backend.py:1224
sherpa/sim/sample.py:184
sherpa/astro/ui/utils.py:8763
sherpa/astro/ui/utils.py:8768
~~~~

For what is worth, a workaround exists:

~~~~
import warnings
warnings.filterwarnings('ignore', category=FutureWarning)
~~~~

PR #77, opened 2015-08-14
Bug #64 comparison to none

olaurino:
# Release Note
Replace `==` and `!=` comparisons to `None` with `is` and `is not`

# Description
The Release Note says it all. Moreover, there is a pep8 clean-up of the touched files.

### DIFF
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -1,4 +1,4 @@
-# 
+#
 #  Copyright (C) 2010  Smithsonian Astrophysical Observatory
 #
 #
@@ -44,14 +44,16 @@ __all__ = ('RMFModel', 'ARFModel', 'RSPModel',
            'ARFModelPHA', 'ARFModelNoPHA',
            'RSPModelPHA', 'RSPModelNoPHA',
            'MultiResponseSumModel', 'PileupRMFModel', 'RMF1D', 'ARF1D',
-           'Response1D', 'MultipleResponse1D','PileupResponse1D',
+           'Response1D', 'MultipleResponse1D', 'PileupResponse1D',
            'PSFModel')
 
 
 class RMFModel(CompositeModel, ArithmeticModel):
+
     """
     Base class for expressing RMF convolution in model expressions
     """
+
     def __init__(self, rmf, model):
         self.rmf = rmf
         self.model = model
@@ -61,9 +63,12 @@ class RMFModel(CompositeModel, ArithmeticModel):
 
         # FIXME: group pairs of coordinates with one attribute
 
-        self.elo = None; self.ehi = None  # Energy space
-        self.lo = None;  self.hi = None   # Wavelength space
-        self.xlo = None; self.xhi = None  # Current Spectral coordinates
+        self.elo = None
+        self.ehi = None  # Energy space
+        self.lo = None
+        self.hi = None   # Wavelength space
+        self.xlo = None
+        self.xhi = None  # Current Spectral coordinates
 
         # Used to rebin against finer or coarser energy grids
         self.rmfargs = ()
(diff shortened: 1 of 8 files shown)


DougBurke:
@olaurino - is this just for a few files or the whole code base (perhaps this could be added in the release notes or description)?

olaurino:
I started with the ones that were reported, as those are affected by the Numpy comparison to None override. At the same time, as I found more None comparison in the files I was modifying, I fixed them as well. I was planning to look for more instances of the issue and see if it was worth going through all of them in the code base. I'll do that next.

olaurino:
As a reminder to myself, I also have to replace the Copyright year.

DougBurke:
Just to note that I made some changes like this in #70/#71 (just to avoid repeated work/even-more complexity in getting #71 pulled into the repository ;-).

olaurino:
@DougBurke I know... we'll see if this results in conflicts, but that would be easy to fix once your PRs are merged (I guess that will be merged first).

sherpa/astro/instrument.py
```diff
@@ -228,9 +231,11 @@ def filter(self):
             # If PHA grid is in angstroms then convert to keV for
             # consistency
             if (bin_lo[0] > bin_lo[-1]) and (bin_hi[0] > bin_hi[-1]):
-                bin_lo, bin_hi = DataPHA._hc/pha.bin_hi, DataPHA._hc/pha.bin_lo
+                bin_lo, bin_hi = DataPHA._hc / \
```
DougBurke: Rather than have the awkward  continuation line, why not just break it into two lines (which I think also emphasises better what is going on):

    bin_lo = DataPHA._hc / pha.bin_hi
    bin_hi = DataPHA._hc / pha.bin_lo

sherpa/astro/instrument.py
```diff
@@ -322,9 +326,11 @@ def filter(self):
...
```
DougBurke: Again, I suggest splitting into two lines rather than have a continuation line.

sherpa/astro/instrument.py
```diff
@@ -421,9 +425,11 @@ def filter(self):
...
```
DougBurke: Another "avoid a continuation line by splitting into two lines" case.

sherpa/fit.py
```diff
@@ -417,15 +422,16 @@ def myformat( hfmt, str, lowstr, lownum, highstr, highnum ):
         elif False == in_low and True == in_high and mymethod:
```
DougBurke: How about changing the boolean checks - e.g. here

    elif not in_low and in_high and mymethod:

there's a few more like this in this block of code.

sherpa/fit.py
```diff
@@ -661,27 +671,31 @@ def sigmarej(self, statfunc, pars, parmins, parmaxes, statargs = (),
...
+                "'hrej' value for sigma rejection method must be a number")
         if (not (hrej > 0)):
-            raise SherpaErr("'hrej' must be greater than zero") 
+            raise SherpaErr("'hrej' must be greater than zero")
 
         lrej = self.itermethod_opts['lrej']
         if (type(lrej) != int and
-            type(lrej) != float):
-            raise SherpaErr("'lrej' value for sigma rejection method must be a number")
+                type(lrej) != float):
```
DougBurke: The two `type(lrej)` checks could be written on the same line.

sherpa/fit.py
```diff
@@ -661,27 +671,31 @@ def sigmarej(self, statfunc, pars, parmins, parmaxes, statargs = (),
...
+            raise SherpaErr(
+                "'lrej' value for sigma rejection method must be a number")
         if (not (lrej > 0)):
```
DougBurke: I think it's clearer to have

    if (lrej <= 0)

sherpa/fit.py
```diff
@@ -1131,16 +1144,16 @@ def report_progress(i, lower, upper):
...
+                type(self.method) is not LevMar):
             warning(self.method.name + " is inappropriate for confidence " +
                     "limit estimation")
 
         oldmethod = self.method
         if (hasattr(self.estmethod, "fast") and
-            bool_cast(self.estmethod.fast) is True and
-            methoddict is not None):
-            if (isinstance(self.stat, Likelihood) ):
+                bool_cast(self.estmethod.fast) is True and
```
DougBurke: Remove the `is True` ?

sherpa/fit.py
olaurino: Plus, there are better ways of checking if an object is a number (e.g. what if it was a long?).

olaurino:
@DougBurke I implemented the changes you suggested and some more on top of them. Can we merge?

DougBurke:
I'm sure there's more opportunities in the `utils.py` files for clean up (they were too big for github to let me review), but I don't think that's a reason to hold up this particular PR.

PR #78, opened 2015-08-17
Fixes and PEP 8/pyflakes updates for utils

olaurino:
This PR replaces #70.

# Release Note
OutOfBoundErr exceptions are properly caught in some sherpa.utils functions are properly caught.
There were several places where screen output used either print or sys.stderr.write: these have been changed to use the debug and warning loggers instead (respectively). Several routines which use min and max as variable names have been changed. A number of other pyflakes/PEP-8 style violations were addressed, e.g. a more Pythonic use of boolean variables (e.g. not xxx rather than False == xxx).

# Description
These commits fix a potential error, in that `OutOfBoundErr` exceptions were not being caught (although it's not actually clear that this is ever thrown). There were several places where screen output used either `print` or `sys.stderr.write`: these have been changed to use the `debug` and `warning` loggers instead (respectively). Several routines which use `min` and `max` as variable names have been changed. There are very-minor documentation changes for Sphinx (mainly change `:` to `::` when introducing some maths).

It also addresses a number of pyflakes/PEP-8 violations: more Pythonic use of boolean variables (e.g. `not xxx` rather than `False == xxx`); add spaces between operators; remove spaces before/after various brackets; line length; two blank lines between top-level declarations; explicit module export; remove blank characters at the ends of lines.

The commits have been broken up into separate sections to try and make reviewing the changes somewhat easier.

The reason for these changes is that I want to make some changes to the test class code, but found a number of issues that were being flagged up by

```
pyflakes "$1"
pep8 --ignore=E221,E701,E202 --repeat "$1"
```

so decided to clean these up first.

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -105,13 +105,14 @@ _guess_ampl_scale = 1.e+3
 ###############################################################################
 
 
-
 # Default numeric types (these match the typedefs in extension.hh)
 SherpaInt = numpy.intp
 SherpaUInt = numpy.uintp
 SherpaFloat = numpy.float_
 
-################################################################################
+###############################################################################
+
+
 class TraceCalls(object):
     """ Use as a decorator on functions that should be traced. Several
         functions can be decorated - they will all be indented according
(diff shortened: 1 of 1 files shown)


DougBurke:
Looks good to me.

olaurino:
@wmclaugh (as you were assigned to it). I reviewed this PR and I think it should be merged.

PR #80, opened 2015-08-17
Feature/xspec remove xopen source redefinition warning

DougBurke:
Avoid warnings about _POSIX_C_SOURCE and _XOPEN_SOURCE symbols being redefined when
compiling _xspec.cc. The trick is to make sure Python.h is the first file to be included (it is included by sherpa/array.hh which is included by sherpa/extension.hh).

This is the first part of breaking up PR #63 into smaller chunks.


olaurino:
I reviewed this PR and I think it can be merged.

PR #81, opened 2015-08-18
Ensure Xspec models return 0's in case of error.

DougBurke:
# Release Notes

Ensure that XSPEC models which fail - for instance, because a data file it needs is missing - return 0's for all bins, rather than random values. This should make it more obvious that something has gone wrong (for instance if the XSPEC chatter level is low enough not to show any error messages, as is the case for the default setting used by Sherpa, namely 0).

If the model returns an invalid value - defined as an infinite value or NaN - then the model class will raise an error (using the Python `FloatingPointError` exception).

# Details

The XSpec model API does not provide a way of indicating that a
model failed. The common case is that an error message is printed,
often at a chatter level greater than the 0 used by default by Sherpa
(so that users do not see it) and the routine returns without filling
in the flux array. The flux (and error) arrays are now created so that
there contents are set to 0 rather than being undefined. This provides
a more-consistent, and more-obvious, signal to the user that a model
has failed (the result is 0). I believe this is what Xspec itself does when
using the Xspec fitting application.

Another option would be to use an 'out-of-bound' value (e.g. NaN or Inf) 
and then check for these after calling the model, so that an error can be 
thrown, but I am concerned with the extra checks and with coming up 
with a choice of a value that can never be returned by any Xspec model.

This is the second part of breaking up PR #63 into smaller chunks: #80 was the first.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2010  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2010, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -16,6 +16,8 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+
+import numpy
 
 import string
 from sherpa.models import Parameter, ArithmeticModel, modelCacher1d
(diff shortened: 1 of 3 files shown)


DougBurke:
An open question is what error should be raised if an XSPEC model returns a non-finite value. The current choice is `FloatingPointError`, but:

 - should it be a different Python error type

 - or should it be a Sherpa specific error type - the closest appears to be `sherpa.utils.err.ModelErr`. Whatever is chosen, it would require adding a new label/message pair for that error class.

olaurino:
Did the conversation with Keith provide any input to this PR?

Regarding the open question above: I agree FloatingPointError does not seem to be the best choice. Also, I believe throwing the exception is a change to the signature (or contract, rather) of the method, and as such it is probably not backward compatible. I am not against it, but I thought that should be properly stressed.

By the way, we should add a Release Note section before we merge this one.

PR #82, opened 2015-08-18
Change spectrumNumber/ifl from 0 to 1.

DougBurke:
# Release Note
The XSpec "spectrum number" value is now set to 1 rather than 0, as this value is 1-based in Xspec.

# Description
The XSpec "spectrum number" value is not used by Sherpa. It was set to
0, but this is not consistent with Xspec, as the value is 1-based.

A value of 0 has also been seen to cause strange behavior in an XSpec user
model (not shipped with Sherpa).

I also added some comments on how this value could be set by the user,
in the hope of supporting the smaug user model, but this is a long-term
change and may not work out as currently documented.

This is the third part of breaking up PR #63 into smaller chunks after #80 and #81.

### DIFF
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -35,6 +35,17 @@ typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
 typedef float FloatArrayType;
 
 
+// The spectrum number is set to 1. It used to be 0, but some code
+// (a user model, so not included in the XSpec model library being
+// built against) has been seen to behave strangely with a value of 0,
+// so a value of 1 is being used "just in case". A keyword argument
+// could be added so that the user can override this, but it is
+// only really woth doing once write access to the XFLT keywords
+// is added (and then working out how to take advantage of it; perhaps
+// a wrapper model that provides a parameter-like interface to the value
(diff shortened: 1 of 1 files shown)


olaurino:
I updated the description to add a release note.

I don't see anything that should prevent this PR to be merged, however I would be more comfortable if we better specified in the description:
  * which xspec user model was affected by this change.
  * whether the xspec numbering for this field is 0- or 1-based. In the second case this PR would be a no-brainer whether or not it affects a user model.

DougBurke:
I can't remember which user model it was, and it's not one we distribute anyway (I have code to build a module to interface to XSPEC user models as part of the CIAO contributed scripts, and I'd like to move this functionality into Sherpa itself, but that's a longer term issue).

I am *fairly* confident XSPEC uses a 1-based numbering scheme. I'm off for the rest of the week, but I'll try and confirm that.

DougBurke:
Here's the response from Keith Arnaud:

     The ifl/spectrumNumber is one-based.

olaurino:
Then, as I said, this is a no-brainer. I updated the PR description.

The comment you added in the code is now slightly outdated. I'll leave it to you to decide whether you want to update it in a new commit or leave it as is. In any case you can label this to merge.

DougBurke:
I'm currently digging up my garden, can you label this Omar?

PR #84, opened 2015-08-19
Clean up of error code/messages from XSPEC module.

DougBurke:
# Release Note
Clarified error messages in Xspec extension. Also, changed the class of the exception from `RuntimeError` to more appropriate exception types, in particular `LookupError`, `ValueError`, `KeyError`.

This is a backwards-incompatible change, in that code that caught the `RuntimeError` will not catch the new error.

# Description
Change RuntimeError to more appropriate values, as well as some
re-wording of error messages (although it is likely that some won't
ever fire). This includes and extends a change included in PR #63 in
_sherpa_init_xspec_library().

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -295,10 +295,9 @@ int _sherpa_init_xspec_library()
 
 
   if ( !getenv("HEADAS") ) {
-    // Raise appropriate error message that XSPEC initialization failed.
     PyErr_SetString( PyExc_ImportError,
 		     (char*)"XSPEC initialization failed; "
-		     "check HEADAS environment variable" );
+		     "HEADAS environment variable is not set" );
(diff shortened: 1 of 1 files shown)


olaurino:
This change is not backward compatible. If users had code that caught RuntimeException that code would not capture the new exceptions (I checked and it does not look like e.g. ValueError extends RuntimeError). As in a similar case in a different PR, I am not against this change as long as we are OK with the consequences and we properly stress them in the Release Notes.

DougBurke:
I view our use of `RuntimeError` in Sherpa as a crutch, so would like to replace them with more-semantically meaningful values (hence the PR). We have never guaranteed the exception raised by these routines, so whilst I agree documenting them is good, I don't think it's a significant breaking change.

One thing to think about is whether this module should be raising errors derived from the `sherpa.utils.err.SherpaErr` class rather than default Python ones.

PR #86, opened 2015-08-20
Bug #39 plotting counts vs channel

### LINKED ISSUE #39: Importing Bugs #13898 - plotting counts vs. channel
This is import a bug #13898 from an old bug tracking system.
 
Currently Sherpa plots of counts vs channel are given by 

set_analysis(1,"channel","counts", factor=0) 
plot_data() 

However, in this case the Y axis shows counts/channel and this is not what one expects.  Here the counts are divided by the channel number (or  channel number in a group), so for a high number of channel the values seem smaller.  What we want is to plot the counts in each channel.

I can generated such plot for PHA data set using 

add_curve(get_data().channel, get_data().counts)) 

In the case of grouped data this would need to be slightly different.

olaurino:
Resolve #39.
# Release Note

Sherpa incorrectly divided counts by the number of channels in a bin, so that, for instance, `get_counts()` would return `counts per channel` rather than `counts`. Similarly, the default `plot_data` function would show a plot of `counts/channel` vs `channel`. Now Sherpa always assumes that a channel always has width 1, even when it is the result of a grouping scheme, so that `get_counts().sum()` is invariant. The plot's `ylabel` has been updated accordingly from `counts/channel` to `counts`.
# Description

This is a rather simple fix, but I wonder whether there was a good reason for Sherpa to behave the way it did. I am pushing this sooner rather than later, so that we can start the review. So far I tested the fix visually, but I am adding more automated tests.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1507,8 +1507,8 @@ class DataPHA(Data1DInt):
                 ylabel += '/keV'
             elif self.units == 'wavelength':
                 ylabel += '/Angstrom'
-            elif self.units == 'channel':
-                ylabel += '/channel'
+            # elif self.units == 'channel':
+                # ylabel += '/channel'
 
         if self.plot_fac:
             ylabel += ' X %s^%s' % (self.units.capitalize(), str(self.plot_fac))
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1439,7 +1439,7 @@ class DataPHA(Data1DInt):
             elif self.units == 'wavelength':
                 ebin = self._hc/elo - self._hc/ehi
             elif self.units == 'channel':
-                ebin = ehi - elo
+                ebin = 1
             else:
                 raise DataErr("bad", "quantity", self.units)
 
--- a/sherpa/astro/tests/test_data.py
+++ b/sherpa/astro/tests/test_data.py
@@ -1,4 +1,4 @@
-# 
+#
 #  Copyright (C) 2007, 2015  Smithsonian Astrophysical Observatory
 #
 #
@@ -22,6 +22,7 @@ from sherpa.astro.data import DataPHA
 from sherpa.utils import SherpaTestCase, test_data_missing
 from sherpa.all import *
 from sherpa.astro.all import *
+from sherpa.astro import ui
 import logging
 logger = logging.getLogger('sherpa')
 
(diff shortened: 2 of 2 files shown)


olaurino:
@DougBurke, @anetasie are we sure that the `counts/channel` calculation was a bad idea? From Nick's question is seems so, but maybe it was intentional to have `counts/channel` just as one would have `counts/keV`.

I am just being the advocate's devil, as I think it is more rational to have 

Also, we must be sure that this change does not affect anything elsewhere in the code, although all the tests are passing.

DougBurke:
I am not at all convinced that you do not want coubbts/channel  but I
haven't really thought about it

olaurino:
@DougBurke @anetasie I need to be sure about the expected behavior of get_counts() and, in turn, of what should be plotted, so that I can add the tests. So I am holding off on this one for now.

anetasie:
This is tricky.

Sherpa has 'set_analysis' which defines the analysis space and the y-axis for plotting.
For ungroup data it is straightforward:
`set_analysis(1,'energy','counts',0)`
`set_analysis(1,'channel','counts',0)`
would result in plotting `counts`  on y axis for ungroup data.
After the data is grouped the y axis would contain `counts/channel`.

So the question is what should be plotted in the case of grouped data - total number of counts in a group at the center of a new group, or the number of counts in a group/number of channels that goes into that group. I was voting for total number of counts in a new group instead of the division per channel, because if you group to have 10 counts per group then you do not expect to see 2 counts in a new grouped spectrum. Note that this is only for the case when the analysis is set to 'channel' as one do want to see the correct counts/keV values per group, so the division needs to be performed here.

Following XSPEC - the units on the Y-axis in 'plot count' show 'counts/channel' but the number of counts is set to the total number of counts as summed over the channels belonging to the group with the new channel numbering.

olaurino:
Should we bring the discussion to sherpadev?

olaurino:
I resuscitated this PR as #86 was added to the priority list for 4.11.1. @DougBurke @anetasie any input on whether this is the desired change? I'd consider add more testing, but looking back at the conversation, it wasn't clear to me this is the desired change or not.

anetasie:
@olaurino In the previous chat I said:
_I was voting for total number of counts in a new group instead of the division per channel, because if you group to have 10 counts per group then you do not expect to see 2 counts in a new grouped spectrum._

The total number in a group is what one would expect in this case.

olaurino:
@anetasie there was an open action a while ago according to which you and @DougBurke needed to confirm this PR had the required fix or else what the requirement was. As we got no input on this for a while, we suggested we closed this PR, which we then did. I would like to make sure we are on the same page and know how to move forward.

anetasie:
There was a comment from @DougBurke at #38 about ignoring the setting at `set_analysis()`. Has this been resolved?

olaurino:
I don't think I understand what to do with this PR.

DougBurke:
There is #575 which looks related (at least some of the behavior).  I am a bit reluctant to make changes here until we have a bunch of commands and plots and agree that this is what we want to happen [*], since otherwise I think it's just chipping away at the problem and we are never sure we've not broken something elsewhere.

[*] and then turn these eventually into tests, which probably can be done in a plotting back-end agnostic manner, as we can check the attributes of the `*Plot` object. You still need backend tests, but I would hope that there's not too much logic in the `plot` method (or whatever it's called).

DougBurke:
I've created https://gist.github.com/DougBurke/57dc8dab328d4428240059c34d04d147 which has an example script and plot_data plots for ungrouped and grouped data to help explore the `plot_data` behavior (with matplotlib) for a range of `analysis`, `type`, and `factor` settings (for `set_analysis`).

PR #87, opened 2015-08-20
remove mutable default arguments

olaurino:
# Release Note
Some methods in `sherpa/fit.py` assigned mutable objects to default arguments. This has now been fixed. More instances of this issue have been identifies (Bug #95) and will removed in the future.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -299,8 +299,8 @@ class ErrorEstResults(NoNewAttributesAfterInit):
                'sigma', 'percent', 'parnames', 'parvals', 'parmins',
                'parmaxes', 'nfits')
 
-    def __init__(self, fit, results, parlist=[]):
-        if (parlist == []):
+    def __init__(self, fit, results, parlist=None):
+        if (parlist is None):
             parlist = [p for p in fit.model.pars if not p.frozen]
 
         from sherpa.estmethods import est_success, est_hardmin, est_hardmax, est_hardminmax
@@ -431,7 +431,9 @@ class ErrorEstResults(NoNewAttributesAfterInit):
 
 class IterFit(NoNewAttributesAfterInit):
 
-    def __init__(self, data, model, stat, method, itermethod_opts={'name': 'none'}):
+    def __init__(self, data, model, stat, method, itermethod_opts=None):
+        if itermethod_opts is None:
+            itermethod_opts = {'name': 'none'}
         # Even if there is only a single data set, I will
         # want to treat the data and models I am given as
         # collections of data and models -- so, put data and
@@ -515,7 +517,7 @@ class IterFit(NoNewAttributesAfterInit):
         return cb
 
     def primini(self, statfunc, pars, parmins, parmaxes, statargs=(),
-                statkwargs={}):
+                statkwargs=None):
         """An iterative scheme, where the variance is computed from
         the model amplitudes.
 
@@ -556,7 +558,8 @@ class IterFit(NoNewAttributesAfterInit):
                http://adsabs.harvard.edu/abs/1995ASPC...77..331K
 
         """
-
+        if statkwargs is None:
+            statkwargs = {}
         # Primini's method can only be used with chi-squared;
         # raise exception if it is attempted with least-squares,
         # or maximum likelihood
@@ -638,7 +641,7 @@ class IterFit(NoNewAttributesAfterInit):
         return final_fit_results
 
     def sigmarej(self, statfunc, pars, parmins, parmaxes, statargs=(),
-                 statkwargs={}):
+                 statkwargs=None):
         """Exclude points that are significately far away from the best fit.
 
         The `sigmarej` scheme is based on the IRAF `sfit` function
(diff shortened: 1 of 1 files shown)


DougBurke:
Do the tests exercise these arguments?

sherpa/fit.py
```diff
@@ -299,8 +299,8 @@ class ErrorEstResults(NoNewAttributesAfterInit):
                'sigma', 'percent', 'parnames', 'parvals', 'parmins',
                'parmaxes', 'nfits')
 
-    def __init__(self, fit, results, parlist=[]):
-        if (parlist == []):
+    def __init__(self, fit, results, parlist=None):
+        if (parlist is None):
```
DougBurke: The brackets aren't needed around the check ;-)

DougBurke:
I logged each function and ran tests (with ds9 and xspec), and here are the unique results (the number and first word of each line indicate which of the changes above created this output, and the rest of the string gives the parameter value on input)

```
1 ErrorEstResults parlist=[<Parameter 'fwhm' of model 'g1'>, <Parameter 'pos' of model 'g1'>, <Parameter 'ampl' of model 'g1'>]
1 ErrorEstResults parlist=[<Parameter 'fwhm' of model 'g'>, <Parameter 'pos' of model 'g'>, <Parameter 'ampl' of model 'g'>]
1 ErrorEstResults parlist=[<Parameter 'gamma' of model 'p1'>, <Parameter 'ampl' of model 'p1'>]
1 ErrorEstResults parlist=[<Parameter 'gamma' of model 'p1'>, <Parameter 'ampl' of model 'p1'>, <Parameter 'fwhm' of model 'g1'>, <Parameter 'pos' of model 'g1'>, <Parameter 'ampl' of model 'g1'>]
1 ErrorEstResults parlist=[<Parameter 'kT' of model 'mek1'>, <Parameter 'norm' of model 'mek1'>]
1 ErrorEstResults parlist=[<Parameter 'nH' of model 'abs1'>, <Parameter 'PhoIndex' of model 'p1'>, <Parameter 'norm' of model 'p1'>]
1 ErrorEstResults parlist=[<Parameter 'nH' of model 'xs1'>, <Parameter 'gamma' of model 'p'>, <Parameter 'ampl' of model 'p'>]

2 IterFit itermethod_opts={'name': 'none'}

5 fit statkwargs=None

6 Fit itermethod_opts={'name': 'none'}
6 Fit itermethod_opts=None
```

So, there seems to be no test of `primini` or `sigmarej`, and only the `Fit` class really exercises the different argument values (i.e. the empty/None case).

DougBurke:
Also at (a simple check for `=[` and `={` in files, but not all of them):

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/logposterior.py#L49

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/instrument.py#L72
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/instrument.py#L252
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/instrument.py#L330

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/models/template.py#L134

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/estmethods/__init__.py#L150
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/estmethods/__init__.py#L204
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/estmethods/__init__.py#L270

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/optmethods/__init__.py#L91

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/ui/utils.py#L166
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/ui/utils.py#L4925
https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/ui/utils.py#L6581

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/utils/numpytest.py#L554

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/utils/__init__.py#L556

https://github.com/sherpa/sherpa/blob/bug-fix-mutable-default-arguments/sherpa/astro/ui/utils.py#L9120

olaurino:
@DougBurke thanks for jump-starting my sherpa day! ;-)

olaurino:
An update on this: I looked into most of the issues that @DougBurke "grepped out". It's a slippery slope. Some of the updates would touch very old code that was not using idiomatic Python to begin with. A "safe" change of that code would probably be even less idiomatic, but a proper change would require a deeper review.

The point is that the mutable default arguments are always used read-only, as far as I can tell, and they are not usually used, since the methods are called by other sherpa functions without relying on the default argument. It is even unlikely for the user to do the wrong thing and get surprising results.

I will try to come up with a set of safe changes, but I'll create a new issue for those instances I won't fix. That issue will likely be open for some time, as it will need to wait for a deeper review of the code. I have been finding several places that would take advantage of such deep review, but given the current priorities I'll need to hold off on those ones as well.

DougBurke:
I've labelled this merge, and we can have a new issue for the remaining points @olaurino made.

Issue #46, opened 2015-06-16
BUG: sherpa.astro.ui.pack_pha/save_data fail with PyFITS backend for CIAO PHA files


DougBurke:
## Summary

The `ui.unpack_pha` and `ui.save_data` functions fail when:

a) the PyFITS backend is in use
b) the PHA file uses the LONGSTRN OGIP convention, as CIAO does

The error raised is `ValueError: Illegal value: ...`.

***EDIT***: it's not the `LONGSTRN` keyword that is the problem, it is the presence of "commentary" cards (i.e. key=HISTORY, COMMENT, or blank) which causes the problem. Chandra files have both COMMENT and HISTORY cards, and the failure reported above is due to the COMMENT cards that follow the `LONGSTRN` keyword in the header.

There are deprecation warnings, but these are mentioned in #27 so are not considered here.

## Details

I have installed pyfits using conda (using the conda.binstar.org/sherpa channel), so that I have

```
% conda list pyfits
# packages in environment at /export/local/anaconda/envs/sherpa-master:
#
pyfits                    3.3.0                np18py27_1
```

With the latest master branch of Sherpa, I get, when run from its top-level directory:

```
% ipython
Python 2.7.10 |Continuum Analytics, Inc.| (default, May 28 2015, 17:04:42)
Type "copyright", "credits" or "license" for more information.

IPython 3.1.0 -- An enhanced Interactive Python.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import sherpa

In [2]: sherpa.__version__
Out[2]: '4.7+427.g57e2d92'

In [3]: from sherpa.astro import ui
WARNING: imaging routines will not be available,
failed to import sherpa.image.ds9_backend due to
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available

In [4]: ui.load_pha('sherpa/astro/datastack/tests/data/3c273.pi')
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
but not used; to use them, re-read with use_errors=True
[...]

Issue #27, opened 2015-05-27
pyfits/astropy backend is using deprecated functionality


DougBurke:
# Summary

The pyfits back end uses deprecated functions and classes - it includes `new_table`, `CardList`, and `append` but I have not done an exhaustive check. Once #6 is resolved we should address these issues.

The error, whereby the Chandra PHA file can not be written out is discussed in #46.

# Details

The following test script:

```
% cat test_fits.py
# does writing out a FITS file cause warnings?
import sherpa
from sherpa.astro import ui
print("*** Sherpa version: {}".format(sherpa.__version__))
ui.load_arrays(1, [1,2,3], [4,5,6])
ui.save_data(1, 'test_fits.dat', ascii=True, clobber=True)
print("*** created test_fits.dat ***")
ui.save_data(1, 'test_fits.fits', ascii=False, clobber=True)
print("*** created test_fits.fits ***")
```

causes the following messages when writing the FITS version of the data. 

## When using pyfits

I meant to use the master branch, but used PR #26 by accident. As the code changes there are not functional, it is not the cause of the error (although the line numbering may be out slightly).

```
% conda list pyfits
# packages in environment at /home/naridge/local/anaconda/envs/sherpa-pep8:
#
You are using pip version 6.1.1, however version 7.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
pyfits                    3.3.0                np18py27_1  

% python test_fits.py 
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available
*** Sherpa version: 4.7+375.g1693f34
*** created test_fits.dat ***
/home/naridge/local/anaconda/envs/sherpa-pep8/lib/python2.7/site-packages/sherpa/astro/io/pyfits_backend.py:877: PyfitsDeprecationWarning: The new_table function is deprecated as of version 3.3 and may be removed in a future version.

        Use :meth:`BinTableHDU.from_columns` for new BINARY tables or :meth:`TableHDU.from_columns` for new ASCII tables instead.
  tbl = pyfits.new_table(pyfits.ColDefs(collist))
*** created test_fits.fits ***
```

## When using astropy

This is using PR #6 

```
% conda list astropy
# packages in environment at /home/naridge/local/anaconda/envs/sherpa-astropy:
#
[...]

PR #89, opened 2015-08-25
Bug #27 update pyfits deprecated interfaces

DougBurke:
# Release Note

The `astropy.io.fits`/`pyfits` interface used deprecated functionality. The code was updated to use the replacement classes/methods when available, falling back to the original code if not. The interfaces that were changed were, when the new symbols are availble, to:

 - use `astropy.io.fits.BinTableHDU.from_columns` rather than `astropy.io.fits.new_table`

- use `astropy.io.fits.Header` rather than `astropy.io.CardList`

# Code changes

The code went through some minor clean up (PEP8 style fixes and then some minor re-factoring to reduce repeated code), as well as changing from using the `pyfits` module name to `fits` (since astropy is now the preferred package) before removing the deprecated functionality. Fortunately for the routines used here, the replacement routines/classes have the same API, so it's an easy change.

The [PyFITS changelog](http://www.stsci.edu/institute/software_hardware/pyfits/release) was consulted - particularly for version 3.3 - to check what functionality we are using that has been deprecated.

# Testing

The full smoke test was run (with the testing submodule installed and with the XSPEC module compiled against version 12.9.0b and with ds9 available) using astropy version 1.0.4. I have *not* tested against pyfits (in particular to see if it works with old versions).

I used the following two scripts to check that the deprecation warnings are gone (I wasn't sure how/if this could be done as part of the smoke test). These are based on examples given in #27.

a) writing out a FITS binary table

```
# does writing out a FITS file cause warnings?
import sherpa
from sherpa.astro import ui
print("*** Sherpa version: {}".format(sherpa.__version__))
ui.load_arrays(1, [1,2,3], [4,5,6])
ui.save_data(1, 'test_fits.dat', ascii=True, clobber=True)
print("*** created test_fits.dat ***")
ui.save_data(1, 'test_fits.fits', ascii=False, clobber=True)
print("*** created test_fits.fits ***")
```

Before this PR this complained about

```
WARNING: AstropyDeprecationWarning: The new_table function is deprecated and may be removed in a future version.
        Use :meth:`BinTableHDU.from_columns` for new BINARY tables or :meth:`TableHDU.from_columns` for new ASCII tables instead. [astropy.utils.decorators]
```

whereas I know get the following (it's dirty because the `setup.cfg` is changed to build XSPEC and I haven't hidden it from git):

```
*** Sherpa version: 4.7+530.g6dacc53.dirty
*** created test_fits.dat ***
*** created test_fits.fits ***
```

b) writing out a PHA file

Unfortunately the PHA files we have available in sherpa all seem to use the `LONGSTRN` convention, which means that writing them out triggers #46 - but with

```
import sherpa
from sherpa.astro import ui
print("*** Sherpa version: {}".format(sherpa.__version__))

ui.load_pha(1, './sherpa/astro/datastack/tests/data/3c273.pi')
ui.save_data(1, 'test.fits', ascii=False, clobber=True)
print("*** created test.fits")
```

we used to get the following warnings:

```
WARNING: AstropyDeprecationWarning: The CardList class has been deprecated; all its former functionality has been subsumed by the Header class, so CardList objects should not be directly created.  See the PyFITS 3.1.0 CHANGELOG for more details. [astropy.io.fits.card]
WARNING: AstropyDeprecationWarning: The append function is deprecated and may be removed in a future version.
        Use :meth:`Header.append` instead. [astropy.utils.decorators]
```

### DIFF
--- a/sherpa/astro/io/meta.py
+++ b/sherpa/astro/io/meta.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2009  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2009, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -18,19 +18,17 @@
 #
 
 
-import inspect
-import sys
 from sherpa.utils import NoNewAttributesAfterInit
-from sherpa.utils.err import NotImplementedErr
 
 
 __all__ = ('Meta',)
 
 
 class Meta(NoNewAttributesAfterInit):
+    """A key-value store."""
 
     def __init__(self):
-        self.__header={}
+        self.__header = {}
         NoNewAttributesAfterInit.__init__(self)
 
     def __getitem__(self, name):
@@ -57,7 +55,7 @@ class Meta(NoNewAttributesAfterInit):
         return keys
 
     def has_key(self, key):
-        return self.__header.has_key(key)
+        return key in self.__header
 
     def values(self):
         return [self.__header[key] for key in self.keys()]
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -297,6 +297,29 @@ def _get_file_contents(arg, exptype="PrimaryHDU", nobinary=False):
     return (tbl, filename)
 
 
+def _find_binary_table(tbl, filename, blockname=None):
+    """Return the first binary table extension we find. If blockname
+    is not None then the name of the block has to match (case-insensitive
+    match), and any spaces are removed from blockname before checking.
+
+    Throws an exception if there aren't any.
+    """
+
+    if blockname is None:
+        for hdu in tbl:
+            if hdu.__class__ is fits.BinTableHDU:
+                return hdu
+
+    else:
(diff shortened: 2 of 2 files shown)


olaurino:
I updated the Release Note so that it can just be copied and pasted in the eventual release notes we are going to gather.

Also, in the release notes it might probably be a good idea (at least, this is what I believe @wmclaugh will ask me) to be more specific about the deprecated calls that were replaced.

You say:

    I have not tested against pyfits (in particular to see if it works with old versions).

This should have been done by Travis CI, right? but I can add more pyfits versions if necessary.

In order to avoid the `setup.cfg` issues (and the annoyance of having to remember not to commit it) you can use this command:
````
$ git update-index --assume-unchanged setup.cfg
````

If we ever change setup.cfg in the repo itself then you will need to undo this change to the git index and then reapply your changes. I included some documentation on this on our internal UDF.

DougBurke:
*I* have not done any testing against pyfits when testing this, but the
Travis CI tests will (at least one version, and have actually run by now).

DougBurke:
I've edited the release notes to add in an explicit note about what deprecated functionality has been replaced.

olaurino:
This is OK to merge as far as I am concerned. Thanks @DougBurke

Issue #41, opened 2015-06-15
Importing Old Issues -  incorporate the background component as an argument to calc_stat_func 


anetasie:
This is an issue sent by a user to the CXC helpdesk pointing out the limitations of the calc_stat_function.
This issue limits the user_stat functionality. There is no workaround..

"So I'm trying to define a w-statistic function for use in fitting some data, and I'm trying to follow the examples and documentation at http: //cxc.harvard.edu/sherpa/faq/user_stat.html and
http: //cxc.harvard.edu/sherpa/ahelp/load_user_stat.html, and the definition of the w-statistic at
http: //heasarc.nasa.gov/xanadu/xspec/manual/XSappendixStatistics.html (at
 the end of the second cstat section).

 My question is, how does a statistic-calculating function gain access to the
 backgrounds from the pha file? According to the documentation the function
 definition should look like
 ---
 def calc_stat_func(data, model, staterror=None, syserror=None, weight=None):
 ---
 and that doesn't have a "background" argument."

DougBurke:
This is addressed in #90

PR #90, opened 2015-08-25
add background data to user-defined statistics

dtnguyen2:
# Release Note
Added background data to the UserStat class 

# Description
This PR fixes issue #41 

Note, the user stat examples have to be modified to add the bkg arg to stat func
Can make backwards compatible by adding **kwargs instead of bkg=None as it is done now
no pep8 mod done

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -22,7 +22,8 @@ import logging
 import os
 import signal
 from numpy import arange, array, abs, iterable, sqrt, where, \
-    ones_like, isnan, isinf, float, float32, finfo, nan, any
+    ones_like, isnan, isinf, float, float32, finfo, nan, any, zeros, append, \
+    int
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
     bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import FitErr, EstErr, SherpaErr
@@ -31,8 +32,8 @@ from sherpa.estmethods import Covariance, EstNewMin
 from sherpa.models import SimulFitModel
 from sherpa.optmethods import LevMar, NelderMead
 from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, LeastSq, \
-    Likelihood
-
+    Likelihood, UserStat 
+import sherpa
 
 warning = logging.getLogger(__name__).warning
 info = logging.getLogger(__name__).info
@@ -448,6 +449,7 @@ class IterFit(NoNewAttributesAfterInit):
         # Data set attributes needed to store fitting values between
         # calls to fit
         self._dep = None
+        self.bkg = None
         self._staterror = None
         self._syserror = None
         self._nfev = 0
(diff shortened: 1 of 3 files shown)


sherpa/fit.py
```diff
@@ -31,8 +32,8 @@
 from sherpa.models import SimulFitModel
 from sherpa.optmethods import LevMar, NelderMead
 from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, LeastSq, \
-    Likelihood
-
+    Likelihood, UserStat
```
DougBurke: It looks like UserStat is not needed in this module.

sherpa/fit.py
```diff
@@ -31,8 +32,8 @@
 from sherpa.models import SimulFitModel
 from sherpa.optmethods import LevMar, NelderMead
 from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, LeastSq, \
-    Likelihood
-
+    Likelihood, UserStat 
+import sherpa
```
DougBurke: This import isn't used so can be removed.

sherpa/fit.py
```diff
@@ -482,7 +510,8 @@ def _get_callback(self, outfile=None, clobber=False):
             warning(e)
 
         self._dep, self._staterror, self._syserror = self.data.to_fit(
-            self.stat.calc_staterror)
+	    self.stat.calc_staterror)
+        self.bkg, junk1, junk2 = self.get_bkg_data( self._dep )
```
DougBurke: Can this be changed to

    self.bkg, _, _ = self.get_bkg_data(self._dep)

to be explicit that we do not care about the second and third items returned by the call?

sherpa/stats/__init__.py
```diff
@@ -147,7 +148,8 @@ def __init__(self, name='cash'):
         Likelihood.__init__(self, name)
 
     @staticmethod
-    def calc_stat(data, model, staterror=None, syserror=None, weight=None):
+    def calc_stat(data, model, staterror=None, syserror=None, weight=None, \
+                      bkg=None):
```
DougBurke: Should we verify that bkg is None here (to catch the case when a user is calling something incorrectly)? For instance, add

    if bkg is None:
        raise ValueError("The background can not be included in the statistic")

but with an improved error message?

Or is this limiting, and we may actually get cases where bkg is set here but its safe to ignore?

This holds for all the remaining stats which have a bkg parameter but do not use it.

sherpa/fit.py
dtnguyen2: The 2nd and 3rd arguments will be used for wstat ( data size and response
time, respectively).  They are needed for simultaneous fit of more then 1
data.

sherpa/stats/__init__.py
dtnguyen2: As suggested by Omar, I'm currently updating the user stat class so as
to be backwards compatible.  The new calc_stat will have the following
signature

def calc_stat(data, model, staterror=None, syserror=None, weight=None,
**kwargs):

The UserStat class may or may not have the bkg, see backwards
compatibility comment above.  So I don't think we should raise an
exception if there is not bkg data.  For the wstat, an exception shall
be raised since as you remember from the conversation with K. Arnaud
that the bkg data must be used

dtnguyen2: In the file sherpa/stats/__init__.py:

# truncation_flag indicates whether or not model truncation
# should be performed.  If true, use the truncation_value from
# the config file.
truncation_flag = config.get('statistics', 'truncate').upper()
truncation_value = float(config.get('statistics', 'trunc_value'))
if (bool(truncation_flag) is False or truncation_flag == "FALSE" or
    truncation_flag == "NONE" or truncation_flag == "0"):
    truncation_value = -1.0

If the truncation_value is not set by the user, we give it a default value
of -1.0.  Personally, I think it should be a small value (after all our
default value is 1.0e-25).  Setting it to -1.0 as it is now really changes
the stats.  As a reminder, truncation_value is used when the argument to
the log function is <=0.

PR #91, opened 2015-08-27
bug #46 handle commentary cards with astropy

DougBurke:
This PR has been replaced by #102 so has been closed.

# Release Notes

Fix for writing out FITS files using the `save_pha` and `save_table` commands - and the associated `pack_pha` and `pack_table` commands - when using the astropy I/O back end (issue #46). It does not affect CIAO users (i.e. users of the crates back end).

Adds a basic test suite for the sherpa.astro.io module.

# Notes

The primary fix is to handle the so-called commentary cards (i.e. those with a keyword of `HISTORY`, `COMMENT`, or blank) properly, when converting from the Sherpa data structure (the header field in the DataXXX object) to an astropy FITS header. This is a general-purpose change (so both FITS tables, images) but testing has focussed on the PHA case,  as that was where the issue was first seen. More tests are needed for sherpa.astro.ui, but that's outside the scope of this PR. I also wanted to see how the new tests work with the crates back end before extending them.

This conflicts with PR #89 (it re-uses part of these changes, but not all of them, so a little bit of merging will be required).


DougBurke:
@olaurino If you look at https://travis-ci.org/sherpa/sherpa/jobs/77506619 then you can see the test fails because it can't load the pyfits module. As this PR adds in tests for a "low level" module, then it could be that the way I'm importing code doesn't meet the expectations of the backend-handling code, but if so, why is it only in this particular way of building/testing Sherpa (the other builds for this run pass - e.g. https://travis-ci.org/sherpa/sherpa/builds/77506608 )?

olaurino:
For better of worse (should we change this?) you cannot import sherpa.astro.io if you don't have a backend installed. The failing build seems to be the one in which pyfits/astropy is not installed. There is one more build where the FITS I/O backend is not installed (x.7) but that build uses `import sherpa; sherpa.test()`, which is the CIAO smoke test. However, the issue with `sherpa.test()` (and one of the reasons why I implemented the new infrastructure) is that it does not fail when a backend cannot be imported.

The solution is to import sherpa.astro.io inside elements decorated with `@unittest.skipIf` including, if necessary, `setUp` methods.

DougBurke:
I can make this change.

DougBurke:
A probably-preferable alternative is to move the tests to sherpa.astro.ui (i.e. test the file I/O routines for this module), since it should cover the same code in sherpa.astro.io and get the ui code as well. I didn't do this originally because I wasn't sure I'd be able to test enough of the options. However, as the io routines are essentially just thin wrappers around the backend functionality, I think that moving the tests isn't that much of a problem, and saves issues here.

I'll look at whether this move makes sense, unless someone hollers.

olaurino:
@DougBurke if I understand your latest comment, generally speaking I would prefer them here, as unit tests. We should *also* test the ui with more unit and integration tests. On the other hand, pragmatically, you are right that moving the tests to sherpa.astro.ui may give us a little bit more coverage: it's a shortcut, one that might make sense at this point, but not "the right thing", imo.

DougBurke:
It depends - in part - on whether we think of the code in `sherpa.astro.io` as something we expect users to call, or if it is only intended for use by the higher-level modules. If it's the latter case, then we may not need unit tests on these since, at present at least, the higher-level code that calls these is basically just

  - check arguments
  - call the sherpa.astro.io routine

I think there's some benefit in identifying what parts of our code are expected to be called by outside code - in which case we may need to keep the API more stable - from those parts that are more internal, and may have a less-stable API. But that's a way-different ticket to this one.

DougBurke:
I've moved the tests to sherpa/astro/ui from sherpa/astro/io. I think this is ready for review now, but will probably want to be cleaned up (the commits squashed to hide the creation then removal of the io tests) before merging.

olaurino:
This should be reviewed again after #89 is merged, as they conflict. Also, #89 introduces a change in the way the pyfits module is imported, which will result in this PR to fail the tests.

DougBurke:
@olaurino I've redone this patch set to avoid the conflict (which also happens to tidy up the commit history). I am closing this PR since it has been replaced by #102

PR #94, opened 2015-08-27
Feature/wstat

dtnguyen2:
# Release Notes
Implement wstat as described at the following url:
https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixStatistics.html

# Description
The bkg file associated with the response file must be available in order to use wstat,
an exception is raised if bkg file is not available.
This PR is based on #90

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -32,8 +32,7 @@ from sherpa.estmethods import Covariance, EstNewMin
 from sherpa.models import SimulFitModel
 from sherpa.optmethods import LevMar, NelderMead
 from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, LeastSq, \
-    Likelihood
-
+    Likelihood, WStat
 
 warning = logging.getLogger(__name__).warning
 info = logging.getLogger(__name__).info
@@ -478,6 +477,7 @@ class IterFit(NoNewAttributesAfterInit):
         bkg_dep = []
         data_size = None
         response_time = None
+
         len_datasets = len(self.data.datasets)
 
         data_size = zeros(len_datasets, dtype=int)
@@ -493,8 +493,8 @@ class IterFit(NoNewAttributesAfterInit):
                 bkg_dep = append(bkg_dep, tmp_bkg_dep)
                 response_time[2 * index] = mydata.exposure
                 response_time[2 * index + 1] = bkg.exposure
-            # else:
-            #     raise FitErr( 'no bkg file is supplied, use cstat instead' )
+            elif type(self.stat) is WStat:
+                raise FitErr('no bkg file is supplied, use cstat instead')
 
         return bkg_dep, data_size, response_time
 
@@ -511,7 +511,11 @@ class IterFit(NoNewAttributesAfterInit):
 
         self._dep, self._staterror, self._syserror = self.data.to_fit(
             self.stat.calc_staterror)
-        self.bkg, junk1, junk2 = self.get_bkg_data(self._dep)
+
+        self.bkg, datasize, responsetime = self.get_bkg_data(self._dep)
+        if type(self.stat) is WStat:
+            self._staterror = datasize
+            self._syserror = responsetime
 
         self._nfev = 0
         if outfile is not None:
@@ -958,7 +962,10 @@ class Fit(NoNewAttributesAfterInit):
         model = self.data.eval_model_to_fit(self.model)
         bkg = None
 
-        bkg, _, _ = self._iterfit.get_bkg_data(dep)
+        bkg, datasize, responsetime = self._iterfit.get_bkg_data(dep)
+        if type(self.stat) is WStat:
+            staterror = datasize
+            syserror = responsetime
         if 0 == len(bkg):
             return self.stat.calc_stat(dep, model, staterror, syserror)[0]
         else:
@@ -1029,7 +1036,7 @@ class Fit(NoNewAttributesAfterInit):
     def fit(self, outfile=None, clobber=False):
         dep, staterror, syserror = self.data.to_fit(self.stat.calc_staterror)
         if not iterable(dep) or len(dep) == 0:
-            #raise FitError('no noticed bins found in data set')
+            # raise FitError('no noticed bins found in data set')
             raise FitErr('nobins')
 
         if ((iterable(staterror) and 0.0 in staterror) and
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -23,7 +23,7 @@ import os
 import signal
 from numpy import arange, array, abs, iterable, sqrt, where, \
     ones_like, isnan, isinf, float, float32, finfo, nan, any, zeros, append, \
-    int
+    int, ones
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
     bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import FitErr, EstErr, SherpaErr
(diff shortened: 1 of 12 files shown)


sherpa/stats/__init__.py
```diff
@@ -474,29 +492,93 @@ def __getstate__(self):
...
+    .. [1] The description of the W statistic (`wstat`) in
+           https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixStatistics.html
+
+    """
+
+    def __init__(self, name='wstat'):
+        Likelihood.__init__(self, name)
+
+    @staticmethod
+    def calc_stat(data, model, staterror=None, syserror=None,
```
cdeil: Thanks for implementing this!

As a future user of this feature, I'd like to ask / comment:

This is shoehorning wstat into a `Likelihood.calc_stat` interface that was designed for chi2 stats, right?
I.e. arguments are re-used for something completely different than their name says: `exposuretime = syserror`

I guess there is no better way to get wstat into the existing API?
If so, can you at least explain what's going on here in the docstring?

dtnguyen2: If I make the following changes:

    @staticmethod
    def calc_stat(data, model, datasize=None, exposuretime=None,
                  weight=None, **kwargs):

        if 'bkg' in kwargs.keys():
            bkg = kwargs.get('bkg')
        else:
            raise FitErr('no bkg file is supplied, use cstat instead')
        return _statfcts.calc_wstat_stat(data, model, datasize,
exposuretime,
                                         bkg, truncation_value)

Then it breaks symmetry with the other classes within the stats/__init__.py
file and we currently call calc_stat with 4 arguments (without keywords)
but I don't think I'm totally against this.  Hopefully no future
modifications will require the staterror and syserror keywords.

BTW, I thought about adding  two additional arguments datasize and
exposuretime to the calc_stat method of the various Statistic classes but
the problem with this approach is that it breaks backwards compatibility
with the user's existing user-stat scripts or it  makes the logical if
statement in fit.py too complicated and be error prone.

sherpa/fit.py
```diff
@@ -470,9 +471,36 @@ def __init__(self, data, model, stat, method, itermethod_opts={'name': 'none'}):
...
+            mydata = self.data.datasets[index]
+            if hasattr(mydata, 'response_ids') and \
+                    hasattr(mydata, 'background_ids') and \
+                    len(mydata.response_ids) and len(mydata.background_ids):
+                bkg = mydata.get_background(mydata.background_ids[0])
+                tmp_bkg_dep = bkg.get_dep(True)
+                data_size[index] = tmp_bkg_dep.size
+                bkg_dep = append(bkg_dep, tmp_bkg_dep)
+                exposure_time[2 * index] = mydata.exposure
+                exposure_time[2 * index + 1] = bkg.exposure
```
DougBurke: For the wstat case, a simple hack to include the BACKSCAL ratios, would be

```
                try:
                    bexp = bkg.exposure * bkg.backscal / mydata.backscal
                except AttributeError:
                    bexp = bkg.exposure
                exposure_time[2 * index + 1] = bexp
```

which seems to give me the same stat values as XSPEC does. I do not like this solution, as (as also mentioned by @cdell) this feels like trying to work around the limitations of an interface, but I don't have any concrete suggestions for improvements.

DougBurke:
As mentioned by @anetasie we need to add WStat support to the MCMC/pyBLoCXS code. I had to add `WStat` to https://github.com/sherpa/sherpa/blob/f4e8aa346a9f577a8d435eb46a6fe85015dc9e18/sherpa/sim/__init__.py#L27 and https://github.com/sherpa/sherpa/blob/f4e8aa346a9f577a8d435eb46a6fe85015dc9e18/sherpa/sim/__init__.py#L494 (and the text of the error message on the next line)

DougBurke:
And added to https://github.com/sherpa/sherpa/blob/f4e8aa346a9f577a8d435eb46a6fe85015dc9e18/sherpa/plot/__init__.py#L76

DougBurke:
There is a problem when re-grouping the data (I have not checked against yesterday's check in yet). If the grouping is changed then you get an error saying

    TypeError: statistic input array sizes do not match

whenever the statistic is next calculated (e.g. by `fit` or `calc_stat`). An example script is shown below. My guess is that the grouping isn't being applied to the background (and, if so,  it may be tricky to work out how best to handle this).

# Example

The program

```
% cat doit.py
from sherpa.astro import ui
ui.load_pha(1, 'sherpa/astro/datastack/tests/data/3c273.pi')
ui.set_stat('wstat')
ui.set_method('simplex')
ui.notice_id(1, 0.5, 7)
ui.set_source(1, ui.xsphabs.gal * ui.xspowerlaw.pl)
pl.norm = 1e-4
s1 = ui.calc_stat(1)
print("stat 1 = {}".format(s1))
ui.group_counts(1, 5)
s2 = ui.calc_stat(1)
print("stat 2 = {}".format(s2))
```

gives

```
% python doit.py
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273.pi' 
read ARF file sherpa/astro/datastack/tests/data/3c273.arf
read RMF file sherpa/astro/datastack/tests/data/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi' 
read background file sherpa/astro/datastack/tests/data/3c273_bg.pi
stat 1 = 1081.43303654
  File "doit.py", line 11, in <module>
    s2 = ui.calc_stat(1)
  File "<string>", line 1, in calc_stat
  File "/lagado.real/sherpa/sherpa-wstat/sherpa/ui/utils.py", line 7830, in calc_stat
    return f.calc_stat()
  File "/lagado.real/sherpa/sherpa-wstat/sherpa/fit.py", line 990, in calc_stat
    bkg=bkg_data)[0]
  File "/lagado.real/sherpa/sherpa-wstat/sherpa/stats/__init__.py", line 585, in calc_stat
    truncation_value)
TypeError: statistic input array sizes do not match
```

If I manually hack the background to have the same grouping as the data then it works. Adding in the following after the `group_counts` call means that the program runs:

````
d = ui.get_data(1)
b = ui.get_bkg(1) 
b.grouping = d.grouping
b.mask = d.mask
```

which gives

```
% python doit2.py
...
stat 1 = 1081.43303654
stat 2 = 1161.32718342
```

dtnguyen2:
I'm pretty sure this behavior has been around for a while. When you sent
out the bug report about wstat misbehaving after calc_data_sum, I had a
fixed for it forcing all filter and masks for data and background to match
but we decided that we will address this issue outside of wstat etc... So
perhaps we should think about fixing group instead?

dtnguyen2:
I implemented the wstat verbatim from the following url:
heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixStatistics.html
However, recently I stumbled on the following url from K. Arnaud:
heasarc.gsfc.nasa.gov/docs/xanadu/xspec/wstat.ps
On page 5, it says something to the effect: 
"...to give a quadratic in b_i whose solution is:....."
I have to look at this more closely but it looks like he is using the grade school formula 
(-b +/- sqrt(b^2-4ac))/2a for one of the roots (see equation a_i) of quadratic equation which is not numerically stable.

DougBurke:
@dtnguyen2 one interesting project on converting equations into more-stable forms is herbie http://herbie.uwplse.org/ (here's one link to it http://cs.stackexchange.com/questions/44263/program-transformations-for-numeric-stability but I've been tracking it for some time, just never got around to doing anything with it).

PR #101, opened 2015-09-11
Fix handling of non-contiguous grids in XSPEC models

### LINKED ISSUE #42: Importing Old Issues - Failures in XSPEC regression tests
This is an issue from the  CIAO 4.7 ECR related to the test failures.

"We have noticed some sporadic Xspec failures when running the sherpa regression tests. For example, when some models are  called multiple times or with certain array sizes, they may  return arrays of zeros or nans. The issue also exists in  CIAO 4.6 using Xspec 12.8.0. "

The status of the tests should be checked after  PR #34  is merged.

### LINKED ISSUE #62: XSpec model evaluation for non-continuous grids (Xspec models that call sumdem, such as apec, can crash)

DougBurke:
# Release Note

Fixes handling of non-contiguous bins - i.e. when a model is called with both xlo and xhi arguments but the bins do not fully cover the energy, or wavelenth, range. This fixes #62 (for XSPEC 12.8.2; switching to XSPEC 12.9.0 should also fix it) and #56. It also fixes an (un-reported) problem with handling of non-contiguous grids when using a table model, where a crash was likely.

When an XSPEC model is called with both low and high values for the grid - i.e. with two arguments - then the two arrays are checked to have the same length, and a ValueError is raised if this condition does not hold. This is a breaking change, but the results are not guaranteed to be correct if the two arrays are not the same length.

The experimental interface for XSPEC convolution models has changed, so that the function call takes pars, fluxes, xlo, with optional xhi whereas before it was pars, xlo, xhi, fluxes. This is a breaking change, but this is in the low-level API that is not documented to users, and adds useful functionality (the ability to have xhi be optional). The cpflux convolution model has been added. Note that these models do not have Python classes associated with them as they are still an experimental interface. 

The test suite has been updated to test the new and changed functionality in this PR. The choice of models is made so as to avoid known problematic models (with a version check where relevant).

It is believed that the changes in this PR fixes #42, although this is hard to prove conclusively as the cause of #42 have never been identified, but the clean up and fixes here are likely fixes.

# Other notes

This has been tested with XSPEC libraries 12.8.2q and 12.9.0b.

The code has seen a significant change, to improve clarity and to re-inforce the similarites between the various templates. A future change could be to re-write the templates to avoid the existing code duplication - as suggested by @dtnguyen2 - but that is outside the scope of this PR.

No performance testing has been done to see if the new approach is slower than the old one. Any performance tweaks should be done after the code has been consolidated, if found to be necessary.

The code for convolution models has been separated from that for "C-style" models as it is a lot cleaner to do so.

This should complete the set of fixes taken from #63, although the code is slightly different as different choices were made (based on experience developing both patches). Differences to those parts of #63 not already merged are:

- In #63 the grid sent in to "float *" models was converted straight away to float, rather than being handled as a double until it needed to be converted to float. This lead to some surprising behavior when using a model expression that contained both single- and double-precision models, so for nw this change has been scrapped
- #63 added checks for the grid being monotonic and to not have overlaps; these have been removed since at present it's not obvious that this is not going to cause problems due to float/double conversion tolerances (related to the previous point)
- the test suite is a lot simpler, since it avoids the comparison of the results of a non-contiguous array with the "parent" contiguous array (i.e. the grid from which the non-contiguous grid was created). This comparison is problematic since although the results should be the same in principle, in practice the XSPEC code often re-bins data, or uses the existing grid it is sent, so that you can get larger-than expected differences from the comparison. This lead to a confusing set of tolerances and manual choices of models to test or not in #63. The new approach is to compare the results for non-contiguous grids when evaluated on energy and wavelength grids, which I think is a sensible compromise here (given the other checks the test suite does for the contiguous case)

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec.py
+++ b/sherpa/astro/xspec/tests/test_xspec.py
@@ -122,6 +122,7 @@ class test_xspec(SherpaTestCase):
 
     def setUp(self):
         from sherpa.astro import xspec
+        ui.clean()
         self._defaults = xspec.get_xsstate()
 
     def tearDown(self):
@@ -187,7 +188,13 @@ class test_xspec(SherpaTestCase):
         import sherpa.astro.xspec as xs
         mdl = xs.XSpowerlaw()
 
+        # Check when input array is too small (< 2 elements)
         self.assertRaises(TypeError, mdl, [0.1])
+
+        # Check when input arrays are not the same size (when the
+        # low and high bin edges are given)
+        self.assertRaises(TypeError, mdl, [0.1, 0.2, 0.3], [0.2, 0.3])
+        self.assertRaises(TypeError, mdl, [0.1, 0.2], [0.2, 0.3, 0.4])
 
     def test_xspec_models(self):
         import sherpa.astro.xspec as xs
@@ -228,6 +235,23 @@ class test_xspec(SherpaTestCase):
             assert_allclose(evals1, wvals1,
                             err_msg=emsg + "energy to wavelength")
 
+    def test_tablemodel_checks_input_length(self):
+
+        # see test_table_model for more information on the table
+        # model being used.
+        #
+        ui.load_table_model('mdl',
+                            self.make_path('xspec/tablemodel/RCS.mod'))
+        mdl = ui.get_model_component('mdl')
+
+        # Check when input array is too small (< 2 elements)
+        self.assertRaises(TypeError, mdl, [0.1])
+
+        # Check when input arrays are not the same size (when the
+        # low and high bin edges are given)
+        self.assertRaises(TypeError, mdl, [0.1, 0.2, 0.3], [0.2, 0.3])
+        self.assertRaises(TypeError, mdl, [0.1, 0.2], [0.2, 0.3, 0.4])
+
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_xspec_tablemodel(self):
         # Just test one table model; use the same scheme as
@@ -243,6 +267,8 @@ class test_xspec(SherpaTestCase):
         # when used in the test suite it appears that the tmod
         # global symbol is not created, so need to access the component
         tmod = ui.get_model_component('tmod')
+
+        self.assertEqual(tmod.name, 'xstablemodel.tmod')
 
         egrid, elo, ehi, wgrid, wlo, whi = make_grid()
 
--- a/sherpa/astro/xspec/tests/test_xspec.py
+++ b/sherpa/astro/xspec/tests/test_xspec.py
@@ -291,6 +291,73 @@ class test_xspec(SherpaTestCase):
         # assert_allclose(evals2, wvals2, rtol=rtol,
         #                 err_msg=emsg + "two args")
 
+    def test_convolution_model_cflux(self):
+        # Use the cflux convolution model, since this gives
+        # an easily-checked result. At present the only
+        # interface to these models is via direct access
+        # to the functions (i.e. there are no model classes
+        # providing access to this functionality).
+        #
+        import sherpa.astro.xspec as xs
+
+        # The energy grid should extend beyond the energy grid
+        # used to evaluate the model, to avoid any edge effects.
+        # It also makes things easier if the elo/ehi values align
+        # with the egrid bins.
+        elo = 0.55
+        ehi = 1.45
+        egrid = numpy.linspace(0.5, 1.5, 101)
+
+        mdl1 = xs.XSpowerlaw()
+        mdl1.PhoIndex = 2
+
+        # flux of mdl1 over the energy range of interest; converting
+        # from a flux in photon/cm^2/s to erg/cm^2/s, when the
+        # energy grid is in keV.
+        y1 = mdl1(egrid)
+        idx, = numpy.where((egrid >= elo) & (egrid < ehi))
+
+        # To match XSpec, need to multiply by (Ehi^2-Elo^2)/(Ehi-Elo)
+        # which means that we need the next bin to get Ehi. Due to
+        # the form of the where statement, we should be missing the
+        # Ehi value of the last bin
+        e1 = egrid[idx]
+        e2 = egrid[idx + 1]
+
+        f1 = 8.01096e-10 * ((e2 * e2 - e1 * e1) * y1[idx] / (e2 - e1)).sum()
+
+        # The cflux parameters are elo, ehi, and the log of the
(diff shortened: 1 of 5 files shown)


dtnguyen2:
I'm ok with this

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -50,60 +166,306 @@ def test_create_model_instances(self):
...
+            mdl = cls(model)  # use an identifier in case there is an error
+
+            evals2 = mdl(elo, ehi)
+            wvals2 = mdl(wlo, whi)
+            emsg = "{} non-contiguous model evaluation failed: ".format(model)
+
+            assert_allclose(evals2, wvals2,
+                            err_msg=emsg + "energy to wavelength")
+
+    def test_tablemodel_checks_input_length(self):
```
DougBurke: This is missing the decorator

    @unittest.skipIf(test_data_missing(), "required test data missing")

PR #102, opened 2015-09-14
Bug #46 astropy handle commentary cards

DougBurke:
This replaces #91 - the commit history is a lot cleaner and does not conflict with the `master` branch.

Resolve #46 

# Release Notes

Fix for writing out FITS files using the save_pha and save_table commands - and the associated pack_pha and pack_table commands - when using the astropy I/O back end (issue #46). It does not affect CIAO users (i.e. users of the crates back end).

Fix for when the `notice2d_id`, `notice2d_image`, and the `ignore` version functions are called with an invalid identifier (i.e. an identifier that i snot an integer or string value). The error is now an `ArgumentTypeErr` with the message "'ids' must be an identifier or list of identifiers". It was a `NameError` with the message "global name '_argument_type_error' is not defined".

Improves the `sherpa/astro/ui/tests/test_astro_ui.py` test suite.

# Notes

The primary fix is to handle the so-called commentary cards (i.e. those with a keyword of HISTORY, COMMENT, or blank) properly, when converting from the Sherpa data structure (the header field in the DataXXX object) to an astropy FITS header. This is a general-purpose change (so both FITS tables, images) but testing has focussed on the PHA case, as that was where the issue was first seen. More tests are needed for sherpa.astro.ui, but that's outside the scope of this PR.

The NameError fix was just that the _argument_type_error routine has not been imported from the necessary module.

The code has seen PEP-8 fixes and minor refactorings.

The tests are not added to `sherpa/astro/io/tests/test_astro_io.py` since there are issues with

PR #105, opened 2015-09-19
Ensure astropy and crates behave the same with gzipped files

DougBurke:
# Release Notes

Change the behaviour of the AstroPy back end so that it matches that of Crates when given a file name which does not exist, but a compressed version, with the suffix `.gz` does. The Crates behavior is to read the file. This extends to PHA files whose ancillary files - e.g. those stored in the `BACKFILE`, `ANCRFILE`, and `RESPFILE` keywords - are given as unzipped names, but only the gzipped names exist on disk.

As an example: if `pha.fits.gz` exists but `pha.fits` does not, then

    load_pha('pha.fits')

will now load the file with either back end. If the response files are set to `arf.fits` and `rmf.fits` (via the `ANCRFILE` and `RESPFILE` keywords), but only the `.gz` versions exist, then they will now also be loaded by the AstroPy back end. 

Tests have been added to cover this case.

# Notes

This requires an update to `sherpa-test-data`, which adds several gzipped files for the test suite.

I have checked the changes against a CIAO build (with crates) and they pass.

This commit depends on PR #102 (although with the rebase to include #108 this may now not be as obvious).

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2007  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2007, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -17,7 +17,15 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-from itertools import izip, repeat
+"""
+Provide I/O routines to Sherpa, focussing on FITS.
+
+This module should not be imported directly if there is no
+FITS support present (i.e. neither the crates or astropy
+back ends are in use).
+"""
+
+from itertools import izip
 import logging
 import os
 import os.path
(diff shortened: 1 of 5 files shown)


DougBurke:
The `sherpa-test-data` update is sherpa/sherpa-test-data#2

DougBurke:
The test failures all appear to be due to

```
py.error.EACCES: [Permission denied]: mkdir('/home/.cache',)
```

which would seem to indicate it's not due to my commit. At least, it's not *obvious to me* what the cause of this failure is: is it a Travis-CI issue (or an update to one of the Python modules we use during testing), or is it that I need to add the new test to a file (e.g. a manifest), or ...?

DougBurke:
I also see the same failures in PR #107 which suggests that it's a setup issue (since that PR only changes existing files rather than adds new ones).

DougBurke:
Note that the test failure here was a problem with the Travis-CI test suite and pytest-2.8.0. It has been fixed in #108.

DougBurke:
Merging in #108 has fixed the test failures.

DougBurke:
The test failures are not due to the code changes here, but some configuration issue with the Travis tests (the failures are `conda not found`). The `.travis.yml` file here is a bit behind the `master` version, which could be the problem.

DougBurke:
I am closing this in favour of #151.

PR #107, opened 2015-09-20
Normalize plot labels (PHA data sets)

DougBurke:
# Release Notes
    
There are two main changes for plots of PHA data sets (and related quantities, such as the source, model, and ARF):
    
- plots with matplotlib now use the LaTeX support - so that 'cm^2' is
  now displayed as a superscript; previously they were displayed
  directly. This does not change the display with the ChIPS backend.
    
- plots created with `plot_source` used a different format to other
  plots when analysis=wavelength, in that LaTeX symbols were used for
  Angstrom and lambda (in other plots the string 'Angstrom' is used
  instead). The source plots now match the other plots.

There have also been a number of PEP 8 changes that do not change the behaviour of the code.

# Notes

The choice to replace `\AA` by the word `Angstrom` was the easier change. It would also be possible to go the other way - i.e. replace `Angstrom` by `\AA`, but it would have been a larger change, and so requires some discussion. This can be done after this PR is merged.

The plot labels are created in multiple places - e.g. data sets (`sherpa.astro.data`) and within the plot code - which makes the logic of where labels are created a bit hard to track down.

# Rebase

The branch was rebased on Sep 23 2015 10:18AM EST to pull in the Travis-CI fixes from PR #108.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2008  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2008, 2015  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


DougBurke:
I've put example before and after `plot_source` outputs with the matplotlib backend at https://gist.github.com/DougBurke/eaed0e3ba9f63832f6cc

DougBurke:
Note that the test failure here was a problem with the Travis-CI test suite and pytest-2.8.0. It has been fixed in #108.

DougBurke:
Merging in #108 has fixed the test failures.

anetasie:
I guess you did not attempt to update the location of the plot title which is running a bit too close to the axis labels. Would this update require a change in all the plots?

olaurino:
Rebasing, original tip was at 362ff1e

olaurino:
Aside from the question on the opportunity of creating a new public method `to_latex` I think this is OK to be merged.

olaurino:
@DougBurke any thoughts on the opportunity of removing the public method "to_latex"? I don't have strong feelings against it, so I'll let you decide.

olaurino:
I'll merge it as is.

PR #109, opened 2015-09-26
Support matplotlib 1.5RC1

### LINKED ISSUE #103: Can not create plots if use matplotlib 1.5
Using the rc1 version of matplotlib, I am unable to plot since the pylab backend is reported not to have loaded, even though it does work (i.e. the plot from line 1 is displayed), as shown below. Note that the test suite (`python setup.py test`) passes, so it's obviously not testing out the plotting backends particularly well.

```
% ipython --pylab
Python 2.7.10 |Continuum Analytics, Inc.| (default, Sep 15 2015, 14:50:01) 

IPython 4.0.0 -- An enhanced Interactive Python.
%quickref -> Quick reference.
Using matplotlib backend: Qt4Agg


In [1]: plt.plot([1,2,5,8])
Out[1]: [<matplotlib.lines.Line2D at 0x7f20b44c8d90>]

In [2]: from sherpa.astro import ui
WARNING: failed to import sherpa.plot.pylab_backend; plotting routines will not be available
```

The problem (or at least the first one, there could be more) is that the way we access default options no-longer works in 1.5.

```
In [1]: import sherpa.plot.pylab_backend
WARNING: failed to import sherpa.plot.pylab_backend; plotting routines will not be available
KeyError                                  Traceback (most recent call last)
<ipython-input-1-06f45ada1491> in <module>()
----> 1 import sherpa.plot.pylab_backend

/home/naridge/data/my-sherpa/sherpa-matplotlib/sherpa/plot/pylab_backend.py in <module>()
     95           overplot=False, clearwindow=True,
     96           yerrorbars=False,
---> 97           ecolor=_errorbar_defaults['ecolor'],
     98           capsize=_errorbar_defaults['capsize'],
     99           barsabove=_errorbar_defaults['barsabove'],

KeyError: 'ecolor'
```

From earlier in the file we have

```
_errorbar_defaults = get_keyword_defaults(pylab.Axes.errorbar)
```

Using matplotlib 1.5rc1 I get

```
In [2]: import pylab

In [3]: import sherpa.utils

In [4]: sherpa.utils.get_keyword_defaults(pylab.Axes.errorbar)
Out[4]: {}
```

whereas the output for 1.4.3 I get

```
In [1]: import pylab

In [2]: import sherpa.utils

In [3]: sherpa.utils.get_keyword_defaults(pylab.Axes.errorbar)
Out[3]: 
{'barsabove': False,
 'capsize': 3,
 'capthick': None,
 'ecolor': None,
 'elinewidth': None,
 'errorevery': 1,
 'fmt': u'',
 'lolims': False,
 'uplims': False,
 'xerr': None,
 'xlolims': False,
 'xuplims': False,
 'yerr': None}

```

I installed 1.5rc1 via

```
% conda create -c tacaswell python=2.7 astropy ipython matplotlib
```

and the issue is probably related to the 'Axes and artist' point at http://matplotlib.org/devdocs/api/api_changes.html#changes-in-1-5-0

I'm hoping that it's just a case of finding where to extract this information from (and that it's a backwards-compatible change).

### LINKED ISSUE #113: numpy 1.10: histo test failure
# Test failure

Using numpy 1.10.0, the test_histo test will fail with

```
E   TypeError: histogram() got an unexpected keyword argument 'new'
```

This is in `sherpa/astro/tests/test_astro.py`.

# Warning note, possibly not relevant

The numpy 1.9 release notes - currently at http://docs.scipy.org/doc/numpy/release.html as I couldn't find a versioned URI - has this in the 'Future changes' section

- Default casting for inplace operations will change to ‘same_kind’ in Numpy 1.10.0. This will certainly break some code that is currently ignoring the warning.

The Sherpa code has been somewhat lax about numpy warnings - some modules override the current value "globally" (i.e. without restricting to a particular function), and this can lead to odd behaviour depending on the order modules are imported. I have not checked to see if we have any code that ignores the warning mentioned above, but it should probably be done since numpy 1.10.0 has just been released: https://mail.scipy.org/pipermail/scipy-user/2015-October/036709.html

A run of the test suite with numpy 1.10.0 doesn't seem to show any problem related to this particular change.

DougBurke:
Resolve #103.
Resolve #113.

# Release Notes

Support matplotlib 1.5 RC1.

# Details

Switch extraction of keyword defaults from `pylab.Axes.errorbar` to `pylab.errorbar` to support running with matplotlib 1.5RC1. See #103 for details. This should be a backwards-compatible change.

This change allows the matplotlib backend to be loaded and a simple plot created (by `plot_data`). I have not done any investigation to see if further changes may be needed.

Also, in order to make the tests with matplotlib 1.5 pass, we had to fix outstanding issue #113.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,11 @@ addons:
       - flex
       - bison
 
+# We use matrix include so that xspec builds (which are more complicated and long) are started first
 matrix:
   fast_finish: true
   include:
-    - env: XSPEC="true" FITS="astropy" INSTALL_TYPE=develop TEST=submodule
+    - env: XSPEC="true" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5
       python: "2.7"
       sudo: required
       dist: trusty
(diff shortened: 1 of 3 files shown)


olaurino:
Rebasing. Original tip was at b705d52

olaurino:
@DougBurke I added a test and reviewed the travis builds. Would you like to review my changes or should I just merge this?

DougBurke:
The only real question I have is if the test you added does fail without the change (i.e. I'm being paranoid and want to know if the test is actually catching this case, because I have been known to write tests that end up not testing what I think they're testing!). However, I'm not paranoid enough to say that this should stop this being merged.

olaurino:
That's a very fair question, so much fair that yes, I did test that. It's just a matter of 

~~~~
git checkout master -- sherpa/plot/pylab_backend.py
python setup.py -q test --pytest-args="sherpa/plot/tests/test_pylab.py"
~~~~

Of course you need to make sure you have matplotlib 1.5 installed.

To get back to a clean state:
~~~~
git reset --hard 
~~~~

As I said before I usually try to have a failing test before I commit a fix. In this case I had to do the other way around, as the fix was already there, but it's close enough :)

PR #110, opened 2015-09-28
datastack module documentation update

DougBurke:
# Release Notes

Update the `sherpa.astro.datastack` module documentation to include information from the
CIAO ahelp documentation and to match the style used by the `sherpa.astro.ui` module.

# Details

There are no code changes here, only documentation updates (and moving the
copyright statement to the start of the file). The documentation changes
include a major expansion of the module-level docstring, addition of
various functions, and updating to numpy-style formatting. The new text
is based on the CIAO ahelp documentation for the datastack module.

### DIFF
--- a/sherpa/astro/datastack/__init__.py
+++ b/sherpa/astro/datastack/__init__.py
@@ -32,6 +32,14 @@ commands used to load data, set source models, get and set parameters,
 fit, and plot. It is intended for use when simultaneously fitting a
 common source model to multiple data sets.
 
+Acknowledgements
+----------------
+
+The Datastack code was developed by Thomas Aldcroft and originally
+provided as an external package for Sherpa, available at
+http://cxc.harvard.edu/contrib/datastack/. The code was added to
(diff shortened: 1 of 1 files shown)


anetasie:
We need to give credit to Tom Aldcroft, as the code is basically taken  as was  written and designed in his datastack module.

DougBurke:
I actually looked at this, and found this numpy guideline suggesting that
authors shouldn't be in docstrings

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-modules

We have a similar issue with the pyblocxs code in #111 which I haven't
really addressed either

olaurino:
I am OK with merging this PR, as long as we have a plan for giving credit to the author of the original source code, since the PR removed this information.

DougBurke:
I have added an acknowledgments section. Longer term we need a better solution (e.g. once the planned on-line documentation is available), at which point this can be re-worked.

PR #114, opened 2015-10-09
force numpy 1.9/1.8 in Travis build

olaurino:
# Release Note
This PR reverts the changes made by PR #89, as there have been some (yet unknown) incompatibilities with PR #101. In order to support the CIAO 4.8 release, we are giving priority to #101 and reverting #89.

We will need to look into the reason for the test errors triggered by #89.

Also, the current pyfits conda binaries list numpy 1.8* as dependency, making it incompatible with 1.9, so when testing pyfits I force numpy to 1.8 instead.


olaurino:
Closing this PR as we are making the change in PR #101

PR #115, opened 2015-10-09
Revert pr 89

olaurino:
#Release Note
Reverting changes made by PR #89 as they conflict with PR #101. Unfortunately the origin of this issue is still unknown, and should be investigated. In any case PR #89 does not impact CIAO, while we want PR #101 to be included in CIAO 4.8.


dtnguyen2:
I don't think the following is true anymore:

For UserStat: if response has bkg then user must add bkg to the model func
otherwise it can be omitted


because I made sure that the UserStat will be backwards compatible.
Otherwise, probably, all of the current user's stats functions do have bkg
data but the function does not require to have the bkg argument so without
an update, all current scripts would fail

PR #116, opened 2015-10-09
Bug #27 update pyfits deprecated interfaces

### LINKED ISSUE #27: pyfits/astropy backend is using deprecated functionality

olaurino:
This PR resuscitates PR #89 that was reverted by #115 for milestone 4.8rc1.

Resolve #27.

# Post-release note

This PR was reverted by #115 since it caused problems for #101 (in that it broke the detection of XSPEC table files loaded via `load_table_model`).

# Release Note

The `astropy.io.fits`/`pyfits` interface used deprecated functionality. The code was updated to use the replacement classes/methods when available, falling back to the original code if not. The interfaces that were changed were, when the new symbols are availble, to:

 - use `astropy.io.fits.BinTableHDU.from_columns` rather than `astropy.io.fits.new_table`

- use `astropy.io.fits.Header` rather than `astropy.io.CardList`

# Code changes

The code went through some minor clean up (PEP8 style fixes and then some minor re-factoring to reduce repeated code), as well as changing from using the `pyfits` module name to `fits` (since astropy is now the preferred package) before removing the deprecated functionality. Fortunately for the routines used here, the replacement routines/classes have the same API, so it's an easy change.

The [PyFITS changelog](http://www.stsci.edu/institute/software_hardware/pyfits/release) was consulted - particularly for version 3.3 - to check what functionality we are using that has been deprecated.

# Testing

The full smoke test was run (with the testing submodule installed and with the XSPEC module compiled against version 12.9.0b and with ds9 available) using astropy version 1.0.4. I have *not* tested against pyfits (in particular to see if it works with old versions).

I used the following two scripts to check that the deprecation warnings are gone (I wasn't sure how/if this could be done as part of the smoke test). These are based on examples given in #27.

a) writing out a FITS binary table

```
# does writing out a FITS file cause warnings?
import sherpa
from sherpa.astro import ui
print("*** Sherpa version: {}".format(sherpa.__version__))
ui.load_arrays(1, [1,2,3], [4,5,6])
ui.save_data(1, 'test_fits.dat', ascii=True, clobber=True)
print("*** created test_fits.dat ***")
ui.save_data(1, 'test_fits.fits', ascii=False, clobber=True)
print("*** created test_fits.fits ***")
```

Before this PR this complained about

```
WARNING: AstropyDeprecationWarning: The new_table function is deprecated and may be removed in a future version.
        Use :meth:`BinTableHDU.from_columns` for new BINARY tables or :meth:`TableHDU.from_columns` for new ASCII tables instead. [astropy.utils.decorators]
```

whereas I know get the following (it's dirty because the `setup.cfg` is changed to build XSPEC and I haven't hidden it from git):

```
*** Sherpa version: 4.7+530.g6dacc53.dirty
*** created test_fits.dat ***
*** created test_fits.fits ***
```

b) writing out a PHA file

Unfortunately the PHA files we have available in sherpa all seem to use the `LONGSTRN` convention, which means that writing them out triggers #46 - but with

```
import sherpa
from sherpa.astro import ui
print("*** Sherpa version: {}".format(sherpa.__version__))

ui.load_pha(1, './sherpa/astro/datastack/tests/data/3c273.pi')
ui.save_data(1, 'test.fits', ascii=False, clobber=True)
print("*** created test.fits")
```

we used to get the following warnings:

```
WARNING: AstropyDeprecationWarning: The CardList class has been deprecated; all its former functionality has been subsumed by the Header class, so CardList objects should not be directly created.  See the PyFITS 3.1.0 CHANGELOG for more details. [astropy.io.fits.card]
WARNING: AstropyDeprecationWarning: The append function is deprecated and may be removed in a future version.
        Use :meth:`Header.append` instead. [astropy.utils.decorators]
```

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -42,6 +42,8 @@ try:
 except ImportError:
     import pyfits as fits
 
+from numpy.compat import basestring
+
 import os
 from itertools import izip
 from sherpa.utils.err import IOErr
@@ -269,12 +271,10 @@ def _try_vec_or_key(hdu, name, size, dtype=SherpaFloat, fix_type=False):
 
 def read_table_blocks(arg, make_copy=False):
 
-    filename = ''
-    hdus = None
-    if type(arg) is fits.HDUList:
+    if isinstance(arg, fits.HDUList):
         filename = arg[0]._file.name
         hdus = arg
-    elif type(arg) in (str, unicode, numpy.str_) and is_binary_file(arg):
+    elif isinstance(arg, basestring) and is_binary_file(arg):
         filename = arg
         hdus = fits.open(arg)
     else:
@@ -323,11 +323,11 @@ def _get_file_contents(arg, exptype="PrimaryHDU", nobinary=False):
     file is a binary file (via the is_binary_file routine).
     """
 
-    if type(arg) == str and (not nobinary or is_binary_file(arg)):
+    if isinstance(arg, basestring) and (not nobinary or is_binary_file(arg)):
         tbl = fits.open(arg)
         filename = arg
-    elif type(arg) is fits.HDUList and len(arg) > 0 and \
-            arg[0].__class__ is fits.PrimaryHDU:
+    elif isinstance(arg, fits.HDUList) and len(arg) > 0 and \
+            isinstance(arg[0], fits.PrimaryHDU):
         tbl = arg
         filename = tbl[0]._file.name
     else:
@@ -347,14 +347,14 @@ def _find_binary_table(tbl, filename, blockname=None):
 
     if blockname is None:
         for hdu in tbl:
-            if hdu.__class__ is fits.BinTableHDU:
+            if isinstance(hdu, fits.BinTableHDU):
                 return hdu
 
     else:
         blockname = str(blockname).strip().lower()
         for hdu in tbl:
             if hdu.name.lower() == blockname or \
-                    hdu.__class__ is fits.BinTableHDU:
+                    isinstance(hdu, fits.BinTableHDU):
                 return hdu
 
     raise IOErr('badext', filename)
@@ -391,7 +391,7 @@ def get_column_data(*args):
 
     cols = []
     for arg in args:
-        if not (type(arg) in (numpy.ndarray, list, tuple) or arg is None):
+        if arg is not None and not isinstance(arg, (numpy.ndarray, list, tuple)):
             raise IOErr('badarray', arg)
         if arg is not None:
             vals = numpy.asarray(arg)
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -1,4 +1,4 @@
-# 
+#
 #  Copyright (C) 2011, 2015  Smithsonian Astrophysical Observatory
 #
 #
@@ -29,7 +29,7 @@ from itertools import izip
 from sherpa.utils.err import IOErr
 from sherpa.utils import SherpaInt, SherpaUInt, SherpaFloat, is_binary_file
 from sherpa.io import get_ascii_data, write_arrays
-from sherpa.astro.io.meta import *
+from sherpa.astro.io.meta import Meta
 
 import logging
 warning = logging.getLogger(__name__).warning
@@ -47,6 +47,7 @@ __all__ = ('get_table_data', 'get_image_data', 'get_arf_data', 'get_rmf_data',
            'get_pha_data', 'set_table_data', 'set_image_data', 'set_pha_data',
            'get_column_data', 'get_ascii_data')
 
+# Should we drop support for earlier versions?
 try:
     # pyfits-1.3 support
     _VLF = fits.NP_pyfits._VLF
@@ -74,7 +75,6 @@ def _has_hdu(hdulist, id):
 
 
 def _has_key(hdu, name):
-    #return hdu.header.has_key(name)
     return name in hdu.header
 
 
@@ -121,7 +121,7 @@ def _get_wcs_key(hdu, key0, key1, fix_type=False, dtype=SherpaFloat):
 
     if _has_key(hdu, key0) and _has_key(hdu, key1):
         return numpy.array([_try_key(hdu, key0, fix_type, dtype),
-                            _try_key(hdu, key1, fix_type, dtype)],dtype)
+                            _try_key(hdu, key1, fix_type, dtype)], dtype)
     return ()
 
 
@@ -135,6 +135,7 @@ def _add_keyword(hdrlist, name, val):
 
     else:
         hdrlist.append(fits.Card(name, val))
+
 
 def _try_col(hdu, name, dtype=SherpaFloat, fix_type=False):
     if name not in hdu.columns.names:
@@ -152,6 +153,7 @@ def _try_col(hdu, name, dtype=SherpaFloat, fix_type=False):
 
     return col
 
+
 def _try_tbl_col(hdu, name, dtype=SherpaFloat, fix_type=False):
     if name not in hdu.columns.names:
         return (None,)
@@ -168,9 +170,10 @@ def _try_tbl_col(hdu, name, dtype=SherpaFloat, fix_type=False):
 
     return numpy.column_stack(col)
 
+
 def _try_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
     if name not in hdu.columns.names:
-        return numpy.array([None]*size)
+        return numpy.array([None] * size)
 
     col = hdu.data.field(name)
 
@@ -183,9 +186,10 @@ def _try_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
         col = col.astype(dtype)
 
     if col is None:
-        return numpy.array([None]*size)
+        return numpy.array([None] * size)
 
     return col
+
 
 def _require_col(hdu, name, dtype=SherpaFloat, fix_type=False):
     col = _try_col(hdu, name, dtype, fix_type)
@@ -193,11 +197,13 @@ def _require_col(hdu, name, dtype=SherpaFloat, fix_type=False):
         raise IOErr('reqcol', name, hdu._file.name)
     return col
 
+
 def _require_tbl_col(hdu, name, dtype=SherpaFloat, fix_type=False):
     col = _try_tbl_col(hdu, name, dtype, fix_type)
     if len(col) > 0 and col[0] is None:
         raise IOErr('reqcol', name, hdu._file.name)
     return col
+
 
 def _require_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
     col = _try_vec(hdu, name, size, dtype, fix_type)
@@ -205,21 +211,22 @@ def _require_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
         raise IOErr('reqcol', name, hdu._file.name)
     return col
 
+
 def _try_col_or_key(hdu, name, dtype=SherpaFloat, fix_type=False):
     col = _try_col(hdu, name, dtype, fix_type)
     if col is not None:
         return col
     return _try_key(hdu, name, fix_type, dtype)
 
+
 def _try_vec_or_key(hdu, name, size, dtype=SherpaFloat, fix_type=False):
     col = _try_col(hdu, name, dtype, fix_type)
     if col is not None:
         return col
-    return numpy.array([_try_key(hdu, name, fix_type, dtype)]*size)
+    return numpy.array([_try_key(hdu, name, fix_type, dtype)] * size)
 
 
-
-## Read Functions ##
+# Read Functions
 
 def read_table_blocks(arg, make_copy=False):
 
@@ -232,13 +239,14 @@ def read_table_blocks(arg, make_copy=False):
         filename = arg
         hdus = fits.open(arg)
     else:
-        raise IOErr('badfile', arg, "a binary FITS table or a PyFITS.BinTableHDU list")
+        raise IOErr('badfile', arg,
+                    "a binary FITS table or a PyFITS.BinTableHDU list")
 
     cols = {}
     hdr = {}
 
     for ii, hdu in enumerate(hdus):
-        blockidx = ii+1
+        blockidx = ii + 1
 
         hdr[blockidx] = {}
         header = hdu.header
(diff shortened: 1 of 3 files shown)


olaurino:
Rebasing. Original tip was at 3cb1cc7

olaurino:
5e89203 should fix the regression introduced by commit 5e5eabff.

olaurino:
Lowering priority as the blocking issue was resolved. I'll keep reviewing the PR so that hopefully it can soon be merged.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -163,62 +206,67 @@ def _try_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
...
-
-## Read Functions ##
+# Read Functions
 
 def read_table_blocks(arg, make_copy=False):
 
     filename = ''
     hdus = None
-    if type(arg) is pyfits.HDUList:
+    if type(arg) is fits.HDUList:
```
olaurino: We should use `isinstance` rather than comparing the class. Even better, we should rely on duck typing, although this is probably more complicated at this stage.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -163,62 +206,67 @@ def _try_vec(hdu, name, size=2, dtype=SherpaFloat, fix_type=False):
...
         filename = arg[0]._file.name
         hdus = arg
     elif type(arg) in (str, unicode, numpy.str_) and is_binary_file(arg):
```
olaurino: There are more robust ways of checking if a variable is a string. It is probably not worth trying to employ patterns that are both Python 2 and Python 3- compliant at this stage, as this would most likely require the addition of more OTS packages (e.g. `six` or `future`) we still cannot add.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
+    file is a binary file (via the is_binary_file routine).
+    """
 
-
-def get_header_data( arg, blockname=None, hdrkeys=None ):
-
-    filename = ''
-    if type(arg) == str and is_binary_file(arg):
-        tbl = pyfits.open(arg)
+    if type(arg) == str and (not nobinary or is_binary_file(arg)):
```
olaurino: ditto for the string type check.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
     return filename, cols, hdr
 
 
+# @DougBurke thinks that nobinary may just be a sign of a
```
olaurino: So, what to do here? My inclination would be to leave things as they are. Maybe add a TODO, maybe not.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
+# missed copy-and-paste edit, as it's not at all obvious why
+# it isn't included there (alternatively, it may make sense to
+# just remove the binary check entirely and let the file I/O
+# fall over if it can't read the file).
+#
+# @DougBurke thinks that the exptype argument also seems to be
```
olaurino: again, we should probably leave it as is to reduce the risk of regressions.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
-    filename = ''
-    if type(arg) == str and is_binary_file(arg):
-        tbl = pyfits.open(arg)
+    if type(arg) == str and (not nobinary or is_binary_file(arg)):
+        tbl = fits.open(arg)
         filename = arg
-    elif ( (type(arg) is pyfits.HDUList) and
-           (len(arg) > 0) and
-           (arg[0].__class__ is pyfits.PrimaryHDU) ):
+    elif type(arg) is fits.HDUList and len(arg) > 0 and \
```
olaurino: use isinstance instead.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
+            arg[0].__class__ is fits.PrimaryHDU:
```
olaurino: ditto.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
+    if blockname is None:
         for hdu in tbl:
-            if blockname is None:
-                if hdu.__class__ is pyfits.BinTableHDU:
-                    break
-                else:
-                    continue
-            elif (hdu.name.lower() == str(blockname).strip().lower()):
-                break
+            if hdu.__class__ is fits.BinTableHDU:
```
olaurino: ditto.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -237,45 +285,76 @@ def read_table_blocks(arg, make_copy=False):
...
+                return hdu
 
-        else:
-            raise IOErr('badext', filename)
+    else:
+        blockname = str(blockname).strip().lower()
+        for hdu in tbl:
+            if hdu.name.lower() == blockname or \
+                    hdu.__class__ is fits.BinTableHDU:
```
olaurino: ditto

olaurino:
Rebasing. Original tip was at f42bceb

sherpa/astro/io/pyfits_backend.py
```diff
@@ -65,6 +90,23 @@
...
+# if available. Note that this should only be done for binary
+# tables, as fits.TableHDU.from_columns should be used for
+# ASCII tables.
+#
+if hasattr(fits.BinTableHDU, "from_columns"):
+    _new_table = fits.BinTableHDU.from_columns
+else:
+    _new_table = fits.new_table
+
+# fits.CardList is deprecated
```
DougBurke: This (the definition of `_new_header`) isn't needed as it is already at line 86.

sherpa/astro/io/pyfits_backend.py
DougBurke: As discussed (off GitHub), there may be some benefit to clean up here, but not in this PR.  The comment probably needs to be clearer that the issues here are:

* is the check for a binary file even useful here; perhaps the code should just try and open the file and we deal with the error (it's "more pythonic")

* if we do want to include a binary-file check, then all the callers of `_get_file_contents` should be reviewed to see if they should include the binary-file check

* the `exptype` argument lets the error message be changed; however, does this make sence since the code that does call this function with `exptype="BinTableHDU"` aren't guaranteed that there is a binary table in the file (very likely, but it is possible to write a FITS file with only a `PrimaryHDU` or have an `ImageHDU` [or whatever it's called] falling it [*]).

[*] I haven't checked the FITS standard to see if either of these are valid, but that doesn't mean that they can't be created. It could be that `astropy` will error out if it reads such a file, but I haven't checked this either.

sherpa/astro/io/tests/test_io.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @skipIf(not has_package_from_list("sherpa.astro.io.pyfits_backend"), "this is a pyfits/astropy test")
+    def test_warnings_are_gone_pha(self):
+        pha = self.make_path("threads", "pha_intro", "3c273.pi")
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            with NamedTemporaryFile() as f:
+                ui.load_pha(pha)
+                ui.save_data(1, f.name, ascii=False, clobber=True)
+            assert len(w) == 0
```
DougBurke: Should there be a new-line at the end of this file?

sherpa/astro/io/tests/test_io.py
```diff
@@ -0,0 +1,59 @@
...
+from unittest import skipIf
+from tempfile import NamedTemporaryFile
+import warnings
+
+
+class test_89_issues(SherpaTestCase):
+    def setUp(self):
+        ui.clean()
+
+    @skipIf(not has_package_from_list("sherpa.astro.xspec"), "xspec required")
```
DougBurke: Do we gain anything from this test (i.e. `test_mod_fits`) given that the XSPEC tests have what looks to be a more-extensive test `text_xspec_tablemodel` https://github.com/sherpa/sherpa/blob/ca2db23857a4da09ca7fb3adb7b94b2871da67f3/sherpa/astro/xspec/tests/test_xspec.py#L305

I think the conditions for running both tests are the same (both fits and xspec) and the XSPEC version includes these tests - but I'm in a hurry and may have missed something.

If we do keep this test then don't we need to add a `@skipIf(test_data_missing)` decorator to make sure that the RCS.mod file is available?

olaurino: When I fix a bug I want to have a failing unit test that exercises the bug and fails before I add a fix. The xspec integration tests are too complicated and take too long to run, so I added a new simple one that exercises the specific issue.

However it is true that we need one more skipif.

sherpa/astro/io/pyfits_backend.py
olaurino: I'll add a TODO with your questions.

sherpa/astro/io/pyfits_backend.py
olaurino: Good catch. This was a side effect of the rebase/conflict resolution.

sherpa/astro/io/tests/test_io.py
olaurino: Yes.

olaurino:
071869c addresses Doug's comments. As soon as Travis builds are green I'll merge it.

PR #117, opened 2015-10-14
Release/4.8.0

olaurino:
# Release Note
The Python help strings are no longer generated from the ahelp files but from the standard Python docstrings embedded in the code.

In accordance with the new CIAO standards, the field `sherpa.__versionstr__` was removed and the field `sherpa.__version__` now contains a Python string compliant with the PEP440 standard. The `ipython` startup scripts were updated to compensate for this change and to make sure that the changes are backward compatible, i.e. that running a different (maybe older) version of CIAO will result in such startup scripts to replaced as needed.

A method introduced in PR #94 was failing to catch a edge case triggered by the srcextent tool in CIAO 4.8b2. The method has now been updated.

# Description
This PR tracks the post-code-freeze changes to support the CIAO 4.8 release. At this point only urgent fixes in response to testing can be considered and accepted. Any changes to the Release Notes should go in the section above.

### Version change
As agreed with the other CIAO teams, the sherpa.__versionstr__ field was removed, and the version string is now held by the sherpa.__version__ field. For Sherpa this field is automatically determined from the git status. When running "python setup.py version" (of course, from the release branch) you should get "4.8.0". The same output should be given by "python -c 'import sherpa; print sherpa.__version__".

More work will be needed to make sure that the change does not have side effects, and that the ancillary scripts are updated accordingly. This is a CIAO-wide change, not just a Sherpa one.

### Release Notes
I added a RELEASE_NOTES file with the Release Notes for this release. As I had to add it to MANIFEST.in, I also added CITATION and CONTRIBUTING.md. The RELEASE_NOTES file will be updated before the next release by adding new release notes on top of it.

### Test Changes
Running the tests on CIAOX with the test data folder uncovered some issues:
  - a test fits file contained a column of a datatype unsupported by pycrates/dm. The file was replaced with a different file with the same numerical content but different data types.
  - some tests did not change the logging level and resulted in spurious messages when running them. I added some setUp/tearDown logic to compensate.

### get_valid_args
Commit 5ba2a7a introduces a change in response to a regression in the srcextent script in CIAO 4.8b2.


olaurino:
The push build on Travis fails because in this branch Travis does not have the conda path fix. However, the PR Travis build (the result of the merge of this branch with master) succeeds.

PR #119, opened 2015-10-15
Startup scripts update

olaurino:
# Description
Update startup scripts.

In particular, we need to compensate for the disappearance of `__versionstr__` and the fact that `__version__` is now a string.

From now on, since Sherpa will have its own versioning scheme independent of CIAO, we will need to state what versions of the CIAO packages are supported by a Sherpa version.
In this PR, we state that ChIPS >=4.7.0 and Crates >=4.7.0 are required. These values are conservative, and it is possible that Sherpa 4.8.0 would work just fine with earlier versions.

The `sherpa` wrapper script now uses the `ciao_app_setup.sh` script that will be available in CIAO for setting up user startup scripts, or update them in case of a version mismatch. This obsoletes the `sherpa_py` script, which will be removed.

More clean-up will be needed as there are obsolete files and statements to be removed. Also, I would like to move the CIAO-specific files to a specific folder rather than having them mixed within the main package directory.

### DIFF
--- a/sherpa/00-sherpa_startup.py
+++ b/sherpa/00-sherpa_startup.py
@@ -51,7 +51,7 @@ import ipython_cxc
 import ahelp_interface
 
 def _initialize_sherpa_app_():
-    # These are the minimum CIAO version supported by this version of Sherpa.
+    # These are the versions Sherpa was tested and packaged with.
     min_chips_version = "4.8.0"
     min_crates_version = "4.8.0"
 
--- a/sherpa/00-sherpa_startup.py
+++ b/sherpa/00-sherpa_startup.py
@@ -1,5 +1,5 @@
 # -*- Mode: Shell-Script -*-  Not really, but shows comments correctly
-# 
+#
 #  Copyright (C) 2012, 2015  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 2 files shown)


DougBurke:
@olaurino you say that 'but we know for sure that Sherpa 4.8.0 is
incompatible with Crates 4.8.x.'. I hope that is just a typo in the crates
numbering. Also, do you mean chips 4.7.0, >= 4.7.0, or 4.8.0 in the
preceeding sentence?

olaurino:
@DougBurke it was indeed a typo, but I reviewed the PR that compensated changes in the pycrates API and I realized that our change was backward-compatible, so even with the typo corrected, that statement is actually inaccurate and Sherpa 4.8.0 should indeed be compatible with pycrates 4.7.

What troubles me is that we don't have proper tests that prove it. While we cannot use CIAOX to test against the current development state of CIAO, maybe we can create Travis CI builds that exercise the released versions of CIAO?

olaurino:
For the record, I am referring to PR #32 and in particular to commit 683f2d4

olaurino:
As for chips, I meant >4.7.0. I'll fix the description and change the minimum crates version required to 4.7.0

olaurino:
I have to say the original check for older versions of pycrates and chips puzzled me. The fact that Sherpa is now standalone may somehow make these checks more relevant, but hardly so: the startup scripts are only run in CIAO, where they may only be relevant if somebody is fiddling with the tools, mixing and matching different versions of the tools, which keeps me wondering about the relevance of such checks.

On the other hand, we are not currently documenting the dependency requirements, which is something we should add anyway (maybe to the README?).

My question to @anetasie @DougBurke and @wmclaugh is: should we keep (and in that case fix) these checks? Should they just warn the user [fiddling with CIAO components versions] that the versions of chips and pycrates are not the ones that Sherpa 4.8.0 was supposed to be packaged with? How common should this case be for us to care?

olaurino:
I rebased the branch as I was getting a minor conflict after the recent merge, and I also fixed the creates version in the wrapper script while rephrasing the warning message:
https://github.com/sherpa/sherpa/commit/d5053bdda16bf4e2aa5f3e681b9b2b1ee76293b0#diff-91030e9cfe6144aa55e5b53c64431a9fR83

PR #120, opened 2015-10-16
update version in README

olaurino:
Update version in README


olaurino:
@DougBurke do you think I can merge this? In particular, is the statement about XSpec and wcslib still relevant? I believe it is, but just double-checking.

Also, the README refers to Xspec 12.8.2... are there any changes needed there as well?

PR #121, opened 2015-10-20
Clean up ciao files

olaurino:
# Description
This PR removes ciao-related files and some ciao-related code that was also obsolete.

The files removed by this PR will be moved to a different location, either in the CIAO source code repository or in a different git repository, and appropriately edited according to the new CIAO setup.


DougBurke:
Looks good to me.

Should this be against master or the release/4.8.0 branch?

olaurino:
master. This is intended as part of a post-ciao-release clean-up.

DougBurke:
Do we need a milestone for this work (i.e. is there more planned), or can it just be 4.8?

olaurino:
@DougBurke sorry for not replying earlier. Milestones are important so that we track the issues we decide to address, as it helps tracking progress. We may add milestones to PRs in some cases, but I don't think that's too important, given the experience with the past development cycle.

So no, I don't think we need a milestone for this.

olaurino:
Rebasing, original tip was cd2a754

PR #122, opened 2015-10-20
Remove the check for non-finite values from XSPEC models

DougBurke:
The XSPEC interface changes in Sherpa 4.8.0 development included a
change where the sherpa.astro.xspec.XSModel class would check that the
model had created only finite values, raising an error if this
constraint did not hold. This was suggested as a good thing to do
during a meeting between the Sherpa developers and Keith Arnaud.

This change caused problems with some of the external regression tests
in CIAO, which use a problematic (i.e.  does not confirm to OGIP
standards) RMF file with an energy bin that includes 0 keV. This
causes the XSpowerlaw model to create a non-finite value in this bin
when the photon index is positive.

This commit reverts this change (by commenting out the code that added
the check), and adds in explicit checks in the test suite.  Hopefully
the check for non-finite values can be added back in once the handling
of this case has been reviewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_install:
   - export PATH=/home/travis/miniconda/bin:$PATH
   - conda update --yes conda
   - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy=1.9 matplotlib
-  - conda config --add channels https://conda.binstar.org/sherpa
+  - conda config --add channels https://conda.anaconda.org/sherpa
   - pip install -r test_requirements.txt
   - if [ ${TEST} == package ];
      then pip install ./sherpa-test-data;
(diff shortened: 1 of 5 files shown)


olaurino:
@DougBurke I tweaked some documentation, namely the README and the RELEASE_NOTES files.

The change in the README is to mention anaconda.org rather than binstar.org. Travis had issues yesterday since binstar.org was not reachable. They fixed things today, but in any case binstar.org has been deprecated for anaconda.org. As I changed the values in the Travis configuration, I thought for consistency I should make that change too.

I also removed the part where Xspec models check for finite values in the RELEASE NOTES.

If you could quickly give me your thumbs up I can merge this PR. Thanks!

DougBurke:
:+1: (I'm going to assume the tests pass ;-)

olaurino:
They do. Thanks.

I am going to issue a new PR to remove the Xspec build from the "allowed to fail" set, since the new setup seems reliable so far, and it it's not we'll know by making this change.

PR #126, opened 2015-10-26
fix typo in startup script

olaurino:
# Description
After testing the new startup scripts in CIAOD I realized there was a typo in one of the updates startup scripts. This PR fixes the typo.

### DIFF
--- a/sherpa/00-sherpa_startup.py
+++ b/sherpa/00-sherpa_startup.py
@@ -78,7 +78,7 @@ def _initialize_sherpa_app_():
 
     versions = zip(min_crates_version.split('.'), crates_version.split('.'))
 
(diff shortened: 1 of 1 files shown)


olaurino:
@DougBurke do you think I can merge this?

DougBurke:
So, just to check: this looks like it's fixing a typo you added recently (and we didn't pick up in our code review), rather than a bug in the code that's been there a long time?

Also, we don't have anything in the Travis-CI tests that checks that this code works. The testing for this happens off-to-the side since it's part of CIAO rather than the standalone release?

olaurino:
Yes, and Yes.

DougBurke:
Looks fine for merging.

PR #127, opened 2015-11-19
fix miniconda path

olaurino:
It looks like the default value for the miniconda installation prefix has changed with the latest versions, and both the old and new Travis builds would fail for Sherpa.

This PR explicitly sets the prefix for the miniconda installation. It also uses an environment variable instead of having the same string repeated over and over again in the configuration file.

The Travis builds are currently running, but I already made sure that they get past the point where they were failing.


DougBurke:
Looks good to me. I've labelled it as ps:merge.

Issue #50, opened 2015-06-22
group.so and stk.so are installed outside of the sherpa namespace


taldcroft:
After doing a `python setup.py install` I see the following:
```
$ ls /Users/aldcroft/anaconda/envs/sherpa/lib/python2.7/site-packages/*.so
/Users/aldcroft/anaconda/envs/sherpa/lib/python2.7/site-packages/group.so*
/Users/aldcroft/anaconda/envs/sherpa/lib/python2.7/site-packages/stk.so*
```
Installing modules outside of the package namespace is not a  typical behavior.

olaurino:
This is a known issue and derives directly from CIAO, where these modules are built outside of Sherpa and installed in the global namespace for all the CIAO components to use.

Sherpa should be a good Python citizen and should avoid installing these modules in the global namespace, so I will take this as a RFE on both the build process (avoid installing group and stk in the global namespace) and the code (Sherpa should try to import these modules from the `sherpa` namespace and fall back to the global namespace). Not sure what the priority for this change will be, though.

In my tests, `pip` correctly identifies those modules as part of the Sherpa installation and correctly removes them during a `pip uninstall sherpa`: in my experience this is not always true, and I don't think I could find any patterns to explain why sometimes that does not happen.

cdeil:
It would be great if this could be fixed for the next Sherpa release (see https://github.com/sherpa/sherpa/issues/75#issuecomment-129862316)

DougBurke:
For reference (as part of the "we need to revamp the sherpa build project") I need to read https://discuss.python.org/t/building-extension-modules-the-2020-way/5950/33 and related

hamogu:
I noted that - at least in the one case I'm currently looking at - I get
`error: [Errno 2] No such file or directory: '/Users/guenther/code/sherpa/build/lib/python3.11/site-packages/stk.so'` when I install as editable for development with `pip install -e .`, but not when I install normally `pip install .`

That's (a) good because I can work around out (it's a little more cumbersome to have to run the install every every change in a Python file, but possible) and (b) because it might give a hint (thought I don't quite know which one) what causes this.

DougBurke:
I wonder - and this is a completely random thought - whether we need

  pip install -e . --no-build-isolation

DougBurke:
A quick build later suggests that may not be much of a help ....

PR #132, opened 2015-12-04
131 extensions not built

### LINKED ISSUE #131: several extensions (psf, region,  wcs) are not built
The file system structure I work on are unique, and we have our own equivalent of pip. Installing sherpa, I discovered that the extensions defined in helpers/extensions/\_\_init\_\_.py although later added into distribution.ext_modules through config, are not built during the procedure of build_ext.

The current strategy in sherpa packaging, specifically, in the overrode cmd_build and subclassed build(), it tries to call sherpa_config.build_configure(), in which the self.distribution.ext_modules will get written in several extensions defined in helpers/extensions/\_\_init\_\_.py. But, since what's defined in ext_modules are processed in build_ext procedure, which in time precedes build procedure, the expanded part of ext_modules won't have the opportunity to get processed, hence, those extenions won't be built. This has ultimately resulted in sherpa.test() failure, with a ImportError on looking for _psf module while importing sherpa.utils.

In the effort of mine trying to get this to work for my purpose, I "rescheduled" this to happen in the very beginning of build_ext, and it worked.

olaurino:
Fix #131 

#Description
Some extensions (wcs, region, psf) were not always built as part of the sherpa build process because of some unnecessary checks for the `extern/built` file in `sherpa_config` prevented `python setup.py develop` to have those extensions declared. The issue was triggered in some edge cases with the `develop command` (i.e. if the file `extern/built` existed the extensions would not be declared and built). Since the extensions were not declared before the `build_ext` command was run, special setups may have not been able to build those extensions, as reported in #131.

This fix allows the extensions to always be declared by the `sherpa_config` command, even though the `extern` dependencies are not being built.

Also, the `build` command override is now redundant, so it was removed.

There was also an unrelated issue with the `stk` and `group` extensions, where sherpa would try and copy them during the `develop` phase even if they were disabled and so they could not be found. That is what commit 6b95ec6 is about. This seemed like a good PR to include this change as well.


chaobin:
Please allow me to get back on this on Monday, and thanks for the quick fix and update!

chaobin:
Olaurino, using the fix, I can see the configure step now happens before build_ext (thus it should be able to add those extensions into ext_modules to be built later). But I seemed to get an error "error: No such file or directory" after that (I didn't have this with the original code).

Also, in install.py, line 28 and line 30 will actually result in twice calls to build_configure() as the sherpa_config.run() internally calls self.build_configure(), is that necessary?

Back to the bigger picture, I filed this issue because I suspect that this is a problem to a standard "pip install" procedure regardless of the specifics of my environment. But if it is not, there is no need to cater for my needs since I have well grasped the changes needed to teach it to obey in my environment :-)

I appreciate the excellent effort in open sourcing this superb project, and your prompt response on the issue.

chaobin:
I tested in a standard environment using pip, the installation were able to configure the ext_modules and those extensions are later built. So the environment at my work should be an edge case. Thanks for looking into this again.

olaurino:
@chaobin, thanks for testing the patch. Sometimes edge cases may uncover issues we want to fix anyway. So please feel free to add more information to the original issue #131.

The install override, at this point, has become redundant as well, so I'll look into removing it. Thanks for the catch!

olaurino:
The install override has to stay because it addresses setups without setuptools (like CIAO). However, as suggested by @chaobin I removed the code that would result in the sherpa_config.build_configure method to run twice during an install.

DougBurke:
I'm happy for this to be merged.

DougBurke:
Does this change "fix" #50 at all?

olaurino:
It doesn't, unfortunately.

DougBurke:
:-( 

Not that  this changes whether this should be merged or not.

olaurino:
If you feel #50 should be fixed as a priority we can work on that. Group and stack library are not imported in so many different places, so a conditional import should not be too hard to add. Changing their locations when installed by standalone sherpa should not be hard either. I just never looked into that issue because #50 was never included as a priority.

olaurino:
Rebasing. Original tip was at 580e51e

PR #137, opened 2016-01-05
Update region library code and configure.ac

olaurino:
[release note]
Update region library to v4.8
[/]

# Description
This PR updates the region library OTS source code to version 4.8, released as part of CIAO 4.8.
However, with respect to the code taken from CIAO code base, the library contained in Sherpa is stripped of the `regtest` folder containing the tests, as they have not been [autoconfiscated](https://en.wikipedia.org/wiki/Autoconfiscation) yet.

No other OTS libraries were updated.


DougBurke:
I don't think this needs a detailed review, since it's "infrastructure". So I say merge it as the tests pass.

olaurino:
Rebasing. Original tip was 8b9bde9

PR #138, opened 2016-01-05
Bug/improve save all

olaurino:
# Note

This PR replaces #98.

[OL] I squashed the commit history into one single commit after rebasing. This seems to make sense as the whole change, however large, is rather atomic:
  * move serialization code to its own module (+ refactor and add docstrings)
  * add tests for the serialization code
  * include a file handler as a possible argument for save_all, as per tests

The one change that does not seem so atomic is the one to `test_astro.py`. It is not clear why this change is required in the first place, and it might imply that something is wrong with the skipping of `xspec` tests. I'll keep looking into it.

# Release Notes
[release note]
This commit fixes and enhances the output of the `save_all` command. 

User-visible changes:

- added a new argument to `save_all`: if `outfile` is `None` then the `outfh` argument is used 
  to define the output handle (the argument can be any file-like argument, such as a file
  handle like `sys.stdout` or the output of `open`, or a `StringIO` object)
- setting the `clobber` argument to `save_all` now means that the output file (the `outfile`
  argument, if not `None`) is deleted if it already exists; prior to this, the file would be
  appended to instead
- the source expression is now saved correctly for most cases (e.g. when not using
  `set_full_model`); this is bug #97 but also affects non-PHA data sets
- the background model expression was not always written out correctly when using PHA
  data sets
- quality and grouping arrays of PHA data sets are now stored as 16-byte integers rather
  than a floating-point value (this has no affect on the results, but matches the OGIP standard)
- fixed up saving the grouping and quality arrays of background PHA data sets (this would only
  be an issue if the background is being fit, rather than subtracted)
- basic data sets created with the `load_arrays` function are now written out by `save_all`
  as part of the script; this is intended for small datasets and may have problems with
  precision if used with floating-point arrays
- calls to `load_psf` are now correctly restored (they may not have been written out correctly
  if multiple data sets were loaded)
- user models are now written out to disk; this consists of two parts:
  - writing out the function that defines the model, which may or may not be possible (if
    not, a place-holder function is added to the output and a warning displayed).
  - the necessary calls to `load_user_model` and `add_user_pars` are now included in
    the output
- the Python code created by `save` all has undergone several minor changes:
  - it now explicitly imports the `sherpa.astro.ui` module, so that it can be run from the
    IPython prompt using the `%run <filename>` command, or directly as `python <filename>`
  - it uses the `create_model_component` function rather than `eval` to create model
    components (this is CXC bug 12146)
  - many optional arguments to functions are now given as `name=value` rather than
    being a positional argument, to make it clearer what the script is doing.
  - calls to `load_data` have been replaced by more-specific versions - e.g. `load_pha`
    and `load_image` - if appropriate
  - there have seen several minor syntactic clean ups to better follow the suggestions from PEP8 

When writing out code that defines a user-model, there is no attempt to make sure that
modules used by the function are available. These will need to be added, either directly
or imported, manually to the output.
[/]

Internal changes:

- adds a test suite; this requires a small change to the `test_astro.py` test script just to make
  sure that the default XSPEC settings are restored on test tear-down (to avoid the new tests
  from failing). The test suite covers some of the basic situations expected to be served by
  `save_all` - such as basic arrays, PHA datasets - but does not cover all functionality
- moves the code out to `sherpa/astro/ui/serialize.py` from `sherpa/astro/ui/utils.py`, as the
  latter is rather large
- code has been refactored to promote sharing and improve readability
- code has seen PEP 8 fixes and other simplifications, such as:
  - replace `x == True` and `y == False` by `x` and `not y`
  - remove extra brackets in conditionals
  - use `isinstance(x, basestring)` rather than `type(x) == str`, which has been a problem for
    other parts of the Sherpa UI when given a string stored in a numpy array (since these
    aren't treated as `str` but do match the `isinstance` check)

#  Notes

The test suite has only been tested on one machine. It is likely that it will need changing to account for running on different configurations (in particular 32 vs 64 bit or different OS values); in particular, the regression tests that ensure the textual output is correct may need updating to allow numerical differences or changed to a different approach.

**ADDED NOTE** the tests include regression tests, where the output is compared against an expected string. These strings are currently stored within `test_serialize.py` itself, but it may make sense to move them into the `sherpa-test-data` repository, once the patch has been accepted (and the approach vindicated by more testing, to make sure it's worth approach).

Not all functionality is tested - e.g. serializing sessions that use image data, or an iterative fit, or more corner cases in various options. This set of changes is large enough as is, and significantly improves the overall coverage of Sherpa (from around 50% to 55%) that I think it's worth looking at now. The coverage of `serialize.py` is 68% (when the equivalent before this PR was 0% for the code that is now in `serialize.py`).

A lot of the logic in `serialize.py` should really be moved into the classes, so that rather than this code being changed whenever a new statistic needs to be serialized (say), the object itself can be queried. I'm wondering about a `Serializable` class which could be imported from to provide a `serialize` method (names open to discussion), or if not a class, just make sure that each object we want to serialize provides a method. This is way outside the scope of this PR, but some of the refactoring done in this PR was done to highlight places where this would be useful.

# Example

With the following code

```
from sherpa.astro import ui
ui.load_arrays(1, [1,2,3], [5,6,7])
ui.set_source(ui.const1d.m1)
```

the difference in the output of `save_all` from CIAO 4.7 (`out.ciao47`, which is what `master` gives) and this PR (`out.pr98`) is:

```
% diff out.ciao47 out.pr98 
1a2
> from sherpa.astro.ui import *
5c6,9
< load_data(1,"")
---
> load_arrays(1,
>             [1, 2, 3],
>             [5, 6, 7],
>             Data1D)
7c11
< ######### Set Image Coordinates 
---
> ######### Set Image Coordinates
21c25
< notice_id(1,"1.0000:3.0000")
---
> notice_id(1, "1.0000:3.0000")
44c48
< eval("const1d.m1")
---
> create_model_component("const1d", "m1")
60c64
< set_full_model(1, const1d.m1)
---
> set_source(1, const1d.m1)
```

which shows several of the enhancements and bug fixes, namely:

- import of the `sherpa.astro.ui` module
- simple datasets created by `load_arrays` are now written out to file
- `create_model_component` is used rather than `eval`
- the source model is written out correctly (using `set_source` and not `set_full_model`)

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -22,12 +22,20 @@ import logging
 import os
 import os.path
 from sherpa.utils import SherpaTestCase, test_data_missing
-from sherpa.utils import has_package_from_list, has_fits_support
-import sherpa.astro.ui as ui
+# from sherpa.utils import has_package_from_list, has_fits_support
+from sherpa.utils import has_fits_support
+from sherpa.astro import ui
 from sherpa.astro.data import DataPHA
 
 logger = logging.getLogger('sherpa')
 
+try:
+    from sherpa.astro import xspec
+    has_xspec = True
+except ImportError:
+    has_xspec = False
+
+# has_xspec = has_package_from_list("sherpa.astro.xspec")
 
 class test_threads(SherpaTestCase):
 
@@ -45,10 +53,17 @@ class test_threads(SherpaTestCase):
         self.old_level = logger.getEffectiveLevel()
         logger.setLevel(logging.CRITICAL)
 
+        # Store XSPEC settings, if applicable
+        if has_xspec:
+            self.old_xspec = xspec.get_xsstate()
+
     def tearDown(self):
         os.chdir(self.startdir)
         ui._session.__dict__.update(self.old_state)
         logger.setLevel(self.old_level)
+
+        if has_xspec:
+            xspec.set_xsstate(self.old_xspec)
 
     def assign_model(self, name, obj):
         self.locals[name] = obj
@@ -62,7 +77,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_pha_intro(self):
@@ -121,7 +136,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_simultaneous(self):
@@ -138,7 +153,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_sourceandbg(self):
@@ -174,7 +189,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_pileup(self):
@@ -214,7 +229,7 @@ class test_threads(SherpaTestCase):
         # Since any other I/O backend would not understand DM syntax,
         # I make this test a no-op unless CRATES is used as the I/O
         # backend.  SMD 05/23/13
-        if (True == self.is_crates_io):
+        if self.is_crates_io:
             self.run_thread('radpro_dm')
             self.assertEqualWithinTol(ui.get_fit_results().statval, 217.450, 1e-4)
             self.assertEqualWithinTol(ui.get_fit_results().rstat, 6.21287, 1e-4)
@@ -301,7 +316,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_spectrum(self):
@@ -330,7 +345,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_xmm(self):
@@ -358,7 +373,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_proj(self):
@@ -433,7 +448,7 @@ class test_threads(SherpaTestCase):
 
     @unittest.skipIf(not has_fits_support(),
                      'need pycrates, pyfits or astropy.io.fits')
-    @unittest.skipIf(not has_package_from_list('sherpa.astro.xspec'),
+    @unittest.skipIf(not has_xspec,
                      "required sherpa.astro.xspec module missing")
     @unittest.skipIf(test_data_missing(), "required test data missing")
     def test_proj_bubble(self):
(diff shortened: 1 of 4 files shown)


sherpa/astro/tests/test_astro.py
```diff
@@ -22,12 +22,20 @@
...
 
 logger = logging.getLogger('sherpa')
 
+try:
+    from sherpa.astro import xspec
+    has_xspec = True
+except ImportError:
+    has_xspec = False
+
+# has_xspec = has_package_from_list("sherpa.astro.xspec")
```
olaurino: Was there something wrong with "has_package_from_list"?

sherpa/astro/ui/utils.py
```diff
@@ -12364,7 +12367,7 @@ def calc_kcorr(self, z, obslo, obshi, restlo=None, resthi=None,
     # Session Text Save Function
     ###########################################################################
 
-    def save_all(self, outfile=None, clobber=False):
+    def save_all(self, outfile=None, clobber=False, outfh=None):
```
olaurino: tInstead of having to introduce a new argument and conditional checks on arguments combinations (e.g. if filename is None then use the file handler, if it is not None, etc.), couldn't we just overload the outfile argument and make it hold either None (stdout), or a string (filename), or an object (file handler)?

sherpa/astro/ui/utils.py
```diff
@@ -12401,13 +12409,19 @@ def save_all(self, outfile=None, clobber=False):
...
         -----
 
-        Items which are not saved include:
+        This command will create a series of commands that restores
+        the current Sherpa set up. It does not save the set of commands
+        used. Not all Sherpa settings are saved. Items not fully restored
+        include:
 
-        - user models
+        - data created by calls to `load_arrays`, or changed from the
```
olaurino: the docstring to save_all says that data created with calls to load_arrays is not saved, but that seems to be actually be the case, both by reading the code and by performing a simple test. Am I missing something?

sherpa/astro/ui/utils.py
```diff
@@ -12401,13 +12409,19 @@ def save_all(self, outfile=None, clobber=False):
...
+          version on disk - e.g. by calls to `sherpa.astro.ui.set_counts`.
 
         - any optional keywords to comands such as `load_data`
           or `load_pha`
 
+        - user models may not be restored correctly
+
         - only a subset of Sherpa commands are saved.
```
olaurino: This is rather vague, for a section that should indeed describe what is missing. This is not an issue of the PR per se, as this part is unchanged from the current code, but I thought it was worth mentioning, as maybe it could be fixed at this round.

olaurino:
@DougBurke not sure you get the notifications, as I recreated your PR, so just in case, here is a notification that I added some comments :)

olaurino:
A note by @DougBurke in the code mentions that the logic in `serialize.py` might be moved into the classes/modules being serialized. I certainly agree with this, but the question is whether we should do it at this round or not. If we did, we should first come up with more tests, and then start refactoring the code with the tests as a safety net against regressions.

olaurino:
I suggest we merge this as is. We need to add more tests, and we may want to refactor the logic in an object oriented fashion, but unless @DougBurke and @anetasie have a different opinion, this is hardly a priority.

olaurino:
Sorry, the previous comment was premature: I am working on some commits to improve the PR as suggested in the above comments, but I have not published them yet. On the other hand, the general meaning of the comment above regarding more tests and refactoring stands.

DougBurke:
I haven't reviewed anything of this yet. One of the reasons the code was moved to a separate module was because `save_all` and `save_state` (I think that was its name) were very similar and I wanted to combine them where possible. However, we've since removed `save_state` and accepted a very-cut-down version of these changes. I wonder if, rather than working through all the changes needed to get this to merge cleanly, it would be better to:

* take the tests (removing or commenting out the string buffer versions, maybe this means converting some which just used a string buffer to write out a file instead) and either note those that fail as "expected" or fix them so they pass with the current save_all.

* then address each test that is either failing or had to be edited

The idea to make it a bit-more obviously a test driven process than the original development.

sherpa/astro/tests/test_astro.py
DougBurke: I can't remember now; I think there was some reason (given I have a commented out `has_package_from_list` line), but I don't know if it was a sensible or correct reason.

sherpa/astro/ui/utils.py
DougBurke: I guess you'd do it as "check for `None`, then check for a file handle, otherwise assume a string (I'm a little wary of checking for strings, since it has bitten us in the past when we've not done it correctly - e.g. np.string_).

Overloading outfile is in many clearer for the code, and is something we do a lot of, so it wouldn't be a terrible choice. I think I wanted to separate out the two here just to be a bit cleaner/clearer in the interface, but would be happy either for

* the string buffer support to be dropped for now

* your suggestion be used (i.e. drop the `outfh` argument, which isn't a very useful name)

sherpa/astro/ui/utils.py
DougBurke: Sorry, I'm unclear on what you're asking here. Does save_all still not save data created with a load_arrays call?

sherpa/astro/ui/utils.py
DougBurke: The code I added to try and restore user models is not strongly tested [*], and I think it's very hard,if not impossible, to guarantee that it can be restored in all cases.

[*] or I am hallucinating that I added such code

olaurino: I meant "only a subset of Sherpa commands are saved" is a little vague, and it would be good to list what is not saved. Again, not an issue of this PR, but something we may have wanted to address here.

sherpa/astro/ui/utils.py
olaurino: According to my tests load_arrays is indeed recorded by save_all, which is why I was wondering why the docstring says that data created using load_arrays is not restored. Maybe I am missing something.

sherpa/astro/ui/utils.py
DougBurke: Ah. I was focusing on the user-models part! Unfortunately, listing what is not saved requires reviewing the whole code base to see what has been missed!

sherpa/astro/ui/utils.py
DougBurke: It's likely a failure-to-update-the-docstring condition.

olaurino:
Rebasing. Original tip was at 3dc8a25

olaurino:
I added some commits addressing the comments I posted earlier. @DougBurke if you wanted to double-check them and give me your thumbs up, I'd appreciate it.

olaurino:
> The idea to make it a bit-more obviously a test driven process than the original development.

I am all for it.

At this point I don't think we should go back to a pre-serialize.py state. Given a priority for this (and other) task(s), I would rather use it as a stepping stone to make save_all more sustainable and more robust.

sherpa/astro/ui/serialize.py
```diff
@@ -0,0 +1,1183 @@
...
+            _output(cmd, fh)
+        else:
+            # just in case
+            msg = "Set notice range of id={} bkg_id={} to {}".format(cmd_id,
+                                                                     fvals,
+                                                                     bkg_id)
+            _output('print("{}")'.format(msg), fh)
+
+
+def _save_data(state, funcs, fh=None):
```
DougBurke: With the removal of `save_state` (or whatever it was called), we should no-longer need the `funcs` argument. The values set in `save_all` (line 1169 below) should really be used in this routine, rather than use funcs as a proxy.

DougBurke: This is fixed in the commit I just added.

olaurino:
@DougBurke do you think we can merge this?

DougBurke:
I'm okay with this being merged.

PR #140, opened 2016-01-07
enable ds9 tests

olaurino:
This PR enables tests depending on DS9 to run on Travis CI. Also, tests are not silently skipped on headless systems (or wherever the `DISPLAY` env variable is unset).

PR #141, opened 2016-01-07
Tests cleanup

olaurino:
This PR simplifies the tests-skipping logic and makes it more robust to changes.

It is based on #116 for simplicity, as I was annoyed by the complexity of the process while adding tests for that PR. Since it also changes the way some tests added by #116 are decorated for being skipped, I'll just leave it like this.

This is likely to conflict with existing PRs, in particular #140.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -68,15 +68,6 @@ class test_ui(SherpaTestCase):
         ui.load_table(1, self.fits, 4, ('R', "SUR_BRI", 'SUR_BRI_ERR'),
                       ui.Data1DInt)
 
-    def test_load_table_fits(self):
-        # QUS: why is this not in the sherpa-test-data repository?
-        this_dir = os.path.dirname(os.path.abspath(__file__))
-        ui.load_table(1, os.path.join(this_dir, 'data',
-                                      'two_column_x_y.fits.gz'))
-        data = ui.get_data(1)
-        self.assertEqualWithinTol(data.x, [1, 2, 3])
-        self.assertEqualWithinTol(data.y, [4, 5, 6])
-
     # Test table model
     def test_table_model_ascii_table(self):
         ui.load_table_model('tbl', self.singledat)
@@ -522,6 +513,21 @@ class test_save_pha(SherpaTestCase):
             oldval = getattr(self._pha, key)
             self.assertEqual(oldval, newval, msg=key)
         """
+
+
+@requires_fits
+class test_basic_io(SherpaTestCase):
+    def setUp(self):
+        ui.clean()
+
+    def test_load_table_fits(self):
+        # QUS: why is this not in the sherpa-test-data repository?
+        this_dir = os.path.dirname(os.path.abspath(__file__))
+        ui.load_table(1, os.path.join(this_dir, 'data',
+                                      'two_column_x_y.fits.gz'))
+        data = ui.get_data(1)
+        self.assertEqualWithinTol(data.x, [1, 2, 3])
+        self.assertEqualWithinTol(data.y, [4, 5, 6])
 
 if __name__ == '__main__':
 
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -34,9 +34,10 @@ import logging
 logger = logging.getLogger("sherpa")
 
 
+@requires_fits
+@requires_data
 class test_ui(SherpaTestCase):
 
-    @requires_data
     def setUp(self):
         self.ascii = self.make_path('threads/ascii_table/sim.poisson.1.dat')
         self.fits = self.make_path('1838_rprofile_rmid.fits')
@@ -55,15 +56,11 @@ class test_ui(SherpaTestCase):
         self.func = lambda x: x
         ui.dataspace1d(1, 1000, dstype=ui.Data1D)
 
-    @requires_data
-    @requires_fits
     def test_ascii(self):
         ui.load_ascii(1, self.ascii)
         ui.load_ascii(1, self.ascii, 2)
         ui.load_ascii(1, self.ascii, 2, ("col2", "col1"))
 
-    @requires_data
-    @requires_fits
     def test_table(self):
         ui.load_table(1, self.fits)
         ui.load_table(1, self.fits, 3)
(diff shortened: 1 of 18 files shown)


DougBurke:
+1 for the simplified decorator support.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -128,8 +118,10 @@ def test_filter_table(self):
         ui.load_filter(self.filter_single_log_table, ignore=True)
 
 
+@requires_data
+@requires_fits
 class test_more_ui(SherpaTestCase):
-    @unittest.skipIf(test_data_missing(), "required test data missing")
+    @requires_data
```
DougBurke: Is the decorator on `setUp` needed as there's now one on the class?

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -29,7 +29,7 @@
 
 class test_ui(SherpaTestCase):
```
DougBurke: It looks like `@requires_fits` should be moved to the `class test_ui` statement and removed from each method.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -257,20 +241,17 @@ def test_psf_model2d(self):
     # bug #12503
     def test_psf_pars_are_frozen(self):
         ui.load_psf('psf', ui.beta2d.p1)
-        self.assertEqual([], p1.thawedpars)
+        self.assertEqual([], ui.beta2d.p1.thawedpars)
```
DougBurke: I think this would be better written as

    p1 = ui.get_model_component("p1")
    self.assertEqual([], p1.thawedpars)

I'm not sure there's much of a difference, but with my version you also check that the "global" parameter mapping has been updated.

sherpa/image/tests/test_image.py
```diff
@@ -50,10 +50,10 @@ def get_arr_from_imager(im):
     data_out = numpy.float_(data_out)
     return data_out
 
+
+@requires_ds9
 class test_image(SherpaTestCase):
     if (os.environ.has_key("DISPLAY") == True):
```
DougBurke: Can this be changed to

    if "DISPLAY" in os.environ:

olaurino: Yes. However, PR #140 gets rid of the `if` altogether.

sherpa/plot/tests/test_plot.py
```diff
@@ -115,9 +115,9 @@ def test_splitplot(self):
         # sp.addplot(rp)
 
 
+@requires_data
 class test_contour(SherpaTestCase):
```
DougBurke: Other than checking the contour routines do not crash, it doesn't look like these are doing much. Can we add a comment to note this (perhaps as a TODO to indicate needs a closer look)?

sherpa/ui/tests/test_ui.py
```diff
@@ -38,9 +36,9 @@ def calc(self, p, x, *args, **kwargs):
         return p[0]*x+p[1]
 
 
+@requires_data
 class test_ui(SherpaTestCase):
```
DougBurke: Hmmm, more examples of tests without explicit tests!

sherpa/utils/__init__.py
```diff
@@ -306,15 +309,39 @@ def has_package_from_list(*packages):
...
+def requires_fits(test_function):
     """
     Returns True if there is an importable backend for FITS I/O.
     Used to skip tests requiring fits_io
     """
-    return has_package_from_list('pyfits',
-                                 'astropy.io.fits',
-                                 'pycrates',
-                                 )
+    packages = ('pyfits',
```
DougBurke: Should `astropy.io.fits` be listed before `pyfits` here? I assume that `requires_package` works from left to right, and I would rather the astropy version be tested on the off chance someone has both installed.

sherpa/image/tests/test_image.py
DougBurke: Okay.

sherpa/astro/ui/tests/test_astro_ui.py
olaurino: I understand your point, but that feature should have its own unit tests. I find it annoying that many files show up with "undefined global name"s in my IDE, even though they are fine. I'll change it to use get_model_component for now (not sure why I didn't in the first place), and I am sure there is plenty of places were the tests assume the pollution of the global namespace.

sherpa/utils/__init__.py
olaurino: Hmm... I don't think that makes a difference, actually. In case someone has both installed, the condition will be satisfied by pyfits, but then astropy will be selected as a backend. The purpose of this check is just to understand whether a test has to be run or skipped, not which backend to load.

sherpa/plot/tests/test_plot.py
olaurino: Good idea. I started looking into ways to test the output of the plotting backends, but that's not a priority unless we make it such.

olaurino:
Rebasing to disentangle from #116. Original tip:  535e003

sherpa/astro/ui/tests/test_astro_ui.py
olaurino: No, it's not. Thanks.

sherpa/astro/ui/tests/test_astro_ui.py
olaurino: correct, thanks.

olaurino: @requires_data can also be promoted. There is in principle one test that might run without test data, but it's in a class with a setUp method that requires test data, so that method would not be run anyway (and maybe that's a problem)

olaurino:
@DougBurke commits 15e8296 (an edit of the original one) and 8aaffbe should address your comments.

I'll wait for your thumbs up to merge. Thanks.

olaurino:
@DougBurke the new commits should address your remaining comments (and should fix the failing tests, but we'll need Travis to tell us).

olaurino:
ouch, I should have rebased it. Hold on.

olaurino:
@DougBurke this should be it.

DougBurke:
I'm happy for this to be merged, but I just thought about moving the `<system>_required` routines (and other related code) from `sherpa.utils` to `sherpa.utils.test` (or something similar), since it's a bit cleaner. This could of course be done as a separate PR, but it might make sense to do it now, since this is where these new decorators are being added.

There's also likely a few more chances for consolidation of the test decorators from individual methods to the test class, but that can be done in more-focused PRs after this has been merged.

olaurino:
@DougBurke I agree with the move to sherpa.utils.test, but that would require moving SherpaTest too.

And sherpa.utils would most likely benefit from a refactoring beyond the test utils, as it is currently >2500 lines long.

I'd leave it alone for now. Maybe we can add a new issue for this?

DougBurke:
I'm okay with merging this as is (although it'll mean more work for #151 :-) and opening a new ticket for moving the test routines out of `sherpa.utils` (as a small step to cleaning up that module).

PR #144, opened 2016-01-14
Add missing skipIf and ensure legacy test runner fails when tests fails

olaurino:
This PR adds a missing `skipIf` to a test. The Travis CI build that was supposed to catch this issue did not fail because the legacy test runner that is triggered by that build (what we call the "smoke" test, i.e. a test run with the `sherpa.test()` function) was not failing even though the tests were.

So, this PR also ensures that failing tests make the Travis build fail by throwing an Exception. This behavior was tested locally.


DougBurke:
Looks good to me.

olaurino:
Perfect, then I'll merge it. Thanks!

DougBurke:
Once this goes in there's an extra change needed for #141 (i.e. to update this skipIf decorator to use the new scheme from #141), but it looks like work is needed on 141 anyway to get it to merge.

olaurino:
Correct.

PR #145, opened 2016-01-14
make xspec verbose during tests, in case of failure

olaurino:
This PR sets the xspec chatter level to 50 by default when running tests.

Despite the chatter level, pytest will redirect the stdout and err, and none will be printed to the console if everything is OK.

However, if an error occurs, the output from xspec will be logged along with the error.

This PR should not impact the current CIAO testing infrastructure.

### DIFF
--- a/helpers/test.py
+++ b/helpers/test.py
@@ -17,6 +17,15 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+
+def set_xspec_chatter():
+    try:
+        from sherpa.astro import xspec
+        xspec.set_xschatter(50)
+    except ImportError:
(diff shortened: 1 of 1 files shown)


DougBurke:
Note that - at present - the maximum value of the XSPEC chatter level is 25 - e.g. see https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSchatter.html - but setting it to 50 should not be a problem here (I know I suggested 50; I can never remember what the max value is).

DougBurke:
The copyright statement in `helpers/test.py` needs updating to include `2016`.

DougBurke:
Other than the copyright date, this PR is good to be merged.

olaurino:
For the record, I checked with @jbudynk that this change should not interfere with CIAO's testing infrastructure.

PR #151, opened 2016-01-19
Ensure astropy and crates behave the same with gzipped files

DougBurke:
# Release Notes

Change the behaviour of the AstroPy back end so that it matches that of Crates when given a file name which does not exist, but a compressed version, with the suffix `.gz` does. The Crates behavior is to read the file. This extends to PHA files whose ancillary files - e.g. those stored in the `BACKFILE`, `ANCRFILE`, and `RESPFILE` keywords - are given as unzipped names, but only the gzipped names exist on disk.

As an example: if `pha.fits.gz` exists but `pha.fits` does not, then

    load_pha('pha.fits')

will now load the file with either back end. If the response files are set to `arf.fits` and `rmf.fits` (via the `ANCRFILE` and `RESPFILE` keywords), but only the `.gz` versions exist, then they will now also be loaded by the AstroPy back end.

Tests have been added to cover this case (for both I/O backends).

# Notes

This is a rebase of PR #105.

This is particularly useful when loading Type 3 PHA files - as produced
by the Chandra Source Catalog.

An explicit import statement is used in sherpa/astro/ui/utils.py (PEP8 fix).

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -269,12 +269,21 @@ def _try_vec_or_key(hdu, name, size, dtype=SherpaFloat, fix_type=False):
 # and 'foo.fits.gz' exists, then it will use this). This requires some
 # effort to emulate with the astropy backend.
 
-def is_binary_file(fname):
+def _infer_and_check_filename(filename):
+    fname = filename  # keep filename unchanged
+    if not os.path.exists(fname):
+        fname += '.gz'  # try to find a gzipped version, following CXC/Crates conventions.
+        if not os.path.exists(fname):  # fail fast
+            raise IOErr('filenotfound', filename)  # error message reports the original filename requested
+    return fname
+
+
+def is_binary_file(filename):
     """Do we think the file contains binary data?
 
     Parameters
     ----------
-    fname : str
+    filename : str
        The file name.
 
     Returns
@@ -289,10 +298,7 @@ def is_binary_file(fname):
     attempts to transparently support reading from a `.gz`
     version if the input file name does not exist.
     """
-
-    if not os.path.exists(fname):
-        fname = fname + '.gz'
-
+    fname = _infer_and_check_filename(filename)
     return sherpa.utils.is_binary_file(fname)
 
 
@@ -312,16 +318,8 @@ def open_fits(filename):
        The return value from the `astropy.io.fits.open`
        function.
     """
-
-    try:
-        return fits.open(filename)
-    except IOError as e:
-        try:
-            # This will try 'foo.gz.gz'.
-            return fits.open(filename + '.gz')
-        except IOError:
-            # Raise the original error
-            raise e
+    fname = _infer_and_check_filename(filename)
+    return fits.open(fname)
 
 
 def read_table_blocks(arg, make_copy=False):
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -47,7 +47,8 @@ from numpy.compat import basestring
 import os
 from itertools import izip
 from sherpa.utils.err import IOErr
-from sherpa.utils import SherpaInt, SherpaUInt, SherpaFloat, is_binary_file
+from sherpa.utils import SherpaInt, SherpaUInt, SherpaFloat
+import sherpa.utils
 from sherpa.io import get_ascii_data, write_arrays
 from sherpa.astro.io.meta import Meta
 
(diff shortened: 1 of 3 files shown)


DougBurke:
This is a replacement for PR #105 - where I have just rebased the commits (and possibly lost one unrelated code-cleanup change, but that's can be handled by a different PR, if true).

olaurino:
Tests introduced in this PR are not running (and making the tests fail) in the snapshot branch, e.g.:
https://travis-ci.org/sherpa/sherpa/jobs/103685001

because of #141 being merged.

This branch should be rebased on top of the new master and edited accordingly. I'll do it tomorrow.

DougBurke:
I guess that #156 will also be an issue.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -263,6 +264,66 @@ def _try_vec_or_key(hdu, name, size, dtype=SherpaFloat, fix_type=False):
...
+
+    Notes
+    -----
+    This is a wrapper around the version in sherpa.utils which
+    attempts to transparently support reading from a `.gz`
+    version if the input file name does not exist.
+    """
+
+    if not os.path.exists(fname):
+        fname = fname + '.gz'
```
olaurino: Maybe this should fail fast if neither files exist, rather than passing the call downstream and wait for is_binary_file to fail. It would also be more explicits, i.e. without relying on the implementation of is_binary_file. I had to open up that file to understand what would happen if neither file existed.

olaurino:
Rebasing to resolve conflicts. Original tip was cce6cb1.

After fixing the conflicts the tests would fail anyway, so I compensated for changes in the testing infrastructure and amended the original commit in order to keep it "green".

sherpa/astro/io/pyfits_backend.py
DougBurke: Several issues here:

* I am not convinced `is_binary_file` is a good idea in the first place, but this is not the PR to work on this issue

* if the routine were to fail if both `fname` and `fname` + `.gz` do not exist (I think that's what you are suggesting) then we need to make sure the same error is raised as is done by `is_binary_file`: is it `IOError` or `IOErr` (hopefully no other choice). This would then lead to a coupling of concerns - e.g. if we change `is_binary_file` to raise a different error then it would also have to be changed here.

* We should probably have a discussion at some point about Sherpa's error class (I believe I was one of the people originally wishing for a sherpa-related error class, so I shouldn't complain, but I'm not sure the pain is worth the gain). If we got rid of the possibility of `IOErr` then it's more obvious that raising `IOError` here is an okay thing to do. Actually, I think this is likely to be part of our Python 3 transition discussion, so should probably be thought about then.

The item is the issue here, so it's a discussion between code clarity and having repeated code.

olaurino: Note that if fname does not exist and fname+gz does not exist either, is_binary_file will tell you that fname+gz does not exist, which is confusing, because you never asked to load fname+gz. I have a patch for this, I am just trying to figure out whether it makes sense to add a unit test, given that we still can't use mock (it's in the process of being approved as a CIAO OTS).

olaurino: Of course, me submitting a patch does not imply we need to take it. Just that it's easier to discuss about two existing alternatives.

DougBurke: I'm away doing other things this afternoon, so won't be back at this until Monday.

olaurino:
I added a new commit. We can remove it if necessary. By the way, as far as I can tell I agree that is_binary_file seems to add complexity that might not be required, but as you said this is not for this PR to fix.

olaurino:
@DougBurke what do you think about my latest commit?

DougBurke:
@olaurino I am not sold on your addition, but neither do I think it's terrible, so I vote to keep it in. I think we need to come up with a consistent story for how file I/O should be done in general (do we do lots of file exists/is binary checks, or just let the code try and open the file [*]), and revisit this all at a later date.

[*] which can then probably support more back-end specific options (e.g. if one backend supports URLs); it would also might be useful to allow streams or `StringIO` inputs as well.

DougBurke:
I did a few quick spot checks to check that the CIAO version still behaves and it seems to, so it looks good to me. I've just pushed a minor change to update the copyright year and remove an unused import from the new test file.

olaurino:
The push Travis build is failing because of the mkl/nomkl issue, and the pr build is failing because of the libgfortran issue. Let's rebase this PR once we fix those issues for good and they are in the master branch. Ok?

DougBurke:
Okay with me.

DougBurke:
@olaurino - can we have a PR with just the libgfortran change in? I note there's several PRs where you've applied it, and it would be nice to get just that change in (since it's a travis-only change it doesn't need to wait for CIAO testing) rather than having to wait for one of the PRs it is in to be merged? Or is there a PR about to be merged that it could be added to?

olaurino:
@DougBurke that makes sense. I had to test these PRs so I cherry-picked the fixes in more than one branch. I now created a PR that I am going to merge as soon as Travis is green, just for the libgfortran fix. We'll need to rebase the others when necessary.

olaurino:
Rebasing to get the nomkl and libgfortran fixes. Original tip was at fabc2ae.

olaurino:
If Travis is happy this can be marked as `pr:merge`.

DougBurke:
Is this code in ciaod?

olaurino:
I believe it is, with the old commits 0e55187 and 71493d6.

DougBurke:
I think we should hold off pr:merge until we've got the CIAOX build working, so we can check that everything has been integrated correctly (just trying to avoid too many things that could go wrong at once).

olaurino:
Rebasing to include a fix to the travis builds. Original tip was at 6a21e86.

olaurino:
@DougBurke this is in today's CIAOD. Next CIAOX would have this PR if we merged it today. I would suggest we merge it.

DougBurke:
If the tests have passed then I am happy (I don't have time to actually try
it out in ciaod today)

PR #153, opened 2016-01-19
fix chi2datavar comparison <=0 bug

dtnguyen2:
The comparison test within calc_chi2datavar _errors is too stringent, after all sqrt(x) where x >=0 is legit.

### DIFF
--- a/sherpa/stats/tests/test_stats.py
+++ b/sherpa/stats/tests/test_stats.py
@@ -308,8 +308,7 @@ class test_stats(SherpaTestCase):
         ui.load_arrays(1, xy, xy, Data1D)
         ui.set_stat('chi2datavar')
         err = ui.get_staterror()
-        for index, yyy in enumerate(xy):
-            assert numpy.allclose(err[index], numpy.sqrt(yyy), 1.0e-7, 1.0e-7)
+        numpy.testing.assert_allclose(err, numpy.sqrt(xy), rtol=1e-7, atol=1e-7)
 
 
 def tstme(datadir=None):
--- a/sherpa/stats/tests/test_stats.py
+++ b/sherpa/stats/tests/test_stats.py
@@ -20,14 +20,14 @@
 import os.path
 
 import numpy
-import unittest
 
-from sherpa.utils import SherpaTest, SherpaTestCase
 from sherpa.utils import requires_data, requires_xspec, requires_fits
+from sherpa.utils import SherpaTest, SherpaTestCase
+from sherpa.data import Data1D
 
 from sherpa.models import PowLaw1D
 from sherpa.fit import Fit
-from sherpa.stats import Stat, Cash, LeastSq, UserStat, WStat
+from sherpa.stats import Cash, UserStat, WStat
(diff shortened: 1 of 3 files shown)


DougBurke:
I'd like to see a test added for this.

sherpa/stats/tests/test_stats.py
```diff
@@ -299,6 +299,15 @@ def test_wstat_error(self):
         fit = Fit(data, self.model, WStat(), NelderMead())
         self.assertRaises(StatErr, fit.fit)
 
+    def test_chi2datavar(self):
+        num = 3
+        xy = numpy.array(range(num))
+        ui.load_arrays(1, xy, xy, Data1D)
+        ui.set_stat('chi2datavar')
+        err = ui.get_staterror()
+        for index, yyy in enumerate(xy):
```
DougBurke: Numpy includes several routines for testing whether arrays match, within a tolerance, which would avoid the explicit looping (actually, `numpy.allclose` also does so as well, but it just returns `True` or `False` and the testing routines work with `unittest`). The routines are in [`numpy.testing`](http://docs.scipy.org/doc/numpy/reference/routines.testing.html) and probably the one wanted here would be [`numpy.testing.assert_allclose`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose), which should let you replace the for loop by (untested, assuming the symbol has been explicitly imported)

    assert_allclose(err, numpy.sqrt(xy), rtol=1e-7, atol=1e-7)

It also provides a more-useful error message than the `assert` statement used on line 309.

olaurino:
Rebasing, original tip was at 957584e

sherpa/stats/tests/test_stats.py
DougBurke: I should also note that we do have Sherpa specific "test an array for equality" routines  somewhere in `SherpaTest or `SherpaTestCase` but I think these should probably be marked as deprecated, since there seems little advantage to them compared to the numpy ones (but this needs review, and is not really part of this particular PR).

olaurino:
Rebasing for fixing dependency issue in one of the travis builds. Original tip was at 7a10adc.

olaurino:
@DougBurke @dtnguyen2 if I understand correctly as long as Travis is happy with the change I made to implement @DougBurke's suggestion (99902f5) this PR is good to go, right?

DougBurke:
Looks good to me. I just checked and the test does fail with the existing code (I used Sherpa in CIAO 4.8 rather than master for this):

```
% sherpa
CIAO 4.8 Sherpa version 1 Thursday, December 3, 2015

sherpa-1> xy = np.arange(3)
sherpa-2> load_arrays(1, xy, xy, Data1D)
sherpa-3> xy
          array([0, 1, 2])
sherpa-4> set_stat('chi2datavar')
sherpa-5> get_staterror()
ValueError: calculation of errors has failed using current statistic
```

I note that the test could be cleaned up a little bit (e.g. use of `np.arange(..)` rather than `np.asarray(range(..))`, but that doesn't need to be done in this PR.

PR #155, opened 2016-01-20
add argument to get_draws for supplying a covariance matrix

olaurino:
# Release Note
The `get_draws` function now accepts a user-provided covariance matrix. If no covariance matrix is provided, the covariance matrix computed by the default implementation is used. Note that `covar()` must be invoked before invoking `get_draws` if no covariance matrix is provided, otherwise `get_draws` will exit with an error.

# Description
This PR introduces a new attribute to the `get_draws` function, for allowing users to provide a custom covariance matrix.

# Testing
Tests provide as much coverage as possible. Unfortunately, without `mock` some advertised behavior cannot be checked (at least not in a simple, sustainable way). For instance, it cannot be checked that if no covariance matrix is provided then `extra_output` is used. This means that some lines are not covered by tests.

I planned to add additional tests in a similar fashion to what was done with #196 (datastack tests), but it would be complicated to include commits from #196 and then keep up with changes. I don't think it's worth it in this case. We'll add tests if possible in a future pass.

An additional test is provided in c6b0345, but for now it is not included in the PR because it requires a change in sherpa/sherpa-test-data@03600ca, and I am not sure this is worth the footprint.

### DIFF
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -206,7 +206,6 @@ import numpy
 import logging
 info = logging.getLogger("sherpa").info
 _log = logging.getLogger("sherpa")
-_level = _log.getEffectiveLevel()
 
 _tol = numpy.finfo(numpy.float).eps
 
@@ -670,6 +669,7 @@ class MCMC(NoNewAttributesAfterInit):
             raise ValueError("Fit statistic must be cash, cstat or wstat, not %s" %
                              fit.stat.name)
 
+        _level = _log.getEffectiveLevel()
         mu = fit.model.thawedpars
         dof = len(mu)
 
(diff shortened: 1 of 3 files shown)


olaurino:
Rebasing so to make the travis builds pass. Original tip was at 2b2b743.

anetasie:
@olaurino Thanks for the test. 
I realized that we also need to test the data ids in _get_draws_. By default it runs on one data id and if one wants to run on more than one id then the list of _otherids_ needs to be given. Perhaps there is such a test already included in the sherpa tests.

olaurino:
@anetasie may I suggest we merge this (if it looks good to you) and we create a new issue for the additional tests? In any case there weren't many tests for `get_draws` before this PR, the ones I found did not check the validity of the output, and I could not find any tests for `otherids`.

anetasie:
I just labeled it for the merge.

PR #156, opened 2016-01-21
WIP Move test code out from sherpa.utils

DougBurke:
# Status

This is on hold until after the Python 3 migration has been completed. It can be reviewed at that time (added 30 June 2016).

# Release Notes

There is no functional change in the PR.

Move the test code out from `sherpa.utils` to `sherpa.utils.test` and
update the test code to match. Some minor PEP8 cleanup (mainly handling
of space characters, but several changes to use explicit import
statements).

# Notes

This is based on the discussion of PR #141 where it was noted that
the `sherpa.utils` module is too large and needs separating out. This
is the first step in the process.

The PR adds a new file: `sherpa/utils/test.py`. Does this need to be added to any manifest-like file in the distribution?

# Test of the changes

The test suite passes with these changes, with the same number of tests
run (and tests skipped, as the tests were ran without ds9 or XSPEC
support) as before the code change.

### DIFF
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2007,2014,2015  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2007, 2014, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -109,9 +109,7 @@ def test(level=1, verbosity=1, datadir=None):
     (including the discipline-specific ones)
 
     """
-    # import sherpa.all
-    # import sherpa.astro.all
-    from sherpa.utils import SherpaTest
+    from sherpa.utils.test import SherpaTest
     SherpaTest().test(level, verbosity, datadir)
 
 
--- a/sherpa/astro/datastack/tests/test_datastack.py
+++ b/sherpa/astro/datastack/tests/test_datastack.py
@@ -18,10 +18,9 @@
 #
 
 
-from sherpa.utils.test import SherpaTestCase, requires_fits
+from sherpa.utils.test import SherpaTest, SherpaTestCase, requires_fits
 import os
 import sys
-import unittest
 from sherpa.astro import ui
 from sherpa.astro import datastack
 from acis_bkg_model import acis_bkg_model
@@ -690,9 +689,6 @@ class test_query(SherpaTestCase):
 
 if __name__ == '__main__':
 
-    from sherpa.utils import SherpaTest
-
-    import sys
     if len(sys.argv) > 1:
         datadir = sys.argv[1]
     else:
--- a/sherpa/astro/datastack/tests/test_datastack.py
+++ b/sherpa/astro/datastack/tests/test_datastack.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2014, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2014, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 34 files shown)


DougBurke:
This could have been made less invasive, by having `sherpa.utils` re-export the functions from `sherpa.utils.test`, but I wanted to make sure I had caught everything, and I don't think that we need to enforce API stability for the test routines.

olaurino:
It is probably a good idea to re-export the names anyway, to reduce the risk of breaking both our code and users', especially now that we know sherpa is a dependency for external projects.

DougBurke:
I'm not convinced that I can get the re-export case to work, due to import loops. If the following is placed in `sherpa.utils.__init__.py`

```
class SherpaTestCase(sherpa.utils.test.SherpaTestCase):
    """Deprecated: use sherpa.utils.test.SherpaTestCase"""
    def __init__(self, *args, **kwargs):
        m = 'SherpaTestCase is deprecated: use sherpa.utils.test.SherpaTestCase'
        warnings.warn(m, DeprecationWarning, stacklevel=2)
        sherpa.utils.test.SherpaTestCase.__init__(*args, **kwargs)
```

I get test failures like

```
sherpa/astro/datastack/tests/test_datastack.py:21: in <module>
    from sherpa.utils.test import SherpaTest, SherpaTestCase, requires_fits
sherpa/utils/__init__.py:2582: in <module>
    class SherpaTestCase(sherpa.utils.test.SherpaTestCase):
E   AttributeError: 'module' object has no attribute 'utils'
```

This would presumably go away if I moved `sherpa/utils/test.py` to `sherpa/test.py` (or anywhere out of `sherpa/utils/`); is this a sensible change?

olaurino:
@DougBurke I can see at least one issue with your code, i.e. that you are not passing `self` to `SherpaTestCase.__init__`. However, that wouldn't explain the error you get anyway. I suspect there is something wrong with the code in `utils/__init__.py` that is preventing the module to load. I don't think it's an issue with import loops. Do you have a commit I can look at?

DougBurke:
@olaurino - you can try https://github.com/DougBurke/sherpa/tree/playground/separate-test-utils but I'm not 100% sure I've got everything correct there. The tests for me fail with

```
______________ ERROR collecting sherpa/utils/tests/test_utils.py _______________
sherpa/utils/tests/test_utils.py:21: in <module>
    from sherpa.utils import SherpaFloat
sherpa/utils/__init__.py:2574: in <module>
    class SherpaTest(sherpa.utils.test.SherpaTest):
E   AttributeError: 'module' object has no attribute 'utils'
```

but then again they did so before the last check in (which is where I think I added in the "backwards compatibility" symbols, but can't be 100% sure as this was a few days ago and I've already forgotten what I was doing).

olaurino:
I made some changes and added some tests in a branch on my fork. I'll issue a PR against your fork for your convenience, then you decide what to do with it.

DougBurke:
I've integrated your changes @olaurino. I guess we'd need to take off pr:hold to check that this works in CIAO? At least, after we've resolved the current travis issues (eg #191)?

olaurino:
That's what I'd guess as well. I did merge #191, so I believe this branch can be rebased on top of master (and it looks like there are conflicts to address).

DougBurke:
I was going to try rebasing this against master, but if you are going to merge #178 soon I'll hold off (there shouldn't be any overlap but I'd rather avoid extra work if I can help it).

olaurino:
Rebasing to resolve conflicts and get travis build fixes. Original tip was at 919d71d.

olaurino:
The rebase also included copying over the `run_thread` function from PR #164 and its new `sherpa-test-data`.

DougBurke:
I removed the priority:low label since we now have the milestone as 4.8.2, and it's not a low-priority for that release.

DougBurke:
Given that this is going to conflict with #210, is it worth closing this PR? We could then have a low-priority item to break up `sherpa.utils` module (i.e. expand the scope to more than just the testing code), perhaps related to any documentation work (since a better module structure is likely to rpoduce better docs).

olaurino:
This is too out of date to be rebased. We could implement the same changes in a brand new PR. Closing.

PR #158, opened 2016-01-22
fix ascii file region issue, add unit tests

olaurino:
# Release Note
[release-note]
Fixed bug that prevented region ascii files to be read in standalone Sherpa.
[/]

# Description
The problem was that if the region was read from a file, the standalone region extension code wrapped the filename inside a "region()" string, but this caused the low level region library call to fail. Removing the unnecessary wrapping makes the function happy again. In the patched region extension we use in CIAO we always call dmRegParse whether or not the input is a filename, and we always wrap the region expression, or filename, inside "region()".

This PR also introduces some basic unit tests.

### DIFF
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -93,7 +93,7 @@ static regRegion* parse_string( char* str, int fileflag ) {
   regRegion *reg = NULL;
   std::string input(str);
   if( fileflag ) {
-      input = "region(" + input + ")";
+//      input = "region(" + input + ")";
   	  reg = regReadAsciiRegion( (char*)input.c_str() , 0 ); // Verbosity set to 0
   } else {
 	  reg = regParse( (char*)input.c_str() );
--- a/sherpa/astro/utils/tests/test_region_unit.py
+++ b/sherpa/astro/utils/tests/test_region_unit.py
@@ -1,0 +1,48 @@
+#
+#  Copyright (C) 2016  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


anetasie:
I tested the fix in 'region-tests' branch and my script works.

The user would see an error when the filter expression was contained in the ascii file, so the following expression would failed. 
    ignore2d('file.reg')  
I checked that this is now working as expected.


In [5]: load_image("image.fits")

In [6]: ignore2d('jet.reg')

In [7]: show_data()
Data Set: 1
Filter: !Pie(4028.61,4119.64,2.74445,6.70732,313,360)
name      = image.fits
x0        = Float64[4096]
x1        = Float64[4096]
y         = Int16[4096]
shape     = (64, 64)
staterror = None
syserror  = None
sky       = physical
 crval    = [ 4013.5511, 4103.8305]
 crpix    = [ 0.5, 0.5]
 cdelt    = [ 0.5, 0.5]
eqpos     = world
 crval    = [ 227.4951,  57.0422]
 crpix    = [ 4096.5, 4096.5]
 cdelt    = [-0.0001, 0.0001]
 crota    = 0
 epoch    = 2000
 equinox  = 2000
coord     = logical
"

anetasie:
I tested the fix in the region-tests branch and confirmed that it fixes the problem.

2D filters supplied in the ascii file were not parsed. This PR fixes the problem.
Here are the steps in my script related to the failure which was fixed.
   
    In [5]: load_image("image.fits")
    In [6]: ignore2d('jet.reg')
    In [7]: show_data()
          Data Set: 1
          Filter: !Pie(4028.61,4119.64,2.74445,6.70732,313,360)
          name      = image.fits
.....

PR #159, opened 2016-01-22
enforce usage of the python executable when building deps

### LINKED ISSUE #19: Set python used to launch setup.py when building extern
This ticket replaces one of the issues reported in #15.

The `configure` script of external python extensions coming from CIAO libraries tries to infer the python version and picks up the first `python` available in the `PATH`. However, when users explicitly use a different version of python (e.g. `python2.7` in an environment where `python` points to Python3, as reported in #15) the `configure` script should not try to infer what version of Python to run, but just use the one specified by the user.

A workaround exists: `export PYTHON=python2.7` (BASH) would tell `configure` what is the Python executable the user wants to run.

olaurino:
Resolve #19 

# Release Note
[release-note]
The `python` executable used to call `setup.py` is now injected in the environment that kicks off the external dependencies build. This means that if `python2` is used to call `setup.py` but `python3` is the first `python` executable in the `$PATH`, `python2` will be used to run both the python and the external builds.
[/]

# Testing
I tested this locally by playing with the $PATH and using a python 3 conda environment. I could set up a Travis build to automatically test this, if somebody argues for it during review, but in my opinion this is not necessary. It would be good if the original reporter tested this fix, though, and I'll ask him in the comments.

### DIFF
--- a/helpers/deps/__init__.py
+++ b/helpers/deps/__init__.py
@@ -21,6 +21,7 @@
 from subprocess import call
 from multiprocessing import cpu_count
 import os
+import sys
 
 def clean_deps():
     prefix = os.getcwd()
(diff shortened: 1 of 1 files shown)


olaurino:
@cdeil this should fix #19. It would be nice if you could confirm it, as you originally reported the issue.

Please let me know if you have any questions.

DougBurke:
@olaurino I agree with you that we do not need a Travis build to test this use case.

cdeil:
Yes, this does fix #19 for me. Thank you!

With master I still see Python 3 used and the build fail if I run:
```
python2 setup.py install --user
```
With this branch the build / install / tests work.

DougBurke:
Assuming the tests pass then this can be merged.

Issue #15, opened 2015-04-25
grplib uses Python from PATH


cdeil:
I'm getting this configure error for grplib when I try `python2.7 setup.py install` for Sherpa:
https://gist.github.com/cdeil/39e9404e91954cb67270#file-gistfile1-txt-L284

It tries to use python3.4, which is the `python` on my PATH.
Instead it should use the Python with which I call setup.py.

taldcroft:
@cdeil - yeah, that looks like a problem in configure.  At least for now the easiest workflow for sherpa development is going to be an isolated Anaconda environment or something similar.

olaurino:
@cdeil thanks for the report

I see two issues here:
  - as you reported, the build is not taking into account that you want to force python2.7, despite the fact that python3 has the precedence in your PATH setup.
  - the macro that should ensure python's version is 2.7 is not doing its job: you should get a meaningful error message if building with python3.

Regarding the first issue, here is a workaround: just `export PYTHON=python2.7` before you run `python2.7 setup.py install`. Then the `configure` script will not try to infer the Python version itself. I will take this report as a suggestion for a feature enhancement so to make this transparent to the user: if python2.7 is used to run `setup.py` it should also be used to build the python extensions built externally.

Regarding the second issue, it is an actual bug. Unfortunately, a fix would require to rerun the `autotools`, and I need to coordinate this with the `grplib` maintainers, as this code is taken from the `CIAO` codebase.

I will create two separate tickets, as I believe these two issues will be addressed separately.

olaurino:
This issue has been replaced by #19 and #20

olaurino:
@cdeil if you agree I would like to close this issue: your report will be addressed in #19 and #20 separately.

cdeil:
@olaurino – Sure, splitting separate issues into different tickets is always a good idea. Closing this one now.

PR #160, opened 2016-01-22
fix configure scripts so they fail if python3 is used

### LINKED ISSUE #20: AX_PYTHON_DEVEL macro fails to report Python 3 is not supported
This ticket replaces one of the issues reported in #15.

When python3 is detected, the `configure` script should gracefully fail informing the user python3 is not supported.

However, this does not happen (see https://gist.github.com/cdeil/39e9404e91954cb67270#file-gistfile1-txt-L276).

The `AX_PYTHON_DEVEL` call in `configure.ac` should be changed to `AX_PYTHON_DEVEL([<'3.0'])`
and `autoreconf` should be run. We need to coordinate this with the maintainers of the group library.

olaurino:
Resolve #20 

# Description
This PR fixes a configure issue by which python 3 was not detected when building the extern dependencies.

I applied the fix to our version of the group and stk library. These fixes should be made into the CIAO codebase as well, but that can happen at a later time, especially considering that we are currently discussing a possible CM change to these libraries.

# Testing
I tested this fix locally together with #159.


extern/grplib-4.7/configure
```diff
@@ -13898,7 +13898,8 @@ $debug ||
 if test -n "$CONFIG_FILES"; then
 
 
-ac_cr=''
+ac_cr='
+'
```
olaurino: I just noticed git automatically removed a spurious CR (Carriage Return ^M) character from here. This should not have any effect on the scripts, but I'll double check.

olaurino: I removed this change from the commit, just to be safe.

DougBurke:
@olaurino so, is the issue that checks like `[>'2.7' <'3.0']` weren't error-ing out for Python 3 users?

What happens if a user tries to use Python 2.6 to build? Do we still get an error (i.e. I'm slightly concerned about the removal of the `>2.7` check)?

olaurino:
We do, because of this line:
https://github.com/sherpa/sherpa/pull/160/files#diff-28700aed04a9a4f1ab9ff8f6bcb2dbe2R15

In any case, I checked the documentation of the macro, and I can indeed use ==2.7 in order to pinpoint the version number. I'll make this change and test it. I don't think it makes too much of a difference, but it's safer.

DougBurke:
Sounds sensible to be more-explicit here. It's unlikely that people will use an older Python, but just in case!

olaurino:
I tried making the change and I now remember why I didn't do it that way to begin with. The more I work with this macro the less I like it. I have a patch ready to be submitted, but I still have to properly test it. Eventually, the macro itself should be changed, but I am holding off any deeper changes until we decide the future of these OTS.

olaurino:
This seems to work. It's ugly, but it's the least extensive change I could come up with.

DougBurke:
I would say looks good to me, but its `autoconf`, so that's not possible ;-)

I think this is good to be merged, unless you want @cdeil to check it as this is part of the fix for #15 ?

cdeil:
My default Python on the command is Python 3.5 and both with master and this branch I see:
```
$ python setup.py install
  File "setup.py", line 50, in <module>
  File "/Users/deil/code/sherpa/helpers/__init__.py", line 21, in <module>
    from clean import clean
ImportError: No module named 'clean'
```
I think that's what most users and devs will see, i.e. they don't even get far enough in the build to see this change in the configure.
I don't have 2.6 available any more for testing, so I don't know if the build error reporting is improved with this PR.

As I've mentioned elsewhere already, I think maybe starting by making `python setup.py egg_info` work with Python 3 and then trying to port parts of Sherpa to Python 3 in small pieces (while always keeping Python 2 compat) would be nice (i.e. a realistic option to get going, for me and others to contribute). I don't know if this comment has any relevant on this PR. I guess no, since Python 3 is prevented before and after this change by this configure?

DougBurke:
@cdeil - thanks for the feedback.

So, I would say this PR can be merged, but we have more to do to address
the original issue.

olaurino:
That is why I didn't ask @cdeil to confirm this one, and I just tested it together with #159.

PR #166, opened 2016-01-29
update travis ci to include new version of xspec

olaurino:
This PR introduces a new xspec build in Travis CI, with xspec binaries for xspec v12.9.0i.

I am keeping the old version for now, but that can be up for discussion.

Travis seems to be happy.


DougBurke:
:+1:

Issue #150, opened 2016-01-18
iminuit


dtnguyen2:
Finally got a chance to read the mail titled "shared model package?" from C. Deil dated Dec 2, 2015 as forwarded to sherpadev from Aneta. Perhaps, we should consider shipping Sherpa with iminuit.  In theory, one could make it so that if the user has installed iminuit then the user has another optimization option to use otherwise no love loss (kind of like xspec now).

cdeil:
If adding a MINUIT wrapper to Sherpa sounds useful enough to you, this is something I could implement.

Having MIGRAD as an optimizer, HESSE as a parabolic error estimator and MINOS as a profile error estimator would be nice. I don't know how they compare in terms of performance and robustness with what's already available in Sherpa, i.e. if the methods / implementation is sufficiently different to warrant adding it. My guess is that it's a useful option to have available.

@dtnguyen2 - Adding just a wrapper that does `import iminuit` (see https://github.com/iminuit/iminuit) might be the better solution? If you bundle MINUIT it adds quite a few files / build time (see https://github.com/iminuit/iminuit/tree/master/Minuit), though if that is preferably, it's of course possible as well.

dtnguyen2:
@cdeil - I think it would be useful to have MINUIT added to sherpa.  You are already familiar with MINUIT and if you don't mind doing the work then by all means do it.  If you need help with incorporating it to sherpa then let us know but basically there are only two files (sherpa/optmethods/__init__.py and sherpa/optmethods/optfcts.py) to modify.  Just make sure you return the same quantities like the other optimization methods.

cdeil:
I see that @dtnguyen2 tried this in #194 but it wasn't merged.

I am trying to develop something in `gammapy.utils.fitting`, trying to create a uniform interface to MINUIT and Sherpa. See https://github.com/gammapy/gammapy/pull/1922 . See also my comment in https://github.com/sherpa/sherpa/issues/133#issuecomment-438462083 .

One issue I only realised now is that Sherpa has "stateless" optimiser and error estimator backends, whereas `iminuit.Minuit` is stateful, i.e. probably one has to hold on to the `Minuit` object in between computations like `migrad`, `hesse`, `minos`, `contour`, ... So it doesn't really fit in the Sherpa design.

I've asked the MINUIT / iminuit experts about this question here:
https://groups.google.com/forum/#!topic/scikit-hep-forum/KRxFdkEj3H0
but I'm not very hopeful.

[...]

PR #168, opened 2016-02-01
Errors in sherpa.opt pointed out by pyflakes

cdeil:
I had a look at `sherpa/optmethods/opt.py` today because I want to start working on #150 next week.

PyCharm pointed out a few errors when I opened up the file.
I think they are real and this is code that isn't covered by tests?
It's also possible to see those with e.g. pyflakes:
```
$ pyflakes-2.7 sherpa/optmethods/opt.py 
...
sherpa/optmethods/opt.py:82: undefined name 'InputError'
sherpa/optmethods/opt.py:224: undefined name 'nparx'
sherpa/optmethods/opt.py:243: undefined name 'result'
sherpa/optmethods/opt.py:244: undefined name 'result'
...
```

I think there's also real issues pyflakes can find in other files ... concretely I saw this:
```
$ pyflakes-2.7 sherpa/estmethods/ | grep 'undefined name'
sherpa/estmethods/__init__.py:316: undefined name '_est_funcs'
sherpa/estmethods/__init__.py:440: undefined name '_est_funcs'
```

@olaurino @DougBurke - How do you want to proceed here? Merge this PR as-is and then you continue fixing the issues pyflakes reports and adding regression tests? Or give me pointers how to fix these issues and what kinds of tests to add?

If you're considering adding a travis-ci build that runs static code analysis, in my experience [prospector](https://pypi.python.org/pypi/prospector) and `pyflakes` are easy and quite good, but `pyflakes` finds more issues and is better configurable on what to report / ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 # C extensions
 *.so
+*.so.dSYM
 
 # Build outputs
 *.o
--- a/sherpa/optmethods/opt.py
+++ b/sherpa/optmethods/opt.py
@@ -79,7 +79,7 @@ class Opt( object ):
         xmin = numpy.asarray( xmin, numpy.float_ )  # Make a copy
         xmax = numpy.asarray( xmax, numpy.float_  ) # Make a copy
         if ( x.shape != xmin.shape ) or ( x.shape != xmax.shape ):
-            raise InputError( x, xmin, xmax )
+            raise InputErr( x, xmin, xmax )
         if _outside_limits( x, xmin, xmax ):
             raise OutOfBoundErr( x, xmin, xmax )
         return x, xmin, xmax


dtnguyen2:
opt.py is not used by sherpa, we are in the process of removing the file

cdeil:
@dtnguyen2 - Ah, OK.

What about the error I mentioned in `sherpa/estmethods/__init__.py`?
```
$ pyflakes-2.7 sherpa/estmethods/ | grep 'undefined name'
sherpa/estmethods/__init__.py:316: undefined name '_est_funcs'
sherpa/estmethods/__init__.py:440: undefined name '_est_funcs'
```
https://github.com/sherpa/sherpa/blob/master/sherpa/estmethods/__init__.py#L316
https://github.com/sherpa/sherpa/blob/master/sherpa/estmethods/__init__.py#L440

This doesn't work, because the import is
```
import sherpa.estmethods._est_funcs
```
and not
```
from sherpa.estmethods import _est_funcs
```
right?

Is this also old code that will be removed, or should that be fixed?


PS: For Astropy and most affiliated packages we have the coverage report online, and generated for all pull requests:
https://github.com/astropy/astropy#project-status
https://coveralls.io/github/astropy/astropy
I find this very useful, it makes it quick to check which code is covered without having to produce a coverage report locally.
Add this for Sherpa?
(It's easy to set up via travis-ci, let me know if you want this and if I should make a pull request adding that to one of the builds.)

DougBurke:
@cdeil - we are looking at doing something with the coverage reports, but haven't done so yet. Given that we have a bunch of "optional" packages, so the coverage result depends on the environment, I assume it's easy to say "only report the results from this particular build" (or some other manner)?

cdeil:
@DougBurke - Right, what Astropy and most other packages do is to pick one build where all dependencies are available and generate the main coverage report for that one.

You can see how it's done technically here:
https://github.com/astropy/astropy/blob/master/.travis.yml#L106

The [coveralls](https://pypi.python.org/pypi/coveralls) command line tool is used to submit the report to the free https://coveralls.io/ service, which shows nice coverage reports that one can link to, e.g. to point out missing test coverage to contributors, or just to browse which parts of the code aren't covered yet.

olaurino:
My experience with coveralls is that it does more harm than good. It may work better with Python than with Java, but with our Java projects I disabled it: even when robust tools like sonarqube (and common sense) detected coverage increase, coveralls for some reason signaled a decrease, staining perfectly fine PRs. I think it's good practice to run coverage analysis locally on small units during development, especially to check that one is testing what they think they are testing, so I never found coveralls particularly useful and I don't miss it. Plus, one can obtain 100% line coverage and have meaningless tests anyway. That being said, we may review the decision we made about coveralls in the future.

cdeil:
@olaurino - I agree with your caveats / concerns concerning coverage and coveralls, but in the end I do think it's a bit useful to have. For Gammapy I set up coveralls, but I turned off the automatic commenting on pull requests, because I also found the "coverage has increased or decreased by 0.XX %" not helpful.
The way I use it is:
1. Sometimes check overall coverage and browse which code isn't covered. Having it online can be useful to point out some non-tested function or method to other contributors and ask them to add a test for this (that's usually a good "getting started" or "good first pull request")
2. Sometimes for PRs that add a lot of new code, link to the coverage report and point out the main chunks of new code that aren't covered.

PS: locally for Sherpa I get a coverage of 45% :
https://gist.github.com/cdeil/062380f016b7ebdc8e59#file-gistfile1-txt-L417

Is it normal / useful that the coverage report includes `test_*` files?
Am I running the tests correctly or am I doing something wrong?
Having a "reference" coverage report for master online for Sherpa where all dependencies are present and the tests are run correctly would be useful IMO.

olaurino:
> Is it normal / useful that the coverage report includes test_* files?

They say it's normal (it's the default behavior), but I have to say at first I was surprised too. The reasoning seems to be that you may want to know if some of your tests are not running, or in any case you don't have full coverage of your tests.

I have locally a `pytest-cov` file that disables that, with the following content:

````
[run]
omit = *test_*
branch = True
````

> Am I running the tests correctly or am I doing something wrong?

I don't think you are doing anything wrong. We know that the coverage of tests accompanying code is not enough, but reasons for that go back in time and there is no simple/rapid fix. We are trying to add new coverage whenever possible, even if it's not in the specific scope of a PR (e.g. we added tests for several stats, not just for `wstat`, when we included `wstat`, and I am adding plenty of tests in the scope of the otherwise simple #22 fix). There are tests that run in CIAO, but that do not accompany the code. Some of them were "ported" several years go, some were not. The full code coverage (line and branch combined) is of about 60% when all the dependencies are present, including ds9 and xspec.

I am planning to add plenty of unit tests in the scope of the python3 migration.

> Having a "reference" coverage report for master online for Sherpa where all dependencies are present and the tests are run correctly would be useful IMO.

The decision was made before we published the code to GitHub (we used to have the configuration commented out in the travis config), but we can discuss that again. In any case thanks for your input!

olaurino:
I went back to this PR, and I noticed that we got somehow carried away of the main topic.

Given that `opt.py` is going away, it remains the problem of `_est_funcs`. Maybe I am forgetting something basic about Python that I should know (and tracking PEPs about imports is hard indeed), but the code actually works, despite the fact that I thought it shouldn't. And it is indeed covered by our tests.

Once 'sherpa.estmethods._est_funcs` is imported, `_est_funcs` appears in the globals of `estmethods.__init__.py`. I wonder if this works by accident because of the implementation or because it is somehow part of the spec.

DougBurke:
Also, do we have a PR yet for the removal of `opt.py`?

olaurino:
@cdeil I am going to create a new PR getting rid of `opt.py` and keeping your change for `_est_funcs`: for this latter change, even if it is working and we have coverage I guess it's safer to get your change.

olaurino:
Replaced by #192.

Issue #133, opened 2015-12-16
Add install option for pure-Python Sherpa core?


cdeil:
Is it possible to add an option to the Sherpa build system to install a minimal pure-Python core part of Sherpa?

For [gammapy](https://gammapy.readthedocs.org/en/latest/), [threeml](https://threeml.stanford.edu/) and [naima](http://naima.readthedocs.org/en/latest/) we are debating (have been for a year) whether we can go all-in on Sherpa modeling or whether it has to remain an optional dependency, because it's hard to install of the Fortran / C extensions.
So e.g. currently in Naima there's [models](https://github.com/zblz/naima/blob/master/naima/models.py) and [sherpa wrapper models](https://github.com/zblz/naima/blob/master/naima/sherpa_models.py).

I know the eval function of most of the Sherpa built-in models is written in C extensions, but the `Parameter` and `ArithmeticModel` classes, i.e. the core model classes, are pure-python.

Would it be possible / hard / easy to add an option to the Sherpa build system and make something like `pip install sherpa[core]` work and have the core model classes available, but not have any C / Fortran?

olaurino:
@cdeil, thanks for the suggestion, we'll discuss this.

I have some questions:
  * Did you have issues installing the fortran/c extensions? One of the goals we had when redesigning the install process was to make it easy to install sherpa with a default, streamlined, python-friendly process. If there are still issues to be ironed out it would be good to know. Maybe we need to better document the source build process and options?
  * Would conda binaries be an option for you? Could you take advantage from more frequent binary builds from the master branch?

Thanks!

cdeil:
> we'll discuss this.

Thanks!
As I said, it was just a thought.
If it's a lot of work to implement, it might not be worth the effort because at least for Gammapy, it'll probably remain so that to do likelihood fitting, i.e. most users, we'll rely on Sherpa optimizers and error estimators, i.e. the C extensions.

> Did you have issues installing the fortran/c extensions?

Yes ... I currently have `/opt/local/bin/gfortran-mp-5` as Fortran compiler and I don't know how to tell the Sherpa build to pick it up. I admit I haven't tried to find out ... maybe it is documented how to pass this info to the Sherpa build?

> Would conda binaries be an option for you? Could you take advantage from more frequent binary builds from the master branch?

Yes, conda is a great option.
More frequent binary builds would be helpful, I think.
E.g. we could start using your `wstat` and remove our implementation of wstat in Gammapy now.

[...]

Issue #76, opened 2015-08-11
Support Python 3


cdeil:
Feature request: please port Sherpa to support Python 3.

I know this is a big project that will take time and resources.

But I wanted to file the issue to make it clear that this is important to us ([Gammapy](https://github.com/gammapy/gammapy), used to analyse data from [HESS](https://www.mpi-hd.mpg.de/hfm/HESS/), [CTA](https://portal.cta-observatory.org/Pages/Home.aspx) and possibly other telescopes).
Our motivation is that we want to build Gammapy on Sherpa and it's the only package in our stack that doesn't support Python 3 (see our dependency stack [here](https://indico.cern.ch/event/344485/session/142/contribution/695/attachments/1136524/1626432/gammapy-icrc2015-poster.pdf)). If Gammapy doesn't work on Python 3 it can't in good faith be put forward as a possible analysis package for new projects like CTA. E.g. the [ctapipe](https://cta-observatory.github.io/ctapipe/) package that's being proposed as the official low-level CTA processing pipeline (doesn't use Sherpa) will be Python 3 only. So getting there within the next ~ year is important to us.

I'm willing to help with this in the coming months, but when it comes to the Fortran / C extensions, my skills and familiarity with the Sherpa code base (same for the other Gammapy developers as far as I know) are probably not sufficient to make this happen.

anetasie:
Supporting Python 3 is in the plans and the team is evaluating the required work. At this moment there is no schedule for the update.

cdeil:
That's great to hear, thank you!
:tada:

matteobachetti:
+1 For this

olaurino:
We added this issue to the 4.8.2 milestone as we are targeting this version for adding Python 3 support. Some work in this sense is currently tracked in PR #229.

PR #169, opened 2016-02-01
First tiny step towards Python 3 support

cdeil:
Work in progress ... do not merge!

This PR is a first tiny step towards Python 3 support (see #76).

As I've mentioned at https://github.com/sherpa/sherpa/pull/160#issuecomment-178194122 and https://github.com/sherpa/sherpa/issues/133#issuecomment-175887835, I think it could be useful to try to start the port to Python 3 with small steps. The changes in this PR have no effect on Python 2, but they do allow running `python3 setup.py egg_info` with Python 3, i.e. a Python 3 test build for `python setup.py egg_info` could be added here on travis-ci and then future PRs would try to expand Python 3 support in steps.

@olaurino - Do you think getting started like on Python 3 porting is at all useful?

I have some time to work on this in the coming weeks, but I don't know if this is a good way or time to get started, and if you prefer many small PRs that are easy to review, or fewer large PRs that e.g. change all imports in the codebase to implicit relative imports, touching many files.

### DIFF
--- a/helpers/__init__.py
+++ b/helpers/__init__.py
@@ -18,13 +18,13 @@
 #
 
 
-from clean import clean
-from develop import develop
-from install import install
-from sdist import sdist
-from test import PyTest
-from sherpa_config import sherpa_config
-from xspec_config import xspec_config
+from .clean import clean
+from .develop import develop
+from .install import install
+from .sdist import sdist
+from .test import PyTest
+from .sherpa_config import sherpa_config
+from .xspec_config import xspec_config
 
 commands = {
     'clean': clean,
--- a/helpers/clean.py
+++ b/helpers/clean.py
@@ -19,7 +19,7 @@
 
 
 from distutils.command.clean import clean as _clean
-from deps import clean_deps
+from .deps import clean_deps
 
 class clean(_clean):
     def run(self):
--- a/helpers/deps/__init__.py
+++ b/helpers/deps/__init__.py
@@ -38,7 +38,7 @@ def build_deps(configure):
     if not os.path.exists('extern/built'):
         prefix=os.getcwd()
         os.chdir('extern')
-        os.chmod(configure[0], 0755)
+        os.chmod(configure[0], 0o755)
         env = os.environ.copy()
         env['PYTHON'] = sys.executable
         out = call(configure, env=env)
--- a/helpers/install.py
+++ b/helpers/install.py
@@ -21,7 +21,7 @@
 from numpy.distutils.command.install import install as _install
 import os
 
-from deps import build_deps
+from .deps import build_deps
 
 class install(_install):
     def run(self):
--- a/helpers/sdist.py
+++ b/helpers/sdist.py
@@ -20,7 +20,7 @@
 
 from distutils.command.sdist import sdist as _sdist
 from numpy.distutils.misc_util import get_data_files
-from deps import clean_deps
+from .deps import clean_deps
 
 class sdist(_sdist):
 
--- a/helpers/sherpa_config.py
+++ b/helpers/sherpa_config.py
@@ -18,8 +18,8 @@
 #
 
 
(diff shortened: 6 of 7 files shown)


DougBurke:
@cdeil I know that work on Python 3 support is planned for this year. I
don't know if we have come up with an approach yet, but @olaurino can say
more about that.

olaurino:
> Do you think getting started like on Python 3 porting is at all useful?

@cdeil of course I do! Thanks a lot for starting it!

I would have probably started by adding a Python 3 travis build, expecting it to fail pretty badly, but also counting how many tests pass, watching the number raising at each iteration. Of course, without the first changes you published, tests wouldn't even run, so your start is just as good as mine ;)

As I mentioned in a different comment, my plan was also to add unit tests to limit the scope of the python 3 migration on a unit-by-unit base while also improving coverage, using Sherpa 4.8.0 (under python2) as the oracle (the source of "truth" about the test results) for Python 3 runs.

I am happy we all work on this branch, but since I was not planning to start the Python 3 migration just yet (other than the "prototype" I have locally for estimating the amount of work required), you are free to beat me and complete it. ;)

I did find some unexpected issues in the pure python migration "prototype" (e.g. multiprocessing-related stuff does not work anymore under Python 3). My local solution was good enough for a prototype, but not for production. If you run into such strange issues, please let me know before you spend any time on it, I might already know the cause and we can work a solution together.

I have a branch in my own fork with outdated code from my "prototype". Please feel free to take a look, but with a grain of salt as this is just an attempt to get as many tests as possible to pass in the least amount of time, and there may be errors. Also, the C/C++ changes are not python 2 compatible:
https://github.com/olaurino/sherpa/tree/python3

cdeil:
OK, thanks for the comments.

I'll add a Python 3 travis-ci build and probably some more fixes here tomorrow, but I'll try to keep this first PR simple and easy to review / not touching too many lines.

Do you want to test / target Python 3.4 and / or 3.5?

I think with a 2.7 and 3 compatible code base, everything that works in 3.4 should work in 3.5 and the other way around. So the only cost of supporting 3.4 as well is one extra travis-ci build and it's worth adding that extra build (just to be on the safe side, that it's really working in both 3.4 and 3.5)?

cdeil:
Oh wait ... I now see you already have a ton of good changes here:
https://github.com/sherpa/sherpa/compare/master...olaurino:python3
But there's merge conflicts...

Do you want to rebase your `python3` branch and continue there?
Or is it OK to leave your branch as a prototype that will never be merged, and re-do the changes in PRs against current master?

olaurino:
I took a lot of shortcuts in that branch, I would just keep it there for reference.

Are you familiar with tools like futurize? http://python-future.org/automatic_conversion.html

it's not 100% accurate, but it gives you a head start.

cdeil:
I've used this in the past and it was pretty good: https://pypi.python.org/pypi/modernize

@olaurino Would it be OK to use and bundle six in the Sherpa repo?
It's a single `six.py` file with a few 100 lines and used by a ton of projects, including Astropy.
If yes, do you prefer bundling the file or adding it as a dependency?
(I'm +1 to bundle, just to avoid having to explain the extra dependency to users.)

olaurino:
There are several reasons why I think bundling `six.py` would be a good idea, but I used `future` rather than `six` and `futurize` rather than `modernize` in the past, so I'll need to review the differences. In the meantime, please feel free to add `six.py` to this branch. The pure python stuff does not concern me too much anyway, except for those issues that no tools can fix and we'll need to work out ourselves anyway.

olaurino:
@cdeil I started building on your first commit (thanks again!) a couple of weeks ago, maybe more, and I'll start tracking that work in a new PR with a local branch and I'll close this one. Please feel free to contribute to that branch.

olaurino:
Closing in favor of #229.

PR #173, opened 2016-02-04
Template tests: fix missing test and PEP8 changes

DougBurke:
# Release Notes

The template test suite had two methods with the same name; they have now been renamed to ensure that they are both run. There was also come code cleanup of the tests: improved logic and PEP8 changes.

### DIFF
--- a/sherpa/models/tests/test_template.py
+++ b/sherpa/models/tests/test_template.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2011, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -18,7 +18,6 @@
 #
 
 
-import unittest
 from sherpa.models import TableModel, Gauss1D
 from sherpa.models.template import create_template_model
 from sherpa.utils import SherpaTest, SherpaTestCase, requires_data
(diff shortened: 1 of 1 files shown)


sherpa/models/tests/test_template.py
```diff
@@ -64,35 +63,44 @@ def test_restore_is_discrete(self):
...
             return
-        self.fail('Fit should have failed: using gridsearch with wrong parvals')
+        self.fail('Fit should have failed: gridsearch with wrong parvals')
 
-    # TestCase 3.2 discrete templates fail when probed for values they do not represent (continuous method with discrete template)
-    def test_load_template_model_without_interpolation(self):
+    # TestCase 3.2 discrete templates fail when probed for values they do
+    # not represent (continuous method with discrete template)
+    #
+    def test_load_template_model_without_interpolation_32(self):
```
olaurino: Good catch, I wonder why this wasn't caught as it was in other cases.

sherpa/models/tests/test_template.py
```diff
@@ -64,35 +63,44 @@ def test_restore_is_discrete(self):
...
-            self.assertEqualWithinTol(2023.46, ui.get_fit_results().parvals[1], 0.001)
+        pvals = ui.get_fit_results().parvals
+        pmin = pvals[0]
+        pmax = pvals[1]
+        if pmax < pmin:
+            (pmin, pmax) = (pmax, pmin)
+
+        tol = 0.001
+        self.assertEqualWithinTol(2023.46, pmin, tol)
+        self.assertEqualWithinTol(2743.47, pmax, tol)
```
olaurino: I would contend that, aside maybe adding a couple of lines to assign the observed values before the comparison, the original code was less complex, but I don't care too much, so that's ok.

sherpa/models/tests/test_template.py
```diff
@@ -134,16 +142,15 @@ def tearDown(self):
...
         self.model(self.x)
 
 #    def test_template_query_index(self):
 #        expected = 5
 #        result = self.model.query_index([0,1,0,1])
-#        self.assertEqual(expected, result,
-#                         "Expected %s instead of %s" % (str(expected), str(result)))
-
+#        self.assertEqual(expected, result)
+#
```
olaurino: it would have helped, at the time, adding a couple of comments as to why these tests were commented out. Have you tried running them?

olaurino:
I am OK with the changes, let's merge it. Thanks!

sherpa/models/tests/test_template.py
DougBurke: I haven't tried running these!

PR #162, opened 2016-01-26
load_table_model updates

DougBurke:
# Status

I have closed this PR since the initial changes have made it in through other PRs and its gone back to development stage (in part because I don't have the time to work on it). I do expect these changes to make it back to a PR at some point in the future.

Not finished. PR is to get Travis-CI feedback on the changes.

# Release Notes

Improve documentation of `sherpa.ui.load_table_model`.

Improve documentation of the I/O related routines in `sherpa.ui` (remove mention of the Sherpa I/O module as this is not relevant here).

Enhance the test suite for `sherpa.ui.load_table_model` and apply PEP8 fixes. 

# Notes

I am planning to fix #143 but need to improve the current documentation and testing to make sure nothing breaks. 

**EDIT** I am breaking out some of the changes into separate PRs so that these improvements are not held up by a discussion of how to organize the tests (and improve them): #173 #174 #175

PR #175, opened 2016-02-04
Improve documentation on ASCII file support and minor bug fix

DougBurke:
Release Notes
-------------

Improve the documentation for file I/O in `sherpa.ui`, including

* be explicit about ASCII support only (including remove mention of FITS support)
* add documentation for the support routines (e.g. those in `sherpa/io.py`) that actually do the I/O

Improve documentation for `load_table_model`, in particular the file support.

Fix unknown error classes in `sherpa.ui.utils`: these are in "hard to exercise" code paths (since there have been no reports about the unknown symbols).

There have been PEP8-related code changes.

Notes
-----

Ideally the fix for the unknown error classes would be in a separate PR but I don't have the time to rework this.

### DIFF
--- a/sherpa/io.py
+++ b/sherpa/io.py
@@ -126,6 +126,99 @@ def get_column_data(*args):
 
 def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
                    comment='#', require_floats=True):
+    """Read in columns from an ASCII file.
+
+    Parameters
+    ----------
+    filename : str
+       The name of the ASCII file to read in.
+    ncols : int, optional
+       The number of columns to read in (the first ``ncols`` columns
+       in the file). This is ignored if ``colkeys`` is given.
+    colkeys : array of str, optional
+       An array of the column name to read in. The default is
+       ``None``.
+    sep : str, optional
+       The separator character. The default is ``' '``.
+    dstype : data class to use, optional
+       Used to check that the data file contains enough columns.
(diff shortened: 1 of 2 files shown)


olaurino:
It looks good to me. I'd say merge it.

However, should we make it a task (maybe even part of the python 3 migration) that we unify `sherpa.astro.ui` and `sherpa.ui` to reduce a lot of code duplication and functionality overlap?

DougBurke:
@olaurino - there's a lot of places we could look at for review (e.g. I'm looking at `sherpa.astro.io` and `sherpa.io` in #162)! The `ui` modules do have a lot of overloaded code with subtly (and not-so-subtly) different semantics, and I think they are too large (they've been a pain when I was looking at sphinx). Perhaps it's worth noting things like this down on the wiki, rather than in issues (I did start with a page on model evaluation which I need to get back to finishing off), so they don't get lost and can perhaps provide more of an overview of what we'd like to do?

olaurino:
Let's make it an item for the next meeting, then. I can see pros and cons of using the wiki and tracking issues. Consolidation of `sherpa.astro.[ui|io|??]` and `sherpa.[ui|io|??]` would be a big task anyway.

olaurino:
And it goes without saying, any time we spend fixing these issues is time not spent implementing science features. Which is why making it part of the python 3 migration might be a good idea: a smaller code base can be ported more easily than a larger code base.

Issue #177, opened 2016-02-05
numpy 1.10.4 + mkl on anaconda leads to test failure


DougBurke:
With the master branch - b1e905e5fdbd1579d0fe8757584619fed27fde43 - I have just seen the following test failure (from manually running `sherpa/utils/tests/test_utils.py`, but you also get it from `python setup.py test`).

```
======================================================================
FAIL: test_parallel_map (sherpa.utils.tests.test_utils.test_utils)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/djburke/sherpa/sherpa-test/sherpa/utils/tests/test_utils.py", line 214, in test_parallel_map
    self.assert_((numpy.asarray(result) == numpy.asarray(pararesult)).all())
AssertionError: False is not true
```

This is in a new conda environment - I noticed it updated numpy, so maybe that's the cause.

```
% conda list numpy
# packages in environment at /home/djburke/miniconda2/envs/sherpa-test:
#
numpy                     1.10.4                   py27_0  
```

DougBurke:
If I downgrade to `numpy 1.10.2` then the tests pass, so it looks like something in either `numpy 1.10.4`, or the associated `mkl` package, is causing the problem:

```
conda update numpy
Fetching package metadata: ......
Solving package specifications: ................
Package plan for installation in environment /home/djburke/miniconda2/envs/sherpa-test2:

The following NEW packages will be INSTALLED:

    mkl:   11.3.1-0     

The following packages will be UPDATED:

    numpy: 1.10.2-py27_0 --> 1.10.4-py27_0

Proceed ([y]/n)? y

Unlinking packages ...
[      COMPLETE      ]|###################################################| 100%
Linking packages ...
[      COMPLETE      ]|###################################################| 100%
```

DougBurke:
It appears to be a tolerance issue - changing to use `numpy.testing.assert_allclose` seems to fix this problem. We should really review this particular test as to see if it is a sensible choice.

DougBurke:
I have two different approaches to work around this issue:

* #178 which compares the arrays using a tolerance rather than using equality 

* #180 change the `.travis.yml` file to force numpy to be '< 1.10.4'

My suggestion would be for #180, as a very short-term fix, then we can decide if #178 is a sensible fix or if the test should be changed.

olaurino:
There is a third option, which is probably the one I like the best for the time being, assuming it works: opt out of `mkl` and keep installing the latest 1.10 version available.

[...]

PR #178, opened 2016-02-05
Fix bug #177 - tests failing with numpy-mkl 1.10.4

DougBurke:
Release Notes
-------------

Update the tests to support numpy 1.10.4 with the Intel Math Kernel Library (mkl), as provided by anaconda.

Notes
------

This updates several tests to check for numeric equivalence using a
tolerance rather than absolute equality. There have also been PEP8
changes, and updates to use more-recent unittest methods (e.g.
assertTrue rather than assert_ and assertGreater).

Several tests should probably be re-written to use vectorized routines
rather than iterating over elements, but that is outside the scope of
this fix.

The problem test could be re-written to use a different numerical
operation, which could remove the need for a tolerance check, but I
felt that was also out-of-scope here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,10 @@ matrix:
       python: "2.7"
       sudo: required
       dist: trusty
+    - env: XSPECVER="12.9.0i" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 MKL=nomkl
+      python: "2.7"
+      sudo: required
+      dist: trusty
     - env: XSPECVER="12.9.0i" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5
       python: "2.7"
       sudo: required
@@ -45,7 +49,6 @@ matrix:
 env:  FITS="astropy" INSTALL_TYPE=install TEST=smoke MATPLOTLIBVER=1.5
 
 before_install:
-  - export MKL="nomkl"
   - export LIBGFORTRANVER="1.0"
   - export SHERPA_CHANNEL=https://conda.anaconda.org/sherpa
   - export MINICONDA=/home/travis/miniconda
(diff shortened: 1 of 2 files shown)


DougBurke:
It's unclear to me, looking at the travis logs, what version of numpy they are using (but at least it means the tests still pass with one version of numpy!),

DougBurke:
It looks like the Travis CI tests are now pulling in `numpy 1.10.4` - e.g. see this failure at https://travis-ci.org/sherpa/sherpa/jobs/107481018#L4481 [there are other, expected, failures in this particular test, but not this particular one].

olaurino:
This PR should probably include changes to `.travis.yaml` that test both `mkl` and `nomkl` numpy. @DougBurke I can work on that if you agree.

DougBurke:
I have limited Sherpa time, so please add in any changes you think necessary @olaurino .

olaurino:
Rebasing. Original tip was at 46ef55a.

olaurino:
If travis is happy I would like to merge this sooner rather than later. @DougBurke I wouldn't wait for ciaod. If the tests passed there, there is no reasons why they shouldn't pass now that they are more lenient, right?

DougBurke:
I am happy for this to be merged now. If the CIAOD tests fail then there's really something strange going on!

olaurino:
Rebasing to include #193. Original tip was at  872fe66.

olaurino:
I looked more closely at the test failures #177 refers to.

It is likely we are just dealing with floating point errors involved in the calculations, especially considering that the values in the test are determined randomly.

However, I have a couple of comments regarding the test and the `parallel_map` implementation itself:
  * A try, except block makes the test pass in some cases without doing anything. Actually if there is only 1 cpu that doesn't even happen. Probably this is a leftover of some old version of Python that did not even have the `multiprocessing` module? This should be fixed (as noted by some of @DougBurke comments).
  * Is the choice of the function and data in the test adequate? Why not using something simpler and deterministic? (again, as noted by @DougBurke in his comments).
  * Why doesn't `parallel_map` use the simple approach used in the test, i.e. `multiprocessing.Pool.map` in the first place?

The first two points can be addressed in this PR, the third probably not.

DougBurke:
I agree that the `try/except` around the multiprocessing imports should be removed.

I doubt there are any notes as to why this particular function was tried (perhaps it is used somewhere in the Sherpa internals?); I don't have time to come up with something simpler, and I fear that any long delays with this PR are going to cause problems delaying other PRs.

olaurino:
@DougBurke I was assuming I was going to make the changes. As long as you can give me your blessing after I push a new commit I am fine :)

olaurino:
> I doubt there are any notes as to why this particular function was tried (perhaps it is used somewhere in the Sherpa internals?);

I found that `eigvalsh` (not `eigvals`) is used in some modules, but not as an argument to `parallel_map`, so I suspect this is not why that specific function was used. Given the nature of this test (a unit test for `parallel_map`) anything as simple as `sum` may be used, I guess.

olaurino:
@DougBurke I just pushed 0087d1d. Travis workers run on two cores according to the docs, so the tests should be meaningful. The current `parallel_map` implementation has several `if` clauses with no corresponding tests, but I know I will have to change that implementation for the Python 3 migration, so I'll try and take care of the tests in that context.

DougBurke:
I'm a bit concerned that the test doesn't check for ordering of results - i.e. each task is exactly the same, so if there's a change in ordering then we wouldn't catch it in this test. I don't think we explicitly document the ordering requirement of parallel_map, but given as we do not return any information to determine it, I think it's fair to say that we expect the output order to match the sequence order of the input. So maybe change the iterable from

    vals = np.asarray([1, 2, 3])
    iterable = [vals] * numtasks

to something like

    iterable = [numpy.arange(1, 2+2*i) for i in range(numtasks)]

so that the sums are different, and hence we check the ordering.

olaurino:
That's an excellent point. I'll make the change you suggested.

DougBurke:
Assuming Travis is okay, looks good to me.

PR #100, opened 2015-09-10
Fix set_source output from save_all

DougBurke:
# Release Notes

This fixes a bug whereby the `save_all` function would not create the necessary `set_source()` line. This does not fix all cases, but it does the simple PHA case such as

    load_pha('src.pi')
    set_source(xsphabs.gal * powlaw1d.pl)
    save_all('test.out')

This is bug #97.

It also ensures that files created by `save_all` can be run using IPython's `%run` directive, by explicitly importing the `sherpa.astro.ui` module.

The `sherpa.astro.ui.save_session` function has been removed: this was never documented and was never meant to be released as it did not work correctly. This cleanup also removes the `sherpa.ui.utils.loggable` class.

# Details

This is an *extremely* simplified version of PR #98 which only addresses one problem, and does not add a test suite. It is however very easy to review.

# Test

With the simple test case

```
% cat test.py
from sherpa.astro import ui

ui.load_data(1, 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi')

ui.notice(0.5, 6)

ui.set_source(ui.powlaw1d.pl)

pl.gamma = 0.7
pl.ampl = 1e-4

print("statistic = {}".format(ui.get_stat_name()))
print("          = {}".format(ui.calc_stat()))
```

then with the `master` branch

```
In [2]: %run test.py
/home/naridge/local/anaconda/envs/sherpa-save-all-simple/lib/python2.7/site-packages/IPython/kernel/__init__.py:13: ShimWarning: The `IPython.kernel` package has been deprecated. You should import from ipykernel or jupyter_client instead.
  "You should import from ipykernel or jupyter_client instead.", ShimWarning)
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi' 
but not used; to use them, re-read with use_errors=True
read ARF file sherpa-test-data/sherpatest/threads/pha_intro/3c273.arf
read RMF file sherpa-test-data/sherpatest/threads/pha_intro/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273_bg.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273_bg.pi' 
[...]

PR #179, opened 2016-02-11
Remove the TraceCalls class.

DougBurke:
The sherpa.utils.TraceCalls class is not used in Sherpa and so has
been removed. Was it part of the loggable prototype work that was
removed in PR #100? See changeset 4ba7b8985d8bc3ec2f86fa15c5c0785959501222

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
These tests are failing because of #177.

olaurino:
The `TraceCalls` class was created on 2013-05-02 by @dtnguyen2, according to ClearCase.

It seems to have a goal similar to the `loggable` prototype we were working and removed, but it does not seem to be used anywhere, at least in Sherpa. Maybe it's used somewhere else in CIAO or Level 3? I am ok with removing it as long as @dtnguyen2 is.

dtnguyen2:
it was used for lev 3, but all calls to it has been removed so go ahead

PR #180, opened 2016-02-11
Avoid numpy 1.10.4 in Travis builds (work around for #177)

DougBurke:
This is an attempt to avoid numpy 1.10.4 in the Travis builds to avoid
triggering #177. It is not a sensible long-term solution.


olaurino:
I think #181 is a better stop-gap measure. I agree that in any case we need something like #178 for the long run.

DougBurke:
Closing in favor of #181

PR #181, opened 2016-02-11
opt out of mkl

olaurino:
Workaround #177.

This PR opts out of the new default `mkl` feature in conda packages for `numpy` and related packages. This works around #177, but it does not fix it. A proper fix would make sure that tests pass both with and without `mkl`. In this sense #178 (plus probably adding at least one more travis build) is the way to go eventually.

With respect to #180, this PR does not pins the `numpy` version: it just makes sure `nomkl` is selected as a feature in the conda environment.

See https://www.continuum.io/blog/developer-blog/anaconda-25-release-now-mkl-optimizations


DougBurke:
This is obviously not a good solution long term (neither was #180), but I'd like to see something merged so that the tests can go back to passing.

PR #185, opened 2016-02-25
Protect XPA command to avoid shell confusion.

DougBurke:
# Release Notes

Fix the problem where if the working directory contained a file called `x` or `y` then the `sherpa.astro.ui.image_data()` function would fail with the message

     DS9Err: Could not display image

A test has been added to check this case, as well as a general clean up to slightly improve
the testing. The code has been updated for PEP8 fixes.

# Details

This fix is to avoid `sherpa.astro.ui.image_data()` from failing when
the working directory contains a file called `x` or `y` (and possibly
other combinations). It is a limited fix; ideally the code should be
updated to take better advantage of the `subprocess` module (e.g.
using a sequence rather than a string for the arguments, so that
there is no need to "protect" arguments).

Is there any reason why the XPA communication can not be done with https://github.com/ericmandel/pyds9 ?

### DIFF
--- a/sherpa/image/DS9.py
+++ b/sherpa/image/DS9.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2006-2010  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2006-2010, 2016  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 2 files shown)


olaurino:
I am happy to merge this branch.

PR #187, opened 2016-03-02
Fix #92 - issue a more meaningful warning when FITS backends are not available

### LINKED ISSUE #92: Give a more helpful error message when no FITS backend can be imported
See comment https://github.com/sherpa/sherpa/pull/91#issuecomment-135457766

When no FITS backend is installed, currently the code just prints the bare ImportError from Python, but that's misleading since users can install both astropy and pyfits.

I am setting this to the 4.8 milestone as it does not impact CIAO users.

olaurino:
I don't think this is worth a release note anymore, but just in case
# Release Note
Fix #92: a more meaningful message is given to the user when `sherpa.astro.io` is imported directly and no fits backends are available.

# Notes
Some tests would fail if no FITS backends were present. This PR also fixes this situation by adding the proper `@requires_fits` decorator to tests that were missing it.

# Testing
I would be more comfortable using `mock` and/or `pytest` for writing a test that ensure that the warning is issued. On the other hand, since tests should run just fine when no backends are installed, we may also want to create a new travis build. This is probably overkill in this case, but I am happy to do it if people feels this should be done. Note that in order to be meaningful, it should be a full build including xspec.

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -50,7 +50,14 @@ if io_opt.startswith('pycrates') or io_opt.startswith('crates'):
 elif io_opt.startswith('pyfits'):
     io_opt = 'pyfits_backend'
 
-backend = __import__(io_opt, globals(), locals(), [])
+try:
+    backend = __import__(io_opt, globals(), locals(), [])
+except ImportError:
+    raise ImportError("""Cannot import selected FITS I/O backend {}.
+    If you are using CIAO, this is most likely an error and you should contact the CIAO helpdesk.
+    If you are using Standalone Sherpa, please install astropy (pyfits should also work, but it's deprecated)."""
+                      .format(io_opt))
+
 warning = logging.getLogger(__name__).warning
 info    = logging.getLogger(__name__).info
 
--- a/sherpa/astro/ui/tests/test_serialize.py
+++ b/sherpa/astro/ui/tests/test_serialize.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


DougBurke:
A few nit picks:

* should the docstring of `sherpa/astro/io/__init__.py` be updated to reflect the new functionality

* it's not obvious to me why `sherpa/astro/io/__init__.py` exports routines like `read_arrays` - which explicitly calls the backend - when no backend is available. Should the logic that only adds in `read_table_blocks` to `__all__` be extended for the other routines, or am I missing something here?

* the copyright dates need updating to include 2016

olaurino:
> should the docstring of sherpa/astro/io/__init__.py be updated to reflect the new functionality

Done (f0f2f88).

> it's not obvious to me why sherpa/astro/io/__init__.py exports routines like read_arrays - which explicitly calls the backend - when no backend is available. Should the logic that only adds in read_table_blocks to __all__ be extended for the other routines, or am I missing something here?

I hit "enter" too soon when creating the pull request :)

I was still not sure what would make more sense: reraise the exception with a better message? Try and load the module with a warning but without exposing any functions? I am still not sure. What do you think?

> the copyright dates need updating to include 2016

Done (f0f2f88).

olaurino:
@DougBurke I quickly created f371bce. What do you think? The advantage of this approach is that it just issues a warning rather than failing if there is no backend. I am not really sure there is any advantage in this. Maybe I should just reraise the exception with a better message.

olaurino:
Rebasing, original tip was at f371bce.

olaurino:
@DougBurke this is ready to review (assuming the build is green according to travis)

olaurino:
The errors in the Travis build do not have anything to do with the PR contents, but they seem to be originating from libgfortran issues. I found several reports of similar issues, and I am looking into them.

olaurino:
For some reason libgfortran=1.0 now needs to be explicitly installed through conda. I couldn't reproduce the issue locally, but this fixes the travis builds. I'll use this fix for now, but then I'll need to understand whether this has an impact on the way the conda binaries for sherpa need to be created.

DougBurke:
I think the libgfortran issues are related to the use of the `nomkl` flag. When I just tried a new conda env using `nomkl` I saw that libgfortran was installed and I got test failures from `python setup.py test` like

```
>   import _minpack
E   ImportError: /home/djburke/miniconda2/envs/sherpa-test/lib/python2.7/site-packages/numpy/core/../../../.././libgfortran.so.3: version `GFORTRAN_1.4' not found (required by /home/djburke/sherpa/sherpa-test/sherpa/optmethods/_minpack.so)
```
If I try in an environment that uses the mkl-enabled numpy then I don't see these errors (although then may get the #177 error)

olaurino:
That's interesting, I had tried that but I couldn't reproduce the issue. Thanks!

olaurino:
Rebasing to solve conflicts. Original tip was at d4a04ba.

olaurino:
@DougBurke could you please confirm we can label this to be merged (pending CIAOD testing)?

DougBurke:
Looks good to me.

PR #188, opened 2016-03-03
93 - Try to install test deps with sherpa_test, or provide info

### LINKED ISSUE #93: Some issues with sherpa_test
I recently built a dev cut of the conda binary (for supporting our gammapy colleagues), and I found some minor issues with sherpa_test while testing the package.

Since this does not impact CIAO users I am setting this ticket to the 4.8 release milestone.

Here are the issues:
  * if pytest is not installed sherpa_test suggests that you run `pip install -r test_requirements.txt`, which is fine when you are working with the source code, but not when you are working with binaries. The suggestion should be improved for dealing with both cases.
  * sherpa-test-data does not install the datasets when called with `python setup.py install`, only when using `pip install .` This should be either fixed or documented.
  * `sherpa_test -d` (for overriding the discovery of the test data folder) did not seem to work. I did not have time for looking into it, as there is a workaround (install sherpa-test-data as a python package), but I need to double-check.

### LINKED ISSUE #167: How to select / run Sherpa tests during development?
I'd finally like to start contributing to Sherpa by taking a stab at #150.

Would it be possible to extend the info on how to hack on Sherpa, maybe here?
https://github.com/sherpa/sherpa/blob/master/CONTRIBUTING.md#checklist-for-contributed-code

In Astropy my workflow is that I edit files in the git repo and then run tests repeatedly like this:
```
python setup.py test [options for test selection, reporting]
```
I think this runs this in the background:
```
python setup.py build_ext -i
```

I have a setup with Macports where this is working for me with Sherpa:
```
python setup.py install --user
cd /tmp # leave sherpa repo so that the installed version is picked up
sherpa_test # run tests from installed version
cd - # go back to Sherpa repo
```

- Is it possible to run the tests from the git repo, without having to cd away?
- How can I select which tests to run?

(Apologies if I've asked this before! I don't remember ...)

This is not working for me:
```
$ python2.7 setup.py test
running sherpa_config
warning: sherpa_config: built configure string['./configure', '--prefix=/Users/deil/code/sherpa/build', '--with-pic', '--enable-standalone', '--disable-maintainer-mode', '--enable-stuberrorlib', '--disable-shared', '--enable-shared=libgrp,stklib', '--enable-fftw', '--enable-region', '--enable-group', '--enable-stk', '--enable-wcs']

running xspec_config
running develop
error: can't create or remove files in install directory

The following error occurred while trying to add or remove files in the
installation directory:

    [Errno 13] Permission denied: '/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/test-easy-install-60417.write-test'

The installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages

Perhaps your account does not have write access to this directory?  If the
installation directory is a system-owned directory, you may need to sign in
as the administrator or "root" account.  If you do not have administrative
access to this machine, you may wish to choose a different installation
directory, preferably one that is listed in your PYTHONPATH environment
variable.

For information on other options, you may wish to consult the
documentation at:

  https://pythonhosted.org/setuptools/easy_install.html

Please make the appropriate changes for your system and try again.
```

olaurino:
# Release Note
Fix #93 The `sherpa_test` script now tries to install the test dependencies before running the tests (but not the `sherpatest` package, which should be installed by the user if necessary, due to its footprint). If this is not possible, and the necessary dependencies (`pytest`) are not found, then a meaningful message is given to the user with instructions on how to install the dependencies.

# Description
The idea behind the `sherpa_test` console script was to provide a single, portable test command that could be run the same way whether Sherpa was being installed from source or through the conda command. It was not supposed to be necessarily equivalent to running "python setup.py test", but similar in scope and easy of use. However, there were some quirks by which this was not true, and for instance the command would fail with a meaningless message if one ran it after installing the conda binary (`please run pip install -r test_requirements.txt` does not make sense in that context).

This PR addresses several issues in the following way:
  * `pytest` is now added as a dependency for the conda package. This way `sherpa_test` should always work with conda binaries.
  * `pytest-cov` is not required by the `pytest.ini` default config, and coverage reports are not provided by default anymore. This is required because conda does not have a pytest-cov package, and it would be too unstable to try and install it from the console script (I tried). So, `pytest-cov` is also not listed as a dependency anymore. Travis builds will log the coverage report on the console. Developers/users willing to generate the coverage report will need to install `pytest-cov` themselves and pass the coverage configuration they want to `pytest`, if any. [I found myself changing the `pytest-cov` configuration all the time depending on what format I wanted/needed, and the long terminal coverage report generated by default was just annoying when debugging, so to me this new setup is also an improvement.]

# Note
This PR also fixes #167 by adding some documentation to the README.


olaurino:
See https://github.com/sherpa/sherpa/pull/187#issuecomment-192305438 for the reasons of the travis build failures.

sherpa/__init__.py
```diff
@@ -116,11 +116,25 @@ def test(level=1, verbosity=1, datadir=None):
...
-        sys.exit(1)
+    def install(package_name):
+        try:
+            import pip
+            pip.main(['install', package_name])
+        except:
+            print """Cannot import pip or install packages with it.
+            You need pytest, and possibly pytest-cov, in order to run the tests.
+            If you downloaded the source code, please run 'pip install -r test_requirements.txt' first.
+            """
```
DougBurke: Should it say "please run 'pip install ...' from the source directory first." ?

I don't know when `clitest` can be called - I assume it is called by `sherpa_test` - and I just feel it's worth being explicit about the location of the `test_requirements.txt` file. If `clitest` can only ever be (sensibly) called from the directory containing `setup.py` (and hence `test_requirements.txt`) then it's not really an issue.

sherpa/__init__.py
```diff
@@ -116,11 +116,25 @@ def test(level=1, verbosity=1, datadir=None):
...
-    try:
-        import pytest, pytest_cov
-        errno = pytest.main(['--pyargs', 'sherpa', '-rs'])
-        sys.exit(errno)
-    except ImportError:
-        print """Cannot import pytest and pytest-cov.
-            Please run 'pip install -r test_requirements.txt' first"""
-        sys.exit(1)
+    def install(package_name):
+        try:
```
DougBurke: Would it be worth having separate try/except blocks for `import pip` and `pip.main` so that a user can identify what failed? Or is there going to be enough screen output that you can work it out?

This is not a strong suggestion.

.travis.yml
```diff
@@ -52,11 +54,11 @@ before_install:
   - ./miniconda.sh -b -p $MINICONDA
   - export PATH=$MINICONDA/bin:$PATH
   - conda update --yes conda
-  - if [ -n "${FITS}" ]; then conda install --yes -c ${SHERPA_CHANNEL} ${FITS}; fi
+  - if [ -n "${FITS}" ]; then conda install --yes -c ${SHERPA_CHANNEL} ${FITS} $MKL; fi
   - if [ -n "${MATPLOTLIBVER}" ]; then MATPLOTLIB="matplotlib=${MATPLOTLIBVER}"; fi
   - if [ -n "${NUMPYVER}" ]; then NUMPY="numpy=${NUMPYVER}"; fi
   - echo ${MATPLOTLIB} ${NUMPY} ${FITS}
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION pip ${MATPLOTLIB} ${NUMPY} nomkl
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION pip ${MATPLOTLIB} ${NUMPY} $MKL libgfortran=$LIBGFORTRANVER
```
DougBurke: I thikn it would be safer - in the sense of much-less-likely to end up with a mis-matched set of libraries - if we could just have a single `conda install` line with all the requirements listed on it. This way conda can resolve all the dependencies cleanly. I have occasionally seen issues when updating/downgrading dependencies (in particular if they need different numpy ABIs). I think it would also have the advantage of making it a bit easier to trace what is/isn't being installed in a particular run.

The other approach to take would be to say that if a problem does exist we can quickly update the this file, and it's not worth the engineering time to make this "clean".

sherpa/__init__.py
olaurino: That's a good idea.

clitest is bound to the sherpa_test console script in setup.py... it is called when sherpa_test is called, so there is no guarantee it's called from the right folder.

sherpa/__init__.py
olaurino: all exceptions are caught and then re-raised after the message has been printed, so the actual exception is displayed to the user.

.travis.yml
olaurino: You are right. The problem is that pyfits is only available from our channel, and I don't want to risk that other packages are taken from our channel other than pyfits. I didn't want the conditions to be too complex, but it is true that bad things can happen if the fits backend is astropy, which is the case we care the most. I'll change the script to install everything on the same line.

olaurino:
Rebasing to avoid conflicts, previous tip was at a388c16.

olaurino:
@DougBurke do you think we can merge this (pending CIAOD testing?)

DougBurke:
Do we want to address my first comment, namely https://github.com/sherpa/sherpa/pull/188#discussion_r55047255  (I think you answered my second one)?

olaurino:
Yes, we do.

DougBurke:
Changes look fine with me (assuming various tests pass ;-)

PR #189, opened 2016-03-04
Fix crates dm syntax issue

olaurino:
This PR fixes a test that was failing in CIAO due to some DM syntax introduced by the crates backend. It should not impact the standalone tests.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -18,6 +18,7 @@
 #
 
 import os
+import re
 import unittest
 import tempfile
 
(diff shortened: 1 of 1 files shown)


sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -428,7 +429,11 @@ def save_arrays(self):
         rtol = 0
         atol = 1e-5
         self.assertIsInstance(out, Data1D)
-        self.assertEqual(out.name, ofh.name, msg="file name")
+
+        # remove potential dm syntax introduced by backend before checking for equality
+        out_name = re.sub("\[.*\]", "", out.name)
```
DougBurke: I think this should be anchored to the end of the string - e.g. `"\[.*\]$"` - but is it worth it? I guess first I should check that `re` supports this!

olaurino: I had the same thought right after I pushed the commit, and then I thought it wasn't worth being changed because if the backend didn't happen to append the dm syntax then other parts of the test would fail.

DougBurke: I'm happy leaving the code as-is, as there's a comment explaining why the line is needed.

olaurino:
I set the priority to blocker. This PR needs to be merged sooner rather than later so we can go back to building CIAOX from the master branch. I'll make sure it gets merged into the snapshot branch so that it can be picked up by CIAOD. Once CIAOD passes we can merge this PR.

olaurino:
Rebasing to include #193. Original tip was at d152bfb.

olaurino:
I checked that the CIAOD tests pass on OSX and Linux64. I'll merge this.

PR #190, opened 2016-03-04
fix #22 - datastack can be used if no plotter available

### LINKED ISSUE #22: datastack module and optional requirements

olaurino:
# Release Note
Fix #22 - The datastack package can now be used even if there is no available plotting backend. In this case, plotting functions will not be available, but the rest of the datastack functionality will.

# Description
The main goal of this PR is to fix #22. However, it achieves other goals as well. In particular:
  * the package has been refactored to be more testable and modular, encapsulating each backend extension into its own module.
  * a new `sherpa.utils.logging` provide the utility function `config_logger` that allows to easily config a custom logger. The code is rather similar to the original function defined in the datastack package, but it can now be used by other components.
  * a new `public` decorator allows to dynamically define the `__all__` variable for a module.
  * datastack wrapper functions now properly override the documentation of the wrapped function.

# Testing
This PR was supposed to be accompanied by proper tests. However, these tests use the `mock` library and some advanced feature of `pytest` in order to simulate the precence or absence of the optional dependencies backing up sherpa back-ends. Since these new OTS have not yet been approved for inclusion in CIAO, I am leaving the tests out. I did run the tests locally, and they are included in commit 5f35850. The commit is available in the repository, and will be run with by travis to prove that the tests are successful, at least in standalone mode.


olaurino:
Rebasing to include travis build fixes. Original tip was at c363e76.

olaurino:
Rebasing to resolve conflicts. Original tip was at 7eeba31.

PR #88, opened 2015-08-24
Add CONTRIBUTING.md and CITATION documents

DougBurke:
Adds a `CONTRIBUTING.md` file - based on that of AstroPy - which will be seen when users submit a Pulll Request on GitHUb.

The contents of the CITATION document are taken from the "Citing Sherpa in a Publication" section of http://cxc.harvard.edu/sherpa/

PR #83, opened 2015-08-18
Continue S-Lang removal

DougBurke:
# Release Note
Removed S-Lang scripts, files, and references in the code.

# Description
I forgot to include this in PR #69.
[OL] Rebased and included Release Note section.

PR #192, opened 2016-03-04
Small fixes

olaurino:
This PR replaces #168: it fixes few pep8 issues and removes an obsolete file.


sherpa/utils/__init__.py
```diff
@@ -1633,7 +1642,7 @@ def run_tasks(procs, err_q, out_q, num):
     # return list(numpy.concatenate(results))
     # Remove extra dimension added by split
     vals = []
-    [ vals.extend(result) for result in results ]
+    [vals.extend(result) for result in results]
```
DougBurke: I've been tempted a few times to rewrite this line as it is not particularly clear/pythonic!

olaurino:
Rebasing to get travis build fixes, original tip was at ea1d638.

olaurino:
@DougBurke could you please confirm we can merge this?

sherpa/utils/__init__.py
```diff
@@ -119,6 +120,7 @@
 
 
 class NoNewAttributesAfterInit(object):
+
```
DougBurke: Is there a style guide that suggests a blank line here? I don't think it should hold back the PR; just wondering whether we want to be consistent with docstrings on classes (which is a larger issue than this PR).

DougBurke:
Looks good to me (the comment about the blank lines I've just made is not something that should stop this PR from being merged).

sherpa/utils/__init__.py
olaurino: https://github.com/hhatto/autopep8/issues/194

> As referenced in issues #83, #88, and #178, an earlier version of PEP 257 made the weird recommendation of inserting a blank line before the docstring of a class.
> 
> However, this recommendation has since been removed. The current PEP 257 only recommends adding a blank line after the docstring, not before:
> 
> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.
> (The commit that changed it: 9b715d8246db)
> 
> As far as i understand, autopep8's E309 fixer does not currently distinguish between docstrings and other class members when inserting blank lines, so this issue cannot be worked around with --ignore=E309 alone, if you still want autopep8 to insert blank lines before non-docstring members.

olaurino: Not sure what to do, maybe I should just replace the pep8 commit with a different one.

DougBurke: If you were to do anything, I'd suggest just adding a commit that removes these lines (as it seems to me to be a safer option). However, I am not sure it needs to be done in this PR (unless pep8/pyflakes/whatever) now complains about these lines: I only noticed them from the git diff view.

olaurino:
@DougBurke we can merge it now, right?

DougBurke:
Looks good to me.

PR #193, opened 2016-03-08
use matplotlib 1.4 for pyfits build

olaurino:
This PR fixes a conflict in the `pyfits` travis build. The `pyfits` conda binary pins the `numpy` dependency to 1.8. A conflict arise later on when trying to install `matplotlib` 1.5.

A number of PRs are currently failing for this reason. PRs should be rebase on top of master when this is merged, which should happen as soon as the Travis builds are green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       python: "2.7"
       sudo: required
       dist: trusty
-    - env: FITS="pyfits" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5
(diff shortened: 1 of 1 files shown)


olaurino:
It looks like the travis builds using the google containers (for us, the ones that build xspec) are timing out when installing packages with `apt-get`. I haven't seen any reports for this just yet. I'll open a ticket if I still can't find one.

olaurino:
They fixed the issue that was preventing the GCE builds to use APT. I restarted the PR builds.

@DougBurke I would like to merge it as soon as Travis is fine with it. Is that ok?

DougBurke:
@olaurino - I have no time for sherpa at the moment so you'll have to make
that call.

PR #194, opened 2016-03-17
Feature/minuit

dtnguyen2:
Make iminuit/minuit, optionally (ie users are responsible for typing 'pip install iminuit'), available for the masses.

### DIFF
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2015, 2016  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -29,6 +29,12 @@ warning = logging.getLogger(__name__).warning
 
 __all__ = ('GridSearch', 'OptMethod', 'LevMar', 'MonCar', 'NelderMead')
 
+try:
+    import iminuit
+    __all__ = optfcts.add2__all__(__all__, 'Minuit')
+except ImportError:
+    pass
+
 
 class OptMethod(NoNewAttributesAfterInit):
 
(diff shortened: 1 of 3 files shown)


sherpa/optmethods/__init__.py
```diff
@@ -29,6 +29,12 @@
 
 __all__ = ('GridSearch', 'OptMethod', 'LevMar', 'MonCar', 'NelderMead')
 
+try:
+    import iminuit
+    __all__ = optfcts.add2__all__(__all__, 'Minuit')
```
DougBurke: You should be able to remove `optfcts.add2__all__` and just say

    __all__ += ('Minuit', )

and similarly in ` sherpa/optmethods/optfcts.py `

sherpa/optmethods/tests/test_optmethods.py
```diff
@@ -39,13 +38,22 @@ def tst_all( self, name, fct, fmin, x0, xmin, xmax,
...
-##         self.tst( stogo.stogo, name + '_stogo', fct, fmin,
-##                   x0, xmin, xmax, iprint=iprint )
         self.tst( optfcts.montecarlo,  name + self.mc , fct, fmin,
                   x0, xmin, xmax, iprint=iprint, maxfev=8192*len(x0) )
         self.tst( optfcts.lmdif, name + self.lm, fct, fmin,
                   x0, xmin, xmax, iprint=iprint )
 
+        #
+        # If the following lines are uncommented out, we get errors
+        # when running the optimization test
```
DougBurke: Are the errors because the iminuit test itself fails, or is it a more general error (e.g. unable to import the module or some other error)?

sherpa/optmethods/tests/test_optmethods.py
```diff
@@ -39,13 +38,22 @@ def tst_all( self, name, fct, fmin, x0, xmin, xmax,
...
+        #
+        # try:
+        #     import iminuit
+        #     self.tst(optfcts.midnight, name + '_minuit', fct, fmin,
+        #              x0, xmin, xmax, iprint=iprint)
+        # except ImportError:
+        #     pass
+
```
DougBurke: Rather than protecting the test in a try/except block, we should use the `unittest.skipIf` decorator (or one of the wrappers around such as `sherpa.utils.requires_package`), since then we have better tracking of the total number of tests. @olaurino can (hopefully) provide info on the preferred syntax here.

dtnguyen2: Made the __all__ += changes, but get the following error when I tried to
push:

Password for 'https://dtnguyen2@github.com':
To https://github.com/sherpa/sherpa
 ! [rejected]        feature/minuit -> feature/minuit (non-fast-forward)
error: failed to push some refs to 'https://github.com/sherpa/sherpa'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.

What's a fella to do?

sherpa/optmethods/tests/test_optmethods.py
dtnguyen2: minuit failed to solve a few (can't remember now how many but if I have to
guess it was 6, I can check if inquiring minds want to know) problems that
the other methods can

sherpa/optmethods/tests/test_optmethods.py
DougBurke: Talk to @olaurino ;-)

sherpa/optmethods/tests/test_optmethods.py
DougBurke: Ahhh. That's a more-serious issue than I was hoping for. Do you think it's okay for these particular tests to fail with this optimiser, or should they pass (or do we need to tweak tolerances)?

DougBurke:
We also need to add a note somewhere in the documentation about this optional support - presumably `README.md`.

sherpa/optmethods/tests/test_optmethods.py
dtnguyen2: An optimization method, no matter how well written, is not expected to pass
all tests.  There are probably some tweaks that one can do but this was a
low effort on my part and I did not want to investigate it.

DougBurke: Thanks for the info. I haven't looked at the optimiser tests so wasn't sure what the range of acceptable behaviour was.

Issue #147, opened 2016-01-15
calc_stat_info broken with wstat


joleroi:
I would like to use WStat for gammapy.  The latest dev version I find with Anaconda is ``4.7+524.g2d013d1``. It does not have Wstat, though.

Would it be possible to release a new dev version?

DougBurke:
We are working through setting up an anaconda version of the `release/4.8.0` branch. The current plan is to have this done by the end of January.

olaurino:
@joleroi if you want to try the release candidate (I can't guarantee it will be the final cut) you can use the binaries in the test channel:
conda install -c http://conda.anaconda.org/sherpa/channel/test sherpa

joleroi:
Thank you very much! 

No the following problem appeared. I am trying to run the PHA intro (http://cxc.harvard.edu/sherpa/threads/pha_intro/) using WStat with the following script.

```
import sherpa.astro.ui as sau

sau.load_pha("3c273.pi")
sau.set_source(sau.powlaw1d.p1)
sau.guess(p1)
sau.set_stat("wstat")
sau.fit()
stats = sau.get_stat_info()
```
Which gives this error
```
Traceback (most recent call last):
  File "fit_wstat.py", line 8, in <module>
    stats = sau.get_stat_info()
  File "<string>", line 1, in get_stat_info
  File "/home/kingj/Software/miniconda3/envs/sherpadev/lib/python2.7/site-packages/sherpa/ui/utils.py", line 7549, in get_stat_info
    return self._get_stat_info()
  File "/home/kingj/Software/miniconda3/envs/sherpadev/lib/python2.7/site-packages/sherpa/astro/ui/utils.py", line 9411, in _get_stat_info
    statinfo = f.calc_stat_info()
  File "/home/kingj/Software/miniconda3/envs/sherpadev/lib/python2.7/site-packages/sherpa/fit.py", line 1058, in calc_stat_info
    stat, fvec = self.stat.calc_stat(dep, model, staterror, syserror)
  File "/home/kingj/Software/miniconda3/envs/sherpadev/lib/python2.7/site-packages/sherpa/stats/__init__.py", line 594, in calc_stat
    raise StatErr('usecstat')
sherpa.utils.err.StatErr: No background data has been supplied. Use cstat
```

DougBurke:
[...]

PR #195, opened 2016-03-17
Generalize calc_stat API + example of how to have several datasets with different fit statistics

dtnguyen2:
When I was asked to pass the bkg data to the UserStat, I made a mistake by

making the code backwards compatible.  Then when I had to add WStat to Sherpa,
things went from bad to very bad.  After all, WStat needs the exposure time and
backscale ratio.  Not to mention the size of the data sets if one is to use it
for fitting simultaneous data sets (different data sets could have different
exposure time and/or backscale ratios).

The old classes for the Stats are not very easily extensible; Moreover, The
changes should also enable the user to do simultaneous fit using different
statistics for different data set.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -27,8 +27,7 @@ from numpy import arange, array, abs, iterable, sqrt, where, \
     ones_like, isnan, isinf, float, float32, finfo, nan, any, zeros, append, \
     int, ones
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
-    bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp, \
-    get_valid_args
+    bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import FitErr, EstErr, SherpaErr
 from sherpa.data import DataSimulFit
 from sherpa.estmethods import Covariance, EstNewMin
@@ -460,7 +459,7 @@ class IterFit(NoNewAttributesAfterInit):
         # Data set attributes needed to store fitting values between
         # calls to fit
         self._dep = None
-        self.bkg_data = None
+        self.extra_args = None
         self._staterror = None
         self._syserror = None
         self._nfev = 0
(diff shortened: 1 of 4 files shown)


dtnguyen2:
Rebasing to get travis fixes. Original tip was at ab96227.

DougBurke:
From a quick look this looks like a good change, in that it is more extensible. In fact, this part is probably more important than the example showing how to use multiple statistics in a fit (at least to my mind).

@cdeil - does gammapy make use of user statistics yet (I don't think the changes here should break any existing code, but it's worth checking)?

DougBurke:
@olaurino now that this has been rebased (actually done a while ago) can the pr:hold label be removed? Or are there still conflicts with other PRs?

olaurino:
Hmm... right now there are conflicts with master, according to GitHub. In any case as I mentioned CIAOD is not being built with the snapshot branch, so pr:hold doesn't really make any difference. Hopefully we will go back to the normal CIAO routine as the python 3 changes settle down.

DougBurke:
It may well conflict with #253 although that should be an easy fix.

olaurino:
Rebasing and resolving conflicts. Original tip was at 84fde99.

I also changed the commit message to adhere to our standards: I simply added a short line at the biginning of the message and left the rest as the extended description.

DougBurke:
The test failure here is - I assume - fixed by #263 ?

olaurino:
@DougBurke yes. I'll rebase as soon as I can.

DougBurke:
I don't understand why the test `test_wstat_calc_stat_info` - which was introduced as part of the fix for #147 - fails due to the code added here. Looking at the fix in #263 for this, I would have expected the code to fail when #147 was addressed (i.e. why does this test not fail in #253 ?)

olaurino:
@DougBurke there is of course a chance I made a mistake while rebasing, but that seemed straightforward.

olaurino:
Rebasing, original tip: 7fe126e

olaurino:
Just to confirm, @DougBurke, you reviewed this PR and it's ready to be merged (as long as Travis is happy, *ça va sans dire*). Right?

DougBurke:
Short answer: I am happy for this to be merged into master if travis passes.

Longer answer: So, if my thoughts of a rewrite in https://github.com/sherpa/sherpa/pull/269#issuecomment-245900117 are sensible, then I think the `calc_stat` method should be rewritten (or an additional method added that uses `Data` objects instead of arrays). However, that's in the (possible) future and doesn't stop this from being merged (and there are extra tests added too, which helps in any possible rewrite).

PR #196, opened 2016-03-17
DO NOT MERGE - Tests datastack backend support

olaurino:
The sense of this PR is just to have Travis create a pr-build with the tests for #190. We are not including the tests in that PR because they rely on mock and pytest, which are not yet approved as CIAO OTS packages. This PR proves that #190 actually works when merged into master.


olaurino:
Rebasing after having rebased #190.

sherpa/astro/datastack/__init__.py
```diff
@@ -1086,40 +312,56 @@ def wrap(*args, **kwargs):
         return wrapfunc(*args, **kwargs)
 
     wrap.__name__ = func.__name__
-    wrap.__doc__ = func.__doc__
+    if not hasattr(DataStack, func.__name__):
```
DougBurke: Can the logic be swapped here, as to me it reads  better - i.e.

    if hasattr(...):
        doc = getattr(...)
    else:
        doc = func.__doc__

DougBurke:
Would it make sense to pull out the logging changes into a separate PR (just to make it a bit easier to see what's going on), or do they rely on the mock changes so it makes sense to do them here?

It's not obvious to me how the logging changes made here - e.g. the addition of https://github.com/sherpa/sherpa/blob/tests-datastack-backend-support/sherpa/utils/logging.py - interact with other parts of the system - e.g. should https://github.com/sherpa/sherpa/blob/tests-datastack-backend-support/sherpa/__init__.py#L50 be updated to take advantage of this, or is it unrelated? I'm particularly sensitive to this since the logging output is used in the sphinx branch I'm working on, and it would be nice to have all the logging code centralized (for instance, in my branch I keep the `Formatter` from `sherpa/__init__py` around rtaher than deleting it as i found it useful to have).

DougBurke:
@olaurino - switching to the mock-related code. I haven't used mock  before, so can you provide some high-level docs for why you are using it: is it to allow both chips and matplotlib interfaces to be tested when either neither or only one package is selected, or are you mocking out the plotting api completely (i.e. you use the mock interface even when matplotlib is installed), or something else?

olaurino:
I have had "write documentation for pytest and mock in Sherpa" in my todo list for a long time, so your point is well taken. Long story short:

> is it to allow both chips and matplotlib interfaces to be tested when either neither or only one package is selected

Yes. In the context of optional backends, this allows to unit-test our interface to the backends even if they are not installed, checking that the backends, which have their own public API, are correctly called.

More generally, you mock/stub collaborators of the classes/functions you are developing so that you can test them in isolation: either because they are not available yet (but then you also have to do some sort of duck typing, there are no interfaces in Python), or because you don't want/need the extra complexity introduced by the collaborator: if I have a piece of code that is calling the optimizer and taking the results back to do some extra processing, I don't need the optimizer to actually run. I can simply mock it, ensure it is properly called, and fake a result that is good enough to test the rest of the functionality.

> are you mocking out the plotting api completely (i.e. you use the mock interface even when matplotlib is installed),

Yes, sometimes. In an ideal world (but no so ideal, this is really something we can and should do) you develop a unit and focus on its specific functionality: you unit-test it by mocking/stubbing everything else and you focus on getting it right. Then you write functional/integration tests by running the same (or very similar) tests with the actual dependencies, to make sure units work together as expected.

I plan to make extensive use of mock when working on the python 3 migration, because that should allow me to test individual units without having to migrate all of their dependencies first.

olaurino:
By the way, I think I learned better and cleaner ways to mock dependencies.

Compare how things are done here in PR #210 to unit-test that the smoke tests exits with an error code if XSPEC is not present:
https://github.com/sherpa/sherpa/pull/210/files#diff-261c8e9aad659715d105b9a4fbb4959dR47

With this much messier code:
https://github.com/sherpa/sherpa/pull/196/files#diff-034a373736ee7489197f3289d41e3cedR49

DougBurke:
So does this mean we should re-implement this PR to follow the approach used in #210 (in other words close this PR), or do you want to merge this and then change it?

olaurino:
I would consider this refactoring. We already have tests for the functionality, so it should be safe to rearrange them in a cleaner way. It shouldn't take long, but the priority is probably low.

DougBurke:
@olaurino so are you saying you are still interested in merging this PR, and then once it is merged refactor it to match the approach taken in #210 ?

olaurino:
Sorry, I should have been more clear: I meant that I would rather refactor it before merging it.

DougBurke:
Saves me some work :+1:

olaurino:
I don't think it makes sense to keep this PR open, as we finally included pytest and changed the way we write new tests. I rebased this branch, fixed a couple of python 3 issues, and removed a couple of dubious tests I am not even sure how they had worked to begin with. I pushed the new branch to my fork. I'll see what to do with it after I get the results from Travis, but in the meantime I'm closing this PR.

PR #202, opened 2016-03-31
DO NOTE MERGE - tweak g95 configuration to work for CIAO's OSXM builds

olaurino:
# Description

This change allows to tweak the g95 compiler options so that Sherpa can be built in CIAO with `g95` i.s.o. `gfortran`.

For now this is a test. I don't like the way paths and library names are hard coded. If the test works I'll try and rework the PR so that the options can be configured from the command line or by tweaking the `setup.cfg` file.

This seems also related to #5, as options could be provided to support compilers other than `gfortran`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,9 @@ except:
         "Could not import setuptools.\n"
         "This might lead to an incomplete installation\n"
     )
+
+import platform
+
 from numpy.distutils.core import setup
 
 from helpers.extensions import static_ext_modules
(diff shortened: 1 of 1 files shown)


DougBurke:
This change, as is, is not going to be applied to the repository. It is being kept as a reminder that we need to:

1. document this approach (for current users), e.g. perhaps as part of #5 

2. determine whether there is a need to make this easier, perhaps by adding command-line flags or settings in the `cfg` file

PR #206, opened 2016-04-13
Update the release documentation

DougBurke:
Re-order and re-word the installation instructions - in particular the test section - and adjust the version numbers given in the XSPEC installation section, as well as point out the known problems with the "build using CIAO binaries" approach.

Several other minor clean ups, including changing the alignment to better match an 80-column width, for those people reading directly rather than a rendered version.

Ideally the binary section would split out the "how to use a virtual environment" from the miniconda installation section (as it is relevant for both `1a` and `1b`) but I did not want to try that now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,14 @@
 Sherpa
 ======
 
-Sherpa is a modeling and fitting application for Python. It contains a powerful language for combining simple models
-into complex expressions that can be fit to the data using a variety of statistics and optimization methods.
-It is easily extensible to include user models, statistics, and optimization methods.
-It provides a high-level User Interface for interactive data-analysis work,
-such as within an IPython notebook, and it can also be used as a library
-component, providing fitting and modeling capabilities to an application.
+Sherpa is a modeling and fitting application for Python. It contains a
+powerful language for combining simple models into complex expressions
+that can be fit to the data using a variety of statistics and
+optimization methods.  It is easily extensible to include user models,
+statistics, and optimization methods.  It provides a high-level User
+Interface for interactive data-analysis work, such as within an
+IPython notebook, and it can also be used as a library component,
+providing fitting and modeling capabilities to an application.
 
 What can you do with Sherpa?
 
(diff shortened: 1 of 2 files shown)


olaurino:
Does it make sense to issue this PR against the release branch? We won't repackage the binaries, but we would need to rerun the download tests and source build tests if we let GitHub produce a new tarball associated with the release.

README.md
```diff
@@ -6,14 +6,19 @@
 
 - [Sherpa](#sherpa)
 - [How To Install Sherpa](#how-to-install-sherpa)
-  - [Binary installation](#binary-installation)
+  - [Binary installation using Anaconda](#binary-installation)
```
jbudynk: the link is broken. I think it should be `#binary-installation-using-anaconda`.

README.md
```diff
@@ -6,14 +6,19 @@
...
     - [1a. Anaconda](#1a-anaconda)
     - [1b. Starting from scratch](#1b-starting-from-scratch)
+    - [1c. Other packages](#1c-other-packages)
   - [Source Build](#source-build)
     - [2a. Extract the source tarball](#2a-extract-the-source-tarball)
     - [2b. Get the code from the GitHub repository](#2b-get-the-code-from-the-github-repository)
     - [2c. Build Sherpa](#2c-build-sherpa)
     - [2d. Testing the build](#2d-testing-the-build)
+    - [2e. Development mode](#2e-development-mode)
+  - [Testing Sherpa](#test)
```
jbudynk: this link should also be updated. Should be `#testing-sherpa`

jbudynk:
I'm happy to see  "Other Packages" in its own section, as I'd forget to install those when creating a new sherpa environment. Besides the minor broken TOC links, looks good.

DougBurke:
This commit should fix the two broken links found by @jbudynk - ta.

DougBurke:
Will re-submit this against master (#207).

DougBurke:
Actually, submitting this against `master` will have to wait for `branch/4.8.1` to be merged.

PR #207, opened 2016-04-15
Update the release documentation

DougBurke:
[This is a resubmit of #206 but against the master branch, now that the 4.8.1 release branch has been merged]

Re-order and re-word the installation instructions - in particular the test section - and adjust the version numbers given in the XSPEC installation section, as well as point out the known problems with the "build using CIAO binaries" approach.

Several other minor clean ups, including changing the alignment to better match an 80-column width, for those people reading directly rather than a rendered version.

Ideally the binary section would split out the "how to use a virtual environment" from the miniconda installation section (as it is relevant for both 1a and 1b) but I did not want to try that now.

PR #208, opened 2016-04-15
update Zenodo DOI

olaurino:
# Description

This PR adds a Zenodo DOI badge for 4.8.0 previous release. The DOI badge that was already present automatically points to the latest release. Probably because of caching issues the images look the same, but I checked that they point to different pages.

I am not really sure this is a good idea, so I didn't play too much with the formatting and location. Maybe instead of adding a badge we should list the DOIs in the CITATION file?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,11 @@
 [![Build Status](https://travis-ci.org/sherpa/sherpa.svg?branch=master)](https://travis-ci.org/sherpa/sherpa)
 [![DOI](https://zenodo.org/badge/683/sherpa/sherpa.svg)](https://zenodo.org/badge/latestdoi/683/sherpa/sherpa)
 
-**Previous Releases DOIs**
-
-4.8.0: [![DOI](https://zenodo.org/badge/doi/10.5281/zenodo.45243.svg)](http://dx.doi.org/10.5281/zenodo.45243)
-
-
-
-
 <!-- TOC *generated with [DocToc](https://github.com/thlorenz/doctoc)* -->
 **Table of Contents**
 
 - [Sherpa](#sherpa)
+- [License](#license)
 - [How To Install Sherpa](#how-to-install-sherpa)
   - [Binary installation](#binary-installation)
     - [1a. Anaconda](#1a-anaconda)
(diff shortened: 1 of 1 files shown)


DougBurke:
Is Travis having a bad day? The build errors here and for #207 - e.g. 

  https://travis-ci.org/sherpa/sherpa/jobs/123376234 
  https://travis-ci.org/sherpa/sherpa/jobs/123369206

look more like system issues than with us (plus both these commits are doc-only changes).

olaurino:
There was a glitch in the `apt` support that even failed one of the builds of the master branch. I'll restart the builds.

DougBurke:
I was wondering about how the Zenodo links dealt with different release versions. I think this is fine for now, and we can tweak the layout if necessary at a later time (I would not want to do a larger change here as it is likely to make integration with #207 more work than it needs).

DougBurke:
Given we already have a History section, perhaps the new section could be made a subsection of that. Having looked at the page as it would be rendered on GitHub I think it's a bit distracting where it is now, but I don't have a really strong opinion about it.

DougBurke:
@olaurino just to note that I'm waiting for #207 to be accepted before giving the go-ahead for this.

DougBurke:
I've moved some of the information around: the previous DOI to the end of the file as part of the existing history section, and then moved out the license information hidden in the history section into its own section.

I like the idea of including the zenodo reference in the `CITATION` document, but I think it can wait for another PR:

 - I want to talk to ADS/AAS/library folks about how to cite Zenodo
 - it's not clear to me how to provide a "latest" link to Zenodo, which we'd need (since we don't know the Zenodo URI until after the release has happened, we can't add it to the `CITATION` document, unless there's a way to "reserve" the URI)

DougBurke:
The test failures are not to do with this patch: the messages look something like

```
Fetching package metadata: ....Error: HTTPError: 503 Server Error: Service Unavailable: Back-end server is at capacity for url: https://conda.anaconda.org/sherpa/noarch/repodata.json.bz2: https://conda.anaconda.org/sherpa/noarch/
```

I'll wait for a bit and then re-run them.

DougBurke:
I've restarted the checks (and continued doing so until they passed).

@olaurino if the edits I've made fine look good to you than can you label this pr:merge (or say it's okay and I can do so)?

olaurino:
Looks good to me. Thanks.

PR #209, opened 2016-04-20
Fix docstrings for group_snr() and group_adapt_snr().

jbudynk:
# Release Note

Docstrings used to say "Combine the data so that each bin has a signal-to-noise
ratio of at least `minimum`" but the end of the groups are marked only if the
SNR *exceeds* the given SNR.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2008, 2015, 2016 Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


DougBurke:
@jbudynk sorry for taking so long to look at this. I think the parameter descriptions should be updated to make this clear: e.g. lines 1032 and 1147

jbudynk:
No problem. Nice catch! I'll update those as well.

jbudynk:
I've updated the `snr` parameter descriptions for the two grouping methods.

DougBurke:
Thanks for the update. One final change - which I forgot to check for - is to update the copyright date to include `, 2016` .

jbudynk:
Ah, right. Updated the copyright date. I amended the commit since it's a minor change (was 781c667).

DougBurke:
@jbudynk - thanks. Once travis gives the green light I'll label these as pr:merge, but I need to check the current status of our internal CIAO build with @olaurino before merging.

jbudynk:
Sounds good. Thanks for the review.

Issue #123, opened 2015-10-21
Files opened with the wrong mode (r, w, etc.) in sherpa regression test


jbudynk:
A warning message is shown if someone tries to open a read-only files. Running the sherpa regression tests with the data directory specified in a CIAO environment will give a warning:

```
% source /proj/xena/ciaod/Linux64/ciao_1021/bin/ciao.csh
CIAO configuration is complete... 
CIAO 4.8 Wednesday, October 21, 2015
  bindir      : /proj/xena/ciaod/Linux64/ciao_1021/bin
  CALDB       : 4.6.9
% python -c "import sherpa; sherpa.test(datadir='/data/regression_test/master/in/sherpa')"
Found 7/7 tests for sherpa.astro.tests.test_datastack
Found 1/1 tests for sherpa.astro.models.tests.test_models
...
................................./proj/xena/ciaod/Linux64/ciao_1021/lib/python2.7/site-packages/pycrates/io/dm_backend.py:140: UserWarning: File '/data/regression_test/master/in/sherpa/aref_Cedge.fits' does not have write permission.  Changing to read-only mode.
  warnings.warn("File '" + fname + "' does not have write permission.  Changing to read-only mode.")
/proj/xena/ciaod/Linux64/ciao_1021/lib/python2.7/site-packages/pycrates/io/dm_backend.py:140: UserWarning: File '/data/regression_test/master/in/sherpa/aref_sample.fits' does not have write permission.  Changing to read-only mode.
  warnings.warn("File '" + fname + "' does not have write permission.  Changing to read-only mode.").....
```

The files the test is trying to read in do not need to be edited, so this read-only mode shouldn't affect the results of the regression tests. I have write permissions to the files in `/data/regression_test/master/in/sherpa` and can confirm that files haven't been modified after running the tests.

The test in which these warnings pops up is in `sherpa/astro/sim/tests/test_astro_sim.py`. Specifically, the test calls `pragbayes.py:ARFSIMFacory.read()` method, which uses the `read_table_blocks()` method of the io backend. This function does not accept a mode string, neither in the pycrates nor in the pyfits implementations.

This was an issue that was brought up last year, but was never fixed. @olaurino had done some investigations into this issue last year, which is summarized here.

[...]

PR #210, opened 2016-04-29
New smoke test

olaurino:
# Description
This PR introduces some changes to the way tests work, especially in CIAO. Not much changes for Standalone Sherpa, except that we are officially bringing in `py.test` and `mock` as dependencies for the tests, so we can start using features from these packages.

This PR introduces a new smoke test as a single unittest `TestSuite`, with a convenience high level function in the main namespace. I included some simple tests, but we can beef it up at will, knowing that this is now simply a smoke test. It might be worth adding ds9 tests with the same enforcement mechanism I added for xspec and pycrates/astropy.

The new function `smoke` replaces the old `test` and provides some convenient flags to improve the smoke test effectiveness. In particular, one can request a specific fits backend to be present as well as xspec. Failure to load these modules results in a failure of the smoke test.

A new `sherpa_smoke` console script will be available to both standalone and CIAO users (as we now have `setuptools` in CIAO) as an entry point to the `smoke` function, with options available to make the tests require xspec and/or a specific fits backend. Note that the smoke test will run and skip some tests if xspec and fits backends are not available.

The console script and `smoke` function also accept an option to always make the smoke test fail, which can be used to test the new setup, making sure that when the smoke test fails the appropriate responses are given to the user.

I also removed the `SherpaTest` class and all references to it, as well as the main methods in the tests file, which would make no sense at this point.

Finally, the PR changes the travis builds. So far we had a special build to run the "smoke test", which was really just a subset of unit and regression tests that did not require any test data. I removed the special smoke test build, and I added smoke test runs to the most complex (xspec enabled) and the most basic (no xspec, no matplotlib, no astropy, no setuptools) build.

As we are going to include setuptools in CIAO as well, we can probably get rid of the "nosetuptools" build.

# Testing
The smoke test has its own unit tests which make sure it properly outputs error messages and tries to exit with an error status if backends are not available (even in builds that have the backends or xspec), and that it runs successfully when they are indeed present.

The test classes in the smoke test suite (so far only one class) are also run as part of the unit and regression tests. Test cases requiring backends are skipped accordingly.


DougBurke:
Does the readme need updating to describe the new test suite?

DougBurke:
It wasn't clear to me whether this replaces `sherpa_test` in the standalone build or do we now have both?

olaurino:
> Does the readme need updating to describe the new test suite?

Possibly.

> It wasn't clear to me whether this replaces sherpa_test in the standalone build or do we now have both?

We have both. `sherpa_test` is a means to run all the tests, if you have the test data package/submodule, or the unit tests if you don't. So far, the latter has been analogous to the CIAO "smoke test": now `smoke_test` has that function.

What we want to advertise is up to us: we still need to have both because it's a good thing you can run the whole tests suite with the source and binary distributions, but it's also good that users can run a proper smoke test rather than more extensive tests.

The change originated from the need we had internally to run the Sherpa/CIAO regression tests in a more sustainable way, but also we didn't want the smoke test to require `py.test`.

sherpa/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
```
DougBurke: Why is this file needed?

sherpa/astro/utils/smoke.py
```diff
@@ -0,0 +1,124 @@
...
+
+    test_suite = SmokeTestSuite(require_failure=require_failure)
+    runner = unittest.TextTestRunner(verbosity=verbosity)
+    result = runner.run(test_suite)
+    if result is None or result.failures or result.errors or result.unexpectedSuccesses:
+        sys.exit("Test failures were detected")
+
+
+class SmokeTest(unittest.TestCase):
+    def setUp(self):
```
DougBurke: Somewhere I think we need to specify what we mean by a "smoke test of Sherpa" - i.e. what level of functionality is being tested. I don't know if it's here, the `smoke` module, or just in the documentation (but I think it would be good to have something in the code, close to where the tests are, so that future readers understand the scope).

sherpa/conftest.py
olaurino: It was a placeholder for a prototype that I had to migrate sooner or later. We decided to do it sooner, so there is a new commit with actual code in it.

The file contains project-specific pytest plugins, e.g., as shown below, fixtures that run automatically before and after any test case is run.

DougBurke:
@olaurino - the failure for the @c069c90 build is https://travis-ci.org/sherpa/sherpa/jobs/129431585 which you fixed. However, why did only one of the test runs fail here (i.e. it looks like the test is deterministic, so I would expect to get the warning in each build/test, so shouldn't the new checks cause each test run to fail)? Is it because the "check for warnings and errors" is only done in a subset of the tests, or something else?

olaurino:
That is the only build with numpy=1.9, isn't it? I assumed it was a warning that was gone in more recent versions of numpy. Note that I did not check whether the warnings made sense or whether they could be fixed (unless they were trivial, like the Exception.message ones). That is why I sent the email to sherpadev: if we want to check those warnings and address them, let's do it (by removing them from the white list and make sure the tests pass), but that was not in the scope of my changes: at this round, all current warnings are "expected" warnings.

DougBurke:
Ahhh, that is a reasonable sounding explanation!

DougBurke:
I just tried

    % python setup.py develop
    % sherpa_smoke
    ----------------------------------------------------------------------
    Ran 4 tests in 0.385s

    OK (skipped=2)

So far so good. I then tried `sherpa_test` and it failed as `mock` is not installed (output below). Should it not have been picked up automatically, or do I have to remember to manually install `mock` (or do a `pip install -r blahblah`) step now?

```
% sherpa_test
Collecting pytest
  Downloading pytest-2.9.1-py2.py3-none-any.whl (161kB)
    100% |████████████████████████████████| 163kB 2.3MB/s 
Collecting py>=1.4.29 (from pytest)
  Downloading py-1.4.31-py2.py3-none-any.whl (81kB)
    100% |████████████████████████████████| 92kB 6.4MB/s 
Installing collected packages: py, pytest
Successfully installed py-1.4.31 pytest-2.9.1
You are using pip version 8.1.1, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
platform linux2 -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /home/djburke/miniconda2/envs/sherpa-smoke/bin/python
cachedir: .cache
rootdir: /home/djburke/sherpa/sherpa-smoke, inifile: pytest.ini
collected 323 items / 1 errors 

sherpa/astro/datastack/tests/test_datastack.py::test_design::test_case_1 PASSED
sherpa/astro/datastack/tests/test_datastack.py::test_global::test_case_2 PASSED
...
... a lot of successful and skipped tests happen now
...
SKIP [1] sherpa/ui/tests/test_ui.py:183: required test data missing
SKIP [1] sherpa/astro/tests/test_astro.py:271: required test data missing

==================================== ERRORS ====================================
___________ ERROR collecting sherpa/astro/utils/tests/test_smoke.py ____________
sherpa/astro/utils/tests/test_smoke.py:22: in <module>
    import mock
E   ImportError: No module named mock
============== 214 passed, 109 skipped, 1 error in 35.46 seconds ===============
```

DougBurke:
To be picky: it looks like some of the test files - e.g. `sherpa/sherpa/astro/ui/tests/test_nan.py` and ` sherpa/sherpa/astro/ui/tests/test_astro_ui.py` although this is not an exhaustive check - will need their copyright date updating to include `2016`. I'd do it myself, but I think it's likely to lead to more problems than it solves :-)

sherpa/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+# Whilelist of known warnings. One can associate different warning messages to the same warning class
+known_warnings = {
+    DeprecationWarning:
+        ["unorderable dtypes; returning scalar but in the future this will be an error",
+         "Non-string object detected for the array ordering. Please pass in 'C', 'F', 'A', or 'K' instead",
+         "using a non-integer number instead of an integer will result in an error in the future", ],
+    UserWarning:
+        ["does not have write permission.  Changing to read-only mode.", ],
```
DougBurke: As I mentioned in https://github.com/sherpa/sherpa/issues/123#issuecomment-170118512 I strongly believe we should change the code to explicitly set "mode=read-only" rather than "read-write", so we can hopefully remove this warning from the set.

olaurino:
mock is installed automatically if you run python setup.py test.

However, sherpa_test is supposed to run even when you don't have setup.py in the first place, so I'll add mock to the list of dependencies checked and installed by sherpa_test.

I don't expect these dependencies to change for a long time, so It does not make sense to do something cleaner so avoid having to repeat the test dependencies. However, I am keeping it simple and I won't bother to make the code compliant with python 3 just yet. I'll worry about that later.

Thanks for testing this out!

DougBurke:
@ad935b9 seems to have done the trick for me. ta.

DougBurke:
@olaurino can you re-run the tests for the last check in; the failures are from `apt-get` being unable to access something on the net, and so are unlikely to be related to your change. I can never remember how to re-run a test (maybe I can't in this case as it's not my PR?)

olaurino:
What scares me is that this time I did not get a notification... Travis had several incidents investigations going on in the past few days, probably they had to throw away a number of notifications.

olaurino:
@DougBurke assuming you are logged in, if you open up a specific failing run, there should be a "refresh" icon in the top right corner of the UI, which can be used for rerunning the build. It would be useful to know if you can refresh of the PRs or just yours. I don't think there are enough permission differences between our accounts for preventing yours from restarting builds, but you never know.

DougBurke:
Ahhh, I see the refresh icon. I think I was looking for text rather than an image earlier today. Ta.

olaurino:
This PR has been tested on CIAOD for about a week, and things seem to be stable on the CIAO side.

@DougBurke do you have any objections that we merge this?

DougBurke:
@olaurino - I was hoping for some documentation updates:

  - to the readme to say what users should (or shouldn't) do (it can come in after #207 and #208 are merged, to avoid conflict, but it would be good to have it on record as changes are coming)

  - somewhere in the code to say what we mean by a smoke test - i.e. what level of functionality should be tested - so that anyone adding new code or a module knows whether they need to add something to the smoke-test functionality or not.

It also wasn't clear to me from the code when or how to use the "specify your own data directory" functionality you added in near the end. Is that something to be documented in the README.md?

olaurino:
@DougBurke those are fair concerns. Can I include what you just wrote in a new issue? The trick is that we need to coordinate these changes with CIAO's CM, and there is a code freeze coming up, so unless you have strong objections I would try and get this merged for CM to pick up in tonight's build.

DougBurke:
Also, don't we need a "release notes" section in the PR (or something similar)? I think the existing text contains enough information, but not really in the right form.

[I started this before seeing your comment about trying to get this change in asap; I don't know how you pick up the relelease notes from the PRs so you can better answer how the lack of any notes is going to impact the next release]

DougBurke:
Overall: I would prefer a bit more documentation (both in the code and in other places) to really understand what is going on here, but I am happy for the existing code to go in, with the understanding that we do get improved documentation on this functionality in the near future.

I am unclear on how you build the release notes for a release from the PR documentation, so I'll assume that what we have is okay there.

olaurino:
The release notes depend on what we want to advertise up front to users, in particular standalone users.

PR #211, opened 2016-05-03
Cleanup of documentation and code in sherpa.astro.utils

DougBurke:
# Release Notes

The `calc_kcorr` function is now exported by `sherpa.astro.utils`.

Minor changes to the documentation in `sherpa.astro.utils` to conform to Sphinx standards.

The code was changed to remove some PEP8 and pyflakes warnings, including the use of explicit imports and adding or removing spaces, parentheses, or blank lines.

# Notes

This came out of my sphinx work, where I noticed that calc_kcorr was not exported, and then followed up with some documentation clean up, and then code clean up. I have tried to avoid making sweeping changes to make this an easy PR to review.

The PEP8/pyflakes settings I use still complain about the region symbols that are exported, since it doesn't think they are used even though they both get added to __all__. I did not think it worth making this PR more complex to remove the warning.

### DIFF
--- a/sherpa/astro/utils/__init__.py
+++ b/sherpa/astro/utils/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2008, 2016  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


olaurino:
:+1: Merge it!

Issue #212, opened 2016-05-18
Refine documentation for new smoke tests


olaurino:
As mentioned by @DougBurke in https://github.com/sherpa/sherpa/pull/210#issuecomment-220151005

We need to add some documentation updates to PR #210:
- to the readme to say what users should (or shouldn't) do (it can come in after #207 and #208 are merged, to avoid conflict, but it would be good to have it on record as changes are coming)
- somewhere in the code to say what we mean by a smoke test - i.e. what level of functionality should be tested - so that anyone adding new code or a module knows whether they need to add something to the smoke-test functionality or not.

It also wasn't clear to me from the code when or how to use the "specify your own data directory" functionality you added in near the end. Is that something to be documented in the README.md?

We merged PR #210 so it could go in the next CIAOX build, but we were in the middle of the review (and of commitments to other projects).

olaurino:
@DougBurke, I am not sure what it makes sense to advertise to the user. The changes in #210 were meant to fix some issues we were having on the CIAO side of sherpa testing, and as a bonus I think we streamlined the way the Smoke Test works as it is now a well defined subset of tests that ensures Sherpa is correctly installed along with its dependencies. I don't think end users are too much interested in the different levels of testing, but the code should be documented enough to understand what is going on, I agree, and the README should advertise the interface we want to advertise. To me, that's still the `sherpa_test` command and the test data.

I don't think developers should be too much concerned with the Smoke Test, and I don't expect it to change too often.

I will try and add some docs to the code for the `pytest` plugin and the Smoke Test functions. We can work out together what to add to the README.

DougBurke:
We've moved to "fully embracing" `pytest`, and there have been `sherpa_test` changes, so I'm going to close this issue. There probably is still some documentation to add but let's worry about this with a more-targeted issue if anyone finds something missing.

PR #213, opened 2016-05-18
add docstrings to smoke test classes and functions

olaurino:
Address #212
 
# Description
This PR adds some docstrings to #210, to explain the purpose and extent of the smoke test.

We may want to add a commit to change the README, but it may not be worth providing users with too many testing options. We might just leave the current `sherpa_test` documentation unchanged and do not advertise the Smoke Test option explicitly to standalone users, which do not have the limitations of the CIAO users (e.g. no mock, no pytest).

### DIFF
--- a/sherpa/astro/utils/smoke.py
+++ b/sherpa/astro/utils/smoke.py
@@ -32,19 +32,26 @@ logger = logging.getLogger("sherpa")
 
 def run(verbosity=0, require_failure=False, fits=None, xspec=False):
     """
-    Run the smoke tests, i.e. a small set of tests designed to verify that the installation does not have
+    Run the smoke tests.
+
+    A smoke test is a small set of tests designed to verify that the installation does not have
     any obvious issues.
 
     Parameters
     ----------
-    verbosity non-negative int. Set verbosity level: the higher the level, the more verbose the test. This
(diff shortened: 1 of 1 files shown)


DougBurke:
Ideally the docstrings would follow astropy/numpy guidelines - e.g. http://docs.astropy.org/en/stable/development/docrules.html

@olaurino - would you like me to fix up these docs to better-match the guidelines (at some point in the near-ish future)?

olaurino:
@DougBurke I thought I was following those guidelines. After checking the documentation again I can see that I am missing a `newline` after the argument type. Am I missing something else?

DougBurke:
It's a bit hard to tell from the GitHub diff view how long some of these lines are: the first line of the docstring is meant to be "short" to act as a summary view (or at least, that's what Sphinx seems to like). I do find I sometimes end up creating sphinx documentation just to check that the docstrings look okay!

olaurino:
The lines are all less than 120 characters, which is our stanrdard.

I made some changes. Please let me know if you think they are sufficient.

Also, if there is anything we, as developers, should do to check that the documentation is correctly formatted (e.g. run sphinx build, or a special script), let's make sure we document that.

Issue #149, opened 2016-01-15
bug with filter/grouping


dtnguyen2:
Mike McCollough reported a bug while running specfit.  I was able to trace the problem to the following the code of get_staterror in the file sherpa/astro/data.py.  In this particular case, specfit is normally with a notice region and a requirement of a minimum of 16 counts per bin.  As a result, within the aforementioned function: filter is True so apply_filter (which internally calls apply_grouping) returns an array with one of the elements < 16 (the last element of the cnts array: 0) which is obviously a bug.  I suspect that once this bug is fixed, then a lot of our tests will fail.  I pity the poor soul that has to fix this bug.

```
        if filter:
            cnts = self.apply_filter(cnts)
        else:
            cnts = self.apply_grouping(cnts)
```

the cnts array looks like the following, note the last 0

(Pdb) print cnts
[ 18.  16.  17.  24.  16.  19.  23.  22.  17.  17.  19.  16.  17.  16.  19.
1.  16.  19.  16.  16.  16.  17.  16.  16.  16.  16.  16.   0.]

In short there is something wrong with the filter/group

jbudynk:
@dtnguyen2 and I were investigating the issue, and it doesn't look like a bug anymore.

If using `group_counts(num)`, it is possible for the last array in a grouped dataset to have less than `num`. 

If we reach the end of a dataset, and the summed counts for the last bin are less than `num`, the group is marked as incomplete and given a quality of 2, meaning they're "bad" and should be ignored in analysis (See [extern/grplib/grplib.c::grp_do_num_counts()](https://github.com/sherpa/sherpa/blob/master/extern/grplib-4.7/src/grplib.c#L130) at around line 139). 

One can ignore these poor quality points with `sherpa.astro.ui.ignore_bad()` before analysis.

For example:

```
from sherpa.astro.ui import *

# load the 3c273.pi file in sherpa-test-data/sherpatest/
load_pha('3c273.pi')
group(16)
print get_data().to_fit()[0]

[17.0, 16.0, 17.0, 16.0, 18.0, 21.0, 17.0, 23.0, 18.0, 21.0,
  22.0, 21.0, 19.0, 21.0, 17.0, 17.0, 17.0, 17.0, 21.0, 17.0,
  20.0, 17.0, 18.0, 17.0, 18.0, 17.0, 16.0, 16.0, 17.0, 17.0,
[...]

PR #225, opened 2016-06-13
Bug #149 filter grouping

jbudynk:
# Release Note

Fixes bug #149. If a dataset is filtered and then grouped with `group_counts(grpnum)`, the last bin in the grouped array, while less than `grpnum`, is included in the array. The sherpa documentation states that `group_counts(grpnum)` will group the dataset counts so that each bin contains at least `grpnum` counts.

This fix ensures that each bin in a grouped dataset upholds to the given user condition, be it enough counts, SNR, etc.

A number of simple grouping tests for each grouping method (`group_snr(), group_adapt(), etc.`) have also been added.
# Status

This bug is fixed in the sense that the steps in #149 result in the expected output. However, this fix only works for datasets that have been filtered before being grouped. We still need to discuss and fix what happens to unfilled bins in grouped datasets. 

Currently, if no filters have been applied to the dataset, the data is simply grouped by the external grplib functions, which by default leave the last unfilled bin in the grouped dataset. The unfilled groups are given a quality of 2; the user can ignore the unfilled group with `ignore_bad()`.

For now, I've put the PR on hold because one test, `sherpa/astro/tests/test_data.py::test_grouping::test_grouping_with_previous_quality_flags()`, will fail until we decide how to handle grouping, filtering, and quality flags. I can set this test to fail if we do not want to tackle this issue in this PR.

Note that a number of fitting tests are failing because the grouping logic has changed, which affects the fit results as the groups are different.
# Description

This bug is part of a bigger issue on grouping, filtering, and quality flags. We need to decide what the behavior of grouping should be when a bin is left unfilled - must it never be added to the grouped array in the external grplib code? Should sherpa handle ignoring unfilled bins internally? Should we make the user ignore the unfilled groups themselves? And so on.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -877,13 +877,7 @@ class DataPHA(Data1DInt):
 
         self.grouping, self.quality = group_func(*args, **kwargs)
 
-        # for fixing bug #149. Find the last group. If it's bad quality,
-        # remove it from the grouped view. This could be fixed in
-        # the grouping library; instead of assigning quality=2 to unfilled
-        # groups, just ignore them completely from the grouped view.
-        # last_bad_group_index = numpy.where(self.grouping == 1)[0][-1]
-        # if self.quality[last_bad_group_index] == 2:
-        #     self.grouping[last_bad_group_index:] = 0
+        # Ignore bad groups automatically (for bug #149)
         self._ignore_bad_groups()
 
         self.group()
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -837,6 +837,37 @@ class DataPHA(Data1DInt):
         # self.quality_filter used for pre-grouping filter
         self.quality_filter = qual_flags
 
+    def _ignore_bad_groups(self):
+        """Exclude grouped channels marked as bad.
+
+        Ignore any bin in the PHA data set which has a quality value
+        that is equal to 2.
+
+        Raises
+        ------
+        sherpa.utils.err.DataErr
+           If the data set has no quality array.
+
+        See Also
+        --------
+        ignore : Exclude data from the fit.
+        notice : Include data in the fit.
+
+        Notes
+        -----
+        Bins with a non-zero quality setting are not automatically
+        excluded when a data set is created.
+
+        """
+        if self.quality is None:
+            raise DataErr("noquality", self.name)
+
+        # ignore bad groups (where quality flags == 2)
+        qual_flags = ~numpy.asarray(self.quality == 2, bool)
+
+        # self.quality_filter used for pre-grouping filter
+        self.quality_filter = qual_flags
+
     def _dynamic_group(self, group_func, *args, **kwargs):
 
         keys = list(kwargs.keys())[:]
@@ -844,20 +875,22 @@ class DataPHA(Data1DInt):
             if kwargs[key] is None:
                 kwargs.pop(key)
 
-        old_filter = self.get_filter(group=False)
-        do_notice = numpy.iterable(self.mask)
-
         self.grouping, self.quality = group_func(*args, **kwargs)
+
+        # for fixing bug #149. Find the last group. If it's bad quality,
+        # remove it from the grouped view. This could be fixed in
+        # the grouping library; instead of assigning quality=2 to unfilled
+        # groups, just ignore them completely from the grouped view.
+        # last_bad_group_index = numpy.where(self.grouping == 1)[0][-1]
+        # if self.quality[last_bad_group_index] == 2:
+        #     self.grouping[last_bad_group_index:] = 0
+        self._ignore_bad_groups()
+
         self.group()
         self._original_groups = False
 
-        if do_notice:
-            # self.group() above has cleared the filter if applicable
-            # No, that just sets a flag.  So manually clear filter
-            # here
-            self.ignore()
-            for vals in parse_expr(old_filter):
-                self.notice(*vals)
+        # self.group() above has cleared the filter if applicable; no need to
+        # re-filter the spectrum. See DataPHA._set_grouped()
 
         # warning('grouping flags have changed, noticing all bins')
 
@@ -1651,7 +1684,8 @@ class DataPHA(Data1DInt):
         # Go on if we are also supposed to filter the source data
         ignore = bool_cast(ignore)
         if lo is None and hi is None:
-            self.quality_filter = None
+            # commented out to fix bug #149. TODO: is this OK to remove?
+            #self.quality_filter = None
             self.notice_response(False)
 
         elo, ehi = self._get_ebins()
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -317,12 +317,13 @@ class test_threads(SherpaTestCase):
     # As of CIAO 4.5, can filter on channel number, even when
     # data are grouped! Test results should exactly match CIAO 4.4
     # fit results in grouped/fit.py
+    # jbudynk [sherpa 4.8.1]: fit results updated to fix bug #149.
     def test_grouped_ciao4_5(self):
         self.run_thread('grouped_ciao4.5')
-        self.assertEqualWithinTol(ui.get_fit_results().statval, 18.8316, 1e-4)
-        self.assertEqual(ui.get_fit_results().numpoints, 46)
-        self.assertEqualWithinTol(self.locals['aa'].gamma.val, 1.83906, 1e-4)
-        self.assertEqualWithinTol(self.locals['aa'].ampl.val, 0.000301258, 1e-4)
+        self.assertEqualWithinTol(ui.get_fit_results().statval, 18.5383, 1e-4)
+        self.assertEqual(ui.get_fit_results().numpoints, 45)
+        self.assertEqualWithinTol(self.locals['aa'].gamma.val, 1.81842, 1e-4)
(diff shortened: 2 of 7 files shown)


DougBurke:
@jbudynk - it looks like the tests are failing in multiple places, not just `sherpa/astro/tests/test_data.py::test_grouping::test_grouping_with_previous_quality_flags()` - is this expected with the current PR?

jbudynk:
Yes - because the grouping is different, we expect some differences in the fit results for some tests. I haven't reconciled all of them because (1) I want to make sure I'm following the expected behavior of the tests, and (2) it's a little tedious and I need more time :) . 

Thanks for pointing this out - I'll add that to the status.

I mainly wanted to get this up so we have a place for discussing the grouping issues, and so it's easier to view the code changes.

jbudynk:
Hmm... I am noticing more than just the statistics-related tests failing... all the tests in sherpa/astro/tests/test_data pass on my system (except test_grouping_with_previous_quality_flags(), which is expected to fail at this point), but I see them failing on Travis. I'll take a look...

jbudynk:
Rebased. Original tip: 87077c2

DougBurke:
@jbudynk - there still seems to be some unexpected failures arfter your rebase - e.g.

 - 1006.6 also fails in `test_ui.test_restore_usermodel` https://travis-ci.org/sherpa/sherpa/jobs/137350474#L4690

 - 1006.7 also fails with a problem loading in the io module https://travis-ci.org/sherpa/sherpa/jobs/137350475#L4553 - perhaps the test should be marked as `requires_fits` (although I'm not convinced that I understand what's actually gone on in this test failure, so I could be going doen the wrong rabbit hole)?

I suggest adding a "expected_to_fail" decorator (I forget how to do this) to `sherpa/astro/tests/test_data.py::test_grouping::test_grouping_with_previous_quality_flags` just so we know everything else is working with these patches. Then when we've resolved the test we can remove this hint to check everything is okay.

olaurino:
> I suggest adding a "expected_to_fail" decorator (I forget how to do this) to sherpa/astro/tests/test_data.py::test_grouping::test_grouping_with_previous_quality_flags just so we know everything else is working with these patches. Then when we've resolved the test we can remove this hint to check everything is okay.

Since this test was added by this PR my counter-suggestion would be to remove it from this PR, and add it to a new branch backing a new PR, so we make sure we fix it, even though we decide it's not going to be fixed in the scope of this PR. I am not a fan of marking tests as expected to fail, even when the testing framework allows it.

DougBurke:
I'll leave it up to @jbudynk which she prefers to do ;-)

DougBurke:
@jbudynk it looks like there are still some issues with this PR, but not for all cases. Is it understood why this is happening, or do you need to look into it?

jbudynk:
For the most part, yes. The fitting tests are most likely failing because the datasets were grouped beforehand. Since the groupings for those datasets may have changed, the fit results will change too. I still need to investigate what the right results should be...

I've added a fix for the `test_ui.restore_usermodel()` failure in the latest commit. I was catching the wrong type of exception elsewhere in the code.

I've also removed the quality flags test since this issue can be tackled in another PR. Right now, we're just trying to make sure the last unfilled bin in the array is not included in the grouped view of the dataset. If it turns out that we'll need that test, we can add it back in.

DougBurke:
I don't quite understand the pattern of failures - e.g. why does run#2 with XSPEC pass but runs #4 and #5 with XSPEC fail? Is it something like run#2 doesn't load in the sherpa-test-data repository so the tests aren't run in that build, or is it something else?

DougBurke:
Run #7 also looks to be a completely different failure (hence the separate comment) - e.g. https://travis-ci.org/sherpa/sherpa/jobs/137564449#L4542 - fails with

    ImportError: Cannot import selected FITS I/O backend pyfits_backend.

jbudynk:
Huh... yeah not sure what's happening there. I was focusing on the fit results failures. I haven't looked at these other failures yet to be honest. But I'll take a look at them.

jbudynk:
I'm adding the quality flag test back in... it's probably best to tackle both issues at once so we don't have to possibly rewrite grouping and/or fit results tests twice.

DougBurke:
@olaurino is the build failure for the setuptools test (run 1) the same issue that you mentioned on email; that is, it is time to remove this particular build from the test suite?

https://travis-ci.org/sherpa/sherpa/jobs/138910868#L410

i.e. cherry-pick your commit 43df53d

olaurino:
> is the build failure for the setuptools test (run 1) the same issue that you mentioned on email; that is, it is time to remove this particular build from the test suite?

Yes, that's correct, @DougBurke.

jbudynk:
Hmm... I notice an issue with the current fix. Right now, the last bin in a grouped dataset is *always* removed if the dataset was filtered. For unfilled bins, this is an OK fix, but if the group condition is satisfied in the last bin, we don't want to remove that last bin from the grouped view. This is the fix introduced in 5f20972f7f2f, in `sherpa/data.py`.

You can see an issue with `sherpa/astro/tests/test_astro.py::test_threads::test_xmm()`. The dataset is grouped with `group_width(10)`, and the notice range is channel 40 to 201. The last bin, while is contains 10 channels and has a good quality flag, is removed from the filtered view. That should not happen.

Unfortunately, I hadn't covered this issue in the tests because only one of the tests filters the dataset before grouping, and that is for Nick Lee's bug report with quality flags (the one just added back in today). I'll add a test that sets broad filter on one of the simple grouping tests so we can catch this issue.

DougBurke:
Once we've resolved this issue I'd like to make sure that at least one test case is added to the `sherpa.astro.ui.save_all` test suite - i.e. `sherpa/astro/ui/tests/test_serialize.py` - to make sure that the correct filter and quality flags are saved. This can be done in a separate PR, but I wanted to write it down somewhere so that I don't forget to do it.

jbudynk:
Rebasing on master, with new commits. Original tip was 3203cd5.

jbudynk:
I've got a fix for the issue of making sure each grouping condition is met, and there are no "bad quality" groups shown when viewing the "grouped" view of a dataset (e.g., using `sherpa.astro.ui.get_data().get_dep(filter=True)`). 

However, there is a change I don't like and would like some feedback on and more testing to see if this breaks any other behavior.

I made a new function, `_ignore_bad_groups()`, which works similarly to `ignore_bad()`, but doesn't undo the previously-set filters. It only ignores channels with quality == 2.

To understand the changes better, I'll summarize how grplib's code comes into play. When a dataset is grouped, and a specific group doesn't meet the grouping criteria, the channels in that group are assigned a value of 2. This quality array is assigned to `self.quality` in DataPHA. So in `sherpa.astro.data::DataPHA._dynamic_group()`,  `_ignore_bad_groups()` is called, which sets `self.quality_filter`, a boolean array representation of `self.quality`. This boolean array is used to further filter the groups in `DataPHA.apply_grouping()`:

Originally, I thought we could change the grouping array that comes out of grplib, so that when we found group qualities of 2, we would set the `self.grouping` in those bins to 0, which in grplib's terms means "unused group". However, back when Stephen was working here, "0" groups were considered groups themselves (for filtering reasons?? I guess in case there is a big "ignore" region in the middle of a dataset, and you group the dataset, that you can see that big gap in the data). The relevant code is in [sherpa/include/sherpa/astro/utils.hh::_do_group()](https://github.com/sherpa/sherpa/blob/master/sherpa/include/sherpa/astro/utils.hh#L355) by line [374](https://github.com/sherpa/sherpa/blob/master/sherpa/include/sherpa/astro/utils.hh#L374). All group values >= 0 are considered "groups", and all other values are considered to be inside a group.

jbudynk:
Tests `sherpa/astro/ui/tests/test_serialize.py::test_ui::test_canonical_pha_grouped`, `sherpa/astro/ui/tests/test_serialize.py::test_ui::test_canonical_pha_back`, and `sherpa/astro/tests/test_astro.py::test_threads::test_setfullmodel` are erroring due to a VisibleDeprecationWarning in [DataPHA.get_noticed_channels()](https://github.com/sherpa/sherpa/blob/bug-%23149-filter-grouping/sherpa/astro/data.py#L1580). [DataPHA.get_mask()](https://github.com/sherpa/sherpa/blob/bug-%23149-filter-grouping/sherpa/astro/data.py#L1582) returns an array of different dimensions than the PHA channels.

DougBurke:
The deprecation warnings are part of issue #273 (or, at least, they now are; the original 273 report doesn't explicitly list this example).

jbudynk:
On a note related to grouping but not to this issue:

in [DataPHA._dynamic_group()](https://github.com/sherpa/sherpa/blob/master/sherpa/astro/data.py#L856), the following is stated: 

        self.grouping, self.quality = group_func(*args, **kwargs)
        self.group()
        self._original_groups = False

        if do_notice:
            # self.group() above has cleared the filter if applicable
            # No, that just sets a flag.  So manually clear filter
            # here
            self.ignore()
            for vals in parse_expr(old_filter):
                self.notice(*vals)

self.group() *does* clear the filter, if applicable. It calls [`_set_grouped()`](https://github.com/sherpa/sherpa/blob/master/sherpa/astro/data.py#L253), which does the whole `self.ignore(); for vals in parse_expr(old_filter):` called in _dynamic_group(). The `if do_notice:` conditional can be removed.

olaurino:
The warnings here are different, so this is NOT related to #273 (which btw should be fixed in the context of a different bug fix by a PR I am about to issue).

olaurino:
This PR is no longer relevant, so while it might still contain code and/or tests that will be useful in the future, I am going to close it.

PR #231, opened 2016-06-30
remove nosetuptools travis build

olaurino:
This PR removes a travis build that is not required anymore (the purpose of the build was to test CIAO support, but CIAO now includes setuptools), and that is actually erroring on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,8 +93,6 @@ before_install:
   - if [ ${TEST} == smoke ];
      then git submodule deinit -f .;
     fi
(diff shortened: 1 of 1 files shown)


DougBurke:
Can line 96 now be removed as well (to avoid leaving in confusing code)?

https://github.com/sherpa/sherpa/blob/cf2fff33e486f8c4b7a1c2989da721e6e1ffe6aa/.travis.yml#L96

olaurino:
Good idea.

DougBurke:
I'm happy for this to be merged, but I'd like to wait until travis has validated the 9534e12 commit before labelling it pr:merge (I know this has been validated in other builds, but I'd like to make sure here).

PR #243, opened 2016-08-23
Add Python 3.5 to the list of supported interpreters.

DougBurke:
Adds the string `Programming Language :: Python :: 3.5` to the classifiers list in setup.py.

The tests fail because I do not have the latest fixes from the Python3 branch. It couldn't possibly be because of my code changes :-)


olaurino:
Rebased. Original tip a0196a4.

PR #245, opened 2016-08-24
Update copyright year to include 2016.

DougBurke:
This also changes the case of one of the copyright statements. Note that
files in extern/ are not changed in this commit, and there is no
copyright statement or license in sherpa/include/capsulethunk.h
(I assume it's the same as the Python version).


olaurino:
capsulethunk.h is taken off the python documentation:
https://docs.python.org/2/howto/cporting.html

To be perfectly honest, we probably don't need it since we don't support the python versions this file is supposed to cover. We might just as well remove it, I believe.

In any case, I am merging this :)

Thanks!

PR #247, opened 2016-08-25
Mention Python 3.5 support in README

DougBurke:
Update the README.md file to mention that Python 3.5 support is available. It's a bit complicated since, until there's a binary version released with this support, the documentation needs to distinguish the two cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 [![Build Status](https://travis-ci.org/sherpa/sherpa.svg?branch=master)](https://travis-ci.org/sherpa/sherpa)
 [![DOI](https://zenodo.org/badge/683/sherpa/sherpa.svg)](https://zenodo.org/badge/latestdoi/683/sherpa/sherpa)
 [![GPLv3+ License](https://img.shields.io/badge/license-GPLv3+-blue.svg)](https://www.gnu.org/copyleft/gpl.html)
+![Python version](https://img.shields.io/badge/Python-2.7,3.5-green.svg?style=flat)
 
 <!-- TOC *generated with [DocToc](https://github.com/thlorenz/doctoc)* -->
 **Table of Contents**
@@ -73,6 +74,8 @@ How To Install Sherpa
 =====================
 
 Sherpa can be installed from a binary distribution or built from sources.
+The 4.8.1 release is only available for Python 2.7, while 4.8.2 will
+also support Python 3.5.
 
 The binary distribution is suited for people wanting to have Sherpa up
 and running as soon as possible in its standard form.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,8 +89,13 @@ with the Conda installation (sections [1a](#1a-anaconda) and [1b](#1b-starting-f
 
 1. Binary installation (Anaconda)
 
+The binary distribution of Sherpa 4.8.1 only supports Python 2.7.
+
 2. Source build (from a source tarball or the GitHub repository)
 
+A source build is - at present - the *only* way to use Sherpa
```
olaurino: Let's remember to change this when we release, as at that point it won't probably be true any more ;)

DougBurke: Well, we'll have to change a few other things in the file then, so hopefully we'll notice!

DougBurke:
I've just merged this into master, but it doesn't seem to have been picked up by GitHub (i.e. it hasn't automatically been closed). I did the same thing I did for the other PRs I merged today (I'm pretty sure) so I don't know why this wasn't automatically closed. I'm going to close it, but I don't know if @olaurino wants to check things?

DougBurke:
Here's the history of what I did

```
  601  git fetch origin
  602  git checkout master
  603  git fetch origin pull/247/head:pr/247
  604  git merge --no-ff pr/247 -m 'Merge #247 (DougBurke) - Mention Python 3.5 support in README'
  605  git status
  606  git stash pop
  607  git status
  608  git diff
  609  git log
  610  git status
  611  git log
  612  less README.md 
  613  git diff master
  614  git diff --help
  615  git diff --cached master
  616  git log
  617  git diff bda80835e5a230c70d9065a0aad5dccc1dc9cdc0
  618  git status
  619  git push origin master
  620  history
```

where

```
% git remote -v
origin	http://github.com/sherpa/sherpa (fetch)
origin	http://github.com/sherpa/sherpa (push)
```

the git stash pop was just to get at a setup.cfg file with the xspec configuation in it, so I could test it if I wanted, but I didn't do so. So I don't know why it things some commits are unmerged.

olaurino:
@DougBurke the problem is that the PR was opened against the `python3` branch. In that sense, it's true that the commits are unmerged. In this case, I don't think it's a problem, because there were not commits piggy-backing, and only one commit was supposed to be merged, either in `master` or `python3`.

We should (both) be more careful next time. In particular, when I merged the python 3 PR I did realize there were PRs still depending on that branch, because I could not delete the `python3` branch in the branches page. I didn't have time to go through the open PRs to find the offending one, and then I didn't get back to it.

DougBurke:
So, apart from being a learning experience, we are okay with this
particular branch. Yay!

DougBurke:
PS thanks for looking

Issue #238, opened 2016-07-25
Rewrite sherpa/tests/test_stat.py so it does not depend on fits, xspec, or data


DougBurke:
The statistical tests require xspec, a FITS backend, and sherpa-test-data to run (see below). We should not need any of these, as it is important to be able to verify the stats when the XSPEC or the FITS backend is not available. A quick look at the tests show that they do things like run a fit and then compare results: this also seems excessive for a test of the stats, as it is more of an integration or end-to-end test.

I suggest that the tests should just be of "simple" numerical models and data, and without needing any I/O. The existing tests in this module should probably not be lost, but moved to a different location.

```
% python setup.py test -a sherpa/stats/tests/test_stats.py 
...
=========================== short test summary info ============================
SKIP [1] sherpa/stats/tests/test_stats.py:299: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:268: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:245: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:235: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:240: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:262: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:257: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:279: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:251: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:305: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:230: xspec required
SKIP [1] sherpa/stats/tests/test_stats.py:273: xspec required

========================== 12 skipped in 0.48 seconds ==========================
```

olaurino:
I completely agree.

DougBurke:
Note that #287 contains a number of low-level tests of the statistic code that do not depend on external data files or XSPEC (and so are always run, whatever optional functionality is available). I leave the review for #287 to determine whether it is sufficient to close this issue.

DougBurke:
I am assuming that with #287 merged we can close this (the tests mentioned here haven't been changed, but we have since added a number of tests that are self-contained).

PR #249, opened 2016-08-26
Apply noticed channels to backscale ratio array in WStat

### LINKED ISSUE #248: backscal column not treated properly for WStat
If you provide the background scaling factor not as a header keyword but as data column the implementation of wstat cannot deal with this:

```
import sherpa.astro.ui as sau
import numpy as np

sau.load_pha('/home/kingj/Software/sherpa/sherpa-test-data/sherpatest/3c273.pi')
sau.ungroup()
source_backscal = sau.get_data().get_backscal()
bkg_backscal = sau.get_bkg().get_backscal()
n_channels = len(sau.get_data().channel)

sau.get_data().backscal = np.ones(n_channels) * source_backscal
sau.get_bkg().backscal = np.ones(n_channels) * bkg_backscal

sau.set_stat('wstat')

sau.set_model('bbody.source')
sau.notice(2,20)
sau.fit()
```

raises this error

```
  File "analyse.py", line 17, in <module>
    sau.fit()
  File "<string>", line 1, in fit
  File "/home/kingj/Software/sherpa/sherpa/astro/ui/utils.py", line 9262, in fit
    self._fit(id, *otherids, **kwargs)
  File "/home/kingj/Software/sherpa/sherpa/astro/ui/utils.py", line 9366, in _fit
    res = f.fit(**kwargs)
  File "/home/kingj/Software/sherpa/sherpa/fit.py", line 58, in run
    result = func(fit, *args, **kwargs)
  File "/home/kingj/Software/sherpa/sherpa/fit.py", line 1107, in fit
    init_stat = self.calc_stat()
  File "/home/kingj/Software/sherpa/sherpa/fit.py", line 1021, in calc_stat
    bkg=bkg_data)[0]
  File "/home/kingj/Software/sherpa/sherpa/stats/__init__.py", line 601, in calc_stat
    truncation_value)
TypeError: statistic input array sizes do not match
```

The problem is that the information about noticed channels is not forwarded to the backscal array, uncommenting the sau.notice() line 'solves' the error, uncommenting the sau.ungroup() line introduces it again

```
(Pdb) bkg_data['data_size']
array([888])
(Pdb) len(bkg_data['backscale_ratio'])
1024
```

I would be happy to try fixing this myself and am thus grateful for any hint where to start.

joleroi:
Fix #248
# Release Note

Sherpa now handles the case of the background scaling factor not being provided as a header keyword but as an array in `wstat`. Noticed channels are now properly applied to the backscale ratio array.
# Notes

We should provide more thorough and isolated tests for `wstat` and other statistics, which is captured in #238. This work is however out of the scope of this PR.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -523,7 +523,10 @@ class IterFit(NoNewAttributesAfterInit):
                 exposure_time[2 * index + 1] = bkg.exposure
                 ratio = vectorize_backscale_ratio(bkg.backscal,
                                                   mydata.backscal,
-                                                  data_size[index])
+                                                  len(mydata.channel))
+                noticed_channels = array(mydata.get_noticed_channels(),
+                                         dtype=int)  
+                ratio = ratio[noticed_channels-1]
                 backscale_ratio = append(backscale_ratio, ratio)
             else:
                 return result
--- a/sherpa/stats/tests/test_stats.py
+++ b/sherpa/stats/tests/test_stats.py
@@ -197,6 +197,9 @@ class test_stats(SherpaTestCase):
         self.data = read_pha(pha_fname)
         self.data.notice(0.5, 7.0)
 
+        # hard-code number of channels
+        self.data.backscal = self.data.backscal * numpy.ones(1024)
+         
         bkg_fname = self.make_path("9774_bg.pi")
         self.bkg = read_pha(bkg_fname)
 


DougBurke:
Thanks for working on this. We are obviously missing some tests here - with regard to filtering data - so would it be possible to add some? As noted in #238 our stats tests are not well organised, so it's not clear to me whether code should be there or in some of the higher level areas (e.g. sherpa/astro/ui/utils), or both.

It also may not be worth making large changes just at the moment, as I hope #229 is going to land soon, which is quite an invasive change.

DougBurke:
Having a quick look at `sherpa/stats/tests/test_stats.py`, it would be great if we had some tests that just evaluated a given model (i.e. without the need to fit it) for a range of situations - e.g. focussing on wstat for this case

- all data
- ignored data
- grouped data (ideally wstat is used without grouping, but if you read the XSPEC notes on it there's a suggestion to group so the background has at least one count per bin)
- ignored and grouped data (since we know from other tickets that there can be some complexities here)

and to include "standard" PHA (e.g. with `backscal` a scalar) and "grating style" (where `backscal` is a vector) data. However, that's a bit much for this PR. @joleroi as you're looking at it, what do you think the minimum testing that makes sense for this change?

joleroi:
I had a look at #238 and #229 and ``sherpa/stats/test/test_stats.py`` and indeed, fixing all issues with the test suite is a little optimistic for this PR :wink: .
Just adding a line here https://github.com/sherpa/sherpa/blob/master/sherpa/stats/tests/test_stats.py#L199
that introduces a ``backscal`` vector to the data would suffice to validate the change introduce in this PR, I think. But I agree that it's not a very nice solution.

DougBurke:
I can confirm that the following fails on `master` for only the `wstat` test, so if it now passes in your branch, could you add it (or something similar) to this PR @joleroi  (I'm off out to see Raiders of the Lost Ark with backing by the Boston pops so won't be around until next week)?

```
% git diff sherpa
diff --git a/sherpa/stats/tests/test_stats.py b/sherpa/stats/tests/test_stats.py
index fdc64be..a5bee71 100644
--- a/sherpa/stats/tests/test_stats.py
+++ b/sherpa/stats/tests/test_stats.py
@@ -197,6 +197,9 @@ class test_stats(SherpaTestCase):
         self.data = read_pha(pha_fname)
         self.data.notice(0.5, 7.0)
 
+        # hard-code number of channels
+        self.data.backscal = self.data.backscal * numpy.ones(1024)
+
         bkg_fname = self.make_path("9774_bg.pi")
         self.bkg = read_pha(bkg_fname)
 
```

joleroi:
> I'm off out to see Raiders of the Lost Ark with backing by the Boston pops

Awesome!

I added your test, so if tests pass you can merge, I guess.

DougBurke:
Thanks for this. There'll probably be a little delay in merging this, as we have #252 going in that is slowing us down a bit. I also need to check with @olaurino if we need any further documentation (in terms of the commit message or PR comment) since we use these to build the "change log" in a release.

olaurino:
Thanks @joleroi.

As a rule, we would like to see a regression test for each bug fix: a new test that doesn't pass without the fix and then passes with it. Changing an existing test is not what I would expect, but in this case I can probably live with that.

This PR does not have conflicts with the python3 branch, and it's also a rather small change, so it should be safe to merge it without having to wait for the release branch to be merged. I'll test the fix manually in CIAO (we are temporarily building CIAO daily builds from a different branch than usual while the Python 3 changes settle down).

I'll also update the PR description using the bug report.

DougBurke:
@olaurino - do you want to label this pr:merge?

joleroi:
Can you estimate the time scale when this is going to be in master? The spectral fitting code in ``gammapy`` does not work without this PR and I was wondering if I need to find a workaround in case it takes longer to get it merged.

DougBurke:
It's a holiday here, so there won't be a response from @olaurino until
tomorrow at the earliest

olaurino:
@joleroi from my point of view this could get merged. However, we had #195 in the 4.8.2 milestone. At this point, I would just like to make sure that #249 and #195 do not collide. So I am rebasing #195, resolving the conflicts with the new master, and then I'll try to merge them and make travis run on the resulting merge.

olaurino:
Unfortunately, it looks like there is an issue with #249 and #195. After rebasing #195, I created a temporary branch based off the most recent `master`, plus:
  * PR #263 (to make sure all tests pass on Python 3)
  * PR #195 
  * PR #249.

I pushed the branch ([`stat-changes`](https://github.com/sherpa/sherpa/tree/stat-changes)) so it could run through Travis, and there are some regressions:
https://travis-ci.org/sherpa/sherpa/builds/157951479

I couldn't look into it yet, though. Maybe the fix is trivial.

@DougBurke given that we are close to the 4.8.2 release, we probably need to re-prioritize #195 and #249.

Personally, I would also be more confident with the changes if we had specific tests that excercised the different requirements/scenario, directly.

DougBurke:
The error is - with a version of the code where I've changed the error message in `sherpa/include/sherpa/stat_extension.hh:wstatfct` to display the values being compared

```
        return _statfcts.calc_wstat_stat(data, model,
                                         extra['data_size'],
                                         extra['exposure_time'],
                                         extra['bkg'],
                                         extra['backscale_ratio'],
>                                        truncation_value)
E       TypeError: statistic input array sizes do not match: nelem/model/background/backscale=46/46/46/1024 data/exposure=1/2

sherpa/stats/__init__.py:606: TypeError
```

The data value is meant to be half the exposure value, so the problem is that the extra['backscale_ratio'] has a length of 1024 when it is expected to be 46. This is for the test

```
self = <test_stats.test_stats testMethod=test_wstat_calc_stat_info>

    def test_wstat_calc_stat_info(self):
        ui.load_pha("stat", self.make_path("3c273.pi"))
        ui.set_source("stat", ui.powlaw1d.p1)
        ui.set_stat("wstat")
>       ui.fit("stat")
```

(as there are two failures).

My guess is that this is because of the dataset being grouped. It is "natural" for wstat to be used with ungrouped data (and I think that's how @joleroi / gammapy are using it), but it's not wrong to use it with grouped data either. So the background data needs to account for both filtering and grouping.

DougBurke:
As part of my investigation into this, I wanted some regression tests that exercise the different problem cases with wstat. This has resulted in PR #267. It is not quite what I would really like here, as these are only regression tests and do not actually test that the numbers are being calculated correctly, but they do cover most of the cases we care about.

DougBurke:
I've tried merging in PR #267 to the `stat-changes` branch, and running *just* `sherpa/stats/test/test_wstat.py` I get

```
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_grouped_all FAILED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_grouped_filtered FAILED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_ungrouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_grouped_all FAILED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_grouped_filtered FAILED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_ungrouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_grouped_all xfail
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_grouped_filtered xfail
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_ungrouped_filtered FAILED
```

The last test fails because it's marked "xfail" but now passes (as expected, since this is bug #248), but as can be seen there's four other failures (which are all with grouped data), which all pass in the `master` branch.

DougBurke:
@joleroi - does PR #269 work for you (i.e. you get sensible results from it, and it doesn't error out)?

olaurino:
Closing this as it is superseded by #269 and #287.

joleroi:
@DougBurke @olaurino 
Sorry, I lost track of all the changes this issue triggered (I still don't know the sherpa code base well enough to really follow what's going on). Is there anything I should test at the moment?

DougBurke:
@joleroi - sorry, it's a messy issue, so hard to follow. @olaurino has just merged #287 which is my second attempt to fix the bug you originally found (i.e. #248) and some other related issues. This replaces my previous fix from #269 (which seemed to work for you https://github.com/sherpa/sherpa/issues/248#issuecomment-245910515 )

So, my claim is that the master branch should now work for you. If you could try it out and check my claim as soon as possible that would be wonderful (it's a holiday here Friday so there may not be much response from us until next week).

PR #250, opened 2016-08-26
Remove use of capsulethunk.h as do not support Python 2.6 or 3.0/1.

DougBurke:
Since we don't support the Python versions that do not provide the capsule API (i.e. Python 2.6 and Python 3.0/1) then I don't think we need the include file.


olaurino:
I'm closing this PR and opening another one against master.

PR #252, opened 2016-08-26
Release/4.9b1.5

olaurino:
Release branch to support CIAO 4.9 beta1.

It includes the Python 3 branch as well as a couple of independent commits.

Note that I had to revert PR #243 because it messed with some CIAO patches. It's something easy to fix, but I didn't want to change the patches as this point.

### DIFF
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -952,8 +952,7 @@ def get_pha_data(arg, make_copy=True, use_background=False):
         # Chandra Level 3 PHA files
         for ii in range(phadataset.get_ncrates()):
             block = phadataset.get_crate(ii + 1)
-            hduclas2 = block.get_key('HDUCLAS2')
-            if hduclas2 is not None and hduclas2.value == 'BKG':
+            if (_try_hdr_key(block, 'HDUCLAS2') == 'BKG'):
                 pha = block
 
     if pha is None or pha.get_colnames() is None:
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 #
-# Copyright (C) 2014, 2016  Smithsonian Astrophysical Observatory
+# Copyright (C) 2014  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -46,7 +46,6 @@ except:
         "Could not import setuptools.\n"
         "This might lead to an incomplete installation\n"
     ), file=sys.stderr)
-
 from numpy.distutils.core import setup
 
 from helpers.extensions import static_ext_modules
@@ -127,7 +126,6 @@ meta = dict(name='sherpa',
                 'Programming Language :: C',
                 'Programming Language :: Fortran',
                 'Programming Language :: Python :: 2.7',
-                'Programming Language :: Python :: 3.5',
                 'Programming Language :: Python :: Implementation :: CPython',
                 'Topic :: Scientific/Engineering :: Astronomy',
                 'Topic :: Scientific/Engineering :: Physics'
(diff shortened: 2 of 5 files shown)


DougBurke:
@olaurino I'm happy (and impatient :-) for this to be merged, with the remaining PRs against the `python3` branch to be added afterwards - that is #247 and #250 and readd #243. Is there anything left to be done before this can be merged?

olaurino:
To me, the only issue at this point is coordinating this with CIAO's CM. But I would argue that we could merge this stuff, rebase the release branch and then, as long as the release branch has no diffs with its previuos incarnation, we are good. I am working with this assumption in my local repo so I can fix more stuff for 4.8.2: I have a bunch of branches that look like master, release, and new topic branches after this PR has been merged. I'll have more info tomorrow.

olaurino:
Rebased. Original tip was d679f71.

olaurino:
@DougBurke I have to say when I was commenting on this PR I thought we were in the Python 3 discussion. Sorry for the confusion. This PR should be merged when the CIAO (internal) release is out. But now that #229 is merged, this PR has been rebased and now only has the few commits it should have had.

jbudynk:
Rebased. Original tip was 42b5da3.

sherpa/astro/io/crates_backend.py
```diff
@@ -952,8 +952,7 @@ def get_pha_data(arg, make_copy=True, use_background=False):
         # Chandra Level 3 PHA files
         for ii in range(phadataset.get_ncrates()):
             block = phadataset.get_crate(ii + 1)
-            hduclas2 = block.get_key('HDUCLAS2')
-            if hduclas2 is not None and hduclas2.value == 'BKG':
+            if (_try_hdr_key(block, 'HDUCLAS2') == 'BKG'):
```
olaurino: @jbudynk it would be good to have a test for this change. The issue only shows in CIAO but in a test for an external tool.

There are several options, and some require some infrastructure work, so part of this could be done in a separate PR.

One option is to create an integration test that requires a FITS backend. This is the easiest one, although in this case it could only be tested in CIAO. A better option would be to create a unit test for the crates backend, mock pycrates and make it return byte strings as it does now. This requires some preliminary work similar to what I did for the chips_backend. But even in this case we would need an integration test, too.

In any case we should not fix anything if we don't have a test in place to test the fix. When I review these fixes I checkout the PR, I temporarily remove the fix while keeping the tests, then I make sure that the tests fail. Then I checkout the PR again and run the tests, seeing that they pass.

I know you ran the tests in CIAO directly using an external test, but since this issue was not picked up by a Sherpa test in the first place, we need to add one in Sherpa, as we would with any other bug.

DougBurke: I hadn't realised this was found external to the sherpa tests, so +1 for having a test for this (especially as we have possible changes in the crates backend about the "string" type, so we want to make sure there's a test).

olaurino: And sure enough, the change in crates is coming right now. Crates should be returning strings soon :)
I asked that we have a CIAOX build with the crates changes but not the sherpa ones to compensate. I would like to see tests failing, and I'll probably add one test for this issue myself. Again, I want to see the new test fail before I push a fix for the next CIAOX build.

DougBurke: @olaurino - are we happy with the support for the crates changes? Is it safe to label this pr:merge now?

olaurino: Yes, we are happy.

Since this is the release branch for CIAO, and we are still not clear, technically we shouldn't merge this PR. I was thinking I would merge this branch in the 4.8.2 release branch, though, so we get these changes. What I expect is that if we merge 4.8.2 first (i.e. we release 4.8.2 before b1.5 is "released"), then CIAO4.9b1.5 will result merged as well, which is the same as if we merged it now. However, what's more likely to happen is that the b1.5 CIAO cut will be "released" before that. Then it's just a question of whether we want to rebase the 4.8.2 release branch or not. The double quotes in "released" (for CIAO) make me think we could do ad-hoc stuff in this specific case, but it's a good exercise nevertheless, and I am not a big fan of ad-hoc anyway :)

PR #253, opened 2016-08-29
make sure background is taken into account in calc_stat_info

### LINKED ISSUE #147: calc_stat_info broken with wstat

olaurino:
This PR fixes #147.

# Release Note
`calc_stat_info` call failed when `wstat` was selected, as the background was not taken into account. This issue has now been fixed.

# Notes
I did some minor refactoring, but I really dislike the `isinstance` usage. If stats need info from the `Fit` objects, then we should provide a reference to these object at least at the method level (if not in the constructor) and remove the `isinstance` procedural dispatching. However, this deeper refactoring is more risky, might break the established API, and should be independent of just fixing the bug.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -987,6 +987,24 @@ class Fit(NoNewAttributesAfterInit):
         """
         self.model.guess(*self.data.to_guess(), **kwargs)
 
+    def _calc_stat(self):
+        dep, staterror, syserror = self.data.to_fit(self.stat.calc_staterror)
+        model = self.data.eval_model_to_fit(self.model)
+        bkg_data = self._iterfit.get_bkg_data(dep)
+        valid_args = get_valid_args(self.stat.calc_stat)
+        if is_in('bkg', valid_args):
+            if isinstance(self.stat, UserStat):
+                return self.stat.calc_stat(dep, model, staterror,
+                                           syserror=syserror,
+                                           bkg=bkg_data['bkg'])
+            else:
+                return self.stat.calc_stat(dep, model, staterror,
+                                           syserror=syserror,
+                                           bkg=bkg_data)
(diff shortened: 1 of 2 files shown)


DougBurke:
Ideally we'd have tests of the lower-level routines (e.g. the `calc_stat_info` in the `Fit` class) that exercises this functionality (we'd still need the test you added, but at least we could more-reliably/cleanly check the results), but that's for a later PR. In other words, I'd like to know we had tests that check the numerical values returned by this code, but that shouldn't hold up this fix.

olaurino:
So, can I merge this?

DougBurke:
I was out yesterday. I've labelled this pr:merge but I have **not** been able to test this against the CIAO build.

One question I have is whether we should either include the bug number in the test name - e.g. `test_wstat_calc_stat_info_147` - or include it as a comment of the test, so that when we come back to the test and wonder "what is this doing" we have a reference. It's not a perfect scheme, but is a "quick and easy" scheme.

olaurino:
> One question I have is whether we should either include the bug number in the test name - e.g. test_wstat_calc_stat_info_147 - or include it as a comment of the test, so that when we come back to the test and wonder "what is this doing" we have a reference. It's not a perfect scheme, but is a "quick and easy" scheme.

That's a good point. I used to do it regularly, but then probably git and github spoiled me ;)

I'll add a comment so I don't change any actual code. I think it's fair to make sure there is one way or the other to track the bug a test is associated with: I wouldn't be too strict as to whether it's the method name or a comment, as there can be different scenarios in which they could both make sense.

PR #255, opened 2016-08-30
Add aliases for parameter names so to support new xspec names (fix #74)

### LINKED ISSUE #74: Should Xspec model parameter names be changed to match Xspec 12.9.0?
The 12.9.0 release of Xspec includes this [change](https://heasarc.gsfc.nasa.gov/xanadu/xspec/CHANGELOG.txt):

```
- Special characters have been removed from all model parameter names
and (generally) replaced with underscores.  This is to make their name
strings more easilty accessible in PyXspec.
```

Historically we have manually changed the parameter names so that they can be used as Python field names, but it does not look like there was a single set of rules or guide lines for the translation. One possibility is for us to use the new Xspec names, perhaps with temporary support for the existing names as aliases for the new names (to not break backward compatibility, but mark these as deprecated, if this is possible).

A diff of the `model.dat` file between 12.8.2a and 12.9.0 shows some of the changes (note that there are also changes to the limits, and a new model, but that's out of scope of this issue):

```
% diff heasoft-6.16/Xspec/src/manager/model.dat heasoft-6.17/Xspec/src/manager/model.dat 
254c254
< tau-y    " "  1.0   0.05    0.05    3.0       3.0       0.1

---
> tau_y    " "  1.0   0.05    0.05    3.0       3.0       0.1
256c256
< H/R_cyl " "  1.0   0.5     0.5     2.0       2.0       -1.

---
> HovR_cyl " "  1.0   0.5     0.5     2.0       2.0       -1.
295c295
< Ab>He    " "  1.0   0.1     0.1     10.       10.0      -0.01 

---
> Ab_met   " "  1.0   0.1     0.1     10.       10.0      -0.01 
345,346c345,346
< Lc/Ld    " "  0.1    0.      0.     10.       10.        1.e-2
< fin     " "   1.e-1  0.0    0.0     1.        1.        -1e-6

---
> LcovrLd  " "  0.1    0.      0.     10.       10.        1.e-2
> fin      " "   1.e-1  0.0    0.0     1.        1.        -1e-6
386c386
< l_h/l_s  " "  1.    1e-6    1e-6    1.e6      1.e6      0.01

---
> l_hovl_s " "  1.    1e-6    1e-6    1.e6      1.e6      0.01
389c389
< l_nt/l_h " "  0.5   0.      0.      0.9999    0.9999    0.01

---
> l_ntol_h " "  0.5   0.      0.      0.9999    0.9999    0.01
399c399
< Ab>He    " "  1.0   0.1     0.1     10.       10.0      -0.01 

---
> Ab_met   " "  1.0   0.1     0.1     10.       10.0      -0.01 
408c408
< l_h/l_s  " "  1.    1e-6    1e-6    1.e6      1.e6      0.01

---
> l_hovl_s " "  1.    1e-6    1e-6    1.e6      1.e6      0.01
411c411
< l_nt/l_h " "  0.5   0.      0.      0.9999    0.9999    0.01

---
> l_ntol_h " "  0.5   0.      0.      0.9999    0.9999    0.01
421c421
< Ab>He    " "  1.0   0.1     0.1     10.       10.0      -0.01 

---
> Ab_met   " "  1.0   0.1     0.1     10.       10.0      -0.01 
456c456
< <kT>    keV     1.0   0.0808  0.0808  79.9      79.9      0.01

---
> meankT   keV     1.0   0.0808  0.0808  79.9      79.9      0.01
460,465c460,465
< D       kpc    10.0  0.0   0.0      10000.     10000.   -1
< i       deg     0.0  0.0   0.0      90.0        90.0    -1.
< Mass    solar   1.0  0.0   0.0      100.0      100.0    0.01
< Mdot    1e18   1.0   0.0   0.0      100.0      100.0    0.01
< Tcl/Tef " "    1.7   1.0   1.0      10.0       10.0    -1.
< refflag " "    1.0   -1.0  -1.0     1.0        1.0     -1.

---
> D        kpc    10.0  0.0   0.0      10000.     10000.   -1
> i        deg     0.0  0.0   0.0      90.0        90.0    -1.
> Mass     solar   1.0  0.0   0.0      100.0      100.0    0.01
> Mdot     1e18   1.0   0.0   0.0      100.0      100.0    0.01
> TclovTef " "    1.7   1.0   1.0      10.0       10.0    -1.
> refflag  " "    1.0   -1.0  -1.0     1.0        1.0     -1.
485c485
< Tcol/Teff " "   1.5   1.0     1.0     2.0      2.0        -1.

---
> TcoloTeff " "   1.5   1.0     1.0     2.0      2.0        -1.
493c493
< lineE   keV     6.4   -0.1    0.1     0.1       100.0     100.0

---
> lineE   keV     6.4   0.1     0.1       100.0     100.0   -0.1
497c497
< a        " "    0.998 0.0     0.0     0.998     0.998      0.1 

---
> a        " "    0.998 0.01     0.01     0.998     0.998      0.1 
554a555,559
> nlapec         3  0.         1.e20           C_nlapec    add  0
> kT      keV     1.    0.008   0.008   64.0      64.0      .01
> Abundanc " "    1.    0.      0.      5.        5.        -0.001
> Redshift " "    0.   -0.999  -0.999   10.       10.       -0.01
> 
582c587
< redshift " "   0.1     1.0     1.0     1.5     2.0     0.1

---
> redshift " "   0.1   1.0e-5   1.0e-5   1.5     2.0     0.1
626c631
< logL/LEdd " "   -1.    -10.      -10.       2    2     0.01

---
> logLoLEdd " "   -1.    -10.      -10.       2    2     0.01
641c646
< logL/LEdd " "   -1.    -10.      -10.       2    2     0.01

---
> logLoLEdd " "   -1.    -10.      -10.       2    2     0.01
691c696
< HighECut keV    95.   0.01    1.      50.       200.     -0.01

---
> HighECut keV    95.   0.01    1.   100.       200.     -0.01
818c823
< He/H     " "    1.0   0.      0.      100.      100.      0.01

---
> HeovrH  " "    1.0   0.      0.      100.      100.      0.01
873c878
< <kT>    keV     1.0   0.0808  0.0808  79.9      79.9      0.01

---
> meankT  keV     1.0   0.0808  0.0808  79.9      79.9      0.01
1119c1124
< <kT>    keV     1.0   0.0808  0.0808  79.9      79.9      0.01

---
> meankT   keV     1.0   0.0808  0.0808  79.9      79.9      0.01
1426c1431
< E_B-V   " "    0.05  0.      0.      10.       10.       0.001

---
> E_BmV   " "    0.05  0.      0.      10.       10.       0.001
1512c1517
< E_B-V   " "    0.05  0.      0.      10.       10.       0.001

---
> E_BmV   " "    0.05  0.      0.      10.       10.       0.001
1563c1568
< lx/ld     " "   0.3   0.02    0.02      100    100        0.02

---
> lxovrld   " "   0.3   0.02    0.02      100    100        0.02
1580c1585
< "z"      " "    0.0   0.0     0.0     1.e5      1.e6      1.e-6

---
> z        " "    0.0   0.0     0.0     1.e5      1.e6      1.e-6
1584c1589
< E_B-V    " "   0.1     0.0     0.0     100.    100.     0.01

---
> E_BmV    " "   0.1     0.0     0.0     100.    100.     0.01
1623c1628
< E_B-V   " "    0.05  0.      0.      10.       10.       0.001

---
> E_BmV   " "    0.05  0.      0.      10.       10.       0.001
1628c1633
< E_B-V    " "   0.1     0.0     0.0     100.    100.     0.01

---
> E_BmV    " "   0.1     0.0     0.0     100.    100.     0.01
1712c1717
< Sig@6keV keV    1.00  0.0     0.0       10.     20.        .05

---
> Sig_6keV keV    1.00  0.0     0.0       10.     20.        .05
1748c1753
< Sig@6keV keV    1.00  0.0     0.0       10.     20.        .05

---
> Sig_6keV keV    1.00  0.0     0.0       10.     20.        .05
```

DougBurke:
# Release Note

The XSPEC models have been updated to use parameter names that match the new XSPEC naming scheme introduced in XSPEC 12.9.0. The old parameter name can still be used as aliases for the new name, but are deprecated and will be removed in a next major release. Note this allows any models to define aliases for their parameters.

# Details

The approach is not specific to the XSPEC models, but they are the only ones which need it at this time.

- [X] validate proof of concept
- [X] add tests for proof of concept
- [X] add deprecation warning
- [X] add tests to check that deprecation warning is created when expected
- [X] implement for XSPEC models
- [X] validate new XSPEC parameter names against latest release (compared against XSPEC `12.9.0o`).
- [X] add documentation to models to describe old parameter names
- [X] add documentation to model class describing approach
- [X] validate in end-user testing: interactive environment, including `save_all`
- [X] validate in end-user testing: script

# Issues

The warning message - since it uses the Python `DeprecationWarning` system - does not display to users by default. I thought it would appear, but I haven't seen it from some quick testing. I wonder if sherpa (in CIAO) should adjust the warnings status (or maybe if you import `sherpa.astro.ui`, but I'm a bit more concerned about that approach).

# Example

The XSbmc model has a parameter called `logA` in Sherpa 4.8 but it has been renamed to `log_A`; as shown below, both can be used (and `.logA` is essentially an alias for `.log_A`):

```
>>> ui.xsbmc.mdl
<XSbmc model instance 'xsbmc.mdl'>
>>> print(mdl)
xsbmc.mdl
   mdl.kT       thawed            1         0.01          100        keV
   mdl.alpha    thawed            1         0.01            4           
   mdl.log_A    thawed            0           -6            6           
>>> mdl.log_a.val
0.0
>>> mdl.log_a = 1.1
>>> mdl.log_a.val
1.1000000000000001
>>> mdl.loga
<Parameter 'log_A' of model 'mdl'>
>>> mdl.loga.val
1.1000000000000001
>>> mdl.loga = 2
>>> mdl.log_a.val
2.0
>>> print(mdl)
xsbmc.mdl
   mdl.kT       thawed            1         0.01          100        keV
   mdl.alpha    thawed            1         0.01            4           
   mdl.log_A    thawed            2           -6            6           
```

If you turn up the warnings you see a `DeprecationWarning`:

```
>>> import warnings
>>> warnings.filterwarnings('once')
>>> ui.xsbmc.mdl
<XSbmc model instance 'xsbmc.mdl'>
>>> print(mdl)
xsbmc.mdl
   mdl.kT       thawed            1         0.01          100        keV
   mdl.alpha    thawed            1         0.01            4           
   mdl.log_A    thawed            0           -6            6           
>>> mdl.loga = 2
sherpa/models/model.py:180: DeprecationWarning: Parameter name logA is deprecated for model XSbmc, use log_A instead
  warnings.warn(wmsg, DeprecationWarning)
>>> mdl.loga = 2
>>>
```

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1181,7 +1181,7 @@ class XSc6pvmkl(XSAdditiveModel):
     by using the exponential of the 6th order Chebyshev polynomial.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -1258,7 +1258,7 @@ class XSc6vmekl(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -1335,7 +1335,7 @@ class XScemekl(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -1397,7 +1397,7 @@ class XScevmkl(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -1588,7 +1588,7 @@ class XScompPS(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``tauy`` and ``HRcyl`` will be removed in the next release as
+              ``tauy`` and ``HRcyl`` might be removed in future releases as
               they have been replaced by ``tau_y`` and ``HovR_cyl``
               respectively, to match the XSPEC naming convention.
 
@@ -1801,7 +1801,7 @@ class XSdisk(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``NSmass`` will be removed in the next release as it has been
+              ``NSmass`` might be removed in future releases as it has been
               replaced by ``CenMass``, to match the XSPEC naming convention.
 
     Attributes
@@ -1852,7 +1852,7 @@ class XSdiskir(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``LcLd`` will be removed in the next release as it has been
+              ``LcLd`` might be removed in future releases as it has been
               replaced by ``LcovrLd``, to match the XSPEC naming convention.
 
     Attributes
@@ -1956,7 +1956,7 @@ class XSdiskline(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``RinM`` and ``RoutM`` will be removed in the next release as
+              ``RinM`` and ``RoutM`` might be removed in future releases as
               they have been replaced by ``Rin_M`` and ``Rout_M``
               respectively, to match the XSPEC naming convention.
 
@@ -2166,7 +2166,7 @@ class XSequil(XSAdditiveModel):
     particular the keyword "NEIVERS".
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``kT_ave`` will be removed in the next release as it has been
+              ``kT_ave`` might be removed in future releases as it has been
               replaced by ``meankT``, to match the XSPEC naming convention.
 
     Attributes
@@ -2358,7 +2358,7 @@ class XSgrad(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``TclTef`` will be removed in the next release as it has been
+              ``TclTef`` might be removed in future releases as it has been
               replaced by ``TclovTef``, to match the XSPEC naming convention.
 
     Attributes
@@ -2422,7 +2422,7 @@ class XSgrbm(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``temp`` will be removed in the next release as it has been
+              ``temp`` might be removed in future releases as it has been
               replaced by ``tem``, to match the XSPEC naming convention.
 
     Attributes
@@ -2535,7 +2535,7 @@ class XSkerrd(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``TcolTeff`` will be removed in the next release as it has been
+              ``TcolTeff`` might be removed in future releases as it has been
               replaced by ``TcoloTeff``, to match the XSPEC naming convention.
 
     Attributes
@@ -2682,7 +2682,7 @@ class XSlaor(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``RinG`` and ``RoutG`` will be removed in the next release as
+              ``RinG`` and ``RoutG`` might be removed in future releases as
               they have been replaced by ``Rin_G`` and ``Rout_G``
               respectively, to match the XSPEC naming convention.
 
@@ -2745,7 +2745,7 @@ class XSlaor2(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``RinG`` and ``RoutG`` will be removed in the next release as
+              ``RinG`` and ``RoutG`` might be removed in future releases as
               they have been replaced by ``Rin_G`` and ``Rout_G``
               respectively, to match the XSPEC naming convention.
 
@@ -2853,7 +2853,7 @@ class XSmeka(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -2957,7 +2957,7 @@ class XSmkcflow(XSAdditiveModel):
     on the cosmology settings set with ``set_xscosmo``.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -3416,7 +3416,7 @@ class XSnsmax(XSAdditiveModel):
     ``XSnsmaxg``.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``specfile`` will be removed in the next release as it has been
+              ``specfile`` might be removed in future releases as it has been
               replaced by ``_specfile``, to match the XSPEC naming convention.
 
     Attributes
@@ -3466,7 +3466,7 @@ class XSnsmaxg(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``specfile`` will be removed in the next release as it has been
+              ``specfile`` might be removed in future releases as it has been
               replaced by ``_specfile``, to match the XSPEC naming convention.
 
     Attributes
@@ -3522,7 +3522,7 @@ class XSnsx(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``specfile`` will be removed in the next release as it has been
+              ``specfile`` might be removed in future releases as it has been
               replaced by ``_specfile``, to match the XSPEC naming convention.
 
     Attributes
@@ -3864,7 +3864,7 @@ class XSplcabs(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``nmax`` and ``FAST`` will be removed in the next release as
+              ``nmax`` and ``FAST`` might be removed in future releases as
               they have been replaced by ``_nmax`` and ``_FAST``
               respectively, to match the XSPEC naming convention.
 
@@ -4289,7 +4289,7 @@ class XSsrcut(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``breakfreq`` will be removed in the next release as it has
+              ``breakfreq`` might be removed in future releases as it has
               been replaced by ``break_``, to match the XSPEC naming
               convention.
 
@@ -4464,7 +4464,7 @@ class XSvbremss(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``HeH`` will be removed in the next release as it has been
+              ``HeH`` might be removed in future releases as it has been
               replaced by ``HeovrH``, to match the XSPEC naming convention.
 
     Attributes
@@ -4514,7 +4514,7 @@ class XSvequil(XSAdditiveModel):
     particular the keyword "NEIVERS".
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``kT_ave`` will be removed in the next release as it has been
+              ``kT_ave`` might be removed in future releases as it has been
               replaced by ``meankT``, to match the XSPEC naming convention.
 
     Attributes
@@ -4569,7 +4569,7 @@ class XSvgnei(XSAdditiveModel):
     particular the keyword "NEIVERS".
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``kT_ave`` will be removed in the next release as it has been
+              ``kT_ave`` might be removed in future releases as it has been
               replaced by ``meankT``, to match the XSPEC naming convention.
 
     Attributes
@@ -4728,7 +4728,7 @@ class XSvmeka(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -4856,7 +4856,7 @@ class XSvmcflow(XSAdditiveModel):
     on the cosmology settings set with ``set_xscosmo``.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``switch`` will be removed in the next release as it has been
+              ``switch`` might be removed in future releases as it has been
               replaced by ``_switch``, to match the XSPEC naming convention.
 
     Attributes
@@ -5973,7 +5973,7 @@ class XSgabs(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``Tau`` will be removed in the next release as it has been
+              ``Tau`` might be removed in future releases as it has been
               replaced by ``Strength``, to match the XSPEC namin
               convention.
 
@@ -6269,7 +6269,7 @@ class XSredden(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``EBV`` will be removed in the next release as it has been
+              ``EBV`` might be removed in future releases as it has been
               replaced by ``E_BmV``, to match the XSPEC naming convention.
 
     Attributes
@@ -6437,7 +6437,7 @@ class XSswind1(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``logxi`` will be removed in the next release as it has been
+              ``logxi`` might be removed in future releases as it has been
               replaced by ``log_xi``, to match the XSPEC naming convention.
 
     Attributes
@@ -6657,7 +6657,7 @@ class XSuvred(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``EBV`` will be removed in the next release as it has been
+              ``EBV`` might be removed in future releases as it has been
               replaced by ``E_BmV``, to match the XSPEC naming convention.
 
     Attributes
@@ -6857,7 +6857,7 @@ class XSxion(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``lxld`` will be removed in the next release as it has been
+              ``lxld`` might be removed in future releases as it has been
               replaced by ``lxovrld``, to match the XSPEC naming convention.
 
     Attributes
@@ -6930,7 +6930,7 @@ class XSzdust(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``method`` and ``EBV`` will be removed in the next release as
+              ``method`` and ``EBV`` might be removed in future releases as
               they have been replaced by ``_method`` and ``E_BmV``
               respectively, to match the XSPEC naming convention.
 
@@ -7122,7 +7122,7 @@ class XSzxipcf(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``logxi`` will be removed in the next release as it has been
+              ``logxi`` might be removed in future releases as it has been
               replaced by ``log_xi``, to match the XSPEC naming convention.
 
     Attributes
@@ -7168,7 +7168,7 @@ class XSzredden(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``EBV`` will be removed in the next release as it has been
+              ``EBV`` might be removed in future releases as it has been
               replaced by ``E_BmV``, to match the XSPEC naming convention.
 
     Attributes
@@ -7212,7 +7212,7 @@ class XSzsmdust(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``EBV`` will be removed in the next release as it has been
+              ``EBV`` might be removed in future releases as it has been
               replaced by ``E_BmV``, to match the XSPEC naming convention.
 
     Attributes
@@ -7826,7 +7826,7 @@ class XScompth(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``AbHe`` will be removed in the next release as it has been
+              ``AbHe`` might be removed in future releases as it has been
               replaced by ``Ab_met``, to match the XSPEC naming convention.
 
     Attributes
@@ -8298,7 +8298,7 @@ class XSlogpar(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``pivotE`` will be removed in the next release as it has been
+              ``pivotE`` might be removed in future releases as it has been
               replaced by ``_pivotE``, to match the XSPEC naming convention.
 
     Attributes
@@ -8348,7 +8348,7 @@ class XSoptxagn(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``logLLEdd`` will be removed in the next release as it has
+              ``logLLEdd`` might be removed in future releases as it has
               been replaced by ``logLoLedd``, to match the XSPEC naming
               convention.
 
@@ -8437,7 +8437,7 @@ class XSoptxagnf(XSAdditiveModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``logLLEdd`` will be removed in the next release as it has
+              ``logLLEdd`` might be removed in future releases as it has
               been replaced by ``logLoLedd``, to match the XSPEC naming
               convention.
 
@@ -8746,7 +8746,7 @@ class XSheilin(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``redshift`` will be removed in the next release as it has
+              ``redshift`` might be removed in future releases as it has
               been replaced by ``z``, to match the XSPEC naming
               convention.
 
@@ -8845,7 +8845,7 @@ class XSzbabs(XSMultiplicativeModel):
     The model is described at [1]_.
 
     .. note:: Deprecated in Sherpa 4.10.0
-              ``redshift`` will be removed in the next release as it has
+              ``redshift`` might be removed in future releases as it has
               been replaced by ``z``, to match the XSPEC naming
               convention.
 
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -7817,25 +7817,28 @@ class XSeqtherm(XSAdditiveModel):
         XSAdditiveModel.__init__(self, name, (self.l_hovl_s, self.l_bb, self.kT_bb, self.l_ntol_h, self.tau_p, self.radius, self.g_min, self.g_max, self.G_inj, self.pairinj, self.cosIncl, self.Refl, self.Fe_abund, self.Ab_met, self.T_disk, self.xi, self.Beta, self.Rin, self.Rout, self.redshift, self.norm))
 
 
+# It is not obvious from the XSPEC documentation what theta, showbb,
+# and RefOn are.
+#
 class XScompth(XSAdditiveModel):
     """The XSPEC compth model: Paolo Coppi's hybrid (thermal/non-thermal) hot plasma emission models.
 
     The model is described at [1]_.
 
+    .. note:: Deprecated in Sherpa 4.9.0
+              ``AbHe`` will be removed in the next release as it has been
+              replaced by ``Ab_met``, to match the XSPEC naming convention.
+
     Attributes
     ----------
-    l_hl_s
-        The ratio of the hard to soft compactness, l_h / l_s.
-    l_bb
-        The soft photon compactness.
+    theta
+    showbb
     kT_bb
         The temperature of the blackbody if greater than 0.
         When less than zero then the absolute value is used as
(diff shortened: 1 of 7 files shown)


DougBurke:
Rebased (main areas of overlap were Python 3 support and model documentation). Original tip was 6553b5f89bf74fd1fa885a5db1edb5e234be0e98

sherpa/models/model.py
```diff
@@ -121,16 +169,64 @@ def __str__(self):
...
+                        return val
+
+            rpars = self.__dict__.get('_renamedpars')
+            if rpars is not None:
+                for (oname, nname) in rpars:
+                    if oname.lower() == lname:
+                        # Easier to call itself with the proper name
+                        # than to loop through __dict__ again. This
+                        # should not be a common case.
+                        ans = getattr(self, nname)
```
olaurino: Lines 226, 227, and 228 are not covered by the tests. Since this code is recursive it would be better to test it.

sherpa/models/model.py
```diff
@@ -121,16 +169,64 @@ def __str__(self):
     def __iter__(self):
         return iter([self])
 
-    # Make parameter access case insensitive
     def __getattr__(self, name):
```
olaurino: There is a lot of complexity in this method. At the very least, I would suggest it be broken down into smaller pieces to aid readability and maintainability. On the other hand, wouldn't it be much easier to just build a dictionary of key value pairs upon instantiation? The keys would be the acceptable (lower case) names and the values would be the parameters, then `__getattr__` would simply be a lookup in the dictionary. If it is safer to defer this action after the instantiation, for whatever reason, I would create the map the first time `__getattr__` is called and use it during subsequent calls.

Even better, why not overriding `__setattr__` to add `Parameter`s to an index when they are added in the initializer? I would find this closer to the design I would have expected for parameters all along, i.e. to be declared as class members (with instance values) rather than instance members. In any case, I would then implement a method for declaring alternative names, which takes the alternative name and the reference to the parameter itself as arguments. I am less concerned about the mechanism by which alternative names are registered to trigger a warning, which is more of an implementation detail.

sherpa/models/tests/test_model.py
```diff
@@ -149,11 +158,77 @@ def test_complex_expression(self):
...
+        return self._basemodel.calc(*args, **kwargs)
+
+
+class test_model_parametercase_instance(test_model_renamed):
+
+    def setUp(self):
+        self.m = ParameterCase()
+
+    def test_name(self):
+        self.assertEqual(self.m.name, 'parametercase')
```
olaurino: Warnings are not actually tested, are they?

DougBurke: No. I added them to `sherpa/conftest.py` to make sure they didn't fail the tests, but have no idea how we are supposed to check if a warning fired or not.

sherpa/models/model.py
```diff
@@ -121,16 +169,64 @@ def __str__(self):
...
+                        for par in pars:
+                            if par.name == nname:
+                                warn(oname, nname)
+                                return par
+
+        else:
+            for key in self.__dict__:
+                try:
+                    kname = key.lower()
+                except AttributeError:
```
DougBurke: Looking at the code, it appears that I feared that an element of the objects dictionary might not be a string (or, rather, might not have a `lower` method), and so wanted to guard against it. I don't know if it's possible for a non-string key to be used in `__dict__`

olaurino:
@DougBurke, should we address the issues I highlighted and get them done for 4.9.1?

DougBurke:
I had a brief discussion with @anetasie earlier today about the two XSPEC related PRs (so this and #299) and our conclusion was that they could be held back from this release (i.e. drop the 4.9.1 milestone from them).

olaurino:
Trying new Resolve Conflict feature in GitHub. Original tip was at 5f95714.

olaurino:
I don't like it, it merges the master branch into the feature branch rather than rebasing the feature branch over the master branch. I think I'll just force-push from my local repo, where I had already rebased and resolved the conflict.

sherpa/models/model.py
olaurino: While it is possible to have non strings as keys, keys of the special `self.__dict__` dictionary should always be strings. If they are not, something is wrong in the code or how the client is using it, so I wouldn't worry about that case. I am currently attempting a refactoring that could greatly simplify the code, and I am going to remove this check, which also means I won't add a test case where `__dict__` has a non-string key.

sherpa/models/tests/test_model.py
```diff
@@ -149,11 +158,77 @@ def test_complex_expression(self):
...
+
+    def calc(self, *args, **kwargs):
+        for par in self.pars:
+            setattr(self._basemodel, par.name, par.val)
+
+        self._basemodel.integrate = self.integrate
+        return self._basemodel.calc(*args, **kwargs)
+
+
+class test_model_parametercase_instance(test_model_renamed):
```
olaurino: @DougBurke I guess it is intentional that this test, as well as the one it's extending, are all running the same set of core tests, as they are run with different custom models. But I wanted to double check :)

sherpa/models/model.py
olaurino: I think you just had to:
  * introduce the renamed parameters before accessing the parameters
  * access a parameter through its alias rather than through the "canonical name"

to trigger that code, at least that's what I think I am doing now. I want to make sure the tests cover the necessary cases before I test the refactoring.

sherpa/astro/xspec/__init__.py
```diff
@@ -874,18 +874,28 @@ class XSbmc(XSAdditiveModel):
...
+    log_A
         The log of the A parameter: see [1]_ for more details.
     norm
         The normalization of the model: see [1]_ for an explanation
         of the units.
 
+    Notes
+    -----
+    In Sherpa 4.9.0 the ``logA`` parameter has been renamed to ``log_A``.
+    It can still be accessed using ``log_A`` for the time being, but this
```
olaurino: I believe it should say `logA` on this line. Also, the version should be changed from Sherpa 4.9.0 to Sherpa 4.10.0 (we still have to fix the ciao/sherpa versioning scheme, I'll send a proposal for that). And a few lines above, I am relaxing the statement "`logA` will be removed in the next release" with a "might be removed in future releases".

sherpa/models/tests/test_model.py
DougBurke: The idea of this test is to see what happens when the parameter names - e.g. the second argument of the `Parameter` constructor - does not have the same case as the attribute name it is stored with. I forget now which XSPEC model had/has this, but I wanted to make sure we tested this case. So, I want the same tests run, on essentially the same model that is set up slightly differently. As you can see from the `calc` method, it is basically just passing the values through to the `Sin` model.

Does this answer your question?

sherpa/models/tests/test_model.py
DougBurke: I've got a better handle now on how I'd check the warnings now - e.g. https://github.com/sherpa/sherpa/pull/383/files#diff-4b23a06bd3b4b5489b772aaccd41ac42 - but I'm happy for you to add these tests as you see fit ;-)

olaurino: That also allows you to find issues in my tests, good move ;)

sherpa/astro/xspec/__init__.py
DougBurke: Good catch (for the `log_A` vs `logA` confusion on line 896).

sherpa/models/tests/test_model.py
```diff
@@ -149,11 +158,77 @@ def test_complex_expression(self):
...
             self.assertEqual(m(self.x), func(self.m(self.x)))
+
+
+# Test support for renamed parameters by sub-classing
+# the Sin model (which lets the tests be re-used).
+#
+class RenamedPars(Sin):
+
+    def __init__(self, name='renamedpars'):
+        self._renamedpars = [('norm', 'ampl')]
```
olaurino: Hmm... is this really a case we want to support? I.e. that a model might be an extension of an existing model, just defining a rename?

The approach I was hoping we could take is for `Parameter` to be aware of its *deprecated aliases*. However, this test case seems to make that approach a dead end, in that it's actually the `Model`'s responsibility to define the deprecated aliases. There is a redundancy in that being defined as instance attributes (not a great idea to begin with, imo), parameters have two names: one defined by the model, i.e. the name of the attribute, and one being the name of the parameter being passed to the parameter's initializer. So we are in a grey area where knowing the parameter's name is a responsibility of both classes.

What breaks the tie is whether we want to support this test case or not, which I don't really find too compelling, especially since we are introducing this "renaming" business ourselves, and I think that making it a parameter's responsibility to know of its aliases (which the model then needs to index, though) makes for a cleaner design.

DougBurke: a) I could come up with a hand-wavy case for allowing renames, but it would not be very strong

b) this test case is - I think - in part to support some of the tests I wrote which use a renamed class to make writing the tests less onerous (switching to pytest parameterize may avoid this; I haven't thought that through)

c) how would a user/code accessing "model.deprecatedparametername" get passed on to the correct parameter? Would you have to query each parameter and ask it "do you recognize this alias"?

olaurino: a) I think it's an arbitrary decision, then. Do we want it or not?

b) which is why in my yet-to-push rework I am changing the implementation of the `RenamedPars` model to inherit the methods from `Sin` but then I short-circuited `Sin` altogether in the initializer. Rewriting the tests in `pytest` would be useful, but at this point I would leave them alone so we don't risk changing the coverage.

c) a `Parameter` has an `aliases` attribute with the deprecated aliases. The model uses it to build a dictionary name/alias -> parameter which is how `__getattr__` returns the right parameter. If the requested attribute has a name in the list of aliases for that parameter, then the warning is issued. I still have to update the xspec models to use my new stuff, but maybe I should publish it first. And by the way, tests should be failing but don't... I am wondering how much of the renamed parameters we should test in the xspec models.

DougBurke: a) I suggest publishing what you have, so we can see / try it out (it would be nice to get it into the CIAO daily build for this).

b) I agree that re-writing the tests to use pytest is not for this PR (or, whatever PR gets used)

c) I feel we have to test some of the renamed parameter stuff. It may not need to be attached to the XSPEC models, since this is intended as a general purpose feature, so as long as the inheritance works okay we don't need to try every possible model combination. Or am I missing part of the problem?

olaurino: Did you get notified of my PR against your repo?

DougBurke: I don't tend to look at the email address that I use gor my github account during work hours. Let me check. I got an email about PR #384, and emails about these comments (including 'Did you get notified of my PR against your repo).

olaurino: @DougBurke I issued a PR against your repo a few months ago, but I don't see it merged. Please let me know what to do. Should I just force push my rework, issue a new PR, or wait for you to review the rework the way I had pushed it?

olaurino: Wait, I think I misinterpreted my notes. It's been a while. Let me double check.

olaurino: I think I just have to close this PR in favor of #384. @DougBurke do you agree?

olaurino: I am taking it as a yes and closing this PR.

olaurino:
Closing in favor of #384.

sherpa/models/tests/test_model.py
olaurino: I didn't misinterpret my notes after all. I had issued a PR against yours so you could sign off the changes, which you never did explicitly... I'll reopen this PR and push all my commits on top of it, then I'll close the other PR which is not against master.

olaurino:
@DougBurke as for this issue:

> The warning message - since it uses the Python DeprecationWarning system - does not display to users by default. I thought it would appear, but I haven't seen it from some quick testing. I wonder if sherpa (in CIAO) should adjust the warnings status (or maybe if you import sherpa.astro.ui, but I'm a bit more concerned about that approach).

If we want the warning to appear to the user by default maybe we should use a different kind of warning type. Otherwise we should treat it as all the other deprecation warnings, which are not displayed by default. Or am I missing something?

olaurino:
Rebasing, tip was e6d3ffe

olaurino:
For the record, here is an email I got from @DougBurke with the explanation of some changes we reverted back:

Here's the current set of parameter renames/aliases from Omar's PR, based on my original PR.

My suggestions are

a) we keep the aliases for those I've labelled "I think this is okay"

b) virtually all the "potential problem" cases, we actually remove the alias (and so there's actually no change for these).

This is for those parameters whose model.dat lines begin with $ or * (these are the characters I replaced with _). So, we would leave the switch parameter as "switch" and not rename it to "_switch".

c) I think this leaves us with one case in the "potential problem" list, which is the break parameter. I am pretty sure that break is a Python reserved symbol and so we can't have a class attribute/descriptor/whatever they're called be named "break". We have handled this so far by having it called "breakfreq". The current suggestion is to use the name "break_" (and leave breakfreq as an alias). I would leave it like this (ie call it break_ and leave an alias of breakfreq), but I do not have a strong opinion on this.

# Potential problem cases

```python
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 1, 0, 1, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 1, 0, 1, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._specfile = Parameter(name, '_specfile', 1200, 0, 1.0e6, 0, 1.0e6, alwaysfrozen=True, aliases=["specfile"])
self._specfile = Parameter(name, '_specfile', 1200, 0, 1.0e6, 0, 1.0e6, alwaysfrozen=True, aliases=["specfile"])
self._specfile = Parameter(name, '_specfile', 6, 0, 1.0e6, 0, 1.0e6, alwaysfrozen=True, aliases=["specfile"])
self._nmax = Parameter(name, '_nmax', 1, alwaysfrozen=True, aliases=["nmax"])
self._FAST = Parameter(name, '_FAST', 0, alwaysfrozen=True, aliases=["FAST"])
self.break_ = Parameter(name, 'break_', 2.42E17, 1.E15, 1.E19, 0.0, hugeval, 'Hz', aliases=["breakfreq"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 1, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._method = Parameter(name, '_method', 1, 1, 3, 1, 3, alwaysfrozen=True, aliases=["method"])
self._energy00 = Parameter(name, '_energy00', 0.5, min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy00"])
self._energy01 = Parameter(name, '_energy01', 1., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy01"])
self._energy02 = Parameter(name, '_energy02', 1.5, min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy02"])
self._energy03 = Parameter(name, '_energy03', 2., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy03"])
self._energy04 = Parameter(name, '_energy04', 3., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy04"])
self._energy05 = Parameter(name, '_energy05', 4., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy05"])
self._energy06 = Parameter(name, '_energy06', 5., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy06"])
self._energy07 = Parameter(name, '_energy07', 6., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy07"])
self._energy08 = Parameter(name, '_energy08', 7., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy08"])
self._energy09 = Parameter(name, '_energy09', 8., min=0.0, max=10.0, units='keV', alwaysfrozen=True, aliases=["energy09"])
self._redshift = Parameter(name, '_redshift', 0.0, alwaysfrozen=True, aliases=["redshift"])
self._model = Parameter(name, '_model', 0, 0, 1, 0, 1, alwaysfrozen=True, aliases=["model"])
self._lyman_limit = Parameter(name, '_lyman_limit', 1, 0, 1, 0, 1, alwaysfrozen=True, aliases=["lyman_limit"])
self._switch = Parameter(name, '_switch', 2, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._switch = Parameter(name, '_switch', 2, 0, 2, 0, 2, alwaysfrozen=True, aliases=["switch"])
self._pivotE = Parameter(name, '_pivotE', 1.0, units='keV', alwaysfrozen=True, aliases=["pivotE"])
```


Most of the following come about because the XSPEC model.dat file starts the name with a $, to indicate
it's not a fittable parameter, and my original code just converted $ to _ rather than ignoring it.
There's also '*' used by XSPEC (I forget what for):

```
  grep '^[^a-zA-Z]' ~/local/u35/ciao-4.9/ots/spectral/manager/model.dat
  $switch    1
  $switch    1
  $switch    1
  $switch    1
  $switch    1     0       0     1      1       -1
  $switch    1     0       0     1      1       -1
  *energy00    keV    0.5   0. 0.   10.  10.  -1.
  *energy01    keV    1.    0. 0.   10.  10.  -1.
  *energy02    keV    1.5   0. 0.   10.  10.  -1.
  *energy03    keV    2.    0. 0.   10.  10.  -1.
  *energy04    keV    3.    0. 0.   10.  10.  -1.
  *energy05    keV    4.    0. 0.   10.  10.  -1.
  *energy06    keV    5.    0. 0.   10.  10.  -1.
  *energy07    keV    6.    0. 0.   10.  10.  -1.
  *energy08    keV    7.    0. 0.   10.  10.  -1.
  *energy09    keV    8.    0. 0.   10.  10.  -1.
  $switch         2
  *pivotE keV 1.0
  $switch    1     0       0    1         1           -1
  $switch    1     0       0     1      1       -1
  $specfile 1200
  $specfile 1200
  $specfile 6
  *nmax   " "   1
  *FAST    " "  0
  $switch         2
  $switch    1
  $switch    1
  $method   " "   1       1       1       3       3       -0.01
  *redshift " "     0.0
  $model    " "     0
  $lyman_limit " "  1
  *nregions " "   1.
  *fracexpo " "   1.
```
Some of the above are for models we don't support (e.g. the last two - nregions and fracexpo - are from the XSPEC pileup model which we do not support).

The only onw I notice is "break_", which I think I had hard-coded, since break is a Python keyword. It is used in
one model:

```
  grep break ~/local/u35/ciao-4.9/ots/spectral/manager/model.dat
  breakE   keV 10.    0.1     0.1     1000.     1000.     0.1
  breakE   keV 10.    0.1     0.1     1000.     1000.     0.1
  Rbreak   " "    20.   1.235   1.235   400.      400.      -0.1
  break  Hz   2.42E17 1.E10   1.E15     1.E19    1.E25    1E10 
  Rbreak    " "   20.   1.235   1.235   400.      400.      -0.1
```

# I think these are okay

```python
self.log_A = Parameter(name, 'log_A', 0.0, -6.0, 6.0, -hugeval, hugeval, aliases=["logA"])
self.tau_y = Parameter(name, 'tau_y', 1.0, 0.05, 3.0, 0.0, hugeval, aliases=["tauy"])
self.HovR_cyl = Parameter(name, 'HovR_cyl', 1.0, 0.5, 2.0, 0.0, hugeval, frozen=True, aliases=["HRcyl"])
self.CenMass = Parameter(name, 'CenMass', 1.4, .4, 10., 0.0, hugeval, units='Msun', frozen=True, aliases=["NSmass"])
self.LcovrLd = Parameter(name, 'LcovrLd', 0.1, 0., 10., 0.0, hugeval, aliases=["LcLd"])
self.Rin_M = Parameter(name, 'Rin_M', 10., 6., 1000., 0.0, hugeval, frozen=True, aliases=["RinM"])
self.Rout_M = Parameter(name, 'Rout_M', 1000., 0., 1000000., 0.0, hugeval, frozen=True, aliases=["RoutM"])
self.meankT = Parameter(name, 'meankT', 1.0, 0.0808, 79.9, 0.0, hugeval, 'keV', aliases=["kT_ave"])
self.TclovTef = Parameter(name, 'TclovTef', 1.7, 1.0, 10.0, 0.0, hugeval, frozen=True, aliases=["TclTef"])
self.tem = Parameter(name, 'tem', +300., +50., +1000., 0.0, hugeval, 'keV', aliases=["temp"])
self.TcoloTeff = Parameter(name, 'TcoloTeff', 1.5, 1.0, 2.0, 0.0, hugeval, frozen=True, aliases=["TcolTeff"])
self.r_br_g = Parameter(name, 'r_br_g', 6.0, 1.0, 400., 0.0, hugeval, frozen=True, aliases=["r_brg"])
self.Rin_ms = Parameter(name, 'Rin_ms', 1.0, 1.0, 400., 0.0, hugeval, frozen=True, aliases=["Rinms"])
self.Rout_ms = Parameter(name, 'Rout_ms', 400., 1.0, 400., 0.0, hugeval, frozen=True, aliases=["Routms"])
self.Rin_G = Parameter(name, 'Rin_G', 1.235, 1.235, 400., 0.0, hugeval, frozen=True, aliases=["RinG"])
self.Rout_G = Parameter(name, 'Rout_G', 400., 1.235, 400., 0.0, hugeval, frozen=True, aliases=[""])
self.Rin_G = Parameter(name, 'Rin_G', 1.235, 1.235, 400., 0.0, hugeval, frozen=True, aliases=["RinG"])
self.Rout_G = Parameter(name, 'Rout_G', 400., 1.235, 400., 0.0, hugeval, frozen=True, aliases=["RoutG"])
self.HeovrH = Parameter(name, 'HeovrH', 1.0, 0., 100., 0.0, hugeval, aliases=["HeH"])
self.meankT = Parameter(name, 'meankT', 1.0, 0.0808, 79.9, 0.0, hugeval, 'keV', aliases=["kT_ave"])
self.meankT = Parameter(name, 'meankT', 1.0, 0.0808, 79.9, 0.0808, 79.9, 'keV', aliases=["kT_ave"])
self.Strength = Parameter(name, 'Strength', 1.0, 0., 1.e6, 0.0, hugeval, aliases=["Tau"])
self.E_BmV = Parameter(name, 'E_BmV', 0.05, 0., 10., 0.0, hugeval, aliases=["EBV"])
self.log_xi = Parameter(name, 'log_xi', 2.5, 2.1, 4.1, 0.0, hugeval, aliases=["logxi"])
self.E_BmV = Parameter(name, 'E_BmV', 0.05, 0., 10., 0.0, hugeval, aliases=["EBV"])
self.lxovrld = Parameter(name, 'lxovrld', 0.3, 0.02, 100, 0.0, hugeval, aliases=["lxld"])
self.E_BmV = Parameter(name, 'E_BmV', 0.1, 0.0, 100., 0.0, hugeval, aliases=["EBV"])
self.log_xi = Parameter(name, 'log_xi', 3, -3, 6, -hugeval, hugeval, aliases=["logxi"])
self.E_BmV = Parameter(name, 'E_BmV', 0.05, 0., 10., 0.0, hugeval, aliases=["EBV"])
self.E_BmV = Parameter(name, 'E_BmV', 0.1, 0.0, 100., 0.0, hugeval, aliases=["EBV"])
self.l_hovl_s = Parameter(name, 'l_hovl_s', 1., 1e-6, 1.e6, 0.0, hugeval, aliases=["l_hl_s"])
self.l_ntol_h = Parameter(name, 'l_ntol_h', 0.5, 0., 0.9999, 0.0, hugeval, aliases=["l_ntl_h"])
self.Ab_met = Parameter(name, 'Ab_met', 1.0, 0.1, 10., 0.0, hugeval, frozen=True, aliases=["AbHe"])
self.l_hovl_s = Parameter(name, 'l_hovl_s', 1., 1e-6, 1.e6, 0.0, hugeval, aliases=["l_hl_s"])
self.l_ntol_h = Parameter(name, 'l_ntol_h', 0.5, 0., 0.9999, 0.0, hugeval, aliases=["l_ntl_h"])
self.Ab_met = Parameter(name, 'Ab_met', 1.0, 0.1, 10., 0.0, hugeval, frozen=True, aliases=["AbHe"])
self.Ab_met = Parameter(name, 'Ab_met', 1.0, 0.1, 10., 0.0, hugeval, frozen=True, aliases=["AbHe"])
self.logLoLEdd = Parameter(name, 'logLoLEdd', -1., -10., 2, -hugeval, hugeval, aliases=["logLLEdd"])
self.logLoLEdd = Parameter(name, 'logLoLEdd', -1., -10., 2, -hugeval, hugeval, aliases=["logLLEdd"])
self.z = Parameter(name, 'z', 0.0, -1.0e-3, 1.0e5, -1.0e-3, 1.0e5, aliases=["redshift"])
self.n = Parameter(name, 'n', 1.e-5, 0.0, 1.0e6, 0.0, 1.0e6, '10^22 atoms / cm^2', aliases=["nHeI"])
self.z = Parameter(name, 'z', 0.0, -1.0e-3, 1.0e5, -1.0e-3, 1.0e5, aliases=["redshift"])
self.z = Parameter(name, 'z', 0.0, 0.0, 1.0e5, 0.0, 1.0e6, aliases=["redshift"])
```

sherpa/models/model.py
```diff
@@ -81,6 +83,48 @@ def cache_model(cls, pars, xlo, *args, **kwargs):
...
+    as object attributes. That is, if the model contains parameters
+    called ``breakFreq`` and ``norm``, and the instance is stored in
+    the variable ``mdl``, then the following can be used to access
+    the parameters::
+
+        print("Break frequency = {}".format(mdl.breakfreq))
+
+        mdl.norm = 1.2e-3
+
+    Support is provided to allow parameter names to be changed,
```
DougBurke: @olaurino - I think this paragraph (about `_renamedpars`) is from my version of the code. Should it be removed or replaced now we have your implementation?

sherpa/models/tests/test_model.py
```diff
@@ -20,9 +20,12 @@
 import logging
 import operator
 import numpy
-from sherpa.utils import SherpaFloat, SherpaTestCase
+import warnings
+from sherpa.utils import SherpaTestCase
```
DougBurke: This should probably be updated to

```
from sherpa.utils.testing import SherpaTestCase
```

sherpa/models/model.py
olaurino: Argh... I thought I had looked for instances of the string `_renamedpars`. Well, at this point the model class could tell the user something about the aliases, but it should probably link to the documentation for `Parameter` to describe the `aliases` argument. The `Parameter` class seems to still have the "old-style" unformatted docstrings, though.

DougBurke: My approach has been to try and add relevant documentation when it impacts the code I've been touching, and when I notice the need. I obviously failed to notice the lack of Parameter documentation when writing this ;-)

sherpa/astro/xspec/__init__.py
```diff
@@ -7517,6 +7871,13 @@ class XSzigm(XSMultiplicativeModel):
         Should photoelectric absorption be included (1), or not (0).
         This parameter can not be thawed.
 
+    Notes
```
DougBurke: Note to self: this note is to be removed.

sherpa/models/model.py
DougBurke: I've removed this paragraph. I'm not sure how best to describe support for parameter aliases, so have left it out for now.

sherpa/models/tests/test_model.py
DougBurke: I've cleaned up this module so this has been done.

sherpa/astro/xspec/__init__.py
DougBurke: Done.

DougBurke:
a) warning message

I think leave as is for now.

b) docs

I believe I have updated the docstrings to match the current behavior (for the models); the low-level Parameter class has a small amount of documentation but nothing significant.

c) does the summary of this PR need changing now the implementation has changed?

d) I think I've finished with all I need to do on this PR. Am I missing anything?

PR #257, opened 2016-08-30
fix docstring for levmar tolerance

### LINKED ISSUE #128: possible documentation mismatch for optimization methods
Some docstrings for the optimization methods seem to be out of sync with the actual code. Most likely this happened because the code was changed at some point but the documentation was not, or only partially updated, resulting in the mismatch.

In particular, the docstrings claim that the default values for some parameters are the square root of the double precision epsilon, although this does not seem to be true, at least not for all of the optimization methods. For instance, in the `LevMar` method docstring it says that the default for `ftol` is `sqrt(DBL_EPSILON) ~ 1.19209289551e-07`. There are several issues with this:
  * `1.192..e-07` is not the double precision (assuming 64 bit floating point) epsilon, but the float32 one.
  * `1.192..e-07` is indeed the epsilon itself, not its square root.
  * the `LevMar` implementation seems to be using the float32 epsilon as its default, rather than its square root. So may other methods.

See https://github.com/sherpa/sherpa/blob/master/sherpa/optmethods/__init__.py#L351

vs

https://github.com/sherpa/sherpa/blob/master/sherpa/optmethods/optfcts.py#L37 and https://github.com/sherpa/sherpa/blob/master/sherpa/optmethods/optfcts.py#L368

olaurino:
Fix #128.

# Release Note

The documentation string for the Levenberg-Marquardt optimization function now correctly states that the parameter default values are equal to the single precision epsilon, rather than the square root of the double precision epsilon.

### DIFF
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -372,8 +372,8 @@ class LevMar(OptMethod):
        free parameters.
     epsfcn : number
        This is used in determining a suitable step length for the
-       forward-difference approximation; default is FLT_EPSILON ~ 1.19209289551e-07,
-       where DBL_EPSILON is the smallest number
+       forward-difference approximation; default is FLT_EPSILON
+       ~ 1.19209289551e-07, where FLT_EPSILON is the smallest number
        x such that `1.0 != 1.0 + x`. This approximation assumes that
        the relative errors in the functions are of the order of
        `epsfcn`. If `epsfcn` is less than the machine precision, it is
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -348,21 +348,21 @@ class LevMar(OptMethod):
     ----------
     ftol : number
        The function tolerance to terminate the search for the minimum;
-       the default is sqrt(DBL_EPSILON) ~ 1.19209289551e-07, where
-       DBL_EPSILON is the smallest number x such that `1.0 != 1.0 +
+       the default is FLT_EPSILON ~ 1.19209289551e-07, where
+       FLT_EPSILON is the smallest number x such that `1.0 != 1.0 +
        x`. The conditions are satisfied when both the actual and
        predicted relative reductions in the sum of squares are, at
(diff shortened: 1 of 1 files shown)


sherpa/optmethods/__init__.py
```diff
@@ -372,8 +372,8 @@ class LevMar(OptMethod):
        free parameters.
     epsfcn : number
        This is used in determining a suitable step length for the
-       forward-difference approximation; default is sqrt( DBL_EPSILON
-       ) ~ 1.19209289551e-07, where DBL_EPSILON is the smallest number
+       forward-difference approximation; default is FLT_EPSILON ~ 1.19209289551e-07,
+       where DBL_EPSILON is the smallest number
```
DougBurke: should this be DBL_EPSILON or FLT_EPSILON? I'm guessing replace DBL_EPSILON by FLT_EPSILON as done elsewhere (e.g. the changes for `gtol`).

I have tried hard to keep the docstring lines "short", so they can be read easily in an 80-column
terminal. Can line 375 be broken up at the ~ (e.g. as done in line 364)?

DougBurke: I should note that by "80-column limit" I really mean "a somewhat random limit that can vary per docstring and depends on what version of emacs and pyflakes/pep8 I have available, but is somewhere near 70-80 ish".

olaurino: I have been working with a 120 char limit in mind (and in the tools config), but I guess you are saying that we want to keep the docstring at the 70-80ish level, right?

DougBurke: I'm targeting a "nice" display of the docstring in an 80-column terminal, so ideally docstrings shouldn't be too long. It is obviously less-than ideal for heavily indented docstrings, so it's not a hard limit. When editing an existing docstring I tend to just try and go with the current line-length limits, unless I'm willing to re-edit the whole thing.

olaurino:
I addressed the comments, so as long as Travis is happy, I am going to merge this PR.

PR #258, opened 2016-08-30
add some docs about other fortran compilers

### LINKED ISSUE #5: Explicitly mention gfortran requirement for installation
The current install docs don't call out the requirement for `gfortran`.  I initially tried `g77` on Mac Mavericks and this failed.

olaurino:
Fix #5.

# Release Note
Added documentation regarding support of different `fortran` compilers. While we only officially support `gfortran`, we have an example on how to patch Sherpa's `setup.py` to use a different compiler with a platform-dependent configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,13 @@ There are two ways to download the source code:
 
  - directly from GitHub as a git repository
 
+**NOTE:** it is possible to build Sherpa with `fortran` compilers other than
+`gfortran`. While this is not supported, [PR #202](https://github.com/sherpa/sherpa/pull/202/files)
+shows how this can been accomplished with `g95` on `OS X` in a specific setup.
+Similar changes are probably required for other compilers or setups.
+Since the `fortran` code is compiled by `numpy`'s [`f2py`](http://docs.scipy.org/doc/numpy-dev/f2py/distutils.html)
+you might want to check their documentation for details.
+
 ### 2a. Extract the source tarball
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +238,13 @@ There are two ways to download the source code:
 
  - directly from GitHub as a git repository
 
+**NOTE:** it is possible to build Sherpa with `fortran` compilers other than
+`gfortran`. While this is not supported, [PR #202](https://github.com/sherpa/sherpa/pull/202/files)
+shows how this can been accomplished with `g95` on `OS X` in a specific setup.
+Similar changes are probably required for other compilers or setups.
+Since the `fortran` code is compiled by `numpy`'s [`f2py`](http://docs.scipy.org/doc/numpy-dev/f2py/distutils.html)
+you might want to check their documentation for details.
```
DougBurke: It would be preferable if "their documentation" was a link to whatever documentation you mean. Or perhaps re-write to something like

The Fortran code is compiled using `f2py` as described in the 
[numpy documentation](http://docs.scipy.orf/doc/numpy-XXXX/yyy).

where the link should probably be to a release version rather than to numpy-dev (or is that the release version, I haven't checked).

olaurino: The tricky part is that *their documentation* in this context is rather sparse, fragmented, and it may not work. I'll go with your suggestion.

olaurino:
I addressed the comments, so I am assuming I can merge the PR.

Issue #106, opened 2015-09-20
axis differences between plot_source() and plot('source') and issues with model plots


DougBurke:
# Summary

The data and axis labels are not the same for
- `plot_source()` compared to `plot('source')`
- `plot_model()` compared to `plot('model')`

when using PHA data sets. It's not clear whether it's the same issue for the two, but we need to look at what the `plot` command is doing.
# Details

The following is seen with the ChIPS backend (CIAO 4.7) and master with the `master` branch. I've seen it with multiple PHA data sets.

```
% ipython --pylab
Python 2.7.10 |Continuum Analytics, Inc.| (default, Sep 15 2015, 14:50:01) 
Type "copyright", "credits" or "license" for more information.

IPython 4.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
Using matplotlib backend: Qt4Agg

In [1]: from sherpa.astro import ui
/home/naridge/local/anaconda/envs/sherpa-astropy-fits-gzip/lib/python2.7/site-packages/IPython/kernel/__init__.py:13: ShimWarning: The `IPython.kernel` package has been deprecated. You should import from ipykernel or jupyter_client instead.
  "You should import from ipykernel or jupyter_client instead.", ShimWarning)
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available

In [2]: ui.load_pha(1, 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi') 
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273.pi' 
but not used; to use them, re-read with use_errors=True
read ARF file sherpa-test-data/sherpatest/threads/pha_intro/3c273.arf
read RMF file sherpa-test-data/sherpatest/threads/pha_intro/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273_bg.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/threads/pha_intro/3c273_bg.pi' 
but not used; to use them, re-read with use_errors=True
read background file sherpa-test-data/sherpatest/threads/pha_intro/3c273_bg.pi

[...]

PR #261, opened 2016-09-02
[skip ci] wip: unit test for plotting and bug 106 **DO NOT MERGE**

olaurino:
This PR is very work in progress. It just adds (failing) unit tests for some plotting functions. Tests are not even properly factored, they are just there to illustrate how to capture issue #106 in unit tests.


DougBurke:
So, we would like to see support for tests for both chips and matplotlib backends. It may not cover all the tests included in this PR.

PR #263, opened 2016-09-06
fix #262 list_data_ids() fails on py3 with mixed id types

### LINKED ISSUE #262: Issue with daset IDs of different types on py3
A `wstat` test is failing on Python 3. This is due to sorting not working as expected.

Here is a way to reproduce the issue:
    
    from sherpa.ui.utils import Session
    session = Session()
    session.load_arrays(1, [1, 2, 3], [1, 2, 3])
    session.load_arrays("bar", [1, 2, 3], [1, 2, 3])
    assert [1, "bar"] == session.list_data_ids()

And the error:
```
========================================================================== FAILURES ==========================================================================
_______________________________________________________________________ test_list_ids ________________________________________________________________________

    def test_list_ids():
        session = Session()
        session.load_arrays(1, [1, 2, 3], [1, 2, 3])
        session.load_arrays("bar", [1, 2, 3], [1, 2, 3])
>       assert [1, "bar"] == session.list_data_ids()

sherpa/ui/tests/test_session.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sherpa.ui.utils.Session object at 0x7ff785ee0a20>

    def list_data_ids(self):
        """List the identifiers for the loaded data sets.
    
            Returns
            -------
            ids : list of int or str
               A list of the data set identifiers that have been created
               by commands like `load_data` and `load_arrays`.
    
            See Also
            --------
            delete_data : Delete a data set by identifier.
            load_arrays : Create a data set from arrays of data.
            load_data : Create a data set from a file.
    
            Examples
            --------
    
            In this case only one data set has been loaded:
    
            >>> list_data_ids()
            [1]
    
            Two data sets have been loaded, using string identifiers:
    
            >>> list_data_ids()
            ['nucleus', 'jet']
    
            """
        # TODO3: I have left the explicit copy in (the "[:]" suffix)
        # when converting to Python 3, but it is probably unnescessary
        keys = list(self._data.keys())[:]
>       keys.sort()
E       TypeError: unorderable types: str() < int()

sherpa/ui/utils.py:2254: TypeError

```

Since the issue is resulting from a PR that has been merged, and is now affecting master, I am giving this issue the highest priority.

olaurino:
Fix #262 

# Release Note
Sherpa was sorting the list of dataset IDs in a non-python3 compliant fashion, which resulted in issues when using strings and integers together as dataset IDs. This has now been fixed.

### DIFF
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -1,0 +1,28 @@
+#
+#  Copyright (C) 2016  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+from sherpa.ui.utils import Session
+
+
+# bug #262
+def test_list_ids():
+    session = Session()
+    session.load_arrays(1, [1, 2, 3], [1, 2, 3])
+    session.load_arrays("bar", [1, 2, 3], [1, 2, 3])
+    assert [1, "bar"] == session.list_data_ids()
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -18,11 +18,14 @@
 #
 
 from sherpa.ui.utils import Session
+from numpy.testing import assert_array_equal
 
 
 # bug #262
 def test_list_ids():
     session = Session()
     session.load_arrays(1, [1, 2, 3], [1, 2, 3])
-    session.load_arrays("bar", [1, 2, 3], [1, 2, 3])
-    assert [1, "bar"] == session.list_data_ids()
+    session.load_arrays("1", [1, 2, 3], [4, 5, 6])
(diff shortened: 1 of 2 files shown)


DougBurke:
I was a bit concerned since we currently allow - but do not document - the use of both x and str(x) as different data set identifiers, where x is an integer. It looks like this is still possible with this patch, but perhaps we should add some tests (like the following) to check this still works (if only to let us know if we ever break it):

```
% git rev-parse HEAD
94dda526b78be5ba32d441d22a1c90ccff7eea0e

% ipython
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) 

IPython 5.1.0 -- An enhanced Interactive Python.
%quickref -> Quick reference.

In [2]: ui.load_arrays(1, [1,2,3], [10,20,30])

In [3]: ui.list_data_ids()
Out[3]: [1]

In [4]: ui.load_arrays("1", [10,20,30], [-10,-20,-30])

In [5]: ui.list_data_ids()
Out[5]: [1, '1']

In [6]: ui.get_data(1).y.min()
Out[6]: 10

In [7]: ui.get_data('1').y.min()
Out[7]: -30

```

DougBurke:
I haven't checked this in our internal build, but have labeled it pr:merge.

olaurino:
@DougBurke I pushed a new commit with tests similar to what you suggested. I removed the pr:merge label while the Travis builds run. If you are happy with the changes I'll merge it as soon as Travis is happy (hoping it's happy ;)).

olaurino:
The order of `1` and `'1'` is not guaranteed, so some Travis builds are failing. I pushed a new commit to compare IDs as sets rather than lists.

DougBurke:
An in a L3 meeting, so haven't looked in detail but it sounds sensible

DougBurke:
I'm happy with this.

Issue #227, opened 2016-06-21
Issues using wstat when grouping/filtering data


anetasie:
I was just grouping/filtering for plotting of the results of fitting using `wstat`. The arrays get confused and  I get errors. See below. This should be seen for standard X-ray PHA files and analysis. I used CSC1 data set here.

```
load_data('acisf06114_000N002_r0025_pha3.fits.gz')
set_model('xsphabs.abs1*powlaw1d.p1')
set_method('simplex')
set_stat('wstat')
notice(0.5,7.)
fit()

group_counts(10)
subtract()
#plot_fit()

fit() # correct error
unsubtract()
fit() # Array size mismatch
```

Here is the screen output with the error:

>   File "/soft/ciao-4.8/ots/lib/python2.7/site-packages/IPython/core/interactiveshell.py", line 3032, in run_code
>     exec(code_obj, self.user_global_ns, self.user_ns)
>   File "<ipython-input-13-749b6e6de45a>", line 1, in <module>
>     fit()
>   File "<string>", line 1, in fit
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/astro/ui/utils.py", line 9259, in fit
>     self._fit(id, _otherids, *_kwargs)
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/astro/ui/utils.py", line 9363, in _fit
>     res = f.fit(*_kwargs)
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/fit.py", line 58, in run
>     result = func(fit, *args, *_kwargs)
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/fit.py", line 1106, in fit
>     init_stat = self.calc_stat()
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/fit.py", line 1020, in calc_stat
>     bkg=bkg_data)[0]
>   File "/soft/ciao-4.8/lib/python2.7/site-packages/sherpa/stats/__init__.py", line 601, in calc_stat
>     truncation_value)
> TypeError: statistic input array sizes do not match

DougBurke:
On a related note (found when trying some variations of this bug), should `calc_stat` fail with an error on subtracted data - i.e. in the following `fit` errors out but `calc_stat` does not.

```
% python
Python 2.7.12 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:42:40) 
[...]

PR #267, opened 2016-09-08
add wstat tests

DougBurke:
# Summary

Add several regression tests for wstat.

# Details

Issue #248 pointed out some problems with wstat (in particular, when used with a dataset with a non-scalar backscal value that had been filtered). The simple fix to this (PR #249) has caused problems elsewhere. In order to help disentangle things, this test suite is a set of simple regression tests, intended to exercise the various options, in particular:

- grouped vs ungrouped
- filtered vs unfiltered
- scalar vs array for the BACKSCAL value

Three of the tests are marked as "Expected Fail", and cover the cases that are fixed in #249 as well as the cases that are causing problems with #249 (which are from tests added to the master branch in PR #195 and being tested as part of #249).

# Notes

I have included PR #263 in this PR in order to give the tests a chance on Travis. This can be rebased once 263 has been merged.

### DIFF
--- a/sherpa/stats/tests/test_wstat.py
+++ b/sherpa/stats/tests/test_wstat.py
@@ -1,0 +1,406 @@
+#
+#  Copyright (C) 2016  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+# Explicit tests of wstat, related to issue 248
+# https://github.com/sherpa/sherpa/issues/248
+# The intention is the functionality of these tests will be moved
+# elsewhere, but for now it is useful to have some testing of
+# the various options.
+#
+# At present the statistic tests are not broken up into unit and
+# integration tests, as most of the functionality is tested either
+# in integration tests (directly or implicitly).
+#
+# These tests are regression tests, designed to check
+# that the wstat statistic can be used with PHA data sets:
+#
+# a) one data set (backscal is a scalar)
+#   a.1) ungrouped, no filter
+#   a.2) ungrouped, energy filter
+#   a.3) ungrouped, wavelength filter
+#   a.4) grouped, no filter
+#   a.5) grouped, energy filter
+#   a.6) grouped, wavelength filter
+#
+#   The filters should be complex (e.g. multiple ignore regions)
+#   but this part is not exercised.
+#
+# b) multiple data sets (backscal is a scalar)
+#
+# Repeat a but for multiple spectra.
+#
+# c) as a but with backscal being a vector
+#
+# d) as b but with backscal being a vector
+#
+# Note that for test a a previously-grouped PHA file is used, whereas
+# for test b one of the files is grouped on the fly.
+#
+# TODO:
+#   - replace the data used for test c with actual grating data
+#
+#   - replace unittest with py.test
+#
+#   - replace with a synthetic data set so the test can be run
+#     without data or the IO module; however, this version does test
+#     the calculation when the responses (i.e. ARF and RMF) are
+#     included.
+#
+#   - validate some/all of the statistic values to check the
+#     answer is meaningful
+#
+#   - should there be any tests that include quality flags?
+#
+#   - to test out if issue 227 is fixed
+#     https://github.com/sherpa/sherpa/issues/227
+#     Actually, DougBurke believes this is due to the behavior
+#     of group_counts, which we already work around below,
+#     search for "Note: this is related to issue 227".
+#
+
+import numpy as np
+
+from sherpa.utils import SherpaTestCase, requires_data, requires_fits
+from sherpa.astro import ui
+
+from unittest import expectedFailure
+
+import logging
+
+logger = logging.getLogger('sherpa')
+
+
+@requires_data
+@requires_fits
+class test_wstat_single_scalar(SherpaTestCase):
+    """Single PHA file with a scalar backscal value"""
+
+    def setUp(self):
+
+        # defensive programming (one of the tests has been seen to fail
+        # when the whole test suite is run without this)
+        ui.clean()
+
+        self._old_logger_level = logger.getEffectiveLevel()
+        logger.setLevel(logging.ERROR)
+
+        ui.set_stat('wstat')
+
+        infile = self.make_path('3c273.pi')
+        ui.load_pha(1, infile)
+
+        ui.set_source(1, ui.powlaw1d.pl)
+
+        # The powerlaw slope and normalization are
+        # intended to be "a reasonable approximation"
+        # to the data, just to make sure that any statistic
+        # calculation doesn't blow-up too much.
+        #
+        ui.set_par("pl.gamma", 1.782)
+        ui.set_par("pl.ampl", 1.622e-4)
+
+    def tearDown(self):
+        ui.clean()
+
+        try:
+            logger.setLevel(self._old_logger_level)
+        except AttributeError:
+            pass
+
+    def _filter_data(self):
+        """Filter the data.
+
+        Use a slightly-complex filter - e.g. not just the
+        end points - to exercise the system.
+        """
+
+        ui.ignore(None, 0.5)
+        ui.ignore(3, 4)
+        ui.ignore(7, None)
+
+    def _check_stat(self, nbins, expected):
+
+        # check the filter sizes (mainly so that these tests
+        # get flagged as in need of a look if anything changes
+        # in other parts of the code, such as filtering and binning
+        #
+        self.assertEqual(nbins, ui.get_data().get_dep(True).size)
+
+        stat = ui.calc_stat()
+        self.assertAlmostEqual(expected, stat, places=7)
+
+    def test_wstat_grouped_all(self):
+
+        # Used git commit 770359b5004374b969ebb63c173f293419397b4c
+        # to create the oracle value, on a linux 64-bit machine.
+        self._check_stat(46, 62.12275005769554)
+
+    def test_wstat_grouped_filtered(self):
+        self._filter_data()
+
+        # Used git commit 770359b5004374b969ebb63c173f293419397b4c
+        # to create the oracle value, on a linux 64-bit machine.
+        self._check_stat(35, 39.84715083397662)
+
+    def test_wstat_ungrouped_all(self):
+        ui.ungroup()
+
+        # Used git commit 770359b5004374b969ebb63c173f293419397b4c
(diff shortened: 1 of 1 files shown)


DougBurke:
Nope. I disagree with the possibility that these tests could have failed.

I'll have to look at it tomorrow to see why one test failed with

```
_______________ test_wstat_single_scalar.test_wstat_grouped_all ________________

self = <test_wstat.test_wstat_single_scalar testMethod=test_wstat_grouped_all>

    def test_wstat_grouped_all(self):

    

        # Used git commit 770359b5004374b969ebb63c173f293419397b4c

        # to create the oracle value, on a linux 64-bit machine.

>       self._check_stat(46, 62.12275005769554)

sherpa/stats/tests/test_wstat.py:142: 

sherpa/stats/tests/test_wstat.py:136: in _check_stat

    self.assertAlmostEqual(expected, stat, places=7)

E   AssertionError: 62.12275005769554 != 124.23384899235153 within 7 places
```

but not on my local build. It is suspicious that the expected value appears to be half the calculated value.

DougBurke:
Apparently

```
% python setup.py  test -a sherpa/stats/tests/test_wstat.py
```

passes

```
% python setup.py  test
```

fails. :-1:

joleroi:
For me (locally) both 
```python setup.py  test -a sherpa/stats/tests/test_wstat.py```
and
```python setup.py test
```
pass

DougBurke:
@joleroi - thanks for checking. In your tests of just `test_wstat.py`, are the tests run or skipped?

sherpa/stats/tests/test_wstat.py
```diff
@@ -0,0 +1,394 @@
...
+        # The powerlaw slope and normalization are
+        # intended to be "a reasonable approximation"
+        # to the data, just to make sure that any statistic
+        # calculation doesn't blow-up too much.
+        #
+        ui.set_par("pl.gamma", 1.782)
+        ui.set_par("pl.ampl", 1.622e-4)
+
+    def tearDown(self):
+        ui.clean()
```
olaurino: Have you tried adding `ui.clean()` to the `setUp` method as well?

More generally, we now can rely on `py.test`, so writing tests should barely require any boilerplate. We don't need to care about logging and startup/shutdown operations can be factored into common fixtures, which can be applied automatically.

Other than that, I agree that we should have better control on the data in unit tests so we can precisely calculate the expected output. Possibly, we should not rely on disk I/O as much as possible, which should also speed up execution of unit tests. Something along the lines of the tests added by #261 (which however haven't been properly refactored yet).

joleroi:
> In your tests of just test_wstat.py, are the tests run or skipped?

```
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_grouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_grouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_scalar::test_wstat_ungrouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_grouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_grouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_two_scalar::test_wstat_ungrouped_filtered PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_grouped_all <- ../miniconda3/envs/headversions/lib/python2.7/unittest/case.py xfail
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_grouped_filtered <- ../miniconda3/envs/headversions/lib/python2.7/unittest/case.py xfail
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_ungrouped_all PASSED
sherpa/stats/tests/test_wstat.py::test_wstat_single_array::test_wstat_ungrouped_filtered <- ../miniconda3/envs/headversions/lib/python2.7/unittest/case.py xfail
```

sherpa/stats/tests/test_wstat.py
DougBurke: I think the `ui.clean` suggestion is just the ticket; let's see if Travis agrees. I don't really understand what setting there could be which would change the statistic calculation in this case, but that's for another day.

I have stuck to `unittest` since that's what I'm familiar with. I'm happy to switch to `py.test` but maybe when there's a bit less of a time crunch ;-)

olaurino: > I don't really understand what setting there could be which would change the statistic calculation in this case,

I guess one would need to look at the logs of the failing build and bisect the tests run before the offending test until one finds the test responsible for the failure. As I mentioned in a separate conversation, this behavior could be benign (tests should be isolated nevertheless) or uncover an underlying, maybe unrelated issue.

olaurino:
Differences in test results may depend on the order in which tests are run, which is platform-dependent. This is always an issue: tests should be isolated and not interfere with each other, but sometimes it might uncover deeper issues.

Now that tests pass, do we have a better understanding of why #195 and #249 are not playing nicely together?

joleroi:
Yes, the fix I added in https://github.com/sherpa/sherpa/pull/249 does not take into account grouped data.
For a scalar backscal value you get an array of len(noticed_channels) while you should get an array of len(groups)
```
ratio = vectorize_backscale_ratio(bkg.backscal,
                                                       mydata.backscal,
                                                       len(mydata.channel))
ratio = ratio[noticed_channels-1] 
```
This should be very easy to fix. I can have a look.

joleroi:
Since @DougBurke added a test for filtered data in https://github.com/sherpa/sherpa/pull/267 and it passes - doesn't that solve https://github.com/sherpa/sherpa/issues/248 and thus make https://github.com/sherpa/sherpa/pull/249 superfluous?

DougBurke:
(short answer: I still think we need #249, it just needs to also handle grouping]

Just got back from dropping the kids off at school. As noted in the comments to this PR there's a couple of combinations to look at - i.e. it's not just "does it work with filtered data", but "does it work with [filtered|unfiltered], [grouped|ungrouped], [scalar|array] backscal" data.

I wrote this test to see how we were doing in these various combinations, and as various fixes come in hopefully we can remove the `expectedFailure` decorators until everything passes.

Note that there's also #227 which I want to add tests for, but this may be an issue higher-up in the session code rather than directly in the stats code.

joleroi:
Ah sorry I was looking for ``pytest.mark.xfail`` I didn't know about the ``expectedFailure`` decorator :see_no_evil: . So you're right that we need #249 . However, thinking about it again, I have no idea how to handle grouping in the case of an array-like backscale ratio for WStat. If the backscale ratio varies within one group you need to find a way to average ``backscal_source / backscal_background``. In gammapy we do this by taking the 'background counts weighted average' but I don't know if you want to go that way (given that probably few people will use WStat with grouped data).

DougBurke:
Well, here's a comment from the [XSPEC page on the wstat statistic](https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixStatistics.html)

> In practice, it works well for many cases but for weak sources can generate an obviously wrong best fit. It is not clear why this happens although binning to ensure that every bin contains at least one count often seems to fix the problem.

so we could see how XSPEC handles this case.

Another option would be to explicitly error out in this case, so that users get a (hopefully) comprehensible error message rather than a generic "array sizes do not match") error, and hope we can fix up the code to remove the need for the error sooner rather than later. If we went this route we would have to agree on what functionality we want to support now, and what we are happy to wait until later to have.

joleroi:
> Another option would be to explicitly error out in this case, so that users get a (hopefully) comprehensible error message rather than a generic "array sizes do not match") error 

:+1:

DougBurke:
I've jotted down some thoughts on background grouping in https://github.com/sherpa/sherpa/issues/227#issuecomment-245600743 - one of the proposed solutions there is for the wstat code to just ignore the grouping information on the background dataset and to use the source grouping. I am unsure about this - there are definite advantages, particularly for users of the `ui` layer, but I haven't thought if it complicates things for uses of the `object` API.

DougBurke:
Looking at `DataPHA` in `sherpa/astro/data.py`, we have 

```
    def _check_scale(self, scale, group=True, filter=False):
        if numpy.isscalar(scale) and scale <= 0.0:
            scale = 1.0
        elif numpy.iterable(scale):
            scale = numpy.asarray(scale, dtype=SherpaFloat)
            if group:
                if filter:
                    scale = self.apply_filter(scale, self._middle)
                else:
                    scale = self.apply_grouping(scale, self._middle)

            scale[scale <= 0.0] = 1.0
        return scale

    def get_backscal(self, group=True, filter=False):
        backscal = self.backscal
        if backscal is not None:
            backscal = self._check_scale(backscal, group, filter)
        return backscal

...

    def apply_grouping(self, data, groupfunc=numpy.sum):
        """

        Apply the data set's grouping scheme to the array data,
        combining the grouped data points with groupfunc, and return
        the grouped array.  If the data set has no associated grouping
        scheme or the data are ungrouped, data is returned unaltered.

        """
        if (data is None) or (not self.grouped):
            return data

        groups = self.grouping
        filter = self.quality_filter
        if filter is None:
            return do_group(data, groups, groupfunc.__name__)

        if (len(data) != len(filter) or len(groups) != len(filter)):
            raise DataErr('mismatch', "quality filter", "data array")

        filtered_data = numpy.asarray(data)[filter]
        groups = numpy.asarray(groups)[filter]
        grouped_data = do_group(filtered_data, groups, groupfunc.__name__)

        if data is self.channel and groupfunc is self._make_groups:
            return numpy.arange(1, len(grouped_data) + 1, dtype=int)

        return grouped_data

...

    @staticmethod
    def _middle(array):
        array = numpy.asarray(array)
        return (array.min() + array.max()) / 2.0
```

so it looks like our general approach to grouping the BACKSCAL array is to use the mean value of the BACKSCAL values in a group. I haven't thought this through whether it's sensible for the statistic. Also, if we went this way where this should be done (since the data object is not sent into the guts of the statistic code, but to be honest the information flow in this part of the code is opaque, at best).

DougBurke:
From a conversation with @xspector 

> In general, XSPEC calculates grouped values of backscal and areascal by a flux-weighted mean unless the data is identified NET in which case a simple mean is used.

but it's also pointed out one should not be using grouped data with what we refer to as wstat.

DougBurke:
TODO: need to review behavior of wstat when the dataset has multiple backgrounds (e.g. `BACKGROUND_UP` and `BACKGROUND_DOWN` for Sherpa grating data). From a *very quick* look at the data it may just be using the first component, but I need to check further.

olaurino:
We should rebase this to remove the #263 commits. Other than that, as we discussed during our meeting, this should be merged.

I have mixed feelings about using `unittest.expectedFailure` rather than `pytest.mark.xfail`, however the new tests still use `unittest`, so I guess that's good for consistency.

I would vote for merging this, as long as it's rebased and Travis keeps the green light (which probably means the `[ci skip]` tag should be removed from the last commit message).

DougBurke:
Rebased against master. Original tip was 4f691545512aef201fa65591fbd1b1ca85c3af18

I took the opportunity to squash the commits.  I have also add regression tests for issue #227 which I developed as part of #269 (these add two new `expectedFailure` cases). It's nice to get a test in that demonstrates the problem before any fix is added.

I did look at `py.test` but quickly decided I don't have the time to learn how to use it and still get things in for 4.8.2.

DougBurke:
So @olaurino - if the travis ci tests remain green, are you happy for me to merge this this evening?

olaurino:
yes!

PR #268, opened 2016-09-09
Fix docstring for sherpa.astro.ui.get_bkg_scale

DougBurke:
Corrects the documentation for the `sherpa.astro.ui.get_bkg_scale` function by removing the `bkg_id` parameter and clarifying that the return value can be either a scalar or an array.


DougBurke:
Rebased from master. Original tip was 5e2916bd0d21853d211e111c41af4a0f127e839b

This is to pick up the fix from PR #263 so that the tests pass.

DougBurke:
Rebased from master. Original tip was 05ff8a45bb61e5d471626de4b3fa03ed7a16e3dd

This is to revert PR #263 since that's now in master.

PR #269, opened 2016-09-09
fix for wstat issues #227 #248

DougBurke:
**EDITED TO ADD** my preferred approach to fixing this problem is in PR #287 - which changes how data is passed from the fit to the stat object.
# Summary

Support a variety of data options when using the `wstat` statistic, including grouped or ungrouped data, filtered or unfiltered data, and scalar or vector backscale arrays. This includes #227 and #248.

I am not 100% convinced this is the approach to take - since it "forces" the background data to match the grouping of the source data set. This is arguably the right thing to do for `wstat`, but it may not be for other statistics, and the routine where this is being calculated is meant to be somewhat general (as far as I can tell).

There is also the possibility that the code does not correctly support datasets with multiple background datasets, but this is outside the scope of this PR (i.e. if it is broken here, it is broken in `master`).

This PR is intended for review and discussion. I am not convinced it will be accepted, but the code changes are fairly small.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -24,8 +24,11 @@ import logging
 import os
 import signal
 from numpy import arange, array, abs, iterable, sqrt, where, \
-    ones_like, isnan, isinf, float, float32, finfo, nan, any, zeros, append, \
-    int, ones
+    ones_like, isnan, isinf, float, float32, finfo, nan, any, \
+    append, int, ones
+
+import numpy as np
+
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
     bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import FitErr, EstErr, SherpaErr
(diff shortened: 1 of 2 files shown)


DougBurke:
The build failure here is due to #262 since I did not have the energy to pull in #263 here.

DougBurke:
Rebasing on upstream master. Original tip was
fd0c344c1911e5b2a1e7b7334a97fd586b30aa6b      
(assuming I've done this right).

DougBurke:
The following can be thought of as "I think the approach used here is okay in the short term, since I think the overall approach needs re-working to move logic into the place where it should be, rather than stuck in the Fit object".

With some thought (thanks to having to prepare the kids their school lunches), I think that the logic in `IterFit.get_bkg_data` (and other parts of `IterFit`/`Fit`) for calculating the arrays to send to the `Stat` object should be pushed into `Stat`. That is, `Stat` should accept `Data` objects rather than arrays. This means that the Stat object can access all the information it could need, rather than having to second-guess in `Fit`/`IterFit`.

I would imagine that some of the higher-level classes in the `Stat` hierarchy would be responsible for this - e.g. `sherpa.stats.Likelihood` and `sherpa.stats.Chi2` - with subclasses (e.g. `sherpa.stats.Wstat`) overriding as needed.

There's some planning (and thought) needed for this, but it seems like a much-more sensible place to have this logic.

dtnguyen2:
I haven't looked at sherpa for a while, but I think Primini and sigma-rejection are also embedded within the fit class.

DougBurke:
As a user, the primini and sigma-rejection methods are presented as a separate piece of functionality - e.g. using the `set_iter_method` routine rather than `set_stat`. I can understand this - the primini case is particularly complicated to know how to handle well - but it would be interesting to see if/how well they could all be separated out. The current approach of having `IterFit` and `Fit` classes looks clunky at first glance, but there's a lot going on in `sherpa/fit.py` and I really don't understand it all!

DougBurke:
I plan to rebase this over the week-end, as #195 has now been merged (and hopefully #267 too). Actually, it may be more of a rewrite than a rebase; we'll see.

DougBurke:
Rebased - let's see how I did.

DougBurke:
@olaurino I am looking into the re-write the Stats API to see if it makes sense, but ran out of time this week-end. About the only thing I did to was write a couple of "unit tests" for the Fit class to act as a regression test when I go about rewriting things. It's my first py.test style test, so was wondering if it made sense. It's at

https://github.com/DougBurke/sherpa/blob/feature/rewrite-stat/sherpa/tests/test_fit_unit.py

DougBurke:
I may have found at least one case where this approach doesn't work (filtered and grouped data), but I need to investigate more. I am more convinced that the logic should be moved into the stats class, and am working on code for this, but I haven't got further than committing the tests.

DougBurke:
Closed in favor of #287

Issue #260, opened 2016-09-02
Test fails on Python 3


cdeil:
Sorry if I've brought this up before!
If so, please close and point me to the answer.

I have Sherpa master installed, with Python 3.5 from Macports and latest stable Setuptools, Numpy, Astropy, Pytest 2.9.

From the git repo:

```
$ python setup.py test
...
    [Errno 13] Permission denied: '/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/test-easy-install-31449.write-test'
```

Full output: https://gist.github.com/cdeil/7bc2fc9301bb8b0e2bf335ff31e6981d

From another folder

```
$ sherpa_test
```

gives a ton of test errors from plotting tests because I don't have the test data:
https://gist.github.com/cdeil/5cf3919209ea1c7913108461170a406c

Here's the first:

```
_______________________________________________________ ERROR at setup of test_plot[model-args0] _______________________________________________________

ui = <module 'sherpa.astro.ui' from '/Users/deil/Library/Python/3.5/lib/python/site-packages/sherpa/astro/ui/__init__.py'>
make_data_path = <function make_data_path.<locals>.wrapped at 0x10b8a3378>

    @requires_fits
    @pytest.fixture(autouse=True)
    def load_data(ui, make_data_path):
        """
        Load dataset before every test.
        """
>       ui.load_data(make_data_path("3c273.pi"))

/Users/deil/Library/Python/3.5/lib/python/site-packages/sherpa/astro/ui/tests/test_chips_unit.py:66: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/deil/Library/Python/3.5/lib/python/site-packages/sherpa/conftest.py:140: in wrapped
    return os.path.join(path, arg)
/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/posixpath.py:89: in join
    genericpath._check_arg_types('join', a, *p)
[...]

Issue #235, opened 2016-07-16
pip / PyPI for Sherpa?


cdeil:
I was asked if `pip install sherpa` is supposed to work. Is it?

I see that the version on PyPI is very outdated (and not a stable tag?):
https://pypi.python.org/pypi/sherpa/5d12147

Maybe you could clarify this question here?
https://github.com/sherpa/sherpa#source-build
(And maybe upload the latest stable source dist to PyPI, or if you don't want to use PyPI, remove the very old one to avoid confusion?)

olaurino:
It was decided in the early stages of development that we would not have the package in PyPI, although I had included it in my prototyping. I think the rationale for that decision does not stand anymore and we could upload the package to PyPI.

As you know, building Sherpa requires several dependencies, so to me the advantage of `pip install sherpa` over `git clone; python setup.py install` is minimal: you need to read some docs regarding the installation process anyway.

I'll confirm that I can upload the package to PyPI. Thanks for the report.

DougBurke:
Do build issues like #260 and #271 mean we should hold off on this? In other words, we don't want to suddenly have a "flood" of bug reports for things we already know about. Or are those issues not really relevant to the 'pip install' use case?

DougBurke:
I also guess that a 'pip install' approach means that the user doesn't get to change the build defaults (okay for most users, but a caveat if they want to take advantage of XSPEC)?

DougBurke:
I'm going to close this issue as we now have https://pypi.python.org/pypi/sherpa/4.9.0 and (after ensuring that numpy is installed)

```
% pip install sherpa
Collecting sherpa
  Using cached sherpa-4.9.0-1.tar.gz
Requirement already satisfied: numpy in ./miniconda2/envs/foo35/lib/python3.5/site-packages (from sherpa)
Collecting six (from sherpa)
  Using cached six-1.10.0-py2.py3-none-any.whl
Building wheels for collected packages: sherpa
  Running setup.py bdist_wheel for sherpa ... done
  Stored in directory: /home/djburke/.cache/pip/wheels/aa/2f/91/9fd77ab80480e5f148e160dc49cd3567dff634095fc1f3ec86
Successfully built sherpa
Installing collected packages: six, sherpa
Successfully installed sherpa-4.9.0 six-1.10.0
% python -c 'import sherpa; print(sherpa._version.get_versions())'
{'full': '187333e830c1cbe9413c90fc77ba52784fd6bd85', 'version': '4.9.0'}
[...]

PR #271, opened 2016-09-10
Travis changes for Python 3.5

DougBurke:
# Summary

Python 3.5 builds must now pass for the overall tests to pass. The versions have been swapped, so that there's now only one Python 2.7 run (testing everything) and all the other runs use Python 3.5.
# Discussion

The first commit should be done, now we have merged the Python 3 support to master (this removes the "AllowFailures" rule for Python 3.5). The second commit swaps the 2.7 and 3.5 builds (_except_ for the pyfits test) since the idea is that for the standalone build we are better off testing the various options with 3.5 rather than 2.7. A 2.7 run is needed, which tests as much as possible, to make sure changes are suitable for CIAO.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+# leaving this as 2.7 for the pyfits test, as I am not sure if that works
+# with Python 3.5
 python:
   - "2.7"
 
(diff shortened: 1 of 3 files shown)


olaurino:
I am not sure we should be doing this at this point, but it's useful to test the change nevertheless. At this point, Python 2 is still what we support better, while the Python 3 support is still young. This change might make the Python 2 builds unstable as we might miss regressions on that platform. In any case this is a good discussion to have.

DougBurke:
The remaining failures appear to be due to warnings - e.g. see https://travis-ci.org/sherpa/sherpa/jobs/158987134

Some of them come from me removing the "inspect.getargspec is deprecated" warning from our list of "allowed" warnings. I had audited the code to remove our use of them, but it appears that the version of astropy.io we're using still uses this routine (which makes sense as I think it is no-linger deprecated as of Python 3.6), so we need to ignore it.

The other failures are more complex.

My view on the python 2.7/3.5 support is slightly different (in that I think the main "customer" for 2.7 is going to be CIAO, and for that we don't actually care about the various different options of develop vs install, smoke vs xxx, as that part is tested by us in our CIAO builds; all we really care about is that all the tests pass when using 2.7; the extra builds that travis tests are - in my view - for the community, which is more likely to use 3.5).

I do think it would make sense to take some of these changes sooner rather than later, in particular d9df7b3 and 3846d82 as they are very simple and "safe" changes, but it's up for debate. 

I think it's useful to have the knowledge of what passes and fails; given the current behaviour of master I think we may want to hold off the PyPI work from #235 until we have a better understanding of what failure messages we may get from the community (some of the other build issues, such as #260 also make me worry about putting Sherpa on PyPI too early). I think this is echoing some of your reservations you had in our recent telecon about 235.

olaurino:
The CIAO vs community point is fair, and as I said it is worth discussing these changes, and they have value anyway. I am not sure we know what is "most likely" to happen. This resource seems to have a different opinion:
https://hynek.me/articles/python3-2016/

However, it is based on PyPI download statistics, which are totally unreliable.

This resource shows a different, possibly more relevant and reliable dataset, but still...:
http://ianozsvald.com/2016/06/20/results-for-which-version-of-python-2vs3-do-london-data-scientists-use/

DougBurke:
I do not understand the travis failures here. The failing test is https://travis-ci.org/sherpa/sherpa/jobs/159029801 which is the astropy, install, package, matplotlib=1.4, numpy=1.9 run. 

The same run but with matplotlib=1.5, numpy=1.10 - i.e. https://travis-ci.org/sherpa/sherpa/jobs/159029802 - passes.

I had taken this to mean there was an issue with the older dependencies (e.g. astropy = 1.0.4), but I get a failure on my local machine with `sherpa_test` with matplotlib=1.5.1 numpy=1.11.1 astropy=1.2.1

Most of the issues seem to be about warning messages, which doesn't strike me as too important, but it looks like there are some failures in test_stat/wstat (at least when I run it); for instance

```
_______________ test_wstat_single_array.test_wstat_ungrouped_all _______________

self = <test_wstat.test_wstat_single_array testMethod=test_wstat_ungrouped_all>

    def test_wstat_ungrouped_all(self):
        ui.ungroup()
    
        # Used git commit 770359b5004374b969ebb63c173f293419397b4c
        # to create the oracle value, on a linux 64-bit machine.
>       self._check_stat(1024, 663.0160968458746)

../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/stats/tests/test_wstat.py:505: 
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/stats/tests/test_wstat.py:482: in _check_stat
    stat = ui.calc_stat()
<string>:1: in calc_stat
    ???
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/ui/utils.py:7869: in calc_stat
    ids, f = self._get_fit(id, otherids)
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/ui/utils.py:7478: in _get_fit
    fit_to_ids, datasets, models = self._prepare_fit(id, otherids)
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/ui/utils.py:7453: in _prepare_fit
    mod = self._get_model(i)
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/ui/utils.py:5521: in _get_model
    return self._add_convolution_models(id, data, model, is_source)
../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/astro/ui/utils.py:8262: in _add_convolution_models
    (data, bkg_srcs))

self = <[AttributeError("'BackgroundSumModel' object has no attribute 'name'") raised in repr()] BackgroundSumModel object at 0x7f46e2741cc0>
srcdata = <DataPHA data set instance '/home/djburke/miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpatest/3c273.pi'>
bkgmodels = {1: <Const1D model instance 'const1d.c'>}

    def __init__(self, srcdata, bkgmodels):
        self.srcdata = srcdata
        self.bkgmodels = bkgmodels
        scale_factor = self.srcdata.sum_background_data(lambda key, bkg:1)
        bkgnames = [model.name for model in bkgmodels.values()]
>       name = '%g * (' % scale_factor + ' + '.join(bkgnames) + ')'
E       TypeError: only length-1 arrays can be converted to Python scalars

../miniconda2/envs/sherpa-268-newer/lib/python3.5/site-packages/sherpa/astro/background.py:33: TypeError
```

My environment was

```
% conda create -n=sherpa-268-newer python=3.5 astropy matplotlib
% source activate sherpa-268-newer
% python setup.py install
% git submodule init
% git submodule updat
% pip install ./sherpa-test-data
% cd ..
% sherpa_test >& log.since.the.output.can.be.long
```

wmclaugh:
At present, the 3.5 base is still a beta release while  2.7 is an officially supported release. I believe the prudent course of direction would be to keep the 2.7 testing as the core for the time being. Once the 3.5 support is no longer considered beta (and we have a more concrete understanding of where the core user base is), I would advocate revisiting changing the baselines.

DougBurke:
@olaurino should we be concerned about the failing test? It is "only" about warnings, so it does not seem really serious,  but I am not confident enough in my understanding of the testing setup to really be sure.

olaurino:
I would be happier if I knew what the warnings are about. The differences between build 9 and 10 are the versions of numpy and matplotlib, so I guess it has to do with those versions on python 3.5, but there is no way to tell from the logs, so this should be tested out locally with that combination of dependencies.

It doesn't look like a serious issue at first glance.

DougBurke:
I'm closing this as #336 has been merged, which has cleaned up the Travis configuration and includes more python 3 tests.

PR #278, opened 2016-09-20
fix #277 - parallel_map not working on py3 with numcores=1

### LINKED ISSUE #277: `parallel_map` not working on python3 when `numcores`=1
function `sherpa.utils.parallel_map` does not work when called with `numcores=1` on Python 3.

olaurino:
Fix #277.

# Release Note
The `utils` function `parallel_map` failed on Python 3 when called with `numcores=1`, i.e. on systems with only one processor/core. This has been fixed.

# Description
Note that the tests have been changed to make them fail (with the fix removed) on all platforms, whatever the number of cores. So, when testing I made sure I would see the 1-core test fail on a system where the old test would pass and then, with the fix back in place, I made sure the test passes again.

### DIFF
--- a/sherpa/utils/tests/test_utils.py
+++ b/sherpa/utils/tests/test_utils.py
@@ -17,6 +17,7 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+import pytest
 import numpy
 import multiprocessing
 from numpy.testing import assert_allclose, assert_equal
@@ -263,9 +264,9 @@ class test_utils(SherpaTestCase):
                 val = utils.neville2d(xx, yy, x, y, fval)
                 self.assertTrue(utils.Knuth_close(answer, val, tol))
 
-    def test_parallel_map(self):
-        ncpus = multiprocessing.cpu_count()
 
+@pytest.mark.parametrize("num_tasks", [1, 3, 100])
+def test_parallel_map(num_tasks):
         numtasks = 8
         f = numpy.sum
         iterable = [numpy.arange(1, 2+2*i) for i in range(numtasks)]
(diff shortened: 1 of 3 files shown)


olaurino:
@DougBurke could you quickly review this, please?

I understand @wmclaugh is leaning towards not getting this into 4.8.2. Whether or not we do, the fix is here.

Note that this needs to be merged before I can package the new binaries, because of tagging issues.

olaurino:
Travis is complaining because the log files (of the full builds including xspec) are getting too long. I am trying to compensate and testing out solutions from my fork.

olaurino:
Travis should be happy now (it was from my fork).

sherpa/utils/tests/test_utils.py
```diff
@@ -263,16 +264,16 @@ def test_neville2d(self):
                 val = utils.neville2d(xx, yy, x, y, fval)
                 self.assertTrue(utils.Knuth_close(answer, val, tol))
 
-    def test_parallel_map(self):
-        ncpus = multiprocessing.cpu_count()
 
+@pytest.mark.parametrize("num_tasks", [1, 3, 100])
+def test_parallel_map(num_tasks):
```
DougBurke: I'm not happy with the function argument being called `num_tasks` as it is too similar to `numtasks` and makes it harder to see what is going on with the tests.

sherpa/utils/tests/test_utils.py
```diff
@@ -263,16 +264,16 @@ def test_neville2d(self):
                 val = utils.neville2d(xx, yy, x, y, fval)
                 self.assertTrue(utils.Knuth_close(answer, val, tol))
 
-    def test_parallel_map(self):
-        ncpus = multiprocessing.cpu_count()
 
+@pytest.mark.parametrize("num_tasks", [1, 3, 100])
```
DougBurke: I htink it would be good to iterate over both `num_tasks` (or whatever it gets called, see next comment) and `numtasks`, so that we have covered at least:

- (1, 1)
- (n > 1, 1)
- (1, n > 1)
- (n, m) where n > m
- (n, m) where n < m
- (n, n) for n > 1

just to make sure we've covered the corner cases of more tasks than processes, more processes than tasks, the same number of tasks and processes, ...

olaurino:
@DougBurke I agree with your comments. I'll close this PR and open a new one against master once the 4.8.2 release branch will be merged.

PR #280, opened 2016-09-30
WIP: Add a python=3.6 build option on Travis

DougBurke:
Python 3.6 is now close to release, so test out the nightly branch which is
hopefully close to this release: see
https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against
for more details.

I picked the Python 3.5 test as the one to copy, and marked it as
allowed-to-fail, as this is intended to be informational since we do
not claim Python 3.6 support yet.

PR #281, opened 2016-10-03
make the travis logs less verbose

olaurino:
This PR should make the travis log less verbose and avoid hitting the log size limits of Travis.

There are far better solutions (including not relying on Travis any more ;-) ), so that the logs can be recovered if the builds fail, but I can't explore them right now and we need the builds to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ before_install:
     fi
 
 install:
-  - python setup.py $INSTALL_TYPE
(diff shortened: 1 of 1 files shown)


olaurino:
@DougBurke could you test if by cherry-picking this commit in a side branch makes PR #280 pass?

DougBurke:
I didn't actually see your comment, but I've already done so ;-)

olaurino:
Great minds think alike ;-)

PR #282, opened 2016-10-03
parallel_map not working on py3 with numcores=1

### LINKED ISSUE #277: `parallel_map` not working on python3 when `numcores`=1

olaurino:
Fix #277.

# Release Note
The `utils` function `parallel_map` failed on Python 3 when called with `numcores=1`, i.e. on systems with only one processor/core. This has been fixed.

# Description
Note that the tests have been changed to make them fail (with the fix removed) on all platforms, whatever the number of cores. So, when testing I made sure I would see the 1-core test fail on a system where the old test would pass and then, with the fix back in place, I made sure the test passes again.

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -46,7 +46,7 @@ from sherpa.utils._psf import extract_kernel, normalize, set_origin, \
 
 from sherpa import get_config
 from six.moves.configparser import ConfigParser, NoSectionError
-from six.moves import xrange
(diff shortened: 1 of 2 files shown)


olaurino:
Rebasing, was 87a5d4a

DougBurke:
I'm happy for this to be merged.

PR #283, opened 2016-10-03
fix 273, sample flux and numpy deprecations

### LINKED ISSUE #273: Numpy VisibleDeprecationWarning
sherpa/astro/flux.py uses a float as an index array, resulting in a deprecation warning from Numpy.

To reproduce:

``` python
from sherpa.astro import ui
ui.load_arrays(1, [1, 2], [1, 2])
ui.set_source(ui.polynom1d.p)
ui.sample_flux()
```

You would get the following warning:

>sherpa/astro/flux.py:163: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  upconfidence = sf[(1.0 - myconfidence) * size - 1]
sherpa/astro/flux.py:164: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future

### LINKED ISSUE #276: issue with sample_flux on Python 3
The function `sample_flux` throws an exception when run under Python 3, in the following case:

``` python
from sherpa.astro import ui

load_pha('sherpa-test-data/sherpatest/3c273.pi')
set_model('polynom1d.p1')
fit()
covar()
scal = get_covar_results().parmaxes
sample_flux(p1, 0.5, 1, num=5, correlated=False, scales=scal)
```

```
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/pdb.py", line 1661, in main
    pdb._runscript(mainpyfile)
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/pdb.py", line 1542, in _runscript
    self.run(statement)
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/bdb.py", line 431, in run
    exec(cmd, globals, locals)
  File "/home/vagrant/test_kenny.py", line 1, in <module>
    from sherpa.astro.ui import *
  File "<string>", line 1, in sample_flux
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/astro/ui/utils.py", line 11521, in sample_flux
    bkg_id=bkg_id)
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/astro/ui/utils.py", line 11383, in sample_energy_flux
    scales)
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/astro/flux.py", line 91, in sample_flux
    samples = sampler.get_sample(fit, samples, num=num)
  File "/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/sim/sample.py", line 260, in get_sample
    val, scale, int(num)) for val, scale in izip(vals, scales)]
TypeError: zip argument #2 must support iteration
Uncaught exception. Entering post mortem debugging
Running 'cont' or 'step' will restart the program
> /home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/sim/sample.py(260)get_sample()
-> val, scale, int(num)) for val, scale in izip(vals, scales)]
```

olaurino:
Fix #273.

# Release Note
The `sample_flux` function was not working under Python 3 if the `scales` argument was provided. This has been fixed. Also, a `DeprecationWarning` was issued by `numpy` because during the `sample_flux` execution values were extracted from an array with non-integer indices. This has also been fixed.


olaurino:
There are instances of the deprecation warning that are still there and totally unrelated to the fix. I am not sure whether I should fix them in this PR or revert the changes to the deprecation warning in this PR and fix all of them in one PR.

olaurino:
@DougBurke I'll let you decide if you have a minute. I am fine either way but leaning towards making two PRs.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -551,3 +551,12 @@ def test_load_table_fits(self):
         data = ui.get_data(1)
         self.assertEqualWithinTol(data.x, [1, 2, 3])
         self.assertEqualWithinTol(data.y, [4, 5, 6])
+
+
+def test_bug_276(make_data_path):
```
olaurino: Need the `@requires_data` annotation.

olaurino:
Actually, I think I am going to revert the known warnings and keep the code changes related to the deprecation warnings in this PR. I'll just have to fix some more warnings when the time comes.

olaurino:
Finally, in PR #284 I added the `requires_data` decorator to the `make_data_path` fixture itself, so any test using that fixture is also marked to be skipped if the data files are not available.

olaurino:
Rebasing, tip was at c1a4357.

olaurino:
Rebasing, was 4ee092d

olaurino:
The failure in travis is weird. GitHub creates a hidden merge commit to test that the PR can be merged, and travis uses this commit to run the tests in a PR build. In this case, it looks like the PR build is based on the old merge commit, so the travis log changes are not there. I am not sure if it's an issue in GitHub or Travis, but I'll try resetting the build by amending the last commit.

sherpa/astro/flux.py
```diff
@@ -160,8 +160,10 @@ def print_sample_result(title, arg):
         for x in oflxiflx:
             sf = numpy.sort(x)
             median = numpy.median(sf)
-            upconfidence = sf[(1.0 - myconfidence) * size - 1]
-            loconfidence = sf[myconfidence * size - 1]
+            upconfidence_index = int((1.0 - myconfidence) * size - 1)
```
wmclaugh: It would be cleaner from a tracking perspective to keep the change as a straight forward fix and then add your suggestion of using numpy.percentile as a separate request. Doing so would allow more concise testing and reduce the risk of inadvertent side effects (subtleties) being introduced.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -551,3 +551,13 @@ def test_load_table_fits(self):
...
+
+
+@requires_fits
+def test_bug_276(make_data_path):
+    ui.load_pha(make_data_path('3c273.pi'))
+    ui.set_model('polynom1d.p1')
+    ui.fit()
+    ui.covar()
+    scal = ui.get_covar_results().parmaxes
+    ui.sample_flux(ui.get_model_component('p1'), 0.5, 1, num=5, correlated=False, scales=scal)
```
DougBurke: A newline is "needed" here (according to GitHub's diff).

sherpa/astro/flux.py
DougBurke: As noted in the code review, I've added an issue to track this change (#286), so it isn't forgotten.

sherpa/astro/ui/tests/test_astro_ui.py
olaurino: Since we are talking about review practices, one of the main advantages of working on the same repo rather than on individual forks is that people can push commit to a PR, especially in simple cases like this.

I would encourage us all to do it. It's good to coordinate this kind of changes, i.e. leaving a review comment saying: "there should be a newline at the end of file, I am going to push a commit to fix that".

Of course, it's just a suggestion, and I don't think it makes any sense to turn this into a policy. Just a reminder we can take advantage of the decision to work on the same repo.

DougBurke: I've added the newline.

PR #284, opened 2016-10-03
fix 275, string representation of data classes under py3

### LINKED ISSUE #275: issue with data representation on python 3
Printing a dataset results in an exception on Python 3.
``` python
from sherpa.astro.ui import *
load_arrays(1, [1, 2, 3], [1, 2, 3], DataPHA)
print(get_data())
```
Since the exception is raised in the `BaseData` class this is likely to be applicable to other data types, not only `DataPHA`.

```
<ipython-input-4-f8304f59701f> in <module>()
----> 1 print(get_data())

/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/astro/data.py in __str__(self)
    403         try:
    404             self._fields = filter((lambda x: x != 'header'), self._fields)
--> 405             ss = BaseData.__str__(self)
    406         finally:
    407             self._fields = old

/home/vagrant/miniconda2/envs/sherpa3/lib/python3.5/site-packages/sherpa/data.py in __str__(self)
    109         """
    110 
--> 111         fields = self._fields + getattr(self, '_extra_fields', ())
    112         fdict = dict(izip(fields, [getattr(self, f) for f in fields]))
    113         return print_fields(fields, fdict)

TypeError: unsupported operand type(s) for +: 'filter' and 'tuple'
```

olaurino:
Fix #275.

# Release Note
Data classes `DataPHA`, `DataARF`, `DataRMF`, `DataIMG`, and `DataIMGInt` in `sherpa.astro.data` would throw an exception if users tried to print them as strings, under Python 3. This has been fixed.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -107,7 +107,7 @@ class DataARF(Data1DInt):
         old = self._fields
         ss = old
         try:
-            self._fields = filter((lambda x: x != 'header'), self._fields)
+            self._fields = tuple(filter((lambda x: x != 'header'), self._fields))
             ss = BaseData.__str__(self)
         finally:
             self._fields = old
@@ -181,7 +181,7 @@ class DataRMF(Data1DInt):
         old = self._fields
         ss = old
         try:
-            self._fields = filter((lambda x: x != 'header'), self._fields)
+            self._fields = tuple(filter((lambda x: x != 'header'), self._fields))
             ss = BaseData.__str__(self)
         finally:
             self._fields = old
@@ -401,7 +401,7 @@ class DataPHA(Data1DInt):
         old = self._fields
         ss = old
         try:
-            self._fields = filter((lambda x: x != 'header'), self._fields)
+            self._fields = tuple(filter((lambda x: x != 'header'), self._fields))
             ss = BaseData.__str__(self)
         finally:
             self._fields = old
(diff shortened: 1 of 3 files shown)


olaurino:
Rebasing, was 12b2e24

sherpa/astro/data.py
```diff
@@ -107,7 +107,7 @@ def __str__(self):
         old = self._fields
```
DougBurke: Given that this is repeated for all the three classes (or at least it seems to be from a quick look), can we not move this it into a separate routine? It's probably not worth the hassles of doing this in a nice OO way, but I thought I'd ask.

olaurino: You just reminded me the comment I wanted to add to the description of the fix. I had your same impression and urge. The code does look the same, and even without adding an intermediate class the __str__ method might be pushed up. I'll have to check that the code is exactly the same, though.

olaurino: By the way, something else I found out while reviewing this is that the base class's `__init__` is called without passing it any arguments apart from `self`. Then it looks like the base class inspects the frame to get the arguments passed to the calling class, which is weird. But again, I didn't look any closer.

DougBurke:
Weird is one word that springs to mind when reviewing this code!

PR #287, opened 2016-10-10
rewrite sherpa.stats.Stat.calc_stat and simplify sherpa.fit.Fit (fix #227 #248 #289 #292).

### LINKED ISSUE #227: Issues using wstat when grouping/filtering data

DougBurke:
# Status

Finished (there is a question about the support for weighted columns).

# Summary

The aim is to move logic from the `sherpa.fit.Fit` class into the `sherpa.stats.Stat` class, namely the choice of what data is actually needed to calculate the given statistic. It is the approach discussed in https://github.com/sherpa/sherpa/issues/248#issuecomment-251795903

As well as (hopefully) cleaning up the API, it addresses the following issues: #227, #248, #289, #292.

Although it does not directly address #238 - in that it does not substantially change `sherpa/stats/tests/test_stat.py` (the changes made to this file are related to the code changes, and do not address the fact that many tests should ideally be run without the need for I/O), there are a number of tests added to `sherpa/stats/tests/test_stat_unit.py` and `sherpa/tests/test_fit_unit.py` that test the low-level statistic code without the need for an I/O library.

It does not address #293.

# To do

- [X] change `calc_stat_from_data` so that it accepts a single dataset/model (at present the inputs must be `*SimulFit*` objects)
- [X] review remaining `sherpa.fit.Fit` methods to see which need to use `calc_stat_from_data` rather than `calc_stat`: have used this approach to convert `calc_chisqr`
- [ ] review the remaining 'to do' notes and questions, including:
  - [X] tests of background-subtracted data
  - [ ] how is the weight column passed to the statistics object
  - [X] is the current wstat implementation really restricted to the first background component (e.g. if there are multiple ones in grating data). I believe it is (see #293)
- [X] delete the old `calc_stat` routine and rename `calc_stat_from_data`
- [X] the calculation of qval and reduced statistic should also be moved from the `sherpa.fit.Fit` class to the `sherpa.stats.Stat` class (this could be done in a separate PR). Note that the conversion of `calc_chisqr` has already moved extra logic into the `Stat` class, but that qval/rstat is also used by the `esterrors` method, which looks complicated. **MOVED** into the "Follow up changes" section below
- [X] should there be improvements to the docstrings to explain what is going on (e.g. stats) **MOVED** into the "Follow up changes" section below.
- [X] should there be any discussion about the data API - e.g. there doesn't seem to be a nice way to extract the data for a background component to match the source dataset, or for getting at some of the `DataPHA` extras, such as the BACKSCAL values grouped to match the data (it may be that the current API is okay and that the extra complexity needed here is okay). **MOVED** into the "Follow up changes" section below.

# Design notes

The `sherpa.fit.Fit` object needs to calculate the statistic value for a given set of data (including any filtering, grouping) and models. It does this by deconstructing the data objects (i.e. the subclasses of `sherpa.data.Data`) to get the "raw" data - e.g. count per bin, model value per bin - and then sending this to the `sherpa.stats.Stat.calc_stat` method to calculate the statistic.

There are several issues with this approach: 
- The code does not use the `sherpa.data.Data` API to get at the data, instead doing things manually - e.g. the backscal values. This is the cause of the original wstat error #248 (and problems with the original fix in #249) , because the grouping/filtering isn't being applied correctly as there really isn't a simple API to get at these values. The PR #269 addresses this issue (i.e. fixes up the code), but that doesn't address the following points
- The decision about what items a given stats class needs is being made by the fit object, not the stats object itself. This leads to an API that is brittle and not easily extensible. For example, to calcuate the wstat statistic we need to send in items such as the backscal ratio and exposure times, which meant that the fit object had to be changed. Other statistics may need different elements, but there's no way of supporting this without also changing the fit object.
- There are other places in the fit object where knowledge about how the stats object behaves is required. For example, the decision of whether this is a "chi-square" or "chi-square-like" statistic in `calc_chisqr` and `calc_stat_info`.

The proposed solution is to change the interface so that the fit object sends the stat object the data _object(s)_, rather than raw arrays, and then the stat object can pull out the data it needs. This should be more extensible (e.g. easier to add in new statistics); in particular I plan to improve on the "multi statistic example" using this new interface. 

I do not claim that this is the final solution, since I think that there are design issues for the interplay between the statistic and data objects: as mentioned above, there is no nice API to get at all the information needed by wstat, which means that the sherpa.stats.WStat class probably has too much knowledge of the `sherpa.data.Data` class in it, but
1. this is actually true of the existing code, where the logic is embedded within the fit class
2. addressing this particular issue is a much larger change than this work

# Outstanding issues

The only regression that I am aware of in this PR is that the old code path supported a weights column for chi-square based statistics. This functionality has currently been removed. Note that no tests fail because of it, so it appears to have been untested. I'm also not sure how users of the `sherpa.astro.ui` API would have access to this (i.e. if I have called `load_data("somefilename")` then how do I set/change the weights? I do not think it would be hard to add back in support for the weights keyword (I think it's mainly a case of supporting `**kwargs` in the `sherpa.stats.Stat.calc_stat[chisqr]` routine), but have not investigated this.

In both the `master` branch and this PR, the WSTAT calculation only uses the first background component if multiple components (e.g. Chandra HETG datasets with `BACKGROUND_UP` and `BACKGROUND_DOWN` columns) are present. The PR adds a warning message to tell the user, but this code path has not been tested. So this PR does not address #293.

I added explicit tests of the sigma-rejection iterated-fit routine in this PR (to check that the changes did not change its behavior) but there are no tests of the primini iterated-fit routine.
# Follow up changes

These are changes that could be made in new PRs if this one is accepted.

As mentioned above, there is some code in the `Fit` class - e.g. calculation of the "q" value - that should probably be in the relevant `Stat` class.

There are a number of cases where explicit checks for a `Chi2` instance is made (excluding `LeastSq`) which really should just be a check that they have the concept of `staterror` (i.e. that `calc_staterror` does not return `None`). An example of this is the "sigmarej" iterated-fit method. This is potentially an invasive change since there are some cases where `calc_staterror` returns something (e.g. `LeastSq`) and it probably should be changed to return `None`. An associated change would be to to remove the statistical and systematic error values returned by the `sherpa.data.Data.to_fit` method, requiring code that needs them to ask for them explicitly (in other words, remove the assumption that the data has these error values which means you need to special case likelihood statistics where this is not relevant).

This PR is not the place to make the data docstring changes discussed above - e.g. how to access various components, such as raw data, data to be fit, models evaluated on the data grid. Also, how t special case things like background data sets or the BACKSCAL scalar/array which are only supported by one or more sub classes (in this case `sherpa.astro.data.DataPHA`)? I have added method-level docstrings to at least explain what the statistic routines are doing, but we don't have any overview-level documentation where the overall information flow and procedures should be documented.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -41,13 +41,6 @@ info = logging.getLogger(__name__).info
 __all__ = ('FitResults', 'ErrorEstResults', 'Fit')
 
 
-# def get_valid_args(func):
-#     valid_args = func.func_code.co_varnames[:func.func_code.co_argcount]
-#     kwargs_length = len(func.func_defaults)  # number of keyword arguments
-#     valid_kwargs = valid_args[-kwargs_length:]  # because kwargs are last
-#     return valid_kwargs
-
-
 def evaluates_model(func):
     """
     Fit object decorator that runs model startup() and teardown()
@@ -352,10 +345,11 @@ class ErrorEstResults(NoNewAttributesAfterInit):
         self.parmins = ()
         self.parmaxes = ()
         self.nfits = 0
-        success = True
+        # The success flag is currently unused
+        # success = True
         for i in range(len(parlist)):
-            if results[2][i] != est_success:
-                success = False
+            # if results[2][i] != est_success:
+            #     success = False
             if (results[2][i] == est_hardmin or
                     results[2][i] == est_hardminmax):
                 self.parmins = self.parmins + (None,)
@@ -733,11 +727,6 @@ class IterFit(NoNewAttributesAfterInit):
         if grow < 0:
             raise SherpaErr("'grow' factor must be zero or greater")
 
-        # Keep record of current and previous statistics
-        #
-        # In this algorithm however, they do not seem to be used
-        # previous_stat = float32(finfo(float32).max)
-        # current_stat = statfunc(pars)[0]
         nfev = 0
         iters = 0
 
@@ -753,7 +742,13 @@ class IterFit(NoNewAttributesAfterInit):
             else:
                 mask_original.append(array(d.mask))
 
+        # QUS: why is teardown being called now when the model can be
+        #      evaluated multiple times in the following loop?
+        #      Note that after the loop  self.model.startup
+        #      is called, so either I [DJB] or the code has them the
+        #      wrong way around
         self.model.teardown()
+
         final_fit_results = None
         rejected = True
         try:
@@ -841,6 +836,8 @@ class IterFit(NoNewAttributesAfterInit):
 
         self._dep, self._staterror, self._syserror = self.data.to_fit(
             self.stat.calc_staterror)
+
+        # QUS: shouldn't this be teardown, not startup?
         self.model.startup()
 
         # N.B. -- If sigma-rejection in Sherpa 3.4 succeeded,
@@ -1020,9 +1017,7 @@ class Fit(NoNewAttributesAfterInit):
 
         """
 
-        # TODO: does this need to be updated due to changes in _calc_stat?
-        #       also, this logic would be better in the stat class
-        #       than here
+        # TODO: This logic would be better in the stat class than here
         #
         stat, fvec = self._calc_stat()
         model = self.data.eval_model_to_fit(self.model)
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -977,20 +977,43 @@ class Fit(NoNewAttributesAfterInit):
         """
         self.model.guess(*self.data.to_guess(), **kwargs)
 
+    # QUS: should this have an @evaluates_model decorator?
+    #
     def _calc_stat(self):
-        dep, staterror, syserror = self.data.to_fit(self.stat.calc_staterror)
-        model = self.data.eval_model_to_fit(self.model)
-        extra_args = self._iterfit.get_extra_args(dep)
-        return self.stat.calc_stat(dep, model, staterror, syserror=syserror,
-                                   extra_args=extra_args)
+        """Calculate the current statistic value.
+
+        Returns
+        -------
+        statval, fvec : number, array of numbers
+            The overall statistic value and the "per-bin" value.
+        """
+
+        # Casting into simultaneous objects is not great (and should
+        # be done in __init__ if we are going to do this).
+        #
+        # Note that, because of Fit.simulfit, we can not assume that
+        # self.data and self.model are single (i.e. not simultaneous)
+        # objects.
+        #
+        if isinstance(self.data, DataSimulFit):
+            data = self.data
+        else:
+            data = DataSimulFit('simulfit data', (self.data,))
+
+        if isinstance(self.model, SimulFitModel):
+            model = self.model
+        else:
+            model = SimulFitModel('simulfit model', (self.model,))
+
+        # TODO: is there anything missing here that
+        #       self._iterfit.get_extra_args calculates?
+        return self.stat.calc_stat_from_data(data, model)
 
     def calc_stat(self):
-        """Calculate the fit statistic for a data set.
+        """Calculate the statistic value.
 
-        Evaluate the model for one or more data sets, compare it to
-        the data using the current statistic, and return the value.
-        No fitting is done, as the current model parameter, and any
-        filters, are used.
+        Evaluate the statistic for the current model and data
+        settings (e.g. parameter values and data filters).
 
         Returns
         -------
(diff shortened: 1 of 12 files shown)


DougBurke:
Note that the current test failures don't appear to be because of the changes in this code, but something in Travis/conda/... It's one particular build (run 4) that fails, and I restarted that particular tun from a previously-passing build in this PR (0a2fe90) and it now fails. Unfortunately, due to the log now being hidden (PR#281) it's hard to track down the problem.

olaurino:
I think I understand the issue:

https://travis-ci.org/sherpa/sherpa/jobs/167694569

I'll suggest a change that would make it pass as soon as I've tested it.

DougBurke:
@olaurino and suddenly the tests are passing again!

DougBurke:
@joleroi - if you have time could you check if this PR works as well as #269 ?

DougBurke:
I've added a comparison of the coverage from master (at the time the PR branch was created) to 1121527 which you can find at https://gist.github.com/DougBurke/a935ed97caef8bd3c302784ea719ff7a (it includes tar files of the HTML output if you want to compare the results).

The overall coverage increases fro m61% to 63% with this PR.

olaurino:
I am investigating an issue with the `srcextent` script in CIAO (it's hard to reproduce in standalone mode, so I haven't opened a GitHub issue yet, the issue was reported to me by @wmclaugh). This PR removes the line that triggers that issue, so I tried to test this PR against today's CIAOD (it took forever to set things up, argh) but there is still a failure, although a different one. I'll send an email to our development list with the details, but before this PR can go in, we need to make sure we harmonize this PR with the CIAO scripts, one way or the other.

DougBurke:
It's a bit hard for me to help work on a bug which I know nothing about!

sherpa/fit.py
```diff
@@ -977,20 +961,26 @@ def guess(self, **kwargs):
...
+
+        Returns
+        -------
+        statval, fvec : number, array of numbers
+            The overall statistic value and the "per-bin" value.
+        """
+
+        # TODO: is there anything missing here that
+        #       self._iterfit.get_extra_args calculates?
+        return self.stat.calc_stat(self.data, self.model)
```
olaurino: This is not necessarily a show stopper, but wouldn't it be more generic to pass the entire object, rather than just `self.data` and `self.model`? This way more specific statistics' `calc_stat` methods could subset more specific fit objects and use more information that would not be available otherwise, providing more flexibility.

sherpa/stats/__init__.py
```diff
@@ -64,10 +68,173 @@ def __repr__(self):
...
+            data parameter.
+
+        Returns
+        -------
+        statval, fvec : number, array of numbers
+            The statistic value and the per-bin "statistic" value.
+
+        """
+
         raise NotImplementedError
```
olaurino: Should we consider using proper abstract base classes? I guess this is a question for another day.

sherpa/fit.py
```diff
@@ -449,17 +491,17 @@ def __init__(self, data, model, stat, method, itermethod_opts=None):
         # models into the objects needed for simultaneous fitting,
         # if they are not already in such objects.
         self.data = data
-        if (type(data) is not DataSimulFit):
+        if type(data) is not DataSimulFit:
```
DougBurke: I decided that I didn't want to address these issues in this PR. I felt it was big enough as is!

There's a bunch of areas like this - as you note below - that should be cleaned up, but I am not convinced that we yet have a reliable enough set of tests within sherpa to validate any such changes.

sherpa/stats/__init__.py
DougBurke: I leave that decision for other people to make!

PR #298, opened 2016-11-03
Move XSPEC table support to load_xstable_model and deprecate its support in load_table_model (Fix #270)

### LINKED ISSUE #270: Add a "load_xspec_table_model" command and deprecate XSPEC support in load_table_model
This is related to #142 and to discussions with users. I think it is confusing to use load_table_model for both XSPEC and Sherpa style table models. They are (subtly) different concepts, so it makes sense to separate them out. It also leads to better error messages, since at present if a user tries to load an XSPEC table model file with `load_table_model` but there's a problem (e.g. no XSPEC support, or the file does not match the XSPEC table model definition) then it silently falls through to the Sherpa table model support. I am pretty sure that some data can - in some cases - be read in, but it is not meaningful. Note that the error handling (e.g. fall through) cases in load_table_model is quite complex (mainly down to the complex handling of cases in the I/O calls it uses), so anything that can help simplify things is good in my book.

So, I say we add an explicit function for loading XSPEC table models - e.g. `load_xstable_model` or `load_xspec_table_model` - in `sherpa.astro.xspec` which can be re-exported in `sherpa.astro.ui`. This function is then used by `sherpa.astro.ui.load_table_model` (i.e. it is just a move of the code), and a deprecation warning is added to the code path if the call succeeds (e.g. to remove this support in the next major release of Sherpa after 4.9).

I plan to look at this for the 4.9 release, unless someone beats me to it.

DougBurke:
# Status

Finished. [argh - somehow firefox managed to overwrite my original summary, so re-creating it]

# Summary

Add the `load_xstable_model` routine to the `sherpa.astro.ui` module, which supports loading XSPEC additive or multiplicative (atable and mtable) models. The support for these models is still available via `load_table_model` in this release, but it is deprecated. The `read_xstable_model` routine has been added to the `sherpa.astro.xspec` module.

This addresses #270.

# Details

The read logic has been moved to `sherpa.astro.xspec.read_xstable_model`, which returns a model instance. This is then used by the `sherpa.astro.ui.utils.load_xstable_model` routine to create the model and add it to the session. The `load_table_model` routine calls `load_xstable_model` and creates a `DeprecationWarning` warning if the call succeeds.

There is a slight difference in `load_table_model` and `load_xstable_model`, in that the former calls `sherpa.utils.is_binary_file` before calling `read_xstable_model` whereas there is no equivalent check in `load_xstable_model`. This is intentional as it matches the old behavior of `load_table_model`, but I believe that we should move away from using `is_binary_file`, since it is based on a *very* simple heuristic, and instead rely on the IO library to raise a sensible exception.

# Example

The two routines - `load_table_model` and `load_xstable_model` - behave the same:

```
>>> ui.load_table_model('m1', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
>>> ui.load_xstable_model('m2', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
>>> print(m1)
xstablemodel.m1
   m1.tau       thawed            1            1           10           
   m1.beta      thawed          0.1          0.1          0.5           
   m1.t         thawed          0.1          0.1          1.3           
   m1.norm      thawed            1            0        1e+24           
>>> print(m2)
xstablemodel.m2
   m2.tau       thawed            1            1           10           
   m2.beta      thawed          0.1          0.1          0.5           
   m2.t         thawed          0.1          0.1          1.3           
   m2.norm      thawed            1            0        1e+24           
```

If we turn on warnings, we see a warning from `load_table_model` (not shown, but `load_xstable_model` does not raise this warning):

```
>>> import warnings
>>> warnings.filterwarnings('always')
>>> ui.load_table_model('m1', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
/home/djburke/sherpa/sherpa-fix-270/sherpa/astro/ui/utils.py:9018: DeprecationWarning: Use load_xstable_model to load XSPEC table models
  DeprecationWarning)
>>> ui.load_xstable_model('m2', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
>>> ui.load_table_model('m1', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
/home/djburke/sherpa/sherpa-fix-270/sherpa/astro/ui/utils.py:9018: DeprecationWarning: Use load_xstable_model to load XSPEC table models
  DeprecationWarning)
```

If you change the warning to "once" then the warning is only displayed the first time it is used:

```
>>> warnings.filterwarnings('once')
>>> ui.load_table_model('m1', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
/home/djburke/sherpa/sherpa-fix-270/sherpa/astro/ui/utils.py:9018: DeprecationWarning: Use load_xstable_model to load XSPEC table models
  DeprecationWarning)
>>> ui.load_table_model('m1', 'sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod')
>>> 
```

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
@@ -24,7 +24,6 @@
 #
 
 import os
-import tempfile
 
 import pytest
 
@@ -32,8 +31,6 @@ import pytest
 
 from sherpa.astro import ui
 from sherpa.utils import requires_data, requires_fits, requires_xspec
-
-tmpdir = tempfile.gettempdir()
 
 
 # Note that the logic in load_table_model is quite convoluted,
@@ -48,38 +45,32 @@ tmpdir = tempfile.gettempdir()
 #
 
 @requires_fits
-@pytest.mark.skipif(not(os.path.isdir(tmpdir)),
-                    reason='temp directory does not exist')
-def test_load_table_model_fails_with_dir():
+def test_load_table_model_fails_with_dir(tmpdir):
     """Check that the function fails with invalid input: directory
 
-    The temporary directory is used for this (the test is skipped if
-    it does not exist).
+    The temporary directory is used for this.
     """
 
     ui.clean()
     assert ui.list_model_components() == []
     with pytest.raises(IOError):
-        ui.load_table_model('tmpdir', tmpdir)
+        ui.load_table_model('tmpdir', str(tmpdir))
 
     assert ui.list_model_components() == []
 
 
 @requires_fits
 @requires_xspec
-@pytest.mark.skipif(not(os.path.isdir(tmpdir)),
-                    reason='temp directory does not exist')
-def test_load_xstable_model_fails_with_dir():
+def test_load_xstable_model_fails_with_dir(tmpdir):
     """Check that the function fails with invalid input: directory
 
-    The temporary directory is used for this (the test is skipped if
-    it does not exist).
+    The temporary directory is used for this.
     """
 
     ui.clean()
     assert ui.list_model_components() == []
     with pytest.raises(IOError):
-        ui.load_xstable_model('tmpdir', tmpdir)
+        ui.load_xstable_model('tmpdir', str(tmpdir))
 
     assert ui.list_model_components() == []
 
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
@@ -31,21 +31,27 @@ import pytest
 # from numpy.testing import assert_almost_equal
 
 from sherpa.astro import ui
-from sherpa.utils import requires_data, requires_fits
+from sherpa.utils import requires_data, requires_fits, requires_xspec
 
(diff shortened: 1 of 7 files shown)


DougBurke:
Oops - I meant to merge the fixup commit (26076ca) into an earlier commit. I may therefore rebase this PR.

DougBurke:
So, looks like there's issues over scope - e.g. when can `sherpa.astro.xspec` be imported. I think there may also be issues over the I/O backends (when can you import them, since `sherpa.astro.io.__init__.py` has a warning about importing this module directly).

DougBurke:
Rebased to fix up commits (to avoid partially/fully broken commits). Tip was 0f9799a

DougBurke:
Rebased to fix up failing test suite and to see if Travis will actually pick up the build. Tip was c42cfcd.

DougBurke:
I agree that there should ideally be a test for this warning, but I don't know how to do it. I got as far as 
adding the warning to the list of "things to be ignored" in `sherpa/conf.py`, but that is definitely not a test!

olaurino:
Closing in favor of #307.

PR #299, opened 2016-11-04
Add support for the XSPEC nlapec model (fix #228)

### LINKED ISSUE #228: XSPEC module missing nlapec model from XSPEC 12.9.0
Two issues here:
1. the XSPEC module is missing the `nlapec` model https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSmodelNlapec.html added in XSPEC 12.9.0
2. how do we deal with "optional" models (e.g. this particular model isn't in XSPEC 12.8.2)?

At this time I would say that we drop support for XSPEC 12.8.2, since there are known issues with this version, but it is something to think about (e.g. #52), since we could come across this with a new version of XSPEC.

For this _particular_ model - e.g. `nlapec` - it is possible to use the `apec` model together with the poorly-documented `APECNOLINES` (_note_ not sure of this name) xset variable to replicate it, so there is a work around in this case.

DougBurke:
# Status

Finished.

# Summary

Provide support for the XSPEC nlapec model: if the underlying XSPEC model library does not support the model (it was introduced with XSPEC version 12.9.0) then the model component will error out with the message "The XSPEC nlapec model is not available" when it is evaluated.

This addresses #228.

# Details

There does not appear to be a nice, clean, way to do build-time
detection of the presence (or absence) of a model in the XSPEC library.
This is in part because there's no guarantee that the runtime environment
is the same as the build time one (i.e. you can build Sherpa against one
version of XSPEC and then point to a different one when you use Sherpa).
It it also because the two ways to check if the model are present are:

1) check it is present in libXSFunctions, but it is not just as simple
   as using ctypes.cdll.LoadLibrary(<path to this>) since it requires
   symbols from other modules

2) parse the model.dat file and look for it there, but the problem with
   this approach is that (at present) there is no requirement for this
   file to be present when building the XSPEC module in Sherpa.

This patch goes with option 2, by guessing where the model.dat file
is (if it can't be found then the nlapec model is assumed to not
exist).

The patch adds the model class `sherpa.astro.xspec.XSnlapec` - i.e. users of the `sherpa.astro.ui` API can use `xsnlapec.mdl` to create a model component - which is present even if XSPEC does not support the model. It is only when the model is evaluated that a `sherpa.utils.err.ModelErr` is raised indicating that the model is not available. As we don't have an existing concept of "optional model" I did not want to go overboard in engineering support for it (e.g. adding a class attribute to indicate if the model is actually usable or some other approach).

### DIFF
--- a/helpers/xspec_config.py
+++ b/helpers/xspec_config.py
@@ -17,6 +17,7 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+import os
 
 from numpy.distutils.core import Command
 from .extensions import build_ext, build_lib_arrays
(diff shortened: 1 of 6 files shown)


DougBurke:
Rebased to fix up a broken commit. Tip was d2a903e

olaurino:
@DougBurke do we want to have this refactored and included in 4.9.1?

helpers/xspec_config.py
```diff
@@ -25,6 +26,58 @@ def clean(xs):
...
+        ../spectral/manager/model.dat.
+
+    Returns
+    -------
+    flag : bool
+        True if model.dat can be found and nlapec is found in it,
+        False otherwise.
+    """
+
+    fname = os.path.join(libdir, '../../spectral/manager/model.dat')
```
olaurino: No matter what I did I couldn't get the tests to pass... it took me a while to find the culprit: this line is hardly portable... in my setup, where I was using the libraries from `/soft/ciao/ots/lib`, the `fname` was a non-existent path, so the macro was not being set, so the model function was not built.

DougBurke: Perhaps this location should be a config option in the `xspec` section of `setup.cfg`. 

Or we use the HEADAS environment variable and say it's `$HEADAS/../manager/model.dat`, but then this fails on CIAO builds, where it appears to be `$HEADAS/manager/model.dat` (or `$CIAO_HEADAS/...`). 

So the configuration option sounds best (from my not-really-looing-too-closely approach). Note that we don't necessarily want the path to the model.dat file, but to the directory containing the `model.dat` file, as we may one day want to also parse `mixmodel.dat`

olaurino: Note that the model is not built, but then in the tests the condition for assuming nlapec is not there is to check the version of xspec, which is why the tests fail.

olaurino:
This way of supporting optional models is unsustainable.

To fix #299 we just need to add the model and drop support for 12.8.2 (Which we did already, as we don't test anymore against this version).

If we want to support different versions of xspec we need to find a better approach. The best one would be obviously to generate the xspec extension and models automatically. Alternatively we can try to come up with a way to "manually" add optional models, but it must be a more sustainable way than the one in this PR.

I'd be happy to start working on one of these options, but I wonder what the priority would be, and really which path we want to take.

For now, we just need to add the nlapec model. Would you like me to do that?

DougBurke:
@olaurino I'm happy for you to write a patch that just adds in the nlapec model (and notes in the documentation that XSPEC 12.9.0 is the minimum-supported version). This also allows the XSPEC test to be cleaned up a little, since we no-longer have to avoid evaluating the `XSnteea` model.

I think that's all we can hope for for the 4.10 release (I'll ping Keith and see if they have any concrete release plans for XSPEC for the rest of the year, since this could tweak these priorities).

DougBurke:
Should I close this PR now that we have #390, and delete the branch?

olaurino:
Closing in favor of #390.

PR #305, opened 2016-11-10
fix #303 - fix python 3 issues with some session commands

### LINKED ISSUE #303: Python 3 issues with `set_*log` and `show_bkg_model`
The `set_xlog`, `set_ylog`, and `show_bkg_model` functions raise an exception in Python 3. To reproduce:

```
from sherpa.astro.ui import *
set_xlog()
```

(or `set_ylog()`)

and

```
from sherpa.astro.ui import *
load_arrays(1, [1,2,3], [1,2,3])
show_bkg_model()
```

olaurino:
# Release Note
The `set_xlog`, `set_ylog`, and `show_bkg_model` functions were not compatible with Python 3. This has now been fixed (Issue #303).

# Notes
Admittedly, the tests are rather simplistic. They just check you don't get an exception, which technically speaking is what the bug report pointed out, but clearly not enough as apparently such functions did not have any unit tests to begin with. I would rather focus on other currently open issues and PRs rather than adding proper unit tests for these function, which I don't expect to be a quick feat.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -16,25 +16,11 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-
-from sherpa.ui.utils import Session
-from numpy.testing import assert_array_equal
+from sherpa.astro.ui.utils import Session
 
 
 # bug #303
-def test_set_log():
+def test_show_bkg_model():
     session = Session()
-    session.set_xlog()
-    session.set_ylog()
-
-
-# bug #262
-def test_list_ids():
-    session = Session()
-    session.load_arrays(1, [1, 2, 3], [1, 2, 3])
-    session.load_arrays("1", [1, 2, 3], [4, 5, 6])
-
-    # order of 1 and "1" is not determined
-    assert {1, "1"} == set(session.list_data_ids())
-    assert_array_equal([4, 5, 6], session.get_data('1').get_dep())
-    assert_array_equal([1, 2, 3], session.get_data(1).get_dep())
+    session.load_arrays(1, [1, 2], [1, 2])
+    session.show_bkg_model()
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -17,6 +17,8 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 from sherpa.astro.ui.utils import Session
+from sherpa.utils import requires_data, requires_fits
+from sherpa.models import PowLaw1D
 
 
 # bug #303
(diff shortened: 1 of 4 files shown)


sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -0,0 +1,26 @@
...
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+from sherpa.astro.ui.utils import Session
+
+
+# bug #303
+def test_show_bkg_model():
```
DougBurke: Is it worth adding another test where the id is set explicitly in the call? e.g.

```
def test_show_bkg_model_id():
    session = Session()
    session.show_bkg_model()
    session.show_bkg_model("xx")
```

I ask because I see slightly-different behavior when no data files have been loaded (this is with the current master branch):

```
% python
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
>>> ui.show_bkg_model()
>>> ui.show_bkg_model('xx')
  File "<string>", line 1, in show_bkg_model
  File "/lagado.real/sherpa/sherpa-294/sherpa/astro/ui/utils.py", line 542, in show_bkg_model
    all += self._get_show_bkg_model(id, bkg_id)
  File "/lagado.real/sherpa/sherpa-294/sherpa/astro/ui/utils.py", line 364, in _get_show_bkg_model
    bkg_ids.extend(self._background_sources.get(id, {}).keys())
AttributeError: 'dict_keys' object has no attribute 'extend'
```

Ideally we'd also have a test that included a case where a background model was set up. From a quick look it doesn't look like we have many existing tests - e.g. onlu `sherpa/astro/ui/tests/test_chips_unit.py` and `sherpa/astro/datastack/tests/test_datastack.py` seem to use the `set*_bkg_model`, and `sherpa/astro/ui/tests/test_serialize.py` use `set_bkg_source`

olaurino: Good catch. I'll also try and include a test with the background set.

sherpa/ui/tests/test_session.py
```diff
@@ -21,6 +21,13 @@
 from numpy.testing import assert_array_equal
 
 
+# bug #303
+def test_set_log():
```
DougBurke: Can you also add

```
def test_set_linear():
    session = Session()
    session.set_xlinear()
    session.set_ylinear()
```

(or add the set_x/ylinear calls to `test_set_log`?) I don't expect there to be a problem there, but these functions are directly related to the set_x/ylog calls, so would be good to test both).

I didn't want to add to the PR directly in case you were working on it for my previous comment.

sherpa/astro/ui/tests/test_astro_session.py
olaurino: It looks like it doesn't make much of a difference when the bug is fixed, but I also added `show_bkg_source()` and `show_bkg_source('xx')`.

sherpa/ui/tests/test_session.py
DougBurke: If you did want to make these tests a bit more functional, you can check the setting of the get_*_plot_prefs dictionary before and after the call - e.g. as shown below:

```
>>> from sherpa.astro.ui.utils import Session
>>> session = Session()
>>> session.get_data_plot_prefs()['xlog']
False
>>> session.get_data_plot_prefs()['ylog']
False
>>> session.set_xlog()
>>> session.get_data_plot_prefs()['xlog']
True
>>> session.get_data_plot_prefs()['ylog']
False
>>> session.set_ylog('model')
>>> session.get_data_plot_prefs()['ylog']
False
>>> session.get_model_plot_prefs()['ylog']
True
>>> session.set_ylog()
>>> session.get_data_plot_prefs()['ylog']
True
>>> session.get_model_plot_prefs()['ylog']
True
>>> session.set_ylinear()
>>> session.get_data_plot_prefs()['ylog']
False
>>> session.get_model_plot_prefs()['ylog']
False
```
I'll see if I can come up with a simple example test (or two).

PR #311, opened 2016-11-18
Fix over-zealous code clean up in PR #287 affecting sigmarej

DougBurke:
# Summary

Fits using the sigmarej iterated-fit method were broken if a filter
had been applied to the data before the fit and there are any bins
that get ignored at larger bin values than the filtered-out data. This
bug was introduced by PR #287.

# Details

I added a test, checked that it failed before fixing the code and
fixed after the code. This deals with the "while not(newmask...)" line
change.

I manually checked the change for the "if current_par.frozen:" change,
by inserting errors in both branches and checking the test suite errored
out.

The last change - "not(any(newmask))" - does not have any test as it
is for an error condition and it is not obvious how to create a test
that triggers this.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -806,7 +806,7 @@ class IterFit(NoNewAttributesAfterInit):
                     j = 0
                     kmin = 0
                     for i in xrange(0, ressize):
-                        while newmask[j] is False and j < filsize:
+                        while not(newmask[j]) and j < filsize:
                             j = j + 1
                         if j >= filsize:
                             break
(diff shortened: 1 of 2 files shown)


olaurino:
I would approve this PR, but it needs to be issued against the release/ciao4.9 branch (well, closed and re-issued). @DougBurke would you like to do that? Or should I?

DougBurke:
@olaurino it would be a lot quicker if you could do it. Thanks.

olaurino:
Closing in favor of #312.

Issue #240, opened 2016-08-22
pytest warning due to use of deprecated functionality


DougBurke:
I just tried sherpa master with the latest anaconda setup (see below for details) and got this warning (fortunately the 'yellow text on a white background' hasn't been copied).

```
============================ pytest-warning summary ============================
WC1 None passing a string to pytest.main() is deprecated, pass a list of arguments instead.
========== 273 passed, 54 skipped, 1 pytest-warnings in 65.89 seconds ==========
```

I couldn't easily see which test caused this (due to a small scroll-back buffer for my history). Here's the setup I used

```
% conda create -n=sherpa-master python=2.7 matplotlib astropy ipython
% source activate sherpa-master
% python setup.py develop
% python setup.py test
...
% conda list
# packages in environment at /some/path/anaconda/envs/sherpa-master:
#
Using Anaconda Cloud api site https://api.anaconda.org
astropy                   1.2.1               np111py27_0  
backports                 1.0                      py27_0  
cairo                     1.12.18                       6  
cycler                    0.10.0                   py27_0  
decorator                 4.0.10                   py27_0  
fontconfig                2.11.1                        6  
freetype                  2.5.5                         1  
get_terminal_size         1.0.0                    py27_0  
glib                      2.43.0                        1  
harfbuzz                  0.9.39                        1  
ipython                   5.1.0                    py27_0  
ipython_genutils          0.1.0                    py27_0  
libffi                    3.2.1                         0  
libpng                    1.6.22                        0  
libxml2                   2.9.2                         0  
matplotlib                1.5.1               np111py27_0  
mkl                       11.3.3                        0  
numpy                     1.11.1                   py27_0  
openssl                   1.0.2h                        1  
pango                     1.39.0                        1  
path.py                   8.2.1                    py27_0  
pathlib2                  2.1.0                    py27_0  
pexpect                   4.0.1                    py27_0  
pickleshare               0.7.3                    py27_0  
pip                       8.1.2                    py27_0  
pixman                    0.32.6                        0  
prompt_toolkit            1.0.3                    py27_0  
ptyprocess                0.5.1                    py27_0  
pycairo                   1.10.0                   py27_0  
pygments                  2.1.3                    py27_0  
pyparsing                 2.1.4                    py27_0  
pyqt                      4.11.4                   py27_4  
python                    2.7.12                        1  
python-dateutil           2.5.3                    py27_0  
pytz                      2016.6.1                 py27_0  
qt                        4.8.7                         4  
[...]

PR #318, opened 2016-11-21
remove requires_data decorator from make_data_path fixture

olaurino:
This fixes the CIAO issue SH-3.

The `make_data_path` fixture was decorated with `requires_data` so to simplify the creation of tests that use the fixture, which are assumed to require external data. By decorating the fixture, tests were skipped without requiring the tests themselves to be decorated.

However, this does not seem to work in CIAO, were the tests are always skipped.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -21,7 +21,7 @@ import pytest
 import os
 import sys
 import re
-from sherpa.utils import SherpaTestCase, requires_data
+from sherpa.utils import SherpaTestCase
 
 from six.moves import reload_module
 
@@ -148,7 +148,6 @@ def pytest_configure(config):
 
 
 @pytest.fixture(scope="session")
-@requires_data
 def make_data_path():
     """
     Fixture for tests requiring the test data dir. It returns a function that can be used to make paths by using
--- a/sherpa/models/tests/test_template_unit.py
+++ b/sherpa/models/tests/test_template_unit.py
@@ -25,8 +25,10 @@ import pytest
 import six
 
 from sherpa.ui.utils import Session
+from sherpa.utils import requires_data
 
 
+@requires_data
 def test_309(make_data_path):
 
     idval = 'bug309'


DougBurke:
I recently noted that some tests which used `make_data_path` seemed to imply `requires_data` which I thought was a useful simplification. I am concerned that I may have started to assume this in my tests.

However, I guess that as the travis tests still pass with this change then there can't be any such problem cases.

olaurino:
I believe our Travis builds end up skipping tests that require data even if they don't have the decorator, because they are skipped for other reasons. Do we want to create a new Travis build to try and exercise this case, maybe in this PR? Or do we just want to figure out what tests are missing the decorator locally and use this PR to decorate them?

DougBurke:
We are waiting for travis config to be rebuilt to provide better testing of various build options before accepting this PR.

olaurino:
@DougBurke does build answer our questions?
https://travis-ci.org/sherpa/sherpa/jobs/201706463

olaurino:
Rebasing, original tip was 7e2a0a9.

I expect Travis to fail because a test is missing the `requires_data` decorator and the new travis builds include a bare bone build with no test data and no optional dependencies.

DougBurke:
@olaurino so is the idea that you'll now add back in the decorator you removed in 19933ee and show that the tests now pass?

olaurino:
exactly.

olaurino:
I am also piggybacking this PR to fix the second part of #240 with a new commit.

helpers/test.py
```diff
@@ -51,7 +51,7 @@ def run_tests(self):
             import pytest
             if not self.pytest_args:
                 self.pytest_args = ['sherpa',]
-            errno = pytest.main(self.pytest_args)
+            errno = pytest.main([self.pytest_args,])
```
DougBurke: If the previous condition was fired, then this is going to

    pytest.main([['sherpa',], ])

Is this okay?

olaurino: I thought about that right after I committed. It's probably not right, let's see what Travis thinks. I'll remove the commit if that's the case.

DougBurke: I see the tests for the previous stuff have failed anyway; looks like missing the coverage module?

Isn't automated testing fun?

olaurino: Nope, I shouldn't have pushed the commit before the PR builds ran. PR builds do not use the specific commit, but the hidden PR references, so when you update the PR while the builds are still running, the builds use the latest code. I'll remove the latest commit and rerun the builds.

DougBurke:
@olaurino what's the status of this PR now?

olaurino:
I am working on other priorities, I'll get back to this hopefully tomorrow.

olaurino:
In the meantime, I restarted the Travis PR build.

DougBurke:
Yay - it passes again!

DougBurke:
Did you add in the fix for the second half of #240, or has that been removed?

olaurino:
It's been removed. It's more complicated than I expected, so I didn't want to add risk/complexity to the rather simple change in this PR.

DougBurke:
I don't know why the CIAO build is failing, and it seems to me that really we should fix that rather than remove the decorator, but as the tests pass here I think we should just merge this PR (I guess there may be times when you want to allow the use of `make_data_path` even without the data being available).

olaurino:
It looks like with pytest some command line calls do not behave like others, all things being equal. In CIAO testing conditions are different than in the standalone realm, so we have to rely on a "manual" command line configuration that seems to have an impact on the order in which stuff in conftest.py is loaded. The time required to debug the pytest issue, figure out if it's a bug in the first place, and report it to them is much more than the one required to work around the behavior, which might be expected to begin with.

olaurino:
To be more precise, here is the command line used to run the tests in CIAO:

> py.test /proj/xena/ciaod/Linux64/ciao_0301/lib/python2.7/site-packages/sherpa-4.9.0+33.g5402a75-py2.7-linux-x86_64.egg/sherpa --test-data /data/regression_test/master/in/sherpa --verbose

For some reason when run like this, i.e. referring py.test to the folder where sherpa is rather than loading sherpa as a module as we do in standalone realm, `sherpa` as a module is imported after the `conftest.py` file has been processed by pytest. Which means that the data folder is searched only after `conftest.py`'s fixture `make_data_path` is defined, which is when the decorator is firstly (and solely) used. Note that the other difference is that the test data is neither imported as a package nor available in the expected relative path in the sherpa folder, which is why we introduced the `--test-data` option in CIAO.

The result is that all tests requiring data are skipped in CIAO if the fixture (rather than the tests) is decorated.

This could probably be reproduced in standalone world, but for some mysterious reason I never managed to do it... Note that this is reverting back something that I had done to simplify things... after this PR things go back to what they were. All in all it is much much easier and less time consuming to have to remember to put the decorator than even just describing this pesky issue ;-)

PR #326, opened 2016-12-22
Skip test if no plotting backend is installed

DougBurke:
# Summary

Skip the `sherpa.ui.tests.test_session.test_set_log` test when no plotting backend is installed.

# Details

The test added to fix bug #303 fails if there is no plotting backend available, with the following error:

```
=================================== FAILURES ===================================
_________________________________ test_set_log _________________________________

    def test_set_log():
        session = Session()
>       assert not session.get_data_plot_prefs()['xlog']
E       KeyError: 'xlog'

sherpa/ui/tests/test_session.py:32: KeyError
```

The fix is to create a decorator that skips tests if neither chips or matplotlib is installed and apply it to this test. For now I have decided not to export this decorator directly from `sherpa.utils`.

PR #335, opened 2017-02-01
Fix `setup.py` install command

olaurino:
# Release Note
The `setup.py install` command was not enforcing the installation of the dependencies listed in `setup.py`. This has been fixed.

# Description
It was reported that `six` would not be installed with Sherpa when doing a simple `python setup.py install`. It turns out one needs to call a special method from an extension of the `install` command for this to work.

### DIFF
--- a/helpers/install.py
+++ b/helpers/install.py
@@ -27,4 +27,4 @@ class install(_install):
     def run(self):
         self.get_finalized_command('xspec_config', True).run()
         self.get_finalized_command('sherpa_config', True).run()
-        _install.run(self)
+        _install.do_egg_install(self)


DougBurke:
The error was (so I can remember what to test), that in a clean conda environment with numpy installed:

    python setup.py install
    sherpa_test

would fail with an error like

    pkg_resources.DistributionNotFound: The 'six' distribution was not found and is required by sherpa

olaurino:
Dependencies are installed by the `pip install test_dependencies.txt` command in travis. I'll try and see if I can (re)move that command, which might just be a relic of when we didn't use pytest. I'll make this part of the travis updates. If you want we can wait for travis to fail without merging this PR so we make sure this PR fixes that issue.

DougBurke:
I'd like to wait and make sure the tests fail on travis before applying the fix, so that we are clear what has and hasn't happened.

olaurino:
This is a build resulting from a deep clean up of the travis file (which is still wip), building on top of #326. Among other things, I removed the pip install of the dependencies, and you can see the error you reported:
https://travis-ci.org/olaurino/sherpa/jobs/197785379

Here is the travis configuration of the failing build:
https://github.com/olaurino/sherpa/blob/update-travis/.travis.yml#L41

DougBurke:
Great. So is the idea we wait for your WIP rewrite of the Travis configuration file and then rebase this PR onto that, or merge this and then the Travis rewrite will get it for free (after it has been rebased)?

olaurino:
I don't think I care too much... but I'll confirm this PR fixes that build and then I just think we should merge this one. I'll issue a PR with the new Travis configuration when it looks ready to go, but we'll maybe iterate over it to make sure it covers what it should.

olaurino:
Compare:
https://travis-ci.org/olaurino/sherpa/builds/197816842

with:
https://travis-ci.org/olaurino/sherpa/builds/197870918

Where the only difference is that the second build has seen this PR merged. I hope this is enough evidence that this PR did indeed fix the issue :)

olaurino:
This has the pesky side effect of breaking `bdist_wheel`, which is run by `pip` when using the PyPI package. I currently don't have a solution other than reverting this PR back. setuptools commands are still a mystery to me.

olaurino:
I might have a fix, but while I am getting inspiration from the setuptools source code itself, it seems like a hack. I'll push it tomorrow morning for review.

PR #337, opened 2017-02-12
Feature/update fit documentation

DougBurke:
# Summary

Update the documentation in the `sherpa.fit` module.

# Details

This updates the docstrings (both existing documentation and adding new ones), adds a few comment strings,  and uses the `functools.wraps` function to create a decorator that copies over the function signatures as well as the docstring.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -23,6 +23,9 @@ from six.moves import xrange
 import logging
 import os
 import signal
+
+from functools import wraps
+
 from numpy import arange, array, abs, iterable, sqrt, where, \
     ones_like, isnan, isinf, float, float32, finfo, nan, any, int
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
@@ -42,18 +45,15 @@ __all__ = ('FitResults', 'ErrorEstResults', 'Fit')
 
 
 def evaluates_model(func):
+    """Fit object decorator that runs model startup() and teardown()
     """
-    Fit object decorator that runs model startup() and teardown()
-
-    """
+    @wraps(func)
     def run(fit, *args, **kwargs):
         fit.model.startup()
         result = func(fit, *args, **kwargs)
         fit.model.teardown()
         return result
 
-    run.__name__ = func.__name__
-    run.__doc__ = func.__doc__
     return run
 
 
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -653,13 +653,13 @@ class IterFit(NoNewAttributesAfterInit):
         Raises
         ------
         sherpa.utils.err.FitErr
-            This exception is raised if the statistic is not
-            supported. This method can only be used with
-            Chi-Square statistics with errors.
+            This exception is raised if the statistic is not supported.
+            This method can only be used with Chi-Square statistics
+            with errors.
 
         Notes
         -----
-        The following keys are looked for in the ``itermethod_opts``
+        The following keys are looked for in the `itermethod_opts`
         dictionary:
 
         ========  =========  ===========
@@ -670,19 +670,21 @@ class IterFit(NoNewAttributesAfterInit):
         maxiters  int        The maximum number of iterations.
         ========  =========  ===========
 
-        The variance in bin i is estimated to be:
+        The variance in bin i for iteration j is estimated to be:
 
         .. math::
 
-           \sigma(j)_i^2 = S(i, t_s(j-1)) + (A_s/A_b)^2 B_{off}(i, t_b(j-1))
+           \sigma(j)_i^2 = S(t_s(j-1))_i + (A_s/A_b)^2 B(t_b(j-1))_i
 
-        where j is the number of iterations that have been carried out
-        in the fitting process, B_off is the background model
-        amplitude in bin i of the off-source region, and t_s(j-1) and
-        t_b(j-1) are the set of source and background model parameter
-        values derived during the iteration previous to the current
-        one. The variances are set to an array of ones on the first
-        iteration.
+        where the subscript i indicates the bin number, j is the
+        number of iterations that have been carried out in the
+        fitting process, S is the model-predicted amplitude of
+        the source in the bin, B is the background (off-source)
+        model amplitude, A_s and A_b are the source and background
+        areas (or scaling factors), and t_s and t_b are the set
+        of source and background model parameter values derived in
+        the previous iteration (indicated by the j-1 subscipt).
+        The variances are set to an array of ones on the first iteration.
 
         In addition to reducing parameter estimate bias, this
         statistic can be used even when the number of counts in each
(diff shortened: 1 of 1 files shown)


DougBurke:
The test failure is to the log overflowing on one test case. I don't believe this is anything to do with this PR, but am not sure how to confirm this statement.

olaurino:
I restarted the offending build... I am guessing wget and conda might log more stuff if the network is slow, and that contributes to the logs. In my rework of the travis configuration I am silencing conda and wget as much as I can.

sherpa/fit.py
```diff
@@ -23,6 +23,9 @@
 import logging
 import os
 import signal
+
+from functools import wraps
```
olaurino: Neat. You can always learn something new.

sherpa/fit.py
```diff
@@ -572,18 +646,40 @@ def primini(self, statfunc, pars, parmins, parmaxes, statargs=(),
...
+                             values.
+        maxiters  int        The maximum number of iterations.
+        ========  =========  ===========
 
         The variance in bin i is estimated to be:
 
-        sigma^2_i^j = S(i, t_s^(j-1)) + (A_s/A_b)^2 B_off(i, t_b^(j-1))
+        .. math::
+
+           \sigma(j)_i^2 = S(i, t_s(j-1)) + (A_s/A_b)^2 B_{off}(i, t_b(j-1))
```
olaurino: Hmm... can sphinx render latex equations? If it can, I'd rather leave the original text alone. Either way I am going to trust @DougBurke's and @anetasie's judgment on this, so I won't require any changes to this.

DougBurke: This output be seen at http://hea-www.harvard.edu/~dburke/playground/sherpa/fit/api/sherpa.fit.IterFit.html#sherpa.fit.IterFit.primini - so it is being rendered via a LaTeX-like system. There is a discussion to be had over how much extra markup we want to include in the docstrings to make the HTML output nicer while making the docstrings harder to read.

There's actually two reasons for changing this equation:

a) making it latex-able
b) changing the syntax - it appears that t_a^b was being used to say "t_a" at iteration b, and I've changed this to "t_a(b)" which I think is a bit less confusing, but it'll be interesting to see what @anetasie thinks

sherpa/fit.py
DougBurke: I was glad to find it available for both Python 2.7 and 3.5, as I didn't fancy writing the logic to copy over the function syntax as part of the decorator. We should probably review elsewhere we use decorators to see if we can use `functools.wraps`, but I decided that was not for this PR.

sherpa/fit.py
olaurino: Shouldn't it be `\sigma_i^2(j)`, or `\sigma^2_i(j)` though?

DougBurke: There's a number of different ways it could be done (including leaving it as it was), and I'm not sure what's best. So this discussion is good, if it comes to a conclusion...

DougBurke:
I see the build still fails. Looking at the build you can see that numpy keeps on getting up/downgraded, which adds to noise and why your rewrite is eneded @olaurino, but it's not obvious to me what is causing the failure here.

olaurino:
I thought I had already consolidated the conda install lines, but I see that's not true. I'll add a commit to #338.

olaurino:
I'll try to rebase it once #338 has been merged (not sure I can force push to @DougBurke's branch, though).

@DougBurke your request was for one between me and @anetasie to review this? Or did you mean we should both do it? I was expecting @anetasie to review this, but maybe I misunderstood.

DougBurke:
I was hoping @anetasie could find time to look at this. I think the documentation is lower-level than is her normal interest for Sherpa, but it is partly about how we document it, and how "unreadable" we want to make the docstrings in the aim of giving "nice" HTML.

olaurino:
Rebasing this. Original tip was 9a2b55c.

olaurino:
@DougBurke assuming Travis is OK with the rebase, should I merge this or should I wait for @anetasie's review?

DougBurke:
I'll see if I can get @anetasie to look at it.

sherpa/fit.py
anetasie: One can make the indexing in the subscripts`\sigma_{i,j}^2`
I would prefer the indices not be enclosed inside the brackets, but if both j and i are defined than it should be ok.

anetasie: S needs to be defined - model predicted source amplitude (counts) in a bin i

sherpa/fit.py
```diff
@@ -1128,6 +1327,56 @@ def simulfit(self, *others):
```
anetasie: If I understand this correctly `est_error()` calculates the confidence bounds for the model parameters. Suggest to change to

'Calculate the low and high confidence bounds for one or more of the thawed parameters in the fit.'

sherpa/fit.py
DougBurke: One of the reasons I changed it was because sphinx/mathjax/whatever didn't like the original format when treated as latex. It also wasn't immediately clear to me that the ^j and ^(j-1) were referring to iteration number, as I was thinking exponentiation (this was without really reading the text).

The original was (I believe)

    sigma^2_i^j = S(i, t_s^(j-1)) + (A_s/A_b)^2 B_off(i, t_b^(j-1))

so if someone can come up with a "better" form then I'll use it.

DougBurke:
Rebasing. Original tip was 3959ed0.

DougBurke:
The latest version of the formula looks like the following screen shot; hopefully this is enough to get this accepted as this PR has been sitting around long enough.

this

anetasie:
This looks good. My only comment is my initial confusion of using 't' for a set of model parameters. I guess this is sort of like 't' for 'theta, with 'theta' for a set of model parameters in the notation in the stats books. So it is OK.

DougBurke:
@anetasie - I was trying to make minimal changes to the original, which used the `t` symbol:

```
    The variance in bin i is estimated to be:
    
    sigma^2_i^j = S(i, t_s^(j-1)) + (A_s/A_b)^2 B_off(i, t_b^(j-1))
    
    where j is the number of iterations that have been carried out
    in the fitting process, B_off is the background model
    amplitude in bin i of the off-source region, and t_s^(j-1) and
    t_b^(j-1) are the set of source and background model parameter
    values derived during the iteration previous to the current
    one. The variances are set to an array of ones on the first
    iteration.
```

DougBurke:
Doh. I can't spell subscript. It is also not a subscript, so I'll have to remove it anyway.

olaurino:
@anetasie @DougBurke: Is this settled?

DougBurke:
I hope so, but I am somewhat biased!

olaurino:
Alright, I am merging this. We can fix it later if really necessary.

PR #338, opened 2017-02-14
Update travis

olaurino:
This PR consolidates Travis builds and updates most dependencies, in particular:
  * language is now C, which is more appropriate for a project that uses conda to get Python. Also, the C environment reduces the number of required build dependencies, in fact:
  * apt dependencies (used for all non-xspec builds) have been reduced.
  * smoke test is now run at the end of all builds, in addition to the other tests. The test probes for the dependencies that are known to be installed in each build, so e.g. if astropy is installed the smoke test checks the module can be imported so the proper tests are run... if it can't, it will fail.
  * we now test for python 2.7, 3.5, and 3.6, although the 3.6 build is still allowed to fail (and it does fail because of some unhandled warnings).
  * numpy v1.11 and up are being tested (i.e. 1.9 and 1.10 are not tested anymore)
  * pyfits build has been removed
  * there is now a bare bone 2.7 build with no optional dependencies and no test data, to make sure most tests are properly skipped and dependencies properly installed in this particular case (see #326 and #335)
  * libgfortran dependency has been updated to 3.0 (1.0 was needed by the pyfits package and its obsolete dependencies)
  * ds9 was updated to version 7.5, and xpa to version 2.1.17
  * conda now installs quietly, hoping this reduces the size of the logs on Travis
  * `pip install -r test_requirements.txt` is not run explicitly anymore. We now have machinery that should always install dependencies.
  * builds are less verbose
  * there are fewer builds. Hopefully they still probe all useful cases (although the matrix does not cover all the possible combinations), in particular:
     * full builds (including xspec, ds9, and all of the optional dependencies) are tested with python 2.7, 3.5, and 3.6
     * a bare bone build checks tests are properly skipped when no optional dependencies and no test data are installed
     * both `develop` and `install` commands are used
     * test data is not installed, it is installed as a package, or it is available as a submodule
     * there is a build with only matplotlib installed (no xspec, no astropy)
     * there is a build with only astropy installed (no xspec, no matplotlib)
     * numpy is installed as both v 1.11 and at its latest version available in conda
     * matplotlib is installed as 1.5 and 2.0

Two more changes are piggybacking in this PR:
  * a fix to #240.
  * a fix to the smoke test, which did not work on Python3 if the verbosity argument was passed to it.

Note: xspec has not been updated to 12.9.1 because the package is too big, and it easily takes more than 10 minutes to download, which results in the build failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,25 +36,35 @@ matrix:
     - env: XSPECVER="12.9.0i" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
 
 before_install:
+  # General configurations
   - export LIBGFORTRANVER="3.0"
   - export SHERPA_CHANNEL=sherpa
   - export XSPEC_CHANNEL=cxc/channel/dev
   - export MINICONDA=/home/travis/miniconda
+
+  # Install, update, and configure conda
   - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
   - ./miniconda.sh -b -p $MINICONDA
   - export PATH=$MINICONDA/bin:$PATH
   - conda update --yes conda
+  - conda config --add channels ${SHERPA_CHANNEL}
+  - conda config --add channels ${XSPEC_CHANNEL}
+
+  # Figure out requested dependencies
   - if [ -n "${MATPLOTLIBVER}" ]; then MATPLOTLIB="matplotlib=${MATPLOTLIBVER}"; fi
   - if [ -n "${NUMPYVER}" ]; then NUMPY="numpy=${NUMPYVER}"; fi
   - if [ -n "${XSPECVER}" ];
      then export XSPEC="xspec-modelsonly=${XSPECVER}";
     fi
   - echo ${MATPLOTLIB} ${NUMPY} ${FITS} ${XSPEC}
-  - conda config --add channels ${SHERPA_CHANNEL}
-  - conda config --add channels ${XSPEC_CHANNEL}
+
+  # Create and activate conda build environment
+  # We create a new environment so we don't care about the python version in the root environment.
   - conda create --yes -n build python=$TRAVIS_PYTHON_VERSION pip ${MATPLOTLIB} ${NUMPY} ${MKL} libgfortran=$LIBGFORTRANVER $XSPEC $FITS
   - source activate build
+
+  # XSPEC and DS9
   - if [ -n "${XSPECVER}" ];
      then DS9_SITE=http://ds9.si.edu/download/centos6/;
      XPA_SITE=http://ds9.si.edu/download/centos6/;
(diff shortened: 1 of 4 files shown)


olaurino:
@DougBurke I was not planning to work on this just yet, but given the current failures in travis I thought we needed this PR as a baseline.

olaurino:
I now remember why I stopped commenting lines in the travis configuration file. I forgot, I started commenting the file back again and now I got an error. Bottom line: I haven't figured out a way to add comments in blocks. I added some comments where they didn't make travis choke nevertheless.

olaurino:
Bad news. One of the xspec builds keep failing with the log overflow issue on this PR.

DougBurke:
@olaurino I haven't looked at this yet. Apologies. One thing I'm interested to know is how easy is it to expand the new setup. I ugly-hacked support for sphinx into the old system - e.g. https://github.com/DougBurke/sherpa/blob/4315acd235d5932feb14b30ceaed98337bee0cb7/.travis.yml#L37 - and is this going to be cleaner/easier with the new setup?

olaurino:
As far as I can tell, it's pretty much the same. I took the opportunity to consolidate the builds, address some long standing issues, remove obsolete stuff, but the approach is the same. We could clean things up by refactoring the logic in standalone shell scripts, but the configuration is easy enough that I prefer keeping it in a single file, despite the ugliness of a shell script written in yaml.

.travis.yml
```diff
@@ -1,122 +1,117 @@
...
+    # Install sherpatest package rather than relying on relative location of the submodule
+    # Also, install matplotlib 2.0 and do not install astropy (checks tests are properly skipped)
+    - env: INSTALL_TYPE=install TEST=package MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="2.7"
+    # Install astropy without matplotlib (checks tests are properly skipped)
+    - env: INSTALL_TYPE=develop TEST=submodule FITS="astropy" TRAVIS_PYTHON_VERSION="3.5"
   allow_failures:
-    - env: XSPECVER="12.8.2q" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5
-
-env: FITS="pyfits" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.4 # 1.5 does not support numpy 1.8, required by pyfits
+    - env: XSPECVER="12.9.0i" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
```
DougBurke: Isn't this the same as the test above at line 27?

olaurino: Yes, which is required by the travis config specification:
https://docs.travis-ci.com/user/customizing-the-build#Matching-Jobs-with-allow_failures

DougBurke: Ahhh. I had taken "allow failures" to be a place that you can specify runs that can fail, rather than this subtly-different meaning.

We need a comment in the allow-failures section to say why we allow this to fail (so we remember in 6 months time when we are reviewing the build).

olaurino: > I had taken "allow failures" to be a place that you can specify runs that can fail

Well, that's what it is, it's just that you need to match the exact same configuration as the build itself, so it reads somewhat redundant.

olaurino:
I just added a comment.

olaurino:
The master branch is currently failing because of issues fixed by this PR. @DougBurke I'll merge this PR since I believe I added the comment you were asking for and I am not sure if you are working today and can respond to this.

DougBurke:
Thanks - the comment you added addresses my requirements.

PR #302, opened 2016-11-07
add sphinx documentation (fix #197)

DougBurke:
This PR is ready for review, although I am likely to try and improve the documentation.

Summary
========

Create the Sherpa documentation using Sphinx to build, Travis-CI to test, and Read The Docs to build and display. The aim for the documentation is to build a sensible skeleton, but not to have a feature-complete set of documentation.

Notes
=====

There are three parts to getting sphinx documentation

  1. the infrastructure to get sphinx to build the documentation
  2. the contents of the documentation (mainly the .rst files but also fixes and additions to the
      module docstrings)
  3. getting docs built and available to the community

Status
=====

The outline of the documentation is present, as are much of the concepts, but there are missing areas and areas that need work.

All three points above have been addressed in this PR. I have set up a test build on Read The Docs - https://sherpa-test.readthedocs.io/en/latest/ - which requires manual re-building when new commits are made (so may not reflect the latest changes). Once the PR is integrated into the Sherpa branch I have a Read The Docs account set up to build the documentation from the master branch on every commit.
It is possible to use this PR but build the documentation in some other manner (e.g. via GitLab) as there are only a few small, text-only, configuration files specific to the Read The Docs environment:

  - `readthedocs.yml`
  - `docs/environment.yml`
  - `docs/rtd-pip-requirements`

In order to get the documentation to build on Read The Docs, which does not support "install any package you need", the build is done without requiring a functional Sherpa installation. It does this by mocking required packages *and* the compiled modules within Sherpa. The only required Python packages for the build (other than Sphinx and `sphinx_rtd_theme`) are `numpy` (since `setup.py` requires it), and `six` (although this probably also be mocked).

Previous issues with evidence of mocked symbols - e.g. in the routines provided in `sherpa.astro.xspec`, as described in https://sherpa-test.readthedocs.io/en/latest/model_classes/xspec_model.html - has been fixed.

The output from the example code is *not* auto-generated: this is different to my previous approach in  https://github.com/DougBurke/sherpa/tree/feature/sphinx-docs which was nice, but required the ability to build Sherpa *and* there was some **interesting** interaction between the Sherpa support code and some of the Sphinx utility routines with screen output.

Notes
=====

[...]

PR #339, opened 2017-02-22
Tweak documentation for Sphinx

DougBurke:
This is a documentation-only change. There is no meaningful change to the
documentation; the changes are mostly sphinx-related (to avoid problems
generating HTML documentation, often because the suggested standards are
not 100% clear on what is supported), as well as a few minor clean ups
found while reviewing the documentation.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2010, 2015, 2016, 2017  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -197,7 +197,7 @@ class Session(sherpa.ui.utils.Session):
         Raises
         ------
         sherpa.utils.err.IOErr
-           If ``filename`` already exists and ``clobber`` is ``False``.
+           If `filename` already exists and `clobber` is ``False``.
 
         See Also
         --------
@@ -244,7 +244,7 @@ class Session(sherpa.ui.utils.Session):
         Raises
         ------
         IOError
-           If ``filename`` does not exist.
+           If `filename` does not exist.
 
         See Also
         --------
@@ -412,7 +412,7 @@ class Session(sherpa.ui.utils.Session):
            otherwise it is taken to be the name of the file to
            write the results to.
         clobber : bool, optional
-           If ``outfile`` is not ``None``, then this flag controls
+           If `outfile` is not ``None``, then this flag controls
            whether an existing file can be overwritten (``True``)
            or if it raises an exception (``False``, the default
            setting).
@@ -420,7 +420,7 @@ class Session(sherpa.ui.utils.Session):
         Raises
         ------
         sherpa.utils.err.IOErr
-           If ``outfile`` already exists and ``clobber`` is ``False``.
+           If `outfile` already exists and `clobber` is ``False``.
 
         See Also
         --------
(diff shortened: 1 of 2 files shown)


DougBurke:
I claim the Travis failure here is not due to this PR. I'll update it once #338 is merged, but am asking for a review from @olaurino and @anetasie

sherpa/astro/ui/utils.py
```diff
@@ -2338,10 +2346,10 @@ def set_staterror(self, id, val=None, fractional=False, bkg_id=None):
         val : array or scalar
            The systematic error.
         fractional : bool, optional
-           If ``False`` (the default value), then the ``val`` parameter is
-           the absolute value, otherwise the ``val`` parameter
+           If ``False`` (the default value), then the `val` parameter is
+           the absolute value, otherwise the `val` parameter
            represents the fractional error, so the absolute value is
-           calculated as ``get_dep() * val`` (and ``val`` must be
+           calculated as ``get_dep() * val`` (and `val` must be
```
olaurino: Shouldn't `get_dep() * val` be in single backquotes as interpreted text?

sherpa/astro/ui/utils.py
```diff
@@ -2410,10 +2418,10 @@ def set_syserror(self, id, val=None, fractional=False, bkg_id=None):
         val : array or scalar
            The systematic error.
         fractional : bool, optional
-           If ``False`` (the default value), then the ``val`` parameter is
```
olaurino: Shouldn't `False`, `None`, etc. be in single backquotes like `load_arrays` etc?

DougBurke: The single and double tick marks have a different meaning: `foo` is a link/reference whereas ``foo`` is a "code fragment"; see http://www.sphinx-doc.org/en/stable/rest.html and links therein. We could actually have None/False/... as links, as they could link to the actual Python documentation for those items, but it's not clear whether it's worth it. So we want ``get_dep() * val`` to remain double quotes. 

The trouble with `foo` is that the documentation system can get confused and dump a warning about  an invalid link. I haven't really seen that with my latest attempt at the docs (#302), but I think I saw it in previous attempts.

There's also "tweaks" to the rules. For instance, the astropy (essentially the NumPy ones) docstring rules, which I'm trying to follow, say to use `foo` when referring to parameter values of a function [*], but it's not 100% clear if that is only in the parameter block or in the whole method docstring.

[*] see section 4 of http://docs.astropy.org/en/stable/development/docrules.html#sections

I admit that the current documentation is a mix. I am trying to move towards a consistent style, but haven't ironed out the edges yet.

DougBurke: For reference, the set_staterror call - when rendered in HTML - looks like http://hea-www.cfa.harvard.edu/~dburke/playground/sherpa/ui/api/sherpa.astro.ui.set_staterror.html

(I can't guarantee that this link will remain valid long term, but hopefully this PR will be merged before the link is changed).

DougBurke:
Other examples are the two `unpack_data` calls (as they have slightly-different implementations):

http://hea-www.cfa.harvard.edu/~dburke/playground/sherpa/ui/api/sherpa.ui.unpack_data.html
http://hea-www.cfa.harvard.edu/~dburke/playground/sherpa/ui/api/sherpa.astro.ui.unpack_data.html

The first one is particularly improved by this PR (I unfortunately don't have a HTML rendition of the old version).

olaurino:
Rebasing, original tip was 092c654.

olaurino:
Again, should I merge this (assuming Travis' blessings)?

DougBurke:
Unlike #337 these are more "mechanical" changes rather than substantive changes to the documentation, so I think these can be merged without a review by @anetasie

PR #343, opened 2017-03-06
Fix user statistics regression

### LINKED ISSUE #341: Problem fitting user defined models with custom fit statistic
## Summary

I am unable to fit user defined models with a user defined fit statistic calculation in Sherpa version 4.9.0. This works using version 4.7, but stopped working in all versions starting with 4.8.2 through to 4.9.0. The following file can give you an idea of how I am using sherpa in this fashion, however I have a much simpler example below which will make my real-world example make more sense.
https://github.com/sot/xijafit/blob/f5268236f5f22dead2c2ac570b4b3e9234d66e48/xijafit/fit.py

The methods defined in the file above are derived from code originally written by Tom Aldcroft (@taldcroft) to fit these models, and are used to help fit Ska Xija thermal models for Chandra operations.

I believe the error I get is a result of a change made in the commit labeled "generalize Stat API", included in version 4.8.2 where the calc_stat method in the UserStat class was modified so that a valid/proper return value was removed. Please compare line 233 in the first link to line 519 in the second link.

Version 4.7
Look at the function definition on line 233:
https://github.com/sherpa/sherpa/blob/ba26a5f6e7cb9afa999ee2c16b435a6a12f821be/sherpa/stats/__init__.py

Version 4.8.2
Look at the function definition on line 519:
https://github.com/sherpa/sherpa/blob/183084aee89ec828ce872e4c54ad9e021ae23c54/sherpa/stats/__init__.py

Version 4.9.0
Look at the function definition on line 736:
https://github.com/sherpa/sherpa/blob/aa98d19553d1e62d3c65ba5abf33562e5e7dbe4c/sherpa/stats/__init__.py

This change in question was made in the commit labeled "generalize Stat API":
https://github.com/sherpa/sherpa/commit/576085b736a7aafad524dbfc35056283461aae00#diff-568ea47e7d1cda4a6191dbfc20376a45

## Details

First I'll note what does work. Defining a user model using the methodology spelled out in the 1D example in the following link works as expected.

http://cxc.harvard.edu/sherpa/threads/user_model/

I modified the example in the above link to produce the following code which works in version 4.7 and 4.9.0:

```
import sherpa.astro.ui as ui

def myline(pars, x):
    return pars[0] * x + pars[1]

x = np.arange(20)
y = x ** 2
ui.load_arrays(1, x, y)
ui.load_user_model(myline, "myl")
ui.add_user_pars("myl", ["m","b"])
ui.set_model(myl)
myl.m=30
myl.b=20
ui.fit()
ui.plot_fit()
```

My use case is a bit more complex, where I need to define my own methods for updating the model parameters and calculating the fit statistic, along with a custom model that queries the input data directly. Here is a simplified version that demonstrates my use case.

```
import sherpa.astro.ui as ui
import numpy as np

class ExampleModel(object):
    """ Class to define model
    """
    def __init__(self, x, y):
        self.x = np.array(x) # x needs to be an iterable
        self.y = np.array(y) # y needs to be an iterable
        self.parvals = [1, 2] # initial guess
        self.parnames = ("m", "b")

    def calc_stat(self):
        return np.sum(np.abs(self.y - self.model()))

    def model(self):
        return  self.parvals[0] * self.x + self.parvals[1]


class CalcModel(object):
    """ Class to update model parameters
    """
    def __init__(self, model):
        self.model = model

    def __call__(self, pars, x):
        self.model.parvals = pars
        return np.ones_like(x)


class CalcStat(object):
    """ Class to determine fit statistic
    """
    def __init__(self, model):
        self.model = model
        self.min_fit_stat = None
        self.min_par_vals = self.model.parvals

    def __call__(self, _data, _model, *args, **kwargs):
        fit_stat = self.model.calc_stat()

        if self.min_fit_stat is None or fit_stat < self.min_fit_stat:
            self.min_fit_stat = fit_stat
            self.min_parvals = self.model.parvals

        print('\n FIT STAT = {}\n'.format(fit_stat))

        return fit_stat, np.ones(1)


xdata = np.arange(20)
ydata = xdata ** 2
newmodel = ExampleModel(xdata, ydata)

dummy_data = np.zeros(1)
dummy_times = np.arange(1)
ui.load_arrays(1, dummy_times, dummy_data)

method = 'simplex'
ui.set_method(method)

ui.load_user_model(CalcModel(newmodel), 'simplemodel')
ui.add_user_pars('simplemodel', newmodel.parnames)
ui.set_model(1, 'simplemodel')

calc_stat = CalcStat(newmodel)
ui.load_user_stat('customstat', calc_stat, lambda x: np.ones_like(x))
ui.set_stat(customstat)

ui.fit(1)
```

This example works in version 4.7, however I get the error trace shown below when using version 4.9.0, but only after I run the `ui.fit(1)` command, I do not receive any errors prior to running this fit command. I did not test my example code in 4.8.2, however I did confirm I get a very similar error in that version when running my real-world case. 

```
<ipython-input-34-d42e83603d80> in <module>()
----> 1 ui.fit(1)

<string> in fit(id, *otherids, **kwargs)

/proj/sot/ska/dev/arch/x86_64h-darwin-10.12/lib/python3.5/site-packages/sherpa/astro/ui/utils.py in fit(self, id, *otherids, **kwargs)
   9260         """
   9261         kwargs['bkg_only'] = False
-> 9262         self._fit(id, *otherids, **kwargs)
   9263
   9264     def fit_bkg(self, id=None, *otherids, **kwargs):

/proj/sot/ska/dev/arch/x86_64h-darwin-10.12/lib/python3.5/site-packages/sherpa/astro/ui/utils.py in _fit(self, id, *otherids, **kwargs)
   9364                     i).mask & numpy.isfinite(self.get_data(i).get_x())
   9365
-> 9366         res = f.fit(**kwargs)
   9367         res.datasets = ids
   9368         self._fit_results = res

/proj/sot/ska/dev/arch/x86_64h-darwin-10.12/lib/python3.5/site-packages/sherpa/fit.py in run(fit, *args, **kwargs)
     57     def run(fit, *args, **kwargs):
     58         fit.model.startup()
---> 59         result = func(fit, *args, **kwargs)
     60         fit.model.teardown()
     61         return result

/proj/sot/ska/dev/arch/x86_64h-darwin-10.12/lib/python3.5/site-packages/sherpa/fit.py in fit(self, outfile, clobber)
   1099             raise FitErr('statnotforbackgsub', self.stat.name)
   1100
-> 1101         init_stat = self.calc_stat()
   1102         # output = self.method.fit ...
   1103         output = self._iterfit.fit(self._iterfit._get_callback(outfile, clobber),

/proj/sot/ska/dev/arch/x86_64h-darwin-10.12/lib/python3.5/site-packages/sherpa/fit.py in calc_stat(self)
   1014         #     shell.interact()
   1015
-> 1016         return self._calc_stat()[0]
   1017
   1018     def calc_chisqr(self):

TypeError: 'NoneType' object is not subscriptable
```

This is telling me that the `_calc_stat` function has a return value of `None`, suggesting my custom fit statistic method, as defined using the `CalcStat` class in my example, is missing or not set properly. This is the clue that led me to the theory I stated above in the summary, please let me know if my theory is not accurate or if I am missing an important step that would fix my problem.

Thank You,
Matthew Dahmer

olaurino:
# Release Note
Fix #341. A number of regressions were introduced in version 4.8.1 up to version 4.9.0, so user statistics that were properly working in version 4.7 have not been working any more. This has been fixed, and a number of regression tests have been added.

# TODO:

  * [x] Refactor to avoid repetitions and improve maintainability
  * [ ] Should we improve the error message when the statistic does not return a float?
  * [ ] Should we add a way to get the user statistic from the API without relying on the pollution of the root namespace?
  * [ ] Should we add documentation on how to use classes, rather than functions, as user statistics?
  * [ ] Should we remove any mention of weights? Note that this would change the advertised API for user statistics. Not that I see this as an issue, since the functionality has been broken since v4.8.1 anyway
  * [ ] Should we pass the full data and model objects to the user statistic method? That is a significant change to the API, although a good one. We could make it backward compatible by providing a switch in `load_user_stat` for functions implementing the new API. Maybe out of scope for this PR, though, and not a priority.

# Notes
The tests in this PR make sure the [examples in the documentation](http://cxc.harvard.edu/sherpa4.4/statistics/#userstat) work again. Plus, it also probes for 

In issue #341 the reporter suggests some code that reproduces the issue. However, such code would not work with this PR either, because the user statistic does not return a float, which the underlying C code doesn't like. **Should we make sure that this case provides a more helpful error message? Should we do it in this PR?**

I could not find a way to get the user statistic from the API, which means you have to rely on a variable being present in the root namespace. This is a little more complex to fix, **but there is a question whether this should be possible**.

I think there is a potential confusion as to how the `UserStat` class should be used to create custom statistic *classes*, rather than *functions*, which probably lead to the regression. At first glance, the pattern might seem to be that `UserStat` is a kind of abstract class users can subclass and for which they have to implement the `calc_stat` method. That is not the case though. It looks more like a wrapper class that provides a statistic implementation to wrap callable objects and functions. In this sense the only way users can implement a statistic as a class rather than a function is by creating a callable class, which is what the OP did. Or, they should know how exactly to subclass `UserStat`. So, **I am not sure whether we are OK on the documentation side, or we should address the issue of showing how to create a user statistic class.**

**Do we really need to keep support for the weights?** They are not mentioned anywhere in the documentation, although they are included in the signature (as `None`) for user statistics. We have being punting this, but maybe it's time to remove such keyword?

### DIFF
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -734,7 +734,10 @@ class UserStat(Stat):
         return self.errfunc(data)
 
     def calc_stat(self, data, model):
-        raise StatErr('nostat', self.name, 'calc_stat()')
+        if not self._statfuncset:
+            raise StatErr('nostat', self.name, 'calc_stat()')
+        else:
+            return self.statfunc(data, model)
 
 
 class WStat(Likelihood):
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2009, 2015, 2016  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2009, 2015, 2016, 2017  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -192,6 +192,13 @@ class Stat(NoNewAttributesAfterInit):
         self._check_sizes_match(data, model)
         return data, model
 
+    def _get_fit_model_data(self, data, model):
+        data, model = self._validate_inputs(data, model)
+        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
+        modeldata = data.eval_model_to_fit(model)
+
+        return fitdata, modeldata
+
     # TODO:
     #  - should this accept sherpa.data.Data input instead of
     #    "raw" data (i.e. to match calc_stat)
@@ -279,9 +286,7 @@ class Likelihood(Stat):
         return data, model
 
     def calc_stat(self, data, model):
-        data, model = self._validate_inputs(data, model)
-        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
-        modeldata = data.eval_model_to_fit(model)
+        fitdata, modeldata = self._get_fit_model_data(data, model)
 
         return self._calc(fitdata[0], modeldata, None,
                           truncation_value)
@@ -493,11 +498,7 @@ class Chi2(Stat):
         raise StatErr('chi2noerr')
 
     def calc_stat(self, data, model):
-
-        # TODO: HOW TO GET THE WEIGHTS?
-        data, model = self._validate_inputs(data, model)
-        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
-        modeldata = data.eval_model_to_fit(model)
+        fitdata, modeldata = self._get_fit_model_data(data, model)
 
         return self._calc(fitdata[0], modeldata,
                           fitdata[1], fitdata[2],
@@ -735,17 +736,16 @@ class UserStat(Stat):
         return self.errfunc(data)
 
     def calc_stat(self, data, model):
-        data, model = self._validate_inputs(data, model)
-        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
-        modeldata = data.eval_model_to_fit(model)
         if not self._statfuncset:
             raise StatErr('nostat', self.name, 'calc_stat()')
-        else:
-            return self.statfunc(fitdata[0],
-                                 modeldata,
-                                 staterror=fitdata[1],
-                                 syserror=fitdata[2],
-                                 weight=None)  # TODO weights
+
+        fitdata, modeldata = self._get_fit_model_data(data, model)
+
+        return self.statfunc(fitdata[0],
+                             modeldata,
+                             staterror=fitdata[1],
+                             syserror=fitdata[2],
+                             weight=None)  # TODO weights
 
 
 class WStat(Likelihood):
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -741,7 +741,11 @@ class UserStat(Stat):
         if not self._statfuncset:
             raise StatErr('nostat', self.name, 'calc_stat()')
         else:
-            return self.statfunc(fitdata[0], modeldata)
+            return self.statfunc(fitdata[0],
+                                 modeldata,
+                                 staterror=fitdata[1],
+                                 syserror=fitdata[2],
+                                 weight=None)  # TODO weights
 
 
 class WStat(Likelihood):
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -695,6 +695,7 @@ class UserStat(Stat):
 
         if statfunc is not None:
             self.statfunc = statfunc
+            self._calc = statfunc
             self._statfuncset = True
 
         if errfunc is not None:
@@ -734,10 +735,13 @@ class UserStat(Stat):
         return self.errfunc(data)
 
     def calc_stat(self, data, model):
+        data, model = self._validate_inputs(data, model)
+        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
+        modeldata = data.eval_model_to_fit(model)
         if not self._statfuncset:
             raise StatErr('nostat', self.name, 'calc_stat()')
         else:
-            return self.statfunc(data, model)
+            return self.statfunc(fitdata[0], modeldata)
 
 
 class WStat(Likelihood):
--- a/sherpa/stats/tests/test_user_stat.py
+++ b/sherpa/stats/tests/test_user_stat.py
@@ -20,6 +20,8 @@
 import numpy as np
 import pytest
 
+from pytest import approx
+
 from sherpa.astro import ui
 from sherpa.data import Data1D
 from sherpa.utils.err import StatErr
@@ -43,12 +45,12 @@ def test_user_stat_unit():
     ui.set_model(1, 'polynom1d.p')
 
     ui.load_user_stat('customstat', calc_stat, lambda x: given_stat_error)
-    ui.set_stat(customstat)
+    ui.set_stat(eval('customstat'))
 
     try:
         ui.fit(1)
     except StatErr:
-        pytest.fail("Call should not be throwing and exception (bug #341)")
+        pytest.fail("Call should not be throwing any exception (bug #341)")
 
     # Test the result is what we made the user stat return
     assert 3.235 == ui.get_fit_results().statval
@@ -91,15 +93,15 @@ def test_user_model_stat_docs():
     ui.clean()
     ui.load_arrays(1, x, y)
     ui.load_user_stat("mystat", my_stat_func, my_staterr_func)
-    ui.set_stat(mystat)
+    ui.set_stat(eval('mystat'))
     ui.load_user_model(myline, "myl")
     ui.add_user_pars("myl", ["m", "b"])
-    ui.set_model(myl)
+    ui.set_model(eval('myl'))
 
     ui.fit()
 
-    assert abs(1 - ui.get_par("myl.m").val) < 0.01
-    assert abs(3 - ui.get_par("myl.b").val) < 0.01
+    assert ui.get_par("myl.m").val == approx(1, abs=0.01)
+    assert ui.get_par("myl.b").val == approx(3, abs=0.01)
 
 
 def test_341():
@@ -163,8 +165,9 @@ def test_341():
 
     calc_stat = CalcStat(newmodel)
     ui.load_user_stat('customstat', calc_stat, lambda x: np.ones_like(x))
-    ui.set_stat(customstat)
+    ui.set_stat(eval('customstat'))
 
     ui.fit(1)
-    assert abs(1 - ui.get_par("simplemodel.m").val) < 0.00001
-    assert abs(3 - ui.get_par("simplemodel.b").val) < 0.00001
+
+    assert ui.get_par("simplemodel.m").val == approx(1, abs=0.00001)
+    assert ui.get_par("simplemodel.b").val == approx(3, abs=0.00001)
--- a/sherpa/stats/tests/test_user_stat.py
+++ b/sherpa/stats/tests/test_user_stat.py
@@ -1,0 +1,46 @@
+#
+#  Copyright (C) 2017  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+import numpy as np
+import pytest
+
+from sherpa.astro import ui
+from sherpa.utils.err import StatErr
+
+
+def test_341():
+    def calc_stat(data, _model):
+        return 3.235, np.ones_like(data.get_y())
+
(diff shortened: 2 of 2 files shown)


DougBurke:
It's going to take me a bit of time to get around to looking at this.

olaurino:
It looks like I forgot to mention in the description that I was going to add some more tests for user statistics. While implementing a very simple one, I realized there are still issues with the implementation. I'll look more into it tomorrow.

olaurino:
@matthewdahmer hopefully the new commits fix your issues.

@DougBurke I would appreciate some input on the questions in the TODO section above.

DougBurke:
Sorry - a bit busy at the moment. Will try to get time to look at this asap.

matthewdahmer:
@olaurino Repost from the original issue page: All my tests worked perfectly, PR #343 fixes the issue I was having, thank you.

DougBurke:
@olaurino - for your question "**Do we really need to keep support for the weights?**", I think it is a good move to drop support. I removed some unused code for weights in #287. While weights may be a useful addition, I don't think having half-implemented, completely unused, untested, and undocumented code paths to support weights is a good idea.

To be precise, I would suggest dropping weights support in this PR for the code that you are touching. I do not think it worth doing a global sweep to remove all remaining weights-based code in *this* PR.

sherpa/stats/__init__.py
```diff
@@ -734,7 +735,17 @@ def calc_staterror(self, data):
         return self.errfunc(data)
 
     def calc_stat(self, data, model):
-        raise StatErr('nostat', self.name, 'calc_stat()')
+        data, model = self._validate_inputs(data, model)
+        fitdata = data.to_fit(staterrfunc=self.calc_staterror)
+        modeldata = data.eval_model_to_fit(model)
+        if not self._statfuncset:
```
DougBurke: Wouldn't it be better to have this check *before* validating the input, extracting the data, and evaluating the model - e.g.

    def calc_stat(...):
        if not self._statfuncset:
            raise OhNoes()
        data, model = self._validate_inputs(...)
        ....

sherpa/stats/tests/test_user_stat.py
```diff
@@ -0,0 +1,170 @@
...
+
+    ui.set_model(1, 'polynom1d.p')
+
+    ui.load_user_stat('customstat', calc_stat, lambda x: given_stat_error)
+    ui.set_stat(customstat)
+
+    try:
+        ui.fit(1)
+    except StatErr:
+        pytest.fail("Call should not be throwing and exception (bug #341)")
```
DougBurke: Is this a typo in the string (should it read "any" rather than "and")?

sherpa/stats/tests/test_user_stat.py
```diff
@@ -0,0 +1,170 @@
...
+    ui.load_arrays(1, x, y)
+    ui.load_user_stat("mystat", my_stat_func, my_staterr_func)
+    ui.set_stat(mystat)
+    ui.load_user_model(myline, "myl")
+    ui.add_user_pars("myl", ["m", "b"])
+    ui.set_model(myl)
+
+    ui.fit()
+
+    assert abs(1 - ui.get_par("myl.m").val) < 0.01
```
DougBurke: Is it worth changing these to (untested)

    np.testing.assert_allclose(ui.get_par("my1.m").val, 1, rtol=0, atol=0.01)

since this gives us a slightly-more-informative error message on failure?

sherpa/stats/tests/test_user_stat.py
```diff
@@ -0,0 +1,170 @@
...
+    ui.load_user_model(CalcModel(newmodel), 'simplemodel')
+    ui.add_user_pars('simplemodel', newmodel.parnames)
+    ui.set_model(1, 'simplemodel')
+
+    calc_stat = CalcStat(newmodel)
+    ui.load_user_stat('customstat', calc_stat, lambda x: np.ones_like(x))
+    ui.set_stat(customstat)
+
+    ui.fit(1)
+    assert abs(1 - ui.get_par("simplemodel.m").val) < 0.00001
```
DougBurke: Same comment as above about using `np.testing.assert_allclose`

DougBurke:
I think this may be worth a different issue/PR, but I note that you can "see" and "set" the `userstat` statistic, which leads to an error. In the following I think that `ui.list_stats` shouldn't include "userstat", and we shouldn't be able to call `ui.set_stat("userstat")`.

```
In [1]: from sherpa import ui

In [2]: ui.load_arrays(1, [1,2,3], [2,4,5])

In [3]: ui.set_source(ui.const1d.mdl)

In [4]: mdl.c0 = 4.1

In [5]: ui.get_stat_name()
Out[5]: 'chi2gehrels'

In [6]: ui.calc_stat()
Out[6]: 0.6952047189401317

In [7]: ui.set_stat('userstat')

In [8]: ui.get_stat_name()
Out[8]: 'userstat'

In [9]: ui.calc_stat()
NotImplementedError                       Traceback (most recent call last)
<ipython-input-9-48a57ef953ff> in <module>()
----> 1 ui.calc_stat()

<string> in calc_stat(id, *otherids)

~/sherpa/sherpa-pr343/sherpa/ui/utils.py in calc_stat(self, id, *otherids)
   7876         """
   7877         ids, f = self._get_fit(id, otherids)
-> 7878         return f.calc_stat()
   7879 
   7880     def calc_chisqr(self, id=None, *otherids):

~/sherpa/sherpa-pr343/sherpa/fit.py in calc_stat(self)
    994         """
    995 
--> 996         return self._calc_stat()[0]
    997 
    998     def calc_chisqr(self):

~/sherpa/sherpa-pr343/sherpa/fit.py in _calc_stat(self)
    975         # TODO: is there anything missing here that
    976         #       self._iterfit.get_extra_args calculates?
--> 977         return self.stat.calc_stat(self.data, self.model)
    978 
    979     def calc_stat(self):

~/sherpa/sherpa-pr343/sherpa/stats/__init__.py in calc_stat(self, data, model)
    736 
    737     def calc_stat(self, data, model):
--> 738         data, model = self._validate_inputs(data, model)
    739         fitdata = data.to_fit(staterrfunc=self.calc_staterror)
    740         modeldata = data.eval_model_to_fit(model)

~/sherpa/sherpa-pr343/sherpa/stats/__init__.py in _validate_inputs(self, data, model)
    186             # StatErr('nostat', self.name, '_calc')
    187             #
--> 188             raise NotImplementedError("_calc method has not been set")
    189 
    190         data, model = self._bundle_inputs(data, model)

NotImplementedError: _calc method has not been set

In [10]: quit()
```

DougBurke:
For my suggestion of having `UserStat` essentially be only for "internal" use by `load_user_stat` to really work, I think we would also need to add an equivalent to `sherpa.ui.utils.add_model` but for user-written sub-classes of `sherpa.stats.Stat`. That is, a function that would register a user statistic model class with the session API.

sherpa/stats/tests/test_user_stat.py
DougBurke: Or perhaps cleaner to use the pytest features for checking numeric closeness - https://docs.pytest.org/en/latest/builtin.html#comparing-floating-point-numbers ?

DougBurke: Hmm, here's sample output from tests using a manual check (as in the current code) and using pytest.approx. I'm not sure one is necessarily better than the other, but I think using `approx` (or one of the other variants) is a bit more explicit about what the test is doing.

# manual

```
    def test_for_omar1():
        import numpy as np
        a = 0.1
        b = 0.2
>       assert (np.abs(a - b) < 0.05)
E       AssertionError: assert 0.10000000000000001 < 0.05
E        +  where 0.10000000000000001 = <ufunc 'absolute'>((0.1 - 0.2))
E        +    where <ufunc 'absolute'> = <module 'numpy' from '/data/dburke2/local/Linux/anaconda/envs/sherpa-warnings/lib/python3.6/site-packages/numpy/__init__.py'>.abs
```

# pytest.approx

```
    def test_for_omar2():
        import pytest
        a = 0.1
        b = 0.2
>       assert a == pytest.approx(b, abs=0.05)
E       AssertionError: assert 0.1 == 0.2 ± 5.0e-02
E        +  where 0.2 ± 5.0e-02 = <class '_pytest.python.approx'>(0.2, abs=0.05)
E        +    where <class '_pytest.python.approx'> = <module 'pytest' from '/lagado.real/sherpa/sherpa-warnings/.eggs/pytest-3.1.1-py3.6.egg/pytest.py'>.approx
```

sherpa/stats/__init__.py
olaurino: Yes, that's part of the refactoring I was mentioning, but thanks... it's been a while since I looked at this, there is no guarantee I would have remembered.

sherpa/stats/tests/test_user_stat.py
olaurino: Yes, fixed.

sherpa/stats/tests/test_user_stat.py
olaurino: Agreed, `approx` looks like the best approach in this case.

olaurino:
Rebased to get latest updates, original tip was 0696329

olaurino:
@DougBurke There a number of comments in this conversation, I would like you to confirm I addressed all of the ones you requested for this review. If there is something else I should add please let me know.

olaurino:
By the way, I'll remove the latest commit (already covered in #371) and rebase this PR so it can take advantage of it.

DougBurke:
I'm dismissing my review as it is not meant to stop the approval of this PR, more to discsuss where to go from here.

DougBurke:
@olaurino we should probably create an issue for your remaining "todo" items/questions, so we don't forget them.

olaurino:
That sounds good to me, although maybe we should agree on the todo list offline?

PR #351, opened 2017-04-10
fix handling of AREASCAL column in PHA files (fix #350)

### LINKED ISSUE #350: problems with PHA files with an AREASCAL column (or keyword value not equal to 1)
We've had a CXC helpdesk ticket about problems when a PHA file has an AREASCAL column. It happens because the `DataPHA` class doesn't handle this case properly: it's handled by `get_y` when it probably should be `get_dep`, there are issues with how the errors are scaled, and there's problems with background handling.

An example script that sets up a problem case (doesn't show all problems):

```
import numpy as np

from sherpa.models.basic import Const1D, StepHi1D

channels = np.arange(10)

# The idea is that there's two constant regions, with values
# of 20 and 40: i.e.
#
#    Const1D + StepHi1D
#
# where const1d.c0 = 20, stephi1d.xcut = 5.5,stephi1d.ampl = 20
#
# However, as areascal is 0.5 for the last 4 bins, the counts
# are all close to 20
#
# counts = np.asarray([0, 12, 21, 25, 18, 24, 47, 32, 39, 38])
#
counts = np.asarray([0, 12, 21, 25, 18, 24, 23, 16, 20, 19])

quality = np.zeros(10, dtype=np.int16)
quality[0] = 1
areascal = np.ones(10)
areascal[6:] = 0.5



dset = DataPHA('test', channel=channels, counts=counts,
               quality=quality, areascal=areascal)

dset.ignore_bad()

m1 = Const1D()
m2 = StepHi1D()
m1.c0 = 20
m2.xcut = 5.5
m2.ampl = 20

mdl = m1 + m2

from sherpa.astro.plot import DataPlot, ModelPlot

dplot = DataPlot()
mplot = ModelPlot()

dplot.prepare(dset)
mplot.prepare(dset, mdl)

from sherpa.stats import Chi2XspecVar

cvar = Chi2XspecVar()

print("stat val = {}".format(cvar.calc_stat(dset, mdl)))
```

The statistic value should be 8.713777196614727 (first element of the tuple) but with the current master branch we get 99.04558497693509.  There are problems with data and model plots - although the latter is hard to show with this example.

I have some preliminary changes which address some of these problems, but it's not yet ready for review.

DougBurke:
# Summary

Add support for handling the AREASCAL value (either scalar of vector) for PHA data sets. This array is used in XMM RGS data to handle missing chips.

# Details

The AREASCAL value is provided as part of a PHA data set, either as a keyword (scalar) or column (vector). It can be provided with both source and background data sets. The OGIP PHA standard which describes the value https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/node7.html
says

> AreaSc, a 4-byte REAL scalar giving the area scaling for this row (channel). In most cases AREASCAL = 1.0, since the instrumental effective area is provided within the spectral calibration files (specifically by the SPECRESP column of the ARF in the general case - see George et al. 1992a). Internally, XSPEC divides the observed spectrum by AreaSc before comparing it to the model spectrum.

However, this doesn't really make much sense for use with Poisson-based statistics, where you don't want to be changing the data values. Conversations with Keith Arnaud have made it clear that XSPEC does treat it as "part of the exposure time", which makes sense for Poisson-based statistics. This suggests it could be part of the ARF or RMF (as I don't want to change the exposure time from a scalar to a vector), but it is complicated since the data (the AREASCAL values) is associated with a PHA file and not a response file. Fortunately we have the PHA variants of the ARF/RMF/RSP instrument models, which are easy to modify to handle the multiplication by the AREASCAL value.

The DataPHA class needs to be changed:

a) to support handling the AREASCAL when calculating the contribution of the background (when doing background subtraction) and the background error (e.g. `get_dep` and `get_staterror`)

b) to apply the area scaling to the data points when creating plots (e.g. `get_y` and `get_yerr`).

The final change is to pass the AREASCAL values to the WSTAT calculation: this is done by changing the exposure times from scalars to arrays (multiplying the exposure time by the area scaling factor for each channel). This is the only change that requires recompilation.

Tests have been added, including comparison with statistic and data values taken from XSPEC.

There may still be differences to XSPEC when the PHA file contains grouping information (since Sherpa and XSPEC do not handle the grouping of scaling factors the same way), and with the `chi2xspecvar` statistic when 0-count channels are found and background subtraction is turned on (since Sherpa and XSPEC treat these differently; see #356). Neither of these are dircetly related to the code here.

### DIFF
--- a/sherpa/astro/tests/test_astro_pha_scale.py
+++ b/sherpa/astro/tests/test_astro_pha_scale.py
@@ -181,6 +181,52 @@ def expected_basic_chisquare_errors_bgnd():
     return np.sqrt(expected)
 
 
+def expected_basic_chisquare_errors_scaling_bgnd():
+    """Return the expected error values (chi square) after bg subtraction.
+
+    The errors are calculated using the Chi2DataVar algorithm.
+
+    It assumes:
+       source exposure = 20
+       background exposure = 15
+
+       source backscal = 0.01
+       background backscal = 0.1
+
+    ignore_bad() is assumed to have been called.
+    """
+
+    # Calculate the errors based on the counts. There are no
+    # counts less than 1, so this is just the square root of
+    # the observed data value, which then has to be scaled by
+    # the exposure, backscal, and area scaling.
+    #
+    # Since ignore_bad has been called we ignore the first bin.
+    counts = expected_basic_counts(scale=False)[1:]
+
+    exposure_src = 20.0
+    exposure_bg = 15.0
+
+    backscal_src = 0.01
+    backscal_bg = 0.1
+
+    ascal = expected_basic_areascal()[1:]
+
+    bgcounts = expected_basic_counts_bgnd(scale=False)[1:]
+    bgascal = expected_basic_areascal_bgnd()[1:]
+
+    denom = ascal
+    expected = counts / (denom * denom)
+
+    # The background counts are re-scaled to match the source
+    # settings.
+    denom = bgascal * backscal_bg * exposure_bg / \
+        (backscal_src * exposure_src)
+
+    expected += bgcounts / (denom * denom)
+    return np.sqrt(expected)
+
+
 def setup_basic_dataset():
     """Create a basic PHA data set with an AREASCAL value.
 
@@ -411,7 +457,10 @@ def test_get_staterror_no_bgnd():
 
 
 def test_get_staterror_bgnd():
-    """What does get_staterror return when bgnd is subtracted."""
+    """What does get_staterror return when bgnd is subtracted.
+
+    This is the easy case, since BACKSCAL and EXPOSURE are 1.
+    """
 
     dset = setup_basic_dataset_bgnd()
     dset.ignore_bad()
@@ -422,4 +471,30 @@ def test_get_staterror_bgnd():
                                 staterrfunc=stat.calc_staterror)
 
     expected = expected_basic_chisquare_errors_bgnd()
+    assert_allclose(errors, expected)
+
+
+def test_get_staterror_scaling_bgnd():
+    """What does get_staterror return when bgnd is subtracted.
+
+    This expands on test_get_staterror_bgnd by having different
+    BACKSCAL and EXPOSURE values in the source and background
+    data sets.
+    """
+
+    dset = setup_basic_dataset_bgnd()
+    dset.exposure = 20.0
+    dset.backscal = 0.01
+    dset.ignore_bad()
+    dset.subtract()
+
+    bset = dset.get_background()
+    bset.exposure = 15.0
+    bset.backscal = 0.1
+
+    stat = Chi2DataVar()
+    errors = dset.get_staterror(filter=True,
+                                staterrfunc=stat.calc_staterror)
+
+    expected = expected_basic_chisquare_errors_scaling_bgnd()
     assert_allclose(errors, expected)
--- a/sherpa/astro/tests/test_astro_pha_scale.py
+++ b/sherpa/astro/tests/test_astro_pha_scale.py
@@ -1017,8 +1017,9 @@ def test_cstat_comparison_xspec(make_data_path):
 
     ui.clean()
     ui.set_data(dset)
-    ui.set_source(ui.xspowerlaw.pl)
-    ui.set_par('pl.norm', 1e-4)
+    # use powlaw1d rather than xspowerlaw so do not need XSPEC
+    ui.set_source(ui.powlaw1d.pl)
+    ui.set_par('pl.ampl', 1e-4)
 
     ui.set_stat('cstat')
     ui.set_analysis('channel')
@@ -1063,8 +1064,8 @@ def test_wstat_comparison_xspec(make_data_path):
 
     ui.clean()
     ui.set_data(dset)
-    ui.set_source(ui.xspowerlaw.pl)
-    ui.set_par('pl.norm', 1e-4)
+    ui.set_source(ui.powlaw1d.pl)
+    ui.set_par('pl.ampl', 1e-4)
 
     ui.set_stat('wstat')
     ui.set_analysis('channel')
--- a/sherpa/astro/tests/test_astro_pha_scale.py
+++ b/sherpa/astro/tests/test_astro_pha_scale.py
@@ -36,7 +36,7 @@ data and/or model values are shown).
 
 """
 
-# import pytest
+import pytest
 import six
 import warnings
 
@@ -55,6 +55,13 @@ def expected_basic_areascal():
     areascal[0] = 0.0
     areascal[6:] = 0.5
     return areascal.copy()
+
+
+def expected_basic_areascal_bgnd():
+    """Return the expected areascal values for the background."""
+
+    return np.asarray([0.0, 2.0, 2.0, 2.0, 2.0,
+                       4.0, 4.0, 4.0, 4.0, 4.0]).copy()
 
 
 def expected_basic_counts(scale=False):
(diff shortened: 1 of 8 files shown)


DougBurke:
With regard to the test failures: I wonder if some test - or piece of code in Sherpa - has changed the NumPy warnings and hasn't reset them: i.e. it called `np.seterr` but does not restore things once it has finished.

anetasie:
The definitions of AREASCAL and BACKSCAL are on page 11 of the OGIP/92-007 memo updated on May 2009 [(https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007.ps)].
This could probably be included somewhere in the code:

> 7.0 AreaSc, a 4-byte REAL scalar giving the area scaling for this row (channel). In most cases AREASCAL = 1.0, since the instrumental effective area is provided within the spectral calibration files (specifically by the SPECRESP column of the ARF in the general case – see George et al. 1992a). Internally, XSPEC divides the observed spectrum by AreaSc before comparing it to the model spectrum.

In general we may advise users to work with the ARF file instead of scaling the data and then think about the systematic errors this introduces to the entire process.

I looked at the updated code and it looks good. I'm still wondering about the calculation of the staterrors. Because the A is applied to the observed data  this should be the standard error propagation for the data before doing any background subtraction. The standard error for N counts is sqrt(N), and for N/A counts would be sqrt(N)/A. However, with the grouping the AREASCAL needs to be applied to the data in each channel in the group and the error calculated for each channel before combining the errors in quadrature for the entire group. I'm not sure how this will behave in the cases of the weighted errors, so need to go through the examples. But, this introduces systematic errors that we ignored and leave to the users to worry about. Modifying the observed data is a really bad idea.

anetasie:
For the Poisson Likelihood calculations - the AREASCAL would just rescale the data. Not a perfect solution as this would introduce systematics that are not taken into account in the final analysis.

DougBurke:
@anetasie I'm out today.

Sherpa appears to have several approaches to dealing with calculating values in a group. The standard ones are sum and "take the average of the min and max values [I think, going from memory]". I believe XSPEC often uses counts (or perhaps flux) weighting in this case.

I think we have  a bunch of things to talk about here, as I am having trouble getting the same answer as XSPEC when source and/or background channels have 0 counts in them (this is before worrying about AREASCAL).

DougBurke:
We (well @anetasie) have had a discussion with @xspector about how AREASCAL is handled in XSPEC, in particular for likelihood-based statistics like CSTAT (where scaling the counts array is the wrong thing to do), and indeed XSPEC applies the AREASCAL to the exposure time. I've confirmed that we currently get different results to XSPEC for the CSTAT statistic when AREASCAL is not unity (as expected).

I think that the best way to support this is to follow XSPECs lead, and apply AREASCAL to a factor other than the counts (as is currently done in the master branch and "fixed" in this PR for the chi-square case). However, I am not sure there's a good place to do this with the current API (which includes model expressions). Thought is needed here!

DougBurke:
I am moving towards applying the AREASCAL factor to the exposure time (in which case it multiplies the value rather than used as a divisor as it is with the counts), since this appears to be the more-sensible approach for likelihood statistics (where you should not be manipulating the count values). The value is probably going to be applied in the `sherpa.astro.instrument.*PHA` classes, since there the instrument response can access the AREASCAL values. This means that the correction applied in the `sherpa.astro.data.DataPHA.get_dep` routine is no-longer needed *except* to handle background subtraction, but is still needed in the `get_y/get_dy` routines for plotting purposes.

Actually, it isn't really needed for plotting, but it makes for a "nicer" display, where you don't see huge vertical shifts when the AREASCAL values change from 0.5 to 1 (for the RGS case).

DougBurke:
I have been able to get agreement with the ca99c97 commit when comparing XSPEC and Sherpa results for a RGS data using CSTAT (XSPEC) and WSTAT (Sherpa, since there was a background dataset), when no grouping or filtering was applied. So that's nice.

sherpa/astro/tests/test_astro_pha_scale.py
```diff
@@ -0,0 +1,1191 @@
...
+from sherpa.models.basic import Const1D, StepHi1D
+from sherpa.stats import Chi2DataVar, CStat
+from sherpa.utils.err import ArgumentErr, DataErr
+
+from sherpa.utils import requires_data, requires_fits
+
+from sherpa.astro import ui
+
+
+# pytest.mark.parametrize and requires_data do not seem to be playing
```
olaurino: This is ugly. I tested changing the `requires_*` decorator to use `pytest.mark.skipif`. It seems to work, however `pytest` is not guaranteed to be on the user's system, so `sherpa.utils` should have a trap to only define the decorators if `pytest` is importable. Even better, the decorators might be defined in their own module, a first step towards splitting `sherpa.utils`. Do we want to do it now?

sherpa/stats/__init__.py
```diff
@@ -849,11 +850,15 @@ def calc_stat(self, data, model):
 
             bset = dset.get_background(bid)
 
+            # TODO: the following should be reviewed to see what
```
DougBurke: Tests could be added, but I don't have time to do so at the moment.

sherpa/astro/tests/test_astro_pha_scale.py
DougBurke: I have limited time to make any significant changes.

I'm not 100% convinced whether the diagnosis of the problem I made is correct (it seems to be, but allI did was fins what appeared to be a workable workaround, rather than to diagnose the problem(.

olaurino: I would like to fix this. the main issue however is that I need to import pytest in sherpa.utils, but pytest might not be available. What to do then? If pytest is not available then the tests cannot be run in the first place. So, I'll try importing pytest and if the import fails I'll make the decorator throw and exception instead, prompting the user to install pytest. Then the question is if we should add some testing, and the snowball keeps getting larger (especially because one would need to simulate the absence of pytest itself, with unknown consequences on the tests being run). I could settle for some form of manual testing, but I'll try to add some testing as well, and if it turns out to be less than trivial I'll give up and test manually.

DougBurke: While I agree that it is ugly, I don't personally see it as something that should stop this PR. Can it not be a separate issue?

olaurino: I have a commit I'll push to my fork to have Travis run on it. It was easier than I expected to write the unit tests. Assuming Travis is happy with it I'll let you decide whether to include the commit in this PR or in a new one. I would consider this a normal refactoring required by issues encountered when fixing the original bug. OTOH it is a somewhat extensive refactoring. As long as we refactor **soon** I am OK either way.

olaurino: See https://github.com/olaurino/sherpa/commit/ba29928eaab81ff4617cb2b53c99033aad5ccee6
I think we also need to rebase it on top of the current master, I only see 5 travis jobs, rather than 7.

olaurino: The problem now is that the old unittest-style tests (which were mostly skipped in my local tests, apparently) are not skipped anymore, as the `pytest.skipif` does not seem to have effect. I'll keep looking: even if we had to have two different sets of decorators, I'd rather have them cleaned up properly.

olaurino:
Rebasing, original tip was 6733229.

sherpa/astro/tests/test_astro_pha_scale.py
olaurino: I think it's all good now. I also rebased the branch to make sure it has the latest changes to the tests from master, including above all the ds9 test changes from #368. Again, up to you if you want me to issue a new PR or add a commit to this one. The new commit is https://github.com/olaurino/sherpa/commit/bf9980bca0c2e4a6191884305135f29876c14e36

DougBurke: I've left a couple of minor comments on your commit. I would rather see it as a separate PR as I think it's just cleaner to review and understand.

olaurino: I'll merge this PR and issue a new one. Thanks.

PR #352, opened 2017-04-10
Add low-level support for the XSPEC rgsxsrc convolution model

DougBurke:
# Summary

Add low-level support for the rgsxsrc convolution model in XSPEC. This is intended for R&D into
fully supporting XSPEC convolution models in Sherpa.

# Details

This model was part of XSPEC version 11 but removed in initial versions
of XSPEC 12, being added back in XSPEC 12.8.1. Somehow I missed this
so there was no support for it in the XSPEC module. I have also
added commented-out support for the convolution models added in XSPEC
12.9.1, since we do not have a way of tailoring the XSPEC module
to the XSPEC version it is being compiled againts.

Adding support for rgsxsrc requires adding new template code, since
it is the only convolution model that uses the F77 XSPEC model
interface (the others use the C interface). I did the minimum job
here (using copy and paste), rather than refactoring the templates
as there is now a lot of repeated logic.

As the convoultion models are only provided at a very-low level, there
are now tests of them, so none is added here for rgsxsrc. Note that
the idea of this interface is to allow us to experiment with what
is needed in Sherpa to properly support XSPEC convolution-style
models.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -278,7 +278,7 @@ void xsmtbl(float* ear, int ne, float* param, const char* filenm, int ifl,
 // versions.
 //
 void C_cflux(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
-  // void C_lumin(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
+  // void C_clumin(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);  12.9.1
 void C_cpflux(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 void C_xsgsmt(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 void C_ireflct(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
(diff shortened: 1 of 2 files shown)


olaurino:
Is this something we want to add to CIAO 4.10? I assumed it was before reviewing it, but now I understand there is a lot of work to do yet.

DougBurke:
Now we have moved to XSPEC 12.9.1 I think there are several more models that could be added to this PR, but it is a change I'd like to get into 4.10.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -1293,10 +1308,11 @@ static PyMethodDef XSpecMethods[] = {
   XSPECMODELFCT_CON(C_PartialCovering, 1),
   XSPECMODELFCT_CON(C_rdblur, 4),
   XSPECMODELFCT_CON(C_reflct, 5),
+  XSPECMODELFCT_CON_F77(rgsxsrc, 1),
   XSPECMODELFCT_CON(C_simpl, 3),
   XSPECMODELFCT_CON(C_zashift, 1),
   XSPECMODELFCT_CON(C_zmshift, 1),
-  
+
```
DougBurke: Hmmm, why did I not add commented-out lines saying

    XSPECMODELFCT_CON(C_clumin, 4),

for the commented-out convolution models? (note that in the comments `C_lumin` is wrong and should be `C_clumin`).

olaurino:
Should I wait for this PR to get more models or should we work to merge it as is? One concern I have is that I have no way to test this PR, not even manually, and I don't know enough about these models for my review to be meaningful. I know it doesn't break any existing Travis build. Maybe that's enough in this case?

On the other hand, is it possible these models belong in the external package that uses them? It might be a stupid question, that just proves I don't know enough about this models :)

DougBurke:
I should address the comment I made above - i.e. https://github.com/sherpa/sherpa/pull/352#discussion_r136488551 - but this is only a change to the comments, not to the functionality. I can also look at adding some sort of low-level test.

There's no "nice" way for an external module to build the necessary interface code (the compiled code that this adds), so at the moment I strongly believe we need it here.

Since we've decided for the short term to target XSPEC 12.9.0 then I would like this to get merged sooner rather than later. Once we've worked out the optional model support for the 12.9.1 models we could look at adding optional support for the other convolution models (those not in 12.9.0 but are in 12.9.1), but that would be a separate PR.

olaurino:
Any way I can manually test this or should I just trust you it works?

DougBurke:
Given that this adds a "new" interface class - that is the C++ wrapper code we write to interface between Python scope and the XSPEC library - it would be good to add a test. However, the only way to test this interface is to actually use a model that uses this interface, and at this time the `rgsxsrc` model is the only one that uses it. This model requires both an XMM RGS spectrum and at least one external file (I can't remember the details), so it is involved to write a test.

On the plus side, there are no downstream users of this functionality at present, so if it's broken we don't break anything!

I do need to go back and clean up the comments.

olaurino:
I understand we can't write a test, easily or at all. I was just wondering if I could test is manually. I guess you are adding this model for a reason, and that you tested it yourself somehow. In any case I'll add a request for changes so I don't look again at this PR in a week wondering what to do with it.

DougBurke:
About to rebase. Tip was ebd62d1

DougBurke:
@olaurino - I'll try to package you up some data and a simple test script for your amusement. It's data from a colleague, so I don't want to make it public.

DougBurke:
Commit f299252 is the change to the comment-ed out code.

If we wait for #395 to land then this could be updated to take advantage of the `XSPEC_12_9_1` define to allow conditional compilation of the 12.9.1 models.

DougBurke:
I have also sent @olaurino a test script to verify that the code is doing something not-completely terrible.

DougBurke:
Aha - I still have my comment from https://github.com/sherpa/sherpa/pull/352#discussion_r136488551 to look at.

DougBurke:
About to rebase. Tip was f299252

DougBurke:
I've finnished the tweaking I wanted to do. So we can either review/merge or wait for #395 and then optionally include the 12.9.1 models.

olaurino:
rebasing to resolve conflicts, was 5f22c8e

olaurino:
I know what's going on... while resolving conflicts I missed one of them, which resulted in a broken c++ file.

PR #358, opened 2017-04-27
Improve testing of non-Chandra OGIP files (fix #357)

### LINKED ISSUE #357: pyfits backend fails to read in a SWIFT RMF
From https://github.com/sherpa/sherpa/issues/354#issuecomment-295712027

If you use the pyfits/astropy IO backend, you can't load in the SWIFT RMF swxpc0to12s6_20130101v014.rmf - one location for this file is http://darts.jaxa.jp/pub/legacy.gsfc.nasa.gov/caldb/data/swift/xrt/cpf/rmf/swxpc0s6_20130101v014.rmf

With pycrates (in CIAO 4.9) I can load it in:

```
sherpa-1> rmf = unpack_rmf('swxpc0to12s6_20130101v014.rmf')
sherpa-2> print(rmf)
name     = swxpc0to12s6_20130101v014.rmf
detchans = 1024
energ_lo = Float64[2400]
energ_hi = Float64[2400]
n_grp    = UInt64[2400]
f_chan   = UInt32[2400]
n_chan   = UInt32[2400]
matrix   = Float64[2457600]
offset   = 0
e_min    = Float64[1024]
e_max    = Float64[1024]
```

but with astropy 1.3.2 installed

```
In [2]: rmf = sherpa.astro.ui.unpack_rmf('swxpc0to12s6_20130101v014.rmf')
IOErr                                     Traceback (most recent call last)
<ipython-input-2-8f4410910367> in <module>()
----> 1 rmf = sherpa.astro.ui.unpack_rmf('swxpc0to12s6_20130101v014.rmf')

<string> in unpack_rmf(arg)

/home/djburke/sherpa/sherpa-rgs/sherpa/astro/ui/utils.py in unpack_rmf(self, arg)
   5363 
   5364         """
-> 5365         return sherpa.astro.instrument.RMF1D(sherpa.astro.io.read_rmf(arg))
   5366 
   5367     # DOC-TODO: add an example of a grating/multiple response

/home/djburke/sherpa/sherpa-rgs/sherpa/astro/io/__init__.py in read_rmf(arg)
    181     read_rmf( RMFCrate )
    182     """
--> 183     data, filename = backend.get_rmf_data(arg)
    184     return DataRMF(filename, **data)
    185 

/home/djburke/sherpa/sherpa-rgs/sherpa/astro/io/pyfits_backend.py in get_rmf_data(arg, make_copy)
    747         data['matrix'] = None
    748         raise IOErr('badfile', filename,
--> 749                     " MATRIX column not a variable length field")
    750 
    751     good = (data['n_grp'] > 0)

IOErr: 'swxpc0to12s6_20130101v014.rmf' is not a filename or  MATRIX column not a variable length field
```

I haven't looked to see if this is due to a change in the backend code since this was written or, more likely, if it's just a bug in our code.

DougBurke:
# Summary

Fixes issue #357, which meant that a Swift RMF could not be read in when the AstroPy back end was in use. The problem was that the code did not support RMF matrices that were not stored as variable-length arrays. The PR adds tests of reading in Swift PHA, ARF, and RMF files, as well as checks that you can not read in a RMF as a PHA (and various other options), and a simple check that the files work with the `sherpa.astro.ui` API.

# Details

The test that all the files work together calculates the statistic for a model over a fixed energy range, comparing the values to those calculated manually using XSPEC. To do this the code changes the RMF and ARF, to replace the 0-energy lower-edge of the first bin by a small positive value.

I would like to remove pr:hold, but this needs sherpa/sherpa-test-data#11 to be accepted first. I also don't know why the travis tests haven't been run for any of these commits (perhaps because I didn't handle the submodule update correctly?).

### DIFF
--- a/sherpa/astro/tests/test_astro_data_swift_unit.py
+++ b/sherpa/astro/tests/test_astro_data_swift_unit.py
@@ -58,6 +58,8 @@ except ImportError:
     io = None
     backend = None
 
+from sherpa.astro import ui
+
 
 # PHA and ARF are stored uncompressed while RMF is gzip-compressed.
 # Use the same name (no .gz suffix) to access all files.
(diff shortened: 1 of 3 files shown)


DougBurke:
@olaurino in fd13e89 I have to add a warning to `sherpa/conftest.py`. However, I really only want this warning ignored in the one test; it *should not* be ignored if created anywhere else in the system. I don't think the current `capture_all_warnings` fixture in `sherpa/conftest.py` supports this.

Some options would be (off the top of my head):

 - to extend the metadata used by `capture_all_warnings` so that you can specify a test, or file, in which the warning should be ignored
 - to add a decorator to a test to say "please ignore the following warnings if raised by this test"

Probably many others. I currently prefer the second approach, but no idea if either are possible.

I have tried (in #351) to use the `recwarns` fixture to try and remove a raised warning from within a test (i.e. check it was raised, and then remove it from the global database), but this did not work reliably, so I ended up removing it.

DougBurke:
Just to note that we have external validation that this fixes a similar problem (this time not a SWIFT response); https://github.com/sherpa/sherpa/issues/357#issuecomment-306219412

sherpa/conftest.py
```diff
@@ -64,7 +64,14 @@ def pytest_addoption(parser):
...
         ],
     RuntimeWarning:
-        [r"invalid value encountered in sqrt", ],
+        [r"invalid value encountered in sqrt",
+         # The following is added for the test_can_use_swift_data
+         # test in sherpa/astro/tests/test_astro_data_swift_unit.py;
+         # as the test checks this warning is created it would be useful
+         # if the test could eitehr clear the warnings, or tell this
+         # fixture to ignore it, rather than making it a global flag
+         r"divide by zero encountered in divide",
```
olaurino: I might be doing something wrong, but if I remove these lines I don't get a failure from the tests. @DougBurke could you please confirm you saw this failure before adding the warnings to the whitelist?

olaurino: I should add I tried this on Python 3, I'll try with Python 2, maybe that's the catch.

olaurino: Nope, just tried in Python 2, the tests pass even though I removed the warnings from the white list.

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -0,0 +1,454 @@
...
+import numpy as np
+from numpy.testing import assert_allclose
+
+import pytest
+
+from sherpa.utils import requires_data, requires_fits
+from sherpa.utils.err import IOErr
+from sherpa.astro.data import DataARF, DataPHA, DataRMF
+
+# Should each test import io instead of this? Also, do we have a
```
olaurino: importing io in each test could be done with a fixture, but I am not sure what that would buy you.

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -0,0 +1,454 @@
...
+# better way of determining what the backend is?
```
olaurino: I don't think we do. We might provide a name to the backends, but that would not be much better than what you did here. The backend "interfaces" are somewhat underspecified, so I am not sure how much sense it makes to add a name.

DougBurke: The reasons for these comments are - in part - because there's the following comment in `sherpa/astro/io/__init__.py`

> This module should not be imported directly if there is no
FITS support present (i.e. neither the crates or astropy
back ends are in use).

so I'm always a bit nervous about how best to import and use it.

sherpa/conftest.py
olaurino: @DougBurke any comments on my findings? I'd try running CI with the warnings removed from the whitelist. It bothers me that there isn't anything in the pytest docs explaining why the warnings wouldn't bubble up after you capture them with pytest itself, but then again maybe I did something wrong and Travis wouldn't be happy. This is the only (minor) issue preventing me from merging this PR.

DougBurke: I managed to forget about this comment until this morning. I haven't looked at it yet (unlikely to get around to it until tonight now). I don't know if it's related that in pytest 3 they added support for reporting warnings, so I don't know if any of the things we are doing interact strangely with that code (I doubt it but just mention it). My experience with the warning code has been some fragility, in that some things don't seem reliable. This is an aside, and we should work out what's going on here before merging the code.

olaurino:
Rebasing to get testing fixes/improvements from master. Original tip was 32478f5

sherpa/conftest.py
olaurino: I can confirm that all the tests pass on Travis after removing the warnings from the whitelist.

DougBurke: I can remove these from the whitelist. I may have just added them "in case" (it's one of the reasons why I think a more fine-grained approach to marking tests that can/do raise a warning would be useful).

olaurino: I'll push the commit I already have.

DougBurke:
Is this okay to be marked as approved now, or are we waiting for any more code changes?

Issue #360, opened 2017-05-02
Sherpa tests with Python 3.6


cdeil:
Hi, I updated from Python 3.5 to 3.6. Everything works OK (all Gammapy tests pass). There's just a few things that I noticed during build and when running `sherpa_test` that I thought you might want to know about. Feel free to close this issue any time if those aren't relevant.

First I got this: https://gist.github.com/cdeil/94e3f758e5501db7fc014a402accbdcb
I guess `mock` could be mentioned here?
https://github.com/sherpa/sherpa#testing-sherpa

After installing mock I got this: https://gist.github.com/cdeil/0306b8e16954599bdab18574997733d8
I guess there's a few missing `requires_data` decorators in your tests?
Shouldn't this be caught by your CI?

During install this is printed in red. Is is normal?
```
zip_safe flag not set; analyzing archive contents...
sherpa.__pycache__.__init__.cpython-36: module references __file__
sherpa.__pycache__._version.cpython-36: module references __file__
sherpa.astro.datastack.tests.__pycache__.test_datastack.cpython-36: module references __file__
sherpa.astro.models.__pycache__._modelfcts.cpython-36: module references __file__
sherpa.astro.ui.__pycache__.serialize.cpython-36: module MAY be using inspect.getsource
sherpa.astro.ui.tests.__pycache__.test_astro_ui.cpython-36: module references __file__
sherpa.astro.utils.__pycache__._pileup.cpython-36: module references __file__
sherpa.astro.utils.__pycache__._region.cpython-36: module references __file__
sherpa.astro.utils.__pycache__._utils.cpython-36: module references __file__
sherpa.astro.utils.__pycache__._wcs.cpython-36: module references __file__
sherpa.astro.utils.__pycache__.smoke.cpython-36: module references __file__
sherpa.estmethods.__pycache__._est_funcs.cpython-36: module references __file__
sherpa.models.__pycache__._modelfcts.cpython-36: module references __file__
sherpa.optmethods.__pycache__._minim.cpython-36: module references __file__
sherpa.optmethods.__pycache__._minpack.cpython-36: module references __file__
sherpa.optmethods.__pycache__._saoopt.cpython-36: module references __file__
sherpa.optmethods.__pycache__._tstoptfct.cpython-36: module references __file__
sherpa.stats.__pycache__._statfcts.cpython-36: module references __file__
sherpa.tests.__pycache__.test_stack.cpython-36: module references __file__
[...]

PR #363, opened 2017-06-07
fix up tests that need the requires_data decorator

DougBurke:
# Summary

Fix up several tests that were missing the `requires_data` decorator, and adds an AstroPy-with-no-sherpa-test-data build to Travis-CI to check.

# Details

This addresses part of #360 (and maybe a little bit of #260). I think in this case adding an extra build to Travis is okay, but it could be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,12 @@ addons:
     packages: &default_apt
       - gfortran
 
-# Barebone build to check tests pass when most of the tests must be skipped.
-# We need to use TEST=none to remove the test-data submodule.
-# Barebone build with no test data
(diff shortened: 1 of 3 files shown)


olaurino:
For some reason it looks like this PR makes Travis skip one of the jobs (the "barebone" build which is usually job n.1). master has 6 job, and while this PR is supposed to add a job, the resulting number of jobs is still 6. I am looking into it. Also, this is the PR that conflicts with #364. I'll merge this one before I rebase/merge #364.

olaurino:
I added a commit that moves jobs around in the travis configuration file. I am not really sure what was wrong, but the new configuration seems to work and probably makes more sense anyway.

PR #364, opened 2017-06-07
Remove ResourceWarning warnings from image tests on Python 3.6

DougBurke:
# NOTE

This PR has been superceeded by #368 - which contains the DS9 warning clean up code from this PR but excludes the decorator and `--show-warnings` operator. I've added back pr:hold on this one as there's likely to be conflict, but have left this open for the short term.

--------------------------------------------------------------------------------------------------------------------------------

# Summary

Update the DS9 code so that external processes are cleaned up properly. This removes the `ResourceWarning` warnings raised when running the `sherpa/image/tests/test_image.py` test.

For developers:

- added an explicit decorator to skip certain warnings: this is currently unused
- added a `--show-warnings` option to the test suite to support diagnosing failures due to unexpected warnings being created by a test. This is only needed for developing Sherpa.

# Details

The current test setup supports skipping warnings, but it is a global setting - we ignore all occurrences of a warning. This approach lets us be explicit about which tests are expected to create which warnings. I have found some issues with the current approach to handling warnings - in particular when I wanted to write tests that check that a warning is generated in a piece of code - which is one of the drivers of trying a different (less global) approach.

# Possible updates

Some tests could be converted to use this decorator in this PR. Is it worth it? Should the decorator work be pulled into a separate PR?

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -38,11 +38,13 @@ except ImportError:
 
 
 TEST_DATA_OPTION = "--test-data"
-
+SHOW_WARNINGS_OPTION = "--show-warnings"
 
 def pytest_addoption(parser):
     parser.addoption("-D", TEST_DATA_OPTION, action="store",
                      help="Alternative location of test data files")
+    parser.addoption(SHOW_WARNINGS_OPTION, action="store_true",
+                     help="Display the warnings created in a test")
 
 
 # Whilelist of known warnings. One can associate different warning messages
@@ -97,8 +99,11 @@ if have_astropy:
     known_warnings.update(astropy_warnings)
 
 
+# Can this be replaced by the warning support added in pytest 3.1?
+# See https://docs.pytest.org/en/latest/warnings.html#warnings
+#
 @pytest.fixture(scope="function", autouse=True)
-def capture_all_warnings(request, recwarn):
+def capture_all_warnings(request, recwarn, pytestconfig):
     """
     This fixture will run automatically before and after every test function is executed.
     It uses pytest's infrastructure to get all recorded warnings and match them against the while list. If an
@@ -111,6 +116,7 @@ def capture_all_warnings(request, recwarn):
     ----------
     request standard injected service for pytest fixtures
     recwarn injected pytest service for accessing recorded warnings
+    pytestconfig injected service for accessing the configuration data
 
     """
     def known(warning):
@@ -125,7 +131,13 @@ def capture_all_warnings(request, recwarn):
         warnings = [w for w in recwarn.list
                     if type(w.message) not in known_warnings or not known(w)]
 
-        assert 0 == len(warnings)
+        nwarnings = len(warnings)
+        if nwarnings > 0 and pytest.config.getoption(SHOW_WARNINGS_OPTION):
+            print("*** Warnings created: {}".format(nwarnings))
+            for i, w in enumerate(warnings):
+                print("{}/{} {}".format(i + 1, nwarnings, w))
+
+        assert 0 == nwarnings
 
     request.addfinalizer(fin)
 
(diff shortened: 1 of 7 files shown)


DougBurke:
I have no idea why the test fails, as it is in a completely-different part of the system: sherpa/ui/tests/test_ui.py:test_get_draws.test_covar_as_argument

https://travis-ci.org/sherpa/sherpa/jobs/240496540#L768

Since this is testing the output of `get_draws` I wonder if it just happened on a "lucky" random seed that ended up causing the 10 draws to not have any acceptable jumps? I'm tempted to re-run this test, but I want to leave the log - linked to above - around for reference.

DougBurke:
So, the test failure above does appear to be sporadic, as the only change in 08a49b5 is to the Travis-configuration file.

DougBurke:
Here's a comment on these ResourceWarnings which I'm ignoring in this change - https://mail.python.org/pipermail/python-dev/2016-August/145830.html - which does make it sound like we do have some code changes (rather than skipping the warnings).

DougBurke:
So I've just commited a "fix" for the issue; it is easy for most cases (calling xpaget or xpaset, which are tools we run and wait for them to finish), but the ds9 case itself is more complex (since this is run more as a background process with no direct communication), and I had to perform an ugly hack to remove the `ResourceWarning` case.

So this has now morphed into a "fix the issue" rather than "allow per-test warnings" PR. I've left the warning code in even though it's not used.

DougBurke:
Adding Pr:Hold back as we need to get some PRs merged first (there's a conflict somewhere).

DougBurke:
I've also had some success writing a fixture to mark a test as "expect this warning", rather than a decorator. I'll try and post the code soon, as maybe it's a better approach?

olaurino:
Rebasing, original tip was at 7f131f9

DougBurke:
sorry - been in the L3 meeting so not 100% following this. I agree on separating out the decorator step as I think there's other ways to achieve this: as mentioned above I have also written a pytest fixture which can be used to enforce/skip a warning message, and maybe that's a better approach (or at least should compare and contrast).

sherpa/conftest.py
```diff
@@ -38,11 +38,13 @@
 
 
 TEST_DATA_OPTION = "--test-data"
-
+SHOW_WARNINGS_OPTION = "--show-warnings"
 
 def pytest_addoption(parser):
     parser.addoption("-D", TEST_DATA_OPTION, action="store",
                      help="Alternative location of test data files")
+    parser.addoption(SHOW_WARNINGS_OPTION, action="store_true",
```
olaurino: Ah, the problem is that I wasn't passing 'sherpa' as an argument. I am not sure if the 'test' command should be implemented differently, in any case the following command both work as expected:

````
python setup.py test -a "-kds9 sherpa"
python setup.py test -a "-kds9 --show-warnings sherpa"
````

File: sherpa/image/DS9.py [hunk 2]
```diff
@@ -454,11 +450,26 @@ def doOpen(self):
...
         )
 
+        # Trick to stop a ResourceWarning warning to be created when
+        # running sherpa/tests/test_image.py
+        #
+        # Adapted from https://hg.python.org/cpython/rev/72946937536e
+        # but I am completely mis-using it, so it is unclear how
+        # sensible a hack this is.
+        #
+        p.returncode = 0
```
olaurino: I believe this statement to be more appropriate inside the `if self.isOpen()` condition. It's still a slight hack, but at least it's more semantically appropriate to say that the return code is 0 when things are OK, and leave it undefined when they are not, e.g. when the timeout is reached and the window is still not open (if I interpret the code correctly).

I tested this in a commit I am going to push for simplicity. If @DougBurke disagrees I can always revert back to the original code.

DougBurke:
Rebased. Original tip was a614232

DougBurke:
Apparently I have managed to confuse things a bit. Not sure how. I am probably going to rewrite this PR as it's easier (for me) than trying to disentangle the decorator code anyway.

PR #366, opened 2017-07-05
add xvfb plugin to test requirements

olaurino:
This PR uses a pytest plugin to always run tests with an Xvfb server, so that GUI tests do not get displayed in the current window. While this fixes a major annoyance for developers running the integration tests, it also simplifies the travis configuration, where Xvfb was configured explicitly.


DougBurke:
Cool. It looks like `pytest-xvfb` defaults to `800x600x16` whereas we were using `1280x1024x16`. Is it worth setting the defaults to use 1280 by 1024, or - given that your commit using 800 by 600 worked - leave it for now?

DougBurke:
I would like to separate out our buil/install/test instructions from the README.md file into separate file (or files), leaving a generic "how to install from conda" set in the README.md. Then we can (hipefully) better include requirements for various stages (such as `mock` - see #360 - and `pytest-xvfb`) in the appropriate places. I am *not* proposing to make this change in this PR.

DougBurke:
I've checked with Bill Joye and he doesn't see a problem with using a 800 by 600 pixel framebuffer, so I'm fine with this.

DougBurke:
@olaurino - I just ran a test against master, so including this change, and the ds9 tests still created visible ds9 windows. So, is the plugin actually running? And why isn't this a problem for the CI tests?

Or is this only meant to be run on the CI server?

Here's part of the test output (the error is unrelated, and is coming from AstroPy 2.0):

```
platform linux -- Python 3.6.1, pytest-3.1.2, py-1.4.34, pluggy-0.4.0
rootdir: /home/djburke/sherpa/sherpa-py36, inifile: pytest.ini
plugins: xvfb-1.0.0
collected 1366 items 

sherpa/astro/datastack/tests/test_datastack.py .E....
```

DougBurke:
Aha, I had missed this crucial note from earlier:

```
pytest-xvfb could not find Xvfb. You can install it to prevent windows from being shown.
```

so that clarifies that. Something to add to our testing documentation, when we write it!

PR #368, opened 2017-07-05
Remove ds9 warnings when run under Python 3.6

DougBurke:
# Summary

Update the DS9 code so that external processes are cleaned up properly. This removes the `ResourceWarning` warnings raised when running the `sherpa/image/tests/test_image.py` test. The Travis python 3.6 build is no longer allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,6 @@ matrix:
     # A relatively basic installation - NumPy and AstroPy - with no test data; this is similar to the preceeding
     # test, so perhaps could be combined?
     - env: INSTALL_TYPE=develop TEST=none FITS="astropy" TRAVIS_PYTHON_VERSION="3.5"
-  allow_failures:
-    # Python 3.6 build fails because of unhandled warnings coming (mostly?) from the ds9 integration.
-    - env: XSPECVER="12.9.0i" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
 
 before_install:
   # General configurations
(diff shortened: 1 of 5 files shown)


DougBurke:
This replaces PR #364. It removes the decorator and `--show-warnings` option.

DougBurke:
Should we update the python-support button in `README.md` to now say "2.7,3.5,3.6" (either as part of this PR or once this has landed)?

olaurino:
would the commit I added do?

DougBurke:
LGTM - you can see the badge at https://github.com/DougBurke/sherpa/blob/remove-ds9-warnings-with-py36/README.md

There's some other clean up needed in README.md, but I think it's more extensive than is warranted for this PR (e.g. there's mention of the latest release being 4.8.2), and should be done in a "Here's the documentation for the 4.9.1 release" PR).

olaurino:
Unfortunately, this code does not work in Python 2, where Popen does not implement the context manager protocol. I am exploring some options, but maybe we should just revert this? The real problem is that we didn't pick up the issue from the CI jobs, so I am making sure we add a ds9 smoke test, and that we require ds9 in the CI jobs that install it as a dependency.

DougBurke:
I guess the 

```
with Popen(...) as blah:
    moreblah()
```

could be changed to

```
blah = Popen(...)
try:
    moreblah()
finally:
    blah.<not sure exactly what methods to call here>()
```

but, as indicated, I'm not sure what needs to be explicitly closed.

Another option would be to add the ResourceWarning messages to `conftest.py` so that we can pass the tests with Python 3.6 (as these particular warnings are not "serious").

DougBurke:
Or there are some 2.7 solutions in https://stackoverflow.com/questions/30421003/exception-handling-when-using-pythons-subprocess-popen

I think we could define `_Popen` in Python pre 3.2 to have the necessary `enter`/`exit` methods to act as a context manager.

DougBurke:
With python 2.7 I get this:

```
% python -c 'import sherpa.astro.ui'
'AttributeError: __exit__'
```

as a quick check to see if it's still happening.

DougBurke:
See PR #373

PR #370, opened 2017-07-07
Refactor testing utilities to work with pytest

olaurino:
This was discussed in the context of #351 but we decided to issue a new PR.
This PR makes the `requires_*` decorators work with pytest, in order to achieve a better integration with other pytest markers, which required an ugly workaround in #351 itself.

I took the opportunity to split `sherpa.astro.utils`, as requested several times by @DougBurke, but the module still defines the utilities in its own namespace, for backward compatibility.

### DIFF
--- a/sherpa/utils/testing.py
+++ b/sherpa/utils/testing.py
@@ -1,0 +1,279 @@
+#
+#  Copyright (C) 2017  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+import numpy
+import unittest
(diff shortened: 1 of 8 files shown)


sherpa/utils/tests/test_utils_error_unit.py
```diff
@@ -0,0 +1,91 @@
...
+import pytest
+
+from six.moves import reload_module
+
+SKIP_MESSAGE = "This test should have been skipped"
+
+
+def test_decorator_exception_when_pytest_is_absent(monkeypatch):
+    """
+    Test that when pytest is not installed (simulated with monkeypatch) the decorators raise an exception whe
```
DougBurke: typo in comment: `whe` -> `when` ;-)

sherpa/utils/testing.py
```diff
@@ -0,0 +1,279 @@
...
+    for package in packages:
+        try:
+            importlib.import_module(package)
+            return True
+        except:
+            pass
+    return False
+
+
+def requires_package(msg=None, *packages):
```
DougBurke: Should this decorator have a different definition if `HAS_PYTEST` is `False`, since it requires `pytest.mark.skipif`?

DougBurke: I guess it comes down to whether we think of `requires_package` being an internal routine to this module or if there's ever chance it could be used directly in a test. I would prefer the second one but haven't really thought about it that deeply.

olaurino: I implicitly assumed that was not going to be used (I didn't really think about it), but it's not a safe assumption, so I'll fix that. Thanks for catching it.

DougBurke:
One question I have: do any of the existing travis builds run the smoke test without `pytest` installed, so we can check that `pytest.mark.skipif` isn't being called by accident there?

olaurino:
Yes we do, and it's the bare bone one (build number one in Travis), pytest is not installed (yet, it is installed later when the regression tests are run) but all the skippable is skipped.

PR #371, opened 2017-07-08
Hide AstroPy 2.0 deprecation warning for use of clobber parameter

DougBurke:
# Summary

Ensure that the tests can pass with AstroPy 2.0, which has introduced a warning message to note that the `clobber` parameter is deprecated.

# Details

I have left the 1.3 version of the warning suppression in, since it was in a released package (if only briefly).

PR #373, opened 2017-07-10
Fix #368 so that ds9 works with Python 2.7

DougBurke:
For Python versions prior to 3.2 add in explicit code to make _Popen work as a context manager.

# Testing

I ran `python setup.py test -a sherpa/image/tests/test_image.py ` with Python 2.7 before and after this fix and verified that the tests are now run (with no errors) rather than being skipped. I also tried

```
% python -V
Python 2.7.13 :: Continuum Analytics, Inc.
% python -c 'import sherpa.astro.ui'
```

and there is also no-longer any message about DS9 failing to load because of an `__exit__` attribute error.

I have not made any changes to the test suite as it looks like @olaurino is working on that.

### DIFF
--- a/sherpa/image/DS9.py
+++ b/sherpa/image/DS9.py
@@ -237,8 +237,30 @@ def setup(doRaise=True, debug=False):
 
         if doRaise:
             raise RuntimeErr('badwin', _ex)
+
+    elif sys.version_info < (3, 2):
+        class _Popen(subprocess.Popen):
+            # Add in the necessary methods so it can be used as a
+            # context manager (for Python 3.1 and earlier).
+            #
+            # See https://stackoverflow.com/a/30421047
+            #
+            def __enter__(self):
+                return self
+
+            def __exit__(self, type, value, traceback):
+                if self.stdout:
+                    self.stdout.close()
+                if self.stderr:
+                    self.stderr.close()
+                if self.stdin:
+                    self.stdin.close()
+                # Wait for the process to terminate, to avoid zombies.
+                self.wait()
+
     else:
         _Popen = subprocess.Popen
+
     return _SetupError
 
 


DougBurke:
Looks like the tests are no-longer being skipped: see

https://travis-ci.org/sherpa/sherpa/jobs/252211443#L936

DougBurke:
The two test failures are because we are now running with astropy 2.0 for some of the tests, and the clobber warning (fixed in pr #371) causes the failures.

There does seem to be some strange interaction with PR #369 in one of the failing tests - see

https://travis-ci.org/sherpa/sherpa/jobs/252211447#L747

I was going to suggest this was an error due to not having `pytest` installed, but actually that doesn't look correct.

sherpa/image/DS9.py
```diff
@@ -237,8 +237,30 @@ def __init__(self, *args, **kargs):
 
         if doRaise:
             raise RuntimeErr('badwin', _ex)
+
+    elif sys.version_info < (3, 2):
```
DougBurke: An alternative to this (that is a bit-more Pythonic) is

    elif hasattr(subprocess.Popen, '__enter__'):
        _Popen = subprocess.Popen
    else:
        class _Popen(subprocess.Popen):
            ... define __enter__ and __exit__

olaurino: I don't mind using `sys.version_info`, it will also be easier to spot these changes if and when we decide not to support Python 2 any more.

olaurino:
Rebasing to get the changes to the warnings whitelist (#371). Original tip was 7468530

sherpa/image/DS9.py
DougBurke: I think it would only really be a concern if we were trying to still support old versions of Python (pre 2.7), so I'm happy sticking with `sys.version_info` for now :-)

PR #375, opened 2017-07-11
add ds9 to smoke test and refactor dependency smoke test

olaurino:
This PR adds a `ds9` option to the smoke tests. When DS9 is supposed to be installed (as in some of the Travis jobs and in CIAO installations) one can have the smoke test to fail if DS9 is absent or not useful (see for instance this Travis job: https://travis-ci.org/olaurino/sherpa/jobs/252469728).

I'll rebase this PR when #373 is merged so that this PR can pass the tests.

Note that I also used this change to refactor the smoke test in a saner, more readable way, which also results in a better output.

### DIFF
--- a/sherpa/astro/utils/smoke.py
+++ b/sherpa/astro/utils/smoke.py
@@ -27,10 +27,12 @@ import logging
 import sys
 import os
 
+from sherpa.utils.err import RuntimeErr
+
 logger = logging.getLogger("sherpa")
 
 
-def run(verbosity=0, require_failure=False, fits=None, xspec=False):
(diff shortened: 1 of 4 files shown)


olaurino:
Rebased, original tip was 0410ab26

sherpa/astro/utils/smoke.py
```diff
@@ -27,10 +27,12 @@
 import sys
 import os
 
+from sherpa.utils.err import RuntimeErr
```
DougBurke: Is this import needed?

olaurino: I used to have my IDE set up to optimize imports automatically. I still think in those terms. Thanks for catching it.

PR #376, opened 2017-07-11
Release/4.9.1

olaurino:


README.md
```diff
@@ -551,6 +549,8 @@ extend (whether to other areas of Astronomy or in other domains).
 Release History
 ---------------
 
+4.9.1: 01 August 2017 [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.825839.svg)](https://doi.org/10.5281/zenodo.825839)
```
DougBurke: The badge at the top of the file will also need updating with this info to, I believe.

DougBurke: Oops - disregard this comment. It's just me being stupid.

DougBurke: Oops II: no, I was right the first time. It should be a the top of the file too. I promise this is the last comment here.....

olaurino: Hmm, the one at the top is a pointer to the latestdoi, isn't it? So it will point to the right version when we publish it. Or am I missing something?

DougBurke: Okay, so this wasnt' my last comment on this. So, zenodo has a "latest" version URL then that doesn't have to be changed when there's a release? Sweet.

DougBurke:
Presumably the recipes folder has been deleted because that code is now in a separate repository?

olaurino:
Yes, it's been in a separate repository for a while, at least since 4.8.2.

PR #379, opened 2017-07-21
Use a line not span to draw horizontal line (fix #378)

### LINKED ISSUE #378: y=0 line is not drawn in residual-style plots with matplotlib 2.0
With Sherpa 4.9.0 and matplotlib 1.5.3 the following creates a plot where there is a line drawn to indicate y=0 (as intended). If you switch to matplotlib 2.0 the y=0 line is not drawn. This behavior also holds in the 4.9.1 release candidate.

```
from sherpa import ui

ui.load_arrays(1, [1, 2, 3], [10, 17, 20])
ui.set_source(ui.const1d.mdl)
mdl.c0 = 15

ui.plot_resid()
```

DougBurke:
# Summary

Ensure that  the line at y=0 (residual) or y=1 (ratio) is drawn for residual or ratio plots when using matplotlib 2.0.

# Details

This is not needed for the 4.9.1 release.

My guess is that the handling of zero-width spans has changed between
version 1.5 and 2.0 of matplotlib. It is also not clear why axhspan washere
being used in the first place, since axhlie is used in another part of
the module.

There is an attempt to make the line match the axis (by checking what the
linewidth is), since without this change it is visually distracting in
matplotlib 2.0.

As we don't have a systematic check of these plots I have not tried to add any tests of this change. I have manually verified this works with both maplotlib 2.0 and 1.5 using python 3.5 on a linux machine.

### DIFF
--- a/sherpa/plot/pylab_backend.py
+++ b/sherpa/plot/pylab_backend.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2010, 2015, 2017  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -250,11 +250,25 @@ def plot(x, y, yerr=None, xerr=None, title=None, xlabel=None, ylabel=None,
         if val is not None:
             getattr(line, 'set_' + var)(val)
 
+    # Should the color for these lines be taken from the current axes?
+    #
+    # Using black (color='k') and the default line width (of 1) in
+    # matplotlib 2.0 produces a slightly-discrepant plot, since the
+    # axis (and tick labels) are drawn with a thickness of 0.8.
+    #
+    #
+    try:
+        lw = axes.spines['left'].get_linewidth()
+    except:
+        # should restrict to a particular error, but I do not know
+        # the axis structure enough to know what can go wrong
+        lw = 1.0
+
     if xaxis:
-        axes.axhspan(ymin=0, ymax=0, xmin=0, xmax=1)
+        axes.axhline(y=0, xmin=0, xmax=1, color='k', linewidth=lw)
 
     if ratioline:
-        axes.axhspan(ymin=1, ymax=1, xmin=0, xmax=1)
+        axes.axhline(y=1, xmin=0, xmax=1, color='k', linewidth=lw)
 
     #pylab.draw()
 


olaurino:
I would be OK with this change. However, I took the opportunity to check what it would take to add a proper test. Not much, apparently, so I am pushing a new commit to see whether Travis agrees. We can roll back the changes if needed.

The bigger issue is how to deal with the changes in CIAO-world, as the new commit introduces two new test-only dependencies: pytest-mpl and nose. But one step at a time.

olaurino:
Rebasing to resolve conflict (I added with the new commit, but we can still roll back as much as we need to). Original tip were e1b65ee and my new 23cc693

DougBurke:
So will `pytest-mpl` be required even if I am testing Sherpa without `matplotlib` installed (outside of CIAO)?

olaurino:
At this point Travis is failing because the `sherpa_test` script doesn't install the dependencies. I wouldn't think that requiring `pytest-mpl` (and installing it if not present) is a problem, but maybe it is?

Also, inside CIAO we actually don't need any of this. The complexity lies in all the different ways in which the tests can be run from both the source code and the binaries, inside and outside CIAO.

olaurino:
I did some testing in CIAO, and this should work in that set up as well, where the tests are skipped because `matplotlib` is missing and the `pytest.ini` file is not used. Assuming we can make Travis happy, I guess we have two options. One is to include the tests now and make sure it's a sustainable solution. The other is to defer the inclusion of the tests, and just take the "trivial" bug fix, which we have tested manually and independently. I would go with the first option, because we are fairly far from a release and it would be easy to remove the tests if we changed our mind. I don't think this change would require any changes in the CIAO setup.

olaurino:
So, this might be less sustainable than I thought. Travis tests are failing with image comparison differences which may depend on the actual version of matplotlib used to create and then compare the images. I thought that using a style might improve this, but actually it might even make things worse, because the style might change across versions. I'll make one more try by removing the style. If it doesn't work we should probably just go back to not adding the tests.

olaurino:
Still seeing errors, including one having to do with an "Invalid DISPLAY variable". This is taking way longer that what I intened, so I'll revert back to the original fix, approve, and merge the PR. We can try and enable plot tests at a later time.

DougBurke:
For your "differences in image output" issue, Tom (pytest-mpl author) suggests that it seemed to be okay if you used a reasonable tolerance - e.g. https://www.reddit.com/r/Python/comments/3b6oo1/pytestmpl_a_pytest_plugin_to_compare_the_output/cskfzjr/ - but this is an old post and things/API may have changed since then.

PR #381, opened 2017-07-26
Fix `bdist_wheel` command

olaurino:
# Release Note
We reviewed our customization of the `setuptools` commands as we realized that a recent bug fix (#335) introduced a regression in the `bdist_wheel` command used by `pip` to build Sherpa from source distributions. Both commands should now properly work.

# Notes
See [this comment](https://github.com/sherpa/sherpa/pull/335#issuecomment-277039877) for how to reproduce the original issue with dependencies not being installed during an `install`.

The other issue is that using the current master branch `python setup.py bdist_wheel` produces and error:

````
running install
warning: sherpa_config: built configure string['./configure', '--prefix=/export/working/sherpa/build', '--with-pic', '--enable-standalone', '--disable-maintainer-mode', '--enable-stuberrorlib', '--disable-shared', '--enable-shared=libgrp,stklib', '--enable-fftw', '--enable-region', '--enable-group', '--enable-stk', '--enable-wcs']

Checking .pth file support in build/bdist.linux-x86_64/wheel/
/export/miniconda/miniconda/envs/sherpa-pypi/bin/python -E -c pass
TEST FAILED: build/bdist.linux-x86_64/wheel/ does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    build/bdist.linux-x86_64/wheel/

and your PYTHONPATH environment variable currently contains:

    ''

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://setuptools.readthedocs.io/en/latest/easy_install.html#custom-installation-locations


Please make the appropriate changes for your system and try again.
````

With this PR both issues should be fixed.

I believe the implementation of the install command predates the realization we could do a lot with aliases in `setup.cfg`, which is what this PR does for `install` now, thus simplifying the whole thing.


olaurino:
I tested that this fixes the pip installation. While we should indeed add bdist_wheel to the tests, I am wondering if Travis is the right place to do it. Or maybe we should start having builds that are only triggered when we merge into the master branch, but I am not sure this is possible in Travis.

PR #383, opened 2017-08-06
Replace 0-energy bins in ARF and RMFs (fix #332)

### LINKED ISSUE #332: loading pha with energy_lo=0 gives nan/inf
Not sure if this has been addressed in another bug already.

When loading a pha file where the lowest energy bins starts at 0 (e.g. 0-0.01), calc_stat returns a nan/inf if cstat is selected. I think this is because some models try to integrate and divide by 0.

This occurs for all files downloaded from swift.ac.uk.
You can find some example files: https://github.com/JohannesBuchner/BXA/tree/feature-auto/examples/sherpa/swift

A workaround is:
load_pha(id, 'interval0pc.pi')
if get_rmf(id).energ_lo[0] == 0: get_rmf(id).energ_lo[0] = 0.01
if get_arf(id).energ_lo[0] == 0: get_arf(id).energ_lo[0] = 0.01

xspec handles this gracefully.

DougBurke:
# Release Note

Sherpa now performs some validation of the energy ranges of ARF and RMF files against the OGIP standard. If the ranges are inconsistent then the code will return with an error. If one energy bound is 0 the bound is replaced with a small number to avoid numerical issues like division by zero. A new configuration option `minimum_energy`, if present, allows users to override Sherpa's default behavior.

Fix #332.

# Status

Ready for comment.

# Details

This is to look at how to address #332 - handling of ARF/RMF where an energy bound is 0, which is against the spec [*] but some responses are written like this. The aim is to make a series of commits so I can check that the various tests and fixes work.

The first commit adds some basic tests of the low-level instrument API; this repeats some of the existing tests but does so without the need for the test data or an I/O library.

There are several clean up commits, and then the first major one implementing the functionality is in 8def70e - which adds the ability to replace 0-energy bins in ARFs either explicitly (when using the `DataARF` constructor), or via a new configuration element - the `minimum_energy` setting of the new `ogip` section in the `sherpa.rc` or `sherpa-standalone.rc` file [**] - for users of the UI layer. Commit a62041f expands this to RMFs, so finishing off the initial version of this patch.

[*] well, I thought it was against the spec, but it actually doesn't seem to be spelled out explicitly. XSPEC definitely replaces 0-energy elements.

[**] I'm not sure how this gets installed in the standalone build

# Notes

At present the choice of what to do when the minimum ENERG_LO is 0 is either replace it or error out. I think this should probably be changed to replace it or let it through. The latter is not ideal (in that it leads to possible problems, as shown by #332), but is more flexible (e.g. if the user wants to handle the case differently). I may make this change to the PR. I would still expect to error out if any energies are negative (not that I've ever seen this).

I'm not sure of the usability aspects of this patch, namely

  - a warning message for both ARF and RMF (I could imagine that in XSPEC this may only happen
    once)
  - using the config file to set the replacement value/behavior (the other obvious option would be a
    parameter for each function that can load an ARF or RMF, but I don't think that makes sense
    for the UI later)


DougBurke:
An example:

# Before this commit

In this case I'm using CIAO 4.9, but the same is true of the current `master` branch:

```
% sherpa
CIAO 4.9 Sherpa version 1 Friday, December 2, 2016

sherpa-1> load_pha('sherpa-test-data/sherpatest/target_sr.pha')
sherpa-2> load_arf('sherpa-test-data/sherpatest/target_sr.arf')
sherpa-3> load_rmf('sherpa-test-data/sherpatest/swxpc0to12s6_20130101v014.rmf.gz')
sherpa-4> get_arf().energ_lo[0]
          0.0
sherpa-5> get_rmf().energ_lo[0]
          0.0
sherpa-6> set_source(xspowerlaw.pl)
sherpa-7> calc_stat()
/home/djburke/local/u35/ciao-4.9/lib/python3.5/site-packages/sherpa/astro/instrument.py:190: RuntimeWarning: divide by zero encountered in true_divide
  self.lo, self.hi = DataPHA._hc / self.ehi, DataPHA._hc / self.elo
          nan
```

Note that the statistic evaluates to `nan` (this doesn't happen with the Sherpa `powlaw1d` model since it has a special case for a bin starting at e=0), and the responses have a 0-energy bin.

# New behavior

With a62041f I get

```
% ipython
Python 3.5.3 |Continuum Analytics, Inc.| (default, Mar  6 2017, 11:58:13) 
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [2]: ui.load_pha('sherpa-test-data/sherpatest/target_sr.pha')

In [3]: ui.load_arf('sherpa-test-data/sherpatest/target_sr.arf')
/home/djburke/sherpa/sherpa-fix-invalid-ogip-responses/sherpa/astro/data.py:178: UserWarning: The minimum ENERG_LO in the ARF 'sherpa-test-data/sherpatest/target_sr.arf' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

In [4]: ui.load_rmf('sherpa-test-data/sherpatest/swxpc0to12s6_20130101v014.rmf.g
   ...: z')
/home/djburke/sherpa/sherpa-fix-invalid-ogip-responses/sherpa/astro/data.py:178: UserWarning: The minimum ENERG_LO in the RMF 'sherpa-test-data/sherpatest/swxpc0to12s6_20130101v014.rmf.gz' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

In [5]: ui.get_arf().energ_lo[0]
Out[5]: 1e-10

In [6]: ui.get_rmf().energ_lo[0]
Out[6]: 1e-10

In [7]: ui.set_source(ui.xspowerlaw.pl)

In [8]: ui.calc_stat()
Out[8]: 722908286.2342303

```

Now a warning message is provided to point out that the response has been modified, the modification value of 1.0e-10 comes from the sherpa resource file (which defaults to 1.0e-10), and the `calc_stat` call does not return `nan`.

olaurino:
Rebasing and resolving conflicts. Original tip was at a62041f. @DougBurke the conflicts were about docstrings. I think the master branch had the most accurate docstrings, but it would be good if you could confirm, e.g. by trying to merge master into a62041f to view the conflicts, and then diffing `sherpa/astro/io/__init__.py` from a62041f to 6ce4e416 to check I did something sensible with those resolutions.

sherpa/astro/io/__init__.py
```diff
@@ -58,6 +58,30 @@
 io_opt = config.get('options', 'io_pkg')
 io_opt = str(io_opt).strip().lower()
 
+# Python 3 allows the following (although should move to the dictionary
+# style provided by the mapping protocal access.
+#
+# ogip_emin = config.get('ogip', 'minimum_energy', fallback='1.0e-10')
+if config.has_option('ogip', 'minimum_energy'):
```
olaurino: I must be missing something, because I can't find where this configuration option is used. I see it's used to set `data['ethresh']` in `read_arf` and `read_rmf` but then I can't find any usage of this.

sherpa/astro/data.py
```diff
@@ -83,8 +84,149 @@ def _notice_resp(chans, arf, rmf):
...
+    Several checks are made, to make sure the parameters follow
+    the OGIP standard. If the checks fail a DataErr is raised.
+    When ethresh is set, the case where the low-edge of the start bin
+    is zero is treated as a special case rather than an error.
+    If a replacement is made (i.e. the low edge is set to ethresh) then
+    a warning is displayed.
+
+    Parameters
+    ----------
+    rtype : {'ARF', 'RMF'}
```
olaurino: It looks like this argument is only used in the error/warning messages, which seems odd. Also, there is a lot of potential for refactoring, with the goal of reducing the complexity. My feelings are not too strong about it, at least until the time comes (if ever) when we have to add more validation criteria. So unless @DougBurke thinks we should follow up on this, and as long as we discuss the cases that are not currently covered by tests, I am fine with the current implementation.

olaurino:
> [**] I'm not sure how this [sherpa-standalone.rc] gets installed in the standalone build

Unlike CIAO, it doesn't, as described in: http://cxc.cfa.harvard.edu/contrib/sherpa/#configuration-files

Maybe we should better document how to create (or, rather, copy and edit) the configuration file, but the fact that nobody ever complained makes me think this feature is not used much.

DougBurke:
@olaurino - for the docstring changes in the rebase. When I did it I got a slightly different result - since I merged the two docstrings slightly. However, I don't think they are significantly different and so could be adjusted with a later commit (e.g. if we resolve the ogip configuration block issue you mention above https://github.com/sherpa/sherpa/pull/383#issuecomment-355123109).

For reference I ended up with:

```
def read_arf(arg):
    """Create a DataARF object.

    The ogip configuration block is used to control the behavior
    in cases where the data does not match the OGIP standard.

    Parameters
    ----------
    arg
        If a string then the name of the file, otherwise it is
        backend specific (e.g. a Crate for pycrates).

    Returns
    -------
    data : sherpa.astro.data.DataARF
        The ARF.
    """
```

and

```
def read_rmf(arg):
    """Create a DataRMF object.

    The ogip configuration block is used to control the behavior
    in cases where the data does not match the OGIP standard.

    Parameters
    ----------
    arg
        If a string then the name of the file, otherwise it is
        backend specific (e.g. a RMFCrate for pycrates).

    Returns
    -------
    data : sherpa.astro.data.DataRMF
        The RMF.
    """
```

sherpa/astro/data.py
DougBurke: When you talk about refactoring, how would you approach it (not this function itself, but the functionality in the PR)? I did what I felt made most sense to me, but that doesn't mean there aren't better approaches out there (i.e. I did not spend a lot of time exploring options).

sherpa/astro/io/__init__.py
DougBurke: data is used to set up the keyword arguments when initializing the `DataARF` or `DataRMF` objects - e.g. lines 371 and 397 below which say

```
    return DataARF(filename, **data)
    return DataRMF(filename, **data)
```

The constructors for these classes use the `ethresh` argument to set up the minimum energy threshold, which eventually gets passed to the `validate_ogip_energy_ranges` routine you commented on above.

sherpa/astro/io/__init__.py
```diff
@@ -337,6 +361,13 @@ def read_arf(arg):
 
     """
     data, filename = backend.get_arf_data(arg)
+
+    # It is unlikely that the backend will set this, but allow
+    # it to override the config setting.
+    #
```
DougBurke: I wish now I'd been a bit-more explicit in this comment: why am I checking for `emin` being set to decide whether to set `ethresh`? I know I was concerned about not making the code too rigid, but I'd need to go and review the `get_arf_data` and `get_rmf_data` routines in both backends to see exactly what they are doing

sherpa/astro/data.py
olaurino: Validating input files is an important functionality, one that I feel is currently fragmented or, conversely, too centralized in the back end methods, which are rather opaque and complex, and there is a lot of duplication. But a) this does not need to be addressed by this PR and b) the function introduced by this PR is complex but not too much, so as I said I don't feel strongly about it. We can refactor the next time we need to add validation rules, if ever. As to how, I'd create classes for each input file type, rmfs, arfs, but also maybe mission file types (e.g. swift vs chandra vs xmm), with a factory method/function that returns the correct class based on the header metadata, with such classes being independent of the back end. But again, this is too much to ask in the scope of this PR (or others we have reviewed and merged in the recent past), and we have more pressing priorities anyway.

DougBurke:
I believe these failures are explained by #425 and #426

sherpa/astro/data.py
```diff
@@ -83,8 +84,149 @@ def _notice_resp(chans, arf, rmf):
...
+      - elo is monotonic (ascending or descending)
+      - when emin is set, the lowest value in elo is >= 0,
+        otherwise it is > 0.
+      - ethresh (if set) is less than the minimum value in ENERG_HI
+
+    A failed check raises a DataErr exception.
+
+    """
+
+    if ethresh is not None and ethresh <= 0.0:
```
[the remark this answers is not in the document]
DougBurke: Added in 2d02ee0

sherpa/astro/data.py
```diff
@@ -83,8 +84,149 @@ def _notice_resp(chans, arf, rmf):
...
+                      'has at least one bin with ENERG_HI < ENERG_LO')
+
+    # if elo is monotonically increasing, all elements will be True
+    #                         decreasing,                      False
+    #
+    # so the sum will be number of elements or 0
+    #
+    increasing = numpy.diff(elo, n=1) > 0.0
+    nincreasing = increasing.sum()
+    if nincreasing > 0 and nincreasing != len(increasing):
```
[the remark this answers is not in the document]
DougBurke: Fixed in a7c7b01 (which I haven't pushed yet)

sherpa/astro/data.py
```diff
@@ -83,8 +84,149 @@ def _notice_resp(chans, arf, rmf):
...
+        raise DataErr('ogip-error', rtype, label,
+                      'has a non-monotonic ENERG_LO array')
+
+    if nincreasing == 0:
```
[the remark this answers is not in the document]
DougBurke: Added in 63c2a31 or 7cb8deb.

sherpa/astro/data.py
```diff
@@ -83,8 +84,149 @@ def _notice_resp(chans, arf, rmf):
...
+    e0 = elo[startidx]
+    if ethresh is None:
+        if e0 <= 0.0:
+            raise DataErr('ogip-error', rtype, label,
+                          'has an ENERG_LO value <= 0')
+    else:
+        # TODO: should this equality be replaced by an approximation test?
+        if e0 == 0.0:
+
+            if ehi[startidx] <= ethresh:
```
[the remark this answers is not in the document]
DougBurke: Added in 63c2a31 or 7cb8deb.

sherpa/astro/data.py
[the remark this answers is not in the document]
DougBurke: It's not clear whether this condition is valid (in that I believe the OGIP standard means that the energy array should always increase). I have added a test to exercise the condition just in case.

DougBurke:
I've added tests to exercise some of the error conditions I added that @olaurino pointed out were not being exercised. In doing so

a) there's an error that is being raised as `ValueError` (when creating a `DataARF` or `DataRMF` objecting and setting the `ethresh` parameter to a non-positive value); this could be converted to an error derived from `sherpa.utils.err.SherpaErr` but I don't think it's a useful change.

b) There is a test added to check when the energy grid is monotonically decreasing, but I don't think that the file would be considered to meet the OGIP standard if written out. I do not know if there are files out there that look like this (I have seen some apparently-odd things when looking at some grating-data responses, but don't have an example at hand).

c) I have tried to answer questions about the use of the configuration file/setup, but I'm not sure I've addressed your concerns. Do we maybe need more tests?

sherpa/astro/io/__init__.py
olaurino: Ah! Thanks! Hardly explicit, but that's not your fault :+1:

DougBurke: Does this need commenting, or restucturing?

olaurino: No, actually I am not sure how I missed it to begin with. I'll blame python ;)

olaurino: Just to say it, this is an example of the issues I see in the backend implementation. The `get_arf_data` is almost identical in the two backends, and really what differs is just how the rows and columns are extracted from the file according to different APIs (the various `_try_*` functions). So, when you read this code, a lot of things and validations happen in different places. Some data is populated and to some extent validates in the so-called backend, now some data is validated/populated in the `io` module, while some other info is populated/validated in the constructor (`validate_ogip_energy_ranges`). For what is worth I now understand your original question about where the best place to put the validation is. Let me see if I can do something sensible to untangle this in the context of this PR.

olaurino:
@DougBurke I pushed some minor refactoring. Please let me know what you think. I wish I could have done more, and I have a deeper attempt at refactoring sitting in my repo for a future attempt.

olaurino:
It's hard to say if the error is at all related to my refactoring... but I knew xspec was going to be acting up close to the code freeze and/or release.

olaurino:
Ah, I think this is related to the numpy 1.14 issue (#426). I worked around it in #419. I am rebasing this PR. Original tip was 9c23a8a

olaurino:
Apparently the latest numpy 1.13 conda package also brings in the libgfortran-ng package, which is making the xspec jobs on travis fail.

olaurino:
As I did for #419 I'll work around by pinning numpy to the latest version that worked with the master branch. For the CIAO release we don't particularly care, because we already have tests there and numpy is at version 1.12. But we'll need to fix #426 before the next standalone release.

olaurino:
@DougBurke I'd like to get your final OK on my changes and, if so, merge this PR.

sherpa/astro/data.py
```diff
@@ -84,107 +84,110 @@ def _notice_resp(chans, arf, rmf):
             arf.notice(bin_mask)
 
 
-def validate_ogip_energy_ranges(rtype, label, elo, ehi, ethresh=None):
-    """Check the lo/hi values are > 0, handling common error case.
+class DataResponse(Data1DInt):
```
DougBurke: I'm reluctant to add classes with no documentation, as I've no idea what this is meant to represent. Which will be useful when we look at this in six months time having forgotten this PR.

I assume that the class is an implementation detail - i.e. not really anyone who creates `DataARF` and `DataRMF` classes need to worry about or is going to be exported from this module - but the name is rather generic. Perhaps `DataOGIPResponse`?

DougBurke: If it's renamed to include `OGIP` in the class name then the validate method can drop "ogip" - e.g.  `_validate_energy_ranges`

.travis.yml
```diff
@@ -26,7 +26,7 @@ matrix:
       sudo: required
       dist: trusty
     # As above, xspec 12.9.1n
-    - env: XSPECVER="12.9.1" NUMPYVER=1.13 FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
+    - env: XSPECVER="12.9.1" NUMPYVER="1.13.3=py36ha12f23b_0" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
```
DougBurke: Perhaps there should be a comment added here to explain the reason for the pinning (just pointing to an issue/comment on GitHub is enough, I think).

olaurino:
Thanks for the comments. As soon as travis is happy I'll merge this.

JohannesBuchner:
Thank you @DougBurke!

DougBurke:
Hope it works for you @JohannesBuchner !

PR #384, opened 2017-08-25
Feature/support deprecated parameter names

olaurino:
I still need to fix the xspec models aliases (but tests are passing so we probably need to add some tests for the xspec models). Before I do that I would like to make sure @DougBurke agrees on the new design. (Plus, I have to figure out why some Travis jobs are failing, ugh.)

One change might be to make it the model class's responsibility to define the aliases. On one hand, it's the model that has to provide clients with the parameters through its attributes. On the other hand, since the `Parameter` already knows about its name, it seems to make sense it also knows about its aliases. Also, if in the future we decide to clean up the design and make parameters class level instances, e.g. through the [descriptor](https://www.smallsurething.com/python-descriptors-made-simple/) protocol, that's easier to do if the renaming is done at the parameter level, as you would still need a model initializer to setup the aliases otherwise, or some special class level descriptors just for that.

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -131,7 +131,6 @@ class Model(NoNewAttributesAfterInit):
         self.type = self.__class__.__name__.lower()
         self.pars = tuple(pars)
         self.is_discrete = False
-        # Should this set up the _renamedpars attribute if not set?
         NoNewAttributesAfterInit.__init__(self)
 
     def __repr__(self):
@@ -171,7 +170,13 @@ class Model(NoNewAttributesAfterInit):
 
     def __getattr__(self, name):
         """Access to parameters is case insensitive."""
-        lname = name.lower()
+
+        if "_par_index" == name:
+            if self.__dict__.get('_par_index') is None:
+                self.__dict__['_par_index'] = {}
+            return self.__dict__['_par_index']
+
+        lowered_name = name.lower()
 
         def warn(oname, nname):
             wmsg = 'Parameter name {} is deprecated'.format(oname) + \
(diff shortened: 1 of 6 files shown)


sherpa/models/model.py
```diff
@@ -171,72 +170,44 @@ def __iter__(self):
...
 
     def __setattr__(self, name, val):
         par = getattr(self, name.lower(), None)
         if (par is not None) and isinstance(par, Parameter):
+            # When setting an attribute that is a Parameter, set the parameter's
+            # value instead.
             par.val = val
         else:
             NoNewAttributesAfterInit.__setattr__(self, name, val)
+            if isinstance(val, Parameter):
```
DougBurke: Should we add a check here to ensure that the attribute name has the same case as the parameter name? As I mentioned in the original PR, I'm not sure that this is an intended constraint, but if it is there should be some enforcement/checks.

Or doesissue no-longer matter with the new design?

olaurino: Are you asking if we should enforce that the parameter name and the model's attribute it is assigned to have the same case? I wouldn't think that is a valid constraint. I am not really sure we should enforce them to be the same either, to be honest, but maybe we do? If we don't, does that mean we should use the parameter name (which might be different than the attribute name) as another alias for the model to respond to? We don't check that there aren't parameters overwriting other parameters either. Do we care?

DougBurke: Given that I found an issue with my implementation when this didn't hold, perhaps all I need to do is convince myself it's not a problem with the new design and then leave it at that?

olaurino: Hmm... Am I missing at least one test case?

DougBurke: Note that this is now all fuzzy in my memory, so maybe I'm making a mountain out of a non-existant molehill, but I'm thinking it's related to our discussion at https://github.com/sherpa/sherpa/pull/255#discussion_r135114243

olaurino: I see. I'll edit the relevant test to see if we are missing something.

olaurino: Doing so actually uncovered a bug both in the code and in the test.

DougBurke: Yay for testing the testing.

olaurino: Say I have a model that defines a parameter with name `Ampl` and it assigns it to `self.ampl`, with alias `NORM`. The client calls `model.norm`. What should the warning say? What's the "right" name of the parameter? The current code would say "Ampl", but maybe it should be "ampl", shouldn't it? The problem is that there are two names, one is the name of the parameter, one is the name of the model attribute.

DougBurke: Well, this is partly why I was talking about enforcing a name check at object creation time :-)

However, since the case of the name is "not important" - in that the case is ignored when using the model attribute to access the parameter name - I would use the case of the parameter name in any error message. That is stick with `Ampl` in the warning.

I doubt many users are going to see this warning (as I think the default is for Python to ignore these warnings), at least from the various REPLs.

olaurino: That's good enough for me, and what the code I was about to push does. In the end, I just had to edit the tests so they covered this case and compensate for the different expected results.

olaurino:
@DougBurke if your pair of eyes (fresher than mine, it's not like you were visually perusing hundreds of FITS keywords) could double check I didn't mess up anything in e6d3ffe, my pair of eyes would be grateful.

DougBurke:
We should review some of these "new names" with @anetasie to make sure they make sense; for instance `xszigm` has `_redshift`, `_model`, and `_lyman_limit` since the parameter file has

```
zigm           3  0.0        1.0e20          zigm      mul  0
*redshift " "     0.0
$model    " "     0
$lyman_limit " "  1
```

but I'm not convinced now that it's worth adding the leading `_` (as replacement for `*`/`$`). That may have come about because of my mis-understanding when setting up the initial list of parameter renames. I wonder what PyXSpec does here?

olaurino:
@DougBurke any news on this?

DougBurke:
Not yet. The issue is that, as mentioned above, my list of "renamed" parameters include a number which change `lyman_limit` to `_lyman_limit` because the code I wrote way back was taking the XSPEC "name" and converting the leading charater (such as `$` or `*`) to an underscore. This is probably not needed.

Issue #319, opened 2016-11-22
image_getregion returns byte string on Py3


olaurino:
Originally reported by @nplee on sherpadev.

from the Python 3 build of CIAOX, when "image_getregion" is invoked in Sherpa, a byte string is returned:

````
sherpa-3> image_getregion()
          b''
sherpa-4> image_getregion().decode()
````

I didn't have time to reproduce the issue, but I am opening a ticket so it does not fall through the cracks again.

**Comment by DougBurke** [comment]
This also happens (unsurprisingly) for `image_xpaget`:

```
sherpa-1> image_open()
sherpa-2> image_xpaget('version')
          b'sherpa 7.5\n'
```

**Comment by DougBurke** [comment]
This does cause problems - e.g. from a session using Python 3.6 where I had added a region to an image, the `notice2d_image` command fails as shown below (`ui` is `sherpa.astro.ui` here):

```
In [10]: ui.image_getregion()
Out[10]: b'circle(58.716049,55.822531,30.764058);'

In [11]: ui.notice2d_image()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-aa5314fa3698> in <module>()
----> 1 ui.notice2d_image()

<string> in notice2d_image(ids)

/lagado.real/sherpa/sherpa-warnings/sherpa/astro/ui/utils.py in notice2d_image(self, ids)
   6490             if (coord == 'world'):
   6491                 coord = 'wcs'
-> 6492             regions = self.image_getregion(coord).replace(';', '')
   6493             self.notice2d_id(id, regions)
   6494 

TypeError: a bytes-like object is required, not 'str'
```

PR #385, opened 2017-08-29
fix image_getregion in Python 3 (#319)

DougBurke:
This is a proof-of-concept. It has shown that we need to decide at what "level" of the API do we want to convert from byte string to string: is it just `sherpa.ui.utils.image_getregion` (and `image_xpaget`), or the `sherpa.image.ds9_backend` module, or `sherpa.image.DS9`?

### DIFF
--- a/sherpa/image/ds9_backend.py
+++ b/sherpa/image/ds9_backend.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 #
-#  Copyright (C) 2007, 2016  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2016, 2017  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -25,6 +25,11 @@ from sherpa.utils.err import DS9Err
 
 imager = DS9.DS9Win(DS9._DefTemplate, False)
 
+
+# TODO: the except blocks would ideally catch explicit errors; the present
+#       catch-anything approach means that we lose potentially-useful
+#       information on the type of error.
+#
 
 def close():
     if imager.isOpen():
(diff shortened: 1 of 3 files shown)


sherpa/image/ds9_backend.py
```diff
@@ -23,9 +23,16 @@
...
 from sherpa.utils.err import DS9Err
 
+import six
+
 imager = DS9.DS9Win(DS9._DefTemplate, False)
 
 
+# TODO: the except blocks would ideally catch explicit errors; the present
+#       catch-anything approach means that we lose potentially-useful
+#       information on the type of error.
```
olaurino: Yeah, unfortunately Python2 lacks a standard mechanism for chaining exceptions, but I believe they fixed that, finally, in Python 3. Not sure whether it makes sense to fix it now in a 2+3 fashion.

olaurino:
Thanks for beating me to this. I'll review it asap, and try to figure out where the decoding should happen, if not where you added it.

DougBurke:
I've removed pr:hold as I think it's ready for review to see if the fix is in the right place.

olaurino:
There are a number of conflicting PRs at this point, so not all of them are getting into the snapshot branch.

DougBurke:
Do you want to put pr:hold back on this one (and others)?

olaurino:
I'll check again tomorrow.

olaurino:
Rebasing and squashing, was 280eeb3

olaurino:
I moved the code that forces the string to be decoded to the `DS9.DS9Win` class. It is riskier to move it there because it might have unintended consequences. Several tests failed and I had to compensate, so hopefully that covers the unintended consequences. I expect users might be impacted by this if their scripts, like some of our tests, expected a bytes string rather than a proper string.

olaurino:
@DougBurke I'd like you to sign off on the latest change before merging. Thanks!

DougBurke:
Should we add documentation to the two `xpaget` routines to say which ones return "native" strings and which ones are coerced to "whatever we call the Python string" type?

Issue #234, opened 2016-07-08
ignore_bad() filter issues with models (Nick's bug)


jbudynk:
# Description

The user extracted the spectrum with specextract, then used FTools' GRPPHA to bin the spectrum to at least five counts per bin and flag specific channels as bad (1-35,501-1024). 

Simplifying the steps taken: 

```
load_pha("18076_bin5.pi") 
rmf = get_rmf() 
arf = get_arf() 
bkg_scale = get_bkg_scale() 
set_full_model(rmf(arf(xswabs.abs1*xszwabs.abs2*powlaw1d.pl1)))+bkg_scale*rmf(arf(xswabs.bkgabs*(xsapec.bkgapec))+(powlaw1d.bkgpl+xsgaussian.bkgg1+xsgaussian.bkgg2)))) 
ignore_bad() 
fit() 
*** 
WARNING: data set 1 has associated backgrounds, but they have not been 
subtracted, nor have background models been set 
ValueError: shape mismatch between 'ARFModelPHA: 1070' and 'BinaryOpModel: 
1063' 
***
```

throws an error. I'm pretty sure it has to do with the 'bkg_scale' component of the model, where the APEC component is convolved with the response while the power-law+Gaussian component is not.  Extending this out to using the ungrouped data in this session, and using group_counts and adding quality flags to the data from within Sherpa, it also fails: 

```
load_data(2,"18076.pi") 
quality = numpy.zeros(len(get_data().channel)) 
for i in range(len(quality)): 
     if i+1<=35: 
         quality[i]=-1 
     elif i+1>=501: 
         quality[i]=-1 
set_quality(2,quality) 
group_counts(2,5) 
set_full_model(2,rmf(arf(xswabs.abs1*(xszwabs.abs2*powlaw1d.pl1)))+bkg_scale*(rmf(arf(xswabs.bkgabs*(xsapec.bkgapec))+(powlaw1d.bkgpl+xsgaussian.bkgg1+xsgaussian.bkgg2)))) 
ignore_bad(2) 
fit(2) 
*** 
WARNING: data set 2 has associated backgrounds, but they have not been 
subtracted, nor have background models been set 
ValueError: shape mismatch between 'ARFModelPHA: 1070' and 'BinaryOpModel: 
1063 
*** 
```

Starting a new Sherpa session and loading the ungrouped spectrum of that source gives no issues. The fit works fine. 

```
import numpy
load_data("18076.pi") 
[...]

Issue #233, opened 2016-07-08
Grouping methods overwrite quality flags (part of Nick Lee's bug)


jbudynk:
# Description

The quality flags of a `DatasetPHA` are overwritten when a grouping method is called. Channels that were originally marked as bad (quality > 0) can be marked as good (quality == 0) after grouping so long as the group condition is met for those channels.

The quality flags should not be overwritten after grouping. The quality flags are a "static" part of the dataset, meaning only the user should be given the power to edit these flags; grouping should not affect them. 

If the user saves a grouped dataset to file, then it could make sense to _merge_ the bad group flags with the basic quality flags. 

# Notes

It is still undecided how bad quality groups should be marked. Some ideas are:

 - adding a "bad_group_quality" array to the DataPHA class.
 - re-designing the grouping and filtering methods and objects for datasets.

PR #388, opened 2017-08-31
Address indexing DeprecationWarning/IndexError from NumPy for PHA data with filter + ignore_bad (fix #361)

### LINKED ISSUE #361: numpy 1.13 error "boolean index did not match indexed array along dimension 0" (VisibleDeprecationWarning with numpy 1.12)
The following code creates a `VisibleDeprecationWarning` (I'm using one of the data sets in sherpa/sherpa-test-data/) with numpy 1.12 but it fails with numpy 1.13:

```
from sherpa.astro import ui

ui.load_data('/path/to/sherpa-test-data/sherpatest/q1127_src1_grp30.pi')

ui.set_stat('chi2xspecvar')
ui.set_source(ui.xspowerlaw.pl * ui.xsphabs.gal)

ui.calc_stat_info()

ui.notice(0.5, 8.0)
ui.ignore_bad()

ui.calc_stat_info()

ui.notice(0.5, 8.0)
ui.calc_stat_info()
```

The output (here using a Python 3.5 build of CIAO 4.9 but also seen in the standalone) is (where I've elided the initial output):

```
...
/home/djburke/local/u35/ciao-4.9/lib/python3.5/site-packages/sherpa/astro/data.py:1557: VisibleDeprecationWarning: boolean index did not match indexed array along dimension 0; dimension is 1024 but corresponding boolean dimension is 996
  chans = chans[mask]
/home/djburke/local/u35/ciao-4.9/lib/python3.5/site-packages/sherpa/astro/data.py:1557: VisibleDeprecationWarning: boolean index did not match indexed array along dimension 0; dimension is 1024 but corresponding boolean dimension is 996
  chans = chans[mask]
Dataset               = 1
Statistic             = chi2xspecvar
Fit statistic value   = 9.20204e+10
Data points           = 410
Degrees of freedom    = 407
Probability [Q-value] = 0
Reduced statistic     = 2.26094e+08
```

The `VisibleDeprecationWarning` warnings are coming from the last `calc_stat_info` call - i.e. after the `ignore_bad` and `notice` calls.

Using NumPy 1.13 (Python 3.6) it now errors out:

```
  File "../dep.py", line 16, in <module>
    ui.calc_stat_info()
  File "<string>", line 1, in calc_stat_info
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/ui/utils.py", line 7547, in calc_stat_info
    output = self._get_stat_info()
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/astro/ui/utils.py", line 9468, in _get_stat_info
    statinfo = f.calc_stat_info()
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/fit.py", line 1192, in calc_stat_info
    stat, fvec = self._calc_stat()
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/fit.py", line 1123, in _calc_stat
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/stats/__init__.py", line 501, in calc_stat
    fitdata, modeldata = self._get_fit_model_data(data, model)
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/stats/__init__.py", line 198, in _get_fit_model_data
    modeldata = data.eval_model_to_fit(model)
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/data.py", line 503, in eval_model_to_fit
    total_model.append(data.eval_model_to_fit(func))
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/astro/data.py", line 1267, in eval_model_to_fit
    return self.apply_filter(modelfunc(*self.get_indep(filter=True)))
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/astro/data.py", line 643, in get_indep
    return (self.get_noticed_channels(),)
  File "/home/djburke/sherpa/sherpa-mine-numpy113/sherpa/astro/data.py", line 1760, in get_noticed_channels
    chans = chans[mask]
IndexError: boolean index did not match indexed array along dimension 0; dimension is 1024 but corresponding boolean dimension is 996
```

DougBurke:
The first commit only added tests: it is expected to fail with NumPy 1.13 due to an `IndexError` and
with earlier NumPy warnings due to an uncaught `DeprecationWarning`.

The second commit is the simplest fix I could come up with. It is not clear whether it is sensible or sufficient, and how it interacts with other issues that @jbudynk has been looking at, such as #234, #233, and #149.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2182,6 +2182,18 @@ class DataPHA(Data1D):
         chans = self.channel
         mask = self.get_mask()
         if mask is not None:
+
+            # This is added to address issue #361
+            #
+            # If there is a quality filter then the mask may be
+            # smaller than the chans array. It is not clear if this
+            # is the best location for this. If it is, then are there
+            # other locations where this logic is needed?
+            #
+            if self.quality_filter is not None and \
+               self.quality_filter.size != mask.size:
+                chans = chans[self.quality_filter]
+
             chans = chans[mask]
         return chans
 
--- a/sherpa/astro/ui/tests/test_filtering.py
+++ b/sherpa/astro/ui/tests/test_filtering.py
@@ -1,0 +1,166 @@
+#
+#  Copyright (C) 2017, 2018  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+# Additional tests of filtering.
+
+import logging
+
+import pytest
+
+from sherpa.astro import ui
+from sherpa.utils.testing import requires_data, requires_fits
+
+
+# Check that a logging message is created: this is based on the
+# code in https://stackoverflow.com/a/1049375 and
+# https://stackoverflow.com/a/20553331
+#
+# Alternatively, we could require textfixtures. Can this be
+# simplified once we drop Python versions < 3.4?
+#
+# Could this cause problems when running the whole test suite?
+#
+class MockLoggingHandler(logging.Handler):
+    """Mock logging handler to check for expected logs.
+
+    Messages are available from an instance's ``messages`` dict,
+    in order, indexed by a lowercase log level string (e.g., 'debug',
+    'info', etc.).
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.messages = {'debug': [], 'info': [], 'warning': [], 'error': [],
+                         'critical': []}
+        super(MockLoggingHandler, self).__init__(*args, **kwargs)
+
+    def emit(self, record):
+        "Store a message from ``record`` in the instance's ``messages`` dict."
+        self.acquire()
+        try:
+            self.messages[record.levelname.lower()].append(record.getMessage())
+        finally:
+            self.release()
+
+    def reset(self):
+        self.acquire()
+        try:
+            for message_list in self.messages.values():
+                message_list.clear()
+        finally:
+            self.release()
+
+
+def setup_model(make_data_path):
+    """Set up a model that is reasonably close to the data.
+
+    Returns the expected statistic values for various filters.
+    """
+
+    infile = make_data_path('q1127_src1_grp30.pi')
+
+    ui.clean()
+    ui.load_pha(infile)
+    ui.subtract()
+
+    ui.set_stat('chi2datavar')
+    ui.set_source(ui.powlaw1d.pl)
+
+    pl = ui.get_model_component('pl')
+    pl.ampl = 5.28e-4
+    pl.gamma = 1.04
+
+    # These statistic values were created using CIAO 4.9 on a
+    # Ubuntu machine. The quality=2 values are for high energies
+    # (above ~ 10 keV or so), and so a filter of 0.5-8.0 keV should
+    # give the same answer with or without ignore_bad.
+    #
+    return {'all': 2716.7086246284807,
+            'bad': 2716.682482792285,
+            '0.5-8.0': 1127.7165108405597}
+
+
+@requires_fits
+@requires_data
+def test_filter_basic(make_data_path):
+    """Test out issue 361 without ignore_bad calls.
+
+    This should not trigger the bug behind issue 361.
+    """
+
+    stats = setup_model(make_data_path)
+
+    s1 = ui.calc_stat()
+    assert s1 == pytest.approx(stats['all'])
+
+    ui.ignore_bad()
+    s2 = ui.calc_stat()
+    assert s2 == pytest.approx(stats['bad'])
+
+    ui.notice(None, None)
+    s3 = ui.calc_stat()
+    assert s3 == pytest.approx(stats['all'])
+
+    ui.notice(0.5, 8.0)
+    s4 = ui.calc_stat()
+    assert s4 == pytest.approx(stats['0.5-8.0'])
+
+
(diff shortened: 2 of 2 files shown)


DougBurke:
The first three failures all use NumPy < 1.13 and so fail because a `DeprecationWarning` is created (although the tests themselves end up failing because of some sort an issue with how I implemented #369), and the last failure (1683.6 https://travis-ci.org/sherpa/sherpa/jobs/270236444) fails because it uses NumPy 1.13 and so the code raises an `IndexError`.

I wanted to see what the current behavior was before trying to work out what to do with the test to make it work here (the conceptually simplest would be to fix the code so there's no error :-)

DougBurke:
As mentioned in the updated comment above, the fix in 04982e9 is the simplest I could think of and I think it needs to pass more-strenuous tests than the one here before it could be considered for inclusion.

olaurino:
Rebasing, original tip was at 04982e9.

olaurino:
I was hoping (irrationally) that this might fix 233. Unsurprisingly, it doesn't. And it doesn't need to. To me, in order to do some more "strenuous" testing we need test cases that would cover the complexity of the issues that have to do with the quality, grouping, and filtering, which might not make sense. Either we decide to defer fixing this issue until we start addressing the bigger picture or we take this fix with its specificity. I'll review the actual fix as soon as I can, though.

DougBurke:
I am rather nervous about taking this fix without better testing - or someone with a better idea of how all the pieces fit together can see if it's a good idea!

olaurino:
This is likely to snowball, so maybe we should keep this bug with the others related to grouping and filtering and address it once we have enough requirements from SDS to compose the bigger picture to begin with?

DougBurke:
I've removed this from the 4.10 milestone given yesterdays face-to-face meeting: the consensus was that it's not clear whether this fix could interact with other parts of the system. If we want to support recent NumPy with the 4.10 milestone then we should revisit this.

DougBurke:
About to rebase to reflect changes on the master branch. Original tip was 440b713

DougBurke:
I don't know what travis is doing - the test failure was due to a timeout but is now due to `/home/travis/build/sherpa/sherpa/sherpa/models/tests/test_psf_rebin.py` which doesn't even exist in this branch.

olaurino:
Either we take this fix or we close this PR and resuscitate it in the future when we are ready to tackle the other overarching issues. The good news is that I tried to merge this into master and it worked despite having been open for a while.

DougBurke:
Whatever happened to @jbudynk work on the binning/filtering work? I would hope there's at least a set of tests that could be added so that we can understand the current behavior, and find out if any PRs we make change this behavior (which may be a sensible change or an undesired change)? One of the problems I have with making changes to any of this code is not knowing what the downstream consequences are, and having more tests will hopefully help that.

olaurino:
I believe you mean #225, which we closed a while ago to reduce the PR clutter. We could keep track of these closed but useful PRs in a GitHub project until we tackle the issue?

DougBurke:
Ta - I did mean #225 but was too lazy to search.

I assume that the tests from #225 would fail if added to master (without the associated code changes in #225)? What I really want is a bunch of tests so that we can feel confident that we know if we've changed downstream behavior. Maybe a New-Year project for me then!

olaurino:
one issue in #225 is that some of those tests derive from @jbudynk's research into what the expected behavior of the grouping and filtering is supposed to be, but we don't have requirements for that.

When I tried to take charge of the project I hit the same issue. I started working on BDD-style tests to make sure we were on the same page as far as what the expected behavior is, and I remember asking for SDS input, then we just dropped the project. [See the `.feature` files in https://github.com/olaurino/sherpa/tree/grouping-tests/features]

We can't work on this stuff if we don't start from defining the expected behavior. If you added a bunch of tests *deciding* what the desired outcome is in a bunch of scenarios, that would be useful indeed, assuming we want to give this project a priority.

DougBurke:
We (well, @kglotfelty) has run the SDS sherpa regression test with this PR applied to the latest CIAOX build, and there were no problems. While that is not really a sufficient test - in that I can't guarantee that we hit this use case - it is at least a sign it doesn't break things elsewhere.

DougBurke:
Well, technically the change wasn't made in 2019... :)

DougBurke:
I'm currently just trying to rebase this PR (to run it with all the changes and because now we have numpy 1.17 on travis to test), but let me if this isn't wanted.

DougBurke:
I rebased this since you couldn't reliably run the test suite with python 3.7/numpy 1.17 for this PR without it (there were a number of tests, unrelated to this PR, that were failing without the rebase).

sherpa/astro/data.py
```diff
@@ -2182,6 +2182,18 @@ def get_noticed_channels(self):
...
+            # If there is a quality filter then the mask may be
+            # smaller than the chans array. It is not clear if this
+            # is the best location for this. If it is, then are there
+            # other locations where this logic is needed?
+            #
+            if self.quality_filter is not None and \
+               self.quality_filter.size != mask.size:
+                chans = chans[self.quality_filter]
+
             chans = chans[mask]
```
wmclaugh: Unless I am missing something, the code in lines 2193-2195 are a no-op since line 2197 will override the value of chans.

DougBurke: While I haven't looked at this for a while, I think what is important hers is that `chans` and `mask` have the same size. Without this addition there are times when this doesn't hold - i.e. we need to subset `chans` to match the size of `mask`.

Does this make sense?

DougBurke:
Looking at the tests I think we can clean them up slightly (the method used to check that a warning message has been created when we expect to) now we are python 3.5 or later.

sherpa/astro/data.py
DougBurke: The issue I have with this PR is that it's not obvious to me if this is just a band aid, hiding a more serious problem, of if this is the best place to place this fix. However, at some point a band aid is better than nothing.

wmclaugh: So for the code segment, are we certain that the quality_flag True elements is always equal to the mask size?      
     (np.where(self.quality_filter == True))[0].size == mask,size

DougBurke: All I know for sure is that there is a case (#361) where the fix above stopped the warning. I was hoping that this PR would lead to someone who did understand the filtering/grouping internals would be able to say "yes" or "this should be done elsewhere", but that hasn't happened. So now I think we either take this under the assumption that it doesn't seem to make things worse and does fix a known issue, or decide that it's too risky and we should wait until someone can fix this properly.

PR #390, opened 2017-08-31
Add nlapec model

olaurino:
# Release Note
The `nlapec` XSpec model has been added. Note that XSPEC 12.9.0 is now required.

# Notes
This is supposed to replace #299. I edited the commit to have @DougBurke as author as he authored to code I edited.

This PR piggybacks #389 to run the tests with xspec 12.9.1n.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -113,6 +113,7 @@ void xsmekl_(float* ear, int* ne, float* param, int* ifl, float* photar, float* 
 void C_xsmkcf(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 // void C_xneq(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 void C_nei(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
+void C_nlapec(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 // void xshock_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
 void C_npshock(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 void nsa_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
@@ -1155,6 +1156,7 @@ static PyMethodDef XSpecMethods[] = {
   XSPECMODELFCT_NORM( xsmekl, 6 ),
   XSPECMODELFCT_C_NORM( C_xsmkcf, 6 ),
   XSPECMODELFCT_C_NORM( C_nei, 5 ),
+  XSPECMODELFCT_C_NORM( C_nlapec, 4 ),
   XSPECMODELFCT_C_NORM( C_npshock, 7 ),
   XSPECMODELFCT_NORM( nsa, 5 ),
   XSPECMODELFCT_NORM( nsagrav, 4 ),
--- a/README.md
+++ b/README.md
@@ -433,14 +433,15 @@ Then, build Sherpa in the standard way:
 XSPEC
 -----
 
-Sherpa does not support
-[`XSPEC`](https://heasarc.gsfc.nasa.gov/xanadu/xspec/) models by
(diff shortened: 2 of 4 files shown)


olaurino:
Hmmm... After reading the description of #299 I thought the nlapec model was in 12.9.0i, but I guess it's not. Or am I missing something @DougBurke. I'll do some more digging.

olaurino:
All right. I must have messed up somewhere because the model is in 12.9.0i and the 12.9.1 test is failing too.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -1155,6 +1156,9 @@ static PyMethodDef XSpecMethods[] = {
   XSPECMODELFCT_NORM( xsmekl, 6 ),
   XSPECMODELFCT_C_NORM( C_xsmkcf, 6 ),
   XSPECMODELFCT_C_NORM( C_nei, 5 ),
+#ifdef XSPEC_HAS_NLAPEC_MODEL
```
olaurino: ugh. I thought I had removed those lines.

DougBurke:
We should remove mention of 12.8.2 from `README.md`, that is remove the paragraph

```
The XSPEC module is designed for use with XSPEC versions 12.9.0 and
12.8.2, although 12.9.0 is the preferred version. Use with other
versions may lead to build problems or incomplete support for the
XSPEC models.
```

and instead say,  at the start of this section, something like

```
Sherpa can be built with support for XSPEC (link), although support is not enabled by
default. The current supported XSPEC versions are 12.9.0 and 12.9.1, and it is expected
that it will build against newer versions, but without support for new models or features.

To build the XSPEC support in Sherpa, the `xspec_config` option ... continue with the existing paragraph...
```

olaurino:
Should we just have a file for the dependencies we support/test? That would be easier to update.

DougBurke:
It might be easier for us, not so sure it's easier for the casual reader (although those fancy hyperlink things do make this easier than in the old days).

DougBurke:
Here's a list of models that appear in heasoft 6.22 that are not in Sherpa/XSPEC:

X-Spec model names not in Sherpa:
  - btapec   type=add
  - bvtapec   type=add
  - bvvtapec   type=add
  - carbatm   type=add
  - hatm   type=add
  - nlapec   type=add
  - slimbh   type=add
  - smaug   type=add
  - snapec   type=add
  - tapec   type=add
  - voigt   type=add
  - vtapec   type=add
  - vvtapec   type=add
  - ismabs   type=mul
  - tbfeo   type=mul
  - tbgas   type=mul
  - tbpcf   type=mul
  - tbrel   type=mul
  - xscat   type=mul
  - cflux   type=con
  - clumin   type=con
  - cpflux   type=con
  - gsmooth   type=con
  - ireflect   type=con
  - kdblur   type=con
  - kdblur2   type=con
  - kerrconv   type=con
  - lsmooth   type=con
  - partcov   type=con
  - rdblur   type=con
  - reflect   type=con
  - rfxconv   type=con
  - rgsxsrc   type=con
  - simpl   type=con
  - vashift   type=con
  - vmshift   type=con
  - xilconv   type=con
  - zashift   type=con
  - zmshift   type=con
  - pileup   type=acn

The pileup model, since it is type=acn, can be ignored, and the type=con results are being handled separately (#352), so that leaves 13 type=add models (one of which is `nlapec`) and 6 type=mul models that need to be reviewed for addition.

olaurino:
Rebasing to remove commit from #389. Original tip: 364e03f

olaurino:
This can probably be merged. There were a couple of open questions/issues regarding the version strings in the README (and whether to list the dependencies in an ad-hoc file).

olaurino:
I modified the README and the docstring in `xspec.__init__.py`. That should be enough to merge this PR, shouldn't it?

DougBurke:
Does this need to be rebased to pick up PR #408 (which removed the show-warnings code)?

DougBurke:
I'm happy with the changes, but the travis build shoulf be green (I think that the failure is unrelated to this build). We need to make sure that the changelog for the release that contains this PR explicitly says "The minimum supported version of XSPEC is now 12.9.0.". Should the PR commit be updated to include this information so you don't forget it when creating the release notes?

olaurino:
I wasn't expecting the Travis error, but I always wait for it to be green. I'll rebase as the failure has to do with code we changed anyway.

olaurino:
Rebasing, original tip was f98ad97

PR #391, opened 2017-09-01
make requires_data a function so it works in CIAO too

olaurino:
I just found out that since #370 those Sherpa tests that require the data folder have always been skipped in CIAO (they are executed or appropriately skipped in the Travis jobs).

This PR most likely ensures that tests are also properly executed in CIAO (we'll see on Tuesday, I just tested manually that the code works, but you never know). Note that there is no reason tests in CIAO should be skipped because of the test data missing.

The deepest reason why this is happening is obscure, but it has to do with the fact that tests are executed in a different way in CIAO, by explicitly calling pytest on the location where Sherpa is installed. This seems to yield to a different order in which `conftest.py` is executed with respect to when the tests are collected. Long story short, the tests are first marked to be skipped and then the data directory is made available to `SherpaTestCase`. There is probably a better implementation that would make this issue go away, but our setup is rather complex, as we have `unittest` and `pytest` style tests mixed together, and we want the data folder to be available in different ways, i.e. through the git submodule, as a package, or through the `--test-data` pytest argument. It doesn't make sense to review this now, but for future reference this could be a good starting point:
https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

### DIFF
--- a/sherpa/utils/testing.py
+++ b/sherpa/utils/testing.py
@@ -178,8 +178,14 @@ if HAS_PYTEST:
     #  standard library. The decorator will be defined if pytest is missing, but if the tests are run they throw
     #  and exception prompting users to install pytest, in those cases where pytest is not installed automatically.
 
-    requires_data = pytest.mark.skipif(SherpaTestCase.datadir is None,
-                                       reason="required test data missing")
+    def requires_data(test_function):
+        """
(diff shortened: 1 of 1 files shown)


olaurino:
Rebasing, original tip was cc3325f

olaurino:
I'll merge it as soon as I get the green light from Travis.

PR #394, opened 2017-09-07
XSPEC build and functionality improvements

DougBurke:
# Release notes

Add support for reading and changing the path to the XSPEC "manager" directory, and reading the current XSPEC "model" directory (`get_xspath_manager`, `set_xspath_manager`, and `get_xspath_model`). These are intended for the power user and so require an explicit import from `sherpa.astro.xspec`.

There are several improvements to the build and interface to the XSPEC model library: these have no user-visible changes.

# Details

At present this code is intended to build with XSPEC 12.9.0. There are additional routines which could be implemented if we 

- bumped out minimum version to XSPEC 12.9.1
- started using the C++ `FunctionUtility` code rather than `xsFortran`, but there are implications on the build (see the Issues section below)
- take advantage of the "optional" machinery introduced in #395

# Status

Ready for review. I've left it as pr:hold since I expect it to clash with other XSPEC-related PRs we have going.

This adds basic tests for existing XSPEC functionality, namely reading
and writing values such as the abundance and cross section tables,
as well as the chatter settings.

One test is marked as "expected fail" as it is not clear why it returns
a chatter setting of 50 rather than 0. Actually, this is down to Sherpa test suite changing the chatter level for the tests. Options are to stop this or change the expected value in the test; I prefer the former but not strongly.

The code has been changed to use XSPEC's own include file for declaring functions, rather than using our own internal copy of this file.

Interfaces to three functions have been added (read and write of the manager path, read the model path) . The "state" mechanism we use for the XSPEC module has been updated to support this (only the manager path), and tests added for the state handling (there were none).

# Issues

The state-handling tests have shown up an issue that could be addressed, but it needs thinking about and may best be done outside the PR: the problem is that if a user has used `set_xsxset` or `set_xspath_manager` and then used `load_xsstate` to load in a new XSPEC state, there is no guarantee that the changes made will be reverted. There is a relatively-clear way to do this for the xset case (explicitly set the currently-set variables to `""` before loading in a new state), but not so obvious for the path (since there's no "default" value like `""` here).

I would have liked to use the C++ `FunctionUtility` interface rather than the `xsFortran` wrapper around it which we currently use, since it has more features, and the XSPEC team suggest it's the better choice for C++ code. However, there are include issues with this - the include files in an XSPEC installation don't have the same directory structure as the build environment, which causes include problems - which makes it only really feasible to go with this approach if we use the XSPEC build location rather than the install location. This is not unprecedented - it is the only way to build user models when using the XSPEC executable - but isn't really what we want for Sherpa.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -185,11 +185,12 @@ def get_xsstate():
     set_xsstate
     """
 
+    # Do not return the internal dictionary but a copy of it.
     return {"abund": get_xsabund(),
             "chatter": get_xschatter(),
             "cosmo": get_xscosmo(),
             "xsect": get_xsxsect(),
-            "modelstrings": modelstrings}
+            "modelstrings": modelstrings.copy()}
 
 
 def set_xsstate(state):
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -68,6 +68,9 @@ except AttributeError:
 # http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSxset.html
 modelstrings = {}
 
+# Store any path changes
+xspecpaths = {}
+
 
 def get_xsxset(name):
     """Return the X-Spec model setting.
@@ -160,6 +163,40 @@ def set_xsxset(name, value):
         modelstrings[name] = get_xsxset(name)
 
 
+# Since these are not included in __all__, create a symbol in this
+# module to avoid "unused symbol" if they were to be imported
+# directly from _xspec.
+#
+get_xspath_manager = _xspec.get_xspath_manager
+get_xspath_model = _xspec.get_xspath_model
+
+
+def set_xspath_manager(path):
+    """Set the path to the files describing the XSPEC models.
+
+    Parameters
+    ----------
+    path : str
+        The new path.
+
+    See Also
+    --------
+    get_xspath_manager : Return the path to the files describing the XSPEC models.
+
+    Examples
+    --------
+    >>> set_xspath_manager('/data/xspec/spectral/manager')
+    """
+
+    _xspec.set_xspath_manager(path)
+    spath = get_xspath_manager()
+    if spath != path:
+        raise IOError("Unable to set the XSPEC manager path " +
+                      "to '{}'".format(path))
+
+    xspecpaths['manager'] = path
+
+
 # Provide XSPEC module state as a dictionary.  The "cosmo" state is
 # a 3-tuple, and "modelstrings" is a dictionary of model strings
 # applicable to certain models.  The abund and xsect settings are
@@ -168,6 +205,13 @@ def set_xsxset(name, value):
 # cosmo, xsect, and xset.
 # http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/Control.html
 # http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/Setting.html
+#
+# The path dictionary contains the manager path, which can be
+# explicitly set. It could also contain the model path, but there
+# is no XSPEC routine to change that; instead a user would set the
+# XSPEC_MDATA_DIR environment variable before starting XSPEC.
+# Should this path be included?
+#
 def get_xsstate():
     """Return the state of the XSPEC module.
 
@@ -176,8 +220,8 @@ def get_xsstate():
     state : dict
         The current settings for the XSPEC module, including but not
         limited to: the abundance and cross-section settings, parameters
-        for the cosmological model, and any XSET parameters that have
-        been set.
+        for the cosmological model, any XSET parameters that have been
+        set, and changes to the paths used by the model library.
 
     See Also
     --------
@@ -190,7 +234,8 @@ def get_xsstate():
             "chatter": get_xschatter(),
             "cosmo": get_xscosmo(),
             "xsect": get_xsxsect(),
-            "modelstrings": modelstrings.copy()}
+            "modelstrings": modelstrings.copy(),
+            "paths": xspecpaths.copy()}
 
 
 def set_xsstate(state):
@@ -200,8 +245,9 @@ def set_xsstate(state):
     ----------
     state : dict
         The current settings for the XSPEC module. This is expected to
-        match the return value of ``get_xsstate``, and so requires the
-        keys: 'abund', 'chatter', 'cosmo', 'xsect', and 'modelstrings'.
+        match the return value of ``get_xsstate``, and so uses the
+        keys: 'abund', 'chatter', 'cosmo', 'xsect', 'modelstrings',
+        and 'paths'.
 
     See Also
     --------
@@ -211,21 +257,37 @@ def set_xsstate(state):
     Notes
     -----
     The state of the XSPEC module will only be changed if all
-    the required keys in the dictionary are present.
+    the required keys in the dictionary are present. All keys apart
+    from 'paths' are required.
     """
 
-    if (type(state) == dict and
-        'abund' in state and
-        'chatter' in state and
-        'cosmo' in state and
-        'xsect' in state and
-        'modelstrings' in state):
+    if type(state) == dict and \
+       'abund' in state and \
+       'chatter' in state and \
+       'cosmo' in state and \
+       'xsect' in state and \
+       'modelstrings' in state:
+
+        h0, q0, l0 = state["cosmo"]
+
         set_xsabund(state["abund"])
         set_xschatter(state["chatter"])
-        set_xscosmo(state["cosmo"][0], state["cosmo"][1], state["cosmo"][2])
+        set_xscosmo(h0, q0, l0)
         set_xsxsect(state["xsect"])
         for name in state["modelstrings"].keys():
             set_xsxset(name, state["modelstrings"][name])
+
+        # This is optional to support re-loading state information
+        # from a version of XSPEC which did not provide the path
+        # information.
+        #
+        try:
+            managerpath = state['paths']['manager']
+        except KeyError:
+            managerpath = None
+
+        if managerpath is not None:
+            set_xspath_manager(managerpath)
 
 
 def read_xstable_model(modelname, filename):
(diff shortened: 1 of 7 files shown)


DougBurke:
Darn it; I think the travis changes are going to have to be more extensive. I should look at #389 to see if this helps in setting the library names (as they can depend on the XSPEC version).

olaurino:
> One test is marked as "expected fail" as it is not clear why it returns
> a chatter setting of 50 rather than 0.

It could be because the chatter level is set to 50 in the tests setup, https://github.com/sherpa/sherpa/blob/2d0b5272e2cdc07ef62c9e914029ebb336ac34ca/helpers/test.py#L25 so that if something fails we get to see the logs. It was helpful to track down some xspec issues when the tests failed often on Travis, it hasn't been that helpful lately, so we can revise that.

DougBurke:
Aha. That would explain it. I have been pulling my hair out trying to think how the initialization code could fail to be fired!

As I mention in the comments for some of these tests, they ideally would be run straight after importing XSPEC, so we can test the defaults, but that is not really realistic. At some level checking the default level is not as important as checking we can get/set the values.

Or we could think about having the requires_xspec decorator set the chatter to 50 when calling the function, resetting it - and then having a specialized version that doesn't do this just for this test (a random idea; may well be better options out there if we do want to test the default value).

olaurino:
rebasing, was 4fbc314. @DougBurke it wouldn't be bad if you could double check the conflict resolutions I made.

olaurino:
One issue I had locally is that some new tests exercising the path functions fail when `$HEADAS` ends with `/`.

olaurino:
@DougBurke There are a number of things you mention you would have done differently but had issues, plus the opportunity to use #395. I am going to assume I can review this PR as is and we can improve things if needed/possible.

sherpa/astro/xspec/__init__.py
```diff
@@ -160,6 +163,40 @@ def set_xsxset(name, value):
...
+    get_xspath_manager : Return the path to the files describing the XSPEC models.
+
+    Examples
+    --------
+    >>> set_xspath_manager('/data/xspec/spectral/manager')
+    """
+
+    _xspec.set_xspath_manager(path)
+    spath = get_xspath_manager()
+    if spath != path:
```
olaurino: Hmm... when I saw there were no tests for the error condition I tried to set this to the string '%$', and I got the same string from `get_xspath_manager`, so it looks like this error condition is redundant? Or is it really possible it's going to fail?

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -297,6 +297,13 @@ PyObject* xspecmodelfct( PyObject* self, PyObject* args )
           XSpecFunc( &fear[0], &npts, &pars[0], &ifl,
                      &result[0], &error[0] );
 
+	} catch(std::exception& exc) {
+
+          std::ostringstream err;
+          err << "XSPEC model evaluation failed: " << exc.what();
+          PyErr_SetString( PyExc_ValueError, err.str().c_str() );
+          return NULL;
+
```
olaurino: Interesting. Any way we can test this/these?

DougBurke: The only way I can think of would be to build - as part of the test - an xspec model which we can trigger a failure in certain conditions (e.g. have a parameter as 0 or 1 and if 1 then it throws an exception). Otherwise we could try and look through xspec models to see if any do raise an exception, and then see of there's a way to trigger that condition in the test (I'm thinking of changing the path to the data files, since a model may fail then). However, that's not robust (since it depends on internals of the XSPEC model library),

sherpa/astro/xspec/__init__.py
DougBurke: I honestly don't know. You'd have to dig into the XSPEC model library code to see if this can happen. I'm just trying to be a bit defensive here, but maybe it's not worth it?

sherpa/include/sherpa/astro/xspec_extension.hh
DougBurke: From looking at the commit message I see that I added this because it helped diagnose a problem with a XSPEC user model!

DougBurke: Summary of the following: it's not obvious that this condition can actually be triggered, but there is no explicit statement on this from the XSPEC docs (so we should probably ask them).

Hmmm, I think my comment about a user model must have referred to my work on building XSPEC user models which (currently) does not use this template code - i.e. I saw the error and had to add a similar throw/catch block in that code, so hence added it here also just in case. My attempts to trigger a similar situation with XSPEC models has been less fruitful. I looked through the models for ones with an obvious raise and found the laor model (amongst others). I can trigger the condition, but it's obviously getting caught/handled within the XSPEC interface itself. As an example, using CIAO (but the same behavior is seen on this branch with XSPEC 12.9.1p):

a) default set up

```
% sherpa
CIAO 4.9 Sherpa version 1 Friday, December 2, 2016

sherpa-1> from sherpa.astro import xspec
sherpa-2> xspec.get_xsversion()
          '12.9.0o'
sherpa-3> xspec.set_xschatter(50)
sherpa-4> xslaor.mdl()
          <XSlaor model instance 'xslaor.mdl'>
sherpa-5> mdl([1.1, 1.11, 1.12])
          array([ 0.53513175,  0.46486822,  0.        ])
```

b) trigger a failure in the laor model

The model loads in data files which it gets from XSPEC_MDATA_DIR env var, if set, otherwise it is based on HEADAS. So setting XSPEC_MDATA_DIR to an invalid path triggers the error, which is converted to a message printed to stderr (I believe), and then exits. Our code makes sure that the return array in this case is all 0's.

```
% export XSPEC_MDATA_DIR /dev/null/
% sherpa
CIAO 4.9 Sherpa version 1 Friday, December 2, 2016

sherpa-1> from sherpa.astro import xspec
sherpa-2> xspec.get_xsversion()
          '12.9.0o'
sherpa-3> xspec.set_xschatter(50)
sherpa-4> xslaor.mdl()
          <XSlaor model instance 'xslaor.mdl'>
sherpa-5> mdl([1.1, 1.11, 1.12])

***XSPEC Error: In RDLAOR, FITSIO error = 104
Failed to open /dev/null/ari.mod
          array([ 0.,  0.,  0.])
```

olaurino:
How about we remove the stuff we don't know how to test so we can keep reviewing this and we add the extra stuff if and when we need it?

sherpa/include/sherpa/astro/xspec_extension.hh
DougBurke: A discussion with @xspector says that XSPEC aims to catch all the errors, so this particular check is likely not that useful.

olaurino:
I'd like to merge this. I'd prefer not to add any code that we can't even test, though.

sherpa/include/sherpa/astro/xspec_extension.hh
olaurino: I am confused as to why the exception is not bubbling up. Shouldn't a `ValueError` Python exception be raised if any C++ exception is thrown by the wrapper xspec function? Unless I am missing something that is not the case. Or am I missing something?

DougBurke: I believe that there is some "hidden" XSPEC wrapper code that lies between the implementation (that is, the function that is pointed to in the `model.dat` file) and the function name we call, and this catches C++ exceptions (or at least any of the Yellow/RedAlert exceptions that they define). So we never see the exception. Or I don't understand C++ exceptions well enough, which is a perhaps-more-realistic scenario.

The error message I show above appears to come from this bit of code from `XSFunctions/xslaor.cxx`:

```
   std::ostringstream errMsg;
   errMsg << "In RDLAOR, FITSIO error = ";
   fitsfile *fptr=0;

   // Open the ari.mod file
   fits_open_file(&fptr,cFileName, 0, &status);
   if (status)
   {
      errMsg << status << "\nFailed to open " << modFileName << "\n";
      throw YellowAlert(errMsg.str());
   }
```

A quick look at `XSUtil/Error/Error.cxx` has

```
YellowAlert::YellowAlert()
{
}

YellowAlert::YellowAlert (const string& message)
{
  *IosHolder::errHolder() << "\n***XSPEC Error: " << message << std::flush;
}
```

I'm not sure if this helps you.

DougBurke: Here's the info I got from Craig Gordon when I asked if they had a policy on exception handling:

> I can't recall any cases where exceptions currently propagate entirely out of the model functions library, but we've never established a rule guaranteeing that that will always be the case.  So unless maintaining those try/catch blocks is problematic, I recommend keeping them.  If they're at a high enough level of the code, I wouldn't think it would affect performance.
>
> Just for your future reference, all Xspec exceptions inherit from either YellowAlert or RedAlert classes (defined in src/XSUtil/Error/Error.h).  RedAlerts basically perform the role of abort() calls: something so bad has happened that Xspec should terminate immediately.  Therefore in our high level try/catch blocks, we often just catch YellowAlert& while letting RedAlerts go uncaught to end the program.  Of course if there is something you want to do even for RedAlerts, you would do "catch (...)" as in your current code.

olaurino: > Our code makes sure that the return array in this case is all 0's.

I probably misunderstood this line in one of your previous comments. I thought you were claiming that our catch code was responsible for returning a zero array, so that it was called. If that's not the case then I am not confused anymore, because we have enough evidence to believe that the catch code is never ever called. Right?

DougBurke: Our code zeroes out the return array before calling the XSPEC model. So if the XSPEC model doesn't do anything (presumably because of an error condition) then we will return all 0's. This was done because we used to just send in an array of uninitialized memory, which meant that if there was an error we would get random values out.

If the XSPEC model has changed some of the values before the error then we'll see those changes, but I don't think it's worth worrying about that case.

Does this help you? At present I think the catch code I added could be removed if you feel it is potentially unsafe. If we did remove it we wouldn't be in a worse case than we are know (i.e. in the released code), and we can worry about other PRs.

sherpa/astro/xspec/__init__.py
olaurino: I resisted the urge to remove this check we can't test.

olaurino:
Unless I hear otherwise from @DougBurke, if and when Travis gives us the green light I'll merge this (I'll probably need to rebase this, I forgot).

olaurino:
Rebasing (was 39c531d).

PR #396, opened 2017-09-10
fix set_xlog and related commands using Python 3 (fix #393)

### LINKED ISSUE #393: set_xlog/ylog fail with Python 3
The `sherpa.astro.ui.set_xlog` and `set_ylog` commands fail with Python 3.5 but not with Python 2.7. The following is using the CIAO 4.9 version of Sherpa but I expect the same behavior with the master branch. I have not audited the code to see if there are other functions which fail in a similar manner - such as `set_xlinear` - but I expect them to.

A) Python 2.7

```
sherpa-1> get_data_plot_prefs()['xlog']
          False
sherpa-2> get_data_plot_prefs()['ylog']
          False
sherpa-3> set_xlog()
sherpa-4> set_ylog()
sherpa-5> get_data_plot_prefs()['xlog']
          True
sherpa-6> get_data_plot_prefs()['ylog']
          True
sherpa-7> sys.version
          '2.7.11 (default, Apr 27 2016, 13:22:12) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)]'
sherpa-8> sherpa._version.get_versions()
           {'full': '10736c4d70fcd3e7d686670e1594103a1819098e', 'version': 'ciao4.9'}
```

B) Python 3.5

```
sherpa-1> get_data_plot_prefs()['xlog']
          False
sherpa-2> get_data_plot_prefs()['ylog']
          False
sherpa-3> set_xlog()
TypeError: 'dict_keys' object is not subscriptable
sherpa-4> set_ylog()
TypeError: 'dict_keys' object is not subscriptable
sherpa-5> get_data_plot_prefs()['xlog']
          False
sherpa-6> get_data_plot_prefs()['ylog']
          False
sherpa-7> sys.version
          '3.5.1 (default, Apr  6 2016, 12:14:18) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)]'
sherpa-8> sherpa._version.get_versions()
          {'full': '10736c4d70fcd3e7d686670e1594103a1819098e', 'version': 'ciao4.9'}
```

DougBurke:
# Release notes

Fix use of commands that set the plot state using the `sherpa.astro.ui` module, such as `set_xlog` and `set_xlinear`, when using Python 3.

# Details

The fix is very easy. The complication comes in how best to test this. Tests were added for #303 to the session object, but they only tested the `sherpa.ui` version, not the `sherpa.astro.ui` version. So I have added a copy of these tests updated to use the `astro` session module. This is obviously not-a-good-idea™ since it relies on manually-duplicating tests. There are also some existing tests of the astro session object in `test_astro_session.py` that should be merged with these tests (just to make it clearer what is being tested).


DougBurke:
Hmmm, this is an interesting failure:

https://travis-ci.org/sherpa/sherpa/jobs/273863209#L848

Do the test file names have to be unique - e.g. we can not have two `test_session.py` files in different directories?

PR #399, opened 2017-09-11
Python 3 fixes and new tests for pha2 datasets

DougBurke:
# Apology

It's a bit of a mis-match of code - I originally wanted to add a PHA2 data set as a test case, and then noticed Python 3 failures, and it spiraled out of control from there.

# Release Notes

Fixes several problems when using Sherpa with Python 3: failures when calling `list_bkg_ids` and `list_response_ids` and the parameter querying after `paramprompt(True)` has been called.

There is also a change to avoid a `FutureWarning` being raised in the astropy backend when reading in PHA2 data files.

Tests have been added to test the reading of a PHA2 format dataset, and new files added to the `sherpa-test-data` repository for this purpose.

# Background

Note: this commit requires an update to the `sherpa-test-data` repository - see sherpa/sherpa-test-data#12

This is intended to provide some tests and fixes for #397 but for that I need a PHA2 file, and we don't have one, so the first order of business is to add one.

Add a PHA2 file to the test suite, courtesy of TGCat, and then perform some basic read and evaluation tests, comparing the results to those from reading in the same data from PHA1 files.

This increases test coverage of several modules by 1 to 2 %.

These tests uncovered a `FutureWarning` due to the use of `if None in ...` which I have changed to `if ... is None`, as I think that makes sense in this context. I don't understand why I didn't see this in my local builds.

Finally added in tests of list_bkg_ids and list_response_ids which fail with Python 3, and added a fix for these.

In checking the .keys fixes (looking for other possible cases) I discovered that paramprompt fails on Python 3, so have added tests and a fix for this (let's see if Travis agrees).

I plan to look at the coverage info to see if any other tests could be added here. I also plan to add some checks of a range of statistics as regression tests, to improve confidence that things are being handled correctly for the multiple-response case for things like the areascal handling, and wstat statistic. Actually, I've shelved this plan for now.


DougBurke:
Argh, these tests are failing because I used `unittest.mock`. Don't have time to dig into this just now.

DougBurke:
I expect these to fail due to #406

DougBurke:
About to rebase. Original tip was f636ec8.

DougBurke:
This needs #408 merged since the failures look to be due to #406.

DougBurke:
About to rebase. Original tip was 4da0541

olaurino:
Rebasing to fix conflict: original tip was 621743d

sherpa/astro/io/pyfits_backend.py
```diff
@@ -929,7 +929,17 @@ def get_pha_data(arg, make_copy=False, use_background=False):
...
             counts = _try_vec(hdu, 'COUNTS', num, fix_type=True)
-            if None in counts:
+            # This was
+            #   if None in counts:
+            # but that caused a FutureWarning about
+            # 'comparison to `None` will result in an elementwise
+            #  object comparison in the future.'
+            # so it has been changed to the following, which I think
+            # is the intended meaning. If the counts array does contain
+            # "None" values then the data isn't a valid OGIP file
```
olaurino: I am not sure not being valid OGIP is necessarily a good reason for changing the semantics of the check. However, in this case I don't think the problem is OGIP-compliance, as the `_try_vec` function can itself return an array of `None`s in some cases, and it doesn't seem like it can return a `None` unless pycrates or astropy do, which I am not sure is a possible case (NaNs maybe, but `None`s?). For better or worse I think we should keep what appears to be the original semantics of the check, i.e. whether any elements in `counts` are `None`s. I'll see if I can add a test case for this.

olaurino: I mean, we could also change the implementation of `_try_vec` not to return an array of `None`s, of course.

olaurino: Hmm... apparently I has already fixed at least one of these comparisons: https://github.com/sherpa/sherpa/commit/61fbfcdc16fd9f4bdc599e5f2c2db41560abf82a

sherpa/astro/ui/tests/test_astro_supports_pha2.py
```diff
@@ -0,0 +1,363 @@
...
+        assert b.name == infile
+
+        b = ui.get_bkg(i, bkg_id=2)
+        validate_pha(b, bkg=False)
+        assert b.name == infile
+
+    ui.clean()
+
+
+# TODO: how to test what messages are logged when load_pha is
```
olaurino: I can probably add this.

sherpa/astro/ui/tests/test_astro_supports_pha2.py
```diff
@@ -0,0 +1,363 @@
...
+    assert s9 == sm1
+    assert s10 == sp1
+
+    # The values were calculated using CIAO 4.9, Linux64, with
+    # Python 3.5.
+    #
+    assert s9 == pytest.approx(1005.4378559390879)
+    assert s10 == pytest.approx(1119.980439489647)
+
+    ui.clean()
```
olaurino: I am a little concerned by these `clean` statements, which are not executed unless the tests pass. I'd rather create a fixture for it, to make sure `clean` is called before and after a test, whether or not it passes. I have been thinking about an automatic fixture to run before and after every test, but I don't want to stir the pot too much, so maybe for now it is enough to just include the fixture when needed.

sherpa/ui/utils.py
```diff
@@ -5630,60 +5614,59 @@ def get_model(self, id=None):
...
+                    except Exception as e:
+                        info("Please provide a float value; " + str(e))
+                        continue
+
+                try:
+                    self.set_par(par, val, minval, maxval)
+                    break
+                except Exception as e:
+                    info(str(e))
+                    continue
```
olaurino: I have a strong urge to refactor this code. It's probably not worth the effort, but that's what we say every time we see code like this... maybe we should start building some muscle memory for this kind of refactorings. @DougBurke what do you think?

sherpa/astro/io/pyfits_backend.py
olaurino: We don't have any tests exercising the "there are `None`s in the array". I would like to add one, but it looks like we either need a complex stub of the backend or a data file without the counts. The first option is better on the long run (you could "mock" a file for test purposes without having to actually have a physical file) but it would probably take a while to implement. I'll keep working on the other items, then I'll evaluate when I am done if it makes sense to add more improvements.

DougBurke: Just to comment on your first comment: this happens in the routine `get_pha_data`, which is designed to read OGIP-standard files, so I think it's okay to assume we have a valid OGIP file here.

OGIP files generally have a COUNTS column, but can occasionally have a RATE column. You can see the same logic in the PHA-1 logic just above at  https://github.com/DougBurke/sherpa/blob/69950b2aec6e3e8526942c6814c574f15a5eaa5d/sherpa/astro/io/pyfits_backend.py#L870 which uses "if blah is None" and not "if None in blah".

So, a None returned by `_try_vec` indicates that the column doesn't exist (or perhaps there was a problem reading the data somehow, but I'm too lazy to read the code for `try_vec` just now). Actually, I've just looked at `_try_vec` and I see what you mean. I think there is a potential divergence here, and we should add a PHA-1 file with a RATE column rather than a COUNTS one as I'm not sure it'll work sensibly.

I am not convinced that `_try_vec` should return an array of None values if the column does not exist, but we'd need to audit the module to see how it is called to make that call.

sherpa/ui/utils.py
DougBurke: Well, as my code is a minor refactoring of the existing code, I can't really complain about it. I tried to make mine a minimal change to the old logic (to make it easier to follow), coupled with some test-coverage related changes (to make sure all the code was tested).

Given that before this test we don't have any tests of this logic, my suggestion would be to hold off refactoring this code in this commit. Then we have some test code that we have some confidence in to test out your refactoring. Maybe it would be worth removing my refactoring of this routine from this commit to make this a bit more obvious?

olaurino: Nah, I am OK with your changes, not with the underlying code. However, I am spending more time that I should on trying to lay out some work for important things like testing plots (well, not today) and testing log output. In any case I checked the test coverage and some of the error cases, as I think you mentioned in the comments, are not exercised, so at the very least I would try and add those tests in. Thing is, if we don't refactor it now we'll never refactor it as we'll be caught in other stuff.

sherpa/astro/ui/tests/test_astro_supports_pha2.py
DougBurke: I have a simple fixture that does this in https://github.com/sherpa/sherpa/pull/412/commits/35c228774669ec932b92da35f4cd73ebe8487fb2 - we might want two versions, or be able to parametrize whether it's sherpa.ui or sherpa.astro.ui being used.

However, for these particular tests, I wonder if we should just change to creating a session object within the test and use that. I'm using the session layer both as a test of the high-level functionality (i.e. that it all works together), and because it's quicker to write the test at this level then have  a bunch of tests that check whether a particular class (e.g. Fit or Stat or Model or ..) works here, since really we should rely on the tests for those components.

sherpa/astro/ui/tests/test_astro_supports_pha2.py
```diff
@@ -39,6 +39,16 @@
 from sherpa.astro.data import DataPHA
 
 
+@pytest.fixture(autouse=True)
```
DougBurke: I have an alternate version in https://github.com/sherpa/sherpa/pull/412/commits/35c228774669ec932b92da35f4cd73ebe8487fb2 which doesn't require request. I don't understand if one is better than the other, or whether there's any difference in behavior or what versions they work in or ...

olaurino: From the `pytest` documentation:

https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code
> Note that if an exception happens during the setup code (before the yield keyword), the teardown code (after the yield) will not be called.
> An alternative option for executing teardown code is to make use of the addfinalizer method of the request-context object to register finalization functions.

That's why I prefer to be explicit with the finalizer, although in this case the likelihood of `ui.clean()` throwing an exception is rather low, and anyway in that case not executing `ui.clean()` during tear down doesn't seem to make any difference.

olaurino: On the other hand, I think this fixture should probably be autoused for all tests, and while I have the temptation to make it a global fixture, we should probably make it `autouse=True` at least for new tests. That also prevents test functions from having a dangling, unused parameter.

olaurino:
By the way, if we approve this I need to file an OTS request for CIAO for the `pytest-catchlog` plugin. Alternatively we can decide we don't want to test the log messages output.

sherpa/astro/ui/tests/test_astro_supports_pha2.py
DougBurke: One issue with `autouse=True` is that there are plenty of tests which don't use the session interface, so it's just extra work. Then again, it shouldn't be a significant run-time cost. I would also suggest in this case that it calls both `sherpa.ui.clean()` and `sherpa.astro.ui.clean()` - since they refer to different session objects. I don't think this needs to be done or decided upon to accept this PR.

sherpa/__init__.py
```diff
@@ -169,20 +169,32 @@ def install(package_name):
...
+
+
+def clitest():
+    plugins = _install_test_deps()
     import pytest
     import os
     sherpa_dir = os.path.dirname(__file__)
-
-    errno = pytest.main([sherpa_dir, '-rs'])
+    errno = pytest.main([sherpa_dir, '-rs'], plugins=plugins)
```
DougBurke: So, `plugins` is the list of packages that had to be installed whilst running the tests (i.e. that weren't already installed). So presumably `pytest` wants to know about these additions "for reasons"?

olaurino: Yeah, I meant to add comments when I finally figured it out, then I forgot.

olaurino: Just pushed a commit with comments clarifying this point.

DougBurke: Thanks. It's useful in 6-months time and we don't know what's going on anymore...

DougBurke:
As the screen output for some of these commands can be considered part of the UI, it makes sense to test them. It will at least tell us when something has changed. However, whether the cost of getting the new package into the CIAO build infrastructure is worth it is not my call (as I wouldn't be the one doing it :-). It might be useful (for that process) to know if there are other parts of the system where we might add extra checks of the logging output (ie is this PR going to be the only user of it, or do we think it can be used elsewhere)?

olaurino:
Don't worry about CIAO, I will :) I just need you to sign off, from your side of the fence, on the inclusion of the plugin for Sherpa testing for the sake of Sherpa testing. I already checked that it's feasible to add the extra plugin to the lot, if decide we actually need it.

That being said, we can definitely start using the plugin elsewhere. I noticed in some old tests that there is actually a way to test logging without using the plugin, but it's way more cumbersome than with the plugin. I vote for its inclusion :)

olaurino:
It turns out `pytest-catchlog` was merged into the `pytest` core package and released with it in v3.3.0, which is still not available in the `conda` default channel.

If I install `pytest` 3.3.1 using `pip`, then `sherpa_test` and `python setup.py test` work fine, even though `sherpa_test` keeps installing the plugin as a `pip` package. As a matter of fact, `pytest` itself reports it as a warning:
> pytest-catchlog plugin has been merged into the core, please remove it from your requirements.

I would leave the code as is so it works with older and newer versions of `pytest`, in particular the one available currently in `conda`. On the other hand, on the CIAO side it might just be necessary to upgrade `pytest` to the latest version rather than requesting the new plugin.

Issue #406, opened 2017-10-26
possible new warning to ignore: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88


DougBurke:
With a clean build of master branch (except for updating config to build XSPEC and to automatically print any warnings that are caught, since I can no longer get the --show-warnings flag to work), I get a failure in the datastack tests because of the following

```
--------------------------- Captured stdout teardown ---------------------------
*** Warnings created: 2
1/2 {message : RuntimeWarning('numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88',), category : 'RuntimeWarning', filename : '/home/djburke/miniconda2/envs/sherpa-rmf-nobins/lib/python3.5/importlib/_bootstrap.py', lineno : 222, line : None}
2/2 {message : RuntimeWarning('numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88',), category : 'RuntimeWarning', filename : '/home/djburke/miniconda2/envs/sherpa-rmf-nobins/lib/python3.5/importlib/_bootstrap.py', lineno : 222, line : None}
```

This appears to be a "known conda issue" - e.g. see https://github.com/ContinuumIO/anaconda-issues/issues/6678 - so we may want to add this to our skipped warnings.

DougBurke:
This is causing travis tests to fail - e.g. https://travis-ci.org/sherpa/sherpa/jobs/293414606

PR #407, opened 2017-10-26
Ensure 0-length array is an error in filter_resp (fix #405)

### LINKED ISSUE #405: RMF code doesn't handle "no channels" well; invalid read in C++ scope
The code

```
from sherpa.astro.ui import unpack_rmf
rmf = unpack_rmf('3c273.rmf')
rmf.notice([])
```

leads to an invalid memory read in C++ scope - e.g. valgrind reports with Sherpa in CIAO 4.9

```
==9652== Invalid read of size 4
==9652==    at 0xF2172D7: is_in<unsigned int, long int, unsigned int> (utils.hh:182)
==9652==    by 0xF2172D7: int sherpa::astro::utils::_filter_resp<unsigned int, unsigned int, double, double, unsigned char, long>(unsigned int const*, long, unsigned int const*, long, unsigned int const*, long, unsigned int const*, double const*, long, unsigned int, std::vector<unsigned int, std::allocator<unsigned int> >&, std::vector<unsigned int, std::allocator<unsigned int> >&, std::vector<unsigned int, std::allocator<unsigned int> >&, std::vector<double, std::allocator<double> >&, unsigned char*) (utils.hh:269)
==9652==    by 0xF217B8F: _object* sherpa::astro::utils::filter_resp<sherpa::Array<double, 12>, sherpa::Array<unsigned int, 6>, unsigned int, double>(_object*, _object*) (_utils.cc:269)
```

This is because there is no explicit check that the array sent to the `notice` method is not empty in C++ scope.

This can lead to the `rmf.notice` call returning either all `True` or all `False` (should be all `False`).

DougBurke:
# Release Note
Add an explicit check in the C++ filter_resp code to error out if the noticed channels array is empty.

# Notes
This fixes #405.

This is currently the minimal change needed to avoid reading from uninitialized memory with a RMF. I have not audited the other C++ code to see if similar checks should be added. The behavior of the higher-level code (i.e. the Python objects) is up for debate.

# Details

It is not yet clear to me whether we should error out when asking to filter a RMF with an empty channel array, but it seems to fit into other parts of how Sherpa deals with this. Note that it might make sense for the low-level code (such as that changed here) to error out, but the RMF code to explicitly check for the "no channel" case and to handle it separately (e.g. `sherpa.astro.data.DataRMF.ignore`).

### DIFF
--- a/sherpa/astro/utils/src/_utils.cc
+++ b/sherpa/astro/utils/src/_utils.cc
@@ -1,5 +1,5 @@
 // 
-//  Copyright (C) 2009  Smithsonian Astrophysical Observatory
+//  Copyright (C) 2009, 2017  Smithsonian Astrophysical Observatory
 //
 //
 //  This program is free software; you can redistribute it and/or modify
@@ -252,6 +252,12 @@ namespace sherpa { namespace astro { namespace utils {
 			    &matrix,
 			    &offset) )
       return NULL;
+
+    if (0 == noticed_chans.get_size()) {
+      PyErr_SetString( PyExc_ValueError,
+		       (char*)"There are no noticed channels" );
+      return NULL;
+    }
 
     if( EXIT_SUCCESS != mask.zeros( 1, n_grp.get_dims() ))
       return NULL;
--- a/sherpa/astro/utils/tests/test_astro_utils_unit.py
+++ b/sherpa/astro/utils/tests/test_astro_utils_unit.py
@@ -25,7 +25,6 @@ from sherpa.utils import requires_data, requires_fits
 
 
 # See https://github.com/sherpa/sherpa/issues/405
-@pytest.mark.xfail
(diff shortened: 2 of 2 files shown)


DougBurke:
I  claim these failures are due to #406 (although only looked at one such failing case).

DougBurke:
About to rebase, to pick up #408. Tip was 830d7d3

DougBurke:
About to rebase. Tip was 595b08b (although the last two are cherry-picks, so really interested in a262479).

olaurino:
Rebasing. Was 5d04230

olaurino:
@DougBurke this was marked as WIP, but I think it can be merged, right?

DougBurke:
I think I'd left it marked WIP since there are other things that could be looked at because of this (e.g. how should the UI layer handle this case), but I think it's fine to merge this PR.

PR #408, opened 2017-10-27
Ignore binary-incompatibility warnings seen in anaconda environments (fix #406)

DougBurke:
# Summary

These commits only affect the testing code.

Add a new warning to our white list to allow tests to pass on Travis-CI and when using anaconda environments. See https://github.com/ContinuumIO/anaconda-issues/issues/6678 for a discussion of the reason for the warning

    numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88

The --show-warnings option for the tests has been removed since it was found to be unstable - potentially depending on the package versions installed - and not actually that useful. The contents of any unexpected warnings are now displayed at the end of the test, as this was found to be more useful anyway. To avoid excess output, repeated warnings are only shown once.

PR #409, opened 2017-10-31
EmissionGaussian model when skew parameter is not unity (fix #403)

### LINKED ISSUE #403: problem with skewed sherpa.astro.optical.EmissionGaussian model
Via the CSC help desk.

The `sherpa.astro.optical.EmissionGaussian` model has problems when skew is not set to 1.0:

a) the documentation is wrong: `d2(x) = d(x) / skew` not `d(x) / s` when `x > pos`

b) the code is wrong: `left = (arg <= pos)` should be - thanks to @kglotfelty - `left = (x <= pos)`, which is only used when `skew` is not 1.

It's a small, simple fix; more effort is needed to add/validate the tests and check against the "specification": http://specview.stsci.edu/javahelp/Components.html#Emission_Gaussian

DougBurke:
# Summary

Fix the `sherpa.astro.optical.EmissionGaussian` code for the case when the `skew` parameter is not unity. The documentation has also been updated.

# Details

The new tests are more "property based" than functional, in that it's not obvious what the correct values should be (I couldn't see any unit tests for specview from a quick look). The new tests check expected behavior - e.g. the peak is at the center of the gaussian, and that skewed values are larger on x>pos (skew>1) or x<pos (skew < 1).

I have cherry-picked my commits from #408 to try and get a clean travis run.

### DIFF
--- a/sherpa/astro/optical/__init__.py
+++ b/sherpa/astro/optical/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2016  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2011, 2016, 2017  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


DougBurke:
I'll rebase this at some point soon now that #408 as been merged.

olaurino:
Rebasing. Original tip f4ae3ed

DougBurke:
I think we should have a quick discussion about the testing of this change - i.e. is it sufficient - before merging.

olaurino:
The only other thing we could do I can think of is to run the original Java model on a fake dataset and see what the results are, but it might be overkill.

Or maybe create an ascii file that exactly reproduces the python model on an arbitrary grid with some parameters and let specview fit it, and make sure the parameters are the same as the original model.

Other ideas?

PR #412, opened 2017-12-02
support ROSAT PSPC (and similar) RMF files with AstroPy 

DougBurke:
# Release Notes

Add explicit tests for reading in, and using, ROSAT PSPC PHA and RMF files. Fix a bug in the AstroPy back-end handling of the ROSAT RMF file.

# Other notes

Hidden away now in this PR is a commit that adds support for a Travis-CI build that uses the CIAO 4.9 Crates module (Python 2.7 only), to provide immediate feedback for changes that might affect the Crates backend. There is no support for XSPEC or ChIPS in this build. This was used to verify that the changes didn't break anything obvious. This commit has been backed out for now, but could be resurrected at a future date (it moved the logic for Travis testing into external scripts rather than all being encoded in the `.travis.yml` file, which could in itself be a useful update even if the CIAO-specific options are not supported).

# Details

The logic for creating the test environment has been moved out of the Travis yaml file and into scripts in the `scripts/` directory. I like the idea, as it potentially makes it easier to re-create a particular Travis test (e.g. if one of them is failing) locally. The CIAO build is fragile - in the sense that it relies on CIAO 4.9 which requires kludges to support working around some installation issues. It could be separated from the bug fix, but was useful to help me make sure I did not break anything on the Crates side (in particular in setting up useful tests).

This PR can be thought of as an follow-up to #358, in that the code it corrects the code added in that commit to handle the more-general case (i.e. the Swift RMF format is "simpler" than the ROSAT PSPC one).

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -376,7 +376,7 @@ class DataRMF(DataOgipResponse):
                  n_chan, matrix, offset=1, e_min=None, e_max=None,
                  header=None, ethresh=None):
 
-        energ_lo, energ_hi = self._validate_energy_ranges(name, energ_lo, energ_hi, ethresh)
+        energ_lo, energ_hi = self._validate(name, energ_lo, energ_hi, ethresh)
 
         self._fch = f_chan
         self._nch = n_chan
@@ -402,6 +402,19 @@ class DataRMF(DataOgipResponse):
         if 'header' not in state:
             self.header = None
         self.__dict__.update(state)
+
+    def _validate(self, name, energy_lo, energy_hi, ethresh):
+        """
+        Default validation: do nothing and just pass the arrays through.
+        Subclasses may override this method to perform different validations.
+
+        :param name: The name/label of the current file
+        :param energy_lo: The lower bounds of the energy bins
+        :param energy_hi: The upper bounds of the energy bins
+        :param ethresh: The lowest energy value, if any
+        :return: A tuple of energy bounds arrays, validated and possibly adjusted
+        """
+        return self._validate_energy_ranges(name, energy_lo, energy_hi, ethresh)
 
     def apply_rmf(self, src, *args, **kwargs):
         "Fold the source array src through the RMF and return the result"
@@ -452,6 +465,16 @@ class DataRMF(DataOgipResponse):
 
     def get_ylabel(self):
         return 'Counts'
+
+
+# FIXME There are places in the code that explicitly check if an object is an instance of sherpa.astro.data.DataRMF.
+# So it's safer to make DataRosatRMF a subclass of the default class, although in principle they should be siblings
+# and subclasses of the same superclass.
+class DataRosatRMF(DataRMF):
+    ui_name = "ROSAT RMF"
+
+    def _validate(self, name, energy_lo, energy_hi, ethresh):
+        return energy_lo, energy_hi
 
 
 class DataPHA(Data1DInt):
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -405,14 +405,23 @@ class DataRMF(DataOgipResponse):
 
     def _validate(self, name, energy_lo, energy_hi, ethresh):
         """
-        Default validation: do nothing and just pass the arrays through.
-        Subclasses may override this method to perform different validations.
+        Validate energy ranges and, if necessary, make adjustments.
+        Subclasses may override this method to perform different validations
+        or skip validation altogether.
 
-        :param name: The name/label of the current file
-        :param energy_lo: The lower bounds of the energy bins
-        :param energy_hi: The upper bounds of the energy bins
-        :param ethresh: The lowest energy value, if any
-        :return: A tuple of energy bounds arrays, validated and possibly adjusted
+        Parameters
+        ----------
+        name : str
+            The name/label of the current file
+        energy_lo, energ_hi : NumPy array
+            The lower bounds of the energy bins. The arrays must have the same size
+        ethresh : float
+            The lowest energy value
+
+        Returns
+        -------
+        energy_lo, energ_hi : NumPy array
+            The energy values to use for the bin boundaries
         """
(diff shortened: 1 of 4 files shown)


DougBurke:
This came out of @olaurino code review of #399 - in particular my comment at https://github.com/sherpa/sherpa/pull/399#discussion_r154169174

I felt it worth exploring the rate behavior, and it's probably easier to do so in a separate PR at the moment.

DougBurke:
So, turns out that crates is throwing the error when reading the RMF file. I'll report that internally.

DougBurke:
Well, that was a spectacular success. Perhaps the build logic should be factored out into scripts so that they can be tested separately, and make .travis.yml cleaner?

DougBurke:
About to rebase to deal with change in `sherpa-test-data`. Original tip was cb493a3

DougBurke:
About to rebase to make it look like I knew what I was doing all along. Tip was 88665e4

DougBurke:
About to rebase to simplify the commits. Original tip was 6634dfd

DougBurke:
About to rebase. Original tip was a572d55

DougBurke:
@olaurino - can I remove the pr:hold label from this, or are there still build conflicts?

olaurino:
I don't really know. Remove it and if something comes up we'll address that.

olaurino:
Yeah, unfortunately it must be conflicting with something else, so it doesn't get merged into the snapshot branch. Most likely the issue is with PR 419 since the conflict is in the travis configuration file. Which reminds me I owe you a reply in that thread.

DougBurke:
About to rebase. Original tip was 613a261

olaurino:
@DougBurke I'd be in favor of removing the new travis job and taking the fix. As for moving some of the logic out of the .travis.yml I never thought the setup part was complex enough to do that, but I'd be in favor of that as well.

DougBurke:
@olaurino can you cherry pick 661cd4b into a new PR. I'd split this into two commits to make it easy to do so, but don't have much time this week for Sherpa

olaurino:
It seems to make more sense to reset this to the previous commit. As long as you have a branch pointing to the current tip git shouldn't GC the commit if you want to create a new PR.

olaurino:
Rebasing (was 63d8bce). There, now even GitHub will keep it around forever ;)

DougBurke:
@olaurino - I'm a bit confused here. I've just edited the summary to say "this fixes a bug, but hides away the travis changes", but it seems that now the only things left are the travis changes. What happened to the actual bug fix? Has that already been merged?

olaurino:
Yeah, I think I saved the wrong commit last night, I was just fixing it.

olaurino:
Interesting. It looks like after merging #383 this PR now gets errors because the new test files don't validate.

olaurino:
@DougBurke how does it look?

sherpa/astro/data.py
```diff
@@ -403,6 +403,19 @@ def __setstate__(self, state):
             self.header = None
         self.__dict__.update(state)
 
+    def _validate(self, name, energy_lo, energy_hi, ethresh):
+        """
+        Default validation: do nothing and just pass the arrays through.
```
olaurino: Oops, I left this comment by mistake. Waiting for more comments before fixing.

DougBurke: I don't think this comment is right. It looks to me like the default is to actually do the validation. Or am I reading this wrong?

sherpa/astro/data.py
```diff
@@ -454,6 +467,16 @@ def get_ylabel(self):
         return 'Counts'
 
 
+# FIXME There are places in the code that explicitly check if an object is an instance of sherpa.astro.data.DataRMF.
+# So it's safer to make DataRosatRMF a subclass of the default class, although in principle they should be siblings
+# and subclasses of the same superclass.
+class DataRosatRMF(DataRMF):
```
DougBurke: In principle this is `DataNoValidationRMF`, which is subclassed to `DataRosatRMF`, but I think that's over-engineering for now (the thought being that there could be other cases where we decide to do no validation).

sherpa/astro/data.py
```diff
@@ -403,6 +403,19 @@ def __setstate__(self, state):
...
+        """
+        Default validation: do nothing and just pass the arrays through.
+        Subclasses may override this method to perform different validations.
+
+        :param name: The name/label of the current file
+        :param energy_lo: The lower bounds of the energy bins
+        :param energy_hi: The upper bounds of the energy bins
+        :param ethresh: The lowest energy value, if any
+        :return: A tuple of energy bounds arrays, validated and possibly adjusted
+        """
```
DougBurke: We've been using something like the following (although I am never sure what textual label to use for NumPy array here):

    Parameters
    ----------
    name : str
        The name/label of the current file.
    energ_lo, energ_hi : NumPy array
        The lower and upper bounds of the energy bins. The arrays must have
         the same size
    ethresh : float
        The lowest valid energy value.

    Returns
    -------
    energ_lo, energ_hi : NumPy array
        The energy values to use for the bin boundaries

olaurino: Right. I thought I had already configured my IDE to use the right style. I'll check again.

DougBurke:
The only quibbles I have are over comments and names, but neither are serious.

sherpa/astro/data.py
olaurino: In principle, I agree.

olaurino:
Pushed a new commit fixing the docstring. For the class hierarchy I agree, but we don't need the intermediate class at this point. If we do in the future it will be easy to introduce one without changing the API and without affecting existing code, I think.

olaurino:
@DougBurke the special test that checked for the crates exception is now failing as the upstream bug was fixed. I am going to remove the test, but I wanted to double check we don't need to make simplifications to the code somewhere else. I don't think so, as the tests were added with this same PR, but wanted to double check.

olaurino:
I forgot, there was a test being skipped as it didn't work on crates, I am also going to remove the skip condition.

PR #419, opened 2017-12-14
Update requirements with pytest 3.3.

olaurino:
Release Note
===========
Sherpa now requires `pytest` version 3.3.0 or later for testing. This change only impacts users and developers who want to run the full unit and regression tests suite. New capabilities in `pytest` 3.3.0 include the ability to catch and test log messages, as well as improved support for testing warnings.

Notes
=====
We need pytest 3.3 because it contains capabilities for capturing and testing the logs and warnings. For warnings, versions <3.1 make verified warnings pop up and fail the teardown stage of some tests, while version 3.3.0 embeds the catchlog plugin for catching and testing logs.


DougBurke:
It looks like it would make sense to squash these commits before merging. Or is there a use in keeping them separate?

Also, we need an update to the `README.md` file

DougBurke:
So, the failed runs seem to be

a) an XSPEC failure https://travis-ci.org/sherpa/sherpa/jobs/327000893 

b) an issue with the serialization tests, where the accuracy has changed 

I don't understand why b is suddenly happening now; I assume there's some other change related to this (e.g. an update to the numpy package being used), or some other reason?

olaurino:
The xspec failure seems to be too much of a coincidence, since it only happens in the 12.9.1 build but this PR does not have anything to do with xspec. Also the build for the commit before the README changes was ok: https://travis-ci.org/sherpa/sherpa/builds/326843938

My concern is that this is that something changed in the travis infrastructure that is causing both issues. On one hand we have the option to go back to the old images, but on the other this might be a serious issue, and a very hard one to debug.

My guess is that if I push the master branch to my fork the build, which has already succeeded, will fail. That's what I'll try. In the meantime I also have build issues in CIAO world to fix, as including PR 394 in the snapshot branch is preventing Sherpa to be built in CIAO.

olaurino:
As I expected, the current master branch, which was working fine yesterday, would not pass travis checks now. After barking up several wrong trees I finally think I figured it out. The latest commit on the master branch was built with numpy 1.11 and numpy 1.13, while the new travis jobs have been run with 1.11 and 1.14. I could reproduce the issue locally and confirm that in the same environment simply switching from numpy 1.13 to numpy 1.14 makes the issue appear.

Now, the problem seems to arise when numpy values are printed, which is why the offending test is test_serialize.py and the default parameters for the optimization methods (e.g. `EPSILON`) are numpy values. numpy 1.14 has a new algorithm for the representation of floating point values(see: https://github.com/numpy/numpy/blob/master/doc/release/1.14.0-notes.rst#float-printing-now-uses-dragon4-algorithm-for-shortest-decimal-representation).

I think I can set a numpy printing options that prints out the "legacy" values and have the tests pass. It seems to work locally, but I'll need to push the change to Travis and make sure it works. Also, I think I need to update Travis so it runs numpy 1.13 on at least one job.

The problem however is that with numpy 1.14 the xspec tests are segfaulting. I think it's because numpy is bringing in libgfortran-ng as a conda dependency, since if I force its removal and rebuild the xspec extension the tests pass. I'll just have travis run the xspec jobs with numpy 1.13 and open a new issue we'll have to address at some point.

Of course the question is whether we should update the serialization code to keep up with numpy (but what about backward compatibility?), but in the meantime we need to make sure the tests keep passing.

olaurino:
@DougBurke this is ready (again) to be reviewed and possibly merged. Since it fixes the Travis jobs this needs to be merged before the other PRs. I could have created an ad-hoc PR, but I didn't and this one seemed small enough to just fix the travis issue "inline".

olaurino:
I am double checking the xspec issue. It could be it's just relevant to the specific conda environment we are creating on travis, and I was reproducing it locally just because I was starting from that environment. If I start from scratch building the xspec extension with numpy 1.14 I don't get the segfault. So I think the issue is just specific to the travis jobs, and it probably just needs me to create new xspec, and maybe cfitsio and ccfits conda packages for testing.

PR #420, opened 2017-12-14
import mock on both python 2 and 3

### LINKED ISSUE #317: test `test_smoke.py` imports `mock` as a python 2 dependency
`test_smoke.py` should import `mock` from the standard library in Python 3. Since `mock` is also available as an independent package on Python 3, a workaround is to simply install `mock` as part of the dependencies, which is what happens both in CIAO and on Travis since `mock` is listed in the `test_requirements.txt` file and it is installed as part of the regression tests tools in CIAO.

olaurino:
Fix #317.

### DIFF
--- a/sherpa/astro/utils/tests/test_smoke.py
+++ b/sherpa/astro/utils/tests/test_smoke.py
@@ -17,7 +17,11 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-import mock
+try:
+    import mock # Python 2
+except ImportError:
+    from unittest import mock # Python 3
+
 import pytest
 from sherpa import smoke
 


sherpa/astro/utils/tests/test_smoke.py
```diff
@@ -17,7 +17,11 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-import mock
+try:
```
DougBurke: Would it be better to follow https://github.com/sherpa/sherpa/blob/master/sherpa/ui/tests/test_session.py#L32 and import the python 3 version first - e.g.

```
try:
    from unittest import mock
except ImportError:
    import mock
```

(then when we remove python 2 support it's hopefully easier not to make a mistake!)

olaurino: It would have been slightly better (but then again we could be even better off using the `pytest-mock` plugin, or using `six.moves`, or something like that, but I think we are OK.

PR #422, opened 2017-12-15
Improve error message with wrong xspec version

olaurino:
Improve the handling of XSPEC versions mismatch.

### DIFF
--- a/sherpa/astro/xspec/utils.py
+++ b/sherpa/astro/xspec/utils.py
@@ -93,14 +93,17 @@ class include_if(object):
     def __call__(self, model_class):
         if not self.condition:
             model_class.version_enabled = False
-            model_class._calc = self._disabled(model_class.__name__)
+            model_class._calc = self._disabled(self.get_message(model_class))
 
         return model_class
 
+    def get_message(self, model_class):
+        return self.DISABLED_MODEL_MESSAGE.format(model_class.__name__)
+
     @staticmethod
-    def _disabled(cls_name):
+    def _disabled(message):
         def wrapped(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


sherpa/astro/xspec/utils.py
```diff
@@ -93,4 +95,5 @@ class version_at_least(include_if):
     the xspec version detected at runtime is equal or greater than the one provided to the decorator.
     """
     def __init__(self, version_string):
-        include_if.__init__(self, _xspec.get_xsversion() >= version_string)
+        xspec_version = LooseVersion(_xspec.get_xsversion())
```
DougBurke: Would it make sense to make `xspec_version` a module-level "constant", since it shouldn't be something that changes within a test run?

olaurino: Done.

DougBurke:
Should there be an explicit check of this message in the tests? In other words, a test with a hard-coded knowledge of what the error message should look like. The idea being that if the error message text is changed, we will get a test failure, which would then let us know to update any accompanying documentation. For this particular test it is likely overkill.

olaurino:
> Should there be an explicit check of this message in the tests?

Yes, that's why this is still WIP and I didn't update the description. I need to make sure it's actually working whether it's through a unit test of even just a local test. I don't even know whether that line is being hit during the tests. I'll get to it asap.

olaurino:
Rebasing to resolve conflicts and get the latest changes from master (was ffc4cc4)

olaurino:
I was hoping I could test this, even just interactively, I just haven't had any time to do it.

PR #427, opened 2018-01-19
Add support for XSPEC models added in 12.9.1 (fix #331)

### LINKED ISSUE #331: Add support for XSPEC 12.9.1 models
XSPEC 12.9.1 has just been released and includes new models and a change to the table-model code. I was able to build Sherpa against a pre-release of the 12.9.1 release, so users should be able to use this new version, but they will not be able to use the new models (and the `nlapec` model which was introduced in 12.9.0). Note that some of these models are convolution models, which we do not fully support at this time in Sherpa.

The list of new models, from the release notes at https://heasarc.gsfc.nasa.gov/xanadu/xspec/CHANGELOG.txt

```
XSPEC 12.9.1 release notes
Jan 2017
___________________________

New features:

- New Models:

    carbatm          - NS non-magnetic C atmosphere 
    clumin           - calculates luminosity
    hatm             - NS non-magnetic H atmosphere 
    ismabs           - High-resolution ISM absorption
    rfxconv          - reflection from ionized disk
    slimbh           - slim accretion disk
    snapec           - galaxy cluster spectrum using SN yields
    (b)(v)(v)tapec   - apec model with different temperatures for continuum
                       and emission lines
    tbfeo            - tbabs model with O and Fe only variable
    tbgas            - tbabs model with no grains
    tbpcf            - tbabs model with partial covering
    tbrel            - tbabs model that allows negative columns
    vashift          - velocity shift an additive component
    vmshift          - velocity shift a multiplicative component
    voigt            - emission line with a Voigt profile
    xilconv          - reflection from ionized disk
    xscat            - dust scattering out of the beam 
```

DougBurke:
# Release Notes

Add support for models added in XSPEC 12.9.1: bvtapec, bvvtapec, tapec, vtapec, vvtapec, carbatm, hatm, ismabs, slimbh, snapec, TBfeo, TBgas, TBpcf, TBrel, voigt, xscat.

Version 12.9.0 of XSPEC can still be used, in which case the models can be generated - e.g. a user can say xstapec.mdl to create a Python object - but it will error out when evaluated.

The Si and S elemental parameters for the ismabs model have been renamed from the XSPEC versions since they are not unique using the caseinsensitive matching used by Sherpa: so SI, SII, SIII and SiI, SiII, SiIII have been renamed S_I, S_II, S_III and Si_I, Si_II, and Si_III respectively.

Low level support for the following convolution models from XSPEC 12.9.1 has also been added, but there is no Python model support for these: clumin, rfxconv, vashift, vmshift, xilconv.

# Status

Fix #331.

The idea is to address #331 - this is a WIP and all the new models have been added, but we need to work out what to do with the ismabs parameter names, since it has parameters SI, SII, and SII which conflict - when compared in a case-insensitive manner as is done in Sherpa - for the parameters Si, SiI, and SiII. The proposed solution is to use an underscore after the element name *just* for these two elements: so `Si_I` and `S_II` which are now unique when compared in a case-insensitive manner.

I'm leaving this pr:hold until the existing XSPEC prs are resolved (I think just #394).

# Details

For XSPEC 12.9.0 users, the following models will result in a run-time error when they are evaluated.

Added support for the following additive XSPEC models from 12.9.1:
    bvtapec, bvvtapec, tapec, vtapec, vvtapec, carbatm, hatm, slimbh, snapec, voigt, xscat

Adds support for the following multiplcative XSPEC models from 12.9.1:
    ismabs, TBfeo, TBgas, TBpcf, TBrel, 

This adds low-level support (that is, functions in `sherpa.astro.xspec._xspec` are provided, but no Python classes wrapping these, which is what has been done for other XSPEC convolution models) for the following XSPEC convolution models from 12.9.1:
  clumin, rfxconv, vashift, vmshift, xilconv

# Notes

There are no Python models associated with the convolution models, so no tests have been added (there are existing tests of the interface to the convolution models, namely those interfaced using the
XSPECMODELFCT_CON macro, as these new ones are).

This commit marks the first - to my knowledge - interface to a model marked as "c_XXX" in the `model.dat` file. The same interface as used for "C_XXX" models is used. This needs some review.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -8791,5 +8791,205 @@ class XSbvvtapec(XSAdditiveModel):
         XSAdditiveModel.__init__(self, name,
                                  (self.kT, self.kTi, self.H, self.He, self.Li, self.Be, self.B, self.C, self.N, self.O, self.F, self.Ne, self.Na, self.Mg, self.Al, self.Si, self.P, self.S, self.Cl, self.Ar, self.K, self.Ca, self.Sc, self.Ti, self.V, self.Cr, self.Mn, self.Fe, self.Co, self.Ni, self.Cu, self.Zn, self.Redshift, self.Velocity, self.norm))
 
+
+@version_at_least("12.9.1")
+class XStapec(XSAdditiveModel):
+    """The XSPEC tapec model: APEC emission spectrum with separate continuum and line temperatures.
+
+    The model is described at [1]_. The ``set_xsabund`` and ``get_xsabund``
+    functions change and return the current settings for the relative
+    abundances of the metals. The ``set_xsxset`` and ``get_xsxset``
+    functions are used to set and query the XSPEC XSET parameters, in
+    particular the keywords "APECROOT", "APECTHERMAL", "APECVELOCITY",
+    "APECNOLINES", and "APEC_TRACE_ABUND".
+
+    Attributes
+    ----------
+    kT
+        Continuum temperature, in keV.
+    kTi
+        Line temperature, in keV.
+    Abundanc
+        The metal abundance of the plasma, as defined by the
+        ``set_xsabund`` function and the "APEC_TRACE_ABUND" xset
+        keyword (which defaults to 1.0).
+    Redshift
+        The redshift of the plasma.
+    norm
+        The normalization of the model: see [1]_ for an explanation
+        of the units.
+
+    See Also
+    --------
+    XSbtapec, XSvtapec, XSvvtapec
+
+    Notes
+    -----
+    This model is only available when used with XSPEC 12.9.1 or later.
+
+    References
+    ----------
+
+    .. [1] https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSmodelTapec.html
+
+    """
+
+    __function__ = "C_tapec"
+
+    def __init__(self, name='tapec'):
+        self.kT = Parameter(name, 'kT', 1.0, 0.008, 64.0, 0.008, 64.0, 'keV')
+        self.kTi = Parameter(name, 'kTi', 1.0, 0.008, 64.0, 0.008, 64.0, 'keV')
+        self.Abundanc = Parameter(name, 'Abundanc', 1.0, 0.0, 5.0, 0.0, 5.0)
+        self.Redshift = Parameter(name, 'Redshift', 0.0, -0.999, 10, -0.999, 10)
+        self.norm = Parameter(name, 'norm', 1.0, 0.0, 1.0e24, 0.0, hugeval)
+        XSAdditiveModel.__init__(self, name,
+                                 (self.kT, self.kTi, self.Abundanc, self.Redshift, self.norm))
+
+
+@version_at_least("12.9.1")
+class XSvtapec(XSAdditiveModel):
+    """The XSPEC vtapec model: APEC emission spectrum with separate continuum and line temperatures.
+
+    The model is described at [1]_. The ``set_xsxset`` and ``get_xsxset``
+    functions are used to set and query the XSPEC XSET parameters, in
+    particular the keywords "APECROOT", "APECTHERMAL", "APECVELOCITY",
+    "APECNOLINES", and "APEC_TRACE_ABUND".
+
+    Attributes
+    ----------
+    kT
+        Continuum temperature, in keV.
+    kTi
+        Line temperature, in keV.
+    He, C, N, O, Ne, Mg, Al, Si, S, Ar, Ca, Fe, Ni
+        The abundance of the element in solar units. The trace element
+        abundances are determined by the APEC_TRACE_ABUND xset
+        keyword (the default is 1.0).
+    Redshift
+        The redshift of the plasma.
+    norm
+        The normalization of the model: see [1]_ for an explanation
+        of the units.
+
+    See Also
+    --------
+    XSbvtapec, XStapec, XSvvtapec
+
+    Notes
+    -----
+    This model is only available when used with XSPEC 12.9.1 or later.
+
+    References
+    ----------
+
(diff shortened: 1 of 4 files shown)


DougBurke:
The problem with the `ismabs` model is shown in the display of the default model parameters: note how Sherpa thinks the `S` and `Si` parameters are linked when they should not be (I think one of the checks I originally added for the paramter-alias work, but was then I believe removed in the rewrite by @olaurino *may* have picked this problem up):

```
>>> mdl = xspec.XSismabs()
>>> print(mdl)
ismabs
   ismabs.H     thawed          0.1            0       100000      10^22
   ismabs.HeII  frozen            0            0       100000      10^16
   ismabs.CI    thawed         33.1            0       100000      10^16
   ismabs.CII   frozen            0            0       100000      10^16
   ismabs.CIII  frozen            0            0       100000      10^16
   ismabs.NI    thawed         8.32            0       100000      10^16
   ismabs.NII   frozen            0            0       100000      10^16
   ismabs.NIII  frozen            0            0       100000      10^16
   ismabs.OI    thawed         67.6            0       100000      10^16
   ismabs.OII   frozen            0            0       100000      10^16
   ismabs.OIII  frozen            0            0       100000      10^16
   ismabs.NeI   thawed           12            0       100000      10^16
   ismabs.NeII  frozen            0            0       100000      10^16
   ismabs.NeIII frozen            0            0       100000      10^16
   ismabs.MgI   thawed          3.8            0       100000      10^16
   ismabs.MgII  frozen            0            0       100000      10^16
   ismabs.MgIII frozen            0            0       100000      10^16
   ismabs.SiI   linked            0         expr: ismabs.SII      10^16
   ismabs.SiII  linked            0        expr: ismabs.SIII      10^16
   ismabs.SiIII frozen            0            0       100000      10^16
   ismabs.SI    thawed         2.14            0       100000      10^16
   ismabs.SiI   linked            0         expr: ismabs.SII      10^16
   ismabs.SiII  linked            0        expr: ismabs.SIII      10^16
   ismabs.ArI   thawed         0.25            0       100000      10^16
   ismabs.ArII  frozen            0            0       100000      10^16
   ismabs.ArIII frozen            0            0       100000      10^16
   ismabs.CaI   thawed         0.22            0       100000      10^16
   ismabs.CaII  frozen            0            0       100000      10^16
   ismabs.CaIII frozen            0            0       100000      10^16
   ismabs.Fe    thawed         3.16            0       100000      10^16
   ismabs.redshift frozen            0            0           10           
```

DougBurke:
Commit 6f68ef1 "hacks around" the ismabs problem by renaming both sets of parameters from

    SiI, SiII, SiIII, SI, SII, SIII

to

    SiliconI, SiliconII, SiliconIII, SulfurI, SulfurII, SulfurIII

One tweak would be to only rename one of the elements (maybe Sulfur as it is less characters to type), leaving `SiI, SiII, SiIII` as is. However, they are hard to read.

olaurino:
An alternative might be to introduce `_`s between the element name and the roman numeral. Would that make sense? It might or might call for having similar aliases for all the other parameters that have to do with line names.

DougBurke:
After a discussion with @anetasie I am going to rename SI/SII/SIII to SulfurI/SulfurII/SulfurIII. Will get to that soon (i.e. today).

DougBurke:
Actually, we've decided to change Si and S to Si_ and S_ but leave the other elements alone.

DougBurke:
With the latest commit, the output for the `ismabs` model is now:

```
In [1]: from sherpa.astro import xspec

In [2]: mdl = xspec.XSismabs()

In [3]: print(mdl)
ismabs
   ismabs.H     thawed          0.1            0       100000      10^22
   ismabs.HeII  frozen            0            0       100000      10^16
   ismabs.CI    thawed         33.1            0       100000      10^16
   ismabs.CII   frozen            0            0       100000      10^16
   ismabs.CIII  frozen            0            0       100000      10^16
   ismabs.NI    thawed         8.32            0       100000      10^16
   ismabs.NII   frozen            0            0       100000      10^16
   ismabs.NIII  frozen            0            0       100000      10^16
   ismabs.OI    thawed         67.6            0       100000      10^16
   ismabs.OII   frozen            0            0       100000      10^16
   ismabs.OIII  frozen            0            0       100000      10^16
   ismabs.NeI   thawed           12            0       100000      10^16
   ismabs.NeII  frozen            0            0       100000      10^16
   ismabs.NeIII frozen            0            0       100000      10^16
   ismabs.MgI   thawed          3.8            0       100000      10^16
   ismabs.MgII  frozen            0            0       100000      10^16
   ismabs.MgIII frozen            0            0       100000      10^16
   ismabs.Si_I  thawed         3.35            0       100000      10^16
   ismabs.Si_II frozen            0            0       100000      10^16
   ismabs.Si_III frozen            0            0       100000      10^16
   ismabs.S_I   thawed         2.14            0       100000      10^16
   ismabs.S_II  frozen            0            0       100000      10^16
   ismabs.S_III frozen            0            0       100000      10^16
   ismabs.ArI   thawed         0.25            0       100000      10^16
   ismabs.ArII  frozen            0            0       100000      10^16
   ismabs.ArIII frozen            0            0       100000      10^16
   ismabs.CaI   thawed         0.22            0       100000      10^16
   ismabs.CaII  frozen            0            0       100000      10^16
   ismabs.CaIII frozen            0            0       100000      10^16
   ismabs.Fe    thawed         3.16            0       100000      10^16
   ismabs.redshift frozen            0            0           10           

In [4]: 
```

DougBurke:
I'm assuming the test failure is the known issue that could be solved by pinning numpy as talked about in https://github.com/sherpa/sherpa/pull/383/commits/ee24a91ddaaa20a9e8c9b5a7ca6f45409c6c5b58

olaurino:
Rebasing to resolve trivial conflicts, but I'll trigger the CI jobs anyway. Was 6ae621b.

PR #428, opened 2018-02-08
handle function name changes in XSPEC 12.9.1

DougBurke:
# Release Note
Sherpa now supports multiple versions of the XSPEC models from the 12.9.0 and 12.9.1 series. Some models recommend using the C-style interface over the older FORTRAN one, which may also resolve some memory access issues. For CIAO 4.10 users this means the interfaces to XSPEC models have been updated to the 12.9.1n versions. For Standalone Sherpa users this means they can build and run Sherpa against a larger range of XSPEC versions, and Sherpa will pick the XSPEC low level model function accordingly. Note that Sherpa has been tested against XSPEC 12.9.0i, 12.9.0o, and 12.9.1n.

# Notes

Several models in XSPEC 12.9.1 have changed from a FORTRAN to C style interface. Can we support them?

The first commit is actually just to see how working around the numpy version failure works and doesn't address the subject line. The second line supports conditional compilation for the compiled code - that is, it should build against XSPEC 12.9.0 and 12.9.1. However, the Python module code only supports a single version, which I have chosen to be 12.9.1.

Is there a way to make the function name depend on the XSPEC version? That is, for https://github.com/DougBurke/sherpa/blob/69fb779d8d09723ba235df0289047dd762472608/sherpa/astro/xspec/__init__.py#L531 make `__function__` depend on the XSPEC version? Something like

    __function__ = if_at_least((12,9,1), "C_apec", "xsaped")

I wasn't sure how this would interact with the new support for conditional compilation, which turned the value of the data stored `__function__` into a string rather than a symbol.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1,6 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016, 2017, 2018
-#         Smithsonian Astrophysical Observatory
+#  Copyright (C) 2010, 2015-2018  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -595,7 +594,8 @@ class XSapec(XSAdditiveModel):
 
     """
 
-    __function__ = "xsaped"
+    # __function__ = "xsaped"
+    __function__ = "C_apec"
 
     def __init__(self, name='apec'):
         self.kT = Parameter(name, 'kT', 1., 0.008, 64.0, 0.0, hugeval, 'keV')
@@ -686,7 +686,8 @@ class XSbapec(XSAdditiveModel):
 
     """
 
-    __function__ = "xsbape"
+    # __function__ = "xsbape"
+    __function__ = "C_bapec"
 
     def __init__(self, name='bapec'):
         self.kT = Parameter(name, 'kT', 1., 0.008, 64.0, 0.0, hugeval, 'keV')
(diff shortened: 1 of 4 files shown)


DougBurke:
As expected, the XSPEC 12.9.0 tests failed. Well, I'd have been worried if they didn't! The 12.9.1 test passed though.

olaurino:
> __function__ = if_at_least((12,9,1), "C_apec", "xsaped")

I can implement this and then we can change it if we come up with something better. Not sure if we need the tuple version complexity though... can't we just use a string?

DougBurke:
I was just free-form riffing when coming up with that syntax!

olaurino:
After trying different options I went for something simpler and more explicit.

olaurino:
Oops, I didn't realize I was committing changes to `setup.cfg`. I usually have the file `--assume-unchanged`. I'll force push a new commit.

DougBurke:
LGTM

olaurino:
Resolving conflicts. Was 88da351

olaurino:
@DougBurke do you think we have enough evidence I didn't mess up the conflict resolution?

DougBurke:
The code behind the XSPECMODELFCT_NORM/XSPECMODELFCT_C_NORM templates will fail to compile if you mix up the model type (i.e. send in a fortran style to the _C_ variant or C style to the other one). So if it compiles and runs on both XSPEC 12.9.0 and 12.9.1 then - unless we have lost one of the models - it should be fine. As you didn't have to touch the Python module (__init__.py) in the conflict resolution, then I think we also have evidence that you didn't miss a function (since the `__function__` names would end up causing an error when running the model [theoretically at least].

olaurino:
Thanks. I'll merge then.

DougBurke:
I added an explicit list of the models to the release-notes section above, and checked against your conflict-resolved changes, and the numbers agree (14 models), so I think we are indeed golden.

PR #438, opened 2018-03-13
Change from error to warning for OGIP violations

DougBurke:
# Summary

Given that users can not easily change a response file, and previous
versions of Sherpa would allow the responses to be used, this commit
changes some of the errors recently introduced (in PR #383) into
warnings. The errors for the first bin edge being <= 0 are still
left in because users of the sherpa.astro.ui module will find that
these files are auto-corrected for them (by PR #383).

# Details

The first commit is unrelated, but fixes a "false" failure you get if the HEADAS environment variable ends with a / character (I obviously normally do not include this character), which messes up the file name comparison. Normalizing the paths fixes this problem.

The code in the validation routine is now more complex, but it's not time to properly address this.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -22,6 +22,7 @@ Classes for storing, inspecting, and manipulating astronomical data sets
 """
 
 import os.path
+import warnings
 
 import numpy
 from sherpa.data import BaseData, Data1DInt, Data2D, DataND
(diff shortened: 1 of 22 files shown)


sherpa/astro/data.py
```diff
@@ -94,66 +95,81 @@ class DataOgipResponse(Data1DInt):
     """
     _ui_name = "OGIP Response"
 
-    # FIXME For a future time when we'll review this code in a deeper way: we could have better separation of concerns
-    # if the initializers of `DataARF` and `DataRMF` did not rely on the `BaseData` initializer, and if the class
-    # hierarchy was better organized (e.g. it looks like children must not call their super's initializer.
-    # Also, I'd expect validation to happen in individual methods rather than in a large one, and nested ifs should be
-    # avoided if possible.
+    # FIXME For a future time when we'll review this code in a deeper way: we
```
olaurino: Just to say it, we decided early on to have a limit of 120 lines.

sherpa/astro/data.py
```diff
@@ -94,66 +95,81 @@ class DataOgipResponse(Data1DInt):
...
 
         rtype = self._ui_name
 
         if ethresh is not None and ethresh <= 0.0:
             raise ValueError("ethresh is None or > 0")
 
+        # Only display the first warning, since it doesn't really
+        # help users to have multiple warnings for the same response.
+        #
+        warned = False
```
olaurino: I believe this can be done in a cleaner way using the warnings framework, but I'll have to check.

DougBurke:
Can we get #439 merged into ciao-4.10 and/or master?

DougBurke:
One issue with the warning stuff is that I'm just trying to stop the case of

    warning 1: something is wrong with the data
    warning 2: something different is wrong but with the same bit of data

The warning stuff I've seen lets you stop seeing multiple copies of the same message (i.e. warning 1 or warning 2) but not clear it will only let through one of these if both are created.

Then again, maybe this is over-thinking it and we can just let the user see the multiple messages.

olaurino:
If there are multiple messages users should see both of them but only once. I think that makes sense.

olaurino:
Btw I had a very short week last week. I am going to merge #439. Just let me know if you'd like me to work on this PR.

olaurino:
I started working on this and I'll push a commit soon.

DougBurke:
Sorry - was out yesterday and busy on other things at the moment!

olaurino:
Rebasing (was 69abbaaca)

olaurino:
@DougBurke it would help if you could please confirm that:
  * the test cases were correctly updated
  * as usual, that you are ok with the changes I made.

DougBurke:
Looks fine to me (the GitHub changes view is a bit annoying as it includes unrelated changes, but that's no fault of this PR).

It might be worth attaching a note to the `_assert_userwarning` function that this does not test the order of the messages, just the content. I think that's fine here, since we aren't too bothered about the order, and the test method is short enough that it's obvious from the code. I don't have a strong feeling on this and if it would stop it getting merged today then we don't need to do it.

olaurino:
I hadn't realized the PR was issued against the release branch. As I mentioned in https://github.com/sherpa/sherpa/pull/430#issue-168694613 I am trying to issue PRs against master and then cherry-pick the merge commit in the release branches, which seems beneficial especially for ciao. I am closing this manually.

Issue #426, opened 2018-01-11
Travis jobs would fail with xspec and numpy 1.14


olaurino:
When using numpy 1.14 the xspec Travis jobs would fail with a segmentation fault. I believe this is due to numpy 1.14 bringing in the libgfortran-ng package. Most likely rebuilding the xspec conda package will fix this.

See https://github.com/sherpa/sherpa/pull/419#issuecomment-356750619 and https://github.com/sherpa/sherpa/pull/419#issuecomment-356966044 for more info and context.

olaurino:
It looks like the latest conda build of numpy 1.13.3 also brings in libgfortran and makes our xspec conda package unhappy. I tried rebuilding the xspec conda package by including libgfortran-ng but that doesn't seem to help, so there must be something else (also, I couldn't rebuild xspec 12.9.1n because the old patch file is gone from their server and I had to fall back on 12.9.1u).

PR #439, opened 2018-03-14
Fix travis jobs by pinning numpy to an old version

olaurino:
Latest versions of the numpy conda package include a dependency on libgfortran-ng, which seems incompatible with our conda packages for xspec. This PR makes sure an older version of the numpy package is installed, thus fixing the failing jobs.

We'll need to discuss the way forward (which is basically #426, except that now several older numpy conda packages have been rebuilt with libgfortran-ng as a dependency).


DougBurke:
I think the comment from line 29 (or something similar) should either be repeated for these tests, or perhaps moved before these tests and changed to something like

    Explicit version dependencies (for NumPy) are to ensure that the XSPEC build works. See issue # 426.

DougBurke:
Other than that, LGTM. Let's :shipit:

PR #444, opened 2018-04-19
Improve caching

### LINKED ISSUE #442: Performance Improvements
Model evaluation
    - Model caching, not to recompute models with frozen parameters
    - Paralellization with intelligent parsing of the model expression
 Slowness in fitting X-ray spectra
    - issues in simultaneous fitting of multiple data (more general) - do  not reevaluate the models with the
      same parameters multiple times
 Parallelization for MonCar and Nelder-Mead methods:
   - lower priority until the two parts as described above are done

dtnguyen2:
# Release Note
When fitting multiple datasets simultaneously Sherpa now tries to cache model evaluations to improve performance.

# Notes

This PR addresses one element of #442, in particular:

    2b/ Slowness in fitting X-ray spectra:
          - issues in simultaneous fitting of multiple data (more general) - do not reevaluate the models with
            the same parameters multiple times

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -25,8 +25,8 @@ Classes for storing, inspecting, and manipulating astronomical data sets
 import os.path
 import logging
 import warnings
-
 import numpy
+
 from sherpa.data import BaseData, Data1DInt, Data2D, DataND, Data
 from sherpa.utils.err import DataErr, ImportErr
 from sherpa.utils import SherpaFloat, pad_bounding_box, interpolate, \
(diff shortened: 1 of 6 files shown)


olaurino:
@dtnguyen2 as you pointed out this does not fixes the whole #442, so I'm editing the description to clarify that, rather than saying that this fixes the issue.

anetasie:
@dtnguyen2 I did run a few tests and I can see some improvements. I did not run the XSPEC models in my test. I'll do that in CIAOX after we have the first release candidate.
One issue that I'm not clearly seeing from the Profile is whether or not the model is evaluated for each data set even if the arf energy grid is the same, but rmf matrix is different.

olaurino:
Rebasing to fix a tiny conflict.

olaurino:
Rebasing to resolve conflicts (I also did some clean up). Was cca7754.

Issue #460, opened 2018-05-16
Can we remove support for pyfits now (i.e. just support astropy.io.fits)?


DougBurke:
We have a little bit of code that supports different versions of pyfits, which is untested. I think it's a good time to drop pyfits support (we've labelled it as deprecated for a while now) and just support astropy.io.fits.

DougBurke:
In `sherpa.astro.io.pyfits_backend.py` we have

```
try:
    from astropy.io import fits
except ImportError:
import pyfits as fits

...

# Should we drop support for earlier versions? The code is only tested
# using recent (3.3 or later) versions, so there are no guarantees
# for earlier versions. Should the import order below be reversed,
# so that later versions are tried first?
#
try:
    # pyfits-1.3 support
    _VLF = fits.NP_pyfits._VLF
except AttributeError:
    # pyfits-2.3.1 support
    try:
        _VLF = fits.core._VLF
    except AttributeError:
        # pyfits-3.0 support
        _VLF = fits.column._VLF

# fits.CardList is deprecated
if hasattr(fits, 'Header'):
    _new_header = fits.Header
else:
    _new_header = fits.CardList


# As fits.new_table is deprecated, use the from_columns method
# if available. Note that this should only be done for binary
# tables, as fits.TableHDU.from_columns should be used for
# ASCII tables.
#
if hasattr(fits.BinTableHDU, "from_columns"):
    _new_table = fits.BinTableHDU.from_columns
else:
_new_table = fits.new_table
```

The `Header`/`from_columns` change is - I believe, "relatively" new (i.e. it changed in `astropy`) - so this would be kept (although should add in the version of `astropy` when the change was made so we know when to drop this code). Hopefully we can remove the various attempts to find the `_VLF` symbol: we just need to check that the versions of astropy we want to support support this.

DougBurke:
I installed astropy version 1.0.6 from anaconda (note that I believe AstroPy v 1.0 is now consider "unsupported" but I couldn't find an official document to point to here), and note that it has `astropy.io.fits.column._VLF` and not the `core` or `NP_pyfits` variants, as shown below. So I'll put together a simple PR to clean up the imports here.

```
% python
Python 2.7.15 |Anaconda, Inc.| (default, May  1 2018, 23:32:55) 
[GCC 7.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from astropy.io import fits
>>> import astropy
>>> astropy.__version__
u'1.0.6'
[...]

PR #462, opened 2018-05-17
Remove pyfits support (address #460)

DougBurke:
The pyfits support has long been marked as deprecated and we have
not been testing it, so it's not even clear if it worked.

# Notes

One thing I have not done is a systematic check to ensure that all symbols affected by this change have tests. Note that I haven't added any functionality, but it's not clear to me that our test suite really exercises all these code paths, so it might be a good time to expand the test coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,8 +205,7 @@ You can start using Sherpa by starting a Python shell, or you can install
 `ipython` and use it as a more convenient shell. We recommend that you also install
 `ipython-notebook` and `matplotlib` so that you can use the nice `ipython` notebook
 features and the seamless integration with `matplotlib` for plotting from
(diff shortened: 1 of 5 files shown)


olaurino:
I ran all the tests with coverage using the latest astropy, and I can see that the imports for the older versions of pyfits are exercised. In some cases, where the backend uses `hasattr` the IDE tells me that the alternative doesn't make sense. The backend has 78% coverage. Most of the stuff that is not covered is error handling, plus the `set_image_data` function, which maybe I am skipping because I don't have ds9 installed where I am running the tests. In any case it doesn't seem to have anything to do with pyfits. I wouldn't really know how to make this more systematic, but I think the risk here is minimal, so I'd take the changes.

PR #466, opened 2018-05-30
Fix bounding box out-of-bounds memory read

DougBurke:
# Release Note
Avoid an out-of-bounds memory read when calling `pad_bounding_box` (when the data is not matching the expected conditions for this call).

# Notes

This bug was noted when adding documentation for this routine. It is not clear whether the chosen approach - just to avoid copying over elements once the kernel size has been met - is sensible; an alternative would be to error out (i.e. make it a pre-condition that the valid mask elements must match, or be smaller than, the kernel size). This is a "low-level" routine, and it is unclear to me how it is expected to be used, so I opted for the more-permissive approach in this PR.


olaurino:
Rebasing, was 61fa41e

Issue #425, opened 2018-01-11
numpy 1.14 changes the way values are printed, affecting `serialize.py`


olaurino:
From https://github.com/sherpa/sherpa/pull/419#issuecomment-356750619:

> Now, the problem seems to arise when numpy values are printed, which is why the offending test is test_serialize.py and the default parameters for the optimization methods (e.g. EPSILON) are numpy values. numpy 1.14 has a new algorithm for the representation of floating point values(see: https://github.com/numpy/numpy/blob/master/doc/release/1.14.0-notes.rst#float-printing-now-uses-dragon4-algorithm-for-shortest-decimal-representation).
> 
> I think I can set a numpy printing options that prints out the "legacy" values and have the tests pass. It seems to work locally, but I'll need to push the change to Travis and make sure it works. Also, I think I need to update Travis so it runs numpy 1.13 on at least one job.

> Of course the question is whether we should update the serialization code to keep up with numpy (but what about backward compatibility?)

olaurino:
For the record, the issue was e.g. in this Travis job:
https://travis-ci.org/sherpa/sherpa/jobs/328210223

PR #467, opened 2018-06-06
Add simple MacOS Travis job

### LINKED ISSUE #199: Improving build configuration (e.g. .travis.yml)
I think that our current `.travis.yml` file could be improved since
- it is hard for me to identify what is going on in a particular build (there's always going to be some configuration complexity somewhere, but hopefully the current situation can be improved)
- I think it would be better if all the conda packages were installed with one call, to ensure that everything is compatible (the case I've seen involved changes to the NumPy ABI and switching between packages that were built against different versions of NumPy caused build failures).
- it is not clear to me how easy it is going to be to extend to support multiple python versions (maybe I'm missing something obvious in this one!)

Switching to using  [tox](https://tox.readthedocs.org/en/latest/) is one possibility, but I'm sure there are other ways to do this. If anyone has any ideas, please holler.

### LINKED ISSUE #426: Travis jobs would fail with xspec and numpy 1.14

olaurino:
This PR introduces a Travis job for macOS. The job does not include the xspec and ds9 tests, which will be included in a different PR when the xspec conda package for macOS will be available.

Note that in order to make the builds possible on Travis both for macOS and Linux, and in order to make them both simple and not too different from each other, the travis configuration is a complete overhaul of the previous one. As a result, this will fix #199 (@DougBurke please confirm) and it will also fix #426, as there is now special configuration for making sure that the xspec extension links against libgfortran-3 and not the newer versions brought in by numpy. The new builds now use the conda built-in compilers.

I can't reproduce #425 with the new jobs, so probably that should be closed too. I don't have time to investigate, I'd rather close the issue and open it again if it comes back (@DougBurke again please confirm).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,23 +12,19 @@ matrix:
     # We need to use TEST=none to remove the test-data submodule.
     - env: INSTALL_TYPE=develop TEST=none NUMPYVER=1.11 TRAVIS_PYTHON_VERSION="2.7"
     # Full build (including ds9 and xspec), Python 2.7, setup.py develop, astropy, matplotlib 1.5
-    # Note we are pinning down the numpy version to the latest that worked with our xspec conda package
-    - env: XSPECVER="12.9.0i" NUMPYVER="1.11.3=py27_0" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="2.7"
+    - env: XSPECVER="12.9.0i" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="2.7"
       sudo: required
       dist: trusty
     # As above, Python 3.5
-    # Note we are pinning down the numpy version to the latest that worked with our xspec conda package
-    - env: XSPECVER="12.9.0i" NUMPYVER="1.11.3=py35_0" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.5"
+    - env: XSPECVER="12.9.0i" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.5"
       sudo: required
       dist: trusty
     # As above, Python 3.6, setup.py install
-    # Note we are pinning down the numpy version to the latest that worked with our xspec conda package
-    - env: XSPECVER="12.9.0i" NUMPYVER="1.11.3=py36_0" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
+    - env: XSPECVER="12.9.0i" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
       sudo: required
       dist: trusty
     # As above, xspec 12.9.1n
-    # Note we are pinning down the numpy version to the latest that worked with our xspec conda package
-    - env: XSPECVER="12.9.1" NUMPYVER="1.13.3=py36ha12f23b_0" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
+    - env: XSPECVER="12.9.1" NUMPYVER="1.14" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
       sudo: required
       dist: trusty
     # Install sherpatest package rather than relying on relative location of the submodule
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,18 +37,18 @@ matrix:
     - env: INSTALL_TYPE=develop TEST=none FITS="astropy" TRAVIS_PYTHON_VERSION="3.5"
 
 before_install:
-  - . travis/setup_conda.sh
+  - source travis/setup_conda.sh
 
   # XSPEC and DS9
   - if [ -n "${XSPECVER}" ];
-      then . travis/setup_xspec_ds9.sh;
+      then source travis/setup_xspec_ds9.sh;
     fi
 
 install:
     - python setup.py $INSTALL_TYPE &> install.log
 
 script:
-  - . travis/test.sh
+  - source travis/test.sh
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,12 @@ language: c
 
 sudo: false
 
-addons:
-  apt:
-    packages: &default_apt
-      - gfortran
-
 matrix:
   fast_finish: true
   include:
+    # macOS build
+    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
+      os: osx
     # Barebone build to check tests pass when most of the tests must be skipped.
     # We need to use TEST=none to remove the test-data submodule.
     - env: INSTALL_TYPE=develop TEST=none NUMPYVER=1.11 TRAVIS_PYTHON_VERSION="2.7"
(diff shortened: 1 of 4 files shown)


olaurino:
I've had a lot of issues today with Travis jobs timing out while installing the conda dependencies or running `apt-get update`.

travis/setup_conda.sh
```diff
@@ -0,0 +1,55 @@
...
+# We create a new environment so we don't care about the python version in the root environment.
+conda create --yes --quiet -n build python=$TRAVIS_PYTHON_VERSION pip ${MATPLOTLIB} ${NUMPY} $XSPEC $FITS ${COMPILERS}\
+  libgfortran=${LIBGFORTRANVER}
+
+source activate build
+
+# It looks like on some systems (well, linux) the F90 variable needs to be set. Not sure why
+export F90=${F77}
+
+# This is required to make sure that the CIAO python extensions being build pick the correct flags
```
DougBurke: pointless typo warning "being build pick" -> "being built pick"

travis/setup_xspec_ds9.sh
```diff
@@ -0,0 +1,32 @@
...
+# untar them
+THIS_DIR=$(pwd);
+cd $MINICONDA/bin; tar xf $THIS_DIR/$DS9_TAR; tar xf $THIS_DIR/$XPA_TAR; cd -;
+
+# install build dependencies
+sudo apt-get update
+sudo apt-get install -qq libwcs4 wcslib-dev libx11-dev libsm-dev libxrender-dev;
+
+# Some relevant environment variables
+export HEADAS=$MINICONDA/envs/build/Xspec/spectral;
```
DougBurke: Shouldn't HEADAS be set to .../envs/build/Xspec/spectral/x85_64-blah-blah-libc2.15.0?

See step 8 of https://heasarc.gsfc.nasa.gov/lheasoft/install.html

In which case XSPEC_INCLUDE_PATH is just ${HEADAS}/include/

travis/setup_xspec_ds9.sh
```diff
@@ -0,0 +1,32 @@
...
+export LD_LIBRARY_PATH=$MINICONDA/envs/build/Xspec/x86_64-unknown-linux-gnu-libc2.15-0/lib/;
```
DougBurke: I've never had to set LD_LIBRARY_PATH to this to get Sherpa to work, but maybe I've been lucky (possibly because I've tended to build a full XSPEC installation rather than a models-only build, and maybe the installation is subtly-different for these two approaches).

DougBurke:
Is the idea that the macOS version of XSPEC/ds9 testing will be part of this PR or it will come later? I'm fine with either.

olaurino:
As I mentioned in the description, it's coming with its own PR.

travis/setup_xspec_ds9.sh
olaurino: I don't usually do that either, but it depends on the setups, and I didn't question the previous setup. I'll try removing that line and see what happens.

.travis.yml
```diff
@@ -43,85 +37,18 @@ matrix:
...
-  # Install test data as a package, then remove the submodule
-  - if [ ${TEST} == package ];
-     then pip install ./sherpa-test-data;
-     git submodule deinit -f .;
+      then . travis/setup_xspec_ds9.sh;
     fi
 
 install:
-    - python setup.py $INSTALL_TYPE &> install.log
+    - python setup.py $INSTALL_TYPE
```
DougBurke: Do we not have to worry about the screen output anymore (i.e. the removal of the pipe-to install.log) part? I forget exactly why this was added in the first place.

olaurino: Oops, good catch, I was supposed to put the redirect back in after debugging the build. It's interesting that no jobs failed, though. A lot of times the jobs would fail for the logs were too long. Not sure if they removed the limitation, but we don't really need to see all the output from the build, so I'll add the redirect back.

travis/setup_xspec_ds9.sh
olaurino: Maybe I am missing something... `$MINICONDA/envs/build/Xspec/spectral` only contains the `manager` and `modelData` folders, so indeed it's what `HEADAS` should point to. While the include files are not there, they are located in the platform-dependent folder. Am I missing something?

travis/setup_conda.sh
```diff
@@ -0,0 +1,55 @@
...
+SHERPA_CHANNEL=sherpa
+XSPEC_CHANNEL=cxc/channel/dev
+MINICONDA=$HOME/miniconda
+
+if [[ ${TRAVIS_OS_NAME} == linux ]];
+then
+    MINICONDA_OS=Linux
+    export COMPILERS="gcc_linux-64 gxx_linux-64 gfortran_linux-64"
+    sed -i.orig "s|#extra-fortran-link-flags=|extra-fortran-link-flags=-shared|" setup.cfg
+else  # osx
```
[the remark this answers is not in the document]
olaurino: I don't think Travis has any plans for supporting Windows, and neither do we. If that changes we'll notice and it will be easy to fix (except that everything else will break on Windows or not even run), so it's safe to assume it's either mac or linux.

travis/setup_xspec_ds9.sh
olaurino: I think it's different with models-only, `HEADAS` is the environment variable that the Sherpa extension needs to point to the folder containing `manager` and `modelData`. That's how we've always done it in Sherpa. We don't run the init script and we don't require users to run it. There used to be instructions on how to build the models-only package, the document seems to be gone now (https://heasarc.gsfc.nasa.gov/docs/software/lheasoft/xspec-models-only.txt), I wouldn't change the way we've done things so far, but maybe we should double-check our installation instructions work?

travis/setup_xspec_ds9.sh
```diff
@@ -0,0 +1,33 @@
...
+ds9_tar=ds9.centos6.7.5.tar.gz;
+xpa_tar=xpa.centos6.2.1.17.tar.gz;
+
+# Fetch them
+wget --quiet $ds9_site$ds9_tar;
+wget --quiet $xpa_site$xpa_tar;
+
+# untar them
+THIS_DIR=$(pwd);
+cd $MINICONDA/bin; tar xf $THIS_DIR/$ds9_tar; tar xf $THIS_DIR/$xpa_tar; cd -;
```
DougBurke: I'm guessing the travis failures are because MINICONDA needs to be changed to miniconda in this file?

travis/test.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Run smoke test
+cd /home;
+sherpa_smoke ${smokevars};
+
+# Run regression tests using sherpa_test
+if [ ${TEST} == package ] || [ ${TEST} == none ];
+    then cd $HOME;
+    sherpa_test;
+fi
```
DougBurke: github is complaining about a lack of a new line here; do we care about that?

travis/setup_xspec_ds9.sh
DougBurke: Also, now this is a separate script, this need not all be on one line.

olaurino: Yeah, my IDE can be smart at refactoring, but I need to be smarter ;)

travis/test.sh
olaurino: Yeah, we should please our masters!

travis/setup_xspec_ds9.sh
```diff
@@ -1,32 +1,36 @@
...
 # Fetch them
-wget --quiet $DS9_SITE$DS9_TAR;
-wget --quiet $XPA_SITE$XPA_TAR;
+wget --quiet $ds9_site$ds9_tar
+wget --quiet $xpa_site$xpa_tar
 
 # untar them
-THIS_DIR=$(pwd);
-cd $MINICONDA/bin; tar xf $THIS_DIR/$DS9_TAR; tar xf $THIS_DIR/$XPA_TAR; cd -;
+THIS_DIR=$(pwd)
```
DougBurke: Should THIS_DIR be lowercase too? It is not clear to me what rule you are using for deciding on the case of a symbol here ;-)

olaurino: yes :)

In bash I have a tendency to export each and every variable I use, which is not really the way to go, so I am trying to be more disciplined. In this sense, only environment variables are upper case, and you are right, `THIS_DIR` should be lower case.

olaurino:
Either Travis or Anaconda are having serious bandwidth issues and most builds are timing out.

DougBurke:
Fingers crossed for this last build...

travis/setup_xspec_ds9.sh
olaurino: I am reorganizing the script for the inclusion of the xspec and ds9 tests, so I'll take care of that there.

DougBurke:
So @olaurino  do you want approval for merging this PR as is, or wait for the reorganization you mention above https://github.com/sherpa/sherpa/pull/467#discussion_r193847284 ?

olaurino:
I was assuming you were OK with the changes... I don't know how long it will take to have a functioning xspec conda package to test (I am taking a couple of days off next week) so I'd rather merge this now.

travis/setup_conda.sh
```diff
@@ -0,0 +1,55 @@
...
+    compilers="gcc_linux-64 gxx_linux-64 gfortran_linux-64"
+    sed -i.orig "s|#extra-fortran-link-flags=|extra-fortran-link-flags=-shared|" setup.cfg
+else  # osx
+    miniconda_os=MacOSX
+    compilers="clang_osx-64 clangxx_osx-64 gfortran_osx-64"
+
+    # This is required on macOS when building with conda
+    sed -i.orig "s|#extra-fortran-link-flags=|extra-fortran-link-flags=-undefined dynamic_lookup -bundle|" setup.cfg
+
+    # It looks like xvfb doesn't "just work" on osx travis, so...
```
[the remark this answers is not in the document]
olaurino: Right, I'll see if I can include it in the next round of changes. Not a big deal, but it's good to remove a moving part from the other jobs.

PR #473, opened 2018-07-18
Add macOS tests and fix failing tests.

olaurino:
This PR makes sure some tests run successfully both on Linux and macOS. Some of them provide equal results (within the default tolerance) when the optimization routine is changed to LevMar, while others need to be compared within a larger tolerance. @dtnguyen2 did more research to figure out the numerical differences, and it turns out they are due to the different platform: when the same compiler is used to compile the optimization routine on Linux and macOS the optimization finds different minima on different platforms.

Note that this PR also enables the complete set of tests including xspec and ds9 on the macOS platform.


olaurino:
Compare
https://travis-ci.org/sherpa/sherpa/jobs/405430545

to

https://travis-ci.org/olaurino/sherpa/jobs/395489842

PR #475, opened 2018-07-18
Fix unecessary runtime warning (#402)

### LINKED ISSUE #402: optfcts.py - Unnecessary Runtime warning
On lines 489 and 677 of optfcts.py, there is a check on the bounds and in parallel there is a check on whether the value is a NaN.  This check needs to be done in series to avoid the Runtime error that numpy throws in the bounds check function.  

The offending code is: 

    def stat_cb0(x_new):
        if _outside_limits(x_new, xmin, xmax) or _my_is_nan(x_new):
            return FUNC_MAX
        return orig_fcn(x_new)

A fix would look like this as an option:
    def stat_cb0(x_new):
        if _my_is_nan(x_new):
            return FUNC_MAX
        if _outside_limits(x_new, xmin, xmax) :
            return FUNC_MAX
        return orig_fcn(x_new)

dtnguyen2:
# Release Note
A condition check in the optimization code was modified so as not to produce a warning when the value is a NaN.

Fixes #402

### DIFF
--- a/sherpa/optmethods/optfcts.py
+++ b/sherpa/optmethods/optfcts.py
@@ -488,7 +488,7 @@ def minim(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, step=None,
 
     orig_fcn = stat_cb0
     def stat_cb0(x_new):
-        if _outside_limits(x_new, xmin, xmax) or _my_is_nan(x_new):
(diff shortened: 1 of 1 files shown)


DougBurke:
Can we change the language of the commit message/PR text from "the complainer" to "the original reporter"?

olaurino:
Rebasing, rewording, and cleaning up the PR.

PR #481, opened 2018-08-01
Remove Fortran code

dtnguyen2:
# Release Note
Sherpa does not rely on any Fortran routines anymore, except those compiled in XSPEC. The existing Fortran code, which was used mainly in some of the optimization routines, has been replaced by equivalent C++ code. This means that `gfortran` is no longer required for building Sherpa, although a version of `libgfortran` compatible with the one used to link the XSPEC libraries would still be needed at runtime if the XSPEC extension is built.

# Notes
1) Had to changed on sherpa thead (lev3fft.py) test: changed src.fwhm from 0.0441858 to 0.044178 to get it to pass.
2) Enabled LevMar to return covariance matrix and changed output of fitted pars to include errors


dtnguyen2:
LevMar does not handle parameters limits/bounds as well as possible, ie it does not transform.  Having the code in C/C++ now should make transition easier for the implementer.

olaurino:
The code did not compile with certain compilers as the C++ methods were incorrectly called. I fixed that.

There were more residual usages of the fortran flags that prevented the build from happening, and I removed those too.

One potential issue with this PR is that it requires a new dependency, `autograd`, which does not have a conda package available. I added the dependency to the python code, which should hopefully make the travis tests work, but I wouldn't know today how to package the conda binaries for release, as autograd cannot be included as a dependency. There is a package in conda-forge, but up until now I have tried to stay away from conda-forge, as it seems to be more harm than good.

olaurino:
Unfortunately including `autograd` clashes with `numpy` 1.11, which we still use as a baseline version for some of the tests. I tried updating it to 1.12, let's see what happens.

olaurino:
I am also bumping the minimum version of matplotlib we test against to 2, as 1.5 was incompatible with numpy 1.12. At this point I am just trying to get the tests to pass, we'll need to figure out what to do with the dependencies and the tests.

dtnguyen2:
autograd is not needed (it was some left-over work which should have been rm from my work on autodiff).  I tried to push the changes but was unable to do cause I guess Omar had messed with it:

olaurino:
All right. I'll revert the changes I made, cancel all the tests on travis, and push again. Fingers crossed.

olaurino:
@dtnguyen2 two tests are failing on Linux and four on macOS. They are all tests requiring xspec.

See
https://travis-ci.org/sherpa/sherpa/jobs/416088554

and
https://travis-ci.org/sherpa/sherpa/jobs/416088556

sherpa/optmethods/__init__.py
```diff
@@ -336,71 +336,10 @@ def __init__(self, name='gridsearch'):
...
-    ----------
-
-    .. [1] J.J. More, "The Levenberg Marquardt algorithm:
-           implementation and theory," in Lecture Notes in Mathematics
-           630: Numerical Analysis, G.A. Watson (Ed.),
-           Springer-Verlag: Berlin, 1978, pp.105-116.
-
-    """
 
+class LevMar(OptMethod):
```
DougBurke: Some of the docstring should be retained; at least the

```
"""Levenberg-Marquardt optimization method."""
```

part but ideally more.

olaurino:
Rebasing to clean history and fix conflicts. Was a5be52b

DougBurke:
I did a fit with this branch and got the following screen output, where the `+/- error` term is new. I assume this is part of the change here. We need to document this: what do we say about these error estimates (how are they calculated and how reliable are they)?

```
In [25]: ui.fit()
Dataset               = 1
Initial fit statistic = 675034
Final fit statistic   = 6738.63 at function evaluation 9
Data points           = 3827
Degrees of freedom    = 3825
Probability [Q-value] = 5.63697e-165
Reduced statistic     = 1.76173
Change in statistic   = 668296
   mdl.c0         244.128      +/- 0.996649    
   mdl.c1         -0.00677439  +/- 0.000155806 

```

dtnguyen2:
As reliable as the fit, but more reliable as sherpa's covar command.  The
LevMar method has a built-in estimated covariance matrix.  The following is
the excerpt from Numerical Recipes:

    Once the acceptable minimum has been found, one wants to set lamda = 0
and compute the matrix
                                                       [C] = [alpha]^-1
which, as before, is the estimated covariance matrix of the standard errors
in the fitted parameters a (see next section).

Of course, XSpec gives you a warning about using the values etc...

sherpa/optmethods/src/_saoopt.cc
```diff
@@ -606,8 +891,10 @@ static PyMethodDef WrapperFcts[] = {
   FCTSPEC(difevo, py_difevo),
   FCTSPEC(nm_difevo, py_difevo_nm),
   FCTSPEC(lm_difevo, py_difevo_lm),
+  FCTSPEC(cpp_lmder, py_lmder),
```
DougBurke: Do we need `cpp_lmder` crreated/exported as I can't see a place where it is being used?

dtnguyen2:
One can also get the covar matrix via the extra_output, for example:

fit_results = get_fit_results()
covar = fit_results.extra_output['covar']

I can't seem to find it now, but as I remember it XSpec gives the user some
kind of warning about the covarerr from the resulting fit.

dtnguyen2:
Not sure which version you are using but the following should be in the
said file:

static PyMethodDef WrapperFcts[] = {


  FCTSPEC(difevo, py_difevo),
  FCTSPEC(nm_difevo, py_difevo_nm),
  FCTSPEC(lm_difevo, py_difevo_lm),
  FCTSPEC(cpp_lmdif, py_lmdif),
  FCTSPEC(neldermead, py_nm),
  { NULL, NULL, 0, NULL }

};

not:

static PyMethodDef WrapperFcts[] = {


  FCTSPEC(difevo, py_difevo),
  FCTSPEC(nm_difevo, py_difevo_nm),
  FCTSPEC(lm_difevo, py_difevo_lm),
  FCTSPEC(cpp_lmder, py_lmder),
  FCTSPEC(cpp_lmdif, py_lmdif),
  FCTSPEC(neldermead, py_nm),
  FCTSPEC(minim, py_nm_minim),
  { NULL, NULL, 0, NULL }

};

The latter version is my new attempt at gettting rid of the Fortran code in
the stand alone version of sherpa, but some of the tests are failing on
some versions of macs and I haven't had a chance to look at the issues yet.

DougBurke:
Dan, I'm reviewing this PR as requested by Omar, so I'm meant to be looking at your new version.

DougBurke:
Ta - so it sounds like `cpp_lmder` is probably a useful thing to have around for future experiments (e.g. maybe providing an optimiser which can use analytical derivatives?).

DougBurke:
@olaurino I've had a look at the code and - modulo the fact that I've skipped over the C++ side of it completely - it looks fine to me (as of commit 03244c2 aka "increase tolerance for tests that failed with older xspec").

You can consider this an "official" sign off, assuming that the only remaining changes are "test wibbles" (or perhaps doc fixes). I think the fact that the best-fit results in the tests aren't changing here (at least within the tolerance) is as good a sign as my review. Let me know if there's something else ion this you wanted a comment on.

olaurino:
I wouldn't mind an additional check on the README changes. I think I'll merge everything tomorrow morning and kick off the release process.

DougBurke:
I'm concerned about #503 and the excessive memory usage we are seeing. I
believe there code that causes there regressions in the ciao build are also
in 4.10.1

README.md
```diff
@@ -509,9 +504,6 @@ used, but the full path should be in your own copy of the file):
     have been issues seen using the CIAO binaries on certain OS X
     systems.
 
-In all cases, the same version of `gfortran` should be used to build
-Sherpa and XSPEC, in order to avoid possible incompatibilities.
-
 If there are problems building, or using, the module, then the other
```
DougBurke: We have this paragraph which talks about `gfortran_xxx` settings. This seems at odds with our brave, new, Fortran-Free world. I think I added this paragraph, but I don't know of a situation where we've had to change these settings (which is not a solid proof that they are not needed, I admit).

I think we should leave this paragraph in for now, but when we get to the XSPEC 12.10.0 PRs we can look at whether this paragraph still has any use.

dtnguyen2:
A couple of comments about the tolerance.  When I changed the minpack code (lmdif) from Fortran to C++ I only had to change one test: 
 the sherpa thead (lev3fft.py) test: changed src.fwhm from 0.0441858 to 0.04417
note that the fit statistics remained unchanged for this thread so I suspect that is due to the resolution of the data.  Subsequent tests that failed which Omar had to lowered the tolerance were due to errors in the covariance errors (I have added the covar errors from the Fortran code before switching to C++).  So we are talking about the errors on the errors an insignificant amount. 

If you look carefully at the LevMar.cc file, you will notice that I have valgrind the c++ code (outside of sherpa).  One can compile the test from typing the following (documented in the code itself):
g++ -ansi -pedantic -Wall -O3 -I. -I../../../include -I../.. -DtestLevMar.cc LevMar.cc -o levmar
The model functions are run on the tests published in the following paper (the same authors as the original lmdif code itself):

//
// J. MORE', B. GARBOW & K. HILLSTROM,
// "Algorithm 566: Fortran Subroutines for Testing Unconstrained 
// Optimization Software.", ACM TOMS, VOL. 7, PAGES 14-41 AND 136-140, 1981
//
 
Whenever possible I generalized the test model functions from a fixed number of parameters to a general number of free parameters.  For example, the Rosenbrock in the aforementioned paper were written for npar=2,  in my test version npar can be any even number etc...

I included the covariance matrix for the Levenberg-Marquardt algorithm, cause I will (someday) intend to do the same for the NelderMead algorithm since they are useful for Lev3 work (get_draws needs a covariance matrix and sherpa's covar command is iffy at best, never mind about covariance from int_unc).  As I mentioned before, I have always wanted to transform the fitted parameters to handle parameter boundaries.  In theory, I could do that from Fortran, but that is not going to be handy for my next job.

I have thoroughly tested the C++ code, and I very confident that it will be fine

README.md
olaurino: I'll clarify that part. Gfortran is not needed to build the xspec extension. Only libgfortran is, and it needs to be the same (or same version) as the one used to build xspec.

olaurino:
I share your concern on #503. However we are several weeks behind schedule and I need to start working on the actual release. We can still remove #503 if necessary. Or, if we reach a conclusion today, I can remove it right away. Either way, I'll start working on the release.

README.md
```diff
@@ -509,13 +504,16 @@ used, but the full path should be in your own copy of the file):
...
 options may need to be set - in particular the `gfortran_lib_dirs` and
 `gfortran_libraries` settings.
 
+It is important that the `gfortran`-related options above point to a
```
olaurino: @DougBurke does this seem OK?

olaurino: Actually, I am merging this. If we need to reword we can do it anyway.

README.md
olaurino: Maybe it would help to include a setup.cfg.xspec with some of the options preset to a common scenario, with placeholders for the actual paths the user will have to change.

PR #482, opened 2018-08-09
Diag arf/rmf

dtnguyen2:
1) create_arf and create_delta_rmf currently returns sherpa.astro.data.DataARF and sherpa.astro.data.DataRMF class, respectively.  However, get_arf() and get_rmf() return sherpa.astro.instrument.ARF1D and sherpa.astro.instrument.RMF1D, respectively.  Do you want create_arf and create_delta_rmf to be consistent with get_arf and get_rmf. 
2) Currently the default names for create_arf and create_delta_rmf are 'test-arf' and 'delta-rmf', what do you want the default names to be?
3) I need details specs for create_non_delta_rmf()


anetasie:
Input via image file would be the easy user interface to ```create_non_delta_rmf()```. The definitions of how to deconstruct the image into ```n_grp, n_chan, f_chan``` are given in @DougBurke prototype. 
The definition of ```n_grp, n_chan, f_chan``` is in the HEASARC doc:

https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html
Thease are the definitions of the keys in each row:

n_grp -number of channels subsets in a row
n_chan - the number of channels in a subset
f_chan - the first channel in each subset

anetasie:
The image  file can be generated from the standard RMF file with the CIAO tool ```rmfimg```

dtnguyen2:
dmlist tells me that image is 1024x900; However, if I use pycrates to read the image I get an image of 900x1024.  What is a fella supposed to do?

devel12-143: dmlist test_rmfimg.fits blocks

Dataset: test_rmfimg.fits

Block    1: RMF_IMAGE                      Image      Real4(1024x900)

devel12-148: python
Python 2.7.11 (default, May 27 2017, 18:57:15)
[GCC 4.8.3] on linux2
>>> from pycrates import *
>>> a = read_file('test_rmfimg.fits')
>>> print(a.get_image().values.shape)
(900, 1024)

dtnguyen2:
added create_non_diag_rmf; However, I think create_diag_rmf and create_non_diag_rmf should be consolidated into a function called create_rmf with fname default to None.  If fname == None or fname != None then create_rmf should behave like create_diag_rmf or create_non_diag_rmf, respectively

anetasie:
@dtnguyen2 - implementation as one function with an optional input of a filename for a non-square matrix would be good.

dtnguyen2:
I implemented create_rmf which is a combo of create_non_diag_rmf and create_diag_rmf, the question is should we keep or deprecate the create_non_diag_rmf and create_diag_rmf or not.  There is a symmetry with create_rmf and the already implemented create_arf.

sherpa/astro/ui/utils.py
```diff
@@ -4733,6 +4733,162 @@ def pack_table(self, id=None):
...
+
+            return sherpa.astro.data.DataRMF(name, detchans=nchans,
+                                             energ_lo=rmflo, energ_hi=rmfhi,
+                                             n_grp=dummy, n_chan=dummy,
+                                             f_chan=f_chan, matrix=matrix,
+                                             offset=startchan,
+                                             e_min=e_min, e_max=e_max,
+                                             ethresh=ethresh)
+
+        def create_non_delta_rmf(rmflo, rmfhi, fname, startchan=1,
```
olaurino: this function is never exercised by the tests.

sherpa/astro/ui/utils.py
```diff
@@ -4733,6 +4733,162 @@ def pack_table(self, id=None):
...
+            nchans = rmflo.size
+            def calc_grp_chan_matrix(fname):
+                try:
+                    data, filename = \
+                        sherpa.astro.io.backend.get_image_data(fname)
+                    matrix = data['y']
+                    n_grp = []
+                    n_chan = []
+                    f_chan = []
+                    for row in (matrix > 0):
```
olaurino: This condition does not seem to make sense, but I might be wrong. However, since there are no automated tests, I can't run it in a debugger.

dtnguyen2:
Aneta has updated the requirements doc to handle the case when the fit stat is chisqr, I got mail from Aneta yesterday specifiying that the multivariate normal distributions shall be:

NormalParameterSampleFromScaleMatrix

anetasie:
@dtnguyen2 this additional requirement was for EW error in PR #484

dtnguyen2:
load_data knows how to read the associated ARF and RMF files. create_rmf returns a DataRMF class so in order to set the newly created RMF to the pha files one has two options: 

load_data('3c273.pi')
a = get_data()
a.set_rmf(drmf)

or

from sherpa.astro.io import read_pha
a = read_pha('3c273.pi')
a.set_rmf(drmf)

dtnguyen2:
I was wrong (sad but true) in my previous comment, you actually had a third option:
lo=np.linspace(0.1,10,1000)
hi=np.linspace(0.2,10.1,1000)
drmf=create_rmf(lo,hi,e_min=lo,e_max=hi)

load_data('3c273.pi')
set_rmf(1,drmf)

That is rm the single quote in the set_rmf statement.

anetasie:
I get an argument type error with the single quote

```
set_rmf(1,'drmf')

ArgumentTypeErr                           Traceback (most recent call last)
<ipython-input-8-39a97df5a826> in <module>()
----> 1 set_rmf(1,'drmf')

<string> in set_rmf(id, rmf, resp_id, bkg_id)

/Users/aneta/git/sherpa/sherpa/astro/ui/utils.py in set_rmf(self, id, rmf, resp_id, bkg_id)
   5497         if type(rmf) in (sherpa.astro.instrument.RMF1D,):
   5498             rmf = rmf._rmf
-> 5499         _check_type(rmf, sherpa.astro.data.DataRMF, 'rmf', 'an RMF data set')
   5500 
   5501         data = self._get_pha_data(id)

/Users/aneta/git/sherpa/sherpa/ui/utils.py in _check_type(arg, argtype, argname, argdesc, nottype)
     74     if ((not isinstance(arg, argtype)) or
     75             ((nottype is not None) and isinstance(arg, nottype))):
---> 76         _argument_type_error(argname, argdesc)
     77 
     78 

/Users/aneta/git/sherpa/sherpa/ui/utils.py in _argument_type_error(argname, argdesc)
     68 
     69 def _argument_type_error(argname, argdesc):
---> 70     raise ArgumentTypeErr('badarg', argname, argdesc)
     71 
     72 

ArgumentTypeErr: 'rmf' must be an RMF data set
```

dtnguyen2:
create_rmf returns a DataRMF, so in the following statement:

drmf=create_rmf(lo,hi,e_min=lo,e_max=hi)

drmf is a class; However, your statement:

set_rmf(1,'drmf')

you are trying to set a string 'drmf' when you should have said:

set_rmf(1, drmf)

BTW, you may want to change lo and hi to the following:

lo=np.linspace(0.1,10,1024)
hi=np.linspace(0.2,10.1,1024)

That is 1000 --> 1024 to match the number of channels in your pha file, player's choice here though

anetasie:
@dtnguyen2 yes, create_rmf() needs to much the channel number  of the PHA file.

anetasie:
@dtnguyen2 Sherpa should issue an error if `set_rmf()` is called with a wrong number of channels from `create_rmf()`. Currently `set_rmf(drmf)` will not issue an error if the channel number is wrong, `fit()` will result in  `Matched filter error`, which should be caught earlier, by `set_rmf()`.

dtnguyen2:
I can make the changes to set_rmf if data._x != rmf.detchans but I need to know exactly the error statement you want to print (Remember: the past agreement was that SDS must write the exact error message to be written out otherwise SDS will complain about the message I will write for example: ....)

Technically this is not a diag-RMF/ARF since set_rmf bug/enhancement can be triggered with any non-matching rmf (ie not necessary a user-generated rmf).

anetasie:
@dtnguyen2 Agree. the error msgs is not diag-ARF/RMF issue and should be considered with `set_rmf` issues in general. At this moment `set_rmf()` will accept any rmf with incorrect channel numbers, which results in error later on in fitting. I guess this should be tracked in a different issue allowing this PR to be merged.

anetasie:
@dtnguyen2 Can we add the following example to 'create_rmf()' doc string?

```
Note that the number of channels in the diagonal rmf should match the number of channels in the PHA file. Here is an example for a PHA file 'acis.pi':

load_data('acis.pi')
arf1 = get_arf()
rmf1 = get_rmf()

Create an drmf with a linear space corresponding to 1024 channels:

lo=np.linspace(0.1,10,1024)
hi=np.linspace(0.2,10.1,1024)
drmf=create_rmf(lo,hi,e_min=lo,e_max=hi)

set_rmf(1,drmf)
a = get_rmf()
set_rmf(rmf1)

Use in a model expression with default rmf/arf and the diagonal a matrix for different model components:

set_full_model(rmf1(arf1(powlaw1d.p1))+a(const1d.d1))
```

olaurino:
Rebasing to resolve conflicts. Was 7389101

sherpa/astro/tests/test_astro_diag_arfrmf.py
```diff
@@ -0,0 +1,1669 @@
...
+    ArithmeticConstantModel, BinaryOpModel
+from sherpa.astro.instrument import ARF1D, ARFModelNoPHA, ARFModelPHA, \
+    Response1D, RMF1D, RMFModelNoPHA, RMFModelPHA, \
+    RSPModelNoPHA, RSPModelPHA
+from sherpa.fit import Fit
+from sherpa.astro.data import DataARF, DataPHA, DataRMF
+from sherpa.models.basic import Const1D, Polynom1D, PowLaw1D
+from sherpa.utils.err import DataErr
+from sherpa.utils.testing import requires_xspec
+from sherpa.utils import _utils, requires_data, requires_fits
```
DougBurke: The `requires_data` and `requires_fits` sumbols need to be moved to the line above (ie got from `sherpa.utils.testing`).

DougBurke: I don't want to commit the changes myself in case it conflicts with something @olaurino is doing (and I don't have time)

olaurino:
I am pushing the import fix as well as a round of refactoring to make the Session object more stupid and the implementation a tad bit more encapsulated. I also moved part of the checks to the main ARF/RMF classes since I checked with @dtnguyen2 that's where they belonged. I would have liked to have more time to make the refactoring, but this is already better than nothing.

sherpa/astro/instrument.py
```diff
@@ -1141,3 +1142,120 @@ def get_kernel(self, data, subkernel=True):
...
+        raise ioerr
+
+
+def _validate_create_rmf_input(rmflo, rmfhi, startchan, fname=None):
+    try:
+        assert rmflo.size == rmfhi.size
+    except AssertionError:
+        raise ValueError("rmflo and rmfhi must have the same size, not {} and {}".format(rmflo.size, rmfhi.size))
+
+    try:
```
DougBurke: Why not just 'if startchan < 0: raise ValueError(...)' ?

DougBurke: And similarly the check on `rmflo.size == rmfhi.size`?

sherpa/astro/instrument.py
```diff
@@ -1141,3 +1142,120 @@ def get_kernel(self, data, subkernel=True):
...
+        n_chan = numpy.asarray(n_chan, dtype=numpy.int16)
+        matrix = matrix.flatten()
+        matrix = matrix[matrix > 0]
+        return n_grp, f_chan, n_chan, matrix
+    except sherpa.utils.err.IOErr as ioerr:
+        print(ioerr)
+        raise ioerr
+
+
+def _validate_create_rmf_input(rmflo, rmfhi, startchan, fname=None):
```
DougBurke: Given that

1. this routine is not used by `create_non_delta_rmf`
2. I believe `DataRMF` has checks in for this (well, I assume that's what `DataRMF._validate` does, but I haven't got time to check)

is this routine really worth it? It can be nice to get an error "close" to the call site, and that can be specialized for the context of the call site, but I don't know if it is worth it here.

However, we could always remove these checks in a later PR.

sherpa/astro/instrument.py
olaurino: Yes, that's part of what I wanted to avoid with the refactoring. Thanks.

sherpa/astro/instrument.py
olaurino: I think I was trying to get rid of that function altogether and make those checks in the response classes themselves. Let me see if I can still do that or if, as you say, the checks are already there. Of course tests should be doing that, but you know...

sherpa/astro/data.py
```diff
@@ -163,6 +163,9 @@ def _validate_energy_ranges(self, label, elo, ehi, ethresh):
 
         rtype = self._ui_name
 
+        if elo.size != ehi.size:
+            raise ValueError("rmflo and rmfhi must have the same size, not {} and {}".format(elo.size, ehi.size))
```
DougBurke: I suggest changing this message to

```
"The energy arrays must have the same size, not {} and {}" .format(elo.size, ehi.size)
```

since the "useful" name of the arrays (that is, the input column name or function argument name) is "long lost" at this point.

sherpa/astro/data.py
```diff
@@ -397,7 +400,7 @@ def __init__(self, name, detchans, energ_lo, energ_hi, n_grp, f_chan,
 
         energ_lo, energ_hi = self._validate(name, energ_lo, energ_hi, ethresh)
 
-        if offset <=0:
+        if offset <0:
```
DougBurke: Oops - I hadn't noticed that in my review. Good catch.

olaurino:
I need a release note on this PR. @dtnguyen2? @anetasie?

PR #483, opened 2018-08-15
An initial release of simultaneous fit on multicores (slower for most…

dtnguyen2:
# Release Note
Part of the work on sherpa's **Performance Improvements** by performing simultaneous fit on multiple cores.  Note, this is not to be confused with PR #444 and PR #503.

# Description
Sherpa allows the modeling of multiple, independent data sets; However, sherpa evaluates the independent data sets sequentially.  This PR distributes the evaluations of the multiple independent data sets using the multi-cores built-in the user's workstation.  The current default setting for this PR is to evaluate the multiple independent data sets sequentially since the overhead for distributing the workload across multi-cores is high if the evaluation of the data sets is not time consuming.  However, the user/tester can change the default setting at runtime by entering the fit as: ```fit(numcores=N)``` where N is an integer greater then one.

# Notes
The original requirements for this PR can found at #442, although it is hard to say which specific category this PR fits in the initial requirements.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,11 +1,6 @@
 #
-<<<<<<< HEAD
 #  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020
 #      Smithsonian Astrophysical Observatory
-=======
-#  Copyright (C) 2010, 2015, 2016, 2017, 2018
-#             Smithsonian Astrophysical Observatory
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -9998,11 +9993,7 @@ class Session(sherpa.ui.utils.Session):
         # validate the kwds to f.fit() so user typos do not
         # result in regular fit
         # valid_keys = sherpa.utils.get_keyword_names(sherpa.fit.Fit.fit)
-<<<<<<< HEAD
-        valid_keys = ('outfile', 'clobber', 'filter_nan', 'cache')
-=======
         valid_keys = ('outfile', 'clobber', 'filter_nan', 'numcores')
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
         for key in kwargs.keys():
             if key not in valid_keys:
                 raise TypeError("unknown keyword argument: '%s'" % key)
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -9993,7 +9993,7 @@ class Session(sherpa.ui.utils.Session):
         # validate the kwds to f.fit() so user typos do not
         # result in regular fit
         # valid_keys = sherpa.utils.get_keyword_names(sherpa.fit.Fit.fit)
-        valid_keys = ('outfile', 'clobber', 'filter_nan', 'numcores')
+        valid_keys = ('outfile', 'clobber', 'filter_nan', 'cache', 'numcores')
         for key in kwargs.keys():
             if key not in valid_keys:
                 raise TypeError("unknown keyword argument: '%s'" % key)
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,6 +1,11 @@
 #
+<<<<<<< HEAD
 #  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020
 #      Smithsonian Astrophysical Observatory
+=======
+#  Copyright (C) 2010, 2015, 2016, 2017, 2018
+#             Smithsonian Astrophysical Observatory
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -9803,7 +9808,7 @@ class Session(sherpa.ui.utils.Session):
 
         return fit_to_ids, datasets, models
 
-    def _get_bkg_fit(self, id, otherids=(), estmethod=None):
+    def _get_bkg_fit(self, id, otherids=(), estmethod=None, numcores=1):
 
         fit_to_ids, datasets, models = self._prepare_bkg_fit(id, otherids)
 
@@ -9812,7 +9817,7 @@ class Session(sherpa.ui.utils.Session):
 
         fit_to_ids = tuple(fit_to_ids)
 
-        f = self._get_fit_obj(datasets, models, estmethod)
+        f = self._get_fit_obj(datasets, models, estmethod, numcores)
 
         return fit_to_ids, f
 
@@ -9993,15 +9998,21 @@ class Session(sherpa.ui.utils.Session):
         # validate the kwds to f.fit() so user typos do not
         # result in regular fit
         # valid_keys = sherpa.utils.get_keyword_names(sherpa.fit.Fit.fit)
+<<<<<<< HEAD
         valid_keys = ('outfile', 'clobber', 'filter_nan', 'cache')
+=======
+        valid_keys = ('outfile', 'clobber', 'filter_nan', 'numcores')
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
         for key in kwargs.keys():
             if key not in valid_keys:
                 raise TypeError("unknown keyword argument: '%s'" % key)
 
+        numcores = kwargs.get('numcores', 1)
+
         if fit_bkg:
-            ids, f = self._get_bkg_fit(id, otherids)
+            ids, f = self._get_bkg_fit(id, otherids, numcores=numcores)
         else:
-            ids, f = self._get_fit(id, otherids)
+            ids, f = self._get_fit(id, otherids, numcores=numcores)
 
         if 'filter_nan' in kwargs and kwargs.pop('filter_nan'):
             for i in ids:
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -1,6 +1,11 @@
 #
+<<<<<<< HEAD
 #  Copyright (C) 2008, 2015, 2016, 2017, 2019, 2020
 #     Smithsonian Astrophysical Observatory
+=======
+#  Copyright (C) 2008, 2015, 2016, 2017, 2018
+#             Smithsonian Astrophysical Observatory
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -30,9 +35,10 @@ from sherpa.models.regrid import EvaluationSpace1D, EvaluationSpace2D
 from sherpa.utils.err import DataErr
 from sherpa.utils import SherpaFloat, NoNewAttributesAfterInit, \
     print_fields, create_expr, calc_total_error, bool_cast, \
-    filter_bins
+    filter_bins, parallel_map_funcs
 
 
+<<<<<<< HEAD
 __all__ = ('Data', 'DataSimulFit', 'Data1D', 'Data1DInt',
            'Data1DAsymmetricErrs', 'Data2D', 'Data2DInt')
 
@@ -89,6 +95,9 @@ class DataSpace1D(EvaluationSpace1D):
             return self
 
         data = self.grid[0]
+=======
+__all__ = ('Data', 'DataSimulFit', 'Data1D', 'Data1DInt', 'Data2D', 'Data2DInt')
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 
         data = self.filter.apply(data)
         return DataSpace1D(self.filter, data)
@@ -866,20 +875,45 @@ class DataSimulFit(NoNewAttributesAfterInit):
 
     """
 
+<<<<<<< HEAD
     def __init__(self, name, datasets):
         self.name = name
+=======
+    def __init__(self, name, datasets, numcores=1):
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
         if len(datasets) == 0:
             raise DataErr('zerodatasimulfit', type(self).__name__)
         self.datasets = tuple(datasets)
         NoNewAttributesAfterInit.__init__(self)
 
     def eval_model_to_fit(self, modelfuncs):
+<<<<<<< HEAD
         total_model = []
 
         for func, data in zip(modelfuncs, self.datasets):
             total_model.append(data.eval_model_to_fit(func))
 
         return numpy.concatenate(total_model)
+=======
+        if self.numcores == 1:
+            total_model = []
+            for func, data in izip(modelfuncs, self.datasets):
+                tmp_model = data.eval_model_to_fit(func)
+                total_model.append(tmp_model)
+            return numpy.concatenate(total_model)
+        else:
+        # best to make this a different derived class
+            funcs = []
+            datasets = []
+            for func, data in izip(modelfuncs, self.datasets):
+                funcs.append(func)
+                datasets.append(data.get_indep(filter=False))
+            total_model = parallel_map_funcs(funcs, datasets)
+            all_model = []
+            for model, data in izip(total_model, self.datasets):
+                all_model.append(data.apply_filter(model))
+            return numpy.concatenate(all_model)
+>>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 
     def to_fit(self, staterrfunc=None):
         total_dep = []
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -1,11 +1,6 @@
 #
-<<<<<<< HEAD
 #  Copyright (C) 2008, 2015, 2016, 2017, 2019, 2020
 #     Smithsonian Astrophysical Observatory
-=======
-#  Copyright (C) 2008, 2015, 2016, 2017, 2018
-#             Smithsonian Astrophysical Observatory
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -38,7 +33,6 @@ from sherpa.utils import SherpaFloat, NoNewAttributesAfterInit, \
     filter_bins, parallel_map_funcs
 
 
-<<<<<<< HEAD
 __all__ = ('Data', 'DataSimulFit', 'Data1D', 'Data1DInt',
            'Data1DAsymmetricErrs', 'Data2D', 'Data2DInt')
 
@@ -95,9 +89,6 @@ class DataSpace1D(EvaluationSpace1D):
             return self
 
         data = self.grid[0]
-=======
-__all__ = ('Data', 'DataSimulFit', 'Data1D', 'Data1DInt', 'Data2D', 'Data2DInt')
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 
         data = self.filter.apply(data)
         return DataSpace1D(self.filter, data)
@@ -875,29 +866,18 @@ class DataSimulFit(NoNewAttributesAfterInit):
 
     """
 
-<<<<<<< HEAD
-    def __init__(self, name, datasets):
-        self.name = name
-=======
     def __init__(self, name, datasets, numcores=1):
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
         if len(datasets) == 0:
             raise DataErr('zerodatasimulfit', type(self).__name__)
+        self.name = name
         self.datasets = tuple(datasets)
+        self.numcores = numcores
         NoNewAttributesAfterInit.__init__(self)
 
     def eval_model_to_fit(self, modelfuncs):
-<<<<<<< HEAD
-        total_model = []
-
-        for func, data in zip(modelfuncs, self.datasets):
-            total_model.append(data.eval_model_to_fit(func))
-
-        return numpy.concatenate(total_model)
-=======
         if self.numcores == 1:
             total_model = []
-            for func, data in izip(modelfuncs, self.datasets):
+            for func, data in zip(modelfuncs, self.datasets):
                 tmp_model = data.eval_model_to_fit(func)
                 total_model.append(tmp_model)
             return numpy.concatenate(total_model)
@@ -905,15 +885,14 @@ class DataSimulFit(NoNewAttributesAfterInit):
         # best to make this a different derived class
             funcs = []
             datasets = []
-            for func, data in izip(modelfuncs, self.datasets):
+            for func, data in zip(modelfuncs, self.datasets):
                 funcs.append(func)
                 datasets.append(data.get_indep(filter=False))
             total_model = parallel_map_funcs(funcs, datasets)
             all_model = []
-            for model, data in izip(total_model, self.datasets):
+            for model, data in zip(total_model, self.datasets):
                 all_model.append(data.apply_filter(model))
             return numpy.concatenate(all_model)
->>>>>>> an initial release of simultaneous fit on multicores (slower for most, ie a lot, of cases :)
 
     def to_fit(self, staterrfunc=None):
         total_dep = []
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -888,7 +888,7 @@ class DataSimulFit(NoNewAttributesAfterInit):
             for func, data in zip(modelfuncs, self.datasets):
                 funcs.append(func)
                 datasets.append(data.get_indep(filter=False))
-            total_model = parallel_map_funcs(funcs, datasets)
+            total_model = parallel_map_funcs(funcs, datasets, self.numcores)
             all_model = []
             for model, data in zip(total_model, self.datasets):
                 all_model.append(data.apply_filter(model))
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2737,6 +2737,7 @@ def parallel_map(function, sequence, numcores=None):
 
     return run_tasks(procs, err_q, out_q, numcores)
 
+
 def parallel_map_funcs(funcs, datasets, numcores=None):
     """Run a sequence of function on a sequence of inputs in parallel.
 
@@ -2754,7 +2755,6 @@ def parallel_map_funcs(funcs, datasets, numcores=None):
     datasets : a list or tuple of array_like
        The data to be passed to ``func``. The number of elements in
        datasets must match the number of elements of funcs.
-
     numcores : int or None, optional
        The number of calls to ``funcs`` to run in parallel. When
        set to ``None``, all the available CPUs on the machine - as
@@ -2770,15 +2770,14 @@ def parallel_map_funcs(funcs, datasets, numcores=None):
 
     Notes
     -----
-    Due to the overhead involve in passing the functions and datasets
+    Due to the overhead involved in passing the functions and datasets
     to the different cores, the functions should be very time consuming
-    to compute.  A similar requirement (the function is expected to be
-    run on computations that take a significant amount of time to run.)
-    to the the parallel_map function
-
-    An ordered iterable (ie tuple or list) should be used to pass multiple
+    to compute (of order 0.1-1s).  This is similar to the ``parallel_map``
+    function.
+    
+    An ordered iterable (i.e. tuple or list) should be used to pass multiple
     values to the multiple functions. The lengths of the iterable funcs and
-    datasets must be equal. The coressponding funcs and datasets are passed
+    datasets must be equal. The corresponding funcs and datasets are passed
     to the different cores to distribute the work in parallel. There is no
     guarantee to the ordering of the tasks.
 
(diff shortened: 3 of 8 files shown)


dtnguyen2:
lowered the tolerance on the error of the error, but when I tried to push it back it gave me the following error, see below.  I tried to use the commands Omar suggested previously:

git reset --hard
git rebase @{u}
git push

to no avail

(sherpa3) [dtn@devel12 sherpa]$ git push myfork no-optmethods-Fortran
To https://github.com/dtnguyen2/sherpa
 ! [rejected]          no-optmethods-Fortran -> no-optmethods-Fortran (fetch first)
(sherpa3) [dtn@devel12 sherpa]$ git reset --hard
HEAD is now at 62aef807 Some tests are failing on 1 of the errors in the covar matrix, hoping relaxing tol should fix it
(sherpa3) [dtn@devel12 sherpa]$ git rebase @{u}
Current branch no-optmethods-Fortran is up to date.
(sherpa3) [dtn@devel12 sherpa]$ git push
To https://github.com/dtnguyen2/sherpa
 ! [rejected]          no-optmethods-Fortran -> no-optmethods-Fortran (fetch first)

sherpa/data.py
```diff
@@ -490,19 +490,31 @@ class DataSimulFit(Data):
 
     """
 
-    def __init__(self, name, datasets):
+    def __init__(self, name, datasets, numcores=1):
```
DougBurke: Given that `numcores` is only used in `eval_model_to_fit`, it looks like it would be better if it were an argument to `eval_model_to_fit` rather than an attribute of the `BaseData` object. Or does the "complicated" way we implement/use `BaseData` mean that this doesn't work?

dtnguyen2: For now, let's leave it the constructor as is for testing purposes. The default value for numcores=1; However, if you want to run with multicores, say 3, then run your script with
```
fit(numcores=3)
```
I tested this claim by adding a printing statement in the constructor of the DataSimulFit
```
(sherpa3) [dtn@devel12 sherpa]$ git diff sherpa/data.py
diff --git a/sherpa/data.py b/sherpa/data.py
index de2d0302..701c2eb2 100644
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -491,6 +491,7 @@ class DataSimulFit(Data):
     """
 
     def __init__(self, name, datasets, numcores=1):
+        print('DataSimulFit::__init__ numcores =', numcores)
         if len(datasets) == 0:
             raise DataErr('zerodatasimulfit', type(self).__name__)
         datasets = tuple(datasets)
```

(sherpa3) [dtn@devel12 integrate]$ python fit_3_spectra_simple.py | grep numbers
DataSimulFit::__init__ numcores = 1

Changing the script from fit() to fit(numcores=3)

(sherpa3) [dtn@devel12 integrate]$ diff fit_3_spectra_simple.py tmp.py
68d67
<
85c84
<     fit()
---
>     fit(numcores=3)

(sherpa3) [dtn@devel12 integrate]$ python tmp.py | grep numcores
DataSimulFit::__init__ numcores = 3

dtnguyen2: For fun, I moved the ```numcores``` from the constructor of the DataSimulFit class and moved it to the ```eval_model_to_fit```  as Doug suggested.  Unfortunately, the ```fit(numcores=3)``` has no effect.  Not a good idea in my opinion, I'm willing to move it but I suspect the user interface is not going to be pretty but that is just my opinion

DougBurke:
This is going to need rebasing before it is easy for us to test.

anetasie:
There are several cases of the simultaneous fit of multiple data sets:
1/ all data set ids have exactly the same model components and model expressions:
2/ each data set id has specific separate model components with different parameters and model expressions.
3/ some data sets have some common model components with the same parametrs,  but different model expressions.
In each case the multicore fits would need to be handled differently. Is this done in this PR?

dtnguyen2:
The enhancement in this PR was based on the script and data that you gave me which was taken from: ```/data/sherpa/TestIssues/SimulFit3```.  So I suspect it is case 1 only.

dtnguyen2:
For simultaneous fit, each data set has to have a model expression associated with it.  This PR just distribute the matched model expression with the data set to different cores, see the the function ```parallel_map_funcs(funcs, datasets, numcores=None):``` for details, so I think the three cases are covered.  BTW, I made _numcores=1_ as the default cause I think for most cases it is _not_ cost effective to distribute the work on multiple cores.

dtnguyen2:
Here is another data point for you:  Took the data/script from one of SDS regression tests, ie took: _/data/lenin2/Test/CIAO4.11/SherpaRegressionBeta1/45_
I commented out all sherpa's commands after the call to ```fit()``` cause I'm only interested in the time it takes do the simultaneous fit with/out use of multicores.

```
(sherpa) [dtn@devel12 cachebug]$ diff fit_4.py fit_4_numcores.py
39c39
< fit()
---
> fit(numcores=4)

(sherpa) [dtn@devel12 cachebug]$ time python fit_4.py > a

real	3m20.413s
user	3m20.227s
sys	0m0.162s

(sherpa) [dtn@devel12 cachebug]$ time python fit_4_numcores.py > c
/export/sherpa/sherpa/utils/__init__.py:496: VisibleDeprecationWarning: sctypeNA and typeNA will be removed in v1.18 of numpy
  v = '%s[%d]' % (numpy.typeNA[v.dtype.type], v.size)

real	3m13.753s
user	9m36.160s
sys	0m28.494s

(sherpa) [dtn@devel12 cachebug]$ diff a c | wc
      0       0       0
```

So we managed to reduce the computing time from ~3m20 to ~3m14 using 4 cores for the 4 data sets, not much of an improvement. 

```
(sherpa) [dtn@devel12 cachebug]$ grep nfev a
nfev           = 711
```

As time consuming as the script/data is ~.28secs per model evaluation for the 4 data sets using 4 cores, it is still enough to take advantage of the distributed computation.

dtnguyen2:
@hamogu, working on fixing the messed-up rebase

dtnguyen2:
just noticed that one test passed, so it should be ready by tomorrow

sherpa/data.py
```diff
@@ -866,20 +866,33 @@ class DataSimulFit(NoNewAttributesAfterInit):
...
+                total_model.append(tmp_model)
+            return numpy.concatenate(total_model)
+        else:
+        # best to make this a different derived class
+            funcs = []
+            datasets = []
+            for func, data in zip(modelfuncs, self.datasets):
+                funcs.append(func)
+                datasets.append(data.get_indep(filter=False))
+            total_model = parallel_map_funcs(funcs, datasets)
```
hamogu, line 891: I'm sure you thought about it, but I'm curious why you don't use `eval_model_to_fit` in this case, something like
```
for ...
    funcs.append(data.eval_model_to_fit)
    arg_for_funcs.append(func)
total_model = parallel_map_funcs(funcs, arg_for_funcs)
```
Since the two branches for 1 or more cores are different, I suggest a small comment in the code.

sherpa/utils/__init__.py
```diff
@@ -2736,6 +2737,56 @@ def parallel_map(function, sequence, numcores=None):
...
+
+    funcs_size = len(funcs)
+    datasets_size = len(datasets)
+    if funcs_size != datasets_size:
+        msg = "input funcs (%d) and datsets (%d) size must be same" % \
+            (funcs_size, datasets_size)
+        raise TypeError(msg)
+
+    if not _multi or datasets_size == 1 or \
+            (numcores is not None and numcores < 2):
```
hamogu, line 2760: Is there a usecase for negative numcores or are you just catching both 0 and 1 here?

sherpa/utils/tests/test_utils.py
```diff
@@ -388,3 +389,14 @@ def test_filter_bins_unordered():
     assert len(flags) == len(expected)
     for got, exp in zip(flags, expected):
         assert got == exp
+
+def test_paralle_map_funcs():
```
hamogu, line 393: typo "parallel"  ;-)

sherpa/utils/tests/test_utils.py
```diff
@@ -388,3 +389,14 @@ def test_filter_bins_unordered():
...
+    for arg in range(1, 5):
+        func = [numpy.sum]
+        funcs = arg * func
+        datas = [numpy.arange(1, 2+2*i) for i in range(arg)]
+        result = []
+        for func, data in zip(funcs, datas):
+            result.extend(numpy.asarray(list(map(func, data))))
+        assert_equal(result, utils.parallel_map_funcs(funcs, datas, arg))
+
```
hamogu, line 402: I think we need a "real" test, too, that uses all the sherpa machinery with setting up data, defining a model through the ui etc.

I know this is marked as WIP, so I understand it's just not here yet, but I'm just saying we'll need it eventually.

sherpa/utils/__init__.py
```diff
@@ -2736,6 +2737,56 @@ def parallel_map(function, sequence, numcores=None):
...
+    out_q = manager.Queue()
+    err_q = manager.Queue()
+    lock = manager.Lock()
+
+    # if datasets is less than numcores, only use len datasets number of
+    # processes
+    numcores = min(funcs_size, datasets_size)
+
+    procs = []
+    for ii, chunk in enumerate(datasets):
```
hamogu, line 2784: `for func, chunk in zip(funcs, datasets):` is functionally the same but makes more clear what's going on here.

sherpa/astro/ui/utils.py
```diff
@@ -9993,15 +9993,17 @@ def _fit(self, id=None, *otherids, **kwargs):
         # validate the kwds to f.fit() so user typos do not
         # result in regular fit
         # valid_keys = sherpa.utils.get_keyword_names(sherpa.fit.Fit.fit)
-        valid_keys = ('outfile', 'clobber', 'filter_nan', 'cache')
+        valid_keys = ('outfile', 'clobber', 'filter_nan', 'numcores')
         for key in kwargs.keys():
```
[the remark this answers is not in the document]
dtnguyen2: you are right

sherpa/data.py
dtnguyen2: The original ```parallel_map``` assumes the arguments are numpy arrays to do its magic, which is not the case here.

I don't have data and not very familiar with what models are _time consuming_ so I relied on SDS giving me, the one test which goes from 3:20 to 3:14secs using 4 cores is the one that @anetasie gave me.

BTW, if one wants to have a runtime option of setting ```numcores``` then do tell how that can be done, enjoy.

sherpa/utils/__init__.py
dtnguyen2: users are creative with their options

sherpa/astro/ui/utils.py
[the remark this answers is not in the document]
dtnguyen2: done

sherpa/utils/__init__.py
hamogu: So, how about `numcores == 1`? (I know you copied that just from the function above, but I still wonder why it's written this way.

sherpa/data.py
hamogu: I'm thinking hard a about a model I can try for this, hopefully I get an idea over the weekend.

hamogu:
The Travis failures looks unrelated, so I restarted that build. Let's see what happens.

anetasie:
I build @dtnguyen2 branch and run a test fitting three data sets using a composite model with non-overlapping components or parameters across the data sets. There are three types of runs for the timing comparison:

1/ simultaneously with default setting using `fit()`; 
2/ fit each of the data set using `fit(1), fit(2), fit(3)`; 
3/ fit with `fit(numcores=3)`.

The fit was performed with `Nelder-Mead` method, and `wstat` statistics

Summary Results:

1/ The default `fit()` option takes much longer than the fit split into 3 cores, 6051s compare to 3805s.

2/ Total time for the fit split into individual data sets `fit(1), fit(2), fit(3` is shorter than the simultaneous fit of the 3 data sets in (1), or with `fit(numcores=3)`.
The difference in  time is relatively large, a factor of 3 when summed individual  times `(462.914+355.90+229.52) = 1048.334  vs. 3805s` for the `numcores=3` option.


More details on the fit is given in a google doc: https://docs.google.com/document/d/1WdQW-7zxCKdwbYUDhyQ-W3KtUwG5gn9OIfcRk19igv4/edit?usp=sharing

dtnguyen2:
@anetasie, I'm in the process of writing the docs to the underlying function ```parallel_map_funcs``` to distribute fitting to the different cores for simultaneous fitting to satisfy @hamogu's request;  However, I would like to point out to you that the sherpa's ```parallel_map``` has the following blurb:

> .. the function is expected to be run on computations that take a significant amount of time to run....

As of right now, I have the following doc for ```parallel_map_funcs``` (work in progress, so I might change some wordings etc...)

>    Due to the overhead involve in passing the functions and datasets           
>   to the different cores, the functions should be very time consuming         
>    to compute.  A similar requirement (the function is expected to be          
>    run on computations that take a significant amount of time to run.)         
>    to the the parallel_map function                                            

That is, I don't consider the functions in your writeup very time consuming.  Don't worry, I will try to prove my claim with a concrete example.

dtnguyen2:
I misunderstood @anetasie report, so let's me see if I can figure out what was reported.
The default runtime option for numcores in the fit function is ```1```, so by typing ```fit()``` one is running on one core. Whereas ```fit(numcores=3)``` means the three datasets will be run on three cores. So @anetasie report does show that there is a benefit in running using multiple cores when simultaneous datasets using a composite model with non-overlapping components or parameters across the data sets:

> 1/ The default fit() option takes much longer than the fit split into 3 cores, 6051s compare to 3805s.

The results when running with the commands: ```fit()``` and ```fit(numcores=3)``` are _identical_.  
However, while the total time for ```fit(1), fit(2), fit(3)``` might be less then when one runs with ```fit(numcores=3)```, 1084secs vs 3805secs,  but the resulting fits are not identical.  For example for the simultaneous fit:
```
nhz1.nH        85.1347 
....
nhz2.nH        109.144 
...
nhz3.nH        62.1531
```

versus:

for ```fit(1)```
```
nhz1.nH        86.5598
```
for ```fit(2)```:
```
nhz2.nH        102.49 
```
for ```fit(3)```:
```
nhz3.nH        63.5885
```
so I'm not sure what to make of the shorter running time if the fitted parameters are different.

anetasie:
@dtnguyen2 I guess the difference in the final parameter values is a question for the optimizer. Is there a difference between the optimization search in case of `fit()` vs. `fit(1)`, `fit(2)`, `fit(3)`. Why do we get a different parameter value?

anetasie:
Also, it looks like the total fit statistics in the individual `fit(1), fit(2), fit(3)` sums to `5.79412+15.3907+59.0306=80.21542` lower value than in the simultaneous fit of `80.3306`.
So, in this case the parameters could be different.

sherpa/utils/__init__.py
dtnguyen2: I experimented with @hamogu's suggestion:

```
@@ -2757,7 +2835,7 @@ def parallel_map_funcs(funcs, datasets, numcores=None):
         raise TypeError(msg)
 
     if not _multi or datasets_size == 1 or \
-            (numcores is not None and numcores < 2):
+            (numcores is not None and numcores == 1):
         return list(map(funcs[0], datasets))
```
I'm using the script and data that @anetasie gave me, the one that makes life more miserable for the user if using multicores rather than a single core because of the large overhead in multi-processing:

```
(sherpa) [dtn@devel12 sherpa]$ python fit_3_spectra.py.orig 1
...
fit with numcores == 1 took 24.588085412979126 secs

(sherpa) [dtn@devel12 sherpa]$ python fit_3_spectra.py.orig 3
...
fit with numcores == 3 took 53.463428020477295 secs
```
If the code is left unchanged: only one core is used; However, if the change as suggested by @hamogu the code runs with multiple cores (don't ask me how many cores were used):

```
(sherpa) [dtn@devel12 sherpa]$ python fit_3_spectra.py.orig -1
...
fit with numcores == -1 took 54.06742238998413 secs

(sherpa) [dtn@devel12 sherpa]$ python fit_3_spectra.py.orig 0
...
fit with numcores == 0 took 53.430519342422485 secs
```

So if ```numcores == 1``` is to be substituted for ```numcores < 2```, perhaps one should also add in the following check:

```
if numcores <= 0:
    raise ValueError('numcores must be > 0')
```

sherpa/data.py
dtnguyen2: BTW,  a good thing I looked at this comment cause the prototype for ```parallel_map_funcs``` is:
```
def parallel_map_funcs(funcs, datasets, numcores=None):
```

so the call to ```parallel_map_funcs``` as written, see comment above, was missing the ```numcores``` option so not sure how this ever worked, yikes!   Anyway, when I tried @hamogu's suggestion:

```
(sherpa) [dtn@devel12 sherpa]$ git diff sherpa/data.py
diff --git a/sherpa/data.py b/sherpa/data.py
index b9af0528..5c9faa6a 100644
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -885,10 +885,14 @@ class DataSimulFit(NoNewAttributesAfterInit):
         # best to make this a different derived class
             funcs = []
             datasets = []
+            arg_for_funcs = []
             for func, data in zip(modelfuncs, self.datasets):
-                funcs.append(func)
-                datasets.append(data.get_indep(filter=False))
-            total_model = parallel_map_funcs(funcs, datasets)
+                funcs.append(data.eval_model_to_fit)
+                arg_for_funcs.append(func)
+            total_model = parallel_map_funcs(funcs, arg_for_funcs, self.numcores)
+            #     funcs.append(func)
+            #     datasets.append(data.get_indep(filter=False))
+            # total_model = parallel_map_funcs(funcs, datasets, self.numcores)
             all_model = []
             for model, data in zip(total_model, self.datasets):
                 all_model.append(data.apply_filter(model))
```
I get:
```
(sherpa) [dtn@devel12 PR483]$ fit_3_spectra.py 3
Process Process-2:
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/export/sherpa/sherpa/utils/__init__.py", line 2590, in worker
    out_q.put((ii, list(vals)))
  File "/export/sherpa/sherpa/astro/data.py", line 1663, in eval_model_to_fit
    return self.apply_filter(modelfunc(*self.get_indep(filter=True)))
  File "/export/sherpa/sherpa/astro/data.py", line 1221, in apply_filter
    elif len(data) != len(self.counts):
TypeError: object of type 'numpy.float64' has no len()
Process Process-3:
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/export/sherpa/sherpa/utils/__init__.py", line 2590, in worker
    out_q.put((ii, list(vals)))
  File "/export/sherpa/sherpa/astro/data.py", line 1663, in eval_model_to_fit
    return self.apply_filter(modelfunc(*self.get_indep(filter=True)))
  File "/export/sherpa/sherpa/astro/data.py", line 1221, in apply_filter
    elif len(data) != len(self.counts):
TypeError: object of type 'numpy.float64' has no len()
Process Process-4:
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/export/miniconda/envs/sherpa/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/export/sherpa/sherpa/utils/__init__.py", line 2590, in worker
    out_q.put((ii, list(vals)))
  File "/export/sherpa/sherpa/astro/data.py", line 1663, in eval_model_to_fit
    return self.apply_filter(modelfunc(*self.get_indep(filter=True)))
  File "/export/sherpa/sherpa/astro/data.py", line 1221, in apply_filter
    elif len(data) != len(self.counts):
TypeError: object of type 'numpy.float64' has no len()
  File "./fit_3_spectra.py.orig", line 136, in <module>
    tst(numcores)
  File "./fit_3_spectra.py.orig", line 127, in tst
    fit(1, 2, 3, numcores=numcores)
  File "<string>", line 1, in fit
  File "/export/sherpa/sherpa/astro/ui/utils.py", line 9907, in fit
    self._fit(id, *otherids, **kwargs)
  File "/export/sherpa/sherpa/astro/ui/utils.py", line 10013, in _fit
    res = f.fit(**kwargs)
  File "/export/sherpa/sherpa/fit.py", line 56, in run
    result = func(fit, *args, **kwargs)
  File "/export/sherpa/sherpa/fit.py", line 1231, in fit
    init_stat = self.calc_stat()
  File "/export/sherpa/sherpa/fit.py", line 1116, in calc_stat
    return self._calc_stat()[0]
  File "/export/sherpa/sherpa/fit.py", line 1097, in _calc_stat
  File "/export/sherpa/sherpa/stats/__init__.py", line 916, in calc_stat
    data_model = data.eval_model_to_fit(model)
  File "/export/sherpa/sherpa/data.py", line 892, in eval_model_to_fit
    total_model = parallel_map_funcs(funcs, arg_for_funcs, self.numcores)
  File "/export/sherpa/sherpa/utils/__init__.py", line 2869, in parallel_map_funcs
    return run_tasks(procs, err_q, out_q, numcores)
  File "/export/sherpa/sherpa/utils/__init__.py", line 2623, in run_tasks
    vals.extend(r)
TypeError: 'NoneType' object is not iterable
```
 Will stick with the original code, apart from the additional call to pass ```numcores```.

sherpa/utils/__init__.py
hamogu: I guess 0 has some special meaning (maybe "default"). Anyway, it was just a suggestion. IF you have good reason to do it this way, I'm fine with it, too.

hamogu:
I can confirm that this approach works. I used the following model:
```
class SlowGauss(sherpa.models.basic.Gauss1D):
    def calc(self, **args, **kwargs):
        time.sleep(.1)
        return super().calc(*args, **kwargs)
```
and made just a couple very simple datasets. Then, I vary the time in sleep. I find that for my data and for my machine, you start to save with the multicore option once the model takes about 0.1 s to evaluate on a single dataset. I'm sure the absolute number depends on the number of datapoints etc, but as an order-of-magnitude that's probably OK.

So, to move this forward, I suggest we try to tie up the loose ends in the discussion above and merge this. Do I think that this is the best we can do to make sherpa parallel? Certainly not. But is it an improvement over what we have now? Yes, for certain special cases - and thus, it should be merged when ready. We can always make a better implementation later.

There is a lot of in-line discussion above, I'll try to summarize the points that, in my opinion, still need to be done before we merge this:

- [ ] Like Doug (see above) I'm unhappy with the `numcore` option living in `DataSimulFit`. To me, that's not a data option, it's a fit option, so it should logically live in `sherpa.fit.Fit`. (In fact, when I wrote my test script, I had to get into the code here to find how to make it because I intuitively expected this to be a fit parameter and not a data parameter.). Is there any way we can move it there? If there is no way to do that, we can merge now and possibly change that later. This is going to be a little used feature. I'm not worried about breaking backwards compatibility later.

- [ ] This needs docs, so the user can find out about the option. Since I'm new to the process @anetasie has to tell me: Are we adding it to the .rst files in `/doc/` as part of PR or is SDS (we) going to write those in a separate PR? We should say that "around 0.1 s" is where this starts to help.

- [ ] We need a real test with Data Objects etc. The test just need to check that everything runs, not that it actually saves time.

- [ ] The typo "parallel" that I marked inline should be fixed ;-)

hamogu:
Here is the full script I used for testing. Just modify the number in sleep and numcores before and run again:
```
from time import sleep
from datetime import datetime
import numpy as np
from sherpa.data import Data1D, DataSimulFit
from sherpa.models.basic import Gauss1D
from sherpa.models import SimulFitModel
from sherpa.optmethods import LevMar
from sherpa.fit import Fit

x = np.linspace(-5, 5, 200)


def cheap_gauss(x, x0, sigma):
    y = np.exp(-0.5 * ((x - x0)**2) / sigma**2)
    return y + np.random.normal(0, 0.2, x.shape)

d1 = Data1D('1', x, cheap_gauss(x, 1., .8))
d2 = Data1D('2', x, cheap_gauss(x, 1.2, .7))
d3 = Data1D('3', x, cheap_gauss(x, .8, .7))


class SlowGauss(Gauss1D):
        sleep(.1)
        return super().calc(*args, **kwargs)


g1 = SlowGauss()
g2 = SlowGauss()
g3 = SlowGauss()

sdata = DataSimulFit('all', (d1, d2, d3), numcores=1)
smodel = SimulFitModel('all', (g1, g2, g3))
sfit = Fit(sdata, smodel, LeastSq(), LevMar())

start = datetime.now()
res = sfit.fit()
print(datetime.now() - start)

print(res)
```

hamogu:
And, similar to @anetasie said above: I do see slightly different fit results even when using the same script bunt since add some randomness to my data that's expected. Fit results in both cases are OK.

dtnguyen2:
Using the data and script that I have:
```
(sherpa) [dtn@devel12 PR483]$ time fit_3_spectra.py.orig 1 > out.txt
fit with numcores == 1 took 24.51454472541809 secs

real	0m26.191s
user	0m25.940s
sys	0m0.204s

(sherpa) [dtn@devel12 PR483]$ time fit_3_spectra.py.orig 3 > out.txt
fit with numcores == 3 took 54.90108942985535 secs

real	0m56.466s
user	2m9.710s
sys	0m7.679s
```

So the extra 30 seconds for the three datasets (with 76 data points) is the overhead which means that the model function has to be more time consuming then ~.22 secs to see some benefits from the using ```parallel_map_funcs```

```
(sherpa) [dtn@devel12 PR483]$ grep function out.txt 
Final fit statistic   = 90.1923 at function evaluation 135
(sherpa) [dtn@devel12 PR483]$ grep points out.txt 
Data points           = 76
```

>>> 30. / 135.
0.2222222222222222

hamogu:
I already got a benefit with 0.1 s (30 s with numcores=1 and 15 s with numcores=3), but I'm not surprised that the break-even point depends on memory, CPU type, concurrent processes etc. and differes by a factor of a few. 
E.g. if you do numcores=1 but only have one physical core (even if hytherthreading etc. makes it look to the OS as if there were multiple cores), then you pay all the overhead for little of the benefit (which will make the break-even point come at longer times). I tested on a machine running little else at the time with 2 sockets and 6 cores per socket, each of them hyperhreaded into 2 virtual CPUs.

(I don't think we need to go any further in the analysis here, my point is that the docs should say that this should not be used if a single call the the model is significantly faster than one second).

anetasie:
> This needs docs, so the user can find out about the option. Since I'm new to the process @anetasie has to tell me: Are we adding it to the .rst files in /doc/ as part of PR or is SDS (we) going to write those in a separate PR? We should say that "around 0.1 s" is where this starts to help.

Yes, we need to update the docs ,rst files when this is merged. The question is where to include this information?

hamogu:
I would add a sentence here: https://sherpa.readthedocs.io/en/latest/fit/index.html#simultaneous-fits

Eventually, I might want a separate page on performance with several options and a more general discussion, but since this is of fairly limited applicability (most or models are way faster than 0.1 s) I think a short sentence is enough for now.

anetasie:
4.12 release had an update to optimization methods, `levmar` and `moncar` to set the `numcores`  in the opt method (see PR #596). I'm wondering  how the `numcores` setting in the method  via `set_method_opt` relates to the new setting of numcores in `fit`. Is there any overhead/difference/advantage of setting the numcores in methods?

sherpa/utils/__init__.py
```diff
@@ -2736,6 +2737,110 @@ def parallel_map(function, sequence, numcores=None):
...
+       The return values from the calls, in the same order as the
+       ``sequence`` array.
+
+    Notes
+    -----
+    Due to the overhead involve in passing the functions and datasets
+    to the different cores, the functions should be very time consuming
+    to compute.  A similar requirement (the function is expected to be
+    run on computations that take a significant amount of time to run.)
+    to the the parallel_map function
```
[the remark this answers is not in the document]
hamogu: Fixed by direct commit.

dtnguyen2:
> 4.12 release had an update to optimization methods, levmar and moncar to set the numcores in the opt method (see PR #596). I'm wondering how the numcores setting in the method via set_method_opt relates to the new setting of numcores in fit. Is there any overhead/difference/advantage of setting the numcores in methods?

The opt methods ```LevMar``` and ```MonCar``` do have the runtime option ```numcores``` (default is 1)  and the optimization algorithm will use ```parallel_map``` which does come with an overhead.  I had sent a warning of this in the following mail: [lmdifwithparallel_map](https://icxc.harvard.edu/pipe/sherpadev/2019/0038.html)

The setting of ```numcores``` with the ```set_method_opt``` is unrelated with the current PR (setting ```numcores``` with the call to ```fit()```).  In this PR, the setting of ```numcores``` only makes sense if one is doing simultaneous fitting of more than one data set; Whereas, the setting of ```numcores``` with the ```set_method_opt```  can be used with a single data set, of course the caveat about time consuming model function still applies.  In theory, although I have never tested it: one should be able to set ```numcores``` with the ```set_method_opt``` and ```fit``` commands if one is using one of the aforementioned optimization algorithms to simultaneous datasets.  Just like one could set ```numcores``` with the ```set_method_opt``` while running ```conf```

hamogu:
@anetasie Since this is otherwise ready to merge, do you want to add docs or do you want me to either open a PR to @dtnguyen2 branch, or , for simplicity, add rst docs to this branch directly?

PR #484, opened 2018-08-23
Equivalent Width errors

dtnguyen2:
# Release Note
Several optional arguments (params=None, error=False, otherids=(), niter=1000) , were added to eqwidth. If error = True, then get_draws shall be run if the fit stat is one of the following: (Cash, CStat, WStat) otherwise a multivariate normal distribution shall be run to generate the samples. The optional niter parameter is used to generate the number of samples. The optional parameter otherids is only used if get_draws is used internally when multiple data sets were used to fit. Alternatively, the user can enter the samples via the params option where the samples must be a numpy.ndarray of dimension (num, npar).


anetasie:
@dtnguyen2 @olaurino `get_draws` that is used in calculating error on `eqwidth` can only be used with likelihood statistic, so I get a Value Error for any other stat:
```
ValueError: Fit statistic must be cash, cstat or wstat, not chi2gehrels
```

For `chi2` statistics the sampling should be run using a multivariate normal

anetasie:
@dtnguyen2 for the UI consistency with the eqwidth() behaviour, we should print a value of the EW with error to the screen. Now this information is given at the top of the printout with the full parameter arrays printed later. I suggest to print the EW and errors at the bottom of the full array, so it stays on the screen for easy reading.  

Currently:

```
eqwidth(p1,p1+g1) 
```
 prints to the screen just one value

```
eqwidth(p1,p1+g1,error=True)
```
print the EW and error and then the full parameter arrays.
```
ew=eqwidth(p1,p1+g1,error=True)
```
prints to the screen 'get_draws' information (@DougBurke I'm wondering if we really need that) and no EW information which is somewhat confusing.

dtnguyen2:
Here is the requirements for eqwidth:

https://docs.google.com/document/d/1e4l88tevfwUHWubaGwRaRadBLOahKacuI3UYYRZVmFY/edit

I met with Aneta yesterday and explained to her that the output of eqwidth, along with the docs, and she agrees that the output should not be changed;  However, Aneta wants me to modify eqwdith to handle the case when the stat used in the fit is from the Chi-family (I should add that this was not in the requirement doc).

anetasie:
@dtnguyen2 I just updated the spec document to include a note on chi2 stat.

dtnguyen2:
Currently, if the fit stat is one of (Cash, CStat, WStat) then eqwidth calls get_draws to get the sampe parameters, if the stat is from one of the ChiSqr family then the multivariate normal distributions, ie NormalParameterSampleFromScaleMatrix, is used to get the sample parameters.  The question is what to do if it is a user-Stat?  This is not specified so eqwidth just throw an exception.  Let me know what has to be done here.

anetasie:
@dtnguyen2 I suggest NormalParameterSampleFromScaleMatrix.

dtnguyen2:
Just had a phone conversation with @anetasie: eqwidth shall use get_draws if fit stat is one of the following: Cash, CStat or WStat else the multivariate normal dist shall be used.  The recently checked in code should have the aforementioned feature.

anetasie:
@dtnguyen2 The simulations for calculating errors work ok but there is an issues with the model parameters that are obtained with `get_model' after the simulations. They are not set to the original fitted values which 'get_fit-results()` shows. It looks like they are being reset to incorrect values from the simulations array.

anetasie:
@dtnguyen2 here is an example of the script that gives incorrect parameter setting after the simulations for eqwidth:

```
from sherpa.astro.ui import *
import matplotlib
%matplotlib
import numpy as np

load_pha("3c273.pi")
notice(0.5,7.0)
set_stat("chi2datavar")
set_method("simplex")

set_model('powlaw1d.p1+gauss1d.g1')
g1.fwhm=0.1
g1.pos=2.0
freeze(g1.pos,g1.fwhm)
fit()
print(get_fit_results())

par_orig = get_fit_results().parvals
np.random.seed( 2345 )
eqwidth(p1,p1+g1)
e=eqwidth(p1,p1+g1,error=True, niter=100)
```

This is the screen output, where the thawed parameter values in get_mode() are set incorrectly
```
In [24]: get_fit_results().parvals
Out[24]: (1.90552729021528, 0.00017387966749774392, 1.2794150760589706e-05)


print(get_model())
apply_rmf(apply_arf((38564.608926889 * (powlaw1d.p1 + gauss1d.g1))))
   p1.gamma     thawed   3.8141e-05          -10           10           
   p1.ref       frozen            1 -3.40282e+38  3.40282e+38           
   p1.ampl      thawed  2.11123e-05            0  3.40282e+38           
   g1.fwhm      frozen          0.1  1.17549e-38  3.40282e+38           
   g1.pos       frozen            2 -3.40282e+38  3.40282e+38           
   g1.ampl      thawed  2.09846e-06 -3.40282e+38  3.40282e+38           

```

anetasie:
@dtnguyen2 btw, for `cstat,cash, wstat` - the parameter setting in `get_model` after the simulations is OK, so the issue is only with the `chi2` cases.

anetasie:
@dtnguyen2 This update worked!

anetasie:
@dtnguyen2 I was just looking at the documentation part. We need to add that sampling from multivariate normal is performed in the case of chi2 statistics.

dtnguyen2:
Technically, the sampling is from multivariate normal if stat is not one of the following : [Cash, CStat or WStat] since the statistic could be a user defined one.  However, let me know exactly what you want to say for the doc and I will update it.  Thanks in advance

dtnguyen2:
As requested by @olaurino, the release Notes for eqwidth:

Several optional arguments (params=None, error=False, otherids=(), niter=1000) , were added to eqwidth.  If error = True, then get_draws shall be run if the fit stat is one of the following: (Cash, CStat, WStat) otherwise a multivariate normal distribution shall be run to generate the samples.  The optional niter parameter is used to generate the number of samples.  The optional parameter otherids is only used if get_draws is used internally when multiple data sets were used to fit.  Alternatively, the user can enter the samples via the params option where the samples must be a numpy.ndarray of dimension (num, npar).

anetasie:
@dtnguyen2 The text looks good.

olaurino:
Rebasing to resolve conflicts, was 0b18449

PR #485, opened 2018-08-29
bug was fixed in crates, remove obsolete skips and tests

olaurino:
This PR simply removes some tests and skip conditions that were required because of a bug in pycrates that prevented some files to be read with python 3. As that bug is now fixed, and this resulted in an ad-hoc test to fail in CIAO, I am removing the failing test and the skip conditions that prevented other tests to run within CIAO.


olaurino:
Note the bug was originally mentioned here: https://github.com/sherpa/sherpa/pull/412#issuecomment-348769088

PR #486, opened 2018-08-30
Fix #476 - `ZeroDivisionError` in `calc_stat_info`

### LINKED ISSUE #476: Should this ZeroDivisionError be caught in calc_stat_info?
From a CIAO helpdesk ticket (number 21329) ... the user has a dataset, where after grouping there is only 1 bin.  With their model they have DOF=0.  When they do a `calc_stat_info` it returns a raw `ZeroDivisionError` exception.  Should this maybe be caught and the stats set to `NaN` ? 

Here's a mock'ed up example --

```python
>>> foo = np.arange(100)
>>> goo = np.zeros(100)
>>> goo[10]=10
>>> load_arrays(1, foo, goo, DataPHA)
>>> group_counts(1, 100)
>>> set_full_model(1,const1d.fudge)
>>> fit()
Dataset               = 1
Initial fit statistic = 442.443
Final fit statistic   = 2.08554e-29 at function evaluation 5
Data points           = 1
Degrees of freedom    = 0
Reduced statistic     = nan
Change in statistic   = 442.443
   fudge.c0       0.1         
```

So far, so good.  The `Reduces statistic` is `nan`, but then when they try:

```python
>>> calc_stat_info()
ZeroDivisionError: float division by zero
```
Should this exception be caught presumably like in the `fit()` output ? 

(Above is from sherpa in CIAO 4.10 python3, Linux64)

olaurino:
# Release Note
Sherpa did not catch divisions by zero in the `calc_stat_info` command. That has been fixed.

# Notes
This PR fixes #476.

I am no comfortable with the way the UI function is implemented, as it should be using the internal machinery rather than performing the calculation. The whole reason why the bug was there is that changes to the way the statistic is calculated are not reflected in the UI function `calc_stat_info`. However, this is a big change and it might not be worthwhile to do it now, unless @kglotfelty feels strongly we should.

### DIFF
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -21,11 +21,14 @@
 There is a copy of this test in sherpa/ui/astro/tests/ that needs
 to be kept up to date.
 """
+import numpy
 
+from sherpa.astro.data import DataPHA
 from sherpa.utils.testing import requires_plotting
 from sherpa.ui.utils import Session
+from sherpa.astro.ui.utils import Session as AstroSession
 from numpy.testing import assert_array_equal
-from sherpa.models import parameter
+from sherpa.models import parameter, Const1D
 
 import pytest
 
@@ -248,3 +251,22 @@ def test_list_model_ids():
 
     s.set_model('const1d.mdla')
     assert set(s.list_model_ids()) == set([1, 'ma', 'mb'])
+
+
+# Fix 476
+def test_zero_division_calc_stat():
+    ui = AstroSession()
+    x = numpy.arange(100)
+    y = numpy.zeros(100)
+    ui.load_arrays(1, x, y, DataPHA)
+    ui.group_counts(1, 100)
+    ui.set_full_model(1, Const1D("const"))
+
+    # in principle I wouldn't need to call calc_stat_info(), I could just
+    # use _get_stat_info to reproduce the issue, However, _get_stat_info is not a public
+    # method, so I want to double check that calc_stat_info does not throw an exception.
+    # So, first we try to run calc_stat_info and make sure there are no exceptions.
+    # Then, since calc_stat_info only logs something and doesn't return anything, we use
+    # a white box approach to get the result from _get_stat_info.
+    ui.calc_stat_info()
+    assert ui._get_stat_info()[0].rstat is numpy.nan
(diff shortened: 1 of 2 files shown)


olaurino:
By the way, at first I tried to user `numpy`'s divide method to deal with the possible issues including division by zero, but that would have returned `inf` rather than `nan`, so I decided to stick to the `try`/`catch` block.

sherpa/ui/tests/test_session.py
```diff
@@ -248,3 +251,22 @@ def test_list_model_ids():
 
     s.set_model('const1d.mdla')
     assert set(s.list_model_ids()) == set([1, 'ma', 'mb'])
+
+
+# Fix 476
+def test_zero_division_calc_stat():
+    ui = AstroSession()
+    x = numpy.arange(100)
```
kglotfelty: Interesting that haven't needed to `import numpy` before adding this test.  If you wanted to avoid it you can also just do

```python
x = list(range(100))
y = [0]*100
```

[That is lists' work too.]

olaurino: By the time the test is run numpy has been probably imported so many times already, I don't think it matters really.

olaurino:
@kglotfelty do you think I can merge this? Thanks!

kglotfelty:
I've not been indoctrinated into whole review process so I'm not sure what the 
formal procedure is ... but yes, this code change looks OK to me.  I haven't 
done any testing/independent verification -- just reviewed the diffs.

olaurino:
That's good enough for me. I know you'd test things in CIAO anyway. Thanks!

DougBurke:
@olaurino - sorry to comment on a closed pr, but do we need an issue for your concerns about the `calc_stat_info` implementation, so we don't forget about it?

olaurino:
Possibly :-)

olaurino:
Actually, I really don't know. The problem is more general than that single function, as there are more functions that have too much logic in the session "facade" (at least, it should be a facade, but it's more like a [God Object](https://en.wikipedia.org/wiki/God_object)). So no, it's a known problem, sooner or later we should decide whether we want to fix it or leave it be.

DougBurke:
Outside of the session object, there are other places where we hard-code
knowledge which probably shouldn't be (e.g. checking for instances of
subclasses of sherpa.stats.Stat to determine behavior).

PR #487, opened 2018-09-04
PSF bin size warning

olaurino:
# Release Note
Sherpa assumes that the PSF image and the data have the same pixel size. When this is not true Sherpa ignores the difference, which results in a larger PSF being applied. From now on Sherpa will issue a warning when the PSF image pixel size and the data image pixel size are different.

# Note
It looks like we are not going to be able to implement the whole PSF rebinning logic into 4.10.1. This PR at least makes sure that a warning is issued when the PSF image and the data image do not have the same binning size.

### DIFF
--- a/sherpa/instrument.py
+++ b/sherpa/instrument.py
@@ -16,6 +16,7 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+import warnings
 
 from six import string_types
 
@@ -550,6 +551,8 @@ class PSFModel(Model):
     def fold(self, data):
         # FIXME how will we know the native dimensionality of the
         # raveled model without the values?
+        self._check_pixel_size(data)
+
         kargs = {}
 
         kshape = None
@@ -697,7 +700,6 @@ class PSFModel(Model):
 
         indep, dep, kshape, lo, hi = self._get_kernel_data(data, subkernel)
 
-        dataset = None
         ndim = len(kshape)
         if ndim == 1:
             dataset = Data1D('kernel', indep[0], dep)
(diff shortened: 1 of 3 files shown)


anetasie:
@olaurino I do see a warning when I use a differently binned psf image. Sherpa also performs the fit
and issues the warning after the fit. I'm wondering if we should just stop fit and error out when the sizes do not match.

juramaga:
Do we really want to error out when the sizes do not match? I think a warning might be enough. I can think of some users who might want to assess PSF issues in their results by deliberately use a PSF that does not match the image pixel size.

anetasie:
@juramaga I agree that in the future when sherpa allows for the use of different size binning a warning or some other msg instead of the error can be used. Current Sherpa code assumes that the psf and image bins match (even if they don't) and the convolution proceeds with mismatch bins, so the result is incorrect and it cannot be used for a test you have suggested.

olaurino:
I had implemented this as it was listed in the "Arbitrary Grids" document signed off by SDS, i.e. SH-54: 

> Sherpa assumes the pixel size to be the same as the image, which is wrong.
> For the time being we will just issue a warning when the pixel sizes do not match.

Before I change the implementation, I'd like to note that I fear this might be more complicated than expected, and we ran out of time. The code that currently checks the pixel sizes match is called every time the model is "folded", which is simple enough. To selectively error out only when a fit is performed (i.e. not when a model is evaluated outside of a fit, by the user or for e.g. calculating the statistic) would require more work and testing.

The agreement was we would have just a stop-gap measure to inform the user something is off in case we couldn't provide the full implementation of the PSF rebinning, which is currently the case.

@anetasie please confirm you want me to change the requirement in the requirements document and the code so it errors out during a fit when the PSF and image bin sizes do not match.

anetasie:
@olaurino I guess we will leave the warning, which is as you said less invasive at this moment. It is better than the current silent behavior. I will approve this PR as is.

DougBurke:
The test failures are pretty easy to fix - and come from a change I made in the sphinx documentation. In `sherpa/models/tests/test_psf_rebin.py` change

```
from sherpa.utils import requires_data, requires_fits
```

to

```
from sherpa.utils.testing import requires_data, requires_fits
```

PR #489, opened 2018-09-10
Update ds9

olaurino:
Utility PR to update the version of ds9, as 7.5 was removed from the web pages. Some PRs already have the changes, other will be rebased as necessary. I am going to merge this PR as long as Travis is OK. No review needed.

### DIFF
--- a/travis/setup_xspec_ds9.sh
+++ b/travis/setup_xspec_ds9.sh
@@ -11,7 +11,7 @@ then
     sudo apt-get install -qq libwcs4 wcslib-dev libx11-dev libsm-dev libxrender-dev
 
     # set os-specific variables
-    ds9_os=centos6
+    ds9_os=ubuntu14
     xspec_root=$miniconda/envs/build/Xspec/x86_64-unknown-linux-gnu-libc2.15-0
 
     # Newer versions (or conda builds) of numpy bring in a libgfortran-ng dependency, which our
--- a/travis/setup_xspec_ds9.sh
+++ b/travis/setup_xspec_ds9.sh
@@ -36,7 +36,7 @@ fi
 
 ### DS9 and XPA
 # Tarballs to fetch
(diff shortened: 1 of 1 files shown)


DougBurke:
Is it worth adding

https://github.com/sherpa/sherpa/pull/302/commits/db1481bc6193be3f18cbb9161c038fef91e08526

which makes the tests fail quickly/early if DS9/XPA can't be downloaded (for whatever reason), rather than waiting until all the XSPEC tests have been run (which takes a long time)?

olaurino:
I don't know. When DS9 fails to install you get an error in the smoke test and all the ds9 tests are skipped, so there is still a lot of information you can get from a failed build: it did (or did not) fail because of your changes. Unless of course you are working on the imaging code.

If we wanted the jobs to fail for this kind of reasons, though, I'd just set the bash flags so that the script exits with an error if any statement errors out (which is what we do with the basic setup and testing, e.g. https://github.com/sherpa/sherpa/blob/master/travis/setup_conda.sh#L1). I would reserve this treatment to the scripts that just set up the basic environment or run the tests, i.e. all the other scripts, as it's done now.

olaurino:
Maybe I am missing something, or we are not talking about the same thing. XSPEC-related stuff would not be impacted by db1481b at all. On the other hand, all db1481b would do is to stop jobs too early for code changes completely unrelated to DS9 in case DS9 is not available (say the download server is down), which I keep thinking is undesirable in most cases for exactly the reason you mention: it's unlikely that changes are related to DS9, so we want at the same time the tests to run to validate the changes, and the job to fail because of DS9 missing. That's why that's the only script without `-e` in the first place. We could catch an error with retrieving ds9 and print a meaningful message if that helps, but I'd still prefer the job to continue and run all the tests unrelated to ds9 before failing. I am going to merge this PR, we can add db1481b later if necessary.

PR #493, opened 2018-09-18
fix import

olaurino:
It looks like a previous merge needed an import to be fixed. This PR should fix the tests on master. It's urgent, so no review needed.


DougBurke:
I suggest merging this rather than trying to get the last build to pass, since it's failing on a download (and I've seen this beflaky all day), rather than a failure due to the code.

olaurino:
Travis is having a really bad day.

PR #494, opened 2018-09-20
Tweak the eqwidth docstring

DougBurke:
This removes an erroneous piece of text which messed up the docstring
parsing in the Sphinx documentation, adds new lines to break up the
sections, and rewrites the return section slightly (it's not obvious
from the astropy/NumPy docstring "rules" what the best way to document
routines that return a variable number of arguments).

Also took out un-needed \ lines and updated the copyright year.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016, 2017  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2010, 2015, 2016, 2017, 2018  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


DougBurke:
You can see the current rendering at https://sherpa.readthedocs.io/en/latest/ui/api/sherpa.astro.ui.eqwidth.html#sherpa.astro.ui.eqwidth (well, until it gets fixed with this or another PR ;-)

DougBurke:
Ideally there would be an example showing the return when error=True, but I have no idea what that would look like.

dtnguyen2:
check out file:
sherpa/astro/ui/tests/test_eqwidth_err.py

olaurino:
@dtnguyen2 @DougBurke can this be merged?

DougBurke:
I think it can, as it improves the current documentation. I think it would be useful to add an example showing the errors=True case, but that does not need to be done in this PR.

PR #503, opened 2018-10-04
Fix OutOfMemory regression from #444

### LINKED ISSUE #518: Repeated test code related to ARF/RMF creation
It looks like #482 ended up repeating some tests - e.g. `test_rspmodelnopha_delta_call` is in both `sherpa/astro/tests/test_astro_diag_arfrmf.py` and `sherpa/astro/tests/test_instrument.py`. I haven't checked closely, so there may be differences, but I note that they both fail the same way in https://travis-ci.org/sherpa/sherpa/jobs/440614460#L4148 which suggests that they are the same.

I think we should also update `sherpa/astro/tests/test_instrument.py` so that it uses the "top-level" create-an-arf / rmf functions added in #482 rather than having its own version (which predates #482), to reduce code duplication (and maybe improve test coverage).

dtnguyen2:
Should fix the miss-match ARF bug @anetasie sent in recently

Incidentally, this fixes #518


sherpa/data.py
```diff
@@ -192,6 +192,9 @@ def eval_model_to_fit(self, modelfunc):
     def _get_indep(self, filter=False):
         return self.get_indep(filter=filter)
 
+    def _get_indep(self, filter=False):
```
DougBurke: This is a duplicate of the function on line 192 and should be removed

DougBurke:
Not directly related to this PR, buti s for this functionality: Is there a way to turn off the cache behavior?

DougBurke:
So, what was "special" about this case (which has now been added as a test; yay) that meant non of our existing tests caught the problem? Was it the fact this test doesn't apply any channel filtering at all (and if so I'ma bit surprised we didn't have an existing test), or is it more subtle than that?

DougBurke:
I should note it's also a bit hard to review this PR since there are a lot of commits but not much code change; does it need rebasing?

olaurino:
It looks like this PR needs some TLC. I'll see what I can do.

dtnguyen2:
The following is the results of running a slightly modified fit_3_spectra_simple.py script which can be found in:
  
/data/sherpa/TestIssues/SimulFit3

(sherpa3) [dtn@devel12 integrate]$ grep secs simple.*caching*
simple.allcaching:method = simplex, stat = wstat, in 29.616645097732544 secs
simple.arfcachingonly:method = simplex, stat = wstat, in 40.82709860801697 secs
simple.modelcachingonly:method = simplex, stat = wstat, in 29.58698296546936 secs
simple.nocaching:method = simplex, stat = wstat, in 41.15684127807617 secs

(sherpa3) [dtn@devel12 integrate]$ grep Maximum simple.*caching*
simple.allcaching:	Maximum resident set size (kbytes): 258824
simple.arfcachingonly:	Maximum resident set size (kbytes): 258248
simple.modelcachingonly:	Maximum resident set size (kbytes): 256896
simple.nocaching:	Maximum resident set size (kbytes): 257792

As with the results using the fit_3_spectra.py script, the model caching does seem to speed things but ARF caching is not by much but that is also expected since ARF folding is nothing more then element wise multiplications.  In all cases, memory usage is about the same.

SDS should decide ASAP which version:
1) no caching
2) all caching
3) model caching only
4) ARF caching only
they want to promote to the daily build so it can be tested.  Thanks in advance

DougBurke:
It would be good to know the results when fitting a single dataset - i.e.
when we don't expect the cache to make much of a difference. From my brief
reading of the cache code I don't expect there to be much of a speed up,
but perhaps I'm missing something.

Doug

dtnguyen2:
The bug report titled "ciaot sherpa test now running out of memory" sent in by Kenny has four distinct data sets so we expect model and ARF folding caching not to be useful and this is what I got for time (a bit slower using the caching since code has to check the cache dictionaries but always come up empty handed) and memory usage (about the same, although I don't quite understand why caching is slightly less then when using caching).
BTW, in all cases the fit results are identical for all the combinations of caching and for the three (Aneta's 2 scripts and Kenny's script) tests that I have run

(sherpa3) [dtn@devel12 cachememoryissues]$ grep seconds donotusecache.txt 
	User time (seconds): 5166.45
	System time (seconds): 6.40
(sherpa3) [dtn@devel12 cachememoryissues]$ grep Maximum donotusecache.txt 
	Maximum resident set size (kbytes): 104428
(sherpa3) [dtn@devel12 cachememoryissues]$ grep Maximum allcaching.txt 
	Maximum resident set size (kbytes): 98580
(sherpa3) [dtn@devel12 cachememoryissues]$ grep seconds allcaching.txt 
	User time (seconds): 5439.89
	System time (seconds): 6.50

dtnguyen2:
The current code contains caching for  model evals and ARF folding, so perhaps it is best to promote this to ciao daily so SDS can test.  After all it is easier to remove code then to add them later.

olaurino:
With the standalone release almost out of the way, this is one of my next priorities and I just started investigating what black magic I need to employ in order to get this to a mergeable state. I just need @DougBurke to sign off on the "bare diff" so we can merge it into master.

As for

> given that the cache code here is similar to that in sherpa.models.modelCacher1d, should we be using the same implementation in both, or is that premature at this stage?

Yes, I agree on both accounts. We need to refactor, but it doesn't make sense to refactor if we still don't know if it's useful functionally. If that's still up in the air, let's figure that part out first.

DougBurke:
@olaurino - note that there's a duplicate function from the first commit I called out which needs fixing - i.e. https://github.com/sherpa/sherpa/pull/503/commits/4c715d43e1d3156379dd4523e9533e990af1667b

I think maybe just dropping that commit would be enough.

One thing I don't understand is "where" the cache lives. Is there a single cache (i.e. `myCacheARF` instance), which is my guess, or are there multiple ones (I don't know the semantics of the decorator protocol well enough)? However, this shouldn't stop the merging of this PR, since it does fix a problem with the existing code.

I would like to be able to check the cleaned-up PR before it gets merged, but I'm out Monday so it is not a hard requirement (since I think we want this in sooner rather than later).

olaurino:
@dtnguyen2 I am guessing you did not push the sherpa-test-data commit with the data files you meant to add. That's something I can't fix.

olaurino:
Rebasing, but I couldn't add the additional data files, which do not seem to be available in the `sherpa-test-data` repo. Original tip was 66a3e3e. Test should fail because of the missing data files.

olaurino:
> One thing I don't understand is "where" the cache lives. Is there a single cache (i.e. `myCacheARF` instance), which is my guess,

That is the case. The decorator *replaces* the `my_arf_fold` function definition with its own, which wraps the original function with a callable object. There is a lot I would change or I would have done differently, e.g. `my_arf_fold` has a name argument it doesn't need, it's only needed by the decorator, which is weird. Rather than using a decorator, the `my_arf_fold` function could simply initialize a cache object for each ARF (maybe using the ARF object as a weak key, so when it's garbage-collected the cache is also garbage-collected). So yes, there's a lot of room for refactoring, but I hope I answered your question.

DougBurke:
@olaurino - that commit history looks a lot cleaner. Thanks.

I'm happy for this to be merged (once the `sherpa-test-data` issue is resolved).

olaurino:
Yeah, not so fast. It looks like other than the missing file there are a couple of regressions which should be investigated, I didn't think this PR would change any values, but it looks like it does, so there might be a bug somewhere.

DougBurke:
I added #518 which I noted when looking at one of the test failures, but that is a code clean up that can be done later rather than now.

olaurino:
It looks like #503 never triggered the Travis jobs, probably because of the conflicts that prevented a merge right off the bat (and I am assuming the fork does not have Travis enables, so push jobs are not triggered either). So I can't confirm that the tests always failed, which would confirm it's a regression, but from the looks of it, it's a regression.

Plus, what @DougBurke said in #518

dtnguyen2:
> > One thing I don't understand is "where" the cache lives. Is there a single cache (i.e. `myCacheARF` instance), which is my guess,
> 
> That is the case. The decorator _replaces_ the `my_arf_fold` function definition with its own, which wraps the original function with a callable object. There is a lot I would change or I would have done differently, e.g. `my_arf_fold` has a name argument it doesn't need, it's only needed by the decorator, which is weird. Rather than using a decorator, the `my_arf_fold` function could simply initialize a cache object for each ARF (maybe using the ARF object as a weak key, so when it's garbage-collected the cache is also garbage-collected). So yes, there's a lot of room for refactoring, but I hope I answered your question.

innitialize a cache object for each ARF will guarantee that caching will never work, cause the jth ARF does not have access to the ith ARF cached values (where i < j).

olaurino:
I believe @dtnguyen2 and I pushed the correct `sherpa-test-data` commit and then updated this PR accordingly. Tests should still fail, but only if there is an actual regression in some existing tests.

DougBurke:
So, most of the errors in the "simple" build - https://travis-ci.org/sherpa/sherpa/jobs/440705143 - look to just be a by-product of the recent clean up in 8ec64fc (i.e. just change `test-arf` to `user-arf` or vice versa) - but we still have the failure in `test_rspmodelnopha_delta_call` at https://travis-ci.org/sherpa/sherpa/jobs/440705143#L4142 which is concerning

olaurino:
Can anybody reproduce the regression on their systems? If I run the offending test on my workstation (RHEL 7, Conda Python 3.5.6) it doesn't fail.

olaurino:
Never mind, I guess I was selecting the wrong test -.- It's Friday.

olaurino:
Ah! it wasn't me. It's just that when run by itself the test is ok, it fails when run with others. Which might still be an issue with the caching mechanism.

DougBurke:
Well, I would claim the fact the test runs okay on its own but not after other runs indicates it is a problem with the cache, but unfortunately I've seen odd behavior like that before with our test suite (I do think it must be with the cache code, but it's just not as clear cut as we'd like).

I wonder if we have some other test that uses the same ARF name as the one used here, but with different data, hence the cache problem?

Maybe we should just merge the binning fix and comment out/remove the ARF cache for now (since it looks - from initial data provided by @dtnguyen2 - that the ARF cache doesn't seem to make a significant difference)?

olaurino:
> but unfortunately I've seen odd behavior like that before with our test suite

True, however that usually happens with old-style tests that did not clean before and after themselves. `test_instrument` is all new-style pytest tests, they pass if you disable the ARF caching, they don't when you enable it, and this does look like the caching not doing its job, so when you run several tests you get to a point the cache returns a value it shouldn't return.

I am fine with merging the code with the ARF caching disabled, just let me know if that's the case.

olaurino:
Just to clarify, if I only run `test_instrument` the offending test fails. Just that:

```
python setup.py test -a "-v -k test_instrument"
```

And if I manually disable the ARF caching, the issue goes away.

Earlier @dtnguyen2 seemed to have a lead as to what was going awry, so I'll wait for an update from both @DougBurke and @dtnguyen2 before disabling ARF caching and merging this.

DougBurke:
Note: I am not going to be able to comment on this until Tuesday at the earliest now.

dtnguyen2:
By making the change from test-arf --> user-arf in the test_instrument.py 
and one can't use ARF's filename for caching ARF rsp then the test_instrument.py passes just fine.  However, I'm having problems with two XSPEC tests but this is the first time I'm using 12.10

XSPEC::getAbundance: Invalid element:  entered, returning 0.
XSPEC::getAbundance: Invalid element:  entered, returning 0.
XSPEC::getAbundance: Invalid element:  entered, returning 0.
XSPEC::getAbundance: Invalid element:  entered, returning 0.
XSPEC::getAbundance: Invalid element:  entered, returning 0.
XSPEC::getAbundance: Invalid element:  entered, returning 0.

dtnguyen2:
It looks like @olaurino beat me to it by changing the 'test-arf' to 'user-arf'.  The remaining error in the test_instrument.py is because the ARF caching was using the ARF (file)name as part of the cache which does not work for create_arf so this forces me to use the actual rsp in the ARF file.

There are still two errors, in the  sherpa/astro/xspec/tests/test_xspec.py test which I don't think has anything to do with the caching (I commented out all the calls to caching and the same two errors still show up).

dtnguyen2:
Not on the computer right now but I think I have Xspec 12.10.? And when I run the Sherpa test I get two Xspec errors.  The errors have nothing to do with the caching code so I don't think Xspec was properly built

PR #508, opened 2018-10-06
Ensure fields are initialized before use (pileup code)

DougBurke:
# Summary

This avoids a warning message from the compiler, and fixes a potential attempt to free unallocated memory if there was a memory-allocation failure.

# Details

I do not think this is a serious issue, as an attempt to call `free` on an unitialized variable would only happen in an error case (when `XMALLOC` fails), so things are already going wrong in this case. It does just remove one extra warning from the compilation output.

It is also not a 4.10.1 lien.


DougBurke:
I note that really the `delete_kernel` call should be made in other places earlier in `init_kernel` if other `XMALLOC` calls fail - and if this were to be done then all the pointers in `pileup_kernel_t` should be set to `NULL` at the start, but that this leak would only occur if there's a memory-allocation failure.

PR #513, opened 2018-10-09
Add a simple DataPlot test with PHA data

DougBurke:
This tests out the handling of the error calculation for data
objects. At present the only test is that the plot can be
prepared for each statistic value. The primary motivation for
this test is because the error calculation is known to be
wrong for non-Chi2XXX statistic classes due to a recent
change from the Sphinx documentation. This commit adds a simple
test which can validate any change.

### DIFF
--- a/sherpa/astro/tests/test_astro_plot.py
+++ b/sherpa/astro/tests/test_astro_plot.py
@@ -17,11 +17,15 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 import numpy
-from sherpa.utils.testing import SherpaTestCase
+from sherpa.utils.testing import SherpaTestCase, requires_data, requires_fits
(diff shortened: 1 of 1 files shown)


DougBurke:
@olaurino - just in case this is useful for your work, although I would guess you have written your own version of this.

DougBurke:
There's some oddity with the number of tests run when using `pytest.params` - it seems to think there are two per call, which I am guessing is due to the test itself failing and there being some warnings - but if you remove them on the fixed branch I think the number should be fine. I don't think it's worth worrying about, but just thought I'd mention it here.

olaurino:
@DougBurke I can approve this but I wouldn't mind you double checking the fix. Thanks a lot!

anetasie:
I just checked the plots in the 4.10.1 conda build. I confirm that the bug is fixed.

PR #517, opened 2018-10-11
Remove use of deprecated imp (fix #516)

### LINKED ISSUE #516: module imp is deprecated (seen when building with python 3.7)
With python 3.7 we see the following message when `setup.py` is run:

```
% python setup.py develop
setup.py:26: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
...
```

It looks like we can either:

1. change from 'imp' to 'importlib' once we drop Python 2.7 support (since `importlib` in Python 2.7 only contains one routine)
 2. change how we check for NumPy - e.g. perhaps just an 'import numpy' rather than mess around with fancy tricks

I haven't a clue which is better, or if there's another option.

DougBurke:
# Description

Avoid a warning message when building Sherpa with Python 3.7.

Update the classifiers section of the Python build configuration to reflect recent changes.

# Details

This looks a sensible change, but I then read answers like https://stackoverflow.com/a/41575848 and wonder why they don't just follow the approach I took here, which makes me worry slightly.


olaurino:
> This looks a sensible change, but I then read answers like https://stackoverflow.com/a/41575848 and wonder why they don't just follow the approach I took here, which makes me worry slightly.

That approach is fancier, and we might get there. On the other hand, the check for `setuptools` was meant to work for CIAO. We can probably either assume `setuptools` is installed or just include the standard bootstrap scripts ourselves.

Numpy was required for the Fortran extensions, we can now probably remove all the numpy imports for `setup.py` and the `helpers`, as with Fortran gone we can use the basic `setuptools`/`distutils` commands.

This PR does get rid of the warning, so I'll just merge it.

DougBurke:
@olaurino - do we need a ticket to remind ourselves to look at doing the clean up you mention in https://github.com/sherpa/sherpa/pull/517#issuecomment-429967676 ?

PR #520, opened 2018-10-15
reintroduce test that was dropped in conflict resolution

olaurino:
@DougBurke reported to me (he tried to post on GitHub but it didn't seem to work from the airport) that I had probably inadvertently dropped some assertions from a test while I was resolving a conflict in #503. He is probably right. I didn't want to stir the pot too much for the PR, so I merged it as-is, and then issued this PR.


sherpa/astro/tests/test_astro.py
```diff
@@ -539,6 +539,16 @@ def test_proj(self):
...
         self.run_thread('proj_bubble')
 
+        fit_results = ui.get_fit_results()
+        covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
+        assert covarerr[0] == approx(0, rel=1e-4)
+        assert covarerr[1] == approx(8.74608e-07, rel=1e-3)
+        
+        # Fit -- Results from reminimize
+        assert self.locals['mek1'].kt.val == approx(17.8849, rel=1e-2)
+        self.assertEqualWithinTol(self.locals['mek1'].norm.val, 4.15418e-06,
```
DougBurke: Should we have both checks using the same form (preferably the pytest `approx` rather than the `assertEqualWIthinTol` method)? I imagine this would be a clean-up of the original code (ie is not due to the merge issue that caused the tests to be dropped)?

DougBurke: Actually, it turns out we have these checks are repeated below  see line 578. However, in that case the tests use an absolute tolerance (was 0.01, now 0.057) rather than a relative one, and an absolute tolerance of ~1e-2 for the normalization, which is ~4e-6, is obviously pointless.

So, I think we can drop this PR but may want to review this test

DougBurke: Or perhaps replace the existing tests (e.g. lines 578 and 579) with the "new" ones (lines 548 and 549).

olaurino: I believe I addressed these comments in the new commits (I also rebased to get the latest changes from master).

olaurino: @DougBurke could you please confirm if this is resolved?

DougBurke:
@olaurino - if we look at the whole test now, we have

a) a mixture of `pytest.approx` and `numpy.assert_allclose`

b) because of a, we have a mixture of relative and absolute tolerances

c) a mixture of referring to the expected values in the tests either by value (e.g. `12.3423`) or by name (e.g. `kt_val`) [values/names made up for this example]. I probably added the use of names (to try and make it easier to see what is going on), but maybe it isn't worth it here, especially as the names are only used once (or, in the case of `kt` and `norm`, aren't actually used anymore I think)

I am most concerned about point b, since, as noted in the comments

```
    norm_min_covar = -8.847916e-7
    atol = 0.057
    assert_allclose(covar.parmins[1], norm_min_covar, atol=atol, rtol=0)
```

doesn't strike me as much use as a test.

So, ideally, I think we should move the other tests to pytest.approx and see if rtol=0.01 works for them (then tweak them if they don't). However, this then becomes a test re-write [*], more than a "add back in a test", so I don't know if you feel like you want to do that at this time.

[*] since really these points probably apply to other tests in this file too

olaurino:
I have to say this seems like more work than the PR was supposed to do. In any case I'll try pushing a new commit replacing `assert_all_close` with `approx` (I hadn't even paid enough attention to realize `assert_all_close` was being used to compare scalars). Let's see where I can get. I'll also have to check why the tolerances had changed to the point they were not longer meaningful. I don't remember where all this is coming from.

olaurino:
Unsurprisingly, the test is failing. I could mark this as expected to fail. I could use the new values as new baselines. That seems tautological, but this is a *run-thread* regression test, so as long as somebody tells me the new values are correct I am fine with updating them. Or I could mark the test as expected to fail, but that's not great.

DougBurke:
I've added 1% relative tolerances for the error bounds. This should be okay as these are regression tests (and matches the tolerance used for the best-fit results). I haven't been able to run this locally (my old build of XSPEC 12.9.0 ends up crashing the sherpa test suite for some reason and I can't build with 12.10.1 until #533 is addressed), so I'm hoping that it all just works on travis.

olaurino:
I believe all the comments were addressed, @DougBurke unless you have other comments I'd assume this is good to go?

DougBurke:
LGTM - I don't know if you want to keep or remove 901e17e (as I haven't tested it in the case that a download fails).

olaurino:
No that's fine. I created a new branch to test that but it looks like I cherry-picked both commits, so obviously it won't trigger the issue. I wanted to test that off to the side and then merge this.

olaurino:
The test was successful. The warning message has a couple of spurious "\n" but we can fix that in a different PR if necessary (see https://travis-ci.org/olaurino/sherpa/jobs/476507462#L307).

olaurino:
When testing the update with xspec 12.10.0e I get the error at the end of this job:
https://travis-ci.org/olaurino/sherpa/builds/481553857

Any suggestions on how to deal with it?

dtnguyen2:
It took a while, but I managed to find it.  Here is my explanation of why sherpa's implementation of covar is troublesome:

https://icxc.harvard.edu/pipe/catdev/2014/0085.html

The following excerpts 

I don't know exactly what algorithm MINUIT HESSE uses. From what I've seen, yes, it gives more accurate results than Sherpa covariance.

were taken from the following thread:

https://github.com/sherpa/sherpa/issues/548

MINUIT has what I consider to be a (slightly) better implementation.  I do know MINUIT's attempt at the covariance calculation which I can re-send it again if you guys care to see but it is in the algps.pdf file that I have sent before.

As to what to do with this test?  Get rid sherpa's covar, why give users something that we know do not give reliable results?

DougBurke:
@olaurino in https://github.com/sherpa/sherpa/pull/520#issuecomment-455700283 are you talking about the `Test known to crash XSPEC 12.10.0` messages?

If so, then it comes from https://github.com/sherpa/sherpa/blob/1a9600cbbc2935ea2eb4fce05f4992a6dc4a4406/sherpa/astro/xspec/tests/test_xspec_unit.py#L700 which has the text 

```
XSPEC 12.10.0 (if not manually patched) will crash if an
XSPEC table model is evaluated beyond its grid (this is only
an issue for programs like Sherpa that use XSPEC as a "library").
```

and it is because of the changes made in XSPEC 12.10.0 which cause problems for access to the "models only" library (the output stream needed by the code is no longer set up when the models-only version of the library is initialized; HEASARC sent us a patch for 12.10.0 for use in CIAO but this has since been fixed in XSPEC 12.10.1).

Does this help?

olaurino:
@DougBurke I meant this failure:

````
________________________ test_threads.test_proj_bubble _________________________
self = <test_astro.test_threads testMethod=test_proj_bubble>
    @requires_fits
    @requires_xspec
    def test_proj_bubble(self):
        self.run_thread('proj_bubble')
    
        fit_results = ui.get_fit_results()
        covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
        assert covarerr[0] == approx(0, rel=1e-4)
        assert covarerr[1] == approx(8.74608e-07, rel=1e-3)
    
        # Fit -- Results from reminimize
        assert self.locals['mek1'].kt.val == approx(17.8849, rel=1e-2)
        assert self.locals['mek1'].norm.val == approx(4.15418e-06, rel=1e-2)
    
        # Fit -- Results from reminimize
    
        # The fit results change in XSPEC 12.10.0 since the mekal model
        # was changed (FORTRAN to C++). A 1% difference is used for the
        # parameter ranges from covar and proj (matches the tolerance for
        # the fit results).
    
        # Covar
        #
        # TODO: should this check that parmaxes is -1 * parmins instead?
        covar = ui.get_covar_results()
>       assert covar.parmins[0] == approx(-0.328832, rel=0.01)
E       AssertionError: assert -0.3363520598126048 == -0.328832 ± 3.3e-03
E        +  where -0.328832 ± 3.3e-03 = approx(-0.328832, rel=0.01)
sherpa/astro/tests/test_astro.py:512: AssertionError
````

DougBurke:
@olaurino I think this is discussed in the XSPEC 12.10.1 PR at https://github.com/sherpa/sherpa/pull/534#issuecomment-453138816

olaurino:
I'll have to look at the changes and see if I can apply them to 12.10.0 as well (of course everything is on hold because the ds9 download page is not reachable and the jobs are hanging).

In the meantime (and slightly OT), what versions of xspec should we test in travis? I was assuming 12.9.1, 12.10.0, and 12.10.1, but it's up to you really.

DougBurke:
I am most concerned with 12.10.0 and 12.10.1. Given that we still have 12.9.0/1 code then ideally we'd have a 12.9.1 test (especially given the change in XSPEC internals between 12.9.1 and 12.10.0). The downside to including as 12.9.1 test is that the XSPEC tests really slow the system down, so perhaps we should aim for 12.9.1 and 12.10.1, with 12.10.0 being optional (we currently have CXC-internal test runs with 12.10.0 so we know that we do have some testing of this version).

PR #497, opened 2018-09-22
WIP a rebased version of PR 465 (XSPEC 12.10.0)

DougBurke:
I've rebased PR #465 but am adding a new PR in case I mess up anything @olaurino had done with that branch. The aim here is to update the sphinx documentation and see how that goes.

PR #521, opened 2018-10-16
Docs: update build installation instructions for XSPEC 12.10.0

DougBurke:
# Summary

Update the README and Sphinx documentation to note building against XSPEC 12.10.0.

This doesn't need a release note since it is just part of the newly-added XSPEC 12.10.0 support. It is the only part of the #497 rebase of #465 that hasn't been added.

# Details

Changes include:

 - the addition of the `xspec_version` setting
 - when using XSPEC 12.10.0 some of the settings are different.

I'm not sure what problems there still are when building against the XSPEC 12.10.0 models-only option, so have just left a comment about "there can be problems".

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,27 +447,29 @@ newer versions, but without support for new models or features.
 
 To build the XSPEC support in Sherpa, the `xspec_config` section of the
 `setup.cfg` file will need changing to point to the libraries, and to
-turn on the extension. In all cases, set
+turn on the extension, **and** the `xspec_version`` option must be
+set to the XSPEC version. So, in all cases, set
 
-    with-xspec=True
+    with-xspec = True
+    xspec_version = 12.10.0
+
+(changing the XSPEC version to the appropriate value, noting that
+the patch level must not be included).
 
 The remaining settings depend on how the XSPEC libraries have
-been built (in the examples below, environment variables are
-used, but the full path should be in your own copy of the file):
+been built. In the examples below, the `$HEADAS` environment variable 
+**must be replaced** by the actual path to the HEADAS installation,
+and the versions of the libraries - such as `CCfits` - may need to
+be changed:
 
  1. If the full XSPEC 12.10.0 system has been built, then use
 
         xspec_include_dirs = $HEADAS/include
         xspec_lib_dirs = $HEADAS/lib
-        xspec_libraries = XSFunctions XSModel XSUtil XS hdsp_2.9
+        xspec_libraries = XSFunctions XSModel XSUtil XS hdsp_3.0
         cfitsio_libraries = cfitsio
         ccfits_libraries = CCfits_2.5
         wcs_libraries = wcs-5.16
-
-    The environment variable `$HEADAS` should be expanded out, and the
-    version numbers of the `wcs`, `cfitsio`, and `CCfits` libraries
-    may need to be changed, depending on the version of XSPEC (the
-    values given above are valid for XSPEC 12.10.0).
 
  2. If the full XSPEC 12.9.x system has been built, then use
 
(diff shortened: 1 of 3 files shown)


olaurino:
We released 4.10.1 without xspec 12.10.0 support, while we are trying really hard to get such support in CIAO 4.11. I expect the work we are doing in CIAO to have an impact on the user experience for those users willing to build Sherpa with XSPEC support. I would use this PR to review the installation process once we have ironed out the CIAO side.

Also, there is a lot of duplication at this point between the README and the sphinx docs. I'd suggest we simplify the README a lot and just point to the official documentation, possibly with an extended qualifier, so previous versions refer to the appropriate RTD tag. @DougBurke what do you think?

olaurino:
Let me add that the reason for the difference between CIAO and Standalone in building XSPEC has to do with the fact that in CIAO we are now building the XSPEC libraries using the standard, proper linking strategy, so to reduce the chance of obscure segmentation faults at runtime. The XSPEC libraries will now have information about their dependencies in their headers, rather than leaving it to Sherpa to link the necessary dependencies, which is non-standard, error-prone, and has given us innumerable headaches over the years.

Standalone users will be stuck with the custom, error-prone libraries that do not provide their own linking information, and to accommodate both CIAO and Standalone builds users might be required to provide a little more information in the configuration file rather than relying on defaults (and yet experience the occasional obscure run-time segmentation fault).

On the other hand, as of CIAO 4.11 it should become much easier and less error prone to build Sherpa from sources with XSPEC support using the CIAO distribution, which is currently not recommended by our instructions (not sure why, I never had problems with that, but I guess we had reasons).

DougBurke:
I have been meaning to suggest that we just make the README point to the documentation now - except perhaps for a "if you have conda, do this ..., otherwise see the README" - but never quite got around to it.

DougBurke:
Should we look at expanding out variables for the `xspec_lib_dirs` and `xspec_include_dirs` options? That is, if a user has in their config

    xspec_lib_dirs = $HEADAS/lib
    xspec_include_dirs = $HEADAS/include

then we expand out these variables for them?

DougBurke:
For building against the CIAO-supplied XSPEC. I've seen problems when trying to build code using a gfortran compiler significantly newer than the version used to build XSPEC in CIAO.

DougBurke:
(sorry for breaking things up, but GH pr's don't thread very well...)

@olaurino, for the build issues you discussed in https://github.com/sherpa/sherpa/pull/521#issuecomment-430402586 - am I to take it that you've managed to tweak the XSPEC build, as suggested by the HEASARC folks, to add in the correct transitive dependencies for the models-only build? If so, we have the option of telling people how to do that, or suggesting they just build the full XSPEC distribution (since it isn't significantly different in download size/installation complexity) and - at least when I tried it prior to the recent changes in Sherpa - seemed to work. 

Well, we could document both options, depending on how complex the changes needed to the XSPEC builds are.

olaurino:
Indeed, that's an option. However we had to manually edit a complex configure script as part of the process, and our edits only make sense for CIAO, where we want to build some of the dependencies (cfitsio) ourselves with a specific versions, other XSPEC dependencies (CCfits) need to pick up that specific version, while other can just be built as part of XSPEC. Plus a number of environment variables to correctly set. Instructions for a general user would probably be simpler, I just can't tell right now.

olaurino:
> for the build issues you discussed in [#521 (comment)](https://github.com/sherpa/sherpa/pull/521#issuecomment-430402586) - am I to take it that you've managed to tweak the XSPEC build, as suggested by the HEASARC folks, to add in the correct transitive dependencies for the models-only build?

Yes.

DougBurke:
UX question: I just tried building master with XSPEC 12.9.1, so set `xspec_version = 12.9.1` and you see at the start of the build (this was adding in XSPEC to an already-built installation, so it didn't have to build anything else):

```
% python setup.py develop
running sherpa_config
warning: sherpa_config: built configure string['./configure', '--prefix=/lagado.real/sherpa/sherpa-no-erf/build', '--with-pic', '--enable-standalone', '--disable-maintainer-mode', '--enable-stuberrorlib', '--disable-shared', '--enable-shared=libgrp,stklib', '--enable-fftw', '--enable-region', '--enable-group', '--enable-stk', '--enable-wcs']

running xspec_config
Found XSPEC version: 12.9.1
running develop
running build_scripts
...
```

Is it even worth reporting this version anymore (I was trying to reword this line, to something like `Building support for XSPEC version: ...`, but then wondered if it was useful or not).

olaurino:
Sorry, I had missed your questions.

> Is it even worth reporting this version anymore

Yes, for debugging, but we should probably change the message.

> Should I look at "slimming down" `README.md` in this PR, or another one?

another one.

PR #523, opened 2018-10-18
Ensure the ui modules have a unique list of exported symbols (fix #502)

### LINKED ISSUE #502: sherpa.astro.ui erf confusion
With the current master branch

```
% python -c 'from sherpa.astro import ui; print(len(ui.__all__)); print(len(set(ui.__all__)))'
695
694
```

This is because `erf` appears in it twice - once because it is imported from `sherpa.utils` and then it gets added because of the `sherpa.models.basic.Erf` model.

I think we just remove the explicit import of `erf` from `sherpa.utils` here. I've checked and there is no reference to the `erf` symbol in `sherpa.astro.ui.utils` (which is what is used to populate `sherpa.astro.ui`), so it should be safe.

I think this is a minor issue (almost cosmetic).

DougBurke:
# Summary

Remove the `sherpa.utils.erf` symbol from the two `ui` modules as it is "over-written" by the ModelWrapper-created symbol when applied to `sherpa.models.basic.Erf`.

Tests have been added to point out if this ever happens again: for example, if any new models are added to Sherpa, they could conflict with an existing symbol, and these tests will pick that up.

# Details

I originally thought this was a minor issue, but it strikes me that it is possible to add a Sherpa model and accidentally shadow an existing symbol that a user could be relying on. Of course, we are exporting too many symbols in the `sherpa.ui` and `sherpa.astro.ui` modules, but that's a discussion for another day. 

#502 only mentions `sherpa.astro.ui` but when fixing it I noticed it also happens for `sherpa.ui` too (which is not too surprising given their ancestry/relationship).

### DIFF
--- a/sherpa/astro/ui/__init__.py
+++ b/sherpa/astro/ui/__init__.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2007  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2007, 2018  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


olaurino:
> Of course GitHub's decision to order commits by their version of reality, not mine, means the ordering looks odd here (I'd used a rebase to move things around).

That's weird. It looks fine in git log.

PR #524, opened 2018-10-18
Be explicit about "invalid escape characters" in strings

DougBurke:
# Summary

Convert strings that contain "invalid" escape characters, which newer versions of Python complain about, to raw strings by preceding them with a 'r' character.

This avoids the test suite from reporting DeprecationWarnings.

# Details

I was surprised the tests didn't fail because of the DeprecationWarnings, but I know there have been changes to how warnings are handled by pytest. I decided not to add any tests to show the original problem as it didn't seem worth it. I manually ran Sphinx on this PR and things looked good.


DougBurke:
This is related to this Python 3.6 change: https://bugs.python.org/issue27364

PR #525, opened 2018-10-19
fixed based line radpro_dm test and changed double --> real (template)

dtnguyen2:
1) radpro_dm should have the same covariance as radpro.
2) changed a few (I think 3) double to real since they are within the C++ template code

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -293,9 +293,9 @@ class test_threads(SherpaTestCase):
 
             fres = ui.get_fit_results()
             covarerr = sqrt(fres.extra_output['covar'].diagonal())
-            self.assertEqualWithinTol(covarerr[0], 9.37344776, 1e-4)
-            self.assertEqualWithinTol(covarerr[1], 0.51259645, 1e-4)
-            self.assertEqualWithinTol(covarerr[2], 0.06911017, 1e-4)
+            assert covarerr[0] = approx(9.37344776, rel=1e-4)
+            assert covarerr[1] = approx(0.51259645, rel=1e-4)
+            assert covarerr[2] = approx(0.06911017, rel=1e-4)
             assert fres.statval == approx(217.450, rel=1e-4)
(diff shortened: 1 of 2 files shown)


sherpa/astro/tests/test_astro.py
```diff
@@ -293,9 +293,9 @@ def test_radpro_dm(self):
 
             fres = ui.get_fit_results()
             covarerr = sqrt(fres.extra_output['covar'].diagonal())
-            assert covarerr[0] == approx(6.98861, rel=1e-4)
-            assert covarerr[1] == approx(0.383065, rel=1e-4)
-            assert covarerr[2] == approx(0.101712, rel=1e-4)
+            self.assertEqualWithinTol(covarerr[0], 9.37344776, 1e-4)
```
DougBurke: A question for @olaurino - should we try and stick with the use of `pytest.approx` here, rather than the `assertEqualWithinTol` method? I would like to be able to remove the `SherpaTestCase` class (or whatever it is called) in the long term, *but* I haven't looked at the meaning of the tolerances in the two paths (e.g. does `pytest.approx` only provide relative, absolute, or both tolerances, and how does it map to the third parameter of `assertEqualWithinTol`).

olaurino: Hmm... I thought I had already replied to this. We should be consistent in the use of `pytest.approx`.

olaurino:
Note, this PR addresses an issue with the CIAO tests, as the `radpro_dm` test only runs with crates, and it's manually skipped otherwise. That's also wrong. I'd like to merge this asap so we can see the tests go back 100% in CIAO.

DougBurke:
Ideally (for the short term at least) we'd have the same test code for `radpro` and `radpro_dm` (to ensure the same results are being returned).

olaurino:
Argh, this slipped through the cracks. As it just fixes a regression test we are not going to take it for CIAO 4.11, but I'll merge this so it can at least be tested in CIAOD.

Issue #507, opened 2018-10-06
pileup test fails with XSPEC: covariance matrix is different


DougBurke:
# Summary

The replacement of the Fortran optimisation code with C++ has lead to the pileup test to fail with XSPEC 12.9.1p. The errors are in the diagonal elements of the covariance matrix; the best-fit parameters and statistic are the same (within the error tolerance). What is is about my system that is different enough to the Travis tests?

# Details

I have master checked out - i.e. 3540951 - built against XSPEC 12.9.1p

```
% git rev-parse HEAD
354095107245d5190c6cfe54741c5f4c2d4f87f9
% python -c 'from sherpa.astro import xspec; print(xspec.get_xsversion());'
12.9.1p
``` 

The pileup test fails with:

```
% python setup.py test -a sherpa/astro/tests/test_astro.py::test_threads::test_pileup
running sherpa_config
warning: sherpa_config: built configure string['./configure', '--prefix=/home/djburke/sherpa/sherpa-fix-sphinx-docs/build', '--with-pic', '--enable-standalone', '--disable-maintainer-mode', '--enable-stuberrorlib', '--disable-shared', '--enable-shared=libgrp,stklib', '--enable-fftw', '--enable-region', '--enable-group', '--enable-stk', '--enable-wcs']

running xspec_config
Found XSPEC version: 12.9.1p
running develop
running build_scripts
running egg_info
running build_src
build_src
building extension "sherpa.estmethods._est_funcs" sources
building extension "sherpa.utils._utils" sources
building extension "sherpa.models._modelfcts" sources
building extension "sherpa.optmethods._saoopt" sources
building extension "sherpa.optmethods._tstoptfct" sources
building extension "sherpa.stats._statfcts" sources
building extension "sherpa.utils.integration" sources
building extension "sherpa.astro.models._modelfcts" sources
building extension "sherpa.astro.utils._pileup" sources
building extension "sherpa.astro.utils._utils" sources
building extension "sherpa.utils._psf" sources
building extension "sherpa.astro.utils._wcs" sources
building extension "sherpa.astro.utils._region" sources
building extension "sherpa.astro.xspec._xspec" sources
building data_files sources
build_src: building npy-pkg config files
writing sherpa.egg-info/PKG-INFO
writing dependency_links to sherpa.egg-info/dependency_links.txt
writing entry points to sherpa.egg-info/entry_points.txt
writing requirements to sherpa.egg-info/requires.txt
[...]

PR #527, opened 2018-10-19
Switch pileup test from unittest to pytest style

DougBurke:
# Summary

An internal change to the tests (change the pileup thread from unittest to pytest style) added whilst investigating #507.

# Details

There is no significant functional change here, except that

a) the covariance check is moved later, so we can see whether the
   best-fit parameter values have changed (given the test
   tolerance), since at some level these are "more important" than the covariance
   checks (added to help track down what is going on with #507)

b) the logging has been left at its default level, which provides
   more screen output on failure.

There may be some subtle differences due to how the Sherpa "state"
is saved by the test_threads class compared to the clean_astro_ui
fixture. The difference is in whether the ui._session dict is
saved/restored, or whether sherpa.astro.ui.clean() is called.

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -211,62 +211,6 @@ class test_threads(SherpaTestCase):
         self.assertEqual(ui.get_fit_results().numpoints, 4881)
         self.assertEqual(ui.get_fit_results().dof, 4877)
 
-    # This has been moved to its own pytest-style test below (test_thread_pileup)
-    # but this has been left in for reference.
-    #
-    @requires_fits
-    @requires_xspec
-    def _test_pileup(self):
-        self.run_thread('pileup')
-
-        fr = ui.get_fit_results()
-        covarerr = sqrt(fr.extra_output['covar'].diagonal())
-        assert covarerr[0] == approx(684.056 , rel=1e-4)
-        assert covarerr[1] == approx(191.055, rel=1e-3)
-        assert covarerr[2] == approx(0.632061, rel=1e-3)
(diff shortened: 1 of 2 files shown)


DougBurke:
About to rebase. Original tip was 5e27178

olaurino:
@DougBurke I refactored some of the infrastructure so it's easily reusable. I ran the tests locally and everything seemed fine, but Travis will tell us if I missed some edge cases. And of course please let me know if you have comments on the changes, or if you think they should be different.

DougBurke:
I'm okay with the refactoring. I am a bit wary of the "magic" behind pytest making it hard to know where some of these fixtures come from, especially for ones where we only use the fixture in one file, but I don't feel strongly enough to say your code should be changed ;-)

olaurino:
I can see your point. If that's the only place where we ever expect to use the `run_thread` fixture we could move it back, but I think it's useful to make it a fixture anyway. And I also think it's good to have the `clean_ui` fixture, which at some point we should run automatically before and after every test. If you don't feel strongly about it I'll merge this as is. Thanks!

DougBurke:
Merge away.

Just a few notes about the clean fixture (I do agree we want to use it more
liberally): it only needs to be run on tests that use the ui interface; do
we need one for sherpa.ui and one for Sherpa.astro.ui?; I wonder if there’s
cases when we wouldn’t want to use it?

olaurino:
Note we could disable autousing fixtures if it made sense to do so:
https://stackoverflow.com/a/38763328

I would be more comfortable with making sure that the `ui` session is cleared before and after every test, and I think we are moving in that direction. On the other hand doing so might hide issues in the management of the session or in the tests themselves, so I am really ambivalent about it.

PR #530, opened 2018-10-22
ensure XSPEC 12.10.0 uses ATOMDB 3.0.9 by default

DougBurke:
# Summary

Ensure that the default AtomDB version is 3.0.9 when using XSPEC 12.10.0. This is to fix an issue with the models-only XSPEC installation in XSPEC 12.10.0 which uses a default version of 3.0.7 but only provides data files for 3.0.9. Without this change models such as XSapec would return all 0's (unless the AtomDB version was set manually by the user).

# Details

A test is added to ensure that the model evaluation returns at least bin > 0. This means that the default grid had to be increased up to 7 keV for the XSgaussian model. Several models are excluded from this test (XSismabs, XSkerrd*, XSmkcflow, XSvmcflow) since they evaluate to 0. They should be looked at, but it is a different issue (and does not seem to be a regression with XSPEC 12.10.0 support).

I have also updated the XSPEC evaluation tests to change them from unittest to pytest style so that each model is run as an independent test. This improves coverage in the case of a failure (the previous version would stop at the failed model) *and* ensures that the screen output when there is a failure is just related to that model (and not all of them, which was hard to track through on Travis when developing this PR). I have managed to resist converting the whole test to pytest style, which means that there is some repeated logic in the test suite. One immediate advantage of this new test is that I was able to work out why the two `cflow` models return 0 (since I could now see the relevant output from the model evaluation): the default redshift parameter for these two models is set to 0 (I checked that this holds in the XSPEC 12.10.0 model.dat file and it does, so it is not a transcription error on our part) but the model code requires it to be > 0.

You may note that I've forgotten all the rules about string creation in C/C++ code in 1ae44cc so have gone the "lets be overly safe" route to ensure the string is null terminated ;-)

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec.py
+++ b/sherpa/astro/xspec/tests/test_xspec.py
@@ -68,8 +68,14 @@ def remove_item(xs, x):
 # results for any model). For now stick with testing most of
 # the models.
 #
-def get_xspec_models(xs):
-    """What are the XSpec model names to test."""
+def get_xspec_models():
+    """What are the XSpec model names to test.
+
+    This must only be called after making sure that XSPEC support
+    is available.
+    """
+
+    import sherpa.astro.xspec as xs
 
     # The alternate approach is to use that taken by
     # test_create_model_instances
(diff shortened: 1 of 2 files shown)


DougBurke:
Hmmm, the default parameters for `XSgaussian` (6.5 keV peak, sigma = 0.1 keV) means that we would need to increase the grid maximum energy from 5 to ~6.5 (say 7 keV).

however, I am now seeing issues with XSismabs, which I can't get to return anything but 0's!

DougBurke:
Doh. This "model > 0 in at least one bin" is being more annoying than I expected (and not helped by some stupid testing on my part).

DougBurke:
About to rebase to clean up the tests. Original tip was 80e001b

DougBurke:
I think I've finished with this PR now ;-)

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -611,3 +575,138 @@ def __init__(self, name='zbabs'):
     expected = XSzbabs()([1, 2, 3])
 
     assert_array_equal(expected, actual)
+
+
+# The following repeats logic in self.assertFinite above.
+# The plan is to move the SherpaTestCase derived tests into
+# plain pytest-style tests, and then self.assertFinite
+# will be removed, but that day is not today.
```
olaurino: If I am not mistaken `assertFinite` only uses `self` to use a specific implementation of the assertions, which I am assuming are equivalent to the `pytest`-powered assertions in `assert_is_finite`. Have you considered changing the implementation of `assertFinite` as just a call to `assert_is_finite`? Possibly making `assertFinite` a  static method?

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -611,3 +575,138 @@ def __init__(self, name='zbabs'):
...
+    #
+    # Some models seem to return 0's, so skip them for now.
+    #
+    # The *cflow models return 0's because:
+    #     XSVMCF: Require z > 0 for cooling flow models
+    # but the default value is 0 but mkcflox/vmcflow
+    # have a default redshift of 0 in XSPEC 12.10.0 model.dat
+    #
+    if modelcls in [xs.XSismabs, xs.XSkerrd, xs.XSmkcflow, xs.XSvmcflow]:
+        # perhaps should check that all values == 0 so that
```
olaurino: I agree with checking the zeroness of the returned value, so we would also check against pesky build issues. Although I have no idea how the specific models are implemented and whether what I said makes any sense for these models.

sherpa/astro/xspec/tests/test_xspec.py
DougBurke: Well, long-term I want to get rid of `assertFinite`, but replacing it in the current code base would just tweak other tests (due to the imprecisely-defined meaning of the `model` argument) and I did not want to jump down that rabbit hole with this PR. I could do if you would rather it get cleaned up now (not a huge amount of work)?

DougBurke: although, I did manage to bork my XSPEC 12.10.0 installation last night in tracking down some other installation, so would rather not make changes just at this time (as I can't easily validate them).

DougBurke:
@olaurino - I would prefer to merge this PR as is (in large part because it would take me time to fix up my local set up so I could safely make any changes). 

We can address the two points you made with later PRs.

PR #534, opened 2018-10-24
XSPEC 12.10.1 support and mtable fix

DougBurke:
# Release Note
Sherpa now supports XSPEC 12.10.1. Several issues were fixed to ensure compatibility with this release. Note that there are known issues with version 12.10.1 (no patches), so at least v12.10.1a must be used.

# Notes
I had labelled this as WIP since I hoped HEASARC were going to make the `tabint` routine accessible from C (the current PR uses the C++ interface), but 12.10.1b has been out long enough now (and the current Government shutdown) means that it's not clear if it is worth waiting.

# Summary

Allow Sherpa to be built against XSPEC 12.10.1 and provide access to the new models, namely:

    XSzbknpower
    XSzlogpar
    XSagnsed, XSqsosed
    XScph, XSvcph
    XSrelline, XSrelline_lp, XSrelline_lp_ext
    XSkyrline

Low-level support is provided for the new convolution models - kyconv, tdrelconv, tdrelconvlp, tdrelconvlpext - but there are no Python classes that will let users use these functions.

Switched to using the new (to XSPEC 12.10.1) routines for handling table models.

Note that XSPEC 12.10.1 has changed the "models-only" approach so that it will process the users ~/.xspec/Xspec.init or (if missing) the system $HEADAS/../spectral/manager/Xspec.init file when the code is initialized (first time a routine from XSPEC is called). This could cause surprising results for users who are not aware of it (in part because the default photo-ionization cross section changes from `bcmc` to `vern`).

It is strongly recommended that XSPEC 12.10.1 not be used; please ensure that it is patched to at least XSPEC 12.10.1a.

# Details

Developing this PR did reveal several issues (in early versions of XSPEC 12.10.1 and in our interface to models). These have been fixed (either by patches to XSPEC 12.10.1 or with other PRs).

Changed to use a single routine for both additive and multiplicative table models (`tabint`) rather than two routines (`xsatbl` and `xsmtbl`), as using the old routines causes a crash (XSPEC is not initializing a variable that may be used by these routines). This was worked around for us in XSPEC 12.10.e by a patch to XSPEC (used in CIAO 4.11) but that was a short-term solution. In XSPEC 12.10.1 the `tabint` routine is the solution. Note that a macro is still used to create the symbol in `_xspec.cc`, to reduce surface changes, although perhaps a better solution could be developed.  Note that exponential table models could be supported with this interface, but it would need a test case (from what I have seen it is up to the user to know which are `mtable` and which `etable`, so it could require a change in the user API). As we don't support etable in the existing code base I do not think it is a reason to hold up this PR.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -112,6 +112,11 @@ int _sherpa_init_xspec_library();
 
 extern "C" {
 
+#ifdef XSPEC_12_10_1
+void agnsed_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+void qsosed_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+#endif
+
 #ifndef XSPEC_12_9_1
 void xsaped_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
 void xsbape_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
@@ -166,6 +171,10 @@ void jet_(float* ear, int* ne, float* param, int* ifl, float* photar, float* pho
   
 void xsgrbm_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
 void spin_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+
+#ifdef XSPEC_12_10_1
+void kyrline_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+#endif
 
 #ifndef XSPEC_12_9_1
 void xslorz_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
@@ -302,6 +311,12 @@ void ismabs_(float* ear, int* ne, float* param, int* ifl, float* photar, float* 
 void slimbbmodel(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
 #endif
 
+#ifdef XSPEC_12_10_1
+void tdrelline_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+void tdrellinelp_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+void tdrellinelpext_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+#endif
+
 // XSPEC table models
 void xsatbl(float* ear, int ne, float* param, const char* filenm, int ifl, 
 	    float* photar, float* photer);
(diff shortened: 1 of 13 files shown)


sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -404,9 +408,8 @@ int _sherpa_init_xspec_library()
     FPATDV(atomdbVersion1210);
 #endif
 
-  } catch(...) {
+  } catch(std::exception& e) {
```
DougBurke: Oops - I didn't mean to commit this change.

DougBurke:
I believe - but need to look into it more - that there's a problem with multiplicative table models in Sherpa (there's definitely something odd,just need to check that the test I added that is failing with pre XSPEC 12.10.1 builds is actually valid)

DougBurke:
Darn it. I'd assumed that was going to work...

DougBurke:
About to rebase to clean up the commit history slightly. Original tip was 8667254

DougBurke:
About to rebase again; this time to move the table-model fixes first. Original tip was c68ace4.

DougBurke:
About to rebase (which will remove some of the early pre-12.10.1 fixes developed during this PR as they have now been merged in #539). Original tip was 07ae493

DougBurke:
About to rebase. Tip was 9ef52ad (although this message isn't really needed anymore)

DougBurke:
I've closed #536 as I no longer have the laptop available, so can't continue investigation of the possible issues with ubuntu and recent gcc versions.

DougBurke:
About to rebase to pick up #582 fix. From local testing it looks like the changes in #520 cause problems - which means we need to look at these values / tolerances more closely for this regression test - but we won't see this on Travis until we get an XSPEC 12.10.x (for x=1 and maybe x=0?) build

DougBurke:
I've added a commit that fixes the cross-section table and loosens the tolerance for some of the parameters to address the problem I discuss above https://github.com/sherpa/sherpa/pull/534#issuecomment-453138816 - this is commit 01b7499

DougBurke:
I'll rebase this once #588 has landed.

DougBurke:
I am not sure if it is a problem or not, but looks odd. Maybe it's a remnant of the way `setup.py` was set up way back?

olaurino:
It's not a problem for the changes we are making. It might mean we are not testing some `INSTALL_TYPE`s  properly, but that's something we can check later.

DougBurke:
I agree it's not a relevant issue for this PR (or related ones).

DougBurke:
So we have a successful run on Travis with XSPEC 12.10.0. I'll rebase once #590 has made it to master so we have 12.10.1 builds on Travis.

DougBurke:
I've rebased this onto master to get the latest travis changes. I have not tried to cherry pick the #591 commit.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: Although I think it might make sense to be able to provide the actual error message to the user, since it may not be due to the HEADAS env. variable setting.

DougBurke:
@olaurino I agree with you that there is a lot of duplication in the templates used to wrap the XSPEC functions. I am not enough of a C++ user to be the person to attempt to clear this up ;-)

Issue #490, opened 2018-09-10
Should option parsing of ~/.sherpa.rc be more robust (i.e. not require all sections)?


DougBurke:
I created a simple Sherpa configuration file, intending to over-ride *only* the `options.plot_pkg` setting:

```
% cat ~/.sherpa.rc 
[options]
plot_pkg : pylab
```

Unfortunately our configuration-parsing code doesn't like this:

```
% python -c 'import sherpa.ui'
Traceback (most recent call last):
  File "/opt/ciao-4.10/ots/lib/python3.5/configparser.py", line 1135, in _unify_values
    sectiondict = self._sections[section]
KeyError: 'statistics'

During handling of the above exception, another exception occurred (I'm using the CIAOX branch here but master branch should perform very similarly):

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/ciao-4.10/lib/python3.5/site-packages/sherpa-4.10.0+28.g6692c01-py3.5-linux-x86_64.egg/sherpa/ui/__init__.py", line 20, in <module>
    import sherpa.all
  File "/opt/ciao-4.10/lib/python3.5/site-packages/sherpa-4.10.0+28.g6692c01-py3.5-linux-x86_64.egg/sherpa/all.py", line 37, in <module>
    from sherpa.fit import *
  File "/opt/ciao-4.10/lib/python3.5/site-packages/sherpa-4.10.0+28.g6692c01-py3.5-linux-x86_64.egg/sherpa/fit.py", line 38, in <module>
    from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, \
  File "/opt/ciao-4.10/lib/python3.5/site-packages/sherpa-4.10.0+28.g6692c01-py3.5-linux-x86_64.egg/sherpa/stats/__init__.py", line 48, in <module>
    truncation_flag = config.get('statistics', 'truncate').upper()
  File "/opt/ciao-4.10/ots/lib/python3.5/configparser.py", line 778, in get
    d = self._unify_values(section, vars)
  File "/opt/ciao-4.10/ots/lib/python3.5/configparser.py", line 1138, in _unify_values
    raise NoSectionError(section)
configparser.NoSectionError: No section: 'statistics'
```

Should we allow config files which only define a subset of values (this would make them more "future proof", in the sense that if we add a new configuration option/section then old `.sherpa.rc` files should continue to work)?

DougBurke:
[...]

PR #537, opened 2018-10-30
remove verbosity.level option and tracebacklimit hack

olaurino:
# Release Note
The `verbosity.level` option in the `sherpa.rc` and `sherpa-standalone.rc` files has been removed. This option has been causing issues with several versions of IPython including IPython 6.5 and 7. When the option is set to `0` then a chain of exceptions is triggered by any legit exception and in certain case the IPython console crashes. Standalone users should be fine. CIAO users relying on the default `0` in the `sherpa` shell should also be fine. CIAO users setting the option to anything but `0` and the `sherpa` shell would now be required to set `sys.tracebacklimit` manually, as documented for `chips`.

# Notes

See https://github.com/ipython/ipython/issues/9978#issuecomment-433428148 and https://github.com/ipython/ipython/issues/9978#issuecomment-433434648 for a description of the issue, which does not have anything to do with Sherpa.

### DIFF
--- a/sherpa/sherpa-standalone.rc
+++ b/sherpa/sherpa-standalone.rc
@@ -22,10 +22,6 @@ trunc_value: 1.0e-25
 minimum_energy: 1.0e-10
 
 [verbosity]
-# Sherpa Chatter level
-# a non-zero value will
-# display full error traceback
-level      : 10000
 # NumPy arrays have a threshold size print option, such that when an
 # array greater than that size is converted to a string, for printing,
 # only the first few and last few elements are shown.  If the array is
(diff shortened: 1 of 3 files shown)


DougBurke:
LGTM but I'm checking it out to just try it out manually.

olaurino:
I had an issue when testing CIAOT as I had the "new" sherpa.rc with verbosity.level commented out, and the sherpa shell would not even start. We should probably leave the verbosity.level setting in the default rc file rather than removing it, so it works for both new and old versions of CIAO/Sherpa.

DougBurke:
See #490 ;-)

Issue #538, opened 2018-11-04
XSismabs model is wrong


DougBurke:
The `XSismabs` model is a multiplicative-style XSPEC model, and the XSPEC class itself is correctly identified - i.e.

```
In [1]: from sherpa.astro import xspec                                          

In [2]: mdl = xspec.XSismabs()                                                  

In [3]: isinstance(mdl, xspec.XSMultiplicativeModel)                            
Out[3]: True

In [4]: isinstance(mdl, xspec.XSAdditiveModel)                                  
Out[4]: False
```

Unfortunately the `sherpa.astro.xspec._xspec.ismabs` routine is incorrectly marked as being an additive model - the `_NORM` variant of the template is used:

```
% grep ismabs, sherpa/astro/xspec/src/_xspec.cc 
  XSPECMODELFCT_NORM(ismabs, 31),
```

This means that before returning to Python, the model output is multiplied by the last element in the parameter list, which is the redshift value. The default value for this parameter is 0, which means that the model will always return 0 (although a non-zero redshift is likely to be set by the user).

PR #539, opened 2018-11-04
Fix XSPEC models: mtable table models and ismabs (fix #535 #538)

### LINKED ISSUE #535: XSPEC multiplicative table models are wrong
In investigating adding support for XSPEC 12.10.1 I noticed that XSPEC multiplicative table models were internally being treated as additive models, which means that the model was being multiplied by the last parameter value of the model. This is

a) going to make the normalization of the fit invalid
b) make the fit less stable (as the normalization is going to be coupled to the last parameter of the table model)

I believe I have fixed this in #534 (only relevant for XSPEC 12.10.0 and earlier).

The actual fix - which is the removal of "_NORM" from one line - (and tests) could be separated out of #534 if this was felt necessary.

DougBurke:
# Summary

This PR fixes:

- #535 XSPEC multiplicative table models (`mtable` models loaded with `load_xstable_model`)
- #538 incorrect evaluation of the `ismabs` model
- #540 the `kerrd` model evaluated to 0 for XSPEC 12.10.0 and later

For the first two, the models were being multiplied by the value of the last parameter of the model. At best this would lead to fits where the overall model amplitude was incorrect but could be corrected. However, it is likely that such fits were not robust (due to the accidental correlation between the overall amplitude and the last parameter of the model), and so it is not guaranteed that the best-fit would be found.

The kerrd model was due to a change in the naming of the XSPEC functions which had not been reflected in the XSPEC `model.dat` file in XSPEC 12.10.0 or 12.10.1b, which meant that the kerrd model in XSPEC 12.10.0 or later was wrong in Sherpa.

# Details

The fix for the first two errors is to switch from the `XXX_NORM` to `XXX` macro when wrapping the XSPEC-supplied functions (`xsmtbl_` and `ismabs_`). Tests were added to catch the errors (for `ismabs` it relies on the redshift parameter being at its default value of 0, which ends up ensuring that the model evaluates to 0 at all bins).

The kerrd fix requires conditional use of the `C_kerrdisk` routine prior to XSPEC 12.10.0 and `C_kerrd` for XSPEC 12.10.0 and later. This was confirmed by Craig Gordon of HEASARC in an email discussion.

There are no checks that the models are calculating the "correct" answers, since this is hard to do for Sherpa. The approach taken instead is to try and validate the answers (e.g. that the models evaluate to at least one bin with a value > 0).

As well as these changes, a test was added to ensure that the table model (whether additive or multiplicative) can be evaluated on energies outside the model-defined range. This is done because this situation caused crashes with XSPEC 12.10.0 (as the XSPEC code was using an uninitialized variable).

The commits are arranged to try and show a failing test and then the next commit fixes it. This breaks down at the end where I made a stupid mistake which required another commit (i.e. 4ebf672 fixes 552aec0), but this fails because of #540 which needed to be fixed in a4bafbb and then tested in eccda95. Note that there is currently no test that this fixes the behavior for XSPEC releases newer than 12.9.1, since the Travis builds do not have a more-recent XSPEC version available for testing.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -3162,7 +3162,7 @@ class XSkerrd(XSAdditiveModel):
 
     """
 
-    __function__ = "C_kerrdisk"
+    __function__ = "C_kerrd" if equal_or_greater_than("12.10.0") else "C_kerrdisk"
 
     def __init__(self, name='kerrd'):
         self.distance = Parameter(name, 'distance', 1., 0.01, 1000., 0.0, hugeval, 'kpc', True)
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -44,7 +44,7 @@ import six
 from six.moves import xrange
 
 import string
-from sherpa.models import Parameter, ArithmeticModel, modelCacher1d, RegriddableModel1D
+from sherpa.models import Parameter, modelCacher1d, RegriddableModel1D
 from sherpa.models.parameter import hugeval
 
 from sherpa.utils import guess_amplitude, param_apply_limits, bool_cast
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -1250,7 +1250,7 @@ static PyMethodDef XSpecMethods[] = {
 #ifdef XSPEC_12_10_0  
   XSPECMODELFCT_NORM(jet, 16),
 #endif
-  XSPECMODELFCT_NORM(ismabs, 31),
+  XSPECMODELFCT(ismabs, 31),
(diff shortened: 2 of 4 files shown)


DougBurke:
The last two commits are "expected" to fail - 552aec0 because I had forgotten the `return` line, and 4ebf672 because it turns out `kerrd` only appears to return 0's for XSPEC 12.10.0 and later (email sent to the HEASARC folks).

DougBurke:
Would you want this rebased/squashed (mainly to merge 552aec0 and 4ebf672 ) or are you happy with this as is?

olaurino:
This is small enough that it doesn't really matter, imo.

olaurino:
Note, I am going to merge a bunch of PRs into master, so they'll be in CIAOD. However, I will only cherry-pick the ones that need to go into CIAO 4.11, and in any case not now as I haven't had a final word on what needs to go there. Let's make sure to have that conversation before the next CIAOT is cut.

PR #546, opened 2018-11-13
Small code cleanup in LevMar

cdeil:
This PR contains a small code cleanup in the LevMar class to match the other classes in that file.

In  69b5484 the docstring indentation is fixed to be the usual.

In  5fe6551 I changed `optfcts.lmdif` to `lmdif`. Both are in the module namespace and are the same object, but `lmdif` is explicitly imported at the top. I'm not sure where the `optfcts` symbol comes from in that file `sherpa/optmethods/__init__.py`, I think like this it is a bit clearer and more consistent with the other similar init methods in that file.

### DIFF
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -441,7 +441,7 @@ class LevMar(OptMethod):
 
         """
     def __init__(self, name='levmar'):
-        OptMethod.__init__(self, name, optfcts.lmdif)
+        OptMethod.__init__(self, name, lmdif)
 
 
 class MonCar(OptMethod):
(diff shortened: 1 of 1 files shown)


DougBurke:
Thanks for this. Thanks to ADASS and the forthcoming CIAO release we are currently holding off what is going in to master (just in case you wonder why it doesn't get applied immediately :-)

olaurino:
Oh but this is simple, so I am ready to merge it as soon as Travis is happy. I still need to get back to the gammapy folks regarding the PSF issue. I have a simple fix, but I was hoping I could provide them with a decent workaround before we cut a new release.

olaurino:
Thanks @cdeil!

PR #551, opened 2018-11-15
fix gammapy/gammapy#1905, catch errors with psf pixel size

olaurino:
# Release Note
Sherpa did not properly catch errors when checking the PSF pixel size matches the image pixel size. This has now been fixed.

# Notes
This originated from gammapy/gammapy#1905. Under normal circumstances I wouldn't issue the PR without a test, but I would like the fix to be reviewed as soon as possible. I'll work on the test hopefully today.

In the meantime @cdeil might confirm it fixes their issue.

Another thing I don't like is to have a blanket `catch`. I could only catch `AttributeError`, but I am concerned it would just kick the can a little further down the road but there are a number of things that can go wrong and are not `AttributeError`, so I am not sure where to draw the line. Input is appreciated.


sherpa/instrument.py
```diff
@@ -717,10 +717,13 @@ def get_kernel(self, data, subkernel=True):
...
+            try:
+                psf_pixel_size = self.kernel.sky.cdelt
+                data_pixel_size = data.sky.cdelt
+
+                if not numpy.allclose(psf_pixel_size, data_pixel_size):
+                    warnings.warn("NOTE: The PSF pixel size ({}) does not correspond to the Image Pixel Size ({})"
+                                  .format(psf_pixel_size, data_pixel_size
+                    ))
+            except:
+                warnings.warn("Cannot determine the PSF or Data pixel size, ignoring.")
```
DougBurke: If I were to see the warning:

```
Cannot determine the PSF or Data pixel size, ignoring.
```

I wouldn't know what was being ignored. I think this is meant to mean:

```
Unable to check whether the PSF and data pixel sizes match.
```

but maybe it's more (or less)?

sherpa/instrument.py
```diff
@@ -717,10 +717,13 @@ def get_kernel(self, data, subkernel=True):
...
-                ))
+            try:
+                psf_pixel_size = self.kernel.sky.cdelt
+                data_pixel_size = data.sky.cdelt
+
+                if not numpy.allclose(psf_pixel_size, data_pixel_size):
+                    warnings.warn("NOTE: The PSF pixel size ({}) does not correspond to the Image Pixel Size ({})"
+                                  .format(psf_pixel_size, data_pixel_size
+                    ))
+            except:
```
DougBurke: I personally would vote for being explicit and catching only  `AttributeError`

a) I have just wasted several hours on a project of mine because a typo in a block ended up being ignored because of a bare `except`

b) I don't really buy the concept of "other things than an AttributeError could go wrong here". At the moment we are only concerned with the attribute access. Sure, the return values could do something strange in the `allclose` call, but if that happens then I'd rather we see this error, as it is a change in behavior.

anetasie:
I agree with @DougBurke on having more informative  warning message.

olaurino:
@cdeil yes, I think that's what I'll do. Thanks for the feedback.

olaurino:
Added a test. I made sure locally that the test fails without the fix, but I am not force-pushing another change.

I think I addressed all the comments, so I am going to merge this as long as Travis is happy.

sherpa/instrument.py
olaurino: Done

sherpa/instrument.py
olaurino: I used the text suggested by @cdeil

PR #552, opened 2018-11-15
put back `verbosity.level` option in sherpa.rc

olaurino:
This PR reverts back some of the change in #537. In particular it introduces the `verbosity.level` option back in the `.rc` files for both CIAO and Standalone, with an additional comment that the option is ignored by Sherpa > v.4.10.1.

This way the configuration files can work with both old and new versions of Sherpa.

A much better approach would be #490, but we can't do that now.

### DIFF
--- a/sherpa/sherpa-standalone.rc
+++ b/sherpa/sherpa-standalone.rc
@@ -22,6 +22,11 @@ trunc_value: 1.0e-25
 minimum_energy: 1.0e-10
 
 [verbosity]
+# Sherpa Chatter level
+# a non-zero value will
+# display full error traceback
+# This option is only used by Sherpa up to version 4.10.1
+level      : 10000
 # NumPy arrays have a threshold size print option, such that when an
 # array greater than that size is converted to a string, for printing,
 # only the first few and last few elements are shown.  If the array is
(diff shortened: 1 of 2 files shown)


sherpa/sherpa-standalone.rc
```diff
@@ -22,6 +22,11 @@ trunc_value: 1.0e-25
 minimum_energy: 1.0e-10
 
 [verbosity]
+# Sherpa Chatter level
+# a non-zero value will
+# display full error traceback
+# This option is only used by Sherpa up to version 4.10.1
```
DougBurke: Technically (e.g. in case of a 4.10.2 release) this - and the version in `sherpa.rc` - should be more like

```
This option is not used by Sherpa 4.11 but is left in so as to not cause problems
with systems that may switch between different Sherpa versions (see #490)
```

Even if we fix #490 we can't remove these settings until the pre-#490 fixed code is no-longer in common use.

I didn't change the PR in case this is "too much" and because I didn't want to start a new set of travis tests.

DougBurke:
I'm fine with merging this (assuming that the travis tests pass). I've added a note about how I think the comment could be improved (e.g. for when we look at this in a year's time and have forgotten all about this issue), but I am not requiring it.

PR #557, opened 2018-11-21
fix the nfev for NelderMead opt

dtnguyen2:
A fix to the original bug report:

https://icxc.harvard.edu/pipe/sherpadev/2018/0597.html


dtnguyen2:
Modified Aneta's original script by changing the call to fit from
fit()
to
fit(outfile='tmp.txt', clobber=True)
(sherpa3) [dtn@devel12 tmp]$ diff script.py /data/sherpa/TestIssues/NormBeta/script.py 
33,34c33,34
< fit(outfile='tmp.txt', clobber=True)
< #conf()
---
> fit() 
> conf()
37,39c37,41
< #eqwidth(bg,bg+NeIXr)
< #eqwidth(bg,bg+NeIXi)
< #eqwidth(bg,bg+NeIXf)
---
> eqwidth(bg,bg+NeIXr)
> eqwidth(bg,bg+NeIXi)
> eqwidth(bg,bg+NeIXf)
> 
> 
(sherpa3) [dtn@devel12 tmp]$ python script.py
WARNING: Gauss-Kronrod integration failed with tolerance 2.22045e-16, trying lower tolerance...
integration succeeded with tolerance 1.19209e-07
Dataset               = 1
Statistic             = cstat
Initial fit statistic = 362.018
Final fit statistic   = 77.2516 at function evaluation 816
Data points           = 69
Degrees of freedom    = 64
Probability [Q-value] = 0.123634
Reduced statistic     = 1.20706
Change in statistic   = 284.766
   bg.c0          86.9056     
   NeIXr.pos      13.4503     
   NeIXr.ampl     20.3681     
   NeIXi.ampl     3.32053     
   NeIXf.ampl     10.4619

Sherpa now prints 816 nfevs.  I believe this is correct since the file tmp.txt (remember fit was called with fit(outfile='tmp.txt', clobber=True));

(sherpa3) [dtn@devel12 tmp]$ grep -v nfev tmp.txt | wc
    817    5719   74347

The difference in one nfev is because of the following code in the class FitResults class:

class FitResults(NoNewAttributesAfterInit):
    def __init__(self, fit, results, init_stat, param_warnings):
        _vals = fit.data.eval_model_to_fit(fit.model)
        _dof = len(_vals) - len(tuple(results[1]))

That is the first line of the __init__ method of the FitResults class calls the function one more time (note: outside of the optimization method).  The model function doesn't actually get evaluated since the cached value is used. 

(sherpa3) [dtn@devel12 tmp]$ tail -1 tmp.txt
8.160000e+02 7.725162e+01 8.690560e+01 1.345033e+01 2.036813e+01 3.320532e+00 1.046188e+01
(sherpa3) [dtn@devel12 tmp]$ grep "8.690560e+01 1.345033e+01" tmp.txt
8.150000e+02 7.725162e+01 8.690560e+01 1.345033e+01 2.036813e+01 3.320532e+00 1.046188e+01
8.160000e+02 7.725162e+01 8.690560e+01 1.345033e+01 2.036813e+01 3.320532e+00 1.046188e+01

In theory one could add 1 to the printed final nfev but this is a choice for SDS to decide.

PS  I have also made some changes to the file optmethod/tests/tstopt.hh since I did not properly merged the code to test the optimization routines outside of our smoke test.

DougBurke:
So, is the problem that the existing code over-reports `nfev` (because it includes evaluations when the parameter was outside the parameter limit)?

dtnguyen2:
yes

DougBurke:
@olaurino - I have looked at the code, but I think this is one of those cases where SDS aren't really the best people to review a PR, so I'd prefer it if you could review this one

olaurino:
I was missing a lot of context, so while this is a small change it took me a while to review it. Some info below should help me and other people reading this, now and possibly in the future.

First of all there is a lot of noise because of the changes to the tests. I now understand these tests benchmark the optimization routines along specifically shaped loss functions. Whether we should include such tests in our automated tests is a question for another day.

The actual change in this PR is seemingly simple: NeldearMead sometimes proposes points in the parameter space that are outside the defined boundaries. When that happens I believe we force the loss function to assume the value of infinity so it becomes a "wall" to the simplex. However, we were counting that as a function evaluation, while one might argue that it's not, so this PR counts such occurrences and removes them from the `nfev` figure.

There are a bunch of things I am concerned about, and they go beyond the simplicity of the change.

First, it's not clear this is a regression or that the PR "fixes" it. When running the original script (which should be included in the PR description for reference) I cannot reproduce the exact number of function iterations we were having with CIAO 4.10. There might be a very good reason for this, in the end we replaced the fortran optimizer with the C++ one. However it is not clear whether we are changing the semantics of `nfev` with this PR or we are indeed "fixing a regression" in that we are changing the semantics back to what it was supposed to be.

Also, it looks like by the time the code counts the evaluation as outside the bounds there is no way to prevent the models from being evaluating. In some of the benchmarks there is a factor of ten difference in the `nfev` with or without the out-of-bounds evaluations, which are taking CPU time to perform anyway.

So I have some open questions:
   1. Could we make sure the stat value goes to infinity when the parameters are outside the boundaries *without* evaluating the models? I couldn't find a way to do it.
   2. Assuming we can't change the behavior in 1., does it make sense to remove the hits outside the boundaries from `nfev` in the first place? If I was a user I'd like to know that the optimizer is spending 95% of the time outside of the boundaries of the parameter space, rather than thinking that 5% of the actual evaluations took a long time for no reason. Granted, that's probably a pathological case, but my question is really about the definition of `nfev` and the information we'd hide from the user by removing some evaluations from `nfev`.
   3. Why exactly are we not getting the same number of iterations as CIAO 4.10 even after the fix? There is probably a good reason, but since this PR's goal was to "fix" this to begin with, if it doesn't fix it, then I'd go back to asking question 2.

There is an additional and more general issue I discussed with @dtnguyen2 regarding how to deal with the parameters at the boundary, as what we do in 1. is not perfect either way, but I am sure this is not in the scope of this PR.

dtnguyen2:
After a long discussion with @olaurino where he pointed out that minim still calls the stat function if any of the pars are outside the limits I made the changes to be consistent with my version of NelderMead

olaurino:
@dtnguyen2 I could not figure out whether returning from that function was enough, as I couldn't trace back which callback would get called and whether such callback would need to change the values of the parameters. It's useful for me to get to know this part of the code, but I can't claim I understand the chain of calls. I'll try running the "unit" tests to see if there are any changes in the optimization. And while I don't think I want to spend too much time on this, I'd still prefer to see that the number of model evaluations has changed as well.

olaurino:
It looks like there are significant differences in some of the benchmarks when adding the return statement. The differences are *only* in the benchmarks impacted by the `nfev` "fix", but somewhat surprisingly not all of them (`GoldsteinPrice` and `Michalevicz5` are the only exceptions) and so potentially most if not all cases where the parameters go out of bounds could be impacted by the change.

``` diff
< Minim_BrownBadlyScaled  572     0       1.10634e-17     1e+06,1.99977e-06,1e+06,1.99819e-06,1e+06,2e-06
---
> Minim_BrownBadlyScaled  -1788   0       1.9515e+12      81515.1,595289,173981,168207,792843,1.55898e-06
36,38c36,38
< Minim_McCormick 1287    -1.91   -nan    -nan,-nan
< Minim_BoxBetts  84      0       1.02735e-14     1,10,1
< Minim_Paviani   -76813  -45.7   -2.75503        5.78989,7.49795,5.06157,7.23616,4.91172,5.56942,7.55769,6.10409,6.90818,7.98112
---
> Minim_McCormick 76      -1.91   -1.91322        -0.547198,-1.5472
> Minim_BoxBetts  -2      0       1.79769e+308    1.1,9.9,1.39
> Minim_Paviani   1604    -45.7   -45.7785        9.35027,9.35027,9.35027,9.35027,9.35027,9.35027,9.35027,9.35027,9.35027,9.35027
44,46c44,46
< Minim_Levy5     -312    -11.504 38.8617 3.96386,3.99615,3.99623,3.99624,3.99945
< Minim_Levy6     -300    -11.504 47.8504 3.96386,3.99615,3.99623,3.99623,3.99624,3.99945
< Minim_Levy7     -376    -11.504 56.8392 3.96386,3.99615,3.99623,3.99623,3.99623,3.99624,3.99945
---
> Minim_Levy5     -230    -11.504 38.8617 3.96386,3.99615,3.99623,3.99624,3.99945
> Minim_Levy6     -363    -11.504 47.8504 3.96386,3.99615,3.99623,3.99623,3.99624,3.99945
> Minim_Levy7     -430    -11.504 56.8392 3.96386,3.99615,3.99623,3.99623,3.99623,3.99624,3.99945
48,49c48,49
< Minim_SixHumpCamel      86      -1.03   -1.03163        0.089842,-0.712656
< Minim_Branin    77      0.397889        0.397887        9.42478,2.475
---
> Minim_SixHumpCamel      101     -1.03   -1.03163        -0.089842,0.712656
> Minim_Branin    85      0.397889        0.397887        9.42478,2.475
57c57
< Minim_Easom     -19     -1      -0      25.55,25.55
---
> Minim_Easom     33      -1      nan     -nan,-nan
```

The columns are `name, nfev, answer, stat, par`.

Note how some benchmarks that didn't converge now converge, and some that did now don't. Note also that in some cases the number of function evaluations is very different. And the final results (minimum, parameter values) are different in many cases.

So, it looks like adding the return statement makes a big difference. It could be because it's changing the way the out-of-bounds case is handled, or simply because it's not doing what it is supposed to do. As I am not familiar with the individual benchmarks I have no way to check what the true minimum should be.

However I ran the original script posted by @anetasie on the modified version of the code (*with* the return statement) and I get exactly the same fit as the one posted by @dtnguyen2 , so *without* the return statement.

So other options are that some benchmarks may have issues of their own *or* that @anetasie's original script does not trigger any differences when the return statement is there. The latter seems unlikely as all of our tests seem to pass in this PR, but I can't tell how many of our tests hit the out-of-bounds case.

I also changed the code to print how many times models are evaluated by instrumenting the `modelCacher1d` decorator's wrapper function, so caching doesn't impact the count. Without the return I get 3276 model evaluations, which seems to make sense, there are 4 models in the model expression, so they are evaluated `3276/4=819` times, which is close enough to the reported 816 function evaluations in the fit (I expect there may be few more evaluations from outside the fit). When I run the script *with* the return statement (current HEAD of this PR) I get again 3276 hits. That seems to imply that even with the return statement the models are always evaluated. Again, this is not important for this PR but it seems to suggest that when NelderMead is run and the walk gets out of the user-defined bounds there may be a number of model evaluations that could be spared.

I cannot tell whether the return statement is doing what it's supposed to do, as in particular it doesn't prevent useless model evaluations from happening. And in general I am not sure this PR is "fixing" the change @anetasie originally reported, as the number of function evaluations probably changed for reasons independent from the one this PR is fixing, and this PR is simply "hiding" part of the evaluations that are actually happening.

dtnguyen2:
In the latest version, not yet checked in, I get the following results for the BrownBadlyScaled function.

devel12-185: minim | grep BrownBadlyScaled
Minim_BrownBadlyScaled	4547	0	1.10634e-17	1e+06,1.99977e-06,1e+06,1.99819e-06,1e+06,2e-06

The minim with the return gives better results for McCormick and the Paviani functions.  However, the previous version of minim, the one without the return term, does give better results for the BoxBetts and Easom functions.  I'm still working on resolving this.  As for the remaining functions such as Levy[5-7], SixHumpCamel, Branin: the two versions give identical results with the exception of the nfevs.  So I decided to put the following code 

    std::cout << "ninarB\n";

in the function Branin. As you can see:

devel12-170: minim | grep ninarB | wc
     85      85     595

devel12-192: minim | grep Branin
Minim_Branin	85	0.397889	0.397887	9.42478,2.475

The nfevs in the version with the return gives the correct number of function evaluations.

dtnguyen2:
Looking at the results from Omar's diff, it looks like the no-return version is
better for the BrownBadlyScaled, BoxBetts ad Easom funcdtions but not as good
for the McCormik and Paviani functions. Never mind for now the differences
in the nfevs between the two versons, with and without the return statement in
the eval_usr_func.  The goal is to get minim to return the best results for all
the aforementioned test functions.

The current version, correctly gets the answer for all the aforementioned test
function except for the Paviani (at minimum, all the parameters are supposed to
be around 9.35).

devel12-110: minim | grep BrownBadlyScaled
Minim_BrownBadlyScaled	2151	0	7.60681e-20	1e+06,1.99991e-06,1e+06,2.0001e-06,1e+06,2e-06

devel12-111: minim | grep McCormick
Minim_McCormick	93	-1.91	-1.91322	-0.547198,-1.5472

devel12-112: minim | grep BoxBetts
Minim_BoxBetts	94	0	8.5606e-16	1,10,1

devel12-113: minim | grep Paviani
Minim_Paviani	1497	-45.7	-45.6242	9.32565,9.39821,9.34252,9.46265,9.35731,9.36868,9.28165,9.299,9.40574,9.21271

devel12-114: minim | grep Easom
Minim_Easom	-4096	-1	-0	25.55,25.55


As for the nfevs,  the function eval_usr_func now calls the function
reflect_about_boundary to handle the boundary conditions.  The new function
guarrantees that the parameters are within bounds so there the original program
will handle the nfev calculations.  I did a spot check to make sure that the
nfevs are done correctly:

(sherpa3) [dtn@devel12 src]$ minim | grep delacSyldaBnworB | wc
   2151    2151   36567

(sherpa3) [dtn@devel12 src]$ minim | grep kcimroCcM | wc
     93      93     930

(sherpa3) [dtn@devel12 src]$ minim | grep stteBxoB | wc
     94      94     846

(sherpa3) [dtn@devel12 src]$ minim | grep inaiavaP | wc
   1497    1497   13473

(sherpa3) [dtn@devel12 src]$ minim | grep mosaE | wc
   4096    4096   24576

$$$
This section is probably more of an interest for catdev.

I have added the McKinnon function, see the following paper:

https://pdfs.semanticscholar.org/7015/b0113f85db63ebbd3a1fe799ac367f3abc35.pdf

since it does show behaviors akin to where the centroid was found to be outside
the ECF90REG.

If the starting parameters for the Easom function are: x[0] = x[1] = 25, then
the resulting fit does not wander very from the initial starting parameters:

devel12-114: minim | grep Easom
Minim_Easom	-4096	-1	-0	25.55,25.55

However, if the initial parameters are set at: x[0] = x[1] = 6.4, then the
nelder-mead algorithm does indeed converge to the correct global minimum:

(sherpa3) [dtn@devel12 src]$ minim | grep Easom
Minim_Easom	91	-1	-1	3.14159,3.14159

However, for the McKinnon function the starting paramaters are:
x[0] = x[1] = 1.0 which is not too far from the global minimum of
f( 0.0, -0.5) = - 0.25 and yet the final fitted parameters are closed to
limits.

(sherpa3) [dtn@devel12 src]$ minim | grep McKinnon
Minim_McKinnon	-4098	-0.25	-2.376e+11	-100,-100

I should probably trace the behavior of the MLE fitted which were sent to QA
see if they have the same behavior as the McKinnon function.
$$$

anetasie:
Reading the full discussion it is not clear what is the current status of this PR. Do we need to run any more tests?

dtnguyen2:
Not sure what you are asking. I did a few checks to make sure the number of function evaluations are correct which was the original problem, along the way I improved the algorithms to solve more test problems.

anetasie:
@dtnguyen2 who needs to review and sign off for merging the code into master?

dtnguyen2:
Anyone not named dtn

DougBurke:
Last time I looked I didn't feel qualified to offer substantive comments on the code, so punted to @olaurino . I would guess given that we are focusing on a 4.11.0 release it will have to wait until that has calmed down a bit.

dtnguyen2:
Any implementation of any optimization algorithm should never return NaN, yet if one looks closely at the results of the minim (before this PR), you will see that for the McCormick function returns nan:

< Minim_McCormick 1287    -1.91   -nan    -nan,-nan

which hopefully we all agree is not ideal. If you look closely at lines 393-396 and 582-585 of sherpa/optmethods/optfcts.py contain the following code:

```
    def stat_cb0(x_new):
        if _my_is_nan(x_new) or _outside_limits(x_new, xmin, xmax):
            return FUNC_MAX
        return orig_fcn(x_new)
```

so back in the day when minim was in Fortran (ie when the dinosaurs were roaming the earth) I did not bother to debug the root cause of the problem and decided to implement a work around instead.  To me, a side benefit from this PR is that the mis-behavior is now fixed.

sherpa/optmethods/tests/tstoptfct.hh
```diff
@@ -2,7 +2,7 @@
 #define tstoptfct_hh
 
 // 
-//  Copyright (C) 2007, 2017  Smithsonian Astrophysical Observatory
+//  Copyright (C) 2007, 017, 2019  Smithsonian Astrophysical Observatory
```
DougBurke: The leading 2 has been lost from 2017. A highly-technical comment I know!

sherpa/optmethods/src/minim.hh
```diff
@@ -29,26 +29,41 @@ namespace sherpa {
...
+             SIMP, VC, IFAULT, NEVAL, LIMITS );
+      return;
     }
 
   private:
 
     Func usr_func;
     Data usr_data;
 
+    void reflect_about_boundary( int npar, std::vector<real>& par,
```
DougBurke: This is not a lien on this PR, but I assume this is not a general-purpose operation - i.e. it should only be used here, and not everywhere else we have code that calls `sherpa::Opt<Data, real>::are_pars_outside_limits`?

DougBurke:
@dtnguyen2 I thought the idea was for me to review this once you'd fixed the conflicts?

dtnguyen2:
I think I resolved the conflicts now

DougBurke:
@dtnguyen2 I don't know what's happened, but I see commits from PR 616 here. This doesn't look right.

dtnguyen2:
To much of mess to rebase and/or cherry pick so PR #706

PR #562, opened 2018-12-02
Add code and tests for "low degree-of-freedom behavior"

DougBurke:
# Note

Closed as some code has been accepted as part of #562 and the remainder was easier to include as a new PR #567

---

The idea is to document the current behavior, so we can work out how any changes we might make change the runtime behavior.

At the moment there is a simple check of Cash and CStat behavior with LevMar, NelderMead, and MonCar optimisers. This calls the `calc_stat_info` method of the `Fit` object, with dof=1, 0, -1. There are later additions to add tests of the `fit` method too.

**EDITED TO ADD**

- There are now tests of both `fit` and `calc_stat_info` methods.
- Some of these commits have been absorbed into #564 
- The PR is no-longer just added tests, since it now also includes code changes to address some of the issues brought up by the test (e.g. see #565)

PR #564, opened 2018-12-02
fix LevMar when dof < 0

dtnguyen2:


dtnguyen2:
I haven't added a test case for lmdif when dof < 0 (hey it is Sunday after all), will do it 'soon'.

Note: I normally run the full sherpa test before checking in the code, but I'm getting the following Xvfb error:
...
INTERNALERROR>     '\n    '.join(stderr.splitlines())))
INTERNALERROR> pytest_xvfb.XvfbExitedError: Xvfb exited with exit code 1
INTERNALERROR> Xvfb stdout:
INTERNALERROR>     
INTERNALERROR> Xvfb stderr:
INTERNALERROR>     _XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed
INTERNALERROR>     _XSERVTransMakeAllCOTSServerListeners: server already running
INTERNALERROR>     (EE) 
INTERNALERROR>     Fatal server error:
INTERNALERROR>     (EE) Cannot establish any listening sockets - Make sure an X server isn't already running(EE)

and unfortunately the following command: pkill Xvfb has not effect.  I can change the warning/error msg to whatever SDS wants, after this is what was agreed by SDSl

DougBurke:
I've added issue #566 to talk about the test failure here, since it isn't related to this PR but appears to be an "instability" in the Travis tests.

olaurino:
argh, I had missed that PR. You added way more tests. So I'll just cherry-pick your commit(s) and add it to this PR instead of my quick test. Thanks!

DougBurke:
I did try to clarify the status of things in my sherpadev email yesterday!

Note that I have some actual code changes related to calc_stat_info that would clean up these tests and fix what I feel is an unresolved issue from your #486 fix. Can I push them to the same PR (and you just cherry pick the earlier commits)?

olaurino:
> Can I push them to the same PR (and you just cherry pick the earlier commits)?

Yes. you might want to rebase #562 at some point as I am squashing all the commits into one for clarity.

DougBurke:
I was going to say I'll wait to make my changes until you've pushed yours, so I don't "steal" the Travis job, but as you've just done that I don't need to wait ;-)

sherpa/optmethods/optfcts.py
```diff
@@ -721,18 +721,19 @@ def stat_cb1( pars ):
 
     x, fval, nfev, info, fjac = _saoopt.cpp_lmdif( stat_cb1, m, x, ftol, xtol, gtol, maxfev, epsfcn, factor, verbose, xmin, xmax, fjac )
 
-    fjac = numpy.reshape(numpy.ravel(fjac, order='F'), (m, n), order='F')
+    if info > 0:
```
DougBurke: @dtnguyen2 - so the change here is that we only calculate `covar` and update several variables (which make it into the return value later on) if `info` is positive (which presumably means that `cpp_lmdif` didn't fail?

sherpa/optmethods/optfcts.py
```diff
@@ -780,7 +781,10 @@ def stat_cb1( pars ):
     #     rv = (status, x, fval, msg, {'info': info, 'nfev': nfev,
     #                                  'njev': njev, 'covar': covar})
     # else:
-    rv = (status, x, fval, msg, {'info': info, 'nfev': nfev,
-                                 'covar': covar})
+    if info != 1:
```
DougBurke: The logic here is hard to follow, given how info is changed, but can we guarantee at this point that `covar` is defined? I think so, but this depends on `cpp_lmdif` only ever returning a value >= 0

sherpa/tests/test_fit_unit.py
```diff
@@ -2534,3 +2534,267 @@ def test_wstat_rstat_qval_fields_not_none():
...
+# Results from CIAO 4.10 on Linux Ubuntu. The fit results are
+# not consistent between different options (e.g. method and statistic),
+# which is not surprising given that the fit is not-well constrained.
+#
+# The only check here is whether the fit succeeds or not, since this
+# depends on the method.
+#
+@pytest.mark.parametrize("stat", [Cash, CStat, Chi2])
+@pytest.mark.parametrize("method,success",
+                         [pytest.param(LevMar, False,
```
DougBurke: We need the `pytest.param` part removed, since the test with `LevMar` shouldn't be skipped any more.

olaurino: Done.

dtnguyen2:
Omar asked me if I could sit down and explained the changes to him and
Warren.  So here is the explanation that I intend to give them:

lmdif has the following code:

      info = 0;
      iflag = 0;
      nfev = 0;

      /*     check the input parameters for errors. */

      if (n <= 0 || m < n || ldfjac < m || ftol < 0. || xtol < 0. ||
          gtol < 0. || maxfev <= 0 || factor <= 0.) {
    goto TERMINATE;
      }
...
      return info;
      //     last card of subroutine lmdif.

So in the Moritz case m < n, lmdif returns with an error code info=0.
Note, the internal C/C++ lmdif was not changed.
The first change of the code is the following two lines:
if (info > 0 ) {
and
}
in the following section:

    info = lmdif( usr_func, lm_usr_data, m, n, &x[0], &myfvec[0], ftol,
                      xtol, gtol, maxfev, epsfcn, &diag[0], mode, factor,
                      nprint, nfev, &fjac[0], ldfjac, &ipvt[0], &qtf[0],
                      &wa1[ 0 ], &wa2[0], &wa3[0], &wa4[0], bounds );

        if ( info > 0 ) {
          this->covar( n, &fjac[ 0 ], ldfjac, &ipvt[0], ftol, &wa1[0] );

          for ( int ii = 0; ii < n; ++ii )
            if ( fjac[ ii + ldfjac * ii ] > 0.0 )
              covarerr[ ii ] = sqrt( fjac[ ii + ldfjac * ii ] );
            else
              covarerr[ ii ] = 0.0;
        }

That is the covariance matrix does not need to be calculated if there was
no fit.  To me this logic is rather self-explanatory but I'm willing to
elaborate if necessary.
The 2nd change to the code is the following line (aside from the obvious
indentation for python):

if info > 0:

in the following section of optfcts.py:
    x, fval, nfev, info, fjac = _saoopt.cpp_lmdif( stat_cb1, m, x, ftol,
xtol, gtol, maxfev, epsfcn, factor, verbose, xmin, xmax, fjac )

    if info > 0:
        fjac = numpy.reshape(numpy.ravel(fjac, order='F'), (m, n),
order='F')

        if m != n:
            covar = fjac[:n, :n]
        else:
            covar = fjac

        if _par_at_boundary( xmin, x, xmax, xtol ):
            nm_result = neldermead( fcn, x, xmin, xmax,
ftol=numpy.sqrt(ftol), maxfev=maxfev-nfev, finalsimplex=2, iquad=0,
verbose=0 )
            nfev += nm_result[ 4 ][ 'nfev' ]
            x = nm_result[ 1 ]
            fval = nm_result[ 2 ]

Again, since a fit was never made there is no need to calculate the
covariance code. I'm going to get a little off topic here but I'm not happy
with how LevMar currently handles the parameters at the limits (lmdif is an
unconstrained method) etc....

lmdif used to return info as an integer in the range: [0, 8].  The
explanations for the info values were taken verbatim from lmdif code and
are reproduced  in lines 747-768 of the optfcts.py file.  However, SDS at
one point complained that they do not understand the wording of the
explanations and wanted a uniform error wording, hence the _get_sao_fit_msg.

The last code code change that I made is the following lines.  Note: info
has been translated from lmdif-error-code value to sao-error-code value via
the call to _get_sao_fit_msg

    if info != 1:
        rv = (status, x, fval, msg, {'info': info, 'nfev': nfev,
                                     'covar': covar})
    else:
        rv = (status, x, fval, msg, {'info': info, 'nfev': nfev})

Now that I have read the _get_saofit_msg, then line 784 should be change
from
   if info != 1:
to
   if info == 0:
That is only return the covariance matrix if the fit was successful.


In the fortran version of lmdif, the function mylmdif returns with info=0,
that is lmdif was never called, hence the covariance was never calculated
either.  So the changes that I have made, aside from changing info != 1 to
info == 0 should be basically the same as the Fortran version.

sherpa/optmethods/optfcts.py
```diff
@@ -743,6 +743,7 @@ def stat_cb1( pars ):
     if error:
```
DougBurke: Note: I believe the `error` array is also unused and could be removed.

DougBurke: but this is not important for this particular PR

PR #567, opened 2018-12-03
Harmonise the handling of dof is zero or negative in fit and calc_stat_info (fix #565)

### LINKED ISSUE #565: Handling of zero or negative degrees-of-freedom
We have a change in behavior of how "unconstrained" fits behave (e.g. those with a degree-of-freedom of 0 or less) compared to the 4.10 behavior, which is related to the replacement of the Fortran-based optimisation code with C++ code (see PR #481). This is the genesis of the tests added in #562. Here is an example of the change (and is related to #562 as the behavior is different to that checked for in the tests). In many ways I think the new behavior is better, since we get an answer rather than an error, with an invalid (`nan`) value for `rstat`, but perhaps we should also set `qval` to `nan` here too.

I note that @dtnguyen2 has just added #564 to address this (as I was writing this issue ;-), but this may be more related to the fit part of the issue, which the tests in #562 don't yet look at [**edited to add** the #562 tests now include fit, and have been partly merged into #564]

With the code

```
from sherpa.stats import Cash, Chi2, CStat
from sherpa.fit import Fit

d = Data1D('test', [1, 2, 3], [4, 5, 6], [1, 1, 1])
mdl = Polynom1D()
mdl.c1.thaw()
mdl.c2.thaw()
mdl.c0 = 5.1

for stat in [Cash, CStat, Chi2]:
    sobj = stat()
    print("# Stat = {}".format(sobj.name))

    fit = Fit(d, mdl, stat=sobj)
    try:
        statinfo = fit.calc_stat_info()
    except ZeroDivisionError:
        print("\nCAUGHT ZeroDivisionError\n")
        statinfo = None

    print(statinfo)
    print("")
```

WIth CIAO 4.10 I get

```
# Stat = cash
name      = 
ids       = None
bkg_ids   = None
statname  = cash
statval   = -18.2772161919084
numpoints = 3
dof       = 0
qval      = None
rstat     = None

# Stat = cstat

CAUGHT ZeroDivisionError

None

# Stat = chi2

CAUGHT ZeroDivisionError

None

```

With `master` I get

```
# Stat = cash
name      = 
ids       = None
bkg_ids   = None
statname  = cash
statval   = -18.2772161919084
numpoints = 3
dof       = 0
qval      = None
rstat     = None

# Stat = cstat
name      = 
ids       = None
bkg_ids   = None
statname  = cstat
statval   = 0.40863145212838614
numpoints = 3
dof       = 0
qval      = 1.0
rstat     = nan

# Stat = chi2
name      = 
ids       = None
bkg_ids   = None
statname  = chi2
statval   = 2.03
numpoints = 3
dof       = 0
qval      = 1.0
rstat     = nan

```

DougBurke:
Note: this is essentially a cherry-picked version of #562 since some of the commits were merged into #564. I am assuming that @olaurino squashed together 535694f 8f87e44 feb19b0 e136bf0 when creating 33d897a

# Release notes

The `fit` and `calc_stat_info` methods now use the same code to calculate the reduced statistic (`rstat`) and "quality of fit" (`qval`) values. This avoids a `TypeError` when the number of degrees of freedom is negative in `calc_stat_info` and ensures that `NaN` is returned for both values when the goodness-of-fit calculation has failed (e.g. the reduced statistic is zero or less or the statistic is negative), whereas in previous versions the `qval` value could be `None` or `1` depending on the code path.

# Implementation notes
Fix #565.

The calculation of `qval` and `rstat` are moved into a separate routine since the existing code in `sherpa/fit.py` had two different versions. I believe this code should really be in the `sherpa.stats.Stat` class, but that is outside the scope of this PR.

The tests added in #562 are adjusted to give the expected answers (`NaN` for both values, rather than `1.0` for `qval` with `calc_stat_info`) and the "expected fail" marker removed.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -27,16 +27,16 @@ import signal
 from functools import wraps
 
 from numpy import arange, array, abs, iterable, sqrt, where, \
-    ones_like, isnan, isinf, float, float32, finfo, nan, any, int
-from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, igamc, \
+    ones_like, isnan, isinf, float, float32, finfo, any, int
+from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, \
     bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import FitErr, EstErr, SherpaErr
 from sherpa.data import DataSimulFit
 from sherpa.estmethods import Covariance, EstNewMin
 from sherpa.models import SimulFitModel
 from sherpa.optmethods import LevMar, NelderMead
-from sherpa.stats import Chi2, Chi2Gehrels, Cash, CStat, Chi2ModVar, \
-    LeastSq, Likelihood, WStat
+from sherpa.stats import Chi2, Chi2Gehrels, Cash, Chi2ModVar, \
+    LeastSq, Likelihood
 
 warning = logging.getLogger(__name__).warning
 info = logging.getLogger(__name__).info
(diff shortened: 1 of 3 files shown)


DougBurke:
@olaurino - when you look at this I'd appreciate a careful check of the tests that are (in master) marked as `xfail` but are fixed by this PR. I am always concerned with blanket statements like "this test is expected to fail" since there can be a number of reasons why it is failing, and occasionally it's not what we expect. GIven that this was all thrown together quickly, and then there was some fun with partial merging of a rebased version of this PR, some fresh eyes would be appreciated.

olaurino:
I started reviewing this yesterday and I hope we can merge it today. My only issue with this PR is not something that has to do with the PR, but the pre-existing code (and the usage of `isinstance` to define per-class branches, which is weird. Let me see if I can refactor your changes in a more OOP fashion, as for better or worse we are doing OOP.

DougBurke:
I am happy for the implementation to be re-written. I think the tests are really the important part of this - in that I believe the tests make sense - so the important thing is that they should pass (and so as I mentioned above, making sure that the tests make sense is important ;-)

DougBurke:
So feel free to close this PR and open a new one if that is helpful - although note the very-minor code clean up in 56320fd if you go this route

olaurino:
If I checkout the `fit.py` file from master and run the tests from this PR they seem to fail for the "right" reason, i.e. `NaN` vs `1.0` and `igamc domain error, a and x must be positive`, which was caused by what we had already determined during the investigation leading to #564. So unless I am missing something I think the tests are fine.

And the fix is fine as well, I am just going to see if we can make the checks a little bit more encapsulated than they currently are (and were before your fix). I don't think I'll need to open a new PR and I'll just move things around a bit if it makes sense.

olaurino:
I did some refactoring. I would like to make sure the following is the desired behavior

  * cstat, wstat, and all chi2-derived statistics are supposed to have `rstat` and `qval` calculated (and `NaN` in certain conditions).
  * leastsq does not have `rstat` and `qval` calculated (so they are always `None`).
  * likelihood-derived statistics, i.e. cash, prior, and `UserStat` do not have `rstat` and `qval` (so they are always `None`).

My refactoring does not address what I perceive to be some deeper existing design issues. It's not for this PR to fix them.

DougBurke:
I won’t be able to look at this until the New Year. Aneta can answer
these/check they are sensible.

olaurino:
@DougBurke this would be ready to merge if we had an answer from SDS to my question.

DougBurke:
@anetasie - can you comment on the questions Omar asks in https://github.com/sherpa/sherpa/pull/567#issuecomment-449471099

anetasie:
It sounds ok.
Note 'prior' is not a statistics. Not sure where 'prior' is listed in this code.

DougBurke:
@anetasie the `CStat` and `WStat` symbols are not used in this module, so they were removed.  Really we shouldn't have to import any of these statistic classes in this module, but unfortunately the current design means we have to.

olaurino:
> Note 'prior' is not a statistics. Not sure where 'prior' is listed in this code.

Well, in Sherpa the `Prior` class is a descendant of `Stat`, so I had to include it in my analysis. Your statement seems to suggest this should not be the case, but this is how it was implemented.

anetasie:
@DougBurke sounds good.

PR #573, opened 2018-12-14
add release notes that were dropped

olaurino:
It looks like I made a mistake when I merged the 4.10.1 release into master, so I lost some changes to the release notes, which this PR puts back, along with the 4.10.2 release notes.

The reason I lost the commits is that I used the wrong switch on the `git` command line when merging the 4.10.1 release branch and fixing the conflicts.

For future reference here is a detailed description of the issue:
  * A while back we started cherry-picking merge commits from master into the release branches. Earlier we used to issue PRs against the release branch, which was inconvenient for several reasons.
  * However bumping the version number should not make it into the master branch until after the release is merged.
  * So I pushed a couple of commits to the release branch, including the release note changes (the other change was to the README, but that was superseded by later changes into master).
  * When I tried to merge the release branch into master I had some conflicts, and I believe I mistakenly used the `ours` merge strategy rather than the `ours` option in the `recursive` strategy, i.e. I did `git merge -s ours 4.10.1` instead of `git merge -X ours 4.10.1`.

The `ours` option in `recursive` makes sense because if you exclude the version bumping happening in `RELEASE_NOTES` and `README.md` all the changes in the release branch have already been merged into master, and any conflicts are just PRs in master that superseded the ones in the release branch. So the `ours` version should always win.

I guess this incident will make sure I never forget the warning in the git documentation:

> This should not be confused with the 'ours' merge strategy, which does not even look at what the other tree contains at all. It discards everything the other tree did, declaring 'our' history contains all that happened in it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,6 +74,11 @@ macos-python3.6-conda-build:
   variables:
       SHERPA_PYTHON_VERSION: "3.6.*"
 
+macos-python3.7-conda-build:
+  <<: *template_macos_conda_build
+  variables:
+      SHERPA_PYTHON_VERSION: "3.7.*"
+
 linux-python2.7-conda-build:
   <<: *template_conda_build
   variables:
@@ -89,6 +94,11 @@ linux-python3.6-conda-build:
   variables:
     SHERPA_PYTHON_VERSION: '3.6.*'
 
+linux-python3.7-conda-build:
+  <<: *template_conda_build
+  variables:
+    SHERPA_PYTHON_VERSION: '3.7.*'
+
 conda:
   stage: deploy
   tags:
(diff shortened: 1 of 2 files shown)


olaurino:
Not sure why one of the Travis jobs keeps stalling. Well, I guess it's Friday afternoon.

DougBurke:
I've given up counting the number of times I have to restart travis jobs (and is one of the reasons why I wanted the travis tests to fail more obviously if they couldn't run the conda install stages).

olaurino:
it's stalling *during* the conda install phase with a timeout. The only way to make it run is to make conda not-verbose, which however might make the job fail because of the log is too big...

olaurino:
In any case it looks like it got unstuck.

olaurino:
@DougBurke do you think this needs reviewing or should I just merge it?

DougBurke:
I've had a look and I say merge. I have not done any checking to see what may or may not have got lost, as I don't have the mental energy to track down the build before/after the problematic change to see what else could have lost. aka I'm trusting @olaurino 's investigative work here ;-)

PR #578, opened 2018-12-20
Cleanup readme

DougBurke:
# Summary

Remove build instructions from the README and point to the Sphinx documentation instead.

There are no functional changes in this commit, only documentation. Only users who look at the PyPI page  https://pypi.org/project/sherpa/ or the README via the GitHub "home page" are expected to be impacted.

# Details

The installation instructions in the Sphinx docs should be reviewed to see if they are correct and are missing anything in the README (I did add in the "how to pytest a single test" case since this was missing).

One thing I explicitly did not copy over was advice on how to set up a conda environment, as I feel that that can be better explained outside Sherpa (in other words it's not our job). The installation section was getting long, which could scare off people!


README.md
```diff
@@ -82,504 +63,41 @@ version. A copy of the GNU General Public License can be found in the
...
-In order to make Sherpa build its modules against a local `fftw` library, users
-need to change the default Sherpa build configuration as follows.
-
-First, make sure you download the Sherpa source tarball, or get the source
-code from GitHub:
-
-    $ git clone https://github.com/sherpa/sherpa.git
-    $ cd sherpa
+Sherpa is provided for both Linux and Mac OSX operating systems running
+Python 2.7, 3.5, 3.6, and 3.7. It can be installed with by saying
```
olaurino: Is this a typo or an English construct I am unaware of?

README.md
```diff
@@ -82,504 +63,41 @@ version. A copy of the GNU General Public License can be found in the
...
-for more information.
+Source installation is available for platforms incompatible with the
+binary builds, or for when the default build options are not sufficient
+(such as including support for the
+[`XSPEC` model library](https://heasarc.gsfc.nasa.gov/xanadu/xspec/)).
+The steps are described in the
+[building from source](https://sherpa.readthedocs.io/en/latest/install.html#building-from-source)
+documentation.
 
 History
```
olaurino: I believe we are missing a couple of Zenodo DOIs for the latest versions. That's something else I can take care of, and it's just easier to do it in this PR.

olaurino:
> The installation instructions in the Sphinx docs should be reviewed to see if they are correct and are missing anything in the README

I can see a couple of issues in the RTD docs re source builds w/ xspec support. First of all, with xspec 12.10.0 build instructions are dead simple if starting with a CIAO 4.11 installation, which it might make sense to include again. In that case you only need `xspec_version`, `xspec_lib_dirs`, and `xspec_include_dirs`.

Also, `XSFunctions` is misspelled a couple of times (an extra `y`).

`--enable-xs-models-only` is currently broken, as it requires to manually edit some build files to make it build everything Sherpa needs the way Sherpa needs it. Either we review the installation instructions (maybe they fixed it?) or I'd suggest we remove that section altogether and just assume users are using a full HEASOFT installation or a full CIAO installation.

This is not stuff we should necessarily fix with this PR, but if not with this PR we should fix some of the issues soon anyway.

olaurino:
By the way, it might make sense to rearrange the RTD installation page so that the easiest part (`conda install`, `pip install`, `python setup.py install`) is at the top of the page, as in the README. Some people will land on the README, some will land on the RTD page, we should make sure the basic info is highlighted in both places. Details might change, but the simplest forms of installation shouldn't.

README.md
DougBurke: Can it not be both ;-)

Typo.

DougBurke:
For the `--enable-xs-models-only` part - perhaps leave in but say this is currently unsupported and we recommend using the "full" install?

DougBurke:
Also note: I don't have time to make any changes as I'm out until the New Year after today. I just wanted to make sure we had this PR to track ;-)

olaurino:
So would you rather keep this open while we review the RTD instructions? That would be fine with me.

DougBurke:
I was just saying that I won't be making any changes to this PR, so please feel free to make any additions you feel are needed. Maybe review with @anetasie ?

DougBurke:
@olaurino - I've made a few changes to address some of your points for the RTD documentation. Not sure if it's all of them.

Issue #561, opened 2018-11-30
Sherpa in a Jupyter notebook: show_model does not display anything


hamogu:
In a jupyter notebook after loading any data and setting any source, I try:
```
from sherpa.astro import ui
ui.show_model()
```
However, the following works:
```
import sys
ui.show_model(outfile=sys.stdout)
```
and there is no output. The cause if this clearly is that the `show_` commands pass the result to a pager and that does not work inside a jupyter notebook.

I suggest the following:

Detect when we are in a notebook and just redirect to stdout if `outfile=None`. If the output is too long, the notebook can hide it, fold it, or display it all. No need for Sherpa to do anything.

That should be a fairly easy fix, but given the timing of the upcoming release with CIAO 4.11, this might just be a documentation issue (@DougBurke) for now.

DougBurke:
This is essentially #445 but with a more-obvious subject line.

There's no chance for getting this in to CIAO 4.11, unfortunately.

Issue #445, opened 2018-04-20
modernize pager code


olaurino:
While testing the 4.10.0 conda binaries I ran into an issue with the utility function that displays information from sherpa in a pager:
https://github.com/sherpa/sherpa/blob/8e96fb6b86af1e3730724213baa7fa7306e31e16/sherpa/ui/utils.py#L98

There is a lot of hard-coded logic that can be replaced with something more robust. I quickly gave a look around and found this:
https://stackoverflow.com/a/37585086

olaurino:
Let me add that I don't think the issue is likely to happen at all in regular deployments... I am testing in barebone docker containers that do not have `more` installed and do not have `PAGER` set. Regular users usually have both. I worked around this in my tests by setting `PAGER` to `less`, which is present in the docker image I am using.

DougBurke:
I was just having a discussion with a local user today who said that when they used a function like `sherpa.astro.ui.show_model` in a Jupyter notebook then the output when to the terminal running the server, and not the notebook. I suggested passing in a `StringIO` object via the `outfile` parameter and then displaying the contents, but it's not ideal.

kglotfelty:
Adding to Doug's notebook example,   a CIAO helpdesk user [ref 21322] is running sherpa via [spyder](https://github.com/spyder-ide/spyder) and ran into the same problem that `show_*()` routines page the results into the terminal where the GUI was launched (or maybe into some system log if the GUI was launched from an Applications window)

olaurino:
should we make this a higher priority?

DougBurke:
I am okay with increasing the priority, but  I haven't looked at your road map to see how this fits in. Note that there is a chunk of work we could do to make Sherpa work "better" within notebooks (e.g. #345 #347 but I haven't done an exhaustive check to see if there are more), but not for Sherpa 4.11 unless we get external contributors working on it.

I don't think it should a big job code-wise, but maybe more work on the policy/API-design side (i.e. what we want to happen and how to control things).

DougBurke:
I've just tried out the `pydoc.pager` suggestion @olaurino pointed to and it seems to work in both a notebook (where there was no paging) and from IPython. So I think it's a worthwhile change.

PR #583, opened 2019-01-14
WIP update pager code #445 #561

DougBurke:


olaurino:
It looks good. Let me see if I can find a meaningful way of testing the fix, and yes it would be good to add some unit tests for that functionality.

sherpa/astro/ui/utils.py
```diff
@@ -26,7 +27,6 @@
 import numpy
 
 import sherpa.ui.utils
-from sherpa.astro.data import DataARF
 from sherpa.astro.instrument import create_arf, create_delta_rmf, create_non_delta_rmf
 from sherpa.ui.utils import _argument_type_error, _check_type, _send_to_pager
```
olaurino: Should we "fix" the access to functions that are implied to not be publicly accessible and rename those functions?

DougBurke: Sorry - I dont' understand what you mean. Can you give an example?

olaurino: the module imports `_send_to_pager` but the `_` in front of the name suggests this should not be accessed and used outside of the module where it's defined. Same for other functions. It's a minor thing, but since we are there...

olaurino:
The usefulness of unit tests in this case is minor, and I have plenty of things to do with a higher priority. We did some manual testing, which is more relevant. So I'll just merge the PR.

I am bothered by the names of the functions, but I don't want that to stop this to be merged. It's not an issue with this PR.

sherpa/astro/ui/utils.py
olaurino: > We don't have an articulated strategy for naming routines/ I tend to think of `_xxx` routines as ones that "internal" code may use, from other modules even, but shouldn't really be called by the user, or are part of any API or semantic stability we try to keep for "user" routines. I think we would want to come up with some sort of standard here before making large-scale changes.

If we don't have a standard of our own, the language standards should apply. `_*` names should be considered as "private", i.e. implementation details subject to change and not to be used outside of the scope where they are defined.

They should not be imported in other modules (Python prevents them from being imported with a `*` import). I think the current code is sloppy and the `_` should be removed, as clearly the functions are not meant to be private but used outside of the scope where they are defined. I also don't think it's a large scale change. An IDE worth its salt would change this quite easily, and we are talking about three functions not largely used in the code base. I won't let this stop this PR from being merged.

PR #584, opened 2019-01-18
WIP Update xspec

olaurino:
Working on creating an xspec 12.10 job, but having issues... Creating the pull request should allow me to run the job in debug mode.


olaurino:
I believe I can run the travis job in debug mode without opening the PR, which instead means that two jobs are run every time I push a change, so closing this PR for the time being.

PR #585, opened 2019-01-18
rfe plot pvalue

dtnguyen2:
# Release Note

# Description
Handle the case given in:
https://icxc.harvard.edu/pipe/sherpadev/2018/0510.html


dtnguyen2:
Apparently only I can see the following mail search:

 https://mail.google.com/mail/u/0/#search/plot_pvalue/KtbxLthZkHGvbcJQtVBTCwVlTJshlgqcBq

so I'm resending the aforementioned thread so it can be seen (I think) by everyone:

https://icxc.harvard.edu/pipe/sherpadev/2018/0462.html

Will update when more data and scripts (psf, etc...) are made available by SDS

olaurino:
Rebasing and force-pushing to solve a conflict in the submodule

olaurino:
Hmm... something is still wrong...

olaurino:
It looks like github was confused... there was no conflict to begin with.

PR #588, opened 2019-02-01
Add XSPEC 12.10.0e to Travis jobs (Linux), Drop Py2.7/XSPEC job

olaurino:
This PR introduces the following changes in the Travis build:

  * Two XSPEC 12.10.0e jobs with Python 3.6 and 3.7 are added.
  * The XSPEC/Py27 job is removed.
  * Travis build matrix now includes XSPEC 12.9.1 and 12.10.0 for Python 3.5, 3.6, and 3.7.
  * As the build configuration with 12.9.1 and 12.10.0 is drastically different, some changes were made to the configuration scripts.
  * Conda is now verbose again. At first this was a mistake, I just forgot to roll back a debugging change, but it seems to work fine. We had made conda less verbose because the jobs would hit the log file size limitations, but it does not looks like they are anymore, at least for now. On the other hand very often the quietness of conda makes the jobs time out, so hopefully that won't happen anymore either.
  * The XSPEC 12.10.0/Python 3.7 job also tests against Matplotlib 3.
  * Rather than always installing Matplotlib 2.0, the Travis jobs now correctly install the latest version of the 2 series.

Also, since a regression (thread) test started to fail with 12.10.0, I compensated by applying the changes in #534.

## What's missing

@DougBurke suggested we only test 12.10.1, but we still don't have those conda binaries.

Similary, we still don't have macOS binaries.

Having successful builds with 12.10.0 will demonstrate we are compatible with that version and document the changes in configuration users would require.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,25 +6,21 @@ matrix:
   fast_finish: true
   include:
     # macOS build
-    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=2.0 XSPECVER="12.9.1" TRAVIS_PYTHON_VERSION="3.6"
+    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=2 XSPECVER="12.9.1" TRAVIS_PYTHON_VERSION="3.6"
       os: osx
     # Barebone build to check tests pass when most of the tests must be skipped.
     # We need to use TEST=none to remove the test-data submodule.
     - env: INSTALL_TYPE=develop TEST=none NUMPYVER=1.11 TRAVIS_PYTHON_VERSION="2.7"
-    # Full build (including ds9 and xspec), Python 2.7, setup.py develop, astropy, matplotlib 1.5
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="2.7"
+    # Full build (including ds9 and xspec), Python 3.5, setup.py develop, astropy, matplotlib 1.5
+    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1 TRAVIS_PYTHON_VERSION="3.5"
       sudo: required
       dist: trusty
-    # As above, Python 3.5
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.5"
+    # As above, Python 3.6, setup.py install, xspec 12.10
+    - env: XSPECVER="12.10.0e" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2 TRAVIS_PYTHON_VERSION="3.6"
       sudo: required
       dist: trusty
-    # As above, Python 3.6, setup.py install
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
-      sudo: required
-      dist: trusty
-    # As above, xspec 12.9.1n
-    - env: XSPECVER="12.9.1" NUMPYVER="1.14" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
+    # As above, python 3.7, numpy 1.14, matplotlib 2.0
+    - env: XSPECVER="12.10.0e" NUMPYVER="1.14" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=3 TRAVIS_PYTHON_VERSION="3.7"
       sudo: required
       dist: trusty
     # Experimental support for using Sphinx to build the documentation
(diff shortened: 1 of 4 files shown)


olaurino:
I think I'll squash the commit before merging, though.

DougBurke:
Just to confirm that 2b91ad3 should read "matplotlib 1.5 cannot be installed on py3"

.travis.yml
```diff
@@ -11,20 +11,16 @@ matrix:
...
       dist: trusty
-    # As above, Python 3.6, setup.py install
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.6"
+    # As above, Python 3.6, setup.py install, xspec 12.10
+    - env: XSPECVER="12.10.0e" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
       sudo: required
       dist: trusty
-    # As above, xspec 12.9.1n
-    - env: XSPECVER="12.9.1" NUMPYVER="1.14" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2.0 TRAVIS_PYTHON_VERSION="3.6"
+    # As above, python 3.7, numpy 1.14, matplotlib 2.0
```
DougBurke: The comment should read "matplotlib 3.0" I think?

travis/setup_xspec_ds9.sh
```diff
@@ -60,30 +60,27 @@ cd -
...
-  12.9.10e)  # I misspelled the version string, should have been 12.10.0
+  12.10.0e)  
       xspec_version_string="12.10.0"
+      xspec_include_path="$miniconda/envs/build/include"
+      xspec_library_path="$miniconda/envs/build/lib"
       ;;
   *)
-      xspec_version_string="12.9.0"
+      xspec_version_string="12.9.1"
+      sed -i.orig "s/#cfitsio_libraries/cfitsio_libraries/g" setup.cfg
```
DougBurke: So, just to confirm that the `*_libraries` settings only need changing in out XSPEC 12.9.1 build?

DougBurke:
I am also confused by the section

```
It looks like matplotlib 2 cannot be installed with python 3, so I updated the jobs configuration accordingly (matplotlib 3 w/ Python 3.7 and matplotlib 2 w/ Python 3.6). This also means that we are now introducing matplotlib 3 tests.
```

Do you mean that matplotlib 2 can't be installed with python 3.7 not python 3?

olaurino:
> Just to confirm that [2b91ad3](https://github.com/sherpa/sherpa/commit/2b91ad3c63917728b3e4b1b6e477197b0466cb08) should read "matplotlib 1.5 cannot be installed on py3"

There was a conflict between matplotlib 2.0 and python 3.7:
https://travis-ci.org/olaurino/sherpa/jobs/487494366#L536

However, I think there was (and still is) an issue in the travis configuration, as we explicitly set 2.0 (now also 3.0), which is incorrect. We should just use the major version, i.e `2` or `3`. I can update the configuration so that only the major version is used, but I can also keep the job with matplotlib 3, since we are there, right?

travis/setup_xspec_ds9.sh
olaurino: Yes, but that's only true for the xspec binaries we (standalone sherpa and ciao) compile. With a default heasarc installation you'd still need to link the libraries heasarc didn't link explicitly, so you'd need to set `*_libraries` and cross your fingers.

DougBurke:
@olaurino - I was confused because the text says you can't use matplotlib 2 with python 3 and then says "with python 3.6 we use matplotlib 2". I am guessing that the text should have read "can not use matplotlib with python 3.7", and then I wouldn't have been confused.

I don't understand your point about the travis configuration - do you mean that we should change things to just look for "2" or "3" and not "2.0" or "3.0" (so the result will be the same but things are a little cleaner), or is it an actual error that we are using "2.0"/"3.0" (ie some tests may be using the wrong version)?

olaurino:
> I am guessing that the text should have read "can not use matplotlib with python 3.7", and then I wouldn't have been confused.

Yes. Sorry for the confusion.



> or is it an actual error that we are using "2.0"/"3.0" (ie some tests may be using the wrong version)?

This. By setting the variables to "2.0" and "3.0" we were forcing the installation of something in the "*.0" series, rather than e.g. "2.2.3". I just checked that if we leave conda free to install the latest release in the "2" series the conflict is gone.

So yes, when matplotlib 2.1 was released we kept testing matplotlib 2.0

DougBurke:
I think the matplotlib 2 / python 3.7 issue must be down to the conda packaging of matplotlib, since https://matplotlib.org/users/installing.html (which says it's for version 3.0.2) says 'python >= 3.5'. I don't think we need to worry about this, just make sure we have tests that run ;-)

travis/setup_xspec_ds9.sh
DougBurke: Yeah - I was just clarifying that the change was intentional for our setup only ;-)

DougBurke:
The last point in the release notes says:

Rather than always installing Matplotlib 2.0, the Travis jobs now correctly install the latest version of the 2 series.

However, I don't see that in the current commits - ie as of 2b91ad3. Is this a not-yet-commited change (which you talk about in https://github.com/sherpa/sherpa/pull/588#issuecomment-459894971 ) or am I missing something?

olaurino:
> Is this a not-yet-commited change

This.

DougBurke:
Okay - I'll look at it again on Monday (or whenever the changes/squash take place). Overall I'm happy with this but just want to see the final version before pressing accept.

olaurino:
Just pushed the squashed commit which should also address @DougBurke's comments.

We now have 12.10.0 and 12.10.1 xspec conda packages for both linux and macos (thanks to Marie) and hopefully we will soon have updated travis jobs.

.travis.yml
```diff
@@ -6,25 +6,21 @@ matrix:
...
-    # Full build (including ds9 and xspec), Python 2.7, setup.py develop, astropy, matplotlib 1.5
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="2.7"
+    # Full build (including ds9 and xspec), Python 3.5, setup.py develop, astropy, matplotlib 1.5
+    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1 TRAVIS_PYTHON_VERSION="3.5"
       sudo: required
       dist: trusty
-    # As above, Python 3.5
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1.5 TRAVIS_PYTHON_VERSION="3.5"
+    # As above, Python 3.6, setup.py install, xspec 12.10
+    - env: XSPECVER="12.10.0e" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2 TRAVIS_PYTHON_VERSION="3.6"
```
DougBurke: I've noticed that the `INSTALL_TYPE` here has been changed from `develop` to `install`. I don't have a problem with this change but I just wanted to check it was intended.

olaurino: That's fine. It's a way to explore more of the infinite number of possible combinations in which Sherpa can be built.

.travis.yml
DougBurke: I think this is still outstanding, but I don't think it's an important enough change to force delaying merging this PR (given we want this in for #589) - as it is only a change in the comment and the comment can be fixed in another PR (which we're going to have to do when we add XSPEC 12.10.1).

DougBurke:
LGTM (note that the "matplotlib 2.0" comment can be fixed in a separate commit) as we want to get this in

PR #589, opened 2019-02-05
Update macOS Travis job to use XSPEC 12.10.0

olaurino:
This PR updates the macOS Travis job to use XSPEC 12.10.0.

We will soon update it again to use 12.10.1, but I'd like to make sure there is trace of a successful 12.10.0 build. Also, we are changing the channel from which we download xspec, so I'd like to make sure that works as well before proceeding further. I branched off #588 and I'll need to rebase after that one is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   fast_finish: true
   include:
     # macOS build
-    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=3 XSPECVER="12.10.0e" TRAVIS_PYTHON_VERSION="3.7"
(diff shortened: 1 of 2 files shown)


olaurino:
The xspec tests on macOS seemed to be OK, but there was a segfault-like error on `test_plot`. To make sure it's related to matplotlib and doesn't have anything to do with xspec I downgraded matplotlib to v2. Let's see what happens.

DougBurke:
I didn't think `test_plot` did enough to cause any sort of error ...

olaurino:
The job with matplotlib is succeeding. We should investigate why matplotlib 3 makes the macos build fail. I'll set up a branch on my fork to confirm the failure.

DougBurke:
LGTM :shipit:

PR #593, opened 2019-02-07
Release 4.11.0

olaurino:


DougBurke:
Note that 798f62d changes a `.py` file but it is only changes to the docstring. This is added to note a change in behavior for users using XSPEC 12.10.1, since I think the change in the `xsect` settings from `bcmc` to `vern` is going to surprise some users (as it is essentially a quiet change), but I'm not sure how best to document this (probably needs a note added to the #534 item in the release notes).

olaurino:
Not a big deal in this case, considering the scope of the change, but just as reference for the future... the XSPEC docstring change was better suited as its own PR, because it could go directly into `master` (and then cherry-picked here) and it's not directly related to the preparation of the release (bumping version numbers, tweaking installation docs as we test the release, and so on).

Again, this is just to clarify what kind of changes I think should be pushed directly to the release branch and which ones should have their own PR, merged, and then cherry-picked. Not at all a big deal in this specific case.

DougBurke:
I have no idea of the semantics you intend for the release branches, so
thanks for the clarification.

PR #596, opened 2019-02-12
Run optimization algorithms over multiple cores

dtnguyen2:
# Release Note
This PR enables the user to run the optimization algorithms (DifEvo, LevMar and NelderMead) on multi-cores.

# Description
The default setting for this PR is ```numcores=1```; However the user/tester can modify at runtime by either using the ui ```set_method_opt('numcores', N)``` through the config setting of the optimization class for example: ```opt=LevMar(); opt.config['numcores'] =N``` where N is any integer greater then one to indicate the numbers of cores to use.  The user/tester should keep in mind that the overhead in optimizing using multi-cores is rather high and the limited tests have shown that they are only useful if the model evaluation is very time consuming.

# Notes
This PR addresses one element of #442, in particular:
```
2c/ Parallelization for MonCar and Nelder-Mead methods:
      - lower priority until 2a/2b is done
```

Just checked in an implementation of the Differential Evolution (difevo) algorithm to run on multi-cores.  To use it set the option numcores to any number of cores that one wants to use, if not sure how many cores your workstation has then set numcores to 'all'.  The default setting for numcores is 1, so the old C++ implementation will be used. Using the following script:

(sherpa3) [dtn@devel12 numcores]$ cat  cmp.py 
import time
from sherpa.optmethods import optfcts
from sherpa.optmethods import _tstoptfct

def tst(npar):
    x0, xmin, xmax, fmin = _tstoptfct.init('rosenbrock', npar)
    start = time.time()
    result = optfcts.montecarlo(_tstoptfct.rosenbrock, x0, xmin, xmax,
              numcores='all')
    end = time.time()
    print(result)
    print(end -start, 'secs')

    start = time.time()    
    result = optfcts.montecarlo(_tstoptfct.rosenbrock, x0, xmin, xmax,
                                numcores=1)
    end = time.time()    
    print(result)
    print(end -start, 'secs')

    print()
    
    x0, xmin, xmax, fmin = _tstoptfct.init('freudenstein_roth', npar)
    start = time.time()        
    result = optfcts.montecarlo(_tstoptfct.freudenstein_roth, x0, xmin, xmax,
              numcores='all')
    end = time.time()        
    print(result)
    print(end -start, 'secs')

    start = time.time()            
    result = optfcts.montecarlo(_tstoptfct.freudenstein_roth, x0, xmin, xmax,
                                numcores=1)
    end = time.time()            
    print(result)
    print(end -start, 'secs')

The script outputs the results of using the 'moncar' optimization method when numcores is set to 'all' and 1 for the Rosenbrock and Freudenstein and Roth functions. I get the following output:

(sherpa3) [dtn@devel12 numcores]$ python cmp.py 
(True, array([0.99999415, 0.99998881, 1.00000246, 1.00000495, 1.00000122,
       1.00000248]), 6.920901964836967e-11, 'successful termination', {'info': True, 'nfev': 8576})
1.292532205581665 secs
(True, array([0.99999597, 0.99999191, 1.00000595, 1.00001183, 0.99999823,
       0.99999655]), 5.5866566735675604e-11, 'successful termination', {'info': True, 'nfev': 11546})
0.058183908462524414 secs

(True, array([5., 4., 5., 4., 5., 4.]), 4.301090351841087e-17, 'successful termination', {'info': True, 'nfev': 7415})
1.2173118591308594 secs
(True, array([11.41277898, -0.89680525, 11.41277898, -0.89680525, 11.41277898,
       -0.89680526]), 146.95276103772008, 'successful termination', {'info': True, 'nfev': 10895})
0.04735088348388672 secs

As you can see, it takes longer to run on the new implementation using multiple cores (20 on my workstation) then the old implementation but that is expected since the test function evaluations are fast but there is a bit of an overhead for using parallel_map.  The results are not identical because the new implementation is a new and improved algorithm.  As you can see: for the Freudensteing and Roth function when numcores=all vs numcores=1 moncar gets the global vs local minimum, respectively.

Quite frankly, I'm not sure how much faster the new implementation of difevo, but I know it does not suffer from the dreaded 'the curse of dimensionality' as much as the old implementation.  When running the test script with 30 parameters: the new implementation gets the minimum for the Rosenbrock ~5e-8 vs .2 for the old;  Moreover, for the minimum for the Freudenstein and Roth function is 3.08e-16 vs 734.76 when numcores=all vs numcores=1, respectively:

(sherpa3) [dtn@devel12 numcores]$ python cmp.py 30
(True, array([1.00008151, 1.00016187, 1.00004667, 1.00009251, 0.99995188,
       0.99990393, 1.00002637, 1.00005166, 0.99999034, 0.99997893,
       0.99998172, 0.99996206, 1.00001583, 1.00003125, 1.00000188,
       1.00000255, 0.99999647, 0.99999202, 1.00006654, 1.0001321 ,
       1.00004481, 1.00008701, 1.00004369, 1.00008586, 0.99988397,
       0.99976517, 0.99998959, 0.99997734, 0.99989207, 0.99978008]), 5.094045656887898e-08, 'successful termination', {'info': True, 'nfev': 47144})
2.205500841140747 secs
(True, array([1.38855448, 1.92919219, 0.95921491, 0.91486137, 1.11027076,
       1.23449034, 1.1115276 , 1.23590618, 0.94271749, 0.88906997,
       1.13636067, 1.29229779, 1.0226386 , 1.04549794, 1.03787775,
       1.07680124, 1.07804691, 1.1624894 , 0.96957805, 0.9402225 ,
       1.02484322, 1.04886194, 0.89801919, 0.80630727, 0.95060946,
       0.90202392, 0.93390711, 0.87329113, 1.12613221, 1.26883255]), 0.24579409708402042, 'successful termination', {'info': True, 'nfev': 43627})
1.701136589050293 secs

(True, array([5., 4., 5., 4., 5., 4., 5., 4., 5., 4., 5., 4., 5., 4., 5., 4., 5.,
       4., 5., 4., 5., 4., 5., 4., 5., 4., 5., 4., 5., 4.]), 3.086265573690332e-16, 'successful termination', {'info': True, 'nfev': 59277})
2.3580703735351562 secs
(True, array([11.412779  , -0.89680525, 11.41277899, -0.89680525, 11.41277902,
       -0.89680525, 11.41277908, -0.89680525, 11.41277893, -0.89680526,
       11.41277894, -0.89680526, 11.41277903, -0.89680525, 11.412779  ,
       -0.89680525, 11.41277894, -0.89680526, 11.41277899, -0.89680525,
       11.41277903, -0.89680525, 11.41277908, -0.89680525, 11.41277906,
       -0.89680525, 11.41277906, -0.89680525, 11.41277899, -0.89680525]), 734.7638051886005, 'successful termination', {'info': True, 'nfev': 58578})
0.9694828987121582 secs


olaurino:
@dtnguyen2, the PR is missing a Release Note.

It looks like a bunch of unintended/unrelated changes (and entire files) were included in the commits by mistake.

Also, a bunch of tests in this PR fail, a couple of which fail because a test file was removed, most likely by mistake.

dtnguyen2:
See 2c in following doc:

https://docs.google.com/document/d/1wQmHUEviv2K4U5rSK3EH1hYO0ByAVY3SpbgkHG7T7RU/edit#

anetasie:
I just run a simple image fitting test with `moncar` using a relatively large image (253,253) with no psf convolution. The execution time is 49 sec vs 100 sec when I use 3 cores instead of 1. I can definitely see the speed improved and get the same best fit values. This would be a good improvements for our image fitting.

anetasie:
I noticed that the printout on the screen shows the number of function evaluation which is much smaller when I used 3 cores. I suspect that the printout to the screen shows the number of evaluations for one core, instead of 3, but the division of evaluations (3*3218=9654) does not match perfectly.

```
Dataset               = 1
Method                = moncar
Statistic             = cash
Initial fit statistic = -290594
Final fit statistic   = -515548 at function evaluation 8726
Data points           = 63756
Degrees of freedom    = 63751
Change in statistic   = 224954
   c2.c0          0.0680481   
   g1.fwhm        7.97214     
   g1.xpos        127.387     
   g1.ypos        127.061     
   g1.ampl        807.438     



Dataset               = 1
Method                = moncar
Statistic             = cash
Initial fit statistic = -290594
Final fit statistic   = -515548 at function evaluation 3218
Data points           = 63756
Degrees of freedom    = 63751
Change in statistic   = 224954
   c2.c0          0.0680001   
   g1.fwhm        7.97257     
   g1.xpos        127.387     
   g1.ypos        127.062     
   g1.ampl        807.188    

```

olaurino:
Rebasing and force-pushing to resolve a (minor) conflict and get the correct sherpa-test-data commit.

olaurino:
So all our tests are failing because anaconda changed the structure of their repositories (see https://www.anaconda.com/how-we-made-conda-faster-4-7/). I tested a solution locally but that will require rebasing a whole lot of PRs (or the PRs could cherry-pick the relevant commit, either way it's a lot of work).

olaurino:
Rebased once more to get #636, hoping CI will finally run.

olaurino:
Because of several issues, starting with the fact that the submodule's commits hadn't been pushed, the CI tests on this PR never executed. Now it looks like they are failing for Python 2.7:
https://travis-ci.org/sherpa/sherpa/jobs/551858697

I can reproduce the error on my workstation with Python 2.7 and this PR.

olaurino:
Oh, and while the other tests haven't completed yet it looks like there are some other failures with Python 3 as well.

olaurino:
There are still three errors in Travis:
  1. It looks like matplotlib 3 is issuing a new warning. I am inclined to open a new ticket to track it, but to whitelist it for the time being (https://travis-ci.org/sherpa/sherpa/jobs/557964942#L5326)
  2. Some tests for this PR are erroring out on macOS. Given the nature of the tests I think I'd skip them on macOS. As long as they run on Linux and pass, we are fine. The tests depend on the number of CPUs on the virtual machine (https://travis-ci.org/sherpa/sherpa/jobs/557964938#L3820)
  3. 1 test is failing with xspec 12.9.1. The test passes in all the other jobs. I am not sure what to make of that. A few days ago we almost ditched the 12.9.1 job. If we had, we wouldn't had seen this failure. I'd be inclined to skip the test (https://travis-ci.org/sherpa/sherpa/jobs/557964940)

@DougBurke thoughts? Especially on 3.

DougBurke:
So, running things with some-level of multi-core code could explain XSPEC differences (if there are memory-read/write issues in the 12.9.1 code base, say). So I think we should just drop the 12.9.1 XSPEC tests.

For point 2, I am generally a bit wary of "oh, it's okay on linux, must be fine on macOS" style arguments, but I don't have the time to look into this particular case to see whether I'm being too picky.

For the matplotlib warning: I'd really prefer it if our "skip-this-warning" check were more targeted - i.e. we would mark up the particular tests with the warnings it is known to generate. I am just concerned that our current scheme means that we don't notice if a piece of code suddenly starts generating warnings that we've seen before. However, that is a larger issue than this PR.

I note we are still seeing the occasional

```
  /home/travis/miniconda/envs/build/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: Unable to load the ciao_version module to determine version number- defaulting 'group' version to 0.0.0
    return f(*args, **kwds)
```

warning.

olaurino:
> For point 2, I am generally a bit wary of "oh, it's okay on linux, must be fine on macOS" style arguments, but I don't have the time to look into this particular case to see whether I'm being too picky.

This is not the case here. Note that jobs on Travis won't generally exercise parallelization because they run on docker containers and virtual machine that usually have only one core. Unless I am missing something (@dtnguyen2 please read this) the tests are just checking that the multicore code is running, the actual results are being tested in other parts of the test suite, and we did have some failures we had to address.

> For the matplotlib warning: I'd really prefer it if our "skip-this-warning" check were more targeted - i.e. we would mark up the particular tests with the warnings it is known to generate. I am just concerned that our current scheme means that we don't notice if a piece of code suddenly starts generating warnings that we've seen before. However, that is a larger issue than this PR.

Agreed. I wish we had more time for this kind of improvements, but all in all I think the current infrastructure is doing a good job at telling us when new warnings appear, as in this case. This is something we will have to fix before matplotlib 3.3 otherwise the warning will turn into an error.

dtnguyen2:
When I originally checked the code in, @olaurino made a comment to me that if ```numcores != 1``` then how do you know that optimization routine was run on multi-cores?  So I added a check to verify that multi-cores were used.  The test for the fit results on different platforms are still being checked and they pass for all platforms. 

As for point# 2: I made the following changes to the code:
```
        if _ncpus != 1:
            assert fit_results.extra_output['num_parallel_map'] > 0
        else:
            assert fit_results.extra_output['num_parallel_map'] == 0
  ```
point #2 should no longer be an issue

PR #598, opened 2019-02-15
4.11.1 Upgrade Embedded Dependencies

olaurino:
# Release Note
Sherpa now ships and build with new versions of the following dependencies:
  * wcssubs: 3.8.7 -> 3.9.5
  * fftw: 3.3.4 -> 3.3.8
  * CIAO region, stk: 4.9 -> 4.11

# Notes
The group library was unchanged. Upgrading the libraries is not time consuming, but it does require some manual intervention, especially for the CIAO libraries. I added a document in confluence to show what I did and hopefully how to do it in the future.


extern/config.sub
```diff
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Configuration validation subroutine script.
-#   Copyright 1992-2017 Free Software Foundation, Inc.
+#   Copyright 1992-2013 Free Software Foundation, Inc.
```
DougBurke: Was the new code generated with an older version of autotools (or whatever it is) than the old code (ie the drop in max release date from 2017 to 2013)? Does this have any impact (since newer is surely better and shinier)?

DougBurke:
I think it would be useful to note in the changelog the update in the FFTW version that accompanies this change. Oh, I see it came in a second commit (from SH-90) so it probably just hasn't been added to the PR  summary yet.

extern/config.sub
olaurino: Good catch... I thought I was using a more recent version. I will make sure I run autoreconf one more time on the most recent tools I can when everything else checks out. Thanks.

olaurino: So this is weird. I am using autoreconf v2.69, which is the latest version, and the same I'd find on ubuntu 18.10. As I said I had checked before starting, just assumed I made a mistake somewhere. So not really sure where that is coming from. Considering I am working on CentOS there is a chance it's not the "original" 2.69 but a patched repackaged version. I'll see if I can make sense of the discrepancy.

olaurino: So, autoconf hasn't been release since 2012. I guess some other autotools were older in CentOS, so after making all the changes I just reran `autoreconf -fi` from an Ubuntu 18.10 container.

.gitignore
```diff
@@ -99,16 +103,16 @@ MANIFEST
...
+extern/region-4.11/src/regparser.c
+extern/region-4.11/src/regparser.h
+extern/region-4.11/src/regparser.tab.h
+extern/region-4.11/src/regparser.tab.c
+extern/region-4.11/src/reglexer.c.tmp
 
 # Sphinx documentation
 docs/_build
-**/api
+docs/**/api
```
olaurino: @DougBurke I had to make this change because the clause would shadow some fftw code. I'd like to make sure this is fine with you before merging. I am not sure what directories it was meant to ignore.

olaurino: I tried the sphinx build, and I believe my change is ignoring the correct folders, i.e. after a build:

````
[2] $ git status --ignored
On branch upgrade-deps
Your branch is up to date with 'myfork/upgrade-deps'.

Ignored files:
  (use "git add -f <file>..." to include in what will be committed)

        __pycache__/
        build/
        docs/data/api/
        docs/fit/api/
        docs/mcmc/api/
        docs/model_classes/api/
        docs/optimisers/api/
        docs/overview/api/
        docs/plots/api/
        docs/statistics/api/
        docs/ui/api/

````

DougBurke: Sorry for dropping the ball on this. This change looks good (ie it was meant to ignore any of the api/ directories within docs/).

PR #602, opened 2019-02-25
Reduce integration tol

dtnguyen2:
opt routines have tol ~ 1.0e-7 so it doesn't make sense use ~1.0e-16 for tolerance for integration routines so the following change was made:

std::numeric_limits< double >::epsilon()  --> std::numeric_limits< float >::epsilon()


DougBurke:
This looks like it needs an update made to the test-data repository before the tests will run here, and possibly a clean up of the commit history.

PR #603, opened 2019-02-26
Bf optimization flags

zkaufman:
Changes to the region and grplib configure and Makefile.in files so they respect --with-opt flag passed to top level ciao configure script. The same changes are required for libdev/caldb4 (not in sherpa).


DougBurke:
@olaurino is it okay if I leave this with you to review, as other than pointing out the current PR fails (and that I'm not sure we want the `configure[.in].ciao` files added to the repository) I'm not sure I can be much help here!

olaurino:
@DougBurke this was open here because of a misunderstanding and it's about some internal work we are doing in CIAO, which might eventually bubble up in Sherpa, but this PR is not intended for being merged.

PR #606, opened 2019-03-13
stop using numpy.typeNA which is now deprecated

olaurino:
# Release Note
A `numpy` deprecation warning was fixed by removing the usage of `typeNA`, which was not documented and will be removed in a future release of `numpy`. The `print_fields` function has been changed to include a default mapping using the current `typeNA` implementation.

# Notes
First of all, `print_fields` is one of the functions mentioned in #87.

The fix solves the issue described in #605.

This PR changes the logic in `print_fields` because it looks like the `converters` argument is never used in Sherpa, so I simplified the function by removing the logic associated with `converters` altogether. I used the current implementation of `typeNA` to generate a default dictionary of `numpy` array types and their string representation. Anything that is not a `numpy` array is printed as its `str` representation.

There is a small unit test in `test_utils`, which passes with this change. Some tests would fail with a buggy fix, so hopefully the tests are covering enough of the functionality to prevent regressions.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -46,6 +46,20 @@ except ImportError:
     has_xspec = False
 
 
+# In some instances, if some xvfb processes did not stop cleanly
+# pytest-xfvb starts complaining that a virtual screen is already on
+# the following code works around that. The issue is hard to reproduce
+# but I (OL) tested it locally as it reappeared on my workstation.
+try:
+    import random
+    from pyvirtualdisplay import abstractdisplay
+    abstractdisplay.RANDOMIZE_DISPLAY_NR = True
+    abstractdisplay.random = random
+    random.seed()
+except ImportError:
+    pass
+
+
 TEST_DATA_OPTION = "--test-data"
 
 
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -117,7 +117,9 @@ if sys.version_info >= (3, 2):
             [r"invalid value encountered in sqrt",
              # See https://github.com/ContinuumIO/anaconda-issues/issues/6678
              r"numpy.dtype size changed, may indicate binary " +
-             r"incompatibility. Expected 96, got 88"
+             r"incompatibility. Expected 96, got 88",
+             # See https://github.com/numpy/numpy/pull/432
+             r"numpy.ufunc size changed"
              ],
     }
     known_warnings.update(python3_warnings)
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1330,12 +1330,16 @@ def print_fields(names, vals, converters=None):
 
     """
 
+    # This is the part of the deprecated typeNA dictionary Sherpa
+    # would use up to v4.11.0. We included the dictionaty verbatim,
+    # excluding the complex mapping which where wrong in typeNA.
+    # Note only the class -> string mappings have been copied over.
     if converters is None:
         converters = {numpy.bool_: 'Bool',
                       numpy.bytes_: 'Bytes0',
-                      numpy.complex128: 'Complex64',
-                      numpy.complex256: 'Complex128',
-                      numpy.complex64: 'Complex32',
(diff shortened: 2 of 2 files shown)


olaurino:
I added a piggy back commit to fix an issue I have every now and then with pytest, without having to reboot my system just for that when it happens (or having to live with ds9 popping up in my window and taking the mouse focus when I run the tests).

DougBurke:
Is there a reason you hard-coded the "converter" array rather than using `np.sctypeDict` (which I got from https://www.numpy.org/devdocs/release.html#numpy-1-16-0-release-notes )? Although I do seem to see different behavior from them so maybe that's a reason ?

```
In [15]: np.typeNA['u8']
Out[15]: numpy.uint64

In [16]: np.sctypeDict['u8']
Out[16]: numpy.uint64

In [17]: np.typeNA['c8']
Out[17]: 'Complex32'

In [18]: np.sctypeDict['c8']
Out[18]: numpy.complex64
```

olaurino:
Yes, that's the reason. They are not the same dictionary, sctypeDict actually lacks the mapping the function was looking for, and since they claim that typeNA was "buggy" I am just assuming that the original idea of the function was to make the mapping that I ended up hard coding.

Try for instance `np.typeNA[np.float64]` vs `np.sctypeDict[np.float64]` which is an actual call the function in Sherpa would make.

I might be still missing something, that's why I issued two distinct PRs.

olaurino:
btw I hadn't looked at the non-float mappings, so I hadn't noticed that the complex mappings are wrong in typeNA (I guess I understand what they mean by "it's buggy", now). We probably don't need all those mappings, but I can fix them and leave them there just in case.

sherpa/conftest.py
```diff
@@ -117,7 +117,9 @@ def pytest_addoption(parser):
             [r"invalid value encountered in sqrt",
              # See https://github.com/ContinuumIO/anaconda-issues/issues/6678
              r"numpy.dtype size changed, may indicate binary " +
-             r"incompatibility. Expected 96, got 88"
+             r"incompatibility. Expected 96, got 88",
+             # See https://github.com/numpy/numpy/pull/432
+             r"numpy.ufunc size changed"
```
DougBurke: So the ufunc change is *only* happening because we have a mis-match of NumPy versions on Travis? Why is that: are we building against NumPy version 1.x but some some other package we use has a (probably implicit) dependency on NumPy version 1.y (where y != x)? Do we know what package? I am wondering if we can set up travis so that we ensure a constant NumPy version (so hopefully not have to bother with white-listing this version), but maybe we don't want to go to that effort?

olaurino: As it is good practice, we build against an old version of numpy (1.11 seems to be a common choice) and then we test against a number of them. The warning is issued by Cython (See https://github.com/numpy/numpy/pull/432).

Most of our Travis builds have numpy pinned, while some have it float, so it's driven by other packages (astropy, in particular) and will be installed at its latest version. This "lookahead" jobs let any issues with the latest version of numpy surface as soon as possible, as it has happened in the past (e.g. when numpy changed the way they print arrays).

DougBurke: Thanks for the clarification.

DougBurke:
This LGTM - are you planning to merge soon or do more work on it @olaurino ?

olaurino:
Do you have an opinion on the mappings for complex types?
https://github.com/sherpa/sherpa/pull/606/files#diff-0f8e355a5c30baec0c8a9a1b38926a1eR1336

Should I leave them alone? should I "fix" them? should I remove them altogether?

DougBurke:
I don't have a good handle on where user's see this information. My bet would be we don't actually ever call it for ndarray's containing complex numbers, so it probably doesn't matter, but I can't guarantee that.

olaurino:
I added an explanatory comment and fixed the complex names. I am going to assume @DougBurke is still fine with the changes and I am going to merge them when Travis is green.

PR #612, opened 2019-04-01
Fix #609: User Model: unable to change parameter values

### LINKED ISSUE #609: UserModel: unable to change parameter values
I originally thought this was a case-sensitivity, but it is more serious than that. In the following (using CIAO 4.11 but will hold for Sherpa 4.11.0 as well)

a) I can apparently change the parameter value "X", but it doesn't actually change the underlying model (i.e. print(silly) still says X is 2 even after I changed it to 4)

b) I can't use a lower case version of the parameter name to access the parameter

```
% python
Python 3.5.4 (default, Sep 14 2018, 15:42:52) 
[GCC 4.8.4] on linux
>>> import sherpa
>>> sherpa._version.get_versions()
{'version': 'ciao4.11', 'full': '29aca616db21356f16b59fab0de2a2e5590fdca7'}
>>> def sillymdl(pars, xlo, *args):
...     return pars[0] * xlo
... 
>>> ui.load_user_model(sillymdl, 'silly')
>>> ui.add_user_pars('silly', ['X'], parvals=[2.0])
>>> print(silly)
usermodel.silly
   silly.X      thawed            2 -3.40282e+38  3.40282e+38           
>>> silly.X = 4
>>> print(silly)
usermodel.silly
   silly.X      thawed            2 -3.40282e+38  3.40282e+38           
>>> silly.x = 4
  File "/home/dburke/local/ciao-4.11/lib/python3.5/site-packages/sherpa-ciao4.11-py3.5-linux-x86_64.egg/sherpa/models/model.py", line 199, in __setattr__
    NoNewAttributesAfterInit.__setattr__(self, name, val)
  File "/home/dburke/local/ciao-4.11/lib/python3.5/site-packages/sherpa-ciao4.11-py3.5-linux-x86_64.egg/sherpa/utils/__init__.py", line 148, in __setattr__
    (type(self).__name__, name))
AttributeError: 'UserModel' object has no attribute 'x'
>>> 
```

DougBurke:
# Release Note
A regression introduced in Sherpa 4.10.0 presented users from change user-model parameter values through direct access. This issue has been fixed. Several tests were added to reduce the chance of regressions in the future.

# Summary

Fix issue #609 - the inability to change user-model parameter values through direct access - which was broken by some refactoring (probably #255 introduced in Sherpa 4.10.0).

Several tests have been added to exercise simple one-dimensional user models (both for this fix and to check other behavior).

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -396,15 +396,16 @@ def clean_astro_ui():
 
     This also resets the XSPEC settings (if XSPEC support is provided).
 
+    See Also
+    --------
+    clean_ui
+
     Notes
     -----
     It does NOT change the logging level; perhaps it should, but the
     screen output is useful for debugging at this time.
     """
     from sherpa.astro import ui
-
-    # old_lgr_level = logger.getEffectiveLevel()
-    # logger.setLevel(logging.CRITICAL)
 
     if has_xspec:
         old_xspec = xspec.get_xsstate()
@@ -418,7 +419,26 @@ def clean_astro_ui():
     if old_xspec is not None:
         xspec.set_xsstate(old_xspec)
 
-    # logger.setLevel(old_lgr_level)
+
+@pytest.fixture
+def clean_ui():
+    """Ensure sherpa.ui.clean is called before AND after the test.
+
+    See Also
+    --------
+    clean_astro_ui
+
+    Notes
+    -----
+    It does NOT change the logging level; perhaps it should, but the
+    screen output is useful for debugging at this time.
+    """
+    from sherpa import ui
+
+    ui.clean()
+    yield
+
+    ui.clean()
 
 
 @pytest.fixture
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1843,7 +1843,8 @@ class UserModel(ArithmeticModel):
             pars = (self.ampl,)
         else:
             for par in pars:
-                self.__dict__[par.name] = par
+                setattr(self, par.name, par)
+
         ArithmeticModel.__init__(self, name, pars)
 
 
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -390,7 +390,6 @@ def test_user_model_create_pars_full():
 # parameter values for a user model using direct access
 #
 @pytest.mark.usefixtures("clean_ui")
-@pytest.mark.xfail(msg="bug 609")
 def test_user_model_change_par():
 
     mname = "test_model"
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 #
-#  Copyright (C) 2012, 2015, 2016, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2012, 2015, 2016, 2018, 2019  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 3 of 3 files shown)


DougBurke:
@olaurino - I thought I'd try and add some simple tests of `load_user_model`, including one that represents the #609 failure.

olaurino:
awesome, thanks!

DougBurke:
I've got more to add (fixing a logical error that isn't exposed in the
existing commit)

DougBurke:
So I think I have a fix for this. I think the first couple of commits should be squashed together.

sherpa/models/basic.py
```diff
@@ -1843,7 +1843,8 @@ def __init__(self, name='usermodel', pars=None):
             pars = (self.ampl,)
         else:
             for par in pars:
-                self.__dict__[par.name] = par
```
olaurino: This line has not been changed since we created the initial git repository. Any idea what introduced the regression?

DougBurke: As I mentioned in the commit notes; #255 changed the logic for the __getattr__/__setattr__ behavior, adding in information to the model when a parameter was added. This code was not being called here, when editing __dict__ directly. Or at least that's my belief.

DougBurke: [I probably should have added the last sentence as a separate comment rather than when I was editing it because of markdown]

olaurino: Thanks.

For the record, I don't think git bisect requires all the tests to pass at every commit (or even the code to build at every commit) and one can add skipping conditions for compilation errors and other unrelated issues, even when using `git bisect run` to automate the search rather than running interactively. I don't think it matters in this case anyway.

olaurino:
> I think the first three commits should be squashed together before merging.

I am not sure if you were planning to do this or you expect me to.

sherpa/conftest.py
```diff
@@ -407,6 +411,32 @@ def clean_astro_ui():
...
+    clean_astro_ui
+
+    Notes
+    -----
+    It does NOT change the logging level; perhaps it should, but the
+    screen output is useful for debugging at this time.
+    """
+    from sherpa import ui
+
+    # old_lgr_level = logger.getEffectiveLevel()
```
olaurino: could we remove these comments from here (and `clean_astro_ui` while we are at it)? If this is useful, e.g. while debugging a test, we could introduce a new fixture for that.

DougBurke:
For some historical digging, to see if #255 was the culprit, the following script

```
from sherpa import ui
import pytest

def foo(pars, x, *args, **kwargs):
    return x

ui.load_user_model(foo, 'dn')
ui.add_user_pars('dn', ['aA', 'bb', 'XxX'], parvals=[10, 20, 30])

print(dn)

nfail = 0
if dn.aA.val != pytest.approx(10):
    nfail += 1

if dn.bb.val != pytest.approx(20):
    nfail += 1

if dn.XxX.val != pytest.approx(30):
    nfail += 1

dn.Aa = 2
dn.BB = 3
dn.xxx = 4

print(dn)

if dn.aA.val != pytest.approx(2):
    nfail += 1

if dn.bB.val != pytest.approx(3):
    nfail += 1

if dn.xXX.val != pytest.approx(4):
    nfail += 1

if nfail > 0:
    
    print("*** FAILED {}".format(nfail))
    import sys
    sys.exit(1)
```

does not fail with a commit before #255, namely

```
> git checkout 504afc0f29e82b0157ae33e63082742a1ae3aadb
```

 but fails with the #255 fix, namely

```
> git checkout 84fb8a728f8dde70e1aea94adbedd540ca98a4db
``` 

and the failure message is similar

```
  File "../test_um.py", line 22, in <module>
    dn.Aa = 2
  File "/home/dburke/sherpa/sherpa-master/sherpa/models/model.py", line 198, in __setattr__
    NoNewAttributesAfterInit.__setattr__(self, name, val)
  File "/home/dburke/sherpa/sherpa-master/sherpa/utils/__init__.py", line 144, in __setattr__
    (type(self).__name__, name))
AttributeError: 'UserModel' object has no attribute 'Aa'
```

sherpa/conftest.py
DougBurke: I can remove them. Are you okay with the commits being squashed as well?

olaurino: Yes.

DougBurke:
Once travis goes green I plan to push a rebased set of commits and then hopefully that's it for this PR.

DougBurke:
Once this goes green it's ready for final review.

PR #614, opened 2019-04-03
Refactor data classes (fix #563, #627, #628)

### LINKED ISSUE #563: Should we convert to ndarray when creating data objects or fix WStat (and other code) to work with lists?
Our documentation assumes that we support passing in lists or sequences to commands like `load_arrays` or objects like `Data1D`. Deep in other parts of the system - such as calculating stats - it makes it much-more-convenient if we can assume we have ndarrays. This can cause problems, such as shown below where `WStat` can not be used with lists. I think it would make sense for our data constructors (`sherpa.data.Data`-derived classes) to convert to ndarray automatically, but maybe there's a reason not to do this? Another option would be to document that we accepts lists (or rather sequences) and then we would have to audit the code (eg test coverage) to ensure necessary conversion routines are used to avoid problems like the following.

Problem case with current `master` (and CIAO 4.10, so this is nothing new):

```
from sherpa.stats import Cash, Chi2, WStat
from sherpa.fit import Fit

d = Data1D('test', [1, 2, 3], [4, 5, 6], [1, 1, 1])
mdl = Polynom1D()
mdl.c1.thaw()
mdl.c0 = 5.1

for stat in [Cash, Chi2, WStat]:
    sobj = stat()
    print("\nStat = {}".format(sobj.name))

    fit = Fit(d, mdl, stat=sobj)
    fit.calc_stat_info()
```

Running this produces the following:

```

Stat = cash

Stat = chi2

Stat = wstat
  File "../bugs/arrays.py", line 18, in <module>
    fit.calc_stat_info()
  File "/home/djburke/sherpa/sherpa-master/sherpa/fit.py", line 1197, in calc_stat_info
    stat, fvec = self._calc_stat()
  File "/home/djburke/sherpa/sherpa-master/sherpa/fit.py", line 1128, in _calc_stat
  File "/home/djburke/sherpa/sherpa-master/sherpa/stats/__init__.py", line 882, in calc_stat
    nelems.append(y.size)
AttributeError: 'list' object has no attribute 'size'
```

### LINKED ISSUE #627: No input validation for `Data2D`, `Data2DInt` etc.
It seems there is no input validation for the dataset objects, if I create them in the object-oriented way. I can create e.g. the 2D dataset with x0, x1 values that have the wrong dimensionality and will get a cryptic error message only later when I use them in a fit (see example below). I guess that traditionally those checks are done in `load_array` which then decides which type of dataset to create based on the dimensions given, at at least a few rudimentary checks in `Data2D` etc would be useful for people creating the datasets as objects themselves.

In the code below, note how I can create the dataset and a fitter based on this dataset; only when I try to use the fitter to fit something will I get an error message.
```
(saba)~% ipython
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
IPython 7.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np 
   ...: from sherpa.data import Data1D, Data1DInt, Data2D, Data2DInt 
   ...: from sherpa.stats import Chi2 
   ...: from sherpa.optmethods import LevMar 
   ...: from sherpa.estmethods import Covariance 
   ...: from sherpa.models import Gauss2D 
   ...: from sherpa.fit import Fit 
   ...:  
   ...:  
   ...: x0low, x0high = 3000, 4000 
   ...: x1low, x1high = 4000, 4800 
   ...: dx = 15 
   ...: x1, x0 = np.mgrid[x1low:x1high:dx, x0low:x0high:dx] 
   ...: y = (x0-3500)**2 + (x1 - 4500)**2 
   ...:  
   ...:  
   ...: data2 = Data2D('name', x0, x1, y.flatten(), staterror=np.sqrt(y).flatten()) 
   ...: model2 = Gauss2D() 
   ...: fitter = Fit(data2, model2, Chi2(), LevMar(), Covariance()) 
   ...:  
   ...: fitter.fit() 
   ...:                                                                                                     
<ipython-input-1-5a36b22bcd5f> in <module>
     19 fitter = Fit(data2, model2, Chi2(), LevMar(), Covariance())
     20 
---> 21 fitter.fit()

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in run(fit, *args, **kwargs)
     51     def run(fit, *args, **kwargs):
     52         fit.model.startup()
---> 53         result = func(fit, *args, **kwargs)
     54         fit.model.teardown()
     55         return result

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in fit(self, outfile, clobber)
   1228             raise FitErr('binhas0')
   1229 
-> 1230         init_stat = self.calc_stat()
   1231         # output = self.method.fit ...
   1232         output = self._iterfit.fit(self._iterfit._get_callback(outfile, clobber),

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in calc_stat(self)
   1113         """
   1114 
-> 1115         return self._calc_stat()[0]
   1116 
   1117     def calc_chisqr(self):

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in _calc_stat(self)
   1094         # TODO: is there anything missing here that
   1095         #       self._iterfit.get_extra_args calculates?
-> 1096         return self.stat.calc_stat(self.data, self.model)
   1097 
   1098     def calc_stat(self):

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/stats/__init__.py in calc_stat(self, data, model)
    542 
    543     def calc_stat(self, data, model):
--> 544         fitdata, modeldata = self._get_fit_model_data(data, model)
    545 
    546         return self._calc(fitdata[0], modeldata,

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/stats/__init__.py in _get_fit_model_data(self, data, model)
    202         data, model = self._validate_inputs(data, model)
    203         fitdata = data.to_fit(staterrfunc=self.calc_staterror)
--> 204         modeldata = data.eval_model_to_fit(model)
    205 
    206         return fitdata, modeldata

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/data.py in eval_model_to_fit(self, modelfuncs)
    519 
    520         for func, data in izip(modelfuncs, self.datasets):
--> 521             total_model.append(data.eval_model_to_fit(func))
    522 
    523         return numpy.concatenate(total_model)

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/data.py in eval_model_to_fit(self, modelfunc)
    188 
    189     def eval_model_to_fit(self, modelfunc):
--> 190         return modelfunc(*self.get_indep(filter=True, model=modelfunc))
    191 
    192     def _get_indep(self, filter=False):

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/models/model.py in __call__(self, *args, **kwargs)
    263             return self
    265 

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/models/basic.py in calc(self, *args, **kwargs)
   1486     def calc(self, *args, **kwargs):
   1487         kwargs['integrate'] = bool_cast(self.integrate)
-> 1488         return _modelfcts.gauss2d(*args, **kwargs)
   1489 
   1490 

TypeError: array must have 0 or 1 dimensions
```

### LINKED ISSUE #628: Evaluating a model on `Data2DInt` fails in `eval_model_to_fit`
See code below. Note that this works with `Data2D` but not with `Data2DInt` as shown below. It looks like a bug, but given how extensive Sherpa is tested I would be surprised if a bug like that was able to slip through, so it might also be a user error similar to #627 where I did not set something up correctly because I'm using the object-oriented interface.
```
(saba)~% ipython
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
IPython 7.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np 
   ...: from sherpa.stats import Chi2 
   ...: from sherpa.optmethods import LevMar 
   ...: from sherpa.estmethods import Covariance 
   ...: from sherpa.data import Data1D, Data1DInt, Data2D, Data2DInt 
   ...: from sherpa.models import Gauss2D 
   ...: from sherpa.fit import Fit 
   ...:  
   ...: x0low, x0high = 3000, 4000 
   ...: x1low, x1high = 4000, 4800 
   ...: dx = 15 
   ...: x1, x0 = np.mgrid[x1low:x1high:dx, x0low:x0high:dx] 
   ...: y = (x0-3500)**2 + (x1 - 4500)**2 
   ...:  
   ...: data2i = Data2DInt('name', x0.flatten(), x0.flatten()+dx, x1.flatten(), x1.flatten()+dx, y.flatten()
   ...: , staterror=np.sqrt(y).flatten()) 
   ...: model2 = Gauss2D() 
   ...: fitter = Fit(data2i, model2, Chi2(), LevMar(), Covariance()) 
   ...:  
   ...: fitter.fit() 
   ...:                                                                                                     
<ipython-input-1-06090d80a681> in <module>
     17 fitter = Fit(data2i, model2, Chi2(), LevMar(), Covariance())
     18 
---> 19 fitter.fit()

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in run(fit, *args, **kwargs)
     51     def run(fit, *args, **kwargs):
     52         fit.model.startup()
---> 53         result = func(fit, *args, **kwargs)
     54         fit.model.teardown()
     55         return result

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in fit(self, outfile, clobber)
   1228             raise FitErr('binhas0')
   1229 
-> 1230         init_stat = self.calc_stat()
   1231         # output = self.method.fit ...
   1232         output = self._iterfit.fit(self._iterfit._get_callback(outfile, clobber),

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in calc_stat(self)
   1113         """
   1114 
-> 1115         return self._calc_stat()[0]
   1116 
   1117     def calc_chisqr(self):

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/fit.py in _calc_stat(self)
   1094         # TODO: is there anything missing here that
   1095         #       self._iterfit.get_extra_args calculates?
-> 1096         return self.stat.calc_stat(self.data, self.model)
   1097 
   1098     def calc_stat(self):

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/stats/__init__.py in calc_stat(self, data, model)
    542 
    543     def calc_stat(self, data, model):
--> 544         fitdata, modeldata = self._get_fit_model_data(data, model)
    545 
    546         return self._calc(fitdata[0], modeldata,

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/stats/__init__.py in _get_fit_model_data(self, data, model)
    202         data, model = self._validate_inputs(data, model)
    203         fitdata = data.to_fit(staterrfunc=self.calc_staterror)
--> 204         modeldata = data.eval_model_to_fit(model)
    205 
    206         return fitdata, modeldata

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/data.py in eval_model_to_fit(self, modelfuncs)
    519 
    520         for func, data in izip(modelfuncs, self.datasets):
--> 521             total_model.append(data.eval_model_to_fit(func))
    522 
    523         return numpy.concatenate(total_model)

~/anaconda/envs/saba/lib/python3.6/site-packages/sherpa/data.py in eval_model_to_fit(self, modelfunc)
    188 
    189     def eval_model_to_fit(self, modelfunc):
--> 190         return modelfunc(*self.get_indep(filter=True, model=modelfunc))
    191 
    192     def _get_indep(self, filter=False):

TypeError: get_indep() got an unexpected keyword argument 'model'
```

Discovered in astropy/saba#30

olaurino:
# Release Note
Sherpa's basic data classes have been refactored and cleaned up to help facilitating fixing bugs and implementing new features. New tests were added to reduce the chances of introducing regressions in the future.

# Details
This PR fixes #563, fixes #627, and fixes #628.

The fix to #627 hopefully shows how this refactoring was really necessary, as the fix was in before the bug was even reported.

A lot more information on the original, current, and proposed design are available in the internal confluence pages and omitted here for brevity, while other notes from our internal pages are reproduced here because they can inform the code review.

In a couple of instances in the past (e.g. #469 and #383) the current implementation of the basic data classes has been in the way when fixing bugs or improving features. Additionally, when downstream projects tried to extend our data classes they could not find enough documentation within the code or the documentation.

The code was hardly tested at the unit level to begin with and the risk of introducing bugs was significant.

During the refactoring work we have also found that some of the functionality of the base classes was broken, some code was unused, and that the design of the classes had departed from the original design.

This PR introduces ~150 unit tests for the basic data classes. In most cases the tests would pass with the current master branch, as I tried to keep the API unchanged. However I did introduce some regressions intentionally when I thought they would improve the implementation and design while still not breaking the high level API users are more likely to use.

This PR also refactors the basic data classes, with minimal effect to the code outside `sherpa/data.py` and `sherpa/astro/data.py`. I was hoping I could get more into details with `sherpa.astro.data` but I ran out of time. This refactoring can still serve as a blueprint for the refactoring of the astro-specific classes.

Finally, there are very few minor changes made here and there either to support the refactoring (e.g. the removal of `filter`, which was saved and restored in a couple of places, or some minor changes to `EvaluationSpace`, or because they were bothering me during the refactoring with warnings in my IDE. They are very limited in scope and should not add noise to the code review, but I can (re)move them if necessary.

## Intentional regressions
  * DataND name should be undefined. The class did not serve any purpose, was never used directly, and had a misleading name, as Data itself is already an ND-capable class.
  * get_x, get_x0, get_x1, get_err, get_xerr should be undefined for classes where they do not belong, e.g. get_x should not be defined in Data and Data2D(Int), get_x0 should not be defined in Data1D(Int), etc.
  * Same as above for the get_x*label methods
  * Same as above for the to*_plot methods, that only belong to Data1D(Int)
  * Data.notice really has the wrong signature. As this class does not seem to be used directly in the code, I decided to take my chances and change the signature so it makes more sense. Rather than asking to be passed the axislist data the call assumes the object knows itself and can provide that information. The original method was factored out and included in a new class specifically designed to handle filtering (and then grouping, quality, etc). This fixes an issue where the subclasses where redefining the signature of the parent classes.
  * We do not expect to_contour to be defined in Data1D(Int) and to_(component_)plot to be defined in Data2D(Int). And none of them should be defined in Data.

## Potential bugs in the current code
Given the lack of design documents (at least for the Python rework) and unit tests, it is hard to understand whether certain tests are failing because of a bug or (not intuitive?) intended behavior. Here is a list:
  * Data.get_dims does not work, although there is no reason while it should not, especially if one looks at Data as the generic ND data set.
  * Data2D.get_dims when shape is None. The current code deals with the case when self.shape is None. It is not clear whether this case is acceptable (the initializer declares shape as optional though). However the implementation seems incorrect, as in that case the method will return the length of the "meshed" arrays rather than the length of the actual axes. It is not clear whether this is intended or not.
  * Data2D.notice combines the filters on the two axes into one filter for both axes (with a logical AND). This looks like a bug. We probably never noticed because we only use DataIMG.notice2d (note, it's a different method), that uses regions and special logic to deal with filtering in the 2D case.
  * Data2DInt.notice also fails, but because Data2DInt.get_indep doesn't work to begin with, because at some point in the object's lifecycle the mask has a (10, 10) shape while the data's shape is (100, ).
  * The notice issues above also apply to the ignore method, which is really a notice call with the extra ignore=True argument.
  * Data*D.get_indep(filter=True) may return the wrong answer (an unfiltered array) depending on when in the object's lifecycle the method is called. What is more likely to happen is that the user will call the Session's notice method. Then things are OK. However at the unit level when you instantiate the class and then set the filter then the object is in a spurious state and get_indep(filter=True) would return the unfiltered data. I thought I had introduced this bug with some of the arbitrary grid changes but I could reproduce the issue in CIAO 4.10, and plus there are other classes like Data2D that I did not touch and still have the wrong implementation.
  * Most likely for the same reason as get_indep above, the get_y methods seem to fail in the unit tests. And eval_model(_to_fit) as well.
  * Ditto for get_filter(_expr).
  * Ditto for the notice methods in the concrete 1D and 2D classes.

Some of the issues above (the ones that cause `get_indep(filter=True)` when a filter has been set) have to do with the fact that of the two "filtering" fields (`filter` and `mask`) only `mask` seems to be actually used, while `filter` must have been short-circuited at some point in the past and was not used any more. My unit tests tried to set `filter` and expected the filtered results back, but that never happened. Instead, `notice` only results in the `mask` array to be used. I removed the `filter` field altogether, but I can't be sure that the field was not used outside of our tests.

# Questions
  - [ ] `Data2D.get_dims` does not work when shape is None. Should we make it work? What should it return?
  - [ ] `Data1DInt.get_indep` returns a tuple of two arrays, which is the same behavior as `Data2D.get_indep`. In order to understand what the arrays represent (the upper bounds of the same 1D binned independent variable in one case, the two independent variables in the other, the client needs to know which class it is dealing with, which is not right. However, it's not clear whether this can be fixed, as the code seems to necessarily be aware of this ambiguity. For example, the model.calc method can be passed a relatively arbitrary amount (1, 2, or 4) of arrays.
  - [ ] Some of the issues in the current implementation derive from the presence of the `_xlo`, `_xhi`, etc. fields, which maybe were intended to cache the filtered values, but I am not sure. The new implementation does not cache the values, so the filter has to be applied every time the arrays are accessed. I am not sure this is an issue, but it really depends on how and when the accessor methods of the data classes are called. If they were called numerous times during e.g. a fit, then having to apply the filter at every call would hurt performance. Do we need to re-implement a caching mechanism?
  - [ ] `DataSimulFit` has a to_plot method with a weird implementation that almost looks like a debugging thing. Our tests never hit it. Is it even needed? Should it be removed?
  - [ ] `DataPHA` is supposed to extend `Data1DInt`. However the constructor takes a point-like array of channels. `bin_lo` and `bin_hi` are optional and they are generated at run time if they do not exist. Code-wise it looks like `DataPHA` should extend `Data1D` directly, so in order to properly construct the object I changed the base class to `Data1D`. Is this a problem?
  - [ ] Is `DataIMGInt` used at all? There are no tests whatsoever for this class other than some tests introduced for arbitrary grids.
  - [ ] Could we now pull some of the validation in `DataARF` and `DataRMF` up to the `Data1DInt` class?

# Documentation
I did not even try to update the documentation, but it is very likely it needs some updating once this PR is deemed in good shape.
Similarly, before adding more docstrings or updating the current ones I would like to make sure the changes are reviewed.

# Disclaimer
While I tried to keep the API as unchanged as possible, this is a high risk change, especially because of the lack of unit level tests. Some of Sherpa's functionality is tested elsewhere (e.g. Lev3, SDS regression tests), so we need to test this PR through and, if necessary, add more tests to the repo.


olaurino:
According to Travis there are two issues. First, I had left behind the wrong call to `apply_filter` in DataPHA while I was trying to keep it to extend `Data1DInt`. My tests passed [because I was using Python 3, but they fail with Python 2](https://riptutorial.com/python/example/1401/bound--unbound--and-static-methods).

The other thing is that I wanted pytest's `XFAIL` to be strict. The only problem was that the `xspec.set_chatter` test was passing for me while it was marked `XFAIL`, thus making the tests fail, per my request. I thought the `XFAIL` marker was not needed anymore, but apparently it is. I don't think I need `xfail=strict` anymore at this point, so I'll remove that. However it would be good if we could make the `xspec.set_chatter` test pass in a different PR.

DougBurke:
@olaurino - do you think it's worth SDS to run their regression tests on this PR now, or wait a while?

olaurino:
@DougBurke the sooner the better. This is already "stable" enough. Thanks!

DougBurke:
I am "conceptually" concerned about moving `DataPHA` from `Data1DInt` to `Data1D`, as the PHA class is meant to represent "binned" data, it's just that unlike the generic case we know each bin has a width of 1, so we can get away with giving a single value to fully-indicate each bin (rather than having to give xlo/xhi as you do in the general case).

anetasie:
I always assumed that  ```DataPHA``` is sort of equivalent of ```Data1DInt```. It is binned in the sense that the models that will be used for these data will be integrated over the energy bin. The input data array lists channels, but then the ```ARF```  sets the energy grid for the models. For this reason and because there are associated calibration data that are required for modeling the ```PHA``` spectral data, the separate ```DataPHA``` class was introduced. Isthe requirement of ```ARF/RMF``` handled by  ```Data1DInt``` class?

DougBurke:
@anetasie - the ARF/RMF handling is tightly coupled with the DataPHA class, not Data1DInt. For example the `sherpa.astro.instrument.RSPModelPHA` class (although there is a pha-less version; `sherpa.astro.instrument.RSPModelNoPHA`). I don't know how much of the functionality they use comes directly from `Data1DInt` versus `DataPHA` (if any).

olaurino:
I have not received much feedback on these questions, so here is how I am inclined to answer them lacking a definite answer from other team members. Note there is one question that does seem to require consensus before we can merge this.

I'll also be reviewing Kenny's report again to make sure I didn't forget anything (and maybe I'll add a few tests for the output format, if feasible).

  > `Data2D.get_dims` does not work when shape is None. Should we make it work? What should it return?

This PR returns `len(self.get_x0(filter)), len(self.get_x1(filter))` if `shape` is None. I am keeping this behavior.

  > `Data1DInt.get_indep` returns a tuple of two arrays, which is the same behavior as `Data2D.get_indep`. In order to understand what the arrays represent (the upper bounds of the same 1D binned independent variable in one case, the two independent variables in the other, the client needs to know which class it is dealing with, which is not right. However, it's not clear whether this can be fixed, as the code seems to necessarily be aware of this ambiguity. For example, the model.calc method can be passed a relatively arbitrary amount (1, 2, or 4) of arrays.

I am not going to address this for the time being.

  > Some of the issues in the current implementation derive from the presence of the `_xlo`, `_xhi`, etc. fields, which maybe were intended to cache the filtered values, but I am not sure. The new implementation does not cache the values, so the filter has to be applied every time the arrays are accessed. I am not sure this is an issue, but it really depends on how and when the accessor methods of the data classes are called. If they were called numerous times during e.g. a fit, then having to apply the filter at every call would hurt performance. Do we need to re-implement a caching mechanism?

I don't think there is a performance issue here, but I am going to do some profiling to verify this PR is not inadvertently making things less efficient.

  > `DataSimulFit` has a to_plot method with a weird implementation that almost looks like a debugging thing. Our tests never hit it. Is it even needed? Should it be removed?

I am going to remove the method since it clearly looks like it was included by mistake.

  > `DataPHA` is supposed to extend `Data1DInt`. However the constructor takes a point-like array of channels. `bin_lo` and `bin_hi` are optional and they are generated at run time if they do not exist. Code-wise it looks like `DataPHA` should extend `Data1D` directly, so in order to properly construct the object I changed the base class to `Data1D`. Is this a problem?

**We need to reach a consensus on this before we can move forward.**

  > Is `DataIMGInt` used at all? There are no tests whatsoever for this class other than some tests introduced for arbitrary grids.

I don't think this class is used but I cannot just remove it, so I won't.

  > Could we now pull some of the validation in `DataARF` and `DataRMF` up to the `Data1DInt` class?

We probably can, but I am not going to do that at this time.

anetasie:
How all the information that is associated with `DataPHA` class is going to be included in 'Data1DInt' class? There is 'association' between the information that comes with the PHA files which needs to be handled later on. It is not only `ARF/RMF`.  Our Intro Thread shows the basic interaction by the user:
http://cxc.harvard.edu/sherpa4.11/threads/pha_intro/

The UI returns the following for the PHA file plus the associated ARF/RMF/BKG files:

```
Data Set: 1
Filter: 0.1248-12.4100 Energy (keV)
Bkg Scale: 0.134921
Noticed Channels: 1-1024
name           = 3c273.pi
channel        = Float64[1024]
counts         = Float64[1024]
staterror      = None
syserror       = None
bin_lo         = None
bin_hi         = None
grouping       = Int16[1024]
quality        = Int16[1024]
exposure       = 38564.6089269
backscal       = 2.52643646989e-06
areascal       = 1.0
grouped        = True
subtracted     = False
units          = energy
rate           = True
plot_fac       = 0
response_ids   = [1]
background_ids = [1]
'''

PR #615, opened 2019-04-03
Asymmetric Errors

dtnguyen2:
# Release Note
Sherpa now supports asymmetric error bars. Errors can be read through a new `load_ascii_with_errors` high level function, or through the new `Data1DAsymmetricErrs` class. Sherpa uses bootstrap for estimating the uncertainties.

# Notes
The requirements for the PR can be found at the following url:
https://cxcjira.cfa.harvard.edu/confluence/display/SH/Asymmetric+Error+Bars

I wrote some docs for load_ascii_with_errors but I need SDS to write docs for resample_data


dtnguyen2:
Requirements for this PR are in:

https://cxcjira.cfa.harvard.edu/confluence/display/SH/Asymmetric+Error+Bars

anetasie:
@dtnguyen2 What does `resample_data` return now? Is this 1 sigma or best_fit_val-1sigma?
Also, when I run for a fit with 2 model parameters I get 4 values, but there is no par names, so I'm not sure which of the values belong to which parameter.

sherpa/sim/tests/test_asymmetric.py
```diff
@@ -0,0 +1,182 @@
...
+from sherpa.data import Data1DAsymmetricErrs
+from sherpa.optmethods import LevMar
+from sherpa.utils.testing import SherpaTestCase
+from sherpa.models import PowLaw1D
+from sherpa.stats import Chi2Gehrels
+from sherpa.estmethods import Covariance
+from sherpa.sim import ReSampleData
+from sherpa.astro import ui
+
+class test_sim(SherpaTestCase):
```
DougBurke: This class needs the `@requires_data` decorator, since `setUp`  calls `make_path`.

`from sherpa.utils.testing import requires_data`

Also, we're moving away from using `SherpaTestCase` and moving to `py.test` style tests. I don't know if we want to make this a requirement of new tests or not @olaurino.

DougBurke: I tjhink it also needs the `@requires_fits` decorator (form the same location). At least I see some of the travis tests fail with `AttributeError: module 'sherpa.astro' has no attribute 'io'` messages which make it look like this decorator is needed.

sherpa/ui/utils.py
```diff
@@ -10412,6 +10412,30 @@ def get_draws(self, id=None, otherids=(), niter=1000, covar_matrix=None):
...
+        returns the best fit parameters for each realization, and average and
+        standard deviation for the total number of realizations.
+
+        Example
+        -------
+        Account for of asymmetric errors when calculating parameter
+        uncertainties:
+
+        >>> load_ascii_with_errors(1,’test.dat’)
+        >>> set_model(‘polynom1d.p0’)
```
DougBurke: The quotes here appear to be UTF-8 (or some other encoding) - which causes the Python 2.7 tests to fail. My guess is that it's the opening quote. Can we change to `'` for both (as it is likely to cause less problems with the documentation builds too).

DougBurke: So, the error was (from the 0ea2317 run) https://travis-ci.org/sherpa/sherpa/jobs/547315894#L4113 which I was expecting to be fixed by @dtnguyen2 5a1ade4 commit. However, the 006d8a6 commit passes, which is before the fix. So I just don't understand things anymore.

anetasie:
I reviewed and run the updated code. It looks good. I sent the final updates to the documentation example to @dtnguyen2. The code will be ready to merge after the doc update.

sherpa/astro/ui/utils.py
```diff
@@ -1325,6 +1443,12 @@ def load_data(self, id, filename=None, *args, **kwargs):
            The keyword arguments supported by `load_pha`, `load_image`,
            `load_table`, and `load_ascii`.
 
+        Returns
```
DougBurke: `load_data` doesn't return anything, so this needs to be removed.

sherpa/ui/utils.py
```diff
@@ -10412,6 +10412,73 @@ def get_draws(self, id=None, otherids=(), niter=1000, covar_matrix=None):
             fit, covar_matrix, niter=niter)
         return (stats, accept, params)
 
+    def resample_data(self, id=None, niter=1000, seed=123):
```
DougBurke: I think there's a possibility for user confusion here. `resample_data` is defined in `sherpa.ui` (and so will also be available to users of `sherpa.astro.ui`) but `load_ascii_with_errors` is only available to users of `sherpa.astro.ui` (because of it's use of `sherpa.astro.io`).

Can we make a version of `load_ascii_with_errors` that can be used from `sherpa.ui` - or are existing I/O functions for this module not sufficient? (There is also the fact that I think the way we extend functionality in `sherpa.astro.ui` by copying code is not ideal but that's a different PR...).

sherpa/sim/__init__.py
```diff
@@ -758,3 +761,89 @@ def calc_stat(proposed_params):
...
+
+class ReSampleData(NoNewAttributesAfterInit):
+
+    def __init__(self, data, model):
+        self.data = data
+        self.model = model
+        NoNewAttributesAfterInit.__init__(self)
+        return
+    
+    def __call__(self, niter=1000, seed=123):
```
DougBurke: I think it would be better to have `seed=None` rather than `seed=123`, since it will then use a "random" seed by default. Or is that not the intended use case?

sherpa/sim/__init__.py
```diff
@@ -758,3 +761,89 @@ def calc_stat(proposed_params):
         stats = -2.0 * stats
 
         return (stats, accept, params)
+
+
+class ReSampleData(NoNewAttributesAfterInit):
```
DougBurke: I know this is not a well-documented module, but can we add some information here about  what this class is for?

DougBurke:
@anetasie I am a bit concerned - as noted above - about `load_ascii_with_errors` being in `sherpa.astro.ui` but `resample_data` being in `sherpa.ui`. Do we need a version of `load_ascii_with_errors` in `sherpa.ui`, or is `resample_data` useful even without `load_ascii_with_errors`?

anetasie:
@DougBurke - yes, `resample_data` is useful also with other data, so without `load_ascii_with_errors`.

sherpa/sim/__init__.py
dtnguyen2: will change the seed from 123 to None

dtnguyen2:
If ```resample_data``` is "useful also with other data" then that is news to me since the underlying class ```ReSampleData``` requires the ```Data1DAsymmetricErrs```

sherpa/sim/__init__.py
dtnguyen2: Added the following comment to the ```ReSampleData``` class


    The underlying low-level for the resample_data

    Parameters
    ----------
    data
       The data class Data1DAsymmetricErrs
    model
       The model to fit the data

    Returns
    -------
    The class returns the best fit parameters for each realization.
    
    See Also
    --------
    resample_data

    Example
    -------

    >>> load_ascii_with_errors(1, 'gro.txt', delta=False)
    >>> data = get_data(1)
    >>> method = LevMar()
    >>> model = PowLaw1D('p1')
    >>> set_model(1, model)
    >>> fit = Fit(data, model, Chi2Gehrels(), method, Covariance())
    >>> results = fit.fit()
    >>> rd = ReSampleData(data, model)
    >>> rd_results = rd(niter=10)
    >>> print(rd_results)
    rd_results = {'p1.gamma': [-0.2944764463053398, -0.48025660808404, -0.45290026618472473, -0.5238444562856396, -0.3549169211965681, -0.29119982744489403, -0.5136099861402832, -0.49899714779391674, -0.5308025556771122, -0.5645228923615183], 'p1.ampl': [60.757254200932465, 172.7375230511183, 149.6714174684889, 222.60620753447844, 87.33459341889869, 59.257672491708, 212.86125707286197, 194.83475286439503, 233.33238676025326, 275.75315162410527]}

anetasie:
@DougBurke this function was intended for the case with asymmetric errors and not to be a general function, which needs more complex spec to handle different distributions (not only normal) and possible standard bootstraps etc.

PR #616, opened 2019-04-08
enable run-time-option to test cache (default=True)

dtnguyen2:
# Release Note
This PR enables the user to turn off caching at runtime for testing purpose.

# Description
In order to test PR #503 (and to certain extent #444).
SDS made the following RFE:

https://icxc.harvard.edu/pipe/sherpadev/2019/0014.html

So for this PR, the default model caching value is set to True.  However, one
can turn it off at run time by setting cache=False when calling the fit command
as shown below:
(sherpa3)[dtn@devel12 integrate]$ diff False.py True.py
84c84
```
<     fit(cache=False)
---
>     fit()
```

(sherpa3) [dtn@devel12 integrate]$ time False.py &> false.txt

real    0m35.689s
user    0m35.649s
sys    0m2.293s
(sherpa3) [dtn@devel12 integrate]$ time True.py &> true.txt

real    0m18.527s
user    0m18.522s
sys    0m2.283s

They even give the same results:
```
(sherpa3) [dtn@devel12 integrate]$ diff false.txt true.txt
37c37
< method = simplex, stat = cstat, in 34.033469915390015 secs
---
> method = simplex, stat = cstat, in 16.860190629959106 secs
```
etc...

caveat emptor: I modified enough of the startup method to en/dis-able the
caching to pass the built-in test.  I do not claim to have done all of them, in
fact a simple check shows that there are some startup methods that have not
been modified:

(sherpa3) [dtn@devel12 sherpa]$ fstr startup | grep -v .pyc | grep -v cache | grep -v "#" | grep -v _startup_ | grep -v = | grep -v testing
./fit.py:48:    """Fit object decorator that runs model startup() and teardown()
./fit.py:56:            fit.model.startup(True)
./astro/instrument.py:199:    def startup(self):
./astro/instrument.py:200:        self.model.startup()
./astro/instrument.py:201:        CompositeModel.startup(self)
./astro/instrument.py:411:    def startup(self):
./astro/instrument.py:432:        ARFModel.startup(self)
./astro/instrument.py:864:    def startup(self):
./astro/instrument.py:871:        CompositeModel.startup(self)
./astro/instrument.py:919:                self._startup_user_grid(x, xhi)
./astro/instrument.py:1026:    def _startup_user_grid(self, x):
./astro/instrument.py:1047:                self._startup_user_grid(x)
./models/basic.py:1868:    def startup(self):
./models/basic.py:1869:        self.model.startup()
./models/basic.py:1871:        CompositeModel.startup(self)
./models/model.py:208:    def startup(self):
./models/model.py:675:    def startup(self):

Moreover, some functions such as get_draws, sigmarej and some plotting
functions etc... had to have an extra argument (cache=True) to pass the test.

After I had checked in the caching code, Aneta wanted me to cache the ARF
fold code.  My suggestion is to remove tit:
```
@myCacheARF
def my_arf_fold(a, b):
    return arf_fold(a, b)
```
since arf_fold is nothing more then element-wise multiplications of two numpy
arrays which doesn't gain you much by caching it since it waste a bit of memory
and some computation time for checking the cache etc...


DougBurke:
I've restarted the failing travis run (it was the macOS job) since it had timed out / some other error unrelated to these changes.

PR #617, opened 2019-04-12
WIP update pyblocxs: 

dtnguyen2:
# Release Note
The requirements for this PR can be found at:
https://cxcjira.cfa.harvard.edu/confluence/display/SH/Update+pyBlocxs

# Description
Aside from the following minor change, restricting the fit-statistic to only be ```Cash``` to allow the fit statistic to be one of the following ```Cash, CStat``` or ```WStat``` to be consistent with the ```get_draws```. The only work done for this PR was to convert the prototype which can be found at:  
https://github.com/astrostat/pyblocxs
to run on python 3.

# Note
I need the following from SDS:
*1)* What should the function be called, the prototype calls it 'mh_sampling_newdata'.
*2)* I need the doc for the new functionality.
*3)* The function dmvt is defined in the currrent version of sherpa (see file sim/mh.py)
 and in the prototype but they have different functionality.  Should dmvt in the sim/mh.py 
file be replaced?

(sherpa3) [dtn@devel12 pyblocxs]$ grep dmvt /export/sherpa/sherpa/sim/mh.py | grep def
```
def dmvt(x, mu, sigma, dof, log=True, norm=False):
    def dmvt(self, x, log=True, norm=False):
```
versus
sherpa3) [dtn@devel12 pyblocxs]$ grep dmvt mhtest.py | grep def
```
    def dmvt(self, x, mu, sigma, df, log = True, norm = False):
```

*4)* The constant ```con_prob_prev``` is currently hard-wired to 100 in the included 
prototype mhtest.py, should it be an option for the new function.

(sherpa3) [dtn@devel12 sherpa]$ grep con_prob_prev /export/sherpa/sherpa/sim/mhtest.py
```
        con_prob_prev = 100
                    alpha = np.exp(stat_proposal - stat_prev + con_prob_prev - \
                    alpha = np.exp(stat_proposal - stat_prev + con_prob_prev - \
                    con_prob_prev = con_prob_proposal
```

# Caveat
user may encounter the following errors:
```
    raise RuntimeError("Error: con_var is not positive semi-definite")
RuntimeError: Error: con_var is not positive semi-definite
```
or  
```
    raise ValueError("Error: sigma is not symmetric")
ValueError: Error: sigma is not symmetric
```
which I have not attempted to fix

### DIFF
--- a/sherpa/astro/ui/tests/test_mhtest.py
+++ b/sherpa/astro/ui/tests/test_mhtest.py
@@ -1,0 +1,95 @@
+#
+#  Copyright (C) 2019  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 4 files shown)


dtnguyen2:
See PR #1364

PR #622, opened 2019-04-30
Do not create warnings about error bars in plots if no error bars are to be shown (fix #621)

### LINKED ISSUE #621: plots should not complain about missing errors when they are not displaying them
I don't particularly like how we automatically add error bars to values in plots, but can understand where it comes from (primarily from displaying PHA data).

Here's an example when the behavior is not optimal: we get a warning from `plot_data` that errors are being displayed but not used in the fit once we switch to the least-squares statistic. I then turn off the display of error bars in the plot, but I still get the warning.

```
In [1]: from sherpa import ui

In [2]: ui.load_arrays(1, [1, 2, 3], [4, 5, 6])

In [3]: ui.plot_data()

In [4]: ui.set_stat('leastsq')

In [5]: ui.plot_data()
WARNING: The displayed errorbars have been supplied with the data or calculated using chi2xspecvar; the errors are not used in fits with leastsq

In [6]: ui.get_data_plot_prefs()['yerrorbars'] = False

In [7]: ui.plot_data()
WARNING: The displayed errorbars have been supplied with the data or calculated using chi2xspecvar; the errors are not used in fits with leastsq

```

At this point the plot does not show errors, but we still get the WARNING message.

If I create a multi-panel plot then I get the warning several times which is visually distracting - ideally we'd only get the message once (note that in this case the second plot, the residual one, does actually display error bars, so there's an argument that the warning should appear once here, rather than not at all, but you do get the message twice when you don't unset `yerrorbars`).

```
In [8]: ui.set_source(ui.const1d.flat)

In [9]: ui.plot_fit_resid()
WARNING: The displayed errorbars have been supplied with the data or calculated using chi2xspecvar; the errors are not used in fits with leastsq
WARNING: The displayed errorbars have been supplied with the data or calculated using chi2xspecvar; the errors are not used in fits with leastsq

In [10]:   
```

I'd also like a way to turn off the error bars in the residual plot, but we don't seem to have a "resid" style preferences command (at least at the ui layer), but this is really a different issue:

```
In [10]: ui.get*prefs?
ui.get_data_contour_prefs
ui.get_data_plot_prefs
ui.get_model_contour_prefs
ui.get_model_plot_prefs

```

DougBurke:
# Summary

Remove the  `The displayed errorbars have been supplied with the data or calculated using chi2xspecvar...` warnings that appear for data, residual/ratio, and fit plots when the user has explicitly turned off the display of errorbars (by setting the `yerrorbars` plot-preference setting to `False`).

This does **not** address the problem of this message appearing multiple times for "combined" plots like `fit_plot_resid`, as it is not immediately obvious to me how to do this.

# Details

Add several tests to check that we are creating the "The displayed
errorbars have been supplied with the data or calculated using
chi2xspecvar; the errors are not used in fits with ..." warning messages
when creating plots.

These tests do not require a plotting backend. However, given that the
behavior is "stateful", in that the tests could fail if some other
part of the system has changed the plot preferences, then there are
additional checks (beyond those needed to perform the checks) that
indicate whether the system state has been changed.

The caplog fixture, introduced in pytest 3.3, is used to capture the
logging output (which is how the warnings are displayed to the user;
they do not use the warnings module). There are other tests in the
system that change the system logging level, so these tests use the
caplog context mamanger to ensure the tests are run with the standard
logging level.

The tests contain a lot of repeated code, but it's not clear whether it is
worth consolidating it.

There does appear to be repeated calculation of the y error bars in the
plot routines - `data.to_plot` is called and then the errors are re-calculated.
This is outside of the scope of this PR to look at, but I've added some
comments about it.

The development of this PR hit a snag with the existing plot tests, in particular
how the chips setup is mocked. I had to change code to test for attributes
rather than use `isinstance`, as documented below, which is more-idiomatic
Python, but may point to some work needed for the mock code.

The PR should be rebased to clean up the testing code, as I was tracking down
the mock/isinstance issue, but I want to leave the history as is for now.

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -564,6 +564,10 @@ class JointPlot(SplitPlot):
         clearaxes = kwargs.get('clearwindow', True)
         self._clear_window(0, 0, clearaxes)
 
+        print("INTERNAL DEBUGGING: JointPlot")
+        print(" isinstance(plot, FitPlot) = {}".format(isinstance(plot, FitPlot)))
+        print(" type(plot) = {}".format(type(plot)))
+
         # FIXME: should not know about FitPlot, terrible hack to remove label
         if isinstance(plot, FitPlot):
             plot.dataplot.xlabel = ''
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -1274,7 +1274,7 @@ class ResidPlot(ModelPlot):
     x : array_like
        The X value for each point.
     y : array_like
-       The Y value for each point: data-model.
+       The Y value for each point: data - model.
     xerr : array_like
        The half-width of each X "bin", if set.
     yerr : array_like
@@ -1349,7 +1349,7 @@ class RatioPlot(ModelPlot):
     x : array_like
        The X value for each point.
     y : array_like
-       The Y value for each point: data-model.
+       The Y value for each point: data / model.
     xerr : array_like
        The half-width of each X "bin", if set.
     yerr : array_like
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -564,12 +564,34 @@ class JointPlot(SplitPlot):
         clearaxes = kwargs.get('clearwindow', True)
         self._clear_window(0, 0, clearaxes)
 
+        """
         # FIXME: should not know about FitPlot, terrible hack to remove label
         if isinstance(plot, FitPlot):
             plot.dataplot.xlabel = ''
             plot.modelplot.xlabel = ''
         else:
             plot.xlabel = ''
+
+        """
+
+        # Slightly-more Pythonic way to clear the labels than checking
+        # the class of the object.
+        #
+        # The order here could matter if things get changed (such as
+        # FitPlot being re-worked to use a common labelling system)
+        #
+        # Unlike the previous version, this will not fail (here at least)
+        # if xlabel doesn't exist.
+        #
+        if hasattr(plot, 'xlabel'):
+            plot.xlabel = ''
+        elif hasattr(plot, 'dataplot') and hasattr(plot, 'modelplot'):
+            dplot = plot.dataplot
+            mplot = plot.modelplot
+            if hasattr(dplot, 'xlabel'):
+                dplot.xlabel = ''
+            if hasattr(mplot, 'xlabel'):
+                mplot.xlabel = ''
 
         kwargs['clearwindow'] = False
         plot.plot(*args, **kwargs)
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -110,6 +110,25 @@ def _make_title(title, name=''):
         return "{} for {}".format(title, name)
 
 
+def _errorbar_warning(stat):
+    """The warning message to display when error bars are being "faked".
+
+    Parameters
+    ----------
+    stat : sherpa.stats.Stat instance
+        The name attribute is used in the error message.
+
+    Returns
+    -------
+    msg : str
+        The warning message
+    """
+
+    return "The displayed errorbars have been supplied with the " + \
+        "data or calculated using chi2xspecvar; the errors are not " + \
+        "used in fits with {}".format(stat.name)
+
+
 class Plot(NoNewAttributesAfterInit):
     "Base class for line plots"
     plot_prefs = backend.get_plot_defaults()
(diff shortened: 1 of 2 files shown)


DougBurke:
Aha, one of the "checks" I added to make sure things weren't being run "with inappropriate settings" (in this case without y error bars) is being triggered here when there is a back-end selected. I think it's because I was being "too clever" and had included model prefs as well, which I can just remove.

DougBurke:
Hmm, perhaps I should actually try testing this myself with the matplotlib backend in place rather than rely on travis to spot the problems ....

DougBurke:
I do not know why the Travis tests are failing: they succeed on my local machine. So debugging by print statement...

DougBurke:
So in the tests I create `fplot` as a `FitPlot` instance and send it as the `plot` argument to JointPlot's `plottop` method. Here's what I see on Travis

```
INTERNAL DEBUGGING 2
 isinstance(fplot, FitPlot) = True
 type(fplot) = <class 'sherpa.plot.FitPlot'>
```

in the test code, and then when it gets to `plottop`:

```
INTERNAL DEBUGGING: JointPlot
 isinstance(plot, FitPlot) = False
 type(plot) = <class 'sherpa.plot.FitPlot'>
```

Somehow the `isinstance` check is now returning `False`. WAT?

DougBurke:
Oh. Apparently I didn't see this because I didn't have the sherpa data files submodule installed. Obviously I can see why having data files present in a test that doesn't use data files can make it fail. WAT?

DougBurke:
And even with the data files loaded, 

```
% python setup.py test -a sherpa/plot/tests/test_plot_warnings.py 
```

passes but

```
% python setup.py test
```

fails.

DougBurke:
So I was able to fix the tests, but only by changing the code to avoid an `isinstance` call. I don't know why the test was failing, and it might be useful to know what in our environment lead to this failure.

DougBurke:
Oooh. I wonder if the failing test is due to the `mock_chips` fixture, since there's a note in `sherpa/conftest.py` which says (as it re-imports a bunch of modules) `# These are required because otherwise Python will not match imported classes.`.

I wonder if there's some weird interaction between loading in chips and pylab backends.

DougBurke:
It does seem to be due to `sherpa/astro/ui/tests/test_chips_unit.py` (or, rather, the fixtures it calls). I tried reverting the change I made in 82a75ae and so two tests in `test_plot_warnings` fail, but if I changed `test_chips_unit.py` so that the tests in it didn't run (by renaming `test_plot` to `_test_plot`) then the two tests pass.

I think this has some implication for any future plot tests we might add. So should the `mock_chips` fixture be re-worked to address this?

DougBurke:
Oops, the commit log for db73c71 has a typo in it - it should say #621 not #612

sherpa/plot/__init__.py
```diff
@@ -564,13 +585,35 @@ def plottop(self, plot, *args, **kwargs):
         clearaxes = kwargs.get('clearwindow', True)
         self._clear_window(0, 0, clearaxes)
 
+        """
         # FIXME: should not know about FitPlot, terrible hack to remove label
```
wmclaugh: Do we need/want to leave the old code and FIXME note or should we delete it?

sherpa/plot/__init__.py
```diff
@@ -564,13 +585,35 @@ def plottop(self, plot, *args, **kwargs):
...
 
+        """
+
+        # Slightly-more Pythonic way to clear the labels than checking
+        # the class of the object.
+        #
+        # The order here could matter if things get changed (such as
+        # FitPlot being re-worked to use a common labelling system)
+        #
+        # Unlike the previous version, this will not fail (here at least)
```
wmclaugh: It is good that you check for the attribute before accessing it, but the here comment is superfluous.

sherpa/plot/__init__.py
```diff
@@ -39,6 +40,7 @@
 
 warning = logging.getLogger(__name__).warning
 
+# TODO: why is this module globally chaning the invalid mode of NumPy?
```
wmclaugh: chaning <- changing

sherpa/plot/__init__.py
DougBurke: I think we can probably remove this comment: the following `isinstance` check is a sign that something is less-than ideal.

sherpa/plot/__init__.py
DougBurke: Okay - I'll remove it then.

DougBurke:
@wmclaugh I've made the comment changes. Do you want these commits squashed (in particular the earlier ones) or are you happy as is.

wmclaugh:
@DougBurke I'm fine with them as is. I'll merge the commit after the travis tests complete.

PR #626, opened 2019-05-12
add plot_fit_ratio and plot_bkg_fit_ratio functions

DougBurke:
# Overview

Add the `plot_fit_ratio` function to the `sherpa.ui` layer (to match `plot_fit_resid` and `plot_fit_delchi`) and `plot_bkg_fit_ratio` to the `sherpa.astro.ui` layer. Added tests for a number of plot types.

# Details

Most of the commits are adding tests (and can probably be merged into a smaller number of commits, but I want to leave the existing set of commits for initial review).

The tests are very limited, in that although they tend to require a plotting backend, they don't actually check that the plot was actually created. Instead, they test that the plot data structures (such as `DataPlot`) are populated correctly. The `replot` argument is tested since the preference-handling code in the `plot_fit_xxx` routines looked "interesting", and I wanted a reasonable base of tests in case I decided to refactor things.

I think that the `@requires_plotting` decorator could be removed from many (maybe all) of these tests if we required that the dummy plot backend set up sensible default preferences. That's a call for a different PR. I do believe that I see the same issue here as found in the tests for #622 - the way that the chips mocking works can lead to odd downstream behavior.

The `plot_bkg`-related tests show some odd behavior that needs investigating, but is not due to this PR (it may be due to how the synthetic data used in the tests is created).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   fast_finish: true
   include:
     # macOS build
-    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=2 XSPECVER="12.10.1b" TRAVIS_PYTHON_VERSION="3.7"
+    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=3 XSPECVER="12.10.1b" TRAVIS_PYTHON_VERSION="3.7"
       os: osx
     # Barebone build to check tests pass when most of the tests must be skipped.
     # We need to use TEST=none to remove the test-data submodule.
--- a/docs/ui/sherpa_astro_ui.rst
+++ b/docs/ui/sherpa_astro_ui.rst
@@ -288,6 +288,7 @@ burden...
       plot_energy_flux
       plot_fit
       plot_fit_delchi
+      plot_fit_ratio
       plot_fit_resid
       plot_kernel
       plot_model
--- a/docs/ui/sherpa_astro_ui.rst
+++ b/docs/ui/sherpa_astro_ui.rst
@@ -276,6 +276,7 @@ burden...
       plot_bkg_delchi
       plot_bkg_fit
       plot_bkg_fit_delchi
+      plot_bkg_fit_ratio
       plot_bkg_fit_resid
       plot_bkg_model
       plot_bkg_ratio
--- a/docs/ui/sherpa_ui.rst
+++ b/docs/ui/sherpa_ui.rst
@@ -194,6 +194,7 @@ hidden away. This needs better explanation...
       plot_delchi
       plot_fit
       plot_fit_delchi
+      plot_fit_ratio
       plot_fit_resid
       plot_kernel
       plot_model
--- a/sherpa/astro/tests/test_astro_plot.py
+++ b/sherpa/astro/tests/test_astro_plot.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2015, 2018, 2019  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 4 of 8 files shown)


DougBurke:
My guess this is failing for the same reason that I had problems with #622 - the mocking of objects means that `isinstance(fplot, sherpa.plot.FitPlot)` falls over because `sherpa.plot.FitPlot` has been redefined somewhere.

DougBurke:
So, some of the tests fail because the plot preferences are not set up correctly when there's no plot backend. I'm going to gate these tests (i.e. only run them if there is a plot backend), but we could change the plto code to have fall backs for the plot preferences (in this case it's a missing `xlog` setting, but if we fix this there's probably others that would need addressing).

DougBurke:
The tests added for the `plot_bkg` family of routines are producing surprising results (same values for channel vs energy settings) which aren't due to the code changes here, but may be due to how I create the synthetic data used in the tests. This needs investigation but not necessarily as part of this PR.

DougBurke:
The test is failing on the macOS build, which is using python 3.7 and matplolib version 2.2.3. This causes the following warning to be displayed

```
/Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/__init__.py:107
  /Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/__init__.py:107: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import MutableMapping
/Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/rcsetup.py:20
/Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/rcsetup.py:20
  /Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/rcsetup.py:20: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable, Mapping
/Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/colors.py:53
  /Users/travis/miniconda/envs/build/lib/python3.7/site-packages/matplotlib/colors.py:53: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Sized
```

which triggers a test failure.

We could add these warnings to the "expected set" of warnings, or we could upgrade the macOS build to use matplotlib version 3, where I would expect this has been fixed (we don't see it in the linux matplotlib 3 + python 3.7 test on travis).

DougBurke:
I am trying my suggestion of updating the macOS build to use matplotlib 3 to see what happens.

DougBurke:
For reference, here's the diff in the program coverage comparing before this PR to after it (not rebased to the current master branch):

```
% diff ../coverage-before ../coverage-after 
10c10
< sherpa/astro/data.py                                      1355    382    580     99    68%
---
> sherpa/astro/data.py                                      1355    377    580     98    69%
27c27
< sherpa/astro/plot.py                                       271     51     76     24    74%
---
> sherpa/astro/plot.py                                       271     48     76     20    76%
33c33
< sherpa/astro/ui/utils.py                                  1856    684    782    140    58%
---
> sherpa/astro/ui/utils.py                                  1842    663    778    139    59%
58c58
< sherpa/plot/__init__.py                                   1030    353    278     69    60%
---
> sherpa/plot/__init__.py                                   1030    343    278     72    61%
61c61
< sherpa/plot/pylab_backend.py                               222     99     84     11    45%
---
> sherpa/plot/pylab_backend.py                               222     77     84      5    58%
68c68
< sherpa/ui/utils.py                                        1999    970    628     73    48%
---
> sherpa/ui/utils.py                                        1985    915    624     75    51%
74c74
< TOTAL                                                    22804   8504   6690   1036    59%
---
> TOTAL                                                    22776   8388   6682   1029    60%
```

where the columns are

```
Name                                                     Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------------------------------------
```

So, it does increase coverage, in particular for `pylab_backend.py`.

wmclaugh:
@DougBurke - Can you rebase this PR to remove conflicts so I can merge it?

DougBurke:
Okay - I'll have a go now.

PR #630, opened 2019-06-11
Fix "get_int_proj does not work when recalc=True"

### LINKED ISSUE #543: get_int_proj does not work when recalc=True
# Summary

It looks like `get_int_proj` with `recalc=True` has been broken for a long time (we don't seem to have any tests). I originally suspected something to do with the move to Python 3, but that is just because the underlying error (the `nloop` parameter seems not to be being passed through) results in a slightly-different error message.

It would not surprise me if this also holds for `get_int_unc`.

# Details

The following fails, when I expected it to return a structure:

```
% ipython
Python 3.7.0 (default, Oct  9 2018, 10:31:47) 
IPython 7.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %matplotlib                                                             

In [2]: import numpy as np                                                      

In [3]: x = [-13, -5, -3, 2, 7, 12]; y = [102.3, 16.7, -0.6, -6.7, -9.9, 33.2]; 
   ...: dy = np.ones(6) * 5                                                     

In [4]: from sherpa import ui                                                   

In [5]: ui.load_arrays(1, x, y, dy)                                             

In [6]: ui.set_source(ui.polynom1d.poly)                                        

In [7]: poly.c1.thaw()                                                          

In [8]: poly.c2.thaw()                                                          

In [9]: ui.int_proj(poly.c0)                                                    

In [10]: ui.fit()                                                               
Dataset               = 1
Statistic             = chi2
Initial fit statistic = 469.027
Final fit statistic   = 4.01682 at function evaluation 8
Data points           = 6
Degrees of freedom    = 3
Probability [Q-value] = 0.259653
Reduced statistic     = 1.33894
Change in statistic   = 465.01
   poly.c0        -9.38489     +/- 2.91751     
   poly.c1        -2.41692     +/- 0.250889    
   poly.c2        0.478273     +/- 0.0312677   

In [11]: ui.int_proj(poly.c0)                                                   

In [12]: p1 = ui.get_int_proj()                                                 

In [13]: p2 = ui.get_int_proj(poly.c1, recalc=True, min=-5, max=0, nloop=21)    
<ipython-input-13-39aaae7f3a36> in <module>
----> 1 p2 = ui.get_int_proj(poly.c1, recalc=True, min=-5, max=0, nloop=21)

<string> in get_int_proj(par, id, otherids, recalc, min, max, nloop, delv, fac, log, numcores)

~/sherpa/sherpa-hardcode-xspec-atomdb/sherpa/ui/utils.py in get_int_proj(self, par, id, otherids, recalc, min, max, nloop, delv, fac, log, numcores)
  13385             self._intproj.prepare(min, max, nloop, delv, fac,
  13386                                   sherpa.utils.bool_cast(log), numcores)
> 13387             self._intproj.calc(fit, par, self._methods)
  13388         return self._intproj
  13389 

~/sherpa/sherpa-hardcode-xspec-atomdb/sherpa/plot/__init__.py in calc(self, fit, par, methoddict)
   1791                             " for interval projection plot")
   1792 
-> 1793         xvals = self._interval_init(fit, par)
   1794         oldpars = fit.model.thawedpars
   1795         par.freeze()

~/sherpa/sherpa-hardcode-xspec-atomdb/sherpa/plot/__init__.py in _interval_init(self, fit, par)
   1454             raise ConfidenceErr('badlimits')
   1455 
-> 1456         if self.nloop <= 1:
   1457             raise ConfidenceErr('badarg', 'Nloop parameter', '> 1')
   1458 

TypeError: '<=' not supported between instances of 'NoneType' and 'int'

```

It is not a problem with the choice of parameters since I can call `int_proj` with these values and get a plot:

```
In [14]: ui.int_proj(poly.c1, min=-5, max=0, nloop=21)                          

```

dtnguyen2:
# Release Notes
`get_int_proj` did not work when `recalc=True` on Python 3. This has now been fixed.

# Note
This PR fixes #543
As suggested by issue #543, I also added a test for get_int_unc

### DIFF
--- a/sherpa/ui/tests/test_get_int_proj.py
+++ b/sherpa/ui/tests/test_get_int_proj.py
@@ -29,14 +29,18 @@ logger = logging.getLogger('sherpa')
 
 class test_get_int_proj(SherpaTestCase):
 
-    _get_int_proj_recalc_bench = {
+    _get_int_uncproj_recalc_bench = {
         'x': np.array([-5., -4.75, -4.5, -4.25, -4., -3.75, -3.5, -3.25, -3.,
                       -2.75, -2.5, -2.25, -2., -1.75, -1.5, -1.25, -1., -0.75,
                        -0.5, -0.25, 0.]),
-        'y': np.array([110.0185, 90.493 , 72.9534, 57.3995, 43.8316, 32.2494,
-                      22.6531, 15.0427, 9.4181, 5.7794, 4.1265, 4.4594, 6.7782,
-                      11.0829, 17.3734, 25.6497, 35.9119, 48.1599, 62.3938,
-                       78.6135, 96.8191]),
+        'proj': np.array([110.0185, 90.493 , 72.9534, 57.3995, 43.8316,
+                          32.2494, 22.6531, 15.0427, 9.4181, 5.7794, 4.1265,
+                          4.4594, 6.7782, 11.0829, 17.3734, 25.6497, 35.9119,
+                          48.1599, 62.3938, 78.6135, 96.8191]),
+        'unc': np.array([110.774, 91.1094, 73.4447, 57.78, 44.1153, 32.4507,
+                         22.786, 15.1213, 9.4566, 5.7919, 4.1273, 4.4626,
+                         6.7979, 11.1332, 17.4686, 25.8039, 36.1392, 48.4745,
+                         62.8099, 79.1452, 97.4805]),
         'min': -5,
         'max': 0,
         'nloop': 21,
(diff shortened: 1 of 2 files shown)


DougBurke:
So, the problem stems from the `IntervalProjection.prepare` method not being compatible with the `Confidence1D.prepare` method it over-rides (taken from `sherpa/plots/__init__.py`

```
class IntervalProjection(Confidence1D):

    def __init__(self):
        self.fast = True
        Confidence1D.__init__(self)

    def prepare(self, fast=True, min=None, max=None, nloop=20,
                delv=None, fac=1, log=False, numcores=None):
        self.fast = fast
        Confidence1D.prepare(self, min, max, nloop, delv, fac, log, numcores)
```

and

```
class Confidence1D(DataPlot):
...
    def prepare(self, min=None, max=None, nloop=20,
                delv=None, fac=1, log=False, numcores=None):
        self.min = min
        self.max = max
        self.nloop = nloop
        self.delv = delv
        self.fac = fac
        self.log = log
        self.numcores = numcores
```

where the `fast=True` was added at the start and not end of the list. With the move to Python 3 only we could go for [keyword-only arguments](https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments) for these sorts of routines, but it would be a breaking change.

DougBurke:
It looks like this only happens (the addition of the `fast` argument to the `prepare` method) for `IntervalProjection` and `RegionProjection`, and the `get_reg_proj` routine has the correct ordering. There are other callers of the `prepare` method but they are via mediators that parse out the arguments and then use kwargs support, so they should not be vulnerable to this issue.

My original comment about `"It would not surprise me if this also holds for get_int_unc."` in #543 is incorrect because `get_int_inc` doesn't have a `fast` argument.

olaurino:
Rebased to get #636

PR #631, opened 2019-06-14
Add HTML representations of common classes for IPython display (fix #345)

### LINKED ISSUE #345: should we add html (and others?) representations of objects for ipython/jupyter
I have some old experiments - e.g. 
https://nbviewer.jupyter.org/urls/bitbucket.org/doug_burke/ciao-notebooks/raw/917f54ecc759e94445b5f5b92b9dcbf66872d911/HTML%20output%20%28nicer%20views%29.ipynb whose raw form can be found at https://bitbucket.org/doug_burke/ciao-notebooks/src - to show how we could make HTML views of model/parameter values.

It should just be as simple as adding some `_repr_html_` methods to the desired classes - e.g. see http://ipython.readthedocs.io/en/stable/config/integrating.html 

There's some thought required - e.g. do we want to follow pandas and allow for the output "form" to be configurable or not? Also, do we want to support png/latex/svg where appropriate (e.g. DataPlot could return the plot image (once `prepare` has been called).

DougBurke:
# Summary

An initial version of HTML display support for Sherpa objects for users of IPython/Jupyter notebooks. This covers many of the objects that Sherpa users will create.

The Sherpa notebook (`notebooks/SherpaQuickStart.ipynb`) has been updated to take advantage of this support.

# Notes

You can view the updated notebook at https://nbviewer.jupyter.org/github/sherpa/sherpa/blob/65f23d9763d40c5dfb087ee202f7af0fe8e322a8/notebooks/SherpaQuickStart.ipynb

We leverage the `_repr_html_` method of objects, which is a IPython/Jupyter invention, to create HTML representations of objects. The basic approach is to take the `__str__` output and convert it to a HTML table, and then improve from that, but there are some cases where it diverges.

Since this logic is added to a new class method (i.e. `_repr_html`)

Objects that represent, or contain, "plottable" objects will include a SVG representation of the plot. This is normally taken from the plot object - that is, the plot() method of that object is called - but there are several specialized cases where we create our own plots (RMF and images).

If the Jupyter display is unable to display graphics - such as an untrusted notebook or the pylab backend is not selected - then a textual representation is used (this is included as a fall-back in all cases, using standard approaches for _repr_html).

There has been an attempt to move the display logic into sherpa.utils.formatting, but the actual code to generate the HTML tables and SVG are left to functions called by the `_repr_html_` method. This is similar to how we create the `__str__` and `show*` textual outputs. There is the possibility to further reduce code use, but I think we need to have good useage data before we work out improvements to the HTML display code. I have also not attempted to use Jupyter widgets (I can think of things we might want with models and fits, but that can come later).

Tests have been added to check that the `_repr_html_` output works and sensible, but no real tests of the SVG output (other than it exists). The assumption there is that if the plot can generate svg then it's done something useful.

This work has benefitted from previous work adding display support to other Python/NumPy style array packages (e.g. dask and xarray).

Here's the list of objects that I had identified and then those that I ended up implementing:

- [X] parameters
- [X] models
- [X] fit results
- [X] error results (e.g. `conf`, `covar`)
- [X] statistic "info" objects
- [ ] method and statistic objects
- [X] data objects
  - [X] PHA
  - [X] ARF
  - [X] RMF
  - [X] Data1D, Data1DAsummetric
  - [X] Data1DInt
  - [X] Data2D / DataIMG
- [X] plot objects
  - [X] basic display (for when no backend)
  - [X] with plots (when matplotlib is available)
  - [X] "in-between" objects like `IntervalProjection` (both a plot and a "go and do stuff to get the plot" thing)

The images shown in the discussion below do not exactly match the final output, but show various facets that have remained.


DougBurke:
The tests are failing due to #632

DougBurke:
I've run the basic sherpa notebook with this PR and added in a section showing the output to the gist at https://gist.github.com/DougBurke/82b4b0da8b716bc98502fd922872288d

There's a .html file you can download (and shows "nice" styling) and then at the bottom of the page you can find the ipynb itself (which you can view separately) which seems to render with an "old" style on gist (but shows that the information is still readable).

DougBurke:
I've just rebased and tweaked the examples a bit, so that:

- fit/error results no display the "header info" and the tabular date horizontally, rather than vertically
- some style tweaking
- ARF and PHA datasets now include a SVG plot

DougBurke:
I have also added a highly-experimental RMF plot (which shows monochromatic responses). So with

```
arf = ui.get_arf()
rmf = ui.get_rmf()
pha = ui.get_rmf()
```

we get

A) ARF

arf

B) RMF

rmf

C) PHA

pha

hamogu:
I think this is fantastic! Not many package make good use of the rich display (e.g. in astropy it's mostly tables that become html tables, similar in pandas, numpy and scipy don't have any of this as far as I can tell). I have a few thoughts on the questions that you collected above, and I'll put my thoughts here. Most of it is probably only half-backed, so take it or leave it as you see fit.

- Testing is relatively easy for the html output, see https://github.com/astropy/astropy/blob/9c8eb87f921b5d76fce66a3ef410675d37c20263/astropy/table/tests/test_pprint.py#L45 for an example.
- I don't think there needs to be any control over the output. This is a convenience to quickly see what's going on interactively. If you want to control plotting, use a plot method. 
- As much as possible, I suggest to reuse methods at the cost of not being as precise, e.g. all the html for arf, rmf, pha could probably be combined into a single methods that outputs the top 10 of a list of keywords  e.g. hduclass, number channels, obsid, etc. (if a keyword is not present it's skipped, e.g. probability threshold for pha). That would also automatically bring more consistency (e.g. why does the arf have an obsdate, but the rmf does not?)

DougBurke:
I think dask and xarray are taking more advantage than just boring tables ;-)

anetasie:
@DougBurke This looks great! When I saw this PR a long time ago I was wondering what you have had in mind. The examples look good. I would have to check it out in one of my Notebooks.

DougBurke:
@anetasie - note that the look of things will depend on when you clone this branch, as I keep on changing things!

DougBurke:
Here's examples of reading an image in:

# image just read in (logical coordinates)

img-1

# switched to physical (so axes are changed in the plot)

img-2

# apply a spatial filter (so it can zoom into the displayed region)

img-3

DougBurke:
For fun here's the current version of the RMF display. Each line shows the response to a mono-chromatic energy.

# Chandra grating

rmf-grating

# ACIS

rmf-acis

# ASCA

rmf-asca

hamogu:
This probably gets the badge for "PR with largest number of images in the comments".

DougBurke:
@hamogu there can always be moar!

hamogu:
I will also get a badge for "longest PR I've even been asked to review". Almost 2500 lines!

DougBurke:
@hamogu  There's a lot of repeated code that could perhaps be re-factored, but I'm not sure it's worth it until we get the behavior agreed on. So I think the first bit of testing is to actually just use it and see what's annoying and what works (and how tis balance changes over time).

hamogu:
This will also be very useful once we convert all threads to notebook - a project I should put higher on my priority list, I guess.

DougBurke:
@hamogu - one change I've added is to "hide" the fancy output when the notebook is not trusted. In that case you just get a HTML-ified repr() output. The reason for this is that non-trusted notebooks do not run the CSS, which means the styling can look funny. I actually think this is not needed for us, since our use if CSS is pretty basic (i.e. the output should remain readable without it), but I haven't checked for sure (and it's an easy behavior to change).

Reminder to self: need to fix escape issue (e.g. a label like "p(x)<=10"), but that doesn't necessarily play nicely with existing html-ificatino of outputs, so need to review.

hamogu:
@DougBurke I wanted to try this out, so I had to resolve the merge conflict first. That was easy, just that master removed two ``from __future__ import`` where you added a redundant ``import numpy`` (numpy was already imported 5 lines below) and git did not know how to get the order right. It means that you have to pull your own branch now before you continue any work on this, though...

hamogu:
OK, I'm using it now and in general I think it's a HUGE improvement for working in the notebook - in fact this might be an argument to move more people into using notebooks for fitting! I'll post comments as I note things and have ideas on what I like and don't like. Since I'm running through real science and not testing all bits from A to Z, my comments might appear in a random order...

Here is the first one:

For all output in table, I would like to limit the precision more (models, parameters, fit and conf results). I'm looking at a particular model parameters and the value is "0.00017225403622449424" (no joke copied to the last digit). That's not helping with readability and we are guaranteed that Sherpa will never fit anything precise enough to warrant that many digits. 

In the confidence results, I still get 6-8 significant digits, while the upper and lower bound from the confidence show me that the first digit is already uncertain. To make that more concrete, I'm looking at a fit with (sorry, I'm actually looking at the fit on a VPN in a client that does not allow me to easily screen shot) the following numbers:
```
best-fit value   lower bound   upper bound
0.0580842         -0.0497241    0.0593628
```
So, what would I want? Several ideas:

  - Use mathjax to render LateX in the html table like it might be done in a journal: `$0.05^{+0.06}_{-0.5}$`
  - round the relevant digits, so show: `0.05, -0.05, 0.06`
  - Just display fewer digits. How many? It's rare that more than a few are needed, so maybe 4? `0.05808, -0.04972, 0.05936`. Not ideal for readability when comparing many parameters in a table by eye, but better than full precision. I think @DougBurke already put some formatting on there (otherwise they all would be as long as the single parameter above). Do we ever need more than for relevant digits? If we things that's going to happen, this would need a configuration option. That gives me an even better idea: Maybe it could just honor the setting from `np.set_print_options`? No extra code needed on our part to set up and document a configuration system and that leads me to the next option:
  - Similar to how it's right now, fixed number of digits, not "fixed number of significant digits", but with the number fomr `numpy.set_print_options`. Simpler to program. 

Getting the "rounding to relevant digits" right is not trivial and would probably best be done by looking at how other packages do that and re-use some code.

*note* This is an idea from using it, not a strict requirement.

DougBurke:
Thanks. I think it really benefits from being used, rather than just evaluated.

One thing that will (eventually) be useful is knowing which representations you did test and which you didn't, which ones really didn't benefit from the HTML version, and what things that don't have _repr_html_ (or whatever it's called) but should.

hamogu:
What could be added? datastack could get an html representation. That's easy. Datatstacks already have a `show_stack` function that prints an ASCII table. That could easily be reformatted into an html table.

hamogu:
OK, I looked at models, fits, etc. in real live. I did evaluate the plots  for ARF, RMF etc, but that did not come up for me naturally. I usually work with Chandra or XMM data where I already know how they look like and what they are. So, I would not recommend to implement them, but since they are implemented already, I also see no need to remove them. 

So, my suggestion is to get this into 4.13 essentially as is (with rebase and possible minor comments). It's definitely a huge improvement for jupyter notebook users! We can tweak it later if we want to. This is all about presenting intermediate information and there is no expectation of stability (i.e. if this looks different in 4.14 than in 4.13, that's totally fine).

hamogu (approved):

DougBurke:
@hamogu - thanks for the review. 

My concerns about merging this "as is" it that it can (hopefully) be cleaned up a bit to reduce the amount of code duplication. If I don't get it clean then it just may never get fixed up!

hamogu:
When I said "as is", I mean "with the functionality it has now". Obviously, it can't be released "exactly as is", because it also needs to be a rebase. What I mean is that "perfect is the enemy of good". It would be great to have this in 4.13 and not wait just for a code cleanup.

Of course, if this can be improved, even better!

DougBurke:
In some testing I've just found an error in the RMF code. Fix (and test) incoming.

You can test this out with the following (it assumes the notebook is running from within the notebooks/ directory, so the path may need adjusting)

```
from sherpa.astro import ui

ui.load_pha('../sherpa-test-data/sherpatest/9774.pi')
ui.notice(0.5, 7)
```

and then (separate cells for each one)

```
ui.get_data()
ui.get_arf()
ui.get_rmf()
```

The first two will show a plot of the data and some metadata, the last one currently fails.

DougBurke:
I've updated the RMF bug: you now get something like

jupyter-rmf

wmclaugh (approved):

DougBurke:
Yay.

Issue #632, opened 2019-06-14
astropy 3.2 FITS I/O changes - fail to read files, warnings, and test failures


DougBurke:
Our fifth Travis test - using numpy 1.15 - has started to fail because of warnings from AstroPy. If ew compare a run today that failed, we see it is using Astropy 3.2 (which is new)

https://travis-ci.org/sherpa/sherpa/jobs/545724363#L570

whereas a recent run that passed used AstroPy 3.1.2

https://travis-ci.org/sherpa/sherpa/jobs/544274108#L570

Here is where the failures start - https://travis-ci.org/sherpa/sherpa/jobs/545724363#L5268 - and they look like (at least the first few, I haven't scanned through all of them):

```
sherpa/conftest.py:192: AssertionError
--------------------------- Captured stdout teardown ---------------------------
*** Warnings created: 1
1/1 {message : VerifyWarning('Error validating header for HDU #1 (note: Astropy uses zero-based indexing).\n    expect dimension name at position 1 in "(3),(3),(),()->(3)"\nThere may be extra bytes after the last HDU or the file is corrupted.'), category : 'VerifyWarning', filename : '/home/travis/miniconda/envs/build/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py', lineno : 1179, line : None}
```

DougBurke:
As an example of this problem. With CIAO 4.11 I get

```
sherpa In [1]: load_pha('sherpa-master/sherpa-test-data/sherpatest/source.pi')
WARNING: systematic errors were not found in file 'sherpa-master/sherpa-test-data/sherpatest/source.pi'
statistical errors were found in file 'sherpa-master/sherpa-test-data/sherpatest/source.pi' 
but not used; to use them, re-read with use_errors=True

sherpa In [2]: get_data()
Out[2]: <DataPHA data set instance 'sherpa-master/sherpa-test-data/sherpatest/source.pi'>

```

With the master branch and AstroPy 3.2 I get

```
In [1]: from sherpa.astro import ui                                                                                   
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available

In [2]: ui.load_pha('sherpa-test-data/sherpatest/source.pi')                                                          
[...]

PR #633, opened 2019-06-14
Force AstroPy version to be less than 3.2 for Travis tests

DougBurke:
This is a short-term workaround while we update to support AstroPy 3.2. See #632

### DIFF
--- a/travis/setup_conda.sh
+++ b/travis/setup_conda.sh
@@ -48,11 +48,18 @@ fi
 if [ "${DOCS}" == true ];
  then export DOCSBUILD="sphinx sphinx_rtd_theme graphviz";
 fi
+
 echo "dependencies: ${MATPLOTLIB} ${NUMPY} ${FITS} ${XSPEC} ${DOCSBUILD}"
+
+# Hack to force AstroPy < 3.2 (see https://github.com/sherpa/sherpa/issues/632)
+# This is a short-term hack just to get travis tests to pass
+# Should there be better a way of specifying versions for these dependencies?
+#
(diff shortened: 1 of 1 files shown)


travis/setup_conda.sh
```diff
@@ -48,11 +48,18 @@ fi
 if [ "${DOCS}" == true ];
  then export DOCSBUILD="sphinx sphinx_rtd_theme graphviz";
 fi
+
 echo "dependencies: ${MATPLOTLIB} ${NUMPY} ${FITS} ${XSPEC} ${DOCSBUILD}"
 
+# Hack to force AstroPy < 3.2 (see https://github.com/sherpa/sherpa/issues/632)
+# This is a short-term hack just to get travis tests to pass
+# Should there be better a way of specifying versions for these dependencies?
```
olaurino: This is fine. An alternative could be to

olaurino: This should be fine. A simple alternative might be to add the constraint to the yaml configuration, I think. In any case since this has to be fixed to support newer versions of astropy anyway, I think this fix is fine (unless Travis things otherwise for some unforeseen reason).

DougBurke: The only place I could see to put it would be the `FITS=astropy` setting in `.travis.yml`, and I was concerned that even with `FITS='astropy<3.2'` it might not work (in particular since this then gets passed to the test code). Did you have something else in mind?

Not that I think it's worth it for this case, but it might be a useful feature to have (to be able to say "this is tested against AstroPy 3.1").

olaurino:
Rather than showing as merged this PR shows as having a conflict. If I did something wrong I can't seem to understand what.

olaurino:
I guess it was something weird in Github because while I was checking I had done nothing wrong the PR was finally closed and it now shows as merged.

PR #636, opened 2019-06-27
update CI configuration

olaurino:
Anaconda changed their channels and libgfortran=3 was not available anymore. When adding the new `anaconda` channel, which contains libgfortran=3, matplotlib 1 cause a segmentation fault before even running the tests. This PR updates the CI configuration to add the `anaconda` channel as well as to update matplotlib to >=2 on all jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ matrix:
     # Barebone build to check tests pass when most of the tests must be skipped.
     # We need to use TEST=none to remove the test-data submodule.
     - env: INSTALL_TYPE=develop TEST=none NUMPYVER=1.11 TRAVIS_PYTHON_VERSION="2.7"
-    # Full build (including ds9 and xspec), Python 3.5, setup.py develop, astropy, matplotlib 1.5
-    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1 TRAVIS_PYTHON_VERSION="3.5"
+    # Full build (including ds9 and xspec), Python 3.5, setup.py develop, astropy, matplotlib 2
+    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=2 TRAVIS_PYTHON_VERSION="3.5"
       sudo: required
       dist: trusty
     # As above, Python 3.6, setup.py install, xspec 12.10
--- a/travis/setup_conda.sh
+++ b/travis/setup_conda.sh
@@ -38,6 +38,7 @@ fi
 
 conda config --add channels ${sherpa_channel}
 conda config --add channels ${xspec_channel}
+conda config --add channels anaconda
 
 # Figure out requested dependencies
 if [ -n "${MATPLOTLIBVER}" ]; then MATPLOTLIB="matplotlib=${MATPLOTLIBVER}"; fi


olaurino:
There must be something wrong in the way the new requirements are being resolved and installed, so the travis tests for xspec 12.9.1 segfault horribly: https://travis-ci.org/sherpa/sherpa/jobs/551417323#L4449.

It is not necessarily related to xspec, as pytest is failing collecting the tests to begin with, so they never get run.

I could try to figure it out, but would dropping the 12.9.1 test job be an option @DougBurke?

DougBurke:
I was going to say that I think it is okay for us to drop the XSPEC 12.9.1 test job on Travis.

Is this the only Python 2 test we are running, or is that a different job?

olaurino:
> Is this the only Python 2 test we are running, or is that a different job?

No but it is the only Python 3.5 job with all the dependencies installed.

DougBurke:
```
    # macOS build
    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER=2 XSPECVER="12.10.1b" TRAVIS_PYTHON_VERSION="3.7"
      os: osx
    # Barebone build to check tests pass when most of the tests must be skipped.
    # We need to use TEST=none to remove the test-data submodule.
    - env: INSTALL_TYPE=develop TEST=none NUMPYVER=1.11 TRAVIS_PYTHON_VERSION="2.7"
    # Full build (including ds9 and xspec), Python 3.5, setup.py develop, astropy, matplotlib 1.5
    - env: XSPECVER="12.9.1" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=develop TEST=submodule MATPLOTLIBVER=1 TRAVIS_PYTHON_VERSION="3.5"
      sudo: required
      dist: trusty
    # As above, Python 3.6, setup.py install, xspec 12.10
    - env: XSPECVER="12.10.1b" NUMPYVER="1.11" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=2 TRAVIS_PYTHON_VERSION="3.6"
      sudo: required
      dist: trusty
    # As above, python 3.7, numpy 1.15, matplotlib 3
    - env: XSPECVER="12.10.1b" NUMPYVER="1.15" FITS="astropy" INSTALL_TYPE=install TEST=submodule MATPLOTLIBVER=3 TRAVIS_PYTHON_VERSION="3.7"
      sudo: required
      dist: trusty
    # Experimental support for using Sphinx to build the documentation
    # should we not run tests with this build (to save time); i.e. have
    # a specific build just for the documentation and nothing else?
    - env: DOCS=true INSTALL_TYPE=build_sphinx TEST=none TRAVIS_PYTHON_VERSION="3.7"
      sudo: required
      dist: trusty
    # Install sherpatest package rather than relying on relative location of the submodule
    # Also, install matplotlib 2.0 and do not install astropy (checks tests are properly skipped)
    - env: INSTALL_TYPE=install TEST=package MATPLOTLIBVER=2 TRAVIS_PYTHON_VERSION="3.6"
    # Install astropy without matplotlib (checks tests are properly skipped)
    - env: INSTALL_TYPE=develop TEST=submodule FITS="astropy" TRAVIS_PYTHON_VERSION="3.5"
    # A relatively basic installation - NumPy and AstroPy - with no test data; this is similar to the preceeding
    # test, so perhaps could be combined?
    - env: INSTALL_TYPE=develop TEST=none FITS="astropy" TRAVIS_PYTHON_VERSION="3.5"
```

is that I have on my master branch; I think it's up-to-date. So we could just bump it to 12.10.1b rather than remove it. Or is the matplotlib version 1 in this build also cause problems?

olaurino:
https://travis-ci.org/sherpa/sherpa/jobs/551492492

Maybe the problem is indeed matplotlib 1.

DougBurke:
I'm happy to drop both XSPEC 12.9.1 and matplotlib version 1 from these tests.

olaurino:
xspec 12.9.1 and matplotlib 2 on python 3.5 seem to work fine. Not sure why we still hadn't matplotlib 1 jobs. At this point it's your call, we could simply bump matplotlib to 2 (unless we want to consider deeper changes at this time). We should review the test matrix at some point.

DougBurke:
When I wrote that comment I hadn't realised you'd reverted the XSPEC change. I think we should move to matplotlib 2 and XSPEC 12.10 or later, but that doesn't need to be today. If we have a combo that works (and so far e9e42ef looks promising) then use this to get Travis back to green.

PR #637, opened 2019-07-11
fixed the compiler warning for the minim.hh file

dtnguyen2:
# Release Note

A simple fix for the annoying compiler warning in the minim.hh file.

# Details:
The compiler warning:
```
sherpa/optmethods/src/minim.hh:151:10: warning: ‘I’ may be used uninitialized in this function [-Wmaybe-uninitialized]
         R[I-1] = RSQ/W;
          ^
sherpa/optmethods/src/minim.hh:98:11: note: ‘I’ was declared here
       int I, J, K;
           ^
```
should be fixed with the simple change from:
 ```int I, J, K;```
to
```int I=0, J, K;```

### DIFF
--- a/sherpa/optmethods/src/minim.hh
+++ b/sherpa/optmethods/src/minim.hh
@@ -95,7 +95,7 @@ namespace sherpa {
       const real FIVE = 5.0;
       real RSQ = 0.0;
       real W = 0.0;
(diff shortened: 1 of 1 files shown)


olaurino:
Just tested this. Yay.

Issue #342, opened 2017-03-04
'IndexError: tuple index out of range' error when using Confidence via fit.est_errors method


DougBurke:
The following sequence, fitting a simple 1D dataset with errors using chi-square statistics, fails when it comes to using Confidence to estimate the errors. It appears to be due to the two-phase approach (fit with c1 frozen then with c1 thawed). I've put details and test scripts in 
https://gist.github.com/DougBurke/2e5000ed458d261537907b569f2afc02

    create Data1D
    create Polynom1D
    fit c0 and c2 parameters
    fit c0, c1, c2 parameters
    call est_errors with Confidence set as the estmethod

This results in the error:

    IndexError: tuple index out of range

This is not a recent change, as I see the same behaviour in CIAO 4.8 (with Python 2.7) and master (with Python 3.5). It also appears not to happen using the `sherpa.ui` layer.

Note that the error case (`python fail.py 1` in the gist) reports this when evaluating the errors before it errors out

```
   mdl.c0         -9.38489    
   mdl.c1         -2.41692    
   mdl.c2         0.478273    
mdl.c0 lower bound: -2.91751
mdl.c0 upper bound: 2.91751
mdl.c2 -: WARNING: The confidence level lies within (-2.419416e+00, -2.416915e+00)
mdl.c2 lower bound: -0.00125014
mdl.c2 +: WARNING: The confidence level lies within (-2.416911e+00, -2.416915e+00)
mdl.c2 upper bound: 2.44169e-06
```

whereas the successful run reports

```
   mdl.c0         -9.38489    
   mdl.c1         -2.41692    
   mdl.c2         0.478273    
mdl.c0 lower bound: -2.91751
mdl.c0 upper bound: 2.91751
mdl.c1 lower bound: -0.250889
mdl.c2 lower bound: -0.0312677
mdl.c1 upper bound: 0.250889
mdl.c2 upper bound: 0.0312677
```

What is interesting here is that the c2 bounds are very different in the two runs: ~0.03 in the second case but 0.001 and lower in the failing case. My guess is that this is part of the problem that leads to the `IndexError` (i.e. something internal is really messed up).
```

**Comment by DougBurke** [comment]
I think the issue might be that the fit object only thinks there's two thawed parameters, even though there are three. Just before calling `est_errors` (which will fail) we have

```
In [1]: f.thaw_indices
Out[1]: (0, 2)

In [2]: [p.name for p in f.model.pars if not p.frozen]
[...]

PR #640, opened 2019-07-17
fix 'tuple index out of range' error in the fit.est_errors method

dtnguyen2:
# Release Note

A fix for issue #342.

Sherpa's ```Confidence``` class is confused if a parameter is frozen then thawed after a fit resulting with an exception of the form: ```IndexError: tuple index out of range```.  The error does not show up if conf is called from the ui level

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1046,14 +1046,9 @@ class Fit(NoNewAttributesAfterInit):
         NoNewAttributesAfterInit.__init__(self)
 
     def calc_thaw_indices(self):
-        iter = 0
-        self.thaw_indices = ()
-        for current_par in self.model.pars:
-            if current_par.frozen:
-                pass
-            else:
-                self.thaw_indices = self.thaw_indices + (iter,)
-            iter = iter + 1
+        self.thaw_indices = \
+            tuple([i for i, par in enumerate(self.model.pars) if not \
+                   par.frozen])
         
     def __setstate__(self, state):
         self.__dict__.update(state)
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1030,14 +1030,7 @@ class Fit(NoNewAttributesAfterInit):
         # that is, in case an exception is raised and
         # this parameter needs to be thawed in the
         # exception handler.
-        self.thaw_indices = ()
-        iter = 0
-        for current_par in self.model.pars:
-            if current_par.frozen:
-                pass
-            else:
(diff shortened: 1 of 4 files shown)


sherpa/fit.py
```diff
@@ -1052,6 +1045,16 @@ def __init__(self, data, model, stat=None, method=None, estmethod=None,
                                 itermethod_opts)
         NoNewAttributesAfterInit.__init__(self)
 
+    def calc_thaw_indices(self):
+        iter = 0
```
DougBurke: This routine can be simplified to

    self.thaw_indices = tuple([i for i, par in enumerate(self.model.pars) if not par.frozen])

(not tested, I should say), which I find a bit easier to follow.

DougBurke: Oh, I see I'd essentially given this as a workaround in https://github.com/sherpa/sherpa/issues/342#issuecomment-284405268

(but dropping the intermediate array in the comprehension - i.e.

    ... = tuple(i for i, ...)

dtnguyen2:
made the change as discussed on Thu

DougBurke:
I think the code is fine (for fun I ran a test run where the `calc_thaw_indices` method raised a NotImplementedError and a lot of tests failed, so it's definitely well tested), so the only things left I think are

 * changelog
 * I think we need to update the copyright date in `sherpa/fit.py` to include 2019

dtnguyen2:
Doug must have added 2019 already

DougBurke:
I think there's something a little odd in the history of this PR because whenever I switch to it it requires me to update the sherpa-test-data submodule, but there's no change to the test directory here. I tried rebasing it onto the master branch and there's a conflict with 'fit.py' (related to 85239eb I guess) but it also complains about the sherpa-test-data update - e.g. 

```
% diff --git a/sherpa-test-data b/sherpa-test-data
index 4000e17..3f92400 160000
--- a/sherpa-test-data
+++ b/sherpa-test-data
@@ -1 +1 @@
-Subproject commit 4000e170bf2e51591da6305a671af22ea4ecca26
+Subproject commit 3f924001d848187d1d21b3f8267d99e8b02c1c3c
```

(although I don't know which is from this PR and which is from the master branch in this diff output)

DougBurke:
Note to self: We also need to update the sphinx documentation - `docs/fit/index.rst` and `docs/_examples/fit/index.py` - since they use this workaround.

DougBurke:
Perhaps bad form to  add to the PR after approving it, but I've updated the sphinx documentation to remove the workaround for #342.

Issue #641, opened 2019-07-17
sherpa calc_ftest vs. Xspec ftest


dtnguyen2:
The original bug report can be found here:
https://icxc.harvard.edu/pipe/ascds_help/2019/0254.html

The reply to the bug report can be found here:
https://icxc.harvard.edu/pipe/ascds_help/2019/0255.html

# Note

I haven't looked for the specs on calc_ftest, but judging from the docs:
```
    Examples                                                                    
    --------                                                                    
                                                                                
    In this example, the simple model has 5 degrees of freedom and              
    a chi-square statistic of 7.73, while the complex model has 8               
    degrees of freedom and a chi-square statistic of 8.94. The                  
    F test does not provide any evidence that the complex model                 
    is a better fit to the data than the simple model since the                 
    result is much larger than 0.                                               

    >>> calc_ftest(5, 7.73, 8, 8.94)                                            
    0.32480691622314933
```                                                         

The current code is consistent with the docs:

```
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from sherpa.astro.ui import *
>>> calc_ftest(5, 7.73, 8, 8.94)
0.32480691622314933
```

I assume someone must have looked at the docs and approved it; Moreover, the doc has the following entries:

```
    Parameters
    ----------
    dof1 : int
       degrees of freedom of the simple model
    stat1 : number
       best-fit chi-square statistic value of the simple model
    dof2 : int
       degrees of freedom of the complex model
    stat2 : number
       best-fit chi-square statistic value of the complex model
```
which is not true since one can enter the args as arrays/lists/tuples, for example (note, I have already convert the calculations to follow XSPEC):
```
>>> calc_ftest([11, 11], [16.3, 16.3], [10, 9], [10.2, 10.5]) 
array([0.03452353, 0.13819987])
```

anetasie:
The example in the online docs does not look correct, because the more complex model will have a lower number degrees of freedom. 
Our web site provide the example with the expected input:
[http://cxc.harvard.edu/sherpa/ahelp/calc_ftest.html]

I did not know that we can use the arrays as an input to this function.

dtnguyen2:
Using the code in PR #643:
```
>>> calc_ftest(5, 7.73, 8, 8.94)
nan
>>> calc_ftest(2, 20.28, 34, 33.63)
nan
[...]

PR #643, opened 2019-07-19
modified sherpa's calc_ftest to get same results as XSPEC's ftest

dtnguyen2:
# Release Notes
Sherpa calc_ftest does not give the same results as XSPEC's ftest command; Moreover current docs is not entirely correct

# Note
This PR should be considered if sherpa's calc_ftest is to give the same results as XSPEC's ftest, otherwise reject it.  The PR is a 'fix' for issue https://github.com/sherpa/sherpa/issues/641


anetasie:
The 4.11 documentation of the required input into `calc_fstat()` is incorrect.  The first two entries for the function should be delta(dof) and delta(stat), so in the release the `calc_ftest(2, 20.28, 34, 33.63)` will return 0.000328079, while using the code updated by this PR, it will return 'nan'.  If one enters (dof_1, stat_1, dof_2, stat_2) as required by the updated code then `calc_ftest(36,53.92,34,33.64)` returns  0.000328079 the same value.

sherpa/utils/src/_utils.cc
```diff
@@ -84,13 +84,13 @@ static PyObject* ftest( PyObject* self, PyObject* args )
...
   for ( npy_intp ii = 0; ii < nelem; ii++ ) {
 
-    f = ( chisq_1[ii] / dof_1[ii] ) / ( chisq_2[ii] / dof_2[ii] );
-    x = dof_2[ii] / ( dof_2[ii] + dof_1[ii] * f );    
-    result[ii] = incbet( dof_2[ii] / 2.0, dof_1[ii] / 2.0, x);
+    const double delta_dof = dof_1[ii] - dof_2[ii];
+    const double delta_chi = chisq_1[ii] - chisq_2[ii];
+    const double f = delta_chi / delta_dof / (chisq_2[ii] / dof_2[ii]);
+    result[ii] = incbet( dof_2[ii] * 0.5 , delta_dof * 0.5,
+                         (dof_2[ii] / ( dof_2[ii] + delta_dof * f ) ) );
```
wmclaugh: Do we need to handle possible division by zero?

DougBurke:
There were two tests which had timed out in the previous commit, so I just re-started them so we can get a green checkmark.

sherpa/utils/src/_utils.cc
anetasie: `f` is based on the user input and I guess the user will not input 0 there, but who knows....

dtnguyen2: We have had a case where the _dof < 0_ (see PR #564). So delta_dof, chisq_2 and dof_2 could also be 0 and then should one check for the ranges to function incbet (see: [http://cxc.harvard.edu/sherpa/ahelp/incbet.html](url)).  The fun never stops

anetasie: So we need to add a check for 0.

dtnguyen2: Once upon a time, I had to make enhancements to some CIAO code that was full of global variables.  Unfortunately for me, I modified the code correctly, ie using **local** variables instead of the hideous **global** variables that were written throughout the code.  I was told by @wmclaugh that I have to follow convention of the code and he wants it done **incorrectly**, I kid you not!  Take for example, the following function 
```
def sao_arange(start, stop, step=None):
    """Create a range of values between start and stop.

    See also `numpy.arange` and `numpy.linspace`.

    Parameters
    ----------
    start, stop : float
       The start and stop points.
    step : float or None, optional
       If not given the step size defaults to 1.0.

    Returns
    -------
    vals : NumPy array
       The values start, start + step, ... The last point
       is the first position where start + n * step >= stop,
       which means that it can include a point > stop.

    Examples
    --------

    >>> sao_arange(1, 3)
    array([ 1.,  2.,  3.])

    >>> sao_arange(1, 3, 0.6)
    array([ 1. ,  1.6,  2.2,  2.8,  3.4])

    """

    if step is None:
        return _utils.sao_arange(start, stop)
    else:
        return _utils.sao_arange(start, stop, step)
```
defined in the same file (_sherpa/utils/__init__.py_). Logic tells me that one should check for ```start < stop``` but it is obviously not done here.  So the question now is:  "has the follow convention policy been changed?"

wmclaugh:
@dtnguyen2 - Please update the code for this PR to address potential error conditions such as division by zero.

dtnguyen2:
Made the necessary changes, but when I tried to push it was rejected:

(sherpa) [dtn@devel12 sherpa]$ git push myfork calc_fetch!=xspec_fetch 
To https://github.com/dtnguyen2/sherpa
 ! [rejected]          calc_fetch!=xspec_fetch -> calc_fetch!=xspec_fetch (fetch first)

What's a fella to do?

wmclaugh:
@dtnguyen2 - The push was likely rejected because I committed changes (f140268,  5c2d1d6 ) on 10/17 to update the copyright dates in the headers when I was reviewing the PR.  Per the error message, you probably need to pull over those changes.

dtnguyen2:
moved to PR #704

PR #647, opened 2019-07-25
Fix numpy hist warning

olaurino:
# Release Note
When plotting a PDF Sherpa now calls `numpy.histogram` with `density=True|False` rather than `normed=True|False`. There is no change visible to the user other than a warning that will not be issued anymore. `normed=True` was broken for non uniform bins, but our code always produces uniform bins, so we should never hit the problematic case.

# Notes
@anetasie @DougBurke it would be good if you could confirm the change does not produce any differences in real world cases.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -96,7 +96,7 @@ known_warnings = {
          r"numpy.ufunc size changed, may indicate binary ",
          ],
      VisibleDeprecationWarning:
-        [r"Passing `normed=True`*",
+        [
          r"sctypeNA and typeNA will be removed.*",
         ],
 }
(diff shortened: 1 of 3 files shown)


DougBurke:
This looks good to me. I have not actually run this to see if it makes a difference, but the change is simple and shouldn't cause problems...

DougBurke:
I just (quickly) tried it out and I didn't see a problem.

olaurino:
Do you mind if I piggy back this PR to remove the `typeNA` warning from out whitelist? I was going to open a new PR but it'd probably conflict with this one, and it's a one-liner.

DougBurke:
NO problem.

olaurino:
Of course, today Anaconda decided to have a bad day. Most jobs are failing at creating conda environments with 500 errors from their servers. I can reproduce the issue from my command line, although not as consistently as Travis.

DougBurke:
I'm seeing sporadic read-the-docs issues and travis builds have had some
fun recently. It must be the heat.

PR #648, opened 2019-07-25
add warning when chips is selected but cannot be imported

olaurino:
# Release Note
Sherpa now warns the user if Chips is selected as a backend but it is not available in the installation. In this case Sherpa will also try and fall back to the pylab backend.

# Notes
This fixes CIAO-283.
This was included in ciao4.12b1 and then dropped from the subsequent beta because I only committed it to the ciao-specific branch. I was supposed to issue a PR, and here it is.

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -55,13 +55,28 @@ if plot_opt == 'none_backend':
 try:
     importlib.import_module('.' + plot_opt, package='sherpa.plot')
     backend = sys.modules['sherpa.plot.' + plot_opt]
-except:
+except ImportError:
     # if the user inputs a malformed backend or it is not found,
     # give a useful warning and fall back on dummy_backend of noops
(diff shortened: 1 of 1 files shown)


DougBurke:
a) I'm assuming you don't need approval for this?

b) we really should remove the chips smoke tests and chips backend code, but that's for another PR

DougBurke:
Does this need a review by SDS or is it good to go, @wmclaugh ?

DougBurke:
@wmclaugh - do you want this rebased or will you make the needed changes?

DougBurke:
The rebased version is also at https://github.com/sherpa/sherpa/pull/691 - mainly to get a travis run of it.

wmclaugh:
@DougBurke has rebased this as #691 to resolve conflicts. PR#691 is scheduled for inclusion in 4.12.0 so this PR has been closed.

PR #649, opened 2019-07-29
comment out some time consuming tests

dtnguyen2:
# Release Note

Removed a few time consuming tests (test_SumSquares, test_Zakharov, test_Levy, test_Sphere).  The test should finish in reasonable time now

### DIFF
--- a/sherpa/optmethods/tests/test_ncoresopt.py
+++ b/sherpa/optmethods/tests/test_ncoresopt.py
@@ -137,12 +137,12 @@ class test_optmethods(SherpaTestCase):
         self.tst_func(self.ncores_nm, tstopt.Hump, x0, xmin, xmax, 0.0)
         self.tst_func(self.ncores_de, tstopt.Hump, x0, xmin, xmax, 0.0)        
         
-    def test_Levy(self):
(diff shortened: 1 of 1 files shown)


DougBurke:
Can we have a brief comment added to say something like

>>> # test commented out as they take a long time to run, see https://github.com/sherpa/sherpa/pull/649

just so that when we look back at them in 2 years we know why they aren't in use. An alternative to commenting them out is to rename them so they don't begin with 'test_', and then the test harness won't run them, but they are available if you ever want to call them manually (although how you would do that via `pytest` is beyond my ken).

dtnguyen2:
added comments as to why the tests were removed

olaurino:
I'll add this to the 4.11.1 release as that will dramatically speed up the packaging/testing pipeline. The best thing would have been to mark the tests to be skipped unless a specific option is given. I don't have time for this, but somebody might want to do that in the future.

olaurino:
Here is a pointer to the relevant documentation:
http://doc.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

PR #655, opened 2019-07-31
Allow slow tests to be run (not on by default)

DougBurke:
# Summary

The only changes in this PR are developer changes.

Switches `sherpa/optmethods/tests/test_ncoresopt.py` to pytest style from the previous unit-test style. This requires tweaking the expected result from a test (presumably because of the different ways of expressing tolerances for the approximate check).

Uncomment the slow-running tests, but mark them as slow, so that they are only run if the `--runslow` command-line argument is given. That is, the tests will not be run on Travis or for the standard `python setup.py test` run.

Unfortunately it's not obvious to me how to run the full test suite with this command-line argument. I can run this test with it (so the functionality is now available to run these tests) using

```
% python setup.py test -a 'sherpa/optmethods/tests/test_ncoresopt.py --runslow'
```

# Details

The idea is to address @olaurino comment https://github.com/sherpa/sherpa/pull/649#issuecomment-516409016

 - [x] convert the test to pytest
 - [x] add flag
 - [x] use flag to allow tests to be run

Note the pytest conversion required changing the expected minimum for the
Paviani test from -45.7 to -45.778 (which is closer to the expected value for this optimization, I believe). This will need checking.

You can now run the optional tests with

```
% python setup.py test -a 'sherpa/optmethods/tests/test_ncoresopt.py --runslow'
```

but I can't seem to pass this flag through to the whole test suite (it shouldn't really matter,
but makes like easier, and would be needed if we wanted to use this in other tests). That is,

```
% python setup.py test -a --runslow
```

fails, and I can't work out why. I have had similar issues with trying to get the -D flag (to change the test-data directory), so I wonder if there's something odd with how we have got pytest integrated into our system.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -360,10 +360,23 @@ The ``test`` command is a wrapper that calls ``pytest`` under the hood,
 and includes the ``develop`` command.
 
 You can pass additional arguments to ``pytest`` with the ``-a`` or
-``--pytest-args`` arguments.  As an example, a single test can be run
-using the syntax:
+``--pytest-args`` arguments.  As examples, the following two commands
+run all the tests in ``test_data.py`` and then a single named
+test in this file::
 
-    python setup.py test -a sherpa/astro/datastack/tests/test_datastack.py::test_load::test_case3
+    python setup.py test -a sherpa/tests/test_data.py
+    python setup.py test -a sherpa/tests/test_data.py::test_data_eval_model
+
+The full set of options, including those added by the Sherpa test
+suite - which are listed at the end of the ``custom options``
+section - can be found with::
+
+    python setup.py test -a "--pyargs sherpa --help"
+
+and to pass an argument to the Sherpa test suite (there are currently
+two options, namely ``--test-data`` and ``--runslow``)::
+
+    python setup.py test -a "--pyargs sherpa --runslow"
 
 .. note::
 
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -63,9 +63,27 @@ except ImportError:
 TEST_DATA_OPTION = "--test-data"
 
 
+# Follow https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option
+# for adding a command-line option to let slow-running tests be run
+# (not by default), which combines with existing code and options.
+#
 def pytest_addoption(parser):
     parser.addoption("-D", TEST_DATA_OPTION, action="store",
                      help="Alternative location of test data files")
+
+    parser.addoption("--runslow", action="store_true", default=False,
+                     help="run slow tests")
+
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
+        # --runslow given in cli: do not skip slow tests
+        return
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
+            item.add_marker(skip_slow)
+
 
 
 # Whilelist of known warnings. One can associate different warning messages
@@ -214,6 +232,8 @@ def pytest_configure(config):
     This configuration hook overrides the default mechanism for test data self-discovery, if the --test-data command line
     option is provided
 
+    It also adds support for the "slow" test marker
+
     Parameters
     ----------
     config standard service injected by pytest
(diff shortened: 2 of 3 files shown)


DougBurke:
I've asked for reviews from both @dtnguyen2 and @olaurino since

a) to check that the new tolerance for the Paviani test is sensible (Dan)
b) to see if there's a way to pass the `--runslow` flag to the whole test suite (Omar)

DougBurke:
For the record, here's the error I get

```
% python setup.py test -a --runslow
...
... lots of normal output
...
running build_ext
customize UnixCCompiler
customize UnixCCompiler using build_ext
customize UnixCCompiler
customize UnixCCompiler using build_ext
ERROR: usage: setup.py [options] [file_or_dir] [file_or_dir] [...]
setup.py: error: unrecognized arguments: --runslow
  inifile: /lagado.real/sherpa/sherpa-4.11.1/pytest.ini
  rootdir: /lagado.real/sherpa/sherpa-4.11.1

```

olaurino:
I believe that in order to pass the extra configuration flags you need something like:

```
python setup.py test -a "--pyargs sherpa --runslow"
```

I believe the reason is that `python setup.py test` loads `conftest.py` at a different time than with `--pyargs sherpa`. So the built-in pytest options like `-k` can be used with `python setup.py test -a`, but the custom ones need the extra `--pyargs sherpa`.

DougBurke:
Thanks - I'll try that when I get back.

DougBurke:
I can confirm that I see

```
% pythoon setup.py test
...
============ 1887 passed, 104 skipped, 24 xfailed in 237.26 seconds ============
```

and I get four extra run (and a significant increawse in run time) with

```
% python setup.py test -a "--pyargs sherpa --runslow"
...
============ 1891 passed, 100 skipped, 24 xfailed in 334.77 seconds ============
```

DougBurke:
The test failure with b86ef76 are nothing to do with this PR but to do with DS9 - see #661

DougBurke:
This PR has been rebased against the master branch to pick up the ds9 fix from #661.

dtnguyen2:
Tried to merge the changes and update on github but I ran into the following error:

```(sherpa) [dtn@devel12 sherpa]$ git checkout master```
Switched to branch 'master'
Your branch is up to date with 'sherpagithub/master'.
```(sherpa) [dtn@devel12 sherpa]$ git merge --no-ff DougBurke-allow-slow-tests-to-be-optional```
Merge made by the 'recursive' strategy.
 docs/install.rst                          |  19 +-
 sherpa/conftest.py                        |  22 ++
 sherpa/optmethods/tests/test_ncoresopt.py | 449 ++++++++++++++++++------------
 3 files changed, 304 insertions(+), 186 deletions(-)
```(sherpa) [dtn@devel12 sherpa]$ git push origin master```
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

DougBurke:
@dtnguyen2 - what does 'git remote --verbose' return?

dtnguyen2:
```(sherpa) [dtn@devel12 sherpa]$ git checkout DougBurke-change-plot-warnings```
Switched to branch 'DougBurke-change-plot-warnings'
(sherpa) [dtn@devel12 sherpa]$ git remote --verbose
myfork	https://github.com/dtnguyen2/sherpa (fetch)
myfork	https://github.com/dtnguyen2/sherpa (push)
omar	https://github.com/olaurino/sherpa.git (fetch)
omar	https://github.com/olaurino/sherpa.git (push)
sherpagithub	https://github.com/sherpa/sherpa (fetch)
sherpagithub	https://github.com/sherpa/sherpa (push)
upstream	https://github.com/sherpa/sherpa (fetch)
upstream	https://github.com/sherpa/sherpa (push)

Should I try: ```git push myfork master``` then?

DougBurke:
I would talk to @wmclaugh first, because I know Omar documented the process.

The specific problem here for you is that you don't have origin set up, which is why your original push failed with the message
`fatal: 'origin' does not appear to be a git repository`; it should have been

    git push upstream master

but you **have** to first make sure that your upstream "whatever the terminology is" was up to date before you merged in - and given that you have both `upstream` and `sherpagithub` pointing to the same place I would be careful.

DougBurke:
[I restarted the macOS test which had failed because it was due to a download error, ie not the code]

PR #658, opened 2019-08-01
Handle files with MJD_OBS or MJD-OBS keywords (datastack module)

DougBurke:
Summary
=======

The `sherpa.astro.datastack` module has been updated to deal with files that use the `MJD-OBS` keyword instead of `MJD_OBS`. This is only used by the `show_stack` routine, which prints out
several keywords from each file in the stack (that is, it is not used when fitting the data, and is only in an "informational" routine).

Chandra data has historically used the `MJD_OBS` keyword, but will be switching to using `MJD-OBS`.

Details
=======

The test suite has been updated from SherpaTestCase to pytest style, and a basic test of `show_stack` has been added. A test suite for the `sherpa.astro.io.meta.Meta` class was added, along with a `__delitem__` method for the class (this was to make it easy to edit the header of the PHA files in memory rather than on disk).

The PR could be cleaned up to merge some of the initial commits, but it is ready for review now.

### DIFF
--- a/sherpa/astro/datastack/ds.py
+++ b/sherpa/astro/datastack/ds.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 #
-# Copyright (C) 2015, 2016  Smithsonian Astrophysical Observatory
+# Copyright (C) 2015, 2016, 2019  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -19,9 +19,9 @@ from __future__ import print_function
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-import six
 import numpy
 from sherpa.utils.logging import config_logger
+from sherpa.utils.err import IOErr
 from sherpa.astro import ui
 from .utils import load_error_msg, load_wrapper, model_wrapper, \
     simple_wrapper, fit_wrapper, plot_wrapper
@@ -31,7 +31,7 @@ logger = config_logger(__name__)
 
 try:
     import stk
-except:
+except ImportError:
     logger.warning("could not import stk library. CIAO stack files and syntax will be disabled")
 
 # Global list of dataset ids in use
@@ -74,7 +74,7 @@ class DataStack(object):
         for dataid in self.dataset_ids:
             try:
                 del _all_dataset_ids[dataid]
-            except:
+            except KeyError:
                 pass
 
     @property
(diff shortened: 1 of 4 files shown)


DougBurke:
Interesting (perhaps) that the two test failures are with Python 3.7.

DougBurke:
I can at least replicate the failure with Python 3.7

```
>       assert poly1.c0.frozen is False
E       AttributeError: 'NoneType' object has no attribute 'c0'
```

DougBurke:
I've confirmed the test doesn't fail with python 3.5 or 3.6, just 3.7.

DougBurke:
For some reason `sherpa.astro.datastack.ds._all_datastack_ids` isn't being cleared out in the python 3.7 case. My suspicion is that this is some difference in how the pytest "magic" is being handled, rather than anything intrinsic to the module (and could well have been introduced in my initial refactor from UnitTest style).

DougBurke:
The code now looks for MJD-OBS, falling back to MJD_OBS if not found (which itself is optional). There's currently no test of the MJD-OBS handling.

DougBurke:
I've now added a test of the MJD-OBS handling (which ended up requiring adding tests for the Meta class...)

dtnguyen2:
When I saw @DougBurke recent sherpadev mail where he said:

> "...Do we have any documentation on what the Meta class is meant to do (it's not clear to me that it adds anything over a Python dictionary)?..."

I decided to test out his theory, so I commented out the current definition of the class Meta and wrote the following simple one:

```
class Meta(dict):
    """A key-value store."""

    def __init__(self):
        dict.__init__(self)
        return

    def pop(self, name):
        try:
            val = self[name]
        except KeyError:
            return None
        return val
```
So with the exception of the tests in the file _sherpa/astro/io/tests/test_meta.py_ all our current tests pass. So in theory: one can simplify things by redefine the ```Meta``` class (and remove the the _test_meta.py_ file).

DougBurke:
Thanks for the investigation @dtnguyen2 - I don't think we should do anything about it now (with the 4.12 release coming up), but it's another thing we can put on our rather-extensive backburner!

sherpa/astro/datastack/ds.py
```diff
@@ -118,22 +118,40 @@ def show_stack(self):
...
+                        timekey = key
+                        break
+                    except KeyError:
+                        pass
+
+            # This is informational, so do not necessarily need
+            # to report the actual key, but for now try to do so.
+            #
+            if timekey is None:
+                timekey = 'MJD-OBS'
```
wmclaugh: in line 130, timekey is set to None and then changes to MJD-OBS if the keyword is not found. Would it be better to just initialize to MJD-OBS?

DougBurke: I've made this change, and also tweaked the code to change from a hasattr check to try/except.

DougBurke:
I don't know why it thinks the checks are still pending since  https://travis-ci.org/sherpa/sherpa/builds/592597602?utm_source=github_status&utm_medium=notification claims it finished (and passed) about 3 hours ago.

PR #665, opened 2019-08-21
Remove python 2.7 support (fix #498)

### LINKED ISSUE #498: How to handle Python 2.7 support?
I think it would be useful to articulate a plan for Python 2.7 support. My suggestion would be that we keep in python 2.7 support and testing for the CIAO 4.11/Sherpa 4.11 release (as it is too dangerous to remove at this time, and we haven't announced any plans for standalone Sherpa). After release of Sherpa 4.11.0, we immediately remove python 2.7 support by

- [ ] remove testing with python 2.7
- [ ] remove 'import six' statements

This then may help with some work we'd like to do (e.g. clean up of the `BaseData` class) since we can rely on Python 3.5+ semantics.

One downside to this is that it then means that any sherpa 4.11.n (for n > 0) release would not be compatible with Python 2.7.

An alternative would be to have a "python 3" branch to work on, forward porting work to that (or however `astropy` did it), but I don't think it's worth the effort.

Thoughts and suggestions welcome.

DougBurke:
# Summary

Remove support for Python 2.7 from Sherpa. The metadata used by pip now requires python 3.5 or higher (but not version 4), and an attempt to build with an earlier version of python will cause `setup.py` to exit with an error message. The documentation has been updated, including a link to the final Python 2.7-compatible release (Sherpa 4.11.1). The code has been cleaned up to remove the use of the six module (that provided backwards-compatible support) and other parts of the code that supported running under Python 2.7 and 3.5-or-later.

There has been no attempt to re-write the code to take advantage of capabilities present in Python 3.5 that were not in Python 2.7.

# Details

Remove Python 2.7 support in Sherpa.

 - [X] Rework documentation to point out 2.7 isn't supported
    - [X] `README.md`
    - [X] Sphinx docs
 - [X] Remove python 2.7 support in `setup.py`
 - [X] Remove python 2.7 support in tests
 - [X] Remove python 2.7 support when building sphinx docs
 - [X] Remove python 2.7 travis build (replaced with a python 3.7 / numpy 1.16 build)
 - [X] remove 'import six' from all code
 - [ ] Python 2.7 clean up (not needed, but may as well do where appropriate)
      - [ ] remove fall through imports (when modules / functionality has moved)
      - [X] remove `object` base class from objects
      - [ ] can we remove / fix up the future imports now?
      - [ ] anything else?
 - [ ] update the conda build recipies

One problem with testing the six removal is that it is still required by other modules we use, so it is currently hard - if not impossible - to ensure we have a "six free" environment.

I may make some more minor clean ups to address some of the Python 2.7 clean ups, but will not touch the conda recipies.

## Questions

The metadata in `setup.py` has been updated to point out that Python 3.5 or later is needed (and does not support a hypothetical python 4). This means that 'pip' should not try to install it. However, it doesn't stop someone from trying to 'python setup.py install' a source-code installation, so I added an explicit check for that - you can see the output in https://github.com/sherpa/sherpa/pull/665#issuecomment-523559747 

Is this a sensible thing to do?


DougBurke:
The above commit - 47cf063 - should fail for the python 2.7 build on travis (but not the other builds).

DougBurke:
Apparently conda doesn't have numpy 1.17 yet - perhaps I should have checked before pushing 135d0f2! Oh well, it can be rebased out of existence.

DougBurke:
@wmclaugh - do you want me to rebase this to avoid the "update to numpy 1.17, downgrade to 1.16" step, or are we happy to keep this in?

wmclaugh:
@DougBurke - I figure it would be easier for me to just revert the
travis.yml file from my PR and merge yours as is.

DougBurke:
Sorry @wmclaugh I wasn't clear enough. This PR contains 135d0f2 and 337b26c which cancel each other out. I could rebase to remove these (and also take your copyright fix into an earlier commit), which gives a slightly-cleaner history, but do we care that much?

wmclaugh:
@DougBurke-

I realized what you meant after I sent the response. I am not overly
concerned with the history and am fine merging the PR as it stands.

PR #670, opened 2019-08-28
Link boundary of 1 parameter to another par's value (issue #362)

dtnguyen2:
# Release Note

This PR enables sherpa to link a parameter's boundary limit(s) to another parameter's value (issue #362).  One can either link a parameter limit(s) to another's parameter's value separately:
 
```
>>> from sherpa.astro.ui import *
>>> gauss1d.src1
<Gauss1D model instance 'gauss1d.src1'>
>>> gauss1d.src2
<Gauss1D model instance 'gauss1d.src2'>
>>> src2.pos.val = -1
>>> print(src1.pos)
val         = 0.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
link_min    = None
link_max    = None
default_val = 0.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38
>>> link(src1.pos, src2.pos + 2, min=src2.pos - 5)
>>> print(src1.pos)
val         = 1.0
min         = -6.0
max         = 3.4028234663852886e+38
units       = 
frozen      = True
link        = (src2.pos + 2)
link_min    = (src2.pos - 5)
link_max    = None
default_val = 1.0
default_min = -6.0
default_max = 3.4028234663852886e+38
>>> link(src1.pos, src2.pos + 2, max=src2.pos + 8)
>>> print(src1.pos)
val         = 1.0
min         = -6.0
max         = 7.0
units       = 
frozen      = True
link        = (src2.pos + 2)
link_min    = (src2.pos - 5)
link_max    = (src2.pos + 8)
default_val = 1.0
default_min = -6.0
default_max = 7.0
```
or link the boundary parameter limits to another's parameter's value together:

```
>>> gauss1d.src1
<Gauss1D model instance 'gauss1d.src1'>
>>> gauss1d.src2
<Gauss1D model instance 'gauss1d.src2'>
>>> src2.pos.val = -1
>>> print(src1.pos)
val         = 0.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
link_min    = None
link_max    = None
default_val = 0.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38
>>> link(src1.pos, src2.pos + 2, min=src2.pos - 5, max=src2.pos + 8)
>>> print(src1.pos)
val         = 1.0
min         = -6.0
max         = 7.0
units       = 
frozen      = True
[...]

PR #674, opened 2019-09-03
Add link/unlink parameter tests

dtnguyen2:
# Release Note

This PR adds three tests for sherpa's un/link commands. The tests make sure that _all_ the members of the ```Parameter``` class are correct for the low and ui  levels of sherpa's link function.

# Note

The tests are a subset of the tests for PR #670 (Link boundary of 1 parameter to another par's value).  Tests were added to make sure that by adding functionalities to the link command that the current state of the un/link command is backward compatible.

### DIFF
--- a/sherpa/models/tests/test_parameter.py
+++ b/sherpa/models/tests/test_parameter.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2016, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2016, 2018, 2019  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


dtnguyen2:
In the ```__del__``` method, if one changes the following code from:
```
            self.tst_pos(self.src1.pos, self.src1.pos.default_val,
                         min=self.src1.pos.min, max=self.src1.pos.max)
```
to
```
            self.tst_pos(self.src1.pos+1, self.src1.pos.default_val,
                         min=self.src1.pos.min, max=self.src1.pos.max)
```
Then running test will fail miserably so I know that it does indeed get called.  Anyway I have renamed the destructor to ```tst_unlink``` and call it explicitly

DougBurke:
I restarted the macOS travis test as it had timed out.

PR #677, opened 2019-09-06
XSPEC: build against 12.10.1m by removing support for models

### LINKED ISSUE #676: build issues with XSPEC 12.10.1 patch m
From https://heasarc.gsfc.nasa.gov/xanadu/xspec/issues/issues.html

```
12.10.1m

This patch updates rdblur, kdblur, kdblur2, kerrconv for considerable speed improvements 
over previous patches. It also removes the models relline, relline_lp, relline_lp_ext, relconv, 
relconv_lp, relconv_lp_ext since these are older versions to those available through 
the relxill local model package, and have been causing problems for people loading relxill. 
Report added on Aug 14, 2019.This patch modifies heasoft-6.26/Xspec/src/manager/ model.dat, 
src/XSFunctions/ Makefile, funcWrappers.cxx, .h, functionMap.cxx, .h, kerrd.cxx, spin.cxx, 
dospin.f, spinconv.cxx, rdblur.cxx, kdblur.cxx, kdblur2.cxx, src/XSUtil/ Makefile, 
src/XSUtil/Numerics/ Convolution.h.
```

The removal of models is going to cause problems for the build. We could make them conditional, but I propose that we just drop support for the models removed in this patch - they are "convolution" models so are not exposed directly to the user (i.e. there are no model classes that refer to them) and are provided for access for the "interested" users (aka me, who wants to use the new regrid support to finally enable useful support of these models), so I do not expect their removal would affect anyone. They also need dropping post 12.10.1m anyway...

**EDITED TO ADD**

looks like I may have mis-spoke when I said there were no classes providing access to this functionality

DougBurke:
Remove support for XSrelline, XSrelline_lp, and XSrelline_lp_ext models.
These were added in 12.10.1 but removed in 12.10.1m. It does not seem worth
supporting just those versions where they were available.

Remove support for low-level convolution models: tdrelconv, tdrelconvlp,
and trdelconvlpext (as with the relline case, they were added and removed
in XSPEC 12.10.1 but are not exposed to the users as model classes).

# Details

This is to fix #676


DougBurke:
I've asked @anetasie for a review to check that it's okay to remove the three models is okay. I'm not sure who out of @wmclaugh and @dtnguyen2 wants to review it for functionality :-)

DougBurke:
It looks like these were never added to `docs/model_classes/astro_xspec.rst` so don't need to be removed from it.

anetasie:
The models were removed from xspec models, as they are older version of the available functions via `relxill` local model package.

Issue #673, opened 2019-08-30
plot_source messes up calc_energy_flux/plot_fit/...probably others... for PHA data sets


DougBurke:
In the following, the original `plot_fit` call is "correct", but then a call to `plot_source` makes subsequent calls to `plot_fit` to artificially increase the model values (by a large amount). I haven't included the output images, but you can see the y range increases dramatically:

|           | minimum | maximum |
| --- | --- | --- |
| first | -1.4e-4 | 2.7e-3 |
| second | -2.6e-4 | 5.4e-2 |

```
sherpa In [1]: %matplotlib
Using matplotlib backend: TkAgg

sherpa In [2]: load_pha('sherpa-test-data/sherpatest/obs1.pi')
read ARF file sherpa-test-data/sherpatest/obs1.arf
read RMF file sherpa-test-data/sherpatest/obs1.rmf
read background file sherpa-test-data/sherpatest/obs1_bg.pi

sherpa In [3]: group_counts(5)

sherpa In [4]: set_source(powlaw1d.pl)

sherpa In [5]: pl.gamma = 1.5

sherpa In [6]: pl.ampl = 9.5e-7

sherpa In [7]: plot_fit()

sherpa In [8]: from matplotlib import pyplot as plt

sherpa In [9]: plt.ylim()
Out[9]: (-0.00013627224824930666, 0.0027496860510178647)

sherpa In [10]: plt_source()
NameError: name 'plt_source' is not defined

sherpa In [11]: plot_source()

sherpa In [12]: plot_fit()

sherpa In [13]: plt.ylim()
Out[13]: (-0.0025796359269162563, 0.054060323303023801)

```

DougBurke:
It also happens with chips.

DougBurke:
Repeated calls to

    plot_source()
    plot_fit()

increase the "amplitude" of both the source and fit plots. Is this something to do with caching (e.g. the amplitude is being applied to the cached result and then stored again): **complete** guess.

DougBurke:
See how the output of `calc_energy_flux` increases each time `plot_source` is called. It is by a factor of 100, which is 1 / bin-width-in-kev for this dataset (may be a coincidence).

```
sherpa In [1]: load_pha('/lagado.real/sherpa/sherpa-master/sherpa-test-data/sher
          ...: patest/obs1.pi')
read ARF file /lagado.real/sherpa/sherpa-master/sherpa-test-data/sherpatest/obs1.arf
[...]

PR #678, opened 2019-09-09
Cache bug ylimit (fix: deepcopy instead of shallow copy)

dtnguyen2:
# Release Note
The function ```cache_model``` makes shallow copy of the cached function values, this PR added a couple of ```.copy()`` to make sure deep copies are made instead.  This PR is a fix to issue #673 

# Note
In re-reading the code I also noticed that the setting of the ```_use_caching``` member of the class ```ArithmeticModel``` was not set consistently in the member function ```__set_state__```

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -360,10 +360,23 @@ The ``test`` command is a wrapper that calls ``pytest`` under the hood,
 and includes the ``develop`` command.
 
 You can pass additional arguments to ``pytest`` with the ``-a`` or
-``--pytest-args`` arguments.  As an example, a single test can be run
-using the syntax:
+``--pytest-args`` arguments.  As examples, the following two commands
+run all the tests in ``test_data.py`` and then a single named
+test in this file::
 
-    python setup.py test -a sherpa/astro/datastack/tests/test_datastack.py::test_load::test_case3
+    python setup.py test -a sherpa/tests/test_data.py
+    python setup.py test -a sherpa/tests/test_data.py::test_data_eval_model
+
+The full set of options, including those added by the Sherpa test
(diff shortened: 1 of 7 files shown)


DougBurke:
I would like the commit history to be cleared up before I review this. 

I think either rebasing onto a clean, up to date master branch or cherry-picking the commit onto a clean, up to date, master branch should work.

DougBurke:
> perhaps the following code
> 
> ```
>     def teardown(self):
>         self._use_caching = False
> ```
> 
> in the file _models/model.py_ should be changed to:
> 
> ```
>    def teardown(self):
>        self._use_caching = True
> ```
> 
> also

I personally don't have a good feel for how the cache code and the setup/teardown methods are meant to interact. Previously I think the cache was only used for models with no free parameters, and then only within the context of a "setup/teardown" call - e.g. as done by a fit call. This is presumably why there is a unilateral "turn off cacheing" in the teardown method, in case it had been turned on by the setup method.

 We've now changed things so that cacheing is expected to be on, so turning it off in teardown isn't very useful. However, I am not convinced that always turning it on is a great idea either. My current feeling would be that the teardown method shouldn't change the `_use_caching` attribute. In that case we may also want to remove the "turn on caching if there are no free parameters" logic in the setup method as well (since we expect the cache support to already be on), but I think this is getting to the point that we should be reviewing the cache code.

wmclaugh:
@dtnguyen2 - Can you please rebase this so that it only touches the expected file (models.py) instead of 5 files?

dtnguyen2:
It was easier to clean up by issuing a new PR #705, so will close this PR

Issue #587, opened 2019-01-27
plot_fit_resid/delchi should use more area for fit, less for the residuals


DougBurke:
At present `plot_fit_resid` and `plot_fit_delchi` use equal heights for the fit and residual plots, like

![default](https://user-images.githubusercontent.com/224638/51801645-95d28880-220e-11e9-9345-44023c53edef.png)

I think the default should be more space to the fit plot, so something like the following (but perhaps not quite so extreme)

![changed](https://user-images.githubusercontent.com/224638/51801646-9c610000-220e-11e9-8bc8-4829bedb7e9c.png)

but with a configuration setting.

DougBurke:
Closed by #682

PR #682, opened 2019-09-23
jointplot now respects the ratio argument (top plot is larger) with matplotlib backend

DougBurke:
# Summary

The matplotlib back end now makes the main (top) plot taller than the secondary (bottom) plot when using the sherpa.ui.plot_fit_resid and sherpa.ui.plot_fit_delchi routines (this also holds for the sherpa.astro.ui variants).

# Details

With this change the sherpa[.astro].ui.plot_fit_resid/delchi routines now display with the top plot being twice the height of the bottom plot when using the matplotlib (pylab) backend. The code was just not applying the requested `ratio` argument.

There is no test for this as there is currently no good test suite for the plot contents when a plotting backend is available.

This is to address #587 and collides with #626 (although easy to address).

An example of what `sherpa.astro.ui.plot_fit_delchi` looks like with this PR (examples from before the change are available in #587):

resized

The commits should be squashed into a single commit before merging, but left separated for initial review. There is no documentation on what a "joint plot" is meant to represent, so unclear how to handle the `nrows` and `ncols` arguments to `set_jointplot`.


DougBurke:
I cancelled the tests for the 4614e4a build because I realized I needed to add a change, and then one of the tests failed due to network timeouts, so it was easier to just cancel the whole build and let the 697430a tests run instead.

DougBurke:
I'm going to rebase this to fix the conflict.

docs/fit/index.rst
```diff
@@ -368,6 +368,30 @@ The residuals plot shows, on the ordinate, :math:`\sigma = (d - m) / e` where
...
 for each bin respectively.
 
+The use of this style of plot - where there's the data and fit
+in the top and a related plot (normally some form of residual about
+the fit) in the bottom - is common enough that Sherpa provides
+a specialization of
+:py:class:`~sherpa.plot.SplitPlot` called
+:py:class:`~sherpa.plot.JointPlot` for this case. In the following
+example the plots from above are re-used, as no settings have
+changed, so there is no need to pcall the
```
wmclaugh: Do you mean 'call' here?

DougBurke: I think I probably did.

PR #683, opened 2019-09-23
Remove chips from the documentation

DougBurke:
# Summary

Remove mention of ChIPS from the documentation (both on read the docs and in the sherpa.astro.ui and sherpa.ui modules), replacing with matplotlib where appropriate.

# Details

Remove mention of ChIPS from the documentation. There is no change to the code in this PR (i.e. the chips backend is still present and included in the config files).

- [X] from the sphinx build 
- [x] from the docstrings
  - [X] from the description of the `clearwindow` parameter
  - [x] review the examples to see if chips commands are being used
  - [x] the `get_data_plot_prefs` routine includes chips defaults, these should either be replaced
        with matplotlib values or just removed (the latter is "technically better" but not actually
        useful to the user)

There are also some minor un-related doc fixes that have come along for the ride. They could be pulled out into a separate PR but it didn't seem worth it at this time.

This is ready for review.

### DIFF
--- a/docs/ciao.rst
+++ b/docs/ciao.rst
@@ -13,13 +13,12 @@ from the
 `Sherpa GitHub page <https://github.com/sherpa/sherpa>`_,
 with the following modifications:
 
-* the plotting and I/O backends use the CIAO versions, namely
-  :term:`ChIPS` and :term:`Crates`, rather than
-  :term:`matplotlib` and :term:`astropy`;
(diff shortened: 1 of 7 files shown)


DougBurke:
The force push (from 4bcf91c to e3e5365) was to remove a comment (that was a note to fix something that was fixed in this commit) and to address a typo in the commit message.

Issue #431, opened 2018-02-16
iterative fit primini failing in CIAO 4.9


olaurino:
From the CXC help desk:

A helpdesk user reported an issue while fitting a spectrum with iterative fitting using the Primini method, but don't see the issue when using sigma rejection, returning a "TypeError: len() of unsized object".  This fit would work in Sherpa in CIAO 4.8.

Setting things up:

```
load_data("3c273.pi")
subtract()
set_source(xsphabs.abs1 * xsapec.cl1)
```

if I immediately do a fit, it runs, but as soon as I set:

```
sherpa-16> set_iter_method("primini")
sherpa-17> fit()
TypeError: len() of unsized object
```

and if I try backing out the iterative fitting, the problem continues to now persist, even though if I had just the default from the beginning, the fit would run.

```
sherpa-18> set_iter_method("none")
sherpa-19> fit()
TypeError: len() of unsized object
```

It seems like the problem was introduced in Sherpa 4.9.0, where I saw that there's a mention about changes to sherpa.fit.Fit (PR 287, https://github.com/sherpa/sherpa/pull/287), and it's noted:

"I added explicit tests of the sigma-rejection iterated-fit routine in this PR (to check that the changes did not change its behavior) but there are no tests of the primini iterated-fit routine."

Is there any possible workaround to this problem?

olaurino:
here is a minimal script (the dataset is taken from Nick's pool space):

```
from sherpa.astro.ui import *
load_data("3c273.pi")
set_source(powerlaw.p1)
set_iter_method("primini")
fit()
```

With the help of git bisect I found that the commit introducing the bug is 88788da52

Here is the full stack trace:

```
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    fit()
  File "<string>", line 1, in fit
  File "/export/working/sherpa/sherpa/astro/ui/utils.py", line 9262, in fit
    self._fit(id, *otherids, **kwargs)
  File "/export/working/sherpa/sherpa/astro/ui/utils.py", line 9366, in _fit
    res = f.fit(**kwargs)
  File "/export/working/sherpa/sherpa/fit.py", line 59, in run
    result = func(fit, *args, **kwargs)
  File "/export/working/sherpa/sherpa/fit.py", line 1127, in fit
    self.model.thawedparmaxes)
  File "/export/working/sherpa/sherpa/fit.py", line 915, in fit
    statargs, statkwargs)
[...]

PR #689, opened 2019-10-04
WIP: fix primini iterated fit case

DougBurke:
Darn it - this isn't getting the same values as you got from before the break.

# Summary

Allow the iterated method to be set to the `primini` method when analysing PHA data. This is to address bug #431

# Details

Thanks to @olaurino initial analysis to track down the problem. This fix "looks sensible" to me, but I don't understand why the code originally worked, or what commit 637aa33a5d78fbbb85c78f1eef1d1663e049de49 did to cause the problem.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -704,9 +704,20 @@ class IterFit(NoNewAttributesAfterInit):
         staterror_original = []
 
         for d in self.data.datasets:
+            # This loop is only concerned about user-specified
+            # statistical error, not tnat calculated by the
+            # statistic, so the staterrfunc argument is left as None
+            # in the get_staterror call.
+            #
             st = d.get_staterror(filter=False)
             staterror_original.append(st)
-            d.staterror = ones_like(st)
+
+            # could probably use the 'st is None' logic in all
+            # cases
+            if st is None:
+                d.staterror = ones_like(d.get_dep(filter=False))
+            else:
+                d.staterror = ones_like(st)
 
         # Keep record of current and previous statistics;
         # when these are within some tolerace, Primini's method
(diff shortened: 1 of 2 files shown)


DougBurke:
Hmmm, I've run what is essentially the test case against commit 2235e1c738f85a6f675cb295510b9a5d98d30714 which should be before the change that broke the primini fit, and get different results to those in the test, so need to look more. The old values also make a bit-more sense (stat value has decreased rather than increased):

```
get_staterror returns None
Expected stat= 22.392398974457006
          got= 22.392398974457006
Dataset               = 1
Iterative Fit Method  = Primini
Initial fit statistic = 22.3924
Final fit statistic   = 22.3921 at function evaluation 7
Data points           = 42
Degrees of freedom    = 40
Probability [Q-value] = 0.988943
Reduced statistic     = 0.559803
Change in statistic   = 0.000266899
   pl.gamma       1.93114     
   pl.ampl        0.000174135 

get_staterror returns None

Expected stat= 24.469394979244473
          got= 22.392132075293787

Expected gamma= 1.9047630946173293
           got= 1.931138404465634

Expected  ampl= 1.838231409413269e-4
           got= 1.7413534602287168 * 1e-4
```

DougBurke:
So, my current understanding is that I don't understand the primini implementation. I have gone back to an old commit before the break - namely 2235e1c738f85a6f675cb295510b9a5d98d30714 - and commented out the part of the primini iteration where the `staterror` value is updated each iteration, and still get the same fit results (for something that does two iterations). This suggests that the method wasn't doing what it was documented as doing.

There is also some possible inefficiencies (repeated evaluation of the model when not needed) which is related to all this, but may be a different issue.

hamogu:
I think there is a use case for iterative fitting (e.g. https://github.com/sherpa/sherpa/issues/835) and primini is one of the currently implemented ones. However, I think this does not belong into the core package. Instead, I suggest to remove the "primini" entirely and instead add a worked example or thread for "Implementing advanced / iterative fitting methods" where this is all done by the user. "Primini" could then be the example for that section.

DougBurke:
I am closing this as I think we should remove the primini iteration. Unlike Moritz's suggestion I think we just remove it, and don't even try to explain how to do this in the documentation!

Issue #663, opened 2019-08-16
color cycling in matplotlib backend


hamogu:
Using `plot_fit` if might be useful to show data and model in the same color, e.g,. in the following plot, it's not clear if the red model is supposed to go with the green or the blue data (it's a bad fit in either case, but there is no way of telling how bad). Very few people have the default matplotlib color cycle memorized to see the order. (Note that in the old chips backend the situation was even worse where everything was white by default.)
![image](https://user-images.githubusercontent.com/498688/63176675-5c2f6680-c014-11e9-855c-f085b153c826.png)

Related: #251

DougBurke:
The plot options weren't - I believe - designed with the concept of cycling through a set of values (a more-general point that colors, so we should think about that as well, but I guess for now colors is all we should be thinking about). With the matplotlib backend, the default options for data and model plots are:

```
sherpa In [9]: sherpa.astro.ui._session._dataplot.plot_prefs
Out[9]: 
{'barsabove': False,
 'capsize': None,
 'color': None,
 'ecolor': None,
 'linecolor': None,
 'linestyle': 'None',
 'marker': '.',
 'markerfacecolor': None,
 'markersize': None,
 'ratioline': False,
 'xaxis': False,
 'xerrorbars': False,
 'xlog': False,
 'yerrorbars': True,
 'ylog': False}

sherpa In [10]: sherpa.astro.ui._session._modelplot.plot_prefs
Out[10]: 
{'barsabove': False,
 'capsize': None,
 'color': None,
 'ecolor': None,
 'linecolor': None,
 'linestyle': '-',
 'marker': 'None',
 'markerfacecolor': None,
 'markersize': None,
 'ratioline': False,
 'xaxis': False,
 'xerrorbars': False,
 'xlog': False,
 'yerrorbars': False,
 'ylog': False}
```

I'm being a bit sneaky here and directly accessing the session object which underpins the "ui" layer. You can get at this with a more direct

```
>>> from sherpa.plots import DataPlot
>>> print(DataPlot().plot_prefs)
```

as I don't **think** the session class adds in any preference settings (but I could be wrong).

I think there's several things to think about (some of which overlap and some of which are separate):

 - do we want a way to explicitly say "cycle through these values" when setting a plot preference (e.g. I could say `linecolor='brown'` or `linecolor=cycle` where `cycle` is some form of a cycler class, and obviously appropriate from matplotlib to avoid re-inventing the wheel). Or do always cycle?

[...]

Issue #662, opened 2019-08-16
`plot_fit` with matplotlib backend should show model on top of data not below


hamogu:
Plotting a dataset with lots of data point with `plot_fit` makes it very hard to see the model, because for some reasons the datapoints are "on top" and the model is "below". 
![image](https://user-images.githubusercontent.com/498688/63176272-a2d09100-c013-11e9-978c-1a50901b1587.png) 
Notice how orange model is essentially invisible. I checked the zorder and tried to go through the code and I don't understand why that happens, but the plotting code is also hard to follow for me. It looks as if the model get plotted after the data, but I'm not sure because there are several steps to follow through from the backend agnostic code to the actual pylab plotting commands.

DougBurke:
I would expect the model plot to be drawn on top. However, it is drawn "per dataset", so

```
  plot_fit("foo")
  plot_fit("bar", overplot=True)
  plot_fit("bax", overplot=True)
```

should (I believe) draw, in order

  - data for foo
  - model for foo
  - data for bar
  - model for bar
  - data for bax
  - model for bax

It isn't clear to me if you want all data drawn and then all models (so that 'data for bax' does not lie on top of 'model for foo'). I can imagine this would be useful (and may require a "plot n datasets" function call, as I kind-of suggested somewhere in https://github.com/sherpa/sherpa/issues/663#issuecomment-524043316 )

The plotting is controlled by the `FitPlot` class - https://sherpa.readthedocs.io/en/latest/plots/api/sherpa.plot.FitPlot.html - and it's "plot" method is

```
    def plot(self, overplot=False, clearwindow=True):
        if self.dataplot is None or self.modelplot is None:
            raise PlotErr("nodataormodel")
        self.dataplot.plot(overplot, clearwindow)
        self.modelplot.overplot()
```

So it draws the data plot and then overplots the model. The `sherpa.plot.DataPlot.plot` method should be

```
    def plot(self, overplot=False, clearwindow=True):
        Plot.plot(self, self.x, self.y, self.yerr, self.xerr, self.title,
                  self.xlabel, self.ylabel, overplot, clearwindow)
```

and the `sherpa.plot.ModelPlot.overplot` method is a wrapper around it's plot method:

```
    def overplot(self, *args, **kwargs):
        "Add the data to an existing plot."
        kwargs['overplot'] = True
        self.plot(*args, **kwargs)

   def plot(self, overplot=False, clearwindow=True):
[...]

Issue #618, opened 2019-04-12
plot preferences and the backend


DougBurke:
It's not clear to me what the intended semantics of the "plot preferences" dictionaries are: are the keys and values meant to be backend independent or not.

Advantages for them being independent include

 - if you switch backend your code still works and you should still get reasonable plots

Disadvantages include

  - I believe that it's hard to come up with truly backend-agnostic settings, both for the key names and the values, so is it worth the effort coming up with a "data model" for them

From work by @kglotfelty , it looks like we may have a mixture of both - e.g. `xerrorbars` and `ylog` are "agnostic" settings (since we can assume we are going to require backends that support both), but `markerfacecolor` is backend specific, and it's not clear what settings for `linestyle` are suported (since chips and matplotlib use different syntaxes):

```
% env SHERPARC=$HOME/.sherpa.rc python -c 'import sherpa.astro.ui as ui;print(ui.get_data_plot_prefs())'

{'xlog': False, 'xaxis': False, 'color': None, 'linecolor': None, 'marker': '.', 'linestyle': 'None', 'xerrorbars': False, 'ecolor': None, 'barsabove': False, 'markerfacecolor': None, 'ratioline': False, 'capsize': None, 'ylog': False, 'markersize': None, 'yerrorbars': True}


% env SHERPARC=$HOME/.sherpa.chips.rc python -c 'import sherpa.astro.ui as ui;print(ui.get_data_plot_prefs())'

{'linestyle': 0, 'linethickness': None, 'yerrorbars': True, 'xlog': False, 'symbolcolor': None, 'symbolfill': False, 'errcolor': None, 'linecolor': None, 'ylog': False, 'errstyle': 'line', 'ratioline': False, 'errthickness': None, 'symbolsize': 3, 'xaxis': False, 'symbolstyle': 4, 'xerrorbars': False}
```

DougBurke:
This is related to #663 and parts of #662

DougBurke:
I think we are going to have a mix of values that we expect all backends to provide,and some that are backend specific.

DougBurke:
I've re-opened this to note that I strongly believe we have a "mixed" model:

- several fixed options, including `overplot` and `clearwindow`, but also `xlog`, `ylog`, `xerrorbars`, and `yerrorbars` (where appropriate)
- backend -specific, such as `ecolor` and `marker`

We could try to identify sensible options to promote to the "fixed" set (e.g. line or symbol color and style), but I'm not sure it's worth it.

[...]

Issue #251, opened 2016-08-26
Allow plotting options in Sherpa commands


hamogu:
I often find myself in the following situation:
I make a script in Sherpa that outputs some complicated plot (say different datasets and different models plotted on top of each other).

My main problem is that I always have to run "info()" to find out how lines are called and then change them retroactively with

```
set_curve(...)
```

or

```
set_histogram('hist2', ['*.color', 'blue'])
```

This means that my script is littered with hardcodeds string like "crv3"  or "hist1".
That's a) annoying and b) a problem if I e.g. decide to remove the plot statement that generated "crv1". Next time I run the script "crv2" will be called "crv1" and all colors etc. will be off.

It would be great if there was a way to specify properties like color, symbol etc. when making chips plots from Sherpa, e.g. something like:

```
plot_model(2, overplot=True, "color=blue")
```

Matplotlib solves that by allowing my to either set the color in the plot statement "plt.plot(... color='r')", or by returning an object I  can refer to 

   l1 = plt.plot(...)
   l1.something = "blue"

I talked to @DougBurke and he pointed out that Chips supporst different plotting backends. So, here is my suggestion how this can be done relatively easily: Chandge the signature of `plot_data`,  `plot_fit` etc. to

```
plot_data([id], **kwargs)
```

and just pass kwargs to the plotting backend.
Alternatively, accept kwargs called "chips" and "matplotlib", where only
the keyword names like the active backend has any effect. The value of these keywords need to be dictionaries in either chip or mpl style (and can be documented with just "see ahelp chipsopt or matplotlib documentation").

DougBurke:
Just to note that this is with the chips backend - where `set_curve/histogram` are functions from ChIPS - but it is similar with the matplotlib backend, i.e. there's no easy way to override the defaults for a given plot. The current approach would be something like

    get_data_plot_prefs()['linecolor'] = 'green'
    plot_data()
    get_data_plot_prefs()['linecolor'] = <whatever the old value was>

which is not ideal.

We don't really document well out mapping from preferences (e.g. linecolor) to backend-specific option, so we need to think about whether to try and make this generic, or just pass through values for the backend to use directly (which makes the code less portable, but I'm not sure it's worth keeping the API backend-agnostic for "power" users who are unlikely to change backend; at least not until someone writes us an altair/xxx backend ;-)

DougBurke:
[...]

PR #692, opened 2019-10-10
Allow users to override plot preferences when creating a plot

DougBurke:
# Summary

Allow plot preferences to be over-ridden when a plot is created. The various plot and contour calls - for both the object and ui APIs - now have a `**kwargs` argument which gets passed to the underlying plot object. These arguments are merged with the preference values before being passed to the plot backend code.  This allows users of the ui layer to say (assuming the Matplotlib backend)

    ui.plot_data(ylog=True)
    ui.plot_model(marker='s', linestyle='dotted', overplot=True)

and for users of the object API (assuming dplot and mplot are DataPlot and ModelPlot objects that have been set up):

    dplot.plot(ylog=True)
    mplot.plot(marker='s', linestyle='dotted', overplot=True)

This provides minimal customization - namely that provided by the backend's plot, histogram, and contour calls - but should still prove to be useful. More advanced changes are left to the user calling the backend directly. Changes to composite plots - namely ones like `plot_fit` and `plot_fit_resid` - are 
very limited since the plot preferences do not support identifying the components of the plot. This means that you can say

    plot_fit_resid(capsize=5, color='gray', ylog=True)

but this will be applied to the data and model (top plot) and residuals (bottom plot). That is, you can not use this interface just to change the model component of the plot. Users will have to use the existing preferences mechanism - in this case by changing the dictionary returned by get_model_plot_prefs() - if they want more precise changes for the composite case.

Documentation - both docstrings and the Sphinx documentation - have been updated to reflect this change. This includes a significant update of the visualization section of the sherpa web site, since it was previously just a set of example calls with no documentation to tie it together. There have also been updates to the Sphinx documentation to reflect recent changes from other accepted PRs.

# Details

This change was easier than expected, because it ended up working with the way the preference settings are created, so it just requires a merge of the preference and user settings. This merge is biased to select the user settings (i.e. the user choice overrides the preference setting).

At present the merge does not try to restrict the user arguments to those defined by the preferences, but perhaps it should, either to ignore unknown keywords, or to error out (which is actually the current result, but an explicit change might lead to more-readable error messages). Note that the use of multiple inheritance in the plot structures - in particular the source- and model- component plots - causes some potential confusion over which preference setting to use (there's a mixture of "line" and "histogram" style preferences), which requires an explicit choice to be made (by over-riding the `_merge_settings` method added in this PR to use a different preference setting).

This partly addresses #663, some of the discussion in #662, #618, #251 and possibly other tickets.

This PR has seen explicit user testing (but still limited) by @DougBurke and relies in part of the recently-enhanced plot test suite in Sherpa (which found several issues during development).

Notes / todo (items are marked off as they are addressed):

 - [x] the astro-specific plots have not been changed to support this
 - [X] there is no user-level documentation
 - [x] the `plot_model_component` and `plot_source_component` calls are known to be broken,
   and there are probably more (the chips unit test fails)
 - [x] need to test it out interactively to see how it works for composite plots like `plot_fit` and `plot_fit_delchi`.
 - [X] the changes also apply to the `contour_` calls; I have done manual testing of these but
         do not want to add a whole new set of tests for the very-limited support in this PR
 - [ ] the "reg/int" style commands can support this, but have not been fully updated to do so
       (that is the ui interface does not expose this functionality); I think this is okay to leave out at this stage.

The Sphinx-related changes also refer, in part, to changes made in recently-accepted PRs.


DougBurke:
Note to self: this is based on #667 so the rebase will be fun.

anetasie:
Is this PR going to be merged?

wmclaugh:
The PR was reviewed and approved yesterday. The plan is to merge it into
master to get it into the next set of CIAOD builds.

-Warren

anetasie:
Great! Thanks!

DougBurke:
It is going to need to be rebased, and I'd much prefer if #667 (order of
source and data in plot_fit) gets merged first (as that will make things
easier for me).

Doug

wmclaugh:
@DougBurke 

I actually was referring to #667 which was the PR that was reviewed yesterday. #667 has been merged and the travis CI tests are running right now- so that should help for your rebase.

DougBurke:
I've rebased this PR to account for the conflicts, so it's good for a final review now.

docs/examples/simple_interpolation.rst
```diff
@@ -49,12 +49,21 @@ which can be "loaded" into Sherpa using the
...
-   
+
+.. note::
+
+   Creating the :py:class:`~sherpa.data.Data1D` object will - from
+   Sherpa 4.12.0 - warn you that it is converting the dependent axis
+   data to NumPy. That is, the above call will cause the following
+   message to be displayed::
+
+        UserWarning: Converting array [1, 1.5, 1.75, 3.25, 6, 16] to numpy array
```
hamogu: Unless I miss something here it seems that the changes in line 52 and 53 will change this to be just `[6]`. I don't understand the changes in 52 and 53 and the fact that it's not consistent with this new doc note indicates that maybe this line is correct and the 52/53 are not?

docs/examples/simple_user_model.rst
```diff
@@ -99,8 +99,8 @@ data points)::
 
    >>> print(d)
    name      = NaNO_3
-   x         = [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
-   y         = [1, 4.125, 7.5, 10.0417, 12.25, 13.75, 14.0833, 14.7917, 15.5, 14.6667, 14.875]
+   x         = Int64[11]
+   y         = Float64[11]
```
hamogu: Does this no cause inconsistency with the line below if you only set a single value?

docs/examples/simple_interpolation.rst
DougBurke: I don't understand what you mean - but I am too close to this to read it properly. I am just trying to note that the warning we see in the example below is new:

```
>>> d = Data1D('test', [1, 1.5, 2, 4, 8, 17], [1, 1.5, 1.75, 3.25, 6, 16])
/lagado.real/sherpa/sherpa-master/sherpa/data.py:49: UserWarning: Converting array [1, 1.5, 1.75, 3.25, 6, 16] to numpy array
  warnings.warn("Converting array {} to numpy array".format(array))
>>> print(d)
name      = test
x         = Float64[6]
y         = Float64[6]
staterror = None
syserror  = None
>>> 
```

docs/examples/simple_user_model.rst
DougBurke: I don't understand the question. This is just a copy of the screen output of the current behavior. Why the staterror isn't converted to a NumPy array but x and y are is a question for a different time/issue.

docs/examples/simple_interpolation.rst
hamogu: Ah, now I get it. I thought `Float64[6]` means a single float with value 6, but it is in fact six floats in a numpy array. Confusing, but that's not the point of this PR, so consider this solved for now.

docs/examples/simple_user_model.rst
hamogu: Same as above.

Issue #654, opened 2019-07-30
new matplotlib warnings (3.1) about drawstyle/linestyle


DougBurke:
testing out the `plot_pdf` change in #647 I noticed the following warning

```
In [11]: ui.plot_fit()                                                          
WARNING: The displayed errorbars have been supplied with the data or calculated using chi2xspecvar; the errors are not used in fits with wstat

In [12]: ui.plot_pdf(ds[0])                                                     
/home/dburke/sherpa/sherpa-647/sherpa/plot/pylab_backend.py:251: MatplotlibDeprecationWarning: Passing the drawstyle with the linestyle as a single string is deprecated since Matplotlib 3.1 and support will be removed in 3.3; please pass the drawstyle separately using the drawstyle keyword argument to Line2D or set_drawstyle() method (or ds/set_ds()).
  getattr(line, 'set_' + var)(val)

```

We'll have to look at this.

```
% conda list matplotlib
# packages in environment at /home/dburke/anaconda3/envs/sherpa-647:
#
# Name                    Version                   Build  Channel
matplotlib                3.1.0            py37h5429711_0  
```

olaurino:
Note that this is already whitelisted, but it's good to have a ticket open because we need to fix the underlying issue. So thanks for the report.

PR #693, opened 2019-10-12
Address matplotlib linestyle removal in Matplotlib version 3.3 

DougBurke:
# Summary

Support matplotlib version 3.3 (fixes issue #654).

# Details

Matplotlib version 3.3 now requires you to split out the drawstyle and linestyle arguments. This set of PRs changes the code so that the two arguments are now set: linestyle defaults to `solid` and drawstyle to `default` for plots and `step-mid` for histograms. This appears to replicate the old version, and should be backwards compatible (drawstyle is supported in Matplotlib 2.2.5 and 1.5.3).

Earlier versions of this PR had set the drawstyle to plots as well as histograms, but that had unintended consequences (and you can see some of the examples in the discussion below, which are now out of date).

## Examples

Here are the examples of `plot_fit_resid` for DataPHA, Data1D, and Data1DInt after applying this PR (they look very similar to the before-PR version):

example_pha1

example_1d

example_1dint


DougBurke:
One issue with this PR is that it changes `plot_fit` (and `plot_fit_resid/ratio/delchi) for PHA data from

before

to

after

Note how the model is now drawn as a histogram, which is technically corrrect, but the bin edges don't look right (and depending on the `drawstyle` argument can be changed).

This change may be okay, but I wanted to mention it.

You can get back the "pre fix" behavior with something like

    get_model_plot_prefs()['drawstyle'] = 'default'

but I'm reluctant to change the default setting as it needs more testing (e.g. with Data1DInt data  plots0

anetasie:
I like the histogram model plots for the PHA data sets, as they are more informative - you can see the size of the bin being used in the fit. The update default option is good and as @DougBurke said we need to test other plots before making this a default plot option.

DougBurke:
I rebased this onto master (so I could take advantage of the python 3.8 support) and now the macOS travis test failed with some error unrelated to this PR - namely

```
checking whether build environment is sane... configure: error: newly created file is older than distributed files!
Check your system clock
```

I no-longer seem to be able to restart an individual build, so I'm just going to leave as is for now, as this PR isn't ready for final review yet.

DougBurke:
Hmmm, some not great things about this. This is with matplotlib 3.1.3, and the data/models are not necessarily meant to be good fits

a) Data1D case (ie not binned) - the model/source still appears binned/histogram

oned

b) Data1DInt

Here the bins are x=1-2, 2-4, 4-6, 6-10, 10-15, 15-20, 20-22, 22-23 so you can see the histogram doesn't show the bin edges

int1d

c) DataPHA case

This is essentially the same as the Data1DInt case, it's just that you can't easily find out the expected bin edges. But you can see that the bin edges are at the same values as the data points.

I'm not sure why this looks different to the earlier example I gave. I may have changed a preference setting somewhere.

DougBurke:
I've rebased to pick up the new testing fixes and updates. Hopefully that will fix the failing test.

DougBurke:
One **SERIOUS** problem is that it turns Data1D datasets into "histogram-style" plots: e.g.

```
In [1]: from sherpa import ui

In [2]: ui.load_arrays("1d", [1, 2, 4, 6, 7, 8], [2, 1, 2, 3, 2, 1], ui.Data1D)

In [3]: mdl = ui.polynom1d.mdl

In [4]: mdl.c0 = -1

In [5]: mdl.c1 = 1

In [6]: ui.set_source('1d', mdl)

In [7]: ui.plot_fit('1d')
```

fig

PR #694, opened 2019-10-15
WIP Can we plot data with wstat statistic sensibly?

DougBurke:
The aim is to make the output of `plot_fit` be meaningful when `wstat` is used: at present the contribution from the background is not included, so the model appears to under-predict the data.

This PR tests out changes made along the way (we are not yet there yet). It is also based on PR #692 rather than master.

Implemented

 - [X] change internals to allow passing through keyword arguments to the "prepare" method in
       the ui layer rather than using positional arguments
 - [ ] support a way of rescaling backgroud data to match the source aperture
 - [ ] add background contribution to model values or background subtract the data when using wstat


DougBurke:
I've rebased this to account for recent changes to the master branch. Something looks to have changed with the system behavior in the "`Add the rescale arguemnt to plot_bkg/get_bkg_plot`" commit (currently 494d591) since 8 tests added in this PR now fail (when they didn't before). So this needs looking at.

DougBurke:
I've just rebased to remove two "unnescessary" commits, which have been moved to #709 (residual-style plots ignore the ylog setting). The test failures have not been addressed in this rebase.

afruscione:
@DougBurke I get this error when I am building the branch:
```
(sherpatest) antonella@antonellas-mbp sherpa % python setup.py develop
/Users/antonella/git/sherpa/setup.py:45: DeprecationWarning: 

  `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
  of the deprecation of `distutils` itself. It will be removed for
  Python >= 3.12. For older Python versions it will remain present.
  It is recommended to use `setuptools < 60.0` for those Python versions.
  For more details, see:
    https://numpy.org/devdocs/reference/distutils_status_migration.html 


  from numpy.distutils import core
  File "/Users/antonella/git/sherpa/setup.py", line 57, in <module>
  File "/Users/antonella/git/sherpa/helpers/__init__.py", line 25, in <module>
    from .sherpa_config import sherpa_config
  File "/Users/antonella/git/sherpa/helpers/sherpa_config.py", line 25, in <module>
    from numpy.distutils.core import Command
ImportError: cannot import name 'Command' from 'numpy.distutils.core' (/Users/antonella/opt/anaconda3/envs/sherpatest/lib/python3.9/site-packages/numpy/distutils/core.py)

DougBurke:
@afruscione - I haven't looked at this for a year or two now as there's been no interest in it. Given changes to the python build system - e.g instead of

    python setup.py install
    python setup.py develop

we now say

    pip install .
    pip install -e .

I think this needs to be rebased to pick up changes recently made to Sherpa. Unless you really want to try this out I'm unlikely to make the needed changes soon.

PR #696, opened 2019-10-20
Support python 3.8

DougBurke:
# Summary

Support python 3.8. This only tests the basics of Sherpa, and not the plotting (matplotlib) or I/O (enabled by AstroPy) functionality. 

# Details

Switch the "basic" test on Travis to use Python 3.8 rather than 3.7.

Address the failures seen in #730 

- [X] remove syntax warnings from use of `is` with a literal
- [X] remove deprecation warnings about PY_SSIZE_T_CLEAN
- [X] fix index error (use scalar indexes)
- [X] update the documentation to highlight initial python 3.8 support

The first two fixes are simple to understand and implement, as well as they should be backwards compatible back to Python 3.5.

The final issue (index error) is also a simple fix, which I believe to be related to the [Python 3.8 change](https://docs.python.org/3/whatsnew/3.8.html#build-and-c-api-changes):

```
Functions that convert Python number to C integer like PyLong_AsLong() and argument parsing functions like PyArg_ParseTuple() with integer converting format units like 'i' will now use the __index__() special method instead of __int__(), if available. The deprecation warning will be emitted for objects with the __int__() method but without the __index__() method (like Decimal and Fraction). PyNumber_Check() will now return 1 for objects implementing __index__(). PyNumber_Long(), PyNumber_Float() and PyFloat_AsDouble() also now use the __index__() method if available. (Contributed by Serhiy Storchaka in bpo-36048 and bpo-20092.)
```

The code was using a single-element array with a NumPy int64 element in it, and this doesn't have `__index__` defined. The fix is to explicitly convert to a Python int before passing to the compiled code (the error was coming from the `PyArg_ParseTuple` call in `sherpa/utils/src/_psf.cc::set_origin`). There may be a chance for conversion error or overflow, but that would take quite a large array (and if it does exist, is likely present in the original code).

I have also made a change suggested by @serhiy-storchaka in https://github.com/sherpa/sherpa/pull/696#issuecomment-578400676 - replacing `PyErr_Format(etype, msg)` by `PyErr_SetString(etype, msg)` - as it is slightly clearer on intent (but should have no functional change). This was only done for the optmethods test code, so is not in user-facing code.

I have tried with AstroPy and matplotlib installed (via pip), ds9+xpa (conda installed from the CXC channel), and against XSPEC 12.10.1b, with Python 3.8 and the tests pass.

### DIFF
--- a/sherpa/instrument.py
+++ b/sherpa/instrument.py
@@ -1,6 +1,7 @@
 from __future__ import division
 #
-#  Copyright (C) 2008, 2016, 2018, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2008, 2016, 2018, 2019, 2020
+#        Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -296,14 +297,20 @@ class PSFKernel(Kernel):
         # be centered on brightest pixel.
         brightPixel = list(numpy.where(kernel == kernel.max())).pop()
 
-        origin = None
         # if more than one pixel qualifies as brightest, such as const2D
         # use the middle of subkernel -- assumes the user provided center at
         # time of kernel extraction, so that should be middle of subkernel.
+        origin = None
         if (not numpy.isscalar(brightPixel)) and len(brightPixel) != 1:
             origin = set_origin(kshape)
         else:
-            origin = set_origin(kshape, brightPixel)
+            # brightPixel is a NumPy index (int64) which - as of NumPy 1.18
+            # and Python 3.8 - causes a TypeError with the message
+            # "only integer scalar arrays can be converted to a scalar index"
+            # to be thrown here if sent directly to set_origin. So
+            # we convert to a Python integer type.
+            #
+            origin = set_origin(kshape, int(brightPixel))
 
         if self.origin is None:
             self.origin = origin
--- a/sherpa/optmethods/tests/_tstoptfct.cc
+++ b/sherpa/optmethods/tests/_tstoptfct.cc
@@ -37,8 +37,7 @@ static PyObject *bard( PyObject *self, PyObject *args ) {
   npy_intp npar = xpar.get_size( );
   npy_intp mfct = 15 * npar / 3;
   if ( EXIT_SUCCESS != fvec.create( 1, &mfct ) ) {
-    PyErr_Format( PyExc_ValueError,
-		  static_cast<const char*>( "Unable to create 'fvec'" ) );
+    PyErr_SetString( PyExc_ValueError, "Unable to create 'fvec'" );
     return NULL;
   }
   double fval;
@@ -48,16 +47,14 @@ static PyObject *bard( PyObject *self, PyObject *args ) {
     tstoptfct::Bard<double,void*>( mfct, npar, &xpar[0], &fvec[0], ierr,
 				   NULL );  
     if ( EXIT_SUCCESS != ierr ) {
-      PyErr_Format( PyExc_ValueError,
-		    static_cast<const char*>( "error returned for bard function" ) );
+      PyErr_SetString( PyExc_ValueError, "error returned for bard function" );
       return NULL;
     }
   }
   {
     tstoptfct::Bard<double,void*>( npar, &xpar[0], fval, ierr, NULL );
     if ( EXIT_SUCCESS != ierr ) {
-      PyErr_Format( PyExc_ValueError,
-		    static_cast<const char*>( "error returned for Bard Fct function" ) );
+      PyErr_SetString( PyExc_ValueError, "error returned for Bard Fct function" );
       return NULL;
     }
   }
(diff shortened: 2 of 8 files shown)


DougBurke:
We can track https://github.com/gammapy/gammapy/issues/2466 to see when python 3.8 is easily available for testing, as @cdeil seems on the case ;-)

cdeil:
I think the relevant issue to watch is https://github.com/conda-forge/python-feedstock/pull/274 and for Anaconda their blog at https://www.anaconda.com/category/news-blog/ (not sure if there's a way to get notifications for that).

serhiy-storchaka:
It is not related to this PR, but why in code like
```c++
    PyErr_Format( PyExc_ValueError,
		  static_cast<const char*>( "Unable to create 'hi'" ) );
```
you use `static_cast<const char*>`? Does not `"Unable to create 'hi'"` have type `const char[]` automatically coerced to `const char*`?

Also, you can use `PyErr_SetString` instead of `PyErr_Format`.

sherpa/optmethods/tests/_tstoptfct.cc
```diff
@@ -1344,7 +1348,8 @@ static PyObject *wood( PyObject *self, PyObject *args ) {
 
 static PyObject *init_optfcn( PyObject *self, PyObject *args ) {
 
-  int name_length, npar;
+  int npar;
+  Py_ssize_t name_length;
   char* name;
```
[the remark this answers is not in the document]
DougBurke: Thanks for the comments. This is not my code, so I can't comment on its current set up. I will note that it's "test" code, so I think we are perhaps more forgiving here than we'd want to be in other parts of the code base ;-)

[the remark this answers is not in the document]
DougBurke: > Side note: it is better make it `const`. It refers to immutable internal data in the Python string object.

Are you talking about making `name_length` or `npar` (or both) `const` here? If so, I don't see how to do that since we send them to `PyArg_ParseTuple` to be set. Or did you mean something else?

taldcroft:
So I can confirm that sherpa builds OK on my mac catalina with Py=3.8 and latest numpy / astropy, and allows unit tests to pass for one application we have that includes sherpa fitting in tests.  Yay.  

But `sherpa_test` showed some fails:
```
../../miniconda3/envs/sherpa38/lib/python3.8/site-packages/sherpa/estmethods/tests/test_conf.py FFFF                                                                      [ 22%]
../../miniconda3/envs/sherpa38/lib/python3.8/site-packages/sherpa/estmethods/tests/test_estmethods.py ..FF                                                                [ 22%]
../../miniconda3/envs/sherpa38/lib/python3.8/site-packages/sherpa/optmethods/tests/test_ncoresopt.py FFFFFFFFFFFFF.FFFssss                                                [ 30%]
../../miniconda3/envs/sherpa38/lib/python3.8/site-packages/sherpa/ui/tests/test_session.py .......F                                                                       [ 87%]
```
Another application segfaulted, which is not a good sign, but it is a Qt gui app that is only tested thus far on Py 3.6 with very old versions of everything, so not obvious that this is in sherpa.

DougBurke:
@taldcroft thanks for running the tests. Does the `sherpa_test` output provide any info on what went wrong?

taldcroft:
Indeed there are quite-lengthy stack traces.  I'll get them to you one way or another.

taldcroft:
Sent by email.

DougBurke:
I think it would be worth accepting these changes, even if there are other problems (but I am a bit biased)

taldcroft:
> I think it would be worth accepting these changes, even if there are other problems (but I am a bit biased)

Not that I'm a voting member, but I had exactly the same thought yesterday.  A PR merge to master does not have to fix every problem, just the identified problems.  One might just change the title to "Improve support for Python 3.8" to be pedantic.

taldcroft:
And of course there was the practical driver which is that it is slightly easier to pull from and test master than a personal branch.

sherpa/instrument.py
```diff
@@ -296,14 +297,20 @@ def init_kernel(self, kernel):
         # be centered on brightest pixel.
         brightPixel = list(numpy.where(kernel == kernel.max())).pop()
 
-        origin = None
         # if more than one pixel qualifies as brightest, such as const2D
         # use the middle of subkernel -- assumes the user provided center at
         # time of kernel extraction, so that should be middle of subkernel.
+        origin = None
```
wmclaugh: Not necessary for this PR, but a general question:  is there any impact in running set_origin?  Are we doing several instructions that are only applicable if self.origin is set to None? Would it be more efficient for instance to do something like:

    if self.origin is None:
        # if more than one pixel qualifies as brightest, such as const2D
        # use the middle of subkernel -- assumes the user provided center at
        # time of kernel extraction, so that should be middle of subkernel.
        if (not numpy.isscalar(brightPixel)) and len(brightPixel) != 1:
            self.origin = set_origin(kshape)
        else:
            self.origin = set_origin(kshape, brightPixel)

DougBurke: I don't have a good understanding of what's going on with the convolution support, so I don't know the answer to this.

dtnguyen2: understanding what 'center' and 'origin' do is on the first item of my bucket list

taldcroft:
Have you played around with `set_start_method('fork')`?  This worked for me on Py 3.8 on Mac.

```
import multiprocessing as mp

def foo(q):
    q.put('hello')

if __name__ == '__main__':
    mp.set_start_method('fork')
    q = mp.Queue()
    p = mp.Process(target=foo, args=(q,))
    p.start()
    print(q.get())
    p.join()
```

Issue #695, opened 2019-10-18
Initializing generid data silently drops mask


hamogu:
When I make a `Data1D` or `Data2D` dataset from a numpy masked array, Sherpa silently drops the mask and uses the values "behind the mask".
```
import numpy as np

from sherpa.data import Data1D, Data2D
from sherpa.fit import Fit
from sherpa.models import Const1D, Const2D

arr1d = np.ma.array([1.,2.,90.,3.,1.], mask=[False, False, True, False, False])
arr2d = np.ma.array([[1,1.5], [10., 1.2]], mask = [[False, False], [True, False]])

d1 = Data1D('d1', np.array([1,2,3,4,5]), arr1d)
d2 = Data2D('d2', np.array([1,1,2,2]), np.array([1,2,1,2]), arr2d.ravel())
print(d1.mask)

c1 = Const1D('c1')
fit1 = Fit(d1, c1)
print(fit1.fit())
```
The Fit results is (some non-relevant lines removed from output):
```
succeeded      = True
parnames       = ('c1.c0',)
parvals        = (2.8467372929128665,)
statval        = 69.91988835701966
numpoints      = 5
dof            = 4
qval           = 2.359979141585598e-14
rstat          = 17.479972089254915
```
Clearly, the fit is terrible (just look at qval) and the fitted values is too large. `numpoints` shows that the fit used all five datapoints, including the one that should have been masked.
However, I can manually set the mask (**note that I actually need to set the opposite of the mask** because numpy uses `True` to indicate masked values and Sherpa uses `True~ to indicate valid values and `False~ for masked values):
```
d1.mask = ~arr1d.mask
fit1 = Fit(d1, c1)
print(fit1.fit())
```
Now, the fit works better and uses only 4 unmasked datapoints:
```
succeeded      = True
parnames       = ('c1.c0',)
parvals        = (1.5945294293559473,)
statval        = 0.3833606897991587
numpoints      = 4
dof            = 3
qval           = 0.9436579269604303
rstat          = 0.12778689659971956
```

The same works for 2D datasets...
I think its very dangerous to accept masked arrays on input and silently drop the mask. That's even more confusing since there is no obvious way to pass in a mask. I had to discover the `data.mask` attribute and then figure out that the definition is exactly opposite to numpy. Very unfortunate choice!

So, what could be done? I suggest:

[...]

Issue #346, opened 2017-03-31
unexpected behavior with masked arrays


douglase:
Sherpa version 4.9.0, Python 2.7.12,  NumPy 1.11.2 (also seen on NumPy 1.12.1)
This issue was discovered when fitting a masked 2D array but is easily reproducible with the first few lines of the example notebook:

Create a masked numpy array:
```
x = np.arange(-5, 5.1)
y = x*x + 23.2 + np.random.normal(size=x.size)
old_y = x*x + 23.2 + np.random.normal(size=x.size)
y = np.ma.masked_array(old_y,mask=y<35)
e = np.ones(x.size)

sherpa.load_arrays("mydata", x, y, e)
sherpa.plot_data("mydata")
print(y.mask)

```
The mask appears set but the data points still appear.
If you multiply the masked array by 1.0 (or as I noticed this, by taking the sqrt of an array) the behavior changes:

```
y = 1.0*np.ma.masked_array(old_y,mask=y<35)
sherpa.load_arrays("mydata", x, y, e)
sherpa.plot_data("mydata")
print(y.mask)
```

Screenshots:

<img width="555" alt="screen shot 2017-03-31 at 4 29 50 pm" src="https://cloud.githubusercontent.com/assets/1025951/24568255/4b58c3ba-162f-11e7-8ed7-54ee52b69d96.png">

This still produces unexpected behavior, as these new zero values aren't just plotted, they got included in the fit further down the notebook. 
### Work around
Forcing masked values in y to np.nan gives the expected behavior, e.g.
`y = np.ma.filled(1.0*np.ma.masked_array(old_y,mask=y<35),fill_value=np.nan)` then the masked y data points are neither plotted or fitted.

olaurino:
I don't think Sherpa currently supports numpy masked arrays. As a matter of fact, when trying to reproduce the issue in CIAO `load_arrays` fails because the code in the CIAO backend explicitly checks that the arrays are of certain types. This is not to say such behavior is correct, of course. I'll take a closer look to see if we could easily accommodate numpy masked arrays and keep you posted.

In the meantime, Sherpa offers its own method for setting a filter. The catch is that it follows a different convention by default (a positive value or `True` for a "bad" channel, 0 or `False` for a good channel), so you need to negate the mask or set a special keyword, e.g.:

```` python
from sherpa.astro import ui as sherpa
import numpy as np

x = np.arange(-5, 5.1)
y = x*x + 23.2 + np.random.normal(size=x.size)
[...]

PR #698, opened 2019-10-21
Add warnings in documentation about masked arrays loaded into Sherpa

hamogu:
Sherpa currently does not support numpy masked arrays. See #346
and #695 for related bug reports. This does not fix the issue, but
adds documentation to make users more aware of this issue and point
to work-arounds.


DougBurke:
Thanks @hamogu

DougBurke:
[hmmm, the os-x travis job is having trouble downloading the xspec model-only binary package, causing the tests to fail for a reason unrelated to this PR; am on the second restart of that particular job (2754.1 on travis)

PR #699, opened 2019-10-25
a fix for /data/lenin2/Test/CIAO4.11/SherpaRegressionBeta1/45

dtnguyen2:
# Note
PR #616 was written to enable the testers to turn on (or off) caching at runtime to test PR #444  (and possibly PR #503).  This was done by adding the argument ```cache``` to all ```startup``` methods, unfortunately sherpa's tests do not include the instantiation of the ```ARFModelPHA``` class so a recent test by SDS found this bug which is hopefully fixed by this PR.

 SDS has to decide if PR #616, #503 and #699 should be released for public consumption

### DIFF
--- a/sherpa/astro/tests/test_cache.py
+++ b/sherpa/astro/tests/test_cache.py
@@ -39,8 +39,8 @@ logger = logging.getLogger("sherpa")
 @requires_xspec
 class test_ARFModelPHA(SherpaTestCase):
     _fit_using_ARFModelPHA = {
-        'numpoints': 26786,
-        'dof': 26783
+        'numpoints': 1212,
+        'dof': 1208
     }
     
     def setup(self):
@@ -53,23 +53,19 @@ class test_ARFModelPHA(SherpaTestCase):
 
     def test_ARFModelPHA(self):
         from sherpa.astro import ui
-        ui.load_pha(self.make_path("acis_1597_v2_pha2.fits"))
-        ui.load_arf(4, self.make_path("acis_1597_v2_HEG_1_garf.fits"))
-        ui.load_arf(3, self.make_path("acis_1597_v2_HEG_-1_garf.fits"))
-        ui.load_arf(9, self.make_path("acis_1597_v2_MEG_-1_garf.fits"))
-        ui.load_arf(10, self.make_path("acis_1597_v2_MEG_1_garf.fits"))
-        ui.get_data(4).units="wavelength"
-        ui.get_data(3).units="wavelength"
-        ui.get_data(9).units="wavelength"
-        ui.get_data(10).units="wavelength"
-        ui.set_model(4,"xsphabs.abs1*powlaw1d.p1")
-        ui.set_model(3,"abs1*p1")
-        ui.set_model(9,"abs1*p1")
-        ui.set_model(10,"abs1*p1")
-        ui.notice(2, 30)
+        ui.load_pha(self.make_path("3c120_meg_1.pha"))
+        ui.group_counts(20)
+        ui.notice(0.5, 6)
+        ui.subtract()
+        ui.set_model(ui.xsphabs.abs1 * (ui.xsapec.bubble + ui.powlaw1d.p1))
+        abs1.nh = 0.163
+        abs1.nh.freeze()
+        p1.ampl = 0.017
+        p1.gamma = 1.9
+        bubble.kt = 0.5
+        bubble.norm = 4.2e-5
         tol = 1.0e-2
         ui.set_method_opt('ftol', tol)
-        ui.set_method_opt('gtol', tol)
         ui.fit()
         result = ui.get_fit_results()
         assert result.numpoints == self._fit_using_ARFModelPHA['numpoints']
--- a/sherpa/astro/tests/test_cache.py
+++ b/sherpa/astro/tests/test_cache.py
@@ -58,6 +58,8 @@ class test_ARFModelPHA(SherpaTestCase):
         ui.notice(0.5, 6)
         ui.subtract()
         ui.set_model(ui.xsphabs.abs1 * (ui.xsapec.bubble + ui.powlaw1d.p1))
+        ui.set_xsabund('angr')
+        ui.set_xsxsect('vern')
         abs1.nh = 0.163
         abs1.nh.freeze()
         p1.ampl = 0.017
--- a/sherpa/astro/tests/test_cache.py
+++ b/sherpa/astro/tests/test_cache.py
@@ -54,6 +54,11 @@ class test_ARFModelPHA(SherpaTestCase):
     def test_ARFModelPHA(self):
         from sherpa.astro import ui
         ui.load_pha(self.make_path("3c120_meg_1.pha"))
+
+        # remove the RMF to ensure this is an ARF-only analysis
+        # (which is what is needed to trigger the bug that lead to #699)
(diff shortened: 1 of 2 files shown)


DougBurke:
I restarted the two failed tests since they had timed out - as all the other tests are green I expect these two to pass too.

dtnguyen2:
The fix to this PR and PR #616 was to add the cache to the argument of the ```startup``` method, so a simple search for the remaining startup yields:
```
[dtn@devel12 sherpa]$ grep -r startup | grep -v __pycache__ | grep -v cache | grep startup | grep def
astro/instrument.py:    def startup(self):
astro/instrument.py:    def _startup_user_grid(self, x, xhi=None):
astro/instrument.py:    def _startup_user_grid(self, x):
models/basic.py:    def startup(self):
models/model.py:    def startup(self):
models/model.py:    def startup(self):
```
so one has to come up with data and script for the remaining startup to see what else will fail.

sherpa/astro/tests/test_cache.py
```diff
@@ -33,6 +33,46 @@
 import logging
 logger = logging.getLogger("sherpa")
 
+@requires_data
+@requires_fits
+class test_ARFModelPHA(SherpaTestCase):
```
DougBurke: This test also needs a `requires_xspec` decorator

DougBurke:
Is it possible to use the existing pha2 data set in `sherpa-test-data` - namely `3c120_pha2.gz` and the arf/rmf files (`3c120_heg/meg_<order>.arf|rmf.gz` - rather than add a new set of data files?

DougBurke:
I was interested in 

a) can we use existing data in our test suite to trigger this
b) why do we not see this failure in our existing test suite

I believe the answer to b is that this happens when there is no RMF, just an ARF. I think our existing grating/pha2 tests load both RMF and ARF (and the "stanard" PHA tests all have both). We can use the existing `3c120_pha2.gz` + responses to show the problem, but I've managed to re-create it with just a single dataset (a PHA1 "extraction" of the 3C120 grating data). The following script should fail without the fix and run with the fix (once BASEDIR is set to the correct location, and the code uses a make_path function to parallel the test suite)

```
import os
from sherpa.astro import ui

BASEDIR = '/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest'

def make_path(infile):
    return os.path.join(BASEDIR, infile)

ui.load_pha(make_path("3c120_meg_1.pha"))

# remove the RMF to trigger the bug
ui.get_data().set_rmf(None)

ui.group_counts(20)

ui.notice(0.5, 6)
ui.subtract()

ui.set_model(ui.xsphabs.abs1 * (ui.xsapec.bubble + ui.powlaw1d.p1))

abs1.nh = 0.163
abs1.nh.freeze()

p1.ampl = 0.017
p1.gamma = 1.9

bubble.kt = 0.5
bubble.norm = 4.2e-5

tol = 1.0e-2
ui.set_method_opt('ftol', tol)
ui.set_method_opt('gtol', tol)

ui.fit()
result = ui.get_fit_results()
print(result)
```

The result of the fit I get are

```
Dataset               = 1
Initial fit statistic = 842.044
Final fit statistic   = 838.705 at function evaluation 140
Data points           = 1212
Degrees of freedom    = 1208
Probability [Q-value] = 1
Reduced statistic     = 0.694292
Change in statistic   = 3.33904
   bubble.kT      1.65474      +/- 0           
   bubble.norm    2.12601e-05  +/- 9.05638e-05 
   p1.gamma       1.90239      +/- 0           
   p1.ampl        0.0168472    +/- 0.00011549  
datasets       = (1,)
itermethodname = none
methodname     = levmar
statname       = chi2gehrels
succeeded      = True
parnames       = ('bubble.kT', 'bubble.norm', 'p1.gamma', 'p1.ampl')
parvals        = (1.6547364387121473, 2.126006897074647e-05, 1.9023895197834984, 0.016847226417499948)
statval        = 838.7047557945251
istatval       = 842.043793908448
dstatval       = 3.339038113922925
numpoints      = 1212
dof            = 1208
qval           = 1.0
rstat          = 0.6942920163861963
message        = successful termination
nfev           = 140
```

It seems a "reasonable" fit and is physically motivated, but I think I agree with Dan's original test that all we really want is a simple check that the test has run and got a reasonable result.

DougBurke:
If we also want to make sure we have a test of fitting multiple data sets, then we could add another dataset - e.g.

```
import os
from sherpa.astro import ui

BASEDIR = '/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest'

def make_path(infile):
    return os.path.join(BASEDIR, infile)

ui.load_pha(1, make_path("3c120_meg_1.pha"))
ui.load_pha(2, make_path("3c120_heg_1.pha"))

# remove the RMF to trigger the bug
ui.get_data(1).set_rmf(None)
ui.get_data(2).set_rmf(None)

ui.group_counts(1, 20)
ui.group_counts(2, 20)

ui.subtract(1)
ui.subtract(2)

ui.notice(0.5, 6)

mdl = ui.xsphabs.abs1 * (ui.xsapec.bubble + ui.powlaw1d.p1)
ui.set_model(1, mdl)
ui.set_model(2, mdl)

abs1.nh = 0.163
abs1.nh.freeze()

p1.ampl = 0.017
p1.gamma = 1.9

bubble.kt = 0.5
bubble.norm = 4.2e-5

tol = 1.0e-2
ui.set_method_opt('ftol', tol)
ui.set_method_opt('gtol', tol)

ui.fit()
result = ui.get_fit_results()
print(result)
```

which gives for me

```
Datasets              = 1, 2
Initial fit statistic = 1311.62
Final fit statistic   = 1303.16 at function evaluation 120
Data points           = 1879
Degrees of freedom    = 1875
Probability [Q-value] = 1
Reduced statistic     = 0.695018
Change in statistic   = 8.46225
   bubble.kT      1.85603      +/- 0           
   bubble.norm    4.17611e-06  +/- 8.5543e-05  
   p1.gamma       1.87975      +/- 0           
   p1.ampl        0.0165669    +/- 9.70099e-05 
datasets       = (1, 2)
itermethodname = none
methodname     = levmar
statname       = chi2gehrels
succeeded      = True
parnames       = ('bubble.kT', 'bubble.norm', 'p1.gamma', 'p1.ampl')
parvals        = (1.856027445054159, 4.176109275671569e-06, 1.8797453505998039, 0.016566901282489093)
statval        = 1303.1580521246374
istatval       = 1311.6203026767485
dstatval       = 8.462250552111072
numpoints      = 1879
dof            = 1875
qval           = 1.0
rstat          = 0.6950176277998066
message        = successful termination
nfev           = 120
```

DougBurke:
Note - in all these tests we probably want to add

    ui.set_xsabund('angr')
    ui.set_xsxsect('vern')

to avoid possible issues with changes in XSPEC.

DougBurke:
@dtnguyen2 - there's a line missing from the test case (otherwise it doesn't test the code path we want to test). I can fix this and push a fix, if that's okay with you?

DougBurke:
I've put the rebased version up at #701 (so as not to change this). You can see the needed change to the test at /data/lenin2/Test/CIAO4.11/SherpaRegressionBeta1/45 - I'll add the fix to #701 once the travis tests have run and failed (just to check that this test is really showing the bug).

wmclaugh:
Merged PR$701 (rebase) of this PR

PR #704, opened 2019-11-07
change calc_ftest from delta dof to dof (similar to XSPEC)

dtnguyen2:
# Release Notes
Sherpa calc_ftest does not give the same results as XSPEC's ftest command; Moreover current docs is not entirely correct

# Note
This PR should be considered if sherpa's calc_ftest is to give the same results as XSPEC's ftest, otherwise reject it.  The PR is a 'fix' for issue https://github.com/sherpa/sherpa/issues/641

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -51,8 +51,11 @@ debug = logging.getLogger("sherpa").debug
 config = ConfigParser()
 config.read(get_config())
 
-_ncpu_val = config.get('parallel', 'numcores',
-                       fallback='None').strip().upper()
+_ncpu_val = "NONE"
+try:
+    _ncpu_val = config.get('parallel', 'numcores').strip().upper()
+except NoSectionError:
+    pass
 
 _ncpus = None
 if not _ncpu_val.startswith('NONE'):
@@ -178,13 +181,13 @@ def calc_ftest(dof1, stat1, dof2, stat2):
 
     Parameters
     ----------
-    dof1 : int
+    dof1 : int or array/list/tuple of int
        degrees of freedom of the simple model
-    stat1 : number
+    stat1 : number or array/list/tuple of number
        best-fit chi-square statistic value of the simple model
-    dof2 : int
+    dof2 : int or array/list/tuple of int
        degrees of freedom of the complex model
-    stat2 : number
+    stat2 : number or array/list/tuple of number
        best-fit chi-square statistic value of the complex model
 
     Returns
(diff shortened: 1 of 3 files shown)


sherpa/utils/__init__.py
```diff
@@ -51,8 +51,11 @@
 config = ConfigParser()
 config.read(get_config())
 
-_ncpu_val = config.get('parallel', 'numcores',
-                       fallback='None').strip().upper()
+_ncpu_val = "NONE"
```
DougBurke: I don't understand why this has changed - the use of the `fallback` argument with the `get` call means the `try`/`except` block isn't needed. Could this just be because #684 was merged after this code was originally written?

sherpa/utils/__init__.py
```diff
@@ -232,16 +235,11 @@ def calc_ftest(dof1, stat1, dof2, stat2):
 
     Examples
     --------
+    >>> calc_ftest(11, 16.3, 10, 10.2)
```
DougBurke: @anetasie - given that it's easy to mess this up, should we include some text here to say what is being compared and what the result means. So, something like the original text but actually correct?

sherpa/utils/tests/test_utils_unit.py
```diff
@@ -21,16 +21,38 @@
...
 from sherpa.utils import _utils, is_binary_file, pad_bounding_box
 from sherpa.utils.testing import requires_data
-from numpy.testing import assert_almost_equal, assert_array_equal
+from numpy.testing import assert_almost_equal, assert_array_equal, \
+    assert_array_almost_equal
 
 
 def test_calc_ftest():
     """
     this test was created using sherpa 4.8.1 (2507414) as an oracle for the python 3 migration
```
DougBurke: This comment should be removed, since the test value (oracle) wasn't calculated using sherpa 4.8.1, I would guess.

sherpa/utils/__init__.py
anetasie: help(calc_ftest) shows the information about the function and what it means. We may not need the doc update in this PR.
or you just referring to the test.

anetasie: I guess you want to update the descriptions in the Example section.

DougBurke: Yes, I was talking about the examples. This PR removes text explaining the first example. Now, this is good because that text was wrong and the first example has changed, but I think it may be helpful to provide some explanation for the examples.

anetasie: Yes, we will do this in a separate PR

anetasie:
We may have a nicer error msgs when `dof2>dof1`
Now the return is `nan`.

PR #706, opened 2019-11-10
change how minim handles exceeding boundary limits

dtnguyen2:
# Note

When the Fortran opt algo  _minim.f_ was translated to C++, it was reported that the number of function evaluations (nfev) had increased (see [sds](https://icxc.harvard.edu/pipe/sherpadev/2018/0597.html) for details).  Previously, _minim_ handles simple boundary limits as _an infinite potential well_, ie the cost function would return very large value which makes it difficult to count nfev.  The C++ _minimh.hh_ now reflects about the boundary limit by the same amount as was exceeded, with the extra care to not exceed the other boundary limit in the even the limits are very small.

This PR replaces PR #557 since it is difficult to rebase and/or cherry pick

### DIFF
--- a/sherpa/optmethods/tests/test_optmethods.py
+++ b/sherpa/optmethods/tests/test_optmethods.py
@@ -20,9 +20,11 @@ from __future__ import print_function
 
 from sherpa.utils import _ncpus
 from sherpa.utils.testing import SherpaTestCase
+from sherpa.optmethods import _saoopt
 from sherpa.optmethods import optfcts
 from sherpa.optmethods import _tstoptfct
 
+import numpy as np
 
 class test_optmethods(SherpaTestCase):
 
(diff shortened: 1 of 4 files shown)


sherpa/optmethods/tests/test_optmethods.py
```diff
@@ -115,6 +117,20 @@ def test_gaussian(self):
         self.tst( optfcts.lmdif, name + self.lm, _tstoptfct.gaussian,
                   fmin, x0, xmin, xmax )
 
+    def test_mccormick(self):
```
DougBurke: @dtnguyen2 - `test_mccormick` passes on master branch (ie without your change above to use `reflect_about_boundary`). Is this expected?

dtnguyen2:
yes

PR #709, opened 2019-11-12
Residual-style plots ignore the ylog setting (fix #586)

### LINKED ISSUE #586: should set_ylog not change the behavior of residual plots?
If I

```
>>> set_xlog()
>>> set_ylog()
>>> plot_fit_resid()
```

then both the fit and residual plots have X and Y axes drawn with a logarithmic scale. Given that the residual (and delchi) plots are expected to contain negative values, this is not very useful.

I suggest that we make the residual-style plots (`_resid` and `_delchi`) ignore the `ylog` setting of the `ui.get_data_plot_prefs` setting. There is also the question of if they should have their own setting, but I think that's a larger change.

DougBurke:
Residual, ratio, and delchi plots always use a linear scale for the y axis, no matter what the `ylog` setting is.
    
This address #586
    
The plot method for these classes has been over-ridden to force the
`ylog` preference setting to be `False`, and to drop the `ylog` keyword
argument if specified. This means that a command like (from the
ui layer)
    
    set_ylog()
    plot_fit_resid()
    
will use a logarithmic Y axis for the top plot but a linear one for
the bottom plot.

Notes
--------
    
This might be better done as a Mixin / Multi-Inheritance class, since
the logic is the same for all three cases, but I have not looked into
this approach.

This was originally part of the WSTAT plot fix (PR #694) but has been pulled out as it is a separate issue and fix.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -11401,6 +11401,9 @@ class Session(sherpa.ui.utils.Session):
         Display the residuals for the background of a PHA data set
         when it is being fit, rather than subtracted from the source.
 
+        .. versionchanged:: 4.12.0
+           The Y axis is now always drawn using a linear scale.
+
         Parameters
         ----------
         id : int or str, optional
@@ -11437,6 +11440,11 @@ class Session(sherpa.ui.utils.Session):
         plot_bkg_ratio : Plot the ratio of data to model values for the background of a PHA data set.
         set_bkg_model : Set the background model expression for a PHA data set.
 
+        Notes
+        -----
+        The ylog setting is ignored, and the Y axis is drawn using a
+        linear scale.
+
         Examples
         --------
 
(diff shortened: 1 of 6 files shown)


anetasie:
I tested the plots. They look good.

DougBurke:
I can look at adding some. Our automated tests of our plotting code is rather basic, although I have been trying to improve this (but without some relatively-large investment of work it's never going to be great).

DougBurke:
I've just rebased this to make it easier to add more tests to (as I'm using python 3.8).

DougBurke:
I've added some ui-level tests for `plot_resid`, `plot_delchi`, `plot_fit_resid`, and `plot_fit_delchi`.

anetasie:
@wmclaugh I reviewed the PR and the plots looked good. @DougBurke added new UI tests and one of the test failed, because the build did not complete. Is this related to the `pytest` issues?

DougBurke:
Err, I see not test failure.

wmclaugh:
There was an initial test failure reported. Marie restarted the  test this morning. The rerun appears to have been successful.

Issue #717, opened 2019-12-15
Unable to fit with load_multi_arf/load_multi_rmf: startup() takes 1 positional argument but 2 were given


DougBurke:
# Summary

Noted by SDS, Sherpa 4.12 (CIAO 4.12) has the following error:

```
sherpa In [1]: load_pha(1, "460_leg_m1_bin10.pha")                                                                                                               
WARNING: systematic errors were not found in file '460_leg_m1_bin10.pha'
statistical errors were found in file '460_leg_m1_bin10.pha' 
but not used; to use them, re-read with use_errors=True
read background_up into a dataset from file 460_leg_m1_bin10.pha
read background_down into a dataset from file 460_leg_m1_bin10.pha

sherpa In [2]: load_multi_arfs(1, ["460_LEG_-1.garf","460_LEG_-2.garf","460_LEG_-3.garf"], [1,2,3]) 

sherpa In [3]: load_multi_rmfs(1, ["460_leg_-1.grmf","460_leg_-2.grmf","460_leg_-3.grmf"], [1,2,3]) 

sherpa In [4]: mdl = xswabs.abs1 * xsbknpower.bpow1 

sherpa In [5]: set_model(1, mdl)                                                                                            

sherpa In [6]: bpow1.norm = 0.05                                                                                            

sherpa In [7]: plot_fit(ylog=True)                                                                                          
```

The plot works - i.e. you can evaluate the model - but when you come to try to fit it (the `sys.tracebacklimit` setting is to make sure we get a reasonable traceback here) the fit fails

```
sherpa In [8]: import sys                                                                                                   

sherpa In [9]: sys.tracebacklimit = 1                                                                                       

sherpa In [10]: fit()                                                                                                       
WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
  File "/home/dburke/anaconda3/envs/ciao412/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3319, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
TypeError: startup() takes 1 positional argument but 2 were given
```

# Workaround

This can be worked around by monkey-patching the `startup` method of the problematic class before any instances have been created:

```
# patch the startup method of the MultiResponseSumModel *before* any
# instances have been created
#
import numpy
from sherpa.models.model import CompositeModel
from sherpa.astro.instrument import MultiResponseSumModel

def startup_monkey(self, cache):
    pha = self.pha
[...]

PR #719, opened 2019-12-20
Add a test suite for multiple ARF/RMF responses

DougBurke:
# Summary

Add some basic tests of multiple instruments (ARF/RMF) as follow up to #717

# Details

This is a set of tests of simple composite (ie multiple ARF and/or RMF) instruments. The aim is to provide some testing of PHA fitting with multiple responses, such as those provided by load_multi_arfs/load_multi_rmfs (although since the load_multi_xxx routines are simple wrappers around the set_arf/rmf functions these tests just use these lower-level routines).

This was added to support #717 (i.e. it does not necessarily add new tests but just makes a few explicit tests).

Here is the minor difference in coverage, compared to the master branch (but I see this as more important to have some simple tests of combined responses):

```
Name                                                     Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------------------------------------
10c10
< sherpa/astro/data.py                                      1327    300    568     98    74%
---
> sherpa/astro/data.py                                      1327    294    568     96    75%
18c18
< sherpa/astro/instrument.py                                 644    114    178     46    76%
---
> sherpa/astro/instrument.py                                 644     99    178     41    80%
76c76
< TOTAL                                                    24022   6571   6980   1121    68%
---
> TOTAL                                                    24022   6550   6980   1114    69%
```

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
@@ -34,6 +34,7 @@ import pytest
 
 from sherpa.astro import ui
 from sherpa.astro.instrument import create_arf
+from sherpa.astro.data import DataRMF
 
 
 def get_egrid():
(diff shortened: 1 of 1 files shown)


DougBurke:
Rebased to pick up python 3.5 fixes so that the Travis tests should pass.

Issue #720, opened 2019-12-21
importing stk or group modules cause segfault on macOS Mojave


neobar:
Hi, 
The OS is macOS Mojave, and python version is 3.7.4, Anaconda.
I have this problem after building sherpa.
Xspec models are not configured.

$ sherpa_test
Requirement already satisfied: pytest!=5.2.3,>=3.3 in ./anaconda3/lib/python3.7/site-packages (5.3.2)
Requirement already satisfied: py>=1.5.0 in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (1.8.0)
Requirement already satisfied: packaging in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (19.2)
Requirement already satisfied: attrs>=17.4.0 in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (19.3.0)
Requirement already satisfied: more-itertools>=4.0.0 in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (8.0.2)
Requirement already satisfied: pluggy<1.0,>=0.12 in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (0.13.1)
Requirement already satisfied: wcwidth in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (0.1.7)
Requirement already satisfied: importlib-metadata>=0.12 in ./anaconda3/lib/python3.7/site-packages (from pytest!=5.2.3,>=3.3) (1.3.0)
Requirement already satisfied: pyparsing>=2.0.2 in ./anaconda3/lib/python3.7/site-packages (from packaging->pytest!=5.2.3,>=3.3) (2.4.5)
Requirement already satisfied: six in ./anaconda3/lib/python3.7/site-packages (from packaging->pytest!=5.2.3,>=3.3) (1.13.0)
Requirement already satisfied: zipp>=0.5 in ./anaconda3/lib/python3.7/site-packages (from importlib-metadata>=0.12->pytest!=5.2.3,>=3.3) (0.6.0)
=============================================================================================== test session starts ================================================================================================
platform darwin -- Python 3.7.4, pytest-5.3.2, py-1.8.0, pluggy-0.13.1
rootdir: /Users/sxz89
[...]

PR #721, opened 2019-12-23
Docs: update installation notes for conda environment

DougBurke:
# Summary

Add a warning about the need to set PYTHON_LDFLAGS to ' ' on macOS when building within an anaconda environment. This is a documentation-only change.

# Details

This is to address the problem noted in
https://github.com/sherpa/sherpa/issues/720
where you need to set PYTHON_LDFLAGS to ' ' (a space, not
empty string) on macOS when building with conda.

There are several other minor tweaks made (mainly updaing
the text for more-recent versions).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -173,7 +173,8 @@ For example::
     cd sherpa
     git checkout 4.10.0
 
-will use the ``4.10.0`` tag.
+will use the ``4.10.0`` tag (although we strongly suggest using a
+newer release now!).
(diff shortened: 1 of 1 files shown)


DougBurke:
@dtnguyen2 - when building the XSPEC interface we also don't actually need a fortran compiler, we just need to be able to link against the fortran runtime (or may have to link against it; it seems a bit slippery). At least, that's what I believe.

Issue #722, opened 2020-01-03
1D integrated models do not regrid properly for partial overlap of grids


DougBurke:
I have an example where the `regrid` method fails when the grid it is given doesn't fully cover the base grid. There's actually two cases, 

a) a run-time error where the number of array elements don't match (this looks to be from code that is trying to handle this lack of coverage, and I thought we had an issue about this already but perhaps not)

b) or a case where the model is evaluated, but calculates double the expected value

There's a Jupyter notebook, script extracted from the notebook, and HTML version of the notebook at https://gist.github.com/DougBurke/cfa5a83090b464b0875e0d08a52d8ca8

The notebook can be viewed at https://nbviewer.jupyter.org/gist/DougBurke/cfa5a83090b464b0875e0d08a52d8ca8/problems.ipynb

To simplify this I've created tests that show the failure (for the "1D integrated" case, seem to work fine for the 1D case) in #724

DougBurke:
Could case b - the double counting - be due to the fact that Sherpa accidentally includes the predicted model values from the "base" grid in the calculation (rather than only doing this for those bins that lie outside the range given to `regrid`).

I wonder if maybe we should just set points outside the `regrid` range to 0 rather than try to correct for the missing bins?

DougBurke:
The problem of the "double counting" is because of the following in `sherpa/models/regrid.py` in `EvaluationSpace1D.join`: https://github.com/sherpa/sherpa/blob/master/sherpa/models/regrid.py#L330

```
new_xlo = np.unique(np.concatenate((self.x_axis.lo, other.x_axis.lo)))
if self.x_axis.is_integrated:
    new_xhi = np.unique(np.concatenate((self.x_axis.hi, other.x_axis.hi)))
```

This doesn't handle edge differences due to numerical precision - for example:

```
In [16]: xgrid = np.arange(2.5, 5, 0.2)                                         

In [17]: ogrid = np.arange(2, 6,0.1)                                            

In [18]: xgrid                                                                  
Out[18]: array([2.5, 2.7, 2.9, 3.1, 3.3, 3.5, 3.7, 3.9, 4.1, 4.3, 4.5, 4.7, 4.9])

In [19]: ogrid                                                                  
Out[19]: 
array([2. , 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8, 2.9, 3. , 3.1, 3.2,
[...]

PR #724, opened 2020-01-06
Add regrid tests

DougBurke:
# NOTE

Replaced by #764, which is this PR rebased onto the fix in #747

# Details

This PR adds tests to the regrid code based on issue #722 - that is, the handling of partial overlap between the output grid and the user-specified "regrid".

The 1D non-integrated tests pass, but the 1D integrated tests - `test_low_level_regrid1d_int_partial_overlap` - fail because the model is evaluated to twice what it should be (and so they are marked as `pytest.mark.xfail`). Actually, the last one errors out with  `TypeError: 1D model evaluation input array sizes do not match, xlo: 49 vs xhi: 48` (both failures were seen in #722).

These tests are not necessarily sufficient for the overlap case:

a) they only cover the 1D case
b) the model chosen for testing is explicitly set to 0 in the areas where the grids fail to overlap, as it is not clear to me what the expected behavior here is (from the warning message I think that the model is expected to be evaluated on the old grid where there is no overlap, but I am not 100% sure of this).

As a "base case" I added tests that check when the regrid grid is larger than the output grid. Surprisingly enough this fails for the 1D non-integrated case when the bin size is larger than the output bin size. In this case the last two bins are not zero but have significant counts in them. I don't know why, so have marked this particular test xfail too.

I've also added a test of the low-level rebin method to show that when it is being asked to do what I think it should be in this case it does. In other words this does seem to be somewhere in the `regrid` method/class structure.

# Where is the bug?

I think the problem is that the evaluation space has both grids in it - so in our case the union of `np.arange(1, 9, 0.5)` and `np.arange(2.1, 8, 0.5)` (or something similar), which means that the model will evaluate x=4-4.5 and x=4.1-4.5, which is going to essentially double the y values once it has been passed to `sherpa.utils._utils.rebin`. '

Notice the repeated elements in the joined grid (I think this is what is happening in the larger example, but haven't guaranteed it completely) - e.g. the low edges go `..., 1.8, 1.9, 2, 2.1, 2.1, 2.2, 2.2, ...`:

```
In [12]: x1grid = np.arange(1, 9, 0.1)                                          

In [13]: x2grid = np.arange(2.1, 8, 0.1)                                        

In [14]: x2space = EvaluationSpace1D(x2grid[:-1], x2grid[1:])                   

In [15]: x1space = EvaluationSpace1D(x1grid[:-1], x1grid[1:])                   

In [16]: x2space in x1space                                                     
Out[16]: True

In [17]: x1space in x2space                                                     
Out[17]: False

In [18]: jspace = x2space.join(x1space)                                         

In [19]: jspace.grid                                                            
Out[19]: 
(array([1. , 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2. , 2.1, 2.1,
        2.2, 2.2, 2.3, 2.3, 2.4, 2.4, 2.5, 2.5, 2.6, 2.6, 2.7, 2.7, 2.8,
        2.8, 2.9, 2.9, 3. , 3. , 3.1, 3.1, 3.2, 3.2, 3.3, 3.3, 3.4, 3.4,
        3.5, 3.5, 3.6, 3.6, 3.7, 3.7, 3.8, 3.8, 3.9, 3.9, 4. , 4. , 4.1,
        4.1, 4.2, 4.2, 4.3, 4.4, 4.4, 4.5, 4.5, 4.6, 4.6, 4.7, 4.8, 4.8,
        4.9, 4.9, 5. , 5. , 5.1, 5.2, 5.2, 5.3, 5.3, 5.4, 5.4, 5.5, 5.6,
        5.6, 5.7, 5.7, 5.8, 5.8, 5.9, 6. , 6. , 6.1, 6.1, 6.2, 6.2, 6.3,
        6.4, 6.4, 6.5, 6.5, 6.6, 6.6, 6.7, 6.8, 6.8, 6.9, 6.9, 7. , 7. ,
        7.1, 7.2, 7.2, 7.3, 7.3, 7.4, 7.4, 7.5, 7.6, 7.6, 7.7, 7.7, 7.8,
        7.8, 7.9, 8. , 8.1, 8.2, 8.3, 8.4, 8.5, 8.6, 8.7, 8.8]),
 array([1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2. , 2.1, 2.2, 2.2,
        2.3, 2.3, 2.4, 2.4, 2.5, 2.5, 2.6, 2.6, 2.7, 2.7, 2.8, 2.8, 2.9,
        2.9, 3. , 3. , 3.1, 3.1, 3.2, 3.2, 3.3, 3.3, 3.4, 3.4, 3.5, 3.5,
        3.6, 3.6, 3.7, 3.7, 3.8, 3.8, 3.9, 3.9, 4. , 4. , 4.1, 4.1, 4.2,
        4.2, 4.3, 4.4, 4.4, 4.5, 4.5, 4.6, 4.6, 4.7, 4.8, 4.8, 4.9, 4.9,
        5. , 5. , 5.1, 5.2, 5.2, 5.3, 5.3, 5.4, 5.4, 5.5, 5.6, 5.6, 5.7,
        5.7, 5.8, 5.8, 5.9, 6. , 6. , 6.1, 6.1, 6.2, 6.2, 6.3, 6.4, 6.4,
        6.5, 6.5, 6.6, 6.6, 6.7, 6.8, 6.8, 6.9, 6.9, 7. , 7. , 7.1, 7.2,
        7.2, 7.3, 7.3, 7.4, 7.4, 7.5, 7.6, 7.6, 7.7, 7.7, 7.8, 7.8, 7.9,
        8. , 8.1, 8.2, 8.3, 8.4, 8.5, 8.6, 8.7, 8.8, 8.9]))

```
```

### DIFF
--- a/sherpa/models/tests/test_regrid_unit.py
+++ b/sherpa/models/tests/test_regrid_unit.py
@@ -480,6 +480,33 @@ def test_regrid1d_error_grid_mismatch_2(setup_1d):
     assert ModelErr.dict['needspoint'] in str(excinfo.value)
 
 
+@pytest.mark.parametrize("requested, tol",
+                         [ (np.arange(1, 7, 0.1), 1e-7),
+                           (np.arange(1, 7, 0.05), 2e-4),
+                           pytest.param(np.arange(1, 7, 0.2), 0.1,
+                                        marks=pytest.mark.xfail)
+                       ])
+def test_low_level_regrid1d_full_overlap(requested, tol):
+    """Base case of test_low_level_regrid1d_partial_overlap
+    """
+
+    # The range over which we want the model evaluated
+    xgrid = np.arange(2, 6, 0.1)
+    d = Data1D('tst', xgrid, np.ones_like(xgrid))
+
+    mdl = Box1D()
+    mdl.xlow = 3.1
+    mdl.xhi = 4.2
+    mdl.ampl = 0.4
+
+    yexpected = d.eval_model(mdl)
+    assert yexpected.min() == pytest.approx(0.0)
+    assert yexpected.max() == pytest.approx(0.4)
+
+    ygot = d.eval_model(mdl.regrid(requested))
+    assert ygot == pytest.approx(yexpected, abs=tol)
+
+
 @pytest.mark.parametrize("requested",
                          [ np.arange(2.5, 7, 0.2),
                            np.arange(1, 5.1, 0.2),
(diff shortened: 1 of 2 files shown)


DougBurke:
I've just rebased these against the current master branch to make it easier to test with #747 

Note that these tests are written with the current behavior in mind, not those of #747 - so several tests will need to be changed to not assert that a warning message is created

Issue #723, opened 2020-01-03
Warnings when running resample_data()


anetasie:
I see a lot of warnings while running 'resample_data()' in CIAO4.12 Sherpa. Test data for 
asymmetric errors will show the screen printout for every iteration. 

      load_ascii_with_errors(1,'gro.txt')
      set_stat('leastsq')
      set_model(powlaw1d.p1)
      fit()
      sample=resample_data(1,1000)

  

       .... warnings.warn("Converting array {} to numpy array".format(array))
     /Users/aneta/soft/ciao-4.12/ciao-4.12/lib/python3.5/site-packages/sherpa-4.12.0-py3.5- 
     macosx-10.12-x86_64.egg/sherpa/data.py:49: UserWarning: Converting array 
     [13.381319496084906, 13.633260235068068, 14.306213914726577, 13.983557790201848, 
    13.171125623509718, 12.488433532904432, 14.626279812074497, 16.012878283581944, 
    15.258566424800357, 13.690025984816309, 18.514528963412666, 14.876755444565106, 
    16.32593948807569, 11.884746992186537, 11.257004828977017, 11.134748193051466, 
    11.666966554920794, 11.688190461005217, 9.080282498957114, 10.042537574980583, 
    9.794066377984949, 9.180452401453762, 10.331401797264734, 10.69460004113246, 
    8.958842965993878, 7.552101436170255, 7.858390587996477, 10.645408597899696, 
    8.371732877410645, 6.9382374151118675, 8.068750514358847, 5.725444743683091, 
    6.479597857237468, 7.947197940606003, 5.017390444344825, 4.408614337830014, 
    5.470251340952552, 5.400760418773997, 4.926646231399067, 6.669486210793662, 
    3.946138706536805, 5.491861097640214, 20.473864743135888, 9.584644867183084, 
    2.9941695735462095, 3.8880767097610125, 2.8402196610798445, 1.6563930690598663, 
    2.5479535666395163, 2.3527045444059103, 2.774612664648795, 2.2056629549101965, 
    3.6670732010833658, 2.9395946286845893, 3.4041964024021647, 2.261442059019056, 
    1.537098568238905, 4.384837877801289, 0.5012254710457373, 4.472802659515596, 1 
    1.9897275999338797] to numpy array
     warnings.warn("Converting array {} to numpy array".format(array))
     p1.gamma : avg = -0.46824871286051334 , std = 0.07754346661186172
[...]

PR #725, opened 2020-01-07
convert a list to np.array to avoid warning messages

dtnguyen2:
# Note
This PR fixes issue #723 (a lot of warnings while running 'resample_data()' in CIAO4.12 Sherpa)

### DIFF
--- a/sherpa/sim/tests/test_asymmetric.py
+++ b/sherpa/sim/tests/test_asymmetric.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2019, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -18,6 +18,7 @@
 #
 
 import numpy as np
+import warnings
 
 from sherpa.fit import Fit
 from sherpa.data import Data1DAsymmetricErrs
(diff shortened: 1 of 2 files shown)


DougBurke:
The macOS test had failed, but this was due to a (hopefully) random error - namely

```checking whether build environment is sane... configure: error: newly created file is older than distributed files!```

so I've re-started it (it was the only one marked as having failed).

File: sherpa/sim/tests/test_asymmetric.py [hunk 1]
```diff
@@ -182,3 +183,15 @@ def test_ui(self, tol=1.0e-3):
                                   sample['p1.gamma'])
         self.assertEqualWithinTol(self._resample_bench_10['p1.ampl'],
                                   sample['p1.ampl'])
+
+    def test_warning(self):
```
DougBurke: @dtnguyen2 - is there any reason you didn't do the fit using the ui layer? Not that it makes a difference here,but it looks a bit odd. That is

    ui.load_ascii_with_errors(1, self.gro_fname)
    ui.set_model(ui.powlaw1d.pl)
    ui.fit()
    with warnings.bla as w:
        more-blah

dtnguyen2: no real reason, other then laziness by cut and paste the code from the original bug report.  Want me to use the ui?

DougBurke: No - I don't think so. It just looks a bit odd but I don't think it's a major issue (unless there was something that needed it done this way, but there isn't. So it isn't).

PR #732, opened 2020-01-21
Remove unused Python 2.7 compatibility code

DougBurke:
# Summary

Remove unused code from the removal of Python 2.7 support.

# Details

Remove conditional code from Python 2.7 support. This is primarily driven from comments in the code pointing out changes made when adding python 3 support, so this is not guaranteed to catch everything (there is some code in datastack but I'm going to make a separate PR for that as it is more involved).

Changes include:

- remove `inspect.getargspec` as `inspect.signature` is supported by Python >= 3.3
- rework `is_binary_file` to match try/except handling (only catch where we expect an error to be raised) and to use a context manager to handle closing the file; the documentation of this internal routine has also been extended to spell out what "binary" means here
- rework conversion code in crates to be simpler now that there is no need to support Python 2.7
- Remove code that was only run with Python 2.7 (e.g. change "try: py27-call / except: py3-call" to just "py3-call").
- in the DS9 module there was code used only for Python versions earlier than 3.2; this was removed as our minimum supported Python version is 3.5.
- remove conditional compilation for pre-python 3 code from C++ code (both source code and header files)

Note that there is a change to the crates back end which have not been tested (as we don't have crates available in our travis tests). I have tested this off to the side, using the conda CIAO 4.12 CIAO package, and the tests pass.

The increase in coverage is small, but there (note that this doesn't track the minor clean up in the compiled code):

label | statements | missing | excluded | branches | partial | coverage
----- | ---------- | ------- | -------- | -------- | ------- | --------
before | 24304 | 7926 | 0 | 7086 | 1181 | 64%
after | 24246 | 7879 | 0 | 7066 | 1179 | 64%

### DIFF
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2011, 2015, 2016, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2011, 2015, 2016, 2019, 2020
+#                Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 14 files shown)


DougBurke:
There ya go @dtnguyen2 - have some 2020s.

Issue #731, opened 2020-01-17
> perhaps the following code


dtnguyen2:
> perhaps the following code
> 
> ```
>     def teardown(self):
>         self._use_caching = False
> ```
> 
> in the file _models/model.py_ should be changed to:
> 
> ```
>    def teardown(self):
>        self._use_caching = True
> ```
> 
> also

I personally don't have a good feel for how the cache code and the setup/teardown methods are meant to interact. Previously I think the cache was only used for models with no free parameters, and then only within the context of a "setup/teardown" call - e.g. as done by a fit call. This is presumably why there is a unilateral "turn off cacheing" in the teardown method, in case it had been turned on by the setup method.

 We've now changed things so that cacheing is expected to be on, so turning it off in teardown isn't very useful. However, I am not convinced that always turning it on is a great idea either. My current feeling would be that the teardown method shouldn't change the `_use_caching` attribute. In that case we may also want to remove the "turn on caching if there are no free parameters" logic in the setup method as well (since we expect the cache support to already be on), but I think this is getting to the point that we should be reviewing the cache code.

_Originally posted by @DougBurke in https://github.com/sherpa/sherpa/pull/678#issuecomment-532354706_

dtnguyen2:
While working on PR #728 (a fix to issue #717), I decided to take a closer look at the following section of the code
```
    def teardown(self):
        self._use_caching = False
```
First of all, the current code does _not_ pass the runtime cache option in the ```fit()``` command properly.  This can be seen with the following modifications:
1) Add a simple print of the ```_use_caching``` value in the sherpa/models/model.py file:
```
 (sherpa) [dtn@devel12 sherpa]$ git diff sherpa/models/model.py
diff --git a/sherpa/models/model.py b/sherpa/models/model.py
index 5e4c0bf4..f465554a 100644
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -49,6 +49,7 @@ def modelCacher1d(func):
 
     def cache_model(cls, pars, xlo, *args, **kwargs):
         use_caching = cls._use_caching
+        print('from cache_model: use_caching =', use_caching)
         cache = cls._cache
         queue = cls._queue
```
[...]

PR #733, opened 2020-01-22
fix a bug with fit(cache=False) passing the runtime option while fitting

dtnguyen2:
# Note

This PR fixes the issue at the following comment https://github.com/sherpa/sherpa/issues/731#issuecomment-576873292
of issue #731

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 #
-#  Copyright (C) 2010, 2016, 2017, 2018, 2019
+#  Copyright (C) 2010, 2016, 2017, 2018, 2019, 2020
 #      Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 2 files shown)


sherpa/ui/tests/test_ui.py
```diff
@@ -519,3 +526,38 @@ def test_user_model1d_fit():
...
+        return fvec
+
+    def __init__(self, name='myexp'):
+        self.A = Parameter(name, 'A', 1)
+        self.mylambda = Parameter(name, 'mylambda', 2)
+        self.b = Parameter(name, 'b', 3)
+        self.counter = 0
+        ArithmeticModel.__init__(self, name, (self.A, self.mylambda, self.b))
+
+@pytest.mark.usefixtures("clean_ui")
```
DougBurke: We don't need this line since the test doesn't use the ui module.

sherpa/models/model.py
```diff
@@ -543,6 +543,7 @@ def __getitem__(self, filter):
     def startup(self, cache):
         self._queue = ['']
         self._cache = {}
+        self._use_caching = cache
         if int(self.cache) > 0:
```
DougBurke: So, it looks like we have two "variables" to control the cache: the `cache` argument to `startup`, which says "do we cache or not", and the `cache` attribute of the model object, which defines how many items are in the cache (by defining the length of the `_queue` list). Given that we now require python 3.5 at a minimum, we could look at using the [lru_cache](https://docs.python.org/3.5/library/functools.html#functools.lru_cache) decorator rather than writing our own, but that's outside the scope of this PR.

PR #736, opened 2020-01-24
WIP explore test setup as 'python setup.py test' is going away

DougBurke:
Support for 'python setup.py test' is going away, so switch to
using py.test here (it is fortunately an easy change).


DougBurke:
Of course it couldn't be that simple...

DougBurke:
Replaced by work in #784

PR #740, opened 2020-01-31
resample_data: when the error range on a bin includes -1 and no-longer restrict sampling to +/-1 sigma

DougBurke:
# Summary

Fix `resample_data`/`ReSampleData` so that it correctly handles bins where the range of data values (i.e. low to high limit) includes the value -1.

The resampling was restricted to values that lay within the error bounds (ie +/-1 sigma) which has now been lifted. This fixes issue #779 

The return value now includes an entry called "samples" which contains the full data used in the analysis, and NumPy arrays are now used for returning the per-iteration parameter values.

The sampling method now ensures that each iteration starts at the original "best-fit" location (previously it would depend on the previous fit). For well-defined search spaces this will not change the statistical results (e.g. the mean and standard-deviation of the parameter values).

# Details

The current implementation does not work if the error range (i.e. low to hi) for a bin includes the value -1. I have added a test case where each bin has this range, and then fit with a constant model. Each iteration returns -1.

The first commit is expected to fail since it adds the test case. The second commit is white-space (and noting this PR number in the test case for reference). The third commit is where the fix is (assuming the tests pass).

The code is then changed to remove the +/-1 sigma limit on the resampling (issue #779) and the test suite has then been switched over to pytest style.

Further changes add the 'samples' return value and tweaking the internals.


DougBurke:
I've rebased this onto the current master branch (to pick up some python 3.8 changes and the like), as well as adding a fix for #779

DougBurke:
The **only** failed test for the last commit - 515472b  - is in the doc build, and is because of a problem installing packages with conda - see https://travis-ci.org/github/sherpa/sherpa/jobs/675469537

I would restart this one build but I seem to have lost the ability, and I'm planning to add another commit so it doesn't seem worth it.

DougBurke:
To answer https://github.com/sherpa/sherpa/pull/740#pullrequestreview-403911707 - I added extra code to return the actual distribution during testing.

DougBurke:
@anetasie - I gave an example of the change I used to view the pixel distributions in #779 but here it is again. Take this PR and apply the following patch:

```
diff --git a/sherpa/sim/__init__.py b/sherpa/sim/__init__.py
index c0d37dac..d8cda471 100644
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -847,6 +847,7 @@ class ReSampleData(NoNewAttributesAfterInit):
             msg ="{0} {1}".format(ReSampleData.__name__, type(data))
             raise NotImplementedError(msg)

+        ry_all = []
         numpy.random.seed(seed)
         for j in range(niter):
             ry = []
@@ -893,6 +894,7 @@ class ReSampleData(NoNewAttributesAfterInit):
                 ry.append(r)

             ry = numpy.asarray(ry)
+            ry_all.append(ry)
             # fit is performed for each simulated data point
             fit = Fit(Data1D('tmp', x, ry), self.model, LeastSq(), LevMar())
             fit_result = fit.fit()
@@ -908,4 +910,5 @@ class ReSampleData(NoNewAttributesAfterInit):
             print(name, ': avg =', avg, ', std =', std)
             result[name] = pars[name]

+        result['_pixel_data'] = numpy.asarray(ry_all)
         return result
```

and then you get back an entry `_pixel_data` which is a 2D array (niterations by data coordinate). For example you can use it with

```
import numpy as np

from sherpa.data import Data1DAsymmetricErrs
from sherpa.sim import ReSampleData

mdl = Const1D()
mdl.c0 = 10

x = np.asarray([1, 3, 5, 7, 9])
y = np.asarray([11, 12, 9, 9, 13])
dy = np.asarray([2, 4, 10, 1, 10])

d = Data1DAsymmetricErrs('test', x, y, dy, dy / 2)

sim = ReSampleData(d, mdl)

res = sim(seed=7329, niter=1000)
ys = res['_pixel_data']

plt.clf()
for i in range(5):
    lbl = 'y={} -{} +{}'.format(y[i], dy[i], dy[i]/2)
    plt.hist(ys[:, i], alpha=0.5, label=lbl)

plt.xlim(-2, 19)
plt.ylim(0, 200)
plt.gca().set_xticks(range(0, 22, 2))
plt.xlabel('Dependent variable')
plt.legend()
plt.savefig('yvals.png')
print("Created: yvals.png")
```

to create

yvals

DougBurke:
A question for @anetasie - would we want the return value from `resample_data` to include the faked arrays (e.g. a niter by num-bins array)? That is, provide an official version of my hack above?

On the plus side, people might find something interesting to do with the data. On the negative side it could use up memory. It could be made optional.

anetasie:
@DougBurke Yes, it would be good to have the option of returning the resampled data arrays.

DougBurke:

DougBurke:
There have been four changes since @anetasie last looked:

1) the return value now includes a 'samples' key which contains a 2D NumPy array of size niter by num-points that contains the resampled data

2) the parameter values are now returned as a NumPy array rather than a list (given Python, this should not affect people's code but there's always some corner case)

3) each iteration now starts at the "best-fit" location (that is, the parameters of the model that you send in to the resampling method) whereas previously it would start at the best-fit location of the previous iteration. Under the assumption that we are going to be close to the best-fit answer than this should not change results (at least once error values are included)

A low-level routine (i.e. not something sherpa.astro.ui users would call directly) now no-longer changes the model parameter values (that is, we ensure the parameters get reset). This has just changed where this check was being implemented.

4) minor documentation additions

anetasie:
@DougBurke Thanks. I'll take a look.

DougBurke:
I'm wondering if we should also return the statistic value of each iteration.

hamogu:
Yes, please return the statistic value.

anetasie:
@DougBurke Yes, statistic value would be good.

DougBurke:
Done

DougBurke:
Rebased to get the codecov changes

dtnguyen2 (approved):

DougBurke:
@anetasie - as Dan has approved this can you check and sign off (or note any problems)?

DougBurke:
@anetasie - can you approve this?

anetasie:
@DougBurke it's good to go.

Issue #638, opened 2019-07-12
gauss2d models need to have the fwhm/sigma bounds updated by guess


anetasie:
Currently `guess()` does not updated the fwhm or sigma parameters in `gauss2d, normgauss2d, sigmagauss2d` models, so the bounds in fitting are set to the standard 1e-38 and 1e38.  This makes the optimization code wonder off. 

In 1D `gauss models` the `fwhm` is set by guess to a valid range of the data, e.g., energy 0.1-10 keV for X-rays.  In 2D the (x,y) range of the image data would be a valid option and will improve the optimization performance.

PR #741, opened 2020-02-12
Fix issue 368, guess needs to update fwhm/sigma bounds

dtnguyen2:
# Details

Add support for guessing the fwhm or sigma parameters of the Gauss2D, NormGauss2D, and SigmaGauss2D models.

# Note

This PR fixes issue #638 (gauss2d models need to have the fwhm/sigma bounds updated by guess)

### DIFF
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1564,6 +1564,7 @@ class SigmaGauss2D(Gauss2D):
         param_apply_limits(ypos, self.ypos, **kwargs)
         param_apply_limits(fwhm, self.sigma_a, **kwargs)
         param_apply_limits(fwhm, self.sigma_b, **kwargs)
+        param_apply_limits(norm, self.ampl, **kwargs)
 
     def calc(self, *args, **kwargs):
         kwargs['integrate'] = bool_cast(self.integrate)
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 #
-#  Copyright (C) 2010, 2016, 2018, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2010, 2016, 2018, 2019, 2020
+#       Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -1478,9 +1479,11 @@ class Gauss2D(RegriddableModel2D):
     def guess(self, dep, *args, **kwargs):
         xpos, ypos = guess_position(dep, *args)
         norm = guess_amplitude2d(dep, *args)
+        fwhm = guess_fwhm(dep, *args)
         param_apply_limits(xpos, self.xpos, **kwargs)
         param_apply_limits(ypos, self.ypos, **kwargs)
         param_apply_limits(norm, self.ampl, **kwargs)
+        param_apply_limits(fwhm, self.fwhm, **kwargs)
 
     def calc(self, *args, **kwargs):
         kwargs['integrate'] = bool_cast(self.integrate)
@@ -1552,6 +1555,15 @@ class SigmaGauss2D(Gauss2D):
                                  (self.sigma_a, self.sigma_b, self.xpos,
                                   self.ypos, self.theta, self.ampl))
         self.cache = 0
+
+    def guess(self, dep, *args, **kwargs):
+        xpos, ypos = guess_position(dep, *args)
+        norm = guess_amplitude2d(dep, *args)
+        fwhm = guess_fwhm(dep, *args)
+        param_apply_limits(xpos, self.xpos, **kwargs)
+        param_apply_limits(ypos, self.ypos, **kwargs)
+        param_apply_limits(fwhm, self.sigma_a, **kwargs)
+        param_apply_limits(fwhm, self.sigma_b, **kwargs)
 
     def calc(self, *args, **kwargs):
         kwargs['integrate'] = bool_cast(self.integrate)
@@ -1640,8 +1652,10 @@ class NormGauss2D(RegriddableModel2D):
     def guess(self, dep, *args, **kwargs):
         xpos, ypos = guess_position(dep, *args)
         ampl = guess_amplitude2d(dep, *args)
+        fwhm = guess_fwhm(dep, *args)
         param_apply_limits(xpos, self.xpos, **kwargs)
         param_apply_limits(ypos, self.ypos, **kwargs)
+        param_apply_limits(fwhm, self.fwhm, **kwargs)
 
         # Apply normalization factor to guessed amplitude
         norm = (numpy.pi / _gfactor) * self.fwhm.val * self.fwhm.val * \
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019
+#  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019, 2020
 #     Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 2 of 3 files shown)


sherpa/models/basic.py
```diff
@@ -1553,6 +1556,15 @@ def __init__(self, name='sigmagauss2d'):
                                   self.ypos, self.theta, self.ampl))
         self.cache = 0
 
+    def guess(self, dep, *args, **kwargs):
+        xpos, ypos = guess_position(dep, *args)
+        norm = guess_amplitude2d(dep, *args)
```
DougBurke: This isn't used. I think we would want the ampl to be guessed too (it is in Gauss2D).

sherpa/models/basic.py
```diff
@@ -1640,8 +1652,10 @@ def set_center(self, xpos, ypos, *args, **kwargs):
     def guess(self, dep, *args, **kwargs):
         xpos, ypos = guess_position(dep, *args)
         ampl = guess_amplitude2d(dep, *args)
```
DougBurke: I think we would want this to be used to guess the ampl/norm/whatever parameter.

DougBurke:
The two comments aren't strictly related to the bug report (ampl/norm rather than sigma), but we should either drop the un-needed calculation or also guess the amplitude (which we do for Gauss2D, so it would make sense to do so). The NormGauss2D case is slightly-different, since the normalisation is not the max point but the integrated signal, so the calculation would need to be more complex.

sherpa/models/basic.py
dtnguyen2: Added the guessed ampl:
```
(sherpa) [dtn@devel12 sherpa]$ git diff models/basic.py
diff --git a/sherpa/models/basic.py b/sherpa/models/basic.py
index 2e0784e1..799fed36 100644
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1564,6 +1564,7 @@ class SigmaGauss2D(Gauss2D):
         param_apply_limits(ypos, self.ypos, **kwargs)
         param_apply_limits(fwhm, self.sigma_a, **kwargs)
         param_apply_limits(fwhm, self.sigma_b, **kwargs)
+        param_apply_limits(norm, self.ampl, **kwargs)
 
         kwargs['integrate'] = bool_cast(self.integrate)
```

sherpa/models/basic.py
dtnguyen2: Not sure what you mean here, doesn't the following existing code:
```
        # Apply normalization factor to guessed amplitude
        norm = (numpy.pi / _gfactor) * self.fwhm.val * self.fwhm.val * \
            numpy.sqrt(1.0 - (self.ellip.val * self.ellip.val))
        for key in ampl.keys():
            if ampl[key] is not None:
                ampl[key] *= norm
        param_apply_limits(ampl, self.ampl, **kwargs)
```
attempts at guessing the ```ampl``` parameter?

DougBurke: I think I must have just missed that part, as it was separated by a pesky blank line, which threw me off. He guesses wildly.

DougBurke:
Here's the differences in parameter values that we now see from using guess for `gauss2d`, `sigmagauss2d`, and `normgauss2d` (the "-/+" lines show the changes to the parameter values made by guess for a particular dataset):

```
# Diffs for gauss2d.gmdl
-    gmdl.fwhm    thawed           10  1.17549e-38  3.40282e+38
+    gmdl.fwhm    thawed            8        0.008         8000
-    gmdl.xpos    thawed            0 -3.40282e+38  3.40282e+38
+    gmdl.xpos    thawed           16           10           20
-    gmdl.ypos    thawed            0 -3.40282e+38  3.40282e+38
+    gmdl.ypos    thawed          106          100          110
-    gmdl.ampl    thawed            1 -3.40282e+38  3.40282e+38
+    gmdl.ampl    thawed      9.88079   0.00988079      9880.79

# Diffs for sigmagauss2d.sgmdl
-    sgmdl.sigma_a thawed           10  1.17549e-38  3.40282e+38
+    sgmdl.sigma_a thawed            8        0.008         8000
-    sgmdl.sigma_b thawed           10  1.17549e-38  3.40282e+38
+    sgmdl.sigma_b thawed            8        0.008         8000
-    sgmdl.xpos   thawed            0 -3.40282e+38  3.40282e+38
+    sgmdl.xpos   thawed           16           10           20
-    sgmdl.ypos   thawed            0 -3.40282e+38  3.40282e+38
+    sgmdl.ypos   thawed          106          100          110
-    sgmdl.ampl   thawed            1 -3.40282e+38  3.40282e+38
+    sgmdl.ampl   thawed      9.88079   0.00988079      9880.79

# Diffs for normgauss2d.ngmdl
-    ngmdl.fwhm   thawed           10  1.17549e-38  3.40282e+38
+    ngmdl.fwhm   thawed            8        0.008         8000
-    ngmdl.xpos   thawed            0 -3.40282e+38  3.40282e+38
+    ngmdl.xpos   thawed           16           10           20
-    ngmdl.ypos   thawed            0 -3.40282e+38  3.40282e+38
+    ngmdl.ypos   thawed          106          100          110
-    ngmdl.ampl   thawed            1 -3.40282e+38  3.40282e+38
+    ngmdl.ampl   thawed      1866.33      1.86633  1.86633e+06
```

We can see the changes made in this PR if we compare to the CIAO 4.12 output

```
# Diffs for gauss2d.gmdl
-    gmdl.xpos    thawed            0 -3.40282e+38  3.40282e+38
+    gmdl.xpos    thawed           16           10           20
-    gmdl.ypos    thawed            0 -3.40282e+38  3.40282e+38
+    gmdl.ypos    thawed          106          100          110
-    gmdl.ampl    thawed            1 -3.40282e+38  3.40282e+38
+    gmdl.ampl    thawed      9.88079   0.00988079      9880.79

# Diffs for sigmagauss2d.sgmdl
-    sgmdl.xpos   thawed            0 -3.40282e+38  3.40282e+38
+    sgmdl.xpos   thawed           16           10           20
-    sgmdl.ypos   thawed            0 -3.40282e+38  3.40282e+38
+    sgmdl.ypos   thawed          106          100          110
-    sgmdl.ampl   thawed            1 -3.40282e+38  3.40282e+38
+    sgmdl.ampl   thawed      9.88079   0.00988079      9880.79

# Diffs for normgauss2d.ngmdl
-    ngmdl.xpos   thawed            0 -3.40282e+38  3.40282e+38
+    ngmdl.xpos   thawed           16           10           20
-    ngmdl.ypos   thawed            0 -3.40282e+38  3.40282e+38
+    ngmdl.ypos   thawed          106          100          110
-    ngmdl.ampl   thawed            1 -3.40282e+38  3.40282e+38
+    ngmdl.ampl   thawed      2916.15      2.91615  2.91615e+06
```

DougBurke:
One question: we use the same scaling/values for the FWHM parameters (`gauss2d`, `normgauss2d`) as we do for the sigma parameters (`sigmagauss2d`). This is shown in my previous comment, where the `fwhm` and `sigma_a/b` parameters were both set to 8 (with a range of 0.008 to 8000).

Ideally the sigma values would be 2 . sqrt(2 . log(2)) times smaller than the FWHM values.

However, given how simplistic our guess is for this parameter, I am not sure it's worth the extra complication.

The reason why I noted this was I was comparing the predictions made by the guessed models since my assumption was they should be the same, but they're not

a) because of this, the sigmagauss2d variant is different to the other two
b) the normgauss2d "normalization" refers to the integration over an infinite plane, and the scaling used to guess it's normalization doesn't account for the finite nature of the dataset, so there's a constant multiplicative factor for each pixel when comparing the normgauss2d to gauss2d predictions. I definitely do not think this is worth "trying to correct".

DougBurke:
[also, the macOS build on Travis is having troubles finishing at the moment; let's hope this clears up]

DougBurke:
@anetasie I don't think we need to change the guess for `sigma_a/b` parameters of `sigmagauss2d` here. Do you agree (with reference to  https://github.com/sherpa/sherpa/pull/741#issuecomment-615365142 ) ?

anetasie:
@DougBurke I agree we don't need to change the guess for `sigma_a/b` parameters.

Issue #742, opened 2020-02-18
linked params do not follow parameter limits: was segfault with linked parameters + atables


JohannesBuchner:
I seem to be getting segfaults when setting up a atable (with load_table_model), and linking one of its parameters to another parameter of another component. When that parameter exceeds the limit of the table parameters, it tries to look up data that is out of bounds, leading to a segfault.
The correct behavior should prevent the table from exceeding its parameter limits.

dtnguyen2:
Can you make script and data available, thanks in advance

JohannesBuchner:
This script should reproduce the issue with any X-ray spectrum file.
```
from sherpa.astro.ui import *
import os
import numpy

filename = 'g20_a.pha'
load_pha(filename)

# from https://zenodo.org/record/2245189
load_table_model('blob', '/home/user/bin/xrayspectralmodels/clumpybox3.fits')
set_model(blob + xspowerlaw.p * xstbabs.abso)
blob.norm = 1e-10

from sherpa.models.parameter import Parameter
# make a parameter with wide limits
srcnh = Parameter('src', 'nh', 22, 19, 27, 19, 27)
# link to one component
xstbabs.abso.nh = 10**(srcnh - 22)
print(get_model())

print("test with single link ...")
print(calc_stat())
print("test with double link ...")
# link to a link, but blob.nh does not support that wide of a range
blob.nh.link = xstbabs.abso.nh
print(calc_stat()) # --segfault occurs here

```

dtnguyen2:
I got tired of waiting to download your large _clumpybox3.fits_ file and substitute a fits file that I have around but the results, apart from the output of the ```calc_stat```, should not affect the issue you are asking.  Anyway, not sure how your script manages to get the following line to work:
```
blob.norm = 1e-10
```
because sherpa's models usually use ```ampl``` instead of ```norm```;  Moreover, notice also that blob does not have parameter  called ```nh``` either.  I double checked sherpa's class definition of the TableModel:
```
class TableModel(ArithmeticModel):
    def __init__(self, name='tablemodel'):
        # these attributes should remain somewhat private
        # as not to conflict with user defined parameter names
        self.__x = None
        self.__y = None
        self.__filtered_y = None
        self.filename = None
        self.method = linear_interp  # interpolation method
        self.ampl = Parameter(name, 'ampl', 1)
        ArithmeticModel.__init__(self, name, (self.ampl,))
...
(Pdb) blob
[...]

PR #743, opened 2020-02-21
xspec table models: add out-of-bound check to avoid segfault

JohannesBuchner:
This PR tries to address issue #742 .

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -951,6 +951,10 @@ class XSTableModel(XSModel):
         # logic could be in __init__, but that can be changed
         # at a later date.
         #
+        
+        assert all((param.min <= value <= param.max 
+            for value, param in zip(self.pars, p))), "model parameters outside bounds"
+        
         if hasattr(_xspec, 'tabint'):
             tabtype = 'add' if self.addmodel else 'mul'
             return _xspec.tabint(p,
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -953,7 +953,7 @@ class XSTableModel(XSModel):
         #
         
         assert all((param.min <= value <= param.max 
-            for value, param in zip(self.pars, p))), "model parameters outside bounds"
+            for param, value in zip(self.pars, p))), "model parameters outside bounds"
         
         if hasattr(_xspec, 'tabint'):
             tabtype = 'add' if self.addmodel else 'mul'


JohannesBuchner:
The code submitted to issue #742 just after the bug report should trigger the assert. The table model parameter limits are not modified.

dtnguyen2:
The following code:
```
(sherpa) [dtn@devel12 sherpa]$ git diff sherpa/astro/xspec/__init__.py
diff --git a/sherpa/astro/xspec/__init__.py b/sherpa/astro/xspec/__init__.py
index 689f6fe6..cbe4e40b 100644
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -951,6 +951,8 @@ class XSTableModel(XSModel):
         # logic could be in __init__, but that can be changed
         # at a later date.
         #
+        assert all((param.default_min <= value <= param.default_max 
+                    for param, value in zip(self.pars, p))), "model parameters outside bounds"
         if hasattr(_xspec, 'tabint'):
             tabtype = 'add' if self.addmodel else 'mul'
             return _xspec.tabint(p,
```
will trigger the assertion with the original code _and_ the one with the limits changed.  Perhaps, one can be more general with the check with default_min and min but are we going to put this check in all the models that define ```calc``` wouldn't it be better to centralize it somehow?

JohannesBuchner:
A atable fits file can specify a default_min/max with narrower ranges than allowed by the table model grid steps.

hard_min/max may be more useful to look at. However, min/max cannot be changed to exceed hard_min/max, so the PR at this stage is catching this case. The only thing it does not catch is someone changing hard_min/max, but I think if one overrides hard_min/max one shouldn't be surprised to get bad behaviour.

Are you planning to add variations of your script as a test? I couldn't immediately find a test with an atable that calls calc().

I think there are some general issues with the INSTALL_TYPE=develop build.

Regarding generalising to other models -- I don't have an opinion on that. Maybe they already check this in other ways or inadvertently raise exceptions.

DougBurke:
I have a "general" fix in place for all models in #745, but this does not catch the "I expect not many people to do this" use case of calling a models `call` method explicitly. That is, if `mdl` is the model instance (whether an XSPEC table model or a Sherpa-native model),

    mdl(grid)

will error out now if a linked parameter is outside it's min/max range (well, once #745 is applied it will). If a user says

    mdl.call(parvals, grid)

then it is still possible (with #745) to specify an invalid value and cause a crash. @JohannesBuchner fix here would catch this case (but only for XSPEC table models).

As I mentioned, I think the `call` method is a low-level primitive, so we could say that a user is expected to know what they are doing here (that is, we don't enforce any checks in the `call` method). However, we know that for XSPEC table models at least we really should enforce the parameter limits (as much as possible), otherwise we may well crash the Python process. I am not sure I want to have a safety check on the parameter-limits on all `call` methods (there may be times when you may want to call a model outside the min/max range when you are in "low level" code).

So my suggestion is that we accept #743 but do not generalize it beyond the XSPEC table-model case.

sherpa/astro/xspec/__init__.py
```diff
@@ -953,7 +953,7 @@ def calc(self, p, *args, **kwargs):
         #
         
         assert all((param.min <= value <= param.max 
-            for value, param in zip(self.pars, p))), "model parameters outside bounds"
+            for param, value in zip(self.pars, p))), "model parameters outside bounds"
```
DougBurke: I think this would be better as (untested), since it matches the "normal" parameter error for an out-of-bands setting. I have also chose to use the hard bounds here (in XSPEC table-model terms, the `MINIMUM` and `MAXIMUM` values rather than the `TOP` and `BOTTOM` values), since this is a low-level function.

```
for param, value in zip(self.pars, p):
    if value < param._hard_min:
        raise ParameterErr('edge', self.name, 'minimum', param._hard_min)
    if val > param._hard_max:
        raise ParameterErr('edge', self.name, 'maximum', param._hard_max)
````

DougBurke:
A test can be added based on `sherpa/astro/xspec/tests/test_xspec_unit.py::test_xstbl_link_parameter_evaluation` (from #745).

JohannesBuchner:
I would be glad if you go ahead with the work on the PR.

hamogu:
Superceeded by #845 845

PR #745, opened 2020-02-24
ensure that min/max limits are applied to linked parameters before use

### LINKED ISSUE #742: linked params do not follow parameter limits: was segfault with linked parameters + atables

DougBurke:
# Summary

Enforce parameter restrictions (that they must lie within their min/max range) even when the parameter is linked to another (which has a different range). This fixes #742.

# Important note

With this PR, and a model called `mdl` containing a parameter with an invalid value,

    mdl(grid)

will raise a `ParameterErr`. 
However, you can still call the model with any parameter values by using the `call` method, e.g.

    mdl.call(parvals, egrid)

where the `parvals` sequence is not checked. PR #743 would catch this case (but only for XSPEC table models). Note that the `call` method is not expected to be commonly used.

# Details

The first commit adds tests that show the current behavior (that is, the parameter can be evaluated outside it's min/max range). Then the fix is made, but the tests are not updated (so are expected to fail). This picks up an existing test which decided to use an out-of-band value for the link. Examples of these failures can be seen at https://travis-ci.org/sherpa/sherpa/jobs/654689948#L4127  (from an earlier version of this PR before a rebase). The third commit updates the tests to use a more-sensible value (for the existing test), or to check for the `ParameterErr` in the new tests. The fourth commit adds an explicit XSPEC table-model test (to verify the original bug fix).

The check is made when the parameter value is accessed, not when the linked-to parameter is changed, since this is the simplest way to address the issue. Any other way would likely be fragile and *much* more complicated. The implementation can lead to slightly-surprising error messages - e.g. evaluating the model will raise a `ParameterErr` as shown in the example below - but I believe this is reasonable.

```
In [1]: from sherpa.models.basic import PowLaw1D, Const1D                       

In [2]: m1, m2 = PowLaw1D(), Const1D()                                          

In [3]: m1.gamma = m2.c0                                                        

In [4]: print(m1.gamma)                                                         
val         = 1.0
min         = -10.0
max         = 10.0
units       = 
frozen      = True
link        = const1d.c0
default_val = 1.0
default_min = -10.0
default_max = 10.0

In [5]: print(m2.c0)                                                            
val         = 1.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
default_val = 1.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38

In [6]: m2.c0 = 20                                                              

In [7]: m1([1, 2, 3])                                                           
ParameterErr                              Traceback (most recent call last)
<ipython-input-7-bfe3cdbf28bf> in <module>
----> 1 m1([1, 2, 3])

/lagado.real/sherpa/sherpa-playground/sherpa/models/model.py in __call__(self, *args, **kwargs)
    263             return self
    265 

/lagado.real/sherpa/sherpa-playground/sherpa/models/model.py in <listcomp>(.0)
    263             return self
    265 

/lagado.real/sherpa/sherpa-playground/sherpa/models/parameter.py in _get_val(self)
    161             raise ParameterErr('edge', self.fullname, 'minimum', self.min)
    162         if val > self.max:
--> 163             raise ParameterErr('edge', self.fullname, 'maximum', self.max)
    164 
    165         return val

ParameterErr: parameter powlaw1d.gamma has a maximum of 10
```

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016-2018, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016-2018, 2019, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -30,7 +30,9 @@ import numpy as np
 from numpy.testing import assert_almost_equal
 
 from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
+from sherpa.utils.err import ParameterErr
 
+from sherpa.models.basic import Const1D
 
 # It is hard to test many of the state routines, since it requires
 # a full understanding of how they are implemented; the simplest
@@ -90,7 +92,9 @@ def test_chatter_default():
 
     For some reason the chatter setting appears to default to 50,
     rather than 0. It is not at all clear why, so has been marked
-    as xfail.
+    as xfail. DJB now thinks that this is due to the Sherpa test
+    setup, which is designed to make the models be verbose in
+    case there's an error.
 
     """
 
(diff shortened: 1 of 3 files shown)


JohannesBuchner:
The alternative to erring out is to stay at the maximum and print a warning. I think this is the xspec behaviour. 

This can be useful when combining (adding) a power-law with a table model with a limited photon index range, with the powerlaw indices linked. Having the power-law photon index go outside the table's range, and the table model adopting its extreme value, can be a useful approximation.

DougBurke:
I have a feeling we use these parameter errors as "control flow" in various places (i.e. catch them and handle them in some way), so I'm a bit wary about changing the behavior of min/max ranges at this time. I could well be mistaken though.

Issue #404, opened 2017-10-23
sample_flux for multiple data sets


anetasie:
sample_flux uses covar() to set the size of the multivariate gaussian for the simulations. In the current implementation covar() is run on only 1 data set. This is incorrect if the fit was performed on the multiple data sets. sample_flux() needs to expanded to use multiple ids as an input to the function call and then calculate covar/conf using these ids.

DougBurke:
I was looking at this last week, and it's not clear to me what is actually going on, as I originally was worried about this, but now am not so sure. Internally `DataSimulFit` and `SimulFitModel` instances are created, based on the loaded data, and passed to the sampler.

I am concerned that if you have multiple datasets loaded but only want to use a subset then it doesn't work.

Also, this code is somewhat hard to follow, so I could well be mis-reading it.

anetasie:
At this moment only 1 data id is allowed as a subset, which is a limitation, so it is all used for the simultaneous fit or only 1.

olaurino:
Is this issue still relevant?

DougBurke:
I believe it is still relevant. Say I have 3 datasets loaded and I only want the errors on the flux using datasets 2 and 3. I do not think we can do this with the current `sample_flux` function. We could just say in the documentation "don't do this", but I expect we want to support this use case.

kglotfelty:
Just a note that we had a user hit this problem and send in a report via cxc helpdesk (ticket 21004).

> I have 4 observations of a source in Haro 11 (spanning years 2006-2017) and I want to see if the flux has changed. The spectrum looks quite different at the low energies, but contaminant.
> 
> I fit a simple model to each spectrum (xswabs*powlaw) and ran sample_flux to get an idea of the flux plus uncertainty.
> 
> sherpa-60> flux1=sample_flux(comp1,lo=0.3,hi=8.0,Xrays=True,correlated=True,num=100)
> original model flux = 9.1421e-14, + 9.34095e-15, - 7.30664e-15
> model component flux = 9.1421e-14, + 9.34095e-15, - 7.30664e-15
> sherpa-61> flux2=sample_flux(comp2,lo=0.3,hi=8.0,Xrays=True,correlated=True,num=100)
> original model flux = 9.34246e-14, + 8.08303e-15, - 9.67405e-15
> model component flux = 1.01498e-13, + 1.48371e-14, - 1.82864e-14
[...]

PR #754, opened 2020-03-09
sample_flux now returns statistic values for each row

### LINKED ISSUE #751: sample_flux does not fully populate the flux array
I have a case where the flux segment of the `sample_flux` output (the last column of the third returned value) is only partially filled. This is because the code to loop over this array is limited to those with parameter values within the "allowed" range, which is less than the number of iterations (I'm looking at the code and will suggest some clean ups and changes for this).

In the plot below, see how the last trace (which plots the last column) drops to 0 after ~7200 iterations.

```python
f1, f2, vals = ui.sample_flux(id=1, lo=0.5, hi=7, num=10000, correlated=True, scales=covar)
labels = ['flux', 'nH', r'$\gamma$', 'ampl', r'$\chi^2$']

plt.figure(figsize=(8, 10))
for i, lbl in enumerate(labels, 1):
    plt.subplot(5, 1, i)
    plt.plot(vals[:, i - 1])
    plt.ylabel(lbl)

plt.xlabel('Iteration');
```

creates

wrong

DougBurke:
# Summary

The sample_flux command now returns a statistic value for each iteration, even if those rows are not used in the reported flux distribution. Fixes #751.

# Details

There are some associated code clean ups that reduce the amount of information sent to sherpa.astro.flux.calc_sample_flux, and reduce the complexity of the code. The main functional change is that we now loop through all samples, so we can calculate the statistic for each sample. Previously this was only done for those rows that were not "clipped", and so the mapping from parameter row to sample was lost (i.e. you could only guarantee that this held was when no rows were "clipped").

# Not addressed here

Just to remind myself of what still needs looking at (this was my original PR which was much larger and complex but has slowly been merged in pieces thanks to @hamogu so it's now much simpler).

 - Xrays=False (#755)
 - you can't get the flux distribution when modelcomponent is set (#795)
 - work on multiple datasets (#404)
 - what happens with background data sets (e.g flux for a background model); #1048 probably needs fixing first
 - what happens if you've used set_full_model (for this and the *_energy/photon_flux routines)?

### DIFF
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -527,7 +527,7 @@ def sample_flux(fit, data, src,
     return numpy.concatenate((vals, numpy.expand_dims(clipped, 1)), axis=1)
 
 
-def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
+def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
                      confidence):
     """Given a set of parameter samples, estimate the flux distribution.
 
@@ -542,21 +542,17 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
     .. versionchanged:: 4.13.1
        The clipping now includes parameters at the soft limits;
        previously they were excluded which could cause problems with
-       parameters for which we can only calculate an upper limit.
+       parameters for which we can only calculate an upper limit. The
+       id and session arguments have been removed.
 
     Parameters
     ----------
-    id : int or str
-        The dataset identifier. It must exist and have an associated model in
-        session.
     lo : number or None, optional
         The lower edge of the dataspace range for the flux calculation.
         If None then the lower edge of the data grid is used.
     hi : number or None, optional
         The upper edge of the dataspace range for the flux calculation.
         If None then the upper edge of the data grid is used.
-    session : sherpa.ui.utils.Session instance
-        Defines the data, model, and fit.
     fit : sherpa.fit.Fit instance
         The fit object. The src parameter is assumed to be a subset of
         the fit.model expression (to allow for calculating the flux of
@@ -665,11 +661,10 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
             oflx[nn] = mysim[nn, 0]
 
             for par, parval in zip(thawedpars, mysim[nn, 1:]):
-                session.set_par(par.fullname, parval)
+                par.set(parval)
 
             iflx[nn] = calc_energy_flux(data, modelcomponent, lo=lo, hi=hi)
-
-            mystat.append(session.calc_stat(id))
+            mystat.append(fit.calc_stat())
 
         logger.setLevel(orig_log_level)
 
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -641,22 +641,8 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
     #
     orig_model_vals = fit.model.thawedpars
 
-    logger = logging.getLogger("sherpa")
-    orig_log_level = logger.level
-
-    # only change the log level if it is less than error
-    #
-    if orig_log_level < logging.ERROR:
-        logger.setLevel(logging.ERROR)
-
-    # With the move to calling calc_energy_flux directly there should
-    # be no need to change the logging level, but leave in for now.
-    #
+    mystat = []
     try:
-
-        logger.setLevel(logging.ERROR)
-
-        mystat = []
         for nn in range(size):
             oflx[nn] = mysim[nn, 0]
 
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased to include changes in #756 (since some of them originated from this PR), and other changes to master.

anetasie:
@DougBurke Do you have a test for the case where the model component is a background model?
I was having trouble today (in the released version) to get the background model flux.

DougBurke:
@anetasie - I'm concentrating more on calc_energy_flux/sample_energy_flux (and the _photon_ variants) at the time being.

DougBurke:
I cancelled the test run to let other PRs get hold of travis resources, which  is why it is marked as having failed.

DougBurke:
Hmmm, the macOS tests are hitting the 50-minute time limit. This commit does add a number of tests, and perhaps their runtime is sufficient to cause the limit to be exceeded? The alternative is that there's a slow-down in some of the downloads.

DougBurke:
I've pulled a bunch of "unrelated" commits into #802 and #803 so this can remain focused on `sample_flux`

DougBurke:
Note: I don't want to work on this until #803 is accepted as they change the same files, so they could conflict (I had hoped they would't, but as time goes on this hope is fading)

DougBurke:
@anetasie - do you have an example of the problems you were having with the nackground flux (that you mentioned above in https://github.com/sherpa/sherpa/pull/754#issuecomment-620292184 )?

anetasie:
@DougBurke - I need to check to the analysis that I was doing in April. This is a long time ago.

DougBurke:
I have rebased this for all the changes, but have not tried to rework any of the code to take advantage of #803 and #866. This is just meant to be a start

DougBurke:
Rebased to pick up GH Actions.

DougBurke:
Rebased to pick up recent PRs (in particular running tests in parallel)

DougBurke:
AArggh. @hamogu is there something wrong somewhere in the python/conda-verse or is it just Monday? We have conda GitHub action builds failing because we need numpy, but

 a) these same builds ran earlier today
 b) 'conda list' (which I added in 2d57fc8) shows that numpy is installed - https://github.com/sherpa/sherpa/pull/754/checks?check_run_id=1857385695#step:5:47

I think I'm going to have a lie down...

hamogu:
I don't know, but I generally just wait 24 h and try again for something like this.

DougBurke:
I love computers. Just love them. They are so easy to understand...

sherpa/astro/flux.py
```diff
@@ -615,88 +614,55 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
...
-        logger.setLevel(orig_log_level)
+        fit.model.thawedpars = thawedpars
 
     hwidth = confidence / 2
     result = []
-    for x in [oflx, iflx]:
-        result.append(numpy.percentile(x, [50, 50 + hwidth, 50 - hwidth]))
+    for flx in [oflx, iflx]:
+        result.append(numpy.percentile(flx[valid],
+                                       [50, 50 + hwidth, 50 - hwidth]))
```
hamogu, line 658: So, `oflx` is just the same as `iflx`. Why have two different variables in that case?

hamogu (commented):
Just one inline comment where I don't understand something. Maybe it's correct, maybe not, I just don't know.

sherpa/astro/flux.py
```diff
@@ -615,88 +614,55 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
...
+    # a requirement for calling this routine, so is just handled as an
+    # assert (as this is not intended for a user-level routine)
     #
-    # For later restoration
-    #
-    orig_model = session.get_model(id)
-    orig_model_vals = fit.model.thawedpars
+    nthawed = len(thawedpars)
+    npars = samples.shape[1] - 2
+    assert nthawed == npars, (nthawed, npars)
```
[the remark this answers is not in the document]
DougBurke: I had put this in when I started out reworking `sample_flux`, to make sure I didn't accidentally break any iinvariants, and wonder if I should remove it now.

sherpa/astro/flux.py
DougBurke: They are only the same when modelcomponent is None. When modelcomponent is set iflx is set to the flux of modelcomponent. We could make this code copy over the oflx case when modelcomponent is not set, but I thought it would probably just be harder to understand, but I admit to not actually giving it much thought.

hamogu: That's OK, I just wanted to understand better.

sherpa/astro/flux.py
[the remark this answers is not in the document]
hamogu: Doesn't hurt, does it?

hamogu (approved):

PR #756, opened 2020-03-11
calculate_photon_flux/calculate_energy_flux fix and improvement

### LINKED ISSUE #308: calc_energy_flux docstring is confused over lo/hi behavior
The docstring for `sherpa.astro.ui.utils.calc_energy_flux` - shown below - says that when called with `hi=None` the upper limit of the data set is used (`Parameters` section), but in the `Returns` section it points out that if `lo` is not `None` but `hi=None` then an energy density (not flux) is returned. Even then, it is written in such a way that it only applies to X-ray style models. It is also not clear what happens when given a non-integrated data set/model.

```
        Parameters
        ----------
        lo : number, optional
           The minimum limit of the band. Use ``None``, the default,
           to use the low value of the data set.
        hi : number, optional
           The maximum limit of the band, which must be larger than
           ``lo``. Use ``None``, the default, to use the upper value of
           the data set.
        id : int or str, optional
           Use the source expression associated with this data set. If
           not given then the default identifier is used, as returned
           by `get_default_id`.
        bkg_id : int or str, optional
           If set, use the model associated with the given background
           component rather than the source model.
        Returns
        -------
        flux
           The flux from the source model integrated over the given
           band. This represents the flux from the model without any
           instrument response (i.e. the intrinsic flux of the
           source). For X-Spec style models the units will be
           erg/cm^2/s. If ``hi`` is ``None`` but ``lo`` is set then the flux
           density is returned at that point: erg/cm^2/s/keV or
           erg/cm^2/s/Angstrom depending on the analysis setting.
```

### LINKED ISSUE #619: The computed monochromatic fluxes are 0.0, issue with the mask computation?
Hello,

I am using Sherpa (4.11.0+0.g838bd12b.dirty) on an Anaconda environment with Python 3.6.7 and Numpy 1.15.4.

I encountered the following problem: when computing a monochromatic flux with the function calc_energy_flux [e.g. at 10 keV asking: calc_energy_flux(lo=10.,hi=None, id=data_id) ] the result is 0.

I believe I narrowed down the problem digging into the code here, that is what the function calls:

https://github.com/sherpa/sherpa/blob/41809de1a2a2f596deb87e3fdf3ccceb0902a6ca/sherpa/astro/utils/__init__.py#L116-L162

Line 121 causes "dim" (line 130) to be 2. There are 'if' conditions here and there accounting for this dim=2, but I think there should be one also for computing the mask (line 148).  The mask in my example has only "False". What I think it's happening, code-wise, is that there is no energy bin containing the monochromatic energy at which I want to compute the flux.
Line 148 calls this function:
https://github.com/sherpa/sherpa/blob/41809de1a2a2f596deb87e3fdf3ccceb0902a6ca/sherpa/utils/__init__.py#L991-L1055

where at line 1047 the mask is computed: I think that "axis"  here should be different for the match for 'lo" and "hi": in the first case it should be the array containing the low-end of the energy bins (axislist[0] at line 148 above), in the second case it should be the array containing the high-end (so axislist[1]). 

In my example, the mask is "False" because "lo"='hi'=10. (or whatever energy) and the "&" at 1047 is not satisfied. Maybe adding an "if" to compute the mask in a different way with dim=2 would solve it.

It is not a problem of monochromatic fluxes only. In fact, also computing a narrow-band flux with both extremes (lo and hi) within the same energy bin results in a flux=0, for the same problem.

I hope my message was not too confusing and thank you in advance for your help!

DougBurke:
# Summary

Bug fixes and improvements for calculate_photon_flux and calculate_energy_flux:

 - address flux and flux-density calculation issues discussed (fix #619)
 - fix documentation of the lo and hi arguments (fix #308)
 - add model parameter to calc_photon/energy_flux which allows users to easily calculate "unabsorbed" fluxes

# Details

Not listed in the summary section is that new tests have been added, both for the new functionality (checking the ability to calculate the flux of a model component), and for the existing functionality (to show the issues raised in #619).

## flux/flux density calculation (#619)

The problem with the existing code is that it doesn't really handle the case of 1D integrated cases (i.e. bins with low and high edges). The instrument grid would be filtered to select the bins to use for the calculation, but this did not include bins which were partially filled (e.g. for a grid of 0.5-0.6, 0.6-0.7, ... and a low edge of 0.55, the first bin would not be included). You can see the problems this causes in the PDF contained within https://github.com/sherpa/sherpa/issues/619#issuecomment-485779739 - in the test comments I note that before this change I have seen cases where the calculated fluxes were 80 to 90% of the expected value, and after the change are within 5% of the correct value (it depends very much on bin width and range).

For flux density calculations it was worse, since it only seems to have selected a value when the selected energy fell on a bin edge (so with the above grid you could get values for 0.5 or 0.6, but not for 0.55, when the routines would calculate 0).

The current approach taken in this PR is to treat partially filled bins via linear interpolation - that is, for a bin 0.5-0.6 and a low edge of 0.51 or 0.58, 90% or 20% of the flux would be included. To do this I have implemented my own routine for selecting bins from a 1D "integrated" grid (sherpa.astro.utils.range_overlap_1dint). I don't believe this functionality exists elsewhere, but I may be wrong. I also have written this with an over-abundance of caution (a lot of asserts) to ensure the preconitions I believe to be true are in fact true. The logic for searching for which bins the user limits fall within has been left to the numpy.digitize routine; this has the advantage of being a pre-canned routine, but does mean we lose the ability to use sao_fcmp to compare if a value is close to a bin edge (for example, the 3c273.pi RMF we use in many of our tests is nominally defined over the range 0.1 to 11 keV but the first bin actually starts at 0.10000000149011612 keV, which sao_fcmp thinks is the same, but numpy.digitize doesn't). This is primarily an issue for users who request a flux density at the very edge of the response.

The flux density calculation is now constant within a bin of the instrument response, rather than only being defined at the bin edges. This is not ideal, and can be improved upon as discussed in the NOTE section below (note 1).

Since these changes mean that the calc_photon/energy_flux results are different, the calc_kcorr test results need tweaking slightly to account for these changes.

**NOTE 1** I believe a better approach would be to ignore the instrument grid and create an appropriate one based on the user limits (i.e. essentially what is suggested in https://github.com/sherpa/sherpa/issues/619#issuecomment-499412354 ). This completely side steps edge effects and worrying about partial overlap of bins, as well as allowing the calculation of fluxes for energy ranges outside the instrument grid. It would however require changes to the API, since we would need a way of specifying the grid to use (both for fluxes and flux density, and you'd probably want to do different things for these two cases). I decided to try an improve the results first, so we can get an improved test suite, and then we can look at changing the method in a later PR.

**NOTE 2** I am generally unsure about the flux-density calculation (particularly for analysis=angstrom), as it is strongly dependent on the bin size.

**NOTE 3** if the user gave limits that fell within a single bin, the old code would turn that into a flux density, rather than a flux. This is *to me* exceedingly surprising (you don't always remember the binning of the response and we currently don't provide much guidance or information to the user about this grid). I have therefore switched this case to calculate a flux (using linear interpolation within the bin), and only calculatr a flux density when one of these conditions hold: lo=hi and it is not None, lo=value and hi is None, hi=value and lo is None (when lo/hi are within the instrument range).

**NOTE 4** I am pretty sure that eqwidth suffers from the same problems, but I have not looked at this. 

**NOTE 5** I wonder if we should just make calc_energy_flux and calc_photon_flux error out if called on anything but a DataPHA dataset. I have not added any testing of the 1D (non integrated) case here (as this code path shouldn't have changed), so don't know what it actually calculates in this case (particularly for the energy flux case). The calc_photon_flux routine is technically correct (in that it doesn't really make any assumptions about the axis units), it is just named confusingly for the generic case.

**NOTE 6** I don't understand the expected behavior when analysis=channel is selected. It looks like the band limits (or value at which to calculate the flux density) are taken to be in keV, but that surprises me (I expected channel). The tests for analysis=channel are therefore rather limited.

## Addition of the model argument

The addition of the model parameter will be useful for sample_flux, but that is not addressed here. Note that sample_flux uses the long argument name of modelcomponent but model is used here (which I am going to suggest sample_flux change to, and better matches the existing API).

This was an easy change, as it really only needed adding the user argument; the actual code that does the calculation did not need to be changed.

**Note** this functionality (i.e. the model argument) could be added to calc_source_sum (should be easy) and calc_model_sum (significantly harder).

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019
+#  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019, 2020
 #     Smithsonian Astrophysical Observatory
 #
 #
@@ -27,7 +27,6 @@ import pytest
 from sherpa.utils import _ncpus
 from sherpa.utils.testing import SherpaTestCase, requires_data, \
     requires_fits, requires_xspec, requires_group
-from sherpa import get_config
 import sherpa.astro.ui as ui
 from sherpa.astro.data import DataPHA
 
@@ -113,17 +112,20 @@ class test_threads(SherpaTestCase):
             assert calc == approx(-0.57731725, rel=1e-4)
 
             calc = ui.calc_kcorr([1, 1.2, 1.4, 1.6, 1.8, 2], 0.5, 2)
-            expected = [0.93341286, 0.93752836, 0.94325233,
-                        0.94990140, 0.95678054, 0.96393515]
+            # Prior to fixing #619 the expected values were
+            # expected = [0.93341286, 0.93752836, 0.94325233,
+            #             0.94990140, 0.95678054, 0.96393515]
+            expected = [0.93132747, 0.9352768, 0.94085917,
+                        0.94738472, 0.95415463, 0.96121113]
             assert calc == approx(expected, rel=1e-4)
-            
+
         self.run_thread('pha_intro')
         # astro.ui imported as ui, instead of
         # being in global namespace
         fit_results = ui.get_fit_results()
         covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
         cmp_pha_intro(fit_results, self.locals['p1'], covarerr)
-        
+
         self.run_thread('pha_intro_ncpus')
         # astro.ui imported as ui, instead of
         # being in global namespace
@@ -134,7 +136,7 @@ class test_threads(SherpaTestCase):
             assert fit_results.extra_output['num_parallel_map'] > 0
         else:
             assert fit_results.extra_output['num_parallel_map'] == 0
-            
+
     @requires_fits
     def test_pha_read(self):
         self.run_thread('pha_read')
@@ -160,8 +162,7 @@ class test_threads(SherpaTestCase):
             self.assertEqual(fit_results.nfev, 9)
             self.assertEqual(fit_results.numpoints, 11)
             self.assertEqual(fit_results.dof, 9)
-          
-            
+
         # In data1.dat for this test, there is a comment with one
         # word at the beginning -- deliberately would break when reading
         # with DM ASCII kernel, but passes because we have Sherpa code
@@ -171,7 +172,7 @@ class test_threads(SherpaTestCase):
         covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
         cmp_test_basic(fit_results, self.locals, covarerr)
         self.assertEqual(0, fit_results.extra_output['num_parallel_map'])
-        
+
         self.run_thread('basic_ncpus')
         fit_results = ui.get_fit_results()
         covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
@@ -180,7 +181,7 @@ class test_threads(SherpaTestCase):
             assert fit_results.extra_output['num_parallel_map'] > 0
         else:
             assert fit_results.extra_output['num_parallel_map'] == 0
-        
+
     @requires_fits
     @requires_xspec
     def test_simultaneous(self):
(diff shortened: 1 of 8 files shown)


DougBurke:
The 783d2d4 commit fails because I added tests that show the new behavior (handling of edge cases) rather than the old behavior for the flux density calculation.

DougBurke:
Changes since @dtnguyen2 review
 -  rebase to address a conflict (but didn't actually require any manual intervention) 
 - add 2020 copyright years to the files that Dan pointed to

DougBurke:
@dtnguyen2 - I've made the changes you requested. Can you check again, I'm craving all the approval I can get ;-)

DougBurke:
Ta

PR #758, opened 2020-03-13
Fix test break due to pytest 5.4.0 update (issue #757)

### LINKED ISSUE #757: some of our tests now fail due to pytest=5.4.0
Some of Omar's data-regridding tests are now failing because they rely on an "accidental" feature which has now been "fixed". See https://github.com/pytest-dev/pytest/issues/6909 for more details (and it's not just us that has been bitten by this).

For reference, the tests fail with a message like

```
==================================== test session starts =====================================
platform linux -- Python 3.8.1, pytest-5.4.0, py-1.8.1, pluggy-0.13.1
rootdir: /home/dburke
collected 2329 items / 1 error / 2328 selected

=========================================== ERRORS ===========================================
_ ERROR collecting anaconda/envs/sherpa-wtf/lib/python3.8/site-packages/sherpa-4.12.0+121.g2d36b10e.dirty-py3.8-linux-x86_64.egg/sherpa/models/tests/test_regrid_unit.py _
In function "test_evaluation_space2d_overlaps":
Parameter "x_overlaps" should be declared explicitly via indirect or in function itself
!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================== 1 error in 1.23s ======================================
```

It also does not hit all our test configurations on Travis; I don't know if that's just that the new version hasn't made it to all set ups yet or something more interesting on how we set things up.

DougBurke:
# Summary

Update tests to work with the pytest 5.4.0 release. Fixes #757

# Details

The pytest 5.4.0 update changed how "implicit" parameters are
passed around. See

https://github.com/pytest-dev/pytest/issues/6909

for more details.

### DIFF
--- a/sherpa/models/tests/test_regrid_unit.py
+++ b/sherpa/models/tests/test_regrid_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2017, 2018, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2017, 2018, 2019, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -439,7 +439,7 @@ def test_regrid1d_error_grid_mismatch_1(setup_1d):
 
 def test_ui_regrid1d_non_overlapping_not_allowed():
     """Integrated data space must not overlap"""
-    
+
     ui.dataspace1d(1,100,2,dstype=Data1DInt)
     b1 = Box1D()
     ui.set_model(b1)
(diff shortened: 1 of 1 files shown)


DougBurke:
It looks like this change may be reverted in `pytest`, but I think it's still worth merging this PR (it is a bit less magical, if anything in pytest can be set to not be magical).

One concern I had was whether this would work with pytest prior to 5.4.0 but fortunately the travis tests include this condition, and it does.

wmclaugh:
@DougBurke-

Dan has reviewed and approved your fix. I note that in the meantime, pytest has already released v5.4.1 which reverts the change. Does the new release have any implications on this PR?

DougBurke:
Thanks for the quick review. I just upgraded to `pytest 5.4.1` and the tests still pass, thankfully.

PR #747, opened 2020-02-29
reworked regrid to eval usr grid, but 0 every where else

dtnguyen2:
# Note

This PR implements the requirements as written by @DougBurke, see:
[@DougBurke requirements](https://icxc.harvard.edu/pipe/sherpadev/2020/0063.html) for details.

# caveat
Kept the case where data is integrated but model is not, does that ever get used?

PR #764, opened 2020-03-24
Add tests of 1D regrid for partial-overlap of the user and dataspace grids

DougBurke:
# Summary

Improve the tests of regrid for 1D models.

# Details

Add more tests of regridding 1D models (both integrated and non-integrated), focusing on the case where the user grid (that is, the grid given to the `regrid` method) only partially overlaps the  output grid (the dataspace). This replicates and extends the test case given in issue #722 (and was used to test and validate #747).

### DIFF
--- a/sherpa/models/tests/test_regrid.py
+++ b/sherpa/models/tests/test_regrid.py
@@ -270,7 +270,7 @@ def test_runtime_interp():
         regrid_model = mdl.regrid(requested, interp=interp)
         yregrid = regrid_model(xgrid)
         return yregrid
-    
+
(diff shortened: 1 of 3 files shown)


DougBurke:
Darn github and its ordering of commits. I think because Dan's changes are newer than the tests it has decided it doesn't need to run the travis tests on the last commit (56d734a), which I expect to have failed.

DougBurke:
@dtnguyen2 - I'll add your change in (and have a minor clean up of one of the tests - more explicit setting of a tolerance for the equality check) once the tests have run for the previous commit.

PR #765, opened 2020-03-25
Add docs for setting up all dependencies of the source build with conda

hamogu:
Now that we have a conda-based distribution for all of CIAO, we can
make use of it to install all the build dependencies for a Sherpa
source build. I find that particularly convenient, because it does
not require a fortran compiler, nor any knowledge of how or where the
dependencies such as xspec where build.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -35,7 +35,7 @@ Conda can be used to install all the dependencies for Sherpa.
  
 ::
 
-    conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao ciao ds9 ciao-contrib caldb_main astropy
+    conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao ds9 astropy
     conda install -n sherpaciao --only-deps -c https://cxc.cfa.harvard.edu/conda/ciao sherpa
     conda install -n sherpaciao -c anaconda -c astropy sphinx graphviz sphinx-astropy sphinx_rtd_theme
 
@@ -44,8 +44,9 @@ The first line installes the full `CIAO release
 and running tests locally. The last line adds all requirements for
 building the documentation.  Sherpa can use either astropy or crates
 as backend for reading and writing files. The default configuration in
-Sherpa is to use astropy. However, if crates is selected as backend in
-`sherpa.rc`, then astropy can be omitted from the install.
+Sherpa is to use astropy. However, if crates is installed (e.g. by
+installing the `ciao` package) and selected as backend in `sherpa.rc`,
+then astropy can be omitted from the install.
 
 As described in :ref:`build-from-source`, the file ``setup.cfg`` in
 the root directory of the sherpa source needs to be modified to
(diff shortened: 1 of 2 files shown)


docs/development.rst
```diff
@@ -6,12 +6,98 @@ Contributing to Sherpa development
...
+.. _source-install-with-conda:
+
+Install from source in conda
+============================
+
+Conda can be used to install all the dependencies for Sherpa.
+ 
+::
+
+    conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao ciao ds9 ciao-contrib caldb_main astropy
```
DougBurke: For Sherpa development, we don't need to install `ciao-contrib` or `caldb_main` (dropping the latter would save users a lot of bandwidth/time). Similarly, can they get away with just `sherpa` and `ds9` rather than `ciao` (I can't remember what `sherpa` pulls in, so maybe it isn't enough?). I guess it depends if you want to use crates or astropy backends.

hamogu:
I wanted to have the environment be "almost complete" for Chandra purposes. When I do debugging (not often developing as of now) for Sherpa, I usually try out things with Chandra data, but yes, I'll check if I can do with fewer packages. I agree that the minimum should be listed, I can still say "It might be useful to also have the following packages installed: [list of other CXC packages here]".

DougBurke:
Ta for checking. Why did you put this info on a separate page than the other "how to install" pages? I can see the logic for this, but also see the logic for having everything on one page, and wondered what your thought process was.

hamogu:
I thought the other page was already pretty full with a number of different options. In particular there are three cases for how setup.cfg should be edited. The new case has some overlap with those, but on the other hand, it kind of belongs into a separate subsection (in the conda section). In a separate subsection, we would then have two places with edits to `setup.cfg`. It;s just getting messy and I was afraid to confuse people trying to go through. Putting it on a separate page makes all that clearer. Plus, this setup is only useful for development. As a user, just install sherpa from conda, too; so there is a good justification to put this on a separate page (with the appropriate link_

docs/development.rst
hamogu: On second thought, I don't think I need to explain how to install ciao into that environment, too. These are Sherpa docs. Developers who are also CIAO conda users will know, and others don't need to know. Pushed changes.

DougBurke:
A thought: using `conda install ... sherpa` will create an environment with Sherpa already installed. If you then start building with `python setup.py develop`, is there a chance for confusion (e.g. using the wrong version). I know this is why people are moving to setups like `tox`, but I'm not sure we are able to take advantage of that (due to our setup being designed before recent changes/improvements to setuptools and the like).

hamogu:
That's why the "conda install ... sherpa" is on a separate line with `--only-deps`. That does not install sherpa, but only the dependencies listed for sherpa. In the cxc channel those include e.g. fftw. Thus, you don't have to worry about the external dependencies in compiling sherpa, because those are provided as separate conda package and listed as dependencies for sherpa in the json of the CXC conda channel. ds9 and astropy on the other hand are not listed as dependencies, although they are needed for some tests, thus they need to be installed separately (first line of conda installs).

hamogu:
@DougBurke "ciao" is a dependency in the sense that the setup scripts that set the `ASCDS_LIB` are part of the ciao package. And I don't think that that extra dependency is a problem. There are very, very few sherpa contributors who are not also CIAO users (or work at the CXC) and thus have it installed already anyway - and there are other ways to install sherpa for development for those that really don't like ciao.

PR #770, opened 2020-04-03
Replace EmissionVoight/AbsorptionVoigt models by Voigt model (fix #597)

### LINKED ISSUE #597: Voigt profile functions wrong in Sherpa
From a CXC help desk user:

I am pretty sure I have found two quite serious mistakes in the functions defined in Sherpa as "absorptionvoigt" and emissionvoigt". These are supposed to be (copying-and-pasting from the documentation): 
"emission(/absorption) profile, where a Gaussian model of the core is combined with a Lorentzian model of the wings. The approximation presented in Astrophysical Formulae (K. R. Lang, 1980, 2nd ed., p. 220) is used. " 

So, I went to check the reference, where the approximation is given and compared it to the lines of Ciao code at ~/ciao/lib/python/site-packages/sherpa/astro/optical/__int__.py. 
Here is the relevant (for the emissionvoigt model) part of that file: 

```` python
# This model computes emission as a Voigt function -- i.e., with
# a Gaussian core and Lorentzian wings.
class EmissionVoigt(ArithmeticModel):
"""Emission Voigt function."""

def __init__(self, name='emissionvoigt'):
self.center = Parameter(name, 'center', 5000., tinyval, hard_min=tinyval, frozen=True, units="angstroms")
self.flux = Parameter(name, 'flux', 1.)
self.fwhm = Parameter(name, 'fwhm', 100., tinyval, hard_min=tinyval, units="km/s")
self.lg = Parameter(name, 'lg', 1., tinyval, hard_min=tinyval, frozen=True)

# Create core and wings from Gaussian and Lorentz
self._core = EmissionGaussian()
self._wings = EmissionLorentz()

ArithmeticModel.__init__(self, name, (self.center, self.flux, self.fwhm, self.lg))

#@modelCacher1d
def calc(self, p, x, xhi=None, **kwargs):
# Combining two emission components means
# adding them, it appears (at least according
# to addEmission in NarrowBandFunction.java)

core_pars = numpy.array([ p[2], p[0], p[1], 1.0])
wing_pars = numpy.array([ p[3]*p[2], p[0], p[1], 2.0])
return self._core.calc(core_pars, x) + self._wings.calc(wing_pars, x)
````

So, this function in Sherpa computes a voigt profile as the sum of a Gaussian and a Lorentizan: THIS IS WRONG! 
A Voigt profile is the convolution product (very different from a sum, mathematically) of a Gaussian and a Lorentzian, i.e. the weighted integral of the Lorentzian over the Gaussian profile. And it is the same thing for emission and absorption (the only difference is that in one case the profile is added to the continuum and in the other it is subtracted from it). 

The approximation in Lang (1980) is not the sum of the Gaussian and the Lorentzian, but the sum of the Gaussian and an 1/(x-x0)^2  like function function, weighted with the product of the FWHMs of the two original functions (the Gaussian and the Lorentzian). 

The piece of code where the "absorptionvoigt" function is defined (see below) is even more seriously incorrect, I think: because there the Gaussian and the Lorentzian are multiplied by each other! 
Again, there is no difference between absorption and emission: the profile is defined in exactly the same way: it only depends on whether the user "adds" it to the continuum (emission) or subtract it from it (absorption). 

```` python
# This model computes absorption as a Voigt function -- i.e., with
# a Gaussian core and Lorentzian wings.
class AbsorptionVoigt(ArithmeticModel):
"""Absorption Voigt function expressed in equivalent width."""

def __init__(self, name='absorptionvoigt'):
self.center = Parameter(name, 'center', 5000., tinyval, hard_min=tinyval, frozen=True, units="angstroms")
self.ew = Parameter(name, 'ew', 1., tinyval, hard_min=tinyval, units="angstroms")
self.fwhm = Parameter(name, 'fwhm', 100., tinyval, hard_min=tinyval, units="km/s")
self.lg = Parameter(name, 'lg', 1., tinyval, hard_min=tinyval)

# Create core and wings from Gaussian and Lorentz
self._core = AbsorptionGaussian()
self._wings = AbsorptionLorentz()

ArithmeticModel.__init__(self, name, (self.center, self.ew, self.fwhm, self.lg))

#@modelCacher1d
def calc(self, p, x, xhi=None, **kwargs):
# Combining two absorption components means
# multiplying them, it appears (at least according
# to addAbsorption in NarrowBandFunction.java)

core_pars = numpy.array([ p[2], p[0], p[1] / 2.0, 4.0 ])
wing_pars = numpy.array([ p[2] * p[3], p[0], p[1] / 2.0 ])
return self._core.calc(core_pars, x) * self._wings.calc(wing_pars, x)
````

### PR DESCRIPTION
**Pull request by dtnguyen2**
# Summary

Replaces the EmissionVoigt and AbsorptionVoigt models with a single model, Voigt1D. The EmissionVoigt and AbsorptionVoigt models will error out when an instance is created, pointing users to Voigt1D (as the parameter definitions have changed).

# Note

The voigt profile, where requirement is given in
[https://github.com/sherpa/sherpa/issues/597#issuecomment-464845565](https://github.com/sherpa/sherpa/issues/597#issuecomment-464845565)

### CODE REVIEW

**Comment by DougBurke** [comment]
For reference, I am using some stuff from https://stackoverflow.com/questions/59049433/fitting-voigt-function-to-data-in-python as a test case, in particular comparing to the `lmfit` package, in particular their example at https://lmfit.github.io/lmfit-py/builtin_models.html?highlight=voigt#example-1-fit-peak-data-to-gaussian-lorentzian-and-voigt-profiles (although the parameterisation of their Voigt model is different to the one here)

**Comment by dtnguyen2** [comment]
For completeness, you may want to checkout: [xsvoigt](https://cxc.harvard.edu/sherpa/ahelp/xsvoigt.html)

**Comment by DougBurke** [comment]
Just reported a bug in xsvoigt to the XSPEC folks, as it can cause Sherpa to crash when used with certain parameter values.

**Comment by dtnguyen2** [comment]
Does the sherpa version crashed for the same entries?

**Comment by DougBurke** [comment]
Nope.

**Comment by DougBurke** [comment]
Things look good overall. Things to discuss with @anetasie 

 - the alpha and gamma parameters should probably have min/default_min set to 0 (I believe)
 - do we want to expose sigma rather than alpha as the model parameter, since then it is easier to tie gamma and sigma together  (I say this because lmfit has them tied together by default https://lmfit.github.io/lmfit-py/builtin_models.html?highlight=voigt#lmfit.models.VoigtModel ) - we can do this by saying `vmdl.alpha = np.sqrt(2 * np.log(2)) * vmdl.gamma` (when `vmdl` is an instance of `VoigtModel`) but it's not as easy as `vmdl.gamma = vmdl.sigma` ;-)
 - I'm thinking we should keep the `Absorption/EmissionVoigt` models but make their `__init__` routines raise an error telling people to use Voigt instead (this is a usability issue)

File: sherpa/models/basic.py [hunk 1]
```diff
@@ -40,14 +41,34 @@
...
 DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 
+class Voigt(RegriddableModel1D):
+    """Voigt Profile
+    https://scipython.com/book/chapter-8-scipy/examples/the-voigt-profile/
+    """
+
+    def __init__(self, name='voigt'):
+        self.alpha = Parameter(name, 'alpha', 0.1)
```
hamogu, line 58: If I read this correctly, `alpha` and `gamma` are HWHM for the Gaussian and the Lorentian. Instead, I think it would be much better to (a) use more explicit names and (b) be consistent with other models in Sherpa. Looking at the Gauss and Lorentz models, the parameters should use the FWHM as parameters. I suggest `fwhm_g` and `fwhm_l` as names. That way, I can e.g. fit my data with the Gauss model and when I alter find out I need the more complicated Voigt model, then I can just copy the fwhm from the old Gauss as a starting value to this new model. I think that beats both the implementation here and @DougBurke suggestion to use `sigma` as a parameter instead.

sherpa/models/basic.py
```diff
@@ -40,14 +41,34 @@
...
            'NormGauss2D', 'Polynom2D', 'Scale2D', 'UserModel', 'TableModel',
            'Integrate1D')
 
 DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 
+class Voigt(RegriddableModel1D):
+    """Voigt Profile
+    https://scipython.com/book/chapter-8-scipy/examples/the-voigt-profile/
+    """
```
hamogu, line 55: Should this have a reference for Voigt's 1912 paper where the name comes from? http://publikationen.badw.de/de/003395768

sherpa/models/tests/test_basic.py
```diff
@@ -73,4 +76,28 @@ def test_create_and_evaluate(self):
                 self.assertTrue(out.dtype.type is SherpaFloat)
                 self.assertEqual(out.shape, x.shape)
 
-        self.assertEqual(count, 32)
+        self.assertEqual(count, 33)
+
+
+class test_Voigt(SherpaTestCase):
```
hamogu, line 82: Maybe write new tests using pytest so that we can phase out "SherpaTestCase" eventually?

sherpa/models/basic.py
dtnguyen2: Changed the reference:
```
(base) [dtn@devel12 sherpa]$ git diff sherpa/models/basic.py
diff --git a/sherpa/models/basic.py b/sherpa/models/basic.py
index fb118f8c..3f4daa96 100644
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -51,12 +51,12 @@ DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 class Voigt(RegriddableModel1D):
     """Voigt Profile
-    https://scipython.com/book/chapter-8-scipy/examples/the-voigt-profile/
+    http://publikationen.badw.de/de/003395768
     """
 ```

File: sherpa/models/tests/test_basic.py (hunk 3 above)
Comment (dtnguyen2) [reply sherpa/models/tests/test_basic.py:82]: done:
```
(base) [dtn@devel12 sherpa]$ git diff sherpa/models/tests/test_basic.py
diff --git a/sherpa/models/tests/test_basic.py b/sherpa/models/tests/test_basic.py
index 8134c3fc..049c9364 100644
--- a/sherpa/models/tests/test_basic.py
+++ b/sherpa/models/tests/test_basic.py
@@ -23,7 +23,6 @@ from pytest import approx
 
 import sherpa.models.basic as basic
 from sherpa.utils import SherpaFloat, _utils
-from sherpa.utils.testing import SherpaTestCase
 from sherpa.models.model import ArithmeticModel, RegriddableModel1D, RegriddableModel2D
 
 
@@ -31,7 +30,7 @@ def userfunc(pars, x, *args, **kwargs):
     return x
 
 
-class test_basic(SherpaTestCase):
+class TestBasic:
     excluded_models = (ArithmeticModel, RegriddableModel1D, RegriddableModel2D, basic.Const)
 
     def test_create_and_evaluate(self):
@@ -55,7 +54,7 @@ class test_basic(SherpaTestCase):
                 m.load(x,x)
             if isinstance(m, basic.UserModel):
                 m.calc = userfunc
-            self.assertEqual(type(m).__name__.lower(), m.name)
+            assert type(m).__name__.lower() == m.name
             count += 1
 
             try:
@@ -73,13 +72,18 @@ class test_basic(SherpaTestCase):
                 self.fail("evaluation of model '%s' failed" % cls)
 
             for out in (pt_out, int_out):
-                self.assertTrue(out.dtype.type is SherpaFloat)
-                self.assertEqual(out.shape, x.shape)
+                assert out.dtype.type is SherpaFloat
+                assert out.shape == x.shape
 
-        self.assertEqual(count, 33)
+        assert count == 33
 
 
-class test_Voigt(SherpaTestCase):
+def test_basic():
+    my_test_basic = TestBasic()
+    my_test_basic.test_create_and_evaluate()
+
+
+class TestVoigt:
 
     def test_voigt(self):
         def myVoigt(x, alpha, gamma):
@@ -101,3 +105,7 @@ class test_Voigt(SherpaTestCase):
 
         assert voigt_result == approx(myvoigt_result)
         assert voigt_result == approx(scipy_result)
+
+def test_voigt():
+    my_test_voigt = TestVoigt()
+    my_test_voigt.test_voigt()
```

**Comment by dtnguyen2** [comment]
Changed reference, set the min of ```alpha``` and ```gamma``` to ```0.0```:
```
(base) [dtn@devel12 sherpa]$ git diff sherpa/models/basic.py
diff --git a/sherpa/models/basic.py b/sherpa/models/basic.py
index fb118f8c..3f4daa96 100644
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -51,12 +51,12 @@ DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 class Voigt(RegriddableModel1D):
     """Voigt Profile
-    https://scipython.com/book/chapter-8-scipy/examples/the-voigt-profile/
+    http://publikationen.badw.de/de/003395768
     """
 
     def __init__(self, name='voigt'):
-        self.alpha = Parameter(name, 'alpha', 0.1)
-        self.gamma = Parameter(name, 'gamma', 0.1)
+        self.alpha = Parameter(name, 'alpha', 0.1, min=0.0)
+        self.gamma = Parameter(name, 'gamma', 0.1, min=0.0)
         self.pos = Parameter(name, 'pos', 0.0)
         self.ampl  = Parameter(name, 'ampl', 1.0)
         ArithmeticModel.__init__(self, name,
```
A few questions for SDS: 1) change ```alpha``` and ```gamma``` to ```fwhm_g``` and ```fwhm_l```, (and they are to be doubled, right?). 2) I don't quite understand where does sigma fit given the changes to ```alpha``` and ```gamma``` above though

**Comment by hamogu** [comment]
According to https://scipython.com/book/chapter-8-scipy/examples/the-voigt-profile/ (those are the naming conventions you used, right?)
`gamma` is the half-width at half-maximum (HWHM) for the Lorentian, so we want the parameter to be called `fwhm_l` and it is related to the current gamma like this: `gamma = fwhm_l / 2`.

`alpha` is the HWHM of the Gaussian, (often called `sigma`, but not here). Instead we want the parameter to be called `fwhm_g` and it's related to the current parameter `alpha` like this: `alpha = fwhm_g / 2`. There is a relation between alpha and sigma (`alpha = sigma * sqrt(2 * ln(2))`) but we don't need that because your current implementation uses `alpha` and we want to use `fwhm_g` instead, so we only need the relation between those two parameters.

**Comment by hamogu** [comment]
Oh OK, I see where the sigma is used in the code. After all, sigma, HWHM, and FWHM all describe the same thing and they are related by constant factors. Which one we use is mathematically equivalent, but the convention in spectroscopy is to use the FWHM. In statistics, people typically use sigma. I'm sure there are other fields of science where people typically use the HWHM.

**Comment by dtnguyen2** [comment]
@hamogu, your last comment is a great summary but unfortunately does not give me any instructions as to what SDS wants.  I will push what I have done so far and wait for feedback.

**Comment by hamogu** [comment]
@dtnguyen2 Just to be clear, I'll repost the comment 2 above with some bolt font highlighting:

gamma is the half-width at half-maximum (HWHM) for the Lorentian, so **we want the parameter to be called fwhm_l and it is related to the current gamma like this: gamma = fwhm_l / 2**.

alpha is the HWHM of the Gaussian, (often called sigma, but not here). Instead **we want the parameter to be called fwhm_g and it's related to the current parameter alpha like this: alpha = fwhm_g / 2**. There is a relation between alpha and sigma (alpha = sigma * sqrt(2 * ln(2))) but we don't need that because your current implementation uses alpha and we want to use fwhm_g instead, so we only need the relation between those two parameters.

Does that help?

**Comment by hamogu** [comment]
@dtnguyen2 I see what may be confusing. I try again:
SDS would like the Voigt profile to have the following parameters: fmwh_l, fwhm_g, pos, ampl.

**Comment by hamogu** [comment]
Last comment about naming: The model shall be called `Voigt1D` for consistency because we use Gauss1D, Lorentz1D etc. (I'm sorry I did not catch that earlier. I know it's painful to get comments trickle in one by one. Sorry.)

Also, I want to try this out, so I'm resolving the conflict now (which is really silly) so that I can install it locally for me. Please `git pull` your branch before you do any changes locally and you'll get the change I just did to resolve the conflict.

**Comment by hamogu** [comment]
Great! I wanted to try that out by plotting to visually check the shape, but unfortunately, I've hit some installation problem. (Using the same scripts that worked two weeks ago. Grrr. Anyway, that's beyond the scope of this PR.)
Maybe @DougBurke or you can try the following?
```
from sherpa.models import Voigt1D, Gauss1D
from sherpa.astro.models import Lorentz1D
import numpy as np
x = np.arange(-10, 10, .001)
v = Voigt1D()
g = Gauss1D()
l = Lorentz1D()

%matplotlib

for m in [g, l ,v]:
    plt.plot(x, m(x), label=str(m.__class__).split('.')[-1])
```

(And possibly adjust the norm until we have a reasonable comparison).

**Comment by dtnguyen2** [comment]
Regarding the statement:

> (And possibly adjust the norm until we have a reasonable comparison).

How about generating the data using one of the following functions: ```Gauss1D``` or ```Lorentz1D```. Fit the said data using ```Voigt1D``` as the model so the optimizer will adjust the ```norm```

**Comment by DougBurke** [comment]
Too late for me tonight, but if anyone has time they could take the test data I mentioned in https://github.com/sherpa/sherpa/pull/770#issuecomment-611561541 to see how this works.

**Comment by DougBurke** [comment]
I got this error

```
AttributeError: module 'sherpa.models._modelfcts' has no attribute 'wofz'
```

from trying out https://github.com/sherpa/sherpa/pull/770#issuecomment-686851944

**Comment by DougBurke** [comment]
FIG

I have not tried adjusting the parameters as I wasn't sure what @hamogu was after. I'm going to go and try the data I mentioned above.

**Comment by hamogu** [comment]
@DougBurke : The defaults for the FWHM  and amplitude seem to be very different for the three models. Can you (a) make a plot where they are the same and (b) comment if you think we should make the defaults the same?

What I'm hoping for is to see a figure like this: https://scipython.com/static/media/examples/E8/voigt.png and that just needs some adjusting of FWHM and ampl to get the three functions look more similar. I just wanted to visually check that e.g. the wings of the Voigt are wider than the winds of the Gauss, etc.

**Comment by DougBurke** [comment]
With the data file https://raw.githubusercontent.com/lmfit/lmfit-py/master/examples/test_peak.dat as `test_peak.dat` then

```
import numpy as np

from sherpa import ui

ui.set_stat('leastsq')
ui.get_data_plot_prefs()['yerrorbars'] = False

data = np.loadtxt('test_peak.dat')
ui.load_arrays(1, data[:, 0], data[:, 1])

ui.set_source(ui.voigt1d.vmdl)

ui.plot_fit()
```

returns the starting guess, which is obviously not great.

FIG

Fortunately it fits quiclly, and without a guess call:

```
In [3]: ui.fit()
Dataset               = 1
Statistic             = leastsq
Initial fit statistic = 4250.67
Final fit statistic   = 10.9302 at function evaluation 339
Data points           = 401
Degrees of freedom    = 397
Change in statistic   = 4239.74
   vmdl.fwhm_g    0.527003     +/- 0.0502438
   vmdl.fwhm_l    0.262701     +/- 0.0560108
   vmdl.pos       9.24375      +/- 0.0266324
   vmdl.ampl      34.1915      +/- 1.08177

In [4]: ui.plot_fit()
```

FIG

I'm going to assume this is similar to the "unconstrained-gamma" voigt fit reported by lmfit at the end of the section https://lmfit.github.io/lmfit-py/builtin_models.html?highlight=voigt#example-1-fit-peak-data-to-gaussian-lorentzian-and-voigt-profiles

```
[[Model]]
    Model(voigt)
[[Fit Statistics]]
    # fitting method   = leastsq
    # function evals   = 21
    # data points      = 401
    # variables        = 4
    chi-square         = 10.9301767
    reduced chi-square = 0.02753193
    Akaike info crit   = -1436.57602
    Bayesian info crit = -1420.60017
[[Variables]]
    amplitude:  34.1914737 +/- 0.17946860 (0.52%) (init = 65.43358)
    center:     9.24374847 +/- 0.00441903 (0.05%) (init = 9.25)
    sigma:      0.89518909 +/- 0.01415450 (1.58%) (init = 0.8775)
    gamma:      0.52540199 +/- 0.01857955 (3.54%) (init = 0.7)
    fwhm:       2.72573635 +/- 0.01363980 (0.50%) == '1.0692*gamma+sqrt(0.8664*gamma**2+5.545083*sigma**2)'
    height:     10.0872204 +/- 0.03482129 (0.35%) == '(amplitude/(max(2.220446049250313e-16, sigma*sqrt(2*pi))))*wofz((1j*gamma)/(max(2.220446049250313e-16, sigma*sqrt(2)))).real'
[[Correlations]] (unreported correlations are < 0.250)
    C(sigma, gamma)     = -0.928
    C(amplitude, gamma) =  0.821
    C(amplitude, sigma) = -0.651
```

**Comment by DougBurke** [comment]
@hamogu - without knowing what that plot was trying to generate I'm finding it very hard to come up with a sensible parameterisation to get the plot you want. 

I am more concerned about the lmdif naming scheme for their version and how it maps to ours https://github.com/sherpa/sherpa/pull/770#issuecomment-687206952

**Comment by hamogu** [comment]
@DougBurke OK. If you are OK with it, then I'm fine with it, too. I can always suggest to change the default values in a follow-up PR if I still think it's worthwhile after this is in.

**Review by hamogu — APPROVED** [review]

**Comment by DougBurke** [comment]
@hamogu - I am not okay with the parameter names at the moment. I want to understand the lmdif parameterisation first.

File: sherpa/include/sherpa/models.hh [hunk 4]
```diff
@@ -24,9 +24,24 @@
...
+    const DataType p0 = 2 * p[0];
+    const DataType p1 = 2 * p[1];
+    double sigma = p0 / std::sqrt(2 * log(2));
+    std::complex<double> arg = (x - p[2]) + p1 * std::complex<double>(0,1);
+    arg /= sigma * std::sqrt(2);
+    val = p[3] * Faddeeva::w(arg).real() / (sigma * std::sqrt(2 * PI));
+    return EXIT_SUCCESS;
+    
+  }
+
```
DougBurke, line 44: I found I had to change the interpretation of wofz as the scaling between fwhm_l and gamma was inverted:

```
@@ -32,16 +32,19 @@ namespace sherpa { namespace models {
   inline int wofz_point( const ConstArrayType& p, DataType x,
                          DataType& val )
   {
-    const DataType p0 = 2 * p[0];
-    const DataType p1 = 2 * p[1];
-    double sigma = p0 / std::sqrt(2 * log(2));
-    std::complex<double> arg = (x - p[2]) + p1 * std::complex<double>(0,1);
+    // Follow https://en.wikipedia.org/wiki/Voigt_profile after converting
+    // from FWHM.
+    const DataType fwhm_g = p[0];
+    const DataType fwhm_l = p[1];
+    const DataType gamma = fwhm_l / 2;
+    double sigma = fwhm_g / std::sqrt(8 * log(2));
+    std::complex<double> arg = (x - p[2]) + gamma * std::complex<double>(0,1);
     arg /= sigma * std::sqrt(2);
     val = p[3] * Faddeeva::w(arg).real() / (sigma * std::sqrt(2 * PI));
     return EXIT_SUCCESS;
-
+
   }
-
+
```

With this I now get a fit to the above dataset of 

```
In [4]: ui.fit('v')
Dataset               = v
Statistic             = leastsq
Initial fit statistic = 4329.82
Final fit statistic   = 10.9302 at function evaluation 313
Data points           = 401
Degrees of freedom    = 397
Change in statistic   = 4318.89
   vmdl.fwhm_g    2.10801      +/- 0.200975
   vmdl.fwhm_l    1.0508       +/- 0.224043
   vmdl.pos       9.24375      +/- 0.0266324
   vmdl.ampl      34.1915      +/- 1.08177

```

and this now matches the lmdif version as fwhm_l = 2 * gamma and fwhm_g = sigma / sqrt(8 * log(2)) 

It also now better matches the gauss1d and lorentz1d fits to the same data (the datasets are labelled by `g`, `l`, and `v` for gauss1d, lorentz1d, and voigt1d):

```
In [2]: ui.fit('g')
Dataset               = g
Statistic             = leastsq
Initial fit statistic = 4149.03
Final fit statistic   = 29.9943 at function evaluation 80
Data points           = 401
Degrees of freedom    = 398
Change in statistic   = 4119.04
   gmdl.fwhm      2.90164      +/- 0.0632627
   gmdl.pos       9.24277      +/- 0.0268652
   gmdl.ampl      9.81445      +/- 0.185295

In [3]: ui.fit('l')
Dataset               = l
Statistic             = leastsq
Initial fit statistic = 4213.59
Final fit statistic   = 53.7536 at function evaluation 101
Data points           = 401
Degrees of freedom    = 398
Change in statistic   = 4159.84
   lmdl.fwhm      2.30963      +/- 0.0716252
   lmdl.pos       9.24455      +/- 0.0252414
   lmdl.ampl      38.9724      +/- 0.854116

In [4]: ui.fit('v')
Dataset               = v
Statistic             = leastsq
Initial fit statistic = 4329.82
Final fit statistic   = 10.9302 at function evaluation 313
Data points           = 401
Degrees of freedom    = 397
Change in statistic   = 4318.89
   vmdl.fwhm_g    2.10801      +/- 0.200975
   vmdl.fwhm_l    1.0508       +/- 0.224043
   vmdl.pos       9.24375      +/- 0.0266324
   vmdl.ampl      34.1915      +/- 1.08177

```

So you can see the fwhm values of the voigt1d model now better match the gauss1d and lorentz1d fits.

sherpa/models/basic.py
```diff
@@ -40,14 +40,36 @@
...
-           'SigmaGauss2D',
+           'StepLo1D', 'Tan', 'Voigt1D',
+           'Box2D', 'Const2D', 'Delta2D', 'Gauss2D', 'SigmaGauss2D',
            'NormGauss2D', 'Polynom2D', 'Scale2D', 'UserModel', 'TableModel',
            'Integrate1D')
 
 DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 
+class Voigt1D(RegriddableModel1D):
```
DougBurke, line 51: Given that the Voigt1D model is related to Lorentz1D model, shouldnt we have this model in `sherpa.astro.models` instead (which is where Lorentz1D is defined)?

DougBurke:
Here's the best-fit plots of gauss1d (top left), lorentz1d (top right), and voigt1d (bottom left) - so you can see that we are "in between" the gaussian and lorentzian, as desired (this is using my parameterisation I mention in https://github.com/sherpa/sherpa/pull/770#discussion_r485012714 but that only changes the parameter values, not the best-fit model:

FIG

This is better shown by overplotting the models:

FIG

sherpa/models/basic.py
```diff
@@ -40,14 +40,36 @@
...
+    """Voigt1D Profile
+    http://publikationen.badw.de/de/003395768
+    """
+
+    def __init__(self, name='voigt1d'):
+        # fwhm_g = 2 * alpha
+        # fwhm_l - 2 * gamma        
+        self.fwhm_g = Parameter(name, 'fwhm_g', 0.05, min=0.0)
```
DougBurke, line 59: We should use to match the Gauss1D and Lorentz1D models [although technically we should reduce the default values of fwhm_l/g from 10 to something smaller)

        self.fwhm_g = Parameter(name, 'fwhm_g', 10, tinyval, hard_min=tinyval)
        self.fwhm_l = Parameter(name, 'fwhm_l', 10, 0, hard_min=0)

[I you set fwhm_g to 0 you get nan's out, so we should stop that]

sherpa/astro/optical/__init__.py
```diff
@@ -56,10 +56,10 @@
 
 __all__ = ('AbsorptionEdge', 'AccretionDisk', 'AbsorptionGaussian',
            'AbsorptionLorentz', 'EmissionLorentz', 'OpticalGaussian',
-           'EmissionGaussian', 'AbsorptionVoigt', 'BlackBody',
+           'EmissionGaussian', 'BlackBody',
            'Bremsstrahlung', 'BrokenPowerlaw', 'CCM', 'LogAbsorption',
            'LogEmission', 'Polynomial', 'Powerlaw', 'Recombination',
-           'EmissionVoigt', 'XGal', 'FM', 'LMC', 'SM', 'SMC', 'Seaton')
+           'XGal', 'FM', 'LMC', 'SM', 'SMC', 'Seaton')
```
DougBurke, line 62: We want to leave Absorption/EmissionVoigt available so that users who have used these models get some information on what has changed. I suggest

```
diff --git a/sherpa/astro/optical/__init__.py b/sherpa/astro/optical/__init__.py
index 15112f88..42807644 100644
--- a/sherpa/astro/optical/__init__.py
+++ b/sherpa/astro/optical/__init__.py
@@ -41,6 +41,7 @@ import numpy
 from sherpa.models.parameter import Parameter, tinyval
 from sherpa.models.model import ArithmeticModel, RegriddableModel1D
 from sherpa.utils import SherpaFloat, sao_fcmp
+from sherpa.utils.err import ModelErr

 _tol = numpy.finfo(numpy.float32).eps

@@ -57,6 +58,7 @@ _tol = numpy.finfo(numpy.float32).eps
 __all__ = ('AbsorptionEdge', 'AccretionDisk', 'AbsorptionGaussian',
            'AbsorptionLorentz', 'EmissionLorentz', 'OpticalGaussian',
            'EmissionGaussian', 'BlackBody',
+           'AbsorptionVoigt', 'EmissionVoigt',  # Use Voigt1D instead
            'Bremsstrahlung', 'BrokenPowerlaw', 'CCM', 'LogAbsorption',
            'LogEmission', 'Polynomial', 'Powerlaw', 'Recombination',
            'XGal', 'FM', 'LMC', 'SM', 'SMC', 'Seaton')
@@ -96,6 +98,19 @@ def _extinct_interp(xtable, etable, x):
     return out


+class AbsorptionVoigt(ArithmeticModel):
+    """This model has been replaced by Voigt1D"""
+
+    def __init__(self, *args):
+        raise ModelErr("The AbsorptionVoigt model has been replaced by Voigt1D")
+
+class EmissionVoigt(ArithmeticModel):
+    """This model has been replaced by Voigt1D"""
+
+    def __init__(self, *args):
+        raise ModelErr("The EmissionVoigt model has been replaced by Voigt1D")
+
+
 # This model sets in edge (in Angstroms) beyond which absorption
 # is a significant feature to the spectrum or SED.
 class AbsorptionEdge(RegriddableModel1D):
```

dtnguyen2: Haven't tried it yet but I suspect the test of the file ```sherpa/astro/optical/tests/test_optical.py``` would fail

DougBurke: ========================================== FAILURES ==========================================
____________________________ test_models.test_create_and_evaluate ____________________________

self = <test_optical.test_models testMethod=test_create_and_evaluate>

    def test_create_and_evaluate(self):
        x = arange(4500.0,5500.0,1,SherpaFloat)
        count = 0

        for cls in dir(models):
            clsobj = getattr(models, cls)

            if not isinstance(clsobj, type) \
                    or not issubclass(clsobj, ArithmeticModel) \
                    or clsobj in self.excluded_model_classes:
                continue

>           m = clsobj()

sherpa/astro/optical/tests/test_optical.py:44:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'AbsorptionVoigt' object has no attribute 'name'") raised in repr()]
AbsorptionVoigt object at 0x7fc3a2d06250>
args = ()

    def __init__(self, *args):
>       raise ModelErr("The AbsorptionVoigt model has been replaced by Voigt1D")
E       sherpa.utils.err.ModelErr: The AbsorptionVoigt model has been replaced by Voigt1D

sherpa/astro/optical/__init__.py:105: ModelErr
================================ 1 failed, 25 passed in 0.12s ================================


But we can adjust the test to handle these two models in the short term

DougBurke: It's as simple as

```
% git diff sherpa/astro/optical/tests/test_optical.py
diff --git a/sherpa/astro/optical/tests/test_optical.py b/sherpa/astro/optical/tests/test_optical.py
index 4548de45..ddfb444c 100644
--- a/sherpa/astro/optical/tests/test_optical.py
+++ b/sherpa/astro/optical/tests/test_optical.py
@@ -27,7 +27,9 @@ from sherpa.models.model import ArithmeticModel, RegriddableModel1D, Regriddable


 class test_models(SherpaTestCase):
-    excluded_model_classes = (ArithmeticModel, RegriddableModel1D, RegriddableModel2D)
+    # The Absorption/EmissionVoigt models will be removed at some
+    # point after Sherpa 4.13 is released.
+    excluded_model_classes = (ArithmeticModel, RegriddableModel1D, RegriddableModel2D, models.AbsorptionVoigt, models.EmissionVoigt)

     def test_create_and_evaluate(self):
         x = arange(4500.0,5500.0,1,SherpaFloat)
```

DougBurke:
@dtnguyen2 - I'll look at this tomorrow morning. Do you want me to push a docstring-only change to this PR (just to update the model description) or have you got more changes planned?

dtnguyen2:
Made the necessary changes to reproduce the [fit](https://github.com/sherpa/sherpa/pull/770#discussion_r485012714), added the ```Absoption/EmissionVoigt``` class, set the init parameters for ```fwhm_g/l``` to 10. I think I made all the necessary changes.

DougBurke:
The tests are failing because the voigt tests still have the "old" expected values.

anetasie:
I got lost in the whole discussion. How is the amplitude of the model defined?

dtnguyen2:
The amplitude parameter is called ```ampl```:
 
```
(sherpa) [dtn@devel12 sherpa]$ python
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
>>> import sherpa.models.basic as basic
>>> voigt = basic.Voigt1D()
>>> print(voigt)
voigt1d
   voigt1d.fwhm_g thawed           10  1.17549e-38  3.40282e+38           
   voigt1d.fwhm_l thawed           10            0  3.40282e+38           
   voigt1d.pos  thawed            0 -3.40282e+38  3.40282e+38           
   voigt1d.ampl thawed            1 -3.40282e+38  3.40282e+38           
```

[The  amplitude parameter is p[3] in the code](https://github.com/sherpa/sherpa/pull/770#discussion_r485012714)

anetasie:
@dtnguyen2 Is this the value at the peak or the integral of the function?

DougBurke:
@anetasie - it's the same description as the lmfit model (apart from we have fwhm_g,fwmg_l rather than sigma,gamma) as discussed above - https://github.com/sherpa/sherpa/pull/770#issuecomment-687206952

There's output there that shows the relationship between amplitude and height. The fit below is with the parameters (this is the latest version, after Dan fixed up the fwhm_g/l params):

```
   vmdl.fwhm_g    2.10801      +/- 0.200975
   vmdl.fwhm_l    1.0508       +/- 0.224043
   vmdl.pos       9.24375      +/- 0.0266324
   vmdl.ampl      34.1915      +/- 1.08177
```

You can see that the peak of the data and model is ~ 10 but the amplitude is ~34.

FIG

DougBurke:
It also fits in with the existing lorentz1d model, which for this dataset gives

```
   lmdl.fwhm      2.30963      +/- 0.0716252
   lmdl.pos       9.24455      +/- 0.0252414
   lmdl.ampl      38.9724      +/- 0.854116
```

Unfortunately the gauss1d is defined differently, so there's no way to get a consistent ampl here (the gaussian ampl value is ~ 9, so that must be a height):

```
   gmdl.fwhm      2.90164      +/- 0.0632627
   gmdl.pos       9.24277      +/- 0.0268652
   gmdl.ampl      9.81445      +/- 0.185295
```

If I switch to normgauss1d then we get an ampl value more-close to voigt/lorentz:

```
   gg.fwhm        2.90164      +/- 0.0632846
   gg.pos         9.24277      +/- 0.0268652
   gg.ampl        30.3139      +/- 0.572419
```

anetasie:
@DougBurke ok Thanks. lmfit output has two parameters - amplitude and height of the fitted model, so then the amplitude refers to the integral.  `lorentz1d` is normalized and comparison to `normgauss1d` is good then. So the ampl for the Voigt profile does refer to the integral.

I like the lmfit outputs with both amplitude and height, where height is just calculated using the formula and a correct error propagation. Good to have the formula in the documentation.

sherpa/models/basic.py
```diff
@@ -40,14 +40,36 @@
...
+           'Box2D', 'Const2D', 'Delta2D', 'Gauss2D', 'SigmaGauss2D',
            'NormGauss2D', 'Polynom2D', 'Scale2D', 'UserModel', 'TableModel',
            'Integrate1D')
 
 DBL_EPSILON = numpy.finfo(numpy.float).eps
 
 
+class Voigt1D(RegriddableModel1D):
+    """Voigt1D Profile
+    http://publikationen.badw.de/de/003395768
```
anetasie, line 53: @hamogu This reference is in German. Yes, this is the original publication of Voigt profile, but in the documentation we do need to explain the way we implemented it and the meaning of the parameters. We need a proper 'docstring'. I guess this could be done in a separate PR.

anetasie:
I noticed that the docstrings of EmissionVoigt/AbsorptionVoigt has been removed and I could not find the new one for Voigt1D functions. Is the docstring for this PR or a different one?

DougBurke:
I plan tp add docstrings this morning after I've reviewed Dan's latest changes. And updated my kernel. And had a cup of tea.

anetasie:
@DougBurke Yes, have a cup of tea first!

hamogu:
@anetasie I think German is a perfectly fine language for science. We should all speak it ;-)
But seriously, I agree we need a docstring in addition to a scholarly reference (but I do see some value in having that reference, because it is the original publication after all.

DougBurke:
```
    References
    ----------

    .. [1] http://publikationen.badw.de/de/003395768

    .. [2] https://en.wikipedia.org/wiki/Voigt_profile

    .. [3] https://en.wikipedia.org/wiki/Faddeeva_function
```

DougBurke:
I've just pushed updates to the docstrings. I'll add a note to the rtd PR page once it's built

DougBurke:
The only issue I have left is whether we have sherpa.models.basic.Voigt1D (current) or sherpa.astro.models.Voigt1D (since this is where Lorentz1D is defined).

sherpa/models/basic.py
```diff
@@ -40,14 +40,91 @@
...
+    Attributes
+    ----------
+    fwhm_g
+        The full-width half-maximum (FWHM) of the Gaussian distribution.
+    fwhm_l
+        The full-width half-maximum of the Lorentzian distribution.
+    pos
+        The center of the profile.
+    ampl
+        The amplitude of the profile..
```
anetasie, line 68: I guess the amplitude is defined as an integral of the profile and equivalent of the EW - equivalent width. 
In our old documentation we give a bit more details about the profile = the core is described by a Gaussian and the wings by a Lorentzian.

anetasie:
@DougBurke Thanks for the docs. I guess the illustrative figures could be included in the rst doc files.

anetasie:
One more question. The two models `Emission/AbsorptionVoigt` were defined for angstrom scale. Is this also the case with the generic `Voigt1D` model?

DougBurke:
@anetasie 

for https://github.com/sherpa/sherpa/pull/770#issuecomment-690387360 you'll have to try it out. The code doesn't care about the units, so technically it should work

for https://github.com/sherpa/sherpa/pull/770#issuecomment-690385874 - we have no place to realistically put the figure in the rst docs.

sherpa/models/basic.py
DougBurke: I am not 100% convinced the old description was reliable.

I have no idea whether the amplitude is the same as the EW or not. You'll have to look down to where I give the formula describing the model, and see.

I am reluctant to add significantly more text here, as I have provided references for people who want to find out more, and the more text we have, the more we have to update when we want to change things.

anetasie: `lmfit` output gives the ampl and height. I assume that ampl is the integral of the profile, but I have not written the code, so I don't know how it is calculated.  However, the calculation/definition of `ampl` needs to be explained otherwise it is ambiguous parameter.

anetasie:
It would be good to figure out where we could have the documentation that includes such illustrations. These are really good!
Maybe in a notebook?

sherpa/models/basic.py
DougBurke: There's a notes section below that describes exactly how ampl and the other parameters are used/defined - it ends at https://github.com/sherpa/sherpa/pull/770/files#diff-dc85eaecd2ce7d045722ea9aa44fab66R93

anetasie: Yes, it refers to Faddeeva function...

DougBurke:
You could look at how lmdif structures it's documentation for inspiration.

DougBurke:
I've rebased this PR to fix a silly typo in my docs change (I had a '..' at the end of a sentence) but also to clean up the history of this PR (to avoid a merge) and to get the latest master-branch changes.

sherpa/models/basic.py
DougBurke: which is referenced so you can find out what it is.

anetasie:
@DougBurke You mean `lmfit` documentation.  `lmdif` is an implementation of the levenberg-marquart algorithm and I've been so confused when you refered to `lmdif` in this PR.

DougBurke:
I meant lmfit, which was mentioned in first comment: 

> For reference, I am using some stuff from https://stackoverflow.com/questions/59049433/fitting-voigt-function-to-data-in-python as a test case, in particular comparing to the lmfit package, in particular their example at https://lmfit.github.io/lmfit-py/builtin_models.html?highlight=voigt#example-1-fit-peak-data-to-gaussian-lorentzian-and-voigt-profiles (although the parameterisation of their Voigt model is different to the one here)

DougBurke (changes requested):
As I have already mentioned, I strongly believe that the Voigt1D model should be in the sherpa.astro.models module rather than sherpa.models.basic because Lorentz1D is defined in sherpa.astro.models.

dtnguyen2:
> The only issue I have left is whether we have sherpa.models.basic.Voigt1D (current) or sherpa.astro.models.Voigt1D (since this is where Lorentz1D is defined).

Something must have been lost in the translation, I don't see how  the presence of the word ```whether``` in your last comment translate to 

>...  I strongly believe  that the Voigt1D model should be in the sherpa.astro.models module...

anetasie:
@dtnguyen2 I have trouble building your branch. I can build sherpa from the master branch, but not from your branch. Not sure where the problem is.

```
$git status
HEAD detached at dan/fix_issue_597
nothing to commit, working tree clean
$python setup.py develop
  File "setup.py", line 57, in <module>
  File "/Users/aneta/git/sherpa/helpers/__init__.py", line 25, in <module>
    from .sherpa_config import sherpa_config
  File "/Users/aneta/git/sherpa/helpers/sherpa_config.py", line 25, in <module>
    from numpy.distutils.core import Command
ImportError: cannot import name 'Command' from 'numpy.distutils.core' (/Users/aneta/anaconda/anaconda/envs/git-sherpa/lib/python3.7/site-packages/numpy/distutils/core.py)
```

DougBurke:
@anetasie - have you got the latest version, it has been rebased at least once?

DougBurke:
@dtnguyen2 - to comment on https://github.com/sherpa/sherpa/pull/770#issuecomment-691221202 - I personally believe it should be moved, but no-one else on the SDS side has cared to comment on this.

hamogu:
I believe it should be moved, too, so please move it - and I also believe that @DougBurke 's believe anlone is good enough to be considered the "the SDS opinion on where the user would expect this model" without a need of all SDS members to explicitly say "yes, I agree".

dtnguyen2:
@DougBurke, it doesn't matter to me where ```Voigt1D``` resides, but if you want it move too astro than I will move it.  
BTW, do I have to pull before modifying the code?

anetasie:
Yes, I needed to update the branch and now its building ok.

anetasie:
btw. the model is called `voigt1d`, but I don't see that we have a `voigt2d`. Also is `voigt2d` being used in any astro application? If not, then I would vote to update the model name to `voigt`.

hamogu:
@anetasie I asked for the name "voigt1d" to bring it in line with the Gauss1D and Lorents1D model, which can be thought of as "mathematical components" of this model (even though Voigt is implemented independently). And while we don't have a Voigt2D model now, and it's certainly not something we want to implement into sherpa right now, I seem to recall that I've used a PSF fitting package for optical data that had a voigt2D model.  

As a casual user, it's not easy to see which models are useful for the 1D, 2D or multiD-case. That's clearly beyond this issue, but I would welcome a more clear labelling - in fact I'll open a new issue for that.

dtnguyen2:
It was originally called ```Voigt``` but @hamogu thinks it should be ```Voigt1D``` and now @anetasie thinks it should be called ```Voigt```. 

> is good enough to be considered the "the SDS opinion on where the user would expect this model" without a need of all SDS members to explicitly say "yes, I agree".

So what is fella supposed to do?

DougBurke:
I believe it should be called Voigt1D too FWIW,

DougBurke:
@dtnguyen2 - to answer https://github.com/sherpa/sherpa/pull/770#issuecomment-691232138 - you will need to pull the branch as it has been updated. In fact, as it has been rebased you may want to delete your existing branch before hand.

DougBurke:
@dtnguyen2 - I can move the model over if that would help?

dtnguyen2:
@DougBurke, go ahead make my day

anetasie:
I have a test case for the old `AbsorptionVoigt` that was given by the user. The new implementation of the Voigt has a different meaning for the absorber which is usually assumed a multiplicative function. In the new implementation it needs to be subtracted from the continuum. This implementation also changes the units of the parameters, as fwhm values now refer to the X-axis units, while in the current ciao implementation the units of the fwhm were defined in km/s. So, there is more that has changed which needs to be documented. Note that I'm not the spectroscopist, so maybe @hamogu can chime in.

Here is the data set:
[tentative_ratio_1_err.dat.gz](https://github.com/sherpa/sherpa/files/5211288/tentative_ratio_1_err.dat.gz)


Here is the setting for the old `AbsorptionVoigt` model:

```
load_data("tentative_ratio_1_err.dat", ncols=3)
set_model(polynom1d.poly*absorptionvoigt.v1)
set_par("v1.fwhm",1000, min=1., max=1000)
set_par("v1.center",2952, min=2800, max=3200)
set_par("v1.ew",5., min=0.01,max= 100)
set_par("v1.lg",0.98)
freeze("v1.lg") 

```

The plot in CIAO looks like this:
Fig_AbsorptionVoigt

Here is how I've tried to define the model using `Voigt1d`, btw. the model is not 'easily' fit - default fit is definitely wrong in this case, while it is 'reasonable' in the old parameterization.

```
load_data("tentative_ratio_1_err.dat", ncols=3)
set_model(polynom1d.poly-voigt1d.v1)
set_par(v1.pos, 2952, min=2800, max=3100)
set_par(v1.ampl,1.,min=0.01,max= 100)
set_par(v1.fwhm_g,1, min=0.01,max=1000)
link(v1.fwhm_l,0.98*v1.fwhm_g)
```

and the plot for this setting:

Figure_voigt1d

hamogu:
@anetasie: The old models were tweaked for essentially these (or very similar) fits. The new model is more general, so it won't fit out-of-the-box to this specific dataset. However, the same is true for the other models we have, e.g. for the Gauss1D we also don't have a separate "AbsGauss" that includes the continuum, instead you would, just as you do here, fit a model of (continuum + Gauss1D with negative ampl). So, while this specific case gets harder, I would argue that overall the use of the models becomes easier because you can use different absorption profiles (Gauss, Lorentz, Voigt) in the same way. That also makes it easier to fit more than one absorption line: You just add more "voigt1d" to your model.

The same is true for the unit of FWHM. What if I want to fit a radio spectrum in GHz, or even more fundamental, how does Sherpa even know what the x unit is in km/s? How does Sherpa know, if my spectrum is in Ang, nm, or micron? We don't have units attached to our dataspaces. We the previous implementation would have been silently wrong if the x x-axis had any other than the expected unit. Now, the onus to convert from x to km/s (or m/s or whatever) is on the user - but explicitly so.

The default values are probably an artifact of the same heritage: "AbsVoigt" was an optical model and worked well for typical optical spectra. But its defaults probably would have failed in other situations - what we really need is to work on making the fit method as robust as we can (e.g. #835, #836 ), not to tweak defaults for one specific application. (It might be too late, but if we started form scratch, I would want all default values to be 1, e.g. look at the first figure in https://github.com/sherpa/sherpa/pull/770#issuecomment-687196613 : Why are the functions so different in height and width, although their functional forms are quite similar? Should they not all have the same ampl or if we use "ampl" for the integrated area under curve - which is confusing in itself - not have the same integrated area? But I guess that ship has sailed...

So, yes, more changes, which need to be documented, but for the better.

DougBurke:
@anetasie - I don't know how meaningful these results are, but they also show the line with the following source model (not from a fit as there's too much structure in the data). Note this is a different model to the one you tried (poly - voigt) and is closer to the version using the initial model (poly * absorptionvoigt).

```
(polynom1d.poly * (1 - voigt1d.v1))
   poly.c0      frozen            1 -3.40282e+38  3.40282e+38
   poly.c1      frozen            0 -3.40282e+38  3.40282e+38
   poly.c2      frozen            0 -3.40282e+38  3.40282e+38
   poly.c3      frozen            0 -3.40282e+38  3.40282e+38
   poly.c4      frozen            0 -3.40282e+38  3.40282e+38
   poly.c5      frozen            0 -3.40282e+38  3.40282e+38
   poly.c6      frozen            0 -3.40282e+38  3.40282e+38
   poly.c7      frozen            0 -3.40282e+38  3.40282e+38
   poly.c8      frozen            0 -3.40282e+38  3.40282e+38
   poly.offset  frozen            0 -3.40282e+38  3.40282e+38
   v1.fwhm_g    thawed            5  1.17549e-38  3.40282e+38
   v1.fwhm_l    thawed            5            0  3.40282e+38
   v1.pos       thawed         2952 -3.40282e+38  3.40282e+38
   v1.ampl      thawed            2 -3.40282e+38  3.40282e+38
```

FIG

DougBurke:
Oh *bo**ox* - I think I've messed something up with the rebase. Will have to fix tomorrow

DougBurke:
I've added a PseudoVoigt1D model - based on https://lmfit.github.io/lmfit-py/builtin_models.html?highlight=voigt#lmfit.models.PseudoVoigtModel

DougBurke:
I have also added guess routines for both models. They aren't great, but that's because our fwhm guesser isn't great (e.g. #875)

DougBurke:
@anetasie - I've added some example plots in https://sherpa--770.org.readthedocs.build/en/770/examples/fit_peaked_data.html

DougBurke (approved):
I think this is fine now, but it should be checked by @dtnguyen2 for the code changes I made (moving to sherpa.astro.models) and the new documentation I added by @anetasie to catch typos and thinkos.

anetasie:
The docs look good!

anetasie (approved):

Issue #68, opened 2015-07-20
Add Python classes for XSpec convolution models


DougBurke:
The XSpec models come in additive, multiplicative, convolution, pileup, and mixing-model types. Sherpa already supports additive and multiplicative. It does not support the one pileup model provided by XSpec, but does have its own version of this functionality. This leaves convolution and mixing-model types. This request is about the convolution models only, which are documented at http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/Convolution.html

The `_xspec.cc` file already contains interfaces to the XSpec convolution models. To support there use in Sherpa we need Python classes.

I have an initial set of classes in my branch at https://github.com/DougBurke/sherpa/tree/feature/xspec-convolution-models - which is based on the support I wrote for these models in the CIAO contributed-script package. It is not ready for a PR, as I want to experiment a bit more.

Note that to properly use these models from Sherpa will require a rethink of how model evaluation is done (I will write up thoughts on this in another issue at some point in the near future). That is, some XSpec convolution models can be used as is, but some aren't going to be useful until further changes are made in Sherpa itself.

The gist at https://gist.github.com/DougBurke/65dd0716afc061a03d63 contains the output of a script (and the script itself) that shows the `cflux` and `cpflux` models being used - see http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSmodelCflux.html and http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSmodelCpflux.html using https://github.com/DougBurke/sherpa/commit/ccd2ccf97e4a95784c8142bca624c5580dab9fcf

DougBurke:
I've updated the gist at  https://gist.github.com/DougBurke/65dd0716afc061a03d63 to add `example2.py`, which is a version of the script using the API added in changeset https://github.com/DougBurke/sherpa/commit/810cae93ba8eda6ae41ca0efdeab1fd796dbcab9

DougBurke:
I wonder if there's overlap with #43 - at least in the sense that the output grid may not be the same as the grid we want to evaluate the data on.

PR #772, opened 2020-04-04
Add XSPEC 12.11.0 support (HEAsoft 6.27)

DougBurke:
# Summary

Add support for XSPEC 12.11.0 (released March 31 2020).

### Additive

- `XSagnslim`
- `XSzkerrbb`
- `XSbwcycl`

### Multiplicative

- `XSismdust`
- `XSolivineabs`
- `XSlogconst`
- `XSlog10con`

## Model parameter values

Updates the following parameter settings for changes made to the
model.dat file in XSPEC 12.11.0. There is no way to make these
settings depend on the XSPEC model library in use, so we chose to
use the latest values (in many cases these will be compatible
but there may be cases where the values are not valid with older
XSPEC libraries):

 - `XSagnsed`: `Htmax` parameter has changed to lie within the range
   6 to 10, and defaults to 10 (previously you could use 6 to 200
   and the default was 100).

 - `XSkerrbb`: `rflag` and `lflag` are now marked as `alwaysfrozen`
   (they have been changed to '$' parameters in model.dat)

 - `XSxscat`: `Xpos` and `Rext` soft and hard limits have increased;
   they were 0.95/0.95 and 115/119 and are now 0.99/0.999 and
   235/240 respectively

---

# Details

 - [X] support compiling against 12.11.0 (only tested on Linux)
 - [X] check this fix still works with 12.10.1 (ie get a green from Travis CI tests)
 - [X] fix failing test with 12.11.0: `sherpa/astro/ui/tests/test_astro_ui.py::sherpa/astro/ui/tests/test_astro_ui.py`  (it's a regression test that uses APEC and the APEC code has been changed in XSPEC 12.11.0) - see 7c4d4a0 (this is marked as failed because the macOS build timed out and for some reason I can't re-run just that test; you can see it passed in the next commit 62adf64)
 - [X] add support for new models in XSPEC 12.11.0
    - [X] agnslim
    - [X] zkerrbb
    - [X] ismdust
    - [X] olivineabs
    - [X] thcomp (only direct access, no Python class as waiting for #68 to be resolved)
    - [X] logconst
    - [X] log10con
    - [X] bwcycl
    (mixing models not supported)
 - [X] update documentation to include 12.11.0 in supported XSPEC versions
 - [X] update `setup.cfg` comments to include `hdsp` library component (as this is needed in recent versions of XSPEC)
 - [X] add a developer guide for how to handle an XSPEC version increase

### DIFF
--- a/docs/indices.rst
+++ b/docs/indices.rst
@@ -22,7 +22,7 @@ Glossary
    Astropy
        A community Python library for Astronomy:
        https://www.astropy.org/.
-       
+
    CIAO
        The data reduction and analysis provided by the :term:`CXC`
        for users of the Chandra X-ray telescope. Sherpa is provided
@@ -35,7 +35,7 @@ Glossary
        speciality support for X-ray data formats that follow
        :term:`OGIP` standards (such as :term:`ARF` and :term:`RMF`
        files).
-       
+
    CXC
        The `Chandra X-ray Center <http://cxc.harvard.edu/>`_.
 
@@ -45,9 +45,9 @@ Glossary
        image data, and is available from http://ds9.si.edu/. It uses
        the :term:`XPA` messaging system to communicate with external
        processes.
-       
+
    FITS
-       The `Flexible Image Transport System 
+       The `Flexible Image Transport System
        <https://en.wikipedia.org/wiki/FITS>`_ is a common data format in
        Astronomy, originally defined to represent imaging data from radio
        telescopes, but has since been extended to contain a mixture of
@@ -85,7 +85,7 @@ Glossary
        `OGIP/92-007
        <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/ogip_92_007.html>`_
        and
-       `OGIP/92-007a 
+       `OGIP/92-007a
        <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007a/ogip_92_007a.html>`_.
        Confusingly, PHA can also refer to the Pulse Height Amplitude (the
        amount of charge detected) of an event, which is one of the
@@ -98,7 +98,7 @@ Glossary
        dependent on the optical design of the system but can also be
        influenced by other factors (e.g. for ground-based observatories
        the atmosphere can add additional blurring).
-       
+
    RMF
        The Redistribution Matrix Function used to describe the response
        of an Astronomical X-ray detector. It is a matrix containing the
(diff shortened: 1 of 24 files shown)


DougBurke:
About to rebase because of minor conflict with #732 (change to indentation of copyright line, so no functional changes).

DougBurke:
@dtnguyen2 - when I worked on the C++ side for XSPEC in this, and previous PRs, I intentionally tried to keep it simple (i.e. allow repeated bits of code) rather than try and fix everything with a layer of C++ templates, as C++ templates are not my forte. I also based this on existing code in Sherpa, which is why it has a lot of old-style casting.

The XSPEC model templates also have a bunch of commented-out code, from checks that were going to be done but turned out to not-be-a-good-idea-at-the-time™

The question is whether this is a good place to fix these issues, or for them to be done in a separate PR?

You can also link to the line number of a particular commit (although it mike require a bit of work to get to the file if it isn't actually changed in that commit). For example, the latest commit in this PR has

https://github.com/sherpa/sherpa/blob/f7d4903860a7e59d77c7a2a86fbffc5ce00d0aa9/sherpa/include/sherpa/astro/xspec_extension.hh#L174

which is I assume the line 174-289 you mentioned and line 418-... is the 416-530

https://github.com/sherpa/sherpa/blob/f7d4903860a7e59d77c7a2a86fbffc5ce00d0aa9/sherpa/include/sherpa/astro/xspec_extension.hh#L418

DougBurke:
@dtnguyen2 - for the std::transform I came up with something slightly different to yours

```
std::transform(std::begin(ear), std::end(ear), std::begin(fear),
		       [](const double val) -> FloatArrayType { return static_cast<FloatArrayType>(val); });
```

It seems to work.

DougBurke:
@dtnguyen2 - to simply this PR I've started #859 to reduce the duplication in `xspec_extension.hh`

DougBurke:
Rebased to address conflicts in the documentation.

DougBurke:
@dtnguyen2 - I'd like to separate out the C++ changes from this PR (whether #859 or some other commit). Is that okay with you?

dtnguyen2:
sounds good to me

docs/indices.rst
```diff
@@ -35,7 +35,7 @@ Glossary
        speciality support for X-ray data formats that follow
        :term:`OGIP` standards (such as :term:`ARF` and :term:`RMF`
        files).
-       
+
    CXC
        The `Chandra X-ray Center <http://cxc.harvard.edu/>`_.
```
wmclaugh, line 40: Should this be 'https' now?

DougBurke: It should (although the http form will be okay for a while now). There's probably a bunch of other URLs in the docs that need updating.

DougBurke: Updated the http cases (left fftw as is that page still seems to be http only).

PR #782, opened 2020-04-15
Add a pyproject.toml file

hamogu:
PEP 518 specifies the pyproject.toml file and the format is
understood by recent pip versions. On the other hand, there is no
good way to do that in setup.py or setup.cfg.

At the same time, numpy is installed as a wheel for almost all
systems nowadays so getting an extra numpy just for the sherpa
installation is not a big cost any longer.

At a later point, it might makes sense to also move some of the
information in setup.py into setup.cfg but this is a separate
issue.

closes #401
closes #382

Issue #595, opened 2019-02-08
move sherpa/ into src/ ?


DougBurke:
Should we move the `sherpa/` directory into `src/sherpa/` since it *might* help avoid issues - e.g. trying to use sherpa from the installation directory can lead to errors about a missing `_utils` model. Or would this not really help there?

I don't know how related this is to #50 #133 #198

olaurino:
I don't know how that would work with `python setup.py develop`, but it might be worth investigating.

olaurino:
And I wouldn't think it's related to those issues. Our files are arranged in a way that is pretty standard for python packages (although not the only possible model). A developer should know the difference between `develop` and `install` and the order of priority of python modules, i.e. if you are in a folder with a `model.py` and you `import model` that folder takes the precedence over everything else. That's just the way Python works.

DougBurke:
Note that I have seen 'pip install -e .' mentioned as the new "hotness" for
development, but have little experience with it myself.

olaurino:
I think it's more or less the same, with some differences in the way dependencies are treated, see e.g.
https://stackoverflow.com/questions/30306099/pip-install-editable-vs-python-setup-py-develop

hamogu:
Can we close this? I don't see a benefit from moving the name. `pip install -e` works (at least for me) in the current configuration and, as @olaurino pointed out above, the current layout is pretty standard.

DougBurke:
I have, over the years, read documents that have strongly suggested that the way we currently do things is not a great idea - but it is mainly an issue for the developer as @olaurino had mentioned. Whenever I've tried to try and improve our build/CI work I seem to always end up in bizarre corner cases which *might* be related to this, but might not be. Whenever I try to find clarity on-line I fail.

The trouble, as far as I can tell, is that the recent PEP work decoupling the python build process from setuptools has focused on "pure python" packages, and those that require compiling code still have "issues". I have not been keeping on top of it, so I could well have missed something recently, but I don't find the existing documentation particularly helpful.

I can close this, but I would like us to be able to make some movement on #784 - as may be evidenced from this comment I'm not sure I'm the best person to lead that work ;-)

Issue #480, opened 2018-07-30
ELF file OS ABI invalid


pepgalax:
I have both anaconda2 and anaconda3 with sherpa installed and I am getting the error:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pep/anaconda3/lib/python3.6/site-packages/sherpa/astro/ui/__init__.py", line 20, in <module>
    import sherpa.all
  File "/home/pep/anaconda3/lib/python3.6/site-packages/sherpa/all.py", line 37, in <module>
    from sherpa.fit import *
  File "/home/pep/anaconda3/lib/python3.6/site-packages/sherpa/fit.py", line 36, in <module>
    from sherpa.models import SimulFitModel
  File "/home/pep/anaconda3/lib/python3.6/site-packages/sherpa/models/__init__.py", line 22, in <module>
    from sherpa.models.basic import *
  File "/home/pep/anaconda3/lib/python3.6/site-packages/sherpa/models/basic.py", line 28, in <module>
    from . import _modelfcts
ImportError: /home/pep/anaconda3/lib/python3.6/site-packages/sherpa/models/_modelfcts.cpython-36m-x86_64-linux-gnu.so: ELF file OS ABI invalid

when I do:
import sherpa.astro.ui as ui

Any idea on how to solve this?

pepgalax:
With anaconda2, the error is longer but ends to the same issue:

Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import sherpa.ui
ERROR:root:code for hash md5 was not found.
Traceback (most recent call last):
  File "/home/pep/anaconda2/lib/python2.7/hashlib.py", line 147, in <module>
    globals()[__func_name] = __get_hash(__func_name)
  File "/home/pep/anaconda2/lib/python2.7/hashlib.py", line 97, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type md5
ERROR:root:code for hash sha1 was not found.
Traceback (most recent call last):
  File "/home/pep/anaconda2/lib/python2.7/hashlib.py", line 147, in <module>
    globals()[__func_name] = __get_hash(__func_name)
  File "/home/pep/anaconda2/lib/python2.7/hashlib.py", line 97, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type sha1
[...]

PR #784, opened 2020-04-16
WIP: Redo sherpa installation and CI with PEP 518 and tox

hamogu:
In this PR, I'm changing sherpa's setup scripts and CI setup to follow PEP 518 and astropy's examples.
**This is not functional right now.** I'm posting it here to get some feedback from the Sherpa developers to see if this is a route they are interested in. Getting it all right, especially with the binary components requires some work and probably help from core developers.

This largely follows what astropy does.

- [x] replace versioneer with setuptools_scm (not strictly required, but since this relies on setuptools now seems like a small step to take)
- [x] Install python files
- [ ] compile and install extensions
- [x] run tests for python with tox 
- [ ] run tests using compiled extensions with tox
- [ ] run CI with tox
- [ ] write / update contributor docs (currently referring to astropy docs)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,10 +1,68 @@
-[aliases]
-build = sherpa_config xspec_config build
-build_ext = sherpa_config xspec_config build_ext
-develop = sherpa_config xspec_config develop
-install = sherpa_config xspec_config install
-test = develop test
-bdist_wheel = sherpa_config bdist_wheel
+[metadata]
+name = sherpa
+author = Smithsonian Astrophysical Observatory / Chandra X-Ray Center
+author_email = cxchelp@head.cfa.harvard.edu
+license = GNU GPL v3
+license_file = COPYRIGHT
+url = http://cxc.harvard.edu/sherpa/
+description = Modeling and fitting package for scientific data analysis
+long_description = file: README.md
+keywords = astronomy, astrophysics, modeling, models, fitting, X-ray, data analysis
+classifiers =
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+
+    Operating System :: OS Independent
+    Programming Language :: C
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: Implementation :: CPython
+    Topic :: Scientific/Engineering :: Astronomy
+    Topic :: Scientific/Engineering :: Physics
+
+[options]
+# We set packages to find: to automatically find all sub-packages
+packages = find:
+requires = numpy
+zip_safe = False
+tests_require =
+    pytest>=3.3
+    pytest-xvfb
+setup_requires = setuptools_scm
+install_requires = numpy>=1.10
+python_requires = >=3.5
+
+[options.entry_points]
+console_scripts =
+    sherpa_test = sherpa:clitest
+    sherpa_smoke = sherpa:_smoke_cli
+
+[options.extras_require]
+test =
+    coverage
+docs =
+    sphinx-astropy
+    pytest
+
+[options.data_files]
+sherpa = sherpa/sherpa.rc, sherpa/sherpa-standalone.rc
+
+
+[options.package_data]
+* = data/*, data/*/*, data/*/*/*, data/*/*/*/*, data/*/*/*/*/*, data/*/*/*/*/*/*
+astropy = s
+astropy.utils.tests = data/.hidden_file.txt
+astropy.wcs = include/*/*.h
+astropy.wcs.tests = extension/*.c
+sherpa = include/sherpa/*.hh, include/sherpa/astro/*.hh, tests/*
+sherpa.astro.ui = tests/data/*
+
+[tool:pytest]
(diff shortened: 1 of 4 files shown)


DougBurke:
I'm going to close #736 as the idea there was to work out how to update the build infrastructure, but I don't know how.

DougBurke:
I personally am interested in this work, but progress will be slow here as our build team is swamped with other stuff.

One thing I don't have a good handle on is what the current best practices are for python development and testing nowadays (e.g. tox is mentioned as a replacement for 'python setup.py test`, but what other options are there?). There's also a number of build-related issues that are presumably related to the fact that our build system was put together a long time ago; for instance #50 #198 #480 #595 (a rather incomplete list)

hamogu:
@DougBurke Ifyou and I are interested, that might be enough to get this developed. I just have no experience with the compiled modules in sherpa. I reviewed the issues you linked and it seems that they could all be solved by this. At the same time, I noticed that the CI matrix is a little "odd", e.g. we are testing "python setup.py install" and "python setup.py develop" separately, but on the other hand, we have very little coverage in terms of python / numpy / astropy versions.

All I need to continue at this point is a "Yes, there is a realistic chance this will get merged eventually". I just don't want to spend days on something that has no chance of being accepted.


My draft is taken from applying what astropy does - I think there is no shame to take from a project with a much larger pool of developers that explicitly puts time, effort (and money) into developing build tools and documentation for how to build with the expressed goal to make live easier for other scientific python packages.

A few links for my own notes:

- https://packaging-guide.openastronomy.org/en/latest/docs.html
- https://github.com/pypa/manylinux
- https://github.com/joerick/cibuildwheel/
- https://openastronomy-azure-pipelines.readthedocs.io/en/latest/ (which uses the two packages above)

hamogu:
A few notes for me to look at:

- https://packaging-guide.openastronomy.org/en/latest/
- An example using cibuildwheel behind the scenes: https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html 
- https://github.com/pypa/manylinux/
- https://github.com/joerick/cibuildwheel/

That does not mean that I intend to use those tools, but that I should look and see if we should do that or get some ideas to maybe not at all.

setup.py
```diff
@@ -18,131 +18,67 @@
...
-        "Other dependencies will be automatically installed\n"
-        "Please install NUMPY (e.g. pip install numpy) and try again."
-    ), file=sys.stderr)
-    sys.exit(2)
+if 'test' in sys.argv:
+    print(TEST_HELP)
+    sys.exit(1)
 
-from helpers.extensions import static_ext_modules
+DOCS_HELP = """
```
DougBurke, line 52: A minor issue, but the existing documentation only talks about using `build_sphinx` and not `build_docs` - i.e. https://sherpa.readthedocs.io/en/latest/install.html#building-the-documentation

Do we want this message to note the `build_sphinx` option as well?

setup.cfg
```diff
@@ -1,10 +1,68 @@
...
-install = sherpa_config xspec_config install
-test = develop test
-bdist_wheel = sherpa_config bdist_wheel
+[metadata]
+name = sherpa
+author = Smithsonian Astrophysical Observatory / Chandra X-Ray Center
+author_email = cxchelp@head.cfa.harvard.edu
+license = GNU GPL v3
+license_file = COPYRIGHT
+url = http://cxc.harvard.edu/sherpa/
```
DougBurke, line 7: We should probably change this to the rtd page? If nothing else we should change to the https version ;-)

setup.cfg
```diff
@@ -103,3 +161,42 @@ source-dir = docs
 #wcslib_libraries = wcs
 #gfortran_lib_dirs = None
 #gfortran_libraries = gfortran
+
+
+[coverage:run]
+omit =
```
DougBurke, line 167: I'm guessing we need to replace `astropy` by `sherpa`, and probably tweak the directories?

hamogu: Yes. I don't know if we've set up any coverage testing for Sherpa at all, so this will need some thought.

DougBurke:
Given that #782 has been merged can this be rebased now?

hamogu:
Sure. But there are still a number of steps missing, this is very much a WIP. It will probably have to be rebased more than once before it's ready, so I don't think there is a hurry.

DougBurke:
We have code that has a mixture of absolute and relative imports - e.g. 

```
from sherpa.bob import bobby
from .fred import freddy
```

I'm wondering if that could be the cause of the mysterious `cannot import from sherpa._utils` message we can see? Clutching at straws, but it was a problem in some of my recent experiments to support the aim if this PR.

DougBurke:
Just to note that in https://github.com/cxcsds/xspec-models-cxc I was able to combine new-style declarative setup.cfg with old-style "build an extension". That's not to say we won't have problems [e.g. the location of the grpup/stk libraries #50 ] but there is hope. And something we should definitely work on sooner rather than later after 4.14 has stabilised

PR #793, opened 2020-04-19
Accept masked arrays for Data XXX creation

### LINKED ISSUE #346: unexpected behavior with masked arrays

### LINKED ISSUE #695: Initializing generid data silently drops mask

hamogu:
Accept masked arrays for Data XXX creation
    
This change implements changes to use the mask of numpy arrays
when initializing DataXXX. It warns sensibly for non-numpy objects with
a mask in the OO interface when duck-typing, but not in the session
interface, because that whitelists only specific data types anyway.
Thus, this reverses the warnings added to the docs in #698.
    
fixes #695
fixes #346

### DIFF
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -52,7 +52,7 @@ def _check(array):
     return array
 
 
-def _check_indep(array):
+def _check_nomask(array):
     if hasattr(array, 'mask'):
         warnings.warn('Dropping mask attribute for array {}. Masks are supported for dependent variables only.'.format(array))
     return array
@@ -86,7 +86,7 @@ class DataSpace1D(EvaluationSpace1D):
             the x axis of this data space
         """
         self.filter = filter
-        EvaluationSpace1D.__init__(self, _check_indep(x))
+        EvaluationSpace1D.__init__(self, _check_nomask(x))
 
     def get(self, filter=False):
         """
@@ -152,7 +152,7 @@ class IntegratedDataSpace1D(EvaluationSpace1D):
             the higher bounds array of this data space
         """
         self.filter = filter
-        EvaluationSpace1D.__init__(self, _check_indep(xlo), _check_indep(xhi))
+        EvaluationSpace1D.__init__(self, _check_nomask(xlo), _check_nomask(xhi))
 
     def get(self, filter=False):
         """
@@ -219,8 +219,8 @@ class DataSpace2D():
             the second axis of this data space
         """
         self.filter = filter
-        self.x0 = _check(_check_indep(x0))
-        self.x1 = _check(_check_indep(x1))
+        self.x0 = _check(_check_nomask(x0))
+        self.x1 = _check(_check_nomask(x1))
 
     def get(self, filter=False):
         """
@@ -281,10 +281,10 @@ class IntegratedDataSpace2D():
             the higher bounds array of the xhi axis
         """
         self.filter = filter
-        self.x0lo = _check(_check_indep(x0lo))
-        self.x1lo = _check(_check_indep(x1lo))
-        self.x0hi = _check(_check_indep(x0hi))
-        self.x1hi = _check(_check_indep(x1hi))
+        self.x0lo = _check(_check_nomask(x0lo))
+        self.x1lo = _check(_check_nomask(x1lo))
+        self.x0hi = _check(_check_nomask(x0hi))
+        self.x1hi = _check(_check_nomask(x1hi))
 
     def get(self, filter=False):
         """
@@ -339,7 +339,7 @@ class DataSpaceND():
             the tuple of independent axes.
         """
         self.filter = filter
-        self.indep = _check_indep(indep)
+        self.indep = _check_nomask(indep)
 
     def get(self, filter=False):
         """
@@ -504,8 +504,8 @@ class Data(NoNewAttributesAfterInit, BaseData):
         self.name = name
         self._data_space = self._init_data_space(Filter(), *indep)
         self.y, self.mask = _check_dep(y)
-        self.staterror = staterror
-        self.syserror = syserror
+        self.staterror = _check_nomask(staterror)
+        self.syserror = _check_nomask(syserror)
         NoNewAttributesAfterInit.__init__(self)
 
     def _init_data_space(self, filter, *data):
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -54,7 +54,7 @@ def _check(array):
 
 def _check_nomask(array):
     if hasattr(array, 'mask'):
-        warnings.warn('Dropping mask attribute for array {}. Masks are supported for dependent variables only.'.format(array))
+        warnings.warn('Input array {} has a mask attribute. Because masks are supported for dependent variables only the mask attribute of the independent array is ingored and values `behind the mask` are used.'.format(array))
     return array
 
 def _check_dep(array):
@@ -66,7 +66,7 @@ def _check_dep(array):
             return _check(array), ~array.mask
         # We don't know what the mask convention is
         else:
-            warnings.warn('Dropping mask for array {}. Set .mask attribute manually or use "set_filter" function.'.format(array))
+            warnings.warn('Format of mask for array {} not supported thus the mask is is ignored and values `behind the mask` are used. Set .mask attribute manually or use "set_filter" function.'.format(array))
             return _check(array), True
 
 
@@ -399,11 +399,13 @@ class Filter():
     def mask(self, val):
         if (val is True) or (val is False):
             self._mask = val
-        # if val is of type np.bool_ and True, it was failed the previous test because
+        # if val is of type np.bool_ and True, it failed the previous test because
         # "is True" compares with Python "True" singelton.
         # Yet, we do not want to allow arbitrary values that evaluate as True.
-        elif (val is numpy.ma.nomask) or (numpy.isscalar(val) and val and isinstance(val, numpy.bool_)):
+        elif val is numpy.ma.nomask:
             self._mask = True
+        elif numpy.isscalar(val) and isinstance(val, numpy.bool_):
+            self._mask = bool(val)
         elif (val is None) or numpy.isscalar(val):
             raise DataErr('ismask')
         else:
@@ -475,13 +477,20 @@ class Data(NoNewAttributesAfterInit, BaseData):
     Data class for generic, N-Dimensional data sets, where N depends on the number of independent axes passed during
     initialization.
 
-    A data class is the collection of a data space and a number of data array for the dependent variable and
+    A data class is the collection of a data space and a number of data arrays for the dependent variable and
     associated errors.
 
     This class can be extended by classes definining data sets of specific dimensionality. Extending classes should
     override the `_init_data_space` method.
 
-    This classe provides most of the infrastructure for extending classes for free.
+    This class provides most of the infrastructure for extending classes for free.
+
+    Data classes contain a ``mask`` attribute, which can be used ignore certain values in the array
+    when fitting or plotting that data. The convention in Sherpa is that ``True`` marks a values as
+    *valid* and ``False`` as *invalid* (note that this is opposite to the numpy convention). When a `Data`
+    instance is initialized with a dependent array that has a ``mask`` attribute (e.g. numpy masked array),
+    it will attempt to convert that mask to the Sherpa convention and raise a warning otherwise. In any case,
+    the user can set ``data.mask`` after initialization if that conversion does not yield the expected result.
     """
     _fields = ("name", "indep", "dep", "staterror", "syserror")
 
@@ -492,10 +501,10 @@ class Data(NoNewAttributesAfterInit, BaseData):
         name : basestring
             name of this dataset
         indep: tuple of array_like
-            the tuple of independent arrays. If this is an numpy masked array,
-            the mask will used.
+            the tuple of independent arrays.
         y : array_like
-            the values of the dependent observable
+            The values of the dependent observable. If this is a numpy masked array,
+            the mask will used to initialize a mask.
         staterror : array_like
             the statistical error associated with the data
         syserror : array_like
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -52,6 +52,24 @@ def _check(array):
     return array
 
 
+def _check_indep(array):
+    if hasattr(array, 'mask'):
+        warnings.warn('Dropping mask attribute for array {}. Masks are supported for dependent variables only.'.format(array))
+    return array
+
+def _check_dep(array):
+    if not hasattr(array, 'mask'):
+        return _check(array), True
+    else:
+        # We know the mask convention is opposite to sherpa
+        if isinstance(array, numpy.ma.MaskedArray):
+            return _check(array), ~array.mask
+        # We don't know what the mask convention is
+        else:
+            warnings.warn('Dropping mask for array {}. Set .mask attribute manually or use "set_filter" function.'.format(array))
+            return _check(array), True
+
+
 class DataSpace1D(EvaluationSpace1D):
     """
     Class for representing 1-D Data Space. Data Spaces are spaces that describe the data domain. As models can be
@@ -68,7 +86,7 @@ class DataSpace1D(EvaluationSpace1D):
             the x axis of this data space
         """
         self.filter = filter
-        EvaluationSpace1D.__init__(self, x)
+        EvaluationSpace1D.__init__(self, _check_indep(x))
 
     def get(self, filter=False):
         """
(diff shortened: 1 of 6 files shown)


hamogu:
Note that he test failures stem from"
```
PackagesNotFoundError: The following packages are not available from current channels:

  - libgfortran=3.0
```
That's clearly not due to this PR but an unrelated CI issue. I will try to restart the jobs tomorrow and try again.

hamogu:
#794 might fix the CI, so that this can be restarted.

sherpa/data.py
```diff
@@ -52,6 +52,24 @@ def _check(array):
     return array
 
 
+def _check_nomask(array):
```
DougBurke, line 55: When I see a message like "Dropping mask attribute" I think I'd expect for the attribute to be removed in some way. I'm not sure that's the behavior here - the .mask attribute may stay but it is just ignored. Or am I missing something?

I'm not 100% convinced that `del array.mask` would be a sensible thing here, but I wanted to understand the warning message.

sherpa/data.py
```diff
@@ -381,6 +399,11 @@ def mask(self):
     def mask(self, val):
         if (val is True) or (val is False):
             self._mask = val
+        # if val is of type np.bool_ and True, it was failed the previous test because
+        # "is True" compares with Python "True" singelton.
+        # Yet, we do not want to allow arbitrary values that evaluate as True.
+        elif (val is numpy.ma.nomask) or (numpy.isscalar(val) and val and isinstance(val, numpy.bool_)):
+            self._mask = True
```
DougBurke, line 406: I would have expected something like

```
elif val is numpy.ma.mask:
    salf._mask = True
elif numpy.isscalar(val) and isinstance(val, numy.bool_):
    self._mask = val  # do we need to convert from numpy bool to bool?
elif ...
```

because don't we want `_mask` to be False if `val` is `numpy.bool_(0)`? Or does that get handled by another check in the chain?

sherpa/data.py
```diff
@@ -464,17 +487,13 @@ class Data(NoNewAttributesAfterInit, BaseData):
...
-        to-be-ignored value.
-
         Parameters
         ----------
         name : basestring
             name of this dataset
         indep: tuple of array_like
-            the tuple of independent arrays
+            the tuple of independent arrays. If this is an numpy masked array,
+            the mask will used.
```
DougBurke, line 496: what does 'the mask will be used' mean? Is it "the masked array will be used" or "the unmasked values" or ...? I don't use the masked array stuff so I don't know the expected terminology.

Actually, this is the independent array: isn't the mask ignored in this case? Actually, it depends on _init_data_space does, which isn't immediately obvious.

sherpa/tests/test_data.py
```diff
@@ -977,3 +1014,128 @@ def test_data2d_int_eval_model_to_fit(array_sizes_fixture):
...
+    args[posy] = numpy.ma.array(args[posy])
+    session.load_arrays(*args)
+    new_data = session.get_data(data.name)
+    # Sherpa's way of saying "mask is not set"
+    assert new_data.mask is True
+    assert len(new_data.get_dep(filter=True)) == len(args[posy].flatten())
+
+# https://github.com/sherpa/sherpa/issues/346
+@pytest.mark.parametrize('Session', [Session, AstroSession])
+def test_regression_346(Session):
```
DougBurke, line 1133: Not issue #346, but what happens if you try to create a Data object (either directly or via load_arrays) where all points are masked out? It looks like `Data1D` at least supports this, so I imagine it'll result in something like the following, but wanted to check:

```
In [20]: from sherpa.data import Data1D

In [21]: d = Data1D('x', [], [])
/home/dburke/sherpa/sherpa-master/sherpa/data.py:49: UserWarning: Converting array [] to numpy array
  warnings.warn("Converting array {} to numpy array".format(array))

In [22]: print(d)
name      = x
x         = Float64[0]
y         = Float64[0]
staterror = None
syserror  = None
```

sherpa/data.py
```diff
@@ -464,17 +487,13 @@ class Data(NoNewAttributesAfterInit, BaseData):
 
     def __init__(self, name, indep, y, staterror=None, syserror=None):
         """
```
DougBurke, line 489: I think we probably want some description added to the `Data` docstring to describe the handling of masked arrays.

sherpa/data.py
hamogu: I don't want to `del` anything, because we don't want to change the original array that the user had. I've reworded the messages, you'll see that on my next push.

hamogu:
A general thought that will help me clarify docstrings and warning messages: I suggest to accept masks on dependent arrays only. Independent and dependent arrays are input to sherpa separately, so if we don't take care of it in some way, the user could input an x where some values are masked and a y where some other values are masked. What does that even mean? Also, for binned data, we input `xlo` and `xhi`. What about bins where the lower edge is masked, but the upper edge is not. That does not make any sense.

So, I decided that only masks that are defined on the dependent array will be accepted and used to set the filter. Having any mask defined on the independent array will simply raise a warning and continue with the raw values as if the mask never existed (the current behavior for independent arrays with a mask attribute, except for an added warning). 

My thinking is that x defines some way of measurement, e.g. the energy of a spectral bin, the time of data taken, or the input voltage of a diode. `y` is then the measured quantity (e.g. the number of photons, the flux in a time series, or the light output from the diode). `x` is never invalid in that case (i.e. in principle we would make a measurement at any energy, ant time or any voltage), but `y` might be invalid (e.g. we don't have data at some energy, it was cloudy at a specific night, or the diode burned out). 

Of course, these examples only go so far (e.g. there are no negative energies, so one could a argue the in this case the independent variable should be masked), but I think we need to make a choice here to keep code and user interface understandable, so I suggest: **Masks on independent arrays are ignored with a warning, masks on dependent arrays are used to set_filter (if the True/False convention is clear, i.e. numpy masked arrays).**

After all, this is for convenience and to follow the principle of least surprise (if I put in a masked array, the mask does not just disappear silently as it currently does) - a user can always put in raw (unmasked) x and y array and set_filter after at their hearts content if they want something more complicated.

@DougBurke : Do you agree?

anetasie:
@hamogu I agree. Masking on dependent arrays make sense. On the other hand one may want to filter the dependent arrays on the values of Y, so say, ignore the data with `Y<10` for example.  That should be possible with the user function that setups the filter, which internally sets a mask on the dependent array.

hamogu:
@anetasie I'm not sure I understand where you are going with your comment. This is just about being able to use numpy masked arrays in `load_array` and using the mask in the original array to set the filter. Once the data is read in to sherpa, you can do whatever you can do now, i.e. set the filter in any way you like with `set_filter`.

hamogu:
@DougBurke Does this address your concerns?

sherpa/data.py
DougBurke: I guess the next question is whether we have any tests of this to check our logic makes sense? I know our existing test set doesn't do a good job of exploring the edge cases, so it would be interesting to know if there's any conditions that aren't evaluated.

sherpa/tests/test_data.py
DougBurke: @hamogu - do you know what happens in this case ? I didn't see any obvious test changes that exercised this, but I could well have missed it.

sherpa/data.py
hamogu: Well, I've added a LOT of tests (about a dozen, but they are all parameterized, (e.g. with and without mask) so I've lost track of the exact number, but I believe it's > 50.

DougBurke: I meant have you checked the HTML coverage output to see if anything isn't being covered? It does require running the tests with coverage turned on of course, but I think it should just be 'pytest --cov=sherpa` and then `coverage html` (I never remember the option to name the output directory, or indeed the exact commands you need to use).

DougBurke: I've rebased the branch onto master (so I can build it with a half-way recent set of software) and the coverage for the following (with XSPEC/astropy/matplotlib installed and the test dataset) indicates that lines 415 and 419 were never True, so their code was never executed. Which is why there's fancy unicode in the output below

```
409    def mask(self, val): 
410        if (val is True) or (val is False): 
411            self._mask = val 
412        # if val is of type np.bool_ and True, it failed the previous test because 
413        # "is True" compares with Python "True" singelton. 
414        # Yet, we do not want to allow arbitrary values that evaluate as True. 
415        elif val is numpy.ma.nomask: 415 ↛ 416
416            self._mask = True 
417        elif numpy.isscalar(val) and isinstance(val, numpy.bool_): 
418            self._mask = bool(val) 
419        elif (val is None) or numpy.isscalar(val): 419 ↛ 420
420            raise DataErr('ismask') 
421        else: 
422            self._mask = numpy.asarray(val, numpy.bool_) 
```

DougBurke:
@hamogu - do you want me to push the rebased version to this branch (or leave as is/do it yourself)?

sherpa/tests/test_data.py
hamogu: It behaves as it should. Your code above works exactly as before because it passes in lists, not numpy arrays and that behaviour is unchanged. If you use masked array, you can mask as much as you like. The masked datapoints are there and can be retried just as you can do after using `set_mask`. That behaviour is unchanged, too. All the changes is that instead of (assuming y is a numpy masked array)
```
d = Data1D('x', x, y)
d.set_mask(~y.mask)
```
you can now simply do `d=Data1D('x', x, y)`. It really just saves you that one line of typing. This PR is important not because it helps Sherpa do something it could not do before, but because is aligns the default to what users apparently expect (#346, #695) - and that expectation "Don't just ignore masks without even telling me you do that" is totally reasonable.

Full output from the tests @DougBurke asked for:
```
sherpa In [1]: from sherpa.data import Data1D 
          ...:                                                                                                       

sherpa In [2]: d = Data1D('x', [], [])                                                                               
/Users/guenther/anaconda3/envs/sherpaciao/lib/python3.7/site-packages/sherpa-4.12.1+301.gecfad5f4.dirty-py3.7-macosx-10.9-x86_64.egg/sherpa/data.py:49: UserWarning: Converting array [] to numpy array.

sherpa In [3]: print(d)                                                                                              
name      = x
x         = Float64[0]
y         = Float64[0]
staterror = None
syserror  = None

sherpa In [6]: xarr = numpy.ma.array([1,2,3], mask=[True, True, True])                                               

sherpa In [7]: yarr = [1, 2, 1]                                                                                      

sherpa In [10]: d = Data1D('x', xarr, yarr)                                                                          

sherpa In [11]: print(d)                                                                                             
name      = x
x         = Int64[3]
y         = Int64[3]
staterror = None
syserror  = None

sherpa In [12]: d.get_indep()                                                                                        
Out[12]: (array([1, 2, 1]),)

sherpa In [13]: d.get_indep(filter=True)                                                                             
Out[13]: (array([], dtype=int64),)
```

DougBurke: Thanks. It's an edge case to check, but someone always finds them ;-)

DougBurke: I'll leave it up to you to see if it needs an explicit test or not.

DougBurke:
So I'm just learning hot to properly decode the codecov output: the 87.5% coverage of the code added in this PR could be better, (I've noticed that my current PRs are at this level too) but there are reasons it'll never get to 100% for this PR, such as us not having a test on travis that uses crates [this is something I've been thinking about for a while]

hamogu:
In astropy, I usually find is more instructive to look at the differences, i.e. a new piece of code should not add untested lines (with some exceptions).
And while Travis probably should have a setup with crates, I don't think it's worth the effort to merge the codecov outputs together or run the codecov twice.

DougBurke:
@hamogu - the codecov output is already the merge of 4 (I think) runs; it is done automatically for us.

sherpa/data.py
DougBurke: I think we want to make sure to have a test of 'val is numpy.ma.nomask' evaluating to True.

hamogu: This line is only triggered when directly setting the mask (when the mask is inferred from the mask attribute of an array on initialization, then it's already handelded i nthe `_check_y` function and never gets to this point. Added test.

DougBurke: Thanks,

hamogu:
@dtnguyen2 : Ping! Just waiting for your review.

dtnguyen2:
OK, looking at it now

hamogu:
@dtnguyen2 The line you pointed out is not required for this change, thus I did not touch it. But, as you say, it's good to be consistent, so I just pushed the change.

PR #791, opened 2020-04-17
Fix ignore/notice error-ing out when all bins have been set bad #790

DougBurke:
# Aim

Allow notice and ignore to be called on a dataset which has no "good" bins after ignore_bad has been called. The calls do not change anything (a screen message is displayed), which is similar to how you can call ignore multiple times.

# Details

The fix is simple: an internal method now raises a `DataErr` if there are no bins, and this error is caught by notice/ignore (well, a lower level class method) and a screen message is created to say the dataset is being ignored. I am hoping that there are no unforseen consequences, but the main change is that we now get a slightly-more useful error raised in problem cases.

Add test cases for bug #790 (Unable to filter PHA data sets with no "good" groups). These test out channel, energy, and wavelength handling.

PR #794, opened 2020-04-19
Get linux libgfortran from conda-forge

### LINKED ISSUE #792: travis tests on linux fail because libgfortran has been removed from conda
We need libgfortran for our XSPEC builds. We don't for other builds.

e.g. see partial success at https://travis-ci.org/github/sherpa/sherpa/builds/676563145 when removing libgfortran requirement from `travis/setup_conda.sh` in d0ce474b8f3429c51ba9c471a39b79e640931212

hamogu:
Update the Travis build setup so that libgfortran is picked up from the conda-forge channel for both Linux and macOS builds (previously it was only done for macOS). This is needed since libgfortran is no-longer provided for Linux builds on the conda channel, causing the tests to fail on Travis (issue #792).

As libgfortran is only needed when building against the CXC-provided XSPEC model-only library conda package, the changes are only done for the XSPEC builds. This should hopefully speed up the non-XSPEC runs slightly (as there's no need to download libgfortran).

This PR only changes files related to the Travis-CI testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: c
 
-sudo: false
+dist: trusty
+os: linux
 
-matrix:
+jobs:
   fast_finish: true
   include:
     # macOS build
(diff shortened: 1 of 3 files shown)


hamogu:
@DougBurke Do you have permission to restart individual jobs? I can't in this repro, I can only change something in the code so that I can push a second commit to make Travis run again. I'm hoping the failure is transient and not real.

DougBurke:
I've restarted https://travis-ci.org/github/sherpa/sherpa/jobs/677047535 since it looked an odd failure, as you said.

DougBurke:
Hmmm, it still seems to fail. I'm guessing it's pytest-xvfb related.

DougBurke:
[I'm busy this morning/afternoon; will hopefully get some time to look at this today]

hamogu:
I compared the failed log with the last time that particular test passed. A lot of package are coming from conda-forge now that were pulled in from Anaconda before. That's kind of the point here, because we can't get libgfortran otherwise. As a consequence of this, there are a lot of minor version mismatches (e.g. tk 8.6.10 vs 8.6.8, matplotlib 2.2.5 vs. 2.2.3) any one of which could be the cause here.  It might help to change the order of channels to give Anaconda a higher priority, but, as I said above, this does not seem a good use of our time. Other tests with the same matplotlib version pass and we did not change any of the plotting related code. Even if we find the minor version increases that causes this, it could happen again and we get a game of whack-a-mole. In particular, the test that fail is a fairly old version (matplotlib 2, python 3.6) that might not be supported in the next major sherpa release anyway.

DougBurke:
I just added a minor clean up of the travis setup to remove some warnings that Travis has been reporting for a while now. I don't expect this to change the failing test.

hamogu:
Well, since you moved all jobs from "matrix" for top-level jobs, we can make it an "allowed failure" (https://docs.travis-ci.com/user/customizing-the-build/). I think that's appropriate here. The tests actually pass as you can see when inspecting the log by hand. The failure is not due to sherpa - it's from headless testing, which is not a use case we want to test here explicitly. The allowed to fail marker can be removed again later when we don't need it (or change the setup anyway). I'll wait for the tests to finish before I do that though.

DougBurke:
@hamogu - when the tests finish I'm going to add a more-invasive change - to only use conda-forge when we are building XSPEC - as I'd like to see if that works. Can you hold off your edit?

hamogu:
Sure. Maybe it won't even be needed any longer!

DougBurke:
Let's see how much damage this one simple patch can do!

DougBurke:
kermit-flail

hamogu:
Great! Good job! We should probably now rebase all other PRs with failing tests on top of this one, so that we get meaningful check marks (#791 [*], #793, #798)?

*: After backing out the test commenting done there.

Marie-Terrell:
This looks like the perfect way to fix this in the interim. Limiting "conda-forge" to just the xspec jobs actually makes this a bit of a safer holdover fix. 
The travis.yaml updates don't seem like they would be needed to fix the issue at hand, but I see no issue with taking them at the same time. The osx run overrides the "linux" default, so I am not concerned there. 
Re-building the xspec packages to remove the "conda-forge" dep (and the need to explicitly set the libgfortran version) is the next step, but that will come in a later PR. 
Thank you both!

Issue #762, opened 2020-03-20
Cannot regrid a composite model (BinaryOpModel)


DougBurke:
In testing out #747 I discovered that you can't regrid a composite model, at least one that's an instance of `BinaryOpModel`, as shown below.  This is not an issue with #747 as it is also true on the master branch.

Note that `(gal * pl).regrid(...)` is technically different than `gal.regrid(..) * pl.regrid(...)`, but at the moment I can't off-hand think of a case where this makes a significant difference for Sherpa (most `regrid` uses for Sherpa will be with psf- or "convolution"- style models which accept `regrid` and can be sent a `BinaryOpModel`to run on), so I'm not sure how much of an issue this is.

# Example

In the following I can regrid one, or both, of the components of the model expression (e.g. `gal` or `pl`) but not the combination:
  
```
sherpa In [12]: smdl = get_source()

sherpa In [13]: type(smdl)
Out[13]: sherpa.models.model.BinaryOpModel

sherpa In [14]: smdl.name
Out[14]: '(xsphabs.gal * xspowerlaw.pl)'

sherpa In [15]: gal.regrid([0.1, 0.2, 0.3, 0.4])
Out[15]: <RegridWrappedModel model instance 'regrid1d(xsphabs.gal)'>

sherpa In [16]: smdl.regrid([0.1, 0.2, 0.3, 0.4])
AttributeError: 'BinaryOpModel' object has no attribute 'regrid'
```

dtnguyen2:
Just out of curiosity, what happens with simultaneous data (especially if the different data sets have very different data-space?)

anetasie:
@dtnguyen2 the original idea was to use the highest resolution grid. I have not checked what's happening with the current setting.

DougBurke:
My understanding is that simultaneous fits are handled by bundling all the data into a `DataSimulFit` object and the models into a `SimulFitModel` object, and then using this with a `Fit` object. So we could add logic to the *Simul* objects to identify common grids to use.

I don't think this answers @dtnguyen2's question though.

Issue #300, opened 2016-11-06
support evaluating the model on a different grid to the "final" grid


DougBurke:
There are cases when it is necessary to evaluate a model (a single model or a composite one) on a different grid to the desired output grid. An example is convolution style models - such as PSF, RMF, or XSPEC convolution models - which can require the model they are convolving to be evaluated on an extended grid so as to account for signal "outside" the desired grid. Sherpa already has to deal with this in some manner - e.g. ignored bins need to be included when doing the convolution, then excluded from the output - but I'm not a fan of this approach, since it seems to tie a convolution component to a grid, which is not how the rest of the system works.

I have some notes on how models are evaluated on the wiki at https://github.com/sherpa/sherpa/wiki/Model-Evaluation-in-Sherpa but no real solution. I have the feeling that there needs to be some entity other than a model component that decides on the grid, sets things up: allocates memory, converts units if appropriate (such as wavelength to keV for XSPEC models), rebins and regrids the result, handles cacheing the results. However, it's not clear that it is really a viable approach, or at what level of the stack it would live (e.g. in the Model class, a layer above it, or only in the Session layer).

I believe this is relevant to:
- #43 Generalize the image PSF to allow data sub-pixel binning
- #68 Add Python classes for XSpec convolution models
- #301 support XSPEC mixing models

DougBurke:
With #747 we seem to almost have this functionality. However, #762 (the inability to use a different grid on an expression such as `xsphabs.gal * xsvvapec.src` which then gets rebinned to the output grid) means I don't want to close this yet.

hamogu:
This is still on Warren's list wit ha task assigned to me to "review and close out". I'm not quite sure what I'm supposed to do though...
@DougBurke #762 is closed by now. Can we close this issue and open a new one if we run into something similar in the future?

DougBurke:
I agree that we can close this: new issues are more "this doesn't do what I expect" rather than the blanket "we can't even do this" that this issue covered.

hamogu:
@wmclaugh : Please tick of this issue from the task list in confluence.

PR #798, opened 2020-04-22
enable regrid for the BinaryOpModel class

dtnguyen2:
#Note 
This PR adds the regrid method to the BinaryOpModel.  This should fix issue #762 and hopefully issue #300 (?)


hamogu:
I wonder if the test failures are caused by this:
https://github.com/dtnguyen2/sherpa/commit/77a54422824bd03060c93c1a3616dc11298a2264

hamogu:
Scratch my last comment. I now see that you addressed that already. I had just not reloaded the pages recently enough.

sherpa/models/tests/test_regrid.py
```diff
@@ -292,6 +292,17 @@ def tst_runtime_interp(model, requested, interp):
     assert ygot == approx(yexpected)
 
 
+def test_regrid_binaryop():
+    """issue #762, Cannot regrid a composite model (BinaryOpModel)"""
+    import sherpa.astro.ui as ui
+    ui.dataspace1d(-5, 15, 1, dstype=Data1D)
+    b = Box1D()
+    c = Const1D
```
hamogu, line 300: These two lines are not used in the test and can be removed I think.

sherpa/models/tests/test_regrid.py
```diff
@@ -292,6 +292,17 @@ def tst_runtime_interp(model, requested, interp):
...
+    ui.set_source(ui.const1d.c1 + ui.box1d.b1)
+    mdl = ui.get_source()
+    mdl.regrid(np.linspace(-5, 15))
```
hamogu, line 303: I would feel more comfortable if the regridded model was used for anything, i.e. call it on some number and check the return value. Right now, that's going to work (the code change is so simple, what could go wrong?), but if we ever change the implementation, it would be better to test that the regridded model actually works.

dtnguyen2: Regarding the code change is so simple, that is the beauty of recursive code.  BTW, if you comment out the three lines of the code change you will find that test will fail but I will add something

hamogu: I know it will. I suggested that to make sure we test enough for the future. I've seen bugs in other software where code was run and tests thus passed, but the numerical result was wrong because the tests did not check the return value. For example, in this case, the following regrid function would also make the test pass:
```
for part in self:
   pass
```
So, I think it's important to actually *use* the regridded function in some way and make sure that something has actually changed as a result of the regrid.

sherpa/models/model.py
```diff
@@ -556,6 +556,16 @@ def teardown(self):
     def apply(self, outer, *otherargs, **otherkwargs):
         return NestedModel(outer, self, *otherargs, **otherkwargs)
 
+    def regrid_1d(self, *arrays, **kwargs):
+        valid_keys = ('interp')
```
DougBurke, line 560: At the moment this is okay, but it should be

```
valid_keys = ('interp',)
```

to make it a tuple.

sherpa/models/model.py
```diff
@@ -640,6 +642,12 @@ def calc(self, p, *args, **kwargs):
                               type(self.rhs).__name__, len(rhs)))
         return val
 
+    def regrid(self, *arrays, **kwargs):
+        result = self.regrid_1d(*arrays, **kwargs)
+        for part in self:
+            part.regrid(*arrays)
+        return result
+
```
DougBurke, line 650: I have two issues here:

a) this will only work for 1D models

b) does the `part.regrid` actually do anything useful (it returns no value, and I don't think it changes anything within the object)

DougBurke:
Unfortunatey the kwarg stuff fails with XSPEC models (I don't think we have a good story on how we pass extra arguments down to the models):

```
In [1]: import numpy as np

In [2]: from sherpa.astro import ui

In [3]: gal = ui.xsphabs.gal

In [4]: pl = ui.xspowerlaw.pl

In [5]: mdl = gal * pl

In [6]: pl([0.1, 0.2, 0.3, 0.4])
Out[6]: array([0.69314718, 0.40546511, 0.28768207, 0.        ])

In [7]: mdl([0.1, 0.2, 0.3, 0.4])
Out[7]: array([0.00000000e+00, 9.06040486e-26, 3.64853381e-10, 0.00000000e+00])

In [8]: mdl2 = mdl.regrid([0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45])

In [9]: mdl([0.1, 0.2, 0.3, 0.4])
Out[9]: array([0.00000000e+00, 9.06040486e-26, 3.64853381e-10, 0.00000000e+00])

In [10]: mdl2([0.1, 0.2, 0.3, 0.4])
<ipython-input-10-69ee3ff34ab2> in <module>
----> 1 mdl2([0.1, 0.2, 0.3, 0.4])

~/sherpa/sherpa-pr/sherpa/models/model.py in __call__(self, *args, **kwargs)
    263             return self
    265

~/sherpa/sherpa-pr/sherpa/models/model.py in calc(self, p, *args, **kwargs)
    775
    776     def calc(self, p, *args, **kwargs):
--> 777         return self.wrapper.calc(p, self.model.calc, *args, **kwargs)
    778
    779     def get_center(self):

~/sherpa/sherpa-pr/sherpa/models/regrid.py in calc(self, pars, modelfunc, *args, **kwargs)
    592         requested_eval_space = self._make_and_validate_grid(args)
    593
--> 594         return self._evaluate(requested_eval_space, pars, modelfunc, **kwargs)
    595
    596     def _make_and_validate_grid(self, args_array):

~/sherpa/sherpa-pr/sherpa/models/regrid.py in _evaluate(self, data_space, pars, modelfunc, **kwargs)
    686                                             data_space.midpoint_grid,
    687                                             eval_space.midpoint_grid,
--> 688                                             **kwargs)
    689
    690

~/sherpa/sherpa-pr/sherpa/models/regrid.py in eval_non_integrated(self, pars, modelfunc, data_space, eval_space, **kwargs)
    641         my_eval_space = np.unique(np.append(eval_space, data_space[indices]))
    642
--> 643         y_tmp = modelfunc(pars, my_eval_space, **kwargs)
    644         y_interpolate = self.method(data_space, my_eval_space, y_tmp)
    645

~/sherpa/sherpa-pr/sherpa/models/model.py in calc(self, p, *args, **kwargs)
    633     def calc(self, p, *args, **kwargs):
    634         nlhs = len(self.lhs.pars)
--> 635         lhs = self.lhs.calc(p[:nlhs], *args, **kwargs)
    636         rhs = self.rhs.calc(p[nlhs:], *args, **kwargs)
    637         try:

~/sherpa/sherpa-pr/sherpa/models/model.py in cache_model(cls, pars, xlo, *args, **kwargs)
     66                 return cache[digest].copy()
     67
---> 68         vals = func(cls, pars, xlo, *args, **kwargs)
     69
     70         if use_caching:

~/sherpa/sherpa-pr/sherpa/astro/xspec/__init__.py in calc(self, *args, **kwargs)
    838         #  - it allows for the user to find out what bins are bad,
    839         #    by directly calling the _calc function of a model
--> 840         out = self._calc(*args, **kwargs)
    841
    842         # This check is being skipped in the 4.8.0 release as it

TypeError: xsphab() takes no keyword arguments
```

sherpa/models/model.py
DougBurke: I've just tried commenting out the 'for part in self' case and the tests still pass, so I really don't understand what it's doing.

anetasie: Why `def regrid()`  is defined twice:  line 588 and 645?

DougBurke: @anetasie - these are different classes.

dtnguyen2:
This bug has existed before this PR, I modified your script to the following:

```
# mdl = gal * pl                                                                
mdl = gal
```
and I get: 

```
...
  File "/export/sherpa/sherpa/astro/xspec/__init__.py", line 835, in calc
    out = self._calc(*args, **kwargs)
TypeError: xsphab() takes no keyword arguments

> /export/sherpa/sherpa/astro/xspec/__init__.py(835)calc()
-> out = self._calc(*args, **kwargs)
```

DougBurke:
I've looked at how the models are called and I can confirm (with some well placed print calls) that a model like

```
mdl = ui.const1d.con * (ui.powlaw1d.pl + ui.polynom1d.poly)
mdl2 = mdl.regrid(...)
```

does (for `mdl2`) evaluate, and combine, the components at the higher resolution before being binned down to the "dataspace" grid.

:+1:

DougBurke:
@dtnguyen2 - for https://github.com/sherpa/sherpa/pull/798#issuecomment-646191544 I can confirm that the regrid of xsphabs does error out in CIAO 4.12 (and does for other XSPEC models to). I don't understand this because I know I have done testing on the regrid support with XSPEC models.

dtnguyen2:
That is what I was trying to show in my test:

```
    fit = Fit(data, mymodel, leastsq, levmar)
    result = fit.fit()
    assert result.numpoints == x.size
...
    assert (result.nfev + 4) * x.size == mygauss.counter
```

```
    fit = Fit(data, mymodel, leastsq, levmar)
    result = fit.fit()
    assert result.numpoints == x.size
...
    assert (result.nfev + 4) * x.size == mygauss.counter
```

```numpoints``` is always x.size but the ```counter``` within each model functions are different for the two cases

**Comment by dtnguyen2** [comment]
I will do a ```git bisect``` in a second, do you happen to know (roughly) when this was done?

**Comment by DougBurke** [comment]
@dtnguyen2 - I should say I "believe" I did testing with the XSPEC models. However, the more I look at things the more I think that I can't have been using the XSPEC models, since the actual bit that causes the kwargs error has been there for a long time (it's just that until now the methods weren't called with any kwargs).

I'm going to open a new ticket to describe this since there is also a related issue with the XSPEC models when using the regrid code, and I think it's out of scope of this fix.

**Comment by anetasie** [comment]
I'm wondering if the issue is related to the 'integrated' pha type data and xspec expects the edges of the bins to evaluate a model. This fails for me for mdl2:

```
dx = 1
xgrid = np.arange(1, 5, dx)
grid_lo = xgrid[:-1]
grid_hi = xgrid[1:]

mdl = scale1d.s1*(powlaw1d.pl + polynom1d.poly)
mdl(grid_lo,grid_hi)

dx = 0.2
hxgrid = np.arange(1, 5, dx)
hgrid_lo = hxgrid[:-1]
hgrid_hi = hxgrid[1:]
mdl2 = mdl.regrid(hgrid_lo,hgrid_hi)
mdl2([grid_lo,grid_hi])
ModelErr                                  Traceback (most recent call last)
<ipython-input-9-516dc5694162> in <module>
----> 1 mdl2([grid_lo,grid_hi])

/data/qednew/git/sherpa/sherpa/models/model.py in __call__(self, *args, **kwargs)
    263             return self
    265 

/data/qednew/git/sherpa/sherpa/models/model.py in calc(self, p, *args, **kwargs)
    767 
    768     def calc(self, p, *args, **kwargs):
--> 769         return self.wrapper.calc(p, self.model.calc, *args, **kwargs)
    770 
    771     def get_center(self):

/data/qednew/git/sherpa/sherpa/models/regrid.py in calc(self, pars, modelfunc, *args, **kwargs)
    590             return modelfunc(pars, *args, **kwargs)
    591 
--> 592         requested_eval_space = self._make_and_validate_grid(args)
    593 
    594         return self._evaluate(requested_eval_space, pars, modelfunc, **kwargs)

/data/qednew/git/sherpa/sherpa/models/regrid.py in _make_and_validate_grid(self, args_array)
    615         # Ensure the two grids match: integrated or non-integrated.
    616         if self.evaluation_space.is_integrated and not requested_eval_space.is_integrated:
--> 617             raise ModelErr('needsint')
    618         if requested_eval_space.is_integrated and not self.evaluation_space.is_integrated:
    619             raise ModelErr('needspoint')

ModelErr: A non-overlapping integrated grid is required for model evaluation,
e.g. [0.1,0.2],[0.2,0.3]
```
But the evaluation space seems ok:
```
In [10]: print(grid_lo)                                                                         
[1 2 3]
In [11]: print(grid_hi)                                                                         
[2 3 4]
```

**Comment by anetasie** [comment]
This was with this version:
```
In [13]: sherpa.__version__                                                                     
Out[13]: '4.12.0+186.gcff6d05.dirty'
```

**Comment by DougBurke** [comment]
@anetasie the issue I'm talking about is nothing to do with the issue you have.

Your problem is an error, as you have said

```mdl2([grid_lo,grid_hi])```

when it should be

```mdl2(grid_lo,grid_hi)```

(ie no array around the grids)

**Comment by anetasie** [comment]
@DougBurke ok I guess there is a difference between
`mld2([0.1,0.2,0.3])` failure in your comment above and mine, so yes `mdl2(grid_hi,grid_lo)` works fine. Thanks.

**Comment by DougBurke** [comment]
Here is an example of a problem I see when trying out the support for combining 2D models (it's all a bit messy as the 2D interface could/should be improved, but that's a different issue):

```
import numpy as np
from sherpa.astro import ui

y0, x0 = np.mgrid[20:29, 10:20]
y0 = y0.flatten()
x0 = x0.flatten()

gmdl = ui.gauss2d.gmdl
gmdl.fwhm = 14
gmdl.xpos = 15
gmdl.ypos = 24
gmdl.ampl = 10

cmdl = ui.const2d.cmdl
cmdl.c0 = 4

xr1 = np.arange(10, 20, 1)
yr1 = np.arange(20, 29, 1)

rmdlg = gmdl.regrid(xr1, yr1)
rmdlc = cmdl.regrid(xr1, yr1)

shape = (9, 10)
truthg = gmdl(x0, y0).reshape(shape)
truthc = cmdl(x0, y0).reshape(shape)
truth = truthg + truthc

ans1 = rmdlg(x0, y0).reshape(shape)
ans2 = rmdlc(x0, y0).reshape(shape)

if (ans1 != truthg).any():
    print("gauss model difference")

if (ans2 != truthc).any():
    print("const model difference")

rmdl = (gmdl + cmdl).regrid(xr1, yr1)

ans3 = rmdl(x0, y0).reshape(shape)

if (ans3 != truth).any():
    print("combined model difference")
```

I can't guarantee that the final check will work (ie it may need a pytest.approx style test rather than equality), but we don't even get that far because

```
  File "binop.py", line 45, in <module>
    rmdl = (gmdl + cmdl).regrid(xr1, yr1)
  File "/home/dburke/sherpa/sherpa-python38/sherpa/models/model.py", line 581, in regrid
    regridder._make_and_validate_grid(arrays)
  File "/home/dburke/sherpa/sherpa-python38/sherpa/models/regrid.py", line 623, in _make_and_validate_grid
    if np.any(lo[1:] < hi[:-1]) or np.any(lo == hi):
ValueError: operands could not be broadcast together with shapes (9,) (8,)
```

and I believe this is happening because it's assuming 1D in the regrid when evaluating the combined expression.

**Comment by dtnguyen2** [comment]
If a make a one line change to the _sherpa/models/model.py_ file:
```
(sherpa) [dtn@devel12 sherpa]$ git diff sherpa/models/model.py
diff --git a/sherpa/models/model.py b/sherpa/models/model.py
index 170e24b5..1b35a640 100644
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -601,7 +601,7 @@ class UnaryOpModel(CompositeModel, ArithmeticModel):
         return self.op(self.arg.calc(p, *args, **kwargs))
 
 
-class BinaryOpModel(CompositeModel, RegriddableModel1D):
+class BinaryOpModel(CompositeModel, RegriddableModel2D):
 
     @staticmethod
     def wrapobj(obj):
```
I can get, @DougBurke's script to work:

```
(sherpa) [dtn@devel12 sherpa]$ python fubar.py; echo $?
0
```

Unfortunately, it will no longer work for 1D models. Bummer!  Let me think about this one a bit

**Comment by DougBurke** [comment]
Our model stack tends to try to avoid committing to a dimensionality  until you get to the actual model implementations. Maybe we need to rethink this (I think @olaurino had been thinking about this in some of his rework).

**Comment by olaurino** [comment]
IIRC there is now a distinction between 1D and 2D models, because it matters to the regridding code. The least painful way of doing it was to introduce `RegriddableModel1D|2D` as extensions of `ArithmeticModel`. So if I am not mistaken what you would need now is to define the behavior in case of a mismatch and implement error handling code in that case. It might be ugly and require to check the classes explicitly, unless you define an attribute/abstract method in `ArithmeticModel` that the 1D|2D classes will set to `1` or `2`. The hierarchy of models is complex, so there may be a number of corner cases to check for (e.g. what about table and template models?), but tbh the first thing you'll need, I think, is a definition of the error handling cases from a UI point of view when there is a mismatch in the number of dimensions for models that are being stringed together.

Does that make any sense?

**Comment by olaurino** [comment]
Re `BinaryOpModel` how about you make it extend `ArithmeticModel`, you check the dimensionality of the models being stringed together, and you have `BinaryOpModel` delegate the regridding methods to a specific implementation of `RegriddableModel1D|2D` that is instantiated on the fly when the `BinaryOpModel` class is instantiated? Again, at that point you will necessarily have to add an attribute or abstract method to `ArithmeticModel` because then checking the dimensionality of a `BinaryOpModel` based on the class wouldn't work anymore.

**Comment by dtnguyen2** [comment]
I think I have a fix for this, I have included @DougBurke's script in the test file:

```
sherpa) [dtn@devel12 sherpa]$ git diff sherpa/models/tests/test_regrid.py 
diff --git a/sherpa/models/tests/test_regrid.py b/sherpa/models/tests/test_regrid.py
index e4f5a1cb..a02bd32a 100644
--- a/sherpa/models/tests/test_regrid.py
+++ b/sherpa/models/tests/test_regrid.py
@@ -27,7 +27,7 @@ from sherpa.astro.data import DataIMG, DataIMGInt
 from sherpa.astro.ui.utils import Session
 from sherpa.data import Data1DInt, Data1D
 from sherpa.models.basic import Box1D
-from sherpa.models import Const1D, RegriddableModel1D, Parameter, Const2D, RegriddableModel2D, ArithmeticModel
+from sherpa.models import Const1D, RegriddableModel1D, Parameter, Const2D, RegriddableModel2D, ArithmeticModel, Gauss2D
 from sherpa.utils.err import ModelErr
 from sherpa.utils import neville, linear_interp
 from sherpa.utils import akima
@@ -292,7 +292,7 @@ def test_runtime_interp():
     assert ygot == approx(yexpected)
 
 
-def test_regrid_binaryop():
+def test_regrid_binaryop_1d():
     """issue #762, Cannot regrid a composite model (BinaryOpModel)"""
     from sherpa.fit import Fit
@@ -358,6 +358,40 @@ def test_regrid_binaryop():
     assert mygauss.counter == myconst.counter
     assert (result.nfev + 4) * x_regrid.size  == mygauss.counter
 
+def test_regrid_binaryop_2d():
+    y0, x0 = np.mgrid[20:29, 10:20]
+    y0 = y0.flatten()
+    x0 = x0.flatten()
+
+    gmdl = Gauss2D()
+    gmdl.fwhm = 14
+    gmdl.xpos = 15
+    gmdl.ypos = 24
+    gmdl.ampl = 10
+
+    cmdl = Const2D()
+    cmdl.c0 = 4
+
+    xr1 = np.arange(10, 20, 1)
+    yr1 = np.arange(20, 29, 1)
+
+    rmdlg = gmdl.regrid(xr1, yr1)
+    rmdlc = cmdl.regrid(xr1, yr1)
+
+    shape = y0.shape
+    truthg = gmdl(x0, y0).reshape(shape)
+    truthc = cmdl(x0, y0).reshape(shape)
+    truth = truthg + truthc
+
+    ans1 = rmdlg(x0, y0).reshape(shape)
+    ans2 = rmdlc(x0, y0).reshape(shape)
+    assert (ans1 != truthg).any() == False
+    assert (ans2 != truthc).any() == False
+
+    rmdl = (gmdl + cmdl).regrid(xr1, yr1)
+    ans3 = rmdl(x0, y0).reshape(shape)
+    assert (ans3 != truth).any() == False
+
```

The test pass:

```
(sherpa) [dtn@devel12 sherpa]$ pytest sherpa/models/tests/test_regrid.py 
platform linux -- Python 3.7.3, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
rootdir: /export/sherpa, inifile: pytest.ini
plugins: arraydiff-0.3, xdist-1.29.0, forked-1.0.2, openfiles-0.3.2, doctestplus-0.3.0, remotedata-0.3.1
collected 11 items                                                             

sherpa/models/tests/test_regrid.py ...........                           [100%]

========================== 11 passed in 0.88 seconds ===========================
```

I just have to clean up the code

File: sherpa/models/model.py [hunk 5]
```diff
@@ -572,21 +578,25 @@ def regrid(self, *arrays, **kwargs):
...
         """
-        valid_keys = ('interp', )
-        for key in kwargs.keys():
-            if key not in valid_keys:
-                raise TypeError("unknown keyword argument: '%s'" % key)
-        eval_space = EvaluationSpace1D(*arrays)
+        try:
+            self.check_regrid_kwargs(**kwargs)
+        except TypeError as te:
+            raise te
```
DougBurke, line 593: You don't need the `try`/`except` case here as it doesn't add anything (the exact same exception is re-raised).

sherpa/models/model.py
```diff
@@ -540,6 +540,12 @@ def __setstate__(self, state):
     def __getitem__(self, filter):
         return FilterModel(self, filter)
 
+    def check_regrid_kwargs(self, **kwargs):
```
DougBurke, line 549: I can understand pushing this into the `ArithmeticModel` class, but the name is a bit odd as we don't have the concept of regrid at this level of the API.

sherpa/models/model.py
```diff
@@ -572,21 +578,25 @@ def regrid(self, *arrays, **kwargs):
...
-        return regridder.apply_to(self)
+        regridder._make_and_validate_grid(args)
+        return regridder.apply_to(kwargs.get('_self', self))
 
 
 class RegriddableModel2D(ArithmeticModel):
-    def regrid(self, *arrays):
-        eval_space = EvaluationSpace2D(*arrays)
+    def regrid(self, *args, **kwargs):
+        try:
```
DougBurke, line 602: the try/except isn't needed,

sherpa/models/model.py
```diff
@@ -558,7 +564,7 @@ def apply(self, outer, *otherargs, **otherkwargs):
 
 
 class RegriddableModel1D(ArithmeticModel):
-    def regrid(self, *arrays, **kwargs):
+    def regrid(self, *args, **kwargs):
```
DougBurke, line 576: Given that the `_self` argument is now required, it would be useful to have it visible in the signature, and to have documentation about it.

sherpa/models/model.py
dtnguyen2: One could introduce a class called ```RegriddableModel```, from which the classes ```RegriddableModel1D``` and ```RegriddableModel2D``` inherit from and the method ```check_regrid_kwargs``` live there?

sherpa/models/model.py
dtnguyen2: I'm not sure I understand what you are suggesting here.  The only time ```self``` needs to be passed if from the class ```BinaryOpModel``` when it calls the method ```regrid``` of the classes ```RegriddableModel1D``` and ```RegriddableModel2D``` and since the method ```regrid``` can be called by the user, I don't see how one can make ```self``` as required argument.

DougBurke: @dtnguyen2 - what I was saying was that I expect there to be some documentation on what the  _self argument is. I mis-spole when I said it was required, what I meant is that it is part of theAPI.

dtnguyen2:
Managed to remove the passing of ```self``` through the  ```_self``` argument:

```
(sherpa) [dtn@devel12 sherpa]$ grep _self sherpa/models/model.py | wc
      0       0       0
```

However, not sure why I needed to have a couple of instances of the python ```isinstance``` calls in the ```BinaryOpModel``` class.  One would think that if ```regrid1D``` and ```regrid2D``` remain in their respective ```regrid``` methods than in the ```BinaryOpModel``` than one call just call their respective ```regrid``` method.  If I have time, I will continue to see if the call to ```isinstance``` can be removed.

DougBurke:
@dtnguyen2  - how about 

```
class RegriddableModel(ArithmeticModel):
    def check_regrid_kwargs(...):
        as current

    def regrid(...)
        raise NotImplementedError
```

and then move the regrid1D/2D methods into the `RegriddableModel1D` and `RegriddableModel2D` classes as the implementation of the `regrid` method (so there's no `regrid1D` or `regrid2D` methods). Then the `BinaryOpModel` class's `regrid` method should be

```
    def regrid(self, *args, **kwargs):
        for part in self.parts:
            ?? = part.regrid(*args, **kwargs)
```

Note that you use `part` and not `self` to define the object in the `regrid` method.

However, the ?? is to indicate that I think we need to work out what we want here. First we need to return something that represents the composite model, so just calling return on the first part isn't going to work. Secondly, although we could pass the regrid method to the components, we would have to handle those classes with no regrid support, such as ArithmeticConstantModel, and I actually think we don't want that. We want to evaluate the combined model (e.g. a*(b + c)) using the high-res grid, and then regrid it to the required grid. We don't want to be evaluating each model component on the grid, then regrid it onto the required grid, then combine the required grid values.

dtnguyen2:
As you know the code in the methods ```regrid1D``` and ```regrid2D``` used to reside in the ```regrid``` methods of the class ```RegriddableModel1D``` and ```RegriddableModel2D```, respectively. The ```regrid``` method of the class ```BinaryOpModel``` would be:

```
     def regrid(self, *args, **kwargs):
         for part in self.parts:
            return part.regrid(*args, **kwargs)
```

As you can see, the following code:

```
(sherpa) [dtn@devel12 sherpa]$ git diff sherpa/models/model.py
diff --git a/sherpa/models/model.py b/sherpa/models/model.py
index b4a2bb43..3a86c9ac 100644
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -572,19 +572,6 @@ class RegriddableModel(ArithmeticModel):
     def regrid(self, *args, **kwargs):
         raise NotImplementedError
 
-    def regrid1D(self, *args, **kwargs):
-        self.check_regrid_kwargs(**kwargs)
-        eval_space = EvaluationSpace1D(*args)
-        regridder = ModelDomainRegridder1D(eval_space, **kwargs)
-        regridder._make_and_validate_grid(args)
-        return regridder.apply_to(self)
-
-    def regrid2D(self, *args, **kwargs):
-        self.check_regrid_kwargs(**kwargs)
-        eval_space = EvaluationSpace2D(*args)
-        regridder = ModelDomainRegridder2D(eval_space)
-        return regridder.apply_to(self)
-
 
 class RegriddableModel1D(RegriddableModel):
     def regrid(self, *args, **kwargs):
@@ -601,12 +588,20 @@ class RegriddableModel1D(RegriddableModel):
         >>> request_space = np.arange(1, 10, 0.1)
         >>> regrid_model = mybox.regrid(request_space, interp=linear_interp)
         """
-        return self.regrid1D(*args, **kwargs)
+        self.check_regrid_kwargs(**kwargs)
+        eval_space = EvaluationSpace1D(*args)
+        regridder = ModelDomainRegridder1D(eval_space, **kwargs)
+        regridder._make_and_validate_grid(args)
+        return regridder.apply_to(self)
 
 
 class RegriddableModel2D(RegriddableModel):
     def regrid(self, *args, **kwargs):
-        return self.regrid2D(*args, **kwargs)
+        self.check_regrid_kwargs(**kwargs)
+        eval_space = EvaluationSpace2D(*args)
+        regridder = ModelDomainRegridder2D(eval_space)
+        return regridder.apply_to(self)
+
 
 class UnaryOpModel(CompositeModel, ArithmeticModel):
 
@@ -639,12 +634,7 @@ class BinaryOpModel(CompositeModel, RegriddableModel):
 
     def regrid(self, *args, **kwargs):
         for part in self.parts:
-            if isinstance(part, RegriddableModel1D):
-                return self.regrid1D(*args, **kwargs)
-            elif isinstance(part, RegriddableModel2D):
-                return self.regrid2D(*args, **kwargs)
-            else:
-                raise NotImplementedError('unknown class type')
+            return part.regrid(*args, **kwargs)
 
     def startup(self, cache=False):
         self.lhs.startup(cache)
```
unfortunately, the code fails the limit tests:
```
(sherpa) [dtn@devel12 sherpa]$ pytest sherpa/models/tests/test_regrid.py 
platform linux -- Python 3.7.3, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
rootdir: /export/sherpa, inifile: pytest.ini
plugins: arraydiff-0.3, xdist-1.29.0, forked-1.0.2, openfiles-0.3.2, doctestplus-0.3.0, remotedata-0.3.1
collected 11 items                                                             

sherpa/models/tests/test_regrid.py .........FF                           [100%]

```

It is difficult to show what goes on without tracing the code yourself, so the best that one can do is to try print between the two approaches (the one that you suggested which I tried to get it to work and the current PR).  As you can see on line 37c50 only the ```gauss2d``` model is passed to the method ```regrid``` of the class ```RegriddableModel2D``` vs the entire model ```(gauss2d.gmdl + const2d.cmdl)``` to the ```regrid2D``` method of the class ```RegriddableModel``` which was why in the previous iteration of the PR the ```self``` was passed (ie so the entire model expression is passed to the ```regrid``` method.

```
(sherpa) [dtn@devel12 regrid_binaryop]$ diff a b
13a14,22
> RegriddableModel::regrid2D: self = gauss2d.gmdl
>    Param        Type          Value          Min          Max      Units
>    -----        ----          -----          ---          ---      -----
>    gmdl.fwhm    thawed           14  1.17549e-38  3.40282e+38           
>    gmdl.xpos    thawed           15 -3.40282e+38  3.40282e+38           
>    gmdl.ypos    thawed           24 -3.40282e+38  3.40282e+38           
>    gmdl.ellip   frozen            0            0        0.999           
>    gmdl.theta   frozen            0     -6.28319      6.28319    radians
>    gmdl.ampl    thawed           10 -3.40282e+38  3.40282e+38           
17a27,30
> RegriddableModel::regrid2D: self = const2d.cmdl
>    Param        Type          Value          Min          Max      Units
>    -----        ----          -----          ---          ---      -----
>    cmdl.c0      thawed            4 -3.40282e+38  3.40282e+38           
37c50
< RegriddableModel2D::regrid: self = gauss2d.gmdl
---
> RegriddableModel::regrid2D: self = (gauss2d.gmdl + const2d.cmdl)
46c59
< combined model difference
---
>    cmdl.c0      thawed            4 -3.40282e+38  3.40282e+38           
```

DougBurke:
As I mentioned, the following is not sufficient - it just calls regrid on the first element of the parts array, not both sides.

```
     def regrid(self, *args, **kwargs):
         for part in self.parts:
-            if isinstance(part, RegriddableModel1D):
-                return self.regrid1D(*args, **kwargs)
-            elif isinstance(part, RegriddableModel2D):
-                return self.regrid2D(*args, **kwargs)
-            else:
-                raise NotImplementedError('unknown class type')
+            return part.regrid(*args, **kwargs)
```

Also, we actually want it to actually run the combined model with the "over-ridden" grid, and only rebin to the actual grid after the components have been combined. So it really should just be calling either regrid1D or regrid2D at this point. But this takes up back to some ugly code to determine the dimensionality of the parts of the binary expression.

DougBurke:
@dtnguyen2 - although the tests appear to have failed, they only did so (I believe) because of the matplotlib error. Hopefully #901 will get merged soon so you could rebase this.

DougBurke:
@dtnguyen2 - here's a check I just made using this branch: the idea is to check that the regridded model calls its composite parts with the "high-res" grid, not with the output model grid. And it does.

```
import numpy as np

import pytest

from sherpa.models import basic, model


class Wrappable1D(model.RegriddableModel1D):
    """This is a truly awful bit of code."""

    def __init__(self, cls, name):
        self.ncalled = []  # record the number of elements
        self.baseclass = cls
        self.baseclass.__init__(self, name)

    def calc(self, pars, xlo, *args, **kwargs):
        xlo = np.asarray(xlo)
        self.ncalled.append((xlo[0], xlo[-1], xlo.size))
        return self.baseclass.calc(self, pars, xlo, *args, **kwargs)


m1 = Wrappable1D(basic.Const1D, 'm1')
m2 = Wrappable1D(basic.Gauss1D, 'm2')

m2.pos = 5

xregrid = np.arange(0, 20, 0.2)
xdata = np.arange(1.5, 12.5, 0.5)

morig = m1 + m2
mwrap = morig.regrid(xregrid)

y = mwrap(xdata)

# Check both components wre called with the same grid
assert m1.ncalled == m2.ncalled

# Check that m1 was called with the expected grid (ie that
# it is larger than xdata).
#
got = m1.ncalled
assert len(got) == 1
minval, maxval, nbins = m1.ncalled[0]
assert minval == pytest.approx(0)
assert maxval == pytest.approx(19.8)

# Due to the way the grid is created, all we can
# really do is check that the number of bins
# is larger than xdata. For a regression test we
# could check it's 111 but that is
#
assert nbins > xdata.size
```

Note this doesn't check that the results are correct, just that the behavior is as we want.

DougBurke:
I've restarted the tests for the last commit as there was apparently some sort of conda issue yesterday, and I'm hoping it's fixed itself now.

DougBurke:
The code looks good to me, but it needs rebasing before we can merged as it contains changes from an unrelated PR.

DougBurke:
Closed by #919 (the rebased version of this PR)

PR #812, opened 2020-05-21
Add unsubtract and ungroup to datastack

hamogu:
# Summary

Since the datastack has a "subtract" method, it should have the matching
"unsubtract" method and the same for group/ungroup.

# Note
Note that the datastack is not documented within this repro, so the
relevant ahelp page has to be updated separately.


DougBurke:
I think out current test coverage of the datastack module isn't great. These routines should be relatively easy to add tests to in `sherpa/astro/datastack/tests/test_datastack.py`

hamogu:
@dtnguyen2 Test added, checks pass except for the spurious codecov failure due ot random tests in brag_bayes that have nothing to do with this PR and that we discussed yesterday.

DougBurke:
@hamogu - those tests look like the sort of thing we'd want - thanks!

hamogu:
@dtnguyen2 : Just waiting for your review. CI failure is from a problem how codecov is set up and has nothing to do with this PR.

dtnguyen2 (approved):

wmclaugh (approved):

PR #821, opened 2020-06-04
Add hyperlink to similarly named SHERPA package

hamogu:
This other package has a different purpose, but it's close
enough that users could (and have in the past) confused
the two.


hamogu:
CI failure is clearly not related to this PR. I restarted, but they fail again. I hope it's intermittent and I'll try again in a could of days.

PR #832, opened 2020-06-21
Support building with NumPy 1.19

DougBurke:
# Summary

Support NumPy 1.19. Note that this version of NumPy drops support for version Python 3.5.

# Details

It looks like NumPy 1.19 has changed some things

- numpy.distutils no longer exports Comnand, so get it from distutils (where it came from anyway); it's not clear whether this used to behave differently in the past
- tostring now generates deprecation warnings, so we switch to tobytes (which has been available since NumPy 1.9, which predates Python version 3.5)

### DIFF
--- a/helpers/develop.py
+++ b/helpers/develop.py
@@ -36,7 +36,7 @@ try:
                 shutil.copyfile(sherpa_config.group_location, os.path.join(os.getcwd(), 'group.so'))
 
 except ImportError:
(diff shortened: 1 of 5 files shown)


DougBurke:
Since this is more a build issue than code, I've asked if @Marie-Terrell can review this.

DougBurke:
Thanks for the fast review and merging.

PR #839, opened 2020-06-25
Add requires_fits to test_astro_sim tests and switch to pytest

DougBurke:
# Summary

Fixes an uncommon error with a simulation test, which would fail if
XSPEC support is built but there is no support for FITS files (i.e. AstroPy
or Pycrates is not installed)

# Details

In the unlikely case you have XSPEC support but do not have astropy
the tests would fail. This fixes this. It also updates the test to use pytest
rather than SherpaTestCase (and in doing cleans up the code a little bit
to remove some repeated code).


DougBurke:
@wmclaugh - I marked you for review but should I give it to @dtnguyen2 instead?

PR #842, opened 2020-06-26
Fix scaling of staterror when reading PHA file with rate instead of counts

### LINKED ISSUE #800: staterrors in PHA files with RATE column
Some PHA spectra have `RATE` column instead of `COUNTS`. This is allowed by the FITS OGIP standard. Sherpa reads the 'RATE column' and converts the rates per channel to counts per channel. However, by default the `STAT_ERR` is not used and the statistical weights are computed within `fit()` by the selected statistics. For background subtracted data, such as for example in Swift BAT spectra, the calculated errors from rates will not be correct, as this calculation will be performed on the counts (calculated from rates) and the contribution to the error from the background counts will be ignored. There is no background spectrum, or background array available to account for the errors. The `STAT_ERROR` column is present in the PHA file and these errors  should be used. One can read the error using `use_errors` option in `load_pha`, which inputs the error array, but then these input errors are used incorrectly because the exposure factor is being ignored. 

```
load_pha('bat_index_862.pha')
load_pha('bat_index_862.pha', use_errors=True)
set_model(powlaw1d.p1)                                           
fit()
print(get_data().staterror)
print(get_data().counts)
```

The fit returns a huge value of the statistics indicating that the sqrt is taken from the supplied error which is read as 'RATE', so the exposure factor has been ignored.

```
Dataset               = 1
Statistic             = chi2
Initial fit statistic = 3.41608e+26
Final fit statistic   = 8.86361e+16 at function evaluation 53
Data points           = 8
Degrees of freedom    = 6
Probability [Q-value] = 0
Reduced statistic     = 1.47727e+16
Change in statistic   = 3.41608e+26
   p1.gamma       2.19094      +/- 3.92428e-09 
   p1.ampl        0.00256067   +/- 3.43976e-11 
```

```
print(get_data().staterror)                                             
[1.5736572e-06 9.5879363e-07 1.2671527e-06 1.0345638e-06 1.0883365e-06
 7.4098654e-07 7.5431592e-07 4.8542256e-07]

print(get_data().counts) 
[631.63786804 187.89412226 254.27311475 168.07048241 150.80228231
 237.09933564  67.56084161 -10.02169898]
```
The workaround exist and one can correct the error array:

```
sherpa In [11]: er=get_data().staterror*get_exposure()                                                  
sherpa In [12]: er                                                                    
Out[12]: 
array([291.8991634 , 177.84754024, 235.04535511, 191.90221112,
       201.87656716, 137.44629504, 139.91877508,  90.04149099])

sherpa In [14]: get_data().staterror=er                                               
sherpa In [15]: get_data().staterror                                                  
Out[15]: 
array([291.8991634 , 177.84754024, 235.04535511, 191.90221112,
       201.87656716, 137.44629504, 139.91877508,  90.04149099])
```

The question is if for the `load_pha()` option with `use_errors=TRUE` Sherpa should check the header HDUCLAS KEYWORD  for RATE instead of COUNT and in case RATE convert the STAT_ERR in the same way as the RATE to COUNT per channel, so the units of the data and the error match. 
Sherpa should do that to be consistent with the OGIP standards.

The data and response file is in 
`/data/sherpa/TestIssues/BAT`

hamogu:
Note that the tests for this cannot pass until https://github.com/sherpa/sherpa-test-data/pull/23 is merged, so please review even though CI is not green.

fixes #800

### DIFF
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016, 2017, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016, 2017, 2018, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -16,41 +16,43 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+import pytest
 
-from sherpa.utils.testing import SherpaTestCase, requires_data, \
-    requires_fits, requires_xspec
+import numpy as np
+from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
 from sherpa.astro import ui
 
 from tempfile import NamedTemporaryFile
 
+@pytest.fixture
+def setup(make_data_path):
+    ui.clean()
 
-class test_89_issues(SherpaTestCase):
-    def setUp(self):
-        ui.clean()
 
-    @requires_data
-    @requires_fits
-    @requires_xspec
-    def test_mod_fits(self):
-        tablemodelfile = self.make_path("xspec-tablemodel-RCS.mod")
-        ui.load_table_model("tmod", tablemodelfile)
-        tmod = ui.get_model_component("tmod")
-        self.assertEqual("xstablemodel.tmod", tmod.name)
+@requires_data
+@requires_fits
+@requires_xspec
+def test_mod_fits(make_data_path):
+    tablemodelfile = make_data_path("xspec-tablemodel-RCS.mod")
+    ui.load_table_model("tmod", tablemodelfile)
+    tmod = ui.get_model_component("tmod")
+    assert tmod.name == "xstablemodel.tmod"
 
-    @requires_fits
-    def test_warnings_are_gone_arrays(self):
-        ui.load_arrays(1, [1, 2, 3], [4, 5, 6])
-        #  We now have logic in conftest.py to catch white-listed warnings and fail on unexpected ones.
-        #  We just need to make any warnings bubble up, here and in the following test.
-        with NamedTemporaryFile() as f:
-            ui.save_data(1, f.name, ascii=True, clobber=True)
-        with NamedTemporaryFile() as f:
-            ui.save_data(1, f.name, ascii=False, clobber=True)
+@requires_fits
+def test_warnings_are_gone_arrays():
+    ui.load_arrays(1, [1, 2, 3], [4, 5, 6])
+    #  We now have logic in conftest.py to catch white-listed warnings and fail on unexpected ones.
+    #  We just need to make any warnings bubble up, here and in the following test.
+    with NamedTemporaryFile() as f:
+        ui.save_data(1, f.name, ascii=True, clobber=True)
+    with NamedTemporaryFile() as f:
+        ui.save_data(1, f.name, ascii=False, clobber=True)
 
-    @requires_fits
-    @requires_data
-    def test_warnings_are_gone_pha(self):
-        pha = self.make_path("3c273.pi")
-        ui.load_pha(pha)
-        with NamedTemporaryFile() as f:
-            ui.save_data(1, f.name, ascii=False, clobber=True)
+@requires_fits
+@requires_data
+def test_warnings_are_gone_pha(make_data_path):
+    pha = make_data_path("3c273.pi")
+    ui.load_pha(pha)
+    with NamedTemporaryFile() as f:
+        ui.save_data(1, f.name, ascii=False, clobber=True)
+
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -56,3 +56,22 @@ def test_warnings_are_gone_pha(make_data_path):
     with NamedTemporaryFile() as f:
         ui.save_data(1, f.name, ascii=False, clobber=True)
 
+        
+@requires_fits
+@requires_data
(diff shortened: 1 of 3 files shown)


sherpa/astro/io/tests/test_io.py
```diff
@@ -16,41 +16,43 @@
...
-from sherpa.utils.testing import SherpaTestCase, requires_data, \
-    requires_fits, requires_xspec
+import numpy as np
+from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
 from sherpa.astro import ui
 
 from tempfile import NamedTemporaryFile
 
+@pytest.fixture
+def setup(make_data_path):
```
DougBurke, line 28: `sherpa/conftest.py` has `clean_astro_ui`. Can this be used here instead (it's only a simple function, but there is a little bit of subtlety that can rarely bite with the XSPEC settings)?

DougBurke: Although it doesn't look like the fixture is actually being used in any of the tests.

sherpa/astro/io/tests/test_io.py
```diff
@@ -56,3 +56,22 @@ def test_warnings_are_gone_pha(make_data_path):
...
+    Notes
+    -----
+    Test files wehre made with
+    dmtcalc sherpa-test-data/sherpatest/source.pi sherpa-test-data/sherpatest/source_edit.pi expression="stat_err=stat_err/exposure"
+    dmcopy "sherpa-test-data/sherpatest/source_edit.pi[cols channel,pi,RATE=count_rate,stat_err]" sherpa-test-data/sherpatest/source_rate.pi clobber=yes
+
+    '''
+    ui.load_pha("phacounts", make_data_path("source.pi"), use_errors=True)
+    ui.load_pha("pharate", make_data_path("source_rate.pi"), use_errors=True)
+    assert np.allclose(ui.get_data("phacounts").counts,
```
DougBurke, line 74: Can we change the `np.allclose(a, b)` to `b == pytest.approx(a)`, since I think the error message when there is a difference is better (although when there's a large array like here it may not help)?

DougBurke:
A usability question: So, is the situation now that users will have to know to add `use_errors=True` to the `load_pha` (or `load_data`) call when they have a RATE column, or are there cases when you would still want to use `use_errors=False` with a RATE file? If the answer is "must use use_errors otherwise the answers are wrong" then I'm not sure how we'd want to handle this ...

DougBurke:
@hamogu - can we add a check where the data is grouped and filtered, so we can compare the two files as you do with `test_scaling_staterror`? I would hope they are the same, but given the complexity of the group/filter code I think a test would be helpful.

sherpa/astro/io/tests/test_io.py
hamogu: I don't have a strong preference either way, so I'll change it.

hamogu:
Re: Usability question: For a true count rate, everything was fine before with `use_errors=False`. If there is no staterror, the error to use for a fit is calculated on the spot from the counts number and since the counts number was scaled correctly (and still is) that does not change. However, that's not true in the general case, because RATE *could* be something else then counts/time, e.g. it could be a background-subtracted rate for some instruments. In that case, `use_error=False` will give you the wrong answer, but there is already a big warning when reading files with `use_error=False` and an error column exists.

hamogu:
Note that there is an issue here that is more complicated and that I'll punt to a future PR: The rate could be given in units other than cts/s, e.g. counts/ks. In that case, the scaling will be wrong, but at least ti will be the same for counts and err, and thus the relative error (which we care about here) is still the same as long as rate and STAT_ERR have the same units.

sherpa/astro/io/tests/test_io.py
hamogu: I converted the previous tests we had it, but I don't think it's needed, so I'll remove it. (In fact, I think it's good if the tests fail because they *should* not depend on the state. If they do, we need to find out why.)

hamogu:
Also note that I'm testing only the PHA I case. Maybe I should add a PHA II case, too? But that would need another data file It's a separate branch of code, so it should really be tested...

DougBurke:
Thanks for the updates. Looks good to me (and thanks for the PHA2 case).

As a note, I would often suggest faking up the dataset rather than adding a new one, but in this case I think it's important in this case that we are actually checking the data as read in by the IO library (we could create the file as a temporary file, but that seems excessive).

hamogu:
My original intention was to take the input file and then modify it to have RATE instead of COUNTS on the fly. I chose to add files instead because (a): I would have to write the temporary file with both the crates and pyfits backends and there is not "write_pha" that gets the keywords right by itself and (b) I want to avoid circular testing where I use the pyfits backend to write a temporary file and then read it in again without hand-checking that it actually has all the columns that it should have with the values it should have.

hamogu:
These tests require the new files in https://github.com/sherpa/sherpa-test-data/pull/23. So, they will fail in travis until that is merged. If you try this out locally, you need to check out the changes in my branch of sherpa-test-data.

hamogu:
and i guess that the sherpa-test-data PR will only get merged once we have agreement that this PR is doing the right ting, so unfortunately, you'll have to review it although the CI does not pass at the moment.

wmclaugh:
@hamogu - What is the status of [sherpa/sherpa-test-data#23 ](https://github.com/sherpa/sherpa-test-data/pull/23)? Are these ready to merged together?

hamogu:
Yes, since @dtnguyen2 approved this PR, they are ready to be merged. Just out of an abundance of caution, I suggest to merge sherpa/sherpa-test-data#23 first, then I can restart the CI here and we make sure it passes with the merged test data. I tested that locally, but you never know...

hamogu:
@wmclaugh I added a commit that accounts for the updated submodule and now all tests pass, so this can be merged.

PR #845, opened 2020-07-04
xspec table models: add out-of-bound check to avoid segfault

hamogu:
This PR superceeds #743. See there for discussion.

Once the CI passes here, I will transfer the meta-information from #743 (requested reviewers, milestone etc.) to this PR and close #743.

## Summary
Passing an parameter that is out of bounds to an XSPEC table model can segfault the Sherpa session. This was originally discovered with parameters linked to other models, but can happen in other ways, too. This PR adds an explicit out-of-bounds check.


DougBurke:
@hamogu - can you add information to the initial entry in this PR to explain what it does (it makes generating the changelog a lot easier)?

PR #759, opened 2020-03-16
revert PR #444 (caching)

dtnguyen2:
# Summary

Revert the ARF cache added in #444, as well as some of the related code changes, as they caused problems (e.g. #746)

# Note 

This PR reverts part of PR #444

PR #847, opened 2020-07-06
Cache code: make the parameter optional and fix some missing cases

DougBurke:
The cache argument to startup has been made optional, to try and
avoid issues like #728, #733, #759

All startup methods have now been adjusted to take the cache
argument (which should also fix the issue).

The default is to not cache the results (as a safety precaution).

### DIFF
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -148,7 +148,7 @@ class RMFModel(CompositeModel, ArithmeticModel):
         # Used to rebin against finer or coarser energy grids
         self.rmfargs = ()
 
-    def startup(self, cache):
+    def startup(self, cache=False):
         self.model.startup(cache)
         CompositeModel.startup(self, cache)
 
@@ -197,7 +197,7 @@ class ARFModel(CompositeModel, ArithmeticModel):
         # Used to rebin against finer or coarser energy grids
         self.arfargs = ()
 
-    def startup(self, cache):
+    def startup(self, cache=False):
         self.model.startup(cache)
         CompositeModel.startup(self, cache)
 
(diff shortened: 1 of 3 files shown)


DougBurke:
I am doing some R&D on the cache code and noticed that even after recent fixes we still haven't caught all cases of passing the cache to the startup method. To fix this

a) I've made the argument optional (defaulting to False) so this isn't a problem anymore
b) I've made sure we pass the argument to all cases of startup (in these modules anyway) which should catch all cases we have so far

dtnguyen2:
In re-reading @DougBurke intention for this PR, if his intention

> The default is to not cache the results (as a safety precaution).

is not to use caching as the default, then this PR as it is written (never mind about my approval earlier, oops!) does not do that because ```_use_caching``` is initialized to ```True```, see code below.  Setting the default cache arg to ```False``` in the ```startup``` method will only set the cache of the relevant function to false if the calling function does not call with the optional ```cache``` argument, I know it is confusing  but will try to clarify if necessary :(

```
(sherpa) [dtn@devel12 sherpa]$ grep use_caching sherpa/models/model.py
        use_caching = cls._use_caching
        if use_caching:
        if use_caching:
        self._use_caching = True  # FIXME: reduce number of variables?
        if '_use_caching' not in state:
            self.__dict__['_use_caching'] = True
        self._use_caching = cache
                self._use_caching = cache
        self._use_caching = False
```

DougBurke:
I have mentioned in our previous review of the cache code that I don't understand the fall-through logic of the fit cache argument, the startup cache argument, and the various arguments: _use_caching and cache (e.g. there's one part that has int(cache)>0. It would be great if we could clean this up a bit!

PR #848, opened 2020-07-07
Use TRAVIS_NUMCORES for Travis jobs (number of npus)

DougBurke:
It looks like the information I found is no-longer valid, since Travis seems to think we only have 2 cores, so there's no need for the following (which I ended up not getting quite right anyway).

# Details

It is possible that out multi-processing code on Travis is picking
the wrong number of CPUS: the Python multiprocessing module reads
the number from a value which is not guaranteed to be correct on
a VM, as described at
https://github.com/travis-ci/travis-ci/issues/4696

This is fixed by setting up a configuration file, for the Travis
tests, which uses the TRAVIS_NUMCORES environment variable to
fix the number of CPUs.


DougBurke:
And I still didn't manage to make the code do what I wanted...

PR #851, opened 2020-07-08
Updates to the Read-The-Docs build environment

DougBurke:
# Summary 

Update the Read-The-Docs configuration to the latest version (2), and switch to a cleaner build (using conda) for the documentation. The minimum Sphinx requirement is now 1.8 (updated from 1.3).

Several minor document fixes have been added, as well as updates to the minimum-required Sphinx version.

# Notes

The build now uses python 3.7 (previously it was using 3.5 which is not sustaiable). The current Sphinx version being used is version 3.1.2 (i.e. much later than the minimum version).

### DIFF
--- a/docs/ciao.rst
+++ b/docs/ciao.rst
@@ -14,7 +14,7 @@ from the
 with the following modifications:
 
 * the I/O backend uses the CIAO library :term:`Crates` rather than
-  :term:`astropy`;
+  :term:`Astropy`;
 
 * a set of customized IPython routines are provided as part of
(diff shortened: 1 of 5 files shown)


DougBurke:
Note that for this PR I have turned on the read-the-docs support for building PRs, as it is rather useful to check this actually works on RTD...

I will turnoff the PR support once this PR has been reviewed.

DougBurke:
For reference I've turned off the build-PRs-on-read-the-docs. It's nice, but I don't think we benefit that greatly (for most PRs) and don't want to waste energy/resources.

PR #856, opened 2020-07-16
Support setting the id value in load_pha with pha2 files (fix #666)

### LINKED ISSUE #666: Cannot specify dataset id when loading pha2 files
I cannot find a way to assign ids when reading a pha type II file. This is relevant for grating data, where CIAO by default builds PHA II files, with spectra for each order, and for the HETGS for HEG and MEG. 

Is there a way of reading in a PHA II file without overwriting a previously loaded one? I've tried to pass in a number like `load_data(100, 'pha2.fits')` hoping I would then get ids 100, 101, 102, ... but that did not work. I also tried passing in a list of names, but I always get:
`Multiple data sets have been input: 1-12` and  Ioose the previously loaded dataset.

I know I can split a PHA II file into PHA I files (that's what I'm doing now), but it sure would be useful if there was a documented way to tell Sherpa which dataset ids to load the data into.

DougBurke:
The `load_pha` function will now use the id argument when given a PHA2 file, where previously it was ignored and the range 1 to nfiles was used for the identifiers of the datasets. Now the range is set to id to id + nfiles - 1 (when id is an integer), or "{}1".format(id) to "{}{}".format(id, nfiles) when id is a string.

# Details

If infile contains 4 files then the datasets loaded are

load_pha(infile) = [1, 2, 3, 4]
load_pha(5, infile) = [5, 6, 7, 8]
load_pha('x', infile) = ['x1', 'x2', 'x3', 'x4']

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1739,12 +1739,20 @@ class Session(sherpa.ui.utils.Session):
         multiple data sets can be loaded with this command, as
         described in the `sherpa.astro.datastack` module.
 
+        .. versionchanged:: 4.12.2
+           The id argument is now used to define the first identifier
+           when loading in a PHA2 file (previously they always used
+           the range 1 to number of files).
+
         Parameters
         ----------
         id : int or str, optional
-           The identifier for the data set to use. If not given then
-           the default identifier is used, as returned by
-           `get_default_id`.
+           The identifier for the data set to use. For PHA2 files,
+           that is those that contain multiple datasets, the id value
+           indicates the first dataset: if it is an integer then the
+           numbering starts at id, and if a string then a suffix of 1
+           to n is added.  If not given then the default identifier is
+           used, as returned by `get_default_id`.
         arg
            Identify the data to read: a file name, or a data structure
            representing the data to use, as used by the I/O backend in
@@ -1804,12 +1812,33 @@ class Session(sherpa.ui.utils.Session):
         >>> load_pha('bg', 'x2.fits')
 
         If a type II PHA data set is loaded, then multiple data sets
-        will be created, one for each order.
+        will be created, one for each order. The default behavior is
+        to use the dataset identifiers 1 to the number of files.
 
         >>> clean()
         >>> load_pha('src.pha2')
         >>> list_data_ids()
         [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+
+        If given an identifier as the first argument then this is used
+        to start the numbering scheme for PHA2 files. If id is an
+        integer then the numbers go from id:
+
+        >>> clean()
+        >>> load_pha(20, 'src.pha2')
+        >>> list_data_ids()
+        [20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31]
+
+        If the id is a string then the identtifier is formed by adding
+        the number of the dataset (starting at 1) to the end of
+        id. Note that the `list_data_ids` routine does not guarantee
+        an ordering to the output (as shown below):
+
+        >>> clean()
+        >>> load_pha('x', 'src.pha2')
+        >>> list_data_ids()
+        ['x1', 'x10', 'x11', 'x12', 'x2', 'x3', 'x4', 'x5', 'x6',
+         'x7', 'x8', 'x9']
 
         Create the data set from the data read in by Crates:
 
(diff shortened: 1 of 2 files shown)


DougBurke:
@hamogu does this do what you want? If so then I'll pass it on to Dan for review but I wanted to check first.

sherpa/astro/ui/utils.py
```diff
@@ -1739,12 +1739,20 @@ def load_pha(self, id, arg=None, use_errors=False):
         multiple data sets can be loaded with this command, as
         described in the `sherpa.astro.datastack` module.
 
+        .. versionchanged:: 4.12.2
+           The id argument is now used to define the first identifier
+           when loading in a PHA2 file (previously they always used
+           the range 1 to nummber of files).
```
hamogu, line 1745: typo: number

sherpa/astro/ui/utils.py
```diff
@@ -1841,14 +1870,35 @@ def load_pha(self, id, arg=None, use_errors=False):
...
-                self.set_data(id + 1, pha)
-            if num > 1:
-                info("Multiple data sets have been input: 1-%s" % num)
-        else:
+        if not numpy.iterable(phasets):
             self.set_data(id, phasets)
+            return
+
+        # One issue with the following is that if there's
+        # only one file and id is a string then the
```
hamogu, line 1878: You mean "one dataset in the file" and not "one file", right?

sherpa/astro/ui/utils.py
```diff
@@ -1841,14 +1870,35 @@ def load_pha(self, id, arg=None, use_errors=False):
...
+        # output will be "foo1" rather than "foo" (when
+        # id="foo").  DJB thinks we can live with this.
```
hamogu, line 1880: I think we can call that a feature and not a bug!

DougBurke: I just wanted to note this, as it is subtly different to `load_pha('goo', singlefile)` when singlefile is not a PHA2 container, since this just creates a dataset called `goo`.

sherpa/astro/ui/utils.py
DougBurke: I'm updating the comment to be clearer.

sherpa/astro/ui/utils.py
hamogu: It becomes relevant if the want to make a further enhancement (not for this PR), that allows selecting the PHA rows to read. For HETGS we typically extract 12 rows (HEG and MEG from -3 to +3), but if I already know I'm looking at a weak and relatively soft source, I only need MEG-1 and MEG+1. However, for now, I don't think that's a feature that's important to have because (a) the memory demand is low and (b) people can manipulate the PHAII file outsode of sherpa (e.g. tell tg_extract which orders they want). This might become more important in the future. Arcus will have 4 channels * 20 orders = 80 spectra or peopel could use PHAII file to store, e.g. all ACIS spectra from 154 identified sources in the FOC of an observation. However, I would argue that that's their problem. Sherpa just reads "one file" and if you want less, make a smaller file.

DougBurke: I had briefly thought about being able to specify the ids in some way - e.g. you could name them "megp1", "megm1", ... - but that is a "big" change to how we think about ids, whereas this PR is pretty tiny ;-)

hamogu: That's what I do by hand, i.e. using `copy_data`, but that's for a future PR (which is not hard to build on top of this one.) So far, Sherpa is pretty generic, but I can imagine something like:
```
naming_functions = {'CHANDRA': chandra_namer,
                                  ...}
def chandra_namer(pha2):
    return ['{}_{}'.format(['zero','heg', 'meg','leg'][row['TGPART']],row['TGM']) for row in pha2]
```
and in the load function:
```
try:
    ids = nameing_functions[pha2.meta['TELESCOP']](pha2)
except:
   ids = ... what's in this PR because that works for all instruments ...
```
Not sure if that has to be backend dependent or if there is a way to iterate over rows in a backend-independent way.

anetasie:
@DougBurke an @hamogu does this support `datastacks`? I guess that's the question for @hamogu who can try to use it with `datastacks`.

DougBurke:
datastacks is a whole nother bach of complexity; from what @hamogu has
already reported it looks like there are a lot of "issues" (or design
decisions) with datastacks. I wasn't going to delay this fix to resolve
them as they are separate

hamogu:
I agree with @DougBurke. Datastack are a separate issue - but this work will make is much easier to work with them. After all, datastacks don't have to be created from file, in-memory datasets can be added to a datastack.

PR #858, opened 2020-07-20
Minor documentation improvements to ReadTheDocs

DougBurke:
# Summary

Minor updates to the convolution, regrid, and model evaluation sections of the ReadTheDocs documentation.

# Details

Add a 1D PSF convolution example (to go with the existing 2D PSF example) and add example code showing how the regrid method of models works (although there are existing issues that need to be addressed before we can really improve this section).

The PHA model-evaluation section has been rewritten, as there were three different versions which contained incomplete, and occasionally incorrect, advice. We now have a coherent single example.


docs/evaluation/convolution.rst
```diff
@@ -25,32 +25,175 @@ or the redistribution of the counts as a function of energy
 as modelled by the :term:`RMF` when analyzing astronomical
 X-ray spectra.
 
+.. note::
+
+   In the following documentation we are going to use the term
+   "convolution" to refer to any model that requires the data
+   from another model, such as a model that reprocesses emission
+   into another pass band or that renormalises the data, even if
```
hamogu, line 33: pass band or bandpass? You are the native speaker...

docs/evaluation/convolution.rst
```diff
@@ -25,32 +25,175 @@ or the redistribution of the counts as a function of energy
...
+
+ 2. apply the instance to the model components
+
+The convolution kernel is not an instance of
+:py:class:`sherpa.models.model.ArithmeticModel`, and so
+can not be evaluated by passing it the data grid. Instead,
+the kernel is used to wrap another model expression,
+normally done by passing the model to be wrapped as the
+argument to the model, and the result of this is then
+a model that can be evaluated as normal.
```
hamogu, line 53: as normal -> on the data grid (to match the wording from above)

docs/evaluation/convolution.rst
```diff
@@ -25,32 +25,175 @@ or the redistribution of the counts as a function of energy
...
+    psfmodel(point)
+       Param        Type          Value          Min          Max      Units
+       -----        ----          -----          ---          ---      -----
+       point.xlow   thawed          9.5 -3.40282e+38  3.40282e+38
+       point.xhi    thawed         10.5 -3.40282e+38  3.40282e+38
+       point.ampl   thawed            1           -1            1
+
+.. _convolution-1d-fold:
+
+The model can be evaluated bith before and after convolution, by
```
hamogu, line 118: typo "bith"

docs/evaluation/convolution.rst
```diff
@@ -25,32 +25,175 @@ or the redistribution of the counts as a function of energy
...
+evaluation. This method pre-calculates terms needed for the
+convolution (which is done using a fourier transform), and so needs
+the grid over which it is to be applied. This is done by passing in a
+:py:class:`~sherpa.data.Data` instance (in this case the Y data in the
+:py:class:`~sherpa.data.Data1D` instance is not used so is set to
+zero):
+
+    >>> x = np.arange(6, 15)
+    >>> blank = Data1D('blank', x, np.zeros(x.size))
+    >>> psf1.fold(blank)
```
hamogu, line 131: This strikes me as complex. Why not do that when the model is called wit the x-values of whatever data comes in. The result can then be cached so that the fold is called only once. We have caching already in Sherpa, so that can be re-used, alternatively a simple
```
self.folded_x = None
def call (...):
    if self.folded_x != x_value_passed_in:
       self.fold(...)
```
will do. I realize that that's not part of this PR, but I wonder if we can fix that instead of spending the work documenting a complicated procedure.

docs/evaluation/convolution.rst
```diff
@@ -25,32 +25,175 @@ or the redistribution of the counts as a function of energy
...
+
+   Evaluating the kernel - such as `psf` in the example
+   below - with a grid will raise an error like::
+
+       ModelErr: attempted to create ArithmeticFunctionModel from non-callable object of type ndarray
+
+.. _convolution-psf1d:
+
+One-dimensional convolution with a PSF
+======================================
```
hamogu, line 65: What does "one-dimensional convolution with a PSF" even mean? A PSF is usually 2D. I'm trying to come up with a more natural example and if I have an idea I'll comment again. At least, I think this would profit from 1-2 sentences that give an example what this is good for. I think "spectrum", e.g. in the optical: "One example for this is observing a spectrum. The model for the spectrum might have very sharp features, but instrumental effects will broaden those features. Mathematically, we can be expressed as the convolution of the physical spectrum with a broadening function. Note that in X-ray the RMF formalism is used in sherpa, see ...

docs/evaluation/regrid.rst
```diff
@@ -2,21 +2,153 @@
...
 
+.. note::
+
+   The current implementation only really supports the case of
+   2D PSFs (using a higher resolution image for the convolution
+   and then rebinning to match the data). There are ongoing
+   efforts to extend the use of this, in particular issues
+   `829 <https://github.com/sherpa/sherpa/issues/829>`_,
+   `762 <https://github.com/sherpa/sherpa/issues/762>`_,
+   and `300 <https://github.com/sherpa/sherpa/issues/300>`_.
```
hamogu, line 20: We are expecting to resolve those before the next release. Then, we'll forget to update the docs. Maybe we don't have to list those PRs here? Users using the master branch might expect some missing docs? Or is master served on RTD be default (instead of e.g. stable) so that we expect users to hit this text any time?

docs/evaluation/regrid.rst
```diff
@@ -2,21 +2,153 @@
...
+method:
+
+    >>> from sherpa.models.basic import Gauss1D
+    >>> mdl = Gauss1D('orig')
+    >>> rmdl = mdl.regrid(xgrid)
+
+The "new" model is really just a wrapper around the original model,
+which means that it shares the parameter values and settings. This
+is shown below where the ``pos`` and ``fwhm`` parameters of the
+``rmdl`` are labbeled by the label given to ``mdl`` (namely
```
hamogu, line 59: labbeled?

docs/evaluation/convolution.rst
DougBurke: I am not a fan of the current `fold` interface (I think it should either be more explicit or handled differently). However, I have no time or energy to work on this, and I do not believe we have the resources to work on this before the end of the year. So I really would like to get some documentation out there to users.

hamogu: Makes sense.

docs/evaluation/convolution.rst
DougBurke: Sherpa has used the term PSF with 1D kernels for longer than you've been using CIAO ;-) You can think of PSF as LSF in this case. I worry that being too explicit about an example - e.g. "Include the grating response in a model" - hides the meaning for people who aren't bothered about spectra.

docs/evaluation/regrid.rst
DougBurke: I am not convinced they will be fixed.

docs/evaluation/convolution.rst
anetasie: Just to add an example -  1D radial profile analysis can use a 1D PSF generated as a radial profile on the simulated PSF.

docs/evaluation/regrid.rst
DougBurke: I've now taken out this note as at least one of these issues has been addressed.

DougBurke:
@hamogu - I believe I've resolved all your issues other than the question of 1D convolution.

There has been an update to the evaluation/examples section to rewrite the PHA section since it was confused and occasionally wrong.

anetasie (approved):
I did not try to run any of the examples, but it is good to include them in RTD.

hamogu (approved):

docs/evaluation/regrid.rst
```diff
@@ -18,5 +14,137 @@ from the
...
+given two arrays, for the low and high edges of each bin.
+
+::
+
+   >>> imdl = mdl.regrid(xgrid[:-1], xgrid[1:])
+   >>> y1 = mdl(x[:-1], x[1:])
+   >>> y2 = imdl(x[:-1], x[1:])
+   >>> np.testing.assert_allclose(y1, y2, rtol=1e-2)
+
+The fact that the models are integrated across the bins mean that the
```
wmclaugh, line 96: I think it should be 'means' rather than 'mean'

docs/evaluation/examples.rst
```diff
@@ -134,473 +134,386 @@ and
...
+   >>> from sherpa.astro.instrument import RSPModelPHA
+   >>> full = RSPModelPHA(arf, rmf, pha, pha.exposure * mdl)
+
+   Note that the exposure time is not automatically included for you as it
+   is with ``Response1D``.
+
+If we evaluate this model we get a surprise! The grid arguments
+are ignored (as long as something is sent in), and instead the
+model is evaluated on the channel group (hence the evaluated
+model as 1024 bins in this exampe)::
```
wmclaugh, line 390: exampe -> example

docs/evaluation/examples.rst
```diff
@@ -134,473 +134,386 @@ and
...
+:py:class:`~sherpa.astro.plot.ModelHistogram` plotting
+class to display the model data without needing to
+convert anything::
+
+   >>> from sherpa.astro.plot import ModelHistogram
+   >>> mplot = ModelHistogram()
+   >>> mplot.prepare(pha, full)
+   >>> mplot.plot()
+
+The difference to the previos plot is that this one
```
wmclaugh, line 473: previos -> previous

docs/evaluation/regrid.rst
DougBurke: Done

docs/evaluation/examples.rst
DougBurke: Done

docs/evaluation/examples.rst
DougBurke: Done

DougBurke:
Ta for the review @wmclaugh

PR #859, opened 2020-07-20
Clean up of the XSPEC interface code

DougBurke:
# Summary

Internal changes to the XSPEC interface code, which reduces the amount of similar (sometimes identical) code. There is no change to the behavior of the XSPEC models.

# Details

Noted by @dtnguyen2 in a review of #772

 - [X] Switch to using std::transform in XSPEC Fortran-style models
 - [X] Reduce duplication (but there is more, eg #860)

I'm not likely to work on this anymore (modulo PR review) so we can either review this or through it.

### DIFF
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -266,6 +266,125 @@ static bool create_grid(const sherpa::Array<CType, ArrayType> &xlo,
 } /* create_grid */
 
 
+// Similar to create_grid except that it does not support grids that are
+// not contiguous (an error is raised).
+//
+template <typename CType, int ArrayType>
+static bool create_contiguous_grid(const sherpa::Array<CType, ArrayType> &xlo,
+				   const sherpa::Array<CType, ArrayType> &xhi,
+				   std::vector<CType> &ear) {
+
+  int nelem = int( xlo.get_size() );
+  if ( nelem < 2 ) {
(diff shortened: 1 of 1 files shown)


DougBurke:
I've rebased this because of changes to the XSPEC interface (for XSPEC 12.11.0 support).

DougBurke:
It appears that this code barfs on the 98 C++ standard, which we still want to support easily (i.e. without forcing explicit calls to a newer standard).

dtnguyen2:
```
(base) [dtn@devel12 dtn]$ cat tmp.cc
#include <iostream>

int main(){
#if __cplusplus==201402L
  std::cout << "C++14" << std::endl;
#elif __cplusplus==201103L
  std::cout << "C++11" << std::endl;
#else
  std::cout << "C++" << std::endl;
#endif
  
  return 0;
}

(base) [dtn@devel12 dtn]$ g++ -std=c++98 --pedantic tmp.cc; a.out
C++
(base) [dtn@devel12 dtn]$ g++ -std=c++11 --pedantic tmp.cc; a.out
C++11

```

wmclaugh:
@DougBurke - I agree with Dan's suggestion of a conditional compile here to support the c++ 98 standard as well as the c++ 11 standard.  The 11 standard provides improved performance and compatibility with other libraries built against the standard. Whether we would want to eventually drop the 98 standard is a separate issue as is the infrastructure changes to support specifying/propagating (commandline) compiler flag options.

PR #863, opened 2020-07-22
Support an "ideal" RMF where MATRIX is a scalar (pyfits backend, fix #862)

### LINKED ISSUE #862: loading RMF fails in backend pyfits
Because of issue #817, I am using io_pkg: pyfits.

However, loading [this RMF](https://github.com/JohannesBuchner/BXA/blob/master/bxa/sherpa/background/mos_dia.rmf) file fails now, with the error:

```
$ ipython3
Python 3.5.4 (default, Aug 22 2019, 13:24:24) 
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

   ...:                                                                                                                                                                 

In [2]: load_rmf('mos_dia.rmf')                                                                          
IndexError                                Traceback (most recent call last)
<ipython-input-2-8682bea2a0e4> in <module>
----> 1 load_rmf('mos_dia.rmf')

<string> in load_rmf(id, arg, resp_id, bkg_id)

~/bin/ciao-4.12/lib/python3.5/site-packages/sherpa-4.12.0-py3.5-linux-x86_64.egg/sherpa/astro/ui/utils.py in load_rmf(self, id, arg, resp_id, bkg_id)
   5929         if arg is None:
   5930             id, arg = arg, id
-> 5931         self.set_rmf(id, self.unpack_rmf(arg), resp_id, bkg_id)
   5932 
   5933     def get_bkg_rmf(self, id=None):

~/bin/ciao-4.12/lib/python3.5/site-packages/sherpa-4.12.0-py3.5-linux-x86_64.egg/sherpa/astro/ui/utils.py in unpack_rmf(self, arg)
   5846 
   5847         """
-> 5848         return sherpa.astro.instrument.RMF1D(sherpa.astro.io.read_rmf(arg))
   5849 
   5850     # DOC-TODO: add an example of a grating/multiple response

~/bin/ciao-4.12/lib/python3.5/site-packages/sherpa-4.12.0-py3.5-linux-x86_64.egg/sherpa/astro/io/__init__.py in read_rmf(arg)
    387 
    388     """
--> 389     data, filename = backend.get_rmf_data(arg)
    390 
    391     # It is unlikely that the backend will set this, but allow

~/bin/ciao-4.12/lib/python3.5/site-packages/sherpa-4.12.0-py3.5-linux-x86_64.egg/sherpa/astro/io/pyfits_backend.py in get_rmf_data(arg, make_copy)
    745                 # a float.
    746                 end = start + numpy.int(nc)
--> 747                 rowdata.append(mrow[start:end])
    748                 start = end
    749 

IndexError: invalid index to scalar variable.
```

DougBurke:
# Summary

The AstroPy (pyfits) backend could not handle the case of an ideal RMF when the MATRIX column was specified as a scalar and not a 1-element array.

# Details

This is to support issue #862 where a MATRIX column is a scalar, not an array. This is aready supported by Crates, so the Astropy backend needs updating. The change is small (if you can't access the first element of an array then treat it as a scalar), and is an extension to the changes in PRs #358 and #412 which dealt with Swift and ROSAT responses.

A test is added, but rather than adding a file to the Sherpa data corpus, we instead just create the file when needed. It also means that we can easily test the values read in as the data can be made to be easily tested (i.e. a small number of bins).

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -738,13 +738,24 @@ def get_rmf_data(arg, make_copy=False):
             # Need a RMF which ng>1 to test this with.
             if ng == 1:
                 ncs = [ncs]
+
             start = 0
             for nc in ncs:
                 # n_chan can be an unsigned integer. Adding a Python
                 # integer to a NumPy unsigned integer appears to return
                 # a float.
                 end = start + numpy.int(nc)
-                rowdata.append(mrow[start:end])
+
+                # "perfect" RMFs may have mrow as a scalar
+                try:
+                    rdata = mrow[start:end]
+                except IndexError:
+                    if start != 0 or end != 1:
+                        raise IOErr('bad', 'format', 'MATRIX column formatting')
+
+                    rdata = [mrow]
+
+                rowdata.append(rdata)
                 start = end
 
         data['matrix'] = numpy.concatenate(rowdata)
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -16,11 +16,17 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-import pytest
+
+from tempfile import NamedTemporaryFile
+import struct
 
 import numpy as np
+
+import pytest
+
 from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
 from sherpa.astro import ui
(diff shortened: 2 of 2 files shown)


DougBurke:
A test case is needed, which is why this is still a WIP.

DougBurke:
Gah. Another PR that needs rebasing.

DougBurke:
In rebasing this I found that two of the "extra" commits from this PR were no-longer needed (as another PR had converted test_io.py to pytest from SherpaTestCase) so I have dropped them. Which makes this more focussed: a commit to add the change, and one to add the test.

PR #865, opened 2020-07-27
Minor documentation fixes

DougBurke:
# Summary

Several documentation fixes: XSPEC parameter names, avoiding confusion over links on references (Sphinx pages), and adding some basic documentation to the sherpa.astro.background module.

# Details

This is a documentation-only PR.

I noticed these when working on an issue. They are all minor changes, but the XSPEC changes do fix a documentation bug (and are the only changes users of the sherpa.astro.ui layer would note).

### DIFF
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -31,7 +31,7 @@ class BackgroundSumModel(CompositeModel, ArithmeticModel):
     """Combine multiple background datasets.
 
     Define the model expression to be applied to the source
-    region accounting for the backgroud models.
+    region accounting for the background models.
 
     Parameters
     ----------
@@ -40,7 +40,7 @@ class BackgroundSumModel(CompositeModel, ArithmeticModel):
     bkgmodels : dict
         The background components, where the key is the identifier
         in this dataset and the value is the model. This dictionary
-        is not empty.
+        cannot be empty.
 
     """
 
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -1,5 +1,5 @@
-# 
-#  Copyright (C) 2007  Smithsonian Astrophysical Observatory
+#
+#  Copyright (C) 2007, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -17,6 +17,10 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+"""Support multiple background components for a PHA dataset.
+
+"""
+
 from sherpa.models.model import CompositeModel, ArithmeticModel
 from sherpa.utils.err import DataErr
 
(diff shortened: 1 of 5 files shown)


DougBurke:
I would have liked to provide more context to sherpa.astro.background, but I'm having to work out what it's meant to do so I have decided to go with the minimum information I felt useful. Hopefully some forthcoming work will improve our understanding of the code.

DougBurke:
Also, the fix to the reference links is not meant to fix all cases, just a small-number of them, It's only really an issue for uses who build the sphinx documentation.

sherpa/astro/background.py
```diff
@@ -17,13 +17,32 @@
...
 from sherpa.utils.err import DataErr
 
 __all__ = ('BackgroundSumModel',)
 
 
 class BackgroundSumModel(CompositeModel, ArithmeticModel):
+    """Combine multiple background datasets.
+
+    Define the model expression to be applied to the source
+    region accounting for the backgroud models.
```
hamogu, line 34: I don't understand. Does the model define the expression?

sherpa/astro/background.py
```diff
@@ -17,13 +17,32 @@
...
+
+    Parameters
+    ----------
+    srcdata : sherpa.astro.data.DataPHA instance
+        The source dataset.
+    bkgmodels : dict
+        The background components, where the key is the identifier
+        in this dataset and the value is the model. This dictionary
+        is not empty.
```
hamogu, line 43: is not -> cannot be

sherpa/astro/background.py
wmclaugh, line 34: line 34- change 'backgroud' to 'background'.   (Missing 'n')

DougBurke: It's meant to be "model expression". It could have been "model", but I feel that is a bit anemic. However, I am not the best person to judge my words. 

There's no documentation on how this is meant to be used, so I'm relying on its's use in the following (sherpa/astro/ui/utils.py), where we see we need the models to be sent in as a dict (indexed by there "bkg_id" value).

```
    def _add_convolution_models(self, id, data, model, is_source):
        model = \
            sherpa.ui.utils.Session._add_convolution_models(self, id, data,
                                                            model, is_source)
        id = self._fix_id(id)
        if (isinstance(data, sherpa.astro.data.DataPHA) and is_source):
            if not data.subtracted:
                bkg_srcs = self._background_sources.get(self._fix_id(id), {})
                if len(bkg_srcs.keys()) != 0:
                    model = (model +
                             sherpa.astro.background.BackgroundSumModel
                             (data, bkg_srcs))

            pileup_model = self._pileup_models.get(self._fix_id(id))
            if pileup_model is not None:
                resp = \
                    sherpa.astro.instrument.PileupResponse1D(
                        data, pileup_model)
                model = resp(model)

            elif len(data._responses) > 1:
                resp = sherpa.astro.instrument.MultipleResponse1D(data)
                model = resp(model)

            else:
                resp = sherpa.astro.instrument.Response1D(data)
                model = resp(model)

        return model
```

DougBurke: I've not got the energy to change this, so I'm going to punt on it and hope it gets improved whenever someone gets to write proper documentation for this class.

DougBurke:
Thanks for the updates @hamogu and @wmclaugh - I've changed the ones I want to change ;-)

PR #866, opened 2020-07-27
Add parameter-clipping strategy to routines that generate samples (fix #846)

### LINKED ISSUE #846: parameter sampling, soft and hard limits, and the sample*flux functions
So, `sample_energy_flux` creates `num` realizations of the free-parameter distribution (e.g. if the model has nh, gamma, and ampl free then you get a num by 3 array of numbers) which is used to calculate the flux distribution. Since the numbers are generated using a normal distribution (either with correlated parameters or not) then you can end up with parameter values outside the hard-limit range (often seen with the nh parameter), so the code goes through and makes sure the parameters are clipped to the hard limits (that is, values outside the hard-limit range get replaced with values at the hard limit).

The `sample_flux` function currently post-processes this array to remove any flux values with parameters outside the soft-limit range (internally `sample_flux` calls `sample_energy_flux` to do most of the work). Unfortunately there are issues with how this is done (e.g. #751).

I suggest that add an option to `sample_energy_flux` that lets you chose whether to clip the parameters by the soft or hard limit (this would leave in those rows where values are outside the limits but replaces the problem value with the limit). This would then avoid the problem in #751 (and avoids `sample_flux` doing excess work).

One question is that I have left the bins in (since they provide some information related to upper limits, e.g. #457). However, maybe you would want to remove the clipped bins. One thought I have is to add an extra column indicating whether the row was "okay" or "clipped"  (e.g. 1 or 0) so users could filter out the clipped values if they wanted.

DougBurke:
# Summary

The addition of the clip parameter lets users control how parameter values are clipped before use in  sample_energy_flux, sample_photon_flux, plot_energy_flux, plot_photon_flux, get_energy_flux_hist, and get_photon_flux_hist.

# Details

At present only sherpa.astro.flux.sample_flux supports the `clip` parameter. The sherpa.astro.flux.calc_sample_flux is waiting on #754 before it is worth updating (this means that we can not add a clip parameter to sample_flux yet).

### DIFF
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -136,7 +136,8 @@ def calc_flux(data, src, samples, method=calc_energy_flux,
     return numpy.asarray(fluxes)
 
 
-def _sample_flux_get_samples_with_scales(fit, src, correlated, scales, num):
+def _sample_flux_get_samples_with_scales(fit, src, correlated, scales,
+                                         num, clip='hard'):
     """Return the parameter samples given the parameter scales.
 
     Parameters
@@ -162,13 +163,21 @@ def _sample_flux_get_samples_with_scales(fit, src, correlated, scales, num):
         mfree by mfree.
     num : int
         Tne number of samples to return. This must be 1 or greater.
+    clip : {'hard', 'soft', 'none'}, optional
+        What clipping strategy should be applied to the sampled
+        parameters. The default ('hard') is to fix values at their
+        hard limits if they exceed them. A value of 'soft' uses the
+        soft limits instead, and 'none' applies no clipping. The last
+        column in the returned arrays indicates if the row had any
+        clipped parameters (even when clip is set to 'none').
 
     Returns
     -------
-    samples : 2D NumPy array
+    samples, clipped : 2D NumPy array, 1D NumPy array
         The dimensions are num by mfree. The ordering of the parameter
         values in each row matches that of the free parameters in
-        fit.model.
+        fit.model.  The clipped array indicates whether a row had one
+        or more clipped paraneters.
 
     Raises
     ------
@@ -185,6 +194,7 @@ def _sample_flux_get_samples_with_scales(fit, src, correlated, scales, num):
     has not been tested for complex models, that is when fit.model
     is rmf(arf(source_model)) and src is a combination of components
     in source_model but not all the components of source_model.
+
     """
 
     npar = len(src.thawedpars)
@@ -247,10 +257,12 @@ def _sample_flux_get_samples_with_scales(fit, src, correlated, scales, num):
     else:
         sampler = NormalParameterSampleFromScaleVector()
 
-    return sampler.get_sample(fit, scales, num=num)
+    samples = sampler.get_sample(fit, scales, num=num)
+    clipped = sampler.clip(fit, samples, clip=clip)
+    return samples, clipped
 
 
-def _sample_flux_get_samples(fit, src, correlated, num):
+def _sample_flux_get_samples(fit, src, correlated, num, clip='hard'):
     """Return the parameter samples, using fit to define the scales.
 
     The covariance method is used to estimate the errors for the
@@ -272,13 +284,21 @@ def _sample_flux_get_samples(fit, src, correlated, num):
         Are the parameters assumed to be correlated or not?
     num : int
         Tne number of samples to return. This must be 1 or greater.
+    clip : {'hard', 'soft', 'none'}, optional
+        What clipping strategy should be applied to the sampled
+        parameters. The default ('hard') is to fix values at their
+        hard limits if they exceed them. A value of 'soft' uses the
+        soft limits instead, and 'none' applies no clipping. The last
+        column in the returned arrays indicates if the row had any
+        clipped parameters (even when clip is set to 'none').
 
     Returns
     -------
-    samples : 2D NumPy array
+    samples, clipped : 2D NumPy array, 1D NumPy array
         The dimensions are num by mfree. The ordering of the parameter
         values in each row matches that of the free parameters in
-        fit.model.
+        fit.model. The clipped array indicates whether a row
+        had one or more clipped paraneters.
 
     Notes
     -----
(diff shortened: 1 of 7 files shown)


DougBurke:
Note to self: was about to ask for SDS review, but I think we need #803 merged first (otherwise it's confusing to work out what is new in this PR)

DougBurke:
Due to changes to the master branch this needs rebasing (since it's based no #803 which has just been rebased), but I don't have the energy to fix this at the moment.

DougBurke:
I've rebased this and marked it ready for review (by SDS).

hamogu:
In the combined set there are > 3000 lines to look at and for most of the files, github does not even show me the diff because it's so big. I don't even know where to start. That's one problem with these huge changes and PRs that build on top of each other. Reviews just become difficult.

Can this wait until #803 is merged?

DougBurke:
For @hamogu - I'm labelling this as SDS review so we all have it on our radars. 

If you were bored and wanted to look at what this code is doing from a high level you could look at the new `clip` parameter to `sample_energy_flux` and see if the documentation makes sense.

DougBurke:
Rebased because of recent changes and to fix a less-than-ideal rebase.

DougBurke:
Rebased since #803 has been merged.

DougBurke:
So @anetasie and @hamogu - this should be easier to review now.

anetasie (approved):
It looks good. I agree with the default clip setting to the hard limits.

sherpa/astro/flux.py
```diff
@@ -162,13 +163,21 @@ def _sample_flux_get_samples_with_scales(fit, src, correlated, scales, num):
...
 
     Returns
     -------
-    samples : 2D NumPy array
+    samples, clipped : 2D NumPy array, 1D NumPy array
         The dimensions are num by mfree. The ordering of the parameter
         values in each row matches that of the free parameters in
-        fit.model.
+        fit.model.  The clipped array indicates whether a row had one
+        or more clipped paraneters.
```
wmclaugh, line 180: paraneters -> parameters

sherpa/astro/flux.py
```diff
@@ -272,13 +284,21 @@ def _sample_flux_get_samples(fit, src, correlated, num):
...
 
     Returns
     -------
-    samples : 2D NumPy array
+    samples, clipped : 2D NumPy array, 1D NumPy array
         The dimensions are num by mfree. The ordering of the parameter
         values in each row matches that of the free parameters in
-        fit.model.
+        fit.model. The clipped array indicates whether a row
+        had one or more clipped paraneters.
```
wmclaugh, line 301: paraneters -> parameters

sherpa/astro/flux.py
```diff
@@ -345,14 +367,19 @@ def decompose(mdl):
 
 def sample_flux(fit, data, src,
                 method=calc_energy_flux, correlated=False,
-                num=1, lo=None, hi=None, numcores=None, samples=None):
+                num=1, lo=None, hi=None, numcores=None, samples=None,
+                clip="hard"):
```
wmclaugh, line 371: For consistency, shouyld "hard" be 'hard'.  Looks like single quotes are used elsewhere...

sherpa/astro/flux.py
```diff
@@ -452,22 +488,13 @@ def sample_flux(fit, data, src,
...
-    #
-    hardmins = fit.model.thawedparhardmins
-    hardmaxs = fit.model.thawedparhardmaxes
-
-    for pvals, pmin, pmax in zip(samples.T, hardmins, hardmaxs):
-        # do the clipping in place
-        numpy.clip(pvals, pmin, pmax, out=pvals)
-
-    # When a subset of the full mdel is use we need to know how
+    # When a subset of the full model is use we need to know how
```
wmclaugh, line 497: use -> used

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -826,6 +826,15 @@ def test_sample_foo_flux_niter(multi, single, id, niter, correlated,
         flux = single(lo=0.5, hi=7, id=id)
         assert ans[i, 0] == flux
 
+    # check clipped is 0/1; ideally this is a boolean and not a
+    # float represention
```
wmclaugh, line 830: represention -> representation

sherpa/sim/sample.py
```diff
@@ -261,16 +463,71 @@ def get_sample(self, fit, myscales=None, num=1):
...
     def get_sample(self, fit, dof, num=1):
+        """Return the parameter samples.
+
+        Parameter
+        ---------
+        fit : sherpa.fit.Fit instance
+            This defines the thawed parameters that are used to generate
+            the samples, along with any possible error analysis.
+        dof : int
+            The degrees of freedon of the distribution.
```
wmclaugh, line 520: freedon -> freedom

sherpa/sim/sample.py
```diff
@@ -305,32 +588,160 @@ def _sample_stat(fit, samples, numcores=None, cache=True):
...
+
+        Parameter
+        ---------
+        fit : sherpa.fit.Fit instance
+            This defines the thawed parameters that are used to generate
+            the samples, along with any possible error analysis.
+        num : int, optional
+            The number of samples to return.
+        dof : int
+            The degrees of freedon of the distribution.
```
wmclaugh, line 730: freedon -> freedom

wmclaugh:
@DougBurke, this needs a rebase so that it can be merged.

sherpa/astro/flux.py
DougBurke: Ta - I've made the doc changes.

DougBurke:
@wmclaugh - let's hope for less fun with this rebase than my recent PRs!

DougBurke:
I've had to restart the macOS test as it timed out (after a failed test, but in code not related to this PR so it may have failed for the same reason the whole test failed)

DougBurke:
Yay it all passed.

PR #868, opened 2020-07-29
Add delete_pileup_model (fix #441), list_pileup_model_ids, list_psf_ids functions, fix list_models (fix #749)

### LINKED ISSUE #441: Add an easy way to remove a jdpileup model
From a CIAO helpdesk enquiry:

`sherpa.astro.ui.set_pileup_model` is used to add a pileup kernel to a model for a PHA dataset. At present there doesn't appear to be an easy way to remove the pileup kernel. A suggested work around is to copy the data and source model expression to a new "dataset", which doesn't have a pileup kernel - e.g. (untested)

    set_data(2, get_data())
    set_source(2, get_source())

I think we could have it so that calling `set_pileup_model` with `model=None` would remove the pileup kernel; at present it fails with 

```
sherpa-11> set_pileup_model(id=1, model=None)
ArgumentTypeErr: 'model' must be a model object or model expression string
```

which comes from some internal code: `sherpa.ui.utils._set_item`.

### LINKED ISSUE #749: sherpa.ui.list_models with a filter does not return a list but an iterator
```
In [3]: from sherpa.ui import list_models
l
In [4]: type(list_models())
Out[4]: list

In [5]: type(list_models('2d'))
Out[5]: filter

```

This is a consequence of moving to Python 3 . It can be fixed easily (add an explicit conversion to `list` for the filtered values), or we update the documentation.

Given these arrays aren't that large, I think the explicit conversion to a list makes sense for the UI layer.

DougBurke:
# Summary

Add the delete_pileup_model() function to allow a pileup model to be removed from a fit (issue #441), and list_psf_ids() and list_pileup_model_ids() routines to list those datasets with an associated PSF or pileup model. The list_models() routine no-longer returns an iterator but a list when given an option (issue #749).

# Details

This matches the existing set_psf/delete_psf routines (for delete_pileup_model) and list_data_ids/list_model_ids for the list_psf_ids and list_pileup_model_ids routines. 

One possibility for the pileup routines was to drop the "_model", but I wanted to keep consistency with set_pileup_model and I don't expect these routines to be used heavily, so having a long name is not a big concern.

The test_astro_ui.py test suite was updated to use pytest as part of this. In doing so I was able to identify a test previously marked as "skip" - test_psf_model1d  - because it was originally never ran (due to a name clash) and turned out to fail. I reworked the test to loop through the models at the test level (rather than within the test), which showed that only one of the three models was failing. I changed the test so that it passed for this test, and they are now all running. It's not clear whether the result makes sense but it now at least will be caught if we ever work on the psf-centering code.

The test_ui.py tests were also converted to pytest and cleaned up (so finally able to move a comment @olaurino added about how this would be better with pytest parameters).

I have also added some random tests from the session class that I noticed were't being covered (there's plenty more that are needed for sherpa.ui.utils but that can wait).

### DIFF
--- a/sherpa/astro/datastack/__init__.py
+++ b/sherpa/astro/datastack/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2010,2014,2015 Smithsonian Astrophysical Observatory
+# Copyright (c) 2010, 2014, 2015, 2020 Smithsonian Astrophysical Observatory
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -218,14 +218,14 @@ __all__ = ['set_template_id', 'DataStack']
 
 import sys
 import types
+import logging
+
 from sherpa.astro import ui
 from sherpa.utils.logging import config_logger
 from sherpa.utils import public
 from sherpa.astro.datastack.ds import DataStack
-import logging
-from . import plot_backend as backend
-from .utils import model_wrapper, set_template_id, load_wrapper,\
-    load_error_msg, simple_wrapper, fit_wrapper, plot_wrapper
+
+from .utils import set_template_id
 
 logger = config_logger(__name__)
 
--- a/sherpa/astro/datastack/plot_backend/__init__.py
+++ b/sherpa/astro/datastack/plot_backend/__init__.py
@@ -18,8 +18,9 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-from sherpa.plot import plotter
 from importlib import import_module
+
+from sherpa.plot import plotter
 from sherpa.utils.logging import config_logger
 
 logger = config_logger(__name__)
--- a/sherpa/astro/datastack/utils.py
+++ b/sherpa/astro/datastack/utils.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2015, 2016  Smithsonian Astrophysical Observatory
+# Copyright (C) 2015, 2016, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -16,10 +16,12 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+
+import re
+
 from sherpa.utils import public
 from sherpa.utils.logging import config_logger
 import sherpa
-import re
 
 from . import plot_backend as backend
 
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016, 2018, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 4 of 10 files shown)


DougBurke:
I'm going to ask @anetasie and @hamogu just to check that these routines make sense, then once that's been done I'll ping @dtnguyen2 .

I've labelled this as an easy review which it is, as long as you don't want to jump through all the test code that gets converted from unittest to pytest style.

DougBurke:
For @anetasie and @hamogu - the docstrings are

a) delete_pileup_model is https://github.com/sherpa/sherpa/blob/6c534cd94bbf4a7df0a4f5944014707ceec7e02e/sherpa/astro/ui/utils.py#L8955

b) list_pileup_model_ids is https://github.com/sherpa/sherpa/blob/6c534cd94bbf4a7df0a4f5944014707ceec7e02e/sherpa/astro/ui/utils.py#L8983

c) list_psf_ids is https://github.com/sherpa/sherpa/blob/6c534cd94bbf4a7df0a4f5944014707ceec7e02e/sherpa/ui/utils.py#L7192

hamogu:
I'm glad this is being worked on and have to problem in principle. I wonder though about the UI. We already have convolution models, we have rmfs (which are also convolution models), so do we really need to add two more functions to the UI layer? I admit that this is by far the easiest solution (and "Perfect is the enemy of good"), I just want to bring it up to make sure it's a conscious decision to go this way.

Separate question: Should these functions be added to the datastack code in https://github.com/sherpa/sherpa/blob/master/sherpa/astro/datastack/ds.py#L449 as well?

DougBurke:
@hamogu - we already have internal code for the pileup model as a separate "stack" to the convolution and psf code, so I'm not inventing a new thing here. I do agree with you that they do have a lot of similarity, but in no realistic universe do I have the time or energy to replumb the internals to unify this.

DougBurke:
@hamogu - you'll have to tell me whether datastack would benefit from these, as I don't really use it myself.

DougBurke:
Why is one build failing - https://travis-ci.org/github/sherpa/sherpa/jobs/713029271#L4401 - because list_<>_ids() is returning the wrong order - namely

```
>       assert ui.list_pileup_model_ids() == [1, "1"]
E       AssertionError: assert ['1', 1] == [1, '1']
E         At index 0 diff: '1' != 1
E         Use -v to get the full diff
```

The code returns sorted(vals, key=str) which should return [1, '1']. Confused.

hamogu:
@DougBurke 

Yes, datastack would benefit from these, so please add them in.

And no, I did not expect you to do a re-plumbing, I'm just mentioning it so it's documented if people look back and so its on your mind if we ever need to do a replumbing for some other reason and write specs for that at some point.

anetasie:
@DougBurke thanks for this. I managed to built the branch and tried it out to see it in action. It worked fine. One comment to the doc  'list_pileup_model' is not defined and I needed to use `list_pileup_model_ids'.

DougBurke:
@hamogu - it **looks like** we've automatically picked them up in datastack since we have

```
In [9]: from sherpa.astro import datastack

In [10]: [l for l in datastack.__all__ if 'pileup' in l]
Out[10]:
['delete_pileup_model',
 'get_pileup_model',
 'list_pileup_model_ids',
 'set_pileup_model']

```

So I think we don't need to change anything? As I said I'm not a datastack aficionado so may be missing something.

DougBurke:
I've fixed the list_pileup_model documentation typos reported by @anetasie

DougBurke:
@anetasie - I'm going to assume you're okay with this now, so I've asked for @dtnguyen2 to review it.

DougBurke:
Ive rebased, primarily to pick up the codecov changes

DougBurke:
I've rebased this to simplify the commit history (e.g. move all the test clean-up work to before the changes, and then combining fixes (such as the typos mentioned by Aneta) into the commits.

Issue #797, opened 2020-04-22
Failure with spectra that have vector-valued BACKSCAL


hamogu:
Sherpa cannot display fit and data that has a background model and a vector-valued BACKSCAL column.
This is relevant, because `combine_spectra ... bscale_method='counts'` is advertised in the `combine spectra` ahelp as the correct way to handle spectra where the background will be fitted, not subtracted. Thus, it would be natural to expect that sherpa can fit backgrounds spectra made this way.

In this case, sherpa cannot even do a `show_model`, mostly because the name for the model is a string expression that includes the value of the backscal parameters, which does not work if that parameters is a vector.

While fits run, the fact that I can neither show nor plot the result does not give me confidence that sherpa treats the vector-valued BACKSCAL correctly in other methods.

See script below.
The test files are generated by the lines below and are available here: [issue.tar.gz](https://github.com/sherpa/sherpa/files/4517751/issue.tar.gz)
```
combine_spectra 17764_A_grp.pi,19980_A_grp.pi 2017_A_counts bscale_method='counts' clob+
combine_spectra 17764_A_grp.pi,19980_A_grp.pi 2017_A_asca bscale_method='asca' clob+
combine_spectra 17764_A_grp.pi,19980_A_grp.pi 2017_A_time bscale_method='time' clob+
```
Here is the script and error message:
```
from sherpa.astro import ui
ui.load_data(1, 'data/Chandra/2017_A_asca_src.pi')
ui.load_bkg(1, 'data/Chandra/2017_A_asca_bkg.pi')
ui.load_data(2, 'data/Chandra/2017_A_counts_src.pi')
ui.load_bkg(2, 'data/Chandra/2017_A_counts_bkg.pi')
ui.load_data(3, 'data/Chandra/2017_A_time_src.pi')
ui.load_bkg(3, 'data/Chandra/2017_A_time_bkg.pi')
for i in [1,2,3]:
    ui.set_bkg_source(i, ui.scale1d(name='scaleB_early') * ui.xsvapec(name='Bv1_'))
    ui.set_source(i, ui.xsvapec(name='Av1_')) 
```
`show_model` works for 1 and 3 and fails for 2:
```
ui.show_model(2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-5f5362e05497> in <module>
----> 1 ui.show_model(2)

<string> in show_model(id, outfile, clobber)

[...]

Issue #629, opened 2019-05-16
Is something wrong with including background models into the source-model (when fitting both)


DougBurke:
I have a notebook going through my thinking at https://gist.github.com/DougBurke/588ab80aa06e72507269711a35d0c0ed

The problem I have is that with a "faked up" PHA dataset (with a background), to which I've applied a model to both the source and data, and I do not understand how Sherpa has calculated the model expression it uses to fit both, since I'd expect 'source model + scale * background model', but the scale factor returned by Sherpa doesn't match what I'd expect.

This could be user error.

DougBurke:
So, this is a combination of user errors, namely

a) I was ignoring the background scaling factor which does need to be corrected for when adding the background models to the source expression (this corrects the table I generate of "expected" scale values to match the Sherpa values when there's only 1 background component)

b) I was thinking that each background component is fully added to the source model (after applying the scaling factor), when in fact you need to add scale_i * bgnd_i / nbgnd for each background component (that is, normalize by the number of backgrounds). Just because we have multiple background datasets doesn't mean that our background contribution to the source aperture will increase!

as well as an error with the model display when we had multiple backgrounds:

```
In[19]:     ui.get_model('all').name
Out[19]:  'apply_arf((10 * (const1d.cpt + 0.55 * (const1d.bcpt + const1d.bcpt))))'
```

This display is actually ***wrong***, and what it should say is (neglecting excessive bracketing)

    apply_arf(10 * (const1d.cpt + 0.55 * (const1d.bcpt / 2 + const1d.bcpt / 2)))

or the simplified

    apply_arf(10 * (const1d.cpt + 0.55 * const1d.bcpt))

PR #869 addresses this last point.

DougBurke:
**ACTUALLY** I have come back to my conclusion that the background scaling to apply to the background model for the source aperture should **NOT** include the exposure correction (and not obvious what to do for AREASCAL, but possibly only applying the source aperture).

If you fit the background aperture, you fit `bkg_resp(tbkg * background_model)`, which means that the model is actually calculating a rate (per second) value. We then construct the source model and get `src_resp(tsrc * (source_model + bscale * background_model))` and the `bscale` value should **NOT** include the `tsrc/tbkg` correction factor the the exposure time. It should still contain the correction for the aperture size.

DougBurke:
[...]

PR #435, opened 2018-03-07
WIP: Work around failure cases when backscal is an array

DougBurke:
Work on addressing issues seen when a PHA file (e.g. grating dataset) has an array `BACKSCAL` value. This needs more work (e.g. tests being added) and the consequences worked out before it can be merged.

# Details

For grating PHA data sets the background scale value can vary with bin,
and so is returned as an array. This causes the creation of the
BackgroundSumModel instance - which is created whenever a model instance
is being created by sherpa.astro.ui functions like get_model and show_all -
to fail with a TypeError, as '%g' % backscale errors out.

This change traps this error and replaces the actual scaling value by
a token ("bkgscale"), which lets the process proceed. It is unclear
however what the semantics of the string representation of a model is:
is it purely for display purposes, in which this replacement is
okay, or is it meant to be actionable/machine readable (e.g. when
saved via the ui.save_all function)?

PR #869, opened 2020-07-29
Issues with fitting background and source datasets

### LINKED ISSUE #879: ui.set_bkg does more than datapha.set_background; should it?
You can see subtle differences if you have two DataPHA objects, `src` and `bkg`, and say

    sherpa.astro.ui.set_data(src)
    sherpa.astro.ui.set_bkg(bkg)

and

    pha.set_background(bkg)
    sherpa.astro.ui.set_data(src)

Now, there are legitimate reasons for the ui layer to do extra things, since it is a convenience layer, and if so we should try to be explicit in the documentation. However, I am trying to make it easier to use the object API, and it may make sense to move some functionality into set_background (or other methods).

In this case, the extra functionality that set_bkg adds is that

 - it copies over the grouping and quality data from the source if there is no information in the background
 - if there's no background response it copies over the source response and matches the units setting
 - it copies over the rate and plot_fac settings

I believe that this makes sense to be in the lower-level set_background routine (there may be times the user doesn't want this but it is likely to be a rare occurrence, and so can be  handled by the user post-processing the data).

DougBurke:
This PR has been replaced by the following PRs:

1. First is #882 which just provides a number of tests that I added in this PR (as well as reworking tests to use pytest rather than SherpaTestCase). 
2. Then is #884, which fixes #879 and #880
3. Then #888, which supports vector backscales and addresses other issues with fitting background data.

It would be nice if GitHub had [merge train](https://docs.gitlab.com/ee/ci/merge_request_pipelines/pipelines_for_merged_results/merge_trains/) support,

# Summary

When fitting source and background datasets simultaneously, there were times that the scaling factor for including background models into the source aperture was incorrectly including the ratio of exposure (source) to exposure (background), and so the background contribution to the source aperture was being systematically biased when the source and background exposure times are different.

This PR adds support for vector scale values for the conversion of background models from the background aperture to the source aperture (when fitting models to the source and background apertures). Prior to this change Sherpa would error out when generating the combined model terms. **TODO** after all the changes I need to ho back and review this statement.

The logic for setting background settings from the source dataset has been moved from sherpa.astro.ui.set_bkg to sherpa.astro.data.DataPha.set_background to remove potential confusion (issue #879). This has simplified the code path for generating the response used to fit a background model, removing some odd corner cases (issue #880).

The sherpa.astro.ui.get_bkg_scale routine now returns values for one background component, rather than the average value
for all the background components associated with the dataset.  It also now can calculate the scaling factor for background subtraction (units='counts') or when fitting simultaneously (units='rate').

The display of the source "model" expression - the output of `get_model().name` - was incorrect when multiple background datasets with models was present (the per-component model expression was not normalized by the number of background components). This is now fixed. Note that this **only** affected the screen display, the actual model calculation correctly normalized the background signal (for the common case when the same background model was used for all background datasets).

---

# Outstanding issues

 - [ ] should the new ScaleArray class just be replaced by ArithmeticConstantModel?
 - [x] have to warn when dealing with a pileup model
 - [ ] how best to handle the naming of the models used to store the vector scale values, the current approach picks a name 'scale<dataset id>_<bkg_id>' which is going to break if the dataset name has "interesting" characters in it, and has the potential to clobber ser models
 - [ ] can we delete the scale-array model when the dataset is cleared out
 - [x] get_backscal should handle multi bkg better: either give an id or return all, don't just combine them

# Details

As I've dug into the code there are a number of related issues, which makes it hard to separate into simple changes. If needed I can do so, bit I don't think any one particular problem is a lot of code, it's just all over the place.

## 1) How should the background be scaled when fitting?

This is #629 

When subtracting the background data you say

      source - r_exposure * r_backscale * r_areascal * bgnd

where r_x is the ratio of source x to background x (this assumes a
single background dataset but it extends to multiple
datasets).

When modelling the background we have to create the model (excluding
the response term)

      tsrc * (source_model + scale_factor * background_model)

where tsrc is the source exposure time and scale_factor corrects the
background model to the source aperture.

The existing code uses the same scaling factor as above - namely

      scale_factor = r_exposure * r_backscale * r_areascal

but this is wrong. The source and background models return rates
and not counts. This is why the source model starts with 'tsrc * '
since you need to convert to counts. This means that the ratio
between the source and background exposure times DOES NOT FACTOR
into scale_factor. It is unclear to me whether AREASCAL should be
included or not: XSPEC treats it as an exposure correction, so
I am also dropping it here, but this **could be** wrong.

The correct scale factor should therefore be

      scale_factor = r_backscale

that is, we just correct for the aperture size when adding the
background model to the source aperture.

Fortunately this is not likely to be a big problem with existing
users as I **IMAGINE** that the source and background exposures
are often the same, and so it doesn't matter.

It will matter for users fitting a background model to "blank-sky
datasets", where the exposure time is different in the source and
background spectra.

## 2) Using a vector to scale background to source apertures (when fitting)

This is code designed to look into #797 and #435. The first part is a number of code clean ups, adding documentation, and tests. The you get to "EXPERIMENT support vectors for background scale factors (when fitting)" which is a proposed solution to the problem. I have added and/or expanded test cases as needed, and improved some documentation.

Note that there is some overlap in the test cases added in #719 and in this test. They do test similar functionality, but they don't overlap completely so I think it's worth having both,

The sherpa.astro.background.BackgroundSumModel was only used in one place, and is no-longer used, so I've removed it.

One consequence of this PR is that you can not reliably use a pileup model when fitting the source + background of a dataset where there is a vector array defining the scaling of the background-to-source ratio. The code will run but it warns the user, and the output will not be correct (the amount depends on how much pileup there is).

## What is the problem

When fitting source + background, the source model (that is the expression that contains the instrument model) becomes (in principle, but not actually)

    response(exposure * (source_model + bscale * bgnd_model))

where the response is actually 'apply_rmf(apply_arf',  just to make this simple. The `bscale` value has to be a scalar (otherwise it errors out, as discussed below). If you call

    print(get_model())

then you sill see the above. An example is the following, where xsvapec.Av1 is the source model, xsvapec.Bv1 is the background, and the scale factor is 0.1458

```
>>> print(get_model().name)
apply_rmf(apply_arf((48631.388921332 * (xsvapec.Av1 + 0.1458 * (xsvapec.Bv1_)))))
```

This is actually an approximation of the truth, since if we look at the actual model, we have (breaking down the binary operations):

```
>>> mdl = ui.get_model()
>>> print(type(mdl))
<class 'sherpa.astro.instrument.RSPModelPHA'>
>>> mdl.parts[0]
<BinaryOpModel model instance '(48631.388921332 * (xsvapec.Av1 + 0.1458 * (xsvapec.Bv1)))'>
>>> mdl.parts[0].parts[0]
<ArithmeticConstantModel model instance '48631.388921332'>
>>> mdl.parts[0].parts[1]
<BinaryOpModel model instance '(xsvapec.Av1 + 0.1458 * (xsvapec.Bv1))'>
>>> mdl.parts[0].parts[1].parts
(<XSvapec model instance 'xsvapec.Av1'>,
 <BackgroundSumModel model instance '0.1458 * (xsvapec.Bv1)'>)
```

The trick here is that we actually have a sherpa.astro.background.BackgroundSumModel which provides a screen output of "scale * models" but internally it actually evaluates this a bit differently. This is okay, but I wanted to explain this to explain where the BackgroundSumModel class is used.

There's two issues with the BackgroundSumModel instance:

 - the initialisation tries to generate a string like `'%g * models" % bscale` (used in the string output above), **and this fails** when an array is used
 - the scale factors are applied to the model before it has been passed through the instrument response, and so need to have binning matching the instrument grid, and not the channel space. We never hit this problem because of the previous point.

In my original investigation in #435 I only tried to fix the first of these two problems. This let us create the model with a vector background scale, but trying to use it failed with size-mismatch errors (see the comments in that PR).

I note in #629 that the display of the source model was wrong when there are multiple background components (this is only the display, the actual model calculation is correct), The example from #629 was

    apply_arf(10 * (source_model + 0.55 * (bgnd_model + bgnd_model)))

when it should have said (some form of):

    apply_arf(10 * (source_model + 0.55 * (bgnd_mode / 2l + bgnd_model / 2)))

## How does this PR fix it

Instead of using BackgroundSumModel to do the magic of combining the background components I now explicitly create the model expression. The code is in `sherpa.astro.ui.utils._add_convolution_models`.

The process is roughly (for the case of a dataset with a background when we are trying to fit the background):

  - calculate the scale factor for each background component for the dataset
  - for each scale factor which is a scalar (in which case we don't need to do anything fancy)

         bmdls = scale_key1 * background_model_key1 + scale_key2 * background_model_key2 + ...

     where the _key suffix is just to point out this is done for each background component
  - if source_model is the source model expression, and resp is the instrument model,  then create

        out_model = resp(source_model + bmdls)

  - now we need to deal with the remaining vector scale values: if these are scale_j1/background_model_j1
    to _jn then we need to generate

        out_model += scale_j1 * resp(background_model_j1) + ... + scale_j2 * resp(background_model_jn)

    The trick is that we apply the response model to the background model, which converts to channel space,
    and then we can multiply by the array of values in channel space.

Unfortunately if will not work if resp() is non-linear - so resp(a) + resp(b) != resp(a + b) - which is the case of pileup models. I can not see any way around this while keeping the scale factor in channel units (and that's what the input is in).

One trick in the `scale_j1 * resp(..)` step is that `scale_j1` is actually a model (a tablemodel-like instance, added as sherpa.astro.background.ScaleArray, which acts to multiply by a numpy array of values. This is done because it makes the string output of the model a lot easier to handle (i,e, no attempt to include each value of the array). For reasons depending on the RSPModel*PHA classes, this always returns the full channel dataset (whatever the input arguments), in the same way that resp() does when it's a RSPModel... instance (filtering to match the cignore/notice channel range is done at a later stage).

As an example, the case above now becomes (variable names may change depending on exactly what  version of the code is used):

'(apply_rmf(apply_arf((48631.388921332 * xsvapec.Av1))) + (scalearray.scale2_1 * apply_rmf(apply_arf((48631.388921332 * xsvapec.Bv1)))))'

## Optimising for the expected use case

It is expected that the same background model is used for all components. Note that in this case the previous behavior of `sherpa.astro.ui.utils.get_bkg_scale` to return the average scaling factor is correct, it's only if you have different background models that that approach fails.

When we have the same background model used for multiple background components then we can reduce some computation, since we can convert (here assuming it's a scalar correction factor) 

    bmdls = scale_key1 * background_model_key1 + scale_key2 * background_model_key2 + ...

to

    bmdls = (scale_key1 + scale_key2 + ...) * background_model

# Examples

I've read in a grating spectrum and am fitting overly-simple models: the source is an xsvapec and the background is a powerlaw.It's not physically meaningful, but that's not to relevant here.

Here's the names of the various "models" - e.g. the "source" and "model" versions which refer to the models the user sets and the one that Sherpa actually uses [the use of "model" in this case conflicts with the general use of "model" and I wished we'come up with a better term...].

```
In [62]: print(get_source().name)
xsvapec.vsrc

In [63]: print(get_bkg_source().name)
xspowerlaw.bpl

In [64]: print(get_bkg_model().name)
apply_rmf(apply_arf((158805.39418195 * xspowerlaw.bpl)))
```

So far as expected. 

**NOTE** the next command fails in master branch with the error message

```
~/sherpa/sherpa-xspec-12101/sherpa/astro/background.py in __init__(self, srcdata, bkgmodels)
     50         scale_factor = self.srcdata.sum_background_data(lambda key, bkg:1)
     51         bkgnames = [model.name for model in bkgmodels.values()]
---> 52         name = '%g * (' % scale_factor + ' + '.join(bkgnames) + ')'
     53         CompositeModel.__init__(self, name, self.bkgmodels.values())
     54

TypeError: only size-1 arrays can be converted to Python scalars
```

The model name looks okay, but if you look carefully you'll see it is actually different:

```
In [65]: print(get_model().name)
(apply_rmf(apply_arf((158805.16994966 * xsvapec.vsrc))) + (scalearray.scale1_1 * apply_rmf(apply_arf((158805.16994966 * xspowerlaw.bpl)))))
```

The differences are

a) instead of `apply_rmf(apply_arf(src + scale * bgnd))` we now have  `apply_rmf(apply_rmf(src)) + scale * apply_rmf(apply_arf(bgnd))`

b) the scale factor is not a number but a model component: `scalearray.scale1_1`

You can view this component (it has been created for you so the following works), and its y attribute contains the actual scaling array:

```
In [66]: print(scale1_1)
scalearray.scale1_1
   scale1_1.ampl frozen            1 -3.40282e+38  3.40282e+38

In [68]: scale1_1.y.shape
Out[68]: (8192,)

In [69]: scale1_1.y[0:4]
Out[69]: array([0.99999859, 0.99999859, 0.99999859, 0.99999859])
```

If you print out the get_model() you see it as part of the model parameters

```
In [67]: print(get_model())
(apply_rmf(apply_arf((158805.16994966 * xsvapec.vsrc))) + (scalearray.scale1_1 * apply_rmf(apply_arf((158805.16994966 * xspowerlaw.bpl)))))
   vsrc.kT      thawed     0.586205       0.0808       68.447        keV
   vsrc.He      thawed   0.00242171            0         1000
   vsrc.C       thawed   0.00139063            0         1000
   vsrc.N       thawed    0.0320576            0         1000
   vsrc.O       thawed     0.587276            0         1000
   vsrc.Ne      thawed     0.273331            0         1000
   vsrc.Mg      thawed     0.279274            0         1000
   vsrc.Al      thawed     0.979798            0         1000
   vsrc.Si      thawed     0.326917            0         1000
   vsrc.S       thawed      1.07481            0         1000
   vsrc.Ar      thawed      26.8577            0         1000
   vsrc.Ca      thawed      8.06943            0         1000
   vsrc.Fe      thawed     0.199196            0         1000
   vsrc.Ni      thawed  0.000675637            0         1000
   vsrc.redshift frozen            0       -0.999           10
   vsrc.norm    thawed   0.00111689            0        1e+24
   scale1_1.ampl frozen            1 -3.40282e+38  3.40282e+38
   bpl.PhoIndex thawed      4.64582           -2            9
   bpl.norm     thawed  0.000494137            0        1e+24
```

---

# NOTES

From the existing datasets I've investigated, I think there's a general issue in the BACKSCAL values when it's a vector and they come from combined datasets. The values are often <some value when there are 0 counts> and then the actual scale values. This works for when we subtract the background, because the <some value when there are 0 counts> calue gets multiplied by 0 ,so it doesn't matter what the value is. When we are trying to fit the background then we care about each channel (not just those with counts in the background), and then **how** do we deal with the <some value when there are 0 counts>, as it is generally not correct from what I've seen.

This is not directly an issue with this PR ,but it may indicate whether we have any sensible data for fitting.

### DIFF
--- a/docs/evaluation/astro_background.rst
+++ b/docs/evaluation/astro_background.rst
@@ -11,9 +11,17 @@ The sherpa.astro.background module
    .. autosummary::
       :toctree: api
 
-      BackgroundSumModel
+      ScaleArray
+
+   .. rubric:: Functions
+
+   .. autosummary::
+      :toctree: api
+
+      add_response
+
 
 Class Inheritance Diagram
 =========================
 
-.. inheritance-diagram:: BackgroundSumModel
+.. inheritance-diagram:: ScaleArray
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -22,58 +22,11 @@
 """
 
 from sherpa.models.parameter import Parameter
-from sherpa.models.model import ArithmeticModel, CompositeModel, \
-    modelCacher1d
+from sherpa.models.model import ArithmeticModel, modelCacher1d
 
-from sherpa.utils.err import DataErr, ModelErr
+from sherpa.utils.err import ModelErr
 
-__all__ = ('BackgroundSumModel', 'ScaleArray')
-
-
-class BackgroundSumModel(CompositeModel, ArithmeticModel):
-    """Combine multiple background datasets.
-
-    Define the model expression to be applied to the source
-    region accounting for the background models.
-
-    Parameters
-    ----------
-    srcdata : sherpa.astro.data.DataPHA instance
-        The source dataset.
-    bkgmodels : dict
-        The background components, where the key is the identifier
-        in this dataset and the value is the model. This dictionary
-        cannot be empty.
-
-    """
-
-    def __init__(self, srcdata, bkgmodels):
-        self.srcdata = srcdata
-        self.bkgmodels = bkgmodels
-        scale_factor = self.srcdata.sum_background_data(lambda key, bkg:1)
-        bkgnames = [model.name for model in bkgmodels.values()]
-        name = '%g * (' % scale_factor + ' + '.join(bkgnames) + ')'
-        CompositeModel.__init__(self, name, self.bkgmodels.values())
-
-    def calc(self, p, *args, **kwargs):
-        def eval_bkg_model(key, bkg):
-            bmodel = self.bkgmodels.get(key)
-            if bmodel is None:
-                raise DataErr('bkgmodel', key)
-            # FIXME: we're not using p here (and therefore assuming that the
-            # parameter values have already been updated to match the contents
-            # of p)
-            return bmodel(*args, **kwargs)
-
-        # Evaluate the background model for each dataset using the same
-        # grid and apply the background-to-source correction factors.
-        #
-        # This will only work if the scaling factors are scalars,
-        # since if there are any array elements then the models
-        # have been evaluated on the energy/wavelength grid,
-        # but the correction factors are defined in channel space.
-        #
-        return self.srcdata.sum_background_data(eval_bkg_model)
+__all__ = ('ScaleArray', )
 
 
 class ScaleArray(ArithmeticModel):
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -53,8 +53,6 @@ class ScaleArray(ArithmeticModel):
     ----------
     ampl
         The amplitude of the model.
-    y
-        The array of values.
 
     Examples
     --------
@@ -77,7 +75,7 @@ class ScaleArray(ArithmeticModel):
 
     @property
     def y(self):
-        """The data to return"""
+        """The array of values."""
         return self._y
 
     @y.setter
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -21,12 +21,23 @@
 
 """
 
+from collections import defaultdict
+from functools import reduce
+import logging
+import operator
+
+import numpy
+
+from sherpa.astro.instrument import PileupResponse1D
 from sherpa.models.parameter import Parameter
 from sherpa.models.model import ArithmeticModel, modelCacher1d
-
 from sherpa.utils.err import ModelErr
 
-__all__ = ('ScaleArray', )
+
+__all__ = ('ScaleArray', 'add_response')
+
+
+warning = logging.getLogger(__name__).warning
 
 
 class ScaleArray(ArithmeticModel):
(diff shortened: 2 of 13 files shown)


DougBurke:
@hamogu - technically this lets you use your data from #797 but I think that the results will not be as illuminating as you'd hope for (given the discussions we've had about the contents of the BSCALE column in this file)

anetasie:
@DougBurke There should not be an issue with the background data with 0 counts in the simultaneous fit of background and source data. The backscal is applied to the total model when calculating the statistics. The fact that there are 0 counts in the background data should not matter.

DougBurke:
@anetasie - but we are fitting to the background here. Which means we compare the data to (excluding the response terms)

    source_model + scale * background_model

and the problem is that the scale value (when we have an array of them) isn't well defined when tools like combine_spectra is used (which have to use the observed counts and so it's not obvious what to use when there's 0 counts). We end up multiplying the background model by "odd" values. I will try to dig up an example showing how odd the results look.

DougBurke:
Rebased to get new codecov support and recent changes.

DougBurke:
Hmmm, that last commit was marked as failing because of the codecov/project check. What is this malarkey?

PR #873, opened 2020-08-04
Add requires_fits decorators to test_xspec_con.py

DougBurke:
# Summary

Add decorators to two test routines that were missing (it requires XSPEC to be built but no I/O module available, which is an uncommon choice).

# Details

I had forgotten these three decorators. Our test builds don't have
a build with XSPEC but no IO library, which is why this is not seen.


DougBurke:
I'm sure I made the same mistake recently in another test (and fixed) Ahh, found it in #839 - so at least I'm consistently wrong ...!

PR #876, opened 2020-08-06
Add 'est_method_args' to fulfill CAT5 specfit.py requirements.

jbudynk:
For the Chandra Source Catalog 2.0 development (CAT4.5 build), we asked to add a new parameter 'est_method_args' to sherpa.astro.sample_flux(). These changes were added to branch cat45_ciao49.  For the Chandra Source Catalog 2.1 build (CAT5.1), we updated sherpa to release 4.12.0, and still require the 'est_method_args' parameter in sample_flux().

These PR changes are basically the same changes that are on the cat45_ciao49 branch, but put on top of the sherpa 4.12.0 release. The CAT pipelines team would like to have the changes on a new branch so we can pull from there when we build CAT like we did for CAT4.5, without affecting sherpa's main development branch.

Related JIRA issues:
 PTCAT-2733: Update all tools to be python3/ciao4.12 compatible.
 PTCAT-2754: Replace PyChips with matplotlib in specfit.py

### DIFF
--- a/sherpa/sim/sample.py
+++ b/sherpa/sim/sample.py
@@ -108,7 +108,10 @@ class ParameterScale(NoNewAttributesAfterInit):
 
 class ParameterScaleVector(ParameterScale):
 
-    def get_scales(self, fit, myscales=None):
+    def get_scales(self, fit, myscales=None, est_method_args=None):
+
+        if est_method_args is None:
+            est_method_args = {}
 
         scales = []
         thawedpars = [par for par in fit.model.pars if not par.frozen]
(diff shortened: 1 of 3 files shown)


DougBurke:
It  looks like this is failing because there's a whole bunch of infrastructure updates that have landed in the master branch that are needed to get the tests running on Travis.

sherpa/sim/sample.py
```diff
@@ -119,7 +122,8 @@ def get_scales(self, fit, myscales=None):
 
             covar = Covariance()
             covar.config['sigma'] = self.sigma
-            fit.estmethod = Covariance()
```
DougBurke, line 122: Hmmm, this looks like a logical error in out code. It's fixed in this PR but we should check.

PR #882, opened 2020-08-10
Add tests of background handling

DougBurke:
# Summary

Update tests to use the current best practices, and add tests of background handling (primarily when fitting a background model).

# Details

The new tests are not expected to significantly improve the test coverage, but they do provide a number of explicit checks for the planned background rework. It has been separated out of #869 since this is an easy set of tests, and it's useful to have the tests in place before the code gets changed.

The number of commits and files changed in this PR are:

     11 sherpa/astro/ui/tests/test_astro_ui_background.py
      3 sherpa/astro/ui/tests/test_astro_ui_plot.py
      2 sherpa/astro/ui/tests/test_serialize.py
      1 sherpa/conftest.py
      1 sherpa/stats/tests/test_wstat.py
 
Changes are:

## sherpa/conftest.py

Adds the hide_logging fixture, which runs a test with the sherpa logging level set to the ERROR level, and then returns it to it's previous level.

## sherpa/stats/tests/test_wstat.py

This changes the tests from unittest to pytest.

## sherpa/astro/ui/tests/test_serialize.py

This changes the tests from unittest to pytest and removes an old comment. Oh, I've just notices one of the tests can be written to use 'with pytest.raises' but that can wait,

## sherpa/astro/ui/tests/test_astro_ui_plot.py

Some of the test values are adjusted to make it a bit more obvious what is changing: we look at some combination of the ratio of energy-grid width, exposure, backscal, and areascal values in various tests, and I had chosen values that were not necessarily unique for the various values. The big improvement is moving from a uniform energy grid to a vector, but there are also other changes too.

There are also several tests where I remove some calls (explicitly setting the units of the background dataset) which I had added to work around #879 (but I had dropped the ball and never reported it as an issue at that time, September 2019). This is to make the fixes and change in behavior in #884 more obvious.

## sherpa/astro/ui/tests/test_astro_ui_background.py

This is a new set of tests which I created to provide some basic tests of the functionality I plan to test. Some of these are 'upfront' new tests, checking things that appear to be relevant, and some were added during development (i.e. that either encode a useful invariant or ones that show a change in behavior in either #884 or #888. The focus is handling of the background dataset, or datasets, of a PHA file. To make the checks easier, most of the tests use artificial datasets, created on the fly. Of particular interest is the various model expressions created for the source and background datasets (when fitting, it subtracting, the background): checking the string output of the model and evaluating the model.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -73,7 +73,7 @@ def example_pha_data():
     d = ui.DataPHA('example', _data_chan.copy(),
                    _data_counts.copy(),
                    exposure=etime,
-                   backscal=0.1)
+                   backscal=0.2)
 
     a = ui.create_arf(_energies[:-1].copy(),
                       _energies[1:].copy(),
@@ -136,7 +136,7 @@ def example_bkg_model():
     ui.create_model_component('powlaw1d', 'bcpt')
     bcpt = ui.get_model_component('bcpt')
     bcpt.gamma = 0.0  # use a flat model to make it easy to evaluate
-    bcpt.ampl = 1e-1
+    bcpt.ampl = 0.4
     return bcpt
 
 
@@ -519,7 +519,7 @@ def test_get_bkg_model_plot(idval, direct, clean_astro_ui):
 
     # TODO: this is the same output as test_get_bkg_model_plot_energy,
     #       which doesn't make sense.
-    yexp = _arf / 100
+    yexp = _arf / 25
     if direct:
         yexp /= _bexpscale
 
@@ -571,7 +571,7 @@ def test_get_bkg_model_plot_energy(idval, direct, clean_astro_ui):
 
     # TODO: The factor of 100 comes from the bin width (0.1 keV), but
     # why is there a scaling by _bexpscale?
-    yexp = _arf / 100
+    yexp = _arf / 25
     if direct:
         yexp /= _bexpscale
     else:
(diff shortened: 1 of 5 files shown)


DougBurke:
You can see from the coverage that - other than a show_data test which now gets coverage when it didn't before - most of the improvements are to catch some conditional checks which we used to not get. As I mentioned, this is less important than the fact that the tests do provide a number of functional tests which I will be checking (or changing the tests) when the code gets fixed.

dtnguyen2 (approved):

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -92,10 +111,19 @@ def example_pha_data():
...
     """Create an example data set with background
 
     There is no response for the background.
+
+    Parameters
+    ----------
+    direct : bool, optional
+        If True then the background is added to the source
+        and only the source is returned. If False then the
+        return is (source, bgnd) and the backgrouund is not
```
wmclaugh, line 124: typo- backgrouund -> nackground

DougBurke: As I had to rebase because of a conflict I fixed this in the original commit: https://github.com/sherpa/sherpa/pull/882/commits/74c53ef9c2e36c9a553d0b41c053f0ecc124c472#diff-a6ca41fadd396de931aed280f1fbd433R105

DougBurke: Oops - I'm going to have to rebase again, so the commit will change, but the fix has been made.

DougBurke:
I've rebased to fix the typo and, more importantly, deal with the conflict in sherpa/conftest.py.

Issue #880, opened 2020-08-07
Differences in the "full" model created for a background dataset


DougBurke:
The ui level has the following logic in it when creating the "full" model that is used to fit a model to a background dataset (

        if len(bkg.response_ids) != 0:
            resp = sherpa.astro.instrument.Response1D(bkg)
        else:
            resp = sherpa.astro.instrument.Response1D(data)

For context, this is https://github.com/sherpa/sherpa/blob/master/sherpa/astro/ui/utils.py#L9134

So, if the background has a response then we use the background dataset (which will include things like filtering and grouping, as well as exposure time, which may be different to the source dataset). If the background dataset has no response then it uses the source dataset, and so any filtering specific to the background will be lost (and exposure-time differences).

I believe we should always be using the bkg dataset here.

DougBurke:
It is related to #879 (since the presence of a background response can depend on the logic of #879).

DougBurke:
See also #629

DougBurke:
Aha, fixing #879 means that we can turn this check around, so that if there's no response then

 - the user must have reset the response for some reason
 - which means we can error out

Which is much easier than my initial attempt.

PR #884, opened 2020-08-11
Move logic from ui layer to DataPHA class: background responses (fix #879, #880)

### LINKED ISSUE #879: ui.set_bkg does more than datapha.set_background; should it?

DougBurke:
# Summary

Move the logic for adding a background response, if one doesn't exist, from the UI layer to the DataPHA class. This clears up several corner cases. Fixes #879 and #880.

There is the potential for differences in the results of fitting a model to the background dataset due to these changes, but it requires use of methods from the object API to trigger, rather than uses of just the sherpa.astro.ui routines, and the new behavior is more consistent.

# Details

There are several minor changes not directly related to #879 and #880.

## Current behavior

If you have a PHA file with BACKFILE, ANCRFILE, and RESPFILE set (but not for the background file), then you have the following (I use `unpack_pha` as if it holds here then it holds for `load_pha`):

```
>>> pha = sherpa.astro.ui.unpack_pha('3c273.pi')
read ARF file ...
read RMF file ...
read background file ...
>>> arf, rmf = pha.get_background().get_response()
>>> assert isinstance(arf, ui.DataARF)
>>> assert isinstance(rmf, ui.DataRMF)
```

Now I a going to remove the background component and then re-add it (to simulate adding a background to a PHA dataset which doesn't have a BACKFILE keyword)

If I add the background using the DataPHA method (`set_background`) then the response is not set:

```
>>> pha.delete_background()
>>> bkg = ui.unpack_pha('/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest/3c273_bg.pi')
>>> bkg.get_response()
(None, None)
>>> pha.set_background(bkg)
>>> bkg.get_response()
(None, None)
>>> pha.get_background().get_response()
(None, None)
```

If I Instead use the UI layer then note that the background response is set (copied from the source region):

```
>>> ui.set_data(pha)
>>> pha.delete_background()
>>> ui.set_bkg(bkg)
>>> bkg.get_response()
(<DataARF data set instance '...f'>,
 <DataRMF data set instance '...'>)
```

## This PR

This moves the logic for setting the background response (if it isn't set) from `ui.set_bkg` to `DataPHA.set_background` so in both cases above the background response (once associated with a source dataset) is set and not `(None, None)`. It then means that we can assume a background response exists in the UI layer, which simplifies things (and, if the response does not exist, it means the user has removed it on purpose, so we can error out when a response is needed, rather than trying to fix one up when needed).

### DIFF
--- a/sherpa/astro/background.py
+++ b/sherpa/astro/background.py
@@ -62,4 +62,12 @@ class BackgroundSumModel(CompositeModel, ArithmeticModel):
             # of p)
             return bmodel(*args, **kwargs)
 
+        # Evaluate the background model for each dataset using the same
+        # grid and apply the background-to-source correction factors.
+        #
+        # This will only work if the scaling factors are scalars,
+        # since if there are any array elements then the models
+        # have been evaluated on the energy/wavelength grid,
+        # but the correction factors are defined in channel space.
+        #
         return self.srcdata.sum_background_data(eval_bkg_model)
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1142,7 +1142,7 @@ class DataPHA(Data1D):
         return scale
 
     def get_backscal(self, group=True, filter=False):
-        """Return the area scaling of the PHA data set.
+        """Return the background scaling of the PHA data set.
 
         Return the BACKSCAL setting [BSCAL]_ for the PHA data set.
 
@@ -1169,7 +1169,7 @@ class DataPHA(Data1D):
         to the total number of image pixels. The fact that there is no
         ironclad definition for this quantity does not matter so long
         as the value for a source dataset and its associated
-        background dataset are defined in the similar manner, because
+        background dataset are defined in the same manner, because
         only the ratio of source and background BACKSCAL values is
         used. It can be a scalar or an array.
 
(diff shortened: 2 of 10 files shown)


DougBurke:
Rebased to fix conflicts because of recent PRs.

DougBurke:
Rebased to address test conflict.

dtnguyen2 (approved):

DougBurke:
Rebased to fix conflict.

PR #892, opened 2020-08-14
Attempt to support building on C++98 compliant compilers (XSPEC code)

DougBurke:
# Summary

Support building the XSPEC interface using compilers which default to the C++98 standard.

# Details

The std::begin/end templates, and the use of lambda expressions
requite C++11. Rather than try to come up with a C++98 version of
std::transform, I just went back and used the old C-style loop.

The contents of the macro is based on a check against __cplusplus,
which is not universally defined but is hopefully standard on the
platforms we are targeting. Rather than picking a potential version,
which could be confusing, I trigger on whether the value is larger
than the C++98 standard or not (a simplification of the code given
in https://github.com/sherpa/sherpa/pull/859#issuecomment-673294542 )

This should not be needed, as we should be requiring/sending -std=c++11 but there appears to be issues somewhere.

Unfortunately this code is not exercised on the Travis infrastructure (the compilers seem to be new enough that they are defaulting to the C++11 standard).

### DIFF
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -62,6 +62,21 @@ namespace sherpa { namespace astro { namespace xspec {
 typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
 typedef float FloatArrayType;
 
+// Try and support the use of std::transform while still building
+// against C++-98 compilers.
+//
+#if __cplusplus > 199711L
+#define CONVERTARRAY(orig, out, npts)					\
+        std::transform(std::begin(orig), std::end(orig), std::begin(out), \
+                       [](const double val) -> FloatArrayType { return static_cast<FloatArrayType>(val); });
+#else
+#define CONVERTARRAY(orig, out, npts)					\
+	for (int i = 0; i < npts; i++) { \
+          out[i] = (FloatArrayType) orig[i]; \
+        }
+#endif
+
+
 // XSpec models can be called from Sherpa using either
 //   - a single array for the grid
 //   - two arrays for the grid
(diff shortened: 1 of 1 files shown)


DougBurke:
I cancelled the first set of tests as I wanted to get the update requested by @dtnguyen2 in.

DougBurke:
@Marie-Terrell I hope this will work for you (and that the Travis tests pass too...).

hamogu:
It's not just Marie. I just tried to build the master branch on my work machine with CentOS7 and failed because of the default gcc version installed there.

DougBurke:
So more testers then: awesome!

hamogu:
Works on mt machine (i.e. it compiles again).

DougBurke:
@dtnguyen2 - just to note I made the change you requested so it can be re-reviewed.

Marie-Terrell:
Just ran a quick test on two of our machines and it does seem to fix the build errors. Though, I did not run any checks besides making sure the build completed.

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -62,6 +62,21 @@ namespace sherpa { namespace astro { namespace xspec {
...
+// Try and support the use of std::transform while still building
+// against C++-98 compilers.
+//
+#if __cplusplus > 199711L
+#define CONVERTARRAY(orig, out, npts)					\
+        std::transform(std::begin(orig), std::end(orig), std::begin(out), \
+                       [](const double val) -> FloatArrayType { return static_cast<FloatArrayType>(val); });
+#else
+#define CONVERTARRAY(orig, out, npts)					\
+	for (int i = 0; i < npts; i++) { \
```
wmclaugh, line 74: when possible, you should avoid using single character variable names-  int i,   float f, etc...  Using indexes such as   int ii,   float ff  improves readability and searching (ii will generate fewer false hits than i).

Also,  neglecting optimization and loop unrolling,  reversing indexing for simple independent assignments

     for (int ii = npts; ii--; ) { }

will generally provide more efficient code than 

    for (int ii = 0; ii < npts; ii++) { } 

when compiled the same way.

DougBurke: @wmclaugh - all I did here was resurrect the old code (the for loop) that we had before #859 was merged. I didn't want to spend time complicating things with changes to code that we knew already worked (and that we can hopefully remove soon). If someone wants to fix up this PR to make these changes then they are more than welcome to, but I feel I've done enough with this patch as is.

Issue #867, opened 2020-07-28
change random sampling to use the new numpy 1.17 generators


DougBurke:
NumPy 1.17 has introduced an improved way to generate random numbers. It would be good to switch to this once we can drop pre-1.17 support. Information at https://numpy.org/doc/stable/reference/random/index.html

DougBurke:
I wonder if we can use this new functionality to make the testing of our MonteCarlo-related routines more robust, by allowing a known (or fixed) generator to be used when testing these routines (like `get_draws` or `sample_energy_flux`)?

hamogu:
I suggest that we just use the new system whenever we touch code anyway or when a test breaks and we need to examine it more closely. We could go back and change any occurrences in Sherpa, but that's not a high priority item.

DougBurke:
That's okay for the tests. For actual code we'll need to think about how we want to do it - probably just add a parameter, but then what's the default yadda yadda.

DougBurke:
Argh - how do I tell GH that #1753 is not going to magically fix this. Grrr.

PR #893, opened 2020-08-15
Tests: run FullBayes as well as PragBayes (and hopefully all codecov tests)

DougBurke:
# Summary

Increase coverage of the pragbayes and fullbayes samplers and fix an error condition.

# Details

Since FullBayes builds off PragBayes we can run the same tests, particularly as there's actually no check of the validity of the response. There is one change however, since FullBayes can not be run with a 'simarf' (just 'pcaarf'), so we need to check for this error.

There is also an attempt to check for the LimitError check (that the bounds are exceeded during a draw). Since the fits are randomised and we haven't moved to the NumPy 1.17 random interface, which might make it easier to pick a particular seed/random generator for a test this is somewhat ad-hoc. The test itself is written to try and force quickly hitting the limits, with some "interesting" choice of models and parameter limits.

This PR is aimed at fixing some of the recent coverage failures we've seen because the pragbayes coverage has decreased, in code completely unrelated to the pragbayes code. It would be nice to actually be able to test the return values of the draws, but it's not obvious how best to test this (and is something missing in the current code base).

There's one code fix, which is hard to test: if a covariance error is raised when calculating the proposed draw it's meant to reject this iteration and carry on, but if this happens the first time the system is run then the warning message will error out. I've changed the warning message to contain a more-meaningful set of data.

### DIFF
--- a/sherpa/astro/sim/pragbayes.py
+++ b/sherpa/astro/sim/pragbayes.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2016, 2017, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2011, 2016, 2017, 2019, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
--- a/sherpa/astro/sim/pragbayes.py
+++ b/sherpa/astro/sim/pragbayes.py
@@ -231,8 +231,8 @@ class WalkWithSubIters(Walk):
                     try:
                         proposed_params = self._sampler.draw(current_params)
                     except CovarError:
-                        info("Draw rejected: covariance matrix failed. " +
-                             str(proposed_params))
+                        info("Draw rejected: covariance matrix failed: " +
+                             "{}".format(current_params))
                         # automatically reject if the covar is malformed
                         self._sampler.reject()
                         continue
--- a/sherpa/astro/sim/tests/test_astro_sim.py
+++ b/sherpa/astro/sim/tests/test_astro_sim.py
@@ -20,6 +20,8 @@
 
 import logging
 
+import numpy as np
+
 import pytest
 
 from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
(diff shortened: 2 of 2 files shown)


DougBurke:
As I am not able to reliably test the one line I changed the code-coverage test fails here.

anetasie:
@DougBurke I thought that FullBayes functionality has not been fully supported. Need to check

DougBurke:
All the test does is to run the code. It doesn't check the actual results (this is also a problem for the existing pragbayes tests).

DougBurke:
Just updated the copyright year, @dtnguyen2 -eagle-eyes.

DougBurke:
Aha : https://travis-ci.org/github/sherpa/sherpa/jobs/721541457 is a run where one of the random tests failed. We should think about how we should test such functions - allow occasional errors like this one, run many-more iterations, remove the problematic test, or be clevererer.

hamogu:
Suggestion: In CI, run the tests with a fixed random seed. Try to find a seed that covers as much as possible. Then, make extra tests with random seed, that's not run by normal CI (because we really want to test not just one input value). Those tests should be run by hand for any work that touches this code (e.g. they could be copy & paste code in the `test_*.py` or a test that's commented out and an inline comment in the functions to comment that in an try by hand if it works for any code changes).

When we get more of these cases we can do something like astropy: Use a special pytest fixuture (something like `@long_runnig_test`) for tests that run way longer (here run with random seed long enough that all lines are triggered). That fixture evaluates to "skip" unless the tests are called with a `--run-long-tests`. Those tests are skipped by default in CI, but run with a cron job once a week on the master branch.

DougBurke:
I'm hoping, as mentioned in #867, that we can eventually move to the new numpy rng system which might make this better, but that's a long way off.

We could creat a decorator that sets a specific seed for the test (I think this has to be done per test, otherwise things like adding a new test could change things).

hamogu:
I'm refusing to look at code this week, but I think a call `numpy.random.seed(12345)` just before the function in questions is called in the test should be sufficient. That's just one line added to the test and no changes to the actual code.

PR #898, opened 2020-08-18
Move to src layout

DougBurke:
This is an experiment, and will be useful for #784. I'm really doing this because of #903, but being stymied.


DougBurke:
@hamogu - what is the current thinking of the "src/" layout approach (i.e. having src/sherpa/... rather than sherpa/ like we currently do). I find the current packaging landscape for Python more like quicksand, and don't know of any easily-understood documentation. I need something between the "here's a simple python-only install guide" and "this is the raw details of setuptools"...

DougBurke:
I also wonder if we should just require setuptools - there are a few places where it appears to be optional, but with caveats,and I wonder how well the code would work without setuptools (I think these checks are old).

hamogu:
@DougBurke How about this https://packaging-guide.openastronomy.org/en/latest/index.html

DougBurke:
So, I have no idea why the travis develop builds insist on telling me thigs lile

```
    from sherpa.utils._utils import hist1d, hist2d
E   ModuleNotFoundError: No module named 'sherpa.utils._utils'
```

since the module has been built and is mentioned by the egg-generating code. So I wonder if there's something not-quite-right about our extension code (building the compiled modules) - i.e. it's not so much the _utils module but any of the modules.

DougBurke:
Closing this because it turned out to not be useful. I'll leave the branch around, at least for  a while.

PR #899, opened 2020-08-19
Update likelihood descriptions

anetasie:
This PR updates likelihood description in several doc-strings and clarifies descriptions of statistics.

### DIFF
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -288,7 +288,7 @@ class Stat(NoNewAttributesAfterInit):
 
 
 class Likelihood(Stat):
-    """Maximum likelihood function"""
+    """Likelihood functions"""
 
     def __init__(self, name='likelihood'):
         Stat.__init__(self, name)
@@ -341,7 +341,7 @@ class Likelihood(Stat):
 
 
 class Cash(Likelihood):
-    """Maximum likelihood function.
+    """Poisson Log-likelihood function.
 
     Counts are sampled from the Poisson distribution, and so the best
     way to assess the quality of model fits is to use the product of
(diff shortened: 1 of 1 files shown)


DougBurke:
@anetasie - I haven't had a chance to look at this yet (I was out yesterday), but just to note that I've changed some magic behind the scenes so if ou make any more commits to this branch you'll get a build of the documentation on read-the-docs - it'll look something like

https://sherpa--899.org.readthedocs.build/en/899/

anetasie:
@DougBurke ok. the link right now shows 'page does not exist', so I assume the build will happened the next time I commit a doc update.

DougBurke:
Yes. As I said, you need to commit a change for this to happen.

sherpa/stats/__init__.py
```diff
@@ -681,6 +703,14 @@ class Chi2DataVar(Chi2):
     in the fit. The background term appears only if an estimate of the
     background has been subtracted from the data.
 
+    A(B) is the off-source "area", which could be
+    the size of the region from which the background is extracted, or
+    the length of a background time segment, or a product of the two,
+    etc.; and A(S) is the on-source "area". These terms may be defined
+    for a particular type of data: for example, PHA data sets A(B) to
+    `BACKSCAL * EXPOSURE` from the background data set and A(S) to
+    `BACKSCAL * EXPOSURE` from the source data set.
```
DougBurke, line 712: Technically we also have to deal with AREASCAL but I'm not sure that's worth mentioning here (as it is a bit odd).

wmclaugh:
When updating files containing a copyright, please add the current year if it is not already there.

PR #888, opened 2020-08-13
Support vector backscales and bugfix for background modelling

DougBurke:
# Summary

Support fitting backgrounds to PHA datasets which have a variable BACKSCAL array (rather than a scalar), which can come from combining spectra (e.g. the CIAO contrib script `combine_spectra`) or from the data extraction process. In doing so a number of routines related to the scaling of background-to-source aperture data have seen adjustments to behavior and some enhanced functionality (such as `sherpa.astro.ui.get_bkg_scale` and the `sherpa.astro.data.DataPHA.get_background_scale` method). This fixes #797.

One issue with the variable-array support is that it is only correct for linear responses, such as the normal ARF and RMF pairing, where doubling the model amplitude (all other things being equal) will double the number of predicted counts. This is not the case if the `jdpileup` model is used to model the ACIS pileup phenomenon: in this case a warning message will be displayed, but the fit can still be run (there is no guarantee on the accuracy of the result).

There have also been changes to how the background scale factor has been calculated when fitting the background as part of the source spectrum, since it previously included the exposure-time and areascal ratios which are not valid for this case. Note that this will not affect results for the common case when the background has the same exposure time as the source aperture, but users do fit background models (rather than subtracting the data) should check their fit results. This fixes #629.

The sherpa.astro.background.BackgroundSumModel class has been removed as it is no-longer used (this is a low-level class that is not used by the sherpa.astro.ui layer).

# Details

This follows from #884 and the first 13 commits are from #884.

There are a number of commits that either re-work the tests, add new tests, or change existing tests to cover various features that we will want to either check aren't changed, or verify that we do see changes, in the later commits.

For the source aperture, fitting a background model will result (conceptually) in a model expression (for the scalar version of the scaling value) of

    rsp(texp_src * (model_src + sum_i (bgnd_scale_i * model_bkj_i))

Although the Sherpa model language can describe this expression (expanding out the summation), the existing code used a special model class (`BackgroundSumModel`) to do this. I don't see the benefit for adding this extra class, so the code now just creates the expression directly. There was a comment in the `BackgroundSumModel` class about a potential problem with the way it was implemented, regarding what parameter values were going to be used to evaluate the models, so this is now avoided (although I don't believe it was actually a problem, but I didn't look into it closely). Removing the class also means that the correct model expression is used: the string representation of BackgroundSumModel could actually display the wrong value, for complex situations, and I know people have tried to recreate these expressions as they are included in the ASCII save files, based on helpdesk tickets). See the examples section.

[...]

PR #900, opened 2020-08-20
Add basic tests of Un/BinOp model combination.

DougBurke:


DougBurke:
That diff coverage is low. I'll take a look.

DougBurke:
I don't completely believe the codecov results (e.g. see the discussion in #903), as I strongly believe that we are testing things it claims we are not (e.g. there is an explicit test of TableModel.ndim even if codecov [really, the coverage tool]  claims otherwise).

DougBurke:
Err, @dtnguyen2 that change isn't in my PR - https://github.com/sherpa/sherpa/pull/900/files doesn't have a self.val change. Are you perhaps coombining this with #888 (which does have a change a bit like this but it has just moved the ShepaFloat to an earlier part of the routine)?

Issue #672, opened 2019-08-30
ylog setting of model plot prefs not honored for PHA data sets


DougBurke:
I haven't actually checked the chips backend to see if this is an issue with the PHA plot classes the backend code, or somewhere in the UI code (the last seems unlikely given the behavior below) but the `yscale` setting of the model plot is not being honored for PHA data sets but it is for other data:

```
sherpa In [1]: from matplotlib import pyplot as plt

sherpa In [2]: get_model_plot_prefs()['ylog']
Out[2]: False

sherpa In [3]: load_arrays(1, [1, 2, 3], [4, 3, 2])

sherpa In [4]: load_arrays(1, [1, 2, 3], [400, 300, 200])

sherpa In [5]: set_source(polynom1d.mdl)

sherpa In [6]: mdl.c0 = 10

sherpa In [7]: mdl.c1 = 100

sherpa In [8]: plot_model()

sherpa In [9]: %matplotlib inline
UnknownBackend: No event loop integration for 'inline'. Supported event loops are: qt, qt4, qt5, gtk, gtk2, gtk3, tk, wx, pyglet, glut, osx

sherpa In [10]: %matplotlib
Using matplotlib backend: TkAgg

sherpa In [11]: plot_model()

sherpa In [12]: plt.gca().get_yscale()
Out[12]: 'linear'

sherpa In [13]: get_model_plot_prefs()['ylog'] = True

sherpa In [14]: plot_model()

sherpa In [15]: plt.gca().get_yscale()
Out[15]: 'log'

sherpa In [16]: load_pha('sherpa-test-data/sherpatest/3c273.pi')
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273.pi' 
but not used; to use them, re-read with use_errors=True
read ARF file sherpa-test-data/sherpatest/3c273.arf
read RMF file sherpa-test-data/sherpatest/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273_bg.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273_bg.pi' 
but not used; to use them, re-read with use_errors=True
read background file sherpa-test-data/sherpatest/3c273_bg.pi

sherpa In [17]: set_source(powlaw1d.pl)

sherpa In [18]: plot_model()

sherpa In [19]: plt.gca().get_yscale()
[...]

PR #906, opened 2020-08-21
Improve and add support for histogram plots

### LINKED ISSUE #106: axis differences between plot_source() and plot('source') and issues with model plots

### LINKED ISSUE #908: Minor doc error: get_source_plot (and probably others) examples
The `get_source_plot` example (and probably others) contains the following example. Unfortunatelly `splot1` and `splot2` refer to the same object (in this case the id='core' dataset)so the plot doesn't do what it's described.

```
        >>> splot1 = get_source_plot(id='jet')
        >>> splot2 = get_source_plot(id='core')
        >>> splot1.plot()
        >>> splot2.overplot()
```

DougBurke:
# Summary

Improvements to the display of histogram-style plots, in particular for 1D integrated datasets and some model display for PHA data. The histograms now cover the full length of each bin (previously they only showed half the bin for the first and last bins), and gaps in the histogram (where the high edge of a bin is less than the lower edge of the next bin) are now correctly displayed.

The get_xxx_plot and get_xxx_contour routines now accept an optional recalc argument. If False (the default is True) then the current plot object is returned, rather than updating the object. The get_data_plot_prefs and get_model_plot_prefs now take an optional id argument, since the settings depend on the data being displayed.

The plot and contour functions now accept keyword arguments, which are sent to each plot or contour. The commands now can contain multiple arguments, which is primarily useful for plotting background datasets since you can now say `plot("bkg", 1, 1, "bkg", 1, 2)` to plot the first two background components of dataset 1.

When using the matplotlib plotting backend, using the linestyle plot preference will now cause a screen message to be output since it is not used.

Fixes #106, #672

# Details

You can view the [proposed developer documentation](https://sherpa--906.org.readthedocs.build/en/906/developer/index.html#plotting-data-using-the-ui-layer).

The following changes are not necessarily in chronological order:

*) Added a number of tests to check the functionality we are going to change.

I've tried my best to cover all code paths I've added or changed. The parts that are still untested break down into

 - handling of "unlikely" errors, such as a check on an internal routine or the plot code failing. The latter
   should be tested, but that's outside the scope of this PR (ie it is an existing hole ino ur testing
 - energy and photon flux plot tests which are in #803 
 - there's something odd going on with the template component code, as it looks like the check is
   for the wrong instance ot the template-family of classes; this is existing code which needs to be
   looked at outside of this PR
 - there's some PSF and kernel contour code which I haven't added tests for (note, this code already existed
   and all I've done is change things, so this is not new code)

*) remove linecolor attribute from the matplotlib code

The linecolor preference setting in the matplotlib code isn't actually used. The code now warns if a user sets this value, but doesn't remove the attribute as this would break plots that work (but doesn't actually do anything with the `linecolor` settings).

*) matplotlib plotting of histograms

The `histo` low-level plot function in `pylab_backend.py` now explicity creates a stepped version of the data, so if the bins are xlo_i,xhi_i,y_i then it connects up xlo_i,y_i to xhi_i,y_i and then to xlo_i+1,y_i+1. The old version would (with the correct drawstyle argument) draw "flat" bins but it would not cover the whole bins for the first and last bin (and it depends on the drawstyle option).

Whereas the code used to create a single object (axes.errorbar or axes.plot depending on the presence of the error-bar requirement, we now have a separate call for the lines and the error bars.

An improvement to the old code is when bins are not contiguous - the old code would draw a line between the bins, but we now have no line.

*) drawing a histogram model for Data1DInt data

There's two parts: creating new sherpa.plot classes (`ModelHistogramPlot` and `SourceHistogramPlot`) which create and display the data in `xlo,xhi,y` format, and changes to the model-plotting infrastructure to use this new class rather than the default model class (`sherpa.plot.ModelPlot`) when creating plots.

Unfortunately it's a bit messy as we have to deal with model and source plots, and the component variants of these too. These new classes for all these, even we could just use ModelHistogramPlot for most of them [not for the component cases], as this appears to be the way the plot code is set up.

*) drawing a histogram model for DataPHA data

`plot_model` for `DataPHA` data draws a histogram, but when used in `plot_fit` - and also `plot_source` (need to check) - a line (not histogram) was used. I hope the DataPHA changes are easier to read than the Data1DInt changes, as they were made after changing the plotting code described below, but I don't think it's particularly obvious.

We now properly see ignored data in `plot_fit` (by gaps in the data).

*) cleaning up the plotting code

When adding the Data1DInt code I realised there was a lot of cut-n-paste code from sherpa.ui.utils in the sherpa.astro.ui.utils module. I started to change this so that the astro version only deals with astro-sepcific concerns (i.e. mainly DataPHA), relying on the superclass for other cases. So, as an example, get_source_plot is now "if the dataset is a DataPHA instance then do something, otherwise let the sherpa.ui.utils.get_source_plot deal with it". However, in doing this I found that I couldn't easily rework the _prepare_plotobj class, which is integral to the existing code (it determines how the prepare method of the plot object is called), to reduce the duplication, as the logic wasn't clear for determining when a plot type would be  processed (comparing the two _prepare_plotobj structures). I've long found this code to be hard to reason about, and it seemed to be hard to reason about, so I have removed this central location. Now it is the get_xxx_plot call that determines calling the prepare method of  its plot class. So now the get_xxx_lot method determines what plot classes it has to deal with (generally only one, but with DataPHA and Data1DInt cases there are some which support multiple classes).

Moving the prepare logic to get_xxx_plot means that the plot_xxx call now "just" has to call the get_xxx_plot method, and then plot the resulting object. This has simplified the _plot method, which used to have to call _prepare_plotobj; it now just has to deal with plotting things. Note that unfortunately nothing in life is ever this simple, so there is still some extra logic in the more-complicated plot_xxx calls, but this was true even in the old version of the code.

The get_xxx_plot calls have gained a recalc argument (default True), which determines whether the prepare method is called. This was added as a) it is useful, b) it simplifies the plot_xxx calls (as they don't have to repeat logic in the get_xxx_plot calls).

The get_data_plot_prefs and get_model_plot_prefs routines now accept an optional id argument since a model plot may use a histogram or plot preferences, depending on the data being plotted. This is issue #672.

The data() call has seen some internal tweaks and now seems to be behaving sensibly (fixing #106), due to the new approach. It has also been updated to allow keyword arguments to be sent to each plot - e.g. `plot('fit', 'm', 'fit', 'p'.ylog=True)` - and to support multiple arguments for a plot type, which lets multiple background components be displayed with a call like `plot('bkg", 1, 1, 'bkg', 1, 2)`, which shows bkg_id=1 and bkg_id=2

The contour plots were also changed to follow the same scheme, although it is simpler as there's a lot less plot types, and essentially no data-specific classes to worry about.

## Things for the future

I feel that DataPHA and Data1DInt data plots should use a histogram. This is #931 which is six commits on top of this PR (and 2 are clean-up commits).

The image could should be changed to use the same scheme but I decided against including this in this PR as it is a completely separate set of changes.

I have been tempted to add a copy=False argument to the get_xxx_plot and get_xxx_contour routines,whih if set means that a copy of the plot object is returned. This would fix #908, but it does not need to be included in this PR.

# Examples

## For a PHA dataset:

```
logger.setLevel(logging.ERROR)
ui.load_pha('pha', '/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest/3c273.pi')
logger.setLevel(logging.INFO)

ui.notice_id('pha', 0.5, 7)
ui.ignore_id('pha', 2, 3)

ui.set_source('pha', ui.xswabs.mabs * ui.powlaw1d.mpl)
mabs.nh = 0.05
mpl.gamma = 1.98
mpl.ampl = 1.93e-4

ui.plot_fit('pha', xlog=True, ylog=True)
```

### This PR

Look, it is a histogram AND has a gap

pha_fit

### Current plot

pha_fit

## data(..) for `DataPHA`

plot only  takes keyword arguments in this PR hence the try/except here:

```
try:
    ui.plot('source', 'pha', 'model', 'pha', 'fit', 'pha', xlog=True, ylog=True)
except TypeError:
    ui.set_xlog()
    ui.set_ylog()
    ui.plot('source', 'pha', 'model', 'pha', 'fit', 'pha')
```

Note that the source and model plots (top left and right) appear different [and fit just shows the same differences we've seen]. This is issue #106

### This PR

pha_multiplot

### Current Sherpa

pha_multiplot

## For a `Data1DInt` dataset :

```
xlo = np.asarray([10, 12, 16, 20, 22])
xhi = np.asarray([12, 16, 19, 22, 26])
yedges = np.asarray([50, 54, 58, 60, 64])

ui.load_arrays('1dint', xlo, xhi, yedges, ui.Data1DInt)

ui.set_source('1dint', ui.polynom1d.m1dint)
m1dint.c0 = 0
m1dint.c1 = 1

ui.plot_fit('1dint')
```

### This PR

Lool, we now have a gap which wasn't at all obvious in the old (aka "current") version:

FIG

### Current plot

FIG

## playing around with a model plot (`Data1DInt`)

```
ui.plot_model('1dint', color='brown', marker='*', markersize=12, markerfacecolor='yellow', linestyle='--')
```

### This PR

1dint_model

### Current

1dint_model

## plot_pdf

```
>>> ui.plot_pdf([10, 20, 20], bins=3)
```

### This PR

The bins now start at 10 and end at 20, not at 11 and 18 (ish).

AFTER

### Current plot

BEFORE


DougBurke:
Here's a (in needs of work to make it usable) example I've got to improve plot_fit for PHA data:

 - a "proper" histogram, showing the widths of the bins
 - skipping ignored bins 

FIG

DougBurke:
@anetasie and, for when yuo get back @hamogu , this needs some testing from SDS

anetasie:
@DougBurke I have not tried it yet, but the example plots look nice! I was wondering about plotting models on high-res grid, when we use the `regrid` function. Have you looked at this option?

DougBurke:
@anetasie - no. It's a big enough PR as is, and the regrid stuff is unrelated to the code changes made here.

anetasie:
@DougBurke - is there a way to get the model plotted on the full grid, including the ignore data points? It is sometimes good to be able to inspect the model behavior over the ignored data space.

anetasie:
I guess I can `notice()` and use `overplot=True` to get the model plotted over the full dataspace.

DougBurke:
@anetasie - I strongly believe that is a different PR. It is not functionality we really have at the moment and can be looked at after this PR is approved.

It actually is related to #922, where the current plot_model command kind of works as you are mentioning, but that is only because the code is wrong. The plot_model command should also be ignoring the data range, but until #922 is fixed it is showing a random set of data which normally shows the full data range.

<rant>

My belief is that if you want to look at the model with a different notice/filter/group/... setting then I believe you should be setting another dataset to use that setting. The current approach to adding random bits of functionality to plot_fit versus plot_model is really confusing from the viewpoint of the code, and is also not well described to the user. If we want to make it easier to generate model/source plots with different characteristics then we should think about it as a whole, rather than the current approach to making plot_model work differently to plot_fit and supporting lo/hi arguments but only for these 3 specific plot types and not other ones. This would be a post 4.13 item.

hamogu:
I agree with @DougBurke

anetasie:
I have the unevenly space data and there is no ignore/notice applied. When I generate a plot_fit() the histogram does not connect between the data points. I understand that my `xlo,xhi` data space has an empty space. Is this the intended behavior? 
Figure_1
The plot looks a bit strange.

```
from sherpa.astro.ui import *
import matplotlib
%matplotlib

xlo = [2,3,4,6,8]
xhi = [3,4,5,7,9]
y = [11,13,20,18,15]
yerr = [1.,1.,1.,1.,1.]
load_arrays(1,xlo,xhi,y, yerr,Data1DInt)
plot_data()
set_model(gauss1d.g1)
fit()
plot_fit()
```

anetasie (approved):

DougBurke:
Yes, this is the expected behavior. There is no nice way to draw a line between the bins with no data. You can try the same  plot with the master branch but then you should carefully check the edges of the bins; I don't think they'll look correct.

We could use a system like the histogram, as shown below, but then we get a line going down to 0 *for each bin*, which looks terrible for PHA datasets. I *couild* add in extra <edge,0> values so we get a line going up from 0, but I'm not sure how they'd lool with a logged y axis (and this can be done as a separate PR).

```
In [21]: plt.clf()

In [22]: yy = np.asarray([1, 2, 2, 4, 4, 4, 4, 6, 6, 7])

In [23]: plt.hist(yy, bins=6, edgecolor='k')
Out[23]:
(array([1., 2., 0., 4., 0., 3.]),
 array([1., 2., 3., 4., 5., 6., 7.]),
 <a list of 6 Patch objects>)
```

FIG

DougBurke:
For reference, the current  behavior for https://github.com/sherpa/sherpa/pull/906#issuecomment-687247927 is

FIG

And I (very biased) think that the new behavior is better.

We can (with the current code) make these seem like histograms, but I will argue until the cows come home, and then more, that the following are wrong:

```
In [5]: plot_fit(drawstyle='steps')
```

FIG

```
In [6]: plot_fit(drawstyle='steps-mid')
```

FIG

hamogu:
I agree with @DougBurke that the new behaviour is better.

anetasie:
Yes, I do agree this is a better behavior.

DougBurke:
Rebased to match #931.

DougBurke:
Another day, another rebase (this time for #931)

DougBurke:
Rebased to fix conflict.

DougBurke:
Rebased because of conflicts.

DougBurke:
No idea why the docs build failed here - it appears to be an internal RTD issue and not due to this PR.

DougBurke:
Rebased as #803 has landed.

DougBurke:
Rebased to fix an internal error in a test due to #884 having been merged (so that test no-longer failed). In doing so I combined all the multiple test commits into a single commit at the start.

dtnguyen2 (approved):

PR #911, opened 2020-08-25
Docstring: changes for ahelp

DougBurke:
These are formatting changes to the docstrings for several routines
that are useful for SDS in generating ahelp files. There is no
change in content, it is just moved around slightly.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -7357,8 +7357,6 @@ class Session(sherpa.ui.utils.Session):
         they are interpreted as the `id` and `val` parameters,
         respectively.
 
-        Notes
-        -----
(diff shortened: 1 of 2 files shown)


DougBurke:
I'm not going to ask for a review just yet, as there may be some more changes coming soon.

wmclaugh:
@DougBurke - As a general indication of when things are ready to be reviewed, can we  (a) wait until they are ready to be reviewed before adding the 'easy to review' indicator   OR.  even better would be to (b) mark them as draft/WIP.

DougBurke:
@anetasie - these are just moving information around in the docstring, trying to keep a consistent ordering (this is only really useful for me, working on the ahelp updates, as I don't think anyone else cares that the reference section is before or after the examples...).

PR #918, opened 2020-08-29
fix an issue with PHA filtering that affects plot_model

### LINKED ISSUE #917: DataPHA.get_filter(group=False) seems odd
This may come down to the definition of the `group` parameter to `get_filter` as it is undocumented.

```
>>> ui.load_pha('sherpa-test-data/sherpatest/3c273.pi')
>>> d = ui.get_data()
>>> d.get_filter()
'0.124829999695:12.410000324249'
>>> d.get_filter(group=False)
'0.008030000201:14.943099975586'
```

This is a grouped dataset, so maybe there's a reason the `group=False` returns a larger range than the default, but there's been no filter applied here. so it's not clear (this could point to issues with `group=True` instead?).

**NOTE** `get_filter` notes that the bin locations reference the center of the bin, which explains the difference above. Whether that's a good idea is a different question...

If we add a simple filter, we again see that the `group=False` range is larger, why is this?

```
>>> d.notice(0.5, 7)
>>> d.get_filter()
'0.518300011754:8.219800233841'
>>> d.get_filter(group=False)
'0.474500000477:9.862299919128'
```

And I now ignore the 2-3 keV range, and things get odd for `group=False`:

```
>>> d.ignore(2, 3)
>>> d.get_filter()
'0.518300011754:1.919900000095,3.233900070190:8.219800233841'
>>> d.get_filter(group=False)
'0.474500000477:0.781100004911,0.795699983835:9.862299919128'
```

Why is the grouping 0.47:0.78 and 0.79:9.8 ?

**EDITED TO ADD** The reason for the last question is "it's a bug", and the fix is in #918

DougBurke:
# Summary

Fix an issue when applying filters to generate the plot_model and plot_model_component plots for PHA datasets.

Fixes #917

# Details

First of all I add a bunch of tests that I came up with trying to diagnose the problem.

The sherpa.utils.parse_expr and sherpa.utils.expr.create_expr now have docstrings and have seen some clean up. The create_expr changes are more extensive, but I claim is simpler than before, and also fixes a logical issue (that doesn't really have any downstream consequences). Both routines now error out for invalid cases (the parse_expr case had code claiming it raised an error but thanks to the logic of the code the error could never be raised), and the create_expr error condition is actually hit in a number of tests.

Investigating sherpa.astro.data.DataPHA.get_filter() showed (eventually) that the case was when called with group=False. The data to be passed to create_expr is supposed to be ungrouped, and the values (the energy/wavelength/channel grid) was, but the mask was taken from self.mask. The trouble is that self.mask is grouped! The solution is to use self.get_mask() in this case. I have left in a safety check to warn users if mask.sum() does not equal the number of data values, but I do not make this an error because we know there are grouping/filtering issues.

The fix for plot_model/plot_model_component comes "for free" - i.e. it doesn't require a change to the plotting code. I mention this as the driver for the fix since it seems to be the only way to trigger this issue (and is a user-visible change of this PR), even if the actual code changes are all very-low level.

Note that the fix commit - Fix subtle create_expr/DataPHA.get_filter issue - only has to change a small number of tests.

# Example

Using the code

```
from sherpa.astro import ui
ui.load_pha('/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest/3c273.pi')
ui.set_source(ui.powlaw1d.pl)

ui.notice(0.5, 7)
ui.ignore(2, 3)

d = ui.get_data()

print("# filter(group=True)")
print(d.get_filter(group=True, format='%.4f'))

print("# filter(group=False)")
print(d.get_filter(group=False, format='%.4f'))

ui.plot_model()
```

## Current master branch

The screen output is

```
# filter(group=True)
0.5183:1.9199,3.2339:8.2198
# filter(group=False)
0.4745:0.7811,0.7957:9.8623
```

Note that the output of get_filter(group=False) doesn't make any sense: it is not roughly 0.5:2,3:7.

The plot is

filtered_model

or, if we switch from line to symbols to make it more-obvious where the points are [unfortunately I created this with ifferent model parameters, so the shape is different, but the point is that we see points i the 2-3 keV range]:

```
ui.plot_model(linestyle='', marker='s')
```

old

## This PR

The screen output now provides sensible information for the filter(group=False) call:

```
# filter(group=True)
0.5183:1.9199,3.2339:8.2198
# filter(group=False)
0.4745:1.9345,3.0879:9.8623
Created: filtered_model.png
```

and the model plot now excludes the ignored range (2-3 keV):

filtered_model

or, if we switch from line to symbols to make it more-obvious where the points are [the model parameters are different, which is why the plot looks different, but it's the same as the above points-only plot]:

```
ui.plot_model(linestyle='', marker='s')
```

new

I've created the following plot by rebasing this PR onto #906 as it changes how the model plot is shown, in particular it better handles ignored bin ranges, and you get (with `ui.plot_model`):

filtered_model

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1452,20 +1452,23 @@ def parse_expr(expr):
     [(None, None)]
 
     """
-    res = []
+
     if expr is None or str(expr).strip() == '':
-        res.append((None, None))
-        return res
+        return [(None, None)]
+
+    res = []
     vals = str(expr).strip().split(',')
     for val in vals:
         lo, hi = None, None
+
         interval = val.strip().split(':')
-        if len(interval) == 1:
+        ninterval = len(interval)
+        if ninterval == 1:
             lo = interval[0]
             if lo == '':
                 lo = None
             hi = lo
-        elif len(interval) > 1:
+        elif ninterval == 2:
             lo = interval[0]
             hi = interval[1]
             if lo == '':
@@ -1473,7 +1476,14 @@ def parse_expr(expr):
             if hi == '':
                 hi = None
         else:
+            # This check exited but was never hit due to the way the
+            # code was written. It now errors out if a used gives
+            # a:b:c, whereas the old version would have just ignored
+            # the ':c' part. Perhaps we should just keep dropping
+            # it, in case there's existing code that assumes this.
+            #
             raise TypeError("interval syntax requires a tuple, 'lo:hi'")
+
         if lo is not None:
             try:
                 lo = float(lo)
@@ -1484,7 +1494,9 @@ def parse_expr(expr):
                 hi = float(hi)
             except ValueError:
                 raise TypeError("Invalid upper bound '%s'" % str(hi))
+
         res.append((lo, hi))
+
     return res
 
 
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1523,10 +1523,10 @@ def parse_expr(expr):
                 hi = None
         else:
             # This check exited but was never hit due to the way the
-            # code was written. It now errors out if a used gives
+            # code was written. It now errors out if a user gives
             # a:b:c, whereas the old version would have just ignored
             # the ':c' part. Perhaps we should just keep dropping
-            # it, in case there's existing code that assumes this.
+            # it, in case there's existing code that assumes this?
             #
             raise TypeError("interval syntax requires a tuple, 'lo:hi'")
 
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1350,6 +1350,13 @@ def create_expr(vals, mask=None, format='%s', delim='-'):
     delim : str, optional
         The separator for a range.
 
+    Raises
+    ------
+    ValueError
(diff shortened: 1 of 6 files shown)


anetasie:
This looks good!

dtnguyen2 (approved):

sherpa/utils/__init__.py
```diff
@@ -1327,71 +1327,209 @@ def print_fields(names, vals, converters=None):
...
+        elif ninterval == 2:
             lo = interval[0]
             hi = interval[1]
             if lo == '':
                 lo = None
             if hi == '':
                 hi = None
         else:
+            # This check exited but was never hit due to the way the
+            # code was written. It now errors out if a used gives
```
wmclaugh, line 1526: 'used' or 'user'

DougBurke: It should be 'user'.

DougBurke: I've just pushed a commit to fix the typo. This comment is probably too defensive but I just wanted to point out that there is a subtle change in behavior here. I note that the "user-level" documentation - e.g. sherpa.astro.ui.notice - only ever mentions "a", "a:b", "a:", or ":b" for these fragments (so we shouldn't expect anything more).

PR #924, opened 2020-09-02
Updated to ignore the .vscode directory

anetasie:
I'm using VSCode and needed to ignore the local .vscode configuration directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,6 @@ docs/**/api
 
 #generated from python setup.py test
 .pytest_cache
+
+# IDE config
+.vscode


DougBurke:
I've asked @Marie-Terrell to review this as it's a bit more "code setup-ery" rather than actual code.

Marie-Terrell (approved):
I see no problem with this change. Though, I am curious why those checks failed.

DougBurke:
@Marie-Terrell - the document build failed because of #925 which is a "not our problem" issue.

The change in coverage is just part of the "ransomness" of our existing tests which we are trying to improve (I forget the link to a recent tissue about this)

wmclaugh:
Is there any reason why this should be done in the repo's .gitignore file vs your local ~/.gitignore_global?  Not advocating one or the other but would like to understand the implications this change may impose  on others.

anetasie:
@wmclaugh I don't know anything about `.gitignore_global`. I don't have such a file in my local directories and I'm not quite sure how to set this up. Would `.gitignore_global` work  for all the updates performed via VSCode in my local sherpa repository? `.gitignore` file is in the repository.

DougBurke:
So I think the issue is convenience. At what point do we switch from adding things to the system .gitignore to tell users that they need to set there own. I just checked and astropy have .vscode in their .gitignore (https://github.com/astropy/astropy/blob/master/.gitignore). I ssume they have more vscode users than we do!

wmclaugh:
@anetasie- Please see https://docs.github.com/en/github/using-git/ignoring-files for documentation describing set up of a global gitignore file.

anetasie:
Yes, this is the convenience for me. I work with three different computers :-)

PR #928, opened 2020-09-03
Tests: change test_astro_sim.py::test_pragbayes_pcaarf_limits

DougBurke:
# Summary

Use a known seed when testing a MCMC run, introduced in #893, to avoid occasional failures.

# Details

The test_pragbayes_pcaarf_limits test from #893 uses a random seed to
run a MCMC chain which has been engineered to trigger a number
of "problem" cases (e.g. values exceeding the parameter limits).

Unfortunately this test occasionally fails, so I've selected a
random seed to run this test with which works on my machine.

A new fixture has been added - reset_seed - which can be used for
tests for which we want to force a random seed. Note that it
resets the seed to a "random" value after the test as I want to
make sure we only force it when required.

### DIFF
--- a/sherpa/astro/sim/tests/test_astro_sim.py
+++ b/sherpa/astro/sim/tests/test_astro_sim.py
@@ -159,11 +159,21 @@ def test_pragbayes_pcaarf(sampler, setup):
 @requires_data
 @requires_fits
 @pytest.mark.parametrize("sampler", ["pragBayes", "fullbayes"])
-def test_pragbayes_pcaarf_limits(sampler, setup, caplog):
-    """Try and trigger limit issues"""
+def test_pragbayes_pcaarf_limits(sampler, setup, caplog, reset_seed):
+    """Try and trigger limit issues.
+
+    """
 
     from sherpa.astro.xspec import XSAdditiveModel, XSMultiplicativeModel, \
         XSwabs, XSpowerlaw
+
+    # Set the seed for the RNG. The seed was adjusted to try and make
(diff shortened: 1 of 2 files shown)


DougBurke:
The initial version of this patch had a fancier fixture, which set the seed for the test and then reset it after the test. This was over-engineered, so now the fixture just ensures the seed is reset to a "random" value after the test.

dtnguyen2 (approved):
Once I got xspec properly installed, which is not exactly an easy process, the review is straight forward as @DougBurke said. I even commented out the ```reset_seed``` to make sure it was called correctly:

```
__________ ERROR at setup of test_pragbayes_pcaarf_limits[pragBayes] ___________
file /export/sherpa/sherpa/astro/sim/tests/test_astro_sim.py, line 158
  @requires_xspec
  @requires_data
  @requires_fits
  @pytest.mark.parametrize("sampler", ["pragBayes", "fullbayes"])
  def test_pragbayes_pcaarf_limits(sampler, setup, caplog, reset_seed):
E       fixture 'reset_seed' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capture_all_warnings, clean_astro_ui, clean_ui, doctest_namespace, make_data_path, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, restore_xspec_settings, run_thread, setup, test_data_path, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.
```

Issue #571, opened 2018-12-06
inconsistent warning for axis limits when using log scale (set_xlog/ylog) for plots with matplotlib


DougBurke:
This is not a lien on the 4.11 release.

With the CIAO 4.11 code base (so close to `master`) and matplotlib version 2.2.3, I see the following behavior, where some, but not all calls to plot data/model/residuals/... leads to the following warning after calling `set_xlog`:

```
UserWarning: Attempted to set non-positive xlimits for log-scale axis; invalid limits will be ignored.
  'Attempted to set non-positive xlimits for log-scale axis; '
```

Note that the presence of the warnings is inconsistent (ie I would expect it to happen all times or no times) but is consistent (in that it happens after a particular sequence of plot calls). My guess is that we are not clearing up the state from the previous plot correctly, but I haven't looked into the code.

Also note that the warning doesn't really make sense, since the data is > 0 for both axes.

```
% ipython --matplotlib
Python 3.5.4 (default, Sep 14 2018, 15:42:52) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.
Using matplotlib backend: TkAgg

In [1]: from sherpa.astro import ui

In [2]: ui.load_arrays(1, [0.2, 2, 20], [200, 20, 0.2])

In [3]: ui.set_source(ui.const1d.mdl)

In [4]: ui.plot_fit()

In [5]: ui.plot_fit_resid()

In [6]: ui.plot_fit()

In [7]: ui.plot_fit_resid()

In [8]: set_xlog()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-8-1dd939fa3894> in <module>()
----> 1 set_xlog()

NameError: name 'set_xlog' is not defined

In [9]: ui.set_xlog()

In [10]: ui.plot_fit()

In [11]: ui.plot_fit_resid()

In [12]: ui.plot_fit()
/home/djburke/testlocal/ciaot/ciao-4.11/ots/lib/python3.5/site-packages/matplotlib/axes/_base.py:3157: UserWarning: Attempted to set non-positive xlimits for log-scale axis; invalid limits will be ignored.
  'Attempted to set non-positive xlimits for log-scale axis; '

In [13]: ui.plot_fit()

In [14]: ui.plot_fit()

In [15]: ui.plot_fit_resid()

In [16]: ui.plot_fit_resid()
[...]

PR #929, opened 2020-09-04
Support overplot option in plot_fit_xxx (issue #700)

### LINKED ISSUE #700: (overplot=True) with double plots
The option `overplot=True` in the plots which generate one display window allow to overplot different data sets in the same plot window. For example 

      plot_data(1) 
      plot_data(2,overplot=True)

will generate a plot showing two data sets in the same plot. However,
`overplot=True` is  not generating the plots for two data sets in the case of double plots.
For example

    plot_fit_resid(1)
    plot_fit_resid(2,overplot=True)'

will plot the first data set, but then regenerate the second plot after deleting the first one.
There is no Error mgs, so in theory this is allowed, but the current behaviour is consistent with the
4.11 and older versions of Sherpa.

DougBurke:
# Summary

The overplot argument can now be used with the plot_fit_xxx and plot_bkg_fit_xxx routines (e.g. plot_fit_ratio).

Fix #700 

# Details

This allows you to say

    plot_fit_resid(ylog=True)
    plot_fit_resid(2, overplot=True)

The underlying plot objects are poorly documented, so it's not obvious what they were meant to be doing. I have taken the liberty to change things around a little bit to support the overplot setting. Tests have been added to show this is working, and fortunately it's the only use case for JointPlot in the ui code.

The tests trigger #571 so I've added a suppressor to the test code. I  would prefer this to be local to the test, rather than for the whole test suite, but that is a larger change.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -120,6 +120,12 @@ known_warnings = {
             # Matplotlib version 2 warnings (from HTML notebook represention)
             #
             r'Attempting to set identical bottom==top results\nin singular transformations; automatically expanding.\nbottom=1.0, top=1.0',
+
+            # See issue #571 and tests
+            # test_ui_plot.py
+            #    test_plot_fit_xxx_pylab
+            #    test_plot_fiot_xxx_overplot_pylab
+            r"Attempted to set non-positive left xlim on a log-scaled axis.*"
         ],
     RuntimeWarning:
         [r"invalid value encountered in sqrt",
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -124,8 +124,9 @@ known_warnings = {
             # See issue #571 and tests
             # test_ui_plot.py
             #    test_plot_fit_xxx_pylab
-            #    test_plot_fiot_xxx_overplot_pylab
-            r"Attempted to set non-positive left xlim on a log-scaled axis.*"
+            #    test_plot_fit_xxx_overplot_pylab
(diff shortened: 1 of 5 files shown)


DougBurke:
@anetasie - does this work like you'd want for #700 ?

anetasie (approved):

anetasie:
@DougBurke Yes, it looks good and it does what I wanted for #700. Thanks!

anetasie:
@wmclaugh We want this for 4.12.2

DougBurke:
Rebased because of #906 being merged

DougBurke:
I've added the fix from #955 to get the tests to pass.

DougBurke:
Rebased now #955 has been merged

dtnguyen2 (approved):

wmclaugh:
@DougBurke - Can you please rebase so this can get merged.

DougBurke:
@wmclaugh - well, that was a surprisingly more-involved rebase than I expected. It definitely needs to pass travis before merging :-)

DougBurke:
**ARGH** stupid rebases

DougBurke:
So, I don't know if that was something that got lost in the rebase or is because of other changes that have been recently merged. It looks like we needed to add a separate warning check for a Matlpotlib 2 warning compared to Matplotlib 3.

DougBurke:
@wmclaugh - the tests now pass. It's marked as failed because the coverage has changed for the worse. However, if you look this is one of those odd cases, as the coverage has decreased in files that haven't been touched in this PR and are not related to the tests I've added.

PR #931, opened 2020-09-04
Update Data1DInt / DataPHA data plots to use the histogram plot style

DougBurke:
# Summary

Switch the plot_data/plot_bkg plots to draw the data as histograms for Data1DInt and DataPHA plots. By default this doesn't change anything significant (just the x-axis range of the plot), but you can tweak things, such as set `linestyle='-'` to see the bin edges.

This will change the behavior of code that accesses the plot data - e.g. get_data_plot() or the dataplot element of get_fit_plot() - since for DataPHA and Data1DInt datasets the data will no-longer have an x attribute but xlo and xhi. To reduce the need for code changes for existing scripts - as many people use get_model_plot and get_data_plot to get the data - the histogram plots will return (xlo+xhi)/2 when asked for the x attribute.

# Details

It builds off of the changes in #906 to draw models of Data1DInt/DataPHA as histograms, and fixes up a corner case left from that work (the handling of the `xerrorbars` setting).

The choice for what to do when you call get_data_plot(id, recalc=False) is interesting, as we don't know what the data type was before hand - e.g. it could have been Data1D although now it's Data1DInt. This is too hard to fix, so I think it's best to just return the plot object of the current "dataset=id" value.

I've added in a fix for the handling of the background id when sherpa.astro.ui.set_default_id is used - it previously changed both the default dataset id and the default id for the background dataset, but unfortunately we don't actually allow the DataPHA object to change that setting, leading to odd behavior. The easiest solution is to just not change the default background id when set_default_id is called.

# Example - 1D Int

```
In [2]: ui.load_arrays(1, [2, 4, 8], [4, 6, 10], [10, 12, 7], ui.Data1DInt)
In [3]: ui.plot_data()
```

FIG

This should be unchanged compared to the master branch (well, actually this covers the full range of the X axis, so 2-10, whereas the master branch just shows the centers of the bins (so 3-9).

If I tweak things I get:

```
In [4]: ui.plot_data(linestyle='-', marker='')
```

FIG

Here's what this plot looks like with the current code (you can see the restricted range of the axis):

FIG

You can try to make it "more histogramey", but I still don't like it (note this is with the current master branch, not this PR):

```
In [4]: ui.plot_data(linestyle='-', marker='', drawstyle='steps-mid')
```
FIG

# Example - PHA

```
>>> ui.load_pha('sherpa-test-data/sherpatest/3c273.pi')
>>> ui.notice(0.5, 7)
>>> ui.ignore(2.5, 3)
>>> ui.set_source(ui.xsphabs.gal * ui.xspowerlaw.pl)
>>> gal.nh = 0.05; pl.phoindex = 1.98; pl.norm = 1.93e-4
>>> ui.plot_data()
```

This doesn't look much different (although the x-axis range is slightly larger as it covers xlo/first bin to xhi/last bin, rather than (xlo+xhi)/2.

FIG

```
>>> ui.plot_data(linestyle='-', marker='', xlog=True)
```

This shows odd the "proper" histogram display now:

FIG

```
>>> ui.plot_data(xerrorbars=True, xlog=True)
```

We still get error bars (although they look very similar to the linestyle='-' version now):

FIG

```
>>> ui.plot_data(xerrorbars=True, xlog=True)
```

Default plot as normal

FIG

A different version

```
>> ui.plot_fit(xlog=True, ylog=True, linestyle='-', marker='')
```

FIG

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -34,7 +34,7 @@ import logging
 
 warning = logging.getLogger(__name__).warning
 
-__all__ = ('SourcePlot', 'ARFPlot', 'BkgDataPlot', 'BkgModelHistogram',
+__all__ = ('DataPHAPlot', 'SourcePlot', 'ARFPlot', 'BkgDataPlot', 'BkgModelHistogram',
            'BkgFitPlot', 'BkgSourcePlot', 'BkgDelchiPlot', 'BkgResidPlot',
            'BkgRatioPlot', 'BkgChisqrPlot',
            'OrderPlot', 'ModelHistogram')
@@ -57,6 +57,40 @@ def to_latex(txt):
     """
 
     return backend.get_latex_for_string(txt)
+
+
+class DataPHAPlot(sherpa.plot.DataHistogramPlot):
+    """Plot a PHA dataset."""
+
+    histo_prefs = sherpa.plot.get_data_hist_prefs()
+
+    def prepare(self, data, stat=None):
+
+        # Need a better way of accessing the binning of the data.
+        # Maybe to_plot should return the lo/hi edges as a pair
+        # here.
+        #
+        (_, self.y, self.yerr, self.xerr, self.xlabel,
+         self.ylabel) = data.to_plot()
+
+        if stat is not None:
+            yerrorbars = self.histo_prefs.get('yerrorbars', True)
+            self.yerr = sherpa.plot.calculate_errors(data, stat, yerrorbars)
+
+        self.title = data.name
+
+        # Get the X axis data.
+        #
+        if data.units != 'channel':
+            elo, ehi = data._get_ebins(group=False)
+        else:
+            elo, ehi = (data.channel, data.channel + 1.)
+
+        self.xlo = data.apply_filter(elo, data._min)
+        self.xhi = data.apply_filter(ehi, data._max)
+        if data.units == 'wavelength':
+            self.xlo = data._hc / self.xlo
+            self.xhi = data._hc / self.xhi
 
 
 class ModelPHAHistogram(HistogramPlot):
@@ -310,10 +344,11 @@ class ARFPlot(HistogramPlot):
                 self.xhi = data._hc / self.xhi
 
 
-class BkgDataPlot(DataPlot):
+class BkgDataPlot(DataPHAPlot):
     "Derived class for creating plots of background counts"
-    def __init__(self):
-        DataPlot.__init__(self)
+
+    # Is this derived class worth it?
+    pass
 
 
 # was BkgModelPlot; this is not a good class name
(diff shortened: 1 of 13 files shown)


DougBurke:
I've asked for @hamogu and @anetasie to try this out to see if I've missed anything. It's actually a bit tricky to test since there are a lot of changes from #906 - this PR is just to check the plot_data plots (but if you want to do a combined check of this and #906 that's okay ;-).

anetasie:
@DougBurke Is the histogram intended as a default plot `plot_data/plot_fit` to replace the current normal plot? or is this only with an addition of `linestyle` argument?

DougBurke:
The idea is that if you don't change anything then you get the same behavior as the existing code. It's only if you add the linestyle argument or change the preferences that you'll get the new behavior.

We could change this - if so I'd suggest linestyle='-', marker='' but you may prefer something else.

DougBurke:
I've just added a note to the summary section to note that this will change code for people with DataPHA or Data1DInt data that use get_data_plot() or get_fit_plot().dataplot to access the plotted data. You will no-longer have a x attribute but xlo and xhi. It's not a big change, but it could break a lot of code.

This also holds for #906, but that is for people accessing the model data.

One idea would be to add  a "fake" x attribute for these model classes that just returns (xlo+xhi) / 2. Ugly, but it should be relatively simple.

anetasie:
I prefer the non-histogram plot of the counts data with the `xerrorbar` set to True. I don't like the histogram style connecting the points.

DougBurke:
@anetasie - just to check, I believe that this is the behavior you get by default from this PR.

anetasie:
@DougBurke I just build this branch and tried `plot_fit(xlog=True, ylog=True, linestyle='-', marker='')` and the plot does not look like the one you have above. I guess I don't have all the other changes in this branch and will not see histograms? 

`plot_data(xerrorbars=True, xlog=True)` looks good and the default plots look good.
Fig_plot_fit

anetasie:
Wait, I'm not sure I've build the code correctly!

anetasie:
ok. I rebuild the code and get the right version. It looks better! I just mixed up my envs.

I like the `xerrorbars=True`.  Should this be a default option?

anetasie (approved):
It looks good!

DougBurke:
I do agree that `xerrorbars=True` for PHA data has a certain appeal. If we do that I would suggest turning off the marker (ie marker='') to make the plot less busy. I think these can be played with after this has been merged and we can come up with a suggestion after playing with it briefly.

anetasie:
@DougBurke Yes, this sounds good.

DougBurke:
Rebased because of conflicts

DougBurke:
Rebased because of changes because of #884 and #906 cleanup (and finally fixed whatever horrors had befouled this commit history from one rebase too many).

anetasie:
@wmclaugh We want this for 4.12.2. SDS has approved it.

DougBurke:
Since #906 has been merged I have rebased this as it removes a lot of commits!

dtnguyen2 (approved):

wmclaugh:
@DougBurke- Can you please rebase this so I can merge it?

DougBurke:
**argh** where did this conflict come from. I'm 223% sure I was looking at this over the week-end to check!

DougBurke:
@wmclaugh - this should pass the travis tests now.

PR #935, opened 2020-09-10
Remove the RTD Travis build

DougBurke:
# Summary

Remove the travis sphinx build as the ReadTheDocs builder on GitHub is more useful.

# Details

We are now building each PR on ReadTheDocs with the docs/readthedocs.org
check on GitHub. This means we do not need a separate sphinx build on
Travis. Another advantage is that the RTD build lets us view the output
which we fon't have on the Travis build.


Marie-Terrell (approved):
I don't see an issue with this.

PR #938, opened 2020-09-13
WIP Improve pylab plots: support label and linewidth options

DougBurke:
This lets you say

    plot_data(label='Data 1')
    plot_data(2, label='Data 2', overplot=True)
    plot_model(label='Model 1', alpha=0.5, linewidth=2, linestyle='--', overplot=True)
    plot_model(2, label='Model 2', alpha=0.5, linewidth=2, linestyle=':', overplot=True)
    plt.legend()

An interesting one, which I think I like, is

    plot_fit(label='xx')
    plt.legend()

since the label is applied to both data and model fits.

I have added an experiment to the ui layer so that plot_data/model/fit/... adds a default label (normally '<label>: <id>' but can be 'Source/Model: <model name>') to try this out. It basically forces the label option for all backends (or makes the code pylab dependent), and does not yet handle all options (the plot_bkg* ones are missing).

Note there are other larger changes to the pylab code in existing PRs so it'd be easier to merge those first. I'm also leaving this open for the time being to see if there's more functionality we want to add.

# Examples

```
ui.plot_fit()
plt.legend()
```

FIG

```
ui.plot_data(xlog=True, alpha=0.6)
ui.plot_model(overplot=True)
plt.legend()
```

FIG

Creating your own label:

```
ui.plot_data(xlog=True, alpha=0.6, label='The mighty Zorb')
ui.plot_model(overplot=True, label='and her sidekick, Pete')
plt.legend()
```

FIG


DougBurke:
@hamogu and @anetasie  - what do you think of this one?

hamogu:
I like it the label part of it, and mostly like the linewidth part of it.

A minor quibble first: In the `plot_fit` example (first plot), both lines are labelled as "fit", while in reality one of them if the observation (not fit, just read in), while the other is a model (probably fit, since I'm called `plot_fit`, but not necessarily, since I can run plot fit any time a model and data are defined, even if I never called `fit`). I think they should be called 'data' and 'model' as in the other examples.

It's natural for matplotlib users (i.e. every scientific python user) to add labels this way and it only matters if you make a legend. Once we add legends to our threads and use reasonably label ids ('1' is not very meaningful as an id. I usually assign string labels like 'ObsID 1234' so this will be immediately useful for me) this will be great.

For the linewidth, I wonder, what's special about linewidth? How about linestyle or marker or markeredgecolor or solid_cap_style or any other matplotlib argument that the underlying plotting routines understand (for the matplotlib backend). `label` is different in that you set a default that actually depends on the data (like `Data: id`). 
I think I've asked this before, but I've never looked at closely enough at the pylab backend (I'll leave inline comments there now): Why can't the methods on `pylab_backend` not just accept `**kwargs` and pass them on to matplotlib? That way, we don't have to list each and every one and we don't have to do any changes if matplotlib changes?

sherpa/plot/pylab_backend.py
```diff
@@ -240,8 +244,10 @@ def plot(x, y, yerr=None, xerr=None, title=None, xlabel=None, ylabel=None,
          barsabove=_errorbar_defaults['barsabove'],
          xlog=False,
          ylog=False,
+         label=None,
          linestyle='solid',
          linecolor=None,
+         linewidth=None,
          drawstyle='default',
```
hamogu, line 251: Why not:
```
def plot(x, y, yerr=None, xerr=None, title=None, xlabel=None, ylabel=None,
         overplot=False, clearwindow=True,
         xerrorbars=False,
         yerrorbars=False,
         ecolor=_errorbar_defaults['ecolor'],
         capsize=_errorbar_defaults['capsize'],
         barsabove=_errorbar_defaults['barsabove'],
         xlog=False,
         ylog=False, **kwargs)
```

sherpa/plot/pylab_backend.py
```diff
@@ -318,7 +324,9 @@ def plot(x, y, yerr=None, xerr=None, title=None, xlabel=None, ylabel=None,
         yerr = yerr if yerrorbars else None
         axes.errorbar(x, y, yerr, xerr,
                       color=color,
+                      label=label,
```
hamogu, line 327: And then here:
```
axes.errorbar(x, y, yerr, xerr, **kwargs)
```

sherpa/plot/pylab_backend.py
```diff
@@ -332,8 +340,10 @@ def plot(x, y, yerr=None, xerr=None, title=None, xlabel=None, ylabel=None,
     else:
         axes.plot(x, y,
                   color=color,
+                  label=label,
                   alpha=alpha,
                   linestyle=linestyle,
+                  linewidth=linewidth,
                   drawstyle=drawstyle,
```
hamogu, line 347: and so on like above with `**kwargs`?

sherpa/plot/pylab_backend.py
DougBurke: It's not that easy. We set up the defaults knowing what the keywords should be. This could be relaxed, but it needs a careful think about.

Having another backend would help identify things we want to keep and those we can change :-)

DougBurke:
As to the plot_fit using label "fit" for both: I explicitly called this out with "since the label is applied to both data and model fits." and it was done this way because [I would need to check] it's not going to be easy to set up separate labels (otherwise I'd have done that in the first place!). Note that

    plot_fit(label='bob')

is going to use 'bob' for both, as we don't have the concept of multiple "attributes" for the plot (since this is a relatively new addition to the API I added). We could think about label=['x', 'y'] or have label= (for both) and labels= (for individual), and also use it for the other options - e.g. color, alpha, ... - but that's a different PR I think.

DougBurke:
Rebased because #906 was merged.

hamogu:
> is going to use 'bob' for both, as we don't have the concept of multiple "attributes" for the plot (since this is a relatively new addition to the API I added). We could think about label=['x', 'y'] or have label= (for both) and labels= (for individual), and also use it for the other options - e.g. color, alpha, ... - but that's a different PR I think.

I agree. I wanted to call it out because I thought is was odd, but I kind of assumed that you did it because everything else would have been hard. I strongly recommend to **not** implement any of the suggestions in the quote, neither in this PR nor in the next one. I smell feature creep. Instead, this is something for the docstring: "If you want different labels of data and model, plot separately with `plot_data` and `plot_model (overplot=True,...)` or use plotting backend directly".

DougBurke:
Note to self: with #906 merged some logic that had already checked model was an actuao model expression is no-longer being used, hence the test failures. Several options beckon...

DougBurke:
This is also in #1382 so hopefully we an close this down soon, but I want to leave it open just for now.

DougBurke:
The main parts of this are in #1382 so I'm closing this.

hamogu:
#1382 does only implement part of this PR. So that we don't forget, I'm reopening this (and making it a draft). In practice, it's probably easier to copy and paste the remaining lines of code into a new PR after #1382 is merged, but I'd hate to loose the thoughts that @DougBurke already put into this.

PR #942, opened 2020-09-18
Tests: support Crates as well as the AstroPy backend

DougBurke:
# Summary

Update a test to pass when run using the Crates I/O backend.

# Details

The Crates and AstroPy I/O backends are not guaranteed to return
exactly the same datatypes, so change a datatype check to be more
generic (checking the type attribute of the dtype to support synonyms).


dtnguyen2 (approved):
Marie and I looked at this problem and the fix here should enable CIAOD/X to pass.  Thanks for the quick response.

Issue #943, opened 2020-09-21
Setting the default id to other than 1 messes up background datasets


DougBurke:
An example of the problem:

```
In [6]: ui.clean()

In [7]: ui.set_default_id('bob')

In [8]: ui.load_pha('sherpa-test-data/sherpatest/9774.pi')
read ARF file sherpa-test-data/sherpatest/9774.arf
read RMF file sherpa-test-data/sherpatest/9774.rmf
read background file sherpa-test-data/sherpatest/9774_bg.pi

In [9]: ui.set_bkg_source(ui.xspowerlaw.bpl)

In [10]: ui.get_bkg_source()
Out[10]: <XSpowerlaw model instance 'xspowerlaw.bpl'>

In [11]: ui.fit_bkg()
---------------------------------------------------------------------------
IdentifierErr                             Traceback (most recent call last)
<ipython-input-11-73c81ec708b7> in <module>
----> 1 ui.fit_bkg()

<string> in fit_bkg(id, *otherids, **kwargs)

~/sherpa/sherpa-pr/sherpa/astro/ui/utils.py in fit_bkg(self, id, *otherids, **kwargs)
  10015         """
  10016         kwargs['bkg_only'] = True
> 10017         self._fit(id, *otherids, **kwargs)
  10018
  10019     def _fit(self, id=None, *otherids, **kwargs):

~/sherpa/sherpa-pr/sherpa/astro/ui/utils.py in _fit(self, id, *otherids, **kwargs)
  10035
  10036         if fit_bkg:
> 10037             ids, f = self._get_bkg_fit(id, otherids, numcores=numcores)
  10038         else:
  10039             ids, f = self._get_fit(id, otherids, numcores=numcores)

~/sherpa/sherpa-pr/sherpa/astro/ui/utils.py in _get_bkg_fit(self, id, otherids, estmethod, numcores)
   9839     def _get_bkg_fit(self, id, otherids=(), estmethod=None, numcores=1):
   9840
-> 9841         fit_to_ids, datasets, models = self._prepare_bkg_fit(id, otherids)
   9842
   9843         # Do not add backgrounds to backgrounds.

~/sherpa/sherpa-pr/sherpa/astro/ui/utils.py in _prepare_bkg_fit(self, id, otherids)
   9833         # If no data sets have models assigned to them, stop now.
   9834         if len(models) < 1:
-> 9835             raise IdentifierErr("nomodels")
   9836
   9837         return fit_to_ids, datasets, models

[...]

PR #944, opened 2020-09-21
Improve handling of the default id with PHA background datasets (fix issue #943)

DougBurke:
# Summary

The `sherpa.astro.ui.set_default_id` call no-longer sets the default identifier for background ids, which are now kept as the value 1. This avoids several issues when using `set_default_id` with the background components of PHA datasets.

# Details

Until now the default background id was set to match the default dataset id, but this is problematic since there is no way to
pass this identifier to the DataPHA class when creating the data objects, and so there is a mis-match between what the sherpa.astro.ui level and the DataPHA object thinks is the default background identifier.

We now have a separate _fix_background_id method in the sherpa.astro.ui Session, to make it explicit what is being checked, and this routine will return the DataPHA.default_background_id setting now.

We could pass an identifier for the default background dataset to the DataPHA clas, but that is left for another time.

You can see some examples from @anetasie in the discussion on this PR, as well as the tests I added.

# Notes

I originally added this as part of #931 but it is a separate issue so I've made it a separate PR.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -1098,7 +1098,8 @@ def fail(x):
                                       fail(ui.plot_bkg_fit_ratio),
                                       fail(ui.plot_bkg_fit_delchi)])
 def test_pha1_bkg_plot(plotfunc, clean_astro_ui, basic_pha1, hide_logging):
-    """These fail because of issue #943"""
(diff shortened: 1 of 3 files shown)


anetasie:
@dtnguyen2 could you review this PR? It is a bug fix that we want in 4.12.2

DougBurke:
Rebased because of recent merge conflicts in the test suite.

dtnguyen2 (approved):

Issue #831, opened 2020-06-21
numpy 1.19.0 is out, does not support python 3.5, breaks some things we use, and adds deprecation


DougBurke:
So, I can't find 1.19 docs easily on the scipy site (e.g. https://numpy.org/doc/stable/ still returns v1.18 for me), but there is https://github.com/numpy/numpy/releases which says

> This NumPy release is marked by the removal of much technical debt:
support for Python 2 has been removed, many deprecations have been
expired, and documentation has been improved. The polishing of the
random module continues apace with bug fixes and better usability from
Cython.
>
> The Python versions supported for this release are 3.6-3.8. Downstream
developers should use Cython >= 0.29.16 for Python 3.8 support and
OpenBLAS >= 3.7 to avoid problems on the Skylake architecture.

# previous notes

numpy 1.19.0 has just made it onto the system we're using, and it loos to break the sphinx doc build on travis.

**EDITED TO ADD** actually, I believe it is more widespread than just the doc build, but also things like `python setup.py develop`

# success from yesterday

https://travis-ci.org/github/sherpa/sherpa/jobs/700406019#L848

uses numpy 1.18.5

# new failure (no significant changes to the docs)

https://travis-ci.org/github/sherpa/sherpa/jobs/700566553#L848

uses numpy 1.19.0, and the error is

```
% $ python setup.py $INSTALL_TYPE
Traceback (most recent call last):
  File "setup.py", line 57, in <module>
    from helpers.extensions import static_ext_modules
  File "/home/travis/build/sherpa/sherpa/helpers/__init__.py", line 25, in <module>
    from .sherpa_config import sherpa_config
  File "/home/travis/build/sherpa/sherpa/helpers/sherpa_config.py", line 25, in <module>
    from numpy.distutils.core import Command
ImportError: cannot import name 'Command' from 'numpy.distutils.core' (/home/travis/miniconda/envs/build/lib/python3.7/site-packages/numpy/distutils/core.py)
The command "python setup.py $INSTALL_TYPE" failed and exited with 1 during .
```

DougBurke:
I have a test build at https://github.com/sherpa/sherpa/pull/830/commits/486f883c8150dd0c34989b603ff10c1011019d89 to work around this (by fixing NumPy to 1.18) until we sort this out properly.

DougBurke:
Some things to address:

a) numpy.distutils.core does not export Command

[...]

PR #946, opened 2020-09-23
Rework stats tests

DougBurke:
# Summary

Update the stats tests to use pytest.

# Details

This is needed by #947 (and hence #831) because in that PR we stop supporting calling XSPEC models with a single grid, which the code here used before this PR (thanks to not using a response function in the model expression).

The tests are changed from unittest style to pytest, taking some advantage of pytest features. We could actually move most of the tests to a single driver routine, passing in the expected values along with the statistic and tolerance, but I have not decided to go that far.

The main change is that the tests used to 

a) not include the response function, so the models were being evaluated with channel numbers, and so weren't "meaningful"
b) many of the tests used "absorption + powerlaw" rather than "absorption * powerlaw"

Technically this is okay, as all we're doing is checking we find the same minimum as previously, but there is a lot of benefit in having the tests be "meaningful", since we can check that the final statistic is sensible (ie ~ dof for those stats that provide a Chi-square-like meaning) and we can see that the best-fit parameter values are "sensible" (ie they are similar between the different statistics). So the tests now include a response and use absorption * powerlaw for the model.

The fits have changed to using the 0.3-7 keV range. Previously the "source" data was 0.5-7 and the background dataset had no energy filter. Using just the 0.5-7 keV range meant that the fits generally were not able to constrain the nH value of the absorption model, so it often fell to very-low levels. Including the 0.3-0.5 keV range means that the nH values are now better constrained, but I have still elected to fit the log of the nH value (at least for the source datasets) since it's easier to check a value like -2.1 rather than 0.008. Ideally we'd use data with a larger nH value to make this a bit better.

The chi-square tests have been changed to use grouped data (group by 20 for source, by 10 for the background as there's less counts) just to better support the chi-square approximation. The Cash/CSTAT tests use the ungrouped data.

The absorption model was changed from XSphabs to XSwabs since we don't really care about the accuracy of this model (and it's not that relevant for the test data we are using here) and the advantage of XSwabs, other than probably being quicker, is that it is unlikely to be changed in future XSPEC releases, as it is marked as "don't use this for science".

### DIFF
--- a/sherpa/stats/tests/test_stats.py
+++ b/sherpa/stats/tests/test_stats.py
@@ -22,19 +22,20 @@ import logging
 import numpy
 import pytest
 
-from sherpa.utils.testing import requires_data, \
-    requires_xspec, requires_fits
 from sherpa.data import Data1D
 from sherpa.astro.data import DataPHA
 
-from sherpa.models import PowLaw1D, SimulFitModel
+from sherpa.astro import ui
+from sherpa.astro.instrument import Response1D
 from sherpa.data import DataSimulFit
 from sherpa.fit import Fit
+from sherpa.models import Const1D, PowLaw1D, SimulFitModel
+from sherpa.optmethods import LevMar, NelderMead
 from sherpa.stats import Cash, CStat, WStat, LeastSq, UserStat, \
     Chi2Gehrels, Chi2ConstVar, Chi2ModVar, Chi2XspecVar, Chi2DataVar
-from sherpa.optmethods import LevMar, NelderMead
 from sherpa.utils.err import StatErr
-from sherpa.astro import ui
+from sherpa.utils.testing import requires_data, requires_fits, \
+    requires_xspec
 
 logger = logging.getLogger("sherpa")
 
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased to get GH Actions and to address simple conficts.

sherpa/stats/tests/test_stats.py
```diff
@@ -240,388 +242,484 @@ def chi_nobkg(self, data, model, *args, **kwargs):
...
+        )
+    }
+
+    compare_results(_fit_simul_datavarstat_results_bench,
+                    result)
+
+
+@requires_data
+@requires_fits
+def test_wstat_calc_stat_info(hide_logging, make_data_path, clean_astro_ui):
```
DougBurke, line 716: Just to note that in working on #1060 I came across a run (using `pytest -n6` to run in parallel` where this test (that is, the old version, it this PR) failed, because `clean_astro_ui` - or equivalent - had not been used. That is, there are reasons for these changes!

DougBurke:
Rebased to address another conflict.

dtnguyen2 (approved):

DougBurke:
Thanks @dtnguyen2 - I know this wasn't an intellectually pleasing PR to review!

PR #830, opened 2020-06-18
WIP Add xspec regrid support

DougBurke:
Support regrid for XSPEC models

- [X] allow XSPEC models to run when regridded
- [ ] support additive models (evaluated correctly with both 1 and 2 arrays)
- [ ] support multiplicative models (evaluated correctly with both 1 and 2 arrays)
- [ ] support composite models (needs #798)

PR #947, opened 2020-09-24
WIP XSPEC changes: integrate setting and require lo,hi grids

DougBurke:
This has been replaced by #1111 (merged), #1112 (merged), and #1179 (WIP). I'm leaving this up as a reminder until #1179 is in better shape.

# Summary

XSPEC model classes must now be evaluated with bin edges - that is with low,high bins. The support for sending in a single grid and treating it a consecutive set of bins has been removed from the model class interface. This feature is still supported for anyone evaluating the models directly from the sherpa.astro.xspec._xspec module.

The XSPEC multiplicative models are now set to have integrate=False, and the integrate setting of XSPEC models can not be changed.

# Details

Neither of these changes has much impact on their own:

 - XSPEC models, when evaluated by an instrument model, already use the two-argument form (ie send in low,high bin edges)
 - the integrate setting of the XSPEC models isn't really used

However, they will be used in #830

Sherpa models with integrate=False and called with low,high bins actually just ignore the high bins (no matter what the documentation says), so after `mdl.integrate = False`, `mdl(elo, ehi)` and `mdl(elo)` will return the same answer. This isn't true of XSPEC additive models since

  a) you can not call it with `mdl(elo)` thanks to this PR
  b) it is up to the model whether to use elo, ehi, or some function of elo,ehi such as the mid-point

I am therefore, with this PR, subtly changing the interpretation of the integrate flag. There is no documentation within the Sherpa code base to indicate what the semantics of this setting are intended to be, so we can only be guided by the actual implementation (the documentation of the integrate flag in our "user-level" docstrings is written to match the existing behavior rather than being a blueprint for the implementation).

# Notes

This builds on #946 (it requires the clean up of the stats tests) and comes out of work on #830. I have pulled it out to try and make things a bit easier to review.

This actually makes a noticeable (but not spectacular) reduction in the time to run the Sherpa test suite (when you have XSPEC support enabled) since we no-longer have to run each XSPEC model with both one and two grids.

### DIFF
--- a/sherpa/astro/sim/tests/test_astro_sim.py
+++ b/sherpa/astro/sim/tests/test_astro_sim.py
@@ -210,7 +210,9 @@ def test_pragbayes_pcaarf_limits(sampler, setup, caplog, reset_seed):
     pmins = np.asarray(covar_results.parmins)
     pmaxs = np.asarray(covar_results.parmaxes)
 
-    fit.model = myabs * mypl
+    # Make sure we add the response
+    rsp = Response1D(fit.data)
+    fit.model = rsp(myabs * mypl)
 
     fit.model.thawedpars = pvals
     fit.model.thawedparmins = pvals + 2 * pmins  # pmins are < 0
--- a/sherpa/astro/ui/tests/test_serialize.py
+++ b/sherpa/astro/ui/tests/test_serialize.py
@@ -284,7 +284,7 @@ pl.ampl.units   = ""
 pl.ampl.frozen  = False
 
 create_model_component("xsphabs", "gal")
-gal.integrate = True
+gal.integrate = False
 
 gal.nH.default_val = 1.0
 gal.nH.default_min = 0.0
@@ -415,7 +415,7 @@ gpl.ampl.units   = ""
 gpl.ampl.frozen  = False
 
 create_model_component("xsphabs", "ggal")
-ggal.integrate = True
+ggal.integrate = False
 
 ggal.nH.default_val = 2.0
 ggal.nH.default_min = 0.0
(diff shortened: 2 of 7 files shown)


DougBurke:
Rebased to pickup GH actions and address conflicts.

dtnguyen2 (approved):
Not sure why I got a random fail for the the _sherpa/ui/tests/test_ui_plot.py_:
```
sherpa/ui/tests/test_ui.py ...........................................   [ 87%]
sherpa/ui/tests/test_ui_plot.py ........................................ [ 88%]
........................................................................ [ 89%]
........................................................................ [ 91%]
........................................................................ [ 92%]
...........................F.......................................      [ 94%]
sherpa/ui/tests/test_ui_unit.py .....                                    [ 94%]
```

It seems to work fine when I re-ran it:
```
(sherpa) [dtn@devel12 sherpa]$ python setup.py test --pytest-arg=sherpa/ui/tests/test_ui_plot.py
...
sherpa/ui/tests/test_ui_plot.py ........................................ [ 12%]
........................................................................ [ 34%]
........................................................................ [ 56%]
........................................................................ [ 79%]
...................................................................      [100%]

========================= 323 passed in 18.33 seconds ==========================
```

DougBurke:
There are some persistent random failures I occasionally try to squash, but it normally requires a moment of luck to identify the problem!

DougBurke:
For reference, I made a change to `test_ui_plot` as part of #1101 which fixed a case like this - see https://github.com/sherpa/sherpa/commit/d651a78fe143b078091a939efa19ae509a073465 - which this PR doesn't include (that was merged Feb 5 and the last update I made to this PR was in January). So I'm going to assume that it's fixed until we see it again!

Issue #529, opened 2018-10-21
Can we include notebooks in our sphinx docs (e.g. as GammaPy do)?


DougBurke:
I note that GammaPy has a bunch of notebooks which get included in the Sphinx documentation. As an example see https://docs.gammapy.org/0.8/tutorials.html  

I think the important part is that they can be run via binder, with a direct link from the documentation, which means that it is easy (assuming it all works) for a user to try them out and explore, without necessarily having to have everything installed.

I can see this being useful for our "worked examples", such as

- the quick guide: https://sherpa.readthedocs.io/en/4.10.1/quick.html
- simple interpolation: https://sherpa.readthedocs.io/en/4.10.1/examples/simple_interpolation.html
- simple user model: https://sherpa.readthedocs.io/en/4.10.1/examples/simple_user_model.html

I do wonder about how the notebooks integrate with the rest of the documentation - for example, the quick guide has a lot of links to other sections, class definitions, and the like which are all handled by Sphinx directives/commands, and presumably they'd have to be emulated or hard-coded in a notebook. So maybe it's more useful for the worked examples than the quick guide.

cdeil:
@bultako has set up the integration of tutorial notebooks, and handling of example datasets, for Gammapy. The docs how that works in detail aren't written yet, but you can have a look [here](https://github.com/gammapy/gammapy-meetings/blob/master/coding_sprints/2018-10_Madrid/slides/setup-docs-nbs-data.pdf).

Short description:
- we store un-executed notebooks (only input and code cells) in version control
- notebooks are executed via `jupyter nbconvert` and then integrated in the docs using `nbsphinx`.
- we only allow "fast" examples that execute in less than a minute, but overall the Gammapy docs build is now ~ 10 min, so pretty slow. It is possible to only run the notebook execute step, or to only run Sphinx, separately.
- we don't have CI set up at the moment, docs builds are run once in a while by a maintainer, and pushed to Github pages (we ran into build timeouts with Readthedocs, so switched away)
- the integration between markdown cells and the rest of the docs isn't great, we use hardcoded URLs to link from the markdown cells to the Sphinx docs, not some directive.

[...]

PR #957, opened 2020-10-05
Docs: include the example notebook in the RTD pages

DougBurke:
# Summary

Add the example notebook (`SherpaQuickStart.ipynb`) to the Sphinx documentation under the "Notebooks" heading.

# Details

This addresses part of #529 - it doesn't address the validation part, but at present the requirement is that these notebooks have been evaluated before passing to Sphinx - i.e. we do not run the notebook conversion during documentation generation.

Fortunately including a Python notebook is easy, but there's some choice of how to handle this. As I am including the notebook from outside the docs/ directory I have decided to copy the .ipynb file at run time, rather than creating a symbolic link, or having two copies.

The notebooks/ directory contains a nbmapping.dat which indicates where the files get copied to in the docs/ directory: it is a tab-separated pair of notebook name and documentation location.

A new section of the main page ("Notebooks") has been added and the notebook placed there. The decision of whether to have a separate section like this or place the notebooks within sections can be made once we make more notebooks.

You can check the documentation at https://sherpa--957.org.readthedocs.build/en/957/ which now has a new Notebooks entry

jupyter-rtd

and the notebook is

https://sherpa--957.org.readthedocs.build/en/957/SherpaQuickStart.html

The notebook will look better once #631 is merged!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,6 +113,7 @@ extern/region-4.11/src/reglexer.c.tmp
 # Sphinx documentation
 docs/_build
 docs/**/api
+docs/SherpaQuickStart.ipynb
 
 #generated from python setup.py test
 .pytest_cache
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -42,6 +42,7 @@ include sherpa/sherpa-standalone.rc
 recursive-include docs *
 
 # Quick Start Tutorial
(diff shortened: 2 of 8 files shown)


anetasie (approved):
Good idea!

anetasie:
@DougBurke nice addition to the documentation!

PR #960, opened 2020-10-05
Fix model evaluation when changing the integrate setting (fix #958)

### LINKED ISSUE #958: Model evaluation is wrong when integrate setting changed; a cache issue?
This got created twice - #958 and #959. I'm closing #959 but there was some useful commentary I added to it.

# Details

With the following program

```
import numpy as np

m1 = Polynom1D()
m2 = Polynom1D()

m1.integrate = False
m2.integrate = True

xgrid = np.arange(2, 10, 2)
xlo, xhi = xgrid[:-1], xgrid[1:]

m1.c0 = 5
m2.c0 = 5
m1.c1 = 1.5
m2.c1 = 1.5

print("m1(x)")
print(m1(xgrid))

print("m2(x)")
print(m2(xgrid))

print("m1(xlo, xhi)")
print(m1(xlo, xhi))

print("m2(xlo, xhi)")
print(m2(xlo, xhi))

print("Changing m1.integrate")
m1.integrate = True

print("m1(xlo, xhi)")
print(m1(xlo, xhi))
```

I would expect the last line to display the same output as m2(xlo, xhi), since the parameters are the same as is the integrate setting. It isn't - it is in fact returning the integrate=False output for m1,  which I *assume* is because of some bizarre cache error.

I have noticed occasional strange results which I believe due to be related to this behavior.

The output I get is

```
m1(x)
[ 8. 11. 14. 17.]
m2(x)
[ 8. 11. 14. 17.]
m1(xlo, xhi)
[ 8. 11. 14.]
m2(xlo, xhi)
[19. 25. 31.]
Changing m1.integrate
m1(xlo, xhi)
[ 8. 11. 14.]
```

DougBurke:
# Summary

Fix an issue with cache evaluation on 1D models using integrated bins and the user has changed the integrate setting of the model.

# Details

I add several tests of the cache code, using low-level tests that the digest is calculated correctly (although as we just repeat the code from the `sherpa.models.model.py::modelCacher1d` routine there is no external validation). There are two tests that show #958 - the original problem and one that I used to identify the problem. It is worth keeping both as they do test slightly-different scenarios and are not expensive tests to run. A number of other tests are also added to check out various features of the code.

The actual fix is very simple - the integrate setting is taken from the input object (called `cls` in the `modelCacher1d` routine) rather than looking for the setting in `kwargs['integrate']`. It is unclear whether we can have a mis-match between `cls.integrate` and `kwargs['integrate']`; I would claim that this indicates a problem somewhere, but it's not clear why we explicitly add the `integrate` setting to the model calls in the first place.

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -59,8 +59,30 @@ def modelCacher1d(func):
         digest = ''
         if use_caching:
 
+            # Up until Sherpa 4.12.2 we used the kwargs to define the
+            # integrate setting, with
+            # boolean_to_byte(kwargs.get('integrate', False)) but
+            # unfortunately this is used in code like
+            #
+            #    @modelCacher1d
+            #    def calc(..):
+            #        kwargs['integrate'] = self.integrate
+            #        return somefunc(... **kwargs)
+            #
+            # and the decorator is applied to calc, which is not
+            # called with a integrate kwarg, rather than the call to
+            # somefunc, which was sent an integrate setting.
+            #
+            try:
+                integrate = cls.integrate
+            except AttributeError:
+                # Rely on the integrate kwarg as there's no
+                # model setting.
+                #
+                integrate = kwargs.get('integrate', False)
+
             data = [numpy.array(pars).tobytes(),
-                    boolean_to_byte(kwargs.get('integrate', False)),
+                    boolean_to_byte(integrate),
                     numpy.asarray(xlo).tobytes()]
             if args:
                 data.append(numpy.asarray(args[0]).tobytes())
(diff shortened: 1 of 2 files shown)


DougBurke:
@dtnguyen2 had mentioned trying the cache code out and saying that it looks like the kwargs integrate setting was never set (it got list to a rebase in this PR) so I added

    assert 'integrate' not in kwargs

to the `modelCacher1d` code and ran the test suite (without the XSPEC models unfortunately because of laptop upgrade reasons), and the tests that failed - i.e. that had the kwarg set - were regrid related. This makes sense since the regrid code can set the integrate kwarg.

Note that #947 and #830 make changes to XSPEC and Unary/BinaryOp models to pass through the integrate setting from their constituent models (when possible).

anetasie:
@DougBurke Was this going to go into 4.12.2?

DougBurke:
I don't think so. Although it's a bug, I'm not sure anyone-but-me would actually do this/get hit by this. We have a number of integrate-related PRs either waiting for review or WIP (related to the regrid code we didn't get to finish) so I think it makes sense for this all to be done as part of the next release.

DougBurke:
Rebased to get GH Actions

PR #968, opened 2020-10-08
Fix 2d image filtering (fix #965)

### LINKED ISSUE #965: segv from combined notice2d and ignore2d filters
This came up via helpdesk, and while I believe this is expected behavior, rather than just a segmentation fault, it would be helpful if a more useful message is thrown.

In standalone Sherpa (4.12 and 4.11), the usage of a notice2d and ignore2d (and similarly notice and ignore) results in Sherpa breaking with a segmentation fault.  A simplified example where this occurs:

```
from sherpa.astro import ui

ui.load_image("acisf06406_000N021_r0273b_ctsimg3.fits")

ui.set_coord("physical")

ui.notice2d('ellipse(4259.3695,3841.6295,13.728445,13.728445,143.90268)')
ui.ignore2d('ellipse(4259.3708,3841.6298,2.7459362,2.7459362,143.90268)')

```

would SegV , but can be worked around:

```
from sherpa.astro import ui

ui.load_image("acisf06406_000N021_r0273b_ctsimg3.fits")

ui.set_coord("physical")

ui.notice2d('ellipse(4259.3695,3841.6295,13.728445,13.728445,143.90268)-ellipse(4259.3708,3841.6298,2.7459362,2.7459362,143.90268)')

ui.get_filter()
````

This behavior is not seen in the Sherpa distributed via the CIAO binaries and conda distribution, presumably due to the lack of DM in standalone.  It would be helpful if some error message is thrown instead of just breaking Sherpa.

Thanks,
~Nick

[acisf06406_000N021_r0273b_ctsimg3.fits.gz](https://github.com/sherpa/sherpa/files/5342878/acisf06406_000N021_r0273b_ctsimg3.fits.gz)

### PR DESCRIPTION
**Pull request by DougBurke**
# Summary

Fix problems with ignore2d and notice2d when multiple regions are used that could cause a segmentation fault.

# Details

I have no idea when or why #965 started - is it that the external region lib code got updated and our _region module didn't follow some changes, or some other reason?

There are two commits to add some basic image tests which we were not covering. The ones in the first commit are not related to the problem but are nice to have. The second commit adds some tests which we have to skip because they cause Python to crash.

The third commit reworks the sherpa.astro.data.DataIMG.notice2d method slightly, to make it a bit clearer what it is doing. It still errors out though.

The fourth commit is where the fix happens:

    I don't actually know why the old code failed, but I decided to
    simplify the region_mask routine in sherpa.astro.utils._region
    which ended up in reworking that module completely. Since this
    is a low-level module, and it is *only* used by sherpa.astro.data
    I felt we were unlikely to be breaking any external code (these
    routines are poorly documented).

    Changes:

    - the import of sherpa.astro.utils._region is now done by
      sherpa.astro.data rather than importing it via
      sherpa.astro.utils
    - the Region lib has changed from accepting a string, which
      must contain a region, to
      - being able to create with no argument, to create an
        empty region
      - if called with a single string it is a region
      - if called with a string and an integer, then if the
        integer is 0 it is taken to be a region, otherwise
        it is taken to be the name of an ASCII file
        containing the region
    - region_mask has been simplified into three routines:
      a) use Region() to parse the region specification
      b) use region_mask to *ONLY* calculate the mask for
         the region (treated as an include)
      c) use region_combine to combine regions

    Somewhere in the rework of region_mask I appear to have fixed
    issue #965

    The tests in test_astro_data2.py - test_img_get_filter_combined
    and test_img_get_filter_excluded - have had the skip pytest
    mark removed, since they now no-longer crash Python. Unfortunately
    there were several gremlins in the original test (since I
    couldn't run it to test the expected outout) which had to be
    fixed at the same time.

    The tests in test_region_unit have had to be changed since
    region_mask has changed significantly. I've added a number of
    simple tests to improve the coverage of the C++ code.

    I have spent time reviewing the Region, region_mask, and
    region_combine routines to check they do not leak memory,
    running a loop like

    while True:
       ...:     r1 = Region('circle(200,200,20)')
       ...:     r2 = Region('circle(205,205,20)')
       ...:     r = region_combine(r1, r2)
       ...:     region_mask(r, [200, 300, 210], [200, 300, 210])

    while checking top, and (after some fun) the memory use
    looks good.

The fifth commit moves the methods region_mask and region_combine to methods of the Region
class, as it makes more sense to have them as methods (they both require a region, so may as well
make them methods!). This makes the memory leak check

    from sherpa.astro.utils._region import Region
    while True:
       ...:     r1 = Region('circle(200,200,20)')
       ...:     r2 = Region('circle(205,205,20)')
       ...:     r = r1.combine(r2)
       ...:     r.mask([200, 300, 210], [200, 300, 210])

The sixth commit just adds some tests that the coverage checks showed were missing.

Commit 7 removes an unnecessary cast.

Commit 8 fixes a bug introduced in the rewrite: if the first region added is an excluded region then this wasn't being handled correctly.

# Notes

I have my suspicions that the bug is in the old `region_mask` section

```
  if( reverse ) {
    regRegion *tmp = reg;
    reg = regInvert( tmp );
    if( tmp ) regFree( tmp );
  }

  // If possible, append the new region string to the stack
  if( reg_obj != Py_None ) {
    region = (PyRegion*)(reg_obj);
    regRegion *tmp = reg;

    reg = regCombineRegion( region->region, tmp );

    if ( NULL == reg) {
      PyErr_SetString( PyExc_TypeError,
                       (char*)"unable to combine regions successfully" );
      return NULL;
    }

    pyRegion_dealloc( region );
    if( tmp ) regFree( tmp );
  }
  region = (PyRegion*) pyRegion_build( &pyRegion_Type, reg );

  Py_XINCREF(region);

  return Py_BuildValue((char*)"ON", region, mask.return_new_ref());
```

and has several suspicious areas - the freeing of regions associated with the Python region structure, a random pyRegion_dealloc call to an object that is in use, and the Py_XINCREF - but I haven't spent time trying to track
down what the problem was (or, probably were).

### DIFF
--- a/sherpa/astro/utils/tests/test_region_unit.py
+++ b/sherpa/astro/utils/tests/test_region_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -17,7 +17,8 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 import pytest
-from sherpa.astro.utils._region import region_mask, Region
+
+from sherpa.astro.utils._region import Region, region_combine, region_mask
 
 
 region_string = 'Circle(256 ,256 , 25)'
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by dtnguyen2 — COMMENTED** [review]
Regarding the comment on line 45 of the file _sherpa/astro/utils/src/_region.cc:

```
static regRegion* parse_string( char* str, int fileflag ) {

   regRegion *reg = NULL;
   std::string input(str); // what benefit do we get casting to std:string?
   if( fileflag ) {
     reg = regReadAsciiRegion( (char*)input.c_str() , 0 ); // Verbosity set to 0
   } else {
     reg = regParse( (char*)input.c_str() );
   }

   return reg;
 }
```

There is no benefit, in fact I would think that while it is not time and resources consuming you are forcing it to do an extra memory allocation followed by a string copy to put it a C++ class only to access it as a C ```char *```

**Comment by DougBurke** [comment]
@dtnguyen2 - ta. That's what I thought but it's not my code so I couldn't tell. This function is only used in one place I thing, so maybe it could be sent a std::string (too late for me to check). It's got to end up as char* at the end though!

**Comment by DougBurke** [comment]
Checking against CIAO 4.12.1 I do see a difference. For a DataIMG with no filter,

    d.notice2d(shape, ignore=True)
    d._region

returns "!" + shape for CIAO 4.12.1 (this makes sense), but shape for this PR. I need to investigate.

**Comment by DougBurke** [comment]
I now get the correct behavior:

```
d.notice2d()
d.notice2d('circle(100,100,10)', ignore=True)
print(d.get_filter())
```

returns

```
!Circle(100,100,10)
```

I have done some comparisons with CIAO 4.12.1 (for the cases where it wouldn't crash) and I get the same results now.

File: sherpa/astro/utils/tests/test_region_unit.py [hunk 1]
```diff
@@ -17,57 +17,305 @@
...
+def test_region_empty():
+    """The empty region"""
+    r = Region()
+    assert '' == str(r)
+
+
 def test_region_string():
+    """We can create with a string.
+
+    Note the regon gets normalized.
```
wmclaugh, line 45: regon -> region

DougBurke: Ta -fixed

DougBurke:
Ta @wmclaugh.

PR #967, opened 2020-10-08
Testing Python 3.8 and PR # 969

DougBurke:
Let's see what Python 3.8 does to our "full" tests. We can see that without #967 the tests fail, but they pass after.

We should think about updating the Travis tests, but this is for after the 4.12.2 release.

This PR is not intended for merging. Just testing.

PR #969, opened 2020-10-08
Support Python 3.8 for MacOSX

### LINKED ISSUE #730: Python 3.8 test failures
# Still open

There appears to be serious problems on macOS but not Linux (installed via conda).

# Fixed by #696 

As conda now has python 3.8 available (but not with `astropy` or `matplotlib` just yet) I took the latest `master` branch out for a spin and there were several test failures:

```
========================================================================================= test session starts =========================================================================================
platform linux -- Python 3.8.1, pytest-5.3.3, py-1.8.1, pluggy-0.13.1
rootdir: /home/dburke/sherpa/sherpa-pr640, inifile: pytest.ini
plugins: xvfb-1.2.0
collected 2358 items                                                                                                                                                                                  
...
sherpa/models/tests/test_regrid_unit.py ..............................................F.................................                                                                        [ 30%]
...
sherpa/optmethods/tests/test_optmethods.py .E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E..E.E.E.E.E.E.E                                                                                                [ 33%]
...
sherpa/utils/tests/test_psf_rebinning_unit.py FFFFFFFFFFFF.................                                                                                                                     [ 95%]
...
```

It looks like some clean up is needed:

a) `test_optmethods.py`

```
*** Warnings created: 1
1/1 {message : DeprecationWarning("PY_SSIZE_T_CLEAN will be required for '#' formats"), category : 'DeprecationWarning', filename : '/home/dburke/sherpa/sherpa-pr640/sherpa/optmethods/tests/test_optmethods.py', lineno : 110, line : None}
```

b) `test_regrid_unit.py` and `test_psf_rebinning_unit.py`

I wonder if these are NumPy 1.18 errors instead of Python 3.8 (although maybe a Python 3.8 change has lead to these now being created/seen/...?)

```
TypeError: only integer scalar arrays can be converted to a scalar index
```

c) some general warnings

```
sherpa/astro/data.py:2600
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2600: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'logical':

sherpa/astro/data.py:2609
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2609: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'physical':

sherpa/astro/data.py:2618
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2618: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'world':

sherpa/astro/data.py:2811
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2811: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'logical':

sherpa/astro/data.py:2826
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2826: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'physical':

sherpa/astro/data.py:2841
  /home/dburke/sherpa/sherpa-pr640/sherpa/astro/data.py:2841: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    if coord is not 'world':

sherpa/ui/utils.py:258
  /home/dburke/sherpa/sherpa-pr640/sherpa/ui/utils.py:258: SyntaxWarning: "is" with a literal. Did you mean "=="?
    (name is '_sherpa_version') or

sherpa/ui/utils.py:259
  /home/dburke/sherpa/sherpa-pr640/sherpa/ui/utils.py:259: SyntaxWarning: "is" with a literal. Did you mean "=="?
    (name is '_sherpa_version_string')):
```

anetasie:
Add a new multiprocessing_start_method option to `sherpa.rc` and `sherpa-standalone.rc` and `init` code to set start multiprocessing method to `fork` by default.

Fixes #730 

## Functional testing:

Used Python 3.8 and run tests using 'pytest':
/sherpa/sherpa/optmethods/tests/test_optmethods.py
/sherpa/sherpa/optmethods/tests/test_ncoresopt.py

for all the options: no option set in the configuration file, and the specific options 'fork', 'default', 'spawn', and typos in the setting of the multiprocessing_start_method. 

Tests importing sherpa into ipython to check the error msgs.

Still need to test with Python 3.7.

with help from @taldcroft


DougBurke:
I've just merged this into #967 so we'll hopefully get macOS python 3.8 tests from Travis. Assuming I did the merge right. And didn't mess up the travis settings.

sherpa/optmethods/opt.py
```diff
@@ -20,9 +20,9 @@
 #
 
 import numpy as np
-import multiprocessing
 import random
 from sherpa.utils import Knuth_close, _multi, _ncpus, run_tasks, func_counter
+import multiprocessing
```
DougBurke, line 25: My pylint keeps on complaining about having Python models after numpy or sherpa modules. This is not a functional change but a style change. If we were to redo it it "should" be

```
import multiprocessing

import numpy as np

from sherpa.utils import kitchen_sink
```

sherpa/sherpa.rc
```diff
@@ -46,3 +46,13 @@ arraylength: 1000000
...
+[multiprocessing]
+# Define the method by which the multiprocessing package starts
+# parallel processes. Sherpa requires the "fork" method in order to
+# utilize multiprocessing to improve performance and pass all tests. 
+# However, on MacOSX there are known issues with using "fork" that 
+# prompted a change in the multiprocessing default to be "spawn" on 
+# MacOSX. Expert users may consider changing this value to either
+# "spawn" or "default", where the latter will use the 
+# platform-specific start method of the multiprocessing package.
+multiprocessing_start_method: fork
```
DougBurke, line 58: GitHub is warning that the last line doesn't end with a new-line character

sherpa/utils/__init__.py
```diff
@@ -63,8 +63,19 @@
 
 try:
     import multiprocessing
-    _multi = True
+    try:
+        multiprocessing_start_method = config.get('multiprocessing', 'multiprocessing_start_method')
```
DougBurke, line 67: Do we need the try/except - can we not just say

```
blah = config.get('multiprocessing', 'long name', fallback='fork')
```

or does the fallback option not catch all error cases (if so it would mean looking at the other config.get calls we have).

DougBurke (approved):
I've left several minor comments, but it **looks** fine. We;ll have to see what the travis tests from #967 say though!

sherpa/optmethods/opt.py
anetasie: (from @taldcroft) - The intent was to be explicit that the first time multiprocessing is imported anywhere is within `sherpa.utils` where the start method is handled. In reality this may not be necessary. IIRC there can be a problem if the start method is changed after actually using multiprocessing, but I suppose that isn't happening here. I'm all for keeping the linters happy when possible, so we can probably revert this.

sherpa/utils/__init__.py
anetasie: The `config.get` docstring was sufficiently confusing that I wasn't sure about `fallback`. Assuming that works even with a missing section then definitely :+1:.

DougBurke:
The #967 tests look good - the macOS build with XSPEC passes - https://travis-ci.org/github/sherpa/sherpa/jobs/734127166

hamogu:
I did not follow the discussion in #730 enough to have an informed opinion, thus no formal review from me. If this is what @taldcroft recommends then I'm sure it's right ;-)

sherpa/utils/__init__.py
anetasie: @DougBurke  fallback option looks good. I'll update the code in a few min.

sherpa/sherpa.rc
anetasie: I'll update the space

sherpa/optmethods/opt.py
anetasie: I just reverse this change.

anetasie:
@DougBurke thanks for the update on the tests.

DougBurke:
One thing I am concerned a little about is that we import multiprocessing into several modules - sherpa.utils, sherpa.optmethods.opt, sherpa.estmethods - and it's not clear to me that users who just import sherpa.optmethods.opt would get the fix.

Fortunately it looks like the multiprocessing imports are all after importing sherpa.utils, so it seems to be okay. However, there's no documentation on those imports to highlight the ordering (or in the sherpa.utils file too, from my earlier discussion with @anetasie ). We could add comments like '# Must be imported after sherpa.utils' to the multiprocessing imports, but I wonder if long term we should have a sherpa.utils.multiprocessing module (or sherpa.multiprocessing) so that we know the setup has been called. That is definitely for a later PR.

taldcroft:
> we should have a sherpa.utils.multiprocessing module (or sherpa.multiprocessing)

👍

hamogu:
I'm also not convinced that the current approach of (spawning or forking) a new process.
I threw out a few ideas in
https://github.com/sherpa/sherpa/pull/483#pullrequestreview-416579892
https://github.com/sherpa/sherpa/pull/483#issuecomment-638374891
and I've been thinking about this a bit (we are not the only package with this problem. There must be a python package out that that solves that for us...), but I have no solution to show. 
Any new development would be easier if this was all concentrated in just a single module.

anetasie:
We should work out a better and more general solution to the multiprocessing implementation. Also, there are areas of sherpa functionality beyond what is in the code now and considering overall upgrade and design for multiprocessing would be good.

DougBurke:
@anetasie - there is one more change I think this should have. The SHERPA_VERSION line needs updating to 4.12.2

```
%  grep SHERPA sherpa/sherpa.rc
# SHERPA_VERSION 4.12.1
% grep SHERPA sherpa/sherpa-standalone.rc
# SHERPA_VERSION 4.12.1
```

dtnguyen2 (approved):

PR #976, opened 2020-10-13
Update Sherpa version example and Python versions in CONTRIBUTING.md

Marie-Terrell:
This PR is just a minor documentation update to the Sherpa version example and the Python versions in the CONTRIBUTING.md file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,10 +16,10 @@ the code you are using - in particular for Sherpa, NumPy, and Python. The
 Sherpa version can be found using
 
     $ python -c 'import sherpa; print(sherpa.__version__)'
-    4.10.0
+    4.12.1
 
 When using a development version of Sherpa, the value will include the
-Git commit. An example of this is ``4.10.0+177.g3540951``.
+Git commit. An example of this is ``4.12.1+327.ge5183cd4``.
 
 Contributing code
 -----------------
@@ -96,7 +96,7 @@ Example: new functionality, such as
 
 **Software versions**
 
-Development should use Python 3.5 or later (Sherpa does support Python 2.7, 
+Development should use Python 3.6 or later (Sherpa does support Python 3.5,
 but this will be dropped soon).
 
 Ideally, NumPy support should be 1.6 or greater, but please include a comment


DougBurke:
We should probably also change

- docs/index.rst (there's only one line which  mentions 3.5)
- docs/install.rst (there are two lines which mention Python 3.5)

to follow this language.

Marie-Terrell:
The Python version listing update was moreso to remove the comment mentioning support for 2.7. Though, it does now mention that Python 3.5 support would be ending soon. Those files say something along the lines of "Python 3.5, 3.6, or 3.7 (there has been limited testing with Python 3.8)". Are you suggesting switching that up to something along the lines of: "Python 3.5, 3.6, or 3.7 (there has been limited testing with Python 3.8). Python 3.5 support will be dropped soon."

DougBurke:
Ideally we'd have consistent information in the various places we talk about supported versions of Python, which appears to be the CONTRIBUTING.md file and the two docs/*rst files I mentioned. It's probably easiest at this time to change CONTRIBUTING.md to match the other cases (as there's less to do) and say something like

    Python 3.5, 3.6, or 3.7 (there has been limited testing with Python 3.8).

or

    Python 3.5, 3.6, 3.7, and 3.8.

since we hope we'v not got 3.8 working on macOS with #969 (but that requires changing the .rst fiiles too). 


I tend to try and be explicit about the Python versions, rather than say "Python 3.x and up". For instance, I note that Python 3.9 has just come out but we have no experience with it.


I am not sure we need to say "Python 3.5 will be dropped soon" in this file, since we can just update the documentation when we get to it (since "soon" isn't that helpful a signifier). That is, as long as we remember to update the documentation, which, as this PR shows, we haven't been able to do :-)

hamogu:
If, instead of "soon" we say "in 2021", then that's accurate. We still have 3.5support  in 4.12.2, which means that 4.13 will also support 3.5 (even if we don't build, e.g. conda packges for 3.5 and test for 3.5 specifically). On the other hand, I want to start using f-strings and other 3.6+ goodies (we have an issue about data classes), so it's safe to assume that 3.5 won't be supported on 4.14. But, if we forget to update this and someone comes across the text in 2022, they know tat 3.5 has been dropped.

DougBurke (approved):
I don't have the mental energy to work on this, so I think we take this as is.

Issue #970, opened 2020-10-09
Build on aarch64 fails


ggardet:
The error is:
```
[  232s] sherpa/utils/src/cephes/const.c:92:19: warning: implicit declaration of function ‘strtod’ [-Wimplicit-function-declaration]
[  232s]    92 | double INFINITY = strtod("INF");  /* 99e999; */
[  232s]       |                   ^~~~~~
[  232s] sherpa/utils/src/cephes/const.c:92:19: error: initializer element is not constant
[  232s] sherpa/utils/src/cephes/const.c:97:14: error: initializer element is not constant
[  232s]    97 | double NAN = strtod("NAN");
[  232s]       |
```

DougBurke:
We have no experience on aarch64 so any experience/help would be appreciated.

`helpers/extensions/__init__.py` defines how the compiled code is built, but there's no obvious platform-specific rules there.

`helpers/sherpa_config` has a bunch of options for configuring the build, there may be something there.

I'm assuming that as you got to the `sherpa/utils/src/` part of building then the external dependencies - which are defined in `helpers/deps/__init__.py` - built. Whether they work is another question.

ggardet:
If I apply the following patch to build as if `INFINITIES` and `NANS` where not defined:
```
--- sherpa-4.12.1.orig/sherpa/utils/src/cephes/const.c	2020-10-06 16:01:08.401520790 +0200
+++ sherpa-4.12.1/sherpa/utils/src/cephes/const.c	2020-10-07 11:14:37.547135589 +0200
@@ -89,12 +90,13 @@ double LOGSQ2 =  3.46573590279972654709E
 double THPIO4 =  2.35619449019234492885;       /* 3*pi/4 */
 double TWOOPI =  6.36619772367581343075535E-1; /* 2/pi */
 #ifdef INFINITIES
-double INFINITY = strtod("INF");  /* 99e999; */
+//double INFINITY = strtod("INF");  /* 99e999; */
+double INFINITY =  1.79769313486231570815E308;    /* 2**1024*(1-MACHEP) */
 #else
 double INFINITY =  1.79769313486231570815E308;    /* 2**1024*(1-MACHEP) */
 #endif
 #ifdef NANS
-double NAN = strtod("NAN");
+double NAN = 0.0; 
 #else
 double NAN = 0.0;
 #endif
```

I get a failure on tests:
```
[  422s] =================================== FAILURES ===================================
[...]

PR #978, opened 2020-10-15
Use C99 def for INFINITY, NAN, isfinite and isnan to build on aarch64 (fix issue #970)

dtnguyen2:
# Note

Use the math constants (```IFINITIY```, ```NAN```) and funcs (```isfinite```, ```isnan```, ```signbit```) from a C99 compliant compiler if the compiler option ```-std=c99``` or greater is used, otherwise use the quantities as defined by the library.

[See how test was done for details](https://github.com/sherpa/sherpa/pull/978#issuecomment-726316153)

### DIFF
--- a/helpers/extensions/__init__.py
+++ b/helpers/extensions/__init__.py
@@ -147,6 +147,8 @@ utils = Extension('sherpa.utils._utils',
               depends=(get_deps(['extension', 'utils'])+
                        ['sherpa/utils/src/gsl/fcmp.h',
                         'sherpa/utils/src/cephes/cephes.h',
+                        'sherpa/utils/src/cephes/mconf.h',
+                        'sherpa/utils/src/cephes/protos.h',
                         'sherpa/utils/src/sjohnson/Faddeeva.hh']))
 
 modelfcts = Extension('sherpa.models._modelfcts',
--- a/sherpa/utils/src/cephes/const.c
+++ b/sherpa/utils/src/cephes/const.c
@@ -134,7 +134,7 @@ unsigned short LOGE2[4]  = {0x39ef,0xfefa,0x2e42,0x3fe6};
 unsigned short LOGSQ2[4] = {0x39ef,0xfefa,0x2e42,0x3fd6};
 unsigned short THPIO4[4] = {0x21d2,0x7f33,0xd97c,0x4002};
 unsigned short TWOOPI[4] = {0xc883,0x6dc9,0x5f30,0x3fe4};
-#ifdef NOC99COMPILER
+#ifndef __STDC_VERSION__
 #ifdef INFINITIES
 unsigned short INFINITY[4] = {0x0000,0x0000,0x0000,0x7ff0};
 #else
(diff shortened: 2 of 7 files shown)


DougBurke:
So the idea is someone wants to use this some way-out-there compiler (borland-c++++-for-windows-3?) they'd just have to set the new define, but everyone else can just get on with their life and ignore it?

dtnguyen2:
Yes.  As is: if one has a ```C99``` compiler then the code should work.  With a little more work, I can restore (I think by adding one ```#ifdef```statement in the _mconf.h_ file) the code to its original state by uncomment one line.  Will do when I get a chance.

dtnguyen2:
Not sure what I was thinking when I worked on this PR, but I should use a similar ```#ifdef``` as was used in PR #859 so it will be automatic

dtnguyen2:
Using our archaic default compiler:
```
(sherpa) [dtn@devel12 sherpa]$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
A simple prototype:
```
(sherpa) [dtn@devel12 ~]$ cat tmp.c
#include <stdio.h>

#if (__STDC_VERSION__ >= 199901L)
#include <math.h>
#endif

int main() {

#ifndef __STDC_VERSION__
  printf("dinosaurs were roaming the earth when compiler was written\n");
#else
  printf("the compiler is at least C99 compliant\n");
  printf("C99 compliant compiler has %g\n", NAN);
  printf("C99 compliant compiler has %g\n", INFINITY);
  printf("C99 compliant compiler has isfinite %d\n", isfinite(3));
  printf("C99 compliant compiler has signbit %d\n", signbit(3));
#endif

  return 0;
}
```

Test the prototype to make sure it works correctly if the compiler is not C99 compliant:

```
(sherpa) [dtn@devel12 ~]$ gcc tmp.c; a.out
dinosaurs were roaming the earth when compiler was written
```

Test to see if the ```NAN```, ```INFINITY```, ```isfinite``` and ```signbit``` are defined in a C99 compliant compiler:

```
(sherpa) [dtn@devel12 ~]$ gcc -std=c99 tmp.c; a.out
the compiler is at least C99 compliant
C99 compliant compiler has nan
C99 compliant compiler has inf
C99 compliant compiler has isfinite 1
C99 compliant compiler has signbit 0
```
The default compilation options:
```
building 'sherpa.utils._utils' extension
compiling C sources
C compiler: gcc -pthread -B /export/miniconda/envs/sherpa/compiler_compat -Wl,--sysroot=/ -Wsign-compare 
-DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC

```
Thanks to @Marie-Terrell, by entering the following before compiling
```
export CFLAGS="-std=c99"
```
comparing to the default compiler options, there is now an additional option of ```-std=c99```:
```
building 'sherpa.utils._utils' extension
compiling C sources
C compiler: gcc -pthread -B /export/miniconda/envs/sherpa/compiler_compat -Wl,--sysroot=/ -DNDEBUG -g -fwrapv 
-O3 -Wall -Wstrict-prototypes -std=c99 -fPIC
```

DougBurke:
I just kicked the mac test on travis so we can (hopefully) get a clean travis run

DougBurke:
@dtnguyen2 - this looks like it's ready for review -  do you want the "draft" label to be removed?

dtnguyen2:
@DougBurke The "draft" label has been removed

DougBurke (approved):
LGTM

PR #979, opened 2020-10-15
Improve plots of PHA data sets and models when the response grid is not ideal

### LINKED ISSUE #977: plotting PHA data/models when the grids don't quite match
This is the output of

    plot_data(color='green', 'linestyle='-', marker='', yerrorbars=False)
    plot_model(overplot=True)

for a PHA dataset (HRC-S grating). Note that this doesn't happen for all such datasets.

FIG

You can see that the vertical line between some, but not all, bins is missing. This is because the `BIN_LO/BIN_HI` sets of data in the PHA file don't quite match up:

```
sherpa In [34]: d = get_data()

sherpa In [35]: idx = np.arange(8462, 8466)

sherpa In [36]: for x in zip(d.bin_hi[idx], d.bin_lo[idx]):
           ...:     print(x[0], x[1])
           ...:
100.025 100.0125
100.0125 100.0
100.00000000000001 99.98750000000001
99.98750000000001 99.97500000000001

```

You can see that the bin at 100 Angstroms doesn't quite match.

This messes up the code added in #906 / #931, which assumes that the bins exactly match. So, there's two things to think about here:

a) a fix in the plotting code that uses `sherpa.utils.sao_fcmp` to identify bins that are close enough together

b) think about changes to the DataPHA/ARF/RMF classes to insert some validation stage that "corrects" for bins that don't match (in a perfect world with a warning to users that this has happened, but in reality they do not know how to fix this, and the upstream producers of the data may not be able to fix this, so it's not clear it is that useful).

DougBurke:
# Summary

Improve the display (plots) of PHA data and models when the response grid contains minor numerical differences. This fixes #977.

# Details

    The code now checks for "close" edges and replaces them so that
    they are identical. In this case the sherpa.utils.sao_fcmp method,
    with a tolerance of numpy.finfo(numpy.float32).eps, is used to
    identify close values. This works for the problematic case found
    in #977 but is not a guarantee for all cases. if the difference is
    much larger than I argue that the PHA or response file is "not
    ideal" and really it's an issue for the data producer.

    I could not see any obvious code in sherpa.astro.data or
    sherpa.astro.instrument that worries about such differences. This
    should really be addressed but is a much-larger issue than this
    fix.

    One possibility for fixing #977 was to place the logic into the
    sherpa.plot histogram code, so that all histogram plots would
    benefit. I did not do this because for the PHA data we have
    domain knowledge that the bins are not going to relevant at
    very-small values (e.g. float32 eps sizes), whereas in the
    general case we do not. It therefore seems safer to only
    apply a fix for these cases.

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -21,6 +21,11 @@
 Classes for plotting, analysis of astronomical data sets
 """
 
+import logging
+
+import numpy as np
+from numpy import iterable, array2string, asarray
+
 from sherpa.astro.data import DataPHA
 from sherpa.plot import DataPlot, ModelPlot, FitPlot, DelchiPlot, ResidPlot, \
     RatioPlot, ChisqrPlot, HistogramPlot, backend, Histogram
@@ -28,9 +33,8 @@ from sherpa.plot import ComponentSourcePlot as _ComponentSourcePlot
 import sherpa.plot
 from sherpa.astro.utils import bounds_check
 from sherpa.utils.err import PlotErr, IOErr
-from sherpa.utils import parse_expr, dataspace1d, histogram1d, filter_bins
-from numpy import iterable, array2string, asarray
-import logging
+from sherpa.utils import parse_expr, dataspace1d, histogram1d, filter_bins, \
+    sao_fcmp
 
 warning = logging.getLogger(__name__).warning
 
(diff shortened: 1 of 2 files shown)


sherpa/astro/plot.py
```diff
@@ -42,6 +46,42 @@
...
+    hold.
+
+    """
+
+    # Technically idx should be 0 or 1, with no -1 values. We
+    # do not enforce this. What we do is to take all bins that
+    # appear similar (sao_fcmp==0) and set the xlo[i+1] bin
+    # to the xhi[i] value.
+    #
+    equal = sao_fcmp(plot.xlo[1:], plot.xhi[:-1], _tol)
```
hamogu, line 80: I'm fine with this as is, but I'm wondering why not do it with plain numpy?
```
equal = np.close(plot.xlo[1:], plot.xhi[:-1])
```
(could set `tol=_tol`, but default is already chosen to catch round-off errors for 32 bit systems).

DougBurke: I did not know about np.close and I know about sao_fcmp...

dtnguyen2: ```sao_fcmp``` is more robust, checkout @DougBurke's comment:
[...which sao_fcmp thinks is the same, but numpy.digitize doesn't...](https://github.com/sherpa/sherpa/pull/756#issue-386719738)

hamogu: @dtnguyen2 The linked comment is about np.digitize which is made to digitize by looking for exactly identical values. `np.close` and its cousin `np.allclose` will tell you if two values are approximately the same with any tolerance you like (but default parameter are set to test for differences typical for float32 round-off errors). Yes, you could argue that there should be a `np.close_digitze` that performs the digitize operation using np.close, but that's not what's needed here.

dtnguyen2: From the numpy doc:
[absolute(a - b) <= (atol + rtol * absolute(b))](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html)

subtracting two numbers that are close, I guess the numpy folks have a lot of numerical expert.  Whereas, the following comment is from gsl ```fcmp``` which ```sao_fcmp``` is built on to fix a few bugs.

>     /* x1 || x2 is 0 so the following substraction is kosher */

In numpy we trust?

DougBurke: As the code works I'd rather not change anything at this time.

hamogu: I'm sure either is fine
If I want to be snarky then I could point out that 
> is build on to fix a few bugs

does not exactly inspire confidence in the gsl code...

dtnguyen2 (approved):

PR #984, opened 2020-10-19
Address set_xlog/ylog problems with DataPHA/Data1DInt classes (#981)

### LINKED ISSUE #981: set_[xy]log/linear do not work for Data1DInt and DataPHA data
```

In [2]: load_arrays(1, [10, 70, 300], [1, 2,3])

In [3]: load_arrays(2, [10, 70, 300], [50, 100, 500], [1, 2,3], Data1DInt)

In [4]: load_arrays(3, [1, 2, 3], [2, 3, 4], DataPHA)

In [5]: set_xlog()

In [6]: plot('data', 1, 'data', 2, 'data', 3, linestyle='-')
```

FIG

Note that only the top-left plot has a log scale for the x axis.

The problem (I think) is that the set_xxx routines do not know about the new data objects for Data1DInt/DataPHA plots added in #906 and #931. I imagine we need to have a version in both sherpa.ui.utils and sherpa.astro.ui.utils (as only the latter has the PHA models).

DougBurke:
# Summary

Fix the set_xlog/ylog routines for PHA and 1D integrated datasets. Fixes #981

# Details

The first commit adds tests that show issue #981. They require matplotlib (since you only see the problem when
you create the plot, not with the data returned by get_data_plot). Tests that fail are marked as xfail. Unfortunately, thanks to the
way the tests are structured, several tests are marked as xfail even though they pass, but I think that's okay here.

The second commit just moves to a "more Pythonic" approach (just access the attributes we need, handling any possible error, rather than look for a particular class hierarchy).

The third commit handles the set_x/y[log|linear] routine for DataPHA objects. It also fixes a subtle bug introduced in #931 where use of the clean method would not have reset the plot options for DataPA objects (a test has been added to check this fix).

The fourth commit changes the tests since they can be run without requiring matplotlib. This requires a little bit of work as we now have to query a preference setting rather than ask the axis what the scaling is, which ends up requiring more knowledge of the internal data structures.

The fifth commit just extends the number of plot types that get tested.

The sixth commit fixes the problem for the Data1DInt data type, by adding entries to the _plot_types dictionary. This only fixes the case when the 'all' option (the default) is used: it wouldn't work for 'set_xlog("data")' for Data1DInt datasets.

The seventh commit adds more tests. In particular it adds tests for the 'set_xlog("data")' command (which are marked as xfail).

The eighth commit fixes the set_xlog/ylog calls for DataPHA classes and also addresses the set_xlog('data') and set_ylog('model') style options for all data types by changing the _plot_types dictionary so the values are a list of plot types rather than a single plot type. This means we can identify all the data types as "data" rather than having data, astrodata, data1dint, ... It means we can remove the _set_plot_type method in sherpa.astro.ui.Session since it can now just use the parent classes method.

The last commit are typos and noting an apparently-odd construct that Warren noted in the tests.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -3132,7 +3132,7 @@ def test_set_plot_opt_y(cls, datafunc, plotfunc, answer):
     """Does set_ylog/ylinear work?
 
     Unlike test_set_plot_opt_x, the Y axis setting of the plot
-    does not necessarily folloe the set_ylog setting (e.g.
+    does not necessarily follow the set_ylog setting (e.g.
     the residual plots).
     """
 
@@ -3166,6 +3166,10 @@ def test_set_plot_opt_y(cls, datafunc, plotfunc, answer):
             assert p1.dataplot.plot_prefs['ylog'] == answer
             assert p1.modelplot.plot_prefs['ylog'] == answer
     elif plotfunc in ['resid', 'ratio', 'delchi']:
+        # We use plot_prefs even if is_int is True for
+        # the residual-style plots. That is, the ordering
+        # of the checks here is important.
+        #
         assert p1.plot_prefs['ylog'] == answer
     elif is_int:
         assert p1.histo_prefs['ylog'] == answer
@@ -3247,7 +3251,7 @@ def test_set_plot_opt_x_astro(cls, datafunc, plotfunc):
         # Note that for DataPHA datasets the modelplot object is
         # created on-the-fly using sherpa.astro.plot.ModelPHAHistogram,
         # rather than a session._plotobj value that is also set in
-        # session._plot_types, so it doesn't get changed bu set_xlog etc.
+        # session._plot_types, so it doesn't get changed by set_xlog etc.
         #
         # Ideally this would match the dataplot setting but it isn't
         # actually required for the plot to work.
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -2974,7 +2974,11 @@ def example_data1dint():
 
 
 def example_datapha():
-    """Create a DataPHA object"""
+    """Create a DataPHA object.
+
+    The background is set to be equal to the source just
+    because it is easier.
+    """
     chans = np.arange(1, 6, dtype=np.int16)
     d = DataPHA('example.pha', chans, chans)
 
@@ -2982,6 +2986,11 @@ def example_datapha():
     arf = create_arf(ebins[:-1], ebins[1:])
 
     d.set_arf(arf)
+
+    # re-create so we don't have a circular dataset
+    b = DataPHA('background.pha', chans, chans)
+    d.set_background(b)
+
     return d
 
 
@@ -3034,6 +3043,7 @@ def test_datapha_plot_after_clean():
 @pytest.mark.parametrize("plotfunc",
                          ['data',
                           'model',
+                          'source',
                           'fit',
                           'resid',
                           'ratio',
@@ -3112,6 +3122,7 @@ def test_set_plot_opt_x(cls, datafunc, plotfunc):
 @pytest.mark.parametrize("plotfunc,answer",
                          [('data', True),
                           ('model', True),
+                          ('source', True),
                           ('fit', True),
                           ('resid', False),
                           ('ratio', False),
@@ -3187,10 +3198,18 @@ def test_set_plot_opt_y(cls, datafunc, plotfunc, answer):
 @pytest.mark.parametrize("plotfunc",
                          ['data',
                           'model',
+                          'source',
                           'fit',
                           'resid',
                           'ratio',
-                          'delchi'
+                          'delchi',
+                          'bkg',
+                          'bkg_model',
+                          'bkg_source',
+                          'bkg_fit',
+                          'bkg_resid',
+                          'bkg_ratio',
+                          'bkg_delchi',
                          ])
 def test_set_plot_opt_x_astro(cls, datafunc, plotfunc):
     """Does set_xlog/xlinear work? Astro data objects only.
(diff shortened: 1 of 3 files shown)


sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -2954,3 +2956,609 @@ def test_pha1_plot_multiple_args(clean_astro_ui, basic_pha1):
...
+                          ('fit', True),
+                          ('resid', False),
+                          ('ratio', False),
+                          ('delchi', False)
+                         ])
+def test_set_plot_opt_y(cls, datafunc, plotfunc, answer):
+    """Does set_ylog/ylinear work?
+
+    Unlike test_set_plot_opt_x, the Y axis setting of the plot
+    does not necessarily folloe the set_ylog setting (e.g.
```
wmclaugh, line 3135: folloe -> follow

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -2954,3 +2956,609 @@ def test_pha1_plot_multiple_args(clean_astro_ui, basic_pha1):
...
+    # Create the plot
+    plot()
+    p1 = pdata()
+    if plotfunc in ['fit', 'bkg_fit']:
+        assert p1.dataplot.histo_prefs['xlog']
+        # Check the current behavior of the model plot in case it changes.
+        # Note that for DataPHA datasets the modelplot object is
+        # created on-the-fly using sherpa.astro.plot.ModelPHAHistogram,
+        # rather than a session._plotobj value that is also set in
+        # session._plot_types, so it doesn't get changed bu set_xlog etc.
```
wmclaugh, line 3250: bu -> by

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -2954,3 +2956,609 @@ def test_pha1_plot_multiple_args(clean_astro_ui, basic_pha1):
...
+    plot()
+    p1 = pdata()
+    if plotfunc == 'fit':
+        if is_int:
+            assert p1.dataplot.histo_prefs['ylog'] == answer
+            assert p1.modelplot.histo_prefs['ylog'] == answer
+        else:
+            assert p1.dataplot.plot_prefs['ylog'] == answer
+            assert p1.modelplot.plot_prefs['ylog'] == answer
+    elif plotfunc in ['resid', 'ratio', 'delchi']:
```
wmclaugh, line 3168: Will the assertions ever change here? If not, would it be more efficient to remove the 
 
    elif plotfunc in ['resid', 'ratio', 'delchi']:
        assert p1.plot_prefs['ylog'] == answer

since the same code is executed in the default/fallthrough case.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: Fixed

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: Fixed.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: It actually isn't the same as the fall through case. since this check needs to be run before the is_int check (ie if you take it out the test fails). I've added a comment to note this.

DougBurke: I perhaps should have added it to other tests that use the same ordering, but there's unfortunately a lot of similar-but-not-quite-similar-enough code that I felt it worth abstracting over (since I've found that just gets really hard to follow).

DougBurke:
Sorry for the delay but I was out at the Doctor's this morning

anetasie:
I tested the code in the branch and the bug is fixed. I've tried all the plot_ options with 1 and 2 data sets. Here is part of the relevant history of the session using version:    `4.12.2+9.g6c31af1a` on MacOS with Python 3.8

```
plot_fit(2)
notice(None,0.5)
notice(7.0, None)
fit()
plot_bkg()
plot_bkg(2)
notice(7.0, None)
plot_data(2)
notice()
plot_data()
plot_data(2)
notice(0.5,7.)
plot_data()
plot_data(2)
notice()
notice_id(2,0.5,7.)
plot_fit(2)
notice_id(2)
plot_fit(2)
notice_id(2,0.5,7.)
plot_fit(2)
set_xlinear()
plot_fit(2)
plot_fit(1)
ignore(None,0.5)
ignore(7.0,None)
fit()
plot('fit 1','fit 2')
plot('fit1','fit2')
plot("fit", 1, "fit", 2)
set_xlinear()
plot("fit", 1, "fit", 2)
set_xlog()
plot("fit", 1, "fit", 2)
plot("data", 1, "data", 2)
plot_model()
plot_source()
plot_arf()
set_ylog()
plot_arf()
set_ylinear()
plot_arf()
plot_data()
plot_data(2)
plot("data", 1, "data", 2, ylog=True)
plot_data(2)
plot("data", 1, "data", 2, xlog=False, ylog=True)
plot_data()
set_ylog()
plot_data()
plot("data", 1, "data", 2, xlog=False, ylog=True)
plot_data()
```

DougBurke:
@anetasie - just to confirm, you were using a PHA dataset for this set of plots?

anetasie:
@DougBurke Yes, I have tested the PHA data sets only. I did not try the other types of data.

dtnguyen2 (commented):
Is anyone seeing this?:
```
(base) [dtn@devel12 sherpa]$ git pull https://github.com/DougBurke/sherpa.git fix-981
From https://github.com/DougBurke/sherpa
 * branch              fix-981    -> FETCH_HEAD
Already up to date.
(base) [dtn@devel12 sherpa]$ python setup.py develop
  File "setup.py", line 38
    ), file=sys.stderr)
           ^
SyntaxError: invalid syntax
(base) [dtn@devel12 sherpa]$ 
```

Oops, I forgot to type:
```
source activate sherpa
```
before compiling

dtnguyen2 (approved):

DougBurke:
Ta @dtnguyen2

Issue #987, opened 2020-10-23
Fix typo in the error handling of citation


DougBurke:
```
There was a problem retireving the data from Zenodo:
Zenodo has no information for version 4.12.2.
```

should probably say `retrieving`. This only happens for "error" conditions so is not a show-stopper.

PR #1000, opened 2020-10-29
Improve sherpa.citation (fix #994 #987)

### LINKED ISSUE #994: sherpa.citation() doesn't quite work but sherpa.citation('latest') does
Thanks to `hysterical raisons` the new-to-Sherpa-4.12.2 citation function is slightly borked

```
>>> import citation
>>> sherpa.citation()
```

isn't able to pickup the data from Zenodo, because it's missing a `created` field (and possibly others, but that's where it's complaining).

The following does work:

```
>>> sherpa.citation('latest')
```

Note to self: to find

```
>>> import sherpa
>>> import logging
>>> logging.getLogger('sherpa').setLevel(logging.DEBUG)
>>> sherpa.citation('temp-file.txt')
Unable to find metadata: 'created'
```

DougBurke:
# Summary

Fixes the sherpa.citation() command with its default argument (issue #994) and adds release 4.12.2 to the hard-coded list of releases. A typo in a warning message was fixed (#987).

# Details

I had not realised Zenodo was doing pagination and, because I also had not set a "most-recent" order on the query the 4.12.2 response wasn't being included in the first set of results, and so wasn't being found. We now iterate through the pages (working around https://github.com/zenodo/zenodo/issues/1662) if the requested version isn't found (and the search order has been changed so that the latest version is returned first, which means the paging mechanism shouldn't be needed as long as the hard-coded list is kept up to date).

I've added developer documentation describing how to update the sherpa._get_citation_hardcoded routine to add new releases, and added the `scripts/make_zenodo_citation.py` script which will print out the needed update to stdout.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -119,6 +119,26 @@ can be activated and Sherpa can be build from source.
 How do I ...
 ============
 
+Update the Zenodo citation information
+--------------------------------------
+
+The :py:func:`sherpa.citation` function returns citation information
+taken from the `Zenodo records for Sherpa <https://doi.org/10.5281/zenodo.593753>`_.
+It can query the Zenodo API, but it also contains a list of known
+releases in the ``sherpa._get_citation_hardcoded`` routine. To add
+to this list (for when there's been a new release), run the
+``scripts/make_zenodo_release.py`` script with the version number
+and add the screen output to the list in ``_get_citation_hardcoded``.
+
+For example, using release 4.12.2 would create (the author list has been
+simplified)::
+
+  % ./scripts/make_zenodo_release.py 4.12.2
+      add(version='4.12.2', title='sherpa/sherpa: Sherpa 4.12.2',
+          date=todate(2020, 10, 27),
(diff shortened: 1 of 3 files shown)


DougBurke:
The tests are marked as failed because of coverage checks only, as the Travis tests don't use the --runzenodo flag

hamogu (approved):
Would it make sense to wait and get 4.13 in here, too? I though that @Marie-Terrell  has already branched that off. Since this PR certainly won't make it into 4.13, there is no harm in waiting a few more days until that's on zenodo and hardcode it here, too.

Of course, that could just as easily done in the follow-up PR, so I'm approving this regardless and leave it up to @DougBurke if he wants to wait or get this in now.

DougBurke:
Let's have a race between getting this merged or getting 4.13.0 out the door so we can hard-code the Zenodo values!

hamogu:
I think @wmclaugh is holding off merging new stuff until 4.13 is out, so it's a race this PR is guaranteed to loose!

DougBurke:
I've added the 4.13.0 hard-coded data.

PR #1001, opened 2020-10-29
Allow command-line arguments for sherpa_test

DougBurke:
# Summary

Allow command-line arguments to be passed to the sherpa_test script. This allows running optional tests (e.g. the `--runzenodo` argument) and to configure the pytest configuration (e.g. to run coverage checks with `--cov sherpa`).

# Details

Passing the command-line arguments through to the pytest machinery was easy. It allowed me to turn on coverage checks for the `sherpa_test` calls made in our Travis runs, which means a small bump in the test coverage (since these tests tend to be run without all the optional modules, so we get to check the fall-through code paths). Note that the way the `travis/test.sh` script is set up we get either a coverage run from `sherpa_test` or from the `python setup.py test` route, not both with the same setup. [with the change to GitHub actions we don't appear to see this bump so it may be that some other test was added which covered these cases]

The code has been rebased to apply to the GitHub actions code (fortunately the new version is very similar to the travis version so the update was easy). Of couse it now looks like the coverage has decreased, but that really doesn't make much sense.And now it's back to no change...

I did try some experiments with switching the tests to use `pytest` rather than `python setup.py test` but this fails as the build can't find the _utils module (a longstanding oddity that is easy to trigger with CI but harder to reliably diagnose manually...). So I dropped those changes.

### DIFF
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -43,7 +43,10 @@ cd /home
 sherpa_smoke ${smokevars} || exit 1
 
 # Run regression tests using sherpa_test
-if [ ${TEST} == package ] || [ ${TEST} == none ];
-    then cd $HOME
-    sherpa_test || exit 1
+if [ ${TEST} == package ] || [ ${TEST} == none ]; then
+    cd $HOME;
+    conda install -yq pytest-cov codecov;
+    # This automatically picks up the sherpatest modile when TEST==package
+    sherpa_test --cov sherpa --cov-report term || exit 1;
+    codecov;
 fi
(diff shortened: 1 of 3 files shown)


DougBurke:
This was rebased to get GH Actions.

DougBurke:
Rebased to remove the replacement of conda by pip to get the coverage packages (as I've noted issues in other CI-related changes with this logic so let's not change too many things here).

Marie-Terrell (approved):
These changes look good to me.

wmclaugh (approved):

Issue #996, opened 2020-10-27
clobber=True and the show_xxx style commands


DougBurke:
With 4.12.2, if you use `clobber=False` (the default) and set `outfile` (or `filename` or ...) to the name of an existing file in the routines that use the pager (so show_xxx and sherpa.citation) then you get an error. You should get an IOErr, reporting that the file exists and clobber is False, but instead you get an error 

```
NameError: name 'IOErr' is not defined
```

It's a simple fix (I've got it in as part of #991 for now)

DougBurke:
I've broken out the fix from #991 into #1002

PR #634, opened 2019-06-19
Added invitation for native software citation

ivalaginja:
# Summary

Improve information on how to cite Sherpa in the CITATION file and with a new method, sherpa.citation, which will display the citation information provided by Zenodo.

# Details

The sherpa.citation method can either query Zenodo to get the data or look up a hard-coded table (stored in the sherpa module) for past releases. The output is of the form

 - what version of Sherpa you are using
 - the Zenodo citation or a message saying it couldn't be found (the version isn't known, the Zenodo API has changed and no-longer returns the information we are expecting, there is no internet connection, or a glitch in the download)
 - information on how to cite vresions, linking to the Zenodo help page
 - Two "general purpose" references to Sherpa

# Example

If you were using a released version then the citation method would return something like

```
>>> sherpa.citation()
You are using Sherpa 4.12.1.

Sherpa 4.12.1 was released on July 14, 2020.

@software{sherpa_2020_3944985,
  author       = {DougBurke and
                  Omar Laurino and
                  wmclaugh and
                  dtnguyen2 and
                  Marie-Terrell and
                  Hans Moritz Günther and
                  Jamie Budynkiewicz and
                  Aneta Siemiginowska and
                  Tom Aldcroft and
                  Christoph Deil and
                  Brigitta Sipőcz and
                  Katrin Leinweber and
                  Todd},
  title        = {sherpa/sherpa: Sherpa 4.12.1},
  month        = jul,
  year         = 2020,
  publisher    = {Zenodo},
  version      = {4.12.1},
```

If you are using a random commit then you'll see something like the following, except that the response depends on whether the version number exists on Zenodo (in which case you will get that information - ie the `@citation{sherpa_...}` section from above), or a warning that the version isn't known, such as

```
>>> sherpa.citation()
You are using Sherpa 4.13.0+442.g38fb7ea5 (it is not a released version).

There was a problem retireving the data from Zenodo:
Zenodo has no information for version 4.13.0.

Please review the Zenodo Sherpa page at
       https://doi.org/10.5281/zenodo.593753
to identify the latest release of Sherpa. The Zenodo page
https://help.zenodo.org/#versioning provides information on how to
cite a specific version.

If you want a general-purpose citation then please use either of the
following, kindly provided by the SAO/NASA Astrophysics Data System
service:

@INPROCEEDINGS{2001SPIE.4477...76F,
       author = {{Freeman}, Peter and {Doe}, Stephen and {Siemiginowska}, Aneta},
...
```

# Background

[...]

PR #1002, opened 2020-10-29
Fix error with clobber=False for paging - issue #996

DougBurke:
# Summary

Fix an error with clobber=False when the output file exists for several paging commands (e.g. show_data and sherpa.citation). Instead of getting a Sherpa `IOErr` being raised a NameError was being raised.

# Details

The problem was when the output file exists and clobber=False (the default value). Instead of getting an IOErr you'd get a NameError since IOErr wasn't in scope. Adding an import of `IOErr` fixes the problem. A test has been added to cover this case.

The bug was introduced in #634 when the paging code was moved from sherpa.ui.utils to sherpa.utils.

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -40,9 +40,8 @@ import numpy.fft
 # Note: _utils.gsl_fcmp and _utils.ndtri are not exported from
 #       this module; is this intentional?
 from sherpa.utils._utils import hist1d, hist2d
-from sherpa.utils import _utils
-
-from sherpa.utils import _psf
+from sherpa.utils import _utils, _psf
+from sherpa.utils.err import IOErr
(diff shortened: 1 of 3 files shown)


DougBurke:
I don't understand why the code coverage test is marked as failing here.

DougBurke:
Rebasing to pick up GitHub Actions.

DougBurke:
So, I don't understand this as this only consists of tests. Did the problem get fixed in another PR?

DougBurke:
Aha - it's the

    from sherpa.utils.err import IOErr

line that fixes the problem. I'll update the description.

hamogu (approved):
That was easy to review...

DougBurke:
Thx @hamogu

Issue #997, opened 2020-10-27
problems with save_all and iterated fit


DougBurke:
If you have `set_iter_method("sigmarej")` [also 'primini. but you have many more serious problems than this] then the method settings for the sigmarej method are not written out correctly by `save_all`.

It's an easy fix (I have it as part of #991 for now)

PR #1003, opened 2020-10-29
Fix serialization of iter method data - issue #997

DougBurke:
# Summary

If set_iter_fit_method has been called with a value other than 'none' then the output of save_all would be incorrect for the options for the iter-fit method.

# Details

If an iterated-fit statistic was set then the serialization of the option data was wrong - ("%s", %s) rather than key and value. A test case has been added.

This was found when reviewing lint warnings generated with flake8.

### DIFF
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2015, 2016, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2015, 2016, 2019, 2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


DougBurke:
Rebased for GH Actions

sherpa/astro/ui/serialize.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2015, 2016, 2019  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2015, 2016, 2019, 2020  Smithsonian Astrophysical Observatory
```
hamogu, line 2: Needs a 2021 now...

sherpa/astro/ui/serialize.py
```diff
@@ -654,7 +654,7 @@ def tostatement(key, val):
         # is probably what we want but is a change, so leave
         # for now.
         # return 'set_iter_method_opt("{}", {})'.format(key, val)
```
hamogu, line 656: This comment can go away. No need to keep commented return statements around.

hamogu (changes requested):
Two minor things details. Ready to merge after that.

sherpa/astro/ui/serialize.py
DougBurke: The joy of ole PRs...

sherpa/astro/ui/serialize.py
DougBurke: I've actually made a small change to the existing code - use of a f-string - and cleaned up the comment line (to point out a difference to a different `tostatement` routine in this module). There should be no significant functional change.

DougBurke:
Ta for the review @hamogu

hamogu (approved):
Merge when CI passes.

PR #1006, opened 2020-10-29
Add missing collection-API methods to Meta

DougBurke:
# Summary

Improve the Meta API to better-match the collections API.

# Details

The idea was to remove a warning identified by flake8 (the use of .has_key) but this method has been left in for the moment. Instead I've added methods so that it more-closely follows the collections API (e.g. len and `__contains__`).

One reason for this PR is that I have to ask whether we really need the Meta object, as all it is is a dictionary. The only difference is that it has a FITS-style `__str__` method, which could be handled by a function, and an odd copy routine (which returns a copu of the underlying dict, not the Meta class).

As an example, the has_key method doesn't seem to be used in out code base.

### DIFF
--- a/sherpa/astro/io/meta.py
+++ b/sherpa/astro/io/meta.py
@@ -40,6 +40,15 @@ class Meta(NoNewAttributesAfterInit):
     def __setitem__(self, name, val):
         self.__header[name] = val
 
+    def __contains__(self, item):
+        return item in self.__header
+
+    def __len__(self):
+        return len(self.__header)
+
+    def __iter__(self):
+        return self.__header.__iter__()
+
     def get(self, name, default=None):
         return self.__header.get(name, default)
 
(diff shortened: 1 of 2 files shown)


hamogu:
A "blame" shows that it's all from the "initial commit" 7 years ago with minor changes after that (docs, a bug fix). I saw @DougBurke 's email on sherpa-dev. Looking at this code (at given my limited understanding of Sherpa internals), I vote to remove it and replace it's use with a standard dictionary in the codebase. If more things were meant to be done with it, it clearly did not happen, which implies that, if anything else was planned, we don't urgently need them.

DougBurke:
I'm closing this as we plan to remove the Meta class entirely in #1155

hamogu:
"meta class" I get it. Almost a "metaclass" but not quite.

PR #1010, opened 2020-11-02
Update Travis tests to cfitsio 4.49 and xspec 12.11.1

Marie-Terrell:
Xspec now requires fgsl and the fftw fortran wrappers. This now places a minimum gfortran build requirement for Xspec of ~5.x ([for Fortran 2008 C_LOC support](https://heasarc.gsfc.nasa.gov/lheasoft/issues.html)). 
This gfortran version requirement is an issue because the latest version of gfortran available in defaults for Conda on macOS is 4.8.5. A ticket was submitted to Anaconda to hopefully increase the gfortran support of macOS, but that may take a while (as not many people need it or they use Conda-forge). 
For now, we can proceed with adding Xspec 12.11.1 testing for Sherpa on Linux with future methods to be tested in later updates.
This PR switches some of the travis tests (some linux tests only) to use Xspec 12.11.1 and cfitsio 3.49. All other Travis tests have also been updated to use xspec 12.10.1s and cfitsio 3.49. 
The xspec_channel used for testing was also switched from "xspec/label/test" to "xspec/label/dev" for testing purposes and to avoid needing to rebase old PRs if the Conda packages were uploaded before this PR was merged. That is the case because the old Xspec packages used for testing do not have cfitsio pinned properly, so any old PRs using 12.10.1n for testing would attempt to pick up the cfitsio update and would run into library issues. This issue has been rectified for CIAO using automated pins through "run_exports" in the cfitsio and xspec-modelsonly recipes and brought here to resolve this issue moving forward. If we want, we can migrate "xspec/label/test" back to "xspec/label/dev", but any old PRs may need to be rebased and the new packages on the "xspec" Conda channel would need a "dev" label.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ jobs:
   fast_finish: true
   include:
     # macOS build
-    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER="3" XSPECVER="12.10.1n" TRAVIS_PYTHON_VERSION="3.7"
+    - env: INSTALL_TYPE=develop FITS="astropy" TEST=submodule MATPLOTLIBVER="3" XSPECVER="12.10.1s" TRAVIS_PYTHON_VERSION="3.7"
       os: osx
     # Barebone build to check tests pass when most of the tests must be skipped.
     # We need to use TEST=none to remove the test-data submodule.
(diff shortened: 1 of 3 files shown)


DougBurke:
@Marie-Terrell - is this PR now obsolete because if #1035 (in that the changes in this PR have been made to the equivalent #1035 actions) or do we still need to ubdate the GitHub actions to use the newer cfitsio/xspec versions (so I guess we'd still close this but would need another PR)?

Marie-Terrell:
This one would now be obsolete. The equivalent changes have been made on that commit.

PR #1012, opened 2020-11-10
Tests: allow test_ui tests to be run with pytest-xdist

DougBurke:
# Summary

Allow the tests to be run with pytest-xdist.

# Details

The pytest-xdist runner requires that the sequence sent to the pytest parametrize function be repeatable (so that the same
list is used when generating the tests with multiple processes). It turns out that set doesn't meet these requirements, so change from a set to a sorted list. At some-point relatively recently I had added a PR which used sets for the parametrize list - which is a perfectly cromulent thing to do - but it broke the xdist functionality, but I hadn't worked out what the reason was until now.

This does not change the tests that are run, just their potential ordering, and allows you to say

    pytest -n 6

(or some other value) if you have installed pytest-xdist.

### DIFF
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -96,7 +96,7 @@ def setup_covar(make_data_path):
 
 
 @requires_data
-@pytest.mark.parametrize('stat', set(ui.list_stats()) - RIGHT_STATS)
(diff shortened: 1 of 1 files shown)


DougBurke:
Just rebased to run with GitHub Actions. There's been no change to the code in this PR.

hamogu (approved):

PR #1016, opened 2020-11-13
Jupyter notebook representation not ideal if model components don't have unique names  - issue: #1013

### LINKED ISSUE #1013: Jupyter notebook representation not ideal if model components don't have unique names
It's not ideal to have the parameters listed twice under the same name, just because two model have the same name. Might be confusing after the fit when the parameters of the two models are actually different. 

Screen Shot 2020-11-12 at 4 04 42 PM

Not ideal, but also clearly not a dealbreaker. Maybe that's a good incentive for the user to ensure unique names!

@DougBurke

hamogu:
Summary
=======
Fix an error in the HTML display of a model (used in the notebook) when
two model components have the same name.

Details
-------
The previous implementation interated over paramters and the modelname of
the parameter can be ambiguous. Instead, we now iterate over models and then
other paramters within eahc model. That way, we can start a new block in the
html table for each new model, without relying on detection of a name change.

fixes #1013

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -1054,18 +1054,34 @@ def _wrapobj(obj, wrapper):
 
 # Notebook representation
 #
+def modelcomponents_to_list(model):
+    if hasattr(model, 'parts'):
+        modellist = []
+        for p in model.parts:
+            modellist.extend(modelcomponents_to_list(p))
+        return modellist
+    else:
+        return [model]
+
+
 def html_model(mdl):
     """Construct the HTML to display the model."""
 
     # Note that as this is a specialized table we do not use
     # formatting.html_table but create everything directly.
     #
-    nrows = defaultdict(int)
-    for par in mdl.pars:
-        if par.hidden:
-            continue
-
-        nrows[par.modelname] += 1
+    complist = []
+    nrows = []
+    for comp in modelcomponents_to_list(mdl):
+        this_comp_nrows = 0
+        for par in comp.pars:
+            if par.hidden:
+                continue
+            else:
+                this_comp_nrows +=1
+        if this_comp_nrows > 0:
+            complist.append(comp)
+            nrows.append(this_comp_nrows)
 
     out = '<table class="model">'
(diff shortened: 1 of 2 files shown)


DougBurke:
I'm out today so it'll have to wait until next week.

hamogu:
I note that Travis passed, it's codecov that failed for a non-measurable change in coverage, so this is really waiting for a review, not for me to address failures.

hamogu:
@DougBurke : Ping for review.

sherpa/models/model.py
```diff
@@ -1076,56 +1092,47 @@ def html_model(mdl):
...
-        if linked:
-            out += "<td>linked</td>"
-        else:
-            out += '<td><input disabled type="checkbox"'
-            if not par.frozen:
-                out += ' checked'
-            out += '></input></td>'
 
-        out += addtd(par.val)
+    for mcount, n, comp in zip(range(len(complist)), nrows, complist):
```
DougBurke, line 1096: Would (the untested)

    for mcount, (n, comp) in enumerate(zip(nrows, complist)):

work / be better?

DougBurke:
It was a bit annoying to track down what the code was doing, so I tried your test code and got

```
sherpa In [34]: h = m._repr_html_()

sherpa In [35]: iodd = h.find('<th class="model-odd" ')

sherpa In [36]: h[iodd:iodd+120]
Out[36]: '<th class="model-odd" scope="rowgroup" rowspan=6>name</th><td>fwhm</td><td><input disabled type="checkbox" checked></inp'

```

So I can see the priblem here (`rowspan=6`).

sherpa/models/model.py
hamogu: it looks cleaner!

hamogu:
Next time, I'll add a screenshot!

DougBurke:
There are times it can be useful to add the test in one commit and then a change in the next as you can see the difference. Not sure it would have helped me in this case though!

DougBurke (approved):
LGTM

DougBurke:
Sorry for the delay in getting this looked at

PR #1017, opened 2020-11-15
reword remark on normalization of Lorentz function

hamogu:
I was confused by the text as that implies that an additional normalization step takes place. It sounds as if Sherpa calculates the integral, when in fact, the Lorentz function already integrates to 1 for ampl = 1. That's why there is the factor 2 pi in there, which ensures that integral f(x) dx = 1.

@DougBurke As a side note: This is a case that would really profit from writing the equation in LaTeX and activating MathJax for the sphinx build. I was very tempted to just do that and see what happens, but then I realized the the ahelp pages are now also build from the doc strings, and I have no idea if they can deal with that, so I figured I just ask before I do anything about it.

### DIFF
--- a/sherpa/astro/models/__init__.py
+++ b/sherpa/astro/models/__init__.py
@@ -569,14 +569,14 @@ class Lorentz1D(RegriddableModel1D):
     -----
     The functional form of the model for points is::
 
-        f(x) =                A * fwhm
+        f(x) =              ampl * fwhm
                --------------------------------------
                2 * pi * (0.25 * fwhm^2 + (x - pos)^2)
 
-           A = ampl / integral f(x) dx
-
     and for an integrated grid it is the integral of this over
     the bin.
+
+    The area under the function as defined above is 1 if ampl is 1.
     """
 
     def __init__(self, name='lorentz1d'):


DougBurke:
I have mixed feelings about docstrings, as do we want them to be usable from the python prompt or from a web page? I do know that supporting LaTeX for these formulas will probably not complicate the text for users of the python prompt, but (I think) it's a slippery slope. What is the AstroPy approach?

hamogu:
Astropy uses LaTeX for formulas in models in the "notes" section of the docstring: https://docs.astropy.org/en/latest/api/astropy.modeling.functional_models.Lorentz1D.html#astropy.modeling.functional_models.Lorentz1D

Most users can read plain text LaTeX just as well as plain text formulas.
(Actually, in astropy there was some attempt to autogenerate the Latex from the python expression, not sure how far that went. It's certainly overkill here.)

DougBurke:
I suggest trying to add latex support here so we can see what it looks like. There's two parts

a) the text itself
b) any change needed to the read-the-docs sphinx build (which may not be necessary)

hamogu:
I will add some latex to try it out (I've done that before, so it should not take me long to find the switch in sphinx again), but I will do it in a separate PR. This PR is about a simple rewording of the meaning of "normalized" and I'd like @DougBurke or @anetasie to review and merge it, independently of what we eventually decide about using latex or not.

DougBurke (approved):

hamogu:
@wmclaugh : I originally requested @DougBurke @anetasie as reviewers here. That was supposed to be an "or" request (see https://github.com/sherpa/sherpa/pull/1017#issuecomment-729251181) and since @DougBurke has reviewed and approved this, I've removed @anetasie from the list list of requested reviewers and ask for this to be merged.

DougBurke:
@hamogu - just to mention that this was another PR lower-down my priority list (but not as low down as #1034) for merging.

PR #1022, opened 2020-12-03
plot_model_components adds the response if needed

DougBurke:
The `plot_model_component` function will now automatically add the instrument response (for `DataPHA` objects) unless the `add_response` flag is set to `False` (it defaults to `True`). There is no change for any other data type.

No tests have been added yet.

# Example

For a source model of `xsphabs.gal * (powlaw1d.pl + xsgaussian.line)`:

```
plot_fit()
plot_model_component(gal * pl, alpha=0.5, overplot=True)
plot_model_component(gal * line, alpha=0.5, overplot=True)
```

creates

sherpa

Without this we would get the following plot, where both components appear to be 0 [this depends on the model but 0 is a common option, and is because the response isn't being included):

sherpa2


DougBurke:
Rebased to get GH Actions.

DougBurke:
Rebased to move some simple changes first. They'll be converted to separate PRs.

DougBurke:
Rebased as #1057 and #1058 have been merged.

anetasie (approved):

anetasie:
@DougBurke I tried this PR and the plots look good. The only adjustments I had to do was to the limits when I use the log scale, but that's expected.

DougBurke:
@anetasie - I'm not ready for this to be merged yet, as we have questions about the UI to answer, but fortunately there is a conflict so it can't be merged as is.

What do you think about the "flag" that is added to `plot_model_components` as I don't really like it, but there's not a great set of options?

We have to think about trying not to break existing "correct" code, where a user has added the response themselves (or we may have a situation with multiple responses and the user wants to select a particular response or maybe there's a really-complicated model with different responses on it), versus making it easy for the general case of "there's only one response which we can add automatically".

Instead of adding a flag to the call, which users would have to specify when they have included the response themselves, we could try and identify if the model contains a response. I think - for this kind of a routine - this is the better approach, except that we don't have an easy way to ask if a model contains a response or not (there are things we can check but it's not-very-Python-OO code).

anetasie:
@DougBurke I vote for a non-elegant approach of adding a flag. You can add an elegant approach when  we know how to check if the model has a response, which I guess is not a trivial implementation.

hamogu:
I'm against adding a flag now if we plan to remove it later, rather hold this for the next release. That's just from the UI level. I'll try to look through the code now and see if I can make a better suggestion, but here are some thoughts:

- This really is an UI level function, a convenience (That I will use all the time!) for making good plots. It's not one of the fundamental building blocks you use to make those plots. You can make the plot above entirely by hand, just need to code along the lines of (for a model `a * (b + c)`):
```
x = get_plot().x
plt.plot(x, rsp((a * b))(x))
plt.plot(x, rsp((a * c))(x)
```
- That means that we want to code that "just works" in 90% of the cases and at least does not do something bad without throwing an error 99% of the time. If you "know what you are doing" (as @DougBurke put it above for e.g. using difference responses in the same model), then you can build up your own plot line by line if you really need to.
- Can't we just check something like:
```
def flatten_sherpa_model(model):
    if hasattr(model, 'parts'):
        modellist = []
        for p in model.parts:
            modellist.extend(flatten_sherpa_model(p))
        return modellist
    else:
        return [model]

add_resposnse=True
for m in flatten_sherp_model(full_model_we_want_to_plot):
    if isinstance(m, RSPModel):  # Or whatever it's called
       add_response = False
       break
```
If you don't set a response, you get the default one, if you set at least one you don't get any. If you want to plot without a response, go and write your own loop.

DougBurke:
@moritz - as I mentioned above, there's no single one class we can key off to identify a "this model includes a response". I think we;d have to check three classes (and as it's extensible we would potentially have to change the logic if someone came along and added a spectro-polarimetry model case, for instance, but that would not be an issue for > 90% of the users).

If Sherpa were to be designed now I'd think we'd want to be able to ask a model things like "what are the units of the independent axis" so we could identify the channel vs energy/wavelength case (and hence response included or not), but we're not close to that at the moment!

hamogu:
We can test for three classes in just one line of code `isinstance(m, (RSPModel, RMFModel, ThirclassIdonotknowabout))`. If users extend that with a subclass for their own purpose, it's likely that they inherit form one of the existing classes and the `isinstance` check would still work. since this is a convenience method, I don't think we have to worry about users who implement their own response class without inheriting from our classes, just by duck-typing.

A similar argument can be made against checking the unit. What if I make a response in frequency space (which sherpa itself does not currently support)?

**begin rant**
I really want to avoid an extra parameter if we can. Sherpa has (at least it feels like that) more functions and ways to do the same thing than numpy or e.g. matplotllib Even after contributing to the Sherpa  codebase a bit, I still constantly discover new functions, new parameters, etc. It often does not feel "coherent" and the fact that our docs are organized by function (docstrings, ahelp "see alos"), to just skim the surface (readthedocs text) or in threads (ciao website) instead of systematically explain all options does not help. 

Threads are good as examples, but if you want to do something just slightly different, it's hard to find out how because you don't know where to start looking. 
**end rant**

moritz:
wrong moritz, it seems :-)

DougBurke:
OOps - sorry @moritz

DougBurke:
There were two parts of this PR originally - ability to display a model component without having to manually add a response, and the plot_model/source_components call, which splits up a model expression.

The first part was broken out to #1184 and merged.

The second part is now being looked at in #1679 with a significantly-less complex approach than used here. So I'm closing this one.

PR #1028, opened 2020-12-16
Update region lib code to correct build issues with gcc 7.3.0 compile…

wmclaugh:
Corrects issue building with gcc 7.3.0+ compilers with the -Werror=format-security compilation flag. Without the fix, compilation will fail with an error such as:

    region_repr.c: In function 'regWriteAsciiRegion':
    region_repr.c:130:13: error: format not a string literal and no format arguments [- 
   Werror=format-security]
                 snprintf(shapeName, maxlen, Shape->name);
                 ^~~~~~~~
  
This fix was included in CIAO 4.13 as SL-195.


DougBurke:
I've checked and master branch fails with

```
% export CFLAGS="-Werror=format-security"
```

but this passes. Should the `extern/region-4.11` directory be moved to `extern/region-4.13` or would this require more changes to match the CIAO 4.13 region code?

I guess the question is is what does the version number on the `extern/foo-version` directories mean to us?

```
% ls -1 extern
built
compile
config.guess
config.log
config.status
config.sub
configure
configure.ac
depcomp
fftw-3.3.8
grplib-4.9
install-sh
libtool
ltmain.sh
m4
Makefile
Makefile.am
Makefile.in
missing
region-4.11
stklib-4.11
wcssubs
```

DougBurke:
I'm not sure you need a SDS approval for this one, but yo can get one from me anyway :-)

DougBurke (approved):

DougBurke:
One question is that there were un-related changes removing some text files from the fftw code. As we don't really "care" about these (I'm not sure whether we install them or not) I just wanted to check whether these changes were intended here?

wmclaugh:
The removal of the text files in extern/fftw-3.3.8/doc/FAQ/ was unintentional. The deletions were 'untracked changes'. My assumption is that they are not necessary as they are not included in the fftw-3.3.8.tar.gz (available via wget https://www.fftw.org/fftw-3.3.8.tar.gz).

DougBurke:
@wmclaugh - thanks. I agree it's okay to retain the deletions in this PR, I just wanted to check.

PR #1030, opened 2020-12-16
Update lev3fft test to use PHYSICAL not WCS

DougBurke:
# Summary

Update a 2D image test to use physical rather than WCS coordinates for the fit.

# Details

It is not clear why we use WCS for fitting, since it is known that WCS fitting isn't great, particularly when you have a convolution model like the PSF components used here.

This is being done to make some work on FWHM-related guess code easier (since without this change the guess in the updated code will return a FWHM in degrees (i.e. small) which messes up the fit).

Note that technically the new version is a "better" fit (the statistic value is smaller, 559 vs 592) but as the data is very ropey I'm not sure how much  weight we should give this.

We need https://github.com/sherpa/sherpa-test-data/pull/24 merged before we can accept this. I can't guarantee I've got that part right with the submodule. In fact I'm reasonably sure I haven't, but I can fix that once https://github.com/sherpa/sherpa-test-data/pull/24 is merged. The tests currently fail because of this botched change (thanks to things having changed sine the PR was made last year).

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -599,15 +599,15 @@ def test_stats_all(run_thread, fix_xspec):
 def test_lev3fft(run_thread, clean_astro_ui):
     tlocals = run_thread('lev3fft', scriptname='bar.py')
 
-    assert tlocals['src'].fwhm.val == approx(1.6304283807465337e-05)
-    assert tlocals['src'].xpos.val == approx(150.014516168296)
-    assert tlocals['src'].ypos.val == approx(2.6650092327584507)
-    assert tlocals['src'].ampl.val == approx(97.24051178672816)
-    assert tlocals['bkg'].c0.val == approx(0.02087374459610318)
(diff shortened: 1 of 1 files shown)


DougBurke:
Rebased to get GH Actions

dtnguyen2 (approved):
Once [sherpa/sherpa-test-data#24 ](https://github.com/sherpa/sherpa-test-data/pull/24) is merged, I verified the following fit
```
(sherpa-4.14-dev) [dtn@devel12 sherpatest]$ python lev3fft-bar.py
Datasets              = 1, 2
Statistic             = cstat
Initial fit statistic = 6483.84
Final fit statistic   = 558.987 at function evaluation 797
Data points           = 3307
Degrees of freedom    = 3302
Probability [Q-value] = 1
Reduced statistic     = 0.169287
Change in statistic   = 5924.85
   src.fwhm       1.48914     
   src.xpos       3142.86     
   src.ypos       4519.72     
   src.ampl       9.11268     
   bkg.c0         0.0118703   
```
are consistent with the changes  in the _test_astro.py_ file

Issue #875, opened 2020-08-06
Caruana, Guos algoritm for Gaussian guess function?


dtnguyen2:
While working on [PR#741](https://github.com/sherpa/sherpa/pull/741), I noticed that the underlying 1D ```guess``` algorithm for the Gaussian function is not very sophisticated (the location of the peak value is the position, the peak value is the amplitude and to find the fwhm one marches off from the peak value until half the peak value is found etc...) but it is probably good enough for most cases.  

However, Raphael is making good progress on MLE by using simultaneous fitting of crowded fields.  In his prototype, he no longer uses the ```wavedect``` positions as the initial positions for the fitted parameters since they sometimes prove to be unreliable but he uses the sherpa built-in ```guess``` function.  If ever the Gaussian guess has to be improved for MLE, I suggest we look into the _Caruana_ and _Guos_ algorithm.

Moreover, the current sherpa 2D Gaussian guess function does not calculate the two ```sigmas``` so one could in theory "sum up" the bins in the x and y axes to reduce the 2D Gaussian to two 1D Gaussians and apply the aforementioned algorithm.

DougBurke:
@dtnguyen2 - is https://content.sciendo.com/view/journals/jee/66/3/article-p178.xml a good reference? Fortunately it's open access.

dtnguyen2:
Yes, the reference [Caruana, Guos algorithm](https://content.sciendo.com/view/journals/jee/66/3/article-p178.xml) is a clear enough to implement from

DougBurke:
Just to note that I've been tracking down a difference between CIAO 4.12 and 4.13 and one of them was the change in behavior of the guess algorithm for the gauss1d.fwhm component. From https://github.com/sherpa/sherpa/commit/9c29df316488d9f3114ec3adc064d4c863c19baa (#741) the changes to `sherpa/utils/__init__.py.get_fwhm`

a) the code now goes to the right of the peak whereas it started from the left-edge before [there's no a-priori right way, and this version is at least better-connected to the peak whereas the old version would just stop at the first value found, even if not associated to the peak]

We could perhaps go from both sides and take min, max, or mean of the two widths.

b) the old version returned

            return 2.0 * numpy.abs(x[ii] - x_max)	

whereas the new version returns

            return 2.0 * ii

I think it should be returning the old version (so `2 * abs(x[ii] -xmax`), as `ii` is just an index. That is, instead of

    for ii, val in enumerate(y[y_argmax:]):
[...]

PR #1031, opened 2020-12-16
Update fwhm calculation

DougBurke:
# Summary

Update the estimation of FWHM for 1D profiles, and hence the guess method for Gauss1D and related routines. The 2D models use the same routine, so see these changes, but they are known to require more work.

# Details

The FWHM estimation used by the `guess` method of models is based on `sherpa.utils.get_fwhm`. This was changed in #741 but I believe it had an error, as discussed at https://github.com/sherpa/sherpa/issues/875#issuecomment-746378325

We add explicit tests of `get_fwhm` that show the problem (essentially that the index value is used rather than a value appropriate for the axis, although when the simple test fails you end up with the position of the maximum value).

I then apply an improvement suggested at https://github.com/sherpa/sherpa/issues/875#issuecomment-746678714 which improves some of the tests but not all of them.

I then rework the `get_fwhm` algorithm so that it calculates the HWHM for both sides of the peak and then

- if both are found, returns the sum
- if only one is found (as the peak is too close to one edge) then returns 2 * HWHM
- if neither is found (the data never drops to half the peak height) then return half the width of the X array as a heuristic

I then have to update the lev3fft test, but as discussed in that commit (@9b006f3) I strongly believe this will need to be redone because I would hope we accept #1030

I have added an explicit check for negative values and if found we just fall through to the default case. It would be nice to handle this case (for absorption features) but it is tricky to handle generically, so let's just be explicit here about the requirement of this only being intended for positive peaks.

Is this worth taking, or should we just implement #875 (although I think we want to add some of the functionality in this PR, to add tests).

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -599,15 +599,15 @@ def test_stats_all(run_thread, fix_xspec):
 def test_lev3fft(run_thread, clean_astro_ui):
     tlocals = run_thread('lev3fft', scriptname='bar.py')
 
-    assert tlocals['src'].fwhm.val == approx(0.0442234, 1e-4)
-    assert tlocals['src'].xpos.val == approx(150.015, 1e-4)
-    assert tlocals['src'].ypos.val == approx(2.66494, 1e-4)
-    assert tlocals['src'].ampl.val == approx(1.56384, 1e-4)
-    assert tlocals['bkg'].c0.val == approx(-1.51662, 1e-4)
+    assert tlocals['src'].fwhm.val == approx(1.6304283807465337e-05)
+    assert tlocals['src'].xpos.val == approx(150.014516168296)
+    assert tlocals['src'].ypos.val == approx(2.6650092327584507)
+    assert tlocals['src'].ampl.val == approx(97.24051178672816)
+    assert tlocals['bkg'].c0.val == approx(0.02087374459610318)
 
     fres = ui.get_fit_results()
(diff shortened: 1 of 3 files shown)


DougBurke:
Argh - this should have been made from @DougBurke not this repository...

DougBurke:
Rebased to pick up GH Actions.

dtnguyen2:
@DougBurke I was about to review this PR but it looks like there are two ```Conflicting files``` and they look quite complicated and they are probably all your doing so it would be nice if you can take a look at it.

Regarding the the absorption features, does taking the absolute values of the y-values generic enough?

DougBurke:
@dtnguyen2 - I've just rebased this to fix the conflicts.

One of the problems with the current algorithm is that it doesn't consider a "background" component (where background here means "anything other than the gaussian-like peak we are trying to identify"). So all we are doing is trying to find where the maximum peak falls to half its value. For an absorption feature we can identify the minimum value, but it's not guaranteed to be negative: e.g. think of a flat background of 100 counts per bin which falls to 60 counts per bin at the center of the absorption feature - here I would want to identify the FWHM as the point where we get back to 80 counts/bin, but without knowledge of the background level I just can't see how to do this. Hence I punted and made this a "only works for positive features" routine (well, I claim this is what the original code assumed as well, it just never documented this).

It would be nice to do better, but I think that's outside the level of the changes in this PR!

dtnguyen2 (approved):

PR #1034, opened 2020-12-19
Add contextmanager and docs on how to control the output level of sherpa

### LINKED ISSUE #1032: Controlling verbosity of fitting and error estimation in stand-alone sherpa
I'm using Sherpa in a script that fits hundreds of relatively small and simple datasets and I see a lot of output like this
```
gauss1d.pos lower bound:	-0.178148
gauss1d.pos upper bound:	0.178148
```

I would like to switch that off. After doing some code digging, the best solution I cam up with is this.
At the import section of my code:

```
import logging
sherpablog = logging.getLogger('sherpa')
old_level = sherpablog.level
```
And then in the place where I do the fitting:
```
sherpablog.setLevel(logging.WARNING)
my_fittter.est_error()
sherpablog.setLevel(old_level)
```
That seems pretty complicated and I only found it by reading through the source code. So, I have two questions: 

- Is there an easier way?
- I did not see it in the stand-along docs. Did I miss that?

If this is indeed the way to do it, then I'll look for a place in the docs where I can add that.

hamogu:
# Summary
Sherpa uses logging for much of its output, this adds a docs and a context manager for controlling the output level for a particular piece of code.

# More details 
In particular the fit and error estimation routines print a lot of stuff to the screen like fit results or the confidence range for every parameter as it's calculated. That's convenient for interactive use, but can clutter output in the script or notebook. Using python's logging infrastructure the level of logging can be adjusted directly, this PR adds docs for that and a context manager to change the logging level only for a few lines of code (e.g. the lines where fit or confidence are called.)

fixes #1032

### DIFF
--- a/sherpa/utils/logging.py
+++ b/sherpa/utils/logging.py
@@ -18,6 +18,7 @@
 #
 import logging
 import sys
+import contextlib
 
 
 def config_logger(name, level=logging.WARNING, stream=sys.stdout, template=None):
@@ -39,7 +40,7 @@ def config_logger(name, level=logging.WARNING, stream=sys.stdout, template=None)
     return logger
 
 
-class SherpaVerbosity():
+class SherpaVerbosity(contextlib.AbstractContextManager):
     '''Set the output logging level for sherpa as a context.
 
     This changes the logging level globally for all modules in sherpa.
--- a/sherpa/utils/logging.py
+++ b/sherpa/utils/logging.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2015  Smithsonian Astrophysical Observatory
+# Copyright (C) 2015-2020  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -37,3 +37,26 @@ def config_logger(name, level=logging.WARNING, stream=sys.stdout, template=None)
     logger.propagate = False
 
     return logger
+
+
+class SherpaVerbosity():
+    '''Set the output logging level for sherpa as a context.
+
+    This changes the logging level globally for all modules in sherpa.
+
+    Parameters
+    ----------
+    loglevel : string or int
+        New level for logging. Allowed strings are
+        ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and ``CRITICAL``
+    '''
+    def __init__(self, level):
+        self.level = level
+        self.sherpalog = logging.getLogger('sherpa')
+
+    def __enter__(self):
+        self.old = self.sherpalog.level
+        self.sherpalog.setLevel(self.level)
+
+    def __exit__(self, *args):
+        self.sherpalog.setLevel(self.old)
--- a/sherpa/utils/tests/test_logging.py
+++ b/sherpa/utils/tests/test_logging.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2020  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2020, 2021  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -25,19 +25,15 @@ from sherpa.astro import ui
 
 
 def test_logging_verbosity_contextmanager(caplog):
-    # Previous tests might have changed this to a non-default
-    # level already. So, we first make sure the root logger
-    # is set right.
-    sherpalogger = logging.getLogger('sherpa')
-    sherpalogger.setLevel('WARNING')
 
-    logger = logging.getLogger('sherpa.some_module')
-    logger.warning('1: should be seen')
-    assert len(caplog.records) == 1
+    with caplog.at_level(logging.INFO, logger='sherpa'):
+        logger = logging.getLogger('sherpa.some_module')
+        logger.warning('1: should be seen')
+        assert len(caplog.records) == 1
 
-    with SherpaVerbosity('ERROR'):
-        logger.warning('2: Should not be seen')
-    assert len(caplog.records) == 1
+        with SherpaVerbosity('ERROR'):
+            logger.warning('2: Should not be seen')
+        assert len(caplog.records) == 1
 
-    logger.warning('3: should be seen')
-    assert len(caplog.records) == 2
+        logger.warning('3: should be seen')
+        assert len(caplog.records) == 2
--- a/sherpa/utils/tests/test_logging.py
+++ b/sherpa/utils/tests/test_logging.py
@@ -25,6 +25,12 @@ from sherpa.astro import ui
 
 
 def test_logging_verbosity_contextmanager(caplog):
+    # Previous tests might have changed this to a non-default
+    # level already. So, we first make sure the root logger
+    # is set right.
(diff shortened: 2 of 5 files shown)


DougBurke:
Would it make sense to change `sherpa/conftest.py:hide_logging` to use this? In some ways it think it's a "too clever" idea but wanted to see what you think @hamogu

hamogu:
I think it's "too clever" in the sense that we know the fixture works and we'd only save 4 lines of code. Using a fixture vs. using the with statement within each test. I prefer the fixture, because it's a better separation between what is test infrastructure (the fixture) vs what is code to be tested (the code in the text function). 
I will however add a comment to the text fixture that, should it ever break and need to be redone for some reason, we can try to replace it with this.

sherpa/utils/tests/test_logging.py
```diff
@@ -0,0 +1,37 @@
...
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+import pytest
+import logging
+import numpy as np
+import sherpa
+from sherpa.utils.logging import config_logger, SherpaVerbosity
+from sherpa.astro import ui
+
+logger = logging.getLogger('sherpa.some_module')
```
DougBurke, line 26: A really pedantic point, but I think this should be declared within `test_logging_blah_blah` as there's no need for it to be a package global (and the fact we use a "random" name for the logger could potentlally-but-incredibly-unlikely cause confusion if we ever add more tests to this file).

sherpa/utils/logging.py
```diff
@@ -37,3 +37,26 @@ def config_logger(name, level=logging.WARNING, stream=sys.stdout, template=None)
     logger.propagate = False
 
     return logger
+
+
+class SherpaVerbosity():
```
DougBurke, line 42: Should this derive from contextlib.AbstractContextManager (now we're safe to require python 3.6)? To be honest it's not obvious from the documentation what benefits we get, but I do like the implied contract/documentation you get from something like this.

sherpa/utils/tests/test_logging.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_logging_verbosity_contextmanager(caplog):
+    logger.warning('1: should be seen')
+    assert len(caplog.records) == 1
+
+    with SherpaVerbosity('ERROR'):
```
DougBurke, line 32: Ideally I'd like to also see a test with `SherpaVerbosity(logging.ERROR)`  as well (I know that we're just passing on responsibility to `setLevel` so we should be fine, but just to make sure).

DougBurke (commented):
I've made a few comments - none of them are required, but I'll hold off on approval until @hamogu has shot them down.

sherpa/utils/tests/test_logging.py
hamogu: We can add as many tests as we like, but this one makes on sense to me. That's the responsibility of Python itself. I'm already using the most advanved input method (a string). `logging.ERROR` is literally just an int.

sherpa/utils/logging.py
hamogu: Sure. We can do that. Maybe some benefit will be added in later Python versions...

hamogu:
I'm so confused. This passes locally:
```
platform darwin -- Python 3.7.7, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: /Users/hamogu/code/sherpa, configfile: pytest.ini
plugins: xvfb-2.0.0
collected 1 item                                                               

sherpa/utils/tests/test_logging.py .                                     [100%]

============================== 1 passed in 0.60s ===============================
```
In CI, we have the same pytest version. Why does it fail on CI?

hamogu:
There is so much that goes into this seemingly simple test (we test code that resets the logging level, in a test runner that resets the logging level and also redirects the logging output) and I just can't figure out why it passes on my computer, but not on CI. Any ideas anyone? If not, can we maybe not have a test here since writing the test has already taken me much more time than writing the code?

hamogu:
Mmm. It depends on which other tests where run before. There must be some other test that changes the logging level from the default and does not reset it. That's a hard-t-debug thing (in particular since I've got 273 failing locals tests form the multiprocessing that's not set up correctly in my development environment and that makes it very tedious to experiment with the full test suite) . So, better set the level by hand before doing this test.

DougBurke:
One of the reasons I added the `hide_logging` fixture was to try and fix up all the cases where we forget to reset the logging level!

DougBurke:
So, in `sherpa/conftest.py`:

```
@pytest.fixture(autouse=True)
def check_logging():
    """Will this run after hide_logging?"""

    logger = logging.getLogger('sherpa')
    olvl = logger.getEffectiveLevel()
    yield
    nlvl = logger.getEffectiveLevel()
    if nlvl != olvl:
        raise RuntimeError(f"log level change: {olvl} to {nlvl}")
```

is a way to find out those tests that don't clean themselves up (or fix up a previously-borked state...). For a "full" build (only ds9 not available) build it reports the following case:

```
sherpa/astro/utils/tests/test_smoke.py .E...                                                                           [ 36%]
```

This can be addressed by adding `hide_logging` to all the tests in `test_smoke.py`, after which the fix in 13b597c isn't needed (although my suggested tweak to that may be technically safer).

sherpa/utils/tests/test_logging.py
```diff
@@ -25,6 +25,12 @@
 
 
 def test_logging_verbosity_contextmanager(caplog):
+    # Previous tests might have changed this to a non-default
+    # level already. So, we first make sure the root logger
+    # is set right.
+    sherpalogger = logging.getLogger('sherpa')
```
DougBurke, line 31: This is a bit nicer with

```
    with caplog.at_level(logging.INFO, logger='sherpa'):
        logger = logging.getLogger('sherpa.some_module')
        logger.warning('1: should be seen')
        assert len(caplog.records) == 1

        ... rest of the test
```

DougBurke:
I can push a fix for the `test_smoke.py` case if wanted.

sherpa/utils/tests/test_logging.py
DougBurke: I can also push a change for this one too.

hamogu: Sure, I'm happy to take it!

hamogu:
Happy to take your take on `test_smoke`, too ;-)

DougBurke:
The test failure will be fixed by #1043 (I claim) and is unrelated to this PR.

DougBurke (approved):
LGTM. I am a bit biased here though :-)

hamogu:
Pretty sure that the RTD failure is not related to this PR, I suggest to merge.

hamogu:
Can we merge as is since we know that the unrelated RTD failure is fixed in master? I feel that rebasing this on master just to re-run all tests is a waste to GH action minutes.

DougBurke:
I don't see there being any request to rebase this. We just have a backlog of items and I have prioritized this as "lower priority" as it adds a new feature, rather than tweaks tests or the like, which I've prioritised as "higher priority". I expect it to get merged soon (assuming that AAS doesn't take up too much of @wmclaugh's time this week)

Issue #142, opened 2016-01-08
Explanations about TableModel


tibaldo:
I am working on a project for which I need to use arbitrary non-arithmetic spectral models. Therefore, I am planning to use TableModel. To make sure I understand how to use it properly, I did an exercise using a test dataset and comparing the predicted counts (from the plot_model method) between a sherpa powlaw1d model with gamma 2 and a TableModel that I constructed as follows. I took as x values for TableModel the energy boundaries of the arf file, and I calculated the y values as y=N*(x/E0)^-g (I took E0 the same as ref in sherpa powlaw1d). As you can see in the attached plots, the predicted counts look different if I use g=2 for the TableModel, but they are consistent for g=1 (I tweaked the normalization so that the number of counts in the peak is approximately the same for easier visual comparison). I verified that this does not depend on the x grid used for TableModel (for a 10 times finer grid the results are the same). Hence, I cannot figure out why the spectral indices are different between sherpa powlaw1d and the simple arithmetic formula. Probably I misunderstand something about TableModel or powlaw1d. Any explanations or suggestions are appreciated, thanks!
![model_powlaw1d-2](https://cloud.githubusercontent.com/assets/3046743/12194793/c3c4f16e-b5f2-11e5-9019-7511105b58f1.png)
![model_table-1](https://cloud.githubusercontent.com/assets/3046743/12194795/c61574e8-b5f2-11e5-9006-2d6c469263f2.png)
![model_table-2](https://cloud.githubusercontent.com/assets/3046743/12194798/c8277a7e-b5f2-11e5-824d-d6f66300830b.png)

tibaldo:
Sorry, file names disappeared: top is powlaw1d with gamma=2, mid panel is TableModel with g=1, and bottom TableModel with g=2.

DougBurke:
The table models do need better documentation! I think the issue here is that you need to calculate the integral of the model over the bin, so that the y values are in count/s rather than count/s/keV. The plots are automatically converted from count/s to count/s/keV. This is all assuming an integrated dataset (Data1DInt or DataPHA) rather than Data1D.

You should be able to overplot the models on the same graph - assuming you have them set up for datasets 1, 2, and 3 (with different combinations of model and g values):

    plot_model()
    plot_model(2, overplot=True)
    plot_model(3, overplot=True)

As an aside: the table model actually supports two forms of table model

* a simple grid that can be scaled by a normalization, as you are using here

[...]

PR #1039, opened 2020-12-29
Add docs for basic.TableModel

hamogu:
Summary
-------
Adding docs for TableModel

I had to find how this works from reading the code so I might as well
write it down for the next user to stumbple across this obscure model.

### DIFF
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1779,6 +1779,61 @@ class Polynom2D(RegriddableModel2D):
 
 
 class TableModel(ArithmeticModel):
+    """Tabulated values in this model are simply linearly scaled.
+
+    After initializing this model, the independent and dependent
+    arrays need to be loaded using the ``load`` method before the
+    model can be used (see examples below); interpolation is used if
+    the grids of the data and the table do not match.
+
+    This model can also be used as to fit indexed data where the
+    independent axis is not a continuous variable, e.g. the
+    independent variable may hold the index for a number of stars and
+    the dependent variable some measured property for each star. In
+    this case, the independent variable needs to be loaded as ``None``
+    and the length of the dependent array needs to match the length of
+    the data array exactly.
+
+
+    Attributes
+    ----------
+    ampl
+        The linear scaling factor for table values
+
+    See Also
+    --------
+    Const1D
+
+    Examples
+    --------
+    Below is an example for the "indexed" use:
+
+      >>> import numpy as np
+      >>> from sherpa.models.basic import TableModel
+      >>> from sherpa.data import Data1D
+      >>> from sherpa.stats import Chi2
+      >>> from sherpa.optmethods import NelderMead
+      >>> from sherpa.fit import Fit
+      >>> d = Data1D('data', [1,2,3,4,5], [1.2, .4, 2.2, .3, 1.],
+      ...             staterror=[.2, .2, .2, .2, .2])
+      >>> tm = TableModel('tabmodel')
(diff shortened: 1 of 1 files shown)


DougBurke:
I haven't checked - would this allow us to close #142 ?

hamogu:
I'll look at the linked issue and see if I can roll that in here, too.

sherpa/models/basic.py
```diff
@@ -1786,6 +1786,10 @@ class TableModel(ArithmeticModel):
     model can be used (see examples below); interpolation is used if
     the grids of the data and the table do not match.
 
+    When used with an integrated data set (for example,
```
DougBurke, line 1789: May be for a follow-up doc update, but generally with "integrated" or "gridded" or "whatever the term we want to use" the model gets sent both `xlo` and `xhi` - so something like one of

    def evaluator(pars, x, xhi=None):
    def evaluator(pars, x, *args):

[the form depends where in the evaluation chain this happens, I've arbitrarily picked a low-level routine I'd expect to be calling inside the model class]. Does the tablemodel support this or does it only get sent a single array? Or have you not looked into it (the trouble with working on a bit of Sherpa is you then get asked all the questions about it from now on).

For the basic models I tried to add a `Notes` section to explain the difference for point versus grid - e.g.

- https://sherpa.readthedocs.io/en/latest/model_classes/api/sherpa.models.basic.NormGauss1D.html
- https://sherpa.readthedocs.io/en/latest/model_classes/api/sherpa.models.basic.Poisson.html
- https://sherpa.readthedocs.io/en/latest/model_classes/api/sherpa.models.basic.Gauss2D.html

I don't know if we want to formalize this documentation practice (and improve it). I do know there are questions about labelling (eg #1042) which it would be nice to come to some decision on.

hamogu: The table model does not use the `x1` array even if it's given one, hence the "lower edge" description in the docstring. Personally, I could see a new class "IntregratedTableModel" (because of backwards compatibility, I would not want to change this one) that does better, but this PR is about documenting existing features not writing new ones.

DougBurke: Hmm. For better or worse we normally have a model support both point (x) and grid (xlo, xhi) options, but that;s for historical raisins. Something to throw on our to-do pile...

DougBurke: [going off into the bush here]

I see the evaluation function does accept a `xhi` (aka `x1`) argument

https://github.com/sherpa/sherpa/blob/2e8de205fa88323f6708eca6aecceae76c5953f8/sherpa/models/basic.py#L1827

but it ignores it, and more-importantly there's no way to set such an argument with the loaded data

https://github.com/sherpa/sherpa/blob/2e8de205fa88323f6708eca6aecceae76c5953f8/sherpa/models/basic.py#L1801

DougBurke (approved):

DougBurke:
The test failure was a "transient event" (fixed by #1043) unconnected with this PR (an issue with the read-the-docs build) and doesn't need to be re-run

sherpa/models/basic.py
hamogu: I think it needs to accept the `xhi` to work with integrated datasets, but then it does not use it - but at least it won't fail and if the bins are small enough and the table model is smooth enough it might just be OK -- or at least that's what I imaging the original intent to be, but that was before my time. Either way it's not easy to be better and at the same time be flux conserving (it currently almost certainly is not). I'll open a new issue about this so we don't forget, but it's certainly not a high-priority item.

DougBurke: Ta

PR #1043, opened 2021-01-04
RTD: install more packages via pip

DougBurke:
There's a tension between pip and conda, with conda often being
behind pip. At the moment this appears to cause a problem with
indirect requirements - namely parso and jedi - which causes the
RTD build to fail. Hopefully moving to pip for most of the
Python packages will help.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -7,9 +7,9 @@ channels:
 dependencies:
   - python=3.7
   - numpy
-  - sphinx-astropy
   - pandoc
-  - ipykernel
   - pip
   - pip:
+    - ipykernel
     - nbsphinx
+    - sphinx-astropy


hamogu:
Often, pip installs are faster today than conda, too so it may help to save some electrons.

DougBurke:
For reference, here is an example of a failed of the master branch which this is hopefully fixing https://readthedocs.org/projects/sherpa/builds/12687642/ (although it's not entirely clear we can see the failure here as a PR from earlier today passed even though the master branch was failing; I think there may be a subtle cache issue on the master branch version that you don't get on clean PRs)

DougBurke:
@Marie-Terrell - I've asked for your review on this more as an attempt to improve the knowledge of how the RTD docs are built than anything specific for this PR. The reason for this is that  the current docs status is failing for the master branch - aka https://readthedocs.org/projects/sherpa/badge/  which displays as doc build status - due to an issue with some python modules, so this fixes the problem for now (by moving more installation to pip from conda; unfortunately the RTD environment is restricted in what you can / can't do).

Let me know if this isn't a good time and I can get someone else to look at this.

Marie-Terrell (approved):
I have no problem with these changes. I do appreciate keeping me in the loop on this. 
As a follow-up, will this need to be merged and included in 4.13.0?

DougBurke:
Good question. It's reminded me to set up a build for the 4.13.0 branch - when it's finished we can see whether we need this or not.

DougBurke:
Looks like we will want this PR for 4.13.0 as the docs build for the 4.13.0 branch just failed.

PR #1057, opened 2021-01-13
Fix plot_source clearwindow setting

DougBurke:
# Summary

The `plot_source` function was ignoring the `clearwindow` parameter (always using `True`) in calls to `plot_source` for non-PHA data. This was only for `sherpa.astro.ui.plot_source` (so `sherpa.ui.plot_source` did not have this problen).

# Details

This was found during development of #1022 and has been moved into it's own PR. The bug was introduced by #906.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020
+#  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020, 2021
 #      Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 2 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -11515,7 +11515,7 @@ def plot_source(self, id=None, lo=None, hi=None, replot=False,
             return
 
         super().plot_source(id=id, replot=replot, overplot=overplot,
-                           clearwindow=True, **kwargs)
+                            clearwindow=clearwindow, **kwargs)
```
hamogu, line 11518: Is there a specific reason why "overplot" and "clearwindow" are treated separately and are not just part of `kwargs`?
Presumably,  the routines this get passed to also have their defaults for those parameters, so why do we have to set (or even mention) them here at all? They can still be listed in the docs if we want to.
(I have the grand idea of autogenerating some part of the docs that just pastes in the explanation for the parameters that every plotting function takes, but that's a a different matter.)

DougBurke: I can't speak to the original design. I do note the tension between "hey, if we just have a kwargs dictionary we can avoid some repetition" versus "Let's be explicit about the function arguments". I tend to the latter as I **really** dislike user-level functions which just have a `**kwargs` declaration and no easy way to find out what the valid inputs actually are (matplotlib has improved on this area but it requires discipline in the documentation writing).

wmclaugh (approved):

PR #1072, opened 2021-01-22
Update load_data to match load_pha for PHA2 data

DougBurke:
# Summary

Ensure that `load_data` behaves like `load_pha` when given a PHA2 dataset.

# Details

I hadn't realized when writing #856 that the `load_data` call could be used to load in PHA2 data. This rectifies this (and ensures the same logic is used by both functions).

# Example

With CIAO 4.13 we have the following, where the id argument to `load_data` is unused for PHA2 data.

```
sherpa In [1]: load_pha('x', '3c120_pha2')
WARNING: systematic errors were not found in file '3c120_pha2'
statistical errors were found in file '3c120_pha2'
read background_up into a dataset from file 3c120_pha2
read background_down into a dataset from file 3c120_pha2
Multiple data sets have been input: x1-x12

sherpa In [2]: load_data('x', '3c120_pha2')
WARNING: systematic errors were not found in file '3c120_pha2'
statistical errors were found in file '3c120_pha2'
read background_up into a dataset from file 3c120_pha2
read background_down into a dataset from file 3c120_pha2
Multiple data sets have been input: 1-12
```

With this PR this becomes (note that the id's are now x1 to x12 rather than 1 to 12):

```
In [3]: load_data('x', 'sherpa-test-data/sherpatest/3c120_pha2')
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c120_pha2'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c120_pha2'
read background_up into a dataset from file sherpa-test-data/sherpatest/3c120_pha2
read background_down into a dataset from file sherpa-test-data/sherpatest/3c120_pha2
Multiple data sets have been input: x1-x12
```

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1458,19 +1458,79 @@ class Session(sherpa.ui.utils.Session):
                 datas.elo, datas.ehi = calc_err(datas)
             datas.staterror = calc_staterror(datas)
 
-    # DOC-NOTE: also in sherpa.utils
+    def _load_data(self, id, datasets):
+        """Load one or more datasets.
+
+        Used by load_data and load_pha.
+
+        Parameters
+        ----------
+        id : int or str, optional
+           The identifier for the data set to use. For multi-dataset
+           files, currently only PHA2, the id value indicates the
+           first dataset: if it is an integer then the numbering
+           starts at id, and if a string then a suffix of 1 to n is
+           added.  If not given then the default identifier is used,
+           as returned by `get_default_id`.
+        datasets : Data instance or iterable of Data instances
+           The data to load, either as a single item or, for
+           multiple-dataset files, an iterable of them.
+
+        """
+
+        if not numpy.iterable(datasets):
+            self.set_data(id, datasets)
+            return
+
+        # One issue with the following is that if there's
+        # only one dataset in phasets and id is a string then the
+        # output will be "foo1" rather than "foo" (when
+        # id="foo").  DJB thinks we can live with this.
+        #
+        if id is None:
+            id = self.get_default_id()
+
+        num = len(datasets)
+        ids = []
+        for ctr, data in enumerate(datasets):
+            try:
+                idval = id + ctr
+            except TypeError:
+                # id is assumed to be a string
+                idval = id + str(ctr + 1)
+
+            self.set_data(idval, data)
+            ids.append(idval)
+
+        if num > 1:
+            info("Multiple data sets have been input: " +
+                 "{}-{}".format(ids[0], ids[-1]))
+        else:
+            info("One data set has been input: {}".format(ids[0]))
+
+    # DOC-NOTE: also in sherpa.utils without the support for
+    #           multiple datasets.
+    #
     def load_data(self, id, filename=None, *args, **kwargs):
         """Load a data set from a file.
 
         This loads a data set from the file, trying in order
         `load_pha`, `load_image`, `load_table`, then `load_ascii`.
 
+        .. versionchanged:: 4.13.1
+           The id argument is now used to define the first identifier
+           when loading in a PHA2 file to match `load_pha` (previously
+           the range always started at 1).
+
         Parameters
         ----------
         id : int or str, optional
-           The identifier for the data set to use. If not given then
-           the default identifier is used, as returned by
-           `get_default_id`.
+           The identifier for the data set to use. For multi-dataset
+           files, currently only PHA2, the id value indicates the
+           first dataset: if it is an integer then the numbering
+           starts at id, and if a string then a suffix of 1 to n is
+           added.  If not given then the default identifier is used,
+           as returned by `get_default_id`.
         filename
            A file name or a data structure representing the data to
            use, as used by the I/O backend in use by Sherpa: e.g.  a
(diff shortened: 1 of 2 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -1458,19 +1458,79 @@ def calc_staterror(data):
...
+
+        if not numpy.iterable(datasets):
+            self.set_data(id, datasets)
+            return
+
+        # One issue with the following is that if there's
+        # only one dataset in phasets and id is a string then the
+        # output will be "foo1" rather than "foo" (when
+        # id="foo").  DJB thinks we can live with this.
+        #
```
hamogu, line 1489: Will this happen even for PHA type I datasets? If so, I think we should fix it. People would be a little surprised if every dataset named with a string suddenly has a "1" at the end. If (as a I understand that) this happens only for PHAII datasets, then we should declare that a feature and move this comment into the docstring.

hamogu (approved):
Tiny comment on a comment. I'm approving, but please look at it. One suggestion to think about.
In practice, we often have the following situation in HETG or LEGT data: We run chandra_repro and get the datasets with default extraction, which gives me order -3 to +3 for both HEG and MEG, but we really only want to use the first order spectra, because all the others are too noisy to be useful. The correct solution might be to not use default arguments and only extract order -1 and 1, but that's how CIAO works for now. (XMM uses PHA type I files, so it's not as noticeable there). So, should Sherpa provide a switch to select which spectra to read in? I fear feature creep. However, it might be worth a few lines of and Example in the docstring, something like (with correct formatting):
```
Example
----------
Here, we load HETG spectra:
>>> ui.load_data(id='3c120_', '3c120_pha2.fits')

Now, we remove the spectra of higher orders of the HEG where the signal is low:
>>> ui.delete_data('3c120_1')
>>> ui.delete_data('3c120_6')
```

Of course, removing is not strictly necessary, I'm just thinking aobut how a typical user workflow might look like. What do you think @DougBurke?

sherpa/astro/ui/utils.py
DougBurke: This only happens when unpacking a dataset returns multilpe data objects, and this only happens for a PHA2 dataset. It would have been obvious in our tests, but I did add an explicit check for this which is already in the master branch: https://github.com/sherpa/sherpa/blob/40f27543a380aec34e2a5cab264c47a3bbd58a8e/sherpa/astro/ui/tests/test_astro_ui.py#L253

Unfortunately, for reasons which 'll mention on the main discussion, we don't have a test of the 'load a single file from a PHA2' case.

I had thought the documentation was already clear about the id naming scheme, but I am way too close to it!

DougBurke:
So, if you have the crates backend then you can say (well, you will with this PR, ni CIAO 4.13 you have to use `load_pha`):

```
>>> load_data('neg', 'pha2[tg_m=-1]')
WARNING: systematic errors were not found in file '3c120_pha2[tg_m=-1]'
statistical errors were found in file '3c120_pha2[tg_m=-1]'
read background_up into a dataset from file 3c120_pha2[tg_m=-1]
read background_down into a dataset from file 3c120_pha2[tg_m=-1]
Multiple data sets have been input: neg1-neg2
>>> load_data('part', 'pha2[tg_part=1]')  # I can never remember the naming
WARNING: systematic errors were not found in file '3c120_pha2[tg_part=1]'
statistical errors were found in file '3c120_pha2[tg_part=1]'
read background_up into a dataset from file 3c120_pha2[tg_part=1]
read background_down into a dataset from file 3c120_pha2[tg_part=1]
Multiple data sets have been input: part1-part6
>>> load_data('foo', 'pha2[to_m=-1,tg_part=1]')
WARNING: systematic errors were not found in file '3c120_pha2[tg_m=1,tg_part=1]'
statistical errors were found in file '3c120_pha2[tg_m=1,tg_part=1]'
read background_up into a dataset from file 3c120_pha2[tg_m=1,tg_part=1]
read background_down into a dataset from file 3c120_pha2[tg_m=1,tg_part=1]
>>> list_data_ids()
['neg1', 'neg2', 'part1', 'part2', 'part3', 'part4', 'part5', 'part6', 'x']
```

Aha. Note how the last one doesn't say "Multiple datasets". It's because the unpacking code (I guess, would need to check) has decided to not return a list of 1 data objects, just the data object itself, which then means the single-file case we talked about probably never actually happens (although this could be backend dependent; not sure).

sherpa/astro/ui/utils.py
DougBurke: Also note: this case may never happen (a single file being handled by the multi-file code).

DougBurke:
A note as to why we don't have tests of the "load a subset" of data, or the "select a single element" case: I don't believe it's possible with the AstroPy backend as there's no system like the virtual-file specification we have with the Crates/DM backend.

Some of our routines do allow "some" virtual-file-like specifications, but it is restricted to column selection, not filtering.

For the documentation we could have

 - for crates, you can say this as a short cut
 - otherwise you can loop over the files and remove them with something like (untested)

```
for id in list_data_ids():
    hdr = get_data(id).header
    if (hdr['TG_M'] not in [-1, 1]) and (hdr['TG_PART'] != 1]):
        delete_data(id)
```

sherpa/astro/ui/utils.py
hamogu: Then we won't worry about it.

hamogu:
OK, that fine. If it's that complicated (in the sense that it's backend dependent) then let's not worry about it here. I just wanted to bring it up, but your reply makes it obvious that this is not the right place for that. I already approved this PR because I kind of expected that this would be the answer, but I wanted to bring it up just in case.

So, ready to merge from my point of view!

Issue #1064, opened 2021-01-20
fake_pha with id=None can cause weirdness


DougBurke:
If you say `fake_pha(None, ....)` then you might expect it to work as if you'd used the default id value (e.g. `1` for most cases). Unfortunately this isn't the case, which can lead to subtle differences

 - if the default id had a background associated to it it would get forgotten with `id=None`
 - header values from the default id dataset would not be copied over
 - the error message for certain error conditions

I had noted this oddity in the tests I added as a part of #888 but didn't think to note it down as an issue.

I would expect the vast majority of users would not use an `id` value of `None` when calling `fake_pha` (the documentation does not suggest it is possible but the code does support it explicitly).

PR #1073, opened 2021-01-22
Allow fake_pha to be called with an identifier of None

DougBurke:
# Summary

The fake_pha command now treats id=None as the default id. This addresses #1064.

# Details

Shown in the examples is a minor improvement to the error message (reporting `1` rather than `None`). There are also other subtle issues which could arise: for example if dataset 1 has a background then calling `fake_pha(1,..)` would recognize the background whereas `fake_pha(None, ...)` would ignore the background.

I don't expect users to be calling `fake_pha` with `id=None`, but we may as well follow other routines. An alternative would be to require the `id` parameter to be specified, but that seems prescriptive.

# Example

An example of the error case: with CIAO 4.13 we get

```
sherpa In [4]: fake_pha(None, None, None, 100)
DataErr: An RMF has not been found or supplied for data set None
```

and with this PR (which has different logging hence sees more of the traceback):

```
In [4]: fake_pha(None, None, None, 100)
DataErr                                   Traceback (most recent call last)
<ipython-input-4-954b24257789> in <module>
----> 1 fake_pha(None, None, None, 100)

<string> in fake_pha(id, arf, rmf, exposure, backscal, areascal, grouping, grouped, quality, bkg)

~/sherpa/sherpa-play/sherpa/astro/ui/utils.py in fake_pha(self, id, arf, rmf, exposure, backscal, areascal, grouping, grouped, quality, bkg)
   8690
   8691         if rmf is None:
-> 8692             raise DataErr('normffake', id)
   8693
   8694         if type(rmf) in (str, numpy.string_):

DataErr: An RMF has not been found or supplied for data set 1
```

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
@@ -30,10 +30,6 @@ from sherpa.astro import ui
 from sherpa.utils.err import DataErr,IOErr
 
 
-def xfail(arg):
-    return pytest.param(arg, marks=pytest.mark.xfail)
-
-
 @pytest.mark.parametrize("id", [None, 1, "faked"])
 def test_fake_pha_no_rmf(id, clean_astro_ui):
     """Check we error out if RMF is None."""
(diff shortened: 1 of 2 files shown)


hamogu (approved):

DougBurke:
Ugh. For another issue, but `fake_pha` creates a `DataPHA` instance with both `channel` and `counts` being `None`. I don't see much point in being able to create such an "empty" object, and would prefer it if we require these two arguments to be set.

DougBurke:
Rebased to deal with merge conflict.

PR #1076, opened 2021-01-26
sample_flux now uses the id argument

### LINKED ISSUE #752: sample_flux does not work for non-default id
Seen when tracking down #751: `sample_flux` calls several routines - e.g. `calc_energy_flux` and `calc_stat()` - which will give the wrong result when using a non-default dataset:

```python
from sherpa.astro import ui
ui.load_pha(2, 'sherpa-test-data/sherpatest/3c273.pi')
ui.set_source(2, ui.powlaw1d.pl)
ui.fit(2)
ui.covar(2)
ui.sample_flux(id=2, lo=1, hi=7, num=10)
```

fails with this error

```
IdentifierErr: data set 1 has not been set
```

DougBurke:
# Summary

The sample_flux routine now uses the id argument rather than always using the data from the default dataset. Fixes #752

# Details

This has been pulled out of the long-stalled #754, so it includes a couple of documentation and code changes not directly related to #752. More changes will be pulled out of #754.


sherpa/astro/flux.py
```diff
@@ -563,23 +571,18 @@ def print_sample_result(title, arg):
...
-            logger.setLevel(logging.ERROR)
             session.set_source(id, orig_source)
-            logger.setLevel(orig_log_level)
             oflx[nn] = mysim[nn, 0]
             for ii in range(len(thawedpars)):
                 val = mysim[nn, ii + 1]
                 session.set_par(thawedpars[ii].fullname, val)
             session.set_source(id, modelcomponent)
-            iflx[nn] = session.calc_energy_flux(lo, hi)
+            iflx[nn] = session.calc_energy_flux(lo=lo, hi=hi, id=id)
```
hamogu, line 582: I know you are just fixing this routine, but I do note that it has no docstring. Since you just made fixes, you probably know better than anyone what it does and what the parameters are. Could you add a short doctring?
*I would have commented in line 531, but github does not let me add comments in lines that are not changed in this PR.*

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -1154,6 +1176,175 @@ def test_sample_foo_flux_scales_example(multi, make_data_path, clean_astro_ui,
...
+@requires_fits
+def test_bug_276(make_data_path):
+    ui.load_pha(make_data_path('3c273.pi'))
+    ui.set_model('polynom1d.p1')
+    ui.fit()
+    ui.covar()
+    scal = ui.get_covar_results().parmaxes
+    ui.sample_flux(ui.get_model_component('p1'), 0.5, 1, num=5,
+                   correlated=False, scales=scal)
+
```
hamogu, line 1189: Test any result here? The test ensure that it's not throwing an error, but that could fullfilled is `sample_flux` was a no-op. No need to go crazy, just to see if anything happens?

sherpa/astro/ui/utils.py
```diff
@@ -13248,6 +13248,10 @@ def sample_flux(self, modelcomponent=None, lo=None, hi=None, id=None,
         values used, together with the median, upper, and lower
         quantiles of the flux distribution.
 
+        .. versionchanged:: 4.13.1
```
hamogu, line 13251: I noticed that the `Returns` section in the docsstring has an unusual formatting. Should that be modified?

hamogu (commented):
I think this is done and almost good to go. Just a few short questions. The answer might well be "I thought about it, it's beyond the scope of this PR" and that's OK, too. I just wanted to ask.

sherpa/astro/flux.py
DougBurke: I have a whole lot of commits in #754 and I'm pretty sure I add documentation in them.

I just checked and it turns out I haven't added documentation. In part because the work isn't done yet to work out what I want this routine to do... Later PRs make more-significant changes to this routine so I think it makes sense to add doc changes then.

sherpa/astro/ui/utils.py
DougBurke: What is the problem with it? It at least renders "okay" at https://sherpa--1076.org.readthedocs.build/en/1076/ui/api/sherpa.astro.ui.sample_flux.html but I am likely missing something obvious

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
DougBurke: This is an old test, which I've just moved file in this PR, so I don't feel the need to work out what useful tests we can add to it.

There are explicit tests added in this PR which test the sample_flux output, so it's not like we don't have tests.

hamogu (approved):
As I said, I just wanted to to make sure it's all intentional. Fine by me!

PR #1078, opened 2021-01-27
Add error checking for exceptional cases for sample_flux

### LINKED ISSUE #286: Update sample_flux to use numpy.percentile
From a code review of PR #283 - https://github.com/sherpa/sherpa/pull/283#pullrequestreview-2708949 - it was noticed that the code in sample_flux should use numpy.percentile. This is a reminder to make that change and ensure the tests are adequate (since there are issues over how rounding is done to pick the percentile values).

DougBurke:
# Summary

Ensure that sample_flux errors out if the Xrays argument is False (as this code path is currently broken) or if the confidence argument is invalid.

Internal code clean up fixes #286

# Details

This is broken out of #754 to make the changes more manageable. 

Note that the switch to using numpy.percentile, for calculating the limits of a distribution, changes the results slightly (see the test changes in 9f13e86). I have not looked at the existing code to work out "which is better", as I tend to think that the library function will have less gotchas than a hand-written version, but we probably should think about this. 

Some of the changes - in particular parts of 4c458ef - will be reworked in a later PR once I get time to finish them.

### DIFF
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -578,19 +578,11 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
             session.set_full_model(id, orig_model)
             mystat.append(session.calc_stat(id))
             #####################################
-        oflxiflx = [oflx, iflx]
 
-        myconfidence = (1.0 - confidence / 100.0) / 2.0
+        hwidth = confidence / 2
         result = []
-
-        for x in oflxiflx:
-            sf = numpy.sort(x)
-            median = numpy.median(sf)
-            upconfidence_index = int((1.0 - myconfidence) * size - 1)
-            loconfidence_index = int(myconfidence * size - 1)
-            upconfidence = sf[upconfidence_index]
-            loconfidence = sf[loconfidence_index]
-            result.append(numpy.array([median, upconfidence, loconfidence]))
+        for x in [oflx, iflx]:
+            result.append(numpy.percentile(x, [50, 50 + hwidth, 50 - hwidth]))
 
         for lbl, arg in zip(['original model', 'model component'], result):
             med, usig, lsig = arg
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -530,19 +530,14 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
                      confidence):
 
     def simulated_pars_within_ranges(mysamples, mysoftmins, mysoftmaxs):
-        num = len(mysoftmins)
-        for ii in range(num):
-            ii1 = ii + 1
-            tmp = (mysamples[:, ii1] > mysoftmins[ii])
-            selpmasym = mysamples[tmp]
-            tmp = (selpmasym[:, ii1] < mysoftmaxs[ii])
-            mysamples = selpmasym[tmp]
+
+        for i, (pmin, pmax) in enumerate(zip(mysoftmins, mysoftmaxs), 1):
+            parvals = mysamples[:, i]
+            tmp = (parvals > pmin) & (parvals < pmax)
+            mysamples = mysamples[tmp]
+
         return mysamples
 
-    def print_sample_result(title, arg):
-
-        print('%s = %g, + %g, - %g' % (title, arg[0], arg[1] - arg[0],
-                                       arg[0] - arg[2]))
     #
     # For later restoration
     #
@@ -550,7 +545,6 @@ def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
     orig_model_vals = fit.model.thawedpars
 
     orig_source = session.get_source(id)
-    orig_source_vals = orig_source.thawedpars
 
     logger = logging.getLogger("sherpa")
     orig_log_level = logger.level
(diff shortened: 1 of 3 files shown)


hamogu:
@DougBurke I feel like I'm the bottleneck now and I know I'm not even the only one reviewing your PRs. How do you get them open so fast? You are certainly on a run here!

DougBurke:
I have a whole pipeline of them in various states of being, from the ephemerally frivolous to the should-have-been-doing-science type.

hamogu:
Let's try to clean out the pipeline1

DougBurke:
That's why I'm trying to break up some of my larger changes into more-manageable chunks (which makes sense in its own right)!

sherpa/astro/flux.py
```diff
@@ -584,22 +578,16 @@ def print_sample_result(title, arg):
...
-            upconfidence = sf[upconfidence_index]
-            loconfidence = sf[loconfidence_index]
-            result.append(numpy.array([median, upconfidence, loconfidence]))
-
-        print_sample_result('original model flux', result[0])
-        print_sample_result('model component flux', result[1])
+        for lbl, arg in zip(['original model', 'model component'], result):
+            med, usig, lsig = arg
+            msg = '{} flux = {:g}, + {:g}, - {:g}'.format(lbl, med, usig - med, med - lsig)
+            print(msg)
```
hamogu, line 590: This is tricky. I would really want this to be a `logger.INFO` message. I thought we want to move away form print statements in general, because users can't filter or switch them off if they run this in a script. On the other hand, this is inside a block that resets the log level so we want simply do that. 
The code above resets the base sherpa logger to ERROR, so even a new logging instance just for this function would not show INFO messages. Maybe we can make the line above that silences logging of other functions more specific and then create a new logger here that emits the output?

hamogu (commented):
I'm not concerned about minor changes with np.percentile. I'm pretty certain that it's just rounding differences (i.e. if you have 100 samples and ask for the 25.5 percentile, do you pick 25 or 26?). It does not matter here what's "more correct for our problem". If you worry about that and it really matters, then one should have more samples anyway. On the plus side, the cleaned up code is much easier to read and thus easier to maintain - a clear plus.

One inline questions about logging. I do understand that this just reworks existing code that used print before, but since that line is touched it would be good to fix. If @DougBurke declares it to be beyond the scope, I'll open an new issue about "replacing print statements with logging.info in general".

sherpa/astro/flux.py
DougBurke: The problem with breaking up #754 into manageable chunks is that some of the follow-on clean-up code context is missed. There are planned logging changes coming - e.g. 275ea0fd6aeb002fd3ed8178fa9dcf531b85495f and then d40677607ad7d95f7cd385f23d2061e7b67c6115 and then actually adding a test at db3243488e9a15105a3dbd6a58bff84793899b7b - but they are not easy to separate out from some of the other changes I made in that PR.

So I hope to be able to assuage your concerns with Yet Another PR soon ;-)

hamogu (approved):

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -2210,3 +2209,137 @@ def test_sample_flux_pha_component(idval, make_data_path, clean_astro_ui,
...
+    assert np.log10(bflux1) == pytest.approx([-13.15241991, -12.97986006, -13.31382087])
+
+    # Just check that the flux we've got is close to the one calculated by
+    # calc_energy_flux. This is just to test the values we have checked for
+    # the median value of flux1 and bflux1.
+    #
+    # Note that I am slightly concerned that the background flux has been
+    # calculated using the source response, not the background response.
+    # This is too hard to identify with this dataset. We need a test which
+    # has drastically different background response (nad maybe exposure
```
wmclaugh, line 2326: nad -> and

DougBurke: @wmclaugh I've commited a fix for this. There are fairly-easy-i'm-sure to make a suggestion that @hamogu has been making, which makes it easy for me to add them automatically, but I'm just not sure how.

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -2101,8 +2100,8 @@ def test_sample_flux_errors(make_data_path, clean_astro_ui,
     elo = fmed - flsig
     ehi = fusig - fmed
     assert fmed == pytest.approx(77.27966775437665)
-    assert elo == pytest.approx(9.86485842683382)
-    assert ehi == pytest.approx(11.40548578502279)
+    assert elo == pytest.approx(9.773389936230018)
```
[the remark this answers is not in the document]
DougBurke: All I did was grab the expected value and copy it over to the test. While I can understand an argument for saying "only bother with giving values we use in the comparison", I can also come up with 

a) I don't have time to sit down for each value and validate the precision that can be used
b) if we ever decide to switch to a different approximation routine (or pytest.approx changes) then it would be annoying to have to re-calculate all these values
c) having the full value let's us see if there are subtle differences when reviewing code changes (this is definitely clutching at straws)

I would argue it's primarily issue a.

wmclaugh (approved):

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
hamogu: You make a suggestion with the small icon on the top of the edit bad (next to the H B I for "header" "bold" "italics"). If @DougBurke wants to make use o these suggestions to have himself some typing, he needs to go the GH website and click "apply suggestion". That will create a new commit on the branch with the suggestion. (There is also a way to batch them so you get only one commit for all applied suggestions.). This is a true new commit on your fork and GH action will run on it and it can be merged into master etc. If you want to continue to work on that branch locally, then you have to pull form your branch in your fork first.
(Of course, you can also just look and type any changes we suggest as you've done so far.)

PR #1085, opened 2021-01-28
WIP how do we plot PHA data?

DougBurke:
Can we improve our plotting of PHA data, focussing on whether to normalize by the bin width or not?

This is an experiment in seenig if it makes sense to add an explicit flag (similar to the factor and type settings from `set_analysis`), and what that means for the UI.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1367,6 +1367,18 @@ class DataPHA(Data1D):
                         doc='Number of times to multiply the y-axis ' +
                         'quantity by x-axis bin size')
 
+    def _get_plot_norm(self):
+        return self._plot_norm
+
+    def _set_plot_norm(self, val):
+        val = bool(val)
+        self._plot_norm = val
+        for id in self.background_ids:
+            self.get_background(id).plot_norm = val
+
+    plot_norm = property(_get_plot_norm, _set_plot_norm,
+                         doc='Do we normalize y-axis by the bin width?')
+
     def _get_response_ids(self):
         return self._response_ids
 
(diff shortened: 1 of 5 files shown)


DougBurke:
```
>>> load_data('3c273.pi')
>>> load_data(2, '3c273.pi')
>>> set_source(xsphabs.gal * xspowerlaw.pl)
>>> set_source(2, get_source())
>>> gal.nh = 0.05; freeze(gal); fit(1)
>>> notice(0.5, 8)
>>> set_plot_opt(id=2, norm=False, type='counts')  # this is new
>>> plot('fit', 1, 'fit', 2, xlog=True)
```

produces

sherpa

With the existing code there's no way to get "counts" for the Y axis for a grouped dataset like this.

DougBurke:
I am now exploring making `norm` a string rather than a boolean (an enumerated type but using string values). This lets us normalize by "number of channels" even when the X axis is in keV.

```
>>> load_data('3c273.pi')
>>> load_data(2, '3c273.pi')
>>> load_data(3, '3c273.pi')
>>> set_source(xsphabs.gal * xspowerlaw.pl)
>>> set_source(2, get_source())
>>> set_source(3, get_source())
>>> notice(0.5, 6)
>>> gal.nh = 0.05; freeze(gal); fit(1)
>>> set_plot_opt(id=2, norm='channel', type='counts')
>>> set_plot_opt(id=3, norm='none', type='counts')
>>> plot('fit', 1, 'fit', 2, xlog=True)
```

sherpa

This does let you plot counts/s/Angstrom vs energy, which is unlikely to be interesting, but if we are going to go from a choice of 2 values to more, it's not obvious that there's a benefit to having a special "divide by number of channels" option rather than a "divide by <any valid analysis type>", but it is definitely something to think about.

hamogu:
Doug said I had thoughts... so here they are. For gratings we want to be able to make plots like this
image
or like this:
image
Both images are taken from TGCaT (i.e. ISIS) and they are absolutely typical of how grating spectra are displayed. You see the same type of scaling from e.g. SPEX users and also from Sherpa users. `set_analysis('energy'); plot_data()` get's close.

DougBurke:
So, having some way to say "divide by channel, energy, or Angstrom" is the desired approach. Ideally integrated into the data class (can ask to convert an axis to a certain scaling, perhaps), but that can be done later.

PR #1086, opened 2021-01-28
Use the logging infrastructure for sample_flux output

DougBurke:
# Summary

The screen output from sample_flux is now generated by the Sherpa logger and so can be hidden by the user if required.

# Details

This is pulled out of #754 and was coincidentally asked for by @hamogu in reviewing #1078

The sample_flux code still contains some other logging code related to hiding messages during the flux calculation. There are other changes from #754 that will improve that part of the code.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13331,7 +13331,7 @@ class Session(sherpa.ui.utils.Session):
         median and confidence ranges - is controlled by the standard
         Sherpa logging instance, and can be hidden by changing the
         logging to a level greater than "INFO" (e.g. with
-        sherpa.utils.logging.SherpaVerbosity).
+        `sherpa.utils.logging.SherpaVerbosity`).
 
         Examples
         --------
@@ -13368,7 +13368,7 @@ class Session(sherpa.ui.utils.Session):
 
         Run sample_flux after changing the logging level, so that the
         screen output from sample_flux is not displayed. We use the
-        SherpaVerbosity function from sherpa.utils.logging to
+        SherpaVerbosity function from `sherpa.utils.logging` to
         only change the logging level while runnng sample_flux:
 
         >>> from sherpa.utils.logging import SherpaVerbosity
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13250,7 +13250,8 @@ class Session(sherpa.ui.utils.Session):
 
         .. versionchanged:: 4.13.1
            The `id` parameter is now used if set (previously the
-           default dataset was always used).
+           default dataset was always used). The screen output is now
+           controlled by the Sherpa logging setup.
 
         Parameters
         ----------
@@ -13329,7 +13330,8 @@ class Session(sherpa.ui.utils.Session):
         The summary output displayed by this routine - giving the
         median and confidence ranges - is controlled by the standard
         Sherpa logging instance, and can be hidden by changing the
-        logging to a level greater than "INFO".
+        logging to a level greater than "INFO" (e.g. with
+        sherpa.utils.logging.SherpaVerbosity).
 
         Examples
         --------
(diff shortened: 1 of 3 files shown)


DougBurke:
Hmm, I should bump the example to use SherpaVerbosity, which wasn't a thing when I first wrote this code.

hamogu:
> Hmm, I should bump the example to use SherpaVerbosity, which wasn't a thing when I first wrote this code.

I'm sorry for messing up your old commits in that way ;-)

DougBurke:
Well, I'm getting my revenge by assigning this to you to review...

sherpa/astro/ui/utils.py
```diff
@@ -13326,6 +13327,12 @@ def sample_flux(self, modelcomponent=None, lo=None, hi=None, id=None,
         -----
         Setting the Xrays parameter to False is currently unsupported.
 
+        The summary output displayed by this routine - giving the
+        median and confidence ranges - is controlled by the standard
+        Sherpa logging instance, and can be hidden by changing the
+        logging to a level greater than "INFO" (e.g. with
+        sherpa.utils.logging.SherpaVerbosity).
```
hamogu, line 13334:
```suggestion
        `sherpa.utils.logging.SherpaVerbosity`).
```

sherpa/astro/ui/utils.py
```diff
@@ -13359,6 +13366,15 @@ def sample_flux(self, modelcomponent=None, lo=None, hi=None, id=None,
         >>> cmatrix = get_covar_results().extra_output
         >>> ans = sample_flux(correlated=True, scales=cmatrix, num=500)
 
+        Run sample_flux after changing the logging level, so that the
+        screen output from sample_flux is not displayed. We use the
+        SherpaVerbosity function from sherpa.utils.logging to
```
hamogu, line 13371:
```suggestion
        SherpaVerbosity function from `sherpa.utils.logging` to
```

hamogu (approved):
Backticks  in the docstring should make sphinx generate links to the function/module. Not strictly required, but helpful.

Since I know that @DougBurke is online and working right now, I approve anyway because I know that he can do that changes immediately. If not, that's OK, too.

sherpa/astro/ui/utils.py
DougBurke: I have found the backtick-links to be less than successful in my documentation so far, but I'l try again (maybe the fact it's fully qualified name will help)

hamogu: Yes, it needs to be the full name for sphinx to find it. If you don't want the full name to appear in the docs, you write
`~sherpa.utils.logging.SherpaVerbosity`, which will show up in the html as "SherpaVerbosity" with a link.

DougBurke: I use that in the docs/ setup, but I don't really like the extra noise in the docstrings of functions with the sphinx "extras".

DougBurke:
@hamogu - FYI it doesn't look like those backticls worked (this is what I've seen before, and I don't understand it): https://sherpa--1086.org.readthedocs.build/en/1086/ui/api/sherpa.astro.ui.sample_flux.html

hamogu:
They should work (see reference to CATGrating in  https://marxs.readthedocs.io/en/latest/api/marxs.optics.FlatGrating.html#marxs.optics.FlatGrating.order_sign_convention  - click on "source" to see the docstring), even across packages if intersphinx is set up (see e.g. https://marxs.readthedocs.io/en/latest/api/marxs.missions.mitsnl.catgrating.InterpolateEfficiencyTable.html#marxs.missions.mitsnl.catgrating.InterpolateEfficiencyTable - see the references to scipy - click on "source" to see the docstring)

It's something in the sherpa sphinx setup. I think it's VERY useful for the rendered html docs, so useful in fact that I usually read the html docs to have those links instead of reading the docstring in the code even for packages that I write myself (like the one the examples above are taken from).

Changing the sphinx conf.py is beyond the scope of this PR though. However, can you leave the backticks in? When we get that working for sphinx, we may not want to redo every docstring in Sherpa at once, but just upgrade them as we go.

DougBurke:
I wasn't going to waste more CI time removing those backticks!

PR #1089, opened 2021-01-31
WIP move to a more-extensible system for plot types

DougBurke:
This has been replaced by #1516 but I'll leave this around as the approach is slightly different, in case we want to discuss the differences.

# Summary

Internal change to plotting in sherpa.ui and sherpa.astro.ui. As a result there are small changes in the allowed identifiers for datasets. TBD should be able to list the changes

# Details

I've pulled out one change to #1175 but left in this PR (will probably remove at a later time).

Rather than having the per-plot objects at the top level of the session class, which makes it hard to identify what plots are
used in what places, and hard to write somewhat-generic code, place the plots into a dictionary. There was a partial attempt at improving this in #906 and #931 with the _plot_types dictionary, and this attempts to take that idea and apply it fully:

 - the _plot_store dictionary is used to define the "names" of the plots (e.g. "plot", "model", ...) but it is not yet clear
   about composite plots like "fit". The same change is made for _contour_store.
 - each entry is a `sherpa.utils.ObjectStore` item which is used to select the most appropriate object (in this case the plot class)
   given the data object (e.g. `Data1DInt` or `DataPHA`). There are many cases where this is excessive - as we only have
   a single plot type - but there's no attempt to optimize (either run time or for the person reading the code) this use case

Identifying the "most appropriate" type relies on having a common base case for all the data objects, so we can use the MRO depth for the class to select the largest depth that does not exceed the depth of the data object being queried.

The most user-facing change is in the selection of valid identifiers for a data set (and is actually already merged as part of #1105; I need to check if there's any remaining changes in this PR). Since we want to be able to say

    plot('data', identifier)

then we currently restrict identifier so that it does not match the valid set of plot types that can be sent to `plot` (and ditto for `contour`, but it turns out that contour doesn't add new plot types so we can just think about plots). However, this check isn't direct tied to the plot code, and so it has possibly drifted apart (as plot types were changed or added, I guess). Since we know have a "single" source that defines the plot types (the `_plot_store` dictionary) we can use that in the "is a valid identifier". That is a simplification, as - for reasons I'm not sure about - we actually use a different dictionary (which is related to _plot_store), but the idea is the same (and it may be something we can simplify here too).

---

## Examples

Here are examples where the code has changed::

### Example

`get_data_plot` from `sherp/ui/utils.py` has changed from:

```
try:
    is_int = isinstance(self.get_data(id), sherpa.data.Data1DInt)
except IdentifierErr as ie:
    if recalc:
        raise ie

    is_int = False

if is_int:
    plotobj = self._datahistplot
else:
    plotobj = self._dataplot

if recalc:
    plotobj.prepare(self.get_data(id), self.get_stat())
return plotobj
```

to

```
plotobj, data = self._get_plotobj('data', id, recalc=recalc)
if recalc:
    plotobj.prepare(data, self.get_stat())
return plotobj
```

and logic has been moved to the `clean` which defines

```
        self._plot_store = {}
        self._plot_store['data'] = ObjectStore(sherpa.plot.DataPlot())
        self._plot_store['data'].add(sherpa.data.Data1DInt, sherpa.plot.DataHistogramPlot())
```

### Example

The following routine (`get_data_plot`) from sherpa/astro/ui/utils.py has been removed

```
    def get_data_plot(self, id=None, recalc=True):
        try:
            d = self.get_data(id)
        except IdentifierErr:
            return super().get_data_plot(id, recalc=recalc)

        if isinstance(d, sherpa.astro.data.DataPHA):
            plotobj = self._dataphaplot
            if recalc:
                plotobj.prepare(d, self.get_stat())
            return plotobj

        return super().get_data_plot(id, recalc=recalc)
```

This is because we've added the following logic to `clean`, which means we can use the `get_data_plot` routine from sherpa/ui/utils.py (which was shown in the previous example):

```
        # Add PHA plot types. This is ugly.
        #        
       self._plot_store['data'].add(sherpa.astro.data.DataPHA, sherpa.astro.plot.DataPHAPlot())
```

### Example

An example where it isn't as obvious an improvement: `get_fit_plot` from `sherpa/astro/ui/utils.py` which has changed from

```
    def get_fit_plot(self, id=None, recalc=True):

        plotobj = self._fitplot
        if not recalc:
            return plotobj

        d = self.get_data(id)
        if isinstance(d, sherpa.astro.data.DataPHA):

            dataobj = self.get_data_plot(id, recalc=recalc)

            # We don't use get_model_plot as that uses the ungrouped data
            #    modelobj = self.get_model_plot(id)
            # but we do want to use a histogram plot, not _modelplot.
            # modelobj = self._modelplot

            # Should this object be stored in self? There's
            # no way to get it by API (apart from get_fit_plot).
            #
            modelobj = sherpa.astro.plot.ModelPHAHistogram()
            modelobj.prepare(d, self.get_model(id),
                             self.get_stat())

            plotobj.prepare(dataobj, modelobj)
            return plotobj

        return super().get_fit_plot(id, recalc=recalc)
```

to

```
    def get_fit_plot(self, id=None, recalc=True):

        plotobj, data = self._get_plotobj('fit', id, recalc=recalc)
        if not recalc:
            return plotobj

        if not isinstance(data, sherpa.astro.data.DataPHA):
            # This repeats some of the calls we have already made
            return super().get_fit_plot(id, recalc=recalc)

        dataobj = self.get_data_plot(id, recalc=recalc)

        # Unlike the normal fit plot we use a specialised model class.
        #
        # modelobj = self.get_model_plot(id, recalc=recalc)
        modelobj = sherpa.astro.plot.ModelPHAHistogram()
        modelobj.prepare(data, self.get_model(id), stat=self.get_stat())

        plotobj.prepare(dataobj, modelobj)
        return plotobj
```

---

## Test changes

One of the largest changes is in the tests, as some were directly accessing the fields we have now moved into the `_plot_store` object. The intent of these tests is to check the matplotlib backend is "doing the right thing" without actually having to check the hardcopy output (e.g. PNG or SVG output), so instead want to be able to check the plot objects are filled with the correct data (a common check is to make sure that the `xlog` or `ylog` settings are as expected). It is not always obvious that these checks are meaningful after the changes (in particular the model part of a "fit" plot with PHA data, which is handled specially and doesn't really fit into the plot system).

---

# Notes

There has been a change to template models, but it is not-at-all obvious to me that this  has made things "worse" as there appears to be a  difference between how sherpa.ui and sherpa.astro.ui were handling things:

## Template models

The old `sherpa.ui.utils.get_source_component_plot` included the logic

```
        if isinstance(model, sherpa.models.TemplateModel):
            plotobj = self._comptmplsrcplot
        elif isinstance(d, sherpa.data.Data1DInt):
            plotobj = self._compsrchistplot
        else:
            plotobj = self._compsrcplot
```

where `_comptmplsrcplot` was an instance of `sherpa.plot.ComponentTemplateSourcePlot` (note that there was commented-out code for the "model" version using `ComponentTemplateModelPlot`). Note that this logic is ignored for PHA data because we have for `sherpa.astro.ui.utils.get_source_component_plot` the following:

```
    def get_source_component_plot(self, id, model=None, recalc=True):
        if model is None:
            id, model = model, id
        self._check_model(model)
        if isinstance(model, string_types):
            model = self._eval_model_expression(model)

        try:
            d = self.get_data(id)
        except IdentifierErr as ie:
            if recalc:
                raise ie
            d = None

        if isinstance(d, sherpa.astro.data.DataPHA):
            plotobj = self._astrocompsrcplot
            if recalc:
                plotobj.prepare(d, model, self.get_stat())
            return plotobj

        return super().get_source_component_plot(id, model=model, recalc=recalc)
```

That is, if it's a `DataPHA` object then we skip the template-model check. So I argue that we need to review what the template model plot classes are meant to be for and when they should be applied (e.g. what happens if you have a composite model?). If we do need to worry about handling template-models specially then the approach used in this PR probably goes out of the window (or needs major work).

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -89,7 +89,7 @@ del _ncpu_val, config, get_config, ConfigParser, NoSectionError
 
 
 __all__ = ('NoNewAttributesAfterInit', 'SherpaFloat',
-           'OrderedByMRO',
+           'ObjectStore', 'OrderedByMRO',
            '_guess_ampl_scale', 'apache_muller', 'bisection', 'bool_cast',
            'calc_ftest', 'calc_mlr', 'calc_total_error', 'create_expr',
            'create_expr_integrated',
@@ -4388,6 +4388,59 @@ def send_to_pager(txt, filename=None, clobber=False):
 
     with open(filename, 'w') as fh:
         print(txt, file=fh)
+
+
+class ObjectStore:
+    """Allow object hierarchies to map to other objects.
+
+    Given a list of stored (class, value) pairs, allow the
+    most-appropriate value to be returned when given an instance. The
+    intention is that this is used to select from a tree (that is all
+    classes derive from some common ancestor), and outside this use
+    case the store is not guaranteed to work sensibly.
+
+    Parameters
+    ----------
+    default : object
+        The default object to return
+
+    Notes
+    -----
+    Once created and classes added via the `add` method, the store can
+    be queried by calling it, which will return the appropriate value
+    given the called value (or the default if no argument is given).
+
+    Examples
+    --------
+
+    >>> from sherpa import data
+    >>> s = ObjectStore(1)
+    >>> s.add(data.Data1D, 2)
+    >>> s.add(data.Data1DInt, 3)
+    >>> s()
+    1
+    >>> obj = data.Data1DInt('x', [1, 2], [2, 3], [5, 4])
+    >>> s(obj)
+    3
+
+    """
+
+    def __init__(self, default):
+        self.default = default
+        self.store = OrderedByMRO()
+
+    def add(self, cls, value):
+        self.store[cls] = value
+
+    def __call__(self, key=None):
+        if key is None:
+            return self.default
+
+        for k, v in self.store.items():
+            if isinstance(key, k):
+                return v
+
+        return self.default
 
 
 # The Python documentation isn't clear on UserDict vs dict usage but I've
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -32,6 +32,7 @@ import string
 import sys
 from configparser import ConfigParser, NoSectionError
 import pydoc
+from collections import UserDict
 
 import numpy
 import numpy.random
@@ -88,6 +89,7 @@ del _ncpu_val, config, get_config, ConfigParser, NoSectionError
 
 
 __all__ = ('NoNewAttributesAfterInit', 'SherpaFloat',
+           'OrderedByMRO',
            '_guess_ampl_scale', 'apache_muller', 'bisection', 'bool_cast',
            'calc_ftest', 'calc_mlr', 'calc_total_error', 'create_expr',
            'create_expr_integrated',
@@ -4386,3 +4388,63 @@ def send_to_pager(txt, filename=None, clobber=False):
 
     with open(filename, 'w') as fh:
         print(txt, file=fh)
+
+
+# The Python documentation isn't clear on UserDict vs dict usage but I've
+# read stories about the optimisations dict makes which could complicate
+# the code for this class, so stick with UserDict.
+#
+# It is expected that this class is only used with a small number of
+# items, and not used in time-critical places, so it is not optimised
+# for time but readability.
+#
+class OrderedByMRO(UserDict):
+    """Iterating over this dictionary goes from highest to lowest MRO depth.
+
+    It is intended to only store classes and is intended for returnig
+    information specific from the "most specific" class in the store.
+    This replicates the class hierarchy so indicates that something
+    "funky" is going on and maybe you should approach the problem from
+    a different angle!
+
+    Example
+    -------
+
+    With the following dictionary, a user can request the "most
+    relevant" information for an instance, in this case x (assuming
+    it's from the sherpa.data.Data1D hierarchy, although there is no
+    requirement that all objects have a common base class):
+
+    >>> d = OrderedByMRO()
+    >>> d[sherpa.data.Data1D] = "basic info"
+    >>> d[sherpa.data.Data1DInt] = "a bit more specialised"
+    >>> d[sherpa.astro.data.DataPHA] = "even more specialised"
+    >>> for k, v in d.items():
+    ...     if isinstance(x, k):
+    ...         print("found a match")
+    ...         break
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.store = {}
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, key, value):
+        try:
+            m = len(key.__mro__)
+        except AttributeError:
+            raise TypeError("Can only store classes")
+
+        super().__setitem__(key, value)
+        self.store[key] = m
+
+    def __delitem__(self, key):
+        super().__delitem__(key)
+        del self.store[key]
+
+    def __iter__(self):
+        store = list(self.store.items())
+        store = sorted(store, key=lambda x: x[1], reverse=True)
+        keys = [x[0] for x in store]
+        return iter(keys)
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -821,7 +821,7 @@ def check_bkg_fit(plotfunc, isfit=True):
     assert isinstance(mplot, BkgModelPHAHistogram)
 
     # check the "source" plots are not set
-    for plot in [ui._session._dataplot, ui._session._modelplot]:
+    for plot in [ui._session._modelplot]:
         assert plot.x is None
         assert plot.y is None
 
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -820,11 +820,6 @@ def check_bkg_fit(plotfunc, isfit=True):
     assert isinstance(dplot, BkgDataPlot)
     assert isinstance(mplot, BkgModelPHAHistogram)
 
-    # check the "source" plots are not set
-    for plot in [ui._session._modelplot]:
-        assert plot.x is None
-        assert plot.y is None
-
     xlabel = 'Channel' if plotfunc == ui.plot_bkg_fit else ''
 
     # check plot basics
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -822,7 +822,7 @@ def check_bkg_fit(plotfunc, isfit=True):
     fplot = ui.get_bkg_fit_plot(recalc=False)
     mplot = fplot.modelplot
 
-    dplot = ui._session._plot_store['bkg'][0]
+    dplot = ui._session._plot_store['bkg']()
 
     # Since we are checking the store object, we can just check
     # that the data object matches that stored in the fit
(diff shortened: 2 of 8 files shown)


hamogu:
I have to admit I don't understand exactly how this is going to work. Maybe I'm just looking too early and should wait a little longer until more plots and types are added to the dictionary so that it's easier to see. On the other hand I want to try useful feedback before it's all done. 

How about plots like `plot_bkg_fit_resid`? 

Conceptually, I wonder if this is a little similar to [holoviews](https://holoviews.org/getting_started/Introduction.html). They have different plot types that can be combined using a "formula-like" syntax on a data . Holoview overloads operators like + and *, so the above plot would be become something like "Hist(bkg) * Curve(bkg_fit) + Hist(bkg_resid)" or, if one wanted the model also in the look of a histogram as apposed to a smoother curve ""Hist(bkg) * Hist(bkg_fit) + Hist(bkg_resid)".
(An error is raised, if the number of dimensions does not match between data and plottype.) So, the basic principle is the decouple data from representation, similar to what Vega / Altair are also doing. 

If I understand correctly, though, this not not what is happening here, right?
Your dictionary here is more like this:
```
       self._plot_store = {}
        self._plot_store['data'] = (sherpa.plot.DataPlot(),
                                    {sherpa.data.Data1DInt: sherpa.plot.DataHistogramPlot(),
Data2D: CountourPlot,
Data2DInt: ImagePlot})
```
If that's the case, then I don't understand the comment about the MRO. Are you saying that you are looking for the first match in the mro, i.e. if I have
```
class MySuperComplexDataType(sherpa.data.Data1DInt):
  pass
```
then you want to use mro to ensure that that is matched to the entry for `sherpa.data.Data1DInt` because `MySuperComplexDataType` does not have an entry on it's own?

On the other hand, it's going to be more difficult to "force" a less-specialized plot or change the type, e.g. images I want PHA data with a line like in `plt.plot(x,y)` instead of the histogram-like plot that's the default. Can I do that? Do we even want that? Or is that a case of "go back to matplotlib and do it yourself"?

I'm sure this will be better than what we have right now. I just wonder if a little more thinking about the structure could make this even more useful. 

For backwards compatibility, we would still need to have the translations that you enter right now:
```
plot_stor['data'] = {'Data1D': DataPlot, 'Data1DInt': HistogramPlot, ...}

def ui.plot_fit(id):
        dat = get_data(id)
        return plot_store[type(dat)](dat)
```

Enough rambling. Don't want to derail that work...

DougBurke:
The current code is not easily extensible. I think that this should PR should help us identify what can be improved and hint at ways to go about this (whether it's my approach or a complex refit)

DougBurke:
@hamogu - if you're interested abut the MRO stuff...

The existing code has stuff like (using `isa` as shortform for 'is an instance of`)

 - call get_data
 - in sherpa.astro.ui.utils check if data isa DataPHA; if so go boom
 - otherwise fall through to sherpa.ui.utils and check if isa Data1DInt and can go boom
 - otherwise fall through to the base case and go boom

where "boom" means we have identified which plot object to use for whatever we want.

This is easier to reason about than the old version, but involves a lot of repeated code and knowledge that is hard coded (i.e. the mapping from data class to the particular object to make things go boom). What I am playing with is a structure that contains

    (base-object, dict of keyword=data class, value=subclass-of-base-object)

The idea being that given a data object, I can hunt through the dictionary to find the most appropriate match and return the value (falling through to the base object if not found). With the assumption that we have a common base class for all the Data objects (that is, we aren't supporting python-like you-have-the-methods-I-want-so-I-don't-care-if-you're-not-the-"correct"-object) then this search can be done using the mro depth (we want the element from the dictionary with the largest depth). The current code does this brute-force but it could be better.

What I really want for this is to be able to act the data object what plot class to use, but there are reasons why I do not want to just add this knowledge to the data class, such as it doesn't separate logic cleanly and it's actually a bit-more complicated than just returning the appropriate class.

I am sure there are better ways to do this. However, I think that those are "implementation" details and I want to check that the consequences of the change are okay, as I suspect that things like plot_fit are "interesting".

This is all within the context of having a mapping from a "plot type" (e.g. "resid", "fit", ...) to an actual plot and where we (aka Sherpa) control what that mapping means for each dataset. It's an open question whether more user control can be done in this formalism versus adding a separate mechanism.

hamogu:
I am interested, because I have the "implement bokeh/holoviews/plotly/..." backend in the back of my mind. I won't get to that before end of Feb at the earlier thought and I don't have any other useful comment at this point. But I'll keep thinking.

DougBurke:
Rebased so we can run the tests with modern systems  (matplotlib 3.4).

sherpa/utils/__init__.py
```diff
@@ -4188,6 +4188,59 @@ def send_to_pager(txt, filename=None, clobber=False):
         print(txt, file=fh)
 
 
+class ObjectStore:
```
DougBurke, line 4191: So, I have a tree of data classes - e.g. Data1D, Data1DInt, DataPHA - and potentially different values to store for each class. I then come along and say given an instance of one of these classes, what is the appropriate value from the store? The Data1DInt and DataPHA classes are derived from Data1D but they are not derived from each other, and it may be that DataPHA stores a value but Data1DInt doesn't.

We use the MRO depth to determine what value is a good match. This could well be over-engineering...

hamogu: I had to make a similar design decision for a different package a while ago (https://github.com/Chandra-MARX/marxs/blob/main/marxs/visualization/utils.py#L27). There is a similar hierarchy of classes and also the need to set per-object preferences. So, what I settled on there was to has a dictionary for each object and a special look-up so that any key that is not set in that object is looked up in the class dictionary, and if it's not found there, it's checked in the base class etc. That's similar to what you propose here and I found it to work. That way, one can, e.g. change the line width for just one histogram plot, without changing it for all current and future plots. But one can also (by modifying the class dictionary) change all histogram plots (except for those that override the setting in their own instance dictionary. How to make this transparent on the UI layer is a different question...

hamogu:
Now that I'm working on bokeh and know the plotting system better, a few more ideas (not necessarily for this PR): 

- **How is that dictionary / object store populated?** I think it would make sense to add a class (I'm a little fuzzy if we should store classes or instances) at the time the class is imported. That way, it becomes an extensible system in the sense that any plot object class that derives from our base class will automatically be registered (a simpler version of this: https://github.com/astropy/astropy/blob/main/astropy/io/ascii/connect.py and https://github.com/astropy/astropy/blob/main/astropy/io/ascii/core.py#L1134) 
- **Is a dictionary the right format?** Dicts are easy, but they are not good at storing many-to-many relations(dataclass and sub classes. <-> plottype). sqllite3 is probably overkill... (also, it is possible to make a Python installation without sqllite3, even though it's in the standard library. Not a common case we need to worry about thought.)

DougBurke:
[in part to remind myself]

So, before this PR the plot classes are registered on a one-to-one basis - e.g. here's the current `sherpa.ui.utils` and `sherpa.astro.ui.utils` versions

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/ui/utils.py#L290-L365

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/astro/ui/utils.py#L153-L228

The problem with this approach is that it makes it hard to extend (e.g. add a specific new plot type for your new DataFoo class, which I admit is not a high priority for us) or to work out the mapping between "plot types" - that is commands like `plot_data` and `get_resid_plot` - and the classes that are used.

What I'd like is something like the pseudocode of `plot_data` (or, actually, `get_data_plot` since `plot_data` should just call that routine and I forget whether that clean up has already made it  in or not as it's been too-long since I last looked at this):

```
    data = self.get_data(id)
    plotobj = self._get_plot_object("data", data)
```

That is, make the logic for mapping between the data class and the plot "label" automated and extensible. The problem for my implementation is that I want the equivalent of `_get_plot_object` to be able to automate the selection of the most-relevant class. So, manually I could write the following, but I want the logic to be automated.

````
    if isinstance(data, DataFoo):
        return something
    if isinstance(data, DataBar):
        return a-different-thing
    ...
    return the-base-thing
```

The approach I came up with is rather ugly. For instance, for the "data" plots, we have to worry about DataPHA, Data1DInt, Data1D and note that DataPHA extends the Data1D class [*] but plot-wise we extend from the "histogram" style plots (which is what Data1DInt uses). I am not sure that the hierarchy (who extends whom) really matters here, more the "pick the most-relevant class" work.

The "data" case is a good one to pick because we do add DataPHA support in sherpa.astro.ui.utils compared to sherpa.ui.utils.

[*] it's arguable about whether it should extend Data1D or Data1DInt but historically it's extended Data1D and I think we concluded in #614 that it wasn't worth changing

[**] not mentioned here, but the plot-object class names could do with a pass through as it's hard to work out what's used and related (and we also have a bunch of classes related to background plots which are not-very-different to the "source" version). But this is definitely a different issue.

Issue #1070, opened 2021-01-22
Write files to StringIO or using pathlib


hamogu:
(Some parts of) the code require any filename for reading/writing to be a string. It would be nice to accept a later range of objects here (a stream, a URL, a pathlib object). See
https://github.com/sherpa/sherpa/pull/1067#discussion_r562664154

In some cases that might already work (e.g. the astropy backend should deal with at least some of that) but it's certainly not part of the official requirements, it's not tested, and it's not consistent at this point.

DougBurke:
So some parts I've updated to allow a StringIO object, but it's been done piece meal and a system-level view might be nice. One issue we have is that we do a lot of not-particularly-pythonic type checks throughout our code and it's a bit of a rats nest to clean up (which is one of the reasons I've been going after test cases!)

DougBurke:
https://github.com/sherpa/sherpa/blob/307ddbbd4b7ff98e3c9720915f05aa530f4a1262/sherpa/fit.py#L570-L579

is something to remember, from 

- #2054

Issue #878, opened 2020-08-07
Should datasets have a `PrimaryHeader`?


hamogu:
When loading a PHA dataset the header information from the extension that holds the PHA (i.e. the `"SPECTRUM"` extension) is stored in a `dataset.header` attribute. That's used e.g. in the datastack to print out some diagnostic information about each dataset in the stack. However, it's common for telescopes to store only rudimentary information in the headers of the extension, e.g. for XMM, each extension will list the keywords `TELESCOP` and `INSTRUME`, but other vital information such as `OBS-ID` or the target (`OBJECT`) are only stored in the header of the primary hdu. I suggest to add an attribute `.primaryheader` that stores the primary header in the same format that `.header` stores the header of the specific extension. 

This is relatively easy to implement (probably < 5 lines of code per backend) and does not hurt in any way.

DougBurke:
We already have issues with writing out PHA files (I believe) because we don't treat the header properly. That's not to say this idea isn't warranted, but I imagine we would we want the primary block set up correctly (so down-stream consumers can make use of it), so it would be nice to fix this.

There's #488, but that's not what I was thinking about. Aha: it's #203

DougBurke:
Random comment (I'm not really thinking about this issue): would the user be expected to manually check through `header` and then `primaryheader` or would the interface "search both magically". I would strongly suggest the former - make the user be explicit, since I think the PHA case you'd expect to have the keywords in the SPECTRUM block (at least from my reading of the OGIP "standard") and for Chandra cases if you have a WMAP block as the primary header you don't want to get confused with the image settings there.

Issue #1090, opened 2021-02-01
numpy 1.20 causes test warnings and failures


DougBurke:
We're getting this warning from some tests now

```
sherpa/utils/tests/test_utils.py:584
  /home/runner/work/sherpa/sherpa/sherpa/utils/tests/test_utils.py:584: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    [numpy.zeros(5, dtype=numpy.bool),
```

Turns out there's also int and float versions.

DougBurke:
Actually, this may be a bit-more serious as I'm also seeing

```
1/1 {message : RuntimeWarning('numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject'), category : 'RuntimeWarning', filename : '<frozen importlib._bootstrap>', lineno : 219, line : None}
```

DougBurke:
I'd hoped it wouldn't show up on the pip tests - as part of ce20f5b - but it does

https://github.com/sherpa/sherpa/pull/1089/checks?check_run_id=1807303273

PR #1091, opened 2021-02-01
Simplify implementation of pyfits backend

### LINKED ISSUE #841: refactor pyfits_backend with modern astropy functionality?
The pyfits backend as written in the era of pyfits and it does well what it does. However, pyfits became part of astropy about 8 years ago and today we import the fits code from astropy. That means the every time we use that backend, we also have the full astropy available, and quite some of the code in `pyfits_backend.py` duplicates existing astropy code, in particular `astropy.table.Table` (for reading and writing tables like PHA) and `astropy.wcs` (for all the wcs stuff in the pyfits backend) come to mind. 

On the one hand, I'm reluctant to change anything that's not broken and that relies only on very stable code, i.e. astropy >= 0.4 should be good enough (the pyfits interface has been stable since before astropy, while e.g. Table and WCS have changed in the past but are quite mature now), on the other hand, I know that astropy handles quite a few corner cases and extras that we don't have in this code, e.g. reading of Tables over https instead of a local file.

@DougBurke : Thoughts?

hamogu:
These are the general steps that we can take to reduce the code:

- replace table writing with the `astropy.table.Table` interface
- replace table reading with the same interface
- Can we use `astropy.wcs` instead of the WCS implemented here?

At least as of now, this implemenation is not identical to pycrates
any longer, e.g.
- an astropy specific warning is emitted for file operations
  (e.g. file not found, clobber=False and file exists).
  I think that's OK.
- The creates backend automatically tries to add ".gz" if the filename
  is not found. That requires and extra try .. except that I did not
  put in yet (maybe as a decorator for file reading functions)
- Should header be written? Should keywords that are None be written?
  The pyfits code has opinions on this, this implementation simply
  chooses the path of least resistance. If really needed, thos could be
  added back in.

This is a WIP, I'm putting it up here for general discussion and so I
don't forget. The reason I'm looking at this at all (why youch working code?)
is that I needed and ARF/RMF reader/writer for a different project and was
hoping to use the implementation that's alrady in Sherpa. So, I had to
understand what's in here first.

Also related to
- #1070 (at least astropy table can read from URLs and
  read/write to streams).
- #878

closes #841

Note: We could always leave the pyfits backend untouched (safe!) and move
this to a new backend "astropy". That way we can collect some real world
experience and see if it breaks too much. While I don't think we should
maintain both pyfits backend and this indefinitely, there is little harm
in having both for an extended pariod of time (assuming I ever finish this PR!)

Todo

- [ ] Table reading (especially RMF/ARF!)
- [ ] image reading / writing
- [ ] modify tests because this throws different exceptions now (but may we don't have to text all branches any longer, since astropy itself tests, e.g. that overwrite=False does not overwrite an existing file.

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -44,6 +44,7 @@ import numpy
 from numpy.compat import basestring
 
 from astropy.io import fits
+from astropy.table import Table
 from astropy.io.fits.column import _VLF
 
 from sherpa.utils.err import IOErr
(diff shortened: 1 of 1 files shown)


DougBurke:
I have been mulling over renaimg the `pyfits` backend to `astropy` with a fall-through case, but I think having an all-new `astropy` backend sounds like fun.

DougBurke:
Also the test failures are tracked in #1090 (at least the ones unrelated to this PR ;-)

DougBurke:
Looks like this needs rebasing now that #1092 has been merged, which will hopefully fix the test failures.

hamogu:
This has been overtaken by work by @DougBurke -see #1939. Some of the changes here might still be useful as reference, but it's going to be easier to wrote from scratch then to adapt this, so I'm closing.

PR #1092, opened 2021-02-01
Numpy 1.20 warnings

DougBurke:
# Summary

Avoid test failures due to new warnings added by NumPy 1.20

# Details

This does not "fix" the underlying code, but does allow us to run the CI tests to see things still pass. Further changes will need to be made to address the deprecations that NumPy 1.20 is now warning about.

An example of the tests failing without these tests is https://github.com/sherpa/sherpa/pull/1089/commits/ff3853748ebdd1982459527c53b72fb15197e6c7


Marie-Terrell (approved):
These changes look good to me. It would be good to cleanup any of the code causing these warning (if they are specifically from Sherpa), but this seems okay for now.

PR #1094, opened 2021-02-02
Cleanup multi plot code

DougBurke:
# Summary

Rework the code that handles the plot_fit_xxx and plot_bkg_fit_xxx calls. There should be no user-visible changes.

# Details

I noted in #626 - in particular aa472cbbbe1d1074ead8b6328a37393b9f39fe77 - that the handling of "jointplot" could be improved but that there wasn't enough tests of the functionality. This tries to fix both.

The aim is to avoid direct access to the plot objects (e.g. self._dataplot and self._modelplot) and instead use less-specialised functions to get them (in this case we can use the actual plot objects that are being used to create the plots). We can also assume that all the plot types support the `xlog` setting so there's no need to check for it and, in the background case, we know we are dealing with histogram-style plots so saving some accessor shenanighans.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -3611,7 +3611,7 @@ def test_plot_fit_resid_set_xlog(idval, plottype, xscale, clean_astro_ui):
     """Check that set_xlog handling for plot_fit_resid.
 
     What is the X-axis scaling when you call set_xlog(plottype)?
-    We use a rannge of plottypes as the behavior is not always
+    We use a range of plot types as the behavior is not always
     obvious, so let's ensure we test them.
     """
 
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -3579,3 +3579,94 @@ def test_set_plot_opt_explicit_astro():
     assert fig.axes[3].get_yscale() == 'linear'
 
     plt.close(fig)
+
+
+def check_plot2_xscale(xscale):
+    """Are there two plots, y-axis linear, x axis set?
+
+    Any test using this needs @requires_pylab
+    """
+
+    from matplotlib import pyplot as plt
+
+    fig = plt.gcf()
+    axes = fig.axes
+    assert len(axes) == 2
+    assert axes[0].xaxis.get_label().get_text() == ''
+
+    assert axes[0].xaxis.get_scale() == xscale
+    assert axes[0].yaxis.get_scale() == 'linear'
+
+    assert axes[1].xaxis.get_scale() == xscale
+    assert axes[1].yaxis.get_scale() == 'linear'
+
+
+@requires_pylab
+@pytest.mark.parametrize("idval", [None, "bob"])
+@pytest.mark.parametrize("plottype,xscale", [('data', 'log'),
+                                             ('resid', 'log'),
+                                             ('bkg', 'linear'),
+                                             ('bkgresid', 'linear')])
+def test_plot_fit_resid_set_xlog(idval, plottype, xscale, clean_astro_ui):
+    """Check that set_xlog handling for plot_fit_resid.
+
+    What is the X-axis scaling when you call set_xlog(plottype)?
+    We use a rannge of plottypes as the behavior is not always
+    obvious, so let's ensure we test them.
+    """
+
+    setup_example(idval)
+    ui.set_xlog(plottype)
+    ui.plot_fit_resid(idval)
+    check_plot2_xscale(xscale)
+
+
+@requires_pylab
+@pytest.mark.parametrize("idval", [None, "bob"])
+@pytest.mark.parametrize("plottype,xscale", [('data', 'linear'),
+                                             ('resid', 'linear'),
+                                             ('bkg', 'log'),
+                                             ('bkgresid', 'log')])
+def test_plot_bkg_fit_resid_set_xlog(idval, plottype, xscale, clean_astro_ui):
+    """Check that set_xlog handling for plot_bkg_fit_resid.
+
+    This logic could be added to test_plot_fit_resid_set_xlog but it
+    would complicate the setup, so we duplicate the code.
+
+    """
+
+    setup_example_bkg_model(idval)
+    ui.set_xlog(plottype)
+    ui.plot_bkg_fit_resid(idval)
+    check_plot2_xscale(xscale)
+
+
+@requires_pylab
+@pytest.mark.parametrize("plot,yscale", [('data', 'linear'),
+                                         ('ratio', 'linear'),
+                                         ('fit', 'linear'),
+                                         ('bkg', 'log'),
+                                         ('bkg_resid', 'linear'),
+                                         ('bkg_fit', 'log')])
+def test_set_ylog_bkg(plot, yscale, clean_astro_ui):
+    """Check y axis after_ylog('bkg').
+
+    The idea is to check how separate the "background" from
+    "non-background" plots are.  I use ylog since other tests have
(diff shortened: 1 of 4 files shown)


sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3579,3 +3579,163 @@ def test_set_plot_opt_explicit_astro():
...
+
+
+@requires_pylab
+@pytest.mark.parametrize("idval", [None, "bob"])
+def test_plot_fit_resid_handles_data_log(idval, clean_astro_ui):
+    """Check that log handling is correct: data=log
+
+    See also test_plot_fit_resid_handles_resid_log.
+
+    I thought we had tests of this, but apparently not.
```
hamogu, line 3591: I'm not sure that's a useful comment to have in the code, because - as soon as this is merged - it's not correct any longer: We then will have a test (this one). If you want to leave a comment just for me, it might be better to use GH to leave a comment like I do here, so that it does not end up in the commited code?

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3579,3 +3579,163 @@ def test_set_plot_opt_explicit_astro():
...
+    assert axes[0].yaxis.get_scale() == 'linear'
+
+    assert axes[1].xaxis.get_scale() == 'log'
+    assert axes[1].yaxis.get_scale() == 'linear'
+
+
+@requires_pylab
+@pytest.mark.parametrize("idval", [None, "bob"])
+def test_plot_bkg_fit_resid_handles_data_log(idval, clean_ui):
+    """Check that log handling is correct: data=log"""
```
hamogu, line 3647: The code below actually tests for data=linear

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3579,3 +3579,163 @@ def test_set_plot_opt_explicit_astro():
...
+    fig = plt.gcf()
+    axes = fig.axes
+    assert len(axes) == 2
+    assert axes[0].xaxis.get_label().get_text() == ''
+
+    assert axes[0].xaxis.get_scale() == 'log'
+    assert axes[0].yaxis.get_scale() == 'linear'
+
+    assert axes[1].xaxis.get_scale() == 'log'
+    assert axes[1].yaxis.get_scale() == 'linear'
```
hamogu, line 3741: These are a number of almost identical tests. Maybe they should be parameterized along one more axes?
Something like (untested, not sure about the syntax for passing multiple parnames to parameterize. I know I've done it, I just don't recall right now).
```
@requires_pylab
@pytest.mark.parametrize("idval", [None, "bob"])
@pytest.mark.parametrize("setxlog, xloglin", [('data', 'linear'), ('bkg', 'log'), ('resid', 'linear'), ('bkgresid', 'log'])
def test_plot_bkg_fit_resid_handles_bkgresid_log(idval, clean_ui):
    """Check that log handling is correct:
    MORE GENERAL COMMENT HERE
    """

    setup_example_bkg_model(idval)
    ui.set_xlog(setxlog)  # NOTE: not 'resid'
    ui.plot_bkg_fit_resid(idval)

    fig = plt.gcf()
    axes = fig.axes
    assert len(axes) == 2
    assert axes[0].xaxis.get_label().get_text() == ''

    assert axes[0].xaxis.get_scale() == xloglin
    assert axes[0].yaxis.get_scale() == 'linear'

    assert axes[1].xaxis.get_scale() == xloglin
    assert axes[1].yaxis.get_scale() == 'linear'
```
That makes the tests easier to read and thus maintain.

sherpa/ui/tests/test_ui_plot.py
```diff
@@ -2717,3 +2717,62 @@ def test_plot_fit_xxx_overplot_pylab(ptype, caplog):
...
+    fig = plt.gcf()
+    axes = fig.axes
+    assert len(axes) == 2
+    assert axes[0].xaxis.get_label().get_text() == ''
+
+    assert axes[0].xaxis.get_scale() == 'log'
+    assert axes[0].yaxis.get_scale() == 'linear'
+
+    assert axes[1].xaxis.get_scale() == 'log'
+    assert axes[1].yaxis.get_scale() == 'linear'
```
hamogu, line 2778: See comment above about parameterizing more.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: No. It checks what happens when you've set the data "option/name/class" to log - i.e. set_xlog('data'). It just so happens that this setting isn't used in plot_bkg_fit_xxx (or, rather, I just want code to check what the actual behavior is, as this is one of those cases where I can argue for the x axis being linear and log).

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: FYI, I'm pretty sure you need to remove the space to get "setxlog,xloglin" because I think the parser complains with spaces, and you need to add setclog and xloglin parameters to the function (I don't think the relative ordering matters but I try to follow the order of the parametrize call.

It is always easier to copy-n-paste, particluarly when you don't know for sure how the code is meant to behave!

hamogu: I agree that it's easier to write by copy&paste, but it's often easier to maintain when it's shorter. When I read through this I was wondering: How do these tests differ? What's different and what's the same? It's there something subtle that I'm missing? With a parameterization, it's easier to see that these are the same test, just applied to different plots.

But either works and you are working on plotting, so up to you!

DougBurke: Doing the refactoring has uncovered a subtle bug - I should have been using clean_astro_ui not clean_ui (well, I had caught this in 71e18221a404a37e3ceba1aadbc6d84a80cd2639 but I thought that I'd pushed it through to the first commit). That fixes one of the caveats I think,

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: Although this is now outdated with the refactoring you suggested below.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: Done.

sherpa/ui/tests/test_ui_plot.py
DougBurke: Done.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: This has been removed in the refactoring,

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3579,3 +3579,94 @@ def test_set_plot_opt_explicit_astro():
...
+@pytest.mark.parametrize("idval", [None, "bob"])
+@pytest.mark.parametrize("plottype,xscale", [('data', 'log'),
+                                             ('resid', 'log'),
+                                             ('bkg', 'linear'),
+                                             ('bkgresid', 'linear')])
+def test_plot_fit_resid_set_xlog(idval, plottype, xscale, clean_astro_ui):
+    """Check that set_xlog handling for plot_fit_resid.
+
+    What is the X-axis scaling when you call set_xlog(plottype)?
+    We use a rannge of plottypes as the behavior is not always
```
hamogu, line 3614: range - minor typo

hamogu (approved):

PR #1096, opened 2021-02-02
Reduce direct access to plot objects

DougBurke:
# Summary

Changes to the internals of the plot code, to access information via methods rather than direct access, which improves encapsulation and reduces code repetition.

# Details

This is another "small" change from the larger "plot rework" I'm doing (e.g. #1089).

There's two parts in the commit

a) Accessors

The simple case is something like `get_data_contour_prefs` which has gone from

    return self._datacontour.contour_prefs

to

    return self.get_data_contour(id, recalc=False).contour_prefs

The real benefit is for plots which have more-complex requirements (e.g. data plots require a DataPHA check in the astro-ui layer and then falling through to the ui layer where there's a check for Data1DInt). By using the `get_<>_plot/contour` calls rather than direct access we can avoid repeating this logic and, in some cases, avoid the need for subclassing the routines in the first place.

b) In a number of routines the data-accessor has been changed so that it no-longer errors out if the dataset does not exist. 

The change is basically from `d = get_data(id)` to

```
try:
    d = get_data(id)
except IdentifierErr as ie:
    if recalc:
        raise ie
    d = None
```

I added a test to check this functionality (it's related to source/model/source_component/model_component plots).

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020, 2021
-#      Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -10231,34 +10231,6 @@ class Session(sherpa.ui.utils.Session):
     # Plotting
     ###########################################################################
 
-    def get_data_plot_prefs(self, id=None):
-
-        try:
-            d = self.get_data(id)
-            if isinstance(d, sherpa.astro.data.DataPHA):
-                return self._dataphaplot.histo_prefs
-
-        except IdentifierErr:
-            pass
-
-        return super().get_data_plot_prefs(id)
-
-    get_data_plot_prefs.__doc__ = sherpa.ui.utils.Session.get_data_plot_prefs.__doc__
-
-    def get_model_plot_prefs(self, id=None):
-
-        try:
-            d = self.get_data(id)
-            if isinstance(d, sherpa.astro.data.DataPHA):
-                return self._modelhisto.histo_prefs
-
-        except IdentifierErr:
-            pass
-
-        return super().get_model_plot_prefs(id)
-
-    get_model_plot_prefs.__doc__ = sherpa.ui.utils.Session.get_model_plot_prefs.__doc__
-
     def get_data_plot(self, id=None, recalc=True):
         try:
             d = self.get_data(id)
@@ -10367,7 +10339,13 @@ class Session(sherpa.ui.utils.Session):
 
         """
 
-        d = self.get_data(id)
+        try:
+            d = self.get_data(id)
+        except IdentifierErr as ie:
+            if recalc:
+                raise ie
+            d = None
+
         if isinstance(d, sherpa.astro.data.DataPHA):
             plotobj = self._astrosourceplot
             if recalc:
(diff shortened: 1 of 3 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -10433,7 +10417,13 @@ def get_source_component_plot(self, id, model=None, recalc=True):
...
             model = self._eval_model_expression(model)
 
-        d = self.get_data(id)
+        try:
+            d = self.get_data(id)
+        except IdentifierErr as ie:
+            if recalc:
+                raise ie
+            d = None
+
```
hamogu, line 10426: Since this code turns up three times, should it be a separate function?

hamogu (commented):
One stylistic question, looks good overall. I note that calling methods might have a different performance characteristic, but even if that's slightly worse (I did not time it), the better encapsulation would be worth it.

sherpa/astro/ui/utils.py
DougBurke: I think I'd like to check that we're doing something sensible, and worth encapsulating, first!

Maybe, in this case, we should require a dataset to be present, to make sure that you're changing the "correct" plot type (the datatype-specific plots is new functionality added in 4.12.2 or 4.13.0, I forget which)? I don't like that (I want to be able to set prefs before loading data), but requiring a dataset is cleaner.

Or, we can add a dtype parameter (name open for bikeshedding) which defines the plot type to change? Then we need to decide whether it replaces id or only comes into play if there is no dataset available.

Or we can say "when you change the plto preferences, you change all datatype preferences" (modulo some issues about keys not being always present), but that is *hard* to retrofit into the existing system.

hamogu: I apparently did not understand the intention of this change. What is the point of `get_xxx_plot` when there is no data to be plotted? It's to get it of the `get_plot_prefs` functions, right? If so, the code as suggest will do that (as you showed). I know you are working on a bigger development that goes beyond that, so I don't know exactly how that fits in there.

Conceptually, `get_plot_xxx` makes more sense if there is data to plot. So, I guess we should require data to be present, however I'll comment more in #1096.

DougBurke: Previously, we had a patch-work of code that didn't make it easy to work out responsibility, and the same bit of logic was scattered about. I am trying to make the `get_xxx_plot` calls be the place that maps from identifier to relevant plot object. These calls have existed for a long time (i.e. they're not new).

The move to plot-specific types has complicated the interface, but it already somewhat existed (handling of PHA data).

Whether this is the correct semantics of the `get_xxx_plot` (and other routines) is open to question.

hamogu: To be clear: We can just accept this and move and possibly rework it in the next PR. This is really just a minor suggest and I don't want to hold up all the rest of the PR for this one thing that can be decided later. It's really only a few lines. So, I'm going to accept this and continue the discussion in #1097.

hamogu (approved):

sherpa/ui/utils.py
```diff
@@ -10664,15 +10667,11 @@ def get_data_plot_prefs(self, id=None):
...
-            d = self.get_data(id)
-            if isinstance(d, sherpa.data.Data1DInt):
-                return self._datahistplot.histo_prefs
-
-        except IdentifierErr:
-            pass
-
-        return self._dataplot.plot_prefs
+            return plot.histo_prefs
+        except AttributeError:
```
wmclaugh, line 10673: Is the more common case here typically the dataset having the histo_prefs or raising the Attribute error? If we expect the latter, perhaps using hasattr() might be more effiicient.

hamogu: Either way, this is not performance-critical. It's something that will come up no more than a dozen times (no more than a few time per plot if someone runs a script to automatically generate a lot of plots) in a typical few-hour analysis session and probably save < 0.1 ms each time. Discussing this might well take longer than the potential performance gain for all sherpa users ever combined ;-)

DougBurke: @wmclaugh - there is a cottage industry on saying whether to use hasattr or not - e.g. https://hynek.me/articles/hasattr/

PR #1097, opened 2021-02-02
WIP Simplify plot preference access

DougBurke:
# Summary

Add an explicit get_prefs method to return the plot preferences.

# Details

The idea is to see whether we can simplify access to the plot preferences, which I think this does. The question is whether this is the sensible way to do this, versus just adding a ui-layer function like

```
def get_prefs(plotobj):
    try:
        return plotobj.histo_prefs
    except AttributeError:
        return plotobj.plot_prefs
```

which is less generic but probably does most of what we need. Alternatively, should I go the whole way and add a `Preferences` class with a single `get_prefs` entry and then use multiple inheritance to pull it into the plot classes?

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -12435,16 +12435,17 @@ class Session(sherpa.ui.utils.Session):
             # Unlike the plot version we can assume we are dealing
             # with histogram plots here.
             #
-            oldval = plot2.plot_prefs['xlog']
-            dprefs = plot1.dataplot.histo_prefs
-            mprefs = plot1.modelplot.histo_prefs
+            prefs2 = plot2.get_prefs()
+            oldval = prefs2['xlog']
+            dprefs = plot1.dataplot.get_prefs()
+            mprefs = plot1.modelplot.get_prefs()
 
             if dprefs['xlog'] or mprefs['xlog']:
-                plot2.plot_prefs['xlog'] = True
+                prefs2['xlog'] = True
 
             self._jointplot.plotbot(plot2, overplot=overplot, **kwargs)
 
-            plot2.plot_prefs['xlog'] = oldval
+            prefs2['xlog'] = oldval
         except:
             sherpa.plot.exceptions()
             raise
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020, 2021
-#      Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -10231,34 +10231,6 @@ class Session(sherpa.ui.utils.Session):
     # Plotting
     ###########################################################################
 
-    def get_data_plot_prefs(self, id=None):
-
-        try:
-            d = self.get_data(id)
-            if isinstance(d, sherpa.astro.data.DataPHA):
-                return self._dataphaplot.histo_prefs
-
-        except IdentifierErr:
-            pass
-
-        return super().get_data_plot_prefs(id)
-
-    get_data_plot_prefs.__doc__ = sherpa.ui.utils.Session.get_data_plot_prefs.__doc__
-
-    def get_model_plot_prefs(self, id=None):
-
-        try:
-            d = self.get_data(id)
-            if isinstance(d, sherpa.astro.data.DataPHA):
-                return self._modelhisto.histo_prefs
-
-        except IdentifierErr:
-            pass
-
-        return super().get_model_plot_prefs(id)
-
-    get_model_plot_prefs.__doc__ = sherpa.ui.utils.Session.get_model_plot_prefs.__doc__
-
     def get_data_plot(self, id=None, recalc=True):
         try:
             d = self.get_data(id)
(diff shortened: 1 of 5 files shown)


hamogu:
What is the use case for functions to access a object at all, e.g.:
```
def get_plot_prefs(self):
   return self.plot_prefs
```
What can this construct do that we can't achieve with just working with `self.plot_prefs` directly? That looks like significant duplication.
(I'll probably have more questions later, I'm just trying to understand first.)

DougBurke:
For historical reasons the preferences are stored in a xxx_prefs field (e.g. `plot_prefs`, `histo_prefs`, `point_prefs`, `contour_prefs`, maybe more). This used to be fine, since we always knew what plot "type" we were working with (e.g. always a plot or histogram type). With the move to the plot type depending on the data type, I started having to write code like

    try:
        prefs = plotobj.histo_prefs
    except AttributeError:
        prefs = plotobj.plot_prefs

which strikes me as poor design (we shouldn't have to know what plot type we are using here).

An alternative approach would be to drop the prefix for the `xxx_prefs` attribute, so we would just use `plotobj.prefs`. However, there are problems with this as we have plots with multiple inheritance,  such as

```
class Confidence2D(DataContour, Point):

    contour_prefs = backend.get_confid_contour_defaults()
    point_prefs = backend.get_confid_point_defaults()
```

Maybe these could instead of using multiple inheritance just act as container types, but that's a larger question.

However, these are the questions I want asked here!

hamogu:
What is the use case for functions to access a object at all, e.g.:
```
def get_plot_prefs(self):
   return self.plot_prefs
```
What can this construct do that we can't achieve with just working with `self.plot_prefs` directly? That looks like significant duplication.
(I'll probably have more questions later, I'm just trying to understand first.)

DougBurke:
@hamogu - did you mean to duplicate your question?

hamogu:
No, sorry. My browser restarted and must have re-posted when I re-loaded the page.

DougBurke:
Yay for mutability!

hamogu:
From a user perspective, I wonder where (without thinking too much) I would expect the preferences to live. If a plot pref a preference for Confidence2D or it is a preference for DataContour?

In other words, if I change the linewidth in the preferences of a some plot, do I expect that to change only the line width in the fit_res plot, or in all plot that use lines (model, fit [which has the model in it], resid, chi2)? If the latter, then maybe Condidence2D does not need to have any plot_pref on it's own?

hamogu:
Or can we just merge all prefs? Something like (psedo-code, order of calling new and init need to be fixed and some super() calls be inserted)
```
def __new__(cls, ...)
    prefs = {}
    for parent in reversed(cls.__bases__):
        prefs.update(parent.prefs)
def __init__(..., kwargs):
    prefs.update(kwargs)
```

hamogu:
Oh, does not need __new__, can just be done in __init__. Simpler!
```
def __init__(...):
    for parent in reversed(self.__class__.__bases__):
        prefs.update(parent.prefs)
```

DougBurke:
For reference, the preferences are taken from the associated backend code (for "plots", "histogram", "plot", "contour" types, possibly more) with code like

```
class Contour(NoNewAttributesAfterInit):
    "Base class for contour plots"
    contour_prefs = backend.get_contour_defaults()
...
```

This doesn't affect @hamogu's thoughts on merging preferences in the constructor of sub-classes, but I just wanted to be explicit.

As to the discussion of where the preferences are set, there are two use cases here

a) direct access to the plot objects by a user manually creating an instance of a class
b) for users of the utils session objects, which add some wrinkles: these session objects store instances of these plot objects, so preference changes are applied to instances and not the class, which has some downstream consequences

 - we can easily restore the defaults (via the clean methofd) which just creates new plot instances; this requires not having messed with the class preferences
 - we have a set of "named" plots (e.g. 'data') which can be used to set the preferences of a subset of the plot instances

I don't think either of these necessarily stop a design for the classes similar to what @hamogu is playing with, but we need to be aware of the current use case.

hamogu:
> For reference, the preferences are taken from the associated backend code (for "plots", "histogram", "plot", "contour" types, possibly more) with code like

```
 class Contour(NoNewAttributesAfterInit):
    "Base class for contour plots"
    contour_prefs = backend.get_contour_defaults()
```

That might cause problems, if the backend is switched. Once a `Contour` object exists, what is going to happen when we switch to a different backend? As written above, the countour_prefs are going to reflect the first backend  and may contain options that the second backend can't deal with.

Two solutions might be
1) The preferences contain only information that every backend can deal with - in other words, we define a minimum set of preferences (color, linestyle, linewidth, ...) that every backend implementation needs to work with, where "work with" means that the backend code can't crash - it either respects that setting or ignores it gracefully. In that case, the preferences are backend independent. They can be an attribute of the plot class. That has several advantages:

 - It's easy to find for the user.
 - Changes cna be made for one plot (on the plot object) or all of them (on the plot class)
 - decouple backend from plot definition as much as possible.

It makes it harder to have backend specific preferences (use blue lines in matplotlib, but red lines in bokeh), but that's OK, I think.
2) Preferences live in the backend. If a plot object needs them, they are looked up dynamically, and if they are set, they are passed on to the backend, which can check if they are valid etc. I think that's complicated.

DougBurke:
I'm closing this because some of the code here has now been merged, the actual change left is relatively small, and there are other larger plot PRs which are going to cause a conflict, and it's easier to just re-create this once the dust has settled, **if** we feel like it's important (and the discussion so far hasn't really felt like we feel it's that important relative to other work).

PR #1099, opened 2021-02-03
Do not download miniconda manually in GitHub actions

DougBurke:
# Summary

Update the GitHub conda action so that we use the setup-miniconda action rather than manually installing miniconda. This only affects the GitHub-run tests.

# Details

I've separated this from #1069 but the original work to use the `setup-miniconda` action was developed there.

I am not convinced with the cache code - see https://github.com/sherpa/sherpa/pull/1069#issuecomment-772613598

A note I added in #1069:

The rework of the existing conda workflow has moved the conda configuration to etc/environment-<platform>.yml files to simplify the download (we try to have everything in one place). There are actually issues with the conda workflow (not really from this PR)

 - I've had to rely on the conda pytest packages rather than use pip, as there were problems when mixing the two
 - I've tried really hard to move to a more-modern build setup (build with pip, test with pytest) but this just kept on failing for a number of reasons and I've decided (having tried this before) that we need to fix this issue separately
 - the existing case of install-type=install and test=submodule is problematic, since the test run (which uses setuptools) actually ends up re-building the code (because it needs a develop build). So we should review changing these to install-type=develop or improving the build


Marie-Terrell:
The PR looks good. Though I did note that there was an issue with the dependencies installed by Conda echoed to the files (introduced by me in the switch to GitHub Actions). This just made it so that the fits packages would never get echoed in the dependencies list. 
As there is not really a point in having separate "FITS" and "FITSBUILD" variables, these have now been combined. And as long as I did not mess anything up in the process and you are also on board with those updates, then we might want to take these updates too. If not, fee free to revert. I'll wait for the tests to complete and then approve the PR.

DougBurke:
@Marie-Terrell - I think its worth thinking about whether the way I've set up the conda environment files is sensible or not, as I just followed the suggestion of a random piece of internet advice here!

Perhaps they should go into .github/ rather than etc/, given that this is really nly used for GH actions, or we can bike-shed the directory name. Other questions are

 - should we (or do we want to) add some version constraints in them?
 - are they expressive enough (I think there was some code I was playing around with where we still needed to manually extend the conda environment beyond these files, which is not necessarily a bad thing, but we should have more of a think about it than I gave in developing this code ;-)

Marie-Terrell:
Doug, that is a good point. 
I guess if it were in the etc  directory, we may want to have a more "specific" name (referring to the fact that they are used for testing). Though, moving them to the scripts area (or a new "envs" folder) in the .github folder would help with that as well. I think doing both may be a good idea.

In terms of the contents, I think that is good. We do not want to pin down these "build" OTS unless we have to. Though, there is always the possibility that we run into pytest version issues again.

Adding the other OTS (matplotlib, numpy, etc.) as part of the setup_conda.sh is good as those are the ones we want to have the ability to select particular versions or packages. 

Other than that, my only note would be that typically, I have seen the environments having a name section at the top. It may be a good idea to add that too (with the name "build" to correspond to the name used in the environment activation). That would likely get rid of this warning in the tests:

>  Warning: The environment name on "environment-file" is not defined, using "enviroment-activate"!

DougBurke:
I've rebased this because there was a conflict and because I want to see if this suffers from the same conda issues we see with existing PRs (e.g. this test from #1106 - https://github.com/sherpa/sherpa/pull/1106/checks?check_run_id=2296631706

Marie-Terrell:
There seems to be a bit of an order of events issue that may make this a bit difficult. Here are a few bits of info:

- The two numpy 1.18 tests fail with this current setup because the compilers and python get installed first. Even if I remove all of the other things in the environment file except the compilers (python gets added by the yml file), create the env, and then try to just conda install numpy=1.18.* this fails. However, if I install numpy at the same time as the compilers and python, this is no longer an issue.

- If I update one of those tests to 1.19, that test now passes.

- I assume this is because of the compiler/python updates (like what we were seeing on the conda build side).

If we want to move forward with this, we would need to figure out a way to resolve this. This might work if numpy-version could be selected like the python-version at environment creation time. However, that does not seem to be the case (though maybe they would accept a request for that)?

I am a bit more hesitant to the switch seeing this behavior, though I am not entirely against it because this could expose potential GitLab issues eariler. Either way, this does not seem possible at the moment unless we want to consider bumping the numpy 1.18 tests to 1.19. Which, I am open for discussions on.

DougBurke:
I haven't been tracking this issue since I last looked at this, so I don't know the state of the GitHub action I proposed above (ie is it still being updated?) or the latest issues that @Marie-Terrell has noted. I am unlikely to spend time on this in the near future, instead focusing on code (rather than general infrastructure) changes.

Marie-Terrell:
The last time they update the action is back in march, so I cannot necessarily speak to if it is still being maintained. It seems like a nice switch in theory, but it seems to come with its own set of issues (at least in our set up) that are not easily worked around. For now, I think it is project we can abandon to be re-examined at a later date (if the action is still maintained and when we have a bit more time).

DougBurke:
I'll close this down then and we can come back to it in our infinite free time...

PR #1101, opened 2021-02-03
Address upper limit issues with sample_flux (fix #457)

### LINKED ISSUE #457: sample_flux ignoring flux=0 is biasing results
From CIAO helpdesk ticket 21264, a user found that the median value returned by sample_flux was not the same as taking the median of the individual realizations.  For details and files see /data/lenin2/HelpDeskPub/21264/SampleFluxLimits.ipynb .

They surmised that flux=0 values were being excluded when the median value is computed.

Digging through the code it does look like flux values generated when the model parameter are *at* or beyond their soft limits are excluded from the median, eg

https://github.com/sherpa/sherpa/blob/d8b58d91b17e786a7f77f67d7b0241a9265d3c76/sherpa/astro/flux.py#L103-L111

Since the user is doing an upper-limits kind of analysis, their model's normalization were indeed hitting 0, which is the hard limit for the parameters.  Since 0 is the limit, all those samples were excluded when the median was computed which then biases the final flux values higher than it should be.

It's not clear if allowing model values at the limit to be included would be sufficient (ie < vs <= kind of thing).

DougBurke:
# Summary

sample_flux no-longer excludes samples at the parameter bounds (soft) when calculating the flux distribution. This could lead to an over-estimation of the flux for upper limits (Fix #457).

# Address

This is pulled from #754 and there are more changes planned to clean up the code (as it is doing stuff it needn't do), but this is a nice improvement that is easy to review. It also adds some documentation.

Technically the flux distribution when there's an upper or lower limit could be underestimated or overestimated but it's easier to just say "overestimate" in the summary. We can see the overestimate in the test_sample_flux_457 test where the estimated flux values go from

```
assert flux1[0] == pytest.approx(7.5243e-13)
assert flux1[1] == pytest.approx(7.5243e-13 + 3.53865e-14)
assert flux1[2] == pytest.approx(7.5243e-13 - 3.01841e-14)

assert flux2[0] == pytest.approx(8.26722e-13)
assert flux2[1] == pytest.approx(8.26722e-13 + 5.88881e-14)
assert flux2[2] == pytest.approx(8.26722e-13 - 4.99602e-14)
```

to

```
assert flux1[0] == pytest.approx(7.50199e-13)
assert flux1[1] == pytest.approx(7.50199e-13 + 3.64329e-14)
assert flux1[2] == pytest.approx(7.50199e-13 - 2.87033e-14)

assert flux2[0] == pytest.approx(8.19482e-13)
assert flux2[1] == pytest.approx(8.19482e-13 + 5.42782e-14)
assert flux2[2] == pytest.approx(8.19482e-13 - 6.87875e-14)
```

### DIFF
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -442,7 +442,7 @@ def sample_flux(fit, data, src,
 
     See Also
     --------
-    calc_flux
+    calc_flux, calc_sample_flux
 
     Notes
     -----
@@ -529,6 +529,85 @@ def sample_flux(fit, data, src,
 
 def calc_sample_flux(id, lo, hi, session, fit, data, samples, modelcomponent,
                      confidence):
+    """Given a set of parameter samples, estimate the flux distribution.
+
+    This is similar to sample_flux but returns values for both the full
+    and component models (which can be the same). The set of parameter
+    samples is searched to remove rows where the parameters lie outside
+    the soft limits, and then this set of rows is used to calculate the
+    flux of the modelcomponent expression (the flux for the full model
+    is taken from the samples argument). The resulting set of fluxes is
+    used to calculate the median and the confidence range.
+
+    .. versionchanged:: 4.13.1
+       The clipping now includes parameters at the soft limits;
+       previously they were excluded which could cause problems with
+       parameters for which we can only calculate an upper limit.
+
+    Parameters
+    ----------
+    id : int or str
+        The dataset identifier. It must exist and have an associated model in
+        session.
+    lo : number or None, optional
+        The lower edge of the dataspace range for the flux calculation.
+        If None then the lower edge of the data grid is used.
+    hi : number or None, optional
+        The upper edge of the dataspace range for the flux calculation.
+        If None then the upper edge of the data grid is used.
+    session : sherpa.ui.utils.Session instance
+        Defines the data, model, and fit.
+    fit : sherpa.fit.Fit instance
+        The fit object. The src parameter is assumed to be a subset of
+        the fit.model expression (to allow for calculating the flux of
+        a model component), and the fit.data object matches the data
+        object. The fit.model argument is expected to include instrumental
+        models (for PHA data sets). These objects can represent
+        simultaneous fits (e.g. sherpa.data.DataSimulFit and
+        sherpa.models.model.SimulFitModel).
+    data : sherpa.data.Data subclass
+        The data object to use. This is not a DataSimulFit instance.
+    samples
+        The output of sample_flux for the data (assumed to have been
+        created with method=calc_energy_flux and clip='hard').
+    modelcomponent : sherpa.models.Arithmetic instance
+        The source model (without instrument response for PHA data) that
+        is used for calculating the flux. This is not a SimulFitModel
+        instance.
+    confidence : number
+        The confidence value to return, as a percentage (0-100).
+
+    Returns
+    -------
+    (fullflux, cptflux, vals)
+       The fullflux and cptflux arrays contain the results for
+       the full source model and the flux of the `modelcomponent`
+       argument (they can be the same). They have three elements
+       and give the median value, upper quartile, and lower
+       quartile values of the flux distribution. The vals array
+       has a shape of ``(num+1, N+2)``, where ``N`` is the number of
+       free parameters and num is the `num` parameter. The rows of
+       this array contain the flux value for the iteration (for
+       the full source model), the parameter values, and then the
+       statistic value for this set of parameters.
+
+    See Also
+    --------
+    calc_flux, calc_sample_flux
+
+    Notes
+    -----
+    This function is in the process of being documented and
+    re-written, as it contains a lot of redundancy and the semantics
+    of some of the options are unclear.
+
+    The summary output displayed by this routine - giving the median
+    and confidence ranges - is controlled by the standard Sherpa
+    logging instance, and can be hidden by changing the logging to a
+    level greater than "INFO" (e.g. with
+    `sherpa.utils.logging.SherpaVerbosity`).
+
+    """
 
     def simulated_pars_within_ranges(mysamples, mysoftmins, mysoftmaxs):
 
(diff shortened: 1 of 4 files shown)


sherpa/astro/flux.py
```diff
@@ -529,12 +529,91 @@ def sample_flux(fit, data, src,
...
+    .. versionchanged:: 4.13.1
+       The clipping now includes parameters at the soft limits;
+       previously they were excluded which could cause problems with
+       parameters for which we can only calculate an upper limit.
+
+    Parameters
+    ----------
+    id : int or str
+        The dataset identifier. It must exist and have an associated model in
+        session.
```
hamogu, line 551:
```suggestion
        the session.
```

sherpa/astro/flux.py
```diff
@@ -529,12 +529,91 @@ def sample_flux(fit, data, src,
...
+    lo : number or None, optional
+        The lower edge of the dataspace range for the flux calculation.
```
hamogu, line 553: I have not come across the term "dataspace" a lot as a sherpa user. I know we use it internally in specs, in the code, and in our discussions, but I'm not sure this wording is helpful for the user. I don't have any better ideas though... I mean, could this number even be outside the range of the datagrid? Maybe a functional explanation "Use this parameter to do X" would be helpful in addition to the current text.

sherpa/astro/flux.py
```diff
@@ -529,12 +529,91 @@ def sample_flux(fit, data, src,
...
+        If None then the lower edge of the data grid is used.
+    hi : number or None, optional
+        The upper edge of the dataspace range for the flux calculation.
+        If None then the upper edge of the data grid is used.
+    session : sherpa.ui.utils.Session instance
```
hamogu, line 558:
```suggestion
    session : sherpa.ui.utils.Session or sherpa.astro.ui.utils.Session instance
```

sherpa/astro/flux.py
hamogu: Or is that something for later? (Since @DougBurke pulled this out of a larger change).

sherpa/astro/flux.py
```diff
@@ -529,12 +529,91 @@ def sample_flux(fit, data, src,
...
+    confidence : number
+        The confidence value to return, as a percentage (0-100).
+
+    Returns
+    -------
+    (fullflux, cptflux, vals)
+       The fullflux and cptflux arrays contain the results for
+       the full source model and the flux of the `modelcomponent`
+       argument (they can be the same). They have three elements
+       and give the median value, upper quartile, and lower
```
hamogu, line 586: At least in the current version, the code only returns quartiles, if I set `confidence=50`, with e.g. `confidence=90` I get the 5-95% range, not the quartiles.

hamogu (changes requested):
There are some inconsistencies in the documentation. I don't know if that's because it's pulled out of a larger changeset or if it's something that should be addressed here.

sherpa/astro/flux.py
DougBurke: I am reluctant to fix this as I plan to remove this parameter (and a bunch of other ones) in a forthcoming PR...

sherpa/astro/flux.py
DougBurke: This text was taken from the existing documentation from other functions in sherpa.astro.flux, so it is at least consistent. I bet if you look through other bits of code you would find this elsewhere.

I am reluctant to spend significant amount of time on tweaking and discussing documentation in these low-level routines without an effort to improve the "intended" user-layer code (e.g. the ui modules). Any change we make for these would need to be reflected in all those routines.

https://github.com/sherpa/sherpa/blob/b7cb202c27951c6dc2ee2ef0a2b661f1ff80c0db/sherpa/astro/ui/utils.py#L13040-L13042 which is the sherpa.astro.ui.sample_energy_flux routine uses the term "data grid" rather than "dataspace", bt I would argue that `dataspace` is a Sherpa term, since we have `sherpa.astro.ui.dataspace1d` - e.g. https://github.com/sherpa/sherpa/blob/b7cb202c27951c6dc2ee2ef0a2b661f1ff80c0db/sherpa/astro/ui/utils.py#L567

sherpa/astro/flux.py
DougBurke: Well, as sherpa.astro.ui.utils.Session is derived from sherpa.ui.utils.Session I don't feel the need to include all possible subclasses here! I have never felt comfortable with the existing documentation on how to document Python docstrings as I never feel I understand what to do for this very case (amongst other things)

sherpa/astro/flux.py
hamogu: You asked for a review of this PR, not all docstring in the entire sherpa package, so I give you one ;-) 
But that makes sense, fine with me.

sherpa/astro/flux.py
hamogu: Accepted.

sherpa/astro/flux.py
DougBurke: This is a doc error - and is also present in the more-likely-to-be-used `sherpa.astro.ui.sample_flux` docstring (from where this was shamelessly copied) 

https://github.com/sherpa/sherpa/blob/b7cb202c27951c6dc2ee2ef0a2b661f1ff80c0db/sherpa/astro/ui/utils.py#L13317-L13329

hamogu: Can I ask you to fix the doc in `sample_flux` in this PR as well, since we now found it?

sherpa/astro/flux.py
DougBurke: Well, actually my real hope is to be able to remove this function completely...

sherpa/astro/flux.py
DougBurke: Yeah, I just need the time.

DougBurke: I've made a change. I'm not proud of the change, but I've made it.

hamogu (approved):

PR #1104, opened 2021-02-09
Add basic documentation for the low-level optimisation code

DougBurke:
# Summary

Add and update documentation on the interfaces used by the low-level optimization code. Ensure that the plot classes are fully included in the documentation.

# Details

There are two separate changes here:

- updates to the plot documentation to address changes (primarily from #931/#906 but there are some earlier updates that have never been picked up)
- added minimal documentation for the low-level optimization code, to provide context for users (it does not attempt to document the existing functions as this is not my code and I have nothing sensible to say for these functions)

Although intended as being a change to the Read-The-Docs documentation, there are changes to the code, primarily

- changing the export order of symbols (e.g. the order of __all__) which was done to make it easier to keep the RTD docs matching the actual code (and the existing setup was a cause of "fun" in recent rebase work I have had to do)
- adding "all" the classes to the export list (except for some obvious helper classes); this can leak out to user code, but only if you've used either of 'from sherpa.all import *' or 'from sherpa.astro.all import *' in which case I think all bets are off.


sherpa/optmethods/__init__.py
```diff
@@ -17,6 +18,86 @@
...
+name         = simplex
+ftol         = 1.1920928955078125e-07
+maxfev       = None
+initsimplex  = 0
+finalsimplex = 9
+step         = None
+iquad        = 1
+verbose      = 0
+
+A model is fit by sengdg the ``fit`` method a callback, the starting
```
hamogu, line 38:
```suggestion
A model is fit by sending the ``fit`` method a callback, the starting
```

sherpa/optmethods/__init__.py
```diff
@@ -17,6 +18,86 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+"""Optimization classes.
+
+The `OptMethod` class provides an interface to a number of optimisers.
+When creating an optimizer an optional name can be added; this name is
+only used in string representations of the class:
+
+>>> opt = NelderMead()
```
hamogu, line 27: Add import statement from NelderMead, so I can see that it comes from this module.

hamogu (changes requested):
Just a typo and one import statement missing. Otherwise, this is great!

DougBurke:
I have got myself the energy to copy over documentation from __init__.py to optfcts.py (as the former is a wrapper around the latter), at least for those functions where it makes sense, so I'll fold the doc fixes in with them,

dtnguyen2 (commented):
In the _docs/optimisers/index.rst file_, perhaps it's best to say that the Levenberg-Marquardt  is to minimize ```ChiSqr``` stat 
```
 NelderMead           single-shot   fast       OK for refining minima
 Levenberg-Marquardt  single-shot   fast       OK for refining minima
 GridSearch           scatter-shot  slow       OK for smooth functions
 Monte Carlo          scatter-shot  very slow  Good in many cases   
 Monte Carlo          scatter-shot  very slow  Good in many cases
```
After all the example in the _sherpa/optmethods/optfcts.py_ file

```
>>> y = np.asarray([3, 2, 7])
 >>> def cb(pars):
 ...     'Least-squares statistic value from fitting a constant model to y'
 ...     dy = y - pars[0]
 ...     dy *= dy
 ...     return (dy.sum(), dy)
 ...
```
cannot possibly minimize non-ChiSqr statistics such as ```Cash``` or ```Cstat```

sherpa/optmethods/optfcts.py
```diff
@@ -583,6 +735,204 @@ def random_start(xmin, xmax):
 def neldermead(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None,
                initsimplex=0, finalsimplex=9, step=None, iquad=1,
                verbose=0):
+    r"""Nelder-Mead Simplex optimization method.
+
+    The Nelder-Mead Simplex algorithm, devised by J.A. Nelder and
+    R. Mead [1]_, is a direct search method of optimization for
+    finding local minimum of an objective function of several
```
hamogu, line 742:
```suggestion
    finding the local minimum of an objective function of several
```

sherpa/optmethods/optfcts.py
```diff
@@ -583,6 +735,204 @@ def random_start(xmin, xmax):
...
+    variables. The implementation of Nelder-Mead Simplex algorithm is
```
hamogu, line 743:
```suggestion
    variables. The implementation of the Nelder-Mead Simplex algorithm is
```

sherpa/optmethods/optfcts.py
```diff
@@ -583,6 +735,204 @@ def random_start(xmin, xmax):
...
+    a variation of the algorithm outlined in [2]_ and [3]_. As noted,
+    terminating the simplex is not a simple task:
+
+    "For any non-derivative method, the issue of termination is
+    problematical as well as highly sensitive to problem scaling.
+    Since gradient information is unavailable, it is provably
```
hamogu, line 749:
```suggestion
    Since gradient information is unavailable, it is probably
```
(Unless the typo is part of the original reference)

sherpa/optmethods/optfcts.py
```diff
@@ -583,6 +735,204 @@ def random_start(xmin, xmax):
...
+    r"""Nelder-Mead Simplex optimization method.
+
+    The Nelder-Mead Simplex algorithm, devised by J.A. Nelder and
+    R. Mead [1]_, is a direct search method of optimization for
+    finding local minimum of an objective function of several
+    variables. The implementation of Nelder-Mead Simplex algorithm is
+    a variation of the algorithm outlined in [2]_ and [3]_. As noted,
+    terminating the simplex is not a simple task:
+
+    "For any non-derivative method, the issue of termination is
```
hamogu, line 747: Is this is directed quote from [2] or from [3]?

DougBurke: I have no idea. This stuff is just copied over from ___init__.py

sherpa/optmethods/optfcts.py
DougBurke: I have no idea. This stuff is just copied over from `__init__.py`

DougBurke: I'm pretty sure the text is correct: https://www.merriam-webster.com/dictionary/provably

hamogu: It always pays-off to learn from a native speaker. I learnt a new word today.

DougBurke:
This has been rebased so that the tests will pass and a few minor doc updates to address comments from @hamogu and @dtnguyen2

sherpa/optmethods/optfcts.py
DougBurke: Just to note for anyone reading this in the future that I'm not planning to address this issue in this PR.

DougBurke:
@hamogu - unfortunately the rebase means it's hard to track the changes, but I believe I've addressed all the basic comments you made on readability. Any remaining issues are more on the document structure/meaning, which we're punting to another time™.

hamogu (approved):

DougBurke:
@wmclauch - the provable -> probably change you just pushed in b61eaea is worng - it should have stayed as provably (as in https://www.merriam-webster.com/dictionary/provably )

wmclaugh:
@DougBurke - my bad. I've reverted line 756 back to 'provably'.

DougBurke:
@wmclaugh  This should have fixed the conflicts, let's see how CI likes it

PR #1105, opened 2021-02-12
Consolidate and harmonize the validation of dataset identifiers

DougBurke:
# Summary

Simplify the code used to validate dataset identifiers. Several names can no-longer be used as an identifier ('astrocompmodel', 'astrocompsource', 'astrodata', 'astromodel', 'astrosource', 'model_component', and 'source_component') and two can now be used ('energy' and 'photon').

# Details

It's long been a mystery to me over why some identifiers are "protected" so let's try and clear this up! This came out of PR #1089 but is a stand-alone change that I think is useful to have.

For users, this changes the set of valid/invalid keywords (the following is the logic used by `_fix_id`):

```
>>> from sherpa.astro.ui imoort *
>>> old = _session._plot_types.keys() | _session._contour_types.keys()
>>> new =  _session._plot_type_names.keys() | _session._contour_type_names.keys()
```

So the new symbols which can no longer be used as an identifier are:

```
>>> new.difference(old)
{'astrocompmodel',
 'astrocompsource',
 'astrodata',
 'astromodel',
 'astrosource',
 'model_component',
 'source_component'}
```

The following symbols used to be protected but are now allowed:

```
>>> old.difference(new)
{'energy', 'photon'}
```

## Why do we even validate the identifiers?

There is the question of why we even need to invalidate certain identifiers, since shouldn't everything essentially be namespaced so there's no possibility of confusion? I believe it's needed to support the `plot` and `contour` commands since these functions take

    type1 [, arg1_1, arg1_2, ...], type2 [, arg2_1, arg2_2,...], ... 

where `argn_1` is essentially always going to be a dataset identifier (technically it needed't be, but it often is). So, with a command like

    plot('data', 'fit')

is this to show two plots - e.g. `plot_data()` and `plot_fit()` - or is it the same as `plot_data('fit')`? We choose the former by making sure that we can't use a "plot type" as the identifier (in this case `fit`).

## Why do we need to change?

Unfortunately our code seems to actually validate against a slightly-different set of terms than the ones actually used in the `plot` and `contour` commands (which relies on the _plot_type_names dictionary)

 In CIAO 4.13 we can create the same plot three ways

```
sherpa In [6]: plot_data()

sherpa In [7]: plot('data')

sherpa In [8]: plot('astrodata')
```

(the reason for astrodata vs data is historical and I think was meaningful (but confusing) prior to #931 (or related conmits).

Note that we can not use `data` as an identifier (we get an `IdentifierErr`) but we can use `astrodata`, which seems wrong.

```
sherpa In [9]: load_pha('data', '3c273.pi')
read ARF file 3c273.arf
read RMF file 3c273.rmf
IdentifierErr: identifier 'data' is a reserved word

sherpa In [10]: load_pha('astrodata', '3c273.pi')
read ARF file 3c273.arf
read RMF file 3c273.rmf
```

## Open question

There is the question whether we should drop these `astroXXX` plot options. Note that in CIAO 4.12 we get confusing behavior:
`astrodata` is not supported but `astromodel` is, and returns a different plot to `model` in the `plot`call (#906 and #931 cleaned up some of this and, I believe, invalidated the need for the `astroXXX` options):

```
sherpa412 In [11]: plot('astrodata')
ArgumentErr: 'astrodata' is not a valid plot type

sherpa412 In [18]: plot('model')

sherpa412 In [19]: plot('astromodel')
/home/dburke/anaconda/envs/ciao412/lib/python3.7/site-packages/sherpa/plot/pylab_backend.py:330: MatplotlibDeprecationWarning: Passing the drawstyle with the linestyle as a single string is deprecated since Matplotlib 3.1 and support will be removed in 3.3; please pass the drawstyle separately using the drawstyle keyword argument to Line2D or set_drawstyle() method (or ds/set_ds()).
  zorder=zorder)

```

# Comparing behavior

For reference, the existing behavior is

### CIAO 4.12

ciao412-fit

### CIAO 4.13

ciao413-fit

### CIAO 4.12

ciao412-model

### CIAO 4.13

ciao413-model

## plot('model')

We can see that in 4.12 this produces a different plot to `plot_model`.

### CIAO 4.12

ciao412-plot-model

### CIAO 4.13

ciao413-plot-model

## plot('astromodel')

The astromodel plot in 4.12 matches the `plot_model` call.

### CIAO 4.12

ciao412-plot-astromodel

### CIAO 4.13

ciao413-plot-astromodel

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -85,7 +85,7 @@ def test_id_checks_session(session, setting):
 
 
 @pytest.mark.parametrize("session", [Session, AstroSession])
-@pytest.mark.parametrize("setting", ['cdf', 'lr', 'pdf', 'scatter', 'trace'])
+@pytest.mark.parametrize("setting", ['cdf', 'energy', 'lr', 'photon', 'pdf', 'scatter', 'trace'])
 def test_id_checks_session_unexpected(session, setting):
     """These identifiers are allowed. Should they be?"""
 
@@ -99,7 +99,7 @@ def test_id_checks_session_unexpected(session, setting):
                          [(Session, True), (AstroSession, False)])
 @pytest.mark.parametrize("setting", ['arf', 'bkg', 'bkgchisqr', 'bkgdelchi', 'bkgfit',
                                      'bkgmodel', 'bkgratio', 'bkgresid', 'bkgsource',
-                                     'energy', 'order', 'photon'])
+                                     'order'])
 def test_id_checks_astro_session(session, success, setting):
     """Do some common identifiers fail for astro but not default?"""
 
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2016, 2018, 2020  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016, 2018, 2020, 2021
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -19,15 +20,20 @@
 
 import numpy
 
+import pytest
+
 from sherpa.astro.data import DataPHA
-from sherpa.astro.ui.utils import Session
+from sherpa.astro.ui.utils import Session as AstroSession
+from sherpa.data import Data1D
 from sherpa.models import Const1D
+from sherpa.ui.utils import Session
+from sherpa.utils.err import IdentifierErr
 from sherpa.utils.testing import requires_data, requires_fits
 
 
 # bug #303
 def test_show_bkg_model():
-    session = Session()
+    session = AstroSession()
     session.load_arrays(1, [1, 2], [1, 2])
     session.show_bkg_model()
     session.show_bkg_model('xx')
@@ -39,7 +45,7 @@ def test_show_bkg_model():
 @requires_data
 @requires_fits
 def test_show_bkg_model_with_bkg(make_data_path):
-    session = Session()
+    session = AstroSession()
     session.load_data('foo', make_data_path('3c273.pi'))
     session.show_bkg_model()
     session.show_bkg_model('foo')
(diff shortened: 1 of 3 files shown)


DougBurke:
@hamogu I'd like a review of the code, to see if it makes sense, then we can discuss with @anetasie the future if the `astromodel` and related plot options.

anetasie:
@DougBurke this has not been well documented. I guess `astrodata` would use the resolution based on the response files, and  `data` would not? It is not obvious, but is this the case illustrated in `plot('model')` and `plot('astromodel')`?
I'm wondering about plotting models on the higher resolution grid defined by the user. How model defined on the higher resolution grid could be overplotted on the data? Would we need a different identifier?

hamogu:
Agreeing with @DougBurke.  I think we we should have fewer ui functions, with parameters if necessary. We already a large number of them, so many that I loose track of what they do and where to find them. So, no `plot_astromodel`. Also, my long-term hope is to use binning outside of astrodata, so the specific term "astro" is not good. Also, what does "astromodel" even mean, is that as opposed to "biologicalmodel"?

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -19,15 +20,20 @@
...
+from sherpa.data import Data1D
 from sherpa.models import Const1D
+from sherpa.ui.utils import Session
+from sherpa.utils.err import IdentifierErr
 from sherpa.utils.testing import requires_data, requires_fits
 
 
 # bug #303
 def test_show_bkg_model():
-    session = Session()
```
hamogu, line 30: Is there a specific reason this are not parameterized to test both session types? Is that just efficiency of not running unneeded tests? If so, I'm OK with that, I just want to make sure it's intentional.

hamogu (commented):
For simplicity of documentation (=understanding for users) and possibly also code, I suggest that we forbid the superset of identifiers for both session types. I think few users are aware that we do have different types of session objects in the first place. Also, there are so many possible identifiers (just think of all the available integer numbers!) that taking out a few is really not a problem. We could make this list even longer to reserve other identifiers for the future. It's almost impossible that anyone currently uses the session ui and has a dataset that is called "bkgdelchi".

sherpa/astro/ui/tests/test_astro_session.py
hamogu: I'll make this as resolved and then it's good for me.

hamogu (approved):

hamogu:
This is a step forward as-is. Up to @DougBurke and @anetasie if they want to reserve the same keywords for both session types or not and if the want to reserve the "astro*" keywords, even tough they should not have a special meaning for the plot comments IMO. Reserving more strings than currently used is not a problem, reserving fewer is.

anetasie (approved):

PR #1106, opened 2021-02-12
The sample_flux routine now returns the correct information for the clip column

### LINKED ISSUE #966: flux calculations when using set_full_model
`set_full_model` gives the flexibility in setting the model expression and one can calculate the flux for different model components. The updated `sample_energy_flux` handles this case. 
[bkg.example.tar.gz](https://github.com/sherpa/sherpa/files/5344013/bkg.example.tar.gz)

```
sherpa In [17]: sample_energy_flux(0.5,7.,model=pow1)
Out[17]: array([[2.18501706e-12, 5.85561312e-01, 3.31314089e+00, 7.43932636e-04]])
````

However,  `sample_flux` fails. We could improve the current error msg saying `You should use sample_energy_flux or sample_photon_flux instead`

```
sherpa In [20]: sample_flux(pow1,0.5,7.)
IdentifierErr: Convolved model
'(apply_rmf(apply_arf((50454.152030948 * (xsphabs.abs1 * powlaw1d.pow1)))) + (0.06576568179195134 * apply_rmf(apply_arf((50454.152030948 * (const1d.c1 * (((((((powlaw1d.bkg_pow1 + powlaw1d.bkg_pow2) + gauss1d.bkg_g1) + gauss1d.bkg_g2) + gauss1d.bkg_g3) + gauss1d.bkg_g4) + gauss1d.bkg_g5) + gauss1d.bkg_g6)))))))'
 is set for dataset 1. You should use get_model instead.
```

### PR DESCRIPTION
**Pull request by DougBurke**
# Summary

The sample_flux routine now returns correct information for the clip column (that is, it matches the clipping done by this routine). There may be changes to the reported error ranges because of this change.

The sample_flux routine will now point users to calc_energy_flux if set_full_model or set_bkg_full_model was used (as it uses the source model to calculate the statistics which it can't identify from a "full model" expressio). Fixes #966.

# Details

The `calc_sample_flux` function - a "low level" routine in `sherpa.astro.flux` - can be sent `modelcomponent=None` so it is obvious when the "unabsorbed" and "absorbed" fluxes are actually the same.

Ever since #866 landed the `sample_flux` routine returned a clip column that doesn't match what the routine does. We fix that here. A consequence of this is that we get a subtly different median/errors calculated when we hit a case where the parameter values have fallen outside the soft limits: previously the values were excluded but they are now included which tweaks the distribution slightly [**TODO** need to check this statement]

The fix for #966 is included here as I honestly don't have the energy to handle multiple PRs for functionality that I've been waiting over 6 months to get looked at. The fix for #966 is a bit ugly, as we have to ask for information we don't need and we don't have a nice API (I think) to identify when a "full model" is in use (e.g. the error classes raised for the source vs bkg case are different).

---

# Example

With the following script (note, it turns out to not show the main point of this PR but is instructive anyway to understand the changes that have been going on in `sample_flux` since CIAO 4.12):

```
import numpy as np
from sherpa.astro.ui import *

np.random.seed(1783)

load_pha('sherpa-test-data/sherpatest/3c273.pi')
notice(0.4, 6)

set_source(xsphabs.gal * powlaw1d.pl)
fit()

f1, f2, pars = sample_flux(pl, num=10)

print("")
print(f"Number masked: {pars[:, -2].sum()}")

print("# Iteration mask nH statistic")
for i, (mask, stat, nh) in enumerate(zip(pars[:, -2], pars[:, -1], pars[:, 1]), 1):
    print(f"{i:2d}  {int(mask):1d}  {nh:.4f}  {stat}")
```

then in CIAO 4.13 I get the following (and there are a number of issues with the output that are not related to this PR but have been fixed in `main` as I'm about to describe):

```
Dataset               = 1
Initial fit statistic = 5.43148e+10
Final fit statistic   = 21.78 at function evaluation 255
Data points           = 42
Degrees of freedom    = 39
Probability [Q-value] = 0.988288
Reduced statistic     = 0.558462
Change in statistic   = 5.43148e+10
   gal.nH         0.0142957    +/- 0.0940199
   pl.gamma       1.95164      +/- 0.171985
   pl.ampl        0.000184129  +/- 1.7539e-05
WARNING: hard minimum hit for parameter gal.nH
WARNING: Covariance failed for 'gal.nH', trying Confidence...
gal.nH lower bound:     -----
gal.nH upper bound:     0.0446962
original model flux = 9.36103e-13, + 2.09487e-13, - 2.74571e-13
model component flux = 1.35762e-12, + 8.94485e-14, - 2.63429e-13

Number masked: 3.0
# Iteration mask nH statistic
 1  0  0.0153  22.074275610444808
 2  1  0.0000  30.293462367621604
 3  0  0.0265  33.102269142733135
 4  1  0.0000  69.09519054432509
 5  0  0.0758  34.71601379862959
 6  0  0.0977  31.717372812591037
 7  1  0.0000  30.670132360021412
 8  0  0.0775  28.637858608992246
 9  0  0.0475  0.0
10  0  0.0213  0.0
11  0  0.0434  0.0
```

As an aside, if I use CIAO 4.12 then I get the same output for the flux rows and statistic column, but we don't provide the "mask" column as this was added in CIAO 4.13 (or Sherpa 4.12.something).

If I use the `main` branch then I get [here showing differences to the CIAO 4.13 output], which show

a) many of the repeated WARNING messages are no-longer generated
b) the median and one-sigma errors are different; this is because we now use numpy for calculating the percentiles rather than our code and they give different numbers [here we only have 11 flux values so the difference appears large]. However, this isn't the full story as shown below
c) we no-longer have any statistic bins with 0 in them

```
 WARNING: Covariance failed for 'gal.nH', trying Confidence...
 gal.nH lower bound:    -----
 gal.nH upper bound:    0.0446962
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
-original model flux = 9.36103e-13, + 2.09487e-13, - 2.74571e-13
-model component flux = 1.35762e-12, + 8.94485e-14, - 2.63429e-13
+original model flux = 1.11755e-12, + 2.15811e-13, - 2.19039e-13
+model component flux = 1.36636e-12, + 1.85307e-13, - 1.62047e-13

 Number masked: 3.0
 # Iteration mask nH statistic
  1  0  0.0153  22.074275610444808
- 2  1  0.0000  30.293462367621604
- 3  0  0.0265  33.102269142733135
- 4  1  0.0000  69.09519054432509
- 5  0  0.0758  34.71601379862959
- 6  0  0.0977  31.717372812591037
- 7  1  0.0000  30.670132360021412
- 8  0  0.0775  28.637858608992246
- 9  0  0.0475  0.0
-10  0  0.0213  0.0
-11  0  0.0434  0.0
+ 2  1  0.0000  31.773760670194623
+ 3  0  0.0265  30.293462367621604
+ 4  1  0.0000  37.42705248212813
+ 5  0  0.0758  33.102269142733135
+ 6  0  0.0977  69.09519054432509
+ 7  1  0.0000  26.077901060016888
+ 8  0  0.0775  34.71601379862959
+ 9  0  0.0475  31.717372812591037
+10  0  0.0213  30.670132360021412
+11  0  0.0434  28.637858608992246
```

If I compare this PR to the `main` branch then the difference is

```
 WARNING: Covariance failed for 'gal.nH', trying Confidence...
 gal.nH lower bound:    -----
 gal.nH upper bound:    0.0446962
-original model flux = 1.11755e-12, + 2.15811e-13, - 2.19039e-13
-model component flux = 1.36636e-12, + 1.85307e-13, - 1.62047e-13
+original model flux = 9.36103e-13, + 2.85049e-13, - 4.64372e-14
+model component flux = 1.35762e-12, + 1.62419e-13, - 9.16943e-14

 Number masked: 3.0
 # Iteration mask nH statistic
```

so the flux values have changed again, and are actually closer to the original values - but not identical, as this is the difference between the CIAO 4.13 and PR outputs ***just** for the flux lines

```
-original model flux = 9.36103e-13, + 2.09487e-13, - 2.74571e-13
-model component flux = 1.35762e-12, + 8.94485e-14, - 2.63429e-13
+original model flux = 9.36103e-13, + 2.85049e-13, - 4.64372e-14
+model component flux = 1.35762e-12, + 1.62419e-13, - 9.16943e-14
```

So the median value is now the same as it was in CIAO 4.13 but the sigma values have changed. If I change the number of iterations from 10 to 1000 (to get a better handle on the distribution) the difference in these limits is smaller (the - lines are CIAO 4.13 and the + lines are this PR), so I'm not too bothered about this difference (this difference because in 4.13 we dropped the rows which were outside the soft limits but we now include them in the distribution after moving the parameter value to the soft-limit boundary):

```
-original model flux = 1.03093e-12, + 1.62163e-13, - 1.58581e-13
-model component flux = 1.40432e-12, + 1.1435e-13, - 1.20281e-13
+original model flux = 1.03093e-12, + 1.62851e-13, - 1.5572e-13
+model component flux = 1.40432e-12, + 1.15212e-13, - 1.19412e-13
```

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13391,6 +13391,8 @@ class Session(sherpa.ui.utils.Session):
            soft limits, as this could cause an over-estimation of the
            flux when a parameter is only an upper limit. The statistic
            value is now returned for each row, even those that were
+           excluded from the flux calculation. The last-but-one column
+           of the returned `vals` array now records the rows that were
            excluded from the flux calculation.
 
         Parameters
@@ -13454,9 +13456,9 @@ class Session(sherpa.ui.utils.Session):
            free parameters and num is the `num` parameter. The rows of
            this array contain the flux value for the iteration (for
            the full source model), the parameter values, a flag
-           indicating whether any parameter in that row was clipped to
-           the hard-limits of the parameter, and the statistic value
-           for this set of parameters.
+           indicating whether any parameter in that row was clipped
+           (and so was excluded from the flux calculation), and the
+           statistic value for this set of parameters.
 
         See Also
         --------
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13370,7 +13370,6 @@ class Session(sherpa.ui.utils.Session):
                                              numcores=numcores,
                                              samples=scales, clip=clip)
 
-    # DOC-NOTE: are scales the variance or standard deviation?
     def sample_flux(self, modelcomponent=None, lo=None, hi=None, id=None,
                     num=1, scales=None, correlated=False,
                     numcores=None, bkg_id=None, Xrays=True, confidence=68):
@@ -13480,6 +13479,9 @@ class Session(sherpa.ui.utils.Session):
         logging to a level greater than "INFO" (e.g. with
         `sherpa.utils.logging.SherpaVerbosity`).
 
+        This routine can not be used if you have used set_full_model:
+        the calc_energy_flux routine should be used instead.
+
         Examples
         --------
 
@@ -13540,6 +13542,26 @@ class Session(sherpa.ui.utils.Session):
            not isinstance(modelcomponent, sherpa.models.model.Model):
             raise ArgumentTypeErr('badarg', 'modelcomponent', 'a model')
 
+        # We can not have a "full model" expression so error-out nicely here.
+        # Thanks to _get_fit we know we have a model so any error below can
+        # not be because no model is set but must be a "need full model"
+        # error. TODO: should we have a nicer way to determine this?
+        # Also note that it appears we have different ways the two code
+        # paths can error out
+        #
+        if bkg_id is not None:
+            try:
+                self.get_bkg_source(id, bkg_id)
+            except (IdentifierErr, ModelErr):
+                # At present ModelErr is thrown but keep in IdentifierErr
+                # just in case
+                raise IdentifierErr('Please use calc_energy_flux as set_bkg_full_model was used') from None
+        else:
+            try:
+                self.get_source(id)
+            except IdentifierErr:
+                raise IdentifierErr('Please use calc_energy_flux as set_full_model was used') from None
+
         correlated = sherpa.utils.bool_cast(correlated)
 
         # Why is this +1? The original comment was
@@ -13555,6 +13577,9 @@ class Session(sherpa.ui.utils.Session):
             samples = self.calc_energy_flux(lo=lo, hi=hi, id=id,
                                             bkg_id=bkg_id)
         else:
+            # NOTE: the samples are drawn from the full model expression
+            #       as this is how it was originally written
+            #
             samples = self.sample_energy_flux(lo=lo, hi=hi, id=id, num=niter,
                                               scales=scales, clip='soft',
                                               correlated=correlated,
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13556,7 +13556,7 @@ class Session(sherpa.ui.utils.Session):
                                             bkg_id=bkg_id)
         else:
             samples = self.sample_energy_flux(lo=lo, hi=hi, id=id, num=niter,
-                                              scales=scales, clip='hard',
+                                              scales=scales, clip='soft',
                                               correlated=correlated,
                                               numcores=numcores,
                                               bkg_id=bkg_id)
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -13528,22 +13528,17 @@ class Session(sherpa.ui.utils.Session):
         if not Xrays:
             raise NotImplementedError("sample_flux(Xrays=False) is currently unsupported")
 
-        ids, fit = self._get_fit(id)
-        data = self.get_data(id)
-        src = None
+        _, fit = self._get_fit(id)
         if bkg_id is not None:
             data = self.get_bkg(id, bkg_id)
-            src = self.get_bkg_source(id, bkg_id)
         else:
-            src = self.get_source(id)
+            data = self.get_data(id)
 
-        if modelcomponent is None:
-            modelcomponent = src
+        if (modelcomponent is not None) and \
+           not isinstance(modelcomponent, sherpa.models.model.Model):
+            raise ArgumentTypeErr('badarg', 'modelcomponent', 'a model')
 
         correlated = sherpa.utils.bool_cast(correlated)
-
-        if not isinstance(modelcomponent, sherpa.models.model.Model):
-            raise ArgumentTypeErr('badarg', 'modelcomponent', 'a model')
 
         # Why is this +1? The original comment was
         # "num+1 cause sample energy flux is under-reporting its result?"
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -546,7 +546,9 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
        id and session arguments have been removed and the
        modelcomponent argument can now be set to None. The statistic
        value is now returned for each row, even those that were
-       excluded from the flux calculation.
+       excluded from the flux calculation. The information on what
+       rows were excluded is now set, rather than reporting those rows
+       that exceeded the hard parameter limits.
 
     Parameters
     ----------
@@ -594,8 +596,8 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
         parameter. The rows of this array contain the flux value for
         the iteration (for the full source model), the parameter
         values, a flag indicating whether any parameter in that row
-        was clipped to the hard-limits of the parameter, and the
-        statistic value for this set of parameters.
+        was clipped (and so was excluded from the flux calculation),
+        and the statistic value for this set of parameters.
 
     See Also
     --------
@@ -603,10 +605,6 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
 
     Notes
     -----
-    This function is in the process of being documented and
-    re-written, as it contains a lot of redundancy and the semantics
-    of some of the options are unclear.
-
     The summary output displayed by this routine - giving the median
     and confidence ranges - is controlled by the standard Sherpa
     logging instance, and can be hidden by changing the logging to a
@@ -651,6 +649,12 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
                 iflx[nn] = calc_energy_flux(data, modelcomponent, lo=lo, hi=hi)
 
             mystat[nn, 0] = fit.calc_stat()
+
+            # Replace the input "clipped" value (from clip='hard') with
+            # the calculated clip here (essentially clip='soft'),
+            # remembering to invert the flag.
+            #
+            samples[nn, -1] = 0 if valid[nn] else 1
 
     finally:
         fit.model.thawedpars = thawedpars
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -570,7 +570,7 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
         The data object to use. This is not a DataSimulFit instance.
     samples
         The output of sample_flux for the data (assumed to have been
-        created with method=calc_energy_flux and clip='hard').
+        created with method=calc_energy_flux and clip='soft').
     modelcomponent : sherpa.models.Arithmetic instance or None
         The source model (without instrument response for PHA data)
         that is used for calculating the "unabsorbed" flux. This is
@@ -624,14 +624,8 @@ def calc_sample_flux(lo, hi, fit, data, samples, modelcomponent,
     npars = samples.shape[1] - 2
     assert nthawed == npars, (nthawed, npars)
 
-    # Identify any row where a parameter lies outside the min/max range
-    softmins = fit.model.thawedparmins
-    softmaxs = fit.model.thawedparmaxes
-
-    # We have to use columns 1 to n-1 of samples
-    valid = numpy.ones(samples.shape[0], dtype=bool)
-    for col, pmin, pmax in zip(samples.T[1:-1], softmins, softmaxs):
-        valid &= (col >= pmin) & (col <= pmax)
+    # Use the clip column from samples (after inverting the logic).
+    valid = samples[:, -1] == 0
 
     nrows = samples.shape[0]
     oflx = samples[:, 0]       # observed/absorbed flux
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by DougBurke** [comment]
For @anetasie - this PR, which we talked about yesterday, looked more-significant than it should have since it included commits from an earlier PR that had (after writing this PR) had been merged. I've rebased the code so it's easier to see what is new in this PR. I expect the tests to fail because #1121 has not been merged.

**Comment by DougBurke** [comment]
Now that #1121 has merged I've rebased so it should now pass.

**Comment by DougBurke** [comment]
And now a rebase as #1122 has been merged.

**Comment by anetasie** [comment]
ok I will review the clipping options as we discussed last week.

**Comment by DougBurke** [comment]
Rebased to allow the tests to run with updated OTS.

**Comment by DougBurke** [comment]
@anetasie - just a note that this needs reviewing

**Comment by anetasie** [comment]
@DougBurke is there a parameter/flag which turns off the clipping? For the too narrow soft limits this PR will bias the results. 

Here is my simple test:

```
from sherpa.astro.ui import *
load_data('acisf11309_000N023_r0002_pha3.fits.gz')
%matplotlib
notice(0.5,7.)
set_model('powlaw1d.p1')
set_stat('wstat')
fit()
p1.gamma.max=1.7
p1.gamma.min=1.5
f1=sample_flux(p1,0.5,7.,num=1000)
plot_pdf(f1[2][:,1])
p1.gamma.max=3
p1.gamma.min=1
f1=sample_flux(p1,0.5,7.,num=1000)
plot_pdf(f1[2][:,1],overplot=True)

```

The plot shows differences between the parameters returned when the soft limits are set to a narrow range and clipped, or a wider range.
sample

For a comparison the pdf returned in 4.13 - for narrow and broader limits.
sample_4 13

**Comment by anetasie** [comment]
`sample_energy_flux` has the clip column with the returned flag which is printed as float. Should this be integer for easier 'spotting'?

```
e2=sample_energy_flux(0.5,7.,num=1000,clip='soft')
    ...: 
    ...: 

In [72]: print(e2)
[[5.82170586e-13 1.67571741e+00 1.09018095e-04 0.00000000e+00]
 [5.76859390e-13 1.63927497e+00 1.04833704e-04 0.00000000e+00]
 [6.20255021e-13 1.60000000e+00 1.09047020e-04 1.00000000e+00]
 [5.24945534e-13 1.69915619e+00 1.00176202e-04 0.00000000e+00]
 [5.77512267e-13 1.62372919e+00 1.03595671e-04 0.00000000e+00]
 [5.93141437e-13 1.66100345e+00 1.09746184e-04 0.00000000e+00]
 [5.45902668e-13 1.66083380e+00 1.00991755e-04 0.00000000e+00]
 [5.68466307e-13 1.67299174e+00 1.06216155e-04 0.00000000e+00]
 [6.09129545e-13 1.61291893e+00 1.08274586e-04 0.00000000e+00]
 [5.63756034e-13 1.67435173e+00 1.05452654e-04 0.00000000e+00]
....

```

**Review by anetasie — COMMENTED** [review]
Prefer to use clip='hard' from `sample_energy_flux` which seems to be the default setting for `sample_energy_flux`

**Comment by DougBurke** [comment]
Rebased to pick up recent changes and added a fix for #966

File: sherpa/astro/ui/utils.py [hunk 1]
```diff
@@ -13558,8 +13577,11 @@ def sample_flux(self, modelcomponent=None, lo=None, hi=None, id=None,
             samples = self.calc_energy_flux(lo=lo, hi=hi, id=id,
                                             bkg_id=bkg_id)
         else:
+            # NOTE: the samples are drawn from the full model expression
+            #       as this is how it was originally written
+            #
             samples = self.sample_energy_flux(lo=lo, hi=hi, id=id, num=niter,
-                                              scales=scales, clip='hard',
+                                              scales=scales, clip='soft',
```
hamogu, line 13584: @anetasie left a general comment asking why you use soft here. My understanding is that soft limits set the values that are allowed for the fitting, e.g. I have prior knowledge that the temperature of the star is > 0.3 keV (let's say, I have a UV spectrum), so I set the soft limit or 0.3 keV. However, that should not stop the error bar from going below 0.3 keV, so `sample_energy_flux` should be allowed to go down to the hard limits (obviously, it can't go beyond the hard limits, because I can't evaluate the model there). Or maybe my interpretation of what "soft" limits in Sherpa are is wrong?

hamogu (changes requested):

DougBurke:
Just for reference, there have been a whole set of PRs leading up to this to the sample_flux/sample_energy|photon_flux commands to fix various issues including 

- #846 
- #751 
- #457 

(there are a bunch of other issues that were fixed, but these seem the most relevant to the discussion of soft vs hard limits).

The following is mostly for me:

I was going to say that I think that the recent change to the XSPEC parameter limits in #1261 may be relevant because I was assuming the nH hard limit may have been < 0 but that's not the case, but in CIAO 4.13 we can see the nH hard minimum is 0 so that's not the case

```
Sherpa 4.13.0

Python 3.8.2 (default, Mar 25 2020, 17:03:02)
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa

sherpa In [1]: xsphabs.gal
-------------> xsphabs.gal()
Out[1]: <XSphabs model instance 'xsphabs.gal'>

sherpa In [2]: print(gal)
xsphabs.gal
   gal.nH       thawed            1            0       100000 10^22 atoms / cm^2

sherpa In [3]: print(gal.nh)
val         = 1.0
min         = 0.0
max         = 100000.0
units       = 10^22 atoms / cm^2
frozen      = False
link        = None
default_val = 1.0
default_min = 0.0
default_max = 100000.0

sherpa In [4]: gal.nh.min
Out[4]: 0.0

sherpa In [5]: gal.nh.hard_min
Out[5]: 0.0
```

sherpa/astro/ui/utils.py
anetasie: Yes, this was my basic issue - how to use soft/hard limits in `sample_flux`. Is the intended answer use `sample_energy_flux`?

anetasie:
@DougBurke I still have the same plots as shown above and I don't see how to change from `clip=soft` to `clip=hard` using in this branch.  Is the answer to use `sample_energy_flux`?

anetasie (approved):
I approve this PR

DougBurke:
@anetasie - I did not add the ability to change the clip method in this routine as the idea was to make it behave "the same". Now, once I changed things to address #457 then it was never going to be exactly the same, but the idea was to remain similar.

So

a) we can add a clip parameter - which is #1297
b) did I really match the behavior, which I admit to being unable to answer at this time because I have not looked at this code in a long while

DougBurke:
I want to rebase this to pull in all the changes since February, so we don't have a repeat of #1292

anetasie:
@DougBurke adding a clip parameter would be good.

DougBurke:
@anetasie - I don't have time to add a clip parameter in this PR (it would require too much time to add tests for one).

anetasie:
@DougBurke I did not expect you to add this in this PR and this is why I approved this PR and submitted a separate issues.

DougBurke:
Unfortunately GitHub does not allow me to respond to individual comments dircetly - oh come back Google Wave ... - so to answer https://github.com/sherpa/sherpa/pull/1106#issuecomment-874877826 

The short answer is no.

The longer answer is that if we decided to we could use something like an AstroPy table, Pandas frame, or some-lvel of a NumPy "rec array" when returning a bunch of data like we do here, but

1. we'd want to make such a change in a coordinated manner (ie try and get all of them in one go so we don't have an annoying confusion)

2. I'm not convinced it wouldn't break some users (which is not to say we shouldn't do it but it needs work)

I've just created #1298 so we don't forget.

hamogu:
Replying to a comment high above: The flag should probably a boolean and not an int (and certainly not a float).

DougBurke:
I've just said we can't do that with the existing code, and then added an issue about it!

DougBurke:
**THIS IS A LONG DETAILED COMMENT BUT SOMEONE SHOULD READ IT CAREFULLY TO MAKE SURE I'M NOT JUST MAKING THINGS UP**

**ALSO NOTE: THE CODE NOW LOOKS VERY DIFFERENT AFTER 2 YEARS OF ME PUSHING CHANGES, BUT IT'S IMPORTANT HERE TO SEE WHAT THE CODE WAS DOING**

For the discussion of whether we should be using soft or hard limits here:

The original documentation routine doesn't describe what it does - e.g. https://sherpa.readthedocs.io/en/4.10.1/ui/api/sherpa.astro.ui.sample_flux.html . Now, I am pretty sure I wrote this text but there was no existing documentation describing how the data was clipped.

If I look at the 4.10.1 tag https://github.com/sherpa/sherpa/releases/tag/4.10.1 then we get to https://github.com/sherpa/sherpa/commit/a252b75f226362511f7d251e96e190b7ffc4067e and then to the code at https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/ui/utils.py#L11575

Looking at where the samples are defined we get https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/ui/utils.py#L11706-L11710

which takes us to  https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/ui/utils.py#L11569-L11572 

and so the hunt continues... to https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L65-L66

The final bit of spelunking takes us to https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L71-L79

So, in this version of the code we *set* any parameter value outside of the limits to the limit. How is this routine called? Well, it's called with the **hard** limits

https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L93-L95

Now, I would like to say that this is the end of the code spelunking, but it isn't. This is because after all the above, we then take the samples and - in `sample_flux` pass them to another routine at https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/ui/utils.py#L11712-L11714

This is https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L100-L101 which is just after the code we were looking at. This code goes through the samples array we just calculated and re-filters it with

https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L103-L111

which goes through and filters out all samples who are > softmin and < softmax. We shouldn't take those names seriously because it depends on what is sent in, but we can see from https://github.com/sherpa/sherpa/blob/a252b75f226362511f7d251e96e190b7ffc4067e/sherpa/astro/flux.py#L128-L130 that it really is using the soft limits

Note that thie fact it uses `> softmin and < softmax` is the reason for https://github.com/sherpa/sherpa/issues/457 - so this has long been changed to `>= sortmin and <= softmax`.

So, after all this, I stand by the current choice of the code to use the soft minimums, because this is how it was originally encoded.

We could change this, I honestly don't care anymore, but I do not believe the day of the code deadline is the time to make that change.

hamogu:
@DougBurke Sorry, I did not want to ask you to make changes now (I would have left inline changes) - I left the comment here so that I could reference it from the new issue you created (which I also responded to).

hamogu (approved):
I agree that this is an improvement and should go in as is. I also agree that this is what the original code did. (I don't think I ever disputed that.)

But - just because it's how the original code did it, does not mean it's the right thing to do. 
It does mean that this can be merged. It's an improvement and it's not breaking the old "soft limits" choice.

PR #1107, opened 2021-02-14
Add some grating related keywords for _repr_html_ for ARF, RMF, PHA

hamogu:
Summary
=======
Add keywords to the default output in the `_repr_html_` for some X-ray classes.

Details
=======
For grating data, the
diffraction order and the grating type (HETG / LETG) are very important
and should be printed. This only changes the output of those files that actually have the keywords in question
set. My new list adds a few keywords that e.g. CIAO does not set by
default, but that could be set in grating data.


hamogu:
@DougBurke Please let me know what you think, I'll update the tests if you agree to these changes. If you tell me that this is too much because you don't want to make the list of metadata printed that long, then I won't bother to change the tests.

DougBurke:
Just to note that the current tests of the notebook output are "not great". I'm not sure of the best thing to do for them, but at least they are noticing that something has changed in the output in this case...

DougBurke:
Out of interest, are these keywords set for XMM grating data, or is this Chagdra-grating specific?

hamogu:
From the ones I added, Chandra uses GRATING and TG_M, XMM uses INSTRUME (which is included anyway) "RFLORDER" (not ORDER), so I should probably change that. For XMM SOURCEID would also be good. I note that it does not hurt to list too much here, since keywords that are not present won't take up any space (and the chance that a keyword called TG_M or RFLORDER is present if it's not Chandra data is essentially zero)

DougBurke:
Just t make it more obvious, an example Chandra grating ARF

# Before

old

# After

oldold

hamogu:
I'll fix the tests as soon as I get to it so that we can continue with this.

DougBurke (approved):
LGTM

PR #1108, opened 2021-02-14
Read RMFs where N_CHAN is an array

hamogu:
Summary
=======
Allow the pyfits backend to read a wider range of RMF files

Details
======
The old code checked the number of groups, and if that is 1
assumed that N_CHAN would be a scalar. However, the OGIP standard
also allows N_CHAN to the a fixed or variable length array.
In that case, this test is insufficient, because the number of channel subsets would be one in a single row, while still
be stored as array - in that case you end up with `[array([1])]` which will break the for loop below. (This is how I discovered the problem). 
In particular, this will happen if some rows have 1 channel
subset, but others have more than one.

Why is there no test added?
-------------------------------
I'm working on generating rmf files for Arcus, which need channel subsets. I'm not entirely certain that all details of that response are correct, so I don't want to add it to the test data. Instead, take this PR as old-fashioned reading of the OGIP specs and thinking how to map this to Python.

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -735,7 +735,7 @@ def get_rmf_data(arg, make_copy=False):
         rowdata = []
         for mrow, ng, ncs in zip(matrix, n_grp, n_chan):
             # Need a RMF which ng>1 to test this with.
-            if ng == 1:
+            if np.isscalar(ncs):
                 ncs = [ncs]
 
             start = 0
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -735,7 +735,7 @@ def get_rmf_data(arg, make_copy=False):
         rowdata = []
         for mrow, ng, ncs in zip(matrix, n_grp, n_chan):
             # Need a RMF which ng>1 to test this with.
-            if np.isscalar(ncs):
+            if numpy.isscalar(ncs):
                 ncs = [ncs]
 
             start = 0


sherpa/astro/io/pyfits_backend.py
```diff
@@ -735,7 +735,7 @@ def get_rmf_data(arg, make_copy=False):
         rowdata = []
         for mrow, ng, ncs in zip(matrix, n_grp, n_chan):
             # Need a RMF which ng>1 to test this with.
-            if ng == 1:
+            if np.isscalar(ncs):
```
hamogu, line 738:
```suggestion
            if numpy.isscalar(ncs):
```

DougBurke:
Is the RMF read in correctly in CIAO (ie with the crates backend)? Maybe we could create a small synthetic RMF (possibly created on the fly for the test, rather than bothering adding it to the test-data) as a way to test this?

And while I'm asking (I haven't looked at the code yet...), how does this interact with #1091? I'm guessing it's a "this would need to be reworked as part of #1091" but wanted to check. Since I haven't looked at #1091 either...

DougBurke:
Things are coming back to me - one of the ideas in #1091 was to create an all-new astropy backend, so we'd stillneed this fix for the pyfits version...

DougBurke:
For reference, this code was last touched by me in 

- #358 Fix reading of non-variable-length-array FITS matrix https://github.com/sherpa/sherpa/commit/cfad8501f28928242fecc88560cfc753c3ab5a1a
- #412 Fix handling of ROSAT RMF file with AstroPy https://github.com/sherpa/sherpa/commit/649d733c971a9857289df00a9ddf1a326bf195ff

where the latter adds the comment

```
# Need a RMF which ng>1 to test this with.
```

Anti-prophetically this is another case where `ng=1`...

sherpa/astro/io/pyfits_backend.py
```diff
@@ -735,7 +735,7 @@ def get_rmf_data(arg, make_copy=False):
         rowdata = []
         for mrow, ng, ncs in zip(matrix, n_grp, n_chan):
             # Need a RMF which ng>1 to test this with.
-            if ng == 1:
+            if numpy.isscalar(ncs):
```
DougBurke, line 738: So in this case `n_chan = [array([1])]` which means `ncs = array([1])` which means you don't actually want this check to fire, which it would have done, since `ng = 1`.

I guess an alternative approach would be to just say "hey, let's not bother about the input form, let's just always reshape/flatten/whatever the values to the expected shape just before use". This could end up being more lenient, which could be helpful for the wild-west of PHA responses, but I'm not convinced. But I'm also not convinced the current code, which I've only lightly-worked on, is entirely sensible (hence the testing comment).

hamogu:
The comment `# Need a RMF which ng>1 to test this with` is correct - the current code will fail and the fix in this PR will solve that problem. I'm just saying that this problem can *also* occur for `ng=1` cases.

Anyway, to answer the points raised in review: Yes, having a test is usually better, but testing should not become a fetish either. I don't have a real-live HEASARC approved RMF as a test case here. I can read the OGIP specs and see that the current code is not complete. I can also read the OGIP specs and make up an RMF that triggers the problem. However, I think that's problematic for two reasons:

- A practical problem is making that on-the-fly mini-rmf for testing. We don't have RMF **writing** code in sherpa, so there is no easy way to write that test file in a backend-independent way. I would have to write two separate writing routines, just to make sure that the test runs with both backends.
- A different concern is the correctness of the test. A test is only as good as its input data. If the input data is wrong (here: Not fully OGIP compliant), then the test is not testing the right thing. Given the complexity of the ARF and RMF specs, it's probably more likely that my test rmf would be not compliant in some way then that I screw up this single-line code change.

So, I strongly suggest to accept this PR without an added test. Existing tests pass, meaning that I don't break anything. I'm also not removing tests. I'm simply improving one line of code that is only partially tested now (i.e. the line is tested, but with an RMF of the type I have here), and will have exactly the same coverage after this PR.

Relationship with #1091 : To be determined. I suggest "astropy" as a separate backend, so this test is still needed. Also, we can have this fix *right now* while the new backend would still have to be written (OK, I have it 50% ready, but who knows when I get to to the remaining 50%). Lastly, the "new astropy" backend is not fully "new". I'm copying some code from the pyfits backend and only replace those parts that can be simplified with the current astropy versions. I don't yet know if the line of this PR will be replaced or kept.

sherpa/astro/io/pyfits_backend.py
hamogu: If I understand @DougBurke correctly, this is essentially what I'm doing. Put `[]` around it if it's scalar and don't otherwise. I don't think that's "wile west", though, because the OGIP standard says that this could be a scalar, a fixed-length array, or a variable length array. `numpy.isscalar` will work for those cases; in other words, the OGIP standard tells us to accept scalars or arrays.

DougBurke: Yeah, my comment was more a work-through of the code, as I had not understood it originally.

Also, just because the OGIP standard says you can only do a, b, or c doesn't mean some mission doesn't do zeta :-)

DougBurke:
I wonder why we don't have `write_arf` and `write_rmf` in https://sherpa.readthedocs.io/en/latest/overview/astro_io.html#module-sherpa.astro.io - but that's not for this PR.

I really would like to know if we can read this file with the crates backend before accepting this PR, so we're not causing potential headache down the line.

hamogu:
Crates also fails to read this file:
```
/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/sherpa-4.13.0-py3.7-linux-x86_64.egg/sherpa/astro/io/__init__.py in read_rmf(arg)
    386 
    387     """
--> 388     data, filename = backend.get_rmf_data(arg)
    389 
    390     # It is unlikely that the backend will set this, but allow

/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/sherpa-4.13.0-py3.7-linux-x86_64.egg/sherpa/astro/io/crates_backend.py in get_rmf_data(arg, make_copy)
    840     # if rmfdataset.get_current_crate() < rmfdataset.get_ncrates():
    841     #    ebounds = rmfdataset.get_crate(rmfdataset.get_current_crate() + 1)
--> 842     ebounds = _get_crate_by_blockname(rmfdataset, 'EBOUNDS')
    843 
    844     if ebounds is None:

/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/sherpa-4.13.0-py3.7-linux-x86_64.egg/sherpa/astro/io/crates_backend.py in _get_crate_by_blockname(dataset, blkname)
    391     ncrates = dataset.get_ncrates()
    392     for ii in range(1, ncrates + 1):
--> 393         crate = dataset.get_crate(ii)
    394         if crate.name.strip().lower() == blkname.strip().lower():
    395             return crate

/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/pycrates/cratedataset.py in get_crate(self, crate)
    173         # get header and data info, if not already read in
    174         if tmpcrate._get_key_flag() == False:
--> 175             tmpcrate._read_header()
    176         if tmpcrate._get_data_flag() == False:
    177             tmpcrate._read_data()

/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/pycrates/tablecrate.py in _read_header(self)
    233             try:
    234                 backend.read_header(self)
--> 235                 backend.read_subspace(self)
    236 
    237                 if self._get_key_flag() and self._get_data_flag():

/nfs/mkx/a1/linux-x86_64/opt/packages/ciao-4.13/lib/python3.7/site-packages/pycrates/io/dm_backend.py in read_subspace(self, crate)
   1309 
   1310                 vdim = dm.dmGetElementDim(col)
-> 1311                 dtype = dm.dmGetDataType(col)
   1312 
   1313                 # loop over vector column components

SystemError: <built-in function dmGetDataType> returned NULL without setting an error
```

However, ISIS and XSPEC **do** read it correctly, suggesting that the problem is on the reading side and not with the file format.

DougBurke:
Ta for checking. You can see that the crates error is from the internals of crates. Yay. Can I have a copy of the file so I can send a crates issue?

DougBurke (approved):

hamogu:
I wanted to discuss a related issue at the next Sherpa meeting. I think this issue (and some others) would profit from having a small test file in the sherpa repro itself. Yes, I can make one and put it into the test-data, but that's quite some overhead in terms of workflow. Also, the test-data is just a flat collection and it's gets hard to find something. For this issue, I image making five rmfs that are all valid OGIP with almost identical content, but e.g. N_CHAN is number, array, or in header (because OGIP says columns cna be replaced by header keywords, if all entries are identical). Each file would be < 10 kb and I think it would help to have them live in this repro to (a) make the tests run every time and (b) reduce the overhead in  workflow (can test and merge without dealing with out-of-sync submodules, test data lives close to test so it's easy to see what it's good for), (c) [in general, not in this case] they are available all the time so they can be used in readthedocs to build example, and (d) so that everyone has them (they are part of the installation, so that we we even say "Can you try if this code works for you" when a user reports a problem).

We *do* have a a few datafiles in sherpa itself (sherpa/sherpa/astro/datastack/tests/data/) and I've used them for all sort of testing, debugging, etc, simply because they are always available. Let's talk about it, I would suggest that "test data < 25 kB can live in Sherpa itself". If we agree to that, I'll make a range of test files and add tests for all backends (some of which might be "xfail" for now), and then we can use the same file for debugging crates.

PR #1112, opened 2021-03-04
XSPEC: require model evaluation to be sent low and high grid values

DougBurke:
# Summary

XSPEC model classes must now be evaluated with bin edges - that is with low,high bins. The support for sending in a single grid and treating it a consecutive set of bins has been marked as deprecated from the model class interface. This feature is still supported for anyone evaluating the models directly from the sherpa.astro.xspec._xspec module or via the _calc method.

# Details

This was originally #947 but I have since split the PR into #1111 (has been merged) and this PR, with the integrate changes to come in a yet-to-be-determined PR.

XSPEC models, when evaluated by an instrument model, already use the two-argument form (ie send in low,high bin edges), so for the vast majority of use cases this does not change anything for users (that is, a user using `set_source`, `notice`, `fit`, `plot_fit` etc from the UI layer will not see any difference). The only tests that require changes are those that are designed to test the model evaluation (this is because previous PRs, such as #1111 and #946, have made the changes to ensure we use a response when we need to).

There is a possibility that downstream code will break because of this change, so rather than raising an error the code currently creates a `FutureWarning` message, which indicated to the user that things need to be changed. I chose `FutureWarning` rather than `DeprecationWarning` as it's harder to miss and, I believe but things seem to have changed around Python 3.7, is a closer match for the intended message (the semantics of the call has changed).

This actually makes a noticeable (but not spectacular) reduction in the time to run the Sherpa test suite (when you have XSPEC support enabled) since we no-longer have to run each XSPEC model with both one and two argument forms (although hard to show on CI as it is so dependent on the machine).

Normally I would argue that we shouldn't change something that works and could be being relied on, but

 - we have not explicitly documented the XSPEC behavior when sent a single grid, so it's not like we're "breaking an explicit contract"
 - combining XSPEC and non-XSPEC models - e.g. `XSphabs * PowLaw1D` - are currently tricky to understand what is going on when you evaluate the combined model and call it with a single argument
 - there are downstream changes related to model regridding that I believe are easier when we don't have the current behavior (this requires unraveling the `integrate` setting which is not well documented and was part of #947, but I've decided to separate the changes)

I do need to check the documentation (both docstrings and in docs/) to check they don't suggest this functionality is still possible.

# Example

With this PR we have the following behavior (note that the warning message is slightly confusing since it talks about the calc method, but unfortunately this is hard to catch sensibly; if we were throwing an error it would be easier to have a specific error message for the case used here):

```
>>> m1 = xspec.XSpowerlaw()
>>> m1([0.1, 0.2, 0.3], [0.2, 0.3, 0.4])
array([0.69314718, 0.40546511, 0.28768207])
>>> m1([0.1, 0.2, 0.3])
/home/dburke/sherpa/sherpa-master/sherpa/astro/xspec/__init__.py:868: FutureWarning: calc() requires pars,lo,hi arguments, sent 2 arguments
  warnings.warn(emsg, FutureWarning)
array([0.69314718, 0.40546511, 0.        ])
>>> m1([0.1, 0.2, 0.3])
array([0.69314718, 0.40546511, 0.        ])
```

The current behavior of this is

```
>>> m1 = xspec.XSpowerlaw()
>>> m1([0.1, 0.2, 0.3], [0.2, 0.3, 0.4])
array([0.69314718, 0.40546511, 0.28768207])
>>> m1([0.1, 0.2, 0.3])
array([0.69314718, 0.40546511, 0.        ])
```

An earlier version of the PR made it an error instead, as shown below:

```
>>> m1([0.1, 0.2, 0.3])
  File "/home/dburke/sherpa/sherpa-master/sherpa/models/model.py", line 304, in __call__
  File "/home/dburke/sherpa/sherpa-master/sherpa/models/model.py", line 95, in cache_model
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/xspec/__init__.py", line 867, in calc
    raise TypeError(emsg)
TypeError: calc() requires pars,lo,hi arguments, sent 2 arguments
```

# What about the integrate setting

Sherpa models have a `integrate` setting which indicates whether to evaluate the model actoss the grid or at a point (at least, that's my interpretation of this setting; it's not well documented in the code base what it's meant to do). Given this, perhaps XSPEC models

 - shouldn't allow the user to change the setting
 - have it fixed to `True` (although there is the fact that XSPEC multiplicative models technically act like `integrate=False`)

This isn't being done here (it was in #947), but I will add a PR for this (this has issues for how we handle the regrid support)

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2010, 2015-2018, 2019, 2020, 2021
-#         Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -848,6 +848,24 @@ class XSModel(RegriddableModel1D, metaclass=ModelMeta):
 
     @modelCacher1d
     def calc(self, *args, **kwargs):
+        """Calculate the model given the parameters and grid.
+
+        Notes
+        -----
+        XSPEC models must always be evaluated with low and high bin
+        edges. Although supported by the XSPEC model interface the
+        ability to evaluate using an XSPEC-style grid (n+1 values for
+        n bins which we pad with a 0), we do not allow this here since
+        it complicates the handling of the regrid method.
+
+        Keyword arguments are ignored.
+        """
+
+        nargs = len(args)
+        if nargs != 3:
+            emsg = f"calc() requires pars,lo,hi arguments, sent {nargs} arguments"
+            raise TypeError(emsg)
+
         # Ensure output is finite (Keith Arnaud mentioned that XSPEC
         # does this as a check). This is done at this level (Python)
         # rather than in the C++ interface since:
@@ -971,6 +989,13 @@ class XSTableModel(XSModel):
 
     @modelCacher1d
     def calc(self, p, *args, **kwargs):
+
+        # Note the kwargs is ignored
+        nargs = 1 + len(args)
+        if nargs != 3:
+            emsg = f"calc() requires pars,lo,hi arguments, sent {nargs} arguments"
+            raise TypeError(emsg)
+
         # The function used depends on XSPEC version and, prior
         # to XSPEC 12.10.1, the type of table.
         #
@@ -987,16 +1012,15 @@ class XSTableModel(XSModel):
 
         if hasattr(_xspec, 'tabint'):
             tabtype = 'add' if self.addmodel else 'mul'
-            return _xspec.tabint(p,
-                                 filename=self.filename, tabtype=tabtype,
-                                 *args, **kwargs)
+            return _xspec.tabint(p, *args,
+                                 filename=self.filename, tabtype=tabtype)
 
         if self.addmodel:
             func = _xspec.xsatbl
         else:
             func = _xspec.xsmtbl
 
-        return func(p, filename=self.filename, *args, **kwargs)
+        return func(p, *args, filename=self.filename)
 
 
 class XSAdditiveModel(XSModel):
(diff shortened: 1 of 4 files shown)


DougBurke:
@hamogu - as you are the most likely to be a user of this level of functionality I wanted to see what you thought about this.

hamogu:
I'm not a user of this level of functionality, but I could be. However, I support this change. I think sherpa has too many places were it's too flexible with input. Much of that comes from it's interactive heritage (pass an id number or not, pass bin edges or bin midpoints), but it frequently makes for hard to read and hard debug code, as a developer and also as a user (did I use this function wrong, or is it a bug in sherpa? ), so any place where we can clean that up, even if small, is good.

anetasie:
@DougBurke does this PR impact the way xspec user models are used?

DougBurke:
@anetasie I don't understand your question. What XSPEC user models are you talking about?

anetasie:
@DougBurke ok it should work. I was considering the models that the user compiles and which are not part of the distributed XSPEC model library.

DougBurke:
@anetasie - we don't officially support this use case (use external XSPEC models) in Sherpa at the moment so I don't see this as a stopper.

I have added code to CIAO that may let users use these models with Sherpa, depending on the phase of the moon [*] - but in that case it uses the same interface as the in-built XSPEC models so would behave the same. Anything we add to Sherpa to support this would also use the same interface.

[*] thanks to the way that XSPEC is built and installed then there are a lot of issues that can make compiling the code tricky, so we can't easily support this use case

anetasie:
Yes, that's what I was thinking.

anetasie (approved):

DougBurke:
I've created https://github.com/gammapy/gammapy/issues/3298 in case this cases problems for GammaPy.

dtnguyen2 (approved):

Issue #796, opened 2020-04-22
Refactor Session classes in `utils` and `astro.utils`


hamogu:
There is a lot of duplicated code (e.g. functions that are identical up to a line or so). In #793 I found that I needed to do the same change three times (and run separate tests for the two Session classes) because `get_column_data` is defined three times (in `io.py`, `pftils_backend.py` and `crates_backend.py`) with the former two functions being identical and the latter two differeing by two lines. Similarly, the `set_filter` in both sessions are virtually identical. There are probably more, but I only compared the functions above.

hamogu:
I don't know what I meant when I opened this issue. The session classes are pretty good about using `super`. Most of the lines that are duplicated are docstrings where `astro.ui` adds a bkg argument, so the docstring is almost, but not quite a copy of the `ui`. Unless we do a lot of magic in editing docstrings there is no easy way around it and "a lot of magic" does not seem worth the effort at this point. 
(If re-writing from scratch, maybe the general session class could also deal with backgrounds and then the difference would be far smaller, but that's also not worth the effort at this point.)

PR #1113, opened 2021-03-05
XSPEC: support etable table models

DougBurke:
# Summary

Allow XSPEC ETABLE table model files to be read by load_xstable_model by setting the etable parameter. For example if model.fits is an ETABLE XSPEC model then it would be loaded with

    load_xstable_model('emdl', 'model.fits', etable=True)

If the `etable` argument is not given then the model will be treated as a MTABLE model. 

# Details

Unfortunately there's no way to tell from a table-model file whether it's multipicative (mtable) or exponential (etable), so out-of-band information is required (the user has to explicitly say what type is being used). The default assumes mtable.

Note that there is no support added to load_table_model, as the support for XSPEC table models there has been deprecated for a long time.

Note there's no test of this functionality, and I'm not 100% happy with the interface.If etable models could be determined automatically then it would be okay, but maybe we should have something like

  - read_xstable_atable
  - read_xstable_mtable
  - read_xstable_etable

Then again, as the number of "etable" models is low, perhaps it's okay with

  - read_xstable_model(modname, filename, etable=True)

for this case.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -9658,12 +9658,17 @@ class Session(sherpa.ui.utils.Session):
 
         return (x, y)
 
-    def load_xstable_model(self, modelname, filename):
+    def load_xstable_model(self, modelname, filename, etable=False):
         """Load a XSPEC table model.
 
-        Create an additive (``atable``, [1]_) or multiplicative
-        (``mtable``, [2]_) XSPEC table model component. These models
-        may have multiple model parameters.
+        Create an additive (``atable``, [1]_), multiplicative
+        (``mtable``, [2]_), or exponential (``etable``, [3]_) XSPEC
+        table model component. These models may have multiple model
+        parameters.
+
+        .. versionchanged:: 4.13.1
+           The etable argument has been added to allow exponential table
+           models to be used.
 
         Parameters
         ----------
@@ -9671,7 +9676,10 @@ class Session(sherpa.ui.utils.Session):
            The identifier for this model component.
         filename : str
            The name of the FITS file containing the data, which should
-           match the XSPEC table model definition [3]_.
+           match the XSPEC table model definition [4]_.
+        etable : bool, optional
+           Set if this is an etable (as there's no way to determine this
+           from the file itself). Defaults to False.
 
         Raises
         ------
@@ -9690,7 +9698,7 @@ class Session(sherpa.ui.utils.Session):
         Notes
         -----
         NASA's HEASARC site contains a link to community-provided
-        XSPEC table models [4]_.
+        XSPEC table models [5]_.
 
         References
         ----------
@@ -9699,9 +9707,11 @@ class Session(sherpa.ui.utils.Session):
 
         .. [2] http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSmodelMtable.html
 
-        .. [3] http://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/general/ogip_92_009/ogip_92_009.html
+        .. [3] http://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSmodelEtable.html
 
-        .. [4] https://heasarc.gsfc.nasa.gov/xanadu/xspec/newmodels.html
+        .. [4] http://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/general/ogip_92_009/ogip_92_009.html
+
+        .. [5] https://heasarc.gsfc.nasa.gov/xanadu/xspec/newmodels.html
 
         Examples
         --------
@@ -9714,6 +9724,10 @@ class Session(sherpa.ui.utils.Session):
         >>> set_source(xsphabs.gal * xtbl)
         >>> print(xtbl)
 
+        Load in an XSPEC etable model
+
+        >>> load_xstable_model('etbl', 'etable.mod', etable=True)
+
         """
 
         try:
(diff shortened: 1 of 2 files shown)


anetasie:
@DougBurke your choice is good. It would be good to have the test case.
 
This is from the XSPEC manual specifying how the `emodel` should work. Is any of this going to happen in Sherpa as well?
I don't really like this hidden assumed ways this model operates.

_XSPEC will multiply the contents of the model by -1 then take the exponential i.e. this model is for calculating absorption functions. .....
The model value is assumed unity outside the tabulated energies unless the keywords LOELIMIT and HIELIMIT are given in the primary extension of the FITS file. In the latter case the value given in LOELIMIT will be used for energies below those tabulated and HIELIMIT for energies above those tabulated._

DougBurke:
All we do is ask XSPEC to interpret this as an "emodel", so we get all this behavior "for free", and we can't change it [this is assuming you are objecting to the behavior outside the elimit values, which are up to the model writer to set]. In this particular case I don't see the problem you see here, as it is at least explained in the documentation.

However, I don't know which decision you say is good - I gave two options

 - `load_xstable_model` where the user has to know to add "etable=True" to the call
 - bite the bullet and have separate functions, such as `load_xstable_amodel`, `load_xstable_emodel`, `load_xstable_mmodel` [or some other name] and require users to use the correct one (which more-closely mirrors XSPEC)

anetasie (approved):
I don't have a test file to check. I approve the PR as it extends the existing functionality to match the xspec models following the xspec approach.

DougBurke:
Rebased to pick up latest changes and there was actually a conflict even though the UI didn't seem to think there was.

DougBurke:
@hamogu - have you any thoughts on my existential angst over the `load_xstable_model`: do we

- add an optional parameter to indicate that we want the etable support (when the table is multiplicative)
- change our approach to be more like XSPEC and require the user to explicitly specify the type of model they are loading - perhaps `load_xsatable_model`,`load_xsmtable_model`, and `load_xsetable_model` which is close-ish to the XSPEC naming scheme. (if we went with this we could allow load_xstbale_model to exist for another release but tell users they need to change their code).

This PR only implements the first option.

This is only needed because there's no information in an ETABLE model file to indicate it should be treated as ETABLE rather than MTABLE, whereas we can tell the difference between MTABLE and ATABLE. My assumption is that the number of ETABLE files is significantly less than MTABLE/ATABLE, so making users of these by explicit will not affect most people, but I do worry about the conceptual information flow whereas if we make all XSPEC table model users specify exactly what they want then it is clearer to all, at the expense of usability.

hamogu:
I think either is fine. I have, in fact, never seen an ETABLE model. I have a slight preference for not adding any more new functions, i.e. I prefer the approach with an extra keyword implemented here.

A thought I had, but then quickly discarded, was that, in principle, we can add a heuristic. MTABLE models as well as ETABLE models are typically used for absorption and that use case is explicitly called out in the OGIP memo. For absorption, the values in the MATABLE model will typically be between 0 and 1, while in an ETABLE model they will be between 0 and INF. So, we could load in the table, check which fraction of the entries is > 1 and issue a warning if it's a larger number (say, > 10 %). The reasons I'm discarding this option is because (a) etable models are so rare (b) it's a lot of work to check that that check is working.

hamogu (approved):
I approve the concept. I think we need to test it with at least one etable - not necessarily in CI, but just locally checking out this branch would be fine. To be clear: I have not done that test, my approval is for the Python this of things and the docstrings.

DougBurke:
I don't believe there's an etable model in https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/newmodels.html. Actually, I think (based on @nplee work on the original CIAO helpdesk ticket) find one at

http://astrophysicsformulas.net/mytorus/mytextfiles/zeroc/mytorus_Ezero_v00.fits.gz

note that it's not small.

I'm not sure what file I used when testing. I was able (I believe) to get the same plot out in XSPEC and Sherpa when using it, but I can't guarantee that.

DougBurke:
@hamogu - a semi-serious, semi-in-fun question, but in https://github.com/sherpa/sherpa/pull/1022#issuecomment-828432550 you really didn't like adding a parameter, but in this one you are okay with it. I believe the answer is likely "it depends on the circumstance" or "don't add mode options to commonly-used routines", but just wanted to check.

hamogu:
Oh, but here I argue for an extra parameter as the lesser evil to adding another **method or function**. 

I do think that Sherpa suffers from UI/API complexity. Some if that is intrinsic to the problem (there are so many different models for X-ray emission), but most of it is home-made. For various reasons, there are often Sherpa models that almost exactly duplicate XSPEC models (powerlaw, constant, ...). Then, there is the UI layer vs. OO layer duality which essentially duplicates every function, and there is quite some code duplication even where it does not need to be e.g. #1174, #796.

I think that's in fact a problem for several reasons: 
- I don't find the function I need because reading the list of all functions is just too long.
- There is a certain mental overhead of using all those functions I really need only once per project (sample_energy_flux is one of those I rarely use). Every package has some of those, it's hard if you never get in the flow of re=using what you can type from memory.
- More stuff means more maintenance - which currently is work for you and me.
- I don't think it's an accident that we have limited community buy-in (gammapy decided sherpa fitting did not match their requirement, stingray looked at what we did, but then rolled out their own). Even now I'm uncomfortable making certain PRs for sherpa and I'm pretty deep in the weeds of reviewing the ``__contain__`` function...

So, when I read PRs, I try to mentally check if we really, really need this feature or if we can do without it or find some simpler way.

Issue #889, opened 2020-08-14
RFE plot_bkg family of routines to be able to scale the data to match the aperture (optional)


DougBurke:
It makes sense for `plot_bkg`, and related routines, to behave the way they do: when you plot up the background you see the actual data (so divided by the background exposure time - with the default analysis setting - and no correction for BACKSCAL).

There are times however when I'd like to be able to overpot the background and source apertures and have them scaled, so you can compare the values - that is I'd like to be able to say something like (options and naming open for bike-shedding):

    plot_data()
    plot_bkg(scale=True, overplot=True)

The scaling you apply depends on whether the Y axis is in units of rate or counts.

What we probably need to do (if it's not already done) is to send the source dataset along with the background dataset to the prepare method of the background object, and add a preference setting (e.g. 'scale' above ,but I wonder if that's too generic a name) to control whether the scaling happens. Fortunately we already have background-specific classes - e.g. `sherpa.astro.plot.BkgDataPlot` - which would need to override the prepare method. The plot options go through a single _plot method of the sherpa.astro.ui.Session class, which would need to be taught about the extra options [or, possibly preferably, is to push the extra logic out to the plot_xxx functions, but that's not relevant here :-]

For reference, this is the current behavior of Sherpa:

```
import numpy as np

from sherpa.astro.ui import *

# create a source spectrum
#
load_arrays(1, [1, 2, 3], [100, 200, 150], DataPHA)
set_exposure(10)
set_backscal(0.1)

# create a background spectrumm in dataset 2 and then
# copy it over
#
load_arrays(2, [1, 2, 3], [8, 10, 12], DataPHA)
set_exposure(2, 50)
set_backscal(2, 0.3)
b = get_data(2)
set_bkg(b)

# The plots for data and background show
# counts / exposure time (as the default is a rate)
# and no correction for BACKSCAL.
#
p1 = get_data_plot().y == np.asarray([100, 200, 150]) / 10
if not p1.all():
    print("Data is unexpected")

p2 = get_bkg_plot().y == np.asarray([8, 10, 12]) / 50
if not p2.all():
    print("Background is unexpected")
```

DougBurke:
I've done some work, and there's a problem with my suggested approach: the behavior which I am suggested be set by a preference (and so can be passed in to plot_bkg as a named argument) is really intended for the prepare phase, but the preference settings are currently assumed to be sent to the low-level plotting code. There'd have to be some way to remove the scale argument (or whatever we call it) at a late stage.

[...]

PR #1114, opened 2021-03-11
Tests: add basic WSTAT plot checks

DougBurke:
What data is shown when we plot the data and model with stat=WSTAT?

We test both low-level and the UI layer just to check the same
information is being used.

At the moment the WSTAT statistic does not "correct for" the model
signal in the data. The intention is that we will be changing this
in the next commit.


DougBurke:
@anetasie @hamogu  (SDS) can test out this approach without even trying this patch - all you need to do is fit a PHA file with wstat and then compare

```
plot_fit()
```

to

```
subtract()
plot_fit()
unsubtract()
```

The latter is essentially what this PR does.

hamogu:
Thanks for tagging me in this - I did not even know that the wstst statistic exists at all, so I've never wondered how it should be visualized. Turns out, I have a dataset open in a notebook right now where it might be useful.

hamogu:
Unfortunately, I'm not able to use this statistic:
```
StatErr                                   Traceback (most recent call last)
<ipython-input-70-d46b8b286e15> in <module>
      1 ui.set_stat("wstat")
      2 ui.set_model(ui.xsphabs.abs1 * ui.xsapec.ap1)
----> 3 ui.fit()

<string> in fit(id, *otherids, **kwargs)

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/ui/utils.py in fit(self, id, *otherids, **kwargs)
  10032         """
  10033         kwargs['bkg_only'] = False
> 10034         self._fit(id, *otherids, **kwargs)
  10035 
  10036     def fit_bkg(self, id=None, *otherids, **kwargs):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/ui/utils.py in _fit(self, id, *otherids, **kwargs)
  10138                     i).mask & numpy.isfinite(self.get_data(i).get_x())
  10139 
> 10140         res = f.fit(**kwargs)
  10141         res.datasets = ids
  10142         self._fit_results = res

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/fit.py in run(fit, *args, **kwargs)
     54         else:
     55             fit.model.startup(True)
---> 56         result = func(fit, *args, **kwargs)
     57         fit.model.teardown()
     58         return result

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/fit.py in fit(self, outfile, clobber, numcores)
   1240             raise FitErr('binhas0')
   1241 
-> 1242         init_stat = self.calc_stat()
   1243         # output = self.method.fit ...
   1244         tmp = self._iterfit._get_callback(outfile, clobber)

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/fit.py in calc_stat(self)
   1123         """
   1124 
-> 1125         return self._calc_stat()[0]
   1126 
   1127     def calc_chisqr(self):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/fit.py in _calc_stat(self)
   1104         # TODO: is there anything missing here that
   1105         #       self._iterfit.get_extra_args calculates?
-> 1106         return self.stat.calc_stat(self.data, self.model)
   1107 
   1108     def calc_stat(self):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/stats/__init__.py in calc_stat(self, data, model)
    971             nbkg = len(bids)
    972             if nbkg == 0:
--> 973                 raise StatErr('usecstat')
    974 
    975             elif nbkg > 1:

StatErr: No background data has been supplied. Use cstat
```

That's despite a background dataset that's clearly present:
```
ui.get_bkg()
```
image

DougBurke:
@hamogu - dotty, about to take the afternoon off so response will be limited.

Background: XSPEC has the cstat statistic which we have broken up in our infinite wisdom into cstat and wstat (without background and with background).

What happens if you `print(ui.get_data())` - is the `background_ids` array empty or not (as you appear to have it both set and unset)...

hamogu:
```
name           = TYC_2597_735_1.pi
channel        = Float64[1024]
counts         = Float64[1024]
staterror      = None
syserror       = None
bin_lo         = None
bin_hi         = None
grouping       = None
quality        = None
exposure       = 8599.2928494005
backscal       = 0.00011139411906376
areascal       = 1.0
grouped        = False
subtracted     = False
units          = energy
rate           = True
plot_fac       = 0
response_ids   = [1]
background_ids = [1]
```

hamogu:
I agree that this PR is an improvement in the plotting. I'll have to look at the code more closely to see if I have any comments related to the implementation.

sherpa/astro/plot.py
```diff
@@ -112,8 +124,22 @@ def prepare(self, data, stat=None):
         # Maybe to_plot should return the lo/hi edges as a pair
         # here.
         #
+        if isinstance(stat, WStat) and data.background_ids != [] and not data.subtracted:
+            # When using the WStat statistic we want to show the
+            # subtracted data. How are the error-bars handled in
+            # this case?
```
hamogu, line 130: That's a questions for @anetasie! However, personally, I find it looks convincing in the plots.

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -348,3 +352,123 @@ def test_pha_model_with_gaps_977():
...
+
+
+def test_pha_plot_with_wstat_data(caplog):
+    """Can we create a PHA data plot with WSTAT.
+
+    For PHA data, WSTAT "silently" includes the background, which
+    complicates visualization: do we subtract the background from
+    the data, add the background to the model, or something else?
+
+    At the moment we do nothing.
```
hamogu, line 364: Is that true? I thought this PR is there specifically to do something?

hamogu (approved):
I sometimes feel that we are overtesting. Is a separate test for the ui layer really required?

sherpa/astro/tests/test_astro_plot.py
DougBurke: This comment was added in 19cd8fa when we did in fact "do nothing". It hasn't been updated to address the change in behavior of the code.

sherpa/astro/plot.py
DougBurke: We add in fake error bars - using Chi2XspecVar - but I don't know if it just uses the "source" counts or does something fancy to handle combined data.

A quick check suggests that subtracting the background does not change the Y errors (in this case not using WSTAT but I would expect this to be hand;ed generically), which I guess answers this question (we can see that y2 <= y1 but dy2 == dy1):

```
read ARF file 3c273.arf
read RMF file 3c273.rmf

sherpa In [2]: d = get_data()

sherpa In [3]: y1 = d.to_plot()[1]

sherpa In [4]: dy1 = d.to_plot()[3]

sherpa In [5]: subtract()

sherpa In [6]: y2 = d.to_plot()[1]

sherpa In [7]: dy2 = d.to_plot()[3]

sherpa In [8]: y2 < y1
Out[8]:
array([ True,  True, False,  True,  True, False, False, False,  True,
        True, False,  True,  True,  True,  True, False, False,  True,
        True, False,  True,  True,  True,  True, False,  True, False,
        True, False, False,  True, False,  True,  True,  True,  True,
        True,  True,  True,  True,  True,  True,  True,  True,  True,
        True])

sherpa In [9]: (dy1 == dy2).all()
Out[9]: True
```

DougBurke:
I disagree that there's no benefit to the UI tests - it checks that a lot of the low-level plumbing is being handled for us (e.g. adding a response model, selection of the correct plot object) that we hard-code (i.e. do manually) in the lower-level test. We could argue that we don't need the lower-level test instead - since it's implicitly tested by the UI level test - which is more a judgement call. One thing I like about the lower-level tests is it makes it explicit how things can be done, but it is something to think about if we believe our testing budget (e.g. time to run) is excessive.

sherpa/astro/tests/test_astro_plot.py
DougBurke: And fixed.

hamogu:
Oh, I wasn't worried about the time to run tests, more about the time to write them! Sorry, I did not want to say they are "useless", more that we many not need to test both the ui and the oo layer in every case, but I think it would be OK to do it only in some cases.

DougBurke:
For future reference, the problems @hamogu was happening in https://github.com/sherpa/sherpa/pull/1114#issuecomment-802932170 are not easily repeatable / out of the scope of this PR.

anetasie:
@dburke the plots are good. Do you recalculate the errorbars for plotting to include the addition of the background error?

sherpa/astro/plot.py
DougBurke: Note that I'm not sure this example is actually telling me what I think it's telling me...

DougBurke:
[the following should be taken to be a continuance of my discussion above at https://github.com/sherpa/sherpa/pull/1114#discussion_r598772633 ]

@anetasie - this does not change the yerr values **I think** but I don't have a build to check. If you do have a build available thenthe following should tell us

  set_stat('chi2xspecvar')
  unsubtract()
  yerr1 = get_data_plot().yerr
  set_stat('wstat')
  yerr2 = get_data_plot().yerr
  print(yerr2 > yerr1)

I had mentioned above that I thought we weren't correctly handling the y error bars in a plot but for a dataset with chi-square errors we do actually see the y errors increase when you call subtract:

```
sherpa In [30]: unsubtract()

sherpa In [31]: yerr1 = get_data_plot().yerr

sherpa In [32]: subtract()

sherpa In [33]: yerr2 = get_data_plot().yerr

sherpa In [34]: yerr2 > yerr1
Out[34]:
array([ True,  True,  True,  True,  True,  True,  True,  True,  True,
        True,  True,  True,  True,  True,  True,  True,  True,  True,
        True,  True,  True,  True,  True,  True,  True,  True,  True,
        True,  True,  True,  True,  True,  True,  True,  True,  True,
        True,  True,  True,  True,  True,  True,  True,  True,  True,
        True])

sherpa In [35]: (yerr2 > yerr1).all()
Out[35]: True
```

Unfortunately, due to the design of Sherpa I don't think it's going to be "nice" to get the error bar values to change (that is, I fear it will require some non-local changes to support the WSTAT case rather than being able to do this "nicely" within the wstat statistic class).

anetasie:
`chi2datavar` takes into account the background when it is plotting background subtracted data. For `wstat` - when subtracting the data for plotting the errors should also be adjusted. Perhaps getting the errors array for plotting is not so unreasonable, as it should have been already calculated.

DougBurke:
@anetasie - I can promise you it isn't as easy as you might think. We don't have a way to say "we need to treat this statistic as subtracted but it isn't actually subtracted". Previously we've "solved" this by a check for "oh, if the statistic is wstat then do this special case", but that approach is problematic. Maybe - designing on the fly so this has not been thought through - we need to separate the concerns of "background is included somehow" from "we handle the background by subtracting it". I shall pontificate on this for a bit.

anetasie:
It is confusing when for `wstat plot_data()` displays the background subtracted data, while the data is not really background subtracted, i.e. no `subtract()` was issued.  It would be perhaps less confusing if the subtraction is only done in `plot_fit()`, so in the case when the data is overlayed with the model, when one needs to check the fit 'by eye'.
In this PR `plot_data()`  behaviour is different for this statistic only, which is not ideal.

DougBurke:
I've talked with @anetasie about this, but I personally dislike how the way

    plot_fit()

and

    plot_data()
    plot_model(overplot=True)

do not create the same plot (ignoring color differences) for PHA data. Aneta's suggestion about handling `plot_fit` and `plot_data` here feel like the same thing to me.

hamogu:
I agree with @DougBurke. I see @anetasie reasoning, but `plot_fit` should always be the same as `plot_data` + `plot_model`. It's not just the plot itself. I also do `get_plot` and then plot the x/y values I get in other ways, e.g. overlay on something else, plot different models, use matplotlib for plot options that `plot_fit` does not support etc. If `plot_fit` gave me something that I could not make from combining the dataplot and the model plot, that would be very confusing.

anetasie:
I still think it is confusing to display the background subtracted data when the background is not subtracted from the data.

```
sherpa In [8]: print(get_data().subtracted)
False
```

DougBurke:
But we have to have *some* way to indicate what the background contribution for statistics like WSTAT that auto-magically handle the background. We have

 - plotting the data
 - plotting the model
 
At the moment the model evaluation is the "worst affected", but my experiments suggest it's the hardest to include the background in this case.

So we can

 - do nothing, and have to warn users that plotting with wstat is problematic
 - do something

I am all ears on what "do something" should be.

anetasie:
@DougBurke one option would be to add scaled background plot. Current `plot_bkg()` displays the background data without the scaling which does not reflect the background contribution to the data.

DougBurke:
@anetasie - I have a bunch of experiments with scaling the background with #889 being the issue and #694 was my first attempt (which came out of my original work trying to fix the WSTAT plotting). For WSTAT users the problem with this approach is that in a command like `plot_fit_resid` there's no indication to them that the background isn't included in the fit part of the plot; they would have to remember to add in a `plot_back` call, and this is just not a nice user experience (it is too easy to forget).

We *could* think about always adding in the scaled background for WSTAT users, but this would be a **very** large change and come with a lot of other questions (e.g. when to add this data).

anetasie (approved):
I approve this PR and will open the issue for plotting the correct errorbars for background subtracted plot.

anetasie:
@DougBurke I was just running the code in this branch and noticed that the 'subtract/unsubtract' does not update the data in `plot_data` for the WSTAT. I was assuming that the only change was for `plot_fit` only - when the data is being overplotted by the model. However, it looks like both `plot_data` and `plot_fit` show the same data, which I assume is background subtracted. I don't think this was the intended change in this PR. This is really confusing as `get_data()` indicates the subtracted/unsubtracted data, but `plot_data()` does not.

DougBurke:
@anetasie I am not working on this bit of code. I don't have time to think about this at this time. If people really wanted it then we could have spent some time between March and a month ago to work on it.

anetasie:
@DougBurke Agree. We don't want this PR for the release.

hamogu (changes requested):
Since there is agreement not to take it in this form, I'll switch my review to 
"request changes". I had approved it a long time ago and other code has changed underneath it since.

hamogu:
This is a case where the statistic is essentially just wrong and we strongly want to advise users not to use it. I believe that is not our place to patronize our users. If they really want to use WSTAT, then can. However, it is fair for us to say that we don't want to invest further resources to improve something that we know to be flawed.

So, I suggest to not spend any more time on this PR (we don't know what to do anyway), instead we add a warning to be issued:

"WSTAT has been shown to give biased results (URL here to CIAO thread or published paper). It is left in for backwards compatibility, but fitting and plotting for this statistic are not developed any further."

DougBurke:
TODO: rebase this and review.

Issue #1109, opened 2021-02-26
psf.pars not returning list of parameters


kglotfelty:
I know that the PSF stuff is a little bit different than other model components (eg `gauss2d`) but shouldn't the `.pars` return the list of parameters?

```python
sherpa In [1]: load_data("fi/out_broad_thresh.img")

sherpa In [2]: load_psf("mypsf", "tmpjrv7xb58_simulate_p.psf")

sherpa In [3]: set_psf(mypsf)

sherpa In [4]: print(mypsf)
psfmodel.mypsf
   Param        Type          Value          Min          Max      Units
   -----        ----          -----          ---          ---      -----
   mypsf.kernel frozen tmpjrv7xb58_simulate_p.psf
   mypsf.size   frozen   (256, 256)   (256, 256)   (256, 256)
   mypsf.center frozen   (128, 128)   (128, 128)   (128, 128)
   mypsf.radial frozen            0            0            1           
   mypsf.norm   frozen            1            0            1           

sherpa In [5]: mypsf.pars
Out[5]: ()
```

This is using the CIAO 4.13 version on Linux. I can provide files, but any PSF should do.

In this case I'd have expected `mypsf.pars` to return a tuple of `Parameter` objects (or similar) akin to other model components

```python
sherpa In [6]: gauss2d.g1()
Out[6]: <Gauss2D model instance 'gauss2d.g1'>

sherpa In [7]: g1.pars
Out[7]: 
(<Parameter 'fwhm' of model 'g1'>,
 <Parameter 'xpos' of model 'g1'>,
 <Parameter 'ypos' of model 'g1'>,
 <Parameter 'ellip' of model 'g1'>,
 <Parameter 'theta' of model 'g1'>,
 <Parameter 'ampl' of model 'g1'>)
```

DougBurke:
This could be intentional or it's just an error in the convolution-style code. It's unclear, given the lack of documentation there, but my bet is the latter.

If you apply the psf model to a component - e.g.

```
m = mypsf(g1)
```

then I **believe** you can use `m.pars` (but I haven't tried, so confirmation from someone other than me would be nice).

There is the possibility that it only makes sense to have parameters on such a combined object, but then why does `print(mypsf)` talk about parameters...

kglotfelty:
applying the psf just gets you the parameters of the model it's applied to

```python
sherpa In [7]: x = mypsf(g)

sherpa In [8]: x.pars
Out[8]: 
[...]

PR #1116, opened 2021-03-22
Add parameter-based tests for the PSF model

DougBurke:
# Summary

Adds several tests of edge-case handling of parameters for PSF convolution models.

# Details

This was meant to solve #1109 but "Eek. Turns out to not-at-all-be-easy™ so this may just morph into some basic tests..." so we have some test if I ever do work out how to improve #1109.


anetasie:
@DougBurke expanding number of tests is good. Thanks!
Yes, there are at least two ways to define a PSF - the file/array base and the model base.

anetasie (approved):

DougBurke:
@anetasie - these tests are pretty specialised, so they don't really add much, unfortunately. For instance, your comment about two ways to create a PSF model is irrelevant here, as there is no test of that part of the code - this just checks what parameters are reported for some simple cases.

Issue #1087, opened 2021-01-29
plot_source ignores the rate setting of a dataset


DougBurke:
```
get_data().rate = True
plot('source', 'model')
get_data().rate = False
plot('source', 'model')
```

creates 

![sherpa](https://user-images.githubusercontent.com/224638/106279281-297a2d80-620a-11eb-97b3-371811fa6402.png)

and then

![sherpa](https://user-images.githubusercontent.com/224638/106279325-3c8cfd80-620a-11eb-8a37-30fc0bc51e0f.png)

Note that the source plot has not changed y units. Changing get_data().plot_fac does change the source plot.

Is this something we care about or not?

hamogu:
I think we do care.

hamogu:
And is it not the best to stumble on questions like this while celebrating a major birthday?

DougBurke:
Unfortunately I've had this in my back pocket for a day or two. I just like to space out my github profile...

DougBurke:
So, let's consider the case of the model `xspowerlaw.pl`. When we call `plot_source` we evaluate the model `pl`, and when we evaluate `plot_model` we evaluate the model `rmf(arf(exposure * pl))` (for the case of a "standard" PHA case where we are not doing anything fancy). For the `plot_model` case we have an exposure time which we can include or not, whereas for the `plot_source` case we don't. This is important because the models (in this case `pl`) are taken to return a value per-second (or, can be thought of as representing a single-second exposure), which is why the `exposure` term gets automatically added for `plot_model`. 

Therefore, at some level setting `set_analysis(..., type='counts')` or `get_data().rate = False` could be considered to be an error: we don't have an exposure time to use. Now, technically we do have an exposure time (from the data object), so we could use it for the `counts` case, but then we'd be plotting `photons/cm^2/keV` (or, with `plot_fac=1`, `photons/cm^2`), and I'm not sure what use this has (other than "doing something different" when the rate setting is changed).

hamogu:
I have a nagging thought that tells me that we are rapidly nearing the end of where we can get with Sherpa's simplistic units system. Astropy's units system is build to deal exactly with cases like this and it's far more complex. I'm sure we can fix this one thing, but then how about nm as unit of wavelength or frequency?  My understanding is that tools like [3ML](https://threeml.readthedocs.io/en/stable/index.html) use astropy.units for this exact reason.

DougBurke:
Closed by #1828

PR #1118, opened 2021-03-29
Improve test cases for source plots with PHA data

DougBurke:
# Summary

Add additional test cases for source plots with PHA data to check out rarely-used combinations.

# Notes

This was originally intended to address #1087 - hence the addition of tests - but I'm now not sure what to do with that issue, so (unless we get some communal insight soon) it makes sense to at least add the tests.

As sherpa.astro.plot.SourcePlot does not use the "normal" routines for evaluating the data to plot (that is, using the to_plot method of the data object) we need to check the behavior for a number of cases like

  - changing units
  - changing rate
  - changing the factor setting

The test has also changed from using integer values for comparison back to an approximate equality test due to changes in how the data is compared (e.g. multiplying by the bin width, particularly for the wavelength check).

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -274,7 +274,7 @@ class SourcePlot(HistogramPlot):
         sqr = to_latex('^2')
 
         self.xlabel = f'{self.units.capitalize()} ({quant})'
-        self.ylabel = '%s  Photons/sec/cm' + sqr + '%s'
+        self.ylabel = f'%s  Photons/sec/cm{sqr}%s'
 
         if data.plot_fac == 0:
             self.y /= xmid
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -245,9 +245,17 @@ class SourcePlot(HistogramPlot):
             self.units = "energy"
 
         self.xlabel = data.get_xlabel()
-        self.title = 'Source Model of %s' % data.name
+        self.title = f'Source Model of {data.name}'
         self.xlo, self.xhi = data._get_indep(filter=False)
+
+        # Why do we not apply the mask at the end of prepare?
+        #
         self.mask = filter_bins((lo,), (hi,), (self.xlo,))
+
+        # The source model is assumed to not contain an instrument model,
+        # and so it evaluates the expected number of photons/cm^2/s in
+        # each bin (or it can be thought of as a 1 second exposure).
+        #
         self.y = src(self.xlo, self.xhi)
         prefix_quant = 'E'
         quant = 'keV'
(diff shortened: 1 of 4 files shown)


DougBurke:

DougBurke:
Rebased because of conflict.

hamogu (approved):
The tests run and pass, so at the very least they check what's currently the state of the code. The strings are tested, so they are definitely OK. Can be merged as is.

sherpa/astro/plot.py
```diff
@@ -265,23 +273,21 @@ def prepare(self, data, src, lo=None, hi=None):
...
-                                                       prefix_quant), '')
+            self.ylabel = self.ylabel % (f'{prefix_quant} f({prefix_quant})', '')
 
         elif data.plot_fac == 2:
             self.y *= xmid
-            self.ylabel = self.ylabel % ('%s%s f(%s)' % (prefix_quant, sqr,
-                                                         prefix_quant),
-                                         ' %s ' % quant)
+            self.ylabel = self.ylabel % (f'{prefix_quant}{sqr} f({prefix_quant})',
+                                         f' {quant} ')
```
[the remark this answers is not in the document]
DougBurke: I do admit to having an irrational dislike of the `%` version and prefer the f-string, but you can see here I left both in place ;-)

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -25,22 +28,172 @@
...
+    assert sp.xlabel == 'Wavelength (Angstrom)'
+
+    if factor == 0:
+        assert sp.ylabel.startswith('f(lambda)  Photons/sec/cm')
+        assert sp.ylabel.endswith('/Angstrom ')
+    elif factor == 1:
+        assert sp.ylabel.startswith('lambda f(lambda)  Photons/sec/cm')
+        assert sp.ylabel.find('/Angstrom ') == -1
+    elif factor == 2:
+        # This says lmabda^2 f(lambda) ... but the exact format depends on
```
wmclaugh, line 124: lmabda^2 -> lamba^2

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -25,22 +28,172 @@
...
 
 
-def test_sourceplot():
+def check_sourceplot_energy(sp, rate=True, factor=0):
+    """Check for test_sourceplot/test_sourceplot_channel
+
+    rate determines if the plot is for a "rate" or "counts",
+    although at the moment it does not change anything.
+    """
+
```
wmclaugh, line 45: Rather than stub in rate, the API seems straight forward enough to add rate  in later when you are ready to use it (since the rest of the script appears to be using named rather than positional parameters). The runtime savings are negligible but the code looks cleaner without unused hooks/stubs and comments noting they are unused.

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -25,22 +28,171 @@
...
 from sherpa import stats
 from sherpa.utils.err import IOErr
-from sherpa.utils.testing import requires_pylab
 
 
-def test_sourceplot():
+def check_sourceplot_energy(sp, factor=0):
+    """Check for test_sourceplot/test_sourceplot_channel
+
+    Note that the rate setting does not change these tests.
```
wmclaugh, line 42: Is this comment currently applicable?

DougBurke: Yes - `test_sourceplot` has `assert data.rate` and `test_sourceplot_counts` has `data.rate = False` and both call `check_sourceplot_energy`. I wanted some note that the rate setting doesn't change these tests since we may want to change this at some point. I orignially added these tests so I could then go and address #1087 but then realized it would take some review and discussion, and it was better to get tests added of the existing functionality.

PR #1121, opened 2021-04-06
Hide AstroPy 4.2.1 FITS-related warnings

DougBurke:
# Summary

Hides warning messages created by AstroPy 4.2.1 when trying to read in FITS files.

# Details

astropy.io.fits.open now creates warning messages about invalid FITS
structures when given a non-FITS file. Unfortunately this leads to
a lot of confusing messages for the user of some of the top-level
routines - like sherpa.astro.io.load_data - when given an ASCII
file, since the code will try to load a FITS file multiple times
before trying as an ASCII file.

We avoid this by hiding all warnings from astropy.io.fits when
we try to open a FITS file. This means that we will not see useful
warnings that are generated when the input is a FITS file, but I
think that is a valid compromise. We could re-create the warnings,
as we can change to something like

    with warning.catch_warnings() as ws:
        out = fits.open(fname)
        for w in ws:
            <somehow recreate the w>

but it's not completely obvious how to recreate the warnings (as
AstroPy has over-ridden the warnings module and I did not feel
it's worth the effort to understand what's going on here).

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2011, 2015, 2016, 2017, 2018, 2019, 2020, 2021
-#    Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -39,6 +39,7 @@ References
 
 import logging
 import os
+import warnings
 
 import numpy
 from numpy.compat import basestring
@@ -285,7 +286,20 @@ def open_fits(filename):
        function.
     """
     fname = _infer_and_check_filename(filename)
-    return fits.open(fname)
+
+    # With AstroPy v4.2.1 we start to see warnings when this call
+    # fails (i.e. when the argument is not a FITS file).  This leads
+    # to spurious messages to the user, so we just remove all
+    # warnings. This may need tweaking at some point: ideally we would
+    # only want to hide the warnings when it is not a FITS file but
+    # show them when it is a FITS file.
+    #
+    # Note that this is not thread safe.
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore', module='astropy.io.fits')
+        out = fits.open(fname)
+
+    return out
 
 
 def read_table_blocks(arg, make_copy=False):


hamogu (commented):
We should think about `with warnings.catch_warnings(record=True) as w`. Then, we can save the warnings to some internal variable: `pyfits_backend._warnings_we_caught = w`.
That way, we have it to debug for us and if users report problems we can ask them to 
```
print(pyfits_backend._warnings_we_caught)
```
to help us understand what's going on.
In `astropy.io.ascii` we have something similar and it sometimes comes in handy when things go wrong, but you don't know why. If we put that in the code, then we'll always have a way to access the warnings without changing the code.

DougBurke:
@hamogu I assume the idea that this is intended for simple debugging so you don't care about thread safety or multiple calls?

One issue with how our code - like sherpa.astro.io.load_data - works is that it ends up calling the "open fits" call multiple times - e.g. for an ASCII file then it runs through something like

- check it's a PHA file
- check it's a FITS image file
- check it's a FITS table file 
- check it's an ASCII file

so technically we could lose information at the "load_data" level, but I think that is okay (as you can always use the type-specific routines like load_image/load_pha when debugging).

hamogu:
Yes, it's purely for debugging, which is why the suggested name starts with an _, so thread-safety et. does not matter. You could do `pyfits_backend._keep_warnings = []` and then use append. That way, you have all of them, but I did not want to make it complicated and I also did not want a memory hog, if someone reads 100000 files. 

Look at https://github.com/astropy/astropy/blob/master/astropy/io/ascii/ui.py#L43 and all the lines in that module that touch `_read_trace`. I don't think we need something this complete here, but in astropy it came in handy a few times to diagnose really non-obivous failures that were hidden because warnings and exceptions were caught and the user never saw them - often enough that Tom A. went through the trouble of even implementing a `get_read_trace` function.

DougBurke:
I think there's a subtle difference between what AstroPy is doing - tracking success and errors - and what we'd do here (tracking warnings). I think until we get experience of tracking down problems - i.e. perhaps we want to add in a bunch of debug logging or something more like the AstroPy version - we shouldn't try to work too hard on this (ie just do something simple until we find we need something more).

hamogu:
@DougBurke I fully agree, that's why my original suggestion is not thread save and does not preserve everything but just the most recent warning.

DougBurke:
So, this is turning out "interesting" as I was thinking about tracking the warnings from the "error" case, but that is actually not really meaningful since

```
with warnings.check_warnings(record=True) as ws:
    out = fits.open('a.tbl')
    _catch_warnings = ws  # I am going to assume we don't need to deepcopy.copy(ws) here
```

then we only get to record the warning when the open call succeeds. However, that makes sense since you get an error message which is the more-important information.

What we do get is warnings from calls to open that succeed, which is probably more useful to track... So the above text is really just me working through some things!

One actually slightly-useful question is if you know whether I should save a copy of the warning or not. I don't think so (it's not like these objects should be carrying around state/information we need to deepcopy).

hamogu:
I don't know but I would not worry about it. I already regret suggesting it because it's more complicated that I thought ;-)

DougBurke:
So, I got as far as the following, but I don't believe it's worth it and would rather hold off from something like this until we find a use case.

```
% git diff sherpa
diff --git a/sherpa/astro/io/pyfits_backend.py b/sherpa/astro/io/pyfits_backend.py
index cde5e0a7..8301e27b 100644
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -69,6 +69,15 @@ __all__ = ('get_table_data', 'get_image_data', 'get_arf_data', 'get_rmf_data',
            'get_column_data', 'get_ascii_data')


+# Track FITS open warnings from AstroPy in the open_fits routine.
+#
+# This is intended for debugging and is not guaranteed to work in all cases
+# (e.g. it is not guaranteed to be thread safe). Is also not intended to be
+# a stable interface,
+#
+_open_warnings = None
+
+
 def _has_hdu(hdulist, name):
     try:
         hdulist[name]
@@ -284,20 +293,28 @@ def open_fits(filename):
     out
        The return value from the `astropy.io.fits.open`
        function.
+
+    Notes
+    -----
+    AstroPy, from version 4.2.1, can create warnings when trying to open
+    a FITS file. These warnings are hidden by this routine, but can
+    be queried with the _open_warnings module variable. This setting is
+    intended only for debugging purposes.
+
     """
+
+    global _open_warnings
     fname = _infer_and_check_filename(filename)

-    # With AstroPy v4.2.1 we start to see warnings when this call
-    # fails (i.e. when the argument is not a FITS file).  This leads
-    # to spurious messages to the user, so we just remove all
-    # warnings. This may need tweaking at some point: ideally we would
-    # only want to hide the warnings when it is not a FITS file but
-    # show them when it is a FITS file.
+    # AstroPy v4.2.1 can generate warnings (e.g. if given an ASCII table
+    # file when this call errors out and generates warnings) which is
+    # confusing.
     #
     # Note that this is not thread safe.
-    with warnings.catch_warnings():
-        warnings.filterwarnings('ignore', module='astropy.io.fits')
+    _open_warnings = None
+    with warnings.catch_warnings(record=True) as ws:
         out = fits.open(fname)
+        _open_warnings = ws

     return out

```

DougBurke:
I'm happy with taking this as is, but I may be a tad biased... We can think about potential improvements to astropy-IO code as part of #1091

hamogu:
Agreed. Take it as is.

hamogu (approved):

wmclaugh (approved):

PR #1122, opened 2021-04-08
Update ds9 download

DougBurke:
# Summary

Update DS9 tests to use ubuntu18 and darwinhighsierra as ubuntu14 and darwinsierra are no longer supported. The DS9 version has been bumped from 8.2 to 8.2.1 as this is the latest version.

# Details

This only changes the tests, which were failing because of the DS9 changes. There are also a number of "canary" lines added to the output of setup_xspec_ds9.sh to make tracking down this sort of error easier in the future.

The DS9 supported versions are taken from http://ds9.si.edu/download/


Marie-Terrell (approved):
Awesome! Thank you for tracking that down! Everything looks good!

PR #1126, opened 2021-04-14
Tweak plot docs

DougBurke:
# Summary

Adds a new notebook that shows off a number of plots created with matplotlib and exports the ScatterPlot, TracePlot, CDFPlot, PDFPlot, and LRHistogram classes from sherpa.plot.

# Details

This was created to help test out #1125 - testing out a new matlpotlib release - but has been separated from that as it is seperate from it.

A notebook has been added to notebooks/ and added to the readthedocs site which shows off a number of basic plots in matlpotlib. This was used to check that #1125 didn't change anything obvious (as the tests don't really test out the appearance of the plots for various reasons).

In doing this, I noticed that several classes were not being exorted by `sherpa.plot`, which makes it hard to find them - e.g. help(sherpa.plot) - so I addressed this.

One thing to think about is if the notebook title should be changed as it is a bit naff: "A sample of plots".


DougBurke:
@hamogu - since the notebook is included in the read-the-docs build I don't need to send you the notebook but just point you at https://sherpa--1126.org.readthedocs.build/en/1126/ExamplePlots.html

My main concern is that the title - `A sample of plots` - is less-than optimal!

DougBurke:
A more-serious issue is that I made up the PHA/ARF before working out what to plot- they therefore don't really match up well. I think it may make sense to come up with a different functional form for the ARF and/or PHA data for the plots.

hamogu:
I think the title is fine. It's just what it is: A sample of notebooks. 
We might want to rethink that when we get to a dozen notebooks and a need arises to separate notebooks for new users (sherpa quick start), notebook for showing off features that are easier to show in a notebook than in narrative rst docs ("Notebook support in Sherpa" and this one), and possibly other catagories (e.g. threads using the OO interface to work through a full science example), but for now, with just three notebooks, that's too complicated.

hamogu:
I agree that the arf/pha combination is not ideal, but that's not the point here. Maybe it's even good to have a clearly non-sensical shape (I've never seen a sinusoidal shape in a PHA file) to make clear that this is not meant to be real, but just for testing / demonstration.

hamogu (approved):

notebooks/ExamplePlots.ipynb
```diff
@@ -0,0 +1,766 @@
...
+   ],
+   "source": [
+    "plot.DataHistogramPlot.histo_prefs"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Previously we explicitly set the error values, but we can also use one of the chi-square statistics to come up with error values. In this case it's just the square-root of the data value (so, for for $x \\sim 1$ bin, we have an error of 0):"
```
wmclaugh, line 312: Is there an extraneous 'for' in '(so, for for $x...'?

wmclaugh: @DougBurke-  I went ahead and tweaked two lines in ExamplePlots.ipynb:
    "(so, for for..."   -> "(so, for ..."
    "numper" -> "number"

Hope you don't mind.

DougBurke: No problem

Issue #1081, opened 2021-01-27
group_xxx, tabStops, and mask handling


DougBurke:
This may relate to ideas to implement #956. Not sure how this relates to #718

# Issue 1

This is part of the grouping/filtering work, and in this case is because the `mask` attribute is grouped but for the group functions we need the ungrouped data.

```
In [2]: load_data('sherpa-test-data/sherpatest/3c273.pi')
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file sherpa-test-data/sherpatest/3c273.arf
read RMF file sherpa-test-data/sherpatest/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273_bg.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file sherpa-test-data/sherpatest/3c273_bg.pi

In [3]: ignore(None, 0.5)

In [4]: ignore(7, None)

In [5]: d = get_data()

In [6]: group_counts(20, tabStops=~d.mask)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-3b601a5fe044> in <module>
----> 1 group_counts(20, tabStops=~d.mask)

<string> in group_counts(id, num, bkg_id, maxLength, tabStops)

~/sherpa/sherpa-docs/sherpa/astro/ui/utils.py in group_counts(self, id, num, bkg_id, maxLength, tabStops)
   8132         if bkg_id is not None:
   8133             data = self.get_bkg(id, bkg_id)
-> 8134         data.group_counts(num, maxLength, tabStops)
   8135
   8136     # DOC-TODO: check the Poisson stats claim; I'm guessing it means

~/sherpa/sherpa-docs/sherpa/astro/data.py in group_counts(self, num, maxLength, tabStops)
   2561             raise ImportErr('importfailed', 'group', 'dynamic grouping')
   2562         self._dynamic_group(pygroup.grpNumCounts, self.counts, num,
-> 2563                             maxLength=maxLength, tabStops=tabStops)
   2564         for bkg_id in self.background_ids:
   2565             bkg = self.get_background(bkg_id)

[...]

Issue #921, opened 2020-08-31
DataPHA.get_ebins() with analysis=channel returns +- 0.5 bin widths


DougBurke:
```
ui.load_pha(..)
d = ui.get_data()
d.set_analysis('channel')
xlo, xhi = d._get_ebins(group=False)
```

xlo is [0.5, 1.5, ...]
xhi is [1.5, 2.5, ...]

In other places we use channel, channel + 1 rather than channel-0.5, channel+0.5

Which is best?

anetasie:
hmm.... interesting question.  I guess this is mainly relevant to plotting, while in the analysis it should follow the translation between the channel number and the energy range for the channel. How is the channel defined in the response files?

DougBurke:
I agree that this is primarily a plotting issue (although channel numbers do get used in filtering, so something to keep aware of).

This was mainly something I noted but am not planning on addressing right now. Unless by accident because of something else ;-)

PR #1127, opened 2021-04-16
PHA filtering changes

DougBurke:
# Summary

Improve some corner cases for filtering PHA data, including fixing #921 (channel bounds are integers, not half-integers).

# Details

I had originally started out trying to work on #1081 but there turns out to be a number of small corner cases that it makes sense to address. They could be spun out into separate PRs but as they all deal with the same code the rebasing would be excessive, and some of the fixes build on each other.

The things to think about are

a - should the ui layer ignore and notice error out if there's no data? The code had an explicit check for it but it was wrong, so never fired. Given some of our recent work to allow things like multiple calls to subtract not error-ing out then we may want to keep the existing behavior (i.e. the command is essentially a no-op). This would be an easy change to make to this PR.

b - should we allow negative wavelength or energy values in filters? We explicitly support a value of 0 since responses can sometimes contain this (even if against the OGIP standard) and it acts as a nice "smaller-than-everything" value (even if we can use None for that). This PR supports handling negative wavelengths "sensibly" but maybe this should error out. I am reluctant in this PR to actually make that change since I worry about testing it out (and it may be only me who thinks to break the system in this way)

c - the last significant change in this PR adds a bunch of assert statements into the filter code to check that I understand the code (there are odd times when the code decides to allow lo and hi to be swapped and there's no commentary on why this is done, so I don't know if it's just pointless defensive coding or there's some underlying reason to do with our grating responses which the code just doesn't make clear). These asserts should be removed but I'd like to see something like the SDS regression tests run through this code first to make sure we've covered our bases.

The changes are

- internal code clean up (relatively minor, but this is obviously old code that has been changed over time)
- ignore and filter will now report an error if there are no datasets loaded (it might be argued that doing nothing in this case is okay, but the code was meant to report an error, presumably lost in the python 2 to 3 conversions, given the code change)
- ensures that for units=channel the edges are channel,channel+1 not channel-0.5,channel+0.5
- handle negative wavelengths the same way we handle 0 (maybe we should error out in this case?)
- turn a filter range where both low and high limits lie outside the valid range and are both either less than or greater than the valid range into no-ops - previously the first or last bin would get filtered which was surprising (particularly when coupled with the negative wavelength handling)
- explicitly error out if given a string range: the idea is that the support for strings is only in the ui layer and at the object layer you only deal with numbers (or None) - it makes the DataPHA code better match the standard data classes and is a very-minor improvement in code clarity. Actually, I've since discovered that the original code did end up error-ing out if given a string, but thanks to refactoring the code it helps to add an explicit check for this case.
- the channels and counts arrays are converted to NumPy arrays if they are not when a DataPHA object is created - this is not an issue for people loading in a PHA file but can be for those few of us who are creating DataPHA objects directly - as the change cleans up some odd cases (there may be some decision on handling of masked arrays but I feel that can be worked o separately, and may not even be an issue)

# Examples

## ignore/filter with no data

In CIAO 4.13 we have (the `ignore_id` behavior is just included to show this does not change)

```
sherpa In [1]: ignore()

sherpa In [2]: ignore_id(1, None, None)
IdentifierErr: data set 1 has not been set
```

and with this PR we get

```

In [2]: ignore()
... ignoring traceback

IdentifierErr: No data sets found

In [3]: ignore_id(1, None, None)
... ignoring traceback

IdentifierErr: data set 1 has not been set
```

## channel +/- 1 vs 0.5

In CIAO 4.13 we have (this is issue #921)

```
sherpa In [3]: d = DataPHA('x', np.arange(1, 4), np.asarray([1, 2, 0]))

sherpa In [4]: d._get_ebins()
Out[4]: (array([0.5, 1.5, 2.5]), array([1.5, 2.5, 3.5]))
```

and now we have

```
In [4]: d = DataPHA('x', np.arange(1, 4), np.asarray([1, 2, 0]))

In [5]: d._get_ebins()
Out[5]: (array([1, 2, 3]), array([2, 3, 4]))
```

As mentioned, this is an internal routine.

In both cases (before and after this change) we haven't changed the DataPHA plots since it uses integer values for the xlo and xhi values (shown below is the CIAO 4.13 output, it's the same after thie PR)

```
sherpa In [2]: set_analysis('channel')

sherpa In [3]: get_data_plot().xlo
Out[3]:
array([  1.,  18.,  22.,  33.,  40.,  45.,  49.,  52.,  55.,  57.,  60.,
        62.,  66.,  69.,  72.,  76.,  79.,  83.,  89.,  97., 102., 111.,
       117., 125., 131., 134., 140., 144., 151., 157., 165., 178., 187.,
       197., 212., 233., 245., 261., 277., 292., 324., 345., 369., 405.,
       451., 677.])

sherpa In [4]: get_data_plot().xhi
Out[4]:
array([  18.,   22.,   33.,   40.,   45.,   49.,   52.,   55.,   57.,
         60.,   62.,   66.,   69.,   72.,   76.,   79.,   83.,   89.,
         97.,  102.,  111.,  117.,  125.,  131.,  134.,  140.,  144.,
        151.,  157.,  165.,  178.,  187.,  197.,  212.,  233.,  245.,
        261.,  277.,  292.,  324.,  345.,  369.,  405.,  451.,  677.,
       1025.])
```

## converting to numpy arrays

In CIAO 4.13 we have the following - note that there's a comment about converting the counts array (the [1,2,0] array) to Numpy but it turns out d.counts is still a list (it's only d.y that is changed, and we don't use really use the .y fie when accessig data from DataPHA). Also note that the internal _get_ebins routine fails.

```
sherpa In [1]: d = DataPHA('x', [1, 2, 3], [1, 2, 0])
/home/dburke/anaconda/envs/ciao413/lib/python3.8/site-packages/sherpa/data.py:50: UserWarning: Converting array [1, 2, 0] to numpy array.

sherpa In [2]: print(d)
name           = x
channel        = [1, 2, 3]
counts         = [1, 2, 0]
staterror      = None
syserror       = None
bin_lo         = None
bin_hi         = None
grouping       = None
quality        = None
exposure       = None
backscal       = None
areascal       = None
grouped        = False
subtracted     = False
units          = channel
rate           = True
plot_fac       = 0
response_ids   = []
background_ids = []

sherpa In [3]: type(d.channel)
Out[3]: list

sherpa In [4]: type(d.counts)
Out[4]: list

sherpa In [5]: type(d.x)
Out[5]: list

sherpa In [6]: type(d.y)
Out[6]: numpy.ndarray

sherpa In [7]: d._get_ebins()
TypeError: unsupported operand type(s) for -: 'list' and 'float'
```

With this patch we get the following (both channel and counts are stored as NumPy arrays and _get_ebins no longer errors out):

```
In [1]: from sherpa.astro.data import DataPHA

In [2]: d = DataPHA('x', [1, 2, 3], [1, 2, 0])
/home/dburke/sherpa/sherpa-master/sherpa/data.py:50: UserWarning: Converting array [1, 2, 3] to numpy array.
/home/dburke/sherpa/sherpa-master/sherpa/data.py:50: UserWarning: Converting array [1, 2, 0] to numpy array.

In [3]: print(d)
name           = x
channel        = Int64[3]
counts         = Int64[3]
staterror      = None
syserror       = None
bin_lo         = None
bin_hi         = None
grouping       = None
quality        = None
exposure       = None
backscal       = None
areascal       = None
grouped        = False
subtracted     = False
units          = channel
rate           = True
plot_fac       = 0
response_ids   = []
background_ids = []

In [4]: type(d.channel)
Out[4]: numpy.ndarray

In [5]: type(d.counts)
Out[5]: numpy.ndarray

In [6]: type(d.x)
Out[6]: numpy.ndarray

In [7]: type(d.y)
Out[7]: numpy.ndarray

In [8]: d._get_ebins()
Out[8]: (array([1, 2, 3]), array([2, 3, 4]))
```

## filtering with silly values

In CIAO 4.13 we have

```
sherpa In [2]: set_analysis('wave')

sherpa In [3]: get_filter()
Out[3]: '0.999066753912:99.322428665279'

sherpa In [4]: ignore(0, 2)

sherpa In [5]: get_filter()
Out[5]: '2.200017578690:99.322428665279'

sherpa In [6]: notice()

sherpa In [7]: ignore(-1, 2)

sherpa In [8]: get_filter()
Out[8]: '0.999066753912:99.322428665279'
```

whereas we now have

```
In [4]: set_analysis('wave')

In [5]: get_filter()
Out[5]: '0.999066753912:99.322428665279'

In [6]: ignore(0, 2)

In [7]: get_filter()
Out[7]: '2.200017578690:99.322428665279'

In [8]: notice()

In [9]: ignore(-1, 2)

In [10]: get_filter()
Out[10]: '2.200017578690:99.322428665279'
```

Notice that the negative wavelength is treated like 0 now whereas in CIAO 4.13 it didn't appear to do anything (note that the behavior depended on the data range and what values you used)

## filtering with "excessive" values

If you gave a filter with both low and high values "outside" the data range then the first or last bin could get filtered, as shown in the example below. We now just drop the filter in this condition.

With CIAO 4.13, even though I've said ignore 2000 to 2000 keV we actually end up losing the last group - here ~8 to ~12 keV

```
sherpa In [9]: notice()

sherpa In [10]: set_analysis('energy')

sherpa In [11]: get_filter()
Out[11]: '0.124829999695:12.410000324249'

sherpa In [12]: ignore(2000, 3000)

sherpa In [13]: get_filter()
Out[13]: '0.124829999695:8.219800233841'
```

The new behavior is to do nothing

```
In [11]: notice()

In [12]: set_analysis('energy')

In [13]: get_filter()
Out[13]: '0.124829999695:12.410000324249'

In [14]: ignore(2000, 3000)

In [15]: get_filter()
Out[15]: '0.124829999695:12.410000324249'
```

## string filtering check

This only deals with the `notice` and `ignore` methods of the `DataPHA` class - you can still give strings to the `notice` and `ignore` functions from the ui layer. It also doesn't change the user behavior - that is the following is from CIAO 4.13 and we still have the same error raised, it's just now done earlier

```
In [5]: pha.notice("0.5:2.0")
DataErr                                   Traceback (most recent call last)
<ipython-input-5-05cff9c3672d> in <module>
----> 1 pha.notice("0.5:2.0")

~/anaconda/envs/ciao413/lib/python3.8/site-packages/sherpa/astro/data.py in notice(self, lo, hi, ignore, bkg_id)
   3456                 # do nothing (other than display an INFO message).
   3457                 #
-> 3458                 bkg.notice(lo, hi, ignore)
   3459             finally:
   3460                 bkg.units = old_bkg_units

~/anaconda/envs/ciao413/lib/python3.8/site-packages/sherpa/astro/data.py in notice(self, lo, hi, ignore, bkg_id)
   3507         groups = self.apply_grouping(self.channel,
   3508                                      self._make_groups)
-> 3509         self._data_space.filter.notice((lo,), (hi,),
   3510                                        (groups,), ignore)
   3511

~/anaconda/envs/ciao413/lib/python3.8/site-packages/sherpa/data.py in notice(self, mins, maxes, axislist, ignore)
    453         ignore = bool_cast(ignore)
    454         if str in [type(min) for min in mins]:
--> 455             raise DataErr('typecheck', 'lower bound')
    456         elif str in [type(max) for max in maxes]:
    457             raise DataErr('typecheck', 'upper bound')

DataErr: strings not allowed in lower bound list
```

whereas this PR throws the error earlier

```
In [3]: pha.notice("0.5:2.0")
DataErr                                   Traceback (most recent call last)
<ipython-input-3-05cff9c3672d> in <module>
----> 1 pha.notice("0.5:2.0")

~/sherpa/sherpa-master/sherpa/astro/data.py in notice(self, lo, hi, ignore, bkg_id)
   3794             if isinstance(val, str):
   3795                 # match the error seen from other data classes here
-> 3796                 raise DataErr('typecheck', f'{label} bound')
   3797
   3798         # If any background IDs are actually given, then impose

DataErr: strings not allowed in lower bound list
```

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3487,6 +3487,11 @@ class DataPHA(Data1D):
 
         ignore = bool_cast(ignore)
 
+        # This condition is checked for in the _data_space.filter call
+        # at the end of the method, but it is easier to enforce it
+        # here so we do not need to worry about possible type errors
+        # when comparing string and number values.
+        #
         for val, label in zip([lo, hi], ['lower', 'upper']):
             if isinstance(val, str):
                 # match the error seen from other data classes here
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3529,10 +3529,57 @@ class DataPHA(Data1D):
             self.quality_filter = None
             self.notice_response(False)
 
+        # When convert_limit is called, any limit that is outside the
+        # supported range (e.g. lo = 0.01 when the response is 0.1 to
+        # 11 keV) will get clamped to the minumum or maximum value.
+        # This is okay when lo is smaller and hi is larger than the
+        # supported values but is a problem when they are both
+        # smaller or larger than the range, since the result is then
+        # that the first or last group would be filtered. It is
+        # therefore necessary to check for this condition.
+        #
+        # This would be easier if _to_channel were to identify that
+        # the range had been (or should be) clamped, but that is a
+        # larger change than I want to implement, so instead we repeat
+        # some of the checks that the _xxx_to_channel methods make to
+        # see what the minimum and maximum allowed values are.
+        #
+        try:
+            elo, ehi = self._get_ebins(group=False)
+        except DataErr as de:
+            info(f"Skipping dataset {self.name}: {de}")
+            return
+
+        if self.units == 'energy':
+            umin = elo[0]
+            umax = ehi[-1]
+        elif self.units == 'wavelength':
+            umin = self._hc / ehi[-1]
+            umax = self._hc / elo[0]
+        else:
+            # assume channel units
+            umin = self.channel[0]
+            umax = self.channel[-1]
+
+        assert umin < umax, (self.units, umin, umax)
+
+        # Finally we can check that the request is not
+        # outside the valid range "on the same side".
+        #
+        if lo is not None and hi is not None:
+            assert lo <= hi, (lo, hi)
+            if hi <= umin:
+                return
+            if lo >= umax:
+                return
+
         # We do not want a "all data are masked out" error to cause
         # this to fail; it should just do nothing (as trying to set
         # a noticed range to include masked-out ranges would also
-        # be ignored).
+        # be ignored). Note that originally this try/except caught
+        # errors like "RMF has no ebounds data", but this is now
+        # caught earlier in the call to _get_ebins, so it is possible
+        # that the check is no-longer needed.
         #
         # Convert to "group number" (which, for ungrouped data,
         # is just channel number).
@@ -3560,7 +3607,11 @@ class DataPHA(Data1D):
             if hi is None:
                 return
 
-        elo, ehi = self._get_ebins()
+        # At this point lo and hi are in "group number".
+        #
+        # When do we need to flip the limits? It is not clear whether
+        # we always need lo <= hi.
+        #
         if ((self.units == "wavelength" and
              elo[0] < elo[-1] and ehi[0] < ehi[-1]) or
             (self.units == "energy" and
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1760,8 +1760,8 @@ class DataPHA(Data1D):
         group = bool_cast(group)
 
         if self.units == 'channel':
-            elo = self.channel - 0.5
-            ehi = self.channel + 0.5
+            elo = self.channel
+            ehi = self.channel + 1
         else:
             arf, rmf = self.get_response(response_id)
             if (self.bin_lo is not None) and (self.bin_hi is not None):
@@ -1779,8 +1779,8 @@ class DataPHA(Data1D):
                 elo = arf.energ_lo
                 ehi = arf.energ_hi
             else:
-                elo = self.channel - 0.5
-                ehi = self.channel + 0.5
+                elo = self.channel
+                ehi = self.channel + 1
 
         # If the data are grouped, then we should group up
         # the energy bins as well.  E.g., if group 1 is
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3440,8 +3440,53 @@ class DataPHA(Data1D):
             _notice_resp(noticed_chans, arf, rmf)
 
     def notice(self, lo=None, hi=None, ignore=False, bkg_id=None):
+        """Notice or ignore the given range.
+
+        Parameters
+        ----------
+        lo, hi : number or None, optional
+            The range to change. A value of None means the minimum or
+            maximum permitted value. The units of lo and hi are set by
+            the units field.
+        ignore : bool, optional
+            Set to True if the range should be ignored. The default is
+            to notice the range.
+        bkg_id : int or sequence of int or None, optional
+            If not None then apply the filter to the given background
+            dataset or datasets, otherwise change the object and all
+            its background datasets.
+
+        See Also
+        --------
+        get_filter, get_filter_expr, get_mask
+
+        Notes
+        -----
+        If no channels have been ignored then a call to `notice` with
+        `ignore=False` will select just the `lo` to `hi` range, and
+        exclude any channels outside this range. If there has been a
+        filter applied then the range `lo` to `hi` will be added to the
+        range of noticed data (when `ignore=False`).
+
+        So, for an ungrouped PHA file with 1024 channels:
+
+        >>> pha.units = 'channel'
+        >>> pha.get_filter()
+        '1:1024'
+        >>> pha.notice(20, 200)
+        >>> pha.get_filter()
+        '20:200'
+        >>> pha.notice(300, 500)
+        '20:200,300:500'
+
+        """
 
         ignore = bool_cast(ignore)
+
+        for val, label in zip([lo, hi], ['lower', 'upper']):
+            if isinstance(val, str):
+                # match the error seen from other data classes here
+                raise DataErr('typecheck', f'{label} bound')
 
         # If any background IDs are actually given, then impose
         # the filter on those backgrounds *only*, and return.  Do
@@ -3501,12 +3546,12 @@ class DataPHA(Data1D):
                 info(f"Skipping dataset {self.name}: {de}")
                 return None
 
-        if lo is not None and not isinstance(lo, str):
+        if lo is not None:
             lo = convert_limit(lo)
             if lo is None:
                 return
 
-        if hi is not None and not isinstance(hi, str):
+        if hi is not None:
             hi = convert_limit(hi)
             if hi is None:
                 return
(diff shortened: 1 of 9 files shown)


hamogu:
I like this. Detailed look at code coming. I'm torn if ignoring before data is set should be a no-op. On the one hand, I find all these Errors on what I expect to be no-ops (e.g. ungrouping an ungroupd dataset) annoying because it requires a lot of if-then in analysis scripts and in jupyter notebooks where I might not execute things in a linear fashion. On the other hand, it's nice to remind users that ignoring before setting data won't do anything.

hamogu (approved):
I think this is good to go unless @DougBurke wants to make any further changes?

DougBurke:
I think I am done with making changes (offer valid on **this** PR only)

anetasie:
@DougBurke this looks good. I finally tried it.

One weird case to consider: 
The first data set has been filtered, for example `notice(3.,7.)` before the second data set was loaded and then filtered with `notice(4.,7.)`. The current behaviour is to not update the data set 1 with the new filter, and only apply the filter to the second data.  I think this is ok, but we may want to document this case.

```
load_data(1,'9774.pi')
show_data()
notice(3,7.)
show_data()
load_data(2, '9774.pi')
show_data(2)
notice(4,7.)
```

This is what I get with this PR
```
print(get_filter(1))                                                                                          
3.000300049782:7.000699996948

print(get_filter(2))                                                                                          
3.993100047112:7.000699996948
````

**Review by anetasie — APPROVED** [review]

**Comment by DougBurke** [comment]
@anetasie - note that the behavior you see is not to do with this PR - it's how Sherpa works:

- the first `notice` call sets the range for dataset 1 to be 3 to 7 keV as dataset 1 has no filter
- the second `notice` call sets the range for dataset 2 to be 4 to 7 keV as dataset 2 has no filter
- the second `notice` call knows that dataset 1 has had a filter applied to it and so it combines the new filter (4 to 7 keV) with the existing filter (3 to7 keV) which here happens to mean no change.

If you'd made the second call `notice(4, 8)` then you'd have seen the filter for dataset ` change (to the union of 3-7 keV and 4-8 keV, so 3-8 keV).

```
% sherpa
Sherpa 4.13.0

Python 3.8.2 (default, Mar 25 2020, 17:03:02)
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa

sherpa In [1]: load_pha('9774.pi')
read ARF file 9774.arf
read RMF file 9774.rmf
read background file 9774_bg.pi

sherpa In [2]: notice(3, 7)

sherpa In [3]: get_filter(1)
Out[3]: '3.000300049782:7.000699996948'

sherpa In [4]: load_pha(2, '9774.pi')
read ARF file 9774.arf
read RMF file 9774.rmf
read background file 9774_bg.pi

sherpa In [5]: notice(4, 7)

sherpa In [6]: get_filter(1)
Out[6]: '3.000300049782:7.000699996948'

sherpa In [7]: get_filter(2)
Out[7]: '3.993100047112:7.000699996948'

sherpa In [8]:
```

PR #1131, opened 2021-04-26
Test: fix a test failure when XSPEC installed but no fits backend

DougBurke:
# Summary

Avoid running a test we know to fail for an uncommon combination of options.

# Details

This just adds necessary requires_data and requires_fits decorators
to a test. It is expected that if you have XSPEC enabled then you
will also have FITS, so it is a rarely-tested combination, hence
the fact we occasionally find this issue.


DougBurke:
rebased as #1132 has been merged

DougBurke:
I'm asking for a review from @Marie-Terrell but if @harlantc wants to beat her to the punch I'm not going to complain :-)

Marie-Terrell (approved):
Looks good to me!

PR #1132, opened 2021-04-26
Update XQuartz for GH Actions workflow

Marie-Terrell:
xQuartz is now available on GitHub and the old version is no longer accessible via the old bintray link. This PR updates the xQuartz download location and version for the GitHub Actions Conda tests.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -83,16 +83,16 @@ jobs:
     - name: Install XQuartz on macOS
       if: runner.os == 'macOS'
       run: |
-        curl -LO https://dl.bintray.com/xquartz/downloads/XQuartz-2.7.11.dmg
-        hdiutil attach XQuartz-2.7.11.dmg
-        cd /Volumes/XQuartz-2.7.11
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-conda-workflow.yml
```diff
@@ -83,16 +83,16 @@ jobs:
...
+        cd /Volumes/XQuartz-2.8.1
+        sudo installer -pkg /Volumes/XQuartz-2.8.1/XQuartz.pkg -target /
         if [ $? != 0 ] ; then
             exit 1
         fi
         cd -
-        hdiutil detach /Volumes/XQuartz-2.7.11
-        rm XQuartz-2.7.11.dmg
+        hdiutil detach /Volumes/XQuartz-2.8.1
+        rm XQuartz-2.8.1.dmg
```
DougBurke, line 95: We could make the version string a variable, so that updating to the next version would require only changing one place, but I don't think it's worth making another commit just to do that (we could look at updating it in a later PR).

DougBurke (approved):
LGTM (assuming it passes the CI run ;-)

PR #1134, opened 2021-04-28
This is a doc-string only update for load_multi_XXX

### LINKED ISSUE #982: Doc for load_multi_arfs / rmfs confusing or even wrong
This was brought up by a Sherpa I communicated with:

The docstring for `load_multi_arfs` / `rmfs` seems to indicate that this function is useful for e.g. Chandra HETG observations.
The docstring says: ```A grating observation - such as a Chandra HETG data set - may require multiple responses. This function lets the multiple ARFs for such a data set be loaded with one command.```
https://github.com/sherpa/sherpa/blob/master/sherpa/astro/ui/utils.py#L5728

I read this as "If you have a PHA2 files from the HETG, then will load all your arfs at once". This impression is strengthened by an example that seems to load the  arfs for positive and negative orders into dataset 1 and 2:
```
>>> arfs = ['m1.arf', 'p1.arf']
>>> load_multi_arfs(arfs, [1,2])
```
Only after I read the code fir this function I realized that that's not what it's doing. Instead it's loading two arfs for the same dataset (here defailt_id=1 because it's not given.)

I have no idea when I would possibly use `load_multi_xxx`, but certainly not for HETG data. That comes as a phaII file, so what I need to do for HETG is something like:

```
for i, arf_file in zip([3, 4, 9, 10], arf_list):
    load_arf(i, arf_file)
```
or use a datastack.

This is an issue and not a PR because:

- The example is so explicit that I wonder it I'm too confused. Is what I wrote above right?
- I don't know what these functions might be useful for at all (I've never seen or used multiple arfs or rmfs for the same dataset), so I can't make a more meaningful example.

hamogu:
# Summary
Doc-string only change. No entry in release notes needed.

# Details
The previous example implied that a spectrum may contain both positive
and negative orders in the same spectrum. That can happen only after
coadding - and in that case one would co-add the arfs as well.

fixes #982

Since this is a doctring-only update, I'll take the opportunity to see
if I can avoid wasting electrons for CI.

[ci skip]

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -5722,11 +5722,15 @@ class Session(sherpa.ui.utils.Session):
     def load_multi_arfs(self, id, filenames, resp_ids=None):
         """Load multiple ARFs for a PHA data set.
 
-        A grating observation - such as a Chandra HETG data set - may
-        require multiple responses. This function lets the multiple
-        ARFs for such a data set be loaded with one command. The
-        `load_arf` function can instead be used to load them in
-        individually.
+        A grating observation - such as a Chandra LETGS data set - may
+        require multiple responses if the detector has insufficient energy
+        resolution to sort the photons into orders. In this case, the
+        extracted spectrum will contain the signal from more than one
+        diffraction orders.
+
+        This function lets the multiple ARFs for such a data set be
+        loaded with one command. The `load_arf` function can instead
+        be used to load them in individually.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


hamogu:
This is a very simple change. It can wait till after the 4.13.1 release or go in if approved - whatever is easier.

hamogu:
[ci skip] seems to work! Github action is skipped (which makes no sense for this PR), but RTDs still builds (which makes sense since I changed the docstring - also I don't know how to skip RTDs)

PR #1136, opened 2021-04-29
Replace error message "too few columns" with a more generic word

### LINKED ISSUE #244: Minor inconsistency in error message
Today  I saw this:

```
sherpa-57> save_arrays("stuff.tbl", [energies[:-1], energies[1:], flux[:-1]], ["keV","photons/s/cm**2/keV"], ascii=True)
IOErr: Received more column names: 2, than required: 3
```

Surely, the error message should should say that it received "FEWER" column names, not "MORE" since 2 < 3 for most definitions of numbers...

hamogu:
# Summary
Change the wording of an error message. Potentially breaking if someone tests for the wording of the error message in their code, but certainly more correct.

# Details
Change is required, because it could also be "too many" instead of "too few" columns

fixes #244

Can wait till after 4.13.1 is out.

### DIFF
--- a/sherpa/io.py
+++ b/sherpa/io.py
@@ -236,7 +236,7 @@ def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
     colkeys = list(colkeys)
 
     if len(names) > len(args):
-        raise IOErr('toomanycols')
+        raise IOErr('wrongcolnumber')
 
     assert(len(names) <= len(args))
 
--- a/sherpa/io.py
+++ b/sherpa/io.py
@@ -236,7 +236,7 @@ def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
     colkeys = list(colkeys)
 
     if len(names) > len(args):
-        raise IOErr('wrongcolnumber', len(args), len(names))
+        raise IOErr('wrongnumcols', len(args), len(names))
 
     assert(len(names) <= len(args))
 
--- a/sherpa/io.py
+++ b/sherpa/io.py
@@ -236,7 +236,7 @@ def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
     colkeys = list(colkeys)
 
     if len(names) > len(args):
-        raise IOErr('wrongcolnumber')
+        raise IOErr('wrongcolnumber', len(args), len(names))
 
     assert(len(names) <= len(args))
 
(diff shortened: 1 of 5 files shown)


sherpa/io.py
```diff
@@ -236,7 +236,7 @@ def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
     colkeys = list(colkeys)
 
     if len(names) > len(args):
-        raise IOErr('toomanycols')
+        raise IOErr('wrongcolnumber')
```
DougBurke, line 239: Yadda yadda this needs the arguments for the number of columns

sherpa/utils/err.py
```diff
@@ -279,7 +279,7 @@ class IOErr(SherpaErr):
             'badarray': "'%s' must be a Numpy array, list, or tuple",
             'notascii': "file '%s' does not appear to be ASCII",
             'noparamcols': 'No parameter columns found in %s',
-            'toomanycols': "Received more column names: %s, than required: %s",
+            'wrongcolnumber': "Received wrong number of column names: %s, required: %s",
```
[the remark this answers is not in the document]
hamogu: I'm just trying to keep the number of open issues manageable. That large number makes my brain hurt and makes it hard to find the issues I'm really looking for. Plus, it's a very easy way to feel I was at least a little bit productive.

hamogu:
i think I did all the requested changes. Can you review again @DougBurke ?

DougBurke (approved):
LGTM

sherpa/utils/err.py
```diff
@@ -279,7 +279,7 @@ class IOErr(SherpaErr):
             'badarray': "'%s' must be a Numpy array, list, or tuple",
             'notascii': "file '%s' does not appear to be ASCII",
             'noparamcols': 'No parameter columns found in %s',
-            'toomanycols': "Received more column names: %s, than required: %s",
+            'wrongcolnumber': "Expected %d columns but found %d",
```
wmclaugh, line 282: @hamogu: a very trivial point but the connotation I get from the mnemonic 'wrongcolnumber' sounds like an indexing error ( 'ie. indexed column 0 instead if column 1'). Perhaps something like 'wrongnumcols' would improve clarity.

hamogu: Done. (Just pushed.)

wmclaugh (approved):

hamogu:
Test failure is just in the codecov because most of the exceptions are not covered by tests - but they were not before, so I don't feel it's necessary to add tests for all of them right now.

DougBurke:
I agree that the coverage "failure" here is not an issue.

PR #1137, opened 2021-04-30
Correct docstring for fake_pha

hamogu:
The old docstring claims that the backgound is just scaled and not
simulated, but there is definitely a call to `poisson_noise` for the
background in the code.

[skip ci]

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -8677,9 +8677,10 @@ class Session(sherpa.ui.utils.Session):
         expresion is evaluated for each channel to create the expectation
         values, which is then passed to a Poisson random number generator
         to determine the observed number of counts per channel. Any
-        background component is scaled by appropriate terms (exsposure
-        time, area scaling, and the backscal value) before adding to the
-        simulated date. That is, the background component is not simulated.
+        background component is scaled by appropriate terms (exposure
+        time, area scaling, and the backscal value) before is is passed to
+        a Poisson random number generator. The simulated background is
+        added to the simulated data.
 
         Examples
         --------
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -8678,7 +8678,7 @@ class Session(sherpa.ui.utils.Session):
         values, which is then passed to a Poisson random number generator
         to determine the observed number of counts per channel. Any
         background component is scaled by appropriate terms (exposure
-        time, area scaling, and the backscal value) before is is passed to
+        time, area scaling, and the backscal value) before it is passed to
         a Poisson random number generator. The simulated background is
         added to the simulated data.
 


DougBurke:
Hmmm. Now you've made me look at the code.

We have

```
            nbkg = len(d.background_ids)
            b = 0
            for bkg_id in d.background_ids:
                # we do (probably) want to filter and group the scale array
                b += d.get_background_scale(bkg_id) * \
                    d.get_background(bkg_id).counts

            if nbkg > 0:
                b = b / nbkg
                b_poisson = sherpa.utils.poisson_noise(b)
                d.counts = d.counts + b_poisson
```

Let's ignore the check on `len(d.background_ids)` as I think it should be > 0 here by construction but can't guarantee that.

This code evaluates the background contribution from all the components, scaling to match the observation, and then calls `poisson_noise` on that. Conceptually I think we should have

```
            nbkg = len(d.background_ids)
            b = 0
            for bkg_id in d.background_ids:
                b += d.get_background_scale(bkg_id) * \
                    sherpa.utils.poisson_counts(d.get_background(bkg_id).counts)

            if nbkg > 0:
                b = b / nbkg
                d.counts = d.counts + b / nbkg
```

but I don't know if operationally it makes a difference (my gut says yes, but my gut is not a statistician).

hamogu:
I also want to be able to call this function with multiple ARFs and RMFs (e.g. the simulate a LETG/HRC-S observation). I think we should allow that, too. I have a propotype locally, but did not open a PR yet.

To the second question: The sum Poisson distributions is again a Poisson distribution. I think it's OK (even if not very clear) as it's now.

DougBurke:
I was assuming I was too lazy to do a quick google search to show that P(a) + P(b) ~ P(a + b) for Poissons...

DougBurke:
So, we add the background dataset and then use a Poisson realisation of that to add to the source data. This means that

    plot_data()

will contain the random background component but

    plot_bkg()

will show the input data (that is, the data passed to the `bkg` parameter). This is fine, but unfortunately in the documentation when we talk about "background" I **don't think** we are clear about "background component that is being added to the source region" and "background component that represents the background". However, I may be over-thinking this.

hamogu:
Mmm. Good point. Maybe it's not just a doc change, but also requires a change in code now that we look at it. Always dangerous to look at code that you believe to be working - never know what lurks around the corner.

DougBurke:
@hamogu - I don't think there's anything wrong with the code

 - we create a random realization of the background to add to the source aperture
 - we just copy over the background and make it the background you can subtract/fit/whatever with the result

I guess you could randomize the background in the second step but that is (possibly) technically annoying and I'm not sure it's meaningful.

All I was saying was I'm not sure we properly describe what the code does...

Oh, and as you are working on this, I'd ideally want to move the logic out of sherpa.asto.ui.utils and into <someplace>, so that `fake_pha` basically just has to

    input = self.get_data(id) if id in self._data else None
    output = some.module.fake_pha(input, ...)
    self.set_data(id, output)

with the <someplace> version just working with data classes, not the UI layer [although there may be more logic needed]. Then we can use in places where we don't care about the UI layer...

DougBurke (approved):
Let's take this doc change and then we can worry about other possible documentation changes when you add your multi-response support.

anetasie:
@DougBurke are you going to create a separate issue for the details?

sherpa/astro/ui/utils.py
```diff
@@ -8677,9 +8677,10 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
         expresion is evaluated for each channel to create the expectation
         values, which is then passed to a Poisson random number generator
         to determine the observed number of counts per channel. Any
-        background component is scaled by appropriate terms (exsposure
-        time, area scaling, and the backscal value) before adding to the
-        simulated date. That is, the background component is not simulated.
+        background component is scaled by appropriate terms (exposure
+        time, area scaling, and the backscal value) before is is passed to
```
wmclaugh, line 8681: "is is" -> "it is"?

hamogu: fixed.

PR #1138, opened 2021-05-05
Address NumPy 1.20 bool/int/float deprecation warnings

DougBurke:
# Summary

NumPy 1.20 notes that numpy.bool, numpy.int, and numpy.float will be removed and users should just use bool, int, and float instead, so these symbols have been changed. This is a follow on to #1092

# Details

This change should future-proof ourselves against the upcoming numpy release that removes these deprecated symbols (I haven't loked to see if they've announced when that would be).

Thanks to sherpa/fit.py importing the symbols directly this was a little-bit more complex than a simple search and replacement (i.e. numpy.x/np.int -> int). Our code could probably do with a review of when a "generic" float vs a specific type (float32 or float64) is useful, but I think the default is to assume float64-like values.

I am not sure whether we have any CI tests with NumPy 1.20 (but I assume so otherwise why would we have needed #1092 ?).

Note that #1143 and #1144 both came out of this, but they are un-related.

### DIFF
--- a/sherpa/astro/tests/test_astro_data.py
+++ b/sherpa/astro/tests/test_astro_data.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2007, 2015, 2017, 2018, 2020, 2021
-#        Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -70,7 +70,7 @@ class test_filter_energy_grid(SherpaTestCase):
         2.71560001e+00, 2.86159992e+00, 3.08060002e+00, 3.38720012e+00,
         3.56240010e+00, 3.79600000e+00, 4.02960014e+00, 4.24860001e+00,
         4.71579981e+00, 5.02239990e+00, 5.37279987e+00, 5.89839983e+00,
-        6.57000017e+00, 9.86960030e+00], np.float)
+        6.57000017e+00, 9.86960030e+00], float)
 
     _emax = np.array([
         0.2482, 0.3066, 0.46720001, 0.56940001, 0.64240003,
@@ -82,7 +82,7 @@ class test_filter_energy_grid(SherpaTestCase):
         2.58419991, 2.71560001, 2.86159992, 3.08060002, 3.38720012,
         3.5624001, 3.796, 4.02960014, 4.24860001, 4.71579981,
         5.0223999, 5.37279987, 5.89839983, 6.57000017, 9.8696003,
-        14.95040035], np.float)
+        14.95040035], float)
 
     def setUp(self):
         self.old_level = logger.getEffectiveLevel()
@@ -158,7 +158,7 @@ class test_filter_energy_grid_reversed(SherpaTestCase):
         0.37154216, 0.36742872, 0.3641032, 0.36167556, 0.35983625,
         0.35634032, 0.35248783, 0.35085678, 0.34843227, 0.34669766,
         0.34418666, 0.33912122, 0.33720407, 0.33505177, 0.33279634,
-        0.33081138, 0.32847831, 0.32592943, 0.3111549], np.float)
+        0.33081138, 0.32847831, 0.32592943, 0.3111549], float)
 
     _emax = np.array([
         3.06803656, 2.39196181, 2.35973215, 2.34076023, 2.30973101,
@@ -201,7 +201,7 @@ class test_filter_energy_grid_reversed(SherpaTestCase):
         0.37347662, 0.37154216, 0.36742872, 0.3641032, 0.36167556,
         0.35983625, 0.35634032, 0.35248783, 0.35085678, 0.34843227,
         0.34669766, 0.34418666, 0.33912122, 0.33720407, 0.33505177,
-        0.33279634, 0.33081138, 0.32847831, 0.32592943], np.float)
+        0.33279634, 0.33081138, 0.32847831, 0.32592943], float)
 
     def setUp(self):
         self.pha = DataPHA('', np.arange(204, dtype=float) + 1.,
@@ -1030,7 +1030,7 @@ def test_pha_mask_filtered(infile, size, nset, make_data_path):
     pha.notice(0.5, 7)
     pha.ignore(2, 3)
 
-    assert pha.mask.dtype == np.bool
+    assert pha.mask.dtype == bool
     assert pha.mask.size == size
     assert pha.mask.sum() == nset
 
@@ -1279,11 +1279,11 @@ def test_grouping_filtering_binning(analysis, make_data_path):
     assert (pha.grouping == expected).all()
 
     # This is based on the energy results
-    expected = np.zeros(21, dtype=np.bool)
+    expected = np.zeros(21, dtype=bool)
     expected[1:9] = True
     assert (pha.mask == expected).all()
 
-    expected = np.zeros(1024, dtype=np.bool)
+    expected = np.zeros(1024, dtype=bool)
     expected[50:450] = True
     assert (pha.get_mask() == expected).all()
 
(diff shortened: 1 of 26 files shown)


DougBurke:
I've suggested @hamogu review this but that was just ot get the ball rolling. Anyone who wants to is welcome / encouraged to!

sherpa/astro/tests/test_astro_data.py
```diff
@@ -70,7 +70,7 @@ class test_filter_energy_grid(SherpaTestCase):
         2.71560001e+00, 2.86159992e+00, 3.08060002e+00, 3.38720012e+00,
         3.56240010e+00, 3.79600000e+00, 4.02960014e+00, 4.24860001e+00,
         4.71579981e+00, 5.02239990e+00, 5.37279987e+00, 5.89839983e+00,
-        6.57000017e+00, 9.86960030e+00], np.float)
+        6.57000017e+00, 9.86960030e+00], float)
```
hamogu, line 73: This is fine, but there is no need to even have the "float" here. `np.array` defaults to float anyway.

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2007, 2015, 2018, 2019, 2020, 2021  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2015, 2018, 2019, 2020, 2021
```
hamogu, line 2: At some point, we should really make a post-commit hook that takes care of this stupid updating of year automatically!

sherpa/models/basic.py
```diff
@@ -1962,7 +1962,7 @@ def calc(self, p, xlo, xhi=None, **kwargs):
 class Integrate1D(RegriddableModel1D):
 
     def __init__(self, name='integrate1d'):
-        tol = numpy.finfo(float).eps
+        tol = numpy.finfo(float).eps  # should this be DBL_EPSILON?
```
hamogu, line 1965: I think this this fine. The precision of "float" depends on the machine it's run on. For Mac and Linux, it's going to be a 64 bit number, and `np.finfo(float).eps=2.220446049250313e-16`. However, if run on other architectures, it might be different. Unless I miss something, I suggest reverting the comment because that just invites the next reader to wonder why it's there.

sherpa/models/parameter.py
```diff
@@ -41,8 +42,8 @@
 # hugeval = 1.0e+38
 #
 # Use FLT_TINY and FLT_MAX
-tinyval = numpy.float(numpy.finfo(numpy.float32).tiny)
-hugeval = numpy.float(numpy.finfo(numpy.float32).max)
+tinyval = float(numpy.finfo(numpy.float32).tiny)
```
hamogu, line 45: I don't think there is a need for explicit conversion here. `numpy.finfo(numpy.float32).tiny` is already a float (a `flaot32` to be precise and numpy will automatically typecast that (keeping the value though!) when used in any of the following calculations. It's not like `numpy.finfo(numpy.float32).tiny` itself will suddenly become larger or smaller through that typecasting.
```suggestion
tinyval = numpy.finfo(numpy.float32).tiny
hugeval = numpy.finfo(numpy.float32).max
```

sherpa/optmethods/optfcts.py
```diff
@@ -1153,7 +1153,7 @@ class fdJac:
         def __init__(self, func, fvec, pars):
             self.func = func
             self.fvec = fvec
-            epsmch = numpy.float_(numpy.finfo(numpy.float).eps)
+            epsmch = numpy.float_(numpy.finfo(float).eps)
```
hamogu, line 1156: Again, I don't see what this is explicitly typecast at all.
```suggestion
            epsmch = numpy.finfo(float).eps
```

sherpa/fit.py
```diff
@@ -722,7 +724,7 @@ def primini(self, statfunc, pars, parmins, parmaxes, statargs=(),
         # Keep record of current and previous statistics;
         # when these are within some tolerace, Primini's method
         # is done.
-        previous_stat = float32(finfo(float32).max)
+        previous_stat = np.float32(np.finfo(np.float32).max)
```
hamogu, line 727: No need to typecast that at all. `np.finfo(np.float32).max` is already a `np.float32`.
```suggestion
        previous_stat = np.finfo(np.float32).max
```

hamogu (approved):
I don't see anything with a great risk of doing harm here. You never know, but nothing here looks suspicious to me.
There are a few cases where there is extra typecasting done (e.g. casting a float32 to a float32), which is unnecessary and makes the code a little harder to read, but the impact on runtime will be absolutely negligible. Could be address in here, or could just stay as it is.

sherpa/models/basic.py
DougBurke: What I meant was that there is a module-level constant called DBL_EPSILON with the (if I remember correctly) same definition, so should this just say `tol = DBL_EPSILON`?

sherpa/models/parameter.py
DougBurke: I agree that this looks like unnecessary code here.

sherpa/models/basic.py
hamogu: I think it's the same for us, because we run on a system where the `float` equals `float64`. But on other architectures, `float` might actually equal `float32`. Not sure we'll ever make sherpa run on an old rasberry py or whatever, so it probably does not matter in practice, but in principle, the code as written gives `tol` the value that's appropriate for the particular system it's running on.

DougBurke:
Given that we are not going to rush this PR I'll make (some of?) the additional changes @hamogu noted above

sherpa/models/parameter.py
DougBurke: Pedantry awaits!

Is it guaranteed that Python's `float` type is the same as NumPy's `float32` type? I know that there are "Python will automatically type-convert the values when necessary`, but there are times when it can be useful when moving between Python/C/FORTRAN to be able to say "I really want this type here, not something that will magically contort itself to what we want"

DougBurke: actually, I guess the issue is that we are - for the platforms we are interested in - almost guaranteed that `float` and `numpy.float32` have different sizes anyway

```
In [7]: type(numpy.finfo(numpy.float32).eps)
Out[7]: numpy.float32

In [8]: type(float(numpy.finfo(numpy.float32).eps))
Out[8]: float
```

Do we gain or lose anything with conversions here (and my previous comment was actually "the wrong way around" in thinknig about the type of the value storing the data)?

hamogu: Yes, they are different. An explicit conversion does not do any harm, but it's also not necessary. Python and numpy will take care of that automatically if necessary - one of the points of using a high-level dynamically types language as opposed to C (where you would definitely have to know what dtype with variable has) is that you save the developer brainpower needed to think about it. 
And the time used for those automatic conversions is essentially unmeasurably short, so this is purely about saving developer time by not discussing if a conversion is needed or not thinking about it when reading the code; leaving this in or taking it out will have no effect whatsoever on measured runtime or numeric results.

DougBurke:
Note to self - the commit previous to this comment is going to go away - it is best done in a separate PR as it's not related to these numpy changes (it was intended to improve the coverage of a change that was made in this PR but that snowballed quicker than I expected ...).

hamogu (approved):

sherpa/fit.py
```diff
@@ -24,8 +24,10 @@
 
 from functools import wraps
 
+import numpy as np
```
wmclaugh, line 27: Is there a reason why you only moved some of the numpy imports (if. float32, finfo) into the np namespace but left the other imports from numpy in place?  Would it be better to be consistent?

DougBurke: I'm not aware we have a "standard" for our imports, so I felt it was better to leave as is rather than make wholesale changes. I do occasionally try to upgrade to the 'import numpy as np' style to save those precious three characters, but in this particular case I felt it was risky to go through and find all the "naked" symbols (e.g. `arange` and `abs`) and convert them to their "namespaced version" (e.g. `np.arange`, `np.abs`, ...). I think it would be a good thing, but in a separate PR to be done later.

PR #1139, opened 2021-05-06
Fix typo in Zenodo test that made it fail

DougBurke:
# Summary

Fix one of the zenodo tests, which had a typo in the test.

# Details

This test is only run when the --runzenodo flag is given, which we do not use in CI. It is an embarrassing bug, as I appear to still
be unable to type retrieve correctly.


hamogu:
I just saw that and thought "Oh, I'll make a PR later" and then I forgot where exactly I had seen it.

hamogu (approved):

PR #1140, opened 2021-05-07
Fix some remaining master/main renames

hamogu:
# Summary
replace some master -> main strings

# Details
Mostly, I do this to correct the docs/install.rst page, but while
I was at it I replaced a few more, mostly in the comments.
And then thee is an unrelated renaming `master_doc` -> `main_doc`
for a variable in sphinx's `conf.py` which I renames just for good
measure and because it was easier to change thme all then to think
too hard if that one should stay.

[skip ci]
(rtd build will run nevertheless and there are no code changes except
for sphinx)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -273,8 +273,8 @@ source_suffix = '.rst'
 #
 # source_encoding = 'utf-8-sig'
 
-# The master toctree document.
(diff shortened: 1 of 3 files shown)


DougBurke (approved):
:shipit:

PR #1143, opened 2021-05-10
add basic test and documentation for Integrate1D

DougBurke:
# Summary

Add basic testing of the integrate1d model class and improve its documentation.

# Details

I noted that the Integrate1D class had no test in #1138 so this adds a test or two.

I have added some basic documentatino for the `sherpa.models.basic.Integrate1D` class (at least as far as I understand it) and for one of the underlying python-accessible C routines (this is not expected to be used by anyone outside of Sherpa but it's useful to have the optional arguments discussed).

There are documentation improvements I'd like to make (e.g. discussing the `integrate` flag and coming up with a consistent naming scheme for point versus grid/histogram analysis) but this is outside the scope of this PR.

### DIFF
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1954,10 +1954,10 @@ class Integrator1D(CompositeModel, RegriddableModel1D):
     --------
 
     Numericaly integrate the Gauss1D model across the bins defined by
-    xlo and xhi, using an absolute tolerance of 1e-5. Note that the
-    Gauss1D model is used as an example here, since it can be used
-    directly to evaluate the integrated signal, as shown in the last
-    line (creating the ytrue value).
+    xlo and xhi, using an absolute tolerance of 1e-5. Note that you
+    would not use this in practice, since the Gauss1D model already
+    supports integrated bins, but it does allow us to compare the two
+    approaches (the `y` and `ytrue` arrays).
 
     >>> from sherpa.models.basic import Gauss1D, Integrator1D
     >>> gmdl = Gauss1D()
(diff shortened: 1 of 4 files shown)


DougBurke:
Note there is a small amount of information in the old ahelp page - https://cxc.harvard.edu/sherpa4.10/ahelp/integrate1d.html

sherpa/models/basic.py
```diff
@@ -1959,7 +1999,76 @@ def calc(self, p, xlo, xhi=None, **kwargs):
...
+    If `imdl` is an instance of `Integrate1D` and `omdl` the model to
+    integrate, then `imdl(omdl)` creates the integrated form. Note
+    that changes to the `Integrate1D` parameters - such as `epsabs` -
+    must be made before creating this integrated form.
+
+    Examples
+    --------
+
+    The Gauss1D is used as an example here since it - as can most Sherpa
+    models - already be used with an histogram dataset. The evaluation
```
hamogu, line 2036: So the point of using this is to compare the build-in with the "by-hand" integration? Maybe add "in the example we compare the integration build into Gauss1D with the integration done by Integrate1D, but in practice, one would use Integrate1D only on models that do not define integration over bins themselves." just to really drive this point home.

hamogu: i had to think about it, so other users of Sherpa might have to, too.

hamogu (approved):
A few suggestion for expanding the docstring even more. I tried to word it such that you can copy and paste to save you extra effort, because this is already much better than before (no docstring).

sherpa/models/basic.py
```diff
@@ -1926,6 +1926,46 @@ def __init__(self, name='usermodel', pars=None):
 
 
 class Integrator1D(CompositeModel, RegriddableModel1D):
+    """Integrate a one-dimensional model.
+
+    It is expected that instances of this class are created by the
+    `Integrate1D` class and not directly.
```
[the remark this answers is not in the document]
DougBurke: I'm not sold on the example - we could remove that. However, I'm not completely sure about the "do not use this manually" comment - it's something I assume to be true: as this class isn't exported by default that does seem to fit in to the idea of this being an internal class.

sherpa/models/basic.py
DougBurke: I wanted an easy-to-write example, but most of our models already support integration, so this was the best I could come up with on short notice. Then I ran out of energy to explain my thinking ;-)

sherpa/models/basic.py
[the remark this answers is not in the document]
hamogu: You could make it and ABS and see if anything breaks. If it does, we have plenty of time to revert that particular line before the next release.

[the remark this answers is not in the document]
DougBurke: I'm reluctant to make changes here as I don't think it's a much-used piece of code (that might mean we should ensure it's rock solid, but I'm taking it to mean let's focus on other things!).

Is there somewhere I can read on ABC to enforce the requirement (last time I tried to read up on the abstract base class work I found it a bit too abstract for my use case...).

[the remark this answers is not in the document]
hamogu: Let's focus on other things.

sherpa/models/basic.py
DougBurke: I've reworked this example and it's hopefully better now.

hamogu: I have already approved it, so I can't approve it again. Looks good to me.

sherpa/models/src/_modelfcts.cc
```diff
@@ -57,7 +58,32 @@ static PyMethodDef ModelFcts[] = {
...
+		    "    The bin edges, which must match in size and have xhi_i > xlo_i\n"
+		    "errflag : int, optional\n"
+		    "    Currently unused.\n"
+		    "epsabs : number, optional\n"
+		    "    The maximum absolute difference. The default is the 64-bit\n"
+		    "    float epsilon.\n"
+		    "epsrel : number, optional\n"
+		    "    The maximum relative difference. The default is 0.0.\n"
+		    "maxeval : int, optional\n"
+		    "    The maximum bumber of evaluations. The default is 10000.\n"
```
wmclaugh, line 80: should this (bumber) be 'number'?

dtnguyen2:
10,000 is kind of high for the max number of evaluations isn't it?

DougBurke:
@dtnguyen2 - all I'm doing is documenting the current behavior! I didn't write this code so I don't know how it behaves. I also note that it seems to complain even with simple models (e,g, f(x) = constant) when the absolute tolerance is the float64 eps value so we should either increase the default tolerance or "do something else"™ here (or, rather, put it on our list of things to worry about tomorrow).

sherpa/models/src/_modelfcts.cc
DougBurke: Fixed.

PR #1144, opened 2021-05-10
A minor cleanup of estmethods 

DougBurke:
# Summary

A minor cleanup of the sherpa.estmethods code.

# Details

In writing #1138 I noticed a couple of issues with the code in sherpa/estmethods. This is not a complete scrub through of the module, but includes

- moving tests from `SherpaTestCase` to `pytest`
- removal of code talking about supporting Numpy 0.98 as we do not support such an old release
- a cleanup to `projection` to only define arrays when we need to (when `parallel` is `False`) and to avoid building up a list to then convert to NumPy when we know the size needed for the NumPy array 
- improve coverage of `projection` by testing out both the `parallel=True` and `False` branches (at least for those machines which have multiple cores and so can run the parallel code).

### DIFF
--- a/sherpa/estmethods/__init__.py
+++ b/sherpa/estmethods/__init__.py
@@ -351,11 +351,10 @@ def covariance(pars, parmins, parmaxes, parhardmins, parhardmaxes, sigma, eps,
     # simpler inv function for inverting matrices does not appear to
     # have the same issue.
 
-    invfunc = numpy.linalg.inv
     inv_info = None
 
     try:
-        inv_info = invfunc(info)
+        inv_info = numpy.linalg.inv(info)
 
     except numpy.linalg.linalg.LinAlgError:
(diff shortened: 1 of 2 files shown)


hamogu (approved):

DougBurke:
Oh joy. Another rebase needed.

hamogu:
That's why it's great to stay < 50 open PRs or so... At some point collisions can't be avoided. You are just too active! But, seriously, I much prefer a larger number of small PRs over one big monolithic one because it's so much easier to review.

DougBurke:
Unfortunately there are some cases (cough cough #1161) that end up being larger than originally anticipated. Especially if people come along and start adding their own PRs to change highly-related code ;-)

The trick is to be able to pull put what we want into easily digestible changes and still be able to review the larger commit on a reasonable timescale.

Issue #1152, opened 2021-05-21
coverage and the GH Actions


DougBurke:
From a recent CI run we can see https://github.com/sherpa/sherpa/pull/1151/checks?check_run_id=2634818447#step:6:953 which is the test step og the conda "Linux Minimum Setup" run. What is of note here are the lines saying

    Error running `['git', 'log', '-1', '--pretty=%B']`: returncode=128, output=b''

and other variants. In fact, if you look earlier in this run at https://github.com/sherpa/sherpa/pull/1151/checks?check_run_id=2634818447#step:6:380 you can see some related messages, in this case

    fatal: not a git repository (or any of the parent directories): .git

I believe this comes from the following logic in `test.sh`: https://github.com/sherpa/sherpa/blob/dde8c22647e8e4f098bf344b7640b17a41a5e265/.github/scripts/test.sh#L45-L52 which runs the tests outside of the sherpa git repository, and so has no access to the git repository (this was added by me as part of #1001).

I wonder if this could be the cause of the coverage notes we see about being unable to identify the correct commit (it would make sense).

Options include

 - avoid the coverage check in this section (which technically means we loose coverage of some edge cases in the testing - such as fall-through code when we don't have the fits or plotting backends) but this is not terrible
 - somehow find out how to "add in" the commit history in the parts that need it (but this would require understanding how the coverage code works and I think is outside our scope/remit)

DougBurke:
I'm closing this because I think/hope that #1339 has "fixed" this

PR #1151, opened 2021-05-20
Fix crash from grouping data using an unsupported method

DougBurke:
# Sumary

Improve error checking and documentation of the low-level grouping code. This requires C++11 support in the compiler.

# Details

The following requires the C++11 standard, but there is no
change in documentation to record this, or any attempt to
ensure this restriction in the setup.

I had not realised that the grouping support was limited to a
hard-coded set of routines, and found this out by crashing Python.
This fixes this, by explicitly checking that the group function
is supported and if it isn't raising a ValueError. Documentation
for do_group is added.

A test of this is added, along with a few basic tests of the
low-level do_group function.

We could use an error from the sherpa.utils.err module but I don't think it's worth it (and more complicated to do given the error is raised from C scope).

# Example

With this commit we have the following - note the `ValueError` that is raised.

```
In [8]: pha.apply_grouping(pha.counts)
Out[8]:
array([17., 15., 16., 15., 16., 15., 18., 18., 15., 18., 15., 15., 19.,
       15., 15., 17., 16., 16., 17., 15., 19., 15., 16., 15., 16., 17.,
       15., 18., 16., 15., 15., 16., 15., 15., 15., 16., 16., 15., 15.,
       16., 16., 15., 16., 15., 15., 20.])

In [9]: pha.apply_grouping(pha.counts, sum)
Out[9]:
array([17., 15., 16., 15., 16., 15., 18., 18., 15., 18., 15., 15., 19.,
       15., 15., 17., 16., 16., 17., 15., 19., 15., 16., 15., 16., 17.,
       15., 18., 16., 15., 15., 16., 15., 15., 15., 16., 16., 15., 15.,
       16., 16., 15., 16., 15., 15., 20.])

In [10]: pha.apply_grouping(pha.counts, lambda x: sum(x))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-39d632c1d5f9> in <module>
----> 1 pha.apply_grouping(pha.counts, lambda x: sum(x))

~/sherpa/sherpa-master/sherpa/astro/data.py in apply_grouping(self, data, groupfunc)
   2350         filter = self.quality_filter
   2351         if filter is None:
-> 2352             return do_group(data, groups, groupfunc.__name__)
   2353
   2354         if len(data) != len(filter) or len(groups) != len(filter):

ValueError: unsupported group function: <lambda>

```

Without this PR the last command crashes Python:

```
[...]

PR #1153, opened 2021-05-21
CI: remove coverage reporting for sherpa_test

DougBurke:
# Summary

Remove a build from our coverage reports as it is done outside of the git repository (issue #1152).

# Details

Remove an addition added in #1001. The problem is that sherpa_test
is explicitly run outside the Sherpa git repository, and so there
is no wat to identify the commit for the coverage code, causing
problems both in running the coverage and then when sending the
coverage data to our coverage system. It is not-at-all obvious
how to work around this, so I'm just throwing in the towl.

This does mean we lose some coverage of less-used code - i.e.
fall through cases when the fits or plotting backends are not
available - but so be it.


DougBurke:
So the codecov report no-longer complains about things like

```
Current head 35f9c00 differs from pull request most recent head 7288b5a. Consider uploading reports for the commit 7288b5a to get more accurate results
```

but I honestly don't know what causes this - e.g. https://github.com/sherpa/sherpa/pull/1151 which is currently still open had this error but a PR which has just been merged such as https://github.com/sherpa/sherpa/pull/1143 didn't complain.

DougBurke:
I give up. I'm now seeing the message

```
exclamation Current head 90d9abd differs from pull request most recent head fd3eadd. Consider uploading reports for the commit fd3eadd to get more accurate results
```

(this is from the first commit as the second commit isn't running tests). So I just don't understand. I wonder if our history got confused at one point and that is causing problems now?

DougBurke:
I'm closing this as I think the work @harlantc did on #1339 means this isn't needed anymore

Issue #555, opened 2018-11-17
Should we have the changelog as part of the Sphinx documentation?


DougBurke:
AstroPy (and others) have a changelog in the documentation, which I find very useful (we do this for CIAO after all ;-).

There's a summary page

http://docs.astropy.org/en/stable/whatsnew/3.0.html

and a detailed page

http://docs.astropy.org/en/stable/changelog.html#changelog

Can we do this for the Sherpa Sphinx documentation?

hamogu:
+1 - I find it useful. It does require discipline (or a bot to check) to make sure every PR adds to the changelog and in astropy it requires a lot of rebases, because often PRs add to the same section of the changelog.

hamogu:
Update: Astropy changed to towncrier for their changelog system: https://pypi.org/project/towncrier/
(or see the astropy developer docs: https://docs.astropy.org/en/latest/development/workflow/maintainer_workflow.html#updating-and-maintaining-the-changelog). That add one step to the release procedure (run towncrier to turn the news fragments files into the sorted readme), but no rebases are ever needed now for changes to the changelog.

hamogu:
@DougBurke @Marie-Terrell : Food for thought post release.

Adding some more details since this issue is so old, that the original reference to astropy does not refer to the same CHANGELOG system that is used to refer to. 

I have used two different systems for changelogs in the past, that write the main changelog as part of each PR and allow for automatic corss-linking from the changelog to the relevant GH PR, so that it's easy to go back and read the details discussion, see the code diff on the GH web interface etc.
Both approaches have been used by astropy at some point. I'm going to summarize pros and cons now. There certainly are other ways to deal with that.

[...]

PR #1155, opened 2021-05-23
Remove Meta class for fits hdu headers

hamogu:
# Summary
Remove Meta class for header information that is used only rudimentary and acts almost, but not quite, like a `dict`.

# Details
The Meta class for meta information acted almost like a dict, but did not implement the full collections API. In fact, in several places where it was used it was converted to a dict first or a dict was assigned to the header instead of the meta class - apparently without breaking anything else downstream. 
Thus, I'm making it a dict now, saving us the extra implementation. This ain't broken, so why fix it? @DougBurke touched this in #1006 and this is an easier (as in "it removes more code") approach and closes #1006 if accepted. The one change this will bring for users is that the Meta class formatted itself like we are used from fits (with 'KEY       = VAL') while now it's going to be displayed like a normal Python dict on the command line. I believe that this is a good thing. Only people who dig into the Python objects will ever get the header information as an object that they can print with "str(data.header)", and users who do that know what a dict is and how it looks. If there is any need we could keep that formatting around as a utility function, but it does not seem to be used either.

However, there maybe places I have missed (e.g. in "ui.show_data" or something like that) and that are not covered by tests, I'll check a few "ui.show_X" commands manually, if this passes CI. I've locally run tests with the pyfits backend, but want to see the full set of CI and the easiest way to do that is to push a PR.


sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -431,17 +431,12 @@ def test_save_data_data1d_no_clobber(tmp_path):
...
-    from sherpa.astro.io.meta import Meta
-
     ui.load_arrays(1, [1, 2, 3], [5, 4, 3], ui.DataPHA)
 
     # The code requires DataPHA to have a "valid" header so
     # fake one. Ideally we would not require it.
     #
-    hdr = Meta()
-    ui.get_data().header = hdr
+    ui.get_data().header = {}
```
DougBurke, line 439: I wonder if we should just make `DataPHA` set the `header` attribute duriing initialization?

hamogu: I think that's a good idea! Before I put that in though, I want to make all the current tests pass. There are some real failures that I need to understand because it indicates that there is some subtle difference between the dict and the old implementation. I suspect that it's something that I can fix, but I need to understand what's going on.

hamogu:
OK, I think this is ready for full review.

sherpa/astro/instrument.py
```diff
@@ -1279,7 +1279,7 @@ def create_arf(elo, ehi, specresp=None, exposure=None, ethresh=None, name='user-
 
 def create_delta_rmf(rmflo, rmfhi, offset=1,
                      e_min=None, e_max=None, ethresh=None,
-                     name='delta-rmf'):
+                     name='delta-rmf', header=None):
```
DougBurke, line 1282: If we do this then we should also do this for `create_non_delta_rmf` and `create_arf`.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -686,7 +682,7 @@ def get_rmf_data(arg, make_copy=False):
             data['matrix'] = hdu.data.field('MATRIX')
 
         data['header'] = _get_meta_data(hdu)
-        data['header'].pop('DETCHANS')
+        data['header'].pop('DETCHANS', None)
```
DougBurke, line 685: Obviously some of these "change to pop(foo, None)" are good as it simplifies the code (removes the try/except handing) but in some cases like this I think the file is invalid if it doesn't contain the keyword, so we technically don't need the safety check.

In this case a few lines earlier we have

        data['detchans'] = SherpaUInt(_require_key(hdu, 'DETCHANS'))

and I would hope there's an equivalent in the crates version but can not guarantee that as that could be deep in the crates code which we don't get to see.

I'm not requesting a change here, just bringnig it up for a small discussion.

sherpa/astro/tests/test_astro_data_notebook.py
```diff
@@ -67,7 +62,7 @@ def check(r, summary, name, label, nmeta):
     assert '<div class="dataval">{}</div>'.format(name) in r
 
 
-@pytest.mark.parametrize('header', [None, TEST_HEADER])
+@pytest.mark.parametrize('header', [None, {}, TEST_HEADER])
```
DougBurke, line 65: We should be able to drop the None case here (unless we think there is code out in sherpa that does end up setting header to None)?

sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -431,20 +431,8 @@ def test_save_data_data1d_no_clobber(tmp_path):
...
-
     ui.load_arrays(1, [1, 2, 3], [5, 4, 3], ui.DataPHA)
-
-    # The code requires DataPHA to have a "valid" header so
-    # fake one. Ideally we would not require it.
-    #
-    hdr = Meta()
-    ui.get_data().header = hdr
-
     out = tmp_path / "data.dat"
```
DougBurke, line 435: dang it - it's not easy to be able to cheer on removed lines from this interface!

sherpa/astro/io/pyfits_backend.py
hamogu: I usually try to flexible when reading and strict when writing. For Sherpa, there is no harm done if, for whatever reason, header has not DETCHANS keyword at this point, since we are not using the popped value anyway, so we might as well accept a situation where it's not there. Plus, this is just replicating the old behavior as the old Meta always returned `None` if a keyword was not found. 

In some instances (I'm not sure know if this was one of them) there are code path which pop a keyword earlier and tests actually fail with a "KeyError" in one of the those pop commands, where it's a totally unnecessary fail because the return vaue of the pop is not used. Refactoring those code path is beyond the scope of this PR.

sherpa/astro/tests/test_astro_data_notebook.py
hamogu: Not in sherpa itself, but users may have written code for `dat = DataPHA(..., header=None)` so I want to make sure that still works.
I think we could remove that, but in that case we should issue a DeprecationWarning first for at least one full version, and there is no harm done keeping it around.

sherpa/astro/instrument.py
```diff
@@ -1288,6 +1288,10 @@ def create_delta_rmf(rmflo, rmfhi, offset=1,
 
     .. versionadded:: 4.10.1
 
+    .. versionchanged:: 4.13.2
+
+       Added header keyword
+
```
hamogu, line 1294: Do we really need a "versionchanged" for adding a keywords that won't usually be used? I put it here because that's where other functions have it, but personally I feel that `versionchanged` can be reserved for breaking changes; we don't need to keep a changelog of each function in the docs. For things that are as small as this one we have the changelog. Let me know if you agree, because I'd be happy to remove this.

DougBurke: I have "complicated" thoughts on issues like this because combing through multiple changelogs to find out when something was changed is not a great experience. However, there's plenty of changes I'm making that are more significant than adding a keyword to a rarely-used function that never get called out (because I forgot to add it or the behavior is from combination of things so a versionchanged line isn't useful to add!). On the other hand, having five or more sections listing things that have changed does not improve the documentation experience! What I would like to discuss at some point is #555 but that's not relevant for this PR.

I don't believe there's a way to come up with a hard-and-fast rule for when something like this is needed or not (e.g. things like removing a parameter or changing some fundamental behavior is more important in general than adding an optional argument, but if that optional argument adds new functionality to an often-requested function then ...). So we have to go with our guts.

I think in this case the versionchanged` line is not needed.

sherpa/astro/instrument.py
DougBurke: Am I right in thinking this comment can be closed with your new changes?

hamogu: yes, done.

sherpa/astro/instrument.py
hamogu: done.

sherpa/astro/instrument.py
```diff
@@ -1234,11 +1234,16 @@ def get_kernel(self, data, subkernel=True):
...
 
 
-def create_arf(elo, ehi, specresp=None, exposure=None, ethresh=None, name='user-arf'):
+def create_arf(elo, ehi, specresp=None, exposure=None, ethresh=None,
+               name='user-arf', header=None):
     """Create an ARF.
 
     .. versionadded:: 4.10.1
 
+    .. versionchanged:: 4.13.2
```
wmclaugh, line 1243: See version changed conversation a line 1302. The information is better placed in change logs/release notes. One reason is to avoid mismatches or incorrect information (for example when code changes are delayed or subsequently modified and notes aren't updated. In this case, we are not expecting to have a 4.13.2 release. The release schedule was modified to align with the CIAO beta2 schedule so the upcoming release is 4.14.0.

Issue #1160, opened 2021-05-25
Changing the grouping messes up the filter


DougBurke:
This is with CIAO 4.13 but it still holds in Sherpa 4.13.1

```
sherpa In [1]: load_data('3c273.pi')
WARNING: systematic errors were not found in file '3c273.pi'
statistical errors were found in file '3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file 3c273.arf
read RMF file 3c273.rmf
WARNING: systematic errors were not found in file '3c273_bg.pi'
statistical errors were found in file '3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file 3c273_bg.pi

sherpa In [2]: notice(0.5, 7)

sherpa In [3]: get_dep(filter=True).shape
Out[3]: (42,)

sherpa In [4]: set_grouping(np.ones(1024))

sherpa In [5]: get_dep(filter=True).shape
DataErr: size mismatch between mask and data array
```

Setting the `grouping` array - whether via `set_grouping` or the `grouping` attribute of a `DataPHA` object - does not re-calculate the mask.

DougBurke:
I created https://github.com/DougBurke/sherpa/tree/fix-1160 based on code taken from #1161 but it didn't immediately work as I'd expected and then I got pulled into fixing a number of other issues. This message is so I don't forget.

DougBurke:
This bug is still present even after #1216 has been merged.

DougBurke:
#1470 
has some changes which should help, but I am not planning on addressing this issue the PR

DougBurke:
So, I have one way of fixing this, but it's a bit tricky. Consider the case when we have

| channel | noticed? |
| ----------- | -------------|
| 1           | True        |
| 2           | True        |
| 3           | True        |
| 4           | False       |
| 5           | True        |
| 6           | True        |
| 7           | True        |

If we group this up then what happens to any group that contains channels 3 and 4 or 4 and 5? The current behavior is that we treat these "mixed" bins as being included/noticed. This is the way `notice` works when given a low or high value that falls within a group - you select all of that group. The `ignore` command also works this way (but in this case you ignore the group if the low or high edge falls within this group.

[...]

PR #1161, opened 2021-05-25
WIP Track actual pha filter ranges

### LINKED ISSUE #1024: PHA data, models, and notice - what is going on
This is with CIAO 4.13.0 developmemt but I've seen this with the master branch:

# OKAY

```
...

sherpa In [3]: notice(0.5, 5)

sherpa In [4]: plot_data()
```

sherpa

# WAT

The only difference is a call to `plot_model` before the `notice` call, and it looks like the `notice` has been ignored!

```
...

sherpa In [3]: plot_model()

sherpa In [4]: notice(0.5, 5)

sherpa In [5]: plot_data()
```

sherpa

DougBurke:
WIP - this is likely to need reworking if #1208, #1215, #1216 ... get merged.

This PR contains commits that have also been added as #1168 #1171 #1172 #1173 - I have removed the code fro 1168, 1172, and 1173 from this PR (1171 sets up tests that then get changed so it's still needed). 

# Summary

Track the requested filter range of PHA data sets so that when the dataset is ungrouped, or the grouping scheme is changed, then the filters can be re-applied. This makes the resulting filters better match the intent of the user. Fixes #1024 and #1160 in the process. There have been documentation updates and additions.

This is a breaking change, in that a set of commands like the following (e.g. a notice then ignore followed by a group change) can lead to a slightly-different range of data being included:

    load_pha('3c273.pi')
    group_counts(10)
    notice(0.5, 7)
    notice(1.1, 1.2)
    get_filter()
    -> '0.489099994302:1.058499991894,1.255600035191:8.314700126648'

    group_counts(20)
    get_filter()
    -> '0.510999992490:1.058499991894,1.211799979210:10.110499858856'  BEFORE THIS PR
    -> '0.510999992490:1.058499991894,1.335900008678:10.110499858856'  WITH THIS PR

Because of this, commands like `plot_model` and `plot_order` can show slightly-different channel ranges.

There is no change to the handling of quality channels, and hence the `ignore_bad` command.

# Details

This work places the logic in the DataPHA class, which makes sense as this is where we have to deal with mapping from units to channels and grouping. There are two places that functionality could go - somewhere further up the Data inheritance chain, or in the Filter class - but I don't think they fit. It should be thought about though when reviewing this PR.

I had originally tried tracking the exact channel range that matches the filter requests - e.g. if notice(0.5, 7) maps to channels 53 to 470 but thanks to groups we end up using 48 to 590 then I tracked the 53-470 range in a structure similar to the _filter_store. These channel ranges were then used when changing grouping or switching grouping on or off. However, this didn't seem to work out well: it's the mapping that an individual command has to do when deciding what to do with those channels in the edge bins which are partially selected and partially un-selected - in my previous approach I'd lost the sense of whether it was a notice or ignore and this was a problem. I could take the part where the filter is stored in channel units (which would avoid some work), but a nice feature of the current approach is that it uses notice/ignore so all the constraints and updates they make are done for us, rather than having to worry about the book-keeping.

When a `DataPHA` object is created we add

        # Track the requested noticed/ignored data range, so that it
        # can be re-created when swaping between grouped and ungrouped
        # data.
        #
        # We start off with an explicit "notice everything" call, as this
        # means less special cases. At least, that's the hope.
        self._filter_store = [(None, None, 'channel', False)]

The idea is that each `notice` or `ignore` call appends to this store, adding the low, high, current units, and a flag indicating if it's an `ignore call`. This is done by the `_filter_add` method: it's only used once, shown below, in the `notice` method so we could just avoid the method but I liked the chance to add some text to the method, and wasn't sure how many times it would be used.

        # Add the requested filter setting, *before* it is converted
        # into a channel range.
        #
        self._filter_add(lo, hi, ignore)

Note that `_filter_add` has a special case when it erases the existing `_filter_store` array - this is when `lo=hi=None`, i.e. the user is requesting the whole range either be noticed or ignored. At this point we reset the store to `[(None, None, 'channel', ignore)]` (essentially the same as when we created the `DataPHA` object. All other cases just append to the store. Note that we could be clever here and identify any call that ends up filtering all the data - e.g. `notice(0, 1000000)` or when all the existing filters end up mapping to either including or excluding all data - but I have not done that. Thanks to how filtering and grouping works, a user could end up removing all bins thanks to a particular grouping scheme, but if we went ungrouped we'd still have data. We could handle a user request of lo/hi values that exceed the range (i.e. treat them as we do `lo=hi=None`) but I don't think it's worth it - the existing `notice` logic is messy enough as is for what is potentially a small space and time saving.

This is all that `notice` and `filter` do - just add each filter to the store. An important invariant is that the first element of `_filter_store` is `(None, None, 'channel', ignore)` - i.e. it either selects all channels or removes all channels. This means that subsequent commands (which we're about to talk about) work on a known state.

The 'grouped' setting - handled by the `_set_group` method has changed slightly and is now

    def _set_grouped(self, val):
        val = bool(val)

        if val and self.grouping is None:
            raise DataErr('nogrouping', self.name)

        if self._grouped == val:
            return

        self._grouped = val

        # Apply the existing filters, recreating the list as we go along.
        # This used to obly be done if self.mask was set but now rely
        # on the filter store to determine what needs to be re-evaluated.
        #
        self._filter_recreate()

The change is that the method used to end with:

        # As the grouping status is being changed, we need to reset the mask
        # to be correct size, while still noticing groups within the filter
        #
        if numpy.iterable(self.mask):
            old_filter = self.get_filter(group=val)
            self._grouped = val
            self.ignore()
            for vals in parse_expr(old_filter):
                self.notice(*vals)

         self._grouped = val

So the difference is that the logic for recreating the filter has been moved to the `_filter_recreate` method and that we now always do it if the grouping has changed. But what if the grouped fllag has not changed (but is True) but the actual grouping data has changed (this is issue #1160). Well, I have turned the `grouping` attribute, which stores the grouping data, into a property with the following setter:

```
    def _set_grouping(self, val):
        if val is None:
            # Should we clear the quality array too?
            self._grouping = None
            if self.grouped:
                self.grouped = False

            return

        # We should have more validation of quantities in the DataPHA
        # case, so add some basic ones here. The chosen errors could
        # well be changed; this is a first pass.
        #
        if self.channel is None:
            # Technically we could have grouping with no channels but
            # it doesn't make much sense (DJB is hoping to change __init__
            # to require channel and counts to be set).
            #
            raise DataErr('ogip-error', 'dataset', self.name,
                          'has no channels')

        val = numpy.asarray(val)
        try:
            if len(self.channel) != len(val):
                raise DataErr('mismatch', 'channel', 'grouping')
        except TypeError:
            # if val is a scalar
            raise DataErr('mismatch', 'channel', 'grouping')

        self._grouping = val

        # Recreate the filter if the grouping is set.
        #
        if self.grouped:
            self._filter_recreate()
```

This means that if we change the groupnig array we end up calling `_fitler_recreate`. I have also added some sanity checks on the array (something I think we should do more of in the `DataPHA` object).

The issue of what to do with the `quality` array and `ignore_bad` is currently ignored.

The `_filter_recreate` method - which is now used in other places, unike `_filter_add`, re-applies the requested filter. Previously we calculated the filter range as a string, parsed it to a bunch of notice statements, and then ran it. Instead we can re-run the actual `notice` and `ignore` calls the user made - which will then be evaluated with the new grouping scheme (or the ungrouped data). This is an easy command to write - we just have to deal with unit changes:

    def _filter_recreate(self):
        """Re-apply the current set of filters.

        This is intended to be used if the grouping has been changed:
        that is turning on or off the grouping or the grouping scheme
        has changed. It will attempt to create the overall filter
        as close to the user-requested range as possible.

        """

        # The _filter_store array is designed so that it starts with
        # either a "notice all" or "ignore all" command. Therefore
        # when we create it the first notice call will clear the
        # _filter_store list, resulting in the list being rebuilt.
        # We could avoid doing this but by going through notice we
        # aim to keep al the invariants.
        #
        ounits = self.units
        ostore = self._filter_store[:]
        try:
            for lo, hi, units, ignore in ostore:
                self.units = units
                self.notice(lo, hi, ignore)

        finally:
            self.units = ounits

Note that as `ostore` starts with `(None, None, ...)` - as previously mentioned - then the first `self.notice` call is going to hit `_filter_add` with the `lo=hi=None` case which will delete the existing `_filter_store` array. So, we end up re-creating the `_filter_store` array (and is why we explicitly copy it into a separate variable `ostore`). This is a little subtle, and I don't know if the commentary above is enough for when we come back in a year.

I think this is conceptually nicer than converting a filter range to a string and then deconstructing that to get the required limits. Note that there is actually still a use case for this particular logic (`sherpa.astro.plot.ModelHistogram.prepare` where we want to get a range as close to the range used by the grouped data as we can, rather than the range that more-closely matches the user-requested range).

An important part of the design is that I looked at the  implementation of the `notice` method and decided that there is enough complexity there that while it might be useful to store "lower-level" details of the filter state to allow a more-surgical version of `_filter_recreate` (e.g. not having to mess around with changing the `units` setting) this is potentially dangerous, and it is better to stay at this "higher" level.

I was planning to address the ignore_bad command and the handling of quality channels here, but as I looked into it:

a) the logic is separated from the notice/ignore command so it can be tackled separately
b) it's looking to be a big issue (e.g. #1166) that is better handled in a separate PR

# What to worry about?

What happens if a user tries to direct modify the mask array - e.g. `pha.mask[0] = True`? This approach doesn't know about this.

The old system, which converts the mask into a filter expression (e.g. `0.5:2.1,2.2:8.0`) is okay if you directly edit the mask array. So this is a major problem with this new approach. Dang it. We *could* say "don't edit the mask attribute" for DataPHA objects, but that strikes me as a poor compromise. So I think 'll try to add some subset of the changes here for testing and documentation purposes while I mull on things.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1233,7 +1233,6 @@ class DataPHA(Data1D):
     syserror
     bin_lo
     bin_hi
-    grouping
     quality
     exposure
     backscal
@@ -1335,7 +1334,9 @@ class DataPHA(Data1D):
             self._filter_recreate()
 
     grouping = property(_get_grouping, _set_grouping,
-                        doc='The grouping scheme.')
+                        doc='The grouping scheme.\n\nAn array, matching the ' +
+                        'channels attribute, where a 1 means the start\n' +
+                        'of a group and -1 continues the previous group.')
 
     def _get_subtracted(self):
         return self._subtracted
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1295,6 +1295,48 @@ class DataPHA(Data1D):
     grouped = property(_get_grouped, _set_grouped,
                        doc='Are the data grouped?')
 
+    def _get_grouping(self):
+        return self._grouping
+
+    def _set_grouping(self, val):
+        if val is None:
+            # Should we clear the quality array too?
+            self._grouping = None
+            if self.grouped:
+                self.grouped = False
+
+            return
+
+        # We should have more validation of quantities in the DataPHA
+        # case, so add some basic ones here. The chosen errors could
+        # well be changed; this is a first pass.
+        #
+        if self.channel is None:
+            # Technically we could have grouping with no channels but
+            # it doesn't make much sense (DJB is hoping to change __init__
+            # to require channel and counts to be set).
+            #
+            raise DataErr('ogip-error', 'dataset', self.name,
+                          'has no channels')
+
+        val = numpy.asarray(val)
+        try:
+            if len(self.channel) != len(val):
+                raise DataErr('mismatch', 'channel', 'grouping')
+        except TypeError:
+            # if val is a scalar
+            raise DataErr('mismatch', 'channel', 'grouping')
+
+        self._grouping = val
+
+        # Recreate the filter if the grouping is set.
+        #
+        if self.grouped:
+            self._filter_recreate()
+
+    grouping = property(_get_grouping, _set_grouping,
+                        doc='The grouping scheme.')
+
     def _get_subtracted(self):
         return self._subtracted
 
@@ -1416,12 +1458,19 @@ class DataPHA(Data1D):
         self.bin_lo = bin_lo
         self.bin_hi = bin_hi
         self.quality = quality
-        self.grouping = grouping
+
+        # We do not want to trigger thelogic in _set_grouping here,
+        # so access the _grouping attribute rather than grouping.
+        self._grouping = grouping
         self.exposure = exposure
         self.backscal = backscal
         self.areascal = areascal
         self.header = header
+
+        # We do not want to trigger the logic in _set_grouped here,
+        # so access the _grouped attribute rather than grouped.
         self._grouped = (grouping is not None)
+
         # _original_groups is set False if the grouping is changed via
         # the _dynamic_groups method. This is currently only used by the
         # serialization code (sherpa.astro.ui.serialize) to determine
@@ -2485,22 +2534,9 @@ class DataPHA(Data1D):
             if kwargs[key] is None:
                 kwargs.pop(key)
 
-        old_filter = self.get_filter(group=False)
-        do_notice = numpy.iterable(self.mask)
-
         self.grouping, self.quality = group_func(*args, **kwargs)
         self.group()
         self._original_groups = False
-
-        if do_notice:
-            # self.group() above has cleared the filter if applicable
-            # No, that just sets a flag.  So manually clear filter
-            # here
-            self.ignore()
-            for vals in parse_expr(old_filter):
-                self.notice(*vals)
-
-        # warning('grouping flags have changed, noticing all bins')
 
     # Have to move this check here; as formerly written, reference
     # to pygroup functions happened *before* checking groupstatus,
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -33,7 +33,7 @@ from sherpa.data import Data1DInt, Data2D, Data, Data2DInt, Data1D, \
 from sherpa.models.regrid import EvaluationSpace1D
 from sherpa.utils.err import DataErr, ImportErr
 from sherpa.utils import SherpaFloat, pad_bounding_box, interpolate, \
-    create_expr, parse_expr, bool_cast, rebin, filter_bins
+    create_expr, bool_cast, rebin, filter_bins
 from sherpa.utils import formatting
 
 # There are currently (Sep 2015) no tests that exercise the code that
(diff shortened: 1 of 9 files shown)


sherpa/astro/data.py
```diff
@@ -1439,6 +1486,13 @@ def __init__(self, name, channel, counts, staterror=None, syserror=None,
         self._plot_fac = 0
         self.units = 'channel'
         self.quality_filter = None
+        # Track the requested noticed/ignored data range, so that it
+        # can be re-created when swaping between grouped and ungrouped
+        # data.
+        #
+        # We start off with an explicit "notice everything" call, as this
+        # means less special cases. At least, that's the hope.
```
hamogu, line 1494: I'm always encourage by these slight upbeat comments!

DougBurke: The approach has so far seemed to work out well, so I'm just waiting for the other shoe to drop...

Or when I realize it won't actually help us in our plans to fix all the things™.

DougBurke:
Rebased to address test conflicts from #1138

DougBurke:
Rebased to remove #1168 and #1172

DougBurke:
It's too hard to integrate this with recent changes to the code (or, rather, I'm not convinced enough that this approach is worth it, so I don't want to spend more time trying to rebase this than I already have), so I'm closing this.

PR #1163, opened 2021-05-27
XMM/RGS triggers a notice/ignore bug because channels are in reverse-energy order in RMF

hamogu:
# Summary
Recent work on the ignore / notice logic in #1127 implicitly assumes that the energy grid in PHA/ARF/RMF files is in increasing order, which does not hold true for XMM/RGS data.

*The problem was introduced in #1127, which is not yet included in any released version, so a separate changelog entry for this PR might not be necessary?*

# Details
Since #1127 Sherpa errors out at https://github.com/sherpa/sherpa/pull/1127/files#diff-39d64572ec8588c14cccff43dc2d85801af7524041351c8434f80963e847204bR3569

Before #1127, XMM data could be read, noticed, and ignored by sherpa. This PR fixes that and adds tests.

Note that CI will fail until https://github.com/sherpa/sherpa-test-data/pull/25 is merged, which adds the required XMM data files for the test.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3612,7 +3612,7 @@ class DataPHA(Data1D):
 
         # Ensure the data is in ascending order for create_expr.
         #
-        if self.units == 'wavelength':
+        if len(x) > 0 and x[-1] < x[0]:
             x = x[::-1]
             mask = mask[::-1]
 
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3759,8 +3759,8 @@ class DataPHA(Data1D):
             umax = max(hc / elo[[0, -1]])
         else:
             # assume channel units
-            umin = min(self.channel[0], self.channel[-1])
-            umax = max(self.channel[0], self.channel[-1])
+            umin = self.channel[0]
+            umax = self.channel[-1]
 
         assert umin < umax, (self.units, umin, umax)
 
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3748,15 +3748,19 @@ class DataPHA(Data1D):
             return
 
         if self.units == 'energy':
-            umin = elo[0]
-            umax = ehi[-1]
+            # It's not guaranteed that channels are in increasing energy
+            # but we can certainly assume that they are monotonic.
+            # Thus, the min of the first and last entry is the minimum
+            # energy of the elo array
+            umin = min(elo[0], elo[-1])
+            umax = max(ehi[0], ehi[-1])
         elif self.units == 'wavelength':
-            umin = hc / ehi[-1]
-            umax = hc / elo[0]
+            umin = min(hc / ehi[[0, -1]])
+            umax = max(hc / elo[[0, -1]])
         else:
             # assume channel units
-            umin = self.channel[0]
-            umax = self.channel[-1]
+            umin = min(self.channel[0], self.channel[-1])
+            umax = max(self.channel[1], self.channel[-1])
 
         assert umin < umax, (self.units, umin, umax)
 
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3760,7 +3760,7 @@ class DataPHA(Data1D):
         else:
             # assume channel units
             umin = min(self.channel[0], self.channel[-1])
-            umax = max(self.channel[1], self.channel[-1])
+            umax = max(self.channel[0], self.channel[-1])
 
         assert umin < umax, (self.units, umin, umax)
 
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -871,22 +871,6 @@ def test_pha_grouping_changed_filter_1160(make_test_pha):
     assert d4 == pytest.approx([2, 3])
 
 
-def test_pha_reverse_sorted():
-    '''This case for a PHA set with reverse sorted channels is a little
-    non-sensical, but it is a very easy way to hit the problem in #1163.
-
-    Really, we want to test that with a larger test that reads real XMM
-    data, which uses reverse energy ordering (see next test).
-
-    So, if thise tests starts to fail, but the next one passes, this one can
-    probably be removed. It's here now because it's fast, easy, and does not
-    require the large test data.
-    '''
-    d = DataPHA('x', np.array([4, 3, 2, 1, 0]), np.array([5, 4, 3, 2, 1]))
-    d.notice(2, 4)
-    assert len(d.get_dep(filter=True)) == 3
-
-
 @requires_fits
 @requires_data
 def test_xmmrgs_notice(make_data_path):
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -908,7 +908,7 @@ def test_xmmrgs_notice(make_data_path):
     session.set_analysis('wave')
     session.notice(18.8, 19.2)
     dat = session.get_data()
-    assert len(dat.get_dep(filter=True)) == 40
+    assert len(dat.get_dep(filter=True)) == 41
 
     dat.ignore(10, 19.)
     assert len(dat.get_dep(filter=True)) == 20
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -899,8 +899,8 @@ def test_xmmrgs_notice(make_data_path):
     dat.units = 'wave'
     dat.notice(18.8, 19.2)
     assert len(dat.get_dep(filter=True)) == 41
-    assert dat.get_filter(format='%.2f') == '19.20:18.80'
+    assert dat.get_filter(format='%.2f') == '18.80:19.20'
 
     dat.ignore(10, 19.)
     assert len(dat.get_dep(filter=True)) == 20
-    assert dat.get_filter(format='%.2f') == '19.20:19.01'
+    assert dat.get_filter(format='%.2f') == '19.01:19.20'
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -887,28 +887,20 @@ def test_pha_reverse_sorted():
     assert len(d.get_dep(filter=True)) == 3
 
 
-@pytest.fixture
-def read_test_image(make_data_path):
-    from sherpa.astro.io import read_image
-    filename = 'acisf07999_000N001_r0035_regevt3_srcimg.fits'
-    d = read_image(make_data_path(filename))
-    d.name = 'test.img'
-    return d
-
-
 @requires_fits
 @requires_data
 def test_xmmrgs_notice(make_data_path):
     '''Test that notice and ignore works on XMMRGS dataset, which is
     ordered in increasing wavelength, not energy'''
-    from sherpa.astro.ui.utils import Session
-    session = Session()
-    session.load_data(make_data_path('xmmrgs/P0112880201R1S004SRSPEC1003.FTZ'))
-    session.load_rmf(make_data_path('xmmrgs/P0112880201R1S004RSPMAT1003.FTZ'))
-    session.set_analysis('wave')
-    session.notice(18.8, 19.2)
-    dat = session.get_data()
+    from sherpa.astro.io import read_pha, read_rmf
+    dat = read_pha(make_data_path('xmmrgs/P0112880201R1S004SRSPEC1003.FTZ'))
+    rmf = read_rmf(make_data_path('xmmrgs/P0112880201R1S004RSPMAT1003.FTZ'))
+    dat.set_rmf(rmf)
+    dat.units = 'wave'
+    dat.notice(18.8, 19.2)
     assert len(dat.get_dep(filter=True)) == 41
+    assert dat.get_filter(format='%.2f') == '19.20:18.80'
 
     dat.ignore(10, 19.)
     assert len(dat.get_dep(filter=True)) == 20
+    assert dat.get_filter(format='%.2f') == '19.20:19.01'
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -869,3 +869,46 @@ def test_pha_grouping_changed_filter_1160(make_test_pha):
     pha.grouping = [1, 1, -1, 1]
     d4 = pha.get_dep(filter=True)
     assert d4 == pytest.approx([2, 3])
+
+
+def test_pha_reverse_sorted():
+    '''This case for a PHA set with reverse sorted channels is a little
+    non-sensical, but it is a very easy way to hit the problem in #1163.
+
+    Really, we want to test that with a larger test that reads real XMM
+    data, which uses reverse energy ordering (see next test).
+
+    So, if thise tests starts to fail, but the next one passes, this one can
+    probably be removed. It's here now because it's fast, easy, and does not
+    require the large test data.
+    '''
+    d = DataPHA('x', np.array([4, 3, 2, 1, 0]), np.array([5, 4, 3, 2, 1]))
+    d.notice(2, 4)
+    assert len(d.get_dep(filter=True)) == 3
+
+
+@pytest.fixture
+def read_test_image(make_data_path):
+    from sherpa.astro.io import read_image
+    filename = 'acisf07999_000N001_r0035_regevt3_srcimg.fits'
+    d = read_image(make_data_path(filename))
+    d.name = 'test.img'
+    return d
+
+
+@requires_fits
+@requires_data
+def test_xmmrgs_notice(make_data_path):
+    '''Test that notice and ignore works on XMMRGS dataset, which is
+    ordered in increasing wavelength, not energy'''
+    from sherpa.astro.ui.utils import Session
+    session = Session()
+    session.load_data(make_data_path('xmmrgs/P0112880201R1S004SRSPEC1003.FTZ'))
+    session.load_rmf(make_data_path('xmmrgs/P0112880201R1S004RSPMAT1003.FTZ'))
+    session.set_analysis('wave')
+    session.notice(18.8, 19.2)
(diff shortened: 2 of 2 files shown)


DougBurke:
Thanks for checking with the XMM case. I'm just knee deep in other PHA filtering changes so even though this looks like a simple change I'm not able to get to it today.

hamogu:
You are giving me too much credit. I found this while debugging a separate problem that I first discovered on XMM data in 4.13.1, but then could not reproduce in main, because this stopped me from even setting up the data I needed (separate PR coming). So, it's not so much an intentional check of XMM data as more of an accidental discovery - and I approved #1127, so I only have myself to blame for not catching that there.

hamogu:
I take back what I said about the CI failing. I apparently put the files needed into this PR. I think that's good for reviewing, but I may have to do some git magic (that I don't know how to do right now) to clean up this PR and not include the files that are part of the submodule here, but just update the hash that the submodule is pointing to.

DougBurke:
Don't look at it as a problem, look at it as a way to extend our "How do I?" section of the developer docs! (i.e. the "how do I add a new dataset to the sherpa-test-data repository" question) ;-)

hamogu:
I should know - I added the last few files to that repro, too!

sherpa/astro/data.py
```diff
@@ -3563,15 +3563,19 @@ def notice(self, lo=None, hi=None, ignore=False, bkg_id=None):
...
         elif self.units == 'wavelength':
-            umin = self._hc / ehi[-1]
-            umax = self._hc / elo[0]
+            umin = min(self._hc / ehi[[0, -1]])
+            umax = max(self._hc / elo[[0, -1]])
         else:
             # assume channel units
-            umin = self.channel[0]
-            umax = self.channel[-1]
+            umin = min(self.channel[0], self.channel[-1])
```
DougBurke, line 3577: One of the things I'd like to see (not in this PR) is for `DataPHA.__init__` and it's attributes to make a numbre of sanity checks - e.g. that `.channel` and `.counts` have the same size and matches the DATCHANS value [if we store it?], that `.channel` is in increasing order, ... - so that we can more-safely access this related information.

Coming back to this change - the OGIP standard has channel stored in increasing order, so I think this check should not be needed. I'd be interested to know if the xmm data doesn't match this.

sherpa/astro/data.py
```diff
@@ -3563,15 +3563,19 @@ def notice(self, lo=None, hi=None, ignore=False, bkg_id=None):
...
+            umax = max(self.channel[1], self.channel[-1])
```
DougBurke, line 3578: If we do this -ie allow channel to be in decreasing order - then shouldn't the first argument to `max` be `self.channel[0]`?

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -816,3 +816,46 @@ def test_pha_invalid_units(invalid, make_test_pha):
...
+    probably be removed. It's here now because it's fast, easy, and does not
+    require the large test data.
+    '''
+    d = DataPHA('x', np.array([4, 3, 2, 1, 0]), np.array([5, 4, 3, 2, 1]))
+    d.notice(2, 4)
+    assert len(d.get_dep(filter=True)) == 3
+
+
+@pytest.fixture
+def read_test_image(make_data_path):
```
DougBurke, line 838: Is this routine used anywhere?

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -816,3 +816,46 @@ def test_pha_invalid_units(invalid, make_test_pha):
...
+    d.name = 'test.img'
+    return d
+
+
+@requires_fits
+@requires_data
+def test_xmmrgs_notice(make_data_path):
+    '''Test that notice and ignore works on XMMRGS dataset, which is
+    ordered in increasing wavelength, not energy'''
+    from sherpa.astro.ui.utils import Session
```
DougBurke, line 851: I don't think we need the session class here - we should be able to

    from sherpa.astro.io import read_pha, read_rmf
    dat = read_pha(...)
    rmf = read_rmf(...)
    dat.set_rmf(rmf)   # or whatever the method is called
    
    dat.units = 'wave'
    dat.notice(18.8, 19.2)

    # don't need the session.get_data call
    assert len(dat.get_dep(filter=True)) == 41
    
    dat.ignore(10, 19)
    assert len(...) == 20

It might be good to check the output of `dat.get_filter(format='%.4f')` after both filters [adjust the format setting to something sensible for this data set].

hamogu:
Mmm. I take this all back. XMM/RGS IS ordered in increasing energy. I don't know what let me to the conclusion that it's not. Yet, the notice/ignore bug is real and is solved by this PR, so I need ot dig deeper to understand why there is a problem and why I can solve it in this way.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -816,3 +816,46 @@ def test_pha_invalid_units(invalid, make_test_pha):
...
+    session = Session()
+    session.load_data(make_data_path('xmmrgs/P0112880201R1S004SRSPEC1003.FTZ'))
+    session.load_rmf(make_data_path('xmmrgs/P0112880201R1S004RSPMAT1003.FTZ'))
```
DougBurke, line 854: I note you add four files to `sherpatest` but only seem to use 2 here. Are the header keywords set up so that the other files automatically loaded here? Or is the plan to use these files with other tests yet-to-be-written?

sherpa/astro/tests/test_astro_data2.py
hamogu: No. I wonder how it got here...

sherpa/astro/tests/test_astro_data2.py
hamogu: The other two are a background spectra. They are both relatively small files. I don't have any other tests ready to go on my disk, but in the interest of making the data reusable for other tests, I wanted to have the minimum viable number of data files so that we *could* be used for other tests, e.g. background subtraction, plotting with background spectra etc. It's going to be a pain to find the exact matching versions in the archive again (because XMM also goes through repro cycles, which may add keywords etc.), so I think we should invest the extra 39 kB to hae then when we need them.

sherpa/astro/data.py
hamogu: Channel is indeed in increasing order, but OGIP does not require channel to be proportional to energy. After all, the whole concept of "channel" is needed precisely because the channel may not have a simple (and known) mapping to energy. In the XMM/RGS RMFs, the channel number is increasing, but not the energy. Instead is is increasing with wavelength (not shown in the EBOUNDS extension, but we know how to convert energy to wavelength).
```
(ciao) [Hanss-MacBook-Air:sherpa-test-data/sherpatest/xmmrgs] hamogu% dmlist "P0112880201R1S004RSPMAT1003.FTZ" blocks 
 
Dataset: P0112880201R1S004RSPMAT1003.FTZ
 
Block    2: MATRIX                         Table         6 cols x 4000     rows
Block    3: EBOUNDS                        Table         3 cols x 3600     rows
(ciao) [Hanss-MacBook-Air:sherpa-test-data/sherpatest/xmmrgs] hamogu% dmlist "P0112880201R1S004RSPMAT1003.FTZ[EBOUNDS]" data
 
Data for Table Block EBOUNDS
 
ROW    CHANNEL    E_MIN                E_MAX
 
     1          1         3.0918762684         3.0996062756
     2          2         3.0841853619         3.0918762684
     3          3         3.0765318871         3.0841853619
     4          4         3.0689170361         3.0765318871
     5          5         3.0613391399         3.0689170361
     6          6         3.0537991524         3.0613391399
     7          7         3.0462958813         3.0537991524
     8          8         3.0388295650         3.0462958813
     9          9         3.0313994884         3.0388295650
    10         10         3.0240061283         3.0313994884
    11         11         3.0166482925         3.0240061283
```
In fact, for grating spectra, "wavelength" is the natural coordinate. It's linear from the origin, it's the quantity we usually care about. However, Chandra orders in energy, even for grating spectra:
```
(ciao) [Hanss-MacBook-Air:sherpa/sherpa-test-data/sherpatest] hamogu% dmlist "3c120_heg_-1.rmf.gz[EBOUNDS]" data
 
Data for Table Block EBOUNDS
 
ROW    CHANNEL    E_MIN                E_MAX
 
     1          1     0.57720756530762     0.57727473974228
     2          2     0.57727473974228     0.57734197378159
     3          3     0.57734197378159     0.57740914821625
     4          4     0.57740914821625     0.57747644186020
     5          5     0.57747644186020     0.57754367589951
     6          6     0.57754367589951     0.57761090993881
     7          7     0.57761090993881     0.57767820358276
     8          8     0.57767820358276     0.57774549722672
     9          9     0.57774549722672     0.57781279087067
    10         10     0.57781279087067     0.57788014411926
    11         11     0.57788014411926     0.57794749736786
```
I think it is a moot point for us why exactly XMM chose to do it that way - in my opinion *not* supporting XMM is not an option. 

Note that the XMM data is sane, in the sense that the channel number *is* increasing, the number of columns is correct etc. It just does not follow the implicit assumption build into #1127.

sherpa/astro/data.py
hamogu: Yes, it should, and do think we have to allow this. We won't convince XMM to change their file format and reprocess all data just because it's more convenient for us.

sherpa/astro/tests/test_astro_data2.py
hamogu: I know... it's use a lot higher up in the file and I accidentally copied it when I copied one of the other tests to use as template for how to write this one. Fixed.

sherpa/astro/data.py
hamogu: On the other hand, the MATRIX extension is ordered in **increasing** energy, but OGIP does not say that the row number in the MATRIX extension has to correspond to the order in the EBOUNDS extentions. To a certain degree, the ordering in MATRIX is arbitrary, since each row says with channels (through F_CHAN and N_CHAN) are mapped to which (through the MATRIX). Now, I can't say that I *understand* why they did it this way or that I agree that it is a good choice, (it certainly caused me enough confusion to change the title of this PR twice), but have to deal with it - which is waht this PR is about.

hamogu:
> Mmm. I take this all back. XMM/RGS IS ordered in increasing energy.

See details in this comment:
https://github.com/sherpa/sherpa/pull/1163#discussion_r643258528

In short, the RMF MATRIX extension is ordered in increasing energy, but the EBOUNDS extension is in decreasing energy.

sherpa/astro/data.py
DougBurke: I don't understand your comment here: `self.channel` is in increasing order for XMM PHA data, so why are you suggesting we change their file format. In other words, why is `umin = self.channel[0]; umax = self.chanel[-1]` not going to work? The fact that E_MIN/MAX is in decreasing order in the XMM means we have to fix our code in other places, but not in these two lines of code I think.

hamogu: Yes, this will work. 

(Forget about the other comment. I was just thinking out loud. *If* we could make XMM work with increasing energy, then this PR would not be needed. Bot that won't happen, so this PR IS needed.) I'm not sure we have to fix any other places. I work on XMM very often and I do open issues or PRs when I run into something like here. It used to work and it will work after this and the plotting PR again. But if you are aware of other areas, please do flag them.

hamogu: So, to the specific question: Of course, `umin = self.channel[0]; umax = self.chanel[-1]` works, but then the assert statement in the next line fails the assert: https://github.com/sherpa/sherpa/pull/1163/files#diff-39d64572ec8588c14cccff43dc2d85801af7524041351c8434f80963e847204bR3580

DougBurke: In https://github.com/sherpa/sherpa/pull/1163#discussion_r643258528 you say that channel is in increasing order, so I don't see how `assert umin < umax` could fail.

What does

```
% dmlist xmmrgs/P0112880201R1S004SRSPEC1003.FTZ"[cols channel,counts][#row=1:10]" data,clean
```

return?

hamogu:
```
(ciao) [Hanss-MacBook-Air:sherpa/sherpa-test-data/sherpatest] hamogu% dmlist xmmrgs/P0112880201R1S004SRSPEC1003.FTZ"[cols channel,counts][#row=1:10]" data,clean
#  CHANNEL COUNTS
      1          0
      2          0
      3          0
      4          0
      5          0
      6          0
      7          0
      8          0
      9          0
     10          0
```
(The first few channels are always empty...) Why exactly are you asking for this range?

DougBurke: Basically it's just to check that channel is indeed increasing. So I really do not see how the `assert` fails - **unless** it's failing because it's not being triggered by the `umin = self.channel[0]/umax = self.channel[-1]` lines but why one of the other conditions.

hamogu: Ahh, Ok, now I understand. I will revert this change. It should only be needed for the energy and wavelength case.

hamogu: @DougBurke I apologize. This thread has gotten too long and I had taken this as a general discussion, while I now realize that you asked specifically about those tow lines.

DougBurke: For that I think I'll assign another PR to you .....

hamogu: As if you had not done that otherwise... but I like how this is going. I feel we're making more progress in Sherpa than we did in a while.

hamogu:
@DougBurke I think I resolved all your comments. Can you approve if you agree. Also, can you approve https://github.com/sherpa/sherpa-test-data/pull/25 ?
As far as I know, the PR in the submodule has to be merged first because this repro needs to refer to the commit hash of the merge commit. (Meaning that I need to rebase and include that once the PR to sher-test-data is in.)

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -816,3 +816,22 @@ def test_pha_invalid_units(invalid, make_test_pha):
...
+    '''Test that notice and ignore works on XMMRGS dataset, which is
+    ordered in increasing wavelength, not energy'''
+    from sherpa.astro.io import read_pha, read_rmf
+    dat = read_pha(make_data_path('xmmrgs/P0112880201R1S004SRSPEC1003.FTZ'))
+    rmf = read_rmf(make_data_path('xmmrgs/P0112880201R1S004RSPMAT1003.FTZ'))
+    dat.set_rmf(rmf)
+    dat.units = 'wave'
+    dat.notice(18.8, 19.2)
+    assert len(dat.get_dep(filter=True)) == 41
+    assert dat.get_filter(format='%.2f') == '19.20:18.80'
```
DougBurke, line 833: We want get_filter to return 18.80:19.20 here, which is a problem.

My bet is that the old check in `get_filter` is no-longer needed with this dataset

https://github.com/sherpa/sherpa/blob/a6f8cc20091d9675cfb5f45054e0c11850d01926/sherpa/astro/data.py#L3428-L3432

I bet we still need it for other datasets (e.g. Chandra) since if I add `raise NotImplementedError` within the  if check then a number of tests fail (and they also fail if I comment out the section, which is more useful), so we need to change the

    if self.units == 'wavelength':

to something like the untested

    if x[-1] < x[0]:

hamogu:
Fixed sorting (we are certainly getting better in checking for sorting issues!) and reworded summary.

sherpa/astro/data.py
```diff
@@ -3427,7 +3427,7 @@ def get_filter(self, group=True, format='%.12f', delim=':'):
 
         # Ensure the data is in ascending order for create_expr.
         #
-        if self.units == 'wavelength':
+        if len(x) > 0 and x[-1] < x[0]:
```
DougBurke, line 3430: I was wondering about a len check. I **think** at this point we're guaranteed that x and mask are not empty, but it doesn't hurt to add the check here.

DougBurke (approved):
LGTM - although needs sherpa/sherpa-test-data#25 to get merged first then this can get rebased

DougBurke:
I just noticed this on the XSPEC page, so it's not just us ;-)

```
Note!!: A bug in the heasp library is causing plotting problems in XSPEC for cases where EBOUNDS energies are grouped and listed in descending order of energy (as often used for gratings). The patch fix for this is not currently integrated into the XSPEC patch fixes below. But it is available separately and can be easily applied FROM HERE.
```

hamogu:
But we can still fix this first!

DougBurke:
Well, there's no a conflict to resolve, unfortunately, but this time it's not my fault but whoever wrote #1178...

hamogu:
Fortunately, I happen to know that piece of code, so it's a fairly easy rebase... Now, of course, I'm going to fall into the failing conda downloads...

DougBurke:
@hamogu - yay more conflicts (and this one probably is my fault) for you to work on next week!

hamogu:
I'm waiting for https://github.com/sherpa/sherpa-test-data/pull/25 to be merged first because then I'll have to fiddle with this anyway to make sure the submodule points to the merge commit.

DougBurke:
Sounds like a plan.

hamogu:
rebased and picked up the new commits in the submodule.

DougBurke:
I see the tests fail on

```
___________________________ test_pha_reverse_sorted ____________________________

    def test_pha_reverse_sorted():
        '''This case for a PHA set with reverse sorted channels is a little
        non-sensical, but it is a very easy way to hit the problem in #1163.
    
        Really, we want to test that with a larger test that reads real XMM
        data, which uses reverse energy ordering (see next test).
    
        So, if thise tests starts to fail, but the next one passes, this one can
        probably be removed. It's here now because it's fast, easy, and does not
        require the large test data.
        '''
        d = DataPHA('x', np.array([4, 3, 2, 1, 0]), np.array([5, 4, 3, 2, 1]))
>       d.notice(2, 4)

sherpa/astro/tests/test_astro_data2.py:886: 
```

I'm guessing we're hitting the case of "we should remove this test as it's not meaningful"?

hamogu:
It's one that I had already removed but that then came back from the dead due to an insufficient rebase...

DougBurke (approved):
Just re-approving given the rebases,

PR #1164, opened 2021-05-28
Fix plots with wavelength

hamogu:
# Summary
For some data files, plots with a reversed grid (i.e. plots in wavelength) were drawn with disjoint lines since #906.

# Details
Since #906 the plotting code checks if grids are continuous and breaks the line if that's not the case (e.g. bins in between have been filtered out as bad). That comparison (`xlo[1:] = xhi[:-1]`) breaks down if arrays are reverse ordered or low and high are exchanged. I'm not sure if all four combinations of reversing order and  exchanging low and high occur in practice, but I have seen at least three of them, so I'm now making the code robust to all four possible combinations. To that end, I'm reassigning the `xlo` and `xhi` in the plotting code, such that `xlo` is actually the lower bin boundary in all cases. This change requires changing some tests which relied on the old (I would argue sloppy) behavior.

In principle, the same problem can occur for the code that test if two bin boundaries are "close enough" i.e. within numerical precision. I don't have an exploit, since all my test files are well behaved in this respect, but it seems safe to apply the same fix there, just in case.

I discovered this problem in XMM plotting:
index

but because of #1163 I have trouble loading XMM files for now. So, I verified that the same problem occurs in main with Chandra grating data. Here is an example that's fixed with the current code:
index2


DougBurke:
I thought I had checked the wavelength case with Chandra data for #906. Oh dear.

I haven't looked at the code for this PR yet but - given the other wavelength-ish issues, cough cough #1163, and the various comments I've been adding along the lines of "I don't know if we need to bother about checking this case" then we should consider either documenting the behavior for the common cases for out data accessors or making them behave in a consistent manner, no matter what the raw data is. Outside the scope of this PR though.

DougBurke:
@hamogu - just to check, but your fixed plot looks okay if you then do a `ignore(19.1, 19.14)` or equivalent (I don't see why it shouldn't but the existing code is messy...)

hamogu:
> and the various comments I've been adding along the lines of "I don't know if we need to bother about checking this case" then we should consider either documenting the behavior for the common cases for out data accessors or making them behave in a consistent manner, no matter what the raw data is. Outside the scope of this PR though.

I think this PR should take care of the stuff that's related to plotting. #1163 is a at least a good step that way. It also adds an XMM test, so we'll find at least the most obvious things. And I work with XMM data (which is how I found these two), so I know I can run through a full analysis otherwise. While @DougBurke is of course right in principle, I don't think we are that far off from "making them behave consistently", we just need to include XMM/RGS in testing and keep this in mind for other grouping/filtering PRs in the future. For a moment I thought about re-ordring on read-in but that's so messy when you think about ARFs and RMFs that may be rad in at a later time or be modified by code in some way.

DougBurke:
@hamogu - I just meant that if you ignored a range within the plotted range then does the plot still look okay - which is why I picked an ignore range within the displayed wavelengths.

hamogu:
This is what I expect, but you made the plots, so you know best how they should look like:

index4

DougBurke:
That's what I expect to. And this works now for the XMM case to?

hamogu:
Not before #1163 is merged... but if you have both it works. (Pasting image here not because I'm afraid @DougBurke would not believe me, but because I want a record of what exactly we tested if any further problems come up later and "blame" points to this PR).
index5

sherpa/astro/plot.py
```diff
@@ -64,25 +64,43 @@ def _check_hist_bins(plot):
...
+    # Deal with xhi <-> xlo switches. Those can occor when converting
+    # from energy to wavelength.
+    # Deal with reversed order. Can happen when converting from energy
+    # to wavelength, or if input PHA is not ordered in increasing energy.
+    # But is both are happening at the same time, need to switch twice, which
+    # is a no-op. So, we get to use the elusive Python XOR operator.
+    if (xlo[0] > xhi[0]) ^ (xhi[0] > xhi[-1]):
+        xlo, xhi = xhi, xlo
+    equal = sao_fcmp(xlo[1:], xhi[:-1], _tol)
+    print(equal)
```
DougBurke: We should remove this debug line

sherpa/plot/tests/test_utils.py
```diff
@@ -0,0 +1,46 @@
...
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+import numpy as np
+
+from sherpa.plot.utils import histrogram_line
+
+
+def test_histrogram_line():
```
DougBurke: Can we add a docstring so that when we come back to this test we have some idea of what it's to test (a common problem when I look back at tests I've written for Sherpa)?

sherpa/plot/utils.py
```diff
@@ -0,0 +1,88 @@
...
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+import numpy as np
+
+__all__ = ('intersperse', 'histogram_line')
+
+
+def intersperse(a, b):
+    '''Interleave two arrays a and b
```
DougBurke, line 26: So this routine makes n assumptions on the ordering? ie `intersperse([1, 7, 3, 4], [100, -200, 34, 5])` will work (even if not really meaningful, and would fail because they're not numpy arrays?)

sherpa/plot/utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    See https://stackoverflow.com/questions/5347065/interweaving-two-numpy-arrays#5347492
+    # for interleaving arrays.
```
DougBurke, line 29: is the leading # some docstring bit of cleverness, or a typo?

sherpa/astro/plot.py
hamogu: Maybe you need to reload? I don't see that print line and github mark the view of this comment as "outdated"

sherpa/plot/utils.py
DougBurke: OOh. I see this is my code from `pylab_backend` that is being spilled out.

sherpa/astro/plot.py
DougBurke: I'm viewing by commits rather than the whole commit. I see it's removed in the next commit.

sherpa/plot/utils.py
hamogu: I don't know. This routine was originally written by @DougBurke I just moved it to a different files for (a) easier testing and (b) availability from different plotting backends (it's originally from the pylab_matplotlib, where I shoudl now import this version and remove the definition there. Alas, I forgot...

DougBurke: I'd originally put it within another routine, so it was really local to that routine. Here it is now top-level (but not exported) routine. Do we need more-stringent documentation for cases like this?

I was going to mention that we need to remove the code from `pylab_backend`.  I was going to ask whether it's worth splitting this out of the pylab code as we don't use it outside of the pylab code (other than in a test added in this PR). Or am I missing something?

hamogu: We don't currently use it elsewhere. But we will probably need it for any other backend, if we want the plots to look similar. Plus the code is not matplotlib specific at all, so it does not hurt to have it in it's own file with its own tests.

hamogu:
@DougBurke I just removed the code from pylab, so it's not calling the version in utils (which is also the version covered by tests). While you are at it, you might want to look again.

DougBurke:
Unfortunately I have to go off and review a couple of papers, so don't have time for more Sherpa at the moment.

hamogu:
I also broke it again, so no hurry.

DougBurke:
Do we want #1164 or #1163 merged first, or does it not matter?

DougBurke:
And apologies for causing a rebase thanks to #1118

hamogu:
These should be independent and can be merged in any order.

hamogu:
rebased

DougBurke:
Can we clarify the summary as it's not a problem for all files but some files (so that when we come to think about documenting this in December we don't get in a tizzy)?

sherpa/plot/utils.py
hamogu: I think current lelvel of documentation is sufficient for a very short, not exported helper function.

sherpa/astro/plot.py
```diff
@@ -64,25 +64,49 @@ def _check_hist_bins(plot):
...
+    if len(xlo) == 0:
+        return xlo, xhi
 
     # Technically idx should be 0 or 1, with no -1 values. We
     # do not enforce this. What we do is to take all bins that
     # appear similar (sao_fcmp==0) and set the xlo[i+1] bin
     # to the xhi[i] value.
     #
-    equal = sao_fcmp(plot.xlo[1:], plot.xhi[:-1], _tol)
+    # Deal with xhi <-> xlo switches. Those can occor when converting
```
DougBurke, line 97: occor -> occur but I think the first two lines can get removed as it repeated in "Deal with reversed order." I think

sherpa/plot/utils.py
hamogu: Actually, it is exported (it's listed in `_all__`) . It's still a pretty simple helper routine.

DougBurke: For now I'd suggest removing `intersperse` from `__all__`. If needed we can add it later/

hamogu:
Also re-worded summary, so should be all ready now.

DougBurke (approved):
LGTM

hamogu:
rebased and fixed conflicts

PR #1165, opened 2021-05-28
Add a section on running the tests to the developer docs

hamogu:
# Summary
Add a section on running the tests to the developer docs

[skip CI]

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -41,7 +41,47 @@ Sometimes a PR is still in development and known to fail the tests or
 simply does not touch any code, because it only modifies docstrings
 and the documentation. In that case, `[skip ci]` can be added to the
 commit message to prevent running the github actions tests to save
-time, energy, and limited free recources.
+time, energy, and limited recources.
+
+Run tests locally
+-----------------
+Before you issue a pull request, we ask to run the test suite locally.
+Assuming everything is set up to install Sherpa from source, it can be
+installed in development mode with ``pip``::
+
+  pip install -e .
+
+"Development mode" means that the tests will pick up changes in the
+Python source files without running ``pip`` again (which can take some
+time). Only if you change the C code, you will have to explicitly run
+the installation again to see the changes in the tests. After the installation,
+``pytest`` can run all the tests. In the sherpa root directory call::
+
+  pytest
+
+``pytest`` supports a number of options which are
+`detailed in the pytest documentation <https://docs.pytest.org/>`_. A
+particularly useful option is to run only the tests in a specific file.
+For example, if you changed the code and the tests in the `sherpa.astro.ui`
+module, one might expect tests for this module to be the most likely to fail::
+
+  pytest sherpa/astro/ui/tests/test_astro_ui.py
+
+Once everything looks good, you can do a final run of the entire test suite. A
+second option useful for develoment is ``--pdb`` which drops into the
+`interactive Python debugger <https://docs.python.org/3/library/pdb.html>`_
+when a test fails so that you can move up and down the stac and inspect the
+value of individual variables.
+
+The test suite can be sped up by running tests in parallel. After installing
+the `pytest-xdist <https://pypi.org/project/pytest-xdist>`_ module
+(``pip install pytest-xdist``), tests can be run in parallel on several cores::
+
+  pytest -n auto
+
+will autoselect the number of cores, an explicit number can also be given
+(``pytest -n 4``).
+
 
 
 How do I ...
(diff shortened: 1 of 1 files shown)


docs/developer/index.rst
```diff
@@ -41,7 +41,47 @@ Sometimes a PR is still in development and known to fail the tests or
 simply does not touch any code, because it only modifies docstrings
 and the documentation. In that case, `[skip ci]` can be added to the
 commit message to prevent running the github actions tests to save
-time, energy, and limited free recources.
+time, energy, and limited recources.
```
DougBurke, line 44:
```suggestion
time, energy, and limited resources.
```

docs/developer/index.rst
```diff
@@ -41,7 +41,47 @@ Sometimes a PR is still in development and known to fail the tests or
...
+particularly useful option is to run only the tests in a specific file.
+For example, if you changed the code and the tests in the `sherpa.astro.ui`
+module, one might expect tests for this module to be the most likely to fail::
+
+  pytest sherpa/astro/ui/tests/test_astro_ui.py
+
+Once everything looks good, you can do a final run of the entire test suite. A
+second option useful for develoment is ``--pdb`` which drops into the
+`interactive Python debugger <https://docs.python.org/3/library/pdb.html>`_
+when a test fails so that you can move up and down the stac and inspect the
```
DougBurke, line 73:
``` suggestion
when a test fails so that you can move up and down the stack and inspect the
```

docs/developer/index.rst
```diff
@@ -41,7 +41,47 @@ Sometimes a PR is still in development and known to fail the tests or
...
+
+The test suite can be sped up by running tests in parallel. After installing
+the `pytest-xdist <https://pypi.org/project/pytest-xdist>`_ module
+(``pip install pytest-xdist``), tests can be run in parallel on several cores::
+
+  pytest -n auto
+
+will autoselect the number of cores, an explicit number can also be given
+(``pytest -n 4``).
+
```
DougBurke, line 84: Other things we could document include

- running with test coverage
- do we want to say "you can also run `sherpa_test` or smoke tests" - e.g. from https://sherpa.readthedocs.io/en/latest/install.html#testing-the-sherpa-installation - if you really want (but they should not provide any more testing than `pytest` does)

So I think I've talked myself out of the second one, but is it worth mentioning coverage?

DougBurke:
Interesting - why did the `[skip CI]` line get ignored (as the tests were run)?

docs/developer/index.rst
```diff
@@ -41,7 +41,47 @@ Sometimes a PR is still in development and known to fail the tests or
...
+-----------------
+Before you issue a pull request, we ask to run the test suite locally.
+Assuming everything is set up to install Sherpa from source, it can be
+installed in development mode with ``pip``::
+
+  pip install -e .
+
+"Development mode" means that the tests will pick up changes in the
+Python source files without running ``pip`` again (which can take some
+time). Only if you change the C code, you will have to explicitly run
```
DougBurke, line 56: Technically for us it's C++ and not C code, but I'm not sure how pedantic we should be here.

hamogu, line 56:
```suggestion
time). Only if you change the C++ code, you will have to explicitly run
```

docs/developer/index.rst
hamogu: I don't mind adding coverage, but I don't actually know how to run that locally. I guess I could learn it, so that I can document after. But maybe you know and can suggest a sentence (as a comment here is fine, I can paste it in)? I wanted to write these instructions to save time by writing down what I already know, not invest time in finding out about options I don't know yet.

DougBurke: For reference it is something like (I don't have a pristine environment to check I'm not missing something):

    pip install pytest-cov

and then

    pytest --cov sherpa

You then get coverage information out after the test summary and then you can call `coverage report` or `coverage html -d report` which will create `report/index.html` and friends. This may need a 'pip install coverage' or 'conda install coverge' [I'm not sure if this is picked up by installing `pytest-cov`]. However this is more for your benefit. For this we could just add (markup untested to see whether it's readable):

```
Test coverage can be included as part of the tests by installing the
`pytest-cov <https://pypi.org/project/pytest-cov/>`_ module
(``pip ninstall pytest-cov``) and then adding the ``-cov sherpa`` option::

  pytest -cov sherpa
```

hamogu: I don't have a pristine environment, but I definitely have one without coverage installed, so I'll try that right now and then update this.

DougBurke (approved):
LGTM

DougBurke:
@hamogu - just to note you can `pytest -n auto --cov sherpa` if you really want (I am not suggesting we add it to the docs).

hamogu:
Sure, and I can throw `--pdb` into the mix and some more. That's for the lined docs of the tools themselves. Here, listing what we think are the top 3 of useful options is more than enough, I agree.

PR #1170, opened 2021-05-29
Update Python and particularly number versions in documentation

hamogu:
# Summary
Just updating some of our docs pages that list dangerously old numpy versions.

# Details
While I did not test it, I very much doubt that numpy 1.6 is still sufficient to run Sherpa. At the very least, I'm certain that such an old numpy is not compatible with the Python version we require (3.7+). Note that I already updated the text to what we just decided for 4.14. Technically, there will be a 4.13.2 release before, but we say that older versions might work, so that's still correct, and we simply will forget to update this text again.

[skip ci]


DougBurke (commented):
There are a few other places we should mention Python 3.9 - README.md, docs/index.rst, not sure about others.

I am just going through the hoops with my first python 3.9 test run and things look okay (not got to the XSPEC build yet).

**EDITED TO ADD** yay, got an XSPEC+matplotlib+astropy+data files test run to pass with python 3.9. We should revour our CI (not for this PR)

hamogu:
I've searched for "3.7" to get all the instances in the text where it happens. However, that means that this PR cannot be merged yet. I was just updating the numpy and related references before, while now it says "is tested against 3.9", which is not true yet.

Marie-Terrell (approved):
There is one more change at the top of setup.py that we will need here before releasing. This is a python (3, 5) -> (3, 7) change (plus the error message that goes with it). However, this needs to stay for now while the CIAO Conda Python 3.6 builds (using setup.py) are still going. We have a change prepped off to the side to stop these builds, but we are trying to group this with another change going out soon. I will open another PR to make that change as soon as we can.

PR #1171, opened 2021-05-29
Improve the grouping/filtering tests

DougBurke:
# Summary

Add a number of tests for corner cases of filtering and grouping of PHA data.

# Details

This has been broken out of #1161. It is essentially all new tests which I developed to exercise changes I made in #1161 but are useful to have however we address things.

There are three code changes

- the removal of a duplicated field from the DataPHA class
- added a comment about an undocumented field in the DataPHA class
- fixed a bug in the OrderPlot class (used by `plot_order`); given we haven't really documented this option I don't think it's worth documenting this fix in the summary

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
@@ -32,6 +32,9 @@ import numpy as np
 
 import pytest
 
+from sherpa.utils.testing import requires_data, requires_fits, \
+    requires_plotting
+
 from sherpa.astro import ui
 from sherpa.astro.instrument import create_arf
 from sherpa.astro.data import DataRMF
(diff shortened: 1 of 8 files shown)


sherpa/astro/ui/tests/test_astro_ui_multi_response.py
```diff
@@ -452,3 +455,89 @@ def test_eval_multi_arfrmf_reorder(clean_astro_ui):
...
+    # Also compare to the fit plot (which is grouped)
+    #
+    assert fplot.modelplot.xlo[0] == pytest.approx(0.46720001101493835)
+    assert fplot.modelplot.xhi[-1] == pytest.approx(9.869600296020508)
+
+    # How does the overall model plot y values compare to the three
+    # orders?
+    #
+    # Unfortunately the selected groups are different so can not
+    # directly compare.
```
hamogu, line 540: Could ungroup for that...

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3672,3 +3672,45 @@ def test_set_ylog_bkg(plot, yscale, clean_astro_ui):
...
+    """
+
+    ui.load_pha(make_data_path('3c273.pi'))
+    ui.set_source(ui.powlaw1d.pl)
+
+    ui.plot_model()
+    ui.notice(0.5, 5)
+
+    f = ui.get_filter()
+    assert f == '0.518300011754:4.869099855423'
```
hamogu, line 3716: Not necessarily for this PR, because strong comparison of numbers are a little bit of a pain. However, in astropy tests like this have bitten us later when someone added tests on e.g. 32 bit systems, run the tests on a big endian cluster or any other case where rounding errors can be a little different from the 64 bit work station linux that we usually use.
So instead, one might use:
```
assert float(f.split(':')) == pytest.approx(0.518300011754)
```
I'm sure we have other places in the code where that happens, so we would have to fix that anyway if it comes up, but we can avoid adding new cases.

hamogu (approved):
Wow. A lot of new tests. Good! As usual, I'm reviewing that the tests make sense and are not actively harmful, not that they are complete and cover any possible case.

DougBurke:
Rebased because of test conflict.

Issue #1041, opened 2021-01-02
Why does `get_dep(filter=True)` return binned data, but `get_indep(filter=True)` does now?


hamogu:
I thought I could just use `get_dep(filter=True)` and `get_indep(filter=True)` to get the x,y for a plot, but the tow functions are curiously asymmetric in that one of them applies the grouping while the other one does not.

```
from sherpa.astro.data import DataPHA
dat = DataPHA('pha', np.arange(20), np.arange(20))

dat.group_width(2)
dat.notice(None, None)
dat.notice(5, 15)
dat.get_indep(filter=True), dat.get_dep(filter=True)
```

hamogu:
Notice the different length of the arrays.

DougBurke:
I find the handling of grouping by the `DataPHA` class inconsistent and not-at-all documented (if there is any documentation it is only stuff I've added to describe the current behavior, not the intent behind the code [*]). It's one of the reasons why we haven't made much headway with the existing grouping and filtering issues.

Note that the `filter` option for many of these functions defined in the base `Data` class (which makes sense), but that `DataPHA` also adds a `group` option, and I really don't understand how these two options interact (e.g. if one is `False` and the other is `True`).

[*] there may be some cases where there was original documentation, so it may be worth just checking the initial import to see what's changed since I could have been wrong in my documentation updates

PR #1173, opened 2021-05-31
Improve DataPha documentation

DougBurke:
# Summary

Minor improvements to the documentation of the DataPHA class.

# Details

Another PR from #1171. This updates some documentaion based on my understanding of the code, and points out some awkwardness in the class (_get_ebins and _get_indep are similar, but different enough they probably do need to be separate methods, but there is no existing design documentation to understand the architecture).

This PR does include one code change to _get_ebins which just clarifies the logic of an if/else chain.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1764,9 +1764,13 @@ class DataPHA(Data1D):
     def _get_ebins(self, response_id=None, group=True):
         """Return the low and high edges of the independent axis.
 
-        This method is badly named as it will return values in
-        either channel or energy units, depending on the units
-        setting and the associated response information.
+        This method is badly named as it will return values in either
+        channel or energy units, depending on the units setting and
+        the associated response information. It also returns the
+        approximation of the mapping from channel space to energy -
+        that is the E_MIN and E_MAX columns from the RMF EBOUNDS block
+        rather than from the ENERG_LO and ENERG_HI columns from the
+        MATRIX block.
 
         Parameters
         ----------
@@ -1786,6 +1790,10 @@ class DataPHA(Data1D):
             used. If the group flag is set and the data set is grouped
             then it uses the grouping settings, otherwise the data is
             for each channel. No filtering is applied.
+
+        See Also
+        --------
+        _get_indep
 
         Examples
         --------
@@ -1858,6 +1866,65 @@ class DataPHA(Data1D):
         return (self.channel,)
 
     def _get_indep(self, filter=False):
+        """Return the low and high edges of the independent axis.
+
+        Unlike _get_ebins, this returns values in the "native" space
+        of the response - i.e. for a RMF, it returns the bounds from
+        the MATRIX rather than EBOUNDS extension of the RMF - and not
+        the approximation used in _get_ebins.
+
+        Parameters
+        ----------
+        filter : bool, optional
+            It is not clear what this option means.
+
+        Returns
+        -------
+        lo, hi : ndarray
+            The low and high edges of each bin, in either keV or
+            Angstroms.
+
+        Raises
+        ------
+        sherpa.utils.err.DataErr
+            The data set does not contain a response.
+
+        See Also
+        --------
+        _get_ebins
+
+        Notes
+        -----
+        If the PHA file contains multiple responses then they aare
+        combined to create the overall grid.
+
+        Examples
+        --------
+
+        >>> pha.units = 'energy'
+        >>> elo, eho = pha._get_indep()
+        >>> elo.shape
+        (1090,)
+        >>> pha.channel.shape
+        (1024,)
+        >>> elo[0:5]
+        array([0.1 , 0.11, 0.12, 0.13, 0.14])
+        >>> ehi[0:5]
+        array([0.11      , 0.12      , 0.13      , 0.14      , 0.15000001])
+        >>> (elo[1:] == ehi[:-1]).all()
+        True
+
+        >>> pha.units = 'wave'
+        >>> wlo, who = pha._get_indep()
+        >>> wlo[0:4]
+        array([112.71289825, 103.32015848,  95.37245534,  88.56013348])
+        >>> whi[0:4]
+        array([123.98418555, 112.71289825, 103.32015848,  95.37245534])
+        >>> (wlo[:-1] == whi[1:]).all()
+        True
+
+        """
+
         if (self.bin_lo is not None) and (self.bin_hi is not None):
             elo = self.bin_lo
             ehi = self.bin_hi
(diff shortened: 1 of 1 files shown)


sherpa/astro/data.py
```diff
@@ -1761,27 +1761,87 @@ def get_specresp(self, filter=False):
...
+            default is to use the default response identifier.
+        group : bool, optional
+            Should the current grouping setting be applied. This is
+            only used if the "grouped" attribute is set.
+
+        Returns
+        -------
+        lo, hi : ndarray
+            The low and high edges of each bin, in either channels or
+            keV: energy is used if the units setting is not energy and
```
hamogu, line 1788: Is this really true: "energy is used if the units setting is not energy"? Maybe there is a "not" too many?

hamogu: Or it shoud say "is not channel" instead of "is not energy"?

sherpa/astro/data.py
```diff
@@ -1761,27 +1761,87 @@ def get_specresp(self, filter=False):
...
+        """
         group = bool_cast(group)
 
         if self.units == 'channel':
             elo = self.channel
             ehi = self.channel + 1
+        elif (self.bin_lo is not None) and (self.bin_hi is not None):
+            elo = self.bin_lo
+            ehi = self.bin_hi
+            if (elo[0] > elo[-1]) and (ehi[0] > ehi[-1]):
```
hamogu, line 1839: I wonder if we run into trouble with XMM/RGS here, which I've seen reverse-energy sorted (I think. Maybe I'm just confused. It's the end of a long holiday weekend.). I know that this is just the old code moved up by a few lines, so it's probably fine, but I'm not sure I can follow the logic in my head to make sure.
I guess the XMM case will end up in the else clause. But why are we guaranteed that `bin_lo` or `bin_hi` are `None`, so we know that we can get down to the next part of the logic?

sherpa/astro/data.py
```diff
@@ -1761,27 +1761,87 @@ def get_specresp(self, filter=False):
...
     def _get_ebins(self, response_id=None, group=True):
+        """Return the low and high edges of the independent axis.
+
+        This method is badly named as it will return values in either
+        channel or energy units, depending on the units setting and
+        the associated response information. It also returns the
+        approximation of the mapping from channel space to energy -
+        that is the E_MIN and E_MAX columns from the RMF EBOUNDS block
+        rather than from the ENERG_LO and ENERG_HI columns from the
+        MATRIX block.
```
hamogu, line 1773: But if no rmf is loaded, it takes the values from the arf (if present)

hamogu (changes requested):
At the minimum, there is small change in the docs that's needed (and extra "not"?). I  agree we can put it in like this because it is a lot better documented than it was before. 

But any work on this begs the question is the can (and should) clean this up. the `_get_ebins` is an underscore method we can remove at will. There is also #1041... 
I'm not trying to force that work into this PR - it is an improvement and I know @DougBurke is working on something else and just spun this off - but I also think the current way of doing it with two almost identical methods and broken options (`get_indep(filter=True)`) shoudl be cleaned up eventually.

hamogu:
Both methods are from "initial commit" 7 years ago, so there is little help in finding out what the intentions was from that side...

sherpa/astro/data.py
DougBurke: The code used to be

    if a: ... do a things
    else:
        set up arf, rmf variables
        if b: ... do b things and do not use arf or rmf variables
        elif c: ... do c things
        elif d: ... do d things
        else: ... do boring things

to

    if a: ... do a things
    elif b: ... do b things
    else:
        setup arf, rmf variables
        if c: ... do c things
        elif d: ... do things 
        else: ... do boring t hings

The important thing is that the `arf` and `rmf` variables are not used in the `a` or `b` checks (they are in `c` and `d`). We could move the `arf, rmf = ..` line to before this check and then this becomes a straight `if/elif/...` chain, which might be a bit easier to scan, but not sure it's worth it. For reference the `get_response` method (which returns `arf` and `rmf`) can be safely moved before this section (i.e. it's not going to fall over if condition a is True).

DougBurke:
The `_get_ebins` and `_get_indep` routines are - I have come to the conclusion after finding `_get_indep` last week, doing different things. Whether this means we need two methods or not is a different thing. However, it'll be a lot of work before we can think about cleaning this up. The `_get_indep` method isn't actually used in `sherpa.astro.data` so I tried removing it but then all the plot tests failed (I didn't review to see if there are other tests that could fail, i.e. it may be used in other places too). So this is just to say that while I do want to improve this code, I'm personally not likely to target this until we've resolved our filtering/grouping work.

sherpa/astro/data.py
DougBurke: This is one of those situations when properly describing the behavior ends up just repeating the code logic, so I didn't bother. If an ARF is present (but no RMF) then it uses the ENERG_LO/HI columns from the SPECRESP block. And technically the names MATRIX and SPECRESP aren't correct, as there are different variants. What I really wanted to say was that it's the ENERG_XX values and not E_XXX values (but, as you point out this is only meaningful for RMF responses).

sherpa/astro/data.py
DougBurke: Hopefully the re-worked version is better.

sherpa/astro/data.py
DougBurke: I've got around this by starting this sentence with the comment "If a RMF is loaded" (or words to that effect). This means that if we have an ARF-only analysis the reader is left clueless, but in that case we only have one way to specify energies.

hamogu:
Of course, there is also `get_indep` which is also almost the same. I'm convinced that we can roll that all into one public function with tests and even more options than we have now, but still be backwards compatible in the public method. But I also agree that this is not a priority and should definitely wait until the dust has settled on the other grouping and rebinning changes to avoid all sorts of tricky mix-ups.

hamogu (approved):

hamogu:
So, I'm approving this for what it's now, and I'll open a new issue for `get_indep`, `_get_indep`, and `_get_ebins` and copy over some of this discussion.

sherpa/astro/data.py
```diff
@@ -1810,6 +1871,65 @@ def get_indep(self, filter=True):
...
+        sherpa.utils.err.DataErr
+            The data set does not contain a response.
+
+        See Also
+        --------
+        _get_ebins
+
+        Notes
+        -----
+        If the PHA file contains multiple responses then they aare
```
wmclaugh, line 1903: aare -> are

DougBurke: This is the olde englishe spellinge.

DougBurke: Fixed. Ta.

PR #1175, opened 2021-06-02
Store UI contour plots in a dictionary, not directly

DougBurke:
This is now part of #1516 but I'm leaving this open so we can keep track of it.

# Summary

Change how the contour objects are stored internally.

# Details

This has been pulled out of #1089 as it is related to that work but can be reviewed separately.

This only affects users who are poking around in the internals of the session class, and it's all changes to "private" attributes so it is okay to change this.

The idea is to store the individual contour-plot objects in a dictionary rather than as separate attributes in the session class. This makes it easier to track from the contour data command to the object that is used by the command, and potentially makes it easier to extend in sub-classes of the Session class (although this is not currently used as we don't add any contour support in the sherpa.astro.ui module). This is a simpler version of an approach I plan to apply to the plot objects.

# Note

We now have both _contour_store and _contour_type_names storing essentially the same information. I had tried to remove _contour_type_names but it would interfere with #1089 so I have left this clean up once we can also update the plot types to a similar scheme . For plots we have multiple names that map to the same plot object which complicates things and, as we use the same code for handling both contour and plots, changes to one require either complicating the code - i.e. having separate paths for the two cases - or making the same change to both (which I don't want to do here).

### DIFF
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -335,7 +335,7 @@ class Session(NoNewAttributesAfterInit):
         # What objects are used to display the contour plots - e.g.
         # contour('data') or contour_data(). See also the
         # _contour_type_names dictionary which defines the actual
-        # mappnig from user command to the key of this dictionary
+        # mapping from user command to the key of this dictionary
         # (which, for contours, is the identity function).
         #
         self._contour_store = {}
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -332,14 +332,21 @@ class Session(NoNewAttributesAfterInit):
         self._traceplot = sherpa.plot.TracePlot()
         self._scatterplot = sherpa.plot.ScatterPlot()
 
-        self._datacontour = sherpa.plot.DataContour()
-        self._modelcontour = sherpa.plot.ModelContour()
-        self._sourcecontour = sherpa.plot.SourceContour()
-        self._fitcontour = sherpa.plot.FitContour()
-        self._residcontour = sherpa.plot.ResidContour()
-        self._ratiocontour = sherpa.plot.RatioContour()
-        self._psfcontour = sherpa.plot.PSFContour()
-        self._kernelcontour = sherpa.plot.PSFKernelContour()
+        # What objects are used to display the contour plots - e.g.
+        # contour('data') or contour_data(). See also the
+        # _contour_type_names dictionary which defines the actual
+        # mappnig from user command to the key of this dictionary
+        # (which, for contours, is the identity function).
+        #
+        self._contour_store = {}
+        self._contour_store['data'] = sherpa.plot.DataContour()
+        self._contour_store['model'] = sherpa.plot.ModelContour()
+        self._contour_store['source'] = sherpa.plot.SourceContour()
+        self._contour_store['fit'] = sherpa.plot.FitContour()
+        self._contour_store['resid'] = sherpa.plot.ResidContour()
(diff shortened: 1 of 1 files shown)


sherpa/ui/utils.py
```diff
@@ -332,14 +332,21 @@ def clean(self):
...
-        self._sourcecontour = sherpa.plot.SourceContour()
-        self._fitcontour = sherpa.plot.FitContour()
-        self._residcontour = sherpa.plot.ResidContour()
-        self._ratiocontour = sherpa.plot.RatioContour()
-        self._psfcontour = sherpa.plot.PSFContour()
-        self._kernelcontour = sherpa.plot.PSFKernelContour()
+        # What objects are used to display the contour plots - e.g.
+        # contour('data') or contour_data(). See also the
+        # _contour_type_names dictionary which defines the actual
+        # mappnig from user command to the key of this dictionary
```
hamogu, line 338:
```suggestion
        # mapping from user command to the key of this dictionary
```

hamogu (commented):
The changes are easy enough. However, since they (according to the text in the description) get us only half-way there, while a full solution interferes with something else I wonder that the benefit of merging this at this point would be vs waiting until it can all be addressed together.

DougBurke:
I strongly suggest for merging this now. Fixing the issue I noted about above requires a lot of changes in the plot code, which doesn't intersect with this PR, so waiting is just going to cause merge conflicts down the line.

hamogu (approved):
OK, that's fine. I just wanted to bring this up. But we do have a large number of open PRs and conflicts get more and more likely causing extra work when waiting for little benefit in this case.

wmclaugh:
I agree with Moritz's comment regarding getting 'half way there'. While it is more convenient to be able to put in smaller patches, it is cleaner to complete the changes as an atomic unit. Partial implementations leave often leave code inconsistent and can make things more difficult to track down maintain downstream. Too often,  code is left with "To do's" that are never done.

hamogu:
Only that Moritz has since been convinced that this is a reasonable step forward as is and not a case of half-assed implementation. So, @wmclaugh agrees with my original thoughts, but not with what I'm currently thinking.

DougBurke:
**RANT**

Waiting until it the fix can be done "in one shot" only works if we ever get around to reviewing patches in a timely manner, which to be honest **we fail at miserably**. One reason for this is that large PRs are hard to review. I now spend a lot of my time working on breaking up the PRs I make into more-digestable chunks [this is an example of such a PR] - which is generally a good idea, as it makes it easier to review, understand, and to point out whether it's even worth it. Unfortunately, that means that some changes are going to be left in a "not-as-good-as-they-could-be-in-an-ideal-world" state. I would *much rather* take a good, but not complete, fix than to end up spending all my time chasing rebases, which I have been doing for the last two years of working on Sherpa.

Unfortunately GitHub does not make it easy to have a set of related PRs so you can more-easily break up PRs into more-manageable chunks. I hear about GitLab's merge trains - https://about.gitlab.com/blog/2020/12/14/merge-trains-explained/ - as a possible solution, and that GitHub may be working on something similar, but I have no experience in the former and no knowledge of the latter.

hamogu:
I believe this should be merged as it is and I hope we can convince @wmclaugh of doing just that. Reopneing.

dtnguyen2 (dismissed):

DougBurke:
I have just rebased #1089 and it was frustrating. I am not sure what to do about this PR yet as it will just point out the issues I mentioned above in https://github.com/sherpa/sherpa/pull/1175#issuecomment-859556652

PR #1172, opened 2021-05-29
Improve data pha docs

DougBurke:
# Summary

Improve the documentation of the filtering and grouping of PHA data.

# Details

This has been pulled out of #1161. There's more to be done (e.g. add a section on DataIMG handling to the data area as an astronomy-specific file type) but I think this is an improvement as is (and the image stuff is definitely out-of-scope for this PR).

This PR really needs #1168 to be merged (to avoid some warnings from the sphinx documentation about links to the logging code). Note that #1168 is now merged.

PR #1177, opened 2021-06-02
minor rework of image handling for ui layer 

DougBurke:
# Summary

Improve the testing of DS9 from the UI layer and change how image objects (used to send data to DS9) are stored in the session class, following similar changes to plots and contours. It fixes a discrepancy between the image_model_component and get_model_component_image routines.

The sherpa.image module has been added to the documentation.

# Details

This extends the previous changes (eg #1175, #931, #906 and the WIP #1089) to the image set of commands. Fortunately, as with contours (#1175) it is an uneventful change - there are two main parts

1. rather than have self._fooimage, self._barimage, ... we have self._image_types as a dictionary with keys for foo, bar, ... which makes it easier to identify and find these objects
2. access to these objects is now always done by the get_foo_image class, which has the responsibility for calling it's prepare_image method correctly. This has two advantages in that we only ever have to get this logic right once per image type (see below for  a discussion of where we were not actually doing it right before), and it is much easier to over-ride in subclasses (although for the image case we don't actually do this)

The documentation changes are minimal - it really just adds the sherpa.image to the list of documented modules but does not add any description of the classes as I do not have time to do that here and it would be better done in a separate PR like #1172 does for the DataPHA class.

Tests are added to make sure these code paths are covered. We do have existing DS9 tests but they are at a lower level than these tests from the UI layer. In writing these tests I realized there was a bug in that image_model_component() and get_model_component_image() weren't using the same data (the image_mode_component was including any convolution component whereas get_model_component_image wasn't). The solution is to fix get_model_component_image and then make image_model_component use it, rather than having multiple ways of setting the data. The tests cover this case (so when the BUGFIX commit is hit one of the tests had to be changed).

Note that these tests can not reliably be combined with pytest-xdist, so

    pytest -n auto

may fail on these tests. This is because we only create a single DS9 instance which is then connected to by all the tests, resulting in potential corruption if two DS9 tests are run at the same time (the state is undefined in this case). We could look at supporting a per-instance DS9 connection, but that is a large change. For now, if you have DS9 and XPA installed then I suggest you have to run the tests serially. This is also (probably) true of the existing DS9 tests, it's just that no-one has really run into this combination before (and with more ds9 tests there's more chance for test collision). I've added a note to the pytest-xdist section of the developer docs.

### DIFF
--- a/sherpa/ui/tests/test_ui_image.py
+++ b/sherpa/ui/tests/test_ui_image.py
@@ -225,13 +225,14 @@ def test_get_model_component_image_with_convolution(session):
     assert obj.eqpos is None
     assert obj.sky is None
 
-    # Note that the convolution component is NOT applied
-    # to the data!
-    y = d.eval_model(m.parts[0]).reshape((9, 9))
+    # Note that the convolution component is applied
+    # to the data.
+    psf = s.get_model_component('psf')
+    y = d.eval_model(psf(m.parts[0])).reshape((9, 9))
     assert obj.y == pytest.approx(y)
 
     # sanity check
-    assert y[3, 3] == pytest.approx(100.0)
+    assert y[3, 3] == pytest.approx(42.25302)
 
 
 @requires_ds9
--- a/sherpa/ui/tests/test_ui_image.py
+++ b/sherpa/ui/tests/test_ui_image.py
@@ -18,9 +18,9 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-"""Basic tests of the miage functionality in sherpa.ui. The hope is that we
+"""Basic tests of the image functionality in sherpa.ui. The hope is that we
 can run some tests even without DS9/XPA available. Although tests are
-run sith both the base and Astro session classes, there's no tests of
+run with both the base and Astro session classes, there's no tests of
 Astro-specific functionality (e.g. DataIMG).
 
 Notes
--- a/sherpa/ui/tests/test_ui_image.py
+++ b/sherpa/ui/tests/test_ui_image.py
@@ -1,0 +1,692 @@
+#
+#  Copyright (C) 2021
+#  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


sherpa/ui/tests/test_ui_image.py
```diff
@@ -0,0 +1,693 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Basic tests of the miage functionality in sherpa.ui. The hope is that we
```
hamogu, line 21:
```suggestion
"""Basic tests of the image functionality in sherpa.ui. The hope is that we
```

(I also love to fix really irrelevant typos in docstrings for tests!

sherpa/ui/tests/test_ui_image.py
```diff
@@ -0,0 +1,693 @@
...
+can run some tests even without DS9/XPA available. Although tests are
+run sith both the base and Astro session classes, there's no tests of
```
hamogu, line 23:
```suggestion
run with both the base and Astro session classes, there's no tests of
```

hamogu (approved):
I really know nothing about XPA and talking to ds9. However, I can say that the changes in the Python code are good, shorten the files and makes them much easier to read. Since the ds9/xpa handling did not change, I also know that's still good. And the added tests are clearly an improvement; the only thing I cannot say for certain is if they are fully comprehensive or if more should be added. But this is an improvement either way and should be merged.

I marked up two totally inconsequential typos and I'm happy to merge this with or without fixing those.

sherpa/ui/tests/test_ui_image.py
DougBurke: It is a spectacularly-good typo.

sherpa/ui/tests/test_ui_image.py
DougBurke: What, no love for the dark side?

sherpa/ui/tests/test_ui_image.py
```diff
@@ -0,0 +1,693 @@
...
+Astro-specific functionality (e.g. DataIMG).
+
+Notes
+-----
+A number of tests send data to an external DS9 process (which will be
+started if needed) and then check the values displayed by the process.
+These tests are highly-likely to fail if the tests are run in
+parallel, such as with
+
+    pytest -n auto
```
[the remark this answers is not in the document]
DougBurke: I couldn't work out how to add either a "these tests must be run in serial" (which doesn't really fit into the xdist scheme) or a "skip if run in parallel" case.

PR #1178, opened 2021-06-03
Define physical constant hc in fewer places

hamogu:
# Summary
hc (as in "Planck constant * speed of light") is a physical constant that does not depend on the DataPHA and thus it should not be defined as a class attribute. 

(Is a summary really needed for this change?)

# Details
hc was actually defined in more than one place (and it's still separate in one test file) and in particular there is no need for it to be an hidden (`_hc`) attribute to the DataPHA class. I need this change for other work to avoid circular imports (DataPHA is imported into instrument.py purely to access `DataPHA._hc` and I want to remove that import), but I'm pulling it out as a separate PR because it's ready already and I believe this is a good change in any case.

### DIFF
--- a/sherpa/astro/__init__.py
+++ b/sherpa/astro/__init__.py
@@ -16,6 +16,14 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+'''Physical constants for astronomical spectra
+
+This module contains some physical constants needed to perform conversions
+that are common in the `sherpa.astro` module, e.g. from energy to wavelength.
+'''
 
 hc = 12.39841874  # nist.gov in [keV-Angstrom]
-charge_e = 1.60217653e-09  # elementary charge [ergs] 1 keV, nist.gov
+"""The product of Plank constant (h) and speed of light (c) using values from nist.gov in units appropriate for converting between keV and Angstroms."""
+
+charge_e = 1.60217653e-09
+'''elementary charge from nist.gov in units appropriate for converting between keV and erg.'''
--- a/sherpa/astro/__init__.py
+++ b/sherpa/astro/__init__.py
@@ -23,7 +23,7 @@ that are common in the `sherpa.astro` module, e.g. from energy to wavelength.
 '''
 
 hc = 12.39841874  # nist.gov in [keV-Angstrom]
-"""The product of Plank constant (h) and speed of light (c) using values from nist.gov in units appropriate for converting between keV and Angstroms."""
+"""The product of Planck constant (h) and speed of light (c) using values from nist.gov in units appropriate for converting between keV and Angstroms."""
 
 charge_e = 1.60217653e-09
 '''elementary charge from nist.gov in units appropriate for converting between keV and erg.'''
--- a/sherpa/astro/__init__.py
+++ b/sherpa/astro/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2007, 2021  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


hamogu:
Also I notice that @wmclaugh is merging stuff (which is good!) but as a sport I'm of course trying to keep more than 40 open PRs ;-)

DougBurke:
Perhaps more a background note than anything, but we also define these values in C++ scope - see https://github.com/sherpa/sherpa/blob/b09626f0267b7c0be9691b08bd70d0dec534483d/sherpa/include/sherpa/constants.hh#L83-L122

We could expose these from sherpa.utils._utils and then re-export them from sherpa.utils so we ony define them in one place, but is it worth the complexity versus a comment saying we must ensure these values match?

hamogu:
I saw them there, too, and decided not to bother to export to Python.  They hopefully match now and they won't be changed ever.

sherpa/astro/__init__.py
```diff
@@ -17,4 +17,5 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-# Nothing in here yet
+hc = 12.39841874  # nist.gov in [keV-Angstrom]
```
DougBurke, line 20: Can we add a docstring after these rather than a comment - e.g.

```
hc = silly-number
"""The product of h * c using values from nist.gov appropriate for converting between keV and Angstroms."""
```

(but better written).

We will then get this included in the documentation if we ever add sherpa.astro to the documentation. Which we probably should.

hamogu:
Added docs. Not 100% happy with where they are, but it's the best place to put them right now.

DougBurke:
Yay - https://sherpa--1178.org.readthedocs.build/en/1178/overview/astro.html

DougBurke:
Looking at the coverage reports reminds me that we need to create or come up with test cases that exercise these uncommon options. Definitely not for this PR.

DougBurke (approved):
LGTM

sherpa/astro/__init__.py
```diff
@@ -16,5 +16,14 @@
...
 #
+'''Physical constants for astronomical spectra
 
-# Nothing in here yet
+This module contains some physical constants needed to perform conversions
+that are common in the `sherpa.astro` module, e.g. from energy to wavelength.
+'''
+
+hc = 12.39841874  # nist.gov in [keV-Angstrom]
+"""The product of Plank constant (h) and speed of light (c) using values from nist.gov in units appropriate for converting between keV and Angstroms."""
```
wmclaugh, line 26: Plank -> Planck

sherpa/astro/__init__.py
```diff
@@ -16,5 +16,14 @@
...
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+'''Physical constants for astronomical spectra
 
-# Nothing in here yet
+This module contains some physical constants needed to perform conversions
+that are common in the `sherpa.astro` module, e.g. from energy to wavelength.
+'''
+
+hc = 12.39841874  # nist.gov in [keV-Angstrom]
```
wmclaugh, line 25: it would be good to cite the actual nist.gov (or other source) where the exact constant value used was pulled from as within nist.gov there may be various pages w/ slightly different precision values.

hamogu: I agree. That would be good. But I did not look this up in nist.gov I merely moved code that we had before to a different location and unfortunately I don't know who pulled that constant in an where he or she found it because this pre-dates the move to git: https://github.com/sherpa/sherpa/blame/572231b222136fa829590be9523da0d529328bc6/sherpa/astro/utils/__init__.py#L47

sherpa/astro/__init__.py
hamogu, line 26:
```suggestion
"""The product of Planck constant (h) and speed of light (c) using values from nist.gov in units appropriate for converting between keV and Angstroms."""
```

sherpa/astro/__init__.py
wmclaugh: I dug through clearcase and found that the constant was modified on 05/30/08 and changed from:

       _hc = 12.398419058  # h*c in keV-Angstrom
to:
       _hc = 12.39841874  # nist.gov in [keV-Angstrom]

release notes posted to sherpadev included the note:

     o Adjusted hc constant in sherpa/astro/data.py to nist.gov standard.

hamogu: @wmclaugh Thanks for checking. I suggest we just leave it at the current value with the current comment.

wmclaugh (approved):

Issue #829, opened 2020-06-18
XSPEC model issues with the regrid code


DougBurke:
[I am sure I did some testing with the XSPEC models, but it may have been with some low-level code, since there are two issues I've found testing #798]

1) if you call `.regrid` on an XSPEC model to create a "regridded" model then it will error out when you try to evaluate it with an error like

```TypeError: xsphab() takes no keyword arguments```

2) the XSPEC models should be treated as integrated for the `XSAdditiveModel` cases whether sent an energy grid or elo,ehi arrays. At the moment the first version is treated as an interpolation, and so does not work well. We may have to do something "clever" about `XSMultiplicativeModel` (in the opposite direction; ie always treat as interpolation and not binned)

---

I have a simple fix for issue 1 - change the `calc` method of `XSModel` - which I think makes sense to do:

```
-        out = self._calc(*args, **kwargs)
+        #
+        # The XSPEC functions don't expect any extra arguments (at least
+        # as they are coded here), so we just drop any. This isn't
+        # ideal.
+        #
+        # out = self._calc(*args, **kwargs)
+        out = self._calc(*args)
```

For issue 2, I think we should be able to add an override to `regrid` in the `XSAdditiveModel` class which can handle this. I need to think about the `XSMultiplicativeModel` class (and the proposed `XSConvolutionKernel/Model` classes from #750)

DougBurke:
For reference I show a multiplicative model (top) and an additive model (bottom) evaluated directly in blue, using rebinning (orange), and interpolation (green). Note that the orange and green data use a different grid than the blue line (it is one with half the bin size and starts/ends slightly inside the end range of the blue line). The gray line is labelled "truth" as it is the model evaluated on this second grid. 

![ex](https://user-images.githubusercontent.com/224638/123867058-deb92a80-d8fb-11eb-9487-15490c4f3c1a.png)


It's hard to see for the top plot - the multiplicative model - but in this case the interpolation scheme (green) is the better match to the "truth". For the additive case it is actually the rebin scheme (orange) that more-closely matches the "truth". This makes sense if you think about how the multiplicative and additive models are defined - the former calculates a "scaling" value for each bin whereas the latter calculates photons/cm^2/s for each bin. It does mean that we need to expand the regrid code to support this.

The code to generate this is below:

```python
import numpy as np
from matplotlib import pyplot as plt

from sherpa.utils._utils import rebin
from sherpa.utils.akima import akima
[...]

PR #1179, opened 2021-06-03
Improve handling of the integrate flag for models

DougBurke:
I'm coming back to review this and so the text below is not guaranteed to match up with the code behavior.

# Summary

Improve the handling of the integrate setting for models: it is now passed on for composite models, such as an expression like "mdl1 * (mdl2 - mdl3)", and XSPEC models no longer allow users to change the setting (any attempt will be ignored, which will allow scripts created by save_all using an older version of Sherpa to still work).

# Details

This is a rework of the last part of #947 and then merging in work done in #830. This is closely related to the regrid work but has been broken out of that (see #830, which depends on this PR). This PR depends on

- #1803

The ideas are that

- adding more tests to cover some of the issues we have uncovered
- allow the integrate to be passed "up" the model classes, so that composite models can have an integrate setting [1]
- special case the XSPEC models since we know that should send the data "integrated" arrays - that is low and high edges - **but** we also know there's a difference for additive and multiplicative models that maps well to `integrate=True` and `integrate=False` respectively when regridding data [2], [3]

[1] technically it makes to copy the integrate flag for something like an unary op model, but it's less clear what happens for binary op models where you may be combing models with `integrate=True` and `integrate=False`. We could set `integrate` to `None` in this case (or even record all the integrate settings in some way) but I am not sure it would help downstream users. I think the current code will fall over to a `True` setting for this case (see the paragraph starting "One interesting outcome" below).

[2] XSPEC models now have a fixed `integrate` setting, and attempting to change it does nothing. That is, you can say "mdl.integrate = False` for an additive model, but the field will stay set to `True`. I originally made this be an error, **but** that would break any use of `save_all` which contains a multiplicative XSPEC model, for no actual user benefit. So now we just do nothing. We could use the warning logger, or a warnings FutureWarning to create a message, but I am not sure many people actually change this setting, so all we'd then do would be make users of `save_all` see even more screen messages they would then learn to ignore.

[3] XSPEC convolution models are a special case - I believe the way that XSPEC works, you aren't able to apply a convolution model to anything but an additive model (that is, a model expression that "ends" with an additive component), which would imply XSPEC convolution models should be marked as `integrate=True`, so this is what we do

One interesting outcome of the current code, and something we could try to clean up, is that models tend now to default to `integrate=True` (thanks to `ArithmeticModel`) and that this can lead to things like "unary op model is sent something with no `integrate` setting, so we try not to set the `integrate` flag, but it magically gains it from the `ArithmeticModel` hierarchy".

It is hard to update the documentation to highlight this change, as there's no single `ui` routine (for example) we can document. Something to think about.

This code is going to conflict with all the existing XSPEC changes we have (e.g. #1396 https://github.com/sherpa/sherpa/pull/1609 https://github.com/sherpa/sherpa/pull/1615 https://github.com/sherpa/sherpa/pull/1630)

# Question

How useful is the `integrate` setting? In some ways I'd like to get rid of it, but that would be quite a large change for little obvious user benefit.

We could use the integrate setting to validate model expressions, in a similar manner to the way we do with model dimensionaluity, but

 - the dimensional check can be made when the model expression is being created, whereas the integrate setting would need to be checked whenever the composite model is evaluated
 - for XSPEC models we can come up with something (like you can multiply but not add multiplicative components) but is this true for the general case (and coming up with rules is more onerous than I'd like for a simple check)

# Trying to change the integrate setting

In #947 I had made code like

    mdl = XSapec()
    mdl.integrate = False

raise an error, because the integrate setting of this model (an additive model) is True. In this version it just ignores the user setting, so the following would hold after the above:

    assert mdl.integrate

Normally I'd be all for raising an error here, but it would make any script saved by `save_all` that contains an XSPEC multiplicative model to fail when evaluated because before this change the file would include a line `mdl.integrate = True` for these models  (because prior to this PR the multiplicative models did have integrate set to True), which would mean annoying changes for users that doesn't actually improve anything.  Hence the change to the setter just ignoring the user value.

# Example

## Generic

In CIAO 4.15 we have

```
sherpa-4.15.0> mdl1 = 2 + const1d.foo1

sherpa-4.15.0> mdl2 = 2 + scale1d.bar1

sherpa-4.15.0> foo1.integrate
True

sherpa-4.15.0> bar1.integrate
False

sherpa-4.15.0> mdl1.integrate
AttributeError: 'BinaryOpModel' object has no attribute 'integrate'

sherpa-4.15.0> mdl2.integrate
AttributeError: 'BinaryOpModel' object has no attribute 'integrate'

sherpa-4.15.0> tmp = -bar1

sherpa-4.15.0> tmp.integrate
AttributeError: 'UnaryOpModel' object has no attribute 'integrate'
```

With this PR we get

```
>>> mdl1 = 2 + const1d.foo1
>>> mdl2 = 2 + scale1d.bar1
>>> (foo1.integrate, bar1.integrate)
(True, False)
>>> (mdl1.integrate, mdl2.integrate)
(True, True)
>>> tmp = -bar1
>>> tmp.integrate
False
```

Now, should `2 + bar1` have an integrate setting of `True`? Ideally not.

I note that when combining "actual" models we get the `integrate` setting combined as you might expect:

```
>>> foo1.integrate, bar1.integrate
(True, False)
>>> (foo1 + bar1).integrate
True
>>> (foo1 + foo1).integrate
True
>>> (bar1 + bar1).integrate
False
>>> (bar1 + foo1).integrate
True
```

[as a reminder, in CIAO 4.15 composite models do not have an `integrate` setting:

```
sherpa-4.15.0> (foo1 + bar1).integrate
AttributeError: 'BinaryOpModel' object has no attribute 'integrate'
```
]

## XSPEC

In CIAO 4.15 we have

```\sherpa-4.15.0> mdl = xsphabs.gal * xsapec.src

sherpa-4.15.0> (gal.integrate, src.integrate)
(True, True)

sherpa-4.15.0> mdl.integrate
AttributeError: 'BinaryOpModel' object has no attribute 'integrate'

sherpa-4.15.0> src.integrate = False

sherpa-4.15.0> (gal.integrate, src.integrate)
(True, False)
```

We now get

```
>>> mdl = xsphabs.gal * xsapec.src
>>> (gal.integrate, src.integrate)
(False, True)
>>> mdl.integrate
True
>>> (gal + gal).integrate     # NOTE: combing two False create a combined model with integrate=False
False
>>> src.integrate = False
>>> (gal.integrate, src.integrate)
(False, True)
>>> mdl.integrate
True
```

### DIFF
--- a/sherpa/astro/ui/tests/test_serialize.py
+++ b/sherpa/astro/ui/tests/test_serialize.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2015, 2016, 2018, 2019, 2020, 2021
+#  Copyright (C) 2015, 2016, 2018, 2019, 2020, 2021, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -329,7 +329,7 @@ pl.ampl.units   = ""
 pl.ampl.frozen  = False
 
 create_model_component("xsphabs", "gal")
-gal.integrate = True
+gal.integrate = False
 
 gal.nH.default_val = 1.0
 gal.nH.default_min = 0.0
@@ -460,7 +460,7 @@ gpl.ampl.units   = ""
 gpl.ampl.frozen  = False
 
 create_model_component("xsphabs", "ggal")
-ggal.integrate = True
+ggal.integrate = False
 
 ggal.nH.default_val = 2.0
 ggal.nH.default_min = 0.0
@@ -582,7 +582,7 @@ gpl.ampl.units   = ""
 gpl.ampl.frozen  = False
 
 create_model_component("xsphabs", "ggal")
-ggal.integrate = True
+ggal.integrate = False
 
 ggal.nH.default_val = 2.0
 ggal.nH.default_min = 0.0
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1117,9 +1117,19 @@ class XSModel(RegriddableModel1D, metaclass=ModelMeta):
     represents the upper edge of the last bin. This means that for
     an input grid of ``n`` points, the returned array will contain
     ``n`` values, but the last element will be zero.
+
+    The ``integrate`` setting is fixed for XSPEC models: it is set to
+    ``True`` for Additive models and ``False`` for Multiplicative
+    models. Convolution models are expected to be used with "additive"
+    models (or a mixture of multiplicative and additive models) and
+    so they are set to ``True``. The integrate setting is not used
+    to determine how to evaluate the models, but it is used by the
+    RegriddableModel1D class to determine how to rebin the data.
+
     """
 
     version_enabled = True
+    _integrate = True
 
     def __init__(self, name, pars):
 
@@ -1151,6 +1161,24 @@ class XSModel(RegriddableModel1D, metaclass=ModelMeta):
 
         super().__init__(name, pars)
 
+    @property
+    def integrate(self):
+        """The integrate setting.
+
+        It can not be changed for XSPEC models.
+        """
+        return self._integrate
+
+    @integrate.setter
+    def integrate(self, newval):
+        # We could raise an error if the setting is changed, but there
+        # are save files created before we made the integrate
+        # un-changeable that would fail for no good reason. So we just
+        # ignore any request to change the setting. It is possible for
+        # the user to change _integrate but that is undefined
+        # behavior.
+        #
+        pass
 
     @modelCacher1d
     def calc(self, p, *args, **kwargs):
@@ -1307,6 +1335,9 @@ class XSTableModel(XSModel):
                                   hugeval)
             pars.append(self.norm)
 
+        # The integrate setting is the same as the addmodel parameter.
+        #
+        self._integrate = addmodel
         XSModel.__init__(self, name, pars)
 
     def fold(*args, **kwargs):
@@ -1367,7 +1398,7 @@ class XSMultiplicativeModel(XSModel):
 
     """
 
-    pass
+    _integrate = False
 
 
 class XSConvolutionKernel(XSModel):
@@ -1446,8 +1477,7 @@ class XSConvolutionKernel(XSModel):
     """
 
     def __repr__(self):
-        return "<{} kernel instance '{}'>".format(type(self).__name__,
-                                                  self.name)
+        return f"<{type(self).__name__} kernel instance '{self.name}'>"
 
     def __call__(self, model):
         return XSConvolutionModel(model, self)
@@ -1585,9 +1615,7 @@ class XSConvolutionModel(CompositeModel, XSModel):
     def __init__(self, model, wrapper):
         self.model = self.wrapobj(model)
         self.wrapper = wrapper
-        CompositeModel.__init__(self,
-                                "{}({})".format(self.wrapper.name,
-                                                self.model.name),
+        CompositeModel.__init__(self, f"{self.wrapper.name}({self.model.name})",
                                 (self.wrapper, self.model))
 
     # for now this is not cached
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1201,6 +1201,14 @@ class XSModel(RegriddableModel1D, metaclass=ModelMeta):
             warnings.warn(emsg, FutureWarning)
             # raise TypeError(emsg)
 
+        # The XSPEC code does not recognize the integrate flag.
+        # In fact, should we remove all keywords here?
+        #
+        try:
+            del kwargs["integrate"]
+        except KeyError:
+            pass
+
         # Ensure output is finite (Keith Arnaud mentioned that XSPEC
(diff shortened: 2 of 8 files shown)


hamogu:
I'm very opposed to just silently ignoring the setting. And, I've never used `save_all`. Maybe I should, I just never knew about it.
Anyway, a compromise could be to issue a warning "Setting the integrate setting on XSPEC models has no no effect. See URL HERE or release notes for details" or "This warning can be safely ignored, but don't do this from now on". Essentially, a Deprecation warning. Don't say "Hah. Error. Got you." Instead say "Setting integrate does not do anything, but we've always ignored that in the past. Now you at least get this warning and in the future it will be an error."

You could even make that an error for Additive models that don't have the `save_all` backwards compatibility problem.

hamogu:
Apart from the discussion on "should there be a message if I try to change it", the implementation seems uncontroversial. Since the setting is not used, cannot be changed and, for some models, also does not make any sense, I just want to ask: Do we really need an integrate setting at all? Or, if we need it, can be set is to `None` or the string "unused" so signify that it's in fact, not used.

DougBurke:
So, I would like to remove the `integrate` setting - which I believe would mainly be an issue for people using the Const*D models who should be using Scale*D instead (but our documentation points them to use the const version which we should address outside this PR).

However, I also want a way for us to identify whether a model behaves like the XSPEC additive versus multiplicative models, which I believe is important for us to get the regrid code working (and is the hidden driver behind this work). For that I think we need some way to say "hey, are you actually integrating the model or not" and I am co-opting the under-defined integrate setting to provide that. Perhaps they should be separate as this avoids some of this confusion (ie do we need to worry about existing code and the  integrate setting).

hamogu:
Either way works for me. I'd be OK with having as is and just adding an Exception/Warning for the cases where there is no backwards-compatibility issues (the Arithmetic models). We could then remove it two years from now.
Or, we could just add a warning for all and make `read_all` (is that the opposite of `save_all`?) just silence that warning. That way tou cna use te setting in any way way you like, I get the warning I want, and users of `save_all` are not unduly burdened with meaningless messages.

anetasie:
> So, I would like to remove the `integrate` setting - which I believe would mainly be an issue for people using the Const_D models who should be using Scale_D instead (but our documentation points them to use the const version which we should address outside this PR).
> 
> However, I also want a way for us to identify whether a model behaves like the XSPEC additive versus multiplicative models, which I believe is important for us to get the regrid code working (and is the hidden driver behind this work). For that I think we need some way to say "hey, are you actually integrating the model or not" and I am co-opting the under-defined integrate setting to provide that. Perhaps they should be separate as this avoids some of this confusion (ie do we need to worry about existing code and the integrate setting).**EDITED TO ADD** which doesn't really change this PR as such, but if we do decide to remove the integrate setting then I think we'd allow the integrate setting but warn users if they try to set it.

@DougBurke Do you want to remove the 'integrate' setting for all the models or only for XSPEC models?

DougBurke:
If we were to do it it would be for all models. However, as I have discussed above there are concerns I have over the regrid code - e.g. https://github.com/sherpa/sherpa/issues/829 - and so we may need some way of indicating how a model works when given lo/hi bins (is it integrating the data or not). XSPEC models have this by design (we have the XSAdditiveModel and XSMultiplicativeModel) but we need something for all models (whether it's a class or an attribute). I'm trying to get #830 to a place where we can discuss this but it's one of those annoying issues which doesn't really lend itself well to single PRs.

DougBurke:
I'm going to make this a draft PR for now, just to make sure it doesn't accidentally get merged.

anetasie:
If this is for all the models then it is quite a large change to our handling of the model evaluations. There is no distinction between multiplicative and additive models in Sherpa now, so each model could be combined in a general expression not dependent on the data type. If the integration is set by data then the multiplications or convolution expressions will be impacted by this change. As you described above this update requires more thoughts.

DougBurke:
@anetasie - it doesn't change how you combine models, or use convolution-style kernels, or how models work. There are two related but separate issues:

1. does changing the integrate flag make sense - I don't think people make use of the integrate setting: in part because we don't really advertise it but in part because I don't think it's that useful to be able to change and we have other ways we could handle if fir users who do want to switch the integrate setting
2. how does `regrid` work work for data evaluated on an integrated grid (which is #829)

anetasie:
@DougBurke I need to go back and check these other related issues.

DougBurke:
Rebased because of conflicts.

hamogu:
This is still marked as WIP, but it also has a request to review for me attached. So, just checking: Is this just open "for reference" as other PRS (like #1179) take charge of those changes or is there any actualy review here needed by me at this point?

DougBurke:
I don't think it needs a review at this time. I have a document I'm slowly working on trying to explain why I care about the integrate flag and what it has to do with the regrid code. It's slow progress, but probably worth SDS reviewing this before any PRs.

DougBurke:
Rebased as multiple things have changed, in particular #1617 has merged. Need to get back into this as it's not clear what is going on here...

PR #1180, opened 2021-06-07
Minor improvements to model/parameter documentation

DougBurke:
# Summary

Minor fixes and improvements to the documentation (both docstrings and RTD) for the model- and parameter-related code.

# Details

I would like to be able to better describe the regrid code, in particular, but I don't have time to investigate what the code is actually doing, so these additions and changes are not particularly deep.

This also fixes a minor warning from the RTD build due to #1135 .

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -125,6 +125,47 @@ parameter:
     >>> m2.FWHM = 10
     >>> m2.FwHm = 10
 
+Linking parameters
+------------------
+
+One parameter can be made to reference one or more other parameters, a
+process called "linking". The lniked is no-longer considered a free
+parameter in a fit since it's value is derived from the other
+parameters. This link can be a simple one-to-one case, such as
+ensuring the fwhm parameter of one model is the same as the other:
+
+    >>> m2.fwhm = m1.fwhm
+
+It can be more complex, such as ensuring the position of one line
+is a fixed distance from another:
+
+    >>> l2.pos = l1.pos + 23.4
+
+It can even include multiple parameters:
+
+    >>> l3.ampl = (l1.ampl + l2.ampl) / 2
+
+Requesting the parameter value will return the evaluated expression,
+and the expression is stored in the link attribute:
+
+    >>> l1.ampl = 10
+    >>> l2.ampl = 12
+    >>> l3.ampl.val
+    11.0
+    >>> l3.ampl.link
+    <BinaryOpParameter '((l1.ampl + l2.ampl) / 2)'>
+
+The string representation of the model changes for linked parameters
+to indicate the expression:
+
+    >>> print(l3)
+    l3
+       Param        Type          Value          Min          Max      Units
+       -----        ----          -----          ---          ---      -----
+       l3.fwhm      thawed           10  1.17549e-38  3.40282e+38
+       l3.pos       thawed            0 -3.40282e+38  3.40282e+38
+       l3.ampl      linked           11 expr: ((l1.ampl + l2.ampl) / 2)
+
 Model evaluation
 ================
 
@@ -202,6 +243,37 @@ The parameter order matches the pars attribute of the model:
 
     >>> print([p.fullname for p in mdl.pars])
     ['gauss1d.fwhm', 'gauss1d.pos', 'gauss1d.ampl']
+
+Model expressions
+=================
+
+The CompositeModel class is the base class for creating model
+expressions - that is the overall model that is combined of one or
+more model objects along with possible numeric terms, such as a
+model containing two gaussians and a polynomial:
+
+    >>> from sherpa.models.basic import Gauss1D, Polynom1D
+    >>> l1 = Gauss1D('l1')
+    >>> l2 = Gauss1D('l2')
+    >>> l1.pos = 5
+    >>> l2.pos = 20
+    >>> l2.ampl = l1.ampl
+    >>> c = Polynom1D('c')
+    >>> mdl = l1 + (0.5 * l2) + c
+
+The resulting model can be evaluated just like an individual
+component:
+
+    >>> x = np.arange(-10, 40, 2)
+    >>> y = mdl(x)
+
+This model is written so that the amplitude of the `l2` component is
+half the `l1` component by linking the two `ampl` parameters and then
+including a scaling factor in the model expression for `l2`. An
+alternative would have been to include this scaling factor in the link
+expression:
+
+    >>> l2.ampl = l1.ampl / 2
 
 Model cache
 ===========
@@ -716,6 +788,9 @@ class Model(NoNewAttributesAfterInit):
 class CompositeModel(Model):
     """Represent a model with composite parts.
 
+    This is the base class for representing expressions that combine
+    multiple models and values.
+
     Parameters
     ----------
     name : str
@@ -726,6 +801,27 @@ class CompositeModel(Model):
     Attributes
     ----------
     parts : sequence of Model
+
+    Notes
+    -----
+    Composite models can be iterated through to find their
+    components:
+
+       >>> l1 = Gauss1D('l1')
+       >>> l2 = Gauss1D('l2')
+       >>> b = Polynom1D('b')
+       >>> mdl = l1 + (0.5 * l2) + b
+       >>> mdl
+       <BinaryOpModel model instance '((l1 + (0.5 * l2)) + polynom1d)'>
+       >>> for cpt in mdl:
+       ...     print(type(c))
+       ...
+       <class 'BinaryOpModel'>
+       <class 'sherpa.models.basic.Gauss1D'>
+       <class 'BinaryOpModel'>
+       <class 'ArithmeticConstantModel'>
+       <class 'sherpa.models.basic.Gauss1D'>
+       <class 'sherpa.models.basic.Polynom1D'>
 
     """
 
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -114,6 +114,17 @@ attribute as you can say:
 
     >>> m2.fwhm = 20
 
+Accessing parameter values
+==========================
+
+The model class is set up so that any attribute access is case
+insensitive, so the following are all ways to change the ``fwhm``
+parameter:
+
+    >>> m2.fwhm = 10
+    >>> m2.FWHM = 10
+    >>> m2.FwHm = 10
+
 Model evaluation
 ----------------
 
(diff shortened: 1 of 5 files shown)


sherpa/models/parameter.py
```diff
@@ -525,6 +616,20 @@ def eval(self):
...
 class UnaryOpParameter(CompositeParameter):
+    """Apply an operator to a parameter expression.
+
+    Parameters
+    ----------
+    arg : Parameter instance
+    op : function reference
+        The ufunc to apply to the parameter value.
+    opstr : str
+        The symbol used to represemt the operator.
```
wmclaugh, line 627: represemt -> represent

sherpa/models/parameter.py
```diff
@@ -538,6 +643,23 @@ def eval(self):
...
+    Parameters
+    ----------
+    lhs : Parameter instance
+        The left-hand side of the expression.
+    rhs : Parameter instance
+        The right-hand side of the expression.
+    op : function reference
+        The ufunc to apply to the two parameter values.
+    opstr : str
+        The symbol used to represemt the operator.
```
wmclaugh, line 657: represemt -> represent

sherpa/models/regrid.py
```diff
@@ -460,22 +371,18 @@ def overlaps(self, other):
...
-        """
-        Return the grid representation of this dataset. The grid is always a tuple, even if the
-        dataset is 1-D and not integrated. This is due to the existing architecture of Sherpa's
-        model classes and the fact that there is no signature difference among 1-D and 2-D models,
-        as 1-D models can receive 1 or 2 arrays and 2-D models can receive 2 or 4 arrays.
+        """The grid representation of the space.
 
-        The x and y arrays in the grid are one-dimentional representations of the meshgrid obtained
-        from the x and y axis arrays, as in `numpy.meshgrid(x, y)[0].ravel()`
+        The x and y arrays in the grid are one-dimentional
```
wmclaugh, line 376: one-dimentional -> one-dimensional

DougBurke: At last: a typo that was not mine but in the original text.

sherpa/models/model.py
```diff
@@ -642,22 +666,40 @@ def _set_thawed_pars_maxes(self, vals):
                 p._max = v
 
     thawedparmaxes = property(_get_thawed_par_maxes, _set_thawed_pars_maxes,
-                              doc='Access to the maximum limits for the thawed parameters')
+                              doc='The maximum limits of the thawed parameters.\n\n' +
```
hamogu, line 669: While you are at it, I think it would be worthwhile to at least mention the difference between limits and hard limits. No deep discussion, but one line of text would be good either here or in the hardparmin/max or both.

sherpa/models/model.py
```diff
@@ -642,22 +666,40 @@ def _set_thawed_pars_maxes(self, vals):
...
+                                  doc='The hard maximum values for the thawed parameters.\n\n' +
+                                  'See Also\n' +
+                                  '--------\n' +
+                                  'thawedparhardmins, thawedparmaxes\n')
 
     def reset(self):
-        """Reset the parameter values."""
+        """Reset the parameter values.
+
+        Restores each parameter to the last value it was set to.
```
hamogu, line 699: it was *manually* set to? It was *explicitly* set to? it was set to by calling `model.parname = x`? I know what you mean, but running fit also sets the parameters (that's why I can access them as `model.FWHM` after the fit)

sherpa/models/regrid.py
```diff
@@ -311,13 +244,12 @@ def overlaps(self, other):
         return self.x_axis.overlaps(other.x_axis)
 
     def __contains__(self, other):
-        """
-        check if this space properly contains the `other` space, i.e. if the `other` space is contained
-        within the boundaries of `self`.
+        """Is `other` contained in the boundaries of this space?
```
hamogu, line 247: Question: Does "contain" mean that start/stop of other are within the start/stop boundaries of the self spaceor does it mean that the exact numbers are part of it. On other words, is [3,4,5] contained in [2.5, 3.5, 4.5, 5.5] or not?

sherpa/models/regrid.py
```diff
@@ -502,12 +409,23 @@ class ModelDomainRegridder1D():
...
+        the interpolated y values. The default is to use
+        sherpa.utils.akima.akima.
+
     Attributes
     ----------
-    method
-        The function that interpolates the data from the internal
-        grid onto the requested grid. The default is akima. This is
-        *only* used for point grids, as integrated grids use a simple
+    method : function reference
```
hamogu, line 425: You mean "function", not "function reference" right? A function is an object, so that's the object you would put in. I'm not aware of any "reference" or "pointer" type in python/

hamogu (changes requested):
This is clearly an improvement as is. I have a few suggestions for clarification.

sherpa/models/model.py
DougBurke: Actually, I was explicitly vague here because I don't know, and you can't easily tell from the code, the rules. I think it is all cases where you change the parameter value (so it's fits that will get reset but not explicit changes), but it relies on the reset method of the parameter which includes the mysterious `_guessed` attribute which I don't have time/energy to track down...

https://github.com/sherpa/sherpa/blob/d2668e402b4165a81ca2f49830d17a6560a26b27/sherpa/models/parameter.py#L501-L512

So I'm glad you called this out as I don't know what best to say here.

sherpa/models/regrid.py
DougBurke: I don't know. I didn't write this code and I do not understand it.

sherpa/models/regrid.py
DougBurke: There should be function references. I thought there was specific terminology for "hey, I'm being all first-order and sending around callable thingies" that was more than just 'function'. Maybe "callable"?

hamogu: "callable" is good. In fact, under the hood a function is just a callable object...

sherpa/models/model.py
hamogu: OK, let's just leave as you have written it for now.

sherpa/models/regrid.py
hamogu: It's just checking the start/stop boundaries, not the actual values in the middle. (Just look at the only line of code in this function). I should have expanded that before I wrote the comment... here is suggested text:
"Are all elements of `other` within the low end to high end range of this space?"

sherpa/models/model.py
DougBurke: I stuck some text in the thawedparhardmins/maxes attributes and also added some text to the top-level module documentation. Let me know if it's sufficient.

sherpa/models/regrid.py
DougBurke: I've changed this - it's a dunder method so I don't think most people will see it!

DougBurke:
@hamogu - as part of the callable discussion I realised the code never actually ensured the method value (aka interp kwarg) was checked for being a callable, so we now do (and added a test). This is really low-level code so I don't think we need to do too beyond what I've done. But it does require a look at by someone-not-me™.

sherpa/models/parameter.py
```diff
@@ -19,6 +19,171 @@
...
+    default_val = 8.0
+    default_min = -3.4028234663852886e+38
+    default_max = 3.4028234663852886e+38
+
+The `link` attribute stores the expression:
+
+    >>> p.link
+    <BinaryOpParameter '(2 * other.beta)'>
+
+If the linked expression ends up to the parameters limits being
```
hamogu, line 168:
```suggestion
If the linked expression ends violating the parameters limits 
```

sherpa/models/model.py
hamogu: It's great.

sherpa/models/parameter.py
DougBurke: Ouch. That was a terrible sentence. I've put it out of its misery and re-wrote the paragraph (mainly just the first line). Hopefully it's more readable.

sherpa/models/regrid.py
```diff
@@ -460,22 +371,18 @@ def overlaps(self, other):
...
+        representations of the meshgrid obtained from the x and y axis
+        arrays, as in `numpy.meshgrid(x, y)[0].ravel()`
 
         Returns
         -------
         tuple
-            A tuple representing the x and y axes. The tuple will contain four arrays if the dataset is
-            integrated, two otherwise.
-        """
+            A two (x, y) or 4 (x, y, xhi, yhi) tuple.
```
hamogu, line 383:
```suggestion
            A two element (x, y) or 4 element (x, y, xhi, yhi) tuple.
```
At first, I thought I'll get two (x,y) tuples or even 4 tuples. Then I wondered what those 4 tuples mean. I suggest a rewording to avoid any confusion.

sherpa/models/regrid.py
```diff
@@ -502,13 +409,16 @@ class ModelDomainRegridder1D():
     instead it creates an instance that is used to evaluate
```
hamogu, line 409: An instance of what? Of itself? Of an Evaluation space?

sherpa/models/regrid.py
```diff
@@ -538,8 +448,27 @@ def __init__(self, evaluation_space=None, name='regrid1d', **kwargs):
...
+        sherpa.utils.akima.akima.  The callable should accept `(xout,
+        xin, yin)` arguments and interpolate the `(xin, yin)` data
+        onto the `xout` grid, returning the interpolated data.
+        """
+        return self._method
+
+    @method.setter
+    def method(self, method):
+        if not callable(method):
+            raise TypeError(f"method argument '{repr(method)}' is not callable")
```
hamogu, line 466: Fine. But of course Python would fail later anyway when it tries to call that argument and it's not callable.

hamogu (changes requested):

sherpa/models/regrid.py
DougBurke: I prefer to catch an error like this earlier - close to the error - rather than later when the "this is not callable" error gets raised by some code unrelated to this setting.

hamogu: As I said, it's "fine". I just wanted to point out that it's not strictly necessary.

sherpa/models/parameter.py
```diff
@@ -485,6 +727,15 @@ def set(self, val=None, min=None, max=None, frozen=None,
 
 
 class CompositeParameter(Parameter):
+    """Represent a parameter with composite parts.
```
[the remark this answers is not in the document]
hamogu: That's great!

sherpa/models/regrid.py
DougBurke: I didn't change this text in this PR so I'm reluctant to add more changes if I can help it here.

If you are interested, the example provides some context, but I think if we were to fix this then we'd have to fix the 2D case to, and then end up explaining the whole file, which is just not going to happen any time soon.

hamogu: understood.

hamogu (approved):
All my points are addressed except for RTD not rendering bullet points. However, that's true for all locations which should have bullet points in the docs and is not specific to this PR at all. So, good to go. I'm also restarting tests now that #1187 is in.

Issue #1061, opened 2021-01-17
confusion over responses and unpack_arf/rmf


DougBurke:
More a note-to-self at the moment. I'm not sure if this can/should be cleaned up or just documented explicitly.

In playing around with PHA responses I've found that the DataPHA.get_responses routine returns different data to unpack_arf (DataARF vs ARF1D). Thanks to the design of ARF1D it isn't a "big" issue in many cases (as ARF1D should just fall-through to the underlying ARF for attribute access, this is unchecked but `__getattr__ has logic in it that suggests this is so). However, it did just catch me out:

```
In [4]: pha.get_response()
Out[4]:
(<DataARF data set instance 'sherpa-test-data/sherpatest/3c273.arf'>,
 <DataRMF data set instance 'sherpa-test-data/sherpatest/3c273.rmf'>)

In [5]: arf, rmf = pha.get_response()

In [6]: help(arf)


In [7]: help(rmf)


In [8]: arf.__class__.__name__
Out[8]: 'DataARF'

In [9]: rmf.__class__.__name__
Out[9]: 'DataRMF'

In [10]: pha.__class__.__name__
Out[10]: 'DataPHA'

In [11]: arf = unpack_arf('sherpa-test-data/sherpatest/3c273.arf')

In [12]: arf.__class__.__name__
Out[12]: 'ARF1D'
```

So `unpack_arf` returns `ARF1D` but the response is `DataARF`.

I was expecting `unpack_arf` to return `DataARF`.

This plays into an underlying question I have over the large number of sherpa.astro.instrument classes and why they are all needed.

hamogu:
I've come across a similar question in the session classes (#796) mad @DougBurke sounds as if the unexplained duplication of functionality might be more common. Maybe @olaurino can give some background or point to an early design experiment that explains the original intent?

hamogu:
When I say "similar" I mean "I don't understand why extra classes are needed or why code is duplicated" in cases that might have nothing to do with the ARF (which is the point of this issue).

DougBurke:
The `ARF1D` is doing something different to the `DataARF` class, as it acts to "add a response to a model expression" when called. So there is some reason for different functionality, but there are a number of similar classes in `sherpa.astro.instrument` and it's hard to follow the intended logic. In this particular question I'm (originally) focused on `unpack_arf`, I guess, and why it doesn't return a `DataARF`.

We could also move the response-generating functionality into `DataARF` and co, but I think that combines to many concerns.

[...]

PR #1181, opened 2021-06-07
Attempt to unify set_arf/rmf and load_arf/rmf

hamogu:
I am proposing to make `load_arf/rm`f work not just with filenames, but also with the ARF/RMF objects that I get from `get_arf/rmf`. If that is done, `set_arf/rmf` would essentially not be needed any more and could be removed. (Note that I'm not trying to remove them in this PR because they are a legacy interface I don't want to break, but they can stay around as a alias to `load_arf/rmf`).

My motivation for this change is:
- When I read "load_arf" I something think of "load into dataset" not necessarily "load from file".
- If these functions were more similar, then `set_multi_arfs/rmfs` could accept both filename and ARF/RMF objects. 
- If these functions were more similar, then `fake_pha` would work not only with filename, but also with ARF/RMF objects. That way, if I run a number of simulations, I only have to read the ARF/RMF file once and can keep it in memory. 

Unfortunately,
- I don't quite understand the differentce between ARF1D and DataARF and when to use the one or the other.
- The argument names of `load_arf/rmf` and `set_arf/rmf` are almost but not quite identical. In one case, the important argument is called `arg`, the other `arf` or `rmf`. That's not a real problem, it's just annoying for paramerizing tests.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
@@ -157,8 +157,9 @@ def make_rmf2():
     return rmf1, rmf2, dset
 
 
+@pytest.mark.parametrize("set_arf", [ui.set_arf, ui.load_arf])
 @pytest.mark.parametrize("id", [None, 1, "three"])
-def test_set_multi_arfs(id, clean_astro_ui):
+def test_set_multi_arfs(id, set_arf, clean_astro_ui):
 
     arf1, arf2, dset = make_arf2()
 
@@ -166,10 +167,10 @@ def test_set_multi_arfs(id, clean_astro_ui):
     d = ui.get_data(id=id)
     assert d.response_ids == []
 
-    ui.set_arf(id=id, arf=arf1, resp_id=1)
+    set_arf(id, arf1, 1)
     assert d.response_ids == [1]
 
-    ui.set_arf(id=id, arf=arf2, resp_id=2)
+    set_arf(id, arf2, 2)
     assert d.response_ids == [1, 2]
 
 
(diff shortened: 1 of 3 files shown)


hamogu:
(Note that `fake_pha` already has this capability, but with the following code:
```
        if rmf is None:
            raise DataErr('normffake', id)

        if type(rmf) in (str, numpy.string_):
            if os.path.isfile(rmf):
                rmf = self.unpack_rmf(rmf)
            else:
                raise IOErr("filenotfound", rmf)

        self.set_rmf(id, rmf)
```
With the change that I'm working on here, that would simply be:
`self.load_rmf(id, rmf)`. Checking is files exist should really not be in the purview of `fake_pha`.

The difference between DataARF and ARF1D is also dicussed in #1061 where the idea to combine or unify the two classes in some way is already discussed.

DougBurke:
Not commenting on the validity or not of the overall aim, but if we did this then we'd have to do it for a bunch of other commands too - a partial list (because there's a lot of load_xxx commands and they don't all count): load_bkg/set_bkg, load_image/set_image and load_data/set_data.

hamogu:
And `set_arf` and `set_rmf` can continue to be the advertised way in the docs, they would just be implemented as an alias or thin wrapper around `load_arf/rmf`

DougBurke:
Just to note that I don't have your inate dislike of the existing naming schemes and behavior, but that's because I've been using this code for too long and I've just internalized it.

anetasie:
@hamogu Is there any difference between set_arf/rmf and the proposed change to load_arf/rmf?

anetasie:
Just a comment which may be useful (or not) for the discussion here. The original idea with `set_arf/set_rmf` was to be able to have several RMF/ARFs available and then select the ones used for the specific dataset ids, or one specific ARF/RMF which could be used with multiple dataset ids. I'm not sure that this particular functionality is used too often. Typically the appropriate ARF/RMF filenames are present in the header and `load_pha()` will load all the files.

DougBurke:
Conceptually I can see the difference between `load_arf` and `set_arf` as written. Whether it is helpful is another matter.

I can safely say I don't understand the difference between ARF1D and DataARF (and similar for RMF). Does it add useful functionality or is it a case of re-implementing the same thing in different places?

I think it's safe to see that this is not going to make 4.15 though.

Issue #1133, opened 2021-04-28
Consider the structure of our documentation


hamogu:
Sherpa has (at least it feels like that) more functions and ways to do the same thing than numpy or e.g. matplotllib Even after contributing to the Sherpa codebase a bit, I still constantly discover new functions, new parameters, etc. It often does not feel "coherent" and the fact that our docs are organized to be read by function (docstrings, ahelp "see also"), to just skim the surface (readthedocs text) or in threads (ciao website) instead of systematically explain all options does not help.

Threads are good as examples, but if you want to do something just slightly different, it's hard to find out how because you don't know where to start looking and looking things up by function requires knowing what those functions are called.

I know it's a lot of work to write good documentation and the fact that we serve two different communities (through the stand-alone release and as part of CIAO) and then we have two totally different interfaces (UI layer and OO) does not make it easier.

Still, I'm missing pages like 
- https://docs.scipy.org/doc/scipy/reference/tutorial/stats.html (concept)
- https://docs.scipy.org/doc/scipy/reference/stats.html (list)
- https://matplotlib.org/stable/users/interactive_guide.html (concept)
- https://docs.astropy.org/en/latest/time/index.html (concept and list in one page)

that explain the concept and/or list all available options for one topic (e.g. all plotting commands with a shot explanation what they do beyond the name of the function I get in he "see also" list).
https://cxc.cfa.harvard.edu/sherpa/snapshot.html is pretty close to what I have in mind, but not complete in any sense (because it, again, it meant to just skim the surface).

I'm not advocating to write that *right now* because that's too much work, but I'd like to know or discuss (if not yet decided) what our vision for the future is.

Can we unify the CIAO website and readthedocs? Generate them from the same input files? Do we keep them totally separate (one for the ui layer, one for OO), but increase linking between them? Where do we add more content? Where would a more systematic overview go?

I know @DougBurke has done a lot of work to unify docstrings and ahelp files and has probably thought about this...

DougBurke:
To be honest what would help is people contributing (like this suggestion is) to where to go next and then **actually writing** the documentation. So far I see people say the documentation is lacking, and then stopping there. I can not write all the documentation.

I do **not** see the CIAO and readthedocs pages being unified - they serve different use cases and the different backends (crates vs astropy is the main issue but there are some other subtle differences in region filtering) have enough functional differences to make it hard to argue that a CIAO user (my main paying customer) should spend a lot of time reading about stuff they don't care about.

[...]

PR #1182, opened 2021-06-07
Simulations with multiple responses

hamogu:
# Summary
Enable simulations with multipe ARF/RMF present (e.g. LETG/HRC).

# Details
This PR greatly extends the abilities of `fake_pha`. It can now deal with multiple responses, multiple background, and background model (instead of just drawing from the noise PHA distribution). 
`fake_pha` used to be available only on the UI layer, not as OO function. That is a little unwieldy, because the best way to set u pa complex simulations with potentially multiple ARF/RMF pairs, multiple backgrounds, background models etc, is in the OO interface. In the OO, one can load everything that's required into a `DataPHA` set and the run the simulation. In contrast, the `fake_pha` on the UI later has a long list of parameters and then "if X not None, set X" code. That's cumbersome to extend. Therefore, only a subset of the capabilities is included into the UI layer at this point.

Moving `fake_pha` to the OO layer also required the `get_response` function to be moved. In turn, that suggests reorganizing the `background.py` (which does not primarily deal with backgrounds) and the new `fake.py` files. It also requires a re-ordering of our docs to make a good space to explain this.

## Note
I'm asking for review already although this is not 100% done, because I expect that there will be a fee changes based on the discussion and I don't want to word-smith every last word in inline comments until there is consensus that this is close to the final form. 

## Note to self
Additional capabilities and simplifications would be possible with #1181, but this PR should not wait for that to happen. If #1181 happens, the fake_pha code in the UI layer can be simplified considerably.


sherpa/astro/background.py
```diff
@@ -61,20 +63,58 @@ def add_response(session, id, data, model):
...
+    return get_response_for_pha(data, model, bkg_srcs, pileup_model, id)
 
-    resp = session._get_response(id, data)
-    if data.subtracted:
-        return resp(model)
 
-    bkg_srcs = session._background_sources.get(id, {})
-    if len(bkg_srcs) == 0:
-        return resp(model)
+def get_response_for_pha(data, model, bkg_srcs={}, pileup_model=None, id=None):
```
hamogu, line 76: I don't like the "id" parameter, but I did not find another way to pass it down from the UI layer to get the correct error messages.

sherpa/astro/data.py
```diff
@@ -1427,7 +1429,15 @@ def __init__(self, name, channel, counts, staterror=None, syserror=None,
         self.exposure = exposure
         self.backscal = backscal
         self.areascal = areascal
-        self.header = header
+        if header is None:
+            self.header = {'HDUCLASS': "OGIP", 'HDUCLAS1': "SPECTRUM",
```
hamogu, line 1433: Some tests used to initialize the header like this. Note sure why... but instead it makes sense to do that here.

sherpa/astro/data.py
```diff
@@ -1761,6 +1771,44 @@ def get_specresp(self, filter=False):
...
+           model has been associated with the data set.
+
+        See Also
+        --------
+        get_response, get_arf, get_rmf
+        """
+        # import is here because sherpa.astro.instrument depends on
+        # sherpa.astro.data. Importing here instead of on the top
+        # avoids a circular import.
+        from sherpa.astro import instrument
```
hamogu, line 1800: I'm not a big fan of how this introduces a circular dependency (only worked around by moving the import into the method), but conceptually I think this functionality definitely belongs into the data class. Another facet of the discussion in #1061. This would go away if we did not distinguish between data and instrument ARF and RMf - a distinction that's not clear to me anyway.

sherpa/astro/data.py
```diff
@@ -1761,6 +1771,44 @@ def get_specresp(self, filter=False):
 
         return newarf
 
+    def get_full_response(self, pileup_model=None):
+        """Calculate the response for the dataset.
```
hamogu, line 1775: Need this function in the OO layer. Used to be in UI only before.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -99,10 +100,11 @@ def test_fake_pha_incompatible_rmf(id, clean_astro_ui):
     ebins = np.asarray([1.1, 1.2, 1.4, 1.6, 1.8, 2.0])
     elo = ebins[:-1]
     ehi = ebins[1:]
+    arf = ui.create_arf(elo, ehi)
```
hamogu, line 103: The order of checks in `fake_pha` has changed. Now, we need an arf to even get to the point where the error message that is tested here will be thrown.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -176,7 +178,9 @@ def test_fake_pha_basic(id, has_bkg, clean_astro_ui):
     # is summing the counts (to average over the randomness)
     # and then a simple check
     #
-    assert faked.counts.sum() > 200
+    assert (faked.counts.sum() > 200) and (faked.counts.sum() < 3000)
```
hamogu, line 181: tightened a few tests. I did not see random failures,  I think it's OK. If we see those tests fail for no reason and it becomes a problem we can always loosen the requirements again.

sherpa/astro/ui/utils.py
```diff
@@ -8987,9 +8975,7 @@ def _add_convolution_models(self, id, data, model, is_source):
         id = self._fix_id(id)
 
         # Add any convolution components from the sherpa.ui layer
-        model = \
-            sherpa.ui.utils.Session._add_convolution_models(self, id, data,
-                                                            model, is_source)
+        model = super()._add_convolution_models(id, data, model, is_source)
```
hamogu, line 8978: unrelated. It just bugged me how that needed three lines and line breaks.

hamogu (commented):
I left some "review-style" inline comments to guide a review. (Although I may move some files again...)

sherpa/astro/background.py
```diff
@@ -61,20 +63,58 @@ def add_response(session, id, data, model):
...
+    Parameters
+    ----------
+    data : sherpa.astro.data.DataPHA instance
+        The dataset (may be a background dataset).
+    model : sherpa.models.model.ArithmeticModel instance
+        The model (without response or background components)
+        to match to data.
+    bkg_srcs : dict
+        Keys in the dictionary need to be the background ids in the dataset
+        ``data``, and the values are the corresponsing source models.
```
wmclaugh, line 94: corresponsing -> corresponding

sherpa/astro/background.py
```diff
@@ -61,20 +63,58 @@ def add_response(session, id, data, model):
...
+    pileup_model : None or `sherpa.astro.models.JDPileup` instance
+        Pileup model for the dataset if needed, or ``None`` for no pileup
+        model.
+    id : string
+        A string to label the dataset in warning messages. If this is set
+        ``None`` the name of the dataset it used. Thus paramters is mainly
```
wmclaugh, line 100: paramters -> parameters

sherpa/astro/background.py
```diff
@@ -61,20 +63,58 @@ def add_response(session, id, data, model):
...
+        needed if this function is called from the UI layer, where datasets and
+        models have ids that are not stored in an attibute of the dataset
```
wmclaugh, line 102: attibute -> attribute

sherpa/astro/fake.py
```diff
@@ -0,0 +1,176 @@
+#
+#  Copyright (C) 2007  Smithsonian Astrophysical Observatory
```
wmclaugh, line 2: 2007 -> 2021 ?

hamogu:
Yes, the CI tests fail, but with some really obscure installation error that has almost certainly nothing to do with the PR. I've restarted it but get the same message but it's not the same as #1187. Either way, I would ask to review the content first and ignore the CI for now. Tests pass locally and since this is not necessarily easy-to-review there will probably iterations and some time to shape out the CI.

sherpa/astro/background.py
DougBurke: Requiring id is a bit of a code smell (we have this in other places, in particular model expressions passed as a string) since we really shouldn't have the upper-level concepts moving down to this code. That's not to say this particular change is bad (I haven't looked beyond this response) but we should see what would be needed to remove this argument.

DougBurke: I will note though that `add_response` takes an `id` value so the question might be whether we expect this routine to be used outside of the session class?

sherpa/astro/ui/utils.py
```diff
@@ -8727,17 +8740,26 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+
         if id in self._data:
             d = self._get_pha_data(id)
         else:
-            # Make empty header OGIP compliant
-            # And add appropriate values to header from input values
-            d.header = dict(HDUCLASS="OGIP", HDUCLAS1="SPECTRUM",
-                            HDUCLAS2="TOTAL", HDUCLAS3="TYPE:I",
-                            HDUCLAS4="COUNT", HDUVERS="1.1.0")
+            d = sherpa.astro.data.DataPHA('', None, None)
```
DougBurke, line 8762: My current thinking is that we shouldn't allow a user to create an empty `Data` object - we should make them supply this (and ideally make the independent/depentent data immutable [I think this is actually a step too far] or at least don't change their size). However, that's not for this PR...

What would make the above easier to do (at least for the `DataPHA` case is if we can re-arrange the logic of this method do that we move the `DataPHA` call until after we've created `rmf0`, so we can do something like (I haven't checked this is sufficient or if it isn't actually this easy because of the information flow).

```python
if id in self._data:
    d = self._get_pha_data(id)
    if len(d.channel) != rmf0.detchans:
        raise Oops

    for resp_id in d.resonse_ids:
        d.delete_response(resp_id)

    ... maybe something more

else:
    chans = sao_arange(1, rmf0.detchans)
    counts = chans * 0
    d = DataPHA('faked', chans, counts) 
```

We might also want to stop digging around in the internals of the Session class and we should probably just use try/except (although I don't like having so much code unrelated to the exception handler running within it):

````
try:
    d = self._get_pha_data(id)
    ... check number of channels
    for d.response_ids:
        ...
except IdentifierErr:
    .... need to create the object
```

sherpa/astro/ui/utils.py
```diff
@@ -8635,18 +8637,25 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+           background `~sherpa.stro.data.DataPHA` datasets.
+
         Parameters
         ----------
         id : int or str
            The identifier for the data set to create. If it already
            exists then it is assumed to contain a PHA data set and the
            counts will be over-written.
-        arf : filename or ARF object
+        arf : filename or ARF object or list of filenames
```
DougBurke, line 8651: I have two related questions here as we look at the `fake_pha` command:

1. I dislike the way it over-writes my existing dataset and would prefer it if it creates a new one - I think the ship has long sailed on that and we should just make sure we can easily "copy" a dataset (there is `copy_data` which does this, I just don't know if it also copies responses or just points to the same response).
2. if I have a PHA file loaded with arf and rmf why should I give the arf and rmf, why not just use them

I think of fake_pha as either being used either for a proposal - in which case you are not necessarily going to have a loaded-PHA file to use - or to see how significant a feature or model is for an observation (when you probably have everything set up in the PHA file when it's loaded). Do these two use cases have "different" interfaces? Probably not.

sherpa/astro/ui/utils.py
```diff
@@ -8635,18 +8637,25 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
            dataset identifier is used, rather than creating a
            completely new dataset.
 
+        .. versionchanged:: 4.13.2
+           Multiple ARFs and RMFs can now be used for the simulation.
+           Furthermore, background models can be specified instead of
+           background `~sherpa.stro.data.DataPHA` datasets.
```
DougBurke, line 8643:
```suggestion
           background `~sherpa.astro.data.DataPHA` datasets.
```

I am somewhat loathe to add too many sphinx-isms to the documentation as I find them distracting and, for the occasional user, rather baffling (e.g. I know in the past I've looked at numpy and/or matplotlib docs and had to spend time carefully reading the docs). Of course, I've also added `.. versionchanged` lines which are the same thing pand have at least one header for a module which uses this very `~xxx` syntax], but they are a bit-more obvious and I can easily handle them when I create ahelp files, which is less the case for things like `~sherpa.astro.data.DataPHA`.  So I'm not saying no to this, just that we should think about this a bit.

sherpa/astro/ui/utils.py
```diff
@@ -8635,18 +8637,25 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
            dataset identifier is used, rather than creating a
            completely new dataset.
 
+        .. versionchanged:: 4.13.2
```
DougBurke, line 8640:
```suggestion
        .. versionchanged:: 4.14.0
```

I don't believe we are having a 4.13.2 release.

sherpa/astro/background.py
hamogu: If by "this" you mean the new "get_response_for_pha", then the answer is yes. I need to for the faking, that's why I moved it out of the session class. Note that id does not do anything except appearing in the string of the warning that's emitted when a pile-up model is present. So the code itself is "clean" - a good indication that this should indeed be a generic function and not part of the session class. I considered rewording the warning to something so generic that it does not need the id in the string, but, from the UI layer, it's actually very useful to see which dataset caused the warning. Another option would be to pass the warning back up, catch it, and replace it with the text we want in the session class. But that's more complex than it's worth.

sherpa/astro/ui/utils.py
hamogu: And we might not want the changelog in the docstring anyway...

sherpa/astro/ui/utils.py
hamogu: "Do we want links in sphinx" is a bigger question than this PR. We should sit down at some point and think about what we want to do. Personally, I find them very useful in the rendered sphinx docs and would advocate to use them in many more places ;-) but I don't know what implications that has for generating ahelp files, if this markup should be stripped etc. #1133 could be the place to discuss that. (Just as a short note: For the marx docs, I've written a [custom sphinx extention](https://github.com/Chandra-MARX/doc/blob/main/source/marxmarkup.py) to make use of the sphinx mark-up for marx parameters values and to generate links to ahelp pages. I imagine that we could do a similar thing here if we need it.)

sherpa/astro/ui/utils.py
hamogu: Well, "the ship has sailed". My change here is actually incremental. You could always run a `fake_pha` with either a filename or an arf object. I did not change that. What I did change, is the ability to pass in a list of filenames (and a list of ARF objects should also be possible. Need to check if I put that in the code and just forgot it for the docs or not. In the same vain, point (2) is a change to the existing UI that goes beyond just adding the ability to run multiple ARF/RMF. In the past, you could use an id that was taken and it would just be overwritten with the new ARF and RMF. The same happens here.

I like the suggestion to reuse ARFs and RMFs is present (say with "arf=None", which could be the default), but I'm not sure we want roll this into this PR or leave it for a follow-up since it's a separate issue really.

sherpa/astro/ui/utils.py
```diff
@@ -8755,6 +8777,37 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+
+        if d.channel is None:
+            d.channel = sao_arange(1, rmf0.detchans)
+
+        else:
+            if len(d.channel) != rmf0.detchans:
+                raise DataErr('incompatibleresp', rmf.name, str(id))
+
+        # at this point, we can be sure that arf is not a string, because
+        # if it was, it went through load_Arf alreayd above.
```
hamogu, line 8800:
```suggestion
        # if it was, it would have gone through load_arf above.
```

sherpa/astro/ui/utils.py
hamogu: I'm not sure I share the concern about "empty" Data objects. The object is not returned to the caller until it is complete. So, the user never sees the empty object. I guess we can also collect all the pieces (counts, channels, exposure time) in independent variable and then create the DataPHA once it's all ready. The user won't see the difference. I'll look if that's easy to do.

DougBurke: Well, the empty DataPHA object isn't a problem for the user of this code. It is a problem if I go and change DataPHA.__init__ to ensure that channel and counts are set - this is, I believe, the only place in our code where that would fail, so I'm eager to tweak the code here!

hamogu: We need to discuss and settle this before I can do any further work:
For the OO interface, I've settled on an interface where user sets up the DataPHA object with arf, rmf, exposure time etc - essentially all but the channels and counts. Channels and counts will then be filled in by `fake_pha`. That only works if the user can make an empty `DataPHA` object. The alternative is what the old UI method did - you pass in arf, rmfs, exposure time, etc. as separate parameters and then the methods puts them all into a the `DataPHA` object. I think that's cumbersome and complex - just look at the old `fake_pha` UI code because that code needs to duplicate  the load_arf/set_arf code to check if the arf is an object or a filename. If it's a filename it check if that file exists, ... I don't think "checking is a file exists" should be in the purview of `fake_pha`. It gets even more complex if we want to set up background, multiple backgrounds, or multiple arfs/rmfs (the goal of this proposal). Now, I also need to replicate the interface (and some of the code) for `set_multi_arf`, `set_multi_rmf`, and `load_background` . Why would I want to do that? We do have methods that do that already. Then, we could make and object that holds the ARF, RMF, background, exposure time, OGIP compliant header etc. so that we can pass it around, e.g. from the UI layer to the `fake_pha` in the OO layer - but we do have such an object already: `DataPHA`.

So, I think that `fake_pha` should be limited to actually do the faking and leave the setup (loading, setting arf, rmf, background, exposure time) to other existing routines, at least in the OO layer. In the UI layer, we keep the legacy interface, but, since that calls the OO layer to do the actual faking, we do need to set up the empty `DataPHA` object.

I also think that conceptually, that is the right way to think about it: We are creating a new dataset, so we should be able to create a new `DataPHA` object. The only thing that's unusual (=different from reading a fits file) is that we have the meta information first and the channels and counts later. However, that's how simulations work: We set up the conditions and then we run the simulation to get the counts.

In summary: I think any way that avoids making an "empty" `DataPHA` is going to be very cumbersome in some way or another. What we can do is to load the arf first and then create the `DataPHA` with some dummy entries to avoid the `None,None` in the signature. However, that's mostly cosmetic and does not address the deeper point that we'll only know the counts **after** we've set up everything else.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: Would it make sense to use an explicit seed and add the `reset_seed` fixture to reset the seed at the end of the test. Then we can make the tests much-more exact.

sherpa/astro/ui/utils.py
DougBurke: I haven't read this thoughtfully but I think we can have the best of both worlds - we can allow users / the code to change the counts array as long as it is of the correct size. So, we could for the DataPHA case require the channel array to be sent to `__init__` and it can set up an empty counts array if not set. This wouldn't need to be done before this PR is done.

What I really would like is for our code to be able to ensure that things have the correct sizes which means we then we can be more confident that downstream code isn't going to fall over because we've just tried to combine the counts and syserror columns and they don't match. 

However, saying all this is a bit academic as I need to look at #1197 as I suspect my idealism is about to be dashed by reality.

hamogu: Let's think about the OO interface for the moment, because that's cleaner (the UI has code around it to set up things with the right dataset numbers etc.), so in OO, I'd like to do:
```
fakes = DataPHA(None, None)
fakes.exposure = 5000
fakes.load_arf('XXX.arf')
fakes.load_rmf('XXX.rmf')
fake_pha(fakes)
```

If, instead, I need to make fake channel and count arrays then it goes something like this:
```
arf = load_arf('XXX.arf')
counts = np.zeros_like(arf.channels)
fakes = DataPHA(arf.channels, counts)
fakes.exposure = 5000
fakes.load_arf('XXX.arf')
fakes.load_rmf('XXX.rmf')
fake_pha(fakes)
```
(conceptually, I don't actually know on top of my head if `.counts` is the right thing). Why am I loading the arf twice? What's the real purpose of that intermediate "counts" array? 
I'm not saying it can't be done this way, I'm just saying that it's not "free" as in "it complicates the API", so I want to make absolutely sure that that's worth the trouble.

DougBurke: > The TL;DR of this comment is that my hopes are likely not going to work (that is, forcing validation checks when we make changes to the DataPHA object rather than when we evaluate the object to get data out) and so we should just ignore these thoughts and carry on with this PR.

To be picky you can't use the ARF but the RMF (arf's do not have a channels attribute and do not contain the correct number of rows in geeral), so it would be something like

```python
rmf = unpack_rmf('XXX.rmf')
counts = np.arange(rmf.offset, rmf.detchans + rmf.offset)
fakes = DataPHA(counts, np.zeros_like(counts))
fakes.exposure = 5000
fakes.set_rmf(rmf)
fakes.set_arf(unpack_arf('XXX.arf'))
fake_pha(fakes)
```

(here using existing features not some of the work you've been playing with).

An important part of the PHA concept is the range of channels - which can be larger than the actual channels argument, but this is not something we really support at the moment. What I really want is to be able to validate arguments - such as the quality array - have the correct length when set. This is easier if we force the channels to be set when `DataPHA` is called, but as you show it's a bit annoying. What we could do is allow the values to be None but as soon as something is set that defines the channels (which I was thinking would be setting the channels but in your example it's setting the RMF) then we can enforce it. I now have serious concerns about my whole idea since we have not really focused on ARF/RMF ordering but if we were to do this we'd need to enforce RMF first (unless it's a square RMF matrix so we could set ARF first).

hamogu: Thinking about this more: I totally understand what @DougBurke is coming from: Our lives as coders get a lot simpler if we can validate an object in the `__init__` and after that we can assume that everything fits. However, for the simulation, that's a wrong way around - and simulations are an important use case for us.
So, I think we are stuck with checking consistency every time the counts, channels, arf, rmfs are changed e.g. in the the set_arf, set_rmf, or in a property setter for counts and channels.

Bottom line: Let's, as @DougBurke suggests, "ignore these thoughts [at least for now] and carry on with the PR"

hamogu:
Added the ability to use ARF and RMF that were previously set on a dataset. I think this suggestion by @DougBurke makes a lot of sense, even to it requires a few more if cases and paths though `fake_pha`. I've addressed one of the other comments as well, but several of the discussion above are not actionable (or we decided to not tackle them in this PR). So, does this address al points sufficiently?

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: I could do that if you prefer, but I also like to have the test run with a different random seed every time just to make sure we did not happen to catch some special case with the seed I choose. If these tests fail more than 1:1000 times, we can do that. For now, I think we are safe.

sherpa/astro/ui/utils.py
hamogu: done.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: Fixed seed for one test that I've seen fail and leave free for the others (that I have not seen fail ever).

docs/evaluation/examples.rst
```diff
@@ -149,15 +149,18 @@ is used to show the grid used by
...
-Handling PHA datasets
-=====================
-
-:term:`PHA` data is more complicated than other data types in Sherpa because
-of the need to convert between the units used by the model (energy
-or wavelength) and the units of the data (channel). As a user you
-will generally be thinking in keV or Angstroms, but the
-:py:class:`~sherpa.astro.data.DataPHA` class has to convert to
-channel units internally.
+X-ray data (`~sherpa.astro.data.DataPHA`)
```
DougBurke, line 152: This doesn't create a link and so should just probably be `(sherpa.astro.data.DataPHA)` or removed completely - see https://sherpa--1182.org.readthedocs.build/en/1182/evaluation/examples.html#x-ray-data-sherpa-astro-data-datapha

docs/evaluation/examples.rst
```diff
@@ -300,6 +303,11 @@ For now we want to make sure we complete our analysis in
 energy units::
 
    >>> pha.set_analysis('energy')
+   >>> pha.notice(0.3, 7)
```
DougBurke, line 306: Why did you add these lines? You need to update `docs/_examples/evaluation/examples/pharesponses.py` to reflect this (I use these scripts to make it easy to regenerate the text and figures) and then update the output below where it's changed. Or just remove these lines.

docs/evaluation/examples.rst
```diff
@@ -517,3 +525,18 @@ earlier doesn't group the model to match the data)::
 
 Note that this example has not tried to subtract the background
 or fit it!
+
+Reference/API
```
DougBurke, line 529: I don't think you meant to add this here. We already have these linked to in the data section: https://sherpa--1182.org.readthedocs.build/en/1182/data/index.html#reference-api

I've been trying to only add the Reference/API sections to the "foo/index.rst" files and not "sub" files like "foo/bar.rst".

docs/evaluation/examples.rst
hamogu: I don't remember the PR number, but I had to rebase this because the docs where re-organized by @DougBurke in some other PR after I opened this. So, I'll go with "remove".

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
    >>> from numpy.random import standard_normal, normal
    >>> yconst = ymdl + standard_normal(ymdl.shape) * 10
    >>> ydata = ymdl + normal(scale=np.sqrt(ymdl))
+
+
+.. _data_pha_fake:
+
+X-ray data (`~sherpa.astro.data.DataPHA`)
```
DougBurke, line 47: The title will need changing as above

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+=========================================
+
+In principle, the same steps apply when simulating :term:`PHA` data
+(`~sherpa.astro.data.DataPHA` objects), however, the mechanics are a
```
DougBurke, line 51:
```suggestion
(:py:class:`~sherpa.astro.data.DataPHA` objects), however, the mechanics are a
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+little more complicated because we need to account for the
+instrumental response (:term:`ARF` and :term:`RMF`) and possibly also
+the background, which may contribute to the source spectrum that we
+want to simulate.  Readers not interested in X-ray data analysis may
+want to skip this section.
+
+Sherpa offers a dedicated function `sherpa.astro.io.fake_pha` for
```
DougBurke, line 58:
```suggestion
Sherpa offers a dedicated function :py:func:`sherpa.astro.io.fake_pha` for
```

I think it's `:py:func:` - you'd have to `python setup.py build_sphinx` and then check

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+simulations of :term:`PHA` data. A `~sherpa.astro.data.DataPHA` object
```
DougBurke, line 59:
```suggestion
simulations of :term:`PHA` data. A :py:class:`~sherpa.astro.data.DataPHA` object
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+needs to be set up with the responses for the source and an exposure
+time. So, we first create a `~sherpa.astro.data.DataPHA` object with
```
DougBurke, line 61:
```suggestion
time. So, we first create a :py:class:`~sherpa.astro.data.DataPHA` object with
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+The simplest case: Simulate the source spectrum only
+----------------------------------------------------
+
+With this model, it is now easy to run the simulation, which will
+calculate the expected number of counts in each spectral channel
+(where the number and width of the channels is given by the responses)
+and then draw from a Poisson distribution with this expected
+number. Thus, the simulated number of counts in each channel is always
+an integer and includes Poisson noise - running
+`~sherpa.astro.io.fake_pha` twice with identical settings will give
```
DougBurke, line 104:
```suggestion
:py:func:`~sherpa.astro.io.fake_pha` twice with identical settings will give
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+slightly different answers. With default settings, the input model is
+convolved with the RMF and multiplied by the ARF, and properly scaled
+for the exposure time.
+
+  >>> from sherpa.astro.fake import fake_pha
+  >>> fake_pha(data, srcmdl)
+
+We bin the counts into bins of at least 5 counts per bin and display
+an image of the simulated spectrum (see :ref:`data_pha` for details)::
```
DougBurke, line 113: This doesn't create a link - just the text `(see data_pha for details)`.

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+  >>> data.group_counts(5)
+  >>> from sherpa.plot import DataPlot
+  >>> dplot = DataPlot()
+  >>> dplot.prepare(data)
+  >>> dplot.plot(xlog=True, ylog=True)
+
+Sometimes, more than one response is needed, e.g. in Chandra LETG/HRC-S
+different orders of the grating overlap on the detector, so they all
+contribute to the observed source
+spectrum. `~sherpa.astro.fake.fake_pha` works if one or more responses
```
DougBurke, line 126:
```suggestion
spectrum. :py:func:`~sherpa.astro.fake.fake_pha` works if one or more responses
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+are set for the input `~sherpa.astro.data.DataPHA` object.
```
DougBurke, line 127:
```suggestion
are set for the input :py:class:`~sherpa.astro.data.DataPHA` object.
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+
+It is also possible that the input model already includes the
+appropriate responses and scalings. In this case, ``is_source=False``
+can be set to indicate to sherpa that the model is not a naked source
+model, but includes all required instrumental effects. In this
+way, `~sherpa.astro.fake.fake_pha` can be used with arbitrarily
```
DougBurke, line 133:
```suggestion
way, :py:func:`~sherpa.astro.fake.fake_pha` can be used with arbitrarily
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+
+Weak spectral features can be hidden in a large (instrumental or
+astrophysical) background, thus it can be important to include
+background in the simulation.
+
+Sample background from a PHA file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+One way to include background is to sample it from a
+`~sherpa.astro.data.DataPHA` object. To do so, a background need to be
```
DougBurke, line 154:
```suggestion
:py:class:`~sherpa.astro.data.DataPHA` object. To do so, a background need to be
```

docs/evaluation/simulate.rst
```diff
@@ -45,3 +40,181 @@ NumPy or SciPy to evaluate noise, such as ``numpy.random.standard_normal``.
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+If more than one background is set up, then the expected backgrounds
+will be averaged before the Poisson draw. For some backgrounds that
+expected value might be the value of the counts in a PHA file,
+while for others the expected value might be calculated from a model.
+
+
+
+Reference/API
```
DougBurke, line 208: As I mentioned above I've tried not to put reference sections outside of the `index.rst` file but

a) I may be lying and you can find examples elsewhere
b) not sure what to do here, but I think I'd prefer it to be at a higher level

sherpa/astro/background.py
```diff
@@ -61,20 +63,58 @@ def add_response(session, id, data, model):
...
-    if len(bkg_srcs) == 0:
-        return resp(model)
+def get_response_for_pha(data, model, bkg_srcs={}, pileup_model=None, id=None):
+    """Create the response model describing the source and model.
+
+    Include any background components and apply the response
+    model for the dataset.
 
+    This is essentially the object-oriented version of
+    `shera.astro.background.add_response`.
```
DougBurke, line 83:
```suggestion
    `sherpa.astro.background.add_response`.
```

sherpa/astro/fake.py
```diff
@@ -0,0 +1,179 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+'''Simulate PHA datasets
+
+This module contains the epynoumous function used to simulate
+`~sherpa.astro.data.DataPHA` datasets.
```
DougBurke, line 22: epynoumous looks like i've tried to spell it...

I'd suggest just

```
The fake_pha routine is used to create simulated DataPHA objects.
```

sherpa/astro/fake.py
```diff
@@ -0,0 +1,179 @@
...
+
+    This function replaces the counts in a PHA dataset with simulated counts
+    drawn from a model with Poisson noise. For the simulations, all the details
+    already set up in the PHA dataset will be used, including the exposure
+    time, one or more ARFs and RMFs, area and background scalings,
+    grouping, and data quality arrys.
+
+    Including a background component is optional; if requested, the background
+    will be a Poisson draw from the average of all backgrounds that have been
+    set for the input `~sherpa.astro.data.DataPHA`. For each background
```
DougBurke, line 45: This would need to be either

- :py:class:`~sherpa.astrp.data.DataPHA`
- `sherpa.astro.data.DataPHA`

sherpa/astro/fake.py
```diff
@@ -0,0 +1,179 @@
...
+        already contains the ARF and the RMF, then the background models should
+        do so, too. This setting has no effect if ``add_bkgs=False``.
+
+        For all background ids not listed in this dictionary, the
+        counts will be drawn from the PHA data of the background data set.
+    id : str
+        String with id number if called fomr UI layer. This is only used for
+        certain error messages.
+
+        Examples
```
DougBurke, line 88: The indention here is wrong - needs to be shifted left by 4 characters I think

hamogu:
Thanks @DougBurke fora very careful review of the documentation. To be honest, I had not yet checked that it all renders correctly because I was waiting that we agree on the code first (no need to polish the docs until the names of all keywords etc. are settled). 

I will revise the docs based on @DougBurke suggestions and check that they render correctly.

sherpa/astro/tests/test_fake_pha.py
```diff
@@ -0,0 +1,227 @@
+#
```
DougBurke, line 1: Ideally we could rely on the tests here and have a simple test at the UI layer to check it has passed the info through (since the UI layer no-longer does very much) but I don't have the energy to determine what UI tests could be removed now...

hamogu: Most of the UI tests are very short in run time (they fake 4 channels or so), so I decided to leave them in. At the very least, that guarantees backwards-compatibility in the sense that I did not assign an entirely new meaning to the setups already tested in the UI layer.

docs/evaluation/examples.rst
hamogu: Again, I think that's a sign of a bad rebase. Fixing!

docs/evaluation/simulate.rst
hamogu: Looking at the higher level I see this: 
```

Reference/API
=============

This section describes some of the classes used to evaluate the
models, in particular in supporting models evaluated against
:py:class:`~sherpa.astro.data.DataPHA` objects.

.. toctree::
   :maxdepth: 2

   astro_background

```

In other words, the next higher level does NOT contain the API docs themselves, instead it refers to a stub page with the APIT docs for `astro_background`. I don't think that `fake` warrants it's own stub_page; instead, I think it should refer to this section here.

I would argue that instead of moving this into its own stub page (to make it look like the docs for `astro_background`, the sub page for `astrop_models` should be expanded to make it look more like this page (with narrative text and the API just at the bottom of that.

DougBurke: There's definitely room for us to have a more-principled set up in the documentation, as you've mentioned in #1133, but we (aka me as the reviewer) shouldn't hold back documentation improvements here because of that.

docs/evaluation/simulate.rst
```diff
@@ -163,7 +164,7 @@ The ``fake_pha`` function simulates the source spectrum as above, but
 then it samples from the background PHA. For each bin, it treats the
 background count number as the expected value and performs a Poisson
 draw. The background drawn from the Poisson distribution is than added
-to the simulated source spectrum (the sum of two Poisson distribution
+to the simulated source spectrum (the sum of two Poisson distributiona
```
DougBurke, line 167:
```suggestion
to the simulated source spectrum (the sum of two Poisson distributions
```

DougBurke:
@hamogu - before I approve this I just wanted to check you're finished with this now (ie the comment `I'm asking for review already although this is not 100% done` is out of date)?

DougBurke:
I don't understand - aren't https://github.com/sherpa/sherpa/pull/1182/commits/b8f21f54fa85ffd225b793d3ddb42bc88f01004a and https://github.com/sherpa/sherpa/pull/1182/commits/e09ce49098677818edd4ddd2a8667c49784b2cd7 the same change?

DougBurke (approved):

PR #1183, opened 2021-06-08
Check parameter link behavior

DougBurke:
# Summary

Minor tweaks to the parameter tests.

# Details

I started this because I was seeing strange behavior when writing #1180 and wanted tests. This lead me to first rewrite the tests to 

a) remove SherpaTestCase
b) switch from creating throw-away classes to just using functions as we don't gain anything from the classes

In doing so I noted an "accidental pass" in one of the tests (setattr was being used wrongly but in such a way that it created the error we were looking for, but we should have been looking for a slightly-different error in this case), and then some of the link tests have been "enhanced" to use different min/max values as it provides a check of the behavior (more than the previous version, which was essentially always checking that the limits were hugeval or -hugeval as we never changed anything).

I eventually added a test to check the behavior I was trying to investigate, but it turns out that my original analysis was wrong (I was checking the expression 2 * b then changing q and wondering why the link was't changing. doh!). So, overall I think this is a minor improvement to the parameter tests and gives me more confidence in the #1180 documentation (or, rather, that we don't need to re-write that documentation).

### DIFF
--- a/sherpa/models/tests/test_parameter.py
+++ b/sherpa/models/tests/test_parameter.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2007, 2016, 2018, 2019, 2020, 2021
-#                Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -25,7 +25,6 @@ from numpy import arange
 import pytest
 
 from sherpa.utils import SherpaFloat
-from sherpa.utils.testing import SherpaTestCase
 from sherpa.models.parameter import Parameter, UnaryOpParameter, \
     BinaryOpParameter, ConstantParameter, hugeval
 from sherpa.utils.err import ParameterErr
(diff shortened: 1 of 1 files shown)


sherpa/models/tests/test_parameter.py
```diff
@@ -25,200 +25,281 @@
...
+        #
+        assert minval == -10
+        assert maxval == 10
+        tst_pos(gauss.link, pos, -20, 20)
+    assert gauss.default_val == pos
+
+    # Note: these are not minval/maxval but the defaults.
+    # How, where, and why are they set?
+    # assert gauss.default_min == minval
+    # assert gauss.default_max == maxval
```
hamogu, line 252: This makes no sense. Maybe it did at some point, but it does not now. I suggest to just remove those 4 lines of comments.

hamogu (changes requested):
This is good to go except for the comments. There is no need for the frequent comparison to the "the original test". Once this is merged, this are the tests. Period. If somebody wants to dig out the history, they can check the git history (or even clearcase). What if I change those tests again? Am I going to put in text about how this differs form the old original vs the new original?

sherpa/models/tests/test_parameter.py
DougBurke: I've taken out all the comments I could find.

DougBurke:
Rebased (just because I know there have been CI-related changes recently and I didn't want to be caught by it here) and removed the old-test comments that @hamogu wanted removing. Should be ready for a final review once CI has gone green.

hamogu (approved):

PR #1184, opened 2021-06-09
Fix plotting of model components for PHA data (fix #1020)

### LINKED ISSUE #1020: plot_model_component does not work well for PHA data
You need to include the response in the model component, which is not easy to get (better recently now #888 has been merged but still not ideal). That is, you for something like

```
set_source(xstbabs.amdl * (xsvapec.m1 + xsvapec.m2))
```

you would say something like (safest)

```
rsp  = _session._get_response(1, get_data())
```

or (less general - e.g. doesn't handle pileup, multiple responses or fancy background cases)

```
rsp = sherpa.astro.instrument.Response1D(get_data())
```

and then

```
plot_model_component(rsp(m1), alpha=0.5)
plot_model_component(rps(m2), alpha=0.5, overplot=True)
```

or perhaps

```
plot_model()
plot_model_component(rsp(amdl * m1), alpha=0.5 ,overplot=True)
plot_model_component(rsp(amdl * m2), alpha=0.5, overplot=True)
```

sherpa

If you don't include the response you get garbage results.

We should probably add the response in automatically. The questions then are:

- if we want to use a different response then how would that work (is just checking for a subclass of `sherpa.astro.instrument.RSPModel` sufficient? I doubt it as what about RMF-only responses?)
- if I've used set_full_model then we don't have a way to "get" a response, which would require the user to supply their own (i.e. a use case for the first point)
- what happens if I really don't want to use a response (perhaps add a flag for this case)

DougBurke:
# Summary

The use of plot_model_component and get_model_component will now automatically add the response for PHA data sets, if there is one. Model expressions which contain a response will not be changed. Fixes #1020.

# Details

This is a re-work of the code I wrote in #1022 

1. just deals with the `plot_model` case, not the new functionality of the `plot_model_components` part of #1022, which was the most problematic
2. drops the `add_response` keyword and now makes a best-case estimate of whether a model contains a response, using the sherpa.astro.instrument.has_pha_response function (which is similar to a routine @hamogu suggested in #1022 but taking advantage of the `__iter__` support we have)

What we lose is the ability of a user to plot a model component without a response (e.g. say you have fitted a phenomenological model for the particle background which ignores all responses). Users who do this would have to be told how to create the data to plot themselves.

One "trick" in this is that the documentation for `sherpa.ui.utils.plot_model_component` has been updated when technically it should be `sherpa.astro.ui.utils.plot_model_component`. Given that the astro version is our primary user base I think this is okay (the text is not wrong for the sherpa.ui layer, it is just you don't have DataPHA data here). I actually hope to add an astro-secific plot_model_components routine (to add support for the `bkg_id` parameter) in which case we can move the text to the "correct" place. I note that our existng sherpa.ui documentation occasionally documents DataPHA-related documentation (e.g. `plot_model`), for similar reasons (not wanting to copy a function just to tweak the docstring).

# Note

In CIAO 4.13 the output of `plot_model_component(src)` would be numerically wrong but have the nice title:

    'Model component: xsapec.src'

Now we get the correct data, but the title is a tad-more ungainly (see `Example 3` below):

    'Model component: apply_rmf(apply_arf((38564.608926889 * xsapec.src)))'

There are tricks we could add to try and fix this, but I don't think it's worth it and it would probably only be a heuristic, so could fail at some point.

# Examples

The three examples really just plot the same data, but it may be useful to see the differences explicitly.

## Example 1

With the following script I create plots from CIAO 4.13 (labelled `before`) and this PR (`after`):

```
from sherpa.astro.ui import *
load_pha('3c273.pi')
notice(0.5, 7)
set_source(xsphabs.gal * xsapec.src)
gal.nh = 0.05
gal.nh.freeze()
src.kt = 3.96
src.norm = 5.233e-4

plot_model(alpha=0.5, ylog=True)
plot_model_component(src, alpha=0.5, overplot=True)
plt.savefig('fig.png')
print("Created: fig.png")
```

### before

before

### after

after

The blue line shows the `plot_model` output - i.e. `xsphabs * xsapec` - and the yellow line just the `xsapec` component. You can see #1020 in the `before` plot - the yellow line is way below the blue line and is completely wrong. In the `after` case we can see that `plot_model_component(src)` creates a plot very-similar to `plot_model()`, as it should, since the absorption is small in this case.

## Example 2

For the plot

```
plot_fit(ylog=True)
plot_model_component(src, alpha=0.5, overplot=True)
```

### before

before2

### after

after2

## Example 3

For the plot

```
plot_model_component(src)
```

### before

before3

### after

after3

### DIFF
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -1441,3 +1441,57 @@ def calc_grp_chan_matrix(fname):
     except sherpa.utils.err.IOErr as ioerr:
         print(ioerr)
         raise ioerr
+
+
+def has_pha_response(model):
+    """Does the model contain a PHA response?
+
+    Parameters
+    ----------
+    model : Model instance
+        The model expression to check.
+
+    Returns
+    -------
+    flag : bool
+        True if there is a PHA response included anywhere in the
+        expression.
+
+    Examples
+    --------
+
+    >>> rsp = Response1D(pha)
+    >>> m1 = Gauss1D()
+    >>> m2 = PowLaw1D()
+    >>> has_pha_response(m1)
+    False
+    >>> has_pha_response(rsp(m1))
+    True
+    >>> has_pha_response(m1 + m2)
+    False
+    >>> has_pha_response(rsp(m1 + m2))
+    True
+    >>> has_pha_response(m1 + rsp(m2))
+    True
+
+    """
+
+    # The following check should probably include ResponseNestedModel
+    # but it's not obvious if this is currently used.
+    #
+    def wanted(c):
+        return isinstance(c, (RSPModel, ARFModel, RMFModel))
+
+    if wanted(model):
+        return True
+
+    # This check relies on a composite class like the RSPModel is
+    # included in the __iter__ method (see CompositeModel._get_parts)
+    # otherwise the following would have had to be a recursive call
+    # to has_pha_instance.
+    #
+    for cpt in model:
+        if wanted(cpt):
+            return True
+
+    return False
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -10434,9 +10434,8 @@ class Session(sherpa.ui.utils.Session):
     def get_model_component_plot(self, id, model=None, recalc=True):
         """Return the data used to create the model-component plot.
 
-        .. versionchanged :: 4.13.2
-           For PHA data, the response model is now automatically added
-           by the routine unless the model contains a response.
+        For PHA data, the response model is automatically added by the
+        routine unless the model contains a response.
 
         Parameters
         ----------
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -27,7 +27,7 @@ import numpy
 
 import sherpa.ui.utils
 from sherpa.astro.instrument import create_arf, create_delta_rmf, \
-    create_non_delta_rmf
+    create_non_delta_rmf, has_pha_response
 from sherpa.ui.utils import _argument_type_error, _check_type
 from sherpa.utils import SherpaInt, SherpaFloat, sao_arange, \
     send_to_pager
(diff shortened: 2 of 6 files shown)


DougBurke:
I've asked for both @anetasie and @hamogu to check that this works sensibly. I'd imagine the technical review would be done by Moritz!

This is a simplified version of #1022 which is just focussed on issue #1020 so we don't need to worry about the "new in that PR" plot_model_components command works. What we do have is a version of #1022 without the "add_response" flag. This is **much more** user friendly but means that users who really  do not want to add the response in (e.g. fitting a component to explain the particle background without a response term) would have to create the data to plot themselves. I think this is an okay compromise.

sherpa/astro/instrument.py
```diff
@@ -1432,3 +1432,57 @@ def calc_grp_chan_matrix(fname):
     except sherpa.utils.err.IOErr as ioerr:
         print(ioerr)
         raise ioerr
+
+
+def has_pha_response(model):
```
hamogu, line 1437: That's an interesting idea. I might use that heuristic in #1182, as well to get rid of the "is_source" parameter. Not sure, but I'm definitely thinking about it.

sherpa/astro/ui/utils.py
```diff
@@ -10431,13 +10431,78 @@ def get_fit_plot(self, id=None, recalc=True):
 
     get_fit_plot.__doc__ = sherpa.ui.utils.Session.get_fit_plot.__doc__
 
-    # copy doc string from sherpa.utils
     def get_model_component_plot(self, id, model=None, recalc=True):
+        """Return the data used to create the model-component plot.
+
+        .. versionchanged :: 4.13.2
+           For PHA data, the response model is now automatically added
+           by the routine unless the model contains a response.
```
hamogu, line 10439:
```suggestion
           by the routine unless the model contains a response already.
```

sherpa/astro/ui/utils.py
```diff
@@ -10431,13 +10431,78 @@ def get_fit_plot(self, id=None, recalc=True):
...
+
+        Return the full source model (``fplot``) and then for the
+        components ``gal * pl`` and ``gal * gline``, for the data set
+        'jet':
+
+        >>> fmodel = xsphabs.gal * (powlaw1d.pl + gauss1d.gline)
+        >>> set_source('jet', fmodel)
+        >>> fit('jet')
+        >>> fplot = get_model('jet')
+        >>> plot1 = get_model_component_plot('jet', pl*gal)
```
hamogu, line 10491: This sounds as if `get_model` and `get_model_component_plot` are essentially identical. Yes, one of them is names `plot` and the other one is not. As a reader, I wonder what the difference is? While one get me a model object and the other one a figure object?

hamogu (approved):
I just have one question on the a docstring, but that can go either way. So, I'm approving already.

sherpa/ui/utils.py
```diff
@@ -12633,6 +12626,10 @@ def plot_model_component(self, id, model=None, replot=False,
         expression for a data set, including any instrument response.
         Use `plot_source_component` to display without any response.
 
+        .. versionchanged :: 4.13.2
```
wmclaugh, line 12629: no planned 4.13.2 release- better to remove 'versionchanged'

sherpa/astro/ui/utils.py
DougBurke: This is likely an error in the example code - it probably (will have to check) should have said

    >>> fplot = get_model_plot('jet')

DougBurke:
Rebased to address conflict and also addressed review comments.

sherpa/ui/utils.py
DougBurke: Done

sherpa/astro/ui/utils.py
```diff
@@ -10431,13 +10431,78 @@ def get_fit_plot(self, id=None, recalc=True):
 
     get_fit_plot.__doc__ = sherpa.ui.utils.Session.get_fit_plot.__doc__
 
-    # copy doc string from sherpa.utils
     def get_model_component_plot(self, id, model=None, recalc=True):
+        """Return the data used to create the model-component plot.
+
+        .. versionchanged :: 4.13.2
```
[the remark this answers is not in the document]
DougBurke: Done

sherpa/astro/ui/utils.py
DougBurke: I've made this change (replacing get_model with get_model_plot)

sherpa/astro/ui/utils.py
DougBurke: I haven't made this change as (for me) it doesn't add anything.

DougBurke:
@anetasie - can you try this out to see if it works for you?

anetasie (approved):

Issue #499, opened 2018-09-26
Add a "js-based plotting library" plot backend for use in Jupyter notebooks


DougBurke:
Should we look at Vega/Altair/Plotly/... based plot backends to provide interactive graphic capabilities from within a jupyter notebook. With the matplotlib backend you just get PNGs (at least by default).We could create more interactive visualizations with a JS-based plotting library.

I realize that Vega, Altair, and Plotly are not all the same thing (one is a specification rather than a library), and that there are likely other candidates.

olaurino:
Could be a good gsoc project.

DougBurke:
Maybe we need a "gsoc" label ;-)

DougBurke:
Also, and this may deserve its own issue: it would be nice to use https://github.com/ericmandel/js9 to display images from within a notebook. I have talked to Eric about this in the past but have got lost in the attempts to integrate JS9 into Python notebooks (see https://github.com/ericmandel/js9/issues/30)

anetasie:
When is the gsoc deadline? Do we need to post our project?

DougBurke:
I know other projects are worrying about GSOC, but I haven't been following the Astro group this year.

wmclaugh:
The application perioid runs from January 15 - February 6.

https://developers.google.com/open-source/gsoc/timeline


Google Summer of Code 2019 Timeline
DateEvent
2018
November 13 Program announced
2019
January 15 20:00 UTC Mentoring organizations can begin submitting
applications to Google
February 6 20:00 UTC Mentoring organization application deadline
February 6 - February 25 Google program administrators review organization
applications
February 26 12:00 UTC List of accepted mentoring organizations published
February 26 - March 25 Potential student participants discuss application
ideas with mentoring organizations
March 25 18:00 UTC Student application period begins
April 9 18:00 UTC Student application deadline
May 6 18:00 UTC Accepted student proposals announced
Community Bonding Period
<http://googlesummerofcode.blogspot.com/2007/04/so-what-is-this-community-bonding-all.html>
Students
get to know mentors, read documentation, get up to speed to begin working
on their projects
May 27 Coding officially begins!
June 24 18:00 UTC Mentors and students can begin submitting Phase 1
evaluations
June 28 18:00 UTC Phase 1 Evaluation deadline
Work Period Students work on their project with guidance from Mentors
July 22 18:00 UTC Mentors and students can begin submitting Phase 2
evaluations
[...]

Issue #1188, opened 2021-06-13
Several datastack plotting functions don't work


hamogu:
I've discovered that several of the utility plotting functions in the datastack module don't work (or at least, don't work as advertised).

```
cd sherpa/astro/datastack/tests/data
from sherpa.astro import ui
from sherpa.astro import datastack
# Load with the UI
datastack.load_pha('@3c273.lis', use_errors=True)
# Now get the datastack in OO interface (which is harder to use for loading see #727 
ds = datastack.DATASTACK
ds.plot_data()
ds.plot_xlim(1, 6)
```
The plot itself works, but the last line gives me an error, although the docs claim that this is exactly how it's supposed to be used (note however that the method is called "xlim" in the example call signatures, but `plot_xlim` in the signature: 
```
Signature: ds.plot_xlim(*args, **kwargs)
Docstring:
Get or set the x limits of the current axes.

Call signatures::

    left, right = xlim()  # return the current xlim
    xlim((left, right))   # set the xlim to left, right
    xlim(left, right)     # set the xlim to left, right
```
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-0f3362b96de1> in <module>
      1 ds.plot_data()
----> 2 ds.plot_xlim(1, 6)

~/anaconda3/envs/ciao/lib/python3.7/site-packages/matplotlib/pyplot.py in xlim(*args, **kwargs)
   1563     if not args and not kwargs:
   1564         return ax.get_xlim()
-> 1565     ret = ax.set_xlim(*args, **kwargs)
   1566     return ret
   1567 

~/anaconda3/envs/ciao/lib/python3.7/site-packages/matplotlib/axes/_base.py in set_xlim(self, left, right, emit, auto, xmin, xmax)
   3326                 f"Attempting to set identical left == right == {left} results "
   3327                 f"in singular transformations; automatically expanding.")
-> 3328         reverse = left > right
   3329         left, right = self.xaxis.get_major_locator().nonsingular(left, right)
   3330         left, right = self.xaxis.limit_range_for_scale(left, right)

TypeError: '>' not supported between instances of 'DataStack' and 'int'
```

I think the problem is with the wrapper that loops over those functions for plots made.

[...]

PR #1191, opened 2021-06-17
Make it easier to switch the plotting backend

### LINKED ISSUE #348: allow the user to specify multiple back ends in their ~/.sherpa.rc file
It would be useful if we could change the backend (plotting or I/O) while running Sherpa (or at least interactively). This is a "power user" feature, and would probably require a lot of work.

A simpler approach (hopefully) would be to allow users to specigy multiple backends in their ~/.sherpa.rc file, with the order listing preference, and then loop through them until one works or none do [I'm allowing for the possibility of there being > 2 backends, even if this is not the case at present].

With this, I could set

```
[options]
plot_pkg   : pylab chips
```

which would prefer the matplotlib backend but allow the use of ChIPS if matplotlib isn't available (or can't be loaded). I assume that a space-separated list is okay here (i.e. we do not need to worry about backend names with a space in them; as we control these names that should be a restriction we can add if necessary).

hamogu:
# Summary
Make is easier to switch the plotting backend (note that we currently only have on backend implemented anyway)

# Details
There are three parts to this: 
- Code changes to always use "plot.backend", not "backend" alone except in the module where "backend" is set the first time. That way, if a new module is assigned to "plot.backend" it will be used everywhere. Otherwise, in certain modules, "backend" might still be set at what is was when that specific module was imported for the first time. 
- Similarly, remove `begin = backend.begin` and similar conveniences in `plot.__init__.py` because those methods get bound to the first plotting backend that is loaded and will not be updated when the plotting backend is switched. Instead, always call the functions of the plotting backend directly.
- There is added documentation on how to switch a backend. That  documentation is hidden away in the module docstring on purpose because we currently only implement one backend; if we have more than one again in the future, we can call it out more prominently. 
- The `datastack` module used to implement its own plotting backends. Thus, switching the main plotting backend would not work (or cause wired interactions) for plots from the datastack module. This PR moves the additional functions needed for the datastack module into the main plotting setup and does away with the datastack specific backend. 

## Other notes

- This is complementary to #1185, which deals with the I/O backend. Together with #1185, this PR would close #348, but since I don't know if this or #1185 will be merged first, I did not set an autoclose for github. We can close #348 manually once both are in.
- This is somewhat difficult to thoroughly test with only one real backend (the do-nothing dummy backend helps a little, but another real backend would be better). Yet, I'm working on a PR to address #499 and once we have a second backend again, we will need this. Should it turn out later that some corner cases remain, we can address them later. At the very least, this should do no harm for now and reduce code complexity (by removing the extra backend in datastack).


DougBurke:
I haven't looked at the code yet but I note that - unlike the I/O backend - the choice of plot backend has some impact - in the preference settings of the various plot objects. This may be addressed by your first point of the details section but I wanted to be explicit here.

I also saw some "interesting" behavior when I added some multi-backend tests, where some code [probably in a test and not actual code would fail when the backend had been swapped because `FooPlot` wasn't `FooPlot` - unfortunately I forget the details.

sherpa/plot/pylab_backend.py
```diff
@@ -865,3 +862,43 @@ def plotfunc():
...
+       The dataset.
+    ids : array_like
+       The identifier array from the DataStack object.
+
+    See Also
+    --------
+    select_plot
+
+    """
+    plt.figure(ids.index(dataset['id']) + 1)
```
DougBurke, line 884: I'm a bit concerned with this (in that I can finally see it's been moved from the datastack code to here). The parameter description says `dataset` is a `str` or `int` but here you're accessing it by `id`, some something is wrong.

As we don't have a test of this we can't tell.

DougBurke:
I am likely to suggest leaving in the config code, as I did for #1185

sherpa/astro/plot.py
```diff
@@ -141,17 +138,17 @@ def prepare(self, data, stat=None):
         _check_hist_bins(self)
 
 
-class ModelPHAHistogram(HistogramPlot):
+class ModelPHAHistogram(shplot.HistogramPlot):
```
DougBurke, line 141: I think that separating out the sherpa.plot and sherpa.astro.plot namespaces, as you've done here, is a good idea as the naming of the different classes is "not great"™ so explicitly having to identify where the name comes from should help in the long run.

sherpa/plot/pylab_backend.py
hamogu: I think this is related to #1188  - the datastack plots that are supposed to summarize the stack in some way (not clear how - no docs, no example plot and does not work in main) are all broken right now. if at all possible, I don't want to roll that into the current PR though. They were broken (and obviously not tested) before, and they are broken now. This code is just copied over. I do think that datastack plotting needs a closer look and I do think that datastack is underused. However, I also strongly feel that this is not the PR to do that; here, I just want to make it easier to switch the backends and to do that I need to merge the datastack plotting code into this module to reduce the number of places where the plotting backend is defined independently.

 #1188 is open to remind us of the problem for now. In fact, I'm doing two science projects with datastacks now, so I hope that I get some ideas there who type of plots we actually want.

hamogu:
I removed any changes to sherpa.rc parameters and thus DS/CM review is no longer necessary. 
@DougBurke I'm not sure if you've done a full review or just dropped a few comments where things were obvious. LEt me know what you think.

hamogu:
I don't understand the CI failure that suddenly appeared last night. It's clearly not related to this PR, so I'm asking for review anyway under the assumption that it's just some intermittent network failure that causes trouble when pulling in that file and I hope that it goes away by itself.

sherpa/plot/__init__.py
```diff
@@ -26,8 +26,30 @@
...
+Which backend is used?
+----------------------
+
+When this module is first imported, Sherpa tries to import the
+backends installed with Sherpa in the order listed in
+`sherpa.plot.try_backends`. The first module that imports
+successfully is set as the active backend. The following command prints the
+name and the location on disk of that module::
+
+   >>> import sherpa import plot
```
Marie-Terrell, line 38:
```suggestion
   >>> from sherpa import plot
```

Marie-Terrell:
Other than the one comment, this looks good from the CIAO side for me.

hamogu:
I still need to deal with the plot_prefs, so I marked this as WIP for now. I have several ideas how that could be done, but for reference here is the problem (which @DougBurke of course pointed out above in https://github.com/sherpa/sherpa/pull/1191#issuecomment-883659724). In the current implementation, the plot_prefs are set when the plotting classes are first imported e.g.
```
class TracePlot(DataPlot):

    plot_prefs = backend.get_model_plot_defaults()
```
So, all instances of that class generated at later time will have the same plot_prefs. Unfortunately, those plot_prefs contain backend-dependent stuff because they are made for one specific backend. 

## Possible solutions

1) Get the prefs from the active backend when an instance is created instead of at class import time.
2) Remove the backend depended pref's altogether. I think this is possible through a combination of changes. It will cause some of the plots to look slightly different and it's a change of API, but I think it's worth it. The plot_prefs where just introduced to users in the "A sample of plots" notebook and users can only use them in the OO interface. I think they are probably few users that rely on setting the plot_prefs. On the other hand, my proposal below will simplify code considerably while keeping almost all of the current functionality.

## Detailed proposal
There are only a limited number of attributes that are currently set in the plot_prefs and they fall in different groups:
1) Sherpa specific stuff, like `ratioline=True`. Those are backend independent in the sense that they need to be implemented for each backend and can just be set as default in the corresponding call to the `backend.plot(   .... ratioline=True ... )`. 
2) marker and linestlye: Let's make those required for all backends (i.e. any backend needs to accept those, even if it might not use them), so that we can also make the defaults in the call to plot like so:
```
class RandomFancyPlot(Plot):
    def plot(self, overplot=False, clearwindow=True,
                  ratioline=True, marker='+',
                  **kwargs):
        Plot.plot(self, self.x, self.y, yerr=self.yerr, xerr=self.xerr,
                  title=self.title, xlabel=self.xlabel, ylabel=self.ylabel,
                  overplot=overplot, clearwindow=clearwindow, 
                  ratioline=ratioline, marker=marker,
              **kwargs)
```
3) all others. Could be a lot, but so far it's only capsize, barsabove, ecolor (all for error bars). However, those are, in the current pylab_backend implementation set to the matplotlib default in all but two cases. So I really just need a solution for those two cases. I'll think about that, and will push a suggestion when I have one.

hamogu:
Note to self: JointPlot also has matplotlib specific settings (wspace, hspace, ...), but that's a little bit of an odd one anyway.

hamogu:
next note: `get_data_hist_prefs` is part of "__init__.py" while all other `get_XXX_prefs` are in the backend code.

hamogu:
`CDFPlot` has  definitions in the class itself (not the plotting function and not the backend).

DougBurke:
Note to self to look at - I wrote a "backend switcher" pytest fixture to let us test "current backend" and "no backend" (when both are available) which are currently only used for the notebook tests. There are issues with using this because it wasn't "clean" - you could find references to the old backend even when switching to a new one. 

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/conftest.py#L547-L578

This should be fixed in this PR so can we use this fixture in more places? Do we need to rework the fixture?

Examples of it's use

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/astro/tests/test_astro_data_notebook.py#L65-L84

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/tests/test_data_notebook.py#L30-L53

https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/plot/tests/test_plot_notebook.py#L64-L77

hamogu:
Just a note on @DougBurke note:
I have a general testing (all backends) here: https://github.com/sherpa/sherpa/pull/1356/commits/437b0b45f254faaeb721fe8adb8f70625c4e98aa#diff-b5a57fe7261c2e0386005771e11e622236f9ea9f959a934ba6ac007d12755a59R120

and a bokeh specific fixture (similar to the pylab tests that @DougBurke linked above) here:
https://github.com/sherpa/sherpa/pull/1356/commits/437b0b45f254faaeb721fe8adb8f70625c4e98aa#diff-93e04ce55f23af3b9d75f1493e108936d12afb48d402e799afaa9acca85d2b3bR32

So, to answer the question that @DougBurke said: I think we should rework the code to add a parameter (`@plot_backend("pylab")`) so we can use it for any backend. Then we'll also have an "all_backends", but we don't need a no backend any longer (except to test that setting the backend to something crazy like "5" or "dadada" raises a reasonable error) , because the dummy_backend will always be available. We can bike-shed if this should be a pytest fixture or a decorator. Either way, I see a use case for also having a context manager `with plottingbackend('pylab'):` that's exposed to users and then we could also use that in tests (similar to how we have the `with sherpa.utils.SherpaVerbosity` that's fur users but also used in testing). 

Note however, that in #1375 I suggest that backends should be classes, not modules, so that will be slightly (a really only ever so slightly) different for this testing code.

hamogu:
Trivial rebase done. CI passes, ready for the review. (@Marie-Terrell still listed as reviewer, although those parts that needed CM review are no longer included in this PR. I'm still happy to have her comments and I also can't figure out how to remove someone, whose review had been requested in the past.)

sherpa/plot/dummy_backend.py
```diff
@@ -16,6 +16,17 @@
...
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+'''A dummy backend for plotting.
+
+This backend implements only minimal functionality (some formatting of
+strings as HTML or LaTeX which are usually used as axis labels), but no real
+plotting capabilities. It is here to ensure that the `sherpa.plot` module can
+be imported, even if no plotting backend is installed.
+
+In this sense, this backend can be understood as the "base" for backeneds.
```
wmclaugh, line 26:
```suggestion
In this sense, this backend can be understood as the "base" for backends. 
```

sherpa/plot/dummy_backend.py
```diff
@@ -16,6 +16,17 @@
...
+'''A dummy backend for plotting.
+
+This backend implements only minimal functionality (some formatting of
+strings as HTML or LaTeX which are usually used as axis labels), but no real
+plotting capabilities. It is here to ensure that the `sherpa.plot` module can
+be imported, even if no plotting backend is installed.
+
+In this sense, this backend can be understood as the "base" for backeneds. 
+The string-formatting is implemented here so that other backends don't have to
+dublicate that; they can call the functions here.
```
wmclaugh, line 28:
```suggestion
duplicate that; they can call the functions here.
```

sherpa/plot/__init__.py
```diff
@@ -26,8 +26,30 @@
...
+
+When this module is first imported, Sherpa tries to import the
+backends installed with Sherpa in the order listed in
+`sherpa.plot.try_backends`. The first module that imports
+successfully is set as the active backend. The following command prints the
+name and the location on disk of that module::
+
+   >>> from sherpa import plot
+   >>> print(plot.backend)
+
```
DougBurke, line 40: This now conflicts with the preceding paragraph (I tried to include it in this but as it is unchanged GH decided not to let me), which says

> Classes provide access to common plotting tasks, which is done by the plotting backend defined in the ``options.plot_pkg`` setting of the Sherpa configuration file.

So do no-longer allow the user to specify the backend?

sherpa/plot/__init__.py
```diff
@@ -26,8 +26,30 @@
...
+------------------
+
+After the initial import, the backend can be changed by loading one of
+the plotting backends shipped with sherpa (or any other module that
+provides the same interface):
 
+  >>> import sherpa.plot.pylab_backend
+  >>> plot.backend = sherpa.plot.pylab_backend
+
+"""
```
DougBurke, line 51: Is the following now true (text probably needs a re--work)?  If so we should review the docs/ and notebooks. 

Actually, thinking about this, this is only really relevant for users writing "library-style" code rather than someone at a notebook prompt plotting things. Which means the text definitely needs work...

```suggestion

Using the plot classes
----------------------

To allow the backend to be changed, the classes from this module should always be
referenced explicitly with the full path, rather than importing the names directly.
That is:

    >>> from sherpa import plot

and then refernce the objects directly (e.g. `plot.DataPlot`) rather than importing
it directly like

    >>> from sherpa.plot import DataPlot

"""
```

sherpa/plot/dummy_backend.py
```diff
@@ -36,14 +47,19 @@
...
+           'initialize_plot', 'select_plot')
 
 
 lgr = logging.getLogger(__name__)
 
 # Identify the backend
 name = 'dummy'
 
+warning = lgr.warning
+warning('Failed to import usable sherpa.plotting backends.' +
```
DougBurke, line 60:
```suggestion
lgr.warning('Failed to import usable sherpa.plotting backends.' +
```

Not exactly the most-important comment ever

DougBurke:
With this PR I get the following error:

```
% python -c 'from sherpa import plot'
  File "/home/dburke/anaconda/envs/sherpa-master/lib/python3.8/site-packages/sherpa/plot/__init__.py", line 257, in <module>
    class Plot(NoNewAttributesAfterInit):
  File "/home/dburke/anaconda/envs/sherpa-master/lib/python3.8/site-packages/sherpa/plot/__init__.py", line 259, in Plot
    plot_prefs = backend.get_plot_defaults()
AttributeError: 'NoneType' object has no attribute 'get_plot_defaults'
```

This is with an old rc file, which will be a common occurrence for users who try to use this the first time they try to use a version of CIAO with this code:

```
% python
Python 3.8.12 (default, Oct 12 2021, 13:49:34)
[GCC 7.5.0] :: Anaconda, Inc. on linux
>>> from configparser import ConfigParser
>>> from sherpa import get_config
>>> config = ConfigParser()
>>> config.read(get_config())
['/home/dburke/.sherpa-standalone.rc']
>>> config.get('options', 'plot_pkg', fallback='WAT')
'pylab'
>>>
```

Note that this is `pylab` and not `pylab dummy`.

The problem is that I don't actually have matplotlib installed:

```
>>> import matplotlib
ModuleNotFoundError: No module named 'matplotlib'
```

In `sherpa/plot/__init__.py` do we need to add, after

```
for plottry in plot_opt:
    try:
        backend = importlib.import_module('.' + plottry,
                                              package='sherpa.plot')
        break
    except ImportError:
        pass
```

something like the following untested code

```
if backend is None:
    # dummy_backend will report a "plottig is not supported" message
    backend = importlib.import_module('.dummy_backend', package="sherpa.plot')
```

I think this should also catch the case where the user has given an unsupported backend name - e.g. "matplotlib" rather than "pylab".

sherpa/plot/dummy_backend.py
DougBurke, line 60: Ohh. There's also a tiny typo

```suggestion
warning('Failed to import usable sherpa.plotting backend.' +
```

sherpa/plot/pylab_backend.py
```diff
@@ -211,9 +205,9 @@ def histo(xlo, xhi, y, yerr=None, title=None, xlabel=None, ylabel=None,
           overplot=False, clearwindow=True,
           xerrorbars=False,
           yerrorbars=False,
-          ecolor=_errorbar_defaults['ecolor'],
-          capsize=_errorbar_defaults['capsize'],
-          barsabove=_errorbar_defaults['barsabove'],
+          ecolor=None,
+          capsize=None,
+          barsabove=False,
```
DougBurke, line 210: Is the idea here being that the defaults for `plt.errorbar` are unlikely to change so we can just hard-code them?

If you git-blame the `_errorbar_defaults` setting you can see it was written a long time ago, in a galaxy far away, when matplotlib was significantly younger, so it may have made sense to make the values dependent on the backend.

This is also the only "plot object" that has this sort of functionality, so I don't think oit's a bad change, just checking what the thinking was.

sherpa/astro/plot.py
```diff
@@ -165,17 +162,17 @@ def prepare(self, data, stat=None):
...
     The filtering and grouping from the PHA datset
     are used to create the bins for the model.
     """
 
-    histo_prefs = backend.get_model_histo_defaults()
+    histo_prefs = shplot.backend.get_model_histo_defaults()
 
     def __init__(self):
-        HistogramPlot.__init__(self)
+        super().__init__()
```
DougBurke, line 175: :+1: 

We have a lot of old code that doesn't use super() so it's nice to update them when we get the chance!

DougBurke:
We could update `sherpa.plot.__init__.merge_settings` to be just

    return {**prefs, **user}

sherpa/plot/__init__.py
hamogu: It's in the docstring of the module, because that a place where normal users won't see it, but it's likely that future developers (or very advanced users who write their own library-style plot classes) find it.

I plan to make narrative docs with section headings like "for developers: How to implement a plotting backend" and "for developers and advanced users: How to define your own new plot type". However, we already agreed that we'll change the backend modules to backend classes in (my step 1) after this is merged, so I don't want to invest too much time in making detailed docs for something that's going to be out of date literally next week. 

That's one of the drawbacks of chopping the implementation into a series of PRs. I'll reword the text here to that it's not wrong, but I won't make it "good". But please keep that documenation issue in mind when you review the following PRs, because the docs clearly will have to be flashed out more and more over time.

sherpa/astro/data.py
```diff
@@ -283,15 +283,16 @@ def _extract_fields(obj, summary):
 def html_pha(pha):
     """HTML representation: PHA"""
 
-    from sherpa.astro.plot import DataPHAPlot, backend
+    from sherpa.astro.plot import DataPHAPlot
+    from sherpa import plot
 
     ls = []
 
     plotter = DataPHAPlot()
```
DougBurke, line 291: I guess it's okay to have

    plotter = DataPHAPlot()

here rather than an explicit symbol like `sherpa.astro.plot.DataPHAPlot` since we've just imported the symbol in this routine (I don't know the rules to know if Python caches this or will re-import the symbols)?

I would like to think the tests would catch this (if it wasn't working) but things are a bit squirrely so I'm not 100% sure of it.

sherpa/plot/pylab_backend.py
hamogu: It's a step towards allowing full functionality without matplotlib present. `_errorbar_defaults` looks at the signature of `plt.errorbar`, so `plt` must be around for that to work. I want to move to (but this is just one of the tiny first steps that way) a set of defaults that don't require importing a plotting library first. For the reasons you listed above, hardcoding it here is fine in my opinion, and `ecolor=None` actually means "default" (as defined in the matplotlib style file) in matplotlib language, so this is also a way to allow users to influence our plots with matplotlic style files if they want to (again, that's just a thought that that's a good direction to go in, I'm not claiming that these three lines of code are a full solution)

sherpa/astro/data.py
hamogu: I'm not 100% sure either. I'll have a look. However, I find that testing this in a good way is hard without a second functioning backend. Sure, I can use the Dummy backend, but if I don't see anything, I'm not quite sure if that's because it silently failed or because it (correctly) did not produce a plot. So, while I'm going to look into this a little more here, there might be a few cases like that that are not fully resolved here, but that certainly will pop up when the full bokeh backend is merged in and tested. 

The important point is that this is not doing any harm. It worked before with the pylab backend, and it's going to continue to work with one backend in the future. The only thing that might happen is that this PR does not fully live up to it's promise - elimination 100% of the imports that hinder backend-switching. BUT, that's not introducing any new problems, it's just "not fixing all existing problems". If I can't figure that out here, I'll add an issue to make sure I'll get back to it when I write/update the step 1 code.

hamogu:
Note: This PR claims to add new functionality, but this functionality is not really tested in here, as it says in the summary above. I will add real, reliable testing in the next step (links to proto-types are listed here: https://github.com/sherpa/sherpa/pull/1191#issuecomment-986918444). I tried to write good tests with backends and modules and it's just much harder than with backends as classes (one more reason why I want to move that way), so I don't want to spend time on writing complicated tests that are absolute literally next week. For this PR, I suggest the following procedure: We test interactively (as @DougBurke has done above) and I fix all that comes up that way, then we merge and revisit after converting the backend to classes. There is already a PR open for that, so this is not a case where we risk getting stuck with an incomplete implementation.

sherpa/plot/pylab_backend.py
DougBurke: Getting away from the actual change here,but I'm confused. This module requires matplotlib to exist, so I don't see how it helps in your overall goal of decoupling from the specific backend, but maybe that needs to come from a follow-on PR.

sherpa/astro/data.py
DougBurke: I am wondering if there's a step between stages 0 and 1 in your master plan to stop and review the plotting tests, perhaps re-arranging some of them and looking at the backend-switcher stuff?

hamogu: The testing will be in Step 1 A (turn backends into classes). Step 1 B will be "add bokek class".

DougBurke:
For the AttributeError case I mentioned above this did indeed seem to work, both for when pylab is the only option but is not installed, or for if I made a typo (e.g. pyla rather than pylab).

```
if backend is None:
    from . import dummy_backend as backend
```

The other option would be something like

```
plot_opt = plot_opt + ["dummy_backend"]
```

before the loop.

sherpa/plot/pylab_backend.py
hamogu: Mm, @DougBurke is right. What I said above makes no sense, matplotlib is (and always will be) required for this backend. I'm sorry - I wrote this code many months ago and must have mixed up up this change with what I had thought about for some other change. However, we agree that this is not a bad change, even if I can't recall exactly why I did it initially, so I'll leave it. At the very least (and that might have well been my motivation) it simplifies the module by removing and additional look-up that one has to track mentally.

sherpa/plot/dummy_backend.py
DougBurke: Just to note this typo (`backends` -> `backend`) is still present (but the line has changed so I don't think the suggestion can be applied directly).

sherpa/astro/data.py
hamogu: Yes, I thought about this more. It is OK to import `DataPHAPlot` because that is a Sherpa plot object, not a backend. The Plot objects are (or at least will be) backend-indenpendent. The change here is that it previously also imported "backend" which we don't want. It should use "plot.backend" instead.

hamogu:
@DougBurke @wmclaugh I believe I've done the changes you suggested; please check.

wmclaugh:
@Marie-Terrell - when you get a chance please assess the impact of merging this on patches for the CIAOD builds.

Marie-Terrell:
> @Marie-Terrell - when you get a chance please assess the impact of merging this on patches for the CIAOD builds.

Just checked. The patches are fine against these changes.

DougBurke:
[this is the "I wanted to look at an issue" comment I made in yesterday's meeting]

Somewhere I seem to remember an issue I had when trying to test some plotting code where changing the backend could lead to issues because some class or object was remembering the previous backend. Unfortunately I can't find any evidence from the code - my best bet was that it was in #631 - where I added the `override_plot_backend` test fixture but I couldn't see anything from this or anything obvious in the tests.

hamogu:
I'll add significant tests in the next step (when there are > 1 backends with some functionality; see discussion above); if this issue really exists, then I'll find (and fix) it at that point. I suggest to not hold up this PR for something that we can't find "any evidence for in the code".

DougBurke:
Not to be addressed here (as it is existing behavior) but I'll mention it as part of the testing and we can create an issue if we feel it's needed.

If there's no available plotting backend then

1. the warning message is technically incorrect
2. we don't get any user-actionable information when a plot command is called

With this PR (but it would be similar with the current main branch) here is point 1:

```
% python
Python 3.8.12 (default, Oct 12 2021, 13:49:34)
[GCC 7.5.0] :: Anaconda, Inc. on linux
>>> from sherpa import plot
```

It says that the plotting routines are not available. Technically they are available, it's just that they do nothing.

And now for point 2:

```
>>> plot.backend.name
'dummy'
>>> load_arrays(1, [1, 2, 3], [4, 5, 6])
>>> set_source(const1d.foo)
>>> plot_fit()
>>>
```

The `plot_fit` command has returned but there's no indication to the user that it hasn't done anything.

Issue 1 is easy to address, with a tweak to the warning message. I am not-at-all-sure we should do anything about point 2 for various reasons, including-but-not-limited-to

- causing the `plot_xxx` commands to error out would be annoying (e.g it would make it hard to run scripts when you have no plot backend available); technically also we'd probably want the error to be raised by the plot object's `plot` method rather than in the UI `plot_xxx` routine
- creating a warning message each time a `plot_xxx` command is called (or, the plot object's `plot` method is called) would get old real quickly

sherpa/plot/__init__.py
```diff
@@ -22,12 +22,34 @@
...
 
-"""
+Which backend is used?
+----------------------
+
+When this module is first imported, Sherpa tries to import the
+backends installed with Sherpa in the order listed in
+the ``sherpa.rc`` startup file. The first module that imports
+successfully is set as the active backend. The following command prints the
+name and the location on disk of that module::
```
DougBurke, line 36:
```suggestion
backends installed with Sherpa in the order listed in the
``options.plot_pkg`` setting from the ``sherpa.rc`` startup file. The
first module that imports successfully is set as the active
backend. The following command prints the name and the location on
disk of that module::
```

The copyright year also needs updating but I can't seem to do this via the GH page.

sherpa/sherpa-standalone.rc
```diff
@@ -1,7 +1,7 @@
 # SHERPA_VERSION 4.14.1
```
DougBurke, line 1: This should probably get changed to `4.14.2` but I don't know what the rule we have been using for this (it is purely decorative).

sherpa/sherpa.rc
```diff
@@ -1,7 +1,7 @@
 # SHERPA_VERSION 4.14.1
```
DougBurke, line 1: This should probably get changed to 4.14.2 but I don't know what the rule we have been using for this (it is purely decorative).

[need a clone-comment button...]

hamogu:
I created two new issues and added tracking them in the same project (https://github.com/sherpa/sherpa/projects/6). I believe I'll address both of them in the next step. For the point 2, we should discuss details in the new issue, but briefly, I suggest that we do use a warning (because an error is too much) and try how that goes in real work. I think in practice, most users don't call plots that done display anything so often that seeing that warning multiple times is a real problem. If it is, we can use https://docs.python.org/3/library/warnings.html#the-warnings-filter and set a default warning filter *for this specific warning* (don't change the global state for all warnings!) to appear only once.

sherpa/plot/__init__.py
hamogu: Maybe it's finally time that I set up a git pre-commit hook to do that  automatically!

DougBurke (approved):
Let's do this.

PR #1192, opened 2021-06-17
Remove Python 3.6 and add Python 3.9

Marie-Terrell:
This update includes:

- Xspec re-builds to support readline 8.1 (needed to support Python 3.9)
- Removal of the recent workaround added in PR#[1187](https://github.com/sherpa/sherpa/pull/1187). This is now fixed with the xspec readline updates above (allows us to grab more recent micro versions of Python from Conda without the issue).
- The GitHub Actions pip and Conda tests and the Conda builds have been updated to support NEP-29 for July 26th (Python 3.7+ and numpy 1.18+). 
- Though the Conda build meta.yaml was a little out of date. It had a python 2.7 fallback that has now been switched to 3.7. The minimum numpy support for the Sherpa Conda builds have been bumped to 1.18.* for Python 3.7/3.8 and 1.19.* for Python 3.9 (to avoid conflict).

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -27,9 +27,18 @@ jobs:
           - name: Linux Minimum Setup 
             os: ubuntu-latest
             python-version: 3.8
-            numpy-version: 1.17
+            numpy-version: 1.18
             install-type: develop 
             test-data: none
+
+          - name: Linux Full Build (Python 3.9)
+            os: ubuntu-latest
+            python-version: 3.9
+            install-type: develop
+            fits: astropy
+            test-data: submodule
+            matplotlib-version: 3
+            xspec-version: 12.11.1
 
           - name: Linux Full Build (Python 3.8)
             os: ubuntu-latest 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-conda-workflow.yml
```diff
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   sherpa_channel: sherpa
-  xspec_channel: "xspec/channel/test"
+  xspec_channel: "xspec/label/test"
```
DougBurke, line 7: Is this an intended change or should it go back to `xspec/channel/test`?

.github/workflows/ci-conda-workflow.yml
```diff
@@ -43,33 +52,24 @@ jobs:
...
+            python-version: 3.9 
             install-type: install
             test-data: package
-            matplotlib-version: 2
+            matplotlib-version: 3
 
           - name: Linux Build (w/o Matplotlib, Xspec, or test data)
             os: ubuntu-latest
             python-version: 3.7
+            numpy-version: 1.18
```
DougBurke, line 72: Is the idea here that we are now being "more declarative" in our setup, by explicitly mentioning the required versions of packages?

The only place I could see this being an issue would be if we had a "forward facing" test run where we just want to pick the latest versions available to catch possible problems early - but that's not the place of this particular test.

.github/workflows/ci-conda-workflow.yml
Marie-Terrell: So, the "channel" designation is actually the old designation and "label" is the new one (for a while now) for Anaconda cloud. They are interchangeable for now, though I erred on the side of the new one in case the old one is eventually deprecated.

.github/workflows/ci-conda-workflow.yml
Marie-Terrell: I was using this more as a "minimum support version" test case with some of the other builds being more exploratory

.github/workflows/ci-conda-workflow.yml
DougBurke: Or was this a change from #1187 that is now being reverted?

Marie-Terrell: Nope. Reverting that change was removing the "_CONDA_PYTHON_SYSCONFIGDATA_NAME" variable settings.

.github/workflows/ci-conda-workflow.yml
DougBurke: This was me mainly thinking out wide, in case anything I was thinking made (or didn't make) sense...

DougBurke (approved):
LGTM

PR #1194, opened 2021-06-20
datastack: Improve show_stack and add _repr_html_ for stack

hamogu:
# Summary
Improve datastack display (via `show_stack` and in the notebook) for datastacks.

# Details
The previous code listed only the OBSID and MJD_OBS and the name of the datstack. This works well when using a datstack for a time series of spectra. However, for other use cases where all spectra in the datstack from from the observation (and sometimes even form the same PHA II file) that display is useless. In particular for grating data (where HEG and MEG positive and negative orders all have the same OBSID and MJD_OBS)  or polarimetry data (IXPE) all displayed values are identical - not very useful.
Now, the table is constructed from a longer list of header keywords, but only keywords that are present and actually differ between rows are displayed; the list of keywords used can be changed as a module-level variable if needed.
I also added a html represenation for notebook display.

# Notes
The previous implementation went to great length to unify treatment for files that have MJD_OBS or MJD-OBS set. CIAO changed that header keywords a few years ago. I don't think that complexity is warranted any longer. I just treat the columns separately. In practice, cases where people mix files from a 3 year old CIAO version with a current CIAO version in the same datastack are going to be rare and columns with non-existent keywords are not displayed, so users will see either MJD-OBD or MJD_OBS but not both.

# Example
old:
Screen Shot 2021-06-20 at 6 51 43 AM

new:
Screen Shot 2021-06-20 at 6 52 04 AM

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -367,7 +367,7 @@ def html_pha(pha):
                           ('HDUCLAS2', 'Data stored'),
                           ('HDUCLAS3', 'Data format'),
                           ('HDUCLAS4', 'PHA format'),
-                          ('XFLT0001', 'Stokes (polarimetry)')])
+                          ('XFLT0001', 'XEPC filter 0001')])
 
     if meta is not None:
         ls.append(formatting.html_section(meta, summary='Metadata'))
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -366,7 +366,8 @@ def html_pha(pha):
                           ('CHANTYPE', 'The channel type'),
                           ('HDUCLAS2', 'Data stored'),
                           ('HDUCLAS3', 'Data format'),
-                          ('HDUCLAS4', 'PHA format')])
+                          ('HDUCLAS4', 'PHA format'),
+                          ('XFLT0001', 'Stokes (polarimetry)')])
 
     if meta is not None:
         ls.append(formatting.html_section(meta, summary='Metadata'))
--- a/sherpa/astro/datastack/ds.py
+++ b/sherpa/astro/datastack/ds.py
@@ -138,7 +138,7 @@ class DataStack():
         header : list of strings
             Column names
         data : list of lists
-            Eahc list is a row, corresponding to one dataset
+            Each list is a row, corresponding to one dataset
 
         '''
         tab = []
--- a/sherpa/astro/datastack/ds.py
+++ b/sherpa/astro/datastack/ds.py
@@ -20,6 +20,7 @@
 import numpy
 from sherpa.utils.logging import config_logger
 from sherpa.utils.err import IOErr
+from sherpa.utils.formatting import html_table, html_from_sections
 from sherpa.astro import ui
 from .utils import load_error_msg, load_wrapper, model_wrapper, \
     simple_wrapper, fit_wrapper, plot_wrapper
@@ -34,6 +35,18 @@ except ImportError:
 
 # Global list of dataset ids in use
 _all_dataset_ids = {}
+
+keys_to_display = ['OBJECT', 'TELESCOP', 'OBS_ID', 'INSTRUME', 'FILTER',
+                   # Standard is MJD_OBS, but older CIAO used 'MJD-OBS'
+                   # Usually, we'll have the one of the other with few cases
+                   # of mixed occurance. Thus, we can just treat them as
+                   # separate columns
+                   'MJD_OBS', 'MJD-OBS',
+                   # grating spectroscopy
+                   'GRATING', 'ORDER', 'TG_M',
+                   # polarimetry
+                   'XFLT0001']
+'''List of keys displayed with show_stack (if present in header)'''
 
 
 class DataStack():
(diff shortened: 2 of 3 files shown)


sherpa/astro/datastack/ds.py
```diff
@@ -108,47 +121,70 @@ def clear_stack(self):
...
+        cases, only a few of a relatively long list of keywords will
+        differ between datasets, printing out only the relevant
+        information.
+
+        Returns
+        -------
+        header : list of strings
+            Column names
+        data : list of lists
+            Eahc list is a row, corresponding to one dataset
```
wmclaugh, line 141: Eahc -> Each

sherpa/astro/datastack/tests/test_datastack.py
```diff
@@ -723,134 +724,60 @@ def test_default_instantiation(ds_setup):
...
     lines = captured.out.split('\n')
-    assert len(lines) == 3
-    assert lines[0] == l0
-    assert lines[1] == l1
-    assert lines[2] == ''
-    assert captured.err == ''
+    assert len(lines) == 2
+    assert 'empty datastack' in lines[0]
+    assert lines[1].strip() == ''
+     # Now, the check the html represention
```
wmclaugh, line 742: represention -> representation

sherpa/astro/data.py
```diff
@@ -366,7 +366,8 @@ def html_pha(pha):
                           ('CHANTYPE', 'The channel type'),
                           ('HDUCLAS2', 'Data stored'),
                           ('HDUCLAS3', 'Data format'),
-                          ('HDUCLAS4', 'PHA format')])
+                          ('HDUCLAS4', 'PHA format'),
+                          ('XFLT0001', 'Stokes (polarimetry)')])
```
DougBurke, line 370: Technically I think that the `XFLTxxxx` keywords are generic - that is you could have one in a non-polarimetry dataset.

sherpa/astro/datastack/ds.py
```diff
@@ -108,47 +121,70 @@ def clear_stack(self):
...
         for dataset in self.filter_datasets():
-            obsid = "N/A"
-            time = "N/A"
-            timekey = 'MJD-OBS'
+            row = [dataset['id'], dataset['data'].name]
             try:
                 hdr = dataset['data'].header
+                row.extend([hdr.get(k, '') for k in keys_to_display])
+            except AttributeError:
+                row.append([''] * len(keys_to_display))
```
DougBurke, line 151: I'm guessing this is to check for obects which don't have a header? Your recent changes mean that we don't getthis by default, but I guess a user could do something like `pha.header = None`.

sherpa/astro/datastack/ds.py
```diff
@@ -108,47 +121,70 @@ def clear_stack(self):
...
-                                                      'OBS_ID', obsid,
-                                                      timekey, time))
+        A brief summary of each data set in the stack is printed to the
+        standard output channel. The information displayed depends on the
+        type of each data set.
+        """
+        header, data = self._summary_for_display()
+        print('|'.join(header))
+        for row in data:
+            print('|'.join(row))
```
DougBurke, line 187: This has a different feel to the other `show_xxx` routines, which tend to use more spaces, but I think the existing routines have a varied enough approach that we don't have a consistent guide.

sherpa/astro/datastack/ds.py
```diff
@@ -108,47 +121,70 @@ def clear_stack(self):
...
+        """Return a HTML (string) representation of the stack
+        """
+        header, data = self._summary_for_display()
+        htab = html_table(header, data, rowcount=False,
+                          summary=f'datastack with {data.shape[0]} datasets')
+        return html_from_sections(self, htab)
 
-            except AttributeError:
-                pass
+    # QUS: should this use the logger instead of print?
```
DougBurke, line 176: The existing `show_xxx` routines all use `sherpa.utils.send_to_pager` [at least those in sherpa[.astro].ui.utils do] so we should probly do the same.  I'm not a great fan of `send_to_pager` but its what we do...

sherpa/astro/datastack/ds.py
DougBurke: [This is not a request to change code, just so I can understand it as there's no test coverage]

sherpa/astro/data.py
DougBurke: So I'm not sure what we do here.

sherpa/astro/datastack/ds.py
DougBurke: [a general comment, not a request for changes]

sherpa/astro/datastack/ds.py
DougBurke: Of course, I don't know how the `show_xxx` routines work in the noteook/lab interface.

DougBurke (changes requested):
Deciding what to do with XFLT and whether send_to_pager works in the notebooks are the issues I see.

sherpa/astro/datastack/ds.py
hamogu: Yes.

sherpa/astro/datastack/ds.py
hamogu: This is not new code, it's merely moved to a different line: https://github.com/sherpa/sherpa/pull/1194/files#diff-1b2333a505749af538cd4ba44a468fce606ee4cea032f1e6544ce66aa85b3debL111

This `shows_stack` function did not `send_to_pager` before and it does not do so now. All I have changed in the content of the message. (And there is some refactoring, because `_repr_html` needs similar information.
So, just to clarify: Do you think I should bring is in-line with what the other `show_xxx` functions do because I'm touching this function now?
I don't know what the original motivations was to use or not use `send_to_pager` in `show_stack`.

hamogu:
I've corrected the label for XFLT and modified `show_stack` to use `send_to_pager` in line with what the other `show_xxx` functions do, addressing all points from the review.

hamogu:
The red checkbox is only a marginal decrease in codecov. I believe that this PR is ready for re-review by @DougBurke and (if approved) can be merged regardless.

sherpa/astro/data.py
```diff
@@ -366,7 +366,8 @@ def html_pha(pha):
                           ('CHANTYPE', 'The channel type'),
                           ('HDUCLAS2', 'Data stored'),
                           ('HDUCLAS3', 'Data format'),
-                          ('HDUCLAS4', 'PHA format')])
+                          ('HDUCLAS4', 'PHA format'),
+                          ('XFLT0001', 'XEPC filter 0001')])
```
DougBurke, line 370:
```suggestion
                          ('XFLT0001', 'XSPEC filter 0001')])
```

sherpa/astro/datastack/tests/test_datastack.py
```diff
@@ -723,134 +724,60 @@ def test_default_instantiation(ds_setup):
...
-       If None the value is missing.
-    key2 : str or None
-       The keyword being used for the "MJD OBS" value: file 2.
-       If None the value is missing.
-
-    """
+def test_show_empty_datastack(capsys):
+    '''Test output for an empty datastack.
 
+    For a change, we try the OO interface here/
```
DougBurke, line 730:
```suggestion
    For a change, we try the OO interface here.
```

DougBurke: The mostest important fix.

DougBurke (changes requested):
Just the two typos.

The original datastack code was developed outside Sherpa by @taldcroft and then brought into Sherpa so I imagine we never really looked at how `show_stack` was implemented when we did so.

hamogu:
Fixed the typos according to @DougBurke suggestions about 30 s ago, so this is hopefully ready. Asking @DougBurke for final review.

DougBurke:
Just waiting for the tests to pass.

hamogu:
Rebased to pick up recent changes in main.

DougBurke:
Why are these tests still waiting to run? eg https://github.com/sherpa/sherpa/actions/runs/1072355480

DougBurke:
```
Can't find any online and idle self-hosted or hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-latest'
Waiting for a self-hosted or a hosted runner to pickup this job...
```

That doesn't look great...I'm going to cancel the conda runs (as none of them have run) and see if we can restart them, assuming this is a transient failure on GitHub's side.

DougBurke (approved):

PR #1199, opened 2021-06-30
Move the regrid code out into a separate method

DougBurke:
# Summary

Very minor code reorganization for the regrid code.

# Details

The idea is to make it easier for sub-classes to access the regrid code, but all this does is move the logic out into a separate routine *and* cleans up the logic a little bit. There is no functional change in this PR.

This was part of #830 but I think it is a very-minor improvement in itself purely for the logic of `_evaluate`.

### DIFF
--- a/sherpa/models/regrid.py
+++ b/sherpa/models/regrid.py
@@ -486,8 +486,9 @@ class ModelDomainRegridder1D():
     def calc(self, pars, modelfunc, *args, **kwargs):
         """Evaluate and regrid a model
 
-        Evaluate the model on the internal grid and then interpolate
-        onto the desired grid.
+        Evaluate the model on the internal grid and then convert it
+        onto the desired grid either preserving flux (rebinning) or
+        via interpolation.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


sherpa/models/regrid.py
```diff
@@ -586,6 +586,21 @@ def eval_non_integrated(self, pars, modelfunc, data_space, eval_space,
 
         return y
 
+    def eval_integrated(self, pars, modelfunc, data_space, eval_space,
+                        **kwargs):
+        """Evaluate a grid with low and high bins.
+
+        Note that data_space and eval_space are assumed to have two
+        elements (low- and high- edges).
```
hamogu, line 594: This function is called with `data_space.grid`, yet the parameter is simply called `data_space`. I think that's confusing. I would suggest to actually pass in the dataspace itself (which is closer to how it was before this was separated to its own function) to avoid that confusion.

hamogu (changes requested):

sherpa/models/regrid.py
DougBurke: An alternative is to rename these parameter names. The reason I suggest this is that if you look at `eval_non_integrated` (which isn't changed in this PR) it is also sent the "actual" grid - via `data_space.midpoint_grid` - which makes it clear at the `_evaluate` call sight that it is different to the integrated case.  

So I propose

```
    def eval_non_integrated(self, pars, modelfunc, data_grid, eval_grid,
                            **kwargs):

    def eval_integrated(self, pars, modelfunc, data_grid, eval_grid,
                        **kwargs):
```

How does that sound?

DougBurke: And improving the docstrings for these these methods...

hamogu: That's a good solution, too.

DougBurke: I've rebased this PR to use add docstrings and use a more-consistent naming scheme for the parameters

hamogu (approved):

Issue #1212, opened 2021-07-16
What to do about non-integer channel limits in PHA filters?


DougBurke:
Going down the rabbit-hole on #1129 is taking me to surprising places... Today's adventure is the realization that we should decide on what non-integer channel values mean in a filter expression. I would be surprised if any user used non-integer values for the channel limits, so whatever we do should be okay as long as we document it, but it's technically possible (and we have some tests that do so, for better or worse).

In CIAO 4.13 we have

```python
sherpa In [1]:     x = np.asarray([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
          ...:     y = np.asarray([0, 0, 0, 2, 1, 1, 0, 0, 0, 0])
          ...:
          ...:     pha = DataPHA('416', x, y)
          ...:     pha.notice(3.5, 6.5)

sherpa In [2]: pha.mask
Out[2]:
array([False, False, False,  True,  True,  True, False, False, False,
       False])

sherpa In [3]: x[pha.mask]
Out[3]: array([4, 5, 6])
```

This PHA file has no response so the filtering is in channel units, and asking for `3.5 - 6.5` has returned us the channels 4, 5, and 6.

I could argue that really we should think of channels as "ranges", so channel 3 covers anything with a channel value >= 3 and < 4, in which case the filter above should really select channels 3, 4, 5, and 6. There are complications with this interpretation, which I'll mention below.

We could just say that any channel limit is converted to an integer (either nearest or we just take the floor), which would either select `4-6` or `3-6`.

We could make it an error to use non-integer channel limits. This would probably make the code easier, and is easy to understand/document, but it can be annoying in Python (although given that channels are integer values I think enforcing this constraint would be a good thing for anyone manipulating channels to recognize).

# Treating channels as a range

This is more for background, or potential future discussion as part of #1129, as for this issue it doesn't matter.

We want `notice(4, 6)` to correspond to channels 4, 5, 6 and at present it does so - e.g.

```python
sherpa In [4]: pha.get_filter()
Out[4]: '4:6'
```

If we were to treat channels as ranges then really we'd want the `get_filter` expression to return `4:7` if we follow #1129 (lower edge of the first bin to upper edge of the last bin), but then you are saying that `notice(4, 7)` matches channels `4, 5, 6`, but this contradicts the start of this section. This can be handled, but it probably adds special cases to the code which we don't want.

DougBurke:
[...]

Issue #1129, opened 2021-04-21
filter expressions for PHA datasets could be improved


DougBurke:
The filter expression for a PHA dataset reports the center of the bin when I think it would make sense to use the low edge (for lower) and high edge (for upper).

The following is with CIAO 4.13 but the exact version doesn't matter too much as this hasn't changed for a while:

```
sherpa In [1]: load_pha('3c273.pi')
WARNING: systematic errors were not found in file '3c273.pi'
statistical errors were found in file '3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file 3c273.arf
read RMF file 3c273.rmf
WARNING: systematic errors were not found in file '3c273_bg.pi'
statistical errors were found in file '3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file 3c273_bg.pi

sherpa In [2]: get_filter()
Out[2]: '0.124829999695:12.410000324249'

sherpa In [3]: ungroup()

sherpa In [4]: get_filter()
Out[4]: '0.008030000201:14.943099975586'
```

Notice that the filter expression has changed because I've changed from grouped to ungrouped data. The actual filtered data range (which is this case is the full dataset) hasn't actually changed, just the reported values.

```
sherpa In [5]: notice(0.5, 7)

sherpa In [6]: get_filter()
Out[6]: '0.503699988127:7.000699996948'
```

This seems sensible, but as shown below the limits 0.5037 and 7.0006 are actually the mid-points of the bins, and I claim that it would be better to report here (modulo number of decimal places):

    0.496399...:7.00799...

```
sherpa In [7]: pha = get_data()

sherpa In [8]: elo, ehi = pha._get_ebins()

sherpa In [9]: elo[pha.mask][0], ehi[pha.mask][0]
Out[9]: (0.49639999866485596, 0.5109999775886536)

sherpa In [10]: elo[pha.mask][-1], ehi[pha.mask][-1]
Out[10]: (6.9934000968933105, 7.007999897003174)

sherpa In [11]: (0.4964 + 0.5111) / 2
Out[11]: 0.50375

sherpa In [12]: (6.9934 + 7.0080) / 2
Out[12]: 7.0007
```

If we report the bin edges then the response to

    load_pha('3c273.pi')
    get_filter()
    ungroup()
    get_filter()

would be to return the same string for both `get_filter` calls (in this case it would be '0.0014600000577047467:14.950400352478027')

hamogu:
[...]

Issue #416, opened 2017-12-13
filter and then group lead to inconsistent results


olaurino:
The script below seems to give inconsistent results:

``` python
import numpy as np
from sherpa.astro import ui
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [0, 0, 0, 2, 1, 1, 0, 0, 0, 0]
ui.load_arrays(1, x, y, ui.DataPHA)
ui.notice(3.5, 6.5)
ui.group_counts(3)
print(ui.get_data().get_dep(filter=True))
np.testing.assert_equal([0, 0, 0, 0, 0, 2, 2, 2, 2, 2], ui.get_data().quality)
np.testing.assert_array_equal([3, 1], ui.get_data().get_dep(filter=True))
```

I would have expected the resulting array to be `[3, 1]`, with channels 6 to 10 having an associated quality value of 2. While the quality array is the one I would expect, the resulting filtered, grouped dependent array is empty, which does not seem to make sense.

**EDIT:** It looks like there is more to this bug, so I am expanding it.

First, one can get a reasonable result by using `ignore_bad`, which is surprising since `ignore_bad` claims to work in a different way, i.e. it should remove the filters... it even prints out a warning about it:

```python
import numpy as np
from sherpa.astro import ui
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [0, 0, 0, 2, 1, 1, 0, 0, 0, 0]
ui.load_arrays(1, x, y, ui.DataPHA)
ui.notice(3.5, 6.5)
ui.group_counts(3)
ui.ignore_bad()
print(ui.get_data().quality)
print(ui.get_data().get_dep(filter=True))
np.testing.assert_array_equal([3], ui.get_data().get_dep(filter=True))
```

The first obvious surprise is that ignoring some channels results in getting back good groups we wouldn't get without `ignore_bad`, as in the first case.

But there's more. `ignore_bad` only meets your expectations if you use the grouping function without tab stops, probably because the quality array changes in different ways when using tab stops: 

```python
import numpy as np
from sherpa.astro import ui
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [0, 0, 0, 2, 1, 1, 0, 0, 0, 0]
ui.load_arrays(1, x, y, ui.DataPHA)
ui.notice(3.5, 6.5)
tab_stops = ~ui.get_data().mask
ui.group_counts(3, tabStops=tab_stops)
[...]

Issue #288, opened 2016-10-15
An ignore filter for a partial channel is being ignored


DougBurke:
I am not sure if this is related to issue #149 but it looks a little different, so I'm opening a new ticket just in case. Is this a user expectation issue - i.e. you need to ignore all channels in a group for it to be ignored?

In the following (using CIAO 4.8 but also repeatable with master), I try to apply a filter to ignore channels (3.1, 4) but it is being ignored when the data is grouped, but not when it isn't. I use the low-level object API to create a `DataPHA` object with a simple grouping scheme.

If I ignore(3, 4) then the middle group is ignored (the middle group covers channels 3 to 4). If I instead use ignore(3.1, 4) then no bins are filtered.

```
sherpa-1> channels = np.arange(1, 6)
sherpa-2> counts = np.asarray([5, 5, 10, 10, 2])
sherpa-3> grouping = np.asarray([1, -1, 1, -1, 1], dtype=np.int16)
sherpa-4> d = DataPHA('x', channels, counts, grouping=grouping)
sherpa-5> d.mask
          True
sherpa-6> d.ignore(3, 4)
sherpa-7> d.mask
          array([ True, False,  True], dtype=bool)
sherpa-8> print("*** This is expected")
*** This is expected
sherpa-9> xlo = [1, 3, 5]
sherpa-10> xhi = [3, 5, 6]
sherpa-11> d.notice(None, None)
sherpa-12> d.mask
           True
sherpa-13> d.ignore(3.1, 4)
sherpa-14> d.mask
           array([ True,  True,  True], dtype=bool)
sherpa-15> print("*** This was not expected")
*** This was not expected
```

If I ungroup the data then ignore(3.1, 4) does ignore the bin.

```
sherpa-16> d.ungroup()
sherpa-17> d.mask
           array([ True,  True,  True,  True,  True], dtype=bool)
sherpa-18> d.notice(None, None)
sherpa-19> d.mask
           True
sherpa-20> d.ignore(3.1, 4)
sherpa-21> d.mask
           array([ True,  True,  True, False,  True], dtype=bool)
```

DougBurke:
I've added tests in #922, showing these tests, and how they get changed by #922

PR #1215, opened 2021-07-19
Restrict the values used in PHA notice/filter calls

DougBurke:
# Summary

Ensure that the low and high limits for notice and ignore calls for DataPHA objects are sensible (hi >= lo and that for energy or wavelength filters they are >= 0). When filtering DataPHA objects in channel units the lo and hi arguments must be integers otherwise an error is raised.

# Details

In working towards #1129 I realized that the code is a little cleaner if we reject invalid lo/hi values, so for a PHA dataset we now raise a DataErr for `notice` and `ignore` calls when both `lo` and `hi` are set:

 - hi < lo
 - lo or hi values < 0 when filtering in energy or wavelength space
 - non-integer values for channel filters (issue #1212)

**To think about** We could add similar restrictions for other data classes (at least that hi should be >= lo) and I don't know whether to add that here or not.

**To think about** I have used the `DataErr` "bad" error as there isn't really a direct match in DataErr. I could have invented a new error entry for this but I am not sure it's worth it.

Technically we could error out if the `hi` value <= 0 (and ditto for the `lo` value) for energy or wavelength filtering, but using 0 as the low value is a usability "fix" (it's a lot easier to say 0 than have to worry about what the minimum supported energy or wavelength is), and keeping the same check for `hi` is "okay". Using 0 for `lo` is really the same as using `None`, but it's less characters to type.

The non-integer channel error is the **largest usability change** and as discussed in #1212 we could still allow it but either using int(x) [round down for positive values] or pick the nearest integer. If we wanted to go this way (which would be easy to add here) I would vote for int(x) and not nearest integer since if we think of channel 2 really being 2 <= x < 3 then 2.9 should go to 2.

Most of the changes are actually tests for future changes to the filtering code (notice/ignore handling) but I wanted them in before I made those changes. Some of the tests - such as those added in #288 and #416 - have had to be adjusted because they used non-integer channel values. In these cases I've generally  tried to create two copies of the test: a) a version using integer channel values (which probably don't actually probe the original case being tested, and b) a version with a response added so we can keep the non-integer notice/ignore calls. Note that there isn't a way to change a non-integer channel filter to an energy filter and always keep the same lo/hi values and so some subtle changes to the tests have had to be made. A number of other tests have had to be changed because I wrote them to check the behavior when you give a negative energy/wavelength range to a notice/ignore call and so there the negative had to be replaced by a number like 0.

This PR is the second in a sequence: #1208, #1215, #1216, #1219 (hopefully this is all).

# Examples

## energy filters

Before this PR

```python
>>> load_pha('3c273.pi')
read ARF file 3c273.arf
read RMF file 3c273.rmf
>>> set_analysis('energy')
>>> notice(-5, 2)
>>> get_filter()
'0.124829999695:1.985600054264'
>>> notice()
>>> notice(-5, -5)
>>> get_filter()
'0.124829999695'
>>> notice()
>>> notice(20, 10)
>>> get_filter()
'12.410000324249'
```

Notice that the -5,-5 and 20,10 filters end up creating odd filter expressions (these are because of problems to be fixed in a later PR in this series but by error-ing out we don't have to worry about it).

With this PR everything fails...

```python
>>> load_pha('3c273.pi')
read ARF file 3c273.arf
read RMF file 3c273.rmf
>>> set_analysis('energy')
>>> notice(-5, 2)
  File "<string>", line 1, in notice
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/ui/utils.py", line 6759, in notice
    sherpa.ui.utils.Session.notice(self, lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 4739, in notice
    d.notice(lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 3718, in notice
    raise DataErr('bad', 'lo argument', 'must be >= 0')
sherpa.utils.err.DataErr: unknown lo argument: 'must be >= 0'
>>> notice(-5, -5)
  File "<string>", line 1, in notice
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/ui/utils.py", line 6759, in notice
    sherpa.ui.utils.Session.notice(self, lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 4739, in notice
    d.notice(lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 3718, in notice
    raise DataErr('bad', 'lo argument', 'must be >= 0')
sherpa.utils.err.DataErr: unknown lo argument: 'must be >= 0'
>>> notice(20, 10)
  File "<string>", line 1, in notice
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/ui/utils.py", line 6759, in notice
    sherpa.ui.utils.Session.notice(self, lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 4739, in notice
    d.notice(lo, hi, **kwargs)
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 3700, in notice
    raise DataErr('bad', 'hi argument', 'must be >= lo')
sherpa.utils.err.DataErr: unknown hi argument: 'must be >= lo'
```

## channel filters

Before this PR

```python
>>> import numpy as np
>>> d = DataPHA('x', np.asarray([1, 2, 3, 4, 5]), [1, 2, 3, 4, 5])
>>> d.notice(2.4, 4)
>>> d.get_filter()
'3:4'
>>> d.notice()
>>> d.notice(2, 3.5)
>>> d.get_filter()
'2:3'
```

Note that it's not immediately obvious what 2.5-4 and 2-3.5 are selecting: it's not nearest neighbor and it's not int(val) - so by error-ing out we ignore this headache! {it's actually using the Data1D rules but that's not directly relevant here but would be in future PRs in this series}.

After this PR both `notice` calls are an error

```python
>>> import numpy as np
>>> d = DataPHA('x', np.asarray([1, 2, 3, 4, 5]), [1, 2, 3, 4, 5])
>>> d.notice(2.4, 4)
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 3712, in notice
    raise DataErr('bad', 'lo argument', 'must be an integer channel value')
sherpa.utils.err.DataErr: unknown lo argument: 'must be an integer channel value'
>>> d.notice(2, 3.5)
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 3714, in notice
    raise DataErr('bad', 'hi argument', 'must be an integer channel value')
sherpa.utils.err.DataErr: unknown hi argument: 'must be an integer channel value'
```

### DIFF
--- a/sherpa/astro/tests/test_astro_data.py
+++ b/sherpa/astro/tests/test_astro_data.py
@@ -1999,22 +1999,22 @@ def test_energy_filter_roundtrip(make_data_path):
 
     pha = read_pha(make_data_path('3c273.pi'))
 
-    fall = pha.get_filter()
-    assert fall == '0.124829999695:12.410000324249'
+    fall = pha.get_filter(format='%.5f')
+    assert fall == '0.12483:12.41000'
 
     pha.notice(0.5, 7)
     pha.ignore(1, 2)
 
-    expected = '0.518300011754:0.970899999142,2.058600068092:8.219800233841'
-    f1 = pha.get_filter()
+    expected = '0.51830:0.97090,2.05860:8.21980'
+    f1 = pha.get_filter(format='%.5f')
     assert f1 == expected
 
     pha.ungroup()
-    f2 = pha.get_filter()
-    assert f2 == '0.474500000477:0.985500007868,2.036700010300:9.862299919128'
+    f2 = pha.get_filter(format='%.5f')
+    assert f2 == '0.47450:0.98550,2.03670:9.86230'
 
     pha.group()
-    f3 = pha.get_filter()
+    f3 = pha.get_filter(format='%.5f')
     assert f3 == expected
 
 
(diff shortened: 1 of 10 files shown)


hamogu:
I've thought about this for a while, but I think this is a good improvement. I'm not approving it, since it's will marked as WIP and I have not looked through the tests added here in detail.

I'll put down some notes below for the record:
- This is making the interface cleaner in the sense that some ambiguous stuff that used to work (non-integer channels) simply don't work anymore, so, as @DougBurke says, we don't have to document, test, etc. these cases. By not allowing this, we guide the user to do the right thing.
- I worry about simplicity for the user. I can imagine cases, where I look at channel 50:99 and then half the values (50/2:99/2). That's going to fail now, since `99/2` is not an integer.  This is not quite as exotic as it sounds, because grating order 2 will be at exactly twice the wavelength (and thus in some instruments, the channel) as grating order 1. However, I don't think it's a common case and one the user can easily fix with a clear error message (e.g. `99//2` will get you an integer and then at least you know exactly what happens).

DougBurke:
For the integer channel case - I am open to supporting non-integer channels as long as we have a clear statement of that should happen (if we were to do it I would argue it's int(value) so the fractional part is dropped). This would mean the tests changed here would not changing but more would need to be added.

However, I do quite like the "user must realize they are using channels and treat them appropriately" view!

I was going to add a point to say that either approach changes things to how we handle `Data1D` case - you can give it integer values for the independent axis but there's no special case behavior there, but I do think that `DataPHA` data is special since you can treat it both as `Data1D` or `Data1DInt` depending on your viewpoint.

hamogu:
Report from SDS meeting: The SDS/Sherpa team has discussed this and agreed that these changes should be done. This does not replace an explicit code review of *how* the changes are done, but we agree that the UI as written here, incl. changes to the current behavior, are matching SDS expectations.

DougBurke:
As #1208 has been merged I've rebased this to get the latest changes (as 1208 gained an extra typo in a comment fix) and marked as ready for review (so we can better check how these changes are covered by the tests).

hamogu (changes requested):
I have only on functional question, which may be solved by "is dealt with later". The test of the PR also has a "to think about" note which asks if we should put the same checks and limitation on filters for non PHA datasets (lo < hi). I agree we should do that, but since this is part of a series of PRs that concentrate on PHA data, which is somewhat different, I'll open a separate issue for that so that we can move forward here without further discussion.

DougBurke:
[There's two parts here

a) what is the test doing {the tests you highlight show values << data but it also holds for values >> data as I use below}
b) what should the behavior be for this case

and at some point we care about b but I want to explain a as well]

@hamogu - unfortunately I believe the tests that previously were "-20, -5" and are now "0, 1e-10" (or some other minimal value) hit the "special case" logic we have for notice regarding the first notice call. We have historically handled

    notice()  # ie ensure all data is available
    notice(0.5, 7.0)
    notice(5, 8)

as selecting the 0.5-7.0 range and so ignore < 0.5 and > 7.0, and then add the 5-8 range (and not ignore < 5, > 8) in the second call.

We therefore need to thin about what

    notice()
    notice(10000, 20000)

would do when the data covers << 10000. Do we want this call to ignore all the data (since technically 100000-20000 doesn't overlap the data) or do we treat it like we do for 

    notice()
    notice(0.5, 8)
    notice(10000, 20000)

where the 10000-20000 call wouldn't change anything.

So what we want is tests that check this behavior before I make changes and then after the change so we can see what has/hasn't changed. Previously we could use negative values as an "easy" check but we have to be more careful here (hence the "interesting" ranges like 0,1e-10).

So, the main question is what should the first `notice` call do; should it be special or not? If it's not special then this simplifies the code very-slightly, and simplifies the documentation, but would break user scripts and all our threads (at least that's my claim).

If we allow the first call of `notice` to be special, then how special do we make it, which is at some level what I think your question is about? If so then I have a whole section in https://github.com/sherpa/sherpa/pull/1216 discussing this - see the "energy filters outside the response" heading.

sherpa/astro/tests/test_astro_data.py
```diff
@@ -1524,10 +1524,10 @@ def test_notice_channel_grouping_outofbounds_ungrouped(lo, hi, expected, make_da
 @requires_data
 @requires_fits
 @pytest.mark.parametrize("lo,hi,expected",
-                         [(-5, 2000, '0.1248:12.4100'),
+                         [(0, 2000, '0.1248:12.4100'),
                           (0.7, 2000, '0.6716:12.4100'),
-                          (-5, 4.2, '0.1248:4.1391'),
-                          (-20, -5, '0.1248:12.4100'),
+                          (0, 4.2, '0.1248:4.1391'),
+                          (0, 1e-10, '0.1248:12.4100'),
```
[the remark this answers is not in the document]
DougBurke: See https://github.com/sherpa/sherpa/pull/1215#issuecomment-888428804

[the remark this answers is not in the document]
hamogu: I agree that it comes down to if and how "special" the first call is. I've never understood (and as a consequence always been uncomfortable with) that special first notice call (*). In fact, as a user (before I started reviewing PRs like this) I did not know that that first call was supposed to be special. I noticed that it did not work as I intended and that notice calls did not give repeatable results (i.e. if I repeat the first notice call, I get a different result - except that I never understood that it's only for the **first** call that's different, I just found "notice calls don't give repeatable results). To work around that, my sherpa scripts are littered with statement like `ignore(None, None); notice(None, None)` in an attempt to get to a repeatable state where the following notice/ignore calls give repeatable and predictable results.

So, I strongly dislike the "first notice call is special" design and I will continue to try to work around that in my own analysis. (I find it especially problematic in notebook, where cells can the executed in random order and, if I restart, I don't necessarily run all cells above). However, I understand that that's part of Sherpa's initial design, it's been around for a long time, and we don't want to break user scripts if we can avoid it.

So, I'm accepting that it's special, and at that point don't care how special it is - I'll try to avoid using it as much as I can. (Or how about a "First notice call is special" configuration item?  ;-)  )

(*) I'm sure it's explained somewhere in the docs or threads, but either I never read the right part (most users don't read all thread beginning to end, but pick one that looks close to what they want - I and use it that way, too); or it's a recent addition as want not sufficiently explained when I was a beginning sherpa user.

hamogu (approved):

sherpa/astro/tests/test_astro_data.py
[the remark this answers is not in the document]
DougBurke: I would argue that we haven't historically documented this well - it's just something you don't really think about too much.

Calling `ignore()` first is one way to get an explicit "clean" state for repeatable calls. But that's outside the scope of this particular PR!

DougBurke:
@wmclaugh - can we merge this (so I can get the next PR in this sequence reviewed) or are you waiting for something?

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -222,6 +224,114 @@ def test_error_on_invalid_channel_grouped2(chan):
...
+    pha.ignore_bad()
+    assert pha.get_filter() == ''
+    assert pha.get_xerr() == pytest.approx([1, 1, 1])
+
+
+def test_pha_get_xerr_all_bad_channel_group():
+    """get_xerr handles all bad values [channel]
+
+    The behavior with grouping is different, presumably because
+    we assume we have groping when we have a quality array.
```
wmclaugh, line 247:
```suggestion
    we assume we have grouping when we have a quality array.
```

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -222,6 +224,114 @@ def test_error_on_invalid_channel_grouped2(chan):
...
+    pha.ignore_bad()
+    assert pha.get_filter() == ''
+    assert pha.get_xerr() == pytest.approx([2.0, 3.0, 4.0])
+
+
+def test_pha_get_xerr_all_bad_energy_group():
+    """get_xerr handles all bad values [energy]
+
+    The behavior with grouping is different, presumably because
+    we assume we have groping when we have a quality array.
```
wmclaugh, line 289:
```suggestion
    we assume we have grouping when we have a quality array.
```

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -355,13 +543,29 @@ def test_416_b(caplog):
...
     pha = DataPHA('416', x, y)
-    pha.notice(3.5, 6.5)
+
+    rmf = create_delta_rmf(x, x + 1, e_min=x, e_max=x + 1,
+                           name='416')
+    pha.set_arf(rmf)
+    pha.set_analysis('energy')
+
+    # When using channels this used notice(3.5, 6.5)
+    # but using energy space we need to use a differnt
```
wmclaugh, line 564:
```suggestion
    # but using energy space we need to use a different
```

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -431,8 +454,20 @@ def test_can_use_swift_data(make_data_path):
     # this range is used. Alternatively, channel 801 could have been
     # excluded explicitly.
     #
+    # With changes to PHA filtering we now have to change the
+    # filter range again to include the 0.29-0.30 bin.
+    #
     # ui.notice(0.3, 8.0)
-    ui.notice(0.3, 7.995)
+    # ui.notice(0.3, 7.995)
+    # ui.notice(0.299, 8.0)
```
wmclaugh, line 462: Do we need this extra commented line?

DougBurke: These were added as I struggled to select the same range of channels as the code changed (it turns out we do things differently to XSPEC ...) and these comments are the scars to prove it. I can remove them.

DougBurke: I've cleaned up that comment (more than just removing the unnecessary lines) and noted that we didn't actually need to change the filter command (we may in one of the later PRs - I spent a  long time moving commits around to get a meaningful set of PRs and maybe that change sneaked through, or maybe it was due to some work that I then changed - so that will be interesting to check in those PRs, but not relevant here).

Issue #1211, opened 2021-07-14
Are filters semi-closed?


DougBurke:
This is using CIAO but it holds with the main branch (it has sprung to my attention from working on 
#1129):

```python
sherpa In [111]: edges = np.arange(0.4, 2.6, 0.2)

sherpa In [112]: d = Data1DInt('example', edges[:-1], edges[1:], np.ones(edges.size - 1))

sherpa In [113]: d.xlo
Out[113]: array([0.4, 0.6, 0.8, 1. , 1.2, 1.4, 1.6, 1.8, 2. , 2.2])

sherpa In [114]: d.xhi
Out[114]: array([0.6, 0.8, 1. , 1.2, 1.4, 1.6, 1.8, 2. , 2.2, 2.4])

sherpa In [115]: d.get_x(True)
Out[115]: array([0.5, 0.7, 0.9, 1.1, 1.3, 1.5, 1.7, 1.9, 2.1, 2.3])
```

So `get_x` returns the center of each bin (and the `True` argument says to use the current filter, but we do't have a filter at the moment. Let's add a filter - I've chosen this to match a single bin, namely 1.2-1.4:

```python
sherpa In [116]: d.notice(1.2, 1.4)

sherpa In [117]: d.get_x(True)
Out[117]: array([1.1, 1.3, 1.5])

sherpa In [118]: d.xlo[d.mask]
Out[118]: array([1. , 1.2, 1.4])

sherpa In [119]: d.xhi[d.mask]
Out[119]: array([1.2, 1.4, 1.6])
```

So we actually selected the bins

    1.0 - 1.2
    1.2 - 1.4
    1.4 - 1.6

Ideally I'd expect the `get_x` call at line 117 to just return `[1.3]`.

We also see something different for `DataPHA` objects (but it's not quite the same).

My current assumption is that we're treating `notice(lo, hi)` as `(lo, hi)` - i.e. inclusive of both ends - rather than `(lo, hi]` [I think that's the standard nomenclature]. Note that thanks to the fun of floating-point math it is always trickier than you'd expect, but we should be able to get what I want (but I wonder what will break in doing so).

dtnguyen2:
()-exclusive vs []-inclusive convention
<https://en.wikipedia.org/wiki/Interval_(mathematics)>

DougBurke:
So I was close ....

DougBurke:
[...]

PR #1216, opened 2021-07-20
filter improvements for DataPHA and Data1DInt

### LINKED ISSUE #926: PHA: notice/ignore with filters outside the valid data range
It looks like the filtering code for PHA fdata sets does not handle the case of "outrageous" ranges - that is, ones that do not overlap the range of the data. This example is with CIAO 4.12.1 but it's also true of the master branch and is being looked at as part of #920 and but I'm not sure I'll get to it as part of #922 

```
sherpa In [1]: import logging; logging.getLogger('sherpa').setLevel(logging.ERROR)

sherpa In [2]: load_pha('/home/dburke/sherpa/sherpa-master/sherpa-test-data/sherpatest/3c273.pi')

sherpa In [3]: logging.getLogger('sherpa').setLevel(logging.INFO)

sherpa In [4]: get_filter()
Out[4]: '0.124829999695:12.410000324249'

sherpa In [5]: notice(200, 300)

sherpa In [6]: get_filter()
Out[6]: '12.410000324249'
```

I've asked to notice 200-300 which is >> the energy range of the data. For a notice call this would normally not change the data except that this is the first notice call, so it should be treated as "all I want to use is the 200-300 keV range", which should mean that all data is ignored.

If I now ask to ignore the 200-300 keV range (after resetting things) we also see odd behavior, where I would have expected the noticed range to be unchanged, but you can see that the upper limit has actually changed.

```
sherpa In [7]: notice()

sherpa In [8]: get_filter()
Out[8]: '0.124829999695:12.410000324249'

sherpa In [9]: ignore(200, 300)

sherpa In [10]: get_filter()
Out[10]: '0.124829999695:8.219800233841'
```

It's essentially the same thing if we use values below the data range - e.g.

```
sherpa In [13]: get_filter()
Out[13]: '0.124829999695:12.410000324249'

sherpa In [14]: ignore(-300, -200)

sherpa In [15]: get_filter()
Out[15]: '0.277400001884:12.410000324249'
``` 

The problem comes from the fact that the code basically clips the ranges to the data range, so ignore(200,300) -> ignore(max energy, max energy) and this then causes the last bin to be dropped. Similarly for notice(200,300) -> notice(max energy,max energy) and then this causes the last bin to be included. It is surprisingly tricky to address sensibly....

DougBurke:
# Summary

Energy and wavelength filters (with notice and ignore) for PHA data are now treated as lo <= x < hi in all cases, and channel filters are lo <= x <= hi (where the channel values must be integers). For Data1DInt cases the notice and  ignore filters are treated as lo <= x < hi. This changes address a number of corner cases.

This can result in slightly different fit results because the data used in the fit can be changed and hence changing the number of degrees of freedom (by one or two depending on whether the first and last bins have been removed).

A number of private routines in the DataPHA class have been removed (_channel_to_group, _energy_to_channel, _wavelength_to_channel, and _to_channel).

# Details

This is related to issue #1211 .

There are also changes made to address #926 which definitely improve things and makes things more consistent with `Data1D` (see the `energy filters outside the response` example below).

This PR is the third in a sequence: #1208, #1215, #1216, #1219 (hopefully this is all).

# Examples

## Data1DInt example

Before this PR we get

```python
>>> d = Data1DInt('x', [1, 2, 4], [2, 3, 5], [1, 1, 1])
/home/dburke/anaconda/envs/ciao414b1/lib/python3.8/site-packages/sherpa/data.py:50: UserWarning: Converting array [1, 1, 1] to numpy array.
>>> d.mask
True
>>> d.notice(1, 2)
>>> d.mask
array([ True,  True, False])
```

With this PR we get

```python
>>> d = Data1DInt('x', [1, 2, 4], [2, 3, 5], [1, 1, 1])
/home/dburke/sherpa/sherpa-master/sherpa/data.py:88: UserWarning: Converting array [1, 1, 1] to numpy array.
>>> d.mask
True
>>> d.notice(1, 2)
>>> d.mask
array([ True, False, False])
```

So we can see that `notice(1, 2)` is now selecting just the 1-2 bin, not 1-2 and 2-3.

## PHA1 example (fake data)

Here we fake up a response so we have energy bins 0.1, 0.2, ...

```python
>>> import numpy as np
>>> from sherpa.astro.instrument import create_delta_rmf
>>> pha = DataPHA('x', np.asarray([1, 2, 3, 4, 5]), [1, 1, 1, 1, 1])
>>> egrid = np.asarray([0.1, 0.2, 0.3, 0.4, 0.5, 0.6])
>>> elo = egrid[:-1]
>>> ehi = egrid[1:]
>>> rmf = create_delta_rmf(elo, ehi, e_min=elo, e_max=ehi)
>>> pha.set_rmf(rmf)
>>> pha.units = 'energy'
```

Prior to this PR we got

```python
>>> pha.notice()
>>> pha.notice(0.1, 0.2)
>>> pha.mask
array([ True,  True, False, False, False])
>>> pha.notice()
>>> pha.notice(0.5, 0.6)
>>> pha.mask
array([False, False, False, False,  True])
>>> pha.notice()
>>> pha.notice(0.6, 0.7)
>>> pha.mask
array([False, False, False, False,  True])
>>> pha.notice()
>>> pha.notice(0.3, 0.4)
>>> pha.mask
array([False, False,  True,  True, False])
>>> pha.notice()
>>> pha.notice(0, 0.05)
>>> pha.mask
array([ True, False, False, False, False])
```

and now we get

```python
>>> pha.notice()
>>> pha.notice(0.1, 0.2)
>>> pha.mask
array([ True, False, False, False, False])
>>> pha.notice()
>>> pha.notice(0.5, 0.6)
>>> pha.mask
array([False, False, False, False,  True])
>>> pha.notice()
>>> pha.notice(0.6, 0.7)
>>> pha.mask
array([False, False, False, False, False])     # in an earlier version this was True
>>> pha.notice()
>>> pha.notice(0.3, 0.4)
>>> pha.mask
array([False, False,  True, False, False])
>>> pha.notice()
>>> pha.notice(0, 0.05)
>>> pha.mask
array([False, False, False, False, False])    # in an earlier version this was True
```

Note that I am using the `mask` attribute to find the bins that were selected by the `notice` call. The bins are 0.1-0.2, 0.2-0.3, 0.3-0.4, 0.4-0.5, and 0.5-0.6

| requested | before PR | after PR |
| ------------- | -------------- | ----------- |
| 0.1 - 0.2 | T T F F F | T F F F F |
| 0.5 - 0.6 | F F F F T | F F F F T |
| 0.6 - 0.7 | F F F F T | all F |
| 0.3 - 0.4 | F F T T F | F F T F F |
| 0 - 0.005 | T F F F F | all F |

We can see improvements - e.g.

- `notice(0.3, 0.4)` now only selects the 0.3-0.4 bin and not 0.3-0.5 (and similar for `notice(0.1, 0.2)`
- filters that have hi < 0.1 or lo >= 0.6 now remove all the bins rather than all but the first or last bin (this matches how `Data1D` works and fixes #926

Not shown here but I have examples where a filter to select a single bin can end up selecting the bin before, the bin, and the bin after before the PR that now select a single bin.

## PHA2 example (real data)

With

```python
>>> load_pha('3c120_pha2.gz')
>>> load_arf(10, '3c120_meg_1.arf.gz')
>>> load_rmf(10, '3c120_meg_1.rmf.gz')
>>> group_width(10, 2)
>>> set_analysis('wave')
>>> notice(21.4, 21.7)
```

prior to this PR we got

```python
>>> get_data_plot(10).xlo
array([21.71, 21.7 , 21.69, 21.68, 21.67, 21.66, 21.65, 21.64, 21.63,
       21.62, 21.61, 21.6 , 21.59, 21.58, 21.57, 21.56, 21.55, 21.54,
       21.53, 21.52, 21.51, 21.5 , 21.49, 21.48, 21.47, 21.46, 21.45,
       21.44, 21.43, 21.42, 21.41, 21.4 ])

>>> get_data_plot(10).xhi
array([21.7 , 21.69, 21.68, 21.67, 21.66, 21.65, 21.64, 21.63, 21.62,
       21.61, 21.6 , 21.59, 21.58, 21.57, 21.56, 21.55, 21.54, 21.53,
       21.52, 21.51, 21.5 , 21.49, 21.48, 21.47, 21.46, 21.45, 21.44,
       21.43, 21.42, 21.41, 21.4 , 21.39])
```

Note that the first and last bins are (after re-arranging the limits)

    21.39-21.40  and  21.70-21.7

and neither of these are within the requested noticed range of 21.4-21.7 if we think a filter is 21.4 <= x < 21.7. 

With this PR we get

```python
>>> get_data_plot(10).xlo
array([21.7 , 21.69, 21.68, 21.67, 21.66, 21.65, 21.64, 21.63, 21.62,
       21.61, 21.6 , 21.59, 21.58, 21.57, 21.56, 21.55, 21.54, 21.53,
       21.52, 21.51, 21.5 , 21.49, 21.48, 21.47, 21.46, 21.45, 21.44,
       21.43, 21.42, 21.41])
>>> get_data_plot(10).xhi
array([21.69, 21.68, 21.67, 21.66, 21.65, 21.64, 21.63, 21.62, 21.61,
       21.6 , 21.59, 21.58, 21.57, 21.56, 21.55, 21.54, 21.53, 21.52,
       21.51, 21.5 , 21.49, 21.48, 21.47, 21.46, 21.45, 21.44, 21.43,
       21.42, 21.41, 21.4 ])
```

that is the first and last bins are 21.40-21.41 and 21.69-21.70

## energy filters outside the response

With the "standard" PHA file

```python
>>> from sherpa.astro.io import read_pha
>>> pha = read_pha('3c273.pi')
>>> pha.mask
```

The energy range of the response is ~0.0015 to 14.95 keV:

```python
>>> pha._get_ebins(group=False)[0][0]
0.0014600000577047467
>>> pha._get_ebins(group=False)[1][-1]
14.950400352478027
```

Picking filters outside this range we have, before this PR sequence

```python
>>> pha.notice(1e5,1e6)
>>> pha.mask
array([False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
        True])
>>> pha.notice()
>>> pha.notice(1e-5, 1e-4)
>>> pha.mask
array([ True, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False])
```

You can see that the last or first bin is selected, which is wrong. We now get with this PR

```python
>>> pha.notice(1e5, 1e6)
>>> pha.mask
array([False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False])
>>> pha.notice()
>>> pha.notice(1e-5, 1e-4)
>>> pha.mask
array([False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False])
```

So now we no longer always select the first or last bin, but ignore all bins. This matches the `Data1D` behavior, where the first `notice` call, if outside the valid range of the independent axis, ignores all points:

```python
>>> d = Data1D('ex', [1, 2, 3], [1, 2, 3])
/home/dburke/anaconda/envs/ciao414b1/lib/python3.8/site-packages/sherpa/data.py:50: UserWarning: Converting array [1, 2, 3] to numpy array.
>>> d.mask
True
>>> d.notice(20, 30)
>>> d.mask
array([False, False, False])
```


hamogu:
This is difficult; but fortunately it does not matter that much in practice. If you analysis depends on including or excluding a single bin, you've either made your bins too wide or you don't have enough data to confidently say anything anyway. In other words, if users complain that they can't write their paper now, they should not have written this paper in the first place.

That said, from a documentation point, I'd much like all filters to be [] or all [). How am I supposed to remember that the same dataset includes the endpoint if I talk about channels, but not if I talk about energy?

I'm less worried about edges cases, like a bin that starts at 0.200000234 instead of 0.2000000 and thus is or is not included. We use the numbers we load and it's not our fault if they are not what the user things they should be.

DougBurke:
For a continuous range (like energy or wavelength) I don't think you can sensibly define the ranges with [lo, hi] semantics.

If we say notice(20.1, 20.4) means 20.1 <= x <=20.4 and your bins are 19.9-20.0,20.0-20.1, 20.1-20.2, 20.2-20.3,20.3-20.4,20.4-20.5,20.5-20.6 [either energy or wavelength] then do we select the 20.4-20.5 bin or not? We should because 20.4<= 20.4 (for the lower limit on the 20.4-20.5 bin check). Oh, we should also include the 20.0-20.1 bin as well as 20.1 <= 20.1 (from the upper limit on the 20.0-20.1 bin check). But then when I come [in #1219] to come up with my filter expression to explain this filter I would choose '20.0:20.5" (as they are the start and end point of the selected bins). But then if you call notice(20.0, 20.5) on this dataset you'd actually end up now selecting the 19.9-20.0 and 20.5-20.6 bins as well, changing the number of selected bins.

hamogu:
Report from SDS meeting: The SDS/Sherpa team has discussed this and agreed that these changes should be done. This does not replace an explicit code review of *how* the changes are done, but we agree that the UI as written here, incl. changes to the current behavior, are matching SDS expectations.

DougBurke:
[my plan is to leave this as draft until we merge the previous PR in the chain at which point I can rebase it to make it easier to see what was changed]

DougBurke:
I've rebased this now that #1215 has been merged.

DougBurke:
@hamogu - I've marked this as ready for review now that #1215 has been merged. This is where it starts getting fun...

One thing to note is that this specifically adds "extra" special cases to filter and ignore calls when the range is outside the data range - i.e. it's part of what you were asking about at https://github.com/sherpa/sherpa/pull/1215#issuecomment-888428804 and is definitely something we can change/review/think about.

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -411,8 +411,8 @@ def test_can_use_swift_data(make_data_path, clean_astro_ui):
     # bin.
     #
     # Unfortunately, using a range of 0.3-8.0 gives 771 bins
-    # in XSPEC - channels 30 to 800 - but 772 bins in Sherpa,
-    # channels 30 to 801.
+    # in XSPEC - channels 30 to 800 - but a 770 bins in Sherpa,
```
wmclaugh, line 414:
```suggestion
    # in XSPEC - channels 30 to 800 - but 770 bins in Sherpa,
```

sherpa/data.py
```diff
@@ -487,14 +492,45 @@ def apply(self, array):
...
+           otherwise points < max are included.
+        axislist: sequence of arrays
+           The axis to apply the range to. There must be the same
+           number of elements in mins, maxes, and axislist.  The
+           number of elements of each element of axislist must also
+           agree (the cell values do not need to match).
+        ignore : bool, optional
+           If True the range is to be ignored, otherwise it is
+           included.  The default is to include the range.
+
```
wmclaugh, line 517: Since the other parameters are described here, integrated should also be

sherpa/astro/tests/test_astro_data_swift_unit.py
DougBurke: I'm going to hold off fixing this until I have a bunch of changes.

sherpa/data.py
DougBurke: Why did I not add this argument to the docs? I'm not particularly happy with the need for an `integrated` flag here so if we can come up with a different solution that would be great!

sherpa/data.py
```diff
@@ -487,14 +492,45 @@ def apply(self, array):
...
+        See Also
+        --------
+        apply
+
+        """
+
+        # If integrated is True then we should have an even number
+        # of axisist elements, but we do not require this.
```
DougBurke, line 525:
```suggestion
        # of axislist elements, but we do not require this.
```

so @wmclaugh doesn't have to point it out ;-)

sherpa/data.py
```diff
@@ -487,14 +492,45 @@ def apply(self, array):
...
+        Parameters
+        ----------
+        mins : sequence of values
+           The minimum value of the valid range (elements may be
+           None).  When not None, it is treated as an inclusive limit,
+           so points >= min are included.
+        maxes : sequence of values
+           The maximum value of the valid range (elements may be
+           None).  When not None, it is treated as an inclusive limit,
+           so points <= max are included, when upper is True,
```
wmclaugh, line 507: Where is upper set/accessed?

sherpa/utils/__init__.py
```diff
@@ -1022,36 +1029,60 @@ def filter_bins(mins, maxes, axislist):
...
     Repeat the above calculation by combining filters for x >= 1.5
-    and x <= 2.3.
+    and x <= 4 (note that the grid must be repeated for each
+    filter):
+
+    >>> filter_bins([1.5, None], [None, 4], [xs, xs])
+    array([False,  True,  True,  True, False])
+
+    For integrated data sets the lower and upper edges should be sent
+    separately with the max and min limits, alng with setting the
```
wmclaugh, line 1046:
```suggestion
    separately with the max and min limits, along with setting the
```

sherpa/data.py
DougBurke: Just to note that I didn't design this code - I'm just trying to document it. 

Saying that,  I don't understand the question.

wmclaugh: Reading the test 'when upper is True' sounds like upper is a parameter to the function (which it doesn't appear to be) so I was asking for clarification of what determines when 'points <= max are included'.

DougBurke: Aha. This is botched documentation on my part - I think early on I had used a parameter called `upper` but I've changed this - so now I think it is likely related to `integrated` flag but I'll have to review things as the semantics of the two parameters are probably different.

anetasie:
@DougBurke I was just playing with code from this PR. If I have more than 1 data set and set the filter outside the data range the filter does not seem to be applied to all the datasets, only to the first one. 

```
from sherpa.astro.ui import *
%matplotlib

load_data(1,'P0784610301PNS009SRSPEC0002.FTZ')
load_pha(2,'P0784610301M1S007SRSPEC0002.FTZ')
load_pha(3,'P0784610301M2S008SRSPEC0002.FTZ')

load_bkg(1, 'P0784610301PNS009BGSPEC0002.FTZ')
load_bkg(2, 'P0784610301M1S007BGSPEC0002.FTZ')
load_bkg(3, 'P0784610301M2S008BGSPEC0002.FTZ')

notice(20.,40.) 
```

With this branch I get

```
In [4]: print(get_filter(1))
14.967499732971

In [5]: print(get_filter(2))
0.002499999944:9.714999914169

In [6]: print(get_filter(3))
0.002499999944:9.670000076294

```

CIAO 4.12 and 4.14 shows one bin - the last valid bin in these data, so still not an empty set.

```
sherpa In [2]: notice(20.,40.)

sherpa In [3]: print(get_filter(1))
14.967499732971

sherpa In [4]: print(get_filter(2))
9.714999914169

sherpa In [5]: print(get_filter(3))
9.670000076294
```

anetasie:
For some reason I'm not able to upload the tar/zip file with the data.

DougBurke:
@anetasie - can you report the output of

```python
load_data(1,'P0784610301PNS009SRSPEC0002.FTZ')
load_pha(2,'P0784610301M1S007SRSPEC0002.FTZ')
print(get_filter(1))
print(get_filter(2))

notice(20.,40.) 
print(get_filter(1))
print(get_filter(2))

r1 = get_data(1)._get_ebins()
r2 = get_data(2)._get_ebins()
print(r1[0][0], r1[1][-1])
print(r2[0][0], r2[1][-1])
```

The behavior is covered in the "energy filters outside the response" section above, but it is a bit abstuse without real-world data like this. I claim that the old behavior here is wrong but it is not clear what we want the new behavior to be (try to be clever or try to be predicatable). 

The above output would be useful to see from #1219 since then the `get_filter` output actually represents the whole energy range rather than the current-to-this-behavior which  makes it harder to see (and why I Include the _get_ebins output as well above, whic display the full range of the two grids).

anetasie:
Here is the output. I guess the data 1 overlaps with the filter and data 2 does not. It would be good to issue an error msgs/warning about filter being outside the valid data range. 

```
In [2]: load_data(1,'P0784610301PNS009SRSPEC0002.FTZ')
   ...: load_pha(2,'P0784610301M1S007SRSPEC0002.FTZ')
   ...: print(get_filter(1))
   ...: print(get_filter(2))
......

0.002684994834:14.967499732971
0.002499999944:9.714999914169

In [3]: notice(20.,40.)
   ...: print(get_filter(1))
   ...: print(get_filter(2))
14.967499732971
0.002499999944:9.714999914169

In [4]: r1 = get_data(1)._get_ebins()
   ...: r2 = get_data(2)._get_ebins()
   ...: print(r1[0][0], r1[1][-1])
   ...: print(r2[0][0], r2[1][-1])
0.0 20.479999542236328
0.0 12.0
```

DougBurke:
I strongly disagree with error-ing out here - imagine having three datasets that cover a range of x values, not all of them overlaping (e.g. multu-wavelength data): you wouldn't want it to fail because one dataset didn't cover the range you gave.

Similarly, I dislike the idea of printing warning messages for this.

hamogu:
For the "energy filters outside the response", I think we should be predictable and not clever. I'm not sure that's consistent with what I said before, but I'm getting around to that.
In my mind, I'm now thinking of the first (and thus special) notice call like this `notice(a, b)` is really `ignore(None, None); notice(a,b)`. Thus, if the first notice call is outside the PHA range, it should really not select anything. However, if the later notice calls are outside the range, then they are simply no-ops. Is that a reasonable way to think about it?

DougBurke:
So, my understanding of the code for the "first" notice call matches @hamogu (note that you can get back to this state with notice()), so, with the current main branch (i.e. not with this PR) we have the following (I'm using Data1D rather than DataPHA to show it's a generic property)

```python
>>> from sherpa.ui import *
>>> load_arrays(1, [1, 2, 4, 8, 9, 10], [1, 2, 3, 4, 5, 6])
>>> get_filter()
'1.0000:10.0000'
>>> notice(3, 8.5); get_filter()
'4.0000:8.0000'
>>> ignore(); notice(3, 8.5); get_filter()
'4.0000:8.0000'
>>> notice(); notice(3, 8.5); get_filter()
'4.0000:8.0000'
```

What happens if the value is outside the data range?

```python
>>> notice(); notice(20, 30); get_filter()
''
>>> >>> get_dep(filter=True)
array([], dtype=int64)
```

So, this has actually caused all points to be ignored. Which makes sense, if you can convert the `notice` call to `ignore(); notice(20,30)` and then remember that because there's no data > 10 then we filter out everything. This is potentially confusing since if make this notice call after filtering out data then it doesn't change anything:

```python
>>> notice(); notice(3, 8.5); notice(20, 30); get_filter()
'4.0000:8.0000'
```

Note that this PR does not change this behavior for Data1D. I'm going to add another comment to show what happens with Data1DInt.

DougBurke:
Actually, the behavior is the same for Data1DInt - a first call to notice where the range is outside the limits is not changed,

```python
>>> from sherpa.ui import *
>>> load_arrays(1, [1, 2, 4, 8, 9, 10], [2, 4, 8, 9, 10, 11], [1, 2, 3, 4, 5, 6], Data1DInt)
>>> get_filter()
'1.5000:10.5000'
>>> notice(3, 8.5); get_filter()
'3.0000:8.5000'
>>> ignore(); notice(3, 8.5); get_filter()
'3.0000:8.5000'
>>> notice(); notice(3, 8.5); get_filter()
'3.0000:8.5000'
>>> notice(); notice(20, 30); get_filter()
''
```

hamogu:
I agree that it's confusing and I don't like it - I never did. The fact that it's not even just the just notice call, but that you can get back to that special state makes it even more confusing. I did not understand that special behavior until @DougBurke opened this series of PRs and explained it to me - and I've used Sherpa for years (I always assumed it's buggy or I'm stupid because my notice/ignore calls gave different results depending on what I did before).

I know that @DougBurke and @anetasie are strong opposed to changing this "special" behavior because it's in so many threads (and probably user scripts), and by now hardwired into their brains. I understand that, but I also don't see a way to make this work logically within those constraints. I really don't know what to suggest here...

DougBurke:
The DataPHA behavior is different but is not changed by this PR (which I don't understand, see end of comment):

```python
>>> import numpy as np
>>> load_arrays(1, [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], np.ones(10), DataPHA)
>>> from sherpa.astro.instrument import create_arf
>>> erange = 0.1 * np.arange(1, 12) + 0.5; elo = erange[:-1]; ehi = erange[1:]
>>> elo
array([0.6, 0.7, 0.8, 0.9, 1. , 1.1, 1.2, 1.3, 1.4, 1.5])
>>> ehi
array([0.7, 0.8, 0.9, 1. , 1.1, 1.2, 1.3, 1.4, 1.5, 1.6])
>>> arf = create_arf(elo, ehi)
>>> set_arf(arf)
>>> get_analysis()
'energy'
>>> get_filter()
'0.650000000000:1.550000000000'
>>> notice(0.94, 1.28); get_filter()
'0.950000000000:1.250000000000'
>>> ignore(); notice(0.94, 1.28); get_filter()
'0.950000000000:1.250000000000'
>>> notice(); notice(0.94, 1.28); get_filter()
'0.950000000000:1.250000000000'
>>> notice(); notice(20, 30); get_filter()
'0.650000000000:1.550000000000'
```

The above was done with the main branch but it's the same as with this PR.I need to go back and check things just to make sure I understand what's going on (as I've been away fro mthis for a while)

DougBurke:
Aha, so, if I go back to the CIAO 4.14 beta 1 release (so before this PR sequence was merged) then we see some differences for this `notice(); notice(outside range); get_filter()` check for Data1DInt and DataPHA cases (i.e. some of the changes have already been made to main) that we are talking about.

The script [check_filtering.txt](https://github.com/sherpa/sherpa/files/7040547/check_filtering.txt) can be used (called `.txt` so I can attatch it to this comment):

In CIAO 4.14 beta 1 we have

# Data1DInt

The `notice(); notice(20, 30)` call would error out with `IndexError: tuple index out of range` 

# DataPHA

The `notice(); notice(20, 30)` call would return a filter that selected just the last bin: so in this case `get_filter()` would return `1.550000000000`

DougBurke:
For @hamogu - given that we've had this special case then I don't see a way to remove it without breaking a lot of existing code. What we can do is try to make sure what we do do "makes some sort of sense", and I think we've identified something that doesn't quite make sense (but is not actually due to this PR but is part of this sequence).

anetasie:
@DougBurke @hamogu I'm used to the special case and if there is no easy way to change it then we will have it. More examples of this behaviour in our notice/ignore docs might help.

sherpa/data.py
```diff
@@ -487,14 +492,45 @@ def apply(self, array):
...
-    def notice(self, mins, maxes, axislist, ignore=False):
+    def notice(self, mins, maxes, axislist, ignore=False, integrated=False):
+        """Select a range.
+
+        Parameters
+        ----------
+        mins : sequence of values
+           The minimum value of the valid range (elements may be
+           None).  When not None, it is treated as an inclusive limit,
+           so points >= min are included.
```
hamogu, line 503: Should we explain why this is a sequence (presumably for multi-D data?)

sherpa/data.py
```diff
@@ -487,14 +492,45 @@ def apply(self, array):
...
+        maxes : sequence of values
+           The maximum value of the valid range (elements may be
+           None).  When not None, it is treated as an inclusive limit,
+           so points <= max are included, when upper is True,
+           otherwise points < max are included.
+        axislist: sequence of arrays
+           The axis to apply the range to. There must be the same
+           number of elements in mins, maxes, and axislist.  The
+           number of elements of each element of axislist must also
+           agree (the cell values do not need to match).
```
hamogu, line 513: I know @DougBurke is documenting existing code, and not responsible for the design of these parameters. Unfortunately, I don't know what they are supposed to be and this explanation does not make me much wiser. I have the same number of elements in min and axislist, and the number of elements of each element also has to agree but the values do not. I think an example would help greatly to clearify. Something like

```
notice([1,4], [2,7], [[2,3], [0,2]])
```
would then apply the limits 1..2 to axis 2 and 3 and the limits 4..7 to axis 0 and 2? But is that interpretation is correct and a sequence of values is simple like calling notive several times on each element in that sequence, then why does the number of elements in axis list be the same for each element? Why not apply one range to axis 0 and another one to axis 2 and 3? That makes no sense, so apparently I don't understand either the format or the intention or both.

sherpa/utils/__init__.py
```diff
@@ -993,24 +993,31 @@ def pad_bounding_box(kernel, mask):
...
 
+    The ranges are treated as inclusive at both ends if integrated is
+    False, the default, otherwise the lower limit is inclusive but the
+    upper limit is exclusive.
+
     Parameters
     ----------
     mins : sequence of values
        The minimum value of the valid range (elements may be None).
-       When not None, it is treated as an inclusive limit, so
```
hamogu, line 1003: Like above, I don't understand what that can be a sequence of values. Apparently, I miss something fundamental.

hamogu (changes requested):
I seem to fundamentally misunderstand how the arguments for notice should look like. I left comments on the docstrings where the docstring could be improved to help me. Hopefully that will help be to understand the code better, which in turn will allow me to review it better.

sherpa/data.py
DougBurke: So, as far as I understand it, this interface is unlikely to be used by anyone, as the classes that extend the `Data` class (I think that's where this is defined) have their own `notice` call, which has a different interface.

sherpa/data.py
DougBurke: As I said, I don't think this is really user actionable, it's more for any poor soul that wants to write their own data class.

sherpa/data.py
DougBurke: Ahh, this is from the `Filter` class, which the `Data` class calls with

```python
    def notice(self, mins, maxes, ignore=False, integrated=False):
        self._data_space.filter.notice(mins, maxes, self.get_indep(),
                                       ignore=ignore, integrated=integrated)
```

and, for example, the `Data1D.notice` method is

```python
   def notice(self, xlo=None, xhi=None, ignore=False):
... docstring ...

        Data.notice(self, (xlo,), (xhi,), ignore)
```

I didn't write this code, I am just trying to document it. I would say that this code is not directly used by the user unless they really go digging through the code, so I am reluctant to spend more time documenting this at this time.

DougBurke: I think some of the potential confusion is that this code has to deal with `Data1D` (one axis), `Data1DInt` (two axes, if the low and high values), `Data2D` classes (two axes, but independent), and `Data2DInt` classes [*] (which have two pairs of low/hi axes). This is all handled by a single routine but it does **seem to** require some knowledge by the calling classes (e.g. what the low and high edges are and differences between the nD and nDInt cases). This is all based on me trying to reverse-engineer the code so I may be missing something.

One solution is to just remove this documentation.

[*] we support at the low level but don't really have experience with)

sherpa/astro/tests/test_astro_data_swift_unit.py
DougBurke: Addressed in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf

sherpa/data.py
DougBurke: Addressed in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf

sherpa/data.py
DougBurke: Addressed in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf

sherpa/data.py
DougBurke: Addressed in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf

sherpa/utils/__init__.py
DougBurke: Addressed in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf

sherpa/data.py
DougBurke: There have been some minor doc changes to this method in https://github.com/sherpa/sherpa/pull/1216/commits/73ca7cc38e045b2883205f5a11b497a90e8a49bf but I'm not sure they help

DougBurke:
I've rebased because I needed to address some minor conflicts in #1219 and address review comments from @wmclaugh. I also wanted to check that other recent changes we've merged didn't change any of the tests.

DougBurke:
And there were test failures because of those rebases so I've fixed them. This is in the handling of data which starts at channel  0 and they are currently regression tests rather than tests from first principles, so I've accepted these changes (parts of the two tests had commented-out tests that now pass, so it was mainly a case of switching back to them).

sherpa/utils/__init__.py
DougBurke: I think this is similar to https://github.com/sherpa/sherpa/pull/1216#discussion_r704488780

DougBurke:
So, I've finally got to removing the special case where the first `notice` call would ignore the filter it it did not overlap the data. This now means that DataPHA matches Data1D, I would say back to how it used to work but PHA filtering was strange here, and removes some complexity (although the code change was on;y to remove two lines). The tests that did change are only tests I added to catch this (and similar) cases, and didn't change any "real world" test cases we have.

So, for a Changdra dataset which is only defined up to ~15 keV we now have that the first time you say `notice(20, 30)` then all points are ignored

```python
>>> pha = sherpa.astro.io.read_pha('3c273.pi')
>>> pha.notice(20, 30)
>>> pha.get_filter()
''
```

As mentioned, this is what `Data1D` does:

```python
>>> d = Data1D('ex', [1, 2, 3], [1, 2, 3])
/home/dburke/sherpa/sherpa-master/sherpa/data.py:88: UserWarning: Converting array [1, 2, 3] to numpy array.
>>> d.notice(20, 30)
>>> d.get_filter()
''
```

sherpa/data.py
DougBurke: There's a few more doc changes in https://github.com/sherpa/sherpa/pull/1216/commits/6b6749b62b51eba4471d0234f20f817103649a1a

DougBurke:
So @hamogu 

- I've removed the extra special-case on-top-of a special case logic I had called out in my original PR summary as needs looking at - this is for the first `notice(lo, hi)` call when `lo/hi` do not overlap the data range of the PHA data set and we now ignore all points (which matches `Data1D`); the behavior before this PR is discussed in #926 and one of the examples in the PR summary
- this doesn't remove the original special-case behavior of the first notice call, but I claim this is **not** the PR to do this in (and I can not think of a way to actually do it, as you have also mentioned)
- I have added some documentation to the `sherpa.data.Filter.notice` call (not sure I added much to `sherpa.utils.filter_bins`) but I argue that this is **real low-level stuff** and I don't believe we need to significantly improve this documentation at this time; the only other solution I have is to just remove this bit of documentation which I don't think helps

As far as I'm aware this addresses the issues we've talked about but maybe I'm missing something.

sherpa/data.py
hamogu: It's a as good as it gets now and that's good enough. Perfect is the enemy of good.

hamogu (approved):

PR #1190, opened 2021-06-15
Restore minim Fortran

dtnguyen2:
# Summary

This PR restores the Fortran version of ```minim``` because the variability pipeline in L3 was not getting the same results as the CAT2.0 runs by using the C/C++ translated version . A runtime option, ```useminimC``` with a default value of ```True```, was introduced to allow the user to select which version of ```minim``` to use.

When the C/C++ translation to the ```minim``` code was done there was only 26 tests and the tests did not show any issues with the translation.

```
(test-minmin) [dtn@devel12 sherpa]$ git diff
diff --git a/sherpa/optmethods/optfcts.py b/sherpa/optmethods/optfcts.py
index 8abf7cc8..9eff10b1 100644
--- a/sherpa/optmethods/optfcts.py
+++ b/sherpa/optmethods/optfcts.py
@@ -643,7 +643,7 @@ def montecarlo(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, verbose=0,
 #
 def neldermead( fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None,
                 initsimplex=0, finalsimplex=9, step=None, iquad=1,
-                verbose=0, useminimC=True ):
+                verbose=0, useminimC=False ):
 
     x, xmin, xmax = _check_args(x0, xmin, xmax)
```

```
(test-minmin) [dtn@devel12 sherpa]$ python setup.py test --pytest-args=sherpa/optmethods/tests/test_optmethods.py 
...
sherpa/optmethods/tests/test_optmethods.py ..........................    [100%]

============================== 26 passed in 3.16s ==============================
```

However, sometime after the C/C++ translation was done (yes I know, should have added more tests before), more tests were added and when the Fortran version of ```minim``` was used to run against the tests it looks like one of the tests did not get the same results as the C/C++ version.  So obviously, there was a minor bug (if a lot of tests had failed then it would probably be easier to find the offending code).

```
============================= test session starts ==============================
platform linux -- Python 3.7.3, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
rootdir: /export/sherpa, inifile: pytest.ini
plugins: xvfb-2.0.0, arraydiff-0.3, xdist-1.29.0, forked-1.0.2, openfiles-0.3.2, doctestplus-0.3.0, remotedata-0.3.1
collected 233 items                                                            

sherpa/optmethods/tests/test_optmethods.py ....xxxx..................... [ 12%]
[...]

PR #1217, opened 2021-07-20
restore Fortran version of minim & fix the C/C++ version of minim to give identical results to the Fortran version

dtnguyen2:
# Note

This PR restores the Fortran version of minim and changes how the parameter boundary violations in the  C/C++ version of minim are done (from reflecting about the limits to stopping at the limits). 

It turns out the translation of the C/C++ version of minim was done correctly, after all it has been used for quite a while now; However, in the process of translating to the C/C++ version the violation of the parameter boundary limits were changed to reflection about the limits.  As an aside, there was a discussion about making the option of how ```neldermead``` handles the boundary conditions  available for future releases since the fitted results depend on how the boundary conditions are handled.

# User's guide: 
To use the Fortran version of minim, the user should set the ```neldermead``` option ```useminimC``` from the default value of ```True``` to ```False```.   The L3 variability pipeline should use this option if the Fortran version is to be used.

To use the C/C++ version of minim, the user has two options:

1. To get identical results as the Fortran version of minim, change the ```neldermead``` default value of  ```reflect``` from ```True``` to ```False```.  The L3 variability pipeline should use this option if the C/C++ version is to be used.
2. To use ```neldermead``` as the ciao412, then do not change the options ```useminimC``` nor ```reflect```.  The rest of the L3 scripts, ie non-variabilty pipelines, should use this option.

### DIFF
--- a/helpers/extensions/__init__.py
+++ b/helpers/extensions/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2014, 2016, 2017, 2018, 2020
+#  Copyright (C) 2014, 2016, 2017, 2018, 2020, 2021
 #       Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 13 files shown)


DougBurke:
Repeating my coments from https://github.com/sherpa/sherpa/pull/1190#issuecomment-863392267

A few (related) thoughts before I dig in deep but

1. Rather than have an extra flag for the `Neldermead` class perhaps we subclass the `Neldermead` class (e.g. `NeldermeadFortranIsASilyName`) so that you have "neldermead" and "neldermeadfortranisasillyname" as optimisers (from the UI layer) - much of the lower-level code could probably remain the same
2. can we make the fortran compilation optional so you don't need a fortran compiler (and then you just don't get the fortran version)?

I think 2 is the more important one since it makes it easier for our CI infrastructure. The first point is only relevant if we end up having to keep code in for a release (and it's not obvious to me just now what the best approach to take is).

dtnguyen2:
oops, pushed into the wrong branch. see PR #1221

PR #1218, opened 2021-07-21
Update fftw from version 3.3.8 to 3.3.9

wmclaugh:
**Summary**

This change updates the version of the fftw module in sherpa's external dependencies (extern) directory to utilize version 3.3.9 (www.fftw.org) instead of the previous version 3.3.8

**Notes**
Included with this fix are the reinsertion of autoreconf changes from that were removed in #1204. Note that the changes were removed when autoreconf was executed on a RedHat 7.9 machine instead of an Ubuntu 18 machine. Although both used version 2.69, the ubuntu version of the utility provides support for --runstatedir. That feature is introduced in version 2.70.


DougBurke (approved):
I guess I should have made a fuss about https://github.com/sherpa/sherpa/pull/1204#pullrequestreview-710717144 then!

Issue #531, opened 2018-10-23
Should we use the C++ rather than C style interface for XSPEC models with name C_xxx in the model.dat file


DougBurke:
Many XSPEC models are now of the "C_" prefix variety - that is they are "native" C++ models. See https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/node299.html#AppendixAddModels for a discussion of `name` vs `c_name` vs `C_name` for the routine names.

We currently use the C style interface for these models - ie from `sherpa/include/sherpa/astro/xspec_extension.hh` we have

```
template <npy_intp NumPars, bool HasNorm,
void (*XSpecFunc)( const double* energy, int nFlux,
		const double* params, int spectrumNumber,
		double* flux, double* fluxError,
		const char* initStr )>
PyObject* xspecmodelfct_C( PyObject* self, PyObject* args )
{
```

Should we add a C++ style interface and use that? The reason is the claim from the XSPEC doc that

> The C-style call can clearly be compiled and implemented by either a C or a C++ compiler: however we recommend using the C++ call if the model is written in C++ as it will reduce overhead in copying C arrays in and out the XSPEC internal data structures.

and presumably it would mean a template something like

```
template <npy_intp NumPars, bool HasNorm,
void (*XSpecFunc)( const RealArray& energy,
		const RealArray params, int spectrumNumber,
		RealArray& flux, RealArray& fluxError,
		const string& initStr )>
PyObject* xspecmodelfct_CXX( PyObject* self, PyObject* args )
{
```

although I'm sure there'd be some fun with the array type

DougBurke:
As of XSPEC 12.12.0 there is now also models that use the double-precision FORTRAN interface (e.g. `F_ismabs`) which would be nice to support.

PR #1223, opened 2021-07-25
XSPEC 12.12.0 support

DougBurke:
# Summary

Support the new additive and multiplicative models in XSPEC 12.12.0 - XSgrbjet, XSwdem, XSvwdem, XSvvwdem are additive and XSzxipab is multiplicative - and note the additional abundance tables that are available when running with XSPEC 12.12.0 (`lpgp` and `lpgs`).

# Details

The initial commit fixes up #1260 (the commit-log erroneously claims #1261) for changes from #1259 (as they were merged contemporaneously). This is technically not needed for XSPEC 12.12.0 but these changes were useful to validate the changes I had made (e.g. that the model parameters matched the `model.dat` file).

Commit "Initial XSPEC 12.12.0 support" makes the necessary changes so we can build against XSPEC 12.12.0 (the build has changed so that the location of the include files has moved [but this is a useful change on XSPECs part]).

Commit "Add the XSPEC 12.12.0 models (low level only)" adds the new models in this release to the compiled code, but they can't be used from Sherpa as the model classes haven't been added. These changes are, thanks to #1260, pretty mechanical with `scripts/add_xspec_model.py`.

Commit "Add the XSPEC 12.12.0 models (usable from Sherpa)" adds the model classes so these new models can be used. This is also a partly-mechanical process with `scripts/add_xspec_model.py`, but it needs more work to add necessary code to support conditional compilation and expanding the documentation. The `scripts/check_xspec_update.py` script has been used to check that the result matches the HEASARC 6.29 model.dat file.

The other changes are in the documentation, partly driven by #1268 to improve the linking in the readthedocs documentation (mechanical things like changing a double backtick to a single back tick or ' and fixing up some links).

# Notes

I have an ongoing support ticket with HEASARC as there's some odd behavior with the grbjet model. The tests skip this model because of this (the model can occasionally return all zeros). As it's a new model it won't break anyone's code if they can't use this model reliably (it is an XSPEC issue).

I also reported an issue with the log(xi) parameter name for the zxipab model and I expect it to be renamed to log_xi in a patch (as the other parameters named this are). We can not use log(xi) as the parameter name so we are using log_xi in anticipation of this change. Similarly there is a conflict with the p and P parameters of XSvvwdem so I have renamed "p" to "inv_slope" for all three *wdem models for consistency. I've also been in contact with Keith about this.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -360,6 +360,14 @@ void kyconv_(float* ear, int* ne, float* param, int* ifl, float* photar, float* 
 void thcompf_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
 #endif
 
+// XSPEC 12.12.0 changes
+#ifdef XSPEC_12_12_0
+// Note: have dropped the leading 'c_' for this model to follow xsgrbcomp
+void xsgrbjet(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
+
+void zxipab_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
+#endif
+
 }
 
 // This routine could be called when the module is being initialized,
@@ -368,7 +376,11 @@ void thcompf_(float* ear, int* ne, float* param, int* ifl, float* photar, float*
 // that they call _sherpa_init_xspec_library before calling any
 // XSPEC routine.
 //
-// Sun's C++ compiler complains if this is declared static
+// Sun's C++ compiler complains if this is declared static.
+//
+// TODO: should we expose this so that anyone who wants to use a user-model
+//       can access it?
+//
 int _sherpa_init_xspec_library()
 {
 
@@ -1409,6 +1421,15 @@ static PyMethodDef XSpecMethods[] = {
   XSPECMODELFCT(olivineabs, 2),
   XSPECMODELFCT_C(C_logconst, 1),
   XSPECMODELFCT_C(C_log10con, 1),
+#endif
+
+#ifdef XSPEC_12_12_0
+  XSPECMODELFCT_C_NORM( xsgrbjet, 14 ),  // follow xsgrbcomp and drop the leading c_
+  XSPECMODELFCT_C_NORM( C_vvwDem, 37 ),
+  XSPECMODELFCT_C_NORM( C_vwDem, 21 ),
+  XSPECMODELFCT_C_NORM( C_wDem, 8 ),
+
+  XSPECMODELFCT(zxipab, 5),
 #endif
 
   { NULL, NULL, 0, NULL }
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -100,16 +100,22 @@ int _sherpa_init_xspec_library();
 // gammap
 // gammq
 //
-#include "xsFortran.h"
-
-// TODO: is this defined in an XSPEC header file?
-#define ABUND_SIZE (30) // number of elements in Solar Abundance table
-
-// C_<model> are declared here; the other models are defined in
+// TODO:: switch to C++ FuntionUtility interface rather than use xsFortran.h
+//
+// funcWrappers: C_<model> are declared here; the other models are defined in
 // functionMap.h but that requires using the XSPEC build location
 // rather than install location.
 //
+#ifdef XSPEC_12_12_0
+#include "XSFunctions/Utilities/xsFortran.h"
+#include "XSFunctions/funcWrappers.h"
+#else
+#include "xsFortran.h"
 #include "funcWrappers.h"
+#endif
+
+// TODO: is this defined in an XSPEC header file?
+#define ABUND_SIZE (30) // number of elements in Solar Abundance table
 
 extern "C" {
 
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -256,7 +256,7 @@ Update the XSPEC bindings?
 --------------------------
 
 The :py:mod:`sherpa.astro.xspec` module currently supports
-:term:`XSPEC` versions 12.11.1 down to 12.9.0. It may build against
+:term:`XSPEC` versions 12.12.0 down to 12.9.0. It may build against
 newer versions, but if it does it will not provide access to any new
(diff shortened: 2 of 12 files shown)


DougBurke:
@olaurino - I think this is something we should look at but not for 4.14, so I'll keep it in my back pocket... The descriptor stuff looks a lot like the DataClass stuff I've read about (which I'm guessing is not an accident).

DougBurke:
Rebased to pick up OTS-required PRs so it will build.

DougBurke:
This has been rebased as #1261 and #1228 have been merged. I've also pulled out the setup changes to #1270 as it is completely unrelated to this PR.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -354,6 +360,14 @@ void kyconv_(float* ear, int* ne, float* param, int* ifl, float* photar, float*
 void thcompf_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
 #endif
 
+// XSPEC 12.12.0 changes
+#ifdef XSPEC_12_12_0
+// Note: have dropped the leading 'c_' for this model to follow xsgrbcomp
+void xsgrbjet(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr);
+
+void zxipab_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);
```
anetasie, line 368: It seems that 'xs' is missing from `zxipab` model name here

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -1405,6 +1423,15 @@ static PyMethodDef XSpecMethods[] = {
   XSPECMODELFCT_C(C_log10con, 1),
 #endif
 
+#ifdef XSPEC_12_12_0
+  XSPECMODELFCT_C_NORM( xsgrbjet, 14 ),  // follow xsgrbcomp and drop the leading c_
+  XSPECMODELFCT_C_NORM( C_vvwDem, 37 ),
+  XSPECMODELFCT_C_NORM( C_vwDem, 21 ),
+  XSPECMODELFCT_C_NORM( C_wDem, 8 ),
+
+  XSPECMODELFCT(zxipab, 5),
```
anetasie, line 1432: It seems that 'xs' is missing from zxipab model name also here

anetasie (commented):
`zxipab` model seem to miss 'xs' in the name in two places. Is this intended?

sherpa/astro/xspec/src/_xspec.cc
DougBurke: No it isn't. This is the correct name for this part of the code.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: No, this is the correct name for this part of the code.

DougBurke:
The file `sherpa/astro/xspec/src/_xspec.cc` uses the function name from the `model.dat` file. This is at a level "below" the names that @anetasie is worried about, where the names are taken from the `class XS...` lines in `sherpa/astro/xspec/__init__.py` and in these the `...` is the XSPEC model name (so in this case there's a `class XSzxipab(...)` line.

As I'm boring people, note that `zxipab` uses the same function and model name, but other models don't (e.g. `zxipcf` model uses the function `C_zxipcf` [*] and `zredden` uses the function `xszrcd`). 

```
% cat ~/local/heasoft-6.29/spectral/manager/model.dat
...

zxipab       5 0.01         1.e20            zxipab    mul 0 0
nHmin  10^22     0.01  1e-07   1e-07     1000      1e+06       1e-2
nHmax  10^22     10    1e-07   1e-07     1000      1e+06       1e-2
beta    " "      0.0   -10     -10       10        10          1e-2
log(xi) " "      3     -3      -3        6         6           1e-2
redshift " "     0.    0.      0.       10.       10.         -0.01

zxipcf       4 0.01         1.e20           C_zxipcf   mul 0
Nh     10^22     10    0.05    0.05      500       500         1e-1
log_xi " "      3     -3      -3        6         6           1e-2
CvrFract " "     0.5   0.      0.        1.        1.          1e-2
Redshift " "    0.   -0.999  -0.999   10.       10.       -0.01

zredden        2  3.72e-4    9.92e-3         xszcrd    mul  0
E_BmV   " "    0.05  0.      0.      10.       10.       0.001
Redshift " "    0.   -0.999  -0.999   10.       10.       -0.01

...
```

---

[*] actually, the function name in the `model.dat` file is not technically the function name but an optional prefix plus the function name, where the optional prefix selects if it's a C, C++, or double-precision FORTRAN function (`c_`, `C_`, `F_`). If there's no prefix it's a FORTRAN model. [**]

[**] The XSPEC model library actually provides wrappers for the models so we can call them from C, C++, or FORTRAN, but this is getting ridiculously-far from the original question... Issue #531 talks about this a bit, but this has only been added to keep this comment from ending.

sherpa/astro/xspec/src/_xspec.cc
anetasie: ok it is confusing as the model above has 'xs' included and it is called 'xsgrbjet'

DougBurke: Yes, that's because we have

```
% egrep 'grbjet|zxipab' ~/local/heasoft-6.29/spectral/manager/model.dat
grbjet         13   1.e-20     1.e+20         c_xsgrbjet   add  0
zxipab       5 0.01         1.e20            zxipab    mul 0 0
```

Note that above we have to say `zxipab_` but `xsgrbjet`, which I can explain but you shouldn't care about.

Welcome to my world...

anetasie (approved):

DougBurke:
I'm going to have to do a little work to merge this as there's changes needed to support the recenty-merged #1259 work

DougBurke:
I've rebased it for the #1259 changes but there's some more tweaks I want to add related to address non-obvious conflicts between #1259 and #1260

DougBurke:
I've rebased to add an initial commit that fixes up #1261 for changes from #1259 (as they were merged contemperanously). This is technically not needed for XSPEC 12.12.0 but these changes were useful to validate the changes I had made (e.g. that the model parameters matched the model.dat file).

DougBurke:
@Marie-Terrell - I'm done with rebasing this.

PR #1227, opened 2021-07-26
Support AstroPy 4.3 for tests

### LINKED ISSUE #1226: astropy 4.3 and it's FITS checks are causing us problems
I have a test that manually creates a FITS file for a test. astopy-4.3 now notes that it's not a "correct" file and throws an error

```
E               OSError: No SIMPLE card found, this file does not appear to be a valid FITS file
```

Of course, we are now getting version 4.3 on the pip installs in our CI run.

This should be an easy change to the test.

DougBurke:
# Summary

The AstroPy FITS reader is less tolerant of invalid FITS files in AstroPy 4.3 which trips one of our tests. The test has been updated with a FITS file that doesn't trigger the error case. There is no functional change in this commmit.

# Details

Fixes #1226 

The FITS checking meant that a test that created a test FITS file (which was dome rather than adding one to the sheropa-test-data directory) was no-longer valid for AstroPy, so tweak the output to appease both AstroPy 4.3 and fverify (at least for the header lines, it still claims the fill character is wrong but this is less important).

I've tested this out as part of #1223 but pulled it out into its own PR.

### DIFF
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -150,8 +150,12 @@ def fake_rmf(outfile):
     def hdr(key, value):
         if isinstance(value, str):
             value = "'{}'".format(value)
-        elif isinstance(value, bool):
-            value = 'T' if value else 'F'
+        else:
+            if isinstance(value, bool):
+                value = 'T' if value else 'F'
+
+            # add spacing to make FVERIFY happy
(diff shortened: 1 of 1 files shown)


hamogu (approved):
That's why most people use a library to write fits files and don't write them by hand... but it's good to know we can if we want to!

PR #1228, opened 2021-07-26
Fix three parameter names for XSPEC models

DougBurke:
# Summary

Fix three XSPEC models which had some internal confusion over parameter names: `XSzkerrbb` should have uses `fcol` but had instead used `hd`, and `XSzashift` and `XSzmshift` used `Redshift` instead of `Velocity`. The old names have been kept as aliases.

# Details

In working on #1223 I added some checks to ensure we don't create XSPEC parameters with a different name to the attribute setting, which showed three cases where we had done this. We now use the correct name and leave the old name as an alias so old code will still work.

To make sure this doesn't happen we have included assertions that will fail if someone tries to create a model incorrectly. These are not mapped to Python errors as this is very low-level code and I believe using an assert is enough. There is a discussion to be had about where some of the checks should be made - at the moment this is done in the base XSPEC model class as we have historically changed XSPEC model parameters, or added new XSPEC models, **much more** than other models, and moving this code to an earlier class would require more thinking about than I want to do here. We could remove the `Internal improvements ...` commit and just avoid this.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -7205,6 +7205,10 @@ class XSzkerrbb(XSAdditiveModel):
 
     The model is described at [1]_.
 
+    .. versionchanged:: 4.14.0
+       The fcol parameter was incorrectly labelled as hd: both names
+       can be used to access this parameter.
+
     Attributes
     ----------
     eta
@@ -7264,7 +7268,9 @@ class XSzkerrbb(XSAdditiveModel):
         self.Mbh = Parameter(name, 'Mbh', 1., 0., 100., 0.0, hugeval, 'M_sun')
         self.Mdd = Parameter(name, 'Mdd', 1., 0., 1000., 0.0, hugeval, 'M0yr')
         self.z = Parameter(name, 'z', 0.01, 0., 10., 0.0, 10, frozen=True)
-        self.fcol = Parameter(name, 'hd', 1.7, 1., 10., 0.0, hugeval, frozen=True)
+        self.fcol = Parameter(name, 'fcol', 1.7, 1., 10., 0.0, hugeval, frozen=True,
+                              # Parameter was mis-labelled until 4.14.0
+                              aliases=['hd'])
         self.rflag = Parameter(name, 'rflag', 1., -100., 100., -hugeval, hugeval, alwaysfrozen=True)
         self.lflag = Parameter(name, 'lflag', 0., -100., 100., -hugeval, hugeval, alwaysfrozen=True)
         self.norm = Parameter(name, 'norm', 1.0, 0.0, 1.0e24, 0.0, hugeval)
@@ -12886,6 +12892,10 @@ class XSvashift(XSConvolutionKernel):
 
     The model is described at [1]_.
 
+    .. versionchanged:: 4.14.0
+       The Velocity parameter was incorrectly labelled as Redshift:
+       both names can be used to access this parameter.
+
     .. versionadded:: 4.12.2
 
     Attributes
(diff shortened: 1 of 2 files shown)


DougBurke:
@anetasie - I've asked for your review for this more on the overview rather than code changes. I'll bother @hamogu once you've signed off. Basically it's that until now we had xszkerrbb had a parameter called hd but it should have been fcol andthe xsvashift/xsvmshift models used Redshift rather than Velocity as the parameter name (they did this slightly differently so the behavior is not quite the same). After this commit we 

a) use the correct name so "print(mdl)" will return the correct name
b) allow the old name to be used as well as the new name

The following are all done with CIAO 4.14 beta 1 so are before this PR

# xszkerrbb

```python
sherpa In [1]: xszkerrbb.mdl
-------------> xszkerrbb.mdl()
Out[1]: <XSzkerrbb model instance 'xszkerrbb.mdl'>

sherpa In [2]: print(mdl)
xszkerrbb.mdl
   mdl.eta      frozen            0            0            1
   mdl.a        thawed            0           -1       0.9999
   mdl.i        frozen           30            0           85        deg
   mdl.Mbh      thawed            1            0          100      M_sun
   mdl.Mdd      thawed            1            0         1000       M0yr
   mdl.z        frozen         0.01            0           10
   mdl.hd       frozen          1.7            1           10
   mdl.rflag    frozen            1         -100          100
   mdl.lflag    frozen            0         -100          100

sherpa In [3]: mdl.hd
Out[3]: <Parameter 'hd' of model 'mdl'>

sherpa In [4]: mdl.fcol
Out[4]: <Parameter 'hd' of model 'mdl'>

sherpa In [5]: mdl.fcol = 3

sherpa In [6]: mdl.fcol
Out[6]: <Parameter 'hd' of model 'mdl'>

sherpa In [7]: print(mdl)
xszkerrbb.mdl
   mdl.eta      frozen            0            0            1
   mdl.a        thawed            0           -1       0.9999
   mdl.i        frozen           30            0           85        deg
   mdl.Mbh      thawed            1            0          100      M_sun
   mdl.Mdd      thawed            1            0         1000       M0yr
   mdl.z        frozen         0.01            0           10
   mdl.hd       frozen            3            1           10
   mdl.rflag    frozen            1         -100          100
   mdl.lflag    frozen            0         -100          100
```

So in this case we could use `hd` or `fcol` as the parameter name.

# xszashift/zmshift

Here I use create_model_component as these are convolution models so creating them in a stand-alone call (rather than in a set_source call) is a bit more awkward

```python
sherpa In [21]: clean()

sherpa In [22]: create_model_component('xszashift', 'shifty')
Out[22]: <XSzashift kernel instance 'xszashift.shifty'>

sherpa In [23]: print(shifty)
xszashift.shifty
   shifty.Redshift frozen            0       -0.999           10

sherpa In [24]: shifty.redshift = 3

sherpa In [25]: print(shifty)
xszashift.shifty
   shifty.Redshift frozen            3       -0.999           10

sherpa In [26]: shifty.velocity = 30
AttributeError: 'XSzashift' object has no attribute 'velocity'
```

In this case you couldn't actually use the correct parameter name.

DougBurke:
I rebased as #1227 was merged and then I had to pick up other recent changes.

anetasie:
@DougBurke I agree that the change of the default names for these parameters makes sense. 
`xsxkerbb` and `xskerbb` need to have the same parameter name - so `fcol` or `hd` but consistent for both. `fcol` is probably more meaningful as it does suggest a change in temperatures. On the other hand, the parameter indicates a hardening of the spectrum which perhaps is more appropriate for X-ray spectra, but who would understand that `hd` means hardening, so `fcol` is perhaps better name.

I agree that `redshift` is a better name for the `zashift/zmshift` models, as the units of the model is the redshift not velocity.

anetasie (approved):

DougBurke:
@anetasie - note that I am not coming up with these parameter names - I'm taking them from the XSPEC `model.dat` file. My aim is to use the XSPEC names wherever possible.

PR #1231, opened 2021-07-27
Remove __init__.py from tests directory

DougBurke:
# Summary

Clean up of a test directory. There is no functional change in this commit.

# Details

Noted during work on #1223 - I don't think we want to make this directory into a module (we don't for any other test). I can't think of a reason for keeping it but maybe there's a subtle reason?


DougBurke:
This file was added as part of https://github.com/sherpa/sherpa/pull/229 (the transition to Python 3) but not obvious why. I hadn't noted this in the review of that PR at the time, so shame on me.

hamogu (approved):
I think at some point the past, `__init__.py` files where needed for test discovery. Seems that that is no longer necessary (the number of tests run did not change due to this PR), so I agree that this file can be removed.

Issue #1232, opened 2021-07-27
Warnings from XSPEC support (clang)


DougBurke:
From a recent macOS build I noted the following warnings. Unfortunately you don't see this with `pip install -e .` whic I dislike (and I'm not sure we get this warning with `gcc` so maybe I wouldn't have seen it anyway):

```
creating build/temp.macosx-10.9-x86_64-3.8/sherpa/astro/xspec
creating build/temp.macosx-10.9-x86_64-3.8/sherpa/astro/xspec/src
compile options: '-DXSPEC_12_9_0 -DXSPEC_12_9_1 -DXSPEC_12_10_0 -DXSPEC_12_10_1 -Isherpa/include -Isherpa/utils/src -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/include/ -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/lib/python3.8/site-packages/numpy/core/include -Ibuild/src.macosx-10.9-x86_64-3.8/numpy/distutils/include -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/include/python3.8 -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/include/python3.8 -c'
extra options: '-msse -msse2 -msse3'
x86_64-apple-darwin13.4.0-clang++: sherpa/astro/xspec/src/_xspec.cc
clang-10: warning: -Wl,-export_dynamic: 'linker' input unused [-Wunused-command-line-argument]
In file included from sherpa/astro/xspec/src/_xspec.cc:27:
sherpa/include/sherpa/astro/xspec_extension.hh:248:16: warning: implicit conversion of NULL constant to 'bool' [-Wnull-conversion]
        return NULL;
        ~~~~~~ ^~~~
               false
sherpa/include/sherpa/astro/xspec_extension.hh:366:16: warning: implicit conversion of NULL constant to 'bool' [-Wnull-conversion]
        return NULL;
        ~~~~~~ ^~~~
               false
sherpa/include/sherpa/astro/xspec_extension.hh:248:16: warning: implicit conversion of NULL constant to 'bool' [-Wnull-conversion]
        return NULL;
        ~~~~~~ ^~~~
               false
sherpa/include/sherpa/astro/xspec_extension.hh:900:7: note: in instantiation of function template specialization 'sherpa::astro::xspec::create_grid<double, 12>' requested here
[...]

PR #1236, opened 2021-07-28
Remove compiler warnings

DougBurke:
# Summary

Avoid compiler warnings from clang, fixing #1232 and several other warnings.

# Details

I added a very-small note about setting CC/CXX to set the compiler and noted you can build with clang/clang++ as well as gcc/g++ to the RTD documentation.

The first commit is the most contentious, as it removes the `register` keywords. I believe that compilers nowadays ignore this setting (or don't give it much regard, see https://www.drdobbs.com/keywords-that-arent-or-comments-by-anoth/184403859 for one "rant") so I don't expect it will slow down code. It is also verboten as of C++17 as you get the following warning.

```
warning: 'register' storage class specifier is deprecated and incompatible with C++17 [-Wdeprecated-register]
```

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -138,7 +138,7 @@ The prerequisites for building from source are:
 
 * Python versions: 3.6, 3.7, 3.8
 * Python packages: ``setuptools``, ``numpy``
-* System: ``gcc``, ``g++``, ``make``, ``flex``,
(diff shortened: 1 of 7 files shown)


dtnguyen2 (approved):
I'm working getting sherpa to run on Apple M1 and it looks like some of the changes in this PR are identical to the ones I have done so I will revert my changes so I won't have to deal with conflicts later on.

PR #1241, opened 2021-07-30
to get sherpa to run Apple M1 (arm)

dtnguyen2:
# Note

This PR enables sherpa to run on Apple M1, without XSpec.

```
../../../../proj/port_ots/macOS-BigSur-ARM/Python-3.9.4.macOS-BigSur-ARM/lib/python3.9/site-packages/_pytest/cacheprovider.py:428
  /proj/port_ots/macOS-BigSur-ARM/Python-3.9.4.macOS-BigSur-ARM/lib/python3.9/site-packages/_pytest/cacheprovider.py:428: PytestCacheWarning: could not create cache path /.pytest_cache/v/cache/nodeids
    config.cache.set("cache/nodeids", sorted(self.cached_nodeids))

../../../../proj/port_ots/macOS-BigSur-ARM/Python-3.9.4.macOS-BigSur-ARM/lib/python3.9/site-packages/_pytest/stepwise.py:49
  /proj/port_ots/macOS-BigSur-ARM/Python-3.9.4.macOS-BigSur-ARM/lib/python3.9/site-packages/_pytest/stepwise.py:49: PytestCacheWarning: could not create cache path /.pytest_cache/v/cache/stepwise
    session.config.cache.set(STEPWISE_CACHE_DIR, [])

-- Docs: https://docs.pytest.org/en/stable/warnings.html
===== 3705 passed, 924 skipped, 71 xfailed, 2 warnings in 99.23s (0:01:39) =====
```

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -32,6 +32,7 @@ import string
 import sys
 from configparser import ConfigParser, NoSectionError
 import pydoc
+import platform
 
 import numpy
 import numpy.random
@@ -1303,10 +1304,8 @@ def print_fields(names, vals, converters=None):
         converters = {numpy.bool_: 'Bool',
                       numpy.bytes_: 'Bytes0',
                       numpy.complex128: 'Complex128',
-                      numpy.complex256: 'Complex256',
                       numpy.complex64: 'Complex64',
                       numpy.datetime64: 'Datetime64',
-                      numpy.float128: 'Float128',
                       numpy.float16: 'Float16',
                       numpy.float32: 'Float32',
                       numpy.float64: 'Float64',
@@ -1323,6 +1322,10 @@ def print_fields(names, vals, converters=None):
                       numpy.uint8: 'UInt8',
                       numpy.void: 'Void0'
                       }
+        if platform.processor() != 'arm' or platform.machine() != 'arm64':
+            converters[numpy.complex256] = 'Complex256'
+            converters[numpy.float128] = 'Float128'
+
 
     width = max(len(n) for n in names)
     fmt = '%%-%ds = %%s' % width
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1322,10 +1322,14 @@ def print_fields(names, vals, converters=None):
                       numpy.uint8: 'UInt8',
                       numpy.void: 'Void0'
                       }
-        if platform.processor() != 'arm' or platform.machine() != 'arm64':
+        try:
             converters[numpy.complex256] = 'Complex256'
+        except AttributeError:
+            pass
+        try:
             converters[numpy.float128] = 'Float128'
-
+        except AttributeError:
+            pass
 
     width = max(len(n) for n in names)
     fmt = '%%-%ds = %%s' % width
--- a/sherpa/utils/src/cephes/const.c
+++ b/sherpa/utils/src/cephes/const.c
@@ -91,8 +91,6 @@ double TWOOPI =  6.36619772367581343075535E-1; /* 2/pi */
 #ifdef INFINITIES
 #ifndef INFINITY
 double INFINITY = 1.0/0.0;
-#else
-double INFINITY =  1.79769313486231570815E308;    /* 2**1024*(1-MACHEP) */
 #endif
 #endif
 #ifdef NANS
--- a/sherpa/utils/src/cephes/const.c
+++ b/sherpa/utils/src/cephes/const.c
@@ -89,14 +89,16 @@ double LOGSQ2 =  3.46573590279972654709E-1;    /* log(2)/2 */
 double THPIO4 =  2.35619449019234492885;       /* 3*pi/4 */
 double TWOOPI =  6.36619772367581343075535E-1; /* 2/pi */
 #ifdef INFINITIES
-double INFINITY = strtod("INF");  /* 99e999; */
+#ifndef INFINITY
+double INFINITY = 1.0/0.0;
 #else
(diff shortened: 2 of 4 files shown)


sherpa/utils/__init__.py
```diff
@@ -1323,6 +1322,10 @@ def print_fields(names, vals, converters=None):
                       numpy.uint8: 'UInt8',
                       numpy.void: 'Void0'
                       }
+        if platform.processor() != 'arm' or platform.machine() != 'arm64':
```
DougBurke, line 1325: I think it would be better to just check whether the attributes exist rather than base it on the platform type (e.g. windows may not have these symbols, and whilst we currently don't support a windows build we should try to avoid hard-coding logic when we can):

```python
try:
    converters[numpy.complex256] = 'Complex256'
except AttributeError:
    pass

try:
    converters[numpy.float128] = 'Float128'
except AttributeError:
    pass
```

Note above code is untested...

DougBurke:
Are we waiting for SDS review on this?

dtnguyen2:
Still working on this

sherpa/utils/__init__.py
dtnguyen2: made the changes you requested

DougBurke: Ta

DougBurke (approved):
FWIW LGTM (I am just assuming that the defines make sense as we currently don't have any CI to test this out).

sherpa/utils/src/cephes/const.c
```diff
@@ -89,14 +89,16 @@ double LOGSQ2 =  3.46573590279972654709E-1;    /* log(2)/2 */
 double THPIO4 =  2.35619449019234492885;       /* 3*pi/4 */
 double TWOOPI =  6.36619772367581343075535E-1; /* 2/pi */
 #ifdef INFINITIES
-double INFINITY = strtod("INF");  /* 99e999; */
+#ifndef INFINITY
```
wmclaugh, line 92: The preprecessor conditionals here change the behavior compared to the original code due to the nesting. Is this intended?

in the original-  if INFINITIES is not defined, INFINITY is set to 1.79...

    #ifdef INFINITIES
        double INFINITY = strtod("INF");  /* 99e999; */
    #else
        double INFINITY =  1.79769313486231570815E308;    /* 2**1024*(1-MACHEP) */
    #endif

in the modified code- if INFINITIES is not defined, nothing happens...

    #ifdef INFINITIES
        #ifndef INFINITY
            double INFINITY = 1.0/0.0;
        #else
            double INFINITY =  1.79769313486231570815E308;    /* 2**1024*(1-MACHEP) */
        #endif
    #endif

dtnguyen2: The following simple program will test if one can use ~1.79e308 and 0 for infinity and NaN, respectively; Moreover a check if 1.0/0.0 and 1.0/0. - 1/0.0 are legit definitions for infinity and NaN:

```
devel12-107: cat tst_inf_nan.c 
#include <stdio.h>
#include <math.h>
#include <stdlib.h>

/*double tmp = sqrt(-1); */


double myinf = (1.0/0.0);
double not_inf = 1.79769313486231570815E308;
double mynan = (1.0/0. * 0.0);
double not_nan = 0.0;

int main() {

  printf("myinf = %g\tis myinf inf: %d\n", myinf, isinf(myinf));
  printf("not_inf = %g\tis not_inf inf: %d\n", not_inf, isinf(not_inf));

  printf("mynan = %g\tis mynan nan: %d\n", mynan, isnan(mynan));
  printf("not_nan = %g\tis not_nan nan: %d\n", not_nan, isnan(not_nan));

  return 0;
}
```

As you can see, if one uses the builtin ```isinf``` and ```isnan`` to test out a few numbers:

```
devel12-108: gcc tst_inf_nan.c 
devel12-109: a.out
myinf = inf	is myinf inf: 1
not_inf = 1.79769e+308	is not_inf inf: 0
mynan = nan	is mynan nan: 1
not_nan = 0	is not_nan nan: 0
```
QED


There are two problems with the line:
```
double INFINITY = strtod("INF");  /* 99e999; */
```
1) The function ```strtod``` takes two arguments:
```
NAME
       strtod, strtof, strtold - convert ASCII string to floating-point number

SYNOPSIS
       #include <stdlib.h>

       double strtod(const char *nptr, char **endptr);
       float strtof(const char *nptr, char **endptr);
       long double strtold(const char *nptr, char **endptr);
```
2) modern compilers will only allow one to call a function from within a function.  If I comment out the call to the ```sqrt``` function like:

```
#include <stdio.h>
#include <math.h>
#include <stdlib.h>

double tmp = sqrt(-1);


double myinf = (1.0/0.0);
double not_inf = 1.79769313486231570815E308;
double mynan = (1.0/0. * 0.0);
double not_nan = 0.0;

int main() {

  printf("myinf = %g\tis myinf inf: %d\n", myinf, isinf(myinf));
  printf("not_inf = %g\tis not_inf inf: %d\n", not_inf, isinf(not_inf));

  printf("mynan = %g\tis mynan nan: %d\n", mynan, isnan(mynan));
  printf("not_nan = %g\tis not_nan nan: %d\n", not_nan, isnan(not_nan));

  return 0;
}
```
So I can't compile it:

```
devel12-115: gcc tst_inf_nan.c
tst_inf_nan.c:5:1: error: initializer element is not constant
 double tmp = sqrt(-1);
 ^
devel12-116: 
```

harlantc (approved):
This looks good to me.

dtnguyen2:
Just realized that when I made the changes: checking if the platform is an ```arm``` as suggested by @DougBurke then I probably should have deleted the following import:
```
import platform
```

PR #1246, opened 2021-08-11
Image region filters are now specified correctly (Fix #1245)

### LINKED ISSUE #1245: image filters
PR #968 fixed crashes seen in #965 where multiple filters would segfault. In doing some work to extend #1244 to fix #437 I've found an error with multiple filters. Using the `img.fits` file from the test data area we can first check that all pixel values are > 0 in this file (useful later on):

```
% dmstat img.fits cen-
PRIMARY
    min:        1.3310212125e-05              @:        ( 16 44 )
    max:        0.99995952845         @:        ( 96 95 )
   mean:        0.50414351214
  sigma:        0.2887851084
    sum:        5041.4351214
   good:        10000
   null:        0
```

Let's try adding some filters:

```python
>>> load_image('img.fits')
>>> get_filter()
''
>>> notice2d("Circle(50,50,30)")
>>> calc_data_sum2d("circle(50,50,30)")
1424.1254567484866
>>> image_data()
```

sherpa1

```python
>>> notice2d()
>>> get_filter()
''
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> calc_data_sum2d("Ellipse(40,75,30,20,320)")
946.35223559734
>>> image_data()
```

sherpa2

So, we have two regions with counts of ~1400 and ~950 respectively. They overlap, so all we can say is that the total of the combined filter should be > 1400 (since no pixel value is negative):

```python
>>> notice2d()
>>> get_filter()
''
>>> notice2d("Circle(50,50,30)")
>>> get_filter()
'Circle(50,50,30)'
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> get_filter()
'Circle(50,50,30)&Ellipse(40,75,30,20,320)'
>>> image_data()
```

sherpa3

So we can see they've been combined. Let's check the number of counts in this region (for some reason `calc_data_sum2d ignores the filter so we need to give it):

```python
>>> calc_data_sum2d('Circle(50,50,30)&Ellipse(40,75,30,20,320)')
489.92763453030784
```

Hold on, it's << 1400. The reason is that the filter expression uses & to combine the shapes - it should use | (ie or, not and), which gives a much better output.

```python
>>> calc_data_sum2d('Circle(50,50,30)|Ellipse(40,75,30,20,320)')
1880.5500578155188
```

If I use the two filters in one call then we see incorrect behavior for the & case:

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)&Ellipse(40,75,30,20,320)")
>>> image_data()
```

sherpa4

where I've added on the two regions to show we are seeing the intersection, versus the expected:

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)|Ellipse(40,75,30,20,320)")
>>> image_data()
```

sherpa6

If you start removing shapes then you get some interesting behavior

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)")
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> ignore2d("RotBox(30,30,10,5,45)")
>>> image_data()
>>> get_filter()
'Circle(50,50,30)&Ellipse(40,75,30,20,320)&!RotBox(30,30,10,5,45)'
```

sherpa7

[this image is correct, it's just that the filter expression is wrong]

The addition of the negation should lead to a more-complex expression (due to the region library) where the negation gets applied to both shapes, but it doesn't. If we use the CIAO region library we can see this (but note that '+' maps to '|' and '*' to '&'):

```python
>>> from region import CXCRegion
>>> r1 = CXCRegion("Circle(50,50,30)")
>>> r2 = CXCRegion("Ellipse(40,75,30,20,320)")
>>> r3 = CXCRegion("RotBox(30,30,10,5,45)")
>>> r1 + r2
circle(50,50,30)+ellipse(40,75,30,20,320)
>>> r1 | r2
circle(50,50,30)+ellipse(40,75,30,20,320)
>>> r1 * r2
circle(50,50,30)*ellipse(40,75,30,20,320)
>>> (r1 + r2) - r3
circle(50,50,30)*!rotbox(30,30,10,5,45)+ellipse(40,75,30,20,320)
```

See how the rotbox has been applied to the circle here (technically it should be applied to both but the region library knows that the box doesn't overlap the ellipse so can be dropped).

DougBurke:
TODO: the RTD documentation could be improved to add an image case to https://sherpa.readthedocs.io/en/latest/data/index.html#astronomy-data and something to match https://sherpa.readthedocs.io/en/latest/developer/index.html#pha-filtering - however I claiim that is not required for this PR, just a nice to have, and shouldn't stop anyone from reviewing it.

# Summary

Update the image filter code so that the region description matches what the filter actually is: rather than use `&` to combine regions we now use `|` as it is a logical or rather than logical and. This is mainly going to affect users who 

- have used the output of `get_filter()` as input to `calc_data_sum2d`
- have very-complex spatial filters
- are restoring Sherpa sessions created by `save` or `save_all`.

Fixes #1245.

# Details

Note that the changes here are all in very-low-level code: users (even those using the object layer and not the ui layer) are not going to directly interact with Region objects - therefore I feel okay making relatively large API changes as anyone who is affected by these changes is probably doing something they shouldn't be! This can be thought of as the logical conclusion of #968 - I had thought I'd got it right then but obviously not. I am not an expert in the region library so I don't know how the filters seemed to be doing the right thing while the string representation was wrong (creating & rather than |). If I had added tests that had tried to re-create a filter from the region `__str__` method then perhaps I'd have found this when developing #968 (and I want to point out that these problems were not added in #968, they existed in the code before that PR its just that using this code quickly lead to a segfault so you couldn't really appreciate this error) .

I was going to clean up the region code initializer as it can be a lot-more readable (following advice from the Python documentation), but then realized that needs C99 style designated initalizers which we don't have in C++. The initalization code is now more-resistant to errors (but in reality it's not clear how useful this is as if it fails then something is seriously wrong anyway). These changes are suggested by https://docs.python.org/3/c-api/index.html and references therein.

After a few commits that either adds tests or cleans up the test code, we get to the first important change: remove support for the empty region. This is something that may have been present before #968 but I did add tests for this functinoality in #968. However, we don't actually use this capability (the ability to create an "empty" region, which is not well defined) so I removed it. Note that this only changes the low-level region test code (where I had added explicit tests of this behavior, including some with notes like "it's not obvious what this should do but let's just check we get back a known response"), which shows we don't use empty regions in Sherpa.

The next change is similar in that it changes low-level routines (but that are used by Sherpa, but only in one place in sherpa.astro.data.DataIMG). The region library had a combine method for combining two regions, with a flag to say whether it should be a subtraction or not. I have removed the combine method and replaced it with union (for adding together two regions) and subtract (for subtracting the second region from the first). As well as being (I feel) a bit more obvious when in use, it does improve the code slightly (as the two methods are simpler) at the expense of a little-bit of replicated code.

The final change is to correct these two methods so that they use the correct method for combining the regions: for union we use `regUnionRegion` and subtract we use `regIntersectRegion` whereas they both used to use `regCombineRegion` (which is what was used before #968). My guess is that this code was never updated when the region library gained these routines (since the use of `regCombineRegion` was present in the first git commit). I chose `regUnionRegion` and `regIntersectRegion` based on work Kenny did in generating the SDS prototype code for what became the region.CXCRegion Python class in CIAO. I have left in support for the invert method even though we don't use it in sherpa.astro.data.DataIMG "just in case". It is in this commit that I cange how we handle the `notice2d(); ignore2d(shape);` case - rather than inverting the region we subtract it from the catch-all "Field()" case as that fixes some odd region outputs we were seeing in the tests. 

# Notes

In general the test changes are in the `test_region_unit.py` file, which is the low-level tests for the Region class. This is important to test but it is not user facing. The number of test changes required for the "user-level" code is much smaller (but that may just suggest we don't have enough tests of the image filtering functionality).

Sherpa's use of the region library is much "smaller" than the CIAO region module (in particular region.CXCRegion), as all we really need are union, subtract, and possibly invert. We could think about adding support for intersection and possibly even area calculations, but that is not really relevant for this PR.

#1247 adds tests that are likely to change with this code so we'd want that merged first.

# Examples

## Example from 1245

Following the example in #1245 we now have

```python
>>> load_image('img.fits')
>>> get_filter()
''
>>> notice2d("Circle(50,50,30)")
>>> calc_data_sum2d("circle(50,50,30)")
1424.1255
>>> get_filter()
'Circle(50,50,30)'
>>> image_data()   # matches the result from #1245
```

```python
>>> notice2d()
>>> get_filter()
''
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> get_filter()
'Ellipse(40,75,30,20,320)'
>>> calc_data_sum2d("Ellipse(40,75,30,20,320)")
946.3522
>>> image_data()   # matches the result from #1245
```

We have the first difference here in the `get_filter` `output` - it uses `|` rather than `&` to combine the regions  (and it should be `|` or `+` as this is OR - see https://cxc.harvard.edu/ciao/ahelp/dmregions.html - as `&` is AND which isn't technically correct, even though the actual filter as shown by `image_data` hasn't apparently changed):

```python
>>> notice2d()
>>> get_filter()
''
>>> notice2d("Circle(50,50,30)")
>>> get_filter()
'Circle(50,50,30)'
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> get_filter()
'Circle(50,50,30)|Ellipse(40,75,30,20,320)'
>>> image_data()   # matches the result from #1245
```

We can see the difference of `&` vs `|` in the `calc_data_sum2d` output:

```python
>>> calc_data_sum2d('Circle(50,50,30)&Ellipse(40,75,30,20,320)')
489.9276
>>> calc_data_sum2d('Circle(50,50,30)|Ellipse(40,75,30,20,320)')
1880.55
```

The second version, using `|`, makes more sense for what we are trying to do (this is just repeating the text from #1245 here...)

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)&Ellipse(40,75,30,20,320)")
>>> get_filter()
'Circle(50,50,30)&Ellipse(40,75,30,20,320)'
>>> image_data()   # matches the result from #1245
```

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)|Ellipse(40,75,30,20,320)")
>>> get_filter()
'Circle(50,50,30)|Ellipse(40,75,30,20,320)'
>>> image_data()   # matches the result from #1245
```

In the following I noted in #1245 that the image_data was correct but the filter expression was wrong - you can know see it is "correct" - we have `circle1-rotbox` combined with `circle2` (the region library has been smart enough to recognize that `rotbox` doesn't overlap `circle2` so it can simplify `circle1-rotbox | circle2-rotbox` to the filter below:

```python
>>> notice2d()
>>> notice2d("Circle(50,50,30)")
>>> notice2d("Ellipse(40,75,30,20,320)")
>>> ignore2d("RotBox(30,30,10,5,45)")
>>> get_filter()
'Circle(50,50,30)&!RotBox(30,30,10,5,45)|Ellipse(40,75,30,20,320)'
>>> image_data()   # matches the result from #1245
```

One thing we now have is that applying the filter from get_filter to an unfiltered image will now correctly create the filter, when previously it didn't because it was using & rather than |. 

## Just negating things

One interesting piece I had to do is handle a case of "first filter is a subtraction" a little differently - we now expilctly add a `Field()` filter because without it strange things could happen (we could have used a bounding box around the image rather than `Field()` which would have some advantages but also difficulties). So we now have:

```python
>>> image_data()
>>> notice2d()
>>> ignore2d("circle(50,50,30)")
>>> get_filter()
'Field()&!Circle(50,50,30)'
>>> image_data()    # sorry, forgot to grab a screenshot
>>> ignore2d("Ellipse(40,75,30,20,320)")
>>> get_filter()
'Field()&!Circle(50,50,30)&!Ellipse(40,75,30,20,320)'
>>> image_data()
```

sherpa

```python
>>> ignore2d("RotBox(30,30,10,5,45)")
>>> get_filter()
'Field()&!Circle(50,50,30)&!Ellipse(40,75,30,20,320)&!RotBox(30,30,10,5,45)'
>>> image_data()
```

sherpa

(note that it does exclude a few more pixels than the previous version).

Before this PR the result was the same, but without the Field statements:

```python
>>> load_image('img.fits')
>>> get_filter()
''
>>> ignore2d("circle(50,50,30)")
>>> get_filter()
'!Circle(50,50,30)'
>>> image_data()
>>> ignore2d("Ellipse(40,75,30,20,320)")
>>> get_filter()
'!Circle(50,50,30)&!Ellipse(40,75,30,20,320)'
>>> image_data()
>>> ignore2d("RotBox(30,30,10,5,45)")
>>> get_filter()
'!Circle(50,50,30)&!Ellipse(40,75,30,20,320)&!RotBox(30,30,10,5,45)'
>>> image_data()
```

However, with the changes to how `union` and `subtract` work for the Region object (note: this is all internal stuff that users, even those using the object layer, would use) if we didn't add the field() statement the filter expressions after we removed the second region would have been `""`, which is just wrong (although the image_data output would have been correct)

## an example reading regions from files

https://gist.github.com/DougBurke/a14a78a11d5d9fed2a32c402d05b263e shows a notebook run with the main branc (`test_old.ipynb`) and this PR (`test_new.ipynb`) which shows that the filter returned by `get_filter` can not reliably be used before this PR but can with this one. The filters created by the `notice2d` calls select the same pixels in this case (which is reasonably complex), it's just that if you try to re-create the filter you get a problem.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4209,7 +4209,10 @@ class DataIMG(Data2D):
         #
         if self._region is None:
             if ignore:
-                reg.invert()
+                # add an explicit "whole field" constructor to avoid
+                # possible issues with stringification of multiple
+                # ignores.
+                reg = Region('field()').subtract(reg)
 
             self._region = reg
         elif ignore:
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4212,8 +4212,10 @@ class DataIMG(Data2D):
                 reg.invert()
 
             self._region = reg
+        elif ignore:
+            self._region = self._region.subtract(reg)
         else:
-            self._region = self._region.combine(reg, ignore)
+            self._region = self._region.union(reg)
 
     def get_bounding_mask(self):
         mask = self.mask
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -730,7 +730,7 @@ def test_img_get_filter_exclude(make_test_image):
     shape = 'ellipse(4260,3840,3,2,0)'
     d.notice2d(shape, ignore=True)
 
-    expected = '!' + shape.capitalize()
+    expected = 'Field()&!' + shape.capitalize()
     assert d.get_filter() == expected
 
 
@@ -758,7 +758,7 @@ def test_img_get_filter_combined(make_test_image):
     d.notice2d(shape2)
 
     shape2 = shape2.replace('rect', 'rectangle')
-    shape = shape1.capitalize() + '&' + shape2.capitalize()
+    shape = shape1.capitalize() + '|' + shape2.capitalize()
     assert d.get_filter() == shape
 
 
@@ -787,6 +787,9 @@ def check_ignore_ignore(d):
     mask1 = ~Region(shape1).mask(d.x0, d.x1).astype(bool)
     assert d.mask == pytest.approx(mask1)
 
+    expected = 'Field()&!' + shape1.capitalize()
+    assert d.get_filter() == expected
+
     shape2 = 'rect(4258,3830,4264,3841)'
     d.notice2d(shape2, ignore=True)
 
@@ -794,7 +797,7 @@ def check_ignore_ignore(d):
     assert d.mask == pytest.approx(mask1 & mask2)
 
     shape2 = shape2.replace('rect', 'rectangle')
-    expected = '!' + shape1.capitalize() + '&!' + shape2.capitalize()
+    expected = 'Field()&!' + shape1.capitalize() + '&!' + shape2.capitalize()
     assert d.get_filter() == expected
 
 
@@ -810,9 +813,23 @@ def test_img_get_filter_included_excluded(make_test_image):
 def test_img_get_filter_excluded_excluded(make_test_image):
     """Simple get_filter check on an image.
 
-    Here we want to check the behavior when d.mask is False.
-    I am not sure this makes sense, but this is done to
-    show the current behavior.
+    Here we want to check the behavior when d.mask is False.  I am not
+    sure this makes sense, but this is done to show the current
+    behavior. Note that d.notice2d(ignore=True) is meant to ignore all
+    points but it (currently) doesn't add a region since we know from
+    the mask all the points are ignored and so there's no need to add
+    a "no region" filter: if you have !field() and then union s1 we
+    would get
+
+        !field()|s
+
+    but this is the same as
+
+        s
+
+    Instead if we do !field().subtract(s) then it's the same as
+    !field(). There probably is something we could improve here.
+
     """
     d = make_test_image
 
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -798,26 +798,6 @@ def check_ignore_ignore(d):
     assert d.get_filter() == expected
 
 
-def check_ignore_ignore2(d):
-    """Check removing the shapes works as expected."""
-
-    shape1 = 'ellipse(4260,3840,3,2,0)'
-    d.notice2d(shape1, ignore=True)
-
-    mask1 = ~Region(shape1).mask(d.x0, d.x1).astype(bool)
-    assert d.mask == pytest.approx(mask1)
-
-    shape2 = 'rect(4258,3830,4264,3841)'
-    d.notice2d(shape2, ignore=True)
-
-    mask2 = ~Region(shape2).mask(d.x0, d.x1).astype(bool)
-    assert d.mask == pytest.approx(mask1 & mask2)
-
-    shape2 = shape2.replace('rect', 'rectangle')
-    expected = '!' + shape1.capitalize() + '&!' + shape2.capitalize()
-    assert d.get_filter() == expected
-
-
 def test_img_get_filter_included_excluded(make_test_image):
     """Simple get_filter check on an image.
 
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -1,5 +1,6 @@
 //
-//  Copyright (C) 2009, 2016, 2020  Smithsonian Astrophysical Observatory
+//  Copyright (C) 2009, 2016, 2020, 2021
+//  Smithsonian Astrophysical Observatory
 //
 //
 //  This program is free software; you can redistribute it and/or modify
@@ -178,6 +179,9 @@ static PyMethodDef pyRegion_methods[] = {
 };
 
 
+// Unfortunately we can not use C99-style designated initializers here
+// so we need to keep all the "null" values.
+//
 static PyTypeObject pyRegion_Type = {
   // Note that there is no semicolon after the PyObject_HEAD_INIT macro;
   // one is included in the macro definition.
@@ -319,8 +323,10 @@ PyInit__region(void)
     return NULL;
 
   Py_INCREF(&pyRegion_Type);
-
-  PyModule_AddObject(m, (char*)"Region", (PyObject *)&pyRegion_Type);
+  if (PyModule_AddObject(m, (char*)"Region", (PyObject *)&pyRegion_Type) < 0) {
+    Py_DECREF(&pyRegion_Type);
+    Py_DECREF(m);
+  }
   return m;
 
 }
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -37,7 +37,8 @@ typedef struct {
 
 
 // Declare symbols
-static PyObject* region_combine( PyRegion* self, PyObject* args, PyObject *kwargs );
+static PyObject* region_union( PyRegion* self, PyObject* args, PyObject *kwargs );
+static PyObject* region_subtract( PyRegion* self, PyObject* args, PyObject *kwargs );
 static PyObject* region_invert( PyRegion* self, PyObject* args );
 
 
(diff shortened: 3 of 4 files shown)


anetasie:
I just tried a few examples using this branch with ds9. It looks good. I must say that I was not aware of the | vs & impact on `calc_data_sum2d()`. I guess we need to include some of these example in the CIAO documentation.

anetasie (approved):
I run the branch to check the user functionality. It looks good.

DougBurke:
https://cxc.cfa.harvard.edu/ciao/ahelp/dmregions.html

dtnguyen2:
Source the CAT5.4 build:

```
devel12-121: source /proj/cm/Integ/install.CAT5.4/bin/ciao.csh -o
CIAO configuration is complete... 
The CIAO setup for this window has changed from
CAT5.4  Friday, September 17, 2021
  bindir      : /proj/cm/Integ/install.CAT5.4/bin
  CALDB       : Not installed
To:
CAT5.4  Friday, September 17, 2021
  bindir      : /proj/cm/Integ/install.CAT5.4/bin
  CALDB       : Not installed
```

```Python
devel12-122: awk 'NR >= 4251 && 4260 >= NR' < `which mles.py`
    def calc_rdata(self, region2sort, ids2fit):
        self.load_region_pixmask(region2sort, 'calc raw data to sort')
        rdata = 0
        for id in ids2fit:
            set_coord(id, 'physical')
            rdata += calc_data_sum2d(get_filter(id), id=id)
            if self.params['verbose'] > 3:
                print('calc_data_sum2d[%d] = %g' % (id, rdata))
        return rdata
```

MLE does use the output of ```get_filter()``` to input to ```calc_data_sum2d```.

PR #1247, opened 2021-08-11
Allow save_all to work well with images with no filters (Fix #437)

### LINKED ISSUE #437: save_all with image datasets can create empty regions which confuse notice2d_id
# Summary

`save_all`, when called with a image dataset with no spatial filter applied (by Sherpa), stores a line saying `notice2d_id(<idval>, "")`, which then fails when executed with a `unable to parse region string successfully` error.

A user can manually delete (or edit out) this line, but we should fix it so that the line either isn't added to the output of `save_all`, or sending in an empty region filter to `notice2d_id` works (or both).

# Details

This is with our test version of CIAO 4.10 but it happens in CIAO 4.9 (so has been around for a while).

```
sherpa-1> load_image(1, "fimg/broad_thresh.img")
sherpa-2> save_all('delme.shp')
```

In the output file we have this `notice2d_id` call:

```
% grep notice delme.shp 
notice2d_id(1, "")
```

When loaded in (the `sys.tracebacklimit` is needed to actually see the error when using ipython v4, since it works around an issue that has, I believe, been fixed in ipython v5 and later):

```
sherpa-1> sys.tracebacklimit = None
sherpa-2> %run delme.shp
/home/djburke/data/aneta/12715/delme.shp in <module>()
     21 ######### Filter Data
     22 
---> 23 notice2d_id(1, "")
     24 
     25 

<string> in notice2d_id(ids, val)

/home/djburke/testlocal/u35/ciao-4.10/lib/python3.5/site-packages/sherpa-ciao4.10-py3.5-linux-x86_64.egg/sherpa/astro/ui/utils.py in notice2d_id(self, ids, val)
   6426             _check_type(self.get_data(id), sherpa.astro.data.DataIMG,
   6427                         'img', 'a image data set')
-> 6428             self.get_data(id).notice2d(val, False)
   6429 
   6430     def ignore2d_id(self, ids, val=None):

/home/djburke/testlocal/u35/ciao-4.10/lib/python3.5/site-packages/sherpa-ciao4.10-py3.5-linux-x86_64.egg/sherpa/astro/data.py in notice2d(self, val, ignore)
   2450              mask) = region_mask(self._region, val,
   2451                                  self.get_x0(), self.get_x1(),
-> 2452                                  os.path.isfile(val), ignore)
   2453             mask = numpy.asarray(mask, dtype=numpy.bool_)
   2454         else:

TypeError: unable to parse region string successfully
sherpa-3> 
```

DougBurke:
# Summary

Fix a problem with the output of the `save_all` command if an image had been loaded but no region filter applied (fix #437): the output file would contain a `notice2d_id` call with a filter of `""` which would fail if you tried to load the file back in. It now no-longer creates this line.

# Details

All I do is to stop the problematic notice2d_id call from being created. I could have changed noticed2_id to either do nothing, or except everything, if given a filter of '', but I decided not to do this at this time. I note that the logic of get_filter here - returning '' when there's no region filter applied - is somewhat different to other cases (where '' implies "all data has been filtered out) and so I decided I didn't want to go making changes here in this area as it requires more time than I have for the 4.14 release.

### DIFF
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -333,8 +333,10 @@ def _handle_filter(state, id, fh):
         cmd = 'notice_id({}, "{}")'.format(cmd_id, fvals)
         _output(cmd, fh)
     elif ndims == 2:
-        cmd = 'notice2d_id({}, "{}")'.format(cmd_id, fvals)
-        _output(cmd, fh)
+        # Only output the filter if it does anything
+        if fvals != '':
+            cmd = 'notice2d_id({}, "{}")'.format(cmd_id, fvals)
+            _output(cmd, fh)
     else:
         # just in case
         _output('print("Set notice range of id={} to {}")'.format(cmd_id,
--- a/sherpa/astro/ui/tests/test_serialize.py
+++ b/sherpa/astro/ui/tests/test_serialize.py
@@ -872,7 +872,6 @@ set_coord(1, 'logical')
 
 ######### Filter Data
 
-notice2d_id(1, "")
 
 
 ######### Set Statistic
@@ -1502,7 +1501,6 @@ def test_restore_usermodel():
     assert ui.calc_stat(3) == pytest.approx(statval)
 
 
-@pytest.mark.xfail
 @requires_data
 @requires_fits
 def test_restore_img_no_filter_no_model(make_data_path, clean_astro_ui):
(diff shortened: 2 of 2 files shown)


hamogu (approved):
Please open an new issue to address this problem in more depth later (after 4.14) - maybe you did this already, but I lost track if all the new issues in the last few days.

Issue #1249, opened 2021-08-12
PHA mask handling and grouping


DougBurke:
This issue is to remind me in the future to look back at this idea.

We currently create a mask at the current grouping size (so if you have 46 groups then you have 46 elements of mask) - for the `DataPHA` class.

I wonder if it would be better if we always kept the mask in ungrouped form and then rebin it when needed with the current grouping scheme [which would need some decision about partially-filled groups].

I believe (based on work I'm doing to fix #1024) this would help some issues we have with grouping/ungrouping/filtering, but I have not thought about it deeply enough to decide whether it's worth trying.

DougBurke:
Hmm. The work in #1470 suggests that we probably already have too much code that assumes a "grouped" mask, and there are cases it could be useful.

One option would be to allow either the ungrouped or grouped data to be sent in, but that is potentially confusing for the user.

I to think it would be good to be able to identify un-grouped vs  grouped data, but it would probably require a large re-work of the data object. Perhaps if we look at #1450 and #1449 ?

Issue #1220, opened 2021-07-23
what should we do when all data has been ignored?


DougBurke:
This is a problem across our data objects (i.e. it's not just a DataPHA issue). I am using the main branch behavior but I don't think it significatly changes once we get #1219 (and its precursors) merged

When all the data has been masked out we can get a `DataErr` thrown, or `''` returned, or (for PHA) `"No noticed bins"`. This is not ideal.

Note that there's a `get_filter` issue and a more-general issue:

a) for `get_filter` we should have the same behavior however we've ignored the data
b) when we've filtered out all the data do we expect to always raise a `DataErr` when accessing the data, so `get_filter` but also data accessors either called directly like `get_x`, `get_xerr`, `get_indep`, `get_y`, `get_quality`, ..., or indirectly when plotting data, fitting, .... I think it's probably a per "approach" ca;; (ie for fitting throwing an error may make sense but for data accessors we might want to return an empty array)

# What is going on for `get_filter`

The `mask` attribute has three settings: `True` (created with `notice()`), `False` (created with `ignore()`), and an array of bools (created once you specify a limit with a `notice` or `ignore` call).  These three states can lead to different `get_filter` outputs **I think**, but there is also a DataErr that is being raised from some code paths and not others. For `mask` I understand the `True` and array of boleans, but I wonder if we need the `False` state.

# Examples

These examples shows`get_filter` for the moment- I'll add more later

## Data1D

```python
>>> from sherpa.data import Data1D
>>> d = Data1D('x', [1,2,3], [1,1,1])
/home/dburke/anaconda/envs/ciao414b1/lib/python3.8/site-packages/sherpa/data.py:50: UserWarning: Converting array [1, 1, 1] to numpy array.
  warnings.warn("Converting array {} to numpy array.".format(array))
>>> d.get_filter(format='%.4f')
'1.0000:3.0000'
>>> d.ignore()
>>> d.get_filter(format='%.4f')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dburke/anaconda/envs/ciao414b1/lib/python3.8/site-packages/sherpa/data.py", line 1084, in get_filter
    x = self.get_x(filter=True)
  File "/home/dburke/anaconda/envs/ciao414b1/lib/python3.8/site-packages/sherpa/data.py", line 995, in get_x
    return self.get_evaluation_indep(filter, model, use_evaluation_space)[0]
[...]

PR #1250, opened 2021-08-12
Stop odd interaction between plot_model and PHA filters (Fix #1024)

### LINKED ISSUE #1024: PHA data, models, and notice - what is going on

DougBurke:
# Summary

plot_model(), if done before calling notice or ignore, could lead to strange filters (e.g. filter not getting applied). This has been fixed. Fixes #1024 

# Details

The sherpa.astro.plot.ModelHistogram class is used by `plot_model` when plotting PHA data. The logic it used to temporarily ungroup the data just for the plot looks okay but invalidates an invariant used by the filter code, which triggers the behavior seen in #1024. I have re-written the code to be *really* explict, which fixes the problem (i.e. we only bother doing the filter re-creation - needed because we don't have #1249 or something similar - when it is needed, rather than whenever the data is grouped).

There is one small change in behavior if you have filtered out all points using ignore calls with explicit ranges since prior to this case the plot would be created by plot_model, after miraculously removing all filters (i.e. an example of #1024), whereas we now error out. If you had called ignore() - i.e. with no arguments - then you get an error (before and after). This is partly related to #1220

# Notes

This PR conflicts with #1216/#1219 (in that the tests will need to be updated to reflect the changes made in #1219 I would guess). As this is an easy review we could do this first and I can update the other PRs.

This has been pulled out of #1161 as it's not clear what I'm going to do with that PR.

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -204,32 +204,57 @@ class ModelPHAHistogram(HistogramPlot):
 class ModelHistogram(ModelPHAHistogram):
     """Plot a model for a PHA dataset with no grouping.
 
-    The model is drawn at the native resolution of
-    the instrument response, or ungrouped channels.
-    The model is filtered to the minimum and maximum
-    of the PHA dataset but any ignored ranges within
-    this are ignored.
+    The model is drawn at the native resolution of the instrument
+    response, or ungrouped channels.
+
     """
 
     def prepare(self, data, model, stat=None):
 
-        old_filter = parse_expr(data.get_filter())
-        old_group = data.grouped
+        # We could fit this into a single try/finally group but
+        # it makes it harder to see what is going on so split
+        # it out into separate plot types:
+        #   - ungrouped
+        #   - all data has been masked out (we let prepare
+        #     throw any errors or plot no data)
+        #   - grouped but no filter (or the filter doesn't
+        #     remove any points).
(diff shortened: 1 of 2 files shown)


anetasie:
@DougBurke I just compiled and tried.  I could see issue in 4.14 and confirmed that this is not fixed with this PR.

anetasie (commented):
I checked the functionality, but did not check the code. I leave it for @hamogu.

DougBurke:
@anetasie in https://github.com/sherpa/sherpa/pull/1250#issuecomment-898664737 you say

> I could see issue in 4.14 and confirmed that this is not fixed with this PR.

So you mean `that this is now fixed` ?

DougBurke:
If it is fixed @anetasie can you approve the review jus tto make it obvious you are happy with it?

anetasie (approved):

hamogu (approved):

Issue #47, opened 2015-06-16
BUG: sherpa.astro.ui.save_table fails with crates backend: "Unable to create column HEADER: unsupported datatype ('numpy.object_')"


DougBurke:
Using CIAO 4.7 (OS-X), with Sherpa started up in the top-level of the code distribution (for easy access to a PHA test file):

```
% sherpa
-----------------------------------------------------
Welcome to Sherpa: CXC's Modeling and Fitting Package
-----------------------------------------------------
CIAO 4.7 Sherpa version 1 Thursday, December 4, 2014

sherpa-1> import sys
sherpa-2> sys.tracebacklimit = 10
sherpa-3> load_pha('sherpa/astro/datastack/tests/data/3c273.pi')
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file sherpa/astro/datastack/tests/data/3c273.arf
read RMF file sherpa/astro/datastack/tests/data/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file sherpa/astro/datastack/tests/data/3c273_bg.pi
sherpa-4> save_table(1, '/tmp/tbl.fits', ascii=False, clobber=True)
  File "/export/local/ciao-4.7/ots/lib/python2.7/site-packages/IPython/core/interactiveshell.py", line 2883, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-6a29ee71142f>", line 1, in <module>
    save_table(1, '/tmp/tbl.fits', ascii=False, clobber=True)
  File "<string>", line 1, in save_table
  File "/export/local/ciao-4.7/lib/python2.7/site-packages/sherpa/astro/ui/utils.py", line 3162, in save_table
    ascii, clobber)
  File "/export/local/ciao-4.7/lib/python2.7/site-packages/sherpa/astro/io/__init__.py", line 391, in write_table
    backend.set_table_data(filename, data, names, ascii=ascii, clobber=clobber)
[...]

PR #1253, opened 2021-08-13
More numpy 1.20 warning fixes

DougBurke:
# Summary

Avoid warnings from NumPy 1.20 about using the deprecated np.int symbol.

# Details

PR #1138 was supposed to fix the NumPy warnings but mu grep-fu must have been lacking. I've added tests to `save_filter` (and related functions) which show the issue (i.e. a reason why I missed them in #1138 is because we had no test that executed these lines)

I also add a test of #47 (currently marked as xfail) so we'll know when that fails (I wanted to make sure that we had tests for the clean-up code changes I made).

I've marked this as a blocker issue as I would want to have numpy 1.20 issues fixed before we release.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -4352,24 +4352,32 @@ class Session(sherpa.ui.utils.Session):
         if filename is None:
             id, filename = filename, id
         _check_type(filename, string_types, 'filename', 'a string')
-        d = self.get_data(id)
-        if bkg_id is not None:
-            d = self.get_bkg(id, bkg_id)
         id = self._fix_id(id)
 
+        if bkg_id is not None:
+            d = self.get_bkg(id, bkg_id)
+        else:
+            d = self.get_data(id)
+
         # Leave this check as d.mask is False since d.mask need not be a boolean
+        # and we want different errors if mask is True or False (and leave as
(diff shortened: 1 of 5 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -4352,24 +4352,32 @@ def save_filter(self, id, filename=None, bkg_id=None, ascii=True,
...
+
         if bkg_id is not None:
             d = self.get_bkg(id, bkg_id)
-        id = self._fix_id(id)
+        else:
+            d = self.get_data(id)
 
         # Leave this check as d.mask is False since d.mask need not be a boolean
+        # and we want different errors if mask is True or False (and leave as
+        # the iterable check to catch 'd.mask is True or any other value that
```
harlantc, line 4364:
```suggestion
        # the iterable check to catch 'd.mask' is True or any other value that
```

harlantc (approved):
This all looks good to me. 

I do have a question out of curiosity. I see you do `outfile = tmp_path / "stuff.file"` for the tests. I'm not familiar with this, is this a test-specific syntax or an overload?

DougBurke:
Tmp_path is described at https://docs.pytest.org/en/6.2.x/tmpdir.html and
is part of pytest

Issue #1235, opened 2021-07-28
XSPEC: should we bite the bullet and bump the chatter level to 10?


DougBurke:
# Sumary

We should seriously look at changing the XSPEC chatter level from 0 to 10 as it should help users when there's a problem. It will however add screen output when models get evaluated/

# Details

We have defaulted to an XSPEC chatter level of 0 for a long while, as it causes extra screen output when a model is first initialised (and possibly other times). However, rhe default level is 10 in XSPEC and perhaps we should change to this.

I say this because we've just dealt with the nth person who reported getting apec model results being 0, and this is because their ~/.xspec/Xspec.init file had an old ATOMDB_VERSION setting - so for CIAO 4.13 if you have

```bash
% grep ATOMDB ~/.xspec/Xspec.initATOMDB_VERSION:  3.0.7
```

which you can have from an old Xspec run, then you get

```
% sherpa
-----------------------------------------------------
Welcome to Sherpa: CXC's Modeling and Fitting Package
-----------------------------------------------------
Sherpa 4.13.0

Python 3.8.2 (default, Mar 25 2020, 17:03:02)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Using matplotlib backend: Qt5Agg

sherpa In [1]: xsapec.mdl()
Out[1]: <XSapec model instance 'xsapec.mdl'>

sherpa In [2]: mdl([0.1, 0.2, 0.3], [0.2, 0.3, 0.4])
Out[2]: array([0., 0., 0.])
```

Note the output is all 0's, which is wrong.

If you bump the chatter setting to 10 then you get some idea there's a problem:

```
% sherpa
-----------------------------------------------------
Welcome to Sherpa: CXC's Modeling and Fitting Package
-----------------------------------------------------
Sherpa 4.13.0

Python 3.8.2 (default, Mar 25 2020, 17:03:02)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Using matplotlib backend: Qt5Agg

sherpa In [1]: set_xschatter(10)

sherpa In [2]: xsapec.mdl()
Out[2]: <XSapec model instance 'xsapec.mdl'>

sherpa In [3]: mdl([0.1, 0.2, 0.3], [0.2, 0.3, 0.4])
[...]

PR #1254, opened 2021-08-13
FIX: #1235 change XSPEC chatter to 10 from 0

DougBurke:
# Summary

Change the default X-Spec chatter level from 0 to 10. This may result in screen output the first time an XSPEC model is evaluated (e.g. from a plot_model or fit call). There have been a number of cases where APEC models appeared to be creating no output because of missing data files (due to old ~/.xspec/Xspec.init settings) which would have been more obvious with this change. To get back to the previous behavior users can call `set_xschatter(0)`.

# Details

The chatter level is explicitly set to 10 rather than 0 (it appears
to be the default value but let's be explicit in case things change).
I had wondered whether it would be read from the ~/.xspec/Xspec.init
file but it doesn't appear so.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -50,6 +50,10 @@ are now taken from the user's XSPEC configuration file - either
 for these settings. The default value for the photo-ionization table
 in this case is now ``vern`` rather than ``bcmc``.
 
+The default chatter setting - used by models to inform users of
+issues - was set to 0 (which hid the messages) until Sherpa 4.14.0,
+when it was changed to 10 (to match XSPEC).
+
 Supported models
 ----------------
 
@@ -167,7 +171,7 @@ def get_xschatter():
     --------
 
     >>> get_xschatter()
-    0
+    10
     """
 
     return _xspec.get_xschatter()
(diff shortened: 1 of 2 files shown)


DougBurke:
I've marked this as for SDS review as I'd really like people other than me to try Sherpa sessions and a number of "complex" models (e.g xsapec , xstbabs, vertually any relatively-complex reflection/continuum models) and see if it is too distracting.

You don't need to check out this PR, you can just start your sherpa session with

    from sherpa.astro.xspec import set_xschatter
    set_xschatter(10)

and get the same behavior.

anetasie (approved):

wmclaugh:
It might be worth exploring updating the Xspec initialization in sherpa to use the 'log level chatter' value from the $HOME/.xspec/Xspec.init instead of changing the default from 0 to 10. This would allow users to not have to add explicit commands to continue to utilize the old less verbose behavior (ie #81).

DougBurke:
This would require us to parse that file ourselves, as I can's see any
obvious  for that in the XSPEC model interface. That is potentially risky
(as it ties us to our understanding of the file format which could change).

An alternative is for users to add their own config file to the ipython
directory to set it if they don't like the setting (or always use
set_xschatter(0) in the scripts)

Doug

PR #1258, opened 2021-08-17
added clause for xspec 12.12.0

grburgess:
The structure of XSPEC includes changed in the latest version.

This adds a switch to handle that.


DougBurke:
@grburgess can you check to see if this does anything different to #1223?

grburgess:
@DougBurke oops! sorry, I didn't see that. There is nothing different in this PR so I will close.

DougBurke:
Ta. If you have time to try out #1223 and check it works for you that would be helpful (as external validation).

PR #1259, opened 2021-08-17
XSPEC parameter changes - you can change the hard limits of most XSPEC parameters

### LINKED ISSUE #1202: XSPEC vs Sherpa soft and hard limits
XSPEC `soft` and `hard` limits are defined in the `newpar` function as

> The four arguments of the range specification determine the range of acceptable values for the parameter. The soft limits should include the range of expected parameter behavior. Between the hard and soft limits, the parameter is made stiffer to adjustment by the minimization routine invoked by the fit command. The parameter is never allowed to have a value at or outside the hard limits.

This is different to out use of `soft` and `hard` - we don't allow the parameter to be set outside of the `soft` limits apart from within certain fit conditions (which I do not know the details of) and we use the `soft` limits when working out how much the parameter can vary (e.g. in MCMC or in a `guess` call). I don't believe we have this "stiffening" feature within our optimization methods.

This can occasionally cause problems - e.g. see #1200 - and it's not obvious how to handle this. We could think about overloading the parameter behavior so that for XSPEC models we can handle some of the changes (e.g. setting a value outside the soft but within the hard limits) but it's not obvious there's a principled approach we can do.

Also note that even with the above XSPEC documentation, some XSPEC model parameters accept values outside the hard limits - these tend to have a hard minimum of 0 and if you set the parameter to a negative value then the parameter is interpreted in a certain way (and I think is assumed/required to be fixed). For example, `optxagnf` has a `fpl` (aka par#10) parameter which is described as (paraphrased fromm https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSmodelOptxagn.html )

> The fraction of the power below rcor which is emitted in the hard comptonisation component. If this parameter is negative then only the hard Compton component is used.

but the `model.dat` entry has `fpl` as >= 0:

```
optxagnf         11  0.03       1.e20           optxagnf  add  0
mass    solar  1e7     1.0     1.0     1.e9    1.e9     -.1
dist    Mpc    100    0.01    0.01    1.e9    1.e9     -.01
logLoLEdd " "   -1.    -10.      -10.       2    2     0.01
astar   " "     0.0    0.      0.      0.998   0.998    -1
rcor     rg    10.0   1.      1.      100.    100.     0.1
logrout " "   5.0   3.0    3.0       7.0     7.0      -1e-2
kT_e     keV  0.2   0.01    0.01     10      10        1e-2
tau      " "   10.  0.1    0.1      100      100       0.1
Gamma    " "  2.1   1.05     1.05     5.        10.       0.01
fpl     " "   1.e-4 0.0  0.0  1.   1.        1e-6
Redshift " "  0.    0.      0.      10.       10.         -0.01
```

DougBurke:
# Summary

XSPEC model parameters now use the same limits for soft and hard (the hard limit from the model.dat file). This is handled by the XSBaseParameter class which is used for XSPEC table models. The XSParameter class, which extends this and is used by most other parameters, allows the user to change the hard limits of the model. This follows XSPEC, and allows using some models which are documented as supporting a value outside the normal parameter range (normally by setting the value to a negative value). Note that this is potentially dangerous (it could crash the program) so should be used carefully. It is strongly suggested that any parameter set to a value outside of the original limits is also frozen.

# Details

Fixes #1202. I have investigated several different approaches (one of which you can see in this set of commits but has since been changed) and ended up with a similar system to XSPEC: we allow the hard limits to be changed and then it's up to the user to be careful.

To do this we have two classes: `XSBaseParameter` extends `Parameter` and `XSParameter` extends `XSBaseParameter`.

`XSBaseParameter`, has two changes:

1. It stores the original soft and hard limits in the `_xspec_soft/hard_min/max` attributes; this is mainly so we can track what the orignanl values were (for debugging/testing) but is currently unused
2. It uses the hard_min/max range for the soft limits when calling `Parameter.__init__`.

The `XSParameter` class

1. allows a user to set the `hard_min` and `hard_max` values. Any change to `hard_xxx` is copied to `xxx`.
2. over-rides the set method to allow `hard_min` and `hard_max` attributes.

It is interesting to see what tests fail thanks to these changes. Other than some internal changes (e.g. the addition of an XSxxx class we need to filter out of some checks) we can see changes in things like error analysis when the upper limit that was calculated is now outside the valid range for that parameter, and presumably subtle differences thanks to limit changes and how the optimizer deals with hitting the boundary.

One interesting choice is when to use Parameter, XSBaseParameter, or XSParameter. For table models I use XSBaseParameter since we know that it is dangerous for them to set a value outside the default hard range (see #743). For most other parameters we use XSParameter. The norm parameter (last one in additive models) is not technically an XSPEC parameter (i.e. we apply the parameter when evaluating the model) so we can just use the standard Parameter class for this.

# Notes

I had experimented with a version where we set the `hard_min` and `hard_max` values to `-hugeval` and `hugeval` but this caused a number of our "sample flux" tests to fail because we suddenly were trying to set nH to a value < 0 when sampling the distribution. So this version ended up being rejected.

The original implementation let a user set a value outside the hard limits, automatically freezing it. In some ways this is a nicer interface, as it's less to do, but the code to implement this started to fill up with special-case behavior that I felt wasn't worth the effort to implement, test, and - most-importantly - document to the user.

So I ended up with an XSPEC-like approach. I do not expect this feature to be used often so we do not need to spend time improvng the ergonomics.

# Example

The `fpl` parameter from [optxagn](https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSmodelOptxagn.html) is labelled as

> The fraction of the power below rcor which is emitted in the hard comptonisation component. If this parameter is negative then only the hard Compton component is used.

```python
>>> from sherpa.astro.xspec import get_xsversion
>>> get_xsversion()
'12.11.1d'
>>> xsoptxagn.mdl
<XSoptxagn model instance 'xsoptxagn.mdl'>
>>> print(mdl)
xsoptxagn.mdl
   mdl.mass     frozen        1e+07            1        1e+09      solar
   mdl.dist     frozen          100         0.01        1e+09        Mpc
   mdl.logLoLEdd thawed           -1          -10            2
   mdl.astar    frozen            0            0        0.998
   mdl.rcor     thawed           10            1          100         rg
   mdl.logrout  frozen            5            3            7
   mdl.kT_e     thawed          0.2         0.01           10        keV
   mdl.tau      thawed           10          0.1          100
   mdl.Gamma    thawed          2.1          0.5           10
   mdl.fpl      thawed       0.0001            0          0.1
   mdl.fcol     frozen          2.4            1            5
   mdl.tscat    frozen       100000        10000       100000
   mdl.Redshift frozen            0            0           10
>>> print(mdl.fpl)
val         = 0.0001
min         = 0.0
max         = 0.1
units       =
frozen      = False
link        = None
default_val = 0.0001
default_min = 0.0
default_max = 0.1
>>> mdl.fpl.min
0.0
>>> mdl.fpl.hard_min
0.0
```

But we can't set it negative!

```python
>>> mdl.fpl = -1
...
sherpa.utils.err.ParameterErr: parameter mdl.fpl has a minimum of 0
```

Now we can! The preferred style is to use the set method since you can change the minimum, value, and frozen flag in one call:

```python
>>> mdl.fpl.set(val=-1, hard_min=-1, frozen=True)
>>> print(mdl.fpl)
val         = -1.0
min         = -1.0
max         = 0.1
units       =
frozen      = True
link        = None
default_val = -1.0
default_min = 0.0
default_max = 0.1
>>> mdl.fpl.min
-1.0
>>> mdl.fpl.hard_min
-1.0
```

You can do it separately

```python
>>> mdl.fpl.hard_min = -1
>>> mdl.fpl = -1
>>> mdl.fpl.freeze()
```

Note that setting `hard_min` also sets `min` (and `hard_max` will set `max`) as a convenience and because we don't really have a soft vs hard distinction for XSPEC parameters like we do for Sherpa parameters.

Note that normal Sherpa parameters can not have their hard range changed - we can see this with the `norm` parameter of this model:

```python
>>> type(mdl.fpl)
<class 'sherpa.astro.xspec.XSParameter'>
>>> type(mdl.norm)
<class 'sherpa.models.parameter.Parameter'>
>>> mdl.norm.hard_min = -10
 ...
>>> mdl.norm.set(hard_min=-10)
...TypeError: set() got an unexpected keyword argument 'hard_min'
```

# Example

The serialization code used by `save_all` needed to be updated to recognize a model which has had it's hard-limit range changed. As an example

```python
>>> xsoptxagn.mfl
<XSoptxagn model instance 'xsoptxagn.mfl'>
>>> mfl.fpl.set(hard_min=-1, val=-1, frozen=True)
>>> save_all('test.shp')
```

This creates, in `test.shp`

```
mfl.fpl.hard_min    = -1.0
mfl.fpl.default_val = -1.0
mfl.fpl.default_min = 0.0
mfl.fpl.default_max = 0.1
mfl.fpl.val     = -1.0
mfl.fpl.min     = -1.0
mfl.fpl.max     = 0.1
mfl.fpl.units   = ""
mfl.fpl.frozen  = True
```

# Alternatives / Future work

Perhaps instead of tweaking `hard_min`/`hard_max` behavior I should tweak `min`/`max` (as they can be thought of being synonymous) so you'd just say

```python
>>> mdl.fpl.set(min=-1, val=-1)
```

and this would auto-magically shift the hard limits as well as the soft? I think this sounds sensible.

I'd like to add a `XSNormParameter` class (name to be bike-shedded) which is just for the `norm` parameter of additive models, as there are times it's useful to be able to determine which parameter "scales" the overall model. The XSAdditiveModel class should then add this automatically if not already set in the calling class. But this is all for a later PR.

Probably need to improve docs

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -535,38 +535,6 @@ def test_proj(run_thread, fix_xspec):
 @requires_data
 @requires_fits
 @requires_xspec
-@pytest.mark.parametrize('parallel', [False, True])
-def test_proj_bubble(parallel, run_thread, fix_xspec):
-
-    # fit_results is unused
-    def cmp_thread(fit_results, mek1, covarerr):
-
-        assert covarerr[0] == approx(0, rel=1e-4)
-        assert covarerr[1] == approx(8.74608e-07, rel=1e-3)
-        assert mek1.kt.val == approx(17.8849, rel=1e-2)
-        assert mek1.norm.val == approx(4.15418e-06, rel=1e-2)
-
-    check_thread(run_thread, 'proj_bubble', parallel, cmp_thread,
-                 ['mek1'])
-
-    # Proj -- Upper bound of kT can't be found
-    #
-    proj = ui.get_proj_results()
-    assert proj.parmins[0] == approx(-12.048069, rel=0.01)
-    assert proj.parmins[1] == approx(-9.510913e-07, rel=0.01)
-    assert proj.parmaxes[1] == approx(2.403640e-06, rel=0.01)
-    assert proj.parmaxes[0] is None
-
-    conf = ui.get_conf_results()
-    assert conf.parmins[0] == approx(-12.1073, rel=0.01)
-    assert conf.parmaxes[0] == approx(62.0585, rel=0.01)
-    assert conf.parmins[1] == approx(-9.5568e-07, rel=0.01)
-    assert conf.parmaxes[1] == approx(2.39937e-06, rel=0.01)
-
-
-@requires_data
-@requires_fits
-@requires_xspec
 def test_counts(run_thread, fix_xspec):
     tlocals = run_thread('counts')
 
@@ -776,13 +744,13 @@ def test_thread_pileup(run_thread):
 @requires_data
 @requires_fits
 @requires_xspec
-@pytest.mark.usefixtures("clean_astro_ui")
-def test_proj_bubble(run_thread):
+@pytest.mark.parametrize('parallel', [False, True])
+def test_proj_bubble(parallel, clean_astro_ui, fix_xspec, run_thread):
 
     # How sensitive are the results to the change from bcmc to vern
     # made in XSPEC 12.10.1? It looks like the mekal best-fit
     # temperature can jump from ~17.9 to 18.6, so require bcmc
-    # in this test.
+    # in this test (handled by fix_xspec).
     #
     # Note that the error on kT is large, so we can expect that
     # changes to the system could change these results. In particular,
@@ -790,20 +758,6 @@ def test_proj_bubble(run_thread):
     # analysis they are > 10 or even unbound, so it is likely that the
     # covariance error results can change significantly.
     #
-    xspec.set_xsxsect('bcmc')
-    models = run_thread('proj_bubble')
-
-    fit_results = ui.get_fit_results()
-    covarerr = sqrt(fit_results.extra_output['covar'].diagonal())
-    assert covarerr[0] == approx(0, rel=1e-4)
-    assert covarerr[1] == approx(8.74608e-07, rel=1e-2)
-
-    # Fit -- Results from reminimize
-    assert models['mek1'].kt.val == approx(17.8849, rel=1e-2)
-    assert models['mek1'].norm.val == approx(4.15418e-06, rel=1e-2)
-
-    # Fit -- Results from reminimize
-
     # The fit results change in XSPEC 12.10.0 since the mekal model
     # was changed (FORTRAN to C++). A 1% difference is used for the
     # parameter ranges from covar and proj (matches the tolerance for
@@ -811,6 +765,16 @@ def test_proj_bubble(run_thread):
     # 10% for the kT errors, as there is a significant change seen
     # with different XSPEC versions for the covariance results.
     #
+    # fit_results is unused
+    def cmp_thread(fit_results, mek1, covarerr):
+
+        assert covarerr[0] == approx(0, rel=1e-4)
+        assert covarerr[1] == approx(8.74608e-07, rel=1e-3)
+        assert mek1.kt.val == approx(17.8849, rel=1e-2)
+        assert mek1.norm.val == approx(4.15418e-06, rel=1e-2)
+
+    check_thread(run_thread, 'proj_bubble', parallel, cmp_thread,
+                 ['mek1'])
 
     # Covar
     #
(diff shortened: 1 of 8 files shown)


hamogu:
I prefer the @DougBurke  "idea of setting the Sherpa soft limits to the XSPEC hard limits and then the Sherpa hard limits to -big value to +big value, with the understanding that the user is now responsible if they go outside the Sherpa soft-limit range" over the code that's currently in here; maybe with a modified error message for setting parameters outside the soft limit. You could override `_set_val` with a message like "[old message here]. Some XSPEC models use out-of-range parameters as flags for some special condition; to enable this functionality do this first: param.min=-1e38, then reset and freeze the parameter value. Note that param.min should be reset to the current values param.min=0.1234 if this parameter is used for fitting again later." (OK, I know the text is long and not good. Maybe different wording could help.)

anetasie:
@hamogu and @DougBurke agreed with setting Sherpa soft to XSPEC hard, and Sherpa hard to gig values.

DougBurke:
Dang it - switching to a model which allows the hard limits to ~ - 3e28 to ~ 3e34 messes up a bunch of our flux simulation tests (as we end up trying to set nh < 0 which doesn't end well). Blargh.

DougBurke:
This has been rebased because of #1261 and re-worked. Unfortunately my great idea of just setting the hard limits to -hugeval to hugeval breaks some existing code (the sample flux tests show this) so I realised we could just copy XSPEC and allow the user to change the hard limits of the parameter. At least for non-XSPEC-table-model parameters. This then clearly defines a point where the user has entered dragon-infested waters, rather than trying to be clever.

hamogu (approved):

DougBurke:
@anetasie - I'd like you to look at this again in its final form to make sure you agree / think about documentaton / ...

anetasie (approved):
I approve the PR. @DougBurke I have a test case for when this is merged to CIAOX and I can try it out.

Issue #52, opened 2015-06-24
XSPEC module is specific to a single patch / should autogenerate the interface code


DougBurke:
I believe that a lot of the XSPEC module code - i.e. most of `_xspec.cc` and `__init__.py` - should be auto-generated based on the contents of the `model.dat` file in the XSPEC distribution. The main advantage to this approach is that it makes it possible for users to compile against the XSPEC module that they are using. That is, handle changes to the `model.dat` file that includes:

- parameter changes (e.g. a change to the min/max range)
- changes to the C/C++/Fortran model function being called (e.g. if the model has been re-written and
      now uses a different function)
- new models
- a more-consistent naming scheme for parameter names that don't map to Python (there is no
  existing scheme, as far as I can tell, and it looks like different rules were used for different models)
- removal of old models (not sure this has happened, but it could)

At present, users are forced to use an interface that assumes a single patch level - at present it's 12.8.2e - which should work for any 12.8.2 build, but leaves the user in a slightly-strange place if used against a different patch level (i.e. what bugs have/have not been fixed). If there's a 12.8.3 with new models then they will not pick them up until the code has been manually edited (which we have found to be problematic), and then there may be problems compiling the module against an older version of XSPEC.
 
There are some down sides to auto-generating most of the code - such as documentation, complicating the build process, and tracking down bugs (although I'm less bothered about the last point since the code that needs to be generated is quite simple).

I'd also like to expand access to the "utility" functions in XSPEC (in particular the keyword-related ones that *might* allow us to support some of the "mixing" models), which is related (some of them can be auto-generated, if there's an easy way to check for the functionality being present), but isn't an essential part of this request.

I plan to work on this, and have existing code that I can use as a starting point, but that's going to take a bit of time to put together.

DougBurke:
Ahhh, there's a new release of XSPEC being tested, so I guess this is more relevant that I realised this morning!

anetasie:
Yes. The new patch was just released and include the updates to [NEI models](https://heasarc.gsfc.nasa.gov/xanadu/xspec/issues/issues.html)

DougBurke:
Aneta, I was talking about the next release, which is in the testing stage but not-yet released, not the latest patch.

DougBurke:
Note that Xspec 12.9.0 has now been released. It contains one new model, a few changes to the API functions (none relevant to the functions we currently use), and changes in the parameter names (e.g. see #74) and values.

DougBurke:
[...]

PR #1261, opened 2021-08-17
update to match recent XSPEC model.dat settings

DougBurke:
# Summary

Update the XSPEC models to match the latest model.dat file (from HEASOFT 6.29 / XSPEC 12.12.0). This primarily changes the hard limits of the models, but there are some changes to the default parameter values as well, as well as some unit changes, including adding and removing units.

# Details

Background: Ideally I'd like to auto-generate the XSPEC interface (see #52) and this can be seen as part of this work (i.e. making our interface as close to an auto-generated one as we can get).

Background: although I am using the heasoft-6.29 `model.dat` file you eget very-similar results if you use an XSPEC 12.11.1 file.

The first commit is technically un-nescessary but moves us closer to this auto-generated ideal: it moves the model class definitions to match the `model.dat` order, which just makes it easier when manually scanning through the two files.

There are three reasons for our code having different parameter values to the `model.dat` file:

- errors when creating the model classes (e.g. cut-n-paste issues) which will be greatly reduced once we merge #1260
- the model.dat file has changed over time and we haven't noticed this (we can now easily check this once #1260 is merged)
- our understanding of the XSPEC "hard limit" has been different to XSPEC, so we often have completely ignored the hard-limit ranges and just created our own ranges (rather randomly). This is a problem as described in #1202 and is partly addressed in #1259 (we need this to "complete" this fix).

Note that one thing I have not changed are the nH units for many parameters - they are given as `10^22` in the `model.dat` file but we label them `10^22 atoms / cm^2`. Ideally I'd like to make this change but it impacts a number of our serialization tests and would be a large problem for the SDS regression suite. As there are a lot of Sherpa changes going in at this time I have elected not to make this particular change at this time. 

# Tests

It is interesting to see which of our tests change/fail when you make changes like this (i.e. those that are "sensitive" to some part of the system, such as the allowed range of a parameter). I note that sherpa/astro/tests/test_astro.py::test_proj_bubble in particular has a best fit kT ~ 17 and an error range from covariance < 1 but from other methods it's >~ 10 or even unbounded. So I don't think it's surprising when we can see a change like the covariance kT error going from ~0.35 to ~0.65.

# How did I make these changes?
 
I took a build with #1260 and then I could say

```
% ./scripts/check_xspec_update.py ~/local/heasoft-6.29/spectral/manager/model.dat
Model agnslim
  [ 1] par logmdot default: -1.0 -> 1.0
  [ 2] par astar softmin: -1.0 -> 0.0
  [ 3] par astar hardmin: -1.0 -> 0.0

Model apec
  [ 1] par kT hardmin: 0.0 -> 0.008
  [ 2] par kT hardmax: 3.4028234663852886e+38 -> 64.0
  [ 3] par Abundanc hardmax: 3.4028234663852886e+38 -> 5.0
  [ 4] par 3 name: redshift -> Redshift
  [ 5] par redshift hardmax: 3.4028234663852886e+38 -> 10.0

Model bapec
  [ 1] par kT hardmin: 0.0 -> 0.008
  [ 2] par kT hardmax: 3.4028234663852886e+38 -> 64.0
  [ 3] par Abundanc hardmax: 3.4028234663852886e+38 -> 5.0
  [ 4] par Redshift hardmax: 3.4028234663852886e+38 -> 10.0
  [ 5] par Velocity hardmax: 3.4028234663852886e+38 -> 1000000.0

Model bbody
  [ 1] par kT hardmin: 0.0 -> 0.0001
  [ 2] par kT hardmax: 3.4028234663852886e+38 -> 200.0
  [ 3] par norm units: L39 / (D10)**2 ->

Model bbodyrad
  [ 1] par kT hardmin: 0.0 -> 0.0001
  [ 2] par kT hardmax: 3.4028234663852886e+38 -> 200.0

Model bexrav
...
```

I can then update the XSPEC `__init__.py` file and re-check.

### DIFF
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -784,6 +784,12 @@ def test_proj_bubble(run_thread):
     # temperature can jump from ~17.9 to 18.6, so require bcmc
     # in this test.
     #
+    # Note that the error on kT is large, so we can expect that
+    # changes to the system could change these results. In particular,
+    # the covariance errors on kt are < 1 but from other error
+    # analysis they are > 10 or even unbound, so it is likely that the
+    # covariance error results can change significantly.
+    #
     xspec.set_xsxsect('bcmc')
     models = run_thread('proj_bubble')
 
@@ -810,10 +816,10 @@ def test_proj_bubble(run_thread):
     #
     # TODO: should this check that parmaxes is -1 * parmins instead?
     covar = ui.get_covar_results()
-    assert covar.parmins[0] == approx(-0.328832, rel=0.1)
-    assert covar.parmins[1] == approx(-8.847916e-7, rel=0.01)
-    assert covar.parmaxes[0] == approx(0.328832, rel=0.1)
-    assert covar.parmaxes[1] == approx(8.847916e-7, rel=0.01)
+    assert covar.parmins[0] == approx(-0.653884, rel=0.1)
+    assert covar.parmins[1] == approx(-8.94436e-07, rel=0.01)
+    assert covar.parmaxes[0] == approx(0.653884, rel=0.1)
+    assert covar.parmaxes[1] == approx(8.94436e-07, rel=0.01)
 
     # Proj -- Upper bound of kT can't be found
     #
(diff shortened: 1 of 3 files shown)


DougBurke:
@anetasie - do you have any comments on this? Note that I think this really needs to be followed by #1259 so that we can use a value outside the hard limits (because of how XSPEC works).

PR #1266, opened 2021-08-26
Splitup model expressions

DougBurke:
Is this useful functionality (it was originally built to allow a `plot_model_components` command to be able to display the individual components without requiring the user to manually create the expressions)?

It is a relatively large code change for what feels like a small user-level change.

What does it mean for future "fancy" model classes?

# Summary

Allow a model expression to be split up into separate "additive" terms (based on the XSPEC nomenclature).

# Details

The aim is to be able to separate out a model expression into individual meaningful components (in XSPEC terminology a set of multiplicative models applied to an additive model). These expressions can then be used downstream - e.g. to allow a user to plot up all the individual components of a model in a "meaningful" way. The implementation does **not** use the XSPEC additive and multiplicative classes, since this is meant to be a generic solution. Instead it tries to expand the binary operations we support to change an expression like

    a * (b + c)

to

    a * b + a * c

# Notes

The `expand` and `separate` methods are added to the classes and also there's functions with these names which essentiallly just call the underlying methods.

# TODO

Need to look at instrument-style responses (e.g. convolution/RMF, multiplication/ARF, and non-linear/pileup)  to decide if you want to expand the contents - i.e. when you have

    resp(1234.56 * a * (b + x))

should it come back with

    resp(1234.56 * a * b + 1234.56 * a * x)

or not bother (since we can't guarantee that

    resp(a + b) == resp(a) + resp(b)

[we know it isn't the case for pileup]

# Examples

With the following setup to create four model components (in real life they'd have different models but that's not relevant here):

```
>>> from sherpa.models.basic import Const1D
>>> a = Const1D('a'); b = Const1D('b'); c = Const1D('c'); d = Const1D('d')
```

A very-simple expression. Note that it does not `expand` (i.e. the call returns the input`) but we can `separate` the additive terms:

```
>>> mdl = a + (b * c - d * a)
>>> mdl
<BinaryOpModel model instance '(a + ((b * c) - (d * a)))'>
>>> mdl.name
'(a + ((b * c) - (d * a)))'
>>> mdl.expand()
<BinaryOpModel model instance '(a + ((b * c) - (d * a)))'>
>>> mdl.separate()
[<Const1D model instance 'a'>, <BinaryOpModel model instance '(b * c)'>, <UnaryOpModel model instance '-((d * a))'>]
>>> for expr in mdl.separate():
...     print(expr.name)
...
a
(b * c)
-((d * a))
```

Here's the expression I meant to write, which we can see does `expand`, which you can then `separate` to get the additive terms (but only after expansion):

```
>>> mdl2 = a * (b * c - d * a)
>>> mdl2.name
'(a * ((b * c) - (d * a)))'
>>> mdl2.expand()
<BinaryOpModel model instance '((a * (b * c)) - (a * (d * a)))'>
>>> mdl2.expand().name
'((a * (b * c)) - (a * (d * a)))'
>>> mdl2.separate()  # NOTE: this doesn't do anything, you need to expand first
[<BinaryOpModel model instance '(a * ((b * c) - (d * a)))'>]
>>> mdl2.expand().separate()
[<BinaryOpModel model instance '(a * (b * c))'>, <UnaryOpModel model instance '-((a * (d * a)))'>]
>>> for expr in mdl2.expand().separate():
...     print(expr.name)
...
(a * (b * c))
-((a * (d * a)))
```

Here's some more:

```
>>> m1 = a + b
>>> m2 = c + d
>>> (m1 * m2)
<BinaryOpModel model instance '((a + b) * (c + d))'>
>>> (m1 * m2).expand()
<BinaryOpModel model instance '((((a * c) + (b * c)) + (a * d)) + (b * d))'>
>>> for cpt in (m1 * m2).expand().separate():
...     print(cpt.name)
...
(a * c)
(b * c)
(a * d)
(b * d)
>>> m1 = a - b
>>> (m1 * m2).expand()
<BinaryOpModel model instance '((((a * c) - (b * c)) + (a * d)) + (b * d))'>
>>> for cpt in (m1 * m2).expand().separate():
...     print(cpt.name)
...
(a * c)
-((b * c))
(a * d)
(b * d)
>>> (m1 * m2) - (d - c)
<BinaryOpModel model instance '(((a - b) * (c + d)) - (d - c))'>
>>> ((m1 * m2) - (d - c)).expand()
<BinaryOpModel model instance '(((((a * c) - (b * c)) + (a * d)) + (b * d)) - (d - c))'>
>>> for cpt in ((m1 * m2) - (d - c)).expand().separate():
...     print(cpt.name)
...
(a * c)
-((b * c))
(a * d)
(b * d)
-(d)
-(-(c))
```

Here's a more-realistic example, showing that we need to handle responses:

```
>>> set_source(xsphabs.gal * (powlaw1d.pl + xsapec.blob))
>>> get_source().name
'(xsphabs.gal * (powlaw1d.pl + xsapec.blob))'
>>> get_source().expand().name
'((xsphabs.gal * powlaw1d.pl) + (xsphabs.gal * xsapec.blob))'
>>> for cpt in get_source().expand().separate():
...     print(cpt)
...
(xsphabs.gal * powlaw1d.pl)
   gal.nH       thawed            1            0        1e+06 10^22 atoms / cm^2
   pl.gamma     thawed            1          -10           10
   pl.ref       frozen            1 -3.40282e+38  3.40282e+38
   pl.ampl      thawed            1            0  3.40282e+38
(xsphabs.gal * xsapec.blob)
   gal.nH       thawed            1            0        1e+06 10^22 atoms / cm^2
   blob.kT      thawed            1        0.008           64        keV
   blob.Abundanc frozen            1            0            5
   blob.redshift frozen            0       -0.999           10
   blob.norm    thawed            1            0        1e+24
>>> get_model().name
'apply_rmf(apply_arf((38564.608926889 * (xsphabs.gal * (powlaw1d.pl + xsapec.blob)))))'
>>> get_model().expand().name
  File "/home/dburke/sherpa/sherpa-master/sherpa/models/model.py", line 946, in expand
    raise NotImplementedError()
NotImplementedError
```

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -936,6 +936,15 @@ class CompositeModel(Model):
 
         raise NotImplementedError()
 
+    def expand(self):
+        """Expand the model expression to separate out "additive" terms.
+
+        Note that it will re-use model components rather than
+        re-creating them.
+        """
+
+        raise NotImplementedError()
+
 
 class SimulFitModel(CompositeModel):
     """Store multiple models.
@@ -1266,6 +1275,45 @@ class UnaryOpModel(CompositeModel, ArithmeticModel):
 
         return self.__class__(args[0], self.op, self.opstr)
 
+    def expand(self):
+        """Expand the model expression to separate out "additive" terms.
+
+        Returns
+        -------
+        expanded : UnaryOpModel instance
+        """
+
+        try:
+            expanded = self.arg.expand()
+        except AttributeError:
+            return self
+
+        return self.create(expanded)
+
+    def separate(self):
+        """Separate out the 'additive' terms in the model expression.
+
+        Returns
+        -------
+        models : list of Model instances
+            The additive terms of the model. Summing the elements will
+            return the equivalent model expression to that stored in the
+            object.
+
+        Notes
+        -----
+        We separate out the components and then apply the operator to
+        each term.
+
+        """
+
+        try:
+            out = self.arg.separate()
+        except AttributeError:
+            return [self]
+
+        return [self.create(m) for m in out]
+
 
 class BinaryOpModel(CompositeModel, RegriddableModel):
     """Combine two model expressions.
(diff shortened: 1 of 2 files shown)


DougBurke:
Rebased just to match the latest changes.

DougBurke:
The code in

- #1679 

(my latest attempt at adding a `plot_model_components` command) is signifcantly simpler than this so I'm going to close this down as it's not clear the extra complexity is worthwhile.

PR #1268, opened 2021-08-28
Doc: switch default sphinx role

hamogu:
## Summary
Multiply the number of working references on readthedocs, while at the same time reduce verbosity in the docstrings.

## Details
I proposal to switch the sphinx "default role" (the rules that sphinx uses to for \`text like this\`) to "obj". Before, text in single backticks would just be rendered in a different font and only text marked up with an explicit sphinx role (e.g. `:py:func:`) would generate a link. Given the multiple uses of the docstring (source for sphinx, ahelp, and read directly in e.g. ipython with tab completion), that extra role markup can be distracting and we've used it very sparingly in the past. On the other hand, Sherpa has many specialized classes and it's very convenient to just click on "sherpa.models.model.SimulFitModel" to see what that class does and how to initialize it instead of reaching though all of readthedocs to find the documenation of that class.
Fortunately, sphinx offers a  way around that dilemma: The "default role" (https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html). I've set this to "obj", but "any" could be an alternative.

Also improving the intersphinx and xref numpy settings means that a lot more words will be linked automatically, further reducing the need for backticks.

In particular,

- in parameter descriptions, certain words (including "None") will be recognized even without backticks.
- numpy xref brings in a list of more words, most importantly "array", which always means " numpy array" for use and is sometimes even written as "numpy array instance" - a very long way to say "array". Now, the link will make clear what type of array it is. For people investigating the doctring in Python itself, it should be clear that an array is a numy array (and in fact in many cases a dask or pnumpy array might work, too).

While working on the docs, I'm also starting to do the following clean-up:

- remove excess use of "instance" as in "a sherpa.models.parameter.Parameter instance". Everything in Python is an instance and we don't say "float instance" either. If the parameter to a function really did not require an instance, but the class itself, then the type would be "class" or "type".

## Status

I've (mostly) fixed-up one particular file as an example to show the benefits (will link to old and new version of one or more html help files, once sphinx as run and rendered them). 

## Plan
I'm asking @DougBurke for a review in this stage to check that (a) this is not a stupid idea  in general and (b) this would not cause undue burden when generating ahelp files. (In that case, I can probably help wit ha custom sphinx extension).

If we agree to merge something like this, I will extend this PR to fix up more of the doc strings, in particular all relating to fitting. However, we have so many open PRs, that it's not a good strategy to make one giant PR of everything because that would lead to too many merge conflicts as this or other PRS are mergers and all remaining PRS need rebasing. Instead, I would go through module by module and open separate PRS that can be reviewed and merged on a much shorter timescale.

Again, note that having unresolved links for the "any" role only gives us more sphinx warnings, but does not break the look of the documentation or make the CI fail.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,10 @@ import os
 import shutil
 import sys
 
+from sphinx_astropy.conf.v1 import intersphinx_mapping, default_role
+
 import sphinx_rtd_theme
+
 
 # Based on http://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
 #
@@ -146,11 +149,16 @@ extensions = [
     # of warning messages (about missing links) that I don't have time
     # to investigate.
     'sphinx.ext.napoleon',
-    # 'numpydoc.numpydoc'
+    # 'numpydoc.numpydoc',
+    'sphinx.ext.intersphinx',
+    'sphinx_astropy.ext.intersphinx_toggle',
     'sphinx_astropy.ext.edit_on_github',
     # notebooks
     'nbsphinx'
 ]
+# Imported from sphinx_astropy so we don't have to maintain the list
+# of servers
+# intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
 
 # notebook support
 # - for now never execute a notebook
@@ -313,8 +321,11 @@ exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
-#
-# default_role = None
+# Imported from sphinx_astropy
+# default_role = 'obj'
+
+# Setting from sphinx_astropy for numpydoc xref settings
+numpydoc_xref_param_type = True
 
 # If true, '()' will be appended to :func: etc. cross-reference text.
 #
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1339,7 +1339,7 @@ class Fit(NoNewAttributesAfterInit):
         `sherpa.utils.err.EstErr`
            If any parameter in parlist is not valid (i.e. is not
            thawed or is not a member of the model expression being
-           fit), or if the statistic is `~sherpa.optmethods.LeastSq`,
+           fit), or if the statistic is `~sherpa.stats.LeastSq`,
            or if the reduced chi-square value of the current parameter
            values is larger than the ``max_rstat`` option (for
            chi-square statistics).
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -227,7 +227,7 @@ class FitResults(NoNewAttributesAfterInit):
        statistic value, or a larger value, if the assumed model is
        true and the current model parameters are the true parameter
        values. This will be `None` if the value can not be calculated
-       with the current statistic (e.g.  the Cash statistic).
+       with the current statistic (e.g. the Cash statistic).
     rstat : number or None
        The reduced statistic value (the `statval` field divided by
        `dof`). This is not calculated for all statistics.
@@ -1006,11 +1006,12 @@ class Fit(NoNewAttributesAfterInit):
     stat : `sherpa.stats.Stat` or `None`, optional
        The statistic object to use. If not given then
        `Chi2Gehrels` is used.
-    method : sherpa.optmethods.OptMethod instance or None, optional
-       The optimiser to use. If not given then `LevMar` is used.
-    estmethod : sherpa.estmethod.EstMethod instance or None, optional
+    method : `sherpa.optmethods.OptMethod` instance or None, optional
+       The optimiser to use. If not given then `sherpa.optmethods.LevMar`
+       is used.
+    estmethod : `sherpa.estmethods.EstMethod` or None, optional
        The class used to calculate errors. If not given then
-       `Covariance` is used.
+       `sherpa.estmethods.Covariance` is used.
     itermethod_opts : dict or None, optional
        If set, defines the iterated-fit method and options to use.
        It is passed through to `IterFit`.
@@ -1115,7 +1116,7 @@ class Fit(NoNewAttributesAfterInit):
 
         See Also
         --------
-        `calc_chisqr`, `calc_stat_info`
+        calc_chisqr, calc_stat_info
 
         """
 
@@ -1287,7 +1288,7 @@ class Fit(NoNewAttributesAfterInit):
 
         Parameters
         ----------
-        *others : :py:class:`sherpa.fit.Fit` instances
+        *others : `sherpa.fit.Fit` instances
             The ``data`` and ``model`` attributes of these arguments
             are used, along with those from the object.
 
@@ -1324,7 +1325,7 @@ class Fit(NoNewAttributesAfterInit):
             the associated optimisation method instance to use. This
             is only used if the method is changed, as described in
             the Notes section below.
-        parlist : sequence of `sherpa.model.parameter.Parameter` or None, optional
+        parlist : sequence of `sherpa.models.parameter.Parameter` or None, optional
             The names of the parameters for which the errors should
             be calculated. If set to `None` then all the thawed
             parameters are used.
@@ -1338,7 +1339,7 @@ class Fit(NoNewAttributesAfterInit):
         `sherpa.utils.err.EstErr`
            If any parameter in parlist is not valid (i.e. is not
            thawed or is not a member of the model expression being
-           fit), or if the statistic is `LeastSq`,
+           fit), or if the statistic is `~sherpa.optmethods.LeastSq`,
            or if the reduced chi-square value of the current parameter
            values is larger than the ``max_rstat`` option (for
            chi-square statistics).
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -70,7 +70,7 @@ class StatInfoResults(NoNewAttributesAfterInit):
     ids : sequence of int or str
        The data set ids (it may be a tuple or array) included in the
        results.
-    bkg_ids: sequence of int or str, or `None`
+    bkg_ids: sequence of int or str, or None
        The background data set ids (it may be a tuple or array)
        included in the results, if any.
     statname : str
@@ -82,16 +82,15 @@ class StatInfoResults(NoNewAttributesAfterInit):
     dof: int
        The number of degrees of freedom in the fit (the number of
        bins minus the number of free parameters).
-    qval: number or `None`
+    qval: number or None
        The Q-value (probability) that one would observe the reduced
        statistic value, or a larger value, if the assumed model is
        true and the current model parameters are the true parameter
        values. This will be `None` if the value can not be calculated
        with the current statistic (e.g. the Cash statistic).
-    rstat: number of `None`
+    rstat: number or None
        The reduced statistic value (the `statval` field divided by
        `dof`). This is not calculated for all statistics.
-
     """
 
     # The fields to include in the __str__ output.
@@ -158,8 +157,8 @@ def _cleanup_chi2_name(stat, data):
 
     Parameters
     ----------
-    stat : a sherpa.stats.Stat instance
-    data : a sherpa.data.Data or sherpa.data.DataSimulFit instance
+    stat : `sherpa.stats.Stat`
+    data : `sherpa.data.Data` or `sherpa.data.DataSimulFit`
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


hamogu:
As an example, compare the existing https://sherpa.readthedocs.io/en/latest/fit/api/sherpa.fit.Fit.html with the new https://sherpa--1268.org.readthedocs.build/en/1268/fit/api/sherpa.fit.Fit.html
Notice how the later has a lot more working links while it at the same time removes all ":py:class:" etc. from the docstrings, only a few unobtrusive backticks have been added.

DougBurke:
So, presumably the plan of attack is to take a bunch of watnings like

```
/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py:docstring of sherpa.ui.utils.Session.unpack_data:74: WARNING: 'any' reference target not found: colkeys
```

and then change the corresponding line, which I believe is

```
  3665          If the `colkeys` argument is used then a case-sensitive
  3666          match is used to determine what columns to return.
```

to say "``colkeys``" instead.

DougBurke:
Aside:

In general I am burnt out/jaded by my previous attempts to get the sphinx documentation started as nothing ever quite seemed to work right, in part because it looks like I was missing this setting. So I'm glad someone else is looking at it!

hamogu:
Yes, that is the plan of attack. > 50% of the warning can be fixed by looking for the string "None" and adding extra backticks. It's a not difficult at all, just a little tedious, so I did not want to go through all files unless we agree that we want to go ahead with that. (And even then I'll do one module at a time, e.g. not touch the grouping anf filtering functions until those PRs are in) so we don't waste our time rebasing stuff over and over again.

sherpa/fit.py
```diff
@@ -1132,16 +1131,15 @@ def calc_chisqr(self):
 
         Returns
         -------
-        chisq : array or `None`
+        chisq : array or ``None``
            The chi-square value for each bin of the data, using the
-           current statistic (as set by `set_stat`).  A value of
-           `None` is returned if the statistic is not a chi-square
+           current statistic (as set by `sherpa.astro.ui.set_statset_stat`).
+           A value of ``None`` is returned if the statistic is not a chi-square
```
DougBurke, line 1137:
```suggestion
           current statistic. A value of ``None`` is returned if the
           statistic is not a chi-square
```

So, this is at a much "lower" level then sherpa.astro.ui/sherpa.ui, so my  original inclusion of the text mentioning `set_stat` is wrong and should be removed (which also fixes the `set_statset_stat` duplication here).

sherpa/fit.py
```diff
@@ -1118,7 +1117,7 @@ def calc_stat(self):
 
         See Also
         --------
-        calc_chisqr, calc_stat_info
+        `calc_chisqr`, `calc_stat_info`
```
DougBurke, line 1120: Do we really have to add these now? You can see that these used to generate links, without the backticks - e.g. https://sherpa.readthedocs.io/en/latest/fit/api/sherpa.fit.Fit.html#sherpa.fit.Fit.calc_stat and I'm just hoping we can save some time

DougBurke (changes requested):
I've marked one required change and a question, but over-all LGTM and I agree with the approach of "fixing warnings" in a bunch of small commits rather than having a flag day

sherpa/fit.py
hamogu: No, we don't. Thanks for pointing that out. In fact, we can remove even more of the backticks in other locations.

hamogu:
@DougBurke I updated the details on the top. Additional setting reduce backticks even further; I'm now working to the smallest possible number of backticks. In some cases, we might even want more again (`None` instead of None, although both will render a link).
At this stage, I'm asking you to look at the rendered docs again for my example file, and, if you want to, more widely. 

If you agree to this type of layout, then I'll add a few more files to this PR, but as I said above, I do not want to do this sherpa-wide, but piece by piece to avoid merge conflicts in modules that otherwise have a lot of work done on them right now. This makes the docs look better, but it not an absolute necessity in any particular location.

docs/conf.py
```diff
@@ -313,8 +321,11 @@ def getmtime(infile):
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
-#
-# default_role = None
+# Imported from sphinx_astropy
+# default_role = 'obj'
```
DougBurke, line 325: The PR describes as you setting this value, but it's commented out here. I'm confused now!

hamogu: I'm indirectly setting it by importing it from sphinx_astropy: https://github.com/sherpa/sherpa/pull/1268/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R35

I'm doing that because I think there is more that we can import in general (e.g. the graviz-dot settings), so we don't have to maintain the settings should they change, but I also don't just want to import everything at this point (that messes up our html style).

DougBurke: Aha. The `v1` on that import worries me slightly - is there already a `v2` (in other words are we tying ourself to code that we willhave to change once astropy decides `v1` is so old it can be removed)?

sherpa/fit.py
```diff
@@ -1000,23 +998,22 @@ class Fit(NoNewAttributesAfterInit):
...
-       (i.e. be a SimulFitModel object when data is a DataSimulFit).
-    stat : sherpa.stats.Stat instance or None, optional
+    model : `sherpa.models.model.Model` or `sherpa.models.model.SimulFitModel`
+       The model to fit to the data. It should match the ``data`` parameter
+       (i.e. be a `SimulFitModel` object when data is a `DataSimulFit`).
+    stat : `sherpa.stats.Stat` or `None`, optional
        The statistic object to use. If not given then
-       :py:class:`~sherpa.stats.Chi2Gehrels` is used.
+       `Chi2Gehrels` is used.
     method : sherpa.optmethods.OptMethod instance or None, optional
```
DougBurke, line 1009: Looking at https://sherpa--1268.org.readthedocs.build/en/1268/fit/api/sherpa.fit.Fit.html I was surprised that the :look" changed at the `method` parameter from having links to just italicised text, but then I realized that you haven't added the "``` around the `sherpa.optmethods.xxx`, `None`, and `sherpa.estmethod.xxx` items

sherpa/fit.py
```diff
@@ -1000,23 +998,22 @@ class Fit(NoNewAttributesAfterInit):
...
-       The optimiser to use. If not given then
-       :py:class:`~sherpa.optmethods.LevMar` is used.
+       The optimiser to use. If not given then `LevMar` is used.
```
DougBurke, line 1010: Although `LevMar` is now rendered as a "clickable link" at https://sherpa--1268.org.readthedocs.build/en/1268/fit/api/sherpa.fit.Fit.html it doesn't actually work - ie it doesn't link to anything. Actually, I note that this holds for other terms that are used in this section that are not defined in this module (e.g. `Chi2Gehrels`, `Covariance`, `SimulFitModel`). I think you'd need to give the full path - e.g. "`sherpa.optmethods.LevMar`" .

docs/conf.py
hamogu: There is no `v2` yet (and there might never be). The whole point of having this versioned is that `v1` can stay around forever and it's up to the projects that use it to consciously switch to `v2` should it ever come. Astropy uses a similar scheme in other area (e.g. your choice of cosmological parameters) and to my knowledge none of the old version has ever been removed since the inception of the project.

sherpa/fit.py
hamogu: "None" is in fact one of those special words that is picked up for a link even without any formatting. Even though it's formattted like normal text (italics here), it is a link. We can have is "special" by adding backticks. The question is: Do we want that?
That's something I'd like to hear about in this review.

As for the `sherpa.optmethod.xxx`, I simple missed those because they were not links before (I searched for ":py:" to spot the locations that I needed to change), but of course they should become links now! I think the most efficient way to tackle this PR is to go through a bit more carefully once we've decided on the remaining open questions (see above: How should NONE be formatted?). However, we are inconsistent now, and if I miss a few instances tat could be turned into links right now, that OK. This PR will be an improvement in any case and we can fix anytihng I missed at a later point or next time we work on that docstring anyway.

sherpa/fit.py
hamogu: Yeah, I was experimenting. I can, however, do "~sherpa.optmethods.LevMar" , which will render as "LevMar" in the html, but still show the full link in the docstring itself obviously.
Which version do we prefer?

(Eventually we can turn sphinx warnings into errors which will catch cases like that, but for now we have too many other warnings.)

docs/conf.py
DougBurke: Ta - just wanted to know what the strategy was.

sherpa/fit.py
DougBurke: So, "You need to `~sherpa.frobnicate.bar` to get the answer" is a lot less messy than the old "You need to :py:class:`~sherpa.frobnicate.bar` to get the answer" so I think in general we should go for the linkable version (so `~sherpa.optmethods.LevMar`), but I don't think it's a hard-and-fast rule. Does this seem workable?

Out of interest, is it

    ~sherpa.optmethods.LevMar
    `~sherpa.optmethods.LevMar`

ie do you need the back-ticks?

sherpa/fit.py
DougBurke: In this PR we (aka you ;-) do **not** have to fix every issue in fit.py (or whatever you want to use as the example) to make it worth while to merge.

DougBurke: For the `None` vs None vs .. question, what does astropy do? Or are we different enough it's not worth aiming for consistency on the small issues like this?

sherpa/fit.py
hamogu: Yes, you do. (Exept for some very special cases, e.g. the None in a parameter, or a list of numpy specifics (like "array-like"). I think some of those speical rules also work better with numpydoc, which is like slightly better than napoleon for it's formatting, but that's a different issue.

sherpa/fit.py
hamogu: I think we are different enough. For one, they use numpydoc instead of napoleon and their own doc theme.

hamogu:
I think this is enough for this PR, so @DougBurke can review again when he's feeling ready.

Note for future: When we change back to numpydoc (currently using napoleon, but I hope that we can do that in the future), more functionality will be enabled, most notably "array_like" will link to the numpy docs for what formats can be interpreted as arrays. This requires no further changes to the docstring; this is why I left all the "array_like" in there and did not repace them with e.g. numpy.ndarray.

sherpa/fit.py
```diff
@@ -1338,7 +1339,7 @@ def est_errors(self, methoddict=None, parlist=None):
         `sherpa.utils.err.EstErr`
            If any parameter in parlist is not valid (i.e. is not
            thawed or is not a member of the model expression being
-           fit), or if the statistic is `LeastSq`,
+           fit), or if the statistic is `~sherpa.optmethods.LeastSq`,
```
DougBurke, line 1342:
```suggestion
           fit), or if the statistic is `~sherpa.stats.LeastSq`,
```

sherpa/io.py
```diff
@@ -185,8 +185,8 @@ def get_ascii_data(filename, ncols=1, colkeys=None, sep=' ', dstype=Data1D,
 
     Data lines are separated into columns - splitting by the
     ``sep`` comment - and then converted to NumPy arrays.
-    If the ``require_floats`` argument is ``True`` then the
-    column will be converted to the `sherpa.utils.SherpaFloat`
+    If the ``require_floats`` argument is `True` then the
+    column will be converted to a floating point number
```
DougBurke, line 189:
```suggestion
    column will be converted to a floating-point number
```

really-important change

sherpa/io.py
```diff
@@ -259,18 +259,18 @@ def read_data(filename, ncols=2, colkeys=None, sep=' ', dstype=Data1D,
        The name of the ASCII file to read in.
     ncols : int, optional
        The number of columns to read in (the first ``ncols`` columns
-       in the file). This is ignored if ``colkeys`` is given.
+       in the file). This is ignored if `colkeys` is given.
```
DougBurke, line 262: Not for this PR, but we need to decide whether we want to use ``ncols`` or `colkeys` (and it's probably something we may want to normally do one way but allow to change).

I currently don't have a preference.

DougBurke:
LGTM other than two minor comments.

DougBurke (approved):
LGTM

PR #1269, opened 2021-08-29
Reduce concurrent runs to github actions

hamogu:
## Summary
Run GH actions only of pull requests and direct push to main, but not for pushed to all branches

## Details
With current main, I see GH actions run twice every time I update a PR with a new commit: Once in the sherpa/sherpa repro (where the PR is) and once in my fork (hamogu/sherpa), see e.g.
https://github.com/sherpa/sherpa/pull/1182/checks?check_run_id=3454426337 and  https://github.com/hamogu/sherpa/actions/runs/1179131780 .
If I don't keep hamogu/sherpa/main up-to-date (and I usually don't, because I don't use that branch at all; I work from sherpa/sherpa/main, which is the most up-to-date version), then the results of the runs can differ as in the example above which is confusing.  Also, it's a waste of electrons and CPU cycles.

I *think* that the root cause is that actions runs for both the PR (because `on: pull_request` is set) as well as on all changes to a branch like hamogu/sherpa/multisim in the examples above, because `on: push` is set). So, I suggest to limit this to direct pushes to the main branch in this PR.
I can't claim to 100% understand github actions, but this makes sense to me and it's what e.g. astropy does, where I don't see the double runs.

Of course, it's possible that other people use the CI in they local branches before they open a PR and want this feature. (I run local tests in one environment and when those pass, I open a PR where the entire CI runs on the theory that, once it passes in one environment, it's usually good enough to be looked at). @DougBurke  what do you think?

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -1,6 +1,12 @@
 name: Conda CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - main
+    tags:
+    - '*'
+  pull_request:
 
 env:
   sherpa_channel: sherpa
@@ -24,11 +30,11 @@ jobs:
             matplotlib-version: 3
             xspec-version: 12.10.1s
 
-          - name: Linux Minimum Setup 
+          - name: Linux Minimum Setup
             os: ubuntu-latest
             python-version: 3.8
             numpy-version: 1.18
-            install-type: develop 
+            install-type: develop
             test-data: none
 
           - name: Linux Full Build (Python 3.9)
@@ -41,7 +47,7 @@ jobs:
             xspec-version: 12.11.1
 
           - name: Linux Full Build (Python 3.8)
-            os: ubuntu-latest 
+            os: ubuntu-latest
             python-version: 3.8
             install-type: develop
             fits: astropy
@@ -55,13 +61,13 @@ jobs:
             numpy-version: 1.19
             install-type: install
             fits: astropy
-            test-data: submodule 
+            test-data: submodule
             matplotlib-version: 3
             xspec-version: 12.10.1s
 
           - name: Linux Build (w/o Astropy or Xspec)
             os: ubuntu-latest
-            python-version: 3.9 
+            python-version: 3.9
             install-type: install
             test-data: package
             matplotlib-version: 3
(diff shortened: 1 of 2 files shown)


DougBurke:
I think I'll punt this to @Marie-Terrell / @harlantc as I don't have strong feelings about it. I am pretty sure @Marie-Terrell has used the existing feature to test out changes on their branch before making a PR. I have occasionally done this but normally just create a WIP PR.

I will note it's annoying to suddenly see emails from GH actions on your branch as well as the sherpa branch but it's not that often.

Marie-Terrell (approved):
So, we did swap around a bit between this and the current method when we were first switching to GitHub Actions. Though, we ended up sticking to the current method because it was the last one we had implemented before merging the change. At the time, the tests would not run at all prior to merging the first instance of GitHub Actions to the repo.

At this time, I actually do currently use the testing in my repo before I even create a PR. I do this as sometimes the configuration/dependency/conda/pip changes sometimes exhibit slightly different behavior in the workflows than I see when running my initial tests. Though, I am willing to work through those potential issues with a "DRAFT" added to my PR instead. 

In terms of the impact of this on Releases and branches: 

- We will still get testing on our sherpa releases as we use tags and a PR for the releases. 
- We will lose testing on cat branches. I know this is a special case, but just documenting it. Though, these often fail on the branches anyway because the dependencies are newer at that point or we needed to change something in the testing. They also get their own special testing off to the side anyway. Not to mention we hope to reconcile these differences and not have a separate CAT branch anyway (and this helps push us in that direction).

I guess at this point, it is more of a question of: 
- Whether we prefer people having those tests passing in your fork (on top of any manual testing already done) before submitting the PR
- And, whether we want we people to have the ability to test their changes in their fork before submitting a PR
The second question here being arguably more important than the other. 

Either way is valid and as you mention, other projects are using this method to avoid double testing. So, I'll approve the change and leave the final decision to Warren.

hamogu:
We could always add the CAT branches to the list of "whitelisted" branches in same way that "main" is listed. We could even add one or more branches like "marie_check_ci" to that list and, as long as you use those names in your local fork, the branch will still run on every push. That would require uses who want to that use some special branch name, but still cut down on runtime and noise for all others.

Marie-Terrell:
Thank you, that is a good point! Agreed. We could do that now or if/when we need it. Though, I don't think it will provide value at the moment since the test on the cat branch would fail anyway at the moment.

DougBurke:
@hamogu - can you remove my request for a review (so maybe we can get this merged at some point). I don't have any skin in the game here so if CXCCM are happy then I'm happy (and I don't think me just rubber-stamping it helps!).

PR #1270, opened 2021-08-30
Cleanup setup

DougBurke:
# Summary

Update the Python supported status in the package metadata (drop 3.5, add 3.8) and bump the minimum version to 3.6.

# Details

This has been broken out of #1223 as it is completely unrelated.

The main reason was to re-order the data setup to make it easier to compare to the on-disk (with a `find | sort` call) as I find these lists can quietly go out of date. The sherpa.optmethods package data has been updated to mention C++ files, but I don't know what these are / how they are used, just that we have them.

Ideally we would re-work all of this to meet more-recent standards/proto-standards (eg #784) but that is a longer-term project so let's do minor fixes. This probably overlaps with #1170 but I haven't checked

We should probably include Python 3.9 in the list - and maybe drop Python 3.6?


DougBurke:
@hamogu I don't know if you have any thoughts? I'd like @Marie-Terrell or @harlantc to review this to see if it makes sense (I am trying hard not to get sucked into learning about the intricacies of python packaging so I could be doing plenty of stupid things here)

hamogu:
I think this is good - a similar PR was opened extactly 100 issues ago see #1170 - however, that one missed the requirements file, but it includes other stuff instead. We should look at both together or combine them into a single PR.

Marie-Terrell (approved):
Looks good to me.

Issue #1251, opened 2021-08-13
Inconsistent behavior in save_arrays when optional parameter ascii=True/False


dtnguyen2:
Using the following script:
```
(sherpa-4.14-dev) [dtn@devel12 sherpa]$ cat tmp.py
import numpy as np
from sherpa.astro.ui import *

def tst(arg):
    tmp=np.arange(3)
    ascii = {0: False, 1: True}
    save_arrays("tmp.txt",[tmp, tmp, tmp], fields=["a", "b"], clobber=True,
                ascii=ascii[arg])

if '__main__' == __name__:
    import sys
    arg = int(sys.argv[1])
    tst(arg)
```

```save_array``` writes out the table even though the length of the second and third arguments are not equal when the optional argument is set as: ```ascii=True```

```
(sherpa-4.14-dev) [dtn@devel12 sherpa]$ python tmp.py 1
(sherpa-4.14-dev) [dtn@devel12 sherpa]$ cat tmp.txt
#a b
0 0 0
1 1 1
2 2 2
```
However, if the optional argument  is set as:```ascii=False```, it throws an exception (the correct behavior):
```
(sherpa-4.14-dev) [dtn@devel12 sherpa]$ python tmp.py 0
Traceback (most recent call last):
  File "tmp.py", line 13, in <module>
    tst(arg)
  File "tmp.py", line 8, in tst
    ascii=ascii[arg])
  File "<string>", line 1, in save_arrays
  File "/export/sherpa/sherpa/astro/ui/utils.py", line 3973, in save_arrays
    ascii=ascii, clobber=clobber)
  File "/export/sherpa/sherpa/astro/io/__init__.py", line 712, in write_arrays
    backend.set_arrays(filename, args, fields, ascii=ascii, clobber=clobber)
  File "/export/sherpa/sherpa/astro/io/pyfits_backend.py", line 1162, in set_arrays
    raise IOErr("wrongnumcols", len(args), len(fields))
sherpa.utils.err.IOErr: Expected 3 columns but found 2
```

PR #1276, opened 2021-09-02
Fix Inconsistent behavior in save_arrays when optional parameter ascii = True/False (Fixes issue #1251)

wmclaugh:
**Summary**

The behavior of the  save_arrays routine differs when using writing ascii vs binary files. In particular, an exception thrown when providing a different number of column names from data columns occurs for fits files but not ascii files. (Issue #1251 
Details)

**Details**

The pyfits back end save_arrays() routine was calling write_arrays for ascii files before handling exceptions for things such as clobber settings and inconsistent array sizes. The crates backend already correctly handled both instances. This change relocates the ascii write to occur after the checks so that the pyfits ascii and binary threads are consistent.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -33,7 +33,7 @@ from sherpa.astro.data import DataPHA
 from sherpa.astro.instrument import RMFModelPHA
 from sherpa.astro import ui
 from sherpa.data import Data1D
-from sherpa.utils.err import ArgumentErr, DataErr, IdentifierErr, StatErr
+from sherpa.utils.err import ArgumentErr, DataErr, IdentifierErr, IOErr, StatErr
 from sherpa.utils.logging import SherpaVerbosity
 from sherpa.utils.testing import requires_data, requires_fits, \
     requires_group, requires_xspec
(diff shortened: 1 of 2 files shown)


sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -714,6 +714,23 @@ def read_func(filename):
     save_arrays(colnames=colnames, fits=False, read_func=read_func)
 
 
+
+@requires_fits
+@pytest.mark.parametrize("ascii_type", [False, True])
+def test_save_arrays_colmismatch_errs(ascii_type):
```
DougBurke, line 720: You can use the tmp_path fixture to create a file name - so something like

```python
def test_save_arrays_colmismatch_errs(ascii_type, tmp_path):
    outfile = tmp_path / 'bogus.dat'
    outfile = str(outfile)  # Sherpa I/O doesn't handle Path objects
```

but I guess the point is that we actually don't expect the file to get created so is it worth the effort?

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -714,6 +714,23 @@ def read_func(filename):
...
+    with pytest.raises(IOErr) as exc:
+        a = numpy.asarray([1, 3, 5])
+        b = numpy.asarray([4, 6, 8])
+        c = a*b
+        fields = ["odd", "even"]
+        #ofh = tempfile.NamedTemporaryFile(suffix='sherpa_test')
+        #ui.save_arrays(ofh.name, [a, b, c], fields=fields,
```
DougBurke, line 727:
```suggestion
```

No need to keep this comment

DougBurke (changes requested):
Can we have the commented-out lines removed?

I don't think we need to bother with the `tmp_path` fixture I mentioned (see https://docs.pytest.org/en/latest/how-to/tmp_path.html for more info)

DougBurke (approved):

PR #1279, opened 2021-09-09
rm the platform cause no longer need it

dtnguyen2:
#Note 

rm the ```platform``` module since it is not needed


DougBurke (approved):
LGTM

PR #1285, opened 2021-09-10
Added an option to reflect minim about the boundary

dtnguyen2:
# Note

This PR adds the option on how the optimization ```minim``` will behave if the free parameter is beyond the limit.  The default parameter ```reflect``` is True (will reflect by an equal amount about the limit).  If ```reflect``` is set to False then model function will return ```DBL_MAX``` (~1e308) and therefore it will not be included in the simplex.


DougBurke (approved):
I do not have confidence in my C++ to be able to make meaningful comments on this part of the code (I am assuming that as the tests still pass and this has been validated externally in the L3 stuff then it's okay).

We do need to add documentation for the reflect option but that can be done in a separate PR.

Thanks for updating the "i" to p" (or whatever it was) from the original commit.

PR #1293, opened 2021-09-14
RTD: add a "how to optimize a function" example

DougBurke:
# Summary

Add a section to the ReadTheDocs site explaining how to "optimize a function" as this is something we can do but only really with the low-level code.

# Details

This is an offshoot of our discussions with 3ML but it's something I wanted to know how to do. There are some holes in our documentation - in particular for the lmdif-needed least-squares return value - but it's a start. The question is whether it's a useful start.


docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
+********************
+Optimizing functions
```
hamogu, line 2: I suggest "finding minima" or something like that. When I read it first, I thought this is about optimizing the code to make functions run faster. Maybe I'm too much of a coder an normal users won't think that way.

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+.. note::
+   The Sherpa test suite contains a number of test cases based on
+   functions like this, taken from
+   `More, J. J., Garbow, B. S., and Hillstrom, K. E.. Testing unconstrained optimization software. United States: N. p., 1978. Web. doi:10.2172/6650344. <https://www.osti.gov/biblio/6650344>`_.
+
+The function is::
+
+  def rosenbrock(x, y):
+      a, b = 2, 10
+      return (a - x) * (a - x) + b * (y - x * x) * (y - x * x)
```
hamogu, line 53:
```suggestion
      return (a - x)**2 + b * (y - x * x)**2
```
personal preference. I think it's easier to read because it's shorter, but either way is fine.

docs/extra/optimize_function.rst
hamogu: Maybe "minimizing functions"?

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+
+  def to_optimize1(args):
+      x = args[0]
+      y = args[1]
+      pa = (2 - x) * (2 - x)
+      pb = 10 * (y - x * x) * (y - x * x)
+      stat = pa * pa + pb * pb
+      return stat, None
+
+The function can tehn be monimized by passing the routine, the
```
hamogu, line 96:
```suggestion
The function can then be minimized by passing the routine, the
```
I know you put that in jsut ot maek suer we pya attentions.

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+starting point, and the bounds, to the optimizer- in this case the
+:py:func:`~sherpa.optmethods.optfcts.minim` routine:
```
hamogu, line 98:
```suggestion
`~sherpa.optmethods.optfcts.minim` routine:
```

Should work here, too. But not 100% certain, need to check output at RTD if we go with this suggestion. (If it does not work, I need to tweak the conf setting more - not not for this release.)

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+found the minimum location.
+
+In order to use :py:func:`~sherpa.optmethods.optfcts.lmdif` we need
+the per-parameter statistic, which we can get with a small tweak::
+
+  def to_optimize2(args):
+      x = args[0]
+      y = args[1]
+      pa = (a - x) * (a - b)
+      pb = b * (y - x * x) * (y - x * x)
```
hamogu, line 154:
```suggestion
      pa = (a - x)**2
      pb = b * (y - x * x)**2
```
Again, just my personal preference.

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+We can re-create the :ref:`Fitting the NormGauss1D and Lorentz1D models <fit_peaked_data_normgauss1d_lorentz1d>`
+section - at least for the :py:class:`~sherpa.models.basic.NormGauss1D` fit.
+
+The normalized gaussian model can be expressed as
+
+::
+
+  def ngauss(x, ampl, pos, fwhm):
+      term = 4 * np.log(2)
+      numerator = ampl * np.exp(-term * (x - pos) * (x - pos) / (fwhm * fwhm))
```
hamogu, line 185: Again `**2` to reduce the line length? This is long in my browser window, but of course it depends on the size of your screen and settings.

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+      return numerator / denominator
+
+and the data we are going to fit is read from a file:
+
+  >>> d = np.loadtxt('test_peak.dat')
+  >>> x = d[:, 0]
+  >>> y = d[:, 1]
+
+In this case the "callback" routine we want to give the minimizer a
+routine which will apply a least-squares minimizer to compare the
```
hamogu, line 196:
```suggestion
In this case the "callback" routine we want to give the minimizer is a
routine which will apply a least-squares minimizer to compare the
```

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+  def to_optimize1(args):
+      x = args[0]
+      y = args[1]
+      pa = (2 - x) * (2 - x)
+      pb = 10 * (y - x * x) * (y - x * x)
+      stat = pa * pa + pb * pb
+      return stat, None
+
+The function can tehn be monimized by passing the routine, the
+starting point, and the bounds, to the optimizer- in this case the
```
hamogu, line 97:
```suggestion
starting point, and the bounds, to the optimizer - in this case the
```

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+  def cb(pars):
+      model = ngauss(x, pars[0], pars[1], pars[2])
+      delta = model - y
+      statval = (delta * delta).sum()
+      # Keep a record of the parameters we've visited
+      store.add_row([statval] + list(pars))
+      return statval, delta
+
+.. note::
+   This function would normally be written as a closure to ensure
```
hamogu, line 211: I don't know what a "closure" is, I would use a callable class that keep that info as a class attribute. But either way, the point is to point out that that's a simply example, so whatever works here.
Or maybe a callable class is just one example to implement a  "closure"? I need to google and learn that term!

docs/extra/optimize_function.rst
```diff
@@ -0,0 +1,297 @@
...
+   that the use of global terms like `x`, `y`, `ngauss`, and `store` do not
+   cause problems.
+
+For the starting point and bounds I'm using to use a "guess" based on
```
hamogu, line 215:
```suggestion
For the starting point and bounds I'm using a "guess" based on
```

hamogu: The rest of the test is written with "we do this". Now, you switch to "I". Intentional?

hamogu (changes requested):
The content it great, the examples are great. Very useful addition to the docs. I just have a few things related to typos or formatting.

anetasie:
@DougBurke Thanks for the examples. Playing with 3D plots is fun!
The new section looks good.

anetasie:
> @DougBurke Thanks for the examples. Playing with 3D plots is fun!
> The new section looks good.

btw. I note the same typo as @hamogu comment above.

dtnguyen2 (commented):
I know that this is about _how to optimize a function_
```
def cb(pars):
     model = ngauss(x, pars[0], pars[1], pars[2])
     delta = model - y
     statval = (delta * delta).sum()
     # Keep a record of the parameters we've visited
     store.add_row([statval] + list(pars))
     return statval, delta
```
but should we mention that a history of the parameters visited during a fit can be done with an extra argument to the ```fit()``` function with the optional ```outfile```?

DougBurke:
@dtnguyen2 - this is talking at a much lower part of the software stack than the outfile argument (ie there's 'fit' call here).

docs/extra/optimize_function.rst
DougBurke: I note that the scipy docs are "Optimization"

https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html

and we have a big section on "Optimisers" at https://sherpa.readthedocs.io/en/latest/optimisers/index.html so I'm less concerned about this issue than you are. I don't particularly like the title but I do not have the energy/time to come up with a better one.

I will add a link to the scipy docs to let users know what we're trying to do here.

docs/extra/optimize_function.rst
DougBurke: How about

```suggestion
      return (a - x)**2  + b * (y - x**2)**2   # :-)
```

docs/extra/optimize_function.rst
DougBurke: Look. You thought we were learning about how to speed up code, not monomize functions!

hamogu: monomized, sure. But monimized? https://in.pinterest.com/srsirisharao/monomize/

docs/extra/optimize_function.rst
DougBurke: Interestingly they are **slightly** different, but not enough that I care about...

```
TEST1 `~sherpa.optmethods.optfcts.minim` routine

TEST2 :py:func:`~sherpa.optmethods.optfcts.minim` routine
```

creates

image

docs/extra/optimize_function.rst
DougBurke: I'm re-wording this entire single-sentence paragraph.

docs/extra/optimize_function.rst
DougBurke: A closure "encloses" the symbols it uses and I have seen it used in python discussions - even Guido has used it ...

https://www.python.org/dev/peps/pep-3104/#id47

docs/extra/optimize_function.rst
DougBurke: I'm ambidextrous... We shoulf=d really scrub the documnetation for this sort of change in style - I do it a lot - but then I decide that consistency is a hobgoblin of minds that aren't knee deep in documentation and forget...

docs/extra/optimize_function.rst
DougBurke: I've rewritten this to say something "a class or a closure"

DougBurke:
I think I've addressed the review comments.

I am not convinced I'm doing a good job of describing what the callback routine should be returning - e.g. the least-squared difference in the first case is not well elucidated and the terms needed by lmdif is hand-waved over.

We do have https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html as a model

docs/extra/optimize_function.rst
hamogu: OK, good point. Fine with me.

hamogu (approved):

PR #1294, opened 2021-09-14
RTD: Fix some sphinx warnings (sherpa.models.regrid)

DougBurke:
# Summary

Fix a minor warning when generating the Read The Docs pages.

# Detaiils

The line

    blah(foo, bar, **kwargs)

was triggering something in sphinx - I think the '**' phrase - so protect them (add single quotes) and in a few other places. For reference, the warning messages this removes are

```
/home/dburke/sherpa/sherpa-xspec/sherpa/models/regrid.py:docstring of sherpa.models.regrid.ModelDomainRegridder1D.eval_integrated:5: WARNING: Inline strong start-string without end-string.
/home/dburke/sherpa/sherpa-xspec/sherpa/models/regrid.py:docstring of sherpa.models.regrid.ModelDomainRegridder1D.eval_non_integrated:5: WARNING: Inline strong start-string without end-string.
```

### DIFF
--- a/sherpa/models/regrid.py
+++ b/sherpa/models/regrid.py
@@ -418,7 +418,7 @@ class ModelDomainRegridder1D():
         The interpolation function: it should accept three arrays: the
         output x values and the x, y values to interpolate, and return
         the interpolated y values. The default is to use
-        sherpa.utils.akima.akima.
+        `sherpa.utils.akima.akima`.
 
     Examples
     --------
@@ -428,7 +428,7 @@ class ModelDomainRegridder1D():
     linearly-interpolated onto the desired grid (1 to 8,
     spacing of 0.7). In this example there is no benefit to
     this approach - it is easier just to evaluate
-    ``internal_mdl`` on the grid ``x`` - but it illustrates
+    `internal_mdl` on the grid `x` - but it illustrates
     the approach.
 
     >>> from sherpa.models import Gauss1D, Const1D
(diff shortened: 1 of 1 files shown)


sherpa/models/regrid.py
```diff
@@ -565,7 +565,7 @@ def eval_non_integrated(self, pars, modelfunc, data_grid, eval_grid,
             The parameters for the model.
         modelfunc : callable
             The model to evaluate. It is called as
-            modelfunc(pars, x, **kwargs)
+            `modelfunc(pars, x, **kwargs)`
```
hamogu, line 568:
```suggestion
            ``modelfunc(pars, x, **kwargs)``
```
We now use single backticks for stuff that's linked (like sherpa.utils.akima.akima), but that means to format it as "just text that looks like code" we know need ot use double backticks.

sherpa/models/regrid.py
```diff
@@ -619,7 +619,7 @@ def eval_integrated(self, pars, modelfunc, data_grid, eval_grid,
             The parameters for the model.
         modelfunc : callable
             The model to evaluate. It is called as
-            modelfunc(pars, lo, hi, **kwargs)
+            `modelfunc(pars, lo, hi, **kwargs)`
```
hamogu, line 622:
```suggestion
            ``modelfunc(pars, lo, hi, **kwargs)``
```

hamogu (changes requested):

sherpa/models/regrid.py
hamogu: Note that this will work (so I'm going to approve), but I hope that at some point in the future we can active sphinx warnings to links that don't resolve to an object to catch typos, moved objects etc. `modelfunc(pars, x, **kwargs)` won't resolve to a Pytohn class in the docs, so this would raise a warning when we do that, but we can also deal with it when we get to it (maybe never).

hamogu (approved):

PR #1296, opened 2021-09-14
Document the neldermead reflect flag from #1285

DougBurke:
# Summary

Add documentation for the reflect keyword added in #1285 for the neldermead optimizer

# Details

The first commit cleans up the use of ` vs `` which changes how the ReadTheDocs displays the text but it does not change any functionality.

The second commit adds a brief description of the reflect keyword from #1285.

Of note is there isn't really any existing documentation the optimizer settings as used by the UI layer - e.g. the get/set_method_opt routines (as it's not clear how best to  do this), so we don't need to add any documentation there.

### DIFF
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -671,6 +671,11 @@ class NelderMead(OptMethod):
     verbose : int
        The amount of information to print during the fit. The default
        is `0`, which means no output.
+    reflect : bool
+       When a parameter exceeds a limit should the parameter be
+       reflected, so moved back within bounds (`True`, the default) or
+       should the model evaluation return DBL_MAX, causing the current
+       set of parameters to be excluded from the simplex.
 
     Notes
     -----
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -200,21 +200,21 @@ class OptMethod(NoNewAttributesAfterInit):
         ----------
         statfunc : function
            Given a list of parameter values as the first argument and,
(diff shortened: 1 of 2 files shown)


DougBurke:
@dtnguyen2 - I'm just looking for whether the text I added in the second commit makes sense. I based it on you r text from #1285 but may have mangled it.

anetasie (approved):
Text looks good on RTD

dtnguyen2 (commented):
I don't think lines 559 and 774 
```
       the default is sqrt(DBL_EPSILON) ~ 1.19209289551e-07, where
       DBL_EPSILON is the smallest number x such that ``1.0 != 1.0 +                         
                     
```
are not correct.
```Python
def montecarlo(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, verbose=0,
               seed=74815, population_size=None, xprob=0.9,
               weighting_factor=0.8, numcores=1):

def neldermead(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None,
               initsimplex=0, finalsimplex=9, step=None, iquad=1,
               verbose=0, reflect=True):

```
should be changed to
```
       the default is FLT_EPSILON ~ 1.19209289551e-07, where
       FLT_EPSILON is the smallest number x such that ``1.0 != 1.0 +                         
```

Perhaps we can add that ```epsfcn``` (default is ```FLT_EPSILON```) should be set to ```DBL_EPSILON``` if all the model functions are in double precisions since it will minimize the errors in the numerical derivates used in the optimization method.

Issue #1040, opened 2021-01-01
`DataPHA` object fails display if channel does not start from 0


hamogu:
I found the following behaviour surprising:

This works fine:
```
from sherpa.astro.data import DataPHA
x = np.arange(15)
d = DataPHA('test',x, np.arange(15))
d.group_counts(4)
d
```
This is in the jupyter notebook, so the last line gives me a nice plot and summary of my data object.

However, if the channel number does not start at 0, e.g. I replace the second line with `x = np.arange(15) + 10` then I get instead:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in _group_to_channel(self, val, group, response_id)
   1902         try:
-> 1903             return mid[val]
   1904         except IndexError:

IndexError: index 12 is out of bounds for axis 0 with size 12

During handling of the above exception, another exception occurred:

DataErr                                   Traceback (most recent call last)
~/anaconda3/envs/ciao/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in _repr_html_(self)
   1446         """Return a HTML (string) representation of the PHA
   1447         """
-> 1448         return html_pha(self)
   1449 
   1450     def __getstate__(self):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in html_pha(pha)
    273     # Should this only be displayed if a filter has been applied?
    274     #
--> 275     fexpr = pha.get_filter_expr()
    276     bintype = 'groups' if pha.grouped else 'channels'
    277     nbins = pha.get_dep(filter=True).size

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in get_filter_expr(self)
   3419     def get_filter_expr(self):
   3420         return (self.get_filter(format='%.4f', delim='-') +
-> 3421                 ' ' + self.get_xlabel())
   3422 
   3423     def notice_response(self, notice_resp=True, noticed_chans=None):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in get_filter(self, group, format, delim)
   3401 
[...]

Issue #1305, opened 2021-09-24
DataPHA objects without ARF/RMF only work for channel numbers starting with 1


hamogu:
I create a `DataPHA` object from scratch because I'm fitting data that's not in an OGIP compliant format (it's been taken in a ground-based instrumentation lab for testing early development models of X-ray instrumentation). Because of it's origin, I don't have an ARF and RMF, so I'm just passing in `bin_lo` and `bin_hi`. When I start the channel numbering with 0 strange things happen. I saw this first in a plot, but traced it back to `DataPHA.get_x()`. Note the order order of the energies listed.
```
import numpy as np
from sherpa.astro.data import DataPHA
bins = np.array([.2, .4, .6, .8, 1.])
testpha = DataPHA(name='testdata', channel=np.arange(4), counts=[5, 6, 7, 8],
              bin_lo=bins[:-1], bin_hi=bins[1:])
testpha.set_analysis('energy')
testpha.get_x()

>>> array([0.9, 0.3, 0.5, 0.7])
```

It also fails for channel numbers starting > 1, but with a different error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in _channel_to_energy(self, val, group, response_id)
   1910         try:
-> 1911             return (elo[val] + ehi[val]) / 2.0
   1912         except IndexError:

IndexError: index 4 is out of bounds for axis 0 with size 4

During handling of the above exception, another exception occurred:

DataErr                                   Traceback (most recent call last)
<ipython-input-63-29682d4539c6> in <module>
      5               bin_lo=bins[:-1], bin_hi=bins[1:])
      6 testpha.set_analysis('energy')
----> 7 testpha.get_x()

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in get_x(self, filter, response_id)
   3092         # We want the full channel grid with no grouping.
   3093         #
-> 3094         return self._from_channel(self.channel, group=False, response_id=response_id)
   3095 
   3096     def get_xlabel(self):

~/anaconda3/envs/ciao/lib/python3.7/site-packages/sherpa/astro/data.py in _channel_to_energy(self, val, group, response_id)
   1911             return (elo[val] + ehi[val]) / 2.0
   1912         except IndexError:
-> 1913             raise DataErr('invalidchannel', val)
   1914 
[...]

PR #1306, opened 2021-09-24
Make `DataPHA.get_x()` work if channels are not counted starting at 1.

hamogu:
## Summary
Make `DataPHA.get_x()` work if channels are not counted starting at 1.

## Details
See #1305 for details. OGIP allows any number as starting number to count channels, but calls out 0 and 1 as examples. Of course, this may break in other parts of the code, but let's fix them as we find them.

### Note
I'm pushing this up without running the tests locally because it's a pretty simple change and I don't have right now time to debug why I can't build sherpa locally any more. (some libraries not found).

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2224,7 +2224,7 @@ class DataPHA(Data1D):
 
     def _channel_to_energy(self, val, group=True, response_id=None):
         elo, ehi = self._get_ebins(response_id=response_id, group=group)
-        val = numpy.asarray(val).astype(numpy.int_) - 1
+        val = numpy.asarray(val - self.channel[0]).astype(numpy.int_)
         try:
             return (elo[val] + ehi[val]) / 2.0
         except IndexError:
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1359,6 +1359,8 @@ class DataPHA(Data1D):
         The statistical and systematic errors for the data, if
         defined.
     bin_lo, bin_hi : array or None, optional
+        lower / upper bound for channels, in keV. If an ARF and/or RMF
+        is loaded, the values in the ARF will be used instead.
     grouping : array of int or None, optional
     quality : array of int or None, optional
     exposure : number or None, optional
@@ -1972,7 +1974,6 @@ class DataPHA(Data1D):
             resp = instrument.Response1D(self)
 
         return resp
-
 
     def _get_ebins(self, response_id=None, group=True):
         """Return the low and high edges of the independent axis.
--- a/sherpa/astro/tests/test_astro_data.py
+++ b/sherpa/astro/tests/test_astro_data.py
@@ -2622,3 +2622,20 @@ def test_pha_channel0_subtract():
         np.array([0, 1, 2, 0, 3, 1], dtype=np.int16)
     assert p1.get_dep(filter=True) == pytest.approx(expected[1:4])
     assert p0.get_dep(filter=True) == pytest.approx(expected[2:5])
+
+
+@pytest.mark.parametrize("channels", [[0, 1, 2, 3],
+                                      [1, 2, 3, 4],
+                                      [5, 6, 7, 8]])
+def test_pha_channel_noarfrmf(channels):
+    '''Check correct energies are returned for channels starting at 0 or 1.
+
+    Regression test for https://github.com/sherpa/sherpa/issues/1305
+    '''
+    bins = np.array([.2, .4, .6, .8, 1.])
+    testpha = DataPHA(name='testdata',
+                      channel=np.arange(4),
+                      counts=np.array([5, 6, 7, 8]),
+                      bin_lo=bins[:-1], bin_hi=bins[1:])
+    testpha.set_analysis('energy')
+    assert testpha.get_x() == pytest.approx([.3, .5, .7, .9])


DougBurke:
For PRs like this I've been trying to get into the habit [*] of

a) add tests
b) change code

where the tests in a may either be marked as xfail (for the cases where we can say 'this is obviously wrong') or just document the current behavior.

It makes it a bit easier to see what the changes in b do to the tests.

[*] which normally requires excessive use of rebasing to split apart commits and moving them around.

DougBurke:
So we have

- channel starting at 1 (I'd like to say that all issues are fixed but we have quality handling and probably some edge cases still hiding around)
- channel starting at 0
- channel starting at something other than 1 or 0 (I'd put this as less important but it should hppefully come along for free when handling the 0 case, as long as we  add tests)

and then there's PHA files which only contain a subset of the valid set of channels (which is a related, but separate, issue, in part because I'd put it at a lower priority than the channel=0 case).

I would like to have a "test" PR that adds a bunch of tests - not just for get_x - and in particular some "end-to-end" tests (e.g. including grouping and filtering and fitting) so that we can see how much a change like this one fixes things. I did recently add some very-basic channel=0 tests but they aren't really sufficient (e.g. they didn't highlight this problem).

hamogu:
For the record: To avoid the excessive rebasing mentioned by @DougBurke above, I've added the test first, verified locally that it failed except when the channels start at 1 and then implemented the (obvious) fix.

While end-to-end test are nice, I was really working on something else and don't want to do full sherpa spelunking now. This is an obvious bug, I added a test, I fixed it and everything else is out of the scope of this PR. Note also, that the fix in in a code back that is "unusual", i.e. only when no arf or rmf are set - a case we only see in development work on other crazy circumstances, while the "end-to-end" tests that were suggested above should most importantly cover our real use cases for DAtaPHA: X-ray data with arf / rmf.

hamogu:
I realize now that I run into this before - this PR should close #1040.

DougBurke:
In #1040 I was trying to point out that we need to think about how we read in files like this (ie with TLMINn != 1 for whatever n is the channel column) as I think that there are issues in both the pyfits and crates backends. Which is part of my "if we're going to fix this, let's try and actually make it workable" vibe from above. Last time I tried however I ended up being discouraged because of the changes we'd need to make to DataPHA (I think, I forget now).

hamogu:
I suggest to do this fairly isolated change because it goes in the right way - and it only effects DataPHA objects created "by-hand", which is different from reading anything from a file with either backend.

hamogu:
@DougBurke: I know it's been a while. Can I ask you to have a look again (the failure is only codecov, not something that matters here)?
I totally admit that it does not solve all problems we could possibly have with TLMIN != 1, but it is a very isolated change that goes in the right direction. At some point, we do want to be able to work with all TLMIN more broadly, but it's a mess. This change is only for a single code path - a DataPHA created by hand in init, not read in from a file. That's something we barely do, so there is essentially no risk of side effects or confusing users. It's clearly just a bug fix.

And later, when we get to solve all TLMIN problems, then we will be happy that we have this one in already. On the other hand, we we close this now, then we'll forget about it and will painfully rediscover it later. I discovered it twice already (in #1040 and in #1305), I don't want to discover it again.

DougBurke:
@hamogu - what is the underlying cause that makes you keep on hitting this? Is it that in your code you have `channel=np.arange(4)` - i.e. a channel range starting from 0 - or are you reading things in some other way. 

Also, why (for #1305) do you want a `DataPHA` object if you don't have a response (is it to get the grouping or some other reason)?

hamogu:
I want DataPHA in my example because it's a PHA set and I want to group it. If we had a different "DataBinned" class I could use that instead. Also, some of the plotting works without an RMF, I get the nice display of the data object in the notebook (listing the stuff that's relevant for PHA). 

Why do I keep hitting this? Because I work on calibration and instrument development. I've hit this in ARCUS work and in work for the Chandra ACIS gain calibration - both cases where we don't have a complete OGIP compliant set of PHA, ARF, RMF. In Arcus, I'm running the simulations to make the RMF, but in the process I need to fit PHA data. In Chandra calibration, we want to calibrate the GAIN, so the usual channels-to-energy conversion is not valid (I want to get that, after all) and this I can't (shouldn't) use the normal Chandra ARFs and RMFs. So, I do get `DataPHA` that I set up manually, e.g. I chop out something from the middle of a larger Chandra PHA file or I exctract (manually) something from just one Chandra or Arcus chip, and thus the channels range of the grating spectrum does not start at 0.

To repeat what I said above: 
> I totally admit that it does not solve all problems we could possibly have with TLMIN != 1, but it is a very isolated change that goes in the right direction.

I'm not quite sure I understand @DougBurke s reluctance to review this here. It's a bug, I implemented a test of the exiting behavior, verified that it failed, and then fixed the code.

hamogu:
\begin{rant}
@DougBurke asked me for "real-world examples". I assume that mean "example that use PHA files". Well, we don't have that because this code path is only used when `DataPHA` is created "by hand". If we don't want to support that, let's openly declare "Sherpa DataPHA objects can only be created from OGIP compliant files" and remove this code (and I'll write my own `MoritzDataPHA` class that add this functionality back in for me). But, as of now, the this code is here and it's incomplete, so what not fix it.
\end{rant}

Issue #816, opened 2020-05-26
Test examples in the docs


hamogu:
The CI should include a run that tests the examples in the docs (pytest extension 'doctest-plus' work well for that). That should catch cases like https://github.com/sherpa/sherpa/pull/815 and makes sure that the examples are up-to-date with any code / parameter names changes.

hamogu:
xref #784

DougBurke:
I haven't done this in the past

a) for lack of time

b) there is *an* argument to be made about being careful to ensure that document updates are intentional (that is, to make sure some change hasn't happened "by accident"). I don't think this argument is should stop improvements to the automation of the documentation, but it's something I think about as there have been (and probably still are) cases where code and documentation disagree, and then how do we work out which is correct?

hamogu:
Of course for cases where docs and code diverge, we need to make a decision, what the "right" things to do should be.

But it needs to be done - because we want users to be able to use the docs to understand how to use sherpa. We don't want them to accidentally diverge (because then users need to read the code to find how the options are called) and the only way to stop that is to run the doc examples. 

I promise that in 90% of the cases failures are either typos or intentional changes not applied to the docs.

BTW: Nice to have you back!

DougBurke:
I have to admit that my earlier comment was due to me reading this wrongly - it was more aimed at a discussion Omar and myself have had about generating the documentation output at build time rather than this issue.

hamogu:
This is not complete, but all the infrastructure is in place and we test more an more of the examples in the docs. I feel this does not need an extra issue any more; we just address this as we work on docs. Thus, I'm closing this issue.
An alternative would be to turn this issue into a list to be ticked of module my module, but that's just duplicating the exclusion list for doctestplus in the configuration files.

PR #1308, opened 2021-09-28
Docs: fix get_filter examples

DougBurke:
# Summary

Fix the examples in the get_filter documentation.

# Details

Somehow some of the examples I added were borked, in part due to all the work done to split up the code into separate PRs and then the endless rebases. I've also added a few more examples to the DataPHA.get_filter documentation.

If we had doctest set up we would have caught this. The trouble is that setting up doctest is more work than I have time, energy, or interest in...

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3774,9 +3774,7 @@ class DataPHA(Data1D):
     def get_filter(self, group=True, format='%.12f', delim=':'):
         """Return the data filter as a string.
 
-        For grouped data, or when the analysis setting is not
-        channel, filter values refer to the full range of each
-        channel or group.
+        The filter expression depends on the analysis setting.
 
         .. versionchanged:: 4.14.0
            Prior to 4.14.0 the filter used the mid-point of the bin,
@@ -3787,9 +3785,8 @@ class DataPHA(Data1D):
         group : bool, optional
             Should the filter reflect the grouped data?
         format : str, optional
-            The formatting of the numeric values (this is
-            ignored for channel units, as a format of "%i"
-            is used).
+            The formatting of the numeric values (this is ignored for
+            channel units, which uses ``format="%i"``).
         delim : str, optional
             The string used to mark the low-to-high range.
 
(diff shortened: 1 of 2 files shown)


anetasie:
@DougBurke This looks ok and it should be included in the release. @hamogu are you going to review it?

anetasie:
@DougBurke does this PR relates also to #1161 work?

DougBurke:
@anetasie - No. It relates to the merged  #1208, #1215, #1216, #1219 work (probably 1216 or 1219).

hamogu:
We have #816 open for doc tests, but, alas, I did not have time/energy to deal with that either so far... One step at a time!

sherpa/astro/data.py
```diff
@@ -3799,22 +3796,41 @@ def get_filter(self, group=True, format='%.12f', delim=':'):
...
-
-        >>> pha.get_filter(format=%.4f')
-        '0.5183:8.2198'
-        >>> pha.get_filter(group=False, format=%.4f')
-        '0.4745:9.8623'
+        '33:676'
+
+        The filter expression shows the first selected channel to the
+        last one, and so is independent of whether the data is grouped
+        or not:
```
hamogu, line 3807: I don't like that wording. Below, you show "the selected range does depend on whether the data is grouped or not". It can't be, at the same time, depend and not depend on grouping. Maybe "the group argument to get_filter does not change the filter expression (it is only here for consistency with other functions), but grouping and ungrouping the data *does* change the selected range"  ?

hamogu (changes requested):

sherpa/astro/data.py
DougBurke: Unfortunately I don't have time to make changes other than simple typo fixes. This behavior is related to #1309 - which I realized from writing this PR - and I claim that at the moment you can have different behavior because it's technically a different meaning of "group" (the question is should we, and fixing this would be an actual code change which we can't get in before the release, as it isn't in a critical code path). 

For users of sherpa.astro.ui it doesn't matter as the `get_filter` call there does not have a `group` argument. This is only relevant for users of the `sherpa.astro.data.DataPHA.get_filter` call.

hamogu (approved):
This is better than not having it, even if it's not perfect. With the current deadline, let's take this and worry about any further improvement later.

PR #1310, opened 2021-09-29
Update the python check for python setup.py installs to 3.7

Marie-Terrell:
This was held out of #1270  to avoid causing issues for CIAO that still included Python 3.6 builds at the time. This is no longer the case and we can finally update the check for python setup.py based builds to require Python 3.7+.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,9 +23,9 @@ import sys
 # python_requires will stop pip, but also let users who are using
 # 'python setup.py develop'.
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
For reference I would hope we can remove this check from `setup.py` when we move to PEP-517/... builds as it would be encoded in the setup.cfg/pyproject.toml file instead. But that's for the future!

PR #1314, opened 2021-10-06
test: add requires_data decorator

DougBurke:
# Summary

Add a needed decorator for a test. There's no functional change to the code.

# Details

This would only be triggered if you had a FITS backend but the test data was not available, in which case the test would fail because there was no data.


DougBurke:
This was likely (aka I'm sure it was but haven't gone to check) introduced in whatever PR I added which converted the `sherpa/astro/tests/test_astro.py` from "unittest" to "pytest" style.

Issue #1316, opened 2021-10-07
python 3.10: python setup.py install fails in grplib-4.9 / stklib-4.11


DougBurke:
```
% pythons setup.py install
...
... looks good
...
... but then
...
checking for python3.1... /home/dburke/anaconda/envs/py310/bin/python
checking for a version of Python >= '2.1.0'... yes
checking for a version of Python >='2.7'... yes
checking for the distutils Python package... no
configure: error: cannot import Python module "distutils".
Please check your Python installation. The error was:
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
configure: error: ./configure failed for grplib-4.9
```

Unsurprisingly you get the same problem with `pip install .` but it's easier to see the errors with the above.

DougBurke:
You can turn off the problem components in the `setup.cfg` file (I'm surprised that `region-4.14` doesn't fail, but maybe that's because we don't actually build a python interface in `extern/` like we do the other two):

```
%  git diff
diff --git a/setup.cfg b/setup.cfg
index 8cacc6fd..b845937d 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,14 +16,14 @@ bdist_wheel = sherpa_config bdist_wheel
 #configure=None

 # GROUP Python module
-#disable-group=True
+disable-group=True
 #group-location=build/lib/python2.7/site-packages/group.so

 # Additional flags to be passed to libgrp build
 #group-cflags=

 # File Stack Python module
-#disable-stk=True
+disable-stk=True
 #stk-location=build/lib/python2.7/site-packages/stk.so

 # FFTW Library
```

DougBurke:
With

- `disable-group` and `disable-stack`
- PR #1317 
- PR #1318 

I was able to get a build with Python 3.10 + astropy + matplotlib passing the tests (including the data tests). So that's good.

DougBurke:
Hmmm. It looks like I now can build with python 3.10. I know there have been a large number of updates to setuptools recently, and maybe there's also been a change to python 3.10/distutils.

DougBurke:
However, trying to build XSPEC support throws up the following - it may be related to #1328 which recently got merged

```
pip install -e .
Obtaining file:///home/dburke/sherpa/sherpa-master
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #1317, opened 2021-10-07
Add requires group test decorators

DougBurke:
# Summary

Ensure the tests that require the group library are marked with the requires_group decorator.

# Details

To hit the need for this we need to be building with the following set in `setup.cfg`, which is rare, but necessary to get python 3.10 testing started thanks to #1316 

```
disable-group=True
```


hamogu (approved):
Always amazed at the speed @DougBurke can churn out PRs!

PR #1319, opened 2021-10-07
HACK to build stk/group libraries with Python 3.10

### LINKED ISSUE #1316: python 3.10: python setup.py install fails in grplib-4.9 / stklib-4.11

DougBurke:
This is just to let people build with Python 3.10 but is not something that should be merged.....

- we change the `configure` and `m4/AX_PYTHON_DEVEL.m4` files in `extern/stklib-4.11/
- we only change `configure` in `extern/grplib-4.9` (since we're not re-building `configure`)
- this should be handled by `autotools`/whatever it's called.

It fixes #1316

### DIFF
--- a/extern/grplib-4.9/configure
+++ b/extern/grplib-4.9/configure
@@ -12204,7 +12204,7 @@ $am_python_setup_sysconfig
 if can_use_sysconfig:
     sitedir = sysconfig.get_path('purelib', vars={'base':'$am_py_prefix'})
 else:
-    from distutils import sysconfig
+    from setuptools._distutils import sysconfig
     sitedir = sysconfig.get_python_lib(0, 0, prefix='$am_py_prefix')
 sys.stdout.write(sitedir)"`
      case $am_cv_python_pythondir in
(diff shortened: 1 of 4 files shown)


DougBurke:
So, this looks like it's not needed anymore, but I'll leave it as draft for a little bit,

DougBurke:
With recent changes this doesn't appear to be needed anymore so I'm closing.

Issue #903, opened 2020-08-21
Trying to understand the coverage reports


DougBurke:
@hamogu - do you have any experience here?

On my machine, if I `python setup.py develop` and then

a) run `python setup.py -q test -a "--cov sherpa --cov-report term"`
b) run `pytest --cov=sherpa --cov-report=term`

and then use `coverage report`, I get different results. I've looked at `helpers/test.py` which is our code around the `setup.py test` command, and I can't see anything obvious. However, we obviously have some strange behavior with `setup.py test`as I'm currently hitting odd things in experiments in #898 , so maybe I'm missing something obvious.

* The `setup.py` approach returns a value similar to the codecov output (not too surprising, as this is how the coverage is created on Travis-CI), that is

```
TOTAL                                                    24480   8169   7040   1104    64%
```

* The `pytest` approach is better, as it returns

```
TOTAL                                                    24480   6494   7040   1123    69%
```

[This is for a run with matplotlib, astropy, the data directory, but no DS9]

I **don't** understand.

I actually believe the `pytest` results more that the `setup.py` - for instance the setup.py tests claim that in the following, lines 76 , 83, and 85 are never tested, but lines 77-81  are tested. The `pytest` results agree on lines 83 and 85, but point out that line 76 has been tested. Which it must have been tested otherwise 77-81 could never have been tested...

```
76 def compile_energy_grid(arglist): 

77    elo = numpy.unique(numpy.concatenate([indep[0] for indep in arglist])) 

78    ehi = numpy.unique(numpy.concatenate([indep[1] for indep in arglist])) 

79 

80    in_elo = numpy.setdiff1d(elo, ehi) 

81    in_ehi = numpy.setdiff1d(ehi, elo) 

82    if len(in_elo) > 1: 82 ↛ 83

83        ehi = numpy.concatenate((ehi, in_elo[-(len(in_elo) - 1):])) 

84    if len(in_ehi) > 1: 84 ↛ 85

85        elo = numpy.concatenate((elo, in_ehi[0:len(in_ehi) - 1])) 

86 
```

DougBurke:
I'm closing this.

PR #1321, opened 2021-10-14
OLD: Add codecov.yml to fix path issue and update codecov uploader

### LINKED ISSUE #1320: CI: codecov uploaded change
Looks like we have to update the GH action to upload the coverage report to 

https://about.codecov.io/blog/introducing-codecovs-new-uploader/

although the GH action is meant to be automatically updated so need to check 

https://github.com/marketplace/actions/codecov

(although I'm not sure we make use of this yet?)

harlantc:
# Summary
Fix path parsing issue on codecov as well as updated for the new codecov uploader

# Details

In investigating the codecov reports jumping around a lot and just not making a whole lot of sense (#903), it seems the file tree for every report has been split between the normal project root, "sherpa" and a failed attempt by codecov to fix some more complex paths from conda builds to "work/sherpa/sherpa/sherpa", causing inconsistencies in the reports. trying to examine these files on codecov in the 'work/sherpa/sherpa/sherpa' dir reports as not found, leading to strange behavior. This means all our reports since the beginning is not accurate. 

This PR is to fix with codecov's [path fix](https://docs.codecov.com/docs/fixing-paths) and also introduce a way to configure reports for enhancements in the future if we want them (such as reorganizing reports by test type, system, etc, or changing how reports are evaluated and displayed).

This PR also fixes #1320, changing the codecov installation in CI from conda/pip (the deprecated bash version), instead, using the codecov actions from Github. This also plays into the previous issue surprisingly. The codecov Github action exposes the selection of reports, which revealed that reports returned to codecov was several files, all but one of them being coverage reports. These were (likely) interfering with codecov, resulting in the strange issues we've been seeing.

## specific changes
Add a codecov.yml
- sets path fix from "work/sherpa/sherpa" to "" to correct weird path settings
- sets the number of builds to 8, in order to match those returned from CI jobs.
- tell codecov to only report 'informative'ly, to keep reports from failing CI runs (TBD if this is kept)

Move codecov install for all CI tests into a codecov task
- This action uses the new uploader, solving the deprecation issue
- exposes the file selection for each report, setting it results in reports, not including unrelated files codecov guessed to be correct, possibly confusing reports.
- change coverage reports from "term" to "xml", codecov prefers xml anyway, and couldn't find them with term.

### DIFF
--- a/.github/workflows/ci-pip-workflow.yml
+++ b/.github/workflows/ci-pip-workflow.yml
@@ -86,7 +86,6 @@ jobs:
 
         pytest --cov=sherpa --cov-report=xml
 
-
     - name: upload coverage
       uses: codecov/codecov-action@v2
       with:


DougBurke:
@harlantc - thanks for looking at this. It may be related to #1152 (at least it points to #1001 as the place I probably added this "duplicate" coverage issue)

PR #1328, opened 2021-10-21
Versioneer cleanup

DougBurke:
# Summary

Update the versioneer module we use from 0.14 to 0.21 (latest).

# Details

This PR is based on #1336 to work around the #1334 issue

Technically we don't need to update versioneer since "it works now", but given the changes in the python build ecosystem I think it's worth trying to keep things up to date. Unfortunately things aren't that simple because

1. version 0.15 of versioneer updated the configuration to use `setup.cfg` rather than `setup.py` (this is an easy fix, and fits in with the general changes in the python build ecosystem to be more configuration based)
2. we had manually modified the versioneer code to support passing in our over-ridden versions of the build commands (this is the more awkward issue, described below)

The first two commits are test related (the first is unrelated to this PR but it changes the same test file so I did it here and just uses  tmp_path fixture rather than the I-believe-to-be-deprecated tmpdir fixture; note that we have other tests using tmp_path, and it was added in version 3.9.1 of pytest (or some version < 4) that means it is okay for us to rely on it being available).

There used to be a third commit  ("CI add pip freeze lines" is just to help check a current problem we are seeing - related to #1334 - and can be removed (but it may be useful to ensure we keep something like this for the future?)) but I've removed this as it would conflict with recent changes in #1339 and it wasn't needed here.

The next three commits are where the fun is

1. `setuptools is required so remove fall-through code` just removes some code that would error out if setuptools is not required and is old code from the "olde days" of python, as we now require setuptools via pyproject.toml / #782
2. `be explicit about the build_ext command` is at the heart of the existing over-ride we were making to the `versioneer` code - so we are now explicit in the `helpers.commands` dictionary about where the `build_ext` command is coming from (although in this commit we don't use it, that comes in the next commit)
3. `update to 0.21` is where we update versioneer to the released version of 0.21 and remove our manual adjustments, which is made possible by the versioneer.get_cmdclass routine now accepting a cmdclass input, so it can extend the commands we send in). We also move the configuration from `setup.py` to `setup.cfg`, run `versioneer init` - actually ran the code and accepted the changes, and use full-revisionid when building the docs

# Building Sherpa for distribution via conda/PyPI/...

I do not have any experience with building the package for release, but as far as I can tell from the documentation it should work the same, but this **needs testing**.

# One change I've noticed

There is also a change to the `_versions.get_versions()` output which I don't know if anyone but me will notice (well, the docs build did notice but that is included in this change now). With the current CIAOT build as an example we have

```
>>> import sherpa
>>> sherpa.__version__
'4.14.0'
>>> sherpa._version.get_versions()
{'version': '4.14.0', 'full': '53ea6bdd55a0c40cc83c7e2ad428d06d4eb58c25'}
>>>
```

whereas with this we get the following

```
>>> import sherpa
>>> sherpa.__version__
'4.14.0+15.gb9cb7a1d.dirty'
>>> sherpa._version.get_versions()
{'date': '2021-10-21T08:51:09-0400', 'dirty': True, 'error': None, 'full-revisionid': 'b9cb7a1df78f49d174d9cbd59e40e506fc64e3bf', 'version': '4.14.0+15.gb9cb7a1d.dirty'}
>>>
```

The extra fields are not an issue - since it's a dictionary', but the change from `full` to `full-revisionid` could **maybe** be an issue. (it turns out we actualy made us if this field when building the docs, so wahay for having the docs build as part of our tests to find this!).


DougBurke:
FYI @hamogu - I wanted to get this in part because it would make it easier to switch to setuptools-scm in the future if we decided to (primarily the removal of the additions we had made to the versioneer code).

Marie-Terrell (approved):
Thank you! This looks good. I ran the version change through our various build types. This does actually cleanup the lingering missing version info in the case of setup.py install without setup.py build first when working from a clone. I have prepped some CIAO patch updates to go along with the changes here.

Marie-Terrell:
By the way, to add a quick note (does not preclude this change going in): The sherpa-test-data repo will still have the older versioneer until that is updated separately.

DougBurke:
I admit to not having any knowledge of the `sherpa-test-data` repo. So that also has a `setup.py` file in it? I guess so, as it is technically installable.

Marie-Terrell:
Yes. Though, it should be a quick change there because the install is much more simple than here.

PR #1329, opened 2021-10-21
Build updates (move towards PEP-517 support)

### LINKED ISSUE #1545: installing sherpa; interesting version differences
With the main branch I have run the following (in a build where XSPEC support is not selected)

```
pip install .
cd ..
sherpa_smoke
```

with both python 3.9 and 3.10, and the 3.9 version fails with

```
AttributeError: module 'sherpa.astro.xspec' has no attribute 'XSAdditiveModel'
```

It is similar with `sherpa_test` (except the error there is some strange error about `utils`, but I think it's essentially because of this error).

If we look at the installed distribution we have, for Python 3.9

```
% ls $CONDA_PREFIX/lib/python3.9/site-packages/sherpa/astro/xspec
tests
% ls $CONDA_PREFIX/lib/python3.9/site-packages/sherpa/astro/xspec/tests/
__pycache__
```

(it's interesting that the `tests` directory exists but is empty) but for Python 3.10 the directory does not exist

```
% ls $CONDA_PREFIX/lib/python3.10/site-packages/sherpa/astro/xspec
ls: cannot access '/home/dburke/miniconda3/envs/sherpa-work/lib/python3.10/site-packages/sherpa/astro/xspec': No such file or directory
```

Now, if we try to "load" the xspec module then we get a surprising success with Python 3.9 and the expected failure with Python 3.10:

```
% python -c 'import sherpa.astro.xspec'
%
```

```
% python -c 'import sherpa.astro.xspec'
ModuleNotFoundError: No module named 'sherpa.astro.xspec'
```

If I install with `python setup.py install` then the xspec directory does not exist in either case, and things work fine.

So, I'm assuming that something in setuptools/distutils has changed over time and it changes wether the `sherpa/astro/xspec/tests/` directory gets installed or not. My best guess is that because we list it in `MAINFEST.in` indiretly or in `setup.py` via the `packages` and/or `package_data` options we somehow get it copied over.

I noticed this when re-working the PEP-517 PR #1329 but unfortunately it still seems to happen there too.

DougBurke:
# Summary

Move the Sherpa build system towards a more static configuration (PEP 517)

# Details

October 22 2021 saw the release of setuptools 58.3.0 where the `install` method was marked as deprecated - https://setuptools.pypa.io/en/latest/history.html#v58-3-0 - so there's "more urgency" to this. **Added April 14 2022** I've been involved in a discussion on how to build C extensions thread at https://discuss.python.org/t/modern-way-to-build-packages-with-c-cython-extensions/15050 which - at the time of writing - suggests that staying with `python setup.py develop` is reasonable for us (from one respondent).

This work was based on @hamogu's heroic #784 (but doesn't get to some of the good bits from that work, such as supporting tox or switching to `setuptools_scm`; they can come later). It has seen numerous rebases and reworks as I've tried to get parts of this work merged.

The main changes are:

- perform some code cleanup and changes to address newer standards and build systems
  - the code in `helpers/` has been re-arranged so that `helpers.extensions.build_ext` is a bit easier to use (it combines the "grab the settings" and the "build the actual extension" steps) and this happened because I realized we were including the wcs libs/headers when building the region code which is not needed (at least in the non-CIAO build)

- bump to a minimum version of 49.1.2 for setuptools as it is in this release that they include the `distutils` code (which is technically needed to build nicely with >= python 3.10 but it's not obvious if it's required). As part of this we now extend the `numpy.distutils.command.sdist` command rather than `distutils.command.sdist` - this "feels" right but it could have unintended consequences and I can't know remember whether it was needed or not (it's been really hard to track down what any one change does to the build system...)

- there are then two MANIFEST changes - the first one should be uncontroversial but the second one is a bit unclear to me as I am currently reading about how unclear all this is:
  - https://discuss.python.org/t/simple-documentation-for-setup-cfg/11465
  - https://github.com/pypa/setuptools/pull/2835
  
- the next commit is `Move basic options to setup.cfg` which moves over much of the information from `setup.py` to `setup.cfg`, and removes some of the repetition with the use of the `packages = find:` option. Note that we'll come back to that in a commit or two...

- aha: `setup.cfg: explicit about packages and tests` means I go back to being explicit about the packages and tests. This technically should not be needed but I was seeing some odd behavior and being explicit is "not necessarily bad" and potentially helps out with things like the settings queried by xspec_config.

- back to code removal: `Simplify setup.py` removes some code that was to support building when you might not have setuptools - but we now require it. I also ensure `setuptools` is imported in `setup.py` as I was seeing "odd" issues with this - this may not be necessary with setuptools >=49.1.2 (or perhaps later) but I do know that setuptools does "fun" stuff with imports related to distutils. At this point I think you can still say 'python setup.py develop` but I don't want to bother worrying about that.

- in an earlier commit I changed the `clean` command to allow it to work with setuptools >= 49.1.2, but we know that setuptools wants to remove such commands so in `Remove the clean command from setup` I have removed the clean command. I have never used the `clean` command, as I find `git clean -dfx .` to be  "good enough" (and probably better than the current `clean` command). For reference
  - https://github.com/pypa/setuptools/discussions/2838
  - https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

- there are some more replacements of distutils functionality - it's a bit hit or miss on what seems to work, but that's in part because we are trying to use an old setuptools because of some work done in #1458

- looking at the setuptools documentation - in particular https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#options - I noted that the `entry_points` option was listed as added in version 51, so we now require this as a minimum version rather than 49.1.2. Now, it's not obvious to me that this is needed, since I could not obviously trigger problems with an older setuptools version, but I'm not sure it really matters (as version 51.0.0 of setuptools was released December 2020).

- In looking at https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#options I noted that the `data_files` option was marked as deprecated, so I then tried to address this. Unfortunately a combination of the documentation they point to https://setuptools.pypa.io/en/latest/userguide/datafiles.html and our own code made this a surprisingly-involved exploration, even though the actual code change is small. I'm relying on the fact that things still build and the tests still run to validate this change. As I note below, this leads to looking at issue #50, as this is one place we care about data files and locations and the like, but I think that can be looked at separately from this PR.

- there is then a change to basically use `super` rather than the explicit class name. This is mainly a code-style change but there is technically a change in behavior in the announcing of building the stack and group libraries during a develop build - we now only report things when they are actually changed, and separate versions rather than a combined version - but I think these messages aren't seen by default anyway.

- sdist change: I noted earlier that I did not know how to replace the numpy.distutils command `get_data_files` but my claim is that we don't need to, as I could not trigger this bit of code to file. I've replaced the method with something that falls over (rather than removing the method) so we will find out if we do actually use it. However, our CI run doesn't trigger it, and manual calls to `python setup.py sdist` doesn't seem to either.

- sdist: revert the setuptools change. If you ran `python setup.py sdist` then you got a warning from numpy that the setuptools `sdist` command could cause problems - and links to https://github.com/numpy/numpy/pull/7131 (found by @Marie-Terrell ). So I have reverted this change. However, see the `Notes - sdist` section below.

- in an attempt to fix #1545 I have re-implemented the change from #1506 to always include the XSPEC code and tests, even when XSPEC is not being built. The assumption has been that 'import sherpa.astro.xspec' would error out cleanly if XSPEC support was not present, but I was able, in #1545, get to a place where this was not true. I have no idea what causes this condition to happen, and it seems to depend on various versions of python/setuptools/..., but the fact that it can happen makes me worry about it. Now, with this change and an XSPEC-free build a call to 'import sherpa.astro.xspec' would trigger a confusing recursive-import message from `._xspec`, so we now catch this and add a related `XSPEC support is not enabled` import error to be more-understandable (but keep the original error which is needed to debug potential XSPEC library issues for when we do have XSPEC support enabled). 

- finally I switch the GH workflow calls to build with pip rather than directly call setup.py
  - note that we use the `pip --verbose` flag to get the build output included in the log. This is not technically needed, and makes going through the logs slower (more to download) but does give us access to information on how the build works, which I think is worth it.

# Notes

So, in the move to "declarative metadata"

- `setuptools 30.3.0` (Dec 2016) stated to use `setup.cfg` - see https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
- `setuptools 61.0.0` (March 2022) now uses `pyproject.toml` (but `setup.cfg` is still supported) since this is the now-agreed-upon place - see https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

and we are sticking with `pyproject < 60` to follow the NumPy advice as they update their build system - https://numpy.org/doc/stable/reference/distutils_status_migration.html . I think we don't require anything "fancy", so should be able to move away from using `numpy.distutils`, but I still haven't seen any "to build an extension that uses the NumPy C API you just need to do this sparkly thing" information out there.

The release of NumPy 1.23 now causes this warning to be displayed when you build Sherpa:

```
/home/dburke/sherpa/sherpa-pr/setup.py:28: DeprecationWarning: 

  `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
  of the deprecation of `distutils` itself. It will be removed for
  Python >= 3.12. For older Python versions it will remain present.
  It is recommended to use `setuptools < 60.0` for those Python versions.
  For more details, see:
    https://numpy.org/devdocs/reference/distutils_status_migration.html 


  from numpy.distutils.core import setup
```

Some of the changes, such as the `data_files` change, are related to issue #50 - it would be nice to address this, but it doesn't have to be in this PR.

## Notes - sdist

Even with going back to the `numpy sdist` I still see a warning about setuptools, so I just give up (maybe it's saying we should use distutils not numpy.distutils here?)

```
% python setup.py sdist > log
/home/dburke/sherpa/sherpa-pr/setup.py:28: DeprecationWarning: 

  `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
  of the deprecation of `distutils` itself. It will be removed for
  Python >= 3.12. For older Python versions it will remain present.
  It is recommended to use `setuptools < 60.0` for those Python versions.
  For more details, see:
    https://numpy.org/devdocs/reference/distutils_status_migration.html 


  from numpy.distutils.core import setup
/home/dburke/miniconda3/envs/sherpa-pr/lib/python3.9/site-packages/setuptools/_distutils/dist.py:987: UserWarning: 
`build_src` is being run, this may lead to missing
files in your sdist!  You want to use distutils.sdist
instead of the setuptools version:

    from distutils.command.sdist import sdist
    cmdclass={'sdist': sdist}"

See numpy's setup.py or gh-7131 for details.
  cmd_obj.run()
/home/dburke/miniconda3/envs/sherpa-pr/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
```

-- 

# References

- https://www.python.org/dev/peps/pep-0517/
- https://github.com/pypa/setuptools/issues/2220

# Configuration

We can see that `setup.cfg` now contains most of the configuration options that were in `setup.py` (the remaining fields are `version`, `ext_modules`, and `cmdclass`):

```
% bat setup.cfg
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: setup.cfg
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ [metadata]
   2   │ name = sherpa
   3   │ author = Smithsonian Astrophysical Observatory / Chandra X-Ray Center
   4   │ author_email = cxchelp@head.cfa.harvard.edu
   5   │ url = http://cxc.harvard.edu/sherpa/
   6   │ project_urls =
   7   │     Issue Tracker = https://github.com/sherpa/sherpa/issues
   8   │
   9   │ description = Modeling and fitting package for scientific data analysis
  10   │ long_description = file: README.md
  11   │ long_description_content_type = text/markdown
  12   │ license = GNU GPL v3
  13   │ license_files = LICENSE
  14   │ platforms = Linux, Mac OS X
  15   │ classifiers=
  16   │     Intended Audience :: Science/Research
  17   │     License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
  18   │     Programming Language :: C
  19   │     Programming Language :: Python :: 3.7
  20   │     Programming Language :: Python :: 3.8
  21   │     Programming Language :: Python :: 3.9
  22   │     Programming Language :: Python :: 3.10
  23   │     Programming Language :: Python :: Implementation :: CPython
  24   │     Topic :: Scientific/Engineering :: Astronomy
  25   │     Topic :: Scientific/Engineering :: Physics
  26   │
  27   │ [options]
  28   │ python_requires = ~=3.7
  29   │ zip_safe = False
  30   │ install_requires =
  31   │     numpy
  32   │
  33   │ tests_require =
  34   │     pytest-xvfb
  35   │     pytest>=5.0,!=5.23
  36   │
  37   │ # Be explicit as something doesn't seem quite right when using find.
  38   │ #
  39   │ # find sherpa -name __init__.py | sort | sed -e "s|/__init__.py||" -e "s|/|.|g"
  40   │ packages =
  41   │     sherpa.astro.datastack
  42   │     sherpa.astro
  43   │     sherpa.astro.io
  44   │     sherpa.astro.models
  45   │     sherpa.astro.optical
...
```

# Things you can now do

## PEP 517 builds

The following seem to work, but have seem limited testing (also, the builds used to fail but some changes we've done now mean that even with the `main` branch you can build a wheel; it's just that when you come to test the result the `main` branch may fail):

Build using a PEP 517 tool - e.g. [build](https://pypa-build.readthedocs.io/en/latest/)

```
% pip install build
% python -m build .
...
% pip install dist/sherpa-4.14.0+399.g3329a683-cp310-cp310-linux_x86_64.whl   # actual file name may be different ;-)
...
```

I tried changing `setup.cfg` to enable XSPEC support and used `python -m build .; pip install dist/.....whl` and it built and installed an XSPEC-capable Sherpa (at least once I fixed the `MANIFEST.in` file).

You can also try

```
% pip install --use-pep517 . --use-feature=in-tree-build
```

and then the following should work

```
% cd $HOME    # just to make sure you are out of the source directory
% sherpa_test
```

## how to test

This is now the suggested way to test an editable build (ie not an installed one):

```
% pip install -e .
% pytest
```

You can test an installed version (which is different to `python setup.py test`) by saying

```
% pytest --pyargs sherpa
```

although we also have the `sherpa_test` command that essentially does this for us.


.github/scripts/test.sh
```diff
@@ -10,32 +12,26 @@ if [ "`uname -s`" == "Darwin" ] ; then
...
-    codecov;
+    # With an install build we have trouble running the tests
+    # from the installation directory (unable to import sherpa.utils._utils)
+    # and it is not at all clear why this is, so work around it.
+    #
+    if [ ${ITYPE} == install ]; then
+      cd /home
+      # Possibly seeing issues with the coverage checks
+      # pytest --pyargs sherpa --cov sherpa || exit 1;
+      pytest --pyargs sherpa || exit 1;
```
harlantc, line 30: with my new changes, I don't think coverage being in a different place will be a problem. It corrects paths that are different than the installed. It'd be best to keep coverage checks in IMO.

Maybe drop most of this change?

.github/scripts/test.sh
```diff
@@ -10,32 +12,26 @@ if [ "`uname -s`" == "Darwin" ] ; then
...
+    else
+      pytest --cov sherpa || exit 1;
+      codecov;
```
harlantc, line 33:
```suggestion
```

.github/workflows/ci-conda-workflow.yml
```diff
@@ -104,23 +105,69 @@ jobs:
...
+        git submodule deinit -f .
+
+    - name: freeze
+      shell: bash
+      env:
+        TEST: ${{ matrix.test-data }}
+      run: |
+        source ${miniconda_loc}/etc/profile.d/conda.sh
+        conda activate build
+        pip install pytest-xvfb pytest-cov codecov
```
harlantc, line 160: When you rebase you'll want to drop the codecov install on the merge.
```suggestion
        pip install pytest-xvfb pytest-cov
```

.github/workflows/ci-pip-workflow.yml
```diff
@@ -28,67 +28,80 @@ jobs:
...
+    - name: Build Sherpa (install)
+      shell: bash
+      if: matrix.install-type == 'install'
+      run: |
+        pip install . --verbose
+
+    - name: freeze
+      shell: bash
+      run: |
+        pip install pytest-xvfb pytest-cov codecov
```
harlantc, line 81: Codecov will also need to be removed here
```suggestion
        pip install pytest-xvfb pytest-cov
```

harlantc (commented):
Still looking through, but here's a few codecov changes that will be in conflict now #1339 is in.

.github/scripts/test.sh
DougBurke: Yeah - I was waiting for #1339 to re-work this part!

DougBurke: I do like how with #1339 and this work we can make the GH actions a bit more declarative (i.e. move the logic out of a bunch of bash if statements)!

harlantc: Awesome, I just wanted to note it to make it easier to find later. And yeah, I'm seeing the benefits of declarative configuration over "dynamic"/"programmatic" or whatever it's called.

DougBurke:
@harlantc - I've now rebased this to build on your work in #1339. I've also removed the addition of tox (I'll make a new PR for that that is based on this PR) which makes this a little easier to review.

It's interesting to me that the 'GH: switch from setup.py to pip for installation' commit fails but after the next two commits (`use setuptools 49.1.2...` and `xspec_config: rework and add to bdist_wheel` - this one @Marie-Terrell has looked at before and it has consequences for the CIAO build) the tests pass.

harlantc (commented):
Things look good overall to me. I know there might be some ciao patches we'll need, but on the surface ciao and sherpa on conda seems to work out.

Not applicable to this PR, but I noticed the ci workflows use sherpa_test without cov commands, while test.sh does. Is there a reason those don't bother with code coverage?

DougBurke:
@harlanc - I've done some (well, doing some) more clean up work of the GitHub actions moving away from the test.sh file. It was going to be part of adding tox support, but it may well make sense to make it a separate PR for clarity. We can try and fix up the use (or not) of `--cov sherpa` in that PR: my assumption is that it's just a hold over from my initial work where there was some problem at some point with `sherpa_test` and coverage.

harlantc:
@DougBurke I see. I figured it was something of that nature, was more for my own curiosity. I agree the tox support would better fit in another PR, and maybe changing the cov options there would make more sense too. It's not really needed if the other tests do happen to cover the same code (with how I read them, it looks like the tests that don't produce coverage reports don't do anything extra compared to the ones that do).

DougBurke:
@harlantc - #1352 is built on this and extends the GH Action clean up. I think all the `sherpa_test` commands - including those from the pip runs - now include coverage calls but (as I've had to spend a lot of time hand-holding this PR to get it to pass) I could well have missed something,

DougBurke:
Rebased to pick up latest changes.

DougBurke:
Rebased as #1328 has been merged

DougBurke:
Rebased as #1504 has landed.

DougBurke:
Currently being re-worked as I've removed the logic that was in #1506 and have gone down some further clean-up paths.

DougBurke:
I thought I was finished, but after realising I may have to go back to `destutils.sdist` rather than the `numpy` version I bumped against

```
NotImplementedError: Apparently we do need sdist.add_defaults
```

so yay for me putting that easter egg in there....

DougBurke:
@Marie-Terrell I think I've finished with the rebase/rework now. I've left the commits in the order I had them, but me may want to rebase this to avoid the repeated changes to sdist before being merged (but I'm sure there's other issues to address in this PR!).

Marie-Terrell:
Minor change request:
Can you add back in the "if sys.version_info < (3, 7):" section in setup.py (unless there is is another reason we can't have this in)? This is still required if someone uses "python setup.py install" with an older, unsupported version of Python. 

Testing done:
* Sdist comparisons
* Checked for pip install and python setup.py install (pytest, sherpa_test, sherpa_smoke, version check, from clone vs sdist, editable install, and contents comparison)
* Checked if the build stops for unsupported Python versions 
* Checked the CIAO-install and CIAO Conda builds/tests with patch updates

Notes:
* I am seeing slight differences in the include/lib paths with references to the build area, but these do not seem to be causing any issues
* The "Libraries have been installed in:" locations seem to be flipped (with and without the "python#.#/site-packages" bit) in the stk install section (for all but Python 3.10), but this does not actually seem to impact the final install.
* The xspec tests are now always installed and skipped again.
* I still plan to use "python setup.py sdist" for now (as opposed to "python -m build --sdist ." or "python -m build ."). This is just due to some oddities between the existence and contents of some of the setup files in the final source distribution noted while testing this PR. 
* Sherpa does not prevent "python setup.py install" from trying to build with Python 3.6 due to the removal of the "if sys.version_info < (3, 7)" section in setup.py.
* All of the tests passed with xspec or listed and skipped xspec tests if it was not built against it. 
* All of the level 3 tool tests in CIAO that go with Sherpa Sherpa have the same results as the current CIAOX.
* Checked over the notes/script for building the CIAO version of Sherpa and those all still appear to apply (no changes needed).

DougBurke:
I've re-based as I wasn't sure whether we had any "build" changes needed in main.

I've put back the version check that @Marie-Terrell was asking for (I am personally less bothered about this as python 3.6 is now "old" and I wanted to remove as much cruft as possible but I don't see this as requirement).

I also added a minor clean-up of `helpers/deps/__init__.py` based on trying to understand #1548 (but they do not address this issue and I don't think it's a particularly important issue to look at).

DougBurke:
For @Marie-Terrell - as I mention in the details section of the PR, the XSPEC change is because of

> in an attempt to fix https://github.com/sherpa/sherpa/issues/1545 I have re-implemented the change from https://github.com/sherpa/sherpa/pull/1506 to always include the XSPEC code and tests, even when XSPEC is not being built. The assumption has been that 'import sherpa.astro.xspec' would error out cleanly if XSPEC support was not present, but I was able, in https://github.com/sherpa/sherpa/issues/1545, get to a place where this was not true. I have no idea what causes this condition to happen, and it seems to depend on various versions of python/setuptools/..., but the fact that it can happen makes me worry about it. Now, with this change and an XSPEC-free build a call to 'import sherpa.astro.xspec' would trigger a confusing recursive-import message from ._xspec, so we now catch this and add a related XSPEC support is not enabled import error to be more-understandable (but keep the original error which is needed to debug potential XSPEC library issues for when we do have XSPEC support enabled).

What I need to know if this is good, bad, or indifferent.

Marie-Terrell (approved):
For the rest of my review, see my comment from July 18th.
The xspec change isn't ideal, but I think since this catches a case with issues while still throwing a more explanatory error than what was seen before #1458 it is a great compromise (which makes it the best option currently).
Ran through a couple quick checks of the helpers/deps/__init__.py change just to confirm that nothing unexpected changed.
This looks good. Thank you!

setup.cfg
```diff
@@ -1,3 +1,128 @@
+[metadata]
+name = sherpa
+author = Smithsonian Astrophysical Observatory / Chandra X-Ray Center
+author_email = cxchelp@head.cfa.harvard.edu
+url = http://cxc.harvard.edu/sherpa/
```
wmclaugh, line 5:
```suggestion
url = https://cxc.harvard.edu/sherpa/
```

helpers/extensions/__init__.py
```diff
@@ -100,17 +101,61 @@ def build_xspec_ext(library_dirs, include_dirs, libraries, define_macros=None):
...
     return [library_dirs, include_dirs, libraries]
 
+
+def clean(xs):
+    "Remove all '' entries from xs, returning the new list."
+    return [x for x in xs if x != '']
+
+
+def build_ext(command, name, *args, **kwargs):
+    """Set up the requesites for building an extension.
```
wmclaugh, line 118:
```suggestion
    """Set up the requisites for building an extension.
```

setup.py
```diff
@@ -18,39 +18,36 @@
...
-# python_requires will stop pip, but also let users who are using
-# 'python setup.py develop'.
+# Ideally we would rely on the configuration in setup.cfg to avoid
+# this situation, but as we may have users who still use old systems
+# add an error message. As Sherpa releases occur the minimum-supported
+# Python version is going to incrose (beyond 3.7) but it is not really
+# worth spending a lot of time on the error message here.
+#
+# This is done before we load in any non-core modules to avoid people
+# installnig software that they can not use.
```
wmclaugh, line 31:
```suggestion
# installing software that they can not use.
```

setup.py
```diff
@@ -18,39 +18,36 @@
...
 
+import os
 import sys
 
-# python_requires will stop pip, but also let users who are using
-# 'python setup.py develop'.
+# Ideally we would rely on the configuration in setup.cfg to avoid
+# this situation, but as we may have users who still use old systems
+# add an error message. As Sherpa releases occur the minimum-supported
+# Python version is going to incrose (beyond 3.7) but it is not really
```
wmclaugh, line 27:
```suggestion
# Python version is going to increase (beyond 3.7) but it is not really
```

helpers/deps/__init__.py
```diff
@@ -18,34 +19,73 @@
...
+
+        counter = cpu_count() + 1
+        # cflags = '-fPIC'
+        opts = [
+            # "CFLAGS={cflags}",
+            f"-j{counter}"
+        ]
+        make("install", opts, check=True)
+
+        # Create the built file to indicate the dependecies have been
```
wmclaugh, line 88:
```suggestion
        # Create the built file to indicate the dependencies have been
```

DougBurke: @wmclaugh - let me know when you have finished so I don't accidentally add your fixes before you've finished. As that may have just happened here already....

PR #1330, opened 2021-10-21
WIP: minimize C++ code duplications

dtnguyen2:
# Note

This PR, once done (it only has been tested for ```XSPEC``` version 12.10.1), minimizes C++ code duplications.

### DIFF
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -29,6 +29,8 @@
 #include <sstream>
 #include <iostream>
 #include <algorithm>
+#include <stdexcept>
+
 #include "sherpa/fcmp.hh"
 
 // Prior to XSPEC 12.10.1, the table models were split into different
@@ -61,6 +63,10 @@ namespace sherpa { namespace astro { namespace xspec {
 
 typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
 typedef float FloatArrayType;
+typedef void (*XSpecFuncRef)( float* ear, int* ne, float* param, int* ifl, float* photar, float* photer );
+typedef void (*XSpecFuncVal)( float* ear, int ne, float* param, const char* filenm, int ifl, float* photar, float* photer );
+typedef void (*XSpecFuncDouble)( const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr );
+
 
 // Try and support the use of std::transform while still building
 // against C++-98 compilers.
@@ -146,8 +152,8 @@ template <typename CType, int ArrayType>
 static bool create_grid(const sherpa::Array<CType, ArrayType> &xlo,
 			const sherpa::Array<CType, ArrayType> &xhi,
 			std::vector<CType> &ear,
-			std::vector<int> &gaps_index,
-			std::vector<CType> &gaps_edges) {
+			std::vector<int> &gaps_index ) {
+  std::vector<CType> gaps_edges;
 
   int nelem = int( xlo.get_size() );
   if ( nelem < 2 ) {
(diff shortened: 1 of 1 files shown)


dtnguyen2:
Will have to ask CM where one can access ```XSPEC``` version other than 12.10.1 so the following code can be tested:

```C++
#ifdef XSPEC_12_10_1


{
...
}
#else

...
}

#endif
```

BTW,  all the tests pass if one the following section of the code is commented out:
 
```C
#ifdef INIT_XSPEC
	if ( EXIT_SUCCESS != INIT_XSPEC() )
          throw std::runtime_error("Unable to initialize XSpec");
#endif
```

Not sure if it's supposed to do anything

dtnguyen2:
OK let me know when we no longer have to worry about XSPEC version earlier than 12.10.1 so I can remove the ```#else``` code path

DougBurke:
The tests fail because of #1334 - can you rebase this so that we pick up #1336 ?

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -61,6 +63,10 @@ namespace sherpa { namespace astro { namespace xspec {
 
 typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
 typedef float FloatArrayType;
+typedef void (*XSpecFuncRef)( float* ear, int* ne, float* param, int* ifl, float* photar, float* photer );
+typedef void (*XSpecFuncVal)( float* ear, int ne, float* param, const char* filenm, int ifl, float* photar, float* photer );
+typedef void (*XSpecFuncDouble)( const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr );
+
```
DougBurke, line 69: So, XSPEC does define these, but you'd need something like the following untested code

```
#ifdef XSPEC_12_12_0
#include "XSFunctions/Utilities/funcType.h"
#else
#include "funcType.h"
#endif
```

which defines (not all of which we care about):

```
// standard XSPEC11 fortran 77 calls
        typedef void (xsf77Call) (const float* energyArray,
                                  const int& Nenergy,
                                  const float* parameterValues,
                                  const int& spectrumNumber,
                                  float* flux,
                                  float* fluxError);
// double precision fortran 77 calls
        typedef void (xsF77Call) (const double* energyArray,
                                  const int& Nenergy,
                                  const double* parameterValues,
                                  const int& spectrumNumber,
                                  double* flux,
                                  double* fluxError);
// C language calls using C arrays
        typedef void (xsccCall)   (const Real* energyArray,
                                   int Nenergy,
                                   const Real* parameterValues,
                                   int spectrumNumber,
                                   Real* flux,
                                   Real* fluxError,
                                   const char* initString);
// C++ call with initialization string passed.
        typedef void (XSCCall) (const RealArray& energyArray,
                                const RealArray& parameterValues,
                                int spectrumNumber,
                                RealArray& flux,
                                RealArray& fluxError,
                                const std::string& initString);
```

The main issue is the difference on whether arguments are labelled `const` or not.

DougBurke:
@dtnguyen2 - I've created a version of this PR which rebases the code (so we don't suffer from un-related build issues) and then tries to switch over to the XSPEC typedefs I mention in https://github.com/sherpa/sherpa/pull/1330#discussion_r773389022

[I didn't want to change this PR hence the new one]

DougBurke:
My main concern about this rewrite is the error handling. We have C++ style error handling (e.g. throwing std::runtime_error) and Python error handling (which is some form of calling a Python routine to set the error and then return NULL). I need to sit down with paper and pencil and just check that things work out.

dtnguyen2:
The code was changed from
```
PyObject* xspecmodelfct( PyObject* self, PyObject* args )
{

#ifdef INIT_XSPEC
        if ( EXIT_SUCCESS != INIT_XSPEC() )
          return NULL;
#endif
...
}
```
to
```
      class PyArgTupleBase {
      public:
        PyArgTupleBase( ) {
#ifdef INIT_XSPEC
        if ( EXIT_SUCCESS != INIT_XSPEC() )
          throw std::runtime_error("Unable to initialize XSpec");
#endif
        }
};
```
C++ constructor cannot return an error code so an exception must be thrown.  Previously, if the function ```INIT_XSPEC``` returns an error the user has no clue as to what is the failure from.

Another example is, the code was changed from:
```
        if ( !PyArg_ParseTuple( args, (char*)"O&O&|O&",
                        (converter)convert_to_contig_array< FloatArray >,
                        &pars,
                        (converter)convert_to_contig_array< DoubleArray >,
                        &xlo,
                        (converter)convert_to_contig_array< DoubleArray >,
                        &xhi ) )
          return NULL;
```
to
```
          if ( !PyArg_ParseTuple( args, (char*)"O&O&|O&",
                                  (converter)convert_to_contig_array< RealArray >,
                                  &pars,
                                  (converter)convert_to_contig_array< DoubleArray >,
                                  &xlo,
                                  (converter)convert_to_contig_array< DoubleArray >,
                                  &xhi ) )
            throw std::runtime_error("Error Parsing args");
```
etc....

The main entry points for the changes in the new code are in the following functions:

```
template <npy_intp NumPars, bool HasNorm, XSpecFuncRef XSpecFunc>
PyObject* xspecmodelfct( PyObject* self, PyObject* args ) {

  xspecModelFctF<float, FloatArray, XSpecFunc> xspec_model =
    xspecModelFctF<float, FloatArray, XSpecFunc>( args, NumPars, HasNorm );
  try {
    FloatArray result;
    xspec_model.eval( result );
    return result.return_new_ref();
  } catch(std::runtime_error& re) {
    PyErr_SetString( PyExc_TypeError, re.what().c_str() );
    return NULL;
  } catch(...) {
    // Even though the XSPEC model function is Fortran, it could call                   
    // C++ functions, so swallow exceptions here                                        
    PyErr_SetString( PyExc_ValueError, xspec_model.get_err_msg() );
    return NULL;
  }

}

template <npy_intp NumPars, bool HasNorm, XSpecFuncDouble XSpecFunc>
PyObject* xspecmodelfct_C( PyObject* self, PyObject* args ) {

  xspecModelFctC<double, DoubleArray, XSpecFunc> xspec_model =
    xspecModelFctC<double, DoubleArray, XSpecFunc>( args, NumPars, HasNorm );
  try {
    DoubleArray result;
    xspec_model.eval( result );
    return result.return_new_ref();
  } catch(std::runtime_error& re) {
    PyErr_SetString( PyExc_TypeError, re.what().c_str() );
    return NULL;
  } catch(...) {
    PyErr_SetString( PyExc_ValueError, xspec_model.get_err_msg() );
    return NULL;
  }

}
```

As you can see the ```std::runtime_error``` is caught, but I did not call 
```PyErr_SetString( PyExc_TypeError, re.what().c_str() );``` which should be added to alert the user.  Once the error is caught a value of ```NULL``` is returned which is what the old code used to do.

Here is a simple prototype showing that the exception can be caught even if it is thrown from the base class:
devel12-196: cat tmp.cc
```
#include <iostream>
#include <stdexcept>

class Base {
public:
  Base( ) {
    std::cout << "Base()\n";
    throw std::runtime_error("ouch");
  }
};

class Derived: public Base {
public:
  Derived( ) : Base( ) {
    std::cout << "Derived()\n";
  }
};

int main(void) {

  try {
    Derived derived;
  } catch( std::runtime_error& re ) {
    std::cerr << "caught " << re.what( ) << '\n';
    return 1;
  }
 
  return 0;
}

devel12-197: g++ tmp.cc
devel12-198: a.out
Base()
caught ouch
```

A more sophisticated ```try``` & ```catch``` block in the constructor would be:

devel12-200: cat fubar.cc
```
#include <iostream>
#include <stdexcept>

using namespace std;

class Base {
public:
  Base( ) {
    std::cout << "Base()\n";
    throw runtime_error("ouch");
  }
};

class Derived: public Base {
public:
  Derived( ) try : Base( ) {
    std::cout << "Derived()\n";
  } catch( std::runtime_error& re ) { 
    throw re;
  }
};

int main(void) {

  try {
    Derived derived;
  } catch( std::runtime_error& re ) {
    std::cerr << "caught " << re.what( ) << '\n';
    return 1;
  }
 
  return 0;
}
devel12-201: g++ fubar.cc
devel12-202: a.out
Base()
caught ouch

```
but this  is not really necessary.

I think that if one add the following line:
```
PyErr_SetString( PyExc_TypeError, re.what().c_str() );
```
in  the  following functions
```
PyObject* xspecmodelfct( PyObject* self, PyObject* args ) 
PyObject* xspecmodelfct_C( PyObject* self, PyObject* args )
```
and remove the redundant lines  (493 & 527) then I think we should be OK.  That is the following changes should be made:

(base) [dtn@devel12 simplify_psf]$ git diff sherpa/include/sherpa/astro/xspec_extension.hh
```
diff --git a/sherpa/include/sherpa/astro/xspec_extension.hh b/sherpa/include/sherpa/astro/xspec_extension.hh
index 8a98f149..2b922a93 100644
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -490,7 +490,6 @@ static bool create_output(int nbins, T &a, T &b) {
           if ( NumPars != npars ) {
             std::ostringstream err;
             err << "expected " << NumPars << " parameters, got " << npars;
-            PyErr_SetString( PyExc_TypeError, err.str().c_str() );
             throw std::runtime_error(err.str());
           }
           return;
@@ -524,7 +523,6 @@ static bool create_output(int nbins, T &a, T &b) {
           if ( NumPars != npars ) {
             std::ostringstream err;
             err << "expected " << NumPars << " parameters, got " << npars;
-            PyErr_SetString( PyExc_TypeError, err.str().c_str() );
             throw std::runtime_error(err.str());
           }
           return;
@@ -661,7 +659,6 @@ static bool create_output(int nbins, T &a, T &b) {
             std::ostringstream err;
             err << "flux array does not match the input grid: " << nelem
                 << " and " << fluxes.get_size();
-            PyErr_SetString( PyExc_TypeError, err.str().c_str() );
             throw std::runtime_error(err.str());
           }
 
@@ -835,6 +832,7 @@ PyObject* xspecmodelfct( PyObject* self, PyObject* args ) {
     xspec_model.eval( result );
     return result.return_new_ref();
   } catch(std::runtime_error& re) {
+    PyErr_SetString( PyExc_TypeError, re.what().c_str() );
     return NULL;
   } catch(...) {
     // Even though the XSPEC model function is Fortran, it could call
@@ -855,6 +853,7 @@ PyObject* xspecmodelfct_C( PyObject* self, PyObject* args ) {
     xspec_model.eval( result );
     return result.return_new_ref();
   } catch(std::runtime_error& re) {
+    PyErr_SetString( PyExc_TypeError, re.what().c_str() );
     return NULL;
   } catch(...) {
     PyErr_SetString( PyExc_ValueError, xspec_model.get_err_msg() );
```

dtnguyen2:
This PR has been rebased and re-appears as PR #1387

PR #1338, opened 2021-10-27
clean up some C/C++ code in the _psf.cc file

dtnguyen2:
# Note

Minor clean up of C/C++ code

### DIFF
--- a/sherpa/utils/src/_psf.cc
+++ b/sherpa/utils/src/_psf.cc
@@ -267,9 +267,7 @@ static bool same_size_arrays( npy_intp size1, npy_intp size2,
     err << "input array sizes do not match";
     if ( suffix )
       err << " " << suffix;
-    else
-      err << ", ";
-    err << name1 << ": " << size1 << " vs " << name2 << ": " << size2;
+    err << ", " << name1 << ": " << size1 << " vs " << name2 << ": " << size2;
     PyErr_SetString( PyExc_TypeError, err.str().c_str() );
     return false;
   } else
@@ -316,11 +314,11 @@ static PyObject* tcdPyData_convolve( tcdPyData* self, PyObject* args )
                                     1, std::multiplies<long>() );
   
   if( !same_size_arrays( source.get_size(), src_size,
-                         "source size", "source dim", "dimensions, " ) )
+                         "source size", "source dim", "dimensions" ) )
     return NULL;
   
   if( !same_size_arrays( kernel.get_size(), kern_size,
-                         "kernel size", "kernel dim",  "dimensions, " ) )
+                         "kernel size", "kernel dim",  "dimensions" ) )
     return NULL;
   
   const long nAxes = (long) dims_kern.get_size();
@@ -861,7 +859,7 @@ static PyObject* pad_data( PyObject* self, PyObject* args )
   }
 
   if ( !same_size_arrays( kernel.get_size(), size,
-                          "kernel size", "kernel dim", "dimensions, " ) )
+                          "kernel size", "kernel dim", "dimensions" ) )
     return NULL;
   
   npy_intp dims[1];
@@ -906,7 +904,7 @@ static PyObject* unpad_data( PyObject* self, PyObject* args )
   }
 
   if ( !same_size_arrays( kernel.get_size(), padsize,
-                          "kernel size",  "kernel dim",  "dimensions, " ) )
+                          "kernel size",  "kernel dim",  "dimensions" ) )
     return NULL;
   
   npy_intp dims[1];
--- a/sherpa/utils/src/_psf.cc
+++ b/sherpa/utils/src/_psf.cc
@@ -1,5 +1,6 @@
 // 
-//  Copyright (C) 2007, 2016, 2018, 2020  Smithsonian Astrophysical Observatory
+//  Copyright (C) 2007, 2016, 2018, 2020, 2021
+//       Smithsonian Astrophysical Observatory
 //
 //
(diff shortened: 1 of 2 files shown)


DougBurke:
I've just rebased this to pick up the work around in #1336

DougBurke (approved):
LGTM

sherpa/utils/src/_psf.cc
```diff
@@ -250,6 +259,18 @@ static int _convolve( tcdPyData* self, double* source, double* kernel,
   return EXIT_SUCCESS;
 }
 
+static bool same_size_arrays( npy_intp size1, npy_intp size2,
+                              const char *suffix, const char *name1,
+                              const char *name2 ) {
+  if ( size1 != size2 ) {
+    std::ostringstream err;
+    err << "input array sizes do not match" << suffix;
+    err << name1 << size1 << name2 << size2;
```
wmclaugh, line 268: The error should contain all of the formatting- all of the calls to same_size_array are passing in the variable names and formatting text via the suffix, nam1, name2 parameters.  For example:

     err << name1 << ": " << size1 << " vs " << name2 << ": " << size2;

and the arguments passed in as name1 and 2 in the references (line 298, 302, etc...) should only pass in the actual names. Doing so, centralizes the formatting to one location so that it is easier to change and also makes the variable names accurately reflect their purpose.

sherpa/utils/src/_psf.cc
```diff
@@ -273,48 +294,30 @@ static PyObject* tcdPyData_convolve( tcdPyData* self, PyObject* args )
...
     return NULL;
   
-  if( dims_src.get_size() != dims_kern.get_size() ) {
-    std::ostringstream err;
-    err << "input array sizes do not match, "
-	<< "dims_src: " << dims_src.get_size()
-	<< " vs dims_kern: " << dims_kern.get_size();
-    PyErr_SetString( PyExc_TypeError, err.str().c_str() );
+  if( !same_size_arrays( dims_src.get_size(), dims_kern.get_size(),
+                         ", ", "dims_src: ", " vs dims_kern: " ) )
```
wmclaugh, line 298: arguments to same_size_arrays() should not contain formatting (such spaces, colons, commas) those should be handled in the same_size_arrays error string generation. For example, you should be passing in "dims_src" instead of "dims_src: ". (This applies to calls to same_size_array throughout the file.)

sherpa/utils/src/_psf.cc
dtnguyen2: done

sherpa/utils/src/_psf.cc
```diff
@@ -250,6 +259,18 @@ static int _convolve( tcdPyData* self, double* source, double* kernel,
   return EXIT_SUCCESS;
 }
 
+static bool same_size_arrays( npy_intp size1, npy_intp size2,
+                              const char *suffix, const char *name1,
+                              const char *name2 ) {
+  if ( size1 != size2 ) {
+    std::ostringstream err;
+    err << "input array sizes do not match" << suffix;
+    err << name1 << ": " << size1 << name2 << ": " << size2;
```
wmclaugh, line 268: it looks like the 13 calls to this function all pass in " vs ..." as parameter name2. The " vs " segment should be part of the format string here and the name2 should be restricted to the column name

sherpa/utils/src/_psf.cc
```diff
@@ -273,48 +294,30 @@ static PyObject* tcdPyData_convolve( tcdPyData* self, PyObject* args )
...
     return NULL;
   
-  if( dims_src.get_size() != dims_kern.get_size() ) {
-    std::ostringstream err;
-    err << "input array sizes do not match, "
-	<< "dims_src: " << dims_src.get_size()
-	<< " vs dims_kern: " << dims_kern.get_size();
-    PyErr_SetString( PyExc_TypeError, err.str().c_str() );
+  if( !same_size_arrays( dims_src.get_size(), dims_kern.get_size(),
+                         ", ", "dims_src", " vs dims_kern" ) )
```
wmclaugh, line 298: the last parameter in the call to same_size_arrays(),  " vs dims_kern" should only be "dims_kern". The " vs " portion is formatting that should be handled in same_size_arrays. This comment applies to all 13 instances of calls to same_size_arrays in this PR.

DougBurke:
@dtnguyen2 - test_psf_lowlevel.py was added in PR #1332 - see https://github.com/sherpa/sherpa/pull/1338#issuecomment-953326775

It explicitly added tests for some of these messages, so it's not surprising they are now failing.

dtnguyen2:
What a mess, there were only supposed to be two changed files (utils/src/_psf.cc and utils/tests/test_psf_lowlevel.py)  only, but 27?

harlantc:
I think I see what happened, rebasing reversed the order of things. I'll look into fixing the commands I gave you and let you know. fortunately git is forgiving with these things.

DougBurke:
@dtnguyen2 - you can

```
git remove -v
```

which will list the name you have given to the https://github.com/sherpa/sherpa repository. If you have one and it's called bob then

```
git fetch bob
git log --onelibe bob/main..HEAD
```

should list a lot of changes, then

```
git rebase bob/main
git log --onelibe bob/main..HEAD
```

should remove all the extra commands. 

For @harlantc - the above has a 'git merge' in and not a rebase.

harlantc:
@DougBurke Sorry, was talking about some options I had given off to the side. I had Dan rebase, but it looks like the order got mixed up?

PR #1347, opened 2021-11-10
Refactored array classes cause std::vector has non-virtual destructor

dtnguyen2:
# Note

```std::vector``` has non-virtual destructor so in this PR the refactored  Array (1D and 2D) classes no longer inherit from ```std::vector```; Moreover, the boundary limits of the free parameters are now passed to the cost function which should make it easier to apply different boundary algorithms at the limits.

### CODE REVIEW

**Comment by DougBurke** [comment]
@dtnguyen2 - I'm  bit confused about the use of the NDEBUG symbol, but in trynig to work out what's going on I've realized that we do not appear to have any code that uses the `Array1D` class (which means we can remove the `#define` lines as well).

We do have code using `Array2D`.

**Comment by DougBurke** [comment]
This PR also removes some files in `extern/`. While I don't think we really care about those files, it'll make updating the FFTW code more annoying in the future so can you fix this?

**Comment by dtnguyen2** [comment]
@DougBurke - not sure what you mean when you say:

> ... we do not appear to have any code that uses the Array1D class ...

Not counting the ```Array1D``` instances in the definition file of _ include/sherpa/myArray.hh_, there are plenty of use of the class ```Array1D```

```
(base) [dtn@devel12 sherpa]$ pwd
/export/sherpa/sherpa
(base) [dtn@devel12 sherpa]$ grep -r Array1D | grep -v include/sherpa/myArray.hh | wc
    131     847   11206
```

```
(base) [dtn@devel12 src]$ pwd
/export/sherpa/sherpa/optmethods/src

(base) [dtn@devel12 src]$ awk 'NR > 26 && NR < 35' Simplex.hh
  template <typename T> class ParVal : public Array1D<T> {

    friend std::ostream &operator<<(std::ostream &s, const ParVal<T> &P) {
      return P.print(s);
    }

  public:
    ParVal() : Array1D<T>() {}
```

**Comment by DougBurke** [comment]
In my checkout, which is at the current main branch:

```
% git log --format=oneline | head
2616cc08285880c0b3b895462243a5f7964ba773 Merge #1030 (DougBurke) - Update lev3fft test to use PHYSICAL not WCS
991b915b5e8815c355aee0c69dafe2d0a5c89eb8 Merge #1317 (DougBurke) - Add requires group test decorators
cc65d99f49d03b333ee7bb34a9a76dff9b7cccf5 Merge#1326 (hamogu) - Write-up that environment varaible in setup.cfg need to be expanded
76bf4aadad6813850ce8217ffd4a136213a59e4d Merge #1339 (harlantc) - Add codecov.yml to fix path issue and update codecov uploader
54bcb3311d5d972654c2d39df08499d4cf031bc7 Create codecov.yml
9ca64657f7575ef822a725fdffb1b75ea2c7e2cf Docs: make the note an actual note
f557450c03ad234d0d7c70ce399e96f240a6e52d Tests add requires_group for a datastack test
d8ef00261f8698a2447aff1091b25159d20481c5 Add @requires_group for tests requring fits and/or data
46cd13b0e7c298659e8e42034fda8c981217c737 Add requires_group to a number of tests
ed099b81e2353c4c5b59c48511164d14c561a1b5 Merge #1318 (DougBurke) - Adjust a test for python 3.10
```

I get

No match in `Simplex.hh`:

```
% grep Array1D sherpa/optmethods/src/Simplex.hh
```

I don't see any `Array1D` in https://github.com/sherpa/sherpa/blob/main/sherpa/optmethods/src/Simplex.hh - or to be specific about the commit: https://github.com/sherpa/sherpa/blob/14f61c38bce70aa08acfcda9577bc3b6de857544/sherpa/optmethods/src/Simplex.hh

As I said, I see match in any other file other than `myArray.hh`:

```
% find . -type f -exec grep -l Array1D {} /dev/null \;
./sherpa/include/sherpa/myArray.hh
```

Before sending my comment I even removed the `Array1D` code from `myArray.hh` and Sherpa still built. So I don't see where these uses are!

**Comment by DougBurke** [comment]
I now see that the code that uses the Array1D class was added in this PR.

File: sherpa/include/sherpa/myArray.hh [hunk 1]
```diff
@@ -19,567 +16,538 @@
...
-	this->operator[]( ii ) /= rhs;
-      return *this;
-    }
+  virtual void sort() { std::sort(vec.begin(), vec.end()); }
 
-    std::ostream& myprint( std::ostream& os, int num=4, const char* prefix=NULL,
-			   const char* suffix=NULL ) const {
+  Array1D<T> &operator+=(const Array1D<T> &rhs) {
+    MYCMPSIZES(rhs);
+    for (int ii = 0; ii < rhs.size(); ++ii)
```
wmclaugh, line 116:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < rhs.size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
```diff
@@ -19,567 +16,538 @@
...
+  }
 
-      if ( this->size( ) > 0 ) {
-	if ( NULL != prefix )
-	  os << prefix << "[0]=";
-	os << this->operator[]( 0 );
-      }
+  Array1D<T> &operator-=(const Array1D<T> &rhs) {
+    MYCMPSIZES(rhs);
+    for (int ii = 0; ii < rhs.size(); ++ii)
```
wmclaugh, line 123:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < rhs.size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
```diff
@@ -19,567 +16,538 @@
...
-	  os << '\n';
-	else
-	  os << ", ";
-	if ( NULL != prefix )
-	  os << prefix << '[' << ii << "]=";
-	os << this->operator[]( ii );
-      }
+  Array1D<T> &operator*=(const Array1D<T> &rhs) {
+    MYCMPSIZES(rhs);
+    for (int ii = 0; ii < rhs.size(); ++ii)
```
wmclaugh, line 130:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < rhs.size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
```diff
@@ -19,567 +16,538 @@
...
+      vec[ii] /= rhs[ii];
+    return *this;
+  }
 
-      return os;
-    }
+  Array1D<T> &operator+=(T rhs) {
+    for (int ii = 0; ii < this->size(); ++ii)
```
wmclaugh, line 144:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < this->size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
```diff
@@ -19,567 +16,538 @@
...
+      vec[ii] += rhs;
+    return *this;
+  }
 
-  private:
+  Array1D<T> &operator-=(T rhs) {
+    for (int ii = 0; ii < this->size(); ++ii)
```
wmclaugh, line 150:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < this->size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
```diff
@@ -19,567 +16,538 @@
...
+      vec[ii] -= rhs;
+    return *this;
+  }
 
-    void check_size( const Array1D< Real >& arg ) const {
-      if ( this->size() != arg.size( ) )
-	throw std::runtime_error( "Operation attempted on different size arrays" );
-    }
+  Array1D<T> &operator*=(T rhs) {
+    for (int ii = 0; ii < this->size(); ++ii)
```
wmclaugh, line 156:
```suggestion
    for (std::vector<int>::size_type ii = 0; ii < this->size(); ++ii)
```

sherpa/include/sherpa/myArray.hh
DougBurke: I ask as a "not a C++ expert": should we be looking at re-writing these using `std::transform` (I think we can now assume we have this, but I can't say for sure), or is it okay to stick with the more-direct iteration here? The strong advantage in not changing the code is that there's less chance to introduce errors, but do we want to move to a more STL-like interface?

PR #1348, opened 2021-11-12
Minor internal code cleanup tracking errors

DougBurke:
# Summary

Minor internal changes to how some errors are raised. There should be no user-visible differences due to this changes.

# Details

For several cases where errors are caught and re-thrown we now use the `from` statement to say whether to link the original and new error: mostly these are `from None` since this simplifies the traceback, but in at least one case we do link errors (which also improves the traceback). This is only useful if you hit an error and are looking at the tracebaack!

In doing the above I noted we had a routine whose sole purpose was to raise an error. This made it hard to do the `from None` case and whilst this could have been addressed, I felt it was cleared to just raise the error directly - it does not significantly increase the code and, I believe, actually makes it a little easier to follow (since you had to know that the original routine raised an error when reading the code).

There's a couple of unused routines (one is now unused because of the earlier changes) that have been removed, and some minor code-style changes (e.g. moving a check on whether an argument is valid to the start of the routine).

After creating these changes I ran them through CI to see what lines were untested and then added tests to cover those cases.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
@@ -47,8 +47,7 @@ from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
 def test_load_table_model_fails_with_dir(tmp_path):
     """Check that the function fails with invalid input: directory
 
-    The temporary directory is used for this (the test is skipped if
-    it does not exist).
+    The temporary directory is used for this.
     """
 
     tmpdir = tmp_path / 'load_table_model'
@@ -67,8 +66,7 @@ def test_load_table_model_fails_with_dir(tmp_path):
 def test_load_xstable_model_fails_with_dir(tmp_path):
     """Check that the function fails with invalid input: directory
 
-    The temporary directory is used for this (the test is skipped if
-    it does not exist).
+    The temporary directory is used for this.
     """
 
     tmpdir = tmp_path / 'load_xstable_model'
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016, 2018  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2016, 2018, 2021  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


DougBurke:
I'm interested to see what code paths I've changed don't have test cases!

sherpa/ui/tests/test_session.py
```diff
@@ -579,3 +581,317 @@ def test_error_estimate_not_run(name):
...
+
+def test_paramprompt_default():
+    """The default setting is False."""
+
+    s = Session()
+    assert not s._paramprompt
+
+
+@pytest.mark.parametrize("flag", [True, "yes", "yeS", 1, "1", "on", "ON"])
+def test_paramprompt_set_true(flag):
```
hamogu, line 603: Fine. I don think that @DougBurke is overdoing it here though. Do we really 7 tests for this? I know they will be fast, so I'm not asking to revert, but surely if "yeS" passes then  "yes" will pass, too.

hamogu: Note: It is of course possible to make "yeS" pass and "yes" fail, but I can't see any way that could happen accidentally and tests don't protect us against malicious acters who craft the most obscure code possible to introduce such subtle bugs into the code on purpose.

sherpa/ui/tests/test_ui_unit.py
```diff
@@ -122,3 +125,83 @@ def test_save_filter_data1d(tmp_path, clean_ui):
...
+
+
+def test_set_iter_method_type_not_enumeration():
+    with pytest.raises(TypeError) as te:
+        ui.set_iter_method('a random string')
+
+    assert str(te.value) == "a random string is not an iterative fitting method"
+
+
+class NonIterableObject:
```
hamogu, line 144: I sometimes use `None` for that  purpose, but on the other hand `None` has a special meaning in much of sherpa, so this is safer.

hamogu (approved):
Approving (only stylistic comments on docstrings).
I admire the work that went into something like this; I wish we could do without the parampromt stuff entirely and just make users use the Python interface (but I know we can't); and I hope that code spelunking that let's @DougBurke into these depth is at some point done...

sherpa/astro/ui/tests/test_astro_ui_utils_unit.py
```diff
@@ -48,38 +44,40 @@
...
     """
 
+    tmpdir = tmp_path / 'load_table_model'
+    tmpdir.mkdir()
+
     ui.clean()
     assert ui.list_model_components() == []
     with pytest.raises(IOError):
-        ui.load_table_model('tmpdir', tmpdir)
+        ui.load_table_model('tmpdir', str(tmpdir))
```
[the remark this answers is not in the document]
DougBurke: One thing I'm thinking of is if it's worth updating the IO routines to support `pathlib.Path` objects.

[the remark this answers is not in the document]
hamogu: I suggest that we do it if (a) we touch a routine anyway and (b) it's not too much work. For example, astropy support for pathlib objects is spotty (I don't know on top of my head if `astropy.io.fits` is one of the modules that works or not) and thus the pyfits backend might be a challenge in some situations. So, it's not really worth to put a lot of effort into on the sherpa side, if our backends can't deal with it.

sherpa/ui/tests/test_session.py
DougBurke: So, the "boolean" check is added here in part for some potential future change when we might change the logic to be more-Pythonic (the existing logic comes from historical, including - but also pre-dating -S-Lang support)

hamogu (approved):

PR #1351, opened 2021-11-15
Update the sherpa-test-data repo to the latest from main

Marie-Terrell:
It looks like with the merge of #1030 we now have an outdated data reference. This just updates that reference to the latest copy. This should resolve the new regression test failures.


hamogu (approved):
I was just starting to look how my XMM files had disappeared, but @Marie-Terrell beat me in putting them back!

PR #1353, opened 2021-11-18
Fix issue with delete_model_component

### LINKED ISSUE #16: delete_model_component failure when component is part of a source expression
I found this when updating the documentation to match the CIAO version, and it happens in both CIAO and the released version.

The warning message that `delete_model_component` should create when the input model component is part of a source expression does not appear, instead there's an error thrown:
`KeyError: 1`. The `1` comes from `list_model_ids`, and is used to find the source expressions to check. Something in the internals has changed, which makes this check fail, and throw an unrelated, and unhelpful-to-the-user error. As discussed below, it also leaves the system in a slightly-confused state.

```
... hidden warning messages ...

In [2]: ui.load_arrays(1, [1,2,3], [4,5,6])

In [3]: ui.set_model(ui.const1d.mdl)

In [4]: print(mdl)
const1d.mdl
   mdl.c0       thawed            1 -3.40282e+38  3.40282e+38           

In [5]: ui.delete_model_component('mdl')
KeyError                                  Traceback (most recent call last)
<ipython-input-5-3ef4f808567f> in <module>()
----> 1 ui.delete_model_component('mdl')

<string> in delete_model_component(name)

/home/naridge/data/my-sherpa/ltest/lib/python2.7/site-packages/sherpa/ui/utils.pyc in delete_model_component(self, name)
   4318         if mod is not None:
   4319             for key in self.list_model_ids():
-> 4320                 if mod in self._models[key] or mod in self._sources[key]:
   4321                     warning("the model component '%s' is found in model %s" %
   4322                             (mod.name, str(key) + " and cannot be deleted" ))

KeyError: 1

```

The line numbers will not match up since this comes from my own documentation fork.

```
In [6]: ui.list_model_ids()
Out[6]: [1]
```

Looking at the code you can see that the "removal" of the component is done before the warning message is displayed/model restored, which - in this case - leads to the component actually disappearing. Perhaps the logic should be changed to only do the deletion after the check?

```
In [7]: ui.list_model_components()
Out[7]: []
```

I expected this to still return `['mdl']`.

DougBurke:
# Summary

The delete_model_component call could fail in certain circumstances, so fix those cases. Fixes #16 

# Details

The first commit is to add tests to check the existing code. The next commit fixes #16, which is just a problem with the code not recognizing that a key may not exist. The replacement code does slightly more work than it needs to do (i.e. if all the logic were in the if statement it could take advantage of lazy evaluation) but this is not a time critical routine and ease-of-reading is a benefit here.

### DIFF
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -912,7 +912,7 @@ def test_delete_model_component_invalid_argument():
 
 
 def test_delete_model_component_not_a_component():
-    """We could allow the component to be deleted this way"""
+    """Check correct error message for non-exitant model"""
 
     s = Session()
     with pytest.raises(IdentifierErr) as te:
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -895,3 +895,82 @@ def test_paramprompt_eof(caplog):
     assert cpt2.ampl.val == pytest.approx(1)
     assert cpt2.ampl.min < -3e38
     assert cpt2.ampl.max > 3e38
+
+
+def test_delete_model_component_invalid_argument():
+    """We could allow the component to be deleted this way"""
+
+    s = Session()
+    s._add_model_types(sherpa.models.basic)
+
+    tst = s.create_model_component('gauss1d', 'gmdl')
+
+    with pytest.raises(ArgumentTypeErr) as te:
+        s.delete_model_component(tst)
+
+    assert str(te.value) == "'name' must be a string"
+
+
+def test_delete_model_component_not_a_component():
+    """We could allow the component to be deleted this way"""
+
+    s = Session()
+    with pytest.raises(IdentifierErr) as te:
+        s.delete_model_component('tst')
+
+    assert str(te.value) == "model component 'tst' does not exist"
+
+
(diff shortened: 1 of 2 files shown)


hamogu (changes requested):
Wow. A two digit issue to fix!
Just one tiny change to a comment (I don't write into people's branches without talking to them in general, but this is a case where I conceivably could accept my own comment and then approve this. However, absent a consensus that that's OK, I'm just leaving a comment and "requesting changes").

sherpa/ui/tests/test_session.py
```diff
@@ -895,3 +895,80 @@ def test_paramprompt_eof(caplog):
     assert cpt2.ampl.val == pytest.approx(1)
     assert cpt2.ampl.min < -3e38
     assert cpt2.ampl.max > 3e38
+
+
+def test_delete_model_component_invalid_argument():
+    """We could allow the component to be deleted this way"""
```
hamogu, line 901: I think we should. Could be in the PR, but probably better goes into a new one, so we might want to open an issue?

DougBurke: Done - #1357

hamogu (approved):

dtnguyen2 (approved):

PR #1355, opened 2021-11-19
Minor cleanup of the I/O code

DougBurke:
# Summary

Minor internal clean-up of the I/O code. There should be no user-visible changes.

# Details

There's a number of changes, mainly pointed out by pylint. This was originally meant to be a fix for #47 but these changes are worth taking AND the fix for #47 is likely to be larger than I originally expected.

There's technically one user-facing change, as there are two places where

```python
    try:
        something-complex()
    except:
        ignore-this-error()
```

has been changed to only catch `Exception` errors. Ideally we'd restrict further - e.g. to `OSError` - but without careful vetting I can't be sure this is safe. I strongly  believe that restricting to `Exception` is safe, as this only stops "non-Python" errors from being caught - like control-c or OOM cases. I have seen cases where control-c has not been noticed by Sherpa thanks to bits of (old) code like this, which is not ideal.

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019, 2021
-#     Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -54,11 +54,11 @@ References
 """
 
 from configparser import ConfigParser
+import importlib
 import logging
 import os
 import os.path
 import sys
-import importlib
 
 import numpy
 
@@ -86,8 +86,8 @@ else:
            "it must be None or a float > 0"
     try:
         ogip_emin = float(ogip_emin)
-    except ValueError:
-        raise ValueError(emsg)
+    except ValueError as ve:
+        raise ValueError(emsg) from ve
 
     if ogip_emin <= 0.0:
         raise ValueError(emsg)
@@ -111,10 +111,6 @@ __all__ = ('backend',
            'read_table', 'read_image', 'read_arf', 'read_rmf', 'read_arrays',
            'read_pha', 'write_image', 'write_pha', 'write_table',
            'pack_table', 'pack_image', 'pack_pha', 'read_table_blocks')
-
-
-def _is_subclass(t1, t2):
-    return isinstance(t1, type) and issubclass(t1, t2) and (t1 is not t2)
 
 
 # Note: write_arrays is not included in __all__, so don't add to the
@@ -167,7 +163,7 @@ def read_arrays(*args):
         raise IOErr('noarrays')
 
     dstype = Data1D
-    if _is_subclass(args[-1], BaseData):
+    if sherpa.io._is_subclass(args[-1], BaseData):
         dstype = args.pop()
 
     args = backend.get_column_data(*args)
@@ -228,7 +224,9 @@ def read_table(arg, ncols=2, colkeys=None, dstype=Data1D):
     >>> d = read_table('tbl.fits', colkeys=['WLEN', 'FLUX', 'FLUXERR'])
 
     """
-    colnames, cols, name, hdr = backend.get_table_data(arg, ncols, colkeys)
+    args = backend.get_table_data(arg, ncols, colkeys)
+    cols = args[1]
+    name = args[2]
 
     # Determine max number of args for dataset constructor
     sherpa.io._check_args(len(cols), dstype)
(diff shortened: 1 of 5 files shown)


DougBurke:
I'm leaving this as a draft as I want to see how the coverage comes up, as I may need to add some test cases.

DougBurke:
So, there's two un-tested warning lines in the pyfits backend - one of them looks really hard to trigger and the other would require partially populating a WCS header which I'm not sure it's worth doing.

Anything in the crates backend is not checked - this would need something like #1069 to fix

sherpa/astro/io/__init__.py
```diff
@@ -167,7 +163,7 @@ def read_arrays(*args):
         raise IOErr('noarrays')
 
     dstype = Data1D
-    if _is_subclass(args[-1], BaseData):
+    if sherpa.io._is_subclass(args[-1], BaseData):
```
hamogu, line 166: I believe 
```suggestion
    if is_subclass(args[-1], BaseData):
```
would be sufficient. I mean, is it really the task of this function to check that it's not `BaseData` itself? If args[-1] is BaseData and it work, fine, if not, BaseData will throw an exception. But this is fine, too.

sherpa/astro/io/__init__.py
```diff
@@ -228,7 +224,9 @@ def read_table(arg, ncols=2, colkeys=None, dstype=Data1D):
     >>> d = read_table('tbl.fits', colkeys=['WLEN', 'FLUX', 'FLUXERR'])
 
     """
-    colnames, cols, name, hdr = backend.get_table_data(arg, ncols, colkeys)
+    args = backend.get_table_data(arg, ncols, colkeys)
+    cols = args[1]
+    name = args[2]
```
hamogu, line 229: I prefer the old version for readability, but that's personal taste, so I won't argue either way.

sherpa/astro/io/crates_backend.py
```diff
@@ -441,8 +431,6 @@ def read_table_blocks(arg, make_copy=False):
         for name in names:
             cols[ii][name] = crate.get_column(name).values
 
-    if close_dataset:
-        close_crate_dataset(dataset)
```
hamogu, line 445: Why do we not need this any longer? I'm probably missing something obvious, but usually I try to avoid touching anything crate related with a twn-foot-pole...

sherpa/astro/io/__init__.py
DougBurke: Well, pylint didn't like the unused colnames and hdr variable settings. I could have gone with the `_, cols, name, _ = ...` trick, but then it feels like I'm writing Haskell. We're also a ways off getting destructuring in the language (Python 3.10).

sherpa/astro/io/crates_backend.py
DougBurke: In crates there was a memory leak/problem where the solution was to add the snip method to the dataset. `close_crate_dataset` then called this method, if it existed. I can't remember what version of CIAO removed this work around, but it was a while ago...

Aha - CIAO 4.5 removed the need for it (it's at the end of the https://cxc.cfa.harvard.edu/ciao/releasenotes/ciao_4.5_release.html#DataModel right before chips). It looks like we never officially documented when it was taken out.

sherpa/astro/io/__init__.py
DougBurke: Going forward to your "trying not to touch code ..." comment, I have not looked at our various 
is an instance of" routines and compared them to standard routines. In this case it's

```python
def _is_subclass(t1, t2):
    return isinstance(t1, type) and issubclass(t1, t2) and (t1 is not t2)
```

so it does something more than `issubclass`. Although **why** it does these - surely the `isinstance` check is essentially a no-op (this comes from the old days, written in Python 2 times, so maybe it had some use earlier). In fact, it looks like the important thing is the `t1 is not t2` check. We don't want people to be sending in `BaseData`, but to be honest if they do so then things aren't going to work out well for them anyway.

@olaurino left `BaseData` in when he re-worked the data hierarchy in #614 - leaving behind the following

```python
class BaseData(metaclass=ABCMeta):
    """
    Base class for all data classes. Left for compatibility with older versions.
    """
    pass
```

So I think here we are okay in changing to `issubclass`. Maybe it's time to remove `BaseData`, but that would be a different PR ;-)

As an aside, `sherpa.ui.utils` has it's own variant

```python
def _is_subclass(t1, t2):
    return inspect.isclass(t1) and issubclass(t1, t2) and (t1 is not t2)
```

so an alternative for this PR would be to remove `sherpa.io._is_subclass` and replace by `sherpa.ui.utils._is_subclass` (since this PR was trying to reduce the number of times we re-write the same code), but maybe we should just use `is_subclass` there too (but again in another PR).

DougBurke: Hmm. Actually I think there are a number of CIAO and/or L3 scripts that could rely on `BaseData`, so there would be a cost to changing this. But again, for a different PR.

hamogu: Right, I agree that is does more - but I think that "more" does not have to be done *here* but, I agree, not for this PR.

hamogu (approved):
OK, my questions are resolved. Approving.

Issue #347, opened 2017-04-06
add plot support for image data sets


DougBurke:
There are times when it would be good to be able to display image data using the plot backend (matplotlib or ChIPS) rather than DS9 (e.g. in scripts or ipython notebooks). I think this should not be part of the `image_XXX` set of routines, instead either be part of the `plot_XXX` routines, or a new set of routines.

I've put together a proof-of-concept (limited to matplotlib) at https://gist.github.com/DougBurke/5dbaf4398dcc38356abc4bcb1637a71d

This code is not intended for inclusion in Sherpa but just to try this out.

DougBurke:
So, we have

    plot_xxx()
    contour_xxx()
    image_xxx()

and the matching

    plot()
    contour()

commands. We could 

a) add image support to the `plot_xxx` support (adding functionality to the plot backend code to display images)

b) we could add a new series of commands such as

    iplot_data()
    iplot_model()
    iplot_fit()

c) we could add an "image" backend so that you would use `image_data()` and get either DS9 or "matplotlib-style plots"

I am not keen on option c as I think the user interface of the `image_xxx` routines is sufficiently different to a plot that it  would be jaring, and I also could see the use in being able to use both DS9 and <whatever command we come up with> at the same time.

I am leaning to option a as it would be significantly-less work for the implementer, but would welcome thoughts from the community.

anetasie:
@DougBurke could you give an example of option a? would this be something like `plot_image()` or just `image()` to match the other two? and would be just `image(data....)`?

DougBurke:
No, we would use the same commands in option a, so

    plot_data()
    plot_model()
    plot_fit()
    plot_resid()

would create (with matplotlib) plots which match

    image_data()
    image_model()
    image_fit()
    image_resid()

but without the ability to resize/zoom/change scaling or colormaps/add or remove regions that you can in the image_xxx commands.

anetasie:
@DougBurke but these already exist and create images in ds9.  What happens if you have both ds9 and matplotlib in backend?

DougBurke:
Err, no they don't:

```python
sherpa In [1]: load_image('img.fits')

sherpa In [2]: plot_data()
AttributeError: 'DataIMG' object has no attribute 'to_plot'

```

anetasie:
I guess I have to tried the proof-of-concept examples.

DougBurke:
I haven't looked at the proof-of-concept code for a long time so I'm not sure if it worls (or what it actual;y does)

PR #1356, opened 2021-11-20
Add bokeh as a backend

hamogu:
# Status
This would profit from some real-live use to root out any problems before I'm finalizing the tests.

# Summary
This PR adds a [bokeh](https://bokeh.org/) plotting backend. Bokeh is a Python library that generates js output to be displayed in the browser. Those plots are interactive by default (e.g. zoom) and many more features are available (linking different panels, switching individual lines in the plot on and off, anything in js that we implement, e.g. dynamic binning in the plot even without a running python session). Interactive HTML/js plots can be embedded into AJ/ApJ articles, used in websites, interactive posters etc. 

The goal of this PR is simply to reproduce the main features of our matplotlib interface, advanced features can be added later. **Update**: In order to keep PRs managable and have them merged timely, I've decided to put "docs for how to write a backend" into a separate PR.

# Details

This is built on top of #1191 - testing and developing this is much easier with a way to switch the backend dynamically. At the same time, this acts as a test to #1191. It's all good to say that you've cleanly separated backends and backend-independent plotting code, but only with a second, fully functional backend you really know that that's true.

Give that I'm writing a second backend here, I'll also take the opportunity to write down the specs/API of how to do that.

Missing
- [ ] bokeh tests
- [x] narrative docs
- [ ] an example notebook


DougBurke:
:1st_place_medal:

DougBurke:
So, I haven't had the time to look at this, so it may be obvious from the commits, but our existing `_repr_html` code for Sherpa objects - so that they can display "nicely" in notebooks - have a number of places that check if the matplotlib backend is selected and, if so, returns a plot / image rather than a table. This sometimes takes advantage of the sherpa plot hierarchy - in which case we would want the code to just check that the backend is not the dummy backend, but in some places it just uses direct access to matplotlib functionality. I would assume these can be done with bokeh, so I was wondering whether you'd got to look at this code, or whether the notebook changes you talk about are really showing that `plot_data()` is now nicely interactive?

An example from a notebook with a PHA file loaded would be comparing

plot_data()

in one cell with the output of

get_data()

in another cell.

hamogu:
I have not tried all possible cases (arf, rmf, ...), but here is one:

(sorry my screen is not large enough to show both plots in full at the same time)
Screen Shot 2021-12-03 at 1 56 27 PM

From what I have seen, the _repr_html_ generally works by calling XXX_as_html in the backend and the bokeh backend just implements those functions. (Although, in the interest of decoupling backends from plots, it would be useful to have a more general way of expressing that we we don't have to add one `as_html` function per lpottype to each backend. But that's a separate improvement.

DougBurke:
Ta. I was just looking at the notebook tests (for #1374) and it reminded me that we should check things now we are moving to a glorious multi-plot-backend future.

Examples that I know are matplotlib-specific are RMF and image objects. The RMF version could be done via sherpa plots but it would miss some things (partly related to #938 but that is still backend-specific) and the image support would need #347 addressed.

For reference:

RMF plot: https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/astro/data.py#L685-L750

image plot: https://github.com/sherpa/sherpa/blob/a436ccf2478c83fa8628f556726d25de9f729ded/sherpa/astro/data.py#L753-L828

hamogu:
Good point. I think the rmf plot can be made backend agnostic and a display for image objects #347 would be useful in a number of cases. Once that is added, it does not have to be special-cased here. I'm just trying to spin my hear around how to do all that and still have PRs that are small enough to actually review in practice. For once, I'll create a new project to track these. In our last meeting we discussed this to a lever that made clear that is is going to be more than a few isolated PRs. (And a project and easily be removed again if DS disagrees with that way of tracking.)

DougBurke:
I do think the notebook changes are somewhat orthgonal to the "add-a-new-backend" changes, but they are part of the same project (particularly in terms of ensuring the backend abstraction is not-leaky).

hamogu:
Yes, there are a number of orthogonal changes, but then they are also connected in some way. 
So, my goal is to next have a PR that defines the interface for a backend, by writing down a specification and offering a test module to test for "leakiness" in some cases. In practice, that's possible, but unnecessarily hard to do without using this PR (use a second functional backend to see what happens), so I'll probably have to rebase or adapt this a couple of times.

hamogu:
While this is fully functional (despite needing rebasing now), I'm converting it to a draft since it's part of the larger plotting system overhaul that has not yet been approved in full.

DougBurke:
So, as part of looking at #1382 (via the rebased #1550) I've been playing around this morning with adding an [altair](https://altair-viz.github.io/) backend, because the declarative model used by altair - which is basically the [Vega-Lite  schema](https://vega.github.io/vega-lite/) - is likely to break the assumptions that the existing backend assumes:

- how the visualization is created (e.g. the passing around of "objects" that create the plot and then are used to display the thing)
- creation of the plot elements

Although it is also a JS backend it is somewhat different to bokeh, so I'm interested to see the backend here currently uses "global state" to create the output (at some level this is always going to be the case, and it is just how well encapsulated it is).

My guess is that altair is not a good choice for sherpa, and I don't think we want to contort our API to make it work, but just wanted to mention this somewhere.

hamogu:
Sherpa assumes that at plotting backend is stateful, i.e. you can add stuff one by one and the backend keeps track what the "active axis" is and stuff like that. `matplotllib.pyplot` does that, but bokeh does not, that's why this PR has lines like this https://github.com/sherpa/sherpa/pull/1356/files#diff-dbf59ad6ac809e580e0a31e3c990a87ab1692c7c08d3f4474784889235cddaeeR137 to emulate that behavior. Altair needs, at the lest, something similar. I think it could be made work if the plotting module "caches" previous plotting commands in a module-level variable and then combines them in some suitable way.

DougBurke:
Yeah - I'm just thinking about #1729 and whether there's things we can do now to make things workable for "most" backends (but not making it so abstract that it's too hard to design and implement that we never do it). At the moment I quite like the approach in #1729 **but** maybe it's not worth it?

hamogu:
closing in favor of #1899.

PR #1358, opened 2021-11-22
Remove `point` plotting in backend

hamogu:
# Summary
Remove the `point` method in the definition of the plotting backend.

# Details
The definition of the `point` method in the backend was inconsistent with the other plotting methods in argument (e.g. it lacked parameters such as title, xlabel, ...) and in parameter names (what was called symbol here, is called marker
elsewhere).
The `point` method is just a thin wrapper around `plot` itself - `plot` supports scatter plots with symbols, so it can be removed.

For users, this is an API change in that `point_prefs` attribute of the `Confidence2D` plot object now has an entry `marker` instead of `symbol`, so anyone who manually set the `symbol` in that dictionary will have to update their code to set the `marker` instead. (If this sounds esoteric, that's because it is. I doubt that anyone outside the dev group has ever set `symbol` in the `point_prefs` manually.)

# Context
In the context of #1356, I'm trying to understand how plotting backends are
defined and to write down a formal API that we can test existing and future
backends against. I'm also hoping to be able to automate some of the code
translating plotting options from sherpa to different backends (e.g. in Sherpa
`+` is a maker, while in bokeh the same marker is called `"plus"`) and the fact
that the calling signature of `point` was so different from `plot`, `histogram`,
and friends is that problem for that. Since `point` is only a warpper around `plot`,
removing the entire method simplifies the definition of plotting backends at
no cost.

PRs #1191 and #1356 will have to be rebased on this and modified (for #1356 this is a simplification), so it would be great to make a decision on this soon.

# Test
We do not test pixel-by-pixel outcomes, so, for reference, here is an example plot that looks exactly the same before and after applying this PR:
```
import numpy as np
from sherpa.astro import ui
x = np.arange(100)
ui.load_arrays('test', x, x + np.random.rand(100))
ui.set_source('test', ui.polynom1d('poly'))
poly.c1.frozen=False
ui.fit('test')
ui.reg_proj(poly.c0, poly.c1)
```
image

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -88,7 +88,7 @@ backend.init()
 
 plotter = backend
 
-__all__ = ('Plot', 'Contour', 'Point', 'Histogram',
+__all__ = ('Plot', 'Contour', 'Histogram',
            'HistogramPlot', 'DataHistogramPlot',
            'ModelHistogramPlot', 'SourceHistogramPlot',
            'PDFPlot', 'CDFPlot', 'LRHistogram',
@@ -2711,9 +2711,9 @@ class Confidence2D(DataContour, Point):
 
         # Note: the user arguments are not applied to the point
         #
-        point = Point()
-        point.point_prefs = self.point_prefs
-        point.point(self.parval0, self.parval1,
+        point = Plot()
+        point.plot_prefs = self.point_prefs
+        point.plot(self.parval0, self.parval1,
                     overplot=True, clearwindow=False)
 
         if self.log[0]:
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -308,7 +308,7 @@ class Plot(NoNewAttributesAfterInit):
 
         Parameters
         ----------
-        x, y
+        x, y : array-like (scalar, list, or array)
            The data values to plot. They should have the same length.
         yerr, xerr : optional
            The symmetric errors to apply to the y or x values, or `None`
--- a/sherpa/plot/pylab_backend.py
+++ b/sherpa/plot/pylab_backend.py
@@ -30,7 +30,7 @@ from sherpa.utils import formatting
 from sherpa.plot.utils import histogram_line
 
 
-__all__ = ('clear_window', 'point', 'plot', 'histo', 'contour', 'set_subplot',
+__all__ = ('clear_window', 'plot', 'histo', 'contour', 'set_subplot',
            'init', 'get_split_plot_defaults', 'get_plot_defaults', 'begin',
            'end', 'get_data_plot_defaults', 'get_model_plot_defaults',
            'exceptions', 'get_fit_plot_defaults', 'get_resid_plot_defaults',
@@ -191,20 +191,6 @@ def find_zorder(axes):
         return max(zs) + 0.1
 
     return None
-
-
-def point(x, y, overplot=True, clearwindow=False,
-          symbol=None, alpha=None,
-          color=None):
-
-    axes = setup_axes(overplot, clearwindow)
-
-    if color is None:
-        style = '{}'.format(symbol)
-    else:
-        style = '{}{}'.format(color, symbol)
-
-    axes.plot(numpy.array([x]), numpy.array([y]), style, alpha=alpha)
 
 
 def histo(xlo, xhi, y, yerr=None, title=None, xlabel=None, ylabel=None,
@@ -586,7 +572,7 @@ def get_plot_defaults():
 
 
 def get_point_defaults():
-    return get_keyword_defaults(point, 2)
+    return {'marker': None, 'alpha': None, 'color': None}
 
 
 def get_histo_defaults():
@@ -595,7 +581,7 @@ def get_histo_defaults():
 
 def get_confid_point_defaults():
     d = get_point_defaults()
-    d['symbol'] = '+'
+    d['marker'] = '+'
     return d
 
 


sherpa/plot/pylab_backend.py
```diff
@@ -586,7 +572,7 @@ def get_plot_defaults():
 
 
 def get_point_defaults():
-    return get_keyword_defaults(point, 2)
+    return {'marker': None, 'alpha': None, 'color': None}
```
hamogu, line 575: There are other places where we also hardcode definitions instead of pulling them all from `get_keyword_defaults`, so this is not unusual in the context of this backend.
(And, I think it's beneficial to get away from the `get_keyword_defaults` in general, because that risks polluting the general plot objects with backend-specific keywords, which makes backend switching hard, but that is a larger change not part of this PR.)

sherpa/plot/__init__.py
```diff
@@ -2711,9 +2711,9 @@ def contour(self, overplot=False, clearwindow=True, **kwargs):
 
         # Note: the user arguments are not applied to the point
         #
-        point = Point()
-        point.point_prefs = self.point_prefs
-        point.point(self.parval0, self.parval1,
+        point = Plot()
```
DougBurke, line 2714: So `Plot` is clever enough to recognize it's been sent scalar values for x and y and internally converts it into an array. Or does this rely on matplotlib's handlling of scalar values here (in which case we'd want to ensure `Plot` handles scalar values sensibly otherwise it depends on the backend).

hamogu: We will require `plot` in the backend to accept scalars or arrays (but documenting and testing those requirements is a different PR that I have not yet opened), so the onus is on the backend module. Both matplotlib and bokeh accept scalars out of the box with no extra work needed. But if we ever implement a backend module that does not, then the `crazy_backend.plot` code would have to take care of converting it to lists or arrays or whatever the datastructure for `crazy_backend` is.

sherpa/plot/pylab_backend.py
DougBurke: For background info and not directly related to this PR:

Currently the `sherpa.ui.utils.Session` and `sherpa.astro.ui.utils.Session` class (which subclasses the former) - and are the logic between the `sherpa.ui`/`sherpa.astro.ui` modules - create plot objects

- when they are created
-  when their `clean` method is called

We could make it lazy - we only create the actual object when it is first needed - but there is still the issue to decide what the `Session` class should do when the plot backend is changed.

The session approach relies on access to these plot objects - and persisting those objects between calls - so unless the persistence is removed (which I am likely to argue against) it is hard to say we can always just create a new plot object each time (and hence pick up the new plot backend).

Now, I think we could argue that users of the session class (either directly or indirectly) are unlikely to swap plot backends, but I can imagine that I might want to swap between `matplotlib` and `bokeh` depending on whether I'm in a notebook or not. It also makes testing easier if we have a story for this case.

sherpa/plot/__init__.py
DougBurke: I think I wanted to surface whether this was already, or is going to be, something that could rely on the backend or is something we support (i.e. sending in a scalar value to `Plot`). I think it should be the latter. Given that you are changing this in the PR I think there should be s very simple change to the `Plot.plot` docs along the lines of changing

```
    def plot(self, x, y, yerr=None, xerr=None, title=None, xlabel=None,
             ylabel=None, overplot=False, clearwindow=True,
             **kwargs):
        """Plot the data.

        Parameters
        ----------
        x, y
           The data values to plot. They should have the same length.
        yerr, xerr : optional
           The symmetric errors to apply to the y or x values, or `None`
           for no errors along the axis. If given, they should match the
           length of the data.
```

to mention `x`/`y` can be scalars (and `xerr`/`yerr`).

sherpa/plot/pylab_backend.py
hamogu: The goal in #1191 is that make the plots objects work with any backend, so that we would not have to re-create the plot objects, just because the backend is changed; instead the backend is hot-swapable (if that's a word). Plot objects only call "backend.plot", but that "backend" is not a variable held in the object, it's a module level variable. So, you can change it and "backend" would suddenly refer to something else (say bokeh instead of matplotlib). One can think of it analogous to like unix pipes: The plot objects pipes into `backend.histo` and does not need to care what exactly that next step (`backend.histo`) does with the data that's send to it. In that senses, it's somewhat similar to monkey-patching, but not as dangerous (in monkey-patching I  would be changing methods of an existing object, while here I'm simply changing which object is getting the plotting command). So, I would hope that #1191 takes care of this concern. If it does not, that's an incompleteness of #1191 and I should add commits to that.

(With one important exception: Our default plot styling will only use options that all implemented backend understand such as "linestyle="solid", but it is possible to manually set preferences that are backend-specific. Those won't work when the backend is switched. But that's a design feature, not a bug. I will also supply, in the separate PR, a way to validate that all examples and all default plot settings in Sherpa limit itself to a well-defined set of options that all implemented backends understand.)

sherpa/plot/__init__.py
hamogu: Note: I also will (in a later PR) synchronize doctrings between these objects (for example, many functions in `pylab_backend` have incomplete docstrings right now), but that's no excuse to not do this properly here already. 
Done in last commit.

hamogu:
Converted to draft because this is just one of the larger plotting changes where the full concept has not yet been approved.

hamogu:
This is included in #1382, so I'm closing this PR.

PR #1360, opened 2021-11-22
Add missing export statement to clang 12 note

adonath:
This PR adds the missing `export` statement to the installation note on clang 12. After the export the `pip install sherpa` worked successfully on my M1 system.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -391,7 +391,7 @@ C++ compilers to use if not found by ``setup.py``.
    declaration. If so, pre-pending the following to your pip install
    or python setup.py install line should resolve the build issues::
 
-     export CFLAGS='-Wno-implicit-function-declaration'
+     CFLAGS='-Wno-implicit-function-declaration' pip install .
 
    (Note that on MacOS, "gcc" is usually just an alias to Clang. Unless
    you specifically install a gcc from a different source, you are
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -391,7 +391,7 @@ C++ compilers to use if not found by ``setup.py``.
    declaration. If so, pre-pending the following to your pip install
    or python setup.py install line should resolve the build issues::
 
-     CFLAGS='-Wno-implicit-function-declaration'
+     export CFLAGS='-Wno-implicit-function-declaration'
 
    (Note that on MacOS, "gcc" is usually just an alias to Clang. Unless
    you specifically install a gcc from a different source, you are


hamogu:
I believe that you can either export (or "setenv") for C-Shell users OR prepend onto the same line to just set the varaible for that one line, soetihng like (details may change depending on if you use bash or tcsh or something else):
```
CFLAGS='-Wno-implicit-function-declaration' pip install .
```
I like your solution, too, though.
Note that we don't really need to run the full test suite for this change, but I approved it anyway, because github will then remember and automatically run the CI next time you open a PR (https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

hamogu (approved):

adonath:
Thanks @hamogu! I actually prefer the solution you suggested, however it was not immediately clear to me. Probably I read it in the context of the warning before. I'd be happy to change to a more explicit:

```
CFLAGS='-Wno-implicit-function-declaration' pip install .
CFLAGS='-Wno-implicit-function-declaration' pip install sherpa
```
As well...

hamogu:
Both has advantages and disadvantages. Despite what I said, I usually set it as you suggestedso that I can simply run "pip isstall -e ." later (In development, when I run with "-e" I might need to run this often and I know I will forget the CFLAGS again later...). Anyway, I'm find you your suggestion, I'm fine with your new suggestion, too. If you as a new developer, think this was unclear, then we should improve it, so we should leave it to your judgement how to do that best. @DougBurke and I are not longer good in spotting these things, we've done it too often.

adonath:
> I can simply run "pip install -e ." later...

That's actually a fair point to keep the variable set.

wmclaugh:
The issue itself is fixed by adding '-Wno-implicit-function-declaration' to the CFLAGS used when the pip install command is run.  I would advocate that we do not recommend setting the CFLAGS env var as it may pollute user/developer environments or risk overwriting existing environment settings beyond the necessity of setting up the instruction to pip install sherpa (Pre-pending the override limits the scope of the change.).  It also removes dealing with providing examples of setting up env vars in other shells (ie.  export vs set env).  

 As the original text in lines 389-394

      If you are building with Clang version 12, you will
      encounter build issues in the region area related to an implicit
      declaration. If so, pre-pending the following to your pip install
      or python setup.py install line should resolve the build issues::

     CFLAGS='-Wno-implicit-function-declaration'

indicates that the flag should be **prepended** to the pip install command, I would recommend either: (1) leave the instructions as  they are written and close out this issue or (2) changing the text starting at line 391 to be more explicit:

        If you are building with Clang version 12, you may
        encounter build issues in the region area related to an implicit
        declaration. If so, pre-pending the  '-Wno-implicit-function-declaration' 
        C compiler option  to your pip install  or python setup.py install line should
        resolve the build issues. For example::

            CFLAGS='-Wno-implicit-function-declaration' pip install .

docs/install.rst
```diff
@@ -391,7 +391,7 @@ C++ compilers to use if not found by ``setup.py``.
    declaration. If so, pre-pending the following to your pip install
    or python setup.py install line should resolve the build issues::
 
-     CFLAGS='-Wno-implicit-function-declaration'
+     export CFLAGS='-Wno-implicit-function-declaration'
```
hamogu, line 394:
```suggestion
     CFLAGS='-Wno-implicit-function-declaration' pip install .
```

hamogu:
I think we've discussed with one-word change long enough. I've taken the freedom to take @wmclaugh suggestion, to which we can all agree (I think, based on the discussion above) and taken the freedom to push it directly to @adonath 's branch (since he allowed "edits by maintainers" or whatever the button is called). 
@wmclaugh : So this should definitely be good now.

@adonath Thank you for pointing out that the instructions were not as clear for new-comers as they could have been - and I hope you take the length of the discussion as a sign how serious we take your suggestion and won't be discouraged to let us know if you run into any other things that could be improved.

adonath:
Thanks @hamogu and @wmclaugh, the minimal change is already helpful!

PR #1364, opened 2021-11-26
WIP rework PR #617

dtnguyen2:
#Note This PR is basically PR #617 (but without the test script for now) so @anetasie can start testing

### DIFF
--- a/sherpa/sim/mhtest.py
+++ b/sherpa/sim/mhtest.py
@@ -1,0 +1,589 @@
+import numpy as np
+
+from sherpa.sim.mh import dmvnorm
+from sherpa.utils.err import ParameterErr
+from sherpa.astro.io import io_opt
+from sherpa.utils import NoNewAttributesAfterInit, print_fields
+from sherpa.stats import Cash, CStat, WStat
+from sherpa.astro import ui
+
+if io_opt == 'crates_backend':
+    from pycrates import read_file, get_keyval, copy_colvals, get_col_names
+else:
+    from astropy.io import fits
+
+__all__ = ('mhTest',)
+
+class PCA:
+
+    fields = ('filename', 'emethod', 'cmethod', 'errext', 'colnames',)
+    
+    def __init__(self, filename):
+        self.filename = filename
+        self.emethod = None
+        self.cmethod = None
+        self.errext = None
+        self.bias = None
+        self.colnames = None
+        self.component = None
+        self.fvariance = None
+        self.eigenval = None
+        self.eigenvec = None
+        return
+
+    def __str__(self):
+        print('bias = ', self.bias)
+        print('component = ', self.component)
+        print('fvariance = ', self.fvariance)
+        print('eigenval = ', self.eigenval)
+        print('eigenvec = ', self.eigenvec.shape, self.eigenvec)
+        return print_fields(self.fields, vars(self))
+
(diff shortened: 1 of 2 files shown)


sherpa/sim/mhtest.py
```diff
@@ -0,0 +1,589 @@
...
+        self.bias = None
+        self.colnames = None
+        self.component = None
+        self.fvariance = None
+        self.eigenval = None
+        self.eigenvec = None
+        return
+
+    def __str__(self):
+        print('bias = ', self.bias)
```
DougBurke, line 35: there shouldn't be print calls in a __str__ methof

sherpa/sim/mhtest.py
```diff
@@ -0,0 +1,589 @@
...
+from sherpa.sim.mh import dmvnorm
+from sherpa.utils.err import ParameterErr
+from sherpa.astro.io import io_opt
+from sherpa.utils import NoNewAttributesAfterInit, print_fields
+from sherpa.stats import Cash, CStat, WStat
+from sherpa.astro import ui
+
+if io_opt == 'crates_backend':
+    from pycrates import read_file, get_keyval, copy_colvals, get_col_names
+else:
```
DougBurke, line 12: this needs to be something like

```
elif io_opt == "pyfits_backend":
    from astropy.io import fits
else:
    do we want to error out here or just create a warning message?
```

Ideally we'd be using the IO library to access the data rather than call through to the backend directly. I think we can do this with sherpa.astro.io.read_table_blocks (and hence get rid of this and the backend-specific PCA sub-classes below.

It should be something like

```
# note: no error checking
indata = sherpa.astro.io.read_table_blocks(filename)
hdr = indata[2][1]

mypca = PCA(infile)
mypca.emethod = hdr["EMETHOD"]
mypca.cmethod = hdr["CMETHOD"]
mypca.errext = hdr["ERREXT"]
mypca.bias = indata[1][1]["BIAS"]

block = block[1][2]
mypca.colnames = list(block.keys())  # this should work
mypca.component = block["COMPONENT"]
mypca.fvariance = block["FVARIANCE"]
mypca.eigenval = block["EIGENVAL"]
mypca.eigenvec = block["EIGENVEC"]
```

anetasie:
@dtnguyen2 and @DougBurke 
The code that have been already placed in Sherpa to deal with the calibration uncertainties in included in this directory: https://github.com/sherpa/sherpa/tree/main/sherpa/astro/sim

These two files `fullbayes.py` and `pragbayes.py` are used if `set_sampler` is set to `pragbayes` or `fullbayes`.  It is not clear if there is a difference between these functions and the new code.  The ARF file I/O  is set as an option `opt=simarf` for these samplers. It is not clear if that's the best solution, but current setup works. Here is part of my notebook:
https://github.com/anetasie/SherpaNotebooks/blob/master/Bayes.uncertainty-Test.ipynb

The relevant part is here:
image

There are a few things that needs to be done:
(1) check the difference between the code in this PR and the existing pragbayes and fullbayes functions. Do we need the new function or can we fold in the functionality in this PR into the existing code?
(2) understand why there is a failure with changes to the iteration numbers

hamogu:
Keeping this PR open for now, but might be easier to re-do than to continue working form this state given existing conflicts and unresolved implementation questions.

Issue #1369, opened 2021-12-02
should we use black (and enforce) for code formatting


DougBurke:
I actually am not a great fan of the black formatting style, but it is

- consistent
- easily enforced

Using a code formatter (and perhaps also the "I forget the name but it re-orders imports" [*] and stripping trailing spaces, but I think black does this for us) has the advantage that we don't get commits which add a lot of noise due to different preferences (we've had at least one such case recently). black is well used

For reference: https://pypi.org/project/black/

[*] we have some "interesting" import usage so maybe we would not want to do this

DougBurke:
As a reminder to myself, here's example settings for setting up `black` and the like 

https://gist.github.com/tdhopper/ae09efe739247029e096e6dbdd891544

hamogu:
Similar: Might help to run https://github.com/Carreau/velin over our code (and could also be done as a CI check, but I don't think it's worth the effort).

harlantc:
An option would be to put it in the CI pipeline ala https://github.com/marketplace/actions/super-linter, I see they have black under the python linter. I'm not sure what else this would need.

hamogu:
I want to support the suggestion to use black. It's opinionated, so no flamewars over "set this option to "X"; we can just take the defaults. Would be nice to never think about how to format again.
For simplicity, we would have to make one big commit to apply it to all python files in `sherpa/sherpa` (not C code, not `extra`, etc.) and then we can set our editors and hooks to apply formatting.
It's possible to ignore that formatting commit for the purposes of `git blame` (https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) and to mark specific sections of carefully handcrafted code lines (e.g. long formulas broken according to the structure of the math) to not be touched by black: https://stackoverflow.com/questions/58584413/black-formatter-ignore-specific-multi-line-code

Issue #1362, opened 2021-11-23
requires_fits test decorator and the new dummy backend support


DougBurke:
Since #1185 we can now have in the configuration file

```
[options]
io_pkg = dummy
```

If you do this **and** have either/both of `astropy` or `crates` installed then those tests labelled as `requires_fits` are likely to fail because

- the decorator looks to see if it can import either `astropy.io.fits` or `crates` and finds it can, so runs the test
- the code ends up calling the `sherpa.astro.io.dummy_backend` version which is going to fall over

This was found when testing out #1361 but is completely unrelated to that (other than I added a brief comment to `requires_fits` to point out this issue).

So, long-term we need to decide what to do in situations like this: examples include

- error out of the tests early, with some sort of "unsupported use case" error [not friendly]
- tweak `requires_fits` so that it checks what the io_pkg setting is [not ideal as we may have changed things during testing, but maybe okay]
- change the I/O tests to (now we can) switch the I/O backend to the supported backends for each such test, no-matter the `options.io_pkg` setting

My current thinking is that the middle option is the sensible approach, but I haven't given it much thought. I also don't see this as a high priority at this time (the `dummy` value is new, so few people will be using it, and it's also unlikely to be chosen).

DougBurke:
I should point out that we're likely to hit the same issue with `requires_plotting` once #1191 is merged.

hamogu:
I think the proper way of handling that is to have a context manager that sets a backend of your choice. That is useful in general (e.g. to make the next plot with that other backend without changing the backend in general), but also for testing. Some tests may only run for a specific backend (e.g. bokeh specific tests should only be run with bokeh), for other tests we might want to iterate over backends (e.g. ensure that both crates and pyfits can read all files).

PR #1370, opened 2021-12-02
Minor tweaks to handling of PHA settings

DougBurke:
# Summary

Improve the checking of the factor and rate settings of set_analysis and the plot_fac attribute of the DataPHA class, and add a few tests that exercise some corner cases of the DataPHA class.

# Details

The plot_fac value was being silently converted to an integer - so `pha.plot_fac = 0.5` or `set_analysis("energy", factor=0.5)` would set `plot_fac` to 0. We now error out it a non-integer-like value is used (so 2.0 is okay, as it was before, but "`" is no-longer supported). I would rather error out than make the code behave "strangely" for the user. An alternative would be to just warn when the value has been truncated, but I am guessing that this is not-likely to be happening in user code.

The type argument to set_analysis was strange - it would allow any string which, when converted to lower-case, began with "rate" or "counts" - **but** would only set the rate option if `str(type).lower() == "rate"`. This is not a good UX. So we now require the value to be either "counts" or "rate" (and no case conversion or stripping of spaces). The existing documentation lists these as the only two valid arguments.

I threw in some minor "style" changes for `get_areascal/get_backscal`.

As well as testing these changes I've added a few other tests to catch DataPHA code that we weren't testing (edge cases).

There are a few style changes to the tests

- we can now safely import sherpa.astro.io thanks to #1185, and some "standardisation" of checks on the backend - maybe something we want to add to `conftest.py` in a later PR related to #1362
- use of pytest.approx rather than numpy.assert_allclose
- swapping from ' to " to mark a string, following the black standards, to see how this works, related to #1369

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1663,10 +1663,10 @@ class DataPHA(Data1D):
 
         if (rmf is None and arf is None) and \
            (self.bin_lo is None and self.bin_hi is None) and \
-           quantity != 'channel':
-            raise DataErr('norsp', self.name)
+           quantity != "channel":
+            raise DataErr("norsp", self.name)
 
-        if rmf is None and arf is not None and quantity != 'channel' and \
+        if rmf is None and arf is not None and quantity != "channel" and \
            len(arf.energ_lo) != len(self.channel):
             raise DataErr("incompleteresp", self.name)
 
@@ -2478,10 +2478,10 @@ class DataPHA(Data1D):
         7.8504301607718007e-06
 
         """
-        backscal = self.backscal
-        if backscal is not None:
-            backscal = self._check_scale(backscal, group, filter)
-        return backscal
+        if self.backscal is None:
+            return None
+
+        return self._check_scale(self.backscal, group, filter)
 
     def get_areascal(self, group=True, filter=False):
         """Return the fractional area factor of the PHA data set.
@@ -2522,10 +2522,10 @@ class DataPHA(Data1D):
         1.0
 
         """
-        areascal = self.areascal
-        if areascal is not None:
-            areascal = self._check_scale(areascal, group, filter)
-        return areascal
+        if self.areascal is None:
+            return None
+
+        return self._check_scale(self.areascal, group, filter)
 
     def apply_filter(self, data, groupfunc=numpy.sum):
         """Group and filter the supplied data to match the data set.
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1503,7 +1503,7 @@ class DataPHA(Data1D):
             self.get_background(id).rate = val
 
     rate = property(_get_rate, _set_rate,
-                    doc="""Plotting: is the Y axis displayed as a rate?
+                    doc="""Is the Y axis displayed as a rate when plotting data?
 
 When True the y axis is normalised by the exposure time to display
 a rate.""")
@@ -1532,7 +1532,7 @@ a rate.""")
             self.get_background(id).plot_fac = val
 
     plot_fac = property(_get_plot_fac, _set_plot_fac,
-                        doc="""Plotting: how the X axis is used to create the Y axis.
+                        doc="""How the X axis is used to create the Y axis when plotting data.
 
 The Y axis values are multiplied by X^plot_fac. The default
 value of 0 means the X axis is not used in plots. The value
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1500,23 +1500,43 @@ class DataPHA(Data1D):
     def _set_rate(self, val):
         self._rate = bool_cast(val)
         for id in self.background_ids:
-            # TODO: shouldn't this store bool_cast(val) instead?
             self.get_background(id).rate = val
 
     rate = property(_get_rate, _set_rate,
-                    doc='Quantity of y-axis: counts or counts/sec')
+                    doc="""Plotting: is the Y axis displayed as a rate?
+
+When True the y axis is normalised by the exposure time to display
+a rate.""")
 
     def _get_plot_fac(self):
         return self._plot_fac
 
     def _set_plot_fac(self, val):
-        self._plot_fac = int(val)
+        # I'd prefer to check whether val is an integer, but there
+        # may be users who have set the value to 2.0 and it doesn't
+        # seem worth breaking that code.
+        #
+        try:
+            okay = val == int(val)
+        except (ValueError, TypeError):
+            # For when int() can't convert val, which can raise
+            # different errors.
+            okay = False
+
+        if not okay:
+            raise DataErr("bad", "plot_fac setting", val)
+
+        val = int(val)
+        self._plot_fac = val
         for id in self.background_ids:
             self.get_background(id).plot_fac = val
 
     plot_fac = property(_get_plot_fac, _set_plot_fac,
-                        doc='Number of times to multiply the y-axis ' +
-                        'quantity by x-axis bin size')
+                        doc="""Plotting: how the X axis is used to create the Y axis.
+
+The Y axis values are multiplied by X^plot_fac. The default
+value of 0 means the X axis is not used in plots. The value
+must be an integer.""")
 
     def _get_response_ids(self):
         return self._response_ids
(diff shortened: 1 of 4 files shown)


sherpa/astro/data.py
```diff
@@ -1523,23 +1523,43 @@ def _get_rate(self):
...
+            okay = val == int(val)
+        except (ValueError, TypeError):
+            # For when int() can't convert val, which can raise
+            # different errors.
+            okay = False
+
+        if not okay:
+            raise DataErr("bad", "plot_fac setting", val)
+
+        val = int(val)
```
hamogu, line 1552:
```suggestion
        try:
            val == int(val)
        except (ValueError, TypeError):
            # For when int() can't convert val, which can raise
            # different errors.
            raise DataErr("bad", "plot_fac setting", val)
```
I don't understand that extra "okay" variable. If `val = int(val)` works, then val is an int and we don't need okey. If it dies not, then we raise an Error anyway.

sherpa/astro/data.py
```diff
@@ -1523,23 +1523,43 @@ def _get_rate(self):
...
         for id in self.background_ids:
             self.get_background(id).plot_fac = val
 
     plot_fac = property(_get_plot_fac, _set_plot_fac,
-                        doc='Number of times to multiply the y-axis ' +
-                        'quantity by x-axis bin size')
+                        doc="""How the X axis is used to create the Y axis when plotting data.
+
+The Y axis values are multiplied by X^plot_fac. The default
+value of 0 means the X axis is not used in plots. The value
```
hamogu, line 1561: That means the plots don't have an x-axis?

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -1111,6 +1111,58 @@ def test_pha_invalid_units(invalid, make_test_pha):
     assert str(de.value) == f"unknown quantity: '{invalid}'"
 
 
+# currently several variants are allowed, even if they don't work as
+# expected - e.g. type="rates" does not set the rate value to True.
```
hamogu, line 1115: Wait, I thought this PR fixes that so that the variants that don't work as expected no longer work?

hamogu (changes requested):

sherpa/astro/data.py
DougBurke: So, this bit of code went through several revisions, so maybe I just got confused. What is was worried about for your version was that the val that would get used in the `raise` call to create the error message would have already been converted to an integer, and so the message would have been confusing to the user. But I now realize that that thinking is baloney, so I'll make the change.

sherpa/astro/data.py
DougBurke: I was trying to come up with something meaningful to say here, but failed. The problem is that I don't really want to say "the y axis is multiplied by x axis^plot_fac" (even though I do actually say this) as it is logically wrong - the y axis is the y axis. What I'm really trying to say is that the "base units" are multiplied by x^plot_fac values to create the y axis, but there's no definition of "base units" (it depends on the rate setting, for one thing, and the plot type for another). So options include:

- dropping this text, just leaving the first line
- trying to come up with some very basic text that doesn't really say anything (and if so, is it worth keeping)
- coming up with a better way to describe this

Any ideas?

sherpa/astro/tests/test_astro_data2.py
DougBurke: Did I mess something up in my rebasing? These tests shouldn't be marked as xfail (and the comment above should be removed). Let me check

sherpa/astro/data.py
DougBurke: Actually, no. I think I do need the existing code. I want to allow

    factor=1.0

(since this was allowed before, so I don't see it worth breaking code) but disallow

    factor = 1.5

(since this leads to potentially-surprising results) and so you need the current version - or perhaps better would be

```
    try:
        ival = int(val)
        okay = val == ival
    except (...):
        okay = False

    if okay:
        raise ...
```

An alternative to error-ing out would be to create a warning message, but my guess is that not many people have used non-integer-like values (and  that this is technically different to allowing 1.0).

sherpa/astro/tests/test_astro_data2.py
DougBurke: My rebase left these in when they can be removed (ie the xfail tests suddenly started passsing). So I've made the change in "remove un-needed xfail markers".

sherpa/astro/data.py
```diff
@@ -1523,23 +1523,44 @@ def _get_rate(self):
...
+        try:
+            ival = int(val)
+            okay = val == ival
+        except (ValueError, TypeError):
+            # For when int() can't convert val, which can raise
+            # different errors.
+            okay = False
+
+        if not okay:
+            raise DataErr("bad", "plot_fac setting", val)
```
hamogu, line 1552: I think I still see a problem here, but I need to play around with the old version more to see what the differences are. Unfortunately, I might not get to that today.

DougBurke: There's no hurry.

DougBurke: If we didn't want to try and be backwards compatible - i.e. we don't need to support "set_analysis(... , factor=2.0)" - then it would be easier. I note in matplotlib 5.3.0 that they have a bunch of changes like this (I think), but I'm not sure our user community has the same setup as they do.

I'm not particularly happy with the code, but it works (as far as I can tell).

sherpa/astro/data.py
hamogu: How about:
```
'''Scale y values when plotting data

y-axis values can be multiplied by x^plot_fac
where x is the corresponding x-value for each datapoint. 
The default is plot_fac=0, so no scaling is applied. This is useful to show e.g. counts/kev on the y-axis of the plot.
'''
```
(Does my example make any sense?)

sherpa/astro/data.py
hamogu: What exactly is this supposed to accept?
- Do inputs have to be integers, i.e. is there a good physical reason to allow `factor=2`, but not `factor=3/2`? That asks for what scalings make sense. Surely 0 and 1 do make sense, but why not 0.5? I can imagine someone scaling with sqrt(E) or sqrt(lambda), in which case such a plot would make sense or not?
- What inputs do we want to accept? Int, float, also string? I can only trigger the except clause with exotic cases like strings, complex number etc. If the goal of the except clause is to give a better error message for strings, etc., I would suggest:
```
except (ValueError, TypeError):
    raise ArgumentTypeErr("badarg", val, "int or float")
```

Code that I used to try to understand what happens
```
def _set_plot_fac(val):
    '''removed "self" argument and turned output and error into print statements
    so I can call it a couple of times in a row'''
    try:
        ival = int(val)
        okay = val == ival
    except (ValueError, TypeError):
        # For when int() can't convert val, which can raise
        # different errors.
        print("except clause triggered")
        okay = False

    if not okay:
        print(f"{val=} not okay")
        return
            
    print(ival)


0
1
2
val=2.5 not okay
0
1
2
val=2.5 not okay
val='0' not okay
val='1' not okay
val='2' not okay
except clause triggered
val='2.5' not okay
except clause triggered
val='0.' not okay
except clause triggered
val='1.' not okay
except clause triggered
val='2.' not okay
except clause triggered
val='2.5' not okay
```

DougBurke: It accepts an integer. There is no support for non-integer values. adding support for factor=0.5 would be a much-larger PR. This is only about improving the existing support.

When setting plot_fac - either directly as an attribute or via the factor argument of set_analysis - the original code would take the value you send it (`val`) and then call int on it (`int(val)`) - so that if you sent it 0.5 it would silelntly get converted to 0. All I'm really trying to do is ensure we test these assumptions.

So your output above looks like I'd expect it to.

The one question is that you used to be able to set plot_fac to "1" or "2" and this is no-longer supported, but I am pretty sure we never documented support for this. There are a number of places in our code base hwere we can accept strings rather than numbers which comes from "pre history" (e.g. when supporting S-Lang as well as Python and a spill over from the old command language for Sherpa).

hamogu (approved):

sherpa/astro/data.py
hamogu: I've approved it already. I don't think my wording is that much better than what's in the PR now, but feel free to use either or a combination of both.

DougBurke:
I've rebased this to address a merge conflict / slight-change-in-behavior. In doing so I took advantage of moving some code around to remove one of the commits.

PR #1303, opened 2021-09-18
Updated Introductory Section of the MCMC page

anetasie:
Corrected a few words in the Introductory text.

PR #1352, opened 2021-11-16
use tox

DougBurke:
I'm probably going to keep this until #1608 gets merged, so I can rebase it and check we aren't missing anything above support for tox, then am likely to close it, as we currently haven't got a strong need to move to tox or nox

# Summary

Provides initial support for using tox to test the code.

# Details

This has had a long history. I have now moved most of what was in this PR into #1608 as this is now just for adding support of tox. I don't understand tox enough to know how to handle things like

- how to run tests with `matplotlib`, `astropy`, the data directory, and XSPEC
- how to extend `tox` to get it to build the documentation.

# Notes

# Relationship to #1142 

This is built on #1329 and #1608. It almost gets to the place @hamogu was going for in #784 but it does not go as far. The parts of #784 that are not done are

1. change to `setuptools_scm`
2. using tox for building the documentation

and I don't believe either of them are needed to be solved in this PR (i.e. they are for future work).

I could well imagine us deciding not to take the `tox` stuff and just the other changes.

The commits are basically

- use `pytest` rather than `python setup.py test` in the GitHub Actions 
- explicit removal of the `python setup.py test` command (at least as far as we can do so) which we need to do as it will be removed from setuptools at some point
- addition of tox support (very basic, I admit to not really understanding tox yet so I just wanted something basic and we can look at adding support for more test options and using tox in GH Actions in later PRs).

## (really historical)

This no-longer seems to be the case; whether it's because we now force `setuptools < 60`, there have been version changes to our dependendencies, or some change in what we are doing since I originally wrote this last year I do not know. I'm leaving it in just for record:

Now we have setuptools 60 available we seem to need to set `SETUPTOOLS_USE_DISTUTILS=stdlib` to build Sherpa, but there's still a problem with tox:

```
% SETUPTOOLS_USE_DISTUTILS=stdlib tox -e test
test inst-nodeps: /home/dburke/sherpa/sherpa-master/.tox/.tmp/package/1/sherpa-4.14.0+57.g3b58f5cf.dirty.tar.gz
[...]

PR #1376, opened 2021-12-07
Very-minor tweaks to docs/tests

DougBurke:
# Summary

Fix a minor test issue and an unused reference in the documentation.

# Details

I noticed these problems during other unrelated work (from building CIAO and building the sphinx documentation). They are really-minor changes so I put them together so I wouldn't forget about them. The test issue is something I am sure I added, and the unused reference is from #1303 which removed the use of the reference but not the reference itself (rather than remove the reference I just added a reference to the website in the acknowledgements section as it seemed appropriate, and replaces the use of the web link in the text).

### DIFF
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -30,9 +30,9 @@ Acknowledgements
 ----------------
 
 The original version of the code was developed by the CHASC
-Astro-Statistics collaboration http://hea-www.harvard.edu/AstroStat/,
-and was called pyBLoCXS. It has since been developed by the
-Chandra X-ray Center and weas added to Sherpa in version 4.5.1.
+Astro-Statistics collaboration, and was called pyBLoCXS [1]_. It has
+since been developed by the Chandra X-ray Center and weas added to
+Sherpa in version 4.5.1.
 
 Overview
 --------
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -31,7 +31,7 @@ Acknowledgements
 
 The original version of the code was developed by the CHASC
 Astro-Statistics collaboration, and was called pyBLoCXS [1]_. It has
-since been developed by the Chandra X-ray Center and weas added to
+since been developed by the Chandra X-ray Center and was added to
 Sherpa in version 4.5.1.
 
 Overview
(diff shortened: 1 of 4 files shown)


sherpa/sim/__init__.py
```diff
@@ -30,9 +30,9 @@
 ----------------
 
 The original version of the code was developed by the CHASC
-Astro-Statistics collaboration http://hea-www.harvard.edu/AstroStat/,
-and was called pyBLoCXS. It has since been developed by the
-Chandra X-ray Center and weas added to Sherpa in version 4.5.1.
+Astro-Statistics collaboration, and was called pyBLoCXS [1]_. It has
+since been developed by the Chandra X-ray Center and weas added to
```
hamogu, line 34:
```suggestion
since been developed by the Chandra X-ray Center and was added to
```

DougBurke: At least this wasn't my typo.

Well, actually, I think it was my typo. But from old me, not this-PR me...

DougBurke:
@hamogu - sorry, I added another simple fix I has added in #1352 but it doesn't need to be there and fits in here.

DougBurke:
Hmm. The test failure is infrastructure related, not this code change

> Codecov: Uploader shasum does not match -- uploader hash: b09caed14d58bea0436487ae67d80d0bd743c64d96beac0ff35fe0336c504622  codecov, public hash: 4a19a16d3e3ec6966b3499bb166e08fa2226e64acc68edd97060631b01cf3ee3  codecov

I'll see if I can re-run them or this is related to AWS failures my twitterz is complaining about...

DougBurke:
Ahh. The tests are back to green.

hamogu (approved):
I apologize for the delay. I know I looked at this but for some reason apparently did not submit the review.

I don't want to hold this up, but in the spirit of #1306 I'm tempted to ask for real-world examples that justify an isolated change of `==` to `is` without a full review of the code base to see if there are any other instances that this PR does not address ;-)

DougBurke:
Well, I fixed the **only** case of  'please change is to ==' that python complained about. I'm not sure how else to identify other places where this could happen. I also think an isolated change like this is vastly different from the #1306 case where we **know** that there are many other problems.

PR #1378, opened 2021-12-07
Add a Linux conda build using CIAO

DougBurke:
# Summary

Add a CIAO build via conda for Github Actions

# Details

This is a replacement for #1069 with the main change being that it doesn't replace our manual installation of miniconda with the setup-miniconda action.  It does differ a bit in design to the original PR as it more-follows the #1328, #1329, #1352 PRs in that it tries to place most of the logic directly into the workflow rather than having it placed in separate scripts (there are a lot-less options for this test, which helps).

Some items to note

- this is not production ready as there are parts that have been commented out from my efforts to address the DS9 issue
- uses CIAO 4.13 as 4.14 has not been released (should be a relatively easy change)
- limited to the "default" Python for the conda build (which I think is sufficient for us)
- limited to Linux (which is also okay for us as the macOS tests do run slower for whatever reason)
- does not run the ds9 tests as something causes them to fail and it's not obvious what it is.

# Notes

In many ways I'd want the logic in the script

    .github/scripts/setup_conda_ciao.sh

to actually be embedded in the GH action, as it needs to be kept up-to-date with the CIAO version we install (in particular the
XSPEC version), but the advantage of having this as a script is that it's helpful for testing with a local CIAO build.

# What is going on with DS9?

There has been a persistent problem (also found when developing the #1328, #1329, #1352 chain of PRs) with the DS9 tests. I have spent **way** too long trying to determine why the ds9 tests fail. We have checks that they work for Sherpa, and for CIAO
we can test them outside this process so I have elected to forgo them here for now.

I wonder whether it's related to the following code in sherpa/conftest.py (but why would this cause problems?):

    # In some instances, if some xvfb processes did not stop cleanly
    # pytest-xfvb starts complaining that a virtual screen is already on
    # the following code works around that. The issue is hard to reproduce
    # but I (OL) tested it locally as it reappeared on my workstation.
    try:
        import random
        from pyvirtualdisplay import abstractdisplay
        abstractdisplay.RANDOMIZE_DISPLAY_NR = True
        abstractdisplay.random = random
        random.seed()
    except ImportError:
        pass

I was able to call the Sherpa image-open routine via pyvirtualdisplay to show that it should work, but when it gets to the `sherpa_test` run the image tests would fail as it didn't seem to stop the ds9 process (something I have noticed on my machine too). So there's the issue of "shouldn't we be closing down the ds9 application?" and "why are we seeing this problem in this build but not in the other conda builds (not using CIAO)?".


sherpa/conftest.py
```diff
@@ -27,6 +27,7 @@
 
 import pytest
 
+from sherpa.utils.err import RuntimeErr
```
DougBurke, line 30: argh - this isn't needed. I'll remove it at some point.

DougBurke:
I've rebased this to pick up changes and tweaked the config file to see if it still works.

This is still more a "fun project" rather than something to be looked at seriously, for now.

DougBurke:
I've rebased this to pick up recent changes and

- to update the conda workflow to use the latest version
- bump to Python 3.10
- update to CIAO 4.15 (well, it uses the default version, but that's 4.15 now)

DougBurke:
Rebased (in part to remove the "numpy<1.24" temporary fix between 4.15.0 ad 4.15.1)

DougBurke:
Rebased to pick up 4.1.7.0 (give or take the final release), combine the commits, and update to better match the current `ci-conda-workflow` settings.

DougBurke:
It looks like it isn't picking up the codecov report, but that's for future me to worry about.

DougBurke:
Updated to 

- use CIAO 4.17 (although there's no actual version used to it just uses the latest)
- still only uses Python 3.11 and Linux
- now installs microconda rather than any installed conda version

PR #1387, opened 2022-01-04
minimize C++ code duplications (rebase)

DougBurke:
# Summary

Remove code duplication in the XSPEC module.

# Details

This is #1330 which has been rebased to the latest main branch, and then extended to take advantage of XSPEC-provided typedefs. Once we are able to remove support for XSPEC < 12.12.0 we can simplify the code more, but that doesn't need to be included in this PR.

The code to support XSPEC typedefs is partly dependent on how we package XSPEC as a conda package - the "Switch to XSPEC typedefs" commit has a section where the `xsf77Call` and `xsccCall` typedefs are defined which really should be taken from `funcType.h` (pre 12.12.0) or `XSFunctions/Utilities/funcType.h` (>= 12.12.0) but our `xspecmodel-modelsonly` conda build does not include this file [*], so I've hard-coded them (they haven't changed for a LONG time, if ever).

[*] we currently do not have a 12.12.0 version of the xspec-modelsonly conda package

The remaining commits tweak how the C++-to-Python translation of errors is handled. The C++ code throws errors and then the "interface" code  - that is the code that is called by the Python routine to marshal data from Python to C++ and then back again - has to decide what to do with these errors. The chosen approach is based around three classes that extend std::runtime_error (the names are not particularly great):

- `NoError`
- `TypeError`
- `ValueError`

which indicate to the caller how to convert the error message to Python [**].

[**] In Python the error handling is done by setting the error "status" by the `PyErr_SetString` function and then returning `NULL` (actually, there are other related routines that could be called but we'll stick with this one here)

An alternative approach would be to call `PyErr_SetString` at the location of the error and then throw `NoError` so we know we don't need to do anything else. This might be better than having to add `TypeError` and `ValueError` classes.

The reason why I went with this approach was that:

a) there are cases when all we should do is return NULL since a Python C routine has already set up the error string - for this code this is true of `PyArg_ParseTuple` and variants - that is we don't want to replace the error message they have created - which is why `NoError` was added. I have added a test to check some of these cases.

b) some routines can cause a Python `TypeError` or `ValueError` to be raised, depending on the error (e.g. `create_grid`). This lead to the addition of `ValueError` and `TypeError` C++ classes. I have added tests to check some of these cases.

It is likely that this code will not work properly prior to XSPEC 12.10.1 but this was released a long time ago, we don't have a CI run that can test this, and I want to bump the minimum-supported-verision to exclude this case (see #1391).

There is one "interesting" error case left, which is when the library initialization code fails. Instead of an `"Unable to initalize Xspec"` message - which was never very helpful - we'll get some other message (dependent on the model type). However, I plan to remove this case as discussed in #1388 - so I have not spent much time on this case.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -1,4 +1,4 @@
-//  Copyright (C) 2007, 2015-2018, 2019, 2020, 2021
+//  Copyright (C) 2007, 2015-2018, 2019, 2020, 2021, 2022
 //  Smithsonian Astrophysical Observatory
 //
 //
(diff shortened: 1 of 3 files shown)


dtnguyen2 (changes requested):
Not sure how you want to do this, but the following files
```
(testing-PR) [dtn@devel12 dtn]$ ls -l /pool14/dtn/*xspec*
-rw-rw-r--. 1 dtn dtn 23K Jan  6 15:54 /pool14/dtn/test_xspec.py
-rw-rw-r--. 1 dtn dtn 37K Jan  6 16:33 /pool14/dtn/xspec_extension.hh
```
The said files should be placed in:
``` 
 sherpa/astro/xspec/tests/test_xspec.py
 sherpa/include/sherpa/astro/xspec_extension.hh
```
The changes will enable the user to see what is the root cause of the error.

DougBurke:
@dtnguyen2 - is this the code you talked about in https://github.com/sherpa/sherpa/pull/1330#issuecomment-1006056215 ?

dtnguyen2:
yes, the test file has to be updated because an error message is now being printed

dtnguyen2:
I updated the file

```
(testing-PR) [dtn@devel12 simplify_psf]$ ls -l /pool14/dtn/xspec_extension.hh 
-rw-rw-r--. 1 dtn dtn 39K Jan  7 10:49 /pool14/dtn/xspec_extension.hh
```

to include a ``` try &  catch``` block for completeness in the constructor and ran indent on the code that I modified.

```C++
        xspecModelFctC( PyObject* args, npy_intp NumPars, bool HasNorm ) try
          : xspecModelFctBase<Real, RealArray>( args, NumPars, HasNorm ) {
          } catch( std::runtime_error& re ) {
          throw re;
        }
```
I notice that the ```#ifdef XSPEC_12_10_1``` code is still present in the code

```C++
#ifdef XSPEC_12_10_1

{
...
        if ( !PyArg_ParseTupleAndKeywords( args, kwds, (char*)"O&O&|O&s", kwlis\
t,
                        (converter)convert_to_contig_array< FloatArray >,
                        &pars,
                        (converter)convert_to_contig_array< DoubleArray >,
                        &xlo,
                        (converter)convert_to_contig_array< DoubleArray >,
                        &xhi,
                        &filename) )
          return NULL;
...
}

#else

...
}

#endif
```

but I was under the impression that earlier version of 12.10 of Xspec will no longer be supported?  Moreover, I have not modified any code within the aforementioned code in  PR #1330 so no error message will be printed if there is an error.


BTW Not sure why starting from a clean branch
```
(testing-PR) [dtn@devel12 simplify_psf]$ git status
On branch DougBurke-fix_issue_1284
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   setup.cfg

no changes added to commit (use "git add" and/or "git commit -a")
```
```Python
sherpa/astro/ui/tests/test_serialize.py ............FF......             [ 35%]
```

DougBurke:
@dtnguyen2 - I'm currently integrating your code into this branch - there were some changes that I added that made it harder and then there's a change in the table-model code that needs a little love and care with.

At the moment we haven't made a decision on the minimum-supported XSPEC version, so I haven't changed things. I'd like to make that change in a separate PR just to make things a bit cleaner.

DougBurke:
In the middle of rebasing ...

DougBurke:
@dtnguyen2 - I've finished updating the code. It turned out to be a bit more work than your original suggestion - I've tried to add an explanation in the commit messages and the `Details` section of the PR summary above. Let me know what doesn't make sense.

DougBurke:
This has been rebased to pick up #1400 so the CI runs pass

dtnguyen2 (approved):

dtnguyen2 (commented):
So the class ```NoError``` does throw with an error message:

```C++
          if ( !PyArg_ParseTuple( args, (char*)"O&O&|O&",
                                  (converter)convert_to_contig_array< RealArray \
>,
                                  &pars,
                                  (converter)convert_to_contig_array< DoubleArra\
y >,
                                  &xlo,
                                  (converter)convert_to_contig_array< DoubleArra\
y >,
                                  &xhi ) )
            throw NoError("Error Parsing args");
...
```

but there are five instances of catching ```NoError``` without calling ```PyErr_SetString``` so the error message within the ```NoError``` class will not be broadcast to the user.

```C++
template <npy_intp NumPars, bool HasNorm, xsf77Call XSpecFunc>
PyObject* xspecmodelfct( PyObject* self, PyObject* args ) {

  xspecModelFctF<float, FloatArray, XSpecFunc> xspec_model =
    xspecModelFctF<float, FloatArray, XSpecFunc>( args, NumPars, HasNorm );
  try {
    FloatArray result;
    xspec_model.eval( result );
    return result.return_new_ref();
  } catch(NoError& re) {
    return NULL;
  } catch(ValueError& re) {
    PyErr_SetString( PyExc_ValueError, re.what() );
    return NULL;
  } catch(TypeError& re) {
    PyErr_SetString( PyExc_TypeError, re.what() );
    return NULL;
  } catch(...) {
    // Even though the XSPEC model function is Fortran, it could call            
    // C++ functions, so swallow exceptions here                                 
    PyErr_SetString( PyExc_ValueError, xspec_model.get_err_msg() );
    return NULL;
  }

}
```

DougBurke:
Yes, because the cases where `NoError` are thrown are ones from calls to PyArg_ParseTuple which already set the python error message. I am not well versed in the C++ exception semantics to know whether it makes sense to just have a call to 

    throw NoError();

From a quick read on the internet, which is dangerous, I see a suggestion that the catch case should be

  } catch(const NoError& re) {

PR #1396, opened 2022-01-10
XSPEC minimum version is now 12.10.1

DougBurke:
# Summary

Bump the minimum-supported version of XSPEC from 12.9.0 to 12.10.1.

# Details

This requires #1676 which was broken out to make things a bit easier to review and avoid (hopefully) catastrophic conflicts.

There are four main areas to be changed

- documentation
- the actual enforcement of the minimum version (which is in `helpers/xspec_extension.py` and is a small change
- removal of the conditional code in the C++ code (note: we still have some conditional code but that is expected as there are new models in 12.11.* / 12.12.0 that need to be handled)
- the Python code has some tests, documentation, and support for models prior to 12.10.1 that can be removed now

One change I have made is that we now error out if the XSPEC version is too old. As we know the code can't compile against we may as well error out earlier, with a useful error message, rather than later with some undefined symbol or missing include file message. This is more relevant for the follow-up PR to move to 12.12.0 as the minimum supported version!

My long-term aim is to actually require 12.12.0 or later as the minimum version, since that opens up extra changes, but that's a larger change (see #1609, which is built on top if this PR). This change does simplify a lot of code (and XSPEC 12.10.1 is "old", so I would be surprised anyone is still using it with Sherpa).

I have run this locally with XSPEC 12.12.0 (using a local build of XSPEC) and with `xspec-modelsonly=12.10.1` - the CI run handles the `xspec-modelsonly=12.11.1` case (this was with an earlier version of the PR but since then we've added XSPEC 12.12.1 to the CI runs).

PR #1400, opened 2022-01-11
pinning readline to 8.1.0

harlantc:
Pinning readline to 8.1.0 in the conda setup script, as readline 8.1.2 is broken on conda defaults at the moment.

This is to be reverted as soon as conda resolves the readline issue.


harlantc:
@DougBurke basically your solution in #1396 but I put it in the env creation to prevent multiple env updates.

DougBurke (approved):
LGTM

My approach was just to change only those builds that seemed to have problems (that is, those that include XSPEC). If we have to keep this PR for a while then it might make sense to make sure we test a conda build using their "base" environment (in other words use something more-like my approach), but hopefully we can remove this soon.

PR #1142, opened 2021-05-07
Add instruction for source build on Mac, that were previously only part of our internal release notes.

hamogu:
Add instruction for source build on Mac, that were previously only part of the internal release notes.
   
fixes #1141
[skip ci]
(Because it does not touch any code files.)

PR #1401, opened 2022-01-11
Modified region code to remove implicit prototype error

wmclaugh:
**Summary**

This change updates the extern region code in sherpa to resolve a missing prototype which causes build issues with clang 12.0.

**Details**

Changes to remove build error on clang:

     regparser.y:629:5: error: implicit declaration of function 'regYYrestart' is
      invalid in C99 [-Werror,-Wimplicit-function-declaration]

and remove the workaround added to docs/install.rst in #1142:

    .. note::

       Additionally, if you are building with Clang version 12, you will
       encounter build issues in the region area related to an implicit
       declaration. If so, pre-pending the following to your pip install
       or python setup.py install line should resolve the build issues::

         CFLAGS='-Wno-implicit-function-declaration'

       (Note that on MacOS, "gcc" is usually just an alias to Clang. Unless
       you specifically install a gcc from a different source, you are
       likely to run into this problem, even if you believe that your
       compiler is invoked with `gcc`.)

The changes coincide with updates to the CIAO region library code issued in bug fix SL-251.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -384,20 +384,6 @@ C++ compilers to use if not found by ``setup.py``.
 
    That is, the variable is set to a space, not the empty string.
 
-.. note::
-
-   If you are building with Clang version 12, you will
-   encounter build issues in the region area related to an implicit
-   declaration. If so, pre-pending the following to your pip install
-   or python setup.py install line should resolve the build issues::
-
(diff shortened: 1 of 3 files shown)


DougBurke (approved):
LGTM - as much as I understand lex/yacc at least.

PR #1405, opened 2022-01-13
revert readline fix

harlantc:
Looks like readline has been fixed by anaconda, I've tested it on both a minimal install (just python and readline=8.1.2) and a sherpa CI run, and everything looks good.

I wasn't sure if we want to do this with a new commit (as I've done here) or just remove the previous commit. As I'm not sure how to do the second option cleanly, I've just done the first.


DougBurke:
I don't think we can "remove" the last commit without breaking all the things, so just reverting the change as you do in the commit makes sense.

wmclaugh (approved):

PR #1407, opened 2022-01-18
Add Codebase Linting using Super-Linter GitHub Action

harlantc:
So I decided to distract myself from the coming git-ocolypse migration and set up a super linter.

As it's set now, it will do a complete check using black python linter and you can see what would be changed in the whole codebase, but future versions would only check new code.

I know it's likely not to be taken, as there are concerns over massive whitespace changes, but I figured it'd be nice to have this explored.

I may make another check to see if any of the changes (apply them wholesale with no regard to what it will do) to see if it breaks anything.

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -1,0 +1,31 @@
+# This workflow executes several linters on changed files based on languages used in your code base whenever
+# you push a code or open a pull request.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/github/super-linter
+name: Lint Code Base
+
+on:
+  push:
+    branches: [ main, testing_lint ]
+  pull_request:
+    branches: [ main ]
+jobs:
+  run-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


hamogu:
black ensure that the bytecode is unchanged, so it should be impossible to introduce bugs.

hamogu:
In astropy, we're switching from black to ruff for performance. The linting rules are virtually identical (or can be made identical with options) and ruff is about 100 times faster. Could be run as a pre-commit step. It's not hard to set up.

That said, to me this is all mental overhead. In an ideal world, we'd have a linter and autolint (and e.g. astropy does that), but the benefits are marginal for a codebase that has only very few contributors (mostly Doug).  On the other hand, stuff breaks - linters get updated, versions change, setting move from `setup.cfg` to `pyproject.toml` etc. and someone has to stay on top of that. Given our limited resources, I'd favor a model where we run a linter or similar tool locally as needed (e.g. like running 2to3 in the Python2->3 transition, or to take care of certain deprecation warnings or when someone really hates the way the code in some module is inconsistent in style) and put that in a specific PR rather than commiting to maintain another hook, tool, or setup

hamogu:
In the sherpa developer telecon, we decided to close this PR since a lot of new approaches (e.g. ruff) are emerging and it's not clear what is the easiest for developers.

PR #1410, opened 2022-01-23
Add script and docs for pre-commit hook to update copyright

hamogu:
# Summary
Add example script for a git pre-commit hook that checks the copyright is up to date.

# Details
This PR adds a script to the `scripts` directory that can be used as a git pre-commit hook; docs are added to the developer section to explain how to use it. Since git hooks are local (they are not tracked in the repro itself) users have to actively copy their example script into their local `.git/hooks`. In that sense this is a documentation-only PR, it add docs and a script that might be used, but without the user installing (copying into the right place) this script, it won't ever run. Not on CI, not for anyone else and even if it's installed, it can always be skipped for special cases. Thus, this PR can't possibly break anything.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -652,11 +652,11 @@ Never forget to update the year of the copyright notice?
 
 Git offers `pre-commit hooks <https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks>`_
 that can do file checks for you before a commit is executed. The script in
-`sherpa/pre-commit` will check if the copyright notice in any of the files in the
+`scripts/pre-commit` will check if the copyright notice in any of the files in the
 current commit must be updated and, if so, add the current year to the copyright notice
 and abort the commit so that you can manually check before commiting again.
 
-To use this opt-in functionality, simply copy the file to the appropriate location:
+To use this opt-in functionality, simply copy the file to the appropriate location::
 
   cp scripts/pre-commit .git/hooks
 
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -646,6 +646,20 @@ available.
       by addtive and then multiplicative models, using an alphabetical
       sorting - and to the appropriate ``inheritance-diagram`` rule.
 
+
+Never forget to update the year of the copyright notice?
+--------------------------------------------------------
+
+Git offers `pre-commit hooks <https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks>`_
+that can do file checks for you before a commit is executed. The script in
+`sherpa/pre-commit` will check if the copyright notice in any of the files in the
+current commit must be updated and, if so, add the current year to the copyright notice
+and abort the commit so that you can manually check before commiting again.
+
+To use this opt-in functionality, simply copy the file to the appropriate location:
+
+  cp scripts/pre-commit .git/hooks
+
 Notes
 =====
 
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -32,7 +32,7 @@ copyr = re.compile(r"(?P<comment>(#|//))\s*Copyright \(C\) (?P<years>[0-9,\s]+)(
 # Only apply this to C or Python files since sherpa does not use
 # copyright statements in others files, also, do not change in extern
 # directory.
-ftype = re.compile(r"^((?!extern).)*(hh|cc|c|py)$")
+ftype = re.compile(r"^((?!extern).)*(hh|cc|c|h|py)$")
 changed = 0
 
 if __name__ == '__main__':
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -32,7 +32,7 @@ copyr = re.compile(r"(?P<comment>(#|//))\s*Copyright \(C\) (?P<years>[0-9,\s]+)(
 # Only apply this to C or Python files since sherpa does not use
 # copyright statements in others files, also, do not change in extern
 # directory.
-ftype = re.compile(r"^((?!extern).)*(hh|cc|py)$")
+ftype = re.compile(r"^((?!extern).)*(hh|cc|c|py)$")
 changed = 0
 
 if __name__ == '__main__':
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -58,7 +58,7 @@ if __name__ == '__main__':
                               '  ' + m.group('institution'), end='')
                 else:
                     print(line, end='')
-            subprocess.Popen("git add " + f, shell=True)
+            # subprocess.Popen("git add " + f, shell=True)
         except FileNotFoundError:
             # file was removed with "git rm" so there is nothing to check here
             pass
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -29,6 +29,9 @@ import sys
 
 year = str(datetime.datetime.now().year)
 copyr = re.compile(r"(?P<comment>(#|//))\s*Copyright \(C\) (?P<years>[0-9,\s]+)(?P<institution>[\w\s]*)")
+# Only apply this to C or Python files since sherpa does not use 
+# copyright statements in others.
+ftype = re.compile(r"(hh|cc|py)$")
 changed = 0
 
 if __name__ == '__main__':
@@ -36,6 +39,7 @@ if __name__ == '__main__':
     proc = subprocess.Popen("git diff --name-only --staged",
                             shell=True, stdout=subprocess.PIPE)
     filenames = proc.stdout.read().decode().split()
+    filenames = [f for f in filenames if ftype.match(f)]
 
     for f in filenames:
         try:
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+#
+#  Copyright (C) 2022
+#  MIT
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


scripts/pre-commit
```diff
@@ -0,0 +1,63 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+'''Git pre-commit hook to update the copyright year in Sherpa code files
+'''
+from __future__ import print_function
```
DougBurke, line 23: We shouldn't need this, should we?

hamogu: git pre-commit scripts are run with whatever the active Python is in your development environment. Usually, I have a Python 3 active through conda, but my system Python is still 2.7, so we want this script to work in those cases, too. (I don't think it's too crucial, since only developers who know what's happening will use this feature, but it does not hurt to have it.)

DougBurke:
So, with this I get

```
% git commit -m 'bob'
Message from your pre-commit hook script:
I updated copyright statements in one or more files for you.
After doing that, I aborted the commit, so you can check yourself that
it's alright before you commit again.
Commit with ' --no-verify' option to skip this check.
```

So it does what it says on the can. Now if I look I see

```
% git status
On branch delme
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   sherpa/io.py

```

However, how can I check what the change was? I can manually say

```
%  head sherpa/io.py
#
#  Copyright (C) 2007, 2015, 2016, 2019, 2020, 2021, 2022
#
```

which shows that the year was updated to 2022, but it is a bit of an extra step.

I then went and commented out the "git add" line in the pre-commit step and the behavior is now:

```
% git commit -m 'bob'
Message from your pre-commit hook script:
I updated copyright statements in one or more files for you.
After doing that, I aborted the commit, so you can check yourself that
it's alright before you commit again.
Commit with ' --no-verify' option to skip this check.
% git status
On branch delme
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   sherpa/io.py

  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   sherpa/io.py

% git diff
diff --git a/sherpa/io.py b/sherpa/io.py
index f4d6ffba..188d0328 100644
--- a/sherpa/io.py
+++ b/sherpa/io.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2016, 2019, 2020, 2021
+#  Copyright (C) 2007, 2015, 2016, 2019, 2020, 2021, 2022
 #
 #
```

so it makes it easier to see what the script made. I think I'd prefer this for the moment whilst we get used to it, but maybe I'm too conservative here.

DougBurke:
ps I see that this works correctly (this is a case where I'm using my hacked `pre-commit` script), which is nice

```
% git diff
diff --git a/sherpa/logposterior.py b/sherpa/logposterior.py
index 7e2dfc67..67eeb657 100644
--- a/sherpa/logposterior.py
+++ b/sherpa/logposterior.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2009, 2016, 2020, 2021  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2009, 2016, 2020, 2021, 2022
+#  Smithsonian Astrophysical Observatory
 #
```

DougBurke:
Here's a "bug-let". If you are adding a binary file (in this case I'd accidentally added a python pickled file) then the script will fail with an error like the following. In my case it was actually useful as it pointed out that I'd accidentally added something I shouldn't have ....

```
% git commit -m 'WIP tests'
  File "/home/dburke/sherpa/sherpa-main/.git/hooks/pre-commit", line 42, in <module>
    for line in fileinput.input(files=filenames, inplace=True):
  File "/home/dburke/anaconda/envs/sherpa-main/lib/python3.9/fileinput.py", line 249, in __next__
    line = self._readline()
  File "/home/dburke/anaconda/envs/sherpa-main/lib/python3.9/fileinput.py", line 369, in _readline
    return self._readline()
  File "/home/dburke/anaconda/envs/sherpa-main/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

hamogu:
I pushed tow updates that should help that. However, this is never going to be absolutely fool-proof. That's why id does not do the commit, but just aborts, even when if finds and changes things. You can always skip the verification with `git commit  --no-verify` for that particular commit (or rename the script in the `.git/hook` directory so that git does not find it), so it's probably better to be too sensitive then to miss too much. But of course not as sensitive that developers get annoyed - after all that's what this script is supposed to help with in the first place.

harlantc:
thanks, @hamogu for this, I've been writing up a guide for our git migration, and I was considering adding details about pre-commit hooks (though held off as I didn't want to make it too complex). I'll probably point to this PR so developers can get a sense of how they could be helpful.

scripts/pre-commit
```diff
@@ -0,0 +1,72 @@
...
+                          '  Copyright (C) ' + m.group('years').strip() +
+                          ', ' + year)
+                    # sometimes group institution will be empty, because it's
+                    # on the next text line. If not, move it there.
+                    if m.group('institution'):
+                        print(m.group('comment') +
+                              '  ' + m.group('institution'), end='')
+                else:
+                    print(line, end='')
+            subprocess.Popen("git add " + f, shell=True)
```
DougBurke, line 61:
```suggestion
            # subprocess.Popen("git add " + f, shell=True)
```

I prefer the ergonomics of being able to see the suggested change with `git diff`, as (although this is a restricted change) I still feel uneasy just automatically believing an automated up-date system (unless it's been through a lot of testing like `black`...). 

It does make me have to do a little-more work, but I don't think it's that onerous.

Or am I missing something?

DougBurke:
As a check (based on the recent changes) I just tried changing a `*hh` file and it updated correctly too. So that's nice.

docs/developer/index.rst
```diff
@@ -646,6 +646,20 @@ available.
       by addtive and then multiplicative models, using an alphabetical
       sorting - and to the appropriate ``inheritance-diagram`` rule.
 
+
+Never forget to update the year of the copyright notice?
+--------------------------------------------------------
+
+Git offers `pre-commit hooks <https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks>`_
+that can do file checks for you before a commit is executed. The script in
+`sherpa/pre-commit` will check if the copyright notice in any of the files in the
```
DougBurke, line 655:
```suggestion
`scripts/pre-commit` will check if the copyright notice in any of the files in the
```

docs/developer/index.rst
```diff
@@ -646,6 +646,20 @@ available.
...
+current commit must be updated and, if so, add the current year to the copyright notice
+and abort the commit so that you can manually check before commiting again.
+
+To use this opt-in functionality, simply copy the file to the appropriate location:
```
DougBurke, line 659:
```suggestion
To use this opt-in functionality, simply copy the file to the appropriate location::
```

I think the extra `:` makes the output look better (but haven't actually checked)

hamogu:
@DougBurke: Applied your review comments. Thanks for the reminder.

DougBurke (approved):
Ta

scripts/pre-commit
```diff
@@ -0,0 +1,72 @@
...
+import re
+import subprocess
+import sys
+
+year = str(datetime.datetime.now().year)
+copyr = re.compile(r"(?P<comment>(#|//))\s*Copyright \(C\) (?P<years>[0-9,\s]+)(?P<institution>[\w\s]*)")
+# Only apply this to C or Python files since sherpa does not use
+# copyright statements in others files, also, do not change in extern
+# directory.
+ftype = re.compile(r"^((?!extern).)*(hh|cc|py)$")
```
wmclaugh, line 35: For completeness, do you want to include C/C++? The comment says C but the regexp only checks for C++. The code base includes various C files outside of the 'extern directory' (for example in sherpa/utils).

hamogu, line 35:
```suggestion
ftype = re.compile(r"^((?!extern).)*(hh|cc|c|py)$")
```

hamogu: Makes sense. I forgot that when I switched form checking all (which can mess up binary files) to just a selected set of files, so added C files.

scripts/pre-commit
```diff
@@ -0,0 +1,72 @@
...
+import re
+import subprocess
+import sys
+
+year = str(datetime.datetime.now().year)
+copyr = re.compile(r"(?P<comment>(#|//))\s*Copyright \(C\) (?P<years>[0-9,\s]+)(?P<institution>[\w\s]*)")
+# Only apply this to C or Python files since sherpa does not use
+# copyright statements in others files, also, do not change in extern
+# directory.
+ftype = re.compile(r"^((?!extern).)*(hh|cc|c|py)$")
```
wmclaugh, line 35:
```suggestion
ftype = re.compile(r"^((?!extern).)*(hh|cc|c|h|py)$")
```

wmclaugh: @hamogu - thanks. For completeness, c header files typically use a .h extension

wmclaugh (approved):

Issue #1389, opened 2022-01-05
Expose _sherpa_init_xspec_library routine from sherpa.astro.xspec._xspec


DougBurke:
For users of XSPEC user models - which have a compiled module that calls into the XSPEC model library - we ideally want to be able to call `_sherpa_init_xspec_library` to ensure the XSPEC lirary is initialized (it is possible, if unlikely, that the sherpa.astro.xspec module has not been loaded when the user-model library is used).

At the moment I have worked around this by having modules that interface to XSPEC user models call sherpa.astro.xspec.get_xsversion from Python scope, but it is not "good engineering" since it is a social contract.

Ideally we would just call this initializer from the module setup code, but it depends in part on whether we go with #1388 - in fact, I think that #1388 would automatically do this for us since the user-module has to import `sherpa.astro.xspec` which would then automatically initialize the interface so the need for this goes away.

PR #1412, opened 2022-01-24
XSPEC: initialize XSPEC library at load time

### LINKED ISSUE #1388: Do we need to cal the INIT_XSPEC function for each XSPEC routine or just when initializing the module?
Historically we have made the XSPEC module (`sherpa.astro.xspec._xspec`, which is automatically loaded by `sherpa.astro.xspec`) behave lazily - that is it does not actually initialize the XSPEC library until it's needed. The idea (I have always assumed) is that we don't need to pay for initialization we are not going to use (e.g. if I fit image data or am not using PHA data). This initialization is done by the `_sherpa_init_xspec_library` C++ routine (which is also available via the alias INIT_XSPEC), and each call to an XSPEC library or routine starts off by calling this initalizer (it does nothing if it's already been called).

However, over time the system has changed and we now have a somewhat-complicated setup, whereby `sherpa.astro.xspec` imports both `sherpa.astro.xspec._xspec` and `sherpa.astro.xspec.utils`, and `sherpa.astro.xspec.utils` itself loads `sherpa.astro.xspec._xspec` and calls one of the XSPEC routines - in this case `get_xsversion` -  during the import. This is used because the model classes created by `sherpa.astro.xspec` depend on what version of the XSPEC library we are using [*].

This all means that anyone trying to import `sherpa.astro.xspec._xspec` will end up initalizing the XSPEC library, whatever you try. We could try moving `sherpa.astro.xspec._xspec` to a sub-directory (so it's in a different directory to `sherpa.astro.xspec.utils`) but I think we should just call `_sherpa_init_xspec_library` inside `PyInit__xspec` (the module initialization code) and then we can remove the calls to it from all the functions.

---

[*] for example, we may have a model which is only available in XSPEC 12.12.0 or later, so if Sherpa was built against XSPEC 12.11.1 then the class will raise an error if a user tries to evaluate that model

# Implementation

The basic change is basically (ignoring the std::cout line) but I need to check whether we need to do anything to indicate an error:

```
% git diff sherpa
diff --git a/sherpa/astro/xspec/src/_xspec.cc b/sherpa/astro/xspec/src/_xspec.cc
index e3ced008..f48337be 100644
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -378,6 +378,8 @@ int _sherpa_init_xspec_library()

   static bool init = false;

+  std::cout << "DEBUG: init check=" << init << std::endl;
+
   if ( init )
     return EXIT_SUCCESS;

@@ -1438,5 +1440,7 @@ static struct PyModuleDef xspec_module = {

 PyMODINIT_FUNC PyInit__xspec(void) {
   import_array();
+  if ( EXIT_SUCCESS != _sherpa_init_xspec_library() )
+    return NULL;  // WHAT TO RETURN ON ERROR?
   return PyModule_Create(&xspec_module);
 }
```

You would also want to remove all the lines like the following in `_xpsec.cc`:

```
  if ( EXIT_SUCCESS != _sherpa_init_xspec_library() )
    return NULL;
```

You'd probably also want to remove the calls to INIT_XSPEC in `sherpa/include/sherpa/astro/xspec_extension.hh`.

DougBurke:
# Summary

Simplify how and when the XSPEC model library is initialized. Fixes #1388 

# Details

Since we now require the module to be loaded to be able to create sherpa.astro.xspec (as we check the version of the
XSPEC library) then it is simpler to just initialize the XSPEC library when the module is created.

The old code used to be useful - in that it meant only people using an XSPEC model ended up having to initialize the XSPEC model library - but with changes to the way the XSPEC models were created, probably in #395, mean that we now need to call the `get_xsversion` routine from `sherpa.astro.xspec._xspec` when loading the `sherpa.astro.xspec` module (the version checks to support models that may have changed or not be available in the run-time version of XSPEC being used), which means that we may as well initialize the XSPEC library when the module is loaded.

This change means that #1389 is not really needed now - we just require that the xspec module has been loaded rather than exporting the initialization routine.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -17,11 +17,6 @@
 //  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 //
 
-// INIT_XSPEC is used by xspecmodelfct() in xspec_extension.hh, so it needs to
-// be defined before that file is included
-int _sherpa_init_xspec_library();
-#define INIT_XSPEC _sherpa_init_xspec_library
-
 // Have sherpa include first so that Python.h is first, to avoid warning
 // messages about redefining _XOPEN_SOURCE
 #include "sherpa/astro/xspec_extension.hh"
@@ -375,25 +370,23 @@ void xsmtbl(float* ear, int ne, float* param, const char* filenm, int ifl,
 
 }
 
-// This routine could be called when the module is being initialized,
-// but this would cause XSPEC module initialization even if XSPEC
-// functionality is not used. So each function/model has to ensure
-// that they call _sherpa_init_xspec_library before calling any
-// XSPEC routine.
+// The XSPEC initialization used to be done lazily - that is, only
+// when the first routine from XSPEC was about to be called - but
+// the module is now set up so that we need to know the version
+// of XSPEC being used when the sherpa.astro.xspec module is
+// being created. As the version requires a run-time check (that
+// is, the get_version call is made) then we know that we need to
+// initalize the XSPEC code when the Python module is installed.
+// So we no-longer need to support the lazy loading.
 //
-// Sun's C++ compiler complains if this is declared static.
-//
-// TODO: should we expose this so that anyone who wants to use a user-model
-//       can access it?
-//
-int _sherpa_init_xspec_library()
+
+static int _sherpa_init_xspec_library()
 {
 
+  // This routine is only meant to be called once
   static bool init = false;
-
   if ( init )
     return EXIT_SUCCESS;
-
 
   if ( !getenv("HEADAS") ) {
     PyErr_SetString( PyExc_ImportError,
@@ -495,16 +488,12 @@ int _sherpa_init_xspec_library()
   }
 
   init = true;
-
   return EXIT_SUCCESS;
 
 }
 
 static PyObject* get_version( PyObject *self )
 {
-
-  if ( EXIT_SUCCESS != _sherpa_init_xspec_library() )
-    return NULL;
 
   char version[256];
   int retval;
@@ -534,9 +523,6 @@ static PyObject* get_version( PyObject *self )
 static PyObject* get_chatter( PyObject *self )
 {
 
-  if ( EXIT_SUCCESS != _sherpa_init_xspec_library() )
-    return NULL;
-
   int chatter = 0;
 
   try {
@@ -558,9 +544,6 @@ static PyObject* get_chatter( PyObject *self )
 
 static PyObject* get_abund( PyObject *self, PyObject *args )
 {
-
-  if ( EXIT_SUCCESS != _sherpa_init_xspec_library() )
-    return NULL;
 
   char* abund = NULL;
   char* element = NULL;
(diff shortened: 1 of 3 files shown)


dtnguyen2 (commented):
In the file _xspec.cc, is there a reason for not calling ```PyErr_SetString``` when there is an error when an error while parsing: 
```C++
  if ( !PyArg_ParseTuple( args, (char*)"|s", &element ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"i", &chatter ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"s", &table ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"fff", &h0, &q0, &l0 ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"s", &csection ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"ss", &str_name, &str_value ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"s", &str_name ) )
    return NULL;
...
  if ( !PyArg_ParseTuple( args, (char*)"s", &path ) )
    return NULL;
```

As for the the xspec_extension.hh file: since ```INIT_XSPEC()``` has been removed from:
```C++
      class PyArgTupleBase {
      public:
        PyArgTupleBase( ) {
        }

      };
```
Then I think you may as well get rid of the class```PyArgTupleBase ``` so the classes ```PyArgTuple3``` and ```PyArgTuple4``` should be changed from:
```C++
      template <typename RealArray>
      class PyArgTuple3 : public PyArgTupleBase {
      public:
        PyArgTuple3( PyObject* args, npy_intp NumPars, RealArray& pars,
                     DoubleArray& xlo, DoubleArray& xhi ) :
          PyArgTupleBase( ) {
 ...
```
to
```C++
      template <typename RealArray>
      class PyArgTuple3 {
      public:
        PyArgTuple3( PyObject* args, npy_intp NumPars, RealArray& pars,
                     DoubleArray& xlo, DoubleArray& xhi {
..
```
and of course remove the instantiation of the class ```PyArgTupleBase``` in

```C++
{

        PyArgTupleBase();
...
```
and
```C++

        PyArgTupleBase();
...
```

DougBurke:
The contract that `PyArg_ParseTuple` makes is that if it fails then it sets the error message for you (and it will be a more-specific error message than you could make at this point). From https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple we have

> Parse the parameters of a function that takes only positional parameters into local variables. Returns true on success; on failure, it returns false and raises the appropriate exception.

which means that the following is the correct thing to do (technically you can call `PyErr-SetSting` here but you then lose the specific error message that was created, so it's just not done):

```
  if ( !PyArg_ParseTuple( args, (char*)"|s", &element ) )
    return NULL;
```

DougBurke:
I'm going to rebase this to see if it fixes the conda/mac build issues,

DougBurke:
The last commit has also been pulled out into #1547 as it's a generic issue, and not related to this PR.

Issue #813, opened 2020-05-21
Set up airspeed-velocity for sherpa?


hamogu:
Astropy/numpy/scipy have a suite of speed benchmarks that are run regularly through CI see e.g. the astropy website here: https://www.astropy.org/astropy-benchmarks/#/

We could investigate doing the same for sherpa. Since sherpa uses a lot of similar infrastructure (Python, CI) etc. and the benchmarking code asv (https://asv.readthedocs.io/en/stable/) itself is Python code, that should not be too hard. 

The hard part is to write the examples that will be benchmarked!

hamogu:
While I don't use airspeed-velocity, I did set up benchmarks in https://github.com/hamogu/sherpa-benchmarks that serve the same purpose. Of course, more can always be added, but I think I can close this general issue here.

PR #1414, opened 2022-01-25
Improve the handling of image coordinates

### LINKED ISSUE #1380: issues with contours
I'm looking at the coordinates and contour issues which have been listed on the CIAO Sherpa web site
https://cxc.harvard.edu/sherpa/bugs/contour.html

In CIAO 4.14 after setting coordinates to 'world' and resetting back to 'physical' results in an error. The contours are plotted ok before any `set_coord` updates. In the example below the first contour('data') is shown, the rest shows the errors:

```
herpa
Sherpa 4.14.0

Python 3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:50:56) 
IPython 7.30.1 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Using matplotlib backend: MacOSX

sherpa In [1]: load_data('/Users/aneta/git/sherpa-test-data/sherpatest/image2.fits')

sherpa In [2]: show_data()

sherpa In [3]: contour('data')

sherpa In [4]: set_coord('logical')

sherpa In [5]: contour('data')

sherpa In [6]: set_coord('world')

sherpa In [7]: contour('data')
NotImplementedErr: contours on non-uniform grids are not yet supported

sherpa In [8]: set_coord('logical')

sherpa In [9]: contour('data')
NotImplementedErr: contours on non-uniform grids are not yet supported

sherpa In [10]: set_coord('physical')

sherpa In [11]: contour('data')
NotImplementedErr: contours on non-uniform grids are not yet supported

```

DougBurke:
# Summary

Address an error when converting between different coordinate systems for images. Fixes #1380.

The coord attribute of an image (DataIMG, DataIMGInt) can no longer be changed directly. This means that

    img.coord = "physical"

must be changed to

    img.set_coord("physical")

to ensure that the independent axes are calculated correctly.

# Details

Most of the commits are actually minor code clean ups and adding of tests, including ones that show #1380 and the underlying problem that caused this bug.

The first functional change comes in `The coord attribute is now readonly for DataIMG` which stops the `coord` attribue from being settable. At first glance this would appear to be a huge loss in functionaluty, but looking at the code we can see that the user should be using the `set_coord` method of the object rather than directly changing the `coord` attribute. This is because `set_coord` applies a number of changes to the object, such as actually updating the independent axes to match the selected coordinate system. Unfortunately we can't just move this logic into the `_set_coord` routine (which was the previous method used to set the attribute), as we need to be able to change the `coord` attribute during the `__init__` call, when we **DO NOT** want to do the extra work done by `set_coord`. I had previously added a comment to point out that a user being able to set `coord` attribute directly would lead to user confusion, so we now dis-allow it. Fortunately there are no uses of this functionality in the Sherpa code base (for example the UI set_coord routine calls the object's set_coord method). This is a breaking change for downstream users, but I argue that if they had used code like `img.coord = "physical"` then they were now in a state of getting undefined behavior, so it is good that this now fails.

As I write the above I've thought of possible ways that we could make lines like `img.coord = "physical"` work - which would essentially remove the need for the `set_coord` method - but I have not worked through the consequences. Also, we don't technically need this change in this PR as it is technically unrelated to #1380.

The fix for #1380 comes in the appropriately-named "Fix #1380" commit. The reason for the bug is given at https://github.com/sherpa/sherpa/issues/1380#issuecomment-1021400579 and the fix follows the approach mentioned by @hamogu at https://github.com/sherpa/sherpa/issues/1380#issuecomment-1021437084 

- when the object is created we store the original coordinate settings (the coordinate setting and the x0 and x1 arrays) in the `_orig_indep_axis` attribute
- when converting to a particular coordinate system - with the `get_logical`, `get_physical`, and `get_world` methods - we use the `_orig_indep_axis` data to calculate the requested coordinate system (or copy it when they are the same)
- `set_coord` calls the correct `get_logical/physical/world` routine, and so we are always creating the data in the same way

Prior to this commit the `get_logical/physical/world` routines would take the current coordinate system and use it to create the new coordinate system. That means that the sequence

    img.set_coord("physical")
    img.set_coord("world")
    img.set_coord("logical")

would convert logical to physical, then physical to world, then world to logical. This last step turned out to be lossy and so the calculated logical coordinates were now some small amount different than their original values. This would then mean that each "pixel" could now have widths/heights slightly-different from 1, resulting in the contour error (as it requires all the pixels to have the same width/height).

The new approach technically still has this problem, but only if the object is created using the world axes as it's inputs (since then the conversion from world to logical is likely to cause this problem). However, this is not the way we expect the objects to be created (although we do not dis-allow it).

One disadvantage to the new scheme - which was hinted at in a comment removed in this PR - is that we are increasing the memory use of DataMG objects to solve this problem. We now store the original x0, x1 arrays in the object, as well as the computed values when prior to this we only had to store the computed values.

This change does make loading in an old pickle file with an image difficult, as we did not store the original independent axes. To try and support this we add a `_orig_indep_axis` attribute if necessary when un-pickling, and set it to the currently-selected axis. This allow the `set_coord` method to be called on the object but can still show the #1380 data (but only for data unpickled from an old store file). There is no explicit test (I do not think it worth adding a picked file from 4.14.0 to the test data area), but I do add some basic pickle tests to check the current behavior. I have tested it manually. 

Could we solve this in some other way? Well, the problem comes from the fact that we can not guarantee that the WCS transformations are "perfectly invertible": i.e. that invert(apply(x, y)) is not guaranteed to exactly give x,y thanks to floating-point issues, particularly when combining the physical and world systems. We could do something like "always convert logical values to integer values", **but** we do not require that the x0/y0 values are integers when creating a `DataIMG` object, and I am not sure it would be a good change to make.

We could make it so that the `__init__` call is only ever sent in the `logical` coordinate system (that is, we don't allow `coord` as an argument to the constructor) which would then mean we could get rid of the `_physical_to_*` and `_world_to_*` methods (since we only ever call these via `_get_coordsys` and it depends on the coordinate system set when the object was created). However, I think that's for a different PR.

# Example of coord issue

In CIAO 4.14 we see the following

```
sherpa-4.14.0> load_image("image2.fits")

sherpa-4.14.0> img = get_data()

sherpa-4.14.0> l0 = img.get_x0()

sherpa-4.14.0> img.coord
'logical'

sherpa-4.14.0> l0[0:10]
array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

sherpa-4.14.0> img.set_coord("physical")

sherpa-4.14.0> p0 = img.get_x0()

sherpa-4.14.0> p0[0:10]
array([3799.5, 3801.5, 3803.5, 3805.5, 3807.5, 3809.5, 3811.5, 3813.5,
       3815.5, 3817.5])

sherpa-4.14.0> img.coord = "logical"

sherpa-4.14.0> img.get_x0()[0:10]
array([3799.5, 3801.5, 3803.5, 3805.5, 3807.5, 3809.5, 3811.5, 3813.5,
       3815.5, 3817.5])
```

So, the last call has returned the "physical" system when it should have returned the logical units (i,e, `l0`). This is because the `img.coord = "logical"` call does not do the same setup as `img.set_coord("logical")` does. 

And then once we've done this we are "messed up" in that it's not obvious how to get the system back into a sensible state.

```
sherpa-4.14.0> img.set_coord("logical")

sherpa-4.14.0> img.get_x0()[0:10]
array([3799.5, 3801.5, 3803.5, 3805.5, 3807.5, 3809.5, 3811.5, 3813.5,
       3815.5, 3817.5])
```

With this PR we can no-longer change coord except via `set_coord` so it works out correctly:

```

In [2]: load_image("image2.fits")

In [3]: img = get_data()

In [4]: img.get_x0()[0:10]
Out[4]: array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

In [5]: img.coord
Out[5]: 'logical'

In [6]: img.set_coord("physical")

In [7]: img.get_x0()[0:10]
Out[7]:
array([3799.5, 3801.5, 3803.5, 3805.5, 3807.5, 3809.5, 3811.5, 3813.5,
       3815.5, 3817.5])

In [8]: img.coord = "logical"
Input In [8], in <module>
----> 1 img.coord = "logical"

File ~/sherpa/sherpa-main/sherpa/utils/__init__.py:171, in NoNewAttributesAfterInit.__setattr__(self, name, val)
    166     elif not callable(getattr(self, name)) and callable(val):
    167         raise AttributeError(("'%s' object attribute '%s' cannot be " +
    168                               "replaced with a callable attribute") %
    169                              (type(self).__name__, name))
--> 171 object.__setattr__(self, name, val)

In [9]: img.get_x0()[0:10]
Out[9]:
array([3799.5, 3801.5, 3803.5, 3805.5, 3807.5, 3809.5, 3811.5, 3813.5,
       3815.5, 3817.5])

In [10]: img.coord
Out[10]: 'physical'

In [11]: img.set_coord("logical")

In [12]: img.coord
Out[12]: 'logical'

In [13]: img.get_x0()[0:10]
Out[13]: array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])
```

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -935,3 +935,187 @@ We can view the model plot object::
     ylabel = y
     title  = Model
     histo_prefs = {'yerrorbars': False, 'ecolor': None, ... , 'linecolor': None}
+
+
+.. _dataimg_coords:
+
+Coordinate conversion for image data
+------------------------------------
+
+The :py:class:`sherpa.data.Data2D` class provides basic support for
+fitting models to two-dimensional data; that is, data with two
+independent axes (called "x0" and "x1" although they should be
+accessed via the :py:attr:`~sherpa.data.Data2D.indep` attribute).  The
+:py:class:`sherpa.astro.data.DataIMG` class extends the 2D support to
+include the concept of a coordinate system, allowing the independent
+axis to be one of:
+
+- ``logical``
+- ``image``
+- ``world``
+
+where the aim is that the logical system refers to a pixel number (no
+coordinate system), image is a linear transform of the logical system,
+and world identifies a projection from the image system onto the
+celestial sphere. However, there is no requirement that this
+categorization holds as it depends on whether the optional
+:py:attr:`~sherpa.astro.data.DataIMG.sky` and
+:py:attr:`~sherpa.astro.data.DataIMG.eqpos` attributes are set when
+the :py:class:`~sherpa.astro.data.DataIMG` object is created.
+
+Using a coordinate system directly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the independent axes are known, and not calculated via a coordinate
+transform, then they can just be set when creating the
+:py:class:`~sherpa.astro.data.DataIMG` object, leaving the
+:py:attr:`~sherpa.astro.data.DataIMG.coord` attribute set to
+``logical``.
+
+  >>> x0 = np.asarray([1000, 1200, 2000])
+  >>> x1 = np.asarray([-500, 500, -500])
+  >>> y = np.asarray([10, 200, 30])
+  >>> d = DataIMG("example", x0, x1, y)
+  >>> print(d)
+  name      = example
+  x0        = Int64[3]
+  x1        = Int64[3]
+  y         = Int64[3]
+  shape     = None
+  staterror = None
+  syserror  = None
+  sky       = None
+  eqpos     = None
+  coord     = logical
+
+This can then be used to evaluate a two-dimensional model,
+such as :py:class:`~sherpa.models.basic.Gauss2D`:
+
+  >>> from sherpa.models.basic import Gauss2D
+  >>> mdl = Gauss2D()
+  >>> mdl.xpos = 1500
+  >>> mdl.ypos = -100
+  >>> mdl.fwhm = 1000
+  >>> mdl.ampl = 100
+  >>> print(mdl)
+  gauss2d
+     Param        Type          Value          Min          Max      Units
+     -----        ----          -----          ---          ---      -----
+     gauss2d.fwhm thawed         1000  1.17549e-38  3.40282e+38
+     gauss2d.xpos thawed         1500 -3.40282e+38  3.40282e+38
+     gauss2d.ypos thawed         -100 -3.40282e+38  3.40282e+38
+     gauss2d.ellip frozen            0            0        0.999
+     gauss2d.theta frozen            0     -6.28319      6.28319    radians
+     gauss2d.ampl thawed          100 -3.40282e+38  3.40282e+38
+  >>> d.eval_model(mdl)
+  array([32.08564744, 28.71745887, 32.08564744])
+
+Attempting to change the coordinate system with
+:py:class:`~sherpa.astro.data.DataIMF.set_coord` will error out with a
+:py:class:`sherpa.utils.err.DataErr` instance reporting that the data
+set does not specify a shape.
+
+The shape attribute
+^^^^^^^^^^^^^^^^^^^
+
+The ``shape`` argument can be set when creating a
+:py:class:`~sherpa.astro.data.DataIMG` object to indicate that the
+data represents an "image", that is a rectangular, contiguous, set of
+pixels. It is defined as ``(nx1, nx0)``, and so matches the ndarray
+``shape`` attribute from NumPy. Operations that treat the dataset as a
+2D grid often require that the ``shape`` attribute is set.
+
+  >>> x1, x0 = np.mgridp1L4, 1:5]
+  >>> y2 = (x0 - 2.5)**2 + (x1 - 2)**2
+  >>> y = np.sqrt(y2)
+  >>> d = DataIMG('img', x0.flatten(), x1.flatten(),
+  ...             y.flatten(), shape=y.shape)
+  >>> print(d)
+  name      = img
+  x0        = Int64[12]
+  x1        = Int64[12]
+  y         = Float64[12]
+  shape     = (3, 4)
+  staterror = None
+  syserror  = None
+  sky       = None
+  eqpos     = None
+  coord     = logical
+  >>> d.get_x0()
+  array([1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4])
+  >>> d.get_x1()
+  array([1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3])
+  >>> d.get_dep()
+  array([1.80277564, 1.11803399, 1.11803399, 1.80277564, 1.5       ,
+         0.5       , 0.5       , 1.5       , 1.80277564, 1.11803399,
+         1.11803399, 1.80277564])
+  >>> d.get_axes()
+  (array([1., 2., 3., 4.]), array([1., 2., 3.]))
+  >>> d.get_dims()
+  (4, 3)
+
+Attempting to change the coordinate system with
+:py:class:`~sherpa.astro.data.DataIMF.set_coord` will error out with a
+:py:class:`sherpa.utils.err.DataErr` instance reporting that the data
+set does not contain the required coordinate system.
+
+Setting a coordinate system
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The :py:class:`sherpa.astro.io.wcs.WCS` class is used to add a
+coordinate system to an image. It has support for linear (translation
+and scale) and "wcs" - currently tangent-plane projections -
+conversions.
+
+  >>> from sherpa.astro.io.wcs import WCS
+  >>> sky = WCS("sky", "LINEAR", [1000,2000], [1, 1], [2, 2])
+  >>> x1, x0 = np.mgrid[1:3, 1:4]
+  >>> d = DataIMG("img", x0.flatten(), x1.flatten(), np.ones(x1.size), shape=x0.shape, sky=sky)
+  >>> print(d)
+  name      = img
+  x0        = Int64[6]
+  x1        = Int64[6]
+  y         = Float64[6]
+  shape     = (2, 3)
+  staterror = None
+  syserror  = None
+  sky       = sky
+   crval    = [1000.,2000.]
+   crpix    = [1.,1.]
+   cdelt    = [2.,2.]
+  eqpos     = None
+  coord     = logical
+
+With this we can change to the "physical" coordinate system, which
+represents the conversion sent to the ``sky`` argument, and so get the
+independent axis in the converted system with the
+:py:meth:`sherpa.astro.data.DataIMG.set_coord` method:
+
+  >>> d.get_axes()
+  (array([1., 2., 3.]), array([1., 2.]))
+  >>> d.set_coord("physical")
+  >>> d.get_axes()
+  (array([1000., 1002., 1004.]), array([2000., 2002.]))
+  >>> d.indep
+  (array([1000., 1002., 1004., 1000., 1002., 1004.]), array([2000., 2000., 2000., 2002., 2002., 2002.]))
+
+It is possible to switch back to the original coordinate system (the
+arguments sent in as ``x0`` and ``x1`` when creating the object):
+
+  >>> d.set_coord("logical")
+  >>> d.indep
+  (array([1, 2, 3, 1, 2, 3]), array([1, 1, 1, 2, 2, 2]))
+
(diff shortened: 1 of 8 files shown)


DougBurke:
Rebased since GH thought there was a conflict.

docs/developer/index.rst
```diff
@@ -935,3 +935,187 @@ We can view the model plot object::
...
+If the independent axes are known, and not calculated via a coordinate
+transform, then they can just be set when creating the
+:py:class:`~sherpa.astro.data.DataIMG` object, leaving the
+:py:attr:`~sherpa.astro.data.DataIMG.coord` attribute set to
+``logical``.
+
+  >>> x0 = np.asarray([1000, 1200, 2000])
+  >>> x1 = np.asarray([-500, 500, -500])
+  >>> y = np.asarray([10, 200, 30])
+  >>> d = DataIMG("example", x0, x1, y)
```
hamogu, line 978: Usually, images are squares. Here, it's some kind of odd triangle. It's kind of cute that that works, but (a) I'm not sure that's a good idea in general and (b) I'm almost sure it's not a good idea to use that as the canonical example. Instead, I would move this example to the end of the discussion after the reader has seen the "normal" case or alternatively, add a sentence like "This is not the normal case for an image, hang on and see below"

docs/developer/index.rst
```diff
@@ -935,3 +935,187 @@ We can view the model plot object::
...
+^^^^^^^^^^^^^^^^^^^
+
+The ``shape`` argument can be set when creating a
+:py:class:`~sherpa.astro.data.DataIMG` object to indicate that the
+data represents an "image", that is a rectangular, contiguous, set of
+pixels. It is defined as ``(nx1, nx0)``, and so matches the ndarray
+``shape`` attribute from NumPy. Operations that treat the dataset as a
+2D grid often require that the ``shape`` attribute is set.
+
+  >>> x1, x0 = np.mgridp1L4, 1:5]
```
hamogu, line 1028: typo. (that's mey motivation for #813 but we'll see when I get to that).

docs/developer/index.rst
```diff
@@ -935,3 +935,187 @@ We can view the model plot object::
...
+
+Attempting to change the coordinate system with
+:py:class:`~sherpa.astro.data.DataIMF.set_coord` will error out with a
+:py:class:`sherpa.utils.err.DataErr` instance reporting that the data
+set does not contain the required coordinate system.
+
+Setting a coordinate system
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The :py:class:`sherpa.astro.io.wcs.WCS` class is used to add a
```
hamogu, line 1065:
```suggestion
The `sherpa.astro.io.wcs.WCS` class is used to add a
```
should work, too. (Same in other places, but I'm only commenting once). One advantage is that it's a little easier to read the plaintext rst because there is less markup in it. I admit, it matter more in docstrings in the code, because those are read in plain text much more often than rst files...

docs/developer/index.rst
```diff
@@ -935,3 +935,187 @@ We can view the model plot object::
...
+coordinate system to an image. It has support for linear (translation
+and scale) and "wcs" - currently tangent-plane projections -
```
hamogu, line 1067:
```suggestion
and scale) and "wcs" - currently only tangent-plane projection -
```

docs/developer/index.rst
```diff
@@ -935,3 +935,187 @@ We can view the model plot object::
...
+the selected coordinate system (that is, the argument to the
+`set_coord` call). This had the advantage of saving memory, as you
+only needed to retain the current pair of independent axes, but at the
+expense of losing fidelity when converting between the coordinate
+systems. This has been changed so that the original independent axes
+are now stored in the object, in the ``_orig_indep_axis`` attribute,
+and this is now used whenever the coordinate system is changed. This
+does increase the memory size of a `DataIMG` object, and makes it
+harder to load in picked files created with an old Sherpa version (the
+code will do it's best to create the necessary information but it is
```
hamogu, line 1120:
```suggestion
code will do its best to create the necessary information but it's
```

sherpa/astro/data.py
```diff
@@ -4378,12 +4397,28 @@ def __setstate__(self, state):
         # self.__dict__['_get_physical']=(lambda : None)
         # self.__dict__['_get_world']=(lambda : None)
 
+        # Unfortunately we can't re-create the original data used to
+        # create the object, but we can fake it using the last-selected
+        # coordinate system (which could lead to some issues for the
+        # world sytem - see #1390 - but there's little we can do here).
```
hamogu, line 4403:
```suggestion
        # world system - see #1380 - but there's little we can do here).
```

sherpa/astro/data.py
```diff
@@ -4494,50 +4529,60 @@ def _world_to_physical(self, x0=None, x1=None):
...
-        # Destroys original data to conserve memory for big imgs
         good = ('logical', 'image', 'physical', 'world', 'wcs')
         if coord not in good:
             raise DataErr('badchoices', 'coordinates', coord, ", ".join(good))
 
-        if coord.startswith('wcs'):
+        if coord == 'wcs':
             coord = 'world'
-        elif coord.startswith('image'):
+        elif coord == 'image':
```
hamogu, line 4579: Technically, this is an API breaking change, because there could be existing user code that used `set_coord("image_crap")`.... but I think it's good to be explicit!

Just commenting because I recently rediscovered https://xkcd.com/1172/ ;-)

hamogu (changes requested):
Really minor changes requested.

docs/developer/index.rst
DougBurke: There is an interesting discussion to be had on what we mean for the `Data2D` and `DataIMG` cases - perhaps the `DataIMG` case should require a rectangular grid, but that's a different discussion than this PR.

I'll have to look at improving the docs tomorrow.

sherpa/astro/data.py
DougBurke: Actually, given the check just above we know that if we got here and trigger the image then the argument could not have been `"image_your_base_belong_to_us"` - see the error check on lines 4572 to 4575

docs/developer/index.rst
hamogu: Agreed. This is absolutely not what I had expected and I wonder if that's an accident of the design or if there is a real use case. Either way, this PR is certainly not the place to change that!

docs/developer/index.rst
DougBurke: I see what you did there.

docs/developer/index.rst
DougBurke: I looked at replacing this with a grid, but I think it is important to let you know you don't need a grid. So I've added a note to the start of this little section pointing this out.

DougBurke:
I've done some work on the developer docs; not quite what was asked but hopefully okay.

I also rebased the code just to make sure we don't have any interaction with the recently-merged PRs which change the data code and tests.

DougBurke:
Before approval I'd like a quick discussion to see whether the API change here - that is

    d.image = "physical"
    
is now an error is okay, or if I should allow this by essentially making it the same as `d.set_image("physical")` but with an extra check for when called from within `__init__`.

One "nice" side effect from making the code above fail is that users will get explicit feed-back that something has changed. If we make `d.image` silently work correctly then it could lead to some subtle differences that might be hard to identify or understand.

On the other side. I would be mildly surprised if people are using `d.image = ...` in a place where it could break code, so maybe forcing calling code to error out here is not so friendly.

I hope this is a real corner case - i.e. not many users are going to be affected by this, but I can not guarantee it.

hamogu:
It's good to consider strict API backwards compatibility, but in this case I think that it's not a big worry here. As far as I can see, imaging is the less used part of Sherpa. When I see sherpa plots in publications, it's almost always spectral fitting. Maybe I'm biased by my own area of science, but we also don't get a large number of tickets of GH issues for 2D fitting, although we know bugs are present.

All thin points to the fact that there is only a small number of users at the moment AND this change is relatively benign. It's not that old code will suddenly give wrong results silently - quite the opposite in fact.
So, I think that clarity for the future clearly outweighs the minor API change for current users of this feature.

DougBurke:
Rebased to address conflict.

hamogu (approved):
My concerns have been addressed and I agree with the minor API change. I'm going on vacation tomorrow and don't want this to linger until I come back, but because I originally requested changes.

If @anetasie wants to weigh in on the API change, she can "request changes" herself.

PR #1382, opened 2021-12-13
Plot changes: Step 1 A- Draft API for backend classes

hamogu:
# Summary
Sherpa's plotting backends are now classes, not modules.

# Details
This is the first half of the "step 1" described in #1375. It's probably the biggest PR of that series and contains a number of changes and additions:

- Define a base class for backends. That base class essentially defines the API. (I'm not making this an abstract base class though, because it's useful for testing to run plotting directly with this class.)
- Define a `BasicBackend` that derives from the base class and adds warnings for for any operation that does not actually produce output. `BasicBackend` can be used for tests (and is in many cases), but because it issues extra warnings, it messes up tests that capture the output to check specific warnings.
- Define a set of back-end independent plot options. This is written down in the docs and implemented in a test class that will ensure that our defaults only use backend-independent options.
- A context manager to temporarily change the backend.
- Separate backend definitions in the datastack module are removed.
- In the old implementation of plot objects, the defaults for those objects are backend dependent and set at import time, i.e. for whatever backend happens to be active when that class is imported first. Since all classes are defined in `plot/__init__.py` that backend is the first one in `sherpa.rc` that successfully imports. I have ides to address that with more detail in #1375 (in particular, I think those defaults should be set in __init__ based on the backend that's active when an object is initialized, not when the class is imported), but for now, I'm setting the defaults from the `BaseBackend`, i.e. they only contain values that should work in any backend. That means that the defaults for some objects change. I don't think that matters in practice, because (i) any code that sets those options to something else will work, e.g. setting `hspace=0.5` for a SplitPlot will work just as before if the backend is matplotlib (just as before). If nothing is set, then the `SplitPlot.prefs` will be empty, which means that whatever is set a default in the backend is used. The only difference is that, with this PR, I can't do `print(SplitPlot().prefs)` to discover how those options are called. I think that is an acceptable change given that I want to address that in a follow up PR anyway; I think this minor change is a better solution than making this PR even more complex.
- Numerous changes to tests to test the new classes, in particular two new text fixtures (`all_plotting_backends` and `plotting_backends`) that largely replace the `requires_pylab` decorator.
- Docs on backends.
- Removes the "point" plotting function (#1358)
- Adds "label" and "linewidth" to the list of backend-independent options (#938). That is @DougBurke work but I found it hard to use those commits to give him credit since it's essentially implemented new here. So, I hope a mention in this text is enough.
- Define a "registry" of backend where backend classes are keyed to string names.

To be done:
[...]

PR #1415, opened 2022-01-25
RTD fix missing and changed symbols

DougBurke:
# Summary

Update the ReadTheDocs build to cope with changes in #1191

# Details

PR #1191 removed these symbols, so remove them from the documentation, and we don't seem to pick up the backend documentation so remove it.

Found when building the docs for an unrelated PR.

### DIFF
--- a/docs/overview/astro_io.rst
+++ b/docs/overview/astro_io.rst
@@ -8,12 +8,6 @@ The sherpa.astro.io module
 
    .. rubric:: Attributes
 
-   .. autosummary::
-      :toctree: api
-
-      backend
-
-
    .. rubric:: Functions
 
    .. autosummary::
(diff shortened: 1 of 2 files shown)


hamogu (approved):
Fine with me - the entire backend documentation will be redone in step #1382 anyway. If we merge this soon - fine. If not, this might be superceeded in #1832 anyway.

(By the way, I'm looking at sphinx configuration improvements (#1411 was the first PR to come out of that) with the eventual goals to (a) build (most of ) the plots in the docs and doc build time instead of including pre-generated pngs and (b) activate "sphinx warnigns as error" so that the build will actually fail if a PR messes up the sphinx docs. But, there is a long way to go for that...

PR #1418, opened 2022-01-26
Fix minor test issues

DougBurke:
# Summary

Address several minor test issues.

# Details

These were noticed when working on #1414. They look like a combination of cut-n-paste confusion and general-confusion-with-the-arguments-to-Data2DInt

### DIFF
--- a/sherpa/tests/test_data.py
+++ b/sherpa/tests/test_data.py
@@ -1011,7 +1011,9 @@ def test_data2d_int_eval_model_to_fit(array_sizes_fixture):
     from sherpa.models import Gauss2D
 
     x0, x1, dx, y = array_sizes_fixture
-    data2 = Data2DInt('name', x0.flatten(), x0.flatten() + dx, x1.flatten(), x1.flatten() + dx, y.flatten(),
+    data2 = Data2DInt('name', x0.flatten(), x1.flatten(),
+                      x0.flatten() + dx, x1.flatten() + dx,
+                      y.flatten(),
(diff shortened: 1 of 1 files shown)


sherpa/tests/test_data.py
```diff
@@ -707,19 +707,17 @@ def test_data2_get_x0(data):
...
 def test_data2_get_x1(data):
-    numpy.testing.assert_array_equal(data.get_x0(), X0_2D)
+    numpy.testing.assert_array_equal(data.get_x1(), X1_2D)
 
 
 @pytest.mark.parametrize("data", (Data2DInt, ), indirect=True)
 def test_data2_int_get_x1(data):
-    actual = data.get_x1()
-    numpy.testing.assert_array_equal(actual, X1_2D)
+    numpy.testing.assert_array_equal(data.get_x1(), X1_2D)
```
hamogu, line 720: I note that these two tests can now we summarized as:
```
@pytest.mark.parametrize("data", Data2DClasses, indirect=True)
def test_data2_get_x1(data):
    numpy.testing.assert_array_equal(data.get_x1(), X1_2D)
```
That's not important (it only saves three lines) so I'm approving, but if @DougBurke wants to adjust that before merging thats also fine.

hamogu (approved):

sherpa/tests/test_data.py
DougBurke: There's a number of other tests that would benefit from this cleanup, but I don't have the time/energy to do this so I'd rather not do it here.

Issue #1379, opened 2021-12-09
AttributeError: 'DataIMGInt' object has no attribute 'x0lo'


anetasie:
I'm checking the Sherpa/CIAO bug list and the advice for fitting `delta2d` model is to change the `dstype` to `DataIMGInt` or `Data2DInt`. In both cases I get the AttributeError in CIAO 4.14. The error is also thrown after reading fits image data.

```
-----------------------------------------------------
Welcome to Sherpa: CXC's Modeling and Fitting Package
-----------------------------------------------------
Sherpa 4.14.0

Python 3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:50:56) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.30.1 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Using matplotlib backend: MacOSX

sherpa In [1]: dataspace2d([100,100], dstype=DataIMGInt)

sherpa In [2]: show_data()
AttributeError: 'DataIMGInt' object has no attribute 'x0lo'

sherpa In [3]: dataspace2d([100,100], dstype=Data2dInt)
NameError: name 'Data2dInt' is not defined

sherpa In [4]: dataspace2d([100,100], dstype=Data2DInt)

sherpa In [5]: show_data()
AttributeError: 'Data2DInt' object has no attribute 'x0lo'
```

Here is the link to the delta2d issue:
https://cxc.harvard.edu/sherpa/bugs/delta2d.html

DougBurke:
The `*Int` classes for the 2D case have seen essentially no testing or use, so it doesn't surprise me things don't work well.

DougBurke:
My guess is that something in #614 caused this. I'm trying to set up some tests to work out what is going on.

DougBurke:
Just to note - unrelated to this bug - that using Data2DInt/DataIMGInt doesn't really help. As with the Delta1D case for PHA data, the problem is that within a single bin/pixel the model evaluates to a constant, so the optimiser can't improve the fit.

PR #1421, opened 2022-01-27
Address usability issues with Data2DInt and DataIMGInt classes

DougBurke:
# Summary

Fix issues using the Data2DInt and DataIMGInt classes (#1379).

# Details

The first few commits include a number of tests to check basic features of the Data2DInt and DataIMGInt classes. Some of them are marked as xfail and these are fixed in later commits. Some of the tests are left as regression tests, so they will indicate if the current behavior has changed, as there are places where it's not obvious what the correct behavior is (in particular the get_axes call).

The immediate cause of #1379 is that the `__str__` method for the two classes could not find the `x0lo`, `x1lo`, `x0hi`, and `x1hi` attributes. It is likely that this was broken during the rework to the `Data` hierarchy in #614 but I have not verified this. In looking at the code it turns out that a number of other routines, related to accessing the independent axes - such as the `x0` attribute and the `get_x1()` call - are also broken because of these missing attributes.  This was found by adding the tests in the first few commits.

There are a few docstring changes and comments added before we get to the fixes.

The first set of fixes is in the `Fix show error for Data2DInt` commit. This is a deceptively simple fix, which is adding attribute accessors like the following to the `Data2DInt` class:

```
    @property
    def x0lo(self):
        """
        Property kept for compatibility
        """
        return self._data_space.x0lo
```

The docstring is based on the `xlo`/`xhi` accessors for the `Data1Dint` class, and it suggests that we are meant to use some other way to get at this data, but accessing the `_data_space` attribute directly can not be that, and there's no obvious way to get at this, unless it's meant to be the `get_evaluation_indep` method, but this is an undocumented method in the `Data1D` class and isn't defined in the `Data2D` class. Until we have this sorted out using these attributes seems to be the best way to get the code to work. Note that the number of xfail tests has dropped.

The next change in `Fix independent-axis access for DataIMGInt` is essentially the same for the `DataIMGInt` class, but we also need to override the `get_x0` and `get_x1` methods from the parent `Data2D` class (to match what `Data2DInt` does). Perhaps we could have some abstract class to add this "integrated" behavior so we don't need to replicate this, but that is a significantly-larger change than warranted by this issue. Note that the number of xfail tests has dropped.

The last commit just makes a few style changes to DataIMG/2D related methods. The idea is to try and exit a method quickly when it's the "easy" path. This makes the logic for the more-complex path easier to see (in part because we remove four spaces). I don't do too much of this as there are a number of similar changes in #1414 and I don't want to make this to annoying to merge.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4742,6 +4742,42 @@ class DataIMGInt(DataIMG):
     def _init_data_space(self, filter, *data):
         return IntegratedDataSpace2D(filter, *data)
 
+    def get_x0(self, filter=False):
+        indep = self._data_space.get(filter)
+        return (indep.x0lo + indep.x0hi) / 2.0
+
+    def get_x1(self, filter=False):
+        indep = self._data_space.get(filter)
+        return (indep.x1lo + indep.x1hi) / 2.0
+
+    @property
+    def x0lo(self):
+        """
+        Property kept for compatibility
+        """
+        return self._data_space.x0lo
+
+    @property
+    def x0hi(self):
+        """
+        Property kept for compatibility
+        """
+        return self._data_space.x0hi
+
+    @property
+    def x1lo(self):
+        """
+        Property kept for compatibility
+        """
+        return self._data_space.x1lo
+
+    @property
+    def x1hi(self):
+        """
+        Property kept for compatibility
+        """
+        return self._data_space.x1hi
+
     def get_logical(self):
         coord = self.coord
         x0lo, x1lo, x0hi, x1hi = self.get_indep()
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4542,9 +4542,10 @@ class DataIMG(Data2D):
         self._set_coord(coord)
 
     def get_filter_expr(self):
-        if self._region is not None:
-            return str(self._region)
-        return ''
+        if self._region is None:
+            return ""
+
+        return str(self._region)
 
     get_filter = get_filter_expr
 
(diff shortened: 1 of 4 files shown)


sherpa/astro/data.py
```diff
@@ -4577,44 +4578,42 @@ def notice2d(self, val=None, ignore=False):
...
-            mask = mask[x0_lo:x0_hi + 1, x1_lo:x1_hi + 1]
+        x0_lo = x0_i.min()
+        x0_hi = x0_i.max()
+        x1_lo = x1_i.min()
+        x1_hi = x1_i.max()
 
-            mask = mask.ravel()
+        # TODO: subset mask and then ask its shape
+        shape = mask[x0_lo:x0_hi + 1, x1_lo:x1_hi + 1].shape
+        mask = mask[x0_lo:x0_hi + 1, x1_lo:x1_hi + 1]
```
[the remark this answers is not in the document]
DougBurke: `shape` is a tuple and so is immutable, so a copy isn't needed (and doesn't need have a copy method), but the other changes make sense.

DougBurke:
This is more because I now can than because it's needed for this issue, but here's the class hiearchy.

```mermaid
graph TD;
    Data --> Data2D;
    Data2D --> DataIMG;
    Data2D --> Data2DInt;
    DataIMG --> DataIMGInt;
```

It might be nice to have some form of an interface to say "this is what being a `*Int` class is about", but that is way outside the scope of this PR.

sherpa/astro/data.py
```diff
@@ -4577,44 +4578,42 @@ def notice2d(self, val=None, ignore=False):
...
-        if yfunc is not None:
-            y_img = (y_img[0].reshape(*self.shape),
-                     y_img[1].reshape(*self.shape))
-        else:
-            y_img = y_img.reshape(*self.shape)
+        m = self.eval_model_to_fit(yfunc)
+        if numpy.iterable(self.mask):
+            # if filtered, the calculated model must be padded up
+            # to the data size to preserve img shape and WCS coord
+            m = pad_bounding_box(m, self.mask)
```
DougBurke, line 4611: This line, which was not "really" changed in this PR, is not tested, so I'm about to add a test to cover this.

Of course this test lead to issue #1447 but at least we will now have explicit tests of this code so we can see when there's any changes.

DougBurke: I've now added a bunch of tests so this line should now get covered.I can't wait to find out...

DougBurke: And it is now marked as tested.

sherpa/astro/data.py
```diff
@@ -4577,44 +4578,41 @@ def notice2d(self, val=None, ignore=False):
...
+
+        y_img = (y_img, self.filter_region(m))
+
+        y_img = (y_img[0].reshape(*self.shape),
+                 y_img[1].reshape(*self.shape))
```
DougBurke, line 4615: This should probably be re-written (a little bit more-complex than just these four lines)

sherpa/astro/data.py
```diff
@@ -4701,6 +4699,42 @@ def __init__(self, name, x0lo, x1lo, x0hi, x1hi, y, shape=None,
...
 
+    def get_x0(self, filter=False):
+        indep = self._data_space.get(filter)
+        return (indep.x0lo + indep.x0hi) / 2.0
+
+    def get_x1(self, filter=False):
+        indep = self._data_space.get(filter)
+        return (indep.x1lo + indep.x1hi) / 2.0
+
+    @property
```
DougBurke, line 4710: coverage claims that line 4710 was not covered by the tests, which seems a tad odd if line 4711 is covered.

sherpa/astro/data.py
DougBurke: Which I've now just done.

DougBurke:
Rebased to ensure it still works with recent data-related PRs.

hamogu (approved):
As far as I can see all my comments have been addressed.

Issue #1392, opened 2022-01-06
remove the primini method for iterated fitting


DougBurke:
The primini method has been broken for a long while - see #431 - and the original author of the method has suggested that we should remove it from Sherpa. This is a reminder to do so.

PR #1424, opened 2022-02-02
rm primini stat method (issue #1392)

dtnguyen2:
# Note
This PR removes the ```primini``` iterative statistical method.  Please see issue #1392 for details


sherpa/ui/utils.py
```diff
@@ -1862,12 +1855,7 @@ def set_iter_method(self, meth):
                322
                http://adsabs.harvard.edu/abs/1995ApJ...438..322W
 
-        .. [2] "Bias-Free Parameter Estimation with Few Counts, by
-               Iterative Chi-Squared Minimization", Kearns, Primini, &
-               Alexander, 1995, ADASS IV, 331
-               http://adsabs.harvard.edu/abs/1995ASPC...77..331K
-
-        .. [3] http://iraf.net/irafhelp.php?val=sfit
+        .. [2] http://iraf.net/irafhelp.php?val=sfit
```
DougBurke, line 1858: Line 1844 - unfortunately I can't seem to change it on the GH page - will need to be changed to use `[2]` rather than `[3]` - i.e.

```
        [2]_, where after a fit data points are excluded if the value
```

DougBurke:
I'd normally suggest we add some "hey, we've removed the primini method" text added to the documentation, in particular `set_iter_method`, but given that it's been broken for so long I don't think we need to.

Thoughts @anetasie ?

sherpa/ui/utils.py
dtnguyen2: done

anetasie:
I think we do need some kind of documentation warning - say that the method was deprecated when somebody tries to set.
What the result of "list_iter_methods()" with the PR?

DougBurke:
It should just return the following now (not tested, but the documentation has been updated to reflect this).

```
>>> list_iter_methods()
['none', 'sigmarej']
```

DougBurke:
So, with this branch we see that `list_iter_methods` no-longer mentions `primini` and you can't set it:

```

In [2]: list_iter_methods()
Out[2]: ['none', 'sigmarej']

In [3]: set_iter_method("primini")
Input In [3], in <module>
----> 1 set_iter_method("primini")

File <string>:1, in set_iter_method(meth)

File ~/sherpa/sherpa-main/sherpa/ui/utils.py:1882, in Session.set_iter_method(self, meth)
   1880     self._current_itermethod = self._itermethods[meth]
   1881 else:
-> 1882     raise TypeError(f'{meth} is not an iterative fitting method')

TypeError: primini is not an iterative fitting method
```

DougBurke:
@anetasie - this is my suggestion for the documentation - I don't feel like it's worth adding more text to explain why we removed the method - that is better done in the changelog if we feel like doing so.

```
% git diff
diff --git a/sherpa/ui/utils.py b/sherpa/ui/utils.py
index 4c09db21..b7c5465d 100644
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020, 2021
+#  Copyright (C) 2010, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022
 #
 #
@@ -1677,8 +1677,6 @@ class Session(NoNewAttributesAfterInit):
         self._check_method_opt(optname)
         self._current_method.config[optname] = val

-    # Iterative Fitting Methods for CIAO 4.3 testing
-
     def get_iter_method_name(self):
         """Return the name of the iterative fitting scheme.

@@ -1791,6 +1789,9 @@ class Session(NoNewAttributesAfterInit):
         Control whether an iterative scheme should be applied to
         the fit.

+        .. versionchanged:: 4.14.1
+           The "primini" scheme has been removed from Sherpa.
+
         Parameters
         ----------
         meth : { 'none', 'sigmarej' }
@@ -1814,7 +1815,7 @@ class Session(NoNewAttributesAfterInit):

         Notes
         -----
-        The parameters of each scheme are described in
+        The parameters of the schemes are described in
         `set_iter_method_opt`.

         This is a chi-square statistic where the variance is computed
```

dtnguyen2:
Not sure why I'm seeing intermittent test failure:
```Python
sherpa/tests/test_instrument.py .......FFFF..X..                         [ 82%]
```
cause it works fine when I re-test it:
```Python
(testing-PR) [dtn@devel12 sherpa]$ pytest sherpa/tests/test_instrument.py
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /export/sherpa, configfile: pytest.ini
collected 16 items                                                             

sherpa/tests/test_instrument.py .............X..                         [100%]

======================== 15 passed, 1 xpassed in 0.08s =========================
```

anetasie:
The line in the documentation looks good.

DougBurke (approved):
LGTM. The test failures should be fixed by #1428 which has been merged. Do we want this rebased first or can it be merged?

PR #1425, opened 2022-02-03
Allow users to freeze models

### LINKED ISSUE #1080: can we freeze a whole model?
I am lazy - in this case I should say `gal.nh.freeze()` but it's too many characters:

```
In [3]: gal.freeze()
<ipython-input-3-bcdbbe990bd4> in <module>
----> 1 gal.freeze()

~/sherpa/sherpa-docs/sherpa/models/model.py in __getattr__(self, name)
    201             return parameter
    202
--> 203         NoNewAttributesAfterInit.__getattribute__(self, name)
    204
    205     def __setattr__(self, name, val):

AttributeError: 'XSphabs' object has no attribute 'freeze'

```

We could add a freeze method to the model class that goes and freeze all the parameters of the model.

This would help for models with multiple thawed parameters too.

Need to look at how the ui `freeze` function works.

DougBurke:
# Summary

Models can now be frozen or thawed, which just calls the requested on all the parameters of the model (except for "alwaysfrozen" parameters which are skipped). This is only relevant for users who are accessing the objects directly since the ui versions of freeze and thaw already implemented this behavior. Fixes #1080.

The documentation for the freeze and thaw routines in the ui code has been fixed to correctly describe what happens when used with no arguments (nothing is changed).

# Details

In adding the following functionality I realized that the `sherpa[.astro].ui.freeze/thaw` functions were documented incorrectly: they suggested that they did something when called with no arguments when they actually do nothing. I added tests to show that they do nothing and then fixed the documentation. We **could** change the code here, but I strongly advise against it as I think it is likely to cause more problems than to help users.

We were already able to say

```
>>> from sherpa import ui
>>> ui.set_source(ui.polynom1d.m1 + ui.gauss1d.m2)
>>> ui.thaw(m1)
```

but we couldn't say the "equivalent" with objects. Well, now you can:

```
>>> from sherpa.models.basic import Gauss1D
>>> mdl = Gauss1D()
>>> mdl.freeze()
```

This is a minor improvement, but it does actually make implementation of the `ui.thaw(mdl)` version a teensy-bit nicer, as well as moving some code to be "more Pythonic" , in that it uses duck typing/asking for forgiveness rather than instance checks and asking for permission (although this is only a small change so it's not worth spending too much time on it).

# Corner case

Some models have one - or more - parameters that are marked as "always frozen". For example the `ref` parameter of `sherpa.models.basic.LogParabola` is one such. If you try to thaw the individual parameter you get an error (`ParameterErr`) , but when thawing the whole model we just skip this parameter. This is how the `ui.thaw` routine worked, so we replicate this for the `thaw` method added to the model class. 

So, in CIAO 4.14 we have the following, where `thaw(mdl.ref)` fails but `thaw(mdl)` passes:

```
sherpa-4.14.0> logparabola.mdl
-------------> logparabola.mdl()
<LogParabola model instance 'logparabola.mdl'>

sherpa-4.14.0> print(mdl)
logparabola.mdl
   mdl.ref      frozen            1 -3.40282e+38  3.40282e+38
   mdl.c1       thawed            1 -3.40282e+38  3.40282e+38
   mdl.c2       thawed            1 -3.40282e+38  3.40282e+38
   mdl.ampl     thawed            1            0  3.40282e+38

sherpa-4.14.0> thaw(mdl.ref)
ParameterErr: parameter mdl.ref is always frozen and cannot be thawed

sherpa-4.14.0> thaw(mdl)

sherpa-4.14.0> print(mdl)
logparabola.mdl
   mdl.ref      frozen            1 -3.40282e+38  3.40282e+38
   mdl.c1       thawed            1 -3.40282e+38  3.40282e+38
   mdl.c2       thawed            1 -3.40282e+38  3.40282e+38
   mdl.ampl     thawed            1            0  3.40282e+38

sherpa-4.14.0> mdl.thaw()
AttributeError: 'LogParabola' object has no attribute 'thaw'
```

With this PR we can no perform this last line:

```
>>> from sherpa.models.basic import LogParabola
>>> mdl = LogParabola()
>>> mdl.ref.thaw()
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/parameter.py", line 663, in thaw
    self.frozen = False
  File "/home/dburke/sherpa/sherpa-main/sherpa/utils/__init__.py", line 171, in __setattr__
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/parameter.py", line 455, in _set_frozen
    raise ParameterErr('alwaysfrozen', self.fullname)
sherpa.utils.err.ParameterErr: parameter logparabola.ref is always frozen and cannot be thawed
>>> print(mdl)
logparabola
   logparabola.ref frozen            1 -3.40282e+38  3.40282e+38
   logparabola.c1 thawed            1 -3.40282e+38  3.40282e+38
   logparabola.c2 thawed            1 -3.40282e+38  3.40282e+38
   logparabola.ampl thawed            1            0  3.40282e+38
>>> mdl.freeze()
>>> print(mdl)
logparabola
   logparabola.ref frozen            1 -3.40282e+38  3.40282e+38
   logparabola.c1 frozen            1 -3.40282e+38  3.40282e+38
   logparabola.c2 frozen            1 -3.40282e+38  3.40282e+38
   logparabola.ampl frozen            1            0  3.40282e+38
>>> mdl.thaw()
>>> print(mdl)
logparabola
   logparabola.ref frozen            1 -3.40282e+38  3.40282e+38
   logparabola.c1 thawed            1 -3.40282e+38  3.40282e+38
   logparabola.c2 thawed            1 -3.40282e+38  3.40282e+38
   logparabola.ampl thawed            1            0  3.40282e+38
```

### DIFF
--- a/sherpa/models/tests/test_model.py
+++ b/sherpa/models/tests/test_model.py
@@ -1226,7 +1226,6 @@ def test_cache_clear_multiple(caplog):
     assert c._cache_ctr['misses'] == 0
 
 
-@pytest.mark.xfail
 def test_model_freeze():
     """Can we freeze all the parameters in a model?"""
 
@@ -1243,7 +1242,6 @@ def test_model_freeze():
     assert mdl.thawedpars == []
 
 
-@pytest.mark.xfail
 def test_model_thaw():
     """Can we freeze all the parameters in a model?"""
 
@@ -1259,7 +1257,6 @@ def test_model_thaw():
     assert mdl.thawedpars == pytest.approx(expected)
 
 
-@pytest.mark.xfail
 def test_model_freeze_already_frozen():
     """Check it's a no-op rather than an error"""
 
@@ -1271,7 +1268,6 @@ def test_model_freeze_already_frozen():
     assert mdl.thawedpars == []
 
 
-@pytest.mark.xfail
 def test_model_thaw_already_thawed():
     """Check it's a no-op rather than an error"""
 
@@ -1282,7 +1278,6 @@ def test_model_thaw_already_thawed():
     assert mdl.thawedpars == [1.0]
 
 
-@pytest.mark.xfail
 def test_model_freeze_alwaysfrozen():
     """Check it's a no-op rather than an error for an alwaysfrozen parameter"""
 
@@ -1294,7 +1289,6 @@ def test_model_freeze_alwaysfrozen():
     assert mdl.thawedpars == []
 
 
-@pytest.mark.xfail
 def test_model_thaw_alwaysfrozen():
     """Check we skip an alwaysfrozen parameter"""
 
--- a/sherpa/models/tests/test_model.py
+++ b/sherpa/models/tests/test_model.py
@@ -1243,7 +1243,7 @@ def test_model_freeze():
 
 
 def test_model_thaw():
-    """Can we freeze all the parameters in a model?"""
+    """Can we thaw all the parameters in a model?"""
 
     mdl = Polynom1D()
     expected = [2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 20.0]
--- a/sherpa/models/tests/test_model.py
+++ b/sherpa/models/tests/test_model.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2016, 2017, 2020, 2021
+#  Copyright (C) 2007, 2016, 2017, 2020, 2021, 2022
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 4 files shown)


anetasie (approved):

sherpa/models/tests/test_model.py
```diff
@@ -1224,3 +1224,77 @@ def test_cache_clear_multiple(caplog):
...
+    mdl.c7.thaw()
+    mdl.offset.thaw()
+    assert mdl.thawedpars == pytest.approx([1.0, 30, 87, -2])
+
+    mdl.freeze()
+    assert mdl.thawedpars == []
+
+
+def test_model_thaw():
+    """Can we freeze all the parameters in a model?"""
```
wmclaugh, line 1246:
```suggestion
    """Can we thaw all the parameters in a model?"""
```

DougBurke:
I've rebased this to address a conflict.

PR #1434, opened 2022-02-07
Fix aarch64 test failures

DougBurke:
# Summary

Allow the tests to pass on aarch64 (numerical precision checks).

# Details

These values are slightly different on aarch64 and it makes sense to use pytest.approx rather than equality here.

I added the 'CI arch' label so that the aarch run would be included (normally it is only run at the week-end) and the tests pass.


DougBurke:
I wonder why the aarch64 CI run was run three times with this commit - see https://github.com/sherpa/sherpa/actions/workflows/ci-pip-arch.yml and

- https://github.com/sherpa/sherpa/actions/runs/1808478621
- https://github.com/sherpa/sherpa/actions/runs/1808478630
- https://github.com/sherpa/sherpa/actions/runs/1808478643

hamogu (approved):

Issue #1436, opened 2022-02-09
Do we want to add validation to Data classes?


DougBurke:
We've worked through most of the existing issues already, but we can still end up in a situation where a user can do something that causes downstream problems. My assumption is that we can mitigate the vast-majority of them by adding some form of validation to the `Data` hierarchy, but that is yet to be proven by design or code (they're being worked on). I wanted to have an issue to point to for that work.

# Example

Here's an example of the problem, with CIAO 4.14 but any version would do. The problem here is that the independent and dependent axes do not match, so a down-stream consumer of the data - in this case the `get_dep` call - ends up having to catch this case and report an error (or maybe you get a NumPy or Python error raised). Ideally the data could never be created in the first place (aka "Make Invalid States Unrepresentable").

## using the UI interface

```
sherpa-4.14.0> load_arrays(1, [1, 2, 3], [1, 2])

sherpa-4.14.0> ignore(3, None)

sherpa-4.14.0> get_dep()
array([1, 2])

sherpa-4.14.0> get_dep(filter=True)
DataErr: size mismatch between mask and data array
```

## directly using the `Data` interface

```
sherpa-4.14.0> d = Data1D('x', [1, 2, 3], [1, 2])
/home/dburke/anaconda/envs/ciao414/lib/python3.8/site-packages/sherpa/data.py:88: UserWarning: Converting array [1, 2] to numpy array.
  warnings.warn("Converting array {} to numpy array.".format(array))

sherpa-4.14.0> d.get_dep(filter=True)
array([1, 2])

sherpa-4.14.0> d.ignore(3, None)

sherpa-4.14.0> d.mask
array([ True,  True, False])

sherpa-4.14.0> d.get_dep(filter=True)
DataErr: size mismatch between mask and data array
```

# What validations could be done

- convert values to numpy arrays (we do this for the independent and dependent axes but not for other fields, such as `staterror`
- is everything a 1D array
- do we have the correct number of independent axes
- do the various fields (e.g. independent, dependent, errors, ...) have the same size
- do we want to convert data types - normally to float but the PHA case has a number of integer cases

# What would it mean

[...]

PR #1440, opened 2022-02-13
Minor clean up of the data routines

DougBurke:
# Summary

Address several minor code-style changes in the data handling (code and tests).

# Details

This was found working on #1436 but is separate from that work (i.e. can be reviewed separately).

There are two code changes

1. one in handling the `calc_data_sum`/`calc_data_sum2d` (and model variants) routines in the UI layer
2. one in how the `set_dep` method is written for the `DataPHA` class

The first change has two parts

a) removal of unnecessary filters - namely the `notice`/`notice2d` call after summing the data; this is not needed since we explicitly reset the filter in the `fnially` block and so over-ride this call
b) for the 2D case the code would store the coord setting and then reset it, but this is pointless since we don't change the coord setting

Both of these are just changes to stop doing things we don't need to do annd to simplify things for later work (e.g. #1436).

The second change is just a style change, using the more Pythonic `self.channel = val` rather than `setattr(self, "channel") = val`.

For the tests the second commit fixes a logical error in the test: the test just needs a PHA dataset with a different number of channels to the background so it takes the source dataset and removes the last bin from the data fields. We don't actually care what the values are, so I hadn't noticed I'd accidentally set the counts array to match the channel array. This fixes this and simplifies the code a bit to be more explicit about what values it needs to change (ie which are set and an array). I don't know why I wrote the original code to be "generic" - ie to handle any PHA data set it was sent - when it doesn't need to be in this case.

The last commit - `Tests: add explicit tests for calc_data_sum/calc_data_sum2d` - adds some explicit checks of the code in `sherpa.astro.utils`, primarily to ensure we have a case with a quality array, as I noted that this was not being covered by our existing tests (and is related to the code clean up in this PR). I've tried to be systematic, so there's a bunch of unit-like tests checking whether the `calc_data_sum` or `calc_data_sum2d` routines behave as we'd expect (although there are some cases, in particular with the quality array, when it's not obvious what we want the code to do, so in this case the tests are more like regression tests, checking that we will at least know if the behavior changes).

### DIFF
--- a/sherpa/astro/utils/__init__.py
+++ b/sherpa/astro/utils/__init__.py
@@ -123,7 +123,7 @@ def compile_energy_grid(arglist):
 
 
 def bounds_check(lo, hi):
-    """Ensure that the imits of a filter make sense.
+    """Ensure that the limits of a filter make sense.
 
     Note that if only one of them is given we always return
     that value first.
@@ -397,7 +397,7 @@ def _counts(data, lo, hi, func, *args):
 
     It is then time to restore the original filter, which is done in a
     finally block so that any errors calling func or the notice calls
-    does not change the data object.
+    do not change the data object.
 
     The use of the name counts is a bit of a mis-nomer as this could
     be used on non-PHA data, but the user only sees this via
--- a/sherpa/astro/utils/__init__.py
+++ b/sherpa/astro/utils/__init__.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2008, 2016, 2020, 2021  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2008, 2016, 2020, 2021, 2022
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -375,8 +376,8 @@ def _counts(data, lo, hi, func, *args):
         data.notice()  # clear filter
         data.notice(lo, hi)
         counts = func(*args).sum()
-        data.notice()
     finally:
+        # restore the old filter
         data.mask = old_mask
         if old_quality_filter is not None:
             data.quality_filter = old_quality_filter
@@ -386,20 +387,14 @@ def _counts(data, lo, hi, func, *args):
 
 def _counts2d(data, reg, func, *args):
     old_mask = data.mask
-    coord = getattr(data, 'coord', None)
     old_region = getattr(data, '_region', None)
     try:
         data.notice2d()  # save and clear filter
-
         data.notice2d(reg)
         counts = func(*args).sum()
-        data.notice2d()
     finally:
+        # restore the old filter
         data.mask = old_mask
-
-        if coord is not None:
-            data.set_coord(coord)
-
         if old_region is not None:
             data._region = old_region
 
--- a/sherpa/astro/utils/__init__.py
+++ b/sherpa/astro/utils/__init__.py
@@ -123,6 +123,16 @@ def compile_energy_grid(arglist):
 
 
 def bounds_check(lo, hi):
+    """Ensure that the imits of a filter make sense.
+
+    Note that if only one of them is given we always return
+    that value first.
+    """
+
+    # TODO: this error message is not ideal, as it assumes the
+    # units are energy when the values could be wavelengths,
+    # channels, or for a non PHA dataset. See issue #1443
+    #
     if lo is not None and hi is not None and lo > hi:
         raise IOErr('boundscheck', lo, hi)
 
@@ -369,6 +379,31 @@ def _flux(data, lo, hi, src, eflux=False, srcflux=False):
 
 
 def _counts(data, lo, hi, func, *args):
+    """Sum up the data for the range lo to hi.
+
+    The routine starts by saving the current filter, which is a
+    combination of the mask field and - if set, which it is only for
+    PHA data with ignore_bad called - the quality_filter field.
+
+    The existing noticed range of the data set is ignored,
+    and instead the range lo-hi is used (the meaning is slightly
+    different to the notice method when one of the two fields
+    is None but the other is set; see bounds_check).
+
+    At this point the func method is called, with the remaining
+    arguments. It is assumed that this is a method of the
+    data object, but there is no requirement. The return value
+    is summed up and this value will be returned by the routine.
+
+    It is then time to restore the original filter, which is done in a
+    finally block so that any errors calling func or the notice calls
+    does not change the data object.
+
+    The use of the name counts is a bit of a mis-nomer as this could
+    be used on non-PHA data, but the user only sees this via
+    calc_data_sum/calc_model_sum.
+
+    """
     lo, hi = bounds_check(lo, hi)
     old_mask = data.mask
     old_quality_filter = getattr(data, 'quality_filter', None)
@@ -386,6 +421,20 @@ def _counts(data, lo, hi, func, *args):
 
 
 def _counts2d(data, reg, func, *args):
+    """Sum up the data for the given region.
+
+    This is the same as _counts but for a data object that supports
+    the notice2d method (so currently DataIMG and DataIMGInt). It has
+    the same structure as _counts except that the fields used to save
+    and restore the state of the data object match the DataIMG
+    interface (i.e. mask and optional _region rather than mask and an
+    optional quality_filter)
+
+    The use of the name counts is a bit of a mis-nomer as this could
+    be used on non-PHA data, but the user only sees this via
+    calc_data_sum2d/calc_model_sum2d.
+
+    """
     old_mask = data.mask
     old_region = getattr(data, '_region', None)
     try:
--- a/sherpa/astro/utils/tests/test_astro_utils_unit.py
+++ b/sherpa/astro/utils/tests/test_astro_utils_unit.py
@@ -249,7 +249,7 @@ def make_data(data_class):
     # We want to provide PHA tests that check out the grouping and
     # quality handling (but it is not worth trying all different
     # variants), so we have "grp" for grouping and no quality [*], and
-    # "qual" for groupbing and quality.
+    # "qual" for grouping and quality.
     #
     # [*] by which I mean we have not called ignore_bad, not that
     # there is no quality array.
--- a/sherpa/astro/utils/tests/test_astro_utils_unit.py
+++ b/sherpa/astro/utils/tests/test_astro_utils_unit.py
@@ -229,8 +229,9 @@ def test_do_group_single_group(func, expected):
 def make_data(data_class):
     """Create a test data object of the given class.
 
-    Using a string means it is easier to support the various
-    PHA "types" - eg basic, grouping, grouping+quality
+    Using a string means it is easier to support the various PHA
+    "types" - eg basic, grouping, grouping+quality.
+
     """
 
     x0 = np.asarray([1, 3, 7, 12])
@@ -245,7 +246,14 @@ def make_data(data_class):
     if data_class == "pha":
         return DataPHA('pha', chans, y)
 
-    # These are a special case to check out grpuping and quality handling.
+    # We want to provide PHA tests that check out the grouping and
+    # quality handling (but it is not worth trying all different
+    # variants), so we have "grp" for grouping and no quality [*], and
+    # "qual" for groupbing and quality.
+    #
+    # [*] by which I mean we have not called ignore_bad, not that
+    # there is no quality array.
+    #
     grp = np.asarray([1, -1, 1, 1])
     qual = np.asarray([0, 0, 2, 0])
     pha = DataPHA('pha', chans, y, grouping=grp, quality=qual)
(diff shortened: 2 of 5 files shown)


wmclaugh:
Can the title of the PR please be changed to something a little more indicative of the scope of the changes.

sherpa/astro/utils/__init__.py
```diff
@@ -375,8 +376,8 @@ def _counts(data, lo, hi, func, *args):
         data.notice()  # clear filter
         data.notice(lo, hi)
         counts = func(*args).sum()
-        data.notice()
     finally:
+        # restore the old filter
```
hamogu, line 380: Since you just spend a lot (I guess) of time to understand what exactly is happening, I suggest a few comments, e.g. why do we need to reset the quality filter only it it's not None? If there is no good reason,  suggest to always reset it for simplicity (it uses `getattr(..., ... None)` above so maybe was can do that?

What is this function doing anyway? It starts with an underscore, so we don't need a full doctring, but a 1-line description for te future would be nice.

sherpa/astro/utils/__init__.py
```diff
@@ -386,20 +387,14 @@ def _counts(data, lo, hi, func, *args):
 
 def _counts2d(data, reg, func, *args):
     old_mask = data.mask
-    coord = getattr(data, 'coord', None)
```
hamogu, line 389: I assume `coord` is here for the same reason that 
```
        if old_quality_filter is not None:
            data.quality_filter = old_quality_filter
```
is there for the 1d case. You never know if these notice calls mess something up. I've rarely if ever looked at the 2d case, so I don't know if that's possible.

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,269 @@ def test_do_group_single_group(func, expected):
...
+        return Data1D('x1', x0, y)
+
+    if data_class == "1dint":
+        return Data1DInt('xint1', x0, np.asarray([3, 5, 8, 15]), y)
+
+    chans = np.arange(1, 5)
+    if data_class == "pha":
+        return DataPHA('pha', chans, y)
+
+    # These are a special case to check out grpuping and quality handling.
```
hamogu, line 248:
```suggestion
    # This is a special case to check out grouping and quality handling.
```

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,269 @@ def test_do_group_single_group(func, expected):
...
+    #
+    assert str(err.value) == "the energy range is not consistent, 10 !< 2"
+
+
+@pytest.mark.parametrize("data_class", ["1d", "1dint", "pha", "grp", "qual"])
+def test_calc_data_sum_no_range(data_class):
+    """Call calc_data_sum(data)"""
+
+    data = make_data(data_class)
+    assert utils.calc_data_sum(data) == 14
```
hamogu, line 295: Mmm. I would have expected the answer to be different if I `ignore_bad` first, that is in the "qual" case. Do we really want to sum up known bad data? In this example, it's a reasonable number, but I practice, I would expect that bad data just as well could be inf. If that's current behavior and this is a just a regression test, I suggest adding at least a note or maybe split the qual in to a separate test with ` assert utils.calc_data_sum(data) == 10` and xfail?

Or maybe I'm just not understanding what `calc_data_sum` is supposed to be doig.

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,269 @@ def test_do_group_single_group(func, expected):
...
+    assert utils.calc_data_sum(data, **{name: limit}) == expected
+    assert data.get_dep(filter=True) == pytest.approx(orig)
+
+
+@pytest.mark.parametrize("name", ["lo", "hi"])
+@pytest.mark.parametrize("limit,expected", [(0, 0), (1, 0), (2, 2), (3, 0),
+                                            (4, 3), (6, 0), (7, 0), (11, 0), (12, 0),
+                                            (13, 5), (15, 0)])
+def test_calc_data_sum_only_one_limit_1dint(name, limit, expected):
+    """Call calc_data_sum(data, lo or hi): Data1DInt"""
```
hamogu, line 321:
```suggestion
    """Call calc_data_sum(data, lo or hi): Data1DInt and check that is does not mess up the internal state"""
```

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,269 @@ def test_do_group_single_group(func, expected):
...
+@pytest.mark.parametrize("data_class", ["grp", "qual"])
+@pytest.mark.parametrize("name", ["lo", "hi"])
+@pytest.mark.parametrize("limit,expected", [(0, 0), (1, 5), (2, 5), (3, 4),
+                                            (4, 5), (5, 0), (6, 0), (15, 0)])
+def test_calc_data_sum_only_one_limit_pha_grouped(name, limit, expected, data_class):
+    """Call calc_data_sum(data, lo or hi): DataPHA (group/quality)
+
+    It looks like the quality array is ignored (which is probably
+    not a great idea).
+    """
```
hamogu, line 350: Agreed! Suggest same comment above.

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,269 @@ def test_do_group_single_group(func, expected):
...
+                                             ((3, 12), 9),
+                                             ((1, 13), 14),
+                                             ((20, 22), 0)])
+def test_calc_data_sum_filtered_pha_grouped(frange, expected, data_class):
+    """Filter out everything and then check with both lo/hi set: DataPHA (grouped/quality)"""
+
+    data = make_data(data_class)
+    data.ignore()
+    assert not data.mask
+    assert utils.calc_data_sum(data, *frange) == expected
```
hamogu, line 437: The longer I look at this, the less I understand what `calc_data_sum` is good for.

sherpa/stats/tests/test_stats_unit.py
```diff
@@ -497,17 +498,14 @@ def test_stats_calc_stat_wstat_diffbins():
...
-    data.counts = data.channel[:-1]
-    for attr in ['staterror', 'syserror', 'grouping', 'quality',
-                 'backscal']:
-        val = getattr(data, attr)
-        if val is not None:
-            try:
-                setattr(data, attr, val[:-1])
-            except TypeError:
-                # assume a scalar, so leave be
-                pass
```
hamogu, line 510: Yes, much easier to read now.

sherpa/astro/utils/__init__.py
DougBurke: Yes, the 1D case has to worry about the quality array and the 2d case might want to worry about the coord setting (except that it doesn't actually, which is why it's being removed).

sherpa/astro/utils/__init__.py
DougBurke: This can be used for a data object that has no quality filter, so we only want to reset it if the field exists and is not None. We have used `getattr` with a default of `None` so this acts as a way to indicate whether we should bother resetting the field.

There are actually two subtle issues here:

- should we do a "dimensionality" check here, so that using a DataIMG object would lead to an error rather than the current behavior of "do something but if you add a filter it is likely to not make sense", but here I just want to ensure we test the current behavior
- should we change the quality array if it is set; I don't think we should but all I want to do is check/test the current behavior and leave the quality-handling to later PRs (as we know that there are issues with out quality handling)

sherpa/astro/utils/tests/test_astro_utils_unit.py
DougBurke: So, this is the second point of my response above at https://github.com/sherpa/sherpa/pull/1440#discussion_r806262758 

This is not intended to be a PR that fixes all the things. The tests are written to test the current behavior. I really want to see them fail when we get around to fixing the quality handling, but do not want that to be "a subtle" change of x to X in the test output, but a "nope, this test has really failed because the code is now doing something different", as I am getting less hopeful that I'll ever get to fix the quality code.

sherpa/astro/utils/tests/test_astro_utils_unit.py
DougBurke: I don't understand the suggested change.

hamogu: Not the best wording. I meant to explain where there is the line `assert data.get_dep(filter=True) == pytest.approx(orig)`. It's not clear from the test that there is even the possibility that `calc_data_sum` messes up the dependent array, when then, when you look into the `calc_data_sum` code, there is actually a lot going on in the try-finally block. So, this check makes sense - it's just no obvious that it does. 

But I can live without that comment if we can't come up with better wording, not a point to break sweat about.

sherpa/astro/utils/tests/test_astro_utils_unit.py
DougBurke: It actually comes down to the work I did for the 4.14 release to try and "harmonise" the range filtering of ignore/notice. The issue is whether a filter of  `lo,hi` is `[lo, hi]` (inclusive for both the `lo` and `hi` values) - which we have for `Data1D`, or `[lo, hi)` (so exclusive for the `hi` limit) as have do for `Data1DInt`. `DataPHA` is a special case because technically we have `[lo,hi]` for channel and `[lo,hi)` for energy or wavelength filtering [*], but these particular tests only really test the `channel` case (since we rely on the existing `notice` tests to properly check all the many variants of this).

[*] if memory serves, we did actually discuss this in your reviews of the filter code, but it is easy to get confused.

To go off on a bit of a rant - I think the original `calc_data_sum` code assumes we are using energies and hasn't really thought through the cases of using wavelength or channel, or with a non-PHA data set. I am not willing to put a lot of effort into improving this at this time.

sherpa/stats/tests/test_stats_unit.py
DougBurke: I am struggling to come up with a reason for why I wrote the code like this in the first place...

sherpa/astro/utils/tests/test_astro_utils_unit.py
hamogu: That's fine - I'm not asking to improve the meaning of the method. Maybe just your magic "this is a regression test, there may be issues with calculating the sum in channel or wavelength"?

sherpa/astro/utils/__init__.py
DougBurke: For reference, I've created #1444 and #1445 to track these

DougBurke:
I've added documentation which hopefully resolves the review comments from @hamogu but they do need to be read by someone other than me!

hamogu (approved):
That's longer than it needed to be, but more is more in the this case!

sherpa/astro/utils/__init__.py
```diff
@@ -122,6 +123,16 @@ def compile_energy_grid(arglist):
 
 
 def bounds_check(lo, hi):
+    """Ensure that the imits of a filter make sense.
```
wmclaugh, line 126:
```suggestion
    """Ensure that the limits of a filter make sense.
```

sherpa/astro/utils/__init__.py
```diff
@@ -368,15 +379,40 @@ def _flux(data, lo, hi, src, eflux=False, srcflux=False):
...
+    is None but the other is set; see bounds_check).
+
+    At this point the func method is called, with the remaining
+    arguments. It is assumed that this is a method of the
+    data object, but there is no requirement. The return value
+    is summed up and this value will be returned by the routine.
+
+    It is then time to restore the original filter, which is done in a
+    finally block so that any errors calling func or the notice calls
+    does not change the data object.
```
wmclaugh, line 400:
```suggestion
    do not change the data object.
```

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -220,3 +224,362 @@ def test_do_group_single_group(func, expected):
...
+        return Data1DInt('xint1', x0, np.asarray([3, 5, 8, 15]), y)
+
+    chans = np.arange(1, 5)
+    if data_class == "pha":
+        return DataPHA('pha', chans, y)
+
+    # We want to provide PHA tests that check out the grouping and
+    # quality handling (but it is not worth trying all different
+    # variants), so we have "grp" for grouping and no quality [*], and
+    # "qual" for groupbing and quality.
```
wmclaugh, line 252:
```suggestion
    # "qual" for grouping and quality.
```

DougBurke: I had missed this when @hamogu pointed it out in https://github.com/sherpa/sherpa/pull/1440#discussion_r806213780 - grrr

hamogu: I'm glad that the only thing that's still left to worry about is a comment in a test.

Issue #1174, opened 2021-06-01
Unify `get_indep`, `_get_indep`, and `_get_ebins` for DataPHA


hamogu:
`DataPHA` objects have three similar, but distinct functions to access the independent axis in addition to direct access to the  data fields such as `data.channels`, etc. 
The differences are somewhat surprising and the design intent is not clear (see #1173 for discussion). 

- `get_indep` returns the channels, possibly filtered (but see #1041)
- `_get_indep` returns bin edges in keV or Ang, and has an argument `filter`, but that might not do anything.
- `_get_ebins` returns bin edges in channel or energy (but not wavelength), and has parameters `group` and `response_id`.

That looks like we could do all that with just one function, with arguments

- filter (apply filtering / grouping?)
- output units (channel, energy, wavelength, automatic [i.e. energy if arf, rmf available, otherwise channel]).

Of course, that means touching a large number of places with filtering, plotting etc., so we won't do it until the current (summer 2021) work on grouping and filtering is done.

Since two of the methods start with an `_` they are considered private and we can change them at any time.

hamogu:
I forgot that there is also `get_y`, which again is almost, but not quite the same.

DougBurke:
So, `_get_indep` actually is different, in that it returns the energy grid of the response, and so is likely to have a different size to the number of channels.

`get_x` and `get_indep` are "forced" on us because we use Python and objects and class hierarchies.

`_get_ebins` is not ideal™ but we need something like this (but obviously not a "hidden" function).

I am working on this for #2189 (although how long this stays valid is to be determined).

PR #1441, opened 2022-02-13
Clean up the data tests

DougBurke:
# Summary

Clean up of the data tests which were added as part of the data-class rework in #614.

# Details

Apologies that the commits are discussed oiut-of order (ie second one is desvribed first).

The data hierarchy was re-worked in #614 and as part of this @olaurino added a number of basic tests. Some of these tests now test "old" behavior, and are marked as xfail (since they now fail). For instance, we no longer have a `DataND` class, and so trying to create one is now going to fail, but we still have a test for this. The second commit cleans up the test suite to either:

- change the test from being marked `xfail` to one that passes (so, for example. instead of checking whether we can call a particular method we now check to make sure that method is not defined)
- or remove the test (e.g. checking that you can create a `DataND` object when we now no-longer define this class)

It has been a long time now since #614 so it makes sense to clean up the test suite. Note however that this does not remove all the xfail tests in `sherpa/tests/test_data.py`.
 
The first commit enhances the "can we create a data object with no error values" tests. Previously the test was just for the base `Data` class, which is actually not that useful, as we don't expect users to use this class. It makes more sense to test "user" classes, so we now include tests with the data classes defined in `sherpa.data`. One change is that we no-longer test the base `Data` case, but I believe that the old version of the test - which only used the `Data` class - actually had a logical error in how the `Data` object was being created (I believe the `Session` class, which powers the UI interface, does not support the base `Data` class, just the explicit versions, such as `Data1D` and `Data2DInt`, but we don't explicitly say  this anywhere).

The third commit - `Tests: tweak a test of set_dep handling` - changes a test of the `set_dep` routine: first by adding a test with `Data1DInt` to match the existing `Data1D` test; and second by calling `set_dep` with the correct number of elements (in this case 10) rather than the 2 that was in use. It is not clear what a 2-element array means here (since technically the filter is no-longer valid), so using a "sensible" value makes sense. I plan to address what exactly a `set_dep` call does with an invalid length argument as part of #1436 but we need changes like this in first.

This was found working on #1436 but is separate from that work.

### DIFF
--- a/sherpa/tests/test_data.py
+++ b/sherpa/tests/test_data.py
@@ -148,22 +148,23 @@ def test_base_data_instantiation():
         BaseData()
 
 
-@pytest.mark.xfail(reason="DataND did not serve any purpose and had a misleading name")
-def test_base_datand_instantiation():
-    DataND()
-
-
-@pytest.mark.xfail(reason="methods did not belong and were removed")
-@pytest.mark.parametrize("data", (Data, Data2D, Data2DInt, Data1DInt), indirect=True)
+@pytest.mark.parametrize("data", (Data, Data2D, Data2DInt), indirect=True)
 def test_data_get_x(data):
-    with pytest.raises(NameError):
+    with pytest.raises(AttributeError):
         data.get_x()
 
 
-@pytest.mark.xfail(reason="methods did not belong and were removed")
+@pytest.mark.xfail
+@pytest.mark.parametrize("data", (Data1DInt, ), indirect=True)
+def test_data_get_x_special(data):
+    # XFAIL: These classes still provide get_x
+    with pytest.raises(AttributeError):
+        data.get_x()
+
+
 @pytest.mark.parametrize("data", DATA_1D_CLASSES, indirect=True)
 def test_data_get_x0(data):
-    with pytest.raises(NameError):
+    with pytest.raises(AttributeError):
         data.get_x0()
 
 
@@ -189,7 +190,7 @@ def test_load_arrays(data_for_load_arrays):
     numpy.testing.assert_array_equal(new_data.get_dep(), data.get_dep())
 
 
-# DATA-NOTE: In the current Sherpa DataND cannot be correctly loaded using load_arrays
+# DATA-NOTE: In the current Sherpa Data cannot be correctly loaded using load_arrays
 @pytest.mark.xfail
 @pytest.mark.parametrize("data_for_load_arrays", (Data, ), indirect=True)
 def test_load_arrays_data(data_for_load_arrays):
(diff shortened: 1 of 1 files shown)


sherpa/tests/test_data.py
```diff
@@ -141,22 +148,23 @@ def test_base_data_instantiation():
...
-@pytest.mark.xfail(reason="methods did not belong and were removed")
-@pytest.mark.parametrize("data", (Data, Data2D, Data2DInt, Data1DInt), indirect=True)
-def test_data_get_x(data):
-    with pytest.raises(NameError):
+@pytest.mark.xfail
+@pytest.mark.parametrize("data", (Data1DInt, ), indirect=True)
+def test_data_get_x_special(data):
+    # XFAIL: These classes still provide get_x
+    with pytest.raises(AttributeError):
         data.get_x()
```
hamogu, line 162: Note (not for this PR): I do see an argument for providing a `get_x` for all 1d classes. For the binned classes (Data1DInt) it would provide a [2,n] array or a tuple or something, but it is confusing that there are so many way to access the x data with partially overlapping names and functionality (#1174)

sherpa/tests/test_data.py
```diff
@@ -557,34 +572,24 @@ def test_data_1d_to_component_plot(data):
     numpy.testing.assert_array_equal(actual[5], expected[5])
 
 
-@pytest.mark.xfail(reason="methods did not belong and were removed")
-@pytest.mark.parametrize("data", (Data, ), indirect=True)
+@pytest.mark.parametrize("data", (Data, Data1D, Data1DInt), indirect=True)
 def test_data_to_contour(data):
-    with pytest.raises(DataErr):
+    with pytest.raises(AttributeError):
         data.to_contour()
```
hamogu, line 578: Fine. But we also don't defined the methods `to_plot_contours`, `to_image`, or `to_chocolate`. Do we really need tests to prove that at method does not exist? I mean, no harm (so I'll approve) and there is a historic reason for this, but is this really needed?

hamogu (approved):

sherpa/tests/test_data.py
DougBurke: So, this was removed in #614 and I don't think we want to go back on that work without a lot of careful thinking.

sherpa/tests/test_data.py
DougBurke: Well, it ensures that we don't get back into the situation we had before #614 where classes provided methods they could not do anything sensible with (there's likely to always be some problem case where it gets complicated), so I see some value in these checks.

hamogu: OK, understood.

olaurino: I might be stating the obvious, but just for clarity. In the context of #614 I marked tests as `xfail` either to document where I was breaking with the previous API or because what I thought was the expected behavior of a method wasn't actually happening. Weird tests like the one above belong to the first class (I am not sure if any tests of the second class are still there). Three years later the tests in the first class could probably be removed because they might not serve their purpose anymore. Had I broken something by removing a method we should have noticed by now.

PR #1442, opened 2022-02-14
Ensure channel and count fields for PHA are aliases of independent and dependent axes

DougBurke:
# Summary

Ensure that channel and the independent axis and counts and the dependent axis are synonymous for the DataPHA class. 

# Details

The `Data` setup has the independent and dependent axes (also known as "some form of x" and "y"). The `DataPHA` class extends this to include the `channel` and `counts` fields, which are essentially just different names for these two concepts, but the code did not enforce them. We have generally caught most cases where this caused a problem. but not all of them. This PR now makes `channel` an alias for the independent axis, and `counts` an alias for the `y` axis. Note that the independent axis is a tuple of components unlike `channel` so the aliasing requires a little bit of finessing.

The first commit just adds some explicit checks and then the second commit enforces the aliasing. We can see that it's only the new tests we added that need to be changed - that is, we haven't broken any existing PHA functionality.

This was identified as part of working on #1436 but I believe is separate from the validation work and so I've made this separate PR in the hope it is easier to review and we can get it merged sooner rather than later, as I have a bunch of planned changes which are already turning out to be an annoying pain to rebase.

### DIFF
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -1032,7 +1032,6 @@ def test_img_get_filter_compare_filtering(make_test_image):
     assert maska.max() == 1
 
 
-@pytest.mark.xfail
 def test_pha_change_channels(make_test_pha):
     """What happens if we change the channel/count values?
 
@@ -1065,13 +1064,11 @@ def test_pha_change_channels(make_test_pha):
     assert np.all(pha.get_indep()[0] == channels2)
     assert np.all(pha.get_dep() == counts2)
 
-    # XFAIL: These currently return the original data
     assert len(pha.indep) == 1
     assert np.all(pha.indep[0] == channels2)
     assert np.all(pha.dep == counts2)
 
 
-@pytest.mark.xfail
 def test_pha_add_channels(make_test_pha):
     """What happens if we increase the number of channels/counts?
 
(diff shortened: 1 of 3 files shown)


sherpa/astro/tests/test_astro_data2.py
```diff
@@ -1032,6 +1032,89 @@ def test_img_get_filter_compare_filtering(make_test_image):
...
+
+    channels = [1, 2, 3, 4]
+    counts = [1, 2, 0, 3]
+    assert np.all(pha.channel == channels)
+    assert np.all(pha.counts == counts)
+
+    assert len(pha.get_indep()) == 1
+    assert np.all(pha.get_indep()[0] == channels)
+    assert np.all(pha.get_dep() == counts)
+
```
hamogu, line 1051: Should we check `pha.x` and `pha.y`, too?

hamogu (commented):
Looks good in general, but I have a few inline questions. Yes, I *could* go and looks through the code to find out, but I know @DougBurke just looked at that, so I suspect he can answer much faster than I can find that suff.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -1032,6 +1032,89 @@ def test_img_get_filter_compare_filtering(make_test_image):
     assert maska.max() == 1
 
 
+def test_pha_change_channels(make_test_pha):
+    """What happens if we change the channel/count values?
+
+    We have several ways of getting the independent and dependent
+    axes.
```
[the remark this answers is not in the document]
DougBurke: Welcome to legacy software developmemt.

sherpa/astro/tests/test_astro_data2.py
DougBurke: Even though I try to catch everything, we at some point have to assume that "we have other tests that check, even if not under exactly the same condition" and hope we don't miss things (the coverage tests are useful but they unfortunately don't help us pick out certain paths that don't get covered).

PR #1453, opened 2022-02-18
Tests: mark those that require the group module

DougBurke:
# Summary

Note a number of tests that require the group module.

# Details

It is possible to build Sherpa with no group module but it's uncommon and we don't currently have a CI run that checks this. Which means we can end up not marking those cases that are known to fail. Let's fix this.

We could add a pip CI run to test this but

- this is not an often-used combination
- I have a bunch of much-more-important changes that completely re-work the CI runs that I do not want to break at this time


hamogu (approved):

DougBurke:
Ta for coming up for air after the proposals, @hamogu .

PR #1458, opened 2022-02-22
Improve several issues when building and testing Sherpa

DougBurke:
# Summary

Minor improvements to the build and test installation. The use of setuptools has been restricted to match the current advice from NumPy (< 60, from https://numpy.org/devdocs/reference/distutils_status_migration.html).

# Details

This has been rebased onto #1497 (well, a version of that PR) so that the tests pass.

This is a rework of #1335 and is pulled out of #1329 / #1352. The changes are

- one of the CI runs is a bit confused over develop versus install (it used install but than ran the tests, which cause the develop build to be re-build and used); I have tried to rework the CI actions to better handle things but it is not likely to be looked at for a while so this is a simple change
- restrict to `setuptools < 60` following NumPy advice (https://numpy.org/devdocs/reference/distutils_status_migration.html)
- ensure that XSPEC is configured correctly; I believe the CIAO build of Sherpa already makes this change so it will reduce the differences between them, and I have found it necessary to move to PEP 517 and related (e.g. #1329). I have no idea why this was not done a long time ago,

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -52,3 +52,9 @@ echo "dependencies: ${MATPLOTLIB} ${NUMPY} ${FITS} ${XSPEC} ${DOCSBUILD}"
 conda create --yes -n build python=${PYTHONVER} pip ${MATPLOTLIB} ${NUMPY} ${XSPEC} ${FITSBUILD} ${DOCSBUILD} ${compilers}
 
 conda activate build
+
+# Force setuptools earlier than 60 - see issue #1456. At present
+# conda seems to default to a working version, but leave this in
+# to ensure it holds, or we find out once we can no-longer do this.
+#
+conda install 'setuptools < 60'
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -59,7 +59,7 @@ jobs:
             os: ubuntu-latest
             python-version: 3.7
             numpy-version: 1.19
-            install-type: install
+            install-type: develop
             fits: astropy
             test-data: submodule
             matplotlib-version: 3
--- a/.github/workflows/ci-pip-arch.yml
+++ b/.github/workflows/ci-pip-arch.yml
@@ -73,6 +73,9 @@ jobs:
           python3 -m venv --system-site-packages tests
           source tests/bin/activate
 
+          # Ensure we use an old-enough setuptools: see issue #1456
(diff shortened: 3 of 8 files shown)


Marie-Terrell (approved):
The merge conflict due to the change that is now in main needs to be resolved. However, I'll approve of this for the other changes. Here are my notes:

I am fine with the setuptools pinning as that makes sense for now.

As for xspec_config: 
See #1335 for the last time xspec_config was testing for Sherpa. I will make the same note here along with a few other bits:

Without this change if one tries to pip install without Xspec present:
- The xspec tests are installed/listed and skipped 
- The sherpa/astro/xspec section is still present in the build log
- If you try to import sherpa.astro.xspec, you will see: "ImportError: cannot import name '_xspec' from 'sherpa.astro.xspec' ..."
- For minimum Linux python 3.9 with numpy install (no xspec/ds9/astropy/matplotlib): 4894 passed, 1567 skipped, 87 xfailed, 4 xpassed

With this change if one tries to pip install without Xspec present:
- The xspec tests are no longer installed/listed (with the exception of those outside the astro/xspec/tests directory)
- The sherpa/astro/xspec section is no longer present in the build log. The "running xspec_config" bit is now present.
- If you try to import sherpa.astro.xspec, you will now see:  "ModuleNotFoundError: No module named 'sherpa.astro.xspec'" 
- For minimum Linux python 3.9 with numpy install (no xspec/ds9/astropy/matplotlib): 4894 passed, 1461 skipped, 87 xfailed, 4 xpassed

This change is nice because:
- It does further reconcile the CIAO's Sherpa and Standalone Sherpa builds
- At least the ModuleNotFoundError and sherpa/astro/xspec build changes noted above are, in my opinion, cleaner. 

Note: I could see an arguments for or against the test change (sherpa/astro/xspec/tests no longer being installed and run with a skip) noted above.

Overall, I'm okay with this going in as long as everyone is okay with the changes noted above so long as the merge conflict is resolved.

DougBurke:
The conflict was just because of the way I had snuck on #1497 into this PR. As that PR has now been merged I've rebased.

I think I am happy with the build changes that @Marie-Terrell notes above. I think we'll be coming back to these issues - in particular what code/tests are installed once we get to looking at PEP 517 (and related), as there have been changes in how MANIFEST.in and related interact with the various settings for what to include or exclude from a build (or at least that's my interpretation of recent python packaging discussions; I may be wrong in my interpretation though).

PR #1461, opened 2022-02-23
Minor grouping cleanup

DougBurke:
# Summary

Minor internal change to how the group routines work for PHA objects.

# Details

This builds on #1453 (which just cleans up the grouping tests, and is the first commit).

Rather than have code like

    def group_a(self, ....):
        if not do_grouping:
            raise-an-error
        self._dynamic_group(pygroup.a, ..)

    def group_b(self, ....):
        if not do_grouping:
            raise-an-error
        self._dynamic_group(pygroup.b, ..)

we move the check for whether the group code has been loaded into `_dynamic_group`. To avoid the errors from trying to access `pygroup.a`, `pygroup.b`, ... we change `_dynamic_group` so that it is sent the name of the callable, rather than the callable itself. This means that in `_dynamic_group` we now need something like `func = getattr(pygroup, name)`.

To avoid possible regressions with any user code that has been calling `_dynamic_group` , and because I think it may still be useful to send a callable to `_dynamic_group`, we still allow the old calling convention.

One thing to note is that the group module is renamed on import - i.e. there's a line like `import group as pygroup` - since we have too much code that uses `group` as a symbol. Which is wy we talk about `pygrgroup.grpRandomMethod` rather than `group.grpRandomMethod`.

I have also removed a few comments which have been left over from the pre-git days of Sherpa and which no-longer add value to the code.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2926,6 +2926,32 @@ must be an integer.""")
         self.quality_filter = qual_flags
 
     def _dynamic_group(self, group_func, *args, **kwargs):
+        """Group the data using the given function and arguments.
+
+        In order to support the grouping module being optional this
+        routine will error out if it is not loaded (each time it is
+        called). To support Sherpa 4.14.0 and earlier group_func can
+        be a callable, but it is expected to be a string which is the
+        name of the callable from the "group" module (which in this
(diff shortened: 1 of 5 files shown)


hamogu (approved):
I think that's a good change.

PR #1462, opened 2022-02-25
Docs: fix up documentation of ui.set_grouping

DougBurke:
# Summary

Fix the documentation for set_grouping to match the code.

# Details

The documentation suggested that setting the grouping column automatically set the groupng flag, but it does not. A test has
been added to check this and the documentation updated. A few other tests have been added to check the set_grouping/quality handling.

We could change the behavior to match the old documentation.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_unit.py
@@ -1787,10 +1787,72 @@ def test_pha_column_has_correct_size(label, vals, clean_astro_ui):
     """
 
     ui.load_arrays(1, [1, 2, 3, 4, 5], [5, 4, 2, 3, 7], ui.DataPHA)
-    pha = ui.get_data()
 
     with pytest.raises(DataErr) as de:
         # This does not throw an error
         getattr(ui, f"set_{label}")(vals)
 
     assert str(de.value) == f"size mismatch between channel and {label}"
+
+
+@pytest.mark.parametrize("label", ["grouping", "quality"])
+def test_pha_column_checks_pha(label, clean_astro_ui):
+    """Check we error out with a non-PHA class"""
+
+    ui.load_arrays(1, [1, 2, 3, 4, 5], [5, 4, 2, 3, 7], ui.Data1D)
+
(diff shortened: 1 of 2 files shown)


sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -1787,10 +1787,72 @@ def test_pha_column_has_correct_size(label, vals, clean_astro_ui):
...
+    ui.load_arrays(1, [1, 2, 3, 4, 5], counts, ui.DataPHA)
+    pha = ui.get_data()
+
+    assert pha.grouping is None
+    assert not pha.grouped
+    assert ui.get_dep() == pytest.approx(counts)
+
+    ui.set_grouping([1, -1, 1, 1, 1])
+
+    # ui.get_dep returns somethng when grouped
```
hamogu, line 1843:
```suggestion
    # ui.get_dep returns something when grouped
```

sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -1787,10 +1787,72 @@ def test_pha_column_has_correct_size(label, vals, clean_astro_ui):
...
+
+
+def test_pha_what_does_get_dep_return_when_grouped(clean_astro_ui):
+    """Regression test for get_dep with grouped data"""
+
+    ui.load_arrays(1, [1, 2, 3, 4, 5], [5, 4, 2, 3, 7], ui.DataPHA)
+    ui.set_grouping([1, -1, 1, -1, -1])
+    ui.group()
+
+    # Looks like it's returning sum of group / number of channels in group
```
hamogu, line 1857:
```suggestion
    # Looks like it's returning mean of channel values in group
```

hamogu (approved):
Minor suggestions which could be applied but it's really OK either way.

DougBurke:
@anetasie - can you take a look to check whether you agree here: the docs will now explain what we do, but maybe we should do what the docs originally said (although that is significantly more work and will lead to more questions).

anetasie (approved):

PR #1463, opened 2022-03-01
rm a redundant call to calc_h

dtnguyen2:
# Note

Removed a redundant call the the ```calc_h``` method in the class ```fdJac```


DougBurke (approved):
There's some code changes we could do here - e.g.

- move the class outside the function and to the top-level as I believe that would be more efficient but I am not sure how important that is
- make routines use numpy array indexing rather than a c-style for loop
- add some level of documentation

but I'll leave those for other PRs

DougBurke:
The copyright year does need updating though!

dtnguyen2:
done

PR #1468, opened 2022-03-17
Lazy-load image backend to avoid ds9 missing warning on import

taldcroft:
As someone who does not have ds9 installed but uses sherpa, the warning about DS9 being missing every time I import `sherpa.ui` has always annoyed me. 

This had a real consequence for the Chandra operations Python integration testing that we do, which runs unit tests for a couple dozen of our packages and then looks for "warning" or "error" in the output. This resulted in this workaround https://github.com/sot/find_attitude/pull/18 to just turn off that logger warning, but maybe this PR is a better way using PEP 562 introduced in Python 3.7.

Speaking of lazy, I was too lazy to run unit tests. What I did was actually just edit an installed version of sherpa and then do this simple functional test demonstrating that the warning doesn't appear until `backend` is accessed.
```
$ ipython
In [1]: import sherpa.image

In [2]: backend = sherpa.image.backend

In [3]: backend
Out[3]: <module 'sherpa.image.dummy_backend' from '/Users/aldcroft/miniconda3/envs/ska3-next/lib/python3.8/site-packages/sherpa/image/dummy_backend.py'>
```

This was also an excuse to learn about PEP 562, so if this PR is rejected I won't mind. :smile:

### DIFF
--- a/sherpa/image/__init__.py
+++ b/sherpa/image/__init__.py
@@ -29,25 +29,48 @@ References
 
 ..  [XPA] "The XPA Messaging System", https://hea-www.harvard.edu/saord/xpa/
 
-"""
+"""  # noqa
 
+import sys
 import numpy
 from sherpa.utils import NoNewAttributesAfterInit, bool_cast
 
 import logging
 warning = logging.getLogger(__name__).warning
-backend = None
 
-try:
-    from . import ds9_backend as backend
+# Make a reference to this module so __getattr__ below can be used for backend.
+__module__ = sys.modules[__name__]
 
-except Exception as e:
-    # if DS9 is not found for some reason, like inside gdb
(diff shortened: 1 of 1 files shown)


DougBurke:
Can you add a link to PEP 562 as I'm too lazy to hunt for this.

I do want to improve the ds9 code, as we currently have two separate backends we could use but only ever exercise one of them, so there's code to look at. Also, @hamogu has been working on our other backends (I/O and plotting) and it would be good to have a consistent interface (where relevant).

taldcroft:
PEP 562 link added.

taldcroft:
> We are actually in the process of changing a "backend" from a module to a class. 

Perfect, that is an obviously better solution which I'm looking forward to. Closing this PR!

Issue #1334, opened 2021-10-26
GH actions (conda) failures


DougBurke:
A number of the conda builds are reliably failing with the following - this example is taken from a PR by @harlantc and is  https://github.com/sherpa/sherpa/runs/3999185295?check_suite_focus=true but it happens for a number of PRs

```
=================================== FAILURES ===================================
__________________ test_regrid1d_works_with_convolution_style __________________

    def test_regrid1d_works_with_convolution_style():
        """This doesn't really test more than the previous
        model-evaluation tests.
        """
    
        smdl = StepLo1D()
        smdl.xcut = 100
        smdl.ampl = 10
    
        cmdl = Const1D()
        cmdl.c0 = -500
    
        imdl = smdl + cmdl
    
        # Set up the convolution kernel
        #
        gsmooth = Gauss1D()
        psf = PSFModel('psf', gsmooth)
    
        smoothed = psf(imdl)
    
        # This is the model that will be evaluated
        #
        regrid = ModelDomainRegridder1D()
        smoothed_regrid = regrid.apply_to(smoothed)
    
        # Ignoring edge effects, the smoothed step function drops from
        # x=100 down to x~120 (it is not clear why it doesn't smooth
        # below x=100, but maybe I've set up the convolution wrong).
        # So, if the regrid model evaluates x=0 to 200 but the requested
        # grid is from x=102 ro 180, then we should see a difference
        # to evaluating without regrid.
        #
        xfull = np.arange(0, 200, 0.5)
        xout = np.arange(101, 180, 0.5)
    
        regrid.grid = xfull
    
        # fake up a data object for the fold method
        # TODO: it is not clear to me what grid should be used here;
        #       xfull or xout
        #
        d = Data1D('fake', xfull, xfull * 0)
        psf.fold(d)
    
        y_regrid = smoothed_regrid(xout)
    
        # calculate the expected values
        y_full = smoothed(xfull)
    
        # since the grids have the same binning, it is a simple extraction
        idx0, = np.where(xfull == xout[0])
        idx1, = np.where(xfull == xout[-1])
        y_expected = y_full[idx0[0]:idx1[0] + 1]
    
>       assert_allclose(y_regrid, y_expected, atol=1e-10, rtol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=0, atol=1e-10
E       
E       x and y nan location mismatch:
[...]

PR #1472, opened 2022-03-18
tests: remove work-around for #1334 from #1336

DougBurke:
# Summary

Remove a test work around for our CI tests failing (issue #1334)

# Details

In #1334 we noted that on CI runs some builds would fail. It was never obvious why, so we marked them xfail in #1336. Remove this setting. It looks like we can do this now (as I don't know why the problem occurred I don't know if it's some particular version of OTS we use that caused problems., or maybe some of the build changes like #1329, or some combination, or ...).

### DIFF
--- a/sherpa/models/tests/test_regrid_unit.py
+++ b/sherpa/models/tests/test_regrid_unit.py
@@ -845,7 +845,6 @@ class ReNormalizerModel1DInt(CompositeModel, ArithmeticModel):
 
 # TODO: more tests when regridding the models
 
-@pytest.mark.xfail  # see #1334
 def test_regrid1d_works_with_convolution_style():
     """This doesn't really test more than the previous
     model-evaluation tests.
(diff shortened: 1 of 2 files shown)


DougBurke:
:crossed_fingers: we can make this change now

DougBurke:
Well, that shows me the problem still exists!

DougBurke:
I'm going to leave this as we can retest ooccasionally.

DougBurke:
I've assigned this to @harlantc and @Marie-Terrell but it only needs one person to look at it :-)

DougBurke:
Rebased just to check it still works

Marie-Terrell (approved):
Looks good to me

DougBurke:
So, it turns out that there are two tests that are still marked as xfail because of this that I've somehow missed:

```
% find sherpa -name test\*py -exec grep 1334 {} /dev/null \;
sherpa/utils/tests/test_psf_lowlevel.py:@pytest.mark.xfail  # see #1334
sherpa/tests/test_instrument.py:@pytest.mark.xfail  # see #1334
```

I'll add a PR at some point, but it's not a high priority/

Issue #1470, opened 2022-03-17
How to add validation to the data objects?


DougBurke:
# Why validate?

The idea is that if we can ensure that the fields of the `Data`
objects obey certain constraints - such as they have the same length,
are one dimensional, or have a certain type - then the code is easier
to write and the user will not see errors that can be hard to
understand.

The aim is to ensure that the fields match, and that changes made to
the object - such as a filter - make sense. There is already some
validation in place (mainly ensuring that some fields are converted to
NumPy arrays), so the question becomes how much should we do?  I focus
here on the Data class, and those classes that it uses, although there
may be some related work that could be done for other classes.

## Terminology

I am going to try to use the term sequence for an array-like object.
In actuality we will use a NumPy ndarray for these but users can
supply any sequence-like object. Similarly I shall try to use field to
refer to an element of the data object, but sometimes will also use
axis.

For this document I will use the terms independent and dependent axes.
The code also uses terms like `x`, `x0`, `xlo`, `x0lo` for the
independent axis and `y` for the dependent axis. The error axes, or
fields, are `staterror` and `syserror`: these are referred to as
related fields. The `mask` field should also be considered a related
field and has a the ability to be a scalar, in this case a boolean, or
a sequence.

The PHA class also adds a number of related fields, such as grouping,
quality, and the "scaling" fields (`areascal` and `backscal`). Some of
these fields can be scalars or sequences.

## Constraints from the current design

The current design requires the independent and dependent axes to be
specified when creating a `Data` object, but they can be set to
`None`.  This means that it is possible to "build up" a data object,
rather than enforcing the constraints when `__init__` is called. The
argments for the independent and dependent axes are assumed to be
one-dimensional sequences, and the independent axis uses a tuple to
store the different components, such as the low and high edges for the
"integrated" data set and the two dimensions for the 2D case.

Users can change fields either wholesale - that is, replace the
`staterror` field with this sequence or set the `syserror` field to
`None` - or by direct manipulation of the elements of the sequence.

**Question** Supporting multi-dimensional arguments

#1446

The whole design of Sherpa is around one-dimensional sequences of the
independent and dependent axes. This allows us to think about working
on a point in the data space at a time, and just looping over the
sequence of points, and is built in to how we handle filters.

Should we allow 2D data values to be sent in when setting fields for
2D data sets? The code would internally flatten the values.

This should be possible but I have not looked at it yet.

**Question** Should a data object know what dimension it is?

#1460

[...]

PR #1473, opened 2022-03-21
Data test minor cleanup

DougBurke:
# Summary

Minor test clean-ups.

# Details

This is part of my work on #1470 but they are really unrelated to that wirk, they are just test clean-ups. The main changes are that I have finally removed the `create_arf/delta_rmf` functions from the test file as they are now available in `sherpa.astro.instrument` (these routines were originally written for the tests and than got moved over to the instrument code but the test was never updated), and I fixed up a failing test (added in #1440) to still be failing, but to fail properly rather than due to a typo in the test itself.

One of the tests has to create a `DataPHA` instance that has a different number of elements to the original object. This is currently easy, as you can change the fields such as `counts` and `channels`, at will. However, if we ever implement #1470 then this may not be possible, so wthe test now explicitly creates a new `DataPHA` instance (this is `test_stats_calc_stat_wstat_diffbins` in `sherpa/stats/tests/test_stats_unit.py`). Both approaches currently work, but the new version should be valid however we implement #1470.

### DIFF
--- a/sherpa/astro/io/tests/test_io_pha.py
+++ b/sherpa/astro/io/tests/test_io_pha.py
@@ -1276,10 +1276,16 @@ def test_write_pha_missing_column(column, is_ascii, tmp_path):
     At the moment we write out the file.
     """
 
-    chans = np.arange(1, 4, dtype=np.int32)
-    counts = np.arange(1, 4, dtype=np.int32)
+    if column == "channel":
+        chans = None
+        counts = np.arange(1, 4, dtype=np.int32)
+    elif column == "counts":
+        chans = np.arange(1, 4, dtype=np.int32)
+        counts = None
+    else:
+        assert False, "unknown column"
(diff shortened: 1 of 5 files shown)


sherpa/astro/io/tests/test_io_pha.py
```diff
@@ -1276,10 +1276,16 @@ def test_write_pha_missing_column(column, is_ascii, tmp_path):
     At the moment we write out the file.
     """
 
-    chans = np.arange(1, 4, dtype=np.int32)
-    counts = np.arange(1, 4, dtype=np.int32)
+    if column == "channel":
```
hamogu, line 1279: I' confused by "if column == channel, then chans = None" instead of the other way around, but I guess in te context of this test that makes sense.

hamogu (approved):

PR #1475, opened 2022-03-25
Minor cleanup of the data and instrument code

DougBurke:
# Summary

Minor code clean up of the data and instrument classes.

# Details

This is part of the preparation work for #1470 but these commits do not change the behavior of Sherpa. There are a number of lint-style changes - e.g. change

```python
if foo:
    return bar
elif foofoo:
    return barbar
else:
    return barbarbar
```

to

```python
if foo:
    return bar

if foofoo:
    return barbar

return barbarbar
```

or

```python
if foo is not None:
    foo = do-something-to-foo
return foo
```

to

```python
if foo is None:
    return None

return do-something-to-foo
```

as I find these make it clearer what the logic is (and, in the latter case, the `do-something-to-foo` part may consist of multiple parts. A number of changes to use `super()` rather than explicitly label the super-class, have also been made.

There are two "larger" changes 

- the removal of the `DataPHA.set_dep` method, which is now possible because of #1442 (where the channel/independent axis and counts/depenent axis were aliased rather than carrying around the same data).
- removal of the `DataPHA.eval_model` method since it does the same as the super-class, and using it in the `DataPHA.eval_model_to_fit` method rather than repeating the logic.

The changes to `sherpa/instrument.py` are more extensive, but it's mainly re-arranging information so that it's clearer to me. and hopefully other readers, what is going on.

### DIFF
--- a/sherpa/data.py
+++ b/sherpa/data.py
@@ -425,9 +425,15 @@ class DataSpaceND():
         return self.indep
 
 
+# NOTE:
+#
+# Although this is labelled as supporting N-D datasets, the
+# implementation assumes that the data has been flattened to 1D arrays
+# - in particular the notice method. It is likely that we can document
+# this - i.e. that the mask is going to be 1D.
+#
 class Filter():
-    """
-    A class for representing filters of N-Dimentional datasets.
+    """A class for representing filters of N-Dimensional datasets.
     """
     def __init__(self):
         self._mask = True
@@ -444,31 +450,43 @@ class Filter():
 
     @mask.setter
     def mask(self, val):
-        if (val is True) or (val is False):
-            self._mask = val
-        # if val is of type np.bool_ and True, it failed the previous test because
-        # "is True" compares with Python "True" singelton.
-        # Yet, we do not want to allow arbitrary values that evaluate as True.
+        # Special case the numpy.ma.nomask setting (which evaluates to False
+        # but is not a bool).
+        #
+        if val is None:
+            raise DataErr('ismask')
+
+        elif not numpy.isscalar(val):
+            # We could catch an error if the value can not be converted,
+            # but wat values can not be converted to a bool?
+            self._mask = numpy.asarray(val, bool)
+
         elif val is numpy.ma.nomask:
             self._mask = True
-        elif numpy.isscalar(val) and isinstance(val, numpy.bool_):
+
+        elif (val is True) or (val is False):
+            self._mask = val
+
+        elif isinstance(val, numpy.bool_):
             self._mask = bool(val)
-        elif (val is None) or numpy.isscalar(val):
-            raise DataErr('ismask')
+
         else:
-            self._mask = numpy.asarray(val, numpy.bool_)
+            # Note that setting mask to 2.0 will fail, but an array
+            # of 2.0's will get converted to booleans.
+            #
+            raise DataErr('ismask')
 
     def apply(self, array):
         """Apply this filter to an array
 
         Parameters
         ----------
-        array : array_like
+        array : array_like or None
             Array to be filtered
 
         Returns
         -------
-        array_like : filtered array
+        array_like : ndarray or None
 
         Raises
         ------
@@ -488,10 +506,12 @@ class Filter():
         if self.mask is False:
             raise DataErr('notmask')
 
+        # Ensure we always return a ndarray.
+        #
+        array = numpy.asarray(array)
         if self.mask is True:
             return array
 
-        array = numpy.asarray(array)
         if array.shape != self.mask.shape:
             raise DataErr('mismatch', 'mask', 'data array')
         return array[self.mask]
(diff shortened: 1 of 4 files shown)


DougBurke:
Rebased as #1474 (and a lot of other changes) have been merged

sherpa/data.py
```diff
@@ -444,31 +454,43 @@ def mask(self):
...
+            self._mask = numpy.asarray(val, bool)
+
         elif val is numpy.ma.nomask:
             self._mask = True
-        elif numpy.isscalar(val) and isinstance(val, numpy.bool_):
+
+        elif (val is True) or (val is False):
+            self._mask = val
+
+        elif isinstance(val, numpy.bool_):
```
hamogu, line 474: I suggest we keep this comment, it's not obvious to me and I wrote that code originally!
```suggestion
        # if val is of type np.bool_ and True, it failed the previous test because
        # "is True" compares with Python "True" singelton.
        # Yet, we do not want to allow arbitrary values that evaluate as True.
        elif isinstance(val, numpy.bool_):
```

sherpa/instrument.py
```diff
@@ -413,102 +430,114 @@ def _create_tail_grid(axis_list):
...
-        return self._center
+    def _get_field(self, name):
+        """Return the field value"""
+        field = getattr(self, name)
+        if field is not None and len(field) == 1:
+            return field[0]
+
+        return field
+
+    def _set_field(self, name, vals):
```
hamogu, line 473: I know it's a pain to write docs for stuff like this and this is internal, but given that it's all about reformatting, it might be worth it to write down what "vals" is supposed to be is a docstring for the benefit of future developers. Somethink like "vals is any value, scalar or array. If it's not a list, tuple or np.array, this routine will wrap it into a one-element array before setting it. (Why? (I don't know)"

sherpa/instrument.py
```diff
@@ -413,102 +430,114 @@ def _create_tail_grid(axis_list):
...
+
+    def _get_center(self):
+        return self._get_field("_center")
+
+    def _set_center(self, vals):
+        self._set_field("_center", vals)
 
     center = property(_get_center, _set_center, doc='array of size parameters')
 
     def _get_size(self):
```
hamogu, line 496: Given the simplicity of these (it's now just one line of `self._get_field("_size")`) I would prefer to get rid of all those internal helpers altogether and search and replace `xxx._get_size()` directly with `xxx._get_field("_size")`.

hamogu (changes requested):
I noticed that this PR has been sitting for while with no clear action assigned, probably because we were waiting for the 4.14.1 to be done. So, I went ahead and reviewed; I only have a few minor suggestions and will take "I don't agree, let's leave as is now" as an answer.
Since this is not about functionality, but about style of code an readability for future developers, it can go in any time, but let's not wait forever. This type of cleanup is helpful, but also tends to create merge conflicts somewhere if we wait too long.

sherpa/instrument.py
```diff
@@ -540,122 +570,104 @@ def __call__(self, model, session=None):
...
-            self.model = Kernel(dshape, kshape, **kargs)
+            for kwarg in ['is_model', 'lo', 'hi', 'width', 'size']:
+                kwargs.pop(kwarg)
+
+            self.model = Kernel(dshape, kshape, **kwargs)
             return
 
-        if radial:
-            self.model = RadialProfileKernel(dshape, kshape, **kargs)
+        if int(self.radial.val):
```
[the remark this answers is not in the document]
hamogu: I suspected that in practice, this is only used with positive numbers (dimension, number of pixels) and that this guards against 0.000000001, which should be True to be rounded down to 0 because of machine-precision. But this probably needs a comment, give that you don't remember either...

DougBurke:
I've made changes to this code in #1477 (mainly to add more tests and clean up some of the commits), so II'm going to close this (as the code is included in #1477). It would be nice if GH supported some form of stacked commits...

I'll add changes for the comments above to #1477

sherpa/data.py
DougBurke: In #1477 I've re-written this so there's introductory text to explain this (at least, in my current branch which hasn't been pushed yet).

sherpa/instrument.py
DougBurke: The _get/_set_[size|center|...] methods are only used in `property` calls - e.g. line 494 - so there's only one line that calls `_get_field(self, "size")`.

sherpa/instrument.py
DougBurke: I added a note in #1477

sherpa/instrument.py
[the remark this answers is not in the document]
DougBurke: Aha: this is from the original code, which has

```python
(size, center, origin, kargs['norm'], radial) = (self.size, self.center, self.origin, bool_cast(self.norm.val), int(self.radial.val))
```

I have added a comment i #1477 to say `# TODO: why is this not 'self.radial.val > 0'?` but I don't want to change the code here.

PR #1455, opened 2022-02-18
RTD: update the build-from-CIAO documentation

DougBurke:
# Summary

Update the "build with CIAO" documentation to match the CIAO 4.14 conda instructions and automate the process.

# Details

Adds the `scripts/use_ciao_config` script that automates the setting of the `setup.cfg` file for users of CIAO installed with conda, and update the documentation to match.

Note that this is how I found issue #1452 and changes to that could influence this, but they can be handled separately.

PR #1484, opened 2022-04-11
Include ciao region patch

DougBurke:
# Summary

Allow the CXC Data Model library to be used to parse region files, if available. This is primarily intended for building Sherpa as part of CIAO.

# Details

The main aim is to reduce the extra code that CIAO needs to keep around to build Sherpa, which should reduce the chance of build-failures in CIAO if the region code is changed. It should also allow users to take advantage of this support when building Sherpa against conda, as discussed at https://sherpa.readthedocs.io/en/4.14.0/developer/index.html#install-from-source-in-conda - see also #1455 

The way the code works is that if the `setup.cfg` file has set the `sherpa_config.region-use-cxc-parser` value to `True` then we trigger building the `sherpa.astro.utils._region` module with the `USE_CXCDM_PARSER` define/macro, which uses the `dmRegParse` from `ascdm` rather than the parsing code from the region library. The `region-include_dirs` and `region-lib-dirs` settings can be used to indicate the location of `ascdm.h` and `libascdm.<whatever>`, but it is expected that no change is meeded, as they should be the same as `cxcregion.h` and `libregion.<whatever>`. The only value that may need changing is `region-libraries`, from `region` to `region ascdm` (at least it is needed when building Sherpa against conda/CIAO on my machine; it may not be needed when building Sherpa for CIAO).

I think setting some sort of define/macro is realistically the only way to support the ascdm version versus the region parser.

I did think about whether we could automate the detection of this case - that is, remove the need for the `region-use-cxc-parser` option - but I decided that here it's better to be explicit rather than implicit. It's also the case that the automatic-detection approach would require significantly-more work than the approach used here.

What I'm not sure about is whether this works for all builds - so ciao-install rather than conda, and macOS versus Linux. 

I'm interested if people can come up with different approaches.

### DIFF
--- a/helpers/extensions/__init__.py
+++ b/helpers/extensions/__init__.py
@@ -81,12 +81,13 @@ def build_wcs_ext(library_dirs, include_dirs, libraries):
                  libraries=libraries,
                  depends=get_deps(['extension']))
 
-def build_region_ext(library_dirs, include_dirs, libraries):
+def build_region_ext(library_dirs, include_dirs, libraries, define_macros=None):
     return Extension('sherpa.astro.utils._region',
                  ['sherpa/astro/utils/src/_region.cc'],
                  sherpa_inc + include_dirs,
                  library_dirs=library_dirs,
                  libraries=(libraries),
+                 define_macros=define_macros,
                  depends=get_deps(['extension']))
 
 def build_xspec_ext(library_dirs, include_dirs, libraries, define_macros=None):
--- a/helpers/sherpa_config.py
+++ b/helpers/sherpa_config.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2014, 2015, 2016, 2020  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2014, 2015, 2016, 2020, 2022
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -38,6 +39,7 @@ class sherpa_config(Command):
                     ('region-include-dirs', None, "Where the region headers are located, if region is 'local'"),
                     ('region-lib-dirs', None, "Where the region libraries are located, if region is 'local'"),
                     ('region-libraries', None, "Name of the libraries that should be linked as region"),
+                    ('region-use-cxc-parser', None, "If set to True than use the CXC Data Model library to parse region files"),
                     ('wcs', None, "Whether Sherpa should build the embedded wcs library, which is the default behavior: set to 'local' to make Sherpa link against existing libraries on the system.)"),
                     ('wcs-include-dirs', None, "Where the wcs subroutines headers are located"),
                     ('wcs-lib-dirs', None, "Where the wcs subroutines libraries are located"),
@@ -61,6 +63,7 @@ class sherpa_config(Command):
         self.region_include_dirs=None
         self.region_lib_dirs=None
         self.region_libraries='region'
+        self.region_use_cxc_parser = False
         self.wcs=None
         self.wcs_include_dirs=None
         self.wcs_lib_dirs=None
@@ -84,6 +87,12 @@ class sherpa_config(Command):
 
         if self.region_lib_dirs is None:
             self.region_lib_dirs=self.install_dir+'/lib'
+
+        # It is not clear that the other boolean options work
+        # correctly when set to False/false.
+        #
+        if not isinstance(self.region_use_cxc_parser, bool):
+            self.region_use_cxc_parser = self.region_use_cxc_parser.lower() == "true"
 
         if self.wcs_include_dirs is None:
             self.wcs_include_dirs=self.install_dir+'/include'
@@ -113,7 +122,16 @@ class sherpa_config(Command):
             configure.append('--enable-region')
         ld2, inc2, l2 = build_lib_arrays(self, 'region')
         ld, inc, l = (ld1+ld2, inc1+inc2, l1+l2)
-        self.distribution.ext_modules.append(build_ext('region', ld, inc, l))
+
+        # can we find ascdm.h (if so we can support FITS region files as long
+        # as the other settings are set up to include ascdm)?
+        #
+        region_macros = None
+        if self.region_use_cxc_parser:
+            region_macros = [("USE_CXCDM_PARSER", None)]
+
+        self.distribution.ext_modules.append(build_ext('region', ld, inc, l,
+                                                       define_macros=region_macros))
 
         if not self.disable_group:
             configure.append('--enable-group')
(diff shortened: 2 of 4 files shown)


sherpa/astro/utils/src/_region.cc
```diff
@@ -42,14 +50,32 @@ static PyObject* region_subtract( PyRegion* self, PyObject* args, PyObject *kwar
...
+// which case we are restricted to ASCII region files via
+// regReadAsciiRegion.
+//
 static regRegion* parse_string( char* str, int fileflag ) {
 
   regRegion *reg = NULL;
+
+#ifdef USE_CXCDM_PARSER
+
+  std::string input(str);
```
wmclaugh, line 65: Since str is a char* - initializing string class variable 'input' with 'str' can segv if the supplied value for str in parse_string() is set to NULL. The code needs to check that str is  non-NULL before this assignement.

DougBurke: So, this part is just a slight re-wording of the code that CIAO was using, which didn't include any error checking. Fortunately, in this case, as `parse_string` isn't exported and only used in one place (`pyRegion_New`), where the string argument is created. Thanks to the way that `PyArg_ParseTupleAndKeywords` works when given a `s` format argument we are guaranteed that `objS` isn't `NULL`, so the call to `reg = parse_string( objS, fileflag );` isn't a problem.

I can add some validation though as this is not a part of the code where we care performance or excess memory.

DougBurke: This is my proposed change:

```
diff --git a/sherpa/astro/utils/src/_region.cc b/sherpa/astro/utils/src/_region.cc
index 04ce29a3..f15bcf83 100644
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -56,13 +56,12 @@ static PyObject* region_invert( PyRegion* self, PyObject* args );
 // which case we are restricted to ASCII region files via
 // regReadAsciiRegion.
 //
-static regRegion* parse_string( char* str, int fileflag ) {
+static regRegion* parse_string( std::string input, int fileflag ) {

   regRegion *reg = NULL;

 #ifdef USE_CXCDM_PARSER

-  std::string input(str);
   if( fileflag )
     input = "region(" + input + ")";

@@ -71,9 +70,9 @@ static regRegion* parse_string( char* str, int fileflag ) {
 #else

   if( fileflag ) {
-    reg = regReadAsciiRegion( str, 0 ); // Verbosity set to 0
+    reg = regReadAsciiRegion( (char*)input.c_str(), 0 ); // Verbosity set to 0
   } else {
-    reg = regParse( str );
+    reg = regParse( (char*)input.c_str() );
   }
 #endif

@@ -109,7 +108,12 @@ static PyObject* pyRegion_new(PyTypeObject *type, PyObject *args,
                                     &objS, &fileflag ) )
     return NULL;

-  reg = parse_string( objS, fileflag );
+  // The string is not NULL (otherwise the PyArg_ParseTupleAndKeywords
+  // call would have failed) so we don't need to check here.
+  //
+  std::string name(objS);
+
+  reg = parse_string( name, fileflag );
   if (!reg) {
     const char *fmt;
     if (fileflag) {
```

Is this okay or would you like an explicit check (or have this done some other way)?

DougBurke: For reference the description of the `format` argument to `PyArg_ParseTupleXXX` can be found at https://docs.python.org/3.8/c-api/arg.html#strings-and-buffers

DougBurke:
So I've tried to address @wmclaugh review comment (by not-exactly-doing what was asked). Is this okay? I decided not to bother worrying about copy-versus-move semantics (e.g. whether I should use a reference rather than a value) but decided that I didn't think it mattered here.

helpers/extensions/__init__.py
```diff
@@ -81,12 +81,13 @@ def build_wcs_ext(library_dirs, include_dirs, libraries):
                  libraries=libraries,
                  depends=get_deps(['extension']))
 
-def build_region_ext(library_dirs, include_dirs, libraries):
+def build_region_ext(library_dirs, include_dirs, libraries, define_macros=None):
```
DougBurke, line 84: Is this helpful (for when we come back and look at this n years in the future)?

```suggestion
# The _region module can be built using the CXC Data Model interface (which allows
# FITS region files to be read, not just ASCII files). This requires setting the
# USE_CXCDM_PARSER macro, which is why this routine takes an optional
# define_macros argument.
#
def build_region_ext(library_dirs, include_dirs, libraries, define_macros=None):
```

helpers/sherpa_config.py
```diff
@@ -85,6 +88,12 @@ def finalize_options(self):
         if self.region_lib_dirs is None:
             self.region_lib_dirs=self.install_dir+'/lib'
 
+        # It is not clear that the other boolean options work
+        # correctly when set to False/false.
+        #
+        if not isinstance(self.region_use_cxc_parser, bool):
+            self.region_use_cxc_parser = self.region_use_cxc_parser.lower() == "true"
+
```
DougBurke, line 96: I don't really understand how the config value from `setup.cfg` get converted to Python values. I would imagine it just uses string conversion, but some of our options seem to treat `"true"` / `"false"` values as booleans (or maybe all they really care about is if the value is set or not, the value is irrelevant), which makes me wonder just what is going on,

hamogu: By itself, all values are read as strings, put configparser has "getboolean", "getint" and "getfloat" methods that retrieve the value and perform conversion at that same time: https://docs.python.org/3/library/configparser.html#supported-datatypes

DougBurke:
There is an interaction here with #1455 so that can be addressed once the first of these two is merged (that is, adding the new setting to the `use_ciao_config` script added in 1455).

DougBurke:
Hmm. So we can add a test of this functionality as our test-data area (when installed) includes FITS region files, so we can do something like

```
>>> from sherpa.astro.utils._region import Region
>>> r = Region("sherpa-test-data/sherpatest/acisf07999_000N001_r0035_reg3.fits", 1)
>>> print(r)
Ellipse(3145.89,4520.79,37.0615,15.3882,92.2273)
>>> r2 = Region("sherpa-test-data/sherpatest/acisf07999_000N001_r0035_reg3.fits[bkgreg]", 1)
>>> r2
Ellipse(3145.89,4520.79,185.308,76.9408,92.2273)&!Ellipse(3145.89,4520.79,37.0615,15.3882,92.2273)&!Ellipse(3037.38,4613.22,53.836,24.6396,176.674)
```

but we don't have a way to identify when we expect this to work. So we could add (in a follow-up PR) some way to indicate whether the region code supports this (e.g. something set at compile time) and then add a test.

Marie-Terrell (approved):
Looks good. I've tested this on a standalone build (with and without change) and for a CIAO-install build (without our special region patch and with our setup.cfg updated to set the new value to true). I do not see any new failures or changes to regression test runs.

PR #1477, opened 2022-03-29
Add data validation

DougBurke:
# Summary

Improve the validation of arguments when creating Data objects.

Arrays sent to `Data` objects are now converted to ndarrays rather than sometimes being converted and sometimes not. The independent axis is now made read-only, so the only way to change a value is to replace the whole axis, in which case any existing filter or spatial region is removed. The size of a data object is now fixed, so once a field - such as the independent, dependent, or staterror columns - is set the size can not be changed. This means that a new object will have to be created rather than shortening or extending the fields of an existing object. Data objects have now gained a ndim attribute and this is used to prevent a number of error cases, such as trying to evaluate a 2D model on a 1D dataset.

A number of calls will now error out - normally raising a `sherpa.utils.err.DataErr` exception - when they are sent an argument of the wrong size whereas previously the call would have succeeded, but left the data object in an invalid state.

# Details

Fix #1470 and a number of other issues. [Aside: I had copied some of these commits into #1475 to try and reduce the burden of this PR, but I've just closed that as it has diverged enough from changes I've made here to make it hard to manage]. The code has been split up into a large number of small steps, as it makes it easier to see what such a small change makes to the test suite. I've spent a lot of effort breaking up the commit in this way but there are many ways to do this, and this is not guaranteed to be a minimal set of changes. There's also the fact that some changes will end up being re-worked in later commits (normally because of later changes), so please be aware of this when reviewing commits individually.

The aims are to move the validation to as "high-up" the chain as possible - so, for instance, the independent axis checks are made in the underlying `Axis` class (which was not-at-all-obvious to me when I started this) rather than in the `Data.__init__` method. For example, the validation does not generally appear in the ui code (there's always some place where you do need to do this).

There is an attempt to make the error messages include contextual information - that is, not "there is a mismatch between foo and bar" but "The sizes of foo and bar do not match: 23 vs 1024" - but there are cases when more (or less) context could be added and it is always going to be a trade off.

One change made in this PR is that the size of a `Data` object is, once set (that is, a field such as the independent axis, y, or staterror is set), can not be changed. This is a breaking change, but I argue that it is better for the code to be re-written to create a new data object rather than re-size an existing object. There is actually one test that I re-wrote because of this change.

As part of this work, `Data` objects now get `ndim` and `size` attributes, which make it easier to check these values. The`len` command can be used on `Data` obects: it returns the `size` value (unless it's not set, in which case 0 is returned). The `ndim` change works well to extend the support we added to models in #900. There's an example below showing this.

[...]

PR #1486, opened 2022-04-13
Improve filtering/grouping behavior

### LINKED ISSUE #1160: Changing the grouping messes up the filter

DougBurke:
# Summary

Changing the grouping array of a PHA dataset (with set_grouping from the UI layer or by setting the grouping attribute of a DataPHA object) will now re-create the filter, to fix #1160. It is suggested that the filter expression (e.g. as returned by the get_filter routine for uses of the UI layer) is checked to ensure the result is still sensible.

The low-level grouping code has been updated to add data validation to check when it is called improperly.

# Details

The next PR following this is #1637 

## 1160 fix

The user-relevant change is to fix #1160 - which is that grouping the data can mess up the existing filter. This has been addressed by re-filtering the data after a change to the grouping, which ensures that the DataPHA  mask is correctly set up. One consequence of this work is that the resulting filter may be significantly different than before the grouping, since we do not record the intended filter but the result of the filter when applied to the currently-grouped data (see a discussion in #1655). We can help users out by reporting the filter (in the same way we now do for notice/ignore calls) which is done in #1637 as a follow-up PR. An alternative approach would be to clear the filter whenever the grouping changes, but I don't think that will really help as, unless we have something like #1637 users don't know what's happened to the data: we could warn the user the filter has changed but I think that it's not particularly helpful (but this is very-much an opinion and not a fact, others  may disagree here). 

## validation

The sherpa.astro.utils.expand_grouped_mask routine is written in C++ and provided by the sherpa.astro.utils._utils.expand_grouped_mask routine. We add

- documentation
- data validation
  - arguments are 1D arrays
  - arguments are not empty
  - arguments have the correct size (thanks to the way the routine works it's ever-so-slightly-tricky to check)
 
I know @hamogu will claim this (grouping) could be done in Python rather than C++ but I'm not willing to make that change yet (if we were to do that we'd want to check a number of routines, not just remove them one-by-one, in particular with an eye on performance). Technically these changes are not needed, as it doesn't lead to any changes to the code that calls expand_grouped_mask. However, in investigating #1160 I did hit some cases where this validation would have pointed out there was a problem with my proposed solution (rather than just silently creating the wrong answer), so I think it's worth it.

There is a question over whether we should allow the arguments to both be empty (in which case the return would be empty), but I don't believe that's a useful case for us.

This is obviously related to the data-validation work - #1470 / #1477 - but it's technically different and can be looked at separately (and in fact they have been closed now, but not when this was originally written).

---

# Example

## Issues with filtering

In the following we can see that as we regroup things the original filter gets "diluted", in particular we lose the gap between 0.9 and 1 keV. Thanks to the way we do filtering there's no real way to stop this, but at least with #1637 you would at least be able to see this change "directly".

```
In [4]: load_pha("sherpa-test-data/sherpatest/3c273.pi")

In [5]: notice(0.5, 0.9)
dataset 1: 0.00146:14.9504 -> 0.4672:0.949 Energy (keV)

In [6]: notice(1, 2)
dataset 1: 0.4672:0.949 -> 0.4672:0.949,0.9928:2.0294 Energy (keV)

In [7]: notice(2.3, 6)
dataset 1: 0.4672:0.949,0.9928:2.0294 -> 0.4672:0.949,0.9928:2.0294,2.2776:6.57 Energy (keV)

In [8]: group_counts(5)

In [9]: get_filter()
Out[9]: '0.467200011015:0.963599979877,0.992799997330:2.043999910355,2.263000011444:6.613800048828'

In [10]: group_counts(7)

In [11]: get_filter()
Out[11]: '0.467200011015:0.963599979877,0.978200018406:2.043999910355,2.248399972916:6.745200157166'

In [12]: group_counts(20)

In [13]: get_filter()
Out[13]: '0.437999993563:2.087800025940,2.219199895859:6.759799957275'
```

## 1160 fix

The 4.15.0 build does

```
sherpa In [1]: load_pha("3c273.pi")
read ARF file 3c273.arf
read RMF file 3c273.rmf

sherpa In [2]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.4672:9.8696 Energy (keV)

sherpa In [3]: get_dep(filter=True).shape
Out[3]: (42,)

sherpa In [4]: set_grouping(np.ones(1024))

sherpa In [5]: get_dep(filter=True).shape
DataErr: size mismatch between mask and data array: 46 vs 1024

sherpa In [6]: get_filter()
Out[6]: '0.043800000101:0.657000005245,0.700800001621:0.715399980545,0.817600011826:0.846800029278,8.526399612427:8.541000366211,8.643199920654:8.730799674988,8.760000228882:8.847599983215,8.891400337219:8.964400291443,9.095800399780:9.125000000000,9.227199554443:9.256400108337,9.694399833679:9.781999588013,9.811200141907:9.825799942017,9.928000450134:10.015600204468,10.044799804688:10.132399559021,10.161600112915:10.249199867249,10.278400421143:10.366000175476,10.512000083923:10.599599838257,10.803999900818:10.818599700928,10.979200363159:10.993800163269,11.154399871826:11.168999671936'
```

whereas this PR does

```

In [2]: load_pha("sherpa-test-data/sherpatest/3c273.pi")

In [3]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.4672:9.8696 Energy (keV)

In [4]: get_dep(filter=True).shape
Out[4]: (42,)

In [5]: set_grouping([1] * 1024)

In [6]: get_dep(filter=True).shape
Out[6]: (644,)

In [7]: get_filter()
Out[7]: '0.467200011015:9.869600296021'
```

### DIFF
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -1525,7 +1525,6 @@ def test_pha_grouping_changed_no_filter_1160(make_test_pha):
     assert d4 == pytest.approx([1, 2, 3])
 
 
-@pytest.mark.xfail
 def test_pha_grouping_changed_filter_1160(make_test_pha):
     """What happens when the grouping is changed?
 
@@ -1548,8 +1547,6 @@ def test_pha_grouping_changed_filter_1160(make_test_pha):
     d3 = pha.get_dep(filter=True)
     assert d3 == pytest.approx([2, 0, 3])
 
-    # This currently raises a DataErr due to
-    # 'size mismatch between mask and data array'
     pha.grouping = [1, 1, -1, 1]
     d4 = pha.get_dep(filter=True)
     assert d4 == pytest.approx([2, 3])
@@ -1581,7 +1578,6 @@ def test_pha_grouping_changed_1160_grped_no_filter(make_grouped_pha):
     assert pha.get_filter() == ofilter
 
 
-@pytest.mark.xfail
 def test_pha_grouping_changed_1160_grped_with_filter(make_grouped_pha):
     """Test based on work on #1160
 
@@ -1610,15 +1606,16 @@ def test_pha_grouping_changed_1160_grped_with_filter(make_grouped_pha):
     assert pha.get_dep(filter=False) == pytest.approx([1, 2, 0, 3, 12])
     assert pha.get_dep(filter=True) == pytest.approx([3])
 
-    # Change the grouping
+    # Change the grouping; it would be nice if it could have
+    # recognized the requested range was > 1 and <= 5 but the current
+    # code does not support this.
+    #
     pha.grouping = [1] * 5
 
     assert pha.get_dep(filter=False) == pytest.approx([1, 2, 0, 3, 12])
+    assert pha.get_dep(filter=True) == pytest.approx([3])
 
-    # This fails with DataErr: size mismatch between mask and data array: 2 vs 4
-    assert pha.get_dep(filter=True) == pytest.approx([2, 0, 3])
-
-    assert pha.get_filter() == "2:4"
+    assert pha.get_filter() == "4"
 
 
 def test_pha_grouping_changed_1160_ungrped_with_filter(make_grouped_pha):
@@ -1655,7 +1652,6 @@ def test_pha_grouping_changed_1160_ungrped_with_filter(make_grouped_pha):
     assert pha.get_filter() == "4"
 
 
-@pytest.mark.xfail
 @requires_fits
 @requires_data
 def test_1160(make_data_path):
@@ -1683,7 +1679,6 @@ def test_1160(make_data_path):
 
     pha.grouping = [1] * 1024
     assert pha.get_dep(filter=False).shape == (1024, )
-    # This fails with a DataErr: size mismatch between mask and data array
     assert pha.get_dep(filter=True).shape == (418, )
     assert pha.mask.shape == (1024, )
     assert pha.mask.sum() == 418
--- a/sherpa/astro/utils/src/_utils.cc
+++ b/sherpa/astro/utils/src/_utils.cc
@@ -1,5 +1,5 @@
 //
-//  Copyright (C) 2009, 2017, 2021
+//  Copyright (C) 2009, 2017, 2021, 2022
 //  Smithsonian Astrophysical Observatory
 //
 //
@@ -338,23 +338,69 @@ namespace sherpa { namespace astro { namespace utils {
 			    &group ) )
       return NULL;
 
-    if( mask.get_size() == 0 ) {
-      PyErr_SetString( PyExc_TypeError,
-		       (char*)"mask array has no elements" );
+    if (mask.get_ndim() != 1) {
+      PyErr_SetString( PyExc_ValueError, (char*)"mask array must be 1D" );
+      return NULL;
+    }
+
+    if (group.get_ndim() != 1) {
+      PyErr_SetString( PyExc_ValueError, (char*)"group array must be 1D" );
+      return NULL;
+    }
+
+    // Use TypeError for the size=0 checks as this is what historically been
+    // used for this check (when only mask_size was checked).
+    //
+    npy_intp mask_size = mask.get_size();
+    npy_intp group_size = group.get_size();
+    if( mask_size == 0 ) {
+      PyErr_SetString( PyExc_TypeError, (char*)"mask array has no elements" );
+      return NULL;
+    }
+
+    if( group_size == 0 ) {
+      PyErr_SetString( PyExc_TypeError, (char*)"group array has no elements" );
+      return NULL;
+    }
+
+    if (group[0] < 0) {
+      PyErr_SetString( PyExc_ValueError, (char*)"The first element of group is negative" );
       return NULL;
     }
 
     if( EXIT_SUCCESS != res.zeros( group.get_ndim(), group.get_dims() ))
       return NULL;
 
-    npy_intp size = group.get_size();
     npy_intp jj = 0;
 
-    for( npy_intp ii = 0; ii < size; ++ii ) {
-      if( group[ ii ] >= 0 && ii != 0 )
+    // The group array starts a group with a value >= 0 and continues the
+    // group with a value < 0. The first element therefore starts a new
+    // group.
+    //
+    // The assumption is that the difference, in time, to just setting
+    //     res[ii] = mask[jj]
+    // is insignificant.
+    //
+    if( mask[jj] )
+      res[0] = true;
+
+    for( npy_intp ii = 1; ii < group_size; ++ii ) {
+      // Are we starting a new group?
+      //
+      if( group[ ii ] >= 0 ) {
 	++jj;
+	if (jj >= mask_size) {
+	  PyErr_SetString( PyExc_ValueError, (char*) "More groups than mask elements" );
+	  return NULL;
+	}
+      }
       if( mask[jj] )
 	res[ii] = true;
+    }
+
+    if (jj != (mask_size - 1)) {
+      PyErr_SetString( PyExc_ValueError, (char*) "More mask elements than groups" );
+      return NULL;
     }
 
     return Py_BuildValue( (char*)"N", res.return_new_ref() );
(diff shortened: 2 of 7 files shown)


DougBurke:
Aha. I had accidentally been running the tests without XSPEC built so it didn't catch the changed test behavior.

DougBurke:
So, this failure is likely related to #1160. We get, with this code

```
>>> ui.load_pha("3c273.pi")
read ARF file 3c273.arf
read RMF file 3c273.rmf
>>> ui.group_counts(10)
>>> ui.ignore(hi=0.5); ui.ignore(lo=7)
>>> ui.get_bkg().grouped
True
>>> (ui.get_bkg().grouping == ui.get_data().grouping).all()
False
>>> grps = ui.get_grouping()
>>> ui.set_grouping(grps, bkg_id=1)
>>> ui.ungroup(bkg_id=1)
  File "<string>", line 1, in ungroup
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/ui/utils.py", line 7867, in ungroup
    data.ungroup()
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/data.py", line 4316, in ungroup
    self.grouped = False
  File "/home/dburke/sherpa/sherpa-main/sherpa/utils/__init__.py", line 171, in __setattr__
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/data.py", line 1451, in _set_grouped
    old_filter = self.get_filter(group=val)
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/data.py", line 3973, in get_filter
    chans = self.get_noticed_channels()
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/data.py", line 3826, in get_noticed_channels
    mask = self.get_mask()
  File "/home/dburke/sherpa/sherpa-main/sherpa/astro/data.py", line 3865, in get_mask
    return expand_grouped_mask(self.mask, groups)
ValueError: More groups than mask elements
```

and this happens because at this point self.mask has 18 elements but the groups array has 65 groups (which I'm pretty sure is #1160). While you can expand the 18-element mask by 65 groups, it requires reading uninitialized memory.

So I may need to fix #1160 before we can make this change.

anetasie:
@DougBurke this is a good check. 

I'm not sure that this is the right place for the discussion of grouping the background, as we have a separate Google Doc with the written requirements for grouping/filtering of the PHA files.
https://docs.google.com/document/d/1PqScMS4teD04vbHbiDHK271Naarq81DbwOouLIgZhz4/edit?usp=sharing

Both the new GDoc document and the old requirement documents the background data is always ungroup and in the case of background subtraction the grouping from the source is applied to the background data. See the last page of this old document:
https://drive.google.com/open?id=1hfEk3qTR_KJQkO6UXXkExWjC7ca7yCTQ

For background fitting only one would want to potentially group the background in a different way than the source data.

The example in this PR shows the data set which has both the source and background data grouped outside sherpa, and both src/bkg have the same groups. However, in 4.14. and in this PRs `group_counts(10)` is applied to both the source and the background resulting in the bin edges, which don't match between src/bkg (see plot).  I realized that the code you are validating is needed for the `subtract()`. It seems that 'ungroup' in this PR has a different result in `ungroup` in 4.14.

These two figures show the original grouping and the modified grouping with `group_counts(10)`. These are the same in this PR and 4.14.

input_group
new-group-10

Interestingly when I ungroup the background then  I'm no longer able to change the group setting for the background with `set_grouping()` and this is for unfiltered data.

DougBurke:
@anetasie - can you make a separate issue for https://github.com/sherpa/sherpa/pull/1486#issuecomment-1252884568 as it's unrelated to this and it will just get missed otherwise. I'd also suggest trying it with the latest version of Sherpa (e.g. CIAOX / 4.15.0),mainly to check out how the messages created by `notice`/`ignore` work out when messing around with the background.

DougBurke:
Rebsed as #1477 (ad a bunch of other changes) have been merged. It is still expected to fail as #1160 is still open

DougBurke:
See my branch https://github.com/DougBurke/sherpa/tree/fix-1160 which doesn't really fix 1160 at the moment, but does show (for today, at least) a probable example of the calling error that this PR is meant to fix.

DougBurke:
Just to note that this is ready for review and we have other PRs that build on this (#1637 and  #1638)

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -1548,13 +1547,145 @@ def test_pha_grouping_changed_filter_1160(make_test_pha):
...
+    pha.notice(0.5, 6)
+    assert pha.get_dep(filter=False).shape == (1024, )
+    assert pha.get_dep(filter=True).shape == (41, )
+    assert pha.mask.shape == (46, )
+    assert pha.mask.sum() == 41
+    assert pha.get_filter(format="%.2f") == fexpr
+
+    pha.grouping = [1] * 1024
+    assert pha.get_dep(filter=False).shape == (1024, )
+    # This fails with a DataErr: size mismatch between mask and data array
```
hamogu, line 1682: What is "this" that fails? The test passes. Do you mean "should fail"?

hamogu (commented):
I've looked at that Python changes (mostly tests) and skimmed over the C++ changes, looks good to me, but needs a look fomr someone who actually knows C++ (@wmclaugh ?).

This PR is a step forward and should be merged (assuming the C++ changes are OK). All what I'm saying below is beyond the scope of this PR, but are separate points (related to this PR) that we should cinsider in the future:

- The fact that we don't carry around the *intended* filter range, but just have the current filter means (before and after this PR) that repeatedly changing the grouping dilutes the filter expression. Every time the grouping changes, the boundaries of the filter move to the point that filtered regions are included again. That means that users who experiment with different groupging schemes in practice still have to reset the filter by hand after every grouping. The only way around that is to track the fiter ranges that the user put in (@DougBurke has elaborated on the problems that brings elsewhere) and I believe we will eventually end up doing that, but not now and not here.
- As @DougBurke said above, I believe that all the grouping and filtering would be simpler, if it was done in Python, so we would not have to drop to C++ repeatedly, docstrings carry around naturally, etc. It would also easier to review (for me at least), eaiser to build (at least those parts), and to extend by users and again I believe we will eventually end up doing that, but not now and not here. Grouping and filtering is done only a few times per Sherpa session and is not time critical and I predict it's also not going to be significantly slower in Python. It's not broken, so let's not change it unless there is a real need, but keep those thoughts in mind if we ever need to redesign the appraoch (e.g. to carry the intended filter ranges).

sherpa/astro/tests/test_astro_data2.py
DougBurke: It could be a line that should have been removed. I'll have to review.

DougBurke: This test is marked xfail in cd2569d2 and so the text just hasn't been updated when the code was fixed.

DougBurke: Oops - that's the wrong commit. It should be the second commit added as part of this PR.

hamogu (approved):

hamogu:
It's good on the Python side, so I'm approving, but I request @wmclaugh to look at the C++ code (please re-assign to whoever is appropriate).

sherpa/astro/utils/src/_utils.cc
```diff
@@ -338,25 +338,71 @@ namespace sherpa { namespace astro { namespace utils {
...
-      if( group[ ii ] >= 0 && ii != 0 )
+    // The group array starts a group with a value >= 0 and continues the
+    // group with a value < 0. The first element therefore starts a new
+    // group.
+    //
+    // The assumption is that the difference, in time, to just setting
+    //     res[ii] = mask[jj]
+    // is insignificant.
+    //
+    if( mask[jj] )
```
wmclaugh, line 384: While this is correct, it is safer to explicitly   if (mask[0])    here to aid in readability and to avoid accidentally introducing unintended behavior if code is added above this point which changes jj.

DougBurke: Can you suggest what should be said or done here, as I'm not sure what you want? I have to admit that as I wrote this text last April I've forgotten exactly what I was thinking with these comments, so exactly what's meant to happen when the first mask value is < 0 (I claim that the data is then invalid, but we still may have to deal with it), or exactly what the comment about timing is really meant to say is now unclear.

sherpa/astro/utils/src/_utils.cc
```diff
@@ -338,25 +338,71 @@ namespace sherpa { namespace astro { namespace utils {
...
+      res[0] = true;
+
+    for( npy_intp ii = 1; ii < group_size; ++ii ) {
+      // Are we starting a new group?
+      //
+      if( group[ ii ] >= 0 ) {
```
wmclaugh, line 390: You can take advantage of compound conditional short circuiting here-
        if ((group[ii] >= 0) && (++jj >= mask_size) {
          PyErr_SetString( PyExc_ValueError, (char*) "More groups than mask elements" );
	  return NULL;
        }

sherpa/astro/utils/src/_utils.cc
```diff
@@ -513,53 +559,70 @@ static PyMethodDef UtilsFcts[] = {
...
+	      "    -1 continues the group. See [OGIP92_007]_ and [OGIP92_007a]_.\n\n"
+	      "Returns\n"
+	      "-------\n"
+	      "full_mask : array\n"
+	      "    The mask data (booleans) expanded to match the size of group.\n\n"
+	      "Examples\n"
+	      "--------\n\n"
+	      ">>> mask = [True, False, False, True]\n"
+	      ">>> group = [1, -1, 1, 1, -1, 1, -1, -1]\n"
+	      ">>> expand_grouped_data(mask, group)\n"
```
wmclaugh, line 620:
```suggestion
	      ">>> expand_grouped_mask(mask, group)\n"
```

sherpa/astro/utils/src/_utils.cc
DougBurke: I personally find this much harder to read, as it relies on the `++jj` logic firing before the check. I know it's a standard C/C++ idiom, but I do feel like we are combining two things: the ++jj step which is meant to happen, and then a safety check.

sherpa/astro/utils/src/_utils.cc
DougBurke: I'm going to assume this is just about the fact I'd stupidly used `jj` as the argument rather than `0` (and has now been addressed).

sherpa/astro/utils/src/_utils.cc
```diff
@@ -338,25 +338,71 @@ namespace sherpa { namespace astro { namespace utils {
...
+      PyErr_SetString( PyExc_ValueError, (char*)"mask array must be 1D" );
+      return NULL;
+    }
+
+    if (group.get_ndim() != 1) {
+      PyErr_SetString( PyExc_ValueError, (char*)"group array must be 1D" );
+      return NULL;
+    }
+
+    // Use TypeError for the size=0 checks as this is what historically been
```
wmclaugh, line 351:
```suggestion
    // Use TypeError for the size=0 checks as this is what has historically been
```

DougBurke:
@wmclaugh - I did a quick check. I've added notes to the two relevant methods to say that the behavior has changed and re-wrote the summary of the PR to make it more-obvious what routines were changed and how this may affect users. 

Note that #1637 - which follows on from this - is also relevant (as it makes it a bit-more obvious to users of the UI code how the filter has changed because of the grouping code).

For the validation errors that are now raised, I decided not to add more documentation because

1. the code now errors out, so users get useful information that the code was wrong
2. most users won't be affected by this as it's all in low-level code and we've (hoepfully) fixed the cases where this happened

PR #1492, opened 2022-04-27
Make tests pass on 3.10

hamogu:
## Summary
One of the tests fails in Python 3.10 because the error message has been changed.

## Details
The value of an AttributeError that gets raised when a user tries to set a property that does not has a setter has been changed in Python 3.10.
Before, it was "can't set attribute", in 3.10 it became "can't set attribute <attribute name>".

I found this problem when testing #1308 and #1474 and always saw differences in behavior between my local testing (in a 3.10 environment) and our CI.

## Minimal example
Compare the following minimal examples that shows the change in error message
```
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:24:02)
[Clang 11.1.0 ] on darwin
>>> class A:
...
...     _f = 5
...
...     @property
...     def f(self):
...         return self._f
...
>>> a = A()
>>> a.f = 2
```
and
```
Python 3.10.1 | packaged by conda-forge | (main, Dec 22 2021, 01:39:07) [Clang 11.1.0 ] on darwin
>>> class A:
...
...     _f = 5
...
...     @property
...     def f(self):
...         return self._f
...
>>> a = A()
>>> a.f = 2
AttributeError: can't set attribute 'f'
```

### DIFF
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -903,7 +903,7 @@ def test_img_can_not_set_coord(make_test_image):
     with pytest.raises(AttributeError) as ae:
         d.coord = "physical"
 
-    assert str(ae.value) == "can't set attribute"
+    assert  "can't set attribute" in str(ae.value)
 
 
 def test_img_set_coord_invalid(make_test_image):


hamogu:
I suggest that we take this very simple change for the current release, making at least our smallest test suite (no extra data) pass on Python 3.10. 
I locally develop on 3.10, that's how I found it. It's guranteed to do no harm and it would be nice to be minimally 3.10 compatible. Since we just had our call today, I've taken the freedom to assign a milestone - but that's only to make sure that we make a conscious decison to include or not include this. I'm totally OK with DS clearing that milestone; I just wanted to set something to make sure we at least look at it before the feature freeze.

wmclaugh (approved):

PR #1490, opened 2022-04-25
Basic support for XSPEC 12.12.1

DougBurke:
# Summary

Allow Sherpa to be built with XSPEC 12.12.1. It does not provide access to the three new models - polconst, pollin, and polpow - added in this release of XSPEC.

# Details

The main change is to allow the code to compile against XSPEC 1212.1, since it changed the interface to the `tabint` routine (by marking a number of fields const and also exporting the definition). It also swaps over three models - `ismabs`, `ismdust`, and `olivineabs` - from FORTRAN `REAL*4` to `REAL*8` interfaces (as these have just moved in this release), which has meant some additions (as these are the first `xsF77Cal`l models we support, and this release of XSPEC doesn't provide the "wrapper functions" for these models, as confirmed to me by HEASARC).

This PR does not add support for the three new multiplicative models (`polconst`, `pollin`, and `polpow`) as they require the ability to set the XFLT values for a file, and this is not supported by Sherpa. We can add support, but this requires bumping the minimum supported version of XSPEC from 12.9.0 to something later (it's not a simple answer as to what version) and I have had a PR doing this for quite a while (e.g. the closed #1239 and the current #1396) which need to be looked at before we can do this. Even after bumping up the minimum version, which will allow us to set, query, and clear the XFLT database, we still need to do some design work to allow us to connect to the XSPEC `spectrumNumber` argument - e.g. #544 - as we don't have that as a concept for models.

PR #1493, opened 2022-04-28
Remove deprecated distutils.version.LooseVersion

DougBurke:
# Summary

The distutils.version.LooseVersion class is now marked as deprecated so remove its use when building Sherpa and when importing the Sherpa XSPEC module.

# Details

This is taken from #1329 but it's simplified. It also requires #1490 as that makes changes to `xspec_config.py` which we take advantage of here (it could be done without #1490 but they change the same bit of code so one has to be done first). An earlier version of this PR required the `packaging` package to be installed, but as mentioned below that just added more complexity than it was worth, so I've re-worked it to avoid the need for it.

The suggested replacement for `distutils.version.LooseVersion` is `packaging.version.Version` but this requires  an extra package be installed both when building Sherpa and when using Sherpa. As the code to support PEP 517 (in particular the code to allow build pre-requisites to be specified) has not been looked at yet (e.g. #1329) we can't easily add the `packaging` package to the Sherpa build without a lot of work documenting it and telling the user. 

Fortunately, the version check we need does not need the full PEP 440 support provided by `Version` and can be handled with a tuple of (major, minor, micro) versions (all integers). So the code is re-written to use this. Now, full XSPEC version strings are actually not PEP 440 complaint as the trailing patch level - such as the "c" in "12.12.0c" - breaks PEP 440, but the changes started in #1490 (to strip this patch-level value) are used to ensure we have a "sensible" check.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -415,28 +415,18 @@ available.
 
    Current version: `helpers/xspec_config.py <https://github.com/sherpa/sherpa/blob/master/helpers/xspec_config.py>`_.
 
-   When adding support for XSPEC 12.11.1, the code in the ``run``
-   method was changed to include the triple ``(12, 11, 1)``::
+   When adding support for XSPEC 12.11.1, the top-level
+   ``SUPPORTED_VERSIONS`` list was changed to include the triple
+   ``(12, 11, 1)``::
 
-       for major, minor, patch in [(12, 9, 0), (12, 9, 1),
-                                   (12, 10, 0), (12, 10, 1),
-                                   (12, 11, 0), (12, 11, 1)]:
-           version = '{}.{}.{}'.format(major, minor, patch)
-           macro = 'XSPEC_{}_{}_{}'.format(major, minor, patch)
-           if xspec_version >= LooseVersion(version):
-               macros += [(macro, None)]
+     SUPPORTED_VERSIONS = [(12, 9, 0), (12, 9, 1),
+                           (12, 10, 0), (12, 10, 1),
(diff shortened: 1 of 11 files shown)


DougBurke:
Right. So because this was pulled out of #1329 where I was moving the build towards the "newer" setup, the current CI runs fail because with `python setup.py develop` we don't seem to have a way to say "hold on, you need `packaging` to be installed". I had hoped that `pyproject.toml` was sufficient, but obviously not. I have completely forgotten all the build stuff as I worked on it a long while ago, so if anyone spots what needs to be done please holler

DougBurke:
I've finished re-working this and it is now a lot easier to handle (ie no need to worry about installing `packaging` as both build-time and run-time dependencies).

Marie-Terrell (approved):
Looks good!

docs/developer/index.rst
```diff
@@ -413,28 +415,18 @@ available.
...
+     SUPPORTED_VERSIONS = [(12, 9, 0), (12, 9, 1),
+                           (12, 10, 0), (12, 10, 1),
+                           (12, 11, 0), (12, 11, 1)]
 
-   The define should be named ``XSPEC_<a>_<b>_<c>`` for XSPEC release
-   ``<a>.<b>.<c>`` (the XSPEC patch level is not included). This define
-   is used when compiling the XSPEC model interface, to select which
-   functions to include.
+   This list is used to select which functions to include when
+   compiling the C++ interfce code. For reference, the defines are
```
wmclaugh, line 427:
```suggestion
   compiling the C++ interface code. For reference, the defines are
```

PR #1494, opened 2022-04-29
Update AX_PYTHON_DEVEL.m4 from serial 17 to 25

Marie-Terrell:
This should hopefully replicate the work from #1316, but with the update of AX_PYTHON_DEVEL.m4 to serial 25. 

This update covers the cleanup of distutils in the configure files for grplib (and stklib in the case of standalone sherpa). Note: There will still be a few references to distutils in the grplib configure file coming from the AM_PATH_PYTHON macro, but these are only there as a fallback if "import sysconfig" does not work. 

Caveat: There was a manual edit here done for serial 17 to comment out the "final check to see if everything compiles alright" section. I have maintained that edit. 

To avoid bringing in any other changes other than this, I have done the autoreconf on an Ubuntu machine using the autoconf, automake, aclocal, libtool, etc. versions mentioned in the last run and left off the force flag (to leave the other files as is). This does add the depcomp listing back to some of the Makefile.in files which should be okay. If I re-run with force, the depcomp is actually removed at the top, so we may want to look into this a bit more next time we update the autotool versions we run against.


DougBurke (commented):
If you need a review this "looks" okay (as far as I understand autotools/conf/make/..., which is not fine). I assume the plan is to rebase this once #1496 is merged, then hopefully this gets a green light on CI?

Marie-Terrell:
Rebased with #1496

DougBurke (approved):

PR #1495, opened 2022-04-29
Note that we support Python 3.10

DougBurke:
# Summary

Note that Sherpa can be built using Python 3.10 and add two Python 3.10 CI test runs.

# Details

As #1316 appears to have disappeared, let's tell people we support Python 3.10. I've picked "relatively minor" CI runs (i.e. not the full build and install with XSPEC style testing) as we need to review what runs to do properly soon.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -27,7 +27,7 @@ Requirements
 
 Sherpa has the following requirements:
 
-* Python 3.7, 3.8, 3.9, or 3.10
+* Python 3.7, 3.8, 3.9, or (as a beta) 3.10
 * NumPy (the exact lower limit has not been determined,
   1.17.0 or later will work, earlier version may work)
 * Linux or OS-X (patches to add Windows support are welcome)
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -27,7 +27,7 @@ Requirements
 
 Sherpa has the following requirements:
 
-* Python 3.7, 3.8, or 3.9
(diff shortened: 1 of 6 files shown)


hamogu:
When are the dropping 3.7 again? If we drop 3.7 after this release (as in "the next standalone release also goes into CIAO 4.15) AND CIAO 4.15 is 3.8+ then I suggest we wait until this is out and then try to bump all versions on CI by 
0.1, just to save us the work of doing one PR now to update CI, just to do it again in two weeks without merging any features in between (Does not mean we can't try as a draft PR now).

DougBurke:
Rebased as #1496 has been merged

Marie-Terrell:
Could there be a comment update to note that this is more in beta support (for instance we won't be having the python 3.10 Conda package for this release)? Also, can you add one Conda and one pip test for Python 3.10? Mortiz's PR was a good check for the Python support, but I'd still like to maintain the numpy 1.19 and Python 3.7 tests for now.

docs/install.rst
```diff
@@ -27,7 +27,7 @@ Requirements
 
 Sherpa has the following requirements:
 
-* Python 3.7, 3.8, or 3.9
+* Python 3.7, 3.8, 3.9, or 3.10
```
DougBurke, line 30:
```suggestion
* Python 3.7, 3.8, 3.9, or (as a beta) 3.10
```

Marie-Terrell (approved):
Looks good to me. Thank you!

PR #1498, opened 2022-05-04
Update all CI tests by +0.1 Python and numpy versions

hamogu:
I don't necessarily intend for this to get merged and certainly not before the upcoming freeze, but I think it's about time to try out if CI passes when we just update all Python and numpy versions. It's a more thorough test than me just trying if 3.10 works on just my single, local, architecture.

Once we know, we can close (or keep around, whatever seems useful).

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -23,7 +23,7 @@ jobs:
         include:
           - name: MacOS Full Build
             os: macos-latest
-            python-version: 3.8
+            python-version: 3.9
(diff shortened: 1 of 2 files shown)


Marie-Terrell:
Thank you! You beat me to it. I'm asking Doug to add two Python 3.10 tests onto #1495 for now as I don't want to drop Python 3.7 or numpy 1.19 testing just yet.

hamogu:
We tried this out and it worked, we add some 3.10 CI in #1495, so no need to keep this open.

PR #1504, opened 2022-05-08
CI: rework setup

DougBurke:
# Summary

Update the GitHub workflows for Continuous Integration to better take advantage of the workflow language.

# Details

Take the improvements from #1329 and #1352 on changing the CI setup. I found these changes necessary to understand the removal of the test support from `setup.py` (that is, the current code made it hard to make the change, but breaking things up makes it easier). There should be no "functional changes" to the tests here, but I claim it makes it easier to see in the workflow how the various options - in particular the `test-data` and `install-type` matrix elements - interact.

The main work is to move logic that is "implicit" in the old scripts out into the workflow language (by moving the logic into the workflows).Or at least it's easier for me to see and understand. I believe we could make further changes, but that should be done in a separate PR when we come to updating the dependencies.

For macOS the XQuartz handling is now done by `brew` - which means we should have less need to update this step ourselves, at the expense of not having an explicit constraint on what version of XQuartz is installed. This change is based on the information at https://stackoverflow.com/questions/63648591/how-to-install-x11-before-testing-with-github-actions-for-macos

There is also a change in the macOS Xvfb handling, done as part of moving the logic from a script to the workflow. I am not 100% convinced on the need to change the PATH environment variable so I may explore removing it and seeing what works or not. This work was done to address some bizarre macOS failures I found in developing #1329 / #1352 that I couldn't adequately explain.


Marie-Terrell:
There is a bit of redundancy here, but I can see that this is just in preparation for further updates in #1329. 
Just for a bit of background on why tests.sh was not combined with the workflow earlier:

- It was quicker to make the initial switch
- Just in case GitHub Actions started adding limitations or adding a cost we would more easily be able to switch to a new software since the code was not made software specific. 

However, it should be able to bring back scripts should we need them. Separating the steps the way you have hear makes it much easier to navigate the log files, which provides a much greater benefit. 

The only thing that might be able to reduce a few lines is to set the default shell to bash for all of the jobs at once rather than for each job. This shouldn't cause any issues for the actions we pull in, but can easily be backed out if it does: https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs#example-setting-default-run-step-options-for-a-job
Once that's in, I'll add my approval here

.github/scripts/setup_ds9.sh
```diff
@@ -0,0 +1,49 @@
...
+  fi
+}
+
+# Tarballs to fetch
+ds9_tar=ds9.${ds9_os}.8.3.tar.gz
+xpa_tar=xpa.${ds9_os}.2.1.20.tar.gz
+
+# Fetch them
+download $ds9_base_url/$ds9_os/$ds9_tar
+download  $ds9_base_url/$ds9_os/$xpa_tar
```
hamogu, line 41:
```suggestion
download $ds9_base_url/$ds9_os/$xpa_tar
```

DougBurke:
@Marie-Terrell - just to explictily check - are you okay with us relying on brew to update XQuartz (it's a balance between simplifying our code versus relying on other code, and I personally have no  interest in the macOS side of the build)?

Marie-Terrell:
Yeah. Brew for xQuartz is great! Much cleaner. I think there was a reason for not being able to do it at the initial switch (some issue/error), but I can't remember and since it works now that's all that matters

DougBurke:
@Marie-Terrell hopefully this is the change you wanted

Marie-Terrell (approved):
Perfect. Thanks!

PR #1505, opened 2022-05-09
RTD use pip/pytest rather than setup.py for install/develop/test

DougBurke:
# Summary

Change the RTD documentation to use pip and pytest rather than setup.py when building and testing Sherpa.

# Details

Another change taken from #1329 and #1352. We can make this documentation change without moving our CI run over to use pytest (I personally use pip/pytest for all my development work and it has been working fine for a long time).

Change the documentation to use pip and pytest rather than setp.py for installing, developing, and testing. I thought about keeping a line that using "python setup.py foo" was still supported but decided that we didn't need that level of complexity in the
documentation.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -436,7 +436,7 @@ available.
 
 #. Attempt to build the XSPEC interface with::
 
-     python setup.py develop
+     pip install -e .
 
    This requires that the ``xspec_config`` section of the ``setup.cfg``
    file has been set up correctly for the new XSPEC release. The exact
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -10,17 +10,23 @@ refresher (or prefer to act first, and read documentation later),
 the following commands can be used to install Sherpa, depending on
 your environment and set up.
 
-::
+#. Using conda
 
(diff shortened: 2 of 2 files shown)


DougBurke:
There's no particular hurry, as I think we'll be holding off on things while we work on the 4.14.1 release. I just wanted to get things off my mind so I can look at #1382 ....

DougBurke:
It may also be worth looking at the existing documentation to see if it can be improved or stream-lined.

hamogu (approved):

DougBurke:
Hmmmm. I need to check the suggestions since I added something like

    pytest sherpa sherpa/test/test_foo.py

but this doesn't seem to work. So, need to review!

hamogu:
I'm not sure which edit you refer to. I see
```
  pytest sherpa/tests/test_data.py
  pytest sherpa/tests/test_data.py::test_data_eval_model
```
and those work for me. 'pytest sherpa sherpa/test/test_foo.py' would not make sense because it would run tests for all of sherpa and then the test_foo again, but I don't see any new line in this PR that calls pytest with two separate locations?

PR #1506, opened 2022-05-10
do not change the XSPEC configuration when Sherpa is being built

DougBurke:
# Summary

Minor simplification to the XSPEC configuration code.

# Details

I've long wondered about this code, and have removed it in #1329, but https://github.com/sherpa/sherpa/pull/1458#pullrequestreview-962108820 made me think about it again. Naively I imagine that removing the package from the build instructions is "not the right thing to do", but I admit to not having a good-enough understanding of the package ecosystem to know. So this is an attempt to see what happens.

I then removed the "adding the package" change in #1329 so added it here too.

### DIFF
--- a/helpers/xspec_config.py
+++ b/helpers/xspec_config.py
@@ -117,17 +117,12 @@ class xspec_config(Command):
         pass
 
     def run(self):
+        if not self.with_xspec:
+            return
+
         package = 'sherpa.astro.xspec'
         dist_packages = self.distribution.packages
         dist_data = self.distribution.package_data
-
-        if not self.with_xspec:
-            if package in dist_packages:
-                dist_packages.remove(package)
-            if package in dist_data:
-                del dist_data[package]
-
-            return
 
         if package not in dist_packages:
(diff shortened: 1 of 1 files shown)


Marie-Terrell:
So, checking over a few things with this change:

- The sdist contents is the same (no change expected, but run just to be safe)
- If you try to import sherpa.astro.xspec when Sherpa wasn't built with xspec (partially undoing #1458):
```
#You now see this: 
ImportError: cannot import name '_xspec' from partially initialized module 'sherpa.astro.xspec' (most likely due to a circular import) 

#Instead of:
ModuleNotFoundError: No module named 'sherpa.astro.xspec'
```

- If you run the tests without building against xspec, you now see the xspec tests again, but they are skipped (undoes the removal from #1458)

DougBurke:
@Marie-Terrell ta for the review. I honestly don't know enough about the packaging side to understand what should or should not be done here, so I'm glad it's been pulled-out into an easy-to-identify chunk!

Issue #1469, opened 2022-03-17
Convert ds9 backend to class 


hamogu:
Discussion in #1468 , but that PR was closed because it's going to be better to treat ds9 the same way as the other backends, which are currently in the process of being converted to classes. 

Opening an issue here to make sure that's tracked on the project board and not forgotten.

PR #1507, opened 2022-05-10
tests: ensure DS9 window is closed after the tests

DougBurke:
# Summary

Ensure that any DS9 process started by the Sherpa tests is closed when the tests are finished.

# Details

An alternative to this is to just install pytest-xvfb (and ensure you have Xvfb installed), but I am sure I have seen some problems with the DS9 window still being active when the Xvsb session ends (in a similar way to the problems #1497 fixes issues with matpliotlib windows).

Ensure that the DS9 imager is closed after the tests. This is a bit heavy handed as any test run - such as running the tests in one file that doesn't require DS9 - will lead to a warning message being created (as we have to try importing the relevant code), but the pytest machinery seems to hide this.

This follows the approach from #1497 for closing down matplotlib windows after the tests are run.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -27,6 +27,7 @@ from numpy import VisibleDeprecationWarning
 
 import pytest
 
+from sherpa.utils.err import RuntimeErr
 from sherpa.utils.testing import get_datadir, set_datadir
 
 try:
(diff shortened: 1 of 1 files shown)


hamogu (approved):
We might have to tweak the details when we tackle #1469 but for now it's good. Certainly does no harm. Worst thing that happens is that we create a few extra warning when testing. So, we should merge, but keep our eye open in future PRs if we notice anything odd.

PR #1510, opened 2022-05-13
Tests: improve test coverage of plot code

DougBurke:
# Summary

Improve the testing coverage for several plot routines.

# Details

This has been pulled out of #1089 in the hope of getting code actually into Sherpa. These are not huge improvements to the test coverage, but do exercise a few under-used code paths.

This will hopefully not interfere with #1382, but I have just made a bunch of changes to tests that could (they attempt to do something that I imagine #1382 will need, but it does mean they could clash).

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016, 2018, 2020, 2021
+#  Copyright (C) 2016, 2018, 2020, 2021, 2022
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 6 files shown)


DougBurke:
@hamogu - I don't know if this conflicts with your #1382 work - in particular the last commit.

hamogu:
If it does conflict, I'll just have to rebase and deal with it. Isn't that what life is all about? I do ask though, to please continue to review #1382 and not wait for that rebase, if it's needed at all.

hamogu (approved):
The main conflict is probably that the `requires_plotting` decorator goes away in #1382 since we'll always have a plotting backend (even if it's a dummy), but some tests might just not work with a dummy and will need a "requires_pylab" or one of the new fixtures added in that PR. I'll see that the next time I rebase - in either case, it's easy to reconcile.

DougBurke:
I am trynig to get to #1382 but in looking at rebasing #1089 I have had "thoughts" (e.g. #1514) which have been taking up my time-reserved-for-plotting.

PR #1513, opened 2022-05-19
Minor UI tweaks

DougBurke:
# Summary

Add validation to the set_xlog/ylog/xlinear/ylinear calls to ensure the argument is a string. Fiix an issue where the save_model, save_source, save_resid, and save_delchi commands from sherpa.ui (but not sherpa.astro.ui) would error out if not given an explicit identifier. Also minor code cleanups to the session/ui code.

# Details

The "headline change" is the set_xlog (and other routines) now explicitly check the argument is a string. However, by code this is a small change. The bugfix is also a small change (and makes sherpa.ui.utils behave more-like sherpa.astro.ui.utils). There have been a few style changes, moving code around and simplifying things slightly. 

I found the bug because I had added tests to check code that was changed in this PR but that coverage claimed was un-checked. Imagine my surprise when it pointed out an actual bug (I claim that as it is only in the sherpa.ui version of the functions it is not a major fix). I also came across issue #1526 but that is a change in behavior rather than an outright bug, so I have not attempted to address this here.

Before this change

```
>>> set_xlog(True)
  File "<string>", line 1, in set_xlog
  File "/home/dburke/miniconda3/envs/ciao414/lib/python3.9/site-packages/sherpa/ui/utils.py", line 12168, in set_xlog
    self._set_plot_item(plottype, 'xlog', True)
  File "/home/dburke/miniconda3/envs/ciao414/lib/python3.9/site-packages/sherpa/ui/utils.py", line 12115, in _set_plot_item
    plottype = plottype.strip().lower()
AttributeError: 'bool' object has no attribute 'strip'
```

we now get

```
>>> set_xlog(True)
  File "<string>", line 1, in set_xlog
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 12319, in set_xlog
    self._set_plot_item(plottype, 'xlog', True)
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 12223, in _set_plot_item
    _check_str_type(plottype, 'plottype')
  File "/home/dburke/sherpa/sherpa-master/sherpa/ui/utils.py", line 77, in _check_str_type
    raise ArgumentTypeErr('badarg', argname, "a string")
sherpa.utils.err.ArgumentTypeErr: 'plottype' must be a string
```

The save_model/... change is that the code used to (this is from a CIAO 4.14 build):

```
>>> from sherpa import ui
>>> ui.load_arrays(1, [1, 2], [2, 3])
>>> ui.set_source(ui.const1d.mdl)
>>> ui.save_model("x.dat", clobber=True)
  File "<string>", line 1, in save_model
  File "/home/dburke/miniconda3/envs/ciao414/lib/python3.9/site-packages/sherpa/ui/utils.py", line 4131, in save_model
    _check_type(filename, string_types, 'filename', 'a string')
  File "/home/dburke/miniconda3/envs/ciao414/lib/python3.9/site-packages/sherpa/ui/utils.py", line 72, in _check_type
    _argument_type_error(argname, argdesc)
  File "/home/dburke/miniconda3/envs/ciao414/lib/python3.9/site-packages/sherpa/ui/utils.py", line 66, in _argument_type_error
    raise ArgumentTypeErr('badarg', argname, argdesc)
sherpa.utils.err.ArgumentTypeErr: 'filename' must be a string
```

and it now works (and is tested to work)

```
>>> from sherpa import ui
>>> ui.load_arrays(1, [1, 2], [2, 3])
>>> ui.set_source(ui.const1d.mdl)
>>> ui.save_model("x.dat", clobber=True)
>>> quit()
% cat x.dat
#X MODEL
1 1
2 1
```

Note that if I had used `from sherpa.astro import ui` then I would have been able to call `save_model` without an identifier; it is only for the sherpa.ui version (which is why our testing had not found it)

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -47,7 +47,7 @@ from sherpa.models import basic
 from sherpa.models.template import create_template_model
 import sherpa.plot
 
-from sherpa.utils.err import DataErr, IdentifierErr, ModelErr, PlotErr
+from sherpa.utils.err import ArgumentTypeErr, DataErr, IdentifierErr, ModelErr, PlotErr
 from sherpa.utils.testing import requires_data, requires_fits, \
     requires_plotting, requires_pylab, requires_xspec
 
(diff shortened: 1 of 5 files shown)


DougBurke:
Rebased as #1510 has been merged

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3734,17 +3734,74 @@ def test_set_plot_opt_with_plot_y():
...
+
+    The logic for what is and isn't supported in calls like set_xlog
+    has not really been documented, so just test smoe possible names
+    which do work. This is a regression test, so values may be
+    added or removed over time.
+    """
+
+    # All we care about here is that the call does not raise an error
+    s = cls()
+    s.set_ylinear(name)
```
hamogu, line 3791: I suggest to text one x and one log, too, so can be done like this. If we go through that many strings, then we can affard this, too!
```suggestion
    s.set_ylinear(name)
    s.set_xlog(name)
```

sherpa/astro/ui/utils.py
```diff
@@ -62,7 +59,7 @@ class Session(sherpa.ui.utils.Session):
     def __init__(self):
 
         self.clean()
-        sherpa.ui.utils.Session.__init__(self)
+        super().__init__()
```
hamogu, line 62: :+1:

sherpa/plot/__init__.py
```diff
@@ -32,7 +32,7 @@
 When this module is first imported, Sherpa tries to import the
 backends installed with Sherpa in the order listed in the
 ``options.plot_pkg`` setting from the ``sherpa.rc`` startup file.
-The first module that imports successfully is set as the active 
+The first module that imports successfully is set as the active
```
hamogu, line 35: Seems like an unrelated change. It's OK, I can rebase and deal with the conflict in #1382since you already wrote this... but it would be easier without it ;-)

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -3734,17 +3734,74 @@ def test_set_plot_opt_with_plot_y():
...
+                                  "flux",
+                                  "bkg_model", "bkg_source", "bkg_fit", "bkg_resid", "bkg_ratio", "bkg_chisqr", "bkg_delchi"])
+def test_set_opt_invalid(cls, name):
+    """Check we error out if called with an invalid option.
+
+    The logic for what is and isn't supported in calls like set_xlog
+    has not really been documented, so just test smoe possible names
+    which do not work. This is a regression test, so values may be
+    added or removed over time.
+    """
```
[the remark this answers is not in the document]
DougBurke: Well, I know some values will hopefully-soon change (either become valid or invalid here), so there is a reason to check a number of these - i.e. they aren't just "random" strings here.

sherpa/plot/__init__.py
DougBurke: I think I had made a change in this file and then backed that change out, but the "make the file not have trailing spaces" logic made this change. I'll remove it.

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: We do have a bunch of other tests that check these routines work, and given how they are written they "should" work, but I can add them as it doesn't really slow things down. WIll do it in a rebase (for reference, the only relevant commands are

    set_xlog
    set_ylog
    set_xlinear
    set_ylinear

sherpa/astro/ui/utils.py
DougBurke: I have read conflicting information on "super considered harmful", but for most of our cases it isn't an issue (although the model class, and some of the plotting classes, do have multiple inheritance to worry about).

DougBurke:
Rebasing to address some issues and add more tests (and fix a problem the tests showed). I am going to make several commits to make sure the CI code is run enough times.

DougBurke:
Rebased again, as #1523 has been merged (which includes a fix that I had), and also added in some tests I had meant to add (based on looking at the coverage checks) but had somehow lost during a rebase/merge.

DougBurke:
So, the changes from before are really in the first three commits: I've added tests that were highlighted as "not covered" by the coverage reports in the original code. In doing so I found issue #1526 which I am not addressing here (as it needs discussion) and a bug in the sherpa.ui code (but not in sherpa.astro.ui, which is why it hadn't been noted before). This is fixed in the third commit.

The remaining commits are as they were before.

I do want to just check the coverage report: there are definitely a couple of lines that are not tested but these are for a "should not happen" case so I am okay with that.

DougBurke:
Hmmm. The fact that the global symbol table is changed makes this not-as-easy-to-work-around as I had originally thought...

DougBurke:
The coverage report at https://github.com/sherpa/sherpa/pull/1513/files now shows the only two untested code paths are the "fall-through" handlers I've added for plot objects trying to access a preference and not having either `histo_prefs` or `plot_prefs`, which is not something that should happen, so I'm happy leaving these two un-tested. It should be ready \for @hamogu once he has time.

hamogu (approved):

PR #1516, opened 2022-05-21
Consolidate plot and set_xlog handling

DougBurke:
# Summary

Use the same labels for the plot and set_xlog/set_ylog/set_xlinear/set_ylinear functions. As a result, a number of names can no-longer be used as an identifier for a dataset (bkg_chisqr, bkg_delchi, bkg_fit, bkg_model, bkg_ratio, bkg_resid, bkg_source). A number of labels are now considered deprecated and their use will trigger a warning message, displaying the label that should be used. The "energy" and "photon" labels can no-longer be used with the set_x/yxxx functions: instead the get_energy_flux_hist and get_photon_flux_hist calls can be used, such as `get_energy_flux_hist(recalc=False).histo_prefs["ylog"] = True`.

# Details

## Background

This replaces #1089 and #1175. One of the discussion issues on #1175 was a comment I'd made about some code being in a transitional state that would be finished off in a later commit: this addresses this by making the image, contour, and plot commands all use the `_<label>_type` dictionary, and to try and unify the meaning of this (between the plot/contour and set_xlog/... set of commands). The image() command is a little bit different, and set_xlog/.. only is only relevant for the plot commands, so we don't have a clean design for what these structures represent (in the sense of "what do the keys mean"), so having a unified design is a bit hard here. The data-specific plot support added in #906 (or, rather, the more-obviously-designed support for data-specific plots, as there was always some level of support for this) is another reason why there's difference between plots, contours, and images.

Overall this has the same "shape" as the previous attempts, but it has teased out the differences between how set_xlog/... and the plot commands work, and unifies them. It is also less expressive than previous attempts, in that it doesn't try to do much follow-on work to take advantage of the plot changes, as it turns out to be not-quite as effective in reducing the code base as I'd hoped.

In the following I describe the changes made to the code, skipping the intermediate steps. I have added fairly extensive notes to the main commits to describe the per-commit changes so you can follow those, and made sure each "important" step ran through CI to show it works.

The first actual code change (the third commit) is technically a separate bug fix for the handling of calls to `source_component` and `model_component` when using `Data1DInt` data, but it is not a significant-enough fix to be pulled out to a separate commit and it fits in well with the changes in this PR (i.e. it needs to be done to make this PR make sense anyway).

## The original code

The first commit of Sherpa to git is https://github.com/sherpa/sherpa/commit/9ccf0e72acb9f85fdd20a8feea16ea774e1288b3 and contains the Sesion `_plot_types` and `_contour_types` dictionaries, with keys being labels like `"data"` and `"compsource"`, and the values are the plot objects stored as fields, such as `self._dataplot`. The `get_data_plot` call was implemented as

```python
        self._prepare_plotobj(id, self._dataplot)
        return self._dataplot
```

and the `_prepare_plotobj` call started with (this is the sherpa/ui version; the sherpa/astro/ui version had it's own copy rather than falling through to this logic):

```python
    def _prepare_plotobj(self, id, plotobj, model=None):
        id = self._fix_id(id)
        if isinstance(plotobj, sherpa.plot.FitPlot):
            plotobj.prepare(self._prepare_plotobj(id, self._dataplot),
                            self._prepare_plotobj(id, self._modelplot))
        elif isinstance(plotobj, sherpa.plot.FitContour):
            plotobj.prepare(self._prepare_plotobj(id, self._datacontour),
                            self._prepare_plotobj(id, self._modelcontour))
        else:
            if(isinstance(plotobj, sherpa.plot.PSFPlot) or
               isinstance(plotobj, sherpa.plot.PSFContour) or
               isinstance(plotobj, sherpa.plot.PSFKernelPlot) or 
               isinstance(plotobj, sherpa.plot.PSFKernelContour)):
                plotobj.prepare(self.get_psf(id), self.get_data(id))
...
```

A similar scheme was used for contours (you can see that `_prepare_plotobj` deals with both plot and contour classes) and images (this had a separate `_prepare_imageobj` call), although there was no `_image_types` structure. The astro session code over-rode the `_prepare_plotobj` call.

## Recent changes

In trying to fix plot issues I found that I could never work out what was being done, as it ended up going through `_prepare_plotobj` and this hid subtle (or not-so-subtle) issues.

In #906 (Oct 2020) I re-worked the plot and contour code so that access to the `plotobj` is now controlled by the relevant `get_xxx_plot` or `get_xxx_contour` call, rather than going via `_prepare_plotobj`. This lead to `get_data_plot` becoming

```python
        plotobj = self._dataplot
        if recalc:
            plotobj.prepare(self.get_data(id), self.get_stat())
        return plotobj
```

although this is a simple case. We have more-complex code - in particular, as pointed out below, the above code has changed, but the point is that it is now in the relevant call rather than hidden away in code that can't be easily over-ridden. This PR added the `_plot_type_names` and `_contour_type_names` dictionaries which I now want to remove, as I have a better understanding of how the various pieces fit together.

in #1177 (June 2021) I added changes to the image code to copy the #906 changes **and** to extend the logic, so that we do not access fields directly from the session object but do so via an "API" (in this case the API is just a dictionary). I note that the image and contour cases are easier than the plot case as we don't have any extra logic: we always use a particular class for a particular image/contour call. In the plot case we care about the data type (Data1DInt or DataPHA) and there are other places where we appear to care about the model type (this is an area I note below as needing more follow-up, e.g. #1530 but it is not to be fixed here).

The `get_data_plot` code (for the ui case) has changed to become (this was the main branch at one point of the recent development of the PR):

https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/ui/utils.py#L10693-L10713

## Changes in this PR

The basic idea of this PR is to follow #1177 and remove the direct access to fields like `self._dataplot` and access them from a dictionary. This would be for both the plot and contour commands. There main initial complication is that the obvious thing to do - which is re-use the existing `_plot_types` and `_component_types` dictionaries - ends up causing issues with the `set_xlog`/../`set_ylinear` set of commands. This is because the keys of `_plot_types` and `_component_types` are used to indicate the set of valid dataset identifiers and what labels are valid as "plot/contour" names in the `plot()` and `contour()` calls. I note that the contour code does not care about the `set_xlog`/... family of commands but it does about the second plot, and that the image code in #1177 didn't have to worry about either case (although I note that we could add an `image` command in #1517 but that is irrelevant here).
 
So, there is some work needed to be done to rationalize the `set_xlog`/... and `plot()` commands so that they use the same labels and underlying data structures. This leads to changes to the set of valid identifiers: at the moment we allow old and new names but the idea is that the old names are deprecated and will be removed in a future release. The user is noted by a screen message whenever they use an old name.

So, the idea is that we now have in `sherpa.ui.utils.Session`

```python
        self._plot_types = {
            'data': [sherpa.plot.DataPlot(), sherpa.plot.DataHistogramPlot()],
            'model': [sherpa.plot.ModelPlot(), sherpa.plot.ModelHistogramPlot()],
            'model_component': [sherpa.plot.ComponentModelPlot(), sherpa.plot.ComponentModelHistogramPlot()],
            'source': [sherpa.plot.SourcePlot(), sherpa.plot.SourceHistogramPlot()],
            'source_component': [sherpa.plot.ComponentSourcePlot(), sherpa.plot.ComponentSourceHistogramPlot()],
            'fit': [sherpa.plot.FitPlot()],
            'resid': [sherpa.plot.ResidPlot()],
            'ratio': [sherpa.plot.RatioPlot()],
            'delchi': [sherpa.plot.DelchiPlot()],
            'chisqr': [sherpa.plot.ChisqrPlot()],
            'psf': [sherpa.plot.PSFPlot()],
            'kernel': [sherpa.plot.PSFKernelPlot()]
        }
```

and we access the data with

```python
    def get_data_plot(self, id=None, recalc=True):
...
        # Allow an answer to be returned if recalc is False and no
        # data has been loaded. However, it's not obvious what the
        # answer should be if recalc=False and the dataset has
        # changed type since get_data_plot was last called.
        #
        if recalc:
            data = self.get_data(id)
        else:
            data = self._get_data(id)

        # This uses the implicit conversion of bool to 0 or 1.
        #
        idx = isinstance(data, sherpa.data.Data1DInt)
        plotobj = self._plot_types["data"][idx]
        if recalc:
            plotobj.prepare(data, self.get_stat())

        return plotobj
```

For the `sherpa.astro.ui.utils.Session` class this gets augmented with

```python
        self._plot_types['data'].append(sherpa.astro.plot.DataPHAPlot())

...

    def get_data_plot(self, id=None, recalc=True):
        if recalc:
            data = self.get_data(id)
        else:
            data = self._get_data(id)

        if isinstance(data, sherpa.astro.data.DataPHA):
            plotobj = self._plot_types["data"][2]
            if recalc:
                plotobj.prepare(data, self.get_stat())

            return plotobj

        return super().get_data_plot(id, recalc=recalc)
```

The logic for matching the data class to the plot object is intentionally low-tech: it's just a set of `isinstance` calls and we assume that those plots that care about this have the `_plot_types` dictionary set up correctly (that is, the value is a list with 2 or 3 elements in the correct order, depending on which module you are in). I had tried fancier approaches but they didn't make the code simpler, and would make it harder to extend in the future (e.g. for #347).

**NOTES**

1. not all plot types care about the data they are sent (perhaps they should, to better catch things like trying to plot a `DataIMG` dataset, but that's a separate issue and it also depends on what we want to do with #347); in this case the `get_xxx_plot` call is simpler and doesn't need to be over-ridden in the astro module;

2. the `fit` and `bkg_fit` plots complicate things since, for PHA data, the model part of the plot behaves differently to a `model` plot (the fit version groups the model, the model version shows ungrouped data) and the way that this is done is subtly-different between the `fit` and `bkg_fit` versions - I do not try to address this here;

3. the `source` plot has some strange behavior with template models, and this behavior depends on whether you are using PHA data or not - I do not try to address this here but instead raised issue #1530 to track it.

## Interaction with `set_xlog` et al and `plot`

The set_xlog/ylog/xlinear/ylinear commands are affected by these changes because there is an undocumented relationship between dictionary keys and the labels you can use with these commands. This PR attempts to clarify and clean this up. These commands take an optional string (defaulting to `"all"`). When it is not `"all"` it must match a key from either the `_plot_types` dictionary, or the `_plot_types_alias` dictionary (new in this PR). The idea is that the `_plot_types` key must match a corresponding `get_<key>_plot` method, which makes it easier to see what the relationship is between

```python
set_xlog("data")
set_ylinear("resid")
set_xlog("model_component")
```

This last call is new in this PR. Previously you would have had to know to say

```python
set_xlog("compmodel")
```

The code now labels these "special" labels as an alias, so that they still work but the user is warned to change their code, and we can hopefully remove them at some point in the future. We do not use the python deprecation warning system because users of the UI layer are unlikely to see these messages (since they helpfully get hidden by default), so we use the warning logger. The new behavior is:
```
>>> set_xlog("compsource")
WARNING: The argument 'compsource' is deprecated and 'source_component' should be used instead
>>> set_xlog("compmodel")
WARNING: The argument 'compmodel' is deprecated and 'model_component' should be used instead
>>> set_xlog("compmodel")
WARNING: The argument 'compmodel' is deprecated and 'model_component' should be used instead
>>> set_xlog("astrodata")
WARNING: The argument 'astrodata' is deprecated and 'data' should be used instead
>>> set_xlog("astromodel")
WARNING: The argument 'astromodel' is deprecated and 'model' should be used instead
>>> set_xlog("bkgfit")
WARNING: The argument 'bkgfit' is deprecated and 'bkg_fit' should be used instead
 ```

I honestly don't know what commands like "astrodata" meant - I think it was a way you could change only the PHA plot command, but this was never documented and thanks to changes I've made to the plotting code they no-longer have a meaning.

Since the `plot` command now uses the same labels, we get the same behavior - e.g.

```
>>> plot("compsource", mdl)
WARNING: The argument 'compsource' is deprecated and 'source_component' should be used instead
```

**NOTE**

- the message gets repeated, rather than only shown once; we could add an internal cache so that the message is only displayed once per session (well, per "clean" call)

Related changes because of this are

1. the `plot` command now takes the "new" labels, so `plot("bkg_fit")` mean "do the plot_bkg_fit plot" when before you needed to know that `plot("bkgfit")` was the magic
4. the old names are still supported (with the warning message)
5. the old code did support some of these aliases, in particular you used to be able to say `plot("model_component")` as well as the "secret" `plot("compsource")`. I'm not sure if this is from something I did in earlier plot changes or was something that was always there. This has not been changed, but I just wanted to point it out
6. the identifiers that are marked as "forbidden" - that is the strings you can not use as the `id` argument in calls, is actually based around the labels that the `plot` and `contour` calls accepts; this makes sense, as you need to know when parsing `plot(arg1, arg2, arg3, ...)` whether `arg2` is an identifier or a new plot. This means that changes to the values that `set_xlog` take changes this. So we have some changes.

As an example of the last point, in CIAO 4.14:

```
sherpa-4.14.0> set_default_id("astrodata")
IdentifierErr: identifier 'astrodata' is a reserved word

sherpa-4.14.0> set_default_id("data")
IdentifierErr: identifier 'data' is a reserved word

sherpa-4.14.0> set_default_id("bkgfit")
IdentifierErr: identifier 'bkgfit' is a reserved word

sherpa-4.14.0> set_default_id("bkg_fit")

sherpa-4.14.0> set_default_id("compsource")
IdentifierErr: identifier 'compsource' is a reserved word

sherpa-4.14.0> set_default_id("source_component")
IdentifierErr: identifier 'source_component' is a reserved word

sherpa-4.14.0> set_default_id("energy")

sherpa-4.14.0> set_default_id("flux")

```

The current behavior is to make both the aliases and the actual commands be reserved:

```
>>> import sys
>>> sys.tracebacklimit = 0
>>> set_default_id("data")
sherpa.utils.err.IdentifierErr: identifier 'data' is a reserved word
>>> set_default_id("astrodata")
sherpa.utils.err.IdentifierErr: identifier 'astrodata' is a reserved word
>>> set_default_id("bkgfit")
sherpa.utils.err.IdentifierErr: identifier 'bkgfit' is a reserved word
>>> set_default_id("bkg_fit")
sherpa.utils.err.IdentifierErr: identifier 'bkg_fit' is a reserved word
>>> set_default_id("compsource")
sherpa.utils.err.IdentifierErr: identifier 'compsource' is a reserved word
>>> set_default_id("source_component")
sherpa.utils.err.IdentifierErr: identifier 'source_component' is a reserved word
>>> set_default_id("energy")
>>> set_default_id("flux")
>>> 
```

It turns out that it should only be the new bkg_xxx labels that are added to the set of invalid labels - that is bkg_chisqr, bkg_delchi, bkg_fit, bkg_model, bkg_ratio, bkg_resid, bkg_source.

## Consequences

One consequence of these changes is that we now have a consistent set of names that are used

- by the set_xlog/... family of commands;
- plot() and contour() to refer to the plot/contour to display;
- and to determine what labels are not allowed to be used as a dataset identifier (because of the confusion with the plot/contour calls if they are used).

## Some special plot commands: energy and photon

Note: these are only present in the astro version.

As part of this clean up, the `"energy"` and `"photon"` labels have been removed from the `_plot_types` dictionary, which means they can no longer be used with `set_xlog`/..., as shown below. This is a regression in functionality, but the meaning of these labels is unclear, and the `plot_energy_flux`/`plot_photon_flux` commands are a bit different to the other calls.

```
>>> import sys
>>> sys.tracebacklimit = 0
>>> set_xlog("energy")
sherpa.utils.err.PlotErr: Plot type 'energy' not found in ['data', 'model', 'source', 'fit', 'resid', 'ratio', 'delchi', 'chisqr', 'psf', 'kernel', 'source_component', 'model_component', 'order', 'arf', 'bkg', 'bkg_model', 'bkg_fit', 'bkg_source', 'bkg_ratio', 'bkg_resid', 'bkg_delchi', 'bkg_chisqr']
```

I note that we do not document the behavior of "energy" and "photon", so it is not like we are changing behavior we told users they could rely on.

This is most-likely to cause a problem for users who have been using the `"all"` label - e.g. `set_ylog()` or `set_ylog("all")` - without knowing it.

To change these plot preferences users will have to call the relevant `get` call and change the setting manually:

```python
get_energy_flux_hist(recalc=False).histo_prefs["ylog"] = True
```

This is rather ungainly, but I don't see it as a remotely-high-priority to fix.

## Some special plot commands: pdf, cdf, trace, scatter, pvalue 

There are other plot_xxx calls that do not use the `_plot_types` dictionary - e.g. plot_pdf  - that are present in both versions (unlike energy/photon which are only in the astro session). These calls generally require multiple arguments to call - rather than something like `plot_data(2)` or `plot("data", 2)` - and so it is no clear how best to go forward with these.

This means that there are several "plot objects" that are still left as direct attributes of the Session class. However, this is always likely the case because we have things like `self._splitplot` which is used for setting up multiple plots, and `self._intproj` which is used by the `int_proj` command, that are not directly related to `plot` or `set_xlog`.

## Left over fields of the Session class

I will mention two that are left in that are used by the `plot` code, for reference and because they have been mentioned above:

1. `self._comptmplsrcplot = sherpa.plot.ComponentTemplateSourcePlot()`

I do not understand what the template plot code is meant to be doing. It looks half-finished to me. This PR changes (I believe) how the template model display works for DataPHA data as I think the logic has now changed. We don't have tests to check this. We also do not seem to handle the source and model cases the same for the template models (ie there's no `ComponentTemplateModelPlot`). I really don't want to worry about this code in this PR as I believe it can be looked at after this has been merged. I argue that the logic of what is going on is more-obvious after this PR.

Since writing this I created #1530 as I feel it's not something to address in this PR

2. `self._bkgmodelplot = sherpa.astro.plot.BkgModelPHAHistogram()`

PHA plots behave "strangely" in calls to `plot_fit` and `plot_bkg_fit` which took me a long time to understand, even after looking at the code. We want the model to be displayed using the grouped data here, whereas `plot_model` and `plot_bkg_model` display the data ungrouped. This has caused me no-end of pain, and it would be nice to have a "better" way of doing this, but this is existing functionality and this PR is not the place to try and address this.

Saying that, we do have a difference to how `plot_fit` and `plot_bkg_fit` do things - the former generates the plot object each time whereas the latter uses the `self._bkgmodelplot`. I think we should remove this (so make `plot_bkg_fit` behave like `plot_fit`), but I haven't made that change yet.

For reference, the `plot_fit` code is (here using the `main` branch at the time of writing)

https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/astro/ui/utils.py#L10440-L10468

and `plot_bkg_fit` uses

https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/astro/ui/utils.py#L10788-L10803 

which shows the different behavior in how the `modelobj` variable is set.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -10321,15 +10321,16 @@ class Session(sherpa.ui.utils.Session):
     ###########################################################################
 
     def get_data_plot(self, id=None, recalc=True):
-        try:
-            d = self.get_data(id)
-        except IdentifierErr:
-            return super().get_data_plot(id, recalc=recalc)
+        if recalc:
+            data = self.get_data(id)
+        else:
+            data = self._get_data(id)
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if isinstance(data, sherpa.astro.data.DataPHA):
             plotobj = self._dataphaplot
             if recalc:
-                plotobj.prepare(d, self.get_stat())
+                plotobj.prepare(data, self.get_stat())
+
             return plotobj
 
         return super().get_data_plot(id, recalc=recalc)
@@ -10337,15 +10338,16 @@ class Session(sherpa.ui.utils.Session):
     get_data_plot.__doc__ = sherpa.ui.utils.Session.get_data_plot.__doc__
 
     def get_model_plot(self, id=None, recalc=True):
-        try:
-            d = self.get_data(id)
-        except IdentifierErr:
-            return super().get_model_plot(id, recalc=recalc)
+        if recalc:
+            data = self.get_data(id)
+        else:
+            data = self._get_data(id)
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if isinstance(data, sherpa.astro.data.DataPHA):
             plotobj = self._modelhisto
             if recalc:
-                plotobj.prepare(d, self.get_model(id), self.get_stat())
+                plotobj.prepare(data, self.get_model(id), self.get_stat())
+
             return plotobj
 
         return super().get_model_plot(id, recalc=recalc)
@@ -10428,17 +10430,16 @@ class Session(sherpa.ui.utils.Session):
 
         """
 
-        try:
-            d = self.get_data(id)
-        except IdentifierErr as ie:
-            if recalc:
-                raise ie
-            d = None
+        if recalc:
+            data = self.get_data(id)
+        else:
+            data = self._get_data(id)
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if isinstance(data, sherpa.astro.data.DataPHA):
             plotobj = self._astrosourceplot
             if recalc:
-                plotobj.prepare(d, self.get_source(id), lo=lo, hi=hi)
+                plotobj.prepare(data, self.get_source(id), lo=lo, hi=hi)
+
             return plotobj
 
         return super().get_source_plot(id, recalc=recalc)
@@ -10449,8 +10450,8 @@ class Session(sherpa.ui.utils.Session):
         if not recalc:
             return plotobj
 
-        d = self.get_data(id)
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        data = self.get_data(id)
+        if isinstance(data, sherpa.astro.data.DataPHA):
 
             dataobj = self.get_data_plot(id, recalc=recalc)
 
@@ -10463,7 +10464,7 @@ class Session(sherpa.ui.utils.Session):
             # no way to get it by API (apart from get_fit_plot).
             #
             modelobj = sherpa.astro.plot.ModelPHAHistogram()
-            modelobj.prepare(d, self.get_model(id),
+            modelobj.prepare(data, self.get_model(id),
                              self.get_stat())
 
             plotobj.prepare(dataobj, modelobj)
@@ -10545,14 +10546,12 @@ class Session(sherpa.ui.utils.Session):
             id, model = model, id
         model = self._check_model(model)
 
-        try:
-            d = self.get_data(id)
-        except IdentifierErr as ie:
-            if recalc:
-                raise ie
-            d = None
+        if recalc:
+            data = self.get_data(id)
+        else:
+            data = self._get_data(id)
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if isinstance(data, sherpa.astro.data.DataPHA):
             plotobj = self._astrocompmdlplot
             if recalc:
                 if not has_pha_response(model):
@@ -10563,7 +10562,8 @@ class Session(sherpa.ui.utils.Session):
                         # no response
                         pass
 
-                plotobj.prepare(d, model, self.get_stat())
+                plotobj.prepare(data, model, self.get_stat())
+
             return plotobj
 
         return super().get_model_component_plot(id, model=model, recalc=recalc)
@@ -10574,17 +10574,16 @@ class Session(sherpa.ui.utils.Session):
             id, model = model, id
         model = self._check_model(model)
 
-        try:
-            d = self.get_data(id)
-        except IdentifierErr as ie:
-            if recalc:
-                raise ie
-            d = None
+        if recalc:
+            data = self.get_data(id)
+        else:
+            data = self._get_data(id)
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if isinstance(data, sherpa.astro.data.DataPHA):
             plotobj = self._astrocompsrcplot
             if recalc:
-                plotobj.prepare(d, model, self.get_stat())
+                plotobj.prepare(data, model, self.get_stat())
+
             return plotobj
 
         return super().get_source_component_plot(id, model=model, recalc=recalc)
@@ -10656,6 +10655,7 @@ class Session(sherpa.ui.utils.Session):
         if recalc:
             plotobj.prepare(self._get_pha_data(id),
                             self.get_model(id), orders=orders)
+
         return plotobj
 
     def get_arf_plot(self, id=None, resp_id=None, recalc=True):
@@ -10862,6 +10862,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_model(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def get_bkg_plot(self, id=None, bkg_id=None, recalc=True):
@@ -10926,6 +10927,7 @@ class Session(sherpa.ui.utils.Session):
         if recalc:
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def get_bkg_source_plot(self, id=None, lo=None, hi=None,
@@ -11018,6 +11020,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_source(id, bkg_id),
                             lo=lo, hi=hi)
+
         return plotobj
 
     def get_bkg_resid_plot(self, id=None, bkg_id=None, recalc=True):
@@ -11075,6 +11078,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_model(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def get_bkg_ratio_plot(self, id=None, bkg_id=None, recalc=True):
@@ -11132,6 +11136,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_model(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def get_bkg_delchi_plot(self, id=None, bkg_id=None, recalc=True):
@@ -11190,6 +11195,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_model(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def get_bkg_chisqr_plot(self, id=None, bkg_id=None, recalc=True):
@@ -11248,6 +11254,7 @@ class Session(sherpa.ui.utils.Session):
             plotobj.prepare(self.get_bkg(id, bkg_id),
                             self.get_bkg_model(id, bkg_id),
                             self.get_stat())
+
         return plotobj
 
     def _prepare_energy_flux_plot(self, plot, lo, hi, id, num, bins,
@@ -11404,13 +11411,16 @@ class Session(sherpa.ui.utils.Session):
         ...                          id=1, otherids=(2, 3, 4))
 
         """
+
+        plotobj = self._energyfluxplot
         if recalc:
-            self._prepare_energy_flux_plot(self._energyfluxplot, lo, hi, id=id,
+            self._prepare_energy_flux_plot(plotobj, lo, hi, id=id,
                                            num=num, bins=bins, correlated=correlated,
                                            scales=scales, model=model,
                                            otherids=otherids, clip=clip,
                                            numcores=numcores, bkg_id=bkg_id)
-        return self._energyfluxplot
+
+        return plotobj
 
     def get_photon_flux_hist(self, lo=None, hi=None, id=None, num=7500, bins=75,
                              correlated=False, numcores=None, bkg_id=None,
@@ -11539,13 +11549,16 @@ class Session(sherpa.ui.utils.Session):
         ...                          id=1, otherids=(2, 3, 4))
 
         """
+
+        plotobj = self._photonfluxplot
         if recalc:
-            self._prepare_photon_flux_plot(self._photonfluxplot, lo, hi, id=id,
+            self._prepare_photon_flux_plot(plotobj, lo, hi, id=id,
                                            num=num, bins=bins, correlated=correlated,
                                            scales=scales, model=model,
                                            otherids=otherids, clip=clip,
                                            numcores=numcores, bkg_id=bkg_id)
-        return self._photonfluxplot
+
+        return plotobj
 
     def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
                  clearwindow=True, **kwargs):
@@ -11685,8 +11698,8 @@ class Session(sherpa.ui.utils.Session):
 
         """
 
-        d = self.get_data(id)
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        data = self.get_data(id)
+        if isinstance(data, sherpa.astro.data.DataPHA):
             # Note: lo/hi arguments mean we can not just rely on superclass
             plotobj = self.get_source_plot(id, lo=lo, hi=hi, recalc=not replot)
             self._plot(plotobj, overplot=overplot, clearwindow=clearwindow,
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -147,24 +147,11 @@ class Session(sherpa.ui.utils.Session):
         self._background_models = {}
         self._background_sources = {}
 
-        self._dataphaplot = sherpa.astro.plot.DataPHAPlot()
-        self._astrosourceplot = sherpa.astro.plot.SourcePlot()
-        self._astrocompsrcplot = sherpa.astro.plot.ComponentSourcePlot()
-        self._astrocompmdlplot = sherpa.astro.plot.ComponentModelPlot()
-        self._modelhisto = sherpa.astro.plot.ModelHistogram()
-        self._bkgmodelhisto = sherpa.astro.plot.BkgModelHistogram()
-
-        # self._bkgdataplot = sherpa.astro.plot.DataPHAPlot()
-        self._bkgdataplot = sherpa.astro.plot.BkgDataPlot()
+        # The fit-model for PHA data does not get stored in a field
+        # (it is created whenever needed), so we should probably do
+        # the same for this case.
+        #
         self._bkgmodelplot = sherpa.astro.plot.BkgModelPHAHistogram()
-        self._bkgfitplot = sherpa.astro.plot.BkgFitPlot()
-        self._bkgchisqrplot = sherpa.astro.plot.BkgChisqrPlot()
-        self._bkgdelchiplot = sherpa.astro.plot.BkgDelchiPlot()
-        self._bkgresidplot = sherpa.astro.plot.BkgResidPlot()
-        self._bkgratioplot = sherpa.astro.plot.BkgRatioPlot()
-        self._bkgsourceplot = sherpa.astro.plot.BkgSourcePlot()
-        self._arfplot = sherpa.astro.plot.ARFPlot()
-        self._orderplot = sherpa.astro.plot.OrderPlot()
 
         self._energyfluxplot = sherpa.astro.plot.EnergyFluxHistogram()
         self._photonfluxplot = sherpa.astro.plot.PhotonFluxHistogram()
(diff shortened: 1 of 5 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -47,10 +47,39 @@
...
+    bkg : sherpa.astro.data.DataPHA instance
+
+    Raises
+    ------
+    IdentifierErr
+        If there is no bkg_id background component in data.
+
+    """
+
+    assert id is not None  # make sure id has been set
```
DougBurke, line 75: Argh, this is a stupid line. It should either be removed or

```suggestion
    assert idval is not None  # make sure id has been set
```

sherpa/astro/ui/utils.py
```diff
@@ -10422,48 +10437,45 @@ def get_source_plot(self, id=None, lo=None, hi=None, recalc=True):
...
     def get_fit_plot(self, id=None, recalc=True):
 
-        plotobj = self._fitplot
+        plotobj, data = self._get_plotobj("fit", id=id, recalc=recalc)
         if not recalc:
             return plotobj
 
-        d = self.get_data(id)
-        if isinstance(d, sherpa.astro.data.DataPHA):
+        if not isinstance(data, sherpa.astro.data.DataPHA):
```
DougBurke, line 10460: Perhaps we should repeat the logic of `get_fit_plot` from the superclass here, as we have done most of the work already?

sherpa/astro/ui/utils.py
```diff
@@ -10314,37 +10330,36 @@ def _get_stat_info(self):
...
 
-        if isinstance(d, sherpa.astro.data.DataPHA):
-            plotobj = self._dataphaplot
+        The current mechanism to chose a specific type of object
+        is not particularly extensible.
+        """
+
+        # This is ugly as we repeat a lot of the logic of
+        # super()._get_plotobj.
+        #
```
DougBurke, line 10342: I've tried - in this PR and in previous attempts - a number of versions of this call and - as it is always unsatisfactory: we either have to repeat the same logic in most calls to this routine (e.g. the try/except block to access `data`), or we can include it here but there are some cases where it is not needed.

I have noted that this is not particularly extensible, but there is a benefit to a simple system (in earlier versions I used more-complicated classes, but they didn't add anything above the `if isinstance` chain of calls used below). One advantage to this simpler system is that it makes it more-obvious when we are adding extra logic (in particular the template-model/source plots).

Mentioned here, rather than below, is the related issue that when I over-ride the `get_foo_plot` call it can lead to the choice of either copying the logic from the superclass or deciding (for non PHA data) to call the superclass and re-run the call to `_get_plotobj`. Note that `get_fit_plot` and `get_model_component_plot`  show the two approaches.

sherpa/astro/ui/utils.py
```diff
@@ -10539,51 +10551,31 @@ def get_model_component_plot(self, id, model=None, recalc=True):
...
-            d = None
-
-        if isinstance(d, sherpa.astro.data.DataPHA):
-            plotobj = self._astrocompsrcplot
-            if recalc:
-                plotobj.prepare(d, model, self.get_stat())
+        # If this is not a PHA dataset then it is simple (we could
+        # call the super-class but we have already done most of the
+        # work).
+        #
```
DougBurke, line 10561: see lat part of https://github.com/sherpa/sherpa/pull/1516/files#r885618482

sherpa/ui/utils.py
```diff
@@ -319,103 +319,93 @@ def clean(self):
...
+        # with only one plot class, such as "fit", or those that
+        # support a version that depends on the data type, such as
+        # "data". A list is used in both cases, with the first element
+        # being the "generic" case and, for those that support it, the
+        # second being the Data1DInt plot class.
+        #
+        # There is an argument to be made to have the "singleton"
+        # plots, such as "fit", storing the object directly, and not
+        # as a single-element list, but it was felt that having a
+        # consistent access pattern was cleaner.
```
DougBurke, line 358: At least, I see the benefit of keeping a uniform representation here (always use a list), but this can be discussed.

Earlier attempts at fixing this actually created a class to use here (e.g. create an instance for the value rather than a list, and this class then had support for the "default" value and adding extra elements),  but I never found it satisfactory.

sherpa/ui/utils.py
```diff
@@ -319,103 +319,93 @@ def clean(self):
...
+            "psf": sherpa.plot.PSFContour(),
+            "kernel": sherpa.plot.PSFKernelContour()
         }
 
+        # This is used by the get_<key>_image calls to access the
+        # relevant image class. The keys are not included in any check
+        # of valid identifiers. unlike the plot and contour cases, as
+        # there is
+        #
+        # - no image() call that acts like plot() or contour();
```
hamogu, line 406: but there will be - so maybe we should include it already to make sure nobody writes code that uses kays we won't allow later?

sherpa/ui/utils.py
```diff
@@ -2288,8 +2340,29 @@ def get_data(self, id=None):
...
+           The data set. If not given then the default
+           identifier is used, as returned by `get_default_id`.
+
+        Returns
+        -------
+        instance : sherpa.data.Data or None
+
+        """
+
+        return self._data.get(self._fix_id(id))
```
DougBurke, line 2361: This is a very-simple routine, and I feel a bit bad about adding it as a method, but it does (I believe) make the downstream code a bit cleared - i.e. we call `self.get_data` if we require the data to exist and `self._get_data` if the data need not exist.

sherpa/ui/utils.py
```diff
@@ -12105,39 +12161,85 @@ def get_kernel_plot(self, id=None, recalc=True):
...
+            get_model_component_plot(mdl)
+            get_model_component_plot(2, mdl)
+
+        are meaningful calls.
+
+        """
+
+        # This is an internal routine so it is not expected to be
+        # called incorrectly, but make sure we catch any such problem.
+        #
```
DougBurke, line 12214: This comment is added to mainly point out why the two `raise ArgumentxxxErr` calls are not covered by tests.

sherpa/ui/utils.py
hamogu: This is not a full review, just some ideas based on what I'm seeing now; I want to make comments before it's all done an too late to change. So, please don't waste your time answering my suggestions if you already tried that, or think it's stupid and won't work. That said, here are some thoughts:

This entire approach was (and still is, although it's better now) pretty convoluted and requires careful maintenance. When new plot types are added, the `_plot_types` need to be updated, there is the `_plot_types` and `_contour_types` and later possible `_image_types` and then a few others that don't really fall in this scheme.

I am wondering if more of this information should live in the Plot classes themselves, e.g.:

```python
class RatioPlot(...):
    # would accept all Data1DInt and DataPA, but not generic Data1D
    # I know that that's not what RatioPlot does, but giving and example
    _accepted_data_types = (Data1DInt, DataPHA)  
    _plot_area = 'data'  # one of: data, contour, image
   _ plot_name = 'Ratio'

    ... real code here ...
```
(Obviously, all names are open to bike shedding.)

That way, information would not have to be maintained separately here, it would be intrinsic to the plot class. I think that's a better place to keep that information.

`_get_plot_type` would then search through all known plot classes

```python
def _get_plotorcountour_type(plotarea, plotname, data):
    for p in plot_type_list:
        if ((p._plot_area == plotarea) and (p._plot_name == plotname) and 
           any([insinstance(data, allowed_class) for allows_class in p._accepted_data_dypes])):
           return p
    finally:
        raise Error
```
The implementation above could be refined to deal with cases were there is more than one match, e.g. one could select the "most specific" of all plot classes (if one of them accepts all `Data1D` and another only `DataPHA`, then one may wnat to select the latter, while my implementation would pick whatever is first in the list).

How do we get the `plot_type_list`? Either that's hand-curated, but a better approach would be to use a metaclass that puts stuff in a registry when it's imported like I do for the plotting backends in #1382. That's just a few lines of code in the `Plot` and `Histogram` base classes (could be a new class in the tree `PLotHistroImageBase` from which `PLot` and `Histrogram` inherit).

hamogu: I just occurs to me that in practice
```python
    _accepted_data_types = (Data1DInt, DataPHA)  
    _plot_area = 'data'  # one of: data, contour, image, other
```
would not be set for every plot type, but at one point in the inheritance tree and then apply to all derived classes with no further code change. For example `Plot` would get the `    _plot_area = 'data' ` and all derived classes would have that property and `Contour` would get `_plot_area = 'contour'` and we only need to pay attention to special cases with multiple inheritance. 

I've overlooked something that needs to be special cased in the current scheme, then maybe it also has to be special-cased in the proposed scheme and we need an area "other".

DougBurke: I have been trying to get this change merged for several years, so I am somewhat opposed to making more large architectural fixes. This PR is itself a rather cut-down version of the latest approach where I had added extra logic.

Currently I do not believe there is any "nice and simple" way to indicate the mapping between data type and object to use - for example

- source plot for DataPHA accepts lo/hi arguments unlike every-other plot [maybe we should add it to all plot types, but that's a separate discussion/issue/PR]
- plot_fit/plot_bkg_fit does not use the underlying "model" case, again only for DataPHA
- whatever is going on with template models

I've spent a lot of time trying to simplify this code and it has never ended in success: one conclusion is in fact automating the mapping between data-type and plot object logic doesn't really help.

sherpa/ui/utils.py
DougBurke: At the moment the keys for contour and image are already included in the plot table, so it turns out that adding the contour keys into the "what is a valid identifier" doesn't currently add anything, but is done partly for documentation and just-in-case something changes.

hamogu (changes requested):
This is making things better. Is it making things perfect? No. For that we might have to abandon the UI altogether... I think this is mostly ready, but I have a few small comments; "I thought about this and decided on the current version because of X" is good enough for an answer in most cases - I just wanted to make sure these are intentional decisions and not an oversight.

sherpa/ui/utils.py
```diff
@@ -1327,14 +1317,77 @@ def _fix_id(self, id):
             return self._default_id
 
         if not self._valid_id(id):
-            raise ArgumentTypeErr('intstr')
+            raise ArgumentTypeErr("intstr")
+
+        if _is_integer(id):
+            return id
```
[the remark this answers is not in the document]
DougBurke: This is mainly a style thing. Before this check we have `id : Union[int, str]` but after it we have `id : str`. 

This means that I can theoretically say

    def _check_plottype(self, id : str) -> nool:

but I needed to add explicit an `assert isinstance(id, str)` statement when using `mypy` in #1525 . However, this is experimental code that isn't going to get merged, and I may have been mis-understanding quite how typeguards work there

sherpa/ui/utils.py
```diff
@@ -12105,39 +12161,85 @@ def get_kernel_plot(self, id=None, recalc=True):
...
+            The call.
+        kwargs
+            The keyword arguments to apply to each plot or contour.
+
+        Notes
+        -----
+        Each group consists of the name of the plot/contour, followed
+        by optional arguments. So
+
+            self._multi_plot(["data", "data", "org"], plotmeth="plot")
```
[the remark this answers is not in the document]
DougBurke: Just to note: this is a very-thin wrapper around `plot` and `contour`. You would have to change how those are called - i.e.

    plot("data", "data", 2, ylog=True)

would need to be changed to whatever you would come up with. The `_multi_plot` call is basically converting the user input into this design - i.e. a sequence of "command name" followed by arguments - so if `plot` and `contour` are changed then we can simplify this. To be honest the saving isn't going to be huge in terms of the number of lines, but there are things you could think of doing to make it easier to name arguments to send - e.g.

    plot(["data", "id", 1], ["data", "id", 2])

but

a) this is not the place
b) I'm not sure it's worth it.

hamogu (approved):

Issue #491, opened 2018-09-15
Removing warnings from the Sphinx build


DougBurke:
At present the Travis Sphinx build ignores warnings generated by Sphinx. This is not ideal, since it would be good if it could fail on warnings. However, the current build generates plenty of warnings which are not easy to remove. These fall into two camps

1. formatting issues from the docstrings in `sherpa.utils.testing.SherpaTestCase`, which are inherited from 

2. complaints about duplicate target names (for footnotes) from docstrings in `sherpa.ui.utils.Session` and `sherpa.astro.ui.utils.Session` which I *believe* are because of the "funky" way we use these symbols (there are class methods which are re-exported as functions from `sherpa[.astro].ui`)

One way to address the first point would be to update the tests using the `SherpaTestCase` class to us pytest instead, so we could remove the class itself. or we could just manually over-ride the problem routines and write out own docstrings.

I have no idea about the second point.

hamogu:
I believe (but I might be wrong) that the second point is real, not just "funky". It's because the references are always called 1,2,3 as in `[1]`. I believe they have to be manually fixed to unique labels like`[Cash1979]` (that's hoping that no paper is cited twice). Sphinx is not made to have a separate reference section in every function with labels 1,2,3; sphinx citations are meant to be global: http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations

DougBurke:
The first point is (finally) addressed in #1205

hamogu:
After some further investigation I can confirm what I said in https://github.com/sherpa/sherpa/issues/491#issuecomment-618732309: Every reference in sphinx is global and thus we cannot use labels like [1], [2], because we simply have too many of them. The intention of Sphinx is more to act like LaTeX, where `\citep{Bouke2013}` will give you the same reference throughout. You *could* use `\citep{1}`, but nobody does because the "1" would the same reference throughout the text. 

That's not what we do - we use [1] as a throwaway reference in out API and there are dozens of functions wit ha reference called `[1]`, at least a dozen of them refer to ds9 and another 10 or so to the OGIP memo. If we want to keep per-function references, we have to be more create with reference names to makes them unique, e.g. with names as "ds9", "DS9", "ref1", "foot1", etc. 
That's also not ideal because it's inconsistent. 

However, even if one does that (I made branch where I tried), that **still** is not enough, because the same docs are  often build in *tow different locations*:
[...]

PR #1519, opened 2022-05-23
Use direkt html links instead of using `[1]` and a reference section

hamogu:
# Summary
This replaces the use of a "References" section with direct URL links
in particular in cases where the reference is just a plain old HTML
website and not a scholarly publication.

# Details
The motivation for this change is to reduce the number of sphinx warnings
caused the the use of non-unique `[1]` for references see #491
for discussion.
I don't have a good solution for citations to real references that are cited many times, but this is a start and reduces the number of sphinx warnings already by almost 50%. At the same time it shortens the docstrings a little (by removing the "references" section), which, IMHO, makes them slightly better, at least not less, readable. In the rendered HTML, I don't need to be able to read the link (I just click on it), so the fact that the full link is not rendered as ASCII in the HTML is an advantage.


hamogu:
Note: something clear went wrong - I did not want to delete an entire file. Will force push again in a few minutes.

hamogu:
Mmmm. I'm experimenting with autopep8 in VSCode, but it's not supposed to touch lines which I did not edit in this PR. Will revert those changes.

sherpa/ui/utils.py
```diff
@@ -5294,7 +5289,7 @@ def list_models(self, show="all"):
         elif show.startswith("1d"):
             select = lambda x: (not x.startswith('xs')) and (not x.endswith("2d"))
         elif show.startswith("2d"):
-            select = lambda x: x.endswith('2d')
+            def select(x): return x.endswith('2d')
```
DougBurke, line 5292: I don't think you meant to do this...

DougBurke:
So, we use the Sherpa docstrings to create the ahelp files. For example, the first command changed is `create_rmf` which is used to generate https://cxc.cfa.harvard.edu/sherpa/ahelp/create_rmf.html. The code used to do this is in https://github.com/cxcsds/docstring_to_ahelp and I'd need to check the output (i.e. what the rst markup looks like for the replacements), as it'll almost-certainly need some tweaks.

The create_rmf example is interesting as the reference there is hard to read in the HTML and a direct link would be better, *but* for various technical reasons [1] it wouldn't actually be possible to make it a link in the ahelp file for this example.

[1] the ahelp DTD currently doesn't support anything other than text in the "table data" block, which would be nice to fix but is not going to get done any time soon, if ever

sherpa/ui/utils.py
DougBurke: Or maybe you did, but it doesn't make sense in this PR.

hamogu: That's black trying to get rid of all lambda functions. It's functionally equivalent, but yes, it was not an intentional change, it's accidentally left over from when I experimented what black would do.

DougBurke:
@hamogu - I explained above why this is not a link in the HTML; the AHELP DTD doesn't support it for tables. If you look at https://cxc.cfa.harvard.edu/sherpa/ahelp/set_coord.html we again do not support a "link", but I think if you had made this change that we could (since we should be able to make links work in paragraphs in ahelp files).

So I think it is potentially a good thing, but just needs me to check some things.

hamogu:
What I meant, was that the `http://cxc.harvard.edu/ciao/ahelp/rmfimg.html` in the References section is not an html link (just a plain text URL). That's outside the table. 

image

Anyway, I think we want the same thing - better links where we can -  so I'll just shut up now.

sherpa/astro/ui/utils.py
```diff
@@ -6817,10 +6808,10 @@ def notice2d(self, val=None):
 
         Notes
         -----
-        The region syntax support is provided by the CIAO region
-        library [1]_, and supports the following shapes (the
-        capitalized parts of the name indicate the minimum length of
-        the name that is supported):
+        The region syntax support is provided by the `CIAO region
+        library <http://cxc.harvard.edu/ciao/ahelp/dmregions.html>`_,
```
wmclaugh, line 6812:
```suggestion
        library <https://cxc.harvard.edu/ciao/ahelp/dmregions.html>`_,
```

sherpa/astro/ui/utils.py
```diff
@@ -6633,16 +6628,12 @@ def set_coord(self, id, coord=None):
         ``x=1.5`` and ``y=0.5`` to ``y=1.5``.
 
         The 'physical' and 'world' coordinate systems rely on FITS
-        World Coordinate System (WCS) standard [1]_. The 'physical'
+        `World Coordinate System (WCS) standard
+        <http://fits.gsfc.nasa.gov/fits_wcs.html>`_. The 'physical'
```
wmclaugh, line 6632:
```suggestion
        <https://fits.gsfc.nasa.gov/fits_wcs.html>`_. The 'physical'
```

sherpa/astro/ui/utils.py
```diff
@@ -5307,7 +5307,8 @@ def create_rmf(rmflo, rmfhi, startchan=1, e_min=None, e_max=None,
             If None then a "perfect" RMF is generated, otherwise it gives
             the name of the two-dimensional image file which stores the
             response information (the format of this file matches that
-            created by the CIAO tool rmfimg [1]_).
+            created by the CIAO tool
+            `rmfimg <http://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
```
wmclaugh, line 5311:
```suggestion
            `rmfimg <https://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
```

sherpa/astro/ui/utils.py
```diff
@@ -13618,7 +13600,8 @@ def eqwidth(self, src, combo, id=None, lo=None, hi=None, bkg_id=None,
                 covar_matrix=None):
         """Calculate the equivalent width of an emission or absorption line.
 
-        The equivalent width [1]_ is calculated in the selected units
+        The `equivalent width <http://en.wikipedia.org/wiki/Equivalent_width>`_
```
wmclaugh, line 13603:
```suggestion
        The `equivalent width <https://en.wikipedia.org/wiki/Equivalent_width>`_
```

hamogu: Thanks for checking @wmclaugh , but before I address those, I'll wait and see if we can take something like this PR at all or if it messes up the generation of ahelp files too much that @DougBurke does; there is no hurry either way.

sherpa/ui/utils.py
```diff
@@ -16858,13 +16796,8 @@ def image_xpaget(self, arg):
...
         -----
-        The XPA access point [1]_ of the ds9 image viewer lets
-        commands and queries to be sent to the viewer.
-
-        References
-        ----------
-
-        .. [1] http://ds9.si.edu/ref/xpa.html
+        The `XPA access point <http://ds9.si.edu/ref/xpa.html>`_ of the ds9 image
+        viewer lets commands and queries to be sent to the viewer.
```
DougBurke, line 16800:
```suggestion
        The XPA access point of the ds9 image viewer lets commands and
        queries to be sent to the viewer.
```

Turns out this link is now a 404. I've asked Bill if there's a valid link, but we can also just drop it here.

sherpa/ui/utils.py
```diff
@@ -16908,13 +16841,9 @@ def image_xpaset(self, arg, data=None):
 
         Notes
         -----
-        The XPA access point [1]_ of the ds9 image viewer lets
-        commands and queries to be sent to the viewer.
-
-        References
-        ----------
+        The `XPA access point <http://ds9.si.edu/ref/xpa.html>`_ of the ds9
+        image viewer lets commands and queries to be sent to the viewer.
```
DougBurke, line 16845:
```suggestion
        The XPA access point of the ds9 image viewer lets commands and
        queries to be sent to the viewer.
```

Ditto.

hamogu:
@DougBurke Fixed remaining links and squashed commits. Should be good to go.

sherpa/astro/ui/utils.py
```diff
@@ -8551,20 +8539,14 @@ def subtract(self, id=None):
 
         where src_exposure and bg_exposure are the source and
         background exposure times, and src_backscal and bg_backscal
-        are the source and background backscales.  The backscale, read
-        from the ``BACKSCAL`` header keyword of the PHA file [2]_, is
-        the ratio of data extraction area to total detector area.
+        are the source and background backscales.  The backscale, `read
+        from the ``BACKSCAL`` header keyword of the PHA file
+        <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/node5.html>`_,
+        is the ratio of data extraction area to total detector area.
```
DougBurke, line 8546: This isn't rendering nicely

screenshot

I assume it's the fact we have

    blah blah `link text starts but look we have ``SOME KEYWORD`` and more text <actual link>`_ blah bAH

so it's the fact that 

    ``BACKSAL``

is now within a 

    `blah blah`_ 

section.

sherpa/astro/ui/utils.py
```diff
@@ -8551,20 +8539,14 @@ def subtract(self, id=None):
...
```
hamogu, line 8545:
```suggestion
        are the source and background backscales.  The backscale, read
        from the ``BACKSCAL`` header keyword of the `PHA file
        <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/node5.html>`_,
        is the ratio of data extraction area to total detector area.
```

hamogu:
I'll just shorten the part that gets turned into a link. Other tricks are possible, but simple is better than ugly - we also process these strings downstream and we don't really want to monkey with escape sequences etc.

Added commit and let CI run again so you can see the readthedocs build.

sherpa/astro/ui/utils.py
DougBurke: screenshot

Yay....

PR #1521, opened 2022-05-23
Doctestplus

hamogu:
# Summary
Allow doctests to run in Sherpa

# Details
In the first commit, I activate doctests for all of Sherpa. That's going to fail, however, I let it run on CI anyway, just so that you can see how many tests we have and what typical failures are (from the common, trivial on the top, to the rare, but worthy of fixing at the bottom):

- formatting of code
- numpy not imported
- slightly different formatting of of the output
- example contains typos
- example uses variable that are not defined
- example uses functionality that's buggy or outdated.

The far goal is to run doctests for all rst files and all embedded examples of the docs (#816). This would ensure that the examples we give execute at all, and, where we have expected output to compare to, also act as an additional tests. We have about 1000 doctests embedded in our rst docs, so running all those automatically is a significant enhancement in itself.

This is an old issue (#816), the reason I'm looking at this now is is my work in #1382, where I want to ensure that all examples actually work correctly. 

# How to continue?
I fixed up four files that now pass the doctests, which you can locally verify as

```
pytest --doctest-plus /Users/guenther/code/sherpa/sherpa/astro/data.py
pytest --doctest-plus /Users/guenther/code/sherpa/sherpa/data.py
pytest --doctest-rst docs/quick.rst
pytest --doctest-rst docs/evaluation/combine.rst
```

**EDIT**: Updated to current status


.github/scripts/test.sh
```diff
@@ -40,12 +40,15 @@ fi
...
+if [ -n "${XSPECVER}" ] && [ ${TEST} != none ]; then DOCTEST="--doctest-rst"; fi
+
 # Run regression tests using sherpa_test
 if [ ${TEST} == package ] || [ ${TEST} == none ]; then
     cd $HOME;
     conda install -yq pytest-cov;
     
     # This automatically picks up the sherpatest module when TEST==package
-    sherpa_test --cov sherpa --cov-report xml || exit 1;
+    sherpa_test $DOCTEST --cov sherpa --cov-report xml || exit 1;
```
hamogu, line 52: @DougBurke Can I ask for your help? For simplicity, I don't want to run the doctests every time, it's simpler if we run them only when both XSPEC and the TESTDATA are present. In that case, I want to pass the "--doctest-rst" to pytest. On the command line, that's simple: `pytest --doctest-rst` will be sufficient to run the doctests in the modules that I have not blacklisted. However, how do I do that in CI? We seem to have XSPEC only in conda installations and in those we don't run pytest directly, we run `sherpa_test` instead. Is there an easy way to apss through the argument?

Alternatively, I could skip anything that uses XSPEC models, but it would be nice to test those parts, too. See files changes in this PR - I discovered an example where our naming was incocnsistent already.

DougBurke: So, I've spent a lot of time in #1329 / #1352 to try and modernize our CI code, where by "modernize" I really mean "clean up as we no-longer need to follow the Travis setup". From that work I can claim that I do not understand the naming scheme our tests use (e.g. the `TEST == package` vs ... options) and there do seem to be some ways you can set up the tests to be less-than-optimal (e.g. see #1504).

This is all to say that I hope our CI runs will be changed significantly soon, but this means awkwardness for any code that wants to change the behavior in the mean time, and requires the existing PRs to be reviewed.

One thing that may help in the short term is that `sherpa_test` is essentially a wrapper around `pytest` - see https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/__init__.py#L871-L883 - and in #1001 I added support to allow arguments to be passed through (which is why you can add `--cov ...` as shown above, if memory serves), but there are parts of `pytest` that I don't understand so although I'd hope you can just pass through the argument, I can't guarantee it'll work.

For background: we only have XSPEC available in conda because trying to get XSPEC-enabled pip installs would, I predict, be too painful. With the conda build we can pull in our test build of the XSPEC model library via conda and our channel, but for pip I don't know what you'd do. I have spent time looking at separating out the low-level interface to XSPEC models, which could then be a dependency of Sherpa, but my attempts to get HEACIT people to look at this have been less-than successful and we are no well-beyond the scope of this PR.

hamogu:
Now I get a conda error that I can't reproduce locally. Based on some google which brought me to https://github.com/conda/conda/issues/9672 I think might might be due to the fact that Python is first installed in version 3.9.12 and then downgraded to 3.8.13  and conda itself is first installed as 4.12 and then upgraded to 4.13 later in the script and thus the version 4.12 is not found any longer:
```
miniconda/pkgs/conda-4.12.0-py39h06a4308_0/info/test/tests/__init__.py:8: in <module>
    from conda.gateways.logging import initialize_logging
E   ModuleNotFoundError: No module named 'conda'
```
Now, why is it looking for 4.12 in the first place? I don't know and I don't think that this is the place to debug that.

reference: https://github.com/sherpa/sherpa/runs/6759776547

@Marie-Terrell @harlantc : Do you have an idea when you might be able to look into the CI work by @DougBurke ? If that's happening soon, I'll just wait. If not, I'll split these changes into separate parts of "improve docs and run doctestplus locally" and "run doctestplus on CI"; where the latter can wait until the https://github.com/sherpa/sherpa/pull/1329 and https://github.com/sherpa/sherpa/pull/1352 are in, but I don't want to wait forever for the former, because that will just lead to needless rebasing.

hamogu:
I could split out the CI changes into a separate PR. This PR is useful in it's own right as it (a) improves some examples in the documentation and (b) allows to run the doctests locally. That means that the review would have to be done locally and can't rely on CI, but maybe that's a way forward so that I'm not blocked by waiting for the CI work?

hamogu:
Note to self: As discussed in the developer meeting:
- remove CI aspects, so that PR is only about doc changes
- add docs for developers on how to run doctestolus locally
- maybe open new issue to add to CI at a later time

docs/conf.py
```diff
@@ -157,6 +157,7 @@ def get_xsversion(self):
     # notebooks
     'nbsphinx',
     'sphinx_rtd_theme',
+    'pytest_doctestplus.sphinx.doctestplus',
```
hamogu, line 160: No new requirement. `pytest-doctestplus` is part of the `sphinx-astropy` metapackage that we already depend on.

pytest.ini
```diff
@@ -1,3 +1,57 @@
 [pytest]
 addopts = -rs --ignore=setup.py --ignore=test_requirements.txt
 norecursedirs = .git build dist tmp* .eggs
+text_file_format = rst
```
hamogu, line 4: @Marie-Terrell I could not get away without any changes to to `pytest.ini`, so I'm asking for your review, too. However, not that the options are `doctest_*` and `test_file_format` and since we don't (yet) run doctests in CI, there should be no problem.

sherpa/astro/data.py
```diff
@@ -72,7 +72,7 @@
 coordinate 3150,4515:
 
 >>> from sherpa.astro.io import read_image
->>> img = read_image('clus.fits')
+>>> img = read_image('sherpa-test-data/sherpatest/image2.fits')
```
hamogu, line 75: works when sherpa-test-data is installed and will hopefully indicate to a user "put your own file name here".

hamogu:
@DougBurke Do you have a chance to review this? I originally had changes to CI in here to run the doctests automatically, but removed them after the discussion we had at our May meeting. It's still waiting for a formal review though. While this is about formatting of examples in the docs and is not urgent, there is also no reason to not include it in the release.

pytest.ini
hamogu: I originally asked for @Marie-Terrell review, just because I touched the `pytest.ini` file. However, the options that I added are not used in CI, so I have removed her from the list of "required reviews". I'd still welcome any feedback if you have any, of course.

anetasie:
I just installed `pytest` and ran a few doc tests. This is good for catching the typos and errors in the docs and it would help to debug our docs in the long run.  We should try to get this into the main branch.

Is there an option to ignore the errors such as the one below due to the mismatched in the numbers.

```
pytest --doctest-rst docs/quick.rst
======================================== test session starts =========================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/aneta/git/sherpa, configfile: pytest.ini
plugins: anyio-3.5.0, doctestplus-0.11.1
collected 1 item                                                                                     

docs/quick.rst F                                                                               [100%]

============================================== FAILURES ==============================================
________________________________________ [doctest] quick.rst _________________________________________
538 
539 The data can be accessed, e.g. to create a dictionary where the
540 keys are the parameter names and the values represent the parameter
541 ranges::
542 
543     >>> dvals = zip(errors.parnames, errors.parvals, errors.parmins,
544     ... errors.parmaxes)
545     >>> pvals = {d[0]: {'val': d[1], 'min': d[2], 'max': d[3]}
546     ...          for d in dvals}
547     >>> pvals['gerr.pos']
Expected:
    {'val': 1.2743015983545256, 'min': -0.014098074065595156, 'max': 0.014098074065595156}
Got:
    {'val': 1.2743015983545225, 'min': -0.01409807406557051, 'max': 0.01409807406557051}

/Users/aneta/git/sherpa/docs/quick.rst:547: DocTestFailure
========================================= 1 failed in 0.75s ==========================================
```

anetasie (approved):
This is really useful for checking the docs.

anetasie:
I see some failures in the output of the examples where the expected outputs do not agree. I'm wondering if we have to fix examples by hand. 

```
pytest --doctest-rst docs/evaluation/examples.rst
======================================== test session starts =========================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/aneta/git/sherpa, configfile: pytest.ini
plugins: anyio-3.5.0, doctestplus-0.11.1
collected 1 item                                                                                     

docs/evaluation/examples.rst F                                                                 [100%]

============================================== FAILURES ==============================================
_______________________________________ [doctest] examples.rst _______________________________________
045 method, where the parameter values must be specified in the
046 call along with the grid on which to evaluate the model.
047 The order matches that of the parameters in the model, which can be
048 found from the
049 :py:attr:`~sherpa.models.model.Model.pars` attribute of the model::
050 
051     >>> [p.name for p in gmdl.pars]
052     ['fwhm', 'pos', 'ampl']
053     >>> y2 = gmdl.calc([100, 5050, 100], x)
054     >>> y2 / y1
Expected:
    array([ 2.,  2.,  2.,  2.,  2.])
Got:
    array([2., 2., 2., 2., 2.])

/Users/aneta/git/sherpa/docs/evaluation/examples.rst:54: DocTestFailure
```

hamogu:
@anetasie Did you check out this branch first? Differences in white space and in the last few digits of a floating number should be ignored automatically through doctest option flags (see https://github.com/sherpa/sherpa/pull/1521/files#diff-fb6a686182f16eb54af3c628f38593f347f68aba31de903983023c560288d7a1R6-R9).

anetasie:
@hamogu it looks like I build the main branch. Now, I build this PR's branch but I still have issues running the pytest.
It looks like `quick.rst` is not in the list to ignore the floating points. The other test returns a different error which is perhaps a true issue in the docs and comes from running the example script.

```
(sherpa-git) aneta@saos-Air ~ % ipython
Python 3.9.12 (main, Jun  1 2022, 06:36:29) 
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import sherpa

In [2]: sherpa.__version__
Out[2]: '4.14.1+30.g3cd04812'
```

```
pytest --doctest-rst docs/quick.rst              
======================================== test session starts =========================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/aneta/git/sherpa, configfile: pytest.ini
plugins: anyio-3.5.0, doctestplus-0.11.1
collected 1 item                                                                                     

docs/quick.rst F                                                                               [100%]

============================================== FAILURES ==============================================
________________________________________ [doctest] quick.rst _________________________________________
538 
539 The data can be accessed, e.g. to create a dictionary where the
540 keys are the parameter names and the values represent the parameter
541 ranges::
542 
543     >>> dvals = zip(errors.parnames, errors.parvals, errors.parmins,
544     ... errors.parmaxes)
545     >>> pvals = {d[0]: {'val': d[1], 'min': d[2], 'max': d[3]}
546     ...          for d in dvals}
547     >>> pvals['gerr.pos']
Expected:
    {'val': 1.2743015983545256, 'min': -0.014098074065595156, 'max': 0.014098074065595156}
Got:
    {'val': 1.2743015983545225, 'min': -0.01409807406557051, 'max': 0.01409807406557051}

/Users/aneta/git/sherpa/docs/quick.rst:547: DocTestFailure

```

```
pytest --doctest-rst docs/evaluation/examples.rst
======================================== test session starts =========================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/aneta/git/sherpa, configfile: pytest.ini
plugins: anyio-3.5.0, doctestplus-0.11.1
collected 1 item                                                                                     

docs/evaluation/examples.rst F                                                                 [100%]

============================================== FAILURES ==============================================
_______________________________________ [doctest] examples.rst _______________________________________
163 <https://sci.esa.int/web/xmm-newton>`_ or about a dozen other missions.
164 
165 First we will load in a PHA dataset, along with its response
166 files (:term:`ARF` and :term:`RMF`), and have a look at how we can
167 interrogate the object.
168 
169 ::
170 
171    >>> from sherpa.astro.io import read_pha
172    >>> pha = read_pha('sherpa-test-data/sherpatest/9774.pi')
Expected nothing
Got:
    read ARF file sherpa-test-data/sherpatest/9774.arf
    read RMF file sherpa-test-data/sherpatest/9774.rmf
    read background file sherpa-test-data/sherpatest/9774_bg.pi

/Users/aneta/git/sherpa/docs/evaluation/examples.rst:172: DocTestFailure
----------------------------------------- Captured log call ------------------------------------------
INFO     sherpa.astro.io:__init__.py:471 read ARF file sherpa-test-data/sherpatest/9774.arf
INFO     sherpa.astro.io:__init__.py:471 read RMF file sherpa-test-data/sherpatest/9774.rmf
INFO     sherpa.astro.io:__init__.py:553 read background file sherpa-test-data/sherpatest/9774_bg.pi
========================================== warnings summary =========
```

hamogu:
That seems to be a different version string for Sherpa then I get with this branch:
```python
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> ipython
Python 3.10.0 | packaged by conda-forge | (default, Nov 20 2021, 02:27:15) [Clang 11.1.0 ]
IPython 7.30.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import sherpa

In [2]: sherpa.__version__
Out[2]: '4.14.1+31.g3cd04812'

In [3]:                                                                                                                    
Do you really want to exit ([y]/n)? 
```
When I look at a `git log` I see the commit 3cd04812 on the top, which is also the top commit displayed in github for this PR.
```
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> git log
commit 3cd04812ea14a8bc5eeed82978c91796bbf94f9d (HEAD -> doctestplus, github/doctestplus)
Author: H. Moritz Günther <hgunther@mit.edu>
Date:   Fri Jun 24 10:46:12 2022 -0400

    set-up doctest plus in docs and conf files

commit 1a609943161a9ad5e7006b7c056a15c892b13eba
Author: H. Moritz Günther <hgunther@mit.edu>
Date:   Mon May 23 16:39:04 2022 -0400

    Fix up examplary code and doc files for doctest
    
    fix up docs to model evaluation

commit f019fb33d1103e0eedd16f4cc1e2c44489112f50
Merge: 2e0d6b94 0d90b9a2
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> 
```

With this commit, the first example passes:
```
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> pytest --doctest-rst docs/quick.rst
=================================================== test session starts ===================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/guenther/code/sherpa, configfile: pytest.ini
plugins: doctestplus-0.11.2, xvfb-2.0.0
collected 1 item                                                                                                          

docs/quick.rst .                                                                                                    [100%]

==================================================== 1 passed in 2.08s ====================================================
```

In the second one, I see indeed output that is reported as warning/failure. I can fix that one specifically if we want to, but I don't think this should stop this PR:

- This particular case will change when this is merged or rebased, because some of the messages reported (the warnings) have been altered in #1477.
- Since we do not run this test in CI, neither now nor after this PR, it is inevitable that differences creep back in as other PRs are merged that change error messages, warnings, or numerical out (e.g. number of digits reported). For that reason, my long term goal is to enable that in CI, but at our last meeting we decided to get this PR in as is and evaluate the possibility of adding this to CI at a later time.
- This PR is going to improve a number of our docs, but it's not comprehensive, i.e. many more files need to be changed. However, doing them all at once takes a long time and will create a lot of conflicts with other PRs. It's much easier to do many smaller PRs, change just a few files, and get that merged quickly.

Thus, my goal for this PR is now to make the doctest available and improve a select number of our docs as examples, but not to fix each and every instance. In my view, it's enough that many of the files I touched pass right now (or did in Spring 2022 when I opened this), that all changes are improvements, and that all downstream (e.g. generating ahelp files from docstrings) processing works.

That said, I'll rebase this and see how the changes merged since changed the output and how to fix the example @anetasie listed.

hamogu:
I rebased and edited examples that have been recently added to the docs.

Note: I did not check any examples that includes WCS, because my local development sherpa is apparently build without the wcs module. Not sure why and need to look into that, but that's totally independent of this PR and, as I argued above, there is really no need to get this perfect at this time, as long as it's better.

The point is to see if we all are willing to accept that changes that need to be made to make those doctests pass (e.g. extra " #doctest: +SKIP" at the end of some lines in the examples, reformatting some of the outputs - see changed files to see what else I had to change). IF we all agree and merge this PR we can later include more files and be more rigorous.

anetasie:
@hamogu Thanks. I updated and build the branch and this test passes. Yes, this PR is not to fix all the tests. I do think that it improves the overall checks of our docs and we should merge it.

```
pytest --doctest-rst docs/quick.rst
======================================== test session starts =========================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/aneta/git/sherpa, configfile: pytest.ini
plugins: anyio-3.5.0, doctestplus-0.11.1
collected 1 item                                                                                     

docs/quick.rst .                                                                               [100%]

========================================= 1 passed in 2.44s ===
```

docs/developer/index.rst
```diff
@@ -98,6 +98,31 @@ run allows us to generate a coverage report after that::
 The report is in ``report/index.html``, which links to individual
 files and shows exactly which lines were excuted while running the tests.
 
+Run doctests locally
```
DougBurke, line 101: A question for this section. You show runs for four explicit files. Are these the only files that have been converted, or just an example set?

docs/developer/index.rst
```diff
@@ -736,6 +761,7 @@ with:
 A regularly-gridded 2D dataset can be created, but note that the
 arguments must be flattened:
 
+  >>> import numpy as np
```
DougBurke, line 764: Here's questions I have about the testing of each "snippet":

- when do symbols have to be imported or created?
  - Imports: I assume that we don't need to do it for each test, so that this import will cover follow-up tests, but if I just had a section at the top of the page (as I do in some places) that says something like "the following examples assumes these imports ,... and then imports a bunch of modules", then we can use them in all following tests
  - symbols: if I define "foo" somewhere as a variable, is it then present for later tests?

docs/developer/index.rst
```diff
@@ -882,12 +908,16 @@ raised. The aim is to provide some context in the message, such as::
   >>> x = np.asarray([1, 2, 3])
   >>> y = np.asarray([1, 2])
   >>> data = Data1D('example', x, y)
+  Traceback (most recent call last):
+  ...
```
DougBurke, line 912: I'm not a fan of this extra noise, but what ya gonna do!

docs/developer/index.rst
```diff
@@ -1136,11 +1166,11 @@ We can view the model plot object::
     >>> print(plot)
     xlo    = [2,3,5,7,8]
     xhi    = [3,5,6,8,9]
-    y      = [ 8.5,20. ,11.5,13.5,14.5]
+    y      = [ 6.,12., 6., 6., 6.]
     xlabel = x
     ylabel = y
     title  = Model
-    histo_prefs = {'yerrorbars': False, 'ecolor': None, ... , 'linecolor': None}
+    histo_prefs = {'xerrorbars': False, 'yerrorbars': False, ..., 'linecolor': None}
```
DougBurke, line 1173: Hmm. dictionary ordering has calmed down from earlier python versions, so I think it's now safe to assume a given order.

docs/developer/index.rst
```diff
@@ -1079,7 +1109,7 @@ The :py:meth:`sherpa.ui.utils.Session.plot` and
 plots to be created by specifying the plot type as a list of
 argumemts. For example::
 
-    >>> s.plot('data', 'model', 'data', 2, 'model', 2)
+    >>> s.plot('data', 'model', 'data', 2, 'model', 2)  # doctest: +SKIP
```
DougBurke, line 1112: Just to check: the reason for skipping this is not because we don't want to check the screen output here (which doesn't exist, but you might want to enforce this), but because this is a snippet where we haven't set up the requirements - i.e. we haven't loaded data or created models for the default id and id 2)?

That is, we have multiple reasons to skip a test.

I'm also pleased to see that the doctest comment is skipped in the output.

docs/evaluation/combine.rst
```diff
@@ -306,8 +306,8 @@ values)::
     >>> mplot.prepare(d, mdl)
     >>> fplot.prepare(dplot, mplot)
     >>> fplot.plot()
-    >>> plt.plot(x, ystart, label='Start')
-    >>> plt.legend(loc=2)
+    >>> out = plt.plot(x, ystart, label='Start')
+    >>> out = plt.legend(loc=2)
```
DougBurke, line 310: Why are you capturing the output from these two lines? Can we not just label these as `# doctest: +SKIP`?

Hmm, this leads me to question what the SKIP directive does: does it run the line but does not check the output, or does it just not run the line (I assume the former)?

If you really want to capture the output, we could use something like

1. add a semicolon at the end of the line
2. use `_ = plt....`

docs/developer/index.rst
hamogu: rst files are run like one big single piece of code form top to bottom. You have access to everything (imports and variables) that was defined before. There is special markup to define things in a way that doctest knows about is, but sphinx does not show it and also the possibility or a prefix (e.g. run "import numpy as np" before every file), but I have not used either of them here: https://github.com/astropy/pytest-doctestplus

docs/developer/index.rst
hamogu: Yes. Always in insertion order now.

docs/developer/index.rst
hamogu: Yes and yes. (We did not set it up and I don't want to compare the screen output). And yes, with the plugin activates, Sphinx knows about doctest and will remove that markup. The question is if there is any place where we have to worry about that for other downstream code, e.g. do we make ahelp files from rst files or docstrings?

docs/evaluation/convolution.rst
```diff
@@ -92,26 +93,26 @@ since it has sharp edges and so the convolution is more obvious).
     >>> point.xlow = 9.5
     >>> point.xhi = 10.5
     >>> print(point)
-    point
+    pt
```
DougBurke, line 96: It would probably have been easier to change the earlier line to

    >>> point = Box1D('point')

but you'd still have to change the `convolved = psf1(point)` line below.

I'm not saying that you should do this (I think re-iterating the fact that the model name and the variable name do not need to be the same thing has some value!), just musing aloud.

docs/evaluation/combine.rst
DougBurke: There are other places where I have the same question, but I'll not litter the comments with them.

docs/evaluation/examples.rst
```diff
@@ -169,10 +169,12 @@ interrogate the object.
 ::
 
    >>> from sherpa.astro.io import read_pha
-   >>> pha = read_pha('9774.pi')
-   read ARF file 9774.arf
-   read RMF file 9774.rmf
-   read background file 9774_bg.pi
+   >>> pha = read_pha('sherpa-test-data/sherpatest/9774.pi')
```
DougBurke, line 172: Having to explicitly set the path here is only really useful for the test and doesn't really help the user other than making the text longer. Is there any way to add code which is only used by doctest so we could start this section with something like

    import os
    thisdir = os.getcwd()
    os.chdir("sherpa-test-data/sherpatest")

and then at the end

    os.chdir(thisdir)

? I imaging this isn't going to be the only place I'd want to see this.

docs/evaluation/convolution.rst
hamogu: I'll think about it...

docs/evaluation/examples.rst
hamogu: This might be a good case for a (not visible) setup.

docs/evaluation/combine.rst
hamogu: I'm not a big fan on `_ = plt...` because `_` looks somewhat special while it really is a normal variable. In other cases (not in this example) the user might even want to do something with that variable like:
```python
out = plt.legend()
out.title = 'some text'
```
So I try to be as "normal" as I can here.

docs/quick.rst
```diff
@@ -479,7 +480,7 @@ The default error estimation routine is
 
     >>> from sherpa.estmethods import Confidence
     >>> gefit.estmethod = Confidence()
-    >>> print(gefit.estmethod)
+    >>> print(gefit.estmethod)  # doctest: +SKIP
```
DougBurke, line 483: Why is this being skipped?

docs/evaluation/integrate.rst
```diff
@@ -76,12 +76,12 @@ each bin::
...
     array([-22.5,   2.5,  27.5,  52.5])
 
-Thanks to the easy functonal form chosen for this example,
+Thanks to the easy functional form chosen for this example,
 it is easy to confirm that these are the values of the
 integrated model::
   
     >>> (y[:-1] + y[1:]) * 5 / 2.0
-    array([-22.5,   2.5,  27.5,  52.5])
+    array([-50.,  75., 200.])
```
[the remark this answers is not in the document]
hamogu: This is a good example why running the examples in the text is useful - if I was a novice user, that mismatch is potentially confusing. I'll remove the code and adjust the text.

docs/quick.rst
```diff
@@ -502,12 +503,6 @@ attribute. Note that a message is displayed to the screen when each
 bound is calculated, to indicate progress::
   
     >>> errors = gefit.est_errors()
-    gerr.fwhm lower bound:	-0.0326327
-    gerr.fwhm upper bound:	0.0332578
-    gerr.pos lower bound:	-0.0140981
-    gerr.pos upper bound:	0.0140981
-    gerr.ampl lower bound:	-0.0456119
-    gerr.ampl upper bound:	0.0456119
```
DougBurke, line 510: I don't understand why these lines have been removed. They are written out and I'd want users to see this in the output. 

I think the issue is that this screen output is to the Sherpa logger and this is probably something that doctest doesn't handle well. Many years ago I did some work to catch the log output for some reason, but I forgot what that was for but it was docs related.

sherpa/astro/data.py
```diff
@@ -110,11 +110,11 @@
 this range to have at least 20 counts per group:
 
 >>> from sherpa.astro.io import read_pha
->>> pha = read_pha('src.pi')
+>>> pha = read_pha('sherpa/astro/datastack/tests/data/3c273.pi')
 >>> pha.subtract()
 >>> pha.set_analysis('energy')
 >>> pha.notice(0.5, 7)
->>> pha.group_counts(20, tabStops=~pha.mask)
```
DougBurke, line 117: Why did the `tabStops` argument get removed?

sherpa/astro/data.py
```diff
@@ -2102,6 +2104,9 @@ def get_analysis(self):
         Examples
         --------
 
+        >>> from sherpa.astro.io import read_pha
+        >>> pha = read_pha('sherpa/astro/datastack/tests/data/3c273.pi')
```
DougBurke, line 2108: It's going to be very annoying if we have to add imports before each example.

hamogu: We can do that once in the global setup (I need to check where it goes, but something like and ini or `conf.py` file)

docs/quick.rst
hamogu: I'll have to look why I did that, presumably because it's not printed to standard output. I might add a new handler to the logger to print all logger output to stdout, when run as a doctest. Alternatively, we can just skip this example from doctest (not just that one line, because then there is unexpected output, but simply not test that example). That's why it's good to convert just a few files and see where it goes and how we like it.

docs/developer/index.rst
DougBurke: The ahelp files are generated from the sherpa.astro.ui.utils docstrings (and hence sherpa.ui.utils). I'd have to check how the rst code represents such docstrings (ie how to filter them out). I think though that it would take time to get to these docstrings, as they are the most complex to do (as they hit lots of the general issues I'm talking about here as I try to understand the scope).

docs/developer/index.rst
DougBurke: Of course, the way we create any particular dict could change for a variety of reasons, so we may see changes over time (and would have to think about at that time).

Does doctest treat

    {'a': False, 'b': True}
    {'b': True, 'a': False}

the same? For the line highlighted here it may be hard to check since we elide some of the keys, so it would be hard to check.

hamogu: Doctest performs a string comparison of the output. So, two identical dicts can fail if printed in different order. Luckily, now the order of dict items is deterministic in Python, as long as they are generated in a deterministic fashion (which they are for this case). Again, it's a trade-off: Do we want to made doctest happy? Is it worth the extra work or should we just ignore this test or entire output line? Or should we take the fact that the output line is super-long (> 300 characters I think) as a sign that the output is not great to begin with an should be changed?

It's something I don't have a good feeling for yet, but that we'll get experience with as we develop this further. 

In some instances, it might be useful to write code that does not depend on the order e.g. `print(histoprefs['xlog'])` as an example what's in there instead of `print(histoprefs)` or `print(sorted(histroprefs))`; in others we want to show the users how the entire dict looks when printed on screen and then it might be better to just not check that output.

First priority of a docstring should be that it's useful for the reader (which means it's also correct and executes without error). Using it as unit test is a lower priority to me, but does not hurt if we can get it to work with minimal extra effort and without reducing the usefulness of the docs for the reader.

DougBurke:
FYI, I get the same two failures as @anetasie does with a `pytest --doctest-rst` - which appear to be just "numerical noise".

This is with 

```
%git log --oneline upstream/main..HEAD 
3cd0481 set-up doctest plus in docs and conf files
1a60994 Fix up examplary code and doc files for doctest
```

hamogu:
Maybe it's something to do with the fact that I run on Mac and you use Linux or that I run on M1 and you run on intel chips. Either way, we might just not want to worry about it; either just not fix it now (we only run those files by hand and can punt for now) or reduce the number of digits that we compare, e.g. by using `...` or by `np.printoptions(precision=8)` in a (hidden) setup. It indicates that our users also won't get the exact same numbers and the docs are easier to read if we provide just a large, but not a ridiculously large number of digits.

pytest.ini
```diff
@@ -1,3 +1,57 @@
...
+doctest_subpackage_requires =
+    sherpa/astro/io/crates_backend.py = pycrates
+    sherpa/astro/io/pyfits_backend.py = astropy
+    sherpa/plot/pylab_backend.py = matplotlib
+    # Don't know how to check if ds9 and xspec are installed for this purpose.
+    # They are not Python packages, but the following syntax will skip these files
+    # since Python packages named ds9 or xspec don't exist and thus won't be found.
+    sherpa/image/*ds9_backend.py = ds9
+    sherpa/image/DS9.py = ds9
+    scripts/check_xspec_update.py = xspec
```
DougBurke, line 57: Three thoughts:


1. So does this mean that at the moment we will never run tests of these particular files?
2. is the `*` in the `ds9_backend` line intended or just a typo? I'd expect an explicit path here, so I would assume a type
3. I also think that we probably want to just always ignore the `scripts` directory (as noted above) and so we could drop the last line

sherpa/astro/data.py
DougBurke: Which is something we can't resolve in this PR but I wanted to mention it.

docs/developer/index.rst
```diff
@@ -98,6 +98,31 @@ run allows us to generate a coverage report after that::
...
+This guarantees that the examples actually work and don't have typos or outdated
+parameters, which might confuse a user.
+
+Running `pytest --doctest-plus` or `pytest --doctest-rst` (without extra parameters)
+will run all usual tests and also the doctests on all files not specifically
+excluded in `pytest.ini`.
+If you fix examples to pass these tests, remove them from the exclusion list in
+`pytest.ini`! The goal is to eventually pass on all files.
+
+Note: Doctests are not currently run on CI.
```
DougBurke, line 124:
```suggestion
.. note::
   The doctests are not run as part of the CI process. The plan is to add them in once we have more experience with the system.
```

docs/developer/index.rst
```diff
@@ -98,6 +98,31 @@ run allows us to generate a coverage report after that::
...
+
```
DougBurke, line 125: There's currently a hidden requirement to running the doctests, and that is that the `sherpa-test-data` submodule is nistalled/activated/whatever the command is. I think this should be documented, in part because it means that users who have downloaded the code via PyPI/Zenodo/.... won't be able to run the doctests (we do have a separate way to install the test data as a pyton package *but* that is not hooked up to the doctest system, as it installs the data separately and is not something we need to bother about here yet).

docs/evaluation/examples.rst
DougBurke: I don't think we need to resolve this for this PR, but would like it looked into in the glorious future when we have time to do so.

docs/quick.rst
hamogu: The output depends on the machine you are running on because the `numcores` lines actually reports the number of cores. I don't see an easy way to inject a fixed number of cores here for this test, neither do I want to replace it with `...`. 

I think that's one of those cases where it's not worth the effort to use that as a test - let's just skip it.

pytest.ini
hamogu: 1. We will never run the doctests in the files that have a dependence on "xspec" or "ds9" since we don't have python packages with those names. However, I've since learned a better way to accomplish this task (define a pytest fixtures that tests if xspec or ds9 are available and call `pytest.skip("ds9 not available")` when it's not. However, those files are not very high on my priority list.
2. It's intentional to cover `ds9_backend.py` and `pyds9_backend.py`, but not `dummy_backend.py`. I could spell it out, but again this is probably going to be reworked when we get to those files anyway.
3. Ok.

pytest.ini
```diff
@@ -1,3 +1,57 @@
...
+doctest_plus = enabled
+doctest_optionflags =
+    NORMALIZE_WHITESPACE
+    ELLIPSIS
+    FLOAT_CMP
+doctest_norecursedirs =
+    extern
+    sherpa-test-data
+    # Blacklist everything and then remove from this list when it's ready
+    docs/_examples
```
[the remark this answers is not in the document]
hamogu: I think we can do that.

sherpa/astro/data.py
hamogu: UNEXPECTED EXCEPTION: ValueError('grpNumCounts() The tabStops and countsArray have differing length')

Of course, I'm loading a specific, real data file so it might behave different now than the hypothetical `src.pi`. 
Throwing a few `get_filter_expr` in between, we see that, for this data file, even right after the notice call the filter boundary is a lot higher than requested because the bins are so wide. 
```python
In [8]: pha.get_filter_expr()
Out[8]: '0.0015-14.9504 Energy (keV)'

In [9]: pha.subtract()

In [10]: pha.set_analysis('energy')

In [11]: pha.notice(0.5, 7)

In [12]: pha.get_filter_expr()
Out[12]: '0.4672-9.8696 Energy (keV)'

In [13]: pha.group_counts(20)

In [14]: pha.get_filter_expr()
Out[14]: '0.4380-13.4612 Energy (keV)'
```
The root cause is that the data is read in grouped, and thus the length does not match even right after reading the file:
```python
In [24]: pha = read_pha(data_3c273 + '3c273.pi')
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273.pi' 
read ARF file sherpa/astro/datastack/tests/data/3c273.arf
read RMF file sherpa/astro/datastack/tests/data/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi'
statistical errors were found in file 'sherpa/astro/datastack/tests/data/3c273_bg.pi' 
read background file sherpa/astro/datastack/tests/data/3c273_bg.pi

In [25]: len(pha)
Out[25]: 1024

In [26]: len(pha.mask)
Input In [26], in <cell line: 1>()
----> 1 len(pha.mask)

TypeError: object of type 'bool' has no len()

In [27]: pha.notice(0.5, 7)

In [28]: pha.mask.shape
Out[28]: (46,)
```

I have thus removed the `, tabStops=~pha.mask` to make an example that works in all cases and does not depend on loading a data file without initial grouping.

docs/quick.rst
hamogu: The easiest solution is to show the output in a separate block that is not looked at by the doctest because it's not a block that looks like Python code. Slightly not ideal, but much easier and, I would argue, still good for this purpose. (Check tomorrow after I push to GH.)

sherpa/astro/data.py
DougBurke: This is one of those case where I was trying to let users know about the tabStops feature without a lot of fanfare. I think

    , tabstops=~pha.get_mask()

should work (as `get_mask` expands out the mask). Except when there's no filter present, but in this example we do have one.

```
>>> from sherpa.astro.io import read_pha
>>> pha = read_pha("sherpa-test-data/sherpatest/3c273.pi")
>>> len(pha)
1024
>>> pha.mask
True
>>> pha.get_mask().shape
AttributeError: 'NoneType' object has no attribute 'shape'
>>> pha.notice(0.5, 7)
>>> pha.mask.shape
(46,)
>>> pha.get_mask().shape
(1024,)
>>> pha.group_counts(20, tabStops=~pha.get_mask())
>>> 
```

docs/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+import pytest
+
+@pytest.fixture(autouse=True)
+def add_sherpa_test_data_dir(doctest_namespace):
+    doctest_namespace["data_dir"] = 'sherpa-test-data/sherpatest/'
```
DougBurke, line 25: So, we have `sherpa.utils.testing.get_datadir` which could be used here. Or perhaps either of the `test_data_path` or `make_data_path` decorators from `sherpa/conftest.py` (although are they in scope here???). Using these fixtures could be good, since it lets you know that the data directory is required (assuming it all hooks up nicely).

docs/evaluation/simulate.rst
```diff
@@ -65,8 +65,8 @@ and counts empty, because these will be filled in by the simulation::
   >>> from sherpa.astro.data import DataPHA
   >>> from sherpa.astro.io import read_arf, read_rmf, read_pha
   >>> data = DataPHA(name='any', channel=None, counts=None, exposure=10000.)
-  >>> data.set_arf(read_arf('sherpa-test-data/sherpatest/9774.arf'))
-  >>> data.set_rmf(read_rmf('sherpa-test-data/sherpatest/9774.rmf'))
+  >>> data.set_arf(read_arf(data_dir + '9774.arf'))
+  >>> data.set_rmf(read_rmf(data_dir + '9774.rmf'))
```
DougBurke, line 69: Out of interest, where is `data_dir` defined? 

Oh. I see, it's from

        doctest_namespace["data_dir"] = 'sherpa-test-data/sherpatest/'

docs/quick.rst
DougBurke: I've had a quick look and I can't see any obvious PRs where I added the grab-the-log-output-for-the-docs code, which probably means it's lost to the winds.

DougBurke:
Things I don't understand...

I ran a `pytest --doctest-rst` run and got two failures. The first is the precision issue in `docs/quick.rst`, and the second is in `docs/evaluation/examples.rst`, with the output:

```
163 <https://sci.esa.int/web/xmm-newton>`_ or about a dozen other missions.
164 
165 First we will load in a PHA dataset, along with its response
166 files (:term:`ARF` and :term:`RMF`), and have a look at how we can
167 interrogate the object.
168 
169 ::
170 
171    >>> from sherpa.astro.io import read_pha
172    >>> pha = read_pha( data_dir + '9774.pi')
Expected:
    read ARF file .../9774.arf
    read RMF file .../9774.rmf
    read background file .../9774_bg.pi
Got nothing

/home/dburke/sherpa/sherpa-work/docs/evaluation/examples.rst:172: DocTestFailure
-------------------------------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------------------------------
read ARF file sherpa-test-data/sherpatest/9774.arf
read RMF file sherpa-test-data/sherpatest/9774.rmf
read background file sherpa-test-data/sherpatest/9774_bg.pi
---------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------
INFO     sherpa.astro.io:__init__.py:471 read ARF file sherpa-test-data/sherpatest/9774.arf
INFO     sherpa.astro.io:__init__.py:471 read RMF file sherpa-test-data/sherpatest/9774.rmf
INFO     sherpa.astro.io:__init__.py:553 read background file sherpa-test-data/sherpatest/9774_bg.pi
```

This makes sense, because your last commit included

```
    >>> from sherpa.astro.io import read_pha
-   >>> pha = read_pha('sherpa-test-data/sherpatest/9774.pi')
-   read ARF file sherpa-test-data/sherpatest/9774.arf
-   read RMF file sherpa-test-data/sherpatest/9774.rmf
-   read background file sherpa-test-data/sherpatest/9774_bg.pi
+   >>> pha = read_pha( data_dir + '9774.pi')
+   read ARF file .../9774.arf
+   read RMF file .../9774.rmf
+   read background file .../9774_bg.pi
```

and the "read ... file ..." lines are handled by the logger, which we know the code doesn't catch.

**BUT**

If I run just this file it fials in a **DIFFERENT** place , and in a later line than the failure reported above.

```
% pytest --doctest-rst docs/evaluation/examples.rst
===================================================================================================================== test session starts =====================================================================================================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/dburke/sherpa/sherpa-work, configfile: pytest.ini
plugins: anyio-3.6.1, forked-1.4.0, xdist-2.5.0, doctestplus-0.12.0
collected 1 item                                                                                                                                                                                                                                              

docs/evaluation/examples.rst F                                                                                                                                                                                                                          [100%]

========================================================================================================================== FAILURES ===========================================================================================================================
___________________________________________________________________________________________________________________ [doctest] examples.rst ____________________________________________________________________________________________________________________
497 
498 We can use the model including the response to
499 :doc:`fit the data <../fit/index>` (here I am not going
500 to tweak the statistic choice or optimiser which you should
501 consider)::
502 
503    >>> from sherpa.fit import Fit
504    >>> fit = Fit(pha, full)
505    >>> res = fit.fit()
506    >>> print(res.format())
Differences (unified diff with -expected +actual):
    @@ -8,5 +8,5 @@
     Reduced statistic     = 0.716768
     Change in statistic   = 3.34091e+11
    -   phabs.nH       0.0129623    +/- 0.00727297
    -   powlaw1d.gamma   1.78432      +/- 0.0459881
    -   powlaw1d.ampl   7.17014e-05  +/- 2.48851e-06
    +   phabs.nH       0.0129625    +/- 0.00727019  
    +   powlaw1d.gamma   1.78432      +/- 0.0459786   
    +   powlaw1d.ampl   7.17014e-05  +/- 2.48751e-06 

/home/dburke/sherpa/sherpa-work/docs/evaluation/examples.rst:506: DocTestFailure
---------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------
INFO     sherpa.astro.io:__init__.py:471 read ARF file sherpa-test-data/sherpatest/9774.arf
INFO     sherpa.astro.io:__init__.py:471 read RMF file sherpa-test-data/sherpatest/9774.rmf
INFO     sherpa.astro.io:__init__.py:553 read background file sherpa-test-data/sherpatest/9774_bg.pi
====================================================================================================================== 1 failed in 0.96s ======================================================================================================================
```

I am sure this makes sense to someone, somewhere. But that person is not me...

DougBurke:
So, we appear to have something in `sherpa/conftest.py` that changes things (I checked by adding a raise error in the latter and it only gets triggered in exhibit b)

# Exhibit a

Only uses `docs/conftest.py`:

```
pytest --doctest-rst docs/evaluation/examples.rst 
==================================================== test session starts =====================================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/dburke/sherpa/sherpa-work, configfile: pytest.ini
plugins: anyio-3.6.1, forked-1.4.0, xdist-2.5.0, doctestplus-0.12.0
collected 1 item                                                                                                             

docs/evaluation/examples.rst F                                                                                         [100%]

========================================================== FAILURES ==========================================================
___________________________________________________ [doctest] examples.rst ___________________________________________________
497 
498 We can use the model including the response to
499 :doc:`fit the data <../fit/index>` (here I am not going
500 to tweak the statistic choice or optimiser which you should
501 consider)::
502 
503    >>> from sherpa.fit import Fit
504    >>> fit = Fit(pha, full)
505    >>> res = fit.fit()
506    >>> print(res.format())
Differences (unified diff with -expected +actual):
    @@ -8,5 +8,5 @@
     Reduced statistic     = 0.716768
     Change in statistic   = 3.34091e+11
    -   phabs.nH       0.0129623    +/- 0.00727297
    -   powlaw1d.gamma   1.78432      +/- 0.0459881
    -   powlaw1d.ampl   7.17014e-05  +/- 2.48851e-06
    +   phabs.nH       0.0129625    +/- 0.00727019  
    +   powlaw1d.gamma   1.78432      +/- 0.0459786   
    +   powlaw1d.ampl   7.17014e-05  +/- 2.48751e-06 

/home/dburke/sherpa/sherpa-work/docs/evaluation/examples.rst:506: DocTestFailure
----------------------------------------------------- Captured log call ------------------------------------------------------
INFO     sherpa.astro.io:__init__.py:471 read ARF file sherpa-test-data/sherpatest/9774.arf
INFO     sherpa.astro.io:__init__.py:471 read RMF file sherpa-test-data/sherpatest/9774.rmf
INFO     sherpa.astro.io:__init__.py:553 read background file sherpa-test-data/sherpatest/9774_bg.pi
===================================================== 1 failed in 0.95s ======================================================
```

# Exhibit b

Also uses 'sherpa/conftest.py`

```
% pytest --doctest-rst docs/evaluation/examples.rst sherpa/tests/test_stack.py 
==================================================== test session starts =====================================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/dburke/sherpa/sherpa-work, configfile: pytest.ini
plugins: anyio-3.6.1, forked-1.4.0, xdist-2.5.0, doctestplus-0.12.0
collected 6 items                                                                                                            

docs/evaluation/examples.rst F                                                                                         [ 16%]
sherpa/tests/test_stack.py .....                                                                                       [100%]

========================================================== FAILURES ==========================================================
___________________________________________________ [doctest] examples.rst ___________________________________________________
163 <https://sci.esa.int/web/xmm-newton>`_ or about a dozen other missions.
164 
165 First we will load in a PHA dataset, along with its response
166 files (:term:`ARF` and :term:`RMF`), and have a look at how we can
167 interrogate the object.
168 
169 ::
170 
171    >>> from sherpa.astro.io import read_pha
172    >>> pha = read_pha( data_dir + '9774.pi')
Expected:
    read ARF file .../9774.arf
    read RMF file .../9774.rmf
    read background file .../9774_bg.pi
Got nothing

/home/dburke/sherpa/sherpa-work/docs/evaluation/examples.rst:172: DocTestFailure
---------------------------------------------------- Captured stdout call ----------------------------------------------------
read ARF file sherpa-test-data/sherpatest/9774.arf
read RMF file sherpa-test-data/sherpatest/9774.rmf
read background file sherpa-test-data/sherpatest/9774_bg.pi
----------------------------------------------------- Captured log call ------------------------------------------------------
INFO     sherpa.astro.io:__init__.py:471 read ARF file sherpa-test-data/sherpatest/9774.arf
INFO     sherpa.astro.io:__init__.py:471 read RMF file sherpa-test-data/sherpatest/9774.rmf
INFO     sherpa.astro.io:__init__.py:553 read background file sherpa-test-data/sherpatest/9774_bg.pi
================================================ 1 failed, 5 passed in 0.24s =================================================
```

My guess it's `capture_all_warnings`, but I' mnot 100% convinced I understand all the moving parts.

sherpa/conftest.py
```diff
@@ -647,3 +647,8 @@ def cleanup_ds9_backend():
     #
     from sherpa.image import backend
     backend.close()
+
+@pytest.fixture(autouse=True)
+def add_sherpa_test_data_dir(doctest_namespace):
+    doctest_namespace["data_dir"] = 'sherpa-test-data/sherpatest/'
```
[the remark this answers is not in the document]
DougBurke: Nope. We don't want to do that.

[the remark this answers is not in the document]
DougBurke: Here's what I came up with:

```
% git diff .
diff --git a/docs/conftest.py b/docs/conftest.py
index 4bd65e31..e6b72650 100644
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -20,6 +20,20 @@
 
 import pytest
 
+from sherpa.utils.testing import get_datadir
+
+
 @pytest.fixture(autouse=True)
 def add_sherpa_test_data_dir(doctest_namespace):
-    doctest_namespace["data_dir"] = 'sherpa-test-data/sherpatest/'
\ No newline at end of file
+
+    # We could error out here, but we only want the tests that
+    # use it to fail.
+    #
+    path = get_datadir()
+    if path is None:
+        return
+
+    if not path.endswith('/'):
+        path += '/'
+
+    doctest_namespace["data_dir"] = path
diff --git a/sherpa/conftest.py b/sherpa/conftest.py
index 9a0956d2..e7416c23 100644
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -648,7 +648,18 @@ def cleanup_ds9_backend():
     from sherpa.image import backend
     backend.close()
 
+
 @pytest.fixture(autouse=True)
 def add_sherpa_test_data_dir(doctest_namespace):
-    doctest_namespace["data_dir"] = 'sherpa-test-data/sherpatest/'
+
+    # We do not want to use test_data_path/make_data_path as this
+    # errors out when the data is not available.
+    #
+    path = get_datadir()
+    if path is not None:
+        if not path.endswith('/'):
+            path += '/'
+
+        doctest_namespace["data_dir"] = path
+
     doctest_namespace["data_3c273"] = 'sherpa/astro/datastack/tests/data/'
```

The doc tests that require the data fails now fail with

```
068   >>> data.set_arf(read_arf(data_dir + '9774.arf'))
UNEXPECTED EXCEPTION: NameError("name 'data_dir' is not defined")
  File "/home/dburke/miniconda3/envs/sherpa-work/lib/python3.10/doctest.py", line 1350, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest simulate.rst[17]>", line 1, in <module>
NameError: name 'data_dir' is not defined
/home/dburke/sherpa/sherpa-work/docs/evaluation/simulate.rst:68: UnexpectedException
```

I thought about setting the path to something like "data-does-not-exist/", so the error message would come from the read call (and it may catch cases where we have actually defined `data_dir` somewhere), but this looks cleaner to identify.

It doesn't look like there's a nice way to mark these tests as "skip if data directory is not installed".

docs/quick.rst
hamogu: I've wondered, too but I'm not sure where to put it. I don't want to pollute the docstring. I'll add a `#  comment` below the docstring.

docs/conftest.py
hamogu: `conftest.py` is only looked for the the current tree, not in a sibling tree - there is a specific warning in the pytest docs for cases like ours where docs and code are in separate trees. We *could* move the pytest doc into the base dir, but I don't think it belongs there. That's where this is parallel (i.e. a copy of) whats in the `sherpa/conftest.py`. At the minimum, I'll add a comment here to explain that. I did not know about (or forgot) `sherpa.utils.testing.get_datadir`, so I'll look if I can use that.

hamogu:
I've gotten to the bottom of the precision issue. With the default settings for doctestplus, we compare floating numbers anywhere in the output only to a absolute precision of 1e-8, so this should have never triggered the error. The problem is that it uses a regexp to decide what is a number for comparison and that regexp does not find numbers that end in "}" as in `{'val': 1.2743015983545225, 'min': -0.01409807406557051, 'max': 0.01409807406557051}`. Thus, the last number (max) will be compared as a string and that comparison will fail if it's not an exact match.

I opened https://github.com/astropy/pytest-doctestplus/pull/186 and thus I suggest we don't do anything about the offending line in `quick.rst` for now. The people involved in this discussion know why there is this failure and it will go away by itself when doctestplus merges my PR and makes a release. 

Should that not happen by the time we want to switch on a CI job running doctestplus on Sherpa, we can revisit.

docs/evaluation/integrate.rst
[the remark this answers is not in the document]
DougBurke: I think this is the only remaining "outstanding" issue, and it's not even the fault of this PR.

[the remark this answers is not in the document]
hamogu: That's been addressed in https://github.com/sherpa/sherpa/pull/1521/commits/63955817a6a1a434be212c232db77347e87b45a7#diff-af67bde9c745757887133a4bf48b0aa4b0d498f27ee82a8ce708f3569c4a9bc7

but I'm still looking at `data_dir` and skipping tests is the test-data does not exist. We can declare the latter out of scope for this PR if I don't have full solution soon, but give me another hour or so.

sherpa/conftest.py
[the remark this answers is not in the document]
hamogu: > It doesn't look like there's a nice way to mark these tests as "skip if data directory is not installed".

That's because in a narrative text, there is no easy telling where one example begins and the other one ends (recall that variables are kept form the beginning to the end of the file so you can interleave code and explanatory text). So, the only option is really to step testing the rest of the file at that point.

And let's not be more picky than we have to be. We are not going to run doctests that same way we do other CI on different operating systems, with different version of numpy and Python, etc. I envision that we'll add doctests to just one single runner eventually, and for now it's only run locally be developers you want to make sure the docs still work or their new examples are complete. So, I'm going with "if the test data is not available, raise an exception and stop altogether". I don't think we have to try too hard to work around that.

hamogu:
I think I've addressed all remaining issues. See commit message for details. Tests without the datadir present will now look like this:
```
sherpaciao) melkor:/melkor/d1/guenther/soft/python/sherpa> pytest --doctest-rst docs/evaluation/examples.rst --doctest-plus-rtol=1e-3
======================================================================== test session starts =========================================================================
platform linux -- Python 3.8.2, pytest-7.1.3, pluggy-0.13.1
rootdir: /melkor/d1/guenther/soft/python/sherpa, configfile: pytest.ini
plugins: cov-3.0.0, astropy-header-0.2.2, xdist-2.5.0, remotedata-0.3.3, hypothesis-6.54.6, arraydiff-0.5.0, openfiles-0.5.0, mock-3.8.2, astropy-0.10.0, forked-1.4.0, filter-subpackage-0.1.1, doctestplus-0.11.0
collected 1 item                                                                                                                                                     

docs/evaluation/examples.rst s                                                                                                                                 [100%]

========================================================================== warnings summary ==========================================================================
../../../../../../nfs/melkor/d1/guenther/soft/mambaforge/envs/sherpaciao/lib/python3.8/site-packages/pytest_doctestplus/plugin.py:602
  /nfs/melkor/d1/guenther/soft/mambaforge/envs/sherpaciao/lib/python3.8/site-packages/pytest_doctestplus/plugin.py:602: PytestRemovedIn8Warning: The (fspath: py.path.local) argument to DocTestTextfilePlus is deprecated. Please use the (path: pathlib.Path) argument instead.
  See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
    return self._doctest_textfile_item_cls.from_parent(parent, fspath=path)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
====================================================================== short test summary info =======================================================================
SKIPPED [1] docs/conftest.py:47: sherpa-test-data not found
=================================================================== 1 skipped, 1 warning in 0.05s ====================================================================
```

sherpa/conftest.py
```diff
@@ -650,5 +650,25 @@ def cleanup_ds9_backend():
...
+    extra markup out of those files and keep them clean.
+    '''
     doctest_namespace["data_3c273"] = 'sherpa/astro/datastack/tests/data/'
+
+    # We could error out here, but we only want the tests that
+    # use it to fail.
+    #
+    path = get_datadir()
+    if path is None:
+        pytest.skip('this doctest does not work on Windows')
```
DougBurke, line 669: Err. I think this message is a tad mis-leading ;-)

hamogu: But it's true!
(and taken from here: https://docs.pytest.org/en/7.1.x/how-to/doctest.html#skipping-tests)

DougBurke (approved):
LGTM - just waiting for the final green. Thanks for dealing with all my picky comments.

hamogu:
The precision issue we've been seeing when running this locally on different machines is now solved with the latest pytest-doctestplus release: https://github.com/astropy/pytest-doctestplus/releases/tag/v0.12.1

Issue #1467, opened 2022-03-11
Sherpa still requires `-Wno-implicit-function-declaration` on current main


hamogu:
Despite #1401, I still get an error (shortend log below because full output is above max github character limit) when trying to compile sherpa on clang (I have 13.0.0). It does work when I use `CFLAGS='-Wno-implicit-function-declaration'`

```
(base) MoritzAirRoseGold ~/code/sherpa> export PYTHON_LDFLAGS=' '
(base) MoritzAirRoseGold ~/code/sherpa> pip install -e .
Obtaining file:///Users/guenther/code/sherpa
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: numpy in /Users/guenther/mambaforge/lib/python3.9/site-packages (from sherpa==4.14.0+281.gbaa439a5) (1.22.2)
Installing collected packages: sherpa
  Running setup.py develop for sherpa
    ERROR: Command errored out with exit status 2:
     command: /Users/guenther/mambaforge/bin/python3.9 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/Users/guenther/code/sherpa/setup.py'"'"'; __file__='"'"'/Users/guenther/code/sherpa/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
         cwd: /Users/guenther/code/sherpa/
    Complete output (2620 lines):
    running sherpa_config
    warning: sherpa_config: built configure string['./configure', '--prefix=/Users/guenther/code/sherpa/build', '--with-pic', '--enable-standalone', '--disable-maintainer-mode', '--enable-stuberrorlib', '--disable-shared', '--enable-shared=libgrp,stklib', '--enable-fftw', '--enable-region', '--enable-group', '--enable-stk', '--enable-wcs']
    
    checking for a BSD-compatible install... /usr/bin/install -c
    checking whether build environment is sane... yes
    checking for a thread-safe mkdir -p... ./install-sh -c -d
    checking for gawk... no
    checking for mawk... no
    checking for nawk... no
[...]

PR #1522, opened 2022-05-24
Bring "no-implicit-function-declaration" instructions back for MacOS

hamogu:
This partially addresses Sherpa still requires `-Wno-implicit-function-declaration` on current main #1467, but it woudl be even better
if the C-code did not require this flag, I guess.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -401,6 +401,11 @@ C++ compilers to use if not found by ``setup.py``.
      export PYTHON_LDFLAGS=' '
 
    That is, the variable is set to a space, not the empty string.
+   For the clang compiler (which Apple confusingly aliases to ``gcc``) you
+   also need to set::
+
+       export CFLAGS='-Wno-implicit-function-declaration'
+
 
 A standard installation
 ^^^^^^^^^^^^^^^^^^^^^^^


hamogu:
I opened this PR, but I just run into not being able to build sherpa again, until I rememberd that I had opened that issue with instruction a while ago...

DougBurke:
This should have been fixed by #1523 but I don't have a mac machine to test on (and am not going to install clang on my machine just to check) so I'll leave this open until someone can confirm it works. Or doesn't.

hamogu:
Works on my Mac and thus I'm closing this although I have the following thought:

Some Sherpa user (not from our CXC group that works on Sherpa all the time) run into this problem in real live in #1528. I'm not quite sure what to do about it
1) We could make a branch of main, merge this PR, and get that branch build on readthedocs until we release 4.14.2 (where it's not longer needed). 
2) We cut a new release 4.14.2 in the next few weeks with includes #1523, not not necessarily much else.
3) We do nothing, assume that not many users install Sherpa from source on Macs, and if they do, hope that they ask and we can point them to these instructions.

While I prefer 1 or 2 in principle, given our resources, I personally could live with 3.

PR #1523, opened 2022-05-25
Ensure the region-lib code is re-generated when needed

DougBurke:
# Summary

Fix a build issue with clang where the region-library code caused an error message (issue #1142 which was partly addressed by #1401)

# Details

So, the regparser code should have been excluded via the `.gitignore` file, but thanks to the rule being too specific (it was looking for `region-4.11` and the directory is now `region-4.14`) the files got added at some point. So, the change in #1401, which should have fixed this, was not working for users who had not called `python setup.py clean`. This commit removes the offending files and makes the `.gitignore` rule look in `region-*` rather than a specific directory (and matches what is done for `fftw-*`).


DougBurke:
@hamogo - does this work for you (ie building with clang without having to set `CFLAGS`)?

DougBurke:
Note the test failure is not related to this PR but just a general "annoyance" - which I have a probable fix for in feac57833ab50f14fe2ea96c775a6ecee6884939 as part of #1516

wmclaugh:
@DougBurke 

I don't believe this is the correct fix. If I understand correctly, the regparser.c file gets generated. As soon as you perform a "python develop clean" it will remove the file.

DougBurke:
@wmclaugh - so, I have just done `python setup.py clean` and `regparser.c` is removed, along with other files (list below). I then did `python setup.py develop` and it did re-create the `regparser.c` file (and matches the change in this PR, along with a change in the line numbering, as I suggest above, and a whole number of changes because I happen to have bison 3.82 rather than 3.0.4). My guess is that `regparser.c` only gets re-built if it is removed, and is not automatically rebuilt by our `configure` code.

However, this is completely pointless as a user who has just downloaded sherpa is not going to do this two step process: I have never done it, it's not in our build instructions, and I argue shouldn't be. So is there a way to get the `regparser.c` file updated to something that means clang/macOS users do not fail the first time they build Sherpa?

There is also the long-term future of the `clean` command, because as I've noted in either #1329 or #1352 it looks like this command is going to be removed from setuptools.

Here's the list of files `python setup.py clean` removes:

```
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.ascii
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/index.html
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/section1.html
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/section2.html
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/section3.html
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/section4.html
	deleted:    extern/fftw-3.3.9/doc/FAQ/fftw-faq.html/section5.html
	deleted:    extern/region-4.14/src/reglexer.c
	deleted:    extern/region-4.14/src/regparser.c
	deleted:    extern/region-4.14/src/regparser.h
```

wmclaugh:
@DougBurke-

I believe the correct fix is to remove the regparser.c file from the source code.  They won't need to run a clean and will generate the file properly.

DougBurke:
I'm guessing we should also remove `reglexer.c` and `regparser.h`?

wmclaugh:
@DougBurke -  Yes, those files should also be removed as part of the proper fix as they will all get generated.

DougBurke:
@wmclaugh - I've re-done the fix, to 

a) fix the bizarre test failure which is completely unrelated and is not really a failure but is just down to warning messages and how we treat them in our testing
b) remove the three files
c) most  importantly, fix up the `.gitignore` file which allowed these files to be added in the first place (and is, I assume, the real reason for #1401 not working for everyone)

Assuming the tests pass, that is!

DougBurke:
@Marie-Terrell - feel free to re-assign this to @harlantc - or reject the review request entirely

sherpa/conftest.py
```diff
@@ -162,6 +162,13 @@ def pytest_collection_modifyitems(config, items):
             r"unclosed file .*/data.dat'.* closefd=True>",
             r"unclosed file .*/model.dat'.* closefd=True>",
             r"unclosed file .*/resid.out'.* closefd=True>",
+
+            # it would be nice if we could find this out from first principles,
+            # rather than only fiding them from random CI runs
```
wmclaugh, line 167:
```suggestion
            # rather than only finding them from random CI runs
```

DougBurke:
I've rebased this to address the typo in a comment noted by @wmclaugh and also because I found another set of problems from the testing code (warning messages) in #1513 so I've included them here (as well as in #1513) just to make the merges easier.

Issue #1501, opened 2022-05-04
Consider allowing type annotations in newly written code


hamogu:
[PEP 484](http://www.python.org/dev/peps/pep-0484) introduced type annotation for functions and methods and [PEP 526](http://www.python.org/dev/peps/pep-0526) added a similar syntax for variables and attributes.

Type annotations at this point are completely optional, Python itself does nothing with that information. However, it 

- provides additional documentation for developers
- enables the use of static type checkers at compile time such as http://mypy-lang.org/
- can be used to generate parts of the docstrings ([implemented in napoleon](https://github.com/sphinx-doc/sphinx/blob/master/doc/usage/extensions/napoleon.rst#type-annotations) and [under discussion in numpydoc](https://github.com/numpy/numpydoc/issues/196))

Numpy has been adding a lot of type annotations to its code in the last few releases as well as appropriate types to describe numpy arrays etc. (https://numpy.org/devdocs/reference/typing.html) 

I suggest that we try to get into the habit of adding type annotations in new code when it's not too hard (i.e. for cases when it's just str, float etc) in the future, but don't spend effort to convert existing code.

DougBurke:
I was just looking at doing this, but I know there are ongoing changes (e.g. simplifications in Python 3.9 or 3.10) so working out what is and isn't appropriate needs some thought. For instance, we have a lot of code that takes objects that we currently hard-code knowledge of rather than rely on duck typing (e.g. something derived from ArithmeticModel) and what is the best solution here (treat as Any, or as ArithmeticModel, or something else).

My particular play-around got as far as something like (I don't guarantee the following is syntactically correct, and definitely no  comment on semantically correct) but then I started reading about `collections.abc` and then working out whether I needed `MutableMapping` instead and realised I had a bug to work on instead!

```
from typing import Mapping, Tuplle
InterpolationScheme = Tuple[TemplateModel, Dict[Any, Any]]
InterpolationMap = Mapping[str, InterpolationScheme]

interpolators : InterpolationMap = {}
```

I also think we'd require that we actually have some way to check that the annotations are sensible (via running the code with whatever-system-makes sense) to make sure we just aren't adding noise.

hamogu:
There is a lot of development still happening on that front: https://docs.python.org/3.11/whatsnew/3.11.html
so it might be too early to embrace; it's worth keeping an eye on though.

DougBurke:
[...]

PR #1525, opened 2022-05-25
Add UI type annotations

DougBurke:
An example of adding type annotations to try out #1501. What is interesting (in this particular work) is what "interfaces" we end up relying on. There are a number of cases where we basically want to say "we return an object that follows this interface" but our current setup isn't designed like this (we don't really use metaclasses to create protocols/interfaces, instead rely on just straight subclasses. Does this suggest that we might want to use a more metaclass approach, or is it worth not breaking things that aren't broken.

I have been using `mypy` to validate this, (I have now tried with `mypy` using python 3.10 and 3.7). However, even with a number of methods annotated I am not convinced this has correctly caught the best type annotations. I expect that there can be some places where annotating a new method/module may cause a number of related changes to existing annotations. There's also some functionality in 3.10 (or later) that may be helpful (eg type guards), that we could think about requiring type_extensions.

One interesting thing to think about from this work, as it deals with the ui layer, is that we have routines like

    def set_model(id, model=None):

(dropping the `self` argument for exposition) which can actually be used as

    set_model(a_model)
    set_model(idval, a_model)

that is, even though the first argument is labelled `id` it is actually somewhat amorphous in its meaning: if only one argument is given it actually means the `model` argument and the default `id` is used, but if two positional arguments are given then the first argument is the `id`. I decided I didn't want to deal with that in my first play around.

This is based on #1524 #1516 #1513 - currently only the last commit is actually part of this PR
 
I will note that `sherpa/ui/uitils.py` will not validate via `mypy` but that's mnot my fault as it is an existing bit of code and it is rather low-level so I don't feel like worrying about it just yet:

```
% mypy sherpa/ui/utils.py 
sherpa/ui/utils.py:185: error: Argument 1 to "constructor" has incompatible type "Callable[[Any, Any], Any]"; expected "Callable[[Union[Tuple[Callable[..., Any], Tuple[Any, ...]], Tuple[Callable[..., Any], Tuple[Any, ...], Optional[Any]]]], Any]"
Found 1 error in 1 file (checked 1 source file)
```

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -180,12 +180,18 @@ def test_id_checks_session(session, setting):
 
 
 @pytest.mark.parametrize("session", [Session, AstroSession])
-@pytest.mark.parametrize("setting", ['cdf', 'energy', 'lr', 'photon', 'pdf', 'scatter', 'trace',
+@pytest.mark.parametrize("setting", ["cdf", "energy", "lr", "photon", "pdf", "scatter", "trace",
                                      "bkg_model", "bkg_source", "bkg_resid", "bkg_ratio",
                                      "bkg_delchi", "bkg_chisqr", "bkg_fit"
                                      ])
 def test_id_checks_session_unexpected(session, setting):
-    """These identifiers are allowed. Should they be?"""
+    """These identifiers are allowed. Should they be?
+
+    This test has morphed to be a test of "expected" labels that are
+    allowed, so the test name is a bit of a mis-nomer, but is left as
+    is.
+
+    """
 
     s = session()
     s.load_arrays(setting, [1, 2], [1, 2])
@@ -198,7 +204,6 @@ def test_id_checks_session_unexpected(session, setting):
 @pytest.mark.parametrize("setting", ['arf', 'bkg', 'bkgchisqr', 'bkgdelchi', 'bkgfit',
                                      'bkgmodel', 'bkgratio', 'bkgresid', 'bkgsource',
                                      'order',
-                                     # "energy", "photon",  these are currently both valid for astro
                                      "astrocompsource", "astrocompmodel", "astrodata",
                                      "astrosource", "astromodel"])
 def test_id_checks_astro_session(session, success, setting):
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -18,6 +18,8 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from io import StringIO
+
 import numpy
 
 import pytest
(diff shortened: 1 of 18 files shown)


DougBurke:
One thing I wanted to look at was the noise this could add to the RTD documentation (although it's not much different to the `help` output, really). Here's

- a routine that hasn't been annotated



- routines that have been annotated

screenshot1
screenshot2
screenshot3
screenshot4

hamogu:
I have to admit I'm not yet convinced that essentially static typing with worth it in general in a code like sherpa, in particular for the ui stuff where `id` can be given or not as first argument. The "Union[xxx, yyy]" also does not look great to the eye - but maybe I just have to get used to it. 

Re - fixing stuff that's not broken: Even if that was nice, we probably have enough to do. 

So, I do wonder (but have no idea really) (1) if there are limited areas where we can profit from this, (2) if we just add it as we fancy, or (3) just don't do that at all.

DougBurke:
Note to self:

we can add support for copying over annotations, so that after `from sherpa.astro.ui import *` we can see the annotations either with `help` or on RTD, but there do appear to be limitations (we have access to the type synonyms that are used, which means these get copied rather than the underlying `List[Union[int, str]]` type it represents). Python 3.10 also changes/simplifies things, so it's not obvious if this works nicely with earlier code.

```
% git diff .
diff --git a/sherpa/utils/__init__.py b/sherpa/utils/__init__.py
index 3890aa04..47ca8b7e 100644
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1138,8 +1138,7 @@ def bool_cast(val):
 
 
 def export_method(meth, name=None, modname=None):
-    """
-    Given a bound instance method, return a simple function that wraps
+    """Given a bound instance method, return a simple function that wraps
     it.  The only difference between the interface of the original
     method and the generated function is that the latter doesn't
     include 'self' in its argument list.  This means that when the
@@ -1157,6 +1156,11 @@ def export_method(meth, name=None, modname=None):
     function.  Note that the caller is responsible for assigning the
     returned function to an appropriate name in the calling scope.
 
+    Annotations are copied over if available. With all the changes in
+    annotation support in Python 3.10 it is not at all obvious if this
+    is sufficient or will work with the currently-supported Python
+    verions.
+
     """
 
     if type(meth) is not MethodType:
@@ -1207,6 +1211,13 @@ def export_method(meth, name=None, modname=None):
                             new_meth.__closure__)
     new_meth.__doc__ = doc
 
+    # Can we support annotations? There are changes in Python 3.10,
+    # but for now just do the simplest thing in the hope it works.
+    #
+    annotations = getattr(meth, "__annotations__", None)
+    if annotations is not None:
+        new_meth.__annotations__ = annotations
+
     return new_meth
 
```

and

```
>> from sherpa.ui.utils import Session
>>> Session.list_data_ids.__annotations__
{'return': 'IdTypes'}
>>> help(Session.list_data_ids)

>>> from sherpa.ui.utils import IdTypes
>>> IdTypes
typing.List[typing.Union[int, str]]
>>> quit()
```

hamogu:
https://numpy.org/devdocs/reference/typing.html

DougBurke:
I'm leaving this open more as a discussion of things we've found out about typing rather than as something to merge. I should probably just take this and put the relevant info into an issue but ...

DougBurke:
As time goes on this is getting more-complex to rebase, and not worth keeping aronud.

PR #1529, opened 2022-05-28
Add arguments to -i in sed calls in config script

hamogu:
In BSD sed (which is the version installed on MacOS) the argument
after -i is MANDATORY. Without it, one gets error messages like

sed: 1: "setup.cfg": unterminated substitute pattern

and no substitution takes place. The underlying cause is that, on that version of sed, the first part of the substitution pattern is read as the argument for `i`, leading to a pattern that is then too short.
See https://stackoverflow.com/questions/28592043/what-is-wrong-with-my-string-substitution-using-sed-on-mac-os-x for further comments.

Now, I hope that this fix does not break that GNU/Linux use now...

### DIFF
--- a/scripts/use_ciao_config
+++ b/scripts/use_ciao_config
@@ -99,4 +99,4 @@ sed -i="" "s|#xspec_lib_dirs = None|xspec_lib_dirs=${CONDA_PREFIX}/lib|" $cfg
 sed -i="" "s|#xspec_version = .*|xspec_version = ${xspec_version}|" $cfg
 
 # Depending on the version of sed, there might be a temporary file. Remove it if it exists.
-rm -r setup.cfg=
+rm -f setup.cfg=
--- a/scripts/use_ciao_config
+++ b/scripts/use_ciao_config
@@ -98,5 +98,5 @@ sed -i="" "s|#xspec_include_dirs = None|xspec_include_dirs=${CONDA_PREFIX}/inclu
 sed -i="" "s|#xspec_lib_dirs = None|xspec_lib_dirs=${CONDA_PREFIX}/lib|" $cfg
 sed -i="" "s|#xspec_version = .*|xspec_version = ${xspec_version}|" $cfg
 
-# The notgnused version makes a temporary file. Remove it if it exists.
+# Depending on the version of sed, there might be a temporary file. Remove it if it exists.
 rm -r setup.cfg=
--- a/scripts/use_ciao_config
+++ b/scripts/use_ciao_config
@@ -69,31 +69,34 @@ fi
 #
 echo "Updating $cfg"
 
-sed -i "s|#install_dir=build|install_dir=${CONDA_PREFIX}|" $cfg
-sed -i "s|#configure=None|configure=None|" $cfg
+sed -i="" "s|#install_dir=build|install_dir=${CONDA_PREFIX}|" $cfg
+sed -i="" "s|#configure=None|configure=None|" $cfg
 
-sed -i "s|#disable-group=True|disable-group=True|" $cfg
-sed -i "s|#disable-stk=True|disable-stk=True|" $cfg
+sed -i="" "s|#disable-group=True|disable-group=True|" $cfg
+sed -i="" "s|#disable-stk=True|disable-stk=True|" $cfg
 
-sed -i "s|#region=local|region=local|" $cfg
-sed -i "s|#region-include_dirs=build/include|region-include_dirs=${CONDA_PREFIX}/include|" $cfg
-sed -i "s|#region-lib-dirs=build/lib|region-lib-dirs=${CONDA_PREFIX}/lib|" $cfg
-sed -i "s|#region-libraries=region|region-libraries=region ascdm|" $cfg
-sed -i "s|#region-use-cxc-parser=False|region-use-cxc-parser=True|" $cfg
+sed -i="" "s|#region=local|region=local|" $cfg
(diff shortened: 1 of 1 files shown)


DougBurke:
Hmm, https://man.openbsd.org/sed suggests the argument to `-i` is optional but highly-recommended. However, I suggest rather than `" "` we just use

> sed -i.org ....

as we don't really care that there's a backup file created. We could delete it at the end as it doesn't really by use anything (as we do multiple calls rather than all-at-once), which would be something like (untested)

```
cfg=setup.cfg
...

bak=.bak

sed -i${bak} "s|...|" $cfg
sed -i${bak} "s|...|" $cfg
...

rm ${cfg}${bak}
```

hamogu:
Unfortunately, the reality is that MacOS is not fully BSD compatible and
requires (not just strong recommends) that argument, so we'll have to
deal with it. I like your suggestions but won't get to it today.Message
ID: ***@***.***>

hamogu:
From the page @dougBurke cited (and that matches the MacOS version of sed): "If a zero length extension is given, no backup will be saved.". So, `-i ""` as I did it in this PR is the portable way of doing exactly what the script is doing now linux (but fails to do on MacOS): Edit the file in place and do not save a backup. Since we don't need a backup (the file is git-controled and you can get it back with `git reset ...` in case funky things happen that usually would warrant a backup instead of in-place editing (disk full, connection ot network valume lost), I stand by this PR as written. I don't think creating a tepoarary backup that we then delete at the end of the script is worth the effort this this.

DougBurke (approved):

hamogu:
Because, it could not be easy: In GNU sed the **cannot** be a space after the i for this case:  `-i ""` gives the error message reported by @DougBurke. However, on the MacOS sed, *there has to be at least one space*, otherwise I get the same "unterminated substitute pattern" that led me to start this PR in the first place. Grr. Update coming...

hamogu:
OK, that works at least for my two systems... I can obviously squash the commits, but wanted to leave them separate for now to make it obvious for review what I've changed this timel.

hamogu:
I think I've finally found a version that gives identical results on linux and Mac. Spend way more time on this that it's worth... Also, I rebased, so hopefully this is done now.

DougBurke (approved):

scripts/use_ciao_config
```diff
@@ -69,31 +69,34 @@ fi
...
+sed -i="" "s|#fftw-include_dirs=build/include|fftw-include_dirs=${CONDA_PREFIX}/include|" $cfg
+sed -i="" "s|#fftw-lib-dirs=build/lib|fftw-lib-dirs=${CONDA_PREFIX}/lib|" $cfg
+sed -i="" "s|#fftw-libraries=fftw3|fftw-libraries=fftw3|" $cfg
+
+sed -i="" "s|#with-xspec=True|with-xspec=True|" $cfg
+sed -i="" "s|#xspec_include_dirs = None|xspec_include_dirs=${CONDA_PREFIX}/include|" $cfg
+sed -i="" "s|#xspec_lib_dirs = None|xspec_lib_dirs=${CONDA_PREFIX}/lib|" $cfg
+sed -i="" "s|#xspec_version = .*|xspec_version = ${xspec_version}|" $cfg
+
+# The notgnused version makes a temporary file. Remove it if it exists.
```
wmclaugh, line 101:
```suggestion
# The not gnu sed version makes a temporary file. Remove it if it exists.
```

scripts/use_ciao_config
```diff
@@ -69,31 +69,34 @@ fi
...
+rm -r setup.cfg=
```
wmclaugh, line 102: Does this need to be a recursive remove?  Also, it would be preferable to check if the file exists before attempting to remove a file.

scripts/use_ciao_config
hamogu, line 101:
```suggestion
# Depending on the version of sed, there might be a temporary file. Remove it if it exists.
```

scripts/use_ciao_config
hamogu: Recursive remove is the only way for a simple rm to not raise an error if the files does not exist. In a makefile, I would have written `-rm setup.cfg=`. I don't think the `-r` is a security risk here, since the name of the file is spelled out explicitly, not wildcards that could accidentally remove other files.

wmclaugh: I believe using the '-f' option will allow the remove command to not error out if the specified file or directory does not exist. Using the -r will still error out.

scripts/use_ciao_config
```diff
@@ -69,31 +69,34 @@ fi
...
+sed -i="" "s|#fftw-lib-dirs=build/lib|fftw-lib-dirs=${CONDA_PREFIX}/lib|" $cfg
+sed -i="" "s|#fftw-libraries=fftw3|fftw-libraries=fftw3|" $cfg
+
+sed -i="" "s|#with-xspec=True|with-xspec=True|" $cfg
+sed -i="" "s|#xspec_include_dirs = None|xspec_include_dirs=${CONDA_PREFIX}/include|" $cfg
+sed -i="" "s|#xspec_lib_dirs = None|xspec_lib_dirs=${CONDA_PREFIX}/lib|" $cfg
+sed -i="" "s|#xspec_version = .*|xspec_version = ${xspec_version}|" $cfg
+
+# Depending on the version of sed, there might be a temporary file. Remove it if it exists.
+rm -r setup.cfg=
```
hamogu, line 102:
```suggestion
rm -f setup.cfg=
```

Issue #1530, opened 2022-05-31
template-model plotting/handling needs a proper review


DougBurke:
I don't know what the template model code is meant to do, but from looking at the code (in this case for #1516 where I add some tests of template-model plotting to at least have some baseline code to test against) it looks like it was never really thought out fully or tested properly. For example

- looks like source and model plots are handled differently

The following lines are not very meaningful unless you've spent time trying to understand the existing UI code, but the main point here is that we have a "source" plot object but not a "model" one (this is where I noticed it in #1516)

 https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/ui/utils.py#L300-L301

- template-model handling

`get_source_component_plot` accesses this field when the model is a `TemplateModel` instance. However, it turns out that it requires a vrey-specific set of s̶k̶i̶l̶l̶s̶ calls to actually hit this bit of code - in particular, calling `load_template_model(..., template_interpolator_name=None)`

https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/ui/utils.py#L11078-L11119

- general wonkiness in the code

From `sherpa.models.template.create_template_model`, there's a code path which would return `None` here, which does not make sense (if the name is not `None` but is not a key of the dictionary). I think I noted this as part of #1516 but maybe it was some other PR I was working on and I haven't tried to address this)

https://github.com/sherpa/sherpa/blob/feffe3565e329162b1dff81e34d6e3a64ab3fe9b/sherpa/models/template.py#L68-L78

Given the above it would not-at-all surprise me to find other issues. For instance, in the tests I added in #1516 it looks like `Data`D` and `Data1DInt` models are evaluated the same (when the `xlo` axis of `Data1DInt` is the same as the `x` axis of `Data1D`) which suggests it either isn't integrating by default or doesn't know how to. I know from my regrid/XSPEC work that interpolation behaves differently for unintegrated and integrated data (do we conserve "flux" or not) and I imagine these issues would be relevant here. In fact, I think the template interpolation support is really just a different version of the regrid support, and we should proabably unify them, but I haven't really thought about it to see if it really makes sense.

PR #1532, opened 2022-06-03
Add plot related tests

DougBurke:
# Summary

Add more tests to cover corner case scenarios (mostly plotting related).

# Details

This is pulled out of #1516 and adds coverage for corner cases. Note that some of these are not needed with the main branch but could be relevant if we change things in the future (such as #1516), but some areas are definitely new (e.g. they show some issues discussed in #1530).

As we have a number of plot-related changes coming in, it is useful to have these corner cases covered sooner rather than later.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -18,6 +18,8 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from io import StringIO
+
 import numpy
 
 import pytest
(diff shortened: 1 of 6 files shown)


sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -406,3 +413,815 @@ def test_save_error_ascii(session, kwargs, idval, tmp_path):
...
+    #
+    s.set_method("levmar")
+
+    out = StringIO()
+    s.show_method(outfile=out)
+
+    toks = out.getvalue().split("\n")
+    assert toks[0] == "Optimization Method: LevMar"
+    assert toks[1] == "name     = levmar"
+    assert toks[2] == "ftol     = 1.1920928955078125e-07"
```
hamogu, line 497: I'm glad this works, but I'm always wary of hardcoding so much precision. It's prone to failing and creating confusion (e.g. if some user runs it on some other hardware or python version) or on totally unrelated changes because some default numpy printing changed. I suggest to check only what is the minimum we expect for this to make sense, e.g. `assert re.match("1.19[0-9]*e-07")`. We want to test to fail is this is much more or less than 1e-7 or of the formatting of the print statement changes significantly, but we don't need it to fail because the 17th digit is either not printed to differs in some way.

While we have some tests that are classic regression tests ("Make sure we manually check every change"), here I would advocate for a setup that's a little more lenient. If this was not a string comparison, we would use `pytest.approx` for the same effect.

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -406,3 +413,815 @@ def test_save_error_ascii(session, kwargs, idval, tmp_path):
...
+
+    # The following is somethiing that will need to be changed
+    # whenever an underlying object has it's string representation
+    # changed, but it should not be hard to do. The main issue is
+    # going to be numeric precision.
+    #
+    toks = out.getvalue().split("\n")
+    assert toks[0] == "Optimization Method: NelderMead"
+    assert toks[1] == "name         = simplex"
+    assert toks[2] == "ftol         = 1.1920928955078125e-07"
```
hamogu, line 574: Same as comment above.

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -406,3 +413,815 @@ def test_save_error_ascii(session, kwargs, idval, tmp_path):
...
+
+    mclass = getattr(sherpa.plot, f"{label.capitalize()}Plot")
+    assert isinstance(mplot, mclass)
+
+    cclass = getattr(sherpa.plot, f"Component{label.capitalize()}Plot")
+    assert isinstance(cplot, cclass)
+
+    # Check it's not a sub-class.
+    #
+    subclass = getattr(sherpa.plot, f"ComponentTemplate{label.capitalize()}Plot")
```
hamogu, line 855: This is an example where it's not clear to me why this is worth testing. Why could this not be a sub-class? (No need to explain this to me, just read on and click "resolve conversation".) However, that's fine for a Regression test. It passes now, so it clearly tests existing behavior. If it ever changes, someone will have to confirm that ti's an intended change.

hamogu (changes requested):
This is basically ready to approve. I admit that I only skimmed over some of the changes and did not check every line. However, I think in this case, that's sufficient. The vast majority of this PR is some simply "Regression test - the intended behavior is unclear". As such, any test that passes now works - since it works, it's obviously checking current behavior. 

My comments highlight an example where I decided that the fact that it tests something and passes now is good enough without me digging deeper *why* this or that happens. I also have one suggestion to make it more robust to running tests on different architectures.

sherpa/astro/ui/tests/test_astro_session.py
DougBurke: It is a regression test. I don't understand the template plotting code and what it should do, so I'm just testing what it does do now and leaving it to #1530 for future discussion.

sherpa/astro/ui/tests/test_astro_session.py
DougBurke: We have quite a few tests that I expect to fail at some point if there's significant numerical differences. I've actually been quite surprised we haven't seen many already. I think that it's in part many of these values (like the 1.19..e-7) value are "special" - e.g.

```
>>> numpy.finfo(numpy.float32).eps
1.1920929e-07
```

and so they don't suffer from "'rounding/..." issues that a value that was calculated would have. There's still the issue of how-many decimal places get displayed, but I think that is either a python or numpy decision (depending on exactly how it's created).

So in this case I don't really have the energy to change these lines until we start seeing a problem with them, but there are other places where we could be more defensive.

hamogu (approved):

hamogu (approved):
Still fine with me.

Issue #1535, opened 2022-06-15
plot_order and orders argument when there's only one response


DougBurke:
It turns out I've added a note about this to the code already, but I noted this when testing #1382.

https://github.com/sherpa/sherpa/blob/300add21a26573ae838fc71e1d56300f99e62d85/sherpa/astro/plot.py#L585-L597

So,  if you have say three responses and then call

```python
plot_order(orders=4)
plot_order(orders=[1, 2, 3, 4])
```

then you will get an error:

```
PlotErr: '4' is not a valid order
```

However, if you only have one response than both commands do not error out, it just treats every order value you send in as `order=1`. I can see an argument for doing this, but I don't think it's a good argument as it just complicates people's understanding of the order.

DougBurke:
So, instead of

```python
     if len(data.response_ids) > 2: 
         if order < 1 or order > len(model.rhs.orders): 
             raise PlotErr('notorder', order)
```

I think we should have

```python
     if order < 1 or order > len(model.rhs.orders): 
         raise PlotErr('notorder', order)
```

(which would then mean the following block would need to be de-indented).

DougBurke:
@anetasie - do you have any problem with me making

```python
plot_order(orders=2)
```

be an error when your PHA dataset only has one response? You can check with CIAO 4.14 and your latest dataset you are fitting and look at

```python
plot_order()
plot_order(orders=2)
plot_order(orders=[1, 2, 3])
```

I clam the latter two should fail, but currently they do not.

anetasie:
@DougBurke this is ok.  Each order should have a separate response. I did not have a chance to try it.

PR #1536, opened 2022-06-15
Tests: add more tests of get_order_plot/plot_order

DougBurke:
# Summary

Add a test for multiple orders and plot_order.

# Details

Noted in #1382 is that some code in plot_order was not being
evaluated. So

a) make sure we have a plot_order() call with multiple responses,
   not just get_order_plot(), to ensure the plot method is
   called.

   There is no check the plot does anything actually sensible,
   but the assumption is that the get_order_plot checks make sure
   the data is at least sensible.

b) we can run the get_order_plot calls without a plotting backend,
   so do so

c) add checks for what happens if the orders value is invalid; it
   turns out - as I've noted in a TODO comment in the code - that
   the behavior depends on whether the data contains 1 or more
   responses, as discussed at #1535.

   So, the behavior of test_get_plot_order_knows_bad_response_single
   should be thought of as a regression test, as I believe it should
   fail like test_get_plot_order_knows_bad_response_multiple does

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_multi_response.py
@@ -1,5 +1,6 @@
 #
-#  Copyright (C) 2019, 2020  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2019, 2020, 2022
+#  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
I believe this test would have shown up the issue I noted in https://github.com/sherpa/sherpa/pull/1382#discussion_r898195543

hamogu (approved):

PR #1538, opened 2022-06-21
Ensure chi2xspecvar errors match XSPEC when 0 counts are present during background subtraction

### LINKED ISSUE #356: chi2xspecvar does not match XSPEC for bg-subtracted data if src or bg has 0 counts in a channel
I took the `3c273.pi` and `3c273_bg.pi` files from our test suite and edited them so they both had BACKSCAL=1.0, EXPOSURE=1.0, and had no grouping channels. I then compared the errors calculated by XSPEC 12.9.1 and Sherpa in CIAO 4.9 (essentially the same as the current master branch @aa98d19553d1e62d3c65ba5abf33562e5e7dbe4c) with (in Sherpa)

    set_analysis('channels')
    subtract()
    set_stat('chi2xspecvar')

The only channels where the error values were different had the following pattern (there were multiple channels with some of these combination of source and background coutns):

nsrc | nbgnd | Var(sherpa) | Var(xspec)
---- | ----- | ------- | ------
 5  | 0  | 6.0 | 5.0
 7  | 0  | 8.0 | 7.0
 9  | 0 | 10.0 | 9.0
 4  | 0 | 5.0 | 4.0
 0  | 1 | 2.0 | 1.0
 0  | 0 | 2.0 | 1.0
 0  | 2 | 3.0 | 2.0
 3  | 0 | 4.0 | 3.0
 2  | 0 | 3.0 | 2.0
 1  | 0 | 2.0 | 1.0
 6  | 0 | 7.0 | 6.0
10  | 0 | 11.0 | 10.0
 8  | 0 | 9.0  | 8.0
---- ----- ------- ------

`nsrc` and `nbgnd` are the number of counts in the source and background channel, and `Var(x)` is the variance calculated by x. We can see that 

    Var(sherpa) = 1 + Var(xspec)

so it looks like XSPEC is "dropping" the 0-count channel when calculating the errors (apart from when both channels are 0, when the variance is set to 1.0).

At some level all bets are off when you have 0 counts using Gaussian statistics - XSPEC provides this warning:

```
***Warning: Chi-square may not be valid due to bins with zero variance
            in spectrum number(s): 1 
```

but do we want to agree with what XSPEC does here when using `chi2xspecvar`?

**EDIT** I've come to realize that we really should make sure people don't use `chi2xspecvar` when fitting the model to the background (rather than subtracting it), as we don't have a way to replicate what XSPEC does in this case, because XSPEC doesn't actually do this case. But that's a separate issue - #1543

DougBurke:
# Summary

When using the chi2xspecvar statistic for estimating PHA errors, ensure that the errors match those calculated by XSPEC when the background is subtracted and the source or background group contains 0 counts (previously the error for that bin may have been larger than XSPEC calculates).

# Details

Note that, as mentioned in #1543, this does not address the issue of what to do when fitting the background with the chi2xspecvar statistic, but that's because we don't have an equivalent in XSPEC to copy for that case.

I also hope that this is a temporary change, and we can take time to redesign the code to avoid some of the design issues, but I do not want to wait for that to wait for this to land because we have left this bug open for too long, and it is not-at-all-obvious we know what the design should be. As @hamogu has mentioned, IXPE requires combining datasets to get the errors, and that is a different, but highly-relevant, design constraint we'd need to work through.

## Problem

The chi2xspecvar statistic is meant to match XSPEC, but as pointed out in #356 it generated "too large" errors if either (or both) the source and background group contains 0 counts and the background is subtracted. This is because the code was converting 0 counts to an error of 1 and then combining this with the error from the other component to get the overall error, so for instance

```python
src_counts = 4
bkg_counts = 0
````

would then (if we assume the source and background regions have the same exposure time, backscal, and areascal values) we would get from `chi2xspecvar`

```python
src_variance = 4
bkg_variance = 1
```

and hence

```python
variance = 5
error = sqrt(5)
```

However, XSPEC just drops the 0-count channel in this case, so the "correct" error should be

```python
error = 2
```

The handling of when both source and background components have 0 counts is technically the same, it is just that we have to care about which of the components to drop, and that depends on the relative scaling values (exposure, backscal, and areascal).

## Solution

Previous attempts to fix this have looked at changing the stats code, that is the `sherpa.stats.Stat.calc_staterror` method, but this is not designed to handle source and background data, because this is only relevant for PHA data.

The approach I've taken is more restrained, in that it recognizes this is only an issue for PHA data, and so the changes are made to the `sherpa.astro.data.DataPHA.get_staterror` method. This actually makes sense, because it is in this method that we deal with the complications of background subtraction, in particular handle the scaling values we need to worry about for the "source and background are 0" case. However, it is not ideal from the sense of "mixing responsibilities", but I do not see a way to do this in any reasonable time scale (given how long #356 has remained), and I think we are going to have to look at the statistic handing if we want to handle IXPE data properly, and that will likely require quite a large, somewhat different, change so I don't think it's worth making a large design change if we are only going to have to change it again for something else.

I've validated the changes by

1.  seeing that the existing tests I added in #768 now pass when they didn't before
3. realized that those tests aren't sufficient (as they didn't really handle the different scaling regimes) by adding more cases to that test
4. realized that I needed a test with a variable scaling function, so added a test with an array for BACKSCAL

These tests are based on creating a PHA source and background dataset, load them into XSPEC, call `iplot data` and then `wdata`, and note down the last column, which is the error value. Note that if you do this the error XSPEC returns is a rate, and we are checking the counts, so you need to scale by the exposure time (I forgot this at some point and wasted time trying to understand why I was seeing differences).

I've also discussed this with the XSPEC developers who pointed me to the relevant code in XSPEC, which I've referenced in the code. One thing the XSPEC code assumes is that there's only one background, so I've made assumptions on what to do if you have multiple backgrounds (basically that we just average them).

## Notes

Other than tests, the first couple of commits are code clean ups, including removing some attempt to handle the chi2datavar routine that uses the same idea I use (i.e. the change is made to the `DataPHA.get_staterror` method rather than the lower-level statistic classes). However, it is not clear what the code is meant to be doing because

1. we don't have the problem with 0 counts that the code seems to think we do
2. it only triggered if the user was dong something unusual (i.e. they had passed in the low-level method rather than the `chi2datavar` object)

This may in part be because of changes in #153 - which meant that it can handle 0-count bins - but it really is not obvious what the code was meant to be doing and it wasn't being triggered in our tests,

As noted in the commit for 05b4faa93055aa57b84ea331c0bd4ac626b0acce we only trigger this behavior if the user is using "chi2xsecvar", with the check

```python
if staterrfunc == Chi2XspecVar.calc_staterror ...
```

Technically we could say something like

```python
if staterrfunc in [Chi2XspecVar.calc_staterror, sherpa.stats._statfcts.calc_chi2xspecvar_errors  ] ...
```  

as a user could call it with the low-level method, but I elected not to do this as it is not "meant" to be how the method is called.

## Examples

Here is one of the test examples (all channels have 0 counts and we have a variable BACKSCAL)

```
sherpa-4.14.0> data.subtracted
True

sherpa-4.14.0> print(data.counts)
[0. 0. 0. 0.]

sherpa-4.14.0> print(data.get_background().counts)
[0. 0. 0. 0.]

sherpa-4.14.0> print(data.exposure, data.backscal)
100.0 [0.1 0.2 2.  5. ]

sherpa-4.14.0> print(data.get_background().exposure, data.get_background().backscal)
125.0 [ 0.1  0.1  4.  20. ]

sherpa-4.14.0> data.get_staterror()

sherpa-4.14.0> data.get_staterror(staterrfunc=_session._stats["chi2datavar"].calc_staterror)
array([0., 0., 0., 0.])

sherpa-4.14.0> data.get_staterror(staterrfunc=_session._stats["chi2xspecvar"].calc_staterror)
array([1.28062485, 1.88679623, 1.07703296, 1.0198039 ])
```

So the xspecvar statistic returns 1 for each bin in CIAO 4.14. With this PR we get the errors:

```
>>> ui.get_data().get_staterror(staterrfunc=Chi2XspecVar().calc_staterror)
array([0.8, 1. , 0.4, 0.2])
```

I was too lazy to create a fake response so I could just call `set_stat("chi2xspecvar")` and then `calc_stat()` to show the difference, but it would have done (as this is the way users are going to see the differences).

## Worked example

The `sherpa-test-data/sherpatest/qso.pi` file has no grouping, so we can pick a small range where there's some zero's (e.g. 5-6 keV):

```
xspec

		XSPEC version: 12.12.1
	Build Date/Time: Thu Apr 28 18:10:42 2022

XSPEC12>data qso.pi

1 spectrum  in use
 
Spectral Data File: qso.pi  Spectrum 1
Net count rate (cts/s) for Spectrum:1  1.227e-01 +/- 2.040e-03 (99.2 % total)
 Assigned to Data Group 1 and Plot Group 1
  Noticed Channels:  1-1024
  Telescope: CHANDRA Instrument: ACIS  Channel Type: PI
  Exposure Time: 2.98e+04 sec
 Using fit statistic: chi
 Using Background File                qso_bkg.pi
  Background Exposure Time: 2.98e+04 sec
 Using Response (RMF) File            qso.wrmf for Source 1
 Using Auxiliary Response (ARF) File  qso.warf

XSPEC12>ignore 1-342,412-1024
   342 channels (1,342) ignored in spectrum #     1
   613 channels (412,1024) ignored in spectrum #     1

XSPEC12>mo po

              1       0.01(      0.01)         -3         -2          9         10
1:powerlaw:PhoIndex>1.7
2:powerlaw:norm>3e-4

Model powerlaw<1> Source No.: 1   Active/On
 par  comp
   1    1   powerlaw   PhoIndex            1.70000      +/-  0.0          
   2    1   powerlaw   norm                3.00000E-04  +/-  0.0          


Fit statistic  : Chi-Squared                  303.00     using 69 bins.

***Warning: Chi-square may not be valid due to bins with zero variance
            in spectrum number: 1

Test statistic : Chi-Squared                  303.00     using 69 bins.

***Warning: Chi-square may not be valid due to bins with zero variance
            in spectrum number(s): 1 

 Null hypothesis probability of 9.85e-32 with 67 degrees of freedom
 Current data and model not fit yet.
```

So, XSPEC has a chi-square of 303.00 for these bins.

With CIAO 4.14 we get

```
sherpa
Sherpa 4.14.0

Python 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:25:59) 
IPython 8.2.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa

sherpa-4.14.0> load_pha("qso.pi")
read ARF file qso.warf
read RMF file qso.wrmf
read ARF (background) file qso_bkg.warf
read RMF (background) file qso_bkg.wrmf
read background file qso_bkg.pi

sherpa-4.14.0> subtract()

sherpa-4.14.0> set_source(xspowerlaw.pl); pl.phoindex = 1.7; pl.norm = 3e-4

sherpa-4.14.0> set_analysis("channel"); ignore(1, 342); ignore(412, 1024)

sherpa-4.14.0> set_stat("chi2xspecvar"); calc_stat()
60.15435798809412

sherpa-4.14.0> set_stat("chi2datavar"); calc_stat()
147.85824365372247

sherpa-4.14.0> set_analysis("energy"); plot_data(alpha=0.5); plot_bkg(alpha=0.5, overplot=True)
```

Here's the image:

sherpa

The blue are background subtracted, but the background is 0 apart from a few bins (one of which is one where the source=0 but bgnd>0 so the result is negative). Anyway, this shows we have 0's to worry about (including, I think, one case where both source and background are zero).

Here's with this PR (the `chi2datavar` results are just to show that hasn't changed from CIAO 4.14, and `powlaw1d` is used rather than `xspowerlaw` as I don't have XSPEC built for this build, but I don't believe that will cause the small statistic difference I mention below):

```
python
Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0] on linux
>>> load_pha("sherpa-test-data/sherpatest/qso.pi")
read ARF file sherpa-test-data/sherpatest/qso.warf
read RMF file sherpa-test-data/sherpatest/qso.wrmf
read ARF (background) file sherpa-test-data/sherpatest/qso_bkg.warf
read RMF (background) file sherpa-test-data/sherpatest/qso_bkg.wrmf
read background file sherpa-test-data/sherpatest/qso_bkg.pi
>>> subtract()
>>> set_analysis("channel"); ignore(1, 342); ignore(412, 1024)
>>> set_source(powlaw1d.pl); pl.gamma = 1.7; pl.ampl = 3e-4
>>> set_stat("chi2xspecvar"); calc_stat()
303.08668444575886
>>> set_stat("chi2datavar"); calc_stat()
147.85824365372272
>>> set_stat("chi2xspecvar")
>>> calc_stat_info()
Dataset               = 1
Statistic             = chi2xspecvar
Fit statistic value   = 303.087
Data points           = 69
Degrees of freedom    = 67
Probability [Q-value] = 9.52981e-32
Reduced statistic     = 4.52368
```

So we get a statistic of 303.087. It is not quite the 303.00 that XSPEC reports, but is significantly closer than the 60.2 we got in CIAO 4.14.

### DIFF
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -1,6 +1,6 @@
 #
-#  Copyright (C) 2009, 2015, 2016, 2017, 2018, 2019, 2020
-#             Smithsonian Astrophysical Observatory
+#  Copyright (C) 2009, 2015, 2016, 2017, 2018, 2019, 2020, 2022
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -793,8 +793,11 @@ class Chi2XspecVar(Chi2):
     bin.
 
     The calculation of the variance is the same as `Chi2DataVar`
-    except that if the number of counts in a bin is less than 1
-    then the variance for that bin is set to 1.
+    except that if the number of counts in a bin is less than 1 then
+    the variance for that bin is set to 1. Note that this does not
+    handle background subtraction, so to match XSPEC extra logic is
+    present in sherpa.astro.data.DataPHA.get_staterror to handle that
+    case.
 
     See Also
     --------
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -2147,7 +2147,8 @@ class Session(NoNewAttributesAfterInit):
            data.
 
         chi2datavar
-           Chi-squared with data variance.
+           Chi-squared with data variance. If the data has 0 counts then
+           the error for that bin is 0.
 
         chi2gehrels
            Chi-squared with gehrels method [2]_. This is the default method.
@@ -2156,8 +2157,13 @@ class Session(NoNewAttributesAfterInit):
            Chi-squared with model amplitude variance.
 
         chi2xspecvar
-           Chi-squared with data variance (XSPEC-style, variance = 1.0
-           if data less than or equal to 0.0).
+           Chi-squared with data variance to match XSPEC. Errors from
+           zero-count channels (source or background) are ignored if
+           the other channel (background or source) contains counts,
+           or replaced by a minimum value (when both source or
+           background are empty). It should not be used when a model
+           is fit to the background rather than the background is
+           subtracted from the data.
 
         cstat
            A maximum likelihood function (the XSPEC implementation of
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3468,123 +3468,92 @@ must be an integer.""")
         >>> dy = dset.get_staterror(staterrfunc=stat.calc_staterror)
 
         """
-        staterr = self.staterror
 
         filter = bool_cast(filter)
         if filter:
-            staterr = self.apply_filter(staterr, self._sum_sq)
+            filterfunc = self.apply_filter
         else:
-            staterr = self.apply_grouping(staterr, self._sum_sq)
+            filterfunc = self.apply_grouping
 
-        # The source AREASCAL is not applied here, but the
-        # background term is.
+        # This is pulled out into a separate local routine to point
+        # out the same logic is used for the source and background
+        # data objects (and that we always use the source filtering
+        # and grouping, not the background, when calling filterfunc).
         #
-        if (staterr is None) and (staterrfunc is not None):
-            cnts = self.counts
+        def get_error(dataobj):
+            if dataobj.staterror is None:
+                if staterrfunc is None:
+                    return None
 
-            if filter:
-                cnts = self.apply_filter(cnts)
-            else:
-                cnts = self.apply_grouping(cnts)
+                cnts = filterfunc(dataobj.counts)
+                return staterrfunc(cnts)
 
-            staterr = staterrfunc(cnts)
+            return filterfunc(dataobj.staterror,
+                              groupfunc=dataobj._sum_sq)
 
-            # Need to apply the area scaling to the calculated
-            # errors. Grouping and filtering complicate this; is
-            # _middle the best choice here?
-            #
-            """
-            area = self.areascal
-            if staterr is not None and area is not None:
-                if numpy.isscalar(area):
-                    area = numpy.zeros(self.channel.size) + area
-
-                # TODO: replace with _check_scale?
-                if filter:
-                    area = self.apply_filter(area, self._middle)
-                else:
-                    area = self.apply_grouping(area, self._middle)
-                staterr = staterr / area
-
-            """
-
+        staterr = get_error(self)
         if staterr is None:
             return None
 
         if not self.subtracted:
             return staterr
 
-        bkg_staterr_list = []
-
+        # For each background dataset we filter and/or group the
+        # errors to match the source dataset, and then apply the
+        # various scaling factors (areascal, backscal, and exposure
+        # scaling) to correct them to match the source dataset.  The
+        # per-group variance is then stored, so that it can be
+        # combined to create the overall background contribution,
+        # which can then have the "source" side of the scaling values
+        # applied to it before being added to the source erroor term.
+        #
+        bkg_variances = []
         for key in self.background_ids:
             bkg = self.get_background(key)
-            berr = bkg.staterror
-            if filter:
-                berr = self.apply_filter(berr, self._sum_sq)
-            else:
-                berr = self.apply_grouping(berr, self._sum_sq)
-
-            if (berr is None) and (staterrfunc is not None):
-                bkg_cnts = bkg.counts
-                if filter:
-                    bkg_cnts = self.apply_filter(bkg_cnts)
-                else:
-                    bkg_cnts = self.apply_grouping(bkg_cnts)
-
-                berr = staterrfunc(bkg_cnts)
-
-            # This case appears when the source dataset has an error
-            # column and at least one of the background(s) do not.
-            # Because the staterr is not None and staterrfunc is, I think
-            # we should return None.  This way the user knows to call with
-            # staterrfunc next time.
+            berr = get_error(bkg)
             if berr is None:
+                # We do not know how to generate an error, so
+                # return None. An alternative would be to raise an
+                # error, since we have an error from the source.
+                #
                 return None
 
-            bksl = bkg.backscal
-            if bksl is not None:
-                bksl = self._check_scale(bksl, filter=filter)
+            if bkg.backscal is not None:
+                bksl = self._check_scale(bkg.backscal, filter=filter)
                 berr = berr / bksl
 
-            # Need to apply filter/grouping of the source dataset
-            # to the background areascal, so can not just say
-            #   area = bkg.get_areascal(filter=filter)
-            #
-            area = bkg.areascal
-            if area is not None:
-                area = self._check_scale(area, filter=filter)
+            if bkg.areascal is not None:
+                area = self._check_scale(bkg.areascal, filter=filter)
                 berr = berr / area
 
             if bkg.exposure is not None:
                 berr = berr / bkg.exposure
 
-            berr = berr * berr
-            bkg_staterr_list.append(berr)
+            bkg_variances.append(berr * berr)
 
-        nbkg = len(bkg_staterr_list)
+        nbkg = len(bkg_variances)
         if nbkg == 1:
-            bkgsum = bkg_staterr_list[0]
+            bkgvar = bkg_variances[0]
         else:
-            bkgsum = sum(bkg_staterr_list)
+            bkgvar = sum(bkg_variances)
 
-        bscal = self.backscal
-        if bscal is not None:
-            bscal = self._check_scale(bscal, filter=filter)
-            bkgsum = (bscal * bscal) * bkgsum
+        if self.backscal is not None:
+            bscal = self._check_scale(self.backscal, filter=filter)
+            bkgvar = (bscal * bscal) * bkgvar
 
         # Correct the background counts by the source AREASCAL
         # setting. Is this correct?
-        ascal = self.areascal
-        if ascal is not None:
-            ascal = self._check_scale(ascal, filter=filter)
-            bkgsum = (ascal * ascal) * bkgsum
+        #
+        if self.areascal is not None:
+            ascal = self._check_scale(self.areascal, filter=filter)
+            bkgvar = (ascal * ascal) * bkgvar
 
         if self.exposure is not None:
-            bkgsum = (self.exposure * self.exposure) * bkgsum
+            bkgvar = (self.exposure * self.exposure) * bkgvar
 
         nbkg = SherpaFloat(nbkg)
-        staterr = staterr * staterr + bkgsum / (nbkg * nbkg)
-        return numpy.sqrt(staterr)
+        statvar = staterr * staterr + bkgvar / (nbkg * nbkg)
+        return numpy.sqrt(statvar)
 
     def get_syserror(self, filter=False):
         """Return any systematic error.
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -3505,98 +3505,95 @@ must be an integer.""")
                 else:
                     area = self.apply_grouping(area, self._middle)
                 staterr = staterr / area
+
             """
 
-        if (staterr is not None) and self.subtracted:
-            bkg_staterr_list = []
+        if staterr is None:
+            return None
 
-            # for bkg in self._backgrounds.values():
-            for key in self.background_ids:
-                bkg = self.get_background(key)
-                berr = bkg.staterror
-                if filter:
-                    berr = self.apply_filter(berr, self._sum_sq)
-                else:
-                    berr = self.apply_grouping(berr, self._sum_sq)
+        if not self.subtracted:
+            return staterr
 
-                if (berr is None) and (staterrfunc is not None):
-                    bkg_cnts = bkg.counts
-                    if filter:
-                        bkg_cnts = self.apply_filter(bkg_cnts)
-                    else:
-                        bkg_cnts = self.apply_grouping(bkg_cnts)
+        bkg_staterr_list = []
 
-                    # TODO: shouldn't the following logic be somewhere
-                    #       else more general?
-                    if hasattr(staterrfunc, '__name__') and \
-                       staterrfunc.__name__ == 'calc_chi2datavar_errors' and \
-                       0.0 in bkg_cnts:
-                        mask = (numpy.asarray(bkg_cnts) != 0.0)
-                        berr = numpy.zeros(len(bkg_cnts))
-                        berr[mask] = staterrfunc(bkg_cnts[mask])
-                    else:
-                        berr = staterrfunc(bkg_cnts)
-
-                # FIXME: handle this
-                # assert (berr is not None)
-
-                # This case appears when the source dataset has an error
-                # column and at least one of the background(s) do not.
-                # Because the staterr is not None and staterrfunc is, I think
-                # we should return None.  This way the user knows to call with
-                # staterrfunc next time.
-                if berr is None:
-                    return None
-
-                bksl = bkg.backscal
-                if bksl is not None:
-                    bksl = self._check_scale(bksl, filter=filter)
-                    berr = berr / bksl
-
-                # Need to apply filter/grouping of the source dataset
-                # to the background areascal, so can not just say
-                #   area = bkg.get_areascal(filter=filter)
-                #
-                area = bkg.areascal
-                if area is not None:
-                    area = self._check_scale(area, filter=filter)
-                    berr = berr / area
-
-                if bkg.exposure is not None:
-                    berr = berr / bkg.exposure
-
-                berr = berr * berr
-                bkg_staterr_list.append(berr)
-
-            nbkg = len(bkg_staterr_list)
-            assert nbkg > 0  # TODO: should we remove this assert?
-            if nbkg == 1:
-                bkgsum = bkg_staterr_list[0]
+        for key in self.background_ids:
+            bkg = self.get_background(key)
+            berr = bkg.staterror
+            if filter:
+                berr = self.apply_filter(berr, self._sum_sq)
             else:
-                bkgsum = sum(bkg_staterr_list)
+                berr = self.apply_grouping(berr, self._sum_sq)
 
-            bscal = self.backscal
-            if bscal is not None:
-                bscal = self._check_scale(bscal, filter=filter)
-                bkgsum = (bscal * bscal) * bkgsum
+            if (berr is None) and (staterrfunc is not None):
+                bkg_cnts = bkg.counts
+                if filter:
+                    bkg_cnts = self.apply_filter(bkg_cnts)
+                else:
+                    bkg_cnts = self.apply_grouping(bkg_cnts)
 
-            # Correct the background counts by the source AREASCAL
-            # setting. Is this correct?
-            ascal = self.areascal
-            if ascal is not None:
-                ascal = self._check_scale(ascal, filter=filter)
-                bkgsum = (ascal * ascal) * bkgsum
+                # TODO: shouldn't the following logic be somewhere
+                #       else more general?
+                if hasattr(staterrfunc, '__name__') and \
+                   staterrfunc.__name__ == 'calc_chi2datavar_errors' and \
+                   0.0 in bkg_cnts:
+                    mask = (numpy.asarray(bkg_cnts) != 0.0)
+                    berr = numpy.zeros(len(bkg_cnts))
+                    berr[mask] = staterrfunc(bkg_cnts[mask])
+                else:
+                    berr = staterrfunc(bkg_cnts)
 
-            if self.exposure is not None:
-                bkgsum = (self.exposure * self.exposure) * bkgsum
+            # This case appears when the source dataset has an error
+            # column and at least one of the background(s) do not.
+            # Because the staterr is not None and staterrfunc is, I think
+            # we should return None.  This way the user knows to call with
+            # staterrfunc next time.
+            if berr is None:
+                return None
 
-            nbkg = SherpaFloat(nbkg)
+            bksl = bkg.backscal
+            if bksl is not None:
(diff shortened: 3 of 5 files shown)


sherpa/stats/__init__.py
```diff
@@ -793,8 +793,11 @@ class Chi2XspecVar(Chi2):
     bin.
 
     The calculation of the variance is the same as `Chi2DataVar`
-    except that if the number of counts in a bin is less than 1
-    then the variance for that bin is set to 1.
+    except that if the number of counts in a bin is less than 1 then
+    the variance for that bin is set to 1. Note that this does not
+    handle background subtraction, so to match XSPEC extra logic is
```
anetasie, line 798: This statement is a bit misleading. This statistics is for background subtracted data and the only issue is that for the bins with 0 counts this will use this ad hoc correction. Maybe we do need to stress that for low counts data there should be no subtraction and the applied statistics should be cstat or wstat.

sherpa/ui/utils.py
```diff
@@ -2156,8 +2157,13 @@ def set_stat(self, stat):
            Chi-squared with model amplitude variance.
 
         chi2xspecvar
-           Chi-squared with data variance (XSPEC-style, variance = 1.0
-           if data less than or equal to 0.0).
+           Chi-squared with data variance to match XSPEC. Errors from
+           zero-count channels (source or background) are ignored if
```
anetasie, line 2161: Are you really ignoring the errors in zero-count channels? There are non-zero errors in these channels when I run it with this branch build.

```
In [92]: subtract()

In [93]: calc_stat_info()
Dataset               = 1
Statistic             = chi2xspecvar
Fit statistic value   = 120.547
Data points           = 136
Degrees of freedom    = 134
Probability [Q-value] = 0.791007
Reduced statistic     = 0.899604

In [94]: unsubtract()

In [95]: calc_stat_info()
Dataset               = 1
Statistic             = chi2xspecvar
Fit statistic value   = 115.964
Data points           = 136
Degrees of freedom    = 134
Probability [Q-value] = 0.867406
Reduced statistic     = 0.865404
```
Is the errors in the plot_data() calculated for the zero counts channel? Maybe that's what confuses me, as I have been looking at get_data_plot().yerr which changes depending on the subtraction. It is 1 for the non-subtracted data when using `chi2xspecvar`

anetasie:
Here is the comparison of the data plots in XSPEC and Sherpa.  Sherpa shows the errors of 1  for the channels with 0 counts and XSPEC ignores the errors there.  This is a display choice, so probably should be noted as an issue and treated in a different PR.

image

image

sherpa/stats/__init__.py
DougBurke: Any statement here is mis-leading. However, this class, as written, **DOES NOT** handle background subtraction. We only use this class to calculate the errors when given n counts. There is no concept of background subtraction here (and that is, in part, one of the issues, but it is a separate issue). So, all I want to say is that using Chi2XspecVar with background subtracted data relies on other logic - which is in sherpa.asto.data.DataPHA.get_staterror - to get the correct answer.

I was going to say that most people are not going to see this text, but unfortunately anyone who has done 

```python
set_stat('chi2xspecvar')
```

then will see this information with

```python
get_stat()
```

sherpa/ui/utils.py
DougBurke: It does (at least when the other component is not zero). If it didn't, I wouldn't have asked for review because it would have been wrong.

I don't understand what the output you included here means.

DougBurke: You should be able to use `get_staterror` to look at the errors, although you probably want to add `filter=True`

```
sherpa-4.14.0> load_pha("sherpa-test-data/sherpatest/qso.pi")
read ARF file sherpa-test-data/sherpatest/qso.warf
read RMF file sherpa-test-data/sherpatest/qso.wrmf
read ARF (background) file sherpa-test-data/sherpatest/qso_bkg.warf
read RMF (background) file sherpa-test-data/sherpatest/qso_bkg.wrmf
read background file sherpa-test-data/sherpatest/qso_bkg.pi

sherpa-4.14.0> notice(5, 6)

sherpa-4.14.0> set_stat("chi2xspecvar")

sherpa-4.14.0> get_staterror(filter=True)
array([1.41421356, 1.41421356, 2.        , 1.41421356, 2.44948974,
       2.        , 1.73205081, 1.41421356, 2.        , 1.41421356,
       2.44948974, 1.41421356, 1.73205081, 1.        , 1.41421356,
       1.73205081, 1.        , 1.73205081, 2.64575131, 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 2.        ,
       1.        , 1.        , 1.41421356, 1.        , 1.41421356,
       2.44948974, 2.23606798, 1.        , 1.41421356, 1.41421356,
       1.        , 1.41421356, 1.73205081, 2.        , 1.73205081,
       2.        , 1.73205081, 2.        , 1.        , 1.        ,
       1.        , 1.        , 1.41421356, 2.23606798, 1.        ,
       1.41421356, 1.41421356, 2.23606798, 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.        , 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 1.73205081,
       1.73205081, 1.        , 1.41421356, 2.23606798])

sherpa-4.14.0> get_staterror(filter=True, bkg_id=1)
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1.])
```

You can compare to the number of counts in the source and background components with `get_dep`

```
sherpa-4.14.0> get_dep(filter=True)
array([2., 2., 4., 2., 6., 4., 3., 2., 4., 2., 6., 2., 3., 0., 2., 3., 0.,
       3., 7., 2., 2., 2., 2., 4., 4., 1., 1., 2., 1., 2., 6., 5., 0., 2.,
       2., 1., 2., 3., 4., 3., 4., 3., 4., 1., 1., 1., 1., 2., 5., 0., 2.,
       2., 5., 4., 2., 2., 2., 1., 4., 2., 2., 2., 2., 4., 3., 3., 1., 2.,
       5.])

sherpa-4.14.0> get_dep(filter=True, bkg_id=1)
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 1., 1., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0.])
```

You can see here the background is either 0 or 1 counts and chi2xspecvar returns an error of 1 for all of them (which is what it's meant to do).

You can subtract and see the result

```
sherpa-4.14.0> subtract()

sherpa-4.14.0> get_staterror(filter=True)
array([1.43782764, 1.43782764, 2.0167668 , 1.43782764, 2.4631988 ,
       2.0167668 , 1.75138469, 1.43782764, 2.0167668 , 1.43782764,
       2.4631988 , 1.43782764, 1.75138469, 1.03312551, 1.43782764,
       1.75138469, 1.03312551, 1.75138469, 2.65844848, 1.43782764,
       1.43782764, 1.43782764, 1.43782764, 2.0167668 , 2.0167668 ,
       1.03312551, 1.03312551, 1.43782764, 1.03312551, 1.43782764,
       2.4631988 , 2.25107715, 1.03312551, 1.43782764, 1.43782764,
       1.03312551, 1.43782764, 1.75138469, 2.0167668 , 1.75138469,
       2.0167668 , 1.75138469, 2.0167668 , 1.03312551, 1.03312551,
       1.03312551, 1.03312551, 1.43782764, 2.25107715, 1.03312551,
       1.43782764, 1.43782764, 2.25107715, 2.0167668 , 1.43782764,
       1.43782764, 1.43782764, 1.03312551, 2.0167668 , 1.43782764,
       1.43782764, 1.43782764, 1.43782764, 2.0167668 , 1.75138469,
       1.75138469, 1.03312551, 1.43782764, 2.25107715])
```

If I use this PR then I get (the `get_staterror` values should be the same before the `subtract` call but I haven't checked)

```
>>> load_pha("sherpa-test-data/sherpatest/qso.pi")
read ARF file sherpa-test-data/sherpatest/qso.warf
read RMF file sherpa-test-data/sherpatest/qso.wrmf
read ARF (background) file sherpa-test-data/sherpatest/qso_bkg.warf
read RMF (background) file sherpa-test-data/sherpatest/qso_bkg.wrmf
read background file sherpa-test-data/sherpatest/qso_bkg.pi
>>> notice(5, 6)
>>> set_stat("chi2xspecvar")
>>> get_staterror(filter=True)
array([1.41421356, 1.41421356, 2.        , 1.41421356, 2.44948974,
       2.        , 1.73205081, 1.41421356, 2.        , 1.41421356,
       2.44948974, 1.41421356, 1.73205081, 1.        , 1.41421356,
       1.73205081, 1.        , 1.73205081, 2.64575131, 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 2.        ,
       1.        , 1.        , 1.41421356, 1.        , 1.41421356,
       2.44948974, 2.23606798, 1.        , 1.41421356, 1.41421356,
       1.        , 1.41421356, 1.73205081, 2.        , 1.73205081,
       2.        , 1.73205081, 2.        , 1.        , 1.        ,
       1.        , 1.        , 1.41421356, 2.23606798, 1.        ,
       1.41421356, 1.41421356, 2.23606798, 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.        , 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 1.73205081,
       1.73205081, 1.        , 1.41421356, 2.23606798])
>>> get_staterror(filter=True, bkg_id=1)
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1.])
>>> subtract()
>>> get_staterror(filter=True)
array([1.41421356, 1.41421356, 2.        , 1.41421356, 2.44948974,
       2.        , 1.73205081, 1.41421356, 2.        , 1.41421356,
       2.44948974, 1.41421356, 1.73205081, 0.25951556, 1.41421356,
       1.73205081, 0.25951556, 1.73205081, 2.64575131, 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 2.        ,
       1.        , 1.        , 1.41421356, 1.        , 1.41421356,
       2.44948974, 2.23606798, 0.25951556, 1.41421356, 1.41421356,
       1.        , 1.41421356, 1.73205081, 2.        , 1.73205081,
       2.        , 1.73205081, 2.0167668 , 1.03312551, 1.        ,
       1.        , 1.        , 1.41421356, 2.23606798, 0.25951556,
       1.41421356, 1.41421356, 2.23606798, 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.        , 2.        , 1.41421356,
       1.41421356, 1.41421356, 1.41421356, 2.        , 1.73205081,
       1.73205081, 1.        , 1.41421356, 2.23606798])
>>> get_staterror(filter=True, bkg_id=1)
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1.])
```

(note that the background errors haven't changed, it's only the source errors that change). You can see they are smaller than the 4.14 versions. You'll notice that the errors are the same as the before-subtraction errors apart for the three bins which have 1 count in the background and the three bins which have 0's in both source and background

In 4.14 we can look at the ratio of y2 / y1 (after background sbutraction to before subtraction) which is always > 1

```
sherpa-4.14.0> unsubtract()

sherpa-4.14.0> y1 = get_staterror(filter=True).copy()

sherpa-4.14.0> subtract()

sherpa-4.14.0> y2 = get_staterror(filter=True).copy()

sherpa-4.14.0> unsubtract()

sherpa-4.14.0> y2 / y1
array([1.01669768, 1.01669768, 1.0083834 , 1.01669768, 1.0055967 ,
       1.0083834 , 1.01116242, 1.01669768, 1.0083834 , 1.01669768,
       1.0055967 , 1.01669768, 1.01116242, 1.03312551, 1.01669768,
       1.01116242, 1.03312551, 1.01116242, 1.00479908, 1.01669768,
       1.01669768, 1.01669768, 1.01669768, 1.0083834 , 1.0083834 ,
       1.03312551, 1.03312551, 1.01669768, 1.03312551, 1.01669768,
       1.0055967 , 1.00671231, 1.03312551, 1.01669768, 1.01669768,
       1.03312551, 1.01669768, 1.01116242, 1.0083834 , 1.01116242,
       1.0083834 , 1.01116242, 1.0083834 , 1.03312551, 1.03312551,
       1.03312551, 1.03312551, 1.01669768, 1.00671231, 1.03312551,
       1.01669768, 1.01669768, 1.00671231, 1.0083834 , 1.01669768,
       1.01669768, 1.01669768, 1.03312551, 1.0083834 , 1.01669768,
       1.01669768, 1.01669768, 1.01669768, 1.0083834 , 1.01116242,
       1.01116242, 1.03312551, 1.01669768, 1.00671231])
```

We now have the interesting case where the ratio is mostly 1, but there are a few slightly larger than 1 and four actually less than 1

```
>>> unsubtract()
>>> y1 = get_staterror(filter=True).copy()
>>> subtract()
>>> y2 = get_staterror(filter=True).copy()
>>> unsubtract()
>>> y2 / y1
array([1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 0.25951556, 1.        ,
       1.        , 0.25951556, 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 0.25951556, 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.0083834 , 1.03312551, 1.        ,
       1.        , 1.        , 1.        , 1.        , 0.25951556,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        , 1.        ,
       1.        , 1.        , 1.        , 1.        ])
```

anetasie:
Looking at the plots this morning I noticed that in the XSPEC plot you can identify the channels with negative values!

sherpa/stats/__init__.py
anetasie: I just looked at the screen output with `get_stat` and it would be less confusing if you can just remove the last sentence from this description.

anetasie: I was thinking about this sentence:
```Note that this does not
    handle background subtraction, so to match XSPEC extra logic is
    present in sherpa.astro.data.DataPHA.get_staterror to handle that
    case.```

anetasie (approved):

anetasie:
@DougBurke I approved this PR and I suggest to include in the release.

PR #1540, opened 2022-06-27
Tests: fix the random seed for a fake_pha test

DougBurke:
# Summary

Remove randomness from a test.

# Details

This is the last test in test_fake_pha.py that had not fixed ther andom seed. After it failing once-too-many times in a CI run I
decided to fix the seed to make this a repeatable test. The numbers used are a regression test - that is, the counts that get created - but it's hard to come up with a reliable test while allowing the values to vary.

I did add an explicit check that, by adding the background model, we get a vastly-different result (since without it the source model evaluates to 0 for all bins).

### DIFF
--- a/sherpa/astro/tests/test_fake_pha.py
+++ b/sherpa/astro/tests/test_fake_pha.py
@@ -164,9 +164,12 @@ def test_fake_pha_background_pha(reset_seed):
     assert data.counts.sum() < 200
 
 
-def test_fake_pha_bkg_model():
+def test_fake_pha_bkg_model(reset_seed):
     """Test background model
     """
+
+    np.random.seed(5329853)
+
     data = DataPHA("any", channels, counts, exposure=1000.)
 
     bkg = DataPHA("bkg", channels, bcounts,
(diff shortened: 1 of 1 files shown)


sherpa/astro/tests/test_fake_pha.py
```diff
@@ -202,24 +215,19 @@ def test_fake_pha_bkg_model():
...
-    # What we'd like to say is that the predicted counts are
-    # similar, but this is not easy to do. What we can try
-    # is summing the counts (to average over the randomness)
-    # and then a simple check
-    #
-    assert data.counts.sum() > 500
-    assert data.counts.sum() < 1500
-    # This is more likely to fail by chance, but still very unlikely
-    assert data.counts[1] > 1.5 * data.counts[0]
+    assert data.counts == pytest.approx([186, 411, 405])
```
anetasie, line 230: @DougBurke In the comments in line 226 the number of reference counts is different than the one in line 230.
Should these be approximately the same fo the same seed?

DougBurke: No. The predicted signal with no noise is 200, 400, 400. The 186, 411, 405 is the value you get with this particular seed.

anetasie (approved):

PR #1541, opened 2022-06-29
CI bump to Python 3.10 for macOS

DougBurke:
This is to test out https://discuss.python.org/t/pip-install-e-on-macos-suddenly-builds-for-universal2-by-default/16890/18 and is not intended for merging

I didn't update the XSPEC version as I don't know what versions are available on the mac channel.


DougBurke:
Rebased as #1329 has been merged. I expect conda runs to fail because of known issues

DougBurke:
Rebased after merging #1329 just to check

Issue #1542, opened 2022-06-29
C++ warnings from clang from building optmethods


DougBurke:
I was just looking through a mac CI run - https://github.com/sherpa/sherpa/runs/7080787511?check_suite_focus=true - and noted some compilation warnings

```
creating build/temp.macosx-10.9-x86_64-3.8/sherpa/optmethods
creating build/temp.macosx-10.9-x86_64-3.8/sherpa/optmethods/src
INFO: compile options: '-Isherpa/include -Isherpa/utils/src -Isherpa/utils/src/gsl -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/lib/python3.8/site-packages/numpy/core/include -Ibuild/src.macosx-10.9-x86_64-3.8/numpy/distutils/include -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/include/python3.8 -I/Users/runner/work/sherpa/sherpa/miniconda/envs/build/include/python3.8 -c'
extra options: '-msse -msse2 -msse3'
INFO: x86_64-apple-darwin13.4.0-clang++: sherpa/optmethods/src/_saoopt.cc
INFO: x86_64-apple-darwin13.4.0-clang++: sherpa/optmethods/src/Simplex.cc
clang-12: warning: clang-12-Wl,-export_dynamic: 'linker' input unused [-Wunused-command-line-argument]
: warning: -Wl,-export_dynamic: 'linker' input unused [-Wunused-command-line-argument]
In file included from sherpa/optmethods/src/Simplex.cc:25:
In file included from sherpa/optmethods/src/Simplex.hh:23:
In file included from sherpa/include/sherpa/myArray.hh:22:
In file included from /Users/runner/work/sherpa/sherpa/miniconda/envs/build/bin/../include/c++/v1/algorithm:653:
/Users/runner/work/sherpa/sherpa/miniconda/envs/build/bin/../include/c++/v1/memory:891:9: warning: destructor called on non-final 'sherpa::ParVal<double>' that has virtual functions but non-virtual destructor [-Wdelete-non-abstract-non-virtual-dtor]
        __p->~_Tp();
        ^
/Users/runner/work/sherpa/sherpa/miniconda/envs/build/bin/../include/c++/v1/__memory/allocator_traits.h:307:13: note: in instantiation of member function 'std::allocator<sherpa::ParVal<double>>::destroy' requested here
        __a.destroy(__p);
            ^
[...]

PR #1549, opened 2022-07-28
C++ tweaks

DougBurke:
This is to try and "fix" #1542 but as I only dabble in C++ I am not convinced I know what I'm doing here.


DougBurke:
I don't have time/energy/skills to work on this so I'm closing it.

PR #1550, opened 2022-07-29
plot backend - Rebased version of #1382

DougBurke:
This is rebased version of #1382 where most of the commits have been combined (since some no-longer matched their description due to other PRs being merged) and some because it avoids some of the "to be fixed in a later commit" questions.

Some extra work has been done to address issues (but they may be post drop 0 items).


DougBurke:
Argh. This should include both the matplotlib and ds9 "close on exit" fixes, so I don't know why we are getting an unexpected closure. Harumph. At least it's not something I did as @hamogu was seeing this before.

DougBurke:
So, the HACK test looks like it's addressed the "XIO:  fatal IO error 0 (Success) on X server .." error, which suggests that there are some tests which creates plots but are not "protected" by the `all_plot_backends` or `plot_backends` fixtures. Or that somehow these two fixtures are not closing the matplotlib windows. I think the short-term is to take something like the HACK commit 2ea3f4f and we can think about better ways to address this in the future,

There are two other sets of errors

a) a just-started-to-happen error from matplotlib which I don't understand and is not due to this PR as I've seen it in a different PR

b) some tests that fail because the "indep only" test falls over because a plot is trying to access the `barsabove` setting, but this is a matplotlib-only setting

I believe b is likely due to me messing up the rebase in some manner, as I feel like I've discussed this with @hamogu before (but I can't find any evidence of this anymore).

hamogu:
(b) might be real. I know I've looked into it, but I don't remember is that means "looked and fixed" or "looked and working on in commit I have not yet pushed" or "looked and still need to solve". I understand what's happening here. It's to fix by a hack, but a little harder to fix in general, i.e. to allow additional parameter for matplotlib that are not in the general interface, but still pass the tests. Will need to look. Good news is that my vacation and trip is about to end.

DougBurke:
To answer how many tests may not be covered with `plot_backends` or `all_plot_backends` then I get ~340 failing tests [*] if I change the code to be

```
% git diff sherpa/conftest.py
diff --git a/sherpa/conftest.py b/sherpa/conftest.py
index 2e1d56f6..0e02197a 100644
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -610,8 +610,9 @@ def plot_backends(request):
             plt.close(fig="all")
 
 
-@pytest.fixture(autouse=True, scope="session")
-def horrible_hack():
+# @pytest.fixture(autouse=True, scope="session")
+@pytest.fixture(autouse=True, scope="function")
+def horrible_hack(request):
     """Why are some tests failing on GitHub CI with
 
     XIO:  fatal IO error 0 (Success) on X server ":0"
@@ -631,7 +632,12 @@ def horrible_hack():
     except ImportError:
         return
 
+    xs = plt.get_fignums()
     plt.close(fig="all")
+    fixtures = request.fixturenames
+    if len(xs) != 0 and "plot_backends" not in fixtures and "all_plot_backends" not in fixtures:
+        raise NotImplementedError(f"XX: {xs}")
+
 
 
 @pytest.fixture(autouse=True, scope="session")
```

which fails if a test has an open matplotlib window, hopefully filtering out those tests that do have the correct fixture (although I'm not convinced this actually works correctly).

[*] sa there are other test errors going on,and it depends exactly what modules are available and code built, this is an estimate.

hamogu:
There is a reason for many of those not to be covered by the fixtures (or maybe there is a reason for all of them, but I almost certainly was not thorough enough in reviewing exiting tests that passed): When run with those fixtures, the tests get run several times, for example with the BasicBackend, the matplotlib backend and soon also the bokeh backend. Those tests don't take long, but we also don't have unlimited CI resources. So, if a test does not look as if the backend should matter or a very similar functionality is run through all backends by the test just above, then I opted for not decorating every test, just to reduce CI use. Instead, it's just running with whatever the default backend in that installation is. 
Because we run CI with and without matplotlib (and in the future probably with and without bokeh) in different CI runs, we still try out different backends overall, just not in every CI run.

Then, later, the same fixtures also gained the ability to close all matplotlib windows. That was a fix that @DougBurke implemented after I had already written my branch and the got it in by rebasing on main (and some modification). 

So, in principle, those fixtures are not the right tool to deal with matplotlib. In practice, they can be used for that, but it would be cleaner to address that in a different way. In that sense, the "horrible hack" is not all that horrible (except for the name).

How about we make one new fixture with autouse (https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) that closes matplotlib windows if they exist (i.e. add `autouse=True` to `horrible_hack`) and then remove that closing down code from the other two fixtures? It's still a hack, but I think that's OK, it's what teardown functions are for because tests by their very nature are not an exact replication of how a user uses that code - a user would look at a limited, finite number plots and close them when done; while the tests make hundreds of matplotlib plots in very short order.

DougBurke:
Err, doesn't `horrible_hack` already use autouse?

I think the existing fix I had (that I may have messed up in the rebase) already did this - see https://github.com/sherpa/sherpa/blob/2070cd8724f7fabc8fb4726705a79cf70aece490/sherpa/conftest.py#L589-L609 - so are you just suggesting removing the 

```python
        if request.param == 'pylab':
            plt.close(fig="all")
```

bits from the `plot_backends` / `all_plot_backends` fixtures, or maybe leaving them in?

DougBurke:
I've rebased this because #1329 has just been merged (it does not interact with this code, but I wanted to see what happens)

DougBurke:
There's also the RTD failures which are some variant of the following. I assume it is because `docs/conf.py` is not mocking the right thing.

```
Running Sphinx v5.0.2
/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx_astropy/conf/v1.py:226: UserWarning: matplotlib's plot_directive could not be imported. Inline plots will not be included in the output
  "matplotlib's plot_directive could not be imported. " +
WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).
making output directory... done
[autosummary] generating autosummary for: ExamplePlots.ipynb, NotebookSupport.ipynb, SherpaQuickStart.ipynb, bugs.rst, ciao.rst, data/astrodata.rst, data/data.rst, data/index.rst, data/pha.rst, developer/index.rst, ..., plots/pylab_backend_area.rst, plots/utils.rst, quick.rst, statistics/index.rst, statistics/stats.rst, ui/index.rst, ui/sherpa_astro_ui.rst, ui/sherpa_astro_ui_utils.rst, ui/sherpa_ui.rst, ui/sherpa_ui_utils.rst

  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/generate.py", line 394, in generate_autosummary_docs
    name, obj, parent, modname = import_by_name(entry.name)
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/__init__.py", line 654, in import_by_name
    raise ImportExceptionGroup('no module named %s' % ' or '.join(tried), exceptions)
sphinx.ext.autosummary.ImportExceptionGroup: no module named sherpa.plot.pylab_backend.PylabBackend

During handling of the above exception, another exception occurred:

  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/events.py", line 94, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/__init__.py", line 807, in process_generate_options
    encoding=app.config.source_encoding)
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/generate.py", line 399, in generate_autosummary_docs
    name, obj, parent, modname = import_ivar_by_name(entry.name)
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/__init__.py", line 713, in import_ivar_by_name
    real_name, obj, parent, modname = import_by_name(name, prefixes, grouped_exception)
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/ext/autosummary/__init__.py", line 654, in import_by_name
    raise ImportExceptionGroup('no module named %s' % ' or '.join(tried), exceptions)
sphinx.ext.autosummary.ImportExceptionGroup: no module named sherpa.plot.pylab_backend

  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/cmd/build.py", line 275, in build_main
    args.tags, args.verbosity, args.jobs, args.keep_going)
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/application.py", line 256, in __init__
    self._init_builder()
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/application.py", line 314, in _init_builder
    self.events.emit('builder-inited')
  File "/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/1550/lib/python3.7/site-packages/sphinx/events.py", line 103, in emit
    (listener.handler, name), exc, modname=modname) from exc
sphinx.errors.ExtensionError: Handler <function process_generate_options at 0x7fd98c950200> for event 'builder-inited' threw an exception (exception: no module named sherpa.plot.pylab_backend)

Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x7fd98c950200> for event 'builder-inited' threw an exception (exception: no module named sherpa.plot.pylab_backend)
Failed to import sherpa.plot.pylab_backend_area.
Possible hints:
* AttributeError: module 'sherpa.plot' has no attribute 'pylab_backend_area'
* ModuleNotFoundError: No module named 'matplotlib'
Failed to import sherpa.plot.pylab_backend.
Possible hints:
* ModuleNotFoundError: No module named 'matplotlib'
* AttributeError: module 'sherpa.plot' has no attribute 'pylab_backend'
* ```

**Comment by DougBurke** [comment]
So

 - [x] the RTD build now appears to work (at least it doesn't fail, not checked the actual output)
 - [x] the pip tests are passing
 - [ ] the conda tests fail, but this is currently because of some other change (once  that is fixed I expecct the full builds to fail in some places because of my rebasing)

**Comment by DougBurke** [comment]
I'm trying out the fix from #1552 - I expect some conda builds to still fail, but because of potential rebase errors.

**Comment by DougBurke** [comment]
Yay. We're finally back to the expected failures.

**Comment by DougBurke** [comment]
So, I can create a simplified version of the failing test in `sherpa/astro/ui/tests/test_astro_ui_plot.py` which should have the advantage of not needing XSPEC to trigger:

```
@requires_fits
@requires_data
def test_xxxxx(clean_astro_ui, basic_pha1, all_plot_backends):
    """TEST

    """

    ui.plot_model()

```

which fails with

```
sherpa/astro/ui/tests/test_astro_ui_plot.py ..F.                                                                       [100%]

========================================================== FAILURES ==========================================================
________________________________________________ test_xxxxx[IndepOnlyBackend] ________________________________________________

clean_astro_ui = None, basic_pha1 = None, all_plot_backends = None

    @requires_fits
    @requires_data
    def test_xxxxx(clean_astro_ui, basic_pha1, all_plot_backends):
        """TEST
    
        """
    
>       ui.plot_model()

sherpa/astro/ui/tests/test_astro_ui_plot.py:2157: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<string>:1: in plot_model
    ???
sherpa/ui/utils.py:12884: in plot_model
    self._plot(plotobj, overplot=overplot, clearwindow=clearwindow,
sherpa/ui/utils.py:12341: in _plot
    plotobj.plot(**kwargs)
sherpa/plot/__init__.py:632: in plot
    Histogram.plot(self, self.xlo, self.xhi, self.y, title=self.title,
sherpa/plot/__init__.py:536: in plot
    backend.histo(xlo, xhi, y, yerr=yerr, title=title,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (array([0.46720001, 0.48179999, 0.4964    , 0.51099998, 0.52560002,
       0.5402    , 0.55479997, 0.56940001, 0.58399....70671744e-05, 1.67144090e-05, 1.63590216e-05,
       1.59991471e-05, 1.56430344e-05, 1.52859692e-05, 1.49350008e-05]))
kwargs = {'alpha': None, 'barsabove': False, 'capsize': None, 'clearwindow': True, ...}, k = 'barsabove'

    def checked_func(*args, **kwargs):
        for k in kwargs:
            if k not in sig.parameters:
>               raise TypeError(f'{attr.__name__} got keyword argument {k}, ' +
                                'which is not part of the named keyword arguments.')
E               TypeError: histo got keyword argument barsabove, which is not part of the named keyword arguments.

sherpa/plot/backends.py:973: TypeError
```

My guess is that this comes because of how the plot objects are stored in the UI code. This was a problem I think in an original versino of the PR but @hamogu had fixed it. However, with the merging of #1516 it is likely whatever fix was added has been broken. Off to investigate

**Comment by DougBurke** [comment]
I've added a test to show https://github.com/sherpa/sherpa/pull/1550#issuecomment-1205471389 (and now it needs no data/other modules so it should run for all cases). I've also rebased now that #1554 has been merged (so we can remove the change related to #1552).

**Comment by DougBurke** [comment]
Aha: `sherpa/astro/ui/tests/test_astro_ui_plot.py::test_plot_model_all_backends` only fails if matplotlb is installed (which makes sense).

**Comment by DougBurke** [comment]
Aha. I've finally managed to recreate this problem in #1382 (which, although annoying, makes a lot more sense, as I couldn't see anything that would explain the problem only in this PR).

Also note that in this PR I've fixed two typos in some of the commit titles: `blotting` -> `plotting` and `ass` to `pass` (this was a guess as to the correct meaning).

I have tried squashing these commits and separating out the `docs` changes into a separate commit, but it doesn't seem worth it as this stage.

**Comment by hamogu** [comment]
> So, I can create a simplified version of the failing test in sherpa/astro/ui/tests/test_astro_ui_plot.py which should have the advantage of not needing XSPEC to trigger:

This is something I think I knew about and may have worked on. Next week, when I have full day summer camps for Johann again, I'll look at this, evaluate the differences between this branch and #1382, combine them sensibly (which may or may not mean "just take what @DougBurke did here") and try to finish that off.

**Comment by DougBurke** [comment]
@hamogu shall I close this now?

**Comment by hamogu** [comment]
It might be easier for me to actually continue from this PR then from mine, since you've already gone through some of the pain of rebasing. Of course, by now I need to rebase again - and that's all my fault for not pushing this through faster. 
Anyway, I'm waiting for @wmclaugh to tell me that sherpa is frozen for a couple of weeks and then I'll do that, so that I don't have to rebase again because other PRs still get merged. Can we leave this open till I push a new version for #1382?

**Comment by DougBurke** [comment]
No bother - just trying to make sure we have a "sensible" PR list...

**Comment by DougBurke** [comment]
I've rebased this (not because there was a conflict, but just to pick up the latest changes) and have basically squashed all the commits (as a number of them didn't really match the log message thanks to previously-merged changes and the like).

I expect the tests to fail in at least one place.

**Comment by DougBurke** [comment]
So, the expected failure is that one test (I haven't tried an XSPEC build locally, but I would hope we don't get any errors from those build but I have not checked, and the tests may not get to them on the CI run) throws

```
TypeError: histo got keyword argument barsabove, which is not part of the named keyword arguments.
```

I don't understand what is needed to fix this, so I'm leaving it for post-deadline discussion with @hamogu

**Comment by DougBurke** [comment]
So, there is a difference in behavior when the selected backend (in this case pylab, or for the ciao415 case I used dummy in the config file) is not available.

# CIAO/Sherpa 4.15

```
% sherpa
Sherpa 4.15.0

Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:26:04) [GCC 10.4.0]
IPython 8.7.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa

sherpa-4.15.0> load_arrays(1, [1, 2, 3], [1, 2, 3], DataPHA)

sherpa-4.15.0> plot_data()

sherpa-4.15.0>
```

# This PR

```
% python
Python 3.10.8 (main, Nov  4 2022, 13:48:29) [GCC 11.2.0] on linux
>>> ui.load_arrays(1, [1, 2], [1, 2], ui.DataPHA)
>>> ui.plot_data()
WARNING: <class 'sherpa.plot.backends.BasicBackend'> does not implement histogram plotting. No histogram will be produced.
>>> ui.plot_data()
WARNING: <class 'sherpa.plot.backends.BasicBackend'> does not implement histogram plotting. No histogram will be produced.
>>> 
```

# Differences

- the existing code tells me when you import the ui module that the selected backend isn't available, which is helpful (you know early that there's a problem you might want to fix), but can be annoying if you know and don't care
- the existing code doesn't warn when you call an actual "make a plot" call, like `plot_data`
- the new code does not tell you that the backend hasn't been loaded, which makes it hard to know there's a problem
- but whenever you try to make a plot you get a warning

I think we can just add to the load code in `sherpa/plot/__init__.py`

```python
for plottry in plot_opt:
    if plottry in PLOT_BACKENDS:
        backend = PLOT_BACKENDS[plottry]()
        break
else:
    # None of the options in the rc file work, e.g. because it's an old file
    # that does not have dummy listed
    backend = BasicBackend()

    # NEW
    # skip dummy as this is not really an option we need to mention here
    tried_opts = [n for n in plot_opt if n != "dummy"]
    warning(f"Some message to say that unable to find any plot package from {tried_opts}")
```

DougBurke:
To discuss, related to the problems in https://github.com/sherpa/sherpa/pull/1550#issuecomment-1462750971

With this PR (not checked the original) we have

```
% egrep 'histo_prefs ?=' sherpa/plot/__init__.py 
    histo_prefs = basicbackend.get_histo_defaults()
        self.histo_prefs = self.histo_prefs.copy()
                 'histo_prefs = %s') %
    histo_prefs = get_data_hist_prefs()
```

and

```
% egrep 'histo_prefs ?=' sherpa/astro/plot.py 
    histo_prefs = shplot.get_data_hist_prefs()
    histo_prefs = shplot.backend.get_model_histo_defaults()
    histo_prefs = shplot.backend.get_model_histo_defaults()
    histo_prefs = shplot.backend.get_component_histo_defaults()
    histo_prefs = shplot.backend.get_component_histo_defaults()
    histo_prefs = shplot.backend.get_model_histo_defaults()
```

where the `sherpa.plot.get_data_hist_prefs` call is

```python
# I think we want slightly different histogram preferences
# than most (mark by point rather than line).
#
def get_data_hist_prefs():
    """Copy the data preferences to the histogram class"""

    hprefs = basicbackend.get_model_histo_defaults()
    dprefs = basicbackend.get_data_plot_defaults()
    for k, v in dprefs.items():
        if k in hprefs:
            hprefs[k] = v

    return hprefs
```

Now, this leads to the following

```
>>> from sherpa import plot
>>> from sherpa.astro import plot as aplot
>>> 'barsabove' in plot.ModelHistogramPlot().histo_prefs
False
>>> 'barsabove' in aplot.ModelPHAHistogram().histo_prefs
True
>>> plot.backend.name
'pylab'
>>> plot.set_backend("IndepOnlyBackend")
>>> 'barsabove' in plot.ModelHistogramPlot().histo_prefs
False
>>> 'barsabove' in aplot.ModelPHAHistogram().histo_prefs
True
```

The last line is the one that I think causes the problems (since `barsabove` is pylab-specific). 

So, I was going to ask

- should `histo_prefs` ever get a value from `basicbackend` or should it always use `sherpa.backend`?

However, I've realised that `ModelPHAHistogram` has `histo_prefs = shplot.backend.get_model_histo_defaults()` and so it "should" be using the "correct" backend, but of course this is the class definition and so it is not going to be updated when the backend changes.

So, and this is for my understanding as I'm sure @hamogu has already mentioned this in his plans, we could just only ever initialize the `*_prefs` field during `__init__,  but then this is a problem for anyone who wants to change the preferences **before** making a plot (which is a sensible thing to want to do). I then thought about making the prefs access [lazy](https://duckduckgo.com/?q=python+lazy+property&t=canonical&ia=web) but I don't think that would help (ie would probably suffer the same problem).

DougBurke:
So, I think the issue is that

```
% egrep '_prefs ?=' sherpa/astro/plot.py 
    histo_prefs = shplot.get_data_hist_prefs()
    histo_prefs = shplot.backend.get_model_histo_defaults()
    histo_prefs = shplot.backend.get_model_histo_defaults()
    histo_prefs = shplot.backend.get_component_histo_defaults()
    histo_prefs = shplot.backend.get_component_histo_defaults()
    histo_prefs = shplot.backend.get_model_histo_defaults()
```

That is, we get the prefs from the backend and not basicbackend.

The trouble with changing this is that the default prefs no no-longer contain any backend-specific keys, which is not idea for a user who isn't going to change the backend ever and just wants to know what settings can be used.

hamogu:
We're continuing this in #1382, where I took this version, rebased again and will continue work there, thus I'm closing this as suggested by @DougBurke in https://github.com/sherpa/sherpa/pull/1550#issuecomment-1255502346.

PR #1552, opened 2022-08-03
Temporarily Pin Back (py)qt 

Marie-Terrell:
We still need to investigate the real issue is here and what the proper fix is. For now, we can hold this back in our tests to get things going again.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -46,7 +46,7 @@ fi
 echo "dependencies: ${MATPLOTLIB} ${NUMPY} ${FITS} ${XSPEC}"
 
 # Create and activate conda build environment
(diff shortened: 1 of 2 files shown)


Marie-Terrell:
Note: I've tracked down the issue to the qt-main package missing a few dependencies. See the issue created [here](https://github.com/ContinuumIO/anaconda-issues/issues/13007)

I was able to get around the issue by installing the missing dependencies (see [here](https://github.com/sherpa/sherpa/commit/a7647e7a8752a4d424a0eace3e2de7fed74265e3)). However, the workaround here is fine for now until we see what the response is on the issue created above:

- If they fix it we can remove the workaround
- If these are now required, we can switch the fix to the installing the dependencies noted above.

DougBurke (approved):
We could squash this commit into a single commit, or just take the last commit (create a new PR just cherry-picking it), but I leave that decision up to the powers that be.

Marie-Terrell:
Adding these as run requirements is not a desired change for defaults (see the ticket above). So, the long-term fix will need to be to install these deps instead. See #1554 for the fix.

PR #1554, opened 2022-08-04
Install New libxcb Dependencies for Qt

Marie-Terrell:
Based on the feedback: https://github.com/ContinuumIO/anaconda-issues/issues/13007#issuecomment-1205268896

Instead of the workaround in #1552, we will now need to just install these where needed. Which is currently only when we use matplotlib.


DougBurke:
Can this be squashed into a single commit just to make it easier to follow?

wmclaugh:
Sorry @DougBurke, your comment came in while I was merging the PR. Agree it would have been nicer to squash it into one commit.

DougBurke:
No problems.

PR #1555, opened 2022-08-04
Bump the ci-pip-arch workflow to ubuntu-20.04

Marie-Terrell:
It looks like we will begin seeing issues during the brownout period until we update: https://github.com/actions/virtual-environments/issues/6002


DougBurke:
I've given this the "CI: arch" label which triggers the affected action.

DougBurke (approved):
LGTM

Issue #1560, opened 2022-08-15
Improving image filtering


DougBurke:
I'm playing around with #1558 and have the following

```
In [2]: load_image("sherpa-test-data/sherpatest/img.fits")

In [3]: ignore2d("rect(-5, -5, -3, -3)")
1: Field()&!Rectangle(-5,-5,-3,-3)

In [4]: notice2d("circle(10, 10, 10)")
1: Field()&!Rectangle(-5,-5,-3,-3)|Circle(10,10,10)
```

This is technically true, but it's not very helpful. We know the field size, so could do this

- when loading an image, create a filter that matches the box size
  - probably need to also do this whenever a coordinate system change is made, and then we need to think about do we
    always remove any existing filter for this case
- only do it when we add the first filter for `ignore2d` only
- only do it when we add the first filter for both `ignore2d` and `notice2d` (need to check on the logic I cleaned up in #1127 to determine what happens when the first notice filter is outside the data range)

As an example of how this could improve things, the current behavior is basically

```
In [17]: from sherpa.astro.utils._region import Region

In [18]: implicit = Region("field()")

In [19]: implicit
Out[19]: Field()

In [20]: implicit.subtract(Region("rect(-5, -5, -3, -3)"))
Out[20]: Field()&!Rectangle(-5,-5,-3,-3)
```

If we explicitly specify the edge of the image (here not bothering whether we start at 0 or 0.5 or 1...)

```
In [21]: explicit = Region("rect(0,0,100,100)")

In [22]: explicit
Out[22]: Rectangle(0,0,100,100)

In [23]: explicit.subtract(Region("rect(-5, -5, -3, -3)"))
Out[23]: Rectangle(0,0,100,100)
```

Note that here we know that we can drop the rectangle as it's outside the image shape.

This should technically not change any results - it's just a little cleaner for the user. However, there are times a user may want to record the actual regions that have been filtered, even if they are not inside the image, for comparison with other data. I think we just say that it's up to the user to record that information if they want it.

Issue #1558, opened 2022-08-15
Add more information on notice/ignore ranges


DougBurke:
From a comment from the same user who brought us #1557

> One way to mitigate issues like these would be to display the noticed boundaries more often. Running the ignore or notice function could output the values to the screen automatically without having to type show_filter(). It could also be nice when running fit() to output the filter window being used to the terminal screen.

We could add logic to the ui layer for notice/ignore - I don't think I want this to happen with the actual `Data` notice call (for a variety of reasons).

We could add some information to the `FitResults` structure (as this is how information mostly gets displayed from a fit), but I wonder if it could be annoying for multi-data fits, and also should we do it for `ErrorEstResults` too (as it is technically the same thing).

Issue #1430, opened 2022-02-04
filtering inconsistency: what is reported as the filter expression when all data is removed?


DougBurke:
This is with CIAO 4.14 but it's been like this for a while...

The problem is that we have a single routine that handles these two cases differently (actually, there are a couple of ways internally we can handle this but the details aren't important here).

Now, this is not the biggest issue since I believe it's only informational, but this sort of difference in behavior is annoying for user/downstream code. 

# What should be returned for "all-data-is-filtered"

I think that raising an error is excessive here, as it is not unreasonable to end up removing all points. Note that if I were to try and fit this data or calculate the statistic or .. then raising an error may make more sense.

Returning a string like "No noticed bins" is okay for a user, but a bit annoying for any code that may be trying to convert a filter string to something useful (e.g. if the code wants to identify the selected ranges), where the empty string may make more sense.

I do not have an answer here (whether it's worth trying to fix this and, if we do so, what to return). I just wanted to note it down.

# Examples

## PHA data

```
sherpa-4.14.0> load_pha('3c273.pi')
WARNING: systematic errors were not found in file '3c273.pi'
statistical errors were found in file '3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file 3c273.arf
read RMF file 3c273.rmf
WARNING: systematic errors were not found in file '3c273_bg.pi'
statistical errors were found in file '3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file 3c273_bg.pi

sherpa-4.14.0> get_filter()
'0.001460000058:14.950400352478'

sherpa-4.14.0> ignore(None, 5)

sherpa-4.14.0> get_filter()
'5.022399902344:14.950400352478'

sherpa-4.14.0> ignore(lo=4)
```

If we now ask what the filter is we get the empty string:

```
sherpa-4.14.0> get_filter()
''
```

However, if we were to create a filter that ignored all the points we get a different message:

```
sherpa-4.14.0> ignore()

sherpa-4.14.0> get_filter()
'No noticed bins'
```

## Data1D

If we have a non-PHA but still 1D dataset the behavior is slightly different:

```
sherpa-4.14.0> load_arrays(10, [1, 2,3], [1, 2, 3])

sherpa-4.14.0> get_filter(10)
'1.0000:3.0000'

sherpa-4.14.0> ignore_id(10)
```

If we filter out all the data in one go then we get an error raised (in this case `DataErr`):

```
[...]

PR #1562, opened 2022-08-15
report the change in the filter because of calls to notice and ignore

DougBurke:
# Summary

When a `notice` or `ignore` call - including the variants like `notice_id` and `ignore2d` - then the change in the filter is reported to the screen (for users of the ui layer). This is to help users understand what a filter has actually done, as there are times when it may be surprising (in particular with grouped PHA data). The output can be hidden by using the `sherpa.utils.logging.SherpaVerbosity` context manager:

    >>> with SherpaVerbosity("WARN"):
    ...     notice(0.5, 8)

# Details

This is the initial response to #1558 as it is the simplest thing to do. There may be more changes we would like to make - such as enhancing the `FitResults` structure - but that can be done at a later time, as this is hopefully going to provide significant benefit.

Two issues to be aware of:

1. for the `DataIMG` case, the empty string means either "all data", and we handle this by replacing it by `Field()`, or "no data", and we leave this as "" as this is the convention used for the 1D case. See also issue #1560 
2. the filer for background PHA datasets is only shown when explicitly selected - i.e. the `bkg_id` parameter is set. Technically we could show the range when the background is not being subtracted (i.e. even when the user is filtering the "source"), but let's leave that for a potential update, once we understand how well this works.

The implementation is reasonably simple, but there is some annoying repetition in the astro 2d case as we do not pass through an ignore keyword. We should look at improving this in a later PR but I didn't want to change the API of the `notice2d` family of calls in this PR (see #1586). One place I have tried to be careful with is that the logic for `notice`/`ignore` is now to do something like

    ostr = data.get_filter(...)
    data.do-the-filter()
    nstr = data.get_filter(...)
    output_the_old_and_new_filter(ostr, nstr)

but if the `get_filter` call fails then we may not end up doing the filter. Technically the only way to get a failure is if something very wrong has happened, but we have at least two cases where it can happen and (knowing the filter code, there could be other odd cases out there):

1. using `ignore_bad` can leave us in a strange state where some things still work and I don't know what is going on, but let's not break this
2. if a user happens to have used Data2D or Data2DInt then we suffer from #1587 and strange things could happen (in this case it might be helpful to the user to error out as they are likely to get wring results, but again let's not break things needlessly)

In looking at the code I noticed (again) that we do not reliably handle the "add-data-is-filtered" case - e.g. see #1430 :

1. for 1D data we use the empty string, but for `DataIMG` filters the empty string means **both** "all data is noticed" and "all data is ignored"
3. some times we get a `"No noticed bins"` string instead of the empty string (for `DataPHA`), but I have worked around this by checking for `data.mask is False`
4. I don't know if it's still the case, but there have been cases where a filter that removes all data results in an error (there are times this makes sense, but probablly not when just reporting things, but I can't remember the details just now).

We also have some unresolved differences between `get_filter` and `get_filter_expr`, and in whether this output can be handled by the low-level `parse_expr` code (which assumes we use `a:b` to indicate a range, in part because `-` is complicated to read when you have negative coordinates)
 
Edge case: we allow a string to be sent to `notice` and family, which then gets interpreted as a set of filters. This means that we get a line of output after each call, as shown below. Handling this to only create one line would be significantly-more work than I have to put into this PR:

```
>>> ignore()
dataset 1: 15:30 x -> no data
>>> notice("8:12, 18:22")
dataset 1: no data -> 10 x
dataset 1: 10 -> 10,20 x
```

ie it would be nice to just have a single output saying `dataset 1: no data -> 10,20 x`, but that's complicated. Similarly we could try and combine the output from multiple datasets when they are the same, but I don't think it's worth it.

# Examples

## PHA

```

In [2]: for i in range(1, 5):
   ...:     load_pha(i, f"sherpa-test-data/sherpatest/obs{i}.pi")
   ...: 
read ARF file sherpa-test-data/sherpatest/obs1.arf
read RMF file sherpa-test-data/sherpatest/obs1.rmf
read background file sherpa-test-data/sherpatest/obs1_bg.pi
read ARF file sherpa-test-data/sherpatest/obs2.arf
read RMF file sherpa-test-data/sherpatest/obs2.rmf
read background file sherpa-test-data/sherpatest/obs2_bg.pi
read ARF file sherpa-test-data/sherpatest/obs3.arf
read RMF file sherpa-test-data/sherpatest/obs3.rmf
read background file sherpa-test-data/sherpatest/obs3_bg.pi
read ARF file sherpa-test-data/sherpatest/obs4.arf
read RMF file sherpa-test-data/sherpatest/obs4.rmf
read background file sherpa-test-data/sherpatest/obs4_bg.pi

In [3]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.4964:7.008 Energy (keV)
dataset 2: 0.00146:14.9504 -> 0.4964:7.008 Energy (keV)
dataset 3: 0.00146:14.9504 -> 0.4964:7.008 Energy (keV)
dataset 4: 0.00146:14.9504 -> 0.4964:7.008 Energy (keV)

In [4]: notice()
dataset 1: 0.4964:7.008 -> 0.00146:14.9504 Energy (keV)
dataset 2: 0.4964:7.008 -> 0.00146:14.9504 Energy (keV)
dataset 3: 0.4964:7.008 -> 0.00146:14.9504 Energy (keV)
dataset 4: 0.4964:7.008 -> 0.00146:14.9504 Energy (keV)

In [5]: for i in range(1, 5):
   ...:     group_counts(i, 5)
   ...: 

In [6]: notice()
dataset 1: 0.00146:14.9504 Energy (keV) (unchanged)
dataset 2: 0.00146:14.9504 Energy (keV) (unchanged)
dataset 3: 0.00146:14.9504 Energy (keV) (unchanged)
dataset 4: 0.00146:14.9504 Energy (keV) (unchanged)

In [7]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.00146:9.0374 Energy (keV)
dataset 2: 0.00146:14.9504 Energy (keV) (unchanged)
dataset 3: 0.00146:14.9504 -> 0.4818:13.2422 Energy (keV)
dataset 4: 0.00146:14.9504 Energy (keV) (unchanged)

In [8]: ignore_id(2, hi=0.8)
dataset 2: 0.00146:14.9504 -> 0.8906:14.9504 Energy (keV)
```

## PHA (background)

I am just being explicit (checking against the underlying `DataPHA` objects) to check it all makes sense.

```

In [2]: load_pha("sherpa-test-data/sherpatest/3c273.pi")

In [3]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.4672:9.8696 Energy (keV)

In [4]: get_data().get_filter()
Out[4]: '0.467200011015:9.869600296021'

In [5]: get_bkg().get_filter()
Out[5]: '0.467200011015:9.869600296021'

In [6]: notice(bkg_id=1)  # note that this filter is applied to the existing filter (which is why we are clearing it)
dataset 1: background 1: 0.4672:9.8696 -> 0.00146:14.9504 Energy (keV)

In [7]: get_data().get_filter()
Out[7]: '0.467200011015:9.869600296021'

In [8]: get_bkg().get_filter()
Out[8]: '0.001460000058:14.950400352478'

In [9]: notice(0.3, 8, bkg_id=1)
dataset 1: background 1: 0.00146:14.9504 -> 0.2482:9.8696 Energy (keV)

In [10]: get_data().get_filter()
Out[10]: '0.467200011015:9.869600296021'

In [11]: get_bkg().get_filter()
Out[11]: '0.248199999332:9.869600296021'

In [12]: group_counts(1, 3, bkg_id=1)

In [13]: get_data().get_filter()
Out[13]: '0.467200011015:9.869600296021'

In [14]: get_bkg().get_filter()
Out[14]: '0.001460000058:9.942600250244'

In [15]: notice(bkg_id=1)
dataset 1: background 1: 0.00146:9.9426 -> 0.00146:14.9504 Energy (keV)

In [16]: notice(0.3, 8, bkg_id=1)
dataset 1: background 1: 0.00146:14.9504 -> 0.00146:8.176 Energy (keV)

In [17]: notice(bkg_id=1)
dataset 1: background 1: 0.00146:8.176 -> 0.00146:14.9504 Energy (keV)

In [18]: notice(0.5, 7, bkg_id=1)
dataset 1: background 1: 0.00146:14.9504 -> 0.3066:7.2416 Energy (keV)

In [19]: get_data().get_filter()
Out[19]: '0.467200011015:9.869600296021'

In [20]: get_bkg().get_filter()
Out[20]: '0.306600004435:7.241600036621'
```

## image

The output of "Field()" in line 3 is a hack, since get_filter() returns "" which is taken to mean everything, rather than nothing as it does for the 1D case (although note that it also uses "" to indicate all data is ignored):

```

In [2]: load_image("bob", "sherpa-test-data/sherpatest/acisf08478_000N001_r0043_regevt3_srcimg.fits")

In [3]: notice2d("circle(100, 100, 10)")
dataset bob: Field() -> Circle(100,100,10) 

In [4]: notice2d("rect(0, 0, 200, 20)")
dataset bob: Circle(100,100,10) -> Circle(100,100,10)|Rectangle(0,0,200,20) 

In [5]: notice2d("rect(0, 0, 200, 200)")
dataset bob: Circle(100,100,10)|Rectangle(0,0,200,20) -> Circle(100,100,10)|Rectangle(0,0,200,20)|Rectangle(0,0,200,200) 
```

## Image with all or no data

```
In [6]: notice2d()
dataset 1: Circle(4324.5,3827.5,430) -> Field()

In [7]: notice2d()
dataset 1: Field() (unchanged)

In [8]: ignore2d()
dataset 1: Field() -> no data

In [9]: ignore2d()
dataset 1: no data (unchanged)

```

## PHA filtering out all data

What happens when all the data is filtered out - part 1:

```
In [2]: load_pha("sherpa-test-data/sherpatest/3c273.pi")

In [3]: get_filter()
Out[3]: '0.001460000058:14.950400352478'

In [4]: notice(0.5, 7)
dataset 1: 0.00146:14.9504 -> 0.4672:9.8696 Energy (keV)

In [5]: ignore(lo=2)
dataset 1: 0.4672:9.8696 -> 0.467:-1.9418 Energy (keV)

In [6]: ignore(hi=3)
dataset 1: 0.4672:1.9418 Energy (keV) -> no data

In [7]: get_filter()
Out[7]: ''

In [8]: notice(2, 3)
dataset 1: no data  -> 1.9418:3.0806 Energy (keV)

In [9]: notice()
dataset 1: 1.9418:3.0806 -> 0.00146:14.9504 Energy (keV)
```

Part 2: however, because of oddities (eg #1430) we can get different output which I have now fixed (it used to say "No noticed bins Energy (keV)" rather than "no data"):

```
In [10]: ignore()
dataset 1: 0.00146:14.9504 Energy (keV) -> no data

In [11]: ignore()
dataset 1: no data (unchanged)

In [12]: notice(2,3)
dataset 1: no data -> 1.9418:3.0806 Energy (keV)
```

## Data1D basic checks

```

In [2]: load_arrays(1, [1, 2, 3, 4, 10, 20, 25, 30], [1, 2, 3, 4, 5, 6, 7, 8])

In [3]: notice(lo=5)
dataset 1: 1:30 -> 10:30 x

In [4]: notice(lo=2, hi=24)
dataset 1: 10:30 -> 2:30 x

In [5]: ignore(3, 5)
dataset 1: 2:30 -> 2,10:30 x
```

## Data1D filtering out all data

```

In [2]: load_arrays(1, [1, 2, 3, 4, 10, 20, 25, 30], [1, 2, 3, 4, 5, 6, 7, 8])

In [3]: notice()
dataset 1: 1:30 x (unchanged)

In [4]: ignore()
dataset 1: 1:30 x -> no data

In [5]: ignore()
dataset 1: no data (unchanged)

In [6]: notice(lo=5)
dataset 1: no data -> 10:30 x
```

## Data1DInt filtering out all data

```

In [2]: load_arrays(1, [1, 2, 5], [2, 4, 6], [1, 2, 10], Data1DInt)

In [3]: notice()
dataset 1: 1:6 x (unchanged)

In [4]: ignore()
dataset 1: 1:6 x -> no data
```

## DataPHA with ignore_bad

Calling `ignore_bad` does not always change the `get_filter` output (at least not at the moment), as as we use this to determine what to display it may seem like `ignore_bad` is always going to look ike

```
>>> ignore_bad()
dataset 1: 0.00146:14.9504 Energy (keV) (unchanged)
```

which might make it a bit confusing. However, as `ignore_bad` resets the filter it turns out to be useful (along with the existing `WARING` message which has not been changed in this PR):

```
>>> notice(0.5, 8)
dataset 1: 0.00146:14.9504 -> 0.4672:9.8696 Energy (keV)
>>> ignore_bad()
WARNING: filtering grouped data with quality flags, previous filters deleted
dataset 1: 0.4672:9.8696 -> 0.00146:14.9504 Energy (keV)
```

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session_image.py
+++ b/sherpa/astro/ui/tests/test_astro_session_image.py
@@ -1,0 +1,196 @@
+#
+#  Copyright (C) 2021
+#  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Astro-specific tests of the image functionality for Session
+
+Notes
+-----
+A number of tests send data to an external DS9 process (which will be
+started if needed) and then check the values displayed by the process.
+These tests are highly-likely to fail if the tests are run in
+parallel, such as with
+
+    pytest -n auto
+
+as there is no support in our DS9 interface to either create a new DS9
+instance per worker, or to "lock" a DS9 so that the tests can be run
+in serial.
+
+"""
+
+import numpy as np
+
+import pytest
+
+from sherpa.astro.ui.utils import Session as AstroSession
+
+from sherpa.astro.data import DataIMG
+
+from sherpa.utils.err import ArgumentTypeErr
+from sherpa.utils.testing import requires_ds9
+
+
+def example_data():
+    """Create an example data set.
+
+    It's not entirely clear whether this is a valid object, as we do
+    not clearly describe what the units of the intependent axis are
+    meant, or allowed, to be: e.g.  must they by logical units with
+    any coord system added via WCS, or can they be anything, as used
+    here? This means that some of the results (e.g. the filters used)
+    may need to be updated if we ever resolve this.
+
+    """
+
+    x1, x0 = np.mgrid[-4:5, 6:15]
+
+    shp = x1.shape
+
+    x0 = x0.flatten()
+    x1 = x1.flatten()
+    y = 100 / (1 + np.sqrt((x0 - 11)**2 + (x1 + 2)**2))
+
+    return DataIMG('example', x0, x1, y, shape=shp)
+
+
+@requires_ds9
+@pytest.mark.parametrize("idval", [None, "foo"])
+@pytest.mark.parametrize("funcname", ["notice", "ignore"])
+def test_image_xxx2d_not_sent_an_image(funcname, idval):
+    """Check we error out
+
+    This test takes a surprising amount of time as we don't
+    error out before setting up DS9, so we try to reduce the
+    number of parameters.
+
+    If we address #1583 the funcname parameter can be removed, as we
+    would expect the image_data call to fail.
+
+    """
+
+    idarg = 1 if idval is None else idval
+    s = AstroSession()
+    s.load_arrays(idarg, [1, 2, 3], [1, 2, 3])
+
+    # Should this fail? see #1583
+    if idval is None:
+        s.image_data()
+    else:
+        s.image_data(idval)
+
+    func = getattr(s, f"{funcname}2d_image")
+    with pytest.raises(ArgumentTypeErr,
+                       match="^'img' must be a image data set$"):
+        if idval is None:
+            func()
+        else:
+            func(idval)
+
+
+@requires_ds9
+@pytest.mark.parametrize("funcname", ["notice", "ignore"])
+def test_xxx2d_image_no_region(funcname):
+    """What happens when no region has been supplied?"""
+
+    s = AstroSession()
+
+    # This is a bit awkward as we have to send a region to
+    # DS9 with XPA so we can check we can read it back
+    # with notice2d_image.
+    #
+    d = example_data()
+    s.set_data(d)
+    s.image_data()
+
+    func = getattr(s, f"{funcname}2d_image")
+    with pytest.raises(ValueError,
+                       match="^unable to parse region string: ''$"):
+        func()
+
+
+def send_filters_to_ds9():
+    """Filters used by test_xxx2d_image"""
+
+    from sherpa.image import backend
+    backend.xpaset("regions", data="circle(10,0,3)")
+    backend.xpaset("regions", data="box(8,4,8,6,30)")
+
+
+FILTER_NTOT = 81
+FILTER_NSET = 47
+FILTER_A = "Circle(10,0,3)"
+FILTER_B = "RotBox(8,4,8,6,30)"
+
+
+@requires_ds9
+def test_notice2d_image():
+    """A valid region"""
+
+    s = AstroSession()
+
+    # This is a bit awkward as we have to send a region to
+    # DS9 with XPA so we can check we can read it back
+    # with notice2d_image.
+    #
+    s.set_data(example_data())
+    s.image_data()
+
+    d = s.get_data()
+    assert s.get_filter() == ""
+    assert d.mask is True
+
+    send_filters_to_ds9()
+
+    s.notice2d_image()
+    assert s.get_filter() == f"{FILTER_A}|{FILTER_B}"
+
+    assert len(d.mask) == FILTER_NTOT
+    assert d.mask.sum() == FILTER_NSET
+
+
+@requires_ds9
+def test_ignore2d_image():
+    """A valid region
+
+    Unlike test_notice2d_image this uses an explicit
+    dataset identifier.
+    """
+
+    s = AstroSession()
+
+    # This is a bit awkward as we have to send a region to
+    # DS9 with XPA so we can check we can read it back
+    # with notice2d_image.
+    #
+    s.set_data("ex", example_data())
+    s.image_data("ex")
+
+    d = s.get_data("ex")
+    assert s.get_filter("ex") == ""
+    assert d.mask is True
+
+    send_filters_to_ds9()
+
+    s.ignore2d_image("ex")
+    assert s.get_filter("ex") == f"Field()&!{FILTER_A}&!{FILTER_B}"
+
+    # 47 is from test_notice2d_image
+    assert len(d.mask) == FILTER_NTOT
+    assert d.mask.sum() == (FILTER_NTOT - FILTER_NSET)
--- a/sherpa/astro/ui/tests/test_astro_session_image.py
+++ b/sherpa/astro/ui/tests/test_astro_session_image.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2021
+#  Copyright (C) 2022
 #  Smithsonian Astrophysical Observatory
 #
 #
--- a/sherpa/astro/ui/tests/test_filtering.py
+++ b/sherpa/astro/ui/tests/test_filtering.py
@@ -26,7 +26,7 @@ import pytest
 
 import numpy as np
 
-from sherpa.astro.data import DataPHA
+from sherpa.astro.data import DataIMG, DataPHA
 from sherpa.astro import ui
 from sherpa.astro.ui.utils import Session as AstroSession
 from sherpa.data import Data1DInt
@@ -111,13 +111,23 @@ def test_filter_notice_bad_361(make_data_path, caplog):
     with caplog.at_level(logging.INFO, logger='sherpa'):
         ui.ignore_bad()
 
-    assert len(caplog.records) == 6
+    assert len(caplog.records) == 8
 
     lname, lvl, msg = caplog.record_tuples[5]
+    assert lname == 'sherpa.ui.utils'
+    assert lvl == logging.INFO
+    assert msg == '1: 0.0073:14.9504 -> 0.4964:8.0592 Energy (keV)'
+
+    lname, lvl, msg = caplog.record_tuples[6]
     assert lname == 'sherpa.astro.data'
     assert lvl == logging.WARNING
     assert msg == 'filtering grouped data with quality ' + \
         'flags, previous filters deleted'
+
+    lname, lvl, msg = caplog.record_tuples[7]
+    assert lname == 'sherpa.ui.utils'
+    assert lvl == logging.INFO
+    assert msg == '1: 0.4964:8.0592 -> 0.0073:14.9504 Energy (keV)'
 
     s1 = ui.calc_stat()
     assert s1 == pytest.approx(stats['bad'])
@@ -128,15 +138,28 @@ def test_filter_notice_bad_361(make_data_path, caplog):
 def test_filter_bad_notice_361(make_data_path):
     """Test out issue 361: ignore bad then notice.
 
-    This is expected to fail with NumPy version 1.13 and should cause
-    a DeprecationWarning with earlier versions (I do not know when
-    the warning was added).
+    There have been changes in NumPy that this originally was written
+    to catch (version 1.13). This has been left in, but it is
+    currently a regression test as the handling of bad values means
+    that the notice call fails when trying to write out the filter,
+    hence the try/except handling there.
+
     """
 
     stats = setup_model(make_data_path)
 
     ui.ignore_bad()
-    ui.notice(0.5, 8.0)
+
+    # With support for #1558 - writing out the filter as we apply them
+    # - we now get this call failing, so catch this failure to allow
+    # the test to continue. This should be fixed, but it's a
+    # long-standing problem.
+    #
+    try:
+        ui.notice(0.5, 8.0)
+    except IndexError:
+        pass
+
     s1 = ui.calc_stat()
     assert s1 == pytest.approx(stats['0.5-8.0'])
 
@@ -221,7 +244,16 @@ def test_filter_bad_grouped(make_data_path, clean_astro_ui):
     # What happens when we filter the data? Unlike #1169
     # we do change the noticed range.
     #
-    ui.notice(0.5, 7)
+    # With support for #1558 - writing out the filter as we apply them
+    # - we now get this call failing, so catch this failure to allow
+    # the test to continue. This should be fixed, but it's a
+    # long-standing problem.
+    #
+    try:
+        ui.notice(0.5, 7)
+    except IndexError:
+        pass
+
     assert pha.quality_filter == pytest.approx(expected)
 
     # The mask has been filtered to remove the bad channels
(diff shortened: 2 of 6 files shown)


hamogu:
I think this is the right way to go. It frees us somewhat from laying down 100% what the rules are (which of course we do in code, but in the past we found that hard to do in prose that's short enough for a user to read it - who reads the documentation anyway?), because you always see *what* happens, even if the reason *why* that happens in that particular case might still be confusing. 

As a regular Sherpa user, I have the feeling "there should be a configuration option to turn this off so I don't have to see it every time", but as a developer I know that the best way is to educate users on how to use the logging system we already have (`SherpaVerbosity` et al.), because that's cleaner, simpler, less bug-prone, and can be used to switch on an off all sorts of output, not just this one. So, I fully support output using the normal logging procedures for output with no extra on/off switch.

DougBurke:
We can also add a "switch" later if we need.

DougBurke:
@anetasie - you might like to try this out. There's some issues that need working on, but I think it's likely useful, and input earlier rather than later is useful.

anetasie:
The response showing the applied filter is good. I also like the option to have a turn off switch, but this could be added later.

anetasie (approved):

anetasie:
@DougBurke This is still marked as Draft. I approved this PR and suggest to include in the release.

DougBurke:
@anetasie  There are still some questions I described in the commit message above that need discussing / thinking about. 

The immediate ones for this PR are

1. if you filter a background dataset it's a very-subtle indicator (in this case changing "1:" to "1:1:") - should we be more explicit

```
In [18]: notice(0.5, 7, bkg_id=1)
1:1: 0.00146:14.9504 -> 0.3066:7.2416 Energy (keV)
```

2. we have horrible confusion about what an empty filter means: for 1D/PHA data it means everything has been **excluded** whereas for 2D data we have it mean all data is **included**. For this PR I have a "hack" for the 2D case where it replaces an empty string by "Field()" - it is just "informative" but is it sensible (should we do something different for the 1D case too?)

```
In [2]: load_image("bob", "sherpa-test-data/sherpatest/acisf08478_000N001_r0043_regevt3_srcimg.fits")

In [3]: notice2d("circle(100, 100, 10)")
bob: Field() -> Circle(100,100,10) 
```

There are more "general" issues we need to think about to make sure we keep up on our technical debt (and that generally expand upon issue 2).

hamogu:
point 1: Yes, we should be more explicit. How about `dataset 1:` instead of `1:` and `dataset 1: background 1:` instead of `1:1:`?

sherpa/astro/ui/utils.py
```diff
@@ -50,6 +50,28 @@
 __all__ = ('Session',)
 
 
+def _get_image_filter(data):
+    """When reporting filters, we need to handle images separately.
+
+    There is a disconnect between 1D and 2D filters as an empty string
+    means no data has been seected for the former, but all data is
```
wmclaugh, line 57:
```suggestion
    means no data has been selected for the former, but all data is
```

sherpa/astro/ui/utils.py
```diff
@@ -50,6 +50,28 @@
...
+    selected in the latter. For the logging of the filters this makes
+    things awkward, so we over-ride the image case and replace an mpty
```
wmclaugh, line 59:
```suggestion
    things awkward, so we over-ride the image case and replace an empty
```

sherpa/ui/utils.py
```diff
@@ -110,6 +108,94 @@ def get_plot_prefs(plotobj):
...
+       The background identifier (PHA data only).
+
+    Notes
+    -----
+
+    We technically don't require the delim=':', format='%g' arguments
+    to the get_filter call, but I list them so we have consistent
+    code. Ideally this would be encapsulated in the Data class, so we
+    let the object define the best arguments to use, but it's not
+    guaranteed to work well so we are trynig this explicit approach.
```
wmclaugh, line 154:
```suggestion
    guaranteed to work well so we are trying this explicit approach.
```

DougBurke:
As a note, I'd want #1477 to land before merging this one, as they are going to conflict in some way and this is a much-smaller PR to rebase.

DougBurke:
@hamogu and @wmclaugh  - I think I've finished this now. The code has been slightly-reworked from earlier (I basically just merged most of the individual commits) and then added new commits to address some issues we've been discussing, and because in writing the final commit message I got around to covering some of the corner cases I'd found during development. One change has been to try and make sure a `notice` call (or any of the others) won't fail because `get_filter` fails for some reason. I have created a bunch of issues to track potential future work.

anetasie:
@DougBurke it looks good. I tried a few options. 
Thanks for the update.

sherpa/astro/ui/tests/test_astro_session_image.py
```diff
@@ -0,0 +1,196 @@
+#
+#  Copyright (C) 2021
```
hamogu, line 2:
```suggestion
#  Copyright (C) 2022
```
at least I think that's the current year. Or is it still March 2020, about day 968?

sherpa/astro/ui/tests/test_filtering.py
```diff
@@ -171,20 +221,29 @@ def test_filter_bad_ungrouped(make_data_path, clean_astro_ui):
...
 
     # We need to ignore to change the mask.
     #
     ui.ignore(None, 0.5)
+    assert len(caplog.records) == 3
+    clc_filter(caplog, "dataset 1: 0.0073:14.5416 -> 0.511:14.5416 Energy (keV)")
+
     ui.ignore(7, None)
+    assert len(caplog.records) == 4
+    clc_filter(caplog, "dataset 1: 0.511:14.5416 -> 0.511:6.9934 Energy (keV)")
```
hamogu, line 237: Should at least one of those tests use wavelength instead of energy?

sherpa/ui/utils.py
```diff
@@ -110,6 +108,142 @@ def get_plot_prefs(plotobj):
...
+        return None
+
+    except IndexError:
+        # This is a known failure case with ignore_bad handling of
+        # DataPHA.
+        #
+        return None
+
+
+def report_filter_change(idstr, ofilter, nfilter, xlabel=None):
```
hamogu, line 175:
```suggestion
def report_filter_change(idstr, ofilter, nfilter, xlabel=""):
```
Why make a special case with `None` here? Would it not be easier to simply default to an empty string, then there is no need ot document the None case, and no need to deel with in in line 216-220. (Simply use `"ostr += f"no data -> {nfilter} {label}"` later and if there is no label there is an extra " " at the end. Would that be a problem?

Either way, I'm writing here longer than it takes to do either of those, so feel free to ignore.

sherpa/ui/utils.py
```diff
@@ -4974,37 +5115,84 @@ def notice(self, lo=None, hi=None, **kwargs):
...
+        #
+        # The bkg_id argument is astro-specific. The need
+        # to use it to access the correct data object is
+        # problematic.
+        #
+        try:
+            bkg_id = kwargs["bkg_id"]
+            del kwargs["bkg_id"]  # important to remove
+        except KeyError:
+            bkg_id = None
```
hamogu, line 5175: There is a method for that in dicts:
```suggestion
        bkg_id = kwargs.pop("bkg_id", None)
```

sherpa/ui/utils.py
```diff
@@ -5186,8 +5409,38 @@ def notice_id(self, ids, lo=None, hi=None, **kwargs):
...
```
hamogu, line 5423:
```suggestion
        bkg_id = kwargs.pop('bkd_id', None)
```

sherpa/ui/utils.py
```diff
@@ -5186,8 +5409,38 @@ def notice_id(self, ids, lo=None, hi=None, **kwargs):
...
+            data.notice(lo, hi, **kwargs)
+            nfilter = _get_filter(data)
+
+            try:
+                xlabel = data.get_xlabel()
+            except AttributeError:
+                # Data2D case
+                xlabel = None
+
+            report_filter_change(idstr, ofilter, nfilter, xlabel)
```
hamogu, line 5443: I'm a little worried that this is almost (or entirely?) identical to the implementation of `notice`. Is notice not simply a call do `notice_ids` with all datasets selected? If so, why do we copy the implementation? The more lines of code we need, to more it seems to me that we should try to unify. If it's only a few lines, OK, but this PR considerably lengthens the implementation as well as the docstring. 
Something like:
```python
def notice(self, lo=None, hi=None, **kwargs):
    self.notice_ids(self.list_Data_ids(), lo, hi, **kwargs)
```


For the docstring, can't we just have the full examples and description for one of them and say "See docs to notice_id for examples"?
If I were to implement that again, I would make just one function with `ids` as an optional parameter, if given it selects the dataset, if not, you get all datasets. That's too late now, but at least we can try to cut down an duplicated implementation.

hamogu (changes requested):

sherpa/ui/utils.py
DougBurke: In #1586 I've noted that the 2D code needs re-working to remove duplication. We can extend that to here too, but I don't have time/energy/desire to make significant changes (even if they should be logically obviously correct, I've learnt not to trust this belief too much...) at this time.

sherpa/ui/utils.py
DougBurke: So, "foo {xlabel} bar" would lead to two spaces if `xlabel` were the empty string. Technically we can handle this with checking for the empty string, but I like the Optional[string] behavior here, as it is not overloading the meaning of what an empty string is. It does rely on the implicit contract that the `Data` `xlabel` value is always set (which it is with our existing data hierarchy).

sherpa/ui/utils.py
DougBurke: However, saying that, I can remove *some* repetition of code here.

sherpa/astro/ui/tests/test_filtering.py
DougBurke: At some level I'd hope that the existing notice/filter/... code should check this,as all we're really doing is repackaging existing code, but you are right that we should add a test. So I've done so.

sherpa/ui/utils.py
DougBurke: I've moved this code to the top-level in `notice_data_range`. We can look at more-extensive rework for the next release.

sherpa/ui/utils.py
DougBurke: I think this is a stylistic choice so I've not changed it. We can re-visit if we decide to extend/enhance/change the code in the future.

hamogu (approved):

PR #1563, opened 2022-08-19
Add Python 3.10 Conda Builds, Drop Python 3.7 and Numpy <1.20, Switch to Pip for Conda Builds

Marie-Terrell:
Continuing with [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), we can drop Python 3.7 and Numpy <1.20 support for this next release. 

This PR bumps the Conda package builds to 3.8-3.10, switches to using pip instead of setup.py for those builds, and bumps various references to these versions. 

### Notes
As noted during the Python 3.9 addition (#1192), we will need to use Numpy 1.21 for Python 3.10 (Conda does not have a compatible Numpy 1.20.* package for Python 3.10). So, we have two main options:

1. Stay with the current SHERPA_PYTHON_VERSION method and bump the numpy version for all Conda package builds to numpy 1.21 (as done in #1203).
2. Or, switch to using "--python" with a selector for determining the Numpy version (just when using the Conda recipe to build). 

Option 2 is the method that I have currently used in this PR (as it is the same we use in CIAO's Sherpa currently). This change would mean that the Conda recipe will not use the SHERPA_PYTHON_VERSION environment variable to determine the Python version to build for (and will not have the "default" Python version of our minimum supported version). If someone uses the recipe, they will just need to add `--python 3.8` to their call to `conda build ...`. The only people impacted by this change would be those who use the recipes/conda directory to build. 

Option 1 and 2 are both valid. If there is a preference either direction, I can update this accordingly.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -39,14 +39,14 @@ jobs:
           - name: Linux Minimum Setup (Python 3.8)
             os: ubuntu-latest
             python-version: "3.8"
-            numpy-version: 1.18
+            numpy-version: "1.18"
             install-type: develop
             test-data: none
 
           - name: Linux Minimum Setup (Python 3.10)
             os: ubuntu-latest
             python-version: "3.10"
-            numpy-version: 1.21
+            numpy-version: "1.21"
             install-type: develop
             test-data: none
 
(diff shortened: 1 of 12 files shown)


Marie-Terrell:
I just bumped the macOS test to Python 3.10. I'm fine with the update especially since this might catch any issues with the newer updates.

DougBurke (approved):
I am fine with either option but I believe the consensus was option 2, which is what this PR supports so I'm approving it,

PR #1578, opened 2022-09-12
Gitlab: Reduce artifact retention period

harlantc:
GitLab will be reducing the storage amount for private repositories soon, and our sherpa mirror on GitLab has ~5gb in generated artifacts. This PR simply reduces the time from 2 weeks to 3 days, which should be satisfactory as we already do our own backup of artifacts and haven't needed that far back of retention in the first place. This is tracked under CM-559.

This PR is purely GitLab-CI focused and has no impact on the GitHub-CI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,7 +28,7 @@ define-version:
   artifacts:
     reports:
       dotenv: version.env
-    expire_in: "2 weeks"
+    expire_in: "3 days"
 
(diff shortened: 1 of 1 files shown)


Marie-Terrell (approved):
Looks good to me!

Marie-Terrell:
Actually, added one minor change to also reduce the time to save the version.env for to 3 days as well. This is a very small file, so this won't change much anyway

PR #1585, opened 2022-09-18
Tests: work around matplotlib 3.6.0 warning

DougBurke:
# Summary

Allow the tests to pass when run with matplotlib version 3.6.0.

# Details

We have a long history of seeing "strange" errors about setting a logged axis to a negative value in our tests. I haven't the energy to address this (as it does appear to be in mpl), so continue to allow new warning messages, such as this one from matplotlib version 3.6.0. It turns out this is https://github.com/matplotlib/matplotlib/issues/9970 so ignoring the message is the best we can do.


anetasie (approved):
@DougBurke thanks for tracking this.

PR #1589, opened 2022-09-20
Set the language when building the documentation

DougBurke:
# Summary

Ensure the language setting is defined for recent versions of Sphinx

# Details

I thought I had a recent PR that made this change, but I can't find it. My guess is that this change happened in version 5.0 of Sphinx but I'm not spendnig a lot of time investigating it.


hamogu (approved):
This feature was changed in sphinx 5.0, but this PR effectively does not change anything, because `None` selects the sphinx default, and that default is `"en"` (https://www.sphinx-doc.org/en/master/usage/configuration.html). However, no harm in being explicit.

PR #1600, opened 2022-09-29
Change the histogram1d and 2d routines so that they do not sort the input arguments 

### LINKED ISSUE #1601: histogram1d/2d issue
With #1477 now landed, the following code - which is based on the `sherpa.utils.histogram1d` examples - now fails:

```
>>> from sherpa.ui import dataspace1d, get_indep
>>> dataspace1d(1, 10, 1)
>>> lo, hi = get_indep()
>>> print(lo)
[1. 2. 3. 4. 5. 6. 7. 8. 9.]
>>> print(hi)
[ 2.  3.  4.  5.  6.  7.  8.  9. 10.]
>>> from sherpa.utils import histogram1d
>>> n = histogram1d([2, 1, 0, 3, 2, 7, 5, 4, 9, 12], lo, hi)
  File "/home/dburke/miniconda3/envs/CIAOX/lib/python3.8/site-packages/sherpa/utils/__init__.py", line 2105, in histogram1d
    x_lo.sort()
ValueError: sort array is read-only
```

The problem is that the independent axis (which is internally stored away within `DataSpace1D`/`EvaluationSpace1D` as ` `PointAxis` or `InegratedAxis`) is now marked as "read only", and - **for some reason** - the `histogram1d` routine calls `sort` on the input `lo` and `hi` arrays.

There's several possible fixes

1. the user can just copy the arrays - e.g. `histogram1d(vals, lo.copy(), hi.copy())`
2. the `histogram1d` routine can internally call `copy`
3. we do away with sorting the input arguments in `histogram1d` and just make it a pre-condition on calling the routine (e.g. that the data is sorted)
4. we remove the read-only change added in #1477

Option 1 is short-term.
Option 2 is easy.
Option 3 is what I think I'd prefer.
Option 4 is, to me, a non-starter (as it was added to ensure things like this don't happen!).

DougBurke:
# Summary

The histogram1d and histogram2d routines can now fail in certain circumstances (such as when given the independent axes of a dataset, as suggested in one of the examples for histogram1d). Address this by making sure they copy the input arguments before sorting them. Fix #1601 

# Details

With #1477 the output of get_dep is marked read-only, which means that the second example for histogram1d now fails. Add tests for both 1d and 2d histogram variants that check this condition, and tweak the histogram1d example to be a bit easier to follow (and, in future, test).

As noted in #1601 this is option 2. I'd prefer option 3 - that is, remove the sort calls and note that the routines are expected to be called with sorted arguments - but that requires some research.

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2117,7 +2117,7 @@ def histogram2d(x, y, x_grid, y_grid):
     See the `numpy.histogram2d` routine for a version with more options.
 
     .. versionchanged:: 4.15.0
-       The x_grid and x_grid arguments are no-longer changed (sorted)
+       The x_grid and y_grid arguments are no-longer changed (sorted)
        by this routine.
 
     Parameters
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2088,13 +2088,12 @@ def histogram1d(x, x_lo, x_hi):
     >>> xhi = edges[1:]
     >>> y = histogram1d(x, xlo, xhi)
 
-    Given a list of samples (``vals``), bin them up so that
-    they can be used as the dependent axis (the value to
-    be fitted) in a Sherpa data set:
+    Given a list of samples, bin them up so that they can be used as
+    the dependent axis (the value to be fitted) in a Sherpa data set:
 
-    >>> dataspace1d(0.1, 10, 0.1)
-    >>> (lo, hi) = get_indep()
-    >>> n = histogram1d(vals, lo, hi)
+    >>> dataspace1d(1, 10, 1)
+    >>> lo, hi = get_indep()
+    >>> n = histogram1d([2, 3, 2, 8, 5, 2], lo, hi)
     >>> set_dep(n)
 
     """
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2060,6 +2060,10 @@ def histogram1d(x, x_lo, x_hi):
 
     See the `numpy.histogram` routine for a version with more options.
 
+    .. versionchanged:: 4.15.0
+       The x_lo and x_hi arguments are no-longer changed (sorted) by
+       this routine.
+
     Parameters
     ----------
     x : sequence of numbers
@@ -2098,8 +2102,8 @@ def histogram1d(x, x_lo, x_hi):
 
     """
 
-    x_lo = numpy.asarray(x_lo)
-    x_hi = numpy.asarray(x_hi)
+    x_lo = numpy.asarray(x_lo).copy()
+    x_hi = numpy.asarray(x_hi).copy()
 
     x_lo.sort()
     x_hi.sort()
@@ -2111,6 +2115,10 @@ def histogram2d(x, y, x_grid, y_grid):
     """Create 2D histogram from a sequence of samples.
 
     See the `numpy.histogram2d` routine for a version with more options.
+
+    .. versionchanged:: 4.15.0
+       The x_grid and x_grid arguments are no-longer changed (sorted)
+       by this routine.
 
     Parameters
     ----------
@@ -2144,8 +2152,8 @@ def histogram2d(x, y, x_grid, y_grid):
     >>> set_dep(n)
 
     """
-    x_grid = numpy.asarray(x_grid)
-    y_grid = numpy.asarray(y_grid)
+    x_grid = numpy.asarray(x_grid).copy()
+    y_grid = numpy.asarray(y_grid).copy()
 
     x_grid.sort()
     y_grid.sort()
--- a/sherpa/utils/tests/test_utils_unit.py
+++ b/sherpa/utils/tests/test_utils_unit.py
@@ -395,12 +395,11 @@ def test_histogram1d_sort_axis():
     gb = ghi.copy()
     n = histogram1d([2, 3, 4, 5, 6, -1, 7, 20, 8, 9], glo, ghi)
 
-    assert glo == pytest.approx(sorted(ga))
-    assert ghi == pytest.approx(sorted(ghi))
+    assert glo == pytest.approx(ga)
+    assert ghi == pytest.approx(ghi)
     assert n == pytest.approx([1, 3, 1, 2, 1])
 
 
-@pytest.mark.xfail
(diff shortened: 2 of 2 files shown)


hamogu (approved):

sherpa/utils/__init__.py
```diff
@@ -2113,6 +2116,10 @@ def histogram2d(x, y, x_grid, y_grid):
 
     See the `numpy.histogram2d` routine for a version with more options.
 
+    .. versionchanged:: 4.15.0
+       The x_grid and x_grid arguments are no-longer changed (sorted)
```
wmclaugh, line 2120:
```suggestion
       The x_grid and y_grid arguments are no-longer changed (sorted)
```

DougBurke: Ta. One day I will create a PR without a typo or thinko. One day. But not today.

DougBurke:
@anetasie - that error is coming from the C++ histogram code. I believe this is noting to do with this PR - i.e. if you were able to regenerate the same data that you had here for the `sortpar` array (e.g. if you have the session still available we could save the arrays to disk to try in CIAO 4.14) then you should (I believe) see the same error.

I think the only way to get this particular error is when `k >= n` but I haven't looked to see if there's a logical error somewhere (e.g. in `find_bin` or this code). I wonder if it should treat `k >= n` the same way it treats `k < 0`

```
  template <typename ConstArrayType, typename FloatType,
	    typename IntArrayType, typename IntType>
  inline int histogram1d (const ConstArrayType& bv,
			  const ConstArrayType& x_lo,
			  const ConstArrayType& x_hi,
			  IntArrayType& res) {
    IntType n, nbins;
    
    n = bv.get_size();
    nbins = x_lo.get_size();

    for (IntType i = 0; i < n; i++)
      {
        FloatType t = bv[i];
        IntType k = find_bin (t, x_lo, x_hi, nbins);
	if (k >= n)
	  return EXIT_FAILURE;
	if (k >= 0)
	  res[k] += 1;
      }
    
    return EXIT_SUCCESS;
  }
```

As mentioned (or hinted at) in #1602 we should review all this to see whether we can just use numpy for binning data.

Here's a case where the `histogram1d` code fails and **it does not make sense to me**:

```
>>> histogram1d([0, 1.9], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 1, 0, 0], dtype=int32)
>>> histogram1d([0, 1.9, 2], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 1, 1, 0], dtype=int32)
>>> histogram1d([0, 1.9, 2, 1], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 2, 1, 0], dtype=int32)
>>> histogram1d([0, 1.9, 2, 1, 2], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 2, 2, 0], dtype=int32)
>>> histogram1d([0, 2], [0, 1, 2, 3], [1, 2, 3, 4])
  File "/home/dburke/sherpa/sherpa-work/sherpa/utils/__init__.py", line 2108, in histogram1d
    return hist1d(numpy.asarray(x), x_lo, x_hi)
ValueError: histogram1d failed
>>> histogram1d([2], [0, 1, 2, 3], [1, 2, 3, 4])
  File "/home/dburke/sherpa/sherpa-work/sherpa/utils/__init__.py", line 2108, in histogram1d
    return hist1d(numpy.asarray(x), x_lo, x_hi)
ValueError: histogram1d failed
>>> histogram1d([0.1, 2], [0, 1, 2, 3], [1, 2, 3, 4])
  File "/home/dburke/sherpa/sherpa-work/sherpa/utils/__init__.py", line 2108, in histogram1d
    return hist1d(numpy.asarray(x), x_lo, x_hi)
ValueError: histogram1d failed
>>> histogram1d([2, 0.1], [0, 1, 2, 3], [1, 2, 3, 4])
  File "/home/dburke/sherpa/sherpa-work/sherpa/utils/__init__.py", line 2108, in histogram1d
    return hist1d(numpy.asarray(x), x_lo, x_hi)
ValueError: histogram1d failed
>>> histogram1d([0.1], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 0, 0, 0], dtype=int32)
```

anetasie:
@DougBurke I checked in CIAO 4.14 and yes, this also failed in that case, so the failure is not due to this PR or any recent changes in the code.

anetasie:
@DougBurke the second example in the `histogram1d` documentation shows the failures. Now, I'm confused if this PR was going to fix that failure or not.

https://sherpa--1600.org.readthedocs.build/en/1600/overview/api/sherpa.utils.histogram1d.html?highlight=histogram1d

anetasie:
I'm not sure I'm building the right version...

```
In [10]: sherpa.__version__
Out[10]: '4.14.1+249.g032054f8'
```
```
In [2]: import numpy as np

In [3]: dataspace1d(1,10,1)

In [4]: lo, hi = get_indep()

In [5]: print(lo)
[1. 2. 3. 4. 5. 6. 7. 8. 9.]

In [6]: print(hi)
[ 2.  3.  4.  5.  6.  7.  8.  9. 10.]

In [7]: n = histogram1d([2,3,2,8,5,2], lo, hi)
Input In [7], in <cell line: 1>()
----> 1 n = histogram1d([2,3,2,8,5,2], lo, hi)

File ~/git/sherpa/sherpa/utils/__init__.py:2111, in histogram1d(x, x_lo, x_hi)
   2108 x_lo.sort()
   2109 x_hi.sort()
-> 2111 return hist1d(numpy.asarray(x), x_lo, x_hi)

ValueError: histogram1d failed
```

DougBurke:
Well, this looks like it should be the right version. Looking at https://github.com/sherpa/sherpa/commits/main the last version at this moment is labelled as `032054f` which matches your `version` string.

DougBurke:
@anetasie - you are talking about bug #1605 which is fixed in #1607, not in #1600.

This PR fixed the case that `histogram1d` would have failed with the message

```
ValueError: sort array is read-only
```

which is issue #1601

anetasie:
ok. Thanks for clarification. I will include the link to this example in the caveat for the release.

DougBurke:
Note that the `ValueError: histogram1d failed` error is a **very old** bug so I don't think that needs to be explained when describing the fix in this PR ( #1600 )

PR #1597, opened 2022-09-27
Revert "Doctestplus"

wmclaugh:
Reverts sherpa/sherpa#1521

Pull this PR out of the 4.15.0 release cut. The updates will be merged back after the release to provide the functionality to developers while and allow time for additional testing

PR #1604, opened 2022-10-02
Replace utils.histogram from Sherpa

### LINKED ISSUE #1601: histogram1d/2d issue

### LINKED ISSUE #1602: histogram1d issues - does not respect upper bin boundaries
## Problem
I'm testing if `sherpa.utils.histrogram1d` deal with non-consequtive or overlapping bins:

```python
>>> from sherpa.utils import histogram1d
>>> histogram1d(np.arange(0, 3, .1), [0, 1, 2, 2.5], x_hi = [.5, 2.2, 3., 4.])
array([10, 10, 10,  0], dtype=int32)
```
So, it does not raise an error, but the result is wrong. In fact, setting x_hi does not seem any effect - I experimented with different numbers there.

No error or warning  is raised and the result is wrong.

Note 1: I did not test `sherpa.utils.histrogram2d` but I suspect that the same happens there since some of the code is similar.

Note 2: This is a separate issue from #1601, since one is caused in the Python layer and the other in C and it's possible to fix them independently. However, what I suggest (below) would take care of both issues at the same time.

## Suggestion

Post-4.15 histogram and histogram2d should be replaced by:
```python
def histrogram(*args, **kwargs):

    raise NotImplementedError('This function could return wrong results in some cases and has been removed. Use numpy.histogram instead.')
```
Their usage in Sherpa will be replaced by better tested numpy equivalents. I agree with @DougBurke  that it's unlikely that anyone uses those functions directly, so we can afford to remove them, in particularly since they sometimes generated wrong results. There is no other call (except for a test) to the C-functions `hist1d` and `hist2d` so those can be removed also.

hamogu:
This replaces the 1d version of histogram in Sherpa with direct use of te equivalent nuumpy function.

fixes #1602
fixes #1601

Note: One reason I'm pushing now is that I see odd behavior locally, that looks as
if it might be related to #1597 and I want to compare output on CI with
my local output to check what's going on on my
local installation.


hamogu:
Both the numpy and the Sherpa implementations of the histogram have some behavior that looks odd at first in terms of where events are binned that fall below the first or above the last bin. That's not to say that what they implement is wrong, only that it surprised me. Either way, neither one is obviously better than the other. So, we *could* remove some Sherpa code by converting to use numpy (as this PR suggests) but unless we have any maintenance to  do on the sherpa code anyway, that's not the best investment of our time. Given that this PR now has conflicts and does not address an obvious need, I'll close it.

Issue #1605, opened 2022-10-03
bug in histogram code


DougBurke:
Following up from https://github.com/sherpa/sherpa/pull/1600#issuecomment-1265760768

The underlying code uses the following:

https://github.com/sherpa/sherpa/blob/8fee46a003b78ca449dc4a75089e5172b5ba6a54/sherpa/include/sherpa/utils.hh#L616-L638

The problem is that the line

	if (k >= n)
	  return EXIT_FAILURE;

should be checking against `nbins` - the length of the `x_lo`/`x_hi` arrays, and **not** `n`, which is the length of the values array (that is, the values we are binning up).

You can check this with 

```
>>> from sherpa.utils._utils import hist1d
>>> hist1d([0], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 0, 0, 0], dtype=int32)
>>> hist1d([0, 1], [0, 1, 2, 3], [1, 2, 3, 4])
array([1, 1, 0, 0], dtype=int32)
```

but if we change the first array to just `[1]` then it fails:

```
>>> hist1d([1], [0, 1, 2, 3], [1, 2, 3, 4])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: histogram1d failed
```

This is because here `n` is 1, and `find_bins` has returned 1 (correctly), which triggers the error.

# Workaround

You can work around this by extending the initial array (the values to bin) so that they are at least as long as the xlo/xhi arrays. Below I show two variants:

1. picking a value smaller than xlo
2. NaN

I use a list, but you can also do this to a numpy array, it's just a bit-more awkward.

```
sherpa-4.14.0> from sherpa.utils import histogram1d

sherpa-4.14.0> histogram1d([2], [0, 1, 2], [1, 2, 3])
ValueError: histogram1d failed

sherpa-4.14.0> histogram1d([2] + [-1] * 2, [0, 1, 2], [1, 2, 3])
array([0, 0, 1], dtype=int32)

sherpa-4.14.0> histogram1d([2] + [np.nan] * 2, [0, 1, 2], [1, 2, 3])
array([0, 0, 1], dtype=int32)

```

Issue #1606, opened 2022-10-03
Unexpected edge behavior in histogram1d


DougBurke:
Taking care to work around #1605, we have

```
>>> from sherpa.utils import histogram1d
>>> histogram1d([0, 1, 2], [0, 1, 2], [1, 2, 3])
array([1, 1, 1], dtype=int32)
>>> histogram1d([0, 1, 2, 3], [0, 1, 2], [1, 2, 3])
array([1, 1, 2], dtype=int32)
>>> histogram1d([0, 1, 2, 3.01], [0, 1, 2], [1, 2, 3])
array([1, 1, 1], dtype=int32)
>>> histogram1d([-0.01, 0.01], [0, 1, 2], [1, 2, 3])
array([1, 0, 0], dtype=int32)
```

I object to the `[1, 1, 2]` output, which was created by binning up [0, 1, 2, 3] for the bins

    0-1
    1-2
    2-3

I believe it should have returned `[1, 1, 1]` as the ranges are 'lo <= x < hi` (or, at least, that's what makes sense to me).

hamogu:
I think that latter is another manifestation of #1602 
```Python
for i in range(5):
    print(i, histogram1d([0, 1, 2, 3], [0, 1, 2], [1, 2, i]))

0 [1 1 1]
1 [1 1 1]
2 [1 1 1]
3 [1 1 2]
4 [1 1 2]
```

Note that in #1602 I claimed that it would not respect the upper bin boundary at all, but that's wrong. If the upper bin boundary is < 3, the results is [1,1,1], otherwise it's [1,1,2]. However, the result is [1,1,1] even if the upper bin boundary is lower than the lower bin boundary.

Numpy does not have a separate xho and xhi, but just a single array of bin boundaries that must be monotonically sorted, thus avoiding my silly examples like `histogram1d([0, 1, 2, 3], [0, 1, 2], [1, 2, 1])`. However, it also makes the last bin inclusive of the end, but not the other ones:
```Python
np.histogram([0,1,2,3], bins=[0,1,2,3])[0]

array([1, 1, 2])
```

DougBurke:
I do think - and the history of th enumpy histogram code is instructive - that it's actually harder than it originally seems to properly formulate histogram binning...

DougBurke:
From [`numpy.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html):

    Notes
    -----
[...]

PR #1607, opened 2022-10-03
Fix an issue when binning values into a 1D histogram

DougBurke:
# Summary

Fix a problem when binning data into a histogram (issue #1605).

# Details

I'm not sure of the best way to describe the problem - it's something like when the number of values to bin up is smaller than the number of bins, but that's not quite right. So I've left the description rather vague. We haven't had many people complaining about this bug, so I assume it's rare.

### DIFF
--- a/sherpa/include/sherpa/utils.hh
+++ b/sherpa/include/sherpa/utils.hh
@@ -1,4 +1,4 @@
-// 
+//
 //  Copyright (C) 2008, 2021  Smithsonian Astrophysical Observatory
 //
 //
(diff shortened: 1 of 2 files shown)


hamogu:
Do you want to push this into 4.15 or it is for later? The change is easy enough, but I'm not that familiar with the C code. The examples look good enough, but this is a case that we should build and try out, I think.

hamogu (approved):
I *think* I understand what the C++ code does now and the test cases make sense to me. I also recognize that what is called `n` in `find_bins` is called `n_bins` in `histrogram1d` - an easy mistake to make.
Based on my going through this by hand for a few cases, I don't think the EXIT_FAILURE case can ever happen, but that might have to be revisited in the context of #1606.

DougBurke:
@hamogu - I guess it could fail if a value in `x_lo` or `x_hi` were `NaN`? I don't think it's a high-priority issue, either way, and something we can think about down the line :-)

wmclaugh (approved):

PR #1608, opened 2022-10-04
Remove the test setup.py option

DougBurke:
# Summary

Add the test configuration so that we can ensure the necessary testing packages can be installed when the code is built. Support for running the tests with 'python setup.py test' has been removed to match changes in the Python packaging ecosystem.

# Details

As `python setup.py test` has long been deprecated, let's remove it. This was originally part of #1329 and then #1352 but I've pulled it out as a separate entity as it can be done without tying us to using `tox`, which is what #1352 does.

The changes are relatively minimal as the documentation and CI runs have already been converted to use `pytest` (done in #1329). The remaining changes are 

1. a `MANIFEST.in` change thay may or may not be needed (I've seen enough different behavior depending on exactly what is being done and what versions are in use that I am erring on "let's be explicit").
2. add support for a `test` label so that `pip install -e .[test]` will ensure that pytest and pytest-xvfb are installed; this in turn saves a whole line in one of the CI runs
3. document updates
4. remove the `test` option from `setup.py` and provide the user with information on what they should be doing. 
5. the "how to build the documentation" section has moved to using `make html` rather than the `build_sphinx` option as that is also going away (this time in Sphinx 7, see #1714)

PR #1949 build on this to remove the use of `distutils`/`numpy,distutils` needed to build with Python 3.12.


DougBurke:
Rebased since #1669 was merged (it didn't conflict, I just wanted to make sure things still work)

DougBurke:
Rebased just to keep t up to date.

DougBurke:
The conflict comes from #1751 which may go away soon (it's just about whether we install `attrs` in one of the CI actions).

DougBurke:
Rebased to address conflict in setup.py and to get the python 3.11 support added.

DougBurke:
Rebased to pick up NumPy 1.25 support and because of a recent minor conflict from the recent doc updates (in the documentation, which we now don't need because of the recent doc updates...).

DougBurke:
Rebased to address minor documentation conflict and as part of checking out recent CI build issues.

DougBurke:
This was rebased as it was used in #1949

DougBurke:
Rebased to pick up #1938 and hopefully let the macOS tests pass

DougBurke:
Rebased to pick up recent build changes.

DougBurke:
Rebased to pick up recent build changes which lead to a conflict or two.

Marie-Terrell:
Two notes:

- pytest-xvfb: I am a bit concerned about the requirement being added here for testing. That would then mean that this is then always required for CIAO's sherpa testing as well?
- setup.py getting added to manifest: I thought the idea was to move away from setup.py. Is this required as of this PR? How would we be able to remove this in the future (or would this still exist even in the pip world)?

DougBurke:
1. `pytest-xvfb` could be removed (it may only hit me, as work has ds9 installed so it ends up running the tests and that is very annoynig). However, I don't see it as a terrible requirement (in that it's a small module and it does noting if you don't have Xvfb installed), but given recent publicity about supply-side attacks (or even mainter ones) maybe we want to take care.
2. So, `setup.py` is not necessarily going away, just that calling it directly is going away. Whether it needs to be added to the manifest file is a question I **really don't care about**. I am just trying to update our code to be somewhat closer to the recent python packaging standards and when I first did this work multiple years ago there was some notes about files not being included. What the current situation is is currently unclear to me, but it is getting very frustrating keeping this patch set maintained and up to date.

DougBurke:
So, I've hopefully removed the `pytest-xvfb` requirement. I haven't reverted the `oldest-supported-numpy` change because I thought that it was still needed, and looking at some of the build CI reports I see warnings from pip about inconsistent installations (as astropy 6 needs numpy >= 1.22 but the oldest version ends up picking numpy 1.21). I'm not sure if this is a build-time vs install-time issue so let's go one change at a time here.

DougBurke:
@Marie-Terrell - for `oldest-supported-numpy` we have

 1. https://github.com/scipy/oldest-supported-numpy explicitly says that this is deprecated
 2. https://pypi.org/project/oldest-supported-numpy/ doesn't say anything
 3. https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy is more concerned about whether we should add a 'numpy<2' requirement in, as we currently don't test aganist numpy 2 (and, although #2025 should have "fixed" our last numpy issue, we know the tests would fail as discussed in #1864)

I will note that the latest build instructions do somewhat assume we have an updated build (e.g. that `python -m build` will work) which really needs the follow-on PRs installed. They also assume we are following the NEP - err I mean SPEC - https://scientific-python.org/specs/spec-0000/ which moves much faster than we can/do.

So I honestly don't know whether 

- using `numpy`
- using `oldest-supported-numpy`
- using `numpy>=2.0.0rc1`
- using `numpy<2`
- __shrug__

is the best thing to do

DougBurke:
@Marie-Terrell - here is a recent build showing the warning messages I see (when using `oldest-supported-numpy`). This is for the `Linux Full Build (Python 3.9)` conda build - https://github.com/sherpa/sherpa/actions/runs/8991672492/job/24699781592

```
Using pip 24.0 from /usr/share/miniconda/envs/build/lib/python3.9/site-packages/pip (python 3.9)
Obtaining file:///home/runner/work/sherpa/sherpa
  Installing build dependencies: started
  Running command pip subprocess to install build dependencies
  Collecting setuptools<60,>=51.0.0
    Downloading setuptools-59.8.0-py3-none-any.whl.metadata (5.0 kB)
  Collecting wheel
    Downloading wheel-0.43.0-py3-none-any.whl.metadata (2.2 kB)
  Collecting oldest-supported-numpy
    Downloading oldest_supported_numpy-2023.12.21-py3-none-any.whl.metadata (9.8 kB)
  Collecting numpy==1.19.3 (from oldest-supported-numpy)
    Downloading numpy-1.19.3-cp39-cp39-manylinux2010_x86_64.whl.metadata (2.0 kB)
  Downloading setuptools-59.8.0-py3-none-any.whl (952 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 952.8/952.8 kB 11.8 MB/s eta 0:00:00
  Downloading wheel-0.43.0-py3-none-any.whl (65 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 65.8/65.8 kB 21.0 MB/s eta 0:00:00
  Downloading oldest_supported_numpy-2023.12.21-py3-none-any.whl (4.9 kB)
  Downloading numpy-1.19.3-cp39-cp39-manylinux2010_x86_64.whl (14.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.9/14.9 MB 95.0 MB/s eta 0:00:00
  Installing collected packages: wheel, setuptools, numpy, oldest-supported-numpy
  ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
  astropy 6.0.1 requires numpy<2,>=1.22, but you have numpy 1.19.3 which is incompatible.
  contourpy 1.2.1 requires numpy>=1.20, but you have numpy 1.19.3 which is incompatible.
  matplotlib 3.8.4 requires numpy>=1.21, but you have numpy 1.19.3 which is incompatible.
  Successfully installed numpy-1.19.3 oldest-supported-numpy-2023.12.21 setuptools-59.8.0 wheel-0.43.0
  Installing build dependencies: finished with status 'done'
  Checking if build backend supports build_editable: started
...
```

Marie-Terrell:
Thank you for all of the above info. I didn't realize it was being deprecated. I'm good with oldest-supported-numpy -> numpy with all of the above. We can then discuss any other ranges on the 4.17.0 scale.
Would you mind if that change was pulled into a separate PR though (to make it a more explicit change)? I can put in a quick PR to pull this in

Marie-Terrell (approved):

DougBurke:
So, this re-force should have corrected the removal of `pytest-xvfb` but also still get `pytest` installed if you say `pip install -e .[test]`, as well as making sure `pytest-xvfb` is installed in our conda config (in a more-obvious location now - i.e. where we install DS9, which is the only thing that makes us need it).

I've also reverted the `numpy` change.

PR #1609, opened 2022-10-07
XSPEC minimum version is now 12.12.0

DougBurke:
# Summary

Bump the minimum supported XSPEC version to version 12.12.0. 

# Details

Bump the XSPEC minimum to 12.12.0. This is based on #1396 which bumps the minimum version to 12.10.1. This main advantages of moving to XSPEC 12.12.0 are done in a follow-up PR: #1615 There are two small changes here

- we no-longer need to declare the model typedefs (from funcType.h) model names (from functionMap.h) (with earlier releases of XSPEC these were not importable for various reasons)
- we can switch a model to use the model.dat-defined function (it should have been versioned but for whatever reason had not been)

The main aims of this PR are

1. simplify the code (e.g. removing conditional compilation blocks, in particular the table-interpolation code)
2. simplify the code from having the XSPEC headers present
3. allow for future work (#1615)

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -296,7 +296,7 @@ Update the XSPEC bindings?
 --------------------------
 
 The :py:mod:`sherpa.astro.xspec` module currently supports
-:term:`XSPEC` versions 12.12.1 down to 12.10.1. It may build against
+:term:`XSPEC` versions 12.12.1 down to 12.12.0. It may build against
 newer versions, but if it does it will not provide access to any new
 models in the release. The following sections of the `XSPEC manual
 <https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XspecManual.html>`__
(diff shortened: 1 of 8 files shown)


Marie-Terrell:
I hit enter too early on my review. I'm fine with the above, except the github action update (script and workflow). #1687 can land before this one and then the changes to the actions there should cover the bump in the actions.

DougBurke:
Rebased as #1676 was merged which caused minor issues with #1396 causing that to be rebased.

DougBurke:
Rebased as #1396 and #1687 have been merged. Will send the first two commits and then the third commit, to check it all behaves sensibly.

DougBurke:
For @Marie-Terrell - this should be ready for a once over. It shouldn't have changed much but rebasing and integration of your work in #1687 needs a quick check to make sure I haven't done anything stupid.

Marie-Terrell:
No worries! I already checked over it again. Agreed. This is ready to go in

PR #1610, opened 2022-10-11
Allow the XSPEC installation to use the CIAO_HEADAS environment variable

DougBurke:
# Summary

Allow the XSPEC environment variable used by the Sherpa XSPEC interface to be changed from HEADAS to another value via the xspec_env_name configuration variable.

# Details

This is built on #1396 - only the last commit is relevant to this PR.

We have issues in CIAO when trying to use a HEASOFT installation along with CIAO. One possibility is for CIAO/Sherpa to access data via a different environment variable - e.g. CIAO_HEADAS rather than HEADAS - which this patch enables. It has not been tested out.

If the `xspec_env_name` setting in the `xspec_config` section of `setup.cfg` is un-commented / set then it's vaule - such as `CIAO_HEADAS` - will be used to look for the XSPEC environment rather than `HEADAS` when initializing the XSPEC user model library.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -296,7 +296,7 @@ Update the XSPEC bindings?
 --------------------------
 
 The :py:mod:`sherpa.astro.xspec` module currently supports
-:term:`XSPEC` versions 12.12.1 down to 12.9.0. It may build against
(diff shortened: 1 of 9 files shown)


DougBurke:
I don't think we are moving forward with this path, so I'm going to mothball it for now.

PR #1613, opened 2022-10-14
Allow testing of the documentation

hamogu:
# Summary

Allow doctests to run in Sherpa

# Details

This is a redo of #1521, which we pulled before a release. It adds the ability to run doctests in py and rst files. Those tests will run automatically, if doctestplus is installed (not on CI in the current setup) and the file is not explicitly balcklisted for doctestplus. This PR fixes up a few files as examples. Common problems that prevent doctestplus to run a file are 
(from the common, trivial on the top, to the rare, but worthy of fixing at the bottom):

    formatting of code
    numpy not imported
    slightly different formatting of of the output
    example contains typos
    example uses variable that are not defined
    example uses functionality that's buggy or outdated.

The far goal is to run doctests for all rst files and all embedded examples of the docs (https://github.com/sherpa/sherpa/issues/816). This would ensure that the examples we give execute at all, and, where we have expected output to compare to, also act as an additional tests. We have about 1000 doctests embedded in our rst docs, so running all those automatically is a significant enhancement in itself.

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -1,0 +1,52 @@
+#
+#  Copyright (C) 2022
+#  MIT
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+'''Pytest fixtures for running doctests
+
+Pytest looks for conftest.py files that define fixtures only in the
+directory tree that is being tested. When we run doctests in the
+documentation only, it will therefore not find that fixtures that are
+defined in sherpa/conftest.py, thus this file copies over the fixtures we
+need.
+'''
+import pytest
+from sherpa.utils.testing import get_datadir
+
+@pytest.fixture(autouse=True)
+def add_sherpa_test_data_dir(doctest_namespace):
+    '''Define `data_dir` for doctests
+
+    We make this an autouse=True fixture, because that means that
+    the variable `data_dir` is available in every file that we may
+    want to test without any special markup to that file. There is a
+    risk that this is too magical and could confuse developers
+    in the future, but it seems more important that this keeps any
+    extra markup out of those files and keep them clean.
+    '''
+    # We could error out here, but we only want the tests that
+    # use it to fail.
+    #
+    path = get_datadir()
+    if path is None:
+        pytest.skip("sherpa-test-data not found")
+
+    if not path.endswith('/'):
+        path += '/'
+
+    doctest_namespace["data_dir"] = path
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -98,6 +98,44 @@ run allows us to generate a coverage report after that::
 The report is in ``report/index.html``, which links to individual
 files and shows exactly which lines were excuted while running the tests.
 
+Run doctests locally
+--------------------
+If `doctestplus <https://pypi.org/project/pytest-doctestplus/>` is installed
+(and it probably is because it's part of
+`sphinx-astropy <https://pypi.org/project/sphinx-astropy/>`,
+which is required to build the documentation locally), you can
+run the examples in the documentation or in the doctrings of individual
+functions, for example:
+
+   pytest --doctest-plus sherpa/astro/data.py
+   pytest --doctest-plus sherpa/data.py
+   pytest --doctest-rst docs/quick.rst
+   pytest --doctest-rst docs/evaluation/combine.rst
+
+This guarantees that the examples actually work and don't have typos or outdated
+parameters, which might confuse a user.
+
+Running `pytest --doctest-plus` or `pytest --doctest-rst` (without extra parameters)
+will run all usual tests and also the doctests on all files not specifically
+excluded in `pytest.ini`.
+If you fix examples to pass these tests, remove them from the exclusion list in
+`pytest.ini`! The goal is to eventually pass on all files.
+
+.. note::
+   The doctests are not run as part of the CI process.
+   The plan is to add them in once we have more experience with the system.
+
+Some doctests (in the documentation or in the docstrings of individual
+functions) load data files. Those datafiles can be found in the
+`sherpa-test-data <https://github.com/sherpa/sherpa-test-data>` directory
+as explained in the description of the :ref:`development build <developer-build>`.
+There is a `conftest.py` file in the `sherpa/docs` directory and in the `sherpa/sherpa`
+directory that sets up a
+pytest fixture to define a variable called `data_dir` which points to this directory.
+That way, we do not need to clutter the example with long filename, but the
+`sherpa-test-data` directory has to be present as a submodule to successfully pass all
+doctests.
+
 
 How do I ...
 ============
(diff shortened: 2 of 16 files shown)


docs/developer/index.rst
```diff
@@ -98,6 +98,44 @@ run allows us to generate a coverage report after that::
...
 files and shows exactly which lines were excuted while running the tests.
 
+Run doctests locally
+--------------------
+If `doctestplus <https://pypi.org/project/pytest-doctestplus/>` is installed
+(and it probably is because it's part of
+`sphinx-astropy <https://pypi.org/project/sphinx-astropy/>`,
+which is required to build the documentation locally), you can
+run the examples in the documentation or in the doctrings of individual
+functions, for example:
```
DougBurke, line 108:
```suggestion
functions, for example::
```

There's a problem with the current documentation since the following `pytest` lines do not get written as separate lines. I assume that adding a colon here will fix it, but this is just a guess...

screenshot

DougBurke: So, I ran

    pytest --doctest-plus

and it collected 7623 tests, which is the same number as I got from just `pytest`. Is this expected?

As a further check,

    pytest --doctest-plus sherpa/astro/data.py
    pytest sherpa/astro/data.py

both run 13 tests. So how do I know that it's actually running the extra tests - do I just have to assume it is?

docs/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+directory tree that is being tested. When we run doctests in the
+documentation only, it will therefore not find that fixtures that are
+defined in sherpa/conftest.py, thus this file copies over the fixtures we
+need.
+'''
+import pytest
+from sherpa.utils.testing import get_datadir
+
+@pytest.fixture(autouse=True)
+def add_sherpa_test_data_dir(doctest_namespace):
```
DougBurke, line 32: We have some documentation that calls routines that write to files - e.g. `sherpa.io.write_arrays`. Are these just going to be ones we skip, or can we handle this with some magic that runs the test in a temporary directory?

DougBurke: [I'm basically running `pytest --doctest-plus sherpa/random/file.py` and looking at issues !]

DougBurke:
I think the commits should also be squashed as there's no real benefit to having "revert of previous commit" and then a few fix-up commits in our history.

hamogu:
I have a strong preference to leave the first commit untouched, because it's not hand-written. It's generated using advanced git-foo that I don't exactly understand: It's made by reverting the commit that backed out my first attempt at doing this. If there is ever any mix-up in eg. re-basing this PR, then it's going to be way easier for me to handle when the "revert" and the hand-written, new stuff are separate. Comments are not expensive. I'm happy to squash all other commit that are in this PR and that I might be adding the cause of finishing and getting this reviewed.

docs/developer/index.rst
hamogu: To answer the first part of this: With the changes in this PR, doctest-plus is automatically activated for all files that are NOT on the doctestplus blacklist. You can see that the number of tests increases compared to main.
Note that the number shown in my output below is different from what @DougBurke said above, but that's OK because (a) the number of tests run depends on the plotting/IO backend and (b) the number of tests has changed in main when I just rebased locally (not pushed yet). The point is that the settings in this PR increas the number of tests compared to main. The increase is 6, because there are 6 files in the rst files in the docs that are **NOT** blacklisted in `pytest.ini` and doctest run each rst file as a single test.
On main
```
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> pytest
================================================================================================== test session starts ==================================================================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/guenther/code/sherpa, configfile: pytest.ini
plugins: doctestplus-0.11.2, typeguard-2.13.3, xvfb-2.0.0
collecting ... WARNING: Failed to import any usable sherpa.plotting backend. Plotting routines will not be available.
collected 7547 items       
```

On this branch:
```
(sherpaalone) MoritzAirRoseGold ~/code/sherpa> pytest 
================================================================================================== test session starts ==================================================================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/guenther/code/sherpa, configfile: pytest.ini
plugins: doctestplus-0.11.2, typeguard-2.13.3, xvfb-2.0.0
collecting ... WARNING: Failed to import any usable sherpa.plotting backend. Plotting routines will not be available.
collected 7555 items                                                                                         ```

**Comment by hamogu** [comment]
On the other hand, in CI, there is no change, e.g. compare: https://github.com/sherpa/sherpa/actions/runs/4286652711/jobs/7466499212#step:13:120 and https://github.com/sherpa/sherpa/actions/runs/4287930568/jobs/7469345751#step:13:120. This could be changed easily, just by adding doctestplus to the "install-required", but given the problems we had with this PR in an earlier version, I'd like to leave that for a separate PR.

File: docs/developer/index.rst [hunk 3]
```diff
@@ -103,9 +103,25 @@ Run doctests locally
...
+However, many examples were written because doctestplus was used with Sherpa and do not execute
+as written (e.g. they need implicit set-up that's not spelled out),
+so a number of rst files are explicitly excluded from that run in `pyests.ini`.
+When an rst file is fixed to work top-to-bottom, it should be removed from the
+blacklist in `pytest.ini`.
+
+doctestplus can also run the examples in the doctrings of individual
+functions, but again many files are still blacklisted for this functionality in
+`pytest.ini`.
+During development, you can run doctestplus on individual files like so::
```
DougBurke, line 124: This needs some work. Let me suggest updates after the coffee has kicked in and then left my system...

docs/developer/index.rst
```diff
@@ -98,6 +98,53 @@ run allows us to generate a coverage report after that::
...
```
DougBurke, line 124:
```suggestion
This serves two purposes:

  - it ensure that the examples we give are actually correct and match the code,
  - and it acts as additional tests of the Sherpa code base.

The `doctest_norecursedirs` setting in the `pytests.ini` file is used to exclude files which can not be
tested. This is generally because the examples were written before doctestplus support was added, and so
they need to be re-worked, or there is too much extra set-up required that would make the examples
hard-to follow. The file should be removed from this list when it has been updated to allow testing with doctestplus.

During development, you can run doctestplus on individual files like so (the option to use depends on whether it is a Python or reStructuredText file)::
```

Is this better? I feel like we don't need to strongly explain why testing is good.

docs/developer/index.rst
```diff
@@ -98,6 +98,53 @@ run allows us to generate a coverage report after that::
...
+
+   pytest --doctest-plus sherpa/astro/data.py
+   pytest --doctest-plus sherpa/data.py
+   pytest --doctest-rst docs/quick.rst
+   pytest --doctest-rst docs/evaluation/combine.rst
+
+This guarantees that the examples actually work and don't have typos or outdated
+parameters, which might confuse a user.
```
DougBurke, line 132:
```suggestion
```

I think this sentence is superfluous.

docs/conftest.py
hamogu: I have not spend too much thought on that, because I've not chosen one of those files as an example to test here. However, I'm not aware of magic that will do things just magically by just writing a few lines in this file.

The easiest is to just write to the current working directory. That should not be a problem in RTD or GH CI, because each build/test is done in a new, clean container. However, if you run the tests locally you would end up with artifacts lying around (unless you you tox and friends for local testing, which I don't).

In rst files, we can use the following directives (from https://github.com/astropy/pytest-doctestplus)
```
.. testsetup::

    >>> x = 42
```
and
```
.. testcleanup::

    >>> del x
```

They could be used to set the current working dir to a temporary directory and clean it up afterwards. Since each rst file is run as one test, `testsetup` would be inserted at the top and `testcleanup` at the bottom. Sphinx does not render any visible output for these directives. (Files may still end up laying around if the test fails in the middle and never gets to the clean up, but then that file needs to be looked at by hand anyway.)

In docstrings for individual functions that is another matter, we don't want our users to be bothered by sphinx directives. I think pytest setup and teardown (https://docs.pytest.org/en/6.2.x/xunit_setup.html) should work, but I have not tested that. If it's just one line that write a file and file is not needed further down, I would either skip that line (` # doctest: +SKIP`) or the entire doctest (see "skipping doctests" in https://github.com/astropy/pytest-doctestplus). 

While there are great advantages to running the doc examples with doctests and making sure they work, we don't have to bend over backwards if it's just not a good fit; the examples in the docs are examples first and tests only as far as it's useful.

DougBurke: I'm just trying to understand the const-benefit ratio of this (ie how far do we go). Ta.

DougBurke (approved):
LGTM

hamogu:
About leaving files around locally after running doctests, see discussion in https://github.com/astropy/pytest-doctestplus/issues/68

Issue #1225, opened 2021-07-26
XSPEC: change from FORTRAN/C to C++ for utility routines 


DougBurke:
The XSPEC interface provides a FORTRAN/C interface (using `xsFortran.cxx/h` which provides routines like `FPSOLR`, but if you look at the code for this routine you see it's provided by the C++ `FunctionUtility` class (see below).

We currently use the `xsFortran.h` interface, but we should be able to directly use `FunctionUtility` which would provide some extra capability (the ability to set/query/delete the `XFLT` keywords needed by mixing models such as `projct` or the `smaug` additive model is what I am after)

```C++
void FPSOLR(const char* table, int* ierr)
{
   string tableName = string(table);
   tableName = XSutility::lowerCase(tableName);
   if (tableName == "file")
   {
      FunctionUtility::ABUND(tableName);
      *ierr = 0;
   }
   else
   {
      if (FunctionUtility::checkAbund(tableName))
      {
         FunctionUtility::ABUND(tableName);
         *ierr = 0;
      }
      else
      {
         *ierr = 1;
      }
   }
}
```

Technically it should be easy (as we have `xsFortran.cxx` to base it on). In reality ...

DougBurke:
With XSPEC 12.12.0 this should be easy, since the include files are now installed with the correct layout (as far as my initial tests show) but it's not possible with earlier versions of XSPEC (unless we were to change to using the build location and not install location of XSPEC for the `xspec_include_dirs` configuration option).

DougBurke:
I have a version at https://github.com/DougBurke/sherpa/tree/xspec-move-from-xsfortran which makes this change - but does not take advantage of the extra functionality. I expect the build to fail on GitHub actions as they use XSPEC versions prior to 12.12.0 and so can only be built with

    xspec_include_dirs = <build>/Xspec/src <build>/Xspec/src/include

which we don't have packaged up in conda (and we shouldn't have).

I do think we should byump our minimum-supported XSPEC version as we can no-loniger guarantee we can still build with XSPEC 12.9.0 (for instance) and I would like to use functionality added in later versions. But that's a different ticket...

Issue #544, opened 2018-11-10
XSPEC: how do we support the `spectrumNumber`/`ifl` argument?


DougBurke:
XSPEC maintains a mapping between input files (using a simple integer scheme, starting at 1, unlike our int or string identifier) and metadata (primarily the `XFLTxxxx` keywords). Models are sent in the spectrum number, along with the grid, which they can use to access this metadata.

Sherpa currently ignores this, which means we can not support models such as `smaug` and `projct` (and the only way to know which models have to be skipped is by manual inspection of the documentation or code).

There's at least three parts:

- support sending in additional metadata when evaluating a grid, and then passing that through to the model (in this case the spectrum number)
- a mapping between the consecutive integer value XSPEC needs and the free-form Sherpa identifiers (e.g. I may have labelled them "ann1", "ann2", ...), **or** we say that when using these particular XSPEC models the user has to use identifiers 1 to n.
- extracting the relevant metadata from the input PHA files and calling the appropriate XSPEC library functions to set up the information (I think this has been possible since about XSPEC 12.9 for the XFLT keywords, I don't know if there are other keywords that need looking for)

PR #1615, opened 2022-10-19
Take advantage of XSPEC 12.12.0

DougBurke:
Some of these changes have been moved to (in a form that isn't necessarily quite as useful as in this PR, but they don't need C++ changes so are easier to review):

- #2107
- #2109

and this commit has not been updated to reflect that yet. I have pulled out other changes to

- #2215
- #2216 

# Summary

Add support for new functionality in XSPEC provided by XSPEC 12.12.0, such as setting abundance values via atomic number or as an array, and access to the atomdb and nei version settings, XFLT keywords, and the Db database.

# Details

This is built on #1793 which switches from the xsFortran API to the FunctionUtility one. This means that we can just concentrate on new functionality here.

Start taking advantage of the features provided by XSPEC 12.12.0

- use FunctionUtility rather than the xsFortran interface (see #1225)
  - `set_xsabund` can now be sent an array of abundances, not just the name of a table or a filename
  - `get_xsabund` can now be sent the atomic number, as well as the name, and the (optional) table to search
  - `get_xsabundances` has been added, which returns all the abundances for a table as a dictionary, and `set_xsabundances` which takes a dictionary
  - `get_xsxset` now behaves like a dictionary when a keyword does not exist - it raises a `KeyError` rather than returns the empty string
  - add access to the XFLT keyword database with `clear_xsxflt`, `set_xsxflt`, and `get_xsxflt`
    - this is needed for some models, such as `smaug` and the new polarimetry models, although more work is needed to support those models (in particular the `spectrumNumber` value, as discussed in #544)
  - add access to the XSPEC "Db" database (mapping from string to double values), although it's not at all clear how users are meant to use this (or if they are), with `clear_xsdb`, `set_xsdb`, and `get_xsdb`
  - setting the model database path
  - get/set the atomdb/nei versions with `get_xsversion` and `set_xsversion` (the latter is new)
  - the XSPEC state now stores/restores more settings
  - I have added a low-level `get_xsabund_doc` call to get the "nice" name of the abundance table (e.g. `Anders E. & Grevesse N. Geochimica et Cosmochimica Acta 53, 197 (1989)` for the `angr` table)

This is all pretty "low level" stuff, so very-few users are expected to take advantage.

One change I have added is to support the use of `Path` when specifying file / directory names - rather than just using a string. The idea is that you can use either. We can also think about returning `Path` objects: in this PR a new low-level routine does this, but existing routines (which currently return `str` but could easily be changed) have not been updated.

I have added a `show_xsabund` call to the astro UI layer since I noted that XSPEC lists the actual abundances with its `show abund` call, so we now have (taking advantage of these lower-level routines):

```
>>> show_xsabund()
Solar Abundance Table:
angr:  Anders E. & Grevesse N. Geochimica et Cosmochimica Acta 53, 197 (1989)
  H : 1.000e+00  He: 9.770e-02  Li: 1.450e-11  Be: 1.410e-11  B : 3.980e-10
  C : 3.630e-04  N : 1.120e-04  O : 8.510e-04  F : 3.630e-08  Ne: 1.230e-04
  Na: 2.140e-06  Mg: 3.800e-05  Al: 2.950e-06  Si: 3.550e-05  P : 2.820e-07
  S : 1.620e-05  Cl: 3.160e-07  Ar: 3.630e-06  K : 1.320e-07  Ca: 2.290e-06
  Sc: 1.260e-09  Ti: 9.770e-08  V : 1.000e-08  Cr: 4.680e-07  Mn: 2.450e-07
  Fe: 4.680e-05  Co: 8.320e-08  Ni: 1.780e-06  Cu: 1.620e-08  Zn: 3.980e-08
``` 

# Issues

What are the best names / API here? I've tried to fit into the existing settings (e.g `get_xsXXX` and `set_xsXXX`) where appropriate, but have tried to make use of optional arguments and using dictionaries rather than map closely to either the `xsFortran` interface or the newer `FunctionUtilities` interface (note that there are differences between the two, e.g. the old interface, and our code, has "xset" whereas the C++ code talks about the model string database).

Unfortunately, just providing access to the XFLT database is not sufficient, we need to work out how to pass around the `spectrumNumber`/`ifl` argument to models and to map between data objects and these identifiers (see 

- #544
 
). The logic for this (or at least the initial support for this) has been pushed to 

- #1617
 
(and doesn't currently require this PR in an attempt to avoid ridiculously-long chains of PRs as they are already hard enough to handle as is).

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -811,7 +811,7 @@ def test_get_xsstate_keys():
     ostate = xspec.get_xsstate()
     assert isinstance(ostate, dict)
 
-    for key in ["abund", "chatter", "cosmo", "xsect",
+    for key in ["abund", "chatter", "cosmo", "xsect", "xflt",
                 "modelstrings", "paths"]:
         assert key in ostate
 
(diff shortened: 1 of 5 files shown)


DougBurke:
Rebasing now that a number of PRs have landed (in particular we now have XSPEC builds with CCfits installed similar to a HEASOFT installation and recen releases to test against, so we may not need the work-arounds I had previously added).

DougBurke:

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -402,6 +402,86 @@ def test_abund_element_z():
...
+    # testing the XSPEC implementation itself, just our use of it.
+    #
+    h = xspec.get_xsabund('H', 'wilm')
+    he = xspec.get_xsabund('He', 'wilm')
+    si = xspec.get_xsabund('Si', 'wilm')
+    ar = xspec.get_xsabund('Ar', 'wilm')
+    k = xspec.get_xsabund('K', 'wilm')
+    fe = xspec.get_xsabund('Fe', 'wilm')
+
+    assert h == pytest.approx(1.0)
```
DougBurke, line 425: The test logic should be moved into a subroutine as it's used three times and we want to avoid repetition and make it more obvious where the same thing is being checked

DougBurke:
Rebased to address a conflict in a test file.

DougBurke:
Rebasing because of #1998 being merged.

DougBurke:
Rebased as https://github.com/sherpa/sherpa/pull/1997 has been merged (there's no conflicts, but I just want to make sure we test the latest version).

DougBurke:
Rebased on top of #2052

DougBurke:
Rebased to address a conflict.

DougBurke:
I have done a quick rebase to match 4.17.0 and so address some of the conflicts (since I added some of the functionality here in 4.17.0 using the old FORTRAN-style API from XSPEC).

**However** I have not gone through to check whether the commits make sense - some of them will not and will need to be cleaned up, rewritten, or removed. The docstrings will also need updating to get the right `versionadded`/`versionchanged` information.

DougBurke:
I am leaving this open for now, as a memory of some changes that could be made, but it should be reviewed to see if there's anything here that isn't in other (also moth-balled) PRs.

PR #1617, opened 2022-10-27
Pass keyword arguments to models

DougBurke:
# Summary

Allow keyword arguments to be passed to models and ensure these arguments are respected by the model cache.

# Details

There is a kind-of follow-up PR: 
- #1671

There is currently no use case that requires passing keyword arguments to models, but

1. it's something that might be useful
2. we partially support it already
3. there is a use for XSPEC models but this requires something like #1615 to have been merged (to allow the `spectrumNumber`/`ifl` keyword to be used to set up data)
4. it turns out we need to deal with kwargs for XSPEC models to handle model regridding (see #830) 

One of the commits actually allows us to pass the `spectrumNumber` value to XSPEC models. It currently does nothing, because

a) we have no models that use this
b) we have no way to pass this information around either at the UI layer or in the object layer (e.g. during a `Fit` evaluation)
c) we have no way to set the data that the models need

An alternative to `spectrumNumber` as the argument name is `ifl`, which is the 'older' term that XSPEC uses, but as this is a low-level feature users are not expected to use directly [although currently how we plan to expose this to users has not been worked through] I decided to go with the more-descriptive term.

As part of this work the documentation has seen some minor updates. 

# Cache

We have a cache system around 1D models so that when sent a set of parameters and a grid we can avoid re-calculating the model, and just return the cache (with a very-small cache size so it's only really helpful when the same model is being evaluated for different datasets that have the same RMF grid, or at least that's the intent; I find it hard to validate the behavior of this code but that's a different concern). The cache is controlled by the `modelCacher1d` decorator and the logic is in the `sherpa.models.model.ArithmeticModel` class. Fortunately we now have some documentation on this I can point to - https://sherpa.readthedocs.io/en/4.15.0/evaluation/cache.html

Keyword arguments to a model need to be added to the cache key (otherwise what is the point of sending them in). This is relatively easy to do but it probably limits the values that can be used for these keywords - the current version assumes we can convert them to a `ndarray`:

```python
# Add any keyword arguments to the list. This will
# include the xhi named argument if given. Can the
# value field fail here?
#
for k, v in kwargs.items():
    data.extend([k.encode(), numpy.asarray(v).tobytes()])
```

You can get away with passing simple tuples and dicts to `asarray`, so it's not as restrictive as it would first seem. An alternative would be to come up with some way to create a `key` from a set of values (I'm sure there are many Python packages to do so but we don't want to be pulling in extra code to do so).

```
>>> import numpy as np
>>> np.asarray(True).tobytes()
b'\x01'
>>> np.asarray("x").tobytes()
b'x\x00\x00\x00'
>>> np.asarray(("x", True)).tobytes()
b'x\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00T\x00\x00\x00r\x00\x00\x00u\x00\x00\x00e\x00\x00\x00\x00\x00\x00\x00'
>>> np.asarray({"x": True}).tobytes()
b'\x00\xd8V\n#\x7f\x00\x00'
```

# Notes

This somewhat conflicts with existing XSPEC PRs, some of which have now been merged

- #1396
- #1615
- #830 (maybe)

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016-2018, 2019, 2020, 2021, 2022
+#  Copyright (C) 2016-2018, 2019, 2020, 2021, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -1593,3 +1593,50 @@ def test_model_can_send_spectrumnumber_combine_non_xspec():
 
     assert y21 == pytest.approx(y12)
     assert y12 == pytest.approx([2, 2, 2])
+
+
+@requires_xspec
+def test_xspec_model_kwarg_cached_not_needed():
+    """Does the cache recognize when a kwarg has changed?
+
+    This is for a XSPEC model that is known to not depend on the
+    spectrumNumber/ifl argument (i.e. the definition in model.dat has
+    a second 0 after the model type, in this case saying "add 0 0" or
+    "add 0" as the 0 is inferred).
+
+    At the moment Sherpa does not wrap any of the models (smaug,
+    polcost, pollin, polpow, pileup) which have a 1 instead of a 0.
+
+    For this case we test the current behavior (which ignores the
+    spectrumNumber argument), as it is not clear what the best
+    behavior is.
+
+    """
+
+    from sherpa.astro import xspec
+
+    mdl = xspec.XSpowerlaw()
+
+    # check the cache is working
+    assert mdl._cache_ctr['hits'] == 0
+    assert mdl._cache_ctr['misses'] == 0
+
+    egrid = np.arange(0.2, 0.6, 0.1)
+    elo = egrid[:-1]
+    ehi = egrid[1:]
+
+    y1 = mdl(elo, ehi)
+    assert mdl._cache_ctr['hits'] == 0
+    assert mdl._cache_ctr['misses'] == 1
+
+    y2 = mdl(elo, ehi, spectrumNumber=2)
+    assert mdl._cache_ctr['hits'] == 1
+    assert mdl._cache_ctr['misses'] == 1
+
+    assert y2 == pytest.approx(y1)
+
+    y3 = mdl(elo, ehi, spectrumNumber=1)
+    assert mdl._cache_ctr['hits'] == 2
+    assert mdl._cache_ctr['misses'] == 1
+
+    assert y3 == pytest.approx(y1)
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1608,7 +1608,7 @@ def test_xspec_model_kwarg_cached_not_needed():
     value, since it does not change the model output, but it is tricky
     to do, so we do make each call be different. This is a
     pessimisation (i.e. it potentially slows down model evaluation)
-    and could be changed in the future). It depends on how often the
+    and could be changed in the future. It depends on how often the
     spectrumNumber argument will be added to model calls.
 
     """
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1604,9 +1604,12 @@ def test_xspec_model_kwarg_cached_not_needed():
     At the moment Sherpa does not wrap any of the models (smaug,
     polcost, pollin, polpow, pileup) which have a 1 instead of a 0.
 
-    For this case we test the current behavior (which ignores the
-    spectrumNumber argument), as it is not clear what the best
-    behavior is.
+    Technically we could make the cache ignore the spectrumNumber
+    value, since it does not change the model output, but it is tricky
+    to do, so we do make each call be different. This is a
+    pessimisation (i.e. it potentially slows down model evaluation)
+    and could be changed in the future). It depends on how often the
+    spectrumNumber argument will be added to model calls.
 
     """
 
@@ -1627,13 +1630,13 @@ def test_xspec_model_kwarg_cached_not_needed():
     assert mdl._cache_ctr['misses'] == 1
 
     y2 = mdl(elo, ehi, spectrumNumber=2)
-    assert mdl._cache_ctr['hits'] == 1
-    assert mdl._cache_ctr['misses'] == 1
+    assert mdl._cache_ctr['hits'] == 0
+    assert mdl._cache_ctr['misses'] == 2
 
     assert y2 == pytest.approx(y1)
 
     y3 = mdl(elo, ehi, spectrumNumber=1)
-    assert mdl._cache_ctr['hits'] == 2
-    assert mdl._cache_ctr['misses'] == 1
+    assert mdl._cache_ctr['hits'] == 0
+    assert mdl._cache_ctr['misses'] == 3
 
     assert y3 == pytest.approx(y1)
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1534,7 +1534,6 @@ def test_model_can_send_spectrumnumber_combine():
     assert args[2][1] == "con"
 
 
-@pytest.mark.xfail
 @requires_xspec
 def test_model_can_send_spectrumnumber_combine_non_xspec():
     """Check the spectrumNumber is sent through with non-XSPEC models.
@@ -1585,7 +1584,6 @@ def test_model_can_send_spectrumnumber_combine_non_xspec():
 
     args.clear()
 
-    # This fails
     y21 = comb21(elo, ehi, spectrumNumber=4)
     assert len(args) == 1
     assert args[0][0] == 4
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1380,7 +1380,6 @@ def test_xsparameter_limits(base):
     assert p.hard_max == pytest.approx(8)
 
 
-@pytest.mark.xfail
 @requires_xspec
 @pytest.mark.parametrize("clsname", ["XSpowerlaw", "XSphabs"])
 def test_model_can_send_spectrumnumber_indiv(clsname):
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1378,3 +1378,218 @@ def test_xsparameter_limits(base):
 
     assert p.max == pytest.approx(8)
     assert p.hard_max == pytest.approx(8)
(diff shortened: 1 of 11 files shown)


hamogu:
Depending on how exactly this ends up being designed, this may work towards #1651. That issue does have a real use case today - and again we want to the model to make use of some additional information that's in the header of the PHA file (the MJD-OBS in that case). 

That could be addressed by allowing a keyword on model creation as in this PR - but it would be even better to read the keyword at run-time so that the same model can be used to model different spectra and I don't have to create a new model instance for every spectrum to bake in that keyword at model creating time. The latter is probably beyond the scope for this PR, but it does not hurt to keep it in mind when looking at this.

DougBurke:
Hmm. This interacts interestingly with the regrid code (in particular, adding support for regrid to XSPEC models - see #830).

hamogu:
Is this read for a review, but you just did not request my review to be polite? Or is there something still missing? (Trying to get through my reviews in preparation for the next meeting...)

DougBurke:
It's ready for review. I just don't like to overload people with too-many review requests. There is some interaction with the regrid/integrate-flag work I sent you and Aneta a document about recently, related to #830, but I don't think that should hold this work up)

hamogu:
This is the PR where the following comment was supposed to go:

requesting @wmclaugh as a reviewer as well, because I'm not comfortable to sign off on the xspec_extension.hh changes myself.

DougBurke:
For reference, I do not see this as being a lien on the 4.15.1 release. It would be nice to have, but it doesn't add any new functionality to users as is (requires more changes first).

sherpa/models/model.py
```diff
@@ -129,8 +130,8 @@
 ------------------
 
 One parameter can be made to reference one or more other parameters, a
-process called "linking". The lniked is no-longer considered a free
-parameter in a fit since it's value is derived from the other
+process called "linking". The linked parameter is no-longer considered
+a free parameter in a fit since it's value is derived from the other
```
hamogu, line 134:
```suggestion
a free parameter in a fit since its value is derived from the other
```

sherpa/models/model.py
```diff
@@ -387,75 +392,95 @@ def modelCacher1d(func):
...
 
         return vals
 
     return cache_model
 
+# It is tempting to convert the explicit class names below into calls
+# to super(), but this is problematic since it ends up breaking a
+# number of invariants the classes rely on. An example is that
+# instances of Unary/BinaryOpModel classes should not cache-related
+# attributes, but they can do if we change to using super. There is
```
hamogu, line 478: I don't understand this sentence. Do they cache related attributes (relate to what?) or or is this about attributes related to the cache (in that case, there is a verb missing in this sentence)?

sherpa/models/model.py
```diff
@@ -387,75 +392,95 @@ def modelCacher1d(func):
...
+# mode discussion of this in
```
hamogu, line 479:
```suggestion
# more discussion of this in
```

sherpa/models/model.py
```diff
@@ -877,11 +909,17 @@ def __init__(self, name, parts):
 
             for p in part.pars:
                 if p in allpars:
-                    # If we already have a reference to this parameter, store
-                    # a hidden, linked proxy instead
+                    # If we already have a reference to this
+                    # parameter, store a hidden, linked proxy
+                    # instead. This is presumably to ensure that we
+                    # have the correct number of degrees of freedom
+                    # (as pnew is frozen) while stil sending the
```
hamogu, line 916:
```suggestion
                    # (as pnew is frozen) while still sending the
```

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -1378,3 +1378,265 @@ def test_xsparameter_limits(base):
...
+@pytest.mark.parametrize("clsname", ["XSpowerlaw", "XSphabs"])
+def test_model_can_send_spectrumnumber_indiv(clsname):
+    """Check we can send spectrumNumber to individual models.
+
+    All we do is check that the model can be called, and just for
+    additive/multiplicative models that are not expected to use
+    the spectrumNumber argument.
+    """
+
+    from sherpa.astro import xspec
```
hamogu, line 1393: I recently learned about [pytest's `importorskip`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip). I'm not suggesting to change the code, I'm just noticing right now (maybe it's been obvious for you for a long time), that that could replace some of our own requires_X decorators if we aver want to. Not saying that we should, as they work fine, I'm just noticing.

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -1378,3 +1378,265 @@ def test_xsparameter_limits(base):
...
+    "add 0" as the 0 is inferred).
+
+    At the moment Sherpa does not wrap any of the models (smaug,
+    polcost, pollin, polpow, pileup) which have a 1 instead of a 0.
+
+    Technically we could make the cache ignore the spectrumNumber
+    value, since it does not change the model output, but it is tricky
+    to do, so we do make each call be different. This is a
+    pessimisation (i.e. it potentially slows down model evaluation)
+    and could be changed in the future). It depends on how often the
```
hamogu, line 1611:
```suggestion
    and could be changed in the future. It depends on how often the
```

hamogu (commented):
I've marked up a few typos and small stuff, but my main comment is conceptual and should be discussed before we tick off all the details.

As designed here, every model component gets every keyword that is set. Keywords that are not accepted by models are just cleaned out and ignored with `clean_kwargs1/2d`. I guess that's fine for a small number of keywords.

However, I wonder if that boxes us into a corner. What if model A requires the keyword POL='U' (e.g. for the u component of a polarization vector) and model B requires POL='1' (e.g. for the first component if the IQU polarization data)? In other words, this implementation will run into problems once there are name clashes in the keyword arguments. "Namespaces are honking good idea" says the Zen of Python. "let's do more of those".

So, one might consider specifying keywords in a more complex way that makes it easier to apply certain keywords only to some models (think list with elements `(keyword, val, [list-of-model-names-this applies-to])`. However, is that extra complexity worth it?

For now, we have very, very few use cases of passing in parameters, so maybe names clashes are not our first concern. However, with #1651 in mind, I can see a different situation that exposes the same name-clash problem: Imagine we want to fit several datasets at once and need to set a different `SpectrumNumber` for each. How would one do that?

I understand that this PR is on step in the implementation, that needs to be followed by an increase in XSPEC version et.c to be useful. So, some of those thought might be out of scope for this PR, as @DougBurke says in the summary of the PR ("There is currently no use case that requires passing keyword arguments to models.") but I want to make sure that we think through those situation before we accept this implementation.

Obviously, @DougBurke has thought about this more then I have, but on top of my head, I would have suggested to make those extra arguments a property of the model, as `self.keywords = {}`. That way, the user can set them before the evaluation on a per model bases (thus no need to worry about name clashes). That could be combined with the mechanism here (if keywords are set, use those, but override if passed in like in this PR).

Or maybe I don't understand how this is supposed to be working?

DougBurke:
Thinking aloud in response to @hamogu (and no conclusion reached):

Is there a case where we might want the same keyword passed to multiple components of a model - that is, when you have

   src = cpt1 * (cpt2 + cpt3)

and you evaluate `src` and you want to set `spectrumNumber=2` for use by both `cpt2` and `cpt3` but it's going to get ignored by `cpt1`?

I picked `spectrumNumber` as this is one place I could imaging it happening (i.e. have a concrete use case), but to be honest I can't think of a case where you'd have multiple XSPEC models that use this functionality that you would want to combine. At least these would be consistent (ie mean the same thing) and not have the potential to conflict which is likely with the current implementation.

I do admit that I am not particularly proud of the `clean_kwargs1/2d` solution I came up with.

sherpa/astro/xspec/tests/test_xspec_unit.py
DougBurke: Extending this detour, I have, in another commit somewhere (for the life of me I don't know in what PR), a fixture that does the 'from sherpa.astro import xpec' line for you (and returns the model component you asked for, so it may or may not be relevant here). That could probably benefit from `importorskip`, as could probably the `requires_xxx` routines (although I'm not sure the semantics actually matches this routine, would need to think about, as our requires_xxx stuff allows for a choice of modules, but this is getting far away into the weeds).

I'm going to resolve this as it's basically to say "don't forget importorskip".

DougBurke:
So, I don't  think it's worth @wmclaugh reviewing this at this time as we still have work to do (which probably won't change the xspec_extension.hh but I can't guarantee that).

hamogu:
> Is there a case where we might want the same keyword passed to multiple components of a model - that is, when you have
> 
> src = cpt1 * (cpt2 + cpt3)
> 
> and you evaluate src and you want to set spectrumNumber=2 for use by both cpt2 and cpt3 but it's going to get ignored by cpt1?

Absolutely. That's the use case I have in mind. Think of a IXPE spectrum with the following model: xsphaabs * (polarized_polwerlaw1 + polarized_powerlaw2). The powerlaws "create" the X-rays, they do need to know if I'm asking for the I, Q, or U spectrum. On the other hand, the absorption model is multiplicative, represents the matter between us and the source, and does not care about the polarization direction I'm trying to fit. 

Note 1: We don't currently have models like `polarized_powerlaw` but #1649 shows a proof-of-concept on how they could be implemented.

Note 2: My example might be better served by an implementation where the `polarized_powerlaw` reads a property/keyword/header info that's in the data then by passing keywords to the model from outside, but I feel those two cases overlap enough that we should consider both as we design this case.

DougBurke:
One problem is that models, at the end of the day,  are evaluated without a
data object . You just send in the grid/axes.

We could require a data object,  but you'd have to do a very good job to
convince me it would be worth it.

Now,  we could require an axis object for evaluating a model (there s
reasons this could be good) and maybe we could associate values with the
axes objects,  but then we are back to the same issue as the current
design: keywords are currently chivalry rather than being scoped to a
model.

hamogu:
>  One problem is that models, at the end of the day, are evaluated without a
data object . You just send in the grid/axes.

But, we could (conceptually, have not looked at the actual code right now) pass in a grid and options. That grid is in some way connected to the data object, so when that grid is taken, we would have to also get keywords/metadata and pass those along with it all the time. But that's somewhat forced, too.

Then maybe it could be a property of the model that's set at model creation time or any later time that user wants to change it and, if not passed in in that call itself, it can be taken from the model. That way, one could make it persistent and set is on a per-model basis fi needed. Would that work?

hamogu:
I want to make a proposal: We leave the scope of this PR as it is and merge it soon (pending a review from @wmclaugh for the C code).
We then follow-up with another PR that ads the following: If a model wants to have a keyword, but it's not passed in, it's taken from `model.keywords`. If we agree to that scheme, my use cases can be addressed and this PR with it's more narrow scope can be merged now. I think we had enough discussion to convince me that the current PR does not prevent us from doing that (or another more granular scheme) in the future.

PR #1618, opened 2022-10-28
Tweak ds9 tests

DougBurke:


DougBurke:
@harlantc and @Marie-Terrell - I've created this PR basically for myself, as it is something that annoys me no end when I'm working-from-work. It's not directly CM related, as it should not affect how you build/test Sherpa for standalone or CIAO purposes, but I'd appreciate a review.

sherpa/conftest.py
```diff
@@ -83,17 +83,32 @@ def pytest_addoption(parser):
...
+        skip = pytest.mark.skip(reason=f"need {opt} option to run")
+        for item in items:
+            if label in item.keywords:
+                item.add_marker(skip)
+
+    # Skip tests if --skipxxx given in cli
+    #
+    for label in ["ds9"]:
+        opt = f"--skip{label}"
         if config.getoption(opt):
```
hamogu, line 108: Maybe I'm just a bit stupid, but should this not be the inverse? If "skipds9" is set, then we continue, but if it is not set, then `pytest.mark.skip(reason=f"{opt} was set")`? That looks the same as above for the "runxxx" tests and I would have expected a logical not (`~`) somewhere in this section. Or do I need to start drinking coffee?

hamogu: However, we don't have to spend time to understand this. If it works, it works; let's be efficient and get this in instead of waiting for one of us to make perfectly engeneered code for something that's such a special case about just invoking tests.

DougBurke:
@wmclaugh asked if we could make Sherpa run with multiple DS9 instances, which is another way of addressing this. I need to look back at the DS9 interface to see what would be needed.

DougBurke:
Rebased as #1697 has been merged

anetasie:
@DougBurke did you plan to include this PR in the 4.16 release?

DougBurke:
No.

Issue #1620, opened 2022-11-01
table model review


DougBurke:
I've been looking at the table-model code as a by-product of recent work (eg #1617) and there are some things to think about, including (but probably not limited to)

- it supports `x=None` case, which appears to be primarily used to let 2D images (eg exposure maps) be used, but it's all horribly implicit (and undocumented)
- if `x` is given then we sort the data; is this really something we should be doing
- it doesn't support integrated models, which I guess makes sense, but should we error out if `integrate` is set
- `ndim` is `None` so we can't enforce a dimension check
- there's a woeful lack of basic tests, and if you add some then you find that the error conditions fall over because of logic errors
- it appears to have been written assuming the user is only ever going to interact with it via the UI layer, and not as a separate model, so things like "oh, if you reload the data then the previous `fold` call should be forgotten" does not happen

DougBurke:
I wouldn't say your use is a "niche" case, as it is one of the ways it's meant to work..

So, your use here is with `mdl.load(None, ymodel)`. The code also supports `mdl.load(xmodel, ymodel)`, in which case it will interpolate the dataspace grid onto `(xmodel, ymodel)`. I wonder if we should separate the two options, into different classes? I also wonder how much of the `regrid` code we could use (although this code was written before the regrid code, and maybe we don't want the extra complexity)

hamogu:
That usage confused me initially. It would help to split into two models: `TableModelInterpolatedOnX` and `TableModelExactBins` (names open to bike-shedding). The old `TableModel` could remain in place as a wrapper for backwards compatibility. 
the "exact bins" version can e integrated or not - it's up to the user to provide the right bin values. In that case, maybe we find a way to get rid of the "fold" call, which is pretty annoying. It also would never regrid. The "interpolated" version could gain the ability to `integrate` (integrate the spline or linear interpolation or whatever we use over the bin) and regrid freely like any other arithmatic model, that's what I would expect it to do as a user.

DougBurke:
Just so I don't forget: there's an argument to be had that if we do split the behavior into different classes then the old `load(None, y)` case should not be cached - the model is so simple it is likely not worth the cost of caching the data. Partly influenced by #1622

There's also the fact that the `method` argument should be a key for the cache when `load(x, y)` is used.

PR #1621, opened 2022-11-01
Fix a cache error with TableModel

### LINKED ISSUE #1622: tablemodel load does not cache data
Found as part of #1621. I've realised that the existing tablemodel code does not clear the cache when `load` is called, which means that

```
>>> from sherpa.models.basic import TableModel
>>> mdl = TableModel()
>>> mdl.load([1, 2, 3], [4, 5, 6])
>>> mdl([1, 2, 3])
array([4., 5., 6.])
>>> mdl([1, 2, 3])
array([4., 5., 6.])
>>> mdl.load([1, 2, 3], [100, 200, 300])
>>> mdl([1, 2, 3])     # THIS IS WRONG
array([4., 5., 6.])
```

You can fix this by turning off the cache by setting `_use_caching` to `False`

```
>>> mdl._use_caching = False
>>> mdl([1, 2, 3])  # THIS IS CORRECT
array([100., 200., 300.])
>>> mdl([1, 2, 3])
array([100., 200., 300.])
>>> mdl.load([1, 2, 3], [10, 20, 30])
>>> mdl([1, 2, 3])  # THIS IS CORRECT
array([10., 20., 30.])
```

DougBurke:
# Summary

Fix cache errors with the TableModel class (#1622), add validation of input values, and minor additions to the documentation.

# Details

See #1620 for background, but that includes some possible longer-term changes (eg splitting out the "interpolation" from "non interpolation" models) that need more discussion/thinking about.

The model cache is now cleared whenever `load` is called or the `method` attribute is changed. This fixes #1622. There is a small "optimisation" that could be made (only clearing the cache if the data has changed) but I do not believe it is worth it. There's some discussion to be had over whether caching is useful in this model, at least when interpolation is not in use, but that really needs resolving some of the deeper issues discussed in 1620 and we should take these fixes now.

Following #1477 - which adds validation for the data classes - this PR includes checks for the arguments sent to the `load` method, including: convert to ndarray (which fixes some downstream bugs), ensure 1D, check the lengths match (if both set), and error out if only `x` is set. There are also related checks in `fold` and `calc` to ensure that data has been loaded. In #1477 the errors that were used were essentially already-existing cases (although I forget if I added any), and were `DataErr` exceptions. In this case I elected to raise `ModelErr` exceptions, with similar names to the `DataErr` case. There is an argument to be had that we should rethink the error classes (e.g. go back to things like `ValueError` where appropriate), but that is a completely separate issue and not one that would be easy to resolve.

The `__setstate__` method has been removed as it didn't seem to add anything. It looks like it was intended to support restoring "old" objects written with different attribute names, but the code was present when the git conversion happened, so the changes must have been very old. We have had a number of changes since then (in the Sherpa system) that have likely broken support, so there's no benefit in keeping this particular piece. It's also hard to understand how to map the restored values to the new scheme (e.g. ensuring data is in a ndarray).

The `method` attribute is now defined as such - as a class member - which we then change during `__init__`. This was done for documentation - so that RTD now mentions this attribute - and is a little tricky as if we'd just done `method = linear_interp` then the doctring we wrote would have been clobbered by that from `linear_interp`, which is why the need to set it in `__init__`.

## Examples

Here are some examples of the code behavior before and after this PR.

### Not a ndarray

```
>>> from sherpa.models.basic import TableModel  # main
>>> mdl = TableModel()
>>> mdl.load(None, [1, 2, 3])
>>> mdl([2, 3, 4])
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/model.py", line 652, in __call__
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/model.py", line 442, in cache_model
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/basic.py", line 1901, in calc
    return p[0] * self.__y
TypeError: can't multiply sequence by non-int of type 'numpy.float64'
```

This now works:

```
>>> from sherpa.models.basic import TableModel  # this PR
>>> mdl = TableModel()
>>> mdl.load(None, [1, 2, 3])
>>> mdl([2, 3, 4])
array([1., 2., 3.])
```

### Sanity check

```
>>> from sherpa.models.basic import TableModel  # main
>>> mdl = TableModel()
>>> mdl([1, 2, 3])
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/model.py", line 652, in __call__
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/model.py", line 442, in cache_model
  File "/home/dburke/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/models/basic.py", line 1904, in calc
    (len(self.__y), len(x0)))
TypeError: object of type 'NoneType' has no len()
```

vs

```
>>> from sherpa.models.basic import TableModel  # this PR
>>> mdl = TableModel()
>>> mdl([1, 2, 3])
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/models/model.py", line 652, in __call__
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/models/model.py", line 442, in cache_model
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/models/basic.py", line 2059, in calc
    raise ModelErr("The tablemodel's load method must be called first")
sherpa.utils.err.ModelErr: The tablemodel's load method must be called first
```

### cache

This is the original problem. The second `load` should invalidate the cache but it does not. Fortunately we can use non-ndarray values with the main branch

```
>>> from sherpa.models.basic import TableModel  # main
>>> mdl = TableModel()
>>> mdl.load([1, 2, 3], [10, 20, 30])
>>> mdl.ampl = 0.1
>>> mdl([1.5, 2.5])
array([1.5, 2.5])
>>> mdl.load([1, 2, 3], [4, 5, 6])
>>> mdl([1.5, 2.5])
array([1.5, 2.5])
```

vs

```
>>> from sherpa.models.basic import TableModel  # this PR
>>> mdl = TableModel()
>>> mdl.load([1, 2, 3], [10, 20, 30])
>>> mdl.ampl = 0.1
>>> mdl([1.5, 2.5])
array([1.5, 2.5])
>>> mdl.load([1, 2, 3], [4, 5, 6])
>>> mdl([1.5, 2.5])
array([0.45, 0.55])
```

There are other variants of this issue - e.g. when x is `None` and what happens if you change the interpolation method - which have also been fixed and tested.

### DIFF
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1902,13 +1902,6 @@ class TableModel(ArithmeticModel):
         self.ampl = Parameter(name, 'ampl', 1)
         ArithmeticModel.__init__(self, name, (self.ampl,))
 
-    def __setstate__(self, state):
-        self.__x = None
-        self.__y = state.pop('_y', None)
-        self.__filtered_y = state.pop('_filtered_y', None)
-        self.filename = state.pop('_file', None)
-        ArithmeticModel.__setstate__(self, state)
-
     def load(self, x, y):
         """Set the model values.
 
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -1783,22 +1783,25 @@ class Polynom2D(RegriddableModel2D):
 
 
 class TableModel(ArithmeticModel):
-    """Tabulated values in this model are simply linearly scaled.
+    """Tabulated values are linearly scaled and may be interpolated.
 
     After initializing this model, the independent and dependent
     arrays need to be loaded using the ``load`` method before the
     model can be used (see examples below); interpolation is used if
-    the grids of the data and the table do not match.
+    the grids of the data and the table do not match (this requires
+    that the independent axis is set rather than being left as
+    `None`). The interpolation scheme can be changed by setting the
+    `method` attribute.
 
-    When used with an integrated data set (for example,
-    `Data1DInt`) the independent axis loaded should be the left-edge
-    of the bin, and the dependent axis is the integrated value for that bin.
+    When used with an integrated data set (for example, `Data1DInt`)
+    the independent axis loaded should be the left-edge of the bin,
+    and the dependent axis is the integrated value for that bin.
 
     This model can also be used as to fit indexed data where the
     independent axis is not a continuous variable, e.g. the
     independent variable may hold the index for a number of stars and
     the dependent variable some measured property for each star. In
-    this case, the independent variable needs to be loaded as ``None``
+    this case, the independent variable needs to be loaded as `None`
     and the length of the dependent array needs to match the length of
     the data array exactly.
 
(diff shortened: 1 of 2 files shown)


sherpa/models/basic.py
```diff
@@ -1783,125 +1785,291 @@ def calc(self, *args, **kwargs):
...
+        (sherpa.utils.linear_interp).
+        """
+        return self._method
+
+    @method.setter
+    def method(self, val):
+        self._method = val
+
+        # as the method affects the cache, clear it (we could skip
+        # this if the method has not changed but is it worth it?)
```
hamogu, line 1926: I agree that it's not worth it.

sherpa/models/basic.py
```diff
@@ -1783,125 +1785,291 @@ def calc(self, *args, **kwargs):
...
+
+        # Clear the cache. We could avoid doing this if the
+        # data has not changed but this is not worth the
+        # complexity.
+        #
+        self.cache_clear()
+
+        # A simplified version of sherpa.data._check, which is not
+        # used here to avoid circular dependencies. Rather than raise
+        # DataErr, we raise ModelErr with a similar message.
```
hamogu, line 1969: Agreed with this approch (check, and ModelErr)

sherpa/models/basic.py
```diff
@@ -1783,125 +1785,291 @@ def calc(self, *args, **kwargs):
...
+
+        if x is None:
+            return
+
+        nx = len(self.__x)
+        ny = len(self.__y)
+        if nx != ny:
+            raise ModelErr(f"size mismatch between x and y: {nx} vs {ny}")
+
+        # Ensure the data is sorted. Is this useful?
```
hamogu, line 1995: Some interpolation methods my need that.

sherpa/models/basic.py
```diff
@@ -1783,125 +1785,291 @@ def calc(self, *args, **kwargs):
...
+        #
+        idx = self.__x.argsort()
+        self.__y = self.__y[idx]
+        self.__x = self.__x[idx]
 
     def get_x(self):
+        """Return the independent axis, if set"""
```
hamogu, line 2002: What does "if set" mean? It returns something (e.g. `None`) in any case, right? 
This is not a request to update the docstring, I just want to make sure I understand.

sherpa/models/basic.py
```diff
@@ -1783,125 +1785,291 @@ def calc(self, *args, **kwargs):
...
+                           f"data, ({len(self.__y)} vs {len(mask)})")
+
+        self.__filtered_y = self.__y[mask]
 
     @modelCacher1d
     def calc(self, p, x0, x1=None, *args, **kwargs):
+        """Evaluate the model.
+
+        The load method must have been called first. If both x and y
+        were given then the model is interpoalted onto the x0 grid,
```
hamogu, line 2053: interpoalted -> interpolated

sherpa/models/tests/test_tablemodel.py
```diff
@@ -0,0 +1,687 @@
...
+    """load: x and y (sorted)"""
+
+    mdl = TableModel()
+    mdl.load([12, 14, 17], [3, 2, 7])
+    assert mdl.get_x() == pytest.approx([12, 14, 17])
+    assert mdl.get_y() == pytest.approx([3, 2, 7])
+
+
+def test_load_x_y_unsorted():
+    """load: x and y (sorted)"""
```
hamogu, line 153: sorted -> unsorted

hamogu (changes requested):
Great PR. Extensive tests, clear purpose, easy to read. A few typos, I suspect @Dougburke only put in to check that I'm actually reading the code.

sherpa/models/basic.py
DougBurke: "if set" here means "if not None, otherwise returns None". 

I always struggle with how much to say or not to say here, and I don't think I am consistent. In this particular case I don't think it's worth clarifying (as it is adding documentation to a routine which didn't have documentation ;-)

DougBurke: For reference, "if set" means if `load` was called and `x` was not `None` in this call ;-)

DougBurke: I've changed ", if set" to " or None" as I was editing the file anyway.

hamogu (approved):

sherpa/models/basic.py
```diff
@@ -1783,125 +1785,311 @@ def calc(self, *args, **kwargs):
...
-        self.__x = x
+        """Set the model values.
+
+        Parameters
+        ----------
+        x, y : None or sequence
+           The model values. It is expected that either both are
+           given, and have the same number of elements, or that only y
+           is set, although the model can be cleared by setting both
+           to None. If x is given then the data is saved after bing
```
wmclaugh, line 1954:
```suggestion
           to None. If x is given then the data is saved after being
```

sherpa/models/tests/test_tablemodel.py
```diff
@@ -0,0 +1,687 @@
...
+    """
+
+    mdl = TableModel()
+    assert mdl.filename is None
+
+
+def test_ndim():
+    """What is the ndim?
+
+    At the moment we can use this with nD data but ony really if we do
```
wmclaugh, line 74:
```suggestion
    At the moment we can use this with nD data but only really if we do
```

PR #1623, opened 2022-11-02
sherpa_test: ensure pytest-xvfb is installed

DougBurke:
# Summary

When sherpa_test is called it will ensure that pytest-xdist is installed so that a DS9 window will not be created during the test. 

# Details

Ensure that the xvfb pytest plugin is installed before running the sherpa_test tests. This means that if a user has DS9 installed they will not see DS9 windows pop up during a test. If the user wishes to see the windows appear then they can add the

  --no-xvfb

argument to the sherpa_test call, or run the tests directly with

  pytest --pyargs sherpa

Note that there is some issue in the code for when we have to install pytest-xvfb as part of the test process: should we pass the new module through to pytest.main? The existing code suggests we should, as the module will not have been registered correctly. However, tests shows that the code fails if done like this, so we do not do this. The import code has not been used for a while so it's not clear if this is one or more of

- python version change
- pip change
- underlying build infrastructure change
- something odd with pytest-catchlog when it was available
- ?

### DIFF
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -843,12 +843,17 @@ def _install_test_deps():
             raise
 
     deps = ['pytest>=5.0,!=5.2.3']  # List of packages to be installed
-    pytest_plugins = []  # List of pytest plugins to be installed
+    pytest_plugins = ["pytest-xvfb"]  # List of pytest plugins to be installed
 
     # If the plugins are installed "now", pytest won't load them because they are not registered as python packages
     # by the time this code runs. So we need to keep track of the plugins and explicitly pass them to `pytest.main()`.
     # Note that explicitly passing plugins that were already registered would result in an error, hence this
     # complexity seems to be required.
+    #
+    # Is this still needed? A test with Python 3.10/pip 22.1.2
+    # and "pytest-xvfb" failed when "pytest_xvfb" was added to
+    # installed_plugins.
+    #
     installed_plugins = []
 
     for dep in deps:
@@ -863,7 +868,8 @@ def _install_test_deps():
             importlib.import_module(module)
         except ImportError:
             install(plugin_name)
-            installed_plugins.append(module)
+            # See comment above.
+            # installed_plugins.append(mod)
 
     return installed_plugins
 


DougBurke:
@harlantc @Marie-Terrell this is also a case that annoys me, but it has larger scope, as more people are likely to be calling `sherpa_test` with ds9 available than they are to hit #1618. It is also a clearer CM issue since we now force `pytest-xvdb` to be installed when `sherpa_test` is ran (even if ds9 is not installed). This process should be invisible to the user (that is, it's automatically handled, the user will see more screen output) but maybe I am overstepping the bounds of what we want `sherpa_test` to do.

This is only possible because `sherpa_test` was set up to allow `pytest` plugins to be installed, but the code has never been used (it was used to add a plugin but during development of the PR that plugin got included into `pytest` itself, so we have only ever really used it without requiring anything). The reason I mention this is that it appears that something has changed since the code was originally written, since I've found that if we call `pytest.main` with the `pytest_xvfb` argument sent to it's `lpugins` argument then it falls over horribly. So we end up not changing the "plugins" array, and things seem to work fine. Maybe the plugin system for pytest has changed, or maybe the PEP-related packaging changes have changed this behavior, or maybe it's Maybelline.

wmclaugh:
@DougBurke - Just to note, I have some apprehension on adding dependencies for something that we don't expect the standard user to encounter (ie. we don't provide ds9 for standalone sherpa so adding a dependency for pytest_xvfb doesn't seem like the correct solution to me).

hamogu:
I agree with @wmclaugh . It's already in `tests-require`: https://github.com/sherpa/sherpa/blob/main/setup.cfg#L32
So we should rather see what do do (document?) that those dependencies listed there are actually present when the tests are run. I suspect that there is some setup missing that allows to run the tests without it - but I just install `pytext-xvfb` by hand if ds9 annoys me and and I never invested enough time to figure how this is supposed to work and why it does not.

In contrast, the solution in this PR is very magic to install things within `__init__`. I strongly oppose to install things silently (without explicit user approval) in a function that's supposed to run tests. A user may have a good reasons not to have `pytest-xvfb`, e.g. it interacts badly in some way with another package they have in their environment. We can't just install stuff without asking - that's bad design and what viruses do, not respectable software. In fact,  I suggest to remove the code that could be used to install packages while running tests. We don't use it and I think it's bad practice.

DougBurke:
So, I disagree somewhat about the code to install packages in the sherpa_test code, since we also use it to install pytest if not available.

I also don't quite understand the comment about tests-require, since this is explicitly a "user level" set of tests and not part of the "hey, here's how python should run the tests as part of development". I don't believe that PyPA really has the concept of this sort of a test, so it's not surprising that there's no "official" way to do it. We *could* make `_install_test_deps` hook into the Sherpa configuration to find out the `tests_require` field, but I'm wary of doing that at this time (there's plenty of other work such as #1608 that should be looked at first), and I'm not sure it's actually sensible anyway.

Saying all that, I don't feel so strongly about this that I'll push for it - there are definitely concerns about requiring this particular package (e.g. X/xvfb).

PR #1626, opened 2022-11-07
Update sherpa.astro.utils.xspec for recent changes

DougBurke:
Will probably wait until CIAO 4.15 is released before I ask for this to be reviewed (as I need to check it works well with the CIAO 4.15 code, and that'll be a lot easier next year).

# Summary

Update the sherpa.astro.utils.xspec code to better support compiling XSPEC user models for use in Sherpa (fixes several bugs).

# Details

The code generated by the sherpa.astro.utils.xspec module had a few issues

a) outright errors, like a missing ';', a string that should have been interpolated that was not, and importing the warnings module when needed
b) updates needed because of recent changes to the XSPEC interface code
c) FORTRAN is case insensitive, so it looks like the symbols get converted to lowercase in the API which is an issue if a model name is mixed-case, as it is in at least one XSPEC user model; we avoid this now by making the names be lower case
d) ensure that C and C++ models are also declared (and the C++ models use the XSPEC-internal-ish cppModelWrapper routine to handle the conversion between C++ and C interfaces)

These have mostly been found (or taken from) the `convert_xspec_user_model` code (which is where this code came from in the first case).

I have also taken the time to add typing annotations and to move to use a dataclass rather than a namedtuple. As most of this code is internal to the module it should not cause an issue for users.

### DIFF
--- a/docs/overview/astro_utils_xspec.rst
+++ b/docs/overview/astro_utils_xspec.rst
@@ -13,3 +13,21 @@ The sherpa.astro.utils.xspec module
 
       create_xspec_code
       parse_xspec_model_description
+
+   .. rubric:: Classes
+
+   .. autosummary::
+      :toctree: api
+
+      XSPECcode
+      ModelDefinition
+      AddModelDefinition
+      MulModelDefinition
+      ConModelDefinition
+      MixModelDefinition
+      AcnModelDefinition
+      AmxModelDefinition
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased to pick up the python 3.11 changes

DougBurke:
Replaced by #1730

Issue #215, opened 2016-05-22
How to copy over model documentation via the ModelWrapper class?


DougBurke:
Users who use the UI layer - such as CIAO, but it can also be used in the standalone build - can not get a useful docstring for models.

# Example

## Using the model instance works

If a user calls `help` with a model _instance_, as shown below with the `help(m1)` call, then they get the documentation.

```
In [1]: from sherpa.astro.ui import *
WARNING: failed to import sherpa.astro.xspec; XSPEC models will not be available

In [2]: mdl = beta1d.m1 + const1d.b1

In [3]: help(m1)
Help on Beta1D in module sherpa.astro.models object:

class Beta1D(sherpa.models.model.ArithmeticModel)
 |  One-dimensional beta model function.
 |  
 |  The beta model is a Lorentz model with a varying power law.
 |  
 |  Attributes
 |  ----------
 |  r0
 |      The core radius.
 |  beta
 |      This parameter controls the slope of the profile at large
 |      radii.
 |  xpos
 |      The reference point of the profile. This is frozen by default.
 |  ampl
 |      The amplitude refers to the maximum value of the model, at
 |      x = xpos.
 |  
 |  See Also
 |  --------
 |  Beta2D, Lorentz1D, NormBeta1D
... etc ...
```
## Access via the modelwrapper does not

If they use the model wrapper instance - e.g. `beta1d` or `const1d` in this case - then they just get information on the `ModelWrapper` class:

```
In [4]: help(const1d)
Help on ModelWrapper in module sherpa.ui.utils object:

class ModelWrapper(sherpa.utils.NoNewAttributesAfterInit)
 |  Method resolution order:
 |      ModelWrapper
 |      sherpa.utils.NoNewAttributesAfterInit
 |      __builtin__.object
 |  
 |  Methods defined here:
 |  
 |  __call__(self, name)
 |  
... etc ...
```

This is not particularly informative. I imagine it's possible for the `ModelWrapper` class to copy over the docstring from the wrapped object (in the same way that decorators do), or perhaps there's some other approach that's better. This would then mean that `help(const1d)` would return the same thing as `help(sherpa.models.model.Const1D)` (just to be explicit with the module names).

# Is this actually what we want?

[...]

Issue #95, opened 2015-08-28
Use of mutable default arguments


olaurino:
There are places in the code (usually rather old code) where mutable objects are assigned to default arguments (see for instance https://github.com/sherpa/sherpa/pull/87#issuecomment-133427022). While this is a clear mistake and should be fixed, we are deferring this fix because:
  * The objects are not used read-only. Potential issues may be inadvertently introduced by users, but:
  * The impacted methods are not part of the UI, and they are called by other Sherpa calls, in most cases without relying on the default arguments.

In some cases the simple fix that has been applied to, e.g., PR #87, would introduce even less idiomatic python code, as the original source was not idiomatic to begin with. In these cases, a proper fix would go beyond just removing the mutable objects, and should deal with the issues in the original code.

As this does not seem to be a priority, we are going to re-evaluate this issue after the 4.8 CIAO release.

DougBurke:
One way to check is with `pylint`, as I noticed this particular error today (the line numbers won't make sense as this was with a PR I'm developing, but it's more to point out the error message):

```
% pylint sherpa/ui/utils.py sherpa/astro/ui/utils.py -d all -e W0102
************* Module sherpa.ui.utils
sherpa/ui/utils.py:177:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
sherpa/ui/utils.py:5254:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
sherpa/ui/utils.py:7056:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module sherpa.astro.ui.utils
sherpa/astro/ui/utils.py:10006:4: W0102: Dangerous default value {} as argument (dangerous-default-value)

 ```

and this warning does look like a good match to this issue: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html

**Comment by DougBurke** [comment]
The current problems I see are

```
% pylint sherpa/ -d all -e W0102
************* Module sherpa.instrument
sherpa/instrument.py:80:4: W0102: Dangerous default value [] as argument (dangerous-default-value)
sherpa/instrument.py:80:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
sherpa/instrument.py:249:4: W0102: Dangerous default value [] as argument (dangerous-default-value)
sherpa/instrument.py:249:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
[...]

PR #1638, opened 2022-11-29
Minor clean up of the utils code

DougBurke:
I have separated out the bug fixes from this into a number of other PRs. This is being left as there are some code clean ups to look at.

- #1662
  - #1664 
  - #1665 
  - #1666 
  - #1486 
    - #1637 
      - #1660 
        - #1667

# Summary

Minor code clean up of the utils code, adding pylint notes and fixing a number of issues: help(modelname) is more helpful for ui users #215; do not warn about unit mismatch twice with ignore #1641; fix sherpa.ui.save_source/model/resid for Data2D #1642, and fix an issue with show_bkg when some datasets do not have a background #1645.

# Details

Built on #1637. There is also some changes to address #95 but I do not want to close that just yet as I have not done a full check.

Changes are primarily driven by review of pylint output on the `sherpa/astro/ui/utils.py` and `sherpaui/utils.py` files, including

- the following changes lead to a bunch of uncovered code pathways, so most of them have had tests added (before any change was made) to ensure things still work (coverage is not 100% complete as some code pathways are very-tricky to invoke)
- adding a bunch of pylint disable statements to make it easier to see the problem cases; longer term we should think about what settings we want globally, but we get some benefit here and the change to global stuff should probably be done after we've done some more build updates and can rely on a "modern" system (e.g. setuptools > 60, modern settings of build options)
- replacing repeated code with a separate routine, including
  - the `save_*error` calls
  - access to PHA data which also had to worry about the `bkg_id` parameter; this enhances the internal `_get_pha_data` routine to accept an optional `bkg_id` argument, reducing a lot of boilerplate. There are some other places where we could do something similar but it's a bit harder to do (as it is a wrapper around `get_data` / `get_bkg` and it doesn't fit in quite so nicely, and there's not as many of these as there are of the pha case)
- #95 notes we have a number of routines that accept mutable arguments which can be problematic; this fixes a few of them (as pylint catches this case). Fortunately in these cases we don't rely on the mutable behavior **and** these are internal routines, so we can replace the mutable field with a `None` argument.
- some code was of the form `x = blah(); if y: x = blahblah()` which I've tried to replace with `if y: x = blahblah() else: x = blah()` as it avoids excess work when `y` holds. This is not really saving much time, but does clarify information flow
- f-string changes
- try to be more explicit about encoding on flle input and to use `with open` to handle file I/O (i.e. automatic closing)
- removed unused symbols and clarify imports (try to avoid both direct access to a symbol and the fully-qualified name)
- avoid bare `except` cases, but it is not a huge improvement as we go to catch `Exception` which is still too broad; this does at least avoid accidentally catching a control-c and there are some notes that at least one try/except ladder could be re-written (but not done here)
- removed unnecessary brackets and un-needed code (e.g. no need to check for a method returning `None` if that method throws an error rather than returning `None`).

The sherpa.ui.utils.ModelWrapper class has seen tests added, documentation added, and the behavior changed to address issue #215 - where Sherpa users saying `help(gauss1d)` did not find anything useful. We now get

```
>>> help(gauss1d)
Help on ModelWrapper in module sherpa.ui.utils:

<Gauss1D model type>
    Create a gauss1d model instance.
    
    One-dimensional gaussian function.

    Instances can be created either as an attribute of gauss1d,
    as long as the attribute does not begin with an underscore,
    or by calling gauss1d directly.
    
    Examples
    --------
    
    The model, here called mdl, is returned but it's also stored in
    the session and can be returned with `get_model_component`:
    
    >>> m1 = gauss1d.mdl
    
    If the model has already been created with the same name then
    the old version will be returned, rather than creating a new
    instance:
    
    >>> m1 = gauss1d.mdl
    >>> m2 = gauss1d("mdl")
    >>> m1 == m2
    True

>>> help(xscflux)
<XScflux model type>
    Create a xscflux model instance.
    
    The XSPEC cflux convolution model: calculate flux
    
    Instances can be created either as an attribute of xscflux,
    as long as the attribute does not begin with an underscore,
    or by calling xscflux directly.
    
    Examples
    --------
    
    The model, here called mdl, is returned but it's also stored in
    the session and can be returned with `get_model_component`:
    
    >>> m1 = xscflux.mdl
    
    If the model has already been created with the same name then
    the old version will be returned, rather than creating a new
    instance:
    
    >>> m1 = xscflux.mdl
    >>> m2 = xscflux("mdl")
    >>> m1 == m2
    True

```

which is not perfect, but a darn site better than the previous, which was

```
Help on ModelWrapper in module sherpa.ui.utils object:

class ModelWrapper(sherpa.utils.NoNewAttributesAfterInit)
 |  ModelWrapper(session, modeltype, args=(), kwargs={})
 |  
 |  Method resolution order:
 |      ModelWrapper
 |      sherpa.utils.NoNewAttributesAfterInit
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  __call__(self, name)
 |      Call self as a function.
 |  
 |  __getattr__(self, name)
 |  
 |  __init__(self, session, modeltype, args=(), kwargs={})
 |      Initialize self.  See help(type(self)) for accurate signature.
 |  
 |  __repr__(self)
 |      Return repr(self).
 |  
 |  __str__(self)
 |      Return str(self).
 |  
... continuing like this with no helpful information to the user
```


hamogu:
Not to derail this PR, but just noting that astropy is moving from flake8 or pylint to [ruff](https://github.com/charliermarsh/ruff) for performance reasons: https://github.com/sherpa/sherpa/pull/1638

DougBurke:
I'm marking this as ready for review, but it's not worth looking at until #1486 and then #1637 are closed.

DougBurke:
As basically all the PRs I span off out of this have been merged, with enough changes that make rebasing this not worth it to see if there's anything remaining, I'm closing it.

PR #1639, opened 2022-11-29
Ignore bad clarification

DougBurke:
It only really helps for grouped data; if the data is not grouped then resetting the quality array does not reset things **because** in this case we just piggy-back onto the mask system so you can't "remove" the bad-quality filter.

I also want to add in checks for the quality_filter array and think about what it means to ignore channels within a group.

**note to self** should we always apply the `quality_filter` setting (when set) to the `mask` field (technically get_mask) ? I was thinking we could xor it out when removing/changing it, but that doesn't work (as it's always removed). Perhaps we need an explicit mask field separate from the one changed by notice/ignore? However, still need to think about what happens when grouping - e.g. are the bad-quality channels just ignored [then make the widths harder to track] or do they always cause a group to start/end?

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1623,6 +1623,8 @@ class DataPHA(Data1D):
 
         self.bin_lo = None
         self.bin_hi = None
+
+        self.quality_filter = None
         self.quality = _check(quality)
         self.grouping = _check(grouping)
         self.exposure = exposure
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased as a bunch of related work has been merged.

DougBurke:
Rebased just to pick up recent changes (no conflict, I just wanted to ensure this all still worked together)

DougBurke:
Rebased.

PR #1644, opened 2022-12-05
Deployments via GH Actions

Marie-Terrell:
CM has been managing a mirror of this repository using GitLab that acted as the automated build/deployment system for the Conda packages that the end-users install (uploaded to the "dev" label for non-release builds and for builds before they are released). This mirror recently broke as GitLab no longer supports repo mirroring at the free tier (has not for a while, but we seemed to be grandfathered in until a recent repo move).

For now, we are exploring using GitHub Actions to perform the same action through this PR. We may want to back this change out at a later date and explore other options like webhooks.

For details on why the new workflow looks as it does, see [this project page](https://github.com/sherpa/sherpa/files/10154549/Sherpa_Conda_Deployment.pdf).


.github/workflows/ci-conda-deployment.yml
```diff
@@ -0,0 +1,287 @@
...
+        BUILD_NUM="$(git describe --tags --always | awk '{split($0,a,"-"); print a[2]}')"
+        echo "BUILD_NUM=${BUILD_NUM}"
+        if [ -z "${BUILD_NUM}" ] && [ "${GITHUB_REF:0:20}" != "refs/heads/releases/" ] ; then
+            #Skip the build/test/deploy if directly on a tag and NOT on a release branch
+            echo "skip_pipeline=true" >> $GITHUB_OUTPUT
+        else
+            echo "skip_pipeline=false" >> $GITHUB_OUTPUT
+        fi
+
+  #Linux and macOS build separate as a we can't use container option dynamically
```
wmclaugh, line 56:
```suggestion
  #Linux and macOS build separate as we can't use container option dynamically
```

harlantc (approved):
I think this works as a translation from the GitLab CI/CD system. We have some additional work to move to a more GitHub-focused paradigm with this workflow, which I'm tracking with CM-581 in Jira. Specifically, updating the macOS SDK, finding an alternative method for the `check-skip` step, and combining the build matrix. That'd be a bit more work however and can wait for the time being.

As is, I think this is good to go.

PR #1649, opened 2022-12-09
Add IXPE specific models and datastack

hamogu:
*This PR is not intended to be merged (at least not in the current form). It's purpose is to show a possible implementation of models that depend on more that in the counts array itself - in this case the polarization information kept in a header keyword.*

# Summary
Add a new class of models that combine multiple loaded spectra. 

# Details
In this particular example, that's IXPE spectra, where one vital piece of information (the Stokes parameter) is not part of the counts array, but is found in the PHA file header. This is also a case where the fit really need to look at multiple spectra at once - there is no point in fitting a Stokes U spectrum without having a corresponding spectrum for Stokes I and Q.

The basic idea is to have a class of models with a hidden, `alwaysfrozen` parameter that encodes the letter Stokes I/Q/U for each spectrum. The model then is a multiplicative model that multiplies any normal sherpa model (e.g. a powerlaw) to apply that information.

To make that more convenient for the user, much of the mechanics of how to set up those models are hidden away in a special sub-class of datastack.

A user would use this like this
```Python
ds = IXPEStack()
ds.load_filelist(phafiles)
ds.notice(2., 8.)
ds.set_source("polconst.pol * powlaw1d.power")
ds.fit()
ds.plot_fit()
```

(As submitted here, the code won't run. I've extracted this example from a notebook where I generated the data using the (still proprietory) IXPE simulator in the first few cells. However, we cna probably have a look at the way the models are written and decide if we want to peruse this further or not.)

@anetasie asked for this, also CC: @DougBurke

### DIFF
--- a/sherpa/astro/datastack/ixpestack.py
+++ b/sherpa/astro/datastack/ixpestack.py
@@ -1,0 +1,221 @@
+#
+# Copyright (C) 2022
+# MIT
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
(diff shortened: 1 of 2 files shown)


hamogu:
I want to mentions that this PR serves as a general proof-of-concept for setting up models across multiple spectra. 

I have several non-IXPE use cases for a similar mechanism (instead of "Stokes I, Q, U", the parameter that couples models could be "distance from cluster core" or "time since peak of flare"). In all those cases, we need some meta-information about the spectra (Stokes parameter, location, time) and the model would make use of that in some way. How exactly that would be used is slightly different, but I can imagine an API that's general enough to cover all those cases, while the code in this PR is specific to IXPE in particular.

DougBurke:
For background readers: I am interested in supporting XSPEC models that "need to know magic details about a dataset", and unfortunately they have a very-specialized interface which may or may not play nicely with this proposed interface.

PR #1650, opened 2022-12-12
Add wrappers around some XSPEC model for convenience

hamogu:
# Summary
Wrap some XSPEC themal (VAPEC) models for additional convenience to make it easier to have a temperature grid with tied redshift an abundance.

## Details
Literally every time that I fit a model in Sherpa to an X-ray spectrum I find myself typing some variant of 
```Python
apec2.Ne = apec1.Ne
apec2.O = apec1.O
apec2.Fe = apec1.Fe
```
Sure, I can be automated more with tricks like `for par1, par2 in zip(apec1.pars, apec2.pars:`
or 
```Python
for elem in ['Fe', 'Ne', 'O', 'Si']:
    getattr(apec1, elem).link = getattr(apec2, elem)
```
etc., but that means that every time I need to think about how exactly the pars are defined and how I iterate over them. 

I've looked at how how stellar astronomers do that i.e. those not as deeply involved in sherpa development as I am or not as comfortable with somewhat unusual Python commands like `getattr` and I see that they type out (or copy and paste) `apec2.Ne = apec1.Ne` for every element and every APEC component.

Thus, I propose that a model like the one in this PR that wraps that process into a new Sherpa model can simplify our fitting considerably. It's so useful, because the stellar community essentially only uses a single type of XSEPC model (APEC/VAPEC) for all sources. This is true for CCD level spectra, but even more important for grating spectra, where one typically has the information to fit more temperature components and more individual abundances.

## Feedack request
Since this would the be first "conveneience wrapping" of XSPEC models, I'd like an early thumbs yp/down from @DougBurke and @anetasie before I bother with writing tests etc. I think the idea behind this PR is realtively simple and can be seen from just the code and a screenshot I'm going to post shortly.

## To-Do

- [ ] Docs
- [ ] equivalent classes for APEC, VVAPEC
- [ ] tests

### DIFF
--- a/sherpa/astro/xspec/convenience.py
+++ b/sherpa/astro/xspec/convenience.py
@@ -1,0 +1,103 @@
+#
+#  Copyright (C) 2022
+#  MIT
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+try:
+    from . import _xspec
+except ImportError as ie:
+    # It would be nicer for the user to say "from None" rather than
+    # "from ie" here but it may lose important information (in the
+    # case there is some problem initializing the XSPEC library,
+    # rather than it just not being built).
+    #
+    raise ImportError("XSPEC support is not enabled") from ie
+
+import numpy as np
+from sherpa.models import model
+
+__all__ = ['MultiTvapec', 'VAPECgrid']
+
+
+class MultiTvapec(model.RegriddableModel1D):
+    def __init__(self, name='multiTvapec', n_components=1):
+        '''
+        Alternative implementation idea: 
+            Make many VAPEC models, link parameters and keep a list of them in here.
+            That way, each temperature component would have a dedicated VAPEC instance. 
+            If there is some internal cashing, that might speed up computations - but not that much,
+            since the abundances are already matched.
+        '''
+        # Why do I need this line? Should be done at the end in ArithmeticModel.__init__?
+        self.name = name
+        
+        self.n_components = n_components
+        self.vapec = xspec.XSvapec(name=f'xs')
+        pars = []
+        for i in range(n_components):
+
+            p = model.Parameter(name, f'kT_{i + 1}',
+                                 # +1 because can't set to 0 for i=0
+                                val=min(i + 1, self.vapec.kT.hard_max),
+                                min=self.vapec.kT.min,
+                                hard_min=self.vapec.kT.hard_min,
+                                max=self.vapec.kT.max,
+                                hard_max=self.vapec.kT.hard_max)
+            setattr(self, f'kT_{i + 1}', p)
+            pars.append(p)
+        for i in range(n_components):
+            p = model.Parameter(name, f'norm_{i + 1}',
+                                val=self.vapec.norm.val,
+                                min=self.vapec.norm.min,
+                                hard_min=self.vapec.norm.hard_min,
+                                max=self.vapec.norm.max,
+                                hard_max=self.vapec.norm.hard_max)
+            setattr(self, f'norm_{i + 1}', p)
+            pars.append(p)            
+            
+        for j in range(1, len(self.vapec.pars) - 1):
+            par = self.vapec.pars[j]
+            p = model.Parameter(name, par.name, val=par.val,
+                    min=par.min, hard_min=par.hard_min,
+                    max=par.max, hard_max=par.hard_max,
+                               frozen=True)
+            setattr(self, par.name, p)
+            pars.append(p)
+        model.ArithmeticModel.__init__(self, name, tuple(pars))
+        
+    def calc(self, pars, *args, **kwargs):
+        components = []
+        # Can I replace the following loop with sherpa.utils.parallel_map?
+        # Or do I have to use different vapec objects to avoid some race condition?
+        # Is this even the time critical step? Should check that...
+        for i in range(self.n_components):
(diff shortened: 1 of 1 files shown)


hamogu:
This is how it would look"
Screenshot 2022-12-12 at 7 11 06 AM

DougBurke:
I don't have time to look at the code, where I'm sure it's obvious, and I can make a guess from the screenshot, but is the idea that the model lets you "combine" multiple apec models together, with the elemental abundances automatically shared? If so is it limited to just vapec or can we easily apply it to the other models (there's things like bvvtapec, whive I've never used in anger myself...)

hamogu:
Currently it's written fur just a VSPEC, not even APEC or VVAPEC, but there is nothing that stops me from generalizing this to "the sum n copies of any additive model". In the case of VAPEC, I know that you want to fit temperature and norm separately in 99.9% of the cases (otherwise, what's the point of having multiple components?), while in general, there is nothing in the models description that tells me which of the model parameters I want to be linked and which I want to be unlinked. 

However, if we are interested in the more general way, we can discuss the interface for that. It's easy in the OO case, where that just becomes another parameter for `__init__(... ,basemodel, n_copies, parname_linked=['a', 'b', ...], parnames_not_linked=['c', ])` where we can have some sensible default (e.g. don't link anything called "norm", link everythng unless explicitly listed etc.). It's much harder in the UI with our domain-specific language `generalmodelname.myname` where we can't give parameters to `generalmodelname`. That would require some wrapper function (e.g. `ui.make_and_link_n_copies(basemodelname, parname_linked=['a', 'b', ...])`) to generate the model object which can then be passed around. In that usual UI paradigm

I'm doing
```Python
labsmodel = xspec.XSTBabs(name='labs')
ltempmodel = MultiTvapec(name=f'ltemp', n_components=4)
ui.set_source('late', labsmodel * ltempmodel)
```
but we would have to discuss if that's the way we want it to be done.

DougBurke:
So, I can see the general appeal of a "I want n copies of this model with some linkage of paramters" - e.g. maybe three `NormGauss1D` models where I want the FWHM to be linked. So there could be some benefit to having something generic. However, as @hamogu mentions, how do we know what should be linked? Again, we could come up with various schemes for the API. The question is is it worth it? i'm just thinking about where the functionality should live - in some ways it is orthogonal to this, as we could change this code to use any generic system we come up with, but as we have time for noodling we can noodle.

noodle = think, in this particular case...

hamogu:
Yes, in principle, this could also be used for n Gaussian emission lines. There is a qualitative difference though - if I model emission lines independently, I want to link, maybe, one parameter (width) per model, while with VAPEC or VVAPEC I want to link 15-30 parameters per model. That's a lot more work to do manually. So, we could either do a very generic way of doing that or just hand-code it on a case-by-case bases for those models where we believe it's most likely to come up.

I can see benefits of either approach. It's obvious why one may want a generic way to do that, but I want to argue that there would also be a benefit in the latter: In that case, we could implement the model similar to this PR, maybe just for the XXXAPEC models, and it would be listed as models with a vaguely descriptive name in `list_models` and friends, similar to XXXAPEC. There is some precendent for that, too: XSPEC has the c6??mekl models, which are mekal (an earlier version of what today is apec) models with  polynamial representation of the temperature distribution.

DougBurke:
So, we could think about extending the link command to allow multiple models to be set - e.g something like

    link([mdl2.N, mdl3, N, mdl4.N], mdl1.N)

but that would not remove the utility from something like this.

hamogu:
@anetasie While this is a draft PR, I'd like you feedback on this. We could:
- go with this version, specific to APEC with just minor clean-ups needed (docs, tests, generalize to APEC, VAPEC, VVAPEC)
- implement a more generic way of linking many parameters for many models
- discard this entirely

Whatever road we want to take, we need to decide what road to take and @anetasie input would be very useful here.

anetasie:
@hamogu It is a good idea to make this interface simpler and perhaps we should go with APEC example wrapper. 
I guess in order to finalize the PR the docs and tests are need. Also, I see that this is new and separate code, which should not impact the other parts of Sherpa. I suggest option 1 at this moment, so we can add this to the release.

hamogu:
I think @DougBurke is worried that this is not general enough and we end up with a bunch or really moritz-specific models. I don't tihnk that's the case here (APEC is widely used in this way in the stellar community), but it might be worth a short conversation if we can find an API that is both simple and general and supports this use case.

anetasie:
We did not discuss this PR last week. This is a case for specific models. @DougBurke do we really want to worry about the general case here?

hamogu:
> My concerns include

>    who will maintain this

presumably me...

anetasie:
I guess we need at least one test for this.

hamogu:
I agree, we do need a test or more. That's why it's labelled "draft". The question is if we take it in this form, make is a little more general to at least capture the other apec-type XSPEC models (apec, vapec, vvapec, ...), or make follow one of @DougBurke 's ideas for a more general linking, for example, we could have a function `link_all([model1, model2, model3], exclude = ['norm', 'kT'])` which would link all parameters of model1, model2, and model3 that have identical names expect those listed in the exclude parameter.

Such a function would be much more general, and my code example above would become:
```Python
labsmodel = xspec.XSTBabs(name='labs')
apecmodels = [xspec.XSvapec(name=f'apec{i}') for i in range(4)]
link_all(apecmodels, exclude = ['norm', 'kT'])`
ui.set_source('late', labsmodel * sum(apecmodels))
```

Both [me](https://github.com/sherpa/sherpa/pull/1650#issuecomment-1346695516) and [Doug](https://github.com/sherpa/sherpa/pull/1650#issuecomment-1499493075) made alternative suggestions above, but we have to decide what we want.

While I opened this PR specifically for APEC, I now think a more general solution would be better, something like @DougBurke 's `link([mdl2.N, mdl3, N, mdl4.N], mdl1.N)` but with added capabilities like the example above: `link_all(apecmodels, exclude = ['norm', 'kT'])`
We could "overload" link to do that or define a new function that is a superset.

anetasie:
I agree that a more general solution would be bette, so maybe this is still a draft and not ready for this release.

hamogu:
There are then some usability questions: Do we want to allow linking only over models of the same type? (XSapec to XSApec) or also over multiple types (different parameters, link only parameters that the first model has? That all models share?), e.g. I could image a situation where I want to link the redshift of all models, even though one is an apec and the other something else. However, that might open too many corner cases for the function I envision... if I get my plot stuff sorted out fast enough, I might open a PR for a linking functions to replace this one.

anetasie:
@hamogu Sounds good!

hamogu:
Since we decided not to take this code as is, I've opened an issue about it (#1914) for tracking and I'm closing this PR.

Issue #1027, opened 2020-12-10
How to include functions in parameter links


DougBurke:
I have sometimes wanted to say something like

```
foo.x = 4000 + 10 * np.sin(angle.c0)
```

where `angle.c0` is a parameter, but it fails (at creation time) with an error like:

```
sherpa In [7]: src.xpos = 4000 + 10 * np.cos(const1d.angle)
TypeError: loop of ufunc does not support argument 0 of type Parameter which has no callable cos method
```

or, for non-NumPy ufuncs:

```
sherpa In [24]: src.xpos = 4000 + 10 * math.cos(const1d.angle)
TypeError: must be real number, not Const1D
```

We need to lift the function into the parameter language. Here is a solution which needs testing, and then deciding how to handle multiple arguments (possible to make this support multiple arguments): 

```
class FunctionParameter(sherpa.models.parameter.CompositeParameter):
    """Store a function of a single argument.

    We need to only evalate the function when required.
    """

    @staticmethod
    def wrapobj(obj):
        if isinstance(obj, sherpa.models.parameter.Parameter):
            return obj
        return sherpa.models.parameter.ConstantParameter(obj)

    def __init__(self, arg, func):

        if not callable(func):
            raise ValueError("func argument is not callable!")

        self.arg = self.wrapobj(arg)
        self.func = func

        # Would like to be able to add the correct import symbol here
        lbl = _mklabel(func)
        lbl += f'{func.__name__}({self.arg.fullname})'
        sherpa.models.parameter.CompositeParameter.__init__(self, lbl,
                                                            (self.arg,))

    def eval(self):
        return self.func(self.arg.val)


def _mklabel(func):
    """Create a 'nice' symbol for the function.

    Is this worth it over just using __name__?

    Parameters
    ----------
    func : function reference

    Returns
    -------
    label : str
        A representation of the function.
    """

    if isinstance(func, np.ufunc):
        return 'np.'

    try:
        lbl = f'{func.__module__}.'
    except AttributeError:
        return ''

    if lbl == 'builtins.':
        return ''

    return lbl
```

With this, we can say (`xcm` is just the module where I placed the code):

```
sherpa In [16]: src.xpos = 4000 + 10 * xcm.FunctionParameter(angle.c0, np.cos)

sherpa In [17]: print(src)
gauss2d.src
   Param        Type          Value          Min          Max      Units
   -----        ----          -----          ---          ---      -----
[...]

PR #1653, opened 2022-12-14
Allow arbitrary functions for linking parameters

### LINKED ISSUE #1652: Link model parameters with mathematical functions
Model parameters can be linked directly to other model parameters or if they are linearly related, e.g. `g1.fwhm = 2 * g2.fwhm`, have a constant offset, or e.g. `g1.fwhm = 1 + 2 * g2.fwhm**0.5`. However, linking them with more complex expressions, e.g. exp, log, sin, cos etc. fails. The failure is different depending on if I loaded the mathematical function from `numpy` or the build-in `math` module.
```Python
from sherpa.astro import ui
import numpy as np
from math import exp
ui.load_arrays(1, [1,2,3,4,5], [1,2,3,2,1])
ui.set_source(1, ui.gauss1d.g1)

g1.fwhm = np.exp(g1.ampl)
AttributeError: 'Parameter' object has no attribute 'exp'

Cell In [4], line 1
----> 1 g1.fwhm = np.exp(g1.ampl)

TypeError: loop of ufunc does not support argument 0 of type Parameter which has no callable exp method
```

or, alternatively
```Python
g1.fwhm = exp(g1.ampl)
Cell In [5], line 1
----> 1 g1.fwhm = exp(g1.ampl)

TypeError: must be real number, not Parameter
```

hamogu:
# Summary
Provide a way to use any Python function to be used for linking parameters. Fix #1652
    
# Details
Before this PR, only a few selected mathematical operations where allowed to link parameters, namely those expressed with symbols in Python (+, -, *, **, //, etc.). Internally, those symbols are mapped to numpy ufuncs in `UnaryOpParameter` or `BinaryOpParameter`. Since about numpy 1.13 there is a defined interface to wrap through any ufunc, which gives us access to sin, cos, exp, log etc.

See https://github.com/sherpa/sherpa/issues/1652#issuecomment-1349865957 for an example. There, I want to model how a flare cools down and have many spectra distributed in time. Thus, it makes sense to describe the normalization of the model with an exponential function, essentially `apec_n.norm = apec_0.norm * np.exp(-(t_n-t_0)/decay_time)` where apec_n is the model for the nth spectrum observed at time t_n. `decay_time` is a parameter that I want to fit. Being able to use functions like exp, log, cos, sin can be useful as you can see in that example.
In that particular case, I discovered that `2.71**(-(t_n-t_0) ... ` already works, but we give users much more flexibility if we allow any function to be used.

Code changes are small, because the infrastructure is already there, we just added a bit more flexibility. In fact, we added infinite flexibility, because numpy provides a wrapper to turn any arbitrary Python function into a ufunc - and thus any arbitrary Python function can now be used to link parameters.
    
With such power comes great responsibility. It is easily possible to make links that make no sense or will trip over later.
However, I argue that that's a feature not a bug. Nonsensical links that error out later have always been possible, e.g. with Sherpa 4.14:
```Python
from sherpa.models import Gauss1D
import numpy as np
    
g1 = Gauss1D(name='g1')
g1.fwhm = g1.ampl * np.arange(5)
```
which will fail as soon as the expression is evaluated:
```Python
print(g1.fwhm)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
    
This has never been reported as a problem, likely because it's unlikely to happen accidentally and can be spotted quite quickly. I chose to not add any additional checking to prevent things like that, in order to leave the maximum flexibility for any user models.

### DIFF
--- a/docs/models/index.rst
+++ b/docs/models/index.rst
@@ -438,14 +438,14 @@ can be used in the linking expression, for example::
 
 This includes many commonly used mathematical and trigonometric functions
 such as log, exp, sin, cos, which allows building quite complex parameter
-linkage. Only the numpy versions work here, but not the functions from the
+linkage. Only the numpy versions work here, **not** the functions from the
 build-in ``math`` module, so use `numpy.exp` instead of `math.exp`.
 Many more complex functions are available in
 `scipy.special <https://docs.scipy.org/doc/scipy/reference/special.html>`_;
 any arbitrary Python function can be turned into a ufunc with
 `numpy.frompyfunc <https://numpy.org/doc/stable/reference/generated/numpy.frompyfunc.html#numpy.frompyfunc>`_
 and the interface is also available for
-`C extensions <https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html#creating-a-new-universal-function>`.
+`C extensions <https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html#creating-a-new-universal-function>`_.
 However, if such complex expressions are required to link model parameters
 together, it might be better to write a
 :ref:`dedicated user model <usermodel>` that describes the data with the
@@ -460,7 +460,7 @@ turning values of parameters into arrays (Sherpa optimisers can only deal
 with scalar parameters.) In practice, such mistakes
 are easy to spot when displaying a model; because Sherpa is meant to be
 a general and flexible modelling application that works with (almost)
-arbitrary user-defined models, the code puts as few a-priory restrictions
+arbitrary user-defined models, the code puts as few restrictions
 as possible on the functions used for linking parameters.
 
 .. _parameter_reset:
--- a/docs/models/index.rst
+++ b/docs/models/index.rst
@@ -34,7 +34,7 @@ be set. In this case a one-dimensional gaussian using the
 A description of the model is provided by ``help(g)``.
 
 The parameter values have a current value, a valid range
-(as given by the the minimum and maximum columns in the table above),
+(as given by the minimum and maximum columns in the table above),
 and a units field. The units field is a string, describing the
 expected units for the parameter; there is currently *no support* for
 using `astropy.units
@@ -125,7 +125,7 @@ Changing a parameter
 
 The parameters of a model - those numeric variables that control the
 shape of the model, and that can be varied during a fit -
-can be accesed as attributes, both to read or change
+can be accessed as attributes, both to read or change
 the current settings. The
 :py:attr:`~sherpa.models.parameter.Parameter.val` attribute
 contains the current value::
@@ -220,7 +220,7 @@ values as the hard limits.
 Setting a parameter to a value outside its soft limits will
 raise a :py:exc:`~sherpa.utils.err.ParameterErr` exception.
 
-During a fit the paramater values are bound by the soft limits,
+During a fit the parameter values are bound by the soft limits,
 and a screen message will be displayed if an attempt to move
 outside this range was made. During error analysis the parameter
 values are allowed outside the soft limits, as long as they remain
@@ -233,7 +233,7 @@ Guessing a parameter's value from the data
 
 Sherpa models have a
 :py:meth:`~sherpa.models.model.Model.guess`
-method which is used to seed the paramters (or
+method which is used to seed the parameters (or
 parameter) with values and
 :ref:`soft-limit ranges <params-limits>`
 which match the data.
@@ -334,10 +334,10 @@ related to another: this can be equality, such as saying that
 the width of two model components are the same, or a functional
 form, such as saying that the position of one component is a
 certain distance away from another component. This concept
-is refererred to as linking parameter values. The second case
-incudes the first - where the functional relationship is equality -
+is referred to as linking parameter values. The second case
+includes the first - where the functional relationship is equality -
 but it is treated separately here as it is a common operation.
-Lnking parameters also reduces the number of free parameters in a fit.
+Linking parameters also reduces the number of free parameters in a fit.
 
 The following examples use the same two model components::
 
@@ -426,6 +426,42 @@ directly, it should be multiplied by zero. So, for this example
 the model to be fit would be given by an expression like::
 
    >>> mdl = g1 + g2 + 0 * sep
+
+Complex functional relationships
+--------------------------------
+
+Any `numpy universal function ("ufunc") <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs>`_
+can be used in the linking expression, for example::
+
+    >>> import numpy as np
+    >>> g2.ampl = np.cos(g1.ampl)
+
+This includes many commonly used mathematical and trigonometric functions
+such as log, exp, sin, cos, which allows building quite complex parameter
+linkage. Only the numpy versions work here, but not the functions from the
+build-in ``math`` module, so use `numpy.exp` instead of `math.exp`.
+Many more complex functions are available in
+`scipy.special <https://docs.scipy.org/doc/scipy/reference/special.html>`_;
+any arbitrary Python function can be turned into a ufunc with
+`numpy.frompyfunc <https://numpy.org/doc/stable/reference/generated/numpy.frompyfunc.html#numpy.frompyfunc>`_
+and the interface is also available for
+`C extensions <https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html#creating-a-new-universal-function>`.
+However, if such complex expressions are required to link model parameters
+together, it might be better to write a
+:ref:`dedicated user model <usermodel>` that describes the data with the
+appropriate parameters in the first place.
+
+Not every possible link function makes sense
+--------------------------------------------
+
+With this flexibility, it is possible to define links that make no sense,
+for example taking the logical not of a parameter that represents a mass or
+turning values of parameters into arrays (Sherpa optimisers can only deal
+with scalar parameters.) In practice, such mistakes
+are easy to spot when displaying a model; because Sherpa is meant to be
+a general and flexible modelling application that works with (almost)
+arbitrary user-defined models, the code puts as few a-priory restrictions
+as possible on the functions used for linking parameters.
 
 .. _parameter_reset:
 
--- a/sherpa/models/parameter.py
+++ b/sherpa/models/parameter.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2017, 2020, 2021, 2022
+#  Copyright (C) 2007, 2017, 2020, 2021, 2022, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -647,10 +647,20 @@ class Parameter(NoNewAttributesAfterInit):
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         if not method == '__call__':
             return NotImplemented
+        if hasattr(numpy, ufunc.__name__):
+            name = f"numpy.{ufunc.__name__}"
+        else:
+            # Unfortunately, there is no ufunc.__module__ we could use
+            # This could be a ufinc from e.g. scipy or generated from an
+            # arbitrary Python function with numpy.frompyfunc.
+            # In the latter case, the name will be something like
+            # "func (vectorized)", which looks confusing in out string
+            # represenantion, so we simplify it.
+            name = ufunc.__name__.replace(' (vectorized)', '')
         if ufunc.nin == 1:
-            return UnaryOpParameter(inputs[0], ufunc, ufunc.__name__)
+            return UnaryOpParameter(inputs[0], ufunc, name)
         if ufunc.nin == 2:
-            return BinaryOpParameter(inputs[0], inputs[1], ufunc, ufunc.__name__,
+            return BinaryOpParameter(inputs[0], inputs[1], ufunc, name,
                                      strformat='{opstr}({lhs}, {rhs})')
         return NotImplemented
 
--- a/sherpa/models/parameter.py
+++ b/sherpa/models/parameter.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2017, 2020, 2021
+#  Copyright (C) 2007, 2017, 2020, 2021, 2022
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -257,18 +257,18 @@ def _make_set_limit(name):
     return _set_limit
 
 
-def _make_unop(op, opstr):
+def _make_unop(op, opstr, **kwargs):
     def func(self):
-        return UnaryOpParameter(self, op, opstr)
+        return UnaryOpParameter(self, op, opstr, **kwargs)
     return func
 
 
-def _make_binop(op, opstr):
+def _make_binop(op, opstr, **kwargs):
     def func(self, rhs):
-        return BinaryOpParameter(self, rhs, op, opstr)
+        return BinaryOpParameter(self, rhs, op, opstr, **kwargs)
 
     def rfunc(self, lhs):
-        return BinaryOpParameter(lhs, self, op, opstr)
+        return BinaryOpParameter(lhs, self, op, opstr, **kwargs)
 
     return (func, rfunc)
 
@@ -367,7 +367,7 @@ class Parameter(NoNewAttributesAfterInit):
     val = property(_get_val, _set_val,
                    doc='The current value of the parameter.\n\n' +
                    'If the parameter is a link then it is possible that accessing\n' +
-                   'the value will raise a ParamaterErr in cases where the link\n' +
+                   'the value will raise a ParameterErr in cases where the link\n' +
                    'expression falls outside the soft limits of the parameter.\n\n' +
                    'See Also\n' +
                    '--------\n' +
@@ -631,7 +631,7 @@ class Parameter(NoNewAttributesAfterInit):
         return html_parameter(self)
 
     # Unary operations
-    __neg__ = _make_unop(numpy.negative, '-')
+    __neg__ = _make_unop(numpy.negative, '-', strformat='-{arg}')
     __abs__ = _make_unop(numpy.absolute, 'abs')
 
     # Binary operations
@@ -643,6 +643,17 @@ class Parameter(NoNewAttributesAfterInit):
     __truediv__, __rtruediv__ = _make_binop(numpy.true_divide, '/')
     __mod__, __rmod__ = _make_binop(numpy.remainder, '%')
     __pow__, __rpow__ = _make_binop(numpy.power, '**')
+
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if not method == '__call__':
+            return NotImplemented
+        if ufunc.nin == 1:
+            return UnaryOpParameter(inputs[0], ufunc, ufunc.__name__)
+        elif ufunc.nin == 2:
+            return BinaryOpParameter(inputs[0], inputs[1], ufunc, ufunc.__name__,
+                                     strformat='{opstr}({lhs}, {rhs})')
+        else:
+            return NotImplemented
 
     def freeze(self):
         """Set the `frozen` attribute for the parameter.
@@ -815,17 +826,22 @@ class UnaryOpParameter(CompositeParameter):
         The ufunc to apply to the parameter value.
     opstr : str
         The symbol used to represent the operator.
+    strformat : str
+        Format string for printing this operation. Elements
+        than can be used are `arg` and `opstr`, e.g.
+        `strformat='{opstr}({arg})'`.
 
     See Also
     --------
     BinaryOpParameter
     """
 
-    def __init__(self, arg, op, opstr):
+    def __init__(self, arg, op, opstr, strformat='{opstr}({arg})'):
         self.arg = arg
         self.op = op
         CompositeParameter.__init__(self,
-                                    '%s(%s)' % (opstr, self.arg.fullname),
+                                    strformat.format(opstr=opstr,
+                                                     arg=self.arg.fullname),
                                     (self.arg,))
 
     def eval(self):
@@ -845,6 +861,10 @@ class BinaryOpParameter(CompositeParameter):
         The ufunc to apply to the two parameter values.
     opstr : str
         The symbol used to represent the operator.
+    strformat : str
+        Format string for printing this operation. Elements
+        than can be used are `lhs`, `rhs`, and `opstr`, e.g.
+        `strformat='{opstr}({lhs}, {rhs})'`.
 
     See Also
     --------
@@ -857,13 +877,16 @@ class BinaryOpParameter(CompositeParameter):
             return obj
         return ConstantParameter(obj)
 
-    def __init__(self, lhs, rhs, op, opstr):
+    def __init__(self, lhs, rhs, op, opstr,
+                 strformat='({lhs} {opstr} {rhs})'):
         self.lhs = self.wrapobj(lhs)
         self.rhs = self.wrapobj(rhs)
         self.op = op
-        CompositeParameter.__init__(self, '(%s %s %s)' %
-                                    (self.lhs.fullname, opstr,
-                                     self.rhs.fullname), (self.lhs, self.rhs))
+        CompositeParameter.__init__(self,
+                                    strformat.format(lhs=self.lhs.fullname,
+                                                     rhs=self.rhs.fullname,
+                                                     opstr=opstr),
+                                    (self.lhs, self.rhs))
 
     def eval(self):
         return self.op(self.lhs.val, self.rhs.val)
--- a/sherpa/models/parameter.py
+++ b/sherpa/models/parameter.py
@@ -649,11 +649,10 @@ class Parameter(NoNewAttributesAfterInit):
             return NotImplemented
         if ufunc.nin == 1:
             return UnaryOpParameter(inputs[0], ufunc, ufunc.__name__)
-        elif ufunc.nin == 2:
+        if ufunc.nin == 2:
             return BinaryOpParameter(inputs[0], inputs[1], ufunc, ufunc.__name__,
                                      strformat='{opstr}({lhs}, {rhs})')
-        else:
-            return NotImplemented
+        return NotImplemented
 
     def freeze(self):
         """Set the `frozen` attribute for the parameter.
--- a/sherpa/models/parameter.py
+++ b/sherpa/models/parameter.py
@@ -651,11 +651,11 @@ class Parameter(NoNewAttributesAfterInit):
             name = f"numpy.{ufunc.__name__}"
         else:
             # Unfortunately, there is no ufunc.__module__ we could use
-            # This could be a ufinc from e.g. scipy or generated from an
+            # This could be a ufunc from e.g. scipy or generated from an
             # arbitrary Python function with numpy.frompyfunc.
             # In the latter case, the name will be something like
             # "func (vectorized)", which looks confusing in out string
-            # represenantion, so we simplify it.
+            # representation, so we simplify it.
             name = ufunc.__name__.replace(' (vectorized)', '')
         if ufunc.nin == 1:
             return UnaryOpParameter(inputs[0], ufunc, name)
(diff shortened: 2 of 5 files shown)


sherpa/models/tests/test_parameter.py
```diff
@@ -170,24 +170,51 @@ def setUp_composite():
...
+
+    This implicitly also checks the logic to define a the format string for
+    printing the model representation.
+    '''
+    p, p2 = setUp_composite()
+    comp = p + p2
+    assert repr(comp) == "<BinaryOpParameter '(model.p + model.p2)'>"
+    comp = np.logaddexp2(p, p2)
+    assert repr(comp) == "<BinaryOpParameter 'logaddexp2(model.p, model.p2)'>"
+
```
DougBurke, line 217: I think it would be useful to have a check for the string output of something like `1 + np.e**p.gamma`, in part to convince me that the opstr changes is working as expected (i.e. it's not affecting this one). It would be nice to have the first commit be tests, then the second one changes + new-tests-that-are-now-possible (or removing an xfail). Then it makes it easier to have more confidence in what has/hasn't changed. But it depends on how many of these new tests actually work without the corresponding changes.

hamogu: I re-ordered the commits and added a new test (to be run before the code changes go in), i.e. it works in 4.14.

hamogu: Need to go back and fix up the first commit...

hamogu: OK, I think I separated it now such that the tests work commit-by-commit.

DougBurke:
```
>>> m2 = Const1D()
>>> 2 * np.power(np.e, m2.c0)
<BinaryOpParameter '(2 * power(2.718281828459045, const1d.c0))'>
>>> (2 * np.power(np.e, m2.c0)).val
5.43656365691809
>>> m2.c0 = 2
>>> (2 * np.power(np.e, m2.c0)).val
14.778112197861299
```

So this looks good. And the test I asked about earlier is essentially

```
>>> 2 * (np.e ** m2.c0)
<BinaryOpParameter '(2 * (2.718281828459045 ** const1d.c0))'>
>>> (2 * (np.e ** m2.c0)).val
14.778112197861299
>>> m2.c0 = 1
>>> (2 * (np.e ** m2.c0)).val
5.43656365691809
```

I would need to check, but this looks like it hasn't changed.

DougBurke:
I'm going to think a little about models as part of this (ie should we do the same thing with models), so review may be a little slow.

hamogu:
I think we should do the same for model, but I anticipate that that might need a more discussion to convince everyone.  Thus, I would prefer to make that a separate PR (I have a local draft, but not pushed yet).

DougBurke:
So, with the following code (ran in `sherpa-test-data/sherpatest/`):

```python
import numpy as np
from sherpa.astro import ui

# don't use the reg file as do not have support for this outside of a CIAO build
ui.set_stat("cash")
ui.set_method("simplex")

ui.load_image(1, "acisf08478_000N001_r0043_regevt3_srcimg.fits")
ui.set_coord(1, "physical")
ui.notice2d_id(1, "Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)")

ui.load_image(2, "acisf08478_000N001_r0043_regevt3_srcimg.fits")
ui.set_coord(2, "physical")
ui.notice2d_id(2, "Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)")

ui.set_source(1, ui.gauss2d.mdl1 + ui.scale2d.bmdl1 / 1000)
ui.guess(1, mdl1)
mdl1.fwhm = 5
mdl1.ampl = 0.45
bmdl1.c0 = 10

ui.fit(1)


ui.set_source(2, ui.gauss2d.mdl2 + ui.scale2d.bmdl2 / 1000)
ui.guess(2, mdl2)
mdl2.fwhm = 5
mdl2.ampl = 0.45
bmdl2.c0 = 10

radius = ui.create_model_component("scale2d", "radius")
theta = ui.create_model_component("scale2d", "theta")

radius.c0.min = 0
radius.c0.max = 20
theta.c0.min = -3.14
theta.c0.max = 3 * 3.14

mdl2.xpos = 3145 + radius.c0 * np.cos(theta.c0)
mdl2.ypos = 4520 + radius.c0 * np.sin(theta.c0)

# Add in "fake" components so we know to fit these parameters
#
ui.set_source(2, mdl2 + bmdl2 / 1000 + 0 * (radius + theta))

ui.fit(2)

# Compare xpos/ypos
#
print("\n# Comparison")
print(f"xpos: {mdl1.xpos.val:8.3f} -> {mdl2.xpos.val:8.3f}")
print(f"ypos: {mdl1.ypos.val:8.3f} -> {mdl2.ypos.val:8.3f}")

# Check errors
#
print("\n# Errors")
ui.conf(1)
ui.conf(2)
```

# WIth CIAO 4.15

```
dataset 1: Field() -> Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)
dataset 2: Field() -> Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)
Dataset               = 1
Statistic             = cash
Initial fit statistic = 279.546
Final fit statistic   = 265.795 at function evaluation 387
Data points           = 1516
Degrees of freedom    = 1511
Change in statistic   = 13.7514
   mdl1.fwhm      5.23288     
   mdl1.xpos      3146.36     
   mdl1.ypos      4519.61     
   mdl1.ampl      0.445817    
   bmdl1.c0       11.9838     
AttributeError: 'Parameter' object has no attribute 'cos'

  File "/home/dburke/sherpa/sherpa-main/sherpa-test-data/sherpatest/imgfit.py", line 39, in <module>
    mdl2.xpos = 3145 + radius.c0 * np.cos(theta.c0)
TypeError: loop of ufunc does not support argument 0 of type Parameter which has no callable cos method
```

# With this branch

```
dataset 1: Field() -> Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)
dataset 2: Field() -> Ellipse(3144.52,4518.81,25.2979,19.1119,42.9872)
Dataset               = 1
Statistic             = cash
Initial fit statistic = 279.546
Final fit statistic   = 265.795 at function evaluation 387
Data points           = 1516
Degrees of freedom    = 1511
Change in statistic   = 13.7514
   mdl1.fwhm      5.23288     
   mdl1.xpos      3146.36     
   mdl1.ypos      4519.61     
   mdl1.ampl      0.445817    
   bmdl1.c0       11.9838     
Dataset               = 2
Statistic             = cash
Initial fit statistic = 271.888
Final fit statistic   = 265.795 at function evaluation 611
Data points           = 1516
Degrees of freedom    = 1511
Change in statistic   = 6.09305
   mdl2.fwhm      5.23288     
   mdl2.ampl      0.445817    
   bmdl2.c0       11.9838     
   radius.c0      1.41448     
   theta.c0       -0.279566   

# Comparison
xpos: 3146.360 -> 3146.360
ypos: 4519.610 -> 4519.610

# Errors
mdl1.xpos lower bound:	-0.681136
mdl1.ypos lower bound:	-0.634849
mdl1.fwhm lower bound:	-0.74713
bmdl1.c0 lower bound:	-2.70626
mdl1.ypos upper bound:	0.667707
mdl1.ampl lower bound:	-0.159369
mdl1.xpos upper bound:	0.704789
mdl1.fwhm upper bound:	1.19217
mdl1.ampl upper bound:	0.215634
bmdl1.c0 upper bound:	3.17438
Dataset               = 1
Confidence Method     = confidence
Iterative Fit Method  = None
Fitting Method        = neldermead
Statistic             = cash
confidence 1-sigma (68.2689%) bounds:
   Param            Best-Fit  Lower Bound  Upper Bound
   -----            --------  -----------  -----------
   mdl1.fwhm         5.23288     -0.74713      1.19217
   mdl1.xpos         3146.36    -0.681136     0.704789
   mdl1.ypos         4519.61    -0.634849     0.667707
   mdl1.ampl        0.445817    -0.159369     0.215634
   bmdl1.c0          11.9838     -2.70626      3.17438
radius.c0 lower bound:	-0.679469
theta.c0 lower bound:	-0.476563
mdl2.fwhm lower bound:	-0.74713
bmdl2.c0 lower bound:	-2.70626
mdl2.ampl lower bound:	-0.159368
radius.c0 upper bound:	0.696688
theta.c0 upper bound:	0.488784
mdl2.fwhm upper bound:	1.19217
mdl2.ampl upper bound:	0.215634
bmdl2.c0 upper bound:	3.17438
Dataset               = 2
Confidence Method     = confidence
Iterative Fit Method  = None
Fitting Method        = neldermead
Statistic             = cash
confidence 1-sigma (68.2689%) bounds:
   Param            Best-Fit  Lower Bound  Upper Bound
   -----            --------  -----------  -----------
   mdl2.fwhm         5.23288     -0.74713      1.19217
   mdl2.ampl        0.445817    -0.159368     0.215634
   bmdl2.c0          11.9838     -2.70626      3.17438
   radius.c0         1.41448    -0.679469     0.696688
   theta.c0        -0.279566    -0.476563     0.488784
```

So, we can see that

- we can now do this - that is, create a linked parameter like

      src.xpos = 4000 + radius.c0 * np.cos(theta.c0)

- when we do this, the non-linked parameters agree, and if you deconstruct the linked parameters we get the same result
- the errors for the non-linked parameters agree; it's a lot harder to talk about comparing the errors for the linked parameters

So this is nice (although the syntax for linking parameters and then fitting them is a bit "inside baseball", but that's nothing to do with this PR).

DougBurke:
This addresses #1027 - at least for the NumPy case. I'll have to look at my suggested approach there and see how it compares to this.

sherpa/models/tests/test_parameter.py
```diff
@@ -170,24 +170,55 @@ def setUp_composite():
...
 def test_unop(op):
     porig = setUp_p()
     p = op(porig)
     assert isinstance(p, UnaryOpParameter)
     assert p.val == op(p.val)
 
+def test_unop_string():
+    '''Check that the string output is sensible.
+
+    This implicitly also checks the logic to define a the format string for
```
hamogu, line 186:
```suggestion
    This implicitly also checks the logic to define a format string for
```

hamogu:
I've looked over the discussion in #1027 again. At least for parameter linking, I think the approach here i easier, because numpy does the heavy lifting through the ufunc mechanism - just look at how few lines of code this PR needs to achieve the goal (most of the 118 lines of changes are docs or tests). It's also very natural for the user; there is no need for deep dive into how Sherpa has parameters, no new `FunctionParameter` etc - just type your normal python functions like "model1.par1 = np.sin(model2.par1)" and it just works.

docs/models/index.rst
```diff
@@ -427,6 +427,42 @@ the model to be fit would be given by an expression like::
...
+This includes many commonly used mathematical and trigonometric functions
+such as log, exp, sin, cos, which allows building quite complex parameter
+linkage. Only the numpy versions work here, but not the functions from the
+build-in ``math`` module, so use `numpy.exp` instead of `math.exp`.
+Many more complex functions are available in
+`scipy.special <https://docs.scipy.org/doc/scipy/reference/special.html>`_;
+any arbitrary Python function can be turned into a ufunc with
+`numpy.frompyfunc <https://numpy.org/doc/stable/reference/generated/numpy.frompyfunc.html#numpy.frompyfunc>`_
+and the interface is also available for
+`C extensions <https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html#creating-a-new-universal-function>`.
```
DougBurke, line 448:
```suggestion
`C extensions <https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html#creating-a-new-universal-function>`_.
```

docs/models/index.rst
```diff
@@ -427,6 +427,42 @@ the model to be fit would be given by an expression like::
...
+
+Any `numpy universal function ("ufunc") <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs>`_
+can be used in the linking expression, for example::
+
+    >>> import numpy as np
+    >>> g2.ampl = np.cos(g1.ampl)
+
+This includes many commonly used mathematical and trigonometric functions
+such as log, exp, sin, cos, which allows building quite complex parameter
+linkage. Only the numpy versions work here, but not the functions from the
```
DougBurke, line 441:
```suggestion
linkage. Only the numpy versions work here, **not** the functions from the
```

docs/models/index.rst
```diff
@@ -427,6 +427,42 @@ the model to be fit would be given by an expression like::
...
+Not every possible link function makes sense
+--------------------------------------------
+
+With this flexibility, it is possible to define links that make no sense,
+for example taking the logical not of a parameter that represents a mass or
+turning values of parameters into arrays (Sherpa optimisers can only deal
+with scalar parameters.) In practice, such mistakes
+are easy to spot when displaying a model; because Sherpa is meant to be
+a general and flexible modelling application that works with (almost)
+arbitrary user-defined models, the code puts as few a-priory restrictions
```
DougBurke, line 463:
```suggestion
arbitrary user-defined models, the code puts as few restrictions
```

sherpa/models/parameter.py
```diff
@@ -644,6 +644,17 @@ def _repr_html_(self):
     __mod__, __rmod__ = _make_binop(numpy.remainder, '%')
     __pow__, __rpow__ = _make_binop(numpy.power, '**')
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if not method == '__call__':
+            return NotImplemented
+        if ufunc.nin == 1:
+            return UnaryOpParameter(inputs[0], ufunc, ufunc.__name__)
+        elif ufunc.nin == 2:
```
DougBurke, line 652:
```suggestion
        if ufunc.nin == 2:
```

sherpa/models/parameter.py
```diff
@@ -644,6 +644,17 @@ def _repr_html_(self):
...
+            return BinaryOpParameter(inputs[0], inputs[1], ufunc, ufunc.__name__,
+                                     strformat='{opstr}({lhs}, {rhs})')
+        else:
+            return NotImplemented
```
DougBurke, line 656:
```suggestion
        return NotImplemented
```

DougBurke:
So, we have (when `bar` is a `Const1D` instance):

```
In [54]: foo.c0 = np.exp(2 - (1 - bar.c0))

In [55]: foo.c0.link
Out[55]: <UnaryOpParameter 'exp((2 - (1 - bar.c0)))'>

In [56]: print(foo)
const1d.foo
   foo.c0       linked      403.429 expr: exp((2 - (1 - bar.c0)))           

```

Should this be reporting "np.exp((..." or "numpy.exp((..." instead of `exp`? Ideally yes, but this makes things more complicated and I'm not sure it's that useful [this is to have something to point to when suddenly lots of people start evaluating the string representation of parameters....].

hamogu:
I think `exp` is fine and in fact better than `numpy.exp` because it's closer to how actual math/latex formulas (which our users are familiar with) are written. I thought we are still in the business of hiding the fact that Sherpa in factuses Python underneath to do stuff... needs to look like XSPEC in the UI...

DougBurke:
@hamogu - my comment in https://github.com/sherpa/sherpa/pull/1653#issuecomment-1652246039 was in part because we sometimes rely on the string name for reconstructing data - e.g. either `sherpa.astro.ui.save` or `sherpa.astro.ui.save_all`. Let's check this, with this branch rebased onto upstream/main as I'm using python 3.11:

```
n [1]: from sherpa.astro.ui import *

In [2]: gauss1d.mdl
Out[2]: <Gauss1D model instance 'gauss1d.mdl'>

In [3]: scale1d.param
Out[3]: <Scale1D model instance 'scale1d.param'>

In [4]: mdl.fwhm = 2 * np.exp(np.sqrt((param -23)**2/ 14))
NameError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 mdl.fwhm = 2 * np.exp(np.sqrt((param -23)**2/ 14))

NameError: name 'np' is not defined

In [5]: import numpy as np

In [6]: mdl.fwhm = 2 * np.exp(np.sqrt((param -23)**2/ 14))
File ~/sherpa/sherpa-main/sherpa/models/model.py:591, in Model.__getattr__(self, name)
    589     return parameter
--> 591 NoNewAttributesAfterInit.__getattribute__(self, name)

AttributeError: 'BinaryOpModel' object has no attribute 'sqrt'

Cell In[6], line 1
----> 1 mdl.fwhm = 2 * np.exp(np.sqrt((param -23)**2/ 14))

TypeError: loop of ufunc does not support argument 0 of type BinaryOpModel which has no callable sqrt method

In [7]: mdl.fwhm = 2 * np.exp(np.sqrt((param.c0 -23)**2/ 14))

In [8]: print(mdl)
gauss1d.mdl
   mdl.fwhm     linked      715.438 expr: (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14))))           
   mdl.pos      thawed            0 -3.40282e+38  3.40282e+38           
   mdl.ampl     thawed            1 -3.40282e+38  3.40282e+38           

In [9]: save_all("test-save.py")

In [10]: save("test-save.bin")

In [11]: quit
```

If we look at `test-save.py` (removing all the unrelated text) we can see we do use the "name" field:

```
% cat test-save.py  | grep link
link(mdl.fwhm, (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14)))))
```

So this is potentially an issue. You can see this with

```
% python test-save.py 
  File "/home/dburke/sherpa/sherpa-main/test-save.py", line 72, in <module>
    link(mdl.fwhm, (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14)))))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/ui/utils.py", line 529, in __call__
    _check_str_type(name, "name")
  File "/home/dburke/sherpa/sherpa-main/sherpa/ui/utils.py", line 84, in _check_str_type
    raise ArgumentTypeErr('badarg', argname, "a string")
sherpa.utils.err.ArgumentTypeErr: 'name' must be a string
```

If I hack the code to add the numpy. label (since this script imports numpy and not np) then it works:

```
% diff test-save.py test-save2.py
72,73c72,73
< link(mdl.fwhm, (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14)))))
< 
---
> link(mdl.fwhm, (2 * numpy.exp(numpy.sqrt((((param.c0 - 23) ** 2) / 14)))))
> print(mdl)
% python test-save2.py 
gauss1d.mdl
   mdl.fwhm     linked      715.438 expr: (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14))))           
   mdl.pos      thawed            0 -3.40282e+38  3.40282e+38           
   mdl.ampl     thawed            1 -3.40282e+38  3.40282e+38           
```

For `save`/`restore` it appears to not be an issue (as we've saved the internal state of the objects):

```

In [2]: restore("test-save.bin")

In [3]: print(mdl)
gauss1d.mdl
   mdl.fwhm     linked      715.438 expr: (2 * exp(sqrt((((param.c0 - 23) ** 2) / 14))))           
   mdl.pos      thawed            0 -3.40282e+38  3.40282e+38           
   mdl.ampl     thawed            1 -3.40282e+38  3.40282e+38           
```

hamogu:
Would you prefer to have the "numpy." added just in `save_all` by hacking that particular function or do we also want the display to say "np."?

We could in fact do both and assume that numpy is either imported as "numpy" or as "np" and add a line to the "save_all" to do both imports. Then, we can change the display to return "np." or "numpy." and it would work with either. Of course, it's possible that a user does `import numpy as aertadsadg` but I don't think we neccessarily need to supprt that use case.

DougBurke:
I did try looking if we could identify what name was being used (e.g. `numpy`, `np`, or the always-memorable  `aertadsadg`) but I gave up pretty quickly. And I'm not sure it would have been a sensible thing (e.g. for `save_all` we wouldn't know what import to make at the top of the file...).

I think just saying "np." is enough because

- it's recognizable
- it's shorter

and we can make sure the `save_all` has a `import numpy as np` line. Thoughts?

hamogu:
I agree and I'll look how I get the "np." (or "numpy") into the output in a reliable form. I think it should be easy, I just hae to remember how to get the "fully qualified name" (was `__module__`? - Anyway, easy to find out and I'll look).

What's more difficult is what we do if it's not a numpy function. If it's scipy, that's still not too hard (just use "scipy" and add some magic to the import statement to add import for that). However, numpy allows to turn any Python function into a ufunc, and that could be a function that's defined interactively in that specific session. In that case, we can't fix it and I suggest that we either just issue a warning on "save_all" (can't save function XXXX, which need to be defined separately before "load_all is called in a new session) or just fail. I think that's going to be a very rare case of special uses.

I, for example, did not know that "save_all" even exists before you brought it up on this issue and thus I've clearly never used it. I tent to persist workflows in other ways, typically by having them in a notebook and simply run that top to bottom again when I continue working.

DougBurke:
I'm sure I found an easy way to get the module name from a symbol, but I can no longer remember or find out what I did...

anetasie (commented):
There  are a few fixes to the doc typos that need to be incorporated into the code.

DougBurke:
Doh. What I used in my experiments was simple - the `__module__` attribute, which I think is all we need here:

```
>>> import numpy as np
>>> f = np.sum
>>> f"faked up name: '{f.__module__}.{f.__name__}'"
"faked up name: 'numpy.sum'"
```

DougBurke:
So, with the change above I get

```
>>> mdl.fwhm = 2 * np.exp(np.sqrt((param.c0 - 23) ** 2 / 14))
>>> print(mdl.fwhm)
val         = 715.4376852658457
min         = 1.1754943508222875e-38
max         = 3.4028234663852886e+38
units       = 
frozen      = True
link        = (2 * numpy.exp(numpy.sqrt((((param.c0 - 23) ** 2) / 14))))
default_val = 715.4376852658457
default_min = 1.1754943508222875e-38
default_max = 3.4028234663852886e+38
>>> ui.save_all()
import numpy
from sherpa.astro.ui import *

######### Load Data Sets



######### Set Statistic

...

link(mdl.fwhm, (2 * numpy.exp(numpy.sqrt((((param.c0 - 23) ** 2) / 14)))))


######### Set Source, Pileup and Background Models

>>> 
```

which seems about right.

As a check

```
>>> ui.save_all("foo.boop")
```

and then

```
% python foo.boop
```

hamogu:
That would cover numpy ufuncs, which is the vast majority of ufuncs. However, there are also ufuncs in scipy and you can use `numpy.frompyfunc` to turn any arbitrary Python function into a ufunc. Now, my guess is that 99% of all use cases will just use numpy, and thus I'm personally OK with introducing the bug that we assume ufuncs are automatically numpy funcs, but we should agree on that before we do that.

hamogu:
E.g.
```Python
>>> from scipy.special import gamma
>>> gamma.__name__, gamma.__class__
('gamma', numpy.ufunc)
```

Now, in practice, I don't think many people will couple their parameters using the Gamma function, but I wanted to at least bring it up.

hamogu:
The docs for `save_all` already say " 3. some settings and values may not be recorded.". In fact, it might be considered a security risk to restore arbitrary functions from a file - [someone filed a CVE against numpy in the past because they used the pickle protocol, which can store arbitrary functions](https://security.snyk.io/vuln/SNYK-PYTHON-NUMPY-73513). The numpy team considered that not quite fair, because the whole point of that command is to store arbitrary code; but maybe we don't have to go out of our way to make `save_all` work for non-numpy functions.

sherpa/models/parameter.py
```diff
@@ -644,6 +644,26 @@ def _repr_html_(self):
...
     __pow__, __rpow__ = _make_binop(numpy.power, '**')
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if not method == '__call__':
+            return NotImplemented
+        if hasattr(numpy, ufunc.__name__):
+            name = f"numpy.{ufunc.__name__}"
+        else:
+            # Unfortunately, there is no ufunc.__module__ we could use
+            # This could be a ufinc from e.g. scipy or generated from an
```
DougBurke, line 654:
```suggestion
            # This could be a ufunc from e.g. scipy or generated from an
```

sherpa/models/parameter.py
```diff
@@ -644,6 +644,26 @@ def _repr_html_(self):
     __mod__, __rmod__ = _make_binop(numpy.remainder, '%')
     __pow__, __rpow__ = _make_binop(numpy.power, '**')
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if not method == '__call__':
+            return NotImplemented
+        if hasattr(numpy, ufunc.__name__):
+            name = f"numpy.{ufunc.__name__}"
+        else:
```
DougBurke, line 652: After your mention of scipy.gamma.__class__ being numpy.ufunc, it strikes me we could try something like

```
if isinstance(ufunc, numpy.ufunc):
    name = <not sure what to do here>
```

but, as indicated above, I'm not sue it would be at all helpful (in fact, I imagine you almost certainly have to be a member of numpy.ufunc to get to this code anyway).

I still find it interesting that I can't seem to identify the module that "owns" a particular numpy.ufunc, but life is too short to worry about this.

DougBurke:
Can we add a test to the end of `sherpa/astro/tests/test_serialize.py`. I haven't actually tested this out, so it may need some sort of a tweak.

```
def test_restore_linker_parameter_with_function(clean_astro_ui):
    """Check we can save and restore a numpy ufunc"""

    m1 = ui.create_model_component("gauss1d", "m1")
    m2 = ui.create_model_component("scale1d", "m2")

    delta = numpy.sqrt((m2.c0 - 23)**2)
    m1.fwhm = 2 * numpy.exp(delta / 14)

    m2.c0 = 30

    expected = 2 * numpy.exp(0.5)

    # safety check
    assert m1.link is not None
    assert m1.fwhm.val == pytest.approx(expected)

    restore()

    got = ui.list_model_components()
    assert len(got) == 2
    assert "m1" in got
    assert "m2" in got

    mm1 = ui.get_model_component("m1")
    assert mm1.name == "gauss1d.m1"
    assert mm1.link is not None
    assert mm1.fwhm.val == pytest.approx(expected)
```

sherpa/models/parameter.py
```diff
@@ -644,6 +644,26 @@ def _repr_html_(self):
...
+            # Unfortunately, there is no ufunc.__module__ we could use
+            # This could be a ufinc from e.g. scipy or generated from an
+            # arbitrary Python function with numpy.frompyfunc.
+            # In the latter case, the name will be something like
+            # "func (vectorized)", which looks confusing in out string
+            # represenantion, so we simplify it.
```
wmclaugh, line 658:
```suggestion
            # representation, so we simplify it.
```

sherpa/models/parameter.py
```diff
@@ -815,17 +835,22 @@ class UnaryOpParameter(CompositeParameter):
         The ufunc to apply to the parameter value.
     opstr : str
         The symbol used to represent the operator.
+    strformat : str
+        Format string for printing this operation. Elements
+        than can be used are `arg` and `opstr`, e.g.
```
wmclaugh, line 840:
```suggestion
        that can be used are `arg` and `opstr`, e.g.
```

sherpa/models/parameter.py
```diff
@@ -845,6 +870,10 @@ class BinaryOpParameter(CompositeParameter):
         The ufunc to apply to the two parameter values.
     opstr : str
         The symbol used to represent the operator.
+    strformat : str
+        Format string for printing this operation. Elements
+        than can be used are `lhs`, `rhs`, and `opstr`, e.g.
```
wmclaugh, line 875:
```suggestion
        that can be used are `lhs`, `rhs`, and `opstr`, e.g.
```

sherpa/models/tests/test_parameter.py
```diff
@@ -170,24 +170,72 @@ def setUp_composite():
...
     p, p2 = setUp_composite()
     for comb in (op(p, p2.val), op(p.val, p2),
                  op(p, p2)):
         assert isinstance(comb, BinaryOpParameter)
         assert comb.val == op(p.val, p2.val)
 
+def test_binop_string():
+    '''Check that the string output is sensible.
+
+    This implicitly also checks the logic to define a the format string for
```
wmclaugh, line 218:
```suggestion
    This implicitly also checks the logic to define the format string for
```

sherpa/models/parameter.py
hamogu: You've looked and I've looked. I don't think there is a way to find which file instantiated the `np.unfunc` class that we are looking at here. Future versions of numpy might offer more information, but for now this is all we have if we like it or not.

DougBurke (approved):
Looks good to me. Sorry for the long gestation period.

Issue #1657, opened 2022-12-20
should DataPHA's group/ungroup also change the backgrounds?


DougBurke:
The UI `group` and `ungroup` calls automatically change any related backgrounds. This is not true of the underlying `Data.PHA.group` and `DataPHA.ungroup` methods. As I'm trying to remove as much logic as I can from the UI code, it would make sense to move it to `DataPHA`, but maybe this is one of those times when it is not worth it. Need to think about. 

I note it's not a lot of code, but I would like to keep the UI code for just the data management side as much as possible, rather than setting policy (to mix my metaphors).

DougBurke:
As I noted in the code I wrote to change this, the `notice` routine does something similar - that is, we let the `DataPHA` class handle what to do with the background datasets rather than have it done in the UI layer. And this behavior (for `notice`) was in it in the first version committed to git (I had wondered whether I'd made the change at some point, but I just checked and I hadn't).

PR #1660, opened 2022-12-21
Associated group fixes.

### LINKED ISSUE #1641: ignore warns about mis-matched units twice
Found while working on general code clean up in #1638 and adding test coverage to improve coverage.

This is with 4.15

```
sherpa-4.15.0> from sherpa.utils.logging import SherpaVerbosity

sherpa-4.15.0> with SherpaVerbosity('ERROR'):
          ...:     load_pha(1, "3c273.pi")
          ...:     load_pha(2, "3c273.pi")
          ...: 

sherpa-4.15.0> set_analysis(1, "channel")

sherpa-4.15.0> notice(lo=3)
WARNING: not all PHA datasets have equal analysis quantities
dataset 1: 1:1024 Channel (unchanged)
dataset 2: 0.00146:14.9504 -> 2.8616:14.9504 Energy (keV)

sherpa-4.15.0> notice()
dataset 1: 1:1024 Channel (unchanged)
dataset 2: 2.8616:14.9504 -> 0.00146:14.9504 Energy (keV)

sherpa-4.15.0> ignore(lo=3)
WARNING: not all PHA datasets have equal analysis quantities
WARNING: not all PHA datasets have equal analysis quantities
dataset 1: 1:1024 Channel -> no data
dataset 2: 0.00146:14.9504 -> 0.00146:2.8616 Energy (keV)

```

Calling `ignore` creates the `WARNING` message twice.

It's a minor issue but a **really** easy fix which will be added to #1638

### LINKED ISSUE #1659: DataPHA corner case: remove grouping and the center can not hold
```
>>> src = DataPHA("src", [1, 2, 3, 4], [1, 3, 2, 1])
>>> src.grouping = [1, -1, 1, -1]
>>> src.group()
>>> src.get_dep()
array([1, 3, 2, 1])
>>> src.get_dep(filter=True)
array([4., 3.])
>>> src.grouping = None
>>> src.get_dep(filter=True)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/data.py", line 3827, in get_dep
    dep = self.apply_filter(dep)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/data.py", line 3183, in apply_filter
    gdata = self.apply_grouping(data, groupfunc)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/data.py", line 3315, in apply_grouping
    return do_group(data, groups, groupfunc.__name__)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Changing the `grouping` attribute should clear the `group` flag.

DougBurke:
# Summary

Clean up of the notice/ignore and group/ungroup code to avoid unnecessary work, move logic from the UI layer into the DataPHA class, and ensure that if the grouping field is set to None that the data is not grouped. Fix #1641, #1657, and #1659

# Details

Noticed when working on recent PRs (in particular #1637). These could be separated into separate PRs but I felt they were small enough, and related, to be worth a single PR. Let me know if they should be broken up. There's some more clean-up code in #1667 which could be included here but I decided to separate it out as these are generally "fix some issue, even if minor" whilst #1667 is more straight-up cleanup.

Each commit fixes a bug and maybe does some code cleanup:

- #1641 
  - `ignore` was set up to do the same check as `notice`, but hadn't noticed that because of how the superclass is set up we always go through `notice` anyway (i.e. `ignore` calls `notice`). So we do not need to override `ignore`
- #1657 
  - move the logic for calling group/ungroup for background PHA files to DataPHA from the UI code
- #1659 
  - if the grouping array is cleared (set to None) then clear the group flag

For #1641 I am breaking backwards compatibility, as I am removing a method from the Session class, but it is an internal method (i.e. the name starts with `_`) and is a very-specialized routine. I do not see the need to note this change.

For #1657, I take heart in the fact that notice/ignore behave somewhat similarly - that is the DataPHA code handles the related background components rather than it being in the UI layer (although there's a distinct possibility that I caused this change in a previous PR). I do think this is one case where it makes sense to push the logic to DataPHA rather than have it in the UI layer, even with the logging that the code does (only for rare cases where the background has no grouping array but the source does have). I thought about changing the code from try/except to "check we can do this" but decided not to make that change as it didn't noticeably improve the code (and is somewhat "anti pythonic" anyway).

For #1659, there is the general question of whether we should be able to clear the array (similar for the quality field), but that's a larger question than we want to deal with here, and this code fixes an existing bug (even if it is a definite corner case).

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1929,7 +1929,8 @@ must be an integer.""")
         .. versionchanged:: 4.15.1
            The filter is now re-calculated when the grouping is
            changed. It is suggested that the filter be checked with
-           `get_filter` to check it is still sensible.
+           `get_filter` to check it is still sensible. If set to None
+           then the group flag is cleared.
 
         Returns
         -------
@@ -1965,6 +1966,12 @@ must be an integer.""")
             ofilter = self.get_filter()
 
         self._set_related("grouping", val)
+
+        # If the array has been removed then we need to reset the
+        # group flag.
+        #
+        if val is None and self._NoNewAttributesAfterInit__initialized and self.grouped:
+            self.grouped = False
 
         if ofilter is not None:
             # If the data has been filtered then we want to re-create
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4743,11 +4743,25 @@ must be an integer.""")
         grouping attribute will be used when accessing data values. This
         can be called even if the grouping attribute is empty.
 
+        .. versionchanged:: 4.15.1
+           The grouping status of any background component is now also
+           changed.
+
         See Also
         --------
         ungroup
+
         """
         self.grouped = True
+
+        # Ensure any backgrounds are also grouped.
+        #
+        for bkg_id in self.background_ids:
+            bkg = self.get_background(bkg_id)
+            try:
+                bkg.grouped = True
+            except DataErr as exc:
+                info(str(exc))
 
     def ungroup(self):
         """Remove any data grouping.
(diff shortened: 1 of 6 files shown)


DougBurke:
Rebased just because of all the XSPEC build changes (there was no conflict, I just wanted these to test against the latest XSPEC / build changes)

DougBurke:
Rebased as #1637 (and a lot of other unrelated changes) has been merged.

hamogu (approved):

PR #1663, opened 2022-12-23
Correct the Miniconda Installer for CI

Marie-Terrell:
Switching to the more official method of installing Miniconda for the Conda CI. The old method of chmod'ing and running the installer seems to have broken and is a bit outdated.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -26,8 +26,7 @@ fi
 
 # Download and install conda
 wget https://repo.anaconda.com/miniconda/Miniconda3-latest-${miniconda_os}-x86_64.sh -O miniconda.sh
-chmod +x miniconda.sh
-./miniconda.sh -b -p $miniconda_loc
+bash miniconda.sh -b -p $miniconda_loc
 
 #Source the Conda profile
 source ${miniconda_loc}/etc/profile.d/conda.sh
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -3,7 +3,7 @@ name: Conda CI
 on:
   push:
     branches:
-    - main
+    - '*'
     tags:
     - '*'
   pull_request:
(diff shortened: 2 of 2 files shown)


DougBurke:
Apparently miniconda is already installed, so we could save time and drop this (maybe in a future PR?)

From https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners we get to 
- https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md 
- https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

which both mention miniconda (not sure for macOS about the Intel vs ARM distinction).

Of course, there may be benefit to using the 'latest' miniconda, but I'm not sure it's that important.

Info taken from https://autobencoder.com/2020-08-24-conda-actions/ 

I had originally thought to suggest using one of the setup-miniconda GH actions, but then saw that it's available anyway and made me wonder about why we have these actions. There may be things we could take advantage of (e.g. an environment file)  rather than have the logic coded into the `.github/scripts/*` scripts, but that would be for a follow on PR.

DougBurke (changes requested):
Can we have the commits squashed just to avoid any confusion? 

I think we should look at replacing the wget call with just using the installed miniconda but that can wait for a later PR as it's better to get this fixed first!

harlantc:
I'm of two thoughts on where we get Miniconda. On the one hand, getting the latest conda version could help ID issues a new user might run into, though I can't name any off the top of my head that was found with the actions (Though I did have an issue the other week with an update of conda messing with my zshrc file, removing my nice prompt formatting). On the other, if it's already there, it saves time in setup. However, think Marie had a reason not to use the provided conda due to issues with conda paths? or maybe she was using the Conda GH action.

For squishing, @wmclaugh can do that on merge.

I say we take this for now and go with another PR to look at just using the existing conda install.

DougBurke:
I've thrown together #1669 to test out the "do not re-download miniconda" option, which turned out to be (once I worked through the bugs) not that large a change.

DougBurke:
I'm  closing this as we've taken #1669 instead.

PR #1666, opened 2022-12-23
model-wrapped symbols now reference the actual model (fix #215)

### LINKED ISSUE #215: How to copy over model documentation via the ModelWrapper class?

DougBurke:
# Summary

Wrapped models, accessible from the UI layer, now reference the wrapped model rather than being devoid of any useful information. Fix #215

# Details

This was pulled out of https://github.com/sherpa/sherpa/pull/1638

This adds documentation to the `ModelWrapper` class (for the implementor) and to model-wrapped classes (for the user). It is unclear what information is best for the user, but this is definitely a significant improvement and we can think about updates in a later PR if necessary. For instance, it might be helpful to include the list of parameters, but unfortunately this is rather hard to grab (you can't grab it from the model being wrapped, you'd need to create an instance of that model, which we could do but is not as safe as you might think), so it isn't done here.

There's some "interesting" over-riding of __repr__ / __str__ going on here (i.e. this is existing code, for which I added some tests in #1662), but I didn't want to delve into it as the code works "as is".

# Examples

We now get

```
>>> help(gauss1d)
Help on ModelWrapper in module sherpa.ui.utils:

<Gauss1D model type>
    Create a gauss1d model instance.
    
    One-dimensional gaussian function.

    Instances can be created either as an attribute of gauss1d,
    as long as the attribute does not begin with an underscore,
    or by calling gauss1d directly.
    
    Examples
    --------
    
    The model, here called mdl, is returned but it's also stored in
    the session and can be returned with `get_model_component`:
    
    >>> m1 = gauss1d.mdl
    
    If the model has already been created with the same name then
    the old version will be returned, rather than creating a new
    instance:
    
    >>> m1 = gauss1d.mdl
    >>> m2 = gauss1d("mdl")
    >>> m1 == m2
    True

>>> help(xscflux)
<XScflux model type>
    Create a xscflux model instance.
    
    The XSPEC cflux convolution model: calculate flux
    
    Instances can be created either as an attribute of xscflux,
    as long as the attribute does not begin with an underscore,
    or by calling xscflux directly.
    
    Examples
    --------
    
    The model, here called mdl, is returned but it's also stored in
    the session and can be returned with `get_model_component`:
    
    >>> m1 = xscflux.mdl
    
    If the model has already been created with the same name then
    the old version will be returned, rather than creating a new
    instance:
    
    >>> m1 = xscflux.mdl
    >>> m2 = xscflux("mdl")
    >>> m1 == m2
    True
```

which is not perfect, but a darn site better than the previous, which was

```
Help on ModelWrapper in module sherpa.ui.utils object:

class ModelWrapper(sherpa.utils.NoNewAttributesAfterInit)
 |  ModelWrapper(session, modeltype, args=(), kwargs={})
 |  
 |  Method resolution order:
 |      ModelWrapper
 |      sherpa.utils.NoNewAttributesAfterInit
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  __call__(self, name)
 |      Call self as a function.
 |  
 |  __getattr__(self, name)
 |  
 |  __init__(self, session, modeltype, args=(), kwargs={})
 |      Initialize self.  See help(type(self)) for accurate signature.
 |  
 |  __repr__(self)
 |      Return repr(self).
 |  
 |  __str__(self)
 |      Return str(self).
 |  
... continuing like this with no helpful information to the user
```

### DIFF
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -1112,7 +1112,10 @@ def test_set_default_id_check_invalid(value):
 
 
 class ModelWithDoc(Model):
-    """This has a doc string"""
+    """This has a doc string
+
+    But we don't grab this.
+    """
 
     _hidden = False
 
@@ -1131,12 +1134,29 @@ def test_modelwrapper_init():
     assert repr(wrap) == "<ModelWithDoc model type>"
 
 
+def test_modelwrapper_checks_session():
+    """Check we error out"""
+
+    with pytest.raises(ValueError,
+                       match="session is not a Session instance"):
+        ModelWrapper(ModelWithDoc, None)
+
+
+def test_modelwrapper_checks_model():
+    """Check we error out"""
+
+    s = Session()
+    with pytest.raises(ValueError,
+                       match="modeltype is not a Model instance"):
+        ModelWrapper(s, "Gauss1D")
+
+
 def test_modelwrapper_str_with_doc():
     """Basic ModelWrapper check"""
 
     s = Session()
     wrap = ModelWrapper(s, ModelWithDoc)
-    assert str(wrap) == "This has a doc string"
+    assert str(wrap) == "This has a doc string\n\n    But we don't grab this.\n    "
 
 
 def test_modelwrapper_str_no_doc():
@@ -1147,15 +1167,35 @@ def test_modelwrapper_str_no_doc():
     assert str(wrap) == "<ModelWithNoDoc model type>"
 
 
-def test_modelwrapper_what_is_the_docstring():
+def test_modelwrapper_what_is_the_docstring_has_doc():
     """What do we want help(wrappedmodel) to return"""
 
-    # Check current behavior
-    assert ModelWrapper.__doc__ is None
+    # Check that the class and wrapped docstrings are different.
+    #
+    assert ModelWrapper.__doc__ is not None
+
+    s = Session()
+    wrap = ModelWrapper(s, ModelWithDoc)
+    assert wrap.__doc__ != ModelWrapper.__doc__
+
+    # Check the wrapped docstring references the model and description.
+    #
+    assert wrap.__doc__.startswith("Create a modelwithdoc model instance.\n\n    This has a doc string\n\n    Instances can")
+
+
+def test_modelwrapper_what_is_the_docstring_no_doc():
+    """What do we want help(wrappedmodel) to return"""
+
+    # Check that the class and wrapped docstrings are different.
+    #
+    assert ModelWrapper.__doc__ is not None
 
     s = Session()
     wrap = ModelWrapper(s, ModelWithNoDoc)
-    assert wrap.__doc__ is None
+    assert wrap.__doc__ != ModelWrapper.__doc__
+
+    # Check the wrapped docstring references the model but no description.
+    assert wrap.__doc__.startswith("Create a modelwithnodoc model instance.\n\n    Instances can")
 
 
 @pytest.mark.parametrize("attr", ["_hidden", "_foo_bar"])
--- a/sherpa/ui/tests/test_session.py
+++ b/sherpa/ui/tests/test_session.py
@@ -1114,7 +1114,7 @@ def test_set_default_id_check_invalid(value):
 class ModelWithDoc(Model):
     """This has a doc string
 
-    But we don't grab this.
+    This line is not included in the model-wrapped doc string.
     """
 
     _hidden = False
@@ -1138,7 +1138,7 @@ def test_modelwrapper_checks_session():
     """Check we error out"""
 
     with pytest.raises(ValueError,
-                       match="session is not a Session instance"):
+                       match="session=.* is not a Session instance"):
         ModelWrapper(ModelWithDoc, None)
 
 
@@ -1147,7 +1147,7 @@ def test_modelwrapper_checks_model():
 
     s = Session()
     with pytest.raises(ValueError,
-                       match="modeltype is not a Model instance"):
+                       match="modeltype=Gauss1D is not a Model class"):
         ModelWrapper(s, "Gauss1D")
 
 
@@ -1156,7 +1156,7 @@ def test_modelwrapper_str_with_doc():
 
     s = Session()
     wrap = ModelWrapper(s, ModelWithDoc)
-    assert str(wrap) == "This has a doc string\n\n    But we don't grab this.\n    "
+    assert str(wrap) == "This has a doc string\n\n    This line is not included in the model-wrapped doc string.\n    "
 
 
 def test_modelwrapper_str_no_doc():
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -33,6 +33,7 @@ import numpy
 from sherpa import get_config
 import sherpa.all
 from sherpa.models.basic import TableModel
+from sherpa.models.model import Model
 from sherpa.utils import SherpaFloat, NoNewAttributesAfterInit, \
     export_method, send_to_pager
 from sherpa.utils.err import ArgumentErr, ArgumentTypeErr, \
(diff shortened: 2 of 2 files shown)


sherpa/ui/tests/test_session.py
```diff
@@ -1131,12 +1134,29 @@ def test_modelwrapper_init():
...
+        ModelWrapper(s, "Gauss1D")
+
+
 def test_modelwrapper_str_with_doc():
     """Basic ModelWrapper check"""
 
     s = Session()
     wrap = ModelWrapper(s, ModelWithDoc)
-    assert str(wrap) == "This has a doc string"
+    assert str(wrap) == "This has a doc string\n\n    But we don't grab this.\n    "
```
hamogu, line 1159: Now we do grab it...

sherpa/ui/utils.py
```diff
@@ -332,19 +333,85 @@ def reduce_ufunc(func):
 
 
 class ModelWrapper(NoNewAttributesAfterInit):
+    """Wrap up a model class so we can create instances easily.
+
+    The created instances can be used in a "model language" to create
+    complex expressions. If mdl1 and mdl2 have been created by
+    ModelWrapper then a user can say mdl1.n1 + mdl2.n2 to create model
+    instances named n1 and n2 and then return the expression which
+    represents their sum. Note that the model component is stored in
```
hamogu, line 342: I would understand "note the model components (plural) are stored" (meaning n1 and n2) or "that the model is stored" (singular, but not components). Can you clarify?

sherpa/ui/utils.py
```diff
@@ -332,19 +333,85 @@ def reduce_ufunc(func):
...
+    the session object.
+
+    """
+
+    def __init__(self, session, modeltype, args=(), kwargs=None):
+        # This is an internal class so do not bother with
+        # sherpa.utils.err exceptions.
+        #
+        if not isinstance(session, Session):
+            raise ValueError("session is not a Session instance")
```
hamogu, line 352:
```suggestion
            raise ValueError(f"{session} is not a Session instance")
```
Just a suggestion...

sherpa/ui/utils.py
```diff
@@ -332,19 +333,85 @@ def reduce_ufunc(func):
...
+
+        if not _is_subclass(modeltype, Model):
+            raise ValueError("modeltype is not a Model instance")
```
hamogu, line 355: Not sure if the following suggestion is actually better or not.
```suggestion
            raise ValueError(f"{modeltype} is not a Model instance")
```

sherpa/ui/utils.py
```diff
@@ -332,19 +333,85 @@ def reduce_ufunc(func):
...
         self.args = args
-        self.kwargs = kwargs
+        self.kwargs = kwargs if kwargs else {}
+
+        # Edit the docstring of the new object to hide the
+        # ModelWrapper text and replace it with a reference to the
+        # wrapped class.
+        #
+        mname = modeltype.__name__.lower()
+        prefix = "an" if mname[0] in "aeiou" else "a"
```
hamogu, line 367: Nice attention to detail! I'm not asking for code to detect a silent H or a Y at the beginning of the model name...

hamogu (changes requested):

sherpa/ui/tests/test_session.py
DougBurke: But we don't - we only grab the "This ha a doc string" part of the message. Or at least that's the intent.

sherpa/ui/utils.py
DougBurke: So, I think the message could be

    The session argument is not a Session instance
    f"session={session} is not a Session instance

as I wanted to easily identify which argument is wrong. However, I don't see this really ever getting hit, as this is all internal code, so I'm not sure it's worth the effort. Thoughts?

sherpa/ui/utils.py
DougBurke: Same again, we could say

    f"modeltype={modeltype} is not a Model instance"

sherpa/ui/tests/test_session.py
DougBurke: Well, not in `test_modelwrapper_what_is_the_docstring_has_doc`, where we don't grab the extra text. In `test_modelwraper_str_with_doc` we do grab it because it's the docstring. I could change

    But we don't grab this.

to something like

    This line is not included in the model-wrapped doc string.

which is probably a bit cleared.

sherpa/ui/utils.py
DougBurke: Well,

    mdl1.m1 + mdl2.m2

is a composite model, but that is somewhat irrelevant here. What the line does is create two instances, "m1" from `mdl1` and "m2" from `mdl2`, and these are individually stored in the session object. Maybe this is all too esoteric for this docstring, but I wanted to indicate we can create a model instance within a "model expression", rather than making it look like you have to create components before combining them. Perhaps we need to re-organise this to say something like I(needs to be work-shopped):

    Creates a model instance with a given name - you can say mdl.mname or mdl("mname") - and this stores the
    instance in the session object with the name "mname", over-writing any previous component with this name.
    If mdl1 and mdl2 have been created by ModelWrapper then a user can say mdl1.n1 + mdl2.n2 to create model
    instances named n1 and n2 and then return the expression which represents their sum.

sherpa/ui/utils.py
DougBurke: I should also point out that for this, and the following check, I could not be bothered trying to find a sherpa error class to use, and just went with the normal Python exceptions, as I really don't expect these to ever occur.

There is a discussion we could have about the sherpa error class, but that is a significantly-larger issue that we don't want to touch here.

hamogu: I agree it's not worth the effort; just a suggestion.

sherpa/ui/utils.py
hamogu: I like this suggestion that combines both of our initial ideas, but again, this should, as say, should never be hit, so not worth the effort?

sherpa/ui/utils.py
hamogu: Creates a model instance with a given name - you can say `mdl.mname` or `mdl("mname")` - and either syntax stores the model
instance in the session object with the name "mname", over-writing any previous component with this name.
If mdl1 and mdl2 have been created by ModelWrapper then a user can say `mdl1.n1 + mdl2.n2` to create model
instances named n1 and n2 and then return the expression which represents their sum.

hamogu: Just some slight modification on your suggestion which I think is clearer than the original suggestion.

sherpa/ui/utils.py
DougBurke: I've applied this change, as well as correcting the message, as it should say "Model class" not "Model instance".

sherpa/ui/tests/test_session.py
DougBurke: I've made this change.

hamogu (approved):

PR #1669, opened 2023-01-04
CI do not download miniconda

DougBurke:
# Summary

Use the system conda installation on GitHub CI rather than downloading it ourselves each time.

# Details

Test to see if we can use the installed version of conda rather than download it ourselves each time. An alternative to #1663

I want to see if I can have the conda download happen with a subset (maybe one) test so we can ensure we have the "latest version" tested. Hmmm. I've decided I don't have the time/energy to explore this approach.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash -e
 
 if [ "`uname -s`" == "Darwin" ] ; then
-    miniconda_os=MacOSX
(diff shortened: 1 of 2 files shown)


Marie-Terrell (approved):
Looks good to me! I prefer this over #1663 for a cleaner fix. 

Aside: We can't do this for the Linux side of the deployment workflow at the moment because the docker container we use for Linux does not have the miniconda setup and env variables. However, this is perfectly fine here. We can always install Conda or update it down the road in a different way if something arises.

Issue #1616, opened 2022-10-25
Why does the model calc method require the parameter arrays to be sent in?


DougBurke:
In `sherpa/models/model.py` the base `Model` class contains (this is taken from the very first git commit for Sherpa [*]):

    def calc(self, p, *args, **kwargs):
        raise NotImplementedError

The idea being that we send in the model parameters as the first argument (`p`), and the axis information using `args`. This setup makes the compiled model code easy to write, since we have (again, from the first commit) for the `Box1D` model:

    @modelCacher1d
    def calc(self, *args, **kwargs):
        kwargs['integrate']=bool_cast(self.integrate)
        return _modelfcts.box1d(*args, **kwargs)

Note that here we've dropped the `p` argument for the `calc` method, but it equates to the same thing. That is, `_modelfcts.box1d` is sent the parameters first and then the axis information, and then the C++ code unpacks this with the `modelfct1d` template from `sherpa/include/sherpa/model_extension.hh` which contains the following parsing code (showing that the first argument is a parameter array, and this is again from the first commit):

    static char *kwlist[] = {(char*)"pars",(char*)"xlo",(char*)"xhi",(char*)"integrate", NULL};
    if ( !PyArg_ParseTupleAndKeywords(args, kwds, (char*)"O&O&|O&i", kwlist,
			   (converter)convert_to_array< ArrayType >, &pars,
			   (converter)convert_to_array< ArrayType >, &xlo,
			   (converter)convert_to_array< ArrayType >, &xhi,
			   &integrate) )
      return NULL;

Now, there are times that I take advantage of calling the model's `calc` method with different parameter values to those set in the object, but this is generally done because it can be done, and is not necessarily the simplest code to understand. I wonder if we changed the python-level API to drop sending in the parameter values, so the `Box1D` routine would change to something like

    def calc(self, *args, **kwargs):
        pars = [p.val for p in self.pars]
        return _modelfcts.box1d(pars, *args, integrate=self.integrate)

This would make these routines a bit longer, but we could do something similar to what is done for the XSPEC models and remove a lot of the duplication. The main changes would come from the composite models, since `BinaryOpModel` could be simplified from

    def calc(self, p, *args, **kwargs):
	nlhs = len(self.lhs.pars)
        lhs = self.lhs.calc(p[:nlhs], *args, **kwargs)
        rhs = self.rhs.calc(p[nlhs:], *args, **kwargs)
        try:
            val = self.op(lhs, rhs)
        except ValueError:
[...]

PR #1671, opened 2023-01-11
Rework basic model calc

DougBurke:
This is a follow up to #1617 (to avoid conflicts, it is a separate issue)

# Summary

Re-work the sherpa.models.basic models to use a new Callable super class.

# Details 

This includes "documentation meaning argument lists", as `calc(*args, **kwargs)` has got changed to `calc(p, *args, **kwargs)` to better match the superclass. This has lead to issue #1616 where I wonder why we want the parameters to be sent to `calc`, since the object knows what they are, but that is a much-larger discussion.

The main change is a cleanup of the model classes, such as using a XSPEC-style system for defining the model to call, meaning that `calc` need not be redefined for most models (which better points out those that do things differently).


DougBurke:
Rebased to address conflict.

DougBurke:
Rebased as #1617 has been merged.

DougBurke:
Rebased as #1797 has been merged.

Issue #1642, opened 2022-12-03
ui save_source/model/resid issue for 2D data


DougBurke:
This is only a problem for users of sherpa.ui, not sherpa.astro.ui, and only for Data2D/Data2DInt cases

```
In [1]: from sherpa.ui import *

In [2]: dataspace2d([2, 3], dstype=Data2D)

In [3]: set_source(gauss2d.bob)

In [4]: save_source("delme.dat")
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [4], line 1
----> 1 save_source("delme.dat")

File <string>:1, in save_source(id, filename, clobber, sep, comment, linebreak, format)

File ~/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/ui/utils.py:4458, in Session.save_source(self, id, filename, clobber, sep, comment, linebreak, format)
   4389 """Save the model values to a file.
   4390 
   4391 The model is evaluated on the grid of the data set, but does
   (...)
   4455 
   4456 """
   4457 clobber = sherpa.utils.bool_cast(clobber)
-> 4458 self._save_type('source', id, filename, clobber=clobber, sep=sep,
   4459                 comment=comment, linebreak=linebreak, format=format)

File ~/miniconda3/envs/ciao415-b2-py310/lib/python3.10/site-packages/sherpa/ui/utils.py:4306, in Session._save_type(self, objtype, id, filename, **kwargs)
   4299 funcname = "get_{}_image()".format(objtype)
   4300 """
   4301 imgtype = getattr(self, funcname, None)
   4302 if imgtype is None:
   4303     # raise AttributeErr('attributeerr', 'session', fname)
   4304     raise AttributeError("Unable to find " + funcname)
   4305 """
-> 4306 imgtype = getattr(self, funcname)
   4308 obj = imgtype(id)
   4309 args = [obj.y]

AttributeError: 'Session' object has no attribute 'get_source_image()'

```

Found whilst doing pylint fixes in #1638 

It's an easy fix. The problem is in

https://github.com/sherpa/sherpa/blob/9301250ece6fddfe592a0abf9fa9f688a7caefc8/sherpa/ui/utils.py#L4293-L4326

where we ask for an attribute named `"get_xxx_image()"` rather than `"get_xxx_image"` (for various `xxx`). This was I bug I introduced in when I was doing a previous round of linting (via PEP8) in #175 , and just shows why I know try to make sure we have coverage changes for such changes

PR #1672, opened 2023-01-12
models: f-string changes

DougBurke:
# Summary

Change the models code to use f-strings where appropriate rather than other methods of string interpolation. There is no change in behavior.

# Details

Convert to f-strings for most cases **in the files I'm looking at**. The warning calls, which use the logging infrastructure, have generally used %s-encoded results but passed as arguments (so `("message %s more message", value)` rather than `("message {value} more message")` or `("message %s more message" % value)`), to please pylint.

I have various versions of these changes in a number of PRs, but they are all subtly different (e.g. change quote style, add or remove a new-line character) which just makes rebasing things harder than the need to be. So I've pulled this out into a simple-to review PR.

Unlike other f-string changes I have coverage for all the lines I'm changing here, so we don't need to worry about issues like #1642 (although, to be honest, having the lines be covered in the tests doesn't catch all problems, but tends to catch the major ones).

### DIFF
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -878,8 +878,7 @@ class MultiResponseSumModel(CompositeModel, ArithmeticModel):
         self.models = models
         self.grid = grid
 
-        expr = ','.join([f'{m.name}({source.name})'
-                         for m in models])
+        expr = ','.join([f'{m.name}({source.name})' for m in models])
         name = f'{type(self).__name__}({expr})'
         CompositeModel.__init__(self, name, (source,))
 
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015-2018, 2019, 2020, 2021, 2022
+#  Copyright (C) 2010, 2015-2018, 2019, 2020, 2021, 2022, 2023
 #  Smithsonian Astrophysical Observator
 #
 #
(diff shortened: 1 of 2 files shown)


DougBurke:
Rebased as #1674 has been fixed.

sherpa/models/model.py
```diff
@@ -735,14 +737,14 @@ def _set_thawed_pars_maxes(self, vals):
             v = SherpaFloat(v)
             if v < p.hard_min:
                 p.max = p.hard_min
-                warning(('value of parameter %s maximum is below ' +
-                         'hard minimum; ' +
-                         'setting to hard minimum') % p.fullname)
+                warning('value of parameter %s maximum is below '
+                        'hard minimum; setting to hard minimum',
+                        p.fullname)
```
hamogu, line 742: Could go to an f-string here, too.

hamogu (commented):
This is good as is and it's definitely an improvement over what's in the code before this PR. I'm leaving a few suggestions, but I realize that this is really bike-shedding on a nitpick level. 
If @DougBurke says "no" I'm also going to approve - just giving him a change to look first. Sometimes it's "fun" to do something that's so simple as opposed to looking at the real problems ;-)

sherpa/models/model.py
DougBurke: We could. However, pylint complains about this as this is a logger, and it knows that you can "save time" by not interpolating the string for those cases where the logger is going to drop the message. AKA https://vald-phoenix.github.io/pylint-errors/plerr/errors/logging/W1201.html

Although I admit it's not a useful time saving here, I was just trying to avoid too many red squiggles from my lint checks!

sherpa/astro/instrument.py
```diff
@@ -1048,7 +1048,7 @@ def __init__(self, rmf, model, pha=None):
         self.otherkwargs = None
         self.pars = ()
         CompositeModel.__init__(self,
-                                ('%s(%s)' % ('apply_rmf', self.model.name)),
```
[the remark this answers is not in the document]
DougBurke: I can imagine this is the sort of thing that happened after some initial development work or refactoring that never got noticed.

hamogu (approved):

PR #1673, opened 2023-01-12
Tests: allow tests added in #1662 to work with crates

DougBurke:
# Summary

Allow tests added in #1662 to run with the crates I/O backend.

# Details

The I/O choice has surprising downstream consequences, which these tests added in #1662 help highlight. They are only relevant if you have the crates backend selected, which we currently don't test in our CI runs.

Changes are

- do we support/recognize the `comment` keyword argument (apparently not for crates)
- we don't write out 2D images in ASCII for crates (even though we could)

I have attempted to have the code run "as much as possible" here (so trigger on this crates-only behavior) which makes the tests a little ungainly, but generally can be hidden within a function. One issue we have - but it is actually present with the `@requires_fits/plotting` decorators, is that we check whether a particular library (eg crates or astropy) is importable, but not actually "is the backend we have selected". Technically this can bite us, but only in environments where we have both crates and astropy installed and we select the astropy backend. This is something to think about at a later date.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -3185,7 +3185,7 @@ def check_text_output(path, header, coldata):
     # possible to access this so we hard code the knowledge here. If
     # we ever get another I/O backend we need to revisit this. It also
     # assumes that if crates is available then it is in use, which
-    # is technically incorrect but should pass for out tests.
+    # is technically incorrect but should pass for our tests.
     #
     if has_package_from_list("pycrates"):
         expected = f"#TEXT/SIMPLE\n# {header}\n"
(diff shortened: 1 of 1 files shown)


DougBurke:
@Marie-Terrell if the tests pass CI I'll update the release notes.

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -3157,6 +3175,29 @@ def test_get_conf_opt_none(session):
...
+    """The output format for the various backends is subtly different.
+
+    path is the pathlib object representing the file
+    """
+
+    # This requires knowledge of requires_fits, but we don't make it
+    # possible to access this so we hard code the knowledge here. If
+    # we ever get another I/O backend we need to revisit this. It also
+    # assumes that if crates is available then it is in use, which
+    # is technically incorrect but should pass for out tests.
```
Marie-Terrell, line 3188: "for out tests." ->"for our tests." ?

Marie-Terrell (approved):
This passed in a CIAO install on macOS Catalina and the changes all look good to me. Added one note checking on a comment, but otherwise I'm good with this.

PR #1675, opened 2023-01-13
Bump ds9 to 8.4.1 - Ubuntu 20 and macOS BigSur

### LINKED ISSUE #1674: DS9 location has changed so CI runs break
Looks like the ubuntu18/ directory has gone - which breaks are CI runs - and we probably want to bump to v 8.4.

https://ds9.si.edu/download/

Marie-Terrell:
Bump to the latest DS9 version and update to ubuntu20 version. This resolves #1674


DougBurke (approved):
LGTM!

PR #1630, opened 2022-11-16
Add support for XSPEC 12.13.0

DougBurke:
# Summary

Add explicit support for XSPEC 12.13.0. Sherpa can be built against XSPEC 12.13.0 without this PR, but this adds in support for the new `cglumin` convolution model added in this release.

# Details

Built on #1609, which means that only the last commit is part of this PR. Recent merges have simplified the history of this PR and there's no-need to work-around some differences with the conda build of XSPEC we use for testing, as well as having recent versions of XSPEC available for testing, including on macOS.

This adds "official" support for building with XSPEC 12.13.0 - that is, the build will not warn you that the version is unknown - and adds support for the cglumin convolution model (as long as you have 12.13.0 or later). Fortunately the changes aren't huge.

PR #1676, opened 2023-01-18
update model macro definitions for compiled code

DougBurke:
# Summary

Re-work the macro definitions used to build the compiled models and use this to clean up the handling of the XSPEC models, in particular table models. There is no functional change.

# Details

Note, I have now made this PR depend on #1681 as there are some urgent test updates needed there. This has now been merged, which simplifies this PR.

PR chains that will depend on this:

 - #1396 
   - #1609 
     - #1615
     - #1630
 - #1617
 - #1179 
   - #830

I have a lot of XSPEC follow-up code and I am trying to break things up so that they are

- easy to review
- reduce the chance for PR collisions, or at least make it a lot easier to deal with when we do so

The change to the include files for the model macros is one of the latter.


Marie-Terrell (approved):
Checked with a ciao-install build (xspec 12.12.1s Linux and macOS) and a Conda build (xspec 12.13.0c). All of the changes look good.

wmclaugh (approved):

Issue #1682, opened 2023-01-30
sherpa.astro.fake.fake_pha examples are for the ui layer


DougBurke:
Found when answering a user's questions about PHA data fake. I noticed that

https://github.com/sherpa/sherpa/blob/6f91e8d39698cfe660a20f5d614a8eb0feda3add/sherpa/astro/fake.py#L88-L118

are really for the sherpa.astro.ui.utils.fake_pha call, rather than this lower-level call.

**EDITED TO ADD** I think we need to review our existing `fake` documentation (pun intended) so we can use this as a starting point.

PR #1684, opened 2023-02-01
Fix fake_pha docs

DougBurke:
# Summary

Fix the documentation and some corner cases of sherpa.astro.fake.fake_pha. There are several times when the simulated PHA output would not be correct if the PHA contains at least one background component or a pileup model was used.

# Details

This started out as addressing #1682 - a problem with the documentation - but then I realized there were a few clean ups that could be made, which then lead to looking at the handling of background data and hence finding the bugs and issues in #1685. The following has seen numerous re-writes, as I split code out of this PR into other PRs, so I'm not 100% convinced it makes sense anymore.

This now uses 

- #1733
- #1734

as it was beginning to get a bit too large. It used to include RNG changes for `fake_pha` (to send in a NumPy generator rather than rely on the global state) but I am doing this in #1735 so I've removed the changes here.

We have two main changes - `fake: rework fake_pha` and `utils: fake_pha API additions` - in between which I make a small code-clean-up commit. I am trying to move to a "better" API for fake_pha - both for the UI code and the lower-level API and I do not think I have found it.

One thing to note is that we have two "modes" of simulate when given a background dataset:

- treat the counts as the distribution and simulate that, which is going to be a problem when a bin has 0 counts (i.e. the simulation will not be "great"), and the existing documentation points out this as a case where you might want to use a model
- use a background model to simulate the data

The latter is surprisingly tricky, and is the cause of most of the problems we are about to talk about. There are three problem cases noted in #1685 - namely

1. when there are multiple background components we divide the background signal by an extra `nbkg` term
2. when using a model, we have some interesting messing around with response and exposure times, but we should not be using the background dataset here: we should be using the source PHA and response instead
3. when a pileup model is needed then the current code doesn't work well as it assumes the responses are linear (so can just be added) and this is not one of those cases

I had originally fixed these individually, but it turns out to be a bit if a fool's errand as we can address them in one go by ensuring that

1. we send to `fake.fake_pha` a model which is `response(source_model + scaling * bgnd_model)` as that removes all the complexity of scaling the background models or how we deal with the response (e.g. for a pileup model). This is basically changing `utils.fake_pha` from using `get_source` to using `get_model` (as that does all the response and background scaling), but it means large changes to `fake.fake_pha`, such as the `bkgs`, `add_source`, and `pileup_model` parameters becoming obsolete
2. much of the background logic is now in `utils.fake_pha`, which is not ideal (I want the ui level to have minimal logic) but understandable, in part because of the data management that the ui level does
 
# To think about

So, fake_pha requires that you have a pha loaded. Given you can give the response, should we add a way to send in the model and then this could be dropped - i.e. the need to have a PHA dataset already created (as requested by a user)? Actually, as Aneta points out, this is not true of `sherpa.astro.ui.utils.Session.fake_pha`.

We do not have any pileup-model tests of fake_pha to check this out.

I am not happy with the existing API but as I don't really use this routine it's not clear I understand what people want.

An open question is whether we should be simulating the background data when a background model is used. I think so, but need to think it through, as the interface is not clear, and we have a way to work around this (as hinted at by the existing documentation; use fake_pha on the background dataset and copy it over(.

Should we allow a mode where a user can just simulate a PHA dataset from the existing counts (similar to how we can do the background). Something like "hey, if we vary this data what are the fits like"?

Should we require, when processing backgrounds, that all components have a background or no components do? At the moment we allow a mix. I could come up with a reason for this (e.g. a model background plus an observational component taken from blank-sky like observations), but I'm not 100% sure.


hamogu:
Given that I extensively use `fake_pha` for Arcus and Lynx work, I have some thoughts on what @DougBurke lists as "to think about", but I'll get to that after #1734 is in. This comment it mainly for me to remember to do that.

DougBurke:
Some parts of this could be carved out and done first (e.g. adding the tests) if need be. I have got it to an "okay" place but it's worth thinking about what we want to do and what we are willing to say to the user "you need to deal with this yourself" (this is, in part, from several discussions I've had with users). But baby steps!

DougBurke:
Rebased to pick up changes to #1734

sherpa/ui/utils.py
```diff
@@ -4089,23 +4089,31 @@ def dataspace2d(self, dims, id=None, dstype=sherpa.data.Data2D):
...
-    def fake(self, id=None, method=sherpa.utils.poisson_noise):
+    def fake(self, id=None, method=sherpa.utils.poisson_noise, rng=None):
         """Simulate a data set.
 
         Take a data set, evaluate the model for each bin, and then use
         this value to create a data value from each bin. The default
         behavior is to use a Poisson distribution, with the model
         value as the expectation value of the distribution.
 
+        .. versionchanged:: 4.15.1
```
DougBurke, line 4100:
```suggestion
        .. versionchanged:: 4.16.0
```

sherpa/utils/__init__.py
```diff
@@ -1786,13 +1786,19 @@ def get_error_estimates(x, sorted=False):
     return (median, lval, hval)
 
 
-def poisson_noise(x):
+def poisson_noise(x, rng=None):
     """Draw samples from a Poisson distribution.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 1792:
```suggestion
    .. versionchanged:: 4.16.0
```

DougBurke: This code conflicts with #1735 (although it's actually doing the same thing, mechanically it will lead to a conflcit).

DougBurke:
Rebased because of recent changes, in particular 

- #1734

DougBurke:
Rebased to remove the rng changes.

anetasie:
> So, fake_pha requires that you have a pha loaded. Given you can give the response, should we add a way to send in the model and then this could be dropped - i.e. the need to have a PHA dataset already created (as requested by a user)?

@DougBurke I thought that one does not need the PHA data sets, but only responses and model definition [as in this thread.](https://cxc.harvard.edu/sherpa/threads/fake_pha/) Perhaps I'm reading this out of context.

DougBurke:
@aneta - you are correct. You don't need a PHA loaded as the following works

```
sherpa In [1]: !ls 3c273*
3c273.arf  3c273_bg.pi	3c273.pi  3c273.rmf

sherpa In [2]: md = xsphabs.gal * xsapec.src

sherpa In [3]: help(fake_pha)


sherpa In [4]: set_source(md)

sherpa In [5]: fake_pha(1, arf="3c273.arf", rmf="3c273.rmf", exposure=1e4)
Reading APEC data from 3.0.9


sherpa In [6]: plot_data()
```

The need to `set_source` is perhaps a bit strange, but not that hard to understand.

I wonder if the user who was asking me about this was perhaps confused by our documentation, or I mis-understood the request. We should ask the mysterious person - @adonath - for clarification and whether there are other suggestions that we could look at.

anetasie:
Yes, we should ask for clarification.

`set_source`  is needed for getting the model predicted counts and that was confusing in our docs and @adonath assumed that the data is needed.

Also,  is the background data need to be present?

DougBurke:
The example I showed had no background. I will say I don't like how we handle the background since there's several options we could do, and I don't think it's clear what user's want

- ignore any background
- just use the existing background (ie it just gets added in; not really great, but it's something) 
- simulate the background (treat it as a model we can draw from); this is not ideal and if there's bins with 0 in it is not really useful, but maybe it would be useful
- simulate with a background model
-

DougBurke:

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
...
+        the same shape.
+
+    Notes
+    -----
+
+    The model must include the relevant response, and if a background
+    model is required then it must be included. The model is evaluated
+    to create the predicted counts. If the PHA dataset contains any
+    backgrounds and include_bkg_data is set then the counts are added,
+    after scaling, to the predicted counts. The predicted counts are
```
hamogu, line 82: the predicted counts ... ? From the model? i.e. if the model contains a background component and the backgrounds are set, do I get the background twice in the simulated data?

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
...
+    >>> yarf = 100 * (elo < 3.2) + 50 * (elo >= 3.2)
+    >>> arf = create_arf(elo, ehi, yarf)
+    >>> rmf = create_delta_rmf(elo, ehi, e_min=elo, e_max=ehi)
+    >>> pha = DataPHA("faked", chans, chans * 0)
+    >>> pha.set_response(arf=arf, rmf=rmf)
+    >>> pha.exposure = 50000
+    >>> resp = Response1D(pha)
+    >>> full_model = resp(gal * clus)
+    >>> rng = numpy.random.default_rng()
+    >>> fake_pha(pha, full_model, rng=rng)
```
hamogu, line 133: I don't see an `rng` argument to this function. Something missing fro ma rebase/merge with #1735?

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
...
+    Simulate the data from an absorbed APEC model, where the response
+    is manually created (a "perfect" RMF and the ARF is flat, with a
+    break at 3.2 keV), and then use the Sherpa plot objects to display
+    the simulated data and model. Note that the exposure time must be
+    set (if the model normalization is set).
+
+    >>> import numpy as np
+    >>> from sherpa.astro.data import DataPHA
+    >>> from sherpa.astro.instrument import Response1D, create_arf, create_delta_rmf
+    >>> from sherpa.astro.xspec import XSphabs, XSapec
```
hamogu, line 113:
```suggestion
    >>> # Set up the model we want to use for the simulation
    >>> from sherpa.astro.xspec import XSphabs, XSapec
```

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
...
+    >>> gal = XSphabs()
+    >>> clus = XSapec()
+    >>> gal.nh = 0.12
+    >>> clus.kt = 4.5
+    >>> clus.abundanc = 0.3
+    >>> clus.redshift = 0.23
+    >>> clus.norm = 6.3e-4
+    >>> chans = np.arange(1, 101, dtype=np.int16)
```
hamogu, line 121: This example is so long, some intermediate explanation might help.
```suggestion
    >>> # Generate ARF and RMF. In this example, we generate them from scratch, but in pratice
    >>> # they are typically read in from a file
    >>> chans = np.arange(1, 101, dtype=np.int16)
```

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
...
+    >>> arf = create_arf(elo, ehi, yarf)
+    >>> rmf = create_delta_rmf(elo, ehi, e_min=elo, e_max=ehi)
+    >>> pha = DataPHA("faked", chans, chans * 0)
+    >>> pha.set_response(arf=arf, rmf=rmf)
+    >>> pha.exposure = 50000
+    >>> resp = Response1D(pha)
+    >>> full_model = resp(gal * clus)
+    >>> rng = numpy.random.default_rng()
+    >>> fake_pha(pha, full_model, rng=rng)
+    >>> pha.set_analysis("energy")
```
hamogu, line 134:
```suggestion
    >>> # Plot simulated spectrum and input data
    >>> pha.set_analysis("energy")
```

sherpa/astro/fake.py
```diff
@@ -134,62 +181,21 @@ def fake_pha(data, model,
...
-            bkg_models = {}
-
-        nbkg = len(data.background_ids)
-        b = 0
+    model_prediction = data.eval_model(model)
+
+    if include_bkg_data:
+        # If there are background components then we sum up the data,
+        # after scaling to match the source data set, and use that as a
+        # source term. Note that get_background_scale will correct for
```
hamogu, line 192: It's amazing how someone who understand Sherpa's internals can clean up my crappy code.

sherpa/astro/fake.py
```diff
@@ -21,111 +21,158 @@
 The ``fake_pha`` routine is used to create simulated
 `sherpa.astro.data.DataPHA` data objects.
 '''
+
+from functools import reduce
+import operator
```
hamogu, line 26: Where are those imports used?

sherpa/astro/tests/test_fake_pha.py
```diff
@@ -23,11 +23,16 @@
 At present it is *very* limited.
 """
 
+from functools import reduce
+import operator
```
hamogu, line 27: Where are those imports used?

sherpa/astro/tests/test_fake_pha.py
```diff
@@ -157,24 +161,21 @@ def test_fake_pha_basic(method, is_source, expected1, expected2, has_bkg, reset_
...
-    # or changes due to bin-widths varying with position.
-    #
     assert data.counts == pytest.approx(expected1)
 
     # Essentially double the exposure by having two identical arfs
     #
     data.set_arf(arf, 2)
     data.set_rmf(rmf, 2)
-    fake_pha(data, mdl, is_source=is_source, add_bkgs=False, method=method)
+    resp = data.get_full_response()
```
hamogu, line 170: Should this resp have changed compared to a few line above? If not, no harm pulling that again; I'm just wondering if the intent here is to pull compare this to the response in in line 137 to ensure that the fake_pha did not mess with it.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -50,6 +51,7 @@ def test_fake_pha_no_rmf(idval, clean_astro_ui):
         ui.fake_pha(idval, arf=None, rmf=None, exposure=1000.0)
 
 
+@requires_fits
```
hamogu, line 54: How did this pass without "requires_fits" before? It did, so why do we need to make this as requires_fits now?

sherpa/astro/ui/utils.py
```diff
@@ -9261,27 +9275,92 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+                d.delete_background(bkg_id)
+
+            self.set_bkg(idval, bkg)
+            include_bkg_data = True
+
+        # The source model (which includes any background model
+        # components and the response) is used. The decision here is
+        # whether to set the include_bgnd_flag to True or not. The
+        # idea is that the background components are included IF there
+        # is no source model set. If the used does not want this then
```
hamogu, line 9333:
```suggestion
        # is no source model set. If the user does not want this then
```

sherpa/astro/ui/utils.py
```diff
@@ -9261,27 +9275,92 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+            for bkg_id in d.background_ids:
+                self.delete_bkg_model(idval, bkg_id)
+                d.delete_background(bkg_id)
+
+            self.set_bkg(idval, bkg)
+            include_bkg_data = True
+
+        # The source model (which includes any background model
+        # components and the response) is used. The decision here is
+        # whether to set the include_bgnd_flag to True or not. The
```
hamogu, line 9331: Do you mean
```suggestion
        # whether to set the include_bkg_data to True or not. The
```

hamogu (commented):
This looks pretty good, but it's a very tricky piece of code with multiple backgrounds and different exposure times and scalings. I can't say that I can follow this from first principles to ensure it's correct, but it's better than before. I'm probably the author of the code for the "several bugs" that this PR fixes, so we should not rely on me reading the code alone. I will try this out with Arcus responses in the next few days - that's what originally motivated me to make this code work for multiple responses in the first place. I can't text backgrounds there, but at least I can run through cases with multiple responses.

sherpa/astro/fake.py
DougBurke: Yeah - so this was where I started what became #1735 but when I separated the two I obviously missed something. We may want to add this example back once #1735 is merged (so it depends on which gets done first ;-)

sherpa/astro/fake.py
DougBurke: I found it hard to distill the behaviour, so a good review of the docs is helpful! I keep on thinking that if we find the right way to do this then the description will become easier, but the various different possibilities for backgrounds makes this hard. 

I think I'd like to reduce the different possibilities but there may be some use for some of the current approaches, and almost-certainly someone using them, so I think that's a longer term goal than this PR.

sherpa/astro/fake.py
DougBurke: These are left over from something (maybe the RNG changes or maybe some attempt I made to be a bit-more functional in the code). I'll remove them (and the other files).

sherpa/astro/fake.py
DougBurke: It's one of the issues I have with the "documentation can also be a test" approach, as sometimes it leads to a lot of extra code...

sherpa/astro/fake.py
DougBurke: I'm not convinced the existing documentation describes (or, at the time this code was written, described) the assumptions that various operations require or make, so it can be hard to work out what the current "best approach" is.

sherpa/astro/tests/test_fake_pha.py
DougBurke: So, we've just added a new pair of responses, so I think the response should be different. So we should label it differently at least. The behaviour of the new response is checked in line 173 where we check that the faked data matches the `expected2` array rather than `expected1` (see line 164). Does this help?

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: I'd have to check. There's a possibility that the code paths have changed so that it now ends up calling a routine from `sherpa.astro.io` when it did not before, but it's been so long since I looked at this in detail that I just don't remember,

Alternatively the addition isn't needed....

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -65,6 +67,7 @@ def test_fake_pha_missing_rmf(idval, clean_astro_ui, tmp_path):
         ui.fake_pha(idval, None, str(rmf), 1000.0)
 
 
+@requires_fits
```
DougBurke, line 70: See https://github.com/sherpa/sherpa/pull/1684/files#r1334895272

sherpa/astro/ui/utils.py
```diff
@@ -9261,27 +9275,92 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
...
+        #
+        # If bkg is a DataPHA object then remove all the existing
+        # backgrounds and replace them with the background, and set
+        # include_bkg_data. There is also the need to restore a
+        # background model if set: given that there's only the
+        # possibility of using a single background dataset this
+        # potentially loses information, but it's unclear what the
+        # user really expects here given the existing API.
+        #
+        include_bkg_data = False
```
DougBurke, line 9294: I will note that I'm not happy that all the complexity has moved back into this version of fake_pha, as it makes it hard for users of the "OO" interface to replicate this logic. However, there is a reason for this - the handling of background / model data can be rather different between the UI and "OO" interfaces.

Still, I'd hope we can clean this up so that more of the logic is in `sherpa.astro.fake`, but maybe as a later PR.

sherpa/astro/fake.py
hamogu: Even if it's not a test, it's very useful to a beginner that they can copy and paste the example and it "just works". Sometimes if can't be done, but the risk in those artificially short examples that don't do any of the setup needed is always that it's just a repeat of the function signature and not a working example that someone can take to modify for their own situation. So, I'm absolutely in favor of long examples - even if they are not used as tests.
(In the past, in Sherpa and other packages, I've sometimes looked at the test code to find an actual complete example, but test code is not written for that. It's often littered with pytest parmeterization and fixtures that do not help the beginner.

sherpa/astro/fake.py
hamogu: So maybe the approach is to leave it like this and address that in a follow-up PR? 
One approach might be to have different functions in the user facing API ("fake_pha_with_background_model", "fake_pha_no_bkg", "fake_pha_draw_from_bkg"), another might be to have a `background_method` argument that accepts more complex objects like `sherpa.fake.BackgroundFromModel`, `sherpa.data.NoBackground`, ... The user would then have to instantiate those objects separately, and thus they would have a smaller API that's hopefully easier to understand. Simulating background has become difficult here.

sherpa/astro/tests/test_fake_pha.py
hamogu: Yes, that helps.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: Aha. It is exactly because I've removed a check for file existence in `unpack_arf`/`unpack_rmf` (which would always be run) because it's also a part of the I/O code (but then it can only run in there is actually a working backend).

So there is the question of whether this simplification is worth it (essentially - do we want the second commit)?

sherpa/astro/ui/utils.py
DougBurke: During development I switched flag names, so I think this is an incorrectly-updated comment. Or I always had it not documenting reality.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: OK, understood. Let's keep that change. It just was not obvious to me, so I figured I'll ask you.

sherpa/astro/ui/utils.py
DougBurke: Hmmmm. I need to investigate further. I have in line 9294 `include_bkg_data` set False, then in 9327 it can be set True, but then in line 9340 it's unconditionally set to False, which means the earlier code is pointless. 

So either this extra text and setting is wrong and can be removed, or the earlier stuff is wrong. Or they are noth essentially the same (but you cna see they are not quite the same).

sherpa/astro/fake.py
DougBurke: I've removed this for now.

sherpa/astro/fake.py
DougBurke: I've gone with actual text, rather than comments. How does this work? (and how many typos did I make)

hamogu: That works well.

sherpa/astro/ui/utils.py
DougBurke: Hmm. So I have added tests that check this, and they show that both approaches work. I need to just look into why this is, as from a brief look on Sunday it wasn't obvious that they would.

DougBurke: Doh. They are the same logic.

sherpa/astro/fake.py
DougBurke: The "predicted counts" on line 82 are meant to be the same as the "predicted counts" on line 80.

DougBurke: I've reworked this section (so this first paragraph and then the next one). Whether it's better is hard to say.

sherpa/astro/ui/utils.py
DougBurke: I've removed the second `include_bkg_data` set of changes as they are unnecessary - hence the comment in question has been completely removed. There's also a little-bit if comment clean up (and some code-style changes to use three rather than one-letter variable names (d/m vs pha/mdl).

sherpa/astro/fake.py
```diff
@@ -21,111 +21,180 @@
...
+        then sherpa.utils.poisson_noise is used, otherwise the
+        function must accept a single ndarray, returning a ndarray of
+        the same shape.
+
+    Notes
+    -----
+
+    The model must include the relevant response, and if a background
+    model is required then it must also be included. The simulation
+    requires an array of "predicted counts" (one for each channel).
```
anetasie, line 76: "....requires an array of predicted counts..." - this is confusing, as the simulations generate an array of predicted counts based on the evaluated model as stated in the next sentence. It is not required as input.

sherpa/astro/fake.py
```diff
@@ -21,111 +21,180 @@
...
+    This array is created by evaluating the model and then, if the
+    dataset contains any background datasets and the include_bkg_data
+    flag is set, then the background signal is added, after
+    appropriate scaling between background and source apertures and
+    exposure times.  This "predicted counts" array is then passed to
+    the method argument to create the simulated data.
+
```
anetasie, line 83: I suggest to reword this paragraph:

"The model must include the relevant response, and if a background
    model is required then it must also be included. The model predicted counts array is generated  by evaluating the model and then, if the dataset contains any background datasets and the include_bkg_data
    flag is set, then the background signal is added, after
    appropriate scaling between background and source apertures and
    exposure times.  This "model predicted counts" array is then passed to
    the method argument to create the simulated data with the default method set to 
poisson_noise."

anetasie:
@DougBurke I cannot easily read the code to understand how the background data if present are used. Is the background data used by default when the background is included in the data set with `load_bkg`, or needs to be included in `fake_pha` call? Is the background cound then used with poisson_noise to generate simulated bkg counts?

DougBurke:
Rebased to address recent conflicts, but I haven't addressed the documentation fixes mentioned above.

sherpa/astro/fake.py
DougBurke: Technically I'd argue that the text isn't talking about the input argument here, but I'll see what I can do to clean the text up.

DougBurke:
I've rebased it to fix a conflict (nothing serious) and to rebase it onto #1957 so the tests will pass.

sherpa/astro/fake.py
DougBurke: I've made this change (tweaked very slightly).

DougBurke:
From a question above by @anetasie 

> @DougBurke I cannot easily read the code to understand how the background data if present are used. Is the background data used by default when the background is included in the data set with `load_bkg`, or needs to be included in `fake_pha` call? Is the background cound then used with poisson_noise to generate simulated bkg counts?

So, the documentation (which is what this PR started out to be about but it got out of hand) should tell you enough to let you work this out. As mentioned throughout this PR and comments, it's actually quite tricky since there are several things we can do, and no obvious "this is the one way to do it". Things we could do include (but there are others, I'm sure)

- ignore the background
- just add the existing background data
- resample the background data (i.e. poisson sample the counts, which is not a good idea but I mention it)
- resample given a background model

Now, this is just the background component in the source. There's the question of if you want to resample the background data too (i.e. the load_back/BACKFILE data). For now we don't do this - I think, I haven't looked at the code for a long while! I think the suggestion is that if you want to fake up the background data you have to call fake_pha again for the background and somehow copy it over. 

This is obviously something we should work on, but it is not clear what we want to do or how to go about it. The `fake_pha` call has accreted functionality for the background but I don't htink we've really thought through all the implications.

DougBurke:
I've merged in the current main branch to resolve a conflict. I'd prefer to do this with a rebase as it's more obvious I haven't broken anything but I didn't want to mess up the previous review comments. We can think about a rebase once this passes review.

DougBurke:
Merged in latest changes to pick up #2000 to get back the coverage checks. Really need to rebase before merging ;-)

anetasie:
@DougBurke Thanks for clarifying the background. I agree there are different ways to go about including background in the simulation. I was also able to run some tests by building the branch. This is ready to be approved

hamogu:
@DougBurke Is this ready for final review to make it into the next release or you planning any more changes? I'm sorry if you've just been waiting for me, I may have lost track what the status of this one is.

anetasie (approved):

DougBurke:
@hamogu I don't know what to do with this one.

I believe it fixes a bug in some cases. To be honest I'm not sure how many people would hit the bug (as I forget the details now).

However, I am not particularly happy with how I implemented it 

- it changes things for anyone relying on the low-level API (although this is "relatively" new code)
- is the new API sufficient or useful

So, the question is whether these changes makes sense to you, as the last person to make major changes here.

hamogu (approved):
I agree with @DougBurke that it's not perfect, but perfect is the enemy of good. It's certainly better than the mess I left with the backgrounds. And, as @DougBurke says, in some cases it's not clear what we actually want.
That said, I ike the new API better, it's cleaner and easier (even if not perfect) and I'm not concerned about the changes in the OO layer. It's rarely used for these cases and we just added it not long ago. So, on balance I think we should take this PR and improve from there if need be.

sherpa/astro/fake.py
```diff
@@ -22,38 +23,31 @@
...
-    distribution that might already have very few counts in the first place.
-
-    The backgrounds itself are not changed by this function. To simulate
-    backgrounds as well as the source spectrum, call this function on the
-    source PHA dataset and the background PHA dataset(s) independently.
+    This function replaces the counts in a PHA dataset with simulated
+    counts drawn from a model with Poisson noise. For the simulations,
+    all the details already set up in the PHA dataset will be used,
+    including the exposure time, one or more ARFs and RMFs, area and
+    background scalings, grouping, and data quality arrys.
```
wmclaugh, line 45:
```suggestion
    background scalings, grouping, and data quality arrays.
```

sherpa/astro/fake.py
```diff
@@ -62,76 +56,158 @@ def fake_pha(data, model,
...
+
+    if pileup_model is not None:
+        warn("pileup_model is no-longer used, the model "
+             "must contain any needed pileup model",
+             category=DeprecationWarning)
+
+    if add_bkgs is not None:
+        warn("add_bkgs is no-longer used, as the model "
+             "either contains a backgorund term or the "
+             "backgound datasets are used",
```
wmclaugh, line 195:
```suggestion
             "background datasets are used",
```

sherpa/astro/fake.py
```diff
@@ -62,76 +56,158 @@ def fake_pha(data, model,
...
+             category=DeprecationWarning)
+
+    if pileup_model is not None:
+        warn("pileup_model is no-longer used, the model "
+             "must contain any needed pileup model",
+             category=DeprecationWarning)
+
+    if add_bkgs is not None:
+        warn("add_bkgs is no-longer used, as the model "
+             "either contains a backgorund term or the "
```
wmclaugh, line 194:
```suggestion
             "either contains a background term or the "
```

PR #1687, opened 2023-02-06
GH Actions: Update xspec to min 12.12.0g with conda-forge

Marie-Terrell:
I've added new packages to the xspec channel at label "test". These use "conda-forge" and are ONLY there for testing purposes (not for deployment):

- cfitsio: 4.0.0, 4.1.0 (macOS ONLY since we already have one for Linux), 4.2.0 
- xspec-modelsonly: 12.12.0g (macOS not uploaded due to space), 12.12.1c (macOS not uploaded due to space), and 12.13.0c

These packages include the movement of the CCfits header files into the CCfits directory in the include area. Only the 12.13.0c packages have been tested to the side thus far (and not exhaustively). So, I have this as a draft at the moment.

I can upload the additional macOS packages noted above, but I would need to make a bit more space on that channel.

With this PR, I'm bumping all of the xspec tests to use these new version (12.13.0c for macOS and various versions for Linux) with the minimum of 12.12.0g. This is mean to coincide with #1609.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -33,8 +33,11 @@ conda update --yes conda
 # https://github.com/sherpa/sherpa/pull/794#issuecomment-616570995 )
 # the XSPEC-related channels are only added if needed
 #
+if [ -n "${XSPECVER}" ]; then
+ conda config --add channels conda-forge
+ conda config --add channels ${xspec_channel}
+fi
 conda config --add channels ${sherpa_channel}
-if [ -n "${XSPECVER}" ]; then conda config --add channels ${xspec_channel}; fi
 
 # Figure out requested dependencies
 if [ -n "${MATPLOTLIBVER}" ]; then MATPLOTLIB="matplotlib=${MATPLOTLIBVER}"; fi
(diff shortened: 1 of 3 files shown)


.github/scripts/setup_conda.sh
```diff
@@ -33,8 +33,11 @@ conda update --yes conda
 # https://github.com/sherpa/sherpa/pull/794#issuecomment-616570995 )
 # the XSPEC-related channels are only added if needed
 #
+if [ -n "${XSPECVER}" ]; then
+ conda config --add channels conda-forge
```
DougBurke, line 37: So, presumably there's something in xspec-modelsonly that uses conda-forge ? I'm asking mainly to make it look like I've thought about this ;-)

Marie-Terrell: Yes. The default channel does not have a new enough version of gfortran to build Xspec for macOS (as newer versions of xspec requires fgsl which relies on C_LOC support in gfortran ~5.x). So, we need to jump to conda-forge for at least macOS. 
Since we made the jump for both in CIAO (and based on it not seeming to be an issue when asked a few PR meetings ago), I made the jump to conda-forge for both Linux and macOS. 
It actually makes things easier on CM because we can use the same xspec package for CIAO and Standalone Sherpa testing.

DougBurke: kewlio.

As the kids don't say. But should.

PR #1698, opened 2023-02-15
Update versioneer to 0.28

### LINKED ISSUE #1677: we need to look at versioning of Sherpa
During a `pip install -e . --verbose` run for a PR I'm working on I noticed the following warning, which suggests we need to look at our current versioning strategy. I have been trying to keep well away from knowing anything this...

```
    Adding sherpa ciao4.15-148.gd2a9f3e.dirty to easy-install.pth file
    /tmp/pip-build-env-gqwiizkz/overlay/lib/python3.10/site-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: ciao4.15-148.gd2a9f3e.dirty is an invalid version and will not be supported in a future release
      warnings.warn(
```

DougBurke:
# Summary

Update the version of versioneer included in Sherpa from 0.21 to 0.28. Fix #1677.

# Details

Version 0.21 was released 13 Oct 2021 and 0.28 was released 27 Oct 2022. The main changes appear to be, in no-particular order

- dropping Python 3.6
- adding support for Python 3.10 and 3.11 (although it's not obvious we trigger any python differences)
- you can now install versioneer as part of your dependencies rather than using a vendored version (as we do; this is still experimental and I see no reason to switch over at this time)
- PEP 660 / setuptools editable support (my eyes glaze over with this issue, so I do not understand it, but I use editable installs when testing, so yay I guess, and may be related to #1677)
- some metadata changes which hopefully aren't a problem for us

The first time I tried to build with this I had a failure (a missing `wcs.h` file`), which could indicate some metadata change could be problematic, but I have not been able to replicate this failure.

I don't think the existing code was "bad" but we should keep on top of changes like this given the changes we know are happening with the Python build/configuration infrastructure. It also looks like this has addressed #1677, since before this the output of `pip install -e . --verbose` contained the following warnings

```
 /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/setuptools/dist.py:507: UserWarning: The version specified ('ciao4.15+146.g4ee44439') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: ciao4.15-146.g4ee44439 is an invalid version and will not be supported in a future release
  /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/setuptools/dist.py:507: UserWarning: The version specified ('ciao4.15+146.g4ee44439') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: ciao4.15-146.g4ee44439 is an invalid version and will not be supported in a future release
    /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/setuptools/dist.py:507: UserWarning: The version specified ('ciao4.15+146.g4ee44439') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
    /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: ciao4.15-146.g4ee44439 is an invalid version and will not be supported in a future release
    /tmp/pip-build-env-r4xq4uv6/overlay/lib/python3.10/site-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: ciao4.15-146.g4ee44439 is an invalid version and will not be supported in a future release
```

which do not occur with this PR.

However, the versioning has changed and this is not something I often care about, so some care is needed.

Before we get

```
% python -c 'import sherpa; print(sherpa.__version__)'
ciao4.15+146.g4ee44439
```

and now

```
% python -c 'import sherpa; print(sherpa.__version__)'
4.15.0+150.g0896fd6c
```

which strikes me as being probably-problematic.


DougBurke:
@Marie-Terrell and @harlantc - this strikes me as a "morally correct" change, but I suspect that the change in the version string could have unintended consequences.

An alternative is to use a different package for setting up the version from git (something like setuptools_scm) but I admit to having no knowledge/experience about them.

Marie-Terrell (approved):
Ran through a few combinations of Python versions and setup.py/pip/sdist installs. Looks good to me.

PR #1705, opened 2023-02-22
Codecov actions update and coverage report bump

harlantc:
# Summary
Updates the codecov action to v3 to address the issue with mismatched uploader checksums and bump the report notification threshold from 8 to 12.

# Details
This is a straightforward update to the CI codecov action. codecov action v2 seemed to have a few reports of uploader checksum mismatch issues. Going to v3 seems to have resolved this (unless there's something else on codecov's side that's not performing nominally)

This also bumps the number of reports codecov waits for before sending the report summary. Before, I set it to 8, but still noticed how inaccurate reports can be before it updates completely. Now it matches the total reports sent, 12, so it only reports when all the tests are in, to be a little less annoying. I intended to update this a while ago, but this is the first sensible opportunity.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -2,7 +2,7 @@ codecov:
   require_ci_to_pass: yes
   #this handles notifications for bots (slack, gitter, etc)
   notify:    
-    after_n_builds: 8 #wait for 8 of the builds to finish and send their reports to send a notice, default is 2
+    after_n_builds: 12 #wait for 12 of the builds to finish and send their reports to send a notice, default is 2
     wait_for_ci: yes
   
 coverage:
@@ -31,7 +31,7 @@ comment:
   layout: "reach,diff,flags,files,footer"
   behavior: default
   require_changes: no
-  after_n_builds: 8 # wait for 8 builds to finish and send their reports to make a comment, default 2
+  after_n_builds: 12 # wait for 12 builds to finish and send their reports to make a comment, default 2
   
 # Fix issue with 'work/sherpa/sherpa' being considered a seperate directory 
 fixes:
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -93,7 +93,7 @@ jobs:
 
     steps:
     - name: Checkout Code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3.1.0
       with:
         submodules: 'True'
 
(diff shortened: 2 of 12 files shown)


DougBurke:
Ta for this.

harlantc:
It seems there might be a different issue on codecov's side.

looking at the reports, seeing:
```
[2023-02-22T16:50:03.998Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

[2023-02-22T16:50:03.999Z] ['verbose'] The error stack is: Error: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```
This is apparently related to a GitHub API ratelimit (see blogpost below). It might also be due to not having an upload token, even though codecov says you don't need one. It's related to this issue https://github.com/codecov/codecov-action/issues/837. This leaves us with a few choices
- lower the report threshold. Setting it to 12 (which is actually wrong, it's 13, as @Marie-Terrell helpfully pointed out. I know how to count I swear) would mean we don't get codecov comments until codecov gets that many reports.
- use `fail_ci_if_error: true` as it's not the default... for reasons. this will result in a lot of failed CI jobs I think.
- add a codecov token
- find someway to retry the upload until it works

I think adding the codecov token and retrying jobs might be the first thing to try, at least according to this blogpost https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954

DougBurke:
I think this is now just a "bump the number" PR (thanks to #1706 stealing your thunder), isn't it?

harlantc:
@DougBurke not quite. The underlying issue is that Codecov is getting timed out on GitHub API calls, so if we bump the number of reports codecov expects, we're going to have issues seeing regular reports, as most times at least one reports fails to send, if not more. 

Going to come back around and take another look at this (and partially hope by the time I do, codecov will have figured some of it out).

DougBurke:
So, codecov is now at version 4 and this has marked as a breaking change:

> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token)

See https://github.com/marketplace/actions/codecov

harlantc:
Yep, thanks for checking.

PR #1707, opened 2023-02-22
DS9 use individual names

DougBurke:
# Summary

Allow the user to change the name used to communicate with DS9 via the sherpa.image.set_template function. This supports multiple DS9 instances within a single session but is primarily meant to support multiple Sherpa sessions, each with a separate DS9 instance.

# Details

The communication with DS9 is based on the template value used to create the `DS9Win` object and once set, the same name is used for all communication. The way that DS9 is initialized is (even after the clean up in #2459 and related) still set at module import time, so there is currently no way for a user to change it (without doing some low-level changes which this PR automates).

So, we introduce the `sherpa.image.set_template` and `sherpa.image.get_template` routines to set or retrieve the current name. The use of "template" here is to match the existing `DS9Win` nomenclature, but this could be changed as `set_template` is not the most-obvious name (but `set_name` is I think too generic and my feeling is that this is not likely to be a common operation). We technically don't need to have the `sherpa.image` variants, as we could just get users to call `sherpa.image.backend.set_template` but this feels a bit too much for the intended use case.

These two routines are then replicated in the backend code since the value is used creating the `sherpa.image.backend.image` instance. One tweak is that the validation in `set_template` is done at the `sherpa.image` layer rather than being in the backend code (with only one viable backend the benefit is not huge). 

Note that `set_template` has to re-generate the `image` value due to the way that the `DS9Win` code works (we could change the `imager.template` field but that then invalidates some of the invariants that `DS9Win` uses so the re-create option seems cleanest.

I had originally hoped to use this to allow parallel runs of the testing code when DS9 is present. However, testing with this approach - e.g. by making a `set_template` call with a unique name for each test - turns out to then cause other issues - e.g. warning messages from Xvfb (via `pytest-xvfb`) about clobbering existing sessions and some pipes not being closed on exit - so I have decided not to introduce that logic here.

Given that I think this is a specialized use case I'm happy with making the user

```
from sherpa.image import set_template
```

rather than having it be part of the `Session` class (i.e. available from the `ui` module).

# Example

```
% XPA_METHOD=local python
Python 3.14.4 | packaged by conda-forge | (main, Apr  8 2026, 02:33:53) [Clang 20.1.8 ] on darwin
>>> from sherpa.image import backend, get_template, set_template
>>> ui.load_image(1, "sherpa-test-data/sherpatest/img.fits")
>>> ui.image_data()
```

Let's check the current "template name"

```
>>> get_template()
'sherpa'
```

We can check that this is being used with the "XPA INFO" command:

```
>>> print(ui.image_xpaget("xpa info"))
XPA_VERSION:	2.1.20
XPA_CLASS:	DS9
XPA_NAME:	sherpa
XPA_METHOD:	/var/folders/kz/nt_fs_wj2yzg99cgs1hfz40m0000gr/T//DS9_sherpa.40079
```

Let's change the colormap of this instance:

```
>>> ui.image_xpaset("cmap heat")
```

Now switch template:

```
>>> set_template("foo")
```

If we try "XPA INFO" it fails

```
>>> print(ui.image_xpaget("xpa info"))
...
  File "/Users/burked/code/sherpa/sherpa-1/sherpa/image/ds9_backend.py", line 221, in xpaget
    raise DS9Err('open')
sherpa.utils.err.DS9Err: Imager not open
```

Let's create an image

```
>>> ui.image_data()
>>> print(backend.xpaget("xpa info"))
XPA_VERSION:	2.1.20
XPA_CLASS:	DS9
XPA_NAME:	foo
XPA_METHOD:	/var/folders/kz/nt_fs_wj2yzg99cgs1hfz40m0000gr/T//DS9_foo.40171
```

So, at this point we have two DS9 windows open.

```
>>> ui.image_xpaget("cmap")
'grey\n'
```

If we switch back we get to talk to the original DS9:

```
>>> set_template("sherpa")
>>> ui.image_xpaget("cmap")
'heat\n'
```
 
So, you can switch between the two (or add more) but it is a bit awkward and you have to remember which one you are talking to. I think this is an okay caveat.

The main interest I think is that I can run multiple sherpa sessions from different terminals and they no-longer try to talk to the same DS9 by choosing a "unique" name - e.g.

```
>>> import os
>>> set_template(f"sherpa-{os.getpid()}")
```

### DIFF
--- a/sherpa/image/__init__.py
+++ b/sherpa/image/__init__.py
@@ -23,6 +23,10 @@
 At present the only supported application is DS9 [DS9]_, which is
 connected to via XPA [XPA]_.
 
+Communication is with a single DS9 instance. If multiple instances
+are needed then the ``set_template`` routine can be used to select
+a new template name for the next image call.
+
 References
 ----------
 
@@ -32,8 +36,13 @@ References
 
 """
 
+import string
+
 import numpy
+
 from sherpa.utils import NoNewAttributesAfterInit, bool_cast, display_fields
+from sherpa.utils.err import ArgumentTypeErr
+
 
 import logging
 warning = logging.getLogger(__name__).warning
@@ -54,6 +63,53 @@ except Exception as e:
 __all__ = ('Image', 'DataImage', 'ModelImage', 'RatioImage',
            'ResidImage', 'PSFImage', 'PSFKernelImage', 'SourceImage',
            'ComponentModelImage', 'ComponentSourceImage')
+
+
+def get_template() -> str:
+    """Return the template name used to talk to DS9.
+
+    Returns
+    ----------
+    template : str
+        The name used with for the DS9 instance.
+
+    See Also
+    --------
+    set_template
+
+    """
+
+    return backend.get_template()
+
+
+def set_template(template: str) -> None:
+    """Change the template name used by DS9.
+
+    Calling this will cause a new DS9 instance to be created
+    the next time an image is displayed.
+
+    Parameters
+    ----------
+    template : str
+       The name used with the DS9 instance. It must not contain
+       whitespace or be empty.
+
+    See Also
+    --------
+    get_template
+
+    """
+
+    if not isinstance(template, str):
+        raise ArgumentTypeErr("badarg", "template", "a string")
+
+    if template == "":
+        raise ArgumentTypeErr("badarg", "template", "not empty")
+
+    if any(c in template for c in string.whitespace):
+        raise ArgumentTypeErr("badarg", "template", "a string without whitespace")
+
+    backend.set_template(template)
 
 
 class Image(NoNewAttributesAfterInit):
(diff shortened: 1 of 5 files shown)


wmclaugh:
Would it be better to keep the default ds9 xpa address and allow users/tests to override the default xpa name (ie. option to use PID, UUID, etc..)?  Doing so would simplify external interaction with ds9 (particularly in a CIAO environment).

DougBurke:
The current design makes it hard to switch the name, but I think I see a possible means to doing it. I just need time to think about it.

I don't understand the concern about the interaction with DS9 in a CIAO environment, as I can't think of anything that makes that special.

DougBurke:
@wmclaugh - here's one way to make the name controllable. Not sure if I like it yet - need to use it a bit. It also returns the default to a boring name like "sherpa" ...

I am likely to explore using SAMP to communicate with DS9, which would mean we could drop XPA (at least for the standalone sherpa testing). This "name" field is likely to still be relevant there too, so it is not wasted work.

DougBurke:
Rebased as #1697 has been merged. Ta @wmclaugh

DougBurke:
Rebased because #1969 was merged (which was the original impetus on working on this) and because I've cleaned up the commit history.

PR #1708, opened 2023-02-24
Ensure string id can be used for the background id in the DataPHA notice/ignore call

DougBurke:
# Summary

Allow a string identifier to be used for the bkg_id argument to the DataPHA.notice (and ignore) method, fixing #1709 (it does not affect users using the ui layer).

# Details

This was noticed during the grouping/utils handling work. This is primarily a minor clean up but isn't huge. I then noticed #1709 and so added that fix in (and so the summary is all about the bugfix, but most of the code is clean up).

The #1709 fix is easy - we need to decide if bkg_id is sent a single is or a sequence of ids. Unfortunately the code just used numpy.iterable to decide, not recognizing that  a string is a valid scalar argument here (and that iterable returns true for a string). So the easy fix (as used elsewhere in sherpa) is to add in a "not a string instance" check.

There is also a change to an error message, but it's one that is hard to trigger so I've not listed it in the summary. Before this PR we have

```
>>> DataPHA("x", None, None).response_ids = [1]
  File "/home/dburke/miniconda3/envs/ciao415-py310/lib/python3.10/site-packages/sherpa/utils/__init__.py", line 171, in __setattr__
  File "/home/dburke/miniconda3/envs/ciao415-py310/lib/python3.10/site-packages/sherpa/astro/data.py", line 1737, in _set_response_ids
    raise DataErr('badids', str(id), 'response', str(keys))
sherpa.utils.err.DataErr: 1 is not a valid response id in dict_keys([])
```

whilst we now get

```
>>> DataPHA("x", None, None).response_ids = [1]
  File "/home/dburke/sherpa/sherpa-master/sherpa/utils/__init__.py", line 171, in __setattr__
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 1746, in _set_response_ids
    raise DataErr('badids', str(resp_id), 'response', str(keys))
sherpa.utils.err.DataErr: 1 is not a valid response id in []
```

which has just lost the `dict_keys()` wrapper around the list.

# Notes

I was going to rework how the response_ids/background_id setting is done - as we can just take if from the keys of the _response/_backgrounds dictionaries. However, there's one use case (which I added a test for) that my new code would have handled [1], and it wasn't a huge improvement in the code.

[1] the code as is allows us to do the following

   pha.set_background(..., id="up")
   pha.set_background(..., id="down")
   pha.background_ids = ["up"]
   pha.background_ids = ["down"]

I'm no sure how useful it is, but as I mention I don't think that removing this capability actually helps save much code complexity.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4707,7 +4707,12 @@ will be removed.
         #
         filter_background_only = False
         if bkg_id is not None:
-            if numpy.iterable(bkg_id):
+            # As bkg_id can be
+            #   - int or str
+            #   - iterable of int or str
+            # it's a bit awkward to identify what is meant.
+            #
+            if not isinstance(bkg_id, str) and numpy.iterable(bkg_id):
                 bkg_ids = bkg_id
             else:
                 bkg_ids = [bkg_id]
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1677,8 +1677,8 @@ class DataPHA(Data1D):
         else:
             raise DataErr('bad', 'quantity', val)
 
-        for id in self.background_ids:
-            bkg = self.get_background(id)
+        for bkg_id in self.background_ids:
+            bkg = self.get_background(bkg_id)
             if bkg.get_response() != (None, None) or \
                (bkg.bin_lo is not None and bkg.bin_hi is not None):
                 bkg.units = units
@@ -1693,8 +1693,8 @@ class DataPHA(Data1D):
 
     def _set_rate(self, val):
         self._rate = bool_cast(val)
-        for id in self.background_ids:
-            self.get_background(id).rate = val
+        for bkg_id in self.background_ids:
+            self.get_background(bkg_id).rate = val
 
     rate = property(_get_rate, _set_rate,
                     doc="""Is the Y axis displayed as a rate when plotting data?
@@ -1723,8 +1723,8 @@ a rate.""")
             raise DataErr("bad", "plot_fac setting", val)
 
         self._plot_fac = ival
-        for id in self.background_ids:
-            self.get_background(id).plot_fac = ival
+        for bkg_id in self.background_ids:
+            self.get_background(bkg_id).plot_fac = ival
 
     plot_fac = property(_get_plot_fac, _set_plot_fac,
                         doc="""How the X axis is used to create the Y axis when plotting data.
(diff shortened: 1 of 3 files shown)


sherpa/astro/data.py
```diff
@@ -2721,6 +2723,20 @@ def _channel_to_wavelength(self, val, group=True, response_id=None):
...
+
+        Parameters
+        ----------
+        id : int, str, or None
+            The background identifier, or None. If None then the
+            default_background_id value will be used.
+
+        Returns
+        -------
+        bkg_id : int
```
wmclaugh, line 2736: What is the return value (bkg_id) If id is passed in as a string?

DougBurke: Yeah. This should be "`int or str`". The existing documentation is a bit unclear on this - I had thought that it could only be an int for historical reasons but it turns out not to be so (technically it could be "any type that we can use as a key for a dict" but I think it makes sense to force some structure on this).

So I should review the docs to look at this as part of this.

DougBurke: I've made some changes (in general it looks like we do say "int or str" in most places).

DougBurke:
Rebased to address conflict (and in doing so I merged the tests into a single commit).

DougBurke:
[@hamogu - I'm just putting this on your plate for after-proposal fun and profit]

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -4205,3 +4202,162 @@ def test_pha_ungroup_background_remove(caplog):
...
+    """
+
+    pha = DataPHA("src", [1, 2, 3], [1, 1, 1])
+    b1 = DataPHA("b1", [1, 2, 3], [1, 1, 1])
+    b2 = DataPHA("b2", [1, 2, 3], [1, 1, 1])
+
+    assert len(pha.background_ids) == 0
+
+    pha.set_background(b1)
+    assert pha.background_ids == pytest.approx([1])
```
hamogu, line 4220: I always find it funny to use approz on integers, but I know you have your reasons.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -4205,3 +4202,162 @@ def test_pha_ungroup_background_remove(caplog):
...
+    assert pha.background_ids == pytest.approx(["up", 1])
+
+    # Remove one element.
+    #
+    pha.background_ids = [1]
+    assert pha.background_ids == pytest.approx([1])
+
+    # We can do the following, which is technically a no-op but may
+    # change some internal state. This also tests using an
+    # iterable-that-is-not-a-list.
```
hamogu, line 4246: I would argue that iterable-that-is-not-a-list is not part of the intended API. If it works, great, but I don't think we have to guarantee that for the future.

hamogu (approved):

sherpa/astro/tests/test_astro_data2.py
DougBurke: Can we just compare to `[1]` - I forget (in this case the return value should be a boring Python list and not a ndarray)?

sherpa/astro/tests/test_astro_data2.py
DougBurke: This at least gives us a canary if it does change. Depending on my mood I get annoyed with or approve wholeheartedly of the the Python duck-typing "hey, you can use this as it's an iterable" approach.

sherpa/astro/data.py
```diff
@@ -1740,16 +1740,19 @@ def _set_response_ids(self, ids):
...
 
         self._response_ids = list(ids)
 
     response_ids = property(_get_response_ids, _set_response_ids,
-                            doc=('IDs of defined instrument responses ' +
-                                 '(ARF/RMF pairs)'))
+                            doc="""IDs of defined instrument responses (ARF/RMF pairs).
+
+If set, the identifiers must already exist, and any other responses
+will be removed. The identiers can be integers or strings.
```
wmclaugh, line 1754:
```suggestion
will be removed. The identifiers can be integers or strings.
```

PR #1711, opened 2023-02-27
plot: add linewidth option

DougBurke:
# Summary

Add the linewidth option for data and model plots.

# Details

This is easy to add, although it is "matplotlib" specific (but I'd imagine any plotting backend would let you set this).

This was a request from a recent helpdesk user, and it has been sitting around for ages (this has been taken from #938). It will conflict with the backend rework but I don't want to hold back useful changes for ever.


DougBurke:
@hamogu - in https://github.com/sherpa/sherpa/pull/938/commits/503473305d0a409ab613b8dfb272fa9feac31ad1#r499193215 you suggested using `**kwargs` rather than a specific list of options. Although we could discuss that here, I think it's better off as part of the plot backend work, but YMMV.

hamogu (approved):

PR #1712, opened 2023-02-27
WIP plot multiple datasets at once

DougBurke:
plot_data([1, 2])
    plot_model([1, 2], overplot=True)
    plot_fit([1, 2], xlog=True, color=["k", "g"])

This is an experiment.

It allows users to give a sequence of ids, and to allow the options to be given as a sequence.

Missing documentation, tests, coverage of other plot commands, ...


DougBurke:
I've added a `MultuiPlot` class in 

- #1679  

How does that compare/contrast to this?

DougBurke:
I will note that we did just get a question where this could have been useful

- #2119

DougBurke:
Rebased because a lot has changed.

DougBurke:
Rebased to address a conflict and change the implementation to use #2125

DougBurke:
Rebased onto 

- #2126 

as it's likely to conflict, and it makes sense to be able to ensure it all works together. Assuming it did work together....

sherpa/ui/utils.py
```diff
@@ -12112,6 +12144,70 @@ def get_draws(self,
     # Plot object access
     #
 
+    def _get_plot_objects(self,
+                          ids: IdType | IdTypes,
+                          getfunc: Callable[..., Plot | HistogramPlot],
```
hamogu, line 12150: Where is the symbol "HistogramPlot" imported?

sherpa/ui/utils.py
```diff
@@ -13919,32 +14022,99 @@ def _multi_plot(self,
         get = getattr(self, f"_get_{plotmeth}type")
         check = getattr(self, f"_check_{plotmeth}type")
 
+        # Store the per plot-area positinal arguments in plots and
```
hamogu:
```suggestion
        # Store the per plot-area positional arguments in plots and
```

sherpa/ui/utils.py
```diff
@@ -13919,32 +14022,99 @@ def _multi_plot(self,
...
+            # The important check here really is that the annotation
+            # is set, but it makes sense to check it is the expected
+            # value, in case something changes with the code paths.
+            #
+            if list(pars.keys())[0] == "id" and \
+               pars["id"].annotation == "IdType | None":
+                allow_multiple_ids = True
+            else:
+                allow_multiple_ids = False
+
```
hamogu, line 14058:
```suggestion
            if list(pars.keys())[0] == "id" and \
                pars["id"].annotation == "IdType | None":
                    allow_multiple_ids = True
            else:
                allow_multiple_ids = False

```

sherpa/ui/utils.py
```diff
@@ -13919,32 +14022,99 @@ def _multi_plot(self,
...
+            # identifiers such as
+            #
+            #    plot("data", [1, 2], "model", [1, 2])
+            #
+            # since the "get_xxx_plot" calls do not support multiple
+            # identifiers. This logic is dependent on the code - so
+            # both how arguments are typed and what particular
+            # routines do not support multiple identifiers - so these
+            # heuristics may need updating in the future.
+            #
```
hamogu, line 14107: It's code like this that makes me not like the UI plotting machinery. I know we have it, and I know we need ot continue to support it, but it makes me uncomfortable because it seems so brittle. This routine relies (not only on the line where I'm attaching this comment, but in general) on inspecting annotations and signatures, on constructing functions names from strings and than grabbing them.

I understand why it's done this way, and how users might benefit, but it's complicated.

We should get this PR in, but I would like an open discussion on where the want to go with the plotting interface. Do we want ot continue to add features to the UI interface? Can we do that without increasing code complexity too much? 

I'm not going to disapprove this particular PR, but I do wonder if out time might be spend better into teaching users how to make use of the fact that Sherpa is Python and have them write a simple for loop themselves if they want to stack up multiple plots as opposed to this approach. It's not like `plot("data", [1, 2, 3], color="gray", alpha=[[1, 0.5, 0.2]])` is intuitive either (I would have never thought to try that unless I copy and paste that exact line from the CIAO threads or the docs.)

As more and more of our users get Python-savy, I wonder how much features like this are actually used by users. For example, I don't use them, because I can never remember how the syntax is, I write loops in Python. The one other self-taught Sherpa users that I'm observing "in the wild" (a collaborator who occasionally shares notebooks with me) never uses Sherpa plot commands at all; he des `get_plot_data` and then sets up all plots him self in matplotlib.

So, I'm wondering if it is really the best investment of our time to make the plotting machinary more complex.

sherpa/ui/utils.py
```diff
@@ -13919,32 +14022,99 @@ def _multi_plot(self,
...
+               is_iterable_not_str(getargs[0]):
+                def getid(id, recalc=True):
+                    return getfunc(id, *getargs[1:], recalc=True)
+
+                plotobj = self._get_plot_objects(getargs[0], getid,
+                                                 recalc=True)
+                plots.append(copy.deepcopy(plotobj))
+
+            else:
+                plots.append(copy.deepcopy(getfunc(*getargs)))
```
hamogu, line 14118: I have to admit that I don't fully understand this logic, but apparently it generates plots, so it works.

sherpa/ui/utils.py
```diff
@@ -14408,6 +14596,11 @@ def plot(self,
 
         >>> plot("fit", "nucleus", "fit", "jet")
 
+        A single plot is diplayed with the fit plot for dataset "jet"
+        overlain on that for dataset "nucleus":
+
+        >>> plot("fit", ids=["nucleus", "jet"])
```
hamogu:
```suggestion
        A single plot is displayed with the fit plot for dataset "jet"
        overlain on that for dataset "nucleus":

        >>> plot("fit", ids=["nucleus", "jet"])
```

sherpa/ui/utils.py
```diff
@@ -14454,26 +14647,50 @@ def plot(self,
...
         overplot those from dataset 2:
 
-        >>> plot("data", "resid", cols=1, color="black")
+        >>> plot("data", "resid", color="black")
         >>> plot("data", 2, "resid", 2, overplot=True, color="black", alpha=0.5)
 
+        Draw the data and residuals for the default dataset and then
+        overplot those from dataset 2:
+
+        >>> plot("data", "resid", ids=[1, 2], color="black", alpha=[1, 0.5])
```
hamogu:
```suggestion
        An alternative way to generate the same output is to call ``plot`` once
        and pass in both the default it (1 in this case) and the id for the
        second dataset:

        >>> plot("data", "resid", ids=[1, 2], color="black", alpha=[1, 0.5])
```

sherpa/ui/utils.py
```diff
@@ -14454,26 +14647,50 @@ def plot(self,
...
+
+        Draw the data and residual plots for datasets 1 and 2 with
+        per-plot options:
```
hamogu:
```suggestion
        Draw the data and residual plots for datasets 1 and 2 with
        per-plot options. Here, all data will be shown slight transparent
        (the same value of ``alpha=0.5`` is applied to both datasets), but with
        different line styles: The first dataset will be shown with a dashed line
        and the second with a dash-dotted line as specified in the list of linestyle
        values. The second plot with the residuals will be in orange for both
        datasets, since ``opts2`` only contains a single value and not a list of 
        parameters.
```

sherpa/ui/utils.py
```diff
@@ -14644,6 +14878,11 @@ def plot_model(self,
         >>> plot_model(1)
         >>> plot_model(2, overplot=True)
 
+        Overlay the two datasets on the same plot, both in black with
+        with data set 2 usniog a dashed line style:
```
hamogu:
```suggestion
        with data set 2 using a dashed line style:
```

sherpa/ui/utils.py
```diff
@@ -15237,9 +15520,14 @@ def plot_chisqr(self,
         This function displays the square of the residuals (data -
         model) divided by the error, for a data set.
 
+        .. versionchanged:: 4.17.0
```
hamogu:
```suggestion
        .. versionchanged:: 4.18.0
```

hamogu (changes requested):

hamogu:
@anetasie I checked out the code, do you want to try this out in a sherpa session?

sherpa/ui/utils.py
DougBurke: Doh...

sherpa/ui/utils.py
DougBurke: I don't understand this suggestion.

sherpa/ui/utils.py
DougBurke: I think there's a balance to be had. We definitely have user tickets where we need "just a little bit more" in the existing plot code to support this (e.g. this PR). However, we do not want to support all possible plots, so we need to provide "nice" access to the data that makes up the plots. We also have users who just don't want to learn Python, but that's getting well outside the scope of this PR!

At the moment I think we are now about at the tipping point of "it's not worth adding new functionality" [*], but that's probably because I haven't been using Sherpa for real work recently (so haven't come across annoying things when I just want to quickly plot data)!

[*] I think the "datastack" support could likely be improved but that's likely a bigger thing than just the plot interface.

sherpa/ui/utils.py
DougBurke: So this part is definitely annoying to support and this was the best I could come up with.

sherpa/ui/utils.py
DougBurke: So, the way we create `ahelp` versions of these files is to grab these examples and split them by "text followed by code" is a single example - e.g. https://cxc.cfa.harvard.edu/sherpa/ahelp/plot.html#examples . This means that ahelp users don't quite get the same context (e.g. as written in your suggestion). I'll try to rewrite this

sherpa/ui/utils.py
DougBurke: I've explicitly imported it now. I'm assuming the `from __future__ import annotations` was why it was not a problem before (or it was thanks to importing `sherpa.all` or something similar).

sherpa/ui/utils.py
hamogu: I clearly had some thought when I opened this, but then I didn't edit the code and now I can't recall it. So it can't have been important. resolving conversation.

hamogu (approved):

PR #1667, opened 2022-12-23
Utils cleanup after grouping

DougBurke:
Breaking up into

- #1715
- #1716
- #1717 (it has grown from the version here)
- #1720

# Summary

Some of the clean-up code from #1638

# Details

Todo

PR #1717, opened 2023-03-02
Ensure the background is a DataPHA object

### LINKED ISSUE #1718: check a background is a PHA object
Looking at #1717 I noticed that we don't check a background is a DataPHA object. The following is one example - where fortunately we fail immediately (ie do not leave things in an odd state) but the error-message is not friendly.

```
>>> dataspace1d(1, 100, 1, dstype=DataPHA)
>>> get_data().background_ids
[]
>>> dataspace1d(1, 100, 1, dstype=DataPHA, bkg_id="up")
>>> get_data().background_ids
['up']
>>> dataspace1d(1, 100, 1, bkg_id="down")
  File "<string>", line 1, in dataspace1d
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/ui/utils.py", line 820, in dataspace1d
    self._get_pha_data(id).set_background(dstype('bkg_dataspace1d',
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 2806, in set_background
    if bkg.grouping is None:
AttributeError: 'Data1DInt' object has no attribute 'grouping'
```

OOh. The one I thought might just fail errors out in the same way (so should make it easier to handle):

```
>>> d = DataPHA("x", [1, 2], [1, 2])
>>> d.set_background(Data1D("y", [1, 2], [1, 2]))
  File "/home/dburke/sherpa/sherpa-master/sherpa/astro/data.py", line 2806, in set_background
    if bkg.grouping is None:
AttributeError: 'Data1D' object has no attribute 'grouping'
```

Should we also enforce that the background must match the source - e.g. this is currently allowed:

```
>>> dataspace1d(1, 100, 1, dstype=DataPHA)
>>> dataspace1d(1, 200, 1, dstype=DataPHA, bkg_id="down")
```

DougBurke:
# Summary

Add an explicit check that the background data is a PHA object, to improve error messages. There are internal changes to how the PHA data is accessed in the utils code. Fixes #1718 

# Details

In working on the cleanup for this PR I noted #1718 so added a check that the background is a PHA object, and that it has the same channels as the source dataset (if a user wants to have a different set of channels for the background then they have to treat the datasets as separate - i.e. do  not use the background capability - since the background code implicitly assumes the channels match). Given the recent data validation work added in #1477 I am conscious in not trying to support too many different options, so the rule is that if you add a background then the source dataset must have a channel array set up (we could avoid this, but it would significantly complicate things for little real-world gain).

The main driver was to rework code that looks something like

    d = self._get_pha_data(id)
    if bkg_id is not None:
        d = self.get_bkg(id, bkg_id)

(or the more-recent version where there's a if .. else .. form) by moving the logic for this into _get_pha_data.

There is still repeated code where we don't require a PHA dataset when bkg_id is None and this could do with a similar change. I have tried to move these examples to the same form

    if bkg_id is None:
        d = self.get_data(id)
    else:
        d = self.get_bkg(id, bkg_id)

so it is easier to find if we ever come up with a replacement. Actually, we've now decided to do this (`_get_data_or_bkg`) as part of this PR.

This was taken from https://github.com/sherpa/sherpa/pull/1638 / https://github.com/sherpa/sherpa/pull/1667 as it was meant to be a small, separate, item, but addressing #1718 complicated it a little bit.

# Example

With the main branch - if you set a background that's not DataPHA you get an error about grouping. You can set a background with a different 

```
>>> dataspace1d(1, 10, dstype=DataPHA)
>>> dataspace1d(1, 10, dstype=Data1D, bkg_id=1)
  File "<string>", line 1, in dataspace1d
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/ui/utils.py", line 822, in dataspace1d
    self._get_pha_data(id).set_background(dstype('bkg_dataspace1d',
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/data.py", line 2806, in set_background
    if bkg.grouping is None:
AttributeError: 'Data1D' object has no attribute 'grouping'
>>> dataspace1d(1, 20, dstype=DataPHA, bkg_id=1)
>>>
```

With this branch both the bkg_id lines are errors, but hopefully more descriptive:

```
>>> dataspace1d(1, 10, dstype=DataPHA)
>>> dataspace1d(1, 10, dstype=Data1D, bkg_id=1)
  File "<string>", line 1, in dataspace1d
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/ui/utils.py", line 833, in dataspace1d
    raise ArgumentTypeErr("badarg", "dstype", "set to DataPHA")
sherpa.utils.err.ArgumentTypeErr: 'dstype' must be set to DataPHA
>>> dataspace1d(1, 20, dstype=DataPHA, bkg_id=1)
  File "<string>", line 1, in dataspace1d
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/ui/utils.py", line 837, in dataspace1d
    data.set_background(bkg, bkg_id)
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/data.py", line 2810, in set_background
    raise DataErr("The source and background channels differ")
sherpa.utils.err.DataErr: The source and background channels differ
```

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2794,10 +2794,10 @@ must be an integer.""")
 
         # Check that the background matches the source (i.e. self)
         # dataset.  For this use case we require that the source has
-        # set up it's channel array (ie self.channel can not be None),
+        # set up its channel array (ie self.channel cannot be None),
         # as allowing the background dataset to have channels but not
-        # the source makes tracking the state harder than the ability
-        # is likely worth).
+        # the source makes tracking the state harder and there is no
+        # obvious use case for that.
         #
         if self.channel is None:
             raise DataErr("The channel field must be set before adding a background")
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2792,6 +2792,23 @@ must be an integer.""")
         if not isinstance(bkg, DataPHA):
             raise ArgumentTypeErr("badarg", "bkg", "a PHA data set")
 
+        # Check that the background matches the source (i.e. self)
+        # dataset.  For this use case we require that the source has
+        # set up it's channel array (ie self.channel can not be None),
+        # as allowing the background dataset to have channels but not
+        # the source makes tracking the state harder than the ability
+        # is likely worth).
+        #
+        if self.channel is None:
+            raise DataErr("The channel field must be set before adding a background")
+
+        if bkg.channel is None:
+            raise DataErr("The channel field of the background must be set")
+
+        if len(self.channel) != len(bkg.channel) or \
+           numpy.any(self.channel != bkg.channel):
+            raise DataErr("The source and background channels differ")
+
         id = self._fix_background_id(id)
         self._backgrounds[id] = bkg
         ids = self.background_ids[:]
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022
+#  Copyright (C) 2008, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -128,7 +128,7 @@ from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
     IntegratedDataSpace2D, _check
 from sherpa.models.regrid import EvaluationSpace1D
 from sherpa.stats import Chi2XspecVar
-from sherpa.utils.err import DataErr, ImportErr
+from sherpa.utils.err import ArgumentTypeErr, DataErr, ImportErr
 from sherpa.utils import SherpaFloat, pad_bounding_box, interpolate, \
     create_expr, create_expr_integrated, parse_expr, bool_cast, rebin, filter_bins
 from sherpa.utils import formatting
@@ -2788,6 +2788,10 @@ must be an integer.""")
         external programs, such as XSPEC.
 
         """
+
+        if not isinstance(bkg, DataPHA):
+            raise ArgumentTypeErr("badarg", "bkg", "a PHA data set")
+
         id = self._fix_background_id(id)
         self._backgrounds[id] = bkg
         ids = self.background_ids[:]
--- a/sherpa/astro/ui/tests/test_astro_ui_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_unit.py
@@ -3137,3 +3137,33 @@ def test_set_xxxscal_id(label, clean_astro_ui):
 
     assert got2 / got1 == pytest.approx(2)
     assert got2 == pytest.approx(2.4e-2)
+
+
+@pytest.mark.parametrize("idval", [None, 1, "up"])
+@pytest.mark.parametrize("bkg_id", [1, "up"])
+def test_dataspace1d_is_a_background(idval, bkg_id, clean_astro_ui):
+    """What happens if try to make a PHA background?"""
+
+    ui.dataspace1d(1, 10, id=idval, dstype=ui.DataPHA)
+    ui.dataspace1d(1, 10, id=idval, bkg_id=bkg_id, dstype=ui.DataPHA)
+    expected = [1] if idval is None else [idval]
+    assert ui.list_data_ids() == pytest.approx(expected)
+    assert ui.list_bkg_ids(idval) == pytest.approx([bkg_id])
+
+
+def test_dataspace1d_not_a_background(clean_astro_ui):
+    """What happens if try to make a non-PHA background?"""
+
+    ui.dataspace1d(1, 10, dstype=ui.DataPHA)
+    with pytest.raises(AttributeError,
+                       match="^'Data1DInt' object has no attribute 'grouping'$"):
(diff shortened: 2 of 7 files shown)


sherpa/astro/data.py
```diff
@@ -2788,6 +2788,27 @@ def set_background(self, bkg, id=None):
         external programs, such as XSPEC.
 
         """
+
+        if not isinstance(bkg, DataPHA):
+            raise ArgumentTypeErr("badarg", "bkg", "a PHA data set")
+
+        # Check that the background matches the source (i.e. self)
+        # dataset.  For this use case we require that the source has
+        # set up it's channel array (ie self.channel can not be None),
```
hamogu, line 2797:
```suggestion
        # set up its channel array (ie self.channel cannot be None),
```

sherpa/astro/data.py
```diff
@@ -2788,6 +2788,27 @@ def set_background(self, bkg, id=None):
...
+        # as allowing the background dataset to have channels but not
+        # the source makes tracking the state harder than the ability
+        # is likely worth).
```
hamogu, line 2800:
```suggestion
        # the source makes tracking the state harder and there is no
        # obvious use case for that.
```

sherpa/astro/data.py
```diff
@@ -2788,6 +2788,27 @@ def set_background(self, bkg, id=None):
...
+        #
+        if self.channel is None:
+            raise DataErr("The channel field must be set before adding a background")
```
hamogu, line 2803:
```suggestion
            raise DataErr("The channel field must be set before adding a background.")
```

sherpa/astro/data.py
```diff
@@ -2788,6 +2788,27 @@ def set_background(self, bkg, id=None):
...
+
+        if bkg.channel is None:
+            raise DataErr("The channel field of the background must be set")
```
hamogu, line 2806:
```suggestion
            raise DataErr("The channel field of the background must be set.")
```

sherpa/astro/data.py
```diff
@@ -2788,6 +2788,27 @@ def set_background(self, bkg, id=None):
...
+
+        if len(self.channel) != len(bkg.channel) or \
+           numpy.any(self.channel != bkg.channel):
+            raise DataErr("The source and background channels differ")
```
hamogu, line 2810:
```suggestion
            raise DataErr("The source and background channels differ.")
```

sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -3137,3 +3137,33 @@ def test_set_xxxscal_id(label, clean_astro_ui):
 
     assert got2 / got1 == pytest.approx(2)
     assert got2 == pytest.approx(2.4e-2)
+
+
+@pytest.mark.parametrize("idval", [None, 1, "up"])
+@pytest.mark.parametrize("bkg_id", [1, "up"])
+def test_dataspace1d_is_a_background(idval, bkg_id, clean_astro_ui):
+    """What happens if try to make a PHA background?"""
```
hamogu, line 3145:
```suggestion
    """What happens if we try to make a PHA background?"""
```

sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -3137,3 +3137,33 @@ def test_set_xxxscal_id(label, clean_astro_ui):
...
+
+    ui.dataspace1d(1, 10, id=idval, dstype=ui.DataPHA)
+    ui.dataspace1d(1, 10, id=idval, bkg_id=bkg_id, dstype=ui.DataPHA)
+    expected = [1] if idval is None else [idval]
+    assert ui.list_data_ids() == pytest.approx(expected)
+    assert ui.list_bkg_ids(idval) == pytest.approx([bkg_id])
+
+
+def test_dataspace1d_not_a_background(clean_astro_ui):
+    """What happens if try to make a non-PHA background?"""
```
hamogu, line 3155:
```suggestion
    """What happens if we try to make a non-PHA background?"""
```

sherpa/astro/ui/utils.py
```diff
@@ -2417,8 +2455,9 @@ def set_filter(self, id, val=None, bkg_id=None, ignore=False):
         if val is None:
             val, id = id, val
 
-        d = self.get_data(id)
-        if bkg_id is not None:
+        if bkg_id is None:
+            d = self.get_data(id)
+        else:
             d = self.get_bkg(id, bkg_id)
```
hamogu, line 2461: Since this 4-line pattern is to common, does it make sense to have something equivalent to `_get_pha_data(idval, bkg_id)` like the following?
```Python
def _get_data_or_bkg(idval, bkg_id):
    if bkg_id is None:
        d = self.get_data(id)
    else:
        d = self.get_bkg(id, bkg_id)
```

hamogu (changes requested):
Mostly typos in strings...

sherpa/astro/data.py
DougBurke: Most (too lazy to check the numbers) of our existing errors do not end in a full stop.

sherpa/astro/ui/utils.py
hamogu: I saw that comment, but I found it easier to say it where it's happening.

hamogu (approved):

PR #1720, opened 2023-03-03
Remove mutable argument from a utils routine

DougBurke:
# Summary

Internal code clean up to remove use of a mutable argument from the utils code. There should be no user-visible change.

# Details

pylint complains about mutable arguments - e.g. #95 - and complained about this one. So the aim was to remove the use of `{}` as the argument, which turned out to be easy as we don't actually use the capability. We can just return a dict that the routine creates instead.

There's more changes related to code clean up, done whilst investigating the needed code change.

This was taken from https://github.com/sherpa/sherpa/pull/1638 / https://github.com/sherpa/sherpa/pull/1667 as it is a small, separate, item.


DougBurke:
Rebasing to deal with recent conflict.

hamogu:
@DougBurke certainly wind for the longest-commit-message I've ever seen.

DougBurke:
I've been trying to find the recent-ish commit message for a one-line patch to the linux kernel, that dwarfs this, but my search-foo is broken.

Aha - here we go: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=11933cf1d91d

sherpa/astro/ui/tests/test_astro_ui_background.py
```diff
@@ -1960,3 +1987,54 @@ def test_bkg_analysis_setting_changed(idval, direct, analysis, clean_astro_ui):
...
+    ui.fit_bkg()
+    fres = ui.get_fit_results()
+    assert fres.datasets == (idopt, )
+    assert fres.succeeded
+    assert fres.parnames == ("bmdldn.c0", )
+    assert fres.numpoints == 3
+    assert fres.dof == 2
+    assert fres.istatval == pytest.approx(0.83)
+    assert fres.statval == pytest.approx(2 / 3)
+    assert fres.parvals == pytest.approx([3 + 1/3])
```
hamogu, line 2040: I have trouble building sherpa again; happens every time I don't develop for a few weeks and every time I forget exactly what's wrong and every time it takes ma a while to get up and running again. Thus, I unfortunately won't try this out on my own right now, but I wonder why the fitted parameters is 3.3333, when the input data for the dependent axis in "down" os [1, 0, 0]. Should the best fit not be 0.33333 instead of 3.333333?

hamogu (changes requested):
Do we need all this level of indirection with `_xxx` methods that get something, then pack up, and then unpack again? To me that suggests tha maybe we can get rid of some of the individual functions and inline them where called. It's easier to have one longer function, then to jump back and forth, pack and unpack, if they are called only once.

Alternatively, some of the knowlegde of who calls what in what order should be written down, maybe in the docstring of the utils moduls? The text exisit, it just need to be copied from the issue description or the commit message into the code file.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -1606,3 +1605,76 @@ def test_get_rate_bkg(idval, clean_astro_ui, make_data_path):
...
+    ui.set_bkg(idval, ui.DataPHA("b", [1, 2], [1, 1]))
+
+    ui.set_source(idval, ui.const1d.smdl)
+    ui.set_bkg_source(idval, ui.const1d.bmdl)
+
+    ui.subtract(idval)
+
+    # The hide_logging fixture has changed the loglevel so we need to
+    # change it here to make sure we see the warning.
+    #
```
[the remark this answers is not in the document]
DougBurke: I think this is an outdated comment (from the initial version of the test I wrote before I ended up removing the test-level `hide_logging` fixture. Will remove.

sherpa/astro/ui/tests/test_astro_ui_background.py
hamogu: Understood.

DougBurke: There are places where I try to explain the values (because when I've reviewed them or when writing the tests) I want to explain it enough to be able to find again. Do you think it needed here (my suggestion is no, but YMMV)?

sherpa/astro/ui/utils.py
```diff
@@ -3979,10 +3982,10 @@ def _save_type(self, objtype, id, filename, bkg_id=None, **kwargs):
         args = None
         fields = None
 
-#        if type(d) in (sherpa.data.Data1DInt, sherpa.astro.data.DataPHA):
+#        if type(d) in (sherpa.data.Data1DInt, DataPHA):
 #            args = [obj.xlo, obj.xhi, obj.y]
 #            fields = ["XLO", "XHI", str(objtype).upper()]
```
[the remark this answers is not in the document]
DougBurke: There's an argument to be had that we should do this (ie uncomment the lines), but I don't have the tuits to do it.

For the non-native English speakers

a tuit

sherpa/astro/ui/tests/test_astro_ui.py
DougBurke, line 1633:
```suggestion
```

sherpa/ui/utils.py
```diff
@@ -8296,127 +8298,161 @@ def unlink(self, par):
...
+
+        # Check the data exists. At this point ids is not empty.
+        #
+        out = []
+        for idval in ids:
+            out.append((idval, self.get_data(idval)))
+
+        return out
+
+    def _get_fit_obj(self, store, estmethod, numcores=1):
```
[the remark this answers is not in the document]
DougBurke: I've added a bunch of docstrings for related routines like `_get_fit_obj` for your perusal.

DougBurke:
I think I'm going to look at

> I originally thought about using something with more domain-specific structure (e.g. use a dataclass to have named fields) but it was more effort than it was worth. Now the code has finalised some structure could be added (maybe just use a dict rather than a tuple) but I'm not sure it's worth it (but maybe in 6 months time I'll re-think this)

and see how it ends up looking.

DougBurke:
Rebasing to add another test and to address an infelicity in a recent commit - I'll do it in stages so we can check things still pass.

DougBurke:
@hamogu - I think this is ready for your next review

hamogu (approved):
I'm fine. 

On the procedure I want to note though that re-reviewing is made much harder by rebasing. I understand you rebase to keep the commit timeline clean and to have each commit be a logical unit and in a logical order (tests fist, then code changes), but it also overwrite the old commits that I knew and commented on. So, a re-review is essentially a new review and I have no idea if the fact that I did not spot any issues now means that it's all god or that I just did not pay attention - I can't back out my old comments.

PR #1726, opened 2023-03-08
muzaffer015-patch-1

muzaffer015:
muzaffer015-patch-1

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,24 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,24 @@
...
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.
+
+muzaffer015-yama-1
+Muzafferbayindir
```
hamogu, line 24: What are these lines supposed to do?

Also, it might be useful to not just drop a random file into a PR, but, if you specifically want to contribute to Sherpa, include is in the real documentation (which is rst, not md).

DougBurke:
I'm going to close this at the start of next week unless we get more information on why we should care about it.

Issue #1383, opened 2021-12-21
Following instructions to build sherpa from source with XSPEC causes fatal error


wardhoward:
Dear developer,

I am currently (in tears) trying to build sherpa from source in a conda environment that already contains ciao and astropy since the default conda version of sherpa I originally used does not support any XSPEC models. I successfully downloaded, configured, built, and installed heasoft-6.29 to get XANADU and XSPEC 12.12.0

Next, I downloaded the sherpa source code in my "chandra" conda env following the directions here and here as follows:
HERE: https://sherpa.readthedocs.io/en/latest/install.html
HERE: https://sherpa.readthedocs.io/en/latest/developer/index.html#source-install-with-conda

First, I un-installed and purged my earlier install of sherpa:
conda remove sherpa
find . -name "*sherpa*" to find any leftover cache files, etc
rm <file 1> <file 2> <etc>

Then I installed (well, super-seded) the dependencies into my chandra environment which already had ciao and astropy installed

conda install -n chandra --only-deps -c https://cxc.cfa.harvard.edu/conda/ciao sherpa
conda install -n chandra -c anaconda -c astropy sphinx graphviz sphinx-astropy sphinx_rtd_theme

git clone git://github.com/sherpa/sherpa.git
cd sherpa
git checkout 4.13.0

Next, I open the setup.cfg file and paste the XSPEC version information from the online instructions for heasoft 6.29 linked in the documentation above:
with-xspec = True
xspec_version = 12.12.0
xspec_lib_dirs = $HEADAS/lib
xspec_include_dirs = $HEADAS/include
xspec_libraries = XSFunctions XSUtil XS hdsp_6.29
ccfits_libraries = CCfits_2.6
wcslib_libraries = wcs-7.3.1

Next, I manually extend $HEADAS with my local env in the setup.cfg file as follows, then verify both directories exist on my machine:
xspec_lib_dirs = /home/wshoward/chandra/heasoft-6.29/x86_64-pc-linux-gnu-libc2.23/lib
xspec_include_dirs = /home/wshoward/chandra/heasoft-6.29/x86_64-pc-linux-gnu-libc2.23/include

I also init the heasoft installation from my terminal.
export HEADAS=/home/wshoward/chandra/heasoft-6.29/x86_64-pc-linux-gnu-libc2.23
. $HEADAS/headas-init.sh

Finally, I run python setup.py build
[...]

Issue #1375, opened 2021-12-04
Overarching vision for plot object and backend changes


hamogu:
The goal of this issue is to lay out, in narrative text, the broad line of planned plotting backend improvements; in that sense this text is a design document similar to the grouping/filtering word docs we've been looking at. At the completion of this project, essentially all text in this issue will have been transfered into the sherpa docs.

**I will edit this top description to adjust to any comments that are given in this issue. I take a "thumbs up" as general approval. Details can be discussed in individual issues, but leave comment here for big-picture suggestions.**

This issue is meant as a complement to the project tracking in https://github.com/sherpa/sherpa/projects/6. The project board tracks individual issues, but does not have the narrative text that I'm dumping into this issue for discussion. Individual issues are linked from the project board and therefore not all issues and PRs are listed here individually. 

## Step 1: Backends for plotting
We want backends to be cleanly separated from Sherpa code; they were always supposed to be but since we did have only one backend for a while and did not test for that, dependencies have snug in (#618, https://github.com/sherpa/sherpa/pull/1356#issuecomment-985764507). This will enable cleaner testing (#1229), hot-swapping backends (#348, #1191), simplify the implementation of any new backends (#499, #1356), and simplify maintenance of existing backends.

### Backends as objects, not modules
I proposal that backends should be instances of a class, not modules. There are multiple advantages:
- Objects can use inheritance - see how much code is duplicated in the pylab and the new bokeh backend in #1356 (as of 12/3/21 - #1356 will probably be rebased and re-worked to reduce that). 
- New backends can inherit from existing ones, which simplifies implementation as capabilities can be added piece by piece.
- Users can subclass a backend. For example, I would like to change the way error bars are displayed in matplotlib and show them as shaded regions instead of separate error bars. It's much more Pythonic to subclass, then to moneky-patch the exiting module. This is admittedly a power-user feature, but one that can simplify the use of sherpa as a library for others (like I intend for TGCat and others tried for gammapy in the past).
- Its easier to assign a new object then to switch out a module.
- Sherpa's plotting assumes a stateful backend with the notion of a "current plot" as in `plt.gca`. That is one approach many plotting packages provide (e.g. pylab in matplotlib) and that works well for Sherpa. However, for plotting packages that are more OO, the backend need to track the "current figure" etc. That could be done as a module-level variable in `sherpa.plot.XXX_backend`, but is much more natural as an instance attribute.

[...]

PR #1729, opened 2023-03-11
plot: add sherpa_plot context manager

DougBurke:
# Summary

Add a sherpa_plot context manager to simplify creating a plot with the Sherpa plotting objects (only an issue if calling the plot objects directly, and not with commands like the UI plot_data call).

# Details

Creating a sherpa plot follows the same template, and it fits in well to being a context manager (even though there is no "plot object" we have to attach the open/close logic to, we can use the contextmanager decorator).

This is mainly useful internally, but it may be useful outside of Sherpa.

It should be separate from (but will conflict with) the plot-backend rework in #1382 / #1550. I guess we could add the backend-name as an optional argument to the context manager, to over-ride the backend for just that plot. Not sure how useful that would be, but it sounds cool!

**OOhhhhh** - a thought. We *could* make the backend provide a `sherpa_plot`-like routine, and then this version would just call that. Thoughts

- we could then remove the `setup`/`teardown`/`exceptions` logic as this would now just be handled within the plotend so if you don't need to do anything special with exception-handling you just don't do it
- normally you would add a context manager to a class and have it associated with the lifetime of that object, but that's not true here; is this confusing?

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2009, 2015, 2016, 2018, 2019, 2020, 2021, 2022
+#  Copyright (C) 2009, 2015, 2016, 2018, 2019, 2020, 2021, 2022, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -32,7 +32,7 @@ Which backend is used?
 When this module is first imported, Sherpa tries to import the
 backends installed with Sherpa in the order listed in the
 ``options.plot_pkg`` setting from the ``sherpa.rc`` startup file.
-The first module that imports successfully is set as the active 
+The first module that imports successfully is set as the active
 backend. The following command prints the name and the location
 on disk of that module::
 
@@ -49,8 +49,33 @@ provides the same interface):
   >>> import sherpa.plot.pylab_backend
   >>> plot.backend = sherpa.plot.pylab_backend
 
+Creating a plot
+---------------
+
+The simplest approach is to use the `sherpa_plot` context manager::
+
+    with sherpa_plot():
+        # Now call the plot/overplot or contor/overcontour methods
+        obj.plot()
+
+This handles setting up the backend, handles any error handling,
+and then ends the session. It is equivalent to::
+
+    from sherpa import plot
+    try:
+        plot.backend.begin()
+        # Now call the plot/overplot or contor/overcontour methods
+        obj.plot()
+    except:
+        plot.backend.exceptions()
+        raise
+    else:
+        plot.backend.end()
+
 """
+
 from configparser import ConfigParser
+from contextlib import contextmanager
 import logging
 import importlib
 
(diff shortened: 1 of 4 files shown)


sherpa/plot/__init__.py
```diff
@@ -49,8 +49,33 @@
   >>> import sherpa.plot.pylab_backend
   >>> plot.backend = sherpa.plot.pylab_backend
 
+Creating a plot
+---------------
+
+The simplest approach is to use the `sherpa_plot` context manager::
+
+    with sherpa_plot():
+        # Now call the plot/overplot or contor/overcontour methods
```
wmclaugh, line 58:
```suggestion
        # Now call the plot/overplot or contour/overcontour methods
```

sherpa/plot/__init__.py
```diff
@@ -49,8 +49,33 @@
...
+        obj.plot()
+
+This handles setting up the backend, handles any error handling,
+and then ends the session. It is equivalent to::
+
+    from sherpa import plot
+    try:
+        plot.backend.begin()
+        # Now call the plot/overplot or contor/overcontour methods
```
wmclaugh, line 67:
```suggestion
        # Now call the plot/overplot or contour/overcontour methods
```

DougBurke: I'm thinking that this "equivalent to" section can just be dropped, as I think we want to make `sherpa_plot` (or whatever we decide to bike-shed the name to) would become the way to setup plots, but need to look at what other backends do (related to my excitement above in working out possible synergies with the backend rework).

I'll still take the change though!

hamogu:
This is a slightly different angle to take than #1382 as it would be a change in the API of how to do a plot, which #1382 tries to avoid (not entirely, but as much as reasonable).

A context manager *is* the Pythoninc way to deal with the setup/teardown, so this is worth thinking about. I don't think is collides as much with #1382, as with later steps on where the plot preferences live. My view is that some preferences live in the plot object (symbols/lines for spectral data), while other might live in the backend (e.g. the default color map which will be named differently in matplotlib vs. bokeh, vs. altair). I've written up thoughts about that in multiple places and I'm not going to pull it together in this location again, but it's partially discussed in #1375. So, when I call this context manager, do I get a default version of the backend? Do I get a copy of the current settings? How do I persist settings if the backend is generated in the context manager? Or would I pass a backend *instance* (as opposed the the name/class) into the context manager, because the *instance* is were the user-set preferences live.

All that can be done, but I wonder if it's feature creep - in short, I suggest to keep this idea in the back of our minds at least until #1383 is in.

hamogu:
I had forgotten about this PR and have in fact implemented something similar in https://github.com/hamogu/sherpa/commit/0be22d5ec690a1c63cdb6d5364ed8a0f6a327d83

DougBurke:
@hamogu - do you want me to close this PR down then?

DougBurke:
@hamogu - I plan to close this but should I keep it open?

PR #1730, opened 2023-03-15
Update xspec model creation

### LINKED ISSUE #1345: XSPEC parsing/code creation
4.14.0 adds the sherpa.astro.utils.xspec module. I've been updating the contrib script `contrib_xspec_user_model` to use it (since it is where the code was originally developed) and there appear to be a few issues

- https://github.com/cxcsds/ciao-contrib/pull/525 shows that for `C_<mdl>` routines we need to include the `cppModelWrapper` symbol (which is defined in `libXSFunctions` but doesn't appear to be defined in a XSPE Cinclude file), declare the `<mdl>` C++ routine, and create the `C_<mdl>` routine which calls `cppModelWrapper`
- https://github.com/cxcsds/ciao-contrib/pull/522 points to a missing `;` after defining the `Wrappers` structure

### LINKED ISSUE #1402: Tweaks required to sherpa.astro.utils.xspec as we bump the XSPEC minimum version
So #1387 replaces explicit function declarations with XSPEC typedefs, and if we can bump up to 12.12.0 as the minimum - see #1391 - then we can easily reference them (the version in #1387 has to manually include the typedefs).

This means the `sherpa.astro.utils.xspec.create_xspec_code` routine needs to be updated to something like

```
% git diff sherpa
diff --git a/sherpa/astro/utils/xspec.py b/sherpa/astro/utils/xspec.py
index 0b96b16f..dd3344aa 100644
--- a/sherpa/astro/utils/xspec.py
+++ b/sherpa/astro/utils/xspec.py
@@ -944,13 +944,13 @@ def model_to_compiled(mdl):
     #
     defcode = ''
     if is_fortran:
-        defcode = f'  void {mdl.funcname}'
+        defcode = '  '
         if mdl.language == 'Fortran - single precision':
-            # Note add _ (assume this is needed)
-            defcode += '_(float* ear, int* ne, float* param, int* ifl, float* photar, float* photer);'
+            defcode += 'xsf77Call'
         elif mdl.language == 'Fortran - double precision':
-            # Note add _ (assume this is needed)
-            defcode += '_(double* ear, int* ne, double* param, int* ifl, double* photar, double* photer);'
+            defcode += 'xsF77Call'
+
+        defcode += f' {mdl.funcname}_;'

     return wrapcode, defcode

```

but probably with a `#include <XSFunctions/Utilities/funcType.h>` line

DougBurke:
# Summary

Update the code created by the sherpa.astro.utils.xspec routines to account for changes in the XSPEC interface in Sherpa. This only affects users who call these routines directly.

# Details

Move fixes noticed in `convert_xspec_user_model` from https://github.com/cxcsds/ciao-contrib from the 4.15 release.

The tests I had written were done so in such a way that they just tested the output for the parts that I thought were important: for example they would split the output by line and just checked that certain lines existed. This meant that they didn't notice lines that had not been expanded properly (e.g. f-strings that hadn't had f qualifier for the string) or checking the ordering of certain statements. The code now just does an explicit check for "all" the output as it isn't a lot of text to add and makes the checks more comprehensive. There is an issue that the output now depends on the XSPEC library, so there's a little complication there, but that's handled by a single routine in the tests.

I've also added basic typing rules to this module as it's fairly easy to do so and this module is "external" to most users (i.e. it is really mean to make my life easier as the maintainer of `convert_xspec_model_script`, and it is unlikely that anyone else uses it).

Fix #1345 
Fix #1402

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -369,16 +369,14 @@ These routines are designed to simplify the process but are not
 guaranteed to handle all cases (as the model.dat file syntax is not
 strongly specified).
 
-As an example of their use, the following output is from before
-XSPEC 12.12.0 (released in HEASOFT 6.29) support was added::
+As an example of their use (the output will depend on the current
+Sherpa and XSPEC versions)::
 
-  % ./scripts/check_xspec_update.py ~/local/heasoft-6.29/spectral/manager/model.dat | grep support
-  We do not support grbjet (Add; xsgrbjet)
+  % ./scripts/check_xspec_update.py ~/local/heasoft-6.31/spectral/manager/model.dat | grep support
   We do not support smaug (Add; xsmaug)
-  We do not support vvwdem (Add; vvwDem)
-  We do not support vwdem (Add; vwDem)
-  We do not support wdem (Add; wDem)
-  We do not support zxipab (Mul; zxipab)
+  We do not support polconst (Mul; polconst)
+  We do not support pollin (Mul; pollin)
+  We do not support polpow (Mul; polpow)
   We do not support pileup (Acn; pileup)
 
 .. note::
@@ -387,15 +385,39 @@ XSPEC 12.12.0 (released in HEASOFT 6.29) support was added::
    on the list of models that could be added to
    :py:mod:`sherpa.astro.xspec`
 
-The code needed to add support for the wdem module can be found with::
+Although the ``wdem`` model is included in the XSPEC models, here is
+how the ``add_xspec_model.py`` script can be used for those models
+noted as not being supported::
 
-  % ./scripts/add_xspec_model.py ~/local/heasoft-6.29/spectral/manager/model.dat wdem
+  % ./scripts/add_xspec_model.py ~/local/heasoft-6.31/spectral/manager/model.dat wdem
   # C++ code for sherpa/astro/xspec/src/_xspec.cc
+
+  // Includes
+
+  #include <iostream>
+
+  #include <xsTypes.h>
+  #include <XSFunctions/Utilities/funcType.h>
+
+  #define XSPEC_12_12_0
+  #define XSPEC_12_12_1
+  #define XSPEC_12_13_0
+
+  #include "sherpa/astro/xspec_extension.hh"
 
   // Defines
 
-  extern "C" {
+  void cppModelWrapper(const double* energy, int nFlux, const double* params,
+    int spectrumNumber, double* flux, double* fluxError, const char* initStr,
+    int nPar, void (*cppFunc)(const RealArray&, const RealArray&,
+    int, RealArray&, RealArray&, const string&));
 
+  extern "C" {
+    XSCCall wDem;
+    void C_wDem(const double* energy, int nFlux, const double* params, int spectrumNumber, double* flux, double* fluxError, const char* initStr) {
+      const size_t nPar = 8;
+      cppModelWrapper(energy, nFlux, params, spectrumNumber, flux, fluxError, initStr, nPar, wDem);
+    }
   }
 
   // Wrapper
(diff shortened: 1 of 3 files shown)


hamogu:
This is mostly for developers and not users. On the other hand, it's also relatively low risk - it only prints out stuff that helps developers update XSPEC and won't break users. Do you want me to try to review that to the best of my abilities now or afte the 4.16 freeze?

DougBurke:
It needs to be done before the 4.16 freeze, otherwise it's not useful (as then the CIAO version will still use the old version).

I agree that the scope of this is limited (in that it's basically fr the benefit of code I write for CIAO), so it's less mission critical than other pieces to address all issues.

sherpa/astro/utils/xspec.py
```diff
@@ -972,30 +1044,115 @@ def models_to_compiled(mdls):
...
+    # What includes are needed?
+    #
+    out = marker("Includes")
+    out += '#include <iostream>\n\n'
+    out += '#include <xsTypes.h>\n'
+    out += '#include <XSFunctions/Utilities/funcType.h>\n\n'
+
+    # The Sherpa/XSPEC interface uses a number of defines to control
+    # behavior. These should not be needed for user models, but set
+    # them up. Note that tey depend on the available XSPEC library,
```
hamogu, line 1087:
```suggestion
    # them up. Note that they depend on the available XSPEC library,
```

hamogu (approved):
I'm really not qualified to review much of this, but those small parts that I understand looks OK. Again, this generates code that a human will then have to look at, so if there is anything obviously wrong, it will hopefully be caught at that stage.

docs/developer/index.rst
```diff
@@ -427,22 +464,27 @@ The code needed to add support for the wdem module can be found with::
...
+          XSAdditiveModel.__init__(self, name, (self.Tmax,self.beta,self.inv_slope,self.nH,self.abundanc,self.Redshift,self.switch,self.norm))
 
 
 This code then can then be added to
 ``sherpa/astro/xspec/src/_xspec.cc`` and
 ``sherpa/astro/xspec/__init__.py`` and then refined so that the tests
 pass.
 
+.. note::
+   The outut from ``add_xspec_model.py`` is designed for XSPEC user
```
wmclaugh, line 484:
```suggestion
   The output from ``add_xspec_model.py`` is designed for XSPEC user
```

sherpa/astro/utils/tests/test_astro_utils_xspec.py
```diff
@@ -332,8 +333,82 @@ def test_parse_multiplicative_model():
     assert model.pars[3].hardmax == pytest.approx(1e25)
 
 
+def check_compiled(got, suffix):
+    """Check the compiled output.
 
-def test_create_model_additive():
+    The output depends on the XSPEC version because it will cause
+
+        #define XSPEC_major_micro_minor
```
wmclaugh, line 341: Typically, versions are defined in major.minor.micro. Is the order in XSPEC_major_micro_minor correct? (Comment applies to other references in this file)

sherpa/astro/utils/xspec.py
```diff
@@ -811,47 +854,56 @@ def simple_wrap(modelname, mdl):
...
-        if mdl.flags[0] == 1:
-            out += f"{t2}warnings.warn('support for models like {mdl.clname.lower()} "
-            out += "(variances are calculated by the model) is untested.')\n"
 
-        if nflags > 1 and mdl.flags[1] == 1:
-            out += "{t2}warnings.warn('support for models like {mdl.clname.lower()} "
-            out += "(recalculated per spectrum) is untested.')\n"
+    # If the model needs to be recalculated-per-spectrum turn off the
+    # caching. This needs to be done after the parent class has been
+    # initalized.
```
wmclaugh, line 863:
```suggestion
    # initialized.
```

sherpa/astro/utils/tests/test_astro_utils_xspec.py
DougBurke: As mentioned this was a thinko on my part - see my comment at https://github.com/sherpa/sherpa/pull/1730#discussion_r1332967520

Issue #1240, opened 2021-07-29
XSPEC set_abund from a file only works once


DougBurke:
This is with CIAO 4.14 beta 1 but the exact version doesn't matter

```python
sherpa In [1]: get_xsabund('He')
Out[1]: 0.09769999980926514

sherpa In [2]: set_xsabund('hack.dat')
 Solar Abundance Vector set to file:  User defined abundance vector / no description specified

sherpa In [3]: get_xsabund('He')
Out[3]: 2.0

sherpa In [4]: !head -3 hack.dat
1
2
3

sherpa In [5]: !head -3 hack2.dat
10
11
12

sherpa In [6]: set_xsabund('hack2.dat')
 Solar Abundance Vector set to file:  User defined abundance vector / no description specified

sherpa In [7]: get_xsabund('He')
Out[7]: 2.0
```

Note that the last line returns 2.0 and not 11.0.

The xFortran.cxx interface isn't setting the right things behid the scenes (seems like we should use `RFLABD` and `FPSLFL` is either wrong or needs other calls which aren't documented).

DougBurke:
I thought we could work around this by internally resetting to `angr` (or another table) before applying the new settings but this works apparently randomly.

```python
sherpa In [1]: get_xsabund('He')
Out[1]: 0.09769999980926514

sherpa In [2]: get_xsabund()
Out[2]: 'angr'

sherpa In [3]: set_xsabund('hack2.dat')
 Solar Abundance Vector set to file:  User defined abundance vector / no description specified

sherpa In [4]: get_xsabund('He')
Out[4]: 11.0

sherpa In [5]: get_xsabund()
Out[5]: 'file'

sherpa In [6]: set_xsabund('angr')
 Solar Abundance Vector set to angr:  Anders E. & Grevesse N. Geochimica et Cosmochimica Acta 53, 197 (1989)

sherpa In [7]: get_xsabund()
Out[7]: 'angr'

sherpa In [8]: get_xsabund('He')
Out[8]: 0.09769999980926514

sherpa In [9]: set_xsabund('hack.dat')
 Solar Abundance Vector set to file:  User defined abundance vector / no description specified

sherpa In [10]: get_xsabund('He')
Out[10]: 2.0
```

So far so good

```python
sherpa In [11]: set_xsabund('angr')
 Solar Abundance Vector set to angr:  Anders E. & Grevesse N. Geochimica et Cosmochimica Acta 53, 197 (1989)

[...]

PR #1732, opened 2023-03-21
Tests: add a check for #1240

DougBurke:
# Summary

Add a test for an odd issue with an old version of XSPEC (issue #1240).
 
# Details

Issue #1240 was strange, and appears to have been an issue in XSPEC rather than Sherpa, but add a check to see if it happens again.

The test code is also updated to using the tmp_path fixture from pytest rather than the tempfile module, as it is a bit nicer / more hygenic.

There is the question of whether this is worth it (it doesn't take much time, but it does add to the carbon footprint of our CI).


DougBurke:
I've decided it's not worth the test.

PR #1733, opened 2023-03-28
Allow fake_pha to be called with a list of ARF/RMF names of length 1

DougBurke:
# Summary

Allow fake_pha to be called with a list of ARF and RMF names that only contains one element. Fix #1722

# Details

The problem was that the code used `range(n)` rather than `range(1, n+1)` when loading the ARF/RMF files. This meant that the first response was set with `resp_id=0` rather than `resp_id=1`, and then later on the code called `self.get_response()`, which defaults to looking for `resp_id=1` which is not set. Hence the problem in #1722.

Of course, why is the code calling `get_response()` without an id here, and is that a problem? Well, this happens deep in the code and does actually make sense (it assumes that if there's a single response set up then it has the default id, and it would be up to the caller to change that default id if necessary). Any change to this would require extensive changes that are not warranted at this time.

I originally included this as part of #1684 which started off as being a fix-up of the fake_pha docs,  but has grown somewhat. So I am carving out bits that are easy to review and provide useful user changes.

PR #1734, opened 2023-03-29
Unify the fake and fake_pha calls

DougBurke:
# Summary

The fake_pha call now accepts a method argument to match the fake call.

# Details

Requires
- #1733 

Next in series
- #1684

The fake_pha call now accepts a method argument, to match the fake call.  This is likely not that useful for users, as we really only expect a poisson distribution. There are those users who have RATE PHAs, where we can not assume Poisson stats, but I'm not sure we handle this case particularly well throughout Sherpa anyway (but we can now handle it better in fake_pha with this change). It is, however, very useful for testing the fake_pha call since you can send method the identity function (e.g. `lambda x: x`) and test the actual calculated numbers, rather than have to worry about RNG variations (which make subtle errors impossible to distinguish, as I mention below). 

A number of tests are updated to make use of this (by sending in an identity function) showing some of these problem cases. For example, in sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py there is now

```python
def identity(x):
    return x

@pytest.mark.parametrize("method,expected",
                         [(None, [42, 137, 53]),
                          (identity, [50, 151.25, 50])  # TODO: should be [50, 150, 50]
                          ])
def test_fake_pha_background_model(method, expected, clean_astro_ui, reset_seed):
    ...
```

I have not marked this as `xfail` because

- for the `None` case, it is very hard to know that 137 is coming from the "wrong" distribution (expected value of 151.15 rather than 150)
- when it gets fixed it is easier to see in the difference (in the `git diff` output)
- maybe I'm wrong and it really should be 151.25, but I have tried to add a lot of text to explain what the predicted values are, so we can fix them (e.g. in #1684) 

There are also a large number of corner-cases for the fake-pha code path added. These could be removed (and moved into #1684) if it would make this easier to review.

### DIFF
--- a/sherpa/astro/fake.py
+++ b/sherpa/astro/fake.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2021 MIT
+#  Copyright (C) 2021, 2023 MIT
 #
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -22,8 +22,8 @@ The ``fake_pha`` routine is used to create simulated
 `sherpa.astro.data.DataPHA` data objects.
 '''
 import numpy as np
-import sherpa
-from sherpa.utils.err import DataErr
+from sherpa.utils import poisson_noise
+from sherpa.utils.err import ArgumentTypeErr, DataErr
 from sherpa.astro.background import get_response_for_pha
 
 __all__ = ('fake_pha', )
@@ -31,7 +31,8 @@ __all__ = ('fake_pha', )
 
 def fake_pha(data, model,
              is_source=True, pileup_model=None,
-             add_bkgs=False, bkg_models={}, id=None):
+             add_bkgs=False, bkg_models=None, id=None,
+             method=None):
     """Simulate a PHA data set from a model.
 
     This function replaces the counts in a PHA dataset with simulated counts
@@ -52,6 +53,9 @@ def fake_pha(data, model,
     The backgrounds itself are not changed by this function. To simulate
     backgrounds as well as the source spectrum, call this function on the
     source PHA dataset and the background PHA dataset(s) independently.
+
+    .. versionchanged:: 4.15.1
+       The method parameter was added.
 
     Parameters
     ----------
@@ -82,8 +86,12 @@ def fake_pha(data, model,
         For all background ids not listed in this dictionary, the
         counts will be drawn from the PHA data of the background data set.
     id : str
-        String with id number if called from UI layer. This is only used for
-        certain error messages.
+        String with id number if called from UI layer. This is only
+        used for certain error messages.
+    method : callable or None
+        The routine used to simulate the data. If None (the default)
+        then sherpa.utils.poisson_noise is used, otherwise the function
+        must accept a single ndarray, returning a ndarray of the same shape.
 
     Examples
     --------
@@ -121,6 +129,11 @@ def fake_pha(data, model,
     if len(data.response_ids) == 0:
         raise DataErr('normffake', data.name)
 
+    if method is None:
+        method = poisson_noise
+    elif not callable(method):
+        raise ArgumentTypeErr("badarg", "method", "a callable")
+
     if is_source:
         model = get_response_for_pha(data, model, bkg_srcs={},
                                      pileup_model=pileup_model, id=id)
@@ -132,7 +145,7 @@ def fake_pha(data, model,
 
     # Calculate the source model, and take a Poisson draw based on
     # the source model.  That becomes the simulated data.
-    data.counts = sherpa.utils.poisson_noise(data.eval_model(model))
+    data.counts = method(data.eval_model(model))
 
     # Add in background counts:
     #  -- Scale each background properly given data's
@@ -148,6 +161,9 @@ def fake_pha(data, model,
     # WITHOUT background counts being added in.
     #
     if add_bkgs:
+        if bkg_models is None:
+            bkg_models = {}
+
         nbkg = len(data.background_ids)
         b = 0
         for bkg_id in data.background_ids:
@@ -175,5 +191,5 @@ def fake_pha(data, model,
 
         if nbkg > 0:
             b = b / nbkg
-            b_poisson = sherpa.utils.poisson_noise(b)
+            b_poisson = method(b)
             data.counts = data.counts + b_poisson
--- a/sherpa/astro/fake.py
+++ b/sherpa/astro/fake.py
@@ -54,7 +54,7 @@ def fake_pha(data, model,
     backgrounds as well as the source spectrum, call this function on the
     source PHA dataset and the background PHA dataset(s) independently.
 
-    .. versionchanged:: 4.15.1
+    .. versionchanged:: 4.16.0
        The method parameter was added.
 
     Parameters
--- a/sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
@@ -28,6 +28,7 @@ import numpy as np
 import pytest
 
 from sherpa.utils.testing import requires_data, requires_fits
+from sherpa.astro.instrument import create_arf, create_delta_rmf
 from sherpa.astro import ui
 from sherpa.utils.err import DataErr, IOErr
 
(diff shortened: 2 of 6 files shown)


sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -564,3 +565,186 @@ def test_fake_pha_issue_1568(make_data_path, clean_astro_ui, reset_seed):
...
+    # particularly with different exposure/backscal.
+    #
+    bkg = ui.DataPHA("fakey", [1, 2, 3], bexpected)
+    bkg.exposure = 400
+    bkg.backscal = 0.1
+
+    arf, rmf = setup_fake_pha_test()
+    ui.fake_pha(1, arf, rmf, 100, bkg=bkg)
+
+    # Check what's changed and what hasn't
```
hamogu, line 894: Would it be easier to understand if the assert lines that match the previous test where marked in some way or pulled into a separate function `def fake_pha_with_bgnd_data_XXX` that would be called in both tests to make it obvious which of the lines are the same?

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -107,6 +107,75 @@ def test_fake_pha_incompatible_rmf(idval, clean_astro_ui):
...
+
+    ui.load_arrays(idval, channels, counts, ui.DataPHA)
+    ui.set_rmf(idval, rmf)
+    ui.set_arf(idval, arf)
+
+    cpt = ui.create_model_component('const1d', 'cpt')
+    cpt.c0 = 24
+    ui.set_source(idval, cpt)
+
+    ui.fake_pha(idval, arf=None, rmf=None, exposure=None)
```
hamogu: What is it even supposed to mean if `exposure=None`? It makes no sense to run a simulation for an undefined exposure. I *might* be convinced that we could default to the exposure time of previously loaded data, but if both are `None` what could that possibly do? I don't see any use case for that and I'm in fact surprised that that does not error out with something like "can't multiply array with None` or similar. 

At the very least, I would think that the behavior is probably undefined (not because the code makes it undefined, but because the question asked "fake a spectrum, but I won't tell you how long to fake it" is ill defined. It's almost analogous to "fake a spectrum, but I won't tell you what the size of the telescope is (no ARF) or what the input model is. Neither of those makes sense and gives a result, but apparently this does. Unless I'm missing something fundamental, I think it's a bug and any non-numerical value for the exposure time shoudl be disallowed.

[Note: Because @DougBurke put a lot of effort into sorting his commits, I'm trying now to review commit by commit. It's entirely possible that this changes again in the next commit, which is why I usually review the final PR and not individual commits. I'll find out which strategy causes more confusion...]

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -817,3 +823,59 @@ def test_fake_pha_with_bgnd_data(clean_astro_ui, reset_seed):
...
+
+    assert ui.get_data().grouped
+
+    mdl = ui.create_model_component("polynom1d", "mdl")
+    ui.set_source(1, mdl)
+
+    ui.fake_pha(1, arf=None, rmf=None, exposure=10, grouped=grouped)
+
+    assert ui.get_grouping(1) == pytest.approx(grping)
+    assert ui.get_data().grouped is grouped
```
hamogu, line 952: I now this is a regression test. I wonder if we should write down this behaviour in a docstring somewhere. It's a choice we make, and I can find examples for or against it (e.g. grouping by "min_counts" would be different after a `fake_pha` call, while grouping by channel width would not). 
So, I agree not to touch whatever it's doing now, but if we codify with a test that this is what we want, should be bother to tell the user?

sherpa/astro/fake.py
```diff
@@ -82,8 +86,12 @@ def fake_pha(data, model,
...
         counts will be drawn from the PHA data of the background data set.
     id : str
-        String with id number if called from UI layer. This is only used for
-        certain error messages.
+        String with id number if called from UI layer. This is only
+        used for certain error messages.
+    method : callable or None
+        The routine used to simulate the data. If None (the default)
+        then sherpa.utils.poisson_noise is used, otherwise the function
+        must accept a single ndarray, returning a ndarray of the same shape.
```
hamogu, line 94: This is a clever addition that makes so much sense. Why did we not do that years ago?

sherpa/astro/ui/utils.py
```diff
@@ -8860,6 +8864,11 @@ def fake_pha(self, id, arf, rmf, exposure, backscal=None, areascal=None,
            calling ``fake_pha``. That background dataset needs to include
            the data itself (not used in this function), the background model,
            and the response.
+        method : callable or None, optional
+           If None, the default, then the data is simulated using the
+           sherpa.utils.poisson_noise routine. If set, it must be a
```
hamogu, line 8869:
```
`sherpa.utils.poisson_noise` routine. If set, it must be a
```
To make Sphinx render a link.

hamogu (changes requested):
Suggested changes are minimal, but I've marked a question or two for discussion. I'm OK with pushing that discussion to the already written follow-up PR or we can discuss here - whatever works best for Doug.

sherpa/astro/fake.py
anetasie: What are the choices for 'method'?

hamogu: Any callable (function or callable class) that you define yourself. Since we don't expect users to use this feature are lot, I don't think any more details are needed. There is a sensible default (None, which means Poisson) that should cover 98 % of all use cases.

DougBurke: @anetasie - the docstring says

    method: callable or None

which is, I believe, the Python way to do it (without going into type hints which would be even less intelligible).

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: So, we have a dataset identifier which can be used to define things like the channels, the response, the model, the exposure time. So I claim that `exposure=None` means "take the exposure time from the existing data", just like `rmf=None` does.

It depends if you are saying "I want to simulate the existing data (e.g. to see how reliable the results are)" or "I want to see what happens with a longer exposure". I think both are valid and should really be supported.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: I don't know what behavior you are asking about:

- is the grouping array retained
- is the grouped flag retained

I also don't know what you mean by "grouping by min_counts would be different" because that choice is made before the call - it's made when you call `group_counts` or `group_adapt_snr`. We don't "remember" the choice and re-apply it whenever the data or filter is changed [which is somehting we could think about doing but I would rather not] - we just have a grouping array and a quality array and how those arrays were created has been forgotten/lost.

sherpa/astro/ui/utils.py
DougBurke: In checking whether this already worked (which it doesn't, but sometimes it does) I see that there's an error in the docstring for the preceeding option which I can fix whilst I'm at it:

sherpa

DougBurke: should be done but I don't want to resolve yet.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: I've made a change which might not be quite what you were thinking but does improve things a bit.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: I'll make a PR onto your branch with a suggestion on what and where to write into the docs. That's probably easier than me suggesting things here inside a test which is definitely where the text is NOT going.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: This is good. I usually don't care too much about DRY in tests, but I thought it might help future me (or future you) in this particular case.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: OK. I'll look at the docs to see if I should have known that.

DougBurke:
The tests fail because the read-the-docs builds fails. This because python 3.8 and the version of ipython used do not match (NEP 29) - which is fixed in #1735 - in particular https://github.com/sherpa/sherpa/pull/1753/commits/2527add0467f225474f4eecf50352b500a49c567

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
hamogu: https://github.com/DougBurke/sherpa/pull/4

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
DougBurke: Also note that there is meant to be more-exstensive doc changes in the follow-on PR #1684, which might be a better place to improve things.

hamogu: see https://github.com/DougBurke/sherpa/pull/4 - but if that's part of #1684 (which I have not looked at yet), that's fine, too.

DougBurke:
@hamogu - are you happy with this now?

hamogu (approved):

anetasie (approved):

PR #1735, opened 2023-03-29
Allow use of the NumPy random generator API

DougBurke:
# Summary

Take advantage of the NumPy random-number generator classes to provide random numbers, and move away from using the global state provided by numpy.random.seed and random.seed routines. Users of the UI layer use set_rng and get_rng to set and access the generator, and lower-level code uses a rng parameter to send the generator to the code. The default value for these is set to None, which means that it will still use the legacy NumPy random routines (so setting numpy.random.seed will change the results). To match results for old code that used numpy.random.seed(value) using the new API, use numpy.random.RandomState(value) to create a generator and set it with either set_rng or pass via the rng parameter (there are some cases where we can not guarantee an exact match with previous versions).

# Details

NumPy has for a while now has provided a "better" system for creating random numbers than using a global state via a seed value (it was introduced in NumPy 1.17 which is old enough that it's no-longer supported, so it should be safe to assume we have available). The new approach is to create a generator object and then use that to create random numbers. As noted in #867 we could use the new approach, so this is an attempt to do so. I note that I believe the "old" system is not going to be removed, so we could just make no change, but

- I think the new approach is "better"
- I think the old approach has subtle issues when running code in parallel (via multiprocessing) that the new approach can avoid

The approach taken is (from a bottom-up view)

- provide a sherpa.utils.random module to provide RNG routines that take a rng parameter
  - this might be a "temporary" module, as it's only really needed while we still want to support rng=None, which might be
    something we will want to deprecate, or we want to keep around (even if temporary it would last for a while)
- "low-level" routines that require a RNG will then accept a rng parameter (where low-level means everything below the UI layer)
- the ui layer adds set_rng and get_rng routines that take/return the random generator, so this acts like the set/get stat/method routines, rather than require a user pass around a random generator

I wanted to allow users to continue using the UI layer (at least) with the numpy.random.seed approach - i.e. it should still be possible to call numpy.random.seed(128623) and get the "same" result as you would have done before this PR. This is mostly supported, but it turns out that one module used the random module for generating random numbers, and not numpy.random, so in some cases it's not possible (but in most cases it is). To do this we allow the rng parameter to be set to None (and in fact, that's the default value) [**citation needed**].

For users who want to update their code but still get the same random numbers out (modulo the one case mentioned above about a module that used the random generator), they can use the numpy.random.RandomState class to create a generator. In fact, this is what many of our tests do - they are changed from

    np.random.state(seedvalue)

to either

    rng = np.random.RandomState(seedvalue)
    ui.set_rng(np.random.RandomState(seedvalue))

and the rest of the test has not had to change (although this change has let me remove the reset_seed text fixture).

The suggested use case is either

    rng = np.random.default_rng()
    ui.set_rng(np.random.default_rng())

(to get "true" random numbers, or send in a seed value to get repeatable - although not guaranteed to be repeatable with an updated NumPy library version - set of numbers).

I am actually quite happy with how this has come out, but there a few issues to think about

1. some existing routines send in a seed value as a way to control the random generator
2. how do we create numbers in parallel
3. thanks to the design of some classes, we occasionally seem to have some classes that need the rng parameter to be set when the class is created, and some when a call to that class is made

## 1 - seed parameter

These are mainly (perhaps exclusively) optimization methods, where the seed value is being sent into C++ scope to seed a RNG generator there. We do not want to be messing with that code (an alternative is to get the C++ code to call back into Python scope to access the RNG it wants, but there are many reasons not to do this), but do we want to use the rng generator to calculate the seed value to use, or do we want to just keep the seed argument?

This is relevant since we see the seed value when using the `moncar` optimiser:

```
sherpa-4.15.0> set_method("moncar")

sherpa-4.15.0> get_method_opt()
{'ftol': 1.1920928955078125e-07,
 'maxfev': None,
 'verbose': 0,
 'seed': 74815,
 'population_size': None,
 'xprob': 0.9,
 'weighting_factor': 0.8,
 'numcores': 1}
```

I think we want to leave the seed argument as is: the user can manually set seed however they want - e.g. via a generator if desired. This doesn't quite match the other parts of the API but I can live with that.

I need to review the existing commits to see how I left this. There is some existing support for using the RNG - with `seed=None` for at least one call it will generate a random seed, so having the ability to control the generator is good. However, the current approach leads to `rng` being a member of the optimizer options, which is worrying, as I'm not sure how it will be handled when set (in particular in save files and the like):

```
# current behavior of this branch
>>> ui.set_method("moncar")
>>> ui.get_method_name()
'moncar'
>>> for k, v in ui.get_method_opt().items():
...     print(f"{k:20s} '{v}'")
... 
ftol                 '1.1920928955078125e-07'
maxfev               'None'
verbose              '0'
seed                 '74815'
population_size      'None'
xprob                '0.9'
weighting_factor     '0.8'
numcores             '1'
rng                  'None'
```

## 2 - creating numbers in parallel

The existing code doesn't actually deal with this nicely - we just grab another random number from whatever the state is when we call the multiprocessing code - so we have a chance to do better. 

I have based my approach on https://numpy.org/doc/stable/reference/random/parallel.html which has lead to the `parallel_map_rng` call. I think this is only used in one place - the pvalue code - but I have managed to get myself confused about this so it needs checking. 

For most of our code we have called the random number generator before splitting up the code into chunks to be run in parallel, so the issues with the parallel code are not something to be worried about (at least for these cases).

# 3 - when do we pass in the rng parameter

I mention this just so I don't forget it, but at the moment I am okay with this dichotomy, but we should think about it as we review this PR.

---

# Historical notes

I have found one case in our tests where they failed because one CI run was using NumPy 1.18 and there was a change in NumPy 1.19 that affects the test output - [a change to how SeedSequence works](https://numpy.org/doc/stable/release/1.19.0-notes.html#seedsequence-with-small-seeds-no-longer-conflicts-with-spawning) which `sherpa.utils.parallel.parallel_map_rng` uses - so I've bumped the minimum version of NumPy used in the CI run to 1.19 (as we are well past it's EOL state, according to https://scientific-python.org/specs/spec-0000/). This is no-longer relevant with the 4.15.1 release as this has updated the CI runs to versions > 1.19.

Trying to break up into smaller commits so this can focus on the RNG changes

- #1770 
- #1771
  - #1776 
    - #1778 
- #1772 
  - #1777  
- #1773
- #1775


DougBurke:
Rebased to address recent conflicts, and to remove some workarounds that are no-longer needed. Currently chunking the commits to validate them.

DougBurke:
Rebased as merged

- #1770 
- #1771 
- #1775

DougBurke:
Rebased as merged

- #1773

DougBurke:
Rebased as merged

- #1776

DougBurke:
Rebased as merged

- #1734

DougBurke:
Rebased as merged

- #1778

sherpa/astro/flux.py
```diff
@@ -370,14 +375,17 @@ def decompose(mdl):
...
-                clip='hard'):
+                clip='hard', rng=None):
     """Calculate model fluxes from a sample of parameter values.
 
     Draw parameter values from a normal distribution and then calculate
     the model flux for each set of parameter values. The values are
     drawn from normal distributions, and the distributions can either
     be independent or have correlations between the parameters.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 386: Dang it. Need to update the version number,

DougBurke: The commit message for this PR also has a typo, so I must try to remember to change that too...

sherpa/sim/__init__.py
```diff
@@ -655,6 +656,9 @@ def get_draws(self, fit, sigma, niter=1000, cache=True):
         flag indicating whether the row represents a jump from the
         current location or not.
 
+        .. versionadded:: 4.15.1
```
DougBurke, line 659: Need to change this version number

sherpa/sim/__init__.py
```diff
@@ -839,12 +847,15 @@ def __init__(self, data, model):
...
-    def __call__(self, niter=1000, seed=None):
-        return self.call(niter, seed)
+    def __call__(self, niter=1000, seed=None, rng=None):
+        return self.call(niter, seed, rng=rng)
 
-    def call(self, niter, seed=None):
+    def call(self, niter, seed=None, rng=None):
         """Resample the data and fit the model to each iteration.
 
+        .. versionadded:: 4.15.1
```
DougBurke, line 856: Need to change this version

sherpa/sim/sample.py
```diff
@@ -81,22 +87,21 @@ def multivariate_t(mean, cov, df, size=None):
...
+    normal = random.multivariate_normal(rng, numpy.zeros_like(mean), cov, size=size)
+    x = numpy.sqrt(random.chisquare(rng, df, size=size) / df)
     numpy.divide(normal, x[numpy.newaxis].T, normal)
     numpy.add(mean, normal, normal)
     x = normal
     return x
 
 
-def multivariate_cauchy(mean, cov, size=None):
+# TODO: should pass the size value through
```
DougBurke, line 99: This should really say

> TODO: should this pass the size value through?

sherpa/sim/sample.py
```diff
@@ -748,18 +786,22 @@ def get_sample(self, fit, *, num=1, dof=2, numcores=None):
...
+def normal_sample(fit, num=1, sigma=1, correlate=True, numcores=None,
+                  rng=None):
     """Sample the fit statistic by taking the parameter values
     from a normal distribution.
 
     For each iteration (sample), change the thawed parameters by
     drawing values from a uni- or multi-variate normal (Gaussian)
     distribution, and calculate the fit statistic.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 802: version number to be changed

sherpa/sim/sample.py
```diff
@@ -803,17 +848,20 @@ def normal_sample(fit, num=1, sigma=1, correlate=True, numcores=None):
...
-def uniform_sample(fit, num=1, factor=4, numcores=None):
+def uniform_sample(fit, num=1, factor=4, numcores=None, rng=None):
     """Sample the fit statistic by taking the parameter values
     from an uniform distribution.
 
     For each iteration (sample), change the thawed parameters by
     drawing values from a uniform distribution, and calculate the
     fit statistic.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 862: version number to be changed

sherpa/sim/sample.py
```diff
@@ -840,17 +891,21 @@ def uniform_sample(fit, num=1, factor=4, numcores=None):
...
-def t_sample(fit, num=1, dof=2, numcores=None):
+def t_sample(fit, num=1, dof=2, numcores=None, rng=None):
     """Sample the fit statistic by taking the parameter values from
     a Student's t-distribution.
 
     For each iteration (sample), change the thawed parameters
     by drawing values from a Student's t-distribution, and
     calculate the fit statistic.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 906: version number to be changed

sherpa/sim/simulate.py
```diff
@@ -31,9 +31,12 @@
...
 from sherpa.estmethods import Covariance
-from sherpa.utils import parallel_map, poisson_noise, NoNewAttributesAfterInit
 from sherpa.fit import Fit
 from sherpa.sim.sample import NormalParameterSampleFromScaleMatrix
+from sherpa.utils import NoNewAttributesAfterInit
+from sherpa.utils.parallel import parallel_map_rng
+from sherpa.utils.random import poisson_noise
+
 
 logger = logging.getLogger("sherpa")
```
DougBurke, line 41: change to `getLogger(__name__)` or whatever it should be

sherpa/utils/random.py
```diff
@@ -20,6 +20,11 @@
 
 """
 Basic routines for handling random numbers.
+
+A number of routines have been added to support writing code that
+allows both the legaxy (pre 1.17) NumPy random API to be used
```
DougBurke, line 25: typo; I seem to like saying `legaxy`.

sherpa/optmethods/optfcts.py
```diff
@@ -529,6 +529,9 @@ def montecarlo(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, verbose=0,
     the pool of population, the resulting trial vector is selected if
     it lowers the objective function.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 532: update version number

DougBurke: However, there is already support for using a RNG when `seed=None` so it looks like we would ideally want to allow a `rng` parameter...

I'm just slightly concerned as it ends up leaking through to the method options, and I'm not sure about what would happen if set the method option to a generator:

```
>>> ui.get_method_name()
'levmar'
>>> ui.get_method_opt()
{'ftol': 1.1920928955078125e-07, 'xtol': 1.1920928955078125e-07, 'gtol': 1.1920928955078125e-07, 'maxfev': None, 'epsfcn': 1.1920928955078125e-07, 'factor': 100.0, 'numcores': 1, 'verbose': 0}
>>> ui.set_method("moncar")
>>> ui.get_method_name()
'moncar'
```

Note that the following has a `rng` field:

```
>>> ui.get_method_opt()
{'ftol': 1.1920928955078125e-07, 'maxfev': None, 'verbose': 0, 'seed': 74815, 'population_size': None, 'xprob': 0.9, 'weighting_factor': 0.8, 'numcores': 1, 'rng': None}
```

sherpa/fit.py
```diff
@@ -880,6 +880,9 @@ def fit(self, statfunc, pars, parmins, parmaxes,
 class Fit(NoNewAttributesAfterInit):
     """Fit a model to a data set.
 
+    .. versionchanged:: 4.15.1
```
DougBurke, line 883: version number, if we keep

sherpa/ui/utils.py
```diff
@@ -973,6 +973,53 @@ def clean(self):
...
+
+        """
+
+        return self._rng
+
+    def set_rng(self, rng):
+        """Set the RNG generator.
+
+        .. versionadded:: 4.16.0
+           This replaces the seed argument for certain routines and
```
DougBurke, line 1001: Maybe it replaces the need for the seed parameter, need to review. Seed is used for at least

- moncar method options, which we might want to keep
- resampledata stuff, which we probably want to deprecate
- anything else???

sherpa/ui/utils.py
```diff
@@ -973,6 +973,53 @@ def clean(self):
...
+           the need to call np.random.seed directly in others.
+
+        Parameters
+        ----------
+        rng : numpy.random.Generator or None
```
DougBurke, line 1006: We could require that `rng` be a generator, but this would make it hard for users to match the code to previous versions which used `np.random.seed` to set a "known" random state. The answer could be `you need to change np.random.seed(xxx) to ui.set_rng(np.random.RandomState(xxx))` but this is not "user-friendly".

sherpa/sim/simulate.py
DougBurke: This is part of #1758 and likely better handled separately.

DougBurke:
Rebased as merged both

- #1777
- #1807

In doing so fixed up a few issues noted above.

sherpa/optmethods/optfcts.py
DougBurke: Note: the version number has now been updated to `4.16.0` but the rest remains a valid question.

DougBurke:
@hamogu - now the pre-requisites have been merged it's worth looking at (as it now should just be RNG focussed with little else in it). There are some issues I'm not sure about yet - mainly 

- the fit/optmethods interfaces where maybe we just want to stay with  `seed` argument,
- is `set_rng` a good approach for the UI code (I claim it is, given how we handle stats/method/...) but it's worth thinking about
- what should the deprecation-cycle be (from switching over from have `None` as the default to `np.random.default_rng()`), or do we think this may never happen

but overall I'm pretty happy with how this looks.

DougBurke:
The test failure is a bizarre RTD error and nothing to do with the changes here...

DougBurke:
Added "CI: arch" label so we can test on aarch64, just in case.

DougBurke:
TODO: should think about whether we want to store the state in `save_all` output if set.

hamogu:
I'm leaving general comments here while a I look at this. Not all of them might turn out to be relevant, but I want to write them down so I don't forget:

- I'm OK if different version of Sherpa/numpy give different random numbers. Of course, it's nice if the numbers one gets are consistent about as many Sherpa/numpy combinations as possible, but I think we *don't have to* guarantee that, as long as one gets the exact same answer when using te same seed with the same numpy and Sherpa version.

hamogu (changes requested):

sherpa/optmethods/ncoresde.py
```diff
@@ -326,11 +344,21 @@ class ncoresMyDifEvo(MyDifEvo):
...
+        #
+        if self.rng is None:
+            numpy.random.seed(self.seed)
+
         mypop = self.polytope
         old_fval = numpy.inf
         while nfev < maxnfev:
 
+            # all_strategies has been set up so that each strategy has
+            # it's own generator, so it's okay to use parallel_map here,
```
hamogu, line 359:
```suggestion
            # its own generator, so it's okay to use parallel_map here,
```

sherpa/optmethods/optfcts.py
```diff
@@ -573,6 +576,9 @@ def montecarlo(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, verbose=0,
        efficiency.
     numcores : int
        The number of CPU cores to use. The default is `1`.
+    rng : np.random.Generator, np.random.RandomState, or None, optional
```
hamogu, line 579:
```suggestion
    rng : `np.random.Generator`, `np.random.RandomState`, or None, optional
```

sherpa/sim/__init__.py
```diff
@@ -664,6 +668,10 @@ def get_draws(self, fit, sigma, niter=1000, cache=True):
            values.
         niter : int, optional
            The number of draws to use. The default is ``1000``.
+        rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 671:
```suggestion
        rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/__init__.py
```diff
@@ -856,7 +867,11 @@ def call(self, niter, seed=None):
         niter : int
             The number of iterations.
         seed : int or None, optional
-            The seed value.
+            The seed value (only used if rng is None).
+        rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 871:
```suggestion
        rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/mh.py
```diff
@@ -80,35 +83,47 @@ class CovarError(Exception):
...
-    returns a sample from the multivariate t distribution in the shape of `mu`
+    Parameters
+    ----------
+    mu
+       The current sample.
+    sigma : 2-D array_like, of shape (N, N)
+       The covariance matrix.
+    dof : int
+       The degrees of freedom.
+    rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 100:
```suggestion
    rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/sample.py
```diff
@@ -66,6 +69,9 @@ def multivariate_t(mean, cov, df, size=None):
         generated, and packed in an `m`-by-`n`-by-`k` arrangement.  Because
         each sample is `N`-dimensional, the output shape is ``(m,n,k,N)``.
         If no shape is specified, a single (`N`-D) sample is returned.
+    rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 72:
```suggestion
    rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/sample.py
```diff
@@ -310,6 +315,9 @@ def get_sample(self, fit, *, num=1):
             the samples, along with any possible error analysis.
         num : int, optional
             The number of samples to return.
+        rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 318:
```suggestion
        rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/sample.py
```diff
@@ -413,6 +421,9 @@ def get_sample(self, fit, *, factor=4, num=1):
             the one-sigma error.
         num : int, optional
             The number of samples to return.
+        rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 424:
```suggestion
        rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

sherpa/sim/sample.py
```diff
@@ -452,6 +466,9 @@ def get_sample(self, fit, *, myscales=None, num=1):
             it is calculated from the fit object.
         num : int, optional
             The number of samples to return.
+        rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
```
hamogu, line 469:
```suggestion
        rng : `numpy.random.Generator`, `numpy.random.RandomState`, or None, optional
```

hamogu: OK, enough of these comment. First, please try one and see if that actually works for make the links html in the RTD and then a global "search and replace" on the entire codebase is going to be way more efficient since the line comes up many, many times, sometimes with no and sometimes with numpy. (Also wit hand without optional, see some of my other comments). That still only two search-and-replace operations. Why do I do the work that a computer can do better and faster?

sherpa/ui/utils.py
hamogu: I agree with this assessment. Keep for moncar, get rid of it for everything else.

sherpa/ui/utils.py
hamogu: Why is that not user-friendly? Because it's not automatic?

sherpa/ui/utils.py
```diff
@@ -9422,6 +9470,11 @@ def get_pvalue_plot(self, null_model=None, alt_model=None, conv_model=None,
         get_pvalue_results : Return the data calculated by the last plot_pvalue call.
         plot_pvalue : Compute and plot a histogram of likelihood ratios by simulating data.
 
+        Notes
+        -----
+        The random numbers are generated using the generator returned
+        by`get_rng`.
```
hamogu, line 9476:
```suggestion
        by `get_rng`.
```

sherpa/ui/utils.py
```diff
@@ -11379,6 +11433,9 @@ def get_draws(self, id=None, otherids=(), niter=1000, covar_matrix=None):
         running `get_draws`. The results from `get_draws` is used to
         estimate the parameter distributions.
 
+        The random numbers are generated using the generator returned
+        by`get_rng`.
```
hamogu, line 11437:
```suggestion
        by `get_rng`.
```

hamogu:
I don't have many thoughts on the big picture questions @DougBurke brought up above. Either way is good for me. The only thing that's slightly worrying to me is that it's not clear if moncar always starts wit the same seed, even if the seed of None is set. IT seems tat it did before and fits usually work. Any why would they not? Any seed should be good. So, I'm OK with having a fixed seed in the preferences for reproducibility, we just need to make sure it's clearly spelled out. (On the other hand, I also don't care that much if a moncar optimization is absolutely repeatable. It might differ by numpy version anyway. If my fit depends sensitively on the starting seed, I should be worried.)

sherpa/ui/utils.py
DougBurke: Because it requires users to change code that used to work.

sherpa/astro/sim/pragbayes.py
```diff
@@ -42,10 +43,10 @@
 
 class ARFSIMFactory():
 
-    def __call__(self, filename):
-        return self.read(filename)
+    def __call__(self, filename, rng=None):
```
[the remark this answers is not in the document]
DougBurke: It would be nice if this code had any documentation.

sherpa/utils/random.py
```diff
@@ -0,0 +1,290 @@
...
+    else:
+        simulate = rng.poisson
+
+    x = np.asarray(x)
+    if x.shape == ():
+        x = SherpaFloat(x)
+        if x <= 0.0:
+            x = 0.0
+        else:
+            x = x = simulate(x)
```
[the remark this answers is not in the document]
DougBurke: Because it looked cool? Will fix (but fortunately, thanks to the way Python assignment works, it isn't a bug [note added for anyone reading along at home]).

sherpa/utils/random.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def random(rng):
+    """Create a random value [0, 1.0)
+
+    Parameters
+    ----------
+    rng : np.random.Generator, np.random.RandomState, or None, optional
+        If set, the generator is used to create the random numbers. If
+        not set then the legacy numpy RandomState instance is used.
```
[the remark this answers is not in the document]
DougBurke: I struggled to explain some of this when writing, so thanks for looking Will rework.

sherpa/astro/sim/fullbayes.py
```diff
@@ -21,17 +21,18 @@
...
 
 __all__ = ('FullBayes',)
 
 
 class FullBayes(PragBayes):
 
-    def __init__(self, fcn, sigma, mu, dof, fit, *args):
-        PragBayes.__init__(self, fcn, sigma, mu, dof, fit, *args)
+    def __init__(self, fcn, sigma, mu, dof, fit, *args, rng=None):
+        PragBayes.__init__(self, fcn, sigma, mu, dof, fit, *args, rng=rng)
```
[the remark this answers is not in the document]
DougBurke: I don't think it's the case here (I think that this class structure is pretty vanilla) but there are some classes where the advice from [super used to be harmful but is now just problematic](https://fuhm.net/super-harmful/) is true, so I've slowed down in my general move to using super wherever I see it.

Also, this PR is big enough as is...

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -2877,13 +2877,12 @@ def custom(x):
 
 @pytest.mark.parametrize("session", [Session, AstroSession])
 @pytest.mark.parametrize("idval", [None, "bob"])
-def test_fake_random(session, idval, reset_seed):
+def test_fake_random(session, idval):
```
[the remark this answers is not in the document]
DougBurke: There is a commit that removes `reset_seed`.

sherpa/optmethods/ncoresde.py
```diff
@@ -378,13 +412,21 @@ def __call__(self, fcn, x, xmin, xmax, step=None, maxnfev=None, tol=1.0e-6,
         if maxnfev is None:
             maxnfev = 8192 * npar
 
+        # TODO: this over-writes the seed argument, which looks wrong
```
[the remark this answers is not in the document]
DougBurke: I think it depends on why we have random-ness here. I know I had been strident in the past on not understanding random-ness when fitting (under that assumption that the optimiser had no randomness) but that's obviously wrong here, so maybe this was just added to shut users up.

This code is complex enough to understand that I am not confident in making changes without a lot of thought and discussion, so I'd rather leave this in and make an issue of it.

sherpa/sim/sample.py
DougBurke: To be honest, I am not sure it's needed. I don't understand the rules of when the expansion works, but here is the output I get for the `sherpa.sim.sample.NormalParameterSampleFromScaleMatrix.get_sample` routine

image

and hee is the docstring that created this:

```
 489   │ class NormalParameterSampleFromScaleMatrix(ParameterSampleFromScaleMatrix):
 490   │     """Use a normal distribution to sample parameters (correlated),
 491   │ 
 492   │     The parameters are drawn from a normal distribution based on the
 493   │     parameter errors, and include the correlations between the
 494   │     parameters. The errors will be generated from the fit object or
 495   │     specified directly as a covariance matrix.
 496   │ 
 497   │     """
 498   │ 
 499   │     def get_sample(self, fit, *, mycov=None, num=1, rng=None):
 500   │         """Return the parameter samples.
 501   │ 
 502   │         Parameters
 503   │         ----------
 504   │         fit : sherpa.fit.Fit instance
 505   │             This defines the thawed parameters that are used to generate
 506   │             the samples, along with any possible error analysis.
 507   │         mycov : 2D numpy array or None, optional
 508   │             The covariance matrix to use. If None then it is
 509   │             calculated from the fit object.
 510   │         num : int, optional
 511   │             The number of samples to return.
 512   │         rng : numpy.random.Generator, numpy.random.RandomState, or None, optional
 513   │             If set, the generator is used to create the random numbers. If
 514   │             not set then the legacy numpy RandomState instance is used.
 515   │ 
 516   │         Returns
 517   │         -------
 518   │         samples : 2D numpy array
 519   │             The array is num by npar size, where npar is the number of
 520   │             free parameters in the fit argument.
 521   │ 
 522   │         """
 523   │         vals = numpy.array(fit.model.thawedpars)
 524   │         cov = self.scale.get_scales(fit, mycov)
 525   │         return random.multivariate_normal(rng, vals, cov, size=int(num))
 526   │ 
```

so you can see that (at least on my computer, unfortunately the RTD version for this PR is currently broken as it had some internal error, but hopefully it'll work soon after I fix a conflict and some of the comments so we can check there).

DougBurke: (e.g. `sherpa.fit.Fit` hasn't been picked up here as a link; is the "intersphnix inventory" changes you made a while back that makes it pick up the `numpy.xxx` symbols?

sherpa/utils/random.py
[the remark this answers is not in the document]
DougBurke: My problem is that "If not set then the legacy numpy RandomState" text is technically incorrect (this may be an issue you raise elsewhere) - what is actually used are the numpy.random.xxx routines, so maybe just "legacy numpy random routine" would be better (and then should it label the actual routine, which we can do in this case ?)

sherpa/optmethods/ncoresde.py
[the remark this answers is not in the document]
DougBurke: See https://github.com/sherpa/sherpa/issues/1862

sherpa/sim/sample.py
DougBurke: So, if the text says `numpy.<symbol>` then it should work (for valid symbols) but if the text is `np.<symbol>` then it won't (it becomes something that looks like it could be a link [bold fixed font] but is not guaranteed to be a link).

[this comment is for my own benefit]

DougBurke: So my current idea is not to make these changes, as we don't need them, but let me know if I'm missing a use case where they are useful (or some other reason)

sherpa/utils/tests/test_random.py
```diff
@@ -0,0 +1,165 @@
...
+    """Send in an array, get back an array"""
+
+    out = poisson_noise([1001, 1002, 0.0, 1003, -1004])
+    assert type(out) == np.ndarray
+    assert out.dtype.type == SherpaFloat
+
+    # As the RNG is not set here, technically we could get zeros for
+    # the positive expected values, but this is very unlikely.
+    #
+    ans = np.flatnonzero(out > 0.0)
```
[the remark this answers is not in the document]
DougBurke: but it's technically possible .....

DougBurke:
I've merged in recent changes, rather than rebasing, to try and make @hamogu 's life a little easier, and addressed some of the documentation changes. More to come.

sherpa/sim/sample.py
DougBurke: Here's the RTD version showing the same behavior: https://sherpa--1735.org.readthedocs.build/en/1735/mcmc/api/sherpa.sim.sample.NormalParameterSampleFromScaleMatrix.html#sherpa.sim.sample.NormalParameterSampleFromScaleMatrix.get_sample (just to check it wasn't something magic about my system)

sherpa/utils/random.py
```diff
@@ -0,0 +1,290 @@
...
+    Returns
+    -------
+    out : scalar or array
+       A random realisation of the input array, drawn from
+       the Poisson distribution, as a `SherpaFloat`.
+
+    Notes
+    -----
+    The distribution is calculated by the NumPy poisson generator
+    and all values zero or less are replaced by zero.
```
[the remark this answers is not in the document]
DougBurke: I've re-worked this just to say that all values less than 0 are replaced by 0 and not said anything about the generator. YMMV whether this is an improvement or not.

sherpa/utils/parallel.py
```diff
@@ -157,6 +159,41 @@ def worker(f, ii, chunk, out_q, err_q):
...
+    chunk : sequence
+        The elements to be passed to func.
+    out_q, err_q : manager.Queue
+        The success channel - which will be sent (idx, retval) on
+        success - and the error channel, which will be sent any
+        exception.
+    rng : np.random.Generator, np.randomRandomState, or None, optional
+       If set, the generator is used to create the random numbers. If
+       not set then the legacy numpy RandomState instance is
+       used. Each parallel worker is sent a separate RNG, to ensure
```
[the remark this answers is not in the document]
DougBurke: I have re-worked some of the text in this module. I hate to say "improved" because I'm not sure...

sherpa/astro/ui/tests/test_astro_session.py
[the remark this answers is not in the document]
DougBurke: See 0d236f3118106fd035f7d56de3950df472890972

sherpa/sim/sample.py
DougBurke: I'll close the other "make them links" comments and leave this open as a reminder

sherpa/astro/fake.py
```diff
@@ -86,12 +87,16 @@ def fake_pha(data, model,
...
-        used for certain error messages.
+        String with id number if called from UI layer. This is only used for
+        certain error messages.
     method : callable or None
         The routine used to simulate the data. If None (the default)
         then sherpa.utils.poisson_noise is used, otherwise the function
-        must accept a single ndarray, returning a ndarray of the same shape.
+        must accept a single ndarray, returning a ndarray of the same shape,
+        and an optional rng argument.
+    rng : numpy.random.Generator or None, optional
```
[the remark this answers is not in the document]
DougBurke: Hmm, this can also take numpy.random.RandomState

sherpa/sim/sample.py
hamogu: > is the "intersphnix inventory" changes you made a while back that makes it pick up the numpy.xxx symbols

I think so.

hamogu: Maybe I was more successful than I had thought. It's a good think to get rid of the extra backticks, but I can't say I understand when they are required either. I think in the astropy package that we pull in the name "numpy" might be special-cased to work without backticks.

DougBurke: It's not ideal that we have some special-cased text here, but I just don't want to add the backticks here if I can help it ;-)

sherpa/astro/sim/fullbayes.py
```diff
@@ -21,17 +21,18 @@
...
 
 
 __all__ = ('FullBayes',)
 
 
 class FullBayes(PragBayes):
 
-    def __init__(self, fcn, sigma, mu, dof, fit, *args):
-        PragBayes.__init__(self, fcn, sigma, mu, dof, fit, *args)
+    def __init__(self, fcn, sigma, mu, dof, fit, *args, rng=None):
```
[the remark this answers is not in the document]
DougBurke: But yes, it's woefully undocumented.

DougBurke:
[still working through comments so not yet ready for review again]

sherpa/utils/random.py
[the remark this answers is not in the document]
DougBurke: Hopefully the new text is better.

DougBurke:
I've made a `plot_pvalue` use a fixed seed -this was code I originally wrote in #1757 but is not specific to that PR and makes sense to add here (hopefully reducing the run-time ever-so-slightly whilst being more repeatable).

I've also merged in the current main branch so the tests can be run.

DougBurke:
Turns out `plot_pvalue` isn't as repeatable as we'd like (but at least I understand why).

hamogu:
Let me know if oyu want to to review this to get it in soon, but maybe it's not urgent?

DougBurke:
I'm just working on what I think is my final issue. I'll let you know when it's ready. Ta.

DougBurke:
I am close to having this ready for review. The main changes is 

- no need to send a rng to the fit class since it's only needed for the method, and that is handled separately/directly
- for the UI level there's some complication as we need a rng for the `montecarlo` setting, so we need to tie that to the `set_rng` call (this is not ideal and I have some more thinking to do about this part of the interface)

sherpa/utils/random.py
```diff
@@ -0,0 +1,290 @@
...
+
+A number of routines have been added to support writing code that
+allows both the legacy (pre 1.17) NumPy random API to be used
+(with rng set to None) or the new generator approach.
+
+"""
+
+import numpy as np
+
+# Copy of sherpa.utils.SherpaFloat
```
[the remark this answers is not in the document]
DougBurke: We now have a `sherpa.utils.numeric_types` module just to kee[p `SherpaFloat`, `SherpaInt`, and `SherpaUInt` warm and cozy.

DougBurke:
I've merged in the existing main branch to fix the conflicts and added documentation for the new `numeric_types` module, but that caused my version of sphinx no-end of warning/bleating, so let's see what happens here.

Unfortunately I **have not** looked at the changes I mentioned in https://github.com/sherpa/sherpa/pull/1735#issuecomment-1730043484

DougBurke:
I'm not sure if the documentation build problem is related to https://github.com/sphinx-doc/sphinx/issues/7844 but I am just going to remove the `numeric_types` module from the documentation for now.

DougBurke:
So, I have

- reverted the addition of the `sherpa.utils.numeric_types` module **just** for the RTD documentation as it is not-at-all-clear-to-me what the warning message is really doing and it is not vitally important that we have this module of three symbols documented
- replace the above to at least mention the symbols and their location, even if they are no-longer actual links
- fix a typo

I have thought about the issues I mention above - https://github.com/sherpa/sherpa/pull/1735#issuecomment-1730043484 - and I have decided that the current approach is okay for now, and we can review this behaviour once we have more experience with it.

So I claim it's ready for re-review by @hamogu

docs/overview/utilities.rst
```diff
@@ -12,6 +12,44 @@ There are a number of utility routines provided by Sherpa that may
...
+  (`sherpa.sim.simulate`);
+- when sampling distributions using routines from
+  `sherpa.sim.sample`;
+- and  when simulating data, such as `~sherpa.ui.fake` and
+  `~sherpa.astro.ui.fake_pha`.
+
+The primary way to control the random numbers used by Sherpa is to
+create a `NumPy Random Generator
+<https://numpy.org/doc/stable/reference/random/generator.html>`_ and
+pass it to routines wither with the ``rng`` parameter or, for users of
```
anetasie, line 66: typo here 'wither with...'?

sherpa/astro/fake.py
```diff
@@ -143,9 +148,9 @@ def fake_pha(data, model,
     rmf0 = data.get_rmf(data.response_ids[0])
     data.channel = np.arange(1, rmf0.detchans + 1)
 
-    # Calculate the source model, and take a Poisson draw based on
+    # Calculate the source model, and take a "draw" based on
```
anetasie, line 152: why 'Poisson draw' was changed here? What is the distribution for the assumed draw otherwise?

DougBurke: Because we can now pass in a function to do this, so it need not be a poisson distribution. The obvious choice is the identity function (ie adds no noise) as that is a good way to check the results,  but this is only when you want to check it's doing the "right" thing. I expect essentially everyone to use the default (which is poisson)

anetasie:
@DougBurke that was a lot of work! I ran a few examples, including plot_pvalue and get_draws with this PR. It seems to work, but I have not touched the `rng` setting.

sherpa/astro/fake.py
anetasie: That would be a good example in our docs we should add later.

DougBurke:
@anetasie - thanks for checking. It's useful to check the behaviour without explicitly calling `set_rng` as that is going to be the case for most users.

docs/overview/utilities.rst
```diff
@@ -41,6 +41,45 @@ model:
...
```
DougBurke, line 66:
```suggestion
pass it to routines either with the ``rng`` parameter or, for users of
```

DougBurke:
Argh - conflict. Just checking the merge to fix this (fortunately an easy fix).

DougBurke:
@hamogu - have I missed anything you wanted here? It's valid to say "no time" !

sherpa/utils/tests/test_random.py
```diff
@@ -0,0 +1,165 @@
...
+        a = standard_normal(rng=None, size=SIZE)
+        b = standard_normal(rng=rng, size=SIZE)
+    finally:
+        np.random.seed()
+
+    assert a == pytest.approx(b)
+
+
+def test_chisquare_rng_is_none():
+    """Coverage shows this was not tested, so check it
```
hamogu, line 125: Similarly here. I suggest to decribe wha'ts tested, not why you added it and how ceverage works or doesn't work.

sherpa/utils/tests/test_random.py
```diff
@@ -0,0 +1,165 @@
...
+        a = chisquare(rng=None, df=DF, size=SIZE)
+        b = chisquare(rng=rng, df=DF, size=SIZE)
+    finally:
+        np.random.seed()
+
+    assert a == pytest.approx(b)
+
+
+def test_choice_rng_is_none():
+    """Coverage shows this was not tested, so check it
```
hamogu, line 147: same.

hamogu (approved):
I'm sorry I did not get to it faster. All my concerns have been addressed, but a few suggestions for doctrings. I'm approving already, since those are inconsquential and there is little harm done if this is merged without addressing them.

DougBurke:
I created #1915 for the `save_all` question.

sherpa/ui/tests/test_plot_pvalue.py
```diff
@@ -66,12 +75,14 @@ def test_plot_pvalue(make_data_path, clean_astro_ui, hide_logging):
     g1.fwhm = 0.1
     ui.freeze(g1.fwhm)
 
-    # Could we reduce the number of bins to save evaluation time?
-    # We do want a non-default num value when checking the shapes
-    # of the output attributes.
+    # Pick a small number of bins to try to reduce the runtime of the
+    # test whist still exercising the code.
```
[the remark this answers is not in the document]
DougBurke: I guess `while` is good enough.

DougBurke, line 79:
```suggestion
    # test while still exercising the code.
```

hamogu:
I think all important issues I had are addressed and I did indeed approve this before (Dough just fixed a typo after - no need for me to review again).

sherpa/utils/tests/test_random.py
```diff
@@ -0,0 +1,165 @@
...
+    rng = np.random.RandomState(43)
+    out = poisson_noise([12, 4.5, 0.0, 8.2, -1], rng=rng)
+    assert type(out) == np.ndarray
+    assert out.dtype.type == SherpaFloat
+
+    assert out == pytest.approx([7, 2, 0, 8, 0])
+
+
+def test_standard_normal_rng_is_none():
+    """Coverage shows this was not tested, so check it
```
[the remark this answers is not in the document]
DougBurke: I am going to say these can be addressed in a later PR.

sherpa/utils/tests/test_random.py
DougBurke: ditto.

sherpa/utils/tests/test_random.py
DougBurke: ditto.

Issue #1631, opened 2022-11-21
lazy loading of modules


DougBurke:
This is a long-term issue bit we can start looking at it once python 3.12 comes out and `numpy` builds against it.

Python 3.12 is adding support for lazy loading. We currently have an "interesting" set of imports and some of that can probably be cleaned up by the lazy import work (but we can only take advantage of this once python 3.12 is the minimum supported version!). 

We can however, try out the lazy import - once a python 3.12 + numpy is available - and check

a) does it work
b) do we see any improvements (in loading sherpa)
c) are there cases where the lazy loading needs to be turned off.

We have some modules where we import things and rely on their side effects. Most of them are "can this module be loaded" and so are within a `try/except` block, and hence are inherently `not-lazy`, but I still think there are gremlins. I'd also like our imports to be a bit more direct (e.g. importing `bar` from `sherpa.foo` rather than relying on it being imported fro `sherpa/foo/__init__.py`).

PR #1740, opened 2023-04-06
Make doctestplus a tests_require dependency

hamogu:
#Summary
Add `pytest-doctestplus` to the requirements for tests.

Effectively, this will make our doctests run by default;
currently that is not very many because we've explicitly ignored most files not
fixed in formatting yet, but at least it prevents regressions for those files are *are* ready to be tested.

#Details
We merged lazy loading of modules #1631 which allows us to run doctestplus
(if installed) locally and test that the examples in docstrings and sphinx docs
work, which at the same time uses them as separate tests.
Since that works, this PR explores making doctestplus a testing requirement,
which would install it on also on every CI run and thus
get the doctests checked in CO as well.


hamogu:
We are in fact running our activate (i.e. not blacklisted) doctests on the current main e.g. here: https://github.com/sherpa/sherpa/actions/runs/5094875523/jobs/9159210413?pr=1785#step:9:55 

because I added doctestplus to the tests-require a few months ago for a different reason: https://github.com/sherpa/sherpa/commit/3a2697fa2b5c7bcc0ef1e82f1fdfa5e9bd83b543

That means that this PR is in fact not needed.

PR #1599, opened 2022-09-28
Update Makefile.in and configure as generated by new autotools versions

hamogu:
I'm using automake (GNU automake) 1.16.5 and autoconf (GNU Autoconf) 2.71

I tried to install Sherpa from source on MacOSX 12.5.1 with compilers installed through XCode as "pip install -e ." The pip script runs configure and that prompted me to re-generate the input files with updated version of autotools. I did that and the install completed correctly.

In this PR, I add resulting configure and Makefile.in files.

PR #1741, opened 2023-04-06
extern updates (ciao 4.15 region changes, fftw 3.3.10)

wmclaugh:
**Summary**
Update the sherpa externs directory to fold in the latest stack, group, and region library changes (CIAO 4.15) as well as upgrade from fftw 3.3.9 to 3.3.10.

Details
Folded in changes to region lib code to synch up with CIAO 4.15 baseline. This includes a single change to use const char*in the function declaration for regYYerror. There were no new changes to group or stack library code. 

The fftw version has been changed from 3.3.9 to 3.3.10. This includes removing the fftw-3.3.9 directory and adding an fftw-3.3.10 directory with the accompanying change in configure.ac. Also changed in configure.ac is 'sherpa-dependencies' which is now 4.15 (formally 4.13).

NOTE: These changes are based on automake 1.16.5 and autoconf 2.71 to go with https://github.com/sherpa/sherpa/pull/1599.


DougBurke:
> The diff you're trying to view is too large. We only load the first 3000 changed files.

wmclaugh:
Unfortunately, the changed file set size was a side effect of renaming the fftw-3.3.9 directory to fftw-3.3.10:

    (py38) 94>git diff 4e33a99a ciao415_extern_updates --name-status --output /tmp/git_diffs
    (py38) 95>grep "R100" /tmp/git_diffs  | grep "3.3.9" | grep "3.3.10" | head  -1
    R100    extern/fftw-3.3.9/AUTHORS       extern/fftw-3.3.10/AUTHORS
    (py38) 96>grep "R100" /tmp/git_diffs | grep "3.3.9" | grep -c "3.3.10" 
    2878
   (py38) 97>grep "R0" /tmp/git_diffs | grep "3.3.9" | grep -c "3.3.10"
612

2878 files are identical aside from the directory name change. 612 of the files were modified/differ in the 3.3.9 and 3.3.10 versions.

Any suggestions on how to change the directory name without the extensive diffs welcome. I'd prefer to keep the subdir name appropriate to the version being utilized.

Issue #1746, opened 2023-04-14
is the plot_pvalue stuff working?


DougBurke:
I'm trying to update our tests to actually test the simulation stuff (at present they often just check the code runs). This is complicated for the simulation stuff since it often depends on random numbers, particularly when running with multiple cores.

The following example is basically the test data from `sherpa/utils/tests/test_sim.py`:

```python
import numpy as np

from sherpa.astro import ui

x = np.arange(0.1, 10.1, 0.1)
y = np.array(
    [114, 47, 35, 30, 40, 27, 30, 26, 24, 20, 26, 35,
     29, 28, 34, 36, 43, 39, 33, 47, 44, 46, 53, 56,
     52, 53, 49, 57, 49, 36, 33, 42, 49, 45, 42, 32,
     31, 34, 18, 24, 25, 11, 17, 17, 11,  9,  8,  5,
     4, 10,  3,  4,  6,  3,  0,  2,  4,  4,  0,  1,
     2,  0,  3,  3,  0,  2,  1,  2,  3,  0,  1,  0,
     1,  0,  0,  1,  3,  3,  0,  2,  0,  0,  1,  2,
     0,  1,  0,  1,  1,  0,  1,  1,  1,  1,  1,  1,
     1,  0,  1,  0]
    )
err = np.ones(100) * 0.4


ui.set_method("levmar")
ui.set_stat("chi2datavar")

ui.load_arrays(1, x, y)
ui.set_staterror(1, err)

g1 = ui.create_model_component("gauss1d", "g1")
p1 = ui.create_model_component("powlaw1d", "p1")

ui.set_source(g1 + p1)
ui.fit()

# switch to cstat/neldermead to match what test_lrt uses
ui.set_stat("cstat")
ui.set_method("neldermead")

num = 1000

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=1, recalc=True)
r1 = ui.get_pvalue_results()

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=2, recalc=True)
r2 = ui.get_pvalue_results()

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=4, recalc=True)
r4 = ui.get_pvalue_results()

print("---------------")
```

[...]

PR #1747, opened 2023-04-18
reduce chance of plot_pvalue getting stuck in a bad parameter space 

DougBurke:
# Summary

Ensure that each simulation step for the likelihood ratios code, as used by plot_pvalue, starts with the same parameter values to reduce the chance that the results can get stuck in certain areas of parameter space. The ratios field of the LikelihoodRatioResults structure - e.g. the return value of get_pvalue_results() - can be plotted to see if this has happened (e.g. large ranges where the ratio is close to zero.

# Details

Does resetting the fit after each iteration address #1746 ?

This includes #1751 to get the tests to run. Although as #1755 has now been merger we should remove this with a rebase.

First we need to enhance the simulation tests to actually check the return values, as previously all they did was check the code didn't error out, not that they actually did anything sensible. So the tests have been augmented to check the return values. Things to note are

- this is a regression test, not a test from first principles, as I did not write this code or have anything to do with it so I have no idea what the output should, just checking that we get the same results
- some tests are able to provide some "meaningful" tests, as the routine just adds an extra column to that returned by other routines, so we can use the same test/expected data for both and just filter out what we need
- these tests rely on random numbers, so we explicitly set the seed to a known value, **but**
  - I want to change how we handle random numbers and so want to pass a generator to these routines, but this is a significantly larger task (and doing this is where I found #1746) so I want to get this issue addressed first, and we can worry about changing this code to use a generator in later code (eg #1735)
  - for most tests we can rely on the numpy "guarantee" that we get the same results across machines/platforms (at least with the legacy API, which is what we currently are using), but the likelihood-ratio test is different, since this combines random numbers with fits, and the fit code can lead to differences, in particular between macOS and linux. So the lrt test is very "abbreviated" (in that I only check the first 3 ratios rather than all 25) as otherwise the differences get too large and it's hard to verify on different platforms
  - for the lrt tests an additional complication is that the test results depends on
    - the numcores argument
    - whether the system allows multiprocessing
   so making a test that can run on GH actions (where multiprocessing is always? off) and on my desktop machine complicates things

The actual fix is really small and well contained. There are different places we could make this, but I have tried to go for as safe as possible (i.e. do the fit and reset the parameter values, even for the null model) just to make it clear what is going on, and not to have to worry about whether changes to other code will change this behavior.


DougBurke:
So, if I use the example from #1746 - namely 

```
import numpy as np

from sherpa.astro import ui

x = np.arange(0.1, 10.1, 0.1)
y = np.array(
    [114, 47, 35, 30, 40, 27, 30, 26, 24, 20, 26, 35,
     29, 28, 34, 36, 43, 39, 33, 47, 44, 46, 53, 56,
     52, 53, 49, 57, 49, 36, 33, 42, 49, 45, 42, 32,
     31, 34, 18, 24, 25, 11, 17, 17, 11,  9,  8,  5,
     4, 10,  3,  4,  6,  3,  0,  2,  4,  4,  0,  1,
     2,  0,  3,  3,  0,  2,  1,  2,  3,  0,  1,  0,
     1,  0,  0,  1,  3,  3,  0,  2,  0,  0,  1,  2,
     0,  1,  0,  1,  1,  0,  1,  1,  1,  1,  1,  1,
     1,  0,  1,  0]
    )
err = np.ones(100) * 0.4


ui.set_method("levmar")
ui.set_stat("chi2datavar")

ui.load_arrays(1, x, y)
ui.set_staterror(1, err)

g1 = ui.create_model_component("gauss1d", "g1")
p1 = ui.create_model_component("powlaw1d", "p1")

ui.set_source(g1 + p1)
ui.fit()

# switch to cstat/neldermead to match what test_lrt uses
ui.set_stat("cstat")
ui.set_method("neldermead")

num = 1000

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=1, recalc=True)
r1 = ui.get_pvalue_results()

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=2, recalc=True)
r2 = ui.get_pvalue_results()

print("****************")
np.random.seed(34225)
ui.get_pvalue_plot(null_model=p1, alt_model=g1 + p1,
                   num=num, numcores=4, recalc=True)
r4 = ui.get_pvalue_results()

print("---------------")
```

and get sherpa to return the fitted parameters, then we get:

# before resetting the variables each iteration

TBD

# after resetting the variables each iteration

TBD

DougBurke:
And, for reference,here are the ratios values for the numcores=1, 2, and 4 runs:

# BEFORE THE FIX

sherpa-ratios-old

#AFTER THE FIX

sherpa-ratios-new

You can see the traces look a lot more well-behaved, and you can (if you squint) see a correspindance with the parameter values above (ie when the begin to go off into unwanted parts of the parameter space).

It's not completely obvious but the first 500 elements of `numcores=2` and the first 250 elements of `numcores=4` match the `numcores=1` data, thanks to the way the RNG works. You actually also see this that numcores=2 and 4 match for the 500-725 range (for the same reason). This is something that will change when we change the RNG handling, but neither behavior is "wrong". I just wanted to point it out.

It's also interesting that 500-725 for

anetasie:
@DougBurke is this still a draft? It seems that this is something we should merge relatively soon.

DougBurke:
Ive left it as a draft so you can try it out to check it behaves sensibly. The change I made "makes sense to me" but I'd feel a lot more comfortable if you tried it out for a few cases and see how it compares to CIAO 4.15.

DougBurke:
So, I've made this change. You can use the 

- sim: store the fitted parameters for the alternate model (ppp)

commit to check out the values before the fix, so after

```python
res = get_pvalue_results()
```

you can try

```python
plot_trace(res.ratios, name="ratios")
plot_trace(res.parvals[:, 0], name=res.parnames[0])
plot_trace(res.parvals[:, 1], name=res.parnames[1])
...
```

to see how the parameter values change (which is how I created the plots above).

anetasie:
@DougBurke this looks good and having the additional array of fitted parameters helps in understanding what's happening. That's great.

I run this code with my data in CIAO 4.15 and this PR and for this case the results look the same. The line is strong and the fit does not wonder into the bad location.

It looks good. Thanks.

sherpa/sim/tests/test_sim.py
```diff
@@ -129,91 +134,251 @@ def setup():
...
+
+    It looks like each iteration can generate different results
+    depending on the platform (linux vs macOS) due to numerical
+    differences (assumed) - e.g. the fwhm may be 0.03 in one and 0.06
+    in the other, which can then lead to very different amplitudes,
+    even though the resulting fit is not very different.  Then,
+    because the values are not reset after each fit - see issue #1746
+    - we quickly get rather different ratio values.  At least, that is
+    the current hypothesis. So this test is limited to a small number
+    of iterations, where this does difference does not seem to be a
```
hamogu, line 328: "does difference does". Remove first does.

sherpa/sim/tests/test_sim.py
```diff
@@ -129,91 +134,251 @@ def setup():
...
+    problem. To make it easier to write (since the values change as
+    niter is changed, which makes it a loop to try and identify a
+    sensible number of iterations), we rnu for 25 iterations but only
```
hamogu, line 331: rnu -> run

hamogu (changes requested):
The code is fine, and this change in approach makes sense to me fro ma practical point of view - what's the gain from calculating a long series of numbers, that are very, very far away from the actual fit?

On the other hand, I have this nagging fear that this is a bandaid for some deeper problems and that we should not just change the algorithm. I mean, there are hundreds of papers on variants of the Metropolis-Hastings algorithm and my understanding is that you can't simply say "let's always jump form the same point instead of jumping around from the last point" and expect that all the other statistical properties of the chain stay the same. At the very least, the part of that parameter space that's explored will be different (after all, it's the whole point if this PR to make it different). So, what are the consequences of that? Is the resulting likelihood ratio still chi^2 distributed in the limit of the infinite samples? I have to admit that I don't know, not being a trained statistician. So, what does it mean that it gets "stuck" in the first place? Is that just a quirk of the implementation (e.g. that certain parameters can't be negative? Or is it round-off errors?). Or is it that a sign that the test is not applicable in the fist place? For example in #1746 , the errors on each data value are far too small (according to @DougBurke), and thus the starting fit is already very bad (in a chi^2 sense), so maybe we can't expect that MH algorithm to jump sensibly. If that's the case, than we should not implement this fix, but instead develop a heuristic that point out when there are problems so that the user does not miss it.

Now, I'm not a co-author of the original paper (as @anetasie is), nor am I the secretary of the Astrostatistics Interest Group of the American Statistical Association, as the author of this PR is, so I can't claim the authority to reject this.

In summary: I'm uncomfortable with this change of algorithm and I think "it seems to work" and "it looks good" if not enough to deviate in such significant way from the published papers we claim to implement unless we understand that the root cause of the problem.

However, I leave it to @anetasie and @DougBurke to overrule me. In that case, I would only insist that this be clearly documented in https://sherpa.readthedocs.io/en/4.15.0/mcmc/index.html

sherpa/sim/tests/test_sim.py
anetasie: @hamogu For the MCMC algorithms there are two options - one jumping always from the best location and the other jumping from the new location. There is no fit being done when you run MCMC - just calculation of the likelihood at the point in the parameter space.
plot_pvalue is not an MCMC algorithm to sample parameter space. It performs a simulation and fitting the simulated data at each iteration. The simulated data are from the null model which are then fit with the null and alternative model to calculate likelihood ratio. The algorithm is specified in the last section of Protassov et al paper.

anetasie:
I realized that this comment should be in the thread not at the code lines.

@hamogu For the MCMC algorithms there are two options - one jumping always from the best location and the other jumping from the new location. There is no fit being done when you run MCMC - just calculation of the likelihood at the point in the parameter space.
plot_pvalue is not an MCMC algorithm to sample parameter space. It performs a simulation and fitting the simulated data at each iteration. The simulated data are from the null model which are then fit with the null and alternative model to calculate likelihood ratio. The algorithm is specified in the last section of Protassov et al paper.

sherpa/sim/tests/test_sim.py
DougBurke: I was going to say this is just MC, not MCMC (for whatever you take montecarlo to mean, which I am being very liberal at interpreting here)

hamogu:
OK, then my request comes down to: Please add a sentence like "we use the rule to always jump from XXX"

DougBurke:
@hamogu - where would you want this line to go? I am lost as to where you think it would be useful.

hamogu:
In here: https://sherpa.readthedocs.io/en/4.15.0/mcmc/index.html

DougBurke:
But that document is not talking about this code. This is for the code around the https://sherpa.readthedocs.io/en/4.15.0/ui/api/sherpa.astro.ui.plot_pvalue.html command, which has nothing to do with https://sherpa.readthedocs.io/en/4.15.0/mcmc/index.html

So I'm confused what you want here.

anetasie:
There is no MCMC run in `plot_pvalue` only simple MC and there should not be any reference to the 'jumping rule' there.

anetasie:
@DougBurke I can expand the description of the algorithm in the 'plot_pvalue' help to clarify the simulation part and link to the description in Protassov etal paper. I'm not sure where exactly I should add the text.

DougBurke:
@anetasie - this is one of the problems with how we structure our code. CIAO users are going to come to this from the `plot_pvalue` and `get_plot_pvalue` commands, so you'd probably want something in `plot_pvalue` for these users. If you are using it directly then probably something in the `sherpa.sim.simulate` model. However, this is not well documented. So I suggest looking at https://sherpa.readthedocs.io/en/4.15.0/ui/api/sherpa.astro.ui.plot_pvalue.html and seeing if that can be improved (it's the docstring for `plot_pvalue` in `sherpa.astro.ui.utils.py`) and then we can look at improving other parts of the documentation in later PRs.

hamogu:
I take that as @anetasie will make a suggestion on what to say there. I'll wait for that. @anetasie if you don't have any suggestion by tomorrow evening, then I'll open a new issue to keep track of this and merge this in since it seems that neither my nor your ideas where it could go don't easily work out. However, for this PR that would be a "nice to have", not a necessity, so I think this PR could go in without the extra line in the docs that I requested, if we don't find a space for it.

anetasie:
Updated the docstring in #1759

anetasie (approved):

sherpa/sim/tests/test_sim_unit.py
```diff
@@ -51,43 +52,50 @@ def test_list_samplers_contents():
...
+    assert toks[1].startswith('stats   = [[2.1,3. ')
+    assert toks[4] == 'ratios  = [0.2,0.4,0.3]'
     assert toks[5] == 'null    = 12.3'
     assert toks[6] == 'alt     = 14.4'
     assert toks[7] == 'lr      = 0.2'
     assert toks[8] == 'ppp     = 0.4'
 
 
-def test_lrt_dows_not_like_gaussian():
+# Shold WStat be allowed or not? See #1745 but for now check it errors out.
```
wmclaugh, line 96:
```suggestion
# Should WStat be allowed or not? See #1745 but for now check it errors out.
```

DougBurke:
I've rebased this since #1751 has been reverted, and to pick up the review comments from @hamogu and @wmclaugh

DougBurke:
@hamogu - just to check that you're okay with this now (as the documentation side is partly done by 

- #1759 

and partly for later work, or at least that's my understanding)

hamogu (approved):

hamogu:
I am OK with this. Approved.

PR #1750, opened 2023-04-26
Replace pkg_resources as it is deprecated

### LINKED ISSUE #1723: pkg_resources use discouraged
setuptools 67.5.0 (released 05 Mar 2023) has this to say

> Although pkg_resources has been discouraged for use, some projects still consider pkg_resources viable for usage. This change makes it clear that pkg_resources should not be used, emitting a DeprecationWarning when imported.

linked to https://github.com/pypa/setuptools/pull/3843

We have one use case - 

https://github.com/sherpa/sherpa/blob/859455d218534dc16b0e1526516f70c732fa0e76/sherpa/utils/formatting.py#L29-L39

DougBurke:
# Summary

Remove the use of the deprecated pkg_resources module and replace with use of importlib.resources. Fix #1723

# Detalis

The pkg_resources module has long been "somewhat" deprecated but it looks like, with pip 23.1 and Python 3.12  (e.g. see https://discuss.python.org/t/announcement-pip-23-1-release/25844), this is getting more important. Fortunately it is an easy fix - see
    
    https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-string
    
but we don't need to worry about using the importlib-resources
backport as we are now Python > 3.7. **However**, it turns out that
the importlib.resources.files interface was only added in Python
3.9, and we still support Python 3.8. So we use importlib_resources
got pre-3.9 systems (I investigated other approaches, including
using the "deprecated" importlib.resources interface) but none are
particularly useful (which is why this PR has been rebased so-many times).

Although now we have dropped python 3.8 support we can re-simplify™ this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,6 @@ python_requires = ~=3.9
 zip_safe = False
 install_requires =
     numpy
-    importlib_resources; python_version < '3.9'
 
 tests_require =
     pytest-xvfb
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ python_requires = ~=3.9
 zip_safe = False
 install_requires =
     numpy
+    importlib_resources; python_version < '3.9'
 
 tests_require =
     pytest-xvfb
--- a/sherpa/utils/formatting.py
+++ b/sherpa/utils/formatting.py
@@ -27,10 +27,7 @@ This is aimed at IPython/Jupiter support but may be useful elsewhere.
 
 import contextlib
 import html
-try:
-    from importlib.resources import files
-except ImportError:
-    from importlib_resources import files
+from importlib.resources import files
 
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


DougBurke:
Dang it. Looks like importlib.resources.files which is what we need is in Python 3.9., not 3.8.

We could make importlib_resources conditional...

hamogu:
How much longer are the supporting Python 3.8? I know we've discussed several time before and I always need to look it up again...
[NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) says to drop Python 3.8 about two weeks ago. I know we are doing an incremental release now so we might want to keep it (I know we discussed OTS updates, but I don't have the version numbers and dates in my head) and it's a really small change (in number of code lines), so not a big deal either way.

DougBurke:
With #1753 / #1752 we can simplify this PR>

DougBurke:
This has been rebased to pick up the change to support python 3.9 to 3.11, and suppi=ort for 3.8 has been dropped.

DougBurke:
@Marie-Terrell @harlantc I think think this is good to be reviewed now after the changes you made for the 4.15.1 release. I don't see it as high priority just now!

Marie-Terrell (approved):
Looks good to me!

PR #1751, opened 2023-04-26
Explicitly require attrs module when testing Sherpa

DougBurke:
# Summary

Ensure that the attrs module is available for testing Sherpa.

# Details

The attrs moodule is required for one test so make sure it is installed. It is not clear why this is needed now, as it always should have been needed: what changed to make it a requirement? 

I am not sure all the changes are needed as I forget now exactly what we use (does the `tests_require` stansa in `setup.cfg` actually get used) but it's best to be safe rather than sorry, particularly as there's larger changes I've been waiting to get landed (e.g. #1608).

Without this I find my CI runs fail.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -201,7 +201,7 @@ jobs:
         fi
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
-        conda install -yq pytest-cov
+        conda install -yq pytest-cov attrs
         python setup.py -q test -a "--cov sherpa --cov-report xml"
 
     - name: sherpa_test Tests
@@ -209,7 +209,7 @@ jobs:
       run: |
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
-        conda install -yq pytest-cov
+        conda install -yq pytest-cov attrs
         cd $HOME
         sherpa_test --cov sherpa --cov-report xml
 
--- a/.github/workflows/ci-pip-workflow.yml
+++ b/.github/workflows/ci-pip-workflow.yml
@@ -105,6 +105,7 @@ jobs:
         TEST: ${{ matrix.test-data }}
       run: |
         git submodule deinit -f .
+        pip install -r test_requirements.txt
         pip install pytest-cov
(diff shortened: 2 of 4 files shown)


DougBurke:
Note that #1750 is an example of why this commit is needed.

hamogu:
Since we only use it in one location to set up fixtures for tests, I wonder if an alternative solution would be to just rewrite these tests to plain Python?

I don't know the details of the `attr` module so I can't say authoritatively how long that would take, but it seems that it's just an excersize in bookkeeping and passing all those parameters into the tests. For this purpose, a tuple or a long list of `pytest.parameterize` (as we do in other locations) would do.

We might make the tests more readable (even if slightly longer) by removing this dependency. Given that the `attr` module is shadowed by the more popular `attrs` module (see remark on https://pypi.org/project/attr/) there is bound to be conflict in the future again (I mean, which one of the two is this even supposed to be using?), so removing this dependency altogether might not be a bad investment of an our of time.

(I'm tempted to make a PR, but I know that there are things I should be doing more urgently. Not sure if that knowledge will stop me from doing that ...)

hamogu:
That would have the added advantage to be entirely independent of the other changes to setup.py.

DougBurke:
The alternative is that we decide we really like `attrs` and we should use it more! As I haven't used attrs in anger (last having looked at it many years ago) I'm  agnostic, so if we know how to rewrite the test to avoid attrs we can do that. I just don't have the bandwidth for it.

hamogu:
I'm not saying it's not useful in general, but I'm not convinced that
it's worth it here. Let me see if I get stuck on some telecon in the
next few days when I can mechanically reformat a test setup but not do
anything that requires real thought.Message ID:
***@***.***>

Marie-Terrell (approved):
I'm good either way. However, since this is broken for now I'm good with taking this fix for now (with potentially removal at a later date if not needed).

Marie-Terrell:
(And just to note. This will show up in CIAO's Sherpa Conda tests until the attrs references are removed or unless someone installs attrs. CM can add this to the CIAO's sherpa requirements for now. I assume this is also the case for downloads of the deployed Sherpa packages.)

PR #1770, opened 2023-05-18
Drop python2 ism in the fit code

DougBurke:
# Summary

Internal clean up of the fit code to replace a Python 2 method with a Python 3 version and to use f-strings.

# Details

This is clean-up code which I've pulled out of #1735

- added some corner cases for the fit module (based on reviewing the un-tested code reported in #1735); basically whether the string output of various classes is explicitly checked, but also the "save-each-iteration" part of the fit call
  - it's a bit worrying here as the numbers we test against could depend on platform (e.g Intel vs ARM vs ...) but let's just see how they work with the current CI tests
- one test is whether the `Fit` object can be used in a boolean context
- the reason why we care about the boolean-ness of `Fit` is because we had been using the `__nonzero__` dunder method, but this appears to have been a Python-2 ism, so let's use the modern dunder method `__bool__`. Alternatively, we could just remove it as I'm not convinced people are using it, but as it is not a lot of code, I think we should keep it to avoid the possibility of breaking code (I couldn't actually find much,if anything, on the nonzero dunder method, so there is an argument to just remove it, and you do note that the boolean-ness of Fit is changed by this work)
- I've changed a number of routines to use f strings (there's still some use of the old style "format % vals" approach, but that's because it's building strings to then build later strings, and so it's just not worth touching this at the moment), and to build up a list of strings which we can then concatenate rather than lot's of `str += ...` lines, where it's easy to miss lines (and technically faster but I know Python has spent a lot of effort on making a + b + .. + z to be fast, so I don't expect any actual difference, and this is not performance-sensitive code anyway)
  - there's a few other code-style changes here
- I then add a case to check the fit results when called with an itermethod (as code coverage showed it wasn't covered) and this lead to a few small code improvements to catch error cases

Here's a note on the `__nonzero__` to `__bool__` change: https://portingguide.readthedocs.io/en/latest/core-obj-misc.html#customizing-truthiness-bool

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -18,16 +18,15 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from functools import wraps
 import logging
 import os
 import signal
 
-from functools import wraps
-
 import numpy as np
+from numpy import arange, array, iterable, sqrt, where, \
+    ones_like, isnan, isinf
 
-from numpy import arange, array, abs, iterable, sqrt, where, \
-    ones_like, isnan, isinf, any
 from sherpa.utils import NoNewAttributesAfterInit, print_fields, erf, \
     bool_cast, is_in, is_iterable, list_to_open_interval, sao_fcmp
 from sherpa.utils.err import DataErr, EstErr, FitErr, SherpaErr
@@ -111,6 +110,9 @@ class StatInfoResults(NoNewAttributesAfterInit):
         self.qval = qval
         self.rstat = rstat
 
+        # TODO: should this call
+        # NoNewAttributesAfterInit.__init__(self)
+
     def __repr__(self):
         return '<Statistic information results instance>'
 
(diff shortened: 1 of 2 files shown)


sherpa/fit.py
```diff
@@ -1403,7 +1429,7 @@ def report_progress(i, lower, upper):
             # If maximum number of refits has occurred, don't
             # try to reminimize again.
             if (hasattr(self.estmethod, "maxfits") and
-                    not(self.refits < self.estmethod.maxfits - 1)):
+                    not self.refits < (self.estmethod.maxfits - 1)):
```
hamogu, line 1432: This is technically true, because arithmetic comparison has precedence over logical NOT. However, I had to look that up, because I've definitely done things like "False < 3" accidentally and that also works in Pytohn, leading to hard-to-find bugs.

Just to help future readers, let keep the outer parenthesis in addition to the inner ones that are added here as a bugfix:

```suggestion
                    not (self.refits < (self.estmethod.maxfits - 1))):
```

sherpa/tests/test_fit_unit.py
```diff
@@ -2826,3 +2826,252 @@ def test_fit_ensures_data_and_model_dimensionality_matches(data_class, ddim, mod
...
+
+    toks = str(out).split("\n")
+    for tok, expected in zip(toks, expecteds):
+        assert tok == expected
+
+    assert len(toks) == len(expecteds)
+
+
+def test_fit_results_str():
+    """Just check we so call str() on a fit results instance"""
```
hamogu, line 2842:
```suggestion
    """Just check we can call str() on a fit results instance"""
```

sherpa/tests/test_fit_unit.py
```diff
@@ -2826,3 +2826,252 @@ def test_fit_ensures_data_and_model_dimensionality_matches(data_class, ddim, mod
...
+               "numpoints      = 4",
+               "dof            = 3",
+               "qval           = 0.23114006377865534",
+               "rstat          = 1.4322916666666667",
+               "message        = successful termination",
+               "nfev           = 8"])
+
+
+def test_fit_results_bool_true():
+    """Just check we so call bool() on a fit results instance"""
```
hamogu, line 2929:
```suggestion
    """Just check we can call bool() on a fit results instance"""
```

sherpa/tests/test_fit_unit.py
```diff
@@ -2826,3 +2826,252 @@ def test_fit_ensures_data_and_model_dimensionality_matches(data_class, ddim, mod
...
+    model = Const1D()
+    with pytest.raises(ValueError,
+                       match="^not-a-name is not an iterative fitting method$"):
+        Fit(data, model, itermethod_opts={"name": "not-a-name", "grow": 2})
+
+
+def test_fit_results_with_iteration_str():
+    """Just check we so call str() on a fit results instance with an iteration method
+
+    We have not picked a dataset where the
```
hamogu, line 2898:
```suggestion
    """Just check we can call str() on a fit results instance with an iteration method
    """
```

sherpa/fit.py
DougBurke: Yes - there's a delicate dance to be had between accepting a linting suggestion to remove brackets/precedence/logic and keeping things because they more-clearly express intent!

sherpa/tests/test_fit_unit.py
DougBurke: I'd originally meant to add a point that the database wasn't set up for actually causing the sigmarej outine to fire, hence the incomplete sentence which you've removed, but I went and created something that did actually have an outlier just to make sure, so a double reason to remove the text...

DougBurke:
I had to redo the "apply suggestions from code review" as the original suggestion accidentally added a `"""` when it shouldn't have, which I hadn't noticed, leading to no end of test fun and frolicking. Said frolicking has been put to pasture.

hamogu (approved):

PR #1772, opened 2023-05-19
Tweak resample data code

DougBurke:
# Summary

Internal changes to the ReSampleData class to better handle invalid input.

# Details

This is clean-up code which I've pulled out of https://github.com/sherpa/sherpa/pull/1735. PR #1777 relies on this PR.

In trying to add checks to improve coverage, I realised that the code - which was meant to report an error if ReSampleData was called with an invalid data object - did not actualy work, since it basically required there to be a y attribute, which is not guaranteed for non-1D data types. So, I've moved the check to earlier, taking advantage of the ndim field I recently-ish added (in #1477), rather than check on the data class (I added an earlier commit with a test of this functionality).
    
Should this class warn or error out if given a DataPHA or Data1DInt object? Alternatively, the code could be enhanced to properly handle this (but that is hard for the PHA case as we need to worry about the response).

### DIFF
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -831,7 +831,7 @@ class ReSampleData(NoNewAttributesAfterInit):
 
         # Should this error out if data is an instance of Data1DInt?
         if data.ndim != 1:
-            msg = f"{ReSampleData.__name__} {type(data)}"
+            msg = f"{ReSampleData.__name__} is only implemented for 1D data, got {type(data)} instead."
             raise NotImplementedError(msg)
 
         self.data = data
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2015, 2016, 2018, 2019, 2020, 2021
+#  Copyright (C) 2011, 2015, 2016, 2018, 2019, 2020, 2021, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -773,7 +773,7 @@ class ReSampleData(NoNewAttributesAfterInit):
 
     Parameters
     ----------
-    data : sherpa.data.Data1DAsymmetricErrs instance
+    data : sherpa.data.Data1DAsymmetricErrs or sherpa.data.Data1D instance
        The data.
     model : sherpa.models.model.ArithmeticModel instance
        The model to fit the data. The model parameters are taken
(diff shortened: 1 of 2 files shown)


DougBurke:
@anetasie - just to say that this is one of the PRs I need attention.

anetasie:
@DougBurke I will look at this one while traveling... nothing better to do on the plane.

DougBurke:
I can think of several things I'd refer to do on an :aeroplane: ....

DougBurke:
Rebased just to pick up recent chnages.

sherpa/sim/__init__.py
```diff
@@ -828,6 +828,12 @@ class ReSampleData(NoNewAttributesAfterInit):
 
     """
     def __init__(self, data, model):
+
+        # Should this error out if data is an instance of Data1DInt?
+        if data.ndim != 1:
+            msg = f"{ReSampleData.__name__} {type(data)}"
```
hamogu, line 834: Should we be more wordy here, e.g. "Resample data does not implement support for Data2DInt". I know it's a NotImplementedError, I'm just trying to help a user understand what to do if they hit this error (namely, use a different data class).

anetasie:
@DougBurke Is this resampling only used for the asymmetric errors? If that's the case then I don't think we should use DataPHA. On the other hand I can imagine that Data1DInt might be used.

DougBurke:
@aneta - I didn't write this code and I wasn't involved in its development. All I can do is point to the existing code.

anetasie:
@DougBurke Sure. If it is only used for the data with asymmetric errors then  the resampling is used to estimate the uncertainties with the implicit assumption of the Gaussian errors on data - different sigmas  for upper/lower error to account for assymetry.

DougBurke:
I'm pretty sure the resample code does allow it to be run with symmetric
errors as well, but I haven't looked at this code for a while so I may be
missing something,

anetasie:
Yes, it should be fine for the symmetric errors.

anetasie (approved):

sherpa/sim/__init__.py
DougBurke: I don't have a lot of mental energy for improving the user experience here. If this call fails we don't have any replacement code the user could use - they'd have to do it themselves. I'm not sure how best to explain this without writing really-long error messages which then become too wordy. But as I said , I have little energy for this so if you can come up with a sensible message then I'm all ears.

hamogu:
```suggestion
            msg = f"{ReSampleData.__name__} is only implemented for 1D data, got {type(data)} instead."
```

hamogu:
I did my suggestion inline, but did not change the test for that error. I don't have any other comments, so once you fix that test I'll approve.

hamogu (approved):

PR #1773, opened 2023-05-19
Internal clean up of estmethods

DougBurke:
# Summary

Code clean up of sherpa.estmethods.

# Details

This is clean-up code which I've pulled out of https://github.com/sherpa/sherpa/pull/1735

These changes are driven by pylint, and involve the following, but note [*]

- use of f-strings (although there are plenty of strings that could be converted, the test coverage of these other lines is poor [*] so I don't want to make changes that could break things)
- using modern Python idioms, like
  - do not create a `__init__` call that just calls the superclass
  - do not have `if/elif/.../else` chains when each branch ends in a return or an exception
  - "re-raise" exceptions via "from" to make tracebacks cleaner
  - avoid un-needed brackets
  - use things like "min(xs, val)" and "a < 23" to avoid a "if/then" call
- I have also removed some `__str__` and `__repr__` mehods that were only ever really neded for development/testing (including at least one mis-spelled `__rep__` case)

There's a lot more changes that could be made - this code appears to be written in a Java/C++ style and so uses names like `dir` and `iter` that shadow existing functionality - but I do not want to make the changes too extensive at this time (as I don't plan to be changing this code anytime soon, if ever).

There are some changes that could be made but they will be made either as part of #1735 or some other PR (in particular changes to the parallel code).

[*] coverage is low in part because there's a lot of code written to handle exceptional circumstances (such as finding a better fit during the error analysis, or handling of semi-open intervals, which are hard to find test cases for)

### DIFF
--- a/sherpa/estmethods/__init__.py
+++ b/sherpa/estmethods/__init__.py
@@ -741,10 +741,7 @@ class ConfRootBracket(ConfRootNone):
                     warn_user_about_open_interval([xa, xb])
                     return (xa + xb) / 2.0
 
-                if xa < xb:
-                    return (xa, xb)
-
-                return (xb, xa)
+                return min(xa, xb), max(xa, xb)
 
             return None
 
--- a/sherpa/estmethods/__init__.py
+++ b/sherpa/estmethods/__init__.py
@@ -18,6 +18,7 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from itertools import chain
 import logging
 
 try:
@@ -122,8 +123,7 @@ class EstMethod(NoNewAttributesAfterInit):
     def __getattr__(self, name):
         if name in self.__dict__.get('config', ()):
             return self.config[name]
-        raise AttributeError("'%s' object has no attribute '%s'" %
-                             (type(self).__name__, name))
+        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
 
     def __setattr__(self, name, val):
         if name in self.__dict__.get('config', ()):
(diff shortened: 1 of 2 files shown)


sherpa/estmethods/tests/test_estmethods.py
```diff
@@ -209,3 +211,84 @@ def test_projection(parallel):
...
+                  "tol          = 0.2",
+                  "verbose      = False",
+                  "openinterval = False"
+                  ])
+
+
+def test_estmethod_str_projection():
+    """Simple check."""
+    m = Projection()
+    print(str(m))
```
DougBurke, line 280:
```suggestion
```

DougBurke:
Sorry @hamogu - this one isn't particularly interesting and it doesn't obviously improve anything. It's just clean-up I noticed when working in #1735. This is not really code I want to work on, so I am trying not to make too many changes, in particular because the test coverage isn't great, and trying to find tests that can improve this coverage is hard, as it deals with a lot of edge cases when calculating errors, which are not easy to create from first principles...

sherpa/estmethods/__init__.py
```diff
@@ -763,36 +740,33 @@ def warn_user_about_open_interval(list):
...
-                        return (xa, xb)
-                    else:
-                        return (xb, xa)
-            else:
-                return None
+
+                if xa < xb:
+                    return (xa, xb)
+
+                return (xb, xa)
```
hamogu, line 747:
```suggestion
                return min(xa, xb), max(xa, xb)
```

hamogu (changes requested):
@DougBurke I appreciate these seemingly small clean-ups. We don't want to be changing code just for the sake of changing code; but when we (you much more often than I do) go spelunking into Sherpa's bowls, it helps to find code that's somewhat clean an readable so that I can understand the behavior I'm looking for quickly and go back to my original problem.

I agree with your assessment to do some simple changes while looking at it and leave comments (TODO) for other things that turn up but are not addressed now.

Could be merged as is, but since I reviewed it anyway, one suggestion that cuts another few lines and level of indentation to make the intend of the code more clear.

hamogu (approved):

Issue #1744, opened 2023-04-13
parallel_map does not handle failing functions particularly well


DougBurke:
So two parts

1. when run in parallel we lose traceback info
2. if the routine does fail, is error does not seem to be caught (even though the code explicitly tries to do so)

I'm dumping these here as I am really trying to work on something different and have already spent a lot of time on this digression.

# case 1

in #1738 I add extra information for XSPEC errors. This was found in some code that was being run in parallel, but we tend not to see the traceback, which can make identifying what the problem was hard (although point 2 also comes into play here).

We can add some traceback information when we catch an error (it's not perfect as we can't pickle a traceback, but we can dump it to strings and send them, which the calling code can dump to the error log channel). However, not worth doing until case 2 is understood better.

I think the problem I was trying to debug that lead to #1738 was always going to be hard to identify, as it involved XSPEC code potentially crashing (definitely overwriting memory) and so once that happens all bets were off.

# case 2

This is based on test code I was going to add just to show the code is actually working. But it isn't...

I think (based on reading the code) that the `parallel_map` calls should handle the following error case (by essentially re-throwing the exception). We do when `numcores=1` but once we hit the parallel code it doesn't seem to actually catch the error. Instead the code assumes it has returned an answer and fails to extend an array when it hasn't. You see the `ValueError` message but it isn't an actual error, it has just been printed to the screen:

```
sherpa-4.15.0> import pytest

sherpa-4.15.0> def myfunc(x):
          ...:     if x == 2:
          ...:         raise ValueError("x is 2. It is a bad number")
          ...:     return x / 2
          ...: 

sherpa-4.15.0> from sherpa.utils import parallel_map

sherpa-4.15.0> ans = parallel_map(myfunc, [-2, 0, 1, 3], numcores=1)

sherpa-4.15.0> ans
[-1.0, 0.0, 0.5, 1.5]

sherpa-4.15.0> with pytest.raises(ValueError, match="x is 2. It is a bad number"):
          ...:     parallel_map(myfunc, [-2, 0, 1, 2, 3], numcores=1)
          ...:     print("Should not see this")
          ...: 

```

So far so good. If we use the multiprocessing code then it unravels:

```
sherpa-4.15.0> import sys

sherpa-4.15.0> sys.tracebacklimit = 5

sherpa-4.15.0> with pytest.raises(ValueError, match="x is 2. It is a bad number"):
          ...:     parallel_map(myfunc, [-2, 0, 1, 2, 3], numcores=2)
[...]

Issue #975, opened 2020-10-13
centralize multiprocessing code


DougBurke:
From https://github.com/sherpa/sherpa/pull/969#issuecomment-706138609

> We could add comments like '# Must be imported after sherpa.utils' to the multiprocessing imports, but I wonder if long term we should have a sherpa.utils.multiprocessing module (or sherpa.multiprocessing) so that we know the setup has been called. That is definitely for a later PR.

There are also comments later from @hamogu to think about: https://github.com/sherpa/sherpa/pull/969#issuecomment-706160700

DougBurke:
For example, sherpa/estmethods/__init__.py imports multiprocessing and does not setup the changes in #969

DougBurke:
We now have `sherpa.utils.parallel` so I think we have enough structure.

Issue #1743, opened 2023-04-10
what is parallel_map_funcs doing?


DougBurke:
The documentation for parallel_map_funcs is unclear on exactly what it's doing, but the code is also unclear.

# documentation

This is the example (but I may have added this after to try and illuminate the code so I'm not sure how much it should be trusted):

https://github.com/sherpa/sherpa/blob/f9746d440689fc33d0e7b88f709b14756e96cb92/sherpa/utils/__init__.py#L3143-L3158

# code

Here's what happens when there's no multiprocessing going on:

https://github.com/sherpa/sherpa/blob/f9746d440689fc33d0e7b88f709b14756e96cb92/sherpa/utils/__init__.py#L3177-L3179

This applies the **first** function to each dataset. I think it should be something like the following, but maybe I'm missing something (and I mention this in the example section below):

```python
return [f(d) for f, d in zip(funcs, datasets)] 
```

# example

Here's a run through with CIAO 4.15:

```

sherpa-4.15.0> from sherpa.utils import parallel_map_funcs

sherpa-4.15.0> funcs = [np.sum, np.abs]

sherpa-4.15.0> ds = [np.arange(4), np.arange(-2, 3)]

sherpa-4.15.0> parallel_map_funcs(funcs, ds)
TypeError: 'NoneType' object is not iterable

sherpa-4.15.0> parallel_map_funcs(funcs, ds, numcores=1)
[6, 0]

sherpa-4.15.0> parallel_map_funcs(funcs, ds, numcores=2)
[0, 1, 2, 3, 2, 1, 0, 1, 2]
```

We can see that when `numcores=1` we get the `sum` call applied to both input datasets.

When `numcores=2` - and this is a machine with multiple CPUS - it's not at all clear what it is doing. It appears to be running `sum` on each element of `arange(4)` and then `abs` on each element of `arange(-2, 3)` and then concatenating the results. This is definitely something that could make sense, but if so then this means the `numcores=1`/no multiprocessing case is definitely wrong but not in the way I think it suggest above.

DougBurke:
It looks like it's only used in one place, but it is an important place: model evaluation to match a data set (when multiple data are in use):

https://github.com/sherpa/sherpa/blob/f9746d440689fc33d0e7b88f709b14756e96cb92/sherpa/data.py#L1400-L1464

It does at least point out why the `numcores=1` case in `parallel_map_funcs` isn't triggering an error, as this case never gets to `parallel_map_funcs`.

[...]

PR #1776, opened 2023-05-20
Move the parallel code into a separate modue

DougBurke:
# Summary

Move the `parallal_map` code from `sherpa.utils` to the new `sherpa.utils.parallel` module (the symbols are still accessible from the original location but code should be updated to use the new location).

# Details

This is part of a low-level "break up the large utility-like modules" into smaller modules. In #975 I suggested centralizing the `multiprocessing` code. This can be thought of as addressing this concern, but under a more-generic name, and I don't "finish off" the job to make sure we always import `multiprocessing` via this module (that could be done as a later change).

There's a number of clean ups that go with this change.

This may solve #1744 but this needs more thinking about to make sure.

I do not understand the `parallel_map_funcs` method, and I think the example I added (well, I think I added) a while back is no ideal / shows my lack of understanding. There's only one place we use this method - `sherpa.data.DataSimulFit.eval_model_to_fit` when the `self.numcores` value > 1, which is not the default behavior (e.g. `sherpa.fit` uses the default value of `numcores=1`) so I am not sure if we ever use it.


DougBurke:
Rebased as #1771 has been merged.

DougBurke:
Rebased because of conflict with #1781

DougBurke:
So, this one is the first "major" pieces of #1735 I'm asking review for. It's major because it re-organises the code, rather than anything particularly earth shattering in the code (I think, but I'm biased).

One really minor thing I want to look at is if I can avoid the `NoSectionError` check in the following by making use of the `fallback` argument (which is a lot newer than this code):

```
_ncpu_val = "NONE"
try:
    _ncpu_val = config.get('parallel', 'numcores').strip().upper()
except NoSectionError:
    pass
```

hamogu:
I agree with this incremental clean-up. I've also looked at the multi-processing code at various times over the years, and, just like @DougBurke indicated above for himself, I'm also not happy with the state of it. It's not clear for users how or when to use it, and even when used it's not clear that it's actually beneficial. 
For example, if we are not careful, we might launch *too many* threads, because numpy itself will also spawn threads when it uses some of the BLAS/LAPACK routines underneath, unless that's carefully handled (see e.g. https://github.com/joblib/threadpoolctl). It's not at all clear to me that Sherpa deals with those issues correctly (or at all) - or if it even matters for the models with typically have.

That's why we need to understand what Sherpa is doing right now and with Dan gone, that unfortunately means clean up what's already there - not only to prepare for any future changes, but also so that we can actually assess what's already there. So, I'm happy that @DougBurke spend some time on this and I'm also happy to learn more about what's done in the code by reviewing it.

DougBurke:
So, I am significantly less concerned about this code (the "how do we run stuff inparallel") that @hamogu is, because it's not that bespoke and is fairly standard Python code.

There are cases where it could conflict with other code that uses multiple cores, but that's no different than any other bit of Python code doing this (which is why there are packages like threadpoolctl), but until we see problems with this then I am not going to worry about it.

There is some interesting things to think about random-number generation and multiprocessing, but that's a special-case (and I'm looking at that as part of #1735).

sherpa/utils/parallel.py
```diff
@@ -0,0 +1,462 @@
...
+
+    Returns
+    -------
+    result : list
+        The result from the processes. This may contain more elements
+        than procs, as each process may return multiple results.
+
+    Notes
+    -----
+    Each process sends it's output - a pair with index and a list of
```
hamogu, line 218:
```suggestion
    Each process sends its output - a pair with index and a list of
```

sherpa/utils/parallel.py
```diff
@@ -0,0 +1,462 @@
...
+
+    Sherpa's parallel_map runs a single function to an iterable set of
+    sequence.  parallel_map_funcs is generalized parallelized version
+    of sherpa's parallel_map function since each element of the ordered
+    iterable funcs shall operate on the each element of the datasets.
+
+    Parameters
+    ----------
+    funcs : a list or tuple of functions
+       An ordered iteratble sequence of functions which accepts an element
```
hamogu, line 369:
```suggestion
       An ordered iterable sequence of functions which accepts an element
```

sherpa/utils/parallel.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+# TODO: this routine needs a review
+def parallel_map_funcs(funcs, datasets, numcores=None):
+    """Run a sequence of function on a sequence of inputs in parallel.
+
+    Sherpa's parallel_map runs a single function to an iterable set of
+    sequence.  parallel_map_funcs is generalized parallelized version
+    of sherpa's parallel_map function since each element of the ordered
+    iterable funcs shall operate on the each element of the datasets.
```
hamogu, line 364: Below, it says that the length of the fundcs and datasets needs to be equal. That only makes sense if each func is applied to one-and-only-one if the datasets. In contrast, this section claims that each func operates on each dataset, i.e. that for n funcs and m datasets, I get n*m answers.
Only one can be right.

sherpa/utils/parallel.py
```diff
@@ -0,0 +1,462 @@
...
+    Run the computation (summing up each element of the first input array
+    and calculate the standard deviation of the second input array)
+    on a separate core and return the results (unless the machine only
+    has a single core or the parallel.numcores setting is set to 1).
+
+    >>> import numpy as np
+    >>> funcs = [np.sum, np.std]
+    >>> datasets = [np.arange(3), np.arange(4)]
+    >>> parallel_map_funcs(funcs, datasets, numcores=2)
+    [0, 1, 2, 0.0, 0.0, 0.0, 0.0]
```
hamogu, line 417: Now I get 7 outputs, where I would have expected 2 (one per function). So apparently I didn ot understand who this works.

I can make a suggestion for re-phrasing the expalantion above, but I'm not sure I understand  the purpose.

sherpa/utils/parallel.py
```diff
@@ -0,0 +1,462 @@
...
+        numcores = _ncpus
+
+    # Returns a started SyncManager object which can be used for sharing
+    # objects between processes. The returned manager object corresponds
+    # to a spawned child process and has methods which will create shared
+    # objects and return corresponding proxies.
+    manager = multiprocessing.Manager()
+
+    # Create FIFO queue and lock shared objects and return proxies to them.
+    # The managers handles a server process that manages shared objects that
```
hamogu, line 452:
```suggestion
    # The manager handles a server process that manages shared objects that
```

sherpa/utils/parallel.py
hamogu: Does this mean:
```suggestion
    of sherpa's parallel_map function. In `parallel_map_funcs` a different
    function can be associated with each dataset. For each dataset, the
    corresponding function is run for each element in that dataset.
```

hamogu (changes requested):

hamogu:
I do understand this code better now - I was probably too pessimisting in the beginning - thanks to @DougBurke for pushing me to look at this.

sherpa/utils/parallel.py
DougBurke: I really don't want to touch `parallel_map_funcs` or it's documentation (beyond making it work with doctestplus) because I do not understand what it is meant to be doing (this time I already have an issue about it - see #1743)

sherpa/utils/parallel.py
DougBurke: See https://github.com/sherpa/sherpa/pull/1776#discussion_r1209483512

sherpa/utils/parallel.py
DougBurke: See https://github.com/sherpa/sherpa/pull/1776#discussion_r1209483512

docs/overview/parallel.rst
```diff
@@ -0,0 +1,28 @@
...
+   :py:mod:`sherpa.utils` module.
+
+   .. rubric:: Constants
+
+   .. autosummary::
+      :toctree: api
+
+      _multi
+      _ncpus
+
```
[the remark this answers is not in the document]
hamogu: Maybe it's easier to just leave them in the rst like you did here? I also agree that, in this case, we should document it and probably want users to use them. Seems easier to leave as is, then to make a property. It's just unfortunately that they are named that way.

I would suggest to rename them to `_ncpus` -> `cpus` and `_multi` -> `multi`. Since they are put into a new module in this PR, this is our chance to do that. I don't think it's cimplicated enough that it needs to be a proerty, in can just be can attribute.
The existing modules can then `from parallel import ncpu as _ncpu` and we have the same backwards compatibility that we have in the current commit.

sherpa/utils/parallel.py
hamogu: I was skeptical about the parallel stuff at first and @DougBurke told me not to be afraid. I spend some time reading and I think I have better grasp now, at least to what this function does and I'm comfortable that is does what I wrote in my suggestion.
If you don't want this committed under you name, I can make a PR with just this documentation change when this one is merged and then you'll have to review my text...

docs/overview/parallel.rst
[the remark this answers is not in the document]
DougBurke: So I've now renamed these (still accessible as the old forms if importing from the old location of `sherpa.utils`,  but no code in Sherpa itself should be affected by this, but there may be downstream users in CXC).

sherpa/utils/parallel.py
DougBurke: I'd rather this be done in a separate PR. I wonder if we even need the method at all as my guess it's part of an old attempt to speed things up that never went anywhere, but that would be better discussed in a separate PR ...

DougBurke:
@hamogu - I've made the changes I intend to make based on your feedback (and added a few more tests of corner-case error handling, as well as a little bit of code clean up in `sherpa/utils/__init__.py`).

sherpa/utils/parallel.py
hamogu: I think I've actually used this method in a no-sherpa project, but I'll check. Either way, I can be addressed in a followup. See #1787

hamogu (approved):

hamogu:
Looks good!

Issue #1736, opened 2023-03-30
possible parameter sampling issue


DougBurke:
In looking at #1735 I am looking at how get_sample works, and come across this possible problem

https://github.com/sherpa/sherpa/blob/4e33a99a13f0011892748ce326cb1b5a3d7a3e82/sherpa/sim/sample.py#L751-L801

The problem is that we have

```python
    sampler = NormalSampleFromScaleVector()
    sampler.scale.sigma = sigma
    if correlate:
        sampler = NormalSampleFromScaleMatrix()

    return sampler.get_sample(fit, num=num, numcores=numcores)
```

when I think the setting of the sigma variable should be after the choice of sampler, namely:

```python
    sampler = NormalSampleFromScaleVector()
    if correlate:
        sampler = NormalSampleFromScaleMatrix()

    sampler.scale.sigma = sigma
    return sampler.get_sample(fit, num=num, numcores=numcores)
```

Unfortunately it's a bit awkward to test out as we need a fit object (well, the fit results), but I claim that when `correlate=True` (the default) then the sigma argument is not used (which would imply the results are the same, modulo the RNG), whatever sigma is used.

DougBurke:
Let's try it out, as the UI `normal_sample` routine should call this routine. Unfortunately it looks like I don't understand what the `sigma` parameter is meant to do...

This is with CIAO 4.15 but it should be the same with the main branch.

# setup

Let's fit a very-simple dataset:

```
sherpa-4.15.0> load_arrays(1, [1, 4, 10], [3, 8, 7])

sherpa-4.15.0> set_source(const1d.mdl)

sherpa-4.15.0> fit()
Dataset               = 1
Method                = levmar
Statistic             = chi2gehrels
Initial fit statistic = 6.10601
Final fit statistic   = 1.27829 at function evaluation 4
Data points           = 3
Degrees of freedom    = 2
Probability [Q-value] = 0.527745
Reduced statistic     = 0.639143
Change in statistic   = 4.82773
   mdl.c0         5.39754      +/- 2.00142     

sherpa-4.15.0> conf()
mdl.c0 lower bound:	-2.00142
mdl.c0 upper bound:	2.00142
Dataset               = 1
Confidence Method     = confidence
Iterative Fit Method  = None
Fitting Method        = levmar
Statistic             = chi2gehrels
confidence 1-sigma (68.2689%) bounds:
   Param            Best-Fit  Lower Bound  Upper Bound
   -----            --------  -----------  -----------
[...]

PR #1777, opened 2023-05-21
Change get_sample to require named arguments

DougBurke:
# Summary

The `get_sample` method of the `sherpa.sim.sample.ParameterSample` class now requires named arguments for everything but the first argument (the `fit` object). There are also a number of small style changes to the code.

# Details

This is built on #1772, which has now been merged. It is code which I've pulled out of #1735

There are pylint-inspired fix-ups, such as

- f-string changes
- but for logging calls use %-encoded strings
- change if/elif/.../then calls where each branch is a return or raise an error to multiple 'if' calls
- remove un-needed return calls
- use actual dictionaries rather than dict calls

The main change is the last commit. The `ParameterSample` class defines `get_sample` with a signature accepting the parameters `fit, num=1`. This is an abstract class and could be marked as such, but for now it is left as is. Derived classes have a variety of arguments, and unfortunately they do not always extend the `fit, num=1` pattern. For example, `UniformParameterSampleFromScaleVector` uses the arguments `fit, factor=4, num=1`. This means that any call to `get_sample` has to be careful with the arguments it sends (other than the `fit` argument, which we assume is always the first argument). To make the code easier to read, and to avoid worrying about possible issues - such as "is it `dof, num` or `num, dof`" - this commit
    
- explicitly labels all arguments to `get_sample` other than the first argument
    
  This leads to a very-minor code increase as there's at least one place where I have had to replace a generic
    
           samples = sampler.get_sample(fit, arg, num)
    
  since the second argument (i.e. arg) is named `mycov` in one case and `myscales` in the other case.
    
- marked all-but-the first argument as keyword-only - that is, callers must pass in the keyword.

This is a change in API, as old code could start failing because they have not used keyword arguments.

The reason I did this was because i was worried there was strange behavior because arguments were being mis-used, but I believe this was actually more an issue with general functionality - see #1736 - but the approach here should make it easier to follow the code. The question is, **is it worth it**?

### DIFF
--- a/sherpa/astro/flux.py
+++ b/sherpa/astro/flux.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2009, 2015, 2016, 2019, 2020, 2021
+#  Copyright (C) 2009, 2015, 2016, 2019, 2020, 2021, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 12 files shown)


DougBurke:
Rebased to pick up recent changes and to include two add-tests commits from #1735 I had somehow missed.

DougBurke:
Rebased as #1772 has been merged.

hamogu (approved):
These are mostly internal changes; the change in API will make us safer given the design of those functions, so it should be done.

Issue #1737, opened 2023-04-03
sherpa.optmethods internal code re-arrangement


DougBurke:
Some modules in `sherpa.optmethids` contain code useful for developing the optimisers, but not really helpful for the users. Can this code be moved out into either tests (that need to be triggered) or into scripts into `scripts/`?

DougBurke:
I claim this was done in #1778

PR #1778, opened 2023-05-22
Optmethods cleanup

DougBurke:
# Summary

Move testing code out of the optmethods modules and into a separate test, along with a number of code clean ups. There is no functional change.

# Details

This is built on https://github.com/sherpa/sherpa/pull/1776 which is built on #1771 (which have both now been merged). It is code which I've pulled out of https://github.com/sherpa/sherpa/pull/1735.

Most of the changes (other than the moving of the test code) are pylint inspired.

The first two commits just add tests to try and improve test coverage, but they are generally corner-case changes (i.e. do not significantly improve testing of basic functionality).

First we replace the use of a mutable argument (see #95) in the `Optmethods.fit` method: the `statkwargs` argument used to be set to an empty dictionary and now it defaults to None (which is taken to mean the same thing). The `Neldermead3` and `Neldermead4` initialisers in `sherpa.optmethods.ncoresnm` have a `finalsimplex` argument which defaulted to `[1, 2, 3]` and now defaults to `None` (which is taken to mean the same thing); these are "internal" classes and are not likely to be used by anyone.

The next commit then removes a number of methods which just call the superclass (mainly `__init__` calls) which are unnescessary in Python and actually make it harder to see what's going on (as you need to check that the code is doing the same as the superclass).  Some un-needed `return` calls have also been removed, along with some commented-out code. A few TODO comments have been added for future review, but they are not the focus of this work (which is all about getting #1735 approved). **NOTE** the commit comment references one change (removal of `_ncpus` as unused) which has now been moved into #1776 - if needed I can rebase to fix this but at present I don't see the need.

There are then a number of clean ups, including

- remove return calls that are not needed
- f-string changes
- change 'if/elif/.../then' blocks that return or raise an error at each branch to multiple  'if ..' calls
- change a number of routines to reduce the change of name shadowing of standard python symbols (e.g. change `range` to `xrange`); this is in internal routines which users can not call, but a few changes may technically be noted by the intrepid user

There is then a commit just to remove what looks to be debugging code, which was just printing out a NotImplementedError exception and then re-throwing it. As the Python backtrace code is a lot better nowadays we don't need to do this. It really loks like a debugging feature. I have this in a separate commit just to make things easier to follow.

The code has a lot of methods that take a large-number of positional arguments. I find these hard to read and review, as it's just easy to mess up the order, in particular for an untyped interface like Python. So I added a commit that just converts `foo(arg1, arg2, named1, named2, ...)` to `foo(arg1, arg2, named1=named1, named2=named2, ..._)` to improve the readability. We could go and make these methods take keyword-only arguments, but this is all internal code that I'd rather not mess with more than I am doing, as the coverage is not great. There's also some changes to avoid sending parameters into local subroutines when we already know what the value is (e.g. the tolerance field) which makes it a bit easier to read (particularly when some of the routines were already using the parent value rather than the sent-in argument). There's also one "improvement" with the arguments to a method that could have been marked as  an `@abstractmethod` but I really don't want to mess with this code:

```
-    def init(self, npop, xpar, step, seed):
+    def init(self, npop, xpar, step, seed, factor):
         raise NotImplementedError("init has not been implemented")
```

I then have a commit when I change from using the `optparse` module to `argparse`, as it is the preferred version. I am about to remove this in a follow-up PR, but I just want to clean things up first (as there's a few changes to try and avoid creating quite-so-many global variables).

The final main change is to address #1737 by moving the test code in the optimisation modules into a new `test_opt_original.py` test file.  The optimisation code had "tests" which could be run by evaluating the module directly (run via a "`if __name__ == '__main__':`" block). These tests are not pytest-like tests, and instead run a number of optimisation tests with a variety of methods, displaying the results. These were presumably used to validate and verify changes made to the code (as there is no way to impove the results for all cases).
    
One problem with this arrangement is that there was a lot of code in the modules that were nothing to do with the optimiser, and it makes it hard to see what is tested and what is relevant. So, this commit moves all the "testing" code into test_opt_original.py and makes three of the cases actual tests (those from opt.py). The code from ncoresde.py and ncoresnm.py are left "as is", and can only be run by executing the file directly. They do not actually test anything, just output the results. For example:
    
```
    % python sherpa/optmethods/tests/test_opt_original.py --no_de
    
     rosenbrock  fmin = 0.0
    ncoresNelderMead [0.99999986 0.99999963 1.00000032 1.00000065] = 9.690772711926466e-13 in 2419 nfevs
    
     freudenstein_roth  fmin = 0.0
    ncoresNelderMead [11.41277901 -0.89680525 11.41277899 -0.89680526] = 97.96850735848004 in 1744 nfevs
    
     powell_badly_scaled  fmin = 0.0
    ncoresNelderMead [1.09822720e-05 9.10558361e+00 1.09822720e-05 9.10558361e+00] = 1.0696645337509389e-14 in 7790 nfevs
    
    ...
    
     SumSquares(0, 0, ..., 0) = 0
    ncoresNelderMead [0. 0. 0. 0.]  =  0.0 in 1466 nfevs
    
     Zakharov(0, 0, ..., 0) = 0
    ncoresNelderMead [0. 0. 0. 0. 0. 0. 0. 0.]  =  0.0 in 4701 nfevs
````
    
And here's the full list of options (cobbled together from the existing ncoresnm/de code; I have not changed things so I think the -s flag is actually the inverse of the suggested text):

```    
    % python sherpa/optmethods/tests/test_opt_original.py --help
    usage: test_opt_original.py [-h] [-N NUM] [-u] [-o] [-d] [-c] [-s] [--no_de] [--no_nm]
    
    options:
      -h, --help     show this help message and exit
      -N NUM
      -u, --unc_opt  do not run tst_unc_opt
      -o, --opt      do not run tst_opt
      -d, --difevo   run simple difevo (ncoresde)
      -c, --combine  run nm & difevo (ncoresde)
      -s, --single   run test using 1 core (ncoresnm)
      --no_de        skip ncoresde tests
      --no_nm        skip ncoresnm tests
```
    
It would be nice to run this code as an actual test, since we can at least check that things have not been changed **but**
    
- this is a lot of work
  - need to split up each test into an individual test rather than the current "run all"
  - need to identify what we can test
- it's not clear how much of the code that is run in these "tests" is actually used by Sherpa, and how much is experimental code

The last commit just removes shebang lines which are now unused.

### CODE REVIEW

**Comment by DougBurke** [comment]
Rebased as #1771 has been merged.

**Comment by DougBurke** [comment]
Rebased because of conflict with https://github.com/sherpa/sherpa/pull/1781

**Comment by DougBurke** [comment]
Rebased as #1776 has been merged

**Comment by DougBurke** [comment]
Rebased to add in two relevant "add-tests" commits from #1735 I'd somehow missed.

**Comment by DougBurke** [comment]
@hamogu let me know if you'd like this broken up (e.g. mabe up-to-but-not-including the argparse change could be one PR, or even more fine break up).

**Comment by hamogu** [comment]
No need to break up. I thin I can deal with 1000 lines at a time.

File: sherpa/optmethods/tests/test_opt_unit.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+    xmin = [-1000, -1000]
+    xmax = [1000, 1000]
+    simp = SimplexRandom(func=rosenbrock, npop=4, xpar=x0, xmin=xmin,
+                         xmax=xmax, step=x0 + 1.2, seed=SEED,
+                         factor=10)
+
+    # After the first line, which is the x0 array + function value,
+    # the next three rows have 2 random numbers and then the function
+    # value at that coordinate, and then sorted on the last column.
+    # So this tests how repeatable the RNG us.
```
hamogu, line 63:
```suggestion
    # So this tests how repeatable the RNG is.
```

sherpa/optmethods/tests/test_opt_original.py
```diff
@@ -0,0 +1,955 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""This code used to be in sherpa.optmethods.opt.
+
+It was presumably to make testing the optimisers easier, but was never
```
hamogu, line 23:
```suggestion
It was apparently meant to make testing the optimisers easier, but was never
```

hamogu (approved):
I agree with this code change. It's clean up; unglamerous, but helpful. It's hidden deeply enough into Sherpa that there is no reason to fear that user (expect us) has ever noticed it.
I also agree with leaving the optimization tests there and not running them on CI. This is needed only for development work within the optimizers. They are sufficiently separate that e.g. a change to the plotting code does not touch the numerics of the optimizer and those type of tests are notoriously finicky on CI, e.g. in virtual machines you don't know if the number of CPUs you get is real, or all systems give you only one CPU ever, and those type of tests also tend to break with insignificant changes (e.g. MacOS vs. Linux) and some take a (comparatively) long time to run - so run them only if working on optimizers.

DougBurke:
Ta

PR #1781, opened 2023-05-24
utils: remove the numpy_convolve routine

DougBurke:
# Summary

Remove the sherpa.utils.numpy_convolve routine.

# Details

I noted a while back (Sep 2018, in #303) that numpy_convolve did not seem to be used in our code base. This commit removes it. I note that prior to the comment changes I made in #303 this code does not appear to have been significantly changed since it was originally added in the first git commit (April 2015).

It could instead be marked as a deprecation, with removal coming at a later date, but for such a relatively-simple routine is it worth it?

### DIFF
--- a/docs/overview/utils.rst
+++ b/docs/overview/utils.rst
@@ -7,10 +7,10 @@ The sherpa.utils module
 .. automodule:: sherpa.utils
 
    .. rubric:: Functions
-               
+
    .. autosummary::
       :toctree: api
-               
+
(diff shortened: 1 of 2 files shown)


hamogu (approved):
Agree to remove it. It's not clear how it's supposed to be used anyway and there is also e.g. `np.convolve` or `scipy.signal.fftcolvolve` - there really is no reason for sherpa to provide this utility.

Issue #1764, opened 2023-05-11
Coordinate systems and images


DougBurke:
We can create a `Data2D` or `DataIMG` object with independent axes in whatever coordinate system we want (just like we can with `Data1D` and `Data1DInt`).

# Issue 1

We do not have a good `notice` / `ignore` system for `Data2D`. We do allow spatial filtering (with regions) for `DataIMG`. Should we fix/improve/just look at the 2D case?

# Issue 2 (the main one)

For `DataIMG` we add the concept of coordinate system, in an analogous manner to adding channel, energy, and wavelength to the `DataPHA` class, but the internal implementation is very different, and so the considerations don't quite match.

For one thing, we can create a `DataIMG` object - so we can do things like spatially filter with regions - by creating an object with an independent axis in "any" units (e.g. a physical-like system where we don't start at (1,1)) without using a transform object. However, we can then technically still call `set_coord` to change the values and it gets complicated. 

We could do something like "check the independent axes are really logical, so start at (1,1)" when a user creates an object with physical and/or world transforms, but this would be restrictive (we don't require the logical axes to start at 1,1) and hard to validate.

Is it worth 

- worrying about
- documenting
- trying to clean up

This is not a high priority for me, but I wanted to write this down as an issue before I forget, as I mention it in #1762

Issue #1762, opened 2023-05-11
what is load_image doing with the coord argument?


DougBurke:
This is based on helpdesk ticket 024656 handled by @nplee 

In the following, the "log" dataset is behaving as we'd expect. The "phys" dataset claims it's in a physical system but has returned logical coordinates for the independent axis (I would have expected `x0` and `x1` to match the `ll0` and `ll1` arrays, not `l0` and `l1`).

```
sherpa-4.15.0> load_image("phys", "acisf07999_000N001_r0035_regevt3_srcimg.fits", coord="physical")

sherpa-4.15.0> load_image("log", "acisf07999_000N001_r0035_regevt3_srcimg.fits", coord="logical")

sherpa-4.15.0> get_coord("log")
'logical'

sherpa-4.15.0> get_coord("phys")
'physical'

sherpa-4.15.0> l0, l1 = get_indep("log")

sherpa-4.15.0> x0, x1 = get_indep("phys")

sherpa-4.15.0> l0[0:10]
array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

sherpa-4.15.0> l1[0:10]
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])

sherpa-4.15.0> x0[0:10]
array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

sherpa-4.15.0> x1[0:10]
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])

sherpa-4.15.0> set_coord("log", "physical")

sherpa-4.15.0> ll0, ll1 = get_indep("log")

sherpa-4.15.0> ll0[0:10]
array([3062.31005859, 3063.31005859, 3064.31005859, 3065.31005859,
       3066.31005859, 3067.31005859, 3068.31005859, 3069.31005859,
       3070.31005859, 3071.31005859])

sherpa-4.15.0> ll1[0:10]
array([4333.12988281, 4333.12988281, 4333.12988281, 4333.12988281,
       4333.12988281, 4333.12988281, 4333.12988281, 4333.12988281,
       4333.12988281, 4333.12988281])
```

It may be intended, but it looks like a bug.

DougBurke:
The current `load_image` docstring can be seen at https://sherpa.readthedocs.io/en/4.15.0/ui/api/sherpa.astro.ui.load_image.html and doesn't add any clarity. It calls

[...]

PR #1785, opened 2023-05-26
Add docs to Sherpa data classes

hamogu:
# Summary
Add docstring to most Sherpa dataclasses

# Details
This doc only PR adds docstrings to some of the Sherpa dataclasses. I wanted to used one of them in an SDS analysis, and found that I did not actually know how to set them up. So, looked in the code to find out and added docstrings for next time. I also tried to make up some reasonable examples for the more complex cases. 

There is a lot of apparent change in `pytest.ini`, but that's just because `sherpa/astro/data.py` does pass docstests. Before, `sherpa/astro` was all excluded. In order to get the examples in the Data classes tested, I had to spell everything else out, making this file longer.


sherpa/astro/data.py
```diff
@@ -5379,6 +5425,66 @@ def filter_region(self, data):
...
+    the number of bins in the flattened data array::
+
+        >>> import numpy as np
+        >>> x = np.random.normal(size=1000, loc=1.2)
+        >>> y = np.random.normal(size=1000)
+        >>> xrange = np.arange(-2, 4.1, 0.5)
+        >>> yrange = np.arange(-2, 2.1, 0.5)
+        >>> hist, x0edges, x1edges = np.histogram2d(y, x, bins=(yrange, xrange))
+        >>> x0lo, x1lo = np.meshgrid(x0edges[:-1], x1edges[:-1])
+        >>> x0hi, x1hi = np.meshgrid(x0edges[1:], x1edges[1:])
```
hamogu, line 5482: In this example, `x0lo` ... etc that I pass in are effectively the physical coordinate system, not the logical (they are not integers from 0..n_pixel). That's not an uncommon use case and, in the absence of WCS in this example make sense to me. But is it confusing as an example?

I wrote is like this because I think it explains the use case and then, once I understand #1786 , I can add another example *with* a WCS and explain the difference.
However, I feel it's better to get a short PR with these docs in now then to wait for something in the future.

sherpa/astro/data.py
```diff
@@ -4892,7 +4892,53 @@ def unsubtract(self):
...
+        >>> from sherpa.astro.data import DataIMG
+        >>> import numpy as np
+        >>> x1, x0 = np.mgrid[20:30, 5:20]
+        >>> datashape = x0.shape
+        >>> y = np.sqrt((x0 - 10)**2 + (x1 - 31)**2)
+        >>> x0 = x0.flatten()
+        >>> x1 = x1.flatten()
+        >>> y = y.flatten()
+        >>> image = DataIMG("bimage", x0=x0, x1=x1, y=y, shape=datashape)
+    '''
```
DougBurke, line 4941: There's a few things to think about here (and also in the `DataIMGInt` case you mention below). They don't need to all (or maybe any) addressed here, but I think we should have some understanding of the issues. Some of the following are highly interlinked (or can be read as "you've already said this"):

- technically this example is not the "logical" coordinate system as CIAO thinks of it, because that starts at 1,1 (defining the middle of the first pixel) following the FITS standard.
- however, in practice it only really matters when we have set up the `sky` and/or `eqpos` transforms, which we don't here (but I claim we do actually want to understand what me mean, as I discuss in the last point)
- It's not immediately obvious to me what `DataIMG` adds compared to `Data2D` for this particular use case (there are definitely differences, including the notebook auto-display code, as well as when sending data to DS9, it's just I can't enumerate them)
- on a related note, I struggle to understand why/when we need to send in the `shape` argument
- can we come up with an understanding of when using `DataIMG` is more appropriate than `Data2D`?
- coming back to this example, and the question you raise about the `DataIMGInt` example: in #1764 I have tried to ask the question, but probably not directly enough, of 
  - what do we mean by the x0 and x1 axis coordinates and how does that interact with any `coord` argument oyuo give to the constructor (because, if not careful, we have problems like #1762)
  - what happens when we don't have a transform

sherpa/data.py
```diff
@@ -1986,6 +2019,63 @@ def xhi(self):
...
+    Of course, the same structure can also be used for regularly-gridded data,
+    it just has to be passed in as a flattened array. However, in this case
+    the more specialiazed `DataIMG` class might be more appropriate.
+
+    Parameters
+    ----------
+    name : string
+        name of this dataset
+    x0 : array-like
+        Coordinates in one dimension
```
DougBurke, line 2037: Perhaps

```suggestion
        Coordinates of the first dimension of the independent axis.
```

but this is also ungainly and doesn't quite sound right

sherpa/data.py
```diff
@@ -1986,6 +2019,63 @@ def xhi(self):
...
+    x1 : array-like
+        Coordinates in the other dimension
```
DougBurke, line 2039:
```suggestion
        Coordinates of the second dimension of the independent axis.
```

I don't like "other dimension" here, but am struggling to come up with better words.

sherpa/astro/data.py
DougBurke: There is some text in `Data2D` and `Data2DInt` that suggests why you might prefer to use `DataIMG` / `DataIMGInt` which might better be here (or in both places)?

hamogu:
I tried to address all comments by rewording the docstrings, except for the note on coordinate systems. That really warrants further explanation, but won't get it before #1786 is resolved and I don't think we should wait for it here. Unfortunately, the related questions "What does `coord` do when I set it in object creation?" Is linked to that. As far as I can see, it only sets the default for what is used when I call `get_axes`.

There really are only two things that the IMG images add over the 2D case: WCS and region filtering (Ok, and notebook display, but that's not important at this point, I think). If you don't use either, Data2D is just as good and looks less confusing, because it does not ask for the WCS information and it's clear that the x0/x1 info I pass in are coordinates that will be used for fitting. I believe the same is true for `DataIMG` because the coord only specifies how the *display* is done, but not the fitting.

hamogu:
>  As far as I can see, it only sets the default for what is used when I call get_axes.

Mmm. Not true.

DougBurke:
So, there's two pieces here

- when calling `DataIMG`, what are the "units" of `x0` and `x1`
- once you have a `DataIMG` object, how do you convert between the two

I believe, but have not done the due dilligence, to believe that our code assumes that when calling `DataIMG` that `x0` and `x1` are always in logical units. This makes sense when you are also sending in `sky` and `eqpos` transforms. So, problems arise here if you then try to call `DataIMG` and then set `coord` to a value other than `logical` - what does this really mean? Are you saying `x0` and `x1` are in the given system, or are we to assume `x0` and `x1` are in `logical` units and we just want to change the system to use the given value?

However, we can actually just send in any values for `x0` and `x1` -e.g. we can send in `sky`-like coordinates. Problems arise in this situation if you try to change the `coord` system.

I have added documentation as I go along, but without really understanding this, so I have likely contributed to the confusion. 

I don't think the user in either case of my imaginary example is doing anything wrong, but I do think we need to understand what the various use cases are, and then what we want to do to support them (whether it's just documentation-only or has some code changes).

DougBurke:
The test failure here in https://github.com/sherpa/sherpa/actions/runs/5156005382/jobs/9286405966?pr=1785 looks suspiciously like the last comment I made in #1763 - namely

> I also added a fix for the error found when investigating https://github.com/sherpa/sherpa/issues/1762 - that is, an error call could fail with an AttributeError rather than raise the correct error. Then there's a fix for an issue I found when looking at the lack of coverage of this routine: trying to read in a Data2D object would fail, so this has been fixed (and a test added).

I don't have the time/energy to check up on this - did I miss a case with the `Data*Int` classes? (or why it's only run CI run that shows this problem)

hamogu:
I opened #1789 with a through discussion on what we might or might not want to do to fix current bahavior. However, in the meantime, I think we should just put in those docstrings and examples that document the current behavor. I added a cautionary note to one of them where appropriate. Those objects are not used often (or users would have asked us how to use them), so I'm not worried about documenting something that's not perfect and might be changed. I think it's more important (for us and users) to write down what IS happening and what parameters are used here.

So, (assuming I "fixed" the CI by switching back to logical coordinate system for my example) I suggest to merge something like this and improve it later as we improve the implementation. Note that the doctests add as a regression test of sorts that guarantee that the description in the docs is in fact what the objects does so we will notice if we change that.

In the light of #1789 I decided that it's out of scope to add full regression tests for the current behavior. As I said above, I stumbled on this because I'm using those classes for an actual analysis (the change in the soft ACIS PSF) and I found that I did not know how to initialize them. While working on that, I'll gain some more experience what functionality would be useful and that might help with #1789.

sherpa/astro/data.py
```diff
@@ -185,7 +185,7 @@ def _notice_resp(chans, arf, rmf):
 
         # If the response is mis-matched, determine which energy bins in the
         # RMF correspond to energy bins in the ARF and which are noticed.
-        # Propogate the noticed RMF energy bins to the ARF energy  bins.
+        # Propagate the noticed RMF energy bins to the ARF energy  bins.
```
wmclaugh, line 188:
```suggestion
        # Propagate the noticed RMF energy bins to the ARF energy bins.
```

sherpa/data.py
```diff
@@ -1807,8 +1823,25 @@ def get_yerr(self, filter=False, staterrfunc=None):
 
 
 class Data1DInt(Data1D):
-    """
-    1-D integrated data set
+    """1-D integrated data set.
+
+    This class is designed to handle non-consequtive bins.
```
wmclaugh, line 1828:
```suggestion
    This class is designed to handle non-consecutive bins.
```

sherpa/data.py
```diff
@@ -1986,6 +2019,64 @@ def xhi(self):
...
+    x0 : array-like
+        Independent coordinate values for the first dimension
+    x1 : array-like
+        Independent coordinate values for the second dimension
+    y : array-like
+        The values of the dependent observable. If this is a numpy masked
+        array, the mask will used to initialize a mask.
+    shape : tuple
+        Shape of the data grid for regularly gridded data (optional). This is
+        used return the data as an image e.g. for display,
```
wmclaugh, line 2045:
```suggestion
        used to return the data as an image e.g. for display,
```

sherpa/data.py
```diff
@@ -2161,9 +2252,74 @@ def notice(self, x0lo=None, x0hi=None, x1lo=None, x1hi=None, ignore=False):
...
+        Lower bounds of the bins in the first dimension of the independent coordinate
+    x1lo : array-like
+        Lower bound of the bins in the second dimension of the independent coordinate
+    x0hi : array-like
+        Upper bound of the bins in the first dimension of the independent coordinate
+    x1hi : array-like
+        Upper bound of the bins in the second dimension of the independent coordinate
+    y : array-like
+        The values of the dependent observable. If this is a numpy masked
+        array, the mask will used to initialize a mask.
```
wmclaugh, line 2279:
```suggestion
        array, the mask will be used to initialize a mask.
```

sherpa/astro/data.py
```diff
@@ -4892,7 +4892,61 @@ def unsubtract(self):
...
+    '''Image data set
+
+    This class builds on `sherpa.data.Data2D` to add support for
+    region filters and the ability to describe the data in different coordinate
+    systems, such as the "logical" coordinates (pixels of the image),
+    the "physical" coordinates (e.g. detector coordinates), and the "world"
+    coordinates (e.g. Ra/Dec on the sky).
+    Unlike `sherpa.data.Data2D`, this class does not support sparse data or data
+    with a non-rectangular shape.
+
```
DougBurke, line 4904: I'm not sure this is true. I can create a sparse image

```python
DataIMG("example", [1, 1, 2], [2, 3, 2], [12, 14, 2])
```

There are some things you can't do (without a shape attribute) but you can do some things, like filter the data:

```
sherpa In [1]: load_arrays(1, [1, 1, 2], [2, 3, 2], [12, 14, 2], DataIMG)

sherpa In [2]: image_data()
DataErr: data set '' does not specify a shape

sherpa In [3]: notice2d("circle(1.5, 2.5, 1)")
dataset 1: Field() -> Circle(1.5,2.5,1)

sherpa In [4]: get_dep()
Out[4]: array([12, 14,  2])

sherpa In [5]: get_dep(filter=True)
Out[5]: array([12, 14,  2])

sherpa In [6]: notice2d()
dataset 1: Circle(1.5,2.5,1) -> Field()

sherpa In [7]: notice2d("circle(1.5, 2.5, 0.4)")
dataset 1: Field() -> Circle(1.5,2.5,0.4)

sherpa In [8]: get_dep(filter=True)
Out[8]: array([], dtype=int64)
```

hamogu: Mmm. What is this class good for then? I'll reword to make sure waht it says is actually correct, but all the added functionality is for non-sparse images. I guess it's technically possible to do sparse image, but then it's just like using the parent class.

sherpa/astro/data.py
hamogu: That text does not apply here. As you said in a different location, I can use `DataIMG` for sparse data. There is just no benefit in the `DataIMG` in that case, it simply acts the same way as the parent class. In contrast, if I have anon-sparse image, the parent class still doesn ot give me any of the non-sparse functionality, so in that case you DO want to use a different (i.e. this) class.

hamogu: To the list of questions:

- I agree that all that needs to be resolved, but not in this PR. Here, I'm simply documenting what the current input paramters hare. Some of that is ambiguous, precisely because we do not have the answer to those questions. In those cases, I've kept the description in the docstring short; it can be expanded when we have better answers, but at least a user (i.e. me when I opened that PR) knows hwere to start and what the arguments are called.
- Technically, what I'm putting in is the "physical" coordinate system (numbers from 5-20 and 20-30). And what exactly is supposed (or does) happen wit hcoordinates at that point is fuzzy, so this is one of those locations where I say little and hope that we can expand the docstring once we resolve the coordinate issues.
- I woudl argue that `DataIMG` is more appropriate if you (a) have multiple coordinate systems (e.g. image, physcial, sky), or (b) want regions filters, or (c) niceties like notebook display for non-sparse images.
-

hamogu:
I think all that's left are comments related to the coordinate system like this:

> However, we can actually just send in any values for x0 and x1 -e.g. we can send in sky-like coordinates. Problems arise in this situation if you try to change the coord system.

I want to leave those for #1785 and not address them here in the interest of getting this in. I think those questions need more thinking.

sherpa/astro/data.py
```diff
@@ -5379,6 +5433,84 @@ def filter_region(self, data):
...
+        >>> image = DataIMGInt("binned_image",
+        ...                    x0lo=x0lo.flatten(), x1lo=x1lo.flatten(),
+        ...                    x0hi=x0hi.flatten(), x1hi=x1hi.flatten(),
+        ...                    y=hist.flatten(), shape=hist.shape)
+
+    Note that in this example, we end up with a "logical" coordinate system
+    in ``image`` and no WCS system to convert it to anything else. On the other hand,
+    in FITS standard terminology, the "logical" coordinate system is the
+    "image", counting pixels starting at 1, while here the ``x0lo``` and ``x1lo``
+    actually start at -2. This behavior works for now, but might be revisited.
```
DougBurke, line 5512: Do we want to add a comment like this to the `DataMG` case above?

DougBurke (approved):
I have one small comment (basically, should we copy over some text from the DataIMGInt example to the DataIMG one) but other than that this looks to improve things. There's plenty of things to do, but they are much longer-term projects so we should take the documentation improvements here.

sherpa/astro/data.py
hamogu: done.

hamogu:
I've copied that comment to `DataIMG`, too.

hamogu:
I've rebased this on #1804 and #1806 to generate the plots shown at doc build time. However, I can revert that if we want to take these changes without building the docs on an installed version of sherpa.

PR #1791, opened 2023-06-06
Tests: ensure repeatable test of get_draws

DougBurke:
# Summary

Ensure tests of get_draws use a fixed seed to make the results reproducible. There is no functional changes in this commit.

# Details

This was noted when working on #1772 and one CI run failed. I note that it will need updating once #1735 gets approved but that is a problem for #1735.

The `test_covar_as_none` and `test_covar_as_argument` had the possibility of failing (seen very occasionally) since they did not fix the random seed. So the tests have been updated to

- use a fixed seed
- use the same routine for checking the result, to point out that the aim of these tests is to show the results are the same

Unfortunately, we can't easily check this because

- the fit results, and hence covariance matrices, are different when using `cash` vs `cstat` (not hugely so, but different enough)
- numerical precision when sending in the covariance matrix vs grabbing it from the fit

So we don't actually do much checks of the response. We do add a check on the accept flags as they appear to be constant for this seed choice.

There are also two minor code cleanups in this test

- remove a print statement I'd added long ago when debugging something
- switch to using `SherpaVerbosity` for the module-level hide_logging fixture

### DIFF
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -136,7 +136,7 @@ def check_draws(stats, accept, params):
 
       a) numerical differences, since we only give the covar matrix
          to 8 dp
-      b) the fits, and hence parameter choices, depent on the chosen
+      b) the fits, and hence parameter choices, depend on the chosen
          statistic, leading to subtly different values
 
     So, all we check is
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2012, 2015, 2016, 2018, 2019, 2020, 2021, 2022
+#  Copyright (C) 2012, 2015, 2016, 2018, 2019, 2020, 2021, 2022, 2023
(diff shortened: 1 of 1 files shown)


sherpa/ui/tests/test_ui.py
```diff
@@ -131,44 +125,72 @@ def test_no_covar(stat, clean_ui, setup_covar):
...
+def check_draws(stats, accept, params):
+    """Check the test_covar_as_none/argument output
+
+    Ideally we'd be able to check that test_covar_as_none and
+    test_covar_as_argument return the same answers, but
+
+      a) numerical differences, since we only give the covar matrix
+         to 8 dp
+      b) the fits, and hence parameter choices, depent on the chosen
+         statistic, leading to subtly different values
```
hamogu, line 140:
```suggestion
      b) the fits, and hence parameter choices, depend on the chosen
         statistic, leading to subtly different values
```

hamogu (approved):
I'm approving already despite not knowing for sure if the spelling should be "depent" or "depend". I believe the latter, but I leave that to the native speakers. Either way, it's a comment in the test and functionally irrelevant.

PR #1793, opened 2023-06-08
XSPEC move from the xsFortran API to the FunctionUtility API (where possible)

### LINKED ISSUE #1225: XSPEC: change from FORTRAN/C to C++ for utility routines

DougBurke:
# Summary

Internal change to the XSPEC interface to use the FunctionUtility C++ API rather than the XSFortran API. Fix #1225.

# Details

I've found (since writing all this) a new-ish document providing some level of documentation on the C++ and FORTRAN interfaces: https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/internal/XspecInternalFunctionsGuide.html

Now we have XSPEC 12.12.0 as the minimum-supported version we can use the FunctionUtility C++ API rather than the xsFortran API. To simplify things, this PR just focuses on switching APIs. It does not try to take advantage of the API that would let us do things we currently can't do, as that is left for #1615 . I also explored removing all references to `xsFortran.h` but the table-model support changed significantly between XSPEC 12.12.0 and 12.12.1 and it's too hard to write code that will run with 12.12.0. I will note that I think the code changes are worthwhile doing (the new code is simpler) even without really changing much for the user.

The XSPEC documentation doesn't provide a changelog for when features are added or changed for these APIs, so I am assuming that we are compatible with XSPEC 12.12.0, as I'm currently developing against 12.13.0 (but at the time this code was written I did have access to 12.12.0 builds). I am relying on the CI runs to check for backwards compatibility (and I have done some spot checks to see what is present in 12.12.0 and I believe we should be okay).

Technically there should be no differences, but there is at least one: prior to this we have

```
>>> xspec.set_xsxsect("not-a-number")

***XSPEC Error:  No Initializer for  cross section table named not-a-number
  File "/home/dburke/micromamba/envs/ciao415-py310/lib/python3.10/site-packages/sherpa/astro/xspec/__init__.py", line 497, in set_xsxsect
    _xspec.set_xsxsect(name)
ValueError: could not set XSPEC photoelectric cross-section to 'not-a-number'
```

whereas we now hide the "XSPEC Error" line:

```
>>> xspec.set_xsxsect("not-a-number")
  File "/lagado.real/sherpa/sherpa-main/sherpa/astro/xspec/__init__.py", line 815, in set_xsxsect
    _xspec.set_xsxsect(name)
ValueError: could not set XSPEC photoelectric cross-section to 'not-a-number'
```

I don't think this is worth highlighting in the release notes or documentation (as we've never guaranteed what happens when we call an XSPEC routine I claim we can make changes with abandon for features people are not likely to hit).

# Notes

In a discussion with HEASARC (August 23 2023) when talking about moving over to FunctionUtility or staying with the xsFortran interface, the suggestion was

> It would be a good idea to move away from xsFortran. I try to add everything in FunctionUtility to xsFortran but it is not always possible.

### DIFF
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -23,6 +23,9 @@
 #include <iostream>
 #include <fstream>
 
+// Documentation for the "XSPEC internal functions" is at
+// https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/internal/XspecInternalFunctionsGuide.html
+//
 // xsfortran is only needed to support FNINIT as there's (as of XSPEC 12.13.0)
 // no version of this functionality in FunctionUtility.
 //
@@ -63,13 +66,6 @@ static int _sherpa_init_xspec_library()
   }
 
   // Redirect the stdout channel for the duration of the FNINIT call.
-  // The replacement is set to ostringstream so that we can find out
-  // what was sent to it in case of emergency (but would need code
-  // changes to do this). This used to be done by manually replacing
-  // the std::cout buffer.
-  //
-  // Note that only FNINIT needs this, as the FunctionUtility calls do
-  // not create string output (at least for 12.12.x/12.13.0).
   //
   std::ostream* outStream = IosHolder::outHolder();
   std::ostringstream tmpStream;
@@ -78,7 +74,7 @@ static int _sherpa_init_xspec_library()
 			IosHolder::errHolder());
 
   try {
-    // Initialize XSPEC model library
+    // Initialize XSPEC model library.
     FNINIT();
 
   } catch(...) {
@@ -86,7 +82,11 @@ static int _sherpa_init_xspec_library()
 			  outStream,
 			  IosHolder::errHolder());
 
-    // Raise appropriate error message that XSPEC initialization failed.
+    // The contents of tmpStream could be inspected to see if it
+    // contains useful information for the user, but at this point of
+    // the initialization it is not obvious that it would provide any
+    // extra information.
+    //
     PyErr_SetString( PyExc_ImportError,
 		     (char*)"XSPEC initialization failed; "
 		     "check HEADAS environment variable" );
@@ -97,15 +97,11 @@ static int _sherpa_init_xspec_library()
 			outStream,
 			IosHolder::errHolder());
 
-  // We used to set the chatter to 0 but this misses useful info
-  // (like XSPEC can't find the data files) that would be reported
-  // by XSPEC, so use the default XSPEC setting. It does not appear
-  // that this value is read in from ~/.xspec/Xspec.init so set
-  // it here so we have repeatable behavior.
+  // Set a number of values to their XSPEC defaults (as of XSPEC
+  // 12.14.1, but they have not changed for a long time). It appears
+  // these are not read in from the user's ~/.xspec/Xspec.init file.
   //
   FunctionUtility::xwriteChatter( 10 );
-
-  // Set cosmology initial values to XSPEC initial values
   FunctionUtility::setH0( 70.0 );
   FunctionUtility::setq0( 0.0 );
   FunctionUtility::setlambda0( 0.73 );
--- a/sherpa/astro/xspec/src/_xspec.cc
+++ b/sherpa/astro/xspec/src/_xspec.cc
@@ -23,90 +23,19 @@
 #include <iostream>
 #include <fstream>
 
-// The symbols listed in XSPEC version 12.9.1
-// at https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixExternal.html
-// are given below. Note that this is the C/FORTRAN interface, not the
-// more-featureful FunctionUtility module.
+// xsfortran is only needed to support FNINIT as there's (as of XSPEC 12.13.0)
+// no version of this functionality in FunctionUtility.
 //
-// Functions which are used below:
-// FNINIT	Initializes data directory locations needed by the models. See below for a fuller description.
-// FGABND	Get an element abundance.
-// FGCHAT	Get current chatter level setting for model functions' output verbosity.
-// FPCHAT	Set the chatter level. Default is 10, higher chatter levels produce more output.
-// FGMSTR	Get a model string value (see XSPEC xset command).
-// FPMSTR	Set a model string value.
-// FGDATD	Get the model .dat files path.
-// FPDATD	Set the model .dat files path.
-// FGMODF	Get the model ion data path.
-// FPSLFL	Load values of a file solar abundance table (see abund command).
-// FGSOLR	Get the solar abundance table setting.
-// FPSOLR	Set the solar abundance table.
-// FGXSCT	Get the cross section table setting.
-// FPXSCT	Set the cross section table.
-// csmgh0	Get the cosmology H$_0$ setting (see the cosmo command).
-// csmph0	Set H$_0$.
-// csmgl0	Get $\Lambda_0$.
-// csmpl0	Set $\Lambda_0$.
-// csmgq0	Get q$_0$.
-// csmpq0	Put q$_0$.
-// xs_getVersion (or xgvers)	Retrieve XSPEC's version string.
-//
-// Functions which are not wrapped as their functionality is available:
-// RFLABD	Read abundance data from a file, then load and set this to be the current abundance table. (Essentially this combines a file read with the FPSLFL and FPSOLR functions.)
-//
-// Functions not wrapped as not felt to be that useful:
-// fzsq	Computes the luminosity distance, (c/H$_0$)*fzsq. The function is valid for small values of q$_0$*z for the case of no cosmological constant and uses the approximation of Pen (1999 ApJS 120, 49) for the case of a cosmological constant and a flat Universe. The function is not valid for non-zero cosmological constant if the Universe is not flat.
-//
-// Functions not wrapped since they are not useful as is (they need
-// functionality from 12.9.1 to set the XFLT keywords):
-// DGFILT	Get a particular XFLT keyword value from a data file.
-// DGNFLT	Get the number of XFLT keywords in a data file.
-//
-// Other symbols in xsFortran.h are:
-// DGQFLT       Does a XFLT keyword exist?
-// PDBVAL       Set a database value
-//
-// Symbols in 12.9.1/HEASOFT 6.22 but not in 12.9.0/HEASOFT 6.19
-// FGABNZ
-// FGTABN
-// FGTABZ
-// FGELTI
-// FGNELT
-// FGABFL
-// FPABFL
-// FGAPTH
-// FPAPTH
-// csmpall
-// DPFILT
-// DCLFLT
-// GDBVAL
-// CDBASE
-// FGATDV
-// FPATDV
-//
-// These seem unlikely to be useful for Sherpa
-// xs_getChat
-// xs_write
-// xs_read
-//
-// These are numeric functions which we should have available elsewhere
-// xs_erf
-// xs_erfc
-// gammap
-// gammq
-//
-// TODO:: switch to C++ FuntionUtility interface rather than use xsFortran.h
-//
+#include <XSFunctions/Utilities/xsFortran.h>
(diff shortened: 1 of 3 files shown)


DougBurke:
I tried to remove the xsFortran.h include in `xspec_extension.hh`, as it's only needed for the `tabint` code, but it turns out that XSPEC 12.12.0 does not have `FunctionUtility::tableInterpolate`, and although there's likely a way to access it it is not going to be "nice", so I'm dropping this idea for now.

```
% git diff sherpa/include/
diff --git a/sherpa/include/sherpa/astro/xspec_extension.hh b/sherpa/include/sherpa/astro/xspec_extension.hh
index 9232a8c1..48afb3b7 100644
--- a/sherpa/include/sherpa/astro/xspec_extension.hh
+++ b/sherpa/include/sherpa/astro/xspec_extension.hh
@@ -34,21 +34,8 @@
 
 #include "sherpa/fcmp.hh"
 
+#include <XSFunctions/Utilities/FunctionUtility.h>
 #include <XSFunctions/Utilities/funcType.h>
-#include <XSFunctions/Utilities/xsFortran.h>
-
-// The table interface is defined in XSPEC 12.12.1 in xsFortran.h
-// (and uses const parameters) but it is not defined in 12.12.0.
-//
-#ifndef XSPEC_12_12_1
-extern "C" {
-
-  void tabint(float* ear, int ne, float* param,
-             int npar, const char* filenm, int ifl,
-             const char* tabtyp, float* photar, float* photer);
-
-}
-#endif
 
 
 namespace sherpa { namespace astro { namespace xspec {
@@ -811,31 +798,54 @@ static void create_output(int nbins, T &a, T &b) {
         xspecTableModelTabint( ) : xspecTableModelBase(1) { }
         void call_xspec( std::vector<FloatArrayType>& fear,
                          FloatArray& pars, npy_intp npars,
-                         char* filename, char*tabtype,
+                         char* filename, char* tabtype,
                          FloatArray& result ) {
-          tabint( &fear[0], this->npts, &pars[0], npars,
-                  filename, this->ifl, tabtype,
-                  &result[0], &this->error[0] );
-        }
-      }; // class xspecTableModelTabint
-
-      //
-      // In theory the following should work but have not tested, yet
-      //
-      // template<XSpecFuncVal XSpecFunc>
-      // class xspecTableModel : public xspecTableModelBase {
-      // public:
-      //   xspecTableModel( ) : xspecTableModelBase(1) { }
-      //   void call_xspec( std::vector<FloatArrayType>& fear,
-      //                    FloatArray& pars, npy_intp npars,
-      //                    int char* filename, char*tabtype,
-      //                    FloatArray& result ) {
-      //     XSpecFunc( &fear[0], this->npts, &pars[0], filename,
-      //                this->ifl, &result[0], &this->error[0] );
 
-      //   }
+         // For the moment copy the xsFortran tabint routine, but a
+         // future enhancement would be to move to an interface that
+         // uses XSPEC RealArray arguments directly (this is something
+         // we could take advantage of with a number of other models).
+         //
+         // From include/xsType.h we have
+         //
+         //     typedef double Real;
+         //     typedef std::valarray<Real> RealArray;
+         //
+         // but Sherpa arrays are (FloatArray is specific to this
+         // file, the DoubleArray is defined in sherpa/extension.hh)
+         //
+         //      typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
+         //      typedef float FloatArrayType;
+         //
+         //      typedef double SherpaFloat;
+         //      typedef sherpa::Array< double, NPY_DOUBLE > DoubleArray;
+         //      typedef DoubleArray SherpaFloatArray;
+         //
+         string fileName(filename);
+         string initString("");
+         string tableType(tabtype);
+
+         const size_t npts1 = this->npts + 1;
+         RealArray energyArray(.0, npts1);
+         RealArray parameters(npars);
+         RealArray fluxArray, fluxErrArray;
+
+         for (size_t i=0; i<npts1; i++) energyArray[i] = fear[i];
+         for (size_t i=0; i<(size_t)npars; i++) parameters[i] = pars[i];
+
+         FunctionUtility::tableInterpolate(energyArray, parameters, fileName, ifl,
+                                           fluxArray, fluxErrArray, initString,
+                                           tableType, true);
+
+         for (size_t i=0; i<(size_t)this->npts; i++) result[i] = fluxArray[i];
+         if ( fluxErrArray.size() > 0 ) {
+           for (size_t i=0; i<(size_t)this->npts; i++) this->error[i] = fluxErrArray[i];
+         } else {
+           for (size_t i=0; i<(size_t)this->npts; i++) this->error[i] = 0.0;
+         }
 
-      // }; // class xspecTableModel
+        }
+      }; // class xspecTableModelTabint
 
 
```

DougBurke:
@wmclaugh unfortunately you are on the hook for this as this is firmly in the C++ camp. There is a follow-up PR - #1615 - but that is still experimental and I wanted to get these changes into the system sooner rather than wait for the follow-up work. It is also a chance to make any C++ changes we might want, but I'm no C++ guru so I am not really making significant changes.

DougBurke:
Rebased to address a simple test conflict and because of recent updates to the XSPEC versions ran in CI

hamogu:
needs review by someone with C++ knowledge, target date for review: middle of 2024

DougBurke:
Rebased to address a conflict in #1615 which is based on this PR.

DougBurke:
Rebased to address for conflicts in #1998 but also to pick up other build changes.

DougBurke:
Rebased as #1997 has been merged (there's no conflicts, but I just want to make sure we test the latest version).

DougBurke:
Interesting. I had to re-run the macOS build two times to get the tests to pass - they just segfaulted during the XSPEC tests. I'm guessing there's a memory issue in a XSPEC model, but it's really hard to pin this particular issue down (previous debugging scars to show it)...

DougBurke:
Rebased on top of #2052

DougBurke:

DougBurke:
Rebased to address a conflict in #1615

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -133,262 +62,118 @@ static int _sherpa_init_xspec_library()
...
-  // Stream buffer for redirected stdout
-  std::streambuf* cout_sbuf = NULL;
-  std::ofstream fout;
+  // Redirect the stdout channel for the duration of the FNINIT call.
+  // The replacement is set to ostringstream so that we can find out
+  // what was sent to it in case of emergency (but would need code
+  // changes to do this). This used to be done by manually replacing
+  // the std::cout buffer.
+  //
+  // Note that only FNINIT needs this, as the FunctionUtility calls do
```
wmclaugh, line 71: Does this only need to be handled for FNINIT? I see the same stream handling applied elsewhere in this file (although for STDERR instead of STDOUT)

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -133,262 +62,118 @@ static int _sherpa_init_xspec_library()
...
-  char version[256];
-  int retval;
-
-  try {
+  IosHolder::setStreams(IosHolder::inHolder(),
+			outStream,
+			IosHolder::errHolder());
 
-    retval = xs_getVersion(version, 256);
+  // We used to set the chatter to 0 but this misses useful info
```
wmclaugh, line 100: I would suggest changing 'misses' to something along the lines of drops or obscures since the 0 level chatter actually filters out the info as intended.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: I'll have to review, as it's been a while since I last looked at it. My guess is

1. I had hoped that we could avoid this entirely
2. but we can't, so we could wrap it neatly
3. but then you realize some other routine needs something similar and you don't update the existing documentation

DougBurke:
I've hopefully cleaned up the documnetation for the initialization step. I've also added a link to the new XSPEC documentation on the xsFortran/FunctionUtility interface at https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/internal/XspecInternalFunctionsGuide.html

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -79,19 +79,10 @@ namespace sherpa { namespace astro { namespace xspec {
 typedef sherpa::Array< float, NPY_FLOAT > FloatArray;
 typedef float FloatArrayType;
 
-// Try and support the use of std::transform while still building
-// against C++-98 compilers.
-//
-#if __cplusplus > 199711L
-#define CONVERTARRAY(orig, out, npts)					\
+// Assume std::transform is available on any system we support
```
wmclaugh, line 82: The source fails to compile on my mac intel running clang++ 15.0.0 with std=c++98. 

    In file included from sherpa/astro/xspec/src/_xspec.cc:22:
    sherpa/include/sherpa/astro/xspec_extension.hh:631:11: error: expected expression
          CONVERTARRAY(this->ear, fear);
          ^
    sherpa/include/sherpa/astro/xspec_extension.hh:85:24: note: expanded from macro 'CONVERTARRAY'
                       [](const double val) -> FloatArrayType { return static_cast<FloatArrayType>(val); });

When compiling without your changes  my builds pick up the else case of the macro (ie. __cplusplus == 199711L) when defining CONVERTARRAY.  Replacing lines 82-85 with the following works for my build:

float d2f_convert (double dd) { return static_cast<float>(dd); }
#define CONVERTARRAY(orig, out)                                        \
        std::transform(std::begin(orig), std::end(orig), std::begin(out), d2f_convert);

DougBurke: gah. But yay for testing!

DougBurke: One question is whether we want to support building with `--std=c++98`? We don't really document the supported C++ or C versions, but I thought we were aiming for something like `--std=c++11`?

DougBurke: Of course, the easiest solution is to just revert this particular change.

wmclaugh: On my machine, clang (15.0.0) defaults to using __cplusplus == 199711L (the c++98 standard) and ignores the CXXFLAGS env var. Direct specification of '-std=c++11' on the commandline appears successful from the clang commandline.   As such, I would not suggest not dropping support of c++98 standard until we have a compelling reason to do so (ie. we need to take advantage of features only supported in a newer standard). I'm OK with reverting the code which has the ifdef to not use lambda expressions (introduced in c++11) when compiling against c++98.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: Hopefully the documentation for this routine is better. It's actually a "less is more" since much of the comments were describing how it has changed behaviour, but we don't really care about the old code now.

sherpa/include/sherpa/astro/xspec_extension.hh
DougBurke: So I've reverted this commit.

wmclaugh (approved):

DougBurke:
Thanks for the review @wmclaugh

PR #1795, opened 2023-06-10
Add docs to describe the inputs for the ``calc`` function of a user model

hamogu:
This is another one like #1785: I just need to *use* Sherpa to do some SDS work before I can get back to *developing* Sherpa, but unfortunately, I did not find how to use what I needed, so I had the read the code and thus I decided to make it into a PR, so I know next time.

## Summary
Add docs that describe the arguments of the `calc` function for user models.

## Details
While at it, I slightly modernized the existing example to use `super()` and inlined them in the rst docs, because that way I can run it through doctest-plus. These are the kind of docs we want to have tested to make sure what we write here still works after any changes, so I think the ability to run doctest-plus outweighs the slight inconvenience of having a lot of code verbatim in an rst file. For a user, it looks the same, since the code included is rendered to html the same way anyway.


DougBurke:
For reference, I have some not-really-doc-changes-but-essentially-doc-hanges related to the `calc` method in the first commit of #1671 (they were hard to find/point out but now #1617 has been merged they are much easier to find) which might be useful to think about at the same time - it's basically saynig that we expect

```python
```

because in some code we have code like

```python
        p = args[0]
        xargs = args[1:]
```

and it probably is better to always send in the parameters separately from the grid (aka `args`).

hamogu:
I've updated the example to follow that suggestion. Logically, that should have gone into the first commit in this PR, but I accidentally added it to the second. It's not wrong and I'd like to not spend most my brain power on re-ordering git commits for something this trivial, so please just review to joint PR. 

I don't think we want to prescribe that this is how it HAS to be done (because it's Python, it does not have to be done like that, whatever we do, we can't stop users from writing `calc(self, *args)` or `calc(self, pars, x1, awefad, *args)` if they want to; but we *can* be consistent in code and example and that's what I've done here.

DougBurke:
For reference I've broken out the commit I was talking about into #1797

DougBurke:
The reason I had the poly and trap code separate was

- for ease of testing / development
- to make it easy to recreate the figure(s)

With the doctest changes then we can at least test the code compiles,  and I guess that it runs since we do a fit with the poly model, although I am always suspicious about tests where I can't definitely see them being run ...

I know you have a plan to be able to regenerate all the plots at runtime, but we are not there yet, so do we want to leave the original code and have a note saying that it should be kept up-to-date with the docs?

docs/model_classes/usermodel.rst
```diff
@@ -201,11 +282,49 @@ It also shows one way of embedding models from a different system,
 in this case the
 `two-dimemensional polynomial model 
 <https://docs.astropy.org/en/stable/api/astropy.modeling.polynomial.Polynomial2D.html>`_
-from the AstroPy package.
+from the AstroPy package::
 
-.. literalinclude:: ../code/poly.py
+.. doctest-requires:: astropy
```
DougBurke, line 287: This is being rendered in the text, so is the code actually being tested?

image

docs/model_classes/usermodel.rst
```diff
@@ -21,17 +21,194 @@ external code.
...
+  it supports combining models (e.g. by addition or multiplication) and
+  a cache to reduce evaluation time at the expense of memory use.
+  (`~sherpa.models.model.ArithmeticConstantModel` and `ArithmeticFunctionModel` are less general
+  than `~sherpa.models.model.ArithmeticModel` and can be used to represent
+ a constant value or a function.)
+
+* `~sherpa.models.model.RegriddableModel1D` and
+  `~sherpa.models.model.RegriddableModel2D` which build on
+  `~sherpa.models.model.ArithmeticModel` to allow a model to be
+  evaluated on a different grid to that requested.
```
DougBurke, line 46:
```suggestion
  evaluated on a separate grid and then re-binned or interpolated onto the requested grid.
```

Not sure that's better.

docs/model_classes/usermodel.rst
```diff
@@ -21,17 +21,194 @@ external code.
...
+implements an ``__init__`` method that defines the model parameters
+and a ``calc`` method that, given the parameter values and the coordinates
+at which the model is to be evaluated, returns the model values.
+
+Calculating the model values
+============================
+
+Many Sherpa models are set up to work with both integrated and non-integrated data,
+some might even work with data of different dimensionality. Because of that, the input
+to the ``calc`` method can take several forms:
```
DougBurke, line 58: So, maybe we should say that the default signature is (or should be, I haven't bothered to actually check `Model.calc`)

```python
```

(or use `pars` but need to check with e.g. #1797) and then it's up to the model to decide what to do - i.e. if you have a 1D model you could say

```python
def calc(self, p, x, xhi=None, **kwargs):
```

One concern I have is that we could also have this as one of

```
def calc(self, p, x0, x1=None, **kwargs):
def calc(self, p, lo, hi=None, **kwargs):
... etc ...
```

and so we (as in Sherpa) need to be careful to make sure we don't use named arguments when evaluating models.

If you want to sound Computer-Sciencey then you can throw around [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) but I think our code base is well beyond that.

docs/model_classes/usermodel.rst
```diff
@@ -21,17 +21,194 @@ external code.
...
+    ...     ampl
+    ...        The amplitude of the central (flat) segment (zero or greater).
+    ...     center
+    ...        The center of the central segment.
+    ...     width
+    ...        The width of the central segment (zero or greater).
+    ...     slope
+    ...         The gradient of the slopes (zero or greater).
+    ...     """
+    ...
```
DougBurke, line 194: So, this code/documentation was written before (I hope before) #900 landed, which added some degree of dimension-checking to models. We should probably add a

```python
ndim = 1
```

definition here (and `ndim = 2` to the 2D model). And then mention it earlier in the text.

Note that this is (currently) only used to ensure that models are combined corrcetly (so you can't add 1D + 2D models), and not when it cmoes to model evaluation - that is, to check the arguments that are sent in (because it's hard to do, unless we had something like the structured array approach talked about in #1796 might help).

docs/model_classes/usermodel.rst
hamogu: It didn't show up when I build locally and checked, but now I can see it locally, too. I must have changed something, but not sure what but I'll find it and fix.

Either way though, the test runs (if astropy is installed), which I verified by putting an error in the code and checked that it fails.

docs/model_classes/usermodel.rst
hamogu: These examples derive from `RegriddableModel1D/2D` and don't have to specify `ndim` because they inherit it form their parent class:
```Python

In [1]: from sherpa.models import model
   ...: class LinearModel(model.RegriddableModel1D):
   ...:     def __init__(self, name='linemodel'):
   ...:         self.slope = model.Parameter(name, 'slope', 1, min=-10, hard_min=-1e20, max=10, hard_max=1e20)
   ...:         super().__init__(name, (self.slope, ))
   ...: 
   ...:     def calc(self, pars, *args, **kwargs):
   ...:         if len(args) == 1:
   ...:             return args[0] * pars[0]
   ...:         elif len(args) == 2:
   ...:             xlo, xhi = args
   ...:             return (xlo + xhi) / 2 * pars[0]
   ...:         else:
   ...:             raise ValueError('This model only works in 1 dimension')
   ...: 

In [2]: import numpy as np
   ...: linear = LinearModel()
   ...: 

In [3]: linear.ndim
Out[3]: 1
```

That's why I did not set it here; however, I'll take the suggestion to add it in the text and explain (a) what it is and (b) why it's not set here.

hamogu: OK, I think it stems from a missing black line higher up (something like "sphinx warning: bulleted list did not end on empty line"), which confused it. It's fixed in my local version, I'll check again after I push.

docs/model_classes/usermodel.rst
hamogu: OK, I think it stems from a missing black line higher up (something like "sphinx warning: bulleted list did not end on empty line"), which confused it. It's fixed in my local version, I'll check again after I push.

docs/model_classes/usermodel.rst
hamogu: > we (as in Sherpa) need to be careful to make sure we don't use named arguments when evaluating models.

One more reason to change to an interface that gets the data in some more structured form (#1796), but here I'm just writing down what we do now. I'll add a suggestion with the "default" way to write that, but we don't have to enforce it.

docs/model_classes/usermodel.rst
DougBurke: I'd forgotten that I'd added that default to `RegriddableModel1D/2D`.

For background/any-one reading at  home, we use it in the `*OpModel` class to check when creating a combined model - so you'd get an error when you try to combine

```python
linear + sherpa.models.basic.Gauss2D()
```

DougBurke:
Can this be rebased/updated to match the current code now that

- #1806
- #1808 
- #1813

have been merged?

hamogu:
Yes, it should be updated, but it's not going to happen while the Chandra workshop is still going on.

hamogu:
I rebased this and tests still pass and the [plot appears in the docs](https://sherpa--1795.org.readthedocs.build/en/1795/model_classes/usermodel.html).

DougBurke (approved):
Thanks for all the "unrelated" PRs that got done because of this PR!

hamogu:
And this one is only so I understand how to make a new model that I need to model the recent broadening of the soft PSF on ACIS!

PR #1797, opened 2023-06-12
model: explicity send parameter argument to calc

DougBurke:
# Summary

Ensure that the calc method for models begins with the parameter array (explicitly given) rather than extracting it from the named arguments sent to the routine. This should not change any user code.

# Details

Be expllicit about the required parameter array sent to the calc method of the Model class. This is mainly for documentation, as it shouldn't change the results of any of the existing code, and for future development.


hamogu:
As a matter of process, I suggest we merge this and #1795. They go hand in hand, but were written independently and touch different files, so there is no conflict.

PR #1800, opened 2023-06-14
utils: minor code clean up

DougBurke:
# Summary

Internal clean up of the sherpa.ui.utils module.

# Details

Created when working on something else.

- Clean up the use of _clean_model (make more use of it) and add a little documentation.
- Avoid using a fully-qualified name when we have the symbol imported anyway.
- use `idval` rather than `id` for an internal variable name (although the API uses `id` so that is left as is)
- clean up of logging to use lazy formatting (%s) rather than f-strings
- change a logging call to use warning(..) rather than info("WARNING: ...")

Technically this last one is a user change - it they had used `SherpaVerbosity("WARNING")` then the message is now seen when it wasn't before, but I don't think we need to explain this. Or we could also just not take the last commit, where this change is made.

# Example

## CIAO 4.15

```
sherpa In [1]: load_arrays(1, [1, 2, 3], [1, 2, 3])

sherpa In [2]: set_source(xsphabs.gal * xsapec.src)

sherpa In [3]: guess(gal)
WARNING: No guess found for xsphabs.gal

sherpa In [4]: guess(src)

sherpa In [5]: guess(gal + src)
WARNING: No guess found for (xsphabs.gal + xsapec.src)

sherpa In [6]: from sherpa.utils.logging import SherpaVerbosity

sherpa In [7]: with SherpaVerbosity("ERROR"):
          ...:     guess(gal)
          ...: 

sherpa In [8]: with SherpaVerbosity("WARNING"):
          ...:     guess(gal)
          ...: 
```

## This PR

```

In [2]: load_arrays(1, [1, 2, 3], [1, 2, 3])

In [3]: set_source(xsphabs.gal * xsapec.src)

In [4]: guess(gal)
WARNING: No guess found for xsphabs.gal

In [5]: guess(src)

In [6]: guess(gal + src)
WARNING: No guess found for (xsphabs.gal + xsapec.src)

In [7]: from sherpa.utils.logging import SherpaVerbosity

In [8]: with SherpaVerbosity("ERROR"):
   ...:     guess(gal)
   ...: 

In [9]: with SherpaVerbosity("WARNING"):
   ...:     guess(gal)
   ...: 
WARNING: No guess found for xsphabs.gal
```

### DIFF
--- a/sherpa/ui/tests/test_ui.py
+++ b/sherpa/ui/tests/test_ui.py
@@ -19,6 +19,7 @@
 #
 
 from collections import namedtuple
+import logging
 
 import numpy as np
 
@@ -293,7 +294,7 @@ def test_ui_set_full_model_checks_dimensions_match(clean_ui, setup_ui_2d):
 
 # Bug 12644
 @requires_data
-def test_ui_source_methods_with_full_model(clean_ui, setup_ui_full):
+def test_ui_source_methods_with_full_model(clean_ui, setup_ui_full, caplog):
 
     ui.load_data('full', setup_ui_full.ascii)
     ui.set_full_model('full', 'powlaw1d.p1')
@@ -311,8 +312,21 @@ def test_ui_source_methods_with_full_model(clean_ui, setup_ui_full):
                        match="Convolved model\n'powlaw1d.p1'\n is set for dataset full. You should use get_model_plot instead."):
         ui.get_source_plot('full')
 
-    # Test Case 2
-    ui.set_source('full', 'powlaw1d.p2')
+    # Test Case 2.
+    #
+    # The hide_logging fixture hides the logging output, so we need
+    # to re-enable it for this call.
+    #
+    assert len (caplog.records) == 0
+    with SherpaVerbosity("INFO"):
+        ui.set_source('full', 'powlaw1d.p2')
+
+    assert len (caplog.records) == 1
+    logname, loglvl, logmsg = caplog.record_tuples[0]
+    assert logname == "sherpa.ui.utils"
+    assert loglvl == logging.WARNING
+    assert logmsg == "Clearing convolved model\n'powlaw1d.p1'\nfor dataset full"
+
     ui.get_source('full')
 
     # Test Case 3
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -8955,16 +8955,14 @@ class Session(NoNewAttributesAfterInit):
             try:
                 model.guess(*self.get_data(idval).to_guess(), **kwargs)
             except NotImplementedError:
-                # TODO: change to warning() from info()
-                info('WARNING: No guess found for %s', model.name)
+                warning('No guess found for %s', model.name)
             return
 
         ids, f = self._get_fit(idval)
         try:
             f.guess(**kwargs)
         except NotImplementedError:
-            # TODO: change to warning() from info()
-            info('WARNING: No guess found for %s', self.get_model(idval).name)
+            warning('No guess found for %s', self.get_model(idval).name)
 
     def calc_stat(self, id=None, *otherids):
         """Calculate the fit statistic for a data set.
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -6652,11 +6652,10 @@ class Session(NoNewAttributesAfterInit):
         """
         if model is None:
             id, model = model, id
-        if _is_str(model):
-            model = self._eval_model_expression(model)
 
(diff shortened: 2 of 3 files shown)


hamogu:
I agree with the last commit. For my own education, why do you prefer "%s" over `format` for logging? I mean, I con't care, but I'd like to learn what makes you prefer one over the other here.

sherpa/ui/tests/test_ui.py
```diff
@@ -311,8 +312,21 @@ def test_ui_source_methods_with_full_model(clean_ui, setup_ui_full):
                        match="Convolved model\n'powlaw1d.p1'\n is set for dataset full. You should use get_model_plot instead."):
         ui.get_source_plot('full')
 
-    # Test Case 2
-    ui.set_source('full', 'powlaw1d.p2')
+    # Test Case 2.
+    #
+    # The hide_logging fixture hides the logging output, so we need
+    # to re-enable it for this call.
+    #
```
hamogu, line 319: I do note the that `hide_logging` docstring says, "ignore info", but then sets the lefel to ERROR (meaning it hides INFO *and* WARNING). If `hide_logging` did what it's advertizes as, then we would not have to reset that here, because this is WARNING level. 
However, if you want, we can leave as is - this is just a test where 100% consistency is not required; we are allowed to do inconsistent setups as it makes sense for each test.

sherpa/ui/utils.py
```diff
@@ -6850,6 +6840,26 @@ def delete_model(self, id=None):
         self._sources.pop(id, None)
 
     def _check_model(self, model):
+        """Return a Model instance.
+
+        Converts a string to the model instance if necessary.
+
+        Parameters
+        ----------
+        model : str or a sherpa.models.model.Model object
```
hamogu, line 6849: For rendering in sphinx (if this method is even shown there, but even if not now, maybe we'll at it in the future)
```suggestion
        model : `str` or a `sherpa.models.model.Model` object
```

sherpa/ui/utils.py
```diff
@@ -6850,6 +6840,26 @@ def delete_model(self, id=None):
...
+
+        Returns
+        -------
+        model : sherpa.models.model.Model instance
```
hamogu, line 6853:
```suggestion
        model : `sherpa.models.model.Model` instance
```

hamogu:
Apart from the three small comments, I think this is good to go; I'll approve once those are resolved. (Please ping if I miss it.)

sherpa/ui/utils.py
DougBurke: I've given up trying to work out when symbols get automatically annotated or not and just ignored it completely!

DougBurke:
For https://github.com/sherpa/sherpa/pull/1800#issuecomment-1599140480 :+1: 

For our code I don't think there's really much difference between the two lines

    info("foo bar %s blah blah", self.get_foo_bar(bob))
    info(f"foo bar {self.get_foo_bar(bob)} blah blah")

but pylint likes to warn about it with it's [logging not lazy](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-not-lazy.html) warning. You can tell it not to warn about the use of f-strings for this (but then I don't understand why they allow this form, which is definitely not lazy, but that's for another day) so we could go with the latter.

One "good" thing with this change is that it made me find and fix an issue with our logging override code, but that's been done now, so it's mainly a "style" thing and not something I think we need to codify (i.e. other people don't need to write logging calls like I do).

sherpa/ui/utils.py
DougBurke: For reference, this is not referenced / made visible in the RTD docs

sherpa/ui/tests/test_ui.py
DougBurke: In this particular case I think it useful that we run this test at the INFO level, so we can check what we create for the "standard" user experience.

I don't have the energy/desire to fix up the hide_logging docstring to point out it hides warning too.

DougBurke:
Note that the tests fail because I haven't rebased this to include #1807

hamogu (approved):

sherpa/ui/tests/test_ui.py
DougBurke: I'm not planning to change this unless required - see below (as I didn't want the comment to get lost with future changes/rebases) https://github.com/sherpa/sherpa/pull/1800#issuecomment-1599268830

hamogu: I understand. That's why I approved this yesterday.

Issue #1349, opened 2021-11-12
Integrating models on high resolution grid


anetasie:
I'm getting back to looking at high resolution grid for model evaluation. This links to the discussion in #823. I expected to be able to use `calc_source_sum()` to get the predicted  values for the model on high resolution grid.  The calculation is done on high resolution grid correctly, but again there is a question of the expectation of the results. 

If we have a box model with the constant amplitude set to 10 and  calculated an integral over the space delta(10) - the integral should evaluate to 10x10 = 100 regardless of the bin size. This is not the case and I'm just getting really confused.

```
dataspace1d(1,100,2,dstype=Data1DInt)
set_model(box1d.b1)
b1.xlow=10
b1.xhi=40
b1.ampl.max=100
b1.ampl=10
fake()
calc_source_sum(10,21)  
calc_model_sum(10,21)
calc_source_sum(10,11)
calc_model_sum(10,11)

dx = 0.1
xgrid = np.arange(1, 100, dx)
grid_lo = xgrid[:-1]
grid_hi = xgrid[1:]
rb1 = b1.regrid(grid_lo,grid_hi)
set_model(rb1)
calc_source_sum(10,20)
calc_model_sum(10,20)
```

I get the following results for model on standard or high resolution grid, so for `b1` or `rb1` models. The results for model being 110 or 90 are due to the large dx=2 bin size. I expected to get 100 for model on high resolution grid which is not limited to dx=2, but is defined on dx=0.1

```
sherpa In [21]: calc_model_sum(10,20)
Out[21]: 110.0

sherpa In [22]: calc_source_sum(10,20)
Out[22]: 47.5

sherpa In [25]: calc_model_sum(10,19)
Out[25]: 90.0
```

anetasie:
I guess this is a better output

```
sherpa In [43]: calc_source_sum(11,20)
Out[43]: 45.0

sherpa In [44]: calc_model_sum(11,20)
Out[44]: 100.0
```

Issue #1300, opened 2021-09-15
normalization of plot_source_component on high resolution grid


anetasie:
This is split from the issue #822 - the grid was fixed by PR #823, however there is a remaining question about the normalization in the plots in the binned data fit on high resolution grid described in details in PR #823.

DougBurke:
I believe we are looking at the wrong part of the software stack with the regrid code. We need to be looking at issues like

- #829 
- #840 
- #945 

before we bother with commands like plot_source

anetasie:
@DougBurke thanks for the summary of the `regrid` issues here. I just created a new label `area: regrid` o mark this type of issues.

Issue #840, opened 2020-06-25
2D regrid gives crazy results - is it mixing up x0/x1 axes?


DougBurke:
I'm trying out the regrid support for 2D models. I believe I'm doing the right thing here, but get crazy data when I try to rebin to a higher resolution. I am not doing any PSF convolution here,so it should act as roughly the identity function, depending on exactly what the rebinning is doing. If I give it a regrid that exactly matches the input it does work.

```
import numpy as np
import pytest

from sherpa.models.basic import Const2D, Gauss2D

from matplotlib import pyplot as plt


y0, x0 = np.mgrid[20:29, 10:20]
y0 = y0.flatten()
x0 = x0.flatten()

gmdl = Gauss2D()
gmdl.fwhm = 14
gmdl.xpos = 15
gmdl.ypos = 24
gmdl.ampl = 10

cmdl = Const2D()
cmdl.c0 = 4

# Does the cache affect anything? It shouldn't
gmdl._use_caching = False
cmdl._use_caching = False


xr1 = np.arange(10, 20, 1)
yr1 = np.arange(20, 29, 1)

rmdlg = gmdl.regrid(xr1, yr1)
rmdlc = cmdl.regrid(xr1, yr1)

shape = (9, 10)
truthg = gmdl(x0, y0).reshape(shape)
truthc = cmdl(x0, y0).reshape(shape)

ansg = rmdlg(x0, y0).reshape(shape)
ansc = rmdlc(x0, y0).reshape(shape)

if (ansc != truthc).any():
    print("const model difference")

if (ansg != truthg).any():
    print("gauss model difference")

# use a finer grid: 0.5 rather than 1
xr2 = np.arange(10, 19.1, 0.5)
yr2 = np.arange(20, 28.1, 0.5)

rmdl2g = gmdl.regrid(xr2, yr2)
rmdl2c = cmdl.regrid(xr2, yr2)

ans2g = rmdl2g(x0, y0).reshape(shape)
ans2c = rmdl2c(x0, y0).reshape(shape)

if ans2c != pytest.approx(truthc):
    print("const model difference 2")

if ans2g != pytest.approx(truthg):
    print("gauss model difference 2")
```

When run you get

```gauss model difference 2```

and if I plot up the `truthg` and `ans2g` images I get respectively:

```
In [7]: plt.imshow(truthg, origin='lower')
Out[7]: <matplotlib.image.AxesImage at 0x7f4a12ae3e80>
```

[...]

PR #1803, opened 2023-06-15
Improvements for regrid support for combined models

### LINKED ISSUE #1802: RecursionError when trying to regrid "complex" models
With CIAO 4.15 we have

```
sherpa In [1]: from sherpa.models.basic import Const1D

sherpa In [2]: (Const1D() + Const1D()).regrid([1, 2, 3])
Out[2]: <RegridWrappedModel model instance 'regrid1d((const1d + const1d))'>

sherpa In [3]: (Const1D() * Const1D()).regrid([1, 2, 3])
Out[3]: <RegridWrappedModel model instance 'regrid1d((const1d * const1d))'>

sherpa In [4]: (Const1D() * (1 + Const1D())).regrid([1, 2, 3])
Out[4]: <RegridWrappedModel model instance 'regrid1d((const1d * (1.0 + const1d)))'>

sherpa In [5]: (Const1D() * Const1D() * (1 + Const1D())).regrid([1, 2, 3])
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Where is this recursion coming from? I have also see a similar issue when dealing with the `integrate` flag (when trying to get it to be supported for *OpModel classes).

If I try with python 3.11 / latest main I get the following:

```
>>> (Const1D() * Const1D() * (1 + Const1D())).regrid([1, 2, 3])
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

I can get a slightly-different error if I try the less-well-used division operator

```
>>> (1 / Const1D()).regrid([1, 2, 3])
<RegridWrappedModel model instance 'regrid1d((1.0 / const1d))'>
>>> (1 / (1 + Const1D())).regrid([1, 2, 3])
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 994 more times]
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1378, in regrid
    if not hasattr(part, 'regrid'):
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 577, in __getattr__
    lowered_name = name.lower()
                   ^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded while calling a Python object
``` 

but note that I can say:

```
>>> (Const1D() / (1 + Const1D())).regrid([1, 2, 3])
<RegridWrappedModel model instance 'regrid1d((const1d / (1.0 + const1d)))'>
```

Hmm, it's not due to `ArithmeticConstantModel` - which is used to represent numeric terms like `1` in the above, since this also fails

```
>>> (Const1D() * Const1D() * (Const1D() + Const1D())).regrid([1, 2, 3])
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-main/sherpa/models/model.py", line 1381, in regrid
    return part.__class__.regrid(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

DougBurke:
I've reworked how this is done, which has lead to bigger changes. I haven't yet updated the documentation to reflect this.

# Summary

Improve behavior of the `regrid` method when called with "complex" models (these with multiple components). Fix #1802 

# Details

Allows UnaryOpModels to be regridded and fixes #1802 (in a way that works for both BinaryOp and UnaryOp models). It is only a small part of the changes needed to get the regrid code to handle the cases we care about.

Adding support for UnaryOpModel:

- change from ArithmeticModel to RegriddableModel as a parent class (all that RegriddableModel does is add the `regrid` method)
- add a `regrid` method that follows the approach of BinaryOpModel, which delegates the choice to the first valid component it finds (this is easier than the binary case as we only have one "part" to check, but it can itself be a complex model so we use the same approach as BinaryOpModel and just look for the `parts` attribute rather than using a specific `is instance` check [*]

Fixing #1802 for BinaryOpModel:

The issue is discussed in #1802, which I repeat here. The problem was that the `regrid` method would look at the two components (`lhs`, `rhs`) via the `parts` attribute. It then tries to find the first class which has a `regrid` method, and applies that class method to `self`. The trouble is when one of these two parts is itself a `BinaryOpModel`, because then it will find it's own code and call it, which ends up in an infinite loop, and hence `RecursionError`. The solution I came up with was to recognize that if a part has a `parts` attribute then we need to recurse through them (which can then recurse if it's a highly-nested expression) until we fit an object with a `regrid` attribute we can call.

Two possible alternatives are

- we can do this recursing when we create the `BinaryOpModel` instance and then store it via a `_regrid` field (for instance), which would make `regard` a lot cleaner and would likely stop so much inspection (since at each level of finding a `*OpModel` we know that will already have a `_regrid` setting). I only came up with this idea after submitting the code, and have decided not to pursue it since a) this code seems to work, b) there's no guarantee there isn't little daemons to deal with in this approach, and c) it means all users pay the price, eve if they never call `regard` (although the costs of all approaches is not expected to be huge)

- an alternative is that we could use the `ndim` attribute of the model to decide whether we call `RegriddableModel1D` or `RegriddableModel2D`. This would be **much** simpler but it combines concerns that should be separate, and would not allow sub-classing of these models (or using an alternative method that provides the same `regrid` interface).

[*] An earlier commit, where we only had support for BinaryOpModel, used is instance, but that was on purpose there as at that time it wasn't valid for UnaryOpModel 

# Background

For those that have not really looked at the regrid code before, I apologize. This change is pretty small, but it may be useful to have a larger-picture view. It also touches on elements I don't want to actually address in this particular PR.

We have the `RegriddableModel` class which adds the `regrid` "interface", by which we can ask for a new model that acts as the original model does but it will evaluate the model on the grid sent to `regrid` and will then "rebin" or interpolate onto whatever user grid is given. The actual implementation is done in`RegriddableModel1D` or `RegriddableModel2D`. It was originally restricted to single models (e.g. gauss1d) and not composite models (e.g. gauss1d + powlaw1d), but this is rather restrictive [**]. In #919 Dan added support for calling `regrid` for a "binary" model - that is `BinaryOpModel`) based on initial prototyping done by me, but it was not correct as discussed in #1802. We want to fix this, and to add support for `regrid` to the `UnaryOpModel` class, as this is the other main way to combine models (much-less-used than `BinaryOpModel`). So, the work here is that for a `regrid` call to a "container" class (`BinaryOpModel` / `UnaryOpModel`), how do we find out what `regrid` method we want from the contained elements (via the `parts` array, which comes from the `CompositeModel` class)?

As an aside, the logic that `regrid` uses can be done by any user, so you don't technically need to have a model that provides the `regrid` method. So, we could just drop the idea of being able to regrid a composite model and tell users that they have to do it manually. I don't think this is a good idea, but it is something we could do. However, we still need to be able to handle regridding composite models (even if not automatically), so there are still going to be follow-up PRs!

[**] I claim that it only really makes sense to regrid a convolution-style model (since the only reason you want data from a different grid is because you are going to pass it to some model to "reprocess" the data into the energy range you care about), but I may be missing a use case, and even if I was right we could have something like `convmodel1(...) + convmodel2(...)` where you want both to use this fancy-grid-extension scheme, or just regrid the whole model as it's too awkward to identify just the convolution component.


hamogu:
I have a conceptual question. I guess I should be able to figure it out from the code, but it might be easier to just ask: Assuming we have two model components (say, in a `BinaryOpModel`) and one of the msupports regridding, and the other one does not. What does that mean? Can we regrid the combined model, or not?

Put it another way, I'm thinking of the complex model expression where "basic" models are leaves and `BinaryOpModel` or `UnaryOpModel` or `ConvolutionModel` are nodes. Now, does the entire tree fail to regrid, if there is just one node or leaf that cannot regrid? Or do we identify sub-trees that can regrid, evaluate all parts of that sub-tree on the new grid, and then, inrerpolate/rebin/whatever on the top node of that subtree (so that the output of the subtree is then on the "normal" (not regridded) grid, so that that normal grid can then be passed up into the next higher level (which may not support subgridding)?

Last, there are very few models that cannot support subgridding (there might bem ore where it's not been implemented yet, but that in principle can support that), right? All that comes to mind right now is a `TableModel` with x-axis set to `None`. I'm just trying to understand possible use cases and the fundamental limits.

Why am I asking all this? The code change itself is conceptually simple, but I have not looked (or inderstood) the regrid code before being asked to review this PR and I want to make sure that I'm not approving something that locks us into a solution which I later find out to not agree with what I believe to be the best.

DougBurke:
Good question. At the moment we don't really have a way to say "this model can be regridded and this model can not". We could say that only models that derive from the `RegriddableModel` class can be regridded, and then make the *OpModel classes either

- 1. fail when asked to regrid a composite model that does not handle this
- 2. try something really fancy and only regrid those components that support it, and not the others

If we went this way I'd argue for 1, not 2. Thinking about it I think 2 wouldn't be that hard, as it can be handled by over-riding the `calc` method of the `*OpModel` class, but I wonder if you'd get different results if you wrote models differently - e.g. `m1 * (m2 + m3)` vs `m1 * m2 + m1 * m3` when `m1` is not regriddable.

I do not know how many models we have that do not support regridding - I can only think of the table model case when you haven't given an independent axis (which is a highly-specialized case, and has issues if you filter the data, and ideally would be handled separately from the general table-model code, but that's for a different day), and is the problem case you identified too, but I have not done a deep dive on this.

At the moment I am assuming we do not require the `RegriddableModel` superclass, but I'll have a think about this.  One issue we have is that I don't think we've properly gone and updated `ArithmeticModel` to `RegriddableModel1D` (or `2D`) in all our models.

One thing I have in my mind is that this regrid support is only really relevant with a convolution model, otherwise what's the point. That is, we need a model like `fancy(m1 + m2 * m3)` where we want to evaluate `m1 + m2 * m3` on a different grid, use this when convolving it with whatever `fancy` does, and then regridding onto the original grid. In this case we would only need to add `regrid` support to the convolution model and forget all this complication. However, we could end up with `fancy1(some-model-expression) + fancy2(a-different-model-expression)` and want to regrid both `fancy1` and `fancy2` [or have a complicated expression where the convolution is somewhere within it] and just don't want to bother with identifying just the component to change. Also, maybe we do want to be able to mark a model as "un-regriddable", in which case the "only bother with convolution models" approach loses out.

I am not sure this is particularly helpful though!

hamogu:
This is a bit helpful, in the sense that it clarifies for me that we might need to think a bit harder on what we want. 

What's the difference between "regrid" and "evaluate on some grid"? Leaving aside the odd "Table model with no x axis" (which is needed in some cases, but clearly it's an odd case), all model *can* can be evaluated on different grids - you just need to set an ignore/notice and the input for x changes. How is that different from regridding? The "lower level" models just receive an array for x and return model values for that; that should work even without a regrid method, right?

So maybe the way to do it is to add the regrid method **only** to models that *cause* regridding to be used such as convolution models, i.e. make regrid a method for "call calc of the lower level models, but with this x grid"? So, regrid would then be something like:
```Python
def regrid(self, x, **kwargs):
    xbig = np.concatenate([
                     np.array([.7, .8, .9]) * x.min(),
                     x,
                     np.array([1.1, 1.2, 1.3]) * x.max()])
    ybig = self.calc(xbig, **kwargs)
    return np.interpolate(x, xbig, ybig)
```

All the lower level models (incl `UnOPModel` and `BinaryOpModel`) would then not need a regrid method because they are just called using the normal `calc` method for some value of x.

In fact, maybe there should not be a `regrid` method at all, but the `calc` method of models that require regridding simply do that as part of their normal `calc` method and then call lower-level models with an expanded grid. The user would not notice the difference.

This runs into problems with models like `fancy1(fancy2(model1 + model2))` in that fancy1 will just call fancy2 with a new grid, so how does fancy2 know of that grid is already regridded? It doesn't so it would regrid again, making the grid even wider and causing the result to be interpolated/rebinned twice with possibly loss of precision. But a-priory it's not clear how to solve this problems anyway - fanc1 and fancy might have different requirements on regridding, e.g. one adds only on the left of the x array and one only one the right. The scheme would work, it might just not be ideal.

Also, I would argue that those situations are rare, and we could pass down a flag "no more regridding" to lower levels.

DougBurke:
So, we already have the existing regrid implementation which is being used in prodiuction (for instance the Chandra Source Catalog), so we can't just break things. I believe the use that the CSC makes is only for 2D models [*], and as the 1D and 2D cases are technically separate we can do whatever we want for 1D, but I would want to have a consistent scheme/setup. There are also things I don't understand about the 2D case but that is, I believe, an implementation issue, and discussed in #840 and I think not relevant for the larger picture. 

Although I believe the regrid option only makes sense for convolution style models, I am reluctant to restrict it to just those models as

- I could be wrong (I know @anetasie has used the regrid option in use cases I don't understand)
- I want to support something like the XSPEC [energies](https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSenergies.html) command, which you apply to the whole "dataset/grid", and not just the relevant convolution component and lets you do the equivalent of `plot_source()` and see the model evaluated on the user-defined grid (but `plot_model()` shoes the model evaluated on the "data" grid, ie that provided by the RMF) [**]

I'm not sure I fully understand your issue about "what is special about regrid"? Yes, we should in general be able to evaluate a model on a different grid, that doesn't help you when you say "I need it evaluated on this grid, which is enforced by the data [such as the RMF for PHA data] but I want to internally do the analysis on a different grid and then convert to the 'data grid'". So we need some way to do this operation. But perhaps this is a red herring.

It may be that only being able to regrid convolution models is the cleanest approach, but: we still need to be work out what to do with composite models (the "future reference" point I am about to make is relevant there); as I said, it would be useful to know what @anetasie is actually doing to see if that involves convolution; breaking existing code; and whether we might want to easily view a whole model with the "over-ridden" grid rather than the "what we need to compare to data" grid. But I fear that we may be talking past each other so may have to actually get together in July or later to discuss.

I also want to point out for future reference that "regrid" can mean either

- interpolation onto a different grid
- rebin onto a different grid to conserve "flux"

and understanding what we want to do there depends in part on our discussions on the integrate flag.

---

[*] my understanding is that this is for a convolution-style model (a 2D PSF read in from a file being convolved with a model), so it should fit in the "convolution" narrative 

[**] although I want to be able to support something like XSPEC `energies`, I do not know how we could have `plot_source` use the user-defined grid nicely (it may be easy to do, but you'd not be able to switch between that and the user grid), and there are other issues when it comes to what people expect with `plot_model`with our current code - e.g. 
- #822
- #1300 
- #1349

hamogu:
I support "plotting on a high resolution grid". My point is that that's not conceptually a regrid (which to me means "evaluate on one grid and then rebin/interpolate?), but just an "evaluate on the a different grid and plot that" - and the latter is something that might need UI work in the `plot_model` function, but not in how the model is evaluated.
Of course, the trouble is getting the units right (flux per energy? per wavelength? per bin - and if so, which bin? energy flux or counts flux?)

So, the question is, for that vision, what's "regrid" actually supposed to do?

(And I fully agree with not breaking existing behavior **and** also being consistent between 1D and 2D, if at all possible. I'm not trying to convince you to break something, I'm trying to understand what it's supposed to do so that I can confidently say "yes, this PR works towards that goal".)

hamogu:
@DougBurke: I agree that a quick talk would probably help. 
@anetasie please let us know what you actually use regrid for.

DougBurke:
Hmm. Looking at this again I'm going to try the "detect which regrid class to use when the composite model is created", which is leading me on some interesting digressions into the existing class hierarchy - e.g. `CompositeModel` vs `ArithmeticModel`.

hamogu:
@anetasie What is the status of this? 
Last review note says

> @anetasie please let us know what you actually use regrid for.

Can you tell us your use case so that we can decide how to proceed?

hamogu (commented):
@anetasie ping: What is the status of this?

Issue #1508, opened 2022-05-10
Something odd with our tests and matplotlib handling


DougBurke:
With the current main branch, with matplotlib installed, and my `~/.sherpa-standalone.rc` file containnig

```
% grep plot ~/.sherpa-standalone.rc
# plot_pkg   : pylab dummy
plot_pkg   : pylab
# plot_pkg   : dummy
```

then `pytest` likes to tell me

```
WARNING: Failed to import any usable sherpa.plotting backend. Plotting routines will not be available.
```

which is odd. I've tracked it down to

```
% pytest sherpa/static
==================================================== test session starts =====================================================
platform linux -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/dburke/sherpa/sherpa-master, configfile: pytest.ini
plugins: forked-1.4.0, xdist-2.5.0, xvfb-2.0.0, doctestplus-0.12.0
collected 0 items

=================================================== no tests ran in 0.01s ====================================================
WARNING: Failed to import any usable sherpa.plotting backend. Plotting routines will not be available.
```

So, it tries to find tests in `sherpa/static`, doesn't find anything, but thinks we don't have matplotlib available.

It is **not at all clear** what is happening here, because if I change `sherpa/plot/__init__.py` to the following:

```
 for plottry in plot_opt:
+    lgr.warning(f"XXX  .... trying <{plottry}>")
     try:
         backend = importlib.import_module('.' + plottry,
                                           package='sherpa.plot')
+        lgr.warning("XXX found it")
         break
     except ImportError:
+        lgr.warning("XXX failed")
         pass
 else:
     # None of the options in the rc file work, e.g. because it's an old file
     # that does not have dummy listed
+    lgr.warning("XXX  .... falling over to dummy")
     import sherpa.plot.dummy_backend as backend

+lgr.warning("XXX done all the things")
+
```

then re-run the above we get

```
pytest sherpa/static
==================================================== test session starts =====================================================
platform linux -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/dburke/sherpa/sherpa-master, configfile: pytest.ini
plugins: forked-1.4.0, xdist-2.5.0, xvfb-2.0.0, doctestplus-0.12.0
collected 0 items

=================================================== no tests ran in 0.01s ====================================================
WARNING: XXX  .... trying <pylab_backend>
WARNING: XXX found it
WARNING: XXX done all the things
[...]

PR #1806, opened 2023-06-19
Active sphinx directive to make plots

hamogu:
## Doesn't work yet!

# Summary
Use the Sphinx plot directive to generate (some of) the plots in the docs. 

## Why do we want to do that?
There are several advantages to doing it that way:

- The plots are always up to date. If we change (either directly or indirectly by using a newer version of matplotlib) the default layout, e.g. the order of colors and the plots are regenerated every time the docs are built, then they will always be up to date.
- We don't have to litter the repro with binary png files.
- If we check for sphinx warnings / errors then those scripts act as additional tests: If we break a plot, the CI will fail. (Note: No changes to CI are included in this PR so far.)

Disadvantages:
- Well, if we don't check for sphinx warnings, we might break a plot, so it won't show up in the docs, and we won't know. Need to check.

Note that this is not an all-or-nothing. We can convert the simple plots to this scheme, but leave pre-rendered PNGs in the repro for cases where the plot comes at the end of an analysis with large data or long-running processes.

## Doesn't work yet

Despite all examples passing on the command line and in the jupyter notebook, I get some really strange failures when this hits a model plot. Plotting data is fine, but plotting models is apparently not.

```Python
   >>> import numpy as np
   >>> edges = np.asarray([-10, -5, 5, 12, 17, 20, 30, 56, 60])
   >>> y = np.asarray([28, 62, 17, 4, 2, 4, 125, 55])
   >>> from sherpa.data import Data1DInt
   >>> d = Data1DInt('example histogram', edges[:-1], edges[1:], y)
   >>> from sherpa.plot import DataPlot
   >>> dplot = DataPlot()
   >>> dplot.prepare(d)
   >>> from sherpa.models.basic import Const1D, Gauss1D
   >>> modl = Const1D('base') - Gauss1D('line')
   >>> modl.pars[0].val = 10
   >>> from sherpa.plot import ModelPlot
   >>> mplot = ModelPlot()
   >>> mplot.prepare(d, modl)
   >>> assert len(d.x) == 8
   >>> assert len(d.y) == 8
   >>> # It works up to here.
   >>> # dplot.plot() woudl also work
   >>> # bute the next line fails with
   >>> # ValueError: x and y must have same first dimension, but have shapes (8,) and (0,)
   >>> # despite passing the assert statements above
   >>> mplot.plot()
   >>> # The following lines works again:
   >>> dplot.overplot()
```

Full error is this:
```
/Users/guenther/code/sherpa/docs/plots/index.rst:33: WARNING: Exception occurred in plotting index-2
 from /Users/guenther/code/sherpa/docs/plots/index.rst:
  File "/Users/guenther/mambaforge/envs/sherpaalone/lib/python3.10/site-packages/matplotlib/sphinxext/plot_directive.py", line 517, in _run_code
    exec(code, ns)
  File "<string>", line 17, in <module>
  File "/Users/guenther/code/sherpa/sherpa/plot/__init__.py", line 1592, in plot
    Plot.plot(self, self.x, self.y, title=self.title, xlabel=self.xlabel,
  File "/Users/guenther/code/sherpa/sherpa/plot/__init__.py", line 299, in plot
    backend.plot(x, y, yerr=yerr, xerr=xerr,
  File "/Users/guenther/code/sherpa/sherpa/plot/pylab_backend.py", line 425, in plot
    objs = axes.plot(x, y,
  File "/Users/guenther/mambaforge/envs/sherpaalone/lib/python3.10/site-packages/matplotlib/axes/_axes.py", line 1632, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/Users/guenther/mambaforge/envs/sherpaalone/lib/python3.10/site-packages/matplotlib/axes/_base.py", line 312, in __call__
    yield from self._plot_args(this, kwargs)
  File "/Users/guenther/mambaforge/envs/sherpaalone/lib/python3.10/site-packages/matplotlib/axes/_base.py", line 498, in _plot_args
    raise ValueError(f"x and y must have same first dimension, but "
ValueError: x and y must have same first dimension, but have shapes (8,) and (0,)
```

That makes no sense to me. I've banged my head at a wall for several hours, I've revisited it later, banged my head again the wall again, and it still looks the same. At this point I'm uploading as a draft PR, just so there is a record of what I'm trying.

Maybe, just maybe, this resolves itself after #1382 where I change how backends are accessed. I can't point a finger on it, but it must have something to do with how the plot directive and our backend setting interact. Otherwise it would have to work, right?

Still, if this rings a bell with someone, I'm all ears.

[skip CI]


DougBurke:
I don't know if this is relevant, but in `docs/conf.py` we have  abunch of mocking going on:

DougBurke:
I know there's something odd with the tests and lpotting - see #1508 - but whether it's the same issue that is causing this I don't know

hamogu:
This addresses #816 (though we can argue if it should close that issue or if we want to leave it open until the vast majority of plots are generated this way).

hamogu:
The fist commit here is actually from #1808, so that could be reviewed here on separately there. 
Since it's not a big lien, I'm marking this as ready for review already; I don't care if the review it here altogether or separately in #1808 and here. ( I just developed it in separate PRs to get separate RTD runs and see that I made the configuration correctly.)

docs/plots/index.rst
```diff
@@ -54,16 +54,26 @@ Example
...
-    >>> edges = np.asarray([-10, -5, 5, 12, 17, 20, 30, 56, 60])
-    >>> y = np.asarray([28, 62, 17, 4, 2, 4, 125, 55])
+.. plot::
+   :context:
+   :nofigs:
+   :include-source:
+
+   >>> import numpy as np
+   >>> edges = np.asarray([-10, -5, 5, 12, 17, 20, 30, 56, 60])
+   >>> y = np.asarray([28, 62, 17, 4, 2, 4, 125, 55])
```
DougBurke, line 64: I am too lazy to go hunting in the sphinx documentation to find out why we need to mark up these code segments that are not generating plots. I assume it's because they are doing stuff (e.g. importing modules, setting up variables) that we need in late plot calls.

What I don't understand is why there's a plot directive for the example below which says

```
.. plot::
   :context:
   :nofigs:
   :include-source:

   >>> print(dplot.plot_prefs)
   {'xerrorbars': False, 'yerrorbars': True, 'ecolor': None, 'capsize': None, 'barsabove': False, 'xlog': False, 'ylog': False, 'linestyle': 'None', 'linecolor': None, 'color': None, 'marker': '.', 'markerfacecolor': None, 'markersize': None, 'xaxis': False, 'ratioline': False}
```

As far as I can see, we don't need to include this for plot purposes. Is it perhaps linked to the need to also run these through doctestplus?

hamogu: I am not 100% sure, so I will check again, but I think I did that to make it run in the same context where dplot is defined. My understanding is (but the might be wrong) that the code is split: Code under the plot directive is run to make the plots, and the other code is run separately by doctestplus. That other code would not know about `dplot`, because that's defined inside the plot directive code.

DougBurke: That sounds logical.

DougBurke:
I assume the directive we discussed above, for the "doesn't create a plot but is needed to create a plot" is the reason why we  now get "source code" links like

image

Not sure what I think of them just yet....

hamogu:
For each and every block, I can decide if I want that link to be shown or not by setting [options](https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html#options), in this particular case `:show-source-link: False` should do the trick (there is also a globe option to never show a link like that which is set in `conf.py` for all the sphinx docs). We just have to decide how exactly we want it to look.

Typically, we discuss that examples line by line, so maybe it makes sense to switch that off globally. I'll try that and push here so you can see if you like that.

DougBurke:
I think I prefer the "without adding a source link" by default (and then let us turn it no if we want it), but it's not a huge preference. We do have the "you can hide the prompts and screen output" option already enabled which means users can grab the code relatively easily, but it is more steps / less obvious.

DougBurke:
@hamogu - are you happy with this default now? If so I can approve the PR.

hamogu:
I am happy. 
If we change out mind, we can always change it later. This is not about backwards compatibility, it's just the look of the docs.

DougBurke (approved):
LGTM

PR #1807, opened 2023-06-20
NumPy 1.25 support

DougBurke:
# Summary

Remove deprecation warnings seen when using NumPy 1.25.0.

# Details

NumPy 1.25 deprecates converting a ndarray that is not a 0d value, so for our cases this is

    int(value)
    info("message %g", value)

where value is a ndarray of length 1. The simple fix is to explictly access the first element - so

    answer = int(value[0])
    info("message %g", value[0])

although for the int case (in the PSF convolution instrument code) I have not looked through to determine if value is always guaranteed to be a ndarray, so it is more like

    if np.isscalar(value):
        newval = value
    else:
        newval = value[0]
    answer = int(newval)

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1284,8 +1284,13 @@ class Fit(NoNewAttributesAfterInit):
         def get_par_name(ii):
             return self.model.pars[self.thaw_indices[ii]].fullname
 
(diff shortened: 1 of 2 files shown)


Marie-Terrell (approved):
Looks good to me!

Issue #1714, opened 2023-02-28
build_sphinx option is going away


DougBurke:
Not surprising with the distutils changes in Python 3.12, but

```
running build_sphinx
/home/dburke/miniconda3/envs/sherpa-xspec/lib/python3.10/site-packages/setuptools/_distutils/dist.py:992: RemovedInSphinx70Warning: setup.py build_sphinx is deprecated.
  cmd_obj.run()
Running Sphinx v6.1.3
```

DougBurke:
So, #1608 changes the documentation to tell users to

    cd docs
    make html

and the RTD setup doesn't actually run sphinx, just tells it where the files are, so it should not be a problem.

PR #1808, opened 2023-06-21
Use Sherpa install instead of code only to build docs

hamogu:
# Summary
This PR switches from building the documentations from the source directory, to building it from an installed version of Sherpa.

It's a prerequisite to running (and thus testing!) the plotting included in the docs at doc build time (#1806).

## Details
In the past, we have build Sherpa's documentation from the source directory, not from an installed version. That makes it much faster and easier to build, but it requires mocking out all C extensions, and thus disallows several useful features.

About 5 years ago, Sherpa's C code modules were mocked out because it was hard to build. Now, we've improved that quite a bit, so hopefully we can stop mocking some of those modules which would allow code like #1806 to go forward, leading to greater  consistency and testing of the documentation.

This PR includes the necessary changes to the ReadTheDocs setup and to the developer docs. For ease of review, the actual use of this feature is left for a followup PR (confusingly with a lower issue number, because I pulled this out after the fact):

- #1806


DougBurke:
The reason for mocking stuff was two fold

- complexity of building
- the build took too long and the readthedocs builder (so, not the GitHub Actions, as they weren't around at the time I think, and are actually irrelevant here because the docs build has to be run on the RTD builder) ran out of time; I am worried that whatever we do for the first point is going to be "lost" because of this second issue,  but maybe it's not a problem anymore

At the time (many years ago) there were plenty of questions along the lines of "can we not just upload the built documentation to the RTD site" but it wasn't supported. I have not looked recently to see if that has changed.

You can see the RTD builder page at https://readthedocs.org/projects/sherpa/builds/ which includes a failed build for this PR at https://readthedocs.org/projects/sherpa/builds/21074830/ but at least that's failing for an issue you should be able to see locally and not a RTD build issue (e.g. run-out-of-time).

hamogu:
No, I do need to test this by pushing - the error comes because RTD did not actually build the package because I had not added that to the `.readthedocs.ymal` file - I don't write these files all that often so I didn't notice what was missing...

My hunch is that building sherpa (without XSPEC) has gotten fast and standardized enough that it should be able to be built on RTD now (unlike 5 years ago), which would allow us to autogenerate at least some plots in the docs, so I'm testing that in this PR.

DougBurke:
I just wanted to provide some background and point out the pain points I was aware of for this project.

docs/conf.py
```diff
@@ -4,30 +4,21 @@
...
-# The minimum supported version of Sphinx is 1.8.
-#
-# The minimum requirements are:
-#    numpy  - since setup.py enforces this
-#    sphinx_rtd_theme
-#    sphinx_astropy
+# sherpa.utils._utils).
 #
 # The documentation can be built
 #   a) from the top level with 'python setup.py build_sphinx'
```
DougBurke, line 16: So, if I try

```bash
pip install .
python setup.py build_sphinx
```

I get a whole bunch of errors. I know that support for this is going away

- see #1714

and my PR for setup.py changes 

- #1608 

cleans up the documentation, but misses this line.

DougBurke: So we probably should just drop the "#   a) from the top level" line and change "#    b) from the docs/ .." to "#   .) from the docs/..."

docs/conf.py
```diff
@@ -58,63 +49,49 @@ def __getattr__(cls, name):
...
+# version in the build directory.
+# Thus, any C-extensions that are needed to build the documentation will *not*
+# be accessible, and the documentation will not build correctly.
+# See astropy/docs/conf.py for details where this explanation is taken from.
+
+# -- check that Sherpa is installed ---------------------------------------
+# Since this is a change from the previous way of building the docs
+# it warrants a check and a useful error message.
+try:
+    version = metadata.version('sherpa')
```
DougBurke, line 87:
```suggestion
    sherpa_release = metadata.version('sherpa')
```

We use `sherpa_release` below so may as well set it here rather than change the variable name.

DougBurke: Line 102 which says

```python
sherpa_release = sherpa._version.get_versions()['version']
```

can be deleted, but I can't seem to add a note to point this out (as it's unchanged code).

docs/conf.py
```diff
@@ -4,30 +4,21 @@
...
-#
-# The minimum requirements are:
-#    numpy  - since setup.py enforces this
-#    sphinx_rtd_theme
-#    sphinx_astropy
+# sherpa.utils._utils).
 #
 # The documentation can be built
 #   a) from the top level with 'python setup.py build_sphinx'
 #   b) from the docs/ directory with 'sphinx-build -b html . build/html'
```
DougBurke, line 17:
```suggestion
# To build the documentation:
#   cd docs/
# and then either of
#   make html
#   sphinx-build -b html . build/html
```

docs/install.rst
```diff
@@ -434,9 +434,14 @@ With these installed, the documentation can be built with the
 
     python setup.py build_sphinx
 
-This can be done **without** building Sherpa (either an installation
-or development version), since Mock objects are used to represent
-compiled and optional components.
+Note that this uses the installed version of sherpa, so if you want to make
```
DougBurke, line 437: I think this section needs a minor rework because

a) I can't seem to use `build_sphinx` any more
b) we know `build_sphinx` has been deprecated so may as well remove it
c) at the start of the `building the documentation` section we currently say

```
Building the documentation requires the Sherpa source code and several
additional packages:
```

and we need to update it to mention being able to build Sherpa - so smoething like

```
Building the documentation requires being able to build Sherpa
(:ref:`build-from-source`) and several additional packages:
```

[I haven't run this through sphinx so I'm not sure it will look]

docs/install.rst
```diff
@@ -434,9 +434,14 @@ With these installed, the documentation can be built with the
...
+sure the current repository version is used, you will need to install it with e.g.::
+
+    pip install -e .
+
+before changing to the docs directory. Only very specific modules are mocked out
+because they are hard to build and are not needed for the documentation build
+(currently ds9 and XSPEC).
 
 The documentation should be placed in ``build/sphinx/html/index.html``,
```
DougBurke, line 446: This path is different when you `cd docs; make html`. I just forget what it is now.

docs/install.rst
DougBurke: I note that this is also addressed in #1608 although that version still talks about mock objects, so there will be a small clash once one of these is merged.

docs/install.rst
DougBurke: I think it's `_build/sphinx/html/index.html` within `docs`, but can't guarantee it.

DougBurke (approved):
Thanks. Another gnarly build infrastructure item I didn't have to work on!

PR #1811, opened 2023-06-22
ui: use the default pickle version when dumping the state

### LINKED ISSUE #1243: We should bump our default pickle version from 2 to 4
We currently use in `sherpa.ui.utils.save`

```python
            pickle.dump(self, fout, 2)  # Use newer binary protocol
```

but the default version is now 4 (added in Python 3.4).

I think we should make `version` be an optional argument to `save`, have it default to 4, and then use that in the `dump` call. We don't need the equivalent in `load` since it is determined from the file contents.

Alternatively we could just drop the version field to `dump` and make it just use the default value (which could have long-term issues for users switching between Python/Sherpa versions over time).

DougBurke:
# Summary

Use the default version when pickling data (used by the UI `save` call) rather than an old version. Fix #1243

# Details

Rather than explicitly pick a version (which is now very old) use the default version (the old version was 2 and the current default version is 4). With changes to CIAO (package builds) it's not obvious how well the state file can move between versions anyway.

I have been wondering whether we should include some "state" files in out test data area which can be used to check we can still load them (e.g. updated when a new version is released) but maybe this is better done by SDS testing (as there's a lot of different things to test).

### DIFF
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -1027,7 +1027,8 @@ class Session(NoNewAttributesAfterInit):
 
         fout = open(filename, 'wb')
         try:
-            pickle.dump(self, fout, 2)  # Use newer binary protocol
+            # Use the default version rather than fix a version
+            pickle.dump(self, fout)
         finally:
             fout.close()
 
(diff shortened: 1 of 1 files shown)


DougBurke:
This was an easy fix to do before I leave, and if we never get around to small changes like this when will we ever improve things!

hamogu (approved):
Yes, In understand the choice was made to keep things consistent, but I agree to move on. Diverging too far from the Python default will cause us extra work for little gain, so this is a good change.

PR #1812, opened 2023-06-23
Autoupdate copyright year in sphinx docs

hamogu:
# Summary
Autoupdate copyright year in sphinx docs

## details
The docs are only build if something has been updated in the repro and that means the copyright year should be updated to the current year.
Using `datetime` we can never again forget to update that.

This trick is taken from astropy.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -289,8 +289,8 @@ main_doc = 'index'
 
 # General information about the project.
 project = 'Sherpa'
-copyright = '2019-2022, Chandra X-ray Center, Smithsonian Astrophysical Observatory.'
 author = 'Chandra X-ray Center, Smithsonian Astrophysical Observatory'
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -289,8 +289,8 @@ def getmtime(infile):
 
 # General information about the project.
 project = 'Sherpa'
-copyright = '2019-2022, Chandra X-ray Center, Smithsonian Astrophysical Observatory.'
 author = 'Chandra X-ray Center, Smithsonian Astrophysical Observatory'
+copyright = f"2019–{datetime.datetime.utcnow().year}, " + author
```
DougBurke, line 293: Yay. It also fixes the "multiple full stops" we get from sphinx...

DougBurke (approved):

PR #1816, opened 2023-07-04
Limit precision in the comparison of string outputs

### LINKED ISSUE #1815: New test failures du to numeircal precision
Some tests added in #1770 fail on my Mac (M1 ARM chip, compiling sherpa natively) with current Sherpa main on Python 3.1o with numpy 1.21.4. Test failures persists after upgrading numpy to the current version 1.25. However, a look at the errors clearly shows that the problem is just numerics and not a real sherpa problem, the three failures are
```Python
def test_fit_results_str():
        """Just check we can call str() on a fit results instance"""
    
        data = make_data(Data1D)
        model = Const1D()
        fit = Fit(data, model, stat=Cash())
        fres = fit.fit()
        assert fres.succeeded
    
>       check_str(fres,
                  ["datasets       = None",
                   "itermethodname = none",
                   "methodname     = levmar",
                   "statname       = cash",
                   "succeeded      = True",
                   "parnames       = ('const1d.c0',)",
                   "parvals        = (3.33332926058462,)",
                   "statval        = -4.079456086503793",
                   "istatval       = 6.0",
                   "dstatval       = 10.079456086503793",
                   "numpoints      = 3",
                   "dof            = 2",
                   "qval           = None",
                   "rstat          = None",
                   "message        = successful termination",
                   "nfev           = 9"])

sherpa/tests/test_fit_unit.py:2850: 

out = <Fit results instance>
expecteds = ['datasets       = None', 'itermethodname = none', 'methodname     = levmar', 'statname       = cash', 'succeeded      = True', "parnames       = ('const1d.c0',)", ...]

    def check_str(out, expecteds):
        """Check that out, when split on a newline, matches expecteds"""
    
        toks = str(out).split("\n")
        for tok, expected in zip(toks, expecteds):
>           assert tok == expected
E           AssertionError: assert 'parvals     ...32926058504,)' == 'parvals     ...32926058462,)'
E             - parvals        = (3.33332926058462,)
E             ?                                 --
E             + parvals        = (3.33332926058504,)
E             ?                                ++

sherpa/tests/test_fit_unit.py:2836: AssertionError
```
and

```Python
./sherpa/tests/test_fit_unit.py::test_fit_results_with_iteration_str Failed: [undefined]AssertionError: assert 'parvals     ...00000003157,)' == 'parvals     ...00000000316,)'
  - parvals        = (3.250000000000316,)
  ?                                   ^
  + parvals        = (3.2500000000003157,)
  ?                                   ^^
def test_fit_results_with_iteration_str():
        """Just check we can call str() on a fit results instance with an iteration method
        """
    
        data = Data1D("x", [1, 2, 3, 4, 5], [2, 4, 12, 3, 4])
        data.staterror = [0.8] * 5
        model = Const1D()
        iter_opts = {'name': 'sigmarej', 'maxiters': 5, 'hrej': 5, 'lrej': 5, 'grow': 0}
        fit = Fit(data, model, stat=Chi2(), itermethod_opts=iter_opts)
        fres = fit.fit()
        assert fres.succeeded
    
>       check_str(fres,
                  ["datasets       = None",
                   "itermethodname = sigmarej",
                   "methodname     = levmar",
                   "statname       = chi2",
                   "succeeded      = True",
                   "parnames       = ('const1d.c0',)",
                   "parvals        = (3.250000000000316,)",
                   "statval        = 4.296875",
                   "istatval       = 225.0",
                   "dstatval       = 220.703125",
                   "numpoints      = 4",
                   "dof            = 3",
                   "qval           = 0.23114006377865534",
                   "rstat          = 1.4322916666666667",
                   "message        = successful termination",
                   "nfev           = 8"])

sherpa/tests/test_fit_unit.py:2907: 

out = <Fit results instance>
expecteds = ['datasets       = None', 'itermethodname = sigmarej', 'methodname     = levmar', 'statname       = chi2', 'succeeded      = True', "parnames       = ('const1d.c0',)", ...]

    def check_str(out, expecteds):
        """Check that out, when split on a newline, matches expecteds"""
    
        toks = str(out).split("\n")
        for tok, expected in zip(toks, expecteds):
>           assert tok == expected
E           AssertionError: assert 'parvals     ...00000003157,)' == 'parvals     ...00000000316,)'
E             - parvals        = (3.250000000000316,)
E             ?                                   ^
E             + parvals        = (3.2500000000003157,)
E             ?                                   ^^

sherpa/tests/test_fit_unit.py:2836: AssertionError
```
and

```Python
./sherpa/tests/test_fit_unit.py::test_esterrorresults_str Failed: [undefined]AssertionError: assert 'parvals     ...32926058504,)' == 'parvals     ...32926058462,)'
  - parvals     = (3.33332926058462,)
  ?                              --
  + parvals     = (3.33332926058504,)
  ?                             ++
def test_esterrorresults_str():
        """Basic check of str call."""
    
        data = make_data(Data1D)
        mdl = Const1D("mx")
        fit = Fit(data, mdl, stat=Cash())
        fit.fit()
    
        errs = fit.est_errors()
>       check_str(errs,
                  ["datasets    = None",
                   "methodname  = covariance",
                   "iterfitname = none",
                   "fitname     = levmar",
                   "statname    = cash",
                   "sigma       = 1",
                   "percent     = 68.26894921370858",
                   "parnames    = ('mx.c0',)",
                   "parvals     = (3.33332926058462,)",
                   "parmins     = (-1.0524866849982824,)",
                   "parmaxes    = (1.0524866849982824,)",
                   "nfits       = 0"])

sherpa/tests/test_fit_unit.py:3037: 

out = <covariance results instance>
expecteds = ['datasets    = None', 'methodname  = covariance', 'iterfitname = none', 'fitname     = levmar', 'statname    = cash', 'sigma       = 1', ...]

    def check_str(out, expecteds):
        """Check that out, when split on a newline, matches expecteds"""
    
        toks = str(out).split("\n")
        for tok, expected in zip(toks, expecteds):
>           assert tok == expected
E           AssertionError: assert 'parvals     ...32926058504,)' == 'parvals     ...32926058462,)'
E             - parvals     = (3.33332926058462,)
E             ?                              --
E             + parvals     = (3.33332926058504,)
E             ?                             ++

sherpa/tests/test_fit_unit.py:2836: AssertionError
```

hamogu:
As described in #1815, running the three tests on a Mac with ARM even with recent numpy versions differs numerically, which makes the string comparison fail.

One way to address that would be split the string and compare the formatting of the output string and the numbers separately, but that's more effort. Just limiting the comparison to the first 6-8 digits is enough to ensure that (a) the test do not fail just due to numerics, but would still detect any meaningful change in the output numbers.

fixes #1815

### DIFF
--- a/sherpa/tests/test_fit_unit.py
+++ b/sherpa/tests/test_fit_unit.py
@@ -2833,7 +2833,7 @@ def check_str(out, expecteds):
 
     toks = str(out).split("\n")
     for tok, expected in zip(toks, expecteds):
-        assert tok == expected
+        assert tok[:28] == expected[:28]
 
     assert len(toks) == len(expecteds)
 


DougBurke:
While the suggested fix will work, I worry about it being too brittle, in that if we changed the output so that there's more text then the tests will still pass (once the text is adjusted) but we've "lost" power in the number of decimal places we can check. The simple answer is to just add a comment to `check_str` to point this out. However, I've added this function (or something like it) in several test files so I wonder if we want to

- make it a fixture in `conftest.py` so it's just written once
- expand the checks so we can use a regexp so users can say when to check (it would be nice if we could replicate the `pytest-doctestplus` logic to be able to check a line matches)

So something like the following untested code

```
import re

def check_str(out: str, expecteds: list[Union[str,re.Patern]]]) -> None:
    """Check that out, when split on a newline, matches expecteds

    Parameters
    ----------
    out : str
        The output to check
    expecteds : list of str or Pattern
        The expected output ,slpit on new-line. It is either an exact match
        (str) or a regexp pattern (e.g. to avoid numerical precision issues)

    """

    toks = str(out).split("\n")
    for tok, expected in zip(toks, expecteds):
        try:
            assert expected.match(tok), (expected.pattern, tok)
        except AttributeError:
            assert tok == expected

    assert len(toks) == len(expecteds)
```

DougBurke:
One problem with the current suggestion is that `check_str` is also used in the same file for output like

```
    check_str(out,
              ["# nfev statistic mx.c0",
               "0.000000e+00 6.000000e+00 1.000000e+00",
               "1.000000e+00 6.000000e+00 1.000000e+00",
               "2.000000e+00 5.995167e+00 1.000345e+00",
...
```

where the subscription is not meaningful.

The downside to my approach is that you need to change each number to be a regexp, which is why it would be nice to just use the logic that `pytest-doctestplus` uses when checking with the `# doctest: +FLOAT_CMP` logic

DougBurke:
I've pushed a commit to show what I mean. It uses pytest-doctestplus to do the text comparison (although you can explicitly add a compiled re.Pattern to be explicit).

I expect this change is not sufficient - we will still need to change some of the expected text to handle the mac differences but I can't test this part.

DougBurke:
My changes should be squashed (as it adds a feature and then removes it) but I've left as is for now to avoid confusing the issue. I should have created these in a separate PR to avoid this confusion, but I didn't expect it to get quite so out-of-hand.

DougBurke:
I've added the "CI: arch" label since this runs the aarch64 tests which seem to also show the same issue that @hamogu is having on macOS.

DougBurke:
@hamogu - so #1829 is significantly more complex than this PR, but I prefer it (it tries to behave like `pytest-doctestplus` and, if we ever make it a required package for testing, could simplify the code).

DougBurke:
Relpaced by #1829

PR #1819, opened 2023-07-17
CI: hide NumPy 1.25 array ndim>0 deprecation warnings

DougBurke:
# Summary

Allow the tests to run with NumPy 1.25 and with Crates (from CIAO 4.15) as the I/O backend. There are no functional changes.

# Details

NumPy 1.25 now creates deprecation warnings for converting a non-scalar NumPy value (even if only has a size of 1) to a normal Python value. The Sherpa code base was updated in #1807 but when run with crates (as done in CIAO regression tests) we still see the reports. As we can't fix this in crates (and actually finding where it is coming from has proved to be elusive) we can just hide the warnings using our "standard" set of checks in `sherpa/conftest.py`. The hope is we can revert this change by the end of the year (as it should have been addressed in crates by then).

The first commit is unrelated, and odesn't actually change any tests or code. It just hides the warnings you see if you run the tests without `pytest-doctestplus` installed (as this was annoying me when working on this PR).

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,6 +2,10 @@
 addopts = -rs --ignore=setup.py --ignore=test_requirements.txt
 norecursedirs = .git build dist tmp* .eggs
 text_file_format = rst
+filterwarnings =
+    # Since we can runpytest without loading pytest-doctestplus, hide
+    # the warnings we get when this is done
+    ignore::pytest.PytestConfigWarning
 doctest_plus = enabled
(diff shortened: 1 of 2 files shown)


DougBurke:
This should let the CIAO tests pass. I have tested it with an old-ish CIAOX build (it only reported problems with `test_grouped_ciao4_5` and not `test_proj` so some tweaks may still be needed, but I'd be surprised).

PR #1823, opened 2023-07-21
sherpa_test: tweak how pytest/plugins are installed

DougBurke:
# Summary

Internal changes to how we check for test requirements used by the sherpa_test script.

# Details

We had been doing

    try:
        importlib.import_module("pytest!=1.0,!=5.2.3")
    except ImportError:
        go-and-install-pytest()

which ends up always trying to install pytest (which would be a no-op when already installed). So, this code makes the installation a bit-more verbose, by separating out the package name from the installation-constraint.

It was found when trying to add pytest-doctestplus as a requirement. In doing so I was seeing some bizarre behavior, which made me add the "check" option (another option for specifying a dependency). It isn't actually used here (since this PR does not need any plugin), but I have left the code in.

I added typing rules for the code changed (but did not want to add to more of this module as it turns out it would require some code tweaks).

### DIFF
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -19,7 +19,6 @@
 #
 
 """
-
 Modeling and fitting package for scientific data analysis
 
 Sherpa is a modeling and fitting package for scientific data analysis.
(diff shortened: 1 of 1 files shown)


hamogu:
Before I look at the code in detail (and I admit that's going to be fast, given how few lines are changed), I wonder exactly what use case we target with this code. If we want all users to be always able to run all tests, then why do we not simply add pytest (plus and plugins we select) as mandatory sherpa requirements? I think all of those are easy to install because most (all) of pytest is pure-python. Then, we can rely on them to be installed, and, at the very most, catch errors and print "you did not install sherpa correctly. package XXX should be there".

If we want that to be optional, we can declare them as [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) and, if missing, tell users to run `pip install sherpa[test]`. [*]

In my opinion, automatically downloading stuff from the internet and installing it without explicit confirmation from the user in a routine where one might not expect that (i.e. a routine that runs tests) is bad practice and I can imaging situation where that becomes brittle, such as off-network installations, or cases where outgoing https calls need to be routed via tunnels or VPNs for security.

However, that's somewhat beyond the scope of this PR. I'm also OK with fixing what's broken now (since that's just a tweak to the status quo) and discuss any further changes in a new issue.

*: The reason I can paste this link so quickly is that I had all the tabs open, because I thought about writing a PR to put all the optional testing requirements into `pyproject.toml` or `setup.cfg`explicitly to allow for `pip install sherpa[test]` or `pip install sherpa[docs]` as many other packages do it.

DougBurke:
This code is used by the sherpa_test script/command, so it's a way for sherpa users to run the tests without having to bother with pytest/however we want to set up sherpa testing.

I do have  along-outstanding PR to add "[test]" to pip, but that's a separate issue (I think this code is in #1608 but it's the weekend so I'm not going to chase it down).

hamogu (approved):

sherpa/__init__.py
```diff
@@ -867,51 +866,96 @@ def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=F
...
+
+            You need pytest in order to run the tests.
             If you downloaded the source code, please run 'pip install -r test_requirements.txt'
             from the source directory first.
             """)
             raise
 
-    deps = ['pytest>=5.0,!=5.2.3']  # List of packages to be installed
-    pytest_plugins = []  # List of pytest plugins to be installed
+    # packages are stored as a dictinary with keys
```
wmclaugh, line 897:
```suggestion
    # packages are stored as a dictionary with keys
```

sherpa/__init__.py
```diff
@@ -867,51 +866,96 @@ def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=F
...
+    #   name:  package name, as used by pip
+    #   check: module to check it is installed
+    #          (defaults to name if not given)
+    #   constraint: the constaint to use with pip
```
wmclaugh, line 901:
```suggestion
    #   constraint: the constraint to use with pip
```

sherpa/__init__.py
```diff
@@ -867,51 +866,96 @@ def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=F
...
+    #
+    # So we can have
+    #
+    #  {'name': 'pycheck'}
+    #  {'name': 'pycheck', 'constraint': 'pytest>=5.0,!=5.2.3'}
+    #  {'name': 'pytest-doctestplus',
+    #   'check': 'pytest_doctestplus.output_checker'}
+    #
+    # Note that for pytest plugins, the module name is assumed to be
+    # the name field with hypens replaced by underscores. If necessary
```
wmclaugh, line 912:
```suggestion
    # the name field with hyphens replaced by underscores. If necessary
```

sherpa/__init__.py
```diff
@@ -867,51 +866,96 @@ def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=F
...
+    # this could be updated to be another field in the package
+    # dictonary.
```
wmclaugh, line 914:
```suggestion
    # dictionary.
```

PR #1824, opened 2023-07-21
remove non-existent datastack plotting backend initialization attribute

nplee:
Probably left behind by accident, but the `sherpa.plot.backend.initialize_backend` attribute no longer exists for the datastack plotting routines.  Its presence is causing Sherpa to throw an error:

```
sherpa> datastack.plot_data([])
AttributeError: module 'sherpa.plot.pylab_backend' has no attribute 'initialize_backend'
```
but the removal of the line will allow a figure to be plotted for each dataset in the the stack.

### DIFF
--- a/sherpa/astro/datastack/utils.py
+++ b/sherpa/astro/datastack/utils.py
@@ -235,7 +235,7 @@ def plot_wrapper(func):
         kwargs
            The keyword arguments to the function.
         """
-        shplot.backend.initialize_backend()
+
         for dataset in self.filter_datasets():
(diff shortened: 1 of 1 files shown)


DougBurke:
@nplee - I am reluctant to accept this just yet because I know that @hamogu has worked on this in an existing PR (#1382). We'll keep this open to remind us to check again soon.

DougBurke:
@hamogu - now that #1382 is merged should we take this ? I know you have other datastack changes planned, but this is a simple one.

PR #1780, opened 2023-05-23
Fix model plot display when factor is set for grouped PHA data (issue #1779)

DougBurke:
# Summary

Ensure that model plots can be created for PHA data when set_analysis routine is used to set factor  > 0 and the data is grouped. Fixes #1779 and #1784

# Details

The factor setting comes in to the very-end of the internal "fix units" routine for DataPHA objects, and obviously missed-out on the recent work to improve filtering behavior (eg #1477) as we didn't have a test that triggered the problem. So a test has been added to try and cover the problem (at a  lower level than found in #1779, which used the UI layer). Note that we want to check

- grouped and ungrouped data
- filtered and unfiltered data
- plot_model and  plot_fit (the model part) behavior, which is handled by the two `sherpa.astro.plot` model classes

In working on this I came across #1784 so i have addressed this too. As this is all for "approximate" data it may not matter exactly how we calculate the mid-point of each group, but I prefer the newer version.

We should review the DataPHA class for data access to channel/group values (e.g. `get_x` vs `_get_ebins` vs ...) but that is a significant rework but I'm still birnt out by #1477 and related work, so I do not plan on this for quite some time.

I think this should be looked at and then #1828

PR #1828, opened 2023-07-24
Improve PHA source plots: support type=counts and factor above 1

### LINKED ISSUE #1825: plot_source incorrectly scaled with analysis factor >0
`plot_source` does not scaled properly Y values when `set_analysis` factor option is >0. 
The plot shows the source model plot with factors 0,1,2 and the curves are shifted in normalization, but the shape is the same.
E*f(E) should generate curves with varying shapes.

plt_source_factors

DougBurke:
# Summary

Source plots for PHA data now follow the rate and factor analysis settings. Prior to this the rate setting was ignored and the factor setting produced incorrect values for factor=1 and factor=2). Fix #1825 

# Details

This was written separately to

- #1780 

which we had merged first. Although they are related, the source changes (needed here) are a lot simpler than needed for the model case (which needs to worry about grouping data) so the code paths are different (even though they are doing similar things). It might be nice to refactor this but I do no believe it is worth it (as the refactoring would probably be fairly extensive).

So, for factor=0, we have (assuming type=rate):

    y_0 = mdl(lo, hi) / (hi - lo)

where the model is assumed to create values of photon/cm^2/s for each bin. This means that y is in units of photon/cm^2/s/<X units> where <X units> is energy or angstrom. 

For factor = 1 and we calculate

    y_1 = y_0 * (lo + hi) / 2
    y_2 = y_1 * (lo + hi) / 2

That is, we multiply the factor_0 value by the mid-point of each X bin (and the number of times we do this depends on the factor setting).

So, for factor=0 we plot (lo, hi, y_0), factor=1 we use (lo, hi, y_1), and for factor=2 we use (lo, hi, y_2). This makes sense frm a units viewpoint, but is it what we want? For instance, for factor=1 we could just plot

    mdl(lo, hi)

instead of

    mdl(lo, hi) * (lo + hi)
    --------------------------
          2 * (hi - li)
 
Since `mdl(lo, hi)` calculates a rate, then we need to multiply by the exposure time when the user has called `set_analysis(..., type="counts")`.

This code replicates code that is in the `DataPHA` object - in partiicular the `_fix_y_units` code, but infortunately we can't just use that (to avoid code duplication) since they aren't quite doing the same thing (e.g. the source code doesn't group the data whereas the `_fix_y_units` code does, and the model being evaluated doesn't have a response whereas it does for the DataPHA case. So I think we have to live with this "almost repeated code".

### DIFF
--- a/sherpa/astro/tests/test_astro_plot.py
+++ b/sherpa/astro/tests/test_astro_plot.py
@@ -28,7 +28,7 @@ from sherpa.utils.testing import requires_data, requires_fits
 
 from sherpa.astro.data import DataARF, DataPHA
 from sherpa.astro.instrument import create_delta_rmf
-from sherpa.astro.plot import SourcePlot, \
+from sherpa.astro.plot import SourcePlot, ComponentSourcePlot, \
     DataPHAPlot, ModelPHAHistogram, ModelHistogram, OrderPlot, \
     EnergyFluxHistogram, PhotonFluxHistogram,  _check_hist_bins
 from sherpa.astro import plot as aplot
@@ -202,6 +202,35 @@ def test_sourceplot(caplog, make_basic_datapha):
     check_sourceplot_energy(sp)
 
 
+def test_sourceplot_filtered(caplog, make_basic_datapha):
+    """Filtering only changes the mask attribute"""
+    
+    data = make_basic_datapha
+    data.units = "energy"
+
+    m1 = Const1D('bgnd')
+    m2 = Gauss1D('abs1')
+    src = 100 * m1 * (1 - m2) * 10000
+
+    m1.c0 = 0.01
+    m2.pos = 5.0
+    m2.fwhm = 4.0
+    m2.ampl = 0.1
+
+    sp = SourcePlot()
+    with caplog.at_level(logging.INFO, logger='sherpa'):
+        sp.prepare(data, src, lo=0.35, hi=0.87)
+
+    # The filtering should probably be this, but let's test the
+    # current behavior:
+    #
+    # expected = [False] * 2 + [True] * 6 + [False] * 2
+    expected = [False] * 3 + [True] * 5 + [False] * 2
+    assert sp.mask == pytest.approx(expected)
+    assert len(caplog.records) == 0
+    check_sourceplot_energy(sp)
+
+
 def test_sourceplot_counts(caplog, make_basic_datapha):
     """test_sourceplot but when rate=False is chosen"""
 
@@ -311,6 +340,34 @@ def test_sourceplot_wavelength(caplog, make_basic_datapha):
     check_sourceplot_wavelength(sp)
 
 
+def test_sourceplot_wavelength_filtered(caplog, make_basic_datapha):
+    """Filtering only changes the mask attribute"""
+    
+    data = make_basic_datapha
+    data.units = "wave"
+
+    m1 = Const1D('bgnd')
+    m2 = Gauss1D('abs1')
+    src = 100 * m1 * (1 - m2) * 10000
+
+    m1.c0 = 0.01
+    m2.pos = 5.0
+    m2.fwhm = 4.0
+    m2.ampl = 0.1
+
+    sp = SourcePlot()
+    with caplog.at_level(logging.INFO, logger='sherpa'):
+        sp.prepare(data, src, lo=13, hi=40)
+
+    # Given the filtering for energy didn't quite match, DJB is
+    # slightly surprised this works.
+    #
+    expected = [False] * 2 + [True] * 6 + [False] * 2
+    assert sp.mask == pytest.approx(expected)
+    assert len(caplog.records) == 0
+    check_sourceplot_wavelength(sp)
+
+
 @pytest.mark.parametrize("factor", [1, 2])
 def test_sourceplot_wavelength_facn(factor, caplog, make_basic_datapha):
     """Change plot factor for test_sourceplot_wavelength"""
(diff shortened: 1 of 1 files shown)


DougBurke:
@anetasie - I think this shoulf "fix" your issue #1825, but it needs some careful testing (e.g. to look at the x/y values you get and whether they make sense)

anetasie:
@DougBurke Thanks. This fixes the problems for set_analysis(1,'energy','rate',factor=???) and it looks good.
I'm wondering about 'counts' option, which generates the plots with really high values and as you said the Y value needs to be corrected by the exposure.

anetasie:
I'm also wondering about the translation to wavelength and rate. Need to get some grating spectra with wavelength scales for testing.

DougBurke:
So there's

    set_analysis(idval, "energy" | "wave", "rate", factor=0, 1, 2, ...)

which is the main one. I also added support for

    set_analysis(idval, "energy" | "wave", "counts", factor=0, 1, 2, ...)

but only because we can - it isn't particularly useful that I can see (it's not really counts, but I think that is more a problem with this argument, it should have just said "do I want a rate or not"). I would leave it in, but if we feel it should be removed it's not hard to do so.

anetasie:
@DougBurke which powerlaw model are you using?

anetasie:
I guess my question is related to the definition of the photon index gamma. For factor 2 in powlaw1d I expect the function rising up. This is what I tested and it looked ok

anetasie:
OK the plots look good. I just got confused with gamma vs. factors.

anetasie (approved):

DougBurke:
@hamogu - I've asked for your review just to check the code isn't too crazy, and because you have been thinking about the plotting code recently too.

DougBurke:
I've rebased to deal with a minor conflict from merging #1780 which gave me the chance to combine two commits and fix at least one typo in a comment.

DougBurke:
Here's the script I used for the plots above:

```

from sherpa.astro.ui import *

load_pha("sherpa-test-data/sherpatest/3c273.pi")
notice(0.5, 6)

set_source(xsphabs.gal * powlaw1d.pl)

pl.ampl = 2e-5

def make_plot(units, factor):

    set_analysis(1, units, factor=factor)

    pl.gamma = 2
    plot_source_component(pl, xlog=True, ylog=True)
    ax = plt.gca()
    ax.lines[0].set_label("gamma = 2")

    pl.gamma = 1
    plot_source_component(pl, overplot=True)
    ax.lines[-2].set_label("gamma = 1")

    pl.gamma = 0
    plot_source_component(pl, overplot=True)
    ax.lines[-2].set_label("gamma = 0")

    pl.gamma = -1
    plot_source_component(pl, overplot=True)
    ax.lines[-2].set_label("gamma = -1")

    pl.gamma = -2
    plot_source_component(pl, overplot=True)
    ax.lines[-2].set_label("gamma = -2")

    plt.legend()
    plt.title(f"units={units}  factor={factor}")

    out = f"plot-{units}-{factor}.png"
    plt.savefig(out)
    print(f"Created: {out}")


for unit in ["energy", "wave"]:
    for factor in [0, 1, 2]:
        make_plot(unit, factor)
```

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -48,20 +48,28 @@ def check_sourceplot_energy(sp, factor=0):
     assert sp.xlabel == 'Energy (keV)'
 
     # the following depends on the backend
-    # assert sp.ylabel == 'f(E)  Photons/sec/cm$^2$/keV'
     if factor == 0:
-        assert sp.ylabel.startswith('f(E)  Photons/sec/cm')
-        assert sp.ylabel.endswith('/keV ')
+        if rate:
+            assert sp.ylabel.startswith('f(E)  Photons/sec/cm')
+        else:
```
hamogu, line 54: Line 42 sets "rate=True". Was this intended to be parameterized for rate=True/False?

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -95,18 +108,27 @@ def check_sourceplot_wavelength(sp, factor=0):
     assert sp.xlabel == 'Wavelength (Angstrom)'
 
     if factor == 0:
-        assert sp.ylabel.startswith('f(lambda)  Photons/sec/cm')
-        assert sp.ylabel.endswith('/Angstrom ')
+        if rate:
```
hamogu, line 111: Same. rate is always True in this test.

hamogu (changes requested):

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -195,7 +254,7 @@ def test_sourceplot_counts(caplog, make_basic_datapha):
         sp.prepare(data, src)
 
     assert len(caplog.records) == 0
-    check_sourceplot_energy(sp)
+    check_sourceplot_energy(sp, rate=False)
```
DougBurke, line 257: @hamogu - here is a case when `check_sourceplot_energy` is called with `rate=False`.

sherpa/astro/tests/test_astro_plot.py
```diff
@@ -330,7 +417,67 @@ def test_sourceplot_wavelength_counts(caplog, make_basic_datapha):
         sp.prepare(data, src)
 
     assert len(caplog.records) == 0
-    check_sourceplot_wavelength(sp)
+    check_sourceplot_wavelength(sp, rate=False)
```
DougBurke, line 420: @hamogu and ditto for `check_sourceplot_wavelength`.

sherpa/astro/tests/test_astro_plot.py
hamogu: Then why does this say"if rate"?

hamogu:
I'm sorry,  I missed that. On Aug 22, 2023 10:37 AM, Doug Burke ***@***.***> wrote:
@DougBurke commented on this pull request.



In sherpa/astro/tests/test_astro_plot.py:
> @@ -95,18 +108,27 @@ def check_sourceplot_wavelength(sp, factor=0):
     assert sp.xlabel == 'Wavelength (Angstrom)'
 
     if factor == 0:
-        assert sp.ylabel.startswith('f(lambda)  Photons/sec/cm')
-        assert sp.ylabel.endswith('/Angstrom ')
+        if rate:

I don't understand.
This is not a test itself - it's name doesn't begin with test_ - but is used in tests. It defaults to rate=True but is also called with rate=False - e.g. it is called below at line 420 / see https://github.com/sherpa/sherpa/pull/1828/files#r1300537291
This is also true of the other case you are asking about.

hamogu (approved):

hamogu:
I'm sorry I missed the true/ false thing.  I was reviewing changes on my phone.

DougBurke:
... the perils of PR reviewing on a beach ....

Issue #1834, opened 2023-07-31
XSPEC 12.13.1 has been released


DougBurke:
There are no new models. There are a few model parameter changes:

```
% ./scripts/check_xspec_update.py  ~/local/heasoft-6.32/spectral/manager/model.dat 
Model apec
  [ 1] par 3 name: redshift -> Redshift

Model bapec
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model btapec
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model bexrav
  [ 1] par 9 name: redshift -> Redshift

Model bexriv
  [ 1] par 6 name: redshift -> Redshift

Model brnei
  [ 1] par 5 name: redshift -> Redshift
  [ 2] par Velocity max: 1000000.0 -> 10000.0
  [ 3] par Velocity softmax: 1000000.0 -> 10000.0
  [ 4] par Velocity hardmax: 1000000.0 -> 10000.0

Model bvapec
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model bvrnei
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model bvtapec
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model bvvapec
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
  [ 3] par Velocity hardmax: 1000000.0 -> 10000.0

Model bvvrnei
  [ 1] par Velocity max: 1000000.0 -> 10000.0
  [ 2] par Velocity softmax: 1000000.0 -> 10000.0
[...]

PR #1835, opened 2023-07-31
Note compatability with XSPEC 12.13.1

DougBurke:
# Summary

Note that Sherpa can be built against XSPEC 12.13.1 and update some parameter ranges or defaults to match the new values from this release.

# Details

There are no new models in XSPEC 12.13.1 and no obvious changes to the build (at least for non-macOS platforms) so it is as easy as just updating the documentation. However, there are some parameter-range changes (e.g. see #1834) we can update these (including the `cemekl`/`cevmkl` `Tmax` minimum parameter needing to bump up from 0.01 to 

```
>>> 10**5.5 * 8.6171e-8
0.02724966282543694
```

which we reported to HEASARC thanks to a user seeing crashes in Sherpa).

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -335,9 +335,10 @@ Update the XSPEC bindings?
 --------------------------
 
 The :py:mod:`sherpa.astro.xspec` module currently supports
-:term:`XSPEC` versions 12.13.0, 12.12.1, and 12.12.0. It may build against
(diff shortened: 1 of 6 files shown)


Marie-Terrell (approved):
Looks good to me. I have only one test failure without these changes. The test_xspec_unit.py fix resolves this (tested on Linux RHEL8 and macOS Intel).

PR #1832, opened 2023-07-31
astropy: rework RMF input (fix #1344)

DougBurke:
# Summary

Allow the AstroPy backend to read in HRC-I RMF data. Fixes #1344.

# Details

Rework the RMF input code so that instead of flattening columns containing variable-length rows immediately, we instead do this as part of the final "flattening" process. It was hard for me to understand the old code so I felt this was a cleaner approach, and hopefully we have enough tests to check the various code paths.

The old code allowed some columns - in particular MATRIX - to be optional. We now more-closely follow the OGIP standard and make the columns required. There are more changes that could be made to the AstroPy I/O code - e.g. see #1091 - but I did not want to go that far here.

Surprisingly enough this doesn't seem to conflict with #1841 (at least initially, I have not checked recently).

PR #1836, opened 2023-08-01
Allow ARF and RMF to be written out

### LINKED ISSUE #203: FITS metadata can be wrong in FITS files created with the astropy backend (e.g. save_data)
# Overview

The Crates backend removes certain "structural" FITS keywords from the header block stored with data objects. This means that it's "safe" to copy over the remaining values to FITS output files. This filtering does not happen with the astropy backend, which can result in unwanted keywords being passed through to the output file. This can cause confusion in downstream software.
# Details

This was noticed in testing out the #102 fix in the 4.8.1 release candidate, but is separate from that code. 

The test case is basically

```
ui.load_data('bob.fits')
ui.save_data('/tmp/src.fits', ascii=False, clobber=True)
```

which I have run, for the same input file, for Sherpa in CIAO 4.8 (so using the crates backend) and using astropy 1.1.2 with the 4.8.1 release branch.
- CIAO 4.8 output

```
% dmlist /tmp/src2.fits blocks

Dataset: /tmp/src2.fits

Block    2: SPECTRUM                       Table         2 cols x 1024     rows
% dmlist /tmp/src2.fits cols

Columns for Table Block SPECTRUM

ColNo  Name                 Unit        Type             Range
   1   CHANNEL                           Real8          -Inf:+Inf            
   2   COUNTS                            Real8          -Inf:+Inf            
```
- 4.8.1 branch + AstroPy 1.1.2 output

```
% dmlist /tmp/src.fits blocks

Dataset: /tmp/src.fits

Block    2: SPECTRUM1                      Table         2 cols x 1024     rows
% dmlist /tmp/src.fits cols

Columns for Table Block SPECTRUM1

ColNo  Name                 Unit        Type             Range
   1   CHANNEL                           Real8          1.0:     1024.0      
   2   COUNTS                            Real8          1.0:     1024.0      
```

I've checked that the column data are the same in the two files; there are some header differences which are a little surprising to me but within expected bounds (I think).

The two points to note are:

1) The ranges of the columns are different: for CIAO they are undefined (not particularly useful, but it's a valid FITS file) whereas for the AstroPy output they are identical. This seems surprising, and is in fact wrong for the COUNTS column, as shown below.

2) The block name is different: `SPECTRUM` vs `SPECTRUM1`. Ideally they'd be the same, but I don't think that's too important.

Looking at the values for the two columns shows that the range for the COUNTS column is wrong, since the minimum value is 1 not 0:

```
% dmstat /tmp/src.fits  sig-
CHANNEL
    min:    1         @:    1 
    max:    1024          @:    1024 
   mean:    512.5 
    sum:    524800 
   good:    1024 
   null:    0 

COUNTS
    min:    0         @:    1 
    max:    3         @:    62 
   mean:    0.07421875 
    sum:    76 
   good:    1024 
   null:    0 

```

My guess is that the wrong range is being used when setting up the COUNTS column in the astropy backend, but I haven't looked at the code to confirm this.

### LINKED ISSUE #1695: create_delta_rmf improvement
I keep on writing (admittedly, mostly in tests)

```python
    rmf_src = create_delta_rmf(elo, ehi, e_min=elo, e_max=ehi)
```

The `e_min/e_max` arguments should probably default to this behavior (the default is `None` and that currently means "don't set them").

### LINKED ISSUE #1699: I really want a way to write out an ARF or RMF
We have a way to write out a PHA but not an ARF or RMF. There are times (not often) when I'd find it useful. It's linked to ideas like #1619, but is a separate item.

DougBurke:
# Summary

Support writing out ARF and RMF data to FITS files (and, for ARF, ASCII files). Fix #1699, #1695, and #203.

# Details

[**note** we have just had a helpdesk ticket where this functionality would have been useful, so it's not just me wanting it ;-]

This requires

- an "entry point" in sherpa.astro.io
- support in the backends (mainly to ensure that headers can be written to table files, which should have been present anyway but was not necessarily done), some of this work has now gone into #1841
  - actually, the RMF requires significantly-more work since we need to create two blocks, deal with setting the TLMIN4
    keyword in the MATRIX extension, and working out how to handle Variable-Length Fields for the F_CHAN, N_CHAN,
    and MATRIX columns in the same block 
- plumbing in the sherpa.astro.ui layer

The intent is to create inter-operable files - that is, ones that match the OGIP standards - but as there are many different flavors of RMF in the community there may be some subtle issues the existing tests do not catch (i.e. there may be further tweaks required). This is also partly because we don't currently guarantee full support for PHA files (there are some issues with RATE values and possibly elsewhere), and also because we don't have a good way of defining relevant information - e.g. is the RMF a RSP or "just" a RMF, do we have a source or background response, ... - that would help in setting up some of the metadata that the OGIP standard wants. This is all for future work, as the basic aim is to be able to write out ARF and RMF and then read them back in and get the same data. There are also existing issues over how well we can write out data (with correct data types) which is also "for future consideration".

One complication we have is what we do with FITS header keywords on input: we have previously not had a coherent strategy - we remove WCS keywords for images for both backends but other files and keywords are left alone (somewhat backend dependent,in part because crates removes some "structural" keywords for us anyway). We now are more pro-active, and remove a number of such keywords, such as `TLMIN1`, `PCOUNT`, `BLANK`, `BITPIX`, `NAXIS*`: this was needed because we could end up with invalid files because header values from the input file, which didn't make sense for the output file, were being written out. This is not perfect, and may need more downstream cleanup, but provides a sensible default to begin with. It does mean that the Sherpa "read" routines - such as load_pha, load_image, unpack_rmf, .... - are not guaranteed to provide access to all the data in the input file, but I think I am happy with saying that the user will have to read in any such file with their preferred I/O library if they need this information. The alternative is to only strip out these keywords on output, but that may make it hard to add them if needed [although this would not work for the crates backend].  **There** are some things to think about: I want to support XSPEC models which use XFLT keywords, and Moritz wants to support polarmetric models/other situations where having access to the headers is needed, but I claim that they are not going to be specified via these "structural" keywords but domain-specific keys.

I added in, at the end, a fix for #1695 - which is primarily just to make my life easier when writing tests. It did give me a chance to do a minor spruce up of the related documentation (adding links between methods and reference changes).

Not for this PR, but with this we could allow `save_all` to actually save the data, rather than assuming it has been read in (and do it for all data types, not just PHA). However, this has a lot of "thinking about what's needed" needed first. This is now 
- #1847

# Notes

Writing out ARF data is easy, but for RMF data we have to

- re-construct the original data, as the DataRMF structure has flattened out the on-disk format
- work out what the best format is for the F_CHAN, N_CHAN, and MATRIX columns (scalar, array, or variable-length arrays)
  - I do not guarantee that we match the original on-disk format - for those that have been read in - but it should represent the same actual data

Fix #1699.
Fix #1695.
Fix #203.

### DIFF
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -26,7 +26,7 @@ such as Pulse-Height Amplitudes (PHA) or Pulse-Invariant channels.
 These 'responses' are assumed to follow OGIP standards, such as `OGIP
 Calibration Memo CAL/GEN/92-002, "The Calibration Requirements for
 Spectral Analysis (Definition of RMF and ARF file formats)", Ian
-M. George1, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
+M. George, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
 Michael F. Corcoran
 <https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html>`_.
 
@@ -1369,7 +1369,7 @@ def create_non_delta_rmf(rmflo, rmfhi, fname, offset=1,
         The name of the two-dimensional image file which stores the
         response information (the format of this file matches that
         created by the `CIAO tool rmfimg
-        <http://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
+        <https://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
     offset : int, optional
         The starting channel number: expected to be 0 or 1 but this is
         not enforced.
--- a/sherpa/astro/instrument.py
+++ b/sherpa/astro/instrument.py
@@ -18,23 +18,17 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-"""
-Models of common Astronomical models, particularly in X-rays.
+"""Models of common Astronomical models, particularly in X-rays.
 
 The models in this module include support for instrument models that
 describe how X-ray photons are converted to measurable properties,
 such as Pulse-Height Amplitudes (PHA) or Pulse-Invariant channels.
-These 'responses' are assumed to follow OGIP standards, such as
-[OGIP92-002]_.
-
-References
-----------
-
-.. [OGIP92-002] OGIP Calibration Memo CAL/GEN/92-002, "The Calibration Requirements
-       for Spectral Analysis (Definition of RMF and ARF file formats)",
-       Ian M. George1, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
-       Michael F. Corcoran,
-       https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html
+These 'responses' are assumed to follow OGIP standards, such as `OGIP
+Calibration Memo CAL/GEN/92-002, "The Calibration Requirements for
+Spectral Analysis (Definition of RMF and ARF file formats)", Ian
+M. George1, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
+Michael F. Corcoran
+<https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html>`_.
 
 """
 
@@ -1288,24 +1282,29 @@ def create_delta_rmf(rmflo, rmfhi, offset=1,
     that each channel maps exactly to one energy bin, the
     mapping is monotonic, and there are no gaps.
 
+    .. versionchanged:: 4.16.0
+       The e_min and e_max values will use the rmflo and rmfhi values
+       if not set.
+
     .. versionadded:: 4.10.1
 
     Parameters
     ----------
     rmflo, rmfhi : array
-        The energy bins (low and high, in keV) for the RMF.
-        It is assumed that emfhi_i > rmflo_i, rmflo_j > 0, that the energy
-        bins are either ascending, so rmflo_i+1 > rmflo_i or descending
-        (rmflo_i+1 < rmflo_i), and that there are no overlaps.
-        These correspond to the Elow and Ehigh columns (represented
-        by the ENERG_LO and ENERG_HI columns of the MATRIX block) of
-        the OGIP standard.
+        The energy bins (low and high, in keV) for the RMF.  It is
+        assumed that emfhi_i > rmflo_i, rmflo_j > 0, that the energy
+        bins are either ascending, so rmflo_i+1 > rmflo_i or
+        descending (rmflo_i+1 < rmflo_i), and that there are no
+        overlaps.  These correspond to the Elow and Ehigh columns
+        (represented by the ENERG_LO and ENERG_HI columns of the
+        MATRIX block) of the OGIP standard.
     offset : int, optional
         The starting channel number: expected to be 0 or 1 but this is
         not enforced.
     e_min, e_max : None or array, optional
         The E_MIN and E_MAX columns of the EBOUNDS block of the
-        RMF.
+        RMF. If not set they are taken from rmflo and rmfhi
+        respectively.
     ethresh : number or None, optional
         Passed through to the DataRMF call. It controls whether
         zero-energy bins are replaced.
@@ -1321,6 +1320,7 @@ def create_delta_rmf(rmflo, rmfhi, offset=1,
     See Also
     --------
     create_arf, create_non_delta_rmf
+
     """
 
     # Set up the delta-function response.
@@ -1330,6 +1330,11 @@ def create_delta_rmf(rmflo, rmfhi, offset=1,
     matrix = numpy.ones(nchans, dtype=numpy.float32)
     dummy = numpy.ones(nchans, dtype=numpy.int16)
     f_chan = numpy.arange(1, nchans + 1, dtype=numpy.int16)
+
+    if e_min is None:
+        e_min = rmflo
+    if e_max is None:
+        e_max = rmfhi
 
     return sherpa.astro.data.DataRMF(name, detchans=nchans,
                                      energ_lo=rmflo, energ_hi=rmfhi,
@@ -1343,8 +1348,7 @@ def create_delta_rmf(rmflo, rmfhi, offset=1,
 def create_non_delta_rmf(rmflo, rmfhi, fname, offset=1,
                          e_min=None, e_max=None, ethresh=None,
                          name='delta-rmf', header=None):
-    """
-    Create a RMF using a matrix from a file.
+    """Create a RMF using a matrix from a file.
 
     The RMF matrix (the mapping from channel to energy bin) is
     read in from a file.
@@ -1364,7 +1368,8 @@ def create_non_delta_rmf(rmflo, rmfhi, fname, offset=1,
     fname : str
         The name of the two-dimensional image file which stores the
         response information (the format of this file matches that
-        created by the CIAO tool rmfimg [1]_).
+        created by the `CIAO tool rmfimg
+        <http://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
     offset : int, optional
         The starting channel number: expected to be 0 or 1 but this is
         not enforced.
@@ -1386,11 +1391,6 @@ def create_non_delta_rmf(rmflo, rmfhi, fname, offset=1,
     See Also
     --------
     create_arf, create_delta_rmf
-
-    References
-    ----------
-
-    .. [1] http://cxc.harvard.edu/ciao/ahelp/rmfimg.html
 
     """
 
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -1438,7 +1438,7 @@ def _pack_rmf(dataset):
     # support both types as input.
     #
     if not isinstance(dataset, (DataRMF, RMF1D)):
-        raise IOErr("data set is not an RMF")
+        raise IOErr("data set is not a RMF")
 
     rmfdata = _reconstruct_rmf(dataset)
 
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -64,8 +64,9 @@ import sherpa.io
 from sherpa.utils.err import ArgumentErr, DataErr, IOErr
 from sherpa.utils import SherpaFloat
 from sherpa.data import Data2D, Data1D, BaseData, Data2DInt
-from sherpa.astro.data import DataIMG, DataIMGInt, DataARF, DataRMF, DataPHA, DataRosatRMF
-from sherpa.astro.instrument import ARF1D
+from sherpa.astro.data import DataIMG, DataIMGInt, DataARF, DataRMF, \
+    DataPHA, DataRosatRMF
+from sherpa.astro.instrument import ARF1D, RMF1D
 from sherpa.astro.utils import reshape_2d_arrays
 from sherpa import get_config
 
@@ -114,7 +115,7 @@ info = logging.getLogger(__name__).info
 __all__ = ('backend',
            'read_table', 'read_image', 'read_arf', 'read_rmf', 'read_arrays',
            'read_pha', 'write_image', 'write_pha', 'write_table',
-           'write_arf',
+           'write_arf', 'write_rmf',
            'pack_table', 'pack_image', 'pack_pha', 'read_table_blocks')
 
 
@@ -1121,9 +1122,7 @@ def _pack_arf(dataset):
     }
 
     # Header Keys
-    header = {}
-    if hasattr(dataset, "header"):
-        header = dataset.header.copy()
+    header = dataset.header
 
     # Merge the keywords
     #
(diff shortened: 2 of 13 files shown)


DougBurke:
Rebased as #1841 has been merged.

DougBurke:
Going to rebase and then review because of the merging of 

- #1832

DougBurke:
A note for @hamogu - the backend code (both backends) could use a revamp, but I am reluctant to do that at this time. So although there are some clean up bits in this PR, they're not as extensive as they could be. What I am relying on is that we have a large enough test suite to catch any possible errors. In particular I made sure there are "dog food" tests, where we write out a file and then read it back in, and either

- check it's expected
- check it works as the original version did (e.g. in a fit for writing out PHA+ARF+RMF case).

I have a follow-up PR 
- #1861 

which starts to add some code that may be useful for cleaning up the backends, and

- #1869 

builds on this, but it's still "small" improvements rather than a massive overhaul.

sherpa/astro/ui/tests/test_astro_ui_io.py
```diff
@@ -29,6 +29,9 @@
 from sherpa.astro import ui
 from sherpa.astro.data import DataPHA
 from sherpa.astro.instrument import ARF1D, RMF1D
+from sherpa.utils.err import ArgumentErr
+from sherpa.utils.testing import requires_data, requires_fits, \
```
DougBurke, line 33: Line 28 - which I can't comment on in this interface, is also a `from sherpa.utils.testing` line and can be removed (left over from a less-than-judicious rebase).

hamogu:
> the backend code (both backends) could use a revamp

We should wait with that until Python 3.10 is out minimum version. I forget exactly what it is, but there is some addition to the standard library that would make it easier to convert our backends to use entrypoints, instead of our home-grown "this is how we discover them" code.

hamogu:
Just to comment that I know that this is waiting for my review, but I'm trying to push the plotting PRs to a good place before I get to it.

DougBurke:
No hurry, as we want the plotting changes in.

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+        "TELESCOP": "none",
+        "INSTRUME": "none",
+        "FILTER": "none",
+        "CHANTYPE": "PI",
+        "DETCHANS": "none",
+        "NUMGRP": 0,
+        "NUMELT": 0
+    }
+
+    default_ebounds_header = {
```
hamogu, line 1463: 1/2
```suggestion
    ebounds_header = {
```

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+        "DETCHANS": "none"
+    }
+
+    # Header Keys
+    header = dataset.header
+
+    # Merge the keywords (at least for the MATRIX block).
+    #
+    matrix_header = {**default_matrix_header, **header}
+    ebounds_header = {**default_ebounds_header}
```
hamogu, line 1482: 2/2
```suggestion
```

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+        "CHANTYPE": "PI",
+        "DETCHANS": "none"
+    }
+
+    # Header Keys
+    header = dataset.header
+
+    # Merge the keywords (at least for the MATRIX block).
+    #
+    matrix_header = {**default_matrix_header, **header}
```
hamogu, line 1481: optional. Either works...
```suggestion
    matrix_header = default_matrix_header | header
```

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+
+    # The column ordering for the ouput file is determined by the
+    # order the keys are added to the data dict.
+    #
+    # To allow the backend convert to Variable-Length fields, we
+    # have F_CHAN/N_CHAN/MATRIX either be a 2D ndarray (so fixed
+    # output) OR a list of rows (use VLF). It's not ideal: we could
+    # wrap up in a local VLF type just to indicate this to the
+    # backend.
+    #
```
hamogu, line 1502: But VLF is also optional. The standard even says that you might want to use rectangular matrixes instead of VLF, even is the rectangualr matrix is bigger, because it's much faster to read than VLF. So, don't worry too much about what gets to be a VLF and what gets to be an array.

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+    #  - max(N_GRP) < 4
+    #
+    if len(set(n_grp)) == 1 or n_grp.max() < 4:
+        ny = n_grp.max()
+        f_chan = _make_int_array(f_chan, ny)
+        n_chan = _make_int_array(n_chan, ny)
+    else:
+        f_chan = _make_int_vlf(f_chan)
+        n_chan = _make_int_vlf(n_chan)
+
```
hamogu, line 1365: F_CHAN, N_CHAN are typically very small and by very small I mean that it's only a a few elements per row. Thus the space savings from a VLF are minimal, but the reading speed is slower (in any backend). So, once could argue that it makes sense to always have those as arrays of fixed length. This works and I'm not saying that we need ot change code the @DougBurke already wrote, I'm just suggesting that, if there are any edge cases that turn up later, we could just move it all to a fixed-length array.

hamogu (changes requested):

sherpa/astro/io/pyfits_backend.py
```diff
@@ -78,23 +79,27 @@ def _has_hdu(hdulist, name):
...
+def _try_key(hdu, name, *, fix_type=False, dtype=SherpaFloat):
 
     try:
         key = hdu.header[name]
     except KeyError:
         return None
 
+    # TODO: should this comparison be case insensitive, so that
+    # NONE and None are also matched?
+    #
```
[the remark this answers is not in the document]
DougBurke: So, we have historically had an issue where we have had `None` values get converted to `"None`", causing endless fun later on. Or at least that's my memory of the situation and it's been long enough ago I don't have tickets/PRs to point to.

Saying that, the OGIP standard is not precise when it comes to BACKFILE/RESPFILE/ANCRFILE/CORRFILE - you have an example using "none" in https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/node12.html but I don't believe it's ever exactly specified what case this must be.

sherpa/astro/io/__init__.py
DougBurke: I thought the `|` syntax was a python 3.10-ism but I see from https://docs.python.org/3/library/stdtypes.html#typesmapping it's python 3.9. Which we can assume, so that's nice.

sherpa/astro/io/__init__.py
DougBurke: I'm basically trying to follow the advice given in the OGIP RMF description. We are likely to want to use VLF for the matrix array, so we need to handle VLF data anyway, but this is all internal routine so we can change it if there are found to be problems [ie we don't guarantee exactly how the serialization works, just that it should be "correct"].

This is just background information for the interested reader!

sherpa/astro/io/tests/test_io_response.py
```diff
@@ -183,3 +175,827 @@ def test_read_rmf(make_data_path):
...
+    assert new.e_min[0] == pytest.approx(0.0)
+    assert new.e_max[-1] == pytest.approx(10.239999771118164)
+
+
+@requires_data
+@requires_fits
+def test_write_rmf_fits_rosat_pspc(make_data_path, tmp_path):
+    """Check we can write out an RMF as a FITS file.
+
+    Use the ROSAT PSPCC as it has it's own DataRosatRMF class.  It's
```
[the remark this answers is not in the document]
DougBurke: Well, technically it is the PSPC-C (there was PSPC-B and PSPC-C and this is for PSPC-C)!

sherpa/astro/io/tests/test_io_response.py
```diff
@@ -183,3 +175,827 @@ def test_read_rmf(make_data_path):
...
+    assert new.e_min[0] == pytest.approx(0.01495600026100874)
+    assert new.e_max[-1] == pytest.approx(2.552633047103882)
+
+
+@requires_data
+@requires_fits
+def test_write_rmf_fits_xmm_rgs(make_data_path, tmp_path):
+    """Check we can write out an RMF as a FITS file.
+
+    Try out a grating RMF. Use XMM as smaller than Chandra.
```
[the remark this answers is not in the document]
DougBurke: Whatever gets you into the groove!

sherpa/astro/io/__init__.py
```diff
@@ -46,46 +47,46 @@
...
 from sherpa import get_config
 
 config = ConfigParser()
 config.read(get_config())
 
 # What should we use for the default package?
-io_opt = config.get('options', 'io_pkg', fallback='dummy')
-io_opt = [o.strip().lower() + '_backend' for o in io_opt.split()]
+io_opt = [o.strip().lower() + '_backend' for o in
+          config.get('options', 'io_pkg', fallback='dummy').split()]
```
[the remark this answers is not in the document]
DougBurke: I claim there is a small improvement here because we do not change

```python
io_opt : str
```

to

```python
io_opt : list[str]
```

but that's only relevant for users of systems like `mypy`: for example

```
% cat xs.py
foo = "xx xx xxx"
foo = [f"!{x}!" for x in foo.split()]
```

and

```
% mypy xs.py 
xs.py:2: error: Incompatible types in assignment (expression has type "list[str]", variable has type "str")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

There are different ways around this.

So /end{reflected-snark}

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+        HDUCLAS1 = 'RESPONSE' - extension contains response data.
+        HDUCLAS2 = 'SPECRESP' - extension contains an ARF.
+        HDUVERS = '1.1.0' - version of the file format.
+
+    We do not add the ARFVERSN, HDUVERS1, or HDUVERS2 keywords which
+    are marked as obsolete.
+
+    """
+
+    # The UI layer wraps up a DataARF into ARF1D quite often, so
```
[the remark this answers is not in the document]
DougBurke: I have begun to think there is some architectural reason for this, but I'd need to go and find my Design Patterns book and read it properly to discuss sensibly. The existing documentation is lacking to make us see the reason though. Which I feel is going to be a major area for post 4.16

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+        if cdata is None:
+            raise ArgumentErr("bad", "ARF",
+                              f"{col.upper()} column is missing")
+
+    # The default keywords; these wil be over-ridden by
+    # anything set by the input.
+    #
+    default_header = {
+        "EXTNAME": "SPECRESP",
+        # "HDUNAME": "SPECRESP",    # do we want this or not?
```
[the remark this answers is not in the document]
DougBurke: Related to your comment https://github.com/sherpa/sherpa/pull/1836#discussion_r1330199566 above

DougBurke:
I'm going to check the problem RMF from #1898 with this PR since the crates failure is "ugly" and I want to make sure I haven't made things worse.

sherpa/astro/io/__init__.py
```diff
@@ -675,9 +682,12 @@ def _pack_image(dataset):
...
     data = {}
-    header = {}
-    if hasattr(dataset, 'header') and isinstance(dataset.header, dict):
-        header = dataset.header.copy()
+
+    # Data2D does not have a header
+    try:
+        header = dataset.header
+    except AttributeError:
+        header = {}
```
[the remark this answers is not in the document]
DougBurke: This will actually have to be `getattr(dataset, "header", {})` but I appreciate the heads up (and I note that other lines in this function already use the `getattr` approach...).

sherpa/astro/io/__init__.py
```diff
@@ -702,13 +712,124 @@ def _set_keyword(header, label, value):
...
+    if ukey in ["BITPIX", "BLANK", "BLOCKED", "BSCALE", "BUNIT",
+                "BZERO", "DATAMAX", "DATAMIN", "END", "EXTEND",
+                "EXTLEVEL", "EXTNAME", "EXTVER", "GCOUNT", "GROUPS",
+                "NAXIS", "PCOUNT", "SIMPLE", "TFIELDS", "THEAP",
+                "XTENSION",
+                # other common ones
+                "CHECKSUM", "DATASUM", "CHECKVER",
+                # HEASARC
+                "TSORTKEY",
+                # CIAO specific
```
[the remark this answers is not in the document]
DougBurke: You learn something new every day.

sherpa/astro/io/__init__.py
```diff
@@ -777,6 +893,7 @@ def _pack_pha(dataset):
     #
     default_header = {
         "EXTNAME": "SPECTRUM",
+        # "HDUNAME": "SPECTRUM",  - this is a CIAO-specific keyword
```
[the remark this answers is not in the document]
DougBurke: I'll add a comment above - for https://github.com/sherpa/sherpa/pull/1836/files#r1330194935 - and remove these commented-out lines here.

sherpa/astro/io/__init__.py
```diff
@@ -948,6 +1066,463 @@ def convert(column, dtype):
...
+    # support both types as input.
+    #
+    if not isinstance(dataset, (DataARF, ARF1D)):
+        raise IOErr("data set is not an ARF")
+
+    # Check we have data. Should we allow missing columns?
+    #
+    for col in ["energ_lo", "energ_hi", "specresp"]:
+        cdata = getattr(dataset, col)
+        if cdata is None:
```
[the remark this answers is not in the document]
DougBurke: That's nicer as it points out that we don't actually do anything with the column data (so it is only used for the column-access check).

sherpa/astro/io/__init__.py
DougBurke: This is left over from when I had the different blocks in the same routine, so it matched the logic of the other blocks.

sherpa/astro/io/pyfits_backend.py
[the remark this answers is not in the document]
DougBurke: I will add that this test is I think overly broad since it'll replace the keyword value "This is definitely not none." with None but it does match the crates backend behaviour. I don't want to make too significant a change here at this stage to "improve" this code without a more-throrough review!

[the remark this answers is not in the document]
DougBurke: [and some of my concerns about the None/'None' handling islikely related to just the crates backend]. I'm removing the original comment and replacint it with a "TODO" comment,

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,179 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
...
+        bkg_id : int or str, optional
+           Set if the background ARF should be written out rather
+           than the source ARF.
+        ascii : bool, optional
+           If ``False`` then the data is written as a FITS
+           format binary table. The default is ``True``. The
+           exact format of the output file depends on the
+           I/O library in use (Crates or AstroPy).
+        clobber : bool, optional
+           If `outfile` is not ``None``, then this flag controls
```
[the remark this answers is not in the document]
DougBurke: It should be `filename`.

We can call this as

    save_arf(1, "newarf.fits")

but also (for hysterical raisins)

    save_arf("newarf.fits")

so in this case technically `filename` is None to start with.

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,179 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
...
+        sherpa.utils.err.IOErr
+           If `filename` already exists and `clobber` is ``False``.
+
+        See Also
+        --------
+        create_arf, load_arf, save_pha, save_rmf
+
+        Notes
+        -----
+        The function does not follow the normal Python standards for
```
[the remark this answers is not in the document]
DougBurke: Note there are a lot of places this needs to be done so I'd rather do it as a standalone patch.

sherpa/astro/ui/utils.py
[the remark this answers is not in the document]
DougBurke: But there is some logical inconsistencies with this text...

DougBurke:
Note that we don't have any CI runs to validate the crates changes but I've run them locally against a sherpa-using-CIAO build and they pass (and our integration tests for CIAOX will catch problems too).

sherpa/astro/io/pyfits_backend.py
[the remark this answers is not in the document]
DougBurke: I'll leave this for @hamogu to decide whether it's been resolved or needs more discussion.

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,189 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
...
+           are multiple responses.
+        bkg_id : int or str, optional
+           Set if the background ARF should be written out rather
+           than the source ARF.
+        ascii : bool, optional
+           If ``False`` then the data is written as a FITS format
+           binary table. The exact format of the output file depends
+           on the I/O library in use (Crates or AstroPy).
+        clobber : bool, optional
+           This flag controls whether an existing file can be
```
DougBurke, line 4852: Hopefully the new text for id/filename/clobber/.. is an improvement.

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,179 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
...
+           is determined by the `ascii` argument.
+        resp_id : int or str, optional
+           The identifier for the ARF within this data set, if there
+           are multiple responses.
+        bkg_id : int or str, optional
+           Set if the background ARF should be written out rather
+           than the source ARF.
+        ascii : bool, optional
+           If ``False`` then the data is written as a FITS
+           format binary table. The default is ``True``. The
```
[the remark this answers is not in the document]
DougBurke: As noted elsewhere I've reworked the description of id/filename/ascii/clobber and so they are hopefully better. I'll resolve the other comments and just leave this open.

As to whether we want to allow an ARF to be written out as an ASCII file, I don't see the harm in it (with the understanding that it isn't something you can load back in and fit as the metadata is lost). We have the `ascii` option for `save_pha`, so it matches that.

DougBurke:
I will also note that the follow on PR in #1861 has more clean ups of some of the backend code (but will need to be rebased to address recent changes here).

DougBurke:
Hmm. An interesting failure for the arm/arch build which is actually unrelated to this PR (so I'm going to turn off the CI:arch label

```
_____________________ test_binop_string_with_custom_ufunc ______________________

    def test_binop_string_with_custom_ufunc():
        '''Repeat the previous test, but with a custom ufunc.'''
        def func(a, b):
            return a + b
    
        uf = np.frompyfunc(func, nin=2, nout=1)
        p, p2 = setUp_composite()
        comp = uf(p, p2)
>       assert repr(comp) == "<BinaryOpParameter 'func(model.p, model.p2)'>"
E       assert "<BinaryOpPar..., model.p2)'>" == "<BinaryOpPar..., model.p2)'>"
E         - <BinaryOpParameter 'func(model.p, model.p2)'>
E         ?                     ^^^^
E         + <BinaryOpParameter '?(model.p, model.p2)'>
E         ?                     ^

sherpa/models/tests/test_parameter.py:237: AssertionError

_____________________ test_binop_string_with_custom_ufunc ______________________

    def test_binop_string_with_custom_ufunc():
        '''Repeat the previous test, but with a custom ufunc.'''
        def func(a, b):
            return a + b
    
        uf = np.frompyfunc(func, nin=2, nout=1)
        p, p2 = setUp_composite()
        comp = uf(p, p2)
>       assert repr(comp) == "<BinaryOpParameter 'func(model.p, model.p2)'>"
E       assert "<BinaryOpPar..., model.p2)'>" == "<BinaryOpPar..., model.p2)'>"
E         - <BinaryOpParameter 'func(model.p, model.p2)'>
E         ?                     ^^^^
E         + <BinaryOpParameter '?(model.p, model.p2)'>
E         ?                     ^

sherpa/models/tests/test_parameter.py:237: AssertionError
```

sherpa/astro/io/pyfits_backend.py
```diff
@@ -78,23 +79,27 @@ def _has_hdu(hdulist, name):
     return True
 
 
-def _try_key(hdu, name, fix_type=False, dtype=SherpaFloat):
+def _try_key(hdu, name, *, fix_type=False, dtype=SherpaFloat):
 
     try:
         key = hdu.header[name]
     except KeyError:
         return None
```
hamogu, line 87:
```suggestion
    key = hdu.header.get(name, None)
```

sherpa/astro/io/crates_backend.py
```diff
@@ -106,52 +113,62 @@ def get_filename_from_dmsyntax(filename, blockname=None):
...
+    key = crate.get_key(name)
     if key is None:
         return None
 
-    key = key.value
+    value = key.value
 
-    if str(key).find('none') != -1:
+    # Note that this check seems a bit too permissive, since a value
+    # of "This key is not set to none" will be matched.
```
hamogu, line 150: but since key in FITS are limited to 8 characters, we are probably OK in practice.

sherpa/astro/io/crates_backend.py
```diff
@@ -314,8 +318,7 @@ def _try_col_list(crate, colname, num, make_copy=False, fix_type=False,
         return numpy.array([None] * num)
 
     col = crate.get_column(colname)
-
-    if hasattr(col, 'is_varlen') and col.is_varlen():
+    if col.is_varlen():
```
hamogu, line 321: Do all cols in crates are "is_varlen" and the previous code was just too cautious or why was this needed before, but no longer?

sherpa/astro/io/tests/test_io_response.py
```diff
@@ -652,8 +652,9 @@ def test_write_rmf_fits_swift_xrt(make_data_path, tmp_path):
 def test_write_rmf_fits_rosat_pspc(make_data_path, tmp_path):
     """Check we can write out an RMF as a FITS file.
 
-    Use the ROSAT PSPCC as it has it's own DataRosatRMF class.  It's
-    also a RSP rather than a RMF (ie includes the ARF, I believe).
+    Use the ROSAT PSPC-C example dataset as it has it's own
```
hamogu, line 655:
```suggestion
    Use the ROSAT PSPC-C example dataset as it has its own
```

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,11 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
         sherpa.astro.io.write_pha(filename, d, ascii=ascii,
                                   clobber=clobber)
 
+    # This should be
+    #     def save_arf(self, id, filename=None, *, resp_id=None, ...
+    # but the existing logic used to create the ui moduke does not
```
hamogu, line 4821:
```suggestion
    # but the existing logic used to create the ui module does not
```

hamogu (changes requested):
Almost there...

sherpa/astro/io/pyfits_backend.py
DougBurke: doh,

sherpa/astro/io/crates_backend.py
DougBurke: a) you can actually have a longer key, it just needs "shenanighans"

b) the use of the name "key" for the argument here is a bit confusing and it actually refers to the value being stored, not the name of the key (and crates has a concept of key which contains all the relevant values). I actually have it renamed to `value` by the end of this PR

sherpa/astro/io/crates_backend.py
DougBurke: Columns are represented by CrateData objects and they all have an is_varlen method. I believe this is just old code from when support for this method was being introduced, but that was a **long** time ago.

sherpa/astro/io/pyfits_backend.py
DougBurke: Actually it needs to be

```suggestion
    key = hdu.header.get(name, None)
    if key is None:
        return None
```

and it also suffers from the poorly-named variable, as `name` is the "key"; this is more like value but that's a larger change to make.

sherpa/astro/ui/utils.py
```diff
@@ -4816,6 +4816,189 @@ def save_pha(self, id, filename=None, bkg_id=None, ascii=False,
...
+            resp_id = 1 if resp_id is None else resp_id
+            emsg += f"data set '{id}' (response {resp_id}) does not contain an ARF"
+            raise ArgumentErr(emsg)
+
+        sherpa.astro.io.write_arf(filename, arf, ascii=ascii,
+                                  clobber=clobber)
+
+    # This should be
+    #     def save_rmf(self, id, filename=None, *, resp_id=None, ...
+    # but the existing logic used to create the ui moduke does not
```
DougBurke, line 4918:
```suggestion
    # but the existing logic used to create the ui module does not
```

hamogu (approved):

sherpa/astro/instrument.py
```diff
@@ -18,23 +18,17 @@
...
-
-.. [OGIP92-002] OGIP Calibration Memo CAL/GEN/92-002, "The Calibration Requirements
-       for Spectral Analysis (Definition of RMF and ARF file formats)",
-       Ian M. George1, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
-       Michael F. Corcoran,
-       https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html
+These 'responses' are assumed to follow OGIP standards, such as `OGIP
+Calibration Memo CAL/GEN/92-002, "The Calibration Requirements for
+Spectral Analysis (Definition of RMF and ARF file formats)", Ian
+M. George1, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
```
wmclaugh, line 29: Since we are not including the footnote for the '1' after Ian M. George1, should the 1 be removed.

sherpa/astro/instrument.py
```diff
@@ -1364,7 +1368,8 @@ def create_non_delta_rmf(rmflo, rmfhi, fname, offset=1,
     fname : str
         The name of the two-dimensional image file which stores the
         response information (the format of this file matches that
-        created by the CIAO tool rmfimg [1]_).
+        created by the `CIAO tool rmfimg
+        <http://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
```
wmclaugh, line 1372:
```suggestion
        <https://cxc.harvard.edu/ciao/ahelp/rmfimg.html>`_).
```

sherpa/astro/instrument.py
DougBurke, line 29:
```suggestion
M. George, Keith A. Arnaud, Bill Pence, Laddawan Ruamsuwan and
```

not sure where that came from...

DougBurke:
Yayyyyyyy! Thanks all.

Issue #1817, opened 2023-07-11
'plot_delchi' discrepancies with plot generated using 'get_delchi_plot' arrays


nplee:
Hi all,

This came in through helpdesk, and I'm able to independently generate a similar case using our standard `3c273.pi` example set from the Sherpa threads.  It seems like the errors for the x-axis from `get_delchi_plot` are different than that displayed by `plot_delchi`.  

An example is below:

```
sherpa In [1]: import matplotlib.pyplot as plt

sherpa In [2]: load_data("3c273.pi")
WARNING: systematic errors were not found in file '3c273.pi'
statistical errors were found in file '3c273.pi' 
but not used; to use them, re-read with use_errors=True
read ARF file 3c273.arf
read RMF file 3c273.rmf
WARNING: systematic errors were not found in file '3c273_bg.pi'
statistical errors were found in file '3c273_bg.pi' 
but not used; to use them, re-read with use_errors=True
read background file 3c273_bg.pi

sherpa In [3]: set_source(powlaw1d.p1)

sherpa In [4]: fit()
WARNING: data set 1 has associated backgrounds, but they have not been subtracted, nor have background models been set
Dataset               = 1
Method                = levmar
Statistic             = chi2gehrels
Initial fit statistic = 8.15838e+10
Final fit statistic   = 50.5351 at function evaluation 25
Data points           = 46
Degrees of freedom    = 44
Probability [Q-value] = 0.231118
Reduced statistic     = 1.14852
Change in statistic   = 8.15838e+10
   p1.gamma       1.72329      +/- 0.0501892   
   p1.ampl        0.000147339  +/- 8.26966e-06 

sherpa In [5]: plot_delchi()

sherpa In [6]: delchiplot = get_delchi_plot()

sherpa In [7]: plt.errorbar(delchiplot.x, delchiplot.y, xerr=delchiplot.xerr, yerr=delchiplot.yerr, ls='none', alpha=0.5)
Out[7]: <ErrorbarContainer object of 3 artists>
```

![delchi_plot](https://github.com/sherpa/sherpa/assets/23193862/639daee7-1fe2-43c7-8839-bff2e1282ffc)

```
sherpa In [8]: print(delchiplot)
x      = [ 0.1245, 0.2774, 0.3869, 0.5183, 0.6059, 0.6716, 0.7227, 0.7665, 0.803 ,
[...]

PR #1839, opened 2023-08-03
ui: add get_plot_prefs/get_contour_prefs calls

DougBurke:
# Summary

Add the get_plot_prefs and get_contour_prefs call to simplify access to the plot and contour preferences.

# Details

This was suggested in https://github.com/sherpa/sherpa/issues/1817#issuecomment-1658504691

For users who want to change the delchi plot preferences this replaces

    prefs = get_resid_plot(...).plot_prefs or .histo_prefs

with

    prefs = get_plot_prefs("resid")

It's a small use-of-life improvement, and we could decide to deprecate the get_data/model_plot/contour_prefs() calls.

There is a subtle issue with the "fit" option, since get_fit_plot() returns

    .dataplot
    .modelplot
    .plot_prefs   which is just {}

and the code would (by defualt) return the plot_prefs setting, which does not do anything. I decided to make "fit" error out rather than have it potentially confuse users. Since the plot option also has to worry about "bkg_fit" from the astro layer we can not just check on "fit".

Another issue with the plot command is we allow **kwargs to be set: this means we can easily support the "bkg_id" argument for the bkg-style plots in the astro layer without having to overload get_plot_prefs. The downside is that the documentation is harder to read, as we do not have a bkg_id keyword.

It's essentially the same thing for the contour plots, but simpler, as:

- do not have to deal with plot_prefs vs histo_prefs
- do not have to worry about extra arguments from "bkg" commands in the astro layer

### DIFF
--- a/docs/ui/sherpa_astro_ui.rst
+++ b/docs/ui/sherpa_astro_ui.rst
@@ -89,6 +89,7 @@ burden...
       get_conf_opt
       get_conf_results
       get_confidence_results
+      get_contour_prefs
       get_coord
       get_counts
       get_covar
@@ -146,6 +147,7 @@ burden...
       get_pdf_plot
       get_photon_flux_hist
       get_pileup_model
+      get_plot_prefs
       get_prior
       get_proj
       get_proj_opt
--- a/docs/ui/sherpa_ui.rst
+++ b/docs/ui/sherpa_ui.rst
@@ -53,6 +53,7 @@ hidden away. This needs better explanation...
       get_conf_opt
       get_conf_results
       get_confidence_results
+      get_contour_prefs
       get_covar
       get_covar_opt
       get_covar_results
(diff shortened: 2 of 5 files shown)


DougBurke:
@anetasie - I've assigned this to you for you to check the behaviour out, before passing it to Moritz.

So,

- is this worth doing (it adds more commands we need to document/test/...)?
  - if so, is the documentation good enough
- do we want to deprecate the get_data/model_plot_prefs and get_data/model_contour_prefs commands?

hamogu (approved):

anetasie (approved):

anetasie:
The new commands are easier to use. Thanks.

We could deprecate the get_data/model_prefs commands, but I'm concerned about them being used now. Do we need a deprecation warning saying that they will be removed in the next release?

PR #1840, opened 2023-08-03
Fix #1828 - histogram plot issue

### LINKED ISSUE #1838: histogram plotting bug
With CIAO 4.15 but also the current main branch we have

```
sherpa In [1]: load_arrays(1, [1, 2, 3], [2, 2.5, 4], [12, 5, 11], Data1DInt)

sherpa In [2]: set_source(const1d.foo)

sherpa In [3]: plot_fit()
```

which results in 

plot

The model plot is not showing the middle bin (x=2 to 2.5) correctly. This also happens for plot_model() but the fit output shows that the central bin is known about.

If we look at the data, the model is being evaluated correctly:

```
sherpa In [6]: mplot = get_model_plot()

sherpa In [7]: mplot.xlo, mplot.x, mplot.xhi
Out[7]: (array([1, 2, 3]), array([1.5 , 2.25, 3.5 ]), array([2. , 2.5, 4. ]))

sherpa In [8]: mplot.y
Out[8]: array([1. , 0.5, 1. ])
```

which suggests it's the histogram plotting code I added to the matplotlib backend that is at fault.

If I make the bins consecutive it works:

```
sherpa In [9]: load_arrays(1, [1, 2, 3], [2, 3, 4], [12, 5, 11], Data1DInt)

sherpa In [10]: plot_fit()

sherpa In [11]: plt.savefig("/home/dburke/plot2.png")
```

plot2

If I make each bin separate (ie not connected) we get no model data plotted.

```
sherpa In [12]: load_arrays(1, [1, 2, 3], [1.5, 2.5, 3.5], [12, 5, 11], Data1DIn
           ...: t)

sherpa In [13]: plot_fit()
```

plot3

gah.

DougBurke:
# Summary

Fix up the histogram display for the unlikely case when the bin edges have at least one gap, the low edges are integer values, and the high edges are floating-point values.

# Details

Fix #1838 

We have elected to remove the sherpa.plot.utils.intersperse routine rather than deprecate it as

- this is a relatively-new module so it is unlikely people are using it
- it is a very-low-level module
- it is an easy routine to replace

# Example

Taking the example from #1838 we now get

```

In [2]: load_arrays(1, [1, 2, 3], [2, 2.5, 4], [12, 5, 11], Data1DInt)

In [3]: set_source(const1d.foo)

In [4]: plot_fit()
```

lines

### DIFF
--- a/sherpa/plot/tests/test_plot_utils.py
+++ b/sherpa/plot/tests/test_plot_utils.py
@@ -130,7 +130,6 @@ def test_histogram_line_integers_single_gap():
     assert yh[good] == pytest.approx([12, 12, 3, 3, 15, 15])
 
 
-@pytest.mark.xfail    
 def test_histogram_line_floats_single_gap():
     """Test based on issue #1838
 
--- a/sherpa/plot/tests/test_plot_utils.py
+++ b/sherpa/plot/tests/test_plot_utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2021
+#  Copyright (C) 2021, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -18,6 +18,8 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 import numpy as np
+
+import pytest
 
 from sherpa.plot.utils import histogram_line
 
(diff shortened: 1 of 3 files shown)


sherpa/plot/tests/test_plot_utils.py
```diff
@@ -53,3 +55,105 @@ def test_histrogram_line():
...
+
+    assert xh[good] == pytest.approx([10, 20, 20, 25, 30, 40])
+    assert yh[good] == pytest.approx([12, 12, 3, 3, 15, 15])
+
+
+def test_histogram_line_floats_single_gap():
+    """Test based on issue #1838
+
+    This is just a regression test to document the current behavior
+    (apart from the fact this fails because of #1838).
```
hamogu, line 137:
```suggestion
```

You fixed it with the commit right after the one that added the text.

hamogu (changes requested):
I suggest that we remove the "intersperse" routine. It's such a small utility that I just can't imagine any Sherpa user found it in the Sherpa code and then also decided to use it.
We need to balance our (mental) workload with desire to be backwards compatible and here it's clear to me that the added burden of deprecating it now and adding an issue to remember to remove it next time is not worth the effort. 

I don't think we have a formal deprecation policy anyway that would tell us how many major releases we would have to wait after adding a deprecation warning, so let's just remove it right now.

DougBurke:
I'm happy with removing code deep in the weeds from a module like this. I'll make the change.

hamogu (approved):

PR #1841, opened 2023-08-03
Minor refresh of AstroPy I/O code for better FITS and ASCII output

DougBurke:
# Summary

Minor rework of the sherpa.astro.io.pyfits_backend module to improve on the output (both FITS and ASCII), such as a closer mapping of the data type used on disk and written out to file, and some changes to the header lines of ASCII files.

# Details

I started #1836 since I want to be able to write out ARFs and RMFs. However, I realized the FITS files we were creating didn't match what I wanted, which has lead to this change. I assume it overlaps with #1091 but I have not looked at that PR. I am trying to not make large changes here, as I really want to focus on the new functionality that I haven't been able to write yet, so there are other changes that could be made that I have not looked at.

# Example

## CIAO 4.15 behavior with pyfits backend

```
% SHERPARC=sherpa-standalone.rc sherpa
Sherpa 4.15.0

Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:40:32) [GCC 12.3.0]
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Installed qt5 event loop hook.
Shell is already running a gui event loop for qt5. Call with no arguments to disable the current loop.
sherpa In [1]: from sherpa.astro import io

sherpa In [2]: io.backend
Out[2]: <module 'sherpa.astro.io.pyfits_backend' from '/home/dburke/micromamba/envs/ciao415/lib/python3.10/site-packages/sherpa/astro/io/pyfits_backend.py'>

sherpa In [3]: load_arrays(1, [1,2,3], [12,2,4], DataPHA)

sherpa In [4]: set_exposure(230)

sherpa In [5]: set_backscal(0.125)

sherpa In [6]: pha = get_data()

sherpa In [7]: hdu = io.pack_pha(pha)

sherpa In [8]: print(hdu)
<astropy.io.fits.hdu.table.BinTableHDU object at 0x7f2cdc365c30>

sherpa In [9]: hdu.columns
Out[9]: 
ColDefs(
    name = 'CHANNEL'; format = 'INT32'
    name = 'COUNTS'; format = 'INT32'
)

sherpa In [10]: hdu.header
Out[10]: 
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                    4 / length of dimension 1                          
NAXIS2  =                    3 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    2 / number of table fields                         
HDUCLASS= 'OGIP    '                                                            
HDUCLAS1= 'SPECTRUM'                                                            
HDUCLAS2= 'TOTAL   '                                                            
HDUCLAS3= 'TYPE:I  '                                                            
HDUCLAS4= 'COUNT   '                                                            
HDUVERS = '1.2.1   '                                                            
HDUDOC  = 'Arnaud et al. 1992a Legacy 2  p 65'                                  
TELESCOP= 'none    '                                                            
INSTRUME= 'none    '                                                            
FILTER  = 'none    '                                                            
CORRFILE= 'none    '                                                            
CORRSCAL=                    0                                                  
CHANTYPE= 'PI      '                                                            
RESPFILE= 'none    '                                                            
ANCRFILE= 'none    '                                                            
BACKFILE= 'none    '                                                            
POISSERR=                    T                                                  
EXPOSURE=                230.0                                                  
BACKSCAL=                0.125                                                  
AREASCAL=                  1.0                                                  
SYS_ERR =                  0.0                                                  
QUALITY =                    0                                                  
GROUPING=                    0                                                  
TLMIN1  =                    1                                                  
TLMAX1  =                    3                                                  
DETCHANS=                    3                                                  
TTYPE1  = 'CHANNEL '                                                            
TFORM1  = 'INT32   '                                                            
TTYPE2  = 'COUNTS  '                                                            
TFORM2  = 'INT32   '                                                            
EXTNAME = 'SPECTRUM'           / extension name                                 

sherpa In [11]: hdu.name
Out[11]: 'SPECTRUM'
```

Note the invalid TFORM1/2 keywords.

## This PR

```
In [2]: from sherpa.astro.ui import *

In [3]: load_arrays(1, [1,2,3], [12,2,4], DataPHA)

In [4]: set_exposure(230)

In [5]: set_backscal(0.125)

In [6]: pha = get_data()

In [7]: from sherpa.astro import io

In [8]: hdu = io.pack_pha(pha)

In [9]: print(hdu)
<astropy.io.fits.hdu.table.BinTableHDU object at 0x7f9a45bf6ed0>

In [10]: hdu.columns
Out[10]: 
ColDefs(
    name = 'CHANNEL'; format = 'J'
    name = 'COUNTS'; format = 'J'
)

In [11]: hdu.header
Out[11]: 
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                    8 / length of dimension 1                          
NAXIS2  =                    3 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    2 / number of table fields                         
TTYPE1  = 'CHANNEL '                                                            
TFORM1  = 'J       '                                                            
TTYPE2  = 'COUNTS  '                                                            
TFORM2  = 'J       '                                                            
EXTNAME = 'SPECTRUM'                                                            
HDUCLASS= 'OGIP    '                                                            
HDUCLAS1= 'SPECTRUM'                                                            
HDUCLAS2= 'TOTAL   '                                                            
HDUCLAS3= 'TYPE:I  '                                                            
HDUCLAS4= 'COUNT   '                                                            
HDUVERS = '1.2.1   '                                                            
HDUDOC  = 'Arnaud et al. 1992a Legacy 2  p 65'                                  
TELESCOP= 'none    '                                                            
INSTRUME= 'none    '                                                            
FILTER  = 'none    '                                                            
CORRFILE= 'none    '                                                            
CORRSCAL=                    0                                                  
CHANTYPE= 'PI      '                                                            
RESPFILE= 'none    '                                                            
ANCRFILE= 'none    '                                                            
BACKFILE= 'none    '                                                            
POISSERR=                    T                                                  
EXPOSURE=                230.0                                                  
BACKSCAL=                0.125                                                  
AREASCAL=                  1.0                                                  
SYS_ERR =                  0.0                                                  
QUALITY =                    0                                                  
GROUPING=                    0                                                  
TLMIN1  =                    1                                                  
TLMAX1  =                    3                                                  
DETCHANS=                    3                                                  

In [12]: hdu.name
Out[12]: 'SPECTRUM'
```

We can see the columns now have a TFORM of `J` - which is 32-bit int - e.g. https://docs.astropy.org/en/stable/io/fits/usage/table.html#column-creation - which is closer to the truth (need to check whether J is okay or whether it should be I for 16-bit int, but it's more valid than "INT32".

Note however when the "INT32" columns get written they do get converted to Int2 (not sure the logic of how the replacement is made by AstroPy). The new version is written out as Int4 (because the TFORM is valid).

## ASCII output differences

After

```
>>> load_arrays(1, [1, 2, 3], [5, 2, 3])
>>> load_arrays(2, [1, 2, 3], [2.5, 3, 4], [5, 2, 3], Data1DInt)
>>> save_data(1, "tbl1.dat", ascii=True)
>>> save_data(2, "tbl1.dat", ascii=True)
```

and then if we diff the tbl1,dat file (old vs new)

```
< #X Y
---
> # X Y
```

and for tbl2.dat

```
< #XLO XHI Y
---
> # XLO XHI Y
```

That is, rather than the header line being `#<col1> <col2> ...` we now get `# <col1> <col2> ..` (actually, it's hard to guarantee this for all files as there's a bunch of ways the ASCII gets written out, but I'm pretty sure this holds).

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -45,6 +45,7 @@ import numpy
 
 from astropy.io import fits
 from astropy.io.fits.column import _VLF
+from astropy.table import Table
 
 from sherpa.utils.err import IOErr
 from sherpa.utils import SherpaInt, SherpaUInt, SherpaFloat
@@ -997,23 +998,37 @@ def get_pha_data(arg, make_copy=False, use_background=False):
 
 # Write Functions
 
-def _create_columns(col_names, data):
+def _create_table(names, data):
+    """Create a Table.
 
-    collist = []
-    cols = []
-    coldefs = []
-    for name in col_names:
+    The idea is that by going via a Table we let the AstroPy
+    code deal with all the conversions (e.g. to get the FITS
+    data types on columns correct).
+
+    Parameters
+    ----------
+    names : list of str
+        The order of the column names (must exist in data).
+    data : dict[str, ndarray or None]
+        Any None values are dropped.
+
+    Returns
+    -------
+    tbl : Table
+
+    """
+
+    store = []
+    colnames = []
+    for name in names:
         coldata = data[name]
         if coldata is None:
             continue
 
-        col = fits.Column(name=name.upper(), array=coldata,
-                          format=coldata.dtype.name.upper())
-        cols.append(coldata)
-        coldefs.append(name.upper())
-        collist.append(col)
+        store.append(coldata)
+        colnames.append(name)
 
-    return collist, cols, coldefs
+    return Table(names=colnames, data=store)
 
 
 def check_clobber(filename, clobber):
@@ -1031,25 +1046,20 @@ def set_table_data(filename, data, col_names, header=None,
     if not packup:
         check_clobber(filename, clobber)
 
-    col_names = list(col_names)
-
-    # The code used to create a header containing the
-    # exposure, backscal, and areascal keywords, but this has
-    # since been commented out, and the code has now been
-    # removed.
-
-    collist, cols, coldefs = _create_columns(col_names, data)
-
+    tbl = _create_table(col_names, data)
     if ascii:
-        set_arrays(filename, cols, coldefs, ascii=ascii,
-                   clobber=clobber)
+        tbl.write(filename, format='ascii.commented_header',
+                  overwrite=clobber)
         return
 
-    tbl = fits.BinTableHDU.from_columns(fits.ColDefs(collist))
-    tbl.name = 'HISTOGRAM'
+    hdu = fits.table_to_hdu(tbl)
+    if hdu.name == '':
+        tbl.name = 'HISTOGRAM'
+
     if packup:
-        return tbl
-    tbl.writeto(filename, overwrite=True)
+        return hdu
+
+    hdu.writeto(filename, overwrite=True)
 
 
 def _create_header(header):
@@ -1058,11 +1068,11 @@ def _create_header(header):
     """
 
     hdrlist = fits.Header()
-    for key in header.keys():
-        if header[key] is None:
+    for key, value in header.items():
+        if value is None:
             continue
 
-        _add_keyword(hdrlist, key, header[key])
+        _add_keyword(hdrlist, key, value)
 
     return hdrlist
 
@@ -1073,21 +1083,19 @@ def set_pha_data(filename, data, col_names, header=None,
     if not packup:
         check_clobber(filename, clobber)
 
-    hdrlist = _create_header(header)
-
-    collist, cols, coldefs = _create_columns(col_names, data)
-
+    tbl = _create_table(col_names, data)
     if ascii:
-        set_arrays(filename, cols, coldefs, ascii=ascii,
-                   clobber=clobber)
+        tbl.write(filename, format='ascii.commented_header',
+                  overwrite=clobber)
         return
 
-    pha = fits.BinTableHDU.from_columns(fits.ColDefs(collist),
-                                        header=fits.Header(hdrlist))
-    pha.name = 'SPECTRUM'
+    hdu = fits.table_to_hdu(tbl)
+    hdu.header.extend(_create_header(header))
+
     if packup:
-        return pha
-    pha.writeto(filename, overwrite=True)
+        return hdu
+
+    hdu.writeto(filename, overwrite=True)
 
 
 def set_image_data(filename, data, header, ascii=False, clobber=False,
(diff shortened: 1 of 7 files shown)


hamogu:
Question: How much do we care about the details of how the output looks? Are we OK if it's "valid fits with the right information" or "valid ASCII" or do we want to guarantee the same location of spaces in the header, the same ordering of columns. While that might not be required (e.g. fits columns should be addressed by name, not position), it would show us if anything changed. We clearly don't want to reproduce fits-writer tests here, but we could capture a 2 line ASCII table via `StringIO` and compare it to a hardcoded version in a test. For fits, we could make a small samples fits file for a spectrum (just two spectral bins and the minimal required header info), read it in and write it out again and see how the output files compare (I know that sounds simple, but is not, e.g. if the fits writer added the DATE keyword for when the file is created, so the "how do the files compare" needs to read the file back in or use something like astropy's fitsdiff, which has options to autmatically ignore COMMENT cards and similar).

This is not a review with a request for changes, just a comment to discuss if this is worthwhile or not. Even if the answer is "yes" it does not have to happen in this PR.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -1163,23 +1179,29 @@ def set_arrays(filename, args, fields=None, ascii=True, clobber=False):
...
-    if not ascii and fields is None:
-        fields = [f'col{ii + 1}' for ii in range(len(args))]
-
     if fields is not None and len(args) != len(fields):
         raise IOErr("wrongnumcols", len(args), len(fields))
 
     if ascii:
-        write_arrays(filename, args, fields, clobber=clobber)
+        # Historically, the serialization doesn't quite match the
+        # AstroPy Table version, so stay with write_arrays. We do
```
hamogu, line 1187: Astropy table surely has options to make the format we want (I promise!) but since we have `write_tables` already, we might as well keep it here.

hamogu:
> For fits, we could make a small samples fits file for a spectrum (just two spectral bins and the minimal required header info), read it in and write it out again

I found the tests that do something similar to this, so. I think we are good. My initial comment was based on the warning that https://github.com/sherpa/sherpa/pull/1841/files#diff-6c7a81f3e8a175c6398d968cf0af7b3ce90b444544805f9d7a9558c210637ca0L1175 is not covered by tests - but I think that's OK on closer look.

hamogu (approved):

sherpa/astro/io/pyfits_backend.py
DougBurke: Yeah - I am trying to not significantly change things at this time. For some definition of "significant".

One thing I spent time on when adding tests to the I/O code a while ago was to try and ensure we had "round trip" tests - e.g. that we can read back in a file we create. The idea is that the output text may change slightly (e.g. the column headings) but conceptually it's the same data. Or, at least, that's the plan.

PR #1844, opened 2023-08-10
XSPEC: allow table model redshift parameter range to change (fix #1814)

### LINKED ISSUE #1814: redshift range hard-coded in table models
In sherpa '4.15.0', loading a atable with load_table_model adds a redshift parameter, with hard-coded hard limits z=0-5. I would like to allow higher redshifts.

It seems difficult to work around this. In earlier sherpa versions one could override hard_max.

`__setattr__` disallows replacing the redshift parameter.

DougBurke:
# Summary

Allow the redshift range of XSPEC table models to be changed outside the default 0 to 5 range. This restores a feature that was lost in the 4.15 release. Fix #1814

# Details

In #1259 (Sep 2021) I updated the XSPEC models to generally use the XSParameter rather than Parameter class for parameters. This was to allow users to over-ride the hard min/max ranges, as this is a feature in XSPEC used by a number of models which use a value < 0 (say, where 0 is the minimum parameter value) to change behavior in some way.

However, some parameters were marked as "do not allow the limit to change" (more like the existing Parameter hard limits) using the XSBaseParameter class. This is used for table models where we had experience with crashes and memory corruption if the parameter range were exceeded.

One of the parameters that can - but not always - get added to a XSPEC table model is the redshift parameter, which currently XSPEC (circa 12.13.1 but it has been like this for a long while) sets to 0 with a range of 0 to 5, so we use the same range. The class was set as a XSBaseParameter to match the other table parameters, but there's less need to do this cmopared to the other parameters, as it is not-as-directly related to memory-crashing fun as the other parameters. And a user was setting values to outsize the z=0 to 5 range and the 4.15 release where all this was added broke things for them (#1814).

So we switch just the redshift parameter from XSBaseParameter to XSParameter. Unfortunately we do not have a XSPEC table model we can use to check this out with (in the tests at least, there are models such as the additive table model

https://sites.astro.caltech.edu/~mislavb/download/borus02_v170323/borus02_v170323a.fits

that we can use to test this off to the side).

# Example

```

In [2]: load_xstable_model("joe", "../data/borus02_v170323a.fits")

In [3]: print(joe.redshift)
val         = 0.0
min         = 0.0
max         = 5.0
units       = 
frozen      = True
link        = None
default_val = 0.0
default_min = 0.0
default_max = 5.0

In [4]: joe.redshift.max = 8
ParameterErr                              Traceback (most recent call last)
Cell In[4], line 1
----> 1 joe.redshift.max = 8

File ~/sherpa/sherpa-main/sherpa/utils/__init__.py:135, in NoNewAttributesAfterInit.__setattr__(self, name, val)
    131     if not cname and cval:
    132         raise AttributeError(f"'{type(self).__name__}' object attribute '{name}' "
    133                              "cannot be replaced with a callable attribute")
--> 135 object.__setattr__(self, name, val)

File ~/sherpa/sherpa-main/sherpa/models/parameter.py:230, in _make_set_limit.<locals>._set_limit(self, val)
    227     raise ParameterErr('edge', self.fullname,
    228                        'hard minimum', self._hard_min)
    229 if val > self._hard_max:
--> 230     raise ParameterErr('edge', self.fullname,
    231                        'hard maximum', self._hard_max)
    233 # Ensure that we don't try to set a parameter range, such that
    234 # the minimum will be greater than the current parameter value,
    235 # or that the maximum will be less than the current parameter value.
   (...)
    244 # downgrade the ParameterErr issued here to mere warnings.  Also,
    245 # set the value to the appropriate soft limit.
    246 if hasattr(self, "_NoNewAttributesAfterInit__initialized") and \
    247    self._NoNewAttributesAfterInit__initialized:

ParameterErr: parameter joe.redshift has a hard maximum of 5

In [5]: joe.redshift.hard_max = 10

In [6]: joe.redshift.val
Out[6]: 0.0

In [7]: joe.redshift.max = 8

In [8]: print(joe.redshift)
val         = 0.0
min         = 0.0
max         = 8.0
units       = 
frozen      = True
link        = None
default_val = 0.0
default_min = 0.0
default_max = 5.0

```

If we try this with 4.15 we get:

```
sherpa In [1]: load_xstable_model("joe", "data/borus02_v170323a.fits")

sherpa In [2]: print(joe.redshift)
val         = 0.0
min         = 0.0
max         = 5.0
units       = 
frozen      = True
link        = None
default_val = 0.0
default_min = 0.0
default_max = 5.0

sherpa In [3]: joe.redshift.hard_max = 10
AttributeError: can't set attribute 'hard_max'

sherpa In [4]: joe.redshift.max = 10
ParameterErr: parameter joe.redshift has a hard maximum of 5
```

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -774,7 +774,10 @@ def read_xstable_model(modelname, filename, etable=False):
     .. versionchanged:: 4.16.0
        Parameters with negative DELTA values are now made frozen, to
        match XSPEC. Support for models which use the ESCALE keyword
-       has been added.
+       has been added.  The hard_max and hard_min values of the
+       redshift parameter (for those models that support it) can now
+       be changed. This should be done with care as it could cause
+       memory corruption or a crash.
 
     .. versionchanged:: 4.14.0
        The etable argument has been added to allow exponential table
@@ -1241,15 +1244,20 @@ class XSModel(RegriddableModel1D, metaclass=ModelMeta):
 class XSTableModel(XSModel):
     """Interface to XSPEC table models.
 
-    XSPEC supports loading in user-supplied data files for use
-    as a table model [1]_. This class provides a low-level
-    way to access this functionality. A simpler interface is provided
-    by `read_xstable_model` and `sherpa.astro.ui.load_xstable_model`.
+    XSPEC supports loading in user-supplied data files for use as a
+    `table model
+    <https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSappendixLocal.html>`_. This
+    class provides a low-level way to access this functionality. A
+    simpler interface is provided by `read_xstable_model` and
+    `sherpa.astro.ui.load_xstable_model`.
 
     .. versionchanged:: 4.16.0
-       Parameters with negative DELTA values are now made frozen, to
-       match XSPEC. Support for models which use the ESCALE keyword
-       has been added.
+       The hard_max and hard_min values of the redshift parameter (for
+       those models that support it) can now be changed. This should
+       be done with care as it could cause memory corruption or a
+       crash.  Parameters with negative DELTA values are now made
+       frozen, to match XSPEC. Support for models which use the ESCALE
+       keyword has been added.
 
     .. versionchanged:: 4.14.0
        The etable argument has been added to allow exponential table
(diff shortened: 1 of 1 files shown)


DougBurke:
@JohannesBuchner I don't know if you have time / energy to test out this fix (or just create a throw-away CIAO 4.15 conda environment and make the `XSBaseParameter` to `XSParameter` change in it's `sherpa/astro/xspec/__init__.py` file.

anetasie (commented):
It looks like a good way to deal with the issue.

hamogu (approved):

JohannesBuchner:
> @JohannesBuchner I don't know if you have time / energy to test out this fix (or just create a throw-away CIAO 4.15 conda environment and make the `XSBaseParameter` to `XSParameter` change in it's `sherpa/astro/xspec/__init__.py` file.

Indeed, unfortunately I am not finding the time to look into this at the moment, sorry. But I just want to say thank you for looking into this!

anetasie (approved):

PR #1848, opened 2023-08-15
Tests: add regrid corner case tests

DougBurke:
# Summary

Add tests for corner-cases for the regrid method. Many of these tests are currently regression tests, and check that the code behaviour has not changed, rather than that it works correctly. There is no functional change to the code.

# Details

These tests have been lying around for a while but I am not convinced the PRs they come from are going to make it into 4.16, so I want the actual tests in so I don't keep on re-inventing them, and also so we can easily see when the behaviour changes (hopefully fixed).

- adds explicit tests of #1802 
- taken from 
  - #1179 
  - #1803 
 although much of this code has been re-worked from previous un-merged PRs

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_regrid.py
+++ b/sherpa/astro/xspec/tests/test_xspec_regrid.py
@@ -1,0 +1,639 @@
+#
+#  Copyright (C) 2020, 2023
+#  Smithsonian Astrophysical Observatory
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


sherpa/astro/xspec/tests/test_xspec_regrid.py
```diff
@@ -0,0 +1,639 @@
...
+    mdl2 = xsmodel(name2)
+    cmdl = conv(mdl1 * mdl2)
+
+    rmdl = cmdl.regrid(ebase[:-1], ebase[1:])
+
+    exp = cmdl(eg1, eg2)
+    assert (exp > 0).all()
+
+    y = rmdl(eg1, eg2)
+    assert y == pytest.approx(exp, rel=0.35)
```
hamogu, line 639: It's so large that I wonder if something is wrong here... but as long as essentially everything is expected to fail anyway, I won't worry.

hamogu (approved):

sherpa/astro/xspec/tests/test_xspec_regrid.py
DougBurke: I am treating this as a regression test at the moment, but it is one of the things I have been trying to get to the bottom of with my work on the regrid code, but have not been able to come to a satisfactory story just yet.

PR #1851, opened 2023-08-16
XSPEC: table model improvements

### LINKED ISSUE #1850: XSPEC table model: not recognizing a frozen parameter
I have created an XSPEC table model with a negative DELTA value, which means it should be frozen, and is in XSPEC:

```
% xspec

		XSPEC version: 12.13.1
	Build Date/Time: Tue Aug  1 12:07:46 2023

XSPEC12>mo atable{abslp.tmod}

              1         -1(      0.01)          1          1          1          1
1:galabs:nop>
2:galabs:norm>

Model atable{abslp.tmod}<1> Source No.: 1   Active/Off
 par  comp
   1    1   galabs     nop                 1.00000      frozen
   2    1   galabs     norm                1.00000      +/-  0.0          

```

Note that 'nop' is frozen.

It isn't in CIAO 4.15:

```
sherpa-4.15.0> load_xstable_model("mymodel", "abslp.tmod")

sherpa-4.15.0> print(mymodel)
xstablemodel.mymodel
   mymodel.nop  thawed            1            1            1           
   mymodel.norm thawed            1            0        1e+24           

```

and

```
% dmlist abslp.tmod"[cols name,delta]" data,clean 
#  NAME           DELTA
 nop                            -1.0
```

DougBurke:
# Summary

Address issues with the XSPEC table model support: models that set an Escale parameter (feature added December 2020) are now supported (previously they could cause a segmentation fault) and ensure that parameters with a negative delta value are marked as frozen. Fix #1850 and #1852.

# Details

With the merging of https://github.com/sherpa/sherpa-test-data/pull/28 I have re-worked this so that we can add tests that show the problem and then fix them.

We had not been marking XSPEC table model parameters as frozen that should be, using the standard "delta < 0" check. It's an easy fix, but we currently don't have any test data (although see https://github.com/sherpa/sherpa-test-data/pull/28). Fix #1850

We had not been looking for models which have the ESCALE keyword (in the PRIMARY block) set. This was added December 2020, and, as noted in #1852 it currently causes Sherpa to go boom. This is a an easy fix (we can follow the handling of the REDSHIFT parameter). I've checked and if noth redshift and escale are set then we have redshift then escale in the parameters. Unfortunately the results appear to be garbage when the model is evaluated, but that's not our issue. I have asked Keith about this combination but don't expect a response soon.

The Dec 2020 change also added support for multiple spectra per grid point. This requires XFLT support (these keywords are used to associate data files to the spectrum to use) and we currently do not support this (we don't have a single issue for this, but 

- #544
- #1225 

are related), so we instead explicitly error out for such a file (identified as having NXFLTEXP > 1 in the PRIMARY header).

As the table model support depends somewhat on the XSPEC version in play (for example, there were changes made in 12.13.1 to address a problem we reported for users of the model library) then there may be some clean up to tweak the tests for different XSPEC versions. 

I actually don't check out all the models added in https://github.com/sherpa/sherpa-test-data/pull/28 because I don't know what we expect when both redshift and escale are set in the header (the tests with XSPEC have been less-than spectacular) so I have held off testing this combination until we find there's a use for it.

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1686,7 +1686,7 @@ def test_table_mod_negative_delta_1850(addmodel, redshift, escale, make_data_pat
     tbl = xspec.read_xstable_model('tbl', infile)
 
     # issue #1850; is the parameter recognized as frozen
-    assert not tbl.lscale.frozen
+    assert tbl.lscale.frozen
 
     parnames = [p.name for p in tbl.pars]
     assert parnames[0] == "lscale"
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1698,12 +1698,9 @@ def test_table_mod_negative_delta_1850(addmodel, redshift, escale, make_data_pat
         idx += 1
 
     if escale:
-        # At the moment ESCALE is not recognized
-        #
-        # assert parnames[idx] == "Escale"
-        # assert tbl.escale.frozen
-        # idx += 1
-        assert "Escale" not in parnames
+        assert parnames[idx] == "Escale"
+        assert tbl.escale.frozen
+        idx += 1
 
     if addmodel:
         assert parnames[idx] == "norm"
@@ -1778,7 +1775,6 @@ def test_table_mod_add_redshift(make_data_path):
     assert tbl(elo, ehi) / de == pytest.approx(expected, rel=2e-6)
 
 
-@pytest.mark.xfail  # issue #1852
 @requires_xspec
 @requires_data
 @requires_fits
@@ -1882,7 +1878,6 @@ def test_table_mod_mul_redshift(make_data_path):
     assert tbl(elo, ehi) == pytest.approx(expected, rel=2e-6)
 
 
-@pytest.mark.xfail  # issue #1852
 @requires_xspec
 @requires_data
 @requires_fits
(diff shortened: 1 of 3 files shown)


sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -1661,3 +1661,256 @@ def test_xspec_model_kwarg_cached_not_needed():
...
+    idx = 1
+    if redshift:
+        assert parnames[idx] == "redshift"
+        assert tbl.redshift.frozen
+        idx += 1
+
+    if escale:
+        assert parnames[idx] == "Escale"
+        assert tbl.escale.frozen
+        idx += 1
```
hamogu, line 1703: Should we also check (here or in a different location) that 
```
if not escale:
    assert 'Escale' not in parnames
```
or something like that?

I don't know the XSPEC table models as well. Of course, such an error should not happen, but since this PR special-cases escale, redshift, addmodel it would probably not hurt, right?

DougBurke:
We already test this by virtue of

A) checking each member of parnames
B) checking the length of parnames (to make sure we have checked each
member).

Although I'm on holiday today so haven't checked my logic .

sherpa/astro/xspec/tests/test_xspec_unit.py
DougBurke: So, as mentioned in the main text, we already test this case because we check each element of parnames in this loop:

- parnames[0] is lscale
- parnames[idx] is redshift and or Escale if expected, with idx increasing each time when relevant
- parnames[idx] is norm if an addmodel case, with idx increasing
- len([parnames] == idx

hamogu (approved):

Issue #956, opened 2020-10-04
group_xxx should use the mask to set tabstops by default


DougBurke:
At the moment we have to say (object level)

    pha.group_counts(20, tabStops=~pha.mask)

and (ui level)

    tabs = ~ui.get_data().mask  # may be a more direct call
    ui.group_counts(20, tabStops=mask)

if we want to group only where the data is filtered - which I believe should be the default. The `group_xxx` methods can do this automatically, so I propose

- change the tabStops argument to default to `"filter"` (or some such term), which means to use ~pha.mask

Then users will get the new behavior, and users who have set tabStops to ~pha.mask see no difference. It is only those people who don't want the tabstops set which will need to change, by adding tabStops=None.

Not 100% happy about the "trinary" nature of the comparison: string, sequence or NumPy array, or None, but it's hard to make this backwards compatible.

I'm also embarassed it's taken me this long to think about and suggest this.

DougBurke:
Note for users: you should probably use `pha.get_mask()` rather than `pha.mask` since the forner is guaranteed to give nchannel values whereas the latter gives a "grouped" mask, which is not useful here (although I guess the `guess_xxx` routines could expand out a grouped mask if sent one).

DougBurke:
In #1855 I've actually dropped the "trinary" option:

- tabStops=None (defautl) means use ~self.get_mask()
- tabStops=[0] * nchannel means "no tabstops" (so behaves like 4.15.1 and earlier)
- tabStops=<some array> means use this array as the tabStops

I decided that it likely wasn't worth making it easy to get the pre-4.16 behaviour, but let's see what other people think.

PR #1855, opened 2023-08-21
group_counts et al now default to using the existing mask as the tabstops, unless explicitly given

### LINKED ISSUE #417: improve UI for grouping after filtering
If a user filters the data and the groups it, e.g. according to the number of counts, the grouping is performed by ignoring the filter first, applying the grouping, and then applying the filter again. There is a way to apply the grouping only inside the filtered range(s), but this is hardly user friendly, i.e.:

``` python
from sherpa.astro import ui
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [0, 0, 0, 2, 1, 1, 0, 0, 0, 0]
ui.load_arrays(1, x, y, DataPHA)
ui.notice(3.5, 6.5)
tab_stops = ~ui.get_data().mask
ui.group_counts(3, tabStops=tab_stops)
np.testing.assert_array_equal([3, 1], ui.get_data().get_dep(filter=True))
```

It would be useful to provide an option to the grouping functions to do this seamlessly, e.g.:

``` python
from sherpa.astro import ui
x = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
y = [0, 0, 0, 2, 1, 1, 0, 0, 0, 0]
ui.load_arrays(1, x, y, DataPHA)
ui.notice(3.5, 6.5)
ui.group_counts(3, filter=True)
np.testing.assert_array_equal([3, 1], ui.get_data().get_dep(filter=True))
```

Note: we are closing #149 and replacing it with related issues uncovered while working at `specfit`.

### LINKED ISSUE #1081: group_xxx, tabStops, and mask handling

DougBurke:
# Summary

When calling group_counts, group_width, ..., use the current mask as the tabStops argument by default (actually, the inverse of the mask), that is, when the tabStops argument is set to None (the default). This means that the grouping is only applied to the selected channels, and means that the filter range is less likely to change when filtered data is grouped.

This is a change in behaviour. To get the previous behaviour users need to set tabStops to an array of zeros (length being the number of channels): e.g. for Chandra ACIS imaging-mode data, group_counts(20, tabStops=[0] * 1024) or group_counts(20, tabStops=np.zeros(1024)).

Fix #417 #956 #1081

# Details

[stupid github interface. Fix #1081 ]

There are some minor clean up changes to allow 

- tabStops to be sent as a list rather than needing to be a ndarray, so that users can say `tabStops=[0] * 1024`
- if tabStops is sent a mask that matches the grouped size we allow this, automatically expanding this to the per-channel array in essentially the same way that the DataPHA get_mask routine does [*]; this is to make things easier for users as it has been an issue in my earlier experiments in getting tabStops to work (it's also the case that with follow-up changes it's unlikely to be useful for most users but it's not a large change)

[*] there are some possible issues with quality filtering here, but there's quality-filtering issues elsewhere too

The main change is so that if tabStops is None (the default) then we pick up the existing mask (inverted) and use that, so whereas before you would have to say

    group_counts(20, tabStops=~get_data().get_mask())

you can now say just

    group_counts(20)

This is a **breaking** change. If you want to get the old behavior you have to say either of

    group_counts(20, tabStops=[0] * 1024)
    group_counts(20, tabStops=np.zeros(1024))

changing 1024 to the number of channels in the dataset.

I spent some time trying to work out how to make this "backwards compatible", but I have come to the conclusion that it's actually better that we make this a breaking change - i.e. default to now only grouping within the noticed data range. However, this is something other people may think differently!

There is an argument to be said that we could have some "symbol" for `tabStops` which means "ignore the filter", so you don't have to say `tabStops=[0] * n` and can just say `tabStops="unfiltered"` (*as an example*).

---

# Example

In CIAO 4.15 we have

```
sherpa-4.15.0> load_pha("3c273.pi")
read ARF file 3c273.arf
read RMF file 3c273.rmf

sherpa-4.15.0> ungroup()

sherpa-4.15.0> ignore(hi=0.5)
dataset 1: 0.00146:14.9504 -> 0.511:14.9504 Energy (keV)

sherpa-4.15.0> ignore(lo=7)
dataset 1: 0.511:14.9504 -> 0.511:6.9934 Energy (keV)

sherpa-4.15.0> get_filter()
'0.510999977589:6.993400096893'

sherpa-4.15.0> group_counts(20)

sherpa-4.15.0> get_filter()
'0.437999993563:13.461199760437'
```

So, after calling `group_counts` the filter has increased from 0.51:6.99 to 0.44:13.46 keV.

We can do the grouping only within the filtered range, but only by manually specifying the `tabStops` argument (in the correct manner):

```
sherpa-4.15.0> ungroup()
          ...: ignore(hi=0.5)
          ...: ignore(lo=7)
dataset 1: 0.438:13.4612 -> 0.511:13.4612 Energy (keV)
dataset 1: 0.511:13.4612 -> 0.511:6.9934 Energy (keV)

sherpa-4.15.0> get_filter()
'0.510999977589:6.993400096893'

sherpa-4.15.0> group_counts(20, tabStops=~get_data().get_mask())

sherpa-4.15.0> get_filter()
'0.510999977589:6.993400096893'
```

Note that the filter has not changed.

With this PR we get (you can see we get more information on the filter changes as we call the various group routines, due to #1637

```

In [2]: load_pha("3c273.pi")
read ARF file 3c273.arf
read RMF file 3c273.rmf

In [3]: ungroup()
dataset 1: 0.00146:14.9504 Energy (keV) (unchanged)

In [4]: ignore(hi=0.5)
dataset 1: 0.00146:14.9504 -> 0.511:14.9504 Energy (keV)

In [5]: ignore(lo=7)
dataset 1: 0.511:14.9504 -> 0.511:6.9934 Energy (keV)

In [6]: group_counts(20)
dataset 1: 0.511:6.9934 Energy (keV) (unchanged)
```

So, the `group_counts` call (and all the other `group_xxx` ones) has automatically set `tabStops` for us.

If we want to behave like the old version we have to manually clear the `tabStops`: we need to know the number of channels - here 1024 - and say

```
In [7]: ungroup()
dataset 1: 0.511:6.9934 Energy (keV) (unchanged)

In [8]: notice()
dataset 1: 0.511:6.9934 -> 0.00146:14.9504 Energy (keV)

In [9]: ignore(lo=7)
dataset 1: 0.00146:14.9504 -> 0.00146:6.9934 Energy (keV)

In [10]: ignore(hi=0.5)
dataset 1: 0.00146:6.9934 -> 0.511:6.9934 Energy (keV)

In [11]: group_counts(20, tabStops=[0] * 1024)
dataset 1: 0.511:6.9934 -> 0.438:13.4612 Energy (keV)
```

### DIFF
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -4381,3 +4381,35 @@ def test_img_checks_coord_no_transform(coord):
                        match="^data set 'ex' does not contain a .* coordinate system$"):
         DataIMG("ex", [1, 2, 1, 2], [1, 1, 2, 2], [1, 2, 3, 4],
                 coord=coord)
+
+
+@pytest.mark.parametrize("asarray", [True, pytest.param(False, marks=pytest.mark.xfail)])  # XFAIL: TypeError: grpBinWidth() Could not parse input arguments, ...
+def test_group_xxx_tabtops_not_ndarray(asarray):
+    """What happens if tabStops is not a ndarray?"""
+
+    pha = DataPHA("test", [1, 2, 3, 4, 5], [2, 3, 4, 5, 6])
+    tabstops = [1, 1, 0, 0, 1]
+    if asarray:
+        tabstops = np.asarray(tabstops)
+
+    # This should only group channels 3 and 4.
+    pha.group_width(2, tabStops=tabstops)
+
+    assert pha.get_y() == pytest.approx([2, 3, 4.5, 6])
+    assert pha.mask is True
+    assert pha.get_mask() is None
+
+
+@pytest.mark.parametrize("asarray", [True, pytest.param(False, marks=pytest.mark.xfail)])  # XFAIL: TypeError: grpBinWidth() Could not parse input arguments, ...
+@pytest.mark.parametrize("nelem", [4, 6])
+def test_group_xxx_tabtops_wrong_size(asarray, nelem):
+    """What happens if tabStops is not a ndarray?"""
+
+    pha = DataPHA("test", [1, 2, 3, 4, 5], [2, 3, 4, 5, 6])
+    tabstops = [0] * nelem
+    if asarray:
+        tabstops = np.asarray(tabstops)
+
+    emsg = r"^grpBinWidth\(\) The number of tab stops and number of channels specified in the argument list have different sizes$"
+    with pytest.raises(ValueError, match=emsg):
+        pha.group_width(2, tabStops=tabstops)
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -4415,7 +4415,6 @@ def test_group_xxx_tabtops_wrong_size(asarray, nelem):
         pha.group_width(2, tabStops=tabstops)
 
 
-@pytest.mark.xfail  # XFAIL: ValueError: grpNumCounts() The tabStops and countsArray have differing length
 def test_group_xxx_tabstops_already_grouped():
     """Check what happens if tabStops is sent ~pha.mask when already grouped."""
 
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -4383,7 +4383,7 @@ def test_img_checks_coord_no_transform(coord):
                 coord=coord)
 
 
-@pytest.mark.parametrize("asarray", [True, pytest.param(False, marks=pytest.mark.xfail)])  # XFAIL: TypeError: grpBinWidth() Could not parse input arguments, ...
+@pytest.mark.parametrize("asarray", [True, False])
 def test_group_xxx_tabtops_not_ndarray(asarray):
     """What happens if tabStops is not a ndarray?"""
 
@@ -4400,7 +4400,7 @@ def test_group_xxx_tabtops_not_ndarray(asarray):
     assert pha.get_mask() is None
 
 
-@pytest.mark.parametrize("asarray", [True, pytest.param(False, marks=pytest.mark.xfail)])  # XFAIL: TypeError: grpBinWidth() Could not parse input arguments, ...
+@pytest.mark.parametrize("asarray", [True, False])
 @pytest.mark.parametrize("nelem", [4, 6])
 def test_group_xxx_tabtops_wrong_size(asarray, nelem):
     """What happens if tabStops is not a ndarray?"""
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -469,25 +469,41 @@ def test_416_a():
 
     pha.notice(4.5, 6.5)
 
-    mask = [False, False, False, True, True, True, False, False, False, False]
-    assert pha.mask == pytest.approx(mask)
-
-    # With tabStops not set we use ~mask (as this is in channel space).
+    # There are two ways to get the mask:
+    #   - pha.mask returns the grouped mask (if the data is
+    #     grouped)
+    #   - pha.get_mask() always returns the ungrouped mask
     #
+    mask_ungrouped = [False] * 3 + [True] * 3 + [False] * 4
+    mask_grouped = [False] * 3 + [True] * 2 + [False] * 4
+
+    assert pha.mask == pytest.approx(mask_ungrouped)
+    assert pha.get_mask() == pytest.approx(mask_ungrouped)
+    assert pha.grouping is None
+    assert pha.quality is None
+
+    # The grouping is done only for the noticed data range.
     pha.group_counts(3)
 
-    # We have a simplified mask
-    mask = [False] * 3 + [True] * 2 + [False] * 4
-    assert pha.mask == pytest.approx(mask)
+    assert pha.mask == pytest.approx(mask_grouped)
+    assert pha.get_mask() == pytest.approx(mask_ungrouped)
 
-    # the "full" mask can be retrieved with get_mask
-    mask = [False] * 3 + [True] * 3 + [False] * 4
-    assert pha.get_mask() == pytest.approx(mask)
-
-    grouping = [0, 0, 0, 1, -1, 1, 0, 0, 0, 0]
+    # Check we get the expected grouping: the first 3 and last 4
+    # channels are excluded by the notice call above, so they are 0,
+    # which means we only have 3 channels with data.
+    #
+    grouping = [0] * 3 + [1, -1, 1] + [0] * 4
     assert pha.grouping == pytest.approx(grouping)
 
-    quality = [0, 0, 0, 0, 0, 2, 0, 0, 0, 0]
+    # As with grouoping, the first 3 and last 4 channels are not
+    # changed, so have a quality of 0. For the remaining three
+    # channels we know the first group is okay (this corresponds to
+    # [1, -1] from the grouping array above, correspondinf to counts
+    # [2, 1]) but the second group (the second [1] in grouping,
+    # corresponding to counts of [1]) does not meet the grouping
+    # criteria (it sums to 1!) and so has a quality of 2.
+    #
+    quality = [0] * 3 + [0, 0, 2] + [0] * 4
     assert pha.quality == pytest.approx(quality)
 
     dep = pha.get_dep(filter=True)
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -4413,3 +4413,28 @@ def test_group_xxx_tabtops_wrong_size(asarray, nelem):
     emsg = r"^grpBinWidth\(\) The number of tab stops and number of channels specified in the argument list have different sizes$"
     with pytest.raises(ValueError, match=emsg):
         pha.group_width(2, tabStops=tabstops)
+
+
+@pytest.mark.xfail  # XFAIL: ValueError: grpNumCounts() The tabStops and countsArray have differing length
+def test_group_xxx_tabstops_already_grouped():
+    """Check what happens if tabStops is sent ~pha.mask when already grouped."""
+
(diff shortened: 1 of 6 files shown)


DougBurke:
@anetasie and @hamogu - this is something I have wanted for quite a while - the `group_xxx` calls will not only group within the noticed data range. However, with the current approach it is a **breaking** change, in that users with 4.16 will get different results to 4.15. Fortunately the changes I added to report the filter when notice/ignore is called - #1562 in 4.15.0 - and group/... is called - #1637 in 4.15.1 - mean it is a bit more obvious to the user what is going on.

However, you may find that it is too invasive and we need another way to do this (i.e. do not do this by default) which is why it's still labelled a draft PR.

I do need to add some documentation, but your input is more-than-welcome here.

hamogu:
This is something that I've wanted for a while, too; probably since before I started contributing to sherpa development. In my opinion,  the old behavior is confusing and the new one much better.  Yes,  it's a breaking change,  but it will make it better.

I'm not making this as approved because I did not do a code review at this point.

DougBurke:
I think it's useful to just try it out as well, to make sure that it really does behave as you want @hamogu!

anetasie:
@DougBurke I have a perfect data set for the tabstops and will use it!

anetasie (approved):
Looks good.

DougBurke:
@anetasie - thanks for looking. One thing I wonder about is that if you want the "old" behaviur you have to explicitly set `tabStops` to an array of 0's - e.g. one of

```
group_counts(20, tabStops=[0] * 1024)
group_counts(20, tabStops=np.zeros(0, 1024))
```

This requires you to remember the number of channels and how to create an array of zeros. We could allow the user to set `tabStops` to a particular string - e.g.

```
group_counts(20, tabStops="nofilter")
```

Would this make sense to you? If so,what would be a good string to use (I don't want to use "none" as tabStops=None has a very-different behaviour).

anetasie:
@DougBurke this would be good for back compatibility. 

```
group_counts(20, tabStops="nofilter")
```

DougBurke:
For @anetasie - this "multi-modal" behaviour of keywords is generally frowned upon in the CS world, bit I think the UI layer has long-ago abandoned this for "ease of non-programmer use".

DougBurke:
@anetasie and @hamogu - I've updated this to

- add some documentation
- the ui layer lets you `group_counts(20, tabStops="nofilter")` to get the old behaviour (you can still say tabStops=[0] * nchan, for the appropriate nchan value, but this is to make it easier for users)
  - this is only in the UI layer; the DataPHA version does not support this  option (as it's an ease-of-use change we don't really want in the lower-level code)

Other than that, and all the typos I've added, this is ready for final review.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -472,20 +472,22 @@ def test_416_a():
     mask = [False, False, False, True, True, True, False, False, False, False]
     assert pha.mask == pytest.approx(mask)
 
+    # With tabStops not set we use ~mask (as this is in channel space).
+    #
     pha.group_counts(3)
 
     # We have a simplified mask
-    mask = [True, True]
+    mask = [False] * 3 + [True] * 2 + [False] * 4
```
hamogu, line 480: This is exactly the mask defined in line 472. No need to put it in here again. Same is true for line 482/484 below.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -472,20 +472,22 @@ def test_416_a():
     mask = [False, False, False, True, True, True, False, False, False, False]
     assert pha.mask == pytest.approx(mask)
 
+    # With tabStops not set we use ~mask (as this is in channel space).
+    #
     pha.group_counts(3)
 
     # We have a simplified mask
```
hamogu, line 479: We don't any longer...

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -472,20 +472,22 @@ def test_416_a():
...
-    mask = [True] * 10
+    mask = [False] * 3 + [True] * 3 + [False] * 4
     assert pha.get_mask() == pytest.approx(mask)
 
-    grouping = [1, -1, -1, -1, -1,  1, -1, -1, -1, -1.]
+    grouping = [0, 0, 0, 1, -1, 1, 0, 0, 0, 0]
     assert pha.grouping == pytest.approx(grouping)
 
-    quality = [0, 0, 0, 0, 0, 2, 2, 2, 2, 2]
+    quality = [0, 0, 0, 0, 0, 2, 0, 0, 0, 0]
```
hamogu, line 490: Wait, why has the quality changed? Why is the one in the middle quality 2? I don't see any quality array set anywhere, so I'm confused.

hamogu (changes requested):

sherpa/astro/tests/test_astro_data2.py
DougBurke: So, this is a hold over from when the mask was different. I'll clean up.

However I disagree with line 484 as that line is different (there's an extra bin).

DougBurke: Actually, I mis-spole. This particular line *is* different, since the mask is now grouped. Hopefully the new version of the test will make it clearer.

sherpa/astro/tests/test_astro_data2.py
DougBurke: So, the group call sets both the grouping and quality arrays. I've added text to the test to explain this, but we can view quality here as

  [3] * 0 + [0, 0, 2] + [0] * 4

The first 3 and last 4 channels are not grouped (since we've excluded them by the filter) and so are set to 0. The only channels we care about are the fourth to sixth channels, which have counts of [2, 1, 1]. The grouping groups them into 2 bins, so we have grouping of [1, -1, 1] to indicate this. The first group (of 2, 1) meets the group criteria (of >= 3) and so has a quality of 0. The last channel has a count of 1, which is < 3, and so has a quality of 3.

When we didn't include the tabstops we carried the 2's to the end of the array because these channels had counts of 0 so the group never hit the >= 3 criteria.

hamogu (approved):

PR #1858, opened 2023-08-23
Correctly handle XSPEC table models with both ESCALE and Z parameters

### LINKED ISSUE #1857: XSPEC table model support with both ESCALE and REDSHIFT parameters set
I have reported to  Goddard that the current implementation of XSPEC table models in XSPEC (12.13.1, no patch) is flawed when a model has both REDSHIFT and ESCALE keywords set (which is an odd thing to so, so it's unlikely anyone has done this), as there is a mis-match in the parameter order (shown below). It has been confirmed as a bug.

The easy fix for us - now that #1851 has added support for the ESCALE parameter - is to switch the ordering of the redshift and escale parameters, but I'm waiting to hear from Goddard if that's what then plan to do for XSPEC.

```
% xspec

		XSPEC version: 12.13.1
	Build Date/Time: Tue Aug  1 12:07:46 2023

XSPEC12>dummyrsp 0.4 0.5 10 linear
XSPEC12>mo atable{kze.tmod}

              1         -1(      0.01)          1          1          1          1
1:fze:nop>
              0     -0.001(     0.001)          0          0          5          5
2:fze:z>
              1     -0.001(      0.01)      1e-05      1e-05        100        100
3:fze:Escale>
4:fze:norm>

Model atable{kze.tmod}<1> Source No.: 1   Active/Off
 par  comp
   1    1   fze        nop                 1.00000      frozen
   2    1   fze        z                   0.0          frozen
   3    1   fze        Escale     keV      1.00000      frozen
   4    1   fze        norm                1.00000      +/-  0.0          

XSPEC12>iplot model
 ERROR--YMIN = YMAX in window   1
 ERROR--YMIN = YMAX in window   1
PLT> wdata
READ SERR 1
!
0.405000001 4.99999989E-3 0
0.414999992 4.99999989E-3 0
0.425000012 4.99999989E-3 0
0.435000002 4.99999989E-3 Infinity
0.444999993 4.99999989E-3 Infinity
0.455000013 4.99999989E-3 0
0.465000004 4.99999989E-3 0
0.474999994 4.99999989E-3 0
0.485000014 4.99999989E-3 0
0.495000005 4.99999989E-3 0
```

DougBurke:
# Summary

Update XSPEC table model support for models with both ESCALE and REDSHIFT keywords set to work correctly. Fix #1857 

# Details

I reported to HEASARC that models with both ESCALE and REDHISFT keywords set were producing garbage and they confirmed the bug. They are fixing this in XSPEC 12.13.1a (but it hasn't been released at the time of writing of this PR).

The fix is easy - just switch the ordering of the escale and redshift parameters. I have added tests to show this works (and the results of evaluating the model are as expected).

Note that we have only just added support for the ESCALE keyword setting in XSPEC table models, so we don't need to add documentation about this fix.

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -1359,22 +1359,25 @@ class XSTableModel(XSModel):
         self.addmodel = addmodel
         self.etable = etable
 
-        # Order appears to be
-        #   - z
+        # Order should be (z and escale were incorrectly mixed up
+        # in XSPEC 12.13.1 and earlier):
+        #
         #   - Escale
+        #   - z
         #   - norm
+        #
         # (for those models that support the relevant value)
         #
-        if addredshift:
-            self.redshift = XSBaseParameter(name, 'redshift', 0., 0., 5.,
-                                            0.0, 5, frozen=True)
-            pars.append(self.redshift)
-
         if addescale:
             # Should this just use Parameter?
             self.Escale = XSParameter(name, 'Escale', 1, 0, 1e20,
                                       0, 1e24, frozen=True, units="keV")
             pars.append(self.Escale)
+
+        if addredshift:
+            self.redshift = XSBaseParameter(name, 'redshift', 0., 0., 5.,
+                                            0.0, 5, frozen=True)
+            pars.append(self.redshift)
 
         if addmodel:
             # Normalization parameters are not true XSPEC parameters and
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -1727,6 +1727,23 @@ def test_table_mod_add(make_data_path):
     assert tbl(elo, ehi) == pytest.approx(expected)
 
 
+# Values taken from XSPEC 12.13.1 with
+#
+#    dummyrsp 0.2 2.6 24 linear
+#    mo atable{smod1xx.tmod}
+#    newpar xxxxxx
+#    iplot model
+#    wdata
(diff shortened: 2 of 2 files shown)


sherpa/astro/xspec/__init__.py
```diff
@@ -1359,23 +1359,26 @@ def __init__(self, filename, name='xstbl', parnames=(),
...
         #
-        if addredshift:
-            self.redshift = XSBaseParameter(name, 'redshift', 0., 0., 5.,
-                                            0.0, 5, frozen=True)
-            pars.append(self.redshift)
-
         if addescale:
             # Should this just use Parameter?
             self.Escale = XSParameter(name, 'Escale', 1, 0, 1e20,
                                       0, 1e24, frozen=True, units="keV")
```
DougBurke, line 1374: From looking at the XSPEC 12.13.1a patch it looks like the units of Escale have been removed (which makes sense)

```suggestion
                                      0, 1e24, frozen=True)
```

DougBurke: Need to also check the range as the code looks different to what I saw in XSPEC itself...

DougBurke, line 1374: Actually, it should be

```suggestion
            self.Escale = XSParameter(name, 'Escale', 1, 1e-5, 100,
                                      1e-5, 100, frozen=True)
```

hamogu (approved):
This looks good to me and it woudl make sense to bring that into 4.16, so that we don't release a broken version of ESCALE first and then have to worry about this as an API change later.

PR #1861, opened 2023-08-25
Allow XSPEC table models to be created and written out

### LINKED ISSUE #1859: can we create XSPEC table models
It would make sense for Sherpa users to be able to create XSPEC table models (I have recently helped out a user that needed thos feature). There would need to be additions to `sherpa.astro.io` and the backends.

DougBurke:
# Summary

Provide experimental support for creating XSPEC table models (additive or multiplicative). The output routines can be found in the sherpa.astro.ui.xstable module. Note that XSPEC support is not required to write out an XSPEC table (but it is to be able to read it back in and use it). Fix #1859

# Details

This is built on #1836. As mentioned below, some of the additions here could be integrated with the changes in #1836 but I think that is too large a project at this time.

This was created because of discussions I had with a local user doing cross-telescope calibration work e.g. see https://github.com/sherpa/sherpa/discussions/1849 

The reason the code is labelled experimental is because I am not 100% convinced that I have come up with a good way of sending in the data: when writing out the data there is some repeated information which we don't need to make the user repeat, and hence ensure that the table is consistent, but there's more than one way to avoid this repetition and so we may decide to tweak the API. There are also other parts of the API that may be a bit lacking in polish. I don't expect the changes to be large, but I want to leave there room for the possibility of changes.

As this is labelled experimental I have not added it as a routine in sherpa.astro.io. Instead users will need to use sherpa.astro.io.xstable to import the routines. This functionality is specialized enough that I think that having it in a separate module is good. Thinking it through, I don't think it makes sense to add to sherpa.astro.io directly, and I really don't want to add it to sherpa.astro.ui since (again) it is very specialized.

Some of the internal abstractions I use could be used with the existing I/O calls (ie used as an interchange between sherpa data objects and the crates/pyfits backends), but I have not tried this here as it is too large a project for me at this time.

I added typing information to sherpa.astro.io.xstsable - and mypy will parse it successfully. It did help me find some logical errors, but I have used Any for handling of ndarray-like data (since the last I looked there was no nice simple way for me to say iterable or 1D ndarray, and also some of the typing assumes lists when really it should be more generic (e.g. sequence). 

# Notes

Writing out the data isn't too hard, it's just a bit clunky as there's four HDUs and some columns are "interesting". Fortunately we can use the XSPEC read_xstable_model to verify that we've written it out correctly (although I will caution you that the error messages XSPEC creates for mal-formed files is not very helpful; the default appears to be to segfault).

# Example

Let's write out an additive table model that just varies the location of a gaussian model (this is not a sensible use but it does show off some features)

```
import numpy as np

from sherpa.astro.io import xstable
from sherpa.astro.xspec impost XSgaussian

egrid = np.arange(0.1, 2, 0.01)
elo = egrid[:-1]
ehi = egrid[1:]

mdl = XSgaussian()

# what is the line postion we can interpolate over
linepos = np.arange(0, 2.5, 0.1)

# evalute the model at these parameter values
models = []
for lp in linepos:
    mdl.linee = lp
    models.append(mdl(elo, ehi))

param = xstable.IntParam("pos", 1, 0.01, 0, linepos[0], linepos[-1], values=linepos)

hdus = xstable.make_xspec_table_model("gaussy", elo, ehi, params=[param], spectra=models)
xstable.write_xpsec_table_model("wat.tmod", hdus, clobber=True)
```

This creates a file called `wat.tmod`:

```
% dmlist wat.tmod blocks
 
Dataset: wat.tmod
 
Block    2: PARAMETERS                     Table        10 cols x 1        rows
Block    3: ENERGIES                       Table         2 cols x 189      rows
Block    4: SPECTRA                        Table         2 cols x 25       rows
% dmlist wat.tmod["primary]" header,clean
 --  COMMENT                FITS (Flexible Image Transport System) format is defined in 'Astronomy
 --  COMMENT                and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H
HDUCLASS             OGIP                           format conforms to OGIP standard
HDUCLAS1             XSPEC TABLE MODEL              model spectra for XSPEC
HDUVERS              1.2.0                          version of format
MODLNAME             gaussy                         Model name
MODLUNIT             photons/cm^2/s                 Model units
ADDMODEL             TRUE                           If T then this is an additive table model
REDSHIFT             FALSE                          If T then redshift will be a parameter
ESCALE               FALSE                          If T then escale will be a parameter
LOELIMIT                                0           Model value for energies below ENERGIES block
HIELIMIT                                0           Model value for energies above ENERGIES block
CREATOR              sherpa 4.15.1+213.ge5b465f0.dirty Code that created the file
DATE                 2023-08-25T17:04:16            Date file created
% dmlist wat.tmod data,clean
#  NAME           METHOD     INITIAL              DELTA                MINIMUM              BOTTOM               TOP                  MAXIMUM              NUMBVALS   VALUE[25]
 pos                     0                  1.0     0.00999999977648                    0                    0         2.4000000954         2.4000000954         25                                  0     0.10000000149012     0.20000000298023     0.30000001192093     0.40000000596046     0.50     0.60000002384186     0.69999998807907     0.80000001192093     0.89999997615814         1.0         1.1000000238         1.2000000477         1.2999999523         1.3999999762         1.50         1.6000000238         1.7000000477         1.7999999523         1.8999999762         2.0         2.0999999046         2.2000000477         2.2999999523         2.4000000954
```

This can then be used in Sherpa

```
sherpa In [1]: load_xstable_model("boom", "wat.tmod")

sherpa In [2]: print(boom)
xstablemodel.boom
   boom.pos     thawed            1            0          2.4           
   boom.norm    thawed            1            0        1e+24           

sherpa In [3]: egrid = np.arange(0.05, 3, 0.1)

sherpa In [4]: elo = egrid[:-1]

sherpa In [5]: ehi = egrid[1:]

sherpa In [6]: emid = (elo + ehi) / 2

sherpa In [7]: for p in [0.1, 0.5, 0.8, 1.4, 1.8, 2.2]:
          ...:     boom.pos = p
          ...:     plt.plot(emid, boom(elo, ehi), alpha=0.5, label=str(p))
          ...: 

sherpa In [8]: plt.legend()
Out[8]: <matplotlib.legend.Legend at 0x7f80fe9dc430>

sherpa In [9]: for p in [0.1, 0.5, 0.8, 1.4, 1.8, 2.2]:
          ...:     plt.axvline(p, c='k', alpha=0.5)
          ...: 
```

plot

Now, the line positions look "interesting" but that's for those whose centres lie outside the defined grid (0 to 2), so it's not surprising there's no data there.

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -68,7 +68,7 @@ __all__ = ('get_table_data', 'get_header_data', 'get_image_data',
            'get_column_data', 'get_ascii_data',
            'get_arf_data', 'get_rmf_data', 'get_pha_data',
            'set_table_data', 'set_image_data', 'set_pha_data',
-           'set_arf_data', 'set_rmf_data')
+           'set_arf_data', 'set_rmf_data', 'set_hdus')
 
 
 def _has_hdu(hdulist, name):
@@ -1465,3 +1465,100 @@ def set_arrays(filename, args, fields=None, ascii=True, clobber=False):
     hdu = fits.table_to_hdu(tbl)
     hdu.name = 'TABLE'
     hdu.writeto(filename, overwrite=True)
+
+
+def _add_header(hdu, header):
+    """Add the header items to the HDU.
+
+    Parameters
+    ----------
+    hdu : astropy HDU
+    header : list of sherpa.astro.io.xstable.HeaderItem
+    """
+
+    for hdr in header:
+        card = [hdr.name, hdr.value]
+        if hdr.desc is not None or hdr.unit is not None:
+            comment = "" if hdr.unit is None else f"[{hdr.unit}] "
+            comment += "" if hdr.desc is None else hdr.desc
+            card.append(comment)
+
+        hdu.header.append(tuple(card))
+
+
+def _create_primary_hdu(hdu):
+    """Create a PRIMARY HDU.
+
+    Parameters
+    ----------
+    hdu : sherpa.astro.io.xstable.TableHDU
+       Any data is ignored.
+
+    Returns
+    -------
+    out : astropy.io.fits.PrimaryHDU
+
+    """
+
+    out = fits.PrimaryHDU()
+    _add_header(out, hdu.header)
+    return out
+
+
+def _create_table_hdu(hdu):
+    """Create a Table HDU.
+
+    Parameters
+    ----------
+    hdu : sherpa.astro.io.xstable.TableHDU
+
+    Returns
+    -------
+    out : astropy.io.fits.BinTableHDU
+    """
+
+    if hdu.data is None:
+        raise ValueError("No column data to write out")
+
+    # First create a Table which handles the FITS column settings
+    # correctly.
+    #
+    store = []
+    colnames = []
+    for col in hdu.data:
+        colnames.append(col.name)
+        store.append(col.values)
+
+    out = fits.table_to_hdu(Table(names=colnames, data=store))
+    out.name = hdu.name
+    _add_header(out, hdu.header)
+
+    # Add any column metadata
+    #
+    for idx, col in enumerate(hdu.data, 1):
+        if col.unit is not None:
+            out.header.append((f"TUNIT{idx}", col.unit))
+
+        if col.desc is None:
+            continue
+
+        key = f"TTYPE{idx}"
+        out.header[key] = (col.name, col.desc)
+
+    return out
+
+
+def set_hdus(filename, hdulist, clobber=False):
+    """Write out multiple HDUS to a single file.
+
+    At present we are restricted to tables only.
+    """
+
+    check_clobber(filename, clobber)
+
+    out = fits.HDUList()
+    out.append(_create_primary_hdu(hdulist[0]))
+    for hdu in hdulist[1:]:
+        out.append(_create_table_hdu(hdu))
+
+    out.writeto(filename, overwrite=True)
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -1480,7 +1480,8 @@ def _add_header(hdu, header):
         card = [hdr.name, hdr.value]
         if hdr.desc is not None or hdr.unit is not None:
             comment = "" if hdr.unit is None else f"[{hdr.unit}] "
-            comment += "" if hdr.desc is None else hdr.desc
+            if hdr.desc is not None:
+                comment += hdr.desc
             card.append(comment)
 
         hdu.header.append(tuple(card))
--- a/sherpa/astro/io/xstable.py
+++ b/sherpa/astro/io/xstable.py
@@ -85,8 +85,8 @@ used by Sherpa to decide if the parameter is frozen (value is
 negative) or not, but is used by the optimiser in XSPEC. The values
 field is set to those used to generate the spectral models.
 
->>> param = xstable.IntParam("pos", 1, 0.01, minval, maxval,
-...                          values=linepos)
+>>> param = xstable.Param("pos", 1, 0.01, minval, maxval,
+...                       values=linepos)
 
 With this we can create the necessary information to create the
 table model (`make_xstable_model`) and then write it out
@@ -112,8 +112,8 @@ xstablemodel.mygauss
 Notes
 -----
 
-XSPEC models can be created without XSPEC support in Sherpa, but it is
-needed to read the files in.
+XSPEC models can be created without XSPEC support in Sherpa, but XSPEC
+support is needed to read the files in.
 
 For additive models it is assumed that the model values - that is,
 each bin - have units of photon/cm^2/s. This is easy to accidentally
@@ -133,7 +133,7 @@ import numpy as np
 import sherpa
 
 
-__all__ = ("IntParam", "Param", "make_xstable_model",
+__all__ = ("BaseParam", "Param", "make_xstable_model",
            "write_xstable_model")
 
 
@@ -141,7 +141,7 @@ __all__ = ("IntParam", "Param", "make_xstable_model",
 #
 
 @dataclass
-class Param:
+class BaseParam:
     """Represent a parameter.
 
     The absolute DELTA value is used by XSPEC but ignored by Sherpa.
@@ -150,14 +150,17 @@ class Param:
     default.
 
     """
+
     name: str
     """The parameter name.
 
     This must be unique and ideally a valid attribute name in Python,
     although the latter is not enforced.
     """
+
     initial: float
     """The default value for the parameter."""
+
     delta: float
     """The delta value for fitting.
 
@@ -165,12 +168,16 @@ class Param:
     does not use this value (other than checking the sign) but XSPEC
     does.
     """
+
     hardmin: float
     """The minimum value for the parameter."""
+
     hardmax: float
     """The maximum value for the parameter."""
+
     softmin: Optional[float] = None
     """The "soft" minimum. Defaults to hardmin if not given.."""
+
     softmax: Optional[float] = None
     """The "soft" maximum. Defaults to hardmax if not given."""
 
@@ -209,7 +216,7 @@ class Param:
 
 
 @dataclass
-class IntParam(Param):
+class Param(BaseParam):
     """Represent an interpolated parameter."""
 
     values: list[float] = field(default_factory=list)
@@ -218,6 +225,7 @@ class IntParam(Param):
     Is is required that they range from hardmin to hardmax and are in
     monotonically increasing order (so it can not be an empty list).
     """
+
     loginterp: bool = False
     """Are the values logarithmically interpolated (True) or linearly (False).
 
@@ -365,8 +373,8 @@ def xstable_primary(name: str,
     return TableHDU(name="PRIMARY", header=header)
 
 
-def xstable_parameters(params: list[IntParam],
-                       addparams: Optional[list[Param]]) -> TableHDU:
+def xstable_parameters(params: list[Param],
+                       addparams: Optional[list[BaseParam]]) -> TableHDU:
     """The PARAMETERS block for an xspec table model."""
 
     nint = len(params)
@@ -405,7 +413,6 @@ def xstable_parameters(params: list[IntParam],
         # The main array stores objects as each row can have different
         # lengths.
         #
-        # What is the best way to create an "empty" object array?
         values = np.zeros(ntotal, dtype=object)
         for idx, p in enumerate(params):
             values[idx] = np.asarray(p.values, dtype=np.float32)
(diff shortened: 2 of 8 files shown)


DougBurke:
Rebased to match #1836

DougBurke:
Rebased to pick up recent changes to #1836.

DougBurke:
I'm going to rebase this now that #1836 has been merged to make it a lot-easier to review. Not easy to review, just easier.

DougBurke:
And rebased.

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,754 @@
...
+@dataclass
+class IntParam(Param):
+    """Represent an interpolated parameter."""
+
+    values: list[float] = field(default_factory=list)
+    """The parameter values used to create the model spectra.
+
+    Is is required that they range from hardmin to hardmax and are in
+    monotonically increasing order. That is, it can not remain the
+    empty list.
```
DougBurke, line 220: I think what I meant here is that the `values` field can not  be empty.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -1465,3 +1465,100 @@ def set_arrays(filename, args, fields=None, ascii=True, clobber=False):
...
+    Parameters
+    ----------
+    hdu : astropy HDU
+    header : list of sherpa.astro.io.xstable.HeaderItem
+    """
+
+    for hdr in header:
+        card = [hdr.name, hdr.value]
+        if hdr.desc is not None or hdr.unit is not None:
+            comment = "" if hdr.unit is None else f"[{hdr.unit}] "
```
wmclaugh, line 1482: If hdr.unit is not None, comment is set to the value with an extra space behind it. If hdr.desc is None the extra space is left there. Should the space only be added if you are appending a value from hdr.desc?

sherpa/astro/io/pyfits_backend.py
```diff
@@ -1465,3 +1465,100 @@ def set_arrays(filename, args, fields=None, ascii=True, clobber=False):
...
+            comment += "" if hdr.desc is None else hdr.desc
```
wmclaugh, line 1483: Since the previous line initializes comment, there is no need to append "" to comment when hdr,desc is None.

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,754 @@
...
+
+These routines should be considered experimental, as it is not
+obvious that the interface is as usable as it could be.
+
+Example
+-------
+
+The following example creates an additive table model
+that represents a gaussian model where the only parameters are
+the line positon and the normalization.
```
wmclaugh, line 37:
```suggestion
the line position and the normalization.
```

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,754 @@
...
+
+With this we can create the necessary information to create the
+table model (`make_xstable_model`) and then write it out
+as a FITS file (`write_xstable_model`):
+
+>>> hdus = xstable.make_xstable_model("gaussy", elo, ehi, params=[param],
+...                                   spectra=models)
+>>> xstable.write_xstable_model("example.mod", hdus, clobber=True)
+
+This file can the be used in Sherpa with either
```
wmclaugh, line 99:
```suggestion
This file can then be used in Sherpa with either
```

sherpa/astro/io/pyfits_backend.py
DougBurke: I'll have to look at the `astropy.io.fits` logic

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,754 @@
...
+   mygauss.pos  thawed            1            0          2.4
+   mygauss.norm thawed            1            0        1e+24
+
+Notes
+-----
+
+XSPEC models can be created without XSPEC support in Sherpa, but it is
+needed to read the files in.
+
+For additive models it is assumed that the model values - each those
```
wmclaugh, line 118: perhaps 'each of those' or simply 'those'

DougBurke: Not English sure what writing I was there. I'm changing it to `... - that is, each bin - ..`. Ta.

DougBurke:
In #1869 I extend the `io_types` work (but it doesn't really build on the work here, other than the idea that we are providing a typed API rather than just saying "here, take this dict and hope it's okay").

hamogu:
> but I have used Any for handling of ndarray-like data (since the last I looked there was no nice simple way for me to say iterable or 1D ndarray, and also some of the typing assumes lists when really it should be more generic (e.g. sequence).

I thought that's what https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike is for?

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,752 @@
...
+xstablemodel.mygauss
+   Param        Type          Value          Min          Max      Units
+   -----        ----          -----          ---          ---      -----
+   mygauss.pos  thawed            1            0          2.4
+   mygauss.norm thawed            1            0        1e+24
+
+Notes
+-----
+
+XSPEC models can be created without XSPEC support in Sherpa, but it is
```
hamogu, line 115:
```suggestion
XSPEC models can be created without XSPEC support in Sherpa, but XSPEC support is
```

sherpa/astro/io/pyfits_backend.py
```diff
@@ -1465,3 +1465,101 @@ def set_arrays(filename, args, fields=None, ascii=True, clobber=False):
...
+    out : astropy.io.fits.PrimaryHDU
+
+    """
+
+    out = fits.PrimaryHDU()
+    _add_header(out, hdu.header)
+    return out
+
+
+def _create_table_hdu(hdu):
```
hamogu, line 1509: I beleive that this might not be the easiest way to write the code, but it works and gives correct answers (accordig to the tests) and speed is not a concern, so I'm not going to nitpick if e.g. units are passed to the initial table or are added "by hand" to the header later.

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,752 @@
...
+    default.
+
+    """
+    name: str
+    """The parameter name.
+
+    This must be unique and ideally a valid attribute name in Python,
+    although the latter is not enforced.
+    """
+    initial: float
```
hamogu, line 159: I like an empty line betwene the preceeding docsting and the next attribute. I think it makes the code easier to read. Not a requirement though.

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,752 @@
...
+
+        if self.initial < self.softmin or self.initial > self.softmax:
+            raise ValueError(f"Parameter {self.name} initial={self.initial} outside softmin={self.softmin} to softmax={self.softmax}")
+
+        if self.delta == 0:
+            raise ValueError("delta can not be 0")
+
+
+@dataclass
+class IntParam(Param):
```
hamogu, line 212: I read this as "IntegerParam". Can we call it "InterpolatedParam"?

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,752 @@
...
+            raise ValueError(f"Parameter {self.name} hardmax={self.hardmax} != last value={self.values[-1]}")
+
+        if np.any(np.diff(self.values) <= 0):
+            raise ValueError(f"Parameter {self.name} values are not monotonically increasing")
+
+
+# Perhaps the HeaderItem and Column types can be used in
+# sherpa.astro.io to pass information to and from the backends, rather
+# than the current system. However, that is for later work.
+#
```
hamogu, line 247: We do re-impleent pycrates and astropy.io.pyfits here, though, both of which have classes for header cards, Columns, tables and HDUs. I guess that's the price we pay for having abstraction from the backend. So, passing those in, instead of introducing another layer of abstraction by turning it into dicts and lists to pass to the backend code seems duplication. Also, We might want to move the definitions of these classes into `sherpa.io` and use them inside the other backend functions, too. I think that might actually simplify a lot of the backend code. But I'm OK with placing it here for now and moving it later.

hamogu (approved):
At the risk of looking stupid later, I think this is one of those cases where we need some real-world experience to see how it goes. It's going to be rarely used (few people make table models) and it's also not a complicated code, despite what @DougBurke said above. It's mostly bookkeeping of keywords.

The bigger question for me is if we should push this philsophy of having classes for HDUHeader, Key, HDUTable, Column, etc. into the sherpa.io routines. That would mean writing all our code "closer to the metal" of the fits files. If we do that, pyfits_backend and crate_backend, which currently contains substatial and duplicated logic (with occasional bugs cropping up where they are inconsistent), would become very thin wrappers that only translate from those classes to crates/pyfits.

The more I think about it, the more I like it, but again, it's good to try this out in an experimental modules like here.

sherpa/astro/io/xstable.py
DougBurke: I think I refer to the comment on line 246..... ;-)

DougBurke: I m using this idea in #1869 but at a different level of abstraction (storing data related to a particular "block" of data). So we could be just abstracting ourselves to oblivion (also that code doesn't take advantage of the "HeaderCard" interface provided here when maybe it should).

sherpa/astro/io/xstable.py
```diff
@@ -0,0 +1,752 @@
...
+       Should the redshift parameter be added? The default is False.
+    escale : bool, optional
+       Should the escale parameter be added? The default is False.
+    lolim, hilim : bool, optional
+       The value to used when energies are below or above the values
+       in egrid_lo or egrid_hi.
+    units : str or None, optional
+       The model units. For additive models the default is
+       "photons/cm^2/s" and for multiplicative models it is "".
+    xfxp : sequence of str or None
```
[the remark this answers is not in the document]
DougBurke: The naming on our side isn't great. As we don't actually support this feature of XSPEC tables I am okay with having ugliness (until we are able to use them, Which I'd hoped would be this release but :-(

sherpa/astro/io/xstable.py
DougBurke: As I expect this class to be used more than the Param class I'm renamnig things to

    Param -> BsaeParam
    IntParam -> Param

since `InterpolatedParameter` is a better name but it's too long...

sherpa/astro/io/xstable.py
DougBurke: I also have moved these types into a separate module there - `io_types` - but that's not worth doing here.

sherpa/astro/io/xstable.py
DougBurke: I've gone with my renaming.

sherpa/astro/io/xstable.py
DougBurke: Actually, as I was editing this file I've added spaces.

PR #1863, opened 2023-08-29
Bump test workflows to use xspec 12.13.1

Marie-Terrell:
- Bumps the Linux and macOS build tests to use the latest xspec
- Bumps one other Linux job to cover the switched 12.13.0 test
- Switches from the 10.9 SDK to the 10.14 SDK. We built xspec with 10.14 SDK due to the changes to the clock function, so we've updated Sherpa to build against this version as well.

### DIFF
--- a/.github/scripts/setup_xspec.sh
+++ b/.github/scripts/setup_xspec.sh
@@ -21,6 +21,9 @@ xspec_include_path=${xspec_root}/include/
 
 case "${XSPECVER}" in
   12.13.0*)
+      xspec_version_string="12.13.1"
+      ;;
+  12.13.0*)
       xspec_version_string="12.13.0"
       ;;
   12.12.1*)
--- a/.github/scripts/setup_xspec.sh
+++ b/.github/scripts/setup_xspec.sh
@@ -20,7 +20,7 @@ xspec_library_path=${xspec_root}/lib/
 xspec_include_path=${xspec_root}/include/
 
 case "${XSPECVER}" in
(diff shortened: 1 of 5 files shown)


.github/scripts/setup_xspec.sh
```diff
@@ -20,6 +20,9 @@ xspec_library_path=${xspec_root}/lib/
 xspec_include_path=${xspec_root}/include/
 
 case "${XSPECVER}" in
+  12.13.1*)
+      xspec_version_string="12.13.1"
+      ;;
   12.13.0*)
```
DougBurke, line 26: Can we drop either the 12.12.0 or 12.12.1 rules (just below) now?

Marie-Terrell: I don't have any preference, so I'm fine with dropping either/both from here and the test matrix if we don't need it

DougBurke: If we don't use them in the test matrix then I suggest dropping them (we don't want to be guaranteeing to users that we'll keep the old versions around!).

Marie-Terrell: If that's the case, are you okay with this going in for now and a separate PR being made for dropping them from the test matrix, this file, and any other doc/source updates?

DougBurke: Yeah - let's get this done now so our CI runs get to exercise this.

DougBurke (approved):

PR #1866, opened 2023-08-30
Switch to conda-forge and use new Conda channel

Marie-Terrell:
The proposal for this PR  is to:

- Switches Sherpa from "defaults" to "conda-forge"
- Switches Anaconda references to Conda
- Uses a new channel for the official Conda releases
Each run of the deployment pipeline will store the artifacts on github (to expire after 3 days).

New proposed Conda package release process:

- When an official release is being worked, the Sherpa packages will be uploaded to a test channel for checks
- When the release goes out, those packages will be uploaded to an official channel (proposed as: https://cxc.cfa.harvard.edu/conda/sherpa)

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -505,8 +505,10 @@ The
 contains the ``sherpatest`` package, which provides a number of
 data files in ASCII and :term:`FITS` formats. This is
 only useful when developing Sherpa, since the package is large.
-Example: Installing the 4.14.1 data via pip::
+A version of the test data is released for each `version of Sherpa <https://doi.org/10.5281/zenodo.593753>`_.
 
-   pip install https://github.com/sherpa/sherpa-test-data/archive/4.14.1.zip
+As an example, the 4.15.1 version of the test data can be installed with pip::
 
-And will then automatically be picked up by the ``sherpa_test`` script.
+   pip install https://github.com/sherpa/sherpa-test-data/archive/4.15.1.zip
+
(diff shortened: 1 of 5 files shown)


docs/install.rst
```diff
@@ -501,8 +500,13 @@ Sherpa was configured when built)::
...
+The
+`Sherpa test data suite <https://github.com/sherpa/sherpa-test-data>`_
+contains the ``sherpatest`` package, which provides a number of
+data files in ASCII and :term:`FITS` formats. This is
+only useful when developing Sherpa, since the package is large.
+Example: Installing the 4.14.1 data via pip::
+
+   pip install https://github.com/sherpa/sherpa-test-data/archive/4.14.1.zip
+
+And will then automatically be picked up by the ``sherpa_test`` script.
```
DougBurke, line 512:
```suggestion
A version of the test data is released for each `version of Sherpa <https://doi.org/10.5281/zenodo.593753>`_.

As an example, the 4.15.1 version of the test data can be installed with pip::

   pip install https://github.com/sherpa/sherpa-test-data/archive/4.15.1.zip

The test data will be automatically picked up by the Python tests and the ``sherpa_test`` script.
```

The link I added here isn't quite correct - it doesn't go to a list of releases, just the latest, but
- it does let you find out what the latest release is
- you can get a list of releases from this page

I didn't want to use a URL for the "list all releases" page as it is something that looks like the format could change over time.

DougBurke (approved):
I have not looked at the deployment stuff as I don't have a clue about that part!

PR #1869, opened 2023-08-31
Support multi matrix rmf

DougBurke:
I've moved some logic into #1912 to warn a user when a multi-matrix RMF is found. Need to think more about the imlpementation (the following is a good start but need to think more about the design).

It now looks like I'll be doing the work in #1921 but I'll leave this open as it'll help me review things.

Actually
- #1944
replaces this PR.

# Summary

Support reading in RMF files with multiple matrix blocks from OGIP Calibration Memo CAL/GEN/92-002  https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html. This support should be considered experimental at this time.

As part of this work the handling of error conditions in both the crates and pyfits backend has changed, so that different errors may be reported compared to earlier versions of Sherpa. 

# Details

This requires

- #1861

as I use some of the new types added in that PR.

Can we handle multi-MATRIX RMF blocks? We have a decision on how to handle this - do we extend `DataRMF` or do we handle this within the multiple-response instrument code? When extending `DataRMF` we have

- [+] it should be seamless to the user (e.g. no difference to using a "normal" RMF)
- [-] this overloads the meaning of the `DataRMF` fields and subverts OO methodology - e.g. https://en.wikipedia.org/wiki/Liskov_substitution_principle
  - I will note we have existing code that does this already, so this is not new 
 
Using the multiple-response code might be nice, but it would require significant work on the part of the user, and it's not clear what should be done.

I have therefore opted to add a `DataMultiRMF` class that extends `DataRMF` since I think the ease-of-use argument is very important. I am slapping on a "this is experimental" notice so we can change it if we find this ends up not working well. I will note that I wrote #1907 in large-part because of this PR, and I think it is something to look at post the 4.16 release (no matter what approach we take for this PR).

There are a surprising number of commits before we get to adding multi-matrix support as

- we had cruft to clean up
- I wanted to add support for multi-matrix RMFs in stages to hopefully make it easier to review and so that we can stop at any point and say "this may not do everything" but does at least move the needle
  - in particular, the existing Sherpa code does not have any idea that multi-matrix files exist (and, in fact, the specification does not make it possible to know we have such a file) and so it is possible to read in one of these RMFs and only get one of the matrices used (and we can't guarantee which one it will be); at least know (with some of the commits,  a user would at least know they've lost data up until the commit where we add support)

As with #1861 - which this PR builds on - I've added dataclasses to describe the data we want to send from the backend to the io layer to describe a RMF. This is instead of just passing around a dictionary with keywords we just have to agree on. These structures have guided the change from supporting a "single matrix" to allowing multiple ones.

For testing, as we don't have a real-world example, I have just created a "fake" response. This response is the combination of a "perfect" grid together with a blurry one, and we can see as the commits progress how much of it we can use.

In doing the clean-up work there has been some changes to both the crates and pyfits handling of error conditions. This has generally removed differences between the two backend, as well as centralizing at least one check (so it's now done in the generic IO code rather than in the backends). The main point is that code will still error out but maybe in a slightly-different manner to before. There are **some** cases which would have worked before but now fail, but these are cases where the RMF does not meet the OGIP standard - e.g. see #1871 - and we do not have tests that fail with the new code, so I claim this is over-engineered code we should clean up. At some point. Which is not now. Other than the clean up I have done.

Since AstroPy just reads in the raw data, the changes needed are not too significant. For the crates backend we were using the RMFCrateDataset to read in data, but this does not support multiple matrices (and will not for the 4.16 release) so we have to augment the existing code. We do so by still using RMFCrateDataset, so we still behave the same with radically-invalid input files (i.e. this is only about error-ing out with invalid data) but then we switch to CrateDataset to read in all the data. The API allows a user to send in a RMFCrateDataset in [just like for AstroPy/pyfints they can send a HDUList in] which we support but users of that will not be able to handle multi-matrix RMF files. I do not believe there are any people who do this because we do not document it (i.e. they would need to be calling routines from `crates_backend` directly), and because there is no way to be sure we have been given a multi-matrix file when looking at a single MATRIX block.

# Issues

One issue is that we only have the spec documentation. There aren't really any files lying around with real data (maybe from XRISM?) which makes testing this out and checking it works properly a bit awkward.

You can not write out a multi-matrix RMF (via `write_rmf`). This should be fixable.

The "representation" of a multi-matrix RMF hides all but the first block - e.g. print(rmf) - which is not ideal. This should be fixable. I think the new `plot_rmf` command should work but I can not be sure.

In part because we have some strange non-OGIP cases (#1871) and in part because of the existing API (#1907) I am not 100% convinced there aren't bugs hiding here.

The new RMF does not do any of the fancy "ignore/notice" handling that `DataRMF` does. I should have this "disabled", but I did mention the issue of hidden bugs, didn't I?

### DIFF
--- a/docs/overview/astro_io.rst
+++ b/docs/overview/astro_io.rst
@@ -6,8 +6,6 @@ The sherpa.astro.io module
 
 .. automodule:: sherpa.astro.io
 
-   .. rubric:: Attributes
-
    .. rubric:: Functions
 
    .. autosummary::
--- a/docs/overview/astro_io_io_types.rst
+++ b/docs/overview/astro_io_io_types.rst
@@ -14,9 +14,11 @@ The sherpa.astro.io.io_types module
       Column
       HeaderItem
       TableHDU
+      RMFMatrixData
+      RMFEboundsData
 
 Class Inheritance Diagram
 =========================
 
-.. inheritance-diagram:: Column HeaderItem TableHDU
+.. inheritance-diagram:: Column HeaderItem TableHDU RMFMatrixData RMFEboundsData
    :parts: 1
--- a/docs/overview/astro_io_io_types.rst
+++ b/docs/overview/astro_io_io_types.rst
@@ -1,0 +1,22 @@
+***********************************
+The sherpa.astro.io.io_types module
+***********************************
+
+.. currentmodule:: sherpa.astro.io.io_types
+
+.. automodule:: sherpa.astro.io.io_types
+
+   .. rubric:: Classes
+
+   .. autosummary::
+      :toctree: api
+
+      Column
+      HeaderItem
+      TableHDU
+
+Class Inheritance Diagram
+=========================
+
+.. inheritance-diagram:: Column HeaderItem TableHDU
+   :parts: 1
--- a/docs/overview/astro_io_xstable.rst
+++ b/docs/overview/astro_io_xstable.rst
@@ -1,0 +1,29 @@
+**********************************
+The sherpa.astro.io.xstable module
+**********************************
+
+.. currentmodule:: sherpa.astro.io.xstable
+
+.. automodule:: sherpa.astro.io.xstable
+
+   .. rubric:: Classes
+
+   .. autosummary::
+      :toctree: api
+
+      IntParam
+      Param
+
+   .. rubric:: Functions
+
+   .. autosummary::
+      :toctree: api
+
+      make_xstable_model
+      write_xstable_model
+
+Class Inheritance Diagram
+=========================
+
+.. inheritance-diagram:: IntParam Param
+   :parts: 1
--- a/docs/overview/utilities.rst
+++ b/docs/overview/utilities.rst
@@ -57,4 +57,5 @@ Reference/API
    astro
    astro_io
    astro_io_wcs
+   astro_io_xstable
    astro_utils_xspec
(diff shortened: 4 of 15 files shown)


anetasie:
@DougBurke We don't have a file example, but I'm wondering if `astropy` handles the multi-matrix rmfs in terms of just reading and accessing these `rmf` files. I guess it should.

DougBurke:
@anetasie AstroPy doesn't understand concepts like PHA, ARF, or RMF. It just deals with the underlying data in the FITS files. So it can read these files just fine, because it doesn't care about what is in it (as long as the formatting is correct).

HTH

anetasie:
I was just looking at the code for testing multi-matrix rmf and was thinking that perhaps we could try using this concept on the grating data with overlapping orders, so each order is represented by one matrix block and for the model we would convolve the model with each of them and sum to fit the data? but this may not be a great idea as I don't deal with the grating data at all. @hamogu may have some ideas of whether this would be helpful at all.

DougBurke:
I'm probably going to wait on #1861 getting merged before trying to address these conflicts.

DougBurke:
I've removed the draft label as it is now usable. I do expect to see some more work done to the code to address the listed issues above.

DougBurke:
With relation to @anetasie comment at https://github.com/sherpa/sherpa/pull/1869#issuecomment-1737853645 - this is a **completely different** reality to the existing multi-order code, so it is not at all easy to link the two. Conceptually there are strong similarities, but operationally they are completely different and we'd need to clean up things like 

- #1871
- #1907 

to make meaningful work here.

PR #1872, opened 2023-09-01
docs: note difference to XSPEC for channel analysis

DougBurke:
# Summary

Note the differences between Sherpa and XSPEC for channel analysis of PHA data in the set_analysis documentation.

# Details

This has been keeping me up at nights, so I thought I'd add a note to try and explain the difference in the "channel" handling between Sherpa and XSPEC.

I think the text needs work - I wanted to say that you can use the same channel numbers no-matter the grouping, unlike XSPEC, but it's not technically correct [since the result of the ignore/notice call will depend on the grouping] but perhaps I am over-thinking it and can drop the "caveat" bit in the text.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -6564,6 +6564,18 @@ class Session(sherpa.ui.utils.Session):
         they are interpreted as the `id` and `quantity` parameters,
         respectively.
 
+        Sherpa does not match XSPEC when the `quantity` setting is set
+        to "channel", particularly when the data is grouped. Sherpa
+        retains the original CHANNEL values from the PHA dataset, so
+        calls to `ignore` and `notice` will use the same values even
+        if the grouping changes (although the results of these calls
+        may change depending on how the group edges match the selected
+        ranges), whereas XSPEC counts each group as a single channel.
+        This is also noticeable in calls to `plot_data`, `plot_model`,
+        and plot_fit`, where Sherpa will display each group with its
+        width in the number of PI or PHA channels, rather than having
+        a width of 1 as in XSPEC.
+
         Examples
         --------
 
@@ -6580,13 +6592,13 @@ class Session(sherpa.ui.utils.Session):
         Set data set 1 to use channel units. Plots will use a Y
         axis of count/bin rather than the default count/s/bin.
 
-        >>> set_analysis(1, 'bin', 'counts')
+        >>> set_analysis(1, 'bin', type='counts')
 
         Set data set 1 to use energy units. Plots of this data set
         will display keV on the X axis and counts keV (i.e.
         counts/keV * keV^2) in the Y axis.
 
-        >>> set_analysis(1, 'energy', 'counts', 2)
+        >>> set_analysis(1, 'energy', type='counts', factor=2)
 
         """
         if quantity is None:


DougBurke:
I wonder if it might be a good place to describe how we use the RMF EBOUNDS data to "fake up" energies/wavelengths (e.g. whey you `notice(0.5,6)` when `analysis=energy` it's only approximate.

sherpa/astro/ui/utils.py
```diff
@@ -6564,6 +6564,18 @@ def set_analysis(self, id, quantity=None, type='rate', factor=0):
         they are interpreted as the `id` and `quantity` parameters,
         respectively.
 
+        Sherpa does not match XSPEC when the `quantity` setting is set
+        to "channel", particularly when the data is grouped. Sherpa
```
anetasie, line 6568: I would remove "Sherpa does no match XSPEC..." sentence and reference to XSPEC later on. This is Sherpa documentation and some people may not even know what XSPEC is (really :-))

sherpa/astro/ui/utils.py
```diff
@@ -6564,6 +6564,18 @@ def set_analysis(self, id, quantity=None, type='rate', factor=0):
...
+        retains the original CHANNEL values from the PHA dataset, so
+        calls to `ignore` and `notice` will use the same values even
+        if the grouping changes (although the results of these calls
+        may change depending on how the group edges match the selected
+        ranges), whereas XSPEC counts each group as a single channel.
```
anetasie, line 6573: also here - I would remove the part of the sentence with the reference to XSPEC

sherpa/astro/ui/utils.py
```diff
@@ -6564,6 +6564,18 @@ def set_analysis(self, id, quantity=None, type='rate', factor=0):
...
+        This is also noticeable in calls to `plot_data`, `plot_model`,
+        and plot_fit`, where Sherpa will display each group with its
+        width in the number of PI or PHA channels, rather than having
+        a width of 1 as in XSPEC.
+
```
anetasie, line 6578: I don't understand what a 'width of 1' means. 

We could just have one sentence at the end saying that Sherpa behaves differently than XSPEC and point to their documentation.

anetasie (commented):
I suggested some rewording of the paragraph.

sherpa/astro/ui/utils.py
```diff
@@ -6564,6 +6564,18 @@ def set_analysis(self, id, quantity=None, type='rate', factor=0):
...
 
+        Sherpa does not match XSPEC when the `quantity` setting is set
+        to "channel", particularly when the data is grouped. Sherpa
+        retains the original CHANNEL values from the PHA dataset, so
+        calls to `ignore` and `notice` will use the same values even
+        if the grouping changes (although the results of these calls
+        may change depending on how the group edges match the selected
+        ranges), whereas XSPEC counts each group as a single channel.
+        This is also noticeable in calls to `plot_data`, `plot_model`,
+        and plot_fit`, where Sherpa will display each group with its
```
wmclaugh, line 6575:
```suggestion
        and `plot_fit`, where Sherpa will display each group with its
```

sherpa/astro/ui/utils.py
DougBurke: So, is this whole paragraph not worth including then? I think it might be useful to have something that explains why our plots and XSPEC look different.  I put it at the end of the notes section so it's not the first thing a user sees.

For example here's XSPEC's equivalent of `plot_data`

xspec

and here's our version:

sherpa

sherpa/astro/ui/utils.py
DougBurke: If we don't have the XSPEC reference then I'm not sure we should include this paragraph at all (discussed more above).

sherpa/astro/ui/utils.py
DougBurke: As a user I might get confused because we haven't actually said what the difference is, and looking at the XSPEC documentation isn't going to help (in part because it doesn't really describe what it does, but in part because they are not going to have anything to say about what we do).

sherpa/astro/ui/utils.py
DougBurke: This is the **same** dataset (our favourite `sherpa-test-data/sherpatest/3c273.pi` for anyone who wants to play along at home).

sherpa/astro/ui/utils.py
DougBurke: Explaining the "width of 1" is hard to do (and at some level is the actual point of this paragraph). It's obviously a lot easier if we just don't add the text at all.

If you do care, you can see it above, in the XSPEC output, where each group has an effective width of 1 (as what XSPEC uses for channel numbers does not match the input channel grid), and is essentially the "group number", so each group has a width of 1 by default.For Sherpa we retain knowledge about the original channels (so the bins get placed correctly) an we divide by the number of channels in each group.

sherpa/astro/ui/utils.py
anetasie: Perhaps a better place would be to include this in the CIAO threads or FAQ on CIAO pages.

DougBurke:
Given we don't see an urgent need for this in Sherpa itself I'm closing it.

PR #1877, opened 2023-09-07
un-whitelist deprecation warnings

hamogu:
We have whitelisted a number of deprecation warnings and others in our testing. I believe that some of them are not longer relevant and thus should be removed. I'm pushing this as a PR so we get CI to run and see if any of our current tests fail with this.

Some of them date back to the Python 2.7 era according to a git blame.

Why do I care? I just had tests fail locally with an "np.asscalar\(a\) is deprecated" and I wondered why. I looked into Sherpa and did not find the string "asscalar" anywhere (because @DougBurke probably fixed that years ago). So, why did my test fail locally? Well, I had a recent version of numpy where assaclar has actually been removed and an older version of one of our other dependencies (astropy), which still used it. Since we whitelist this warning (even after Sherpa itself has been fixed) I;d never noticed that that particular environment still had such an old version of astropy. 

So, I might as well try and see how many of those warning we can remove from our whitelist and still pass our current matrix of tests.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -110,16 +110,7 @@ known_warnings = {
         [
             r"unorderable dtypes.*",
             r"Non-string object detected for the array ordering.*",
-            r"using a non-integer number instead of an integer will result in an error in the future",
             r"Use load_xstable_model to load XSPEC table models",
-            #  This does not have to do with Sherpa and is coming from some versions of
-            #  jupyter_client
-            r"metadata .* was set from the constructor.*",
-
-            # Matplotlib version 2 warnings (from HTML notebook represention)
-            #
-            r'np.asscalar\(a\) is deprecated since NumPy v1.16, use a.item\(\) instead',
-            r"Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working",
 
             # NumPy 1.25 warnings that are raised by (mid-2023) crates code.
             # Hopefully this can be removed by December 2023.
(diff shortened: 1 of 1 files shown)


hamogu:
Since CI passed, I'm hoping that this is good. I've asked for a CM review (@Marie-Terrell was suggested by GH first in the list before I saw @harlantc  but input fomr either of you is welcome).

I suggest to put this in sooner rather than later, so that we have a little time left to reactivate any of those white listings if we see warnings and failing tests cropping up in some obscure (i.e. not CI tested) combination of version dependencies.

DougBurke (approved):
This whitelist approach is always going to be something we have to review from time to time,so thanks for doing it. I'm a lot happier nowadays as we have

a) a reasonable coverage of build setups on CI
b) more tests than we used to

Marie-Terrell (approved):
This looks good to me. I'll double check this in CIAO once it goes in

hamogu:
And to @DougBurke I made this exactly because I had a random assortment of packages (in this case numpy 1.25 and astropy 5.0) and in numpy 1.25 the function in question was actually removed because the deprecation period was over, but my astropy was so old that it still called `np.asscalar`. There is no way to make that right except by upgrading astropy, so if a user (e.g. me) has an environment with sherpa 4.16 and dependencies that are so old that these warnings are thrown (i.e. at least four numpy verison back), I think we *want* the warning to be shown to entice users to upgrade whatever dependency throws that warning. 

The other whitelistings I removed where put in 7 years ago in the Python 2.7 era and we know Sherpa does not run on 2.7 any longer anyway.

Issue #1884, opened 2023-09-11
C++ changes?


DougBurke:
During Sunday the following change happened for a conda build (where I have not been changing any code neat this)

```
    building 'sherpa.optmethods._saoopt' extension
    compiling C++ sources
    C compiler: /usr/share/miniconda/envs/build/bin/x86_64-conda-linux-gnu-c++ -Wno-unused-result -Wsign-compare -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /usr/share/miniconda/envs/build/include -fPIC -O2 -isystem /usr/share/miniconda/envs/build/include -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /usr/share/miniconda/envs/build/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /usr/share/miniconda/envs/build/include -fPIC

    creating build/temp.linux-x86_64-3.9/sherpa/optmethods
    creating build/temp.linux-x86_64-3.9/sherpa/optmethods/src
    compile options: '-Isherpa/include -Isherpa/utils/src -Isherpa/utils/src/gsl -I/tmp/pip-build-env-unvhf4f1/overlay/lib/python3.9/site-packages/numpy/core/include -I/usr/share/miniconda/envs/build/include/python3.9 -I/usr/share/miniconda/envs/build/include/python3.9 -c'
    x86_64-conda-linux-gnu-c++: sherpa/optmethods/src/_saoopt.cc
    x86_64-conda-linux-gnu-c++: sherpa/optmethods/src/Simplex.cc
    In file included from sherpa/include/sherpa/MersenneTwister.h:70,
                     from sherpa/optmethods/src/DifEvo.hh:65,
                     from sherpa/optmethods/src/_saoopt.cc:29:
    /usr/share/miniconda/envs/build/x86_64-conda-linux-gnu/include/c++/11.2.0/ctime:80:11: error: 'timespec_get' has not been declared in '::'
       80 |   using ::timespec_get;
          |           ^~~~~~~~~~~~
[...]

PR #1886, opened 2023-09-11
Remove defaults channel in CI to avoid conflicts with conda-forge

Marie-Terrell:
This is an attempt to address #1884. Will check the CI to see if this resolves the issues.

**Background**
Currently, our "deployment" pipeline works because we install Miniforge for Conda (which already has defaults removed). For the development pipeline, the pre-installed miniconda on the GitHub Actions Linux runner does include defaults. We switched to conda-forge for dependencies (adding the channel in the process), but did not remove defaults as an option for the dev workflow. This change does that.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -29,6 +29,8 @@ source ${CONDA}/etc/profile.d/conda.sh
 # update and add channels
 conda update --yes conda
 conda config --add channels conda-forge
+#Remove defaults to avoid conflicts with conda-forge
+conda config --remove channels defaults
 
 # To avoid issues with non-XSPEC builds (e.g.
 # https://github.com/sherpa/sherpa/pull/794#issuecomment-616570995 )


DougBurke (approved):
LGTM

:shipit: 

Do we know what "caused" #1884 (ie could something like this happen again somehow)?

Marie-Terrell:
> LGTM
> 
> :shipit:
> 
> Do we know what "caused" #1884 (ie could something like this happen again somehow)?

So, this was caused because of conflicts between conda-forge and defaults for dependencies. Since the movement to "conda-forge" for dependencies for all builds, we should avoid installing this with packages from defaults. This should not come up again so long as the CI runs are either using something like miniforge or Mambaforge (where the default channels are "conda-forge") or have "defaults" disabled. What likely happened was that we lucked out for a bit, but some dependency updated on "conda-forge" and/or "defaults" resulting in a mix of packages that did not really belong together.

The main time this might come up again is for end users. So, we could potentially note the install line as something like:
`conda install -c https://cxc.cfa.harvard.edu/conda/sherpa -c conda-forge --override-channels sherpa`
Where the "--override-channels" should, theoretically, avoid picking up defaults if someone is using miniconda or anaconda.

DougBurke:
Ta

PR #1891, opened 2023-09-15
ui: add calc_bkg_stat - fix #333

### LINKED ISSUE #333: calc_bkg_stat
In sherpa, one can easily fit the background with fit_bkg, but it is remarkably hard to access the current background statistic. Adding a calc_bkg_stat as a an analog to calc_stat, similar to plot_fit->plot_bkg_fit would be great.

A workaround is:

bkg_stat = [s for s in get_stat_info() if id in s.ids and s.bkg_ids is not None and len(s.bkg_ids) > 0][0]

DougBurke:
# Summary

Adds the calc_bkg_stat, calc_bkg_stat_info, and get_bkg_stat_info routines. Fix #333.

# Details

Adds the routines

  calc_bkg_stat
  calc_bkg_stat_info
  get_bkg_stat_info

to the astro UI session. This is a relatively-easy extension of the existing calc/get_stat[_info] routines. There is one small issue, of what to do when you have multiple background datasets with different background ids, but for now just list all such ids and let the user worry about access to the per-background data.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -776,8 +776,9 @@ class Session(sherpa.ui.utils.Session):
     def calc_bkg_stat_info(self):
         """Display the statistic values for the current background models.
 
-        Returns the statistic values for the background datasets.  See
-        calc_stat_info.
+        Returns the statistics values for background datasets with
+        background models. See `calc_stat_info` for a description
+        of the return value.
 
         .. versionadded:: 4.16.0
 
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -718,6 +718,168 @@ class Session(sherpa.ui.utils.Session):
         txt = self._get_show_bkg_model(id, bkg_id)
         send_to_pager(txt, outfile, clobber)
 
+    def calc_bkg_stat(self, id=None, *otherids):
+        """Calculate the fit statistic for a background data set.
+
+        Evaluate the current background models for the background
+        datasets, calculate the statistic for each background, and
+        return the sum.  No fitting is done, as the current model
+        parameter, and any filters, are used. The `calc_bkg_stat_info`
+        routine should be used if the result for a particular
+        background component needs to be returned.
+
+        .. versionadded:: 4.16.0
+
+        Parameters
+        ----------
+        id : int or str, optional
+           The data set that provides the data. If not given then all
+           background data sets with an associated background model
+           are used simultaneously.
+        *otherids : int or str, optional
+           Other data sets to use in the calculation.
+
+        Returns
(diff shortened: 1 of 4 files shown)


anetasie:
@DougBurke Thank you for adding these functions. My only comment is to the documentation for `calc_bkg_stat_info` which could include the description of the returned values, this is  the same as given by `calc_stat_info`, so maybe include exact description as in  `calc_stat_info` or just a sentence saying that this is the same output for the background.

```
In [7]: calc_bkg_stat_info()
Background 1 in Dataset = 1
Statistic             = cstat
Fit statistic value   = 276.204
Data points           = 512
Degrees of freedom    = 511
Probability [Q-value] = 1
Reduced statistic     = 0.540516
```

sherpa/astro/ui/utils.py
```diff
@@ -718,6 +718,168 @@ def show_bkg_model(self, id=None, bkg_id=None, outfile=None, clobber=False):
...
+
+        """
+        ids, f = self._get_bkg_fit(id, otherids)
+        return f.calc_stat()
+
+    def calc_bkg_stat_info(self):
+        """Display the statistic values for the current background models.
+
+        Returns the statistic values for the background datasets.  See
+        calc_stat_info.
```
anetasie, line 780: Suggest rewording with the direct link to calc_stat_info, or include the full information about the output.

Returns the statistics values for the current background dataset, as in `calc_stat_info` for the source dataset.

DougBurke: I am reluctant to copy the documentation over from `calc_stat_info` as I can tell you from direct experience that it's hard to keep things up to date. I am not sure what you mean by "a direct link". We have:

- for the python user - calling `help(calc_bkg_stat_info)` - the "See Also" section a few lines below this line which already mentions `calc_stat_info`
- for the users of RTD this will become an actual link, which you can see an example of in the RTD build of this PR at https://sherpa--1891.org.readthedocs.build/en/1891/ui/api/sherpa.astro.ui.calc_bkg_stat_info.html
- for CIAO users I convert these files into ahelp files; technically there can be a link between the two but for various technical reasons it can get confusing - e.g. see the `calc_stat_info` "see also" section for [CIAO 4.15](https://cxc.harvard.edu/sherpa/ahelp/calc_stat_info.html) and it's not clear what the version for this command will look like

anetasie: I agree that copying is not a good solution and I thought that updating the wording to specifically say that this as in `calc_stat_info`. For me See Also does not necessarily mean that you are going to see the exact output of `calc_bkg_stat_info`

anetasie: That's good. Thanks

Issue #1887, opened 2023-09-13
Should plotting fail or create an empty plot if there is no data?


hamogu:
This was noticed while working on #1382, but we decided there that it's out of scope for that PR, so I'm opening a separate issue.

Our current pylab backend for plotting is inconsistent about what happens if a plot is empty. When errors are commanded to be part of the plot (such as a DataPlot) the plot is done with `plt.errorbar` which works and thus returns an empty plot. On the other hand, for plots with no errorbars like models, `plt.plot` is used, which errors out when there is no x,y and thus an error is raised. This matters most for the `_repr_html_` in the notebook, which catches the error and returns None, i.e. no plot. Our tests check that behavior extensively in https://github.com/sherpa/sherpa/blob/main/sherpa/plot/tests/test_plot_notebook.py with a check_empty/check_full within most tests.

You can see the inconsistency in the tests: e.g. compare the tests for `DataPlot` and `ModelPlot` (as of writing line 125 and 154, but line number may change over time) in https://github.com/sherpa/sherpa/blob/main/sherpa/plot/tests/test_plot_notebook.py

It seems that this is mostly a check of existing behavior, not an intentional choice of API. That makes is a little difficult to reproduce exactly in a different backend, which handles those things differently. Thus, I'd like to make an API spec that makes DataPlot and ModelPlot (and all the others) behave the same. 

I suggest that "if I ask for a plot, there should always be a plot even if it's empty".

I note that:
- This rarely comes up in normal user workflows. Usually, you make a plot with data (or with a model or with a CDF), and don't worry how that plot would look like if it was empty (no data, no model, no CDF).
- It's mostly relevant for the `_repr_html_` that displays a summary of an object in a notebook. I would argue that it would be overly pedantic to consider this part of the API we try not to break.

PR #1899, opened 2023-09-18
Bokeh_class

hamogu:
This is a fully working implementation, that could be merged as is.
However, there are a few UI improvements that might be worthwile, e.g. there is no color cycline and the bokeh.plot.show could do with a nice wrapper in sherpa.


hamogu:
This PR looks a lot more intimidating than it is, because it is build on top of #1382. Only the last commit (at the time of writing) is actually new. Locally, this passes almost all ploting tests we have, but I've not activated them in this PR, because it only passes "almost" all plotting tests - see #1887. When we get close to merging, I csn simply add bokeh as a new requirement on CI and special case those few tests that fail; but that's not a priority here.

It's more important that @DougBurke and @anetasie take a look at the current functionality.

docs/indices.rst
```diff
@@ -23,6 +23,13 @@ Glossary
        A community Python library for Astronomy:
        https://www.astropy.org/.
 
+   bokeh
+       The bokeh plotting package, which is documented at
+       http://bokeh.org/, is used to provide the plotting support
```
wmclaugh, line 28:
```suggestion
       https://bokeh.org/, is used to provide the plotting support
```

wmclaugh, line 28: "used to provide **the** plotting support in sherpa" should be changed "used to provide plotting support in sherpa" as the word 'the' changes the statement's connotation.

sherpa/conftest.py
```diff
@@ -47,13 +47,41 @@
...
 
-import sherpa.plot
-import sherpa.astro.plot
-from sherpa.plot import dummy_backend
+from sherpa.plot.backends import BaseBackend, IndepOnlyBackend, PLOT_BACKENDS
+from sherpa.plot import TemporaryPlottingBackend
+
+FUNCTIONAL_PLOT_BACKENDS = []
+'''Unfortunately, there is no metadata that tell us which plotting backends are
+dummys and which ones actually produce useful plots. That's because
```
wmclaugh, line 55:
```suggestion
dummies and which ones actually produce useful plots. That's because
```

sherpa/plot/backends.py
```diff
@@ -0,0 +1,961 @@
...
+           'PLOT_BACKENDS',
+           )
+
+PLOT_BACKENDS = {}
+'''global list of all successfully imported plotting backends'''
+
+lgr = logging.getLogger(__name__)
+warning = lgr.warning
+
+backend_indep_colors = list('rgbkwcymk') + [None]
```
wmclaugh, line 42: There are a number of 'Pending' comments to this file in #1382 that apply to this file. For example, "rgb**k**wcym**k**" lists black twice.

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,778 @@
...
+            The number of columns.
+        create : bool, optional
+            If True then create the plots
+        top : int
+            The row that is set to the ratio height, numbered from 0.
+        ratio : float
+            The ratio of the height of row number top to the other
+            rows.
+        """
+        # This function is written to work with an imcomplete grid of plots,
```
wmclaugh, line 679:
```suggestion
        # This function is written to work with an incomplete grid of plots,
```

sherpa/plot/backends.py
hamogu: Then there is no need to comment on them here; I plan to rebase this on main after #1382 is merged.

docs/indices.rst
hamogu: I'll address that in a local change, because that means I should also modify the matplotlib glossary entry to make them match.

sherpa/plot/backends.py
hamogu: @wmclaugh I don't see those comments. If they are "pending" on your account, you have not submitted them yet. There should be a "finish review" button or something like that that will actually post all the comments that you wrote so that I can see them.

wmclaugh: @hamogu-  Apologies- you are correct. They should be visible now.

hamogu:
I added the test and rebased to pick up the uo-to-date RTD configuration file (to fix the RTD failure in the previous run)

DougBurke:
@hamogu - you say you rebased this - in a comment an hour or so ago at the time of writing - but the GH still lists https://github.com/sherpa/sherpa/pull/1899/commits/fafc0cc1040aa926a593c76ec2aa3508f622fadc as the last commit, which was made 4 hours ago. Was your "rebase" comment about the fafc0cc code?

DougBurke:
I've got this far before I've run out of TUITs today:

```
In [1]: from sherpa import plot

In [2]: plot.PLOT_BACKENDS
Out[2]: 
{'BaseBackend': sherpa.plot.backends.BaseBackend,
 'BasicBackend': sherpa.plot.backends.BasicBackend,
 'IndepOnlyBackend': sherpa.plot.backends.IndepOnlyBackend,
 'pylab': sherpa.plot.pylab_backend.PylabBackend,
 'PylabErrorArea': sherpa.plot.pylab_area_backend.PylabErrorArea,
 'BokehBackend': sherpa.plot.bokeh_backend.BokehBackend}

In [3]: plot.set_backend("bokeh")
IdentifierErr                             Traceback (most recent call last)
Cell In[3], line 1
----> 1 plot.set_backend("bokeh")

File ~/sherpa/sherpa-xspec/sherpa/plot/__init__.py:177, in set_backend(new_backend)
    175         backend = PLOT_BACKENDS[new_backend]()
    176     else:
--> 177         raise IdentifierErr('noplotbackend', new_backend,
    178                             list(PLOT_BACKENDS.keys()))
    179 # It's a class and that class is a subclass of BaseBackend
    180 elif isinstance(new_backend, type) and issubclass(new_backend, BaseBackend):

IdentifierErr: 'bokeh' is not a valid plotting backend, choose from ['BaseBackend', 'BasicBackend', 'IndepOnlyBackend', 'pylab', 'PylabErrorArea', 'BokehBackend']

In [4]: plot.set_backend("BokehBackend")

```

DougBurke:
I did just try the example-plots backend and added a call to

    plot.set_backend("BokehBackend")

but then the `<plot object>.plot()` calls didn't seem to do anything. What am I doing wrong [I also took out the `%matplotlb inline` line and restarted the kernel just in case that messed things up. Is there a bokeh-equivalent to this magic (although I'm never sure when this magic is needed anymore).

hamogu:
> I added the test and rebased to pick up the uo-to-date RTD configuration file (to fix the RTD failure in the previous run)

That comment is in the wrong PR... it was meant for the link-parameters PR. I ahve too many tabs open that tool almost, but not quite identical. Sorry for the confusion.

hamogu:
@DougBurke Are you running in the notebook or on a traditional command line?

See https://sherpa--1899.org.readthedocs.build/en/1899/plots/bokeh_backend.html#module-sherpa.plot.bokeh_backend for display help. 

But yes, it should be `plot.set_backend("BokehBackend")` (I fixed the locally already, but did not push again), and yes, the user interface of "how do I get this plot" can be improved. Suggestions welcome.

DougBurke:
So, I start with "jupyter lab" and then have

```python
import numpy as np

from sherpa import data
from sherpa.astro import data as astrodata

from sherpa import plot
from sherpa.astro import plot as astroplot

x1 = [100, 200, 600, 1200]
y1 = [2000, 2100, 1400, 3050]

d1 = data.Data1D('oned', x1, y1)

plot1 = plot.DataPlot()
plot1.prepare(d1)
plot1.plot()
show(plot.backend.current_fig['all_axes'])
```

and a new tab opens up with a file-not-found error (in my case looking for file:///tmp/tmpg6_1hr2n.html but it's obviously a randomly-generated file.

Actually, the file exists but for some reason trying to display it fails. At first I thought it was a permissions error from jupyter lab, but trying "gio open /tmp/...html" failed in the same way. It's too late and wet for me to diagnose. Is firefox being protective of my /tmp/ directory (or something equally annoying to diagnose)? I think so, because if I "firefox /tmp/..." or "gio open /tmp/..." you get a file-not-found error, but if I manually open the file via file/open file it displays. Gah....

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Bokeh Plot</title>
    <style>
      html, body {
        box-sizing: border-box;
        display: flow-root;
        height: 100%;
        margin: 0;
        padding: 0;
      }
    </style>
    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-3.2.2.min.js"></script>
    <script type="text/javascript">
        Bokeh.set_log_level("info");
    </script>
  </head>
  <body>
    <div id="e0448a2b-9e27-4fb9-a01d-57a21641536b" data-root-id="p1063" style="display: contents;"></div>
  
    <script type="application/json" id="p1076">
      {"21276715-a5ae-4bcf-b18f-09fa6758ce03":{"version":"3.2.2","title":"Bokeh Application","roots":[{"type":"object","name":"GridPlot","id":"p1063","attributes":{"rows":null,"cols":null,"toolbar":{"type":"object","name":"Toolbar","id":"p1062","attributes":{"tools":[{"type":"object","name":"PanTool","id":"p1054"},{"type":"object","name":"WheelZoomTool","id":"p1055"},{"type":"object","name":"BoxZoomTool","id":"p1056","attributes":{"overlay":{"type":"object","name":"BoxAnnotation","id":"p1057","attributes":{"syncable":false,"level":"overlay","visible":false,"left_units":"canvas","right_units":"canvas","bottom_units":"canvas","top_units":"canvas","line_color":"black","line_alpha":1.0,"line_width":2,"line_dash":[4,4],"fill_color":"lightgrey","fill_alpha":0.5}}}},{"type":"object","name":"SaveTool","id":"p1061"},{"type":"object","name":"ResetTool","id":"p1059"},{"type":"object","name":"HelpTool","id":"p1060"}]}},"children":[[{"type":"object","name":"Figure","id":"p1033","attributes":{"x_range":{"type":"object","name":"DataRange1d","id":"p1034"},"y_range":{"type":"object","name":"DataRange1d","id":"p1035"},"x_scale":{"type":"object","name":"LinearScale","id":"p1065"},"y_scale":{"type":"object","name":"LinearScale","id":"p1066"},"title":{"type":"object","name":"Title","id":"p1040","attributes":{"text":"oned"}},"renderers":[{"type":"object","name":"GlyphRenderer","id":"p1071","attributes":{"data_source":{"type":"object","name":"ColumnDataSource","id":"p1067","attributes":{"selected":{"type":"object","name":"Selection","id":"p1068","attributes":{"indices":[],"line_indices":[]}},"selection_policy":{"type":"object","name":"UnionRenderers","id":"p1069"},"data":{"type":"map","entries":[["x",{"type":"ndarray","array":{"type":"bytes","data":"ZAAAAMgAAABYAgAAsAQAAA=="},"shape":[4],"dtype":"int32","order":"little"}],["y",{"type":"ndarray","array":{"type":"bytes","data":"0AcAADQIAAB4BQAA6gsAAA=="},"shape":[4],"dtype":"int32","order":"little"}]]}}},"view":{"type":"object","name":"CDSView","id":"p1072","attributes":{"filter":{"type":"object","name":"AllIndices","id":"p1073"}}},"glyph":{"type":"object","name":"Scatter","id":"p1070","attributes":{"size":{"type":"value","value":5},"fill_color":{"type":"value","value":"black"}}}}}],"toolbar":{"type":"object","name":"Toolbar","id":"p1041","attributes":{"tools":[{"id":"p1054"},{"id":"p1055"},{"id":"p1056"},{"type":"object","name":"SaveTool","id":"p1058"},{"id":"p1059"},{"id":"p1060"}]}},"toolbar_location":null,"left":[{"type":"object","name":"LinearAxis","id":"p1049","attributes":{"ticker":{"type":"object","name":"BasicTicker","id":"p1050","attributes":{"mantissas":[1,2,5]}},"formatter":{"type":"object","name":"BasicTickFormatter","id":"p1051"},"axis_label":"y","major_label_policy":{"type":"object","name":"AllLabels","id":"p1052"}}}],"below":[{"type":"object","name":"LinearAxis","id":"p1044","attributes":{"ticker":{"type":"object","name":"BasicTicker","id":"p1045","attributes":{"mantissas":[1,2,5]}},"formatter":{"type":"object","name":"BasicTickFormatter","id":"p1046"},"axis_label":"x","major_label_policy":{"type":"object","name":"AllLabels","id":"p1047"}}}],"center":[{"type":"object","name":"Grid","id":"p1048","attributes":{"axis":{"id":"p1044"}}},{"type":"object","name":"Grid","id":"p1053","attributes":{"dimension":1,"axis":{"id":"p1049"}}}]}},0,0]]}}]}}
    </script>
    <script type="text/javascript">
      (function() {
        const fn = function() {
          Bokeh.safely(function() {
            (function(root) {
              function embed_document(root) {
              const docs_json = document.getElementById('p1076').textContent;
              const render_items = [{"docid":"21276715-a5ae-4bcf-b18f-09fa6758ce03","roots":{"p1063":"e0448a2b-9e27-4fb9-a01d-57a21641536b"},"root_ids":["p1063"]}];
              root.Bokeh.embed.embed_items(docs_json, render_items);
              }
              if (root.Bokeh !== undefined) {
                embed_document(root);
              } else {
                let attempts = 0;
                const timer = setInterval(function(root) {
                  if (root.Bokeh !== undefined) {
                    clearInterval(timer);
                    embed_document(root);
                  } else {
                    attempts++;
                    if (attempts > 100) {
                      clearInterval(timer);
                      console.log("Bokeh: ERROR: Unable to run BokehJS code because BokehJS library is missing");
                    }
                  }
                }, 10, root)
              }
            })(window);
          });
        };
        if (document.readyState != "loading") fn();
        else document.addEventListener("DOMContentLoaded", fn);
      })();
    </script>
  </body>
</html>
```

hamogu:
I don't actually know where clean-up is handled - I consider that out of scope for Sherpa. Sherpa does not specify where to put any file (or to have any file at all), that's in the bokeh layer. 
I understand that that's not a good answer from a CS stand-point, but we're using bokeh as intended here. What we can (and have to!) is to work on the documentation for how to use bokeh by summarizing the most important points and linking to the bokeh docs for more details. For example, a common pattern in bokeh is to specifically set a file name:

https://docs.bokeh.org/en/latest/docs/first_steps/first_steps_7.html

In that case, it'll keep overwriting that one specific file name for every new plot you make.

hamogu:
@DougBurke Can you try a plain bokeh plot (no sherpa involved) and see if that works. If not, then we can fairly confidently say that it's jupyterlab.
```
from bokeh.plotting import figure, show
# prepare some data
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]
# create a new plot with a title and axis labels
p = figure(title="Simple line example", x_axis_label='x', y_axis_label='y')
# add a line renderer with legend and line thickness to the plot
p.line(x, y, legend_label="Temp.", line_width=2)
# show the results
show(p)
```

From: https://docs.bokeh.org/en/latest/docs/first_steps/first_steps_1.html (see there for more details). It might also be specific to jupyerlab, as opposed to the classic Jupyter notebook that I use (though I thought Project Jupyter has merged both projects now that the notebook today is just a specific UI on top of jupyterlab) - you might try https://github.com/bokeh/jupyter_bokeh

DougBurke:
FYI I'm not sure how much time I have today for bokeh.

DougBurke:
Now that #1382 has been merged can we have this rebased, to make it easier to review? [not that I'm likely to do it today]

wmclaugh:
> Plotting and displaying in my browser (chrome on macOS 13) with the Bokeh back end works. Accessing help on the plot object shows a possible discrepancy:
> 
> > > > > plot.set_backend('BokehBackend')
> > > > > plot1 = plot.DataPlot()
> > > > > help(plot1)
> > > > > <... screen output trimmed ...>
> > > > > |  >>> dplot.xlabel = 'length'
> > > > > |  >>> dplot.ylabel = 'data values'
> > > > > |  >>> dplot.plot(**marker=' '**, linestyle='dashed')
> 
> Note that the help example has the marker is set to a space (ascii char 0x20). However, using that marker setting results in the following error message (plot displays):
> 
> > > > > plot1.plot(marker=' ', linestyle='dashed')
> > > > > show(plot.backend.current_fig['all_axes'])
> > > > > ERROR:bokeh.core.validation.check:E-1001 (BAD_COLUMN_NAME): Glyph refers to nonexistent column name. This could either be due to a misspelling or typo, or due to an expected column being missing. : marker=' ' [no close matches] {renderer: GlyphRenderer(id='p1124', ...)}
> 
> Setting the marker to the empty string generates a plot without markers and does not generate the warning.

I  believe this comes from line 1276 of sherpa/plot/__init__.py in Class DataPlot. Since both existing backends (bokeh and pylab) take the empty string for no markers, perhaps the space should be omitted.

    >>> dplot.plot(**marker=' '**, linestyle='dashed')

DougBurke:
I'm re-running the jobs because the error was a bizarre conda issue unrelated to this code.

hamogu:
I'm activating most of the tests for bokeh, but because of #1902 at least three of them fail (and since we parameterize, the actual number of failures is larger). 
Depending on how far we get with this in the next two days, there is probably some say to xfail them (but it's annoying because of the double parameterization using both a pytxt fixture and a `pytest.mark.parameterize` in the same test; alternatively, I'll just put in a `if plot.backend.name == 'BokehBackend': return` (which will have the same effect as an xfail) and note that in #1902 so we don't forget to take it out later.

hamogu:
These are the test failures caused by hard-wiring pyplot into the notebook display for RMFs and images: https://github.com/sherpa/sherpa/actions/runs/6266440316/job/17017429895

As discussed above, that should be addressed by moving those hard-wired pieces to use the plotting backends, but I want to propose that that is not a blocker for this PR (I can set the tests to xfail or use a similar mechanism) - all we are missing is nice output  to the notebook for some cases with `_repr_html_` and there is a fallback text representation that's still pretty good. I think that's exactly the kind of thing that is acceptable behavior difference between our backends.

hamogu:
Here is a question about the UI:

#### First, some background
Bokeh has a different model to update plots and the current state than matplotlib has. There are good reasons for that (for example, if you bokeh plot opens in a new tab in your webbrowser, bokeh needs to write a temporary file. When you call `ui.plot(...., overplot=True) bokeh can't easily know if the browser window is still open, force it to reload etc.). One can also start a separate server process (`bokeh serve`) that keep bi-directional communication, but that brings in a host of other questions. At this stage, I'd rather leave that to power-users and don't default to that for each plot in the sherpa UI.

#### When do we display a plot?
In standard bokeh usage, one simply has to call `show()` to make the plot appear when one is done plotting. That plot can appear in a file, in the browser, or in a notebook cell, depending on the setting.  Remember how we used to call `plt.show()` for matplotlib? This is the same thing.

We then go spoiled by the magic `%matplotlib` in IPython that will automatically keep track of our plots and change them. Well, as I said above, it's not that easy in bokeh; I think they do all they can, bit it's not build into the Jupyter notebook the way matplotlib is.

So, we can either
- Display the plot after every plot command. That's pretty automatic and works fine in the notebook and outside. However, it's annoying if you use repeated `overplot=True` because you'll get a new plot each time - and in the notebook, you end up with a list of plots in the your cell where the top one shows only what you plotted first, and with each overplot, you get another plot appended at the bottom that is like the previous one, just with the overplotted data added.
- Don't display a plot be default and wait for the user to manually call `bokeh.plotting.show(fig)` (or some convenience function we add to the OO and UI, e.g. `sherpa.plot.backend.show()` and `ui.show()`.

It would be so much easier if we could read the users mind and see if their next thought will be to overplot something or if they are done...

If you are playing with this PR, I've currently implemented the first approach for the UI layer and the second one for the OO layer.

That way, the UI "just pops up a plot" in the most common use cases e.g. "ui.plot_fit()", but at the downside of added annoyance for those of us who build up more complex plots with overplot.

Note: In order to save bokeh plots to a permanent file (not just the temporary location it uses to open a webbrowser, a user needs to call `bokeh.io.save(figure)`, so we might want to wrap that into `ui.write()`, too.

DougBurke:
I know when I started looking at using [altair](https://altair-viz.github.io/) as a sherpa backend I hit some of the same pain points regarding the radically different life cycle.

I'd suggest doing minimal work now and labelling this as part of the experimental part that will evolve as we gain more experience with backends other than matplotlib. So if it makes to have different behaviour for UI and OO then we can go for it, but I don't want to set the interfaces they provide in stone.

I should go back to my alrair as it might help see how common it's needs are with bokeh, but that is not going to happen for a while. It doesn't help it was on my laptop that caught fire.

anetasie:
@hamogu I just build the branch and I see the error when importing sherpa. Do I need to update `sherparc` file?

```
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 from sherpa.astro.ui import *

File ~/git/sherpa/sherpa/astro/ui/__init__.py:22
     20 import sherpa.all
     21 import sherpa.astro.all
---> 22 import sherpa.astro.ui.utils
     23 from sherpa.utils import calc_mlr, calc_ftest, rebin, histogram1d, \
     24     histogram2d, gamma, lgam, igamc, igam, incbet, multinormal_pdf, \
     25     multit_pdf
     26 from sherpa.data import Data1D, Data1DInt, Data2D, Data2DInt

File ~/git/sherpa/sherpa/astro/ui/utils.py:30
     26 import warnings
     28 import numpy
---> 30 import sherpa.ui.utils
     31 from sherpa.astro.instrument import create_arf, create_delta_rmf, \
     32     create_non_delta_rmf, has_pha_response
     33 from sherpa.ui.utils import _check_type, _check_str_type, _is_str

File ~/git/sherpa/sherpa/ui/__init__.py:21
      1 #
      2 #  Copyright (C) 2007, 2018  Smithsonian Astrophysical Observatory
      3 #
   (...)
     17 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
     18 #
     20 import sherpa.all
---> 21 import sherpa.ui.utils
     22 from sherpa.utils import calc_mlr, calc_ftest, rebin, histogram1d, \
     23     histogram2d, gamma, lgam, igamc, igam, incbet, multinormal_pdf, \
     24     multit_pdf
     25 from sherpa.data import Data1D, Data1DInt, Data2D, Data2DInt

File ~/git/sherpa/sherpa/ui/utils.py:43
     40 from sherpa.models.model import Model, SimulFitModel
     41 from sherpa.models.template import add_interpolator, create_template_model, \
     42     reset_interpolators
---> 43 from sherpa.plot import set_backend
     44 from sherpa.utils import SherpaFloat, NoNewAttributesAfterInit, \
     45     export_method, send_to_pager
     46 from sherpa.utils.err import ArgumentErr, ArgumentTypeErr, \
     47     DataErr, IdentifierErr, IOErr, ModelErr, ParameterErr, PlotErr, \
     48     SessionErr

File ~/git/sherpa/sherpa/plot/__init__.py:44
     42 from sherpa import get_config
     43 from sherpa.utils.err import ArgumentTypeErr, IdentifierErr
---> 44 from sherpa.plot.backends import BaseBackend, BasicBackend, PLOT_BACKENDS
     45 # PLOT_BACKENDS only contains backends in modules that are imported successfully
     46 # but modules are not discovered by itself. Entrypoints would solve this problem
     47 # but the current implementation does not have this capability.
     48 # See docstring of sherpa.plot.backends.MetaBaseBackend for details.
     49 #
     50 for name in ["pylab", "pylab_area", "bokeh"]:

ImportError: cannot import name 'BasicBackend' from 'sherpa.plot.backends' (/Users/aneta/git/sherpa/sherpa/plot/backends.py)

In [3]: sherpa.__version__
Out[3]: '4.15.1+273.g0709d1d7.dirty'
```

hamogu:
@anetasie Try again. Somehow my editor tried to eat the last 300 lines of that file and - surprise - that means it can't be imported successfully.

anetasie:
@hamogu Thanks. I can import with no errors. Now I have to figure out how to get Bokeh backend which I guess I ned to install first.

anetasie:
I installed bokeh and tried the plots on the command line with `ipython`. The only surprise was that each plot was automatically displayed in a separated tab in my chrome browser. In the documentations of the backends we may want to say when 'bokeh' backend would be primarily used - My guess is that it will be in the notebooks and on the web as the command line use will always use the browser (is this correct assumption?).

I also tried the matplotlib backend in this PR branch and all the standard plots looks good.

anetasie:
I was running a notebook using bokeh backend and it looks good. One needs to get used to the way bokeh interacts, but on Sherpa side it looks ok.

```
fplot = ui.get_fit_plot()
fplot
```
created a bokeh plot, but then I had to use specific show to get the bokeh display. 

```
from bokeh.plotting import show
fplot.dataplot.plot()
show(plot.backend.current_fig)
```

Your docs show this already, so we will have it in RTD. We may add an example notebook at some point to show differences between the two backends, but that's independent on this PR. 

Thanks for this work!

anetasie (approved):

DougBurke:
In reference to https://github.com/sherpa/sherpa/pull/1899#issuecomment-1725774128, using just `jupyter lab` and the suggested code I do indeed see the same behaviour, in that the plot "fails" with a newly-opened tab saying

    Firefox can’t find the file at /tmp/tmpr5mwoc4h.html.

**HOWEVER** I get exactly the same behaviour if I run the same notebook in `jupyter notebook`.

It is unlikely to matter, but this is with

| Package | Version |
| ------- | ------- |
| bokeh | 3.2.2 |
| jupyterlab | 4.0.6 |
| notebook | 7.0.4 |
| jupyter_client | 8.3.0 |
| python | 3.11.4 |

So I don't think I'll be able to provide much real-use testing of this.
| jupyter_core | 5.3.1 |

docs/conf.py
```diff
@@ -140,7 +140,8 @@ def get_xsversion(self):
 ]
 # Imported from sphinx_astropy so we don't have to maintain the list
 # of servers
-# intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
+# but adding to the astropy list.
+intersphinx_mapping['bokeh'] = ('https://docs.bokeh.org/en/latest/', None)
```
DougBurke, line 145: So we don't need the "python" mappnig rule then I guess?

docs/models/index.rst
```diff
@@ -427,6 +427,42 @@ the model to be fit would be given by an expression like::
 
    >>> mdl = g1 + g2 + 0 * sep
 
+Complex functional relationships
```
DougBurke, line 430: Why are we getting this change (and the other changes in this file) in this PR?

docs/plots/bokeh_backend.rst
```diff
@@ -0,0 +1,91 @@
+************************************
+The sherpa.plot.bokeh_backend module
+************************************
+
+The :term:`bokeh` plotting module is not integrated as tightly into
+IPython and Jupyter notebooks as :term:`matplotlib` is, so there is a little
+more setup required to make this work well.
+
```
DougBurke, line 8: An issue I have with the above is that this assumes that this backend is only used with notebooks. This may in fact be the case, but if show we should probably be explicit about this in this paragraph.

hamogu:
The reasons for the test failure is explained in https://github.com/sherpa/sherpa/pull/1899#issuecomment-1730228570
I'm working on a new PR to address that for the pyplot interface and then this PR can be rebased on top of that. Alternatively, I could just xfail them here, but it would be better to make actual plots for the representation in bokeh.

docs/plots/bokeh_backend.rst
hamogu: I can re-order. There is a section explicitly about the IPython terminal below that. My guess is that's is **mostly** going to be used with notebooks, and that's why I address that first, but I can re-order those sections.

sherpa/plot/__init__.py
```diff
@@ -79,7 +79,7 @@
 # might contain e.g. matplotlib specific defaults that are not applicable when other
 # backends are used.
 # Thus, we currently initialize them with the BasicBackend that has only the
-# backend-independent (i.e. those that works for any backend) defaults set.
+# backend-independent (i.e. those that works for any backend) defaults set.``
```
DougBurke, line 82:
```suggestion
# backend-independent (i.e. those that works for any backend) defaults set.
```

I assume these "closing quotes" are not wanted.

docs/models/index.rst
hamogu: No. That's from the already merged #1653. Not sure why GH shows this as new code here. But I will probably reabse / squash commits a couple more times, so I hope that problem goes away by itself.

sherpa/plot/backends.py
```diff
@@ -898,6 +905,12 @@ def hline(self, y, *,
         .. warning::
            No output will be produced by this backend, since the implementation
```
DougBurke, line 906: Above you have changed some `warning` sections to say `function` rather than `backend`; should this one also be changed?

docs/conf.py
hamogu: On top of the file, we do `from sphinx_astropy.conf.v1 import intersphinx_mapping`. We don't need the Python rule, because astropy uses that, too, and thus it's already in `sphinx_astropy.conf.v1.intersphinx_mapping` together with matplotlib, numpy, ans astropy itself. However, astropy does not use bokeh, and thus bokeh is not part of what we import, we have to add it to the dictionary ourselves.

notebooks/ExamplePlots.ipynb
```diff
@@ -11,7 +11,1066 @@
   },
```
hamogu, line 11: I must has accidentially committed changes to this notebook. That was not intentional. I've run it a couple of times to check that I did not mess up the matplotlib plotting but there should be no change in this PR. I'll have to google some git-fu to reverse it out again.

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+        yerr = yerr if yerrorbars else None
+
+        try:
+            color = objs[0].get_color()
+        except (AttributeError, IndexError):
+            pass
+
+        # Do not draw a line connecting the points
+        #
+        # Unlike plot, using errorbar for both cases.
```
DougBurke, line 313: Is this comment still true (it's true for matplotlib but I don't think is a requirement for al backends but I have not really paged in all this code)?

sherpa/plot/__init__.py
wmclaugh, line 82:
```suggestion
# backend-independent (i.e. those that work for any backend) defaults set.``
```

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+        ----------
+        line :
+            The line to change
+        linecolor, linestyle, linewidth : optional
+            The attribute value or None.
+        """
+
+        def set(label, newval):
+            func = getattr(line, 'set_' + label)
+            func(newval)
```
DougBurke, line 351:
```suggestion
```

I don't think we use `set` here, unlike in matplotlib (or the old matplotlib code,I haven't checked recently).

sherpa/plot/backends.py
hamogu: The reason I changed them below is that functions are inherited by BokehBackend, so a text that says "No output will be produced by this backend" is technically wrong - the bokeh backend produces output, just not with that function. In this particular case (hline and vline) all backends we have implement their own version, so I did not have to do that. Still, it might be useful to revisit the wording and come up with something that works for both cases. It's one of those things that you only see once you write a second backend.

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+        x0 : array-like or scalar number
+            lower bin boundary values
+        x1 : array-like or scalar number
+            upper bin boundary values
+        y : array-like or scalar number
+            y values, same dimension as `x0`.
+        {kwargs}
+        """
+        if linecolor is not None:
+            logger.warning("The linecolor attribute ({}) is unused.".format(linecolor))
```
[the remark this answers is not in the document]
hamogu: In fact, we've had those linecolor warnings for years, I'd like to just remove them and fail on using "linecolor" in all backends. However, that's for a different PR.

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+        axes = self.setup_axes(overcontour, clearwindow)
+
+        # Set up the axes
+        if not overcontour:
+            self.setup_plot(axes, title, xlabel, ylabel, xlog=xlog, ylog=ylog)
+
+        if levels is None:
+            levels = 7
+        if isinstance(levels, int):
+            # Matplotlib has defaults for levels and sherpa can pass in "None" to mean
```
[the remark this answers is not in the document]
hamogu: Maybe we don't, but I'm assuming that matplotlib will be the default backend and the one developers are most familiar with for the forseeable future, so referencing back to that did seem to make sense when I wrote this.

hamogu:
@DougBurke : For you not being able to see something: Can you look at the second half of Bryan's answer to a similar problem: https://discourse.bokeh.org/t/using-bokeh-in-spyder/10548
You already diagnosed that it's likely some permissions issue where firefox for some reason is not allowed to read the file automatically. Can you try to set `bokeh.io.output_file('some/other/path/firefox/can/read.html')`? Maybe it just does not have access to your /tmp folder?

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+        """
+        index = self._index_axis(row, col)
+        if index is not None and not clearaxes:
+            # We found the right subplot, now make it the current one
+            self.current_axis = self.current_fig[index]
+            return
+
+        # We found the subplot, but we want to replace it with a new one
+        if index is not None and clearaxes:
+            self.current_fig.children.pop(index)
```
DougBurke, line 663: So, pylint is complainnig to me that "Instance of 'List' has no 'pop' member". This strikes me as odd, because a list does have pop, but perhaps whatever `children` is does not.

There is some strangeness here, since in line 658 above you have `current_fig` as being a sequence (i.e has an `[index]` element) but here `current_fig` also has a `children` argument. Both are entirely possible in the wild world of Python, but it's not clear whether we test this (I don't entirely trust the "line not covered by test" lines from codecov, but I can't realistically test this at the moment to check [*]).

[*] I am old school and stick an `assert False` at the location and re-test

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+                    basenum = index
+                else:
+                    self.current_fig[index].x_range = \
+                        self.current_fig[basenum].x_range
+
+        plotnum = self._index_axis(row, col)
+        if plotnum is None:
+            raise ArgumentErr("No plot at row {} and column {}." +
+                              "Use create=True to create an entirely new, " +
+                              "complete grid of plots.".format(row, col)) from None
```
DougBurke, line 745: The use of "blah {} {}" + ... + "text".format(row,col) doesn't actually pass the row/col arguments to the first bit. So this is one where a f-string change is definitely appropriate.

```suggestion
            raise ArgumentErr(f"No plot at row {row} and column {col}."
                              "Use create=True to create an entirely new, "
                              "complete grid of plots.") from None
```

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,829 @@
...
+            return None
+
+        if summary is None:
+            summary = type(data).__name__
+
+        ls = [formatting.html_svg(image, summary)]
+        return formatting.html_from_sections(data, ls)
+
+    as_html_plot = partialmethod(as_html_plot_or_contour, func="plot")
+    as_html_contour = partialmethod(as_html_plot_or_contour, func="contour")
```
DougBurke, line 816: pylint is complaining to me about both lines (ie same reason for both); there's two items but I'm sure they are essentially  the same issue

- W4903: Using deprecated argument func of method partialmethod() (deprecated-argument)
- E1124: Argument 'func' passed by position and keyword in constructor call (redundant-keyword-arg)

I can't see any recent documentation to say when this behaviour changed

sherpa/plot/testing.py
```diff
@@ -0,0 +1,116 @@
+#
```
DougBurke, line 1: So, I have stared deep into the heart of our build system and it has stared back at me. And won. So I am always "interested" in things that could have some "interesting" consequences for our builds. I believe (too lazy to check) that we include the `test_*py` files in the MANIFEST (either directly, indirectly, or magically) and maybe we get this file for free (e.g. we - or whatever system decides, uses 'test*py' or '*py' as the regexp). Do you know how this works?

sherpa/plot/backends.py
DougBurke: I don't think it's something important (i.e. does not need to be resolved for this PR). It's just the sort of thing you notice when scanning similar changes.

sherpa/plot/bokeh_backend.py
hamogu: I think pylist is misunderstanding here. `func` is the keyword name in `as_html_plot_or_contour`, not an argument to `partialmethod`. Maybe it was a long time ago, but the fact that this works shows that Python interprets this way way I intended. I can rename the argument in `as_html_plot_or_contour` though, just to make pylint happy.

DougBurke:
In reference to https://github.com/sherpa/sherpa/pull/1899#issuecomment-1734362708 - it's probably because I'm using ubuntu, and hence firefox via a snap (so would be similar possibilities about flatpak version) which probably can't see /tmp/ ...

This is what clued me in to the problem, https://askubuntu.com/questions/1389798/how-to-correctly-configure-snapd-firefox-to-open-local-html-file-generated-by but it's not realy helped me with a solution.

This is much older, and so may be wrong now, but it's basically saying I have to bind mount the /tmp directory which is just insane (I can understand the snap restriction, and that this problem is not bokeh or snap's "fault", but it is a UX fail...

Now I just need to know what I can point `bokeh.io.output_file('some/other/path/firefox/can/read.html')` to... I was able to

    mkdir ~/tmp

and then

    bokeh.io.output_file('/home/dburke/tmp/tmp_bokeh.html')

and the bokeh-only code created a plot.

DougBurke:
So is the idea that we get #1906 merged, so this can then be merged into this (or rebased,or whatever) and everything will be copaceptic ;-)

sherpa/plot/testing.py
hamogu: This is not a "test_.py" file, it's part of the real deal; it just happens to have a filename that also starts with T E S T as first four letters. If it was not included, the tests woudl fail, but I've neve added other source files (e.g. `bokeh_backend.py`) manually to MANIFEST and we've always gotten them, so - while I don't understand our build system fully - I'm ptryy sure this is save?

sherpa/plot/bokeh_backend.py
hamogu: I did stick in `assert False` and it does make some tests fail, so this code is used.

sherpa/plot/testing.py
DougBurke: For some reason I thought it was `sherpa/plot/tests/testing.py`...

hamogu: I had it there originally, but I had trouble importing it. Maybe because some test dirs don't have `__init__.py`? We can move it later if you want?

sherpa/plot/bokeh_backend.py
hamogu: I think I put the `assert False` in the wrong location and there is a "children" missing above, but in a code path only triggered to multi-panel plots where individual panels are re-created. We don't have tests to cover use (or use cases) and I don't trust the pylab backend either for that...

sherpa/plot/bokeh_backend.py
```diff
@@ -0,0 +1,854 @@
...
+        x0 : array-like
+            independent axis in the first dimension
+        x1 : array-like
+            independent axis in the second dimension
+        y : array-like, with shape (len(x0), len(x1))
+            dependent axis (i.e. image values) in 2D
+            with shape (len(x0), len(x1))
+        {kwargs}
+        """
+        print(x0, x1, y, aspect, kwargs)
```
DougBurke, line 634:
```suggestion
```

DougBurke (approved):
THere's a print line I've noted but other than that lgtm.

hamogu:
Removing that spare print line will also restart the tests...

DougBurke:
Again the test failure is not related to this code (*).

---

* technically there could be some relation, but I do not see how, and I feel I've seen this problem on other PRs but can't guarantee it.

docs/plots/bokeh_backend.rst
```diff
@@ -0,0 +1,118 @@
+************************************
+The sherpa.plot.bokeh_backend module
+************************************
+
+:term:`bokeh` is a plotting library that can be used to create interactive
+plots, which are rendered in a webbrowser. Those plots allow various options to zoom,
+pan, and scroll out of the box. Many additional interactions can be added by the
+user later, and we also plan make Sherpa use more features of this library
```
wmclaugh, line 8:
```suggestion
user later, and we also plan to make Sherpa use more features of this library
```

docs/plots/bokeh_backend.rst
```diff
@@ -0,0 +1,118 @@
...
+for details. In particular:
+
+  - Bokeh plots cannot easily be exported to png or pdf. They are always
+    rendered in javasript. Bokeh does
+    `offer a mechanism to export plots to static images <https://docs.bokeh.org/en/latest/docs/user_guide/output/export.html>`_,
+    but it requires additional software and drivers.
+  - For display, bokeh saves temporary files that are then automatically opened in a
+    webbrowser. In some linux systems, browsers are not allowed to open
+    files in the local ``\temp`` directory for security reasons, so the user
+    as to manually change the location the bokeh plot is written to with
```
wmclaugh, line 93:
```suggestion
    has to manually change the location the bokeh plot is written to with
```

docs/plots/bokeh_backend.rst
```diff
@@ -0,0 +1,118 @@
...
+    `bokeh.io.output_file`.
+  - Once displayed, bokeh has not control over the webbrowser any longer.
```
wmclaugh, line 95:
```suggestion
  - Once displayed, bokeh no longer has control over the webbrowser.
```

hamogu:
Note that the test that fails here is "Pip CI/Linux Build (w/o XSPEC, Python 3.11) (https://github.com/sherpa/sherpa/actions/runs/6341413247/job/17224999177?pr=1899). That's the only one that I've added bokeh, to (I think) because I did not want to drive up our testing times for something that's still a little experimental. Also, bokeh is pretty independent of Sherpa, in the sense that we don't build C code are binaries that have anything to do with bokeh.

So, it seemed like a reasonable compromise for the time being.

I have a solution to that failure in https://github.com/hamogu/sherpa/pull/6, where  "Pip CI/Linux Build (w/o XSPEC, Python 3.11)" passes: https://github.com/hamogu/sherpa/actions/runs/6385025927/job/17328888999?pr=6

I will move that code into this PR in the next few minutes, but we will now get a different failure in some other build that has noting to do with this PR, but that @DougBurke reported this morning and that probably has something to do with some dependency being upgraded on conda-forge. (We might have to find with one and pin it for the time being).

wmclaugh (approved):

Issue #1903, opened 2023-09-21
are the CI:arch builds broken


DougBurke:
I recently had a PR labelled CI:arch fail with the following (the PR had nothing to do with parameters). I don't know if this is some oddity from merging #1653 (but I didn't thank that PR had included those changes, but I think this test was added in that PR so ..).

Leaving this here so we can refer to it if it happens again

```
_____________________ test_binop_string_with_custom_ufunc ______________________

    def test_binop_string_with_custom_ufunc():
        '''Repeat the previous test, but with a custom ufunc.'''
        def func(a, b):
            return a + b
    
        uf = np.frompyfunc(func, nin=2, nout=1)
        p, p2 = setUp_composite()
        comp = uf(p, p2)
>       assert repr(comp) == "<BinaryOpParameter 'func(model.p, model.p2)'>"
E       assert "<BinaryOpPar..., model.p2)'>" == "<BinaryOpPar..., model.p2)'>"
E         - <BinaryOpParameter 'func(model.p, model.p2)'>
E         ?                     ^^^^
E         + <BinaryOpParameter '?(model.p, model.p2)'>
E         ?                     ^

sherpa/models/tests/test_parameter.py:237: AssertionError

_____________________ test_binop_string_with_custom_ufunc ______________________

    def test_binop_string_with_custom_ufunc():
        '''Repeat the previous test, but with a custom ufunc.'''
        def func(a, b):
            return a + b
    
        uf = np.frompyfunc(func, nin=2, nout=1)
        p, p2 = setUp_composite()
        comp = uf(p, p2)
>       assert repr(comp) == "<BinaryOpParameter 'func(model.p, model.p2)'>"
E       assert "<BinaryOpPar..., model.p2)'>" == "<BinaryOpPar..., model.p2)'>"
E         - <BinaryOpParameter 'func(model.p, model.p2)'>
E         ?                     ^^^^
E         + <BinaryOpParameter '?(model.p, model.p2)'>
E         ?                     ^

sherpa/models/tests/test_parameter.py:237: AssertionError
```

hamogu:
This test tests a custom ufunc, so it relies on numpy to give of the name of that function. If numpy does not provide it, we can't get it. I think the solution is to do something like `self.arg.fullname.replace('?', 'user_func') and special case that in the test on the aarch64 architecture or just xfail the test on arch64. Either is easy to do, but I'll wait until we see that failure at least one more time before I implement that.

DougBurke:
[...]

PR #1904, opened 2023-09-22
Upgrade base image for aarch64 testing

hamogu:
In main, we are getting Debian "bullseye" packages, which contain numpy 1.19.
I wonder if, with a new baseimage and a newer debian version, issue #1903 might resolve itself.

### DIFF
--- a/.github/workflows/ci-pip-arch.yml
+++ b/.github/workflows/ci-pip-arch.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     if: (github.repository == 'sherpa/sherpa' && (github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'CI arch')))
 
     strategy:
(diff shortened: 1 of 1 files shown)


DougBurke:
I created the "arch" build based on the astropy version at the time (I think) and it has seen little update since then, so not too surprising it's out of date. I think it's only been @Marie-Terrell in #1706 who's updated it. I know I've basically forgotten about it (other than to actually run it).

hamogu:
We might decide to take this PR, even if it does not solve #1903, simply to not fall too far behind and get random failures later.

DougBurke:
It's not the fastest build, that's for sure....

hamogu:
Well, this passed. So, that numpy bug in arch64 has been resolved some time between numpy 1.19 and 1.24. Given that the number of Sherpa users we have on Linux ARM 94 is likely very low (if any), I suggest we just take this change to make CI pass again and don't bother to determine or document the exact minimum version of numpy that would work an aarch64 for that one test.

DougBurke (approved):
:shipit:

PR #1906, opened 2023-09-25
Remove hardcoded matplotlib for _repr_html_ of RMF and DataImage2D

### LINKED ISSUE #1385: Add a plot_rmf command to the UI layer?
For the notebook rich-display work I added a way to visualize a RMF - we pick 5 energies spaced across the response (logarithmically spaced) and then show the response to a delta function at those energies. This code is locked away within the jupyter code.

I think we should look at moving it to to a plot_rmf command - so all users can get it (and then the jupyter code can just call it).

As an example, here's a notebook display for an "interesting" RMF I wanted to view whichi I had to start a jupyter lab instance just to see):

screenshot

hamogu:
# Summary

Implement RMFPlot and DataIMGPlot classes and associated UI functions (plot_rmf, get_rmf_plot). Fix #1385

## Motivation
To make the `_repr_html_` work with other backends, we can't have matplotlib hardcoded in the code that generates those representations, see https://github.com/sherpa/sherpa/pull/1899#issuecomment-1730228570

## Details
Sherpa in generally has Plot objects tasked with producing visual information about a Data object (e.g. a DataPHA or an ARF). They have `prepare` method, that is independent of the actual presentation; it just parses the data in the Data objects into some form more suitable for display (for example, it sets up the units and labels of axes, scales the data, calculates ratios of data and model on the same grid for a ratio plot etc. Then, there is `plot` method (called `contour` for some Plot objects), that actually passes that information to the plotting backend. 
The same mechanism is used for the plots of the rich display in the Jupyter notebook, except in two cases where sherpa (for mostly historical reasons) does not have plot classes associated with the data classes (RMF and DataIMG). For those, and those only, the matplotlib plotting commands were hardcoded in the rich display settings bypassing the use of the plotting backend.

This PR add Plot classes for RMF and DataIMG and uses them for the rich display.

- In particular for the RMF, we issue #1385 to add a plot class. That's addressed in this PR.
- This PR will greatly simplify the testing of #1899 (not the implementation) because it will allow me to reuse the existing tests for the pyplot backends when testing bokeh with minimal changes to the tests, since they will both produce comparable rich-display output.


sherpa/astro/plot.py
```diff
@@ -476,6 +458,120 @@ def prepare(self, arf, data=None):
...
+    """
+
+    histo_prefs = shplot.basicbackend.get_model_histo_defaults()
+
+    # TODO: Make that a plot preference
+    # How many monochromatic lines to use
+    n_lines = 5
+
+    # Because this derived from NoNewAttributesAfterInit we need to
+    # make the attribute here it cna be used in prepare
```
DougBurke, line 483:
```suggestion
    # make the attribute here it can be used in prepare
```

My typing skillz strike again....

sherpa/astro/ui/utils.py
```diff
@@ -11193,6 +11194,67 @@ def get_arf_plot(self, id=None, resp_id=None, recalc=True):
...
         return plotobj
 
+
+    def get_rmf_plot(self, id=None, resp_id=None, recalc=True):
+        """Return the data used by plot_arf.
+
+        Parameters
+        ----------
+        id : int or str, optional
+           The data set with an ARF. If not given then the default
```
DougBurke, line 11204:
```suggestion
           The data set with a RMF. If not given then the default
```

sherpa/astro/ui/utils.py
```diff
@@ -11193,6 +11194,67 @@ def get_arf_plot(self, id=None, resp_id=None, recalc=True):
...
+           identifier is used, as returned by `get_default_id`.
+        resp_id : int or str, optional
+           Which ARF to use in the case that multiple ARFs are
```
DougBurke, line 11207:
```suggestion
           Which RMF to use in the case that multiple RMFs are
```

sherpa/astro/ui/utils.py
```diff
@@ -11193,6 +11194,67 @@ def get_arf_plot(self, id=None, resp_id=None, recalc=True):
         plotobj.prepare(arf, data)
         return plotobj
 
+
+    def get_rmf_plot(self, id=None, resp_id=None, recalc=True):
+        """Return the data used by plot_arf.
+
```
DougBurke, line 11200:
```suggestion

        .. versionadded:: 4.16.0

```

Not sure I've got the indenttion/spacing correct

sherpa/astro/ui/utils.py
```diff
@@ -12105,6 +12167,78 @@ def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
...
                    **kwargs)
 
+
+    def plot_rmf(self, id=None, resp_id=None, replot=False, overplot=False,
+                 clearwindow=True, **kwargs):
+        """Plot the RMF associated with a data set.
+
+        Display the energy redistribution from the RMF
+        component of a PHA data set.
+
```
DougBurke, line 12177:
```suggestion

        .. vresionadded:: 4.16.0
        
```

Again, is indentation correct?

sherpa/astro/ui/utils.py
```diff
@@ -12105,6 +12167,78 @@ def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
         self._plot(plotobj, overplot=overplot, clearwindow=clearwindow,
                    **kwargs)
 
+
+    def plot_rmf(self, id=None, resp_id=None, replot=False, overplot=False,
+                 clearwindow=True, **kwargs):
+        """Plot the RMF associated with a data set.
+
+        Display the energy redistribution from the RMF
+        component of a PHA data set.
```
DougBurke, line 12176: This should probably say something about how we 'display' a RMF.

sherpa/astro/ui/utils.py
```diff
@@ -12105,6 +12167,78 @@ def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
...
+
+        Parameters
+        ----------
+        id : int or str, optional
+           The data set with an ARF. If not given then the default
```
DougBurke, line 12181:
```suggestion
           The data set with a RMF. If not given then the default
```

sherpa/astro/ui/utils.py
```diff
@@ -12105,6 +12167,78 @@ def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
...
+           identifier is used, as returned by `get_default_id`.
+        resp_id : int or str, optional
+           Which ARF to use in the case that multiple ARFs are
```
DougBurke, line 12184:
```suggestion
           Which RMF to use in the case that multiple RMFs are
```

sherpa/astro/ui/utils.py
DougBurke: `versionadded` also needs to be spelled correctly.

sherpa/astro/ui/utils.py
```diff
@@ -12105,6 +12167,78 @@ def plot_arf(self, id=None, resp_id=None, replot=False, overplot=False,
...
+        the RMF from data set 2:
+
+        >>> plot_rmf(1)
+        >>> plot_rmf(2, overplot=True)
+
+        Plot the RMFs labelled "rmf1" and "rmf2" for the
+        "src" data set:
+
+        >>> plot_arf("src", "rmf1")
+        >>> plot_arf("src", "rmf2", overplot=True)
```
DougBurke, line 12224:
```suggestion
        >>> plot_rmf("src", "rmf1")
        >>> plot_rmf("src", "rmf2", overplot=True)
```

although we normally just use 1, 2, ... when we have multiple responses (you should be able to use strings, I just haven't checked as there's a nagging worry there's some internal location where we assume an integer key for the responses, but this is all assumption)

DougBurke:
Two issues:

A) `plot_rmf` returns "something"

```
In [30]: ui.plot_arf()

In [31]: ui.plot_rmf()
{}
```

B) axis scaling

I had hard coded the plots to be log-log but we now have plot options, but the default RMF plot uses log axes - e.g.

sherpa

but the plot options for RMF are to use linear axes:

```
In [36]: ui.get_rmf_plot().histo_prefs
Out[36]: 
{'xlog': False,
 'ylog': False,
 'label': None,
 'xerrorbars': False,
 'yerrorbars': False,
 'color': None,
 'linestyle': 'solid',
 'linewidth': None,
 'marker': 'None',
 'alpha': None,
 'markerfacecolor': None,
 'markersize': None,
 'ecolor': None,
 'capsize': None}
```

DougBurke:
I will say that I can go crazy and say

```
In [37]: ui.plot("fit", "resid", "arf", "rmf")
{}
```

and get

sherpa

[this is because of some of the plotting behind the scenes work I've made in the past year for the UI layer]

DougBurke:
I know it's still marked draft, but thought it would be useful to check out.

hamogu:
@DougBurke I appreciate the fast feedback. It's a "draft" because the image display is still missing. I'll look at all your comments, but I have to see if I accept them on the GH interface, or work them in with my local development.

sherpa/astro/ui/utils.py
hamogu: That is - as you see - copied from plot_arf, which uses "arf1" and "arf2" labels... so if it's questionable here, it's questionable there.

sherpa/astro/ui/utils.py
hamogu, line 12176:
```suggestion
        component of a PHA data set. This plot selects a few specific energies
        and generates a plot with several histograms that show the energy
        redistribution for those specific energies.
```

sherpa/astro/ui/utils.py
hamogu: Added your comment, so that I could add it to batch directly...

hamogu:
> plot_rmf returns "something"

No... much simpler... I left a print statement in...

hamogu:
> but the plot options for RMF are to use linear axes:

The preferences strike again... It's unsustainable to define the default for every type of plot in the plotting backends instead of in the plot objects. Sad! Needs to be addressed! - but not in this PR and not before 4.16. It's listed in #1375 as step 2B.

hamogu:
I've also moved iamge plotting. There is a lot of dublicated code and boilerplate functions that need to be added for each plot type. Step 2 in #1375 will address some, but not all of it. For now, we just follow the approach of the existing code and copy and paste the lines of code; that should not be controversial and not hard to review, even though it's much longer thna it has to be.

DougBurke:
I was going to ask why the RTD build keeps failing but this time the build log was actually useful and reported

Possible hints:
* ValueError: not enough values to unpack (expected 2, got 1)
* AttributeError: module 'sherpa.astro.plot' has no attribute 'ImagePlot'
* KeyError: 'ImagePlot'
* ModuleNotFoundError: No module named 'sherpa.astro.plot.ImagePlot'; 'sherpa.astro.plot' is not a package
* ModuleNotFoundError: No module named 'ImagePlot'

hamogu:
Tests passed before I force pushed again just now. All that's missing is to make RTD pass and that's just some stupid attribute that was documented both in the "Attributes" section and with a docstring in the code. So, the second half of this is ready for review by @DougBurke (the first half is unchanged compared to your review last night).

hamogu:
Still one short RTD warning on the label. Please let that stop you from reviewing it. There will almost certainly be changes with respect to the review and I'll fold that fix in there.

Note: Why don't I just check locally that it's fixed? Because I'm working on my bokeh PR, which has a number of new files that are not quite ready yet and thus not committed. So, every time I switch back to this, I need to stash them, then clean the repro and all generated API docs (or I get ton's of warnings about documents not included in toc tree from sphinx) and then rebuild. It just all takes a while to run and do, so II would appreciate it if you can review now with RTD still failing.

DougBurke:
I actually have a sherpa directory which contains multiple check outs of sherpa (called wonderful things like sherpa-work, sherpa-temp, sherpa-pr, sherpa-xspec) so I can have multiple branches open (I have a conda environment for each directory, using the same name so it's easier to match). I should use direnv but ....

This is also not helpful for this PR!

sherpa/astro/plot.py
```diff
@@ -473,16 +473,21 @@ class RMFPlot(shplot.HistogramPlot):
...
     n_lines = 5
 
     # Because this derived from NoNewAttributesAfterInit we need to
-    # make the attribute here it cna be used in prepare
+    # make the attribute here it can be used in prepare
     labels = None
 
+
+    def _merge_settings(self, kwargs):
+        return {**self.rmf_plot_prefs, **kwargs}
```
DougBurke, line 491: I guess I can get back at you and say this can be (I think)

```suggestion
        return self.rmf_plot_prefs | kwargs
```

although is there actually a subtle difference between the two since the original version creates a new dictionary whereas `a | b` will return a if b is empty (and so changes to the return value would then get passed back to `self.rmf_plot_prefs` and could that be an issue. hmm.

sherpa/plot/__init__.py
```diff
@@ -511,6 +511,58 @@ def point(self, x, y, overplot=True, clearwindow=False, **kwargs):
                      **opts)
 
 
+class Image(NoNewAttributesAfterInit):
+    """Base class for image plots
```
DougBurke, line 515: [not really a comment that needs code changes]

One of the reasons I haven't really pushed the "plot image" work is because I feel we haven't come to a conclusion about how to handle 2D data. My feeling is that `Data2D` is for "unstructured" data whereas `DataIMG` is for a 2D matrix (aka an image), **but** this is not entirely born out by the code (e.g. `Data2D` has a `shape` setting), and then there's things that could get moved to `Data2D` - such as image filtering - that is orthogonal to the data representation.

So I am concerned that this design helps to cement this confusion. Fortunately

a) you've labelled it experimental
b) it is only used in one specific place

so it shouldn't break too much if we decide to rework the `Data2D`/`DataIMG` classes in the future.

sherpa/plot/backends.py
```diff
@@ -774,6 +807,20 @@ def as_html_contour2d(self, plot):
                              'min', 'max', 'nloop',
                              'delv', 'fac', 'log'])
 
+    # TODO: Inconsistent. Some plots have as_html_xxx, but others do not.
```
DougBurke, line 810: As I've mentioned before, I do wonder whether we should just have a generic `get_as_html` type routine which takes the name of the "plot type" - e.g. "histogram", "image", ... - rather than having multiple `as_html_xxx` routines (naming is one of the issues here as we already have a `as_html` method with a slighty-different API if memory serves me.

But this is all for a different PR than this one.

hamogu: > As I've mentioned before, I do wonder whether we should just have a generic get_as_html ...
I've suggested one approach in #1375 to this, but that's for a different PR as you say.

DougBurke:
@hamogu - can you update the commit "message" (whatever the first box is called) to match the latest behaviour? I have my own understanding of what the changes are, but it's helpful to know I'm reading things right.

DougBurke:
Just to check - I ran through the three notebooks we have in the repository (with matplotlib) and the results looked the same. Which is the point.

hamogu:
I added detail to the first message in the thread. Let me know if that's sufficient.

DougBurke:
Just tocheck

> Implement an RMFPlot and ImagePlot and associated UI functions (plot_rmf, get_rmf_plot). Fix https://github.com/sherpa/sherpa/issues/1385

should probably be

> Implement RMFPlot and DataIMGPlot classes and associated UI functions (plot_rmf, get_rmf_plot). Fix https://github.com/sherpa/sherpa/issues/1385

DougBurke:
I don't really have anything more to say about the code. I'd just like to see the RTD build go green.

sherpa/astro/plot.py
hamogu: When we redo the way preferences work, we'll address that.

hamogu:
Tada.

DougBurke:
I'm re-running the macOS job as the failure it reported appears to be a not-my-sherpa problem .

DougBurke (approved):
Ta

PR #1909, opened 2023-09-27
Add security warnings to restore functions

hamogu:
Those functions import binary code, and we should be clear about risks involved.


anetasie (approved):

anetasie:
I agree we should include a Warning. For the future release - should we abandon binary files?

DougBurke:
Removing support for the `save`/`restore` commands could remove some internal code, but not a huge amount. I think there is a place for these commands as long as you are aware of the risks (eg if I only ever use it myself and never send/receive files then I'm okay), but it is something to think about long term.

PR #1911, opened 2023-09-28
IO: add missing import for pyfits backend

DougBurke:
# Summary

Ensure a corner-case error condition when writing out responses fails with the correct error.

# Details

I think I missed this in #1836. It would be triggered in a corner case, which is why the tests did not pick it up (as I didn't add any tests for this particular corner case).

Found working on #1869 but it is completely separate to that.

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -47,7 +47,7 @@ from astropy.io import fits
 from astropy.io.fits.column import _VLF
 from astropy.table import Table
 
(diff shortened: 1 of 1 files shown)


DougBurke:
@wmclaugh - this is a bugfix from #1836 - it would only affects users of the pyfits/backend and is a corner case, but it is a simple fix.

wmclaugh (approved):
I see the added ArgumentTypeErr being raised in #1836 for both pyfits and crates backends. The import was added to crates in #1836.

PR #1912, opened 2023-09-28
Warn when reading in a multi-matrix rmf

DougBurke:
# Summary

Ensure that attempts to read in a multi-matrix RMF will warn the user that they are not yet supported.

# Details

This is a very-slimmed down #1869, concentrating on

- adding a test case
- raising a warning (technically an error log message) when the RMF contains multiple MATRIX blocks

so that user's aren't going to accidentally get the wrong result. It is tempered slightly by the fact we don't have any real-world test data to use as a test!

The idea would be we can supply "contrib" code for the 4.16 release which will let users use these files, but without the concerns I have over the code in #1869 (too many changes).

As part of this several error messages have changed - that is, when trying to read in an invalid OGIP response file the error you now get may be slightly different (and the difference depends on the backend). We can improve this somewhat, but

a) that's for later
b) there's always going to be some case you get backend-specific behaviour

This PR starts to use the "generic FITS handling" code added with the XSPEC table PR #1861 but I am attempting to minimize this PR so it doesn't do much (e.g. we want to move the code into a separate module, add better optional behaviour, ... whic we can do later). 

I have run the tests with both the pyfits and crates backends (only the former gets exercised by the CI runs).

# Example

This is from a RMF I have hacked together - I have checked that I get the "expected" output when I pass it to XSPEC (i.e. that it is really a multi-matrix RMF andI understand how it's behaving):

```
>>> rmf = ui.unpack_rmf("multi.rmf")
ERROR: RMF in multi.rmf contains 2 MATRIX blocks; Sherpa only uses the first block!
```

### DIFF
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -18,16 +18,20 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from collections import defaultdict
 import logging
 import os.path
 
 import numpy
 
 from pycrates import CrateDataset, CrateKey, CrateData, TABLECrate, \
-    IMAGECrate, WCSTANTransform, RMFCrateDataset, PHACrateDataset
-import pycrates
-import cxcdm  # work around missing TLMIN support in crates CIAO 4.15
+    IMAGECrate, WCSTANTransform, RMFCrateDataset, \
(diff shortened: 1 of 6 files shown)


hamogu (approved):
It's good to warn.

anetasie:
@DougBurke I agree, it will be good to warn the users.

anetasie (approved):

Issue #1790, opened 2023-06-05
Drop oldest-supported-numpy from build requirements when numpy 1.25 is released


hamogu:
This is based on https://github.com/astropy/astropy/issues/14915

The build system is not necessarily my area of expertise, but if other, similar packages (in this case astropy) work out how to simplify things, I figured we might at least look at that and possibly follow - hence this issue to remember.

The astropy infrastructure and release team has concluded that "once 1.25 is out we can put it as minimum version in our build requirements," and drop oldest-supported-numpy. Since Sherpa uses the exact same mechnism, that should work for us as well.

https://numpy.org/devdocs/release/1.25.0-notes.html#by-default-the-exported-numpy-c-api-is-now-compatible-with-numpy-1-19 (https://github.com/numpy/numpy/pull/23528 and https://github.com/numpy/numpy/pull/23872)

https://numpy.org/devdocs/dev/depending_on_numpy.html#for-downstream-package-authors

hamogu:
Astropy has now a minimum requirement for numpy IN THE BUILDSYSTEM (there is a separate requirement to run it):
```
[build-system]
requires = ["setuptools",
            "setuptools_scm>=6.2",
            "cython==0.29.36",
            "numpy>=1.25,<2",
            "extension-helpers"]
build-backend = "setuptools.build_meta"
```

PR #1916, opened 2023-10-02
BUILD: drop the use of oldest-supported-numpy

DougBurke:
# Summary

Change the build requirements from `oldest-supported-numpy` to `numpy`.

# Details

According to #1790 we no-longer need the oldest-supported-numpy "helper" so let's see if we can remove it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = ["setuptools >= 51.0.0, < 60",
             "wheel",
-            "oldest-supported-numpy",
+            "numpy",
             ]
 build-backend = "setuptools.build_meta"


DougBurke:
Well, that's stopped the trying-to-build-numpy issue (it seems) but now it fails with the always-annoying

    error: [Errno 2] No such file or directory: '/home/runner/work/sherpa/sherpa/build/lib/python3.9/site-packages/stk.so'

which I think is a symptom of 
- #50 
- e.g. see my experiments in 
  - #1546

I don't get this when I try to build (but I also didn't get the original problem, so I may have a different setup). My guess is that I basically always build following 
- #1608 
and so I don't see this sort of problem.

DougBurke:
Hmm. Including this in #1608 doesn't seem to help..

DougBurke:
@hamogu - that's why I've been trying to work on this issue (and failing unsuccessfully) today!

hamogu:
@DougBurke I understand - I just wanted a reference to point to to say "This PR is good, the failure is unrelated".

A different (very short-term) approach would be to find what actually changed in the environment (painful, but possible by comparing CI build logs from last week to those from this week) and pin those requirements. It's not a good solution because the more we pin, the less flexible users are to use sherpa within their environments, but if we are running out of time, then that might be an option for the 4.16 release.

PR #1938, opened 2023-12-14
CI: temporarily remove ds9 tests on macOS

DougBurke:
# Summary

Avoid the DS9 tests on the macOS build to avoid the failure case we currently often, but not always, trigger.

# Details

For the last month or so our `macOS` CI run will often, but not-alwas, fail because of connection errors [*] with DS9.  As well as requiring a re-run (often many re-runs) these failures increase the run-time of the CI run (presumably because there's a timeot that happens because of the connection issues). I have tried to identify what is the problem, as has @Marie-Terrell , but it's hard without direct access to a failing machine. One suggestion is that this only happens with an "old" setup, but it doesn't look like we can force a particular setup (or whatever they are called).

As a work-around this PR changes the `macOS` run so that it does not install ds9, and hence doesn't hit this problem. This is not ideal, but we do have ds9 checks on the linux runs, and I would hope we can `git revert` the second commit soon.

The first commit removes the step that started up Xvfb since it is just a waste of time, since the `pytest-xvfb` plugin starts up Xvfb for us anyway. So this commit should stay.

[*] I think. It's hard to track down, but it's definitely some issue with communicating with DS9.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -117,7 +117,9 @@ jobs:
       run: |
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
-        source .github/scripts/setup_ds9.sh
+        if [ "$RUNNER_OS" != "macOS" ]; then
+          source .github/scripts/setup_ds9.sh
+        fi
         source .github/scripts/setup_xspec.sh
 
     - name: Build Sherpa (install)
@@ -160,6 +162,14 @@ jobs:
         if [ -n "${XSPECVER}" ]; then XSPECTEST="-x -d"; fi
         if [ -n "${FITS}" ] ; then FITSTEST="-f ${FITS}"; fi
         smokevars="${XSPECTEST} ${FITSTEST} -v 3"
+
+        # special case the macOS build as we currently do not
+        # include DS9; as we know the XSPECVER/FITS sittings for
+        # this it is hard-coded
+        if [ "$RUNNER_OS" == "macOS" ]; then
+          smokevars="-x -f ${FITS} -v 3"
+        fi
+
         echo "** smoke test: ${smokevars}"
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -31,8 +31,6 @@ jobs:
             test-data: submodule
             matplotlib-version: 3
             xspec-version: 12.13.1
-            # run xvfb with this display
-            display: :99
 
           - name: Linux Minimum Setup (Python 3.9)
             os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-conda-workflow.yml
```diff
@@ -116,19 +114,18 @@ jobs:
...
         XSPECVER: ${{ matrix.xspec-version }}
+        # Temporarily stop ds9/xpa from installing
+        # on macOS because the tests fail for some
+        # unknown reason
+        OSNAME: {{ matrix.os }}
       run: |
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
-        source .github/scripts/setup_ds9.sh
+        if [ "$OSNAME" != "macos-latest" ]; then
```
Marie-Terrell, line 124:
```suggestion
        if [ "$RUNNER_OS" != "macOS" ]; then
```
I have not tested this in this section of the workflow, but this is used elsewhere in this workflow and would avoid creating an extra var if it works.

.github/workflows/ci-conda-workflow.yml
```diff
@@ -119,16 +117,11 @@ jobs:
       run: |
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
-        source .github/scripts/setup_ds9.sh
+        if [ "$RUNNER_OS" != "macos" ]; then
```
Marie-Terrell, line 120:
```suggestion
        if [ "$RUNNER_OS" != "macOS" ]; then
```
I think this is case sensitive here.

DougBurke: argle bargle arghhhhhhh ;-)

DougBurke: Definitely fixed now .....

DougBurke:
Argh - it's now the smoke test that's complaining. Which makes sense....

DougBurke:
This rebase will fix everything. Honest.

.github/workflows/ci-conda-workflow.yml
```diff
@@ -169,6 +162,14 @@ jobs:
         if [ -n "${XSPECVER}" ]; then XSPECTEST="-x -d"; fi
         if [ -n "${FITS}" ] ; then FITSTEST="-f ${FITS}"; fi
         smokevars="${XSPECTEST} ${FITSTEST} -v 3"
+
+        # special case the macOS build as we currently do not
+        # include DS9; as we know the XSPECVER/FITS sittings for
+        # this it is hard-coded
+        if [ "$RUNNER_OS" == "macOS" ]; then
+          smokevars="-x -f -v 3"
```
DougBurke, line 170:
```suggestion
          smokevars="-x -f ${FITS} -v 3"
```

arghhhhhhhhhh

DougBurke:
Defniitely^2 fixed now.

DougBurke:
This rebase was just to fix the last rebase which was perfect. Just not perfect enough.

DougBurke:
I think I've answered all the review questions @Marie-Terrell . Is there anything more we should look at?

.github/workflows/ci-conda-workflow.yml
```diff
@@ -169,6 +162,14 @@ jobs:
         if [ -n "${XSPECVER}" ]; then XSPECTEST="-x -d"; fi
```
Marie-Terrell, line 162:
```suggestion
        if [ "$RUNNER_OS" != "macOS" ]; then DS9TEST=" -d"; fi
        if [ -n "${XSPECVER}" ]; then XSPECTEST="-x${DS9TEST}"; fi
```
I don't know if this makes things easier or harder to read, so feel free to ignore this suggestion.

DougBurke: I would hope that this is all a temporary change so I don't feel like over-engineering the solution..........

Marie-Terrell: Agreed. And I hope so too!

Marie-Terrell (approved):
Looks good to me!

PR #1926, opened 2023-10-19
Let users know they should use load_xstable_model not load_table_model

DougBurke:
# Summary

Support for XSPEC table models in load_table_model was deprecated in the 4.9 release. Ensure users know that they should be using load_xstable_model instead by adding a warning message whenever it is used.

# Details

This is part of a long sequence of I/O changes I plan to make for 4.16.1. I am trying to make sure PRs are obvious and not-too-large. Next should be #1929 and then #1921.

The warning about the deprecation has been in place since 4.9 but thanks to how Python shows deprecation warnings essentially no-one would see the warning [even with https://peps.python.org/pep-0565/]. So we now have

```
>>> from sherpa.astro import ui
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
>>> ui.load_table_model("foo", "sherpa-test-data/sherpatest/xspec-tablemodel-RCS.mod")
WARNING: Use load_xstable_model to load XSPEC table models
```

There is the argument as to why we want to deprecate this functionality, to which I say

1. it will make the code slightly-nicer to understand (our code related to I/O is a rats nest so any improvememt is a good improvement)

2. over the years I've seen a lot of confusion about table models as their is a difference between reading in an array which we can then re-normalize for any data set versus the full-blown multi-parameter interpolation supported by XSPEC **for PHA data only**

3. We have documented this deprecation for years

There are also a number of test tweaks (to improve coverage) and minor pylint-inspired changes, which are

- explicitly catch `Exception` rather than no exception

  This technically changes behaviour, as we now no-longer catch `KeyboardInterrupt` (i.e. control-c), but this is actually the
  correct behaviour here

- f-string conversions
- apart for for lazy logging
- remove an unused symbol

Issue #1907, opened 2023-09-26
DataRMF.apply_rmf infelicities


DougBurke:
At the moment we have, for `sherpa.astro.data.DataPHA`:

https://github.com/sherpa/sherpa/blob/585c709e3165be5da02a88ed77622deafd6c3e1b/sherpa/astro/data.py#L1258-L1273

which means that we are given an array of numbers (the `src` parameter) and have to magically convolve them by the RMF without knowing what grid they were defined on. Now, if `args is set then we can use them to work out what the grid was but it is a truly atrocious API. I did start trying to improve this as part of #1869 but it quickly gets complicated and we need to resolve some or al l of issues like (but not limited to)

- #1895 
- #1894 
- #1871 
- #1870 
- #1678 
- #1564
- [ran out of patience going through the issues list]

Issue #1871, opened 2023-09-01
some of our PHA response handling goes beyond XSPEC/OGIP standards


DougBurke:
In trying to implement #883 I'm diffing around the response handling, and we support some situations that XSPEC does not, because we are allowing the combination of responses that do not "follow" the OGIP standard. For example, we have

https://github.com/sherpa/sherpa/blob/319b53b6a6e300905698cf317cdd33bf1eb60591/sherpa/astro/tests/test_astro.py#L659-L678

If we try to load this data into XSPEC 12.13.1a we get the following error:

```
xspec

		XSPEC version: 12.13.1a
	Build Date/Time: Mon Jul 31 15:03:37 2023

XSPEC12>data source1.pi

1 spectrum  in use
 
Spectral Data File: source1.pi  Spectrum 1
Net count rate (cts/s) for Spectrum:1  4.237e-01 +/- 7.345e-03
 Assigned to Data Group 1 and Plot Group 1
  Noticed Channels:  1-1024
  Telescope: CHANDRA Instrument: ACIS  Channel Type: PI
  Exposure Time: 7854 sec
 Using fit statistic: chi
 No response loaded.

***Warning!  One or more spectra are missing responses,
               and are not suitable for fit.
XSPEC12>response rmf_1024.fits
Response successfully loaded.
XSPEC12>arf arf_1024.fits
Error: the number of energy bins in the rmf (198) and in the arf (495) are not consistent.
XSPEC12>

```

Now, there may be a reason we want to allow this, but the undocumented code to support this does complicate things. We should probably review our PHA response combination code. It is tangentially related to issues like 

- #1845 
- #1846

DougBurke:
Another example is

https://github.com/sherpa/sherpa/blob/319b53b6a6e300905698cf317cdd33bf1eb60591/sherpa/astro/tests/test_cache.py#L37-L67

If you try something similar in XSPEC (here the file `gack.pha` is the `3c120_meg_1.pha` file with the RESPFILE set to `none` to replicate the test behaviour we get:

```
% xspec

		XSPEC version: 12.13.1a
	Build Date/Time: Mon Jul 31 15:03:37 2023

XSPEC12>data gack.pha

1 spectrum  in use
 
Spectral Data File: gack.pha  Spectrum 1
Net count rate (cts/s) for Spectrum:1  4.366e-01 +/- 3.548e-03
 Assigned to Data Group 1 and Plot Group 1
  Noticed Channels:  1-8192
[...]

PR #1944, opened 2024-01-03
Allow all matrices in a RMF to be used

DougBurke:
# Summary

Support reading in RMF files with multiple matrix blocks from OGIP Calibration Memo CAL/GEN/92-002  https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_002/cal_gen_92_002.html. This support should be considered experimental at this time.

# Details

This requires

- #1926 
- ~~#1929~~
- #2017
- #1921
- #2123

Can we handle multi-MATRIX RMF blocks? We have a decision on how to handle this - do we extend `DataRMF` or do we handle this within the multiple-response instrument code? When extending `DataRMF` we have

- [+] it should be seamless to the user (e.g. no difference to using a "normal" RMF)
- [-] this overloads the meaning of the `DataRMF` fields and subverts OO methodology - e.g. https://en.wikipedia.org/wiki/Liskov_substitution_principle
  - I will note we have existing code that does this already, so this is not new 
 
Using the multiple-response code might be nice, but I do not understand that code so I have gone with this version for now. I will note that in issue #1907 I talk about problems we have associating a model with the grid that was used to create it, and I think it's all related.

I have therefore opted to add a `DataMultiRMF` class that extends `DataRMF` since I think the ease-of-use argument is very important. I am slapping on a "this is experimental" notice so we can change it if we find this ends up not working well.

# Issues

One issue is that we only have the spec documentation. There aren't really any files lying around with real data (maybe from XRISM?) which makes testing this out and checking it works properly a bit awkward.

You can not write out a multi-matrix RMF (via `write_rmf`). This should be fixable.

The "representation" of a multi-matrix RMF hides all but the first block - e.g. print(rmf) - which is not ideal. This should be fixable. I think the new `plot_rmf` command should work but I can not be sure.

In part because we have some strange non-OGIP cases (#1871) and in part because of the existing API (#1907) I am not 100% convinced there aren't bugs hiding here.

The new RMF does not do any of the fancy "ignore/notice" handling that `DataRMF` does. I should have this "disabled", but I did mention the issue of hidden bugs, didn't I?

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2015-2025
+#  Copyright (C) 2008, 2015-2026
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -132,13 +132,12 @@ from __future__ import annotations
 from collections.abc import Callable, Mapping
 import logging
 import os
-from typing import Any, Literal, cast, overload, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Literal, cast, overload
 import warnings
 
 import numpy as np
 import numpy.typing as npt
 
-from sherpa.astro import hc
 from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
     IntegratedDataSpace2D, _check
 from sherpa.models.regrid import EvaluationSpace1D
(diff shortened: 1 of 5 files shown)


DougBurke:
Rebased to pick up #1938

DougBurke:
This has been rebased to pick up changes in #2094. Unfortunately, as this is based on #1921 and I don't want to rebase that, there is going to be a schism. Let's see how bad it gets.

DougBurke:
Rebased as #1921 has been merged.

DougBurke:
Rebasing to pick up recent changes and address a possible conflict

DougBurke:
Rebased to address conflict and the 4.17.0 release (give or take the final commit).

DougBurke:
Rebased to pick up recent build fixes, such as #2174

PR #1946, opened 2024-01-05
versioneer: update to version 0.29

DougBurke:
# Summary

Update the vendored copy of versioneer from 0.28 to 0.29.

# Details

Version 0.29 of versioneer was released in mid 2023 and adds support for pyproject.toml-only configuration, which may be relevant if we move to installing with meson.

The update was as simple as

```
% pip install versioneer
% python setup.py version
... to check existing output
% versioneer install --vendor
% python setup.py version
... to check it still works
```

and then committing the `versioneer.py` and `sherpa/_version.py` changes.


hamogu (approved):

DougBurke:
I will note that this may not be needed since we may drop versioneer completely (as I think it requires setuptools). I don't yet know the expected way to set dynamic version information using mesonpy/meson-python ...

Marie-Terrell (approved):
Looks good. I tried a few different python versions, install types, and sdist creation. All look good.

PR #1952, opened 2024-01-11
Tests: better support of optional region/wcs dependencies

DougBurke:
# Summary

Ensure that the tests can pass if the region or WCS code is not available.

# Details

The idea is to allow Sherpa to be built without the region or WCS libraries. These are low-level Astronomy-specific modules so it makes sense to make them optional, but we don't currently make this easy to do so (that is, this commit does not make it easier to avoid building these two extension modules, which should probably be done, but that should only be done once we've decided how to build with Python 3.12 as that is going to be very invasive for the extension building/where this code would live).

This is mainly changes to the tests, to add decorators, but there are code changes

- addition of `requires_region` and `requires_wcs` decorators
- the `sherpa.astro.io.wcs` module has been changed to make the lack of the WCS code an option, rather than being required There are other ways to do this, but for now it's easiest to do it this way, as it depends on the long-term build changes that are planned.

This was work I did to help with converting to Python 3.12 (the `distutils` changes) but it turned out not to be necessary for that.

### DIFF
--- a/sherpa/astro/io/tests/test_io_img.py
+++ b/sherpa/astro/io/tests/test_io_img.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2021, 2023
+#  Copyright (C) 2021, 2023, 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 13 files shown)


hamogu (approved):
This is one of the PRs that are low stakes to review. As @DougBurke said in the telecon today, all it does is adding decorators to a few tests. There is essentially no risk to that - if we missed a test it's not worse than before and we can add it later; if we marked one too many that's also not a problem because we usually run our tests with wcs and regions.

That's not to say that I did not review the changes (I did), but just noting that there really is no harm in merging this.

PR #1957, opened 2024-02-06
CI: support pytest 8.0.0

### LINKED ISSUE #1956: pytest 8.0.0 changes
We now have pytest 8.0.0 on some of our CI runs and I point out, from the 8.0.0 changelog (currently https://docs.pytest.org/en/stable/changelog.html ) the following nugget

- [#9288](https://github.com/pytest-dev/pytest/issues/9288): [warns()](https://docs.pytest.org/en/stable/reference/reference.html#pytest.warns) now re-emits unmatched warnings when the context closes – previously it would consume all warnings, hiding those that were not matched by the function.

  While this is a new feature, we announce it as a breaking change because many test suites are configured to error-out on warnings, and will therefore fail on the newly-re-emitted warnings.

since I just got a test failure because of this that looks exactly like this issue:

```
-------------------------- Captured stdout teardown ---------------------------
*** Warnings created: 1
1/1 {message : FutureWarning('calc() requires pars,rhs,lo,hi arguments, sent 3 arguments'), category : 'FutureWarning', filename : '/home/runner/work/sherpa/sherpa/sherpa/astro/xspec/__init__.py', lineno : 1591, line : None}

----------- coverage: platform linux, python 3.9.0-final-0 -----------
```

DougBurke:
# Summary

Allow the tests to pass when run with pytest 8.0.0.

# Details

Fix #1956

These two tests created two warning messages, first from the additive model (m1) and then the convolution model (m2), but we only checked one as the other one was swallowed by pytest (since, until pytest 8, the other warning was ignored). Well, pytest 8.0.0 changes this behaviour, so fix the test.

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -835,7 +835,7 @@ def test_xspec_convolutionmodel_requires_bin_edges():
     m2 = xs.XScflux()
     mdl = m2(m1)
 
-    # We get warnings from m1 evalated on the grid and then the
+    # We get warnings from m1 evaluated on the grid and then the
     # convolution model m2.
     #
     emsg1 = r'calc\(\) requires pars,rhs,lo,hi arguments, sent 3 arguments'
@@ -858,7 +858,7 @@ def test_xspec_convolutionmodel_requires_bin_edges_low_level():
     m2 = xs.XScflux()
     mdl = m2(m1)
 
-    # We get warnings from m1 evalated on the grid and then the
+    # We get warnings from m1 evaluated on the grid and then the
     # convolution model m2.
     #
     emsg1 = r'calc\(\) requires pars,rhs,lo,hi arguments, sent 3 arguments'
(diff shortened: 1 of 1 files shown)


DougBurke:
Argh - I've had to temporarily pull in #1954 too

hamogu (approved):

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -835,9 +835,14 @@ def test_xspec_convolutionmodel_requires_bin_edges():
     m2 = xs.XScflux()
     mdl = m2(m1)
 
-    emsg = r'calc\(\) requires pars,lo,hi arguments, sent 2 arguments'
-    with pytest.warns(FutureWarning, match=emsg):
-        mdl([0.1, 0.2, 0.3, 0.4])
+    # We get warnings from m1 evalated on the grid and then the
```
wmclaugh, line 838:
```suggestion
    # We get warnings from m1 evaluated on the grid and then the
```

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -853,9 +858,14 @@ def test_xspec_convolutionmodel_requires_bin_edges_low_level():
     m2 = xs.XScflux()
     mdl = m2(m1)
 
-    emsg = r'calc\(\) requires pars,lo,hi arguments, sent 2 arguments'
-    with pytest.warns(FutureWarning, match=emsg):
-        mdl.calc([p.val for p in mdl.pars], [0.1, 0.2, 0.3, 0.4])
+    # We get warnings from m1 evalated on the grid and then the
```
wmclaugh, line 861:
```suggestion
    # We get warnings from m1 evaluated on the grid and then the
```

PR #1962, opened 2024-02-16
Pick up latest Python micro version in conda workflow

Marie-Terrell:
This resolves the missing crypt.h issue being seen in the conda test workflow by picking up the latest micro version of the python packages for the conda test build workflows.


DougBurke (approved):
We could squash the two commits. But yay.

PR #1965, opened 2024-02-19
Improve the error message when a plot backend is not available.

### LINKED ISSUE #1964: improve error message when plot backend is not available
From a helpdesk ticket, the error message was that `plot_data` (when displaying PHA data) returned

```
WARNING:  does not implement histogram plotting. No histogram will be produced.
```

so it's not correctly including the backend name. Here's the startup info

```
% sherpa -n
Python 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:49:36) [Clang 16.0.6 ]
IPython 8.21.0 -- An enhanced Interactive Python. Type '?' for help.
IPython profile: sherpa
Using matplotlib backend: MacOSX
WARNING: Plotting backend 'chips' not found or dependencies missing. Trying next option.
WARNING: Tried the following backends listed in /Users/meredithpowell/.sherpa.rc: 
['chips']
None of these imported correctly, so using the 'BasicBackend'.
List of backends that have loaded and would be available: ['BaseBackend', 'BasicBackend', 'IndepOnlyBackend', 'pylab', 'PylabErrorArea']
Solar Abundance Vector set to angr:  Anders E. & Grevesse N. Geochimica et Cosmochimica Acta 53, 197 (1989)
 Cross Section Table set to vern:  Verner, Ferland, Korista, and Yakovlev 1996
```

Now, techinically it's "quite hard" to do because the user-requested backend (here `chips`) isn't actually known at the call site (which would be something like `BasicBackend`), but we do not need to solve that particular issue.

The code issue is that the `BasicBackend` has a lot of lines like

https://github.com/sherpa/sherpa/blob/2b7eec14c97742ce83c4d73b4641971c50a0e062/sherpa/plot/backends.py#L897-L899

I think it should use `self.name` instead (or something equivalent), and converted to `'%s .....", self.name`, but that's only really to quieten `pylint`, I'm not sure how other python linters care about https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-not-lazy.html

DougBurke:
# Summary

Improve the error message when a plot backend is not available. Fix #1964 

# Details

There is some minor code-cleanup here, but the main change is that the warning message that gets displayed uses the `self.name` attribute of the backend rather than `self.__class__`, which is a little-more-readable and doesn't contain `<>` characters (which then can get caught in our helpdesk system, leading to the text being lost in error reports).

### DIFF
--- a/sherpa/plot/backends.py
+++ b/sherpa/plot/backends.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2020, 2021, 2022, 2023
+#  Copyright (C) 2007, 2015, 2020 - 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -142,7 +142,6 @@ uses the backend-specific default.'''],
               'levels': ['array-like', 'Levels at which to draw the contours'],
               'aspect': ['str or float', 'Aspect ratio of the plot. Strings "equal" or "auto" are accepted.'],
               'label': ['str', 'Label this dataset for use in a legend'],
-              'levels': ['array-like', 'Levels at which to draw the contours'],
               'ymin' : ['float', '''Beginning of the vertical line in axes coordinates,
 i.e. from 0 (bottom) to 1 (top).'''],
               'ymax' : ['float', '''End of the vertical line in axes coordinates,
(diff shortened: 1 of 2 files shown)


hamogu (approved):
LGTM. Note that this is build on top pf #1962 (required to make CI pass), so either merge #192 first or close that one if this is merged first.

PR #1968, opened 2024-02-22
Remove Conda build 3.25 pin from deployments 

Marie-Terrell:
We no longer need to specify conda-build=3.25 (as newer versions of mamba and others are out). Keeping this now causes compatibility issues. This PR just removed the version specification


harlantc (approved):
LGTM

PR #1970, opened 2024-02-23
Bump the minimum pytest version

### LINKED ISSUE #1960: argh: pytest < 8
So, #1957 lets us run tests with pytest >= 8.0.0. 

However, if you run with pytest 7.4.0 (for example) the changes fail - e.g.

```
________________ test_xspec_convolutionmodel_requires_bin_edges ________________

    @requires_xspec
    def test_xspec_convolutionmodel_requires_bin_edges():
        """Check we can not call a convolution model with a single grid.
    
        This used to be supported in Sherpa 4.13 and before.
        """
    
        import sherpa.astro.xspec as xs
    
        m1 = xs.XSpowerlaw()
        m2 = xs.XScflux()
        mdl = m2(m1)
    
        # We get warnings from m1 evaluated on the grid and then the
        # convolution model m2.
        #
        emsg1 = r'calc\(\) requires pars,rhs,lo,hi arguments, sent 3 arguments'
        emsg2 = r'calc\(\) requires pars,lo,hi arguments, sent 2 arguments'
>       with pytest.warns(FutureWarning, match=emsg1):
E       Failed: DID NOT WARN. No warnings of type (<class 'FutureWarning'>,) were emitted.
E       The list of emitted warnings is: [].

sherpa/astro/xspec/tests/test_xspec_unit.py:843: Failed
___________ test_xspec_convolutionmodel_requires_bin_edges_low_level ___________

    @requires_xspec
    def test_xspec_convolutionmodel_requires_bin_edges_low_level():
        """Check we can not call a convolution model with a single grid (calc).
    
        This used to be supported in Sherpa 4.13 and before.
        """
    
        import sherpa.astro.xspec as xs
    
        m1 = xs.XSpowerlaw()
        m2 = xs.XScflux()
        mdl = m2(m1)
    
        # We get warnings from m1 evaluated on the grid and then the
        # convolution model m2.
        #
        emsg1 = r'calc\(\) requires pars,rhs,lo,hi arguments, sent 3 arguments'
        emsg2 = r'calc\(\) requires pars,lo,hi arguments, sent 2 arguments'
>       with pytest.warns(FutureWarning, match=emsg1):
E       Failed: DID NOT WARN. No warnings of type (<class 'FutureWarning'>,) were emitted.
E       The list of emitted warnings is: [].

sherpa/astro/xspec/tests/test_xspec_unit.py:866: Failed
```

Need to think about how best to handle (other than just saying we need `pytest >= 8` !)

hamogu:
Pytest occasionally changes behavior and it does not seem worth our time to maintain backwards compatibility with old versions of pytest - see https://github.com/sherpa/sherpa/issues/1960

closes #1960

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ install_requires =
 
 tests_require =
     pytest-xvfb
-    pytest>=5.0,!=5.23
+    pytest>=8.0
 
 # Be explicit as something doesn't seem quite right when using find.
 #
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,3 +1,3 @@
-pytest>=5.0,!=5.2.3
+pytest>=8.0
 pytest-xvfb
 pytest-doctestplus


DougBurke:
We should also update `test_requirements.txt`

DougBurke (approved):
I had a look in the docs and we don't have anything that talks about a minimum version, so this should be good to go.

PR #1060, opened 2021-01-16
WIP Remove excess brackets

DougBurke:
# Summary

Remove mathematically un-necessary brackets when displaying a model expression.

The code is there but more tests need adding and some clean up may be possible

The main problem at this time is that "a - (b - c)" gets converted to "a - b - c", which is not-quite-right™.

# Details

The `name` field of the `sherpa.models.model.Model` class is changed so that

- it is now a property (terminology?)
- the supplied value is stored as _name (and currently unused)
- access to `name` will now call the `sherpa.models.token.simplify` routine to be called on the stored model

The `simplify` routine is the routine that removes excess brackets, and is described below. Note that sub-classes of `Model` can override the `name` attribute so that you can have logic like (from `sherpa.astro.insrument` classes deriving from `CompositeModel`):

        CompositeModel.__init__(self, f'apply_rmf(apply_arf({model.name}))', (model,))

Although how this actually works is actually surprising to me, since this ends up setting the `_name` field which isn't then used. So I need to go back and review it, but it does seem to work... It also probably explains why some brackets have not been removed in the `get_model` output below.

TODO.

The basic idea is taken from https://stackoverflow.com/questions/18400741/remove-redundant-parentheses-from-an-arithmetic-expression

# Concerns

The logic is hidden away (the user doesn't interact with it), but it is technically possible for this to either break code (some condition isn't followed leading to a perfectly-valid expression to error out) or it removes a bracket that it shouldn't have. The latter is more concerning, as there's several parts to this work which means it isn't an easily-verified algorithm. My approach is to add a bunch of tests to catch possible problems. It would be nice to use some form of fuzzing or property testing approach.

# Example

Here's a source expression with a lot of terms in the current code:

```
sherpa In [3]: get_source()
Out[3]: <BinaryOpModel model instance '((xsconstant.m1 * xstbabs.m2) * ((((((((((((((((((((xsbremss.m3 + xsbremss.m4) + xsbremss.m5) + xsbremss.m6) + xsgaussian.m7) + xsgaussian.m8) + xsgaussian.m9) + xsgaussian.m10) + xsgaussian.m11) + xsgaussian.m12) + xsgaussian.m13) + xsgaussian.m14) + xsgaussian.m15) + xsgaussian.m16) + xsgaussian.m17) + xsgaussian.m18) + xsgaussian.m19) + xsgaussian.m20) + xsgaussian.m21) + xsgaussian.m22) + xsgaussian.m23))'>

sherpa In [4]: get_source().name
[...]

PR #1972, opened 2024-02-27
Remove excess brackets in model names

### LINKED ISSUE #780: can we avoid excess brackets when displaying model expressions
Something I've been meaning to get to but never do, so I thought I'd best write it down. This is purely a user-experience issue, but can we reduce the number of brackets added in the `name` attribute of a model.

I've just been helping a user who has this lovely expression (from `save_all` but you get it from `get_model`):

```python
set_full_model(1, (apply_rmf(apply_arf((((xswabs.abs1 * xswabs.int_abs) * xsmekal.mek1)))) + (0.0510501847393 * apply_bkg_rmf(apply_bkg_arf(((const1d.bkg1_const * ((((((((powlaw1d.bkg1_pow1 + powlaw1d.bkg1_pow2) + gauss1d.bkg1_g1) + gauss1d.bkg1_g2) + gauss1d.bkg1_g3) + gauss1d.bkg1_g4) + gauss1d.bkg1_g5) + gauss1d.bkg1_g6) + xsapec.bkg1_apec))))))))
```

That's a lot of brackets, and most of them are not needed.

For some situations, such as `save_all`, we could also simplify to avoid the `<model>.<instance>` form and just go with `<instance>`, but I am not advocating that here.

Here's a repeatable example:

```
In [2]: from sherpa.models import basic

In [3]: m1 = basic.Const1D()

In [4]: m2 = basic.Gauss1D()

In [5]: m3 = basic.Polynom1D()

In [6]: m = m1 + m2 + m3

In [7]: m
Out[7]: <BinaryOpModel model instance '((const1d + gauss1d) + polynom1d)'>

In [8]: print(m)
((const1d + gauss1d) + polynom1d)
   const1d.c0   thawed            1 -3.40282e+38  3.40282e+38
   gauss1d.fwhm thawed           10  1.17549e-38  3.40282e+38
   gauss1d.pos  thawed            0 -3.40282e+38  3.40282e+38
   gauss1d.ampl thawed            1 -3.40282e+38  3.40282e+38
   polynom1d.c0 thawed            1 -3.40282e+38  3.40282e+38
   polynom1d.c1 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c2 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c3 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c4 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c5 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c6 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c7 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.c8 frozen            0 -3.40282e+38  3.40282e+38
   polynom1d.offset frozen            0 -3.40282e+38  3.40282e+38

In [9]: m.name
Out[9]: '((const1d + gauss1d) + polynom1d)'

```

Note that `m.name` could be `(const1d + gauss1d + polynom1d)` here

DougBurke:
# Summary

Remove excess brackets from model and parameter expressions. This is purely a cosmetic change, but hopefully makes complicated model expressions easier to read. Fix #780.

# Details

This is back on the #1060 bandwagon. But less significantly invasive.

Can we make model expressions like `xsphabs.gal * (powlaw1d.pl + gauss1d.gl + xsapec.foo)` display as `gal * (pl + gl + foo)` rather than `(gal * ((pl + gl) + foo))`?

Note that I explicitly am not trying to "simplify" expressions - e.g. change `gal + gal` to `2 * gal` of `gal - gal` to `0`, as I want the actual evaluation to match the original requested expression (since there may be reasons a user wants the model to evaluate in a certain manner, and it also makes things a lot easier).

All the various attempts rely on the fact that the arithmetic expressions (model, via `ArithmeticModel` or parameter via `Parameter`)

- use the dunder methods for addition, subtraction, multiplication, ... which create "combined" models containing the
- use specialized versions based on `CompositeModel` or `CompositeParameter` to handle unary and binary operators
  - for models these are `UnaryOpModel` and `BinaryOpModel` 
  - they wrap numeric constants in specialized model/parameter classes that just return the content, so in "2 * mdl"
    the "2" value is actually `ArithmeticConstantModel`
    
        >>> 2 * mdl
        <BinaryOpModel model instance '(2.0 * const1d.mdl)'>
        >>> (2 * mdl).lhs
        <ArithmeticConstantModel model instance '2.0'>

    (this isn't hugely relevant to the discussion, but I thought I'd mention it).

So, in [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) form we have something like the following (but this is meant to be "in-spirit" rather than a faithful conversion, in particular it doesn't handle functions rather than operators for both the unary and binary cases):

    expr :: term | unop op_single term | binop op_binary term term
    op_single :: + | - | abs
    op_binary :: + | - | * | / | ** | %
    term :: numeric-value | object

where `object` refers to `Model` or `Parameter`. In the following XXX means "Model or Parameter" for the class names, as the logic is replicated.

So, when a user enters `a * (b - 3)` then Python automatically creates expression

    BinaryOpXXX(lhs=a,
                    rhs=BinaryOpXXX(lhs=b, rhs=3, op="-"),
                    op="*")

for us. We can then control the "string" version by controlling how the `name` fields are created for the `BinaryOpXXX` and `UnaryOpXXX` terms, since these `.name` fields are used to define the string output (it's not always `__str__` that we care about, e.g. for `Parameter`).

It turns out that we really only care about the `BinaryOpXXX` classes, since although the `UnaryOpXXX` case seems sensible, I've decided it's not worth the complexity, so we "always" add a bracket around the interior term (the unary case handles both "-a" and "abs(a)", so it actually gets a bit complex about the rule - e.g. see

- #1653 

which I want to expand from parameters to models as well).

So, in this PR the logic for removing the excess brackets is moved into the constructor for `BinaryOpXXX`, since this has all the information we need (the operator and the lhs and rhs terms). There are plenty of examples about converting from infix to postfix form (e.g. RPN) and/or handling of the brackets - e.g. https://stackoverflow.com/questions/18400741/remove-redundant-parentheses-from-an-arithmetic-expression . My original attempts were too much, and I realized that 

- much of the discussion on these sites is about parsing the expression, but we have Python doing this for us, so all we care about is how we iterate through the expression tree
- we just want to simplify common expressions that users might create - e.g. we are not trying to simplify `(a + (b - a))` to `b`, or remove every excess pair of brackets, in particular when it comes to subtraction, which is not a common use case for model expressions
- we only need to put the logic into `BinaryOpXXX`, as this is the part of the model language that handles `a <op> b` terms
- you have to be careful because many of the examples do not treat all the things we care about, in particular negation (e.g. what happens with `(a - (b - c))`?)

So, when calculating the `name` field to pass to the parent constructor, we have `lhs`, `rhs`, and the operator term. We can process the two sides and then decide whether they need to add brackets around them. This is slightly tricky as we have to deal with precedences - e.g. to make sure that "(a - (b - c))" doesn't become `a - b - c`. As mentioned, I want to deal with common cases and I am not too bothered if there are one or two too many brackets left! I have extensive tests to try and check the cases we care about.

For the parameter case there's a slight subtly in the choice of `name` vs `fullname` fields, and it's not clear our existing code really handles this well, but it only really matters when the term is a `Parameter` instance (rather than a `CompositeParameter`) and this is only for display (and this behaviour/distinction has been in Sherpa essentially for ever, so nothing I've changed here).

# Notes

One concern I have is over hitting a RecursionError, since this process is recursive. However, the way it is implemented it's not obvious whether we really hit this case. I have been able to create a model expression with > 1000 terms and no recursion error is reported, so I think we are okay.

I was initially concerned that convolution-style models would need to be special cased to benefit from this (e.g. expressions involving a RMF) but fortunately that was wrong, since the convolution models pick this up automatically (via BinaryOpModel).

This simplification only happens when the expression uses the NumPy operators (e.g. `np.add`, `np.multiply`). It would be possible for users to use sub-classes of these routines (or entirely different operators) and we'd miss out on the simplification. To "fix" this we could use the operator symbol rather than operator function but I think we can wait until people start to do this (which I don't see happening).

# Examples

## With CIAO 4.16

```
sherpa In [1]: from sherpa.models import basic

sherpa In [2]: a = basic.Const1D("a")

sherpa In [3]: b = basic.Const1D("b")

sherpa In [4]: c = basic.Const1D("c")

sherpa In [5]: a * b + 2 * c
Out[5]: <BinaryOpModel model instance '((a * b) + (2.0 * c))'>

sherpa In [6]: a - b - c
Out[6]: <BinaryOpModel model instance '((a - b) - c)'>

sherpa In [7]: a - (b - c)
Out[7]: <BinaryOpModel model instance '(a - (b - c))'>

sherpa In [8]: a * 2 * c + 2 * (a + b / c)
Out[8]: <BinaryOpModel model instance '(((a * 2.0) * c) + (2.0 * (a + (b / c))))'>

sherpa In [9]: (a + 2) ** 2
Out[9]: <BinaryOpModel model instance '((a + 2.0) ** 2.0)'>

sherpa In [10]: a + a + a + a - a + a
Out[10]: <BinaryOpModel model instance '(((((a + a) + a) + a) - a) + a)'>

sherpa In [11]: a + a + (a + a) - a + a
Out[11]: <BinaryOpModel model instance '((((a + a) + (a + a)) - a) + a)'>

sherpa In [12]: a - (b + c)
Out[12]: <BinaryOpModel model instance '(a - (b + c))'>
```

## With this PR

```
>>> from sherpa.models import basic
>>> a = basic.Const1D("a")
>>> b = basic.Const1D("b")
>>> c = basic.Const1D("c")
>>> a * b + 2 * c
<BinaryOpModel model instance 'a * b + 2.0 * c'>
>>> a - b - c
<BinaryOpModel model instance 'a - b - c'>
>>> a - (b - c)
<BinaryOpModel model instance 'a - (b - c)'>
>>> a * 2 * c + 2 * (a + b / c)
<BinaryOpModel model instance 'a * 2.0 * c + 2.0 * (a + b / c)'>
>>> (a + 2) ** 2
<BinaryOpModel model instance '(a + 2.0) ** 2.0'>
>>> a + a + a + a - a + a
<BinaryOpModel model instance 'a + a + a + a - a + a'>
>>> a + a + (a + a) - a + a
<BinaryOpModel model instance 'a + a + a + a - a + a'>
>>> a - (b + c)
<BinaryOpModel model instance 'a - (b + c)'>
```

If you have convolution-style models - such as a model expression for a PHA dataset  - then the term inside the `apply_rmf(apply_arf(...))` tokens will have this simplification. You can see the changes to the various tests that show this.

We do not bother with removing brackets around `UnaryOpModel` cases since I didn't think it worth it. This means we simplify

```
sherpa-4.16.0> -a - b
<BinaryOpModel model instance '(-(a) - b)'>
```

to

```
>>> -a - b
<BinaryOpModel model instance '-(a) - b'>
```

rather than to `-a - b`. As most people don't use UnaryOpModel I think this is okay.

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -322,7 +322,7 @@ import logging
 from typing import Callable, Optional
 import warnings
 
-import numpy
+import numpy as np
 
 from sherpa.models.regrid import EvaluationSpace1D, ModelDomainRegridder1D, EvaluationSpace2D, ModelDomainRegridder2D
 from sherpa.utils import NoNewAttributesAfterInit, formatting
@@ -443,18 +443,18 @@ def modelCacher1d(func):
             #
             integrate = kwargs.get('integrate', False)
 
-        data = [numpy.array(pars).tobytes(),
+        data = [np.array(pars).tobytes(),
                 boolean_to_byte(integrate),
-                numpy.asarray(xlo).tobytes()]
+                np.asarray(xlo).tobytes()]
         if args:
-            data.append(numpy.asarray(args[0]).tobytes())
+            data.append(np.asarray(args[0]).tobytes())
 
         # Add any keyword arguments to the list. This will
         # include the xhi named argument if given. Can the
         # value field fail here?
         #
         for k, v in kwargs.items():
-            data.extend([k.encode(), numpy.asarray(v).tobytes()])
+            data.extend([k.encode(), np.asarray(v).tobytes()])
 
         # Is the value cached?
         #
@@ -1075,7 +1075,7 @@ class ArithmeticConstantModel(Model):
     def __init__(self, val, name=None):
         val = SherpaFloat(val)
         if name is None:
-            if numpy.isscalar(val):
+            if np.isscalar(val):
                 name = str(val)
             else:
                 nstr = ','.join([str(s) for s in val.shape])
@@ -1174,19 +1174,19 @@ class ArithmeticModel(Model):
              f"check: {c['check']:5d}")
 
     # Unary operations
-    __neg__ = _make_unop(numpy.negative, '-')
-    __pos__ = _make_unop(numpy.positive, '+')
-    __abs__ = _make_unop(numpy.absolute, 'abs')
+    __neg__ = _make_unop(np.negative, '-')
+    __pos__ = _make_unop(np.positive, '+')
+    __abs__ = _make_unop(np.absolute, 'abs')
 
     # Binary operations
-    __add__, __radd__ = _make_binop(numpy.add, '+')
-    __sub__, __rsub__ = _make_binop(numpy.subtract, '-')
-    __mul__, __rmul__ = _make_binop(numpy.multiply, '*')
-    __div__, __rdiv__ = _make_binop(numpy.divide, '/')
-    __floordiv__, __rfloordiv__ = _make_binop(numpy.floor_divide, '//')
-    __truediv__, __rtruediv__ = _make_binop(numpy.true_divide, '/')
-    __mod__, __rmod__ = _make_binop(numpy.remainder, '%')
-    __pow__, __rpow__ = _make_binop(numpy.power, '**')
+    __add__, __radd__ = _make_binop(np.add, '+')
+    __sub__, __rsub__ = _make_binop(np.subtract, '-')
+    __mul__, __rmul__ = _make_binop(np.multiply, '*')
+    __div__, __rdiv__ = _make_binop(np.divide, '/')
+    __floordiv__, __rfloordiv__ = _make_binop(np.floor_divide, '//')
+    __truediv__, __rtruediv__ = _make_binop(np.true_divide, '/')
+    __mod__, __rmod__ = _make_binop(np.remainder, '%')
+    __pow__, __rpow__ = _make_binop(np.power, '**')
 
     def __setstate__(self, state):
         self.__dict__.update(state)
@@ -1214,7 +1214,7 @@ class ArithmeticModel(Model):
             return
 
         self._queue = [''] * int(self.cache)
-        frozen = numpy.array([par.frozen for par in self.pars], dtype=bool)
+        frozen = np.array([par.frozen for par in self.pars], dtype=bool)
         if len(frozen) > 0 and frozen.all():
             self._use_caching = cache
 
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -316,10 +316,12 @@ non-integrated and integrated datasets of any dimensionality (see
 
 """
 
+from __future__ import annotations
 
 import functools
 import logging
-from typing import Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Optional, \
+    Sequence, SupportsFloat, SupportsIndex, Type, TypeVar, Union
 import warnings
 
 import numpy as np
@@ -363,7 +365,7 @@ __doctest_skip__ = ['ArithmeticModel.cache_status',
                     'SimulFitModel']
 
 
-def boolean_to_byte(boolean_value):
+def boolean_to_byte(boolean_value: bool) -> bytes:
     """Convert a boolean to a byte value.
 
     Parameters
@@ -382,7 +384,7 @@ def boolean_to_byte(boolean_value):
     return bmap.get(boolean_value, b'0')
 
 
-def modelCacher1d(func):
+def modelCacher1d(func: Callable) -> Callable:
     """A decorater to cache 1D ArithmeticModel evalutions.
 
     Apply to the `calc` method of a 1D model to allow the model
@@ -529,17 +531,19 @@ class Model(NoNewAttributesAfterInit):
     ndim: Optional[int] = None
     "The dimensionality of the model, if defined, or None."
 
-    def __init__(self, name, pars=()):
+    def __init__(self,
+                 name: str,
+                 pars: Sequence[Parameter] = ()) -> None:
         self.name = name
         self.type = self.__class__.__name__.lower()
         self.pars = tuple(pars)
         self.is_discrete = False
         NoNewAttributesAfterInit.__init__(self)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<{type(self).__name__} model instance '{self.name}'>"
 
-    def __str__(self):
+    def __str__(self) -> str:
         s = self.name
         sep5 = '-' * 5
         sep4 = '-' * 4
@@ -569,14 +573,14 @@ class Model(NoNewAttributesAfterInit):
 
         return s
 
-    def _repr_html_(self):
+    def _repr_html_(self) -> str:
         """Return a HTML (string) representation of the model
         """
         return html_model(self)
 
     # This allows all models to be used in iteration contexts, whether or
     # not they're composite
-    def __iter__(self):
+    def __iter__(self) -> Iterator[Model]:
         return iter([self])
 
     def __getattr__(self, name):
@@ -633,7 +637,7 @@ class Model(NoNewAttributesAfterInit):
         for alias in val.aliases:
             self._par_index[alias] = val
 
-    def startup(self, cache=False):
+    def startup(self, cache: bool = False) -> None:
         """Called before a model may be evaluated multiple times.
 
         Parameters
@@ -647,7 +651,10 @@ class Model(NoNewAttributesAfterInit):
         """
         raise NotImplementedError
 
-    def calc(self, p, *args, **kwargs):
+    def calc(self,
+             p: Sequence[SupportsFloat],
+             *args,
+             **kwargs) -> np.ndarray:
         """Evaluate the model on a grid.
 
         Parameters
@@ -665,7 +672,7 @@ class Model(NoNewAttributesAfterInit):
         """
         raise NotImplementedError
 
-    def teardown(self):
+    def teardown(self) -> None:
         """Called after a model may be evaluated multiple times.
 
         See Also
@@ -690,7 +697,7 @@ class Model(NoNewAttributesAfterInit):
     def set_center(self, *args, **kwargs):
         raise NotImplementedError
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args, **kwargs) -> Union[Model, np.ndarray]:
         # A bit of trickery, to make model creation
         # in IPython happen without raising errors, when
         # model is made automatically callable
@@ -698,10 +705,10 @@ class Model(NoNewAttributesAfterInit):
             return self
         return self.calc([p.val for p in self.pars], *args, **kwargs)
 
-    def _get_thawed_pars(self):
+    def _get_thawed_pars(self) -> list[SherpaFloat]:
         return [p.val for p in self.pars if not p.frozen]
 
-    def _set_thawed_pars(self, vals):
+    def _set_thawed_pars(self, vals: Sequence[SupportsFloat]) -> None:
         tpars = [p for p in self.pars if not p.frozen]
 
         ngot = len(vals)
@@ -731,10 +738,10 @@ class Model(NoNewAttributesAfterInit):
                           '--------\n' +
                           'thawedparmaxes, thawedparmins\n')
 
-    def _get_thawed_par_mins(self):
+    def _get_thawed_par_mins(self) -> list[SupportsFloat]:
         return [p.min for p in self.pars if not p.frozen]
 
-    def _set_thawed_pars_mins(self, vals):
+    def _set_thawed_pars_mins(self, vals: Sequence[SupportsFloat]) -> None:
         tpars = [p for p in self.pars if not p.frozen]
 
         ngot = len(vals)
@@ -767,10 +774,10 @@ class Model(NoNewAttributesAfterInit):
                              '--------\n' +
                              'thawedpars, thawedarhardmins, thawedparmaxes\n')
 
-    def _get_thawed_par_maxes(self):
+    def _get_thawed_par_maxes(self) -> list[SupportsFloat]:
         return [p.max for p in self.pars if not p.frozen]
 
-    def _set_thawed_pars_maxes(self, vals):
+    def _set_thawed_pars_maxes(self, vals: Sequence[SupportsFloat]) -> None:
         tpars = [p for p in self.pars if not p.frozen]
 
         ngot = len(vals)
@@ -803,7 +810,7 @@ class Model(NoNewAttributesAfterInit):
                               '--------\n' +
                               'thawedpars, thawedarhardmaxes, thawedparmins\n')
 
-    def _get_thawed_par_hardmins(self):
+    def _get_thawed_par_hardmins(self) -> list[SherpaFloat]:
         return [p.hard_min for p in self.pars if not p.frozen]
 
     thawedparhardmins = property(_get_thawed_par_hardmins,
@@ -816,7 +823,7 @@ class Model(NoNewAttributesAfterInit):
                                  '--------\n' +
                                  'thawedparhardmaxes, thawedparmins\n')
 
-    def _get_thawed_par_hardmaxes(self):
+    def _get_thawed_par_hardmaxes(self) -> list[SherpaFloat]:
         return [p.hard_max for p in self.pars if not p.frozen]
 
     thawedparhardmaxes = property(_get_thawed_par_hardmaxes,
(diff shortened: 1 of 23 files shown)


DougBurke:
Hmmm. Do we want the same behavior for parameter expressions (only really relevant if  you link parameters)?

DougBurke:
Rebased to be on top of #1977 as this improves coverage (and the changes here lead to needed changes in the docstrings).

DougBurke:
Rebased because #1977 was merged

DougBurke:
Rebased because #1976 was merged. It's not relevant here but is for #1679 which is based on this PR.

sherpa/models/tests/test_model_op.py
```diff
@@ -410,3 +450,245 @@ def test_binop_arithmeticfunction_works(m1, m2):
...
+                              (abs((a + b) * c), "abs((a + b) * c)"),
+                              (abs((a - b) * c), "abs((a - b) * c)"),
+                              (abs((a - b) / c), "abs((a - b) / c)"),
+                              (abs((a * b) - c), "abs(a * b - c)"),
+                              (abs((a / b) - c), "abs(a / b - c)"),
+                              (a * abs(b * (c + d)), "a * abs(b * (c + d))"),
+                              (abs(b * (c + d)) * (a + d), "abs(b * (c + d)) * (a + d)"),
+                              (-a, "-(a)"),
+                              (+a, "+(a)"),
+                              (-a + b, "-(a) + b"),
```
hamogu, line 515: Want to try one where you throw in a few extra pairs, e.g. `((-(a) + ((b))))`? Of course, no sensible user will ever do that, but it's be good to have a text to the expected output so we know if it changes later for whatever reason.

sherpa/models/tests/test_model_op.py
```diff
@@ -410,3 +450,245 @@ def test_binop_arithmeticfunction_works(m1, m2):
...
+                              ((a * b) * (c + d), "a * b * (c + d)"),
+                              ((a + b) * (c * d), "(a + b) * c * d"),
+                              ((a + (b * c) + d), "a + b * c + d"),
+                              (100 * a * (b + c), "100.0 * a * (b + c)"),
+                              (100 * (a * (b + c)), "100.0 * a * (b + c)"),
+                              (a + b + 2 * c + d + a, "a + b + 2.0 * c + d + a"),
+                              (a + b + c * 2 + d + a, "a + b + c * 2.0 + d + a"),
+                              (a + b * (c - 2) + d + a, "a + b * (c - 2.0) + d + a"),
+                              (a + b * (2 - c) + d + a, "a + b * (2.0 - c) + d + a"),
+                              ((a + b + c) + (c + b + d + a), "a + b + c + c + b + d + a"),
```
hamogu, line 553: I'm not asking to change that, I just want to mention that form like this might be chosen by the user for readibility, e.g., the first group is the star and the second group is the accretion disk. I think those cases are rare in practice and that's a drawback I'm happy to accept it. I just want to mention it.

sherpa/models/tests/test_parameter.py
```diff
@@ -220,11 +223,12 @@ def test_binop_string():
...
     p, p2 = setUp_composite()
     comp = p + p2
-    assert repr(comp) == "<BinaryOpParameter '(model.p + model.p2)'>"
+    assert repr(comp) == "<BinaryOpParameter 'model.p + model.p2'>"
     comp = 1 + p * 2.7**p2
-    assert repr(comp) == "<BinaryOpParameter '(1 + (model.p * (2.7 ** model.p2)))'>"
+    assert repr(comp) == "<BinaryOpParameter '1 + model.p * 2.7 ** model.p2'>"
     comp = np.logaddexp2(p, p2)
-    assert repr(comp) == "<BinaryOpParameter 'numpy.logaddexp2(model.p, model.p2)'>"
+    assert repr(comp) == "<BinaryOpParameter 'numpy.logaddexp2(model.p, (model.p2))'>"
```
hamogu, line 230: This is removing some many brackets, a few more here and there are still worth it. I'm just wandering: Are the new () around `model.p2` design intent or an artifact of the implementation? I'm OK with either (better to keep the implementation simple at the cost of a few extra `()`, I'm just wondering.

sherpa/models/tests/test_parameter.py
```diff
@@ -234,7 +238,135 @@ def func(a, b):
...
+class TestBrackets:
+    """Provide a set of model instances for the tests."""
+
+    a = Parameter('m', 'a', 2)
+    b = Parameter('m', 'b', 3)
+    c = Parameter('m', 'c', 4)
+    d = Parameter('m', 'd', 5)
+
+    # It would be nice to instead use a principled set of states,
+    # but let's just try a somewhat-random set of expressions.
```
hamogu, line 253: If we ever add [hypothesis](https://hypothesis.readthedocs.io/en/latest/) to Sherpa testing, we can go for it ;-)

sherpa/models/tests/test_parameter.py
```diff
@@ -234,7 +238,135 @@ def func(a, b):
...
+                              (abs((a - b) / c), "abs((m.a - m.b) / m.c)"),
+                              (abs((a * b) - c), "abs(m.a * m.b - m.c)"),
+                              (abs((a / b) - c), "abs(m.a / m.b - m.c)"),
+                              (a * abs(b * (c + d)), "m.a * abs(m.b * (m.c + m.d))"),
+                              (abs(b * (c + d)) * (a + d), "abs(m.b * (m.c + m.d)) * (m.a + m.d)"),
+                              (-a, "-(m.a)"),
+                              # (+a, "+(m.a)"),  currently not supported
+                              (-a + b, "-(m.a) + m.b"),
+                              (-a + 2, "-(m.a) + 2"),
+                              # (+a + b, "+(m.a) + m.b"),  currenly not supported
```
hamogu, line 277:
```suggestion
                              # (+a + b, "+(m.a) + m.b"),  currently not supported
                              # (+a + (+b), "+(m.a) + (m.b)"),  currently also not supported
```

sherpa/models/tests/test_parameter.py
```diff
@@ -234,7 +238,135 @@ def func(a, b):
...
+                              ((a ** 2) ** b, "(m.a ** 2) ** m.b"),
+                              (-a ** 2, "-(m.a ** 2)"),
+                              ((-a)**2, "(-(m.a)) ** 2"),
+                              # remainder and integer division, for fun
+                              (a // 2, "m.a // 2"),
+                              ((a + b) // 2, "(m.a + m.b) // 2"),
+                              ((a * b) // 2, "(m.a * m.b) // 2"),
+                              (a % 2, "m.a % 2"),
+                              ((a + b) % 2, "(m.a + m.b) % 2"),
+                              ((a * b) % 2, "(m.a * m.b) % 2")
```
hamogu, line 340: Add something with a function, such as sin or cos?

sherpa/utils/__init__.py
```diff
@@ -4135,3 +4142,156 @@ def send_to_pager(txt, filename=None, clobber=False):
...
+    # as power (aka **) just to make things clear.
+    #
+    lprec = {numpy.power: 4, numpy.remainder: 4,
+             numpy.floor_divide: 4,
+             numpy.multiply: 3, numpy.divide: 3, numpy.true_divide: 3,
+             numpy.add: 2, numpy.subtract: 2}
+
+    rprec = {numpy.power: True, numpy.remainder: True,
+             numpy.floor_divide: True}
+
```
hamogu, line 4182: I suggest moving these definitions to the module level instead of keeping them inside the function. That's for two reasons:
- Performace: Every time this function is called, the dicts are assigned and then garbage collected. Put them into the module level and they will be assigned only once.
- Ability to monkey with it: Though I don't forsee it, if they are the module level, I can modify these dicts if I have some really special case (say, I use parts of sherpa with dax or jax or pytorch tensors instead of numpy arrays). I don't think we have to document that, but it's good to have the possibility to experiment with if needed: `import sherpa.utils; sherpa.utils.lprec[pytorch.operator.power] = 4`

hamogu (changes requested):

sherpa/utils/__init__.py
hamogu: And then say in the docs: "This function uses the module-level dictionaries lprec and rprec."

sherpa/models/tests/test_parameter.py
DougBurke: I believe it is an artifact of the decision I make to not try to be clever with unary-op models, **but** I can not guarantee that. I'll have to look.

sherpa/utils/__init__.py
DougBurke: Maybe there should be a `sherpa.utils.op` module to place this in (name open to bike-shedding), as that makes the scope of such module-level globals a bit-more manageable.

hamogu: That is a good idea!

sherpa/models/tests/test_model_op.py
DougBurke: Off the top of my head, I thing you'd need to add some sort of explicit constructor to group such elements - e.g.

> group(mdl1 + mdl2 + mdl3) + group(foo + bar * baz)

(but with a name other than `group`), as I don't think you can customize `(` and `)` in Python.

I can see the benefit but, as you mentioned, I am willing to live without it.

sherpa/models/tests/test_parameter.py
DougBurke: So, it's an interesting decision in sherpa.utils.get_precedence_rhs which is deciding between ignoring brackets when you have, for a given operator `<op>`

   ... <op> term

For addition and multiplication we don't add the brackets around `term` but for everything else we do. However, this is only relevant for binary expressions written as `lterm <op> rterm`, and `np.logaddexp2` is an example of "do not need ot bother". So I'll have a think about this when I move the code to `sherpa.utils.op_or_some_other_name`

sherpa/models/tests/test_model_op.py
DougBurke: Done

sherpa/models/tests/test_parameter.py
DougBurke: Now I've created `sherpa.models.op` I need to look into this (but wanted to get the updated code tested on CI before trying to do this).

sherpa/models/tests/test_model_op.py
DougBurke: Some tests cover the https://github.com/sherpa/sherpa/pull/1972#discussion_r1545433958 case in more detail, so if I "fix it" we'll see the changes in a few places.

sherpa/utils/__init__.py
DougBurke: This is now in `sherpa.models.op` and they are module-level values, referenced from the documentation.

sherpa/models/tests/test_parameter.py
DougBurke: I've special-cased the `BinaryOpParameter` so that it only bothers with the bracket removal for infix operators, since prefix operators like `np.foobar(lhs, rhs)` do not need special-casing the `lhs` and `rhs` terms.

sherpa/models/tests/test_model_op.py
DougBurke: Which have now been changed.

DougBurke:
So, I believe I've addressed the existing questions. 

I have made some changes based on work I've done in 
- #1974 

which adds some basic typing rules to the model and parameter modules, as well as replacing numpy by np. I don't think they particularly help this particular PR but are an attempt to improve the typing of the modules so we can do more later.

sherpa/models/model.py
```diff
@@ -1070,19 +1083,30 @@ class ArithmeticConstantModel(Model):
...
-                nstr = ','.join([str(s) for s in val.shape])
-                name = f'{val.dtype.name}[{nstr}]'
+                if TYPE_CHECKING:
+                    assert isinstance(val, np.ndarray)
+
+                # For some reason mypy didn't like
+                # '[str(s) for s in store.shape]`
+                dims = map(str, store.shape)
+                nstr = ','.join(dims)
+                name = f'{store.dtype.name}[{nstr}]'
```
hamogu, line 1104: This is complicated, just to make a type checker happy - and that in a dynamically typed language. But fine.

hamogu (approved):

sherpa/models/model.py
DougBurke: No, this was not my favourite bit of code...

PR #1973, opened 2024-02-28
Fix the requires_pylab decorator

### LINKED ISSUE #1971: requires_pylab decorator is wrong
I added this test to `sherpa/ui/tests/test_ui_plot.py`:

```
@requires_pylab
def test_this_should_fail():
    assert False
```

and then try to run it with

```
% pytest sherpa/ui/tests/test_ui_plot.py::test_this_should_fail
pytest-xvfb could not find Xvfb. You can install it to prevent windows from being shown.
============================================== test session starts ===============================================
platform linux -- Python 3.11.4, pytest-7.4.0, pluggy-1.2.0
rootdir: /home/dburke/sherpa/sherpa-xspec
configfile: pytest.ini
plugins: xvfb-3.0.0, doctestplus-1.0.0, anyio-4.0.0, cov-4.1.0, xdist-3.5.0
collected 0 items                                                                                                

================================================ warnings summary ================================================
../../micromamba/envs/sherpa-xspec/lib/python3.11/site-packages/_pytest/mark/structures.py:350
  /home/dburke/micromamba/envs/sherpa-xspec/lib/python3.11/site-packages/_pytest/mark/structures.py:350: PytestCollectionWarning: cannot collect 'test_plot_xxx_components_simple' because it is not a function.
    def __call__(self, *args: object, **kwargs: object):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================== 1 warning in 0.47s ===============================================
ERROR: not found: /home/dburke/sherpa/sherpa-xspec/sherpa/ui/tests/test_ui_plot.py::test_this_should_fail
(no name '/home/dburke/sherpa/sherpa-xspec/sherpa/ui/tests/test_ui_plot.py::test_this_should_fail' in any of [<DocTestModulePlus sherpa/ui/tests/test_ui_plot.py>, <Module sherpa/ui/tests/test_ui_plot.py>])

```

If you remove the `@requires_pylab` line then the test is run (and fails, as expected).

The problem actually is reported by `PytestCollectionWarning`: it doesn't think this is a function. This is because `requires_pylab` is written to be a decorator but actually behaves more like a pytest fixture:

```
   def requires_pylab(test_function):
        """Runs the test with the PylabBackend plotting backend if available,
        otherwise skips the test.
        """
        pylab_backend = pytest.importorskip("sherpa.plot.pylab_backend")
        plt = pytest.importorskip("matplotlib.pyplot")

        with TemporaryPlottingBackend(pylab_backend.PylabBackend()):
            yield

        plt.close(fig="all")
```

Note that it does nothing with `test_function`!

There's actually other things we could look at here too:

- `sherpa.utils.testing` allows `pytest` to be optional, but really it is required for running tests, but we don't want to require `pytest` jus tto use sherpa
- all the `requires_xxx` routines in this module should probably be moved to `sherpa/conftest.py`, which might simplify this

DougBurke:
# Summary

Change the testing code to better-handle those tests that want to check plotting when using the pylab backend. Fix #1971.

# Details

As noted in #1971 the current requires_pylab decorator, which is meant to only run tests when matplotlib is available, and then run the test with the pylab plot backend selected, does not work.

This is because it's trying to be both a decorator and a generator.

The easiest way I have found to do this is to move from requires_pylab being a decorator defined in sherpa.utils.testing to being a pytest fixture defined in sherpa/conftest.py. Actually, I couldn't easily come up with a different solution. That's not to mean there isn't one, but mixing decorators, generators, and pytest is a recipe for disaster so I wanted to make it as simple as I could understand.

I think there's a strong argument to be had to make these tests fixtures rather than decorators, but I am only doing it for requires_pylab because

- it is the only one that needs this
- there are not many users of requires_pylab so it is an easy-to-make change
- requires_pylab is a fairly-new invention so there's less likely to be downstream problems with this change (since it is simpler to remove the sherpa.utils.testing.requires_pylab symbol to make it obvious what is going on)
- doing the other symbols
  - would take too long
  - although I think it would be better, there may be some cases where we would want to retain the decorator

This should technically improve test coverage as it likely runs more code now. Some tests have had been to be updated because they were not actually correct.

This will cause fun for some existing PRs, such as 
- #1679 
I think

# Example

The `main` branch has this many tests

```
=================== 8920 passed, 74 skipped, 99 xfailed, 4 xpassed, 2 warnings in 541.45s (0:09:01) ===================
```

This branch has

```
=================== 9106 passed, 74 skipped, 99 xfailed, 4 xpassed, 2 warnings in 554.39s (0:09:14) ===================
```

That is, we "gain" 186 tests.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -869,3 +869,20 @@ def xsmodel():
         return cls(mname)
 
     return func
+
+
+# Fixtures that control access to the tests, based on the availability
+# of external "features" (normally this is the presence of optional
+# modules). These used to be decorators in sherpa.utils.testing.
+#
+@pytest.fixture
+def requires_pylab():
+    """Runs the test with the pylab plotting backend if available."""
+
+    pylab_backend = pytest.importorskip("sherpa.plot.pylab_backend")
+    plt = pytest.importorskip("matplotlib.pyplot")
+
+    with TemporaryPlottingBackend(pylab_backend.PylabBackend()):
+        yield
+
+    plt.close(fig="all")
--- a/docs/overview/testing.rst
+++ b/docs/overview/testing.rst
@@ -16,6 +16,5 @@ The sherpa.utils.testing module
       requires_fits
       requires_group
       requires_package
-      requires_pylab
       requires_stk
       requires_xspec
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -48,7 +48,7 @@ from sherpa.models.template import create_template_model
 
 from sherpa.utils.err import ArgumentTypeErr, DataErr, IdentifierErr, ModelErr, PlotErr
 from sherpa.utils.testing import requires_data, requires_fits, \
-    requires_xspec, requires_pylab, requires_wcs
+    requires_xspec, requires_wcs
 
 import sherpa.ui.utils
 import sherpa.astro.ui.utils
@@ -1086,7 +1086,6 @@ def check_bkg_source(plotfunc, idval, isfit=True):
     assert np.all(plot.y >= 0)
 
 
-@requires_pylab
 @pytest.mark.parametrize("idval", [None, 1, "one", 23])
 @pytest.mark.parametrize("plotfunc,checkfuncs",
                          [(ui.plot_bkg_delchi, [check_bkg_resid]),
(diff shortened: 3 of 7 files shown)


sherpa/conftest.py
```diff
@@ -869,3 +869,20 @@ def func(name, mname=None):
...
+# modules). These used to be decorators in sherpa.utils.testing.
+#
+@pytest.fixture
+def requires_pylab():
+    """Runs the test with the pylab plotting backend if available."""
+
+    pylab_backend = pytest.importorskip("sherpa.plot.pylab_backend")
+    plt = pytest.importorskip("matplotlib.pyplot")
+
+    with TemporaryPlottingBackend(pylab_backend.PylabBackend()):
```
DougBurke, line 885: Shouldn't I have to import this symbol from somewhere?

There's too much magic in pytest!

DougBurke: Actually, we already explicitly import this symbol for other pytest fixtures, so it's all good.

hamogu (approved):
Mmm, I wonder why that did not work. I'm pretty sure I checked that the number of tests did not go down when I modified the backend stuff. Or maybe I messed that up back then. Or maybe something else (like the pytest version) changes that.

Either way, this is cleaner, actually reduces the number of lines, and makes it work now, so there is no need to do archeology to find out when it stopped working.

PR #1974, opened 2024-02-29
Allow linked parameters to be fit without including them in the model expression

### LINKED ISSUE #777: track linked model parameters not included in source expressions
There are times when I want to transform a model parameter - e.g. change a FWHM to a sigma, or use a log rather than linear scaling - which you can do with a linked parameter. For instance

```python
sigmdl = create_model_component('const1d', 'sigmdl')
mdl.fwhm = 2 * np.sqrt(2 * np.log(2)) * sigmdl.c0  # assuming I've got the constant correct
```

However, if you try to fit this it won't recognize `c0` as a parameter that need varying, and so won't vary the `mdl.fwhm` parameter.

You can work around this by including the extra model in the source expression, multiplying by 0 so that it doesn't do anything - for example

```python
set_source(mdl + 0 * sigmdl)
```

but this is not ideal.

Can we identify these "extra" parameters automatically when doing a fit (we obviously can, it's just how hard would it be to do so).

DougBurke:
# Summary

Treat linked parameters as part of the model expression (via the new `lpars` attribute and `get_thawed_pars` routine) so that they can be included in a fit without including the linked model as part of the model expression. Fix #777

# Details

We would like to be able to fit a model with linked parameters - for instance (using the ui layer here)

    set_source(gauss1d.gmdl + polynom1d.bgnd)
    create_model_component('scale1d', 'sigma')
    gmdl.fwhm = 2 * np.sqrt(2 * np.log(2)) * sigma.c0

where we actually want to fit the `sigma` value of the gaussian rather than the `fwhm`. If you try this you will see that the `fwhm` value never changes, because it has become a frozen parameter and Sherpa doesn't `know` about the `c0` term (when fitting).

For reference we have

```
>>> mdl = get_source()
>>> print(mdl)
(gauss1d.gmdl + polynom1d.bgnd)
(gauss1d.gmdl + polynom1d.bgnd)
   gmdl.fwhm    linked      2.35482 expr: numpy.multiply(2.3548200450309493, sigma.c0)           
   gmdl.pos     thawed            0 -3.40282e+38  3.40282e+38           
   gmdl.ampl    thawed            1 -3.40282e+38  3.40282e+38           
   bgnd.c0      thawed            1 -3.40282e+38  3.40282e+38           
   bgnd.c1      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c2      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c3      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c4      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c5      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c6      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c7      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.c8      frozen            0 -3.40282e+38  3.40282e+38           
   bgnd.offset  frozen            0 -3.40282e+38  3.40282e+38           

>>> mdl.pars
(<Parameter 'fwhm' of model 'gmdl'>,
 <Parameter 'pos' of model 'gmdl'>,
 <Parameter 'ampl' of model 'gmdl'>,
 <Parameter 'c0' of model 'bgnd'>,
 <Parameter 'c1' of model 'bgnd'>,
 <Parameter 'c2' of model 'bgnd'>,
 <Parameter 'c3' of model 'bgnd'>,
 <Parameter 'c4' of model 'bgnd'>,
 <Parameter 'c5' of model 'bgnd'>,
 <Parameter 'c6' of model 'bgnd'>,
 <Parameter 'c7' of model 'bgnd'>,
 <Parameter 'c8' of model 'bgnd'>,
 <Parameter 'offset' of model 'bgnd'>)
>>> mdl.thawedpars
[0.0, 1.0, 1.0]
>>> [p.fullname for p in mdl.pars if not p.frozen]
['gmdl.pos', 'gmdl.ampl', 'bgnd.c0']
```

So we can see that Sherpa doesn't think of `gmdl.fwhm` as a parameter to be fit. This is issue #777

The current work around is to include the model containing the linking parameter into the model expression, but multiplied by 0 so it adds no signal. This works but

- is ugly
- is easy to forget
- requires that we use a model for the linking parameter with the same dimensionality as the model terms.

As an example of this workaround is

```
>>> mdl2 = mdl + 0 * sigma
>>> mdl2.thawedpars
[0.0, 1.0, 1.0, 1.0]
>>> sigma.c0 = 10
>>> mdl2.thawedpars
Out[13]: [0.0, 1.0, 1.0, 10.0]
>>> [p.fullname for p in mdl2.pars if not p.frozen]
['gmdl.pos', 'gmdl.ampl', 'bgnd.c0', 'sigma.c0']
```

I have tried several possible solutions, such as extending `pars` to include the linked parameters, or doing the deconstruction in the `sherpa.fit` module, but they are not ideal since

a) the `.pars` attribute of a models used to send the parameter values to the model calculation code, and this does not want (or need) the linking parameters sent in; adding in the linking parameters here just complicated the existing model code and would likely **break** existing user models

b) we need some of the logic in the model class to answer questions that are related to fits, but does not need creating a `Fit` instance

So, first we add a bunch of tests related to #777 and issues I came across when developing this approach, and being side-tracked by earlier attempts and finding issues like 
- #1978
- #1980 
- #1981 

There are minor code clean ups of `model.py` and `parameter.py` after a bunch of tests are added (some of these tests repeat existing tests, but it's useful to have them in one place so we can check how the parameter handling changes, or does not; these tests are also not expensive to run).

The first significant change is making sure `ConstantParameter` objects are marked as `Frozen`. These are used to create the wrappers around the numeric terms in expressions like 2.345 * sigma.c0 - where the 2.345 gets wrapped up in a `ConstantParameter` term. I had thought I would need to do this to make it easy to extract thawed parameters in parameter expressions, but it turns out that for some reason `ConstantParameter` extends `CompositeParameter`, so we can just use a check like `not isinstance(term, CompositeParameter)` to ignore constant terms. However, I have kept this change in as I think it makes sense.

We then add some typing rules to the files in `sherpa/models`. There's the usual balance of trying to describe the needed types without going overboard: for example, should we treat parameter lists as `Sequence[SupportsFloat]` or some other protocol? I have generally tried to go for a combination of: a) does mypy not complain, and b) does it not seem too outlandish.

There's another bunch of tests - these are added to just check #1982 since I discovered them whilst writing this PR and I have spent enough time rebasing things so the tests come first.

To fix #777 we

- add a `get_thawed_pars` method to the `Model` class which makes it obvious what we are doing, and then take advantage of this (in particular in `sherpa.fit`)
  - as part of this the `pars` field is now a property (so stopping users from seriously borking their models)
    - I have some tests of the pars field as the behaviour surprised me at first, particularly when it came to having repeated models and how the repeated parameters are handled 
  - using `get_thawed_pars` let me clean up some logic in the `sherpa.fit.Fit` class related to the error analysis, where we need to switch the parameters being frozen; the current logic could have created a list of thawed parameters and then used them but instead we created a set of "adapter" values (the `thaw_indices` fields) which were used to map between the number of thawed parameters and the number of all parameters, and had previously caused confusion -e.g. #342 

- provide access to the linked parameters via the `lpars` attribute: this is similar to `pars` but it doesn't duplicate attributes and only contains parameters that are not in the `pars` field (so if `m1.ypos` is linked to `m1.xpos` and `m1.ampl` is linked to `m2.c0` then `m1.lpars` will only contain `m2.c0`); this field also drops the non-parameter elements, such as ArithmeticConstant parameters
 
- we then switch `get_thawed_pars` to returning the thawed parameters in `pars` + `lpars` 
  - with all the preparatory work, this is actually a relatively small change to get #777 fixed

I note that I have not tried to address 

- #1981 

but this is an existing issue (i.e. users have this issue in essentially all versions of Sherpa) rather than one I've introduced here.

### DIFF
--- a/docs/models/index.rst
+++ b/docs/models/index.rst
@@ -13,7 +13,7 @@ section::
 
     >>> import numpy as np
     >>> import matplotlib.pyplot as plt
-    >>> from sherpa import models
+    >>> from sherpa.models import basic
 
 Creating a model instance
 =========================
@@ -22,7 +22,7 @@ Models must be created before there parameter values can
 be set. In this case a one-dimensional gaussian using the
 :py:class:`~sherpa.models.basic.Gauss1D` class::
 
-    >>> g = models.Gauss1D()
+    >>> g = basic.Gauss1D()
     >>> print(g)
     gauss1d
        Param        Type          Value          Min          Max      Units
@@ -51,7 +51,7 @@ of the class name.
 
     >>> g.name
     'gauss1d'
-    >>> h = models.Gauss1D('other')
+    >>> h = basic.Gauss1D('other')
     >>> print(h)
     other
        Param        Type          Value          Min          Max      Units
@@ -78,8 +78,8 @@ plus a flat background - using the
 :py:class:`~sherpa.models.basic.Const1D` class - would be
 represented by the following model::
 
-    >>> src1 = models.Gauss1D('src1')
-    >>> back = models.Const1D('back')
+    >>> src1 = basic.Gauss1D('src1')
+    >>> back = basic.Const1D('back')
     >>> mdl1 = src1 + back
     >>> print(mdl1)
     (src1 + back)
@@ -93,7 +93,7 @@ represented by the following model::
 Now consider fitting a second dataset where it is known that the background
 is two times higher than the first::
 
-    >>> src2 = models.Gauss1D('src2')
+    >>> src2 = basic.Gauss1D('src2')
     >>> mdl2 = src2 + 2 * back
     >>> print(mdl2)
     (src2 + (2 * back))
@@ -341,8 +341,8 @@ Linking parameters also reduces the number of free parameters in a fit.
 
 The following examples use the same two model components::
 
-    >>> g1 = models.Gauss1D('g1')
-    >>> g2 = models.Gauss1D('g2')
+    >>> g1 = basic.Gauss1D('g1')
+    >>> g2 = basic.Gauss1D('g2')
 
 Linking parameter values requires referring to the parameter, rather
 than via the :py:attr:`~sherpa.models.parameter.Parameter.val` attribute.
@@ -398,34 +398,55 @@ Including another parameter
 ---------------------------
 
 It is possible to include other parameters in a link expression,
-which can lead to further constraints on the fit. For instance,
-rather than using a fixed separation, a range can be used. One
-way to do this is to use a :py:class:`~sherpa.models.basic.Const1D`
-model, restricting the value its one parameter can vary.
+which can lead to further constraints on the fit. For example,
+we can fit using the sigma value instead of the FWHM of a
+gaussian::
 
-::
-
-    >>> sep = models.Const1D('sep')
-    >>> print(sep)
+    >>> sigma = basic.Scale1D('sigma')
+    >>> sigma.c0 = 10
+    >>> print(sigma)
     sep
        Param        Type          Value          Min          Max      Units
        -----        ----          -----          ---          ---      -----
-       sep.c0       thawed            1 -3.40282e+38  3.40282e+38
-    >>> g2.fwhm = g1.fwhm + sep.c0
-    >>> sep.c0 = 1200
-    >>> sep.c0.min = 800
-    >>> sep.c0.max = 1600
+       sigma.c0     thawed           10 -3.40282e+38  3.40282e+38
+    >>> g1.fwhm = 2 * np.sqrt(2 * np.log(2)) * sigma.c0
 
-In this example, the separation of the two components is restricted
-to lie in the range 800 to 1600.
+which creates
 
-In order for the optimiser to recognize that it needs to vary the
-new parameter (``sep.c0``), the component *must* be included in the
-model expression. As it does not contribute to the model output
-directly, it should be multiplied by zero. So, for this example
-the model to be fit would be given by an expression like::
+::
 
-   >>> mdl = g1 + g2 + 0 * sep
+    >>> print(g1)
+    g1
+       Param        Type          Value          Min          Max      Units
+       -----        ----          -----          ---          ---      -----
+       g1.fwhm      linked      23.5482 expr: numpy.multiply(2.3548200450309493, sigma.c0)
+       g1.pos       thawed         1200 -3.40282e+38  3.40282e+38
+       g1.ampl      thawed            1 -3.40282e+38  3.40282e+38
+
+and, because ``g2.fwhm`` is still linked to ``g1.fwhm``
+
+::
+
+    >>> print(g2)
+    g2
+       Param        Type          Value          Min          Max      Units
+       -----        ----          -----          ---          ---      -----
+       g2.fwhm      linked      23.5482            expr: g1.fwhm
+       g2.pos       linked         9434    expr: (g1.pos + 8234)
+       g2.ampl      thawed            1 -3.40282e+38  3.40282e+38
+
+.. note::
+
+   Prior to Sherpa 4.16.1 you had to explicitly include any linked
+   parameters into the model expression - e.g. by saying::
+
+       >>> mdl = g1 + g2 * 0 * sigma
+
+   where the ``sigma`` component is multiplied by zero to ensure it
+   does not directly add to the model. This step is **no longer**
+   needed, so you can just fit the model directly.
+
+       >>> mdl = g1 + g2
 
 Complex functional relationships
 --------------------------------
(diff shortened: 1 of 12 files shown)


DougBurke:
To validate this, I use the example data from https://sherpa.readthedocs.io/en/4.16.0/examples/fit_peaked_data.html with the script

```python
import numpy as np

import pytest

from sherpa.fit import Fit
from sherpa.io import read_data
from sherpa.models.basic import NormGauss1D, Scale1D
from sherpa.optmethods import NelderMead

stat = LeastSq()
opt = NelderMead()

d = read_data('test_peak.dat')

mfwhm = NormGauss1D("fwhm")
s0 = mfwhm.fwhm.val  # the starting value
ffwhm = Fit(d, mfwhm, stat, opt)
print(ffwhm.fit())

msigma = NormGauss1D("sigma")
sigma = Scale1D("sigma")
scale = 2 * np.sqrt(2 * np.log(2))
msigma.fwhm = scale * sigma.c0

# Adjust sigma to roughly match the starting position of ffwhm
sigma.c0 = s0 / scale

fsigma = Fit(d, msigma, stat, opt)
print(fsigma.fit())

# compare the results
#
assert msigma.pos.val == pytest.approx(mfwhm.pos.val)
assert msigma.ampl.val == pytest.approx(mfwhm.ampl.val)
assert msigma.fwhm.val == pytest.approx(mfwhm.fwhm.val)
```

# With CIAO 4.16

If I run this with CIAO 4.16 then the output shows both fits succeed but the linked one (the `fsigma` fit) does not try to fit the FWHM value:

```
datasets       = None
itermethodname = none
methodname     = neldermead
statname       = leastsq
succeeded      = True
parnames       = ('fwhm.fwhm', 'fwhm.pos', 'fwhm.ampl')
parvals        = (2.9015671244735737, 9.242770560130534, 30.313543791653892)
statval        = 29.994315740119298
istatval       = 4222.73002033349
dstatval       = 4192.73570459337
numpoints      = 401
dof            = 398
qval           = None
rstat          = None
message        = Optimization terminated successfully
nfev           = 439
datasets       = None
itermethodname = none
methodname     = neldermead
statname       = leastsq
succeeded      = True
parnames       = ('sigma.pos', 'sigma.ampl')
parvals        = (9.237065152694997, 43.11339431561573)
statval        = 1770.9210824770857
istatval       = 4222.73002033349
dstatval       = 2451.808937856404
numpoints      = 401
dof            = 399
qval           = None
rstat          = None
message        = Optimization terminated successfully
nfev           = 223
  File "/home/dburke/sherpa/sherpa-xspec/docs/_examples/examples/fit_peaked_data/links.py", line 34, in <module>
    assert msigma.pos.val == pytest.approx(mfwhm.pos.val)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

# With this PR

We can see both fits contain three parameters and the final statistic value is similar and the same FWHM value is found (modulo numerical differences) even though we fit the sigma value the second time:

```
datasets       = None
itermethodname = none
methodname     = neldermead
statname       = leastsq
succeeded      = True
parnames       = ('fwhm.fwhm', 'fwhm.pos', 'fwhm.ampl')
parvals        = (2.9015671244735737, 9.242770560130534, 30.313543791653892)
statval        = 29.994315740119298
istatval       = 4222.73002033349
dstatval       = 4192.73570459337
numpoints      = 401
dof            = 398
qval           = None
rstat          = None
message        = Optimization terminated successfully
nfev           = 439
datasets       = None
itermethodname = none
methodname     = neldermead
statname       = leastsq
succeeded      = True
parnames       = ('sigma.pos', 'sigma.ampl', 'sigma.c0')
parvals        = (9.242770559875336, 30.313543791792384, 1.2321821074081332)
statval        = 29.99431574011929
istatval       = 4222.73002033349
dstatval       = 4192.73570459337
numpoints      = 401
dof            = 398
qval           = None
rstat          = None
message        = Optimization terminated successfully
nfev           = 435
```

DougBurke:
@anetasie - I know you have used linked parameters before so can you review this for behaviour and documentation?

DougBurke:
@anetasie - I have finished reworking this (which was just to address conflicts with #1972 which will hopefully be merged soon).

DougBurke:
Rebased as #1972 has been merged

anetasie:
@DougBurke this looks good in the example that I ran. I was also looking at linking parameters across the multiple data sets at it seems fine. It took me awhile to understand what you were doing.

anetasie (approved):

DougBurke:
Rebased to address conflicts in the tests (the system couldn't tell this was just adding to the end of the file and so threw up a conflict warning).

PR #1977, opened 2024-03-01
model: improve examples in docstrings

DougBurke:
# Summary

Allow model.py, parameter.py, and regrid.py to be included in the docstring pytest run.

# Details

I have several PRs that work on the model code so it would be good to get this in first - e.g.

- #1679 
- #1972 
- #1974

This involved

- tweaking a couple of examples to actually pass (since the environment they run in is a bit different from how they were created)

- change the choice of variables just to avoid having to create new objects just for the test

- hide some tests where it's just not worth creating objects used in the test (e.g. too complex to create the necessary state and it would hide from the actual example).

Note that one test in regrid.py is technically not-useful but this has already been noted as issue #840 so the example is kept (but we now note the warning message that is created), as this PR is just about testing the existing code.

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023
+#  Copyright (C) 2010, 2016 - 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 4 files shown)


sherpa/models/parameter.py
```diff
@@ -153,7 +158,7 @@
 such as being twice the other parameter:
 
     >>> q = Parameter('other', 'beta', 4)
-    >>> p.val = 2 * p
```
hamogu, line 156: This is exactly want I wanted to achieve with testing the docstrings. When I read this, I was wondering "Oh, what does this do? Is that not circular p.val = 2 * p"? Then, I say that that's one the lines fixed here. We don't want to confuse users with examples that don't work, so it's good to fix them as we get to it.

sherpa/models/model.py
```diff
@@ -351,6 +352,15 @@ def calc(self, p, *args, **kwargs):
            'ArithmeticFunctionModel', 'NestedModel', 'MultigridSumModel')
 
 
+# These tests refer to other variables which are just not worth
+# setting up (or would require too much work to create any useful
+# state, such as the cache statis lines).
```
hamogu, line 357: Should this say the following?
```suggestion
# state, such as the cash statisic lines).
```

hamogu (approved):

sherpa/models/model.py
DougBurke:
```suggestion
# state, such as the cache status lines).
```

I was out by one vowel.

PR #1984, opened 2024-03-11
Rework x errorbar support to better-match community expectations

### LINKED ISSUE #1985: xerr for PHA data does not recognize the wavelength setting
Found when working on a fix to #1817 

The 4.16.0 release (and for a long time before):

```
sherpa In [1]: load_pha("3c273.pi")
read ARF file 3c273.arf
read RMF file 3c273.rmf

sherpa In [3]: notice(0.5, 6); fit()
dataset 1: 0.00146:14.9504 -> 0.4672:6.57 Energy (keV)
Dataset               = 1
Initial fit statistic = 5.43145e+10
Final fit statistic   = 21.4837 at function evaluation 265
Data points           = 41
Degrees of freedom    = 38
Probability [Q-value] = 0.985761
Reduced statistic     = 0.56536
Change in statistic   = 5.43145e+10
   gal.nH         0.0320727    +/- 0.11091     
   pl.gamma       1.98673      +/- 0.183611    
   pl.ampl        0.000194604  +/- 1.83336e-05 

sherpa In [4]: get_data_plot().xerr
Out[4]: 
array([0.1022    , 0.07300001, 0.05839998, 0.0438    , 0.0438    ,
       0.02920002, 0.0438    , 0.02920002, 0.05839998, 0.0438    ,
       0.0438    , 0.05840003, 0.0438    , 0.05839992, 0.08760011,
       0.11679995, 0.07299995, 0.13139999, 0.08760011, 0.11679995,
       0.08759999, 0.0438    , 0.08760011, 0.05839992, 0.10220003,
       0.08759999, 0.11679983, 0.18980002, 0.13140011, 0.14599991,
       0.2190001 , 0.30660009, 0.17519999, 0.2335999 , 0.23360014,
       0.21899986, 0.4671998 , 0.30660009, 0.35039997, 0.52559996,
       0.67160034])

sherpa In [5]: set_analysis("channel")
dataset 1: 33:450 Channel

sherpa In [6]: get_data_plot().xerr
Out[6]: 
array([ 7.,  5.,  4.,  3.,  3.,  2.,  3.,  2.,  4.,  3.,  3.,  4.,  3.,
        4.,  6.,  8.,  5.,  9.,  6.,  8.,  6.,  3.,  6.,  4.,  7.,  6.,
        8., 13.,  9., 10., 15., 21., 12., 16., 16., 15., 32., 21., 24.,
       36., 46.])

sherpa In [7]: set_analysis("wave")
dataset 1: 1.88713:26.5377 Wavelength (Angstrom)

sherpa In [8]: get_data_plot().xerr
Out[8]: 
array([0.1022    , 0.07300001, 0.05839998, 0.0438    , 0.0438    ,
       0.02920002, 0.0438    , 0.02920002, 0.05839998, 0.0438    ,
       0.0438    , 0.05840003, 0.0438    , 0.05839992, 0.08760011,
       0.11679995, 0.07299995, 0.13139999, 0.08760011, 0.11679995,
       0.08759999, 0.0438    , 0.08760011, 0.05839992, 0.10220003,
       0.08759999, 0.11679983, 0.18980002, 0.13140011, 0.14599991,
       0.2190001 , 0.30660009, 0.17519999, 0.2335999 , 0.23360014,
       0.21899986, 0.4671998 , 0.30660009, 0.35039997, 0.52559996,
       0.67160034])
```

Notice that the `xerr` values for wavelength are the same as for energy (but changed for channel). 

So, `xerr` isn't correct when wavelength analysis is in use.

However, we don't really use the `xerr` field other than in a few "odd" cases - such as #1817 - and it really is not clear what it's really meant to be. So I view the downstream consequences of this as "not really affecting" people (but I don't do much grating analysis so may be biased).

DougBurke:
# Summary

Improve the "x errorbar" handling for histogram-style data (including PHA), including support for the wavelength setting (fix #1985), to better match community expectations (e.g. #1817).

# Details

I have hit the point where I have several PRs changing the same code and so re-doing the same "helper-code" cleanup. So I want to get this merged first to avoid doing it differently in different PRs. I did end up finding one issue to address (#1985) and technically this addresses #1817 but I have a follow-up PR planned to do this "better".

These changes are mainly clean ups to the plot code but the actual fix is in the data class (although the `xerr` setting is only really used by the plot class). I don't really want to spend much time on this PR (e.g. expanding the routines that get docstrings added) as other parts of the `sherpa.plot` module are going to change and additions are better placed in later PRs.

Per-commit changes are:

- adds some simple tests of the str output of some classes I found had not been covered (or not as well as could be)
- add tests of the `xerr` handling
  - it is not-a-all clear what semantics we intend for the `xerr` field. I'd go as far as to say it is there just to allow us to draw "error bars" on the independent axis for PHA data
  - so I don't have any compunction in changing it's value (e.g., as discussed below, now returning the half-width) since this is what the community assumes it to be 
- tweak how the tests run when bokeh is available but matplotlib isn't (it would fail)
- "import numpy as np" rather than "import numpy" as it annoys me and it also makes the lines shorter
  - this is not-so important just yet but is more an issue once we add typing rules
- tweak how the jointplot code access plot preferences
  - use the "helper" routine rather than assume we know the plot is a "plot" or "histogram" class
  - this is not needed in this PR but is a useful change, and will be needed in a follow-up PR
- add typing rules to the plot backend handling
  - this changes the ordering of when `backend` is created to make sure it never stores a `None`, which (if we didn't it), would cause all typing code using backend to have to note that it's not None. This way we never construct a None and so we are all safer.
  - as part of this I've tweaked the context-manager code to return the values it should (although obviously it wasn't an issue that it hadn't been returning them)
  - I also have a small tweak needed to enforce a "float" value; I thought I'd addressed this as part of #1959 but obviously not
- re-order the sherpa imports
- make more use of super, although not in all cases as some of our multi-parent classes are "interesting"
- added an explicit check of issue #1986 (these tests already implicity tested the current behaviour but I wanted a simple explicit test so if we wver change this we can easily see in the tests).
- try to make our `xerr` handling sensible
  - rather than `xhi - xlo` return `(xhi - xlo) / 2` which is what the community expects it to be (e.g. see #1817)
  - previously `DataPHA` data returned keV values when `units=wavelength`, which is just wrong (#1985) so fix this, which leads into wondering about how the plot classes handle "wavelength" setting (since in this case it appears `xhi < xlo`), but this is **out of scope** for this PR

Some of this code is taken from #1679.

# Areas not looked at

This is an incomplete list, but I just want to mention them in case review suggests working on them here. As mentioned, I have follow-on PRs where some of these can be addressed.

- the `__str__` methods contain a lot of repeated code; I will address this in a future PR
- because of this I've limited the amount of f-string changes in this PR ...
- adding typing statements to the plot classes
  - this would add a lot of "noise" to this particular PR, which is noisy enough as is
  - it actually shows some issues with our class hierarchy (see next point)
- the class hierarchy is "interesting", such as some classes inheriting from both "point" and "histogram" classes and the `plot` method changing drastically between parent and child classes
  - some of this is addressed in future PRs
  - we could remove the `xerr` field, for instance, but I think this is too large a change at this point in time (since it's in both data and plot classes)
  - the data classes have a `to_plot` method which could be looked into but that is also a large change which is separate from this work

# Example

## CIAO 4.16

This is a fit using `xsphabs.gal * xspowerlaw.xpl` to our standard `3c273.pi` dataset, and then `set_analysis("wavelength")`. Here's the output of

```python
plot_fit(xlog=True)
plot_resid(overplot=True)
plt.xlim(1.8, 30)
```

err1

Note that the x errorbars on the residuals (green) are

a) overlapping at low wavelengths
b) much too small at large wavelengths

This is because they are actually the error bars calculated using the energy setting.

## This PR

err2

We can see that the x errorbars now match the model bins (i.e. the widths agree).

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023
+#  Copyright (C) 2008, 2015 - 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -4263,6 +4263,30 @@ It is an integer or string.
         return self._fix_y_units(err, filter, response_id)
 
     def get_xerr(self, filter=False, response_id=None):
+        """Returns an X "error".
+
+        The error value for the independent axis is not well defined
+        in Sherpa.
+
+        .. versionchanged:: 4.16.1
+           The return value is now half the bin width instead of the
+           full bin width and is now calculated correctly when the
+           analysis is set to "wavelength".
+
+        Parameters
+        ----------
+        filter : bool, optional
+           Should the values be filtered to the current notice range?
+        response_id : int or None, optional
+           What response should be used?
+
+        Returns
+        -------
+        xerr : ndarray
+           The half-width of each bin (or group) in the current
+           analysis units.
+
+        """
         if bool_cast(filter):
             # If we apply a filter, make sure that
             # ebins are ungrouped before applying
@@ -4280,7 +4304,14 @@ It is an integer or string.
                 #
                 return numpy.asarray([])
 
-        return ehi - elo
+        # Issue #748 #1817 noted we should return the half-width
+        # Issue #1985 notes that we need to support wavelength.
+        #
+        if self.units != "wavelength":
+            return (ehi - elo) / 2
+
+        dlam = hc / elo - hc / ehi
+        return dlam / 2
 
     def get_ylabel(self):
         ylabel = 'Counts'
@@ -5399,9 +5430,7 @@ class DataIMG(Data2D):
         return 'x0'
 
     def get_x1label(self):
-        """
-        Return label for second dimension in 2-D view of independent axis/axes
-        """
+        "Return label for second dimension in 2-D view of independent axis/axes."
         if self.coord == 'physical':
             return 'x1 (pixels)'
         if self.coord == 'world':
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -604,7 +604,7 @@ class BkgModelPHAHistogram(ModelPHAHistogram):
     """
 
     def __init__(self):
-        ModelPHAHistogram.__init__(self)
+        super().__init__()
         self.title = 'Background Model Contribution'
 
 
@@ -619,7 +619,7 @@ class BkgModelHistogram(ModelHistogram):
     """
 
     def __init__(self):
-        ModelPHAHistogram.__init__(self)
+        super().__init__()
         self.title = 'Background Model Contribution'
 
 
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -237,11 +237,11 @@ def test_pha_get_xerr_all_bad_channel_no_group():
     pha = DataPHA('name', [1, 2, 3], [1, 1, 1],
                   quality=[2, 2, 2])
 
-    assert pha.get_xerr() == pytest.approx([1, 1, 1])
+    assert pha.get_xerr() == pytest.approx([0.5, 0.5, 0.5])
 
     pha.ignore_bad()
     assert pha.get_filter() == ''
-    assert pha.get_xerr() == pytest.approx([1, 1, 1])
+    assert pha.get_xerr() == pytest.approx([0.5, 0.5, 0.5])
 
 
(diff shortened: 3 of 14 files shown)


DougBurke:
@hamogu I've finally finished with this PR (I was spending time trying to clean up and reorganize the code, but I think I've done all I want on this part of the code as I want to work on the actual improvements).

hamogu (approved):

DougBurke:
Ta @hamogu

PR #1976, opened 2024-03-01
Fix bug in multi-panel plotting in bokeh

hamogu:
## Summary
Fix bug in multi-panel plotting in bokeh

## Details

There are in fact two separate problems:

- The current figure sis a GridPlot and no longer (as originally envisioned by me) a plain Python list, so the indexing needs to index the children of the GridPlot not a plain list.
- Color values need to be translated.

Both should have been caught in the original bokeh backend implementation (and I'm sure the color translation was there at some point, but probably got lost in one of the many rebases I went through). I've added just one test that checks both problems.

closes #1975

PR #1988, opened 2024-03-18
Seperate plot classes

DougBurke:
WIP

Requires #1984

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -347,7 +347,7 @@ class SourcePlot(shplot.HistogramPlot):
                 post += pterm
 
         scale = (self.xhi + self.xlo) / 2
-        for ii in range(data.plot_fac):
+        for _ in range(data.plot_fac):
             self.y *= scale
 
         sqr = shplot.backend.get_latex_for_string('^2')
@@ -382,7 +382,7 @@ class ComponentModelPlot(shplot.ComponentSourcePlot, ModelHistogram):
 
     def prepare(self, data, model, stat=None):
         ModelHistogram.prepare(self, data, model, stat)
-        self.title = 'Model component: %s' % model.name
+        self.title = f'Model component: {model.name}'
 
     def _merge_settings(self, kwargs):
         return {**self.histo_prefs, **kwargs}
@@ -404,7 +404,7 @@ class ComponentSourcePlot(shplot.ComponentSourcePlot, SourcePlot):
 
     def prepare(self, data, model, stat=None):
         SourcePlot.prepare(self, data, model)
-        self.title = 'Source model component: %s' % model.name
+        self.title = f'Source model component: {model.name}'
 
     def _merge_settings(self, kwargs):
         return {**self.histo_prefs, **kwargs}
@@ -638,7 +638,7 @@ class BkgResidPlot(shplot.ResidPlot):
 
     def prepare(self, data, model, stat):
         super().prepare(data, model, stat)
-        self.title = 'Residuals of %s - Bkg Model' % data.name
+        self.title = f'Residuals of {data.name} - Bkg Model'
 
 
 class BkgRatioPlot(shplot.RatioPlot):
@@ -646,7 +646,7 @@ class BkgRatioPlot(shplot.RatioPlot):
 
     def prepare(self, data, model, stat):
         super().prepare(data, model, stat)
-        self.title = 'Ratio of %s : Bkg Model' % data.name
+        self.title = f'Ratio of {data.name} : Bkg Model'
 
 
 class BkgChisqrPlot(shplot.ChisqrPlot):
@@ -738,7 +738,7 @@ class OrderPlot(ModelHistogram):
                 for interval in old_filter:
                     data.notice(*interval)
 
-        self.title = 'Model Orders %s' % str(self.orders)
+        self.title = f'Model Orders {self.orders}'
 
         if len(self.xlo) != len(self.y):
             raise PlotErr("orderarrfail")
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2016, 2019, 2020, 2021, 2022, 2023
+#  Copyright (C) 2010, 2015, 2016, 2019 - 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -129,8 +129,8 @@ class DataPHAPlot(shplot.DataHistogramPlot):
         # Maybe to_plot should return the lo/hi edges as a pair
         # here.
         #
-        (_, self.y, self.yerr, self.xerr, self.xlabel,
-         self.ylabel) = data.to_plot()
+        plot = data.to_plot()
+        (_, self.y, self.yerr, _, self.xlabel, self.ylabel) = plot
 
         if stat is not None:
             yerrorbars = self.histo_prefs.get('yerrorbars', True)
@@ -704,7 +704,7 @@ class OrderPlot(ModelHistogram):
             self.xlo = []
             self.xhi = []
             self.y = []
-            (xlo, y, yerr, xerr,
+            (xlo, y, yerr, _,
              self.xlabel, self.ylabel) = data.to_plot(model)
             y = y[1]
             if data.units != 'channel':
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -604,7 +604,7 @@ class BkgModelPHAHistogram(ModelPHAHistogram):
     """
 
     def __init__(self):
-        ModelPHAHistogram.__init__(self)
+        super().__init__()
         self.title = 'Background Model Contribution'
 
 
@@ -619,7 +619,7 @@ class BkgModelHistogram(ModelHistogram):
     """
 
     def __init__(self):
-        ModelPHAHistogram.__init__(self)
+        super().__init__()
         self.title = 'Background Model Contribution'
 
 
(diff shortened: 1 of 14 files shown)


DougBurke:
Rebased as #1984 has been merged.

DougBurke:
rebased to pick up changes in #1976

sherpa/astro/plot.py
```diff
@@ -576,6 +585,7 @@ class RMFPlot(shplot.HistogramPlot):
     # TODO: Make that a plot preference
     # How many monochromatic lines to use
     n_lines = 5
+    "The numner of lines to draw (only used when `energies` is `None`)."
```
DougBurke, line 588:
```suggestion
    "The number of lines to draw (only used when `energies` is `None`)."
```

sherpa/astro/plot.py
```diff
@@ -643,10 +661,27 @@ def prepare(self,
         elo, ehi = rmf.energ_lo, rmf.energ_hi
         l1 = np.log10(elo[0])
         l2 = np.log10(ehi[-1])
-        dl = (l2 - l1) / (self.n_lines + 1)
 
-        lines = l1 + dl * np.arange(1, self.n_lines + 1)
-        energies = np.power(10, lines)
+        # NOTE: we do not actually record the chosen energies
```
DougBurke, line 665:
```suggestion
# NOTE
# We do not actually record the chosen energies other than in the labels field,
# which is a lossy transform.
#
```

but with spacing

sherpa/plot/__init__.py
```diff
@@ -818,6 +812,11 @@ def prepare(self, data, stat=None):
 
         self.title = data.name
 
+    # This extends the superclass by also plotting yerr but the
+    # current design means we need to re-write this routine rather
+    # than identifying this programatically (over moving yerr support
+    # to the superclass).
```
DougBurke, line 1051: This comment is not as clear as it could be.

sherpa/astro/plot.py
```diff
@@ -480,6 +517,13 @@ class RMFPlot(shplot.HistogramPlot):
     y  = None
     "array_like: The response for a specific energy or channel."
 
+    energies = None
+    """The energies at which to draw the response (in keV).
+
+    If set  to None then `n_lines`  energies will be selected  to span
```
hamogu, line 523:
```suggestion
    If set  to None then `n_lines` energies will be selected  to span
```

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -326,8 +330,65 @@ def test_get_arf_plot(idval, clean_astro_ui):
...
+    assert ap.xlo == pytest.approx(_energies_lo)
+    assert ap.xhi == pytest.approx(_energies_hi)
+
+    assert ap.y == pytest.approx(_arf)
+
+    assert ap.title == 'test-arf'
+    assert ap.xlabel == 'Energy (keV)'
+
+    # the y label depends on the backend (due to LaTeX)
+    # assert ap.ylabel == 'cm$^2$'
```
hamogu, line 350: Not complete, but better than skipping.
```suggestion
    assert 'cm' in ap.ylabel
    assert '2' in ap.ylabel
```

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -326,8 +330,65 @@ def test_get_arf_plot(idval, clean_astro_ui):
...
+    assert ap.xlo == pytest.approx(_wavelength_lo)
+    assert ap.xhi == pytest.approx(_wavelength_hi)
+
+    assert ap.y == pytest.approx(_arf)
+
+    assert ap.title == 'test-arf'
+    assert ap.xlabel == 'Wavelength (Angstrom)'
+
+    # the y label depends on the backend (due to LaTeX)
+    # assert ap.ylabel == 'cm$^2$'
```
hamogu, line 377: same.

hamogu (changes requested):
@DougBurke  wrote

   - the histogram class already had such a class, but not the plot and contour classes, so I used a common name (i.e. prepend Base) and then the histogram case can just subclass the BaseHistogram case
   - an alternative design would be to make the BaseXXX classes "top-level" (so, in Sherpa speak, inherit from NoNewAttributesAfterInit) and then have them include a XXX class (i.e. use a has-a rather than is-a relationship), but that could be done at a later time.

for me the motivation for this change of putting another layer into the hierarchy is not quite clear. Is that mostly in preparation for #1989? If so, a few lines of docstring in the new classes that explains what they are meant to be used for might be useful.

DougBurke:
So, the existing code has `FooXXX` which derives from `XXX` (where `XXX` is `Plot`, `Histogram`, `Contour`). However, they are different: the derived classes generally 

- require a `prepare` call before the `plot` (or `contour`) call
- the `plot`/`contour` call has a different form since it does not need to be sent in the data to display, as it can just grab the data created by the `prepare` call

Ideally I think they are separate hierarchies: that is we have `Plot`, `Histogram`, and `Contour` and what I have called `BasePlot`, `BaseHIstogram`, and `BaseContour` but technically it could be the existing classes [*]. So we would not have `DataPlot` derive from `Plot`.

[*] However, the existing classes are not set up in a way to say "you need a `prepare` call`. For example, we have

- `DataPlot` deriving from `Plot`: https://github.com/sherpa/sherpa/blob/52440b6806d28bf9a6c1f222c492527a670a19af/sherpa/plot/__init__.py#L1310
- `CDFPlot` deriving from `Plot`: https://github.com/sherpa/sherpa/blob/52440b6806d28bf9a6c1f222c492527a670a19af/sherpa/plot/__init__.py#L922
 
so they end up requiring the same logic but there's no "class hierarchy" to tell the reader this - they just have to look at each class.

So I wanted to add a class that said "hey, you need to call `prepare` and then the `plot`/`contour` call takes no data" logic. I disagree with the suggestion that it adds no value, but perhaps I am too close to this code. We actually already have this with the `Histogram` class - i.e. all the histogram classes derive from a class that defines a `prepare` method -  but not for the `DataXXX` and `ContourXXX` classes (e.g. `DataPlot` vs `CDFPlot` from above). Thanks to the existing naming scheme, not wanting to break forwards compatibility, and the desire to have a common name scheme I came up with the `BaseXXX` naming scheme, but we could call it something different.

I will note that I thought I had actually described this for each `BaseXXX` class with the docstring

    Base class for xxx with a prepare method.

It is true that I plan (ie already have, but am currently reworking it) changes that build off these classes in #1989

Would it make sense to break the hiearchy, so that `Plot`/`Contour`/`Histogram`/`Point` are all "singleton" classes [that is, nothing really derives from them] and then we can have a separate set of hierarchies that are the equivalent of my `BaseXXX` classes here?

DougBurke:
I've created #2003 which is the first three commits from this PR. They are hopefully less in need of a discussion and can be taken as is, which will slightly simplify this PR and other planned PRs.

hamogu:
Yeah, I started to type comments for this PR, but then I found that I typed so much code into the comment boxes that it was easier to actually edit the code and see where it takes me. It's not quite finished over the weekend, so I didn't push yet, but will do so tomorrow at the latest so that we can compare my suggestion and your approach here and then decide how to move forward.

hamogu:
I'll rebase on op top of #2003 for easier comparison.

hamogu:
I've tried to implement what I had in mind to look at as an alternative, but it does not work out in code. That means that either my idea is too complicated or I've just not gotten it right yet. Either way, it turns out I would do it on top of this PR, not instead of this PR, so this can move forward. I've clicked "resolve" on all comments related to class names and class structure, which leaves just four small suggestions (a typo, two string tests and a loop that I believe can be replaced with `a**b`.

DougBurke:
I've rebased this because #2003 and #2004 have been merged but I need to update the commit message and review the commentary from @hamogu

DougBurke:
This is going to be a pain to merge. I've stuck in my WIP changes so I don't lose them.

DougBurke:
Rebased to pick up recent changes - in particular thanks

- #2011

and to add some changes suggested by @hamogu. I don't know if the documentation around the `BaseXXX` classes is sufficient - in particular, thanks to #2011 we now have

```mermaid
classDiagram
Plot --|> BasePlot
Histogram --|> BaseHistogram
Contour --|> BaseContour
BasePlot --|> ModelPlot
BaseHistogram --|> HistogramPlot
HistogramPlot --|> ModelHistogramPlot : a historical wart
BaseContour --|> ModelContour
ModelPlot --|> BaseResidualPlot
ModelHistogramPlot --|> BaseResidualHIstogramPlot
ModelContour --|> BaseResidualContour
```

So maybe we want to rename the `BasePlot/xxx` classes to something else (but I quite like `Base` here to indicate this is something we build on but you don't want to create an instance of the class in question, so using it at different parts of the hierarchy may not be too bad).

The follow up to this (currently #2076 but that needs a lot of work) reworks this hierarchy slightly, mainly by breaking the initial `XXX -> BaseXXX` chain (so `XXX` becomes a separate thing with a different name and `BaseXXX` then becomes the "actual" top-level class). Or at least that's the plan; it may not happen.

DougBurke:
@hamogu - it's been long enough ago that I no-longer really remember the state of this. I've rebased it because of recent changes and had a quick look at it and it seems fine (assuming the "adding an explicit class" idea is meaningful).

hamogu:
I admit to having no memory of this, but I will re-review (likely over the weekend).

DougBurke:
Ta. No hurry.

DougBurke:
So, I've just "finished" the follow-on PR to this

- #2076 

and I don't think we are going to get both of these in for 4.17.0, and I think we'd want both in, thanks to the changes.

So I don't think it's worth spending time reviewing this right now. I do plan to extract some of the "dry-by" improvements from both PRs that we can hopefully get in for 4.17.0.

DougBurke:
Rebased to pick up recent merges like #2081

Issue #1993, opened 2024-04-04
cycle checking in linked parameters is incomplete


DougBurke:
This is related to pre-GitHub issue 12287

There is basic cycle checking when linking parameters, so that

```
sherpa In [1]: gauss2d.mdl
-------------> gauss2d.mdl()
Out[1]: <Gauss2D model instance 'gauss2d.mdl'>

sherpa In [2]: mdl.ypos = mdl.xpos

sherpa In [3]: mdl.xpos.val
Out[3]: 0.0

sherpa In [4]: mdl.ypos.val
Out[4]: 0.0

sherpa In [5]: mdl.xpos = 2000

sherpa In [6]: mdl.ypos.val
Out[6]: 2000.0

sherpa In [7]: print(mdl.xpos)
val         = 2000.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
default_val = 2000.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38

sherpa In [8]: print(mdl.ypos)
val         = 2000.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = True
link        = mdl.xpos
default_val = 2000.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38

sherpa In [9]: mdl.xpos = mdl.ypos

sherpa In [10]: mdl.xpos.val
Out[10]: 0.0

sherpa In [11]: mdl.ypos.val
Out[11]: 0.0

sherpa In [12]: print(mdl.xpos)
val         = 0.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = True
link        = mdl.ypos
default_val = 0.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38

sherpa In [13]: print(mdl.ypos)
val         = 0.0
min         = -3.4028234663852886e+38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
default_val = 0.0
default_min = -3.4028234663852886e+38
default_max = 3.4028234663852886e+38
```

So, note that when we set `mdl.xpos = mdl.ypos` it automatically cleared the `mdl.ypos` link. This was the issue for #12287. However, as shown below, this does not catch all cases, and it is easy to get into an infinite loop.

```
sherpa In [1]: gauss2d.mdl
[...]

Issue #1992, opened 2024-04-04
parameter values can be set to None


DougBurke:
This is with CIAO 41.6 but it's been like this essentially for ever

```
sherpa In [1]: gauss1d.mdl
-------------> gauss1d.mdl()
Out[1]: <Gauss1D model instance 'gauss1d.mdl'>

sherpa In [2]: print(mdl.fwhm)
val         = 10.0
min         = 1.1754943508222875e-38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
default_val = 10.0
default_min = 1.1754943508222875e-38
default_max = 3.4028234663852886e+38

sherpa In [3]: mdl.fwhm = None

sherpa In [4]: print(mdl.fwhm)
val         = nan
min         = 1.1754943508222875e-38
max         = 3.4028234663852886e+38
units       = 
frozen      = False
link        = None
default_val = nan
default_min = 1.1754943508222875e-38
default_max = 3.4028234663852886e+38

sherpa In [5]: mdl.fwhm.val
Out[5]: nan
```

There is a similar issue with the `default_val` field.

We do at least check we can not set a value to a non-numeric value:

```
sherpa In [11]: mdl.fwhm = "a string"
ValueError: could not convert string to float: 'a string'
```

We could stop this, but is it worth it? Some may claim that it's a valid thing to do, but I can't think of any real reason off the top of my head.

We could also stop parameters from being set to `nan`/`inf`/... but again is it worth it?

hamogu:
I don't think we have to do anything about it and there might be model (could be user models) where it makes sense to have some parameters set top inf. (Nan might be another matter). It's not something that happens accidentally (you have to type `= np.inf` if you want that and it's pretty easy to spot when looking at the model parameters. 
So, to me that's not a bug and I would suggest to close this issue.

hamogu:
Similarly, the name of parameter values can be changed:
```
mdl.pos.name = 'position'
```
When I do that, the print output stays the same:
```
print(mdl)
gauss1d.mdl
   Param        Type          Value          Min          Max      Units
   -----        ----          -----          ---          ---      -----
   mdl.fwhm     thawed           10  1.17549e-38  3.40282e+38           
   mdl.pos      thawed            0 -3.40282e+38  3.40282e+38           
   mdl.ampl     thawed            1 -3.40282e+38  3.40282e+38   
```
but the notebook representation is now different:
[...]

PR #1994, opened 2024-04-04
Tests: improve coverage of parameter module

DougBurke:
# Summary

Improve the test coverage of the parameter module.

# Details

When working on some recent changes to the model and parameter code, such as #1972 and #1974, I noticed some code paths that were not being checked. So this attempts to add coverage for these lines (although it does not guarantee to do so since there's a cost-benefit trade off here).

Writing these tests lead to noting issues

- #1992
- #1993

but these are not addressed here (but any changes to these issues is likely to be caught by these tests, which was the point in writing them!).

### DIFF
--- a/sherpa/models/tests/test_parameter.py
+++ b/sherpa/models/tests/test_parameter.py
@@ -730,7 +730,7 @@ def test_set_default_to_parameter(check_str):
 def test_set_default_to_parameter_to_parameter(check_str):
     """We support this, so make sure it is tested.
 
-    Maybe I shouldn't have watched Inception. There is the question of
+   There is the question of
     whether it makes sense to support this, but that's a different
     issue.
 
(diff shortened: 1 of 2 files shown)


hamogu (approved):

sherpa/models/tests/test_parameter.py
```diff
@@ -605,3 +622,341 @@ def test_explicit_numpy_combination():
...
+                "default_val = 14.0",
+                "default_min = -3.4028234663852886e+38",
+                "default_max = 3.4028234663852886e+38"
+               ])
+
+
+def test_set_default_to_parameter_to_parameter(check_str):
+    """We support this, so make sure it is tested.
+
+    Maybe I shouldn't have watched Inception. There is the question of
```
wmclaugh, line 733:
```suggestion
   There is the question of
```

PR #2000, opened 2024-04-10
CM-481: update codecov action to v4, add upload token

harlantc:
Codecov seem to want us to start using v4, and that requires the codecov upload token. 

I added the upload token to the CI action secrets.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -204,7 +204,7 @@ jobs:
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
         conda install -yq pytest-cov
-        python setup.py -q test -a "--cov sherpa --cov-report xml"
+        python setup.py -q test -a "--cov sherpa --cov-report xml:${{ github.workspace }}/coverage.xml"
 
     - name: sherpa_test Tests
       if: matrix.test-data == 'package' || matrix.test-data == 'none'
@@ -213,11 +213,12 @@ jobs:
         conda activate build
         conda install -yq pytest-cov
         cd $HOME
-        sherpa_test --cov sherpa --cov-report xml
+        sherpa_test --cov sherpa --cov-report xml:${{ github.workspace }}/coverage.xml
 
     - name: upload coverage
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       with:
-        # only select the reports that we know are coverage reports, from linux conda, mac conda, and pip
-        files: /home/runner/work/sherpa/sherpa/coverage.xml,/Users/runner/work/sherpa/sherpa/coverage.xml,/home/runner/coverage.xml
+        token: ${{ secrets.CODECOV_TOKEN }}
+        files: ${{ github.workspace }}/coverage.xml
         verbose: true
+      
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -2,7 +2,7 @@ codecov:
   require_ci_to_pass: yes
   #this handles notifications for bots (slack, gitter, etc)
   notify:    
-    after_n_builds: 8 #wait for 8 of the builds to finish and send their reports to send a notice, default is 2
+    after_n_builds: 13 #wait for 13 of the builds to finish and send their reports to send a notice, default is 2
     wait_for_ci: yes
   
 coverage:
@@ -31,7 +31,7 @@ comment:
   layout: "reach,diff,flags,files,footer"
   behavior: default
   require_changes: no
-  after_n_builds: 8 # wait for 8 builds to finish and send their reports to make a comment, default 2
+  after_n_builds: 13 # wait for 13 builds to finish and send their reports to make a comment, default 2
   
 # Fix issue with 'work/sherpa/sherpa' being considered a seperate directory 
 fixes:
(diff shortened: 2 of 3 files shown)


DougBurke:
Hmm, looking at the "Coverage upload" step of https://github.com/sherpa/sherpa/actions/runs/8631474927/job/23659963268?pr=2000 shows some interesting things

- it claims to switch to a tokenless upload since this is from a fork; does this matter?
- it can't find the files to upload, which is strange; has the action changed things?

harlantc:
>     * it claims to switch to a tokenless upload since this is from a fork; does this matter?

Looks like that's a feature of v4, https://docs.codecov.com/docs/codecov-uploader#supporting-tokenless-uploads. Once it's merged, we'll get token'd uploads as well. I'd not be surprised if they changed it at some point though.

>     * it can't find the files to upload, which is strange; has the action changed things?

Some reports were always missing AFAIK, but it looks like there are less now than there were before. doing a quick glance, 13 seems to be the magic number, but so far we've only gotten 8 (I suspect 9 with the macSO one that'll eventually finish). I'll try and see where the other 4 went.

DougBurke:
Well, we now have the codecov report back, which is the main thing ;-)

harlantc:
so it looks like the jobs:
https://github.com/sherpa/sherpa/actions/runs/8631474927/job/23659961633?pr=2000
https://github.com/sherpa/sherpa/actions/runs/8631474927/job/23659961916?pr=2000
https://github.com/sherpa/sherpa/actions/runs/8631474927/job/23659963268?pr=2000
https://github.com/sherpa/sherpa/actions/runs/8631474927/job/23659963533?pr=2000
failed to create reports for some reason, but for those same jobs on other commits they did.
For some reason some of the reports are getting places where we expect. I've updated it to just put them in the home dir, with names unique to the sherpa_test and pytest calls, cuz even if it doesn't look like they both run on the same job, it makes it clear which way the report was generated.

harlantc:
Now I'm remembering why I didn't add tokens before, codecov just assumes cuz the PR is from a fork it can't get the secret, even though those are allowed by collaborators. 

so either we accept all our codecovs uploads from fork PRs runs might fail from timeouts, or we just put the token in a plaintext variable. Codecov docs have examples just putting the token in plaintext, which doesn't make sense. Will look into options

harlantc:
I still have the reports at 8, but can bump it to 12, or 13 even as seems to be all the reports generated, but if a report fails to upload it will be annoying.

Also, Codecov will also just bomb out randomly if they get rate limited, and the only workarounds are either 
1) PRs are created from this sherpa project's branches, not forks (suggest against this)
2) we make the codecov token public (also suggest against this, but codecov seems to not care enough to tell us not to do it, which is strange)

I don't like either of those solutions. I'll leave it up to others to decide what they'd like best. 

All reports? some reports? 

change all of sherpa's workflow? make the token public?

.github/workflows/ci-conda-workflow.yml
```diff
@@ -204,7 +204,6 @@ jobs:
         source ${CONDA}/etc/profile.d/conda.sh
         conda activate build
         conda install -yq pytest-cov
-        cd $HOME
```
DougBurke: I'm concerned about removing this, since I think (it's hard to tell) that this is meant to test the installed version.

Of course, it would be easier if we had #1608 merged as I have not looked at the "setup.py test" code for a long while as it's deprecated.

harlantc: Actually, I had added `cd $HOME`, and then removed it after it broke (because that test runs the setup.py, which is located in the work dir, not HOME). 

If you take a look at the final diff, not the diff between commits, you'll see where I added `cd $HOME`.

background: trying to debug why the tests get different results, and it looks like because the tests are run from different locations, causing different coverage reports. it would seem some are running `.../work/sherpa/sherpa/` and others are working from `.../work/sherpa/sherpa/sherpa/` (it breaks down to .../work/sherpa<org name>/sherpa<repo name>/sherpa<the normal dir in the repo>/`). I hadn't rebased/squashed while doing it, so you're seeing those diffs.

now that I've fixed that, the coverage reports something closer to expected, 83.11%.

harlantc:
> a) it will get us back to having codecov reports on our PRs 
> b) except that we may randomly get rate-limited by codecov for a period of time 
Yes, that seems to be a limitation on codecov/github. codecov trying to force everyone they can to use tokens is their attempt to reduce github ratelimits. they make an exception for "tokenless" uploads for fork PRs, but it's still an issue it seems. 

> c) I note that this particular PR claims to drop overall coverage from 82.2% to 81.0% and it is really not-at-all obvious why this is (the codecov report just doesn't make sense here, and this is not the first time this has happened so it's not because of this PR)
I've addressed this, should be working as expected now.

> There are things we could do to address point b, but this is not done in this PR (in part because all the options are "sub-optimal").
yeah, exposing our codecov token would let anyone to upload bogus reports for sherpa. I have one other option to explore, but I don't think it's implemented on Codecov/github's side. Other option is everyone starts working on branches in sherpa directly and not on forks.

DougBurke (approved):
Thanks for answering my questions. Hopefully this is a bit more robust now with your work (modulo rate limits) ;-)

harlantc:
I took a closer look at C, and because I updated where we grab the reports from, it might be the case we're grabbing different reports (which confuses me, I thought we would only have one) but after some cross-checking, it looks like there might be multiple reports generated on some runs, and I've changed which ones we grab

harlantc:
Sorry, hit the wrong button trying to cancel a comment

PR #2004, opened 2024-04-22
Minor improvements to the RMF plot capability

DougBurke:
# Summary

Improve the RMF plot display so that it recognizes the current units setting and allows the choice of energies to be over-ridden by the user.

# Details

This is taken from #1988 and just focuses on the RMF changes, not the larger architectural changes in that PR. It depends on #2003.

With this we have

```
>>> load_pha("sherpa-test-data/sherpatest/3c273.pi")
>>> plot_rmf()
```

sherpa

This matches the existing behaviour. 

However we can now switch to Wavelength space:

```
>>> set_analysis("wave")
dataset 1: 0.829303:8492.07 Wavelength (Angstrom)
>>> plot_rmf()
```

sherpa

Or select the energies to use:

```
>>> get_rmf_plot().energies = [0.2, 0.4, 0.8, 2, 4, 5, 7]
>>> plot_rmf()
```

sherpa

Or the number of lines to use (need to clear the `energies` setting for this to work):

```
>>> get_rmf_plot().energies = None
>>> get_rmf_plot().n_lines = 8
>>> set_analysis("energy")
dataset 1: 0.00146:14.9504 Energy (keV)
>>> plot_rmf()
```

sherpa

## Things to note:

- you can only specify the energies to draw in keV (we could have it instead chose the units based on the current setting, if we renamed the field from `energies`), but is it worth it
- selecting the `energies` or `n_lines` field to set / change does not match the other plot "options", but that's because these are needed in the `prepare` stage and the preferences ar eonly used in the `plot` stage [see issue #889 for some ruminations on this]

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -27,7 +27,8 @@ import logging
 import numpy as np
 
 from sherpa.astro import hc
-from sherpa.astro.data import DataPHA
+from sherpa.astro.data import DataARF, DataPHA, DataRMF
+from sherpa.astro.instrument import ARF1D, RMF1D
 from sherpa.astro.utils import bounds_check
 from sherpa.models.basic import Delta1D
 from sherpa import plot as shplot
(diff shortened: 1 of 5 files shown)


DougBurke:
@hamogu - I know you've "approved" if #1988 but I think it's worth separating the functionality here so we can think a bit more about the lower-level changes, since they do not really interact much with the actual user-facing changes.

hamogu (approved):

PR #1958, opened 2024-02-06
CI: switch to macos-14 for the MacOS build

DougBurke:
Following https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

PR #2009, opened 2024-04-26
Allow both Intel and ARM macOS GitHub runners for the conda builds

DougBurke:
The changes in #2195 invalidate this. Rather than change it I've created #2198 just so that we can see the difference that the `setup-miniconda` action makes. I've marked this as a draft.

# Summary

Switch the GitHub conda workflow to always install miniconda, add a macOS ARM build, and update the macOS Intel build to use the macOS-13 runner since the macOS-12 runner is about to be removed.

# Details

This is a rework of #1958.

There's three parts

a) switching everything to use miniforge by using the `setup-miniconda` action
b) a minor clean up of the workflow actions by splitting up the DS9 and XSPEC installation steps as they are technically separate (even if we only ever have both)
c) add a macOS ARM build

As part of this we have

- select the CIAO `conda` channel for macOS ARM since there's no ARM build of `xspec-modelsonly` in the `xspec` channel
- use version `12.13.1` rather than `12.14.0` of XSPEC for macOS ARM to match the CIAO `conda` channel
- an oddity in that the `default` channel seems to get added when adding the XSPEC channel (for both Linux and macOS) and I have no idea why but I just ensure it gets removed when added
- tweaked the macOS conda install to use Python 3.11 and install bokeh

I have **not** checked to see how this interacts with 

- #2156

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -1,7 +1,17 @@
 #!/usr/bin/env bash -e
 
+# Occasionally useful to know what these values are
+echo "** uname -s: `uname -s`"
+echo "** uname -m: `uname -m`"
+
 if [ "`uname -s`" == "Darwin" ] ; then
-    compilers="clang_osx-64 clangxx_osx-64 gfortran_osx-64"
+    if [ "`uname -m`" == "x86_64" ]; then
+	sys="64"
+    else
+	sys="arm64"
+    fi
+
+    compilers="clang_osx-${sys} clangxx_osx-${sys} gfortran_osx-${sys}"
 
     #Download the macOS 11.0 SDK to the CONDA_BUILD_SYSROOT location for the Conda Compilers to work
     mkdir -p ${GITHUB_WORKSPACE}/11.0SDK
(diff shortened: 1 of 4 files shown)


hamogu:
All I do special is setting `export PYTHON_LDFLAGS=' '` as documented in https://sherpa.readthedocs.io/en/4.16.0/install.html#building-from-source

DougBurke:
Ta @hamogu - our CI actions already do this - and the error actually comes from building the `extern` code so I don't think it's even hit that part of the build yet...

DougBurke:
Rebased as #2135 has been merged which picked up the tolerance changes for the tests.

DougBurke:
From GitHub:

The macOS 12 runner image will be removed by December 3rd, 2024. To raise awareness of the upcoming removal, jobs using macOS 12 will temporarily fail during scheduled time periods defined below:

    November 4, 9:00 AM - 7:00 PM EST
    November 11, 9:00 AM - 7:00 PM EST
    November 18, 9:00 AM - 7:00 PM EST
    November 25, 9:00 AM - 7:00 PM EST

DougBurke:
This has been rebased to pick up the latest main branch (although not much has changed since the last rebase) but also the commits have been cleaned up and re-ordered. The result is the same (apart from we now keep the macOS Intel run but using `macos-13` rather than `macos-12`).

DougBurke:
I've just seen tests fail because they are implementing one of the "temporarily remove the macos-12 runner" periods.

DougBurke:
As #2198 has landed I'm closing this (if we want to move to the conda action we'd probably need to redo this anyway, as things would have changed but on our end and on the action end).

PR #2010, opened 2024-04-26
CI: drop the macOS run (temporary)

DougBurke:
# Summary

Temporarily drop the macOS GitHub action. There is no change to Sherpa.

# Details

The macOS CI runner on GitHub has switched to an ARM build when previously it was Intel. This requires updating our build instructions and working through issues, so the short-term fix is to drop the CI macOS run.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -23,14 +23,14 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


hamogu (approved):

wmclaugh:
Unless there are objections, would prefer to pin macos back to version 12 to keep the CI runs for the time being (via #2012).

DougBurke:
No problem - I'll close this.

Issue #2007, opened 2024-04-24
can not use spawn or forkserver methods for multiprocessing


DougBurke:
[**EDITED TO ADD** I do not know why GitHub thinks #2243 will fix this as it will not]

This is fallout from looking into #1015 but is not related to it (that is, it is not an issue for #2006, or - rather - that PR doesn't change the existing behaviour).

1. we explicitly do not support the `forkserver` method (an easy fix)
2. using `spawn` (and probably `forkserver` but I haven't checked) we get errors because 
    - we need to handle callbacks differently since we can't pickle the local callback routines that we use with the `fork` method
    - likely the same issue, but we need to take more care about sharing information between the process and the parent

As an example, with CIAO 4.16:

# the default fork method

This errors out as expected

```
>>> from sherpa.utils import parallel
>>> parallel.multiprocessing_start_method
'fork'
>>> def func(x):
...     if x == 2:
...         raise ValueError("x")
...     return x ** 2
... 
>>> parallel.parallel_map(func, [1, 2, 3])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/soft/ciao-4.16/lib/python3.11/site-packages/sherpa/utils/parallel.py", line 397, in parallel_map
    return run_tasks(procs, err_q, out_q)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/soft/ciao-4.16/lib/python3.11/site-packages/sherpa/utils/parallel.py", line 264, in run_tasks
    process_tasks(procs, err_q)
  File "/soft/ciao-4.16/lib/python3.11/site-packages/sherpa/utils/parallel.py", line 230, in process_tasks
    raise err_q.get()
ValueError: x
```

# with `spawn`

```
>>> from sherpa.utils import parallel
>>> parallel.multiprocessing_start_method
'spawn'
>>> def func(x):
...     if x == 2:
...         raise ValueError("x")
...     return x ** 2
... 
>>> parallel.parallel_map(func, [1, 2, 3])
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/soft/ciao-4.16/lib/python3.11/multiprocessing/spawn.py", line 122, in spawn_main
    exitcode = _main(fd, parent_sentinel)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/soft/ciao-4.16/lib/python3.11/multiprocessing/spawn.py", line 132, in _main
[...]

PR #2015, opened 2024-04-30
Remove unused code in sherpa.utils and sherpa.optmethods

DougBurke:
# Summary

Removal of unused code in `sherpa.utils` and `sherpa.optmethods`.

# Details

In looking into the `multiprocessing` code (related to #2007) I've realized there's a bunch of code that we either never exercise, or only run as part of a test. So,as this code has been lying around for a while we can remove it. This code is primarily in `sherpa.utils`. 

I do note that this doesn't address all unused code in `sherpa.utils` as I would rather be strategic in what we remove, and this PR is focused on optimization/root-finding code which we have ended up not using.

There are four actual code changes

 1. replace a hand-rolled version with `numpy.sign`; it doesn't save us much but why repeat the wheel
 2. replaced a decorator to count the number of function evaluations by a class, which lead to an internal rework of `sherpa.optmethods.Opt` to remove the `func_counter_bounds_wrappers` method since it didn't add any value to doing the steps directly
 3. removed some commented-out lines and changing things like `x, y, ... = foo(..); return x, y, ...` to `return foo(...)`
 4. the `sherpa.utils.is_in` is replaced by `foo in bar` in `sherpa.fit`, which is the only place it's used (this also pointed out to ne that `sherpa.astro.utils.is_in` exists and has a different API, so removing one of them helps lead to less confusing code  

The second one is a larger change as it changes the interface - e.g. rather than return a mutable list and a wrapped function we now just return a single object which does both - and this has some knock-on consequences in `sherpa.estmethods` and `sherpa.estmethods`, but I claim that the code is a little-bit clearer now. From some simple testing I did it doesn't seem to make any significant runtime change (but that is going to depend on the python version). I actually want this change as part of addressing #2007 but I think it is useful enough to pull out separately.

This code cause down-stream problems - i.e. if code actually uses any of these functions - but they can just grab the code and use it directly if necessary. I note that only one of the symbols that has been removed from `sherpa.utils.__all__`, so they were not widely advertised (i.e. internal only).

The coverage checks do point out that some of the less-well-used paths in the optimization code is not well tested.


hamogu:
I'm excited about this PR because, even if the code were used, it should absolutely be removed because it uses way too many inner loops in Python. Faster (i.e. compiled) versions of similar codes are widely available.

DougBurke:
Rebased to address conflict.

anetasie (commented):
I looked at the code changes. They look fine. We should remove the code which is not used. My only question is if the Hessian was required in calculating the covariance matrix anywhere and perhaps we switched to output from `levmar` for a quick printouts of errors.

DougBurke:
@anetasie - to be honest I haven't been that bothered by why we no-longer use the code! There's a lot of code from Dan that was obviously used during development, testing, and in his possible future plans that we don't actually use. 

If you go to the first commit

https://github.com/sherpa/sherpa/commit/9ccf0e72acb9f85fdd20a8feea16ea774e1288b3#diff-d29a8f76d3d86e53418a2992eaebb8a1f0dae7953c027ed6baf1a4e3361bad1a

then the hessian routine or Hessian class isn't actually used (as far as I can tell), and similarly the NumDeriv classes also don't appear to be used. So they've been long dormant.

anetasie:
@DougBurke I agree.

anetasie (approved):

PR #2017, opened 2024-04-30
IO: separate pack and set commands

DougBurke:
# Summary

Replace the logic of the packup argument to the set I/O commands by adding new `pack_xxx_data` commands (which return an object, and so correspond to the old `packup=True` case) and make the existing `set_xxx_data` calls never return anything (the old `packup=False` behaviour). As part of this, ensure that the dummy backend provides arguments and - for most arguments - types, which means the dummy backend can be thought of as defining the backend interface. There should be little change to users who use the sherpa.astro.io module routines, although some error conditions now raise an IOErr rather than a TypeError or ValueError.

# Details

This is #1929 but without the change to "make the backend use a module" approach that I added during development of #1929. I thought it better to create a new PR so we don't lose the old version (for reference) and since I can rebase things to avoid conflicts. There have also been changes to the code - e.g. some of the code went into #2061 instead.

The PR starts by reworking the dummy backend so that it "defines" the expected API, in terms of what routines need to be provided and what the input and outputs are, along with documentation. I note that the two existing backends do not always agree on the default values for some of these arguments - generally the `make_copy` and `fix_type` arguments so this is a "best guess" and we can iterate in later PRs to address small differences (at present I'm tempted to say that "some arguments are not guaranteed to be used in a backend", but I think until we get a new backend I don't think it's worth spending much time on this).

There is then a commit to fix up some recent type changes - probably from #2061 - just to try and be consistent, and then a "change numpy to np" change for an internal-ish module.

The "selling point" for this PR actually occurs in the fourth commit: ` IO: separate pack and set commands `. This changes the backend code to add `pack_xxx_data` routines and to drop the `packup` argument from the `set_xxx_data` calls. The idea is that the `set_xxx_data` call can just call `pack_xxx_data` and then write out the response but it's not always done like this (since we can go via the `set_table_data` call, for example). However, the individual pack routines are more used in the follow-up PR #1921.

We then add types to the actual backends to try and match the dummy backend.

There is then the start of a bunch of 'cleanup' commits where I try to update the code to

a) add useful types
b) use more-modern python idioms
c) clean up the code a bit
  - there's a lot of code moved around here and with minor tweaks to work better with the python typing programs, but functionally not much actually changes
d) address some changes seen in numpy since this code was first written means that sometimes, to avoid having version-specific code, I have changed how the code creates things (normally associated with converting between a 2D array of numbers and a text file). In particular asarray changed behaviour in 1.24 and 1.26 when dealing with ragged arrays.

I note that I have follow-on code in #1921 which uses this updated code to move the logic from the backend into the frontend, but that really is better done as a follow-on PR (and I need to update that PR to bring it up to date with this version).

There are then two commits

- `try to standardize basic I/O error handling`

  I try to have the routines ail with an IOErr rather than something from the backend or a ValueError/TypeError when possible.

- `treat a HDUList as a context manager (pyfits)`

  This is to address a review comment from #1929 by @hamogu that never got addressed in #2061, because I did not know how to do it then. It looks like a `fits.HDUList` behaves like a `ContextManager` so I can use `contextlib.nullcontext` to wrap the return value so that it skips the close, which lets me change

```
    tbl, fname, close = _get_file_contents(arg, ....)
    try:
        ...
    finally:
        if close:
            tbl.close()
```

  with

```
    cm, fname = _get_file_contents(arg, ....)
    with cm as tbl:
        ...
```

When working on this version I stumbled upon a very-bizarre case where only the Python 3.11 full-build would fail the CI runs. I could never replicate this behaviour. I ended up breaking code up (which should make the individual commits a bit easier to review, but I know this annoys @hamogu when I end up waiting till a later commit to make the change he suggests earlier) and have eventually either squashed this issue or it was a transient bizarro error from bizarro land. The numpy.asarray changes were annoying too.

I will note that there are a number of "obvious" places in the backend I/O code where we need to better handle the PHA offset/TLMIN issue, but that is really for another PR (and we have issues about this already). For example, hopefully after #1921 we will only have to do it in the frontend rather than have to apply changes to both backends as is needed here...

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -54,7 +54,8 @@ import importlib
 import logging
 import os
 import re
-from typing import Any, Optional, Sequence, Union
+from typing import Any, Callable, Mapping, Optional, Sequence, \
+    TypeVar, Union
 
 import numpy as np
 
@@ -67,6 +68,8 @@ from sherpa.io import _check_args
 from sherpa.utils import is_subclass
 from sherpa.utils.err import ArgumentErr, DataErr, IOErr
 from sherpa.utils.numeric_types import SherpaFloat
+
+from .types import KeyType, NamesType, HdrTypeArg, HdrType, DataType
 
 config = ConfigParser()
 config.read(get_config())
@@ -99,12 +102,14 @@ for iotry in io_opt:
     except ImportError:
         pass
 else:
-    # None of the options in the rc file work, e.g. because it's an old file
-    # that does not have dummy listed
+    # None of the options in the rc file work, e.g. because it's an
+    # old file that does not have dummy listed
     import sherpa.astro.io.dummy_backend as backend
 
 warning = logging.getLogger(__name__).warning
 info = logging.getLogger(__name__).info
+
+T = TypeVar('T')
 
 
 __all__ = ('backend',
@@ -117,7 +122,7 @@ __all__ = ('backend',
 # Note: write_arrays is not included in __all__, so don't add to the
 #       See Also section.
 #
-def read_arrays(*args):
+def read_arrays(*args) -> Data:
     """Create a dataset from multiple arrays.
 
     The return value defaults to a `sherpa.data.Data1D` instance,
@@ -176,7 +181,10 @@ def read_arrays(*args):
     return dstype('', *dargs)
 
 
-def read_table(arg, ncols=2, colkeys=None, dstype=Data1D):
+def read_table(arg,
+               ncols: int = 2,
+               colkeys: Optional[NamesType] = None,
+               dstype=Data1D) -> Data:
     """Create a dataset from a tabular file.
 
     The supported file types (e.g. ASCII or FITS) depends on the
@@ -237,7 +245,11 @@ def read_table(arg, ncols=2, colkeys=None, dstype=Data1D):
 
 
 # TODO: should this be exported?
-def read_ascii(filename, ncols=2, colkeys=None, dstype=Data1D, **kwargs):
+def read_ascii(filename: str,
+               ncols: int = 2,
+               colkeys: Optional[NamesType] = None,
+               dstype=Data1D,
+               **kwargs) -> Data:
     """Create a dataset from an ASCII tabular file.
 
     Parameters
@@ -297,7 +309,9 @@ def read_ascii(filename, ncols=2, colkeys=None, dstype=Data1D, **kwargs):
     return dstype(name, *cols)
 
 
-def read_image(arg, coord='logical', dstype=DataIMG):
+def read_image(arg,
+               coord: str = 'logical',
+               dstype=DataIMG) -> Data2D:
     """Create an image dataset from a file.
 
     .. versionchanged:: 4.16.0
@@ -367,9 +381,13 @@ def read_image(arg, coord='logical', dstype=DataIMG):
     # What are the independent axes?
     #
     if issubclass(dstype, (Data2DInt, DataIMGInt)):
-        indep = [x0 - 0.5, x1 - 0.5, x0 + 0.5, x1 + 0.5]
+        data['x0lo'] = x0 - 0.5
+        data['x1lo'] = x1 - 0.5
+        data['x0hi'] = x0 + 0.5
+        data['x1hi'] = x1 + 0.5
     elif issubclass(dstype, Data2D):
-        indep = [x0, x1]
+        data['x0'] = x0
+        data['x1'] = x1
     else:
         raise ArgumentErr("bad", "dstype argument",
                           "dstype is not derived from Data2D")
@@ -383,14 +401,14 @@ def read_image(arg, coord='logical', dstype=DataIMG):
     # not behave sensibly (likely due to #1414 which was to address
     # issue #1380).
     #
-    dataset = dstype(filename, *indep, **data)
+    dataset = dstype(filename, **data)
     if isinstance(dataset, DataIMG):
         dataset.set_coord(coord)
 
     return dataset
 
 
-def read_arf(arg):
+def read_arf(arg) -> DataARF:
     """Create a DataARF object.
 
     Parameters
@@ -417,7 +435,7 @@ def read_arf(arg):
     return DataARF(filename, **data)
 
 
-def read_rmf(arg):
+def read_rmf(arg) -> DataRMF:
     """Create a DataRMF object.
 
     Parameters
@@ -444,7 +462,8 @@ def read_rmf(arg):
     return _rmf_factory(filename, data)
 
 
-def _rmf_factory(filename, data):
+def _rmf_factory(filename: str,
+                 data: Mapping[str, Any]) -> DataRMF:
     response_map = {
         'ROSAT': DataRosatRMF,
         'DEFAULT': DataRMF,
@@ -456,8 +475,12 @@ def _rmf_factory(filename, data):
     return rmf_class(filename, **data)
 
 
-def _read_ancillary(data, key, label, dname,
-                    read_func, output_once=True):
+def _read_ancillary(data: dict[str, str],
+                    key: str,
+                    label: str,
+                    dname: str,
+                    read_func: Callable[[str], T],
+                    output_once: bool = True) -> Optional[T]:
     """Read in a file if the keyword is set.
 
     Parameters
@@ -682,7 +705,7 @@ def read_pha(arg,
     return phasets
 
 
-def _pack_table(dataset):
+def _pack_table(dataset: Data) -> DataType:
     """Identify the columns in the data.
 
     This relies on the _fields attribute containing the data columns,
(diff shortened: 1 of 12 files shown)


DougBurke:
Why does the pyton-3.11 full conda build appear to be stuck at

```
platform linux -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: /home/runner/work/sherpa/sherpa
configfile: pytest.ini
plugins: xvfb-3.0.0, cov-5.0.0
collected 10120 items
sherpa/astro/datastack/tests/test_datastack.py .............             [  0%]
sherpa/astro/io/tests/test_backend_switch.py .......                     [  0%]
sherpa/astro/io/tests/test_io.py ............................            [  0%]
sherpa/astro/io/tests/test_io_img.py ...........                         [  0%]
sherpa/astro/io/tests/test_io_pha.py ...............                     [  0%]
```

???

DougBurke:
Rebasing so I can run the commits individually to try and get more clarity on why the python 3.11 full-build only tests start to fail.

DougBurke:
Well, that was an awful lot of effort to fix an unknown buglet.

DougBurke:
Rebased to address conflicts.

DougBurke:
Rebased on top of #2061 and add the numpy<2.0 fix to the CI runs.

sherpa/astro/io/tests/test_io_pha.py
```diff
@@ -821,10 +821,9 @@ def test_write_pha_fits_with_extras_roundtrip(tmp_path, caplog):
     assert lname == "sherpa.astro.io"
     assert lvl == logging.WARNING
 
-    # message depends on the backend
+    # message depends on the backend. TODO: check
```
DougBurke, line 824: Nope. It still seems to depend on the backend.

sherpa/astro/io/tests/test_io_pha.py
```diff
@@ -902,10 +901,9 @@ def test_pha_missing_backfile(tmp_path, caplog):
...
```
DougBurke, line 904: Ditto.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -65,20 +65,23 @@ def test_fake_pha_missing_rmf(idval, clean_astro_ui, tmp_path):
 
     rmf = tmp_path / 'rmf'
 
+    # the error message depends on the backend:
+    # - pyfits: file '{arf}' not found
+    # - crates: unable to open ...: File ... does not exist.
+    #
```
DougBurke, line 71: This comment is now wrong.

Also, we can simplify the test since we can have a single `pytest.raises` line and just change the match argument.

sherpa/astro/ui/tests/test_astro_ui_utils_simulation.py
```diff
@@ -105,20 +108,23 @@ def test_fake_pha_missing_arf(idval, clean_astro_ui, tmp_path):
 
     arf = tmp_path / 'arf'
 
+    # the error message depends on the backend:
+    # - pyfits: file '{arf}' not found
+    # - crates: unable to open ...: File ... does not exist.
```
DougBurke, line 113: Ditto.

DougBurke:
So @hamogu this is close to being reviewable, for the nth time. I just want to check the last-but-one commit where I have tried to standardize the error reporting as there have been enough changes in this PR (and its antecedents) that a review is warranted (and I have noted a few minor doc changes needed already)

sherpa/io.py
```diff
@@ -475,17 +489,33 @@ def write_arrays(filename: str,
...
+        for arg in args:
+            try:
+                narg.add(len(arg))
+            except TypeError:
+                narg.add(0)
+
+    except TypeError:
+        # args is not iterable. We catch this case when checking the
+        # size of narg.
+        pass
```
DougBurke, line 507: This code is changed twice in this PR. Perhaps that's excessive.

DougBurke:
Finally finished my clean ups and review.

sherpa/astro/io/__init__.py
```diff
@@ -1287,9 +1349,9 @@ def _reconstruct_rmf(rmf):
     """
 
     n_grp = []
```
hamogu, line 1351: While you are at it, why not type `n_grp`? I don't think either is critical, but since we type the others it'll be nice to be consistent.

DougBurke: I was trying to get by with the minimal amount of annotations to please `mypy` but I will note

- it can be advantageous to the reader to know what you expect in somewhat complex cases, as you mention
- tracking the minimal set of annotations is a fools game, as it depends on that version you are using, what software you are using, and what other changes have been made to the code base

Saying that, I'd rather not make too many other changes here as that has an impact on the downstream PRs (in this case #1921 which is being reworked to match these changes in #2085). Will need to look at this a bit more closely.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -924,15 +994,251 @@ def get_rmf_data(arg, make_copy=False):
...
+def _read_multi_pha(hdu,
+                    specnums: np.ndarray,
+                    keys: Sequence[str]) -> list[DataType]:
+
+    # Type 2 PHA file support.
+    num = len(specnums)
+
+    # We want an empty list so we can iterate over it and get back
+    # "None". As this can be used for multiple fields make sure it
+    # isnot writeable.
```
hamogu, line 1088:
```suggestion
    # is not writeable.
```

hamogu (approved):
This is one of those PRs that make my eye bleed when I try ot review them. Nothing interesting to look at, and if any of the type hints are wrong, it doesn't even matter since in Python they are just that "hints". I know it will come in useful later, but that down't make it easier to read now. However, the code itself is well used in our tests (we read a lot of data for all sotrsts of tests) so that alone gives us a high confidence that those minor changes don't break it in major ways.
I did not see anything out of the ordinary and I know that this is just one step in a larger work, so I suggest to merge it and move on.

sherpa/astro/io/__init__.py
```diff
@@ -1516,6 +1583,14 @@ def _pack_rmf(dataset):
     # TODO: is this correct?
     nchan = dataset.offset + dataset.detchans - 1
     dchan = np.int32 if nchan > 32767 else np.int16
+
+    # Technically e_min/max can be empty, but we not expect
```
wmclaugh, line 1587:
```suggestion
    # Technically e_min/max can be empty, but we do not expect
```

sherpa/astro/io/dummy_backend.py
```diff
@@ -19,42 +19,766 @@
...
+from .types import NamesType, HdrTypeArg, HdrType, \
+    ColumnsType, ColumnsTypeArg, DataTypeArg, DataType
+from .xstable import TableHDU
 
 __all__ = ('get_table_data', 'get_header_data', 'get_image_data',
-           'get_column_data', 'get_ascii_data',
-           'get_arf_data', 'get_rmf_data', 'get_pha_data',
+           'get_column_data', 'get_ascii_data', 'get_arf_data',
+           'get_rmf_data', 'get_pha_data',
+           #
```
wmclaugh, line 43: Is there some significance of adding the comment lines before and after the pack_* methods?

sherpa/astro/io/dummy_backend.py
```diff
@@ -19,42 +19,766 @@
...
+
+    Parameters
+    ----------
+    filename : str
+       The name of the ASCII file to read in.
+    ncols : int, optional
+       The number of columns to read in (the first ``ncols`` columns
+       in the file). This is ignored if ``colkeys`` is given.
+    colkeys : array of str, optional
+       An array of the column name to read in. The default is
```
wmclaugh, line 250:
```suggestion
       An array of the column names to read in. The default is
```

sherpa/astro/io/dummy_backend.py
```diff
@@ -19,42 +19,766 @@
...
+    ----------
+    arg
+        The data to read the header values from. This depends on the
+        backend but is expected to be a file name or a data structure
+        supported by the backend.
+    make_copy: bool, optional
+        If set then the returned NumPy arrays are explictly copied,
+        rather than using a reference from the data structure
+        created by the backend. Backends are not required to
+        honor this setting. The default is `True`.
```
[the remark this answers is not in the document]
DougBurke: My feeling is that python documentation guidelines have changed over time, in part because of the increasing prevalence of  language-server-esque capabilities.

In general I'm okay with transitioning to removing the "The default is blah" statements, but I think it's a case-by-case basis (e.g. for the UI code I'd prefer to keep it because it may get used when we create the CIAO ahelp files where the documentation rules are different).

sherpa/astro/io/dummy_backend.py
DougBurke: There's no reason for this Python wise - it was just useful for me to see them separate from the other routines (given the number of times I've tried to get this done). I can remove them if needed.

sherpa/astro/io/dummy_backend.py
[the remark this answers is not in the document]
DougBurke: I created #2088 to track this.

sherpa/astro/io/__init__.py
DougBurke: So, in #1921 (although #2085 is an easier one to follow) I change this code quite significantly, so I want to look at this issue in that PR and not here.

PR #2019, opened 2024-04-30
update clone link and versions and DOI

anetasie:
This is a documentation update to fix the incorrect link in the source install example. I also updated the version and DOI link in the description of the source installation.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -175,21 +175,21 @@ Obtaining the source package
 
 The source code can be obtained as a release package from
 Zenodo - e.g.
-`the Sherpa 4.10.0 release <https://zenodo.org/record/1245678>`_ -
+`the Sherpa 4.16.0 release <https://zenodo.org/record/825839>`_ -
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -175,21 +175,21 @@ Obtaining the source package
...
 
-    git clone git://github.com/sherpa/sherpa.git
+    git clone https://github.com/sherpa/sherpa.git
     cd sherpa
-    git checkout 4.10.0
+    git checkout 4.16.0
 
-will use the ``4.10.0`` tag (although we strongly suggest using a
+will use the ``4.16.0`` tag (although we strongly suggest using a
 newer release now!).
```
DougBurke, line 193: Technically `4.16.0` is the latest release, but by the time this will become "public" (in the sense it's not on the "latest" branch) we'll have released 4.16.1 so it'll be okay ;-)

DougBurke (approved):

PR #2021, opened 2024-05-02
Update Conda Deployment Workflow Action Versions

Marie-Terrell:
Updates the following actions in the deployment workflows:
- actions/checkout@v4.1.1
- actions/upload-artifact@v4
- actions/download-artifact@v4

The download action has this warning, but this is an open bug (okay for now): https://github.com/actions/download-artifact/issues/327

The update to v4 of the upload/download actions required changes to the naming conventions of the artifacts as outlined [here](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md). We previously uploaded each python version to linux-64 and osx-64, but this splits them up (and then downloads all for each OS).

### DIFF
--- a/.github/workflows/ci-conda-deployment.yml
+++ b/.github/workflows/ci-conda-deployment.yml
@@ -211,7 +211,7 @@ jobs:
     - uses: actions/download-artifact@v4
       id: download
       with:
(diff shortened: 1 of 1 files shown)


DougBurke:
I'm glad I just ignored all this in #2020 ;-)

PR #2025, opened 2024-05-05
NumPy 2.0 changes

DougBurke:
# Summary

Ensure the python code will run with NumPy 2.0.

# Details

run `ruff check sherpa --select NPY201` as suggested by https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide and address the issues. Fortunately there were not many (presumably because of #1867).

Note that this only checks the python code. It does not address any C/C++ changes we might need to make in the extension code, or in how those modules are built.


sherpa/utils/__init__.py
```diff
@@ -1780,30 +1780,30 @@ def multinormal_pdf(x, mu, sigma):
...
         raise ValueError("sigma is not symmetric")
     rank = mu.size
-    coeff = 1.0 / (numpy.power(2.0 * numpy.pi, rank / 2.0) *
-                   numpy.sqrt(numpy.abs(numpy.linalg.det(sigma))))
-    xmu = numpy.mat(x - mu)
-    invsigma = numpy.mat(numpy.linalg.inv(sigma))
+    coeff = 1.0 / (np.power(2.0 * np.pi, rank / 2.0) *
+                   np.sqrt(np.abs(np.linalg.det(sigma))))
+    xmu = np.asmatrix(x - mu)
+    invsigma = np.asmatrix(np.linalg.inv(sigma))
```
hamogu, line 1798: matrix itself is deprecated. While it's not removed right now in 2.0, it will be soon (https://numpy.org/doc/stable/reference/generated/numpy.matrix.html#numpy.matrix). Instead, one should just use normal ndarrays, that offer all the same functionality, just with different  indexing.
Do we want to do that right now or in the separate PR?

sherpa/utils/__init__.py
```diff
@@ -1833,27 +1833,27 @@ def multit_pdf(x, mu, sigma, dof):
...
-             (gamma(n / 2.0) * numpy.power(n, rank / 2.0) *
-                 numpy.power(numpy.pi, rank / 2.0) *
-                 numpy.sqrt(numpy.abs(numpy.linalg.det(sigma)))))
-    xmu = numpy.mat(x - mu)
-    invsigma = numpy.mat(numpy.linalg.inv(sigma))
+             (gamma(n / 2.0) * np.power(n, rank / 2.0) *
+                 np.power(np.pi, rank / 2.0) *
+                 np.sqrt(np.abs(np.linalg.det(sigma)))))
+    xmu = np.asmatrix(x - mu)
+    invsigma = np.asmattix(np.linalg.inv(sigma))
```
hamogu, line 1856: same here about matrix already slated for removal. MY feeling is we should get rid of it, while we touch it anyway.

sherpa/utils/__init__.py
DougBurke: I would rather do it in a separate PR because

 1. I don't have time at the moment
 2. this change needs some work (it's not just as simple as changing `asmatrix` to `asarray`, since you have to look at the code to see what it's doing, even if at the end of the day we do just go with the name change)
 3. there are likely other places in the code that `asmatrix` / `matrix` are used, not just the ones that `ruff` found were using `np.mat`

hamogu: Sounds goo to me. I'll open a new issue to keep track of this, but it's not needed for numpy 2.0.

hamogu: see #2026

hamogu (approved):

PR #2022, opened 2024-05-03
Multiprocessing updates

DougBurke:
# Summary

WIP

The aim is to address #2007

# Details

This is built on #2015

WIP

PR #2033, opened 2024-05-13
utils: add typing rules to random and parallel

DougBurke:
# Summary

Add typing statements to the sherpa.utils.parallel and sherpa.utils.random modules. There should be no changes to the functionality of these modules.

# Details

These were created for #2022 but can be considered separately.

These are low-level routines which most people are not expected to use. In adding these typing rules I have

- taken the decision that we want to try to get something that matches the "intent" of the API but it may not actually match the reality (there are fun discussions about `Sequence` vs `np.ndarray` and `float` vs `SupportsFloat` that can hide the actual code). It's also the case that until we have "all" areas typed (or, at least, modules that interact, such as fit/stats/optimizers) that we won't know the "best" type to use.

- created a number of protocols to represent types from the multiprocessing module, since we can not guarantee that this module (or the threading module, which multiprocessing tends to replicate) are present. It does, however, make it more obvious what features we need.

- the multiprocessing.BaseContext class, which would be the obvious class to "type" the sherpa.utils.parallel.context variable (modulo the need for using a protocol) is not useful since this class does no have a Process attribute, whereas all the actual "usable" contexts do. This leads to several "type: ignore[assignment]" statements

- the use of generic types to track the values created/used by the random routines, and to make sure we have the correct callable and queue, can be considered somewhat experimental: this seems to work but we need to see how it is handled long term to know if it's useful or not. It is interesting (to me) that our use of the queue type is technically more constrained than the general version (i.e. our use of the queue assumes that each queue only handles a specific type, rather than just sending around "anything", and the question becomes how do we take advantage of this)

- it would be nice to use type aliases, but we still need to support Python 3.9 (although this may be relaxed soon) and the behavior changes between Python 3.10 (TypeAlias) and Python 3.12 (type and TypeAliasType). This is most keenly felt for the RandomType alias - at the moment we have it but do not directly export it (although we allow other code to use it).

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -881,7 +881,7 @@ or either). The aim is to try and model the intention of the API
 without matching every single possible type that could be used. The
 typing rules are also currently **not** checked in the Continuous
 Integration runs, or required to be run as part of the review process
-of pul requests.
+of pull requests.
 
 .. _handling_nd:
 
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -863,6 +863,26 @@ Notes
 
 Notes on the design and changes to Sherpa.
 
+.. _typing_statements:
+
+Adding typing statements
+------------------------
+
+Typing rules, such as::
+
+  def random(rng: Optional[RandomType]) -> float:
+
+are being added to the Sherpa code base to see if they improve the
+maintenance and development of Sherpa. This is an incremental process
+and it is likely that existing typing statements will need to be
+updated when new rules are added (for instance, it is not always
+obvious when a routine accepts or returns a sequence, a NumPy array,
+or either). The aim is to try and model the intention of the API
+without matching every single possible type that could be used. The
+typing rules are also currently **not** checked in the Continuous
+Integration runs, or required to be run as part of the review process
+of pul requests.
+
 .. _handling_nd:
 
 N-dimensional data and models
--- a/sherpa/utils/parallel.py
+++ b/sherpa/utils/parallel.py
@@ -31,14 +31,140 @@
 
 """
 
+from abc import abstractmethod
 from configparser import ConfigParser
 import inspect
 import logging
-from typing import Final
+from typing import Any, Final, Optional, Protocol, Sequence, TypeVar
 
 import numpy as np
 
 from sherpa import get_config
+from .random import RandomType
+
+# A numer of symbols has been added to this module in release 4.17.0
+# to allow typing statements to be made. This is partly because the
+# multiprocessing module, which provides a number of symbols that
+# would be used in such statements, is optional (and so is the
+# threading module which multiprocessing often uses to define
+# symbols).  Therefore a number of protocols have been added to allow
+# statements to make statements like "this requires a queue which
+# handles exceptions".
+#
+I_contra = TypeVar("I_contra", contravariant=True)
+O_co = TypeVar("O_co", covariant=True)
+T = TypeVar("T")
+
+
+class Callback(Protocol[I_contra, O_co]):
+    """Simple callback of a single argument."""
+
+    def __call__(self, arg: I_contra) -> O_co:
+        ...
+
+
+class CallbackWithRNG(Protocol[I_contra, O_co]):
+    """Simple callback of a single argument plus RNG argument."""
+
+    def __call__(self,
+                 arg: I_contra,
+                 rng: Optional[RandomType]) -> O_co:
+        ...
+
+
+# Since importing multiprocessing and threading are allowed to fail it
+# is hard to type things with a class defined in these modules.
+#
+class SupportsProcess(Protocol):
+    """Label those methods from mulltiprocessing.Process we need."""
+
+    exitcode: Optional[int]
+
+    @abstractmethod
+    def start(self) -> None:
+        ...
+
+    # Technically join accepts an optional timeout argument but we do
+    # not need that.
+    #
+    @abstractmethod
+    def join(self) -> None:
+        ...
+
+    @abstractmethod
+    def terminate(self) -> None:
+        ...
+
+# This typing rule is actually less generic than the actual queue
+# type, since there's no guarantee that it is only used to
+# send/receive a single type. However, the use here does have the
+# queues having a fixed type.
+#
+class SupportsQueue(Protocol[T]):
+    """Represent the multiprocessing Queue class"""
+
+    @abstractmethod
+    def empty(self) -> bool:
+        ...
+
+    # we do not need the block or timeout options
+    @abstractmethod
+    def put(self, obj: T) -> None:
+        ...
+
+    # we do not need the block or timeout options
+    @abstractmethod
+    def get(self) -> T:
+        ...
+
+
+class SupportsLock(Protocol):
+    """Represent the multiprocessing Lock class"""
+
+    @abstractmethod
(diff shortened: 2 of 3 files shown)


sherpa/utils/parallel.py
```diff
@@ -302,22 +456,30 @@ def run_tasks(procs, err_q, out_q, num=None):
...
+        if r is None:
+            raise RuntimeError("task failed")
         vals.extend(r)
 
     return vals
 
 
-def parallel_map(function, sequence, numcores=None):
+
+
```
DougBurke, line 478:
```suggestion
```

sherpa/utils/random.py
```diff
@@ -120,7 +147,11 @@ def random(rng):
     return rng.random()
 
 
-def uniform(rng, low, high, size=None):
+def uniform(rng: Optional[RandomType],
+            low: float,
+            high: float,
+            size=None
```
DougBurke, line 153: I perhaps should have added a note about why I have not added a type for size. This is because it would have to be something like

```
    size: Optional[int | Sequence[int] | ndarray]
```

and I didn't think it helped to say something like this.

docs/developer/index.rst
```diff
@@ -861,6 +861,26 @@ Notes
...
+are being added to the Sherpa code base to see if they improve the
+maintenance and development of Sherpa. This is an incremental process
+and it is likely that existing typing statements will need to be
+updated when new rules are added (for instance, it is not always
+obvious when a routine accepts or returns a sequence, a NumPy array,
+or either). The aim is to try and model the intention of the API
+without matching every single possible type that could be used. The
+typing rules are also currently **not** checked in the Continuous
+Integration runs, or required to be run as part of the review process
+of pul requests.
```
hamogu, line 882:
```suggestion
of pull requests.
```

sherpa/utils/random.py
DougBurke: Comment has now been added.

hamogu (commented):
You say in several places "We don't requite multiprocessing to be available". Yet, it's part of the standard library, so in normal Python installations we always have it. What is a "normal" Python installation? Well, any that's not EMscriptlen and not WASI (https://docs.python.org/3/library/multiprocessing.html) I suspect that many more parts of sherpa will break when we try to compile in to WASI so maybe we don't ahve to be that careful, and can assume that we have multiprocessing and worry about the speical cases later when it's atually needed?
I would not be surprised if multiprcessing gets added to e.g. Python on emscripten before we get around to using Sherpa that way.

sherpa/utils/parallel.py
```diff
@@ -259,15 +410,18 @@ def die():
 
     except KeyboardInterrupt as e:
         # kill all slave processes on ctrl-C
-        die()
+        cleanup_tasks(procs)
```
hamogu, line 413: Much nicer naming!

DougBurke: It's the little things in life!

sherpa/utils/random.py
```diff
@@ -27,17 +27,44 @@
...
+                  ) -> SherpaFloat:
+    ...
+
+@overload
+def poisson_noise(x: Sequence[SupportsFloat],
+                  rng: Optional[RandomType] = None
+                  ) -> np.ndarray:
+    ...
+
+def poisson_noise(x: Union[SupportsFloat, Sequence[SupportsFloat]],
```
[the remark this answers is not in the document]
DougBurke: I had thought I'd seen stuff where the "actual" function still had typing arguments, but I'm happy to be proven wrong.

Looking into https://peps.python.org/pep-0484/ it does say

> The @overload-decorated definitions are for the benefit of the type checker only, since they will be overwritten by the non-@overload-decorated definition, while the latter is used at runtime but should be ignored by a type checker.

which would match your description (and makes sense).

DougBurke:
Rebased to pick up typing changes on #2045 and to address the `overload` discussion above - https://github.com/sherpa/sherpa/pull/2033#discussion_r1611754501

sherpa/utils/random.py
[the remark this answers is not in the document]
DougBurke: So, I've removed this relpication, which is nice. As part of this I've extended the use of overload as some of the functions would return a scalar if `size=None` and an ndarray otherwise, so we want to encode this behavior which `overload` does.

DougBurke:
As well as the overload changes I added a few more asserts in `parallel` to please `pyright`, now that #2045 has landed which cleaned up a complaint that it thought get_config() could return None.

DougBurke:
Just to comment on the "We don't requite multiprocessing to be available" note. I never added multiprocessing support, we just have code that does things like

```python
try:
    import multiprocessing
except ImportError:
    cry()
```

I probably added text just to say it's optional, but it wasn't my original design decision. We could drop it - ie require multiprocessing. I will note that working on issue #2007 and noting the "rapid" changes in the interest for changing the internals, I am not sure what the best approach is here.

hamogu:
While multiprocessing may not have been available in the past (I remember something like that, but I'm not sure about details), today is is part of the standard library and CPython has it on any platform except those explicitly listed (currently emscripten and WASI). I bet that all sorts of other things will break on that, too (e.g. on emscripten there is no local file system to read or write from), so we are probably not compatible with that right now anyway (we don't even test on windows with is much easier to support).

On the other hand, I could see us targeting browser support in the future (e.g. if the ROSES F.7 proposal is funded), so, while it won't work right now, there is some argument to be made that at least we shouldn't introduce any *additional* incompatibilities beyond those that we already have.

So, since the code is written already to have multiprocessing optional, I suggest to leave it as written in this PR.

sherpa/utils/random.py
```diff
@@ -27,16 +27,41 @@
...
+
 import numpy as np
 
 from .numeric_types import SherpaFloat
 
+# This should probably be an explicit type alias but for now leave it
+# like this. Users are expected to use the Generator rather than
+# RandomState form, but both are kept in (e.g. for testing).
+#
+# Once Python 3.10 or 3.12 is the minimum we can make this an actual
```
hamogu, line 41: And 3.10 is hopefully the minimum version very soon...

hamogu (approved):
Just one inconsequential typo, so I'm approving.

sherpa/utils/parallel.py
```diff
@@ -31,14 +31,140 @@
...
+                 arg: I_contra,
+                 rng: Optional[RandomType]) -> O_co:
+        ...
+
+
+# Since importing multiprocessing and threading are allowed to fail it
+# is hard to type things with a class defined in these modules.
+#
+class SupportsProcess(Protocol):
+    """Label those methods from mulltiprocessing.Process we need."""
```
wmclaugh, line 79:
```suggestion
    """Label those methods from multiprocessing. Process we need."""
```

sherpa/utils/parallel.py
```diff
@@ -31,14 +31,140 @@
...
 import logging
-from typing import Final
+from typing import Any, Final, Optional, Protocol, Sequence, TypeVar
 
 import numpy as np
 
 from sherpa import get_config
+from .random import RandomType
+
+# A number of symbols has been added to this module in release 4.17.0
```
wmclaugh, line 45:
```suggestion
# A number of symbols have been added to this module in release 4.17.0
```

PR #2035, opened 2024-05-14
stats: add typing rules

DougBurke:
# Summary

Add typing rules to sherpa.stats and minor internal clean up to the module.

# Details

This is spun out of #2022 (similar to #2033). The idea being that we need all these separate parts to have typing rules to be able to catch any issues.

This improves the typing rules for the routines in `sherpa.stats`. As mentioned in #2022 these are somewhat aspirational, since the reality of how they could be used is "tricky", but they have pointed out some internal issues. The `sherpa.stats.Stat` hierarchy could see some changes but it isn't worth it at this time (i.e. only worth doing if it would add some functionality).

Three "interesting" things this has shown are

- the `calc_staterror` routine is now marked `staticmethod` in the base class since some user-facing classes have it set up - and in fact the xspec variance needs it related to issue #356 - *but* the `UserStat` class needs it to be a normal method (ie have access to the object). The options are
  1. live with this mismatch - this is the option I have picked
  2. make everything a `staticmethod` - which I don't think is possible
  3. fix the code to allow `staticmethod` to be removed, which I just don't want to do at this time
- the tests seem to make interesting use of the class hierarchy, using `UserStat` as their starting point, when they should probably just extend the class they are "extending"
- a lot of the tests require having xspec, the data, and an IO package installed when they really should have most tests run without needing them; there's also a lot of use of UI level code when we should be using the objects directly (with ui tests in the ui area)

So there's more work we could do, but I don't really see it being worth it at this time (other than perhaps fixing the `staticmethod` issue).

### DIFF
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -825,55 +825,50 @@ class UserStat(Stat):
     """
 
     def __init__(self, statfunc=None, errfunc=None, name='userstat'):
+        # These two could be checked by comparing
         self._statfuncset = False
-        self.statfunc = (lambda x: None)
-
         self._staterrfuncset = False
-        self.errfunc = (lambda x: None)
+
+        # statfunc and _calc serve the same purpose but leave as is
+        # in case users use the statfunc field
+        self.statfunc = None
+        self.errfunc = None
 
         if statfunc is not None:
             self.statfunc = statfunc
             self._calc = statfunc
-            self._statfuncset = True
 
         if errfunc is not None:
             self.errfunc = errfunc
-            self._staterrfuncset = True
 
         super().__init__(name=name)
 
     def __getstate__(self):
+        # TODO: Do we still need to do this?
         state = self.__dict__.copy()
-        # Function pointers to methods of the class
-        # (of type 'instancemethod') are NOT picklable
-        # remove them and restore later with a coord init
         del state['statfunc']
         del state['errfunc']
-
         return state
 
     def __setstate__(self, state):
-        # Populate the function pointers we deleted at pickle time with
-        # no-ops.
-        self.__dict__['statfunc'] = (lambda x: None)
-        self.__dict__['errfunc'] = (lambda x: None)
+        # TODO: Do we still need to do this?
+        self.__dict__['statfunc'] = None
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased so that it relies on #2060 - so that we can take advantage of `sherpa.utils.types`.

DougBurke:
Rebased now that #2060 has been merged.

sherpa/stats/__init__.py
```diff
@@ -55,28 +56,40 @@
 
 
 class Stat(NoNewAttributesAfterInit):
-    """The base class for calculating a statistic given data and model."""
+    """The base class for calculating a statistic given data and model.
+
+    .. versionchanged:: 4.17.0
+       Code relying on access to private information may need to be
+       updated, as some fields have changed or been renamed.
```
hamogu, line 63: Do we really need a comment in the code to track changes to private fields? That's what "private" implies - we can change it any time; use at your own risk.

hamogu (approved):
For the testcode, it looks to me like that was modified from some even older (probably pre-git) text version and then adapted to what is tested here. So, I don't think we have to worry about removing things that are not used in current tests. 

As for the messiness of staticmethod or not: (a) I agree that that's beyond a PR that just adds typing, and (b) I think it exposes a problematic design patters. The "pythonic" way of doing it would be to make users define new classes inheriting from our classes. Instead, we have this mixture were we define a class hirachy, but users can set the statistics functions in some other way (e.g. `set_statfunc`). 

Unless we deprecate that mechanism, we'll always have some sort of messiness in typing. Now, I suspect that few, if any users use this mechanism (I think the more sophisticated programmers who know how to use Python to make functions and then pass those function around as an object, are more likely to use inheritance as we do for Sherpa's on classes and the casual users don't define their own statistics), but it's part of the public API, so we certainly can't take it out off-hand in a PR like this.

And it works, so it might be easier (even if not visually appealing) to leave it untouched and spend our energy on other things.

sherpa/stats/__init__.py
DougBurke: I think this is just a reflexive change based on the outcome of #2015 - trying to better warn users about possible changes....

PR #2038, opened 2024-05-14
Add an explicit code on conduct specific to this project

hamogu:
This is a fairly standard software project doe of conduct that's been adopted by many projects in the community. I reviewed the SAO, CfA, and MIT specific codes, but they all don't really apply here, as they typically center on physical visitors to our facilities. Also, it is our long-term goal to make Sherpa a viable source project by itself and not just one of many scripts provided by the CXC. Thus, I think it's better to start with a CoC that's independent of our respective institutions.
However, as a start, I still put "contact the help desk" as the method to report CoC concerns to project leadership.

### DIFF
--- a/code_of_conduct.md
+++ b/code_of_conduct.md
@@ -1,0 +1,134 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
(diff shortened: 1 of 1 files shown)


hamogu (commented):
I note that a Code of Conduct is required for all NASA funded software projects under SPD-41a, see Section 3. D (iv) in https://smd-cms.nasa.gov/wp-content/uploads/2023/08/smd-information-policy-spd-41a.pdf

anetasie (approved):
Yes, we should include it!

code_of_conduct.md
```diff
@@ -0,0 +1,134 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official email address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement per email to
+cxchelp@head.cfa.harvard.edu .
```
wmclaugh, line 64: Not sure if the email should be cxchelp@cfa.harvard.edu (drop the 'head.') or if they need to submit a ticket via https://cxc.cfa.harvard.edu/help/ per [#2139](https://github.com/sherpa/sherpa/pull/2139) - Update helpdesk URL

hamogu, line 64:
```suggestion
cxchelp@cfa.harvard.edu .
```

hamogu:
As I understand it, sending an email will create a ticket. Even if not, email should be fine.

PR #2039, opened 2024-05-14
Add a security reporting policy

hamogu:
This text assumes that someone with "admin" permission in the repro (probably Warren?) activates github private security reporting, which wont' hurt: https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository

If that form is activated, all repro admins will get an email and see the report in the "security" tab on the repro. They can then decide how to continue, e.g. share the report with other developers as needed.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Supported Versions
+
+We only support the latest minor version of `sherpa` with security updates. 
+
+## Reporting a Vulnerability
+
+[GitHub private security vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability)
+should be enabled for `sherpa`. But if that does not work for whatever reason, please report to 
+cxchelp@cfa.harvard.edu.
+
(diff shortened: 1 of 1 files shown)


hamogu (commented):
For context, this is how e.g. astropy handles it. We have a about 4 person security team with admin access (so they are able to see the reports) and then triage. We've gotten a handful reports in about a year and most of them are formally "severe" by stnadard ranking, but give the way our software is used, it's not very likely that they are exploited - e.g. in some cases a user would have to manually construct a manipulated dataset to trigger it. I image that for Sherpa there will be even less activity, but I do think it is good practice to spell out what we would do, how to contact us if there is something, and think about who will read and act on reports.

anetasie:
We did not discuss this. I agree with @hamogu that this process could be good for our project.

DougBurke:
I always think that these things are good to have and we can revisit if things need to be changed, but better to have something than nothing. I haven't looked too closely, but @anetasie if you want to approve this I'm not going to complain.

anetasie (approved):

anetasie:
@DougBurke I just approved it!

PR #2043, opened 2024-05-20
docs: improve the simulation docs

DougBurke:
# Summary

Improve the simulation documentation to better reflect the changes to `fake_pha` made in #1684.

# Details

This was part of #1684 but I never got time to finish it (and hence include it)  for that PR. There's two parts

1. improve the documentation
2. make it so that the plots are generated during the documentation build

I want to focus on the first bit, as the second bit has some complications. If we can use a different dataset that we are guaranteed is present in the build then this would simplify things.

I've made #2122 as the "only-the-doc-changes" to try not to break the review @hamogu has just done. So this can then become the "make the plots be generated during build time" PR.


DougBurke:
@hamogu - this is an "interesting" one. I wanted to update the `fake_pha` documentation given recent changes in #1684, but then it snow-balled a bit as I wanted to create the plots during documentation creation (as part of an attempt to keep our docs honest). This has turned out to be more annoying than I had hoped, but I think I now understand it that we can just look at the new docs in

https://sherpa--2043.org.readthedocs.build/en/2043/evaluation/simulate.html

and compare them to  the last released notes

https://sherpa.readthedocs.io/en/4.16.1/evaluation/simulate.html

It's been a while, but I think the low-level `fake_pha` call now needs a model to include any instrument terms in it. I am not sure if the examples are now meaningful, after #1684 and possibly earlier), so we have

- check the examples make sense
- do the new plots help or hinder
- there are a couple where we don't create plots because they require more setup code (e.g. creating a background model or an extra response) and should we try to impove the text to include this (or hide it away from the reader) or should we drop them

anetasie:
@DougBurke good idea to include plots. Could you add the errorbars to the simulated data for these plots?

DougBurke:
I've split out the commits so there's "doc changes" and then "build the plots during the build" steps, as the latter is surprisingly awkward, and should be done in a follow-up PR.

DougBurke:
@hamogu - the documentation changes in the first commit need a review. I am not convinced I actually finished adding the text, but I wrote this so long ago I don't remember.

I think the second commit will get dropped (as there are issues in doing this that I do not have time to review and it doesn't add to the actual text).

docs/_examples/evaluation/simulate/plots.py
```diff
@@ -0,0 +1,120 @@
...
+        sys.exit(1)
+
+def savefig(name):
+    plt.savefig(f"_static/evaluation/{name}.png")
+    print(f"** Created: {name}.png")
+
+data_dir = str(indir)
+if not data_dir.endswith("/"):
+    data_dir += "/"
+
```
hamogu: Instead of this, you could also do the following below:
`os.path.join(data_dir, '9774_bg.pi')`
That will take care of the "/" at the end of the data_dir string or not. It will also work on windows, which uses "\" as a separator. (Not that we currently support windows, but just saying in general...)

hamogu (commented):
I'm assuming you asked me to review the first two commits and plan to drop the third?

docs/_examples/evaluation/simulate/plots.py
```diff
@@ -0,0 +1,120 @@
...
+from matplotlib import pyplot as plt
+
+import numpy as np
+
+from sherpa.utils.random import poisson_noise
+
+# Assume run from docs directory.
+#
+indir = Path("../sherpa-test-data/sherpatest")
+outdir = Path("_static/evaluation/")
```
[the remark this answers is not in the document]
DougBurke: This is just meant to be a quick helper script that we can hopefully remove as some point soon, so I'm not too worried about improving it!

docs/_examples/evaluation/simulate/plots.py
DougBurke: As I mentioned, I think the last commit is going to get removed for 4.17.0, but useful to think about for the future.

DougBurke: Also, this is meant to replicate the code in `docs/evaluation/simulate.rst` and so I only want to change the "setup" code (since `data_dir` in the actual documentation is provided out-of-band).

DougBurke:
Closing due to #2345

Issue #835, opened 2020-06-22
simplex and levmar optimization failes in what looks like noremal data


hamogu:
I have a what I consider a pretty normal optimization problem: Fit an emission line on a constant background and the simplex optimization fails spectacularly (the model is orders of magnitude above the data for all data points) and levmar is only little better (it drives the line flux to 0, and tries to fit it all with the background only, which gives a better statistic than simplex, but not the right answer). 

There is quite some setup, in that the results are very different for different binnings, notice or ignore regions, but it's not a real pathological case. `int_proj` shows that the statistic is well-behaved (smooth function). I suspect that the problem is in the boundaries conditions that I put in (background and line flux >= 0), but both are physically well motivated and fitting the cash statistics when the model can go negative can have bad consequences (I'll open a separate issue with that next time I come across it). 

A look at the fit output shows me that the fit twiddles both parameters down and then suddenly jumps them up by several orders of magnitude, twiddels them down, and jumps them up again, but never really explores the region with line.ampl below the initial value (which is where the best fit sits.)

If you want to see the best fit, start with line.ampl = 5e-5 or switch to moncar.

Data to reproduce is in `/pool14/guenther/sherpa_simplex_fit`
```
from sherpa.astro import ui

ui.load_data('P0112880201R2S005SRSPEC1003.FTZ')
ui.load_rmf('P0112880201R2S005RSPMAT1003.FTZ')

ui.set_stat('cash')
ui.set_analysis('wave')
ui.group_counts(1)
ui.ignore(None,11.4)
ui.ignore(11.7, None)
ui.notice(11.4, 11.7)

bkg = ui.const1d(name='bkg')
bkg.c0.min = 0  # Fit goes heywire if we ever get negative values in model with cstat or cash
bkg.c0 = 7e-4
line = ui.delta1d(name='line')
line.pos = 11.5467
line.pos.frozen = True
line.ampl = 5e-4
line.ampl.min = 0
ui.set_source(bkg + line)


ui.set_method('simplex')
# or ui.set_method("levmar")
ui.fit(outfile='simplex.dat', clobber=True)

ui.plot_fit()
```

Note that I'm not looking for advice on how to solve this specific problem (I can just switch to "moncar" and it all works fine), but I want to raise this issue because I feel that this is not a very special situation and we should do better than this. Both standard fitters fail with default parameters.

[...]

PR #2047, opened 2024-05-22
Opt cleanup

DougBurke:
# Summary

Improve the typing coverage of the sherpa.optmethods code and internal code clean up.

# Details

This is a follow-on to #2045 and takes in some of the changes in #2015. However, there is less of an attempt to mark symbols as deprecated as these are definitely internal routines for this module. Much of the code removal is to remove debugging information that only someone who could edit the code could get to see. This, along with some typing annotation, and some understanding of the code from #2007 / #2022, has lead to some minor code cleanups. I am also conscious of improving the documentation given https://github.com/sherpa/sherpa/discussions/2042

The typing rules that are added are meant to try and represent the intent of the interfaces, rather than exactly matching the reality. For example, there's subtle differences between a Sequence and ndarray that can cause mypy to complain over things that are "obviously" correct - e.g. can you call len on the object - as well as subtle differences between float and SupportsFloat that do not really change the code. I claim that until we have all the major parts working together it's hard to nail down these types, but that it has actually shown be some problems (or potential problems, for example the `fix logical error with population_size argument" [commit](https://github.com/sherpa/sherpa/pull/2079/commits/0816550159747a31170dd71b295f3e1a0d168c94) which was originally here but I moved into the now-merged #2079).

I have added some notes of possible future work (none of them would be significant changes), but it's hard to make these changes as those code paths are currently not tested.

The idea is that this PR should not cause difference in behavior with code using Sherpa unless they are doing something really funky.

Note that this is part of the work to address #2007, but there's a lot of cleanup needed to understand the code and make sure that changes don't break things, as well as some of the changes needed to address #2007, such as (probably, it's hard to guarantee for sure given all the code that would need to be changed to show this for sure) the changes in `optfcts: extract out the 'infinite potential' idea`.

### DIFF
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2018, 2020, 2021, 2023, 2024
+#  Copyright (C) 2007, 2015, 2018, 2020, 2021, 2023 - 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -43,7 +43,10 @@ match::
 
     callback(pars, *statargs, **statkwargs)
 
-and return the statistic value to minimise.
+and return the statistic value to minimise along with the per-bin
+statistic values (note that per-bin here refers to the independent
+axis of the data being fit, and not the number of parameters being
+fit).
 
 Notes
 -----
@@ -98,7 +101,8 @@ depending on the optimiser:
 >>> print(f"Best-fit value: {res[1][0]}")
 Best-fit value: 4.0
 
-We can see that the model has been updated thanks to this:
+We can see that the model has been updated thanks to the use of
+``cb``, which sets the ``thawedpars`` attribute of ``mdl``:
 
 >>> print(mdl)
 const1d
@@ -109,13 +113,16 @@ const1d
 """
 
 import logging
+from typing import Any, Callable, Sequence
 
 import numpy as np
 
 from sherpa.utils import NoNewAttributesAfterInit, \
     get_keyword_names, get_keyword_defaults, print_fields
-from sherpa.optmethods.optfcts import grid_search, lmdif, montecarlo, \
-    neldermead
+from sherpa.utils.types import ArrayType, OptReturn, StatFunc
+
+from .optfcts import grid_search, lmdif, montecarlo, neldermead
+
 
 warning = logging.getLogger(__name__).warning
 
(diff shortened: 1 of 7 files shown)


DougBurke:
Rebased as I try to make `sherpa.utils.types` a thing.

DougBurke:
Rebased to pick up recent changes and re-ordered the commits.

DougBurke:
Rebased and took the time to change `Optional[xxx]` to `xxx | None` as we no longer need to support python 3.9.

DougBurke:
Rebased as #2207 has been merged. I have also extracted out a `Callback` class to go with the `InfinitePotential` class: these could

- have a more-sensible naming scheme
- be arranged into a hierarchy

but for now I'm not sure this would provide any benefit.

DougBurke:
Rebased to pick up #2208

sherpa/optmethods/optfcts.py
```diff
@@ -445,9 +502,32 @@ def stat_cb0(x_new):
...
+                 xmin: np.ndarray,
+                 xmax: np.ndarray
+                 ) -> np.ndarray:
+    """Create an array of values between xmin and xmax."""
+
+    out = np.zeros_like(xmin)
+    for idx, (minval, maxval) in enumerate(zip(xmin, xmax)):
+        out[idx] = random.uniform(rng, minval, maxval)
+
+    return out
```
hamogu, line 563: This function is unnecessary. Numpy itself allows `low` and `high` argument in the random number generators to be arrays already. Remove this function and inline the lumpy call directly where needed:
```Python
rand_start = random.uniform(rng, xmin, xmax)
```

Performance will be better (no Python loop), but more importantly, the code it easier to understand because it avoids an extra level of indirection.

hamogu (changes requested):

sherpa/optmethods/optfcts.py
hamogu: I know this function was there before, but sine this commit touches it, it might as well remove it altogether. I checked the docs back to lumpy 1.20 and at least the far `low` and `high` can be arrays.

DougBurke: Note that `random.uniform` is actually `sherpa.utils.random.uniform`. Now, we can do what we like with this routine, but it just needs some work to check.

For reference, this routine will take the `rng` argument and do something sensible if `rng` is `None`.

hamogu: It should just pass through the arrays, because `sherpa.utils.random.uniform` call numpy in both cases and will just do the right thing: https://github.com/sherpa/sherpa/blob/main/sherpa/utils/random.py#L188

hamogu: Which will means tat tis function is not needed, `just all sherpa.utils.random.uniform` directly.

DougBurke:
@hamogu - one thing I'm interested in is the "best" way to change local callbacks, where we previously defined a symbol within a routine like

```python
    # fcn is an argument to the enclosing function
    def cb(x):
        return fcn(x)[0]
```

to a new class-based system, like

```python
class Callback:
    def __init__(self, fcn):
        self.fcn = fcn

    def __call__(self, x):
        return self.fcn(x)[0]
```

where we then cam just say (in the function)

```python
    cb = Callback(fcn)
```


I have been doing this because the local callbacks seem to cause an issue for `multiprocessing` when using non-`fork` methods - for more information see #2007 - but I wonder if there's other ways we should do this? The best I could think of was a decorator, but I don't think it would save complexity (but I haven't tried to get it to work). I think most of these classes are going to be local to the module (although I have put some somewhere in `sherpa.utils` in my experiments) so I think it's okay (i.e. this complexity is "hidden within" the module).

hamogu:
Effectively, what you are doing here is to define a protocol. `Callback` is one function that implements that protocol, `InfinitePotential` is the other. They are not derived in a hierarchy, but they have the same call signatures for `__init__` and `__call__` (including the same statement `return self.func(pars)[0]`).

Since you are into typing, you of course have the [option to make the definition explicit for static type checking)](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols) (not sure about the Python version where this becomes available), but as far as Python itself is concerned when running, it's just Duck-typing.

Any other class that follows that definition could also be used in it's place (for example, I have long wanted to make a "Not-quite infinite potential" because I noted that some optimizers get stuck one they hit the infinite part, instead, I think we should have a "sloped infinite potential" which returns "really_big_number + 1e4 * x" where "x" is the distance to the boundary where solutions are valid, to motivate the optimizer to move back into the valid region - see #835.

Of course, since the effect is so simple, a decorator would also work, but I don't think that would be easier to understand when reading the code, nor would it be noticeably faster when running. So, I think classes are good.

sherpa/optmethods/optfcts.py
DougBurke: I've removed this routine (and added a test just to show that `sherpa.utils.random.uniform` is doing the right thing).

There's a 50/50 chance the tests will fail because of some other recent change to OTS packages ....

DougBurke: @hamogu - is this change okay?

hamogu (approved):

sherpa/optmethods/__init__.py
```diff
@@ -135,29 +142,65 @@ class OptMethod(NoNewAttributesAfterInit):
...
+
+        fcn:  Callable[[ArrayType, ...], tuple[float, np.ndarray]]
+        x0:   ArrayType
+        xmin: ArrayType
+        xmax: ArrayType
+
+        ArrayType = Sequence[float] | np.ndarray
+
+    and the remaining arguments are sent in as keyword arguments, and
+    so can specific to the optimization function. The x0, xmin, and
```
wmclaugh, line 164:
```suggestion
    so can be specific to the optimization function. The x0, xmin, and
```

DougBurke:
@wmclaugh - are you waiting for anything more for this PR?

sherpa/utils/tests/test_random.py
```diff
@@ -168,3 +168,35 @@ def test_choice_rng_is_none():
         np.random.seed()
 
     assert a == pytest.approx(b)
+
+
+def test_uniform_ranges():
+    """Check uniform behaves as expected.
+
+    The behaviour tested here is disussed in
```
wmclaugh, line 176:
```suggestion
    The behaviour tested here is discussed in
```

PR #2048, opened 2024-05-22
CI: update the weekly arch run to add ppc64le and to use a supported python version

DougBurke:
# Aim

Support the weekly test builds using aarch64 and ppc46le architectures. This support is considered experimental.

# Details

It looks like the ubuntu_rolling dockerfile has bumped the Python version to 3.12, which we currently do not support. Switch to the 22.04 LTS which uses Python 3.10.

Also update the installation to take advantage of recent changes in the build infrastructure (mainly that we now have a test installation option).

In doing this I noted that the ppc64le architecture can be turned back on - as noted in the AstroPy reports 

- https://github.com/astropy/astropy/pull/11697
- https://github.com/astropy/astropy/pull/14310

this looks like it was down to some "behind the scenes" issue in the build/emulation and not in out code base.

### DIFF
--- a/.github/workflows/ci-pip-arch.yml
+++ b/.github/workflows/ci-pip-arch.yml
@@ -45,7 +45,9 @@ jobs:
       uses: uraimo/run-on-arch-action@v2
       with:
         arch: ${{ matrix.arch }}
-        distro: ubuntu_rolling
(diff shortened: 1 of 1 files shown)


Marie-Terrell (approved):
Looks good to me!

PR #2051, opened 2024-05-28
Optfcts compiler warnings

DougBurke:
Try to address #1542


DougBurke:
This has been replaced by #2494

Issue #2049, opened 2024-05-27
XSPEC tests can crash on macOS with XSPEC 12.14.0


DougBurke:
Noticed after updating 

- #1793
- #1615 

to support XSPEC 12.14.0 (#1997)

I added `--verbose` to the `pytest` command so we can see where the segfault happens (I know, from doing this dance before, that it doesn't mean there's memory corruption in this model, but it's a start):

```
  sherpa/astro/xspec/tests/test_xspec.py::test_evaluate_xspec_model_noncontiguous2[XSbvvnpshock] PASSED [ 49%]
  sherpa/astro/xspec/tests/test_xspec.py::test_evaluate_xspec_model_noncontiguous2[XSbvvpshock] PASSED [ 49%]
  sherpa/astro/xspec/tests/test_xspec.py::test_evaluate_xspec_model_noncontiguous2[XSbvvrnei] PASSED [ 49%]
  Fatal Python error: Segmentation fault
  
  Current thread 0x0000000110d63600 (most recent call first):
    File "/Users/runner/work/sherpa/sherpa/sherpa/astro/xspec/__init__.py", line 1898 in calc
    File "/Users/runner/work/sherpa/sherpa/sherpa/models/model.py", line 473 in cache_model
    File "/Users/runner/work/sherpa/sherpa/sherpa/models/model.py", line 808 in __call__
    File "/Users/runner/work/sherpa/sherpa/sherpa/astro/xspec/tests/test_xspec.py", line 779 in test_evaluate_xspec_model_noncontiguous2
...
   File "/usr/local/miniconda/envs/build/bin/pytest", line 10 in <module>
  
[...]

PR #2052, opened 2024-05-28
XSPEC: do not test bvvnei unless 12.14.0i or later is used

DougBurke:
# Summary

Avoid testing the XSbvvnei model if XSPEC <= 12.14.0i is in use, as the model can cause a crash in this case.

# Details

Calling the XSbvvnei class can cause a seg fault (seen on macOS Intel) if XSPEC 12.14.0h is used. This turns out to be a bug fixed in XSPEC 12.14.0i, so we just skip the model if an earlier version of XSPEC is in use. This model was added in XSPEC 12.14.0.

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec.py
+++ b/sherpa/astro/xspec/tests/test_xspec.py
@@ -117,6 +117,15 @@ def get_xspec_models():
                    "12.14.0c", "12.14.0d", "12.14.0e"]:
         remove_item(model_names, 'XSbsedov')
 
+    # The bvvnei model causes a crashe with XSPEC 12.14.0 to 12.14.0h
+    # (it should be fixed in 12.4.0i and later). The model is not
+    # present before 12.14.0.
+    #
+    if version in ["12.14.0", "12.14.0a", "12.14.0b",
+                   "12.14.0c", "12.14.0d", "12.14.0e",
+                   "12.14.0f", "12.14.0g", "12.14.0h"]:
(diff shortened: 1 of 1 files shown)


DougBurke:
So, this doesn't prove it's fixed the problem in #2049 - but it does at least pass!

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -117,6 +117,15 @@ def get_xspec_models():
                    "12.14.0c", "12.14.0d", "12.14.0e"]:
         remove_item(model_names, 'XSbsedov')
 
+    # The bvvnei model causes a crashe with XSPEC 12.14.0 to 12.14.0h
```
wmclaugh, line 120:
```suggestion
    # The bvvnei model causes a crash with XSPEC 12.14.0 to 12.14.0h
```

wmclaugh: (same typo in existing code in ~line 110)

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -117,6 +117,15 @@ def get_xspec_models():
...
+    # (it should be fixed in 12.4.0i and later). The model is not
```
wmclaugh, line 121:
```suggestion
    # (it should be fixed in 12.14.0i and later). The model is not
```

wmclaugh (approved):

PR #2055, opened 2024-05-30
Bump Xspec 12.14.0h tests to 12.14.0i

Marie-Terrell:
Summary
========
Bumps xspec 12.14.0h to 12.14.0i for testing.

Details
=====
Bumps xspec 12.14.0h to 12.14.0i for testing due to issues noted with XSbvvnei in #2052.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -30,7 +30,7 @@ jobs:
             fits: astropy
             test-data: submodule
             matplotlib-version: 3
-            xspec-version: 12.14.0h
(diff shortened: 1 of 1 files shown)


DougBurke:
So, it looks like the test failures are just intermission install related problems and not due to the actual xspec change here. Is it worth just trying to re-run the failed tests?

Marie-Terrell:
> So, it looks like the test failures are just intermission install related problems and not due to the actual xspec change here. Is it worth just trying to re-run the failed tests?

Yeah. I tried restarting a few times. It looks like the package was malformed. I've uploaded a corrected version and it looks good now.

Issue #2026, opened 2024-05-06
Switch numpy.matrix to numpy.ndarray


hamogu:
In a few places (e.g. https://github.com/sherpa/sherpa/pull/2025#discussion_r1590433818) Sherpa still uses numpy's old "matrix" class, which is almost, but not entirely identical to the normal ndarray. `matrix` is no longer recommended and may be removed from numpy in the future (https://numpy.org/doc/stable/reference/generated/numpy.matrix.html#numpy.matrix). No deprecation date has been announced, but in the long term we should switch from matrix to ndarray.

PR #2058, opened 2024-06-07
Address an error introduced in PR 2025

DougBurke:
# Summary

Fix an error made when addressing NumPy 2.0 changes in issue #2025.

# Details

We first add some minimal tests of the multit_pdf and multinormal_pdf routines (we could perhaps remove these, but for now I leave them in as they might be useful and we don't want to break anyone that uses Sherpa if we can help it). These tests use values calculated using CIAO 4.16 on Linux x86_64. That is, they are not "from first principles" tests, because I did not wite the code or design the interface or have time to investigate too deeply.

We then fix the error added in #2025 - there's two errors

- the routine was using a variable called `np` but we now `import numpy as np`, which breaks things -> the variable is now called `npr`
- to address issues over the matrix call we had switched to `np.asmatrix` but there was a typo in one of the calls and I'd said `np.asmattix`

After fixing things I then go and address issue #2026 where we noted that code now complains about the use of NumPy matrices because we should be using arrays. Fortunately the fixes here are simple as all we need to is make sure to use `@` to indicate matrix multiplication (this syntax was not available when the code here was written). I note that this only addresses the matrix code in these two routines, not in all of Sherpa, which is why we ar enot closing the issue.

### DIFF
--- a/sherpa/utils/tests/test_utils_unit.py
+++ b/sherpa/utils/tests/test_utils_unit.py
@@ -603,12 +603,7 @@ def test_multinormal_pdf_invalid_input(x, mu, sigma, eclass, emsg):
 def test_multinormal_pdf(x, mu, sigma, expected):
     """Very basic regression test"""
 
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore",
-                                message="the matrix subclass",
-                                category=PendingDeprecationWarning)
-        out = multinormal_pdf(x, mu, sigma)
-
+    out = multinormal_pdf(x, mu, sigma)
     assert out == pytest.approx(expected)
 
 
@@ -648,10 +643,5 @@ def test_multit_pdf_invalid_input(x, mu, sigma, eclass, emsg):
 def test_multit_pdf(x, mu, sigma, dof, expected):
     """Very basic regression test"""
 
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore",
-                                message="the matrix subclass",
-                                category=PendingDeprecationWarning)
-        out = multit_pdf(x, mu, sigma, dof)
-
+    out = multit_pdf(x, mu, sigma, dof)
     assert out == pytest.approx(expected)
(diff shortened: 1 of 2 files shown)


sherpa/utils/tests/test_utils_unit.py
```diff
@@ -566,3 +569,89 @@ def test_bool_cast_invalid(arg, badval):
...
+                           0.03157178495439245),
+                          ([2, 1], [0.5, 0.7], [[1, 0], [0, 1]], 2,
+                           0.033798751957335116),
+                          ([1, 2], [0.5, 0.7], [[1, 0], [0, 1]], 2,
+                           0.04100980264678175)])
+def test_multit_pdf(x, mu, sigma, dof, expected):
+    """Very basic regression test"""
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore",
```
hamogu, line 652: I don't see any more matrix subclass in that function, so this warnings filter should not be necessary any more, right?

sherpa/utils/tests/test_utils_unit.py
```diff
@@ -566,3 +569,89 @@ def test_bool_cast_invalid(arg, badval):
...
+                          ([1, 2], [0.5, 0.7], [[1, 0], [0, 1]],
+                           0.06033293935644923)])
+def test_multinormal_pdf(x, mu, sigma, expected):
+    """Very basic regression test"""
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore",
+                                message="the matrix subclass",
+                                category=PendingDeprecationWarning)
+        out = multinormal_pdf(x, mu, sigma)
```
hamogu, line 610: You removed the matrix use here, so this filter should not be necessary.

hamogu (changes requested):

sherpa/utils/tests/test_utils_unit.py
DougBurke: well, only possible >= 53f678611916fe2961fc766802bb6d8925100e9a ....

DougBurke: I've removed the checks. Ta for the review.

hamogu (approved):

PR #2059, opened 2024-06-10
Copy over typing annotations to the sherpa.ui and sherpa.astro.ui routines

### LINKED ISSUE #2053: ui layer does not get sent any typing annotations
Due to the way we "create" sherpa.ui/sherpa.astro.ui, we currently drop the typing annotations from the symbols.

It's probably `sherpa.utils.export_method` but I haven't looked into it too deeply.

DougBurke:
# Summary

Ensure that annotations are added to the sherpa.ui and sherpa.astro.ui routines. Fix #2053

# Details

Fix #2053 by explicitly copying over the `__annotations__` field from the method to the output symbol.

I have made a few changes to the `sherpa.ui.utils` module so we can see if this works - e.g.

- `get_default_id`: [4.16.1](https://sherpa.readthedocs.io/en/4.16.1/ui/api/sherpa.ui.get_default_id.html) 
- `set_default_id`: [4.16.1](https://sherpa.readthedocs.io/en/4.16.1/ui/api/sherpa.ui.set_default_id.html)
- `list_data_ids`: [4.16.1](https://sherpa.readthedocs.io/en/4.16.1/ui/api/sherpa.ui.list_data_ids.html)
- `clean`: [4.16.1](https://sherpa.readthedocs.io/en/4.16.1/ui/api/sherpa.ui.clean.html)

# Notes

I have added some documentation to `export_method` since this is an old routine which has seen some changes and I was wondering if we could use "newer" features to avoid so much low-level code (in other words, can we not just use `functools.wraps`?). The answer is no.

I have added a question to the in-line documentation regarding the existing use of `FunctionType`, but I am not sure it is worth changing this code too much.


hamogu (approved):
I'm looking forward to the day when Sphinx will parse the type annotations and use that to automatically renter the types in the docs, but that's not happening any time soon.

PR #2060, opened 2024-06-12
Simple typing support for sherpa.data

DougBurke:
# Summary

Add some basic typing rules to the sherpa.data module.

# Details

I actually want to be working on #2022 but to do that I want to clean up the code and add typing statements where possible. This is unfortunately a large job so I'm trying to break things up to be easier to review and less invasive.

I've got various versions of this in different PRs so I've decided to create a sherpa.utils.types module so we can provide a common vocabulary where possible. This is intended as an internal module - i.e. should not be relied-upon by external code - since

1. our understanding of what useful types are will change
2. as the typing ecosystem "matures" in Python things will change (e.g. TypeAlias in Python 3.10 and then this being subsumed by TypeAliasType/type in 3.12)

I will say that although this is annoying to do - since some code which is "obvious" is hard to type well, and so we just don't at the moment [*] - it does help point out some corner cases (e.g. what should the data objects do when a routine is called but the independent or dependent axis has not been set).

[*] it also depends on what code you use to do the type checking

### DIFF
--- a/sherpa/astro/tests/test_astro_data.py
+++ b/sherpa/astro/tests/test_astro_data.py
@@ -28,7 +28,7 @@ import pytest
 from sherpa.astro.ui.utils import Session
 from sherpa.astro.data import Data1D, DataARF, DataPHA, DataRMF, DataIMG, DataIMGInt
 from sherpa.astro.instrument import create_arf, create_delta_rmf
-from sherpa.models.basic import Gauss2D
+from sherpa.models.basic import Gauss1D, Gauss2D
 from sherpa.utils import parse_expr
 from sherpa.utils.err import ArgumentTypeErr, DataErr
 from sherpa.utils.numeric_types import SherpaFloat
@@ -36,10 +36,10 @@ from sherpa.utils.testing import requires_data, requires_fits, \
     requires_group, requires_region
 
 
-EMPTY_DATA_OBJECTS = [(DataPHA, [None] * 2),
-                      (DataIMG, [None] * 3),
-                      (DataIMGInt, [None] * 5)]
-
+EMPTY_DATA_OBJECTS_1D = [(DataPHA, [None] * 2)]
+EMPTY_DATA_OBJECTS_2D = [(DataIMG, [None] * 3),
+                         (DataIMGInt, [None] * 5)]
+EMPTY_DATA_OBJECTS = EMPTY_DATA_OBJECTS_1D + EMPTY_DATA_OBJECTS_2D
 
 def _monotonic_warning(response_type, filename):
     return UserWarning("The {} '{}' has a non-monotonic ENERG_LO array".format(response_type, filename))
@@ -3205,7 +3205,7 @@ def test_data_can_not_set_dep_to_scalar_when_empty(data_class, args):
         data.set_dep(2)
 
 
-@pytest.mark.parametrize("data_class,args", EMPTY_DATA_OBJECTS[1:])
+@pytest.mark.parametrize("data_class,args", EMPTY_DATA_OBJECTS_2D)
 @pytest.mark.parametrize("index", ["x0", "x1"])
 def test_data_empty_get_x_2d(data_class, args, index):
     """What happens when there's no data?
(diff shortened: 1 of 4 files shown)


sherpa/data.py
```diff
@@ -1080,7 +1124,9 @@ def indep(self, val):
         self._data_space = self._init_data_space(self._data_space.filter, *val)
         self._clear_filter()
 
-    def get_indep(self, filter=False):
+    def get_indep(self,
+                  filter: bool = False
+                  ) -> Optional[Union[tuple[np.ndarray], tuple[None]]]:
```
DougBurke, line 1129: Hmmm, can it be a `None` or is it only `tuple[None]`, in which case the `Optional` is not needed.

sherpa/utils/types.py
```diff
@@ -0,0 +1,55 @@
...
+StatErrFunc = Callable[..., ArrayType]
+
+# Represent model evaluation. Using a Protocol may be better, but
+# for now keep with a Callable. Ideally the model would just
+# return an ndarray but
+#
+# - an ArithmeticConstantModel can return a scalar
+# - models could return a sequence rather than a ndarray
+#
+ModelFunc = Callable[..., ArrayType]
```
DougBurke, line 55: Untortunately this should also allow returning a scalar value - *or* we change `ArithmeticConstantModel` to return an ndarray (we know the size to return). I shuld have a version where the return value includes a `float`, so need to check on why that isn't here.

sherpa/data.py
DougBurke: This was from an earlier version. The code now better types the optional behaviour (still not perfect, but better than it was).

DougBurke:
Note that the failure of the TESTS build was just when building the docs, and that was an internal error (for some reason it didn't like the rebase) and nothing to do with the code.

sherpa/data.py
```diff
@@ -2144,31 +2367,34 @@ def get_x0label(self):
...
         return 'x0'
 
-    def get_x1label(self):
+    def get_x1label(self) -> str:
         """Return label for second dimension in 2-D view of independent axis/axes.
         """
         return 'x1'
 
-    def get_axes(self):
+    def get_axes(self) -> tuple[int, int]:
```
DougBurke, line 2375: Oops - this is wrong

DougBurke: Fixed in a rebase.

sherpa/utils/types.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Some of these may change to TypeAlias (once Python 3.10 is the
+# minimum) and then type/TypeAliasType once Python 3.12 is the
+# minimum.
+#
+
+# Try to be generic when using arrays as input or output. There is no
+# attempt to encode the data type or shape for ndarrays at this time.
+#
+ArrayType = Union[Sequence[float], np.ndarray]
```
hamogu, line 42: Also allow sequence of int here (I'm thinking counts in a PHA) or do you want that to be a separate type `ArrayIntType`?

sherpa/utils/types.py
```diff
@@ -0,0 +1,55 @@
...
+# Represent statistic evaluation, per-bin.
+#
+StatErrFunc = Callable[..., ArrayType]
+
+# Represent model evaluation. Using a Protocol may be better, but
+# for now keep with a Callable. Ideally the model would just
+# return an ndarray but
+#
+# - an ArithmeticConstantModel can return a scalar
+# - models could return a sequence rather than a ndarray
```
hamogu, line 53: Not in this PR, but it seems like something we should standartize to always be a numpy array. Or maybe now, now that numpy as the "array API" and other new-fangled stuff like xarray or dask might drop in for numpy arrays in many cases.

sherpa/data.py
```diff
@@ -1048,25 +1106,16 @@ def indep(self):
...
-        # subclass.
-        #
-        pass
-
     @indep.setter
-    def indep(self, val):
+    def indep(self,
+              val: Union[tuple[ArrayType, ...],
+                         tuple[None, ...]]
+              ) -> None:
```
hamogu, line 1119: I know this is beyond the scope if the current PR, but just for my curiosity: Wha'ts the use case for setting indep to None? If I have no independent axis, I don't have any data?

sherpa/data.py
```diff
@@ -1144,18 +1203,36 @@ def set_dep(self, val):
...
+    @overload
+    def get_y(self,
+              filter: bool,
+              yfunc: None,
+              use_evaluation_space: bool = False
+              ) -> np.ndarray:
+        ...
+
+    @overload
+    def get_y(self,
```
hamogu, line 1229: Again, beyond the scope of this PR, but if we designed this form scratch, we might make it two different functions instead of one functions with different signatures...

sherpa/data.py
```diff
@@ -1233,8 +1313,7 @@ def get_staterror(self, filter=False, staterrfunc=None):
 
         """
         staterror = getattr(self, 'staterror', None)
-        filter = bool_cast(filter)
-        if filter:
+        if bool_cast(filter):
```
hamogu, line 1316: Good change. Using a recerved work as a variable is rarely a good idea... this is much cleaner.

sherpa/data.py
```diff
@@ -1344,43 +1433,60 @@ def apply_filter(self, data):
...
 
-    def ignore(self, *args, **kwargs):
+    def ignore(self, *args, **kwargs) -> None:
         kwargs['ignore'] = True
         self.notice(*args, **kwargs)
 
-    def eval_model(self, modelfunc):
-        """Evaluate the model on the independent axis."""
+    def _can_apply_model(self, modelfunc: ModelFunc) -> None:
+        """Check we model dimensions match."""
```
hamogu, line 1449:
```suggestion
        """Check if model dimensions match data dimensions."""
```

sherpa/data.py
```diff
@@ -1580,10 +1717,30 @@ def get_xlabel(self):
...
+              use_evaluation_space: bool = False
+              ) -> np.ndarray:
+        ...
+
+    @overload
+    def get_y(self,
+              filter: bool,
+              yfunc: ModelFunc,
+              use_evaluation_space: bool = False
+              ) -> tuple[np.ndarray, ArrayType]:
```
hamogu, line 1739: I suggest to also update the docstring to mention the second signature

hamogu (changes requested):
A few comments, mostly spelling or non-actionable comments. The explanation for this PR  ppoints out a lot of corner cases when e.g .the independent axis is set to `None`. It's **beyond the scope of this PR to fix**, but sometimes an overly complext code indicates that the design is not clean. So, let's think about this for a minute: What is the use for for the dependent or independent axis to be `None`? 
Can we just disallow rthat and say "no axis, no data, and thus no object? We would require it to be set in `__init__` and simply remove the possibility to set it to `None` at a later stage. If we could do that, that would simply not only the typing, but also the code, and reduce the risk of coner case failures or bugs.

sherpa/utils/types.py
DougBurke: There's a couple of issues here

- technically `int` is subsumed by `float` so it's not obvious we'd gain anything by including `int` here
  - note that the numeric typing tower is surprising complex, even when you don't need to worry about IEEE 757/NaN issues, as discussed in many places, such as the recent typing discourse item: https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018
- for PHA data, we can not say "must be an int" for the counts array (dependent axis), since we have to deal with the possibility the data may be a rate rather than a count
  - although we could enforce it for the independent axis, but there's "interesting" things to think about with that as this then changes the interface the base class promises, which can lead to worries over things like the [Liskov Sustitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), although to be fair our existing class structure has many such places like this 

I wanted to say that this change - in particular `ArrayType`, but it holds for any of our initial typing statements - doesn't magically fix up our code. It just points to other areas that we'll have to work on in later PRs, by tweaking up the tying statements, or with code changes, or both. For example, we have many places where we send in something like `Union[float, ArryaType]` and use something like `np.iterable` to check to see whether we have an array - for example this made-up example

```
from typing import Sequence, Union

import numpy as np

ArrayType = Union[Sequence[float], np.ndarray]

def bar(xs: Union[float, ArrayType]) -> int:
    if np.iterable(xs):
        return len(xs)

    return 1
```

which current `mypy` (1.10.0) and `pyright` (1.1.368) both complain about

```
x.py:12: error: Argument 1 to "len" has incompatible type "Iterable[Any]"; expected "Sized"  [arg-type]
```

and

```
  /home/dburke/sherpa/sherpa-xspec/x.py:12:20 - error: Argument of type "Iterable[Any]" cannot be assigned to parameter "obj" of type "Sized" in function "len"
    "Iterable[Any]" is incompatible with protocol "Sized"
      "__len__" is not present (reportArgumentType)
```

sherpa/data.py
DougBurke: So, I am basically trying to type the existing code, and it allows for this. I personally would like to remove the "can be `None`" ability for the independent and dependent axes completely, but this is used in the existing fake-PHA data code, and there are traces throughout the code for allowing `None` values to be returned. So, follow-up PRs could be done to simplify this - for instance, at the moment you can do

```
>>> d = Data1D('x', None, None)
>>> d.dep = None
>>> d.indep = (None, )
>>> d.dep = [1, 2, 3]
>>> d.dep = None
>>> d.dep = [1, 2, 3, 4]
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/utils/__init__.py", line 137, in __setattr__
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/data.py", line 1030, in dep
    self.y = val
    ^^^^^^
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/utils/__init__.py", line 137, in __setattr__
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/data.py", line 997, in y
    self._set_related("y", val, check_mask=False)
  File "/home/dburke/sherpa/sherpa-xspec/sherpa/data.py", line 983, in _set_related
    raise DataErr('mismatchn', 'independent axis', attr, nelem, nval)
sherpa.utils.err.DataErr: size mismatch between independent axis and y: 3 vs 4
>>> d.indep = ([45, 99, 203], )
```

and I could see us removing the ability to set either `indep` or `dep` to `None` (i.e. only allowing the `DatanD(name, None, ..., None)` call). However, there's always the compatibility concern, but that's for future-PR writer to worry about!

sherpa/data.py
DougBurke: However, it doesn't really help, since `filter` is a keyword argument to this (and many other) routines which we are not going to change, so we still have the "using a reserved work" error from tools like `pylint` for the function definition.

sherpa/data.py
hamogu: Thanks for the explanation.

sherpa/utils/types.py
hamogu: Thanks for the explanation. I always learn new things when I review your PRs!

hamogu (approved):
My questions are all answered, so it's only typos in docstrings that are inconsequential an will probably be changed in the next PR anyway, so I'm approving.

sherpa/data.py
DougBurke: Done

sherpa/astro/tests/test_astro_data.py
```diff
@@ -3814,3 +3814,189 @@ def test_pha_group_xxx_with_background(caplog):
...
+                       match=r"object of type 'NoneType' has no len\(\)"):
+        _ = data.get_filter()
+
+
+@pytest.mark.parametrize("data_class,args", EMPTY_DATA_OBJECTS_2D)
+def test_get_filter_when_empty_2d(data_class, args):
+    """This is a regression test.
+
+    Although Data2D hard-codes get_filter, the DataIMG class does
+    change the result (techncally only if region support is availablem
```
wmclaugh, line 3996:
```suggestion
    change the result (technically only if region support is available
```

sherpa/data.py
```diff
@@ -1030,11 +1072,27 @@ def get_dims(self):
...
         return indep_size, self.dep.size
 
+    def _clear_filter(self) -> None:
+        """Clear out the existing filter.
+
+        This is designed for use by @indep.setter.
+
+        """
+
+        # This is currently a no-op. It may be over-ridden by a
```
wmclaugh, line 1087:
```suggestion
        # This is currently a no-op. It may be overridden by a
```

sherpa/utils/types.py
```diff
@@ -0,0 +1,55 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Useful types for Sherpa.
+
+This module should be considered an internal module as its contents is
```
wmclaugh, line 23:
```suggestion
This module should be considered an internal module as its contents are
```

Issue #1939, opened 2023-12-15
planned I/O changes for 4.17


DougBurke:
I have a number of planned changes to the I/O code and wanted to provide something to tie things together. We have

- new code added in 4.16 (e.g. `save_arf` and `save_rmf` along with the experimental "write out an XSPEC table model" routines)
- a lot of technical debt, since the I/O backends have not been updated to reflect changes in both crates and astropy
- a number of annoying bugs where the two backends don't behave the same
  - **note** we can not guarantee bit-for-bit output from the two backends, in fact I claim that we can try to keep the data "semantically" the same but we are to always going to get things like the data formats exactly the same, but we want to better understand when they can be the same and when they can't be, and that's tricky to identify with the existing code
  - we occasionally hit the "backend a can read in this RMF file but backend b can not" and sometimes it's because of an issue with the backend, but not always
- there has been work to define the I/O layer (i.e. that a backend would need to implement) but it is not as clear as it could be 
  - we could follow the plotting approach and move to a class-based approach but I honestly don't see the benefit here compared to the existing module-based approach (since we are unlikely to want to subclass a particular backend)
 
I had been working on this for the 4.16 release, in particular working on supporting "multi matrix" RMF files, but decided that it wasn't going to be possible to get in on time. We now have time. There are two primary changes

1. take advantage of the classes introduced in 4.16 to better represent FITS-like data to send data between the backend code and the generic IO code
2. rewrite the code so that the conversion from FITS-like concepts to the Sherpa data objects (on input) and the reverse (on output) is done in the generic layer, with the backend code just translating between the on-disk format and the new FITS-like classes

In doing this we can pay down some of the technical debt since we are re-working the backend code. So the backend code is going to change (generally that the arguments for the existing calls will change) but the generic interface is not going to see any significant change (hopefully minimal changes), since this is the AI that users can expect to be using.

Code changes are, in order

- #1926 
  - this is a combination of "finally tell users that load_table_model should not be used with XSPEC models" (I'd like to just ban it now but people haven't got the message so I am not sure we want to do this) along with some minor code clean up and test coverage improvements based on the follow-up PRs
- ~~#1929~~ 
  - I have never liked the "packup" argument for the I/O ocde and if you look at the implementation you can see it is like shoe-horning two different routines into one, so this breaks this (this one does actualy change the I/O API a bit, unlike my promise from earlier)
[...]

PR #2061, opened 2024-06-14
Avoid warning messages from AstroPy about unclosed files

DougBurke:
# Summary

Initial cleanup of the astro IO code in preparation for future changes. This fixes an issue when using the AstroPy IO backend (pyfits) which could cause Python to report an unclosed file on exit.

# Details

This is taken from #2017 which was itself taken from #1929. The idea is to reduce the size of that PR to make it easier to review. This PR concentrates on initial clean up of the I/O code, addressing one or two minor improvements, but basically setting the stage for later changes (hence I'd rather not spend time on improving the changes here as that may be wasted given the other changes in #2017, or can be better done after the later changes are made).

This is part of #1939.

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -54,18 +54,16 @@ import importlib
 import logging
 import os
 import re
-from typing import Any, Optional
+from typing import Any, Optional, Sequence
 
 import numpy as np
 
 from sherpa import get_config
 from sherpa.astro.data import DataIMG, DataIMGInt, DataARF, DataRMF, \
     DataPHA, DataRosatRMF
-# leads to circular imports
-# from sherpa.astro.instrument import ARF1D, RMF1D
 from sherpa.astro.utils import reshape_2d_arrays
 from sherpa.data import Data, Data1D, Data2D, Data2DInt
-import sherpa.io
+from sherpa.io import _check_args
 from sherpa.utils.err import ArgumentErr, DataErr, IOErr
 from sherpa.utils.numeric_types import SherpaFloat
 
@@ -172,7 +170,7 @@ def read_arrays(*args):
     dargs = backend.get_column_data(*largs)
 
     # Determine max number of args for dataset constructor
-    sherpa.io._check_args(len(dargs), dstype)
+    _check_args(len(dargs), dstype)
 
     return dstype('', *dargs)
 
@@ -232,7 +230,7 @@ def read_table(arg, ncols=2, colkeys=None, dstype=Data1D):
     name = args[2]
 
     # Determine max number of args for dataset constructor
-    sherpa.io._check_args(len(cols), dstype)
+    _check_args(len(cols), dstype)
 
     return dstype(name, *cols)
 
@@ -293,7 +291,7 @@ def read_ascii(filename, ncols=2, colkeys=None, dstype=Data1D, **kwargs):
     name = args[2]
 
     # Determine max number of args for dataset constructor
-    sherpa.io._check_args(len(cols), dstype)
+    _check_args(len(cols), dstype)
 
     return dstype(name, *cols)
 
@@ -369,8 +367,11 @@ def read_image(arg, coord='logical', dstype=DataIMG):
     #
     if issubclass(dstype, (Data2DInt, DataIMGInt)):
         indep = [x0 - 0.5, x1 - 0.5, x0 + 0.5, x1 + 0.5]
-    else:
+    elif issubclass(dstype, Data2D):
         indep = [x0, x1]
+    else:
+        raise ArgumentErr("bad", "dstype argument",
+                          "dstype is not derived from Data2D")
 
     # Note that we only set the coordinates after creating the
     # dataset, which assumes that data['x'] and data['y'] are always
@@ -382,8 +383,7 @@ def read_image(arg, coord='logical', dstype=DataIMG):
     # issue #1380).
     #
     dataset = dstype(filename, *indep, **data)
-
-    if issubclass(dstype, DataIMG):
+    if isinstance(dataset, DataIMG):
         dataset.set_coord(coord)
 
     return dataset
@@ -633,6 +633,7 @@ def read_pha(arg, use_errors=False, use_background=False):
             if bkg.grouping is None:
                 bkg.grouping = pha.grouping
                 bkg.grouped = bkg.grouping is not None
+
             if bkg.quality is None:
                 bkg.quality = pha.quality
 
@@ -1153,7 +1154,25 @@ def _pack_arf(dataset):
     return data, header
 
 
-def _make_int_array(rows, ncols):
+def _find_int_dtype(rows: Sequence[Sequence[int]]) -> type:
+    """What data type should represent the matrix of integers?
+
+    There is no guarantee that each row has the same number of
+    elements.
+
+    """
+
+    maxval = 0
+    for row in rows:
+        if len(row) == 0:
+            continue
+
+        maxval = max(maxval, np.max(row))
+
+    return np.int32 if maxval > 32767 else np.int16
+
+
+def _make_int_array(rows, ncols) -> np.ndarray:
     """Convert a list of rows into a 2D array of "width" ncols.
 
     The conversion is to a type determined by the maximum value in
@@ -1174,14 +1193,7 @@ def _make_int_array(rows, ncols):
     """
 
     nrows = len(rows)
-    maxval = 0
-    for row in rows:
-        if len(row) == 0:
-            continue
-
-        maxval = max(maxval, row.max())
-
-    dtype = np.int32 if maxval > 32767 else np.int16
+    dtype = _find_int_dtype(rows)
     out = np.zeros((nrows, ncols), dtype=dtype)
     for idx, row in enumerate(rows):
         if len(row) == 0:
@@ -1192,7 +1204,7 @@ def _make_int_array(rows, ncols):
     return out
 
 
-def _make_float32_array(rows, ncols):
+def _make_float32_array(rows, ncols) -> np.ndarray:
     """Convert a list of rows into a 2D array of "width" ncols.
 
     The output has type numpy.float32.
@@ -1219,7 +1231,7 @@ def _make_float32_array(rows, ncols):
     return out
 
 
-def _make_int_vlf(rows):
+def _make_int_vlf(rows) -> np.ndarray:
     """Convert a list of rows into a VLF.
 
     The conversion is to a type determined by the maximum value in
@@ -1236,14 +1248,7 @@ def _make_int_vlf(rows):
 
     """
 
-    maxval = 0
-    for row in rows:
-        if len(row) == 0:
-            continue
-
-        maxval = max(maxval, row.max())
-
-    dtype = np.int32 if maxval > 32767 else np.int16
+    dtype = _find_int_dtype(rows)
     out = []
     for row in rows:
         out.append(np.asarray(row, dtype=dtype))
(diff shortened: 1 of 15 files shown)


DougBurke:
Rebased to pick up #2065 and add another set of tests and clean-up changes.

DougBurke:
@hamogu - you should have seen most of these changes when reviewing #1929 (although this is much smaller than that PR).

hamogu:
Yeah, but that was months ago and I don't remember, so I have to look again.

sherpa/tests/test_sherpa_io.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Although ncols=1 is the default be explicit in this test
+    (colnames, coldata, fname) = get_ascii_data(str(tfile), ncols=1)
+    assert colnames == ["col2", "xcol", "ycol"]
+    assert len(coldata) == 1
+    assert coldata[0] == pytest.approx([-0.2, -0.4])
+
+
+def test_get_ascii_data_more_colnames_than_columns_no_colkeys_ncols2(tmp_path):
+    """What happens if more column names than columns: ncols=2?"""
```
hamogu, line 206: This behavior seems inconsistent with what happens if there are more columns than colnames, but I'm assuming it's not worth discussing because i will be handled in a later PR?

sherpa/tests/test_sherpa_io.py
```diff
@@ -0,0 +1,304 @@
...
+    """What happens if do not give a list of lists?
+
+    Should this error out? Users may be relying on this behavior, so
+    we do not error out, but that seems unlikely.
+
+    """
+
+    tfile = tmp_path / "not-changed.dat"
+    with pytest.raises(IOErr,
+                       match=r"^please supply array\(s\) to write to file$"):
```
hamogu, line 281: I like how the error message politely says "please" and not just "XYZ IS REQUIRED, YOU FOOL"

sherpa/astro/io/__init__.py
```diff
@@ -1147,7 +1155,25 @@ def _pack_arf(dataset):
...
+
+    maxval = 0
+    for row in rows:
+        if len(row) == 0:
+            continue
+
+        maxval = max(maxval, np.max(row))
+
+    return np.int32 if maxval > 32767 else np.int16
+
```
hamogu, line 1174: I know this is not in the critical path, but why not return early?
```suggestion
    for row in rows:
        if len(row) == 0:
            continue
        if np.max(row) > 32767:
            return np.int32
    return np.int16

```

sherpa/astro/io/__init__.py
```diff
@@ -1206,14 +1225,14 @@ def _make_float32_array(rows, ncols):
     """
 
     nrows = len(rows)
-    out = numpy.zeros((nrows, ncols), dtype=numpy.float32)
+    out = np.zeros((nrows, ncols), dtype=np.float32)
     for idx, row in enumerate(rows):
         out[idx, 0:len(row)] = row
 
     return out
```
hamogu, line 1232: This is fine, but for my own education, I looked around for other implementations of the same problem and came upon this:
```Python
import itertools
return np.array(list(itertools.zip_longest(*v, fillvalue=0))).T
```
(from https://stackoverflow.com/questions/38619143/convert-python-sequence-to-numpy-array-filling-missing-values)

I like the existing Sherpa code better (it's easier to read), but I thought I'll just comment on what I learned - I probably don't look at `itertools` often enough.

sherpa/astro/io/crates_backend.py
```diff
@@ -98,28 +105,36 @@ def get_filename_from_dmsyntax(filename, blockname=None):
...
+
+        # Crates only guarantees to support ASCII-like variants but it
+        # should be okay to use a less-restrictive encoding here.
+        #
+        with open(filename, mode='r', encoding="UTF8") as fh:
             last = None
             line = fh.readline().strip()
-            while len(line) > 0 and line[0] in '#%':
+            while len(line) > 0 and line.startswith("#"):
                 last = line
```
hamogu, line 117: This is change in behavior (`%` is no longer allowed). Do we need to document that?

sherpa/astro/io/crates_backend.py
```diff
@@ -285,121 +286,152 @@ def _require_tbl_col(crate, colname, cnames, make_copy=False,
...
 
     if make_copy:
         # Make a copy if a filename passed in
-        dat = pycrates.copy_piximgvals(crate).squeeze()
+        data = np.array(values).ravel()
+
     else:
         # Use a reference if a crate passed in
-        dat = pycrates.get_piximgvals(crate).squeeze()
+        data = np.asarray(values).ravel()
```
hamogu, line 393:
```suggestion
    data = np.array(values, copy=make_copy).ravel()
```

sherpa/astro/io/crates_backend.py
```diff
@@ -525,18 +547,18 @@ def get_column_data(*args):
 
     cols = []
     for arg in args:
-        if isinstance(arg, CrateData):
+        if isinstance(arg, pycrates.CrateData):
             # vals = arg.get_values()
             vals = arg.values
 
-        elif arg is None or isinstance(arg, (numpy.ndarray, list, tuple)):
+        elif arg is None or isinstance(arg, (np.ndarray, list, tuple)):
```
hamogu, line 554: This almost, but not exactly looks like

```suggestion
        elif arg is None or np.isiterable(arg):
```
I suspect that is what this should be to be general.

sherpa/astro/io/crates_backend.py
```diff
@@ -1220,37 +1264,31 @@ def set_image_data(filename, data, header, ascii=False, clobber=False,
...
 
-        tbl.write(filename, clobber=True)
+    write_dataset(tbl, filename, ascii=ascii)
+    return None
 
 
 def set_arf_data(filename, data, col_names, header=None,
-                 ascii=False, clobber=False, packup=False):
+                 ascii=False, clobber=False, packup=False) -> Optional[TABLECrate]:
     """Create an ARF"""
```
hamogu, line 1292: Why is this optional? In case we pass in an empty dataset?

sherpa/astro/io/crates_backend.py
```diff
@@ -1409,112 +1430,78 @@ def mkcol(name, vals, units=None):
...
-
-    if not numpy.iterable(args[0]):
-        raise IOErr('noarrayswrite')
+    # Check args is a sequence of sequences (although not a complete
+    # check).
+    #
+    try:
+        size = len(args[0])
+    except (TypeError, IndexError) as exc:
+        raise IOErr('noarrayswrite') from exc
```
hamogu, line 1448: An alternative would be:
```suggestion
    if not isinstance(args[0], (collections.abs.Sequence, np.ndarray))
        raise IOErr('noarrayswrite') from exc

```
It's slightly different in that is checks the class hirachy, not duck-typing, but in practice, it's probably the same. I don't think either is "better", but my version is shorter ;-)

Same applies for the lines below.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -119,43 +118,22 @@ def _get_meta_data(hdu):
...
     """Add the name,val pair to hdulist."""
-    name = str(name).upper()
-    if name in ['', 'COMMENT', 'HISTORY']:
-        # The values are assumed to be an array of strings
-        for v in val:
-            hdrlist.append(fits.Card(name, v))
 
-    else:
-        hdrlist.append(fits.Card(name, val))
+    # This used to special case name = { "", "COMMENT", "HEADER" } as
```
hamogu, line 131:
```suggestion
    # This was used to special case name = { "", "COMMENT", "HEADER" } as
```
Just to make super clear that this is a comment on past code.
Or, just remove.

sherpa/astro/io/pyfits_backend.py
```diff
@@ -324,90 +291,89 @@ def open_fits(filename):
...
+                       blockname: Optional[str] = None) -> fits.BinTableHDU:
     """Return the first binary table extension we find. If blockname
     is not None then the name of the block has to match (case-insensitive
     match), and any spaces are removed from blockname before checking.
 
     Throws an exception if there aren't any.
     """
 
+    # This behaviour appears odd:
+    # - if blockname is no set then find the first table block
```
hamogu, line 363:
```suggestion
    # - if blockname is not set then find the first table block
```

sherpa/astro/io/tests/test_io.py
```diff
@@ -416,6 +416,36 @@ def test_fits_file_missing_column(make_data_path):
...
 
 
+@requires_fits
+@requires_data
+def test_get_header_data_missing_key(make_data_path):
+    """What happens if a requested key is missing?
+
+    TODO: If get_header_data is useful should we export if from
+    the io, and not backend, level?
+    """
```
hamogu, line 426: Good idea~ I think it would be useful.

hamogu (changes requested):

hamogu (changes requested):
Some comments, some suggestions, some typoes

sherpa/tests/test_sherpa_io.py
DougBurke: We can put that in a later PR....

sherpa/tests/test_sherpa_io.py
DougBurke: Yeah - this is to try and ensure we have tests for the existing behaviour, so if we do change things later on we know about it. Or that's the plan; life has shown that you never catch all issues...

sherpa/astro/io/crates_backend.py
DougBurke: So, this is an interesting one (and I think easier to see in smaller PRs like this than the previous PR). The safest thing is probably just to keep the old behavior, but we have never really discussed '%' support elsewhere and so I thought we could see what happens if we remove it. So we could just document the change. Or try to improve the code.

sherpa/astro/io/crates_backend.py
DougBurke: I really don't like the existing `make_copy` behaviour (we use it less in the pyfits backend). I believe I know where the concern came from, but I wonder if we now still need it. However, if we can just do it with this one line it's less of a concern.

sherpa/astro/io/crates_backend.py
DougBurke: The return value depends on the `packup` argument. So this could be done as an `overload` (ie when packup=False we return None and when True it returns the value), but a later PR changes this code to remove the `packup` option.

sherpa/astro/io/crates_backend.py
DougBurke: Your version doesn't check the case of `args` not being indexable, unless I'm mistaken.

Now, some of my code changes are written to try and improve the typing results, and so some of my thinking could be being swayed by that, so it's good to check on my thinking here.

hamogu: Are there any Sequences that are not indexable? Of course, we could have a multi-D numpy array etc., so it's not a complete check either...

DougBurke: What happens if you send in `arg=None`?

hamogu: Then it will fail, because `None` is no subscriptable. So, let's keep your version. As I said above, I'm not sure testing the hierarchy (as I do) is any better anyway.

sherpa/astro/io/tests/test_io.py
DougBurke: I've made #2071 so we can discuss that there.

sherpa/tests/test_sherpa_io.py
```diff
@@ -0,0 +1,304 @@
...
+    """Regression test"""
+
+    tfile = tmp_path / 'x.dat'
+    with pytest.raises(IOErr,
+                       match=r"^please supply array\(s\) to write to file$"):
+        write_arrays(str(tfile), 23, fields=["x"])
+
+
+def test_get_ascii_data_irregular_data(tmp_path):
+    """What happens when the numbor of columns is not constant?"""
```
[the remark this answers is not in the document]
DougBurke: [also needed in `test_write_arrays_irregular_data`]

sherpa/astro/io/crates_backend.py
DougBurke: I agree that it looks sensible, it's just I don't want to make this change here "just in case". I admit that it is irrational (I have made bigger changes elsewhere in this module)!

sherpa/astro/io/crates_backend.py
hamogu: Agreed; as few notes as we can get away with. I've never seen this used, any maybe nobody ever did.

hamogu (approved):

Issue #1963, opened 2024-02-16
numpy 2.0


hamogu:
Numpy 2.0 will be breaking API and ABI, though they make an effort to have numpy 2.0 compatible code also work on older version of numpy. The general numpy annoucement/tracking issue is here: https://github.com/numpy/numpy/issues/24300 with the following top-level recommendations:

Key guidance for users and downstream package authors

1. If you rely on the NumPy C API (e.g. via direct use in C/C++, or via Cython code that uses NumPy), please add a numpy<2.0 requirement in your package's dependency metadata (for releases only, not on your main dev branch). *Rationale: the NumPy C ABI will change in 2.0, so any compiled extension modules that rely on NumPy are likely to break, they need to be recompiled.*
2. If you rely on a large API surface from NumPy's Python API, also consider adding the same  numpy<2.0 requirement to your metadata. *Rationale: we will do a significant cleanup (see NEP 52), so unless you only use modern/recommended functions and object, your code is likely to require at least some adjustments.*
3. Consider cleaning up your code. E.g. remove from numpy import *, or importing any private modules like numpy.core. See https://github.com/numpy/numpy/blob/main/numpy/tests/test_public_api.py#L114-L126 for what we consider public/private. If it's not in the NumPy docs or in the list of public modules there, don't use it!
4. Plan to do a release of your own packages which depend on numpy shortly after the first NumPy 2.0 release candidate is released (probably in Dec 2023). *Rationale: at that point, you can release packages that will work with both 2.0 and 1.X, and hence your own end users will not be seeing much/any disruption (you want pip install mypacackage to continue working on the day NumPy 2.0 is released).*
5. Consider testing against NumPy nightlies in your own CI. We publish those at https://anaconda.org/scientific-python-nightly-wheels/numpy, and have documented that as a stable location at https://numpy.org/devdocs/dev/depending_on_numpy.html. *Rationale: this will detect potential issues in your code so you can fix them well ahead of the NumPy 2.0 release.*

Given our own constraints, I don't think we'll follow all of that for now, but I think it's worth having this as a Sherpa open issue to track it, even if it might not get into the next release.

Other notes (for releases):
> A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0.dev0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled against NumPy 2.0.

DougBurke:
See also #1864

hamogu:
[...]

PR #2065, opened 2024-06-21
CI: enforce NumPy does not use version 2.0 for the builds

DougBurke:
# Summary

Ensure the CI builds use numpy<2

# Details

At present the Sherpa tests will fail when NumPy 2.0 is selected, mainly because of "not critical" issues (e.g. how numeric values get converted to strings), so enforce a "numpy < 2.0" rule when installing NumPy.

See #1963 for a discussion of the test failures we get with NumPy 2.0.

A number of PRs use this commit.


Marie-Terrell (approved):
Thank you! This looks good to me! I agree with you on not enforcing numpy<2.0 as a requirement yet as this makes it easier to test bumping things up.

PR #2066, opened 2024-06-21
Add basic typing to the model classes WIP

DougBurke:
Based on #2060

- [ ] can we make `ArithmeticConstantModel` return a `ndarray` to simplify the typing?
- [ ] can we enforce the return value of a model is always a ndarray, or should we just require the caller to call `asarray` if needed?


DougBurke:
Rebased as #2060 has been merged.

PR #2068, opened 2024-06-26
CM-887: Updating macos gitlab runner images to version 12 and update the MacOS SDK to version 11.0

harlantc:
This ups the actions macOS Intel version to 12, and updates the macOS SDK to version 11.0, which it consistent with CIAO.

I tested this on my fork, temporarily enabling the conda packaging tests to ensure they worked.

I also added concurrency checks to prevent duplicate actions, as I repeatedly pushed in a short time span, resulting in many duplicate actions being kicked off unnecessarily.

### DIFF
--- a/.github/workflows/ci-conda-deployment.yml
+++ b/.github/workflows/ci-conda-deployment.yml
@@ -125,12 +125,12 @@ jobs:
     runs-on: ${{ matrix.os-type }}
     strategy:
       matrix:
-        os-type: ["macos-11"]
+        os-type: ["macos-12"]
         conda-os: ["MacOSX-x86_64"]
         os-dir: ["osx-64"]
         python-version: ["3.9", "3.10", "3.11"]
     env:
-      CONDA_BUILD_SYSROOT: /opt/MacOSX10.14.sdk
+      CONDA_BUILD_SYSROOT: /opt/MacOSX11.0.sdk
 
     steps:
     - name: Conda Setup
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-conda-deployment.yml
```diff
@@ -142,16 +142,16 @@ jobs:
         source ${conda_loc}/etc/profile.d/conda.sh
         conda create -yq -n builder conda-build conda-verify
 
-    - name: macOS 10.14 SDK
+    - name: macOS 11.0 SDK
```
DougBurke, line 145: One thing I've never understood with this code: we download the macOS SDK in `setup_conda.sh` and also here. Isn't this wasted work? Or am I missing something fundamental, obvious, or both?

Marie-Terrell: So, the setup_conda.sh is only used by the conda test workflow (ci-conda-workflow.yml) the other sdk setup is only in the conda package build workflow (ci-conda-deployment.yml).

This is mostly a product of the old Conda package build being in a separate gitlab mirror (then moved here due to changes with GitLab.com). We needed to add an sdk install step (as we used to already have it installed on the gitlab runner) and pulled what was needed. We couldn't use the setup_conda.sh wholesale due to a lot of the other setup existing in the script.

In the future, we should be able to re-write the setup_conda.sh and move some things around to reduce duplication (as that would be nice)

PR #2069, opened 2024-06-27
Allow the tests to pass with NumPy 2.0

DougBurke:
# Summary

Allow the tests to pass when using NumPy 2.0, as long as pytest-doctestplus is not installed.

# Details

When NumPy 2.0 is installed we have three test failures and this commit addresses two of them

- [x] `test_root.py` failed with `zeroin` for two tests (which turned out to be a duplicate test)
- the representation of NumPy numbers has changed
  - [x] this affects the fit-related structures, which are checked for in our tests
  - [ ] it also affects `pytest-doctestplus` tests, but these are not used in our CI runs 

The root-finding code is not documented. I spent some time in trying to work out why `zeroin` failed for two tests and found that the two tests were actually the same (so one of the commits removes the second test). The issue is that the code relied on "loose" casting rules which were changed in NumPy 2.0

```
>>> import numpy as np
>>> np.__version__
'1.26.4'
>>> type(np.float32(2.0) + 1)
<class 'numpy.float64'>
```

versus

```
>>> import numpy as np
>>> np.__version__
'2.0.0'
>>> type(np.float32(2.0) + 1)
<class 'numpy.float32'>
```

The code made use of a `np.float32` value and previously this has been converted to `float64` but it now remained `float32`, which was a problem because it ended up being used in a statement where 1e120 was used, which is okay for `float64` but caused an overflow with `float32`, resulting in chaos (aka a failing test). The easiest change is to just convert the `float32` to a `float` to avoid this.

There may be other places where we are affected by this subtlety, but our tests don't show any obvious cases!

The second change is in the representation of numbers:

```
>>> import numpy as np
>>> np.__version__
'1.26.4'
>>> p = tuple(np.float64(x) for x in [2, 3])
>>> p
(2.0, 3.0)
>>> 
```

and

```
>>> import numpy as np
>>> np.__version__
'2.0.0'
>>> p = tuple(np.float64(x) for x in [2, 3])
>>> p
(np.float64(2.0), np.float64(3.0))
```

I have found the "obvious" cases - `fit` and `error` results and do an explicit `float(x)` cast, to remove the `numpy-ness` here. However, this is only found in a few test cases. There are plenty of other cases where we see this - as seen in the `doctestplux` output, and so we need to decide what to do (e.g. do we want to try and convert to `float` where we can?), but that is for a later PR. This is to allow NumPy 2 to be used in our CI runs.

For reference, here are some of the types of failures we see with `doctestplus`:

```
2395         >>> (clo == pha.channel).all()
Expected:
    True
Got:
    np.True_
```

```
3000         >>> pha.get_areascal()
Expected:
    1.0
Got:
    np.float64(1.0)
```

```
2959         >>> pha.get_backscal()
Expected:
    2.5264364698914e-06
Got:
    np.float64(2.5264364698914e-06)
```

```
061 The `val` attribute is used to retrieve or change the parameter value:
062 
063     >>> p.val
Expected:
    2.0
Got:
    np.float64(2.0)
```

```
2229         >>> d = Data1DInt('example', xlo, xhi, y)
2230         >>> d.xlo[0], d.xhi[-1]
Expected:
    (0.4, 2.400000000000001)
Got:
    (np.float64(0.4), np.float64(2.400000000000001))
```

```
537     >>> m1 = Gauss1D()
538     >>> m2 = NormGauss1D()
539     >>> m1.pos, m2.pos = 10, 10
540     >>> m1.ampl, m2.ampl = 10, 10
541     >>> m1.fwhm, m2.fwhm = 5, 5
542     >>> m1(10)
Expected:
    10.0
Got:
    np.float64(10.0)
```

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -43,7 +43,7 @@ fi
 # Figure out requested dependencies
 if [ -n "${MATPLOTLIBVER}" ]; then MATPLOTLIB="matplotlib=${MATPLOTLIBVER}"; fi
 if [ -n "${BOKEHVER}" ]; then BOKEH="bokeh=${BOKEHVER}"; fi
-if [ -n "${NUMPYVER}" ]; then NUMPY="numpy=${NUMPYVER}"; else NUMPY="numpy<2.0"; fi
+if [ -n "${NUMPYVER}" ]; then NUMPY="numpy=${NUMPYVER}"; fi
 # Xspec >=12.10.1n Conda package includes wcslib & CCfits and pulls in cfitsio & fftw
 if [ -n "${XSPECVER}" ];
  then export XSPEC="xspec-modelsonly=${XSPECVER}";
--- a/.github/workflows/ci-pip-workflow.yml
+++ b/.github/workflows/ci-pip-workflow.yml
@@ -35,7 +35,7 @@ jobs:
           - name: Linux Build (w/o Xspec; Python 3.11)
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-pkg: 'numpy<2.0'
+            numpy-pkg: 'numpy'
             install-type: install
             test-data: package
             fits-pkg: 'astropy'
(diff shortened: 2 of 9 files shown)


hamogu (approved):
I had a look at this when I wrote https://github.com/cxcsds/ciao-contrib/pull/636, but settled on using the bisection for root finding. So, even if it's not used here, we should think about if it's sherpa's job to supply those routines.

So, yes, fix them here.

PR #2072, opened 2024-07-02
Use the correct case for the symbol name in the parameter warning message

DougBurke:
# Summary

Change how the warning message for deprecated parameter names for models is displayed (only seen when warning messages are turned on).

# Details

This has come out of the "try to add typing to the various bits of Sherpa" work. We have

- an error in `EvaluationSpace1D` where a property was returning the wrong thing
  - as this is an "internal" class, **and** the property is not used much / ever, then it has little downstream consequences
- when catching an exception, name the exception type
- an internal change that just makes the code a bit more obvious what is going on (and provides better information to type checkers)
- change how the parameter name access works, so that we report the actual case used by the user rather than always use the lower-case version, as it is a bit confusing. This is **only** relevant for those parameters that we have "alternative" names - we currently only have these for XSPEC models, but (as shown below) we can show this behavior without needing to have XSPEC installed. Most users will not see this message anyway as the warnings are hidden by default in interactive and script use.

None of these are earth shattering...

# Example

We use the `-Wa` option to always show the warnings. The setup is

```
% python -Wa
Python 3.11.9 (main, Apr 19 2024, 16:48:06) [GCC 11.2.0] on linux
>>> from sherpa.models.model import Model
>>> from sherpa.models.parameter import Parameter
>>> class MyModel(Model):
...     ndim = 1
...     def __init__(self, name='mymodel'):
...         self.foo = Parameter(name, 'foo', 0, aliases=['bar'])
...         super().__init__(name, (self.foo, ))
... 
>>> m = MyModel()
>>> print(m)
mymodel
   mymodel.foo  thawed            0 -3.40282e+38  3.40282e+38   
```

We expect to be aboe to use both `m.foo` and `m.bar` to access the value, in a case-insensitive manner, but that using `bar` will report a warning. The difference is in the case used when reporting the parameter name - previously we always use the lower-case version whereas we now report the case that the user entered.

## CIAO 4.16

```
>>> m.foo.val
0.0
>>> m.bar.val
/soft/ciao-4.16/lib/python3.11/site-packages/sherpa/models/model.py:583: DeprecationWarning: Parameter name bar is deprecated for model MyModel, use foo instead
  warnings.warn(wmsg, DeprecationWarning)
0.0
>>> m.BaR.val
/soft/ciao-4.16/lib/python3.11/site-packages/sherpa/models/model.py:583: DeprecationWarning: Parameter name bar is deprecated for model MyModel, use foo instead
  warnings.warn(wmsg, DeprecationWarning)
0.0
```

## this PR

``` 
>>> m.foo.val
0.0
>>> m.bar.val
/lagado.real/sherpa/sherpa-xspec/sherpa/models/model.py:707: DeprecationWarning: Parameter name bar is deprecated for model MyModel, use foo instead
  warnings.warn(wmsg, DeprecationWarning)
0.0
>>> m.BaR.val
/lagado.real/sherpa/sherpa-xspec/sherpa/models/model.py:707: DeprecationWarning: Parameter name BaR is deprecated for model MyModel, use foo instead
  warnings.warn(wmsg, DeprecationWarning)
0.0
```

### DIFF
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -680,33 +680,34 @@ class Model(NoNewAttributesAfterInit):
     def __iter__(self) -> Iterator[Model]:
         return iter([self])
 
-    def __getattr__(self, name):
-        """Access to parameters is case insensitive."""
+    def __getattr__(self, name: str) -> Any:
+        """Access to parameters is case insensitive. Other fields are exact."""
 
+        # Any model that defines a parameter will end up looking for
+        # the _par_index field, and this is done before Model.__init__
+        # is reached (i.e. setting it there does not help), so create
+        # it if needed.
+        #
         if "_par_index" == name:
             if self.__dict__.get('_par_index') is None:
                 self.__dict__['_par_index'] = {}
             return self.__dict__['_par_index']
 
         lowered_name = name.lower()
-
-        def warn(oname, nname):
-            wmsg = f'Parameter name {oname} is deprecated for model ' + \
-                f'{type(self).__name__}, use {nname} instead'
-            warnings.warn(wmsg, DeprecationWarning)
-
         parameter = self._par_index.get(lowered_name)
-
         if parameter is not None:
             if lowered_name in parameter.aliases:
-                warn(lowered_name, parameter.name)
+                wmsg = f'Parameter name {name} is deprecated for model ' + \
+                    f'{type(self).__name__}, use {parameter.name} instead'
+                warnings.warn(wmsg, DeprecationWarning)
+
             return parameter
 
         NoNewAttributesAfterInit.__getattribute__(self, name)
 
-    def __setattr__(self, name, val):
-        lname = name.lower()
-        par = getattr(self, lname, None)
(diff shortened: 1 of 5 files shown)


sherpa/models/model.py
```diff
@@ -1118,10 +1118,14 @@ def __init__(self, name: str, parts: Sequence[Model]) -> None:
 
         Model.__init__(self, name, allpars)
 
+        # This check should probably be removed since it refers to
+        # loading a very-old session, and that is unlikely to work due
+        # to other changes in the code base.
```
hamogu, line 1123: So, should it be removed? OK with me...

hamogu (approved):
If @dougburke says this can be removed and I think it cna be removed, it can probably be removed. I'm approving, evne eithout removing it, because that does not have to done in this PR, but I'm also OK (i.e. I don't need to review again) with adding another commit to remove it.

sherpa/models/model.py
DougBurke: I've created #2075 to track this, as it's separate from this PR

PR #2076, opened 2024-07-05
Rework plot classes

DougBurke:
Follow on to 
- #1988

This had been crazier, as it used to also include

- using slots rather than NoNewAttributesAfterInit
- using protocol-like classes to determine how to create the string representation and whether something was plottable or contourable

The idea here is to see if breaking the plot hierarchy - so that the class that does the plotting/contouring is separate from the class that has a `prepare` method - is helpful.

### DIFF
--- a/docs/plots/index.rst
+++ b/docs/plots/index.rst
@@ -106,11 +106,13 @@ to their ``prepare`` methods - normally data and model objects,
 but plot objects themselves can be passed around for "composite"
 plots - but there are several classes that accept the values to
 display directly:
-:py:class:`~sherpa.plot.Plot`,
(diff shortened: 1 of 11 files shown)


DougBurke:
Rebased to address conflicts. Still broken.

DougBurke:
Rebased onto the latest version of #1988 and fix up the commits from all the rebasing and let the tests pass. Whether the result is sensible is TBD.

DougBurke:
Now rebased to actually better merge in the changes that have already been merged, and fix up the docs build.This doc fix does suggest that the `xx` -> `xxObject` naming (e.g. `Plot` -> `PlotObject`) is superfluous (but it does make it easier to track down where symbol names may be being used)

PR #2078, opened 2024-07-08
Save the default identifier in save_all output

### LINKED ISSUE #2077: default id is not included in save_all
The following (with CIAO 4.17 drop 1, but the same is true with 4.16.1) shows that save_all does not save changes to the default id. It would be an easy fix.

```
sherpa In [1]: from io import StringIO

sherpa In [2]: out = StringIO()

sherpa In [3]: set_default_id("foo")

sherpa In [4]: load_arrays("foo", [1,2], [1,2])

sherpa In [5]: save_all(out)

sherpa In [6]: txt = out.getvalue()

sherpa In [7]: print(txt)
import numpy
from sherpa.astro.ui import *

######### Load Data Sets

load_arrays("foo",
            [1, 2],
            [1, 2],
            Data1D)


######### Set Statistic

set_stat("chi2gehrels")


######### Set Fitting Method

set_method("levmar")

set_method_opt("epsfcn", 1.1920928955078125e-07)
set_method_opt("factor", 100.0)
set_method_opt("ftol", 1.1920928955078125e-07)
set_method_opt("gtol", 1.1920928955078125e-07)
set_method_opt("maxfev", None)
set_method_opt("numcores", 1)
set_method_opt("verbose", 0)
set_method_opt("xtol", 1.1920928955078125e-07)



sherpa In [8]: 
```

DougBurke:
# Summary

Record the default identifier in save_all output if the user has changed it. Fix #2077

# Details

Fortunately this is an easy fix. I have not added an explicit test of the output - i.e. no check of the actual text file that was create. Instead I rely on "implicit" tests where we load in the output of save_all and check that the default identifier has been changed.

### DIFF
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -1440,6 +1440,18 @@ def _save_dataset(out: OutType,
         _output(out, f"set_syserror({idstr}, {syserr.tolist()})")
 
 
+def _save_id(out: OutType, state: SessionType) -> None:
+    "Does the default id need changing?"
+
+    defid = state.get_default_id()
+    if defid == 1:
+        return
+
+    _output_banner(out, "Default identifier")
+    _output(out, f'set_default_id({repr(defid)})')
+    _output_nl(out)
+
+
 def save_all(state: SessionType, fh: Optional[TextIO] = None) -> None:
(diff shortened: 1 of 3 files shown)


anetasie (approved):
Looks good.

PR #2079, opened 2024-07-10
minor cleanup of sherpa.optmethods

DougBurke:
# Summary

Internal clean up of the sherpa.optmethods code.

# Details

This is a subset of #2047 and comes about because I'm trying to address #2007 but the code in the `optmethods` directory hasn't been looked at in a while. So we (not in order)

- remove unused code (trying to only remove obviously-private routines to avoid issues noted after merging #2015)
- remove some debugging code that the user could never trigger (and we have better ways to log this info if we need to add such  logging)
- replace some code with more-modern numpy idioms (e.g. using `np.isnan`, taking advantage of 'x + y' being automatically "vectorized", and routines like `np.full` and `np.clip`)
- avoid creating multiple local functions when we can just write it once
- remove an attempt to support FORTRAN vs C array ordering for code which is only meant to be sent 1D arrays
- fix an array in an "internal" routine (`difevo_nm`) which would error out if called with `population_size=None` [which we never do but a user could]; in this case I don't think it's worth adding a test for this as we really should do this for all such routines and that's  a lot of noise and it's hard to validate the results of a successful run (other than "does not crash")
- try to avoid reusing the same variable to store different types (e.g. array vs scalar)
- try to clean up tuple handling, since type checkers don't really like "tuple_val += (x, y)" and I believe the code is actually clearer with the new version

The idea here is to make the code easier to review (and to identify regressions if things do change). I will note that some of these changes were found (or strongly hinted at) by type checkers, but adding significant type checker support is for a follow-on PR.

### DIFF
--- a/sherpa/optmethods/optfcts.py
+++ b/sherpa/optmethods/optfcts.py
@@ -131,11 +131,6 @@ def _move_within_limits(x, xmin, xmax):
         x[above] = xmax[above]
 
 
-def _my_is_nan(x):
-    fubar = list(filter(lambda xx: xx != xx or xx is np.nan or np.isnan(xx) and np.isfinite(xx), x))
-    return len(fubar) > 0
-
-
 def _double_check_limits(myx: np.ndarray,
                          myxmin: np.ndarray,
                          myxmax: np.ndarray) -> None:
@@ -466,7 +461,7 @@ def minim(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, step=None,
     orig_fcn = stat_cb0
 
     def stat_cb0(x_new):
-        if _my_is_nan(x_new) or _outside_limits(x_new, xmin, xmax):
+        if np.isnan(x_new).any() or _outside_limits(x_new, xmin, xmax):
             return FUNC_MAX
         return orig_fcn(x_new)
 
@@ -953,7 +948,7 @@ def neldermead(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None,
     orig_fcn = stat_cb0
 
     def stat_cb0(x_new):
-        if _my_is_nan(x_new) or _outside_limits(x_new, xmin, xmax):
+        if np.isnan(x_new).any() or _outside_limits(x_new, xmin, xmax):
             return FUNC_MAX
         return orig_fcn(x_new)
 
--- a/sherpa/optmethods/optfcts.py
+++ b/sherpa/optmethods/optfcts.py
@@ -380,7 +380,7 @@ def grid_search(fcn, x0, xmin, xmax, num=16, sequence=None, numcores=1,
                 raise TypeError(f"{seq} must be of length {npar}")
 
     answer = eval_stat_func(x)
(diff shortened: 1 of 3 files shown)


sherpa/optmethods/optfcts.py
```diff
@@ -122,78 +124,52 @@ def _get_saofit_msg(maxfev, ierr):
...
+def _raise_min_limit(factor: float,
+                     xmin: np.ndarray,
+                     x: np.ndarray
+                     ) -> np.ndarray:
+    """Calculate the new minimum limits."""
 
+    myxmin = x - factor * np.abs(x)
+    below = np.flatnonzero(myxmin < xmin)
+    if below.size > 0:
+        myxmin[below] = xmin[below]
```
hamogu, line 145:
```suggestion
    below = myxmin < xmin
    myxmin[below] = xmin[below]
```

sherpa/optmethods/optfcts.py
```diff
@@ -122,78 +124,52 @@ def _get_saofit_msg(maxfev, ierr):
...
+                     x: np.ndarray,
+                     xmax: np.ndarray
+                     ) -> np.ndarray:
+    """Calculate the new maximum limits."""
+
+    myxmax = x + factor * np.abs(x)
+    above = np.flatnonzero(myxmax > xmax)
+    if above.size > 0:
+        myxmax[above] = xmax[above]
+
```
hamogu, line 160:
```suggestion
    above = myxmax > xmax
    myxmax[above] = xmax[above]

```

sherpa/optmethods/optfcts.py
```diff
@@ -273,31 +224,29 @@ def difevo_lm(fcn, x0, xmin, xmax, ftol=EPSILON, maxfev=None, verbose=0,
...
 
     if population_size is None:
         population_size = 16 * x.size
 
     if maxfev is None:
         maxfev = 1024 * x.size
 
+    # TODO: can we not just call x.size rather than
+    #       np.asanyarray(fcn(x)).size for the last argument?
+    #
```
hamogu, line 240: If `fcn` does not change the size of the array, the answer is clearly : yes.

DougBurke: Unfortunately I'm not convinced we can guarantee we have that condition here [it would be easy to check but even then there's enough going on with this code, including the fact we currently don't seem to have any tests of it, that I'm reluctant to change things too much.

sherpa/optmethods/optfcts.py
hamogu: Yes, even better. From 3 lines to 2 lines to one line. More readable!

sherpa/optmethods/optfcts.py
hamogu: This is not a bit time sink, so not a problem.

sherpa/optmethods/optfcts.py
DougBurke: So I've made these changes, as well as in `_move_within_limits` which then becomes a one-liner so can get put directly into the one routine that called it.

hamogu (approved):

PR #2080, opened 2024-07-12
Calculate absorption models from just one value

hamogu:
## Summary

Speed up Sherpa fits that use XSPEC absorption models by using a specialiazed cache that evaluated absorption models only once to cache the absorption corss sections sigma(E) and then calcualtes exp(-N_H sigma(E)) in Python.

## Details
This introduces a new type of cache, that, instead if caching the model value for each parameter value, cached is for the parameter value=0.1. While the cache in general does not know about which model it is applied to, this is specifically designed for absorption models in XSPEC, which have the form

out(E) = exp(-N_h * sigma(E))

Those models are relatively expensive to calculate because it requires evaluation of a large number of atomic cross-sections. So, instead, we just evaluate them once.
The fiducial value of 0.1 is physically motivated, because a larger number (say N_h=1) already gives no throughput at low energies, and we end up with values of 0, which causes problems for log(0).

This new model is NOT applied to "XSwabs", which is an obsolete model used for historical purposes only and seems to be to inaccurate internally, that the absorption calculated with that law, is not actually an exponentia. It should be, according to the XSPEC docs, but numerically it's not. Instead of debugging that and making choices on how to deal with it, it's simply not touched in this PR.

In a test based on the benchmarking code at
I found a speedup of a factor of 3.
```Python
ui.load_pha(
    1, "../curated-test-data/Chandra/ACIS/acisf04487_001N023_r0009_pha3.fits.gz"
)
ui.group_counts(1, 5)
ui.ignore(None, 0.3)
ui.ignore(7, None)
abs1 = XSphabs("abs1")
abs2 = XSphabs("abs2")
vapec1 = XSvapec("vapec1")
vapec2 = XSvapec("vapec2")
model = abs1 * vapec1 + abs2 * vapec2
ui.set_source(1, model)
import time

abs1.nH = 0.1
abs2.nH = 1
vapec1.kT = 0.5
vapec2.kT = 5
vapec1.norm = 1e-2
vapec2.norm = 1e-2

start_time = time.perf_counter()
ui.fit(1)
end_time = time.perf_counter()
print(end_time - start_time, "seconds")
```

I admit that this is a model that is very favourable for this cache (since there are not many other parameters to fit), but the speedup is significant even in other cases.


hamogu:
@DougBurke I'm not sure how this interacts with your code to include new XSPEC versions, but all the XSPEC models I touch are pretty well established and stable, so hopefully that's not a problem.

A note on applicability:
This could be extended to work with more complex absorption models (e.g. those where the redshift of the absorber is free), but let's do one thing at a time to keep the review manageable. As a stellar observer, I might be biased, but all X-ray fits I've ever done used either XSphabs or XSTBabs, so this would give me a speedup of about a factor 1.5-3 to essentially every Sherpa fit I've ever done - enough to be worth reviewing.

hamogu:
And... apparently one of my rebases has gone wrong and removed 6000 lines of XSPEC models... will fix ASAP.

hamogu:
I'm leaving the test failures for now - please review. The failing test is an example what kind of difference we will get with this PR: Because the of numerical precision with float32 numbers, "exp(log(-nh sigma(e))" is not going to give us an exact answer. Here, the relative tolerance of the tests is 1e-6, while the numerical error is slightly larger than that, but < 1e-5

I personally think that's OK - that's still orders of magnitude better than anything we realistically measure.

Before merging, we can just change the lines that fail to `assert expected == pytest.approx(model, rtol=1e-5)`, but for now, I want to leave it so that @DougBurke and @anetasie can see the difference.

anetasie:
@hamogu this is an interesting idea. I have to install xspec in my test environment to give it a try.

DougBurke:
@anetasie - you can use the CIAO conda channel to install just xspec for you. You'd have to do something like

```
conda install -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge xspec-modelsonly fftw
```

(you don't need fftw but may as well get it too) and then edit `setup.cfg` either manually or with a helper script like

```
wget https://raw.githubusercontent.com/sherpa/sherpa/f400e41e6c23e48c9e978d4cc887bec3f4eff0da/scripts/use_conda_config
chmod u+x ./use_conda_condig
./use_conda_config
```

Note that this would give you XSPEC 12.13.1e (or something close to that).

anetasie:
@DougBurke thanks. I'll try this method.

anetasie:
@DougBurke I must been doing something wrong as I am getting the following error. 

```

In [2]: import sherpa.astro.xspec
ImportError                               Traceback (most recent call last)
File ~/git/sherpa/sherpa/astro/xspec/utils.py:24
     23 try:
---> 24     from . import _xspec
     25 except ImportError as ie:
     26     # It would be nicer for the user to say "from None" rather than
     27     # "from ie" here but it may lose important information (in the
     28     # case there is some problem initializing the XSPEC library,
     29     # rather than it just not being built).
     30     #

ImportError: dlopen(/Users/aneta/git/sherpa/sherpa/astro/xspec/_xspec.cpython-311-darwin.so, 0x0002): Library not loaded: '@rpath/libcblas.3.dylib'
  Referenced from: '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libgsl.25.dylib'
  Reason: tried: '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/bin/../lib/libcblas.3.dylib' (no such file), '/Users/aneta/miniconda3-arm/envs/sherpagit/bin/../lib/libcblas.3.dylib' (no such file), '/usr/local/lib/libcblas.3.dylib' (no such file), '/usr/lib/libcblas.3.dylib' (no such file)

ImportError                               Traceback (most recent call last)
Cell In[2], line 1
----> 1 import sherpa.astro.xspec

File ~/git/sherpa/sherpa/astro/xspec/__init__.py:124
    119 from sherpa.utils.numeric_types import SherpaFloat
    121 # Note that utils also imports _xspec so it will error out if it is
    122 # not available.
    123 #
--> 124 from .utils import ModelMeta, version_at_least, equal_or_greater_than
    125 from . import _xspec  # type: ignore
    128 info = logging.getLogger(__name__).info

File ~/git/sherpa/sherpa/astro/xspec/utils.py:31
     24     from . import _xspec
     25 except ImportError as ie:
     26     # It would be nicer for the user to say "from None" rather than
     27     # "from ie" here but it may lose important information (in the
     28     # case there is some problem initializing the XSPEC library,
     29     # rather than it just not being built).
     30     #
---> 31     raise ImportError("XSPEC support is not enabled") from ie
     33 __all__ = ['ModelMeta', 'include_if', 'version_at_least']
     36 def get_version(version):

ImportError: XSPEC support is not enabled

```

hamogu:
You did not build XSPEC successfully. That likely means one of two things (or both):
- You did not run the script successfully. When you tried "% ./scripts/use_ciao_config", did it say which XSPEC version is found? (see https://sherpa.readthedocs.io/en/4.16.1/developer/index.html#install-from-source-in-conda)
- You have old stuff (binaries from an older build that did not include XSPEC) laying around in your sherpa directory. Try "git clean -dfn" to see what the old stuff is and, if you are sure you don't need to keep any of it, "git clean -dfx" to remove it before you compile/install sherpa.

anetasie:
@hamogu I followed the instructions in RTD and create a sherpaciao environment and this successfully installed sherpa with xspec models. I should be able to test this PR now.

anetasie:
@hamogu Are the redshifted absorption models, such as `xszphabs` model also cached if the redshift is fixed?

> This could be extended to work with more complex absorption models (e.g. those where the redshift of the absorber is free), but let's do one thing at a time to keep the review manageable

hamogu:
No, that would be additional work. There is more that could be done with this scheme of Sherpa doing part of the calculation (red-shfited absorption models and the norm for e.g. apec models come to mind), but I'm trying to get the biggest bank for the buck first. If you and @DougBurke agree that this works, we can do a follow-up PR to make it work for more models later.

(The relative gain is highest for the models with the fewest parameters. With the PR, phabs is called only once (cutting runtime by almost 100%, while for a red-shifted abs, it's only going to cut about 50%.)

This buffering is all internal (at least the way we apply it to XSPEC), so we can change them without worrying about backwards compatibility in my opinion.

DougBurke:
I have found typos to report... I want to think about my answer before reporting the trivial though.

sherpa/astro/xspec/__init__.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015 - 2024
+#  Copyright (C) 2010, 2015, 2024
```
DougBurke, line 2: Unfortunately your "ensure a sensible copyright line" script doesn't handle ranges, so it's messed it up here (and in at least one other flle).

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -805,3 +816,37 @@ def test_apec_redshift_parameter_is_case_insensitive():
...
+
+
+@requires_xspec
+@pytest.mark.parametrize("modelcls", get_xspec_models_with_exp_cache())
+def test_models_with_exp_cache(modelcls):
+    '''Tests that models with the modelCacher1d_exp work correctly.
+
+    Unlike modelCacher1d, does not evaluate the model for every
+    parameter value once, but modelCacher1d_exp assumes a certain
+    functional form and calcualtes model values by itself, this this
```
DougBurke, line 838:
```suggestion
    functional form and calculates model values by itself, this this
```

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -805,3 +816,37 @@ def test_apec_redshift_parameter_is_case_insensitive():
...
+    warrantes a test for every model using this cache.
```
DougBurke, line 839:
```suggestion
    warrants a test for every model using this cache.
```

DougBurke: I think this paragraph could just be as simple as

    Test the models that use modelCacher1d_exp.

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -805,3 +816,37 @@ def test_apec_redshift_parameter_is_case_insensitive():
...
+    '''
+    x = np.arange(0.1, 10, 0.1)
+    xlo = x[:-1]
+    xhi = x[1:]
+    mdl = modelcls()
+    for nh in [1, 0.01, 0.123, 2.3, 11.2]:
+        abs_direct = mdl.calc.__wrapped__(mdl, (nh,), xlo, xhi)
+        abs_cache = mdl.calc((nh,), xlo, xhi)
+        # XSPEC models use float32, so we need to use a relative
+        # tolerance larger than the default
```
DougBurke, line 849: So, it's a bit more than this, as I see differences across platforms and also we cant guarantee that the model uses float32 - some use float64 (although most of the absorption models do use float32). So I'm not sure a comment here is needed.

sherpa/models/model.py
```diff
@@ -488,6 +489,116 @@ def cache_model(cls, pars, xlo, *args, **kwargs):
...
+
+    Unlike the more general `sherpa.models.model.modelCacher1d`, this
+    version is designed for models with a single parameter and a model of
+    the following form:
+
+    math::
+        y(x) = \exp(p * a(x))
+
+    where `p` is the parameter. The cacher will evaluate the model once
+    with `p=1` and then store that. This is useful in particular for
```
DougBurke, line 513:
```suggestion
    with `p=0.1` and then store that. This is useful in particular for
```

I think this is the case. We *could* make the normalization point an argument to the routine (in some manner), but I don't suggest we do that until we decide it's needed.

sherpa/models/model.py
```diff
@@ -488,6 +489,116 @@ def cache_model(cls, pars, xlo, *args, **kwargs):
     return cache_model
 
 
+# This shares a lot of the code with modelCacher1d, but
```
DougBurke, line 492: Would it make sense to move both these routines to something like `sherpa/models/cache.py` and then we just re-export them from here? It wouldn't help with the fact that `modelCacher1d` isn't easily extensible, but does separate out the code a bit.

sherpa/models/tests/test_model.py
```diff
@@ -868,6 +869,24 @@ def check_cache(mdl, expected, x, xhi=None):
...
+
+    Note that a direct calc  gives different values than when
+    using the cache. Obviously, that's not what a real model
+    would do, but it's useful for testing.
+    '''
+
+    def __init__(self, name='expcache'):
+        self.foo = Parameter(name, 'foo', 1, 1e-10, 10)
+        ArithmeticModel.__init__(self, name,
+                                 (self.foo))
```
DougBurke, line 883: shouldn't this be

```suggestion
                                 (self.foo,))
```

sherpa/models/tests/test_model.py
```diff
@@ -868,6 +869,24 @@ def check_cache(mdl, expected, x, xhi=None):
     assert cache[digest] == pytest.approx(expected)
 
 
+class ExpcacheTestModel(RegriddableModel1D):
+    '''A model to test the modelCacher1d_exp decorator.
+
+    Note that a direct calc  gives different values than when
+    using the cache. Obviously, that's not what a real model
+    would do, but it's useful for testing.
```
DougBurke, line 877: Is the point here that `calc` doesn't scale the `a(x)` term - here `np.ones_like` - by an exponential?

sherpa/models/tests/test_model.py
```diff
@@ -899,6 +918,42 @@ def test_evaluate_cache1d(cached):
...
+
+    if cached:
+        check_cache(mdl, np.log(expected), xgrid, pars_in_cache=[False])
+    else:
+        assert len(mdl._cache) == 0
+
+
+    mdl.foo = 1
+
+    # Note that the model is itentiaonally defined such that
```
DougBurke, line 944:
```suggestion
    # Note that the model is intentionally defined such that
````

File: sherpa/models/tests/test_model.py [hunk 10]
```diff
@@ -1245,7 +1338,10 @@ def test_cache_clear_single():
...
 
 
-def test_cache_clear_multiple():
+@pytest.mark.parametrize('model',
+                            [Polynom1D, ExpcacheTestModel])
+
+def test_cache_clear_multiple(model):
     """Check cache_clear for a combined model."""
 
     p = Polynom1D()
```
DougBurke, line 1347: Did you mean this (otherwise `model` doesn't actually get used)?

```suggestion
    p = model()
```

sherpa/models/tests/test_model.py
```diff
@@ -1031,20 +1086,33 @@ class DoNotUseModel(Model):
     _cache_ctr: dict[str, int] = {'hits': 0, 'misses': 0, 'check': 0}
     _queue = ['']
 
+    def __init__(self, name='expcache'):
+        self.foo = Parameter(name, 'foo', 1, 1e-10, 10)
+        super().__init__(name, (self.foo))
```
DougBurke, line 1091:
```suggestion
        super().__init__(name, (self.foo,))

```

sherpa/models/model.py
DougBurke: There is also the thought that maybe `0.1` can still lead to 0's in some cases (or is that not really an issue)?

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -805,3 +816,37 @@ def test_apec_redshift_parameter_is_case_insensitive():
...
+@pytest.mark.parametrize("modelcls", get_xspec_models_with_exp_cache())
+def test_models_with_exp_cache(modelcls):
+    '''Tests that models with the modelCacher1d_exp work correctly.
+
+    Unlike modelCacher1d, does not evaluate the model for every
+    parameter value once, but modelCacher1d_exp assumes a certain
+    functional form and calcualtes model values by itself, this this
+    warrantes a test for every model using this cache.
+    '''
+    x = np.arange(0.1, 10, 0.1)
```
DougBurke, line 841: I was going to suggest decreasing the upper bound since we know not much happens >~ 2 keV, but I just tried with the tbabs model and there's some interesting going on around 7 keV we might want to ensure we catch.

We could decrease the bin width though - here's a comparison of the tbabs model with de=0.1 (top) and de=0.01 (bottom) - it might be nice to check we catch the wiggles well (or get a better handle on how well we can catch them)?

sherpa

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -805,3 +816,37 @@ def test_apec_redshift_parameter_is_case_insensitive():
...
+
+
+@requires_xspec
+def get_xspec_models_with_exp_cache():
+    """Return the XSPEC models that use the modelCacher1d_exp cache.
+
+    Currently, that is: Models that inherit from XSAbsorptionModel.
+    """
+    models = get_xspec_models()
+    return [m for m in models if issubclass(m, XSAbsorptionModel)]
```
DougBurke, line 828: Would it make sense to test the number of models that are found (more just as a regression test than anything else)?

sherpa/models/model.py
DougBurke: If you look at the plots I show with `tbabs` I still get 0's around 0.1 keV when `nh = 0.1` - which might suggest we need to use a smaller value (although I may have used `tbabs` incorrectly as I didn't `set_xsabund('wilm')` or whatever I should have done.

DougBurke:
So, the "major" thoughts I have here are

a) I have been hoping to make our XSPEC interface require less manual intervention - e.g. see

- #52

and this goes against this. Now, maybe what I should be aiming for is that the compiled interface is auto-generated (that is, `sherpa.astro.xspec._xspec`) and the actual Python support (`sherpa/astro/xspec/__init__.py`) is still manually curated, in which case there's no issue here

b) at some level this is a hack, and I worry that some model somewhere is going to be somewhat special and cause the caching code to fail. Fortunately there is an explicit test of this, but maybe the failure happens outside the `nH` range we test.

Now, at some level all of our models are "hacks", in that they are (modulo something really easy like a gaussian) approximations (and even in the gaussian case we still likely rely on an approximation somewhere).

So, what we probably want is some way to easily turn on/off this behaviour - so that users who "care about the purity of the model" or "I just want an answer fast" can both be handled. I think our current caching code does not make this obvious, so should we add something here?

hamogu:
I note that this is not applied randomly: We need to hand-select which model to apply this to. So it's not "some model somewhere", it's models with known physics. In fact, when I tested this, I already found one model where it's not working "wabs", but what's wrong is the model (it's numerical accuracy is too low, so that you see wiggles when you plot the absorption at some wavelengths as a function of N_H), not the caching. Now, wabs should not be used any longer anyway, so I opted to leave this out.

We also know (from physics) that all the absorption models are pretty similar. While they use different absorption cross sections, differences are not that large. That means if 0.1 as fiducial value works for one of them, it's going to work for the other, too.

Last, we also have some knowledge on how X-ray fitting works: These are multiplicative models, and the difference between 0.999 and 0.9999 will never be visible. If you have low counts, that's obvious, but importantly, the uncertainty in the ARF and RMF and in the atomic physics of the atomic cross-sections is larger than that. Anybody who thinks they can fit to that precision is fooling themselves. Even is the formal error might be that small (and I've never seen that) that true (incl systematics) error is not. So the numbers that come out might be different, but the physics is not.

So, we can decrease the bins size and increase the energy range covered for the tests, but is this something that we fundamentally want?

If it turns out that there is no single nh value that works for all inputs, then I've thought about the following options, but both have drawbacks:
- Use ranges, e.g. nh=0..1 uses the cached nh=0.1 and nh>1 uses cached nh=10. Downside to this approach is that one had to be careful to not introduce a step at nh=1, which might confuse the optimizers. That might require some zone of interpolation e.g. between 0.9 and 1.1 to make sure it's smooth. Possible, but more complex than I had hoped for.
- Use ranges in energy where the curve is known to be comfortable away from 0 and 1, e.g. use nh=0.1 as fiducial for energies < 2 keV and nh=10 for energies > 2 keV. Probably easier, but needs to be usable with bins with xlo below the boundary and xhi above the boundary.

So, I hope that we don't need this, but I can implement it if we do.

Why am I saying that? I want this to be "automatic", not "opt-in". We already have an "opt-out" (`use_cache=False`), we might just have to document it better. 

A variant of this idea is to have a more complex cache class, where you can switch between different behaviors on a per-model basis:
- include parameter in cache hash (what we do now for all parameters)
- use parameter = 1 when calculating cached value and scale linearly (for use with `norm`) parameters
- use parameter =0.1 when calculating cached value and scale exponentially (for use with `nh`) parameters - this pr.

Then, users could switch from one to the other if they needed to. But is that worth the added complexity? Probably beccause it will bring benefits to more models (e.g. red-shifted absorption). However, it would be great to iron out the approach with the single case in this example and then generalize in a later PR.

sherpa/models/model.py
hamogu: See my general comment - this might need more complexity. If we go significantly below 0.1 we don't cover the high-energy end any longer. but that can be done, just at the cost of some complexity.

sherpa/astro/xspec/__init__.py
hamogu: I wosh we could just drop the years form the copyright line. Would be way easier/ Many projects don't do it at all and copyright last so many decades that a year more or less has no practical relevance. Anyway, not for me to decide. It is what it is...

anetasie:
One comment on the wiggles in the absorption curve. They help to constrain NH and also sometimes you do want to detect the wiggles indicating transition edge, such as for example Oxygen edge.

hamogu:
Yes - the idea is to replicate the XSPEC model with a relative tolerance of 1e-4 or so - that will preserve all real wiggles.

hamogu:
@anetasie When I talk about "wiggles" in https://github.com/sherpa/sherpa/pull/2080#issuecomment-2246980233 I mean "wiggles in a a plot of transmission vs. N_H for a fixed energy". That particular plot should be a smooth exponential with no wiggles, because out(E) = exp(-N_h * sigma(E)), which, for a fixed E is just proportional to exp(-N_h). If there are wiggles in there for the wabs model, it means the model is wrong - presumably because the accuracy used internally for the computations is too low.

My goal is to cache essentially the "sigma(E)" (with some scaling, e.g. for N_H = 0.1). So if your model contains the XAFS edges, then the cashed version will.

sherpa/astro/xspec/tests/test_xspec.py
hamogu: Sure, doesn't hurt.

sherpa/astro/xspec/tests/test_xspec.py
hamogu: The "wiggles" you talk about here arer the physical absorption edges. If you look closely, you'll the edges fo C, N, O, Si, Fe and and probably a few others at different energies. Those are not affected by this method. We don't interplotate over wavelength, but just over N_H. Just as in all other applications of our caching code, the `xlo,xhi` are part of the key of the cache. If you call with a different set of `xlo,xhi` the orininal model will be evaluated (and cached) again.

The only thing we need to check is that the orange curve is not identical to 0 or 1 in the plot above.

hamogu:
Testing over a larger range is the right call though. This plots shorts the difference between evaluating directly and using my cache, x-axis is energy in keV. Note the y-scale for most plots is 1e-5 or smaller, but there are differences at very small energies and then some of order 0.2% for E > 50 keV at N_H > 1e26. I think the latter case is not physically relevant and we don't have to worry much, but I have to improve the difference for energies below 0.04 keV.
image
(blue = XSPhabs, orange=XScabs, green=XSTBabs)

hamogu:
I've updated this. I still want to do some sleuthing about why exactly I needed to loosen up some of the numerical comparisons in the tests, but first, let me see if this actually passes the full test suite here.

The new idea here is that instead of using one canonical value which can create problems when the log(absorption) iis either 0 and 1 to numerical precision for high or low values of the energy,
I split the energy into different regions that can be used with different canonical values of N_H.

This requires fine tuning at the level of individual models.
It also requires to loosen up some of the numerical tests, because values are not subtely different, but not how e.g. the fitted values of N_H in one test is essentially the same, it's just he fit statistics that differs slightly.

See https://github.com/hamogu/sherpa-opt-benchmark/blob/main/notebooks/optimize_absorption_cache.ipynb for why I chose the values I chose.

sherpa/astro/tests/test_astro.py
```diff
@@ -407,7 +407,7 @@ def cmp_thread(fres, abs2, mek1, mek2, covarerr):
         assert covarerr[2] == approx(0.00377755, rel=1e-3)
         assert covarerr[3] == approx(0.00370543, rel=1e-3)
         assert covarerr[4] == approx(0.0016608, rel=1e-3)
-        assert fres.statval == approx(0.0496819, rel=1e-4)
+        assert fres.statval == approx(0.0496819, rel=2e-2)
```
hamogu, line 410: I would like to understand when the values are different in this PR but (a) the actual fit values are the same, just the statistics is different and (b) I can't reproduce that outside this test. If I run the same thread separately (`fit-spectrum.py` in the sherpa-test-data repro, then I get different numbers. Not very different (for a science application), but different enough that I can't compare them to the numbers in this test. Which is somewhat surprising and means I don't actually understand what this test tests.
```Python
Python 3.11.7 | packaged by conda-forge | (main, Dec 15 2023, 08:41:36) [Clang 16.0.6 ]
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.

   ...: 
   ...: load_pha("data1.pi")
   ...: load_arf("arf1.fits")
   ...: load_rmf("rmf1.fits")
   ...: ignore()
   ...: notice(0.5,7)
   ...: 
   ...: set_source(xsphabs.abs2*(xsmekal.mek1+xsmekal.mek2))
   ...: 
   ...: abs2.nh=1.1
   ...: mek1.kt=0.8
   ...: mek1.norm=0.7
   ...: mek2.kt=2.4
   ...: mek2.norm=1.
   ...: 
   ...: mek1.norm.max=1.e10
   ...: mek2.norm.max=1.e10
   ...: mek1.norm.min=1.e-10
   ...: mek2.norm.min=1.e-10
WARNING: systematic errors were not found in file 'data1.pi'
statistical errors were found in file 'data1.pi'
dataset 1: 0.00146:14.9504 Energy (keV) -> no data
dataset 1: no data -> 0.4964:7.008 Energy (keV)

In [2]: fit()
Dataset               = 1
Initial fit statistic = 18742.8
Final fit statistic   = 129.134 at function evaluation 25
Data points           = 446
Degrees of freedom    = 441
Probability [Q-value] = 1
Reduced statistic     = 0.292821
Change in statistic   = 18613.7
   abs2.nH        1.13768      +/- 0.00152752  
   mek1.kT        0.855161     +/- 0.00111268  
   mek1.norm      0.682236     +/- 0.00366547  
   mek2.kT        2.34117      +/- 0.00365177  
   mek2.norm      1.04396      +/- 0.00164369 
```

DougBurke: So, this is an old test, and it's written in an interesting style. We have long lost any information on what the tests were meant to be showing - are they intended as regression tests or are they there to exercise a particular piece of functionality.

I've found many of the model fits in these tests look poorly constrained, although here at least the two kT values look significantly different. Maybe @anetasie can remember what this test was meant to be showing.

Always beware the "hidden" XSPEC settings - e.g. get_xsabund() and get_xsxect() - when comparing things.

hamogu: I followed the chain. Differences per bin are of order 1e-8, but because the statistic only looks are the difference between data and model (and not the absolute number) it adds up over several bins, leading to a significant change (1.2e-2) in the reduced statistic. The best fit is identical though, so I think this change is safe to accept.

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -754,7 +767,7 @@ def test_evaluate_xspec_model(modelcls):
     assert_is_finite(evals, modelcls, "energy")
     assert_is_finite(wvals, modelcls, "wavelength")
 
-    assert wvals == pytest.approx(evals)
+    assert wvals == pytest.approx(evals, rel=1e-6, abs=2e-5)
```
hamogu, line 770: For XSphabs, there are now numerical difference between evaluating in energy and in wavelength space. The reason is that one of the bin boundaries (0.5 keV) is right at the boundary where I switch from one canonical value to the other and because of numerical differences energy->wavelength->energy does not end up on the same side of the 0.5 keV boundary.

I'm not happy to loosen up a general test that tests dozens of models for this one special case, but I also think it's not worth the complexity of treating just this one model different than the others.

hamogu:
While I think the concept that I propose in this PR is sound (and I have spend quite some time experimenting to find the best parameters) and the speed-up is considerable, I fear that this is doing step 3 before step 1.
As usual in any big code base (and Sherpa is no exception to that) the longer you dig, the more you find. I think a cleaner and more adaptable way to implement this is is to put the caching code into class definition instead of into a decorator. (Or make the decorator a class, so that it can have attributes.)

So, I suggest to not take this for 4.17, but push to the next release. I'm fairly confident that CXC funding won't be cut off right now, so we can afford to do this right. (Though of course it would be nice to deliver the speed improvement to users ASAP.)

sherpa/astro/xspec/tests/test_xspec.py
DougBurke: The wavelength checks are always less important to me ;-)

This change in tolerance doesn't appear terrible anyway.

DougBurke:
We can always provide "special-cased" versions of the absorption models in the CIAO contrib code for 4.17, which would

- let people see how it works
- benefit from the speed ups

at the expense of having to modify their code somewhat.

DougBurke:
I also apologise from dragging you away from this to review my changes.

PR #2081, opened 2024-07-15
Minor plot cleanup

DougBurke:
# Summary

Clean up some plot code to make better use of modern Python idioms. The string display of some of the objects has now changed so that the `name = value` lines are now aligned (except for the preference line).

# Details

This has been pulled out of #1988, concentrating on the "small clean up" stuff rather than the larger changes which were leading to #2076 - and I now believe that we don't have time to properly discuss them for 4.17.0, so we should just take the less controversial changes.

There is little behavioural change here: the only difference is that some of the string representation objects are now more consistent so that the `=` character lines up in the `xxx  = value\nyyy   = value\n...` output.

The other changes are to make more use of `super()` - which is a lot easier now after recent plot changes that have removed some of the interesting hierarchy we had - and to avoid defining methods which are the same as the superclass version.

One positive change, I think, is that it's a bit more obvious when we explicitly subvert the inheritance - since in those case we do not call methods via `super()` but explicitly call the class in question. This is needed because we have a hierarchy where the base classes have a `plot` method which takes the data to plot, but classes built on them change the API and require a call to `prepare` to set up fields in the object and the `plot` call just has `overplot`, `clearwindow`, and `**kwargs` as arguments. This is part of what #1988 and #2076 were about, but we can worry about cleaning that up in the future.

# Example

Note that the `name = value` lines now all line up on the `=` character **except** for the `histo_prefs` line (this was an intentional change as the preferences are slightly different and also generally would require adding even more spaces which just doesn't seem worth it).

## 4.16.0

```
>>> from sherpa.astro.plot import FluxHistogram
>>> p = FluxHistogram()
>>> print(p)
modelvals = None
clipped = None
flux = None
xlo    = None
xhi    = None
y      = None
xlabel = None
ylabel = None
title  = Model
histo_prefs = {'xlog': False, 'ylog': False, 'label': None, 'xerrorbars': False, 'yerrorbars': False, 'color': None, 'linestyle': 'solid', 'linewidth': None, 'marker': 'None', 'alpha': None, 'markerfacecolor': None, 'markersize': None, 'ecolor': None, 'capsize': None}
```

## Now

```
>>> from sherpa.astro.plot import FluxHistogram
>>> p = FluxHistogram()
>>> print(p)
modelvals = None
clipped   = None
flux      = None
xlo       = None
xhi       = None
y         = None
xlabel    = None
ylabel    = None
title     = Model
histo_prefs = {'xlog': False, 'ylog': False, 'label': None, 'xerrorbars': False, 'yerrorbars': False, 'color': None, 'linestyle': 'solid', 'linewidth': None, 'marker': 'None', 'alpha': None, 'markerfacecolor': None, 'markersize': None, 'ecolor': None, 'capsize': None}
```

### DIFF
--- a/sherpa/astro/plot.py
+++ b/sherpa/astro/plot.py
@@ -36,7 +36,7 @@ from sherpa.models.model import Model
 from sherpa import plot as shplot
 from sherpa.utils import parse_expr, dataspace1d, histogram1d, filter_bins, \
     sao_fcmp
-from sherpa.utils.err import IOErr, PlotErr, StatErr
+from sherpa.utils.err import IOErr, PlotErr
 
 warning = logging.getLogger(__name__).warning
 
@@ -695,7 +695,7 @@ class RMFPlot(shplot.HistogramPlot):
             # liable to change.
             #
             energies = [energy for energy in self.energies
-                        if energy >= elo[0] and energy < ehi[-1]]
+                        if elo[0] <= energy < ehi[-1]]
             if len(energies) == 0:
                 raise ValueError("energies must be "
                                  f">= {elo[0]} and < {ehi[-1]} keV")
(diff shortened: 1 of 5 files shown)


hamogu (approved):
Looks good to me. As the title for this PR says: It's not groundbreaking, it's cleaning things up for the next step.

Issue #2083, opened 2024-07-18
How to set the data units for plotting


DougBurke:
I wanted to set some axis labels so calls like `plot_data`, `plot_fit`, `plot_fit_resid`, `plot_ratio`, ... would use them automatically. I came up with

```python
# over-ride the x and y labels (should be a nice way to do this).
#
def xlabelify():
    return "Wavelength [$\AA$]"

def ylabelify(yfunc=None):
    return "Flux [10$^{-17}$ erg/cm$^2$/s/$\AA$]"


get_data().get_xlabel = xlabelify
get_data().get_ylabel = ylabelify
```

We could look at adding `set_xlabel`, `set_ylabel` methods to `sherpa.data.Data`.

There is also the issue of "how can we use astropy/whatever units" but that's a larger change (and in this case wouldn't have helped since the FITS file I was reading in had no units on the columns, I had to read a random page on the internetz to find them, but that's a different issue).

PR #2086, opened 2024-07-21
data: allow axis labels to be changed

DougBurke:
# Summary

Allow the independent and dependent axis labels to be changed. Address #2083

# Details

To complement

  get_xlabel, get_x0label, get_x1label, get_ylabel

add

  set_xlabel, set_x0label, set_x1label, set_ylabel

for the data types. This is trickier than at first thought, since

a) we introduce _xlabel, _ylabel, ... fields to store the labels
   so we can set up the defaults (in the __init__ call) and then
   over-ride them by the user

b) since we have some interesting hierarchies, with __init__ calls
   that can not be just passed to a superclass, we have to either
   create these attributes in more places than we'd want

c) some classes have labels that depend on the object setting
   (DataPHA and DataIMG), so we need to decide how to handle this.
   The assumption taken is that if a user has not called set_*label
   then the get_*label call returns the expected value, but as soon
   as set_xlabel is used then the requested label is always returned.

I thought about adding in UI routines to provide easy access to the user, but for now we rely on the user manually having to call the get_data routine for us.

# Examples

## generic data set

This uses a SDSS spectrum, but the data doesn't really matter (I just used this because it has units on both axes and so we want to label them).

First, the default behaviour

```python
plot_fit_delchi()
plt.ylabel(r"Residual [$\sigma$]")
plt.savefig("no-labels.png")
```

no-labels

With this PR

```python
get_data().set_xlabel("Wavelength [$\AA$]")
get_data().set_ylabel("Flux [10$^{-17}$ erg/cm$^2$/s/$\AA$]")

plot_fit_delchi()
plt.ylabel(r"Residual [$\sigma$]")
plt.savefig("with-labels.png")
```

with-labels

## PHA

The PHA case is special. Here are some default plots

```python
notice(0.3, 6)
plot_data()
```

pha-energy

```python
set_analysis("wave", type="counts", factor=2)
plot_data()
```

pha-wave

Now re-load the data and call

```
notice(0.3, 6)
get_data().set_xlabel("Foo Bar")
get_data().set_ylabel("Bar Foo")
plot_data()
```

pha-energy-2

```python
set_analysis("wave", type="counts", factor=2)
plot_data()
```

pha-wave-2

# Notes

- the `set_*label` call will remain - so a call to `plot_data` or `plot_model` ... will use the new labels.
- there's no way to change the y label for things like `plot_resid`/`plot_delchi`,`plot_ratio` since this is determined by the plot rather than the data, so you still may have to do post-processing work for your visualzation (e.g. the messing around with `plot.ylabel` call here)
- PHA data change the labels depending on choices (e.g. analysis setting, plot as a rate or counts, the factor setting) but once `set_*label` is called then that logic is "lost" (well, you can call `set_*label(None)` to restore it but that is currently not documented/tested).

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -935,9 +935,13 @@ class DataARF(DataOgipResponse):
         self._lo, self._hi = energ_lo, energ_hi
         self.energ_lo = energ_lo
         self.energ_hi = energ_hi
-        Data1DInt.__init__(self, name, energ_lo, energ_hi, specresp)
 
-    def _repr_html_(self):
+        super().__init__(name, energ_lo, energ_hi, specresp)
+
+        self._xlabel = 'Energy (keV)'
+        self._ylabel = None
+
+    def _repr_html_(self) -> str:
         """Return a HTML (string) representation of the ARF
         """
         return html_arf(self)
@@ -982,10 +986,19 @@ class DataARF(DataOgipResponse):
     def get_dep(self, filter=False):
         return self._rsp
 
-    def get_xlabel(self):
-        return 'Energy (keV)'
+    def get_ylabel(self) -> str:
+        """The label for the dependent axis.
 
-    def get_ylabel(self):
+        See Also
+        --------
+        set_ylabel
+
+        """
+        if self._ylabel is not None:
+            return self._ylabel
+
+        # This will pick up the latest plot backend.
+        #
         from sherpa import plot
         return 'cm' + plot.backend.get_latex_for_string('^2')
 
@@ -1076,7 +1089,15 @@ class DataRMF(DataOgipResponse):
             if nelo != self.detchans:
                 raise DataErr(f"detchans mismatch with e_min/max: {self.detchans} vs {nelo}")
 
-        Data1DInt.__init__(self, name, energ_lo, energ_hi, matrix)
+        super().__init__(name, energ_lo, energ_hi, matrix)
+
+        # special case the xlabel behaviour
+        if self.e_min is None or self.e_max is None:
+            self._xlabel = 'Channel'
+        else:
+            self._xlabel = 'Energy (keV)'
+
+        self._ylabel = 'Counts'
 
     # Although we have a Data1DInt-like dataset, the dependent axis
     # does not match the independent axes. So we have to remove the
@@ -1092,7 +1113,7 @@ class DataRMF(DataOgipResponse):
     def y(self, val):
         self._y = val
 
-    def _repr_html_(self):
+    def _repr_html_(self) -> str:
         """Return a HTML (string) representation of the RMF
         """
         return html_rmf(self)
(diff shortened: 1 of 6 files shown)


hamogu:
Without looking at the code yet, I wonder if this is feature creep? If you had asked me before implementing it, I would have said "I'm not convinced we need it, but it's also won't hurt." We won't ever replicate all matplotlib options - nor should we. For really special cases, you can get the matplotlib object (`plt.gca()`) and manipulate it (and similar for bokeh). For example, the first plot could have something like `plt.gct().axes[0].set_ylabel("Flux [10$^{-17}$ erg/cm$^2$/s/$\AA$]")`
If course, the boundary between what we offer and what we don't is a gray area. Is this worth the effort and code complexity?

Well, the effort is already spend, and @DougBurke won't get his time back if I reject this now. But can I ask what promted this addition?

sherpa/astro/data.py
```diff
@@ -935,9 +935,13 @@ def __init__(self, name, energ_lo, energ_hi, specresp, bin_lo=None,
         self._lo, self._hi = energ_lo, energ_hi
         self.energ_lo = energ_lo
         self.energ_hi = energ_hi
-        Data1DInt.__init__(self, name, energ_lo, energ_hi, specresp)
 
-    def _repr_html_(self):
+        super().__init__(name, energ_lo, energ_hi, specresp)
+
+        self._xlabel = 'Energy (keV)'
+        self._ylabel = None
```
hamogu, line 942: May I ask why you chose to set this in the `__init__` call instead of adding them as class attributes? The latter works better for inheritance IMO and is easier to read and set - you could derive a new class that changes the labels without the need to write a new `__init__` method.
Of course, this works, too, so just asking.

DougBurke:
Because I wrote the code linked to in #2083 and I wondered if we could add the ability - since we have the `get_*label` calls it didn't feel significant feature creep to add `set_*label`, but it was not as simple as I'd originally thought (e.g. the discussion I mentioned with the DataPHA behaviour).

sherpa/astro/data.py
hamogu: That's true specifically when init does not call super. Or is that the reason you did not do it? Because our class hierachy is somewhat messed up? Or is it that class attributes can't be overridden later by instance attributes because of the `NoNewAttributesAfterInit` as a base class?

DougBurke: a) I never know what the intended semantic difference is between class and object attributes for Python.

b) our inheritance here is perhaps not what you might think it is, which makes reasoning about some of this interesting - I'd have to look into it to see how much it saves here - e.g. it would hopefully avoid the need for  

https://github.com/sherpa/sherpa/pull/2086/files#diff-83c9f3e1eca305239432be420691a7cb613f4321aa77549f0e33c8ec041c23daR2096-R2107

anetasie:
@hamogu I originally use the threads to figure out how to change/add labels for multiple plots
[[](https://cxc.harvard.edu/sherpa/threads/setplot_manual/index.html#multi_plots)]

This change will make things easier.

anetasie (approved):

PR #2089, opened 2024-07-24
fit: simplify setting of file handle

DougBurke:
# Summary

Internal change to handling of the optional output file option when calling fit. There is no functional change.

# Details

There is no functional change here. The code has been changed to use a context manager to handle closing the file handle during a fit. This wasn't done originally (in #2054) because we do not always want to close the handle - e.g. if the user sends a file handle or sends in None - and so I had come up with code that works but is ugly, as noted in reviews by @hamogu in an unrelated PR that had similar behavior. I realized, in #2017, that we can use the contextlib.nullcontext handler to handle the "do not close" case, so apply that logic here.

One issue is the typing rules, as it's not clear what the best approach is for typing a context handler, so we punt on that.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -18,12 +18,13 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from contextlib import nullcontext
 from functools import wraps
 import logging
 import os
 from pathlib import Path
 import signal
-from typing import Optional, Protocol, Union, \
+from typing import Optional, Protocol, Sequence, Union, \
     runtime_checkable
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


hamogu (approved):

sherpa/fit.py
```diff
@@ -951,6 +952,43 @@ def fit(self, statfunc, pars, parmins, parmaxes,
...
+
+    if outfile is None:
+        return nullcontext(None)
+
+    if isinstance(outfile, WriteableTextFile):
+        fh = outfile
+    else:
+        # os.path.isfile and open accepts strings and Path objects
+        if not clobber and os.path.isfile(outfile):
+            raise FitErr('noclobererr', str(outfile))
```
wmclaugh, line 977: Should 'noclobererr' be 'noclobbererr'?  Not a  lien on this PR but simple a question for future clean up. I note that it is also spelled this way in sherpa/utils/err.py.

DougBurke: Yeah - I spent a bit of time trying to check I had that spelled correctly. I don't know the design decision that came up with the existing label!

PR #2091, opened 2024-07-25
Asymmetric filter plot

### LINKED ISSUE #2090: asymmetric errors - notice/ignore
I entered a black hole  trying to work with the asymmetric errors. Our documentation definitely is lacking. My bad!

I have the data file,  but somehow I'm not able to link to this issue.

1/ 'load_ascii_with_errors()'  - I can read the file with the astropy backend, but it failed in CIAO with colkeys.  I guess this might be the crates issue?

```
sherpa In [1]: load_ascii_with_errors(1,"RLQ_81_data_ascii.dat", colkeys=['Log(X-ray_Lum)','Photon_Index','Photon_Index_min','Photon_Index_ma
          ...: x'],delta=True)
IOErr: Required column 'Log(X-ray_Lum)' not found in ['col1', 'col2', 'col3', 'col4', 'col5', 'col6', 'col7']

sherpa In [2]: load_ascii_with_errors(1,"RLQ_81_data_ascii.dat", colkeys=['col5','col2','col3','col4'],delta=True)
```

Also the 'staterror' is set, but it is not clear how it is calculated, as the errors are asymmetric - is this an average of the input errors?
Is this staterror used in fitting?

```
Filter: 44.6993-45.7980 x
name      = RLQ_81_data_ascii.dat[opt colnames=none]
x         = Float64[52]
y         = Float64[52]
staterror = Float64[52]
syserror  = None
elo       = Float64[52]
ehi       = Float64[52]
```

2 'notice/ignore' seems to work, but actually fails when you try to display the data. 

```
sherpa In [49]: ignore()
dataset 1: 44.6993:45.798 x -> no data

sherpa In [50]: notice(45,46)
dataset 1: no data -> 45.4859:45.3843,45.6892:45.2895,45.9661:45.6546,45.9947,45.1834:45.8926,45.9442,45.8721:45.3682,45.5277,45.318:45.9423,45.2969:45.6138,45.6122:45.8463,45.3026:45.047,45.85:45.6701,45.798 x

sherpa In [51]: plot_data()
ValueError: 'yerr' (shape: (2, 52)) must be a scalar or a 1D or (2, n) array-like whose shape matches 'y' (shape: (28,))
```

DougBurke:
# Summary

Allow 1D data with asymmetric errors to be plotted after a filter. Fix #2090.

# Details

A minimal fix for #2090.

The asymmetric error code is not well integrated with the other data types, and so I expect there to be other areas where there could be a problem. Since we don't have many tests it's hard to find and I do not have time/desire/energy to design more tests to try to cover these issues.

There are minimal documentation fixes discussed in #2090 but I do not plan to do any more here because I do not understand the code well enough to make improvements, and I honestly have no interest in doing so. Also, looking at the code makes me question things like

  - what is the statistic that is returned (it appears to be the `LeastSquares` statistic)
  - can we change the statistic and optimiser
  - the fit is used with symmetric errors but is this obvious to the user
  - what are you meant to do with the results / how do you interpret them / what are the things you should care about

I do at least fix the minimal breakage in the `PylabErrorArea` and `BokehBackend` plot backends so that they can work with asymmetric errors.

The final commit adds a **minimal** notebook showing off the support. I honestly do not understand the logic well enough to make this anything more than minimal. You can view the notebook at https://sherpa--2091.org.readthedocs.build/en/2091/AsymmetricErrors.html


notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,577 @@
...
+    "\n",
+    "print(conf)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "01fc6b66-72b9-4cbf-84fb-455be43c366f",
+   "metadata": {},
+   "source": [
+    "We can draw a vertical rectangle on a pot to show the gamma range (from `parvals + parmins` to `parvals + parmaxes`, since `parmins < 0`). As I plan to use this more-than once I make a helper routine:"
```
anetasie, line 357: typo in line 357 'a plot` instead of `pot`

notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,577 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2a9b27dd-afb9-44c0-83cf-8564d9ac52ba",
+   "metadata": {},
+   "source": [
+    "So, **all** the above used the average error values when calculating statistic values.\n",
+    "\n",
+    "The trick to using the asymmetric errors is to use the `resample_data` call to simulate drawing the data using the asymmetric errors and fitting the resulting simulated data. In the following we explicitly ask for 1000 iterations (although that is the default value) and fix the random-number seed. The call displays information on the distribution of the fit parameters:"
```
anetasie, line 419: In order to get the uncertainties on the fitted parameters we can use 'resample_data'........'simulated data. The bootstrap simulations assume one-sided Gaussian distributions for sampling each data point with sigma equal to elo, ehi.

anetasie: Also, you can link to  resample_data help which has the description:

> 
> Resample data with asymmetric error bars.
> 
> The function performs a parametric bootstrap assuming a skewed normal distribution centered on the observed data point with the variance given by the low and high measurement errors. The function simulates niter realizations of the data and fits each realization with the assumed model to obtain the best fit parameters. The function returns the best fit parameters for each realization, and displays the average and standard deviation for each parameter.

notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,577 @@
...
+   "source": [
+    "print(result.keys())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2470998c-2cbd-47de-86fa-9309e38d5d8c",
+   "metadata": {},
+   "source": [
+    "A trace of the statistic value can be created. Unfortunately the documentation is unclear on what this statistic is, but it **appears** to be the least-squares difference between the simulated data and the best-fit model:"
```
anetasie, line 472: `resample_data` help file describes the output:

> Returns
> :
> sampled – The keys are statistic, which contains the best-fit statistic value for each iteration, samples, which contains the resampled data used in the fits as a niter by ndata array, and the free parameters in the fit, containing a NumPy array containing the fit parameter for each iteration (of size niter).
>

DougBurke: I still claim the documentation is incomplete because it just says "statistic", not what sort of statistic. I believe, from looking at the code, that it is the least-square statistic (you cna see, from the values, that it doesn't match the chi2 values from the fit).

notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,577 @@
...
+    "ui.plot_scatter(result['pl.gamma'], result['pl.ampl'], xlabel=r'$\\Gamma$', ylabel='ampl')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "70f782e9-c851-4432-bf3d-3dcc2f6c1a89",
+   "metadata": {},
+   "source": [
+    "We can compare the fit result from earlier to that from the resampled data. In this case they do not agree, which is interesting but I do not know enough about this data set to understand what this means!"
+   ]
```
anetasie, line 531: Instead of saying you don't know, say that the conf is using symmetric errors as in the fit, so no information about the asymmetric errors for the data points. `conf` and 'reg_proj' will come with a  too narrow range, while the `resample_data` function uses the information about the data.

notebooks/AsymmetricErrors.ipynb
anetasie: the fit statistics - so if you fit with leastsq it should be leastsq, if with chi2 then it should be chi2.

my old notebook used leastsq.

anetasie:
@DougBurke I run the script with the filter set and the data are filtered and fit/displayed correctly. 
Thanks for working on this. Addition of a notebook is great!

notebooks/AsymmetricErrors.ipynb
DougBurke: This was why I said in the intro that all I was doing was a minimal fix as there appears to be a number of issues we could have done better when this was originally written. The "fit to the resampled data" is done in

https://github.com/sherpa/sherpa/blob/45d2379c67e7c49b6ca6c79ffe173075ddf1d5fa/sherpa/sim/__init__.py#L970-L992

and you can see that the statistic and optimization method are hard-coded (line 975) to `LeastSq` and `LevMar`. Ideally these could be set by the user in some manner, but I don't really want to do that in this PR!

DougBurke: I've created #2093 to track this particular issue

notebooks/AsymmetricErrors.ipynb
DougBurke: I've reworked this paragraph to better describe the function. Hopefully it is a bit more useful now.

notebooks/AsymmetricErrors.ipynb
DougBurke: So, I've basically said that it's the least-squares statistic and it currently can't be changed.

notebooks/AsymmetricErrors.ipynb
DougBurke: I've reworked this paragraph a little bit, and added a little bit about using the sampled values to estimate useful information.

notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,651 @@
...
+    "ui.reg_proj(pl.gamma, pl.ampl, nloop=[101, 101])\n",
+    "add_conf_gamma()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2a9b27dd-afb9-44c0-83cf-8564d9ac52ba",
+   "metadata": {},
+   "source": [
+    "So, **all** the above used the average error values when calculating statistic values. This question is, how much different would it be if the fit could use the asymmetric errors?\n",
```
anetasie, line 423:
```suggestion
    "So, **all** the above used the average error values when calculating statistic values. The question is, how much different would it be if the fit could use the asymmetric errors?\n",
```

notebooks/AsymmetricErrors.ipynb
```diff
@@ -0,0 +1,651 @@
...
+    "ui.plot_cdf(result['pl.gamma'], xlabel=r'$\\Gamma$', name=r'$\\Gamma$')\n",
+    "add_conf_gamma()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2c9c1c69-6841-48f2-94a9-24aa8f0734a6",
+   "metadata": {},
+   "source": [
+    "We can use the distribution to estimates errors. Sherpa has useful routines for this but other modules - such as NumPy - can be used. Here we use the `get_error_estimates` routine to return the median value along with the 1-sigma range calculated from the input distribution:"
```
anetasie, line 573:
```suggestion
    "We can use the parameter distribution to estimates errors. Sherpa has useful routines for this but other modules - such as NumPy - can be used. Here we use the `get_error_estimates` routine to return the median value along with the 1-sigma range calculated from the input distribution:"
```

sherpa/sim/tests/test_asymmetric.py
```diff
@@ -125,6 +125,77 @@ def rms(a, b):
     return np.sqrt(a * a + b * b)
 
 
+# Basic data values frmo the gro*txt file. Note the data is unordered.
```
anetasie, line 128:
```suggestion
# Basic data values from the gro*txt file. Note the data is unordered.
```

anetasie (commented):
Left a few fixes to the typos

notebooks/AsymmetricErrors.ipynb
anetasie: Yes, it is better.

anetasie (approved):

sherpa/plot/bokeh_backend.py
```diff
@@ -541,10 +541,20 @@ def plot(self, x, y, *,
             objs.append(xwhisker)
 
         if yerrorbars and yerr is not None:
+            # Are the errors symmetric or not? At the moment this is
+            # decited by checking if yerr is a tuple or not.
```
wmclaugh, line 545:
```suggestion
            # decided by checking if yerr is a tuple or not.
```

sherpa/plot/pylab_area_backend.py
```diff
@@ -76,17 +76,29 @@ def plot(self, x, y, *,
...
         if not overplot:
             self.setup_plot(axes, title, xlabel, ylabel, xlog=xlog, ylog=ylog)
 
         if yerrorbars and yerr is not None:
-            axes.fill_between(x, y - yerr, y + yerr, alpha=.2,
-                              linewidth=0, color=ecolor)
-        if x is None or y is None:
-            return None
+            # Are the errors symmetric or not? At the moment this is
+            # decited by checking if yerr is a tuple or not.
```
wmclaugh, line 90:
```suggestion
            # decided by checking if yerr is a tuple or not.
```

PR #2085, opened 2024-07-19
Clean up of the IO code redux

DougBurke:
rework of #1921

PR #2094, opened 2024-07-31
Improve messages about invalid file types (PHA)

DougBurke:
# Summary

Tweak the warning messages when files can not be read to better indicate the type of file being loaded (for PHA related files).

# Details

So, in recent work I've been trying to standardize the error messages we get from the I/O backends. We can never do it completely, since different backends provide different error messages and we want to retain some of this information, but we can try to normalize it. One change I made was to change a `IOErr(str(exc))` call to something like `IOErr(f"some text including filename: {exc}")`, at least for the crates backend - see 

- #2017
- https://github.com/sherpa/sherpa/commit/4bf833423abda3d6b29a24a9049b621e46e942df

In reviewing changes seen in the CIAO regression testing, I realized that it would make more sense to add context at the caller, rather than in the low-level I/O code. So, have the backend raise `IOErr(str(exc))` and then the caller can catch this and either add context - as we do in the `sherpa.astro.xspec.read_xstable_model` call - or the warning message we create after catching the error when reading in response/background files for PHA data.

It turns out that great minds think alike, and I'd already made some of these changes in #2085.

I don't think there's a "single obviously true" approach for these calls, so we may decide to make more clean ups down the road.

# Example

So, if we copy a PHA file to a new location but do not also copy the `ANCRFILE`, `RESPFILE`, and `BACKFILE` files, we get the following warnings when we try to load the file in. For CIAO 4.16 you can guess what it is trying to load in because the files have sensible names, but with the new code we have a label (ARF, RMF, background) to go along with the error message.

I included both backends here to show that the low-level messages are different, even for the case of "file not found".

The main branch has, for crates, the odd output of 'unable to open xxx.y: File xxx.y does not exist.`, which while true, doesn't really help. I think the proposed version is better and more consistent between the two backends.

## CIAO 4.16 + crates

```
sherpa In [2]: io.backend.__name__
Out[2]: 'sherpa.astro.io.crates_backend'

sherpa In [3]: load_pha("foo.pi")
WARNING: File aciss_1_src7.arf does not exist.
WARNING: File aciss_1_src7.rmf does not exist.
WARNING: File aciss_1_bkg7.pi does not exist.
```

## CIAO 4.16 + pyfits

```
sherpa In [6]: io.backend.__name__
Out[6]: 'sherpa.astro.io.pyfits_backend'

sherpa In [7]: load_pha("foo.pi")
WARNING: file 'aciss_1_src7.arf' not found
WARNING: file 'aciss_1_src7.rmf' not found
WARNING: file 'aciss_1_bkg7.pi' not found
```

## Main branch + crates

```
>>> io.backend.name
'crates'
>>> ui.load_pha("foo.pi")
WARNING: unable to open aciss_1_src7.arf: File aciss_1_src7.arf does not exist.
WARNING: unable to open aciss_1_src7.rmf: File aciss_1_src7.rmf does not exist.
WARNING: unable to open aciss_1_bkg7.pi: File aciss_1_bkg7.pi does not exist.
```

## Main branch + pyfits

```
>>> io.backend.name
'pyfits'
>>> ui.load_pha("foo.pi")
WARNING: file 'aciss_1_src7.arf' not found
WARNING: file 'aciss_1_src7.rmf' not found
WARNING: file 'aciss_1_bkg7.pi' not found
```

## This PR + crates

```
>>> from sherpa.astro import io
>>> from sherpa.astro.io import crates_backend, pyfits_backend
>>> io.backend.name
'pyfits'
>>> io.backend = crates_backend
>>> io.backend.name
'crates'
>>> ui.load_pha("foo.pi")
WARNING: unable to read ARF: File aciss_1_src7.arf does not exist.
WARNING: unable to read RMF: File aciss_1_src7.rmf does not exist.
WARNING: unable to read background: File aciss_1_bkg7.pi does not exist.
```

## This PR + pyfits

```
>>> io.backend = pyfits_backend
>>> io.backend.name
'pyfits'
>>> ui.load_pha("foo.pi")
WARNING: unable to read ARF: file 'aciss_1_src7.arf' not found
WARNING: unable to read RMF: file 'aciss_1_src7.rmf' not found
WARNING: unable to read background: file 'aciss_1_bkg7.pi' not found
```

### DIFF
--- a/sherpa/astro/io/__init__.py
+++ b/sherpa/astro/io/__init__.py
@@ -524,7 +524,7 @@ def _read_ancillary(data: dict[str, str],
 
     except Exception as exc:
         if output_once:
-            warning(str(exc))
+            warning("unable to read %s: %s", label, str(exc))
 
     return out
 
@@ -653,7 +653,7 @@ def read_pha(arg,
 
             except Exception as exc:
                 if output_once:
-                    warning(str(exc))
+                    warning("unable to read background: %s", str(exc))
 
         for bkg_type, bscal_type in zip(('background_up', 'background_down'),
                                         ('backscup', 'backscdn')):
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -90,7 +90,7 @@ def open_crate(filename: str,
     try:
         dataset = CrateDataset(filename, mode=mode)
     except OSError as oe:
-        raise IOErr('openfailed', f"unable to open {filename}: {oe}") from oe
+        raise IOErr('openfailed', str(oe)) from oe
 
     current = dataset.get_current_crate()
     try:
@@ -505,7 +505,7 @@ def read_table_blocks(arg: Union[str, CrateDataset, TABLECrate],
         try:
             dataset = CrateDataset(arg)
         except OSError as oe:
-            raise IOErr('openfailed', f"unable to open {arg}: {oe}") from oe
+            raise IOErr('openfailed', str(oe)) from oe
 
     else:
         raise IOErr('badfile', arg, "CrateDataset obj")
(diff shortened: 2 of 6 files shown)


hamogu (approved):

DougBurke:
@hamogu - ta for the quick review

Issue #774, opened 2020-04-08
Redundant integration code 


dtnguyen2:
While reviewing PR #732, I noticed the following redundant code in the file _utils/src/integration.cc_:

```
  int integrate_1d( integrand_1d fct, void* params,
                    double xlo, double xhi,
                    unsigned int maxeval, double epsabs, double epsrel,
                    double& result, double& abserr )
  {

    if ( NULL == fct )
      return EXIT_FAILURE;

    int retval;
    gsl_function func;
    func.function = fct;
    func.params = params;

    size_t neval = size_t( maxeval );

    gsl_set_error_handler_off ();
 
    retval = gsl_integration_qng( &func, xlo, xhi, epsabs, epsrel, &result,
                                  &abserr, &neval );
    
    if (retval != EXIT_SUCCESS) {
      if( gsl_int_flag ) {
        std::cerr << "WARNING: Gauss-Kronrod integration failed "
                  << "with tolerance " << epsabs << ", trying lower tolerance..."
                  << std::endl;
        
        double tol = std::numeric_limits< float >::epsilon();
        retval = gsl_integration_qng( &func, xlo, xhi, tol, epsrel, &result,
                                      &abserr, &neval );

```
If the initial call to the function ```gsl_integration_qng``` is called with the fourth argument ```epsabs``` fails then it is called again but the fourth argument is now set to: ```std::numeric_limits< float >::epsilon()``` .

The problem is that ```integrate_1d``` is initially called with ```std::numeric_limits< float >::epsilon()``` as the value for ```epsabs```.  This can be seen when the function ```integrate_1d``` is called from the file _include/sherpa/model_extension.hh_:

```
#define TOL (std::numeric_limits< float >::epsilon());
...

template <int (*PtFunc)( const DoubleArray& p, double x, double& val )>
  int integrated_model1d( const DoubleArray& p, double xlo, double xhi,
                          double &val )
  {

    // FIXME: make these user-settable function args!
    double epsabs = TOL;
    double epsrel = 0.0;
    unsigned int maxeval = 10000;

    double abserr = 0.0;

    return integrate_1d( (integrand_1d)(integrand_model1d< PtFunc >),
                         (void*)&p, xlo, xhi, maxeval,
                         epsabs, epsrel, val, abserr );

  }
```
Moreover, I suspect the function ```py_integrate_1d``` has the same redundant issue. 

The author of PR #602, should have noticed this issue.  Oops, my bad.

DougBurke:
[...]

PR #2098, opened 2024-08-09
model: avoid a compilation warning

DougBurke:
# Summary

Avoid a compilation warning about an apparently-unused 

# Details

There can be a complaint when compiling that the
py_integrated_1d symbol is unused. I assume this is actually an issue with the compiler - in this case g++ - since it is used, but maybe the fact it's used within a template leads to a false positive.

An example (from an old build but the line number is close to being correct):

```
In file included from sherpa/astro/models/src/_modelfcts.cc:23:0:
sherpa/include/sherpa/model_extension.hh:99:14: warning: ‘int sherpa::models::py_integrated_1d(double, double, double&, FunctionWithParams<sherpa::Array<double, 12> >*, int, double, double, unsigned int, std::ostringstream&)’ defined but not used [-Wunused-function]
   static int py_integrated_1d(const double xlo, const double xhi, double &val,
              ^~~~~~~~~~~~~~~~
```

A simple solution is to just inline the function, since - to be honest - moving it into a separate routine does not make a huge difference to the reader of the code.

There is some discussion in #774 but that is really about a different issue than this, although the warning message is mentioned there.

### DIFF
--- a/sherpa/include/sherpa/model_extension.hh
+++ b/sherpa/include/sherpa/model_extension.hh
@@ -1,5 +1,5 @@
 // 
-//  Copyright (C) 2007, 2016, 2019, 2020, 2023
+//  Copyright (C) 2007, 2016, 2019, 2020, 2023, 2024
 //  Smithsonian Astrophysical Observatory
 //
 //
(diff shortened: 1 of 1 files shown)


DougBurke:
@Marie-Terrell and @harlantc  - I've labelled this for you as it's somewhat in CM's wheel-house, but feel free to un-assign yourself if it isn't a good fit.

DougBurke:
So, if we bump to gcc 15.2 or clang 21 then we don't see the warning message anymore. So the solution is to just move to a newer tool-chain (it's not worth finding out the minimum version needed).

PR #2102, opened 2024-08-15
Change the URL for XPA

### LINKED ISSUE #2100: link to xpa broken
https://github.com/sherpa/sherpa/blob/50d6bd5a79da01b2161901ba1c8ed25ad0c0e759/docs/indices.rst?plain=1#L132

(it's also broken on ds9 pages).  I'm thinking the only link left is to Eric's github page.

DougBurke:
# Summary

Change the URL used for XPA. Fix #2100.

# Details

As XPA development has stopped, the old SAO page is no longer available, so change to use the GitHub URL (less information, but it is at least valid).

### DIFF
--- a/docs/indices.rst
+++ b/docs/indices.rst
@@ -128,8 +128,7 @@ Glossary
        row on a spectrograph and the corresponding wavelength of light.
 
    XPA
-       The `XPA messaging system
-       <https://hea-www.harvard.edu/saord/xpa/>`_
+       The `XPA messaging system <https://github.com/ericmandel/xpa>`_
        is used by :term:`DS9` to communicate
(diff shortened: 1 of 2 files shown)


anetasie (changes requested):
The new XPA link works, but there is another link on line 139 to the packages goes to error page: https://packages.ubuntu.com/xenial/xpa-tools

Is there another link which would work here?

DougBurke:
Argh. So, there's no "stable" link (that I know of) for a Ubuntu package, since they want them versioned by the Ubuntu release (makes sense). So it's easiest just to remove the link.

anetasie:
I just approved this PR,

DougBurke:
Ta - Duog

PR #2103, opened 2024-08-15
Ensure crates file reads are read-only

DougBurke:
# Summary

Ensure that we read in table data as read-only when using the crates I/O backend.

# Details

I am not sure if this is a regression from #1921. It was found in a CIAO regression test but that was for a new test added in #1921.

There is also a small tweak for a docstring related to the #2086 change.


Marie-Terrell (approved):
I've run this change in a CIAO install with Sherpa. Without this change, there is one test with errors (test_read_table_blocks_sent_image).

With this change, I don't see the error, so this looks good to me! Thank you!

PR #2105, opened 2024-08-16
docs: typo fixes and remove some references.

DougBurke:
# Summary

Fix a typo in a docstring and change how two references are included in the text.

# Details

Reported by @wmclaugh when reviewing the Sherpa ahelp files, which are created from the docstring content. There were 4 problems and this should fix 3 of them; the other one is a typo in the code used to create the ahelp files.

Fix up a typo and remove two reference links which were causing an issue for the ahelp files we generated for CIAO (it's a big discussion revolving around the on-disk file conversion to RST to the format used by sphinx to the ahelp XML format, which leads to a line-break being added between "[1]" and ".", which is surprisingly hard to fix so it's easier to just avoid this in the first place by making sure we don't end a sentence on "[1]".


wmclaugh (approved):
@DougBurke - Thanks for the quick update.

PR #2108, opened 2024-08-21
Asymmetric settings

### LINKED ISSUE #2093: asymmetric errors: allow fit statistic/method to be changed by the user
In https://github.com/sherpa/sherpa/pull/2091#discussion_r1698485540 it was noted that `resample_data` / `ReSampleData` fits the per-iteration values with the least-square statistic and levmar. It would be nice to allow this to be user-controllable (from the `ui` layer it would use the set values, for the OO interface it would be set in some way by the caller).

DougBurke:
# Summary

Allow the statistic and optimizer to be changed when re-sampling data with asymmetric errors. Fix #2093.

# Details

The fix for #2093 turns out to be pretty easy. I decided that the UI layer would see no user-visible change here, but that it would now use the stat/optimiser objects. We could make the `resample_data` take optional `stat` and `method` arguments to allow the user to over-ride this, but I'm not sure it's worth it (although it would make it easier to replicate the previous behavior). Users of the OO code have direct access to this setting.

There are also some code clean ups (which shouldn't change behaviour) and both new and fixes made to the UI typing rules.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1790,9 +1790,13 @@ class Session(sherpa.ui.utils.Session):
                     # If this errors out then so be it
                     return self.unpack_ascii(filename, *args, **kwargs)
 
-    def load_ascii_with_errors(self, id, filename=None, colkeys=None, sep=' ',
-                               comment='#', func=np.average,
-                               delta=False) -> None:
+    def load_ascii_with_errors(self, id, filename=None,
+                               colkeys: Optional[Sequence[str]] = None,
+                               sep: str = ' ',
+                               comment: str = '#',
+                               func: Callable = np.average,
+                               delta: bool = False
+                               ) -> None:
         """Load an ASCII file with asymmetric errors as a data set.
 
         Create a dataset with asymmetric error bars which can be used
@@ -14406,7 +14410,9 @@ class Session(sherpa.ui.utils.Session):
 
     def resample_data(self,
                       id: Optional[IdType] = None,
-                      niter=1000, seed=None):
+                      niter: int = 1000,
+                      seed: Optional[int] = None
+                      ) -> dict[str, np.ndarray]:
         """Resample data with asymmetric error bars.
 
         The function performs a parametric bootstrap assuming a skewed
@@ -14417,6 +14423,12 @@ class Session(sherpa.ui.utils.Session):
         parameters. The function returns the best fit parameters for
         each realization, and displays the average and standard
         deviation for each parameter.
+
+        .. versionchanged:: 4.17.0
+           The resampling now uses the chosen statistic and optimizer
+           (set with set_stat and set_method). Previously the
+           least-squares statistic and Levenberg-Marquardt method were
+           always used.
 
         .. versionchanged:: 4.16.0
            The random number generation is now controlled by the
@@ -14457,6 +14469,8 @@ class Session(sherpa.ui.utils.Session):
         Account for of asymmetric errors when calculating parameter
         uncertainties:
 
+        >>> set_stat("leastsq")
+        >>> set_method("levmar")
         >>> load_ascii_with_errors(1, 'test.dat')
         >>> set_model(polynom1d.p0)
         >>> thaw(p0.c1)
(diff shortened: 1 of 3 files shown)


DougBurke:
Rebased. It removed a few commits, which is nice.

anetasie:
@DougBurke I get an error when I try to use a different statistics to fit the data (chi2gehrels or chi2xspecvar). This was for my test data sets and the staterror column is included in the data array, and I assumed it is an average of upper/lower values entered  here. I don't see any issue when I use 'chi2'.

````
load_ascii_with_errors(1,"RLQ_81_data_ascii.dat", colkeys=['Log(X-ray_Lum)','Photon_Index','Photon_Index_min','Photon_Index_max'],delta=True)
get_data().elo=-1.8*get_data().elo
#plot_data()
set_model('polynom1d.p1')
set_stat('chi2')
fit()
```


```
...
FitErr: zeros found in uncertainties, consider using calculated uncertainties..
....

print(get_data().staterror)
[0.00125009 0.02615687 0.02257724 0.01030793 0.01315787 0.0046353
 0.01975024 0.01523897 0.00373382 0.00558928 0.01499976 0.
 0.04293136 0.02545072 0.01523587 0.         0.00093758 0.03810527
 0.01339787 0.         0.00062505 0.03243021 0.01289937 0.00062505
 0.00414189 0.00250022 0.00125009 0.00250016 0.         0.05626176
 0.02933836 0.13046138 0.02049391 0.         0.0520442  0.02469929
 0.00250021 0.01227709 0.02909116 0.         0.00125011 0.02076223
 0.00250018 0.00690319 0.01662038 0.07124173 0.00125008 0.02096625
 0.01343379 0.03151682 0.03249036 0.00883878]


````

**Comment by DougBurke** [comment]
@anetasie - I was not involved with designing this interface, so I don't know what it's meant to do. There has been minimal testing of it, so I've had to guess at the code, because the documentation was poor. I have four explicit points below:

If I read in one of our test files I get

```
In [2]: load_ascii_with_errors(1, "sherpa-test-data/sherpatest/gro.txt")

In [3]: d = get_data()
```

With this we see (just looking at the first four bins):

```
In [10]: d.staterror[0:4]
Out[10]: array([1.3115, 0.943 , 1.3505, 0.7875])

In [11]: d.elo[0:4]
Out[11]: array([1.004, 0.915, 1.725, 0.757])

In [12]: d.ehi[0:4]
Out[12]: array([1.619, 0.971, 0.976, 0.818])
```

We can check whether `staterror = (elo + ehi) / 2`:

```
In [19]: check = (d.elo + d.ehi) / 2

In [20]: (d.staterror == check).all()
Out[20]: True
```

# POINT 1

So, it looks like the `elo` values are assumed to be positive. So I do not understand what you mean with the line

```
get_data().elo=-1.8*get_data().elo
```

Unless your dataset has a negative "elo" column that you need to negate.

# POINT 2

**HOWEVER**, there is the fact that the current code is not set up to recalculate the `staterror` column if you manually change `elo` or `ehi`. In the following I change elo to 1 for every element, and we can see that `staterror` has not changed.

```
In [21]: store = d.staterror.copy()

In [22]: d.elo = np.ones_like(d.ehi)

In [23]: (d.staterror == store).all()
Out[23]: True
```

So, as far as I can see

- at the moment, if you have been manaully changing the `elo` or `ehi` fields after calling `load_ascii_with_whatever_it_is_called` then you are likely to not get the answer you expect [**THIS IS A GUESS AND NEEDS TO BE CHECKED**]
- if you do need to change the columns from a file then you are going to have to create the data manually, and not use `load_ascii_blah_blah`
  - we could change the class to recalculate staterror if `elo`/`ehi` has been changed but I personally do not see it as a priority

# POINT 3

And to top it off, I honestly have to question myself whenever I compare `chi2` with `chi2XXX`, since I don't know if we say `chi2` but there is a `staterror` column set up then do we use that or just use a gaussian approximation on the counts. That is to say I don't know what is meant to happen here.

# POINT 4

None of the code you show is actually related to the code I have changed. I would expect that if you ran your code with CIAO 4.16 then you'd see the same behavior.

**Comment by DougBurke** [comment]
@anetasie - you can try

```
load_ascii_with_errors(1,"RLQ_81_data_ascii.dat", colkeys=['Log(Xray_Lum)','Photon_Index','Photon_Index_min','Photon_Index_max'],delta=True)

d = get_data()
d.elo=-1.8 * d.elo
d.staterror = (d.elo + d.ehi) / 2

set_model('polynom1d.p1')
set_stat('chi2')
fit()
```

anetasie:
@DougBurke - yes, the errors I reported are not related to the new code and the test data sets gro.txt and gro_delta.txt
give the same results in released code and this PR.

There are issues with the data sets that have negative values for `elo` or `ehi` and the current code assumes that all the input errors are positive and `staterror` are calculated when the data are loaded. The user would need to update the `staterror` separately if this is needed for some reason.  We should add this requirement to the docs.

anetasie (approved):

PR #2109, opened 2024-08-22
Add a set_xsabundances call for XSPEC users

DougBurke:
# Summary

Add a set_xsabundances call to make it easier to change one or more elemental abundances for the XSPEC models and ensure that both the get and set calls are available from the sherpa.astro.ui module.

# Details

Taken from #1615 (where the internal code is nicer) but we can provide the same API to the user with the existing code, which makes it easier to review.

It follows from the get_xsabundances call added in #2017

We could add a more-complex API - such as getting or setting individual elements, or accessing the elements by name or atomic number, but 

- I'm not sure it's worth it
- those lower-level routines are better done when directly interacting with the C++ API, which needs 

  1. #1793 
  2. #1615
  
  and I'm not sure it's worth it **at this time** 

That is, this provides the most "bang" for our "buck"....

Unfortunately, the approach taken does lead to screen output when you call `set_xsabundances`, but that would require #1615 to address (since this is screen output from the XSPEC library and not from Python scope. It's something I can live with.

# Example

In 4.16.0/1 we have (for the ui layer):

```
sherpa-4.16.0> get_xsabund()
'angr'

sherpa-4.16.0> *abund*?
get_xsabund
set_xsabund
```

With this PR we get (note that `show_xsabund` was not added in this PR but is a 4.17.0 changhe):

```

In [2]: ui.*abund*?
ui.get_xsabund
ui.get_xsabundances
ui.set_xsabund
ui.set_xsabundances
ui.show_xsabund
```

So, with

```
In [3]: ui.show_xsabund()
Solar Abundance Table:
angr
  H : 1.000e+00  He: 9.770e-02  Li: 1.450e-11  Be: 1.410e-11  B : 3.980e-10
  C : 3.630e-04  N : 1.120e-04  O : 8.510e-04  F : 3.630e-08  Ne: 1.230e-04
  Na: 2.140e-06  Mg: 3.800e-05  Al: 2.950e-06  Si: 3.550e-05  P : 2.820e-07
  S : 1.620e-05  Cl: 3.160e-07  Ar: 3.630e-06  K : 1.320e-07  Ca: 2.290e-06
  Sc: 1.260e-09  Ti: 9.770e-08  V : 1.000e-08  Cr: 4.680e-07  Mn: 2.450e-07
  Fe: 4.680e-05  Co: 8.320e-08  Ni: 1.780e-06  Cu: 1.620e-08  Zn: 3.980e-08
```

we now can

```
In [5]: avals = ui.get_xsabundances()

In [6]: type(avals)
Out[6]: dict

In [7]: print(avals)
{'H': 1.0, 'He': 0.09769999980926514, 'Li': 1.4500000158901294e-11, 'Be': 1.409999981355492e-11, 'B': 3.979999940728618e-10, 'C': 0.00036299999919719994, 'N': 0.00011200000153621659, 'O': 0.0008510000188834965, 'F': 3.630000122711863e-08, 'Ne': 0.0001230000052601099, 'Na': 2.1400001060101204e-06, 'Mg': 3.7999998312443495e-05, 'Al': 2.9499999527615728e-06, 'Si': 3.550000110408291e-05, 'P': 2.8200000201650255e-07, 'S': 1.6199999663513154e-05, 'Cl': 3.160000119351025e-07, 'Ar': 3.6300000374467345e-06, 'K': 1.3199999671087426e-07, 'Ca': 2.2900001113157487e-06, 'Sc': 1.259999993230565e-09, 'Ti': 9.770000275466373e-08, 'V': 9.99999993922529e-09, 'Cr': 4.679999960899295e-07, 'Mn': 2.4499999540239514e-07, 'Fe': 4.6799999836366624e-05, 'Co': 8.319999977857151e-08, 'Ni': 1.7800000478018774e-06, 'Cu': 1.619999956403717e-08, 'Zn': 3.979999974035309e-08}
```

and even

```
In [8]: avals['Li'] = 0; avals['Cu'] = 0

In [9]: ui.set_xsabundances(avals)
 Solar Abundance Vector set to file:  User defined abundance vector / no description specified

In [10]: ui.get_xsabund()
Out[10]: 'file'
```

after which we can see the changes in `show_xsabund`:

```
In [11]: ui.show_xsabund()
Solar Abundance Table:
file
  H : 1.000e+00  He: 9.770e-02  Li: 0.000e+00  Be: 1.410e-11  B : 3.980e-10
  C : 3.630e-04  N : 1.120e-04  O : 8.510e-04  F : 3.630e-08  Ne: 1.230e-04
  Na: 2.140e-06  Mg: 3.800e-05  Al: 2.950e-06  Si: 3.550e-05  P : 2.820e-07
  S : 1.620e-05  Cl: 3.160e-07  Ar: 3.630e-06  K : 1.320e-07  Ca: 2.290e-06
  Sc: 1.260e-09  Ti: 9.770e-08  V : 1.000e-08  Cr: 4.680e-07  Mn: 2.450e-07
  Fe: 4.680e-05  Co: 8.320e-08  Ni: 1.780e-06  Cu: 0.000e+00  Zn: 3.980e-08
```

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -102,8 +102,10 @@ References
 """
 
 
+from contextlib import suppress
 import logging
 import string
+import tempfile
 from typing import Any, Optional, Union
 import warnings
 
@@ -150,15 +152,15 @@ def get_xsabund(element: Optional[str] = None) -> Union[str, float]:
     Returns
     -------
     val : str or float
-       When `element` is `None`, the abundance table name is
-       returned (see `set_xsabund`); the string 'file' is used
-       when the abundances were read from a file. A numeric value
-       is returned when an element name is given. This value is the
+       When `element` is `None`, the abundance table name is returned
+       (see `set_xsabund`); the string 'file' is used when the
+       abundances were read from a file or vector. A numeric value is
+       returned when an element name is given. This value is the
        elemental abundance relative to H.
 
     See Also
     --------
-    get_xsabundances, set_xsabund
+    get_xsabundances, set_xsabund, set_xsabundances
 
     Examples
     --------
@@ -203,7 +205,7 @@ def get_xsabundances() -> dict[str, float]:
 
     See Also
     --------
-    get_xsabund, set_xsabund
+    get_xsabund, set_xsabund, set_xsabundances
 
     Examples
     --------
@@ -215,6 +217,67 @@ def get_xsabundances() -> dict[str, float]:
 
     return {name: _xspec.get_xsabund(name)
             for name in get_xselements().keys()}
+
+
+# The current interface to the XSPEC model library makes this awkward
+# to write. Once the interface switches to using the FunctionUtility
+# interface this code will be a lot simpler iternally, without
+# changing the API.
+#
+def set_xsabundances(abundances: dict[str, float]) -> None:
+    """Set the abundances used by X-Spec.
+
+    .. versionadded:: 4.17.0
+
+    Parameters
+    ----------
+    abundances : dict
+        The new set of abundances. The keys are the element name
+        (e.g. 'Fe') and the values are the abundances. Any element
+        that is not present is set to 0.0, and other keywords are
+        ignored.
+
+    See Also
+    --------
+    get_xsabund, get_xsabundances, set_xsabund
+
+    Examples
+    --------
+
+    Change the relative abundance of Chromium:
+
+    >>> abunds = get_xsabundances()
+    >>> abunds['Cr'] = 4.2e-6
+    >>> set_xsabundances(abunds)
+
+    """
+
+    # Get the list of elemental values.
+    #
+    elems = get_xselements()
+    out = np.zeros(len(elems))
+    for name, abund in abundances.items():
+        # Skip unknown names. Should this warn the user?
+        with suppress(KeyError):
+            z = elems[name]
+            out[z - 1] = abund
+
+    # Write them to a file and then load them.
+    #
+    # This should set delete_on_close=False but that
+    # requires Python 3.12.
+    #
+    with tempfile.NamedTemporaryFile(encoding="ascii", mode="w",
+                                     delete=False) as fh:
+        for elem in out:
+            fh.write(f"{elem}\n")
+
+        # Make sure data is written to disk. Could this just
+        # be fh.flush() and avoid closing the file?
+        #
+        fh.close()
+
+        set_xsabund(fh.name)
 
 
 # This function is not added to __all__ as it is very specialized.
@@ -365,7 +428,7 @@ def set_xsabund(abundance: str) -> None:
 
     See Also
     --------
-    get_xsabund, get_xsversion, set_xschatter
+    get_xsabund, get_xsversion, set_xsabundances, set_xschatter
 
     Notes
     -----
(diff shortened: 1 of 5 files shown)


sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -363,6 +365,81 @@ def test_abund_get_dict():
...
+
+    We only set a sub-set of elements, with the rest being set to 0.
+
+    """
+
+    from sherpa.astro import xspec
+
+    oval = xspec.get_xsabund()
+
+    # This is a pre-condition. It is't really needed, but it is a better
```
hamogu, line 380:
```suggestion
    # This is a pre-condition. It isn't really needed, but it is a better
```

sherpa/astro/xspec/__init__.py
```diff
@@ -217,6 +219,67 @@ def get_xsabundances() -> dict[str, float]:
...
+    >>> set_xsabundances(abunds)
+
+    """
+
+    # Get the list of elemental values.
+    #
+    elems = get_xselements()
+    out = np.zeros(len(elems))
+    for name, abund in abundances.items():
+        # Skip unknown names. Should this warn the user?
```
hamogu, line 260: As discussed in other PRs, I feel that a better approach is to warn or error out. Typos do happen or people might not remember what elements are or are not in the XSPEC list. If my command returns without error or warning, then I would assume that it worked. 
This is new API and I can't see a case where it's beneficial to fail silently here.

hamogu (changes requested):

sherpa/astro/xspec/__init__.py
DougBurke: There's definitely an argument to be said here that a user who has set 'ar' to 2e-5 should have it pointed out that they should have used 'Ar'.

Now, we could make the keys case insensitive to fix this particular issue,  but we would still want to let a user know that we have ignored their Zirconium setting.

The advantage of warning (to the user): I can send in any-old dictionary and just have the elements I want processed.

The advantage of an error (to me): I can just error out at the first unknown value and so do not have to be concerned about generating many warning messages.

For this particular use case I think error-ing out is better - the user can just ensure that the input is a correctly-labelled dictionary. 

The one thing that would give me pause (ie prefer a warning) was if we thought the number of elements supported by XSPEC were to go up or down in the future. I don't see that likely...

hamogu: If the past is any guidance, XSPEC does almost never change the elements. And even then, I suspect that the most common user scenario is to get the dict using `avals=ui.get_xsabundances()`, modify a few numbers in `avals` and then set is back - and that will work even with a larger set of abundances. So let's make this an error, not a warning.

hamogu (approved):

sherpa/astro/xspec/__init__.py
```diff
@@ -217,6 +219,71 @@ def get_xsabundances() -> dict[str, float]:
             for name in get_xselements().keys()}
 
 
+# The current interface to the XSPEC model library makes this awkward
+# to write. Once the interface switches to using the FunctionUtility
+# interface this code will be a lot simpler iternally, without
```
wmclaugh, line 224:
```suggestion
# interface this code will be a lot simpler internally, without
```

DougBurke:
Rebased to address conflict from merging

- #1793

The conflict was in a test and concerned the ordering of tests - i.e. nothing structural.

Issue #2113, opened 2024-08-24
delete_data does not remove the associated model expression


DougBurke:
Do we want `set_source` / `get_source` to be tied to the dataset, since at present they are not

```
>>> from sherpa import ui
WARNING: imaging routines will not be available, 
failed to import sherpa.image.ds9_backend due to 
'RuntimeErr: DS9Win unusable: Could not find ds9 on your PATH'
>>> ui.load_arrays(1, [1, 2], [10, 12]) 
>>> ui.set_source(ui.const1d.bob)
>>> ui.get_source()
<Const1D model instance 'const1d.bob'>
>>> ui.delete_data()
```

At this point we can not call `get_data`, which makes sense:

```
>>> ui.get_data()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 1, in get_data
  File "/home/dburke/sherpa/sherpa-conda/sherpa/ui/utils.py", line 3031, in get_data
    return self._get_item(id, self._data, 'data set', 'has not been set')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dburke/sherpa/sherpa-conda/sherpa/ui/utils.py", line 2069, in _get_item
    raise IdentifierErr('getitem', itemdesc, id, errdesc)
sherpa.utils.err.IdentifierErr: data set 1 has not been set
```

However, we still can call `get_source`.

```
>>> ui.get_source()
<Const1D model instance 'const1d.bob'>
```

We *could* make `delete_data` remove the model expression associated with the dataset identifier.

I do note that we can call `get_source` before a dataset is created:

```
>>> ui.clean()
>>> ui.set_source(ui.polynom1d.foo)
>>> ui.get_source()
<Polynom1D model instance 'polynom1d.foo'>
>>> ui.get_data()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 1, in get_data
  File "/home/dburke/sherpa/sherpa-conda/sherpa/ui/utils.py", line 3031, in get_data
    return self._get_item(id, self._data, 'data set', 'has not been set')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2114, opened 2024-08-24
Add delete_bkg command - fix #891

### LINKED ISSUE #891: Add delete_bkg to sherpa.astro.ui.utils
We don't provide an easy way to delete a background component of a DataPHA dataset in the sherpa.astro.ui layer. It would be easy to add a delete_bkg routine (also making sure to delete related bkg models).

DougBurke:
# Summary

Add the delete_bkg call. Fix #891.

# Details

This originally had a lot of clean-up documention/typing work which got moved into #2124, hence some of the dicsussion below in the comments does not match the current code.

We add

- `delete_bkg` call to the UI layer
- `get_backgrounds` to the `DataPHA` object, in an attempt to "better type" the code (e.g. to avoid some cases where we know a value is not None because the identifier exists but the code does not)

One question that came up from this work was
- #2113

which I do not plan to address here.

# Issues to resolve

- this code follows the existing `delete_data` behavior and does not error out if a dataset has no backgrounds (it's a little more complicated than this, but this is the main issue) and the question is should it error out
- our code essentially mimics a dictionary (since it does) so maybe we should make this the primary API rather than existing calls
- the current code actually lets you (I **think**) temporarily remove a background but leave it still associated with the parent `DataPHA` whereas the new code does not; this strikes me as a subtle disaster waiting to strike

# Example

```
>>> ui.load_pha("sherpa-test-data/sherpatest/3c273.pi")
>>> ui.subtract()
```

So, what happens if we delete the background?

```
>>> ui.delete_bkg()
dataset 1: background subtraction has been removed
```

And we can no-longer get it:

```
>>> ui.get_bkg()
  File "<string>", line 1, in get_bkg
  File "/home/dburke/sherpa/sherpa-conda/sherpa/astro/ui/utils.py", line 7010, in get_bkg
    raise IdentifierErr('getitem', 'background data set',
sherpa.utils.err.IdentifierErr: background data set 1 in PHA data set 1 has not been set
```

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -1591,8 +1591,7 @@ class DataPHA(Data1D):
         else:
             raise DataErr('bad', 'quantity', val)
 
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             if bkg.get_response() != (None, None) or \
                (bkg.bin_lo is not None and bkg.bin_hi is not None):
                 bkg.units = units
@@ -1607,8 +1606,8 @@ class DataPHA(Data1D):
 
     def _set_rate(self, val):
         self._rate = bool_cast(val)
-        for bkg_id in self.background_ids:
-            self.get_background(bkg_id).rate = val
+        for _, bkg in self.get_backgrounds():
+            bkg.rate = val
 
     rate = property(_get_rate, _set_rate,
                     doc="""Is the Y axis displayed as a rate when plotting data?
@@ -1637,8 +1636,8 @@ a rate.""")
             raise DataErr("bad", "plot_fac setting", val)
 
         self._plot_fac = ival
-        for bkg_id in self.background_ids:
-            self.get_background(bkg_id).plot_fac = ival
+        for _, bkg in self.get_backgrounds():
+            bkg.plot_fac = ival
 
     plot_fac = property(_get_plot_fac, _set_plot_fac,
                         doc="""How the X axis is used to create the Y axis when plotting data.
@@ -1668,6 +1667,9 @@ If set, the identifiers must already exist, and any other responses
 will be removed. The identifiers can be integers or strings.
 """)
 
+    # TODO: why not just use the _backgrounds dict directly rather
+    # than keep this information separate?
+    #
     def _get_background_ids(self) -> list[IdType]:
         return self._background_ids
 
@@ -1675,6 +1677,9 @@ will be removed. The identifiers can be integers or strings.
         if not numpy.iterable(ids):
             raise DataErr('idsnotarray', 'background', str(ids))
 
+        # TODO: we should not have invalid keys in this dictionary
+        # by construction.
+        #
         keys = list(self._backgrounds.keys())
         for bkg_id in ids:
             if bkg_id not in keys:
@@ -2716,11 +2721,30 @@ It is an integer or string.
 
         See Also
         --------
-        delete_background, set_background
+        delete_background, get_backgrounds, set_background
 
         """
         bkg_id = self._fix_background_id(id)
         return self._backgrounds.get(bkg_id)
+
+    def get_backgrounds(self) -> list[tuple[IdType, DataPHA]]:
+        """Return all the background components.
+
+        .. versionadded:: 4.17.0
+
+        Returns
+        -------
+        bkg : list of (int or str, sherpa.astro.data.DataPHA)
+           All background identifiers and datasets associated with
+           this PHA. The list will be empty if there is no associated
+           background.
+
+        See Also
+        --------
+        delete_background, get_background, set_background
+
+        """
+        return list(self._backgrounds.items())
 
     def set_background(self,
                        bkg: DataPHA,
@@ -2854,11 +2878,10 @@ It is an integer or string.
         if bkg_id not in self.background_ids:
             return
 
+        self.background_ids.remove(bkg_id)
         self._backgrounds.pop(bkg_id, None)
         if len(self._backgrounds) == 0:
             self._subtracted = False
-
-        self.background_ids.remove(bkg_id)
 
     def get_background_scale(self,
                              bkg_id: IdType = 1,
@@ -2917,12 +2940,13 @@ It is an integer or string.
         if units not in ['counts', 'rate']:
             raise ValueError(f"Invalid units argument: {units}")
 
-        if bkg_id not in self.background_ids:
-            return None
+        bkg_pha = self.get_background(bkg_id)
+        if bkg_pha is None:
+            return
 
         nbkg = len(self.background_ids)
 
-        def correct(obj):
+        def correct(obj: DataPHA) -> Union[float, numpy.ndarray]:
             """Correction factor for the object"""
             ans = 1.0
 
@@ -2940,11 +2964,15 @@ It is an integer or string.
             return ans
 
         src = correct(self)
-        bkg = correct(self.get_background(bkg_id))
+        bkg = correct(bkg_pha)
         scale = src / bkg / nbkg
         return self._check_scale(scale, group=group, filter=filter)
 
-    def _check_scale(self, scale, group=True, filter=False):
+    def _check_scale(self,
+                     scale: Union[float, numpy.ndarray],
+                     group: bool = True,
+                     filter: bool = False
+                     ) -> Union[float, numpy.ndarray]:
         """Ensure the scale value is positive and filtered/grouped.
 
         Parameters
@@ -3511,8 +3539,7 @@ It is an integer or string.
         the quality value for these channels will be set to 2.
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_bins(num, tabStops=tabStops)
 
         self._dynamic_group("grpNumBins", len(self.channel), num,
@@ -3559,8 +3586,7 @@ It is an integer or string.
         for these channels will be set to 2.
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_width(val, tabStops=tabStops)
 
         self._dynamic_group("grpBinWidth", len(self.channel), val,
@@ -3629,8 +3655,7 @@ It is an integer or string.
         >>> pha.group_counts(20, tabStops=[0] * pha.size)
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_counts(num, maxLength=maxLength, tabStops=tabStops)
 
         self._dynamic_group("grpNumCounts", self.counts, num,
@@ -3688,8 +3713,7 @@ It is an integer or string.
         quality value for these channels will be set to 2.
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_snr(snr, maxLength=maxLength, tabStops=tabStops,
                           errorCol=errorCol)
 
@@ -3744,8 +3768,7 @@ It is an integer or string.
         quality value for these channels will be set to 2.
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_adapt(minimum, maxLength=maxLength,
                             tabStops=tabStops)
 
@@ -3808,8 +3831,7 @@ It is an integer or string.
         quality value for these channels will be set to 2.
 
         """
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             bkg.group_adapt_snr(minimum, maxLength=maxLength,
                                 tabStops=tabStops, errorCol=errorCol)
 
@@ -3875,8 +3897,7 @@ It is an integer or string.
 
         bdata_list = []
 
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for bkg_id, bkg in self.get_backgrounds():
             bdata = get_bdata_func(bkg_id, bkg)
 
             backscal = bkg.backscal
@@ -4102,8 +4123,7 @@ It is an integer or string.
         bkg_counts = []
         bkg_scales = []
 
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
+        for _, bkg in self.get_backgrounds():
             berr, bcounts = get_error(bkg)
             if berr is None:
                 # We do not know how to generate an error, so
@@ -4865,7 +4885,7 @@ It is an integer or string.
         for bid in bkg_ids:
             bkg = self.get_background(bid)
             if bkg is None:
-                # Skip any missing identifier
+                # Skip any missing identifier. Should this error out?
                 continue
 
             old_bkg_units = bkg.units
@@ -4979,11 +4999,7 @@ It is an integer or string.
 
         # Ensure any backgrounds are also grouped.
         #
-        for bkg_id in self.background_ids:
-            bkg = self.get_background(bkg_id)
-            if bkg is None:
-                raise DataErr(f"Internal error: unknown background id {bkg_id}")
-
+        for _, bkg in self.get_backgrounds():
             try:
                 bkg.grouped = True
             except DataErr as exc:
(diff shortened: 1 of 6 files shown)


sherpa/astro/data.py
```diff
@@ -4645,7 +4700,10 @@ def notice_response(self, notice_resp=True, noticed_chans=None):
             arf, rmf = self.get_response(resp_id)
             _notice_resp(noticed_chans, arf, rmf)
 
-    def notice(self, lo=None, hi=None, ignore=False, bkg_id=None):
+    def notice(self, lo=None, hi=None,
```
hamogu, line 4703: Is there a reason that there is no type hint for "lo" and "hi"? 
If we type a function, I would prefer to type all parameters (except in special cases where it's just too hard to do)

sherpa/astro/data.py
```diff
@@ -4807,6 +4865,10 @@ def notice(self, lo=None, hi=None, ignore=False, bkg_id=None):
         #
         for bid in bkg_ids:
             bkg = self.get_background(bid)
+            if bkg is None:
```
hamogu, line 4868: Maybe this is OK, but it look odd to me. `bkg_ids` is returned from `self.background_ids` and I would expect every background to have an identifier. Or is there a situation where that's not the case?`

hamogu: Or is this for the case when no background it set at all? But then the comment in the line below is confusing.

sherpa/astro/ui/tests/test_astro_ui_unit.py
```diff
@@ -3168,6 +3168,117 @@ def test_set_bkg_not_a_background(clean_astro_ui):
...
+    This just checks the code does not error out.
+    """
+
+    ui.delete_bkg(id, bkg_id)
+
+
+@pytest.mark.parametrize("id", [None, 1, "x"])
+@pytest.mark.parametrize("bkg_id", [None, 1, "y"])
+def test_delete_bkg_no_bkg(id, bkg_id, clean_astro_ui):
+    """Check we can delete a background when nothing exists.
```
hamogu, line 3185: I'm not happy with this behaviour. Do we need it for backwards compatibility? I'm worried about silent errors, e.g. I have several datasets and I want to remove the background for one of them. I mistype the id, and don't get an error, so I assume it all worked, when in fact it didn't do anything. 
I would prefer to raise an error when a user tries to remove a background that does not exist.

sherpa/ui/utils.py
```diff
@@ -4069,6 +4069,8 @@ def delete_data(self, id: Optional[IdType] = None) -> None:
         -----
         The source expression is not removed by this function.
 
+        The routine does nothing if the given dataset does not exist.
+
```
hamogu, line 4073: I know it's just a doc change, so this is not a request to change anything, but I general I think it's not good practice to fail silently.

sherpa/astro/ui/utils.py
```diff
@@ -174,19 +179,25 @@ def _check_pha_tabstops(data: DataPHA,
     return np.zeros(data.size)
 
 
-def _save_errorcol(session, idval, filename, bkg_id,
-                   clobber, asciiflag,
-                   get_err, colname):
+def _save_errorcol(session,
+                   idval: Optional[IdType],
+                   filename,
```
hamogu, line 184: Do we support pathlib? If not (or we don't know for sure) we could just type it as string for now.

```suggestion
                   filename: str,
```

sherpa/astro/ui/utils.py
```diff
@@ -654,7 +677,12 @@ def _get_show_bkg_source(self, id=None, bkg_id=None):
 
         return model_str
 
-    def show_bkg(self, id=None, bkg_id=None, outfile=None, clobber=False):
+    def show_bkg(self,
+                 id: Optional[IdType] = None,
+                 bkg_id: Optional[IdType] = None,
+                 outfile=None,
```
hamogu, line 683:
```suggestion
                 outfile: str = None,
```

hamogu: Same below. I'm sure @DougBurke had a good reason to not type this, so I want to discuss here instead of making suggestions for all of them at this point.

sherpa/astro/ui/utils.py
```diff
@@ -2525,7 +2577,9 @@ def load_filter(self, id, filename=None, bkg_id=None, ignore=False,
     # DOC-TODO: prob. needs a review as the existing ahelp documentation
     # talks about 2 cols, but experimentation suggests 1 col.
     #
-    def load_grouping(self, id, filename=None, bkg_id=None, *args, **kwargs):
+    def load_grouping(self, id, filename=None,
```
hamogu, line 2580: Inconsisteny with the doctring, which says filename is a str.

hamogu: Same happens below in different locations where the other parameters of a function are typed in this PR. Again, I want to discuss here and not mark all instances at this stage.

sherpa/astro/ui/utils.py
```diff
@@ -3154,7 +3227,9 @@ def set_exposure(self, id, exptime=None, bkg_id=None):
         d = self._get_pha_data(id, bkg_id)
         d.exposure = exptime
 
-    def set_backscal(self, id, backscale=None, bkg_id=None):
+    def set_backscal(self, id, backscale=None,
```
hamogu, line 3230: Is there a reason to just partially type this function?

hamogu: Same below, but again, I'm not marking all of them.

sherpa/astro/ui/utils.py
```diff
@@ -3491,7 +3580,10 @@ def get_error(self, id=None, filter=False, bkg_id=None):
         return d.get_error(filter, self.get_stat().calc_staterror)
 
     # DOC-NOTE: also in sherpa.utils
-    def get_indep(self, id=None, filter=False, bkg_id=None):
+    def get_indep(self,
+                  id: Optional[IdType] = None,
+                  filter: bool = False,
+                  bkg_id: Optional[IdType] = None):
```
hamogu, line 3586: Should the return value be typed? (Same for a few other modified functions below)

sherpa/astro/ui/utils.py
```diff
@@ -6894,7 +7080,80 @@ def set_bkg(self, id, bkg=None, bkg_id=None):
...
+        See Also
+        --------
+        clean, delete_bkg_model, list_bkg_ids, set_bkg
+
+        Notes
+        -----
+        The source expression for the background is not removed by
+        this function.
+
+        The routine does nothing if the given dataset or background
```
hamogu, line 7114: I would prefer to error out in both cases.

hamogu (changes requested):
Most of the changes are good. I have two major suggestions. I can be convinced otherwise, but I ask to consider both:

1. I think it's dangerous to fail silently, e.g. to not issue an error i I try to remove a background component that does not exist. I know some parts of Sherpa work like that and we don't want to break backwards compatibility, but I'm worried about e.g. `deletete_bgk('src', 'bgk1') not issuing an error. Users will assume that their background `bkg1` has been deleted, and not double check if they had a typo in the string.
2. Typing is optional and is a pain to do. Yet, if we type a function, I would prefer to type all parameters and the return value unless it's really hard (i.e. type at least all the things that the ints or strings).

sherpa/astro/ui/utils.py
DougBurke: Now, I did work on this in a recent PR 

- #2054 
 
 and came up with

https://github.com/sherpa/sherpa/blob/307ddbbd4b7ff98e3c9720915f05aa530f4a1262/sherpa/fit.py#L570-L579

so maybe I can use that.

sherpa/astro/ui/utils.py
DougBurke: I point your attention to the "Notes" section of this (and many other routines in this module):

https://github.com/sherpa/sherpa/blob/307ddbbd4b7ff98e3c9720915f05aa530f4a1262/sherpa/astro/ui/utils.py#L2528-L2576

I have been worrying about how to type these, or whether we even can, for a long time now. However, that was before I knew about [typing.overload](https://docs.python.org/3/library/typing.html#overload). I now think we could use that approach, but

- I have not tied it
- it would be massively invasive, so I don't think now is the time
- it looks like the semantics of `overload` is not entirely resolved - see the [discourse discussion](https://discuss.python.org/t/draft-typing-spec-chapter-for-overloads/60788) - which is always fun to see (although I'm not sure we are likely to hit on these areas)

sherpa/astro/ui/utils.py
DougBurke: Because I do not know the full types that should be used (is it `float`, is it `float | Sequence[float]`, is it `float |  Sequence[float] | np.ndarray`, or ...). 

Typing  of these routines is "easier" once we have the classes better typed (since they define the types).

Here I was focussing on just adding types for "identifiers" (but, as noted above, I got a bit carried away). I htink I'm goong to remove the other types as they are causing too much discussion.

sherpa/astro/ui/utils.py
DougBurke: See https://github.com/sherpa/sherpa/pull/2114#discussion_r1732781845

sherpa/astro/ui/utils.py
DougBurke: But not in this PR.

sherpa/astro/data.py
```diff
@@ -1556,10 +1566,10 @@ def _set_subtracted(self, val):
     subtracted = property(_get_subtracted, _set_subtracted,
                           doc='Are the background data subtracted?')
 
-    def _get_units(self):
+    def _get_units(self) -> AnalysisType:
         return self._units
 
-    def _set_units(self, val):
+    def _set_units(self, val: str) -> None:
```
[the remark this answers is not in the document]
DougBurke: See https://github.com/sherpa/sherpa/pull/2114#discussion_r1732781845

sherpa/astro/data.py
DougBurke: So this is a case where I claim our API doesn't help us. We should have a way to be able to loop through the backgrounds, which would mean we don't need the `if bkg is None` statement to please the type checker

At the moment, since `get_background` returns `DataPHA | None` . the type checker has no way to know that `get_background` can not return `None` here.

I'm not sure the best way to square the circle

- live with the old code, and annoy the type checker gods
- annoy readers with pointless code
- address the API issues, which is just not a good thing to be doing at this stage of the release cycle (well, not when I have other changes I want to get in ;-)

sherpa/astro/ui/utils.py
DougBurke: So, general pathlib support is 

- #1070

and I'd hoped to get to it, but am not sure I will for 4.17.0.

However, this is covered by a combination of 

- https://github.com/sherpa/sherpa/pull/2114#discussion_r1732743640
- https://github.com/sherpa/sherpa/pull/2114#discussion_r1732781845

DougBurke:
So, I'm basically yeeting all the "add types" comments into the void - or is it the abyss you yeet stuff into? - and will remove the extra types.

So, the main discussion is over whether something should be an error or ignored if "it isn't a valid operation".

sherpa/astro/ui/utils.py
DougBurke: I have a complex set of feelings around this

- I do appreciate the "error out" approach
- right up until I process data and write code where I have to explicitly catch the failure (or manually add validation checks) just because for this case the condition wasn't met. An example is calling "subtract" - sometimes it makes sense to error out and say "but you are already subtracted", but in other cases it can be okay to say "oh, you're already subtracted, I don't need to do anything".

I claim there's likely a range of approaches and it depends in part on whether (for Sherpa) it is a class (or low-level routine) or whether it's in the UI layer.
 
These are all generic points, and we need to think about the specifics of this. In this case I am just trying to follow the existing `delete_data` behaviour (kind of, as there are failure cases in the background case we don't have for the source case).

sherpa/astro/data.py
hamogu: For some reason, I've never managed to get a type checker to work with my editor. I know many people do and it should work, but something is off and I never bothered to spend the time to track it down. So I'm fine with annoying the type checker gods ... but you do much more work in the codebase than I do and the typecheckers did help us (more precisely: you) find a few useful things in the past. So, if you prefer "pointless code" that's also fine with me - just click "resolve the conversation"

sherpa/astro/ui/utils.py
hamogu: OK.

sherpa/astro/ui/utils.py
hamogu: Oh, I hate these "designed for easy interactive use". It's easy - unless you want to see what you need to put in. But we're not going to redesign the UI because I don't like it.
OK, I'll live with it.

sherpa/astro/ui/utils.py
hamogu: I believe I was the one who insisted the `subtract` should not give me an error if it's already subtracted because I run into cases where some of my datasets were already subtracted and some are not and then it failed in the middle (#766). So, point taken.

DougBurke:
Rebased to address a simple conflict, from #2112, and added a commit to fix some of the typing rules I've added. There is also the addition of a `get_backgrounds` method which I added after a discussion with @hamogu over logic like

    for bkg_id in foo.background_ids:
        bkg = foo.get_background(bkg_id)
        # we know bkg is not None but the computer does not

so we can now replace it with

    for bkg_id, bkg in foo.get_backgrounds():
        # yay, we know bkg is a DataPHA

sherpa/astro/data.py
```diff
@@ -2716,12 +2721,31 @@ def get_background(self,
...
+           All background identifiers and datasets associated with
+           this PHA. The list will be empty if there is no associated
+           background.
+
+        See Also
+        --------
+        delete_background, get_background, set_background
+
+        """
+        return list(self._backgrounds.items())
```
hamogu: I like this better than the previous use. However, I still wonder if this level of indirection is needed . This is a one-liner. Why not just use `data._background.items()` [note below] directly inline? (While they might be some special cases, where it needs to be used as a list, usually it's just iterated over and then `.items()` is fine - that's interable.

Obviously, this won't hurt performance or anything. I just personally find code that uses too many levels of indirection hard to read. I know that "In computer science, every problem can be solved with one more level of indirection" and there are valid use cases, but it seems overly complicated.

note: If that's the recommended patter, it should probably be made part of the official API and be called `backgrounds` instead of `_backgrounds`.

To be clear, I'm not going to veto this PR over this -- it's a valid design decision -- I'm just pointing out what I personally would prefer.

hamogu: Thinking this further- do we need `self.background_ids`? Isn't that just `self._background.keys()`? Can be simplify a lot more by just making this dict non-private?

hamogu: Last thought before I move to the next commit: If the pattern is just `data._background.items()` then we also don't need any extra tests that show that we really get back what we expect by calling the "get me the list of backgrounds" function because `data._background.items()`  is just Python and not Sherpa functionality - we can rely on the fact that `keys()` or `items()` works on a dict.

DougBurke: So, `background_ids` is interestingly implemented, but it was designed a long time ago when we didn't have the certainty we now have (e.g. that ordering of keys is deterministic now). I wouldn't want to remove `background_ids` since it is an existing API and I don't see any benefit to removing it, but it could be cleaned up (but I'd rather do that in a later PR unless I gain a huge amount of enthusiasm and time).

For background information: this API was also designed to be somewhat language agnostic (as it had to allow working with S-Lang as well as Python) in which case having an explicit API rather than just relying on the Python dict interface was almost a necessity. We are obviously long gone from those requirements, but the old API still exists and I am not convinced of the use of replacing the API (as I don't want to break existing code, including within Sherpa itself), and is it worth having two APIs (explicit `get`/`set`/`delete` calls versus "use dictionary methods)?

Potential advantages to the indirection that the current design has is that we hide the internal implementation and we can check that the keys to the dict are IdType and the values are DataPHA objects when inserting them. We could instead overload the `backgrounds` property to allow this (but it gets a bit messy).

sherpa/astro/ui/tests/test_astro_ui_unit.py
DougBurke: This mirrors `delete_data`:

```
>>> ui.delete_data()
>>> ui.delete_data(1)
>>> ui.delete_data("x")
```

sherpa/astro/data.py
DougBurke: So, with the current setup we can *technically* add a background to a dataset and then "remove" it - by setting the `background_ids` to an array without this identifier - but still have the background associated with the dataset.

I guess this would let you quickly answer the question "how much of a difference does background component xxx make", but I'm not sure it makes that much sense. This does, however, suggest that my `get_backgrounds` helper could break existing code (since it doesn't care about the `background_ids` field). It could be made to do so. Need to think.

DougBurke:
Rebased now that #2124 has been merged.

hamogu:
There are so many open PRs that it's hard to keep track which one is in my court and which one is back on @DougBurke. Here, @DougBurke says ["need to think"](https://github.com/sherpa/sherpa/pull/2114#discussion_r1747044162) so I assume you are not waiting on me, right?

DougBurke:
yeah -with #2124 merged I am less gung-ho on pushing this one, so it's currently back-burnered on my set of burners. Which are burning.

sherpa/astro/data.py
hamogu: We discussed this at the developer telecom and the comment "Need to think" make me think that it's not in my court, but waiting for @DougBurke to finish that thinking. But please let me know if I'm wrong - I don't want to hold thus up unnecessarily.

PR #2115, opened 2024-08-24
doc: add examples of using set_xlabel/ylabel

DougBurke:
# Summary

Add documentation showing the new set_xlabel/ylabel functionality from #2086

# Details

Add a section in the "A sample of plots" notebook.

Update the Session documentation to

- create the plot rather than hard code it
- add a note about accessing the Sherpa objects directly
- add a new plot that shows off the set_xlabel/ylabel changes


anetasie (approved):
This looks good.

PR #2116, opened 2024-08-27
Add explicit support for XSPEC 12.14.1

DougBurke:
# Summary

Add models from XSPEC 12.14.1 and change the parameter settings to match those from 12.14.1, where necessary.

# Details

We do not need any changes to build against 12.14.1, which is nice, but there are new models we can provide.

I have reported two issues to the XSPEC  with 12.14.1 which have been fixed in 12.14.1b (the latest release at the time of writing, Sep 11 2024):

- `ismabs` seems to crash, so we skip this test if using 12.14.1; it has been fixed in 12.14.1b (it's not obvious there ever was a 12.14.1a as they were released at the same time from what I understand)
- the new `sssed` model has an invalid `frac_th` parameter settings in `model.dat`;  I have confirmed with XSPEC 12.14.1b that I got it right.

Adds models:

- additive: `sssed`, `vagauss`, `vgaussian`, `zvagauss`, `zvgaussian`
- multiplicative: `vgabs`, `zgabs`, `zvgabs`

I've run local tests built against XSPEC versions 12.14.1 and 12.14.1b and they pass.

### DIFF
--- a/scripts/check_xspec_update.py
+++ b/scripts/check_xspec_update.py
@@ -24,6 +24,9 @@
 
   ./check_xspec_update.py infile
 
+    --nohard
+    --noswitch
+
 Aim:
 
 Given an XSPEC model.dat file report on differences to the supported
@@ -37,16 +40,26 @@ At the moment this requires a Sherpa build with a working XSPEC
 module; this could be worked around but it's not felt to be worth it
 at this time.
 
+It could take the infile from $HEADAS/../spectral/manager/model.dat
+if not set, but I want to be explicit about what file is being used,
+and it's too easy to have the environment set for a different version
+of XSPEC.
+
 """
 
 import sys
+from typing import Sequence
 
-from sherpa.astro.utils.xspec import parse_xspec_model_description
+from sherpa.astro.utils.xspec import ModelDefinition, \
+    parse_xspec_model_description
 from sherpa.astro import xspec
 from sherpa.models.parameter import hugeval
 
 
-def compare_xspec_models(models, hard=True):
+def compare_xspec_models(models: Sequence[ModelDefinition],
+                         hard: bool = True,
+                         switch: bool = True
+                         ) -> None:
     """Check the sherpa.astro.xspec models to those in models.
 
     This is intended to see what updates are needed when updating the
@@ -58,6 +71,9 @@ def compare_xspec_models(models, hard=True):
     models : list of ModelDefinition
     hard : bool, optional
         Do we bother checking the hard limits (default True)?
+    switch : bool, optional
+        Do we bother checking the limits for the switch parameter
+        (default True)?
 
     """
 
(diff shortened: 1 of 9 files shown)


DougBurke:
Rebased to pick up

- #1793 

It didn't conflict but I wanted to make sure we were using the new code.

scripts/check_xspec_update.py
```diff
@@ -58,6 +71,9 @@ def compare_xspec_models(models, hard=True):
     models : list of ModelDefinition
     hard : bool, optional
         Do we bother checking the hard limits (default True)?
+    switch : bool, optional
+        Do we bother checking the limits for the swich parameter
```
wmclaugh, line 75:
```suggestion
        Do we bother checking the limits for the switch parameter
```

DougBurke: Argh.

anetasie:
@DougBurke @wmclaugh I don't remember what XSPEC version we were planning to support for this release. Are we planning to take this PR?

DougBurke:
@anetasie - CIAO 4.17 will use XSPEC 12.14.0something, not 12.14.1.

If we do not take this PR then users can still build with 12.14.1, they just won't get the new models (this is not always true, but fortunately is in this case). So while it is not ideal, it is not a deal breaker (the new models appear to generally be XRISM-related with making it easier to get widths of lines in km/s rather than energy, so again not what I would claim are generally important).

DougBurke:

DougBurke:
Rebased as #2215 has been merged (there was no conflict but I just wanted to make sure everything was up to date).

sherpa/astro/xspec/__init__.py
```diff
@@ -2642,6 +2642,10 @@ class XSbcph(XSAdditiveModel):
 
     The model is described at [1]_.
 
+    .. versionchanged:: 4.17.0
```
DougBurke, line 2645: argh - need to update the version number

sherpa/astro/xspec/__init__.py
```diff
@@ -11088,6 +11088,125 @@ def __init__(self, name='ssa'):
         XSAdditiveModel.__init__(self, name, (self.te, self.y))
 
 
+@version_at_least("12.14.1")
+class XSsssed(XSAdditiveModel):
+    """The XSPEC sssed model: Shakura & Sunyaev spectral energy distribution
+
+    The model is described at [1]_.
+
+    .. versionadded:: 4.17.0
```
DougBurke, line 11097: bump version number

sherpa/astro/xspec/__init__.py
```diff
@@ -11299,6 +11299,46 @@ def __init__(self, name='tapec'):
         XSAdditiveModel.__init__(self, name, pars)
 
 
+@version_at_least("12.14.1")
+class XSvagauss(XSAdditiveModel):
+    """The XSPEC vagauss model: gaussian line profile in wavelength space with sigma in velocity
+
+    The model is described at [1]_.
+
```
DougBurke, line 11307: Need a `versionadded` comment, I htink

DougBurke:
@hamogu - do you have time to quickly check the python parts of this? It is a relatively-straightforward upgrade so I don't see it being too problematic, but it is too easy to get into a "this is obviously correct" mindset when making this sort of a change!

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -37,7 +37,7 @@
 #    '(XSConvolutionKernel)'
 # in `xspec/__init__.py`
 #
-XSPEC_MODELS_COUNT = 281
+XSPEC_MODELS_COUNT = 289
```
hamogu, line 40: We went from 28 to 289 models?
```suggestion
XSPEC_MODELS_COUNT = 29
```

hamogu (changes requested):
It never amazes me how XSPEC adds new models that look all the same (e.g. absorption with with z=0 fixed vs. z variable). I'm sure they have good reasons...but that's not for us to litigate; we simply take what they offer.

sherpa/astro/xspec/tests/test_xspec.py
DougBurke: Errr, no. We went from 281 to 289 models. As the code originally said.

hamogu: Oh wow. That's a lot of models!

hamogu (approved):

sherpa/astro/xspec/tests/test_xspec.py
DougBurke: For reference, in CIAO 4.17 (which only supports models up to 12.14.0) we have

```
sherpa In [3]: len(list_models("xspec"))
Out[3]: 281

sherpa In [4]: list_models("xspec")[0:5]
Out[4]: ['xsabsori', 'xsacisabs', 'xsagauss', 'xsagnsed', 'xsagnslim']

sherpa In [5]: from sherpa.astro import xspec

sherpa In [6]: xspec.get_xsversion()
Out[6]: '12.14.0k'

```

PR #2118, opened 2024-08-28
Improve support for PHA data starting at channel 0

### LINKED ISSUE #1267: What if channel count does not start at 0 or 1?
I know that XSPEC typically uses 0 or 1, but I believe that's only a recommendation, not a requirement in the OGIP. Sherpa is the same. We should add tests and see if we hardwired those numbers into the code somewhere.

see https://github.com/sherpa/sherpa/pull/1264#discussion_r697046428

### LINKED ISSUE #1979: `create_rmf` channel enumeration clashes with spectral channels
On paper, Sherpa and XSpec do not allow the channel range of the spectrum and that defined in the RMF to differ, although I've been able to get models to fit in Sherpa with a channel mismatch.  Downstream, when trying to plot the convolved model with `plot_model`, `plot_fit`, `plot_bkg_model`, or `plot_bkg_fit`, a `ValueError: RMF data is invalid or inconsistent` is thrown. 

```
In [8]: plot_fit()
Cell In[8], line 1
----> 1 plot_fit()

File <string>:1, in plot_fit(id, replot, overplot, clearwindow, **kwargs)

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/ui/utils.py:14062, in Session.plot_fit(self, id, replot, overplot, clearwindow, **kwargs)
  13979 def plot_fit(self, id=None, replot=False, overplot=False,
  13980              clearwindow=True, **kwargs):
  13981     """Plot the fit results (data, model) for a data set.
  13982 
  13983     This function creates a plot containing the data and the model
   (...)
  14059 
  14060     """
> 14062     plotobj = self.get_fit_plot(id, recalc=not replot)
  14063     self._plot(plotobj, overplot=overplot, clearwindow=clearwindow,
  14064                **kwargs)

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/ui/utils.py:11257, in Session.get_fit_plot(self, id, recalc)
  11248 # We don't use get_model_plot as that uses the ungrouped data
  11249 #    modelobj = self.get_model_plot(id)
  11250 # but we do want to use a histogram plot, not _modelplot.
   (...)
  11254 # no way to get it by API (apart from get_fit_plot).
  11255 #
  11256 modelobj = sherpa.astro.plot.ModelPHAHistogram()
> 11257 modelobj.prepare(data, self.get_model(id),
  11258                  self.get_stat())
  11260 plotobj.prepare(dataobj, modelobj)
  11261 return plotobj

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/plot.py:176, in ModelPHAHistogram.prepare(self, data, model, stat)
    172 if not isinstance(data, DataPHA):
    173     raise IOErr('notpha', data.name)
    175 (_, self.y, _, _,
--> 176  self.xlabel, self.ylabel) = data.to_plot(yfunc=model)
    177 self.y = self.y[1]
    179 if data.units != 'channel':

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/data.py:4819, in DataPHA.to_plot(self, yfunc, staterrfunc, response_id)
   4816 def to_plot(self, yfunc=None, staterrfunc=None, response_id=None):
   4817     return (self.apply_filter(self.get_x(response_id=response_id),
   4818                               self._middle),
-> 4819             self.get_y(True, yfunc, response_id=response_id),
   4820             self.get_yerr(True, staterrfunc, response_id=response_id),
   4821             self.get_xerr(True, response_id=response_id),
   4822             self.get_xlabel(),
   4823             self.get_ylabel())

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/data.py:4237, in DataPHA.get_y(self, filter, yfunc, response_id, use_evaluation_space)
   4235 def get_y(self, filter=False, yfunc=None, response_id=None,
   4236           use_evaluation_space=False):
-> 4237     vallist = Data.get_y(self, yfunc=yfunc)
   4238     filter = bool_cast(filter)
   4240     if not isinstance(vallist, tuple):

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/data.py:1180, in Data.get_y(self, filter, yfunc, use_evaluation_space)
   1178     y2 = self.eval_model_to_fit(yfunc)
   1179 else:
-> 1180     y2 = self.eval_model(yfunc)
   1182 return (y, y2)

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/data.py:1364, in Data.eval_model(self, modelfunc)
   1361 if None not in [mdim, ddim] and mdim != ddim:
   1362     raise DataErr(f"Data and model dimensionality do not match: {ddim}D and {mdim}D")
-> 1364 return modelfunc(*self.get_indep())

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/models/model.py:687, in Model.__call__(self, *args, **kwargs)
    685 if (len(args) == 0 and len(kwargs) == 0):
    686     return self
--> 687 return self.calc([p.val for p in self.pars], *args, **kwargs)

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/instrument.py:565, in RSPModelPHA.calc(self, p, x, xhi, *args, **kwargs)
    563 src = self.model.calc(p, self.xlo, self.xhi)
    564 src = self.arf.apply_arf(src, *self.arfargs)
--> 565 src = self.rmf.apply_rmf(src, *self.rmfargs)
    567 # Assume any issues with the binning (between AREASCAL
    568 # and src) is related to the RMF rather than the ARF.
    569 return apply_areascal(src, self.pha,
    570                       f"RMF: {self.rmf.name}")

File /scratch2/Anaconda/anaconda3/envs/ciao-4.16-py311/lib/python3.11/site-packages/sherpa/astro/data.py:1141, in DataRMF.apply_rmf(self, src, *args, **kwargs)
   1137 if len(src) != len(self._lo):
   1138     raise TypeError("Mismatched filter between ARF and RMF " +
   1139                     "or PHA and RMF")
-> 1141 return rmf_fold(src, self._grp, self._fch, self._nch, self._rsp,
   1142                 self.detchans, self.offset)

ValueError: RMF data is invalid or inconsistent
```

Whether or not a channel mismatch gets caught early if a non-compliant RMF is read into Sherpa is outside the scope of this issue, but is problematic when Sherpa uses a RMF generated by the `create_rmf` function.  The `startchan` argument is only used to set the `offset` attribute in the RMF instance and the channels defined in the `F_CHAN` column are fixed to begin with 1.    

`create_rmf` makes use of the `create_delta_rmf` function:

https://github.com/sherpa/sherpa/blob/75ecce3c69d7fc1ca0962e601c2d6ba40cbf013e/sherpa/astro/instrument.py#L1282-L1352

specifically line:
https://github.com/sherpa/sherpa/blob/75ecce3c69d7fc1ca0962e601c2d6ba40cbf013e/sherpa/astro/instrument.py#L1339

should be changed to:
```
    f_chan = numpy.arange(offset, nchans + offset, dtype=numpy.int16)
```

If you have 4096 channels, with an `startchan` offset of 1 (the default), then `F_CHAN` is expected to span from 1-4096.  Similarly, it's expected that a `startchan` offset of 0 should span from 0-4095, but with how its written, the channels are hardcoded to begin at 1 regardless of `startchan` value.  

When plotting with these tweaked RMFs, no errors are thrown (a lot of XMM EPIC spectra have channels enumerated starting with 0).   So I think the comment in both the `create_delta_rmf` and `create_non_delta_rmf` in `sherpa/astro/instrument.py`: "TODO: should f_chan start at startchan?" I'd say yes.

### LINKED ISSUE #2146: DataPHA.get_x cleanup
Noticed when working in #2118 but I did not want to look into it at this time

https://github.com/sherpa/sherpa/blob/5af8dc13c2bb45307cc085018b36be09a65daec2/sherpa/astro/data.py#L4201-L4205

which ends up calling one of

https://github.com/sherpa/sherpa/blob/5af8dc13c2bb45307cc085018b36be09a65daec2/sherpa/astro/data.py#L2638-L2682

and this is the only use of `_from_channel` . So we can simplify these routines since they know they are called with `val=self.channel`.

The only pace where this does not happen is in explicit tests, which we can change.

DougBurke:
# Summary

Improve the handling of PHA files that start at channel 0. PHA data that are read in from such a file no longer renumber the CHANNEL column to start at 0 rather than 1, which will mean that notice and ignore filter ranges that are given in channel units will need to be reduced by 1.

# Details

Fix #1267.
Fix #1979.
Fix #2146.

We have long had issues over handling "channel is not 1" cases, e.g.

- #1040
- #1267
- #1305
- #1979 
- ~#2127~ (I wrote this when working on this issue and I got confused, so I don't think the issue is technically a problem, although the code I pointed to is in fact strange and gets removed in this PR)

I have added a bunch of tests to check handling of offset=0, 1, 2, and 5, with the idea being that the results should be the same (this may require tweaking the input arguments based on the offset value), and I wanted to make sure we tested

- low level routines (mainly `sherpa.astro.utils._utils.filter_resp`)
- object routines - e.g. for users directly using `DataPHA`/`DataRMF`
- the UI layer - e.g. `load_pha`/`set_source`/`fit` [although I don't use `fit` but the recently-added `calc_model` routine to represent "evaluate model by the response" which we care about

These tests show that we actually seem to support channels starting at > 1 well. However, the channel=0 case appears bizarre because of the existing I/O code that if given CHANNEL=0, 1, ... in the file we auto-convert to CHANNEL=1, 2, .... This is **odd** but not in itself terrible **EXCEPT**

- we do not apply the same logic for DataPHA data we create manually, leading to uncertainty in using such data
- if you write out the data you have read in then it will have changed, which could be surprising

There are two fixes to change behaviour:

1. we no-longer re-number channel=0, 1, ... to channel=1, 2, ...

    This actually didn't change too many tests - those that did change are generally regression tests - e.g. we just want
    to ensure we get back the values we used to, so they need there checked values to be updated - but there is one
    test that had "real-world" implications: if you notice/ignore a dataset which starts at channel=0 then the selected
    channel range is now likely to be different.

2. remove the special-case offset=0 case we were having for RMF files

    This is primarily a fix to the filter_resp C++ code - and a **minimal** one at that, as we just need to remove the two
    problematic lines highlighted in #2127 - but there's some follow-on changes to some of the instrument handling to
    make sure we either allow offset/firstchan/startchan to be different from 1, or fixes for this code. 

I rely on the existing tests to catch problems (and I can confirm that they did).

In comparison/historical note, XSPEC basically 

a) forces the channels to start at 1
b) drops "bad" channels so you never see them
c) if there's grouping then the "channel" number actually refers to the group number, and not the PI/PHA channel value

I assume the existing (changed in this PR) behaviour of "adding 1 if the channels start at 0" was in part because of a).

# Example

In the following I use the SWIFT dataset we have, which starts at CHANNEL=0, and show that filtering/grouping gives the same result when using energy space. If you switch to channel space then the channel values now are 1 less than they were, as is the X axis of the data plot when displayed in channels.

## CIAO 4.16 behaviour

```
sherpa In [1]: load_pha("target_sr.pha")

sherpa In [2]: load_rmf("swxpc0to12s6_20130101v014.rmf.gz")
/home/dburke/micromamba/envs/ciao416/lib/python3.11/site-packages/sherpa/astro/data.py:859: UserWarning: The minimum ENERG_LO in the RMF 'swxpc0to12s6_20130101v014.rmf.gz' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

sherpa In [3]: load_arf("target_sr.arf")
/home/dburke/micromamba/envs/ciao416/lib/python3.11/site-packages/sherpa/astro/data.py:859: UserWarning: The minimum ENERG_LO in the ARF 'target_sr.arf' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

sherpa In [4]: notice(0.5, 6)
dataset 1: 0:10.24 -> 0.5:6 Energy (keV)

sherpa In [5]: group_counts(5)
dataset 1: 0.5:6 Energy (keV) (unchanged)

sherpa In [6]: dp = get_data_plot()

sherpa In [7]: print(dp.x)
[0.535      0.655      0.77500001 0.85499999 0.95999998 1.04500002
 1.11000001 1.39499998 1.76999998 2.45499998 4.505     ]

sherpa In [8]: print(dp.y)
[1.48430207e-02 7.33419700e-03 1.48430207e-02 1.15445738e-02
 8.65842752e-03 2.49362366e-02 1.29876542e-02 2.12043127e-03
 3.99619762e-03 9.36046243e-04 6.94990863e-05]

sherpa In [9]: notice()
dataset 1: 0.5:6 -> 0:10.24 Energy (keV)

sherpa In [10]: set_analysis("channel")
dataset 1: 1:1024 Channel

sherpa In [11]: set_analysis("energy")
dataset 1: 0:10.24 Energy (keV)

sherpa In [12]: notice(0.5, 6)
dataset 1: 0:10.24 -> 0.5:6 Energy (keV)

sherpa In [13]: set_analysis("channel")
dataset 1: 51:600 Channel

sherpa In [14]: notice(lo=41, hi=600)
dataset 1: 51:600 -> 41:600 Channel

sherpa In [15]: set_analysis("energy")
dataset 1: 0.4:6 Energy (keV)

sherpa In [16]: dp = get_data_plot()

sherpa In [17]: print(dp.x)
[0.405      0.41499999 0.425      0.435      0.44499999 0.455
 0.465      0.47499999 0.485      0.495      0.535      0.655
 0.77500001 0.85499999 0.95999998 1.04500002 1.11000001 1.39499998
 1.76999998 2.45499998 4.505     ]

sherpa In [18]: print(dp.y)
[0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 4.15604934e-02
 0.00000000e+00 0.00000000e+00 1.48430207e-02 7.33419700e-03
 1.48430207e-02 1.15445738e-02 8.65842752e-03 2.49362366e-02
 1.29876542e-02 2.12043127e-03 3.99619762e-03 9.36046243e-04
 6.94990863e-05]

sherpa In [19]: set_analysis("channel")
dataset 1: 41:600 Channel

sherpa In [20]: dp = get_data_plot()

sherpa In [21]: print(dp.x)
[ 41.5  42.5  43.5  44.5  45.5  46.5  47.5  48.5  49.5  50.5  54.5  66.5
  78.5  86.5  97.  105.5 112.  140.5 178.  246.5 451.5]

sherpa In [22]: print(dp.y)
[0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 4.15604537e-04
 0.00000000e+00 0.00000000e+00 1.48430192e-04 7.33419772e-05
 1.48430192e-04 1.15445705e-04 8.65842786e-05 2.49362722e-04
 1.29876418e-04 2.12043131e-05 3.99619748e-05 9.36046255e-06
 6.94990865e-07]
```

## This PR

```
In [2]: load_pha("sherpa-test-data/sherpatest/target_sr.pha")

In [3]: load_rmf("sherpa-test-data/sherpatest/swxpc0to12s6_20130101v014.rmf.gz")
/home/dburke/sherpa/sherpa-xspec/sherpa/astro/data.py:873: UserWarning: The minimum ENERG_LO in the RMF 'sherpa-test-data/sherpatest/swxpc0to12s6_20130101v014.rmf.gz' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

In [4]: load_arf("sherpa-test-data/sherpatest/target_sr.arf")
/home/dburke/sherpa/sherpa-xspec/sherpa/astro/data.py:873: UserWarning: The minimum ENERG_LO in the ARF 'sherpa-test-data/sherpatest/target_sr.arf' was 0 and has been replaced by 1e-10
  warnings.warn(wmsg)

In [5]: plot_data()

In [6]: notice(0.5, 6)
dataset 1: 0:10.24 -> 0.5:6 Energy (keV)

In [7]: group_counts(5)
dataset 1: 0.5:6 Energy (keV) (unchanged)

In [8]: plot_data()

In [9]: dp = get_data_plot()

In [10]: print(dp.x)
[0.535      0.655      0.77500001 0.85499999 0.95999998 1.04500002
 1.11000001 1.39499998 1.76999998 2.45499998 4.505     ]

In [11]: print(dp.y)
[1.48430207e-02 7.33419700e-03 1.48430207e-02 1.15445738e-02
 8.65842752e-03 2.49362366e-02 1.29876542e-02 2.12043127e-03
 3.99619762e-03 9.36046243e-04 6.94990863e-05]

In [12]: notice()
dataset 1: 0.5:6 -> 0:10.24 Energy (keV)

In [13]: set_analysis("channel")
dataset 1: 0:1023 Channel

In [14]: set_analysis("energy")
dataset 1: 0:10.24 Energy (keV)

In [15]: notice(0.5, 6)
dataset 1: 0:10.24 -> 0.5:6 Energy (keV)

In [16]: set_analysis("channel")
dataset 1: 50:599 Channel

In [17]: notice(lo=40, hi=599)
dataset 1: 50:599 -> 40:599 Channel

In [18]: set_analysis("energy")
dataset 1: 0.4:6 Energy (keV)

In [19]: dp = get_data_plot()

In [20]: print(dp.x)
[0.405      0.41499999 0.425      0.435      0.44499999 0.455
 0.465      0.47499999 0.485      0.495      0.535      0.655
 0.77500001 0.85499999 0.95999998 1.04500002 1.11000001 1.39499998
 1.76999998 2.45499998 4.505     ]

In [21]: print(dp.y)
[0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 4.15604934e-02
 0.00000000e+00 0.00000000e+00 1.48430207e-02 7.33419700e-03
 1.48430207e-02 1.15445738e-02 8.65842752e-03 2.49362366e-02
 1.29876542e-02 2.12043127e-03 3.99619762e-03 9.36046243e-04
 6.94990863e-05]

In [22]: set_analysis("channel")
dataset 1: 40:599 Channel

In [23]: dp = get_data_plot()

In [24]: print(dp.x)
[ 40.5  41.5  42.5  43.5  44.5  45.5  46.5  47.5  48.5  49.5  53.5  65.5
  77.5  85.5  96.  104.5 111.  139.5 177.  245.5 450.5]

In [25]: print(dp.y)
[0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 4.15604537e-04
 0.00000000e+00 0.00000000e+00 1.48430192e-04 7.33419772e-05
 1.48430192e-04 1.15445705e-04 8.65842786e-05 2.49362722e-04
 1.29876418e-04 2.12043131e-05 3.99619748e-05 9.36046255e-06
 6.94990865e-07]
```

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -4714,12 +4714,16 @@ It is an integer or string.
                ) -> None:
         """Notice or ignore the given range.
 
+        Select or remove a range of channels when analyzing the data.
+        The units for the range, that is the values for the lo and hi
+        argument, is set by the analysis field.
+
         .. versionchanged:: 4.14.0
            PHA filtering has been improved to fix a number of corner
            cases which can result in the same filter now selecting one
            or two fewer channels that done in earlier versions of
-           Sherpa. The ``lo`` and ``hi`` arguments are now restricted based on
-           the units setting.
+           Sherpa. The ``lo`` and ``hi`` arguments are now restricted
+           based on the units setting.
 
         Parameters
         ----------
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -2658,7 +2658,11 @@ will be removed. The identifiers can be integers or strings.
 
         return (lo, hi)
 
-    def _group_to_channel(self, val, group=True, response_id=None):
+    # The _group_to_channel, _channel_to_energy, and
+    # _channel_to_wavelength routines are only used by _from_channel
+    # (it gets set to one of these three).
+    #
+    def _group_to_channel(self, group=True, response_id=None):
         """Convert group number to channel number.
 
         For ungrouped data channel and group numbering are the
@@ -2667,7 +2671,7 @@ will be removed. The identifiers can be integers or strings.
         """
 
         if not self.grouped or not group:
-            return val
+            return self.channel
 
         # The middle channel of each group.
         #
@@ -2676,32 +2680,18 @@ will be removed. The identifiers can be integers or strings.
         # Convert to an integer (this keeps the channel within
         # the group).
         #
-        mid = np.floor(mid)
-        val = np.asarray(val).astype(np.int_) - 1
-        try:
-            return mid[val]
-        except IndexError:
-            raise DataErr(f'invalid group number: {val}') from None
+        return np.floor(mid)
 
-    def _channel_to_energy(self, val, group=True, response_id=None):
+    def _channel_to_energy(self, group=True, response_id=None):
         elo, ehi = self._get_ebins(response_id=response_id, group=group)
-        val = np.asarray(val).astype(np.int_) - 1
-        try:
-            return (elo[val] + ehi[val]) / 2.0
-        except IndexError:
-            raise DataErr('invalidchannel', val) from None
+        return (elo + ehi) / 2.0
 
-    def _channel_to_wavelength(self, val, group=True, response_id=None):
+    def _channel_to_wavelength(self, group=True, response_id=None):
         tiny = np.finfo(np.float32).tiny
-        vals = np.asarray(self._channel_to_energy(val, group, response_id))
-        if vals.shape == ():
-            if vals == 0.0:
-                vals = tiny
-        else:
-            vals[vals == 0.0] = tiny
-
-        vals = hc / vals
-        return vals
+        emid = self._channel_to_energy(group=group, response_id=response_id)
+        # In case there are any 0-energy bins replace them
+        emid[emid == 0.0] = tiny
+        return hc / emid
 
     default_background_id: IdType = 1
     """The identifier for the background component when not set.
@@ -4222,9 +4212,12 @@ It is an integer or string.
         return syserr
 
     def get_x(self, filter=False, response_id=None):
+        if self.channel is None:
+            return None
+
         # We want the full channel grid with no grouping.
         #
-        return self._from_channel(self.channel, group=False, response_id=response_id)
+        return self._from_channel(group=False, response_id=response_id)
 
     def get_xlabel(self) -> str:
         """The label for the independent axis.
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -122,7 +122,7 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import Optional, Sequence, Union
+from typing import Optional, Sequence, Union, overload
 import warnings
 
 import numpy as np
@@ -1041,6 +1041,8 @@ class DataRMF(DataOgipResponse):
         must be in increasing or decreasing order.
     n_grp, f_chan, n_chan, matrix : array-like
     offset : int, optional
+        This must be 0 or greater. This maps to the TLMIN value of
+        the F_CHAN column (when reading from a FITS file).
     e_min, e_max : array-like or None, optional
     header : dict or None, optional
     ethresh : number or None, optional
@@ -1065,8 +1067,13 @@ class DataRMF(DataOgipResponse):
                  header=None, ethresh=None):
         energ_lo, energ_hi = self._validate(name, energ_lo, energ_hi, ethresh)
 
+        # Check it's an integer.
+        if not float(offset).is_integer():
+            raise ValueError(f"offset must be an integer, not {offset}")
+
         if offset < 0:
             raise ValueError(f"offset must be >=0, not {offset}")
+
         self.energ_lo = energ_lo
         self.energ_hi = energ_hi
         self.offset = offset
@@ -1181,20 +1188,36 @@ class DataRMF(DataOgipResponse):
         return rmf_fold(src, self._grp, self._fch, self._nch, self._rsp,
                         self.detchans, self.offset)
 
+    @overload
+    def notice(self, noticed_chans: None) -> None:
+        ...
+
+    @overload
+    def notice(self, noticed_chans: np.ndarray) -> np.ndarray:
+        ...
+
+    # Note that noticed_chans is not a mask, but the channel values.
     def notice(self, noticed_chans=None):
-        bin_mask = None
+        """Filter the response to match the requested channels."""
+
         self._fch = self.f_chan
         self._nch = self.n_chan
         self._grp = self.n_grp
         self._rsp = self.matrix
         self._lo = self.energ_lo
         self._hi = self.energ_hi
-        if noticed_chans is not None:
-            (self._grp, self._fch, self._nch, self._rsp,
-             bin_mask) = filter_resp(noticed_chans, self.n_grp, self.f_chan,
-                                     self.n_chan, self.matrix, self.offset)
-            self._lo = self.energ_lo[bin_mask]
-            self._hi = self.energ_hi[bin_mask]
+
+        # This could also return here if noticed_chans contains all
+        # channels, but that is harder to check.
+        #
+        if noticed_chans is None:
+            return None
+
+        (self._grp, self._fch, self._nch, self._rsp,
+         bin_mask) = filter_resp(noticed_chans, self.n_grp, self.f_chan,
+                                 self.n_chan, self.matrix, self.offset)
+        self._lo = self.energ_lo[bin_mask]
+        self._hi = self.energ_hi[bin_mask]
         return bin_mask
 
     def get_indep(self, filter=False):
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -465,12 +465,25 @@ def test_grouped_ciao4_5(parallel, run_thread, clean_astro_ui):
 
     def cmp_thread(fres, aa, covarerr):
         assert fres.numpoints == 46
-        assert fres.statval == approx(18.8316, rel=1e-4)
 
-        assert covarerr[0] == approx(0.104838, rel=1e-4)
-        assert covarerr[1] == approx(2.43937e-05, rel=1e-4)
-        assert aa.gamma.val == approx(1.83906, rel=1e-4)
-        assert aa.ampl.val == approx(0.000301258, rel=1e-4)
+        # Prior to 4.17.0 the selected channel range of 41 to 167
+        # mapped to 1.159:4.79661 keV and resulted in a best-fit
+        # statistic of 18.8316, gamma = 1.83906, and ampl of
+        # 3.01258e-4. In 4.17 the channels now start at 0 so
+        # the energy range has shifted. Ideally I'd change the
+        # test suite, but that is much harder to do than just
+        # change the fit results.
+        #
+        # The new energy range - as of 4.17 - os 1.18764:5.05439
+        # kev. There is not a huge lot of difference between these
+        # results (i.e between these two ranges).
+        #
+        assert fres.statval == approx(19.0311, rel=1e-4)
+
+        assert covarerr[0] == approx(0.104215, rel=1e-4)
+        assert covarerr[1] == approx(2.52228e-05, rel=1e-4)
+        assert aa.gamma.val == approx(1.83531, rel=1e-4)
+        assert aa.ampl.val == approx(0.000302545, rel=1e-4)
 
     check_thread(run_thread, 'grouped_ciao4.5', parallel, cmp_thread,
                  ['aa'])
@@ -641,7 +654,7 @@ def test_xmm2(run_thread, fix_xspec):
     wmsgs = set([str(w.message) for w in ws])
     assert wmsgs == emsgs
 
-    assert ui.get_data().channel[0] == approx(1.0, rel=1e-4)
+    assert ui.get_data().channel[0] == approx(0.0)
     rmf = ui.get_rmf()
     arf = ui.get_arf()
     assert rmf.detchans == 800
--- a/sherpa/astro/tests/test_astro.py
+++ b/sherpa/astro/tests/test_astro.py
@@ -466,18 +466,6 @@ def test_grouped_ciao4_5(parallel, run_thread, clean_astro_ui):
     def cmp_thread(fres, aa, covarerr):
         assert fres.numpoints == 46
 
-        # Prior to 4.17.0 the selected channel range of 41 to 167
-        # mapped to 1.159:4.79661 keV and resulted in a best-fit
-        # statistic of 18.8316, gamma = 1.83906, and ampl of
-        # 3.01258e-4. In 4.17 the channels now start at 0 so
-        # the energy range has shifted. Ideally I'd change the
-        # test suite, but that is much harder to do than just
-        # change the fit results.
-        #
-        # The new energy range - as of 4.17 - os 1.18764:5.05439
-        # kev. There is not a huge lot of difference between these
-        # results (i.e between these two ranges).
-        #
         assert fres.statval == approx(19.0311, rel=1e-4)
 
         assert covarerr[0] == approx(0.104215, rel=1e-4)
--- a/sherpa/astro/tests/test_astro_data_swift_unit.py
+++ b/sherpa/astro/tests/test_astro_data_swift_unit.py
@@ -650,7 +650,6 @@ def test_1209_background(make_data_path):
     assert d.header["FILTER"] == "NONE"
 
 
-@pytest.mark.xfail
 @requires_fits
 @requires_data
 def test_get_x_channel0(make_data_path):
--- a/sherpa/astro/tests/test_astro_data_swift_unit.py
+++ b/sherpa/astro/tests/test_astro_data_swift_unit.py
@@ -648,3 +648,69 @@ def test_1209_background(make_data_path):
     assert d.header["TELESCOP"] == "SWIFT"
     assert d.header["INSTRUME"] == "XRT"
     assert d.header["FILTER"] == "NONE"
+
+
+@pytest.mark.xfail
+@requires_fits
+@requires_data
+def test_get_x_channel0(make_data_path):
+    """Are we correctly matching up channel and energies?"""
+
+    infile = make_data_path(PHAFILE)
+    rmffile = make_data_path(RMFFILE)
+    arffile = make_data_path(ARFFILE)
+
+    pha = io.read_pha(infile)
+    with warnings.catch_warnings(record=True):
+        warnings.simplefilter("ignore")
+        rmf = io.read_rmf(rmffile)
+        arf = io.read_arf(arffile)
+
+    pha.set_rmf(rmf)
+    pha.set_arf(arf)
+
+    # get_x returns all data, no matter the filter. We can check this
+    # by filtering out everything.
+    pha.ignore()
+
+    pha.set_analysis("channel")
+    x_chan = pha.get_x()
+
+    pha.set_analysis("energy")
+    x_energy = pha.get_x()
+
+    pha.set_analysis("wave")
+    x_wave = pha.get_x()
+
+    assert len(x_chan) == 1024
+    assert len(x_energy) == 1024
+    assert len(x_wave) == 1024
+
+    # Check some expected values. The energy bins are (modulo rounding
+    # issues):
+    #
+    #    0    -  0.01
+    #    0.01 -  0.02
+    #    0.02 -  0.03
+    #    0.03 -  0.04
+    #    0.04 -  0.05
+    #    ...
+    #   10.23 - 10.24
+    #
+    assert x_chan[0:5] == pytest.approx([0, 1, 2, 3, 4])
+    assert x_energy[0:5] == pytest.approx([0.005, 0.015, 0.025, 0.035, 0.045])
+    assert x_wave[0:5] == pytest.approx([2479.68380343, 826.56126781,
+                                         495.93676069, 354.24054335,
+                                         275.5204169])
+
+    # Do not bother checking wavelength since it is calculated from
+    # energy and we have shown above the wavelength mapping is
+    # correct.
+    #
+    assert x_chan[1023] == pytest.approx(1023)
+    assert x_energy[1023] == pytest.approx(10.235)
+
+    # check ascending or descending
+    assert (x_chan[1:] > x_chan[:-1]).all()
+    assert (x_energy[1:] > x_energy[:-1]).all()
+    assert (x_wave[1:] < x_wave[:-1]).all()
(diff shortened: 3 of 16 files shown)


sherpa/astro/tests/test_astro.py
```diff
@@ -465,12 +465,25 @@ def test_grouped_ciao4_5(parallel, run_thread, clean_astro_ui):
...
 
-        assert covarerr[0] == approx(0.104838, rel=1e-4)
-        assert covarerr[1] == approx(2.43937e-05, rel=1e-4)
-        assert aa.gamma.val == approx(1.83906, rel=1e-4)
-        assert aa.ampl.val == approx(0.000301258, rel=1e-4)
+        # Prior to 4.17.0 the selected channel range of 41 to 167
+        # mapped to 1.159:4.79661 keV and resulted in a best-fit
+        # statistic of 18.8316, gamma = 1.83906, and ampl of
+        # 3.01258e-4. In 4.17 the channels now start at 0 so
+        # the energy raneg has shifted. Ideally I'd change the
```
wmclaugh, line 473:
```suggestion
        # the energy range has shifted. Ideally I'd change the
```

sherpa/astro/data.py
```diff
@@ -1434,6 +1434,10 @@ class DataPHA(Data1D):
     The PHA format is described in an OGIP document [OGIP_92_007]_ and
     [OGIP_92_007a]_.
 
+    .. versionchanged:: 4.17.0
+       The first channel is no-longer re-numbered to 1, but matches
```
wmclaugh, line 1438:
```suggestion
       The first channel is no longer renumbered to 1, but matches
```

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -109,6 +101,137 @@ def test_read_pha(make_data_path):
...
+def check_arf(arf):
+    """Does the ARF contain the expected values?"""
+
+    assert isinstance(arf, DataARF)
+
+    nbins = 2400
+
+    # Expect the upper edge of bin j to equal the lower
+    # edge of bin j + 1. Note that we check that the low
+    # bin has been relpaced (i.e. is not 0.0).
```
wmclaugh, line 139:
```suggestion
    # bin has been replaced (i.e. is not 0.0).
```

DougBurke: I actually just found and fixed that in my version. I am going to be doing significant work on this PR so I don't think it's worth much review just yet, as there's plenty of places for me to add typos!

sherpa/astro/tests/test_astro_data_swift_unit.py
```diff
@@ -109,6 +101,137 @@ def test_read_pha(make_data_path):
...
+    # Rather than check each element, use some simple summary
+    # statistics.
+    assert rmf.matrix.min() == pytest.approx(0.0)
+    assert rmf.matrix.max() == pytest.approx(0.05834391713142395)
+    assert rmf.matrix.sum() == pytest.approx(1133.8128197300257)
+    assert np.argmax(rmf.matrix) == 269443
+
+    # Expect the upper edge of bin j to equal the lower
+    # edge of bin j + 1. Note that we check that the low
+    # bin has been relpaced (i.e. is not 0.0).
```
wmclaugh, line 194:
```suggestion
    # bin has been replaced (i.e. is not 0.0).
```

sherpa/astro/tests/test_astro_data_xmm_unit.py
```diff
@@ -0,0 +1,240 @@
...
+    assert rmf.matrix.max() == pytest.approx(5.669764518737793)
+    assert rmf.matrix.sum() == pytest.approx(102430.35533461263)
+    assert np.argmax(rmf.matrix) == 5859948
+
+    for field in ['e_min', 'e_max']:
+        assert getattr(rmf, field).size == nchans
+
+    # Expect the upper edge of bin j to equal the lower
+    # edge of bin j + 1. Note that we check that the low
+    # bin has been relpaced (i.e. is not 0.0).
```
wmclaugh, line 169:
```suggestion
    # bin has been replaced (i.e. is not 0.0).
```

sherpa/astro/ui/utils.py
```diff
@@ -2224,6 +2224,10 @@ def load_pha(self, id, arg=None, use_errors=False):
         multiple data sets can be loaded with this command, as
         described in the `sherpa.astro.datastack` module.
 
+        .. versionchanged:: 4.17.0
+           Channel numbers are no-longer remapped to start at 1, but
```
wmclaugh, line 2228:
```suggestion
           Channel numbers are no longer remapped to start at 1, but
```

DougBurke:
@hamogu - this is one of those "never thought I'd get it done" PRs.

I am relying on the tests to catch the behaviour, which is why I have a bunch of new tests that use offset=0, 1,2, and 5 to try and catch problems. This isn't guaranteed to work (e.g. during writing these tests I did end up writing something that wasn't doing what I thought it was), but I feel I have enough tests now [and with some external validation using some of the files in XSPEC] that I claim that our "offset = 0" woes are over.

Except for corner cases.

Like #1305.

And other cases yet to be found...

sherpa/astro/tests/test_astro.py
```diff
@@ -465,12 +465,25 @@ def test_grouped_ciao4_5(parallel, run_thread, clean_astro_ui):
...
+        # statistic of 18.8316, gamma = 1.83906, and ampl of
+        # 3.01258e-4. In 4.17 the channels now start at 0 so
+        # the energy range has shifted. Ideally I'd change the
+        # test suite, but that is much harder to do than just
+        # change the fit results.
+        #
+        # The new energy range - as of 4.17 - os 1.18764:5.05439
+        # kev. There is not a huge lot of difference between these
+        # results (i.e between these two ranges).
+        #
```
hamogu, line 480: I would remove this comment. It makes sense now for me during the review process, but won't make any sense to a reader in the future. Instead, I suggest to put it as a comment the commit message or into this PR on GitHub. Either way, we'll be able to see it for a while (forever if it's in git, but it's a bit harder to associate with the code lines then in the GH web interface.).
Or add another commit that removes this comment - then we have it in the git history for this file.

A reader in the future only needs to know that the test tests what it tests and that any change compared to that is a regression that needs to be investigated.

hamogu (changes requested):
Except for my one comment on a comment in a test, this is good.

It's definitely a BIG step in the right direction, even if some obscure corner cases remain.  In particular I like how it makes the code less fragile by removing all the special cases for TLMIN=0. It was hard to test and hard to track what's the original channel number, and confusing to have the channel number in Sherpa be different from the channel number in the file in some, but not all cases.

DougBurke:
@nplee - can you comment on if this PR worked with the CIAO sherpa regression tests (just to add more confidence that it's not breaking things. hopefully)?

nplee:
the CIAO-Sherpa regression tests passed using the latest pull built against this week's CIAOX (4.17 September 13, 2024 build) [make sure to build Sherpa with numpy<2, or bad things happen].

sherpa/astro/tests/test_astro.py
DougBurke: I've just added a commit that removes this text (rather than mess around with a rebase).

DougBurke:
The last commit just tweaks some tests - they compare the same PHA data but with channel=0,... to channel=1,... with a notice filter given in channel units. We have two ways to work this

- the old way, which is use the same `lo`/`hi` limits for the two datasets, which then selects different data which we can check
- the new way which is to use `lo-1`,`hi-1` for the channel=0 case and then the calls return the same values

I've chosen the latter because it's easier to see what the test is actually checking.

hamogu (approved):

DougBurke:
@hamogu - when working on #2142 I found a bug in the offset!=1 code so I've just added test and then a code fix (see #2146) here.

sherpa/astro/data.py
```diff
@@ -2676,32 +2680,18 @@ def _group_to_channel(self, val, group=True, response_id=None):
...
 
-    def _channel_to_energy(self, val, group=True, response_id=None):
+    def _channel_to_energy(self, group=True, response_id=None):
         elo, ehi = self._get_ebins(response_id=response_id, group=group)
-        val = np.asarray(val).astype(np.int_) - 1
-        try:
-            return (elo[val] + ehi[val]) / 2.0
-        except IndexError:
-            raise DataErr('invalidchannel', val) from None
+        return (elo + ehi) / 2.0
```
hamogu, line 2687: I'm not a big fan on functions that are just one or two lines long. This is (according to the comment added above) only use in one spot. I would thus suggest to inline all three of those functions into `_channel_to_energy`. The difference in runtime is negligible, but the difference in readability is not.

hamogu (changes requested):
My request for change is stylistic, not functional but I feel that some of the complexity of sherpa stems from a larger number of levels of abstraction into separate functions. That makes sense when we factor out common functionality or when we summarize long pieces of code and wrap them into something that 's easier to understand when reading the code (e.g. `_channel_to_wavelength`) but I feel we should also consider the opposite: If a function is using in only one place and is only one or  lines long I suggest to inline it again.

sherpa/astro/data.py
DougBurke: I don't understand what you mean here. If you look at `def _set_units` you see we basically have

```
if using channels:
    self._from_channel = self._group_to_channel
elif using energy:
    self._from_channel = self._channel_to_energy
elif using wavelength:
    self._from_channel = self._channel_to_wavelength
else
    oops, this should not be possible
```

So it's not clear to me how you would want to inline them.

There is the possibility of doing a more-serious rework of this code, but I do not see the benefit of this **at this time** as I'm trying to focus of other parts of the DataPHA class.

hamogu: Ok, let's worry about that later.

hamogu (approved):

PR #2122, opened 2024-09-05
Update the simulation documentation

DougBurke:
# Summary

Improve the simulation documentation to better reflect the changes to fake_pha made in https://github.com/sherpa/sherpa/pull/1684.

# Details

This is #2043 without the not-yet-resolved "make the simulation docs create the plots during the build" work. That is, this just updates the documentation to better match the code changed by #1684.

I made this rather to make it more obvious the difference (and try not to mess up the review that @hamogu has already made in #2043).


DougBurke:
@hamogu - I've basically ignored your suggestions for the helper scripts as they are meant to be "quick and dirty", will go away after we get #2043 worked out, and there are actually reasons for the code behaving as it does.

Were you able to review the actual text changes in `docs/evaluation/simulate.rst`?

docs/evaluation/simulate.rst
```diff
@@ -62,11 +94,12 @@ time. So, we first create a :py:class:`~sherpa.astro.data.DataPHA` object with
...
-  >>> from sherpa.astro.io import read_arf, read_rmf, read_pha
-  >>> data = DataPHA(name='any', channel=None, counts=None, exposure=10000.)
-  >>> data.set_arf(read_arf(data_dir + '9774.arf'))
-  >>> data.set_rmf(read_rmf(data_dir + '9774.rmf'))
+   >>> from sherpa.astro.data import DataPHA
+   >>> from sherpa.astro.io import read_arf, read_rmf, read_pha
+   >>> data = DataPHA(name='simulation', channel=None, counts=None, exposure=10000.)
+   >>> data.set_arf(read_arf(data_dir + '9774.arf'))
+   >>> data.set_rmf(read_rmf(data_dir + '9774.rmf'))
+   >>> data.channel = np.arange(1, 1025, dtype=np.int16)
```
hamogu, line 102: I wonder if it's more intuitive to have a channel variable first, instead of setting it to None when the Data object is created.
```suggestion
   >>> channels = np.arange(1, 1025, dtype=np.int16)
   >>> data = DataPHA(name='simulation', channel=channels, counts=None, exposure=10000.)
   >>> data.set_arf(read_arf(data_dir + '9774.arf'))
   >>> data.set_rmf(read_rmf(data_dir + '9774.rmf'))
```

docs/evaluation/simulate.rst
```diff
@@ -106,20 +170,41 @@ slightly different answers. With default settings, the input model is
 convolved with the RMF and multiplied by the ARF, and properly scaled
 for the exposure time.
 
-  >>> from sherpa.astro.fake import fake_pha
-  >>> fake_pha(data, srcmdl)
+::
+
+   >>> from sherpa.astro.fake import fake_pha
+   >>> from sherpa.astro.plot import DataPHAPlot
+   >>> fake_pha(data, instmdl)  # NOTE: include the instrument response
```
hamogu, line 177: We have this text above about "running twice will give different answers". Should we tie that back to how the example above used the rng setting to avoid that? e.g. we could say "The example above set the seed of a random number generator "rng" and passed that into the "poisson noise" function. Here, we do not pass anything in and thus we get different answers".
I'm not sure it's worht the effort. Most people want just some spectrum to show for a proposal and don't care if it's slightly different next time, but if we think that's an important point, we should get back to it here. I'm not sure what is moportant for most users here.

docs/evaluation/simulate.rst
```diff
@@ -136,10 +221,14 @@ complex models which may include such components as the instrumental
 background (which should not be folded through the ARF) or arbitrary
 other components::
 
-  >>> fake_pha(data, model=inst_bkg + my_arf(my_rmf(srcmodel)), is_source=False)  # doctest: +SKIP
+   >>> fake_pha(data, model=inst_bkg + my_arf(my_rmf(srcmodel)), is_source=False)  # doctest: +SKIP
```
hamogu, line 224: `is_source` is not longer used, so I don't thing we should set it at all here, neither to True or False. The background is part of the model.

hamogu: And the first half of the paragraph above can probably be deleted.

docs/evaluation/simulate.rst
```diff
@@ -156,9 +245,16 @@ One way to include background is to sample it from a
 loaded into the dataset before running the simulation and, if not done
 before, the scale of the background scaling has to be set::
 
-  >>> data.set_background(read_pha('sherpa-test-data/sherpatest/9774_bg.pi'))
-  >>> data.backscal = 9.6e-06
-  >>> fake_pha(data, srcmdl, add_bkgs=True)
+   >>> bgdata = read_pha(data_dir + '9774_bg.pi')
+   >>> data.set_background(bgdata)
+   >>> data.backscal = 9.6e-06
+   >>> fake_pha(data, instmdl, add_bkgs=True)
```
hamogu, line 251: The code for `fake_pha` says:
```Python
    if add_bkgs is not None:
        warn("add_bkgs is no-longer used, as the model "
             "either contains a background term or the "
             "background datasets are used",
             category=DeprecationWarning)
```
So, I believe the background dataset will be used if present and `add_bkgs` should not be set all all (neither True nor False).
Or am I missing something obvious?

docs/evaluation/simulate.rst
```diff
@@ -186,15 +282,19 @@ numbers as background value and then performing a Poisson draw on them
...
+   >>> fake_pha(data, mdl, add_bkgs=True, bkg_models={'1': bkgmdl})  # doctest: +SKIP
 
 The keys in the `bkg_models` dictionary are the identifiers of the
 backgrounds. Above, we loaded a background with the default identifier
 (which is ``1``).
 
+.. todo::
+   Looks like we need `instmdl` here, but does this mean that
+   bkg_models needs to be sent `resp(bkgmdl)` ?
+
```
hamogu, line 297: Good question. We should know the answer to that, since we wrote the code...

And the code says that:
```Python
    if bkg_models is not None:
        warn("bkg_models is no-longer used, as the model "
             "must contain any background component",
             category=DeprecationWarning)
```

So, the example above must be rewritten to make `bkgmdl` a part of the model expression like `combined_mdl = resp(srcmdl + bkgmdl)` if the `bkgmdl` represents an astrophysical background. Of course, it could also be an instrumental background (e.g. high-energy proton flux) that is independent of the response, in which case `combined_mdl = resp(srcmdl) + bkgmdl` might be more appropriate. I suggest mentioning both possibilites without going into the reasons why one might want to use the one or the other. Suggested text:

If a background model represents and astrophysical component, e.g. the emission of a nebula behind the source we want to simulate, then the background model needs to apply the instrumental response in the same way as the source spectrum: `combined_mdl = resp(srcmdl + bkgmdl)`. On the other hand, the background model might describe electronic noise on a CCD, high-energy protons or some other instrumental effect that is independent of the mirror effective area. In this more complicated case, the instrumental ARF and RMF might not be appropriate; instead the user might set up a different response for the background.

docs/evaluation/simulate.rst
```diff
@@ -186,15 +282,19 @@ numbers as background value and then performing a Poisson draw on them
...
 
 More than one background
 ^^^^^^^^^^^^^^^^^^^^^^^^
```
hamogu, line 300: I can't leave a comment on the exact line below becaue it's not been changed, but if there are multiple background components, theu will now be added, not averaged:
```Python
        for bkg_id in data.background_ids:
            cts = data.get_background(bkg_id).counts
            scale = data.get_background_scale(bkg_id, units="counts")
            model_prediction += scale * cts
```

hamogu (changes requested):
I believe that even after the changes, the documentation is out of sync with the current code in `sherpa.astro.fake.fake_pha` in many, many places. That's not usually what @DougBurke does when he opens a PR, so I'm wondering if I'm missing something obvious or if it's maybe my codebase that for some reason has not merged in upstream changes.
I've copied some of the relvant code lines from `fake.py` into my comment to help the discussion.

docs/evaluation/simulate.rst
DougBurke: Allowing the channel field to be None has been a thorn in my side, so I'm happy to move documentation away from this approach!

docs/evaluation/simulate.rst
DougBurke: So, this change is my attempt to re-work documentation that I may have started but I'm pretty sure you improved at some point. So remembering what is versus what we want has been confusing for me here!

anetasie:
I was just looking at the new text. Additional figures look good and improve documentation.
@hamogu included a few comments with the most important one related to the Multiple backgrounds - is the background for the simulations from multiple backgrounds averaged or added - I'm not sure, but this would need to be clarified before we can take this PR.

anetasie:
This is the comment `astro/fake.py`  code that @hamogu included in his comment, so we need to update the documentation for Multiple background section .

```
if include_bkg_data:
        # If there are background components then we sum up the data,
        # after scaling to match the source data set, and use that as a
        # source term. Note that get_background_scale will correct for
        # exposure time (as units=counts), scaling factors, and the number
        # of background components.
       ```

DougBurke:
So, I do not have the energy to work on this so close to the deadline. It annoys me we can't fix this documentation for 4.17.0 but we can't do everything.

DougBurke:
@hamogu - I assume we can close this because of #2345 ?

hamogu:
We could; we could also wait till #2345 is accepted since technically this is an alternative in case #2345 is not accepted for some reason.

PR #2123, opened 2024-09-05
Data cleanups

DougBurke:
# Summary

Internal clean up of some of the data code.

# Details

This was written as part of #1944 but I know some of the changes have been "re-invented" for other PRs, so let's just get them merged to avoid conflicts.

This is all minor clean ups

- some style changes (e.g. "import numpy as np" and some lazy-logging changes)
- some fix ups to typing rules
- some additions of typing rules point out a place where we technically can return a None but do not deal with that (there is the thought that maybe we can not hit this case here, but it's hard to track down the requirements so easiest to explicitly fail rather than let Python fail with an `AttributeError`)


hamogu (approved):

DougBurke:
Rebased to address the minor conflicts from merging

- #2124

PR #2124, opened 2024-09-07
Improve the documentation of the id/bkg_id/resp_id arguments

DougBurke:
# Summary

Improve the documentation and typing of the various dataset identifier arguments (`id`, `bkg_id`, `resp_id`).

# Details

This was created as part of #2114 but I've separated it out as this part is hopefully less problematic, as it doesn't really add anything to the code. I have also reworked the code so the commits are different to those in #2114, but the content is essentially the same.

The changes:

- improves the documentation of `id`/`bkg_id`/`resp_id` arguments to better match the code
  - this is mainly to better describe when the argument can be None, or either a single value or sequence of them
- add typing rules for these variables.
- there are some minor code changes, which are basically
  - remove unused imports
  - minor pylint changes (e.g. use f-strings)
  - avoid changing the type of a variable within a routine when it confuses mypy

As mentioned in the discussion of #2114, we **could** add more typing rules, but this is surprisingly hard to do in many cases, so I'd rather get some rules added rather than address all cases. I have added some typing rules that are generally "obvious" , such as mark an argument as `str` or a function as returning `None`, but some changes are a bit-more complex. These tend to be those enabled by adding types to internal fields (normally a `dict[str, ...]` or `dict[IdType, ...]`) where we can take advantage of this knowledge.

As mentioned in previous PRs, sometimes the types given are not-quite-true; for example

- I have tended to label all data objects as being `Data`, but this then can cause type checkers to complain because sometimes the base/"virtual" class doesn't provide the methods the actual instances do, and other times it's because there is a type check on the return value which we currently can't easily type (having `TypeGuard` from 3.10 would help, but I can't guarantee it will fix it)
- the plot classes have a similar issue, where the current hierarchy is not useful for typing, and resolving that is a big issue, such as the following, which is why the plot code isn't touched here
  - #2076
- the response instrument code `ARF1D`/`RMF1D` mimics `DataARF`/`DataRMF` so you can often use them interchangeably, but it's awkward to type as `DataARF | ARF1D | None` when forward imports would be needed, so we currently just use `DataARF` and `DataRMF`

This is quite a disruptive PR (i.e. collides with many other PRs) so I'd rather get it in sooner rather than later.

### DIFF
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -27,7 +27,7 @@ import logging
 import os
 import pickle
 import sys
-from typing import Callable, Optional, Union
+from typing import Any, Callable, Optional, Sequence, TypeVar, Union
 
 import numpy as np
 
@@ -66,6 +66,7 @@ _builtin_symbols_ = tuple(BUILTINS.__dict__.keys())
 
 
 ModelType = Union[Model, str]
+T = TypeVar("T")
 
 
 ###############################################################################
@@ -182,7 +183,10 @@ def _get_filter(data):
         return None
 
 
-def report_filter_change(idstr, ofilter, nfilter, xlabel=None):
+def report_filter_change(idstr: str,
+                         ofilter: Optional[str],
+                         nfilter: Optional[str],
+                         xlabel: Optional[str] = None):
     """Report the filter change for ignore/filter.
 
     Parameters
@@ -254,7 +258,10 @@ def report_filter_change(idstr, ofilter, nfilter, xlabel=None):
     info(ostr)
 
 
-def notice_data_range(get_data, ids, lo, hi, kwargs):
+def notice_data_range(get_data: Callable[[IdType], Data],
+                      ids: Sequence[IdType],
+                      lo, hi,
+                      kwargs) -> None:
     """Filter each dataset and report the change in filter.
 
     Parameters
@@ -744,7 +751,7 @@ class Session(NoNewAttributesAfterInit):
     # Standard methods
     ###########################################################################
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.clean()
         self._model_types = {}
         self._model_globals = np.__dict__.copy()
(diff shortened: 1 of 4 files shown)


hamogu (commented):
For something like this I wonder if we should adopt the stance "If it makes the type checker happy, we merge it". I'm going to review manually, as we always do, but, going forward, I wonder if that's the best use of our time for PRs that mostly add typing.

sherpa/ui/utils.py
```diff
@@ -3855,12 +3938,14 @@ def get_dep(self, id=None, filter=False):
...
-    def get_dims(self, id=None, filter=False):
+    def get_dims(self,
+                 id: Optional[IdType] = None,
+                 filter=False):
         """Return the dimensions of the data set.
 
         Parameters
         ----------
-        id : int or str, optional
+        id : int . str, or None, optional
```
hamogu, line 3948:
```suggestion
        id : int, str, or None, optional
```

DougBurke:
I think it's useful to have some form of a review, just in case I've done something stupid or messed up a rebase, but there's definitely cases of

- oh, I've seen most of this before
- typing changes which, if they pass the tests, aren't going to change the runtime behaviour of the code

which are harder to justify spending significant time on.

sherpa/ui/utils.py
```diff
@@ -17072,15 +17410,16 @@ def get_model_image(self, id=None):
...
+                         id: Optional[IdType] = None):
         """Return the data used by image_source.
 
         Evaluate the source expression for the image pixels - without
         any PSF convolution - and return the results.
 
         Parameters
         ----------
-        id : int or str, optional
+        id : int, str, or None. optional
```
hamogu, line 17422:
```suggestion
        id : int, str, or None, optional
```

hamogu (changes requested):
I absolutely admire Doug's dedication to this type of stuff. Just a few small typos that slipped through.

sherpa/ui/utils.py
DougBurke: argh

sherpa/ui/utils.py
DougBurke: argh argh

Or should that be $\rm{argh}^2$

hamogu (approved):

PR #2125, opened 2024-09-10
Allow per plot kwargs for plot_xxx_components calls

DougBurke:
# Summary

Allow per-plot keyword arguments for plot_model_components and plot_source_components.

# Details

Inspired by (but implementation is different)
 
- #1712 

since we have code that can use it now thanks to 
- #1679

Fortunately the implementation is easy, the main conceptual difficulty is determining whether color="black" is a sequence or a single value (since a string can be thought of as both a scalar and a sequence). So we special case this condition.

An alternative would be to make the user identify which are the sequence values by - for example - use a pluralized version - so `alphas`, `colors`, `linestyles`, `labels`, ... - which have to take a sequence. This is conceptually better, but

- I am not sure how to actually code this, since I am not convinced we can guarantee that all existing keywords do not end in `s`, so how do I know that a given keyword is a scalar or not
- for a user I'm not sure it's better

However, maybe I'm wrong.

# Example

You can now

```
In [24]: plot_model(xlog=True)

In [25]: plot_model_components(color="black", alpha=0.5, linestyle=["dashed", "dotted"], overplot=True, label=["continuum", "line"])
```

sherpa-1

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -22,9 +22,45 @@
 
 Classes provide access to common plotting tasks, which is done by the
 plotting backend defined in the ``options.plot_pkg`` setting of the
-Sherpa configuration file. Note that plot objects can be created
-and used even when only the `sherpa.plot.backends.BasicBackend` is
-available.
+Sherpa configuration file. Note that plot objects can be created and
+used even when only the `sherpa.plot.backends.BasicBackend` is
+available, it is just that no graphical display will be created.
+
+Which backend is used?
+----------------------
+
+When this module is first imported, Sherpa tries to import the
+backends installed with Sherpa in the order listed in the
+``options.plot_pkg`` setting from the ``sherpa.rc`` startup file.
+The first module that imports successfully is set as the active
+backend. The following command prints the name of the backend:
+
+   >>> from sherpa import plot
+   >>> print(plot.backend.name)
+
+Change the backend
+------------------
+
+After the initial import, the backend can be changed by loading one of
+the plotting backends shipped with sherpa (or any other module that
+provides the same interface):
+
+  >>> import sherpa.plot.pylab_backend
+  >>> plot.backend = sherpa.plot.pylab_backend
+
+Creating a plot
+---------------
+
+The plot backend acts as a context manager::
+
+    from sherpa.plot import backend
+    with backend:
+        # Now call the plot/overplot or contor/overcontour methods
+        obj.plot()
+
+This handles setting up the backend, handles any error handling,
+and then ends the session.
+
 """
 
 from __future__ import annotations
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -51,12 +51,15 @@ provides the same interface):
 Creating a plot
 ---------------
 
-The plot backend acts as a context manager::
+A plot backend can act as a context manager to apply a specific backend to just one plot
+without globally changing the backend for the rest of the session:
 
-    from sherpa.plot import backend
-    with backend:
-        # Now call the plot/overplot or contor/overcontour methods
-        obj.plot()
+.. doctest-skip::
+
+    >>> from sherpa.plot import backend
+    >>> with backend:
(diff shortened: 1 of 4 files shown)


DougBurke:
Argh - an accidental rebase made the force-push necessary even when it wasn't actually needed.

sherpa/plot/__init__.py
```diff
@@ -22,9 +22,45 @@
...
+The plot backend acts as a context manager::
+
+    from sherpa.plot import backend
+    with backend:
+        # Now call the plot/overplot or contor/overcontour methods
+        obj.plot()
+
+This handles setting up the backend, handles any error handling,
+and then ends the session.
+
```
hamogu, line 63:
```suggestion
A plot backend can act as a context manager to apply a specific backend to just one plot
without globally changing the backend for the rest of the session:

.. doctest-skip::

    >>> from sherpa.plot import backend
    >>> with backend:
    ...     # Now call the plot/overplot or contor/overcontour methods
    ...     obj.plot()

This handles setting up the backend, handles any error handling,
and then ends the session.

```

sherpa/utils/__init__.py
```diff
@@ -2641,6 +2642,13 @@ def is_iterable(arg):
     return isinstance(arg, (list, tuple, np.ndarray)) or np.iterable(arg)
 
 
+# Can this return TypeGuard[Sequence]?
```
hamogu, line 2645: This is a one line function that is used once. Can we avoid this level of indiraction and just put it (wiht the comment) directly in line where it is used?
We can refractor is needed later.

sherpa/ui/utils.py
```diff
@@ -14370,6 +14372,16 @@ def plot_source_components(self,
         >>> set_source(xsphabs.gal * (powlaw1d.pl + xsgaussian.line))
         >>> plot_source_components(alpha=0.6)
 
+        Plot the combnined source and then overplot the two components
```
hamogu, line 14375:
```suggestion
        Plot the combined source and then overplot the two components
```

hamogu (changes requested):

sherpa/utils/__init__.py
hamogu: I meant this comment to apply to the entire function, not this one comment line.

DougBurke: Well, I plan to use this in the follow on commit #1712 but I'm not sure whether I'll be able to get it done in time is to be seen.

sherpa/plot/__init__.py
DougBurke: Given that this is a `.py` file and not a `.rst`, does the `.. doctest-skip::` directive work here?

[I have a hatred for all the extra documentation symbols that get thrown into Python docstrings, but that is a battle I can't retroactively change ;-]

DougBurke: I'm going to accept this change, since we can change the `doctest-skip` line later if needed,

sherpa/plot/__init__.py
```diff
@@ -22,9 +22,48 @@
...
+---------------
+
+A plot backend can act as a context manager to apply a specific backend to just one plot
+without globally changing the backend for the rest of the session:
+
+.. doctest-skip::
+
+    >>> from sherpa.plot import backend
+    >>> with backend:
+    ...     # Now call the plot/overplot or contor/overcontour methods
```
wmclaugh, line 61:
```suggestion
    ...     # Now call the plot/overplot or contour/overcontour methods
```

sherpa/ui/utils.py
```diff
@@ -14510,6 +14526,16 @@ def plot_model_components(self,
         >>> set_source(xsphabs.gal * (powlaw1d.pl + xsgaussian.line))
         >>> plot_model_components(alpha=0.6)
 
+        Plot the combnined model and then overplot the two components
```
wmclaugh, line 14529:
```suggestion
        Plot the combined model and then overplot the two components
```

DougBurke:
I don't know why the tests haven't finished - looks like one of the codecov builds has just got stuck.

Issue #2119, opened 2024-08-30
Plotting PHA data using plot_data() in a for loop


perischind:
Is there a way to use plot_data() in multiple subplots? I wanted to compare the PHA data to a few models, and have tried to create a for loop with four subplots, but this only seems to plot on the last subplot no matter what I do. Thank you!

DougBurke:
There's a couple of different things that could be done, depending on what you want

- try using `plot` - e.g. `plot("fit", 1, "fit", 2, "fit", 3)`

  Unfortunately this does not work well when you want to over plot things (even if you use the overplot and clearwindow arguments). Also, there is not enough support for controlling the "subplot" layout (e.g. number of columns, separation of the plots).

- you can overplot data, e.g.

     plot_data()
     plot_data(2, overplot=True)
     plot_model(overplot=True)
     plot_model(2, overplot=True)

   but this is a single plot, not multiple sub-plots

- you can combine matlpotlib commands with our plots, so something like

     plt.clf()
     for idx, pid in enumerate(["a", "b", "c", "d"], 1):
         plt.subplot(2, 2, idx)
         # you could just say 'plot_fit(pid, overplot=True)' but I wanted multiple plots
         plot_data(pid, overplot=True)
         plot_model(pid, overplot=True)

  where I have assumed the identifiers are "a", "b", ... rather than 1, 2,  ... just to be different 
  One issue with this approach is that you lose the axis labels.

There are some other approaches that could be taken, but hopefully these get you started.

perischind:
This helped! Thank you so much!

DougBurke:
For reference, in 4.17.0 you will be able to say

```python
plot("data", 1, "data", 2, "data", 3, cols=1)
plot("model", 1, "model", 2, "model", 3, overplot=True)
```

Prior to this release the overplot support for `plot` was "not great"™, and also you couldn't change the layout (e.g. with `cols=1` above).

I wasn't able to get to making it easy to say something like

```
plot_fit(All)
```

and have it use all the identifiers, as I ran out of time. Hopefully this will get done for the 4.17.1 release.

DougBurke:
For anyone interested, Sherpa 4.17.0 was released a while ago, and CIAO 4.17 (which includes Sherpa 4.17.0 and is the easiest way to get Sherpa if you want to do spectral X-ray Astronomical data analysis) jas just been released and can be downloaded from https://cxc.cfa.harvard.edu/ciao/download/

It includes things like being able to say

```
[...]

PR #2126, opened 2024-09-12
Allow plot and contour to change the plot layout and have per-plot settings.

DougBurke:
# Summary

Expand the plot and contour commands to allow per-plot options to be given, matching the plot_model_components/plot_source_components commands. Improve the algorithm for deciding how to arrange multiple plots, so it isn't fixed to a maximum of 2 rows. Add the rows and cols arguments so that the layout can be specified by the user.

# Details

When answering 

- #2119 

I remembered how much I found the `plot` command to be lacking with regard to plot control. Well, this adresses some of my complaints (there is an issue to do with `overplot`, but that is addressed in #2129, and this adds useful functionality as is).

Extend the logic of #2125 to allow you to say

```
>>> plot("data", "model", label=["data", "model"], ylog=[False, True], color=["gray", "mauve"])
```

Previously we had no control over the layout, and we had

- single plot -> single plot
- two plots -> a column of two plots
- three or four plots -> two columns, two rows
- 2 * n + 1 to 2 * n + 2 plots -> n + 1 columns, two rows
 
We now allow the user to specify the number of rows or columns (or both) to use, so that

- if both rows and cols are set by the user, use them, unless they don't cover all the plots, then we assume neither are set
- if one is set then use it and calculate the other value
- if neither are set then set the number of colums to ceil(sqrt(nplots)) and then calculate the number of rows

So

```
>>> plot("data", cols=2)
```

now creates a plot in the top-half of the window (there was code special-casing the single-plot behaviour but I found it was easiest to remove it to support #2129 and avoid special-case rules here).

The following will create a 3 column and 2 row display (which matches what it did in 4.16.1):

```
>>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3)
```

We can now change it to 2 columns by 3 rows with any of

```
>>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3, cols=2)
>>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3, rows=3)
>>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3, cols=2, rows=3)
>>> sp = get_split_plot()
>>> sp.rows = 3; sp.cols = 2
>>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3)
```

The last version will persist the setting, so then

```
>>> plot("data", "model")
```

would still use 3 rows by 2 columns.

We can now get a single-column of plots with

```
>>> plot(....., cols=1)
```

# Question

So, should it be

```
>>> plot("data", "model", "resid", cols=1)
```

or

```
>>> plot("data", "model", "resid", ncols=1)
```

Should we use `cols`,`rows` or `ncols`,`nrows`?

### DIFF
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -320,6 +320,55 @@ def notice_data_range(get_data: Callable[[IdType], Data],
         report_filter_change(idstr, ofilter, nfilter, xlabel)
 
 
+@overload
+def calc_multiplot_size(rows: int,
+                        cols: Optional[int],
+                        nplots: int
+                        ) -> tuple[int, int]:
+    ...
+
+@overload
+def calc_multiplot_size(rows: Optional[int],
+                        cols: int,
+                        nplots: int
+                        ) -> tuple[int, int]:
+    ...
+
+def calc_multiplot_size(rows, cols, nplots):
+    """How many rows and columns to use for multi plot/contours?
+
+    Parameters
+    ----------
+    rows, cols
+       The requested sizes, if set (one of them must be set).
+    nplots
+       The nuber of plots.
+
+    Returns
+    -------
+    (nrows, ncols)
+
+    """
+
+    def get(n: int) -> int:
+        return int(np.ceil(nplots / n))
+
+    # Since at least one of rows or cols must be set.
+    #
+    nrows = rows if rows is not None else get(cols)
+    ncols = cols if cols is not None else get(rows)
+
+    if nrows * ncols < nplots:
+        # Go for the nearest (upper) square value for the number
+        # of columns, and calculate the number of rows to match
+        # the data.
+        #
+        ncols = int(np.ceil(np.sqrt(nplots)))
+        nrows = get(ncols)
+
+    return nrows, ncols
+
+
 ###############################################################################
 #
 # Pickling support
@@ -13697,11 +13746,6 @@ class Session(NoNewAttributesAfterInit):
         #
         kwstore = get_per_plot_kwargs(nplots, kwargs)
 
-        if nplots == 1:
-            plotmeth = getattr(self, f"_{plotmeth}")
-            plotmeth(plots[0], **kwstore[0])
-            return
-
         # Store the original values in case they are over-written.
         #
         sp = self._splitplot
@@ -13711,35 +13755,22 @@ class Session(NoNewAttributesAfterInit):
         # What size to use?
         #
         if rows is None and cols is None:
-            nrows = nrows_orig
-            ncols = ncols_orig
-
-        elif rows is not None and cols is not None:
-            nrows = rows
-            ncols = cols
-
-        elif rows is not None:
-            nrows = rows
-            ncols = int(np.ceil(nplots / nrows))
+            # Special case the single-plot call
+            if nplots == 1:
+                nrows = 1
+                ncols = 1
+            else:
+                nrows, ncols = calc_multiplot_size(nrows_orig,
+                                                   ncols_orig, nplots)
 
         else:
-            ncols = cols
-            nrows = int(np.ceil(nplots / ncols))
+            nrows, ncols = calc_multiplot_size(rows, cols, nplots)
 
-        if nrows * ncols < nplots:
-            # Go for the nearest (upper) square value for the number
-            # of columns, and calculate the number of rows to match
-            # the data.
-            #
-            ncols = int(np.ceil(np.sqrt(nplots)))
-            nrows = int(np.ceil(nplots / ncols))
-
+        plotfunc = getattr(sp, f"add{plotmeth}")
         sp.reset(rows=nrows, cols=ncols)
-        plotmeth = getattr(sp, f"add{plotmeth}")
-
         with sherpa.plot.backend:
             for plot, store in zip(plots, kwstore):
-                plotmeth(plot, **store)
+                plotfunc(plot, **store)
 
         # Restore the settings. Is this needed?
         #
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -342,7 +342,7 @@ def calc_multiplot_size(rows, cols, nplots):
     rows, cols
        The requested sizes, if set (one of them must be set).
     nplots
-       The nuber of plots.
+       The number of plots.
 
     Returns
     -------
@@ -13646,7 +13646,7 @@ class Session(NoNewAttributesAfterInit):
         plotmeth
             The call.
         rows, cols
-            The number of rows or columns (to over-ride split plot).
+            The number of rows or columns (to override split plot).
         kwargs
             The keyword arguments to apply to each plot or contour.
 
@@ -13746,7 +13746,7 @@ class Session(NoNewAttributesAfterInit):
         #
         kwstore = get_per_plot_kwargs(nplots, kwargs)
 
-        # Store the original values in case they are over-written.
+        # Store the original values in case they are overwritten.
         #
         sp = self._splitplot
         nrows_orig = sp.rows
@@ -14001,7 +14001,7 @@ class Session(NoNewAttributesAfterInit):
            The keyword arguments can now be set per plot by using a
            sequence of values. The layout can be changed with the
            rows and cols arguments and the automatic calculation
-           no-longer forces two rows.
+           no longer forces two rows.
 
         .. versionchanged:: 4.15.0
            A number of labels, such as "bkgfit", are marked as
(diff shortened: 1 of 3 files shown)


hamogu:
When in doubt, we've typically followed Matplotlib, which is the most well-known library for our users. That argues for [ncols and nrows as in plt.subplots](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html) which is probably the closest equivalent.

hamogu:
I'm unhappy with the semantics of the `plot` command with its variable numbers of arguments, mixing of ids and plot types seemingly randomly (and the associated complexity of ensuring that anything that can be plot type cannot be a data id because then plot could not distinguish between the two - in other words, there is no namespace separation between plot types and data ids), and data_ids the follow, instead of precede that plot type that they mark up:
``` >>> plot("data", "model", "data", 2, "model", 2, "data", 3, "model", 3)```

Why are strings and numbers mixed, and why do the numbers come after the strings? Would not make more sense to say "OK, now select dataset 2 and do the *following* plots with that dataset, rather than saying "oops, and after I listed all the plot types that I want, let me tell you which data I actually want to be plotted"?

It's too late to change, but IMHO a clearer, if much more verbose, syntax would be something like 
```plot((1, "data"), (1, "model"), (2, "data"))```. 
This would allow an easier extension towards plotting in e.g. a 2x2 grid 
```plot([([1, "data"), (1, "model")], [(2, "data")], []])```. 
It would also overcome the problem of overplotting with calls like 
```plot([[(1, "data", 2, "data"), (1, "model", 2, "model")], [(2, "data")], []])```

Now, the plot semantics have been around for a while, but because of these complexities I wonder how much they are used. I never do, because I don't understand how it's meant to be used and we don't have a spec that I can find that actually explains how it's supposed to be used. Instead, I get individual data with e.g. `get_arf_plot()` and then plot in matplotlib. Not shorter, but I can understand what's going to happen.

Since it's been around for a while, why am I even complaining? Sherpa has been around long enough that I can't really ask to remove the plot command, but maybe we can agree that it's horribly confusing and we just let is die quietly and design a new version from the ground up and call it `plot_for_real_this_time`?

Or maybe I should just take the stance that one more function that I don't use in Sherpa doesn't hurt me - there is enough UI stuff that I don't know about?

File: sherpa/ui/utils.py [hunk 1]
```diff
@@ -14049,8 +14168,37 @@ def plot(self, *args, **kwargs) -> None:
 
         >>> plot("bkg", 1, "up", "bkg", 1, "down")
 
+        Draw both data sets in black, but with partial opacity:
```
hamogu, line 14171: This actually draws only one dataset, but with two representations (data and model) - they both have the same id (the one for `get_default_id`)

sherpa/ui/utils.py
```diff
@@ -14049,8 +14168,37 @@ def plot(self, *args, **kwargs) -> None:
...
+
+        >>> plot("data", "model", color="black", alpha=0.5)
+
+        Draw the two plots in black but with different opacities:
+
+        >>> plot("data", "model", color="black", alpha=[1. 0.5])
```
hamogu, line 14177:
```suggestion
        >>> plot("data", "model", color="black", alpha=[1., 0.5])
```

sherpa/ui/utils.py
```diff
@@ -16291,7 +16455,10 @@ def contour_fit_resid(self,
         plot1obj = self.get_fit_contour(id, recalc=not replot)
         plot2obj = self.get_resid_contour(id, recalc=not replot)
 
-        self._splitplot.reset()
+        # This does not use _jointplot because the X axis is not
+        # obviously shared between the twp plots.
```
hamogu, line 16459:
```suggestion
        # obviously shared between the two plots.
```

hamogu:
On the other hand, I see the use of the underlying machinery for data stack plots...

DougBurke:
So, we are nit going to change the behaviour of `plot` in this PR, but I will claim that

```
plot_data(2)
```

and

```
plot("data", 2)
```

have a certain symmetry. There's also the ability to pass in extra positional arguments - e.g.

```
plot_bkg(1, 2)
```

and

```
plot("bkg", 1, 2)
```

Whether this is a "good" thing or not is debatable.

It would be nice to improve the logic for combining plots, but there's some point where you just say "here's the data, do it yourself". I feel the multi-plot code is close to that point, right up until I quickly want to plot a bunch of data and then go "gah, if only...".

sherpa/ui/utils.py
DougBurke: Nope, it draws two datasets - the "up" and the "down" background components for dataset 1.

DougBurke:
> When in doubt, we've typically followed Matplotlib, which is the most well-known library for our users. That argues for [ncols and nrows as in plt.subplots](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html) which is probably the closest equivalent.

So, the reason why I used `cols` and `rows` is twofold:

1. it's one-less character, which seems silly but in lines like `plot("data", 1, "data", 2, cols=1)` I start to care about line length
  - this is not a very-good reason
 
2. because the existing `SplitPlot` class already uses these names - you can see it with

```
sherpa In [1]: print(get_split_plot())
rows   = 2
cols   = 1
plot_prefs = {}
```

Now, up until now this PR these settings are essentially ignored, so no-one should have been using them, but there's still some precedence.

sherpa/ui/utils.py
hamogu: Your comment actually says "both are for 'dataset 1`". So it's about what is a "dataset"? How about 
```suggestion
        Draw data and model for dataset 1 in black, but with partial opacity:
```

DougBurke:
FYI for @moritz as you mentioned follow-on work,

- #2129 is my follow-on to this that fixes up the existing overplot support (which unfortunately doesn't work, see #2129); this is ready for review once this PR is accepted (assuming the review doesn't change behavior/internals)
- #1712 is a long-stalled project to allow users to plot multiple datasets at once; I have been reworking it to match the recent changes here with `plot_model_components`/`plot` but there's still some implementation issues to work through and I'm still not happy about some of the internals
  - one idea I have is to add a `All` symbol (name open for bike-shedding) that would let you say `plot_data(All)` and get all the datasets plotted
   - I need to get back to it to see if `plot("data", [1, 2])` or `plot("data", All)` works or what my plans for that was

Given the impending release deadline I'm not sure how much time I have to work on further expansion, as I'm trying to get some PHA offset / quality filtering work to a useful state. Will have to review my workload.
  -

hamogu:
Since you did the follow-on work though, that means you know that is can be done, which is good enough for me. With this knowledge, I can say that this PR can be accepted once the typos and small doc issues that I flagged are addressed.

sherpa/ui/utils.py
DougBurke: It doesn't draw the model at all. It plots the background dataset components "up" and "down" for dataset 1. I'll address try to improve this in a commit rather than try to address this via the GH web page...

DougBurke: Argh - I'm arguing about the code from the previous example....

DougBurke: I've basically taken your suggestion here now, and also added a few other changes to hopefully better-describe functionality (both as examples and as general text in the notes section).

hamogu (approved):

sherpa/ui/utils.py
```diff
@@ -320,6 +320,55 @@ def notice_data_range(get_data: Callable[[IdType], Data],
...
+
+def calc_multiplot_size(rows, cols, nplots):
+    """How many rows and columns to use for multi plot/contours?
+
+    Parameters
+    ----------
+    rows, cols
+       The requested sizes, if set (one of them must be set).
+    nplots
+       The nuber of plots.
```
wmclaugh, line 345:
```suggestion
       The number of plots.
```

sherpa/ui/utils.py
```diff
@@ -13654,20 +13740,42 @@ def _multi_plot(self, args, plotmeth="plot", **kwargs) -> None:
...
-            return
+        nplots = len(plots)
+
+        # Deconstruct the keyword arguments.
+        #
+        kwstore = get_per_plot_kwargs(nplots, kwargs)
 
-        nrows = 2
-        ncols = int((len(plots) + 1) / 2.0)
+        # Store the original values in case they are over-written.
```
wmclaugh, line 13749:
```suggestion
        # Store the original values in case they are overwritten.
```

sherpa/ui/utils.py
```diff
@@ -15811,28 +15978,35 @@ def contour(self, *args, **kwargs) -> None:
         as returned by `get_default_id` - is used. This is for
         2D data sets.
 
+        .. versionchanged:: 4.17.0
+           The keyword arguments can now be set per plot by using a
+           sequence of values. The layout can be changed with the
+           rows and cols arguments and the automatic calculation
+           no-longer forces two rows.
```
wmclaugh, line 15985:
```suggestion
           no longer forces two rows.
```

PR #2129, opened 2024-09-14
Fix plot overplot

### LINKED ISSUE #2128: plot with overplot
This has long been an annoyance but answering

- #2119 

made me write this up.

With 4.16 we have (the two datasets are the same for this example):

```
>>> plot("data", "data", 2)
```

sherpa-1

If I try to use `overplot` then I get the following, which doesn't make much sense

```
>>> plot("model", "model", 2, overplot=True)
```

sherpa-2

So, the original data has been removed. Now, there is the confusing `clearwindow` argument, but

```
>>> plot("data", "data", 2)
>>> plot("model", "model", 2, overplot=True, clearwindow=False)
```

produces the same display.

I will note the slightly-different result from

```
>>> plot("data", "data", 2)
>>> plot("model", "model", 2, clearwindow=False)
```

sherpa-3

which does not  hide the plot axes/title.

DougBurke:
# Summary

Improve the behaviour of plot() and contour() when called with overplot=True or overcontour=True respectively. Fix #2128.

# Details

Fortunately with the changes made in #2126, in particular sorting out the rows/cols settings and dropping the special case of the "singleton" plot code, this turned out to be easier than  I'd thought. 

So, this PR needs #2126.

So, we need to be able to ask the SplitPlot class what

- the current number of rows and columns are
- how many plots are currently active
- not to clear the window when resetting

At present there is no actual API to do these, so we rely in access to internal fields of the class. Ideally we would add an API but

a) the UI code is basically the only real user of the SplitPlot API at the moment (the JointPlot class subplots SplitPlot but adds it's own API which means we skip this complexity for commands like plot_fit_resid)
b) it's not entirely clear what API we want, so I'd rather live with this code for now until we get more experience of what we want

I will say that the existing SplitPlot API - e.g. addplot, overplot, plot, is a bit covfusing and perhaps could be rewritten to better match the plot vs overplot needs, but I do not think it worth working on at this time.

# Examples

```
>>> plot("data", "data", 2, cols=1)
>>> plot("model", "model", 2, overplot=True)
```

sherpa-1

Now, what should happen when you supply too few plots in the overplot call - e.g.

```
>>> plot("data", "data", 2, cols=1)
>>> plot("model", overplot=True)
```
sherpa-2

For now we allow this, since it didn't seem worth making this an error.

However we do error out if you try to add too many plots [the error message isn't great, but we have overloaded the word "plot" here, particularly if this were a `contour(...)` call] - e.g.

```
>>> plot("data", "data", 2)
>>> plot("model", 1, "model", 2, "model", 3, overplot=True)
...
ArgumentErr: Too many elements for overplot
```

The following, where you try to change the plot layout, is ignored (so the cols and rows arguments are ignored when overplot/contour are given):

```
>>> plot("data", "data", 2, cols=1)
>>> plot("model", "model", 2, rows=5, cols=4,  overplot=True)
```

We could make this an error, but is it worth it?

### DIFF
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -13734,6 +13734,11 @@ class Session(NoNewAttributesAfterInit):
         - try to go with a square display
         - fill in columns first
 
+        The handling of overplot/contour has been improved - see issue
+        #2128 - but this involves using internal features of the
+        SplitPlot class. This could be improved, once we have more
+        experience of how this all works.
+
         """
 
         # This is an internal routine so it is not expected to be
@@ -13755,9 +13760,9 @@ class Session(NoNewAttributesAfterInit):
         check = getattr(self, f"_check_{plotmeth}type")
 
         plots = []
-        args = list(args)
-        while args:
-            plottype = args.pop(0)
+        largs = list(args)
+        while largs:
+            plottype = largs.pop(0)
             _check_str_type(plottype, "plottype")
             plottype = get(plottype.lower())
 
@@ -13766,11 +13771,11 @@ class Session(NoNewAttributesAfterInit):
             # plot/contour type.
             #
             getargs = []
-            while args:
-                if check(args[0]):
+            while largs:
+                if check(largs[0]):
                     break
 
-                getargs.append(args.pop(0))
+                getargs.append(largs.pop(0))
 
             funcname = f"get_{plottype}_{plotmeth}"
             getfunc = getattr(self, funcname)
@@ -13783,6 +13788,14 @@ class Session(NoNewAttributesAfterInit):
 
         nplots = len(plots)
 
+        # What is the overplot/contour handling? To address #2128 we
+        # do not need to worry about the clearwindow option.  Note
+        # that the value (assumed to be a scalar) is not removed from
+        # kwargs.
+        #
+        over = kwargs.get(f"over{plotmeth}", False)
+        # clearwindow = kwargs.get("clearwindow", True)
+
         # Deconstruct the keyword arguments.
         #
         kwstore = get_per_plot_kwargs(nplots, kwargs)
@@ -13793,9 +13806,25 @@ class Session(NoNewAttributesAfterInit):
         nrows_orig = sp.rows
         ncols_orig = sp.cols
 
-        # What size to use?
+        plotfunc = getattr(sp, f"add{plotmeth}")
+
+        # We only care about the size when creating the display, that
+        # is, if overplot/contour is set then we assume the size is
+        # known. There is a check to make sure the user isn't trying
+        # to overplot too-many plots, but we allow less plots to be
+        # overplot than the original.
         #
-        if rows is None and cols is None:
+        # Should this care about clearwindow too?
+        #
+        if over:
+            # Dig inside the SplitPlot class to find this out.
+            #
+            nrows, ncols = sp._used.shape
+            nplots_used = sp._used.sum()
+            if nplots_used < nplots:
+                raise ArgumentErr(f"Too many elements for over{plotmeth}")
+
+        elif rows is None and cols is None:
             # Special case the single-plot call
             if nplots == 1:
                 nrows = 1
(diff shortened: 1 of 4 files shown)


hamogu (approved):

sherpa/ui/utils.py
```diff
@@ -320,6 +320,55 @@ def notice_data_range(get_data: Callable[[IdType], Data],
...
+
+def calc_multiplot_size(rows, cols, nplots):
+    """How many rows and columns to use for multi plot/contours?
+
+    Parameters
+    ----------
+    rows, cols
+       The requested sizes, if set (one of them must be set).
+    nplots
+       The nuber of plots.
```
wmclaugh, line 345:
```suggestion
       The number of plots.
```

sherpa/ui/utils.py
```diff
@@ -13576,10 +13641,12 @@ def _multi_plot(self, args, plotmeth="plot", **kwargs) -> None:
...
         Parameters
         ----------
-        args : list
+        args
             The arguments to the call. It can not be empty.
-        plotmeth : {"plot", "contour"}
+        plotmeth
             The call.
+        rows, cols
+            The number of rows or columns (to over-ride split plot).
```
wmclaugh, line 13649:
```suggestion
            The number of rows or columns (to override split plot).
```

sherpa/ui/utils.py
```diff
@@ -13654,20 +13745,74 @@ def _multi_plot(self, args, plotmeth="plot", **kwargs) -> None:
...
+        over = kwargs.get(f"over{plotmeth}", False)
+        # clearwindow = kwargs.get("clearwindow", True)
 
-        nrows = 2
-        ncols = int((len(plots) + 1) / 2.0)
+        # Deconstruct the keyword arguments.
+        #
+        kwstore = get_per_plot_kwargs(nplots, kwargs)
+
+        # Store the original values in case they are over-written.
```
wmclaugh, line 13762:
```suggestion
        # Store the original values in case they are overwritten.
```

sherpa/ui/utils.py
```diff
@@ -13654,20 +13745,74 @@ def _multi_plot(self, args, plotmeth="plot", **kwargs) -> None:
...
-        plotmeth = getattr(sp, f"add{plotmeth}")
+        nrows_orig = sp.rows
+        ncols_orig = sp.cols
+
+        plotfunc = getattr(sp, f"add{plotmeth}")
+
+        # We only care about the size when creating the display, that
+        # is, if overplot/contour is set then we assume the size is
+        # known. There is a check to make sure the user isn't trying
+        # to overplot too-many plots, but we allow less plots to be
```
wmclaugh, line 13773:
```suggestion
        # to overplot too many plots, but we allow less plots to be
```

sherpa/ui/utils.py
```diff
@@ -13885,6 +14034,13 @@ def plot(self, *args, **kwargs) -> None:
         identifier is given for a plot type, the default identifier -
         as returned by `get_default_id` - is used.
 
+        .. versionchanged:: 4.17.0
+           The keyword arguments can now be set per plot by using a
+           sequence of values. The layout can be changed with the rows
+           and cols arguments and the automatic calculation no-longer
```
wmclaugh, line 14040:
```suggestion
           and cols arguments and the automatic calculation no longer
```

sherpa/ui/utils.py
```diff
@@ -15811,28 +16022,36 @@ def contour(self, *args, **kwargs) -> None:
         as returned by `get_default_id` - is used. This is for
         2D data sets.
 
+        .. versionchanged:: 4.17.0
+           The keyword arguments can now be set per plot by using a
+           sequence of values. The layout can be changed with the rows
+           and cols arguments and the automatic calculation no-longer
```
wmclaugh, line 16028:
```suggestion
           and cols arguments and the automatic calculation no longer
```

sherpa/ui/utils.py
DougBurke: These are from #2126 so once that has been merged (as I've just fixed them) there we can rebase or fix-up this PR to pick them up.

sherpa/ui/utils.py
DougBurke: ditto.

sherpa/ui/utils.py
DougBurke: I'm going to wait for #2126 to get merged before fixing the other typos.

DougBurke:
Rebased to address the conflict from the doc fixes made in #2126 - some of which were noted here - and to address the remaining doc issues thaw @wmclaugh had noted (and I took the time to scrub both the sherpa/ui/utils.py and sherpa/astro/ui/utils.py to replace "no-longer" with "no longer" and "over-ride" with "override").

PR #2121, opened 2024-09-04
UI: add calc_model/source commands

DougBurke:
# Summary

Add calc_model and calc_source routines to return the model expression evaluated per bin.

# Details

If you want to evaluate a model to match the data then the easiest way has been to call get_model_plot and then use the y attribute of the plot object. However, this is annoying because users do not always want to use the plot options - e.g. normalize by the exposure time or divide by the bin width.

So, we add

    calc_model()
    calc_source()

routines which evaluate the model on the relevant grid and return

- the independent axis (if integrated will get both edges)
- the evaluated model value

The return values depend in part on the data object - e.g. the analysis setting for PHA data - and in part the model (i.e. what are its units).

There is an asymmetry in these two routines, since

- calc_model will filter (and group) the data
- calc_source does not filter (and can not group)

We could apply a filter to the calc_source data to "match" the chosen data filter, but it is not ideal (since notice/ignore act on the EBOUNDS energy grid and this is the MATRIX energy grid, which is technically different).

An alternative would be to follow plot_source and add lo/hi arguments so a user can say "only give me the data between these two limits" - but it's easy for a user to do this themselves, and it wouldn't reduce the time taken to make the call.

We **could** have the independent axis get returned as a `PointAxis`/`IntegratedAxis` object,  which may help users, but is also likely to just annoy them.

# Example

```
>>> from sherpa.astro import ui
>>> ui.load_pha("sherpa-test-data/sherpatest/3c273.pi")
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273.pi'
but not used; to use them, re-read with use_errors=True
read ARF file sherpa-test-data/sherpatest/3c273.arf
read RMF file sherpa-test-data/sherpatest/3c273.rmf
WARNING: systematic errors were not found in file 'sherpa-test-data/sherpatest/3c273_bg.pi'
statistical errors were found in file 'sherpa-test-data/sherpatest/3c273_bg.pi'
but not used; to use them, re-read with use_errors=True
read background file sherpa-test-data/sherpatest/3c273_bg.pi
>>> ui.set_source(ui.xsphabs.gal * ui.powlaw1d.pl)
>>> ui.get_analysis()
'energy'
```

Lets get the model values:

```
>>> xvals, yvals = ui.calc_model()
```

What is returned? Well, no filter has been applied so it returns each grouped bin - in this case there are 46 of them - and the units of this array are counts:

```
[...]

PR #2131, opened 2024-09-17
UI: calc_model/source now return Axis objects

DougBurke:
This is an experiment to see whether we should use the `Axis` class in more user-facing situations.

# Summary

The calc_source and calc_model functions now return the independent axis as Axis objects rather than plain ndarrays.

# Details

In #2121 we added `calc_source` and `calc_model` commands to the astro UI layer to make it easier to get the "actual" values for a model/source, rather than have to dreconstruct them from `get_model_plot()` [not quite `get_fit_plot().modelplot` as that is grouped and we don't support that here yet].

One question I has was whether we should use an internal class  (`sherpa.models.regrid.Axis`) for returning the X / independent axis, rather than just a tuple of numpy arrays. Well, this PR asks the question of whether it's worth it.

I like it, since to me it's more obvious, but I can imagine that actually using this would be annoying (since generally you know if you are getting a single value or a pair).

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -48,6 +48,7 @@ from sherpa.sim import NormalParameterSampleFromScaleMatrix
 from sherpa.stats import Cash, CStat, WStat
 from sherpa.models.basic import TableModel
 from sherpa.models.model import Model
+from sherpa.models.regrid import Axis, IntegratedAxis, PointAxis
 from sherpa.astro import fake
 from sherpa.astro.data import DataIMG, DataIMGInt, DataPHA
 import sherpa.astro.instrument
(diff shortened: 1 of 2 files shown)


sherpa/astro/ui/utils.py
```diff
@@ -15783,12 +15784,12 @@ def calc_model(self,
...
         -------
-        xvals, yvals: tuple of ndarray, ndarray
-           The independent axis, which uses a tuple as the number of
-           elements depends on the dimensionality and type of data.
-           The units depends on the data type: for PHA data the
-           X axis will be in the analysis units and Y axis will
-           generally be counts.
+        xvals, yvals: tuple of Axis, ndarray
+           The independent axis, where the size matches the data
+           dimentionality and the values are Axis objects (so point or
```
wmclaugh, line 15789:
```suggestion
           dimensionality and the values are Axis objects (so point or
```

anetasie:
@DougBurke would this be in any way related/helpful for regridding the model?

DougBurke:
@anetasie - this is an experiment I made to see how it "felt". There are, I think, reasonable reasons to do it, but

1. it does not change the **information** that `calc_model` and `calc_source` returns
2. for most users they will not need this level of "indirection" and would probably prefer just to get the data directly as one (or more) NumPy arrays

It has nothing to do with regridding models (as that is more an issue for the arguments you sent to calls like `calc_model` rather than the return values for these calls).

Issue #1818, opened 2023-07-17
doctestplus tests fail when no I/O module is available


DougBurke:
I'm running tests with a minimal install (no matplotlib or astropy) and they are fine, but if I then do

    pip install -r test_requirements.txt

(which just installs `pytest-doctestplus`) then I get failures from `sherpa/astro/data.py` with errors like

```
_________________________________ [doctest] sherpa.astro.data.DataPHA.sum_background_data _________________________________
3805         to divide by the number of background files used.
3806 
3807         Example
3808         -------
3809 
3810         Calculate the background counts, per channel, scaled to match
3811         the source:
3812 
3813         >>> from sherpa.astro.io import read_pha
3814         >>> pha = read_pha(data_3c273 + '3c273.pi')
UNEXPECTED EXCEPTION: NotImplementedError('No usable I/O backend was imported.')
Traceback (most recent call last):
  File "/lagado2.real/local/anaconda/envs/sherpa-main/lib/python3.11/doctest.py", line 1351, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest sherpa.astro.data.DataPHA.sum_background_data[1]>", line 1, in <module>
  File "/lagado.real/sherpa/sherpa-main/sherpa/astro/io/__init__.py", line 528, in read_pha
    datasets, filename = backend.get_pha_data(arg,
                         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lagado.real/sherpa/sherpa-main/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.
/lagado.real/sherpa/sherpa-main/sherpa/astro/data.py:3814: UnexpectedException
```

Installing `astropy` fixes this, but it's not ideal. I don't know why this doesn't cause problems for the "minimal" CI runs which do not install `astropy`, as I thought these `doctest` tests were being run on CI.

DougBurke:
Aha - I've now triggered this on CI as part of working on #1816

hamogu:
Well, without a backend, we can't read files. Those examples need to be marked:

See "Doctest dependencies" in https://github.com/scientific-python/pytest-doctestplus
There are several options, e.g. in setup.cfg or in the file where it's used e.g.:

```Python
doctest_subpackage_requires =
    sherpa/astro/data.py = astropy
```

[...]

PR #2137, opened 2024-09-18
Fix doctests in `sherpa.plot.__init__`

hamogu:
Our backends are now objects and not modules any longer.

I don't know why this caused a failure locally, but did not show up in CI before, in other words, I'm not sure what caused this file to be excluded from doctests in CI, while running them locally.

### DIFF
--- a/sherpa/plot/__init__.py
+++ b/sherpa/plot/__init__.py
@@ -37,6 +37,7 @@ backend. The following command prints the name of the backend:
 
    >>> from sherpa import plot
    >>> print(plot.backend.name)
+   pylab
 
 Change the backend
 ------------------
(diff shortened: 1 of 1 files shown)


DougBurke:
Is #1818 related (currently in a meeting so can't easily check)?

hamogu:
It's related only in the sense that both is about plotting backends. In #1818 a backend is not found, here the test fails because the docs are actually wrong (outdated). What I don't know is why this file run (and failed locally) but not in CI - but I'd rather fix it then do forensics on why that is.

This is exactly the case that doctests are supposed to catch - we change the code and then the example we give in the docs is now wrong. But we know that our doctest coverage is far from complete because we've not rewritten many of the old docs (e.g. to make sure that all required packages are imported etc.), so in CI we run only on a subset of files and apparently not this one.

In contrast, #1818 is about running a doctests that are correct, but fail because no backend is installed in that particular environment.

DougBurke (approved):
I have a dislike of the doctest stuff because of things like this, so fixes are more than welcome!

PR #2141, opened 2024-09-19
Update LaTeX labels and docs for bokeh

hamogu:
## Summary
Correctly render LaTeX labels in bokeh

## Details
Previously, bokeh could only set an entire string in LaTeX, but now it can also do that for substrings in a label, so update code and docs for that.

The main motivation is to improve the docs for bokeh with instructions on LaTeX labels (which are common in astronomy), but take the opportunity to make the plots docs past doctests.

### DIFF
--- a/docs/plots/bokeh_backend.rst
+++ b/docs/plots/bokeh_backend.rst
@@ -68,17 +68,33 @@ The same is true for the UI interface::
     >>> from sherpa.astro import ui
     >>> from sherpa import plot
     >>> from sherpa.astro.data import DataPHA
-    >>> session = Session()
     >>> ui.load_arrays(1, [1, 2, 3], [1, 2, 3], DataPHA)
     >>> ui.plot_data()
     >>> show(plot.backend.current_fig)
     >>> # or
-    >>> save(plot.backend.current_fig, filename='plot.html')
+    >>> save(plot.backend.current_fig, filename='plot.html')  # doctest: +SKIP
 
 Suggestions on how to integrate that better into the Sherpa UI are welcome.
 
 
+Differences to matplotlib
+-------------------------
+The :term:`bokeh` backend is intrinsically different from the :term:`matplotlib` backend,
+but Sherpa will automatically translate some settings.
+For example, if the color `'k'` (black) is passed into a sherpa plotting routine, it
+will be internally translated to bokeh's representation of black (`'black'`).
+However, this is limited to a subset of the available options that are commonly
+used, see :ref:`backend-independent-plotting-options`.
+A user can pass a additional backend specific options to the plotting
+routines, such as `marker="*"` in matplotlib. This will cause an error in
+bokeh, as bokeh does not have a `"*"` marker; a similar look can be achieved
+with `marker="star"`.
 
+Similarly, in bokeh, text for labels can be formatted with LaTeX by
+enclosing it in `'$$'` (e.g. `label=r"$$\alpha$$"`), while in matplotlib
+it is done with `label=r"$\alpha$"`. All labels that sherpa itself sets will be set
+correctly, but if a user wants to manually set a label, they need adjust it
+for the backend that is currently active.
 
 Limitations
 -----------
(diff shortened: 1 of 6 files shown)


docs/plots/index.rst
```diff
@@ -300,12 +300,7 @@ line position (which corresponds to `mdl.pars[2]`):
     >>> from sherpa.plot import IntervalProjection
     >>> iproj = IntervalProjection()
     >>> iproj.calc(f, mdl.pars[2])
-    WARNING: hard minimum hit for parameter base.c0
-    WARNING: hard maximum hit for parameter base.c0
-    WARNING: hard minimum hit for parameter line.fwhm
-    WARNING: hard maximum hit for parameter line.fwhm
-    WARNING: hard minimum hit for parameter line.ampl
-    WARNING: hard maximum hit for parameter line.ampl
+    ...
```
hamogu, line 303: This change is here for two reasons: 

1. We actually talk about plotting changes with a pretty dumb data setup. I don't want to discuss *why* we get these warnings and why they can be ignored.
2. Our warnings are not emitted by the warnings module (which captures in doctests and the output chan be compared using pytest-doctestplus) but by a Logger at the warning level. Thus, it's non-trivial to make that work for a doctest and I think the advantage of testing (and finding e.g. the spelling error in "TemporarypPlottingBackend" above) is bigger than displaying these warnings that are meaningless in the context of this example.

sherpa/plot/bokeh_backend.py
```diff
@@ -197,6 +197,24 @@ def __exit__(self, exec_type, value, traceback):
...
+
+        Returns
+        -------
+        latex : str
+            The text modified as appropriate for a backend so that the LaTeX
+            will be displayed properly.
+
+        """
+        return f"$${txt}$$"
+
```
DougBurke, line 217: So, does this imply a minimum version of bokeh we support (or that if we have an older version we just don't get nice plot labels)?

sherpa/plot/backends.py
```diff
@@ -612,7 +612,7 @@ def get_latex_for_string(self, txt):
         Returns
         -------
         latex : str
-            The text modified as appropriate for a backend so that the LaTeX
+            The test modified as appropriate for a backend so that the LaTeX
```
DougBurke, line 615: Err, I think the original text (of `text`) was correct here. Or am I missing something?

sherpa/plot/bokeh_backend.py
```diff
@@ -209,7 +209,7 @@ def get_latex_for_string(self, txt):
         Returns
         -------
         latex : str
-            The text modified as appropriate for a backend so that the LaTeX
+            The test modified as appropriate for a backend so that the LaTeX
```
DougBurke, line 212: see previous comment

sherpa/plot/bokeh_backend.py
hamogu: For an older version it's just going to look like it does in #1973 (except that you'll see a double $$, instead of a single $). So, no formal minimum version, but this will make it look better on newer version (not sure the exact number where they changed that; there is an entry in the changelog for bokeh 3.4 that could be related, but I don't really want to spend the time to read through and understand the full PR to see if that's the right one)

docs/plots/bokeh_backend.rst
```diff
@@ -68,17 +68,33 @@ The same is true for the UI interface::
...
-
+Differences to matplotlib
+-------------------------
+The :term:`bokeh` backend is intrinsically different from the :term:`matplotlib` backend,
+but Sherpa will automatically translate some settings.
+For example, if the color `'k'` (black) is passed into a sherpa plotting routine, it
+will be internally translated to bokeh's representation of black (`'black'`).
+However, this is limited to a subset of the available options that are commonly
+used, see :ref:`backend-independent-plotting-options`.
+A user can pass a additional backend specific options to the plotting
```
wmclaugh, line 88:
```suggestion
A user can pass an additional backend specific options to the plotting
```

sherpa/plot/backends.py
hamogu: Force pushed to change back.

docs/plots/index.rst
DougBurke: I think when I added this text I didn't really grok the fact we can use "..." to say "skip text when testing"

sherpa/plot/bokeh_backend.py
DougBurke: Just wanted to check if we needed to add a note about bokeh version, but it seems like it's not worth it.

DougBurke (approved):

PR #2142, opened 2024-09-20
Bad

DougBurke:
# Summary

Nope. The more I look at this the more I just don't understand the original design ...

# Details

I've been rushing to get improvements for `ignore_bad` in, as I suddenly realised it could be done. Then it couldn't. Then it could. Then maybe.

So this is the minimal-usable-output of #2142 and attempts to better integrate the quality filtering in with the existing data, rather than just randomly  inserted into some of the code paths. Now, the two big issues with handling "quality-filtered" data are

1. do we treat this just as we do the existing `ignore/notice/mask` case, since no-matter how you do it you have the possibility of a mis-match between data and a filter
2. how do we differentiate the "definitely" bad versus "operationally" bad values (1/5 versus 2)

It really gets hard when we come to working with an interactive environment, since what happens if you do

    ignore_bad()

 which filters out bad data, and then

    group_counts(15) 

which hopefully makes use of the quality filter. Well, one of the results of the group_xxx calls is that they change the quality array (although to maybe the same values), and hence should we now be using the new quality filters or should it require the user to call `ignore_bad` explicitly. Either way is confusing...

There's also fun when filtered-out data lies within a valid group, or you try to filter out a set of channels that are marked as bad quality. Well, fun if you enjoy making no progress.

So, for now we are punting on most of these and just trying to make the existing code "work better".

It should not change any code that doesn't call `ignore_bad`. Anuy code that does use `ignore_bad` is likely to be very finickity so worth a re-look with this new code.

# Examples

Some fun stuff: in 4.16 we have

```
sherpa-4.16.0> load_arrays(1, [1, 2, 3], [2, 3, 0], DataPHA)

sherpa-4.16.0> set_quality([0, 0, 2])

sherpa-4.16.0> get_dep(filter=True)
array([2., 3., 0.])

sherpa-4.16.0> ignore_bad()
dataset 1: 1:3 -> 1:2 Channel

sherpa-4.16.0> get_dep(filter=True)
array([2., 3.])

sherpa-4.16.0> notice()
dataset 1: 1:2 -> 1:3 Channel

sherpa-4.16.0> get_dep(filter=True)
array([2., 3., 0.])
```

Note that the call `notice()` has removed the quality filter. I can understand it, but it's also very annoying for it to clear out the quality array. We should have a separate way to restore the "bad" filter. In this branch we get

```
TO BE CHECKED
```

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -130,7 +130,7 @@ import numpy as np
 
 from sherpa.astro import hc
 from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
-    IntegratedDataSpace2D, _check
+    Filter, IntegratedDataSpace2D, _check
 from sherpa.models.regrid import EvaluationSpace1D
 from sherpa.stats import Chi2XspecVar
 from sherpa.utils import pad_bounding_box, interpolate, \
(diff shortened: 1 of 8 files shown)


DougBurke:
This code is starting to diverge from the main branch, as some of the code (e.g. tests) maay have made its way into the code base and other changes in the data class make merging the code not as easy as it could be. I'm likely to drop this PR soon-ish.

PR #2143, opened 2024-09-20
Improve pre-commit script for ranges of years

hamogu:
## Summary
Improve pre-commit script for ranges of years

## Details
We often use something like "#  Copyright (C) 2010, 2016 - 2023" and the previous version of the script did not deal with that.

Now, we do. While this is not crucial functionality, is annoyed me so much in development, that I decided that it was faster to fix here than to change the copyright years by hand.
In programming this, I made the choice that even a two year span will be listed as a range (i.e. "2023-2024" instead of "2023, 2024").
Note that this script is opt-in in the sense that developers need to copy it into `.git/hooks` for it to run as an automatic pre-commit step.

Note that this script it only used by developers who will generally work the main branch. There is no rason to ruch this through before te release, but I'm opening the PR so we don't forget.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -62,6 +62,7 @@ def combine_years(list_years: list[int]) -> str:
     This combines consecutive years into ranges, e.g. [2010, 2011, 2012]
     becomes "2010-2012".
     '''
+    list_years = list(set(list_years))
     list_years.sort()
     combined = []
     start = list_years[0]
@@ -102,7 +103,7 @@ if __name__ == '__main__':
                     list_years.append(year)
                     changed = 1
                     print(m.group('comment') +
-                          '  Copyright (C) ' + combine_years(list_years), end='')
+                          '  Copyright (C) ' + combine_years(list_years), end='\n')
                     # sometimes group institution will be empty, because it's
                     # on the next text line. If not, move it there.
                     if m.group('institution'):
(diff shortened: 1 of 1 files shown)


hamogu:
It's just the line break. I'll add that back in. Don't know why I didn't see that in my case.

scripts/pre-commit
```diff
@@ -27,14 +27,64 @@ import re
...
+            list_years.extend(range(int(start), int(end) + 1))
+        else:
+            list_years.append(int(y))
+    return list_years
+
+
+def combine_years(list_years: list[int]) -> str:
+    '''Combine a list of years into a string
+
+    This combines consecutive years into ranges, e.g. [2010, 2011, 2012]
```
wmclaugh, line 62: It would be nice to update the code to support duplicates in the input date list (covert the list to be unique). ie... 

    >>> ll=[2011,2012,2010,2011,2016,2012,2015]
    >>> combine_years(ll)
   '2010-2011, 2011-2012, 2012, 2015-2016'

But I would expect the output to look like:

>>> combine_years(list(set(ll)))
'2010-2012, 2015-2016'

hamogu: Since I know that we have a well-kept codebase with careful editing of the list of years in the copyright notice in every file, I was absolutely certain that there can never be an input that is unsorted or contains duplicate entries ;-)

hamogu:
@wmclaugh I think I've resolved your concern in my last commit. I did not want to push that before the release, but since we're back in working mode now, maybe you can have another look and we can get this merged?

Issue #1564, opened 2022-08-25
Remove `bin_lo` and `bin_hi` from `DataPHA`?


hamogu:
Those two fields exist in Chandra grating data pha files. They are not OGIP standard, but OGIP allows PHA files to have additional columns, so it's not wrong to have them in the files. They were intended to give approximate wavelength boundaries for grating files so that the PHA files are readable by themselves (speaking as a member of the CXC/HETG team here). In the context of Sherpa, they are not required, but curiously they are given preference over the values in the ARF/RMF: https://github.com/sherpa/sherpa/blob/main/sherpa/astro/data.py#L2118

In some sense, they allow the analysis of PHA data without an ARR and RMF. For observed data, that should not happen since we always have ARFs and RMFs; if not, one can use `make_diagonal_rmf` to make one. So, this is more useful in development to make a quick PHA plot without loading the responses. 

It is possible that users might use them on their data objects, but they are not advertised in our docs as far as I can see, they are not easily accessible from the UI layer, and users really should have an ARF/RMF.

I suggest to deprecate them, i.e. default to using the ARF/RMF grid in the code linked above, issue a warning message when they are used (maybe with "consider using `make_diagonal_rmf` if you just want to plot the data on some grid") and remove them one or two years later.

Why do I care? Reducing the number of special cases makes it easier to maintain the code.

anetasie:
> In some sense, they allow the analysis of PHA data without an ARR and RMF. For observed data, that should not happen since we always have ARFs and RMFs; if not, one can use make_diagonal_rmf to make one. So, this is more useful in development to make a quick PHA plot without loading the responses.

@hamogu there are cases where you want to use 'make_diagonal_rmf', but fit data with the ARF only. Is this PR removes that functionality?

hamogu:
This is not a PR, just an issue that note the problem, but no, it would not remove that functionality. In that case, you still need to load and assign ARF and an RMF (just, the RMF would not come from a file, but from the `make_diagonal_RMF` function that we already have in Sherpa. If would remove the case, where you load just the PHA file, and no ARF or RMF, which effectively means that you are fitting with ARF = 1 cm^2 and RMF=diagonal. Now, there might still be cases where someone wants to do that, but that someone would only be people who do HETG calibration (i.e. me). Those people could mock up an ARF grid (we have functions for that in Sherpa, too, that we uses in tests) with ARF=1 cm^2 in those very rare cases - but it should not be the default.

anetasie:
[...]

PR #2147, opened 2024-09-22
Improve corner-case handling of filtering and empty data handling

DougBurke:
# Summary

# Details

This is based on #2118, so the first seven commits are from that PR.

This is work I did to try to improve the "bad handling" of PHA code but it is not directly relevant to that project and can be taken as is.

The commits are

- add a number of "bad data" tests (that are not going to be addressed in this PR but useful to have just to check nothing I do here changes them)
- move `__init__` to the start of the `DataPHA` class
  - this is technically a style change, but I have a number of type issues related to the DataPHA class and the many properties that are likely to be addressed by moving this first (so we can define types here)
- I really would like to remove the "we can make an empty dataset" capability, as it leads to lots of issues about what to do when we send in an empty dataset. However, now is not the time to address this. What we can do is explicitly error out if given an empty dataset. This leads to a few minor improvements in the error that gets raised (i.e. explictly saying "this is an empty dataset" rather than an internal error like "you are trying to add 1 to None")
  - this actually changes the Data1DInt behaviour, in that it now errors out rather than returning something, but I claim this is better since it matches the other Data1D classes now, and it is for a "not very important" method (`get_dims`).
- add some "guard rails" to the `DataPHA.quality_filter` field: we don't have a lot of documentation about what this is supposed to do, so this starts to add some (it is used by the `ignore_bad` code, hence it's inclusion in the genesis of this work); at present this only really afects some test code I wrote to catch what setting this field to "silly" values would do
- add some typing to the `Filter` class; this is more so we can catch potential future issues
- the main "useful" change in this PR, and it is related to issue #1220 
  - our mask handling is `mask is True`, `mask is False`, or `mask is an array`
  - `mask is True` and `mask.all() is True` should be isomorphic
  - `mask is False` and `not mask.any() is True` should be isomorphic
  - however, we have historically had subtle differences in the way the code handles this, and this commit tries to improve on this
  - I thought about automatically replacing a mask of all True or all False to the boolean flag, OR vice versa always using an array rather than the special case boolean option, but both of them would require a lot of work to get the tests to pass and this is way too risky to think about at this time
  - so we are "stuck" with these fixes
  - I don't guarantee they catch all cases, but they improve the code base (I claim)
- I then try to add typing statements to `DataPHA`; I manage this for virtually all the code, except for the `exposure`, `areascal`, and `backscal` fields. I would expect them to be easy, **but** I have been burnt too many times by thinking this and would really prefer to work on these in a later PR (since there are issues we want to resolve over how these fields work)


sherpa/astro/data.py
```diff
@@ -2541,10 +2601,17 @@ def _get_ebins(self,
...
+        # is expected to be called with channels set.
+        #
+        if self.channel is None:
+            raise DataErr(f"data set '{self.name}' has no channel information")
+
         if self.units == 'channel':
             elo = self.channel
             ehi = self.channel + 1
         elif (self.bin_lo is not None) and (self.bin_hi is not None):
+            # TODO: review whether bin_lo/hi should over-ride the response
```
hamogu, line 2614: We should just remove bin_hi / bin_lo and always require a response. But not in this PR.

sherpa/astro/data.py
```diff
@@ -3453,7 +3558,10 @@ def ignore_bad(self):
         # self.quality_filter used for pre-grouping filter
         self.quality_filter = qual_flags
 
-    def _dynamic_group(self, group_func, *args, **kwargs):
+    def _dynamic_group(self,
+                       # TODO: remove callable support
```
hamogu, line 3562: I would advocate to ONLY have callable support and move grouping from C to Python ... I don't know if this is the method to do it, but either way it's not for this PR. However, maybe we can discuss that elsewhere not here.

sherpa/astro/data.py
```diff
@@ -3468,6 +3576,9 @@ def _dynamic_group(self, group_func, *args, **kwargs):
         If group_func is a callable then it must return the grouping
         and quality arrays for the new scheme.
 
+        Support for group_func being a callable has been marked as
+        deprecated in 4.17.0 and will be removed in the next release.
```
hamogu, line 3580: Can we say "may be removed" since I'm not convinced yet?

hamogu (changes requested):
Since this builds on a different PR, I've reviewed it "per commit" so I don't have to read the other changes again. Unfortunately, that means that the GH interface does not allow me to make suggestions in most cases, it's just comments. 
In general, this is absolutely a stop in the right direction to make DataPHA less fragile and less unpredictable.

sherpa/astro/data.py
```diff
@@ -1992,6 +1992,41 @@ def quality(self, val):
...
+    # internally with ignore_bad?
+    #
+    @quality_filter.setter
+    def quality_filter(self, val: Optional[ArrayType]) -> None:
+        if val is None:
+            qfilt = None
+        else:
+            # ensure we store boolean values
+            qfilt = np.asarray(val, dtype=bool)
+        self._set_related("quality_filter", qfilt)
```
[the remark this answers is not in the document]
hamogu: Or maybe I just overlooked it. Either way, I would say that you can set the filter and the grouping by hand, but the quality filter is something that's inherent in the data. If you need to change that, change the quality array and reapply the filter. (Side note: Should be automatically do something, like reset or re-apply the quality filter when the quality array is changed?)
I just feel that there is no reasonable use case to set this filter directly and it's better to offer fewer overlapping setters (grouping, ignore/notice, and quality_filter all change which bins are used in fitting).

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -2307,17 +2736,129 @@ def test_grouped_pha_get_y(make_grouped_pha):
...
+    """What is the default get_mask value?"""
+    pha = make_grouped_pha
+    assert pha.get_mask() == pytest.approx([True] * 4 + [False])
+
+
+def test_quality_pha_mask(make_quality_pha):
+    """What is the default mask setting?"""
+    pha = make_quality_pha
+    assert np.isscalar(pha.mask)
+    assert pha.mask
```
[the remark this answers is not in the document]
DougBurke: Technically the interface claims it returns a boolean or an array, but we can make these changes to be explicit.

sherpa/astro/data.py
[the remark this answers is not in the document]
DougBurke: Well, `quality_filter` is a field that already exists. We can make it read-only (and that would likely remove some of the complexities), but I think that is only really worth it if we improve the quality handling (which this PR is explicitly not about).

sherpa/astro/data.py
DougBurke: For completeness, I will link to 

- #1564 
- #1871 

as it can be useful to see what the discussions are.

sherpa/astro/data.py
DougBurke: I can just remove this comment since this is really an internal routine so anyone calling it has to be aware of it being changed randomly!

sherpa/astro/data.py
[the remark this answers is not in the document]
DougBurke: Argh. Removing the setter and making the routines that need to change `self._quality_filter` directly fails because there's some code in `sherpa.astro.utils` that change the `quality_filter`. It is likely that this extra code can be re-written somehow, but that's not going to happen now, so I think we are stuck with this.

DougBurke:
Rebased to address conflicts from #2118 (although there were actually no conflicts to address, so ....) and to address the review comments from @hamogu

sherpa/astro/data.py
DougBurke: I've removed all the comments about removing the callable option.

sherpa/astro/data.py
DougBurke: I have removed this comment

sherpa/astro/data.py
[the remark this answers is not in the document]
DougBurke: So I've left it as is for now.

hamogu (approved):
The changes address my suggestion. I'm approving, assuming that the tests pass.

PR #2155, opened 2024-09-24
Drop support for Python 3.9 and add 3.12

Marie-Terrell:
Summary
========
Drop support for Python 3.9 and numpy < 1.24 and note Python 3.12 experimental support.

Details
======
This PR:
- Drops support for Python 3.9 and numpy < 1.24
- Adds Python 3.12 conda packages and notes that it is experimental

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -38,17 +38,17 @@ jobs:
             matplotlib-version: 3
             xspec-version: 12.14.0i
 
-          - name: Linux Minimum Setup (Python 3.9)
+          - name: Linux Minimum Setup (Python 3.10)
             os: ubuntu-latest
-            python-version: "3.9"
-            numpy-version: "1.21"
+            python-version: "3.10"
+            numpy-version: "1.24"
             install-type: develop
             test-data: none
 
           - name: Linux Minimum Setup (Python 3.11)
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-version: "1.23"
+            numpy-version: "1.24"
             install-type: develop
             test-data: none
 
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-conda-workflow.yml
```diff
@@ -38,17 +38,17 @@ jobs:
...
+          - name: Linux Minimum Setup (Python 3.10)
             os: ubuntu-latest
-            python-version: "3.9"
-            numpy-version: "1.21"
+            python-version: "3.10"
+            numpy-version: "1.24"
             install-type: develop
             test-data: none
 
           - name: Linux Minimum Setup (Python 3.11)
```
DougBurke, line 48: We'd previously done python 3.9 and 3.11 here - perhaps we should now do python 3.10 and 3.12 (i.e. move 3.11 to 3.12)? 

Although the GH interface makes it hard for me to check the whole setup, so maybe it's not relevant.

DougBurke: Or is the idea that we don't want a 3.12 build at the moment, as it's experimental, and we can add this at a later date?

Marie-Terrell: So, I opted for 3.10 with numpy 1.24 to catch the minimum Python and numpy version "supported". If we bumped one of these "minimum support" jobs, I would bump the Python 3.11 one to Python 3.12 and bump the numpy requirement to 1.26.*. Though, since that is already pretty much covered by another job, I figured it would be better to have more 3.11 testing (to minimize overlap).

DougBurke: We can address this in a future PR (i.e. bump versions if need be).

We do at least have one python 3.12 CI run (conda, linux full build) so we know it is being tested.

DougBurke (approved):

PR #2157, opened 2024-09-24
Improve test coverage of ignore_bad

DougBurke:
# Summary

Add more corner-case of quality channel handling for PHA data. There is no functional change here.

# Details

I had been hoping to magically "fix" the quality handling, but there's surprising layers on layers of complexity and design decisions that need to be unravelled. So instead I just want to make sure we add tests that check the current behaviour. Note that these tests may repeat either themselves or other existing tests, but I want to have good test coverage before changing things (some of these are based on code changes I had planned to make). Some of the tests are taken from a previous attempt

- #1639

### DIFF
--- a/sherpa/astro/tests/test_astro_data2.py
+++ b/sherpa/astro/tests/test_astro_data2.py
@@ -1681,6 +1681,60 @@ def test_quality_quality_filter_expr(make_quality_pha):
     pha.get_filter() == "1:9"  # does not exclude bad channels at end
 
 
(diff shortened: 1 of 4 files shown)


hamogu (approved):
There is never anything wrong with adding more regression tests as long as they are clearly labelled as such. On the other hand, there is no clear "go, but we need one more to be complete" - you never are.
In that sense, this is an easy review: The tests make sense, so why not add them?

(Two small typos in comments to tests - Nothing a user ever sees are I really care about.)

PR #2160, opened 2024-09-25
Fixup test failures from #2118

DougBurke:
# Summary

Fix up a test case when writing out the range of a column for a FITS column.

# Details

So, we have two issues here

1. I had only run the changes from #2118 with the AstroPy backend and the crates test needed to be changed (the TLMAX1 value was 10 but is now 4, and this was an explicit change in this PR)

2. the reason why I hadn't noted it was that the AstroPy backend, which has the same test, wasn't changed because of an "issue" with how the header elements are set. The code I wrote did not error out but did not do what I expected for TLMINn and TLMAXn **ONLY**, so they were not updated. Which meant that the corresponding change for the AstroPy backend had not had to be made, so I hadn't looked at the Crates version.

So there's a simple fix to the AstroPy FITS writing to make sure that TLMIN/MAX are written out, and updates to the test to pick up the new values.

### DIFF
--- a/sherpa/astro/io/pyfits_backend.py
+++ b/sherpa/astro/io/pyfits_backend.py
@@ -1191,10 +1191,10 @@ def _create_table_hdu(hdu: TableBlock) -> fits.BinTableHDU:
             out.columns[col.name].unit = col.unit
 
         if col.minval is not None:
-            out.header.append((f"TLMIN{idx}", col.minval))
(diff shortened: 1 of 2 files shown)


Marie-Terrell (approved):
The tests now pass for me in a local CIAO+Sherpa install with the fixes added

Issue #2158, opened 2024-09-25
Python 3.12 macOS Tests


Marie-Terrell:
It looks like we are now seeing this in the Conda macOS Python 3.12 tests: https://github.com/python/cpython/issues/100228

This is causing "errors" in the Conda deployments workflow (at the assert 0 warnings). As there have been many discussions of fork vs spawn for sherpa in the past and would require more investigation before making any sort of default change. With the discussion on the issue above, this is not a "deprecation" but more of a nudge to use spawn for now. So, I assume we just want to accept/filter the warning for now to give more time to review fork vs spawn (post release)?

DougBurke:
See also #2007

DougBurke:
There's also the "was a problem but then stopped being a problem"

- #1950

Marie-Terrell:
> There's also the "was a problem but then stopped being a problem"
> 
> * [multiprocessing and python 3.12 #1950](https://github.com/sherpa/sherpa/issues/1950)

Thank you! I was searching through the issues, but missed this one.

DougBurke:
The #2007 one is perhaps more relevant for the future!

PR #2162, opened 2024-09-25
TEST: ignore fork deprecation warnings

DougBurke:
# Summary

Ignore the deprecation warnings about the use of the fork method for multiprocessing.

# Details

We can not switch to the other multiprocessing methods at present, as discussed in issue #2007, so the easiest thing to do is skip the warnings (e.g. issue #2158).

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -149,7 +149,15 @@ known_warnings = {
             # Technically this should only be needed if bokeh is installed
             # but it doesn't seem worth setting up that machinery.
             #
-            "\nPyarrow will become a required dependency of pandas "
+            "\nPyarrow will become a required dependency of pandas ",
+
+            # See issues
+            # https://github.com/sherpa/sherpa/issues/2158
+            # https://github.com/sherpa/sherpa/issues/2007
+            #
+            # For now we hide them.
+            #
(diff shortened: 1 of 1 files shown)


Marie-Terrell:
I pulled this into another PR to check that the errors in the deployments workflow dissappear, but I am still seeing them.
If you temporarily update to add "pull_request:" to .github/workflows/ci-conda-deployment.yml. Example:
```
on:
  push:
    branches:
    - main
    - ciaox
    - 'release/**'
  pull_request:
```
You should be able to see if this fixes the issues seen in main for macOS Python 3.12 pre-deploy testing. I am still seeing this after adding your MR here, but I might be missing something

DougBurke:
Yeah - because I can't easily trigger these problems it doesn't completely surprise me that I did the wrong thing here...

DougBurke:
This will **definitely** fix it. It can't **possibly** go wrong.

Marie-Terrell:
Awesome! Looks like that did the trick! You can pull the pull request trigger change for this and then this is good to go!

DougBurke:
Removed, but thanks to the way I did it we need to do the full re-build for the tests...

Marie-Terrell (approved):
Thank you!

Issue #1990, opened 2024-03-20
thoughts on model cache work


DougBurke:
Models have `startup`/`teardown` methods which the fit and error-estimation machinery use to allow any work that would benefit the models to be done once, rather than many times. It doesn't currently do much other than supposedly trigger the cache evaluation logic for certain models (e.g. those that have no free parameters). It does mean that calls like `plot_fit` do not benefit from this work,

1. The `startup`/`teardown` approach could (should ?) be added to the MCMC code and things like the flux-error evaluation code, which are both times that behave like this (not sure if we have any other parts that could benefit).
2. The startup/teardown approach is not well tested (and neither is some of the cache code).
3. I think that it's also both too granular and too general at the same time
  - for instance, there may be times when we'd want to do this at either the composite-model case (e.g. when allying a response to a model), or indeed at the multiple-dataset case where we know the same grid is being used for the same set of models so we can just evaluate the model once and share (which could only hold for part of the full model expression, such as different RMF/ARF applied to the same source expression).

There is cache code for 1D models but not 2D models. The cache code takes the parameter values, the integrate setting, the xlo and (if present the xhi) arguments, and any keyword arguments and calculates a numeric hash (this is not cryptographic analysis). 

4. Model writers also have to know to add a `sherpa.models.model.modelCached1d` decorator to the `calc` method otherwise it doesn't come into play. It also is only meant to be used with `ArithmeticModel` sub-classes which might be surprising (can it be made a "compile-time" check / is it worth it?)
5. As mentioned, the cache code is a bit tricky to know if it really is working or not.
6. The cache calculation code has not been compared to model evaluation to determine when it's faster. This is technically going ti be a per-model-with-a-particular-set-of-parameters decision.
7. For XSPEC additive models and many Sherpa models with an `ampl`-like parameter, we could cache the un-normalized value and just do a simple scalar multiplication. I thought I had an issue about this for the XSPEC case but maybe I never created it.
8. The XSPEC models are hard to properly cache since we can not guarantee that the evaluation grid, parameter values, `integrate` setting [which should be fixed for XSPEC models but that's a different set  of tickets], and keyword arguments are **not enough** to uniquely identify the model since you technically should also
    - check the abundance and cross-section table settings
    - worry about any XSET values and/or other hidden-ish database values
  So maybe we need an XSPEC-specific version which gets cleared if any of these extra settings gets changed (safer, if less efficient, caching)?

You can see an XSPEC specialization idea at 
- #1201.

[...]

Issue #767, opened 2020-03-26
should modelCacher1d special case models with a normalization?


DougBurke:
Many models - e.g. all XSModelAdditive components - have a normalization or amplitude which is a linear scaling. That is, `model(x) = norm * model'(x)`. We could take advantage of this in `modelCacher1d` so that we cache the model before normalization (the `model'` term), and apply the normalization when we get the value from the cache. I'm wondering if this will be beneficial for models with a small number of parameters when doing some form of interval analysis (e.g. interval or region projection) 

Some issues

  1. for simple models the extra cost may exceed any benefit to having a cache
  2. how do we identify the models where we can do this

I have an idea for #2 - based on #349,  to add a sub-class of Parameter just for amplitude/normalization parameters - but not sure if this is just me trying to push this ticket rather than it being a sensible idea ;-)

hamogu:
This is significant in terms of complexity - your new sub-class for a model parameter will probably turn out to be more complicated than it initially looks. So the question for me is: Do we have a use case where profiling indicates that this actually saves significant time?

I've never looked at how parameters change. It is common in fitting that the optimizer changes only the normalization? I don't know, but that is relatively easy to find out. How often do people even do interval analysis? Personally, I typically run the fitters, but rarely do interval or regions projection. Of course, that depends on the science we do, but this sounds like good to have in principle, but most likely not worth the effort.

Just my 2c

hamogu:
Here is an investigation **for on particular XSPEC model family** (the code below is for apec, but I've tried vapec as well and I suspect it's the same for the entire family or APEC models). It turns out, that the XSPEC model itself includes significant caching and optimization where the norm seems to be special-cased as suggested in this issue. The experiment below is with Sherpa 4.17:
```Python
>>> import bumpy as np
>>> from sherpa.astro import ui
>>> from sherpa.astro.xspec import XSapec
>>> import time
>>> from sherpa.astro import xspec
>>> xspec.get_xsversion()
'12.14.0k'
>>> en_bins = np.arange(0.2, 10., 0.01)
>>> for i in range(5):
...     apecmodel = XSapec()
...     start_time = time.perf_counter()
...     flux = apecmodel(xlo=en_bins[:-1], xhi=en_bins[1:])
...     end_time = time.perf_counter()
...    print(f"Time taken: {end_time-start_time:.2f} s")
Reading APEC data from 3.0.9

Time taken: 0.33 s
Time taken: 0.02 s
Time taken: 0.02 s
[...]

PR #2166, opened 2024-09-27
Fix bug in cache code that limited cache size to 1.

hamogu:
## Summary
Fix bug that limited cache size to 1.

## Details
While working on #2080, I discovered that the caching code does not do what the docs promise it would. The size of the cache is fixed to 1, although the docs and default setting put it at 5. Given how little memory caches need (for a typical X-ray spectrum with 1024 channels and a float64 data type returned from the model it's about 2kB) but how long many XSPEC models take to evaluate there is no justification to restrict it to size one nor was that intended (from reading docs and code).

I looked at the logs for our fitters at typical X-ray fits, and setting it to 1 actually captures most of the calls, but going slightly higher will be better in some cases at little cost.

So, this PR fixes that bug.

### DIFF
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -56,9 +56,9 @@ to look up a value in the `_cache` dictionary. If it exists then the
 stored value is returned, otherwise the model is evaluated and added
 to the `_cache` dictionary. In order to keep the cache size small, the
 `_queue` array is used to remove an existing value from the store when
-a new value is added. The default size for the `_queue` array is
-a single value, but it can be changed by
-:ref:`the startup method <startup-modelcacher1d>`.
+a new value is added. The size for the `_queue` array is set to
+:py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
+default value for this attribute is 5).
 
 .. _startup-modelcacher1d:
 
@@ -68,9 +68,8 @@ The startup method
 The model :py:meth:`~sherpa.models.model.ArithmeticModel.startup`
 method is automatically called by the :py:meth:`~sherpa.fit.Fit.fit`
 method, but can also be called manually. It sets the `_use_caching`
-attribute and sets the `_queue` array to have
-:py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
-default value for this attribute is 5).
+attribute.
+
 
 Although the default value for the `cache` argument to `startup` is
 set to `False`, the `sherpa.fit.evaluates_model` decorator - which
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -44,6 +44,17 @@ This may be useful if you are evaluating models over a large grid,
 to save memory. For a composite model (e.g. a sum of models) you need
 to set the cache for each component.
 
+When we use the fit method of the UI or the fit method of an optimizer, it is possible to
+turn off caching for all models with the ``cache`` parameter, e.g.
+``ui.fit(..., cache=False)``.
+This works by iterating over all the model components and setting the
+``cache`` attribute to zero. Note that the opposite is not true: setting
+``cache=True`` does not turn on the cache for all model components, it simply
+leaves it at the previous setting. This is because some models may not work with
+caching at all and need to stay at ``cache=0`` at all times.
+The cache has to be manually set to a positive number for all models that should use the cache
+to allow caching again.
+
 How does the cache work?
 ========================
 
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -50,13 +50,12 @@ How does the cache work?
 ========================
 
 The parameter values, integrate setting, and grid values are used to
-create an unique token - the SHA256 hash of the values - which is used
+create a unique token - the SHA256 hash of the values - which is used
 to look up a value in the `_cache` dictionary. If it exists then the
 stored value is returned, otherwise the model is evaluated and added
 to the `_cache` dictionary. In order to keep the cache size small, the
-`_queue` array is used to remove an existing value from the store when
-a new value is added. The size for the `_queue` array is set to
-:py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
+oldest element in the cache is removed when the number of entries becomes
+larger than :py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
 default value for this attribute is 5).
 
 .. _startup-modelcacher1d:
@@ -110,8 +109,7 @@ True
 [0. 1. 1. 1. 0. 0.]
 >>> print(m._cache)  # doctest: +SKIP
 {b'<random byte string>': array([0., 1., 1., 1., 0., 0.])}
->>> print(m._queue)  # doctest: +SKIP
-[b'<random byte string>']
+
 
 Fit and the startup method
 --------------------------
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -29,23 +29,22 @@ When is the cache useful?
 =========================
 
 At present most 1D models use the cache by default when evaluated
-normally, but not during a fit. It is intended to improve fit
-performance - that is, reduce the time taken to fit a dataset - but
-there has been limited effort to evaluate its efficiency.
+during a fit, but not when evaluated separately (e.g. on the command
+line by hand). It is intended to improve fit performance, but the actual
+time saved depends on the model and the data being fit.
 
 Can I turn off this behavior?
 =============================
 
-The `_use_caching` attribute of a model can be set to `False` to stop
-the cache behavior. This may be useful if you are evaluating models
-over a large grid, to save memory, or the model calculation is not
-expensive, and so the extra time used to store the result is not
-beneficial.
+The size of the cache for a specific model component called ``mdl`` can
+be set to zero (``mdl.cache=0``) to turn off the cache behavior.
+This may be useful if you are evaluating models over a large grid,
+to save memory. For a composite model (e.g. a sum of models) you need
+to set the cache for each component.
 
-Note that the :ref:`the startup method <startup-modelcacher1d>` can
-change this value, but it depends if you are calling `startup`
-directly or indirectly, via the ``fit`` and ``est_errors`` methods of
-a fit object.
+Alternatively, caching can be switched off for a specific
+fit call: ``f.fit(cache=False)``.
+
 
 How does the cache work?
 ========================
@@ -63,7 +62,7 @@ default value for this attribute is 5).
 .. _startup-modelcacher1d:
 
 The startup method
-==================
+------------------
 
 The model :py:meth:`~sherpa.models.model.ArithmeticModel.startup`
 method is automatically called by the :py:meth:`~sherpa.fit.Fit.fit`
@@ -82,7 +81,7 @@ have to explicitly pass ``cache=False`` to the fit method::
     f.fit(cache=False)
 
 The teardown method
-===================
+-------------------
 
 The model :py:meth:`~sherpa.models.model.ArithmeticModel.teardown`
 method is run after the fit is done - to match `startup` - and
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -28,10 +28,12 @@ by running a test as shown below,
 When is the cache useful?
 =========================
 
-At present most 1D models use the cache by default when evaluated
-during a fit, but not when evaluated separately (e.g. on the command
-line by hand). It is intended to improve fit performance, but the actual
+At present most 1D models use the cache by default.
+It is intended to improve fit performance, but the actual
 time saved depends on the model and the data being fit.
+Compared to most built-in sherpa models, models in the optional XSPEC model
+library (:py:mod:`sherpa.astro.xspec`) tend to be more complex and
+thus benefit more from caching.
 
 Can I turn off this behavior?
 =============================
@@ -41,10 +43,6 @@ be set to zero (``mdl.cache=0``) to turn off the cache behavior.
 This may be useful if you are evaluating models over a large grid,
 to save memory. For a composite model (e.g. a sum of models) you need
 to set the cache for each component.
-
-Alternatively, caching can be switched off for a specific
-fit call: ``f.fit(cache=False)``.
-
 
 How does the cache work?
 ========================
@@ -58,33 +56,6 @@ oldest element in the cache is removed when the number of entries becomes
 larger than :py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
 default value for this attribute is 5).
 
-.. _startup-modelcacher1d:
-
-The startup method
-------------------
-
-The model :py:meth:`~sherpa.models.model.ArithmeticModel.startup`
-method is automatically called by the :py:meth:`~sherpa.fit.Fit.fit`
-method, but can also be called manually. It sets the `_use_caching`
-attribute.
-
-
-Although the default value for the `cache` argument to `startup` is
-set to `False`, the `sherpa.fit.evaluates_model` decorator - which
-is used to wrap the :py:meth:`~sherpa.fit.Fit.fit`,
-:py:meth:`~sherpa.fit.Fit.simulfit`, and
-:py:meth:`~sherpa.fit.Fit.est_errors` methods - over-rides this value
-and uses a value of `True`. Therefore, to turn off the cache you
-have to explicitly pass ``cache=False`` to the fit method::
-
-    f.fit(cache=False)
-
-The teardown method
--------------------
-
-The model :py:meth:`~sherpa.models.model.ArithmeticModel.teardown`
-method is run after the fit is done - to match `startup` - and
-currently sets the `_use_caching` setting to `False`.
 
 Examples
 ========
@@ -101,8 +72,6 @@ attribute, and see that it has been updated by the model evaluation.
 >>> m = Box1D()
 >>> m.xlow = 1.5
 >>> m.xhi = 4.5
->>> print(m._use_caching)
-True
 >>> print(m._cache)
 {}
 >>> print(m([1, 2, 3, 4, 5, 6]))
@@ -165,9 +134,3 @@ The cache contains 4 elements which we can display::
     [3.39944171 3.39944171 3.39944171]
     [3.40061543 3.40061543 3.40061543]
 
-Note that if we had called::
-
-    f.fit(cache=False)
-
-then the cache would not have been used (e.g. `mdl._cache` would
-have remained empty).
--- a/sherpa/astro/models/tests/test_astro_model.py
+++ b/sherpa/astro/models/tests/test_astro_model.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2016, 2018, 2020, 2021, 2022, 2023
+#  Copyright (C) 2007, 2016, 2018, 2020-2023, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -95,7 +95,6 @@ def test_send_keyword_1d(cls):
     """What happens if we use an un-supported keyword?"""
 
     mdl = cls()
-    mdl._use_caching = False
 
     if cls == models.LineBroad:
         mdl.vsini = 1e6
@@ -118,7 +117,6 @@ def test_send_keyword_2d(cls):
     """What happens if we use an un-supported keyword?"""
 
     mdl = cls()
-    mdl._use_caching = False
 
     # Not guaranteed to produce interesting results
     x0 =  [10, 12, 13, 15]
--- a/sherpa/astro/models/tests/test_astro_model.py
+++ b/sherpa/astro/models/tests/test_astro_model.py
@@ -91,8 +91,8 @@ def test_create_and_evaluate(name, cls):
                                  models.PseudoVoigt1D,
                                  models.Schechter,
                                  models.Voigt1D])
-def test_send_keyword_1d(cls):
-    """What happens if we use an un-supported keyword?"""
+def test_send_keyword_1d_with_cache(cls):
+    """What happens if we use an un-supported keyword and use caching?"""
 
     mdl = cls()
 
@@ -101,6 +101,35 @@ def test_send_keyword_1d(cls):
 
     # Not guaranteed to produce interesting results
     x = [10, 12, 13, 15]
+    y1 = mdl(x)
+    assert len(mdl._cache) == 1
+    y2 = mdl(x, not_an_argument=True)
+    assert len(mdl._cache) == 2
+    assert y2 == pytest.approx(y1)
+
+
+@pytest.mark.parametrize("cls", [models.Atten,
+                                 models.BBody,
+                                 models.BBodyFreq,
+                                 models.BPL1D,
+                                 models.Beta1D,
+                                 models.Edge,
+                                 models.LineBroad,
+                                 models.Lorentz1D,
+                                 models.NormBeta1D,
+                                 models.PseudoVoigt1D,
+                                 models.Schechter,
+                                 models.Voigt1D])
+def test_send_keyword_1d_no_cache(cls):
+    """What happens if we use an un-supported keyword and disable caching?"""
+
+    mdl = cls()
+    mdl.cache = 0
+    if cls == models.LineBroad:
+        mdl.vsini = 1e6
+
+    # Not guaranteed to produce interesting results
+    x = [10, 11.3, 13, 15]
     y1 = mdl(x)
     y2 = mdl(x, not_an_argument=True)
     assert y2 == pytest.approx(y1)
--- a/sherpa/astro/utils/xspec.py
+++ b/sherpa/astro/utils/xspec.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2012 - 2016, 2020 - 2024
+#  Copyright (C) 2012-2016, 2020-2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -526,7 +526,7 @@ def read_model_definition(fh, namefunc: Callable[[str], str]) -> Optional[ModelD
 
 def mpop(array: list[str]) -> Optional[float]:
     """Pop first element from array (converting to float),
-    returning defval if empty.
+    returning None if empty.
     """
 
     try:
--- a/sherpa/astro/utils/xspec.py
+++ b/sherpa/astro/utils/xspec.py
@@ -876,8 +876,7 @@ def simple_wrap(modelname: str, mdl: ModelDefinition) -> str:
     # initialized.
     #
     if nflags > 1 and mdl.flags[1] == 1:
-        out += f"{t2}self._use_caching = False\n"
-
+        out += f"{t2}self.cache = 0\n"
         # Still warn the user that this is not tested.
         out += f"{t2}warnings.warn('support for models like xs{mdl.name.lower()} "
         out += "(recalculated per spectrum) is untested.')\n"
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -473,9 +473,11 @@ def modelCacher1d(func: Callable) -> Callable:
         vals = func(cls, pars, xlo, *args, **kwargs)
 
         # remove first item in queue and remove from cache
+        # if cache is reaching max size.
         queue = cls._queue
-        key = queue.pop(0)
-        cache.pop(key, None)
+        if len(queue) >= cls.cache:
+            key = queue.pop(0)
+            cache.pop(key, None)
 
         # append newest model values to queue
         queue.append(digest)
@@ -1400,8 +1402,7 @@ class ArithmeticModel(Model):
 
     def cache_clear(self) -> None:
         """Clear the cache."""
-        # It is not obvious what to set the queue length to
-        self._queue = ['']
+        self._queue = []
 
         self._cache: dict[bytes, np.ndarray] = {}
         self._cache_ctr = {'hits': 0, 'misses': 0, 'check': 0}
@@ -1498,20 +1499,12 @@ class ArithmeticModel(Model):
         return NotImplemented
 
     def __setstate__(self, state):
-        self.__dict__.update(state)
-
+        # In case old pickle files do not have cache info
+        self.cache_clear()
         if '_use_caching' not in state:
             self.__dict__['_use_caching'] = True
 
-        if '_queue' not in state:
-            self.__dict__['_queue'] = ['']
-
-        if '_cache' not in state:
-            self.__dict__['_cache'] = {}
-            self.__dict__['_cache_ctr'] = {'hits': 0, 'misses': 0, 'check': 0}
-
-        if 'cache' not in state:
-            self.__dict__['cache'] = 5
+        self.__dict__.update(state)
 
     def __getitem__(self, filter):
         return FilterModel(self, filter)
@@ -1522,7 +1515,7 @@ class ArithmeticModel(Model):
         if int(self.cache) <= 0:
             return
 
-        self._queue = [''] * int(self.cache)
+        self._queue = []
         frozen = np.array([par.frozen for par in self.pars], dtype=bool)
         if len(frozen) > 0 and frozen.all():
             self._use_caching = cache
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -287,10 +287,6 @@ which all have the same grid.
 
 The `cache` attribute of the model sets the maximum number of
 entries in a cache. Setting it to 0 disables caching for a model.
-If the cache is actually used is controlled by the `_use_caching`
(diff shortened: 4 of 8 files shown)


hamogu (commented):
Obviously too late for 4.17, but since I tried to push #2080 in and failed, I want to get the work that I did for that PR (and realted findings like this one) into PRs sooner rather than later before I forget what I did. I understand that it will be a week or two before people are ready to review this and related PRs and get stuff merged again.

hamogu:
@DougBurke I didn't 't want to push it into 4.17, but that was several months ago. Maybe you can find time to have a look?

hamogu:
CI Failure is unrelated to this PR, please don't let that stop you from reviewing.

docs/evaluation/cache.rst
```diff
@@ -29,48 +29,45 @@ When is the cache useful?
 =========================
 
 At present most 1D models use the cache by default when evaluated
-normally, but not during a fit. It is intended to improve fit
-performance - that is, reduce the time taken to fit a dataset - but
-there has been limited effort to evaluate its efficiency.
+during a fit, but not when evaluated separately (e.g. on the command
```
DougBurke, line 32: I don't agree with this change. I believe it's quite easy to show that the cache is used outside of a fit. Checking whether it is working at all, or well, during a fit, is trickier.

I am also not sure why I wrote, in the original "but not during a fit". It may have been true, or I may have been confused, or ...

docs/evaluation/cache.rst
```diff
@@ -29,48 +29,45 @@ When is the cache useful?
...
-The `_use_caching` attribute of a model can be set to `False` to stop
-the cache behavior. This may be useful if you are evaluating models
-over a large grid, to save memory, or the model calculation is not
-expensive, and so the extra time used to store the result is not
-beneficial.
+The size of the cache for a specific model component called ``mdl`` can
+be set to zero (``mdl.cache=0``) to turn off the cache behavior.
+This may be useful if you are evaluating models over a large grid,
+to save memory. For a composite model (e.g. a sum of models) you need
+to set the cache for each component.
```
DougBurke, line 43: A meta note, really, but technically we could cache composite models. The questions include

-  whether the cache checks would be faster / slower than the model combination handled by the composite operation
- is it likely both components are going to be cached, or just one (or neither)

This isn't really relevant here, but I wanted to mentino it.

docs/evaluation/cache.rst
```diff
@@ -29,48 +29,45 @@ When is the cache useful?
...
-create an unique token - the SHA256 hash of the values - which is used
+create a unique token - the SHA256 hash of the values - which is used
 to look up a value in the `_cache` dictionary. If it exists then the
 stored value is returned, otherwise the model is evaluated and added
 to the `_cache` dictionary. In order to keep the cache size small, the
-`_queue` array is used to remove an existing value from the store when
-a new value is added. The default size for the `_queue` array is
-a single value, but it can be changed by
-:ref:`the startup method <startup-modelcacher1d>`.
+oldest elements in the cache is removed when the number of entries becomes
```
DougBurke, line 57: either

```suggestion
oldest element in the cache is removed when the number of entries becomes
```

or

```suggestion
oldest elements in the cache are removed when the number of entries becomes
```

sherpa/models/model.py
```diff
@@ -285,10 +285,12 @@
 most effective when the same model is used with multiple datasets
 which all have the same grid.
 
-The `_use_caching` attribute of the model is used to determine whether
-the cache is used, but this setting can be over-ridden by the startup
-method, which is automatically called by the fit and est_errors
-methods of a `sherpa.fit.Fit` object.
+The `cache` attribute of the model sets the maximum number of
+entries in a cache. Setting it to 0 disables caching for a model.
+If the cache is actually used is controlled by the `_use_caching`
```
DougBurke, line 290: I don't understand this sentence.

sherpa/models/model.py
```diff
@@ -473,12 +475,10 @@ def cache_model(cls, pars, xlo, *args, **kwargs):
         vals = func(cls, pars, xlo, *args, **kwargs)
 
         # remove first item in queue and remove from cache
-        queue = cls._queue
-        key = queue.pop(0)
-        cache.pop(key, None)
+        # if cache is reaching max size.
```
DougBurke, line 478:
```suggestion
        # if the cache is full.
```

We **could** change from a `list` to a [deque](https://docs.python.org/3/library/collections.html#deque-objects) object. I have seen some complaints about the  runtime speed of `deque` but have not investigated enough to know if they are meaningful or not. We are also not really doing much with the queue so is the semantic improvement worth it?

sherpa/models/model.py
```diff
@@ -1374,28 +1374,43 @@ def rfunc(self, lhs):
...
 
     def cache_clear(self) -> None:
         """Clear the cache."""
-        # It is not obvious what to set the queue length to
-        self._queue = ['']
-
         self._cache: dict[bytes, np.ndarray] = {}
         self._cache_ctr = {'hits': 0, 'misses': 0, 'check': 0}
 
+        # One might think that this could be a class attribute, but
```
DougBurke, line 1405: This didn't really help me. First of all, what is `AstroModel.__init__` referring to?

There is then the discussion (aluded to above) that we *could* change the behaviour of `CompositeModel` - although I believe that if we were to look at this it would be in a follow-up PR and not in this one.

I think this is one of the cases where we have to be careful about changing explicit calls to `SomeClass.__init__()` with `super().__init__()` as that can lead to fun-to-find issues like this. This is more an aside to myself than anything related to this PR.

Am I right in thinking it's something like

```
# Only clear the _use_caching field if the object has it. This is to ensure that
# CompositeModel-derived classes do not gain this field.
```

sherpa/models/model.py
```diff
@@ -1430,20 +1445,12 @@ def cache_status(self) -> None:
     __pow__, __rpow__ = _make_binop(np.power, '**')
 
     def __setstate__(self, state):
-        self.__dict__.update(state)
-
+        # In case old pickle files do not have cache info
```
DougBurke, line 1448: Given the discussion in `cache_clear` above, I would think we should only call `cache_clear` after checking the `_use_caching` field. Technically it should not matter as written, but I think logically it would make sense.

sherpa/models/tests/test_model.py
```diff
@@ -954,6 +973,7 @@ def test_evaluate_cache1dint():
     xlo, xhi = xgrid[:-1], xgrid[1:]
 
     mdl = Polynom1D()
+    mdl.cache = 1
```
DougBurke, line 976: Can we remove the `_use_caching` line here now?

sherpa/models/tests/test_model.py
```diff
@@ -1079,6 +1108,109 @@ def calc(self, p, *args, **kwargs):
         return numpy.ones(args[0].size)
 
 
+def get_cache_classes():
+    """This is a function because we want to conditionally
+    include an XSPEC model. Within a function, we can simply
+    pass that if XSEPC is not available.
```
DougBurke, line 1114:
```suggestion
    pass that if XSPEC is not available.
```

The number of times I make this tpyo!

DougBurke (commented):
So, I've never had the time to really dive into what on earth the cache code is doing, other than to note the oddities. So thanks for someone else doing it.

We *could* make `_use_caching` a property that

- when read it just returns whether `self.cache > 0` (or whatever the variable is)
- when set to `False` it sets `self.cache = 0`
- the tricky thing is what to do if set to `True` - it could try to pick the default `cache` size either for the class or for `ArithmeticModel`?

The idea would be then we only have one source of truth for if the cache is in use, and we could deprecate the `_use_caching` field. However, I can see the benefit in being able to say "I just want to temporarily stop the cache for this model, but do not want to have to remember what the cache size was" which the current `_use_caching` setting allows for. So ....

Have you seen an improvement in run-times when using the cache with more-than-one element? Or even with the old only-one-element in the cache?

docs/evaluation/cache.rst
hamogu: I think I run into a case where it was not used, but I now I can't reproduce it, so either I confused myself, or I had messed around with the same model object in too many ways and broken the setup. I'll fix that text, but either way, the important point is that it **is** used during a fit - that's why we implemented it in the first place.

docs/evaluation/cache.rst
hamogu: I thought about it, but it's more complicated. There are other improvements in how XSEPC models (really the only ones shipped with Sherpa worth caching at the moment) are cached that are certainly much higher value. 

It's also less likely to improve performance. Imagine a simple XSphabs * XSpowerlaw model. When we fit (and fitting with where the caching is most important, because the model will be called many times unlike, e.g. in plotting where you only call it once anyway), the fitter will typically (e.g. LevMar) change one parameter at the time. That means that either the XSphabs can pull the cached value or the XSPowerlaw can call the cached value, and the other model get re-calculated. However, the composite model will always have at least one parameter changes and thus will *never* pull from the cache.

You can verify that simply by writing out the parameter values that the fit tried (I looked at those files for a few different models). 

Now, for a more complicated model like `(phabs1 * vapec1) + (phabs2 * (vapec2+vapec3)) + ...some bkg model here...`, caching the composite model would help because there are a number of fitting steps where neither phabs1 nor pha2 change their parameters. On the other hand, pulling from one cache (dict) instead of pulling from two and multiplying the numbers does not change much time because all the operations (having parameters, dict lookup, multiplying two small arrays) are very fast anyway.

hamogu:
Currently, two things have to be true to use the cache:
- `cache > 0`
- `_use_caching = True`

That's the current design - we could change it and (as you said below) drop the `_use_caching` altogether. 

The reason I left it like that is that that allows you to use the cache and keep the cached values, but then do something in between that sets `_use_caching=False`. When you return to `_use_caching=True` all the values in the cache are still there and can be used again without being re-calculated. 

Now that I write it down like this though, that sounds like a very strange edge case that we don't have to support in the main code. If you really need it (e.g. for debugging) you could also just make a copy of the `_cache` dictionary by hand. I think I should be more radical and take out the `_use_caching` altogether (with a shim that preserves backwards compatibility, i.e. setting it to `False`, just sets the tie of the cache to 0 and setting it to `True` sets the size to some hardcoded default value).

sherpa/models/model.py
hamogu: Yes, that's the intention. Will reword.

sherpa/models/model.py
hamogu: In the PR, I use a dict. Python spent a lot of effort recently to improve performance of dict AND they also maintain the insertion order (since 3.9 or something like that). With that, there really is not much difference between a dict and a `deque` any more in the way we use it here, so I propose to stick with the more-known `dict` instead of the `deque` (performance aside).

docs/evaluation/cache.rst
hamogu: I did reproduce it. The cache is used outside a fit **but only if the model has never been used in a fit before**. See for example, the following with Sherpa 4.17

```Python

In [2]: ui.load_arrays('data1', [1,2,3,4], [1,2,2,1.1])

In [3]: ui.set_model('data1', ui.box1d(name='boxmodel'))

In [4]: mdl = ui.get_model('data1')

In [5]: mdl._use_caching
Out[5]: True

In [6]: ui.fit(cache=False)

[...]     

In [7]: mdl._use_caching
Out[7]: False
```

This happens **even if the fit is done with `cache=True`**:

``Python

In [2]:  ui.load_arrays('data1', [1,2,3,4], [1,2,2,1.1])

In [3]: ui.set_model('data1', ui.box1d(name='boxmodel'))

In [4]: mdl = ui.get_model('data1')

In [5]: mdl._use_caching
Out[5]: True

In [6]: ui.fit(cache=True)

[...]  

In [7]: mdl._use_caching
Out[7]: False
```

So, what I wrote is (or was) correct **after the model had been used for any fit** (unless you manually do in an reset the private attribute `_use_caching`), not in general.

Anyway, with the total removal of `_use_caching` in the next commit, this deserved rewording anyway.

**Comment by hamogu** [comment]
I pushed a commit that simplifies the caching even further as suggested in the review by removing the private `_use_cache` attributed altogether.

Before that commit, caching happened if two things were true at the same time: The size of the cache is > 0 and `_use_caching == True`. Now there is only one source of truth: The setting of `model.cache`.

Since `_use_caching` was clearly marked as private, we can remove it without worrying about deprecation.

On the face of it, it looks like this will introduce difficulties with `ui.fit(cache=False)` which "runs a fit without caching". That might make a user think that a following run with `ui.fit(cache=True)` would then again use caching. That's not the case after this commit: `ui.fit(cache=False)` will set `cache=0` and will not reset it to a different value after the fit. We could do that in the `teardown` method, but what do we reset is to? To `cache=5`? What if a model should have the cache permanently disabled for some very good reason? So, we would have to keep a temporary copy of the old `cache` value before the fit and then reset to that. That's complicated. And in fact, the old startup and teardown method also didn't do that.

Teardown used to have this line:
`self._use_caching = False`, so after a fit, `use_caching` was always False.

The following is with the Sherpa 4.17, showing that this behavior is pre-existing:
```Python

In [2]: ui.load_arrays('data1', [1,2,3,4], [1,2,2,1.1])

In [3]: ui.set_model('data1', ui.box1d(name='boxmodel'))

In [4]: mdl = ui.get_model('data1')

In [5]: mdl._use_caching
Out[5]: True

In [6]: ui.fit(cache=False)

[...]

In [7]: mdl._use_caching
Out[7]: False
```

This was true even if the fit was done with `ui.fit(cache=True)`.

So, there is a change in behavior in this commit, but it's fixing this bug, so it's for the better.

You will still disable the cache for all models if `ui.fit(cache=False)` and do not reset them after, but at least `ui.fit(cache=True)` no longer alters the size of the cache.

hamogu:
@DougBurke This changes how some XSPEC model Python wrappers are generated - those that do not allow cashing at all. I'm not sure if we have any of those at the moment? Do we? If so, do we have to regenerate all the XSPEC model wrappers in Python as part of this PR or better to do that separately?

hamogu:
I run some benchmarking and the results are inconclusive. That doesn't mean that this fix is not worthwhile, just that, as a I said above, it depends a lot on the data and the optimizer.

I am using the benchmarks in https://github.com/hamogu/sherpa-opt-benchmark/ because I have those ready to run. Unfortunately, I wrote those to compare different optimizers and fitting options within one branch of code. In order to check numerical performance, I use `fake_pha` with low good values. Unfortunately, that means that every time I step out, recompile a different branch of sherpa, and run again, I get different spectra to fit and thus a different number of function evaluations. Even with caching, it's no surprise that it takes longer if you need more steps to find the minimum!

Here are the raw numbers:
Screenshot 2025-02-11 at 2 27 59 PM

where the last column is wit this PR and the time shown is the runtime per function evaluation. I know from profiling that especially the absorption model that I use (phabs) are expensive and that I need to avoid calling those as much as possible.  However, it doesn't come out here.

Now, I *could* improve the test cases I use to read the spectrum from a file such that every fit uses the same input, and I can look for models that profit more from caching than some of those I used here (e.g. a power law is so fast to evaluate that caching can't do much anyway). 
Also, none of the tests in the table fits more than one dataset simultaneously - as case I know to benefit from caching.

I *could* do all that, but I don't think it's worth our time to benchmark this more: I know that the cache is broken in several aspects on main (see all the comments above), I know this PR improves that, and I know that caching is helpful only in limited circumstances right now because many models are either so fast that it doesn't matter or are usually called with different parameters anyway - however, there are cases where it helps a lot (e.g. fitting several spectra simultaneously), those are just not represented in the test cases I have set up since they were set up for other purposes.

DougBurke:
> @DougBurke This changes how some XSPEC model Python wrappers are generated - those that do not allow cashing at all. I'm not sure if we have any of those at the moment? Do we? If so, do we have to regenerate all the XSPEC model wrappers in Python as part of this PR or better to do that separately?

Sorry - I don't see why this is true. Why do you need to change the XSPEC model wrappers? Exactly what do you mean by "model wrappers" - is it the Python code or the C++ code>

hamogu:
I'll rebase later to address conflicts, because if I rebase and force-push now then the link I just posted in https://github.com/sherpa/sherpa/pull/2166#issuecomment-2657817131 would break.

hamogu:
Address the most stupid merge conflicts I've ever seen (white space around the numbers in a copyright line)

DougBurke:
For the comment above: https://github.com/sherpa/sherpa/pull/2166#issuecomment-2648600232

> One reason in particular for a cache size > 1 is when you fit more than one dataset. If you have a cache size of 1 and two datasets with different bins (almost certain, if you bin on counts even if both are ACIS observations), than a cache of size 1 is useless because in every step of the fit XSphabs will first be evaluated for dataset 1 and then for dataset 2 - with the same parameters but different x values and thus the cache won't be hit. Then, the cache value is updated and even if the parameters for XSphabs are the same in the next fit step, if will be re-calculated, but the cached value in a size=1 cache now holds the x-values of dataset 2.
Maybe we should automatically increase the size of the cache to be at least the number of datasets fit. But that's a follow-up PR.

This is technically not true, in the best way that "technically" holds. The grouping used is irrelevant for the cache code, since the cache is only used when the model is evaluated on the "base" energy grid (e.g. the `ENERG_LO` and `ENERG_HI` columns of the `MATRIX` block of the RMF). The grouping is applied **after** the model is evaluated, and this is after the cache code has done its magic. Or has not.

So, if you have several Chandra datasets with the same instrument/grating order then you should find that they use the same energy grid (by default), so **in this case** you should not suffer too much from the cache-size of 1. However, if you are trying to fit Chandra + XMM say then you would hit this issue. In this case an alternative is to create separate absorption models and then link the `nH` parameters. Of course, this is **terrible** UX, since the user has to understand what different grids are in use and then create the models knowing this ....

So I'm not saying this work is not needed, just ...

duty_calls

sherpa/models/model.py
```diff
@@ -1376,25 +1371,31 @@ def rfunc(self, lhs):
 class ArithmeticModel(Model):
     """Support combining model expressions and caching results."""
 
-    cache = 5
-    """The maximum size of the cache."""
+    @property
+    def cache(self) -> int:
+        """The number of entires in the cache.
```
DougBurke, line 1376:
```suggestion
        """The maximum number of entries in the cache.
```

sherpa/models/model.py
```diff
@@ -1376,25 +1371,31 @@ def rfunc(self, lhs):
...
+
+        Changing this value will clear the cache.
```
DougBurke, line 1378: We *could* try to be clever and leave the existing cache alone of we change `.cache` to a larger value, but I do not think it worth the extra complexity (in code, but more importantly  documenting it / making it harder to understand what the rules are)/

sherpa/astro/utils/xspec.py
```diff
@@ -526,7 +526,7 @@ def read_model_definition(fh, namefunc: Callable[[str], str]) -> Optional[ModelD
 
 def mpop(array: list[str]) -> Optional[float]:
     """Pop first element from array (converting to float),
-    returning defval if empty.
+    returning None if empty.
```
DougBurke, line 529: I'm not sure what I was thinking when I wrote this...

docs/evaluation/cache.rst
```diff
@@ -58,33 +56,6 @@ oldest element in the cache is removed when the number of entries becomes
 larger than :py:attr:`~sherpa.models.model.ArithmeticModel.cache` elements (the
 default value for this attribute is 5).
 
-.. _startup-modelcacher1d:
```
DougBurke, line 63: Is there anywhere else in our documentation that we discuss `startup` and `teardown` (other than them being listed as class methods). If not then we probably should, but where. And is it relevant for this PR.

sherpa/astro/utils/xspec.py
```diff
@@ -876,8 +876,7 @@ def simple_wrap(modelname: str, mdl: ModelDefinition) -> str:
     # initialized.
     #
     if nflags > 1 and mdl.flags[1] == 1:
-        out += f"{t2}self._use_caching = False\n"
-
+        out += f"{t2}self.cache = 0\n"
```
DougBurke, line 879: I'm surprised there's no test that needs to be changed because of this. Maybe I just never included a test of this combination.

sherpa/models/model.py
```diff
@@ -1432,37 +1433,20 @@ def cache_status(self) -> None:
...
         return FilterModel(self, filter)
 
     def startup(self, cache: bool = False) -> None:
         self.cache_clear()
-        self._use_caching = cache
-        if int(self.cache) <= 0:
-            return
-
-        self._queue = [''] * int(self.cache)
-        frozen = np.array([par.frozen for par in self.pars], dtype=bool)
```
DougBurke, line 1460: There has always been something about "whether all pars are frozen" or not that has never been well described. There is an argument to say that if all pars are frozen then the model should be cached, but is it worth making a special case of this? Also, did this code really change anything useful?

DougBurke:
I am tempted to say we should make `_use_caching` a `@property` which

- in set it creates a deprecation warning
- if set to `False` then sets `cache` to 0
- if set to `True` then it sets `cache` to 1 (or 5)
- in get returns the boolean `self.cache > 0`

since some code may be using it, and there's no reason to break it (since we previously have not given good advice on what to do).

I think it might be harder than I'd hope because `_use_caching` appears to be something we just add "apparently randomly", so is there a clear class where we can put such a property (I note there was a comment about this with regard to composite models that got removed in this PR but I can't find it anymore).

hamogu:
> I am tempted to say we should make _use_caching a @property which [...]

This echos an earlier comment by @DougBurke in https://github.com/sherpa/sherpa/pull/2166#pullrequestreview-2606149723 . However, I want to push back here:

`_use_caching` is clearly marked as private (by the Python underscore convention), it's not been advertised as a way to switch on and off caching for practical use in real fitting, none of our users has ever talked to me or opened issues about caching, indicating that people don't generally look into or code and assure our private attributes (the comment in https://github.com/sherpa/sherpa/issues/442#issuecomment-1825598704 supports that as it suggested that we add caching, when in fact we had caching for years, supporting the idea that users don't look into how the sausage is made). So, I don't think this is an issue with backwards-compatibility. With this PR, `_use_caching` is certainly not used in Sherpa any more and that didn't impact a lot of our tests - another indication that it's been more an ad-hoc add-on than a well-integrated design decision.

https://github.com/sherpa/sherpa/pull/2166#pullrequestreview-2606149723 already mentions that we want only one source of truth, and we have that by having only one attribute and that is is the public and documented `cache`. 

If we want a more optimized cache for certain models in the vein of #767 and #2080 (which I absolutely want to, because there is significant speed-up we can get that users will actually feel), then supporting a badly defined `_use_caching` on top of `cache` will be even harder (and make `_use_caching` even less well defined).

sherpa/models/model.py
hamogu: In the final version of this PR, there is no `_use_caching` any more. This is more of a compatibility layer for old pickle files. In fact, it might be better to never pickle (or unpickle) the cache at all, because the risk that some hidden state (such as the ATOMDB version or the elemental abundances) have inadvertently changed, not because the user changed a model, but because e.g. the Sherpa version was upgraded. That's bad enough on it's own, but then to have cache values that are restored invites crazy inconsistencies that throw a fit off.

sherpa/models/model.py
hamogu: This is a consequence of removing `_use_caching`. `frozen` was only used in the line just below this one to initialize `_use_caching`, without `_use_caching` we don't need this line any longer. If the cache is activated (i.e. `cache > 0`) then a model called with all parameters frozen will obviously be cached, because it will be called with the same (frozen) parameters every time. There is no need to special treat that situation. `cache > 0` is the default for our models. If someone actively sets the cache to 0 (off) then I'll assume they have a reason (e.g. they are experimenting with a model with frozen parameters, but changes in the XSPEC abundance or cross-section settings). I think those cases are rare, and I don't want to magically and secretly switch caching back on for all-frozen models, if the user actively switched it off before.

(The situation might have been slightly different because of https://github.com/sherpa/sherpa/pull/2166/#discussion_r1949827595 but that's solved with this PR.)

docs/evaluation/cache.rst
hamogu: I suggest to address that in the context of #2244. We don't do a lot with those methods right now, though there are potentially some uses in the future - see #2244. Since we don't use them, it's hard to document what their intended use is (and if that use intended when they were added is actually what we would want to use them for today).

sherpa/models/model.py
DougBurke: I can helpfully come up with arguments for both sides, so I'm not sure I'm much help here ...

DougBurke:
Just to push back on your push back in https://github.com/sherpa/sherpa/pull/2166#issuecomment-2770373091 @hamogu, while I do think that since we have limited time, we do not need to handle all cases, and "messing around with a private variable" does generally fit into the "well, you should  not have been doing that" case, there are times when a Python code base (now talking generally) ends up forcing you to use "private" features. And then, when they are changed, cause loads of problems. Yes, I am now thinking about setuptools ...

I think I have talked about the `_use_caching` attribute before (whether on an issue or in a notebook, but I've just checked https://github.com/DougBurke/sherpa-standalone-notebooks and fortunately can't find it there), and if so then anyone who has followed my advice could reasonably get upset with such a change. I don't think the cost of adding a `_use_caching` deprecation message / workaround is huge, or nor would maintaining it until we remember to take it out, but I do agree that there is a cost. And I haven't actually looked to see if it is as simple as I think.

Here's a few places where people could have found out about `_use_caching`, but I think it's really only the Sherpa team that were involved:

https://github.com/sherpa/sherpa/issues?q=is%3Aissue%20_use_caching

sherpa/models/model.py
hamogu: OK then, I have arguments for clearing the cache, so that's go with that.

hamogu:
I still don't think it's needed (I looked that mentions of `_use_caching` and most them are not relevant, e.g. the mention in an issue that says that we don't use it or the 2D case), but I don't want to argue any longer; it only a dozen lines to at the `_use_caching` property so I did as requested (and added some tests).

docs/evaluation/cache.rst
DougBurke: So do we think it's safe to resolve this comment now?

hamogu: I think so.

sherpa/models/tests/test_model.py
```diff
@@ -1199,8 +1242,8 @@ def test_cache_not_used_in_fit():
     res = fit.fit(cache=False)
     assert mdl.c0.val == pytest.approx(2)
     assert len(mdl._cache) == 0
```
DougBurke, line 1244: Would it make sense for this to be

```suggestion
        assert mdl.cache == 0
```

(and similarly line 1238 checks `mdl.cache` rather than `len(mdl.cache)`)? The idea being that we try not to use "internal" stuff where possible so we can ensure the user API is useful / makes sense. Now, saying that, our tests are full of access to internals (sometimes it is necessary, sometimes it is code that could be updated but we haven't got a need for changing it, and sometimes it's in between). Here I've got myself confused in what the code is now doing, so I'm asking!

DougBurke: [well, after fixing the extra indentation I added...]

hamogu: Here, we want to check that `fit.fit(cache=False)` does not use the cache. The best way to check that it's not use it to check that nothing has been inserted into the dictionary, which is what this line does.

`assert mdl.cache == 0` should actually fail, because the size of cache is > 0 (the default is 5), but even if `cache > 0`, the dict that holds the cached values can be empty because the cache just has not been filled yet.

DougBurke: Ta. I don't think we need to add an API to check the "used" size of the cache as that just feels like TMI ;-)

sherpa/astro/models/tests/test_astro_model.py
```diff
@@ -95,7 +95,6 @@ def test_send_keyword_1d(cls):
     """What happens if we use an un-supported keyword?"""
 
     mdl = cls()
-    mdl._use_caching = False
```
DougBurke, line 98: So, below we call

```python
    y1 = mdl(x)
    y2 = mdl(x, not_an_argument=True)
```

I forget now what happens in the code to "throw-away" this extra argument, but it's not obvious to me whether here we should still ensure there's no cache to make sure we get to the "call the model" rather than the cache just going "oh, we don't care about this argument" - e.g. instead of removing this line we should replace it with

```python
mdl.cache = 0
```

Similar for `test_send_keyword_2d` below.

Maybe we should add a parametrized argument like

```python
@pytest.mark.parametrize("cache_size", [0, 5])
```

and set

```python
mdl.cache = cache_size
```

just to make sure we check both cases without having to worry about the internals of what the cache does.

hamogu: We can make the test even more stringent. If the model is called with the extra parameter, there needs to be a second entry in the cache because the cache hash holds a different hash for the second call than the first. So, what's in there right now and what you suggest are really two different tests. Will split.

hamogu:
I suggest that I squash the last few commits and summarize them all as "response to code review" when this is approved, but for now I have them separate to make it easier for @DougBurke to check I I have addressed his concerns.

DougBurke (approved):
LGTM.

I think there's still work to be done here but this is a useful improvement as is. And this time I mean it ...

hamogu:
> I think there's still work to be done here but this is a useful improvement as is.

Absolutely. And some of that work is going on, see e.g. #2256 and the other open issues we have for cache work like #767 or #1990 or your multiprocessing work.

DougBurke:
@hamogu - looks like there's new boring conflicts to resolve here.

DougBurke:
Those look odd errors; something strange with merge/rebase/whatever?

hamogu:
I tried the gh web interface for the conflict resolution and that screwed up. I'll do it locally later.On Apr 9, 2025 12:18 PM, Doug Burke ***@***.***> wrote:
  Those look odd errors; something strange with merge/rebase/whatever?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



DougBurke left a comment (sherpa/sherpa#2166)
Those look odd errors; something strange with merge/rebase/whatever?

sherpa/astro/models/tests/test_astro_model.py
```diff
@@ -102,6 +102,35 @@ def test_send_keyword_1d(cls):
...
+    """What happens if we use an un-supported keyword and disable caching?"""
+
+    mdl = cls()
+    mdl.cache = 0
+    if cls == models.LineBroad:
+        mdl.vsini = 1e6
+
+    # Not guaranteed to produce interesting results
+    x = [10, 11.3, 13, 15]
+    y1 = mdl(x)
```
DougBurke, line 133: I'm too lazy to check: do we have a test that after `mdl.cache = 0` a call to `mdl(x)` we still have `len(mdl._cache) == 0`?

hamogu: yes: https://github.com/sherpa/sherpa/pull/2166/files#diff-1426e384ca73b9afdca1ae2b6092d7f22f29e67ceef6035c10119eb0f6310592R945-R958

DougBurke: ta.

docs/evaluation/cache.rst
```diff
@@ -44,6 +44,17 @@ This may be useful if you are evaluating models over a large grid,
...
+When we use the fit method of the UI or the fit method of an optimizer, it is possible to
+turn off caching for all models with the ``cache`` parameter, e.g.
+``ui.fit(..., cache=False)``.
+This works by iterating over all the model components and setting the
+``cache`` attribute to zero. Note that the opposite is not true: setting
+``cache=True`` does not turn on the cache for all model components, it simply
+leaves it at the previous setting. This is because some models may not work with
+caching at all and need to stay at ``cache=0`` at all times.
+The cache has to be manually set to a positive number for all models that should use the cache
+to allow caching again.
```
DougBurke, line 56: For reference, this is something we may want to improve in the future, but only after we have some experience with this new, improved, cache and decide if changing the cache status is something people will do. It's not something for this PR.

DougBurke (approved):
I've just checked the PR again after the rebase and still approve it ;-)

DougBurke:
I am biased, as I asked for it, but one of my PRs failed sensibly because of the `_use_caching` behavior

```
sherpa/conftest.py:288: AssertionError
---------------------------------------------------------------------------------- Captured stdout teardown ----------------------------------------------------------------------------------
*** Warnings created: 1
1/1 {message : DeprecationWarning('_use_caching is deprecated.Instead, set `cache` to 0 or a positive integer.'), category : 'DeprecationWarning', filename : '/lagado.real/sherpa/sherpa-main/sherpa/models/model.py', lineno : 1391, line : None}
```

which is what I wanted (rather than it just failing because `_use_caching` had been removed).

PR #2170, opened 2024-10-01
DOC: Example changes to Examples

DougBurke:
# Summary

Change the docstring for calc_source and calc_model to use Examples rather than Example.


hamogu:
While I'm not religiously opposed to this change, I do note that the [numpydoc documentation on docstrings](https://numpydoc.readthedocs.io/en/latest/format.html#examples) uses the plural independent of the actual number of examples in the examples section. I don't know off-hand if that is a requirement or merely and example of how to doc things, but for consistency (and easy of parsing docstrings automatically with sphinx and other tools) there is something to be said for using the exact same set of heading in a doctoring.

DougBurke:
@hamogu - I'm confused. Isn't my change doing exactly what you say that numpy suggest we do? And that's indeed the reason I'm doing it - for consistency (and this consistency is useful for the code I use to generate the sherpa ahelp files).

hamogu (approved):
I guess I must be becoming red-green blind these days...

PR #2171, opened 2024-10-01
specify DS9 region format used in pytest

nplee:
DS9 pytests fails since `xpaset regions` failed to execute if you have a DS9 preference file to default the region file formatting as something other than 'ds9' (e.g. 'ciao').  

By specifying the test's region format [and coordinate system, which is already being done in these tests] being used will supersede whatever setting is found in the user's preference file by doing a `xpaset('regions format ds9')` before the `xpaset('regions', data=imshape)`.

### DIFF
--- a/sherpa/image/tests/test_image.py
+++ b/sherpa/image/tests/test_image.py
@@ -216,6 +216,7 @@ def test_image_getregion(coordsys):
 
     # Use XPA to set a region in the imager
     imshape = 'image; circle 8.5 7.0 0.8'
+    im.xpaset('regions format ds9')
     im.xpaset('regions', data=imshape)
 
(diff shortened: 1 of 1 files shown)


DougBurke (approved):
LGTM (assuming tests still pass ...)

anetasie:
@nplee @DougBurke are we planning to get this to the release?

DougBurke:
No

DougBurke:
For reference, my thinking here is that

1. this issue only affects users who run the tests **and** have both DS9 and the XPA tools in their environment
2. can be addressed by temporarily moving their DS9 preference file

so I don't think it is a critical issue.

PR #2174, opened 2024-10-02
Allow tests to pass with Python 3.13.0

DougBurke:
# Summary

Allow the tests to pass with Python 3.13.0.

# Details

Python 3.13 has changed how docstrings are indented (in that it automatically removes leading indentation).

This affects a few of our tests, so update them to allow for the old and new behaviour. I felt it simpler to just gate on the Python version (sys.version_info) rather than try to write tests that could handle both. The downside to my approach is that the tests repeat themselves. The upside is that it's much easier and cleaner to update them once Python 3.13 becomes our minimum supported version.

There is also a change in a `TypeError` which affects a few of our XSPEC tests: these tests are written to catch the current behaviour, since the code is known to fail, and so we just want to know when anything has changed. So we can easily tweak the message check to work with both pre-3.13 and 3.13 messages.

I have run this test with as extensive a check as I can make it (Linux only) checking

- matplotlib plus bokeh
- DS9 and xpa
- `runslow` tests
- `runzenodo` tests

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -1405,10 +1405,23 @@ def test_show_fit(session):
     assert toks[11] == "Statistic: LeastSq"
     assert toks[12] == "Least Squared Statistic."
     assert toks[13] == ""
-    assert toks[14] == "    The least-square statistic is equivalent to a chi-square"
-    assert toks[15] == "    statistic where the error on each point - sigma(i) - is 1."
+
+    # Python 3.13 has changed the way that leading spaces are handled in docstrings
+    msg1 = "The least-square statistic is equivalent to a chi-square"
+    msg2 = "statistic where the error on each point - sigma(i) - is 1."
+    if sys.version_info >= (3, 13, 0):
+        assert toks[14] == msg1
+        assert toks[15] == msg2
+    else:
+        assert toks[14] == f"    {msg1}"
+        assert toks[15] == f"    {msg2}"
+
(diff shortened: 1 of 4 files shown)


hamogu:
Hi,
I wonder if we should take #2174. Python 3.13 is going to be released well before Sherpa 4.17 makes it out to most users as part of CIAO 4.17. By the time CIAO 4.18 comes around, Python 3.13 is going to be > 1 year old already. 
#2174 just makes a few tests pass and Doug has chosen to gate that by Python version number, so there is no impact from the PR at all below 3.13 making the risk of taking it so late in the cycle very low.

I don’t think it’s essential to do that (CIAO 4.17 with sherpa 4.17 doesn’t even support python 3.12, let alone 3.13), but it would be nice to future-proof - in particular if Chandra funding develops in such a way that the CIAO 4.18 release with Sherpa 4.18 doesn’t happen in the same form we’ve had over the last few years (later / not at all / only if the build is done by the community / CIAO without updates to any OTS / who knows what happens).

Moritz

DougBurke:
One issue is that I claim it improves compatability with 3.13.0, but we don't currently don't have tests to show this (and are not going to, for a while I think....)!

DougBurke:
Rebased to pick up the fix from #2177

DougBurke:
Rebased to pick up recent changes. I have also squashed the two commits into 1 as I didn't think it really helped.

DougBurke:
Rebased to pick up #2208

hamogu (approved):

PR #2177, opened 2024-10-08
Read the Docs build failures with newer sphinx versions and sphinx_rtd_theme deprecation warning

Marie-Terrell:
Summary
=======
Specify sphinx < 8 for the read the doc environment and remove the call to get_html_theme_path no longer needed.

Details
======
With the recent sphinx_rtd_theme release (3.0.0) released yesterday, we now pick up sphinx > 8.0 and encounter a new deprecation warning for get_html_theme_path noting that it is no longer needed. 

Double slash issue with newer sphinx versions
-----------------------------------------------
The sphinx pin here: https://github.com/readthedocs/sphinx_rtd_theme/blob/master/setup.cfg was updated from
```
sphinx >=5,<8
```
to:
```
sphinx >=6,<9
```
With the previous pinning, this forced the use of sphinx <8 in the RTD builds. With the new pinning, we are encountering an issue similar to this: https://github.com/sphinx-doc/sphinx/issues/12782
Which is not released yet (currently targeted for sphinx 8.1.0). This warning is triggering sphinx failures.

get_html_theme_path deprecation warning
--------------------------------------------
With sphinx_rtd_theme 3.0.0 we are also now seeing failures due to the following warning
```
WARNING: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.
```
So, the html_theme_path setting in the conf.py has also been removed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -372,7 +372,6 @@ edit_on_github_doc_root = 'docs'
 html_theme = 'sphinx_rtd_theme'
 
 # html_theme_options = {}
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
 # The name for this set of Sphinx documents.
 # "<project> v<release> documentation" by default.
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -18,5 +18,5 @@ dependencies:
     - ipykernel
     - nbsphinx
     - sphinx-astropy
-    - sphinx_rtd_theme>=1.0.0
+    - sphinx_rtd_theme>=3.0.0
     - bokeh
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - astropy
   - pip
   - graphviz
+  - sphinx<8.0
   - pip:
     - ipykernel
     - nbsphinx


hamogu (approved):
It passes - that's what counts here.

DougBurke:
I do wonder if there's the possibility that if you have `sphinx_rtd_theme < 3.0.0` installed that the behavior would change (in that the theme would not be picked up or it could error out somewhere). So we **could** change `docs/environment.yml` from

```
    - sphinx_rtd_theme>=1.0.0
```

to

```

Marie-Terrell:
> It passes - that's what counts here.

Could you spot check any areas in the RTD built on this PR for any differences that might arise from the changes in theme line change and from: https://about.readthedocs.com/blog/2024/07/addons-by-default/ 

Things looked good in the few spots I checked, but I want to make sure I didn't miss anything obvious

Marie-Terrell:
> I do wonder if there's the possibility that if you have `sphinx_rtd_theme < 3.0.0` installed that the behavior would change (in that the theme would not be picked up or it could error out somewhere). So we **could** change `docs/environment.yml` from

That's a good point. I'll bump that line too as that is what is being picked up, but it would be good to specify this just in case.

DougBurke (approved):
As noted, we may need follow-up changes for the general case, but what's important now is that the RTD docs build.

I've done spot checks on https://sherpa--2177.org.readthedocs.build/en/2177/index.html and it looks okay to me.

hamogu:
I also did spot checks few pages (one with text, one with generated API docs etc.) in the built and they looked fine.

DougBurke:
Hold on. Has anyone tried the search functionality? It doesn't see, to be working for me in https://sherpa--2177.org.readthedocs.build/en/2177/index.html

hamogu:
Doesn't work for me either - I wonder though if that's just because we're looking at docs rendered from a branch, not a full deployment. The only way to find out is to merge this and, if it doesn't work, revert or modify in a follow-up PR.

DougBurke:
If we go to an existing PR then I can get the search to work - e.g. https://sherpa--2022.org.readthedocs.build/en/2022/search.html?q=datapha&check_keywords=yes&area=default

So I think this is due to https://about.readthedocs.com/blog/2024/07/addons-by-default/ - but it could still be a "this only works on the main branch" change, but I would be **very** surprised

Marie-Terrell:
I would have thought so too, but I can search on:
https://sphinx-rtd-theme.readthedocs.io/en/3.0.0/
I cannot search on a RTD build of RTD on a PR:
https://sphinx-rtd-theme--1616.org.readthedocs.build/en/1616/

So, this might be a bug or an indexing thing of some sort on PRs.

DougBurke:
yes, not being able to get a match for the word "sphinx" on https://sphinx-rtd-theme--1616.org.readthedocs.build/en/1616/ is a bit telling!

So I think we should be good to go with this then. Fingers crossed.

Marie-Terrell:
Yeah. You can see some of difference in sherpa_rtd_themes searching here:
https://github.com/readthedocs/sphinx_rtd_theme/pull/1610

Here the search on the left works (no flyout created) and the search in the bottom right read the docs box (creates a flyout when search is clicked) does not work:
https://sphinx-rtd-theme--1609.org.readthedocs.build/en/1609/
Here you now see the flyout is created in both cases and the search fails:
https://sphinx-rtd-theme--1610.org.readthedocs.build/en/1610/

PR #2180, opened 2024-10-10
DOCS: add a what's new / changelog to the documentation

DougBurke:
# Summary

Add a "What's New" area to the documentation.

# Details

It is not yet clear how best to populate this information. That is, how to update a "change log" and then have an overview of these changes.

AstroPy uses towncrier https://pypi.org/project/towncrier/ but it is likely worth using a more-manual system first to work out the kinks.


hamogu:
This would close #555

PR #2190, opened 2024-10-29
Units not analysis

DougBurke:


DougBurke:
@hamogu - this is marked as a draft since Ithink we need to decide whether the change in behaviour we now have is okay. On the one hand I think it makes our code a bit cleaner (but I am biased, and I haven't looked at this for a while). On the other, I do recognize that for our users a change in behavior here could lead to annoyance (since, for the user, it is a regression, and it's not as if we can say "hey look at all the improvements this has lead to").

It really depends on how people load background PHA datasets with no responses, and the steps they take.

hamogu:
Mmm. That seems like a pretty significant change for users with little additional benefit. It makes logical sense, but one might also say "Why does Sherpa are in which order in which those files, I'll make sure all files are loaded when I need them". In that sense, the check should not happen when the units are set, but when they are used ("Unit set to wave, but not response loaded, can't fit/plot/whatever").

I understand that that's not this PR and it's not easily done, I'm just trying to think what users might expect.

DougBurke:
Yeah - so I think this needs some TLC.

PR #2191, opened 2024-10-30
TEST: fix test related to issue #800

DougBurke:
# Summary

Clean up a test. There is no functional change to the code.

# Details

It looks like there was a typo when loading in the data files and so this was just testing that a == a. Change so that we load in a counts and a rate file (see issue #800). There are also some other clean-ups I made as I was trying to understand the test. These changes are essentially style changes.

### DIFF
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2016 - 2018, 2020, 2021, 2023, 2024
+#  Copyright (C) 2016 - 2018, 2020 - 2021, 2023 - 2024
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
@hamogu - when looking at #1564 I noticed that it looked like this test was accidentally using the "rate" file for both the "counts" and "rate" data. So I've fixed this, and also cleaned up the test (just because I was trying to work out what was going on here on the work I'm actually trying to do).

Does this change look sensible?

DougBurke:
@hamogu - are you able to look at this one. it should be easy (but may take a bit of time to remind yourself what the test was doing)?

hamogu:
I will look - but probably not today.

hamogu (approved):
Correction, I can do that today. And I'm sorry it slipped through the cracks for so long.

DougBurke:
No hurry - just wanted to make sure it didn't get lost.

PR #2192, opened 2024-10-30
No longer use bin_lo/hi values from PHA or ARF. Fix #1564

### LINKED ISSUE #1564: Remove `bin_lo` and `bin_hi` from `DataPHA`?

DougBurke:
# Summary

Do not use the BIN_LO/HI columns from Chandra grating PHA/ARF files. Fix #1564.

# Details

As noted on #1564 and #1871 the BIN_LO/HI settings just complicate the code and do not add much (CXC never pushed this as a standard and I do not think any other mission uses these columns). The aim is to

- simplify the code
- make it easier to use a public interface to get at the independent axis values (see #2189, which relies on this PR)

This relies on (a tick indicates that it has been merged **and** removed from this PR)

- [X] #2188
- [X] #2191
- [ ]  #2190

The changes are relatively easy, but a bit annoying, in part because there are a number of tests that set the `bin_lo/hi` fields as they provided an easy way set up things (i.e. without creating a RMF) and so some effort is needed to work out if the test is still relevant.

The `DataPHA` and `DataARF` classes retain `bin_lo` and `bin_hi` fields but they are not used (and, in the `DataPHA` case, are always set to `None`, no-matter what the code tries).  This si to try and reduce the impact this change will have (but it's probably not going to save people who affected by the change). There is a `FutureWarning` the first time that the `DataPHA.bin_lo` field is set to a non-zero value, but I don't know how useful that will be (there is no warning for the `DataARF` case as I don't see that as being that useful to warn about).

The PR finishes off with some internal clean ups in the instrument code

- `pylint` suggestions
- adding some types (where I understand the code)
- some amalgamation and simplification
- some code style changes

More changes here probably require properly tackling #1871 as there is some interesting code here left over from some of the `bin_lo/hi` removals.

# Notes

Before this change you could say

```python
ui.load_pha("3c120_pha2")
ui.set_analysis("wave")
ui.plot_data(4)
```

and see something like

```
dataset 1: 0.333333:7.16 Wavelength (Angstrom)
dataset 10: 1:41.96 Wavelength (Angstrom)
dataset 11: 0.5:20.98 Wavelength (Angstrom)
dataset 12: 0.333333:13.9867 Wavelength (Angstrom)
dataset 2: 0.5:10.74 Wavelength (Angstrom)
dataset 3: 1:21.48 Wavelength (Angstrom)
dataset 4: 1:21.48 Wavelength (Angstrom)
dataset 5: 0.5:10.74 Wavelength (Angstrom)
dataset 6: 0.333333:7.16 Wavelength (Angstrom)
dataset 7: 0.333333:13.9867 Wavelength (Angstrom)
dataset 8: 0.5:20.98 Wavelength (Angstrom)
dataset 9: 1:41.96 Wavelength (Angstrom)
```

sherpa

Now the call will error out with the message

```
DataErr: No instrument response found for dataset 3c120_pha2
```

There's no real easy way to get this to say "Create a diagonal response" to let people know what to do (and, to be honest, they are better off just loading in the actual response at this point anyway, my $0.02).

### DIFF
--- a/docs/data/pha.rst
+++ b/docs/data/pha.rst
@@ -66,8 +66,6 @@ PHA FITS file::
   counts         = Float64[1024]
   staterror      = None
   syserror       = None
-  bin_lo         = None
-  bin_hi         = None
   grouping       = Int16[1024]
   quality        = Int16[1024]
   exposure       = 38564.608926889
@@ -96,8 +94,6 @@ that are set in this file's FITS metadata::
   counts         = Float64[1024]
   staterror      = None
   syserror       = None
-  bin_lo         = None
-  bin_hi         = None
   grouping       = Int16[1024]
   quality        = Int16[1024]
   exposure       = 38564.608926889
@@ -115,8 +111,6 @@ that are set in this file's FITS metadata::
   energ_lo = Float64[1090]
   energ_hi = Float64[1090]
   specresp = Float64[1090]
-  bin_lo   = None
-  bin_hi   = None
   exposure = 38564.141454905
   ethresh  = 1e-10
   >>> print(pha.get_rmf())
@@ -147,8 +141,6 @@ created::
   counts         = Int64[1024]
   staterror      = None
   syserror       = None
-  bin_lo         = None
-  bin_hi         = None
   grouping       = None
   quality        = None
   exposure       = None
@@ -499,8 +491,6 @@ representing the background region.
   counts         = Float64[1024]
   staterror      = None
   syserror       = None
-  bin_lo         = None
-  bin_hi         = None
   grouping       = Int16[1024]
   quality        = Int16[1024]
   exposure       = 38564.608926889
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -1061,13 +1061,6 @@ extra checks, such as the `~sherpa.astro.data.DataPHA.grouping` and
 `~sherpa.astro.data.DataPHA.quality` columns for PHA data which
 are converted to integer values.
 
-One example of incomplete validation is that the
-`~sherpa.astro.data.DataPHA.bin_lo` and
-`~sherpa.astro.data.DataPHA.bin_hi` fields are not checked to ensure
-that both are set, or that they are in descending order, that the
-``bin_hi`` value is always larger than the correspondnig ``bin_lo``
-value, or that there are no overlapping bins.
-
 .. _data_design_errors:
 
 Error messages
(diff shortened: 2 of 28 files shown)


DougBurke:
Tests may appear to be failing because the `macos-12` runners are temporarily being removed. See #2009

anetasie:
@DougBurke I'm wondering if this would break some scripts from our users, specifically Fabrizio Nicastro's ones.

DougBurke:
@anetasie - there is a possibility it would, but it's hard for me to imagine that you can do much in Sherpa without loading in an actual response.

hamogu:
I'm a little reluctant to do a full review of a Pr with > 1000 LOC changed that's still marked as a draft. However, looking over the actual changes in the code (not the tests and docs) I want to leave a comment here that I support going in this direction and this this PR seems to be pretty close to being done, so hope we get it in in the next few weeks.
 
 The bin_lo/bin_hi are essentially a poor-mans version of a delta RMF. Given that we now have tools to build a delta-RMF pretty easily in Sherpa, I agree that the bin_hi/bin_lo support should be removed and the cleaner alternative be documented for people who have a spectrum but not RMF (as can sometimes happen when designing instruments).

DougBurke:
@hamogu - thanks for looking. After reviewing this and trying to remember what I was doing, I think I ran out of steam because - as written - this relies on 

- #2190 

and although that seems like it should be an easy change, it actually points out some hidden state we were using (details more in that PR, brought about because a test failed and I couldn't understand at first why) which meant I wasn't sure it would make sense to pursue at that time. Maybe now is the time to pursue it.

DougBurke:
Replaced by #2358

PR #2195, opened 2024-11-05
Update ci-conda-workflow to use Miniforge

Marie-Terrell:
Summary
-------------
Install Miniforge for use with ci-conda worflow.

Details
-------------
Conda has been updated on the Linux GitHub Actions image (Ubuntu 22.04.5 runner image version: 20241103.1.0). This pulls in recent Conda updates to remove the inclusion of the "defaults" channel by default (opting to let the installers like Miniconda, etc. do this themselves). Related:
- https://github.com/conda/conda/releases/tag/24.9.2
- https://github.com/conda/conda/releases/tag/24.9.0
As the default Conda installed on the GitHub Actions images is "Miniconda" anyway [(Ubuntu 22.04 software details)](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md), we want to move away from this anyway.
There is more discussion to be had about whether it would be better to just use the "setup-miniconda" github action (which allows you to select Miniforge), but considering this has a small footprint and would allow us to more finely control this I am opting for installing Miniforge ourselves (like we already do in the ci-conda-deployment workflow).

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -24,13 +24,7 @@ fi
 
 # Ensure we have set up conda
 #
-source ${CONDA}/etc/profile.d/conda.sh
-
-# update and add channels
-conda update --yes conda
-conda config --add channels conda-forge
-#Remove defaults to avoid conflicts with conda-forge
-conda config --remove channels defaults
+source ${conda_loc}/etc/profile.d/conda.sh
 
 # To avoid issues with non-XSPEC builds (e.g.
 # https://github.com/sherpa/sherpa/pull/794#issuecomment-616570995 )
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -22,10 +22,6 @@ else
    fi
 fi
 
-# Ensure we have set up conda
-#
-source ${conda_loc}/etc/profile.d/conda.sh
-
 # To avoid issues with non-XSPEC builds (e.g.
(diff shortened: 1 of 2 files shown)


Marie-Terrell:
This is listed as failing one check because I keep starting and canceling one check to try to get this to pick up the newer Linux image (with the issue that this is trying to address). This passed the first run (all using the older Linux image without the Conda update).

Marie-Terrell:
On this latest run, I see the following jobs picked up the new image:

- [Linux Full Build (Python 3.10)](https://github.com/sherpa/sherpa/actions/runs/11689510873/job/32552408868?pr=2196#logs)
- [Linux Build (w/o Astropy or Xspec)](https://github.com/sherpa/sherpa/actions/runs/11689510873/job/32552409271?pr=2196#logs)
- [Linux Build (w/o Matplotlib, Xspec, or test data)](https://github.com/sherpa/sherpa/actions/runs/11689510873/job/32552409840?pr=2196#logs)
All three passed, so this looks good and does seem to fix the issue.

Marie-Terrell:
To acknowlege, the pip-arch workflow uses the system python (via apt-get) and ci-pip-workflow uses the "[setup-python](https://github.com/actions/setup-python/tree/main)" action which pulls python from "[python-versions"](https://github.com/actions/python-versions) which pulls from Python.org (and pypy). So, we should no longer have any usage of "defaults" or "Miniconda" with the this PR.

DougBurke:
How does this impact existing PRs which have been waiting for review, such as #2009 ?

PR #2156, opened 2024-09-24
Add macOS ARM Conda Package Builds

Marie-Terrell:
Summary
========
Add macOS ARM Conda builds

Details
=====
This PR only adds macOS ARM Conda package builds to the CI. This does not switch the rest of the macOS testing workflows to ARM as that was completed in #2198.

PR #2198, opened 2024-11-12
Allow both Intel and ARM macOS GitHub runners for the conda builds

DougBurke:
# Summary

Add a macOS ARM build, and update the macOS Intel build to use the macOS-13 runner.

# Details

This is a rework of #1958 and #2009 (#2009 also used the `setup-miniconda` action but #2195 installed it directly so this part was removed).

There's three parts

a) update the scripts to support ARM builds
b) a minor clean up of the workflow actions by splitting up the DS9 and XSPEC installation steps as they are technically separate (this is not technically needed but there are macOS specific rules for the DS9 part which is nice to clean up as part of this)
c) add a macOS ARM build and update the Intel build to `macos-13` since `macos-12` is about to be removed.

As part of this we have

- select the CIAO `conda` channel for macOS ARM since there's no ARM build of `xspec-modelsonly` in the `xspec` channel
- use version `12.13.1` rather than `12.14.0` of XSPEC for macOS ARM to match the CIAO `conda` channel
- tweaked the macOS conda install to use Python 3.11 and install bokeh (for both architectures)

I have **not** checked to see how this interacts with 

- #2156

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -37,12 +37,7 @@ fi
 # the XSPEC-related channels are only added if needed
 #
 if [ -n "${XSPECVER}" ]; then
-  if [[ "`uname -s`" == "Darwin" && "`uname -m`" != "x86_64" ]] ; then
-    # we need to use the CXC conda channel to get the ARM XSPEC build for now
-    conda config --add channels https://cxc.harvard.edu/conda/ciao
-  else
-    conda config --add channels ${xspec_channel}
-  fi
+  conda config --add channels ${xspec_channel}
 fi
 
 # Figure out requested dependencies
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -1,7 +1,17 @@
 #!/usr/bin/env bash -e
 
+# Occasionally useful to know what these values are
+echo "** uname -s: `uname -s`"
+echo "** uname -m: `uname -m`"
+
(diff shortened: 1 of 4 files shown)


Marie-Terrell:
The changes here look good to me (just the one channel workaround can be removed). We will need a fix for the conda package deployment workflow, but that can be in a separate PR (in progress in #2156).

DougBurke:
@Marie-Terrell thanks for looking. I'm not clear exactly what workaround you mean (that we can remove)?

Marie-Terrell:
> @Marie-Terrell thanks for looking. I'm not clear exactly what workaround you mean (that we can remove)?

Sorry. I added a comment directly on the bit. I've uploaded xspec+cfitsio for macOS ARM to the same place as linux and macOS Intel, so you can remove this bit:
```
  if [[ "`uname -s`" == "Darwin" && "`uname -m`" != "x86_64" ]] ; then
    # we need to use the CXC conda channel to get the ARM XSPEC build for now
    conda config --add channels https://cxc.harvard.edu/conda/ciao
  else
```

DougBurke:
Ta @Marie-Terrell - at some point we should update the XSPEC version from 12.13.1 but that can be done at some point in the future...

DougBurke:
Note that this is likely to have a conflict with #2203 (since that PR uses a slightly-modified version of one of the commits here).

Marie-Terrell (approved):
Looks good to me! Thank you!

PR #2201, opened 2024-12-04
Change how invalid limits are checked for within the MCMC code

DougBurke:
# Summary

Simplify a safety check within the MCMC proposed-parameter code.

# Details

This has been part of the specfit est_method_arg update used for the Chandra Source Catalog build - e.g. https://github.com/sherpa/sherpa/pull/2186 - but it is a seperate change from the est_method_arg. There was no documentation to say why this was useful.

It is not obvious why this was done, but there may be cases where the tolerance check for limits lead to problems - e.g.

    np.any(proposed < minvals) or np.any(proposed > maxvals)

is a better check than allowing through a proposed value which is minvals[idx] - delta or maxvals[idx] + delta where

    delta < np.finfo(float).eps

(and given that these values are double, this could happen).

### DIFF
--- a/sherpa/sim/__init__.py
+++ b/sherpa/sim/__init__.py
@@ -218,8 +218,6 @@ from sherpa.utils import random
 
 info = logging.getLogger("sherpa").info
 
-_tol = np.finfo(float).eps
-
 string_types = (str, )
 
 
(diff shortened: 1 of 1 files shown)


DougBurke:
Rebased to pick up test fixes.

anetasie (approved):

DougBurke:
@jmiller515 - this is part of #2186. It's not clear whether this is a needed change, or it just came along for the ride (but I can imagine that in CSC you'd hit cases where the change in behaviour here might be useful). Do you have any info on it or is it just "we need all these changes" or CSC?

PR #2203, opened 2024-12-05
Allow pytest 8.3.4 to be used

### LINKED ISSUE #2202: tests fail because of pytest changes to how it checks for boolean values in pytest 8.3.4
A bunch of our tests have started to fail, presumable because of a `numpy` or `pytest` change:

```
% python -c 'import pytest; from sherpa.data import Data1D; d = Data1D("x", [10,20,30,40,50], [1,2,3,4,5]); d.ignore(14, 45); expected = [True] + [False] * 3 + [True]; print(expected); print(d.mask); assert d.mask == pytest.approx(expected);'
[True, False, False, False, True]
[ True False False False  True]
AssertionError
```

If you look closely you can see the two arrays are displayed differently (because one is a `numpy` array of bool and the other just a list) but semantically the arrays are the same. However, `pytest.approx` doesn't seem to agree with this.

The test failure actually looks like

```
% pytest sherpa/ui/tests/test_ui_unit.py::test_set_filter_masked"[True]"
====================================================== test session starts ======================================================
platform linux -- Python 3.12.6, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/dburke/sherpa/sherpa-xspec
configfile: pytest.ini
plugins: xdist-3.6.1
collected 1 item                                                                                                                

sherpa/ui/tests/test_ui_unit.py F                                                                                         [100%]

=========================================================== FAILURES ============================================================
_________________________________________________ test_set_filter_masked[True] __________________________________________________

ignore = True, clean_ui = None

    @pytest.mark.parametrize("ignore", [False, True])
    def test_set_filter_masked(ignore, clean_ui):
        """What happens when we call set_filter to a filtered dataset?"""
    
        x = np.asarray([10, 20, 30, 40, 50])
        y = np.asarray([2, 3, 4, 5, 6])
        ui.load_arrays(1, x, y)
    
        ui.ignore(lo=15, hi=45)
    
        data = ui.get_data()
>       assert data.mask == pytest.approx([True, False, False, False, True])
E       assert array([ True,...False,  True]) == approx([True,... False, True])
E         
E         (pytest_assertion plugin: representation of details failed: /home/dburke/micromamba/envs/sherpa-xspec/lib/python3.12/site-packages/_pytest/python_api.py:346: TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead..
E          Probably an object has a faulty __repr__.)

sherpa/ui/tests/test_ui_unit.py:1248: AssertionError
----------------------------------------------------- Captured stdout call ------------------------------------------------------
dataset 1: 10:50 -> 10,50 x
------------------------------------------------------- Captured log call -------------------------------------------------------
INFO     sherpa.ui.utils:utils.py:270 dataset 1: 10:50 -> 10,50 x
======================================================= 1 failed in 0.11s =======================================================
```

Now, I have been using

    assert got == pytest.approx(expected)

a lot even for integer-like arrays, where the approx check isn't really needed, because the error message you get if an element of the sequence is different is a lot-more useful than just saying

    assert np.all(got == expected)

but maybe this is just not the way the python testing apparatus is forcing us?

This is with

```
% pip list | egrep 'numpy|pytest'
numpy             2.1.2
pytest            8.3.4
pytest-xdist      3.6.1
% python --version
Python 3.12.6
```

I also tried with `numpy 2.1.3` and got the same result.

Argh - I think it's `pytest 8.3.4` which was released on Dec 1 and includes
-  https://github.com/pytest-dev/pytest/issues/9353 

If I downgrade to `pytest 8.3.3` then the tests pass.

DougBurke:
# Summary

Allow the tests to pass with pytest version 8.3.4 by changing the test code. The GitHub macOS conda action is updated to macos-13 (for the Intel build). There is no functional change here.

# Details

Fix #2202.

In pytest 8.3.3 and earlier
    
        >>> import numpy as np; import pytest
        >>> pytest.__version__
        '8.3.3'
        >>> np.ones(2, dtype=bool) == pytest.approx([True, True])
        True
    
However, pytest 8.3.4 now causes this to fail
    
        >>> import numpy as np; import pytest
        >>> pytest.__version__
        '8.3.4'
        >>> np.ones(2, dtype=bool) == pytest.approx([True, True])
        False
    
This is because of "pytest.approx considers boolean numeric types"

- https://github.com/pytest-dev/pytest/issues/9353
    
The solution is to make the "expected" value be a ndarray, and so
    
        >>> np.ones(2, dtype=bool) == pytest.approx(np.asarray([True, True]))
        True
    
holds with both pytest 8.3.3 and 8.3.4.
    
So this commit basically goes through and updates the tests so that we use a ndarray for boolean arrays.
    
An alternative would be to change from
    
        assert got == pytest.approx(expected)
    
to something like
    
        assert np.all(got == expected)
    
However, the error message when the array lengths are different or an element is different are a **lot less** useful, and the change would be even-more invasive than this change.

I have also added a commit based on the work in #2198 to allow the macOS Intel runner to pass the tests (and bump the python version up to something more up to date).

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -30,13 +30,14 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: MacOS Full Build
-            os: macos-12
-            python-version: "3.10"
+          - name: MacOS Intel Full Build
+            os: macos-13
+            python-version: "3.11"
             install-type: develop
             fits: astropy
             test-data: submodule
             matplotlib-version: 3
+            bokeh-version: 3
             xspec-version: 12.14.0i
 
           - name: Linux Minimum Setup (Python 3.10)
--- a/sherpa/astro/io/tests/test_io_pha.py
+++ b/sherpa/astro/io/tests/test_io_pha.py
@@ -1504,7 +1504,7 @@ def test_write_pha_with_bad_quality(tmp_path):
     counts = chans * 2
     group = [1, -1, -1, 1, -1, 1, 1, 1, -1]
     quality = [0, 5, 0, 0, 0, 0, 0, 2, 2]
-    qfilt = [True, False] + [True] * 5 + [False] * 2
+    qfilt = np.asarray([True, False] + [True] * 5 + [False] * 2)
 
     pha0 = DataPHA("qual", chans, counts, grouping=group,
                    quality=quality)
--- a/sherpa/astro/tests/test_astro_data.py
+++ b/sherpa/astro/tests/test_astro_data.py
@@ -2367,7 +2367,7 @@ def test_pha_check_limit(ignore, lo, hi, evals):
     pha.units = 'energy'
 
     assert pha.mask is True
-    assert pha.get_mask() == pytest.approx([True] * 10)
+    assert pha.get_mask() == pytest.approx(np.ones(10, dtype=bool))
 
     func = pha.ignore if ignore else pha.notice
     func(lo, hi)
@@ -2380,7 +2380,7 @@ def test_pha_check_limit(ignore, lo, hi, evals):
         vin = True
 
     c1, c2, c3 = evals
-    expected = [vout] * c1 + [vin] * c2 + [vout] * c3
+    expected = np.asarray([vout] * c1 + [vin] * c2 + [vout] * c3)
     assert pha.mask == pytest.approx(pha.get_mask())
     assert pha.mask == pytest.approx(expected)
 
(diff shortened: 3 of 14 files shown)


DougBurke:
Argh - the github actions does not follow the recent changes (in part because it was designed with a different workflow) so the simple fix isn't good enough.

DougBurke:
Grrrrrrrrrrrrrrrrrrrrrrrr


rrrrrrrrrrrrrrrrrrrrrrrrrrrr

PR #2211, opened 2025-01-21
Sample sigma

### LINKED ISSUE #2210: ParameterScaleVector ignores sigma parameter
We have

https://github.com/sherpa/sherpa/blob/7f3cea460fe17b79f25586c41faa05934df17265/sherpa/sim/sample.py#L146-L187

So, it looks like we create a `Covariance` object and set it's `sigma` value, but then immediately throw it away and create a new object - e.g.

```python
            covar = Covariance()
            covar.config['sigma'] = self.sigma
            fit.estmethod = Covariance()

            try:
                r = fit.est_errors()
            finally:
                fit.estmethod = oldestmethod
```

This looks wrong.

Found when looking at how to integrate #2186 into the main build.

I will note that #2186 fixes this issue so it is not a problem for CSC, just for us.

I note that `git blame` shows us this code has been present since the initial  git commit.

DougBurke:
This should fix #2210 but it needs some thinking about.

I note that `ParameterSampleVector` and `ParameterSampleMatrix` are contained within other sampling routunes in `sherpa.sim.sample` and so a review of

- can `sigma` be set
- should we be able to change other settings

may be needed.


DougBurke:
@anetasie - I am going to want to talk to you about this when you have some free time

DougBurke:
Rebased to match the current main branch; no changes made.

DougBurke:
As I've had no feed-back on this I think it highly unlikely it will make it in for 4.17.1.

DougBurke:
Rebased to address simple conflict.

DougBurke:
Rebased to pick up recent changes and because I had some time to look at this and provide "better" tests afterunderstanding the code a bit better.

sherpa/sim/sample.py
```diff
@@ -255,29 +279,39 @@ def get_scales(self, fit, myscales=None):
         scales : numpy array
             Two-dimensional scales array (npar by npar elements,
             matching the free parameters in fit). The values are the
-            covariance matrix for the parameters (or the input values if
-            given).
+            covariance matrix for the parameters (or the input values
+            if given).
```
hamogu: Not true any longer. The "values of the covariance matrix" is a mathematically well defined term and does not include scaling by sigma. We have to make a choice here what we actually want: Do we want to return the covariance or do we want to return scales (calculated from the covariance and sigma)? Given the name of the function, I think the latter is appropriate, but then the docstring here should match that.

hamogu (changes requested):

sherpa/sim/sample.py
DougBurke: At some level this gets to my general uncertainty about this: is the `sigma` parameter really meaningful? For those cases where it ends up coming from the covariance matrix (or projection down to 1D) then it feels rather reductive, but there is an argument that it is meaningful when calculated from the data (but it's been long enough for me to forget what exactly that particular code path is doing).

However, we do have at least one user of this functionality (the CSC code), and in other places we have a `sigma` parameter.

This is part of what I want to talk to @anetasie about.

hamogu: If the CSC uses it, let's have it. All I'm asking then is to put some words to that effect into the docstring so that it doesn't just say "the values are the covariance matrix" when that's true only for the (default) `sigma=1`.

DougBurke:
Rebased so it now requires

- #2331 
- #2333 

as those changes (found when working on this PR) are actually separate changes.

sherpa/sim/sample.py
DougBurke: There have been a number of doc changes here, so it would be good to look at this again to see if it is better than before.

There is potential confusion because I can say "covariance matrix from the fit", which has some mathematical meaning. I can also say "send in a covariance matrix which will be used to sample from a distribution", which also has some mathematical meaning. However, for the latter case, I can use any matrix and (as long as it's positive definitive, or whatever the check is) we can get results, even if the matrix isn't formally associated with the fit. So if I, as a user, decide to do something like "get the covariance matrix from the fit, and then scale it by some value, then it's still valid in some sense. [This is also true for the myscales argument for the vector case, as I can just give any-old array of numbers and get a result, as long as it has the right size and is positive, even if the output may be "meaningless"].

What I don't want to add to the documentation is anything resembling that last paragraph, as it's not going to help anyone. I am all for people rewriting what I've written.

DougBurke: [although #2333 needs reviewing first]

DougBurke:
@hamogu and @anetasie - I know you have both had serious concerns about the `sigma` parameter. Do the last changes here help, hinder, or not move the lever?

hamogu:
It is possible to keep accepting the old name, issue a deprecations warning, and have it still work. It's such a ha common case, that astropy has a decorator exactly for this purpose: https://github.com/astropy/astropy/blob/main/astropy/utils/decorators.py#L338

Since that's MIT licensed we could copy and include it (or make astropy a dependency...) BUT I don't think it's worth it in this case; if the routine fails it will be obvious and it's not a very much used part of sherpa or a very much used option to set that parameter.

sherpa/sim/sample.py
```diff
@@ -247,66 +290,85 @@ def get_scales(self, fit, myscales=None):
...
             cov = r.extra_output
+            if cov is None:
+                raise EstErr('nocov')
 
-        else:
+            # Scale the covariance matrix by the square of the
+            # sigma value (which is expected to be 1, although it
+            # can be changed).
+            #
+            cov = self.sigma * self.sigma * cov
```
hamogu: for readability?
```suggestion
            cov = self.sigma ** 2  * cov
```

hamogu (commented):
I like the new text, though I still don't know what the use case is for changing the scale. But if someone has a reasonable use case, I don't see harm in supporting it even if I personally won't need it (and the discussion above suggests that this comes for the CSC).

DougBurke:
I've rebased to address a simple conflict and to apply the suggested change from @hamogu.

DougBurke:
@anetasie - this needs your review.

sherpa/sim/sample.py
```diff
@@ -263,7 +269,7 @@ def get_scales(self,
             raise TypeError("scales option must be iterable of "
                             f"length {npar}")
 
-        return np.asarray(scales).transpose()
+        return np.asarray(scales)
```
anetasie, line 272: Why did you drop transpose() here?

sherpa/sim/sample.py
```diff
@@ -592,8 +610,9 @@ def get_sample(self,
            This defines the thawed parameters that are used to generate
            the samples, along with any possible error analysis.
         mycov : 2D numpy array or None, optional
-           The covariance matrix to use. If None then it is
-           calculated from the fit object.
+           The covariance matrix for the free parameters in the
+           fit. If None then the values are calculated from the fit
+           and scaled by the sigma value of the scale object.
```
anetasie, line 615: When you say "by the sigma value" do you mean 1, 2, 3? or the actual value of sigma error?

sherpa/sim/sample.py
DougBurke: Because the transpose of a 1D NumPy array, which this one is, is the same array:

```
>>> xs = np.asarray([1, 2, 3])
>>> ys = np.asarray([1, 2, 3]).transpose()
>>> xs
array([1, 2, 3])
>>> ys
array([1, 2, 3])
>>> xs == ys
array([ True,  True,  True])
>>> xs.shape
(3,)
>>> ys.shape
(3,)
```

sherpa/sim/sample.py
DougBurke: This object has a `scales` parameter, which contains a `ParameterScaleMatrix` - earlier we have

```python
class ParameterSampleFromScaleMatrix(ParameterSample):
    def __init__(self):
        self.scale = ParameterScaleMatrix()
        ...
```

and this class (`NormalParameterSampleFromScaleMatrix`) derives from `ParameterSamlpeFromScaleMatrix`.

The code for this method has

```python
cov = self.scale.get_scales(fit, myscales=mycov)
```

and `cov` is the covariance matrix we use in the normal calculation. So `cov` is calculated given `mycov` (the input argument) and whatever the `get_scales` method of the `scale` attribute does: this depends on whether `mycov` is set or not. If `mycov` is `None` then the covariance matrix from the fit is used and it is multiplied by the `self.scale.sigma` value (i.e. "the sigma value of the scale object"). We could try to explain this more, but I'm not sure how best to do it (since the text in this reply is both too long and not that helpful).

Issue #2212, opened 2025-01-22
fake_pha channel discrepancy between synthetic spectra and RMF


nplee:
I recently ran into an error running `fake_pha` when including a background PHA file, although I don't think the problem  really has anything to do with the background spectrum itself.  I believe the issue is a discrepancy between the channel value beginning 1 in the synthetic spectrum and 0 in the RMF and background spectrum file.

This example uses the simulation files at:

/data/lagarc/scratch/sherpa-thread_test/nustar/point_sources_for_proposers_v3

or also obtainable from the NuSTAR "Point source simulation files" link at: 
https://www.nustar.caltech.edu/page/response-files 

An example follows, first setting up a simple scenario:

```python
(ciao-4.17-py311) falafel-116: ipython
Python 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from sherpa.astro.ui import *

In [2]: arf = "point_30arcsecRad_1arcminOA.arf"

In [3]: rmf = "nustar.rmf"

In [4]: bgd = "bgd_30arcsec.pha"

In [5]: nustar_arf = unpack_arf(arf)

In [6]: nustar_rmf = unpack_rmf(rmf)

In [7]: nustar_bkg = unpack_pha(bgd)
WARNING: unable to read ARF: File nu10002020001A01_sr.arf does not exist.
WARNING: unable to read RMF: File nu10002020001A01_sr.rmf does not exist.

In [8]: set_source("faked",powlaw1d.p1)
```

An error is then thrown when attempting to include the background spectrum with `fake_pha`

```python
In [9]: fake_pha("faked", arf=nustar_arf, rmf=nustar_rmf, exposure=200000, bkg=nustar_bkg, backscal=nustar_bkg.backscal)
---------------------------------------------------------------------------
DataErr                                   Traceback (most recent call last)
Cell In[9], line 1
----> 1 fake_pha("faked", arf=nustar_arf, rmf=nustar_rmf, exposure=200000, bkg=nustar_bkg, backscal=nustar_bkg.backscal)

File <string>:1, in fake_pha(id, arf, rmf, exposure, backscal, areascal, grouping, grouped, quality, bkg, method)

[...]

PR #2213, opened 2025-01-22
Better handle dataspace1d for DataPHA use cases when offset is not 1

DougBurke:
# Summary

The sherpa.astro.ui.utils.dataspace1d call can now be used to generate a channel grid that does not start at 1 for DataPHA data. The arguments are now checked to ensure they are sensible in this case (consecutive integer values).

# Details

I noticed this when coming up with a work-around for #2212.

We can now say

```python
from sherpa.astro import ui
ui.dataspace1d(0, 1023, dstype=ui.DataPHA)
```

to create a zero-count spectrum with channels 0 to 1023.

Previously I could say

```python
ui.dataspace1d(23, 30, dstype=ui.DataPHA)
```

and that would create a channel grid of `[1, 2, 3, 4, 5, 6, 7, 8]` and this now creates channels of `[23, 24, ..., 30]`. This could cause confusion, but I'm not sure we ever promised the old behaviour.

A more-interesting case is that

```python
ui.dataspace1d(23, 30, step=2, dstype=ui.DataPHA)
```

would create a channel range of `[1, 2, 3, 4, 5]`. It now is an error:

```
>>> ui.dataspace1d(23, 30, step=2, dstype=ui.DataPHA)
  File "<string>", line 1, in dataspace1d
  File "/lagado.real/sherpa/sherpa-xspec/sherpa/astro/ui/utils.py", line 1127, in dataspace1d
    raise DataErr("channel spacing is not 1")
sherpa.utils.err.DataErr: channel spacing is not 1
```

The code will complain if you say

```python
ui.dataspace1d(1.1, 1024, dstype=ui.DataPHA)
```

complaining that the start value is not an integer, but it will allow

```python
ui.dataspace1d(1, 1024.1, dstype=ui.DataPHA)
```

and will treat this as channels=[1, 2, ..., 1025]. I am not sure it's worth catching this particular case.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -4234,3 +4234,87 @@ def test_set_fit_checks_arg(session):
     with pytest.raises(ArgumentTypeErr,
                        match="^'rng' must be a Generator or None$"):
         s.set_rng(1234)
+
+
+@pytest.mark.parametrize("offset", [pytest.param(0, marks=pytest.mark.xfail), 1, pytest.param(5, marks=pytest.mark.xfail)])
+def test_dataspace1d_datapha_offset(offset):
+    """Ensure we can create the correct channel numbers for DataPHA"""
+
+    s = AstroSession()
+    detchans = 10
+    s.dataspace1d(offset, offset + detchans - 1, dstype=DataPHA)
+    d = s.get_data()
+    assert isinstance(d, DataPHA)
+    assert len(d.channel) == detchans
+    assert d.counts == pytest.approx(np.zeros(detchans))
+    chans = np.linspace(offset, detchans - 1 + offset, detchans)
+    assert d.channel == pytest.approx(chans)
+
+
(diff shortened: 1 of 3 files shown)


DougBurke:
@nplee - can you kick the tires on this simple change, just to make sure I'm not missing something (such as in the documentation...)

DougBurke:
This is separate from the needed `fake_pha` work, It just would make the
work-around I gave easier,

Doug

DougBurke:
Rebased to 

- pick up #2174 so I can more-easily test this with Python 3.13
- address missing coverage of an error condition
- minimal improvement of the documentation
- fix the commits so they pass iif run individually (the dangers of moving tests around and not testing fully)

DougBurke:
Given that we've not had users complain about the existing API, and the default is for `step=1, numbins=None`, I'm reluctant to make too many changes here until we get user questions/reports. So I'd prefer to go with what we've got rather than try to tweak the routine too much.

hamogu (approved):
Fine with me; I just wanted to bring it up.

PR #2215, opened 2025-01-24
Minor xspec additions

DougBurke:
# Summary

Minor changes to the XSPEC module, including added documentation to the set_xsabund routine to note the availability of the "felc" table, and set_xsstate no-longer requires all fields to be present.

# Details

This is a small subset of the changes from #1615 that do  not require changing the C++ interface. The changes are (not in order, and not per commit):

- an internal change to the tests so that mypy is less picky
- the read_xstable_model routine can now be sent a filename as a `Path` object as well as a string; this is an "internal" routine so it is not going to change most people's lives
- set_xsabund documentation is updated to mention the "felc" table
- get_xsabund has an extra example in its docstring
- updated the typing in `sherpa/astro/xspec/__init.py` to use Python 10 syntax (can avoid `Union` and `Optional`)
  - as part of this, use an overload for `get_xsabund`
  - which means should we annotate the non-overload version; previously we have not done this because why should we, but
    - I am reading proposed new documentation  https://github.com/python/typing/pull/1839 which doesn't exactly spell it out
    - however, by annotating all versions we do guarantee that `help(set_xsabund)` will see the types
    - it also stops a pyright complaint
    - and seems to mesh with this discussion (that it makes sense to add the types): https://discuss.python.org/t/conflict-between-python-typing-library-docs-and-mypy-docs-regarding-function-overload-usage/75087
 - set_xsstate used to require all settings are present before it would run, which can be useful, but it also means that if we add a new field (as I plan to) , or delete an old field, then old states can not be restored because it wouldn't know about the new settings. So we now allow each setting to be optional.

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -826,7 +826,10 @@ def test_get_xsstate_keys():
 
 
 @requires_xspec
-def test_set_xsstate_missing_key():
+@pytest.mark.parametrize("key",
+                         ["abund", "chatter", "cosmo", "xsect",
+                          "modelstrings"])  # paths is not a required key
+def test_set_xsstate_missing_key(key):
     """Check set_xsstate does nothing if required key is missing.
 
     """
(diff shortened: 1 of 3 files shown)


sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -826,38 +827,68 @@ def test_get_xsstate_keys():
...
+    assert key in ostate
     for val in ostate.values():
         assert val is not None
 
-    # paths is not a required key
+    def not_elem(value, vals):
+        """Pick the first item in vals that is not value"""
+        return [v for v in vals if v != value][0]
+
+    # We need valid arguments as the values will get set woth the
```
hamogu, line 849:
```suggestion
    # We need valid arguments as the values will get set with the
```

hamogu (approved):

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -826,38 +827,68 @@ def test_get_xsstate_keys():
...
         nstate = xspec.get_xsstate()
-        assert nstate == ostate
+        ncopy = nstate.copy()
+        del ncopy[key]
+
+        # The key should be unchanged; the others set to those in
+        # fake. Checking the latter is a bit annoying.
+        #
+        assert nstate[key] == ostate[key]
+        for key, value in fake.items():
```
DougBurke, line 872: argh - this over-writes the input `key` argument. Not a huge issue as such here but it needs cleaning up.

DougBurke: This was found when working no some follow-up code in #2216 but worth cleaning up here.

DougBurke: Addressed by commit f002119

DougBurke:
@hamogu - I made a  small change to a test since you accepted it (found when working on #2216 but worth cleaning up here). Can you just give the last commit a quick check? Ta

DougBurke:
Because it separates the input argument (that is parametrized, aka `miss_key`) from the generic variable name `key`. So I can use `key` in the rest of the loop and not get strange behaviour (as ended up happening in later work where the code would fail because I'd not separated the two concepts).

hamogu:
Makes sense - and that subtlety also explains why I didn't notice the change on review, when Gh highlights just the changes lines and not the others where the word "key" also appears.

PR #2216, opened 2025-01-30
Improve the XSPEC interface

DougBurke:
Most of this has been moved into (and merged) with 

- #2281.

The remaining changes are atomdb/nei version set/get, and this ends up being "interesting" as the XSPEC model library doesn't appear to handle changes particularly reliably (e.g. running `pytest sherpa/astro/xspec` will cause a segfault when `pytest-doctestplus` is installed because some setting has been changed internally within XSPEC).

# Summary

Add and extend routines in sherpa.astro.xspec: get_xsabundances now takes an optional table name, get_xsxset can be called with no argument, set_xsversion has been added to allow the nei and atomdb versions to be changed, get_xsversion can return the nei and atomdb versions, and clear_xsxset will clear the XSET database.

# Details

This PR builds on 

- #2215

The replacement for #1615 (what's left of it). I will note that in working on this code I have simplified some of the choices, so the exact functionality doesn't match #1615, but I believe we have what we need and the extra options provided in the earlier PRs are not actually needed.

Some of these code changes require C++11. They could be rewritten to not use C++11 features, but it does not seem worth it as our expected compilers have long-supported these capabilities.

A number of routines have been added or changed in `sherpa.astro.xspec._xspec` as this is treated as an "internal" module (i.e. if users are calling this directly then they have to expect changes). The user-level changes in `sherpa.astro.xspec` are intended to be more backwards-compatible, as normally we have just added routines or added optional arguments. There is one breaking change however, noted below.

Changes are:

- added `clear_xsxset` as an "obvious" way to clear the XSXSET database (there was a hidden way to do this but you had to know XSPEC internals to do so, by setting the `"INITIALIZE"` key)
- `get_xsxset` can now be called with no name, in which case it returns a dict containing all the settings
- `get_xsxset` now raises a `KeyError` if given an invalid name (to make the dict-like behaviour cleaner) rather than returning an empty string
  - this is an explicit breaking change
- `set_xspath_manager` can now be sent a `Path` or a `str`
- added `set_xspath_model` (which can be sent `Path` or `str`)
- `get_xsversion` can now return the `nei` and `atomdb` versions instead of the library version
- `set_xsversion` has been added to allow the `nei` and `atomdb` versions to be changed
- there have been internal changes to make `set_xsabundances` nicer, but it's not visible to the user, but it does mean that if you have set abundances from a file (or adjusted the values) then they will be recorded in the "xspec state"
- added `get_xsabundances_path` as a way to get the `Path` to the abundances file used by XSPEC
[...]

PR #2106, opened 2024-08-19
Initial support for XSPEC filtering keywords (XFLT)

DougBurke:
- so the main UI should be something like `add_xflt` which takes a dataset (or an identifier for the UI layer) and returns a `SpectrumNumber`, there's some form of `show_xflt`, and `clear_xflt`; we don't really need much more (the lower-level code needs to have more)
- setting the `spectrum` attribute of a model can warn if there's no data, since the idea is that you've got a dataset
- the XSPEC model `startup` method can check whether the `spectrum` values have data for; it would also be nice to check that we don't "mix" `spectrum` values between different source models (or, perhaps more importantly, within the same source expression) but it's hard to get the balance right

I am likely going to do this in 

- #2224 

but I want to leave this available for now

# Summary

Add experimental support for XSPEC models that use filtering keywords (e.g. XFLT0001) and are marked as per-spectrum in the XSPEC model files. This adds the xssmaug, xspolconst, xspollin, and xspolpow XSPEC models and the load_xsxflt command.
 
# Details

To support the per-spectrum XSPEC models we need

- a way to set / get  the XFLT keyword database values - that is associate values with a given "spectrum number"
- a way to tell the XSPEC model to use the correct "spectrum number"

XSPEC can do this all automatically, but we

- do not require an integer value as a dataset identifier (in the UI layer)
- do not have the concept of an identifier when evaluating a model, as all we care about is the grid to evaluate the model on

So, matching these two systems is not simple and means the Sherpa user has to know about this all, unlike the XSPEC user. Which is why this is labelled experimental.

The XSPEC models provided in the model library are not particularly exciting for Sherpa (there is `smaug`, and then the other models require better handling of polarimetric errors than we currently provide). So it  may be more useful for user-weitten models.

However, I had been hoping to use this to then get mixing models - which also use the XFLT database - into Sherpa, but the current API for these models is not currently  designed for external use (i.e. outside XSPEC proper), as reported to me by an XSPEC developer [there are strong reasons for this]. Which perhaps limits the interest in this.

# Examples

## UI

Let's load in a PHA file three times, so we can see whether the model values change with the XFLT keywords:

```
In [1]: from sherpa.astro import ui

In [2]: ui.load_pha(1, "sherpa-test-data/sherpatest/3c273.pi")
...

In [3]: ui.load_pha(2, "sherpa-test-data/sherpatest/3c273.pi")
...

In [4]: ui.load_pha(3, "sherpa-test-data/sherpatest/3c273.pi")
...

[...]

PR #2224, opened 2025-02-10
Extend xspec

DougBurke:
The last part of #1615 and some additions

- adds support for the XFLT keywords
  - need to work on integrating in the last changes from #2106, but I want to change how the `spectrumNumber` argument is sent in; it shouldn't be a property of the model as we want the model to be usable with multiple datasets, but then how do we send it in
- adds support for the model database
  - it's not clear any model makes use of this, but it's easy to add 
- add basic support for the `plot dem` command
  - it's very limited, but that's because the XSPEC API is very limited
 
It requires

- #2215 
- #2216

### DIFF
--- a/docs/model_classes/xspec_model.rst
+++ b/docs/model_classes/xspec_model.rst
@@ -49,6 +49,7 @@ for a general description of the ``sherpa.astro.xspec`` module.
       set_xschatter
       set_xscosmo
       set_xspath_manager
+      set_xspath_model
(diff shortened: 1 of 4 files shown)


DougBurke:
Rebased as #2116 has been merged.

DougBurke:
Rebased as https://github.com/sherpa/sherpa/pull/2252 was merged.

PR #2220, opened 2025-02-01
CI: weekly run on WSL

DougBurke:
# Summary

Add a GitHub workflow which builds and tests Sherpa on WSL. Fix #2218.

# Details

This started out as following `ci-pip-arch.yml` but my work in #1948 fleshed it out somewhat. Since WSL is not a primary platform for Sherpa I have made this workflow run weekly, rather than per-PR, following the `ci-pip-arch.yml` approach.

I elected to add relatively "full" builds using both pip and conda (the latter including XSPEC but not DS9), but only for a single Python version. I would love to be able to rewrite the workflow to avoid repetition (e.g. setting up WSL) but my attempts to do so failed spectacularly and I just didn't feel it worth spending time on. The conda version attempts to mimic the existing conda workflow, but issues over subtle differences in environment (e.g. what is installed), the different requirements (no DS9 here), and the fact that there are issues between transitioning between Windows- and Linux-style environment variable naming, mean that it's not a simple copy.

I did a commit with the WSL build being built as part of the PR, to shake out the many bugs, and then one to turn it into a weekly run. Which of course means you can't see that it passes.

PR #2230, opened 2025-02-13
Note that we need the file program to build Sherpa

DougBurke:
# Summary

Note that the file command is needed to build Sherpa (it is used by extern/configure).

# Details

I was trying to build Sherpa in a docker container to try and work out what's going on in #2220 and in reviewing the output of `extern/configure` realized it wanted the `file` command. So

a) add a note of this to the documentation
b) bump the one GH action that install stuff to include `file` just to make sure
  - as part of this, remove the installation of `git` since this isn't needed to build Sherpa
  - bump the Ubuntu version from `22.04` to `latest` (aka the latest LTS)

It is not really clear what problem not having `file` leads to, if any (since you'd need to track what `configure` was doing with the information), but we may as well note the requirement. Also, most users will already have it.

### DIFF
--- a/.github/workflows/ci-pip-arch.yml
+++ b/.github/workflows/ci-pip-arch.yml
@@ -44,21 +44,21 @@ jobs:
       uses: uraimo/run-on-arch-action@v2
       with:
         arch: ${{ matrix.arch }}
-        # distro: ubuntu_rolling
-        # ubuntu 22.04 uses Python 3.10
-        distro: ubuntu22.04
+        # use the latest LTS version
+        distro: ubuntu_latest
 
         shell: /bin/bash
 
         # Based on AstroPy
         install: |
           apt-get update -q -y
-          apt-get install -q -y git \
+          apt-get install -q -y \
                                 g++ \
                                 gfortran \
                                 flex \
                                 bison \
                                 make \
+                                file \
                                 pkg-config \
                                 python3 \
                                 python3-dev \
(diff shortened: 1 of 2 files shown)


hamogu (approved):
Look good. No harm either way. 
For some systems, configure might use an alternative if file is not present- that the point of configure, to figure out what is present and work around it was much as possible if not. I don't know for `file` specifically iii configure knows alternatives to use and if so for what systems, but for out main targets (Linux, MacOS) `file` is a default tool, so let's just list it in the docs.

.github/workflows/ci-pip-arch.yml
```diff
@@ -44,21 +44,21 @@ jobs:
       uses: uraimo/run-on-arch-action@v2
       with:
         arch: ${{ matrix.arch }}
-        # distro: ubuntu_rolling
-        # ubuntu 22.04 uses Python 3.10
-        distro: ubuntu22.04
+        # use the latest LTS version
+        distro: ubuntu_latest
```
wmclaugh, line 48: Do we want to keep distro pinned to ubuntu 24.04 rather than justr taking 'ubuntu_latest'?

DougBurke: So, the `distro` setting is used by https://github.com/uraimo/run-on-arch-action and has a limited number of values - see https://github.com/uraimo/run-on-arch-action?tab=readme-ov-file#supported-platforms

So, this change switches from ubuntu 22.04 to the latest version, which is currently ubuntu 24.04. According to the documentation there is no `ubuntu-24.04` setting, which is why I picked `ubuntu-latest` as the most sensible choice here.

PR #2231, opened 2025-02-14
Two small astropy-related doc changes

hamogu:
## Summary

- Update linke to astropy dev docs (that we link to in lieu of writing our own dev docs)
- Recommend to install astropy via pip, not conda

## Details

I find that the conda dependency resolver fails (or at least takes > 2 h) to resolve `conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge ds9 ciao astropy`, while it takes only seconds without astropy (This is specifically on MacOS X with Python 3.11, but even if other combinations are faster, this change doesn't hurt). So, install astropy with pip instead, once the conda environment is set up.

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -14,7 +14,7 @@ problems or feature requests on github
 At present we do not have any explicit documentation on how
 to contribute to Sherpa, but it is similar to other open-source
 packages such as
-`AstroPy <https://docs.astropy.org/en/stable/index.html#contributing>`_.
+`AstroPy <https://docs.astropy.org/en/stable/index_dev.html>`_.
 
 The developer documentation is also currently lacking.
 
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -162,12 +162,13 @@ Conda can be used to install all the dependencies for Sherpa, including
 
 ::
 
-    conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge ds9 astropy ciao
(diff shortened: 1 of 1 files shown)


DougBurke:
So, historically you can find quite strident advice that mixing `conda` and `pip` is a "bad idea™":

- https://stackoverflow.com/questions/56134588/is-it-a-bad-idea-to-use-conda-and-pip-install-on-the-same-environment#56141684

The conda documentation is more-nuanced, but bibbles on about things I've never cared about

- https://docs.anaconda.com/working-with-conda/packages/pip-install/

Do we care?

hamogu:
We don't care. We mix it anyway (see installation of the requirements to build the docs just a few lines below). I've never seen problems.

Theoretically, problems can occur when you upgrade a conda package that a pip-installed package depends on. But astropy usually has a wide range of numpy it supports and this is for a development environment. Developers (a) are more likely to recognize problems when they occur and (b) probably nuke the env every now and then anyway, e.g. I do that every time we publish a new CIAO version, which goes with a new Python version, so the env is never so old that such compatibility problems can arise.

DougBurke (approved):

wmclaugh:
@hamogu 

I can run the command 

_conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge ds9 ciao astropy_

on my intel macbook running Ventura and it completes in under 10 minutes without issue. While older threads on the topic of conda with pip are more cautious about mixing the two, it seems to me that more recently the attitude has shifted to try conda first and if the package isn't available/working through conda then fall back to pip to help reduce the possibility of issues.  Perhaps we change the doc to have a similar modus operandi? Not a sticking point- I can merge the change as is if you prefer.

PR #1948, opened 2024-01-08
Switch the build system to meson

DougBurke:
# Summary

Can we build Sherpa with meson?

# Details.

Yes.

Well, kind of.

We can see it building for Linux and macOS (both x86 and arm). It also includes "optional" code such as XSPEC.

**BUT** it builds and the tests pass. So :shipit:

Known issues include

- [ ] I have not tried to verify exactly what's in the sdist or wheel, so there could well be problems (I'm expecting some linkage problems).
  - initial tests have been successful (I have been able to use both the sdist and wheel) but that is on the same system/environment so limited applicability 
- [x] I have not tried building against CIAO and I *fully expect* that to shake out some buglets 
       `pip install . --verbose -Csetup-args=-Dxspec-prefix=$ASCDS_INSTALL -Csetup-args=-Dxspec-libraries=XSFunctions,XSUtil,XS -Csetup-args=-Dbuild-group=false -Csetup-args=-Dbuild-stk=false -Csetup-args=-Dbuild-region=false -Csetup-args=-Dregion-prefix=$ASCDS_INSTALL -Csetup-args=-Dbuild-wcssubs=false -Csetup-args=-Dwcssubs-prefix=$ASCDS_INSTALL`
    - actually, this seemed to work (after shaking out the buglets)
- [X] setting the version
  - [ ] this has fun consequences for building the documentation
  - this has "kind of" been fixed, but it still uses `versioneer` but is this the best way to do it, and it requires that `python3` is available
- [ ] the current code assumes that the FFTW3 module and include directories [?] are already installed, rather than building it ourselves (since I ran out of time trying to glue together meson and autoconf/tools)
- [X] the build documentation has not been updated
  - a first pass has been made
- [ ] the setuptools-related code is still present
- [X] `extern` has been renamed `subprojects` as this seems like a sensible way to go with `meson`, except that these are almost not-really OTS; there's also a question of whether to version-number the directories
- [ ] why do the tests in `test_tests.py` fail; somehow the assert rewrite that `pytest` does is not being triggered, possibly because we now use `--no-build-isolation`;
    - see https://github.com/mesonbuild/meson-python/issues/646

## Editable install success

With an editable install (and `--no-build-isolation`) I edited a `.cc` file and a test (to use the new compiled module feature) and `pytest ..../test...py` actually recompiled just the code I'd changes and ran the test using it. This is noice.

### libraries

Here's the ldd output for `sherpa/astro/_xspec` from CIAO 4.17

```
[...]

PR #2232, opened 2025-02-19
TESTS: mark more that require the group module

DougBurke:
# Summary

Mark several recently-added tests with the requires_group decorator. There is no functional change.

# Details

It would be nice to track these - either by some automated means or with a CI build - but this build configuration is rare so I don't think it's worth it.

An alternative would be to drop the "can build Sherpa without the group module/library" but that is a much bigger issue so we may as well take this simple change anyway.

I found this when working on 
- #1948
but it is completely separate from that work (other than the fact that reworking the build system gets you into places where we don't  have the group module, so you can't run the tests!).


hamogu (approved):
Sure, why not mark them? I agree that going any further is no justified at this time.

PR #2235, opened 2025-02-25
Deal with OGIP products that do not follow the standards

### LINKED ISSUE #2185: Sherpa 4.17.0 cannot read some rmf files
Hi, I have encountered errors reading some of my response files in Sherpa 4.17.0. I also tried Sherpa 4.16.0, and no error appears. 
I have attached the data.

```python
Python 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]
IPython 8.18.1 -- An enhanced Interactive Python. Type '?' for help.
No event loop hook running.
Using matplotlib backend: agg

In [2]: ui.load_data('bgd1B_sr.pha')
WARNING: unable to read RMF: '<' not supported between instances of 'str' and 'int'

In [3]: import sherpa

In [4]: sherpa.__version__
Out[4]: '4.17.0'
```

```python
Python 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:36:13) [GCC 12.3.0]
IPython 8.24.0 -- An enhanced Interactive Python. Type '?' for help.
No event loop hook running.
Using matplotlib backend: agg

In [2]: ui.load_data('bgd1B_sr.pha')
read RMF file bgd1B_sr.rmf

In [3]: import sherpa

In [4]: sherpa.__version__
Out[4]: '4.16.0'
```

[nustarspec.tar.gz](https://github.com/user-attachments/files/17517891/nustarspec.tar.gz)

DougBurke:
# Summary

Some OGIP products from missions occasionally do not follow the OGIP and FITS standards. Provide workarounds for two recent cases from NuSTAR and NICER (TLMIN/MAX values were recorded as a string, and a QUALITY array was stored as a boolean array). Fix #2185.

# Details

The first commit adds tests that show the problem. Rather than add files to the sherpa test data, we manually create the problematic files on the fly. This has the advantage of being able to be precise in showing the problems (e.g. the bits where the OGIP/FITS standards are not being followed) and avoids the need to deal with adding more files to the test data area. There's also the issue that (hopefully) these are "ephemeral" files - e.g. the code that created them has - or will be - updated to fix the issue.

As shown in the tests, both FITS backends fail when reading in these files, but in different ways.

The first fix is for #2185. This is only for the astropy backend because the crates behaviour is internal to the crates module (any attempt to get the tlmin/max values when encoded as a string returns 0, and there's no obvious way to work around this [*]). I did discuss whether AstroPy should work around this in https://github.com/astropy/astropy/issues/17254 but it's felt better to address this problem here.

The second fix is for the QUALITY is boolean issue, which came in from a CIAO helpdesk ticket from NICER data which had a file like the following, with `QUALITY` having a type of `Byte` rather than `Int2`:

```
% dmlist ni3013010109mpu7_sr3c50.pha cols
 
Columns for Table Block SPECTRUM
 
ColNo  Name                 Unit        Type             Range
   1   CHANNEL                           Int4           0:1500               Pulse Invarient (PI) Channel
   2   COUNTS               count        Int4           -                    Counts per channel
   3   SYS_ERR                           Real8          -Inf:+Inf            Fractional systematic error
   4   QUALITY                           Byte                                Spectrum quality flag (0=good)
   5   GROUPING                          Int2           -                    Spectrum bin grouping flag
```

This is okay for AstroPy, which returns a 1D array of np.uint8 values, but for some reason crates reads this data in and creates a 2D array of np.uint8 values, with shape (nrows, 1). So the fix is to manually convert such an array back to 1D. This should be okay because np.uint8 is not a native FITS type, so is only used for Bit/Byte columns, and so it should be safe for the crates backend to make this conversion (also, it should only be being used in cases where we expect 1D arrays).

The remaining four commits are code clean ups, mainly to the tests, and were based on `ruff check`. This is a combination of preemptive  future-proofing and avoiding duplicated test code:

- lint checks reported by `ruff check` which include:
 - remove unused module and symbol imports
 - use `is` rather than `==` for a type check (the alternative is to use `isinstance` but that is not always the correct choice so I want to limit the changes here)
 - be explicit about where some symbols are coming from (e.g. the ui "magic" that both creates a model instance and creates a symbol to represent the instance is something the linters and type checkers find hard to understand)
 - no point in using a f-string if the string contains no "{}" elements
- we can drop `Optional` and `Union` from the typing statements as we now have a minimum of Python 3.10
- we can import several symbols from `collections.abc` rather than `typing` (these changes require Python 3.9)

The duplicated test code is a mixture of

- genuine duplicates, so we can remove one of them
- accidental name shadowing, so we need to fix up one or both names
  - and then in one case fix a test (which had been shadowed so not run) which actually wasn't doing what it should have been doing
  - update a related test to add parametrization over a function argument 

Not all of these files are directly related to the code I needed to change to fix the bugs, but they are somewhat related (that is, they are in the sherpa/astro/ tree). We could move them out to a separate PR but I didn't think it worthwhile.

---

[*] for the problem case in #2185 the existing crates behaviour happens to be the correct one, but this is purely fortuitous....

### DIFF
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2015, 2016, 2019 - 2024
+#  Copyright (C) 2011, 2015, 2016, 2019 - 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -675,8 +675,15 @@ def copycol(cr: TABLECrate,
     Historically Sherpa has converted RMF columns to double-precision
     values, even though the standard has them as single-precision.
     Using the on-disk type (e.g. float rather than double) has some
-    annoynig consequences on the test suite, so for now we retain this
+    annoying consequences on the test suite, so for now we retain this
     behaviour.
+
+    NICER has created QUALITY columns with TFORM=B which crates reads
+    in as a 2D array, but of shape (ncols, 1). So special case this,
+    as we assume there are not many csaes where we are reading in this
+    data (unfortunately it's not easy to check for exactly this
+    condition as crates does not provide direct access to the TFORMn
+    header value).
 
     """
 
@@ -688,6 +695,15 @@ def copycol(cr: TABLECrate,
     vals = col.values
     if dtype is not None:
         vals = vals.astype(dtype)
+    elif vals.dtype == np.uint8 and vals.ndim == 2 and \
+         vals.shape[1] == 1:
+        # Convert boolean (nrows, 1) 2D array, created by the CIAO
+        # data model in CIAO 3.17, to 1D. Fortunately there is no
+        # native FITS TFORM that maps to np.uint8 other than
+        # boolean/bit values, so this should be a relatively safe
+        # expansion.
+        #
+        vals = vals.reshape(vals.size)
 
     # do not expand out variable-length arrays for now.
     #
--- a/sherpa/astro/io/crates_backend.py
+++ b/sherpa/astro/io/crates_backend.py
@@ -18,11 +18,12 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from collections.abc import Sequence
 from collections import defaultdict
 from contextlib import suppress
 import logging
 import os
-from typing import Any, Optional, Sequence, Union
+from typing import Any
 
 import numpy as np
 
@@ -62,8 +63,8 @@ name: str = "crates"
 """The name of the I/O backend."""
 
 
-CrateType = Union[TABLECrate, IMAGECrate]
-DatasetType = Union[str, CrateDataset]
+CrateType = TABLECrate | IMAGECrate
+DatasetType = str | CrateDataset
 
 
 def open_crate(filename: str) -> CrateType:
@@ -160,7 +161,7 @@ def get_filename_from_dmsyntax(filename: str) -> str:
 
 def _try_key(crate: CrateType,
              name: str,
-             dtype: type = str) -> Optional[KeyType]:
+             dtype: type = str) -> KeyType | None:
     """Access the key from the crate returning None if it does not exist.
 
     There's no way to differentiate between a key that does not exist
@@ -201,7 +202,7 @@ def _try_key(crate: CrateType,
 
     # byte strings are to be decoded to strings
     #
-    if dtype == str and isinstance(value, bytes):
+    if dtype is str and isinstance(value, bytes):
         return dtype(value, "utf-8")
 
     return dtype(value)
@@ -258,7 +259,7 @@ def _require_image(crate: IMAGECrate,
 
 
 def _get_crate_by_blockname(dataset: CrateDataset,
-                            blkname: str) -> Optional[CrateType]:
+                            blkname: str) -> CrateType | None:
     """Select the given block name.
 
     Parameters
@@ -292,7 +293,7 @@ def _get_crate_by_blockname(dataset: CrateDataset,
 
 # This should be added to the official backend API.
 #
-def read_table_blocks(arg: Union[str, CrateDataset, TABLECrate],
+def read_table_blocks(arg: str | CrateDataset | TABLECrate,
                       make_copy: bool = False
                       ) -> tuple[BlockList, str]:
     """Read in tabular data with no restrictions on the columns."""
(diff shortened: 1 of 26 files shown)


DougBurke:
@hamogu - note that the two test failures are from the "pip-arch" run which is not done normally - these failures are not related to these changes so I'm ignoring them here (and created 

- #2236 

to track it). I'm assuming there will be some changes from the review so we can get a clean build soon.

sherpa/astro/tests/test_astro_data2.py
```diff
@@ -6158,3 +6169,412 @@ def test_rmf_get_dep_complex():
...
+    chans, counts, quals = make_problem_pha(outpath, offset)
+
+    pha = read_pha(str(outpath))
+    assert isinstance(pha, DataPHA)
+
+    assert pha.channel[0] == offset
+
+    assert pha.channel == pytest.approx(chans)
+    assert pha.counts == pytest.approx(counts)
+    # convert bools to nitegers to please pytest 8.3.4
```
hamogu, line 6508: nitegers -> integers. Not that it matters, but just so you know I actually read every line...

sherpa/astro/io/crates_backend.py
```diff
@@ -688,6 +695,15 @@ def copycol(cr: TABLECrate,
     vals = col.values
     if dtype is not None:
         vals = vals.astype(dtype)
+    elif vals.dtype == np.uint8 and vals.ndim == 2 and \
+         vals.shape[1] == 1:
+        # Convert boolean (nrows, 1) 2D array, created by the CIAO
+        # data model in CIAO 3.17, to 1D. Fortunately there is no
```
hamogu, line 702: 3.17 -> 4.17 ?

hamogu (approved):
I general this is all good except for two typos in comments in tests (which is about the lowest level of importance one can possibly imagine.

As an added note: I was initially less than enthusiastic about typing and this PR is a good example why: The majority of lines changed are just churn as the type annotation rules become better in Python 3.10 compared to 3.9. Maybe the time has come for me to embrace typing as well as it seem that the most common cases (optional arguments, Union) have been settled and the uncommon cases where typing still evolves from 3.12 to 3.13 are, well uncommon and will need much fewer changes when we embrace 3.13.

sherpa/astro/io/crates_backend.py
```diff
@@ -675,9 +676,16 @@ def copycol(cr: TABLECrate,
     Historically Sherpa has converted RMF columns to double-precision
     values, even though the standard has them as single-precision.
     Using the on-disk type (e.g. float rather than double) has some
-    annoynig consequences on the test suite, so for now we retain this
+    annoying consequences on the test suite, so for now we retain this
     behaviour.
 
+    NICER has created QUALITY columns with TFORM=B which crates reads
+    in as a 2D array, but of shape (ncols, 1). So special case this,
+    as we assume there are not many csaes where we are reading in this
```
wmclaugh, line 684:
```suggestion
    as we assume there are not many cases where we are reading in this
```

sherpa/astro/io/crates_backend.py
```diff
@@ -675,9 +676,16 @@ def copycol(cr: TABLECrate,
     Historically Sherpa has converted RMF columns to double-precision
     values, even though the standard has them as single-precision.
     Using the on-disk type (e.g. float rather than double) has some
-    annoynig consequences on the test suite, so for now we retain this
+    annoying consequences on the test suite, so for now we retain this
     behaviour.
 
+    NICER has created QUALITY columns with TFORM=B which crates reads
+    in as a 2D array, but of shape (ncols, 1). So special case this,
```
wmclaugh, line 683: grammar- "but of shape"

DougBurke, line 683:
```suggestion
    in as a 2D array, with a shape of (ncols, 1). So special case this,
```

sherpa/astro/io/crates_backend.py
```diff
@@ -688,6 +696,15 @@ def copycol(cr: TABLECrate,
     vals = col.values
     if dtype is not None:
         vals = vals.astype(dtype)
+    elif vals.dtype == np.uint8 and vals.ndim == 2 and \
+         vals.shape[1] == 1:
+        # Convert boolean (nrows, 1) 2D array, created by the CIAO
+        # data model in CIAO 3.17, to 1D. Fortunately there is no
```
DougBurke, line 702:
```suggestion
        # data model in CIAO 4.17 (and earlier), to 1D. Fortunately there is no
```

PR #2238, opened 2025-03-02
tests: ensure the version matches

DougBurke:
# Summary

Improve testing of the Sherpa version number. There is no functional change.

# Details

We have
    
    sherpa.__version__
    importlib.metadata.version("sherpa")
    
which should be the same, so check this.
    
However, editable installs present an issue, since the `sherpa.__version__` field will get a ".dirty" suffix as soon
as a git-tracked file is edited, but the metadata version will not *if* it was installed before any edits were made (even with no-build isolation, at least for setuptools).
    
This test was suggested by the 2024-10-07 version of
    
https://packaging.python.org/en/latest/discussions/single-source-version/

and added, in part, because experiments with switching to the meson build system from setuptools showed it would be a useful test. However, it may be too annoying given the editable issue, so we may decide to revert this.

### DIFF
--- a/sherpa/tests/test_sherpa.py
+++ b/sherpa/tests/test_sherpa.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2016, 2018, 2020, 2021, 2022
+#  Copyright (C) 2007, 2015, 2016, 2018, 2020-2022, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -18,6 +18,7 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+import importlib.metadata
 import os.path
 
 import pytest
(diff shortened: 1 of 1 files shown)


hamogu (commented):
Right now, I'm working in an editable install wher I've switched branches frequenquntly, I sometimes install to make sure stuff gets compieled when I pull main the main branch with C++ code changes. In other wrods, it's messy. But I get
```
In [4]: got, expected
Out[4]: ('4.17.0+71.gb37e03c7.dirty', '4.17.0+62.ged586b52.dirty')
```

I'm not concerned about that, but I think it's a common situation for developers. Maybe that's my fault and I should run all tests in isolated environments with tox or nox and not right in my conda environment. People tell me that using conda for environments is so 2022...

So, will I have in the future that one test that always fails for me when I run it locally? I guess I can live with that, if I have to. Or is that indicating that my development practices need to be cleaned up? Or should we only do the comparison when at least one of the versions is not dirty?

DougBurke:
So, I guess there's several overlapping thoughts here

1. work out why `versioneer` isn't setting the version correctly; I note that it is not clear how much development is going on with it nowadays, but as we are using the "versioned" copy (that is, we have it included here rather than brought in as part of the build) we can hack it as much as we want (which I'd prefer not to) 
2. work out  away to get @hamogu to build software "properly" (less snarky is to come up with an "approved" development workflow, but I'm not sure I want to do that)
3. hope that moving to a new backend will resolve all these issues (**spoiler** I don't think it does, but we could change to a different versioning system, whether manual or pick an alternative to `versinoeer`)
4. we could make this test optional, that is require a flag to be run (as we do with a few other tests); this is probably better than the "drop the test if the name ends in .dirty" as I think that's a bit too generous)
5. drop this PR as it's not worth our effort

I'll think about item 4, but comments welcome.

hamogu:
We can also just let @hamogu deal with it. He knows about the test, he can ignore it when appropriate. The real concern is if this impacts others people who might try to open a PR to Sherpa, find a constantly failing test (this one) when developing locally that they can't figure out and that seems to appear randomly and then they decide not to open a PR.

DougBurke:
So, I think I'm not going to really push on this, as

- it doesn't add much to our tests
- there are complications
- we may be going to change how the version code is set, for various reasons

but I'll leave this open for now.

DougBurke:
marking as draft as I'm not sure it's worth pursuing at this time

DougBurke:
I am closing this as - until we redo our build system - it isn't worth the confusion it might bring.

PR #2239, opened 2025-03-03
REGION: test CXCDM region parser, when available

DougBurke:
# Summary

Allow the region module to know how it was built (whether reading FITS binary region files is supported). This allows for improvements to the tests.

# Details

The region code allows binary region files to be read in, but ONLY if it was built with the `USE_CXCDM_PARSER` macro set. So we can add a test or two of this behaviour, but how does the test know what behaviour to expect? We could use a heuristic, such as "does this appear to be a CIAO build", but it is **MUCH** better to just encode the macro setting into the module so we can ask it. We do this with the new `sherpa.astro.utils._region.USE_CXCDM_PARSER` constant. It is 1 if it was set and 0 otherwise.

It appears that even though it was added with `PyModuke_AddIntConstant` you can actually change this value. I do not see the point in trying to deal with this as this is a highly-specialized variable which users should not know about.

I wanted this to make it more obvious I was doing the right thing in 

- #1948

### DIFF
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -1,5 +1,5 @@
 //
-//  Copyright (C) 2009, 2016, 2020, 2021, 2022
+//  Copyright (C) 2009, 2016, 2020-2022, 2025
 //  Smithsonian Astrophysical Observatory
 //
 //
@@ -379,6 +379,18 @@ PyInit__region(void)
     Py_DECREF(&pyRegion_Type);
     Py_DECREF(m);
   }
+
+  // Add a symbol to indicate whether the module was built with the
+  // USE_CXCDM_PARSER macro set.
+  //
+  PyModule_AddIntConstant(m, "USE_CXCDM_PARSER",
+#ifdef USE_CXCDM_PARSER
+			  1
+#else
+			  0
+#endif
+			  );
+
   return m;
 
 }
--- a/sherpa/astro/utils/src/_region.cc
+++ b/sherpa/astro/utils/src/_region.cc
@@ -383,13 +383,11 @@ PyInit__region(void)
   // Add a symbol to indicate whether the module was built with the
   // USE_CXCDM_PARSER macro set.
   //
-  PyModule_AddIntConstant(m, "USE_CXCDM_PARSER",
 #ifdef USE_CXCDM_PARSER
-			  1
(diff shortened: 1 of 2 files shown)


DougBurke:
I know this touches the C++ code, but hopefully it's not too hard to understand @hamogu

hamogu:
It's OK, it's not that I'm illiterate. I can read most C++, but don't ask me e.g. to weigh the pros and cons of template argument deduction in class templates.

sherpa/astro/utils/src/_region.cc
```diff
@@ -379,6 +379,18 @@ PyInit__region(void)
     Py_DECREF(&pyRegion_Type);
     Py_DECREF(m);
   }
+
+  // Add a symbol to indicate whether the module was built with the
+  // USE_CXCDM_PARSER macro set.
+  //
+  PyModule_AddIntConstant(m, "USE_CXCDM_PARSER",
```
wmclaugh, line 386: The #ifdef preprocessor wraps lexical value C commands within the conditional text segment. As such, it should be written as:

#ifdef USE_CXCDM_PARSER
	PyModule_AddIntConstant(m, "USE_CXCDM_PARSER", 1);
#else
	PyModule_AddIntConstant(m, "USE_CXCDM_PARSER", 0);
#endif

Alternatively, have you tried using?

    PyModule_AddIntMacro(m, USE_CXCDM_PARSER);

DougBurke: So, I'm not sure what `PyModule_AddIntMacro` does when the macro isn't defined (or is defined but has "no value").

I guess we could change how the code is built so that rather than sending in `-DUSE_CXCDM_PARSER` only when we want the macro set we'd send in either `-DUSE_CXCDM_PARSER=0` or `-DUSE_CXCDM_PARSER=1` but that's a larger change that I was hoping to avoid.

Unless there's a reason I would prefer to go with the initial suggestion - i.e.

```
#ifdef USE_CXCDM_PARSER
PyModule_AddIntConstant(m, "USE_CXCDM_PARSER", 1);
#else
PyModule_AddIntConstant(m, "USE_CXCDM_PARSER", 0);
#endif
```

DougBurke: I can confim that `PyModule_AddIntMacro` needs the macro to exist, so the easiest change is to duplicate the lines.

PR #2242, opened 2025-03-12
Allow building Sherpa without the FFTW library

DougBurke:
# Summary

Allow Sherpa to be built without requiring the FFTW library, which causes the convolution/PSF models to be unusable. Adds in support for automating determining the build information if the FFTW pkg-config information is available.

# Details

This is technically irrelevant if we take 

- #1948

(or equivalent) but we would then want to make sure we can build without support for FFTW in that PR.

If the FFTW library is not available we can not build the `sherpa.utils._psf` module, but we can still do quite a lot. This PR makes the FFTW library be an optional extra rather than required. The main changes are

- add the `disable-fftw` build action (in `setup.cfg`)
- allow the build to automatically determine the FFTW install and library directories via `pkg-config`
  - this is for when the user has not explicitly set the locations in the `setup.cfg` file
- warning users when the `_psf` module is not available
- marking those tests as optional that require the `_psf` module

In general this is as expected, but there are a couple of image tests that have been shown to require the _psf module even though they do not actually do any PSF convolution. This is because they require the "bounding box" routines. We could look at separating this from the PSF code if we felt it worthwhile. However, this is probably not worth it as these code paths require other optional build elements (e.g. the region code), which suggests it is not often users would hit this condition.

Before this, we have updated the CI runs to install FFTW directly (via `apt` or `conda`) rather than building it from the `extern/` directory. Once support for disabling the FFTW build is available we change one of the pip builds to build this way (as a check it works), although technically we would also want to do a build with the data directory and fits support to more-thoroughly check the `@requires_psf` decorator,

See also

- #2298

# Notes

One fun thing is that, with the FFTW build disabled

```python
from sherpa.utils import _psf
assert _psf is None
```

but

```python
import sherpa.utils._psf
```

will raise `ImportError`.


DougBurke:
Rebased to pick up #2445 and updated copyright dates.

DougBurke:
So, this would need rebasing as 

- #2484 

was merged. Is it worth it?

PR #2243, opened 2025-03-13
Fit typing improvements

DougBurke:
# Summary

TBD

# Details

Taken from #2022 

The idea is to take the "add typing rules and clean up the code" changes here, so that we can then work on #2007 in #2022 (which first requires working out what to do with statargs/kwargs, which we do not try to address here, other than hopefully making it clearer how information flows through the fit/optimisation/error estimation code).

### DIFF
--- a/sherpa/estmethods/tests/test_estmethods.py
+++ b/sherpa/estmethods/tests/test_estmethods.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2018, 2021, 2023
+#  Copyright (C) 2007, 2018, 2021, 2023, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -64,6 +64,15 @@ minpars = numpy.array([1, 0, 0])
 hardmaxpars = numpy.array([1.0e+120, 1.0e+120, 1.0e+120])
 hardminpars = numpy.array([1.0e-120, -1.0e+120, -1.0e+120])
 
+# Avoid accidentally changing these values during the tests.
+#
+for ary in [x, y, fittedpars, limit_parnums, maxpars, minpars,
+            hardmaxpars, hardminpars]:
+    ary.setflags(write=False)
+
+del ary
+
+
 gfactor = 4.0 * 0.6931471805599453094172321214581766
 
 
@@ -80,7 +89,7 @@ def report_progress(i, lower, upper):
 
 
 def get_par_name(ii):
-    pass
+    assert False, "should not be called"
 
 
 # Here's the 1D Gaussian function to use to generate predicted
@@ -181,9 +190,6 @@ def test_covar():
                           limit_parnums, freeze_par, thaw_par,
                           report_progress, get_par_name)
 
-    # These tests used to have a tolerance of 1e-4 but it appears to
-    # be able to use a more-restrictive tolerance.
-    #
     expected = standard.diagonal()
     assert results[1] == pytest.approx(expected)
 
@@ -205,10 +211,6 @@ def test_projection(parallel):
                            limit_parnums, freeze_par, thaw_par,
                            report_progress, get_par_name)
 
-    # These tests used to have a tolerance of 1e-4 but it appears to
-    # be able to use a more-restrictive tolerance (given that the
-    # "fitter" doesn't do a fit here).
-    #
     assert results[0] == pytest.approx(standard_elo)
     assert results[1] == pytest.approx(standard_ehi)
 
--- a/sherpa/estmethods/tests/test_estmethods.py
+++ b/sherpa/estmethods/tests/test_estmethods.py
@@ -89,7 +89,7 @@ def report_progress(i, lower, upper):
 
 
 def get_par_name(ii):
-    assert False, "should not be called"
+    pass
 
 
 # Here's the 1D Gaussian function to use to generate predicted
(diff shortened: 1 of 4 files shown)


DougBurke:
@hamogu 

- the test failure is a process failure and nothing to do with this PR
- this has a lot of commits, but they are generally small, so I would suggest that this PR is one where it is better to review per-PR
- there are a lot of things we could change, but 
  - some of them may be done in later commits in this PR
  - some of them may be done in PRs which are based on this
  - some of them are just ones I don't have a lot of time to worry about ;-)

sherpa/optmethods/optfcts.py
```diff
@@ -182,6 +182,18 @@ def _par_at_boundary(low: np.ndarray,
     return False
 
 
+def _update_reported_nfev(result: OptReturn,
```
hamogu, line 187: Does this really need to be a typed function? How often do we need it? I think essentially any function that has only one line should be inlined for readability. When I read the place where it's used `result[4]['nfev'] += nfev` tells me immediately what happens. Instead, I need to go and hunt for the `_update_reported_nfev` function ("update" doesn't imply +1, could also be some other change, so I actually have to read the function if I want to know what happens), see what that does, and then jump back to what I was reading before. Does not make the code clearer.

sherpa/estmethods/__init__.py
```diff
@@ -84,8 +84,11 @@
 #    "Reached a parameter hard maximum"
 #    pass
 #
-class EstNewMin(Exception):
 
+# This class is used to send around the "new parameter values" in
+# sherpa.fit.Fit.est_errors.
+#
+class EstNewMin(Exception):
```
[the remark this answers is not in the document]
DougBurke: I believe there's a python-version constraint here, but I can never find out what that is. I do think that we are well past that constraint, but I mentinon it for completeness [not watched the video, maybe it's mentioned there].

sherpa/fit.py
```diff
@@ -63,67 +69,72 @@ def run(fit, *args, **kwargs):
...
+                 statname: str,
+                 statval: float,
+                 numpoints: int,
+                 model,  # Why do we have this?
+                 dof: int,
+                 qval: float | None = None,
+                 rstat: float | None = None) -> None:
+        self.name: str = ''
+        """The name of the data set, or sets."""
+
```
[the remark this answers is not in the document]
DougBurke: You can view this on https://sherpa--2022.org.readthedocs.build/en/2022/fit/api/sherpa.fit.StatInfoResults.html as #2022 is built on this PR. I see

sherpa

sherpa/optmethods/opt.py
```diff
@@ -204,7 +211,9 @@ def is_max_length_small_enough():
                 xi = self.simplex[ii, :-1]
                 xi_x0 = xi - x0
                 max_xi_x0 = max(max_xi_x0, np.dot(xi_x0, xi_x0))
-            return max_xi_x0 <= ftol * max(1.0, np.dot(x0, x0))
+
+            # force float conversion to please the type checker
```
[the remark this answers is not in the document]
hamogu: The latter; I just read this yesterday: https://arstechnica.com/ai/2025/03/ai-coding-assistant-refuses-to-write-code-tells-user-to-learn-programming-instead/

But it's OK, it's a just a comment in the code. I apologize for making you waste your time to reply to my comment as if it was a serious point of discussion.

sherpa/estmethods/__init__.py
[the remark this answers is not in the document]
hamogu: I'm not sure what Python version constrained you are referring to and why the version would matter.
At this stage, `EstNewMin` is a class that is one line long and does nothing (it only has a `pass` statement), so I simply wondered if we couldn't just use a generic "Exception" instead of `EstNewMin`. I see that we have a code like
```Python
        except EstNewMin as emin:
            # catch the EstNewMin exception and attach the modified
            # parameter values to the exception obj.  These modified
            # parvals determine the new lower statistic.
            raise EstNewMin(pars) from emin
```

so we DO make use of the fact that this is a specialized kind of exception and thus I see why it's written this way. 

I'm not convinced that passing the new parameter values up the chain as properties of an exception is the cleanest way to do that, but it works now and I don't see any reasons to change working code for that at this point.
So, please disregard my comment.

sherpa/optmethods/optfcts.py
hamogu: It's a design choice I dislike, but reasonable people can disagree with my opinion. I'm not going to hold up this PR over a one-line function.

Personally, if I see `_update_reported_nfev` I will look for that function, find it, try to see what it does, and then I'll still have to think if it has the correct semantics in that function (i.e. does it have at least 5 elements and is the fifth element a dict), as opposed to in the place where it's inlined. Same thought process,  just more jumping. But can be done either way.

hamogu (approved):
@DougBurke responded to all my comments except for a simple spelling mistake that can be rolled in later.

sherpa/utils/types.py
```diff
@@ -46,16 +48,21 @@
...
+StatResults = tuple[float, np.ndarray]
+
 # Represent statistic evaluation.
 #
-# Ideally the callable routines would be labelled as accepting
-# [ArrayType] and return np.ndarray, but we do not enforce this yet
-# (and it is not entirely clear whether the stat functions may accept
-# multiple arguments).
+# There is partial support for sending extra information to the
+# statistics function. At present all that is requires is that the
```
DougBurke, line 58:
```suggestion
# statistics function. At present all that is required is that the
```

sherpa/optmethods/opt.py
```diff
@@ -190,17 +183,19 @@ def check_convergence(self,
...
             smallest_fct_val = self.simplex[0, -1]
             largest_fct_val = self.simplex[-1, -1]
             return Knuth_close(smallest_fct_val, largest_fct_val,
                                ftol)
 
-        def is_fct_stddev_small_enough():
+        def is_fct_stddev_small_enough() -> bool:
             fval_std = np.std([col[-1] for col in self.simplex])
-            return fval_std < ftol
+            # Force float comparison to please type checkers, both for
```
DougBurke, line 194:
```suggestion
            # Force float comparison to avoid type checker warnings, both for
```

sherpa/optmethods/opt.py
```diff
@@ -216,7 +211,9 @@ def is_max_length_small_enough():
                 xi = self.simplex[ii, :-1]
                 xi_x0 = xi - x0
                 max_xi_x0 = max(max_xi_x0, np.dot(xi_x0, xi_x0))
-            return max_xi_x0 <= ftol * max(1.0, np.dot(x0, x0))
+
+            # force float conversion to please the type checker
```
DougBurke, line 215:
```suggestion
            # force float conversion to avoid type checker warnings
```

DougBurke:
For reference I believe I've addressed the typo that @hamogu mentioned above, as well as other issues we discussed during the review but hadn't reached "critical" status.

Issue #2247, opened 2025-03-24
python deprecations in build settings


DougBurke:
From #2246 we can see that `uv` complains about (amongst other things):

```
      /home/user/.cache/uv/builds-v0/.tmp0YpTOk/lib/python3.12/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license`
      as a TOML table is deprecated
      !!

              ********************************************************************************
              Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`.

              By 2026-Feb-18, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.

              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************

```

```
      /home/user/.cache/uv/builds-v0/.tmp0YpTOk/lib/python3.12/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers
      are deprecated.
      !!

              ********************************************************************************
              Please consider removing the following classifiers in favor of a SPDX license expression:

              License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)

              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************
```

Yes, we have just re-worked things to use the "latest" standards for building Python, but the "standards" keep on improving...

PR #2248, opened 2025-03-24
Allow combinatinos of models with arbitrarty ufuncs

hamogu:
## Summary
Models can now be combined using arbitrary ufuncs. Before, it was possible ot say `model1 + model2` to greate a combined model, now any numpy ufunc will work, e.g. `np.greater(model1, model2)`.

## Details
This follows on #1653 where the same implementation was used to allow arbitrary ufuncs to link parameters values. In that PR, we already discussed the possibility of extended that to combinations of models, which is implemented in this PR.

Since any Python function can be turned into a ufunc using numpy's machinary, this allows an arbitrary combination of models in addition to +, -, *, / etc that where implemented before. As a general modelling and fitting application, we leave it to our users to find the right applications, but for astronomy two particualr examples could be lightcurves where the data is magnitudes, while the fitted model predicts fluxes. We now can have a model expression like `ui.set_source(-2.5 * np.log10(flux_model))`; the other application I see right now is greater flexibility to model backgrounds in spectra or images with an empiral model that might combine components with "greater", or "heavyside".


hamogu:
This passes locally with an older version of setup tools, the CI failure is due to #2247, thus I would appreciate a review, even though the there is no green tick mark.

hamogu:
rebased since #2249 has been merged

hamogu:
> does this mess up my lovely "strip out excess brackets from model expressions" code? I'm hoping not, but should check

No. Preserving that turned out to be the most time-consuming part of the PR because I had to understand that `float(3.4)` and `np.float64(3.4)` behave differently, because the latter also has an `__array_ufunc__` method, while the former does not.

>as kind-of mentioned in https://github.com/sherpa/sherpa/discussions/2257, do we want to cache model evaluations of the ufunc (almost certainly not here, and maybe never)

Certainly not here. I'm anticipating use cases like `np.log` (for fitting magnitudes when the model gives fluxes) or `np.maximum` (for two instrumental background components, taking only the larger one in each spectral range). If the ufunc itself takes a long time, you probably want to write a user model anyway and not just slap a ufunc on an existing model. Our cache has enough avenues of development that are far more promising.

docs/evaluation/combine.rst
```diff
@@ -50,6 +50,58 @@ a NumPy vector in the expression, but this is only possible in
 restricted situations, when the grid size is known (i.e. the model
 expression is not going to be used in a general setting).
 
+
+Any ufunc can be used to combine models
+=======================================
```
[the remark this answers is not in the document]
hamogu: While it's numpy that defines ufuncs, there are also examples in spicy, and users can define their own. Saying "any numpy ufunc" would imply that .g. spicy ufuncs are not included when in fact they are.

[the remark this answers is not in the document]
DougBurke: Well then you need to change line 66 which says ```Only the numpy versions work here```. Although it does then talk about SciPy I think the text is a bit confusing on what is or is not supported (which is also true given the section title).

docs/evaluation/combine.rst
```diff
@@ -50,6 +50,58 @@ a NumPy vector in the expression, but this is only possible in
...
+    array([1., 1., 1., 1., 2., 4., 2., 1., 1., 1.])
+
+
+Not every possible link function makes sense
+--------------------------------------------
+
+With this flexibility, it is possible to define links that make no sense,
+for example taking the logical not of a parameter that represents a mass or
+turning values of parameters into arrays (Sherpa optimisers can only deal
+with scalar parameters.) In practice, such mistakes
```
DougBurke, line 98:
```suggestion
with scalar parameters). In practice, such mistakes
```

docs/evaluation/combine.rst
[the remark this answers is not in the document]
hamogu: I just had to look that up, but I think by number scipy actually has more ufuncs in https://docs.scipy.org/doc/scipy/reference/special.html than numpy as in total (not saying that our users are likely to combine two models with an exponentially scaled Hankel function of the second kind very often, but they could).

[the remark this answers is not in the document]
hamogu: Will reword. I wanted to express that e.g. `math.sqrt` won't work, but `np.sqrt` does.

docs/models/index.rst
```diff
@@ -72,7 +72,7 @@ more-complicated cases, such as :doc:`convolution models
 Combining models
 ================
 
-Models can be combined and shared by using the standard Python
+The easiest way to combined and share models is by using the standard Python
```
DougBurke, line 75:
```suggestion
The easiest way to combine and share models is by using the standard Python
```

[I am also not 100% happy with my "shared/share" text but I wanted some way to say you can "re-use" components and "share" was the word I came up with; but that's somewhat irrelevant here]

sherpa/models/model.py
```diff
@@ -1357,13 +1357,25 @@ def guess(self, dep, *args, **kwargs):
...
 
-def _make_unop(op: Callable, opstr: str) -> Callable:
-    def func(self):
-        return UnaryOpModel(self, op, opstr)
+# It's hard to come up with a sensible typing rule for the operator
+# in _make_unop/binop so we just use Callable.
+#
+def _make_unop(op: Callable,
+               opstr: str,
+               strformat: Optional[str] = None) -> Callable:
```
DougBurke, line 1365:
```suggestion
               strformat: str | None = None) -> Callable:
```

[since we have dropped Python 3.9 suport]

DougBurke: or support...

sherpa/models/parameter.py
```diff
@@ -743,8 +753,34 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         if ufunc.nin == 1:
             return UnaryOpParameter(inputs[0], ufunc, name)
         if ufunc.nin == 2:
-            return BinaryOpParameter(inputs[0], inputs[1], ufunc, name,
-                                     strformat='{opstr}({lhs}, {rhs})')
+            if ufunc in self.numpy_binop_with_symbols:
```
DougBurke, line 756: Given that we have something like

```python
name = ...
if boolean:
    name = ...
    strformat = ...
    return BinOpPar(arg0, arg1, arg2, name, strformat=strformat)
else
    strformat = ...
    return BinOpPar(arg0, arg1, arg2, name, strformat=strformat)
```

can we not say

```python
if boolean:
    name = ...
    strformat = ...
else:
    name  = ...
    strformat = ...

return BinOpPar(arg0, arg1, arg2, name, strformat=strformat)
```

sherpa/models/parameter.py
```diff
@@ -721,12 +721,22 @@ def _repr_html_(self) -> str:
     __add__, __radd__ = _make_binop(np.add, '+')
     __sub__, __rsub__ = _make_binop(np.subtract, '-')
     __mul__, __rmul__ = _make_binop(np.multiply, '*')
-    __div__, __rdiv__ = _make_binop(np.divide, '/')
     __floordiv__, __rfloordiv__ = _make_binop(np.floor_divide, '//')
     __truediv__, __rtruediv__ = _make_binop(np.true_divide, '/')
     __mod__, __rmod__ = _make_binop(np.remainder, '%')
     __pow__, __rpow__ = _make_binop(np.power, '**')
 
+    numpy_binop_with_symbols = {
```
DougBurke, line 729: So, you have removed `__div__` and `__rdiv__` but not `__add__` etc. Perhaps it's obvious when I go through the individual commits, but I've not done that yet (thanks to listening in on the XMM workshop)?

sherpa/models/tests/test_model_op.py
```diff
@@ -110,18 +105,18 @@ def test_basic_binop_raw(op):
...
 @pytest.mark.parametrize("op,opstr",
-                         [(np.add, '+'), (np.multiply, '*'),
-                          (np.subtract, '-'), (np.divide, '/'),
-                          (np.floor_divide, '//'), (np.true_divide, '/'),
-                          (np.remainder, '%'), (np.power, '**')])
+                         [(operator.add, '+'), (operator.mul, '*'),
+                          (operator.sub, '-'),
+                          (operator.floordiv, '//'), (operator.truediv, '/'),
+                          (operator.mod, '%'), (operator.pow, '**')],
+                          )
```
DougBurke, line 119: Should we still test the `np.XXX` versions (e.g. add in the `operator.XXX` versions to the tests rather than replace the NumPy ones)?

DougBurke:
I was able to create the very-important models

    a = np.log10(g1)
    b = np.log10(g1 + 2 * g1)
    c = np.log10(3 * g1)

and check you get the expected results (modulo numerical wibbles) - e.g.

    b(x) == c(x)
    b(x) == np.log10(3) + a(x)

[for g being Gauss1D so all values are > 0].

And, if you are "clever" enough, you can cause problems - but this is not something we can really try to catch [*]

```
>>> from sherpa.models.basic import Polynom1D
>>> m = Polynom1D("m")
>>> import numpy as np
>>> x = np.arange(-5, 5)
>>> m(x)
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])
>>> m.c1 = 1
>>> m(x)
array([-4., -3., -2., -1.,  0.,  1.,  2.,  3.,  4.,  5.])
>>> m2 = np.log10(m)
>>> m2(x)
/lagado.real/sherpa/sherpa-main/sherpa/models/model.py:1658: RuntimeWarning: divide by zero encountered in log10
  return self.op(self.arg.calc(p, *args, **kwargs))
/lagado.real/sherpa/sherpa-main/sherpa/models/model.py:1658: RuntimeWarning: invalid value encountered in log10
  return self.op(self.arg.calc(p, *args, **kwargs))
array([       nan,        nan,        nan,        nan,       -inf,
       0.        , 0.30103   , 0.47712125, 0.60205999, 0.69897   ])
```

[*] there is some argument that we could try to handle NaN/Inf values you get back from a model, but that is in a separate part of the code and it's not a high priority for me

sherpa/models/tests/test_model.py
```diff
@@ -541,6 +542,24 @@ def test_composite_iter():
     assert parts[3] is out.m2
     assert len(parts) == 4
 
+    assert m(4) == 10
+
+
+def test_composite_iter_ufunc():
+    """Repeat the test, but call a ufunc instead of using the "+" symbol"""
```
DougBurke, line 549: Which test is being repeated? Over time the tests may get moved around and so it's not necessarily obvious if it's the previous test or not.

DougBurke: [I say this because I have written plenty of comments like this and then it has come back to haunt me ;-]

docs/models/index.rst
```diff
@@ -72,7 +72,9 @@ more-complicated cases, such as :doc:`convolution models
 Combining models
 ================
 
-The easiest way to combined and share models is by using the standard Python
+Models can be combined to serve as building blocs for more complex
```
DougBurke, line 75:
```suggestion
Models can be combined to serve as building blocks for more complex
```

sherpa/models/model.py
```diff
@@ -1362,7 +1362,7 @@ def guess(self, dep, *args, **kwargs):
 #
 def _make_unop(op: Callable,
                opstr: str,
-               strformat: Optional[str] = None) -> Callable:
+               strformat: str | None = None) -> Callable:
```
DougBurke, line 1365: I don't trust codecov, as I've seen pieces of code that I know are tested (e.g. I put an `assert False` in them and the tests fail) but it still claims it is not covered by the tests. Is this one of them?

This is for line 1372 but GitHub won't let me add the comment because apparently "Line must be part of the diff" so ya booh sucks to GH...

image

DougBurke: Hmm. I tried a run with this PR and

```
% git diff .
diff --git a/sherpa/models/model.py b/sherpa/models/model.py
index 94b4d593..2e6243c2 100644
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -1368,6 +1368,7 @@ def _make_unop(op: Callable,
         def func(self):
             return UnaryOpModel(self, op, opstr)
     else:
+        assert False
         def func(self):
             return UnaryOpModel(self, op, opstr, strformat=strformat)
 
```

and the tests passed, so it does look like we don't have a test of this option (but this was with a build without XSPEC, although we should be able to test this without XSPEC models).

DougBurke (approved):
LGTM other than the typo I noted in the documentation.

I think there's still work to be done here but this is a useful improvement as is.

DougBurke:
Oops - I think I meant the "there's still things to be done" comment for #2166 ;-)

hamogu:
I will wait for @wmclaugh to review and see if there are any more typos or other changes to address. (I don't want to waste the cycles to just add a "k" if there is more coming anyway.)

docs/models/index.rst
```diff
@@ -114,7 +116,8 @@ FWHM (full-width half-maximum)
 parameters to be the same but to let the position and amplitude
 values vary independently.
 
-More information is available in the
+More information including more complex, but more powerful ways to combined models
```
wmclaugh, line 119:
```suggestion
More information including more complex, but more powerful ways to combine models
```

sherpa/astro/instrument.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015, 2023, 2024
+#  Copyright (C) 2010, 2015, 2023-2025
 #  Smithsonian Astrophysical Observator
```
wmclaugh, line 3:
```suggestion
#  Smithsonian Astrophysical Observatory
```

sherpa/models/model.py
```diff
@@ -1425,12 +1433,70 @@ def cache_status(self) -> None:
...
+                # with the symbol `*`.
+                # This works where both `a` and `b` are ArithmeticModels or when
+                # e.g. `a` is a Python float and `b` is an ArithmeticModel.
+                # However, that is not the case if `a` some other objects that
+                # does have an `__array_ufunc__` method itself, but raises an
+                # NotImplemented error when called with an ArtihmeticModel.
+                # In that case, the `__array_ufunc__` method of the other object
+                # will be called, and, when it fails, it will fall back to call
+                # the `__array_ufunc__` method of the ArithmeticModel.
+                # This leads to be surprinsing behavior that
```
wmclaugh, line 1486:
```suggestion
                # This leads to be surprising behavior that
```

sherpa/models/model.py
```diff
@@ -1635,24 +1707,40 @@ def __init__(self,
...
         self.op = op
-        self.opstr = opstr
 
         p, a =  get_precedences_op(op)
         self.opprec = p
 
-        # Simplify the expression if possible.
+        # Is this an infix or prefix operator? This could be specified
+        # explicitly (and, in fact, could replace the use of the
+        # strformat argument), but for now the behvaiour is inferred
```
wmclaugh, line 1721:
```suggestion
        # strformat argument), but for now the behaviour is inferred
```

sherpa/models/parameter.py
```diff
@@ -743,8 +753,34 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                # We want to replace certain numpy functions with symbols
+                # to make the expression more readable.
+                # For example, `np.multiply(a, b)` should be displayed as
+                # `a * b`.
+                # In many cases, `a * b` will trigger the __mul__ method
+                # of the ArithmeticModel, which will return a BinaryOpModel
+                # with the symbol `*`.
+                # This works where both `a` and `b` are ArithmeticModels or when
+                # e.g. `a` is a Python float and `b` is an ArithmeticModel.
+                # However, that is not the case if `a` some other objects that
```
wmclaugh, line 766: see comment at 1480 of sherpa/models/model.py

sherpa/models/tests/test_model_op.py
```diff
@@ -130,14 +130,126 @@ def test_basic_binop(op, opstr):
...
+
+@pytest.mark.parametrize("op,mdlname",
+                         [
+                          # Test a random selection of numpy ufuncs
+                          (np.add, "polynom2d + gauss2d"),
+                          (np.multiply, "polynom2d * gauss2d"),
+                          (np.heaviside, "numpy.heaviside(polynom2d, gauss2d)"),
+                          (np.greater, "numpy.greater(polynom2d, gauss2d)"),
+                          (np.arctan2, "numpy.arctan2(polynom2d, gauss2d)"),
+                          # and out own, custom made ufunc
```
wmclaugh, line 153: Do you mean 'our own'?

sherpa/models/model.py
```diff
@@ -1425,12 +1433,70 @@ def cache_status(self) -> None:
...
+                # We want to replace certain numpy functions with symbols
+                # to make the expression more readable.
+                # For example, `np.multiply(a, b)` should be displayed as
+                # `a * b`.
+                # In many cases, `a * b` will trigger the __mul__ method
+                # of the ArithmeticModel, which will return a BinaryOpModel
+                # with the symbol `*`.
+                # This works where both `a` and `b` are ArithmeticModels or when
+                # e.g. `a` is a Python float and `b` is an ArithmeticModel.
+                # However, that is not the case if `a` some other objects that
```
hamogu, line 1480:
```suggestion
                # However, that is not the case if `a` is some other objects that
```

sherpa/models/parameter.py
hamogu, line 766:
```suggestion
                # However, that is not the case if `a` is some other objects that
```

PR #2249, opened 2025-03-24
Update project.toml for PEP 639

### LINKED ISSUE #1452: confusion in setup.cfg over the include/lib paths
I am trying to update https://sherpa.readthedocs.io/en/4.14.0/developer/index.html#install-from-source-in-conda to work for CIAO 4.14. In doing so I noticed that we do not have a consistent naming scheme for the lib and include paths. We have some that use the `xxx-dirs` name:

```
% grep -- -dirs setup.cfg
#fftw-lib-dirs=build/lib
#region-lib-dirs=build/lib
#wcs-include-dirs=build/include
#wcs-lib-dirs=build/lib
```

and most that use the `xxx_dirs` name:

```
% grep -- _dirs setup.cfg
#fftw-include_dirs=build/include
#region-include_dirs=build/include
# The xspec_include_dirs and xspec_lib_dirs items should be set
# If you are using the models-only XSPEC build, then the cfitsio_lib_dirs,
# ccfits_lib_dirs, and wcslib_lib_dirs will need to be set if
# The gfortran_lib_dirs should be set if needed.
#xspec_lib_dirs = None
#xspec_include_dirs = None
#cfitsio_lib_dirs = None
#ccfits_lib_dirs = None
#wcslib_lib_dirs = None
#gfortran_lib_dirs = None
```

I noticed this because the wcs include files were not being picked up and changing it to `wcs-incluude_dirs` seemed to do the trick.

However, looking at `helpers/sherpa_config.py` just confuses the matter for me, as the config object uses the `-dirs` form - https://github.com/sherpa/sherpa/blob/fad2a88159ea148c1b3d4eb6358613cd83e5900c/helpers/sherpa_config.py#L30-L52 rather than the `_dirs` form, but maybe the config object automatically switches from `-`-separated to `_`-separated when creating the fieds of the object? 

My guess is the `setup.cfg` lines that say `-dirs` should shuold be changed to `_dirs` but I don't really have time for investigating this.

### LINKED ISSUE #2247: python deprecations in build settings

hamogu:
## Summary
Address setuptools deprecation of lower-kebab-case for names of options

## Details
 no more details.

fixes #2247
fixes #1452

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -261,4 +261,4 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: ${{ github.workspace }}/coverage.xml
         verbose: true
-      
+
--- a/helpers/sherpa_config.py
+++ b/helpers/sherpa_config.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2014 - 2016, 2020, 2022, 2024
+#  Copyright (C) 2014-2016, 2020, 2022, 2024-2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -33,25 +33,25 @@ class sherpa_config(Command):
     description = "Configure Sherpa build options. If in doubt, ignore this command and stick to defaults. See setup.cfg for more information."
     user_options = [
                     ('fftw', None, "Whether Sherpa should build the embedded fftw3 library, which is the default behavior: set to 'local' to make Sherpa link against existing libraries on the system.)"),
-                    ('fftw-include-dirs', None, "Where the fftw3 headers are located, if fftw is 'local'"),
-                    ('fftw-lib-dirs', None, "Where the fftw3 libraries are located, if fftw is 'local'"),
-                    ('fftw-libraries', None, "Name of the libraries that should be linked as fftw3"),
+                    ('fftw_include_dirs', None, "Where the fftw3 headers are located, if fftw is 'local'"),
+                    ('fftw_lib_dirs', None, "Where the fftw3 libraries are located, if fftw is 'local'"),
+                    ('fftw_libraries', None, "Name of the libraries that should be linked as fftw3"),
                     ('region', None, "Whether Sherpa should build the embedded region library, which is the default behavior: set to 'local' to make Sherpa link against existing libraries on the system.)"),
-                    ('region-include-dirs', None, "Where the region headers are located, if region is 'local'"),
-                    ('region-lib-dirs', None, "Where the region libraries are located, if region is 'local'"),
-                    ('region-libraries', None, "Name of the libraries that should be linked as region"),
-                    ('region-use-cxc-parser', None, "If set to True than use the CXC Data Model library to parse region files"),
+                    ('region_include_dirs', None, "Where the region headers are located, if region is 'local'"),
+                    ('region_lib_dirs', None, "Where the region libraries are located, if region is 'local'"),
+                    ('region_libraries', None, "Name of the libraries that should be linked as region"),
+                    ('region_use_cxc_parser', None, "If set to True than use the CXC Data Model library to parse region files"),
                     ('wcs', None, "Whether Sherpa should build the embedded wcs library, which is the default behavior: set to 'local' to make Sherpa link against existing libraries on the system.)"),
-                    ('wcs-include-dirs', None, "Where the wcs subroutines headers are located"),
-                    ('wcs-lib-dirs', None, "Where the wcs subroutines libraries are located"),
-                    ('wcs-libraries', None, "Name of the libraries that should be linked as wcs"),
-                    ('group-location', None, "Location of the group  python module"),
-                    ('disable-group', None, "Disable the group module install"),
-                    ('stk-location', None, "Location of the stack library python module"),
-                    ('disable-stk', None, "Disable the stack library module install"),
-                    ('install-dir', None, "Directory where external dependencies must be installed (--prefix)"),
+                    ('wcs_include_dirs', None, "Where the wcs subroutines headers are located"),
+                    ('wcs_lib_dirs', None, "Where the wcs subroutines libraries are located"),
+                    ('wcs_libraries', None, "Name of the libraries that should be linked as wcs"),
+                    ('group_location', None, "Location of the group  python module"),
+                    ('disable_group', None, "Disable the group module install"),
+                    ('stk_location', None, "Location of the stack library python module"),
+                    ('disable_stk', None, "Disable the stack library module install"),
+                    ('install_dir', None, "Directory where external dependencies must be installed (--prefix)"),
                     ('configure', None, "Additional configure flags for the external dependencies"),
-                    ('group-cflags', None, "Additional cflags for building the grouping library"),
+                    ('group_cflags', None, "Additional cflags for building the grouping library"),
                     ]
 
     def initialize_options(self):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ description = "Modeling and fitting package for scientific data analysis"
 readme = {file = "README.md", content-type = "text/markdown"}
 license = {file = "LICENSE"}
 
+
 authors = [
   { name = "Smithsonian Astrophysical Observatory / Chandra X-Ray Center", email = "cxchelp@cfa.harvard.edu" }
 ]
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,15 +8,15 @@
 #configure=None
 
 # GROUP Python module
-#disable-group=True
-#group-location=build/lib/python2.7/site-packages/group.so
+#disable_group=True
+#group_location=build/lib/python2.7/site-packages/group.so
 
 # Additional flags to be passed to libgrp build
-#group-cflags=
+#group_cflags=
 
 # File Stack Python module
-#disable-stk=True
-#stk-location=build/lib/python2.7/site-packages/stk.so
+#disable_stk=True
+#stk_location=build/lib/python2.7/site-packages/stk.so
 
 # FFTW Library
 # Uncomment to use a local installation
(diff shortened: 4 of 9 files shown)


hamogu (commented):
Alternatively, we could pin setuptools to < 77.0.0 or find some other way to not fail the CI

pyproject.toml
```diff
@@ -12,15 +12,15 @@ dynamic = ["version"]
...
+license = "GPL-3.0-or-later"
+license-files = ["LICENSE", "extern/fftw-3.3.10/COPYING"]
+
 
 authors = [
   { name = "Smithsonian Astrophysical Observatory / Chandra X-Ray Center", email = "cxchelp@cfa.harvard.edu" }
 ]
 
 classifiers = [
-  "Intended Audience :: Science/Research",
```
DougBurke, line 22: Why did you remove this classifier?

hamogu: Ah, no.

DougBurke:
Looking at the CI build failure, I wonder if there's some other changes in setuptools that are causing problems (in this case how the XSPEC config settings are set [*]). We should probably discuss whether we want to go to meson (e.g. #1948) or some other build backend.

[*] or it's just a blip in building Sherpa ...

hamogu:
But something else must have changed in the setup (or the more recent version of setup tools that the CI is apparently picking up)
```
  error: error in setup.cfg: command 'xspec_config' has no such option 'with-xspec'
  error: subprocess-exited-with-error
```

That is not related to the license file, but that's something else.

hamogu:
In the near term, we could also pin a max version of setuptools instead. That would allow us to continue developing other PRs while we figure this out. Going to mason isn't going to happen this week. (Even if the PR is ready, it will need more detailed investigation by DS, I believe).

Or maybe there is an easy solution of the `command 'xspec_config' has no such option 'with-xspec'` error.

setup.cfg
```diff
@@ -65,8 +65,8 @@ source-dir = docs
 
 # XSPEC Models
 [xspec_config]
-# Uncomment (set to True) to build XSPEC extension
-#with-xspec=True
+# Set to True to build XSPEC extension
+with-xspec=False
```
DougBurke, line 69: So, we have `helpers/xspec_config.py` which handles this section, and in it it calls `initialize_options` which is meant to set its `with_xspec` attribute False, but maybe the `user_options` attribute needs updating?

I have never found the documentation to be clear on what is meant to happen in these classes with configuration and set up ... It may be an issue somewhere completely different!

https://github.com/sherpa/sherpa/blob/80ad0889c1df31c943e8a5c7f6f7ee7274486598/helpers/xspec_config.py#L79-L141

hamogu: I'm just experimenting. I have no idea why is stopped working, so I thought "Let's try if they suddenly added a requirement to set the defaults in the setup.cfg file". Could have been, but it's not the case. I figured it faster to try though than to find the docs for that. (and faster on CI then to get matching versions of setuptools etc locally, if I on'y need to try once. That's not the case....)

hamogu: So, not I wonder if the translation `-` -> `_` has changed. If it's not that, then it's something deeper that I won't try to understand and hope somebody else does.

hamogu: Jackpot: After renaming `with-xspec` it goes past that line and now fails on `error: error in setup.cfg: command 'sherpa_config' has no such option 'fftw-include_dirs'`.
It'll be annoying to rename all options in all places (and why on earth is that required now?) but at least it has minimal user impact.

hamogu: > I'll just point to https://github.com/sherpa/sherpa/issues/1452 when I noted some confusion on my part. This has some fall-out as things like scripts/use_ciao_config may need updating, as would the documentation.

That's what "search and replace" is for. ;-)

DougBurke:
And they've reverted the setuptools change: https://setuptools.pypa.io/en/stable/history.html#v78-0-2

hamogu:
It's still deprecated though, "All packages with deprecated configurations are advised to move before 2026." That's essentially right after the 4.18 release, meaning that for 4.18 we either take this change or risk people running into trouble if they accidentally get a newer version of setuptools in 2026. (Or we just rely on them to blow past that date, too.)

DougBurke:
There's also the fact that even with 78.0.2 our build still appears to fail because of (at least in my PR, which does not contain any change from this PR)

```
error: error in setup.cfg: command 'sherpa_config' has no such option 'fftw-include_dirs'
```

All I meant about the reverted change was that it would hopefully give us time to fix it without haste, but apparently not.

DougBurke:
I hacked up https://github.com/sherpa/sherpa/pull/2022/commits/db77f74a46eb2d6e9ac811bc75820769e6ff79c5 to get one of my PRs going.

.github/workflows/ci-conda-workflow.yml
```diff
@@ -38,7 +38,7 @@ jobs:
             test-data: submodule
             matplotlib-version: 3
             bokeh-version: 3
-            xspec-version: 12.14.0i
+            xspec_version: 12.14.0i
```
DougBurke, line 41: There's no reason for changing this (note that `bokeh-version` and `matplotlib-version` above have not been changed).

helpers/sherpa_config.py
```diff
@@ -33,25 +33,25 @@ class sherpa_config(Command):
     description = "Configure Sherpa build options. If in doubt, ignore this command and stick to defaults. See setup.cfg for more information."
     user_options = [
                     ('fftw', None, "Whether Sherpa should build the embedded fftw3 library, which is the default behavior: set to 'local' to make Sherpa link against existing libraries on the system.)"),
-                    ('fftw-include-dirs', None, "Where the fftw3 headers are located, if fftw is 'local'"),
-                    ('fftw-lib-dirs', None, "Where the fftw3 libraries are located, if fftw is 'local'"),
-                    ('fftw-libraries', None, "Name of the libraries that should be linked as fftw3"),
+                    ('fftw_include_dirs', None, "Where the fftw3 headers are located, if fftw is 'local'"),
```
DougBurke, line 36: So, it appears that this change is only an issue for the `setup.cfg` file, and not here, so it is technicaly not necessary, **but**

- I can't find any obvious documentation in `setuptools` that talks about this change (the best I saw was "go and read about it at https://setuptools.pypa.io/en/latest/deprecated/index.html" which at the moment has nothing obvious about this change
- I have been confused about this before - e.g. #1452 - so making all things consistent is probably a good idea

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools >= 64.0.0",  # minimum for PEP 660 editable install
+  "setuptools >= 77.0.0",  # minimum for PEP 639 which changes the license file format in project.toml
```
DougBurke, line 3: Given that version 77.0.0 was only released on March 19 2025 I think this is perhaps too much of a change. I would *hope* that we can use earlier versions even with these changes (perhaps leaving in the license classifier below so that users of older setuptools do not lose information)?

Although maybe the older versions will fall over with the `license` setting? I am too tired / annoyed about all this to go and look.

helpers/sherpa_config.py
hamogu: Yes, I don't think it's required, but I decided that - if we do changes already - then that's a good opportunity to be more consistent. The Inconsistency has caused confusion in the past.

.github/workflows/ci-conda-workflow.yml
hamogu: too much replace all. Will fix.

pyproject.toml
hamogu: I'm going to experiment what the minimum version is, now that I've identified all that we need to do to be compatible with the *newest* version. Expect to see a few CI runs, because that way I'll see that it works for linux, too, and for all the different setups that we use.

hamogu: The change in the license format done here was introduced in 77.0 and is not compatible with previous versions. I'll need to revert that and pin to "<77.0.0" and then we can have a follow-up at some later point and change that.

hamogu:
I reversed the changes to the license and license file keys, which were driving that "setuptools >= 77.0.0" requirement.

DougBurke:
Can we squash these (just so that the history is not so noisy)?

hamogu:
squashed.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools >= 64.0.0",  # minimum for PEP 660 editable install
+  "setuptools == 77.0.1",  # minimum for PEP 660 editable installs
```
DougBurke, line 3: Did you still mean to bump the minimum required version?

hamogu: No, that's an error in a rebase. Will fix. Thanks for catching and sorry that this sloppily slipped through. Pushed fix.

DougBurke: No probs.

DougBurke (approved):
LGTM (assuming the CI runs are green). We will eventually have to deal with the setuptools version 78 changes, but not now!

Thanks for dealing with this as my patience for changes in setuptools is currently sitting somewhere on the complex plane!

hamogu:
updated summary. No more details needed.

Issue #1651, opened 2022-12-13
How can I pass meta-information to a model or fit?


hamogu:
# don't read this issue. Yet.
I know the text below does not make any sense, but it's such a long text, I need to submit the issue so it's not all lost if my browser does stupid stuff. I will edit and make this a useful issue later.

## Summary
Sometimes datasets have metadata associated with it that sets how a model is evaluated and I'm not aware of a good way to accommodate that in Sherpa. There are several ways around that, but in this issue I want to discuss if that's the way we want to going forward, if we can make any of the workarounds that are possible today so easy that it's useful, or bless on of those ways as "canoncial" and write documentation. 

## Motivation
My examples below are all drawn from X-ray spectroscopy, one of sherpa's core use cases, but that's just because I'm working in that field. I could come up with other example that don't use PHA data, if that's useful. 

Often, we use Sherpa to fit just one spectrum of one object, using some combination of models, e.g. an absorbed powerlaw. However, in many cases, we have a set of related spectra:

- a time series
- IXPE: Spectra with different Stokes values (I, Q, and U)
- spatially ordered spectra, e.g. annuli around some center

Sometimes, even a single spectrum has additional information that's not part of the spectrum itself, in my case it was "distance between star and jet front".

Those information is not visible in the PHA data, it's just meta information on the header. How would I set up a model that makes uses of that information, e.g. a spectral model that depends on time? Let's say I want to fit a powerlaw to a flare decay, something like an APEC model, with "apec.norm = function(time)" where function is "A_0 * exp(-t)". That "t" is not in one of the channels in the PHA data, it's in the header. 

## Workarounds
I've done work like this a couple of times and comes up with several approaches, all of which are clunky and ugly.

### Make, many, may models
(Code is conceptual and won't run as written below.)

```Python

```

`datastack` can help under the hood with the model generation.

### Make a model with a sentinel value and add a second dataset
and put the meta information (e.g. the observation time) in the ya value of the second dataset
```Python
d2 = Data1D(x=[123456], y=datapha.meta['OBSTIME'])

class Recol(models.ArithmeticModel):
    ...
    def predicted_spectrum(self, jetspeed, density, launchingradius):
        return VAPEC(kT=jetspeet**2 / XYZ, norm=launchingradius*jetspeed/ abc, ...)
    def calc(self, pars, *args):
        if args[0][0] == 123456:
            if pars[-1] == args[1][0] # encode the OBSTIME 
        else:
[...]

Issue #1406, opened 2022-01-13
Is our API to define a statistic sufficient for IXPE?


hamogu:
In IXPE (and other upcoming X-ray polarization missions), you generally extract three spectra (stokes I, Q, and U) from the same photons list. That is, every detected photon turns up in the I spectrum and some subset of them (depending on the measured polarization angle) also in Q and U. That means that I, Q, and U spectra are not statistically independent. If they were, the fit statistic would just be the sum of the chi^2 of I, for Q, and for U and that's what Sherpa's chi^2 statistic would do when I load the three spectra independently. 
However, in this case, they are not independent. So, the chi^s needs to be modified. The full math is in [Kislat et al. (2015)](https://ui.adsabs.harvard.edu/abs/2015APh....68...45K/abstract) eqn 49, but essentially instead of 

chi^2 = (data - model)^2 / sigma^2

it now includes terms like 

 sigma_Q^2 + sigma_U^2 - covariance(Q, U) 

where Q and U are PHA spectra read from independent files. That means that the function that calculates the statistic needs to get both Q and U passed in at the same time. That's already happening in https://github.com/sherpa/sherpa/blob/main/sherpa/stats/__init__.py#L79 but it's not particularly obvious how to make use of that. I'll investigate if I can exploit this for use with IXPE, but this is certainly not an API that makes it easy to define the statistic we are looking for, nor is there great documentation for a user to discover this fact. A hack to make this work for now might be to rely on secondary information (e.g. the name of the dataset or the order)

To make this general, Sherpa's fitting also needs a general way to get meta-information of spectra and/or a way to receive bundled spectra (e.g. through datastack) to know which files are related and to couple models for Q and U in an appropriate way. So, this issue is necessary, but not sufficient for IXPE data fitting.

Related: #474

Issue #844, opened 2020-07-02
Make background information available to statistics calculation


hamogu:
In many cases, the value of a fit statistic is calculated from the data and the model, e.g. in chi^2 it's (data - model)**2.
However, there are some cases where the value of a statistic also depends on the background. This has been popping up in a number of cases, so it seems worthwhile now to design a proper API.

Use cases and more information can be found in:

- the implementation of wstat
- https://github.com/sherpa/sherpa/issues/356
- https://github.com/sherpa/sherpa/issues/824

This issue is here to design a more general API than the in-promptu solution in e.g. https://github.com/sherpa/sherpa/issues/824 .

PR #2259, opened 2025-03-31
Add typing statements to the optimization code

DougBurke:
# Summary

Add typing statements to the optimization code. As part of this some of the classes and routines have seen changes in behaviour: use of keyword-only arguments, dropping the incomplete support for `statargs` and `statkwargs`, and removal of some optimization code not used by Sherpa. This is only visible to users calling the optimization routines directly.

# Details

This is taken from #2022. The aim here was not to change the code, just add typing statements, but as discussed below that did not quite happen. However, these are all low-level code changes that should not, hopefully, affect users. I have been using `mypy` and `pyright` to assess the code, and you can see certain areas where they agree and some where they do not: this does complicate deciding "what's the best thing to do" ...

In order to be able to change the code I felt I needed to understand the code, which meant adding typing rules to try to clarify things. The optimization code takes advantage of "callbacks" - or local lambda functions - to pass around functions and convert them from one thing to another. So the main aim here is identifying what the function signatures / types are. As part of doing this I have "removed" the `statargs/statkwargs` code - well, left the arguments in but they now default to `None` and if set create a warning log message - because

- the code was incomplete (e.g. the estimation error code never did anything with the arguments)
- given that we are sending around "callbacks", users can create their own to convert their stats function to match the existing interface
- by removing things we can simplify the types (we don't need to worry about extra arguments).

As the commits go by you can see some of my thinking change - e.g. 

- does this take an `ArrayType`, which is a sequence or `ndarray`, or can we specialize it to just `ndarray`
- can a value be None (sometimes it can but tracking when this changes is awkward and I don't guarantee that I get all of them)
- do I want to bother with supporting NumPy float values (e.g. mark a type as `SupportsFloat`) or shall I just say `float` and use `float(x)` in a few places to reduce the number of warnings from the type checkers
- can a value be a scalar or a list (the `***simplex` argument is a good example of this since it has support for being sent a list but generally seems to only use a scalar, but the C++ gets sent a list if memory serves me correctly)

so this is often "best guess". There is at least one place where we start with a function that can take multiple arguments and endup saying "it can only take a single argument" because, as far as I can make out, that's all we ever send in, so it's much clearer to finally be able to specify it. 

There are definitely things that can be done to improve this code, but I think this is a good "chunk" of changes to review, merge, and review downstream changes. And I do have a bunch of changes in #2022 that extend this PR, so I don't really want to be changing things too much here if we can help it.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -793,8 +793,8 @@ class IterFit:
                  pars: ArrayType,
                  parmins: ArrayType,
                  parmaxes: ArrayType,
-                 statargs: Sequence[Any] = (),
-                 statkwargs: Mapping[str, Any] | None = None,
+                 statargs: Any = None,
+                 statkwargs: Any = None,
                  cache: bool = True
                  ) -> OptReturn:
         """Exclude points that are significately far away from the best fit.
@@ -806,6 +806,9 @@ class IterFit:
         has converged or a maximum number of iterations has been reached.
         The error removal can be asymmetric, since there are separate
         options for the lower and upper limits.
+
+        .. versionchanged:: 4.17.1
+           The statargs and statkwargs arguments are now ignored.
 
         Raises
         ------
@@ -830,8 +833,9 @@ class IterFit:
         ========  ==========  ===========
 
         """
-        if statkwargs is None:
-            statkwargs = {}
+
+        if statargs is not None or statkwargs is not None:
+            warning("statargs/kwargs set but values unused")
 
         # Sigma-rejection can only be used with chi-squared;
         # raise exception if it is attempted with least-squares,
@@ -908,9 +912,9 @@ class IterFit:
                     self.stat.calc_staterror)
                 self.model.startup(cache)
                 final_fit_results = self.method.fit(statfunc,
-                                                    self.model.thawedpars,
-                                                    parmins, parmaxes,
-                                                    statargs, statkwargs)
+                                                    pars=self.model.thawedpars,
+                                                    parmins=parmins,
+                                                    parmaxes=parmaxes)
                 model_iterator = iter(self.model())
                 rejected = False
 
@@ -1006,19 +1010,25 @@ class IterFit:
             pars: ArrayType,
             parmins: ArrayType,
             parmaxes: ArrayType,
-            statargs: Sequence[Any] = (),
-            statkwargs: Mapping[str, Any] | None = None,
+            statargs: Any = None,
+            statkwargs: Any = None
             ) -> OptReturn:
+        """
 
-        if statkwargs is None:
-            statkwargs = {}
+        .. versionchanged:: 4.17.1
+           The statargs and statkwargs arguments are now ignored.
+
+        """
+
+        if statargs is not None or statkwargs is not None:
+            warning("statargs/kwargs set but values unused")
 
         if self.current_func is None:
-            return self.method.fit(statfunc, pars, parmins, parmaxes,
-                                   statargs, statkwargs)
+            return self.method.fit(statfunc, pars=pars,
+                                   parmins=parmins, parmaxes=parmaxes)
 
-        return self.current_func(statfunc, pars, parmins, parmaxes,
-                                 statargs, statkwargs)
+        return self.current_func(statfunc, pars=pars, parmins=parmins,
+                                 parmaxes=parmaxes)
 
 
 # What is the best way to annotate the return value here?
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -807,7 +807,7 @@ class IterFit:
         The error removal can be asymmetric, since there are separate
         options for the lower and upper limits.
 
-        .. versionchanged:: 4.17.1
+        .. versionchanged:: 4.18.0
            The statargs and statkwargs arguments are now ignored.
 
         Raises
@@ -1015,7 +1015,7 @@ class IterFit:
             ) -> OptReturn:
         """
 
-        .. versionchanged:: 4.17.1
+        .. versionchanged:: 4.18.0
            The statargs and statkwargs arguments are now ignored.
 
         """
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -39,14 +39,19 @@ reflect      = True
 
 A model is fit by providing the ``fit`` method a callback, the starting
 point (parameter values), and parameter ranges. The callback should
-match::
+match the `StatFunc` type::
 
-    callback(pars, *statargs, **statkwargs)
+    callback(pars)
 
 and return the statistic value to minimise along with the per-bin
 statistic values (note that per-bin here refers to the independent
 axis of the data being fit, and not the number of parameters being
 fit).
+
+.. versionchanged:: 4.17.1
+   The callback is no-longer sent extra information (that is,
+   the statargs and statkwargs values) and is just sent an array
+   of parameter values.
 
 Notes
 -----
@@ -112,7 +117,6 @@ const1d
 
 """
 
-from collections.abc import Sequence
 import logging
 from typing import Any
 
@@ -166,9 +170,7 @@ class OptMethod(NoNewAttributesAfterInit):
     arrays.
 
     The function returns the statistic value and per-bin statistic
-    values for the given values - which are the first argument it is
-    sent, and then it can take other arguments which are set by the
-    statargs and statkwargs arguments sent to the `fit` call.
+    values for the given values.
 
     """
 
@@ -239,10 +241,13 @@ class OptMethod(NoNewAttributesAfterInit):
             pars: ArrayType,
             parmins: ArrayType,
             parmaxes: ArrayType,
-            statargs: Sequence[Any] = (),
-            statkwargs: dict[str, Any] | None = None
+            statargs: Any = None,
+            statkwargs: Any = None
             ) -> OptReturn:
         """Run the optimiser.
+
+        .. versionchanged:: 4.17.1
+           The statargs and statkwargs arguments are now ignored.
 
         .. versionchanged:: 4.16.0
            The statkwargs argument now defaults to None rather than {}.
@@ -263,9 +268,9 @@ class OptMethod(NoNewAttributesAfterInit):
            The maximum allowed values for each model parameter. This
            must match the length of `pars`.
         statargs : optional
-           Additional positional arguments to send to `statfunc`.
-        statkwargs : dict, optional
-           Additional keyword arguments to send to `statfunc`.
+           This is currently unused.
+        statkwargs : optional
+           This is currently unused.
 
         Returns
         -------
@@ -279,16 +284,11 @@ class OptMethod(NoNewAttributesAfterInit):
 
         """
 
-        # Mid 2024 we do not have any tests where either statargs or
-        # statkwargs are not empty.
-        #
-        if statkwargs is None:
-            statkwargs = {}
+        if statargs is not None and statkwargs is not None:
+            warning("statargs/kwargs set but values unused")
 
-        def cb(pars):
-            return statfunc(pars, *statargs, **statkwargs)
-
-        output = self._optfunc(cb, pars, parmins, parmaxes, **self.config)
+        output = self._optfunc(statfunc, pars, parmins, parmaxes,
+                               **self.config)
         (success, pars, fval, msg, imsg) = output
         if not success:
             warning('fit failed: %s', msg)
--- a/sherpa/optmethods/__init__.py
+++ b/sherpa/optmethods/__init__.py
@@ -48,7 +48,7 @@ statistic values (note that per-bin here refers to the independent
 axis of the data being fit, and not the number of parameters being
 fit).
 
-.. versionchanged:: 4.17.1
+.. versionchanged:: 4.18.0
    The callback is no-longer sent extra information (that is,
    the statargs and statkwargs values) and is just sent an array
    of parameter values.
@@ -246,7 +246,7 @@ class OptMethod(NoNewAttributesAfterInit):
             ) -> OptReturn:
         """Run the optimiser.
 
-        .. versionchanged:: 4.17.1
+        .. versionchanged:: 4.18.0
            The statargs and statkwargs arguments are now ignored.
 
         .. versionchanged:: 4.16.0
@@ -284,7 +284,7 @@ class OptMethod(NoNewAttributesAfterInit):
 
         """
 
-        if statargs is not None and statkwargs is not None:
+        if statargs is not None or statkwargs is not None:
             warning("statargs/kwargs set but values unused")
 
         output = self._optfunc(statfunc, pars, parmins, parmaxes,
--- a/sherpa/optmethods/ncoresde.py
+++ b/sherpa/optmethods/ncoresde.py
@@ -237,6 +237,13 @@ class Strategy9(Strategy):
 
 
 class MyDifEvo(Opt):
+    """
+
+    .. versionchanged:: 4.17.1
+       Calling the object now requires named arguments: maxnfev and
+       ftol.
+
+    """
 
     def __init__(self, func, xpar, xmin, xmax, npop, sfactor, xprob, step,
                  seed, rng=None):
@@ -261,14 +268,18 @@ class MyDifEvo(Opt):
                                  rng=np.random.default_rng(sseed))
                            for strat, sseed in zip(strats, sseeds)]
 
-        xpar = np.asarray(xpar)
+        xpar_np = np.asarray(xpar)
+        xmin_np = np.asarray(xmin)
+        xmax_np = np.asarray(xmax)
         if step is None:
-            step = xpar * 1.2 + 1.2
+            step = xpar_np * 1.2 + 1.2
         factor = 10
         self.rng = rng
-        self.polytope = SimplexRandom(func=func, npop=npop, xpar=xpar,
-                                      xmin=xmin, xmax=xmax, step=step,
-                                      seed=seed, factor=factor, rng=rng)
+        self.polytope = SimplexRandom(func=func, npop=npop,
+                                      xpar=xpar_np, xmin=xmin_np,
+                                      xmax=xmax_np, step=step,
+                                      seed=seed, factor=factor,
+                                      rng=rng)
         self.local_opt = self.ncores_nm.algo
 
     # Only used by DifEvo which is currently unused.
@@ -347,8 +358,15 @@ class MyDifEvo(Opt):
 
 
 class ncoresMyDifEvo(MyDifEvo):
+    """
 
-    def __call__(self, tol, maxnfev, numcores=ncpus):
+    .. versionchanged:: 4.17.1
+       Calling the object now requires named arguments and the
+       ordering has been changed to match the superclass.
+
+    """
+
+    def __call__(self, *, maxnfev, ftol, numcores=ncpus):
         nfev = 0
 
         # Set the seed if RNG is not sent in. This used to change
@@ -375,7 +393,7 @@ class ncoresMyDifEvo(MyDifEvo):
                     mypop[index] = result[1:]
 
             self.polytope.sort()
-            if self.polytope.check_convergence(tol, 0):
+            if self.polytope.check_convergence(ftol, 0):
                 break
 
             best = mypop[0]
@@ -383,8 +401,8 @@ class ncoresMyDifEvo(MyDifEvo):
             if best_fval < old_fval:
                 best_par = best[:-1]
                 tmp_nfev, tmp_fval, tmp_par = \
-                    self.ncores_nm(self.func, best_par, self.xmin, self.xmax,
-                                   tol)
+                    self.ncores_nm(self.func, best_par, self.xmin,
+                                   self.xmax, tol=ftol)
                 nfev += tmp_nfev
                 if tmp_fval < best_fval:
                     best_par = np.append(tmp_par, tmp_fval)
@@ -449,7 +467,7 @@ class ncoresDifEvo:
 
         mydifevo = ncoresMyDifEvo(fcn, x, xmin, xmax, npop, sfactor, xprob,
                                   step, seed, rng=self.rng)
-        return mydifevo(tol, maxnfev, numcores)
+        return mydifevo(ftol=tol, maxnfev=maxnfev, numcores=numcores)
 
 
 # This is only used by tests/test_opt_original.py when run directly,
--- a/sherpa/optmethods/ncoresde.py
+++ b/sherpa/optmethods/ncoresde.py
@@ -433,13 +433,11 @@ class ncoresMyDifEvo(MyDifEvo):
     """
 
     .. versionchanged:: 4.17.1
-       Calling the object now requires named arguments and the
-       ordering has been changed to match the superclass.
+       The calling convention has been changed to match its superclass.
 
     """
 
     def __call__(self,
-                 *,
                  maxnfev: int,
                  ftol: float,
                  numcores: int | None = ncpus
--- a/sherpa/optmethods/ncoresde.py
+++ b/sherpa/optmethods/ncoresde.py
@@ -77,6 +77,37 @@ class Strategy:
                  pop: SimplexRandom,
                  icurrent: int
                  ) -> MyOptOutput:
+
+        # The subclasses are assumed to follow:
+        #
+        #   r1, ..., rn = self.init(n)
+        #   trial = pop[icurrent].copy()
+        #   n = self.select_par()
+        #   for counter in self.npar():
+        #
+        # and then either
+        #
+        #       trial[n] += ...
+        #       n = (n + 1) % self.npar
+        #       if self.check_end():
+        #           break
+        #
+        # or
+        #
+        #       if self.check_select(counter):
+        #           trial[n] += ...
+        #           n = (n + 1) % self.npar
+        #
+        # and then ending up with
+        #
+        #  return self.calc(trial, pop)
+        #
+        # Some of the logic in "..." above could be moved out of the
+        # loop (as it can be calculated for the full parameter array
+        # once rather than per element), but this was found to subtly
+        # change some results - presumably because of numerical
+        # instability - and so the code has been left as is for now.
+        #
         raise NotImplementedError
 
     # The arg argument contains parameter values and then the
@@ -94,6 +125,24 @@ class Strategy:
     def init(self, num: int) -> np.ndarray:
         return random.choice(self.rng, range(self.npop), num)
 
+    def select_par(self) -> int:
+        """Select a parameter."""
+
+        return random.integers(self.rng, self.npar)
+
+    # Not all sub-classes are expected to use this.
+    def check_end(self) -> bool:
+        """Should the iteration exit?"""
+
+        return random.random(self.rng) > self.xprob
+
+    # Not all sub-classes are expected to use this.
+    def check_select(self, counter: int) -> bool:
+        """Should the iteration be evaluated?"""
+
+        return random.random(self.rng) < self.xprob or \
+            counter == self.npar - 1
+
 
 class Strategy0(Strategy):
 
@@ -105,11 +154,11 @@ class Strategy0(Strategy):
         # code has been tested using this call.
         _, r2, r3 = self.init(3)
         trial = pop[icurrent].copy()
-        n = random.integers(self.rng, self.npar)
+        n = self.select_par()
         for _ in range(self.npar):
             trial[n] = pop[0][n] + self.sfactor * (pop[r2][n] - pop[r3][n])
             n = (n + 1) % self.npar
-            if random.random(self.rng) > self.xprob:
+            if self.check_end():
                 break
 
         return self.calc(trial, pop)
@@ -125,11 +174,11 @@ class Strategy1(Strategy):
         # code has been tested using this call.
         _, r2, r3 = self.init(3)
         trial = pop[icurrent].copy()
-        n = random.integers(self.rng, self.npar)
+        n = self.select_par()
         for _ in range(self.npar):
             trial[n] = trial[n] + self.sfactor * (pop[r2][n] - pop[r3][n])
             n = (n + 1) % self.npar
-            if random.random(self.rng) > self.xprob:
+            if self.check_end():
                 break
 
         return self.calc(trial, pop)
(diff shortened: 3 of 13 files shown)


DougBurke:
Currently doing some work in #2022 that is likely to impact this PR.

hamogu:
We have several issues (list as far as I can find them below for cross-reference) that are about meta data that a model or a statistic might need beyond the pure numpy arrays of the independent and dependent variables. We do not have a well-designed set or requirements (thus the number of issues), but they all have in common that "more information" needs t o be passed around, e.g.:
- which polarization direction a spectrum represents (#1406, #1651)
- at what time data was taken or what radial distance from the source is was extracted (can't find number right now)
- the background (#844)

similar #1649.

There might be others. Why do I bring that up? I wonder if `statargs/statkwargs` was meant for this use, even though it was never implemented. Or, even if it was not, if `statargs/statkwargs` should be used for that. 
(That's a point for discussion, not a change request for this PR. It's perfectly fine to issue a warning that they are currently unused, because they are unused. That's not stopping us from using them later.)

DougBurke:
Rebased to address conflict.

DougBurke:
In response to https://github.com/sherpa/sherpa/pull/2259#issuecomment-2832830519

It's not clear to me that

a) it's only the statistics class that may need extra information (e.g. IXPE models)
b) that different classes of objects need the same information
c) can we know the information is constant within a fit or error analysis (I would think so but ....)

There is definitely a simplicity of being able to say "let's create a state object that just stores all this extra stuff that we can send around", but it's also a great way to create highly-coupled code!

I don't think this is necessarily different to @hamogu's view point, but I think we can go with the current "deprecate the stat*args  stuff" and wait until we have some actual use cases that will drive a design. That may involve un-deprecating these arguments.

hamogu:
> I don't think this is necessarily different to @hamogu's view point, but I think we can go with the current "deprecate the stat*args stuff" and wait until we have some actual use cases that will drive a design. That may involve un-deprecating these arguments.

I agree. It doesn't serve a purpose right now so it can be deprecated.

DougBurke:
Rebased as #2271 has been merged.

DougBurke:
Rebased to pick up 

- #2306 

because some follow-up work I want to do, to address 

- #2007

turns out to hit the big that #2306 fixes.

hamogu:
I don't think the other matters much. There will be a small rebase either way, but #2303 mostly add new stuff, so the changes should be as separated as we can do when working on the same classes at the same time.

sherpa/optmethods/__init__.py
```diff
@@ -279,16 +284,11 @@ def fit(self,
...
 
-        # Mid 2024 we do not have any tests where either statargs or
-        # statkwargs are not empty.
-        #
-        if statkwargs is None:
-            statkwargs = {}
-
-        def cb(pars):
-            return statfunc(pars, *statargs, **statkwargs)
+        if statargs is not None and statkwargs is not None:
```
hamogu:
```suggestion
        if statargs is not None or statkwargs is not None:
```

sherpa/utils/parallel.py
```diff
@@ -714,12 +715,49 @@ def parallel_map_funcs(funcs, datasets, numcores=None):
...
+    -------
+    ans : list
+       The seed sequences.
+
+    """
+
+    # TODO: Should the seed be created for a larger (i.e. more bits)
+    # data type?
+    #
+    maxval = np.iinfo(np.uint64).max
```
hamogu: It's just a seed. unint64 has enough options for seeds, we don't need that many sequences. THe problem for random number generators is usually the cycle, not the seed.
So, I suggest that this comment is not helpful and can can just be removed. tHis is not an issue that will come up or be studied for the Sherpa code again in our lifetime.
```suggestion
    maxval = np.iinfo(np.uint64).max
```

sherpa/stats/__init__.py
```diff
@@ -1160,3 +1160,34 @@ def calc_stat(self,
...
+# but not always). The statistic functions could return a
+# more-structured result (e.g. a dict or a dataclass), but this would
+# not stop the need for some way to extract just the desired data.
+#
+# There are multiple ways of doing this - such as a class or a
+# decorator - and it's not obvious there's much of a difference.
+# Using a class currently (when supporting Python 3.10 and 3.11) has
+# advantages in expressing the type information.
+#
+class StatCallback:
```
hamogu, line 1176: I suggest to also implement the analogous function to return the second value (the individual chi^2). I'll use it in my PR, and in any case it makes this "complete" and shows both options, maybe even with a class hirachy.

One other thought: This is (despite using a slotted class) a fairly complicated way to say `func(pars)[0]`. It's called in every step of the optimization loop; I'm certain that the performance implications are negligible, but it's one ofr layer in the call stack to go through and understand what's happening when benchmarking or debugging.
To be clear, I'm not suggesting to change this again right now (the code is already here and it's tested and it works), but I do want ot caution that the old joke that "every problem in CS can be solved with one more layer of indirection" can come with a real cost in readability of the code.

sherpa/optmethods/opt.py
```diff
@@ -33,19 +34,55 @@
...
+#
+# sherpa.stats.StatCallback converts a StatFunc into an OptimizerFunc.
+#
+OptimizerFunc = Callable[[np.ndarray], float]
+
+# WorkerFunc could be done as a Protocol, but the argument names are
+# not consistent, and it also requires **kwargs, which has not been
+# handled well by pyright. Concatenate could be used to at least
+# indicate the extra arguments but that requires Python 3.11 (as
+# Concatenate in Python 3.10 can not end with an ellipsis).
```
hamogu, line 69: Aside: When are we setting 3.11 as our minimum version? CIAO already only ships with 3.11. Personally, I would be tempted to go up 3.11-3.13 now and stop supporting 3.10. We may not have that much more time to work on Sherpa, so better to push it to the cutting edge now so that it's further away from deprecation when we can't actively develop any more.

sherpa/optmethods/opt.py
```diff
@@ -84,19 +131,41 @@ class Opt:
...
-        self.npar = len(xmin)
         self.xmin = np.asarray(xmin)
         self.xmax = np.asarray(xmax)
+        self.npar = len(self.xmin)
+        if len(self.xmax) != self.npar:
+            raise ValueError("xmin and xmax must be the same size")
+
+        # The function count should only be done on "valid" ranges,
+        # hence the ordering here.
+        #
```
hamogu: I don't understand this comment, but that's probably OK?

sherpa/optmethods/opt.py
```diff
@@ -140,46 +208,55 @@ def func_bounds(self,
...
+                 step: np.ndarray | None,
+                 seed: int | None,
+                 factor: float | None,
                  rng: RandomType | None = None
                  ) -> None:
         self.func = func
-        self.xmin = xmin
-        self.xmax = xmax
+        self.xmin = np.asarray(xmin)
+        self.xmax = np.asarray(xmax)
```
hamogu: Are these lists? Or could they by numpy arrays already?
In the latter case, I suggest the following to avoid creating a copy if it's not needed (but the arrays are small and this is not a performance bottleneck according to my benchmarking)
```suggestion
        self.xmin = np.asanyarray(xmin)
        self.xmax = np.asanyarray(xmax)
```

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
+                    *,
+                    step: np.ndarray | None = None,
+                    finalsimplex: int | None = 1,
+                    maxnfev: int | None = None,
+                    tol: float = 1.0e-6,
+                    verbose: int = 0
+                    ) -> MyOptOutput:
+        """
+
+        .. versionchanged:: 4.17.1
```
hamogu: Isn't this for the next version? 4.17.1 is already released.

sherpa/optmethods/ncoresnm.py
```diff
@@ -18,41 +18,82 @@
...
+#
+
+
+EPSILON = float(np.finfo(np.float32).eps)
 
 
 class MyNelderMead(Opt):
+    """
+
+    .. versionchanged:: 4.17.1
```
hamogu: I think this will be in 4.18.0.

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
+    def calc_step(self, x: np.ndarray) -> np.ndarray:
         return x + 1.2
 
 
 class NelderMead2(NelderMead0):
 
-    def calc_step(self, x):
-        return abs(x)
+    def calc_step(self, x: np.ndarray) -> np.ndarray:
+        return np.abs(x)
```
hamogu: Doesn't do any harm right here and right now, but this change is also not needed. `abs(x)` will always trigger `x.__abs__(self)`, which, for a numpy array, is equivalent to `np.abs(x)`. That works in the the same way that e.g. `x**3.2` will call `x.__pow__(self, 3.2)`, which, for a numpy array, will do that right thing.

Why do I bother to comment on this change at all? Scientific Python is advancing the array API where numpy might be just one package that provides arrays. `np.abs(x)` will not work easily if `x` is e.g. a jax or dask or xarray object, but `abs(x)` will work for all of those. 
There are clearly bigger problems and hundreds of other lines we will have to change if we decide to support arbitrary array libraries, and who knows if we ever do that, but I'm just trying to look for for that in general if it's a easy as "not to change something". (And I'm surely also guilty of similar edits.)

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
-                             NelderMead3(), NelderMead4(), NelderMead5()]):
-        # NelderMead6(), NelderMead7()]):
-        self.algo = algo
 
-    def __call__(self, fcn, x, xmin, xmax, tol=EPSILON, maxnfev=None,
-                 numcores=ncpus):
+class ncoresNelderMead:
+    """
+
+    .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
    .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
+
+    """
+
+    def __init__(self,
+                 algo: Sequence[NelderMeadBase] | None = None
+                 ) -> None:
+        if algo is None:
+            self.algo = [NelderMead0(), NelderMead1(), NelderMead2(),
+                         NelderMead3(), NelderMead4(), NelderMead5()]
+            # NelderMead6(), NelderMead7()]):
```
hamogu, line 486: I guess this is where the NelderMead6/7 that are now commented out above were supposed to be used. However, I agree that we make a choice ot which degree this should go and don't expect the user to change the algo here so deep in the stack. In other words: I support commenting out the code above.

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
+
+    def __call__(self,
+                 fcn: OptimizerFunc,
+                 x: np.ndarray,
+                 xmin: np.ndarray,
+                 xmax: np.ndarray,
+                 tol: float = EPSILON,
+                 maxnfev: int | None = None,
+                 numcores=ncpus
+                 ) -> MyOptOutput:
```
hamogu: Everywhere I see `MyOptOutput` I wonder if that's a good name. It's kind of "I'm the user this is what I do". This is not the only location obviously, but I finally have suggestions, so this is where I leave the comment. How about `SherpaOptOutputType` or `OptPutPutType` or `OptOutput` ( if that doesn't conflict with a class name somehere)?

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
-class ncoresNelderMeadRecursive(ncoresNelderMead):
-    """ As noted in the paper, terminating the simplex is not a simple task:
-    For any non-derivative method, the issue of termination is problematical as
-    well as highly sensitive to problem scaling. Since gradient information is
-    unavailable, it is provably impossible to verify closeness to optimality
-    simply by sampling f at a finite number of points. Most implementations
-    of direct search methods terminate based on two criteria intended to
-    reflect the progress of the algorithm: either the function values at the
-    vertices are close, or the simplex has become very small. """
+        .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
        .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresnm.py
```diff
@@ -163,229 +223,347 @@ def get_maxnfev(self, maxnfev, npar):
...
-    # there is a need to have a different set of classes.
-    #
-    def __init__(self, algo=[NelderMead0(), NelderMead1(), NelderMead2(),
-                             NelderMead3(), NelderMead4(), NelderMead5()]):
-        ncoresNelderMead.__init__(self, algo)
+        """
 
-    def __call__(self, fcn, x, xmin, xmax, tol=EPSILON, maxnfev=None,
-                 numcores=ncpus):
+        # As self.algo is not empty then results is not empty, but
```
hamogu, line 516: Do you mean:
```suggestion
        # If self.algo is not empty then results is not empty, but
```

sherpa/optmethods/ncoresde.py
```diff
@@ -237,10 +339,27 @@ def __call__(self, pop, icurrent):
 
 
 class MyDifEvo(Opt):
+    """
+
+    .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
    .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresde.py
```diff
@@ -255,134 +374,161 @@ def __init__(self, func, xpar, xmin, xmax, npop, sfactor, xprob, step,
...
+    #     fval_std = numpy.std([col[-1] for col in mypop])
+    #     if fval_std < ftol:
+    #         return True
+    #     return False
 
 
 class ncoresMyDifEvo(MyDifEvo):
+    """
+
+    .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
    .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresde.py
```diff
@@ -255,134 +374,161 @@ def __init__(self, func, xpar, xmin, xmax, npop, sfactor, xprob, step,
...
         if self.rng is None:
-            numpy.random.seed(int(self.seed))
+            np.random.seed(int(self.seed))
 
         mypop = self.polytope
         best_trial = self.strategies[0](mypop, index)
         for ii in range(1, len(self.strategies)):
             trial = self.strategies[ii](mypop, index)
-            if trial[-1] < best_trial[-1]:
+            if trial[1] < best_trial[1]:
```
hamogu: Why this change from the last (index -1) to the second (index 1) strategy in this test? I don't believe the statements are equal (they would be if there were only two strategies, but it looks like there are 6) but I also did not compare runs for the old and new branch to check. I can do that, but I figured it's easier to ask before I try to understand the numerical optimization algorithm in detail.

hamogu: Of course, it depends on what `mypop` etc is, so this code is not trivial to follow. THe change might be correct, or it might be actually equivalent code, but I'm not sure about that without looking deeper - hence I'm just quickly asking before I spend time doing that.

sherpa/fit.py
```diff
@@ -807,6 +807,9 @@ def sigmarej(self,
         The error removal can be asymmetric, since there are separate
         options for the lower and upper limits.
 
+        .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
        .. versionchanged:: 4.18.0
```

sherpa/fit.py
```diff
@@ -1006,19 +1010,25 @@ def fit(self,
...
             parmins: ArrayType,
             parmaxes: ArrayType,
-            statargs: Sequence[Any] = (),
-            statkwargs: Mapping[str, Any] | None = None,
+            statargs: Any = None,
+            statkwargs: Any = None
             ) -> OptReturn:
+        """
+
+        .. versionchanged:: 4.17.1
```
hamogu:
```suggestion
        .. versionchanged:: 4.18.0
```

hamogu (changes requested):

sherpa/stats/__init__.py
DougBurke: For the "one other thought" part:

The trouble is that we have all these local call-backs that do just this, because the lower-level routine doesn't need the second element, so the logic has to live somewhere. So I claim this is not adding a level of indirection, since this indirection already exists. It just makes it a bit more obvious what is going on.

This way we at least have it labelled so someone else who wants to clean it up has more idea what needs to be changed. I hope.

I don't think we want to change the low-level optimization code - which currently expects "func(pars)" to return a single value - to do this.

hamogu: OK, we'll leave it for now and we can always change it later if someone has a better idea.

sherpa/optmethods/ncoresde.py
DougBurke: I shall have to look into this. I had originally gone a lot further with this code, adding more structure to the return types rather than having it just smush everything into a ndarray, but that was much too extensive a change (and, more importantly, lead to differences in the tests), so I took it out. I don't think this particular change is related to the now-extant code, but you never know...

sherpa/optmethods/__init__.py
```diff
@@ -39,15 +39,20 @@
...
 
-    callback(pars, *statargs, **statkwargs)
+    callback(pars)
 
 and return the statistic value to minimise along with the per-bin
 statistic values (note that per-bin here refers to the independent
 axis of the data being fit, and not the number of parameters being
 fit).
 
+.. versionchanged:: 4.17.1
```
DougBurke:
```suggestion
.. versionchanged:: 4.18.0
```

sherpa/optmethods/__init__.py
```diff
@@ -239,11 +241,14 @@ def fit(self,
...
             parmins: ArrayType,
             parmaxes: ArrayType,
-            statargs: Sequence[Any] = (),
-            statkwargs: dict[str, Any] | None = None
+            statargs: Any = None,
+            statkwargs: Any = None
             ) -> OptReturn:
         """Run the optimiser.
 
+        .. versionchanged:: 4.17.1
```
DougBurke:
```suggestion
        .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresnm.py
DougBurke:
```suggestion
    .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresnm.py
DougBurke: [for reference, this code was originally written a long time ago, hence the version skew]

DougBurke:
```suggestion
        .. versionchanged:: 4.18.0
```

sherpa/optmethods/ncoresnm.py
DougBurke: The point here is that we expect `self.algo` to not be empty, but we aren't using a "this sequence is not empty" type [I haven't checked to see if this is true], so the type checkers have to allow for the list to be empty. The assert tells them it isn't empty. So I claim that "As" is correct and that "If" is actually not. However, I am biased - having looked at this code more than I ever wanted - so maybe I'm missing something.

sherpa/optmethods/opt.py
DougBurke: This code was originally written when we still supported Python 3.9 ... I'm not sure this PR is the one to make the Python version change.

sherpa/optmethods/opt.py
DougBurke: I can re-write this. The idea is that we only want to increase the function counter (which is what `FuncCounter` applies) if the data is within bounds. So we want to do the bounds check before `func_count` is called, hence the ordering.

DougBurke: [to be honest, the function counter stuff is, I think, a bit messed up for some cases, not necessarily this one, but I'm just trying to retain the current behavior as much as possible, as working on the "count" is a way-off prospect].

sherpa/optmethods/opt.py
DougBurke: We have some interesting code somewhere around here where we actually do want to make sure we have a copy, since the elements get changed and we don't want that reflected in the input data, but that's not true here.

sherpa/optmethods/ncoresde.py
DougBurke: Note that `trial` is the output of calling the function associated with `self.strategies[ii]` and sending it `(mypop, index)`. So, we're looking at the `__call__` method of `Strategy0` to `Strategy9`. This returns the value of `Strategy.calc()` method (since none of the sub-classes over-ride it). This returns a `MyOptOutput` "thing" (now, originally it returned a `ndarray` where all the data had been combined so that the last element was the function value). `MyOptOutput` is a tuple of `nfev`, `function value`, `pars`, hence we want the second element. We could make `MyOptOutput` "more structured" than a tuple - so that this code would be more readable - but I thought that was too large a change.

That is, we now have

https://github.com/sherpa/sherpa/blob/c37b06e5542a7d64f8550a547185b904fd6ef27a/sherpa/optmethods/ncoresde.py#L113-L123

and we used to have

https://github.com/sherpa/sherpa/blob/c0c4514bed1225595774e0fcad4044bc372595c6/sherpa/optmethods/ncoresde.py#L64-L72

I think the comment for `calc` needs to be updated, since the arg array no-longer gets changed [technically this could be a breaking change, since maybe the code was written to have the stat value be propagated this way, but it looks like the `arg` array is sent a copy of the trial data and so it not actually a problem].

sherpa/optmethods/opt.py
```diff
@@ -33,19 +34,55 @@
...
+#
+OptimizerFunc = Callable[[np.ndarray], float]
+
+# WorkerFunc could be done as a Protocol, but the argument names are
+# not consistent, and it also requires **kwargs, which has not been
+# handled well by pyright. Concatenate could be used to at least
+# indicate the extra arguments but that requires Python 3.11 (as
+# Concatenate in Python 3.10 can not end with an ellipsis).
+#
+MyOptOutput = tuple[int, float, np.ndarray]
```
DougBurke: So, here is where `MyOptOutput` is named. We have three issues

- essentially no documentation
- should it be more structured - e.g. a `dataclass` - so that data access is more obvious than just indexing into a container
- the name sucks

DougBurke: - name suckage

I don't have a strong feeling about this. I thought some of the existing code has `MyOpt` in it, which is why I came up with the name, but now I see no evidence of this. It was probably `MyNCores` and then thinking it's `OptOutput` and I merged it together. So I'll likely rename it to `OptOutput`.

DougBurke: - structure

Is it worth making it a `dataclass` (or something with more structure than a tuple). This PR is already a "breaking change" since the `MyOptOutput` used to just be a single `ndarray` which had the data all combined (hence losing semanting and type information that had to be reconstructed), so using a dataclass is no-more breaking than the existing use of a tuple. It's just all low-level code that only people working on the optimizer are going to come across.

sherpa/optmethods/ncoresnm.py
DougBurke: See discussion at https://github.com/sherpa/sherpa/pull/2259#discussion_r2210721974

sherpa/optmethods/opt.py
DougBurke: Hopefully this is clearer now.

sherpa/optmethods/ncoresde.py
DougBurke: It should now be more-obvious as `OptOutput` is now a `dataclass`. Whether that's a good thing is TBD.

sherpa/optmethods/opt.py
DougBurke: So, I've now 

- changed it to `OptOutput`
- added some documentation
- then changed the `tuple` to a `dataclass`

Whether the dataclass step is worth it is up for discussion.

sherpa/stats/__init__.py
DougBurke: For the main-part of the suggestion. In the follow-on PR

- #2022 

I do create such a class - https://github.com/sherpa/sherpa/blob/1bbbe0cff8d815a5dd62726db0643eab3494782e/sherpa/optmethods/optfcts.py#L1223-L1244 - but there it's only needed in `optfcts.py`. I can introduce it here.

Actually, I've put #2022 onto the back-burner, as it has more changes in than I think I want to include. So the latest variant is actually

- #2309

and there the change is

https://github.com/sherpa/sherpa/blob/a202cf5947f6c6114e7cd9a5293550862a50cbad/sherpa/optmethods/optfcts.py#L1326-L1347

sherpa/optmethods/ncoresde.py
```diff
@@ -531,12 +531,12 @@ def __call__(self,
...
                     best_par = np.append(tmp.pars, tmp.statval)
                     mypop[1] = best_par[:]
-                    self.polytope.sort()
+                    mypop.sort()
                     old_fval = tmp.statval
                 else:
                     old_fval = best_fval
 
-        best_vertex = self.polytope[0]
+        best_vertex = mypop[0]
```
hamogu: If `self.polytope` is exactly the same as `mypop` then why do we need the symbol `mypop` at all? Would it not be clearer to use use one and only one name for the thing?

sherpa/optmethods/opt.py
hamogu: One advantage of this approach is that it is future-proof. If we add an additional element to `OptOutput` in the future, then a dataclass like this can be changed without problems because everything is accessed by name. Scipy uses the same approach in e.g. optimization they return an object and you use `out.nfev`, `out.root` etc. In later versions you can then add `out.extra_message` without backwards-compatibility concerns.

sherpa/optmethods/opt.py
hamogu: No, it's not. I'll ask on sherpa-dev.

sherpa/optmethods/ncoresde.py
hamogu: A strategy now returns an `OptOutput`, so I would strongly prefer to replace `trial[n]` with `trial.xxx`. The new do that, it should, as you say, be easier to see if the -1 or the 1 is correct. (Indexing by number also works, presumably because the data class is defined with slots, but the whole benefit is that we don't have to do it by number any more.)

DougBurke: I don't understand what you mean - can you point to the particular code you are concerned about?

sherpa/optmethods/ncoresde.py
DougBurke: So, this is old code and I am just trying to make it more consistent. We can either have `mypop` or `self.polytope` and `mypop` has the advantage of being shorter (if you look at the code before I started it had a mix of `mypop` and `self.polytope` in). The advantage of using `self.polytope` is that you are reminded each time that you are dealing with an element in `self`. However, I have sometimes written code that will label an attribute of `self` just to make it shorter/easier to read.

sherpa/optmethods/ncoresde.py
DougBurke: If you look at the new code I've added, it now says

https://github.com/sherpa/sherpa/blob/73d9dc249001fa79b5ca5f7d5cbef147159dd6d5/sherpa/optmethods/ncoresde.py#L427-L448

which I think is what you were talking about?

Note also that `OptOutput` is not indexable, which helps remove this possibility of confusion:

```
>>> import numpy as np
>>> from sherpa.optmethods.opt import OptOutput
>>> x = OptOutput(nfev=10, statval=12, pars=np.asarray([1, 2, 3]))
>>> x[1]
TypeError: 'OptOutput' object is not subscriptable
```

DougBurke: Hmm, that `for ii in range(len...)` code could be rewritten to be more Pythonic, but I'm not sure it's worth it.

sherpa/optmethods/ncoresde.py
DougBurke: I've updated the code to just use `self.polytope`.

sherpa/optmethods/ncoresde.py
hamogu: What I mean is: Why does "trial[1]" work at all? I thought (from reading through the code) that `self.strategies[ii](mypop, index)` now returns a `OptOutput`, which, as you say, is not subscriptable.

https://github.com/sherpa/sherpa/pull/2259/files/226d4e44370aceb2c0f164267ce86098b7f4b7b2#diff-677c756a248f02f25bf502c8c6119440dc9e39f216ab12f22efcbfbd67eb01c3R79

DougBurke: Yes, and in the commit that I added `OptOutput` (**edited to add** as a dataclass rather than a tuple) I also changed the code here, as I mentioned in https://github.com/sherpa/sherpa/pull/2259#discussion_r2213549591 , so that it doesn't use `trial[1]` but `trial.statval`. So I'm a bit confused!

hamogu: I know I explicitly re-loaded the webpage, but someone GH still shoed the old version to me. I'm cloning the branch and checking the files locally now.

sherpa/stats/__init__.py
```diff
@@ -1160,3 +1160,62 @@ def calc_stat(self,
...
+
+     """
+
+     __slots__ = ("func", )
+
+     def __init__(self,
+                  func: StatFunc
+                  ) -> None:
+         self.func = func
+
```
hamogu, line 1217: To make is easier to see that it's essentially the same:
```suggestion
class PerBinStatCallback(StatCallback):
     """Return the per-bin statistic values for a set of parameters.

     .. versionadded:: 4.18.0

     See Also
     --------
     StatCallback

     """
```

sherpa/optmethods/optfcts.py
```diff
@@ -790,14 +800,13 @@ def myopt(myfcn, xxx, ftol, maxfev, seed, pop, xprob,
...
+            result = ncores_nm(stat_cb0, x, xmin, xmax, ftol, maxfev - nfev,
+                               numcores)
+            nfev += result.nfev
             # There is a bug here somewhere using broyden_tridiagonal
-            if tmp_fmin < fval:
-                fval = tmp_fmin
-                x = tmp_par
+            if result.statval < fval:
+                fval = result.statval
+                x = result.pars
```
hamogu, line 809: Much easier to read. Thanks!

sherpa/optmethods/ncoresde.py
DougBurke: Ah yes, the joys of having the computer dfo all the work for us .....

sherpa/stats/__init__.py
DougBurke: I'm not sure it makes sense to subclass them here - the return value for the `__call__` method is not the same - so I don't plan to take this suggestion.

Issue #887, opened 2020-08-13
sherpa.models.model.FilterModel - should we remove it?


DougBurke:
In my recent spelunking I've come across the FilterModel class which has no documentation and is only used in one place - a test to check we can create it. I show below it's use, and the fact that it appears to not work. Since it isn't used anywhere I suggest we remove it.

```
>>> from sherpa.models.basic import Const1D
>>> m = Const1D()
>>> m.c0 = 10
>>> m([1, 2, 3])
array([10., 10., 10.])
>>> m2 = m[2, 4 ,6]
>>> type(m2)
<class 'sherpa.models.model.FilterModel'>
>>> print(m2)
(const1d)[2,4,6]
   Param        Type          Value          Min          Max      Units
   -----        ----          -----          ---          ---      -----
   const1d.c0   thawed           10 -3.40282e+38  3.40282e+38
```

So we can create a filter model, but trying to evaluate it raises an error.

```
>>> m2([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dburke/anaconda/envs/ciao412/lib/python3.7/site-packages/sherpa/models/model.py", line 265, in __call__
    return self.calc([p.val for p in self.pars], *args, **kwargs)
  File "/home/dburke/anaconda/envs/ciao412/lib/python3.7/site-packages/sherpa/models/model.py", line 679, in calc
    return self.model.calc(p, *args, **kwargs)[self.filter]
IndexError: too many indices for array
```

anetasie:
@DougBurke just in time for the discussion on code documentation. I just missed it during the meeting.

DougBurke:
It's a really low-priority issue ;-)

hamogu:
I think you missed what this specific model is good for:
https://github.com/sherpa/sherpa/blob/master/sherpa/models/model.py#L546

It's the default `_getitem__` implementation for any ArtithmetricModel. That's not saying that it can't possibly be removed, but that it's part of the machinary on a much deeper level than "only used in a single test".

DougBurke:
Unfortunately the only test that makes use of this functionality is the "existence" test for the model. I changed the `__getitem__` method to raise a NotImplementedError and the only test that failed was the one below. So I claim it's still unused/untested :-)

```
========================================== FAILURES ==========================================
______________________________ test_composite_model.test_filter ______________________________
[...]

PR #2265, opened 2025-04-05
Add new public methods to access the components of a model

### LINKED ISSUE #2221: Accessing the parts of combined models
When working directly with models (not the the UI) I often run into situations like this:

I have a model made up of multiple parts, say the sum of a continuum and several Gaussian lines or a specific line with a complicated shape, such that that one line is fit with a combination of Gaussians or Lorentzians.

For example:
```Python
from sherpa.models import NormGauss1D
from sherpa.astro.models import Lorentz1D

line = NormGauss1D('main') + NormGauss1D('red_peak') + Lorentz1D('wings')
```

Then, I fit my data and want to set specific starting values, parameters links, or caps, but that's cumbersome and requires me to know more about the internal structure of the model (`rhs` and `lhs`) than most people do know or should know: `line.lhs.rhs.fwhm.max = 0.3`

Or course, I could have kept the individual components as separate objects:
```python
main = NormGauss1D('main') 
red_peak = NormGauss1D('red_peak')
wings = Lorentz1D('wings')
line = main + red_peak + wings
```
but that's (a) a lot of boilerplate, (b) get unwieldy if I have more than one spectral line to fit (I need a list of lists that hold the components or need to name them `wings_1`, `wings_2`, ...).

Instead, I propose a new way to get the components of model. In my own code, I often use the following function which really is just one line of code (but might have to be modified to deeply nested models - I didn't test that):
```Python
def flatten_sherpa_model(model):
    '''Flatten a Sherpa model into a list to make easy to iterate over

    Parameters
    ----------
    model : `sherpa.models.model.BinaryOpModel`
        Sherpa model consisting of different parts

    Returns
    -------
    modellist : list of `sherpa.models.model.Model`
        Flat list of individual models
    '''
    return [m for m in model if not hasattr(m, 'parts')]
``` 

An even more convenient alternative would be to get submodels by name like: `line['red_peak'].fwhm.max = 0.3`. Since we often don't give unique names, I suggest to return a list of models with the same name in that case:
```Python
star = xspec.XSphabs(xspec.XSvapec() + xspec.XSvapec())
list_of_two_vapec_models = star['vapec']
```

None of this is hard to implement - you just have to traverse the tree of model components - but I want to start with this issue instead of a PR to see if 
- this is either already implemented and I don't know about it
- get some feedback on the API before writing that code.

Since this is adding a new way to get to model components, not changing an existing way, there are not problems with backwards compatibility.

hamogu:
## Summary

Add new methods to select the components of a model. 

## Details

All `CompositeModels` (which will be at the root of a tree  of models for any model with more than a single component) gain methods to select components either by string name (`mdl.get_components_by_name()`) or by class, e.g. to get all `Gauss1D` components (`mdl.get_components_by_class()`). In order to implement that smoothly, the `mdl._get_parts()` method is now part of the API as `get_parts` with parameters to return only the leaves of the tree and to include or exclude the nodes (`UniaryOpModel` or `BinaryOpModel`).

For convenience, there is also dictionary-like access by name (e.g. `mdl['line2']`) and by class (`mdl[Gauss1D]`).

## Design decisions

- I have chosen to have an inconsistent return type for attribute access: If there is only one element, it's returned directly, if there is more than one, it returns a list. In practice, I anticipate that this will be mostly used in cases where the user knows if the selector is unique (e.g. the name of the spectral line - see my example in the docs). Then, always writing `mdl['S II'][0].fwhm=3` is a lot of extra boilerplate, but I also don't want it to fail with an error just because more than one element matches. "Practicality beats purity". Access by functions consistently returns lists.
- For selectors not found in the model, I raise a `KeyError` because it behaves like a dictionary.
- Currently, the implementation lives in the new methods, the dictionary-like access is just for convenience and syntactic sugar. It could be removed if it feels too magic, but I really, really like it for its convenience and compressed notation.
- One could argue that the dict-like access should only work for strings and not for classes. That might make it a bit less magical -> Decided to make it work for both classes and strings.
- As long as the dictionary-like access is the main avenue of use, long and descriptive names or the methods are OK. If we think that the methods are going to be used interactively or in notebooks, they need to get much shorter names.

closes #2221

### DIFF
--- a/docs/evaluation/combine.rst
+++ b/docs/evaluation/combine.rst
@@ -340,7 +340,7 @@ has many separate emissions lines::
 
     >>> spectral_lines = {'[O I]': 6300, 'Hα': 6563, '[N II]': 6586, '[S II]': 6716, '[S II]': 6731}
     >>> jetlines = [models.Gauss1D(line) for line in spectral_lines.keys()]
-    >>> jetemission = models.Const1D('background') + sum(jetlines)
+    >>> jetemission = models.Const1D('background') + np.sum(jetlines)
     >>> for line, wave in spectral_lines.items():
     ...     jetemission[line].pos = wave
     ...     jetemission[line].fwhm = 0.1
--- a/docs/evaluation/combine.rst
+++ b/docs/evaluation/combine.rst
@@ -272,7 +272,24 @@ it can be evaluated on a grid by passing in an array of values::
 
     >>> sim_model([-1.0, 0, 1])
     array([  1.52587891e-05,   1.00003815e+00,   5.34057617e-05])
-    
+
+In the example above, the model consists of two components ``sim1`` and ``sim2``
+and we keep referencing them by their original variables. In general, more complex models
+with more components can be built, which will then be arranged in a tree where the
+leaves are the original components and the internal nodes are the either
+:py:class:`~sherpa.models.model.BinaryOpModel` (which combine two models) or
+:py:class:`~sherpa.models.model.UnaryOpModel` (with modify just one model on
+the level below, e.g. by taking the absolute value of the output of the lower level model)
+instances. Those models can be quite deep and thus Sherpa provides a syntax to access the
+components of a model tree either by name or by model class. If more than one component
+matches the name or class, a list of all matching components is returned::
+
+    >>> sim_model['sim2']
+    <Gauss1D model instance 'sim2'>
+    >>> sim_model[models.Gauss1D]
+    [<Gauss1D model instance 'sim1'>, <Gauss1D model instance 'sim2'>]
+
+
 Setting up the model
 --------------------
 
@@ -310,6 +327,24 @@ parameter is now linked to the ``g1.pos`` value::
        g2.fwhm      thawed          0.1  1.17549e-38  3.40282e+38           
        g2.pos       linked          0.5       expr: g1.pos + 0.5           
        g2.ampl      thawed            1 -3.40282e+38  3.40282e+38           
+
+
+An alternative way to write the code above is to select the model components by name::
+
+    >>> mdl['g2'].fwhm = 0.1
+
+While not necessary in this example, it can make it easier to keep track of a
+model with many components and simplify code that loops over model components.
+In the following example we want to built a model for the jet from a young star that
+has many separate emissions lines::
+
+    >>> spectral_lines = {'[O I]': 6300, 'Hα': 6563, '[N II]': 6586, '[S II]': 6716, '[S II]': 6731}
+    >>> jetlines = [models.Gauss1D(line) for line in spectral_lines.keys()]
+    >>> jetemission = models.Const1D('background') + sum(jetlines)
+    >>> for line, wave in spectral_lines.items():
+    ...     jetemission[line].pos = wave
+    ...     jetemission[line].fwhm = 0.1
+    >>> jetemission['Hα'].fwhm = 5.
 
 .. note::
 
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -733,7 +733,7 @@ class Model(NoNewAttributesAfterInit):
 
     def get_parts(self,
                   include_composites : bool = True,
-                  remove_duplicates : bool = False) -> Sequence[Model]:
+                  remove_duplicates : bool = False) -> list[Model]:
         """Return the parts of the model.
 
         Parameters
@@ -766,13 +766,19 @@ class Model(NoNewAttributesAfterInit):
 
         Returns
         -------
-        modellist : list
+        parts : list
             The components of the model with the given name.
         """
         parts = [p for p in self.get_parts(remove_duplicates=True) if p.name == name]
         if len(parts) == 0:
             raise KeyError(f"No components found with name '{name}'")
         return parts
+
+    @staticmethod
+    def _checkfunc(p: Model,  cls: type, subclass_ok: bool = True) -> bool:
+        if subclass_ok:
+            return isinstance(p, cls)
+        return type(p) is cls
 
     def get_components_by_class(self, cls: type, subclass_ok: bool = True) -> list[Model]:
         """Return the components of the model with the given type.
@@ -788,13 +794,11 @@ class Model(NoNewAttributesAfterInit):
 
         Returns
         -------
-        list[Model]
+        parts: list
             The components of the model with the given type.
         """
-        if subclass_ok:
-            parts = [p for p in self.get_parts(remove_duplicates=True) if isinstance(p, cls)]
-        else:
-            parts = [p for p in self.get_parts(remove_duplicates=True) if type(p) is cls]
+        parts = [p for p in self.get_parts(remove_duplicates=True) if self._checkfunc(p, cls, subclass_ok)]
+
         if len(parts) == 0:
             raise KeyError(f"No components found with type '{cls}'")
         return parts
@@ -802,6 +806,10 @@ class Model(NoNewAttributesAfterInit):
 
     def __getitem__(self, key: Model | str) -> Model | list[Model]:
         """Return the components of the model with the given name or type.
+
+        .. versionchanged:: 4.17.1
+           Parameter access is now possible with either a string (the name of a model)
+           or a type.
 
         Parameters
         ----------
@@ -1227,9 +1235,9 @@ class CompositeModel(Model):
 
     def get_parts(self,
                   include_composites : bool = True,
-                  remove_duplicates : bool = False) -> Sequence[Model]:
+                  remove_duplicates : bool = False) -> list[Model]:
         # Docstring inherited from base class
-        parts = []
+        parts : list[Model] = []
 
         # Including itself seems a bit strange if it's a CompositeModel
         # but is used by sherpa.astro.instrument.has_pha_instance (and
--- a/sherpa/models/model.py
+++ b/sherpa/models/model.py
@@ -351,7 +351,7 @@ warning = logging.getLogger(__name__).warning
 
 __all__ = ('Model', 'CompositeModel', 'SimulFitModel',
            'ArithmeticConstantModel', 'ArithmeticModel', 'RegriddableModel1D', 'RegriddableModel2D',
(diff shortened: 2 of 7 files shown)


DougBurke:
So, there's always this inconsistency when dealing with a container that behaves like a single object. In this case, `mdl1 = Gauss1D()` and `mdl2 = Gauss1D() + Const1D()` will behave differently because only `mdl2` has the `get_components_by_xxx` and dict-like access. There's potential confusion for the user and the "script" writer:

- for the former we have documentation, but will people read it?
- for the latter, they will have to say things like

```python
try:
   blah = mdl.get_components_by_xxx(...)
except AttributeError:
   blah = [mdl]  # or maybe error out or should it be a dict or ...
```

I'm not saying this is a bad thing, just pointing it out as something to briefly discuss.

hamogu:
Good point. I think that would annoy the heck out of me when I try to use this in practice. Even a single model should allow that access, so that a single model or a container of model behave the same. Obviously, in a single model, there won't be as many components that you can access, but you should be able to access the model itself that way.

hamogu:
This turned out to be much more difficult that I thought it would be, because I got confused by a bug in main: The old `_get_parts` actually does not return all parts, but leaves out some (but not all) of the `BinaryOpModels`. Thus, I struggled to reproduce old tests, until I eventually realized that the new result is correct and the tests needed to be adjusted, not the other way around.

This PR is now split into two commits: The first one implements access by methods, the second one by attribute access. Personally, I think the second one is more valuable, but it requires removing the old attribute access to certain models that could have been used to initialize a `FilterModel`. The code has long had a note that `FilterModel` is indeed unused and, while listed in the public API, not advertised and it's not clear what it was meant for (I've tried again now, and also can't figure out how I would use it). Since these new changes are incompatible with "using attribute access to initialize a FilterModel", the second commit breaks backwards-compatibility and at that point I decided to remove `FilterModel` altogether.

DougBurke:
You may want to link to 

- #887

here

sherpa/models/model.py
```diff
@@ -731,6 +731,75 @@ def __setattr__(self, name: str, val) -> None:
...
+        type : type
+            The type of the component to return.
+        subclass_ok : bool, optional
+            If True, return components that are subclasses of the given
+            type. If False, only return components that are exactly of
+            the given type.
+
+        Returns
+        -------
+        list[Model]
```
DougBurke, line 791: On line 769 we have "modellist : list" and here just "list[Model]". We should gave the same in both.

sherpa/models/model.py
```diff
@@ -731,6 +731,75 @@ def __setattr__(self, name: str, val) -> None:
...
+            The components of the model with the given type.
+        """
+        if subclass_ok:
+            parts = [p for p in self.get_parts(remove_duplicates=True) if isinstance(p, cls)]
+        else:
+            parts = [p for p in self.get_parts(remove_duplicates=True) if type(p) is cls]
```
DougBurke, line 797: This could be something like (untested)

```python
def checkfunc(p: Model) -> bool:
    if subclass_ok:
        return isinstance(p, cls)
    return type(p) is cls

parts = [p for p in self.get_parts(remove_duplicates=True) if checkfunc(p)]
```

We could instead define `checkfunc` "within" the `subclass_ok` check, which would be more efficient but I think it would be harder to follow and we don't care too much about efficiency here.

sherpa/models/model.py
```diff
@@ -731,6 +731,75 @@ def __setattr__(self, name: str, val) -> None:
         for alias in val.aliases:
             self._par_index[alias] = val
 
+    def get_parts(self,
+                  include_composites : bool = True,
+                  remove_duplicates : bool = False) -> Sequence[Model]:
```
DougBurke, line 736:
```suggestion
                 remove_duplicates : bool = False) -> list[Model]:
```

There's a whole thing about contravariant vs covariant types, and I think we want to be explicit in the return type rather than a more-generic type.

DougBurke: oops - missed a space.

sherpa/models/model.py
hamogu: It could be but that's actually 5 lines instead of 4 and another level of indirection. It's not reused in other parts of the code so I don't see a benefit to abstract it out here.
However, if @DougBurke insists I'm happy to do it, I don't feel very strongly about it.

DougBurke: It's not so much the characters saved, but I think it's more obvious what is different when `subclass_ok` is set or not. With the original  version I have to check the `parts = [..]` line twice to spot the difference, where in the new version it is more-obvious (to me) that `subclass_ok` only changes how the model and class are compared.

sherpa/models/model.py
hamogu: I'm now to typing (I think this is the first PR where I try to ype things like that) and mypy as complaining about exactly contravariant vs covariant types. I have ot admit I don't understand it and it also doesn't seem to matter much as a list is just a specialized Sequence, so I figured I just take whatever makes the warning go away. But then it shows up the other way around somewhere else. Maybe it's better ot just NOT type is in that case?

sherpa/models/model.py
hamogu: done.

sherpa/models/model.py
DougBurke: So, the typing docuementation doesn't explicitly call this out at the moment, but you have statements like

> To summarize, function parameters should behave contravariantly and function return types should behave covariantly.

that occur **just before** this link: https://typing.python.org/en/latest/spec/callables.html#passing-kwargs-inside-a-function-to-another-function

Or this may be better: from https://typing.python.org/en/latest/spec/callables.html#parameter-types

> Callable types are covariant with respect to their return types but contravariant with respect to their parameter types.

[this is all about using `Callable` but I assume it somewhat holds elsewhere].

The `mypy` documentation can also be looked at: from https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generics includes

> Most immutable container types, such as [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) and [frozenset](https://docs.python.org/3/library/stdtypes.html#frozenset) are covariant.

So this suggests that the way @hamogu had it originally made sense - but then all the other `get_parts` calls need to be converted to `...) -> Sequence[Model]`. 

I had thought I'd started to get a handle on this but I can no-longer find the documentation I had been following.

hamogu: Thank you for the explanation, but I think I figured it out with an extra typing hint in the middle of the code: `parts : list[Model] = []`. I just needed to tell MyPy what `parts` is supposed to be so it knows that it's not just "a general list" but "a list with Models".

docs/evaluation/combine.rst
```diff
@@ -311,6 +328,24 @@ parameter is now linked to the ``g1.pos`` value::
...
+    >>> mdl['g2'].fwhm = 0.1
+
+While not necessary in this example, it can make it easier to keep track of a
+model with many components and simplify code that loops over model components.
+In the following example we want to built a model for the jet from a young star that
+has many separate emissions lines::
+
+    >>> spectral_lines = {'[O I]': 6300, 'Hα': 6563, '[N II]': 6586, '[S II]': 6716, '[S II]': 6731}
+    >>> jetlines = [models.Gauss1D(line) for line in spectral_lines.keys()]
+    >>> jetemission = models.Const1D('background') + sum(jetlines)
```
DougBurke, line 343: So

```
>>> bob = sum([b1, b2, b3])
>>> print(bob)
0.0 + b1 + b2 + b3
   b1.c0        thawed           10 -3.40282e+38  3.40282e+38           
   b1.c1        frozen            0 -3.40282e+38  3.40282e+38           
...
   b2.c0        thawed            1 -3.40282e+38  3.40282e+38           
   b2.c1        frozen            1 -3.40282e+38  3.40282e+38           
...
   b3.c0        thawed           -5 -3.40282e+38  3.40282e+38           
   b3.c1        frozen            0 -3.40282e+38  3.40282e+38           
   b3.c2        frozen            2 -3.40282e+38  3.40282e+38           
...
```

I wonder if users are going to get confused by the magic "0.0" (because we don't talk to our children enough about [monoids](https://en.wikipedia.org/wiki/Monoid)).

But that's not really relevant for this example

sherpa/models/model.py
```diff
@@ -800,6 +800,33 @@ def get_components_by_class(self, cls: type, subclass_ok: bool = True) -> list[M
         return parts
 
 
+    def __getitem__(self, key: Model | str) -> Model | list[Model]:
+        """Return the components of the model with the given name or type.
+
+        Parameters
```
DougBurke, line 810: I think worth adding `.. versionchanged:: 4.17.1`  text here as it is a significant change.

DougBurke:
Is the `Status` field above still valid? It talks about needing to update documentation, but I think you've now updated it (or were you thinking there's more to do?).

docs/evaluation/combine.rst
hamogu: Solvable:
```Python
>>> jetemission = models.Const1D('background') + np.sum(jetlines)
background + [O I] + Hα + [N II] + [S II]
   background.c0 thawed            1 -3.40282e+38  3.40282e+38           
   [O I].fwhm   thawed           10  1.17549e-38  3.40282e+38           
   [O I].pos    thawed            0 -3.40282e+38  3.40282e+38           
   [O I].ampl   thawed            1 -3.40282e+38  3.40282e+38           
   Hα.fwhm      thawed           10  1.17549e-38  3.40282e+38           
   Hα.pos       thawed            0 -3.40282e+38  3.40282e+38           
   Hα.ampl      thawed            1 -3.40282e+38  3.40282e+38           
   [N II].fwhm  thawed           10  1.17549e-38  3.40282e+38           
   [N II].pos   thawed            0 -3.40282e+38  3.40282e+38           
   [N II].ampl  thawed            1 -3.40282e+38  3.40282e+38           
   [S II].fwhm  thawed           10  1.17549e-38  3.40282e+38           
   [S II].pos   thawed            0 -3.40282e+38  3.40282e+38           
   [S II].ampl  thawed            1 -3.40282e+38  3.40282e+38   
```

hamogu:
Status updated, version changed added.

DougBurke (approved):
LGTM

PR #2268, opened 2025-04-10
Treat norm special in XSEPC additive models

### LINKED ISSUE #767: should modelCacher1d special case models with a normalization?

hamogu:
## Summary

These code changes increase the speed of fitting data with XSPEC additive models by making the cache more effective.

## Details

All XSPEC additive models have a parameter "norm", which is just a multiplicative factor. By using an additional decorator, we now do this multiplication in Python and always pass norm=1 to the XSPEC C/Fortran function. Together with the exisiting cache, that holds the return of an XSPEC  function, this speeds up model fitting, because it increases the number of cache hits. (Now, any change in the norm compared to the last call will pull the result from the cache and just multiply on the new norm.)

There is a slight overhead due to the additional level of indiraction of order 1 ms and some models like APEC seem to have an internal cache in the C-code that has a similar function. Thus, the speed gain depends on what model and what data is fit. Extensive benchmarking is documented in #767 on github. Speedups in fitting can reach 10%-20% for the overall fit for slow models, even if an individual model evaluations takes longer, because fewer evaluations are needed. In the worst case (for the
fastest XSPEC models the added overhead is measurable, but the overall time to fit data is to fast  that this PR does not degrade the performace in practice.

Additionally (see discussion in #767), it seems that for some models a loss of precision in the norm between Python and C/Fortran can cause the optimizer to wiggle around the best-fit value. Keeping the norm in Python keeps the precision and reduces the total number of steps needed in a fit, which contributes to the speed gain.

closes #767


hamogu:
@DougBurke : Please review even though not all checks are reported passing. 

I happened to push during a GH anomaly (https://www.githubstatus.com/incidents/vxfd5br11t6v) which stalled the CI. I have restarted some of the jobs, but others I can't restart manually for some reason. The obvious solution is to push a new commit. I'll have to do that almost certainly after review anyway - so maybe you can just do the review in this partially-passed state for now,

hamogu:
The test failure in https://github.com/sherpa/sherpa/actions/runs/14370913062/job/40317623983?pr=2268 is real (I wonder why I didn't see that locally, but that's a different issue), but it's actually an expected (and welcome!) change in behavior for this PR, so all that needs to be done is to change the test to match the new behavior after applying the changes in this PR. 
Again, I feel it's not worth wasting the cycles to push that change right now, I can lump it in with other changes after an initial review.

sherpa/astro/xspec/__init__.py
```diff
@@ -1752,6 +1753,50 @@ def mkabund(name: str,
...
+    lower precision floats internally. The higher precision can reduce the number of
+    function evaluations needed to reach the same accuracy in the fit.
+    """
+    @functools.wraps(func)
+    def cache_model(cls, pars, xlo, *args, **kwargs):
+        # While the following line looks harmless, benchmarking indicates a
+        # noticable performance hit from the lookup of order 10 ms per fuction call,
+        # thus it is commented out in the production version of the code.
+        # assert cls.pars[-1].name == 'norm'
+        pars_with_norm_1 = (*pars[:-1], 1)
```
DougBurke, line 1795: The additive models have a certain amount of code added in C++ to deal with the `norm` parameter (since XSPEC handles `norm` itself rather than as a parameter of the model). 

So we **could**

- drop that code
- then just have the `calc` routine send `pars[-1]` to the compiled code and apply the norm there

and this would actually not change anything here (I think; it's Friday and I haven't thought this through).

hamogu: If that's the case, we could also drop the norm from the C++ code entirely and apply the norm here instead of there?

DougBurke: We need to have it somewhere in `calc` since this routine is the cache code and so may not be called ;-)

hamogu: This code is guaranteed to be called every time.
Maybe the "cache" in the name is a misnomer here. I mean "modify how we use the cache", but I do that by wrapping the cache+calc block (by which I mean "the `calc` that is already decorated with the `modelCacher1d`), i.e. this decorator is applied on the outside of the "cache+calc" block. It is called every time the (decorated) `calc` is called and works by passing `norm=1` to the `modelCacher1d`, which then in turn may or may not (depending on what's in the cache already) call the actual `calc` method. 

It's essentially like this
```Python
@eval_xspec_with_fixed_norm
@modelCacher_1d
def calc(..)
```

However, in practice, it's a little harder to see, because the `@eval_xspec_with_fixed_norm` is applied to the `calc` of the XSAdditivemodel, which then calls `super().calc` and `super().calc` which is written like this:
```Python
@modelCacher_1d
def calc(..)
```
The effect is that the two decorators are still effectively applied on top of each other as I described above, but they are located many lines of code apart in the file so it's hard to see.

DougBurke: So, #2275 is my PR that changes additive models - **including** table models which are not part of this PR but we could think about here - to apply the `norm` in Python and not in the compiled code.

I have not tried to see what difference it makes to performance.

I have also not tried to see how it would be used by this PR. Technically I think they are completely separate, but maybe we'd want to change where the cache is applied (hand wavy thoughts of applying it to the `._calc` method instead of `calc` but I am not sure that would make sense / work / be a good idea and is likely a distraction here).

sherpa/astro/xspec/__init__.py
```diff
@@ -104,11 +104,12 @@
 from __future__ import annotations
 
 from contextlib import suppress
+import functools
 import logging
 from pathlib import Path
 import string
 import tempfile
-from typing import Any, overload
+from typing import Any, overload, Callable
```
DougBurke, line 112: We should 

```python
from collections.abc import Callable
```

instead.

sherpa/astro/xspec/__init__.py
```diff
@@ -1752,6 +1753,50 @@ def mkabund(name: str,
...
+    Python, the norm retains the full precision of a float, while some XSPEC models use
+    lower precision floats internally. The higher precision can reduce the number of
+    function evaluations needed to reach the same accuracy in the fit.
+    """
+    @functools.wraps(func)
+    def cache_model(cls, pars, xlo, *args, **kwargs):
+        # While the following line looks harmless, benchmarking indicates a
+        # noticable performance hit from the lookup of order 10 ms per fuction call,
+        # thus it is commented out in the production version of the code.
+        # assert cls.pars[-1].name == 'norm'
```
DougBurke, line 1794:
```suggestion
```

Do we need this comment?

sherpa/astro/xspec/__init__.py
```diff
@@ -1791,6 +1836,10 @@ def guess(self, dep, *args, **kwargs):
         norm = get_xspec_norm(dep, self(*args))
         param_apply_limits(norm, self.norm, **kwargs)
 
+    @eval_xspec_with_fixed_norm
+    def calc(self, p, xlo, *args, **kwargs):
```
DougBurke, line 1840: The superclass - in this case `XSModel` has

```
```

so it might be best to keep this argument naming (our classes do not have a good history of that, but I'm trying to rain in the changes a little bit).

sherpa/astro/xspec/__init__.py
DougBurke, line 1795: So, even though the tests pass I don't see how this can work:

```
>>> pars = np.asarray([10, 20, 30, 40, 50])
>>> len(pars)
5
>>> len(pars[:-1])
4
>>> pars1 = (pars[:-1], 1.0)
>>> len(pars1)
2
>>> len(pars1[0])
4
>>> type(pars1)
<class 'tuple'>
>>> type(pars1[0])
<class 'numpy.ndarray'>
>>> type(pars1[1])
<class 'float'>
```

I was expecting to see something like

```python
pars_with_norm_1 = np.append(np.asarray(pars[:-1]), 1.0)
```

[the `np.asarray` call may be superflous].

sherpa/astro/xspec/__init__.py
hamogu: It's a balance between consistency and readability.  In most parts of the models, we use 'p' for parameters (which are part of args here), while the superclass uses p for what all other routines call x0.

DougBurke: I don't see one where the parent class uses `p` for `x0` - most of them use `p, *args, **kwargs` from what I can see. Some may use `pars` rather than `p`, and the "convolution" style models are an "all bets are off" case. I also am ignoring the optical / astro model classes as they are their own thing ....

hamogu: The point is that the first parameter is `xlo`. not `pars, but I'll change that for consistency.

sherpa/astro/xspec/__init__.py
hamogu: Parameters are passed around as Python tuples, not numpy arrays (and that makes sense, because for small tuples, say < a few hundred entries, tuples and lists are often faster than numpy arrays for concatenations etc)
So `(*pars[:-1], 1)` works in these steps:
- `pars[:1]` selects all parameters except for the last one (the norm)
- `*` expands a tuple or list into individual elements
- `, 1` adds one more element (=1) at the end
- `( ...)` turns the combined stuff back into a tuple.

 It's the same syntax used in `func(*args)`, where `args` is a list or tuple, but `func` doesn't get a single parameter that's a list or tuple, but a number of separate parameters equal to the length of the `args` sequence.

Models typically have few parameters. Going through numpy takes about 20 times longer:
```Python
>>> pars = (2, 1e-34, 2.345, 1e-6)
>>> %timeit newpars = (*pars[:-1], 1)
>>> %timeit newpars = np.append(np.asarray(pars[:-1]), 1.0)
67.4 ns ± 0.43 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
1.35 μs ± 9.68 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

DougBurke: do - I'd missed the "`*`".

Note that during a fit it is very likely that the parameters will be sent in as a ndarray, not a list or tuple. I doubt that changes the timing much though (that is, creating the tuple rather than explicitly extending a ndarray (but that when the input is a ndarray it's not as fast as the tuple):

```
In [1]: import numpy as np

In [2]: pars1 = (12.13, 3.4e-6, 125.3, 12.3)

In [3]: pars2 = np.asarray([12.13, 3.4e-6, 125.3, 12.3])

In [4]: %timeit newpars1 = (*pars1[:-1], 1.0)
134 ns ± 0.0528 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit newpars2 = (*pars2[:-1], 1.0)
1.19 μs ± 3.42 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit newpars3 = np.append(pars2[:-1], 1.0)
2.18 μs ± 5.73 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

sherpa/astro/xspec/__init__.py
DougBurke: I still don't understand: we don't care what any argument is called after `p`/`pars` as we just send it through as `*args`:

```python
@superduperdecorator
    return super().calc(p, *args, **kwargs)
```

Or am I missing something important?

sherpa/astro/xspec/__init__.py
```diff
@@ -1752,6 +1753,50 @@ def mkabund(name: str,
...
 
 
+
+def eval_xspec_with_fixed_norm(func: Callable) -> Callable:
+    """Decorator to evaluate an `XSAdditativeModel` with norm=1.
+
+    The norm is then applied in Python, instead of in the XSPEC code.
+    This is a speed optimization. For a single call to an XSPEC model, there
+    is no advantage, because the XSPEC model needs to be called anyway; it is
+    just called with norm=1 instead of the actal norm and the output of the
```
DougBurke, line 1763:
```suggestion
    just called with norm=1 instead of the actual norm and the output of the
```

Although to be honest I think a lot of this can be removed; I just haven't bothered to come up with a suggestion.

sherpa/astro/xspec/__init__.py
```diff
@@ -1752,6 +1753,50 @@ def mkabund(name: str,
                        hard_min=minval, hard_max=maxval, frozen=True)
 
 
+
+def eval_xspec_with_fixed_norm(func: Callable) -> Callable:
+    """Decorator to evaluate an `XSAdditativeModel` with norm=1.
```
DougBurke, line 1758:
```suggestion
    """Decorator to evaluate an `XSAdditiveModel` with norm=1.
```

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -735,6 +737,21 @@ def test_evaluate_xspec_model(modelcls):
 
     Convolution models are skipped (easier to filter out here given the
     current design).
+
+    This test checks two independent things:
+    1. The model can be evaluated with the default parameters
+       and the results agree and on energy and wavelength grids.
+    2. For additative models: Calling the XSPEC model with the norm gives the
```
DougBurke, line 744:
```suggestion
    2. For additive models: Calling the XSPEC model with the norm gives the
```

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -750,6 +767,12 @@ def test_evaluate_xspec_model(modelcls):
     # so there is no need to check that the output of
     # mdl does not contain non-finite values.
     #
+
+    # This second test would be useless if run with a default norm=1
+    if isinstance(mdl, xspec.XSAdditiveModel):
+        mdl.norm = 0.123 * mdl.norm.val
+
+    # 1) Check that energy and wavelangth give the same result
```
DougBurke, line 775:
```suggestion
    # 1) Check that energy and wavelength give the same result
```

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -735,6 +737,21 @@ def test_evaluate_xspec_model(modelcls):
...
+       same answer as calling the model with a decorator that applies the norm
+       in Python and calls XSPEC with norm=1.
+
+    Both asserts are combined into a single test because the test runtime is
+    significant (> 1 min on my machine when iterating over all models), but
+    much of that runtime is needed to initialize the models.
+    If, instead, this was split into two separate tests, that initizalization
+    would have to be done twice which would lead to noticeably longer runtimes
+    in particular on CI, which checks several different environments and
+    platforms.
```
DougBurke, line 754:
```suggestion
    significant (> 1 min on my machine when iterating over all models).
```

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -735,6 +737,21 @@ def test_evaluate_xspec_model(modelcls):
 
     Convolution models are skipped (easier to filter out here given the
     current design).
+
+    This test checks two independent things:
+    1. The model can be evaluated with the default parameters
+       and the results agree and on energy and wavelength grids.
```
DougBurke, line 743:
```suggestion
       and the results agree using energy and wavelength grids.
```

Not sure what you meant `and on` to be here?

sherpa/astro/xspec/__init__.py
hamogu: It's part developer documentation to explain *why* we do this. I'm thinking of another person reading this a few years from now when the CXC is gone and none of the current developers are around to ask questions or a user who studies how we implemented it, maybe even because they want to do something similar in their code wrapping XSPEC in Julia or  whatever the next fancy language after Python will bei.  I can remove it, but a few lines of comment also don't cost us much.

sherpa/astro/xspec/__init__.py
hamogu: No, you are all correct. It's just that the "p" in `calc(self, p, *args, **kwargs)` is not a parameter, it's x0, which is confusing to me. But I promise I'll just rename it and that's fine, too.

hamogu: `p` just seems like an usual way to abbreviate `x0`.

hamogu: Or maybe I'm confused. Either way, it works and I'll just change the name. I don't do anything with it at that location anyway.

sherpa/astro/xspec/__init__.py
```diff
@@ -104,11 +104,12 @@
 from __future__ import annotations
 
 from contextlib import suppress
+import functools
```
hamogu, line 107:
```suggestion
from collections.abc import Callable
import functools
```

sherpa/astro/xspec/__init__.py
hamogu:
```suggestion
from typing import Any, overload
```

sherpa/astro/xspec/__init__.py
```diff
@@ -1791,6 +1836,10 @@ def guess(self, dep, *args, **kwargs):
         norm = get_xspec_norm(dep, self(*args))
         param_apply_limits(norm, self.norm, **kwargs)
 
+    @eval_xspec_with_fixed_norm
+    def calc(self, p, xlo, *args, **kwargs):
+        return super().calc(p, xlo, *args, **kwargs)
```
hamogu, line 1841:
```suggestion
    def calc(self, p, xlo, *args, **kwargs):
        return super().calc(p, *args, **kwargs)
```

sherpa/astro/xspec/__init__.py
DougBurke: `p` should not be `x0` - the first argument for `calc` is always the parameter list. Now, the `__call__` version does ot take parameters, but that's a different routine/story.

DougBurke:
My fix for the test failure (which also happens in #2275 for the same reason as it does here) is

https://github.com/sherpa/sherpa/blob/b323bac9a73ac6be6c17cd77ebfe7b407726d2bf/sherpa/astro/xspec/tests/test_xspec.py#L290-L304

DougBurke:
So, I just tried rebasing this onto main (just before #2281 was merged) and the "runtime cache" tests now fail

```
% pytest sherpa/astro/xspec/tests/test_xspec_caching_performance.py --runspeed
...
=================================================================== 196 failed, 84 passed, 1 warning in 78.18s (0:01:18) ====================================================================
... annoying xspec model screen output that I've never had time to understand
```

The failures look like

```
________________________________________________________________ test_evaluate_additive_xspec_model_normwrapper[XSzpowerlw] _________________________________________________________________
...
        # The norm=1 case has been run for initialization.
        # So, this should pull the value from the cache and be very fast.
        start_time_decorated = time.perf_counter()
        evals = mdl(elo, ehi)
        end_time_decorated = time.perf_counter()
        assert mdl._cache_ctr['hits'] == 1
        # Now, modify the model to remove `sherpa.astro.xspec.eval_xspec_with_fixed_norm`
        # decorator and empty the cache to make sure we are not using the cached value.
        mdl.calc = types.MethodType(inspect.unwrap(mdl.calc), mdl)
        mdl.cache_clear()
        start_time_clean = time.perf_counter()
        evals_no_wrapper = mdl(elo, ehi)
        end_time_clean = time.perf_counter()
        assert mdl._cache_ctr['hits'] == 0
>       assert len(mdl._cache) == 0
E       AssertionError: assert 1 == 0
``` 

So it is either all the `test_evaluate_additive_xspec_model_normwrapper[xxx]` cases, or a large number of them.

Given that a merge or rebase is now needed (presumable because of #2281) I think it's worth doing a rebase and looking at this problem.

What I really wanted to see was whether the runtime behaviour had changed because of this PR, but got stymied by this failure.

hamogu:
I'll get to this this afternoon when I can see we'll enough to read code again (ophthalmology appointment this morning). I would love to get this in,  and it's very close I think.

hamogu:
I'm not sure I need https://github.com/sherpa/sherpa/pull/2268#issuecomment-2824189881 because `XSbbody` now defaults to "no caching" which means that the model is evaluated in XSPEC every time and the original Float32 is passed through from XSPEC because the "special norm treatment" is not done for that particular model. So, I'll leave it for #2275 to address that.

hamogu:
Tests still running, but hopefully good now. See [message of new commit](https://github.com/sherpa/sherpa/pull/2268/commits/30399b8a4f1fba7305d6a02b1022763a15371cee) for details and explanation why I believe the tests are correct the way they are now.

DougBurke:
I've picked up the new version and I still see timing failures, just a lot less of them

```
% pytest sherpa/astro/xspec/tests/test_xspec_caching_performance.py --runspeed | grep ______
___________ test_evaluate_additive_xspec_model_normwrapper[XSbbody] ____________
__________ test_evaluate_additive_xspec_model_normwrapper[XSbbodyrad] __________
___________ test_evaluate_additive_xspec_model_normwrapper[XSdiskpn] ___________
___________ test_evaluate_additive_xspec_model_normwrapper[XSexpdec] ___________
____________ test_evaluate_additive_xspec_model_normwrapper[XSnsa] _____________
__________ test_evaluate_additive_xspec_model_normwrapper[XSnsagrav] ___________
__________ test_evaluate_additive_xspec_model_normwrapper[XSoptxagn] ___________
__________ test_evaluate_additive_xspec_model_normwrapper[XSoptxagnf] __________
__________ test_evaluate_additive_xspec_model_normwrapper[XSpegpwrlw] __________
____________ test_evaluate_additive_xspec_model_normwrapper[XSposm] ____________
___________ test_evaluate_additive_xspec_model_normwrapper[XSredge] ____________
___________ test_evaluate_additive_xspec_model_normwrapper[XSrefsch] ___________
____________ test_evaluate_additive_xspec_model_normwrapper[XSstep] ____________
___________ test_evaluate_additive_xspec_model_normwrapper[XSzbbody] ___________
```

no time to investigate tonight

DougBurke:
Looks like it's both the cache and  a few time checksL

```
% pytest sherpa/astro/xspec/tests/test_xspec_caching_performance.py --runspeed --lf | grep '>' | grep assert
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert run_time_decorated < run_time_clean
>       assert run_time_decorated < run_time_clean
>       assert mdl._cache_ctr['hits'] == 1
>       assert mdl._cache_ctr['hits'] == 1
>       assert run_time_decorated < run_time_clean
>       assert run_time_decorated < run_time_clean
```

hamogu:
I suggest to ignore all `assert run_time_decorated < run_time_clean` failures for this PR. As we discussed before, speed benchmarking is not an exact science. The  results may change with the grid size, the OS, and the weekday. In #2274 I changed only those models where `cache=0` is "obviously" better, i.e. where the test always has the same result, even if I re-run it several times, and for several different grid sizes. In some cases, switching the cache on is "obviously" better and then there are a few cases that are edge cases; and I left both, edge cases and "caching is obviously better", on the default of `cache=5`.

For those edge cases, one really has to do a judgement call. What's the typical use of this model in Sherpa? Will this be used with small or large grids? Do the benchmarking results agree for different computers? So for models where cache and no cache are within 10% or each other or where I observed wide swings in runtime for no obvious reason, I left the cache on by default in #2274. 
Now, I personally think that in those edge cases it might be beneficial to switch the cache off, because the cache not only has a runtime, but also a memory cost, but I wanted to make #2274 simple, by changing only the "obvious" parts. 
We can have a separate PR for the edge cases and discuss in more detail which ones should default to not cache.

(That's also why I initially did not write that as a test to be run with a simple option, but in a way that required a developer to actually look at and edit the source code to make the speed tests run because in that case, the developer would likely read the long explanation of what those "speed tests" can and cannot do.)

hamogu:
The other failures are "test setup" failures. The test code can't evaluate the time for a run with a cache for models that don't have a cache by default, which now are a few. I did not investigate those any further, because I knew that they would fail (because the runtime with cache is longer), when I should have noticed that the failure happens a few lines above (because the runtime with cache can't even be calculated by default).

The question here is what we want though. I'll pus ha fix that will set the cache values explicitly, so that the timing comparison will work for every model. That means that those same tests will continue to fail though, but now they will fail with "assert run_time_decorated < run_time_clean" instead.

Option 1: On the one hand, that's a failure of cache speedup as in "caching slows this down", but on the other hand, that's not quite fair, because we already set `cahce=0`. 

Option 2: So, I could change the condition to "if cache=0 by default, pass is uncached is faster, if cache > 0, pass if cached it faster". Maybe that's closer to how a test should be working (i.e. pass by default), but it does make it a bit harder to follow. Some logging of the runtimes would help people to look at it. (Did I say that these "tests" don't really fit neatly into the pattern of normal testing?), but that's not for this PR.

hamogu:
OK, I pushed a new commit (currently the last one in this PR to switch off cache for three additional XSPEC models

This leaves `XSrefsch` as the sole model that fails the speed test in `test_xspec_caching_performance` on my machine. For that model, I don't understand if the XSPEC model is analytical, uses tables or something else. I'm afraid that the run-time might depend on the parameters of the model and I'm not an expert in that area. For example, for certain settings of inner disk radius and temperature there might be ionic transitions that contribute to the spectrum (slow), for other settings it might be a much simpler continuum (fast). I don't know the model well enough to see what that parameter space is and the number of parameters is so large that a it's not a simpler matter to try that out. Since caching is only somewhat slower for the default values (which are set to be the range typically observed for most models), I feel it's safer to leave caching on in case the run-time does depend on the parameters.

DougBurke:
So, with the latest commit `8415f750` I get speed failures with

```
% pytest sherpa/astro/xspec/tests/test_xspec_caching_performance.py --runspeed | grep ___
___________ test_evaluate_additive_xspec_model_normwrapper[XSagauss] ___________
___________ test_evaluate_additive_xspec_model_normwrapper[XSdiskpn] ___________
__________ test_evaluate_additive_xspec_model_normwrapper[XSezdiskbb] __________
____________ test_evaluate_additive_xspec_model_normwrapper[XSnsa] _____________
__________ test_evaluate_additive_xspec_model_normwrapper[XSnsagrav] ___________
___________ test_evaluate_additive_xspec_model_normwrapper[XSrefsch] ___________
____________ test_evaluate_additive_xspec_model_normwrapper[XSstep] ____________
```

which we can see are speed test failures

```
% pytest sherpa/astro/xspec/tests/test_xspec_caching_performance.py --runspeed | grep '>' | grep assert
>           assert run_time_decorated > run_time_clean
E           assert 2.8217000362928957e-05 > 4.274499951861799e-05
>           assert run_time_decorated > run_time_clean
E           assert 2.6477002393221483e-05 > 4.4423999497666955e-05
            assert run_time_decorated > run_time_clean
>           assert run_time_decorated < run_time_clean
>           assert run_time_decorated > run_time_clean
E           assert 2.824200055329129e-05 > 3.152099816361442e-05
>           assert run_time_decorated > run_time_clean
E           assert 4.107999848201871e-05 > 4.2030002077808604e-05
>           assert run_time_decorated > run_time_clean
E           assert 2.409000080660917e-05 > 0.00012824900113628246
            assert run_time_decorated > run_time_clean
>           assert run_time_decorated < run_time_clean
```

For reference this is with XSPEC 12.14.0i from our conda xspec modelsonly package.

So, this is going back to the point I made about the fact that there are architectural / OS / version / is-it-Tuesday issues, where it's going to be hard to identify exactly which models benefit (my bet, completely plucked from the air, is that the ARM machines may do better at calculating the cache key than the x86_64 machine I have, but maybe it's the other way around, and it's not really relevant here anyway).

We do have to have some way of identifying what is good or not to cache (unless we go with no cache or everyone gets a cache), and this code and set of tests provide us with ways of checking. We can - in the infinite glorious future - think about ways to better quantify the cache cost and whether we should provide users with a way to check on their machines, but that's all for future us™.

I don't think we need to address these `--runspeed` discrepancies in this PR.

sherpa/astro/xspec/__init__.py
```diff
@@ -103,7 +103,10 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from contextlib import suppress
+from collections.abc import Callable
```
DougBurke, line 108:
```suggestion
```

Unless GH is lying to me about line 106 ....

DougBurke (approved):
I have two suggestions for removing imports, but they are not critical.

sherpa/astro/xspec/__init__.py
hamogu: That's an error from the rebase. Sorry.

PR #2271, opened 2025-04-11
Improve test coverage of the optimization code

DougBurke:
# Summary

Improve test coverage for the optimization code. There is no functional change to Sherpa.

# Details

These are tests that were suggested when working on #2022. It makes sense to take them now before that work gets finished. There is also a typo in a comment in `sherpa/fit.py` that I included.

The tests generally extend the checks of the "rng/seed" handling of the optimization code (this was changed in #1735 to better support the modern NumPy RNG code, but this code is complicated to test as it was originally written with a combination of an explicit seed value and also relying on the "system" [1] random state). So whilst reviewing this code I've realized we can improve our test coverage a little bit in a few places.

The optimization code has multi-core support which we haven't really tested in our CI runs. The CI runners on GitHub now have multi-core capabilities, so we can add some explicit checks with `numcores=1, 2, 3` and compare the results. This shows some "interesting" behaviour, in that the results can depend on OS/architecture and it's not clear why. So these tests are really thought of as regression tests, just to check that the code continues working as expected, rather than being a set of tests from "first principles" (e.g. where we know the correct answer).

[1] one thing I had to do as part of #1735 was determine whether the "random state" was coming from `np.random` or `random` and try to rationalize the code so the randomness could be better understood

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -4378,7 +4378,7 @@ def setup_multicore_data(s: Session) -> tuple[ArithmeticModel, ArithmeticModel]:
 #
 @pytest.mark.skipif(not multi, reason="multi-core support not enabled")
 @pytest.mark.parametrize("session", [pytest.param(Session, marks=pytest.mark.session), AstroSession])
-@pytest.mark.parametrize("ncores", [1, 2, 3])
+@pytest.mark.parametrize("ncores", [1, pytest.param(2, marks=pytest.mark.cores), 3])
 def test_method_numcores_levmar(session, ncores):
     """Check we can use numcores>1 with levmar"""
 
@@ -4411,7 +4411,7 @@ def test_method_numcores_levmar(session, ncores):
 
 @pytest.mark.skipif(not multi, reason="multi-core support not enabled")
 @pytest.mark.parametrize("session", [pytest.param(Session, marks=pytest.mark.session), AstroSession])
-@pytest.mark.parametrize("ncores", [1, 3])  # save some time not using ncores=2
+@pytest.mark.parametrize("ncores", [1, pytest.param(2, marks=pytest.mark.cores), 3])
 def test_errors_numcores_levmar_proj(session, ncores):
     """Check we can use numcores>1 with proj"""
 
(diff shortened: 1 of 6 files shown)


sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -4317,3 +4320,255 @@ def test_dataspace1d_datapha_invalid_args(start, stop, step, numbins):
...
+    g1, g2 = setup_multicore_data(s)
+
+    s.set_method_opt("numcores", ncores)
+
+    # The random state is taken from the rng parameter of the
+    # optimizer, which does not fit in well with the current design
+    # (Sherpa 4.16).
+    #
+    # rng = np.random.RandomState(8273)
+    # s.set_rng(rng)
```
DougBurke, line 4564:
```suggestion
```

DougBurke:
@hamogu - do you think you will have time to review this highly-informative PR (it's pulled out of some code I am taking too long to get finished and I'd like to get this tested on DS/SDS builds).

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -4317,3 +4320,249 @@ def test_dataspace1d_datapha_invalid_args(start, stop, step, numbins):
...
+
+    return g1, g2
+
+
+# Note: gridsearch is too slow to test this way
+#
+@pytest.mark.skipif(not multi, reason="multi-core support not enabled")
+@pytest.mark.parametrize("session", [Session, AstroSession])
+@pytest.mark.parametrize("ncores", [1, 2, 3])
+def test_method_numcores_levmar(session, ncores):
```
hamogu, line 4382: This test is also redundant with `test_errors_numcores_levmar_proj` which also generates data and fits it (and then runs `proj` after that. In the interest of run time I suggest to merge the tests - just copy the assert statements for here into the other test and this one won't be needed any more.
One could argue similarly for the `conf` test, but maybe run one of them on `Session` and the other on `Astrosession` so that we have two separate tests for levmar in the end.

Due to the duplication and heavy parameterization, these new tests increase the CI runtime by almost 10%, which seems like a lot give how similar e.g. the code in `Session` and `AstroSession` is.

DougBurke:
Argh - the tests won't run because of the conflict ..... I may try a merge here just to get a "easier to follow" history and then rebase before final merge. Or not.

DougBurke:
So, I've been able to merge in the current branch. I'm about to rebase it so it's a bit nicer, and let me check the new `--runsession` and `--runcores` markers (well, to see if coverage drops when we add them).

DougBurke:
This rebase - up to https://github.com/sherpa/sherpa/pull/2271/commits/69232c4fad77ede982d82bda7f58b1292bbd3fe0 - takes us to the code that Moritz reviewed in https://github.com/sherpa/sherpa/pull/2271#pullrequestreview-2774228765 (i.e. the rebase didn't change anything in these commits other than the commit hash.

hamogu (approved):

PR #2273, opened 2025-04-12
Make sure all datastack-wrapped functions are added to namespace

hamogu:
## Summary
Complete the `datastack` namspace for functions wrapped to operate on datastacks

## Details
 The datastack modules used a decorator called `public` that would add a decorated function to `__all__` to save the programmer to type it out as a string. That decorator also did some checks to ensure that a function was not added twice, which then cause functions are are aliases to other functions not to be added.
The first commit replaces that machinery with direct string access to `__all__`.

The second commit removed the new unsued `public` from utils. It's not used in other parts of sherpa. Technically, `public` could have been used outside of Sherpa by users, but I hope not. (The linked Python issue that is discussed in the now removed code has a lot of voices saying why that might be a bad idea.)

### DIFF
--- a/sherpa/astro/datastack/tests/test_datastack.py
+++ b/sherpa/astro/datastack/tests/test_datastack.py
@@ -812,12 +812,9 @@ def test_operations_datastack_group(ds_setup, ds_datadir):
 def test_wrapped_funcs_in_all():
     """This is a regression test for #2270
 
-    Before #2270 several convenience functions using different
-    types of magic were used and interacted such that not every
-    wrapped function was added to __all__, in particular when one of them
-    is an alias for another. It's very unlikely that that same failure
-    mode will happen again, so this is a very rudamentary test where we are
-    just checking that both functions exisit.
+    Check that aliased routines (here set_model/set_source) are present
+    both in the module and are exported by it. It is not worth checking every
+    possible symbol.
     """
     func = datastack.set_model
     func = datastack.set_source
--- a/sherpa/astro/datastack/tests/test_datastack.py
+++ b/sherpa/astro/datastack/tests/test_datastack.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2014, 2015, 2016, 2018, 2019, 2020, 2021
+#  Copyright (C) 2014-2016, 2018-2021, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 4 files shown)


hamogu (commented):
CC: @taldcroft as the original author. I think this comes from his early phase as a Python programmer where he was sometimes a little too excited by all the magic that's possible.

taldcroft:
@hamogu - hah, blast from the past. And yes, I probably wouldn't write that today. OTOH just last week I decided to challenge my brain writing a descriptor in a dataclass with a metaclass to auto-generate class properties. It took a while to get it straight and I'll probably regret it in the end, but I spent so long doing it that I couldn't just let it go. Sunk-cost coding.

sherpa/astro/datastack/tests/test_datastack.py
```diff
@@ -807,3 +807,19 @@ def test_operations_datastack_group(ds_setup, ds_datadir):
...
+
+def test_wrapped_funcs_in_all():
+    """This is a regression test for #2270
+
+    Before #2270 several convenience functions using different
+    types of magic were used and interacted such that not every
+    wrapped function was added to __all__, in particular when one of them
+    is an alias for another. It's very unlikely that that same failure
+    mode will happen again, so this is a very rudamentary test where we are
+    just checking that both functions exisit.
```
DougBurke, line 820:
```suggestion
    Check that aliased routines (here set_model/set_source) are present
    both in the module and are exported by it. It is not worth checking every
    possible symbol.
```

I don't think we need the history.

We *could* make this a parametrized test (and test more symbols) by sending in the name to test, but - as you said in the now-deleted comment - it's not worth it.

DougBurke (approved):
LGTM. This is technically a breaking change, but unlike `_use_caching` (a recently-discussed symbol) I don't care about its removal, since it's even-less-likely-to-be-used™.

hamogu:
Well, any bugfix is a breaking change: https://xkcd.com/1172/

The code was not behaving as it should have and this PR fixes it.

hamogu:
@wmclaugh The docs built fails due to a timeout when pulling the references to the bokeh docs from the web. There is no change to the docs in here and (a) the docs still pass locally and (b) they passed earlier (school up - all tests passed for commit 524fcce.

Unfortunately, I don't see a button to restart just the readthedocs test, so I suggest to use our judgement and proceed as if it had passed. (Alternatively, I could fix a typo in some unrelated file and push again to trigger the tests again, but that seems slightly silly.)

Issue #713, opened 2019-11-26
Fitting PHA: Sherpa models integrate, but don't work


hamogu:
I'm fitting X-ray data (PHA) files, fitting just a constant:
```
load_data("pha2.gz")
load_arf(1, "leg_-1.arf.gz")
load_rmf(1, "leg_-1.rmf.gz")
set_analysis("wave")
ignore(None, 21.5)
ignore(22.4, None)

set_source(1, const1d.c1)
fit(1)
plot_fit(1)
```
However, the fit is terrible and the plot indicates that it did not work.
![integrateTrue](https://user-images.githubusercontent.com/498688/69652761-269d0f00-1040-11ea-8968-1bb73c74b2cf.png)

However, when I turn integrate off, it suddenly works well:
```
c1.integrate=False
fit(1)
plot_fit(1)
```
![integrateFalse](https://user-images.githubusercontent.com/498688/69652845-4af8eb80-1040-11ea-9fe4-71c1e6c5988b.png)

That's curious, because PHA data is integrated.
As a side note: using xspec models (`xsconstant`) works out of the box without meddeling with the `integrate` flag.

Seen with the sherpa that comes with CIAO 4.11 and CIAO 4.12b2.
Data to reproduce on local HEAD LAN for sherpa developers: `/pool14/guenther/sherpa_bug_integrate`

hamogu:
I see that the fit in the top plot is not a flat line, i.e. is takes the arf and the rmf into account. Not sure if that's a case in the second plot. S, this might well turn out to be a user error instead of a bug, but as a user, I don't know what's going wrong.

DougBurke:
Moritz,

don't have time to discuss this at the moment, but I find there are issues
with understanding the constant model with integrate=True. For non-grating
Chandra PHA files we recommend an energy gridding for the responses with a
constant bin width - normally 0.01 keV - which can lead to a factor of 100
in normalization. If the bin-width is not constant (this is of the
response, not groups) then you can get "garbage" fits.

Doug

dtnguyen2:
Hi Moritz
Finally got a chance to look at this issue but unfortunately the files in the ```/pool14/guenther/sherpa_bug_integrate``` directory are gone.  Please re-populate them and thanks in advance.

dtnguyen2:
@wmclaugh showed me how I can get the files from the snapshot view so need to re-populate the files.

hamogu:
[...]

Issue #340, opened 2017-02-24
The parameters for sherpa.models.basic.Exp are badly defined


DougBurke:
The current `sherpa.models.basic.Exp` model defines the exponential term as `coeff * (x - offset)` which I believe leads to poor behavior. I get better results with `a + b * x` (the names can be changed ;-). My assertion is that the current parameter choice leads to "linked" parameters, which confuses the optimiser, since expanding out the term gives `coeff * x - coeff * offset` (i.e. coeff appears in both terms). Note that I can get the same fit with both versions, but the version in Sherpa requires more hand holding to get the answer. I have *not* tried to work out how sensitive this analysis is to the input data (i.e. did I just happen to hit upon the one problem case, or is this a general issue for this parameterisation).

Details, including an ipython notebook showing a fit which doesn't work with the current model but does with the proposed new version is available at https://gist.github.com/DougBurke/32ac9daa785f95bcbe4a3fdf7a36b86b

It is simple to implement; the main thought would be whether this replaces the `Exp` class (leading to incompatability problems) or a new class, perhaps deprecating the old one. Perhaps there's a better way?

DougBurke:
Given that

    c exp(a + bx) = c exp(a) exp(bx) = d exp(b x)

it's not clear that the model should have a separate amplitude parameter (which is `c` above), since it's degenerate with `a`.

DougBurke:
Also true for the `Exp10` model.

PR #2274, opened 2025-04-14
Set `cache=0` by default for models that are fast to calculate

### LINKED ISSUE #2258: Caching `Scale1D` is actively harmful
I have used `Scale1D` models to mimic a `norm` parameter to speed up computations and often that works because evaluating the scale is way faster than calling into XSPEC C or fortran code, in particular if that code does some complicated calculation itself, see #767 for the original idea and a long discussion of pros and cons and [as a suggestion to a user](https://github.com/sherpa/sherpa/issues/442#issuecomment-1825612981) who [successfully sped up his fits](https://github.com/sherpa/sherpa/issues/442#issuecomment-1825643997).

However, I've just found that we have to be careful in timing comparisons, because `Scale1D` also employs caching and that slows it down by a lot (relatively speaking).

 Here, we just want to fit a scalar scale factor. There is nothing faster then just multiplying with a scalar. However, in the internal hirachy of models `Scale1D` derives from `Const1D` and has a special effected if the `integrate` setting is set to true (which is hopefully isn't because that would mess up the fit). `Scale1D` inherits the cache setting from `Const1D` and now needs to calculate the hash for every call, if it's a "hit" or not, slowing down the model by about a factor of 6 as we can see below. While it's not dominating the overall fit, it's noticable in the run-time for a model where I use the scale to replace the norm.

Using IPython's magic commands:
```Python
from sherpa.models.basic import Scale1D
newscale = Scale1D("newscale")
%timeit newscale(xlo=np.arange(1000), xhi=np.arange(1000) + 1)
newscale._use_caching = False
%timeit newscale(xlo=np.arange(1000), xhi=np.arange(1000) + 1)
```
prints
```
29.7 μs ± 205 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
5.67 μs ± 50.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
In other words, the caching here gives a slow-down of a factor of 5-6.

While the absolute runtime is still small, I can see that adding a `Scale1D` slows models down more than it should. #2256 will help with this, but in this case, no caching is even better than faster caching (and also eliminates the memory use associated with a cache).

hamogu:
## Summary

Based on benchmarking for typical Sherpa use cases, some models are faster to re-calculate than to cache. For those models, the default cache size it now set to 0 to save runtime and memory.

## Details
This also adds a new mark "speed" to select speed tests to run.
That test is written for XSPEC modes, but similar benchmarking has been done for basic Sherpa models (see #2258).

fixes #2258

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -883,6 +883,14 @@ available.
    The ``sherpa/astro/xspec/__init__.py`` file also lists the supported
    XSPEC versions.
 
+#. Should new XSPEC models use caching?
+
+   By default, the generated code does not say anything special about caching, so new
+   models will be cached as usual. Once the models are added, you can run
+   ``pytest --run-speed sherpa/astro/xspec/tests/test_xspec_caching_performance.py``
+   to see if the models are cached. If caching slows down the run, the test may fail.
+   See notes in ``sherpa/astro/xspec/tests/test_xspec_caching_performance.py`` for details.
+
 Never forget to update the year of the copyright notice?
 --------------------------------------------------------
 
--- a/docs/evaluation/cache.rst
+++ b/docs/evaluation/cache.rst
@@ -13,16 +13,17 @@ uses the parameter values, evaluation grid, and integrate setting to
 look for a value from that model's cache. If found the value is returned,
 otherwise the model is evaluated and the result is added to the cache.
 
-Unfortunately it is not always obvious if a model uses caching, or how
-effective it is.
+It is hard to predict how effective caching is so
+the defaults are set for typical Sherpa use cases, based on some benchmarking,
+but your performance might be improved with different settings.
 
 What models are cached?
 =======================
 
-There is unfortunately no easy way to determine whether a model
-uses the cache without either viewing the model definition - looking
-for the application of ``@modelCacher1d`` to the ``calc`` method - or
-by running a test as shown below,
+A model uses the cache if ``@modelCacher1d`` is applied to the ``calc`` method
+**and** `model.cache` is set to a positive integer.
+Unfortunately it is not easy to check weather ``@modelCacher1d`` is applied without
+looking at the source codeb- or by running a test as shown below,
 :ref:`in the example section <example-modelcacher1d>`.
 
 When is the cache useful?
@@ -35,8 +36,15 @@ Compared to most built-in sherpa models, models in the optional XSPEC model
 library (:py:mod:`sherpa.astro.xspec`) tend to be more complex and
 thus benefit more from caching.
 
-Can I turn off this behavior?
-=============================
+By default, the cache is switched off (`mdl.cache=0`) for simple models where the model
+evaluation is fast, sometimes even faster than the hashing that is needed to look up
+a value in the cache. An obvious example is a scale or constant model
+(`sherpa.models.basic.Scale1D` or `sherpa.models.basic.Constant1D`),
+but this also applies to some fast analytical
+models such as an XSPEC black body (`sherpa.astro.xspec.XSbbody`).
+
+Can I turn off this behavior for other models?
+==============================================
 
 The size of the cache for a specific model component called ``mdl`` can
 be set to zero (``mdl.cache=0``) to turn off the cache behavior.
@@ -54,6 +62,40 @@ leaves it at the previous setting. This is because some models may not work with
 caching at all and need to stay at ``cache=0`` at all times.
 The cache has to be manually set to a positive number for all models that should use the cache
 to allow caching again.
+
+How do I set the default in my own models to use or not use the cache?
+======================================================================
+
+In order to be cacheable at all, a model must have the ``@modelCacher1d`` decorator
+applied to the ``calc`` method. The default for an `~sherpa.models.model.ArithmeticModel`
+is a cache size of 5, but this can be changed by setting the
+``cache`` attribute in the model's
+constructor. For example, here we deactivate the cache by default by setting the value to 0,
+but we still decorate the ``calc`` method so that a user can switch if back on for
+individual instances of the model::
+
+    >>> from sherpa.models.model import ArithmeticModel, modelCacher1d, Parameter
+    >>> class MyModel(ArithmeticModel):
+    ...     def __init__(self, name='mymodel'):
+    ...         self.xpos = Parameter(name, 'offset', 0)
+    ...         super().__init__(name, (self.offset,))
+    ...         self.cache = 0
+    ...
+    ...     @modelCacher1d
+    ...     def calc(self, p, *args, **kwargs):
+    ...         # do something
+    ...         return p[0] + args[0]
+    >>> m = MyModel()
+    >>> m.cache = 3  # use the cache in models instance m
+
+Do **not** set `cache = 0` as a class attribute. If you use `~sherpa.models.model.modelCacher1d`
+``cache`` is actually a property that does other things when the cache is set (e.g. reset the
+cache content). Setting a class attribute will lead to errors when the decorated ``calc`` method
+is called, i.e. the following will not work::
+
+    >>> class MyModel(ArithmeticModel):
+    ...     cache = 0  # DO NOT DO THIS!
+
 
 How does the cache work?
 ========================
@@ -99,15 +141,19 @@ case it isn't worth it!). First we set up the data::
 
     >>> import numpy as np
     >>> from sherpa.data import Data1D
-    >>> x = np.arange(1, 4)
-    >>> y = [4, 5, 2]
+    >>> x = np.arange(0, 3)
+    >>> y = [2, 0.3, 0.02]
     >>> data = Data1D('example', x, y)
 
 A simple model is used::
 
-    >>> from sherpa.models.basic import Const1D
-    >>> mdl = Const1D()
-    >>> print(mdl.c0.val)
+    >>> from sherpa.models.basic import Exp10
+    >>> mdl = Exp10()
+    >>> mdl.offset.frozen = True
+    >>> mdl.offset = 1.0
+    >>> mdl.coeff.frozen = True
+    >>> mdl.coeff = -1.0
+    >>> print(mdl.ampl.val)
     1.0
     >>> print(mdl._cache)
     {}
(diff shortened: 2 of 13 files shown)


hamogu:
@DougBurke : Here, I've edited the XSPEC models directly. Will that lead to any difficulty with autogenerating them from the `models.dat` file?

docs/evaluation/cache.rst
```diff
@@ -122,26 +168,23 @@ the `startup` and `teardown` methods are called automatically by
...
     ...
-    [1. 1. 1.]
-    [1.00034527 1.00034527 1.00034527]
-    [3.39944171 3.39944171 3.39944171]
-    [3.40061543 3.40061543 3.40061543]
-
+    [10.   1.   0.1]
+    [10.00345267  1.00034527  0.10003453]
+    [2.01694277 0.20169428 0.02016943]
+    [2.01763916 0.20176392 0.02017639]
```
DougBurke, line 190: I was going to ask why you'd changed this, but I see now that `Const1D` is now "cache averse", so you needed a different model.

I do note that I am not happy with the `Exp/Exp10` models - see #340 - but as I wrote that issue 8 years or so ago I guess we are not going to change them now!

DougBurke:
I am having a moment of `dumb` here - why do the tests in `sherpa/astro/xspec/tests/test_xspec_caching_performance.py` not get picked up by `pytest` by default?

Oh, I see now. It's the test name.

I think I'd prefer a solution like the `runslow` and `runzenodo` flags - so

https://github.com/sherpa/sherpa/blob/cfe28f9915d47cafe3df561799bddb19e451fefc/sherpa/conftest.py#L104-L119

and then

https://github.com/sherpa/sherpa/blob/cfe28f9915d47cafe3df561799bddb19e451fefc/sherpa/tests/test_sherpa.py#L84-L85

Then it's "just" a case of coming up with a `--runXXX` name for `XXX`!

DougBurke:
For https://github.com/sherpa/sherpa/pull/2274#issuecomment-2800352943

The automated code has no way to know whether a new model should be cached or not [*], and so the text it generates will just not include the `self.cache = 0` line. The user would then have to run the tests with the `--runXXX` flag to check (once the model is built/working) and add the line themselves. Which is okay. We could add something to the developer documentation about this.

[*] this would actually be the largest conceptual point here - and I'll address this in a follow-up comment (if only GH had some way to thread the responses)

hamogu:
All those uncertainties are why I'm **not** setting the tests to run every time. I've tried different days of the week (not Tuesdays though, only Thursday, Friday, and Sunday, I think  ;-) ) and did the timing a few dozen times, but yes, strict performance testing is hard, especially if we don't have a complete list of all the data sizes and model values that users use in practice. So, we do have to use a combination of numerical testing and common sense:

For example, multiplying with a number (`Scale1D`) does not take (measurably) longer for different numbers. It's all floating point multiplication. The same is true for simple function evaluations, e.g. a powerlaw. There might be a slight dependence on the exponent but (a) nobody fits an XSPEC model with an exponent of 1e+38 and if you do, the fit will fail very quickly anyway and (b) for simply mathematical formulas that dependence is going to be very, very little.
I looked at the description in XSPEC of all the models that I'm changing the default for and they are almost exclusively simple mathematical formulae (powerlaw, constanst, black body), where I understand that nothing changes fundamentally depending on the scale of the input parameters; and for the sherpa models, it's a constant, a delta function, a box etc. - again models where I understand how they work and know that the range of parameter values does not change the way they are evaluated. 
That's different from models that access underlaying databases to do complex atomic calculations (APEC, absorption models) or ANG models that evaluate anything in GR. For those classes of models, the runtime will often depend on the parameter values (e.g. there are more atomic transitions at lower energies than in a very hot plasma when most elements are fully ionized).

Today, OS's and compilers are pretty similar e.g. we are all using 64 bit systems with so much memory and L2 cache in the CPU that it does not matter for the simple models where I make changes. Most differences between computers and OS'es are going to be relative, i.e. both the cached and uncached version will be faster or slower, but it's unlikely that the cached version becomes slower, while the uncached version becomes faster.

**Most importantly**, I want to point out that we are not going from "a well thought out choice of cache" to "some haphazard thing a single developer cooked up". We made a choice before (to cache everything), but (based on the comments in the code, e.g. for the hash function and all the docs we had before) that choice was based on zero data. So we really go from "randomly apply a cache where someone happened to type it" to "testing on at least one machine for different inputs dozens of times coupled with some analytical understanding".

So, is this guaranteed to be the ideal choice of default values for the cache, averaged Sherpa over all datasets and fits that all users do on different machines? Probably not. But is it significantly closer to that than what we had before? Certainly yes.

docs/evaluation/cache.rst
hamogu: I did not know about #340 - I think we should change it eventually if we take Sherpa's mission as a general modeling and fitting program seriously [*], but it is definitely not part of this PR.

*: While I don't have hard data to back it up, based on my interaction with users and looking at the occasional tickets we get, it seems that Sherpa is probably used 90% of the time X-ray data with XSPEC models, maybe 7% of the time for images and 3% of the time for other things, so the 1D models in `sherpa.models.basic` don't get exercised a lot. (#713 is another example of that.)

sherpa/astro/xspec/tests/test_xspec_caching_performance.py
```diff
@@ -33,17 +33,11 @@
...
-`do_not_test_evaluate_additive_xspec_model_normwrapper` to
-`test_evaluate_additive_xspec_model_normwrapper` and run pytest watching the
-results (possibly with a different number of points in `make_grid_dense`).
-
+To run tests in this file use `pytest --runspeed`.
 
 In tests on a Mac M1 system on Sherpa 4.17 we find the following models
 are consistently faster without the decorator:
-
+`
```
DougBurke, line 40:
```suggestion

```

I assume this is not meant to be?

docs/developer/index.rst
```diff
@@ -333,13 +333,9 @@ The ``sherpa/conftest.py`` file contains general-purpose testing
...
 
-* add to the ``pytest_addoption`` routine, to add the option;
+* to add generic options modify the ``pytest_addoption``;
 
-* add to ``pytest_collection_modifyitems`` if the option adds
-  a new mark;
-
-* and add support in ``pytest_configure``, such as registering
-  a new mark.
+* for new marks, add to ``markers`` in ``pytest.ini``.
```
DougBurke, line 338: I've never felt happy with how we/I documented this.

Your change says to add new markers to ``pytest.ini`` but below you change ``pytest_addoption`` in ``sherpa/conftest.py``.

So am I getting confused by what `mark` means?

hamogu: Yeah no. It's me. Pytest ha sa way [to add marks in pytest.ini](https://docs.pytest.org/en/stable/how-to/mark.html#registering-marks) which is simpler than what we do today and which I tried to use. However, by default it always runs all tests, unless a user types `pytest -m "not speed". In other words, ther is no way to set the default to False.
So, I reverted conftest, but not the docs.

DougBurke:
I think the two commits can be merged together.

There are also other places for models - I think I've marked off the ones you've "de-cached", but maybe you looked at the others and decided that they aren't simple enough:

- [X] `sherpa/models/basic.py`
- [X] `sherpa/astro/xspec/__init__.py`
- [ ] `sherpa/astro/models/__init__.py`  
- [ ] `sherpa/astro/optical/__init__.py`
- [ ] `sherpa/models/template.py`

I am okay with not looking at the `astro` and `template` versions here (I'm pretty sure that by default the template models are going to be complex enough that turning off the cache is not a useful option).

A follow up PR could be to mark these models as "always uncached" - e.g. special case the property access and make sure it can never be set - but it's ot obvious to me that that is worth it, and it's definitely not for this PR.

hamogu:
I wanted to err on the side of caution and only change things where no caching is clearly better. The `astro/models` models are all a bit more numerically complicated, in particular with `integrate=True` (which most of them don't handle well anyway), but I didn't run speed testing on them. I probably should, since I have the machinery ready in this PR.

hamogu:
I added the comment by @DougBurke above with the list of other models that I'm including in speed testing now. Turns out that most models in `astro.optical` profit from using the cache :-)

sherpa/astro/models/tests/test_astro_model.py
```diff
@@ -79,18 +79,12 @@ def test_create_and_evaluate(name, cls):
...
-                                 models.BPL1D,
+@pytest.mark.parametrize("cls", [
                                  models.Beta1D,
-                                 models.Edge,
                                  models.LineBroad,
                                  models.Lorentz1D,
-                                 models.NormBeta1D,
                                  models.PseudoVoigt1D,
-                                 models.Schechter,
-                                 models.Voigt1D])
```
DougBurke, line 93: So, why do these models that you remove fail? Oh, I guess it's just that because they have no cache that the `len(mdl._cache)` checks are no-longer valid.

There was a reason I wrote this test. If you added a test like

```python
@pytest.mark.parametrize("cls", [models.Atten, models.Schechter])  # just pick a few
def test_send_keyword_1d_no_cache(cls):
    """These models do not use a cache. Can we send an unsupported keyword?"""

    mdl = cls()
    x = [10, 21, 13, 15]
    y1 = mdl(x)
    y2 = mdl(x, not_an_argument=True)
    assert len(mdl._cache) == 0
    assert y2 == pytest.approx(y1)
```

Oh, but we essentially have this in `test_send_keyword_1d_no_cache`, so there's no need. But I've already written this comment so I'm going to comment it....

sherpa/models/tests/test_basic_cache.py
```diff
@@ -0,0 +1,109 @@
...
+modellist.sort()
+
+
+def get_runtime_with_cache(mdl, timing = False):
+    """Does the `sherpa.astro.xspec.eval_xspec_with_fixed_norm`
+    wrapper change the results?
+
+    It should not (within numerical tolerances).
+    The hard part in this test is to obtain a "clean" version of the
+    model without the wrapper.
```
DougBurke, line 57: 1. I think this comment is from a different PR
2. There's no point in having the `timing` argument as it's always `True` isn't it?

sherpa/models/tests/test_basic_cache.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    # run once in case there is a startup time for loading
+    x = np.arange(1, 100, 1000)
+    mdl.cache=5
+    evals = mdl(x)
+
+    # The norm=1 case has been run for initialization.
+    # So, this should pull the value from the cache and be very fast.
```
DougBurke, line 65:
```suggestion
```

Another case f the "wrong PR" ?

sherpa/models/tests/test_basic_cache.py
DougBurke: In fact, the routine could just be moved into the test, but I'm not sure it's worth it. 

I think we now have this logic in three or more tests, so could it be moved to `sherpa/conftest.py`? I suggest that would be a "for la ater" PR.

hamogu: I anticipate that one could run this in CI to test that with cache and without cache give the same results (but not run the timing). I don't do that right now, but we do have tests like that that could be refactored to use this routine when we touch them again to I wanted to stake that out by having the `timing` argument in there.

hamogu:
removed the comments that I accidentally copied from the XSPEC tests into the non-XSPEC tests.

DougBurke (approved):
LGTM - ta.

hamogu:
I simply clicked the wrong button... Sorry. Reopened. Thanks for catching @DougBurke !

PR #2275, opened 2025-04-14
Change xspec norm handling

DougBurke:
Change how we handle XSPEC normalization for additive models (including tables).


hamogu:
At least for normal additive models this change could replace #2268 entirely if we apply the `@modelCacher1D` to the `_calc` method (where the norm is already peeled off) instead of the `calc` method (where the norm is still part of the list of parameters).

The table models need a slightly closer look because they have different code path, but I think that a similar approach there should be possible, too. Just make sure that we split out the method that calls the XSPEC code (after peeling off the norm) as a separate method and then apply the case to that.

I thought about that for a#2268, but due to my lack of experience with our C++ interface, I selected a pure-python solution; however, I think the approach here is superior.

DougBurke:
Rebased as #2268 has been merged. This required changing how the code works so

- I have not scrutinized it as closely as I'd like
  - in particular, the changes in `test_xspec.py` need some thinking about to see it they are sensible
- I have not looked at the runtime characteristics

Maybe we need a "dump the evaluation times for models" script set up to help evaluate this.

DougBurke:
@hamogu - this does not need looking at right now (aka I'm about to work on something completely different so I'll just be putting my fingers in my ears and yelling "I can't hear you"), but I've reworked this to account for #2268. This was done quickly so may need re-thinking.

Two immediate thoughts:

1. the change to `test_xspec.py` probably needs some discussion, as it depends on what exactly you wanted to test; I couldn't quickly come up with a replacement for the class `calc` routine so just decided all I cared was that the model evaluation matched  "if you called this from XSPEC, this is what you get" - but maybe we want something else.
2. in my original version of this, I didn't call to `super().calc(...)` and so **potentially** saved a little but of time, at the expense of a) repeating code, and b) not gaining from the cache on the superclass. We could continue that here, but I think we probably prefer the simplicity of the current code rather than trying to micro-optimize the code here.

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -776,19 +780,17 @@ def test_evaluate_xspec_model(modelcls):
...
-    evals_no_wrapper = mdl(elo, ehi)
+
+    # 2) Now, modify the model to remove
+    # `sherpa.astro.xspec.eval_xspec_with_fixed_norm` decorator.
+    #
+    # This turns out to be hard to do, so instead just call the
+    # compiled function directly, which also avoids the cache code.
+    #
+    pars = [p.val for p in mdl.pars]
+    evals_no_wrapper = pars[-1] * mdl._calc(pars[:-1], elo, ehi)
```
hamogu, line 790: Is there a reason why you removed the `mdl.cache_clear()` line here? We may or may not apply that at the `_calc` level in the future (who knows), so better we sure this test doesn't accidentially usee the cache if if the code changes.

sherpa/astro/xspec/__init__.py
```diff
@@ -1863,43 +1867,41 @@ def mkabund(name: str,
...
+
+    Since the XSPEC additive models all have the form::
+
+        pars[-1] * calc(pars[:-1], *args, **kwargs)
+
+    then it is likely better to cache the results without including
+    norm parameter (the last parameter). This decorator takes
+    advantage of this by separating out the norm parameter from the
+    rest, and letting the cache code attached to the superclass
+    (XSModel.calc) handle the decison of what to cache.
```
hamogu:
```suggestion
    (XSModel.calc) handle the decision of what to cache.
```

sherpa/astro/xspec/__init__.py
```diff
@@ -1908,6 +1910,10 @@ class XSAdditiveModel(XSModel):
 
     The XSPEC additive models are listed at [1]_.
 
+    .. versionchanged:: 4.18.0
+       The "norm" parameter is now handled in the `calc` method rather
+       than in the compiled code.
```
hamogu, line 1877:
```suggestion
    .. versionchanged:: 4.18.0
       The "norm" parameter is now handled in the `calc` method rather
       than in the compiled code. (Or, to be more precise, in the 
       eval_xspec_with_fixed_norm decorator that is applied to the calc method.)
```

hamogu: Check if that's correct!

hamogu (changes requested):

sherpa/astro/xspec/tests/test_xspec.py
DougBurke: I'll need to check. This was originally written at the same time as #2268 was (to check some things) and merging the two may have lost something important down the back of the sofa.

sherpa/astro/xspec/__init__.py
DougBurke: Actually, do we need `eval_xspec_with_fixed_norm` decorator any more? Can we not just have

```python
        return p[-1] * super().calc(p[:-1], *args, **kwargs)
```

which will remove a layer of indirection.

DougBurke: I've made this change to see how it goes. It looks like the cache is still being used:

```
>>> from sherpa.astro.xspec import XSapec
>>> mdl = XSapec()
>>> mdl.cache_status()
 apec                       size:    0  hits:     0  misses:     0  check:     0
Reading APEC data from 3.0.9

array([2.10839183, 0.31196176, 0.22008776])
>>> mdl.cache_status()
 apec                       size:    1  hits:     0  misses:     1  check:     1
array([2.10839183, 0.31196176, 0.22008776])
>>> mdl.cache_status()
 apec                       size:    1  hits:     1  misses:     1  check:     2
>>> mdl.norm = 5
>>> mdl.cache_status()
 apec                       size:    1  hits:     1  misses:     1  check:     2
array([10.54195917,  1.5598088 ,  1.10043879])
>>> mdl.cache_status()
 apec                       size:    1  hits:     2  misses:     1  check:     3
```

Note that the number of hits increased by 1 even after changing the norm. If we change a different parameter it does not get changed:

```
>>> mdl.kt.val
np.float64(1.0)
>>> mdl.kt = 4
array([1.87618031, 0.94268214, 0.62276287])
>>> mdl.cache_status()
 apec                       size:    2  hits:     2  misses:     2  check:     4
```

hamogu: The "normal" decorator will cache all input values, which includes the norm. So, we still need some spacial version for XSPEC to not include the norm. However, it seems that version doesn't select the correct parameter to cache or not cache right now.

DougBurke: The normal decorator caches all the parameters it has been sent. I claim that the version I just wrote only sends the non-norm parameters to the model to be cached so we have something like (I believe)

```python
class XSAdditiveModel(...):

    # This calc method does not know anything about the cache (there is no decorator applied to it now)
    def calc(self, pars, *args, **kwargs):
        # This should call the cache code, as the decorator is applied to super().calc
        possibly_cached = super().calc(pars[:-1], *args, **kwargs)
        return pars[-1] * possibly_cached
```

DougBurke: With this code we have

```
>>> mdl = xspec.XSapec()
>>> type(mdl)
<class 'sherpa.astro.xspec.XSapec'>
>>> isinstance(mdl, xspec.XSAdditiveModel)
True
>>> isinstance(mdl, xspec.XSMultiplicativeModel)
False
>>> mdl.cache_status()
 apec                       size:    0  hits:     0  misses:     0  check:     0
>>> mdl([0.1, 0.2, 0.3])
/lagado.real/sherpa/sherpa-work/sherpa/astro/xspec/__init__.py:1576: FutureWarning: calc() requires pars,lo,hi arguments, sent 2 arguments
  warnings.warn(emsg, FutureWarning)
Reading APEC data from 3.0.9

array([2.10839183, 0.31196176, 0.        ])
>>> mdl.cache_status()
 apec                       size:    1  hits:     0  misses:     1  check:     1
array([2.10839183, 0.31196176, 0.22008776])
>>> mdl.cache_status()
 apec                       size:    2  hits:     0  misses:     2  check:     2
array([2.10839183, 0.31196176, 0.22008776])
>>> mdl.cache_status()
 apec                       size:    2  hits:     1  misses:     2  check:     3
array([2.10839183, 0.31196176, 0.22008776])
>>> mdl.cache_status()
 apec                       size:    2  hits:     2  misses:     2  check:     4
>>> mdl.norm  = 2
>>> mdl.cache_status()
 apec                       size:    2  hits:     2  misses:     2  check:     4
array([4.21678367, 0.62392352, 0.44017552])
>>> mdl.cache_status()
 apec                       size:    2  hits:     3  misses:     2  check:     5
```

So you an see the cache was used when we changed just the `norm` of this additive model, which is what the previous code did.

DougBurke: @hamogu - I don't know if I ever properly answered above why I decided not to implement your documentation suggestion but instead re-wrote the code.

PR #2278, opened 2025-04-15
Use `modelCacher1D` (now aliased to `modelCacher`) for 2D models

hamogu:
## Summary
Set up caching for 2D models

## Details
For a long time, `modelCacher1D` has accepted an arbitrary number of arguments. That is useful because 1D models can be at points (x) or integrated over bins (xlo, xhi). However, that means that is seamlessly also works for (x0, x1) in 2D models or (x0lo, x0hi, x1lo, x1hi) in integrated 2D models. This commit adds an alias `modelCacher = modelCacher1D` (for backwards compatibility we should not remove the old name and I don't see a reason to deprecate it either) and starts using is on some more complex 2D models.

This was mentioned in: 
https://github.com/sherpa/sherpa/discussions/2276

### DIFF
--- a/sherpa/astro/models/__init__.py
+++ b/sherpa/astro/models/__init__.py
@@ -1025,6 +1025,7 @@ class Beta2D(RegriddableModel2D):
         ArithmeticModel.__init__(self, name,
                                  (self.r0, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl, self.alpha))
+        self.cache = 0
 
     def get_center(self):
         return (self.xpos.val, self.ypos.val)
@@ -1107,6 +1108,7 @@ class DeVaucouleurs2D(RegriddableModel2D):
         ArithmeticModel.__init__(self, name,
                                  (self.r0, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl))
+        self.cache = 0
 
     def get_center(self):
         return (self.xpos.val, self.ypos.val)
@@ -1193,6 +1195,7 @@ class HubbleReynolds(RegriddableModel2D):
         ArithmeticModel.__init__(self, name,
                                  (self.r0, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl))
+        self.cache = 0
 
     def get_center(self):
         return (self.xpos.val, self.ypos.val)
@@ -1269,6 +1272,7 @@ class Lorentz2D(RegriddableModel2D):
         ArithmeticModel.__init__(self, name,
                                  (self.fwhm, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl))
+        self.cache = 0
 
     def get_center(self):
         return (self.xpos.val, self.ypos.val)
@@ -1484,6 +1488,7 @@ class Sersic2D(RegriddableModel2D):
         ArithmeticModel.__init__(self, name,
                                  (self.r0, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl, self.n))
+        self.cache = 0
 
     def get_center(self):
         return (self.xpos.val, self.ypos.val)
@@ -1551,7 +1556,9 @@ class Disk2D(RegriddableModel2D):
         self.ampl = Parameter(name, 'ampl', 1)  # p[2]
         self.r0 = Parameter(name, 'r0', 1, 0)  # p[3]
         ArithmeticModel.__init__(self, name, (self.xpos, self.ypos, self.ampl, self.r0))
+        self.cache = 0
 
+    @modelCacher
     def calc(self, p, x, y, *args, **kwargs):
         # Compute radii
         r2 = (x - p[0]) ** 2 + (y - p[1]) ** 2
@@ -1606,6 +1613,7 @@ class Shell2D(RegriddableModel2D):
         self.r0 = Parameter(name, 'r0', 1, 0)  # p[3]
         self.width = Parameter(name, 'width', 0.1, 0)
         ArithmeticModel.__init__(self, name, (self.xpos, self.ypos, self.ampl, self.r0, self.width))
+        self.cache = 0
 
     @modelCacher
     def calc(self, p, x, y, *args, **kwargs):
--- a/sherpa/astro/models/__init__.py
+++ b/sherpa/astro/models/__init__.py
@@ -23,7 +23,11 @@ import numpy
 
 from sherpa.models.basic import clean_kwargs1d, clean_kwargs2d
 from sherpa.models.parameter import Parameter, tinyval
-from sherpa.models.model import ArithmeticModel, RegriddableModel2D, RegriddableModel1D, modelCacher1d
+from sherpa.models.model import (ArithmeticModel,
+                                 RegriddableModel2D, RegriddableModel1D,
+                                 modelCacher,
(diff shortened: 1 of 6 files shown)


DougBurke:
I would want to make sure we have some realistic fit scenarios here to see how this changes runtimes, since the data volume is a bit different than the 1D - e.g. 2048^2 rather than 8192 axis elements - which can change things such as

- time to calculate the axis "hash"
- memory storage

Creating such a set is a large enough task that I don't see it happening soon.

DougBurke:
@hamogu - there's a couple of what should be easy conflicts to deal with here. Hopefully.

sherpa/models/model.py
```diff
@@ -374,13 +375,18 @@ def boolean_to_byte(boolean_value: bool) -> bytes:
...
 
-    Apply to the `calc` method of a 1D model to allow the model
+    Apply to the `calc` method of a model to allow the model
     evaluation to be cached. The decision is based on the `integrate`
     setting, the evaluation grid, parameter values, and the keywords
     sent to the model.
 
+    For historic reasons, this decorator was implemented for 1D models
+    first. However, it is general enough that it works in higher dimensions,
+    too. It can be used as either ``@modelCacher1d`` or ``@modelCacher``;
```
DougBurke, line 387: Line 403 - which I can't comment on - is part of the docstring and uses the `@modelCacher1d` name - we should probably change it to `@modelCacher`.

DougBurke: Probably also line 425 if we do this.

DougBurke:
As written this code will not work correctly for `Data2DInt`-type models:

https://github.com/sherpa/sherpa/blob/83b777a8fcb880c9d136779139e6a18084d85513/sherpa/models/model.py#L442-L477

shows that it uses the first axis and - if any more are sent - the second one (so this handles `Data1D`, `Data1DInt`, `DataPHA`, and `Data2D`) - but it will not include any others in the calculation of the "hash value" / digest. Unless I'm missing something.

It shouldn't be a hard change: something like

```python
for ary in args:
    data.append(np.asarray(ary).tobytes())
```

but I haven't tested this or thought it through.

hamogu:
> EDITED TO ADD if a user calls model(xlo, xhi) and then model(xlo=xlo, xhi=xhi) then I think the current scheme "breaks", since the model will get re-evaluated rather than the cache re-used. I think that's okay (since, in part, we don't have an official name on the arguments, so you can not just assume they are xlo and xhi for 1D models, but also because users aren't likely to be changing how the model is evaluated when trying to evaluate the models).

That's an OK breakage. It's (as @DougBurke points out) very unlikely to happen in practice and even if it does, it will only waste some CPU time, but it won't silently give wrong results (since the model is calculated again with the same parameters).  I do not think we should (or can, given the implementation), do something about it.

sherpa/models/model.py
```diff
@@ -440,10 +440,9 @@ def cache_model(cls, pars, xlo, *args, **kwargs):
             integrate = kwargs.get('integrate', False)
 
         data = [np.array(pars).tobytes(),
-                boolean_to_byte(integrate),
-                np.asarray(xlo).tobytes()]
-        if args:
-            data.append(np.asarray(args[0]).tobytes())
+                boolean_to_byte(integrate)]
```
DougBurke, line 443: Thinking out loud: so, this will invalidate the cache contents if a user tries to load in a save/pickled file (since the "schema" used to calculate the digest/hash has changed).

I am okay with this. There's also the possibility that we don't explicitly restore the cache from picked files (I seem to remember @hamogu talking about something like this with some recent changes, but I forget which); I also don't think we need to do anything to change the current system here.

Is it worth documenting somewhere - e.g. a generic "Cache results are not guaranteed to remain consistent if models are restored from a pickled file." type of statement? I don't think so, but thought I'd raise it.

sherpa/models/basic.py
```diff
@@ -1637,7 +1642,6 @@ def guess(self, dep, *args, **kwargs):
         param_apply_limits(fwhm, self.sigma_b, **kwargs)
         param_apply_limits(norm, self.ampl, **kwargs)
 
-    @modelCacher
```
DougBurke: Why is this removed here?

sherpa/astro/models/__init__.py
```diff
@@ -1025,6 +1025,7 @@ def __init__(self, name='beta2d'):
         ArithmeticModel.__init__(self, name,
                                  (self.r0, self.xpos, self.ypos, self.ellip,
                                   self.theta, self.ampl, self.alpha))
+        self.cache = 0
```
DougBurke: So, this could be applied in `sherpa.models.model.RegriddableModel2D` - e.g. as a class attribute, in the same way that `ndim` is set - which would mean we don't need to apply it to each 2D model individually (at least for those that derive from it, which I expect to be all of them).

I think this would be cleaner, but it wouldn't surprise me if it had some issue somewhere else.

sherpa/models/tests/test_caching.py
```diff
@@ -175,6 +175,7 @@ def test_evaluate_cache1dint():
     x1hi = np.array([-10, 1, -8])
 
     mdl = Polynom2D()
+    mdl.cache = 12
```
DougBurke, line 178: The test name here is wrong, shouldn't it be `test_evaluate_cache2d`? I haven't checked to see if there is a repeated `test_evaluate_cache1dint` that this shadows, or whether we need one.

Also, should this tests sending in an integrated dataset but with the `integrate` flag set to `False` - which is an important test. Do we also need to test with it set to `True`, or does another test already cover this?

sherpa/astro/models/__init__.py
hamogu: The problem is that `cache` is a property in a class that is a parent of `RegriddableModel2D`. If I simply add the line `cache=0` to `RegriddableModel2D`, then I'm overwriting that property. In other classes, we added `self.cache=5` to the `__init__` in the Baseclass. Then, it would set that through the property. However, in the 2D models, the base class `__init__` is never called thus that approach won't work unless I add a `super.__init__(...)` for every `__init__` function. It would probably be cleaner to add that call, but it's also more invasive than I wanted to be in this PR.

sherpa/models/basic.py
hamogu: That's wrong, I'll add it back in. Thanks for catching!

sherpa/models/tests/test_caching.py
hamogu: There is a `test_evaluate_cache2d` above, so I'll come up with a new and more correct name for this one.
Most of the 1d cache testing is in `test_model.py`; that file is just so long that I decided to make a new one specifically for testing caching in this PR. THe 1D tests could be moved over, but that's just code churn, so I left them where they are for now.

There are 1D tests with integrate on and off, so I don't think we need ot repeat everything here for 2D, just the things that are specific to 2D (more axes).

sherpa/astro/models/__init__.py
DougBurke: Ta for checking. It doesn't help that our classes are often caught between the two approaches you can use in Python - either be explicit about the parent class name or use `super()` - which confuses things a bit (there is a reason to be explicit, i.e. `super` is not always the answer) and this is definitely not the time to go down that bandwagon!

PR #2280, opened 2025-04-16
Add option `record_steps` to `Fit.fit()` calls

### LINKED ISSUE #2269: Add ability to return a numpy record array instead of/ in addition to a text file in `ui.fit()` and `Fit.fit()`
Sherpa's fitting process can keep a record of all the parameters it tried in the separate fitting steps. Traditionally, that's written to a text file, e.g.
```Python
ui.fit(outfile='mysfitsteps.txt', clobber=True)
```
It's also not hard to capture that in memory:
```Python
import io
fitsteps = io.StringIO()
ui.fit(outfile=fitsteps)

from astropy.table import Table
out = Table.read(fitsteps.getvalue().split('\n'), format='ascii.commented_header')
fitsteps.close()
```
where in this example I used astropy to parse the text file into a table, but pandas or `np.genfromtxt()` would also work with the right parameters. See https://github.com/sherpa/sherpa/issues/836#issuecomment-2794335480 for a more complete discussion.

However, it would be even easier if the fit could return a numpy record array directly to save the user the intermediate steps of capturing and then parsing the text file, e.g. like this `step_table = ui.fit(return_steps=True)` and similar for a fit in the OO interface.

This is mostly useful when trying to understand why certain fits don't work. It's not clear how often users would use that in practice, but it helps to make the process smoother.

hamogu:
## Summary
Add option `record_steps` to `Fit.fit()` calls, which will add a field `record_steps` to the `FitResults` object that the fitter returns. This can help to understand why a fit does not converge.

## Details
The fit function previously had an option to record statistic and parameter values for each step of a fit to a file. This is helpful to understand why a fit may not converge or does not find a global minimum.
However, the file loading and parsing (or capturing in a StringIO object) is somewhat cumbersome.
This commit adds a simpler way to have that information available in the `FitResults` object.

By default `record_steps=False` to match previous behavior and conserve memory.

fixes #2269

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -11350,7 +11350,7 @@ class Session(sherpa.ui.utils.Session):
            file handle instead of a string.
 
         .. versionchanged:: 4.17.1
-           The parameters ``record_steps`` was added to keep parameter
+           The parameter ``record_steps`` was added to keep parameter
            values of each iteration with the fit results.
 
         Parameters
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -11349,6 +11349,10 @@ class Session(sherpa.ui.utils.Session):
            The outfile parameter can now be sent a Path object or a
            file handle instead of a string.
 
+        .. versionchanged:: 4.17.1
+           The parameters ``record_steps`` was added to keep parameter
+           values of each iteration with the fit results.
+
         Parameters
         ----------
         id : int or str, optional
@@ -11364,6 +11368,11 @@ class Session(sherpa.ui.utils.Session):
            overwritten (``True``) or if it raises an exception
            (``False``, the default setting). This is only used if
            `outfile` is set to a string or Path object.
+        record_steps : bool, optional
+            If `True`, then the parameter values and statistic value
+            are recorded at each iteration in an array in the
+            `~sherpa.fit.FitResults` object that you can obtain with
+            `get_fit_results`.
 
         Raises
         ------
@@ -11418,6 +11427,18 @@ class Session(sherpa.ui.utils.Session):
         Simultaneously fit data sets 1, 2, and 3:
 
         >>> fit(1, 2, 3)
+
+        Fit the dataset 'jet' and keep the values for each parameter in every
+        optimization step:
+
+         >>> fit('jet', record_steps=True)
+         >>> fres = get_fit_results()
+         >>> for row in fres.record_steps:
+         ...     print(f"{row['nfev']} {row['statistic']:8.6e} {row['const1d.c0']:6.4f}")
+         [... output here ...]
+
+        (This example assumes that the model for the data set 'jet' has a
+        parameter called 'const1d.c0'.)
 
         Fit data set 'jet' and write the fit results to the text file
         'jet.fit', over-writing it if it already exists:
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1313,7 +1313,7 @@ class Fit(NoNewAttributesAfterInit):
            file handle instead of a string.
 
         .. versionchanged:: 4.17.1
-           The parameters ``record_steps`` was added to keep parameter
+           The parameter ``record_steps`` was added to keep parameter
            values of each iteration in the `FitResults` object that is
            returned.
 
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -221,6 +221,10 @@ class FitResults(NoNewAttributesAfterInit):
        and now contains the covariance matrix estimated at the
        best-fit location, if provided by the optimiser.
 
+    .. versionchanged:: 4.17.1
+       The `record_steps` attribute has been added to the class.
+       If the fit recorded the parameter values of each step in the optimization,
+       this attribute will contain an array with the data.
     """
 
     # The fields to include in the __str__ output.
@@ -1307,6 +1311,11 @@ class Fit(NoNewAttributesAfterInit):
         .. versionchanged:: 4.17.0
            The outfile parameter can now be sent a Path object or a
            file handle instead of a string.
+
+        .. versionchanged:: 4.17.1
+           The parameters ``record_steps`` was added to keep parameter
+           values of each iteration in the `FitResults` object that is
+           returned.
 
         Parameters
         ----------
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -303,6 +303,10 @@ class FitResults(NoNewAttributesAfterInit):
         self.extra_output = results[4]
         """The ``extra_output`` field from the fit."""
 
+        self.record_steps: None | dict = results[4].get('record_steps')
+        """A record of all steps taken during the fit, if requested
+        with `record_steps=True`."""
+
         self.modelvals: np.ndarray = _vals
         """The values of the best-fit model evaluated for the data."""
 
@@ -615,19 +619,21 @@ class IterCallback:
 
     """
 
-    __slots__ = ("data", "model", "stat", "fh", "nfev")
+    __slots__ = ("data", "model", "stat", "fh", "nfev", "record_steps")
 
     def __init__(self,
                  data: DataSimulFit,
                  model: SimulFitModel,
                  stat: Stat,
-                 fh: WriteableTextFile | None = None
+                 fh: WriteableTextFile | None = None,
+                 record_steps: bool = False
                  ) -> None:
         self.data = data
         self.model = model
         self.stat = stat
         self.fh = fh
         self.nfev = 0
+        self.record_steps = [] if record_steps else None
 
     def __call__(self,
                  pars: np.ndarray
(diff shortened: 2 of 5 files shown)


sherpa/fit.py
```diff
@@ -1364,19 +1392,21 @@ def fit(self,
...
-        2.000000e+00 1.899834e+02 1.000345e+00
-        3.000000e+00 4.600000e+01 7.000000e+00
-        4.000000e+00 4.600002e+01 7.002417e+00
-        5.000000e+00 4.600000e+01 7.000000e+00
-
+        >>> out = f.fit(record_steps=True)
+        >>> for n, stat, c0 in zip(out.record_steps['nfev'],
+        ...                          out.record_steps['statistic'],
+        ...                          out.record_steps['const1d.c0']):
+        ...     print(f"{n}, {stat:6.6e}, {c0:8.6f}")
```
DougBurke, line 1400: The f-string has commas after the columns but the output below doesn't have them. Shouldn't the doc-testing have caught this?

sherpa/fit.py
```diff
@@ -1430,8 +1461,9 @@ def fit(self,
                         param_warnings += f"WARNING: parameter value {par.fullname} is at its minimum boundary {par.min}\n"
                     if sao_fcmp(par.val, par.max, tol) == 0:
                         param_warnings += f"WARNING: parameter value {par.fullname} is at its maximum boundary {par.max}\n"
+            record_steps = cb.record_steps
 
-        output = (status, newpars, fval_new, msg, imap)
+        output = (status, newpars, fval_new, msg, imap, record_steps)
```
DougBurke, line 1466: The trouble here is that `FitResults` accepts an `OptReturn` type for its second argument. I am making use of this type in my clean-up code (as I've found having useful type statements added to sherpa.optmethods/estmethods./fit quite helpful in identifying how information flows), and this extra information isn't needed/wanted in the optimizer code.

So I suggest instead that we send the `record_steps`  in some other way to `FitResults`.

**Or** we can take advantage of `OptReturn` having a get-out-of-jail-free type last argument - in this case `imap`, a dictionary, and so do something like

````python
imap[some-key-or-other] = record_steps
output = (status, newpars, fval_new, msg, imap)
```

and then `FitResults` will fish the value out of this dictionary.

sherpa/tests/test_fit_unit.py
```diff
@@ -3137,6 +3137,30 @@ def test_fit_outfile_simple_check_stringio(check_str):
...
+                     -4.073357, -4.079456, -4.079455, -4.079456, -4.079456]
+    expected_c0 = [1.0, 1.0, 1.000345, 2.454143, 2.454991, 3.250567,
+                     3.251690, 3.333285, 3.334435, 3.333329, 3.333329]
+
+    assert fitres.record_steps['nfev'] == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+    assert fitres.record_steps['statistic'] == pytest.approx(expected_stat)
+    assert fitres.record_steps['mx.c0'] == pytest.approx(expected_c0)
+    assert len(fitres.record_steps) == 3
+
+
```
DougBurke, line 3163: A hypothesis-like test would do fits and compare the file output to the recorded values and check they agree. But that seems like a lot of work for little, if any, extra power.

sherpa/fit.py
```diff
@@ -648,6 +660,13 @@ def __call__(self,
             vals.extend([f'{val:5e}' for val in self.model.thawedpars])
             self.fh.write(' '.join(vals) + '\n')
 
+        if self.record_steps is not None:
+            # Store the current parameter values
+            self.record_steps['nfev'].append(self.nfev)
+            self.record_steps['statistic'].append(output[0])
+            for p in self.model.get_thawed_pars():
+                self.record_steps[p.fullname].append(p.val)
+
```
DougBurke, line 669: I wonder, from time to time, whether we should really use more-structured data rather than what is essentially `dict[str, np.ndarray]`, but find myself never finding enough impetus to do it. For example, here where we add a row at a time means either always extending it or doing something clever like expanding the length by 2 every time it fills up and then truncating it when we have finished. But it doesn't feel worth it.

hamogu: I chose this format because it's trivial to convert into an `astropy.table.Table` like so `tab = Table(out.record_steps)` but since we don't have a hard dependency on astropy, only hinted at that in one of the docstrings and didn't include a full test or code example.

hamogu: I don't want to re-invent the wheel of a "general table". Pandas is also widely used, but we don't have a dependency on pandas either, so I decided that this is the simplest way to do it.

sherpa/tests/test_fit_unit.py
hamogu: If both file and table are correct, then they will also be equal. Yes, we could test for all possible inputs, but the more interesting question would be: Even if they are the same, are they correct?
And for such a rarely used feature I didn't want to do something too complex and too run-time intensive on CI.

sherpa/fit.py
hamogu: It would have caught that, if we run doctests on it but `sherpa/fit.py` is still on the balcklist for for doctests (see list in `purest.ini` because we never made all the other docstrings in the file pass - but this is exactly why I want to increase coverage of docstrings!

sherpa/fit.py
```diff
@@ -1364,19 +1392,21 @@ def fit(self,
...
+        ...     print(f"{n}, {stat:6.6e}, {c0:8.6f}")
+        0 1.900000e+02 1.0000
+        1 1.900000e+02 1.0000
+        2 1.899834e+02 1.0003
+        3 4.600000e+01 7.0000
+        4 4.600002e+01 7.0024
+        5 4.600000e+01 7.0000
+
+        This format is also easy to plot, e.g.
+        ``plt.plot(out.record_steps['nfev'], out.record_steps['statistic'])``.
```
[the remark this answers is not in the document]
hamogu: Sure. Will restore the old example in the next commit.

sherpa/fit.py
hamogu: Oh, and I forgot: From what I've read, adding rows to numpy structured arrays (which I intended to use first) is a relatively expensive operation and requires a bit of unwieldy code. There are no convenience functions to do that, because you are not supposed to do it.
On the other hand, in Python, lists are already optimized for this "add one and then another one". When you add to a list it will internally add a few spaces that are empty so that on future adds not new memory allocation is needed. However, I did not speed benchmark this and it's likely irrelevant because the model evaluation will typically much longer than adding a few numbers in any way. So, it we really don't like this, we could go with a numpy record array. A compromise would be to convert to a record array before we return it, so that that more structured format is what the user sees.

DougBurke:
Can we add ``.. versionchanged:: 4.17.1`` sections to `FitResults` and `fit` to mention `record_steps`?

We would probably also want some documentation added to `sherpa.ui.utils.fit` to show how to get the results - I assume it would be similar to the existing `StringIO` example but you need to include `get_fit_results` to get the actual data.

DougBurke:
I assume the foc changes are the reason for the doc build failure

```
/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/2280/lib/python3.10/site-packages/sherpa/astro/ui/__init__.py:docstring of sherpa.astro.ui.fit:129: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/2280/lib/python3.10/site-packages/sherpa/astro/ui/utils.py:docstring of sherpa.astro.ui.utils.Session.fit:129: WARNING: Block quote ends without a blank line; unexpected unindent.
```

hamogu:
Just whitespace off by one. Fixed.

DougBurke (approved):
Thanks for the doc changes.

sherpa/astro/ui/utils.py
```diff
@@ -11349,6 +11349,10 @@ def fit(self,
            The outfile parameter can now be sent a Path object or a
            file handle instead of a string.
 
+        .. versionchanged:: 4.17.1
+           The parameters ``record_steps`` was added to keep parameter
```
wmclaugh, line 11353:
```suggestion
           The parameter ``record_steps`` was added to keep parameter
```

sherpa/fit.py
```diff
@@ -1287,25 +1303,36 @@ def calc_stat_info(self) -> StatInfoResults:
...
+            record_steps: bool = False,
             ) -> FitResults:
         """Fit the model to the data.
 
         .. versionchanged:: 4.17.0
            The outfile parameter can now be sent a Path object or a
            file handle instead of a string.
 
+        .. versionchanged:: 4.17.1
+           The parameters ``record_steps`` was added to keep parameter
```
wmclaugh, line 1316:
```suggestion
           The parameter ``record_steps`` was added to keep parameter
```

sherpa/ui/utils.py
```diff
@@ -9583,6 +9583,10 @@ def fit(self,
            The outfile parameter can now be sent a Path object or a
            file handle instead of a string.
 
+        .. versionchanged:: 4.17.1
+           The parameters ``record_steps`` was added to keep parameter
```
wmclaugh, line 9587:
```suggestion
           The parameter ``record_steps`` was added to keep parameter
```

PR #2107, opened 2024-08-20
XSPEC: add show_xsabund command

DougBurke:
# Summary

Add the show_xsabund command to the sherpa.astro.ui module and get_xsabundances to sherpa.astro.xspec.

# Details

This adds

 - sherpa.astro.ui.show_xsabund (and also to sherpa.astro.ui.utils.Session)

 - sherpa.astro.xspec.get_xsabundances() and, not exported, sherpa.astro.xspec.get_xselements()

The aim is to provide a "useful" show_xsabund() command for the UI user, but the get_xsabundances() command may be useful too.

Unfortunately, with the current interface, we can't get at the description/documenatation for each abundance table, so we can only label it as "`angr`" and not "`angr: Anders E. & Grevesse N. ...`". This could be done but would require adding functionality to the compiled code in `sherpa.astro.xspec._xspec`, and I claim the code here adds a useful capability as is, and we can extend it later.

This code is based on the larger "take advantage of the FunctionUtility interface" PR #1615, which is where a number of the tests are taken, that cover some functionality not actually changed here.

**EDITED TO ADD** When adding the routine (`set_xsabund`) to the read-the-docs documentation I realised that we could add some recently-added routines which had been missed from the PRs that added them (well, I first noted this in a different PR and thought I'd add them here). This does not add anything to sherpa itself, just the pages under https://sherpa.readthedocs.io/en/latest/ui/

# Example

This is the current behavior:

```
>>> from sherpa.astro import ui
>>> ui.get_xsabund()
'angr'
>>> ui.get_xsabund()
'angr'
```

The `show_xsabund` command is based on the XSPEC `show abund` command and returns:

```
>>> ui.show_xsabund()
Solar Abundance Table:
angr
  H : 1.000e+00  He: 9.770e-02  Li: 1.450e-11  Be: 1.410e-11  B : 3.980e-10
  C : 3.630e-04  N : 1.120e-04  O : 8.510e-04  F : 3.630e-08  Ne: 1.230e-04
  Na: 2.140e-06  Mg: 3.800e-05  Al: 2.950e-06  Si: 3.550e-05  P : 2.820e-07
  S : 1.620e-05  Cl: 3.160e-07  Ar: 3.630e-06  K : 1.320e-07  Ca: 2.290e-06
  Sc: 1.260e-09  Ti: 9.770e-08  V : 1.000e-08  Cr: 4.680e-07  Mn: 2.450e-07
[...]

PR #2281, opened 2025-04-17
Provide access to XSPEC abundances

### LINKED ISSUE #2282: buglet: redshift parameter name is displayed strangely for XSeebremss model
```
>>> foo = xspec.XSeebremss("foo")
>>> print(foo)
foo
   foo.T        thawed            1         0.05        1e+10        keV
   foo.eperh    frozen          1.2            0           10           
   0.Redshift   frozen            0       -0.999           10           
   foo.norm     thawed            1            0        1e+24           
```

No matter what you name the model the redshift parameter will be labelled "`0.Redshift`".

This is only a display issue; the model works fine.

It's a silly error (and I found it because of the typing annotations I've been adding).

DougBurke:
# Summary

Minor improvements to the XSPEC interface: get_xsabundances now takes an optional table name, get_xsxset can be called with no argument, and clear_xsxset will clear the XSET database. Fix #2282.

# Details

Fix a minor error in the XSeebremss model (it just affects the string/print output of the model): issue #2282.

The rest are changes from #2216 without the problematic atomdb/nei version settings and some version string accessors that may suffer similar issues (e.g. you can change them but they don't seem to change anything if models have already run).

A number of routines have been added or changed in `sherpa.astro.xspec._xspec` as this is treated as an "internal" module (i.e. if users are calling this directly then they have to expect changes). The user-level changes in `sherpa.astro.xspec` are intended to be more backwards-compatible, as normally we have just added routines or added optional arguments. There is one breaking change however, noted below.

Changes are:

- added `clear_xsxset` as an "obvious" way to clear the XSXSET database (there was a hidden way to do this but you had to know XSPEC internals to do so, by setting the `"INITIALIZE"` key)
- `get_xsxset` can now be called with no name, in which case it returns a dict containing all the settings
- `get_xsxset` now raises a `KeyError` if given an invalid name (to make the dict-like behaviour cleaner) rather than returning an empty string
  - this is an explicit breaking change
- there have been internal changes to make `set_xsabundances` nicer, but it's not visible to the user, but it does mean that if you have set abundances from a file (or adjusted the values) then they will be recorded in the "xspec state"
- added `get_xsabundances_path` as a way to get the `Path` to the abundances file used by XSPEC
  - this is not exported by default as it is expected to be rarely used
- ensure you can't accidentally crash XSPEC by trying to set the abundances to "file" if you haven't actually loaded in any data (either from a file or by manual edits)
- the `set_xsabund` C++ implementation has added comments to explain why we don't use `FunctionUtility::readNewAbnudances()` but still keep the code from when we only were able to use the xsFORTRAN interface (which does not have a "read abundances from a file" routine); basically the error handling and behaviour of `readNewAbundances` mean it's not worth changing our existing code
- easy access to all the abundance tables, not just the currently loaded one, via `get_xsabundances("lodd")` etc
  - in earlier versions of this work I had many more ways to get this information, but for Sherpa users I think being able to just get all the abundances as a dict, labelled by element name, is all that we really need

### DIFF
--- a/sherpa/astro/xspec/__init__.py
+++ b/sherpa/astro/xspec/__init__.py
@@ -142,7 +142,6 @@ warning = logging.getLogger(__name__).warning
 # record these settings manually.
 #
 xsstate: dict[str, Any] = {
-    "modelstrings": {},
     "paths": {},
 }
 
@@ -723,24 +722,71 @@ def set_xsxsect(name: str) -> None:
     _xspec.set_xsxsect(name)
 
 
+def clear_xsxset() -> None:
+    """Remove all existing X-Spec model settings.
+
+    .. versionadded:: 4.17.1
+
+    See Also
+    --------
+    get_xsxset, set_xsxset
+
+    Examples
+    --------
+
+    >>> set_xsxset("POW_EMIN", "0.5")
+    >>> get_xsxset()
+    {'POW_EMIN': '0.5'}
+    >>> clear_xsxset()
+    >>> get_xsxset()
+    {}
+
+    """
+    return _xspec.clear_xsxset()
+
+
+@overload
+def get_xsxset() -> dict[str, str]:
+    ...
+
+@overload
+def get_xsxset(name: None) -> dict[str, str]:
+    ...
+
+@overload
 def get_xsxset(name: str) -> str:
-    """Return the X-Spec model setting.
+    ...
+
+def get_xsxset(name: str | None = None) -> str | dict[str, str]:
+    """Return the X-Spec model setting or settings.
+
+    .. versionchanged:: 4.17.1
+       This routine can now be called with no argument, which means
+       that a dictionary containing all the settings is returned.
+       Asking for an unset keyword now raises a KeyError rather than
+       returning the empty string.
 
     Parameters
     ----------
-    name : str
+    name : str or None, optional
        The name of the setting (converted to upper case before being
        sent to X-Spec). There is no check that the name is valid.
 
     Returns
     -------
-    val : str
-       Returns the value set by a previous call to `set_xsxset` or the
-       empty string, if the value has not been previously set.
+    val : str or dict
+       When name is set, returns the value set by a previous call to
+       `set_xsxset`. When name is not set a dictionary of all the
+       settings is returned.
+
+    Raises
+    ------
+    KeyError
+       When the name has not been set.
 
     See Also
     --------
-    set_xsxset
+    clear_xsxset, set_xsxset
 
     Notes
     -----
@@ -756,9 +802,14 @@ def get_xsxset(name: str) -> str:
     >>> get_xsxset("pow_emin")
     '0.5'
 
+    >>> get_xsxset()
+    {'POW_EMIN': '0.5'}
+
     """
-    name = name.upper()
-    return _xspec.get_xsxset(name)
+    if name is None:
+        return _xspec.get_xsxset()
+
+    return _xspec.get_xsxset(name.upper())
 
 
 def set_xsxset(name: str, value: str) -> None:
@@ -775,12 +826,14 @@ def set_xsxset(name: str, value: str) -> None:
        The name of the setting. It is converted to upper case before
        being used. There is no check that the name is valid.
     value : str
-       The new value of the setting. It must be given as a string.
+       The new value of the setting. It must be given as a string.  If
+       it is the empty string then the name is removed from the model
+       settings.
 
     See Also
     --------
-    get_xsxset, get_xsversion, set_xsabund, set_xschatter, set_xscosmo,
-    set_xsxsect
+    clear_xsxset, get_xsxset, get_xsversion, set_xsabund, set_xschatter,
+    set_xscosmo, set_xsxsect
 
     Notes
     -----
@@ -792,6 +845,10 @@ def set_xsxset(name: str, value: str) -> None:
 
     The model settings are stored so that they can be included in the
     output of `sherpa.astro.ui.save_all`.
+
+    Setting the name argument to "INITIALIZE" will clear the settings,
+    but this is deprecated and the `clear_xsxset` routine must be used
+    instead.
 
     References
     ----------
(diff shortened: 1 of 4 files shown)


DougBurke:
@hamogu - this is a reduced version of 

- #2216 

plus a fix for 

- #2282 

If you have time to look at an XSPEC PR it would be better this than #2216!

When I started looking at adding extra functionality from XSPEC, I was concentrating on low-level things - e.g. get the elemental abundance for a particular element (given by name or Z). I've come to the conclusion that for Sherpa this is overkill, and doing things like "send back a dict" or "send in a dict" is a more-useful interface. The changes here are based on this approach (which we reviewed in #2107 and #2215).

One thing I've dropped (so far) from #1615 is the ability to get the abundance settings from a table that is not the selected one - e.g. we could support

```python
t1 = xspec.get_xsabundances("felc")
t2 = xspec.get_xsabundances("aneb")
t3 = xspec.get_xsabundances()
```

It's a bit harder to do this for `xspec.get_xsabund` as we already have an optional string argument, but it could be added. 

However, a user can explicitly change the table with `xspec.set_xsabund` so I have decided against adding this ability for now.

sherpa/astro/xspec/__init__.py
```diff
@@ -723,24 +758,71 @@ def set_xsxsect(name: str) -> None:
...
-       empty string, if the value has not been previously set.
+    val : str or dict
+       When name is set, returns the value set by a previous call to
+       `set_xsxset`. When name is not set a dictionary of all the
+       settings is returned.
+
+    Raises
+    ------
+    KeyError
+       When the name has not been set.
```
hamogu, line 821:
```suggestion
       When the name passed is not set in X-Spec.
```
(See the text just a few lines above "When the name is not set a dictionary ... is returned", so we can't also have that it raises and error is not set.)

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -667,31 +676,103 @@ def test_abund_change_file_subset(tmp_path):
...
+                     elems2["Ar"],
+                     elems2["K"],
+                     elems2["Fe"])
+
+
+@requires_xspec
+def test_get_xsabundances_table_name_file():
+    """Check we can get a different abundance table to the default.
+
+    Unlike test_get_xsabundances_table_name we load our own abnudances
```
hamogu, line 705:
```suggestion
    Unlike test_get_xsabundances_table_name we load our own abundances
```

hamogu (changes requested):
The tests and Python side looks good (except for soem typos). I'll leave it to @wmclaugh to have a look at the C++ code.

DougBurke:
@hamogu - it (our GitHub overlord) still thinks you have open issues. Is there a way for you to bribe it, so I don't keep on coming back and wondering what I need to still do?

hamogu (approved):

hamogu:
It's probably because I didn't click the right button to re-review. Requesting @wmclaugh for the C code.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -158,6 +164,62 @@ static PyObject* get_abund( PyObject *self, PyObject *args )
...
+			 table );
+  }
+
+  IosHolder::setStreams(IosHolder::inHolder(),
+			IosHolder::outHolder(),
+			errStream);
+
+  // Was there an error?
+  //
+  if( tmpStream.str().size() > 0 ) {
```
wmclaugh, line 208: When checking if strings contain a value "str.empty()" is typically preferred to "str.size()" for readability. Since the old code checked size(), no need to change it.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -158,6 +164,62 @@ static PyObject* get_abund( PyObject *self, PyObject *args )
...
+    // No backwards compatability to worry about, so use the sensible
```
wmclaugh, line 209:
```suggestion
    // No backwards compatibility to worry about, so use the sensible
```

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -345,13 +490,31 @@ static PyObject* get_xset( PyObject *self, PyObject *args  )
 
   char* str_name = NULL;
 
-  if ( !PyArg_ParseTuple( args, (char*)"s", &str_name ) )
+  if ( !PyArg_ParseTuple( args, (char*)"|s", &str_name ) )
     return NULL;
 
+  // If no argument is given then we return a dictonary
```
wmclaugh, line 496:
```suggestion
  // If no argument is given then we return a dictionary
```

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -215,6 +275,15 @@ static PyObject* set_abund( PyObject *self, PyObject *args )
   tableName = XSutility::lowerCase(tableName);
 
   if (tableName == "file") {
+    // Can not use this if no abundances have been loaded (otherwise
+    // XSPEC has been known to crash).
+    //
+    if (!FunctionUtility::abundChanged()) {
```
wmclaugh, line 281: I see instances where the code sets the internal XSPEC s_abundChanged flag to true in this merge request but see no indication that it as ever initialized to false anywhere. Where is the flag initialized? (HEASARC [documentation](https://heasarc.gsfc.nasa.gov/xanadu/xspec/internal/XspecInternalFunctionsGuide.html) on this is limited)

DougBurke: So, as mentioned, the documentation is not clear. 

In a HEASARC download, e.g. `heasoft-6.35.1/Xspec/src/XSFunctions/Utilities/FunctionUtiility.cxx` we have

```C++
bool FunctionUtility::s_abundChanged = false;
```

and the only place it seems to be used is

```C++
void FunctionUtility::readNewAbundances (const string& file)
{
  using namespace std;
  const string name("file");

  s_abundChanged = true;
  s_abundanceFile = file;
  s_abundPath = XSutility::getRunPath();
...
```

The `FunctionUtility.h` file in the same directory has

```C++
inline const bool FunctionUtility::abundChanged ()
{
  return s_abundChanged;
}

inline void FunctionUtility::abundChanged (bool value)
{
  s_abundChanged = value;
}
```

There appears to be only one other place where `abundChanged` is used, and that is to read the value (not set it), and that appears to be in a "save the state to an ascii representation" routine.

sherpa/astro/xspec/src/_xspec.cc
DougBurke, line 208:
```suggestion
  if( !tmpStream.str().empty() ) {
```

DougBurke: I realized that one of the error conditions did not correctly restore the `IosHolder::setStreams` values so I added that. As I was doing that I also changed this (and the one other case) to say `if ( !tmpStream.str().empty() ) {`.

PR #2285, opened 2025-04-19
Allow guess to be used with multiple datasets

### LINKED ISSUE #2284: `guess` fails for DataSimulFit
Guessing works for one dataset:
```Python
from sherpa.fit import Fit, DataSimulFit
d = Data1D("x", [-3, 5, 17, 22], [12, 3, 8, 5])
d2 = Data1D("x2", [1, 2, 3, 4], [23, 12, -3, -3])
m = Const1D()
s = LeastSq()
d12 = DataSimulFit('d12', [d, d2])
f = Fit(d, m, stat=s)
f.guess()
```
but not for more than one (with a not very helpful error message):
```Python
f12 = Fit(d12, m)
f12.guess()
Cell In[23], [line 2](vscode-notebook-cell:?execution_count=23&line=2)
      [1](vscode-notebook-cell:?execution_count=23&line=1) f12 = Fit(d12, m)
----> [2](vscode-notebook-cell:?execution_count=23&line=2) f12.guess()

File ~/code/sherpa/sherpa/fit.py:1201, in guess(self, **kwargs)
   [1192](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1192) def _calc_stat(self) -> StatResults:
   [1193](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1193)     """Calculate the current statistic value.
   [1194](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1194) 
   [1195](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1195)     Returns
   (...)
   [1198](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1198)         The overall statistic value and the "per-bin" value.
   [1199](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1199)     """
-> [1201](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1201)     # TODO: is there anything missing here that
   [1202](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1202)     #       self._iterfit.get_extra_args calculates?
   [1203](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1203)     return self.stat.calc_stat(self.data, self.model)
   [1205](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/~/code/sherpa/sherpa/fit.py:1205) def calc_stat(self) -> float:

AttributeError: 'DataSimulFit' object has no attribute 'to_guess'
```

(I discovered this by looking at the MyPy errors in `fit.py` and then tested the example above to see that that really happens.)

DougBurke:
# Summary

Allow guess to run with multiple datasets. Fix #2284.

# Details

Adding a `to_guess` method to `DataSimulFit` is easy, apart from the fact that we allow `Data` objects to be "partially filled" - that is, the independent or dependent axes can be `None`. In this case it is not clear what to do - perhaps we should just drop that dataset - but for now we error out, as it is obviously not correct to use an empty or partially-empty dataset for the guess. This could be changed later once we have more experience.

Actually, I have now changed it so that guess (on a single dataset, not a simultaneous dataset) is now an error - this requires changing a few tests, but thet are obviously "check the behavior works as before"  so it makes sense they would need to be updated. This makes the handling consistent between the two cases (single vs multiple).

However, it now becomes unclear about how guess should work with multiple datasets, as the independent axis may not agree so we can't just sum the values (and it may not make sense to sum them). To think about. Hmm. The routines in `sherpa.utils.guess` kind-of assume the data is ordered (in the independent axis) so we either enforce this, require the guess routines to accept unordered data, or do something clever, such as send in multiple datasets separately, or run the guess routine on each dataset and "average" the results (but that's not statistically useful).

# Notes

The `ui.guess` routine adds the `otherids` argument, similar to other routines, but it adds it at the end (to avoid breaking other code) - that is

```python
    def guess(self,
              id=None,
              model=None,
              limits: bool = True,
              values: bool = True,
              otherids: Sequence[IdType] = ()
              ) -> None:
```

We could instead have

```python
    def guess(self,
              id=None,
              model=None,
              *,
              otherids: Sequence[IdType] = (),
              limits: bool = True,
              values: bool = True
              ) -> None:
```

which might be better but would be a breaking change. We could do it without the `*` which would also be a breaking change but probably a bit more "dangerous".

# Shelved

So, to do this would really require sending around `PointAxis`/`IntegratedAxis` objects rather than just `ndarray` values, so that we could work out how to "combine" multiple datasets - the existing `guess` routines may assume the data is ordered in the independent axis, and I haven't really thought through whether we need to add/average/whatever the dependent axis. This could just be done for the `to_guess` code, but it is a large change and I do not think it is feasible at this time. I have pulled out some code into

- #2323 

and will close this.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -5152,7 +5152,23 @@ It is an integer or string.
                                        (elo, ehi), ignore=ignore,
                                        integrated=True)
 
-    def to_guess(self):
+    def to_guess(self) -> tuple[np.ndarray, ...]:
+        """Return the dependent and independent axes for guessing.
+
+        .. versionchanged:: 4.18.0
+           It is now an error to call this with either the independent
+           or dependent axes unset.
+
+        Return
+        ------
+        axes : tuple of ndarray
+           The dependent axis and then the independent axes, including
+           any data filtering, grouping, and analysis setting. The
+           dependent axis is a rate if the exposure is set, and
+           corrected by the ARF if set.
+
+        """
+
         elo, ehi = self._get_ebins(group=False)
         elo = self.apply_filter(elo, self._min)
         ehi = self.apply_filter(ehi, self._max)
@@ -5162,6 +5178,10 @@ It is an integer or string.
             elo = lo
             ehi = hi
         cnt = self.get_dep(True)
+        if cnt is None:
+            raise DataErr("The dependent axis of "
+                          f"'{self.name}' has not been set")
+
         arf = self.get_specresp(filter=True)
 
         y = cnt / (ehi - elo)
@@ -5172,12 +5192,20 @@ It is an integer or string.
             y /= arf  # photons/keV/cm^2/sec or photons/Ang/cm^2/sec
         return (y, elo, ehi)
 
-    def to_fit(self, staterrfunc=None):
+    def to_fit(self,
+               staterrfunc: StatErrFunc | None = None
+               ) -> tuple[np.ndarray | None,
+                          np.ndarray | None,
+                          np.ndarray | None]:
         return (self.get_dep(True),
                 self.get_staterror(True, staterrfunc),
                 self.get_syserror(True))
 
-    def to_plot(self, yfunc=None, staterrfunc=None, response_id=None):
+    def to_plot(self,
+                yfunc: ModelFunc | None = None,
+                staterrfunc: StatErrFunc | None = None,
+                response_id: IdType | None = None
+                ):
         return (self.apply_filter(self.get_x(response_id=response_id),
                                   self._middle),
                 self.get_y(True, yfunc, response_id=response_id),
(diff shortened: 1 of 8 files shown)


DougBurke:
Hmmmm. This ain't good:

```

DougBurke:
Ahh. That was in part a silly error on my part, but it does indicate we weren't really testing the code out and trying to add a test of the above suggests the guess isn't really picking up the extra data, so more work is needed.

DougBurke:
Rebased to fix minor conflict and to try and impove the problems seen above. However, we still have a case when there are multiple datasets, as it is probable the output needs to be re-ordered - or perhaps added together  -and that gets messy (how to deal with different grids?).

PR #2286, opened 2025-04-20
Add an interface to ArviZ

### LINKED ISSUE #1942: Add a `to_arviz` method to pyBLoCXS
[Arviz](https://python.arviz.org/en/latest/index.html) is a Python package for "Exploratory analysis of Bayesian models". It provides a lot of plotting and quicklook capability and seems to be a becoming a kind of standard for plotting chains, posteriors, etc. which look much better than the plots Sherpa currently makes for pyBLoCXS (I'm talking about default settings. Both packages use matplotlib under the hood, so of course one can make identical plots for pyBLoCXS.). Arviz  supports a large number of Baysian Analysis packages in the Python world (PyMC3, PyStan, pyro, emceee, ...) out of the box, but not Sherpa. 

I suggest adding a `to_arviz` function to Sherpa's implementation of pyBLoCXS to make it easy to take advantage of Arviz capabilities when installed: https://python.arviz.org/en/latest/getting_started/CreatingInferenceData.html#

hamogu:
## Summary
Add an interface to ArviZ

## Details
ArviZ is a package for the analysis (both plotting and statistical computations) of Baysian models. It has a number of functionalities and default plots that are useful for MCMC chains that we don't have in Sherpa, and thus it is useful to provide an interface to pass Sherpa MCMC results into ArviZ.
This PR defines the interface functions, undates the docs and optional dependencies to include ArviZ. One of the CI runs is modified to install it as a dependency.

fixes #1942

## Note
The first commit in this PR is actually independent of ArviZ, it updates the MCMC docs to generate the plots automatically when sphinx is run instead of using canned pngs.


docs/mcmc/index.rst
```diff
@@ -117,39 +120,41 @@ References
...
+.. plot::
+    :include-source:
+    :context:
+    :nofigs:
+
+    Create a simulated data set:
 
-    >>> np.random.seed(2)
+    >>> import numpy as np
+    >>> np.random.seed(2)  # For reproducibility of this example
```
DougBurke, line 134: Ideally we'd do something like

```
rng = np.random.RandomState(2)  # For reproducibility of this example
```

and then use that below, but that's a larger change. We really should use `np.random.default_rng(seed)` but that's not guaranteed to be reproducible over time and we'd have to validate the results so using `RandomState` is easier for us...

docs/mcmc/index.rst
```diff
@@ -248,46 +262,51 @@ for this example)::
...
-Add in the covariance estimate::
-  
-   >>> xlo, xhi = eres.parmins[1] + eres.parvals[1], eres.parmaxes[1] + eres.parvals[1]
-   
-   >>> plt.annotate('', (xlo, 90), (xhi, 90), arrowprops={'arrowstyle': '<->'})
-   >>> plt.plot([eres.parvals[1]], [90], 'ok')
+    And adding in the covariance estimate using `matplotlib.pyplot.annotate`:
 
-.. image:: ../_static/mcmc/mcmc_pdf_xpos.png
+    >>> xlo, xhi = eres.parmins[1] + eres.parvals[1], eres.parmaxes[1] + eres.parvals[1]
```
DougBurke, line 297:
```suggestion
    >>> xlo = eres.parmins[1] + eres.parvals[1]
    >>> xhi = eres.parmaxes[1] + eres.parvals[1]
```

I'm not sure why I wrote it on one line in the first place.

docs/mcmc/index.rst
```diff
@@ -117,39 +120,41 @@ References
 Example
 =======
 
-.. note::
-
-   This example probably needs to be simplified to reduce the run time
```
DougBurke, line 122: How long does this take? If I remember it's not a particularly fast operation, which was one of the reasons why I'd not bothered to auto-generate the results.

hamogu: If I copy out the example to just run the code (and not build the html on sphinx which needs to be done anyway), it's about 10 s on my machine.

pyproject.toml
```diff
@@ -57,6 +57,13 @@ doc = [
   "bokeh"
 ]
 
+extra = [
```
DougBurke, line 60: Why has the `extra` target been added? I'd have thought that `arviz` would have been added to `doc`.

I know that @astrofrog  has been working on PRP 771 - e.g. https://discuss.python.org/t/pep-771-default-extras-for-python-software-packages/79706/76 - but I don't think it's been approved yet.

docs/mcmc/index.rst
DougBurke: I thought it was slower than that. Your machine is likely a little-bit beefier than the one that RTD uses (or I developed the example on ;-).

DougBurke: [that's assuming these times can be meaningfully compared, which as we've just been discussing with the cache code, is not always true ;-]

DougBurke:
Looking at the output plots - https://sherpa--2286.org.readthedocs.build/en/2286/mcmc/index.html#cdf-for-a-parameter is wrong - you can see it looks to have a single plot axes below the pair of plots. I don't know it this is a problem with this PR or that the existing code is wrong (or something has changed in the plotting code since it was run). All options are possible...

pyproject.toml
hamogu: It should also be added to docs.

What I'm trying to do here is to have a group for all the optional extras that user might want (that don't want to build the docs). Bokeh could go in there and if I make an interface to e.g. export fit results to pandas, then pandas would go in there.

Many users might just want everything and the kitchen sink as long as they can get it with pip.

DougBurke: I'd rather have the discussion of what goes in an `extra` argument in a separate PR, as it isn't related to this one.

DougBurke:
So, bike-shedding time:

- `mcmc_to_arviz`
- `to_arviz`

I am not sure the length is a big problem. I'm also not sure that there's a standard way of doing this (since `arviz` contains some routines for different systems but they aren't just all called `to_arviz`). So should we just stick `mcmc_to_arviz` as it exists and there's no compelling reason to change?

hamogu:
I agree than length is not a problem. This won't be called often. The "to_XXX" pattern is something that we've used in astropy [a number of times](https://docs.astropy.org/en/stable/search.html?q=to_pandas) for similar purposes so astronomers might be familiar with it. Or course, in the ArviZ pattern, that conversion function could live in ArviZ, but I'm (a) not certain that Sherpa is a big as excess or PyStan in this field and (b) Sherpa's ByBloXS is way more specialized (for Poisson data only) then the general packages that ArviZ itself provides interfaces for and thus I'm not sure the developers would take it and (c) is way easier for us to get it into Sherpa and into the sherpa docs and namespace where Sherpa users are more likely to find it.

So, unless there are strong objections, I suggest to keep the name I suggested.

pyproject.toml
hamogu: Will remove on next push. Since most users presumably install for conda anyway, they won't use `pip install "sherpa[extra,test,doc]"` anyway, so they won't benefit for it. If we add more packages, I think we should add such a group, but it doesn't have to be now.

hamogu:
About the timing: I'll try to go down to 500 steps and see what that does, but I suspect that most of the extra time is in installing arviz and making all those plots. That takes time, but it also functions as another test and keeps the plots up-to-date. Compared to when the pngs were made, we already changed several of the colors, default settings etc. I think there is value in having the plots re-made even if it take a bit of time.  

We could not have the ArviZ plot shown (and thus not install and import it on RTDs), but the intention behind this PR is give users a much larger range of plots and statistics then we'll ever write in Sherpa and the cost of just a dozen lines to us. For that, I think, it's worthwhile to invest an extra minute of RTDs run-time (CPU time is much cheaper than developer time, even if we run RTDs a large number of times).

docs/mcmc/index.rst
hamogu: done.

hamogu:
The timing seems to be dominated by the install and `reg_proj`, so I suggest we invest the time.

The figure problem turned out to be the sum of not clearing the figure and not having enough space between the panels.

hamogu:
Looking that the last few RTD runs **not** including this PR I see times around 13 min, but with considerable changes up or down:

- PR 2022: 13:34 min
- PR 2022: 11:53 min
- latest: 12:58 min
- PR 2088: 12:53 min
- PR 2083: 12:59 min
- latest: 12:50 min

While this PR has in the last few runs (I don't think the changes I made should have effected runtime):
- 12:39 min
- 12:26 min

So it seems that this PR does not make RTD builds significantly longer, in line with the runtimes I find locally.

DougBurke:
Ta for checking.

.github/workflows/ci-pip-workflow.yml
```diff
@@ -50,6 +50,7 @@ jobs:
             test-data: package
             matplotlib-pkg: 'matplotlib>=3,<4'
             bokeh-pkg: 'bokeh>=3,<4'
+            arviz-pkg: 'arviz'
```
DougBurke, line 53: We should really re-evaluate all these builds - and the names they use - but that's for a different PR!

sherpa/sim/__init__.py
```diff
@@ -668,6 +668,10 @@ def get_draws(self, fit, sigma, niter=1000, cache=True, rng=None):
            values.
         niter : int, optional
            The number of draws to use. The default is ``1000``.
+        cache : bool, optional
```
DougBurke, line 671: As an aside: this is somewhere I wanted to look at the cache handling, given 

- #2166 
- #2774

but it looks like the code already "should work", which is nice.

There's the difference between turning on/off the cache during a fit/mcmc run/error analysis and the general purpose "always cache" behavior which we are struggling to document/handle correctly, but that's a different concern.

sherpa/sim/__init__.py
```diff
@@ -1004,3 +1008,78 @@ def call(self, niter, seed=None, rng=None,
...
+            raise ValueError("For each chain in the list_of_draws, there must be three elements: " +
+                             "the statistic, the acceptance rate, and the parameter array.")
+        if draw[2].shape[0] != len(pnames):
+            raise ValueError(f"Chain {i} contains {draw[2].shape[0]} dimensions, but the model has {len(pnames)} thawed parameters.")
+        if len(draw[0]) != draw[2].shape[1]:
+            raise ValueError(f"Chain {i} contains {draw[0].shape[0]} steps for the static, but parameter array has {draw[2].shape[1]} steps.")
+        if len(list_of_draws[0][0]) != len(draw[0]):
+            raise ValueError("All chains must have the same number of steps. " +
+                f"Chain {i} contains {draw[0].shape[0]} steps, but the first chain has {len(list_of_draws[0][0])} steps.")
+
```
DougBurke, line 1059: It would help here if we had a way to structure the MCMC data into something more structured than a bunch of tuples, but that is a future-us issue!

docs/mcmc/index.rst
```diff
@@ -298,34 +318,58 @@ the results match reality::
...
+    :context: close-figs
 
+    >>> plt.scatter(pvals[0, :], pvals[4, :], alpha=0.3)
     >>> from sherpa.plot import RegionProjection
     >>> rproj = RegionProjection()
     >>> rproj.prepare(min=[95, 1.8], max=[150, 2.6], nloop=[21, 21])
     >>> rproj.calc(f, mdl.r0, mdl.alpha)
     >>> rproj.contour(overplot=True)
-    >>> plt.xlabel(r'$r_0$'); plt.ylabel(r'$\alpha$')
+    >>> _ = plt.xlabel(r'$r_0$'); plt.ylabel(r'$\alpha$')
```
DougBurke, line 348:
```suggestion
    >>> plt.xlabel(r'$r_0$')
    >>> _ = plt.ylabel(r'$\alpha$')
```

If you do need to suppress the output then I'd have thought the line should have been

```
plt.xlabel(..); _ = plt.ylabel(..)
```

(unless my understanding of how multi-line statements are handled is wrong, which is possible) but I think the need to "hide" the return value means that we should just split up these two lines.

docs/mcmc/index.rst
```diff
@@ -298,34 +318,58 @@ the results match reality::
...
+
+Connections to ArviZ
+--------------------
+`ArviZ <https://python.arviz.org>`_ is a Python package for exploratory analysis of Bayesian models.
+It serves as a backend-agnostic tool for diagnosing and visualizing Bayesian inference and
+provides different plotting and statistical diagnostic functions for MCMC chains, that
+are not implemented in Sherpa itself. If the ArviZ package is installed,
+`~sherpa.sim.mcmc_to_arviz` can be used to convert the MCMC results to an ArviZ InferenceData object.
```
DougBurke, line 356:
```suggestion
`~sherpa.sim.mcmc_to_arviz` can be used to convert the MCMC results to an `ArviZ InferenceData object <https://python.arviz.org/en/latest/api/generated/arviz.InferenceData.html>`_.
```

[although this does tie us to the "current" sphinx rules for how their pages are generated for a particular term, I don't see that changing for a while]

assuming I've not messed up the markup

hamogu: Even better - I'm loading intersphinx for arvix now for exactly that reason. I just forgot to mark this up!

docs/mcmc/index.rst
hamogu: We should split the lines in any case, there is not reasons that needs to be on a single line.

I was hiding output in other places (and then there, but didn't even notice that I had a double statement on a single line) because printing the return values to the screen annoyed me when I played with this example. But it seems that the return values are not rendered on RTD anyway because the entire code is inside a `.. plot` directive, so it's probably more confusing for users to see `_ =` in the docs than it would be if they had some additional output when they copy the example line by line into an interpreter. 

However, our tests with doctest fail if I don't capture the return values in some way.

docs/install.rst
```diff
@@ -494,7 +496,7 @@ build the documentation.
 
 If conda is being used then the other packages can be installed with::
 
-  conda install -c conda-forge pandoc graphviz
+  conda install -c conda-forge pandoc graphviz arviz
```
DougBurke, line 499: Technically we should not need `arviz` since it should have been installed with `pip install .[doc]`, but I don't think it harms us to keep it here.

sherpa/sim/__init__.py
```diff
@@ -1004,3 +1008,78 @@ def call(self, niter, seed=None, rng=None,
...
+def mcmc_to_arviz(mcmc : MCMC, fit: Fit, list_of_draws : list[tuple],
+                  ):
+    """Convert the MCMC results to an ArviZ InferenceData object.
+
+    ArviZ is a Python package for exploratory analysis of Bayesian models and
+    provides a range of tools for visualizing and analyzing MCMC results beyond
+    what is implemented in Sherpa itself.
+
+    This function requires the `ArviZ <https://python.arviz.org>`_ package to be installed.
+
```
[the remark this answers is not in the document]
DougBurke: The reason I suggest this is because I really find it annoying when I go to documentation and go "oh, I can use that", then find I can't because the version of the library/package I'm using is too old (or when it suddenly starts failing when I change machine / environment / ...).

I also find it useful to be able to quickly go through and find all the new, or changed, features we've made in a release!

docs/install.rst
hamogu: you are right.

sherpa/sim/__init__.py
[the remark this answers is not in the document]
hamogu: I forgot that this morning. Maybe I needed tea first... Coming.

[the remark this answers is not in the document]
DougBurke: Well, don't come here then as we are sorely lacking on hot drinks: 

drinks

DougBurke:
Look at that awesomeness:

image

docs/mcmc/index.rst
DougBurke: Don't you just love consistent systems .....

DougBurke (approved):
LGTM (assuming the remaining tests are all green).

Sorry for the pickiness.

sherpa/sim/tests/test_sim.py
```diff
@@ -561,3 +561,100 @@ def test_metropolisMH(setup, caplog):
 
     means = np.asarray([1.06278015, 9.21533855, 2.5736483, 2.5853907, 47.27058904])
     assert params.mean(axis=1) == pytest.approx(means)
+
+
+def setup_no_fit():
+    """Simplified version of the setup fixture that does not perform a fit or confidence
+    to save time. THis is just use to test error messages for non-conforming input."""
```
wmclaugh, line 568:
```suggestion
    to save time. This is just use to test error messages for non-conforming input."""
```

PR #2287, opened 2025-04-21
Use python 3.10 syntax for typing for the ui modules

DougBurke:
# Summary

Use Python 3.10 syntax for typing statements in the ui modules. There are no functional changes in these commits.

# Details

Switch to newer forms of the typing rules

- import symbols from collections.abc rather than typing (Sequence and Callable), first allowed in Python 3.9 I believe
- replace Union[x, y] and Optional[x] with x | y and x | None, which was first allowed in Python 3.10

Although it is likely that the old forms will remain supported for a number of Python releases, it makes sense to avoid technical debt when we can. There is also the fact that the ui routines are the ones that many users will interact with [see below].

I have made these changes in several recent PRs so I thought it worthwhile to just pull them out into a separate easy-to-review PR.

# Example

In CIAO 4.17

```
sherpa> help(get_data)
Help on function get_data in module sherpa.astro.ui:

get_data(id: 'Optional[IdType]' = None) -> 'Data'
    Return the data set by identifier.
    
    The object returned by the call can be used to query and
    change properties of the data set.
...
```

With this PR

```
>>> help(ui.get_data)
Help on function get_data in module sherpa.astro.ui:

get_data(id: 'IdType | None' = None) -> 'Data'
    Return the data set by identifier.

    The object returned by the call can be used to query and
    change properties of the data set.
...
```

Isn't it so much better?

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015 - 2025
+#  Copyright (C) 2010, 2015-2025
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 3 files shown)


hamogu (approved):
This is exciting and looks so good.  Let's go for it!

PR #2288, opened 2025-04-21
ui: add more examples of using SherpaVerbosity

DougBurke:
# Summary

Point out the use of the sherpa.utils.logging.SherpaVerbosity context manager in the documentation of the fit and error-analysis UI routines.

# Details

The `SherpaVerbosity` context manager makes it easy to turn off the output from routines, but it is hard to find if you do not know it exists. So add some more examples to the UI code showing its use.

I looked to see if there were other places and the only other obvious place was `get_draws`, but the screen output from that is small. There may be some of the flux simulation/error analysis places that maybe it could be used, but again I think most of the time the screen output is only a few lines. We could add text, but I didn't want to waste the reader's time on stuff they do not need.

### DIFF
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2015 - 2025
+#  Copyright (C) 2010, 2015-2025
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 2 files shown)


DougBurke:
We could add examples for using `SherpaVerbosity` with `load_data` and `load_pha`, but I want the users to read those messages!

anetasie (approved):

PR #2289, opened 2025-04-23
Update http -> https

### LINKED ISSUE #2172: doc: update http to https
should update to `https` (probably the ADS link before it too).

https://github.com/sherpa/sherpa/blob/ae169c85481920960353e621865c73c8e7fe26c1/sherpa/astro/models/__init__.py#L1335

hamogu:
## Summary
Update links in docs and source code from http -> https where possible

## Details

All servers are hand-checked at least once and in a number of cases the servers or specific pages do no longer exist. In those cases, a best available replacement link is used.

No changes are applied to "extern" or where the copyright of files places limits on how they can be modified.
Similarly, comments in C++ code are generally not updated and for comments in Python code, judgment is used e.g. versioneer, which is just vendored from an external project) is not touched.

closes #2172

## Notes
Yes, I know this is exciting to review. I suggest to not redo my work to check every server in review (when I say "every server", I mean "I check that the ads supports https, but I did not check every single paper on ADS that is referenced").

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,8 +31,6 @@ from sphinx_astropy.conf.v1 import intersphinx_mapping, default_role
 import sphinx_rtd_theme
 
 
-# Based on http://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
-#
 from unittest.mock import MagicMock as BaseMock
 
 
@@ -41,7 +39,7 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 runtime_dir = os.path.split(os.getcwd())[1]
 
 
-# I found this somewhere (probably the rtd link given above). I was
+# I found this somewhere. I was
 # hoping it would support building with Python 2.7 but it doesn't seem
 # to. I have left it in for now, even though Python 2.7 is no longer
 # supported, in case it is being used somewhere.
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -215,7 +215,7 @@ def calc_ftest(dof1, stat1, dof2, stat2):
     .. [1] Protassov et al., Statistics, Handle with Care: Detecting
            Multiple Model Components with the Likelihood Ratio Test,
            Astrophysical Journal, vol 571, pages 545-559, 2002,
-           http://adsabs.harvard.edu/abs/2002ApJ...571..545P
+           https://adsabs.harvard.edu/abs/2002ApJ...571..545P
 
     Examples
     --------
@@ -286,7 +286,7 @@ def calc_mlr(delta_dof, delta_stat):
     .. [1] Protassov et al., Statistics, Handle with Care: Detecting
            Multiple Model Components with the Likelihood Ratio Test,
            Astrophysical Journal, vol 571, pages 545-559, 2002,
-           http://adsabs.harvard.edu/abs/2002ApJ...571..545P
+           https://adsabs.harvard.edu/abs/2002ApJ...571..545P
 
     Examples
     --------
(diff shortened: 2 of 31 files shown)


anetasie (approved):

docs/conf.py
```diff
@@ -31,7 +31,7 @@
 import sphinx_rtd_theme
 
 
-# Based on http://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
+# Based on https://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules
```
wmclaugh, line 34: This gives a 404. The longest stem of the URL that works is  https://read-the-docs.readthedocs.io/en/latest/

sherpa/utils/__init__.py
```diff
@@ -1752,7 +1752,7 @@ def quantile(sorted_array, f):
 
     Based on the description of the GSL routine
     gsl_stats_quantile_from_sorted_data - e.g.
-    http://www.gnu.org/software/gsl/manual/html_node/Median-and-Percentiles.html
+    https://www.gnu.org/software/gsl/manual/html_node/Median-and-Percentiles.html
```
wmclaugh, line 1755: perhaps the link is now- https://www.gnu.org/software/gsl/doc/html/statistics.html#median-and-percentiles

sherpa/utils/akima.py
```diff
@@ -47,8 +47,8 @@
 Requirements
 ------------
 
-* `Python 2.7 or 3.2 <http://www.python.org>`__
-* `Numpy 1.6 <http://numpy.scipy.org>`__
+* `Python 2.7 or 3.2 <https://www.python.org>`__
+* `Numpy 1.6 <https://numpy.scipy.org>`__
```
wmclaugh, line 51: connection to https://numpy.scipy.org times out but https://scipy.org/ works

docs/conf.py
hamogu: Let's just delete it altogether. It's not a reference in the sense of "giving credit" like we have with the numerical algorithms, it's more for developers to look up how this works and if we can't find how it works, we can just not have it.

sherpa/utils/akima.py
hamogu: Since this lists dependencies, it's numpy.org now that we want. But numpy 1.6 is of course years out of date, si this is more of a historical record now anyway.

wmclaugh: Perhaps we could change it to 'https://docs.scipy.org/doc/numpy-1.6.0/reference/'  which points to the old numpy 1.6 documentation (and dates it to May 15, 2011) and makes the association of numpy and scipy
.

hamogu: It doesn't actually use scipy though, it only uses numpy. So it really depends on who our audience is: A developer with a historical interest or a user/developer who gets stuck, and reads the code to see what dependences are listed to get it to run today. I think neither will ever happen in practice, so I don't really care; either is fine with me.

(I'm not going back to test that now or go back through git blame, but it's entirely possible that this code would not run with numpy 1.6 anymore. We changed a lot of file to make them numpy 2.0 compatible and I don't think it's worth the time to check how far back this one will work.)

wmclaugh: I am good with merging this but the github API is currently stuck 'Checking for the ability to merge automatically...'

hamogu: Let me manually rebase and force push. That should trigger it again. I won't do any changes, so we don't have to wait (and waste the electricity) for the tests to run through again. (I'll see if the `[skip CI]` still works.)

hamogu: Skipped GH checked, it's still trying to build RTDs (not sure how to skip that). It passed before and I don't see who that would have changed, but your call if you want to wait for that.

PR #2290, opened 2025-04-23
Minor clean up of utility code

DougBurke:
# Summary

Ensure a deprecated routine can still be used and remove a routine that has never worked.

# Details

Pulled out of #2283 

- add a test for contour preference handling (just to cover a corner case)
- fix a bug, but it is in a deprecated routine (marked as such in 4.17.0, when the bug was introduced) which we do not use and I would be surprised if any external user does
- remove a routine that has been buggy (i.e. will always fail) for the entire time sherpa has been stored in git

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2616,6 +2616,8 @@ def func_counter(func):
     warnings.warn("func_counter is deprecated in 4.17.0: use FuncCounter instead",
                   FutureWarning)
 
+    nfev = [0]
+
     def func_counter_wrapper(x, *args):
         nfev[0] += 1
         return func(x, *args)
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -2565,17 +2565,6 @@ def symmetric_to_low_triangle(matrix, num):
     return low_triangle
 
 
-############################### Root of all evil ##############################
-
-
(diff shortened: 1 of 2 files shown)


sherpa/utils/__init__.py
```diff
@@ -2616,6 +2605,8 @@ def func_counter(func):
     warnings.warn("func_counter is deprecated in 4.17.0: use FuncCounter instead",
                   FutureWarning)
 
+    nfev = [0]
+
```
anetasie, line 2609: Why line 2608 was added? is this to fix a bug?

DougBurke: Yes, as mentioned

- in the details section above
- in the commit log that made the change: https://github.com/sherpa/sherpa/pull/2290/commits/fd0ddbf1dd38b5300bc20302e8dffd810208f96d

DougBurke:
@hamogu - do you think this is worth reviewing for the 4.17.1 release (and do you have time)? None of the other PRs I have open are worth considering at this late stage.

hamogu (approved):
Sure, let's take this. Three uncontroversial changes that clean up some stuff.

DougBurke:
@wmclaugh  This is not something we must have for 4.17.1 (the changes from @hamogu are more important) if there's a decision about what to spend time on!

PR #2292, opened 2025-04-26
Docs for Cash and CStat are wrong about truncate_value

hamogu:
`stats.hh` contains the following lines for both CStat and Cash:
```
    for ( IndexType ii = num - 1; ii >= 0; --ii ) {

      if ( model[ ii ] > 0.0)
	mymodel = model[ ii ];
      else {
	if (trunc_value > 0)
	  mymodel = trunc_value;
	else
	  return EXIT_FAILURE;
      }
```
Clearly, the truncate_value is applied when the MODEL (not the DATA, as the docs state before this PR) is <=0.
(The data in fact has to be >=0, otherwise the likelihood makes no sense anyway.)

### DIFF
--- a/sherpa/stats/__init__.py
+++ b/sherpa/stats/__init__.py
@@ -455,7 +455,7 @@ class Cash(Likelihood):
     .sherpa.rc file:
 
     - if `truncate` is `True` (the default value), then
-      `log(trunc_value)` is used whenever the data value is <= 0.  The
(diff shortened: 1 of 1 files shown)


hamogu:
We shouldn't worry about it, we should just fix it. (It could be done in #2294 as I found it when I worked n that, but I figured that we can just approve and merge this, while #2294 will probably need a lot more discussion).

DougBurke (approved):

PR #2293, opened 2025-04-27
Improve docs for hard and soft limits

### LINKED ISSUE #1480: Clarification of soft vs hard limits
Is the following valid, @dtnguyen2 ?

- Model parameters have soft and hard limits, where the hard limits are either the same size, or larger than, the soft limits. That is a soft limit may be more restrictive than the hard limit.

- The optimizer will respect the soft limits when fitting data. Some optimizers are unconstrained, and for these Sherpa uses a "trick" to handle these constraints (something about an infinite well).

- The error analysis (e.g. confidence, is it relevant for covariance?) use the hard limits instead of the soft limits. 

We have in #1448 support to better-hand;e parameters who hit one of the limits, but I'm not sure if this is talking about the soft or hard limits. I had assumed it was the hard limits but the behavior in https://github.com/sherpa/sherpa/pull/1448#issuecomment-1090474493 suggests it is being applied to "whatever limits are currently in use" (but I could be mis-reading things).

hamogu:
Hard and soft limits are already explained in the docs, but some details that might be useful for developers extending or building on Sherpa have been added. Since the audience of the added information are developers, and this is not typically an area that users need ot care about, the information has been added to a module docstring. I also made some of the relevant module level constants public - a developer might want to reset them.

closes #1480

### DIFF
--- a/docs/optimisers/optfcts.rst
+++ b/docs/optimisers/optfcts.rst
@@ -19,3 +19,6 @@ The sherpa.optmethods.optfcts module
      minim
      montecarlo
      neldermead
+     InfinitePotential
+     EPSILON
+     FUNC_MAX
--- a/sherpa/optmethods/optfcts.py
+++ b/sherpa/optmethods/optfcts.py
@@ -20,20 +20,23 @@
 
 """Optimizing functions.
 
-These functions take a callback, the current set of parameters, the
-minimum and maximum parameter ranges, along with optional arguments,
-and return a tuple containing
+As input, these functions take
+  - a callback function, which should return the current statistic value
+    and an array of the statistic value per bin.
+  - the current set of parameters (a numpy array)
+  - the minimum for each parameter (numpy array)
+  - the maximum for each parameter (numpy array)
+  - any optional arguments
 
-    status, parameters, statistic, message, dict
+The return a tuple containing
 
-where ``status`` is a boolean indicating whether the optimisation
-succeeded or not, parameters is the list of parameter values at the
-best-fit location, the statistic value at this location, a string
-message - when ``status`` is `False` this will give information on the
-failure - and a dictionary which depends on the optimiser.
+ - a boolean indicating whether the optimization succeeded or not
+ - the list of parameter values at the best-fit location
+ - the statistic value at this location
+ - a string message - when ``status`` is `False` this will give information
+   on the failure
+ - a dictionary which depends on the optimizer
 
-The callback should return the current statistic value and an array
-of the statistic value per bin.
 
 Notes
 -----
@@ -66,6 +69,16 @@ model value of 1 and bounded to the range 0 to 10000:
 >>> print(f"Best-fit value: {res[1][0]}")
 Best-fit value: 4.0
 
+How are parameter bounds implemented?
+-------------------------------------
+The optimizers in this module use a simple Infinite Potential
+approach, where the value of the statistic is set to a very large
+number defined by the module level variable `FUNC_MAX`
+if any of the parameter values is outside the limits.
+
+This approach is easy to implement and fast to evaluate, but it does
+introduce a discontinuity at the limits which can in some cases cause
+problems when the best-fit value is close to the limits.
 """
 
 from collections.abc import Sequence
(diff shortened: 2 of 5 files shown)


sherpa/optmethods/optfcts.py
```diff
@@ -20,20 +20,23 @@
...
-succeeded or not, parameters is the list of parameter values at the
-best-fit location, the statistic value at this location, a string
-message - when ``status`` is `False` this will give information on the
-failure - and a dictionary which depends on the optimiser.
+ - a boolean indicating whether the optimization succeeded or not
+ - the list of parameter values at the best-fit location
+ - the statistic value at this location
+ - a string message - when ``status`` is `False` this will give information
+   on the failure
+ - a dictionary which depends on the optimizer
```
DougBurke, line 38: I think it's correct to say something like

```suggestion
 - a dictionary which depends on the optimizer and may be empty, or `None`.
```

sherpa/optmethods/optfcts.py
```diff
@@ -85,20 +98,22 @@
 
 
 __all__ = ('difevo', 'difevo_lm', 'difevo_nm', 'grid_search', 'lmdif',
-           'minim', 'montecarlo', 'neldermead')
+           'minim', 'montecarlo', 'neldermead',
+           'InfinitePotential', 'EPSILON', 'FUNC_MAX',
```
DougBurke, line 102: I would rather we don't export these by default yet. Not until we've gone through and resolved #2022.

docs/optimisers/optfcts.rst
```diff
@@ -19,3 +19,6 @@ The sherpa.optmethods.optfcts module
      minim
      montecarlo
      neldermead
+     InfinitePotential
```
DougBurke, line 22: As mentioned below, I'd rather not expose these just yet.

sherpa/optmethods/optfcts.py
```diff
@@ -20,20 +20,23 @@
 
 """Optimizing functions.
 
-These functions take a callback, the current set of parameters, the
-minimum and maximum parameter ranges, along with optional arguments,
-and return a tuple containing
+As input, these functions take
```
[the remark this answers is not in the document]
hamogu: empty line was missing before list items.

hamogu (commented):
I addressed the comments locally, so that I could test the sphinx build.

Only te last commit is new, te earlier ones just changed the commit hash because I took the opportunity to rebase on main.

sherpa/optmethods/optfcts.py
```diff
@@ -85,20 +98,21 @@
...
+           )
 
 
 warning = logging.getLogger(__name__).warning
 
-# Use FLT_EPSILON as default tolerance
-#
+
 EPSILON = np.float64(np.finfo(np.float32).eps)
+'''Use FLT_EPSILON as default tolerance'''
```
DougBurke, line 109: Is this a helpful documentation string? To me, I have no idea what `FLT_EPSILON` means (I know this is just taken from the original comment, and I think I can work out it means "float 32" epsilon value, but even then as as a part of the documentation does it help). 

Now, looking through other parts of the documentation in this module you can see both `FLT_EPSILON` and `DBL_EPSILON`, so maybe the nomenclature makes sense (and in fact, there are a few other modukes that use one or the other nomenclature).

However, we don't export this symbol, so do we need to add a docstring? This also holds for `FUNC_MAX` (now, this is used from another module, but it is also not included in `__all__`).

Is it useful for people using code-completion/LSP-like stuff (e.g. it can get displayed to the user in their favourite editor)?

DougBurke: [at some level this doesn't really matter, as docstring vs comment doesn't really change the understanding of the code; just wanted to understand the use of docstring here]

hamogu: Many editors (e.g. VSCode which I use) will display the docstring for a symbol when you mouse over it at any location in the code, while they will not display a comment that maybe found in the vicinity of the definition unless you scroll up in the file to see it.

I agree that this comment/docstring is not useful for a user, but, as you say, they won't see it anyway because it's not in `__all__`. However, for a developer who works on this module knowing that this has something to do with the default tolerance *might* be helpful. Sure, a longer text would be even more helpful, but I don't think it's warrented.

PR #2294, opened 2025-04-28
Add penalty for negative models to cstat

### LINKED ISSUE #836: Problem fitting cash/cstat is model goes negative
Fit does terminate at some random negative number, when it should give me a (small) continuum and a (small) line flux.

Data to reproduce is in /pool14/guenther/sherpa_simplex_fit
```
from sherpa.astro import ui

ui.load_data('P0112880201R2S005SRSPEC1003.FTZ')
ui.load_rmf('P0112880201R2S005RSPMAT1003.FTZ')
ui.set_stat('cash')
ui.set_analysis('wave')
ui.group_counts(1)
ui.ignore(None,11.4)
ui.ignore(11.7, None)
ui.notice(11.4, 11.7)

bkg = ui.const1d(name='bkg')
bkg.c0 = 7e-4
line = ui.delta1d(name='line')
line.pos = 11.5467
line.pos.frozen = True
line.ampl = 5e-2
ui.set_source(bkg + line)


ui.set_method('simplex')
ui.fit(outfile='simplex.dat', clobber=True)

ui.plot_fit()
```
Looking at simplex.dat the problem seems to be that the value of the statistic is constant when the model predicts a negative flux. Thus, once the fitter "oversteps" and one parameter becomes negative, there is no incentive for the fitter to move into the right direction (towards positive values) again. I think that's a pretty common problem with any non-physical model (i.e. a powerlaw or apex model won't go negative, but a constant or a Gaussian can). My first idea would be to set lower bounds to prevent this but (a) that gives me the problem in #835 and (b) that's not always easy to do (e.g. if I fit an absorption line on a continuum, the amplitude of the absorption line that is allowed ( = does not produce negative model values) depends on the current value of the constant which is fitted at that same time).

Thus, my question: Can we modify the statistics in such a way that the result for negative model values becomes larger, the further the model is below 0? I understand that we did not invent those statistics, but to me this is a question of numerical optimization, not just how the statistic is defined. In some sense, sherpa should not return a "fit" in the above example, but raise an error (e.g. "Statistic not valid for negative models"). On the other hand, numerical optimizers do use finite step sizes, so a step to a value below 0 can happen if the true parameter is close to 0. A penalty on the statistic that increases with increasing distance from zero would help to nudge the fit in the right direction. 

Note that the setup here is very similar to #835 and uses the same data, but the underlying problem is totally different: In #835 the problem is with the optimizer, here it is with the definition of the statistic.

hamogu:
# Summary

Add a version of the CStat statistic that provides a variable penalty if the model prediction is negative, rather than use a constant term. Fix #836.

# Details

See below.

---

## This PR has to things in it

Since GH does not have threaded discussions on PRs, I'm simply going to label them 1 and 2.

### Motivation
In the past, I have observed (#836) that my fits get stuck with CSTAT or CASH when one or more **model** bins are negative. In this code, I'm trying to fix that by changing how CSTAT handles those bins. 

### CStat with negative bin in the **model**
CStat as defined in the literature will not work with model values that are 0 or negative because of the term log(1/model). In Sherpa CStat, we deal with that by truncating the model values to a small positive value. (At least that's the default. An error cna be raised instead if the user sets the truncate option to False in the .sherpa.rc file.)

THIS IS ALREADY A CRUDGE, just to make it work numerically. In practice, most model will have mostly positive values all the time, but in numerical optimization it can happen that the optimizer tries e.g. an absorption lines that is too deep and get a few bins in the model to a negative value. That's usually not a good fit (since the data is always >=0) and the optimizer would usually step the parameter such that the model is positive again. So, we use the truncation to make sure that it doesn't crash in the middle of the optimization.

### Adding an additional penalty

This solution is incomplete though: Once a value is negative the statistics do not change with how negative it is. A model value of 0 and -10000 will be clipped to the same number and thus the optimizer has no incentive to walk back to positive values.

Again, in many models that is not a problem, e.g. in a Gaussian absorption line that is several bins wide, reducing the amplitude of the line will improve the fit on the edges of the line, even if the center of the line is still truncated and thus the fit will walk back to positive values. But in some cases, e.g. a narrow Gaussian line there might be only one one bin that is negative and stays negative for a range of parameter values. This can happen in two cases:

- The initial guess for the model is wrong. While we typically do not  start with negative values for Poisson data, that can happen accidenitally, e.g. we might describe a model by a continuum and an absorption line. If the absorption line is too deep, the initial model might be negative.
- We start with a model that is all positive, but then the optimizer takes a big step on the absorption line parameter and ends up in negative territory.

So, in addtion to the truncation, this class adds a penalty for negative model values that grows to more negative the value is.
THIS IS STILL A CRUDGE and not grounded in mathematical statistics, it's just a way to make the optimizer work better in some cases. As before, we expect that the optimizer will walk back to positive values and thus the penalty in intermediate steps will not influence the statistic value in the final fit at all.

The scale of the penalty needs to be chosen such that it actually edges the optimizer back to positive values, but not so strong that it creates a giant step in the statistic value which might confuse the optimizer, i.e. we don't want it to be 1e-38, but also not 1e+23. For Poisson data, the natural scale is "1", so we use penalt the sum of all negative model values as penalty.

Again, the exact choice does not matter, because we expect this penalty to apply only in intermediate steps and not in the final fit.

### Example
Below, I'm going to use the following dataset (simplified from the example in #836) and fit is with a constant + absorption line with fixed position:

image

```Python
import numpy as np
from sherpa import data,models, fit, optmethods, stats

data = data.Data1D('data', np.arange(10), np.array([2,3,2,3,0,1,1,3,5,2]))

bkg = models.Const1D('bkg')
line = models.Delta1D('line')
line.pos = 5.
line.pos.frozen = True
source = bkg + line
```

First, I want to visualize what the statistics looks like:
```Python
cstat = stats.CStat()
cstatpenalty = stats.CStatNegativePenalty(name='cstatpenalty') # using the code from this branch

amplgrid, c0grid = np.mgrid[-5: 4: .1, -3: 7: .1]
statgrid = np.zeros(c0grid.shape)
penaltygrid = np.zeros(c0grid.shape)
it = np.nditer(c0grid, flags=['multi_index'])
for x in it:
        bkg.c0 = c0grid[it.multi_index]
        line.ampl = amplgrid[it.multi_index]
        statgrid[it.multi_index] = cstat.calc_stat(data, source)[0]
        penaltygrid[it.multi_index] = cstatpenalty.calc_stat(data, source)[0]

import numpy as np
from astropy.visualization import simple_norm

# Create an ImageNormalize object
norm = simple_norm(statgrid, 'log')
norm2 = simple_norm(penaltygrid - statgrid, 'log')


# Display the image
fig, axes = plt.subplots(ncols=2, figsize=(10, 6))

im = axes[0].imshow(statgrid, origin='lower', norm=norm, 
               extent=[c0grid.min(), c0grid.max(), amplgrid.min(), amplgrid.max()])

im2 = axes[1].imshow(penaltygrid - statgrid, origin='lower', norm=norm2, 
               extent=[c0grid.min(), c0grid.max(), amplgrid.min(), amplgrid.max()])

fig.colorbar(im, ax=axes[0])
fig.colorbar(im2, ax=axes[1])

for ax in axes:
    ax.set_xlabel('c0')
    ax.set_ylabel('ampl')

axes[0].set_title('CStat')
axes[1].set_title('Penalty')
plt.tight_layout()
```

image

*left*: CSTAT. The high numbers are in the region where the constant goes negative - which means all model values are negative and are replaced with the "truncation value" leading to a large jump in the statistic onto a constant level. The teal triangle comes from the one bin at location 5 going negative (when ampl > c0), so the statistic is not as high  (because most model bins are still positive).

*right*: The penalty that I propose to add so that yellow bar and the teal triangle are not longer constant, but have a slope down towards the correct region.

### Does that help the fits converge?
In the next few plots, I compare the path that the default CStat (orange) and my modified version with added penalty (blue) take in the parameter space. Obviously, that depends on the optimizer, so I'm showing two different panels for LevMar and Simplex.

First, let's start in legal territory, where all models bins are always positive. the orange and blue lines are exactly on top of each other, showing that my penalty has no effect on fits that stay in the territory where the Cstat is properly defined. This should represent the vast majority of uses, i.e. most people and most fits won't see a difference from implementing this (code to generate all those plots is added at the bottom of this PR):

image

OK, now lets start the fit in the "wrong" region for a few different starting points (note that LevMar fails with default setting for step size etc. in all cases, thus the left penal only has a single orange dot or very short line:

image
image
image
image

In all those cases, the simplex with penalty find the minimum, while the simplex with default CStat usually does not.

So, is this a perfect solution? No. But I think it's better than what we have now so we should take it.

*The same argument applies to Cash, but for demonstration I've implemented it just for one statistic here.*

## 2) Statistic calculations in Python instead of C

For this PR, I've moved the code to actually calculate CStat from C Python. The advantage is that I don't have to recompile every time I want to try out a different type of penalty, and the numpy version of the code is easier to read (and thus to extend, modify, review) than the C version. While this piece of C code is relatively simple, less complied code also make the installation easier. We could move computations from C to Python independent of my change of the Cstat, and we could also  modify the C code to include the penalty I'm suggesting here and keep the calculation in C. 

*In that sense, choosing whether to implement in C or Python is independent of if we want the penalty I'm suggesting above.*

But, isn't C faster? The following figure compares the runtime of just calculating the statistics value (since everything else is done the same way in Python in either case).
image

Only for small arrays. For spectra with ~1000 bins (typical for X-ray analysis) the difference between C and Numpy is only a factor of 2, of order 1e-5 s. That's hardly noticeable on the scale of a fit. For large arrays, Numpy and C are comparable.

For reference, the code that makes the plot above:
```Python
import numpy as np
import perfplot
from sherpa import stats

def setup_model(n):
    model = np.random.uniform(0.2, 10, n)
    data = np.random.poisson(np.maximum(model, 0))
    return (data, model)

cstat = stats.CStat()
cstatnumpy = stats.CStatNumpy()
cstatpenalty = stats.CStatNegativePenalty()

perfplot.show(
    setup=setup_model,
    kernels=[
        lambda data, model: cstat._calc(data, model, None, stats.truncation_value),
        lambda data, model: cstatnumpy._calc(data, model, None, stats.truncation_value),
        lambda data, model: cstatpenalty._calc(data, model, None, stats.truncation_value),
    ],
    labels=["C", "Numpy", "NumpyPenalty"],
    n_range=[2**k for k in range(1, 18)],  # 23
    xlabel="len(a)",
    # More optional arguments with their default values:
    # logx="auto",  # set to True or False to force scaling
    # logy="auto",
    # equality_check=np.allclose,  # set to None to disable "correctness" assertion
    # show_progress=True,
    # target_time_per_measurement=1.0,
    # max_time=None,  # maximum time per measurement
    # time_unit="s",  # set to one of ("auto", "s", "ms", "us", or "ns") to force plot units
    # relative_to=1,  # plot the timings relative to one of the measurements
    # flops=lambda n: 3*n,  # FLOPS plots
)
```

## And, for reference the code that I used to make the plots with the optimizer paths

```Python
def test_cstatpenalty(c0, ampl, opt=optmethods.NelderMead()):
    bkg.c0 = c0
    line.ampl = ampl
    fitter = fit.Fit(data=data, model=source, stat=cstat, method=opt)
    res = fitter.fit(record_steps=True)
    
    bkg.c0 = c0
    
    line.ampl = ampl
    fitter = fit.Fit(data=data, model=source, stat=cstatpenalty, method=opt)
    resp = fitter.fit(record_steps=True)
    
    return res.extra_output['record_steps'], resp.extra_output['record_steps']

def plot_path(c0, ampl, opts = [optmethods.LevMar(), optmethods.NelderMead()]):

    fig, axes = plt.subplots(ncols=len(opts), figsize=(10, 6),
                             sharex=True, sharey=True)

    for ax in axes:
        ax.set_xlabel('ampl')
        ax.set_ylabel('c0')
        im = ax.imshow(statgrid, origin='lower', norm=norm, 
               extent=[c0grid.min(), c0grid.max(), amplgrid.min(), amplgrid.max()])

    for ax, o in zip(axes, opts):
        for p in test_cstatpenalty(c0=c0, ampl=ampl, opt=o):
            ax.plot(p['bkg.c0'], p['line.ampl'])
            ax.set_title(o.name)

        ax.set_xlim(c0grid.min(), c0grid.max())
        ax.set_ylim(amplgrid.min(), amplgrid.max())
        ax.set_xlabel('c0')
        ax.set_ylabel('ampl')

plot_path(ampl=3.4,c0=5.)
plot_path(ampl=-2,c0=-1)
plot_path(c0=1, ampl=-4)
plot_path(c0=-1, ampl=-3)

# Make larger, but coarser grid
amplgrid, c0grid = np.mgrid[-25: 25: 1, -25: 25: 1]
statgrid = np.zeros(c0grid.shape)
penaltygrid = np.zeros(c0grid.shape)
it = np.nditer(c0grid, flags=['multi_index'])
for x in it:
        bkg.c0 = c0grid[it.multi_index]
        line.ampl = amplgrid[it.multi_index]
        statgrid[it.multi_index] = cstat.calc_stat(data, source)[0]
        penaltygrid[it.multi_index] = cstatpenalty.calc_stat(data, source)[0]

plot_path(c0=-20, ampl=-10)
fig = plt.gcf()
fig.axes[0].set_xlim(-25, 15)
fig.axes[0].set_ylim(-15, 5)
```


hamogu:
@anetasie @DougBurke This is obviously not the the release that gets frozen today, but for the one after that. However, I anticipate that we need some discussion about this, please have a look and let me know what you think.

DougBurke:
just for reference there's `Cash`, `CStat`, and `WStat` that we'd be worried about here (just so we don't forget).

hamogu:
We should look at `WStat`, although the mechanism there is slightly different as it doesn't use the `truncation_value`. However, in the docstring for `WStat` it says 

> In practice, it works well for many cases but for weak sources can  generate an obviously wrong best fit. It is not clear why this happens although binning to ensure that every bin contains at least one count often seems to fix the problem."

Thtis might indicate that there is a similar underlying problem.

DougBurke:
Even though `Cash`/`CStat` are designed for poisson data, there's well-known issues with bins with 0 counts in ...

e.g. https://ui.adsabs.harvard.edu/abs/2009ApJ...693..822H/abstract

which has given rise to a "you want a count or two per bin" urbal legend

hamogu:
Note that the approach I came up with here is very similar to what the optimization package optimagic does: https://optimagic.readthedocs.io/en/latest/how_to/how_to_errors_during_optimization.html#how-to-errors

If I had discovered that earlier, it may have saved me a lot of time trying to come up with this scheme!

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+
+
+class CStatNegativePenalty(CStat):
+    """CStat with negative penalty for negative model values.
+
+    CStat as defined in the literature will not work with model values
+    that are 0 or negative because of the term log(1/model).
+    In Sherpa CStat, we deal with that by truncating the model values to a
+    small positive value.
+    (At least that's the default. An error cna be raised instead if the
```
anetasie, line 580: correct typo 'can'

anetasie:
```suggestion
    (At least that's the default. An error can be raised instead if the
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    user sets the truncate option to False in the .sherpa.rc file.)
+
+    THIS IS ALREADY A CRUDGE, just to make it work numerically.
+    In practice, most model will have mostly positive values all the time,
```
anetasie:
```suggestion
    In practice, most models will have mostly positive values all the time,
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    but in numerical optimization it can happen that the optimizer tries
+    e.g. an absorption lines that is too deep and get a few bins in the
```
anetasie:
```suggestion
    e.g. an absorption line that is too deep and get a few bins in the
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    model to a negative value. That's usually not a good fit (since the data
+    is always >=0) and the optimizer would usually step the parameter such
```
anetasie:
```suggestion
    is always >=0) and the optimizer would usually set the parameter so
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    statistics do not change with how negative it is. A model value of 0
+    and -10000 will be clipped to the same number and thus the optimizer
+    has no incentive to walk back to positive values.
+
+    Again, in many models that is not a problem, e.g. in a Gaussian
+    absorption line that is several bins wide, reducing the amplitude
+    of the line will improve the fit on the edges of the line, even if the
+    center of the line is still truncated and thus the fit will walk back
+    to positive values. But in some cases, e.g. a narrow Gaussian line
+    there might be only one one bin that is negative and stays negative for
```
anetasie:
```suggestion
    there might be only one bin that is negative and stays negative for
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    just a way to make the optimizer work better in some cases.
+    As before, we expect that the optimizer will walk back to positive values
+    and thus the penalty in intermediate steps will not influence the statistic
+    value in the final fit at all.
+
+    The scale of the penalty needs to be chosen such that it actually edges
+    the optimizer back to positive values, but not so strong that it creates
+    a giant step in the statistic value which might confuse the optimizer,
+    i.e. we don't want it to be 1e-38, but also not 1e+23.
+    For Poisson data, the natural scale is "1", so we use penalt the sum of
```
anetasie:
```suggestion
    For Poisson data, the natural scale is "1", so we use the sum of
```

sherpa/stats/__init__.py
```diff
@@ -548,6 +548,112 @@ def __init__(self, name: str = 'cstat') -> None:
...
+    all negative model values as penalty.
```
anetasie:
```suggestion
    all negative model values as the penalty.
```

anetasie (commented):
I left a few typo corrections.

anetasie:
I suggest we take this PR as a new statistics in Sherpa, so as a separate statistics. 
@DougBurke do we use Python or C implementation of the CstatPenalty?

DougBurke:
We will need to do a few things

- add the class to `__all__` so that the UI can use this statistic
- add some documentation (e.g. to `list_stats`/`set_stat`) and perhaps tweak the existing documentation
- remove `CStatNumpy` as I don't think it's needed

I can look at addiing these "mechanical" changes

DougBurke:
So, I have rebased this to pick up the optimization-documentation changes in #2303 

I have then added some changes

- remove the `CStatNumpy` class
- add basic tests that show for "well-behaved" data we get the same results as `CStat` and a negative-model case to show how the penalty works compared to `CStat`
- rework the documentation for the `CStatNegativePenalty` class to better-match the `CStat` class and because I believe some of the documentation is better in the RTD pages (not yet added in this PR, but working on it)
- tweak the `CStatNegativePenalty` code to throw a sherpa error rather than `ValueError`, and to tweak the conditions for when this should happen
- ensure that `cstatnegativepenalty` is available for the ui layer (e.g. Sherpa in CIAO)
- ensure the class is included in the RTD pages

anetasie (approved):
I reviewed @DougBurke updates. They look good. Thanks!

anetasie:
I agree that seeing warning hundreds times is not helpful.

Issue #2223, opened 2025-02-10
Fails to build with GCC 15 due to redefiniton of C23 keywords


glaubitz:
With GCC 15, the compiler has switched to C23 by default meaning that `bool`, `true` and `false` are now built-in keywords.

Thus, `sherpa` now fails to build with GCC 15:

```
[   31s]   /bin/sh ../libtool  --tag=CC   --mode=compile gcc -DPACKAGE_NAME=\"region\" -DPACKAGE_TARNAME=\"region\" -DPACKAGE_VERSION=\"4.14\" -DPACKAGE_STRING=\"region\ 4.14\" -DPACKAGE_BUGREPORT=\"cxchelp@head.cfa.harvard.edu\" -DPACKAGE_URL=\"\" -DPACKAGE=\"region\" -DVERSION=\"4.14\" -DYYTEXT_POINTER=1 -DHAVE_STDIO_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_STRINGS_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_UNISTD_H=1 -DSTDC_HEADERS=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DHAVE_LIBM=1 -DHAVE_FLOAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE__BOOL=1 -DHAVE_MALLOC=1 -DHAVE_REALLOC=1 -DHAVE_SQRT=1 -DHAVE_STRCASECMP=1 -DHAVE_STRCHR=1 -DHAVE_STRPBRK=1 -I.     -g -O2 -MT regShape_utils.lo -MD -MP -MF .deps/regShape_utils.Tpo -c -o regShape_utils.lo regShape_utils.c
[   31s]   region_eval.c: In function ‘regCompareRegion’:
[   31s]   region_eval.c:83:9: error: expected identifier or ‘(’ before ‘true’
[   31s]      83 |     int true  = 1;
[   31s]         |         ^~~~
[   31s]   region_eval.c:84:9: error: expected identifier or ‘(’ before ‘false’
[   31s]      84 |     int false = 0;
[   31s]         |         ^~~~~
```

DougBurke:
oh joy ...

[...]

PR #2298, opened 2025-05-06
Disable ots

DougBurke:
This is similar to #2242 but for the region and wcssubs libraries. It is a bit easier as the code is already set up for these libraries (and the extensions built from them) to be optional.

It may allow building with gcc 15 - e.g. issue #2223 - but I haven't yet checked it out.


DougBurke:
The test failure is some unrelated issue with the documentation build.

DougBurke:
Hmm, trynig this ut with fedora 42 / gcc 15 I see

```
  error: can't copy '/home/sherpa/sherpa/build/lib/python3.13/site-packages/group.so': doesn't exist or not a regular file
  error: subprocess-exited-with-error
```

which I can understand. Argh - we don;t test this option which is why the tets diodn;t show it

PR #2302, opened 2025-05-14
update zenodo link for Sherpa 4.17.1

harlantc:
Update for the new Zenodo version badge and add Release History entry


harlantc:
@DougBurke @wmclaugh I'm not sure what the cause of the error is, nothing seems to have changed between this and the last run.

Should we merge this anyway as it's just a README update?

DougBurke:
@harlantc - for completeness, the failure here is because of

- #2223 

and can be worked around with 

- #2307

but I don't think we need to do a rebase or merge here given the changes (and the pip runs passed)/

DougBurke (approved):

Issue #783, opened 2020-04-16
A few issues with simple bounds


dtnguyen2:
While investigating how to get scipy opt methods to recognize simple bounds, I found a couple of issues with some sherpa code:
1) During the discussion at [boundarywrapper0](https://icxc.harvard.edu/pipe/sherpadev/2020/0151.html) and [boundarywrapper1](https://icxc.harvard.edu/pipe/sherpadev/2020/0152.html), it looks like the resulting fit is dependent on how some unconstrained optimization algorithm handles simple bounds.
2) The following function wrapper:
```
    orig_fcn = stat_cb0
    def stat_cb0(x_new):
        if _my_is_nan(x_new) or _outside_limits(x_new, xmin, xmax):
            return FUNC_MAX
        return orig_fcn(x_new)
```
should be changed to:
```
    def my_stat_cb0(x_new):
        if _my_is_nan(x_new) or _outside_limits(x_new, xmin, xmax):
            return FUNC_MAX
        return stat_cb0(x_new)
``` 
If work is to be done on optimization code, may as well fix the following too:
```
  template < typename Real >
  class Array1D : public std::vector< Real > {
    //
    // base class std::vector has a non-virtual destructor 
    // therefore: Thou shall not inherit from std::vector!
    //
```
Moreover, may as well bite the bullet and transform to handle simple bounds.

PR #2303, opened 2025-05-19
Adding optimizers, options, and docs

hamogu:
## Summary

Add optional dependencies to Scipy and optimagic and warp those as optimizers. Also adding docs on how to solve common fitting problems.

## Details
We have a number of open issues related to optimizers, several of which ask for new algorithms to be added or for better handling of new or existing algorithms (e.g. better handling of bounds, rescaling data etc.).

### Optimagic
In this PR, I'm adding an [Optimagic](https://optimagic.readthedocs.io/en/latest/index.html) optimizer, where optimagic itself has one main (beyond the usual numpy, pytest etc.) dependency (scipy) and a number of optional dependencies. Optimagic wraps > 70 different algorithms (some are wrapped twice in slightly different ways) from about a dozen different packages into a single interface. All those pacakges are optional dependencies, but they include some of the iminuit algorithms (requested in #150), nevergrad (requested in #580), numerical differentiation in LevMar (requested in #611). There are other benefits/options as well such as scaling of the paramter space to account for bounds (requested in #783), ways to get unstuck if a fit ventures into forbidden territory (requested in #836) and extensive methods for comparing and benchmarking algorithms.

Still, Optimagic stays an optional dependency for Sherpa.

### Scipy
Since scipy is so well knwon in the ecosystem and is a required dependency for Optimagic anyway (and for arviz, which we added as optional dependency for Sherpa in the last version), I wrapped those algorithms separately.

### Docs
There are new docs for fitting, resolving the ToDos marked in the documentation as well as new exntensive pages to help people who might be stuck in fitting.

## Notes 

- Some of those changes could be taken independently (e.g. much of the docs or scipy without optimagic), but I did not want to create 3 separate PRs which would almost certainly end up with significant merge conflicts.

- This PR is written to have as little interaction as possible with #2259, though some clean-up might be needed to bring the two together.

- If #2180 / #555 go through before merging this, changelog needs to be written.


docs/fit/index.rst
```diff
@@ -212,46 +232,56 @@ and the best-fit parameter values
...
-   Initial fit statistic = 840.251
-   Final fit statistic   = 96.8191 at function evaluation 6
-   Data points           = 6
-   Degrees of freedom    = 4
-   Probability [Q-value] = 4.67533e-20
-   Reduced statistic     = 24.2048
-   Change in statistic   = 743.432
-      polynom1d.c0   -11.0742     +/- 2.91223     
-      polynom1d.c2   0.503612     +/- 0.0311568   
+.. plot::
```
DougBurke: I always get confused here. Does this need to have the `plot` directive (along with the `:xxx:` options)? Just printing out the format output shouldn't be necessary for the plots, but maybe it makes it get included in the doctests (but in that case why is the following `print(res)` not included, which makes me think it isn't needed. But I've been wrong too many times with the doc/plot code to know anymore).

docs/fit/index.rst
```diff
@@ -263,13 +293,17 @@ The :py:class:`~sherpa.plot.DataPlot`, :py:class:`~sherpa.plot.ModelPlot`,
...
+
+    >>> from sherpa.plot import DataPlot, ModelPlot
+    >>> dplot = DataPlot()
+    >>> dplot.prepare(f.data)
+    >>> mplot = ModelPlot()
+    >>> mplot.prepare(f.data, f.model)
+    >>> dplot.plot()
+    >>> mplot.overplot()
 
 .. image:: ../_static/fit/data_model_c0_c2.png
```
DougBurke: Can we not delete this figure too? 

As mentioned above, I have paged out all the plot/doc knowledge I used to have...

docs/fit/index.rst
```diff
@@ -684,15 +744,23 @@ thawed parameter and varies it until it finds the point where the
...
+Since this can take time for
 complicated fits, the default behavior is to display a message
 when each limit is found (the
 :ref:`order these messages are displayed can change <fit_multi_core>`
 from run to run):
-   
-   >>> from sherpa.estmethods import Confidence
-   >>> f.estmethod = Confidence()
-   >>> conferrs = f.est_errors()
+
```
DougBurke, line 755: Am I right in thinking you've done this so that the test code doesn't complain because it thinks the output of `est_errors` is empty whereas we get the per-limit errors below, and this is because these limits are displayed using the logger and aren't captured by the test harness?

Or is it just that it reads better and nothing to do with the test code?

docs/fit/index.rst
```diff
@@ -765,22 +832,25 @@ of parameters by setting the ``parlist`` parameter of the
...
-   fitname     = levmar
-   statname    = chi2gehrels
-   sigma       = 1
-   percent     = 68.26894921370858
-   parnames    = ('polynom1d.c1',)
-   parvals     = (-2.4169154937357664,)
-   parmins     = (-0.25088931712955054,)
-   parmaxes    = (0.25088931712955054,)
-   nfits       = 2
+.. plot::
```
DougBurke, line 827: My assumption is that you don't actually need this `plot` directive (since it should not be relevant for the next plot, as `IntervalProjection` should not be affected by this), but I can not guarantee it. And I'm too lazy to check myself.

sherpa/optmethods/optscipy.py
```diff
@@ -87,37 +87,38 @@ def convert_bounds_to_scipy(parmins: ArrayType,
     return bounds
 
 
+# At first sight, this looks like it should be a decorator, but instead
+# we make it a higher-order function, because decorated functions cannot
+# be pickled.
+# https://pythonicthoughtssnippets.github.io/2020/08/09/PTS13-rethinking-python-decorators.html
 def wrap_scipy_fcn(func: Callable,
```
DougBurke, line 94: So, the approach I've been taking in 

- #2022

which is basically "when using non-fork multiprocessing methods everything falls apart because of pickling problems", which look essentially what

- https://pythonicthoughtssnippets.github.io/2020/08/09/PTS13-rethinking-python-decorators.html
- https://gael-varoquaux.info/programming/decoration-in-python-done-right-decorating-and-pickling.html

talk about, and you have here with the use of `functools.partial`, is to move all the locally-defined symbols like `stat_wrapper` into "global" symbols, via the `StatCallback` class (and others). I'm not sure if that would avoid the need for `functools.partial` here.

sherpa/optmethods/__init__.py
```diff
@@ -111,756 +111,9 @@
...
-           Wright, M. H. (1996) in Numerical Analysis 1995
-           (Proceedings of the 1995 Dundee Biennial Conference in
-           Numerical Analysis, D.F. Griffiths and G.A. Watson, eds.),
-           191-208, Addison Wesley Longman, Harlow, United Kingdom.
-           https://bemlar.ism.ac.jp/zhuang/Refs/Refs/wright1995numana.pdf
-
-    """
-    def __init__(self, name: str = 'simplex') -> None:
-        super().__init__(name=name, optfunc=neldermead)
+__all__ = buildin.__all__ + optscipy.__all__
```
DougBurke: I'd prefer to split this commit into two

- the first would be moving the contents of `__init__.py` to `buildin.py`
- the second would be the addition of `optscipy.py`

It would just make it easier for me to see how things go here.

sherpa/optmethods/optscipy.py
DougBurke: Actually see

- #2309 

rather than 

- #2022 

(It's the same idea but a newer version).

docs/fit/index.rst
hamogu: Yes we can. Not why I didn't (of, if I did, why it doesn't show up here. GH tell me the lines you commend on are "outdated" some maybe I did in a later commit. Either way, will check.)

docs/fit/index.rst
hamogu: I believe `res` is only defined within the thread that runs the plot stuff (because it's used elsewhere on this page) and the dockets will fail. But, as you, I'm not 100% sure, so I'll check and just they it out.

docs/fit/index.rst
DougBurke: [do not understand why this is labelled outdated; not sure I want to know, as I don't think I was looking at an old version, but maybe I was ???]

hamogu: It can happen when you comment "per commit" and the code was changed in a later commit then the one you looked at while you wrote that comment.

sherpa/optmethods/optscipy.py
hamogu: I beliebe (based on trying this and reading the link that I provided in the code and you in the comment) that it's not `functools.partial` per-se doesn't work, but that we need to be careful in what namespace the function lives that is wrapped. The code I have here passes the tests despite using `functools.partial` below with the appropriate modifications. Some I'm not quite sure what this comment wants me to do? Do you have a case that fails? Or do you not like the approach and want me to try something else?

DougBurke: I am really just trying to point out that there's some potential commonality between these two PRs. 

From a quick look it appears that the approach I took in #2022/#2039 are not necessarily needed (e.g. the introduction of `StatCallback` rather than creating local lambda functions), at least not in the code, you've changed. It makes me wonder whether we can avoid all the changes in #2039 (and some of the PRs we've already merged), but my guess is that we would still need to make some changes, and  I'm not sure I have the time or energy to find out whether that something (for #2039) is easier or harder to understand than the changes I've made [*].

I don't think we have to use exactly the same approach for both, as I'm more concerned about

- being able to understand the code
- we get the same behavior as before (or that changes are understood); and by same behavior I'm really just talking about "the tests still pass"

The links you gave are useful in helping me understand why some of the changes I added in #2309 work.

Do you think it would be worthwhile looking at #2309 and trying to use `partial` instead of the changes I made?

---

[*] as I've been working on this code for a long time I no-longer have an unbiased view.

hamogu: Partial isn't magic either - it's Python from the standard library (which means it's hopefully stable) but I don't think it does anything that couldn't be done without `functools` (i.e. it's not a magic C implementation that has powers you can't achieve in pure Python).  And some of it is style (how expressive do you want to be for the reader?) or "pick your poison" - if you can't have everything nice which part of the code do you make ugly? - so I agree that there is not necessarily a "correct" solution. I'll have a look at #2309 for inspiration and will see what I learn there. But on top of my head I don't think that unifying two different things here just for style is worth our time.

sherpa/optmethods/__init__.py
hamogu: My git-fu for splitting commits is not very good. Is that really necessary? I would probably copy all the changes into temporary directories and then copy them overt a new branch manually one by one. Chances are that I change the order of things else when I do that...

sherpa/optmethods/optscipy.py
hamogu: I like the use the `partial` that I'm using here. However, I don't think that means rewriting #2309 is a good use of our time unless we need to to avoid conflicts. So the real question is how to proceed. Do we want to put this in and then #2309 after that, or should I put this on ice until #2309 is in? While I've tried to keep them independent, #2309 and friends is rewriting some of the interfaces and I feel there is a risk that stuff falls through the cracks if we try to do both at the same time.

At the same time, I do want to get this in before the next release. I know there are some outstanding review comments, but they are all very minor and won't have any significant impact on how this code works or what the user sees in the rendered docs (about which lines of code exactly need to be inside the lot directive in the rst file).

hamogu:
Review comments are pretty minor so far ("Do we really need to enclose those few lines tin `.. plot::` in the rest files?). Does that mean that we all agree that we want this in in principle or does it just mean that nobody has looked at the bigger picture.

This is adding a significant new optional dependency. I think it resolves a number of issues and enhancements we've wanted over the years, and is worth it. If @DougBurke and @anetasie agree, it would be good to write this down here.

I don't want to polish the formatting of the docs, only to be told later that we don't want this at all.

sherpa/optmethods/__init__.py
DougBurke: I know, from experience, that long-lived PRs can be hard to merge if there's a large number of changes that come in from other work before your  work gets to be approved. I have found it particularly annoying when doing changes that have a large "structural" component (e.g. moving code into helper classes/routines) because then it becomes hard to ensure that any merge or rebase you do hasn't ended up losing the changes that have come in from other merged PRs.

I claim that splitting this commit into two, as I suggest, would mean that the "merge hell" would be limited to the first commit (that moves information around), and hopefully make it easier to then update any of your remaining PRs that need to deal with these changes.

However, it's still going to be a mess, however it's done.

DougBurke:
I don't have a problem with the overall goal. The comments I have already made were not "complete", just easy questions I have (not to say that the answers are wasy).

I will note that Dan looked at adding optional optimisers in

- #1448 
- #1465 

but he got caught up in properly handling the bounded parameter case (or that is my understanding of why this never got to the review stage). So I wonder if we are missing something in our documentation/code that says "your optimiser really needs to be able to support this feature/MacGuffin"?

Alternatively we just support what is available and look at possibly retro-fitting MacGuffin support once people hit this case.

Or just say that as no-longer have a MacGuffin expert then all bets are off.

sherpa/optmethods/__init__.py
hamogu: I'll try my best, but it's probably going to rewrite a whole bunch of things.

hamogu:
### On bounds in the the new optimizers

I'm not going to link to the line (because I will rebase shortly to address conflicts and force-push and then the commit IDs won't be available any more), but look at `convert_bounds_to_scipy`. Some of the scipy and many of the optimagic optimizers support bounds. It's not on us to deal with that (like the `InfinitePotential` we use for some build-in fitters), the scipy devs or the devs of the packages that optimagic wraps have already done that. All we need to do is to pass in the the bounds in the correct format. 

So the situation is actually better than in our own optimizers - we leverage work that other people (who in many cases will know more about optimization than I do) have already done. The default optimagic way of treating bounds is more sophisticated than our `InfinitePotential` and more similar to the approach I took in #2294; some of the optimizers are even better internally (they transform bounds). The point is that we don't have to care - other people have solved that and given us an API to use for constraints. In fact, the API allows be details than we need for Sherpa, e.g. you could specify things like `par1 + par2 > par3**2` but we don't need that in Sherpa, so this PR doesn't use it. I'm just mentioning it to point out that both [bounds](https://optimagic.readthedocs.io/en/latest/how_to/how_to_bounds.html) and [contraints](https://optimagic.readthedocs.io/en/latest/how_to/how_to_constraints.html) are handled in that API and we can just use that  without the need for us to implement anything specific in Sherpa.

That's different from Dan's PRs which used the minimization algorithm more directly and tried to treat the bounds within the Sherpa code. (And one of the reasons why I want to use an external library for this that already deal with bounds.)

The tests added in this PR check both constrained and unconstrained optimization.

docs/fit/index.rst
hamogu: Yes, this change is because the text harness doesn't know about the logger and complains if I pretend that there is output when it can't see it.

Note however, that I did not delete those lines (even though it looks like that in the GH website because your comment is set on the deleted lines). I just moved the code up by a few lines and inserted a bit of text in between ("Since this can take time for complicated fits ...") that explains what's happening and also happens to change the formatting in such a way that all tests are happy.

docs/fit/index.rst
hamogu: `c1errs` is used in one of the plotting commands below (around line 900). If this didn't have the `plot` around it, then `c1errs` would not be known in the `plot`-environment below.

sherpa/tests/test_fit_unit.py
```diff
@@ -1877,21 +1877,21 @@ def test_fit_single(stat, usestat, usesys, finalstat):
...
-    (Chi2XspecVar, False, True, fit_xvar_ft, 1e-6),
-    (Chi2ConstVar, False, True, fit_cvar_ft, 1e-6),
-    (Chi2ModVar, True, True, fit_mvar_xt, 1e-6),
-    (Chi2ModVar, True, False, fit_mvar_xf, 1e-6),
-    (Chi2ModVar, False, False, fit_mvar_xf, 1e-6),
-    (Chi2ModVar, False, True, fit_mvar_xt, 1e-6),
-
-    (Cash, True, True, fit_cash, 1e-6),
-    (CStat, True, True, fit_cstat, 2e-4),  # This tests gives different results on Linux and ARM
+    (LeastSq, False, False, fit_lsq, 2e-6),
```
DougBurke: Why should the tolerances for the NelderMead test need to be changed here? I would have naively expected the existing optimisers not to have changed code (even if it has changed location).

hamogu: This has since been pulled out of this PR and is already discussed and merged in https://github.com/sherpa/sherpa/pull/2330/files#diff-3dbf01db46cf521f9d6436699b77586c0f8c47626fe54f702aba01efb57d1673R1892

It shouldn't be here - and that's exactly the type of problem I know I would get when pulling out individual commits and splitting others. I know that git makes all that possible, I'm just not good enough to do it correctly. I'll try to clean that out here.

hamogu (commented):
For some reason, the tests that fail here pass locally. There must be something that's slightly different in the setup (such as an `install -e` gonr wrong locally or something like that). Expect a few pushes with experiments to trigger CI here as I'm trying to find out where the difference comes from.

hamogu:
Turns out I had a different scipy version installed locally, which treated a deprecated parameter different. Current code should work with either version of scipy.

docs/fit/index.rst
hamogu: checked.

DougBurke:
If I install optimagic then I get this doctest failure

```
______________ [doctest] sherpa.optmethods.optoptimagic.Optimagic ______________
[gw9] linux -- Python 3.13.5 /home/dburke/micromamba/envs/sherpa-3/bin/python3.13
298 qval           = 0.33209186143330294
299 rstat          = 1.1466590617725887
300 message        = Optimization terminated successfully.
301 nfev           = None
302 
303 The last element of the fit output is a dictionary which holds the output
304 from optimagic in an `optimagic.optimization.optimize_result.OptimizeResult` object,
305 which can provide further information about the optimization process:
306 
307 >>> result.extra_output['optimagic result'].convergence_report['five_steps']
Expected:
    {'relative_criterion_change': 0.00044802722082809587,
    'relative_params_change': 0.0025520339599334517,
    'absolute_criterion_change': 0.003082406836099949,
    'absolute_params_change': 0.041909631395145426}
Got:
    {'relative_criterion_change': np.float64(0.005707311793861156), 'relative_params_change': np.float64(0.021349474456887934), 'absolute_criterion_change': np.float64(0.03926604472083728), 'absolute_params_change': np.float64(0.7734156200382606)}

/home/dburke/sherpa/sherpa-3/sherpa/optmethods/optoptimagic.py:307: DocTestFailure
----------------------------- Captured stdout call -----------------------------
WARNING: value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING: value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING: value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING: value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING: value of parameter gauss.fwhm is below minimum; setting to minimum
------------------------------ Captured log call -------------------------------
WARNING  sherpa.models.model:model.py:943 value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING  sherpa.models.model:model.py:943 value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING  sherpa.models.model:model.py:943 value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING  sherpa.models.model:model.py:943 value of parameter gauss.fwhm is below minimum; setting to minimum
WARNING  sherpa.models.model:model.py:943 value of parameter gauss.fwhm is below minimum; setting to minimum
```

hamogu:
I'll fix locally for now, but won't push until more changes are in so we don't have to run CI again just for a bunch of doctests that (currently) don't run in CI anyway.

hamogu:
The optscipy test passes for me (on MacOS) with scipy 1.16.1. I don't think there is much we can do about that except skipping some of the output lines (so the doctest wouldn't notice they are different) or increasing the tolerance for this particular case. The parvals are similar enough that we can see that the fit covered to the correct point; it just used a different number of steps.

sherpa/optmethods/__init__.py
```diff
@@ -116,751 +116,9 @@
    const1d.c0   thawed            4 -3.40282e+38  3.40282e+38
 
 """
+from sherpa.optmethods import buildin
+from sherpa.optmethods import optscipy
+from sherpa.optmethods.buildin import *
+from sherpa.optmethods.optscipy import *
```
DougBurke: We could change this to (untested, so I'm just assuming it will work)

```python
from . import buildin, optscipy
from .buildin import *
from .optscipy import *
```

Is it worth it? Will it break things?

DougBurke: Actually, the `optscipy` lines should only be added in the next commit, not this one.

sherpa/optmethods/optscipy.py
```diff
@@ -0,0 +1,502 @@
...
+        if 'x0' in sig.parameters.keys():
+            converted_args['x0'] = x0
+        if 'bounds' in sig.parameters.keys():
+            converted_args['bounds'] = convert_bounds_to_scipy(xmin, xmax,
+                                                                requires_finite_bounds)
+
+        result = func(stat_wrapper, **converted_args, **kwargs)
+        for arg in ['bounds', 'ranges']:
+            if arg in converted_args:
+                result[f'input_{arg}'] = converted_args[arg]
```
DougBurke: From my reading of this code, `converted_args` cann only contain `x0` and `bounds` at most, so why are we also checking tor `ranges` here?

sherpa/optmethods/optscipy.py
```diff
@@ -0,0 +1,502 @@
...
+        _scipy_func = staticmethod(optimize.direct)
+        _requires_finite_bounds = True
+
+
+    __all__.append('Scipy_Minimize')
+    __all__.append('Scipy_Basinhopping')
+    __all__.append('Scipy_DifferentialEvolution')
+    __all__.append('Scipy_DualAnnealing')
+    __all__.append('Scipy_Shgo')
+    __all__.append('Scipy_Direct')
```
DougBurke, line 488: Would it not be better to do something like

```python
__all__.extend(['Scipy_Minimize', 'Scipy_Basinhoppnig', ...])
```

I'm not sure which approach I prefer for readability (I am not concerned about efficiency here).

sherpa/optmethods/buildin.py
```diff
@@ -69,7 +69,7 @@ def optfunc(fcn: StatFunc,
          x0: np.ndarray,
          xmin: np.ndarray,
          xmax: np.ndarray,
-         **kwargs) -> OptReturn:
+         **kwargs) -> OptReturn
```
DougBurke: Err, this change looks wrong to me. I assume it's just a rebase issue (since it doesn't match the commit message).

docs/fit/index.rst
```diff
@@ -485,11 +537,11 @@ function evaluation number, the statistic, and the parameter values::
...
 As can be seen, a solution is found quickly in this situation. Is it
 the same as
-:ref:`the previous attempt <fit_c0_c1_c2>`? Although the final statistic
-values are not the same, they are very close:
+:ref:`the previous attempt <fit_c0_c1_c2>`? The final statistic
+values are the same up to numerical precision:
 
    >>> res3.statval == res2.statval
-   False
+   True
```
DougBurke, line 535: This doesn't match the text or the next line, which has `res3.statval` different to `res2.statval`!

docs/fit/strategies.rst
```diff
@@ -0,0 +1,451 @@
...
+e.g. that set the initial size of steps or the maximum number of iterations.
+For `~sherpa.optmethods.buildin.MonCar` we can make the search more robust
+by making it more likely to jump in directions that do not improve the
+fit statistic, which can allow the optimizer to escape local minima.
+
+    >>> fit.method.weighting_factor = 0.5
+    >>> fit.method.xprob = 0.5
+
+
+Run fits in parallel
```
DougBurke: I think it probably best just not to add this section. We can add it if we ever get time to work on the text (and code) to add here!

sherpa/optmethods/buildin.py
```diff
@@ -180,7 +180,7 @@ def optfunc(fcn: StatFunc,
     def __init__(self,
                  name: str,
                  optfunc: OptFunc,
-                 **kwargs: Any
+                 **kwargs: Any,
```
DougBurke: Is the trailing comma here really needed? I know it's allowed by the Python syntax, and we sometimes use it for dict-like values, but it looks a bit odd to me in a function.

sherpa/optmethods/tests/test_opt_unit.py
```diff
@@ -144,3 +144,22 @@ def test_optmethod_setattr(cls):
     # which would mean changing newval
     #
     assert oldval != pytest.approx(newval)
+
```
DougBurke, line 147: I'm trying to have a consistent approach where we have two blank lines between top-level functions. I know we don't (yet?) have a style guide setting this out in stone (and I'm not suggesting adding one for this PR). I'm not sure what black/ruff/whatever use.

sherpa/optmethods/buildin.py
DougBurke: I wonder if this is just a "rebase" issue, since the comment for this commit talks about adding support for sending in arguments, but this is already allowed here. So perhaps the addition of `**kwargs` needs to be removed from the earlier commit and moved here?

DougBurke:
I've rebased this onto #2309 and there were no conflicts, which is nice. The tests also passed. Which is also nice.

sherpa/optmethods/__init__.py
hamogu: Either relative of absolute imports work. I don't feel strongly either way. Yes, the optscipy line should have been added in the next commit - it happened when the reorganized the commits as requested earlier in the review. will try to fix. (Though nobody will check out intermediate commits...)

sherpa/optmethods/optscipy.py
hamogu: Either works, I don't care strongly either way.

sherpa/optmethods/buildin.py
hamogu: It's a style issue. In general, I see people adding commas in e.g. the following signature:
```Python
def func(a: int,
                some_other_parameter : Literal["opt1", "opt2"],
               ) -> str | None
```
You do that to keep the git commit history clean - if you later add another optional parameter you don't have to change the line `some_other_parameter : Literal["opt1", "opt2"],` just to add the comma; you new commit will only have the line with the new parameter in it, so it doesn't look as if you did changes to `somer_other_parameter` in a `git blame` when really you just added the comma to allow third parameter.

But it's certainly style and either conventions works - I also admit that it's not really needed after a `**kwargs` since that's typically in the end of the function signature and nothing will ever be added behind it.

sherpa/optmethods/tests/test_opt_unit.py
hamogu: Yes, two empty Ines are recommended by PEP 8.

sherpa/optmethods/__init__.py
hamogu: Done while rewriting the commit history to fix the commit at which optscipy is imported.

sherpa/optmethods/optscipy.py
hamogu: done.

docs/fit/strategies.rst
hamogu: removed.

docs/fit/index.rst
hamogu: When I change this to `False` the doctest fails, sp I can't do that. I will however rewrite it to avoid confusion, mayby using numpy isclose.

sherpa/optmethods/buildin.py
hamogu: I depends on we really want to write here. When I say "with the following signature" I mean "hee is the list of parameter you need in this order", not "here is functional code that works". In functional code, the line starting with "def" needs ot end in ":", but then that line needs to be followed by actual code and, to meet the description, also needs to return something. That's not in this example either. 
Maybe the best solutions is (as usual) to expand this further and have an actual working example. But then, that example is already in the "Example" section, so it might be the best to just point there.

hamogu:
I'll push a rebase shortly. Just a comment for the review: Since there was a lot of focus if certain changes as done in the correct commit, I've tried to address the comments by @DougBurke by editing the commit where the actual change was made.
However, there are a few cases where that's difficult - not conceptually, just mechanically because the same file has been repeatedly edited and I create merge conflict for several of the following commits just for a stylistic change. I think there is a high risk that I mess up the rebase when I try to correct that, so  a few changes are in last commit with the comment "address review changes".

If that's not wanted, I suggest to merge more commits together, but that runs counter to @DougBurke 's earlier request to split a commit for clarity. I feel I've made a reasonable effort, but I also believe that there are better uses of our time then to move which lines of code are changes around between consecutive commits.

DougBurke:
So the doc-build now "fails" with the message

/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/2303/lib/python3.10/site-packages/sherpa/optmethods/buildin.py:docstring of sherpa.optmethods.buildin.OptMethod:23: WARNING: Inline strong start-string without end-string.

sherpa/optmethods/optoptimagic.py
```diff
@@ -0,0 +1,328 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+"""Interface to optimagic optimization methods.
+
+This module contains classes that wrap the optimization functions in
+`optimagic <https://estimagic.org>` to match the calling signature and
```
DougBurke:
```suggestion
`optimagic <https://estimagic.org>`_ to match the calling signature and 
```

Is this all that's needed for the doc build?

hamogu:
Based applying this locally, it seems that's all that's required to fix it.

sherpa/optmethods/optscipy.py
```diff
@@ -0,0 +1,498 @@
...
+        converted_args['x0'] = x0
+    if 'bounds' in sig.parameters.keys():
+        converted_args['bounds'] = convert_bounds_to_scipy(xmin, xmax,
+                                                            requires_finite_bounds)
+
+    result = func(stat_wrapper, **converted_args, **kwargs)
+    for arg in ['bounds', 'ranges']:
+        if arg in converted_args:
+            result[f'input_{arg}'] = converted_args[arg]
+    return (result.success, result.x, result.fun, result.message, result)
```
DougBurke: I commented on this before but it looks like it has disappeared during development.

`converted_args` can only contain `x0` and `bounds` keys, so why does line 118 include a check for `ranges`? Should it also be looking in `kwargs` too (as that's the only way I can see that `ranges` can be set)?

hamogu: I saw your comment. Wha't missing here is really in the line above that: Most of the scipy functions take a "bounds" argument, but https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brute.html#scipy.optimize.brute takes "ranges". There are subtle difference in the format that is accepted for those parameters, but they don't matter for how it's implemented here. So, we want to pass in either bounds or ranges, depending on how it's called in the function we are wrapping (thus the `in sig.parameters.keys()`.

hamogu: However, in later development, I decided not support the `scipy.optimize.brute` optimizer and thus I removed the test for ranges above ,and then former to also remove it in line 118. `brute` differs in the interface from the the other scipy optimization methods and it has the same functionality as the sherpa grid search that we already have so there is little benefit of adding it here.

docs/install.rst
```diff
@@ -49,6 +49,14 @@ if installed:
   has only been limited testing. Please
   `report any problems <https://github.com/sherpa/sherpa/issues/>`_
   you find.
+* `scipy <https://www.scipy.org>`_ for the `sherpa.optmethods.optscipy`
+  module, which provides an interface to use several optimizers from Scipy
+  for Sherpa fitting.
+* `optimagic <https://optimagic.readthedocs.io>`_ for the 
+  `sherpa.optmethods.optoptimagic` which provides an interface to `optimagic.minimize`.
```
DougBurke:
```suggestion
  `sherpa.optmethods.optoptimagic` module which provides an interface to `optimagic.minimize`.
```

docs/install.rst
```diff
@@ -49,6 +49,14 @@ if installed:
...
+  Optimagic provides a single interface to dozens of optimizers, many of which
+  rely on other external packages that need to be installed separately, see
+  https://optimagic.readthedocs.io/en/latest/installation.html
```
DougBurke:
```suggestion
  rely on other external packages that need to be `installed separately
  <https://optimagic.readthedocs.io/en/latest/installation.html>`_.
```

docs/install.rst
```diff
@@ -49,6 +49,14 @@ if installed:
   has only been limited testing. Please
   `report any problems <https://github.com/sherpa/sherpa/issues/>`_
   you find.
+* `scipy <https://www.scipy.org>`_ for the `sherpa.optmethods.optscipy`
+  module, which provides an interface to use several optimizers from Scipy
+  for Sherpa fitting.
```
DougBurke:
```suggestion
  module, which provides an interface to several optimizers from Scipy.
```

sherpa/optmethods/optoptimagic.py
```diff
@@ -0,0 +1,328 @@
...
+    def wrap_optimagic_minimize(
+                    stat: StatFunc,
+                    x0: np.ndarray,
+                    xmin: np.ndarray,
+                    xmax: np.ndarray,
+                    **kwargs) -> OptReturn:
+        """Wrap `optimagic.minimize` to the Sherpa interface.
+        """
+
+        bounds = None
```
DougBurke:
```suggestion
```

It's defined below so no need to also have it here.

sherpa/optmethods/optoptimagic.py
```diff
@@ -0,0 +1,328 @@
...
+        if hasattr(kwargs['algorithm'], 'algo_info'):
+            algo_info = kwargs['algorithm'].algo_info
+        else:
+            try:
+                algorithm = getattr(om.algos, kwargs['algorithm'])
+            except AttributeError:
+                raise AttributeError('optimagic does not have an algorithm '
+                                     f'named {kwargs["algorithm"]}. ')
```
DougBurke:
```suggestion
            except AttributeError as AE:
                raise AttributeError('optimagic does not have an algorithm '
                                     f'named {kwargs["algorithm"]}. ') from AE
```

It just makes the chain of exceptions a bit easier to read when they get thrown.

sherpa/optmethods/optoptimagic.py
```diff
@@ -0,0 +1,328 @@
...
+
+        result = om.minimize(stat_wrapper, 
+                             x0,
+                             bounds=bounds,
+                             **kwargs)
+        
+        out = (# I don't think we'll ever get None as a return value, but Mypy complains
+               # that the type declaration in optimagic is "bool | None", while we want "bool".
+               # None certainly isn't a valid success.
+               result.success if result.success else False,
```
DougBurke: How about

```suggestion
               bool(result.success), 
```

or is this too "cute"?

DougBurke:
@anetasie - can you review the documentation changes for this one?

docs/fit/simultaneousfits.rst
```diff
@@ -0,0 +1,146 @@
...
+.. plot::
+    :include-source:
+    :context:
+    :nofigs:
+
+    >>> import numpy as np
+    >>> from sherpa.models import Gauss1D
+    >>> wave_order26 = np.arange(6686, 6720, 1.1)
+    >>> wave_order27 = np.arange(6705, 6735, 1.2)
+    >>> rng = np.random.default_rng()
```
DougBurke: I suggest using an explicit seed here, to make sure the plots are repeatable.

docs/fit/simultaneousfits.rst
```diff
@@ -0,0 +1,146 @@
...
+.. plot::
+    :include-source:
+    :context:
+
+    >>> from matplotlib import pyplot as plt
+    >>> _ = plt.plot(wave_order26, flux_order26, 'o', label='order 26')
+    >>> _ = plt.plot(wave_order27, flux_order27, 'o', label='order 27')
+    >>> _ = plt.xlabel('Wavelength (Angstrom)')
+    >>> _ = plt.ylabel('Flux (normalized)')
+    >>> _ = plt.legend()
```
DougBurke, line 66: The plot I see is

https://sherpa--2303.org.readthedocs.build/en/2303/_downloads/db8048269eeba0a281981c3522483a2f/simultaneousfits-2.png

image


which isn't informative to me. For some reason it's included data with `x ~ 0` (rather than the `x ~ 6700` values we expect).

docs/fit/poisson.rst
```diff
@@ -0,0 +1,118 @@
...
+While very few distributions in science are truly normally distributed, in practice
+they are often close enough that a :math:`\chi^2` statistic can be used and plots can
+display symmetric error bars. Neither is true for Poisson distributed data, in particular
+when there are fewer than 25 counts or so in any bin.
+
+Sherpa implements `sherpa.stats.Chi2Gehrels` which applies a
+weighting scheme to allow a :math:`\chi^2`-like statistic to be used (details and
+references are given in the documentation for that statistic).
+In general, though, Poisson distributed data should be fit by minimizing the
+likelihood. Two options are available in Sherpa:
```
DougBurke: Actually, there are three, as we also have `sherpa.stats.WStat`

docs/fit/poisson.rst
```diff
@@ -0,0 +1,118 @@
...
+in each bin, where the X-axis is the logarithm of the count rate. There are
+even bins with zero counts.
+Thus, we pick `~sherpa.stats.CStat` as the statistic and use the default
+`~sherpa.optmethods.LevMar` optimizer which usually works well for count data.
+
+.. plot::
+    :include-source:
+    :context:
+
+    >>> from sherpa.plot import DataPlot, ModelPlot
```
DougBurke: We could use `DataHistogramPlot` rather than `DataPlot` here, as it shows the histogram edges.

Note that as you have a `Data1DInt` dataset in the fit then the fit will sum up the gauss model across each bin. This may be surprising. You can compare the results to a fit where you first call `xraydistribution.integrate = False`.

This normally doesn't change the "shape" parameters (here `fwhm` and `pos`) but will change the normalization.

```
>>> hist = np.array([45, 50, 43, 30, 27, 32, 5, 8, 0, 4, 1, 0])
>>> edges = np.array([-6.5, -6.3, -6.1, -5.9, -5.7, -5.5, -5.3, -5.1, -4.9, -4.7, -4.5, -4.3, -4.1])
>>> from sherpa.data import Data1DInt
>>> data = Data1DInt("x", edges[:-1], edges[1:], hist)
>>> from sherpa.models import Gauss1D
>>> mdl1 = Gauss1D("m1")
>>> mdl2 = Gauss1D("m2")
>>> mdl2.integrate = False
>>> from sherpa.fit import Fit
>>> from sherpa.stats import CStat
>>> f1 = Fit(data, mdl1, stat=CStat())
>>> f2 = Fit(data, mdl2, stat=CStat())
>>> f1.guess()
>>> f2.guess()
>>> print(mdl1)
m1
   m1.fwhm      thawed            2        0.002         2000           
   m1.pos       thawed         -6.3         -6.5         -4.1           
   m1.ampl      thawed          250         0.25       250000           
>>> print(mdl2)
m2
   m2.fwhm      thawed            2        0.002         2000           
   m2.pos       thawed         -6.3         -6.5         -4.1           
   m2.ampl      thawed          250         0.25       250000           
>>> r1 = f1.fit()
>>> r2 = f2.fit()
>>> print(mdl1)
m1
   m1.fwhm      thawed       1.5021        0.002         2000           
   m1.pos       thawed     -6.25457         -6.5         -4.1           
   m1.ampl      thawed      235.942         0.25       250000           
>>> print(mdl2)
m2
   m2.fwhm      thawed      1.50823        0.002         2000           
   m2.pos       thawed     -6.35456         -6.5         -4.1           
   m2.ampl      thawed      46.9959         0.25       250000           
```

docs/fit/strategies.rst
```diff
@@ -0,0 +1,440 @@
...
+(or at least a subset that we can display in 2D or 3D):
+
+.. plot::
+    :include-source:
+    :context: close-figs
+
+    >>> _ = plt.scatter(fit_result2.record_steps['Halpha.pos'],
+    ...                 fit_result2.record_steps['Halpha.ampl'])
+
+Uups! We forgot to :ref:`fit-strategies-setminmax` and MonCar is a global
```
DougBurke: Not sure whay you meant with `Uups`, so this is a best guess

```suggestion
Woops! We forgot to :ref:`fit-strategies-setminmax` and MonCar is a global
```

sherpa/optmethods/optoptimagic.py
```diff
@@ -0,0 +1,327 @@
...
+
+    >>> opt.algo_options = {'stopping.maxfun': 1000}
+
+Most inputs for optimagic are automatically filled by Sherpa
+------------------------------------------------------------
+This module wraps the `optimagic.minimize` function in such a way that
+all the information that Sherpa already has about the model, the data, and the
+statistic is passed to the optimizer. Sherpa will automatically pick the correct
+format of the statistic function, convert the parameter minimum and maximum
+to the format for optimagic bounds etc.
```
DougBurke:
```suggestion
format of the statistic function and convert the parameter minimum and maximum
to the format for optimagic bounds.
```

It's not clear to me what `etc` means here, so just remove it. Unless there's something explicit that can be used instead?

docs/fit/poisson.rst
hamogu: Despite f1 and f2 having different normalizations, I get the exact same plot when I plot the two with `DataHistrogramPlot` and `ModelPlot` and neither of them has histogram edges:
Screenshot 2025-09-10 at 3 35 52 PM

I don't understand why, but I'll check.

hamogu:
addressed (I hope) comments and rebased to address merge conflict

docs/fit/poisson.rst
DougBurke: Aha.

a) the default for the data histogram plot is to just draw a point. You can add `linestyle='solid'` to draw the edges

b) the model plot needs to use `ModelHistogramPlot` rather than `ModelPlot` to show the edges

With that I get something like 

```python
dplot.plot()
mplot.overplot()
```

image

and

```
dplot.plot(linestyle='solid')
mplot.overplot()
```

image

sherpa/optmethods/optscipy.py
```diff
@@ -243,8 +243,8 @@ class Scipy_Minimize(ScipyBase):
         statname       = chi2
         succeeded      = True
         parnames       = ('const.c0', 'gauss.fwhm', 'gauss.pos', 'gauss.ampl')
-        parvals        = (26.06715425374017, 3.285588554348949, 5.0425169187544565, 42.00624527682233)
-        statval        = 6.879954385987281
+        ...
+        message        = CONVERGENCE: RELATIVE REDUCTION OF F <= FACTR*EPSMCH
```
DougBurke: Looks like this message is also version dependent. I suggest just not testing it.

So delete lines 247 and 248 I guess

PR #2305, opened 2025-05-20
Update `pyproject.toml` for PEP 639

hamogu:
## Summary
Update license specification in `pyproject.toml`  for PEP 639

## Details
PEP 639 defines how licenses should be defined, this PR implements that. It requires setuptools > 77.0.3 to work, which was released on March 20, 2025 (so relatively recent).
On the other hand, setuptool plans to stop supporting the old format on Feb-18-2026 (not too far in the future), so it would be good to make the switch before the Sherpa release that goes into CIAO in Dec 2025.

With a current setuptools, the Sherpa install issues warnings like this:
```
 /tmp/pip-build-env-3q60enjn/overlay/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
  !!

          ********************************************************************************
          Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

          By 2026-Feb-18, you need to update your project and remove deprecated calls
          or your builds will no longer be supported.

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
    corresp(dist, value, root_dir)
  /tmp/pip-build-env-3q60enjn/overlay/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!

          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
```

see, e.g. https://github.com/sherpa/sherpa/actions/runs/15123682552/job/42579807771?pr=2303  (but CI logs are only preserved for a few weeks, so this link won't be forever).


DougBurke:
I think we decided when you last tried this that there was no sensible way to use the new `setup.cfg` file with an "old" `setuptools` version (which is why we need to bump the minimum version)?

DougBurke:
I'm happy with this, but I'd rather we don't take it immediately because

- setuptools 77.0.3 is, as you said, relatively new so I'd prefer to wait a little while
- hopefully we are going to get rid of setuptools

Unless you think we should take it now?

hamogu:
Is the 4.17.2 release independent of CIAO or will the 4.17.2 become the OTS for CIAO 4.18? I think we should add it before the release that becomes CIAO 4.18 in December.

DougBurke:
The current approach has been that we don't have a `.2` release, just the `.0` and `.1` releases - so I expect our next release to be `4.18.0` - e.g. https://github.com/sherpa/sherpa?tab=readme-ov-file#release-history - assuming we still have jobs by then...

hamogu:
In that case, I suggest to do this change before the next feature freeze - and that means it can go into the development branch now.
(It's not that setuptools is a difficult dependency. It's installed automatically by pip if needed and it's unlikely that users even notices this change.)

DougBurke (approved):

PR #2308, opened 2025-05-24
Remove pyds9

hamogu:
## Summary
Remove pyds9 module. 

## Details
The pyds9 project is no longer maintained, and the pyds9 module in sherpa is (a) not used and (b) broken (i.e. it can't be imported because is misses imports). Thus, remove it.

Since the module was broken and could not be used, there is no risk of beaking user code - no user code could have used this anyway.


DougBurke (approved):
I have never understood how the pyds9 stuff was meant to work, but just never had the time to go looking. Ta for looking.

PR #2309, opened 2025-06-05
Allow spawn

DougBurke:
TBD

### DIFF
--- a/sherpa/estmethods/__init__.py
+++ b/sherpa/estmethods/__init__.py
@@ -28,6 +28,7 @@ import warnings
 import numpy as np
 from numpy.linalg import LinAlgError
 
+from sherpa.stats import StatCallback
 from sherpa.utils import NoNewAttributesAfterInit, \
     FuncCounter, OutOfBoundErr, Knuth_close, \
     print_fields, is_iterable, list_to_open_interval, quad_coef, \
@@ -219,8 +220,7 @@ class Covariance(EstMethod):
         if statargs is not None or statkwargs is not None:
             warning("statargs/kwargs set but values unused")
 
-        def stat_cb(pars):
-            return statfunc(pars)[0]
+        stat_cb = StatCallback(statfunc)
 
         def fit_cb(scb, pars, parmins, parmaxes, i):
             # parameter i is a no-op usually
@@ -305,8 +305,7 @@ class Confidence(EstMethod):
         if statargs is not None or statkwargs is not None:
             warning("statargs/kwargs set but values unused")
 
-        def stat_cb(pars):
-            return statfunc(pars)[0]
+        stat_cb = StatCallback(statfunc)
 
         def fit_cb(pars, parmins, parmaxes, i):
             # freeze model parameter i
@@ -415,8 +414,7 @@ class Projection(EstMethod):
         if fitfunc is None:
             raise TypeError("fitfunc should not be none")
 
-        def stat_cb(pars):
-            return statfunc(pars)[0]
+        stat_cb = StatCallback(statfunc)
 
         def fit_cb(pars, parmins, parmaxes, i):
             # freeze model parameter i
--- a/sherpa/estmethods/__init__.py
+++ b/sherpa/estmethods/__init__.py
@@ -564,6 +564,85 @@ def covariance(pars: np.ndarray,
             np.array(error_flags), 0, inv_info)
 
 
+# Use a class purely to make it easy to separate the setup arguments
+# from those needed to call the per-parameter routine while still
+# allowing multiprocessing to work with non-fork methods.
+#
+# As it's not clear what all the types of the values should be
+# this is not written as a dataclass.
+#
+class ProjFunc:
+    """Evaluate an individual parameter with the projection routine."""
+
+    def __init__(self,
+                 *,
+                 stat_cb,
+                 fit_cb,
+                 pars,
+                 parmins,
+                 parmaxes,
+                 parhardmins,
+                 parhardmaxes,
+                 sigma: float,
+                 eps: float,
+                 tol: float,
+                 maxiters: int,
+                 remin,
+                 report_progress
+                 ) -> None:
+        self.stat_cb = stat_cb
+        self.fit_cb = fit_cb
+        self.pars = pars
+        self.parmins = parmins
+        self.parmaxes = parmaxes
+        self.parhardmins = parhardmins
+        self.parhardmaxes = parhardmaxes
+        self.sigma = sigma
+        self.eps = eps
+        self.tol = tol
+        self.maxiters = maxiters
+        self.remin = remin
+        self.report_progress = report_progress
+
+    # Matches LocalEstFunc
+    #
+    def __call__(self,
+                 counter: int,  # unused
+                 singleparnum: int,
+                 lock: SupportsLock | None = None
+                 ) -> tuple[SupportsFloat, SupportsFloat, int, int, None]:
+        """Evaluate the projection for a single parameter."""
+
+        proj_func = _est_funcs.projection
+
+        try:
+            singlebounds = proj_func(self.pars, self.parmins,
+                                     self.parmaxes, self.parhardmins,
+                                     self.parhardmaxes, self.sigma,
+                                     self.eps, self.tol,
+                                     self.maxiters, self.remin,
+                                     [singleparnum], self.stat_cb,
+                                     self.fit_cb)
+
+        except EstNewMin as emin:
+            # catch the EstNewMin exception and attach the modified
+            # parameter values to the exception obj.  These modified
+            # parvals determine the new lower statistic.
+            raise EstNewMin(self.pars) from emin
+
+        if lock is not None:
+            lock.acquire()
+
+        self.report_progress(singleparnum, singlebounds[0],
+                             singlebounds[1])
+
+        if lock is not None:
+            lock.release()
+
+        return (singlebounds[0][0], singlebounds[1][0],
+                singlebounds[2][0], singlebounds[3], None)
+
+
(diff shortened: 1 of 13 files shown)


.github/workflows/ci-pip-workflow.yml
```diff
@@ -103,20 +104,41 @@ jobs:
...
+    # always create it.
+    #
+    - name: setup sherpa rc file
+      env:
+        NCORES: ${{ matrix.ncores }}
       run: |
-        pip install pytest-xvfb
+        cd $HOME
+        orig=`python -c 'import sherpa; print(sherpa.get_config())'`
+        if [ "${NCORES}" == "1" ]; then
```
hamogu: Since NCORES==1 is kind of a special case, should there also (or instead) be a test with ncores=2? Or can we not do that on CI / is is not worth the extra time?

sherpa/optmethods/optfcts.py
```diff
@@ -473,46 +580,19 @@ def grid_search(fcn: StatFunc,
...
 
     if sequence is None:
-        ranges = []
-        for index in range(npar):
-            ranges.append([xmin[index], xmax[index]])
-        sequence = make_sequence(ranges, num)
-    else:
-        if not np.iterable(sequence):
-            raise TypeError("sequence option must be iterable")
+        sequence = make_range_sequence(npar, xmin, xmax, num)
```
hamogu:
```suggestion
        ranges = np.linspace(xmin, xmax, num)
        sequence = np.array(np.meshgrid(*ranges.T)).T.reshape(-1, len(xmin))
```

sherpa/optmethods/optfcts.py
```diff
@@ -408,6 +414,107 @@ def difevo_nm(fcn: StatFunc,
...
+    grid = list(map(np.ravel, mgrid))
+    sequence = []
+    for index in range(mynfev):
+        tmp = []
+        for xx in range(npar):
+            tmp.append(grid[xx][index])
+        sequence.append(tmp)
+
+    return sequence
+
```
hamogu:
```suggestion
```

I know this function was not added in this PR, it's just moved around to make it accessible in a separate process. But, today, this all boils down to 1-2 line of numpy code which I inlined in a comment below, so this entire function is not needed at all.

sherpa/sim/tests/test_sim.py
```diff
@@ -427,12 +406,14 @@ def test_lrt_multicore(setup):
...
-    # The result depends on whether multi-processing is enabled and
-    # if there's more than one core available, which we can check with
-    # the _ncpus setting (although it appears that we still need to
-    # check whether multi-processing is enabled or not).
+    # The result used to depend on whether multi-processing is enabled
+    # and if there's more than one core available, which we can check
+    # with the ncpus setting (although it appears that we still need
+    # to check whether multi-processing is enabled or not). It
+    # no-longer appears to be the case, but the test is left in (but
+    # commented-out) for review.
```
hamogu: This test not commented out. I'm not sure if the "commented out" part of the comment refers to the `test_lrt_multicore` which has been removed and merged with `test_lrt` through the use of `pytest.parameterize`.  Maybe this comment was written in an earlier commit but is no longer valid now?
(I'm reviewing the final stage, not per-commit.)

sherpa/fit.py
```diff
@@ -1087,6 +1087,144 @@ def _check_length(dep) -> int:
     raise FitErr('nobins')
 
 
+# Local classes used by the error-estimation code.
+#
+# FreezePar needs to know about all the thawed parameters so it can
+# return the values of all-but-the-selected parameter, but ThawPar and
+# ParName could be sent the parameter object.
```
hamogu: Mmm. I'm sure this si clear to someone who wrote this code, but not to me. I'll try to make a suggestion once I;ve worked through the other changes and I understand what "but ThawPar and ParName could be sent the parameter object." means and why that motivates the existence of `FreezePar`.

sherpa/fit.py
```diff
@@ -1087,6 +1087,144 @@ def _check_length(dep) -> int:
...
+        # and the hard mins amd maxs of the parameters
+        self.thawedpars[idx].val = pars[idx]
+        self.thawedpars[idx].frozen = True
+        self.parent.current_frozen = idx
+
+        # Identify those parameters that are not frozen.
+        keep_pars = np.ones_like(pars)
+        keep_pars[idx] = 0
+        keep_idx = np.where(keep_pars)
+
```
hamogu:
```suggestion
        keep_idx = ~np.isin(np.arange(len(pars)), idx)

```

sherpa/estmethods/__init__.py
```diff
@@ -251,6 +254,50 @@ def fit_cb(scb, pars, parmins, parmaxes, i):
                           report_progress=report_progress)
 
 
+# This is not expected to be used outside this module.
+# It is used by both Confidence and Projection.
+#
+@dataclass
+class EstFit:
```
hamogu: Do we want to change the name to `_EstFit` to make it clear that it's not public API and we can change it at any time?
I don't think we want to advertize this for use outside of this module at this stage and I'm worried that we'll need to tweek it as the multi-processing code becomes more developed and we'll look into free-threading more in Python 3.14 or later.

(Same comment applies to all new classes in this file.)

hamogu (changes requested):
In general, I think we should take this because of the code clean-up that's done in here, not because of the great performace gain (because, in fact, it's not a performace gain to use other multi-processing methods) and also to future-proof. Free-threading is coming in Python, if we like it or not, and while I haven't tested this code on a free-threaded built, my understanding is that the chamges made here will make it easier to deal with that when we get to it.

In addition to the in-line suggestions, I have two major points I want to raise though:

- In the spirit of this PR, I have suggested even more code clean-up (but easy ones) reducing the number of methods that need to be moved from lambda functions to module-level functions or classes in `estmethods/__init__.py`. For the remaining new classes and methods, I suggest starting names with `_` to make it clear that they are internal and might be changed. I just don't want to be locked into the new API at this point.
- I suggest adding a warning somewhere (maybe when importing sherpa) if multi-processing is not set to "fork". Before this PR, it would not work, so users wouldn't use it. Now, they might do that and might never know that setting "spawn" earns them a performance regression of a factor of a few. @dougburke 's timings above are downright alarming! I'm suggesting a text alolng the lines of `Sherpa is running with multiprocessing set to method {name}. Performance will vary depending on software version, OS, data, and fitting methods, but Sherpa's code has been observed to take several times longer with e.g. method=spawn vs method=fork in some cases. We advice to evaluate performance for your specific application to decide on the most appropriate multiprocessing approach.` OK, maybe that's too wordy. It could also be "Sherpa is running with multiprocessing set to spawn method {name}, see URL_to_docs for a discussion of possible performace issues related to that choice." I think this can be so severe that we can't rely on users simply checking it out on readthedocs themselves, in particualr users who run Sherpa as part of CIAO.

sherpa/fit.py
DougBurke: So, the interface here has a list of parameters and an index into them, to indicate the parameter being looked at. We **could** change this interface (e.g. just pass the desired parameter around), **but** such a change is not needed to get this code to run and a bit larger than I'd like to run with here (plus there may be a reason why we do not want to do it this way).

For example, before this PR we had routines like

```python
        # confidence needs to know which parameter it is working on.
        def get_par_name(idx):
            return thawedpars[idx].fullname
```

This **could** have just been

```python
        # confidence needs to know which parameter it is working on.
        def get_par_name(par):
            return par.fullname
```

.github/workflows/ci-pip-workflow.yml
DougBurke: So, I see this as "run with one core or run with many cores (where many may be 2)". I am not sure it's worth adding an explicit test for ncores=n (when n is not 1). 

**edited to add** this is because the assumption is that nowadays all the CI runs are run on a multi-processor system, and I am not sure it's worth a run where we explicitly restrict ncores to a particular value that is not 1.

hamogu: OK, that's fine. If we discover problems later we can always revisit, but for now, let's assume that this is enough testing.

sherpa/fit.py
hamogu: I wasn't suggesting to change the API, just to change the text of this comment in the code.
How about:
```
# FreezePar needs to know about all the thawed parameters so it can
# return the values of all-but-the-selected parameter, but ThawPar and
# ParName only need one single parameter.
# We could design ThawPar and ParName to accept the parameter object directly,
# but the current interface uses indices consistently for all three classes so that they have 
# a similar API.
```

sherpa/fit.py
DougBurke: It looks like 

```python
keep_idx = np.isin(np.arange(len(pars)), idx, invert=True)
```

should be faster (but I doubt it will be noticeable).

hamogu: Either is fine. This is certainly not speed critical, it's about readability.

sherpa/sim/tests/test_sim.py
DougBurke: This comment refers to the

```python
    if numcores == 2:  # and multi and ncpus > 1:
```

line, and is talking about the commented out part of the check (i.e. `and multi and ncpus > 1`).

We could just remove pretty much everything from this comment as I'm not sure it really buys us anything.

DougBurke:
Rebased as #2022 and #2326 have been merged.

sherpa/sim/tests/test_sim.py
DougBurke: So I've removed the comment now.

hamogu: Fine with me.

PR #2311, opened 2025-06-11
ptcat-3768 - removing max enforcement in sherpa for CAT processing

jmiller515:
removing max enforcement in sherpa as it causes aperture photometry to fail. Updated expected tests failures due to change

### DIFF
--- a/sherpa/models/parameter.py
+++ b/sherpa/models/parameter.py
@@ -368,25 +368,14 @@ hard_min
 
     # 'val' property
     #
-    # Note that _get_val has to check the parameter value when it
-    # is a link, to ensure that it isn't outside the parameter's
-    # min/max range. See issue #742.
-    #
     _val: SherpaFloat  # needed for typing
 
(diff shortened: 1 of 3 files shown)


harlantc:
@jmiller515 I created a new branch for this, `cat58_sherpa4171`, can you update the target branch for this? I was going to myself but it looks like it might be pulling in additional commits from main, and I wasn't sure what you wanted in this.

jmiller515:
@hamogu I agree this is not the correct solution. It's a temporary stopgap to allow the catalog to work. I posted a message to sherpadev back in November about making the same temporary stopgap on sherpa 4.17 (see https://groups.google.com/a/cfa.harvard.edu/g/sherpadev/c/g8IX3WoO5tk/m/MannTZaeBwAJ). I'm happy to discuss alternative approaches, but I would need some science help to ensure that we're not breaking aperture photometry. When we originally discussed this Frank (the lead scientist) didn't have a good solution for how to handle it appropriately in aperture photometry.

PR #2313, opened 2025-06-11
Mark more session tests as optional

DougBurke:
# Summary

Mark more tests that check the Session class as being optional (run with --runsession) as in most cases there is no need to test both the sherpa.ui and sherpa.astro.ui cases. There are no functional changes in this request.

# Details

In #2271 I added the `--runsession` flag for tests to mark tests that are good to run occasionally but do not need to be run on every PR. This is used for tests which test both the sherpa.ui and sherpa.astro.ui `Session` classes.

There are a number of places where it can be added, which can reduce the runtime of the tests slightly. So this PR attempts to identify those places where requiring `--runsession` would be useful. There are some places where we either know there's potentially different paths being run on the two cases, or that the code is fast enough it isn't really worth opting out.

I want to check the codecov report just to check this doesn't reduce our test coverage significantly. **EDITED TO ADD** there are two lines that may have lost coverage, but they are not of a huge concern (since they are part of a common idiom in the UI layer and are only for sherpa.ui not sherpa.astro.ui so I think it's an okay change).

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -3027,7 +3027,7 @@ def test_pha1_get_foo_flux_hist_model(getfunc, ratio,
 
 
 @pytest.mark.parametrize("cls",
-                         [sherpa.ui.utils.Session, sherpa.astro.ui.utils.Session])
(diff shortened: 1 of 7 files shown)


hamogu (commented):
This always marks the `BaseSession` as needing an extra flag. I agree with (and I suggested originally) that always testing both is a waste of time, but I wonder if we should alternate at least a few of the tests to mark the AstroSession as the one that only runs on the special flag?

PR #2316, opened 2025-06-20
TESTS: add a test used by the Chandra Source Catalog

DougBurke:
# Summary

Add a test from the CSC code to exercise #2201. There are no functional changes.

# Details

It is not-at-all clear what this test is exercising, but it appears to be involved with making sure that the minimum value for one of the parameters is never negative, but why this is and what it is for is not explained.

This test was taken from #2186 and has seen minimal changes

- rework to use "modern" Python code (e.g. we can use @ to do the matrix multiplication)
- rework to take advantage of changes made to Sherpa (explicit passing around of the random state)
- export symbols from the module that actually defines them rather than from a module that re-exports the symbol (this is purely a style change)
- avoid importing "private" variables from a module when there's a "global" variable (this is not related to this test, just something I noted when looking at the test code)

If I revert #2201 then the test fails, so this suggests it is a test of that change.

### DIFF
--- a/sherpa/sim/tests/test_sim.py
+++ b/sherpa/sim/tests/test_sim.py
@@ -26,10 +26,12 @@ import numpy as np
 import pytest
 
 from sherpa.data import Data1D
-from sherpa.models import Gauss1D, PowLaw1D
+from sherpa.models.basic import Gauss1D, PowLaw1D
+from sherpa.models.model import ArithmeticModel
+from sherpa.models.parameter import Parameter
 from sherpa.fit import Fit
 from sherpa.stats import Cash, Chi2DataVar, CStat
-from sherpa.optmethods import NelderMead, LevMar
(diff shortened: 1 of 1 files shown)


DougBurke:
@jmiller515 - I am trying to pick out the parts of #2186 and apply them to Sherpa. There is a test that is in the PR which appears completely unrelated to any of the other changes, but it's hard to tell because of the complete lack of documentation. It is **not** related to the `est_method_arg` change for which the PR is labelled, but it could be perhaps related to other changes in the PR - such as the already-landed #2201 code.

Is there any documentation on what this test is for / why it's needed?

DougBurke:
Aha. I've tried reverting 

- #2201

and this test now fails. So it is a test for this.

hamogu (approved):
I hope we find out more about what the actual problems are the catalog needed to save so that we can fix them for good, but having the test it one step in that direction, so let's get it in.

Issue #2291, opened 2025-04-24
review module initilization support for free-threading/subinterpreter work


DougBurke:
So, it's likely a significant amount of work to properly handle free-threading and subinterpreters, because a few of our modules have implicit (e.g. maybe xspec) and explicit (e.g. psf) state handling.

However, we can start looiking at https://discuss.python.org/t/proposal-officially-deprecate-support-for-legacy-single-phase-init-extension-modules/89262

Issue #2062, opened 2024-06-14
What to do with NoGIL and Python 3.13


DougBurke:
In the unlikely event any of us are still here when 3.13 is released

https://github.com/Quansight-Labs/free-threaded-compatibility

hamogu:
In the even more unlikely event that we're still here when 3.13. is the minimum version we support....

I think this is one of those cases where we might just want to force the non-threaded version (which is still the default in 3.13 and might be in 3.14?) and deal with it once 3.15 is out and a significant fraction of our users actually use unlicked versions ...

However, that might be good incentive move some of our C code that is in C for historical reasons and not for performance to Python/Numpy (e.g. many of our statistics functions could be one-liners in numpy instead of C implementations - I doubt it changes the runtime significantly, but it's certainly not in the critical path anyway) because then we don't have to do this work - Quansight and other numpy developers will do it for us.

DougBurke:
In my work in 

- #2022 

to try and address 

- #2007 

I have noted that it looks like the non-spawn methods are significantly slower. It'll be interesting (in this unlikely future) to know how that is related to the GIL/NoGIL stuff...

PR #2319, opened 2025-06-25
Allow Sherpa to be built with lib64 or free-threading

DougBurke:
# Summary

Allow Sherpa to be built with the free-threading build or on a platform that uses the lib64 rather than lib as a directory name (e.g. Fedora 42). There is no promise at this time that the resulting code is correct.

# Details

These were found when trying to build Sherpa with Fedora 42 and with the python 3.13 free-threading conda-forge version. Ideally we could just ask sys/sysconfig/somewhere for the correct name to use, but it's not clear from the documentation what that would be, and I still aim for all this code to be swept away with the meson broom in the future, but this at least allows the build to complete.

This does not guarantee that the build works correctly (I would hope it does on Fedora 42 but there has been no attempt yet to support the free-threaded build, as this requires a lot of work; e.g.

- #2062 
- #2291 
)

### DIFF
--- a/helpers/sherpa_config.py
+++ b/helpers/sherpa_config.py
@@ -79,7 +79,6 @@ class sherpa_config(Command):
     def finalize_options(self):
         incdir = os.path.join(self.install_dir, 'include')
         libdir = os.path.join(self.install_dir, 'lib')
-        pydir = os.path.join(libdir, f'python{version}', 'site-packages')
 
         if self.fftw_include_dirs is None:
             self.fftw_include_dirs = incdir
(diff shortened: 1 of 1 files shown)


hamogu (approved):
This passes tests, so apprantly it does no harm. I can't say how mch it helps (or we we'll ever get a free-trhreaded sherpa), but using more of `sys.platlibdir` etc. ins tead of hardcoding our best guess is what those moduls are for and will help us with other general compatibility (e.g. if we ever get to a Sherpa version that runs on windows).

PR #2320, opened 2025-06-25
Start free threading

DougBurke:
This is **very** early days.

Uses

- #2319 
- #2321 

Note that Sherpa is designed to be very thread-unfriendly, so it's not clear how useful the free-threaded build will be. For instance, something like

```python
data = Data1D('x', x, y)
model = Polynom1D('mdl')
model.c1.thaw()
f = Fit(data, model)
```

is tricky to use in a free-threaded world because `data` and `model` are mutable and so changing them can change the result of `f.fit()` - and vice-versa.


DougBurke:
Rebased to pickup changes and address conflicts.

DougBurke:
Close by #2484 - see #2491 for the closest approach using meson. I won't delete this for now as there are some changes in here that could be useful.

PR #2321, opened 2025-06-26
Minor simplification of the extension code

DougBurke:
# Summary

Minor clean up of the extension code. There should be no functional changes.

# Details

Developed as part of #2320 but is a separate clean-up item.

There are two changes

1. A minor simplification in the `SHERPAMOD` and `SHERPAMODDOC` templates as they can both be written calling a new template, to avoid dupllication (the two macros only differ on whether the documentation for the module is given, and the new code highlights this).
2. Avoid declaring a symbol - normally `init_<modname>` - which we don't actually provide or use in the extension modules.

### DIFF
--- a/sherpa/include/sherpa/extension.hh
+++ b/sherpa/include/sherpa/extension.hh
@@ -1,5 +1,5 @@
 //
-//  Copyright (C) 2007, 2016, 2017, 2020, 2022, 2023
+//  Copyright (C) 2007, 2016-2017, 2020, 2022-2023, 2025
 //  Smithsonian Astrophysical Observatory
 //
 //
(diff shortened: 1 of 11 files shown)


DougBurke:
@hamogu - are you able to review this or would you like someone else to?

sherpa/include/sherpa/extension.hh
```diff
@@ -39,11 +39,11 @@ typedef int (*converter)( PyObject*, void* );
 
 #define CONVERTME(arg) ((converter) sherpa::convert_to_contig_array<arg>)
 
-#define SHERPAMOD(name, fctlist) \
+#define _SHERPAMOD(name, fctlist, doc)	   \
```
hamogu, line 49: Maybe a more descriptive name than just the same name with an underscore?

hamogu (approved):

PR #2328, opened 2025-07-22
Allow plot_fit_xxx to be sent multiple values for an argument to match plot

DougBurke:
# Summary

Allow the plot_fit_xxx and plot_bkg_fit_xxx calls to behave the same as plot("fit", xxx) and plot_model/source_components when sent a plot option as a sequence.

# Details

The `plot`, `plot_source_component`, and `plot_model_component` commands have been able to be sent multiple values for a lpot attribute - such as line color - since ~ 4.17.0 (e.g. PR #2125). So add the same capability to `plot_fit_xxx` and `plot_bkg_fit_xxx`. It is not actually that useful, since in general we don't want different values in the two plots, but it fits in with the other parts of the system. I note thst the plot_model/source_components call, which can take `color=["r", "g", "b"]` for instance, is a bit different in intention (the values are for the individual components within a plot, whereas this "new" feature is essentially the same as the existing `plot()` support shown in the example below).

There are also some other minor internal clean ups that have been added.

This has been pulled out of 

- #1712

# Example

## CIAO 4.17

```
sherpa-4.17.0> load_arrays(1, [1, 2, 5], [2, 7, 6])

sherpa-4.17.0> set_source(scale1d.mdl)

sherpa-4.17.0> fit()
Dataset               = 1
Initial fit statistic = 4.58694
Final fit statistic   = 1.48087 at function evaluation 4
Data points           = 3
Degrees of freedom    = 2
Probability [Q-value] = 0.476906
Reduced statistic     = 0.740436
Change in statistic   = 3.10607
   mdl.c0         4.28065      +/- 1.86146     

sherpa-4.17.0> plot_fit_resid(label=["fit", "resid"])
TypeError: unhashable type: 'list'
```

Since we use this below, it is worth noting that you can say the following in 4.17:

```
sherpa-4.17.0> plot("fit", "resid", label=["fit", "resid"])
```

sherpa

and

```
sherpa-4.17.0> plot("fit", "resid", alpha=[0.5, 1], color=["gray", "green"])
```

sherpa

## This PR

```

In [2]: load_arrays(1, [1, 2, 5], [2, 7, 6])

In [3]: set_source(scale1d.mdl)

In [4]: fit()
Dataset               = 1
Initial fit statistic = 4.58694
Final fit statistic   = 1.48087 at function evaluation 4
Data points           = 3
Degrees of freedom    = 2
Probability [Q-value] = 0.476906
Reduced statistic     = 0.740436
Change in statistic   = 3.10607
   mdl.c0         4.28065      +/- 1.86146     

In [5]: plot_fit_resid(label=["fit", "resid"])
```

sherpa

```
In [11]: plot_fit_resid(color=["gray", "orange"], alpha=[0.5, 1.0])
```

sherpa

Just for fun, to compare to `plot`:

```
In [13]: plot("fit", "resid", color=["gray", "orange"], alpha=[0.5, 1.0])
```

sherpa

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -571,6 +571,46 @@ def test_get_rmf_plot_recalc(idval, clean_astro_ui):
     assert rp.title is None
 
 
+def test_plot_rmf_mpl(clean_astro_ui, requires_pylab):
+    """Very basic check of plot_rmf.
+
+    Requires matplotlib to check the plot output.
+
+    """
+
+    from matplotlib import pyplot as plt
+
+    setup_example(1)
+    ui.plot_rmf()
+
+    fig = plt.gcf()
+    axes = fig.axes
+    assert len(axes) == 1
+    assert axes[0].xaxis.get_scale() == 'log'
+    assert axes[0].yaxis.get_scale() == 'log'
+
+    # Check the lines
+    assert len(axes[0].lines) == 10
+    colors = set()
+    for idx, lbl in enumerate(["0.6 keV",
+                               "0.72 keV",
+                               "0.87 keV",
+                               "1 keV",
+                               "1.2 keV"]):
+        l1 = axes[0].lines[idx * 2]
+        l2 = axes[0].lines[idx * 2 + 1]
+        assert l1.get_label() == lbl
+        col = l1.get_color()
+        assert col == l2.get_color()
+
+        # check we haven't seen this color
+        assert col not in colors
+        colors.add(col)
+
+        assert l1.get_ls() == '-'
+        assert l2.get_ls() == 'None'
+
+
 @pytest.mark.parametrize("idval", [None, 1, "one", 23])
 def test_get_order_plot(idval, clean_astro_ui):
     """Basic testing of get_order_plot: orders=None


sherpa/astro/data.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2015 - 2025
+#  Copyright (C) 2008, 2015-2025
```
hamogu, line 2: I feel we are changing stuff back and forth. You seem to prefer no staples, I think my script uses spaces, so I'll match my script to what @DougBurke does next time I change anything there. Nothing to be done in this PR, I just noted it.

sherpa/astro/ui/utils.py
```diff
@@ -13350,9 +13350,11 @@ def plot_source(self,
         data = self.get_data(id)
         if isinstance(data, DataPHA):
             # Note: lo/hi arguments mean we can not just rely on superclass
-            plotobj = self.get_source_plot(id, lo=lo, hi=hi, recalc=not replot)
-            self._plot(plotobj, overplot=overplot, clearwindow=clearwindow,
-                       **kwargs)
+            recalc = not replot
```
hamogu, line 13353: Is there a specific reason why `not replot` is pulled out into a separate variable here, but not below? I'm fine with either and don't want to make you spend time to change any of them, just curious.

DougBurke: Two reasons:

- most places it's only used once, but sometimes it gets used twice, where I'm more likely to create a variable to store it
- sometimes when I ask my editor to resize the text to fit we get a line that ends "recalc=not" and then "replot" at the start of the next line, which I find hard to parse

sherpa/astro/data.py
DougBurke: No - I have been changing it to remove spaces because I thought I remembered some of your code (maybe when I first tried out your pre-commit code) changed "2024 - 2025" to "2024-2025". I've been pretty-much "team spaces" from the get go, but I honestly don't care.

PR #2329, opened 2025-07-23
XSPEC: drop support for XSPEC 12.12.0 and 12.12.1

DougBurke:
# Summary

Drop support for XSPEC 12.12.0 and 12.12.1.

# Details

XSPEC 12.12.0 was released 2021-07 and XSPEC 12.12.1 in 2022-03. It would be nice to still support old code, but there is no test that we can still build with this old code, and dropping support cleans up the code base slightly.

We don't have an official policy on XSPEC support, but at some point we stop testing old versions (as shown by the fact none of the tests failed here and one XSPEC model here has been changed from optional to required (as it was added in 12.13.0) and so would have caused a built with xspec 12.12.0/1 to fail.

Release dates for XSPEC, from https://heasarc.gsfc.nasa.gov/xanadu/xspec/CHANGELOG.txt

| Version | Date |
| ------- | ---- |
| 12.12.0 | 2021-07 |
| 12.12.1 | 2022-03 |
| 12.13.0 | 2022-11 |
| 12.13.1 | 2023-07 |
| 12.14.0 | 2024-02 |
| 12.14.1 | 2024-08 |
| 12.15.0 | 2025-03 |

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -204,7 +204,7 @@ value) can be used. For most cases, the ``scripts/use_ciao_config``
 script can be used::
 
   % ./scripts/use_ciao_config
-  Found XSPEC version: 12.12.0
+  Found XSPEC version: 12.14.1
   Updating setup.cfg
   % git diff setup.cfg
   ...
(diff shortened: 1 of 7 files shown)


harlantc (approved):
lgtm

anetasie (approved):

PR #2330, opened 2025-07-28
Ensure `Fit.method` gives the same answer as `Fit._iterfit.method`

### LINKED ISSUE #2296: setting the `method` on a `Fit` object after creation has no effect
We can create a `Fit` object with a method, e.g.`fitter = Fit(data, model, stat, method=NelderMead())` and fit with that:
```Python
import numpy as np
from sherpa.models import Const1D
from sherpa.optmethods import NelderMead, MonCar
from sherpa.fit import Fit

data = Data1D('data1', x=np.arange(5),
                y=[3.4, 3.6, 3.3, 3.5, 3.4],
                staterror=[0.1, 0.2, 0.14, 0.09, 0.1])
model = Const1D('model')

fitter = Fit(data=data, model=model, method=NelderMead())

print(fitter.fit())
```
which prints:
```
datasets       = None
itermethodname = none
methodname     = neldermead
statname       = chi2
succeeded      = True
parnames       = ('model.c0',)
parvals        = (3.4306491540656756,)
statval        = 2.369514829472892
istatval       = 2362.5028974552783
dstatval       = 2360.1333826258056
numpoints      = 5
dof            = 4
qval           = 0.6681432947466703
rstat          = 0.592378707368223
message        = Optimization terminated successfully
nfev           = 116
```
Allegedly, one should also be able to change the method after the fact `fitter.method=LevMar()`. That change is recoded in "method name" line but note that the output (in particular the message and the nfev are exactly identical to the previous run:
```Python
fitter.method = MonCar()
model.reset()
print(fitter.fit())
```
Output:
```
datasets       = None
itermethodname = none
methodname     = moncar
statname       = chi2
succeeded      = True
parnames       = ('model.c0',)
parvals        = (3.4306491540656756,)
statval        = 2.369514829472892
istatval       = 2362.5028974552783
dstatval       = 2360.1333826258056
numpoints      = 5
dof            = 4
qval           = 0.6681432947466703
rstat          = 0.592378707368223
message        = Optimization terminated successfully
nfev           = 116
```

In contrast, if I create a new fitter object and set MonCar as the method from the beginning, I get a different output:
```Python
fittermoncar = Fit(data=data, model=model, method=MonCar())
model.reset()
print(fittermoncar.fit())
```
Output:
```
datasets       = None
itermethodname = none
methodname     = moncar
statname       = chi2
succeeded      = True
parnames       = ('model.c0',)
parvals        = (3.4306491540656876,)
statval        = 2.3695148294728914
istatval       = 2362.5028974552783
dstatval       = 2360.1333826258056
numpoints      = 5
dof            = 4
qval           = 0.6681432947466706
rstat          = 0.5923787073682228
message        = successful termination
nfev           = 588
```
(**edited to correct a copy & paste error**)

That's because setting the method does not propagate to the `fitter._iterfit` instance:
```Python
fitter.method, fitter._iterfit.method
```

I'm working on the PR for the documentation about fitting and run into this when I wanted to explain how to try different fitters. I'll include a fix to this bug into the PR I'm doing anyway. (I expect to push that tomorrow or the day after).

hamogu:
## Summary
Ensure `Fit.method` gives the same answer as `Fit._iterfit.method`. FIxes #2296

## Details
Previously, those objects could be different, leading to inconsistent results when a user though that the optimization method had been changed when in reality the previous setting was still active.
One optimization method would be reported in printing, while in practice a different method was used.

This PR changes some of the tests, because the tests contain hardcoded expected values that were actually run with the wong method.

### DIFF
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1145,13 +1145,6 @@ class Fit(NoNewAttributesAfterInit):
         self.data = data
         self.model = model
 
-        # It is not clear why this can not just set self.stat but
-        # there is a comment below that suggests it should not be done
-        # here.
-        #
-        statobj = Chi2Gehrels() if stat is None else stat
-
-        self.method = LevMar() if method is None else method
         self.estmethod = Covariance() if estmethod is None else estmethod
         self.current_frozen = -1
 
@@ -1163,26 +1156,33 @@ class Fit(NoNewAttributesAfterInit):
 
         # Set up an IterFit object, so that the user can select
         # an iterative fitting option.
-        self._iterfit = IterFit(self.data, self.model, statobj,
-                                self.method, iopts)
-
-        # We need to set the statistic *after* creating the _iterfit
-        # attribute
-        #
-        self.stat = statobj
+        self._iterfit = IterFit(self.data,
+                                self.model,
+                                Chi2Gehrels() if stat is None else stat,
+                                LevMar() if method is None else method,
+                                iopts)
 
         super().__init__()
 
     @property
     def stat(self) -> Stat:
-        """Return the statistic value"""
-        return self._stat
+        """Return the statistic object"""
+        return self._iterfit.stat
 
     @stat.setter
     def stat(self, stat: Stat) -> None:
         """Ensure that we use a consistent stat object."""
-        self._stat = stat
         self._iterfit.stat = stat
+
+    @property
+    def method(self) -> OptMethod:
+        """Return the method object"""
+        return self._iterfit.method
+
+    @method.setter
+    def method(self, method: OptMethod) -> None:
+        """Ensure that we use a consistent method object."""
+        self._iterfit.method = method
 
     def __setstate__(self, state):
         self.__dict__.update(state)
--- a/sherpa/fit.py
+++ b/sherpa/fit.py
@@ -1090,6 +1090,10 @@ def _check_length(dep) -> int:
 class Fit(NoNewAttributesAfterInit):
     """Fit a model to a data set.
 
+    .. versionchanged:: 4.18.0
+       Prior to 4.18.0 changing the method field would not change the method used
+       by fit and esterrors, just the label shown when a ``Fit`` object is printed.
+
     .. versionchanged:: 4.17.0
        Changing the stat field now changes the internal IterFit
        object as well.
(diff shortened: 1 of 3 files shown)


sherpa/sim/tests/test_sim.py
```diff
@@ -456,25 +456,26 @@ def test_mh(setup, caplog):
...
-    row20[0] = 20
+    row551 = np.asarray([551, -8975.953,  1.069770,  9.219028,
+                         2.570941,  2.586583,  47.36505])
+    row552 = row551.copy()
+    row552[0] = 552
 
     out.seek(0)
     fitvals = np.loadtxt(out)
-    assert fitvals.shape == (21, 7)
+    assert fitvals.shape == (553, 7)
```
DougBurke, line 475: It is interesting that the fit is a lot more "inefficient" here. If we had infinite-time I'd suggest we look at whether we can get a "simpler" test case (one that requires less evaluations), but I am **not** suggesting it here.

However, the change does make sense (`NelderMead/Simplex` often requires more iterations than `LevMar`).

sherpa/fit.py
```diff
@@ -1163,27 +1156,34 @@ def __init__(self,
...
         self._iterfit.stat = stat
 
+    @property
+    def method(self) -> OptMethod:
+        """Return the method object"""
+        return self._iterfit.method
+
+    @method.setter
+    def method(self, method: OptMethod) -> None:
+        """Ensure that we use a consistent method object."""
```
DougBurke: So, I will note that the same discussion is relevant for `stat` as it is for `method`, and only the latter has been added here (i.e. the existing code already shows the thing I am about to asl about, and I probably wrote the original code).

We have a docstring on both the `@property` and `@property.setter` functions. Only the `@property` is used, so should we change them (both `stat` and `method`) to something like

```
"""Set or get the statistic/method object.""
```

and then we can remove the docstring from the `@foo.setter` version?

hamogu:
I can add in that text, but I think the important location is the changelog. If my code goes that deep into sherpa internals and it suddenly behaves differently, I would either look there or at a git diff. Most Sherpa users uses the UI, so they cannot possibly encounter this and those that do use the OO mostly just create new fit objects every time they need a new method.

It's still a bug that should be fixed, but only a power-power-user would be effected by that - and those are the people who know how to figure out what changed.

hamogu (commented):
Now the Linux build passes on CI, but the ARM build fails (despite passing locally on my ARM system). I think that confirms that this is a numerical issue related to versions of libaries and OS, so I'm just going to increase the numerical tolerance allowed for this test.

sherpa/tests/test_fit_unit.py
```diff
@@ -1876,22 +1876,22 @@ def test_fit_single(stat, usestat, usesys, finalstat):
...
+    (Chi2Gehrels, False, False, fit_gehrels_ff, 1e-6),
+    (Chi2XspecVar, False, True, fit_xvar_ft, 1e-6),
+    (Chi2ConstVar, False, True, fit_cvar_ft, 1e-6),
+    (Chi2ModVar, True, True, fit_mvar_xt, 1e-6),
+    (Chi2ModVar, True, False, fit_mvar_xf, 1e-6),
+    (Chi2ModVar, False, False, fit_mvar_xf, 1e-6),
+    (Chi2ModVar, False, True, fit_mvar_xt, 1e-6),
+
+    (Cash, True, True, fit_cash, 1e-6),
+    (CStat, True, True, fit_cstat, 2e-4),  # This tests gives different results on Linux and ARM
```
DougBurke, line 1892: I am honestly surprised that this is needed (given the other tests and our tests using CStat elsewhere), but the CI tests do show the difference. I don't have the energy to try and find out why (especially not when I do have a lot of internal changes to the optimization code waiting to be finished).

PR #2333, opened 2025-08-01
Fix 2332

DougBurke:


DougBurke:
@anetasie - this is also not the "problematic" part of #2211 but was found when working on it. I'm hoping it's easier to review this part than the more-problematic parts which are left in #2211 (and not related to this particular bug fix).

docs/evaluation/combine.rst
```diff
@@ -441,24 +493,134 @@ manner to models::
     >>> g2.pos
     <Parameter 'pos' of model 'g2'>
     >>> print(g2.pos)
-    val         = 0.504315379302
-    min         = -3.40282346639e+38
-    max         = 3.40282346639e+38
-    units       = 
+    val         = 0.5001684477780305
```
hamogu, line 497: We run doctestplus with `doctest_plus_atol=1e-4` meaning that string output numbers are allowed to differ at the level of 1e-4 (for rtol we use the default, I think it's 1e-6). 
Of course, it's good to have the full precision as shown to the user on a screen, and I'm not requesting changes, I'm just saying if you want to display fewer digits similar to the previous version it wouldn't cause problems.

docs/evaluation/combine.rst
```diff
@@ -441,24 +493,134 @@ manner to models::
...
+What parameters are free to be fit?
+===================================
+
+When :doc:`using an optimiser <../optimisers/index>`, it can be
+necessary to restrict the optimisation to a subset of the parameters
+of the model. Sherpa marks each parameter as frozen or thawed, where
+frozen parameters are **not** changed during a fit.
+
+The string display of a model indicates whether each parameter is
+frozen or thawed - under the ``Type`` column - and the
```
hamogu: The example below, however, doesn't show a frozen parameter in the type column, only a linked one. Maybe freeze one parameter outright and then explain the differences and similarities between frozen and linked?

docs/evaluation/combine.rst
```diff
@@ -441,24 +493,134 @@ manner to models::
...
+How best to access the thawed parameters?
+=========================================
+
+The :py:attr:`~sherpa.models.model.Model.thawedpars`
+attribute of a model will return the current numeric values of the
+thawed parameters. It can also be used to **change** just the thawed
+parameters, by setting it to a sequence of numbers (such as a NumPy
+array or list).
+
+    >>> list(map(float, mdl.thawedpars))  # avoid np.float(...) output with NumPy 2
```
hamogu: Same output (for either numpy version as far are I can tell), but avoiding the 
`map(float, ...)` construct that might not be as familiar as a print statement to most users.
```suggestion
    >>> print(np.array(mdl.thawedpars))
```

sherpa/sim/sample.py
```diff
@@ -86,19 +94,25 @@ def multivariate_t(mean, cov, df, size=None, rng=None):
...
 # TODO: should this pass the size value through to multivariate_t?
-def multivariate_cauchy(mean, cov, size=None, rng=None):
+#
+def multivariate_cauchy(mean: ArrayType,
+                        cov: np.ndarray,
+                        size: tuple[int, ...] | None = None,
+                        rng: random.RandomType | None = None
+                        ) -> np.ndarray:
     """
     This needs to be checked too! A reference to the literature the better
```
anetasie, line 118: Cauchy distribution is the T distribution with 1 degree of freedom.

docs/evaluation/combine.rst
DougBurke: When I'm doing this sort of thing I just want to grab the screen output I get and add it to the documentation. I don't want to then have to think to myself "what precision should I use for each line", as that would not help me writing documentation!

We could think about adding pre/post stanzas to these docs to explicitly call `np.set_printoptions` (just been looking at this for a different PR I think I am not going to work on) but then I'd need to remember to set them when creating documentation like this.

So I'm not sure what the best approach here is.

sherpa/sim/sample.py
DougBurke: This comment has existed for the whole time we've had Sherpa in git!

We only use `multivariate_cauchy` in one place - and that is in a test! So I suggest we remove it rather than try to fix the documentation.

docs/evaluation/combine.rst
hamogu: The best approach is what you do right now. No need to change anything. I just want to make sure we are all aware that the doctests don't actually check the full precision.

hamogu:
I'm in the review, but didn't finish it last night. My comments will come in later today.

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -2580,3 +2581,115 @@ def test_sample_flux_pha_bkg_full_model(idval, make_data_path, clean_astro_ui,
...
+        case _:
+            assert False
+
+    assert ui.get_source(1).thawedpars == pytest.approx([1.0, 0.0])
+    return cpt2
+
+
+@pytest.mark.parametrize("setup", ["A", "B", "C"])
+def test_2337_linked_par_full(setup, clean_astro_ui):
+    """What happens when a linked-parameter is used for sampe_photon_flux?
```
hamogu:
```suggestion
    """What happens when a linked-parameter is used for sample_photon_flux?
```

sherpa/astro/ui/tests/test_astro_ui_fluxes.py
```diff
@@ -2580,3 +2581,115 @@ def test_sample_flux_pha_bkg_full_model(idval, make_data_path, clean_astro_ui,
...
+    assert res.shape == (5, 4)
+    assert res[:, 0] == pytest.approx([118.67499316, 146.11307685, 116.10343197, 125.36669748, 125.86704318])
+    assert res[:, 1] == pytest.approx([14.98264164, 18.13393656, 15.79721333, 15.80157972, 15.8858083 ])
+    assert res[:, 2] == pytest.approx([5.99565486, 7.77280365, 4.44169833, 6.36608732, 6.36504029])
+    assert res[:, 3] == pytest.approx([0] * 5)
+
+
+@pytest.mark.parametrize("setup", ["B", "C"])
+def test_2337_linked_par_subset(setup, clean_astro_ui):
+    """What happens when a linked-parameter is used for sampe_photon_flux?
```
hamogu:
```suggestion
    """What happens when a linked-parameter is used for sample_photon_flux?
```

sherpa/ui/tests/test_ui_unit.py
```diff
@@ -1579,3 +1580,200 @@ def check_ratio(v1, v2):
...
+                                mdl2: ArithmeticModel,
+                                f1: FitResults,
+                                f2: FitResults,
+                                c1: ErrorEstResults,
+                                c2: ErrorEstResults,
+                                swap: bool = False) -> None:
+    """Simple checks that the fit results match for linked pars.
+
+    This is just pulled out of the test to avoid complication, and is
+    just a test that the lniked-parameter fit is working as expected.
```
hamogu:
```suggestion
    just a test that the linked-parameter fit is working as expected.
```

sherpa/astro/flux.py
```diff
@@ -517,11 +555,12 @@ def sample_flux(fit, data, src,
     # but is not sufficient to guarantee the match.
     #
     if npar < mpar:
+        # TODO: will this fail with linked parameters.
         full_pars = dict(map(reversed,
-                             enumerate([p for p in fit.model.pars
-                                        if not p.frozen])))
+                             enumerate(fit.model.get_thawed_pars())
+                             ))
         cols = []
```
hamogu: I didn't understand this line, so I had to disect it and finally figured out that it's 
equivalent to the line below (chaining `map`, `reversed`, and `enumerate` is not the most intautive wat to do it to me). 
However, either work, so you might not want to touch it.
```suggestion
        from iterools import count
        full_pars = {p: i for p, i in zip(fit.model.get_thawed_pars(), count())}
```

sherpa/sim/sample.py
```diff
@@ -199,45 +220,61 @@ def get_scales(self, fit, myscales=None):
...
                     try:
                         t = fit.est_errors(parlist=(par,))
-                        if t.parmins[0] is not None and t.parmaxes[0] is not None:
-                            scale = numpy.abs(t.parmins[0])
-
-                        else:
 
-                            if t.parmins[0] is None and t.parmaxes[0] is not None:
-                                scale = numpy.abs(t.parmaxes[0])
+                        # QUS: could this not be
```
hamogu, line 224: This code seems to just look at the first value or parmin/parmax and say, "hey, let's get a reasonable guess. Let's try parmin. Uups, that's a None. Well, we'll take parmax instead."
I;ve done very similar things in hand-written code for analysis. It's kind of hand-wavy. I don't think we can do much better though (could at least take the max or mean of [parmax,parmin] is both are available), but fundamentally, I'm not sure can can do much better without actually tracking a distribution of numbers.

So, would that belong into this PR or not?

sherpa/astro/tests/test_astro_flux.py
```diff
@@ -0,0 +1,310 @@
...
+                           [1.67167473e-07, 2.21044981e-07, 2.10197208e-07, 2.39277125e-07, 1.81511770e-07],
+                           [1.67167473e-07, 2.21044981e-07, 2.10197208e-07, 2.39277125e-07, 1.81511770e-07])
+                          ])
+def test_sf_full_linked(setup, fluxes1, fluxes2, link, lo, hi):
+    """Check sample_flux with a basic dataset/model: Data1D/DataPHA
+
+    Three things that are checked
+      - behaviour with an upper-limit
+      - checks some, but not all, cases of lo/hi set or unset
+        (the behaviour wheh only one is given is "interesting" for
```
hamogu:
```suggestion
        (the behaviour when only one is given is "interesting" for
```

sherpa/sim/sample.py
DougBurke: I created

- #2340

to track this, so I don't think we want to change it here.

sherpa/astro/flux.py
DougBurke: Or

```python
full_pars = {p: i for i, p in enumerate(fit.model.get_thawed_pars())}
```

hamogu: Even better! Avoids the extra import and `enumerate` is more familiar to most people than `count()`.
At this point we can stop playing [code-golf](https://codegolf.stackexchange.com) since the goal is not to make the shortest version, but just one that's easily readable.

docs/evaluation/combine.rst
DougBurke: Hopefully the new version is better.

sherpa/sim/sample.py
DougBurke: I've added that as #2349

DougBurke:
Hopefully I've addressed all comments now, but I may have missed something.

hamogu (approved):

sherpa/astro/tests/test_astro_flux.py
```diff
@@ -0,0 +1,310 @@
...
+    assert fit.model.thawedpars == pytest.approx([17.076190476182614,
+                                                  6.571428571477132])
+
+
+def setup_data1d_linked_par(link: bool
+                            ) -> tuple[Fit, ArithmeticModel, ArithmeticModel]:
+    """Create simple setup for sample_flux testing
+
+    Fit a straight line but using a "complex" expression so that
+    we can have multiple components with free paramaters. The
```
wmclaugh:
```suggestion
    we can have multiple components with free parameters. The
```

PR #2336, opened 2025-08-05
Fix 2335

DougBurke:
Hmmmm.


DougBurke:
For the structured datatype we have two or three related 2D types (typing this from memory so the details may be slightly wrong)

1. npar by niter which contains the free parameters for each iteration
2. npar + 1 by niter, which is the statistic value followed by the free parameters (or it's the other way around, but for this discussion the order doesn't matter)
3. npar + 2 by niter, which adds in as the final column a 0 or 1 to indicate whether the free parameters have been clipped

So there would probably need to be some internal routines (e.g. add the statistic column).

What would be nice would be to not break existing code too much (although by adding a clip column we possibly causing issues for people). I did try briefly setting `dtype` to a list when creating the `ndarray` but it didn't work immediately (particularly if we want to label the `clipped` column as a boolean or integer rather than a float) and I don't have time/experience to know how to really do it.

DougBurke:
For the "do we really need anything but `clip="hard"` question: I originally added that for the flux code to allow people to run with the same behaviour as older releases. I argue it's a similar issue here. 

I think, after adding it, that I came to the conclusion that `clip="soft"` wasn't as useful as I thought it would, but I forget now and not sure I have time to investigate.

hamogu (approved):
LGTM

PR #2341, opened 2025-08-12
CI: add an  aarch64 workflow

### LINKED ISSUE #2338: add CI run using linux arm64
The `ubuntu-24.04-arm` runners are now available - https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/ - so we should look at adding a build. 

This would likely mean we could stop (or replace) the  weekly `Weekly cron (arch)` action, which also happens to currently be broken.

DougBurke:
# Summary

Add an aarch64 CI run. Fix #2338.

# Details

I just want to make sure we have an aarch64 build being tested. I do not think  it worth adding a full set of tests for this platform. This replaces the "arch" workflow, which used emulation, with the new Linux aarch64 runner.

As part of this work two of the scripts used in the conda GitHub workflows have been reworked to also work when run on the aarch64 platform. The conda setup is currently not used (but is there if we decide to add a conda test) and the ds9 setup is used here (as the workflow is written to emulate the conda setup required by the script).

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -5,12 +5,12 @@ echo "** uname -s: `uname -s`"
 echo "** uname -m: `uname -m`"
 
 if [ "`uname -s`" == "Darwin" ] ; then
-
     if [ "`uname -m`" == "x86_64" ]; then
-	sys="64"
+       sys="64"
     else
-	sys="arm64"
+       sys="arm64"
     fi
+
     compilers="clang_osx-${sys} clangxx_osx-${sys} gfortran_osx-${sys}"
 
     #Download the macOS 11.0 SDK to the CONDA_BUILD_SYSROOT location for the Conda Compilers to work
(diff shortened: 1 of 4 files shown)


DougBurke:
Rebased onto #2453 as it changes the DS9 download code (and updates versions).

DougBurke:
Apparently the aarch runners are transitioning from ARM to GitHub at the moment, so it's not worth looking at until that has finished (mid June at the earliest I believe).

DougBurke:
Rebased as #2484 has been merged, which does not change how the workflow runs  [**edited to add** actually that's not true, as we now need to install the FFTW libraries] but is useful to test against. Also update as #2457 has landed and match #2511 for updates to actions.

DougBurke:
That's "interesting" - there is one test failure and it's a docstring from the scipy optimizer code:

```
_____________ [doctest] sherpa.optmethods.optscipy.Scipy_Minimize ______________
246 ...
247 
248 We can change the method used in the optimization and set some of the
249 options for the optimizer, such as the goal for the precision and the
250 maximum number of function evaluations before the fit stops.
251 
252 >>> optimizer.method = 'TNC'
253 >>> optimizer.options = {'maxfun': 200, 'xtol': 1e-4}
254 >>> model.reset()
255 >>> print(fit.fit())
Differences (unified diff with -expected +actual):
    @@ -1,4 +1,16 @@
     datasets       = None
    -...
    -message        = Converged (|x_n-x_(n-1)| ~= 0)
    -...
    +itermethodname = none
    +methodname     = scipy_minimize
    +statname       = chi2
    +succeeded      = False
    +parnames       = ('const.c0', 'gauss.fwhm', 'gauss.pos', 'gauss.ampl')
    +parvals        = (25.895712467749657, 3.3057804286271653, 5.044191081651596, 42.13425089538124)
    +statval        = 6.883835354730454
    +istatval       = 700.2311623262916
    +dstatval       = 693.3473269715612
    +numpoints      = 10
    +dof            = 6
    +qval           = 0.3317238566482942
    +rstat          = 1.1473058924550756
    +message        = Max. number of function evaluations reached
    +nfev           = 1005

/opt/hostedtoolcache/Python/3.13.14/arm64/lib/python3.13/site-packages/sherpa/optmethods/optscipy.py:255: DocTestFailure
----------------------------- Captured stdout call -----------------------------
WARNING: fit failed: Max. number of function evaluations reached
------------------------------ Captured log call -------------------------------
WARNING  sherpa.optmethods.buildin:buildin.py:289 fit failed: Max. number of function evaluations reached
```

PR #2345, opened 2025-08-19
Make doctests pas again

hamogu:
## Summary
work on docs for readthedocs

## Details
This PR consists of sevaral parts:

- Unify doctsts to (hopefully) pass on numpy 1.x and 2.x
- Fix the examples in `simulate.rst` that explain how to use `fake_pha`. THey have been out of sync with code changes to `fake_pha` since about Sherpa 4.16.1.


DougBurke:
@hamogu - for the last point, can you check how this compares to 

- #2122 
- #2043

as I assume it subsumes some, or maybe all, of those PRs (I have paged out my knowledge of the simulation code and what I was doing with those PRs)?

docs/evaluation/simulate.rst
```diff
@@ -58,39 +64,90 @@ want to skip this section.
...
+In this example, we use a datafile from Sherpa's test data files
+(which are large and not installed by default
+:ref:`but can be downloaded <install-sherpa-test-data>`).
+
+ .. plot::
+    :include-source:
+    :context:
+    :nofigs:
+
+    >>> from sherpa.utils.testing import get_datadir
```
DougBurke, line 86: Technically it would be better to have these two lines (that define `data_dir`) in a separate section (to make it clear it's not part of the "call `read_pha` or `DataPHA`" logic, but I don't think we need to do this here (or maybe a later commit will change things).

docs/evaluation/simulate.rst
```diff
@@ -58,39 +64,90 @@ want to skip this section.
...
+We will need to wrap this model with the response (ARF and RMF), which takes care
+of converting from channels to energy:
+
+.. plot::
+    :include-source:
+    :context:
+    :nofigs:
+
+    >>> from sherpa.astro.instrument import Response1D
+    >>> resp = Response1D(data)
```
DougBurke: We should be able to just say

```
>>> resp = data.get_full_response()
```

but I'm not sure if it's worth changing here.

hamogu:
I see a runtime of 11.10 s for the doctests, and that's with the mcmc tests run and in a dvctest-only run (meaning some of that time is setup time, adding those tests to a CI run that runs all the test anyway will have less impact). I don't think that's too much to ask for. The runtime is dominated the mcmc doctests and that (with the commit I'm about to push) will only run when Arvin is installed - i.e. it will be skipped in most configurations.

hamogu:
The conda check failed for a https timeout in downloading. I've restarted one of the jobs, but I'll save the time to restart all of them, so please review now - if there are any changes I'll need to push again anyway and that will trigger all checks again anyway.

Note that I'm not including "run the doctests on rst file in CI" here. I tried to make commit for that, but it's hard to set up in some configurations (e.g. conda with an installed package, it's easier on a source checkout) and it really only makes sense in a setup with all dependencies installed since most of the "rst" files use matplotlib or bokeh and sometimes other optional dependencies (e.g. arviz).

I think we should add a single CI run that's set up to run the rst files in addition to the normal tests, but we'll need to re-organize the CI runs anyway when we drop numpy 1.X and maybe also Python 3.10, so I suggest that we'll address that in the same time in a separate PR - for now we will continue to run the rst  files with `purest --doctest-rst --doctest-plus` locally when needed.

docs/evaluation/simulate.rst
hamogu: I added some comments around those lines.

sherpa/conftest.py
```diff
@@ -706,7 +706,8 @@ def add_sherpa_test_data_dir(doctest_namespace):
     in the future, but it seems more important that this keeps any
     extra markup out of those files and keep them clean.
     '''
-    doctest_namespace["data_3c273"] = 'sherpa/astro/datastack/tests/data/'
+    sherpa_root = os.path.dirname(os.path.abspath(__file__))
+    doctest_namespace["data_3c273"] = os.path.join(sherpa_root, 'astro/datastack/tests/data/')
```
DougBurke: In https://github.com/sherpa/sherpa/pull/2363 I went with

```python
import importlib.resources
```
...

```python
    # Ensure an absolute path is used and that the string ends in a
    # "/".
    #
    spath = importlib.resources.files("sherpa")
    dpath = spath / 'astro' / 'datastack' / 'tests' / 'data'
    doctest_namespace["data_3c273"] = f"{dpath}/"
```

I don't know whether `importlib,resources.files` is always going to match `__file__` but I  think it should. I also don't know which is "better".

hamogu: Sure. We can do that. Let's be fancy, I'll change that.

docs/evaluation/simulate.rst
```diff
@@ -106,39 +165,57 @@ slightly different answers. With default settings, the input model is
...
-spectrum. :py:func:`~sherpa.astro.fake.fake_pha` works if one or more responses
-are set for the input :py:class:`~sherpa.astro.data.DataPHA` object.
-
-It is also possible that the input model already includes the
-appropriate responses and scalings. In this case, ``is_source=False``
-can be set to indicate to sherpa that the model is not a naked source
-model, but includes all required instrumental effects. In this
-way, :py:func:`~sherpa.astro.fake.fake_pha` can be used with arbitrarily
+spectrum.
+The models that is passed into :py:func:`~sherpa.astro.fake.fake_pha`
```
wmclaugh:
```suggestion
The models that are passed into :py:func:`~sherpa.astro.fake.fake_pha`
```

docs/evaluation/simulate.rst
```diff
@@ -106,39 +165,57 @@ slightly different answers. With default settings, the input model is
...
+can be arbitrarily complex and can contain multiple components. For example,
+:py:func:`~sherpa.astro.fake.fake_pha` can be used
```
wmclaugh: grammer- replace "can be used complex models".  Perhaps "can be used with complex models"  or "can use complex models"?

docs/evaluation/simulate.rst
```diff
@@ -106,39 +165,57 @@ slightly different answers. With default settings, the input model is
...
 complex models which may include such components as the instrumental
```
hamogu:
```suggestion
:py:func:`~sherpa.astro.fake.fake_pha` can simulate
models which may include such components as the instrumental
```

PR #2347, opened 2025-08-20
Install sherpa-test-data in pip is requested

hamogu:
## Summary
Fix bug in pip woorkflow on CI

## Details
The pip setup did not actually install the test data is set to "package", probably due to an editing error when "pyest-xvfb" was added as a step in the workflow.


DougBurke:
I have long been confused over our CI's naming; e.g 

- what `test-data: submodule/package/none` really mean,
- how they are meant to interact with `install-type`, 
- and whether we have the same logic for `pip` and `conda` workflows

and this has likely contributed to issues like this one is fixing.

I'm not sure now is the time to address these issues.

hamogu:
It's actually easier than you think - and this is a very targets fix that I'd like to have (though it's not strictly required) for #23435.

`install_type` sets the way sherpa itself is installed and can be:
  - 'install' = `pip install .`
  - 'develop' = `pip install -e`

`test-data` sets the way the test data is installed and can be:
  - 'none' = no test data
  - 'package' = `pip install sherpa-test-data`
  - 'submodule' = no install, but have it present in the submodule so that test dat can be found there

pip and conda workflows are following roughly the same path, but there is not need to be exactly parallel.

While I think that this setup is more complicated than it has to be (why don't we just pick that the test data will always be installed as a package and don't bother to also support it as an uninstalled submodule? Do we really need to test a development install? I would think it's the responsibility of the pip/setuptools maintainers to ensure that that works?) I agree that now is not the time to change that.

I just want to fix the simple problem that in the pip workflow the test data is not installed when set to "package" because someone accidentally edited that step of the instructions into "pip install aytest-xvfb" at some point in the past.

.github/workflows/ci-pip-workflow.yml
```diff
@@ -92,21 +92,22 @@ jobs:
           NUMPYVER='numpy'
         fi
         pip install ${NUMPYVER} ${FITSBUILD} ${MATPLOTLIBVER} ${BOKEHVER} ${ARVIZVER}
+        pip install pytest-xvfb
```
DougBurke: If this is going to be moved here then why not do it in line 94?

.github/workflows/ci-pip-workflow.yml
```diff
@@ -92,21 +92,22 @@ jobs:
...
 
     - name: Build Sherpa (install)
       if: matrix.install-type == 'install'
       run: |
-        pip install . --verbose
+        pip install .[test] --verbose
```
DougBurke: Why add the test requirement here? There's an argument to be said that we should ensure we can do a typical build (e.g. as documented)  where we don't have the test requirements installed (to make sure things work for user's who don't install them), and the `sherpa_test` command should download the necessary pacakges (I believe).

We also have the "Submodule test with pytest" that installs files via `test_requirements` below, which sort of does the same thing as this here.

One issue I have (with the existing workflow but I think this complicates it a tad more) is that we have `install-type` and `test-data` and they interact in interesting ways to determine what is tested / used / built.

.github/workflows/ci-pip-workflow.yml
hamogu: Either works. I thought it's easier to read to have one line with all the variables and one line with the hardcoded names (currently only one), but I'm fine with putting it on the same line, too, if other people prefer that.
Since pip doesn't do a long dependency resolution like conda does, there shouldn't be a noticeable effect on runtime.

.github/workflows/ci-pip-workflow.yml
hamogu: I don't think it's that "interesting". If the data is available, the test that use the data will be run, if it's not they won't be; that's true for any "install-type". There really is no big and secret interaction between them.

I cna remove the `[test]`. As @DougBurke says, it should not be needed **right here**, but the modules required for the tests will have to be installed somewhere, otherwise the tests won't run. Doesn't have to be on this line, but can happen (automatically or not) elsewhere.

PR #2352, opened 2025-08-25
UI: remove XSPEC table support from load_table_model

### LINKED ISSUE #902: Can we remove support for XSPEC models from load_table_model
In the CIAO 4.9 release I added the warning to the load_table_model function that support for XSPEC models would be removed ([*]). I'd like that time to be now...

[*] Why should it be removed? Well, the current code is very convoluted, and this would help slightly. I also believe it is confusing for users, since the table model interface is *very* basic whereas the XSPEC table model support is much-more complex, and so combining the two is confusing. I would say that the XSPEC table model support is more-closely related to the template model support than table model, but as I have essentially no experience in the template model code I can't say for sure. And I would not want the template model code to have special-case behavior for the XSPEC table models.

```
    def load_table_model(self, modelname, filename, method=sherpa.utils.linear_interp, *args, **kwargs):
        """Load tabular or image data and use it as a model component.
    
        .. note:: Deprecated in Sherpa 4.9
                  The new `load_xstable_model` routine should be used for
                  loading XSPEC table model files. Support for these files
                  will be removed from `load_table_model` in the next
                  release.
    
        A table model is defined on a grid of points which is
        interpolated onto the independent axis of the data set.  The
        model will have at least one parameter (the amplitude, or
        scaling factor to multiply the data by), but may have more
        (if X-Spec table models are used).
```

DougBurke:
# Summary

Remove support for XSPEC table models from load_table_model as this has long been deprecated. Users must use load_xstable_model with XSPEC table models. Fix #902.

# Details

We have long said "do not use load_table_model" with XSPEC table models. It is time to make this a reality.

# Example

With CIAO 4.17 we have

```
sherpa In [1]: load_table_model("foo", "xspec-tablemodel-RCS.mod")
WARNING: Use load_xstable_model to load XSPEC table models
```

We now have the following. I do not want to complicate the `load_table_model` with logic to provide a "nicer" error message here.

```

In [2]: load_table_model("goo", "xspec-tablemodel-RCS.mod")
ModelErr                                  Traceback (most recent call last)
Cell In[2], line 1
----> 1 load_table_model("goo", "xspec-tablemodel-RCS.mod")

File <string>:1, in load_table_model(modelname, filename, method, *args, **kwargs)

File ~/sherpa/sherpa-4/sherpa/astro/ui/utils.py:11054, in Session.load_table_model(self, modelname, filename, method, *args, **kwargs)
  11052 tablemodel.method = method
  11053 tablemodel.filename = filename
> 11054 tablemodel.load(x, y)
  11055 self._tbl_models.append(tablemodel)
  11056 self._add_model_component(tablemodel)

File ~/sherpa/sherpa-4/sherpa/models/basic.py:2079, in TableModel.load(self, x, y)
   2076     return val
   2078 self.__y = _check(y)
-> 2079 self.__x = _check(x)
   2081 # clear the filtered array
   2082 self.__filtered_y = None

File ~/sherpa/sherpa-4/sherpa/models/basic.py:2074, in TableModel.load.<locals>._check(val)
   2072     val + 0
   2073 except Exception:
-> 2074     raise ModelErr(f"Unable to treat array as numeric: {val}") from None
   2076 return val

ModelErr: Unable to treat array as numeric: ['tau         ' 'beta        ' 'T           ']

```

### DIFF
--- a/sherpa/astro/io/tests/test_io.py
+++ b/sherpa/astro/io/tests/test_io.py
@@ -39,7 +39,7 @@ from sherpa.utils.testing import requires_data, requires_fits, \
 def test_mod_fits(make_data_path, clean_astro_ui, caplog):
     """Can we read in an XSPEC table model with load_table_model.
 
-    This is no-longer supported.
+    This is no longer supported.
     """
 
     # It is not really important to know how this fails, but test the
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -10979,7 +10979,7 @@ class Session(sherpa.ui.utils.Session):
         interpolated onto the independent axis of the data set. The
         model has a single parameter, ``ampl``, which is used to scale
         the data, and it can be fixed or allowed to vary during a fit.
-        The ``load_xstable_model`` must be used with XSPEC table
+        The ``load_xstable_model`` routine must be used with XSPEC table
         models.
 
         Parameters


hamogu (approved):

sherpa/astro/ui/utils.py
```diff
@@ -10975,16 +10975,12 @@ def load_table_model(self, modelname, filename,
...
-                  The new `load_xstable_model` routine should be used for
-                  loading XSPEC table model files. Support for these files
-                  will be removed from `load_table_model` in the 4.17
-                  release.
-
         A table model is defined on a grid of points which is
         interpolated onto the independent axis of the data set. The
         model has a single parameter, ``ampl``, which is used to scale
         the data, and it can be fixed or allowed to vary during a fit.
+        The ``load_xstable_model`` must be used with XSPEC table
```
wmclaugh:
```suggestion
        The ``load_xstable_model`` routine must be used with XSPEC table
```

sherpa/astro/io/tests/test_io.py
```diff
@@ -29,37 +29,27 @@
...
 
 
 @requires_data
 @requires_fits
-@requires_xspec
 def test_mod_fits(make_data_path, clean_astro_ui, caplog):
     """Can we read in an XSPEC table model with load_table_model.
 
-    This approach is deprecated. Use load_xstable_model instead.
+    This is no-longer supported.
```
wmclaugh:
```suggestion
    This is no longer supported.
```

Issue #1446, opened 2022-02-15
can DataIMG/2D/... be sent 2D arrays?


DougBurke:
Currently we have to say

```
>>> import numpy as np
>>> x1, x0 = np.mgrid[1:4, 1:5]
>>> from sherpa.astro.data import DataIMG
>>> y2 = (x0 - 2.5)**2 + (x1- 2)**2
>>> y = np.sqrt(y2)
>>> d = DataIMG('img', x0.flatten(), x1.flatten(), y.flatten(), shape=y.shape)
>>> print(d)
name      = img
x0        = Int64[12]
x1        = Int64[12]
y         = Float64[12]
shape     = (3, 4)
staterror = None
syserror  = None
sky       = None
eqpos     = None
coord     = logical
```

We could allow the user to say

```
DataIMG('img', x0, x1, y)
```

- no need to flatten
- we can infer the shape value so the user doesn't need to specify it

This would also hold for `Data2D` and the `*Int` variants. But not Loki.

There is one difference, as we would be explicitly copying the values rather than potentially storing a reference to them, so that `data.x0[1] += 1` would not be reflected in the input argument (although this is not guaranteed when sending in 1D arrays anyway).

PR #2355, opened 2025-08-28
Simple DataIMG(Int) creation from 2D arrays

hamogu:
## Summary
Add convenience methods to generalt `DataIMG` and `DataIMGInt` from 2d arrays with or without WCS information.

## Details
In principle, all of Sherpa's 2D data classes (Data2D, Data2DInt, DataIMG, DataIMGInt) can work with arbitraily sorted and sparse data, since all the relevant arrays (the independent coordinates `x0` and `x1` as well as the dependent data `y) are strored as flattened 1D arrays.
However, in practice we often work with regularly gridded data, e.g. images or even images with a FITS WCS system.

Reading that to turn it into 1D data is a bit cumbersome and has a high potential to mass up the ordering of columns and rows, in particular since the numpy convention (based on C) of [row, column] ordering is the exact opposite of the FITS convention (based on Fortran I believe) of [column, row] ordering. While data fitting will work in any order, output to 2D images will mess up the data ordering and lead to messed up images.

Changing the way how those classes are initialzed however, would be a massive API break that will cause chaos for anyone who already has working code. So, instead, this PR adds two class methods to create `DataIMG/DataIMGInt` instances from a 2D array with `x0`, `x1` arrays or with a WCS. These new methods will check consisteny of the dimensions, flatten the arrys in the correct way, and do not take dublicated information which could potenitally be inconsistent (e.g. when setting up an image with a WCS, the logical coordinate axes are autogenerated from 1..n as required in the FITS convention).

In princple, creation from an array could also be added to `Data2D/Data2DInt`, but the current class hierachy (`DataIMGInt` inherits from `DataIMG`, which inherits from `Data2D`, but `Data2DInt` is not an ancestor of `DataIMGInt`) and the fact the methods like `is_integrated` that exist for 1D data are not consistenly available in the 2D data classes means that a lot of the code would have to be dublicated in different places. I have ideas on how to improve on that by changing the ordering of inheritance, but that is beyond the scope of this PR.

We do however encourage users to use the `DataIMG/DataIMGInt` classes for images (i.e. gridded data), so adding the methods there should be good enough for most use cases.

### DIFF
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -139,7 +139,7 @@ import numpy as np
 import numpy.typing as npt
 
 from sherpa.astro import hc
-from sherpa.data import Data1DInt, Data2D, Data2DInt, Data, Data1D, \
+from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
     IntegratedDataSpace2D, _check
 from sherpa.models.regrid import EvaluationSpace1D
 from sherpa.stats import Chi2XspecVar


sherpa/astro/data.py
```diff
@@ -123,13 +123,14 @@
 from collections.abc import Sequence
 import logging
 import os
-from typing import Any, Callable, Literal, Mapping, cast, overload
+from typing import Any, Callable, Literal, Mapping, cast, overload, TYPE_CHECKING
```
DougBurke: This is not a full review.

We should change line 123 to 

```python
from collections.abc import Callable, Mapping, Sequence
```

rather than import the `Callable` and `Mapping` symbols from `typing`

I'm 100% sure about `Callable` and slightly less sure about `Mapping` ....

In some of the code below you accept a `dict | None` and perhaps that should be `Mapping | None` or `Mapping[str, Any] | None` ?

DougBurke:
Does this address

- #1446 

(at least for `DataIMG`)?

hamogu:
It does address #1446, though maybe not in exactly the way envisioned in that issue. (I didn't know about #1146)
There is a question if we want to overload the `__init__` to automatically detect if we are given 2D arrays instead of having the extra class methods I add here.
I didn't do that because it seems too magical to accept too many different input with the same name and also because having new methods gives a clean way to NOT have certain parameters (instead of accepting all and then raising errors of "If this input is 2D but that parameter is also set, then raise error..."

hamogu:
@DougBurke I think I addressed all points you raised, can you check?

hamogu:
rebased to resolve conflict from recently merged PR.

sherpa/astro/data.py
```diff
@@ -5390,6 +5428,112 @@ def __init__(self, name, x0, x1, y, shape=None, staterror=None,
         self._x0label = None
         self._x1label = None
 
+
+    # This could be moved up Data2D, but the inheritance pathway of the
+    # DataIMG classes makes that a bit awkward, since DataIMG and DataIMGInt
+    # both inherit from Data2D, but DataIMGInt does not inherit from Data2dInt.
```
```suggestion
    # both inherit from Data2D, but DataIMGInt does not inherit from Data2DInt.
```

sherpa/astro/data.py
```diff
@@ -5390,6 +5428,112 @@ def __init__(self, name, x0, x1, y, shape=None, staterror=None,
...
+    @classmethod
+    def from_2d_array(cls,
+                     name: str,
+                     *,
+                     y: npt.NDArray[np.floating],
+                     x0: npt.NDArray[np.floating] | None = None,
+                     x1: npt.NDArray[np.floating] | None = None,
+                     staterror: npt.NDArray[np.floating] | None = None,
+                     syserror: npt.NDArray[np.floating] | None = None,
+                     header: Mapping[str, Any] | None = None) -> "DataIMG":   # -> Self  -- but can't so that al long as we are still 3.10 compatible
```
```suggestion
                     header: Mapping[str, Any] | None = None
                     ) -> "DataIMG":   # change to Self once Python 3.10 support has been dropped
```

sherpa/astro/data.py
```diff
@@ -5390,6 +5428,112 @@ def __init__(self, name, x0, x1, y, shape=None, staterror=None,
...
+        '''Create a `Data2IMG` instance from a 2-dimensional array.
```
```suggestion
        '''Create a `DataIMG` instance from a 2-dimensional array.
```

sherpa/astro/data.py
```diff
@@ -5390,6 +5428,112 @@ def __init__(self, name, x0, x1, y, shape=None, staterror=None,
...
+    def from_2d_array_with_wcs(cls,
+                     name: str,
+                     *,
+                     y: npt.NDArray[np.floating],
+                     staterror: npt.NDArray[np.floating] | None = None,
+                     syserror: npt.NDArray[np.floating] | None = None,
+                     sky: WCS,
+                     eqpos: WCS,
+                     coord: Literal["logical", "image", "physical", "world", "wcs"] = 'logical',
+                     header: Mapping[str, Any] | None = None) -> "DataIMG":   # -> Self  -- but can't so that al long as we are still 3.10 compatible
```
```suggestion
                     header: Mapping[str, Any] | None = None
                     ) -> "DataIMG":   # change to Self once Python 3.10 support has been dropped
```

sherpa/astro/data.py
```diff
@@ -5941,6 +6092,122 @@ def __init__(self, name, x0lo, x1lo, x0hi, x1hi, y, shape=None,
...
+                     *,
+                     y: npt.NDArray[np.floating],
+                     # The superclass has x0, x1, but here we want
+                     # to be explicit that we are using the bounds
+                     x0_bounds: npt.NDArray[np.floating] | None = None,
+                     x1_bounds: npt.NDArray[np.floating] | None = None,
+                     staterror: npt.NDArray[np.floating] | None = None,
+                     syserror: npt.NDArray[np.floating] | None = None,
+                     header: Mapping[str, Any] | None = None,
+                     ) -> "DataIMGInt":  # -> Self:  -- but can't so that al long as we are still 3.10 compatible
```
```suggestion
                     ) -> "DataIMGInt":   # change to Self once Python 3.10 support has been dropped
```

sherpa/astro/data.py
```diff
@@ -5941,6 +6092,122 @@ def __init__(self, name, x0lo, x1lo, x0hi, x1hi, y, shape=None,
...
+    def from_2d_array_with_wcs(cls,
+                     name: str,
+                     *,
+                     y: npt.NDArray[np.floating],
+                     staterror: npt.NDArray[np.floating] | None = None,
+                     syserror: npt.NDArray[np.floating] | None = None,
+                     sky: WCS,
+                     eqpos: WCS,
+                     coord: Literal["logical", "image", "physical", "world", "wcs"] = 'logical',
+                     header: Mapping[str, Any] | None = None) -> "DataIMGInt":
```
```suggestion
                     header: Mapping[str, Any] | None = None
                     ) -> "DataIMGInt":   # change to Self once Python 3.10 support has been dropped
```

sherpa/astro/data.py
```diff
@@ -129,16 +129,17 @@
...
-from typing import Any, Callable, Literal, Mapping, cast, overload
+from typing import Any, Literal, cast, overload, TYPE_CHECKING
 import warnings
 
 import numpy as np
+import numpy.typing as npt
 
 from sherpa.astro import hc
-from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
+from sherpa.data import Data1DInt, Data2D, Data2DInt, Data, Data1D, \
```
```suggestion
from sherpa.data import Data1DInt, Data2D, Data, Data1D, \
```

Turns out this symbol is unused, so no need to import it.

PR #2356, opened 2025-08-29
Update Python version support: drop 3.10 and add 3.14

DougBurke:
# Summary

Drop support for building with Python 3.10 and add support for building with Python 3.14.

# Details

At the moment there is only one change that we take advantage of, that we do not need to install the `tomli` package when building Sherpa. However, Python 3.10 is well into it's [only security fixes](https://devguide.python.org/versions/) phase, and we have historically only focused on supporting three versions. We could retain the 3.10 support but only test with 3.11 and above.

One of the CI runs is changed to build with Python 3.14 (one of the "minimal test" cases), to act as a canary ni case there are Python 3.14 issues, but so far it looks good for advertising 3.14 support.

The CI runs have been updated to use Python 3.11 to 3.14 but the choices made, in particular in the choice of versions of related code, such as NumPy, could do with a review.

There are no actual changes to code to add Python 3.11-only features in this PR.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -53,6 +53,7 @@ echo "dependencies: ${MATPLOTLIB} ${BOKEH} ${NUMPY} ${XSPEC} ${FITSBUILD}"
 echo "compilers:    ${compilers}"
 
 # Create and activate conda build environment
-conda create --yes -n build python"=${PYTHONVER}.*=*cpython*" pip ${MATPLOTLIB} ${BOKEH} ${NUMPY} ${XSPEC} ${FITSBUILD} ${compilers}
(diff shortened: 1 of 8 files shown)


hamogu:
I'm in favor of fully dropping 3.10, not just say "we don't test against it any more". We've been working on adding typing (which has successfully uncovered bugs and problems in the past) and the typing system is still improving from Python version to Python version. For example in 3.11, we can used "typing.Self", e.g. for class methods that generate a instance of the class (https://github.com/sherpa/sherpa/pull/2355/files#diff-39d64572ec8588c14cccff43dc2d85801af7524041351c8434f80963e847204bR5438).

DougBurke:
Rebasing.

DougBurke:
Rebasing to address conflicts and because #2457 was merged.

hamogu (approved):

PR #2358, opened 2025-09-01
Remove bin lohi redux

DougBurke:


sherpa/astro/data.py
```diff
@@ -1913,28 +1915,37 @@ def mask(self, val: ArrayType | bool) -> None:
...
-        return self._bin_lo
+        return None
 
     @bin_lo.setter
     def bin_lo(self, val: ArrayType | None) -> None:
-        self._set_related("bin_lo", val)
+        if val is None:
+            return
+
+        warnings.warn("The bin_lo/hi fields are no-longer used: a diagonal RMF can badded to represent the mapping.", FutureWarning)
```
```suggestion
        warnings.warn("The bin_lo/hi fields are no-longer used: a diagonal RMF can be added to represent the mapping.", FutureWarning)
```

sherpa/astro/io/tests/test_io_response.py
```diff
@@ -245,14 +238,14 @@ def test_write_arf_ascii(make_data_path, tmp_path):
 @requires_data
 @requires_fits
 def test_write_arf_fits_chandra_acis_hetg(make_data_path, tmp_path):
-    """Check we can write out an ARF with BIN_LO/HI as a FITS file.
+    """Check we can write out a Chandra ACIS grating ARF as a FITS file.
+
+    This used to write out BIN_LO/HI values too, until #1564 was fixed.
     """
```
hamogu: No need to dwell on the past. I cna use "git blame" if you want to find out who changed it last and other git tools if I want the full history. In this particular case, I don't think there is much benefit of knowing what it used to do since the bin_lo/hi are removed cleanly. Let's keep comments like this for cases where the history matters for the current design.
```suggestion
    """Check we can write out a Chandra ACIS grating ARF as a FITS file.
    """
```

sherpa/astro/io/tests/test_io_response.py
```diff
@@ -283,14 +276,13 @@ def test_write_arf_fits_chandra_acis_hetg(make_data_path, tmp_path):
...
-
 
 @requires_data
 @requires_fits
 def test_write_arf_ascii_chandra_acis_hetg(make_data_path, tmp_path):
-    """Check we can write out an ARF with BIN_LO/HI as an ASCII file.
+    """Check we can write out a Chandra ACIS grating ARF as an ASCII file.
+
+    This used to write out BIN_LO/HI values too, until #1564 was fixed.
     """
```
hamogu:
```suggestion
    """
```

PR #2359, opened 2025-09-02
Reorganize class hierarchy for Dataclasses and EvaluationSpaces

hamogu:
## Summary
Reorganize hirachy of evaluation spaces and data classes so that they all have a common ancestor.

## Details
`EvaluationSpaceND` will now be the base of all evaluation spaces and this has a simper property `integrated` so the same class can hold either integrated or non-integrated data.
`EvaluationSpace1D` and `EvaluationSpace2D` are merely specializations of that with a slightly different calling structure (The independend axes are passed individually to the constructor instead of as a tuple).
An evaluation space can be flat (i.e. all axes have the same length such as `x0=[1,1,1,2,2,2]` and `x1=[1,2,3,1,2,3]`)
or not (`x0=[1,2]` and `x1=[1,2,3]`). The former case allows evaluation on a non-rectangular grid.
    
Data spaces derive from Evaluation spaces by adding a filter as a mix-in class. This design allows us to reuse
the entire inheritance tree and basically build a parallel tree by just mixing in the mix-in class in every case.
It also means that we can reproduce our legacy DataXD and DataXDInt (with X=1,2) classes with very little code.

To allow this consistency some previously inconsistent features have been changed or removed:
- remove `for_model` method (which always returned `self` independent of the input parameters), `get_evaluation_indep` (which called `for_model`) and the  keyword `use_evaluation_sapce` in `get_x`, get_y` (which didn't do anything).

## Follow-up
The text in the commit messages has lines like "which we intend to change in a future commit". However, I see that this PR is already quite big and might be hard to review. While the user shouldn't see any change in normal Sherpa use and even advanced users who initialize data classes directly won't have to care what changed underneath, this is a significant change in our most fundamental data classes. 

## Benefits
The benefits of having a more consistent set of evaluation spaces and data classes are not fully exploited in this PR, because it's so big already; instead we'll to follow-up PRs. But broadly speaking, they are

- code clarity, consistency, fewer corner cases (the usual)
- Our plotting code works "per data class". With a more consistent data class layout, we need fewer plotting classes. Users who want to "just get the data out" to do their own plots (a common request) can do so with the same UI for all types of data.
- We didn't have IntegratedNDData at all. Integrated data is kind of Sherpa's specialty and I do have two scientific use cases that absolutely need that.
- Capabilities of DataClasses. This will make it easier to move some of the specialties of `DataPHA` that would be useful in other context to the parent classes. Grouping is useful not just for photons, but for any count data (number of galaxies per mass range, size of flares from young stars, ...). On-the-fly grouping and filtering is on of Sherpa's strong suits and can easily be generalized to any integrated 1D data. That may have been possible in the old paradigm, but it easier here.
- Some unused or useless (meaning that they are called but don't change behavior) functions or arguments are already removed in this PR. More can be removed in future PRs. That simplifies the code, makes it more readable and thus makes it easier to use it as a base for special, derived classes. In fact, that's why I started this PR...

### DIFF
--- a/docs/fit/index.rst
+++ b/docs/fit/index.rst
@@ -119,11 +119,12 @@ unless explicitly over-riden with the ``stat`` and
     method    = LevMar
     estmethod = Covariance
     >>> print(f.data)
-    name      = fit example
-    x         = Int64[6]
-    y         = Float64[6]
-    staterror = Float64[6]
-    syserror  = None
+    name       = fit example
+    x          = Int64[6]
+    y          = Float64[6]
+    staterror  = Float64[6]
+    syserror   = None
+    integrated = False
     >>> print(f.model)
     polynom1d
        Param        Type          Value          Min          Max      Units
--- a/docs/quick.rst
+++ b/docs/quick.rst
@@ -91,11 +91,12 @@ axis, and then dependent axis:
     >>> from sherpa.data import Data1D
     >>> d = Data1D('example', x, y)
     >>> print(d)
-    name      = example
-    x         = Float64[200]
-    y         = Float64[200]
-    staterror = None
-    syserror  = None
+    name       = example
+    x          = Float64[200]
+    y          = Float64[200]
+    staterror  = None
+    syserror   = None
+    integrated = False
 
 At this point no errors are being used in the fit, so the ``staterror``
 and ``syserror`` fields are empty. They can be set either when the
@@ -477,11 +478,12 @@ the same, and equal to the true error:
     >>> dy = np.full(x.size, err_true)
     >>> de = Data1D('with-errors', x, y, staterror=dy)
     >>> print(de)
-    name      = with-errors
-    x         = Float64[200]
-    y         = Float64[200]
-    staterror = Float64[200]
-    syserror  = None
+    name       = with-errors
+    x          = Float64[200]
+    y          = Float64[200]
+    staterror  = Float64[200]
+    syserror   = None
+    integrated = False
 
 The statistic is changed from least squares to
 chi-square (:py:class:`~sherpa.stats.Chi2`), to take advantage
@@ -897,13 +899,14 @@ For example, the following code creates a
     >>> yaxis = y.ravel()
     >>> d2 = Data2D('img', x0axis, x1axis, yaxis, shape=(128, 128))
     >>> print(d2)
-    name      = img
-    x0        = Int64[16384]
-    x1        = Int64[16384]
-    y         = Float64[16384]
-    shape     = (128, 128)
-    staterror = None
-    syserror  = None
+    name       = img
+    x0         = Int64[16384]
+    x1         = Int64[16384]
+    y          = Float64[16384]
+    shape      = (128, 128)
+    staterror  = None
+    syserror   = None
+    integrated = False
 
 Define the model
 ----------------
--- a/sherpa/astro/data.py
+++ b/sherpa/astro/data.py
@@ -5506,7 +5506,8 @@ class DataIMG(Data2D):
         #
         self._orig_indep_axis = (self.coord, x0, x1)
 
-        super().__init__(name, x0, x1, y, shape, staterror, syserror)
+        super().__init__(name, x0, x1, y,
+                         shape=shape, staterror=staterror, syserror=syserror)
 
         # special case the x-axis labels
         self._x0label = None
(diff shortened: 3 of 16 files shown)


hamogu:
Rebased. This is already long (even though many lines are just moved, there are number of deep structural changes that warrant thinking about in review). I do want to point out that this actually shortens the code despite adding new docs for classes and functions because it combines duplicate implementations (DRY principle). 
I admit that there is some tension between really keeping evaluation spaces and data spaces separate because they are conceptually different, vs. not reimplementing things. I believe that this PR strikes a better balance than the existing code and in particular it removes our bespoke structure where e.g. the "integrated" was implemented separately (and inconsistently) for 1D and 2D but not at all for 3D.

This PR doesn't really play out the benefits of the new structure but I don't want to make it even longer and I do think it's already useful as is.
I will open follow-up PRs that really get into ND fitting, explore how to make use if this to simplify plotting, and remove more keywords to simplify stuff, but I'll back of until we have an initial review and settle on the general directions this is going to avoid having too many things to rebase on.

hamogu:
The tests pass locally and would do here if I hadn't used Python 3.11 typing syntax. Please review anyway. I can just not type that specific return if we want to stay 3.10 compatible, but I do think we probably want to drop 3.10 sooner rather than later anyway. In that case, I can also rebase on a PR that drops 3.10.

hamogu:
Rebased to pick up the fast the main has moved passed 3.10 now, which will hopefully make CI pass so that we can eventually get this in.

DougBurke:
I restarted the failed conda runs since this looked to be a transient failure unrelated to sherpa. I don't see why the RTD build failed (it mentioned a warning but then I couldn't find the warning ...).

hamogu:
I will investigate the RTD build.

DougBurke:
I *think* the RTD failure is a separate issue. At least I'm being dragged across the hot coals of plot directives trying to understand it, but if you can find out what's going on then all's ++good

hamogu:
I think it might be related. RTD pick up a reference to "shape" and can't resolve it. With this PR, there are a lot more objects with a documented shape, giving Sphinx more choices.

hamogu:
```
generating indices... /home/docs/checkouts/readthedocs.org/user_builds/sherpa/checkouts/2359/docs/mcmc/api/sherpa.sim.sample.multivariate_t.rst:4: WARNING: more than one target found for cross-reference 'shape': sherpa.astro.data.DataARF.shape, sherpa.astro.data.DataIMG.shape, sherpa.astro.data.DataIMGInt.shape, sherpa.astro.data.DataOgipResponse.shape, sherpa.astro.data.DataPHA.shape, sherpa.astro.data.DataRMF.shape, sherpa.astro.data.DataRosatRMF.shape, sherpa.data.Data.shape, sherpa.data.Data1D.shape, sherpa.data.Data1DAsymmetricErrs.shape, sherpa.data.Data1DInt.shape, sherpa.data.Data2D.shape, sherpa.data.Data2DInt.shape, sherpa.data.DataSpace1D.shape, sherpa.data.DataSpace2D.shape, sherpa.data.DataSpaceND.shape, sherpa.data.IntegratedDataSpace1D.shape, sherpa.data.IntegratedDataSpace2D.shape, sherpa.models.regrid.EvaluationSpace1D.shape, sherpa.models.regrid.EvaluationSpace2D.shape [ref.python]
```

The problem is that I'm having other, unrelated Sphinx build problems locally, so debugging this is painful, but I'm sure I'll find it eventually.

DougBurke:
I'm getting lots of sphinx errors from `docs/evaluation/simulate.rst` because the code can't seem to find the data directory for some reason I can't diagnose, if that's any help ;-)

hamogu:
OK, I found it. Passes locally, so I expect it to pass here in a few minutes.

DougBurke:
@hamogu - I think this conflict should be easy to resolve (not sure why they are thought of as a conflict...)

hamogu:
I know. I did that locally, I just haven'd pushed again because I think it's not going to make it into 4.19, so I was going to wait till after.

hamogu:
This is the PR I referred to in the sherpa-dev email: We absolutely need this to make reasonable progress on making ND fitting better, but by and of itself, this PR doesn't help users, so I don't think there is a good reason to push it through in the next few days. Better to concentrate on the things we do want to get in and that users will immediately notice, like XPSEC version updates.

hamogu:
@DougBurke As I said above, I don't think we have other rush this into 4.19, but if you've time to review now or after the freeze that would be great. I also don't want to let this linger so long again that I get merge conflicts again.

DougBurke:
I hope to have some time at the week-end.

PR #2362, opened 2025-09-06
Move caching tests into dedicated file

### LINKED ISSUE #2277: Split up `sherpa/models/tests/test_model.py`
Working on the caching in a number of recent PRs, I find the file size of `sherpa/models/tests/test_model.py` unwieldy and I'd like to separate out all the caching tests into a new `test_caching.py` just to make it easier to scroll around in an editor and see what tests we have and what they do.

That's literally 3 min of work, I just don't want to do that right now because there are are several PRs open that touch that file and it would be a pain to rebase all the time. So, I'm opening an issue so I don't forget to do that as a separate PR when the dust has settled on the other things.

hamogu:
The `test_model.py` simply had gotten so big that it was hard to read in an editor, so moved some files over.

Fixes #2277

### DIFF
--- a/sherpa/models/tests/test_caching.py
+++ b/sherpa/models/tests/test_caching.py
@@ -312,7 +312,7 @@ def test_evaluate_cache1d():
 def test_cache_is_actually_used():
     """Most other tests check that the cache has values in it,
     but not that those values are actually returned.
-    Here, we manipulte the cached value and then call the model
+    Here, we manipulate the cached value and then call the model
     to check that the cached value is used.
     """
     xgrid = np.arange(2, 10, 1.5)


sherpa/models/tests/test_caching.py
```diff
@@ -188,3 +195,814 @@ def test_evaluate_cache2dint():
     mdl._cache[list(mdl._cache.keys())[0]] = 2 * expected
     assert mdl(x0lo, x1lo, x0hi, x1hi) == pytest.approx(2 * expected)
 
+
```
DougBurke, line 198:
```suggestion
```

DougBurke: That's assuming I can count.

sherpa/models/tests/test_caching.py
```diff
@@ -188,3 +195,813 @@ def test_evaluate_cache2dint():
...
+    expected = 5 + 2 * xgrid
+    assert mdl(xgrid) == pytest.approx(expected)
+    # Now, the model has run a second time, so there are two entries in the cache
+    check_cache(mdl, expected, xgrid, cache_size=2)
+
+
+def test_cache_is_actually_used():
+    """Most other tests check that the cache has values in it,
+    but not that those values are actually returned.
+    Here, we manipulte the cached value and then call the model
```
wmclaugh:
```suggestion
    Here, we manipulate the cached value and then call the model
```

PR #2363, opened 2025-09-06
conftest: ensure data path is absolute not relative

### LINKED ISSUE #2163: sherpa-test-directory, when set with `-D`, should be a full path
In testing out #2096 I tried

    % pip install .
    % sherpa_test -n auto -D sherpa-test-data/sherpatest

and a small number of tests fail. This is because the tests assume that the data directory is a full, and not local, path, and so checks for a screen message to start with "blah blah blah  /" when, with the above, you get "blah blah blah  sherpa-test-data...".

For example

```
    def record_starts_with(r, start, level="INFO"):
        """This only checks the start of the message, as it contains
        test-specific information (the location of the file name).
        """
    
        assert r.name == "sherpa.astro.io"
        assert r.levelname == level
    
        # We use this, rather than msg.startswith(start), as the error
        # message is more useful if there is a difference.
        #
        nchar = len(start)
        msg = r.getMessage()
>       assert msg[:nchar] == start
E       AssertionError: assert 'read ARF file s' == 'read ARF file /'
E         
E         - read ARF file /
E         ?               ^
E         + read ARF file s
E         ?               ^

../../micromamba/envs/sherpa-2/lib/python3.11/site-packages/sherpa/astro/tests/test_astro_data_chandra_unit.py:442: AssertionError
```

We could change the tests or change the test code so that the path always gets converted to a full path.

Either way I don't see it as a high priority.

DougBurke:
# Summary

Ensure that the path for the Sherpa test data is an absolute and not relative path to improve testing. Fix #2163.

### DIFF
--- a/sherpa/conftest.py
+++ b/sherpa/conftest.py
@@ -22,6 +22,7 @@ from importlib.resources import files
 import os
 import re
 import logging
+from pathlib import Path
 import time
 
 import numpy as np
@@ -342,7 +343,7 @@ def pytest_configure(config):
     try:
         path = config.getoption(TEST_DATA_OPTION)
         if path:
-            set_datadir(path)
+            set_datadir(str(Path(path).resolve()))
     except ValueError:  # option not defined from command line, no-op
         pass
 


DougBurke:
I'll simplify this since it will conflict with #2345

PR #2366, opened 2025-09-08
Add a  convenience method to set the IO backend and a context manager to set the IO backend temporarily

### LINKED ISSUE #2101: add a way to temporarily change the IO backend
Taken from https://github.com/sherpa/sherpa/issues/1865#issuecomment-1718111209

A request to add something like `TemporaryIOBackend` - this may be hard with the current module-rather-than-class approach.

hamogu:
## Summary
Add a  convenience method to set the IO backend and a context manager to set the IO backend temporarily.

## Details 

 This is implemented analogously to the plotting backend, but differs in the details because the IO backend is a module, not a class.
    
This functionality is probably mostly useful for developers; users generally don't change the backend once it works since they all function identically (unless there is a bug). This is in contrast to the plotting backends, where a user might hae good reasons to change between interactive and statics plots within one session. Thus, the docs for the new feature are more muted (it's just in the module docstring and the docstrings for the functions, not in the narrative docs).

As part if this PR, we also improve the related doctests. 

Fixes #2101

### DIFF
--- a/sherpa/astro/io/tests/test_backend_switch.py
+++ b/sherpa/astro/io/tests/test_backend_switch.py
@@ -79,7 +79,7 @@ def test_set_io_backend(make_data_path, clean_astro_ui):
     try:
         io.set_io_backend('astropy')
     except IdentifierErr:
-        io.set_io_backend('pycrates')
+        io.set_io_backend('crates')
 
     infile = make_data_path("9774.pi")
     ui.load_pha(infile)


sherpa/astro/io/__init__.py
```diff
@@ -32,32 +32,46 @@
...
+  >>> old_backend = io.backend
 
-  >>> import sherpa.astro.io.pyfits_backend
-  >>> io.backend = sherpa.astro.io.pyfits_backend
+  >>> from sherpa.astro.io import set_io_backend
+  >>> set_io_backend('pyfits')
+
+.. testcleanup::
+
+  >>> io.backend = old_backend
```
DougBurke, line 61: can this be

```
>>> set_io_backend(old_backend)
```

or does `set_io_backend` require a string argument (I could jump down to check but I'm too lazy)?

sherpa/astro/io/__init__.py
```diff
@@ -116,30 +144,142 @@
...
 else:
-    # None of the options in the rc file work, e.g. because it's an
-    # old file that does not have dummy listed
-    import sherpa.astro.io.dummy_backend as backend
+    backend = IO_BACKENDS['dummy']
 
+if len(IO_BACKENDS) == 1 and IO_BACKENDS.keys() == {'dummy'}:
+    warning("""Cannot import usable I/O backend.
+    If you are using CIAO, this is most likely an error and you should contact the CIAO helpdesk.
+    If you are using Standalone Sherpa, please install astropy.""")
```
DougBurke, line 183: Thinking aloud here.

This message used to be raised when importing `sherpa.astro.io.dummy_backend` but now it's done here so switching back to the `dummy` backend will not annoy the user with repeated warnings.

hamogu: I think raising it once is enough - you still get an error if you use it. The reason to move it is that I'm now importing all backend as early as possible to populate `IO_BACKENDS`. So, if this was raised on the first import of `dummy` then it would be raised every time because `dummy` will always be present, even if astropy and pycrates are also present.

(As an aside: That means there is a small penalty in import time compared to the previous version because we now import `dummy` in addition to the backend that will be selected, but `dummy` is a single pure-python file so I consider the impact negligible compared to the user convenience of having a list in `IO_BACKENDS` and the ability to switch easily with na session is needed).

sherpa/astro/io/__init__.py
hamogu: It does require a string. For the plotting backend, if and be a string, or a class, or an instance, but here we have a module-based system (which I can't want to change for this simple PR because the benefit of one backend inheriting for another is not there unlike for the plotting backends) and in practice nobody uses "module objects". Of course, everything is an object in Python, so it's possible, but it's not practical. 
`set_io_backend(old_backend.name)` would work but it's not simpler and this is just for a test cleanup. No user will over see it, it's just so that running this test doesn't change the global state and influences other tests that happen to be run later.

sherpa/astro/io/tests/test_backend_switch.py
```diff
@@ -56,6 +57,70 @@ def test_backend_switch(make_data_path):
...
+
+    with pytest.raises(NotImplementedError,
+                       match="No usable I/O backend was imported."):
+         ui.load_pha(pha)
+
+    # Make this test work independent of the fits backend
+    try:
+        io.set_io_backend('astropy')
+    except IdentifierErr:
+        io.set_io_backend('pycrates')
```
DougBurke:
```suggestion
        io.set_io_backend('crates')
```

It's `crates_backend.py` not `pycrates_backend.py`

sherpa/astro/io/xstable.py
```diff
@@ -115,6 +115,11 @@
    mygauss.pos  thawed            1            0          2.4
    mygauss.norm thawed            1            0        1e+24
 
+.. testcleanup::
+
+   >>> import os
+   >>> os.remove("example.mod")
```
DougBurke, line 121: I have wondered where/why some of these files keep popping up in my tree during development! (there's another recent PR with a similar change).

So thanks for the fixes.

DougBurke (approved):

PR #2370, opened 2025-09-09
Update docs still claiming there is only one plotting backend

hamogu:
There is no change to the code.

[skip ci]


docs/plots/index.rst
```diff
@@ -11,9 +11,8 @@ Visualization
 Overview
 ========
 
-Sherpa has support for different plot backends, although
-at present there is only one, which uses the
-:term:`matplotlib` package.
+Sherpa has support for different plot backends, with the default
+:term:`matplotlib` package, see `_Plotting backends_` for details.
```
DougBurke: This text should be improved, but I just wanted to mention that this does not create a link to the backends doc. Instead we get 

image

and the text is not a link.

DougBurke: Text wise, something like (but I'm not particularly happy with it)

```
Sherpa has support for different plot backends, with the default
being the :term:`matplotlib` package. For more information see
:ref:`the backend documentation <Plotting backends>``.
```

(I'm guessing that link will work).

hamogu:
I think I addressed that (and the typos that made the link not work). I've chosen to give a bit more of a shout-out to bokeh because it's (a) new and (b) requires a bit more setup than matplotlib to actually get the plot out. I think at this level (in the "visualization" docs for standalone users)that's justified; I would not push it as much within CIAO but this is addressed to a different audience.

PR #2371, opened 2025-09-09
Pass backend-specific options through to pylab and bokeh

### LINKED ISSUE #2084: can we add fillstyle to the plot options?
This is rather matplotlib-centric, but

```
sherpa In [68]: get_plot_prefs("data")["fillstyle"] = "none"

sherpa In [69]: plot_data()
TypeError: PylabBackend.plot() got an unexpected keyword argument 'fillstyle'
```

vs

```
sherpa In [73]: plt.clf()

sherpa In [74]: plt.plot([1, 2, 3], [4, 2, 17], marker="o", fillstyle="none")
Out[74]: [<matplotlib.lines.Line2D at 0x73b27e018d90>]
```

sherpa

hamogu:
## Summary
Pass backend-specific options through to pylab and bokeh

## Details
Sherpa plotting commands should be passing through keywords that are not part of the sherpa-specific interface to the plotting library that is used as backend underneath.
This was one of the main design goals of the plotting work done for 4.17, and there is even a test for it, but as it turns out that test happens to use the wrong keyword (one that is part of the sherpa-specific set) so we never caught that it was, inf fact, NOT passing through stuff.

Two related small changes are included here:
- remove a print statement that was left over from some debugging
- There is an ambiguity on how to match general (sherpa API) keywords to backend keywords when it's not a 1:1 ratio. For example, if alpha is set, should it apply to the markers, the lines, the hatches or all of them? Hatches are invisible when they are always exactly identical in color, alpha etc to the filling color of a symbol, so we are no longer setting the hatch properties based on the general input and leave it to the user to pass through hatch specific keywords.

fixes #2084

This should have been part if 4.17, so it would be great to get it into 4.18 at least.

### DIFF
--- a/sherpa/plot/bokeh_backend.py
+++ b/sherpa/plot/bokeh_backend.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2023, 2024
+#  Copyright (C) 2023-2025
 #  MIT
 #
 #
@@ -286,6 +286,7 @@ class BokehBackend(BasicBackend):
               label=None,
               linewidth=None,
               linecolor=None,
+              **kwargs
               ):
         """Draw histogram data.
 
@@ -326,7 +327,8 @@ class BokehBackend(BasicBackend):
                 linewidth=linewidth,
                 drawstyle=drawstyle,
                 color=color, marker=None, alpha=alpha,
-                xaxis=False, ratioline=False)
+                xaxis=False, ratioline=False,
+                **kwargs)
 
         # Draw points and error bars at the mid-point of the bins.
         # Use the color used for the data plot: should it
@@ -358,31 +360,9 @@ class BokehBackend(BasicBackend):
                           markerfacecolor=markerfacecolor,
                           ecolor=ecolor,
                           capsize=capsize,
+                          **kwargs
                          )
 
-    # Note that this is an internal method that is not wrapped in
-    # @ translate_args
-    # This is called from methods like plot, histo, etc.
-    # so the arguments passed to this method are already translated and we do
-    # not want to apply the translation a second time.
-    def _set_line(self, line, linecolor=None, linestyle=None, linewidth=None):
-        """Apply the line attributes, if set.
-
-        Parameters
-        ----------
-        line :
-            The line to change
-        linecolor, linestyle, linewidth : optional
-            The attribute value or None.
-        """
-        if linecolor is not None:
-            line.glyph.line_color = linecolor
-
-        if linestyle is not None:
-            line.glyph.line_dash = linestyle
-
-        if linewidth is not None:
-            line.glyph.line_width = linewidth
 
     # There is no support for alpha in the Plot.vline class
     @add_kwargs_to_doc(kwargs_doc)
@@ -392,7 +372,9 @@ class BokehBackend(BasicBackend):
               linecolor=None,
               linestyle=None,
               linewidth=None,
-              overplot=False, clearwindow=True):
+              overplot=False, clearwindow=True,
+              **kwargs
+              ):
         """Draw a vertical line
 
         Parameters
@@ -406,7 +388,9 @@ class BokehBackend(BasicBackend):
         line = Span(location=x, dimension='height',
                     line_width=linewidth,
                     line_color=linecolor,
-                    line_dash=linestyle)
+                    line_dash=linestyle,
+                    **kwargs
+                    )
         axes.add_layout(line)
 
     # There is no support for alpha in the Plot.hline class
@@ -418,7 +402,9 @@ class BokehBackend(BasicBackend):
               linecolor=None,
               linestyle=None,
               linewidth=None,
-              overplot=False, clearwindow=True):
+              overplot=False, clearwindow=True,
+              **kwargs
+              ):
         """Draw a horizontal line
 
         Parameters
@@ -432,7 +418,9 @@ class BokehBackend(BasicBackend):
         line = Span(location=y, dimension='width',
                     line_width=linewidth,
                     line_color=linecolor,
-                    line_dash=linestyle)
+                    line_dash=linestyle,
+                    **kwargs
+                    )
         axes.add_layout(line)
 
     @add_kwargs_to_doc(kwargs_doc)
@@ -458,6 +446,7 @@ class BokehBackend(BasicBackend):
              linewidth=1,
              linecolor=None,
              xaxis=None, ratioline=None,
+             **kwargs
              ):
         """Draw x, y data.
 
@@ -503,7 +492,6 @@ class BokehBackend(BasicBackend):
 
         objs = []
 
-        kwargs = {}
         if label is not None:
             kwargs['legend_label'] = label
         if linestyle != 'noline':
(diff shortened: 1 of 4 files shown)


sherpa/plot/tests/test_bokeh.py
```diff
@@ -53,3 +58,59 @@ def test_bokeh_delchi(caplog, clean_ui):
...
+
+    """
+    bokehbackend = pytest.importorskip("sherpa.plot.bokeh_backend")
+    # previous line guarantees that bokeh is installed
+    from bokeh.embed import file_html
+    from bokeh.resources import CDN
+
+    data = Data1DInt('tst', np.asarray([1, 2, 3]), np.asarray([1, 2, 3]) + 1,
+                      np.asarray([10, 12, 10.5]))
+    pl = DataPlot()
```
DougBurke, line 76:
```suggestion
    pl = DataHistogramPlot()
```

and then also importing it above.

The naming of our plot classes, and what they are meant to support, is lost to the mist of time. 

We also allow users to populate `DataPlot` with a "integrated" data set like `Data1DInt` so maybe this change is not worth it, but I thought I'd check.

sherpa/plot/tests/test_pylab.py
```diff
@@ -110,3 +110,94 @@ def test_warning_plot_hist_linecolor(caplog, requires_pylab):
...
+    out means it was passed through.
+    `test_backend_option_does_not_exist` ensures that.
+    """
+    from matplotlib import pyplot as plt
+
+    edges = np.asarray([-10, -5, 5, 12, 17, 20, 30, 56, 60])
+    y = np.asarray([28, 62, 17, 4, 2, 4, 125, 55])
+
+    d = Data1DInt('example histogram', edges[:-1], edges[1:], y)
+    from sherpa.plot import DataPlot
```
DougBurke: So again should this be `DataHistogramPlot` or does it not matter? Or is using `DataPlot` required for this test since you also create an explicit `Histogram` [*]?

Also can this import be added to the top of the file (purely a style thing, but it makes it easier to see what parts of sherpa are being used in a test).

[*] I do not want to go on about our existing plot class hierarchy here as that's more for #2076

sherpa/plot/tests/test_pylab.py
```diff
@@ -110,3 +110,94 @@ def test_warning_plot_hist_linecolor(caplog, requires_pylab):
...
+    """Test a matplotlib-specific option (linewidths) for contour plots"""
+
+    from matplotlib import pyplot as plt
+
+    x1, x0 = np.mgrid[20:30, 5:20]
+    y = np.sqrt((x0 - 10)**2 + (x1 - 31)**2)
+    contour = Contour()
+    contour.contour(x0, x1, y, vmax=2, vmin=2)
+    # Not testing the actual colors of the contour levels,
+    # but no error was raised, so it worked.
```
DougBurke, line 200: I'm a little-bit confused here since the test docstring suggests it's testing out `linewidth`, this comment suggests it's something to do with `color`, but the actual `contour` call sets no options. So what is actually being tested here?

sherpa/plot/tests/test_bokeh.py
```diff
@@ -53,3 +58,59 @@ def test_bokeh_delchi(caplog, clean_ui):
...
+    newback = bokehbackend.BokehBackend()
+    with plot.TemporaryPlottingBackend(newback):
+        splot.plot(xlog=True, hatch_pattern='/', markersize=20, alpha=0.5)
+        html = file_html(plot.backend.current_fig, CDN, "my plot")
+
+    assert '"hatch_pattern":{"type":"value","value":"/"}' in html
+
+
+def test_warning_plot_linecolor():
+    """We get a warning when using linecolor: DataPlot"""
```
DougBurke:
```suggestion
    """We get a warning when using an  unknown argument: DataPlot"""
```

sherpa/plot/tests/test_bokeh.py
hamogu: Either will work for this test. The point is that the plot commands are passed through to `BokehBackend.plot` and that will happen with both a `DataPlot` and the `DataHistogramPlot`. I'm using `DataPlot` and `ScatterPlot` for bokeh and `DataPlot`, `ModelHistrogramPLot` and `Contour` for matplotlib, just to get some variety without writing a test for every plot class we have (since they all pass it through in the same way to the same underlying function). 
I could switch one of the `DataPlot` uses to a `DataHistrogramPlot` for good measure, but it doesn't really matter.

sherpa/plot/tests/test_pylab.py
hamogu: Oh, here I'm testing `Contour.contour` with two matplotlib specific setting (`vmin` and `vmax`). Those setting change the limits used for the assignment of colors in the contours shown and it's a bit cumbersome to test the actual color assigned (and I suspect also brittle if matplotlib changes the details of the contouring algorithm). However, we test (in a different test) that matplolib raises an error if it is passed an invalid keyword name, so here I just assume that the fact that `vmin` and `vmax` don't raise an error mean that they are passed through to matplotlib (and if you do this test by hand and look at he result, you will see that in fact all contour lines are assigned the same color because the minimum and the maximum of the color scale is the same).
ddba7955-8ab8-482d-ac5c-a37ee4a2e6e6

hamogu: As for the comment: I tried to rush that through before I left the office yesterday. I copied and pasted the boilerplate between the tests, and apparently didn't fix up the docstrings.

sherpa/plot/tests/test_bokeh.py
DougBurke: Just wanted to check.

sherpa/plot/bokeh_backend.py
```diff
@@ -358,31 +360,9 @@ def histo(self, xlo, xhi, y, *,
                           markerfacecolor=markerfacecolor,
                           ecolor=ecolor,
                           capsize=capsize,
+                          **kwargs
                          )
 
-    # Note that this is an internal method that is not wrapped in
```
DougBurke, line 363: How come this method was removed? I assume that, as the tests pass, it just wasn't used and had been leftover from some development or refactoring when writing the bokeh code?

hamogu: The bokeh implementation started as a 1:1 translations from the pylab code, which has the same method, but then the bokeh code never used it because the parameters are set in a different way and its doesn't make sense to refactor out setting the line parameters in the same way it did for matplotlib. So, this method was never used in production, we just didn't notice.

DougBurke: Ta - just wanted to check.

sherpa/plot/pylab_backend.py
```diff
@@ -364,6 +366,8 @@ def setf(label, newval):
         setf('color', linecolor)
         setf('linestyle', linestyle)
         setf('linewidth', linewidth)
+        for k, v in kwargs.items():
+            setf(k, v)
```
DougBurke, line 370: The coverage reports suggest this line isn't used in any of our tests. This surprises me - don't we send in `markerfacecolor` in at least one instance? - but maybe that's going through a different code path. How easy would it be to check this, or is this line actually used in the tests and it's just not picked up by our coverage checks (I thought I'd fixed this but maybe not)?

hamogu: `_set_line` is only called from `vline` and `hline`, which as this point don't accept keyword arguments. So, I guess I could have left out this change since I cannot be reached, but maybe `vline` and `line` *should* accept kwargs (that's for a future PR and only if we actually have a use case for that, so I'm not opening an issue at this point). I just looked like the right thing to do when I went through the code and generally added `**kwargs` where applicable.

(For the bokeh backend, I just removed the equivalent `_set_line` method in a different PR, because the entire method would never be reached.)

PR #2374, opened 2025-09-12
Always re-calculate covariance matrix when needed (get_draws/eqwidth)

### LINKED ISSUE #556: I think `get_draws` is dependent on the order of the ids for multi-fit datasets; documentation *and* UI issues
This is both a documentation issue, and something to think about for the UI, as it is too easy to do the wrong thing here. See also #464 

I have run

```
>>> covar(9, 10)
>>> d3 = get_draws(9, [10], niter=5000)
>>> d3 = get_draws(10, [9], niter=5000)
>>> d3 = get_draws(9, [10], niter=5000)
```

and, probably not surprising to someone who's looked at the code but likely surprising to the user, the second set of draws is *significantly* different to the first. I show the trace of the statistic from these three runs, and we can see the top and bottom are similar, whereas the middle one (the one where I purposefully changed the order) is "not good"

draws

You can see this in the acceptance fractions - 0.45, 0.03, 0.44.

I assume this is because the order of the parameters comes from the order of the datasets as given by the user, and this is assumed to match the order of the covariance matrix. If so it's too easy for the user to mess this up by accident, particularly in more-complicated scenarios than just using two datasets, as I did above.

### LINKED ISSUE #1788: tracking dataset order for covariance and routines like get_draws
`get_draws` wants covariance to have been run, but there's no check that the covariance run matches the data. At some level this is a "user" problem, but we can do things - like check that the `datasets` field of the covariance `ErrorEstResults` object matches that used by `get_draws`.

There's related issues like #404

### LINKED ISSUE #2379: eqwith and otherids
I don't believe eqwith handles the case of `otherids` being set fully (when `error=True`). We have

https://github.com/sherpa/sherpa/blob/d5a42e66fe6c8a7945e94a879cbd9805e443d27b/sherpa/astro/ui/utils.py#L15520

and I believe that should include `otherids`. Need to verify if there's any other change needed.

DougBurke:
# Summary

Catch cases where the covariance information was out of date (get_draws and eqwidth). Improve the handling of the otherids argument for eqwidth. Fix #556. Fix #1788. Fix #2379.

# Details

There are two uses of `get_covar_results` in the UI code (when `covar_matrix` is not set explicitly): `get_draws` and `eqwidth`. For `get_draws` it assumes that the last call to `covar` is valid for the data (error-ing out if it had not been called), whereas `eqwidth` would call `covar`, but **only** if it had not been called; i.e. it would assume that the stored covariance results match the data.

Now, `get_covar_results` contains enough data to work out if it matches the current fit, but it is surprisingly difficult - e.g.

- parameter values
- choice of thawed parameters (and the ordering)
- set of dataset identifiers used

and, actually, there are changes that it can not catch (e.g. changing the source model).

So, I decided that rather than either error-ing out if the `get_covar_results` call was not valid or printing a warning that the data may not be valid, we should just **always** re-calculate the covariance matrix (and hide the screen output just to avoid confusing the user). The **assumption** here is that the covariance-matrix calculation is not expensive relative to the rest of the routine.

This lets us simplify the code slightly, and should avoid problem cases like

- #556 
- #1788

In looking at the code I realized we had a problem when multiple identifiers are given (that is, when `otherids` is set to a non-empty sequence). This is issue

- #2379 

Fortunately this is easy to fix - as we just need to ensure that the `Fit` object knows about the extra identifiers so that the covariance can be calculated correctly. This is **partly** a regression fix thanks to the earlier changes in this PR (related to `eqwidth` always calculating the covariance now), but it would have still been a problem without the earlier fixes if covariance had not been called (which is how the docstring examples are written for `eqwidth`). In #2379 I had originally thought that I'd need to deal with `DataSimulFit` objects for the multi-obsid case, but fortunately the data object is only needed to calculate the independent axis for the model evaluation, so using just one of the data objects is okay [*].

[*] technically we could have a contrived case where the selected dataset doesn't cover the evaluation space for the model, but this is unlikely and the existing code doesn't deal with this situation anyway, so it's not like this code is making things worse.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_session.py
+++ b/sherpa/astro/ui/tests/test_astro_session.py
@@ -3701,6 +3701,22 @@ def test_fit_ignores_repeated_otherid(session, caplog):
 
 
 @pytest.mark.parametrize("session", [pytest.param(Session, marks=pytest.mark.session), AstroSession])
+def test_get_draws_wants_errors(session):
+    """This test was changed in #2374 but the test name kept as is."""
+
+    s = session()
+    s._add_model_types(sherpa.models.basic)
+
+    s.load_arrays(1, [1, 2], [2, 5])
+    s.set_source(1, "const1d.foo")
+    s.set_stat("cash")
+    s.fit()
+
+    # Prior to #2374 this would error out. Now just check it passes.
+    _ = s.get_draws(niter=1)
+
+
+@pytest.mark.parametrize("session", [pytest.param(Session, marks=pytest.mark.session), AstroSession])
 @pytest.mark.parametrize("method", ["proj", "unc"])
 def test_get_int_xxx_otherids(method, session):
     """Corner case.
--- a/sherpa/astro/ui/tests/test_eqwidth_err.py
+++ b/sherpa/astro/ui/tests/test_eqwidth_err.py
@@ -371,9 +371,7 @@ def test_eqwidth_multi_id_chisq(clean_astro_ui):
 def test_eqwidth_multi_id_poisson(clean_astro_ui):
     """Regression test for handling multiple ids (poisson).
 
-    This is only an issue when error=True. At the moment this is
-    treated as a regression test as it is unclear what we want the
-    results to be.
+    This is only an issue when error=True.
 
     """
 
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -15519,8 +15519,7 @@ class Session(sherpa.ui.utils.Session):
         # evaluation, so just use the "default" version and do not
         # bother about multiple datasets. This could be an issue if
         # the model evaluation should be used on an extension of the
-        # grid in this default dataset, but this is a theoretical
-        # concern.
+        # grid in this default dataset.
         #
         data = self._get_data_or_bkg(id, bkg_id)
 
--- a/sherpa/astro/utils/__init__.py
+++ b/sherpa/astro/utils/__init__.py
@@ -953,7 +953,7 @@ def eqwidth(data, model, combo, lo=None, hi=None) -> float:
         else:
             eave = np.abs(xlo[ebin - 1] - xlo[ebin])
         if my[ebin] != 0.0:
-            # TOD: shouldn't this be (continuum - cy[ebin]) / continuum?
+            # TODO: shouldn't this be (continuum - cy[ebin]) / continuum? See issue #2381
             eqw += eave * (cy[ebin] - continuum) / continuum
 
     return eqw
--- a/sherpa/ui/tests/test_ui_unit.py
+++ b/sherpa/ui/tests/test_ui_unit.py
@@ -2382,7 +2382,7 @@ def setup_multiple_datasets_linked():
     # different for the two datasets: m = 2, c_1 = 20, c_2 = -20.
     # However, the exact values do not matter for this test, as long
     # as they are separate enough that generating values for them
-    # result in obviously-different values.
+    # results in obviously-different values.
     #
     # The c_1 parameter is set via a link so that the thawed parameters
     # are in a different order for dataset 1 and 2.
--- a/sherpa/ui/utils.py
+++ b/sherpa/ui/utils.py
@@ -4456,11 +4456,11 @@ class Session(NoNewAttributesAfterInit):
            the default identifier is used, as returned by
            `get_default_id`.
         method : callable or None, optional
-           If None, the default, then the data is simulated using the
+           If None (the default) then the data is simulated using the
            `sherpa.utils.poisson_noise` routine. If set, it must be a
-           callable that takes a ndarray of the predicted values and
+           callable that takes an ndarray of the predicted values and
            an optional rng argument that takes a NumPy random
-           generator, and returns a ndarray of the same size with the
+           generator, and returns an ndarray of the same size with the
            simulated data.
 
         See Also
@@ -12137,7 +12137,7 @@ class Session(NoNewAttributesAfterInit):
         >>> fit(3, 4)
         >>> covar(3, 4)
         >>> cmat = get_covar_results().extra_output
-        >>> res = get_draws(3, [4], covar_matrix=cmat)
+        >>> res = get_draws(id=3, otherids=[4], covar_matrix=cmat)
 
         """
 


DougBurke:
@hamogu - I've just realised we have #2379 from looking at the code. I'm going to try addressing it here too.

DougBurke:
@hamogu - I  think I've finished with this PR. There are things to think about `sherpa.astro.utils.eqwidth` but that's for a different issue / PR.

sherpa/astro/ui/tests/test_astro_session.py
```diff
@@ -3700,21 +3700,6 @@ def test_fit_ignores_repeated_otherid(session, caplog):
...
-def test_get_draws_wants_errors(session):
-
-    s = session()
-    s._add_model_types(sherpa.models.basic)
-
-    s.load_arrays(1, [1, 2], [2, 5])
-    s.set_source(1, "const1d.foo")
-    s.fit()
-
-    with pytest.raises(SessionErr,
```
hamogu, line 3713: I suggest keeping this test as a kind of regression test - just make sure that it does NOT raise an error or warning.

sherpa/astro/ui/tests/test_eqwidth_err.py
```diff
@@ -225,3 +225,218 @@ def test_eqwidth_err_arg_square(clean_astro_ui):
...
+    delta_b = resb[2] - resb[1]
+    assert delta_ab < delta_b
+
+
+def test_eqwidth_multi_id_poisson(clean_astro_ui):
+    """Regression test for handling multiple ids (poisson).
+
+    This is only an issue when error=True. At the moment this is
+    treated as a regression test as it is unclear what we want the
+    results to be.
```
hamogu: You are underselling your test. We don't have an analytic solution (Duh, this is simulated random data), but we do know a lot of the properties we want this to be and that ARE tested.

sherpa/astro/ui/utils.py
```diff
@@ -15480,94 +15487,112 @@ def eqwidth(self, src, combo,
...
-                    sampler = NormalParameterSampleFromScaleMatrix()
-                    tmp = sampler.get_sample(fit, mycov=covar_matrix,
-                                             num=niter + 1, rng=self.get_rng())
-                    params = tmp.transpose()
+        # The dataset is only used to define the grid for the model
+        # evaluation, so just use the "default" version and do not
+        # bother about multiple datasets. This could be an issue if
+        # the model evaluation should be used on an extension of the
+        # grid in this default dataset, but this is a theoretical
+        # concern.
```
hamogu:
```suggestion
        # grid in this default dataset.
```

sherpa/astro/utils/__init__.py
```diff
@@ -892,44 +890,71 @@ def calc_model_sum2d(data, model, reg=None):
...
+    eqw = 0.0
+    for ebin, continuum in enumerate(my):
+        # TODO: calculate the actual bin width for integrated data sets.
         if ebin < (num - 1):
             eave = np.abs(xlo[ebin + 1] - xlo[ebin])
         else:
             eave = np.abs(xlo[ebin - 1] - xlo[ebin])
         if my[ebin] != 0.0:
-            eqw += eave * (cy[ebin] - my[ebin]) / my[ebin]
+            # TOD: shouldn't this be (continuum - cy[ebin]) / continuum?
```
hamogu: Yes, an emission line should have a negative equivalent width. Can we fix that?
I know it's not what this PR set out to do, and it's an API change, but it's fixing a bug. I don't think we should leave a bug in if we are changing the code that generates it anyway.
(I don't think we should leave bus in in general, but sometimes we just don't have time to address them right there and then, while in this case, we absolutely can.)

sherpa/ui/utils.py
```diff
@@ -4454,9 +4455,13 @@ def fake(self,
            The identifier for the data set to use. If not given then
            the default identifier is used, as returned by
            `get_default_id`.
-        method : func
-           The function used to create a random realisation of
-           a data set.
+        method : callable or None, optional
+           If None, the default, then the data is simulated using the
```
hamogu: just simplifying the reading a bit by making the nesting more clear:
```suggestion
           If None (the default) then the data is simulated using the
```

sherpa/ui/utils.py
```diff
@@ -4454,9 +4455,13 @@ def fake(self,
...
+           `sherpa.utils.poisson_noise` routine. If set, it must be a
+           callable that takes a ndarray of the predicted values and
```
hamogu:
```suggestion
           callable that takes an ndarray of the predicted values and
```

sherpa/ui/utils.py
```diff
@@ -4454,9 +4455,13 @@ def fake(self,
...
+           an optional rng argument that takes a NumPy random
+           generator, and returns a ndarray of the same size with the
```
hamogu: I'm not a native speaker, but I pronounce it "en-d-ar-ray" and thus it should be "an ndarray", I think. 
```suggestion
           generator, and returns an ndarray of the same size with the
```

sherpa/ui/utils.py
```diff
@@ -12110,23 +12126,31 @@ def get_draws(self,
...
+        Fit models to datasets 3 and 4, then calculate the covariance
+        matrix (stored in the extra_output field) and pass it through
+        to get_draws (which also evaluates the models for datasets 3
+        and 4):
 
-        ids, fit = self._get_fit(id, otherids)
+        >>> fit(3, 4)
+        >>> covar(3, 4)
+        >>> cmat = get_covar_results().extra_output
+        >>> res = get_draws(3, [4], covar_matrix=cmat)
```
hamogu: I know it's correct as written, but the `[4]` looks a bit odd. I think setting it as a keyword argument makes the example easier to read.
```suggestion
        >>> res = get_draws(id=3, otherids=[4], covar_matrix=cmat)
```

sherpa/ui/tests/test_ui_unit.py
```diff
@@ -2347,3 +2347,260 @@ def check_ratio(vsigma, vsigma1, minval, maxval):
...
+    ui.load_arrays(2, [20, 22, 24, 26],
+                   [28, 22, 28, 38])
+
+    ui.set_stat("cash")
+
+    # Fit y = mx + c to the two datasets, with m being the same but c
+    # different for the two datasets: m = 2, c_1 = 20, c_2 = -20.
+    # However, the exact values do not matter for this test, as long
+    # as they are separate enough that generating values for them
+    # result in obviously-different values.
```
hamogu:
```suggestion
    # results in obviously-different values.
```

hamogu (changes requested):

sherpa/astro/utils/__init__.py
DougBurke: I would rather look at this in a different PR, since I am currently trying to convince myself that this code is actually doing the right thing or is accidentally including the bin width twice. My heart says it is but my initial investigation says no, but I haven't convinced myself.

DougBurke: I just created #2381 so we have somewhere to talk about this.

hamogu: At a minimum, let's fix TOD here -> TODO

sherpa/astro/ui/tests/test_astro_session.py
DougBurke: We do have a bunch of other tests that now do `fit` followed by `get_draws`, but I can add this back in.

sherpa/astro/ui/tests/test_eqwidth_err.py
DougBurke: Well, as this test changed behaviour during this PR I somewhat stand by the original statement, but I can remove it now that we have a better understanding of how the code works (and is tested)!

DougBurke:
```suggestion
    This is only an issue when error=True.
```

sherpa/astro/utils/__init__.py
DougBurke:
```suggestion
            # TODO: shouldn't this be (continuum - cy[ebin]) / continuum? See issue #2381
```

PR #2377, opened 2025-09-14
UI: save_all improvements for PHA2 files. Fix #1882

### LINKED ISSUE #1882: save_all fails miserable for pha2 files
PHA2 files have multiple issues

a) multiple backgrounds, which suffer from a similar issue to #320

b) do not read in the files correctly

The output has

load_pha(1, filename)
load_pha(2, filename)
...
load_pha(12, filename)

so when this is re-run we do 

load_pha(n, filename)

and it fills in datasets n to n+11 with the 12 datasets.

So you end up with 23 datasets, and you have no idea what is what (technically it'll be 12 to 23).

DougBurke:
We now store information when loading a PHA2 file (identified as the only load command that can load in multiple data files from a single call). This information is then used to avoid writing out load_pha/bkg calls that end up being wrong.

I believe that this technique - that is, store the load data call information - should be expanded for all load calls. However, this is a large change that needs work, since there are cases where the new code can fail (but it is expected that users who are doing this are capable of writing out their own Python code to serialize the data).


hamogu (approved):
I like the way where this is going. We've argued for a while that we should keep more meta-data because it allows us to do things like this fix. In that sense, we could probably improve on the format of that meta-data specification to make it more general, but since it's all in variables whose name indicates that they are private, this PR does not preclude us from making such changes in the future and it does allows us to fix the bug in question right now.

I fully support to get this in now to get the real benefit rather than waiting for some overengeeered version we may or may not come up with in the future.

DougBurke:
For reference, when I started on this I did start over-engineering it, but that lead to a lot of code churn which wasn't appropriate at this stage of the release cycle. At least not for this PR ....

PR #2385, opened 2025-09-20
Correct eqwidth sign calculation

### LINKED ISSUE #2381: eqwidth calculation questions
As noted in https://github.com/sherpa/sherpa/pull/2374#discussion_r2352394863 we believe the eqwidth is returning the wrong sign.

I also wonder whether, for X-ray data, whether it is including the bin width one-too-many times (since `model(xlo, xhi)` should be returning the integral of the model for each bin, so we shouldn't need to include the bin width in the calculation. However, my initial investigation (where I try changing the bin size) does not seem to show a significant change as bin width decreases, so I am not actually sure whether this part is a problem.

Example code

```python
import numpy as np

from sherpa.astro import ui

def calc_eqw(continuum, line, binwidth):
    egrid = np.arange(0.1, 4, binwidth)
    elo = egrid[:-1]
    ehi = egrid[1:]

    rmf = ui.create_rmf(elo, ehi)
    arf = ui.create_arf(elo, ehi)
    nbins = len(elo)

    ui.dataspace1d(1, nbins, dstype=ui.DataPHA)
    ui.set_arf(arf)
    ui.set_rmf(rmf)

    return ui.eqwidth(continuum, line)


base_model_s = ui.create_model_component("powlaw1d", "base_model_s")
line_model_s = ui.create_model_component("normgauss1d", "line_model_s")

base_model_s.ampl = 7
line_model_s.pos = 2
line_model_s.fwhm = 0.4

base_model_x = ui.create_model_component("xspowerlaw", "base_model_x")
line_model_x = ui.create_model_component("xsgaussian", "line_model_x")

base_model_x.norm = base_model_s.ampl
line_model_x.linee = line_model_s.pos
line_model_x.sigma = line_model_s.fwhm / np.sqrt(8 * np.log(2))

# The line_model_x.norm value is the flux of the line in units of
# photon/cm^2/s, which is why the sherpa normgauss1d model is used.
#

print("")
print("# Sherpa models")

for binsize in [0.1, 0.05, 0.02, 0.01, 0.001]:
    eqw = calc_eqw(base_model_s, base_model_s - line_model_s, binsize)
    print(f"  {binsize:.3f} : {eqw:.6e}")

print("")
print("# XSPEC models")

for binsize in [0.1, 0.05, 0.02, 0.01, 0.001]:
    eqw = calc_eqw(base_model_x, base_model_x - line_model_x, binsize)
    print(f"  {binsize:.3f} : {eqw:.6e}")
```

Which displays the following

```
# Sherpa models
  0.100 : -2.856543e-01
  0.050 : -2.856993e-01
  0.020 : -2.857119e-01
  0.010 : -2.857137e-01
  0.001 : -2.857143e-01

# XSPEC models
  0.100 : -2.856543e-01
  0.050 : -2.856993e-01
  0.020 : -2.857119e-01
  0.010 : -2.857137e-01
  0.001 : -2.857143e-01
```

This shows the original problem; this is an absorption model but the equivalent width is negative.

Here's the model (for the binsize=0.001 case)

```python
ui.set_source(base_model_s - line_model_s)
ui.plot_model(ylog=True)
```

Image

DougBurke:
# Summary

Correct the eqwidth calculation so that absorption features have a positive width and emission features a negative width (previously the sign was calculated incorrectly but the absolute value was correct). Fix #2381.

# Details

This is based on #2374 as otherwise it would conflict.

The change is easy to make - as it's just changing "line - continuum" to "continuum - line" but first I add explicit PHA checks where we know the expected answer (rather than, as in the other cases, treat it as a regression test).

I am not convinced we are doing the correct thing for `Data1DInt` and `Data1D` cases, but that is a larger change than I want to look at here (and it's not like this code changes the existing behaviour for these cases).


DougBurke:
@anetasie - do we want this for 4.18?

DougBurke:
@anetasie - from https://en.wikipedia.org/wiki/Equivalent_width the definition has EW being positive for absorption features and negative for emission features. Our existing code has it the wrong way around - emission lines have a positive EW and absorption lines a negative EW

e.g. with CIAO 4.17

```
sherpa In [4]: set_source(powlaw1d.cont + gauss1d.line)

sherpa In [5]: line.ampl = 0.5

sherpa In [6]: cont.ampl = 1

sherpa In [7]: eqwidth(cont, cont + line)
Out[7]: 8.699507408295737

sherpa In [8]: eqwidth(cont, cont - line)
Out[8]: -8.699507408295737

```

anetasie:
The convention in for X-ray analysis of Fe-line is to have the positive value. We can supply the reference. I don't think we need to change that for 4.18

DougBurke:
I'd like the reference because we currently point users to https://en.wikipedia.org/wiki/Equivalent_width which has absorption lines being positive and emission lines being negative (or you take the absolute value).

DougBurke:
The trouble with using the absolute value is that you then lose information - e.g. when running `eqwidth(..., errors=True)` you can no-longer identify cases where the sign changes - e.g.

```
midval, loval, hival, params, eqws = eqwidth(..., error=True)`
```

We have at least one test case where `loval` and `hival` have different signs, so using the absolute value would bias the distribution.

Although as someone who doesn't really use equivalent widths maybe this isn't a real issue, but I'd be surprised if that's the case.

anetasie:
We cannot solve everything.

anetasie:
@DougBurke I just build the branch, checked the docs and run examples.

sherpa/astro/ui/utils.py
```diff
@@ -15461,113 +15473,130 @@ def eqwidth(self, src, combo,
...
+        displayed with Sherpa plotting commands.
 
         >>> res = eqwidth(p1, p1 + g1, error=True)
         >>> ewidth = res[0]  # the median equivalent width
         >>> errlo = res[1]   # the one-sigma lower limit
         >>> errhi = res[2]   # the one-sigma upper limit
         >>> pars = res[3]    # the parameter values used
         >>> ews = res[4]     # array of eq. width values
+        >>> plot_pdf(ews)    # probability density
+        >>> plot_cdf(ews)    # cumalitive distribution
```
wmclaugh:
```suggestion
        >>> plot_cdf(ews)    # cumulative distribution
```

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -872,3 +874,113 @@ def test_expand_grouped_mask(mask, group, expected):
...
+POS_GAUSS = NormGauss1D()
+POS_GAUSS.ampl = 4
+POS_GAUSS.pos = 5
+POS_GAUSS.fwhm = 1
+
+
+def setup_eqwidth_datapha() -> DataPHA:
+    """Create a PHA datastet for use with eqwidth tests"""
+
+    # Pick an energy grid that encomposes the line emission of
```
wmclaugh:
```suggestion
    # Pick an energy grid that encompasses the line emission of
```

sherpa/astro/utils/tests/test_astro_utils_unit.py
```diff
@@ -872,3 +874,113 @@ def test_expand_grouped_mask(mask, group, expected):
...
+    #    integral(line / continuum )                               ==
+    #    integral( line ) / c0                                     ==
+    #    ampl / c0
+    #
+    # Which here is
+    #    4 / 10
+    #
+    # For combined == continuum - line then we get -4/10
+    #
+    # (although the George & Fabian paper do not talk about absorption
```
wmclaugh:
```suggestion
    # (although the George & Fabian paper does not talk about absorption
```

PR #1544, opened 2022-07-06
BUILD update setup.cfg based on the conda environment

DougBurke:
This is not needed if we take #2484

# Summary

Add the scripts/use_conda_config script to allow easier configuration when using conda-like environments and update scropts/use_ciao_config to support mamba or micromamba instead of conda.

# Details

The `scripts/use_conda_config` script will change the setup.cfg file to build against conda-provided packages:

- fftw
- xspec-modelsonly

The idea is to make it easy to make these changes (it's something I often do). Is it worth keeping?

There is duplicated logic with `scripts/use_ciao_config` but they are technically separate (you could imagine re-writing `use_ciao_config` to call this first and then post-process the output for the CIAO-specific logic) but I'm not sure it's worth it (unless we re-write these scripts in Python).

# Example

```
% ./scripts/use_conda_config 
Found FFTW version: 3.3.10
Found XSPEC version: 12.12.0
Updating setup.cfg
(sherpa-main) dburke@dburke-Precision-5530:~/sherpa/sherpa-main$ git diff setup.cfg
diff --git a/setup.cfg b/setup.cfg
index df19e1df..329c1011 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,15 +36,15 @@ parentdir_prefix = sherpa-
 
 # FFTW Library
 # Uncomment to use a local installation
-#fftw=local
+fftw=local
 
 # If fftw=local uncomment the following lines and
 # change the default location of libraries and the name
 # of the library to be linked (usually fftw3)
 # (include multiple values by separating them with spaces)
-#fftw-include_dirs=build/include
-#fftw-lib-dirs=build/lib
-#fftw-libraries=fftw3
+fftw-include_dirs=/home/dburke/miniconda3/envs/sherpa-main/include
+fftw-lib-dirs=/home/dburke/miniconda3/envs/sherpa-main/lib
+fftw-libraries=fftw3

 # Region Library
 # Uncomment to use a local installation
@@ -82,7 +82,7 @@ source-dir = docs
 # XSPEC Models
 [xspec_config]
 # Uncomment (set to True) to build XSPEC extension
-#with-xspec=True
+with-xspec=True
 
 # If with-xspec is True, make sure to point Sherpa to right
 # XSPEC-related libraries and to indicate the XSPEC version.
@@ -111,9 +111,9 @@ source-dir = docs
 #
 # The gfortran_lib_dirs should be set if needed.
 #
-#xspec_version = 12.9.0
-#xspec_lib_dirs = None
[...]

PR #2393, opened 2025-10-02
XSPEC: add support for XSPEC 12.15.1

### LINKED ISSUE #2392: XSPEC 12.15.1 has been released but will not work with Sherpa
It looks like there have been changes to the XSPEC function names in 12.15.1 that means that we currently can not build with it.

This is with CIAO 4.17 running the check (so it doesn't recognize 12.15.0)

```
% python ~/sherpa/sherpa-1/scripts/check_xspec_update.py model.dat | grep function
  [ 1] function name change: C_spin to dospin
  [ 1] function name change: C_nthcomp to donthcomp
  [ 1] function name change: rgsxsrc to C_rgsxsrc  **
  [ 1] function name change: ismabs to F_ismabs  **
  [ 1] function name change: ismdust to F_ismdust   **
  [ 1] function name change: olivineabs to F_olivineabs  **
  [ 1] function name change: C_xspwab to xspwab
```

Those labelled ** are not a problem.

There are also changes for models added in 12.15.0 that are not shown here.

# NOTES

I believe there are some issues with the XSPEC 12.15.1 release but I have not reported them yet as HEASARC is closed down. Problems include

 -  XSPEC initalization defaults to atomdb/nei/spex versions of 3.13/3.13/3.08 but the FORTRAN FNINIT code we have to use uses 3.12/3.12/3.08 (at least that is what it seems like)
 - however, the $HEADAS/spectral/modelData/latest.txt uses "3.0.8" for the SPEX version and this will win out if you have a "new" ~/.xspec/Xspec.init file with "SPEX_VERSION: latest" in it. I believe this is wrong and it should be "3.08".

DougBurke:
# Summary

Allow Sherpa to build against XSPEC 12.15.1 and simplify the XSPEC installation steps. Add 13 new additive models: XSbfekblor, XSfekblor, XSzfekblor, XSzbfekblor, XSbfeklor, XSzbfeklor, XSrsapec, XSrsvapec, XSrsvvapec, XSrsgaussian, XSrsrnei, XSrsvrnei, XSrsvvrnei. Fix #2392.

# Details

The documentation for the XSPEC configuration has (hopefully) been improved, and now mentions the (relatively new) HEASARC xspec builds. It turns out that with recent XSPEC builds we do not need to be as explicit as we used to with the "ancillary" libraries as they are referenced via rpath/whatever the magic is. This means that we not need to include "hdsp_<version>" in the XSPEC libraries list, nor do we need to give CCFits and wcslib libraries (again with versions). So the documenation has had the "per-XSPEC-version" list of options removed. I have also changed the XSPEC build step to default to $HEADAS/include and $HEADAS/lib for the XSPEC paths [*] and then allowed environment variables to be used for any of the paths **but only for** the XSPEC options (i.e. not FFTW or region or ...).

[*] Of cource, the CXC-provided `xspec-modelsonly` packages actually have $HEADAS pointing to somewhere different, and you actually want either $HEADAS/../lib or $CONDA_PREFIX/lib instead. Which means that the `scripts/use_ciao_config` and `.github/scrpts/setup_xspec.sh` helper scripts need minor tweaks (because the default options in `setup.cfg` have changed so the scripts don't recognize the lines to change). Other downstream code, like #1544, will also likely need to be updated.

Allows Sherpa to be built with XSPEC 12.15.1 and provide access to the 13 new additive models:

    XSbfekblor     XSfekblor    XSzfekblor    XSzbfekblor
    XSbfeklor      XSzbfeklor
    XSrsapec       XSrsvapec     XSrsvvapec
    XSrsgaussian
    XSrsrnei        XSrsvrnei       XSrsvvrnei

XSPEC 12.15.1 has added support for `"latest"` settings for the apec/atomdb, nei, and spex database versions (the names of this are not clear, e.g. apec vs atomdb, nei vs neiapec). For example:

```
% grep VERSION ~/.xspec/Xspec.init
ATOMDB_VERSION:  latest
SPEX_VERSION:  latest
NEI_VERSION: latest
```

There is now a `latest.txt` file in the `modelData` directory that can be used to find out what "latest" means, but this is not guaranteed to exist, in which case it falls back to default values. It appeas that the default depends on whether you are running XSPEC (the program) or accessing just the model library.

For now we attempt to follow the program versions (as they are newer) as the defaults (when the `latest.txt` file does not exist), but **only** when the interface is built with XSPEC 12.15.1 (or later), as it's not clear what to do with older XSPEC libraries. One option would be to just error out if `latest` is set but there's no `latest.txt` file: I decided against this as it's possible to get a 12.15.1 `Xspec.init` file, which has `latest` values in it, and then switch to use Sherpa (e.g. via CIAO) and it would be annoying for it to then fall over (although there is a strong argument to be said it would make things more obvious about what is going on, but I don't think most users would agree with this choice). 

The `xset` database now gets three values automatically associated with it. This is not technically a breaking change but it can be surprising (e.g. the `save_all` output will now include these three settings if XSPEC is installed/used):

```
>>> xspec.get_xsversion()
'12.15.1'
>>> xspec.get_xsxset()
{'APECROOT': '3.1.3', 'NEIAPECROOT': '3.1.3', 'SPEXROOT': '3.0.8'}
```

# Notes

I believe there are some issues with the XSPEC 12.15.1 release but I have not reported them yet as HEASARC is closed down. Problems include

- XSPEC initalization defaults to atomdb/nei/spex versions of 3.13/3.13/3.08 but the FORTRAN `FNINIT` code we have to use uses 3.12/3.12/3.08 (at least that is what it seems like)
- however, the $HEADAS/spectral/modelData/latest.txt uses "3.0.8" for the SPEX version and this will win out if you have a "new" ~/.xspec/Xspec.init file with "SPEX_VERSION: latest" in it. I believe this is wrong and it should be "3.08".

**EDIT** These have now been reported to the XSPEC team (Jan 14 2026).

Originally this supported `get_xsversion("atomdb")` and `set_xsversion("atomdb", "4.0.1")` but I have removed this because

a) too many ways to set the same information (e.g. `set_xsxset("APECROOT", "4.0.1")`) makes it confusing
b) if you use a pre-XSPEC 12.15.1 library then the different ways to set this information can lead to different behavior (which means we probably don't want to do this until we default to 12.15.1, if ever)

# Examples

XSPEC 12.15.1 installed via HEASARC conda packages, including the data

```
% grep VERSION ~/.xspec/Xspec.init
ATOMDB_VERSION:  latest
SPEX_VERSION:  latest
NEI_VERSION: latest
% cat $HEADAS/spectral/modelData/latest.txt 
ATOMDB_VERSION: 3.1.3
SPEX_VERSION: 3.0.8
NEI_VERSION: 3.1.3
```

```
>>> xspec.get_xsversion()
'12.15.1'
>>> xspec.get_xsxset()
{'APECROOT': '3.1.3', 'NEIAPECROOT': '3.1.3', 'SPEXROOT': '3.0.8'}
```

We can see an apec model and some of the files it loads by bumping the `chatter` setting:

```
>>> xspec.set_xschatter(50)
>>> mdl = xspec.XSapec()
calcMultiTempPlasma : plasmaType = 6

calcMultiTempPlasma : plasmaType = 6
calcMultiTempPlasma : TraceAbund = 1

Reading APEC data from 3.1.3

Reading continuum   data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_coco.fits
Reading line        data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_line.fits
Reading ion balance data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.0_ionbal.fits

Reading continuum data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_coco.fits
Reading line data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_line.fits
Reading ion balance data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.0_ionbal.fits
Not applying thermal broadening
Velocity broadening : 0
Linear interpolation on tabulated temperatures
Not multi-threading over temperatures
Including electron-electron bremsstrahlung
Not doing electron density correction on normalization
Column (rsColumn) for resonance scattering 0

For continuum interpolating between 0.966887 and 1.02418 for 1
with coefficients 0.422019 and 0.577981

For lines interpolating between 0.966887 and 1.02418 for 1
with coefficients 0.422019 and 0.577981

Reading data for temperature 122 0.966887 0.966882

Reading data for temperature 122 0.966887 0.966882
Read 30417 element lines and 30  continuum points for 30 unique elements.

Reading data for temperature 123 1.02418 1.02417

Reading data for temperature 123 1.02418 1.02417
Read 28818 element lines and 30  continuum points for 30 unique elements.

array([2.01612911, 0.30307714, 0.2181515 ])
```

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -248,7 +248,33 @@ by the actual path to the HEADAS installation, and the versions of
 the libraries - such as ``CCfits_2.7`` - may need to be changed to
 match the contents of the XSPEC installation.
 
-1. If the full XSPEC 12.15.0 system has been built then use::
+.. note::
+
+   If XSPEC has been installed via conda - such as with the
+   `HEASARC <https://heasarc.gsfc.nasa.gov/docs/software/conda.html>`_
+   or
+   `CIAO <https://cxc.cfa.harvard.edu/ciao/download/conda.html>`_
+   channels - then the only settings which are needed are::
+
+     with_xspec
+     xspec_version
+     xspec_lib_dirs
+     xspec_include_dirs
+
+* If the full XSPEC 12.15.1 system has been built then use::
+
+       with_xspec = True
+       xspec_version = 12.15.1
+       xspec_lib_dirs = $HEADAS/lib
+       xspec_include_dirs = $HEADAS/include
+       xspec_libraries = XSFunctions XSUtil XS hdsp_6.36
+       ccfits_libraries = CCfits_2.7
+       wcslib_libraries = wcs-8.3
+
+  where the version numbers were taken from version 6.36 of HEASOFT and
+  may need updating with a newer release.
+
+* If the full XSPEC 12.15.0 system has been built then use::
 
        with_xspec = True
        xspec_version = 12.15.0
@@ -258,10 +284,7 @@ match the contents of the XSPEC installation.
        ccfits_libraries = CCfits_2.7
        wcslib_libraries = wcs-8.3
 
-   where the version numbers were taken from version 6.35 of HEASOFT and
-   may need updating with a newer release.
-
-2. If the full XSPEC 12.14.1 system has been built then use::
+* If the full XSPEC 12.14.1 system has been built then use::
 
        with_xspec = True
        xspec_version = 12.14.1
@@ -271,10 +294,7 @@ match the contents of the XSPEC installation.
        ccfits_libraries = CCfits_2.6
        wcslib_libraries = wcs-8.3
 
-   where the version numbers were taken from version 6.34 of HEASOFT and
-   may need updating with a newer release.
-
-3. If the full XSPEC 12.14.0 system has been built then use::
+* If the full XSPEC 12.14.0 system has been built then use::
 
        with_xspec = True
        xspec_version = 12.14.0
@@ -284,10 +304,7 @@ match the contents of the XSPEC installation.
        ccfits_libraries = CCfits_2.6
        wcslib_libraries = wcs-8.2.1
 
-   where the version numbers were taken from version 6.33.1 of HEASOFT and
-   may need updating with a newer release.
-
-4. If the full XSPEC 12.13.1 system has been built then use::
+* If the full XSPEC 12.13.1 system has been built then use::
 
        with_xspec = True
        xspec_version = 12.13.1
@@ -297,10 +314,7 @@ match the contents of the XSPEC installation.
        ccfits_libraries = CCfits_2.6
        wcslib_libraries = wcs-7.7
 
-   where the version numbers were taken from version 6.32 of HEASOFT and
-   may need updating with a newer release.
-
-5. If the full XSPEC 12.13.0 system has been built then use::
+* If the full XSPEC 12.13.0 system has been built then use::
 
        with_xspec = True
        xspec_version = 12.13.0
@@ -310,11 +324,11 @@ match the contents of the XSPEC installation.
        ccfits_libraries = CCfits_2.6
        wcslib_libraries = wcs-7.7
 
-6. If the model-only build of XSPEC - created with the
-   ``--enable-xs-models-only`` flag when building HEASOFT - has been
-   installed, then the configuration is similar, but the library names
-   may not need version numbers and locations, depending on how the
-   ``cfitsio``, ``CCfits``, and ``wcs`` libraries were installed.
+* If the model-only build of XSPEC - created with the
+  ``--enable-xs-models-only`` flag when building HEASOFT - has been
+  installed, then the configuration is similar, but the library names
+  may not need version numbers and locations, depending on how the
+  ``cfitsio``, ``CCfits``, and ``wcs`` libraries were installed.
 
 A common problem is to set one or both of the ``xspec_lib_dirs``
 and ``xspec_lib_include`` options to the value of ``$HEADAS`` instead of
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -219,22 +219,7 @@ XSPEC
 
 Sherpa can be built to use the Astronomy models provided by
 :term:`XSPEC`. To enable XSPEC support, several changes must be
-made to the ``xspec_config`` section of the ``setup.cfg`` file. The
-available options (with default values) are::
-
-    with_xspec = False
-    xspec_version = 12.14.1
-    xspec_lib_dirs = None
-    xspec_include_dirs = None
-    xspec_libraries = XSFunctions XSUtil XS
-    cfitsio_lib_dirs = None
-    cfitsio_libraries =
-    ccfits_lib_dirs = None
-    ccfits_libraries =
-    wcslib_lib_dirs = None
-    wcslib_libraries =
-    gfortran_lib_dirs = None
-    gfortran_libraries =
+made to the ``xspec_config`` section of the ``setup.cfg`` file.
 
 To build the :py:mod:`sherpa.astro.xspec` module, the
 ``with_xspec`` option must be set to ``True`` **and** the
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -218,112 +218,64 @@ XSPEC
 ^^^^^
 
 Sherpa can be built to use the Astronomy models provided by
-:term:`XSPEC`. To enable XSPEC support, several changes must be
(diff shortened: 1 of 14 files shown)


DougBurke:
Rebased to address conflicts.

DougBurke:
Rebased to pick up some changes (with some commits having moved to other PRs).

DougBurke:
Rebased on top of #2444 and rework the ordering (since handling of "latest" needs to be done as soon as Sherpa is built against 12.15.1).

DougBurke:
XSPEC provides a number of ways of setting values. For the ATOMDB/... values this PR supports set_xsxset and set_xsversion. Perhaps we just remove the set_xsversion as (at least for 12.15.1) it looks like they all end up changing the same internal settings (it wasn't clear in earlier versions).

**EDITED TO ADD** I have now removed these additions as they didn't work well with a pre-12.15.1 system and there are too many ways to set the same information.

DougBurke:
I did not mean to assign this to @hamogu and the GH web interface is just playing up too much recently ....

DougBurke:
@nplee - can you check this out (the documentation is meant to suggest how to use the HEASARC XSPEC 12.15.1/HEADAS 6.36 conda build if you don't have a 12.15.1 XSPEC lying around!) ?

nplee:
@DougBurke: I was able to build Sherpa into Conda using both a standalone XSpec 12.15.1 build and the HEASARC's distribution of Conda XSpec 12.15.1.  I just needed to set to use the right `$HEADAS/lib` and `$HEADAS/include` paths in the `setup.cfg` file and use gcc/g++ v14, using v15 resulted in failures.  So long as I used the correct `Xspec.init` file, all the pytests ran to completion.

sherpa/astro/ui/tests/test_serialize.py
```diff
@@ -2542,6 +2544,33 @@ def mymodel_func(pars, x, xhi=None):
...
+        toks = verstr.split('.')
+        patch = ''
+    else:
+        toks = verstr[:-1].split('.')
+        patch = verstr[-1]
+
+    xsver = tuple([int(t) for t in toks] + [patch])
+
+    # Add in the XSET settings, but only for XSPEC 12.15.1 and
+    # later. Since the versions can be uptated, take advantage of the
```
hamogu:
```suggestion
    # later. Since the versions can be updated, take advantage of the
```

hamogu (approved):
Nick tried is out and I know @wmclaugh will look at the C++ code. The Python side and the docs look good to me. 
It's a constant churn to keep up with XSPEC (and to support multiple versions). Not for this PR, but one thought for the future would be to decide if it's worth the effort to keep supporting multiple versions of XSPEC. For example, if we supported the 12.15.1 only and dropped all older ones, then we wouldn't need to worry about edge cases where "latest" exists, but a user has built agains an older XSPEC version etc.

DougBurke:
So, as CIAO doesn't guarantee to use the latest version we can't just drop all-but-the-latest. I do think we want to do a clean up and drop older support (e.g. 12.13.x). It's not clear to me what the XSPEC version scheme means (e.g. would we want to support all 12.14.x or may be just 12.1.41) but again for a different PR (and once CIAO updates it's version).

DougBurke:
The tests fail because of DS9 changes; it would need #2453 (or equivalent) to pass.

DougBurke:
Rebased as the build should be okay now ...

docs/install.rst
```diff
@@ -218,113 +218,64 @@ XSPEC
...
+  xspec_include_dirs = $CONDA_PREFIX/include
+
+The CIAO XSPEC package includes the data files needed to evaluate the
+models.
+
+Other
+"""""
+
+Modern XSPEC builds use run-time search paths to reference most of the
+libraries it uses, so the only settings that should be necessary are
```
wmclaugh:
```suggestion
libraries they uses, so the only settings that should be necessary are
```

DougBurke:
I expect the tests to fail as I haven't rebased it to address some recent changes unrelated to this PR.

DougBurke:
Closed by #2484 - this logic is now in #2492

PR #2397, opened 2025-10-02
TESTS: support HEASARC XSPEC conda builds for XSPEC tests

### LINKED ISSUE #2396: HEASARC XSPEC conda instalation and our test suite
For historical reasons, the CIAO conda installation of XSPEC (the model libraries) does not match the HEASARC approach. As there is now a HEASARC conda XSPEC installation, we actually hit this difference in two tests. We can easily fix the tests, but I wanted to have some documentation.

The test failures are not due to an issue in the code, bit are instead due to assumptions made in the test itself.

Here are the test failures:

```
=============================================================================== FAILURES ===============================================================================
______________________________________________________________________ test_manager_path_default _______________________________________________________________________

    @requires_xspec
    def test_manager_path_default():
        """Check the expected default setting for the manager path.
    
        Ideally this test would be run before any other
        tests of XSPEC are made (i.e. any XSPEC code is called).
        """
    
        # Is this always going to be correct?
        default_path = os.path.join(os.environ['HEADAS'],
                                    '../spectral/manager')
    
        from sherpa.astro import xspec
    
        oval = xspec.get_xspath_manager()
    
        # Normalize the paths to remove double / - e.g. if HEADAS is
        # set to /a/b/c/ rather than /a/b/c.
        #
>       assert os.path.normpath(oval) == os.path.normpath(default_path)
E       AssertionError: assert '/home/dburke...ctral/manager' == '/home/dburke...ctral/manager'
E         
E         - /home/dburke/micromamba/envs/xspec12150/spectral/manager
E         + /home/dburke/micromamba/envs/xspec12150/heasoft/spectral/manager
E         ?                                         ++++++++

sherpa/astro/xspec/tests/test_xspec_unit.py:249: AssertionError
_______________________________________________________________________ test_model_path_default ________________________________________________________________________

    @requires_xspec
    def test_model_path_default():
        """Check the expected default setting for the model data path.
    
        Ideally this test would be run before any other
        tests of XSPEC are made (i.e. any XSPEC code is called).
        """
    
        from sherpa.astro import xspec
    
        # Is this always going to be correct?
        #
        try:
            default_path = os.environ['XSPEC_MDATA_DIR']
        except KeyError:
            default_path = os.path.join(os.environ['HEADAS'],
                                        '../spectral/modelData/')
    
        oval = xspec.get_xspath_model()
    
        # Normalize the paths to remove double / - e.g. if HEADAS is
        # set to /a/b/c/ rather than /a/b/c.
        #
>       assert os.path.normpath(oval) == os.path.normpath(default_path)
E       AssertionError: assert '/home/dburke...ral/modelData' == '/home/dburke...ral/modelData'
E         
E         - /home/dburke/micromamba/envs/xspec12150/spectral/modelData
E         + /home/dburke/micromamba/envs/xspec12150/heasoft/spectral/modelData
E         ?                                         ++++++++

sherpa/astro/xspec/tests/test_xspec_unit.py:275: AssertionError
```

DougBurke:
# Summary

Fix test failures when using the HEASARC XSPEC conda installation. Fix #2396.

# Details

The existing test suite is based on the CIAO-provided version of XSPEC, and this has historically laid files out differently to the HEASARC approach. With HEASARC now providing an XSPEC build via conda, we can see this difference in two tests.

So, the tests have been updated to allow either layout, as really all we care about is that the result is "meaningful". It is not worth trying to use out-of-band data to identify the "expected" layout.

### DIFF
--- a/sherpa/astro/xspec/tests/test_xspec_unit.py
+++ b/sherpa/astro/xspec/tests/test_xspec_unit.py
@@ -235,18 +235,23 @@ def test_manager_path_default():
     tests of XSPEC are made (i.e. any XSPEC code is called).
     """
 
-    # Is this always going to be correct?
-    default_path = os.path.join(os.environ['HEADAS'],
-                                '../spectral/manager')
-
     from sherpa.astro import xspec
 
-    oval = xspec.get_xspath_manager()
+    oval = Path(xspec.get_xspath_manager()).resolve()
+    got = str(oval)
 
     # Normalize the paths to remove double / - e.g. if HEADAS is
     # set to /a/b/c/ rather than /a/b/c.
     #
-    assert os.path.normpath(oval) == os.path.normpath(default_path)
+    # The expected path depends on how XSPEC has been installed,
+    # since the CIAO-related environments do things a little
+    # differently to HEASARC ones, for historical reasons.
+    #
+    headas = Path(os.environ['HEADAS'])
+    heasarc_path = (headas / 'spectral/manager').resolve()
+    ciao_path = (headas / '../spectral/manager').resolve()
+
+    assert got in [str(heasarc_path), str(ciao_path)]
 
 
 @requires_xspec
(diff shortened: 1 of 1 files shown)


hamogu (approved):
I didn't try this out in different environments, but it looks reasonable to me. It passes our current tests and it only modifies tests, so users would not be impacted (unless they run the tests themselves) even if this was wrong in some way.

It's definitely good to make this work for different installation options, so that, maybe one day, the CXC can drop building XSPEC and just use the package that HEASARC is already providing...

DougBurke:
I probably need to add some documentation to our build instructions on how to use the HEASARC conda channel, but that is only really worth it once we can build with XSPEC 12.15.1 - e.g. #2393  - as the HEASARC conda build for 12.15.0 is, for some reason, missing some stuff needed to build things properly.

PR #2403, opened 2025-10-20
Update the Zenodo information

DougBurke:
# Summary

Update the stored Zenodo release data for 4.17.0, 4.17.1, and 4.18.0.

# Details

There are also a few minor changes to the documentation and script used to create this output (a warning if the data is already known about).

For users this should not change the results of getting a citation, other than

- making it faster (as no internet request is made)
- work even without internet (as no internet request is made)

at least for these three versions.


DougBurke:
Rebased as I wanted to fix a typo in the commit message and to separate out #2441

harlantc (approved):
LGTM

PR #2404, opened 2025-10-20
Split `TableModel` into separate classes

### LINKED ISSUE #1056: `basic.TableModel` ignore `xhi` argument in binned datasets
The `basic.TableModel` does work with integrated datasets, but it does so by simply ignoring the upper edge of the bin and evaluating the lower edge only. 

Doing this better requires some thought on integrating and interpolating when the bins don't match up with the bins of the model. (Think "flux conservation").

It seems that the TableModel is not heavily used (since XSPEC Table models - probably the more common use case for Sherpa - are handled differently *), so this might not be a priority issue.

Discussion in #142 and #1039.

*: I did not check if better, but certainly differently.

### LINKED ISSUE #1620: table model review

### LINKED ISSUE #1648: asherpa.astro.ui and fall over when no I/O is provided
I've been looking at improving coverage in #1638 and have found an area I don't want to work on in that PR. There's some support for I/O code to fall back to using the same code as the sherpa.ui code but

a) it doesn't fall back to the sherpa.ui version but re-implements calls, which may or may not be a good idea (I'm leaning towards "not" but there may be a reason why we want to do this)

b) probably because of a, we don't get similar / understandable behavior, which is the reason for this ticket.

This is not high-priority code, as we assume that users of sherpa.astro.io have an I/O module, but as we can support the "no I/O" case we should review this.

I have two 1-column ASCII files that I want to use with `load_table_model`. The only difference is the comment character:

```shell
% cat x.dat
# X
1
2
3
4
%diff x.dat y.dat
1c1
< # X
---
> ! X
```

# Using sherpa.ui

The following shows I can read in these files, and that for `y.dat` I have to pass in an explicit `comment` argument (which makes sense).

```python
>>> from sherpa.ui import load_table_model, get_model_component
>>> load_table_model("mx", "x.dat", ncols=1)
>>> load_table_model("my", "y.dat", ncols=1)
  File "<string>", line 1, in load_table_model
  File "/home/dburke/sherpa/sherpa-pr/sherpa/ui/utils.py", line 7584, in load_table_model
    x, y = self._read_user_model(filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/ui/utils.py", line 7276, in _read_user_model
    data = self.unpack_data(filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/ui/utils.py", line 4514, in unpack_data
    return self._read_data(sherpa.io.read_data, filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/ui/utils.py", line 4348, in _read_data
    return readfunc(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 322, in read_data
    colnames, args, name = get_ascii_data(filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 223, in get_ascii_data
    names, args = read_file_data(filename, sep, comment, require_floats)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 79, in read_file_data
    cols = numpy.column_stack(rows)
  File "<__array_function__ internals>", line 180, in column_stack
  File "/home/dburke/miniconda3/envs/sherpa-pr/lib/python3.10/site-packages/numpy/lib/shape_base.py", line 656, in column_stack
    return _nx.concatenate(arrays, 1)
  File "<__array_function__ internals>", line 180, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 2 and the array at index 1 has size 1
>>> load_table_model("my", "y.dat", ncols=1, comment="!")
>>> get_model_component("mx").get_y()
array([1., 2., 3., 4.])
>>> get_model_component("my").get_y()
array([1., 2., 3., 4.])
```

# sherpa.astro.ui with no I/O module

Basically I 'pip unstall astropy' before doing this... which gives us

```python
>>> from sherpa.astro.ui import load_table_model, get_model_component
WARNING: Cannot import usable I/O backend.
    If you are using CIAO, this is most likely an error and you should contact the CIAO helpdesk.
    If you are using Standalone Sherpa, please install astropy.
```

If I try the same things then all is not well. Let's just try to load in x.dat. This results in a bunch of exceptions which is hard to follow, but the main thin appears to be that the code assumes we have two columns, not 1. Why?

```python
>>> load_table_model("mx", "x.dat", ncols=1)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10246, in load_table_model
    raise Exception("Not a FITS file")
Exception: Not a FITS file

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10050, in _read_user_model
    data = self.unpack_ascii(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 1480, in unpack_ascii
    return sherpa.astro.io.read_ascii(filename, ncols, colkeys, dstype,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/__init__.py", line 291, in read_ascii
    args = backend.get_ascii_data(filename, ncols=ncols, colkeys=colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10061, in _read_user_model
    data = self.unpack_table(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 1286, in unpack_table
    return sherpa.astro.io.read_table(filename, ncols, colkeys, dstype)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/__init__.py", line 231, in read_table
    args = backend.get_table_data(arg, ncols, colkeys)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10257, in load_table_model
    x, y = self._read_user_model(filename, *args,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10074, in _read_user_model
    data = self.unpack_image(filename, *args, **kwargs)
TypeError: Session.unpack_image() got an unexpected keyword argument 'ncols'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in load_table_model
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10263, in load_table_model
    data = sherpa.io.read_data(filename, ncols=2)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 324, in read_data
    return dstype(name, *args)
TypeError: Data1D.__init__() missing 1 required positional argument: 'y'
```

Trying `y,dat` with no `comment` argument actually looks like the sherpa.ui version, which is a bit surprising given the above, but I'll take it (at least once we've gone through the cascading errors):

```python
>>> load_table_model("my", "y.dat", ncols=1)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10246, in load_table_model
    raise Exception("Not a FITS file")
Exception: Not a FITS file

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10050, in _read_user_model
    data = self.unpack_ascii(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 1480, in unpack_ascii
    return sherpa.astro.io.read_ascii(filename, ncols, colkeys, dstype,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/__init__.py", line 291, in read_ascii
    args = backend.get_ascii_data(filename, ncols=ncols, colkeys=colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10061, in _read_user_model
    data = self.unpack_table(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 1286, in unpack_table
    return sherpa.astro.io.read_table(filename, ncols, colkeys, dstype)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/__init__.py", line 231, in read_table
    args = backend.get_table_data(arg, ncols, colkeys)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10257, in load_table_model
    x, y = self._read_user_model(filename, *args,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10074, in _read_user_model
    data = self.unpack_image(filename, *args, **kwargs)
TypeError: Session.unpack_image() got an unexpected keyword argument 'ncols'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in load_table_model
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10263, in load_table_model
    data = sherpa.io.read_data(filename, ncols=2)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 322, in read_data
    colnames, args, name = get_ascii_data(filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 223, in get_ascii_data
    names, args = read_file_data(filename, sep, comment, require_floats)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 79, in read_file_data
    cols = numpy.column_stack(rows)
  File "<__array_function__ internals>", line 180, in column_stack
  File "/home/dburke/miniconda3/envs/sherpa-pr/lib/python3.10/site-packages/numpy/lib/shape_base.py", line 656, in column_stack
    return _nx.concatenate(arrays, 1)
  File "<__array_function__ internals>", line 180, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 2 and the array at index 1 has size 1
```

If I add the `comment` argument I would have expected to match the same error that `x.dat` gave, because in any right-thinking world it would. So, on that note I give you

```python
>>> load_table_model("my", "y.dat", ncols=1, comment="!")
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10246, in load_table_model
    raise Exception("Not a FITS file")
Exception: Not a FITS file

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10050, in _read_user_model
    data = self.unpack_ascii(filename, *args, **kwargs)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 1480, in unpack_ascii
    return sherpa.astro.io.read_ascii(filename, ncols, colkeys, dstype,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/__init__.py", line 291, in read_ascii
    args = backend.get_ascii_data(filename, ncols=ncols, colkeys=colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10061, in _read_user_model
    data = self.unpack_table(filename, *args, **kwargs)
TypeError: Session.unpack_table() got an unexpected keyword argument 'comment'

During handling of the above exception, another exception occurred:

  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10257, in load_table_model
    x, y = self._read_user_model(filename, *args,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10069, in _read_user_model
    y = sherpa.astro.io.backend.get_table_data(filename, *args,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/io/dummy_backend.py", line 44, in get_table_data
    raise NotImplementedError('No usable I/O backend was imported.')
NotImplementedError: No usable I/O backend was imported.

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in load_table_model
  File "/home/dburke/sherpa/sherpa-pr/sherpa/astro/ui/utils.py", line 10263, in load_table_model
    data = sherpa.io.read_data(filename, ncols=2)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 322, in read_data
    colnames, args, name = get_ascii_data(filename, ncols, colkeys,
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 223, in get_ascii_data
    names, args = read_file_data(filename, sep, comment, require_floats)
  File "/home/dburke/sherpa/sherpa-pr/sherpa/io.py", line 79, in read_file_data
    cols = numpy.column_stack(rows)
  File "<__array_function__ internals>", line 180, in column_stack
  File "/home/dburke/miniconda3/envs/sherpa-pr/lib/python3.10/site-packages/numpy/lib/shape_base.py", line 656, in column_stack
    return _nx.concatenate(arrays, 1)
  File "<__array_function__ internals>", line 180, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 2 and the array at index 1 has size 1
```

It is the **same** error as when I didn't set the `comment` keyword.

That's not good.

And it's because somehow the `comment` argument (ie the `kwargs` the user originally defined) have somehow been lost in the chain of cascading errors.

---

# Thoughts

Even if we go with option a it looks like

1. we should review adding some judicious use of `from exc` statements (or possibly `from None` but in this case I think it could lose information)
2. review argument passing

Or we could just require an I/O module to be present, but I'm not ready to go there. Yet.

hamogu:
## Summary
Split the `TableModel` into two separate models: `FixedTableModel` for a model where the number of bins is fixed to be the same number as the data values and `InterpolatedTableModel1D` where a tabulated x/y table is interpolated (and integrated if needed) to match the data values/bins.

## Details

The old `TableModel` has two separate applications: It can be an x/y table that is interpolated in x as needed, or it can just be a table of y values assuming that the number of values exactly matches the number of data values.
    
This is conceptually different and can be confusing to users (and also in the code). This commit introduces two new classes to split the two different functionalities into separate classes (with a common base class).

Adding those two new classes is followed by:

- deprecating `TableModel`
- Adjusting the models generated by UI functions that load a table

in doing so, this includes the changes in #2376, which change the loading functions in a way that is technically breaking API, but in practice I judge that to be a minimal impact which allows for significantly more consistency in the API (I would just treat the change as a "bug fix"):  align the signature of the `load_table_model` in `sherpa.ui` and `sherpa.astro.ui`. This is technically an API breaking change in case someone relied on the ordering of arguments and passed in the interpolation method as third argument instead of a keyword argument.

I think that's unlikely to happen in practice (most people won't pass that in at all and just use the default), but it's separated into it's own commit in case we want to discuss or reverse that change.

Matching the API does not only make the linters happy, it also simplifies documentation as we could now point to (or generate from in some way) the docs of this method from the docs of the superclass method.

In fact, I'm removing keywords that are no longer used, because they are confusing. Again, I don't think any of them are used in practice, but to keep backwards compatibility, we could just keep those unused keywords in the signature and raise a DeprecationWarning if they are set.

Also, this changes some error messages. It raises the same type of exceptions as before (IoErr) but with a different string message. Again, that's technically breaking the API, but I don't think it changes any workflow in practice.

## Questions

We could keep more of API backwards compatible (a re-ordering of keywords is required, but we can keep the old keywords around (in a differenet order) to blunt the impact).

closes #1648
closes #1620
fixes #1056
closes #2376

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -360,8 +360,8 @@ def test_ui_table_model_1d_testfile(clean_astro_ui, comment, ncols):
         f.write(f'{comment}\n1\n2\n3\n')
         f.seek(0)
         ui.load_table_model('tbl', f.name, comment=comment, ncols=ncols)
-        assert tbl.get_x() is None
         assert tbl.get_y() == pytest.approx([1, 2, 3])
+        assert not hasattr(tbl, 'get_x')
 
 
 def test_ui_table_model_1d_testfile_fails(clean_astro_ui):
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2012, 2015 - 2024
+#  Copyright (C) 2012, 2015-2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -22,6 +22,7 @@ from collections import namedtuple
 import os
 import re
 import logging
+from tempfile import NamedTemporaryFile
 
 import numpy as np
 from numpy.testing import assert_allclose
@@ -343,6 +344,39 @@ def test_ui_table_model_fits_table(setup_files):
 @requires_data
 def test_ui_table_model_fits_image(setup_files):
     ui.load_table_model('tbl', setup_files.img)
+
+
+@pytest.mark.parametrize("comment", ['#', '!'])
+@pytest.mark.parametrize("ncols", [1, 2])
+def test_ui_table_model_1d_testfile(clean_astro_ui, comment, ncols):
+    """Test table model with simple 1D ASCII files
+
+    This should work for any IO backend, because it falls through
+    to the sherpa.io ASCII reader is necessary.
+
+    Regression test for https://github.com/sherpa/sherpa/issues/1648
+    """
+    with NamedTemporaryFile(suffix='.dat', mode='w+') as f:
+        f.write(f'{comment}\n1\n2\n3\n')
+        f.seek(0)
+        ui.load_table_model('tbl', f.name, comment=comment, ncols=ncols)
+        assert tbl.get_x() is None
+        assert tbl.get_y() == pytest.approx([1, 2, 3])
+
+
+def test_ui_table_model_1d_testfile_fails(clean_astro_ui):
+    """Test table model with simple 1D ASCII files
+
+    Here the comment character is ! in the file, while the default #
+    is used for the read, so no data lines are found.
+
+    Regression test for https://github.com/sherpa/sherpa/issues/1648
+    """
+    with NamedTemporaryFile(suffix='.dat', mode='w+') as f:
+        f.write(f'!\n1\n2\n3\n')
+        f.seek(0)
+        with pytest.raises(IOErr, match="No column data found in "):
+            ui.load_table_model('tbl', f.name)
 
 
 # Test user model
--- a/sherpa/astro/ui/tests/test_astro_ui.py
+++ b/sherpa/astro/ui/tests/test_astro_ui.py
@@ -352,7 +352,7 @@ def test_ui_table_model_1d_testfile(clean_astro_ui, comment, ncols):
     """Test table model with simple 1D ASCII files
 
     This should work for any IO backend, because it falls through
-    to the sherpa.io ASCII reader is necessary.
+    to the sherpa.io ASCII reader if necessary.
 
     Regression test for https://github.com/sherpa/sherpa/issues/1648
     """
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -59,7 +59,7 @@ from sherpa.data import Data, Data1D, Data1DAsymmetricErrs, Data2D, \
     Data2DInt
 from sherpa.fit import Fit
 import sherpa.io
-from sherpa.models.basic import TableModel, UserModel
+from sherpa.models.basic import UserModel
 from sherpa.models.model import Model
 import sherpa.plot
 from sherpa.sim import NormalParameterSampleFromScaleMatrix, \
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -29,7 +29,6 @@ import os
 from typing import Any
 import sys
 from types import MethodType
-import warnings
 
 import numpy as np
 
@@ -11212,7 +11211,6 @@ class Session(sherpa.ui.utils.Session):
     # (such as fits files) than the super method (only ascii files) and
     # thus we do define a method here just so we can write a new docstring.
     # DOC-NOTE: can filename be a crate/hdulist?
-    # DOC-TODO: how to describe the supported args/kwargs (not just for this function)?
     def load_table_model(self, modelname, filename,
                          *args,
                          **kwargs) -> None:
@@ -11226,6 +11224,13 @@ class Session(sherpa.ui.utils.Session):
         The ``load_xstable_model`` routine must be used with XSPEC table
         models.
 
+        Tables can take many forms. The most common are 1D data (with 2 columns),
+        1D data with integrated bins (wiht colunns xlow, xhigh, y), and 2D data such
+        as an image. This function will try to infer the format of the data from
+        the structure of the file given.
+        Details also depend on the file tome (fits or ascii) and the backend in use
+        (crates, astropy, or just the build in ascii reader).
+
         Parameters
         ----------
         modelname : str
@@ -11234,10 +11239,17 @@ class Session(sherpa.ui.utils.Session):
            The name of the file containing the data, which should
            contain two columns, which are the x and y values for
            the data, or be an image.
-        args
-           Arguments for reading in the data.
+        method : func, optional
+           The interpolation method to use to map the input data onto the
+           coordinate grid of the data set.
+           Sherpa provides `~sherpa.utils.linear_interp`,
+           `~sherpa.utils.nearest_interp`, `~sherpa.utils.neville`, and
+           `~sherpa.utils.neville2d` in `sherpa.utils` but other functions
+           with the same signature can be used.
         kwargs
-           Keyword arguments for reading in the data.
+           The accepted keyword arguments depend on the file type and
+           backend used to read in the data. They are listed in the "notes" section below.
+
 
         See Also
         --------
@@ -11250,14 +11262,63 @@ class Session(sherpa.ui.utils.Session):
 
         Notes
         -----
-        Examples of interpolation schemes provided by `sherpa.utils`
-        are: `linear_interp`, `nearest_interp`, `neville`, and
-        `neville2d`.
+
+        * Keywords for ASCII tables*
+
+            ncols : int, optional
+                The number of columns to read in (the first ``ncols`` columns
+                in the file). This is ignored if ``colkeys`` is given.
+            colkeys : array of str, optional
+                An array of the column name to read in. The default is `None`.
+            sep : str, optional
+                The separator character. The default is ``' '``.
+            dstype : data class to use, optional
+                Used to check that the data file contains enough columns.
+            comment : str, optional
+                The comment character. The default is ``'#'``.
+            require_floats : bool, optional
+                If `True` (the default), non-numeric data values will
+                raise a `ValueError`.
+
+        * Keywords for FITS tables
+
+            ncols: int, optional
+                The number of columns to read in when colkeys is not
+                set (the first ncols columns are chosen).
+            colkeys: sequence of str or None, optional
+                If given, what columns from the table should be selected,
+                otherwise the backend selects. The default is `None`.
+                Names are compared using a case insensitive match.
+            make_copy: bool, optional
+                If set then the returned NumPy arrays are explicitly copied,
+                rather than using a reference from the data structure
+                created by the backend. Backends are not required to
+                honor this setting. The default is `True`.
+            fix_type: bool, optional
+                Should the returned arrays be converted to the
+                `sherpa.utils.numeric_types.SherpaFloat` type. The default
+                is `True`.
+            blockname: str or None, optional
+                The name of the "block" (HDU) to read the column data (useful
+                for data structures containing multiple blocks/HDUs) or None.
+                Names are compared using a case insensitive match.
+            hdrkeys: sequence of str or None, optional
+                If set, the table structure must contain these keys, and the
+                values are returned.  Names are compared using a case
+                insensitive match.
+
+        * Keywords for FITS images
+
+            coord : { 'logical', 'image', 'physical', 'world', 'wcs' }, optional
+                Ensure that the image contains the given coordinate system.
+            dstype : optional
+                The image class to use. The default is `~sherpa.astro.data.DataIMG`.
+
 
         Examples
         --------
 
-        Load in the data from filt.fits and use it to multiply
+        Load in the data from "filt.fits" and use it to multiply
         the source model (a power law and a gaussian). Allow
         the amplitude for the table model to vary between 1
         and 1e6, starting at 1e3.
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -23,10 +23,13 @@ from __future__ import annotations
 from collections.abc import Callable, Sequence
 from contextlib import suppress
 from dataclasses import dataclass
+from functools import partial
 import logging
 import os
 from typing import Any
 import sys
+from types import MethodType
+import warnings
 
 import numpy as np
 
@@ -1368,7 +1371,7 @@ class Session(sherpa.ui.utils.Session):
             return sherpa.astro.io.read_arrays(*args)
         except NotImplementedError:
             # if the astro backend is not set, fall back on io module version.
-            return sherpa.io.read_arrays(*args)
+            return super().unpack_arrays(*args)
 
     # DOC-NOTE: also in sherpa.utils
     # DOC-TODO: rework the Data type notes section (also needed for
@@ -11069,43 +11072,46 @@ class Session(sherpa.ui.utils.Session):
         self._background_sources.get(idval, {}).pop(bkg_id, None)
 
     def _read_user_model(self, filename, *args, **kwargs):
-        x = None
-        y = None
-        try:
-            data = self.unpack_ascii(filename, *args, **kwargs)
-            x = data.get_x()
-            y = data.get_y()
+        """Read in user model data from a file.
 
-        # we have to check for the case of a *single* column in an ascii file
-        # extract the single array from the read and bypass the dataset
-        except TypeError:
-            hdu, _ = sherpa.astro.io.backend.get_ascii_data(filename,
-                                                            *args,
-                                                            **kwargs)
-            y = hdu.columns[0].values
-        except Exception:
+        While some options can be set by keyword arguments,
+        fundamentally, this is trying out a list of possible formats
+        (ASCII with the backend, general table (e. g. fits), image)
+        until one works.
+        If no working backend is loaded, this falls back to the
+        sherpa.ui version, which can read ascii files.
+        """
+        def read_backend(func, filename, *args, **kwargs):
+            hdu, _ = func(filename, *args, **kwargs)
+            cols = hdu.columns
+            if len(cols) == 0:
+                raise IOErr(f"No column data found in {filename}")
+            if len(cols) == 1:
+                return (None, cols[0].values)
+            return (cols[0].values, cols[1].values)
+
+        read_backend_ascii = partial(read_backend,
+                                     sherpa.astro.io.backend.get_ascii_data)
+        read_backend_table = partial(read_backend,
+                                     sherpa.astro.io.backend.get_table_data)
+
+        def read_backend_image(self, filename, *args, **kwargs):
+            data = self.unpack_image(filename, *args, **kwargs)
+            return (None, data.get_y())
+
+        def read_with_super(self, filename, *args, **kwargs):
+            return super()._read_user_model(filename, *args, **kwargs)
+
+        for func in (read_backend_ascii,
+                     read_backend_table,
+                     # The following two need "self", so we make them a method
+                     MethodType(read_backend_image, self),
+                     MethodType(read_with_super, self)):
             try:
-                data = self.unpack_table(filename, *args, **kwargs)
-                x = data.get_x()
-                y = data.get_y()
-
-            # we have to check for the case of a *single* column in a
-            # fits table
-            # extract the single array from the read and bypass the dataset
-            except TypeError:
-                hdu, _ = sherpa.astro.io.backend.get_table_data(filename,
-                                                                *args,
-                                                                **kwargs)
-                y = hdu.columns[0].values
-
-            except Exception:
-                # unpack_data doesn't include a call to try
-                # getting data from image, so try that here.
-                data = self.unpack_image(filename, *args, **kwargs)
-                # x = data.get_x()
-                y = data.get_y()
-
-        return (x, y)
+                return func(filename, *args, **kwargs)
+            except:
+                pass
+        raise IOErr(f"No column data found in {filename} with supported file formats")
 
     def load_xstable_model(self, modelname, filename,
                            etable=False) -> None:
@@ -11200,11 +11206,15 @@ class Session(sherpa.ui.utils.Session):
         self._tbl_models.append(tablemodel)
         self._add_model_component(tablemodel)
 
-    # also in sherpa.utils
+    # implementation just calls super, but that calls `self._read_user_model`
+    # which is overwritten in this class.
+    # As a consequence, this method accepts a wider range of inputs
+    # (such as fits files) than the super method (only ascii files) and
+    # thus we do define a method here just so we can write a new docstring.
     # DOC-NOTE: can filename be a crate/hdulist?
     # DOC-TODO: how to describe the supported args/kwargs (not just for this function)?
     def load_table_model(self, modelname, filename,
-                         method=sherpa.utils.linear_interp, *args,
+                         *args,
                          **kwargs) -> None:
         # pylint: disable=W1113
         """Load tabular or image data and use it as a model component.
@@ -11224,11 +11234,6 @@ class Session(sherpa.ui.utils.Session):
            The name of the file containing the data, which should
            contain two columns, which are the x and y values for
            the data, or be an image.
-        method : func
-           The interpolation method to use to map the input data onto
-           the coordinate grid of the data set. Linear,
-           nearest-neighbor, and polynomial schemes are provided in
-           the sherpa.utils module.
         args
            Arguments for reading in the data.
         kwargs
@@ -11268,22 +11273,7 @@ class Session(sherpa.ui.utils.Session):
         >>> set_source('img', emap * gauss2d)
 
         """
-
-        x = None
-        y = None
-        try:
-            x, y = self._read_user_model(filename, *args, **kwargs)
-        except Exception:
-            data = sherpa.io.read_data(filename, ncols=2)
-            x = data.x
-            y = data.y
-
-        tablemodel = TableModel(modelname)
-        tablemodel.method = method
-        tablemodel.filename = filename
-        tablemodel.load(x, y)
-        self._tbl_models.append(tablemodel)
-        self._add_model_component(tablemodel)
+        super().load_table_model(modelname, filename, *args, **kwargs)
 
     # ## also in sherpa.utils
     # DOC-TODO: how to describe *args/**kwargs
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -11225,11 +11225,11 @@ class Session(sherpa.ui.utils.Session):
         models.
 
         Tables can take many forms. The most common are 1D data (with 2 columns),
-        1D data with integrated bins (wiht colunns xlow, xhigh, y), and 2D data such
+        1D data with integrated bins (with columns xlow, xhigh, y), and 2D data such
         as an image. This function will try to infer the format of the data from
         the structure of the file given.
-        Details also depend on the file tome (fits or ascii) and the backend in use
-        (crates, astropy, or just the build in ascii reader).
+        Details also depend on the file type (fits or ascii) and the backend in use
+        (crates, astropy, or just the build-in ascii reader).
 
         Parameters
         ----------
@@ -11263,7 +11263,7 @@ class Session(sherpa.ui.utils.Session):
         Notes
         -----
 
-        * Keywords for ASCII tables*
+        * Keywords for ASCII tables
 
             ncols : int, optional
                 The number of columns to read in (the first ``ncols`` columns
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -1369,7 +1369,6 @@ class Session(sherpa.ui.utils.Session):
         try:
             return sherpa.astro.io.read_arrays(*args)
         except NotImplementedError:
-            # if the astro backend is not set, fall back on io module version.
             return super().unpack_arrays(*args)
 
     # DOC-NOTE: also in sherpa.utils
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -21,6 +21,7 @@
 from abc import ABCMeta
 from collections.abc import Callable, Sequence
 import logging
+from warnings import warn
 
 import numpy
 
@@ -2020,6 +2021,10 @@ class TableModel(ArithmeticModel):
         self._method = linear_interp
         self.ampl = Parameter(name, 'ampl', 1)
         ArithmeticModel.__init__(self, name, (self.ampl,))
+        warn("TableModel is deprecated and will be removed in a future version. "
+             "Use FixedTableModel (if only setting y values) "
+             "or InterpolatedTableModel1D (if setting both x and y values).",
+             category=DeprecationWarning)
 
     def load(self, x, y):
         """Set the model values.
--- a/sherpa/models/basic.py
+++ b/sherpa/models/basic.py
@@ -18,12 +18,14 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+from abc import ABCMeta
+from collections.abc import Callable, Sequence
 import logging
 
 import numpy
 
-from sherpa.utils import bool_cast, interpolate, linear_interp, \
-    sao_fcmp
+from sherpa.utils import bool_cast, integrate_tabulated_function, interpolate, linear_interp, \
+    integrate, sao_fcmp
 from sherpa.utils.err import ModelErr
 from sherpa.utils.guess import get_position, guess_amplitude, \
     guess_amplitude_at_ref, guess_amplitude2d, guess_bounds, \
@@ -46,6 +48,7 @@ __all__ = ('Box1D', 'Const1D', 'Cos', 'Delta1D', 'Erf', 'Erfc', 'Exp', 'Exp10',
            'StepLo1D', 'Tan',
            'Box2D', 'Const2D', 'Delta2D', 'Gauss2D', 'SigmaGauss2D',
            'NormGauss2D', 'Polynom2D', 'Scale2D', 'UserModel', 'TableModel',
+           'FixedTableModel', 'InterpolatedTableModel1D',
            'Integrate1D')
 
 DBL_EPSILON = numpy.finfo(float).eps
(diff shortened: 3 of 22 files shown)


hamogu:
Test failure is unrelated; the failing test probably just needs a slightly larger numerical range to pass on all Python and numpy versions and all OS that we test. I'll address that in a separate PR, but for now, please assume that this is ready to review.

hamogu:
rebased onto https://github.com/sherpa/sherpa/pull/2453 as it changes the DS9 download code (and updates versions).

@DougBurke if you could find time to review this, that would be great!

sherpa/models/basic.py
```diff
@@ -2184,6 +2187,397 @@ def calc(self, p, x0, x1=None, *args, **kwargs):
                        f"data, ({len(self.__y)} vs {len(x0)})")
 
 
+
+class TableModelBase(ArithmeticModel, metaclass=ABCMeta):
```
DougBurke, line 2199: Are we "safe" using a  `metaclass` here? Two points

1. this is in the middle of the class hierarchy (i.e. this is not the base class); is this okay (e.g. the `Model` class is derived from `NoNewBlahBlah` and I thought this would be an issue, but I could be wrong)
2. I have previously seen some surprising behavior after some work where I tried moving to `super()` rather than calling the parent class directly, which has made me wary of this (unfortunately I can't remember the details). In this case it's probably okay because we don't expect there to be many complicated classes that derive from `TableModelBase`

hamogu: I'm pretty sure it's OK. First, the fact that all the old and new tests are passing is a good sign. 
More specifically:

1. We can add meta classes at any level and a class can have multiple metaclasses. That doesn't change the applicability of the other metaclasses, unless it implements methods of the same name. See code example below.
2. In simple inheritance, `super` and directly calling the superclass is interchangeable. Differences arise when you have a diamond structure or something even more complicated because `super` might then lead to a different call order than selecting which superclass to call directly. That's not an issue here. `super()` will call to `ArithmeticModel` because it's the only normal (not meta) class in the inheritance scheme. (It might be possible to screw this up if you derive a class from `TableModelBase` that also has some other child of `ArithmeticModel` as a parent. But that's (a) unlikely to make any sense and (b) we don't know which behavior would be preferred in that case so I'm not worrying about it.) We could replace "super" with "ArthimeticModel" here and it won't make a difference, if you think that reads better.

Example showing that `NoNewBlaBlah` still works:
```Python
>>> from sherpa.models.basic import InterpolatedTableModel1D
>>> tm = InterpolatedTableModel1D()
>>> tm.load([10, 20, 25, 30], [14, 12, 17, 18])
>>> tm.ampl = 10
>>> tm([15, 20, 27])
array([130., 120., 174.])
>>> tm.sergasdf = 4
Cell In[4], [line 1](vscode-notebook-cell:?execution_count=4&line=1)
----> [1](vscode-notebook-cell:?execution_count=4&line=1) tm.sergasdf = 4

File ~/code/sherpa/sherpa/models/model.py:711, in Model.__setattr__(self, name, val)
    708     par.val = val
    709     return
--> [711](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/debugging/~/code/sherpa/sherpa/models/model.py:711) NoNewAttributesAfterInit.__setattr__(self, name, val)
    713 # If adding a parameter, add the entry to _par_index and
    714 # handle any aliases.
    715 #
    716 if not isinstance(val, Parameter):

File ~/code/sherpa/sherpa/utils/__init__.py:133, in NoNewAttributesAfterInit.__setattr__(self, name, val)
    131 def __setattr__(self, name: str, val: Any) -> None:
    132     if self.__initialized and (not hasattr(self, name)):
--> [133](https://file+.vscode-resource.vscode-cdn.net/Users/guenther/code/sherpa/debugging/~/code/sherpa/sherpa/utils/__init__.py:133)         raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
    135     if self.__initialized and hasattr(self, name):
    136         cname = callable(getattr(self, name))

AttributeError: 'InterpolatedTableModel1D' object has no attribute 'sergasdf'
```

sherpa/models/tests/test_tablemodel.py
```diff
@@ -31,68 +31,82 @@
...
 
-def test_ndim():
+@pytest.mark.parametrize("cls, ndim", [(TableModel, None),
+                                         (FixedTableModel, None),
+                                         (InterpolatedTableModel1D, 1)])
+def test_ndim(cls, ndim):
     """What is the ndim?
 
     At the moment we can use this with nD data but only really if we do
     not set the "x" argument.
```
DougBurke:
```suggestion
```

As the code now kind-of sets ndim the comment is now out-of-date and I'm not sure it's worth adding anything to describe the current behavior.

sherpa/models/basic.py
```diff
@@ -2184,6 +2187,397 @@ def calc(self, p, x0, x1=None, *args, **kwargs):
...
+              len(x0) == len(self._filtered_y)):
+            return p[0] * self._filtered_y
+
+        if len(x0) == len(self._y):
+            return p[0] * self._y
+
+        raise ModelErr("filtermismatch", 'table model',
+                       f"data, ({len(self._y)} vs {len(x0)})")
+
+class InterpolatedTableModel1D(TableModelBase, RegriddableModel1D):
```
DougBurke: What benefit do we have from making this part of `RegriddableModel1D`? 

The regrid stuff was integrated into sherpa without an obvious discussion [*] of the table model class, which already had some form of "regrid" capabilities - e.g. the `calc` method below. So I see two possibilities

- can we use the regrid code to avoid some of the code here 
- if not, then I am concerned about having two different ways to get the same [**] thing done, which could be confusing


---

[*] well, I don't remember it, which doesn't actually mean it wasn't discussed

[**] of course, they aren't quite the same thing

sherpa/astro/ui/utils.py
```diff
@@ -11224,11 +11234,6 @@ def load_table_model(self, modelname, filename,
            The name of the file containing the data, which should
            contain two columns, which are the x and y values for
            the data, or be an image.
-        method : func
```
DougBurke, line 11227: [i have many words that could be said about documenting python code in the presence of kwargs, so some "thoughts" may skip through here. This is definitely based on having to dig through matplotlib documentation hunting down arguments]

In this case the use of `kwargs` lets us drop some repeated code. However, for the user it is, I claim, a demonstrably poorer experience, as the have no idea about the `method` argument or its default value. We don't even have an example showing it (not that having an example would absolve us, I think).

I'm not sure of the best thing to do here. Matlpotlib can (in some cases at least) point to other documentation when describing the `kwargs` argument, but that's not always possible, or makes sense here. One option is to leave in just the documentation. Another would be to not remove the `method=...` argument above. Or leave as is.

hamogu: The DOC has a longstanding "DOC-TODO: how to describe the supported args/kwargs (not just for this function)?" comment above its signature indicating that this problem isn't new. What's new is that this function is now fully redundant. It doesn't do anything but calling the super class and is only here so we get a chance to write down a docstring. I'll expand that to be more useful.

sherpa/models/basic.py
hamogu: I don't think we strictly speaking need it. At least our tests pass without inheriting from `RegriddleModel1D`. I put it there because I think that's what is conceptually does - regriddable models promise to be regriddable, which this model is. The `Regriddable_XXX` classes are actually very small and have only one function (`regrid`) in them, so there isn't any code that could be reused here and I'm not concerned about having two ways to do something that is not the same thing.

sherpa/astro/ui/tests/test_astro_ui.py
```diff
@@ -22,6 +22,7 @@
 import os
 import re
 import logging
+from tempfile import NamedTemporaryFile
```
DougBurke, line 25: An alternative to using `NamedTemporaryFile` is the [tmp_path](https://docs.pytest.org/en/latest/how-to/tmp_path.html) fixture but I'm not sure it's worth changing here.

sherpa/models/basic.py
```diff
@@ -2017,6 +2021,10 @@ def __init__(self, name='tablemodel'):
         self._method = linear_interp
         self.ampl = Parameter(name, 'ampl', 1)
         ArithmeticModel.__init__(self, name, (self.ampl,))
+        warn("TableModel is deprecated and will be removed in a future version. "
```
DougBurke, line 2027: We should add a docstirng at line 1888 ish along the lines of (with left padding added)

```
... deprecated:: 4.19.0
    Use the FixedTableModel or InterpolatedTableModel1D classes instead.
```

DougBurke:
We need somehting like

```
% git diff .
diff --git a/docs/model_classes/basic.rst b/docs/model_classes/basic.rst
index c6c33f12..479acb78 100644
--- a/docs/model_classes/basic.rst
+++ b/docs/model_classes/basic.rst
@@ -19,8 +19,10 @@ The sherpa.models.basic module
       Erfc
       Exp
       Exp10
+      FixedTableModel
       Gauss1D
       Integrate1D
+      InterpolatedTableModel1D
       Log
       Log10
       LogParabola
@@ -34,6 +36,7 @@ The sherpa.models.basic module
       StepHi1D
       StepLo1D
       TableModel
+      TableModelBase
       Tan
       UserModel
 
@@ -49,5 +52,5 @@ The sherpa.models.basic module
 Class Inheritance Diagram
 =========================
 
-.. inheritance-diagram:: Box1D Const1D Scale1D Cos Delta1D Erf Erfc Exp Exp10 Gauss1D Integrate1D Log Log10 LogParabola NormGauss1D Poisson Polynom1D PowLaw1D Sin Sqrt StepHi1D StepLo1D TableModel Tan UserModel Box2D Const2D Scale2D Delta2D Gauss2D SigmaGauss2D NormGauss2D Polynom2D
+.. inheritance-diagram:: Box1D Const1D Scale1D Cos Delta1D Erf Erfc Exp Exp10 FixedTableModel Gauss1D Integrate1D InterpolatedTableModel1D Log Log10 LogParabola NormGauss1D Poisson Polynom1D PowLaw1D Sin Sqrt StepHi1D StepLo1D TableModel TableModelBase Tan UserModel Box2D Const2D Scale2D Delta2D Gauss2D SigmaGauss2D NormGauss2D Polynom2D
    :parts: 1
```

to add the new classes to the Read-The-Docs documentation (although when I just tried it sphinx warned me about

```
/Users/burked/code/sherpa/sherpa-1/sherpa/models/basic.py:docstring of sherpa.models.basic.InterpolatedTableModel1D.method:1: WARNING: duplicate object description of sherpa.models.basic.InterpolatedTableModel1D.method, other instance in model_classes/api/sherpa.models.basic.InterpolatedTableModel1D, use :no-index: for one of the
```

which I have no energy to investigate just now.

There is the question of whether the `TableModelBase` class should be included, as it isn't an exported symbol, but I err on the case of including it.

sherpa/astro/ui/utils.py
```diff
@@ -11245,14 +11262,63 @@ def load_table_model(self, modelname, filename,
 
         Notes
         -----
-        Examples of interpolation schemes provided by `sherpa.utils`
-        are: `linear_interp`, `nearest_interp`, `neville`, and
-        `neville2d`.
+
+        * Keywords for ASCII tables*
```
DougBurke:
```suggestion
        * Keywords for ASCII tables
```

or did you mean to have the asterix here? It looks a bit odd in the sphinx doc build: e.g.

Image

hamogu: No, I only wanted the asterisk in the front (which makes it a bulleted list). I experimented with headings (which I couldn't get sphinx to render at the right level, presumably because it's inside a numpy-style docstring), bold font, numbered and bulleted lists. None is quite perfect for this, but I settled on the list. The extra star must have slipped though from one of the other experiments.

sherpa/astro/ui/tests/test_astro_ui.py
hamogu: They are functionally identical as far as I can tell and I don't care which one to use. If you don't have a preference, let's just leave it as is.
(Maybe "temp_path" is from a time when the "NameTemporaryFile" wasn't in the standard library yet or it offers additional logging for debugging or something like that?)

sherpa/astro/ui/utils.py
DougBurke: I feel your pain for the "trying to work out how to get the documentation look nice" attempt...

hamogu:
> /Users/burked/code/sherpa/sherpa-1/sherpa/models/basic.py:docstring of sherpa.models.basic.InterpolatedTableModel1D.method:1: WARNING: duplicate object description of sherpa.models.basic.InterpolatedTableModel1D.method, other instance in model_classes/api/sherpa.models.basic.InterpolatedTableModel1D, use :no-index: for one of the

This seems to be a known limitation of autodoc that gets confused when the name of a property and the name of a parameter to `__init__` for a class are identical. While I have some general thoughts about autodoc that's beyond the scope if this PR. The parameter is new. In the old table class the only way to change it was to set the property later. I don't like that pattern in general, because it's annoying to make a model and then change it if you want a different default. But since it's new, we don't have to worry about backwards compatibility and I'll just change the name of the parameter.

sherpa/models/basic.py
```diff
@@ -2472,21 +2477,24 @@ class InterpolatedTableModel1D(TableModelBase):
     ndim = 1
 
     def __init__(self, name: str='interpolatedtablemodel1d',
-                 x: Sequence | None=None, y: Sequence | None=None):
+                 x: Sequence | None=None, y: Sequence | None=None,
+                 method: Callable = linear_interp,
+                 integrate_method: Callable = integrate_tabulated_function,
+                 integrate_kwargs: dict = {}):
```
DougBurke: Should this default to `None` - which we then replace with the empty `dict` ? There is the whole "do not use a mutable argument" that `pylint` complains about (but I note `ruff` does not):

https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/dangerous-default-value.html

hamogu: That's a good point. Mutable default argument make for hard-to-debug error messages. Will change.

DougBurke:
Note that the RTD failure is fixed in the main branch (#2473) and it is completely unrelated to this PR

sherpa/astro/ui/utils.py
```diff
@@ -1368,7 +1370,7 @@ def unpack_arrays(self, *args):
             return sherpa.astro.io.read_arrays(*args)
         except NotImplementedError:
             # if the astro backend is not set, fall back on io module version.
```
DougBurke: This comment is not out of date (as the fall back is no longer the `sherpa.io.read_arrays` call). We could fix it but I think it's obvious from the context so lnie 1372 can just be removed.

```suggestion
```

sherpa/models/basic.py
```diff
@@ -18,12 +18,15 @@
...
+from collections.abc import Callable, Sequence
 import logging
+from warnings import warn
 
 import numpy
 
-from sherpa.utils import bool_cast, interpolate, linear_interp, \
-    sao_fcmp
+from sherpa.utils import bool_cast, integrate_tabulated_function, interpolate, linear_interp, \
+    integrate, sao_fcmp
```
DougBurke, line 29:
```suggestion
    sao_fcmp
```

I don't think we use the `integrate` symbol, but I could be mistaken.

sherpa/utils/__init__.py
```diff
@@ -34,7 +34,7 @@
 import string
 import sys
 from types import FunctionType, MethodType
-from typing import Any, Generic, TypeVar, overload
+from typing import Any, Dict, Generic, TypeVar, overload
```
DougBurke:
```suggestion
from typing import Any, Generic, TypeVar, overload
```

We don't need this anymore.

sherpa/utils/__init__.py
```diff
@@ -2344,6 +2345,121 @@ def interpolate(xout, xin, yin, function=linear_interp):
...
+        ind0 = np.nonzero(x == x0[i])[0][0]
+        ind1 = np.nonzero(x == x1[i])[0][0]
+        result[i] = area[ind0:ind1].sum()
+
+    return result
+
+
+def integrate(x0: np.ndarray, x1: np.ndarray, xin: np.ndarray, yin: np.ndarray,
+              interpolate_func: Callable[[np.ndarray, np.ndarray, np.ndarray], np.ndarray] = linear_interp,
+              integrate_func: Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray, Callable, Dict], np.ndarray] = integrate_tabulated_function,
```
DougBurke:
```suggestion
              integrate_func: Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray, Callable, dict | None], np.ndarray] = integrate_tabulated_function,
```

As mentioned above, we don't need `Dict`. I'm pretty sure (but have not checked as much as I should have) that the "mutable fix" you made today means that the "`| None`" change is now also needed.

sherpa/ui/utils.py
```diff
@@ -7950,7 +7943,7 @@ def func2d(pars, x0, x1, x0hi=None, x1hi=None):
         usermodel._file = filename
         if filename is not None:
             _, usermodel._y = self._read_user_model(filename, ncols, colkeys,
-                                                    dstype, sep, comment)
+                                                    sep, comment)
```
DougBurke, line 7946: I think this is the largest conceptual issue I have here. As far as I can see we now longer let the user pass through a `Data2D` or `Data2Dint` case.

I am 99.99% sure we have no tests of this functionality (definitely not in the Sherpa test suite, and probably not in CIAO itself) so it could well be that the code never worked sensibly.

DougBurke:
So, I've recently been looking at out C API and integration, and I wonder whether it's relevant here. However, it is completely undocumented and is all about passing around C routines between modules and is used by `sherpa.models.basic.Integrator1D`. 

So there's possibility some clean up that could be done later to try and unify some of this "integrate" code.

sherpa/ui/utils.py
hamogu: Yes, this is a deliberate decision (which iI intended to put into its own commit so we could reverse it but it turned out that there is no reasonable way to carry that through with the other changes in this PR). Also see discussion on #2376.

The short summary is that, as fas as I can tell, that never really worked in general, was never tested, and never advertised. The basic problem is that there that all those formats look too similar, i.e. a 2D (unwrapped as above) looks exactly like a 1DInt (x_lo, x_hi, y) in the file format. 

\<rant\>
There is a web (literally, I drew it out on the piece of paper) of private and partially hidden  functions like the `_read_user_model` that is called from the public `load_user_model` and many of those involve heuristics and guessing to identify what the format of the file should be, even when a specific data class is passed. There are just too many options. Even a simple Data1D can have anywhere between two (x, y) and four columns (x, y, stat_error, sys_error). How do you distinguish between that and Data2DInt with three to five columns or any of the datasets in between in terms of complexity (Data1DInt, Data2D)? Now, you might say "that's exactly why we had that keyword", but in fact the code is not written that way. Even with that keyword passed in, the underlying code is still making guesses about how many columns there are and what they mean. 
`_read_user_model` used to have multiple try..except blocks nested to encompass that, which I've changed to a simpler loop, but the concept is the same (which is why all old and new tests pass).
It's like a design that contains only edge cases; it's what an LLM would have written if it had been around 10 years ago and the instructions were given by monkeys who also don't know what they want. Ok, I'm joking for the last one: The truth is that the programmer of that code probably wanted to make it work out-of-the-box for almost any format the user throws at it.
We want it to "just work" (TM)! 
And then some other programmer tried the same for the `astro.ui.utils`, with slightly different assumptions. So we have a second version there, that's almost, but not quite giving the same answers leading to subtle inconsistencies that have driven me crazy in debugging before (which resulting in #2376).
But if you make it work so flexibly for almost anything that you throw at it, then it can't be specific at the same time. The vast majority of use cases (I wager about 101 % of them) will do this for some type of 1D data (or  data we thing of as 2SD in our head, but that's unwrapped in Sherpa so that it's Data1D or Data1DInt). I've looked long and hard when I did those changes, because I know about the pain of backwards braking changes, and I cannot see a way to make a sensible use of the `dstype` keyword in the structure that we have.
(You could write essentially separate functions for 1D and 2D but unless we have a use case for that, we shouldn't add that complexity. And even when we have a use case, we should suggest a more direct way to do that.)

We did something similar with astropy, where we have a function to parse almost any ASCII table. That sounds great until you run into edge cases like "is `observation start`" one column or two, separated by a space? What if I want it to be two columns? What about comment characters? What is every line is commented? Try a different format, or assume that user wants to read a table with no data lines? Now, we try not to touch that code any more and if we add new formats, users have to pass them in explicitly. 
It's kind of the same here: We also read ASCII tables (and fits), and the format (in particular in the ASCII case) is simply non-unique, so, mathematically speaking, there can be no way to get all reads right. But the vast majority of use cases (and all tests) are for 1D reads, so we can reduce a lot of that ambiguity by assuming it's always 1D and for the 2D cases ask users to write like 2 lines of python themselves because they know the formats of their input data and don't have to guess:
```
intab = np.loadtxt(filename, parameters)
mydata =Data2DInt(intab[0,:], intab[1, :], .... whatever ...)
--- or ---
ui.load_arrays('mydataset', intab[0,:], intab[1, :], .... whatever ..., Data2DInt)
```

Trying to be 100% backwards compatible here would get us another rat's nest of ill defined behavior, that's hard to test, creates new edge cases, and will probably never be used in the wild. (Because people who built their own Data2DInt models are likely to be users familiar with Python who would rather load the data themselves than spending time to discover exactly how the API works here. For non-specialist cases, we have, e.g. "load_psf"; those people wouldn't encounter this function I changed here.)

\<end rant\>

hamogu:
I agree that some parts of the C API could be redone or replaced by e.g. scipy (which is already on conda-forge so we don't have to built it), but that's definitely out-of-scope here.

hamogu:
And by "redone" I don't mean "rewriting the C-code", but "expose it at the Python level and/or improve the way it's visible to users".
However, unless we have a specific use case, that might not be the bast use of our time. It works now where we need it, so at least I won't touch it unless I have to.

sherpa/models/basic.py
hamogu: I use it in line 2619 in the same file in `IntegratedTableModel1D` like this:
```Python
    @modelCacher1d
    def _calc_integrate(self, x0, x1):
        return integrate(x0, x1, self._x, self._y,
                         interpolate_func=self.method,
                         integrate_func=self.integrate_method,
                         **self.integrate_kwargs)
```

hamogu:
Rebased and pushed to pick up the changes on main to make CI pass. Also including the small fixes suggested by @DougBurke

sherpa/utils/__init__.py
```diff
@@ -2344,6 +2345,121 @@ def interpolate(xout, xin, yin, function=linear_interp):
...
+        ind0 = np.nonzero(x == x0[i])[0][0]
+        ind1 = np.nonzero(x == x1[i])[0][0]
+        result[i] = area[ind0:ind1].sum()
+
+    return result
+
+
+def integrate(x0: np.ndarray, x1: np.ndarray, xin: np.ndarray, yin: np.ndarray,
+              interpolate_func: Callable[[np.ndarray, np.ndarray, np.ndarray], np.ndarray] = linear_interp,
+              integrate_func: Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray, Callable, dict | None], np.ndarray] = integrate_tabulated_function,
```
DougBurke, line 2405: Technically the signature for `integrate_func` is incorrect since `Callable` can't handle keyword arguments and in line 2458 we call it with an explicit keyword argument and then pass it `**kwargs`. However, it expresses the intent and we can look at improving it at some point in the future when  the `Callable` / `Protocol` support has stabilized for the python versions we support.

For anyone following along at home, I noticed this from:

```
% pyright sherpa/utils/__int__.py
...
  /lagado.real/sherpa/sherpa-main/sherpa/utils/__init__.py:2459:44 - error: Expected 2 more positional arguments (reportCallIssue)
...
```

DougBurke:
This commit will require some changes to https://github.com/cxcsds/docstring_to_ahelp but that sort of thing is expected and not a blocker.

DougBurke (approved):
Sorry for the delay in reviewing.

sherpa/utils/__init__.py
DougBurke: So this is not a comment on this PR per se - just something that we can improve down the line.

sherpa/utils/__init__.py
```diff
@@ -2344,6 +2345,121 @@ def interpolate(xout, xin, yin, function=linear_interp):
     return function(xout, xin, yin)
 
 
+def integrate_tabulated_function(x0: np.ndarray, x1: np.ndarray,
+                                 x_points: np.ndarray, y_points: np.ndarray,
+                                 *args, **kwargs):
+    '''Integrate y_points between x0 and x1
+
+    This function is meant to be used for models that linarly interpolate
```
wmclaugh:
```suggestion
    This function is meant to be used for models that linearly interpolate
```

DougBurke: @hamogu is travelling so I'll just accept this change for him.

PR #2405, opened 2025-10-23
Increase tolerance for a test that occasionally fails in CI

hamogu:
## Details
One o the scipy tests occasionally fails in CI, which is obviously due to numerics, so this PR simply increasese the numerical tolerance by a bit.
For an example, see
https://github.com/sherpa/sherpa/actions/runs/18729863538/job/53423975996?pr=2404
(though logs are not retained forever, so this link will become unavailable at some point).


anetasie (commented):
I assume the rtol is only set to check our test and it is not the tolerance in the optimization algorithm itself. If that the case I would approve this PR

DougBurke (approved):
Apologies for the late review.

anetasie (approved):

PR #2407, opened 2025-11-05
sim: add type statements

DougBurke:
TODO

Check how much this is covered by

- #2507

# Summary

Add typing statements to the sherpa.sim code and improve an error check in the set_sampler routine.

# Details

Reviewing the sherpa.sim code it was hard to identify when an object or a class was meant to be sent around, so add type statements to address. There are also some code improvements - e.g. remove reference to a routine from `inspect` that was removed in Python 3.11 and take advantage of the ABC module - as well as a fix to the set_sampler routine when sent an invalid argument (so that you get the expected error message rather than because `issubclass` threw its own error).

There is likely more that could be done, such as avoiding quite so many `for` loops, but that is low priority work for later.


DougBurke:
Rebased to pick up the new build system from

- #2484

PR #2414, opened 2025-11-17
Remove `get_axes`

### LINKED ISSUE #2410: Inconsistencies and errors in `get_axes`
Currently, there is an UI-level `get_axes` function in /sherpa/astro/ui/utils.py (copied here to show in the this issue without the long doctoring):
```Python
    def get_axes(self,
                 id: IdType | None = None,
                 bkg_id: IdType | None = None):
        d = self._get_data_or_bkg(id, bkg_id)
        if isinstance(d, DataPHA):
            return d._get_ebins(group=False)

        if isinstance(d, (Data2D, DataIMG)):
            return d.get_axes()

        return d.get_indep()
```
 and several classes that implement a `get_axes` method. In the examples, I'll use the OO methods, but since the UI function just carries through, that will give the same answers. So:
```Python
import numpy as np
from sherpa.data import Data2D, Data2DInt

import numpy as np
x1, x0 = np.mgrid[20:30:2, 5:20:2]
datashape = x0.shape
y = np.sqrt((x0 - 10)**2 + (x1 - 31)**2)
x0 = x0.flatten()
x1 = x1.flatten()
y = y.flatten()

reg2d = Data2D("regular2d", x0, x1, y, shape=datashape)
reg2dint = Data2DInt("regular2dint", x0 - 0.5, x1 - 0.5, x0 + 0.5, x1 + 0.5, y, shape=datashape)

reg2d.get_axes()
[OUT] (array([1, 2, 3, 4, 5, 6, 7, 8]), array([1, 2, 3, 4, 5]))
reg2dint.get_axes()
[OUT] (array([1, 2, 3, 4, 5, 6, 7, 8]), array([1, 2, 3, 4, 5]))
```

So, for the integrated data, we get the bin centers.
Now, let's try the same thing for an image:
```Python
img2d = DataIMG("img2d", x0, x1, y, shape=datashape)
img2dint = DataIMGInt("img2dint", x0 - 0.5, x1 - 0.5, x0 + 0.5, x1 + 0.5, y, shape=datashape)

img2d.get_axes()
[OUT] (array([1., 2., 3., 4., 5., 6., 7., 8.]), array([1., 2., 3., 4., 5.]))
img2dint.get_axes()
[OUT] (array([-0.5,  0.5,  1.5,  2.5,  3.5,  4.5,  5.5,  6.5]),
     array([-0.5,  0.5,  1.5,  2.5,  3.5]),
     array([0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5, 7.5]),
     array([0.5, 1.5, 2.5, 3.5, 4.5]))
```
Whoa! Now we get 4 axes instead of 2 for the integrated case.

Why is this different? I consider this a bug in `Data2DInt`.

I also note, that `get_axes` only gets called from examples that use `get_axes` to show how it works or from test that test for it; not from any other code, e.g. I would have expected the plotting code to call this, but that's not the case.
Of course, it's a question if we can just get rid of `get_axes` because it's been around for so long, but I suggest to fix the case for `Data2DInt` to also return the bin boundaries (i.e. 4 axes).

*edited to remove some minor typos*

hamogu:
## Summary
Some data objects (and by extension the UI) have a `get_axis` method, but it's inconsistent, insufficiently documented and superfluous. The method has been removed.

## Details
As discussed in #2410, `get_axes` is superfluous and inconsistent. This PR removes it.

## Questions
We can discuss having a period (one full version?) of deprecation first. This could be added to the code (adding back in all the removed tests). I understand the importance of API stability, but in situations with limited developer support that has to be traded off against the available developer resources to do this in a drawn-out deprecations-then-remove process vs. a simple single PR.
This method wasn't heavily advertised, it's inconsistent (so anything build on top of it should probably have chosen other ways to get the axis data), and can easily be replaced with other ways to access the underlying axis data.

fixes #2410


hamogu:
@DougBurke If you could do a review of this, that would be great. It's been a few months old, but I just rebased on pushed again.

sherpa/astro/ui/utils.py
```diff
@@ -3933,103 +3929,6 @@ def get_indep(self,
         d = self._get_data_or_bkg(id, bkg_id)
         return d.get_indep(filter=filter)
 
-    def get_axes(self,
```
DougBurke, line 3936: Perhaps we should keep this as it is one of the only "documented" ways of getting the energy axis information for PHA data. We can ask @nplee if any of the CIAO/Sherpa documentation refers to `get_axes`.

If  kept it would require some work for the Data2D/IMG case.

sherpa/data.py
```diff
@@ -2521,12 +2521,6 @@ def set_x1label(self, label: str) -> None:
         """
         self._x1label = label
 
-    def get_axes(self) -> tuple[np.ndarray, np.ndarray]:
```
DougBurke, line 2560: We should add a ``.. versionchanged: 4.19.0`` [*] note to the class docstring to say that this method has been removed. I don't think it's worth adding it to the subclasses too.

[*] or equivalent if there's something specific for removed elements, but not sure it's worth it here.

sherpa/astro/ui/utils.py
hamogu: There is https://cxc.cfa.harvard.edu/sherpa/ahelp/get_axes.html (which is autogenerated)

All other pages that google finds either have it in the "See also" section or simply in the help sidebar. So it seems to be "documented" as much as any API is, but not advertised in a thread or something like that.

As I said above, it's an API breaking change and the question is if we are comfortable doing that. It seems that @DougBurke answer is "for the Python layer, yes". 

Given that I can't find any reference (outside autogenerated API docs) and the RTD docs that are removed as part of this PR, are we OK with removing it?

I also did a [complete search for "get_axes sherpa" on GitHub](https://github.com/search?q=get_axes+sherpa&type=code&p=1). This should find files where both words appear in the same file; it is possible, but not likely to fool that (e.g. you could have a file in your project "from sherpa.astro import ui" and then a second one that uses "from otherfile.ui import get_axes" but never mentions the word "sherpa"). Github does find 137 code files:
- Most of those are forked repositories of Sherpa (e.g. mine, Marie's etc.) 
- Gammapy defines a "get_axis" method for their own objects and sometimes that appears in the same file as the word "sherpa", but since they don't seem to use sherpa's "get_axis" this is only accidental, and us removing a method they don't use won't hurt them.
- Sometimes sharpa is used and matplotlib's "fig.get_axes()" is the used in the same notebook.

While use can of course have private code that depends on this, it's certainly not in wide use. So OK to remove?

hamogu:
Added "version changed" and rebased.

DougBurke (approved):

PR #2415, opened 2025-11-18
Explicit types for boolean parameters

DougBurke:
# Summary

Add typing statements to the boolean arguments for much of the code.

# Details

This was originally developed as part of https://github.com/sherpa/sherpa/pull/2411 but it makes sense to pull it out into a separate PR. If there are significant changes to be made I'd like to make them in a follow-up PR to simplify the existing PRs that depend on this.

This is primarily to change `foo=False` to `foo: bool = False` (ditto for `True`) in signatures. This is not particularly informative or illuminating, but it's an easy change to make and hopefully easy to review (if not particularly informative). In itself it does not really improve the typing code (there's no reduction in the number of errors from `pyright`, for instance). This started out as being restricted to the UI code, but it has since expanded to include most of the code. Some areas not changed

- the image code, as that is done in #2413 
- the actual backend code (pylab/bokeh) although the superclass (effectively the "dummy" backend) has the changes 
- by not marking a parameter - such as `verbose=False` in `sherpa.utils.NoRichardsonExtrapolation.__init__` I am indicating that "I have not looked at this code, probably because it needs a lot of thought and it is not worth it right now" ....

However, we do get a number of reductions in type errors because of the changes to `sherpa.utils.bool_cast`. This included support for converting an array/iterable of values as well as a scalar one, but this meant it was typed as returning `bool | np.ndarray` and, given the input, it was hard for the typing code to identify when it returned an array. This lead to a lot of errors when `bool_cast` was used as the type checker thought it could return an array when sent to a "scalar boolean value". It turns out the only use of this functionality was in a test case I'd added to check it reported an error when sent an invalid value. So, by removing this support we make `bool_cast` only support the scalar case, and thus reducing the number of errors. 

Using `pyright` version 1.1.407, the number of errors reported from several files were reduced:

|  | `ui/utils` | `astro/ui/utils` | `data` | `astro/data` |
| --- | --- | --- | --- | --- |
| before | 116 | 160 | 88 | 156 |
| after | 105 | 131 | 87 | 148 |

(some of these may be due to the other related typing changes in `sherpa.utils`, but the majority are due to `bool_cast`). Here is an example of the error report from `sherpa.data` that was removed thanks to the `bool_cast` change

```
/home/dburke/sherpa/sherpa-3/sherpa/data.py:782:18 - error: Type "NDArray[Any] | bool" is not assignable to declared type "bool"
  Type "NDArray[Any] | bool" is not assignable to type "bool"
    "ndarray[_AnyShape, dtype[Any]]" is not assignable to "bool" (reportAssignmentType)
```

The first commit is actually an unrelated change:

- I had missed importing `Any` in `sherpa.astro.ui.utils` in #2377, so add that in (fortunately only an issue when running a type checker on the code)

Now, `bool_cast` **could** 

- be removed, as it isn't really useful any more
- or be made to raise a deprecation warning if sent a string argument

I would much rather discuss this at a later time. If pressedm I would have a preference for the deprecation approach, since some users may have code that has `clobber="yes"` (or equivalent) thanks to historical documentation/suggestions and we don't want to make this code fail.

### DIFF
--- a/sherpa/astro/sim/fullbayes.py
+++ b/sherpa/astro/sim/fullbayes.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2016, 2017, 2020, 2021, 2023
+#  Copyright (C) 2011, 2016-2017, 2020-2021, 2023, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -37,9 +37,20 @@ class FullBayes(PragBayes):
                            'old_rr': None}
                           for arf in self.arfs]
 
-    def init(self, log=False, inv=False, defaultprior=True, priorshape=False,
-             priors=(), originalscale=True, scale=1, sigma_m=False, p_M=.5,
-             simarf=None, p_M_arf=.5, sigma_arf=0.1):
+    def init(self,
+             log: bool = False,
+             inv: bool = False,
+             defaultprior: bool = True,
+             priorshape: bool = False,
+             priors=(),
+             originalscale: bool = True,
+             scale=1,
+             sigma_m: bool = False,
+             p_M=.5,
+             simarf=None,
+             p_M_arf=.5,
+             sigma_arf=0.1
+             ):
         # nsubiters is missing from init() to indicate that nubiters=1 for
         # full bayes
 
--- a/sherpa/astro/sim/pragbayes.py
+++ b/sherpa/astro/sim/pragbayes.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2016, 2017, 2019, 2020, 2021, 2023
+#  Copyright (C) 2011, 2016-2017, 2019-2021, 2023, 2025
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -315,9 +315,19 @@ class PragBayes(MetropolisMH):
 
         self.simarf = None
 
-    def init(self, log=False, inv=False, defaultprior=True, priorshape=False,
-             priors=(), originalscale=True, scale=1, sigma_m=False, p_M=.5,
-             simarf=None, nsubiters=10):
+    def init(self,
+             log: bool = False,
+             inv: bool = False,
+             defaultprior: bool = True,
+             priorshape: bool = False,
+             priors=(),
+             originalscale: bool = True,
+             scale=1,
+             sigma_m: bool = False,
+             p_M=.5,
+             simarf=None,
+             nsubiters=10
+             ):
 
         # Note that nsubiters is used as a dummy parameter to indicate the
         # default value.  See the function WalkWithSubIters.__call__()
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -923,7 +923,12 @@ def smoke(verbosity: int = 0,
     smoke.run(verbosity=verbosity, require_failure=require_failure, fits=fits, xspec=xspec, ds9=ds9)
 
 
-def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=False):
+def _smoke_cli(verbosity=0,
+               require_failure: bool = False,
+               fits=None,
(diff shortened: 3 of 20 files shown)


sherpa/astro/sim/fullbayes.py
```diff
@@ -37,9 +37,20 @@ def __init__(self, fcn, sigma, mu, dof, fit, *args, rng=None):
...
-             simarf=None, p_M_arf=.5, sigma_arf=0.1):
+    def init(self,
+             log: bool = False,
+             inv: bool = False,
+             defaultprior: bool = True,
+             priorshape: bool = False,
+             priors=(),
+             originalscale: bool = True,
+             scale=1,
+             sigma_m: bool = False,
```
hamogu, line 48: I'm not sure the type is correct here. I can't claim to understand the inner workings of the MH Sampler, but looking at https://github.com/sherpa/sherpa/blob/main/sherpa/sim/mh.py#L461-L463:
```Python
        self.sigma_m = sigma_m
        if not np.mean(sigma_m):
            self.sigma_m = self._sigma.copy()
```
it seems that sigma_m could be an array or some sort, too.

sherpa/astro/sim/pragbayes.py
```diff
@@ -315,9 +315,19 @@ def __init__(self, fcn, sigma, mu, dof, fit, *args, rng=None):
...
-             simarf=None, nsubiters=10):
+    def init(self,
+             log: bool = False,
+             inv: bool = False,
+             defaultprior: bool = True,
+             priorshape: bool = False,
+             priors=(),
+             originalscale: bool = True,
+             scale=1,
+             sigma_m: bool = False,
```
hamogu, line 326: Same here. Is this really a bool, or is it "bool or array or something else"?

sherpa/utils/__init__.py
```diff
@@ -2642,14 +2650,21 @@ def is_in(arg, seq):
...
 
 
-def is_iterable(arg) -> bool:
-    return isinstance(arg, (list, tuple, np.ndarray)) or np.iterable(arg)
+# The TypeGuard signals intent, even if it is not quite what the code
+# does.
+#
+def is_iterable(arg) -> TypeGuard[Sequence]:
+    # This check used to be more-complex, but it has been simplified to
+    return isinstance(arg, Iterable)
```
hamogu: `is_iterable` is only used 5 times in the code (in two separate files). Since this is a one-liner with a fairly complex (TypeGuard!) markup, I suggest to simply inline it in the five cases it's used. That will make the intent even more clear without checking a separate file.

sherpa/utils/__init__.py
```diff
@@ -2642,14 +2650,21 @@ def is_in(arg, seq):
...
 
 
-def is_iterable_not_str(arg: Any) -> bool:
+# The TypeGuard signals intent, even if it is not quite what the code
+# does.
+#
+def is_iterable_not_str(arg: Any) -> TypeGuard[Sequence]:
     """It is iterable but not a string."""
```
hamogu: This is used only three times in Sherpa - even more reason to simply inline it where it's used.

hamogu (changes requested):
I'm so not looking forward to rebase #2359 on this - I foresee a lot of manual editing. For that reasons, I'm really not a friend of PRs that just change the formatting or add typing all over sherpa, touching essentially every module just to make the upright output look a few percent better.

Working on a module and adding typing while doing useful changes is one thing, but touching everything just to add a few types or selected keywords creates significant busywork for updating other PRs (most of which are written by @DougBurke himself...).
But, alas, it's not for me to decide if we do want such type of work or not. From a technical point, I only have four minor comments.

DougBurke:
I've moved the initial changes to #2415 to try and make things a bit easier to review.

DougBurke:
Rebased now that #2415 has been merged. I do not expect to have this PR reviewed anymore. Instead I want to break things up a bit to make it less boring to review.

PR #2417, opened 2025-11-19
Overload fixes

DougBurke:
Not sure if I shall keep this to merge or for reference in other PRs...

# Summary

Improve and extend the use of overload calls to add more-specific types to routines.  Add types for most of the sherpa.utils module.

# Details

This depends on #2415 just to avoid a minor conflict.

The aims are

- to better use `overload` (see the "Overload changes" section below to explain the reasoning)
- replace `Optional[X]` by `X | None`
- add typing statements to most of `sherpa.utils`
  - this doesn't really fit into the other changes, but I need to add this change somewhere and it's just going to be hard to do without causing merge conflicts
  - there is some numerical code that needs a more-significant review than just "oh, this returns a ndarray", and now is not the time to do it; I did add a "TODO" statement as a marker to spot in later times, but the fact these classes do not have type statements makes them easy to spot
 
## Overload changes

The aim is to try and add useful types. We have a number of routines that behave something like "when sent None, return None, otherwise accept an array-like value and return an ndarray". This can be typed something like

```python
from typing import overload

import numpy as np

@overload
def foo(arg: None) -> None: ...

@overload
def foo(arg: list) -> np.ndarray: ...

def foo(arg: list | None) -> np.ndarray | None:
    if arg is None: return None
    return np.asarray(arg)
```

I originally (i.e. in existing code, not in this PR) started writing these without adding explicit types to the runtime line (the non-`overload`ed version). However, comments in the Python typing ecosphere suggested this was wrong, and are needed. So I have started to do so, such as the version here. It is annoying (particularly for more-complex signatures than this one), but it can be useful when looking at the code. Now, recent experiments with `pyright` version 1.1.407 and `mypy` version 1.18.2 suggest it's no-longer needed, but I plan to keep them.

I have had fun dealing with routines that use a boolean value to trigger different return values, and also when the value is optional. Investigations with `pyright` and `mypy` lead me to code like

```python
@overload
def foo2(val: str, flag: Literal[True]) -> bool: ...

@overload
def foo2(val: str, flag: Literal[False]) -> None: ...

def foo2(val: str, flag: bool) -> bool | None: ...
```

for handling boolean values. For optional values we can just use `None` and apparently do not need to use `Literal[None]`:

```python
@overload
def bar(a: None = None, b: None = None) -> None: ...

@overload
def bar(a: int, b: None = None) -> int: ...

@overload
def bar(a: None, b: int) -> int: ...

@overload
def bar(a: int, b: int) -> int: ...

def bar(a: int | None = None, b: int | None = None) -> int | None: ...
```

Similarly, when an optional value is a boolean we use `Literal[V] = V` for `V` being `True` or `False`:

```python
@overload
def foo4(val: str, flag: Literal[True] = True) -> bool: ...

@overload
def foo4(val: None, flag: Literal[False]) -> None: ...

def foo4(val: str | None, flag: bool = True) -> bool | None: ...
```

There is an interesting case where the `bool` was the last argument and there was an optional argument before that. So I needed both `Literal[False] = False` and `Literal[True] = True` for a value that defaulted to `False`. Fortunately the `smoke_test` caught this (or the normal tests would have also found it).

### DIFF
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -909,7 +909,12 @@ def smoke(verbosity: int = 0,
     smoke.run(verbosity=verbosity, require_failure=require_failure, fits=fits, xspec=xspec, ds9=ds9)
 
 
-def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=False):
+def _smoke_cli(verbosity=0,
+               require_failure: bool = False,
+               fits=None,
+               xspec: bool = False,
+               ds9: bool = False
+               ) -> None:
     from optparse import OptionParser
     parser = OptionParser()
     parser.add_option("-v", "--verbosity", dest="verbosity",
--- a/sherpa/__init__.py
+++ b/sherpa/__init__.py
@@ -923,7 +923,12 @@ def smoke(verbosity: int = 0,
     smoke.run(verbosity=verbosity, require_failure=require_failure, fits=fits, xspec=xspec, ds9=ds9)
 
 
-def _smoke_cli(verbosity=0, require_failure=False, fits=None, xspec=False, ds9=False):
+def _smoke_cli(verbosity=0,
+               require_failure: bool = False,
+               fits=None,
+               xspec: bool = False,
+               ds9: bool = False
+               ) -> None:
     from optparse import OptionParser
     parser = OptionParser()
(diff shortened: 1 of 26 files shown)


DougBurke:
Rebased as a number of changes from this PR have either been copied into merged PRs or similar changes were made elsewhere.

Issue #2421, opened 2026-01-14
pytest 9.1 changes


DougBurke:
Our "basic" pip test has just started to throw warning/errors like

```
>       assert 0 == nwarnings
E       assert 0 == 1

work/sherpa/sherpa/sherpa/conftest.py:305: AssertionError
--------------------------- Captured stdout teardown ---------------------------
*** Warnings created: 1
1/1 {message : PytestDeprecationWarning("\nModule 'sherpa.astro.xspec._xspec' was found, but when imported by pytest it raised:\n    ImportError('XSPEC support is not enabled')\nIn pytest 9.1 this warning will become an error by default.\nYou can fix the underlying problem, or alternatively overwrite this behavior and silence this warning by passing exc_type=ImportError explicitly.\nSee https://docs.pytest.org/en/stable/deprecations.html#pytest-importorskip-default-behavior-regarding-importerror"), category : 'PytestDeprecationWarning', filename : '<doctest sherpa.astro.io.xstable[0]>', lineno : 1, line : None}
_______ ERROR at teardown of [doctest] sherpa.astro.io.xstable.BaseParam _______
```

Need to sit down with a good book and stare at https://docs.pytest.org/en/stable/deprecations.html#pytest-importorskip-default-behavior-regarding-importerror until it gives up it's secrets, since the fact this comes from a `doctest` means t's not at all obvious where to stick the `exc_type=ImportError` argument...

DougBurke:
I'm going to assume this has been closed by #2423

PR #2422, opened 2026-01-24
Change channel location where XSPEC if found

hamogu:
## Summary
Change channel location where XSPEC if found

## Details
From an email on sherpa-dev:
>>> I had consolidated the channels recently, we had a lot of dup packages and needed to save space. Using conda/test would be suggested for in CIAOX/T/sherpa xspec packages, and conda/ciao for released ones.

I'm picking the XSPEC from the last released CIAO version here, since that that we usually want to test compatibility with. PRs that explicitly change the XSPEC version of interface will have to be looked at separately.

## Note
Note that, at a time I'm opening the PR, CI will fail for #2421.
And any PR working on #2421 will fail because of the problem addressed here.  Thus, we will have to merge one of them even with CI failing or make a combined PR for both. I prefer the first approach since there are really separate issues and should be dealt with independently.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  xspec_channel: "https://cxc.cfa.harvard.edu/conda/xspec"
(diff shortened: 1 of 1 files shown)


harlantc (approved):
lgtm

DougBurke (approved):

PR #2429, opened 2026-03-06
Document the `to_plot` method

hamogu:
## Summary
This adds an extensive docstring and also a section to the narrative documentation.

## Rationale
I'm often asked, formally or informally (i.e. not through the issue tracker), how to "just get the values for a plot". 

While the Sherpa plotting system is extensive it can never cover all use cases and it's a bit cumbersome to use for plots that don't fit neatly into the pre-defined plot classes.
And while it offers powerful customization, there is also a steep learning curve if one wants to unleash the full power, e.g. use Sherpa to plot data and a model into an inset on top of some other data (requires `plt.sca()` and the `clearwindow=False` options, none of which are obvious). Sometimes it's just easier to do the whole plot yourself.

In particular, for PHA data (probably the largest user base for Sherpa), however, one cannot just pull the independent and depedent axes form the Data object, because those don't have filtering and grouping applied.
Thus, I'm adding docs for the `to_plot` method that fills exactly this niche.

## Question
I chose to put the PHA example into the `DataPHA.to_plot` docstring and point to it from the narrative docs. It's not obvious it's there, but at least poepe who use iPython to identify the methods of the data objects will see th emethod and many editors will display the docstring on mouse over. 
I'm not sure that's the best location though. It could also be moved into the narrative docs.


DougBurke:
I put some documentation in https://sherpa.readthedocs.io/en/4.18.0/evaluation/examples.html#x-ray-data-datapha (for the PHA case) but this information is a bit spread out (e.g. it took me a little while to find). Also, the `to_plot` method hasn't been a great match to the PHA case (thanks to the whole "is it a `Data1D` or `Data1DInt`" question, but I forget if this was improved when I reworked the plotting code a while ago.

I'll have a look next week.

DougBurke:
I assume this test failure is an example that #2405 fixes.

hamogu:
I think so. I'll let you review first and then worry about how to deal with an unrelated test failure.

sherpa/astro/data.py
```diff
@@ -5189,6 +5189,88 @@ def to_fit(self, staterrfunc=None):
...
+            >>> with as_file(files('sherpa.astro.datastack.tests.data') / '3c273.pi') as phafile:
+            ...     pha = read_pha(str(phafile), use_errors=True)
+            >>> pha.set_analysis('energy')
+            >>> pha.notice(0.5, 7)
+            >>> pha.group_counts(25)
+            >>> x, y, xerr, yerr, xlabel, ylabel = pha.to_plot()
+            >>> print(f"{xlabel}, {ylabel}")
+            Energy (keV), Counts/sec/keV
+
+        We can compare the length of the oringial PHA data with `x` and we see that
```
wmclaugh:
```suggestion
        We can compare the length of the original PHA data with `x` and we see that
```

sherpa/astro/data.py
```diff
@@ -5190,6 +5190,88 @@ def to_fit(self, staterrfunc=None):
...
+        Now, we define a model. Could a blackbody describe this spectrum?
+        When we evaluate the model with the ``yfunc`` argument, we need to
+        include the instrument response:
+
+            >>> from sherpa.astro.models import BBody
+            >>> bb = BBody()
+            >>> bb.kT = 0.5
+            >>> bb.ampl = 0.0005
+            >>> resp = pha.get_full_response(pileup_model=None)
+            >>> x, (ydata, ymodel), xerr, yerr, xlabel, ylabel = pha.to_plot(yfunc=resp(bb))
```
```suggestion
            >>> model = resp(bb)
            >>> x, (ydata, ymodel), xerr, yerr, xlabel, ylabel = pha.to_plot(yfunc=model)
```

just to avoid too much going on in this line?

DougBurke:
I'm not sure my comment at https://github.com/sherpa/sherpa/pull/2429#issuecomment-4145284751 is something that needs addressing in this PR but possibly a future one.

sherpa/astro/data.py
DougBurke: (it would be nice for `to_plot` to return some form of structure, whether just a `dict`, a `dataclass`, or something else, but

a) that's not the reason for this PR, so definitely not anything to add to this one
b) I am not sure it's worth doing

) given the existing issues with `to_plot` and how we use it.

DougBurke: @hamogu - do you want to just say "this is not worth it" so we can close it?

hamogu: It's come up repeatedly when I talk to users. I do think it's worth it unless we're going to replace this entire "to_plot" mechanism before the next release.

DougBurke: I just meant whether you wanted to close this particular suggestion, which was to reword the example, or to apply it. 

I am happy to take the PR but this particular comment is open

hamogu: Yes I agree it would be better to be more structured. It's something I've long wanted to do (and I have an pretty extensive but not quite ready rework of all the inheritance of the `DataXXX` classes), but "perfect is the enemy of good". Let's get this in and I'll add a remark that we're working on reworking that, but you can sue it now.

hamogu:
Added a warning and rebased onto https://github.com/sherpa/sherpa/pull/2453 as it changes the DS9 download code (and updates versions).

sherpa/astro/data.py
```diff
@@ -5201,6 +5201,12 @@ def to_plot(self, yfunc=None, staterrfunc=None, response_id=None):
         The return values are different from the "raw" values of the data in the
         sense that filtering and binning are applied to x, y, and all errors.
 
+        .. warning::
```
DougBurke: I am not convinced a warning here is that useful, other than perhaps for a reader to say "I won't use this call", which isn't that useful.

If we did want to take it then there's some typos to add and we'd also want to add the warning to the other `to_plot` calls.

hamogu: I agree that (a) it needs to be added and typo-corrected if we take it.

This was my attempt to react to your "but it's not a great interface and we should change it" comment. I agree with that, too, but especially given the thread that Sherpa development will be cut or severely reduced, I do want to put in this PR. We never get to make it perfect. 
It's a bit odd to document a "internal-looking" routine and immediately say "don't use it". I'm suggesting this as a compromise between getting something in now (which might be permanent) without preventing us from changing it in the near future if we get to it (because it would then be considered API breaking).

I would think that most uses for this are mostly interactive plotting in notebooks. I don't expect anyone to build an external library on top of this function. So, small changes in API can probably be fixed by users when they encounter it.

@DougBurke What do you think? Take out the warning or fix it up?
I don't have a strong preference, but I do want to get this iPR n sooner rather than later.

DougBurke: I suggest remove the warning.

hamogu: Done. Should be all good then.

DougBurke (approved):
Ta - sorry it took so long for such a small change.

PR #2432, opened 2026-03-12
Allow building with GCC 15 (std=C23)

### LINKED ISSUE #2223: Fails to build with GCC 15 due to redefiniton of C23 keywords

DougBurke:
# Summary

Building Sherpa now requires a C compiler that supports the gnu11 standard and a C++ compiler that supports the C++11 standard. The build will now add the appropriate arguments to the CFLAGS and CXXFLAGS environment variable unless they already contain "-std=" arguments. The c_standard and cxx_standard options in the setup.cfg file can be used to change the standard. This allows Sherpa to be built with GCC 15.

# Details

Originally this PR changed the extern code to allow it to build with GCC 15, but it turns out you can actually do this with just

    CFLAGS="-std=gnu11" pip install .

Rather than make all users have to set this variable, the build now enforces the standard requirements when building by setting (or adding to) the CFLAGS environment variable, apart from two cases

- the environment variable exists and contains the string "-std="
- the sherpa_config.c_standard setting in setup.cfg is set to the empty string

Since we are at it, we explicitly mark the C++ code as requiring C++11 via the cxx_standard setting.

The aim is to provide enough flexibility to let users work with variant compilers, but still allow the default case to build easily with GCC 15.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -22,7 +22,12 @@ if [ "`uname -s`" == "Darwin" ] ; then
     tar -C ${GITHUB_WORKSPACE}/../11.0SDK -xf MacOSX11.0.sdk.tar.xz
 
 else
-    compilers="gcc_linux-64=14.2 gxx_linux-64=14.2 gfortran_linux-64"
+    if [ -n "${COMPILERVER}" ]; then
+        cver="${COMPILERVER}"
+    else
+        cver="14.2"
+    fi
+    compilers="gcc_linux-64=${cver} gxx_linux-64=${cver} gfortran_linux-64"
 
     if [ -n "${MATPLOTLIBVER}" ]; then
         #Installed for qt-main deps which is needed for the QtAgg backend to work for matplotlib
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -65,6 +65,7 @@ jobs:
             numpy-version: "2.3.2"
             install-type: develop
             test-data: none
+            compiler-version: "15.2"
 
           - name: Linux Full Build (Python 3.13)
             os: ubuntu-latest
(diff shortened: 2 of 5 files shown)


DougBurke:
@harlantc - are you able to look at this? I'm hoping this is a temporary change, but I am finding it hard to build Sherpa on one of my main work machines without some sort of fix like this.

DougBurke:
Closed by #2484

PR #2433, opened 2026-03-12
TEST: fix test that requires region support

DougBurke:
# Summary

Mark a test a requiring the region library. There is no functional change.

# Details

This was missed in #2388 (as the test also requires WCS and it's not often we build with one and not the other).


hamogu (approved):
Either way is fine. I would have also been OK with "you can only build with both of region and wcs or none of them" instead of individually switching them on and off, but this is fine, too.

PR #2436, opened 2026-03-12
TEST: update tolerance

DougBurke:
# Summary

Further relax the tolerance for the test of a SciPy optimizer (see #2405). There is no functional change.

# Details

Apparently a tolerance of 1e-4, added in #2405, is still too small, so bump to 1e-3 (the failures seen are close to a tolerance of 1e-4 so hopefully bumping to 1e-3 will fix things).

### DIFF
--- a/sherpa/optmethods/tests/test_scipy_interface.py
+++ b/sherpa/optmethods/tests/test_scipy_interface.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2025
+#  Copyright (C) 2025, 2026
 #  MIT
 #
 #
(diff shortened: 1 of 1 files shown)


hamogu (approved):
In my testing, the number I set was enough to keep it passing (I run it like 20 times or so), but that probably also depends on the OS, the scipy version, the day of the week, and the current temperature in Timbuktu. 
It's scipy's job to test that the optimization works, we only want to check that the interface doesn't crash and the result isn't totally ridiculous. So, we should make it as loose as we need to to not be annoyed by this.

PR #2439, opened 2026-04-02
Correctly document model parameters in some XSPEC models

DougBurke:
# Summary

Correctly label some XSPEC model parameters in the documentation. This changes how the model parameters are displayed in the sphinx/read-the-docs documentation. There is no functional change to the code.

# Details

This was found when updating the code to support XSPEC 12.15.1 in #2393 but it is independent of that work so can be taken now.

The basic change is that some XSPEC models reference model parameter names (which are field names, hence "Attributes" in the terminology of sphinx/docstrings) were being described as "Parameters" (which is meant to be values used when calling the class initiializer, which is definitely not is done here). So

a) change the docstrings where necessary
b) change the code used to autogenerate this (for new models)
c) change the tests for part b

# Examples

Here's the `XSvgabs` read-the-docs page in 4.18.0: https://sherpa.readthedocs.io/en/4.18.0/model_classes/api/sherpa.astro.xspec.XSvgabs.html#xsvgabs

and with this PR: https://sherpa--2439.org.readthedocs.build/en/2439/model_classes/api/sherpa.astro.xspec.XSvgabs.html#xsvgabs

The information is the same but the presentation is slightly different.

### DIFF
--- a/scripts/add_xspec_model.py
+++ b/scripts/add_xspec_model.py
@@ -23,6 +23,7 @@
 """Usage:
 
   ./add_xspec_model.py version infile model
+     --allow_exists
 
 Aim:
 
@@ -34,9 +35,10 @@ any possible version updates.
 
 """
 
-import sys
+import argparse
 
 from sherpa.astro.utils import xspec
+import sherpa.astro.xspec
 
 
 Version = tuple[int, int, int]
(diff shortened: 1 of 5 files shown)


scripts/add_xspec_model.py
```diff
@@ -93,7 +96,9 @@ def create_xspec_model(version: Version,
        The XSPEC version
     models
     modelname
-        XSPEC model name (as given in the model.dat file)
+       XSPEC model name (as given in the model.dat file)
+    allow_exists
+       If False then the model can not already exists.
```
hamogu:
```suggestion
       If False then the model can not already exist.
```

hamogu (approved):
one minor typo... but I don't have to review this again when that's fixed, so I'm clicking "approve"

PR #2398, opened 2025-10-08
model failures from AtomDB file changes in XSpec 12.15.1

nplee:
XSpec 12.15.1 drops the APEC/NEI/eigenvector files associated with AtomDB 3.0.9 which were included in XSpec 12.15.0 and later micro versions of XSpec 12.14.  For the pytests, the AtomDB file versions are explicitly set in the `sherpa/conftest.py` file for consistency sake across different prior versions of XSpec that Sherpa is buildable against; but the lack of the associated files will cause many models to either segmentation fault or core dump while others will return zero counts since AtomDB-dependent CIE model components failed to evaluate.

This update changes the version of AtomDB files used for the tests with XSpec 12.15.1 going alongside PRs #2393 and #2397.  This change will result in some test failures in `sherpa/astro/ui/tests/test_astro_ui_unit.py` with approximation failures and `sherpa/astro/ui/tests/test_serialize.py` with AtomDB version checking.

PR #2252, opened 2025-03-26
Allow Sherpa to build against XSPEC 12.15.0

DougBurke:
# Summary

Allow building against XSPEC 12.15.0. Adds the following 17 models:

- additive: XSfeklor, XSvlorentz, XSvvoigt, XSzfeklor, XSzlorentz, XSzvlorentz, XSzvoigt, XSzvvoigt
- convolution: XSrgsext
- multiplicative: XSlorabs, XSvlorabs, XSvoigtabs, XSvvoigtabs, XSzlorabs, XSzvorlabs, XSzvoigtabs, XSzvvoigtabs

The XSPEC initialization file (~/.xspec/Xspec.init) may need to be removed or the AtomDB version in it updated before using XSPEC models.

# Details

The first commit updates the scripts `add_xspec_model.py` and `update_xspec_functions.py` to reduce friction. The `sherpa.astro.utils.xspec` code used by `add_xspec_model.py` is primarily intended to support XSPEC user models, and so the output required manipulation when used for XSPEC, so the code has been updated to support this "internal" use to better match the Python code used in `sherpa.astro.xspec` (adding a version decorator, a "versionchanged" note, and a reference link with a best-guess for the XSPEC HTML documentation location). The `update_xspec_functions.py` code preciously did not recognize version constraints - the `#ifdef XSPEC_x_y_z` lines used to indicate function changes over time for a model - because there's no encoding of this information in XSPEC itself. The code now takes the existing `sherpa/astro/xspec/src/_xspec.cc` file and uses that to define the version constraints for each model, and uses that to create the new output. It does not create the smallest possible set of `#ifdef` lines, but it is a huge improvement over the previous approach. 

The second commit adds a new helper script - `scripts/update_xspec_docs.py`` which automates the update of the ``docs/model_classes/astro_xspec.rst`` file. It turns out I missed some changes in the XSPEC 12.14.x series so this also updates the ``astro_xspec.rst`` file to reflect these new models.

The last commit adds support for XSPEC 12.15.0. Unfortunately you can not build against 12.15.0 without these changes. There are new models, but no parameter changes.

When testing

- you may need to update or remove `~/.xspec/Xspec.init` to addres AtomDb version updates
- the tests force a particular AtomDB version (3.0.9) that was used in recent XSPEC releases (the default for 12.15.0 is 3.1.2 or so); this is not ideal and we need something like #2224 to handle this more gracefully

There's no conda build of 12.15.0 so we can not test this with the CI runs yet. 

# Notes

[...]

PR #2444, opened 2026-04-15
Simplify the XSPEC tests to not require a particular ATOMDB version

DougBurke:
# Summary

Avoid requiring certain ATOMDB/NEI versions for the XSPEC tests. There is no functional change.

# Details

This is essentially a rework of #2398 by @nplee 

XSPEC 12.15.0 support - added in #2252 - needed to handle changes in ATOMDB version. However, with XSPEC 12.15.1 it has become obvious that the technique used (which was to force a particular set of versions) isn't going to work when we support multiple XSPEC versions (as the set of available ATOMDB models is currently updating rapidly, and so the set of data files included with XSPEC keeps on changing [*]).

So, remove this feature after changing tests that appear to be sensitive to the setting (these are tests which I chose to use an apec model but we can use  a simpler model such as powerlaw which doesn't depend on these data files).

The serialize tests  need updating (to remove the changes made in #2398). There are changes needed to support XSPEC 12.15.1, but they can be done in #2393

---

[*] not to mention the difference between the data provided in a source distribution to that in the XSPEC conda channel

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_unit.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_unit.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2017, 2018, 2020 - 2025
+#  Copyright (C) 2017-2018, 2020-2026
 #  Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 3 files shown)


DougBurke:
@nplee - when you have time can you give this a look (I tried making you a reviewer but the system didn't seem to like that ;-)

nplee:
This works for the purpose of the tests!  Avoidant, but certainly circumvents the issue of dealing with the moving target that's AtomDB.

hamogu (approved):

PR #2446, opened 2026-04-20
BUILD: use the conda fftw package

DougBurke:
# Summary

Use the conda fftw package rather than build it from extern/ for the conda CI runs. There is no functional change.

# Details

Does this make things faster?


DougBurke:
I have merged these changes into #2242 so this can be closed

PR #2382, opened 2025-09-16
Improve save_all for certain PHA situations

DougBurke:
# Summary

The save_all output for PHA data now avoids loading in files it does not need to do (e.g. a background file set in the BACKFILE keyword), which fixes handling of grating data with BACKGROUND_UP and BACKGROUND_DOWN columns. The use_errors flag is now correctly recorded for load_pha calls. The auto_load flag has been added to save_all to allow the output to include these files, to match the output from previous versions. Fix #320. Fix #2383.

# Details

Extend the work added in #2377 to fix #320: this is where a PHA file contains multiple backgrounds (e.g. Chandra HETG data with BACKGROUND_UP/DOWN columns). We take advantage of the "data store" added in #2377 and extend it to all PHA files (not just PHA2 files), and use it to determine when we can avoid loading in ancillary data (by tracking what ancillary data is automatically loaded for us). As part of this (tracking load_bkg calls) we can also fix #2383.

There are at least two use cases for the old behaviour - that is, explicitly including load_arf/rmf/bkg calls even when not needed:

1. to more-closely match previous output (e.g. if the save file is passed through some post-processing or you just want to check what has changed)
2. there is a use for being explicit about all the files that are needed to re-create the state

So, the `save_all` command has gained a new optional parameter, `auto_load` which, when set to `False` (it's default is `True`), will include these `load_arf/rmf/bkg` calls in the output that are technically un-needed.

The resulting serialization code is a bit messy, but it is hopefully not too complex, and covered by both existing and new tests.

PR #2447, opened 2026-04-21
Track filenames on load  WIP

DougBurke:
# Summary

Improve the save_all serialization code so that it better handles datasets and ancillary information - such as ARF and RMF files - that were created rather than read in from a file.

# Details

[this has now been rebased onto #2476 to separate some code clean up and other bug fixes with these changes]

In #2377 the serialization code (that is used to create the `save_all` code) was updated to better handle PHA2 files. This was extended in the #2382 PR to better track the "ancillary" files that can be automatically loaded with a PHA file (ARF, RMF, and background). As noted at the time, the techniques could have been applied to track all loaded files, but it was too large a change at that time. Well, now is the time. These changes should not affect most users of `save_all`: it mainly improves the situation for users who may have manually created a dataset or response rather than read it in from a file.

We add a specialized type (`FileStore`) to track the load call (the function name and the arguments) which can then be used to re-create the call when creating the `save_all` output file. The `Session` object then contains a number of dictionaries that store this data: the key is the dataset identifier and the value is the `FileStore` object. In fact, the PHA support complicates this as without it we would only need to store information for "data" calls but now we have to also store background, ARF, RMF, background ARF, background RMF, and information used to track PHA2 files (which have multiple datasets created with a single call). So we introduce a `sherpa.astro.ui.serialize.Storage` class that contains dictionaries to store `data`, `arf`, `rmf`, `bkg`, `bkg_arf`, and `bkg_rmf`. This class has methods to store information in these structures (but not really for reading out the data, as that is assumed to just use the dictionary access directly). This is then stored as the `_storage` field of the `Session` class. There is a separate `_multi_data_store` structure which could perhaps be merged into `Storage`. TBD.
    
The idea is that `set_data` clears these values and they are filled by the `load_xxx` call (internally these all call `set_data` so we need to fill the store field(s) **after** the `set_data` call).
    
We also need to handle `delete_data` and `copy_data` (an alternative would be to track the actual `copy_data` call rather than converting it to a `load_xxx` call).
    
When tracking the file state we have a "trinary" situation (again because of PHA files):
    
- no direct load_xxx call

  This can be a used manually creating the response object but it can also be via an explicit call to `unpack_xxx` followed by `set_xxx`. This is indicated by there not being a value in the file store (but it's awkward because you either have `idval`, `resp_id` or `idval`, `bkg_id`, `resp_id` so there are a lot of nested dictionaries, which means following the logic is messy as all these have the same type of `IdType`).
    
  Previously the code did not recognize this for ARF/RMF files and backgrounds. The code now attempts to manually re-create the responses, but it is not guaranteed to be correct (e.g. if a `DataRMF` subclass is used or metadata is added and needed).
    
- a `load_xxx` call has been made, so there is a `FileStore` item to store
    
- the response (background, ARF, or RMF) was automatically loaded via the BACKFILE, ANCRFILE, or RESPFILE keywords
    
  In this case a `FileStore` object is synthesized to represent the actual `load_xxx` call, but marked with the `autoloaded` flag set to True. This can then be checked when serializing the code to determine whether to write out the call or not (since we added, in one of the recent changes, the option to always explicitly load the responses (as it can be better to understand what is actually needed).

There is also the fact that `load_xxx` can be called with a "I/O object" (`HDUList` or `Crate`). In this case we just treat this as "manually created" as it's too hard and unreliable to track the provenance of these objects.

 # Note on `FileStore`
    
I had originally set up the FileStore to have a number of subclasses, one for each load_xxx call (and actually, multiple for PHA files to handle different cases), but I decided it actually made the code more opaque.
    
# Four issues
    
a) PHA2 files can load multiple datasets with one call, which complicates the tracking. The `_multi_data_store` dictionary is used to find those ids which were created by a single call. This is partly needed so that a sequence like
    
        load_pha("pha2")  # this creates 1, 2, ..., 11, 12
        delete_data(1)
        delete_data(2)
        delete_data(10)
        delete_data(12)
    
can be re-created. The `autoloaded` flag is set for all the `FileStore` objects (there's actually only one, it is just stored in multiple places), mainly because it semantically makes sense.
    
b) there is *limited* support for tracking cases where a dataset is loaded in via a `load_xxx` call and then the values are changed by the used. This is currently only done for PHA data with the grouping and quality arrays.
    
c) we assume the file names and locations remain consistent: that is, if a user has called
    
        load_image("../bar/foo.fits")
    
then the file has not been moved, deleted, or changed.
    
d) Technically we allow IO backend objects to be sent to the calls rather than the filename (e.g. a AstroPy FITS HDU list). However, this is not well advertised, so maybe we can say that this is only expected to work for `unpack_xxx` calls and not `load_xxx` ones?

Alternate design
----------------
    
The tracking could instead have been made of the `unpack_xxx` calls, since all `load_xxx` calls end up calling `unpack_xxx`, and this would allow cases like `pha = unpack_pha(infile); set_data(pha)` to be tracked. However, it would complicate things even more (since the data cannot just be tracked by `IdType` because at the time of the `unpack_xxx` call we only have the filename), and the `save_all` serialization isn't meant to handle all cases, just "simple" cases (users who are handling more-complex situations are expected to be able to write their own scripts).
    
Known Issues
------------
    
These are issues with the current serialization code (but also the previous version too). We could track some sort of identifier for each file - e.g. the md5sum of the file - which could be used to identify some parts of issues b and c - but I argue that this is beyond the scope of the `save_all` call (i.e. this is just a limitation of how it works).

### DIFF
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -186,7 +186,7 @@ def _save_entries(out: OutType,
     store
        A container with keys. The elements of the container are
        passed to tostatement to create the string that is then
-       written to fh.
+       saved in out.
     tostatement : func
        A function which accepts two arguments, the key and value
        from store, and returns a string. The reason for the name
@@ -374,9 +374,6 @@ def _save_pha_grouping(out: OutType,
     bid
        If not ``None`` then this indicates that the background dataset
        is to be used.
-    fh : None or file-like
-       If ``None``, the information is printed to standard output,
-       otherwise the information is added to the file handle.
     """
 
     _save_pha_array(out, state, pha, "grouping", id, bid=bid)
@@ -1718,10 +1715,10 @@ def save_all(state: SessionType,
 
     _save_id(out, state)
 
-    if fh is None:
-        fh = sys.stdout
+    outfh = sys.stdout if fh is None else fh
 
-    write = lambda msg: fh.write(f"{msg}\n")
+    def write(msg: str) -> None:
+        outfh.write(f"{msg}\n")
 
     # Hard code the required imports.
     #
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -891,7 +891,10 @@ def _save_fit_method(out: OutType, state: SessionType) -> None:
         # return 'set_method_opt("{}", {})'.format(key, val)
         return 'set_method_opt("%s", %s)' % (key, val)
 
-    _save_entries(out, state.get_method_opt(), tostatement)
+    _save_entries(out,
+                  # state.get_method_opt(),  # all values
+                  state._current_method_changed,  # changed values only
+                  tostatement)
     _output_nl(out)
 
 
@@ -921,7 +924,10 @@ def _save_iter_method(out: OutType, state: SessionType) -> None:
         # so it makes no difference.
         return f'set_iter_method_opt("{key}", {val})'
 
-    _save_entries(out, state.get_iter_method_opt(), tostatement)
+    _save_entries(out,
+                  # state.get_iter_method_opt(),  # all values
+                  state._current_itermethod_changed,  # changed values only
+                  tostatement)
     _output_nl(out)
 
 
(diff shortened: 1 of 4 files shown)


DougBurke:
Rebased as #2476 has been merged (and to address a merge conflict).

Issue #2427, opened 2026-03-05
arviz release 1.0 breaks things


DougBurke:
arviz version 1.0.0 was released this week. They've obviously changed the API since our tests now fail with things like

```
    def mcmc_to_arviz(mcmc : MCMC, fit: Fit, list_of_draws : list[tuple],
                      ):
        """Convert the MCMC results to an ArviZ InferenceData object.
    
        ArviZ is a Python package for exploratory analysis of Bayesian models and
        provides a range of tools for visualizing and analyzing MCMC results.
    
        This function requires the `ArviZ <https://python.arviz.org>`_ package to be installed.
    
        .. versionadded:: 4.17.1
    
        Parameters
        ----------
        mcmc : MCMC
            The MCMC object used to run the chain.
        fit : Fit
            The Fit object used to fit the data.
        list_of_draws : list of tuples
            A list of tuples, each containing three elements as returned from
            `MCMC.get_draws`: the statistic, the acceptance rate, and the
            parameter array. Each tuple corresponds to a single chain run for the
            same model and MCMC object.
    
        Returns
        -------
        inference_data: arviz.InferenceData
            An ArviZ InferenceData object containing the posterior and log likelihood data.
        """
    
>       from arviz.data.base import dict_to_dataset
E       ModuleNotFoundError: No module named 'arviz.data'

sherpa/sim/__init__.py:1044: ModuleNotFoundError
```

We now have

```
% python
Python 3.13.12 | packaged by conda-forge | (main, Feb  5 2026, 05:53:46) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import arviz
>>> arviz.__version__
'1.0.0'
>>> import arviz.base
Traceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    import arviz.base
ModuleNotFoundError: No module named 'arviz.base'
>>> import arviz_base
```

DougBurke:
I was wondering why it was "temperamental", since it hasn't been showing up in CI runs, but I think I've realized why

- arviz 1.0.0 requires python >= 3.12
- our test probably uses an earlier python version, so arviz isn't installed there, hence the test fails

DougBurke:
I've looked at this and there's two changes

 - import from `arviz_base` for the dict-related routine
   - it's not clear it we now just require `arviz_base` rather than `arviz` or also want the other `arviz_xxx` packages (this is mainly a documentation issue)
[...]

PR #2448, opened 2026-04-22
Update the documentation to build with recent versions of Sphinx

DougBurke:
# Summary

Allow the documentation to be built with recent versions of Sphinx and update the Read-The-Docs configuration to match.

# Details

The RTD setup has not been updated for a while, so take the time to do so:

- newer build OS (ubuntu-lts-latest)
- newer "python" version (miniforge3-latest; the existing RTD documentation is a bit unclear but it says all the `mambaforge` variants are deprecated so I assume switching to `miniforge3` makes sense)
  - the actual python version has been updated to 3.13
- the upper limit on sphinx has been removed (so we can use 9.1, which is the latest).
- added an upper-limit on arviz since we know version 1.0 is not compatible with Sherpa (issue #2427)

Note that the installation now tends to pick the "latest" version, at least for the low-level code, to (hopefully) reduce the need for future updates.

However, the cost of this is that Sphinx has got "pickier" about how we document methods (which is why we added a "sphinx < 8.0.0" requirement many moons ago), so there are are number of changes needed before we can do this, including

- be somewhat more-precise when giving a "type" for a parameter or a return value
  - this can include the fact that adding text like "len(x0)" or "shape(N)" ends up with Sphinx either failing to find a label for `x0` or for coming up with a link to a shape method which is completely unrelated to the intent, so we remove these (and rely on the text to describe these constraints)
- a module listing accidentally ended up with a repeated entry, which caused some issues 
- fix some documentation issues in the plotting backend
  - use the actual parameter names (`xlo`, `xhi`) rather than `x0`/`x1` for  the histogram routine

As part of this I also "cleaned up" several "See Also" sections, going from one-routine-per-line of the form "label : description", to "label1, label2, ...., labelN" because I originally added them this way because it seemed more-useful and the way it was done. However, it "wastes" more vertical screen space, and has the issue of requiring more work when even the label description changes. So I have been moving to the "multi-label" form when I can, but I just want to do this on an ad-hoc basis rather than change everything at once as it has the potential for causing a lot of review conflicts.

### DIFF
--- a/sherpa/plot/backend_utils.py
+++ b/sherpa/plot/backend_utils.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2022 - 2024
+#  Copyright (C) 2022-2024, 2026
 #  MIT
 #
 #
@@ -117,8 +117,8 @@ def add_kwargs_to_doc(param_doc):
     --------
 
     >>> param_doc = {'c' : ['int', 'thickness of line'],
-    ...              'title' : ['string', 'Title of figure (only use if `overplot=False`)'],
-    ...              'color': ['string or number',
+    ...              'title' : ['str', 'Title of figure (only use if `overplot=False`)'],
+    ...              'color': ['str or number',
     ...                  'any matplotlib color with a really long text attached to it that will not fit in one line of text in the docstring']}
     >>> @add_kwargs_to_doc(param_doc)
     ... def test_func2(a, *, title=None, color='None'):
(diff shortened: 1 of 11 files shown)


sherpa/plot/backend_utils.py
```diff
@@ -141,9 +141,9 @@ def add_kwargs_to_doc(param_doc):
         ----------
         a : int
             Our stuff
-        title : string, default=None
+        title : str, default=None
```
hamogu:
```suggestion
        title : str or None, default=None
```

sherpa/plot/backend_utils.py
```diff
@@ -141,9 +141,9 @@ def add_kwargs_to_doc(param_doc):
...
             Title of figure (only use if `overplot=False`)
-        color : string or number, default=None
+        color : str or number, default=None
```
hamogu:
```suggestion
        color : str or number or None, default=None
```

hamogu (approved):

PR #2452, opened 2026-05-06
BUILD: add the disable_extern build option

DougBurke:
This is not needed if we take #2484

# Summary

Simplify building Sherpa within CIAO by setting the disable_extern setting in setup.cfg.

# Details

If the disable_extern option is set to True then the configure script in the extern directry is not run. This is equivalent to creating the file "extern/built" before building Sherpa (this option remains, so existing builds of Sherpa with CIAO should just work).

The scripts/use_ciao_config script has been updated to set it (and remove changing the "configure" option as it is no-longer relevant).

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -221,8 +221,7 @@ replaced by its actual value, and the ``xspec_version`` line
 should be updated to match the output above::
 
     install_dir=${ASCDS_INSTALL}
-
-    configure=None
+    disable_extern=True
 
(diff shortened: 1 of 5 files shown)


DougBurke:
It is not clear to me that this is worth it (given that I want to move to #1948 or equivalent).

DougBurke:
Closed by #2484 although the meson build currently does not have an equivalent to the proposed `disable_extern` option, but it hopefully isn't needed.

PR #2453, opened 2026-05-11
BUILD: change location of DS9 artifacts

DougBurke:
# Summary

Update the DS9 used in the CI tests. There is no functional change.

# Details

The download/ directory for DS9 contains the "latest" version, whereas the archive/ directory also contains older versions. So switch to use the archive/ directory to avoid issues where the download/ directory gets cleared out of the version that we currently test against.

As part of this, bump the DS9 version from 8.6 to 8.7 which requires bumping up the Ubuntu version, and to add "platform" support to the code (at the moment it's all x86 but if we add back the macOS testing or maybe aarch64 then it will become important to handle).

### DIFF
--- a/.github/scripts/setup_ds9.sh
+++ b/.github/scripts/setup_ds9.sh
@@ -6,7 +6,10 @@
 # would need to identify x86 versus ARM, as well as update the OS).
 #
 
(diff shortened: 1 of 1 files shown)


harlantc (approved):
LGTM

PR #2455, opened 2026-05-11
BUILD: ensure PYTHONPATH set for sherpa_smoke runs

DougBurke:
# Summary

Ensure that the necessary ancillary libraries are available for the sherpa_smoke tests in the CI runs. There is no functional change in the code.

# Details

The develop build needs the PYTHONPATH environment set so that sherpa_smoke can find the group and stk libraries (this is primarily to avoid extra screen messages as the soke test doesn't exercise either the stack or group code).

This requires #2453 in order for the tests to pass.

### DIFF
--- a/.github/scripts/setup_ds9.sh
+++ b/.github/scripts/setup_ds9.sh
@@ -6,7 +6,10 @@
 # would need to identify x86 versus ARM, as well as update the OS).
 #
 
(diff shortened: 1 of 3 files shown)


DougBurke:
This looks like it's not needed if we take #2484

DougBurke:
Closed by #2484

PR #2456, opened 2026-05-13
Reallow ds9 on macos tests

DougBurke:
# Summary

Force XPA_METHOD to local for the CI DS9 tests to allow the macOS/ARM build to include DS9 when testing.  There is no functional change.

# Details

This requires #2453

In #1938 we removed the support for DS9 testing for macOS on the GitHub CI runs because they were failing. It looks like this was because the macOS version of DS9 has trouble with the default XPA_METHOD setting (inet), so switch all the tests to use XPA_METHOD=local (even those not using DS9 because it should not matter). Only the ARM build is supported because the Intel build test runs slowly.

As a side note, the CIAO version of DS9 automatically forces the local setting for XPA_METHOD, which is why this was not seen in CIAO builds for macOS.

There is an argument to be said that we should force the local method for communicating with DS9 (unless over-ridden), but that is a surprisingly more-involved change and IS done in some planned follow-up work: #2459

### DIFF
--- a/.github/scripts/setup_ds9.sh
+++ b/.github/scripts/setup_ds9.sh
@@ -6,7 +6,10 @@
 # would need to identify x86 versus ARM, as well as update the OS).
 #
 
-ds9_base_url=https://ds9.si.edu/download/
+# The download/ version is the "latest" release whereas archive/ includes
+# older releases.
+# ds9_base_url=https://ds9.si.edu/download
+ds9_base_url=https://ds9.si.edu/archive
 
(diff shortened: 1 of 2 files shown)


DougBurke:
I've "fixed" this in Sherpa itself in #2459 (by defaulting to using the `local` value for `XPA_METHOD`) so we do not need this PR (and having it would just lead to merge conflicts). So I'm closing it.

DougBurke:
Actually, the code has been moved to #2465

PR #2457, opened 2026-05-13
Update arviz

### LINKED ISSUE #2427: arviz release 1.0 breaks things

hamogu:
## Summary
Sherpa has a function to turn MCMC output into the format used by the arviz package. This PR updates the API for compatability with arviz >=1.0.0

## Details
There are two commits: The first one changes Sherpa itself, the second one the CI setup.
The second steo is necessary, because we used to run the tests that depends on arviz with Python 3.10 and we need a more modern Python version for that now.

The change is not backwards compatible for arviz versions, i.e. it will no longer work for arviz < 1.0.0. They did a significant cur in API for their 1.0.0 release and this is not a widely used feature in Sherpa. We can expect users to upgrade to the current verion of arviz (actually, at the time of writing, arviz has already releases 1.1).

closes #2427

### DIFF
--- a/.github/workflows/ci-pip-workflow.yml
+++ b/.github/workflows/ci-pip-workflow.yml
@@ -43,15 +43,12 @@ jobs:
             bokeh-pkg: 'bokeh>=3,<4'
             ncores: 1
 
-          - name: Linux Build (w/o Astropy or Xspec)
+          - name: Linux Build (w/o Matplotlib, w/o Astropy or Xspec)
             os: ubuntu-latest
             python-version: "3.11"
             numpy-pkg: 'numpy'
             install-type: install
             test-data: package
-            matplotlib-pkg: 'matplotlib>=3,<4'
-            bokeh-pkg: 'bokeh>=3,<4'
-            arviz-pkg: 'arviz'
             optimagic-pkg: 'optimagic'
 
           - name: Linux Build (w/o Matplotlib, Xspec, or test data)
@@ -62,11 +59,14 @@ jobs:
             fits-pkg: 'astropy'
             test-data: none
 
-          - name: Linux Build (submodule data w/o Matplotlib or Xspec)
+          - name: Linux Build (submodule data without or Xspec)
             os: ubuntu-latest
             python-version: "3.12"
             numpy-pkg: 'numpy'
             install-type: develop
+            matplotlib-pkg: 'matplotlib>=3,<4'
+            bokeh-pkg: 'bokeh>=3,<4'
+            arviz-pkg: 'arviz>=1.0.0'
             fits-pkg: 'astropy'
             test-data: submodule
 
--- a/.github/workflows/ci-pip-workflow.yml
+++ b/.github/workflows/ci-pip-workflow.yml
@@ -43,7 +43,7 @@ jobs:
             bokeh-pkg: 'bokeh>=3,<4'
             ncores: 1
 
-          - name: Linux Build (w/o Matplotlib, w/o Astropy or Xspec)
+          - name: Linux Build (w/o Matplotlib, Astropy or Xspec)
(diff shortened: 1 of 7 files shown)


hamogu:
Of course, the tests fail because we ds9 install doesn't work, so we'll have to wait for #2453 to land before we can expect the tests to run here.

.github/workflows/ci-pip-workflow.yml
```diff
@@ -62,11 +59,14 @@ jobs:
             fits-pkg: 'astropy'
             test-data: none
 
-          - name: Linux Build (submodule data w/o Matplotlib or Xspec)
+          - name: Linux Build (submodule data without or Xspec)
```
DougBurke: So, we have a somewhat esoteric set of tests to cover our bases here and it might be useful to review them. However, that's not for this PR...

At the least we would want to fit the `name` field (e.g. remove the `or `)

```suggestion
          - name: Linux Build (submodule data without Xspec)
```

.github/workflows/ci-pip-workflow.yml
```diff
@@ -43,15 +43,12 @@ jobs:
             bokeh-pkg: 'bokeh>=3,<4'
             ncores: 1
 
-          - name: Linux Build (w/o Astropy or Xspec)
+          - name: Linux Build (w/o Matplotlib, w/o Astropy or Xspec)
```
DougBurke:
```suggestion
          - name: Linux Build (w/o Matplotlib, Astropy or Xspec)
```

I'm not getting into discussing Oxford comma rules here ;-)

DougBurke:
This could be rebased on top of #2453

We can see it working via the docs at https://sherpa--2457.org.readthedocs.build/en/2457/mcmc/index.html#connections-to-arviz compared to https://sherpa.readthedocs.io/en/4.18.0/mcmc/index.html#connections-to-arviz (although the newer version is not quite as "nice", that's outside our wheel-house)

hamogu:
Rebased onto https://github.com/sherpa/sherpa/pull/2453 as it changes the DS9 download code (and updates versions).

DougBurke (approved):

PR #2458, opened 2026-05-13
Minor clean-up of the image code

DougBurke:
# Summary

Improve test coverage of the DS9 code and improve the documentation and typing statements.

# Details

This is the first part of #2413 (after some rearrangement) and depends on (mainly to avoid merge conflicts)

- #2462 
- #2465

There are a lot of changes to add minor levels of documentation and typing statements. Note that not everything is documented or has typing statements added because the follow-on commit 

- #2459
 
deals with that code (by doing the changes here it makes it more obvious what parts of the image code can be removed or cleaned up, and that's left for #2459).

The main user-visible changes are as follows (although I don't expect users to really come across them):

- `ComponentSourceImage` has "moved" in that it now derives from `SourceImage` rather than `ModelImage`; I would be surprised if anyone notices
- `Image.set_wcs` is now a static method, so needs to be called differently; again, I don't expect users to be calling this method directly

Technically the changes to `dummy_backend.py` are "breaking changes", since several routines have gone from `pass` to `return ""` to match the typing statement. Since the dummy code is not really meant to be run I am happy with this. I do not think it worth it to change the backend code to use a class, as was done for the plotting code.

Some of the code changes include

- use of f-strings (apart from logging calls where we want it to be lazy)
- change from `staticmethod(name)` to using the `@staticmethod` decorator as it removes warnings from some lint/type-checker tools and I find it easier to read 
  - also converted a method to being `@staticmethod` (`Image.set_wcs`) which is a breaking change but I do not expect users to be calling these objects directly
- simplify some of the `try/except` handling, including
  - avoid if the exception is just re-thrown
  - try to catch `Exception` rather than any exception (e.g. in case a user is trying to control-c something), although there are some places where we catch `BaseException` and it's a bit of a "dealer's choice" of what we should catch
  - as part of the exception handling try to re-throw the caught exception to make the stack traces easier to follow
- use `super()` rather than the parent class name
- avoid a `return` if the return type is `None`
- the `ds9_backend._set_wcs` routine has had its signature changed (instead of sending in a three-element tuple send the individual arguments); as this routine is intended to be local to this module I am not worried about changing the API
 
This code originally came from #2411, then separated into #2413. In those versions I also spent time doing some clean up of the class structure for the image classes. I have dropped that from these versions for now. I have also split the code up into a number of PRs to make it a bit easier to see where the "interesting" parts are. This part is not interesting, but helped me better understand the code for the more-interesting work to come (although this statement is less true now as I moved some of that "interesting" work to earlier PRs in the sequence).

### DIFF
--- a/sherpa/image/__init__.py
+++ b/sherpa/image/__init__.py
@@ -198,10 +198,16 @@ class DataImage(Image):
         self.eqpos = getattr(data, 'eqpos', None)
         self.sky = getattr(data, 'sky', None)
         header = getattr(data, 'header', None)
-        if header is not None:
-            obj = header.get('OBJECT')
-            if obj is not None:
-                self.name = str(obj).replace(" ", "_")
+
+        # Clear out any previous version.
+        self.name = "Data"
+
+        if header is None:
+            return
+
+        obj = header.get('OBJECT')
+        if obj is not None:
+            self.name = str(obj).replace(" ", "_")
 
     def image(self, shape=None, newframe=False, tile=False):
         Image.image(self, self.y, shape, newframe, tile)
--- a/sherpa/image/__init__.py
+++ b/sherpa/image/__init__.py
@@ -240,6 +240,15 @@ class DataImage(Image):
         Image.__init__(self)
 
     def prepare_image(self, data: Data2D) -> None:
+        """Create the data to display.
+
+        Parameters
+        ----------
+        data
+           The Sherpa data object to display.
+
+        """
+
         self.y = data.get_img()
         self.eqpos = getattr(data, 'eqpos', None)
         self.sky = getattr(data, 'sky', None)
@@ -260,6 +269,18 @@ class DataImage(Image):
               newframe: bool = False,
               tile: bool = False
               ) -> None:
+        """Send the data to the image viewer to display.
+
+        Parameters
+        ----------
+        shape
+           The shape of the data (optional).
+        newframe
+           Should the pixels be displayed in a new frame?
+        tile
+           Should the display be tiled?
+
+        """
         Image.image(self, self.y, shape, newframe, tile)
         Image.set_wcs((self.eqpos, self.sky, self.name))
 
@@ -282,6 +303,17 @@ class ModelImage(Image):
         Image.__init__(self)
 
     def prepare_image(self, data: Data2D, model: Model) -> None:
+        """Create the data to display.
+
+        Parameters
+        ----------
+        data
+           The Sherpa data object to display.
+        model
+           The model to evaluate on the data grid.
+
+        """
+
         self.y = data.get_img(model)
         self.y = self.y[1]
         self.eqpos = getattr(data, 'eqpos', None)
@@ -292,6 +324,18 @@ class ModelImage(Image):
               newframe: bool = False,
               tile: bool = False
               ) -> None:
+        """Send the data to the image viewer to display.
+
+        Parameters
+        ----------
+        shape
+           The shape of the data (optional).
+        newframe
+           Should the pixels be displayed in a new frame?
+        tile
+           Should the display be tiled?
+
+        """
         Image.image(self, self.y, shape, newframe, tile)
         Image.set_wcs((self.eqpos, self.sky, self.name))
 
@@ -304,8 +348,16 @@ class SourceImage(ModelImage):
         self.name = 'Source'
 
     def prepare_image(self, data: Data2D, model: Model) -> None:
-        # self.y = data.get_img(model)
-        # self.y = self.y[1]
+        """Create the data to display.
+
+        Parameters
+        ----------
+        data
+           The Sherpa data object to display.
+        model
+           The model to evaluate on the data grid.
+
+        """
 
         self.y = data.eval_model(model)
         data._check_shape()
@@ -316,7 +368,13 @@ class SourceImage(ModelImage):
 
 
 class RatioImage(Image):
-    """The data divide by the model."""
+    """The data divided by the model.
+
+    See Also
+    --------
+    ResidImage
+
+    """
 
     _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
     """The fields to include in the string output.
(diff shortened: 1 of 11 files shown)


DougBurke:
Rebased to make easier to review as #2465 has been merged

sherpa/image/dummy_backend.py
```diff
@@ -17,44 +18,123 @@
...
+
+from sherpa.astro.io.wcs import WCS
+
+
 imager = None
+"""The DS9 window, or None"""
 
 
-def close(*args, **kwargs):
+def close() -> None:
```
hamogu, line 39: This (and the below) are technically breaking changes since before it would have accepted any parameter (and just ignored it) and not passing in parameters is an error.

sherpa/image/ds9_backend.py
```diff
@@ -158,29 +196,49 @@ def _set_wcs(keys: tuple[WCS | None, WCS | None, str]) -> str:
...
+    keys
+       The eqpos and sky transforms, and the name of the display.
+
+    """
 
     if not imager.isOpen():
         raise DS9Err('open')
 
-    info = _set_wcs(keys)
+    info = _set_wcs(keys[0], keys[1], keys[2])
```
hamogu:
```suggestion
    info = _set_wcs(*keys)
```
Of course, either works. Just so you know that I actually read the code.

sherpa/image/ds9_backend.py
```diff
@@ -158,29 +196,49 @@ def _set_wcs(keys: tuple[WCS | None, WCS | None, str]) -> str:
 
 
 def wcs(keys: tuple[WCS | None, WCS | None, str]) -> None:
+    """Send the WCS information to the image viewer.
+
+    Parameters
+    ----------
+    keys
```
hamogu:
```suggestion
    keys : tuple
```

sherpa/image/ds9_backend.py
```diff
@@ -60,46 +85,59 @@ def get_region(coord):
...
+    """Send the data to the image viewer to display.
+
+    Parameters
+    ----------
+    array
+       The pixel values
+    newframe
+       Should the pixels be displayed in a new frame?
+    tile
+       Should the display be tiled?
```
hamogu:
```suggestion
    array : np.adarray
       The pixel values
    newframe : bool
       Should the pixels be displayed in a new frame?
    tile : bool
       Should the display be tiled?
```

sherpa/image/dummy_backend.py
```diff
@@ -17,44 +18,123 @@
...
+    """Send the data to the image viewer to display.
+
+    Parameters
+    ----------
+    array
+       The pixel values
+    newframe
+       Should the pixels be displayed in a new frame?
+    tile
+       Should the display be tiled?
```
hamogu:
```suggestion
    array : np.ndarray
       The pixel values
    newframe : bool
       Should the pixels be displayed in a new frame?
    tile : bool
       Should the display be tiled?
```

sherpa/image/__init__.py
```diff
@@ -95,25 +108,59 @@ def get_region(coord):
...
+        Parameters
+        ----------
+        array
+           The pixel values
+        shape
+           The shape of the data (optional).
+        newframe
+           Should the pixels be displayed in a new frame?
+        tile
+           Should the display be tiled?
```
hamogu:
```suggestion
        array : np.adnarray
           The pixel values
        shape : tuple or None
           The shape of the data (optional).
        newframe : bool
           Should the pixels be displayed in a new frame?
        tile : bool
           Should the display be tiled?
```

sherpa/image/__init__.py
```diff
@@ -95,25 +108,59 @@ def get_region(coord):
...
         newframe = bool_cast(newframe)
         tile = bool_cast(tile)
         if shape is None:
-            backend.image(array, newframe, tile)
+            vals = array
         else:
-            backend.image(array.reshape(shape), newframe, tile)
+            vals = array.reshape(shape)
+
+        backend.image(vals, newframe, tile)
```
hamogu, line 139:
```suggestion
        tile = bool_cast(tile)
        if shape:
             array = array.reshape(shape)

        backend.image(array, newframe, tile)

```

sherpa/image/__init__.py
```diff
@@ -312,44 +462,109 @@ def __init__(self):
...
+           The shape of the data (optional).
+        newframe
+           Should the pixels be displayed in a new frame?
+        tile
+           Should the display be tiled?
+
+        """
         Image.image(self, self.y, shape, newframe, tile)
-        Image.set_wcs(self, (self.eqpos, self.sky, self.name))
+        Image.set_wcs((self.eqpos, self.sky, self.name))
```
hamogu: Maybe it would be clearer with set_wcs took three separate arguments instead of a tuple? Or do you not want to do that because of backwards compatibility?

hamogu (commented):
Since @DougBurke hasn't asked for a review yet, I'm not approving or "request changes" but just commenting for now.

sherpa/image/__init__.py
DougBurke: Well, the API for `set_wcs` is changed in this PR (by making it a `staticmethod`) so we can change the calling convention (it is not something I expect most users to be calling).

sherpa/image/ds9_backend.py
DougBurke: I'm conflicted here. The HTML output is significantly more informative with the type added, but we technically already have this information with the type annotations. So I've taken to avoiding adding them (as it is just repeated information that needs to be kept up to date).

However, in #2519 I found the HTML display to be so lacking I'm going back to adding the information.

So decisions, decisions!

DougBurke:
@hamogu I should have addressed most of your comments. Thanks for the pre-review review!

DougBurke:
Who would have thought that committing untested code could lead to problems ....

sherpa/image/__init__.py
DougBurke: So I went with changing it.

sherpa/image/dummy_backend.py
DougBurke: I added some documentation to point this out.

hamogu (approved):

PR #2459, opened 2026-05-13
Simplify the initialization of the DS9 module

DougBurke:
# Summary

Simplify the DS9 initialization code and make DS9 8.7 the minimum suggested version.

# Details

Requires
 - #2462 
 - #2465 
 - #2458 

The aim is to clean up the ds9-related code (extending the clean up from #2458) by dropping unused code and updating to the latest ds9 capabilities.

Changes include

- removing commented-out code
- removing methods unused by Sherpa (e.g. `sherpa.image.DS9.DS9Win.showFITSFile`)
- drop the unused `dataFunc` argument from the "xpa set" calls
- greatly simplify how the `sherpa.image.DS9` module sets things up; this is done in steps rather than all-at-once just to make it easier to follow
  - one of the main improvements is no-longer having to create a subclass of `subprocess.Popen` as there is no benefit for Sherpa to allow DS9 to be installed after a Sherpa session has started
- drop the unused extra arguments to `sherpa.image.DS9.DS9Win.showArray`
- fix a rarely-hit bug when sent uint64 data

One interesting change here is that ds9 8.7 properly supports unsigned int64 images and this commit makes several changes to the "what datatypes are supported" logic. Do we want to "enforce" a minimum ds9 version (in that it's up to the user to update if the display doesn't work, and not that we error out if the ds9 version is too old)? I have decided it is not worth the complexity of code and we just note that DS9 8.7 or later is suggested.n

### DIFF
--- a/sherpa/image/DS9.py
+++ b/sherpa/image/DS9.py
@@ -1,6 +1,6 @@
 #
 #  Copyright (C) 2006-2010, 2016-2021, 2025-2026
-#     Smithsonian Astrophysical Observatory
+#  Smithsonian Astrophysical Observatory
 #
 #
 #  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


DougBurke:
Some of these changes can be moved back into earlier commits (e.g. those that fix up typos).

The WCS documentation change can be moved to a separate PR but it needs code in this PR (and was done just because I noted that the WCS code was lacking when reviewing some of the documentation and typing changes made here).

**EDITED TO ADD** this comment is now somewhat out of date.

DougBurke:
Rebased as #2465 has been merged.

DougBurke:
Rebased as #2458 has been merged.

docs/plots/image_backends.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Which backend is used?
+======================
+
+When the `sherpa.image` module is first imported, Sherpa tries to
+import the `sherpa.image.ds9_backend` module, falling back to
+`sherpa.image.dummy_backend` if DS9 and the XPA tools can not be
+found. The logic for sending data to and from DS9, as well as
+starting and stopping the DS9 application, it provided by the
+``sherpa.image.DS9`` module.
```
hamogu, line 20:
```suggestion
`sherpa.image.DS9` module.
```

sherpa/image/DS9.py
```diff
@@ -258,54 +181,34 @@ def xpaset(cmd: str,
...
+#     -------------     --------------------
+#     unsigned 8-bit          32 bit
+#              16-bit         64-bit
+#     unsigned 16-bit
+#              32-bit
+#              64-bit
+#
+# There is no attempt to make this conversion match the capabilities
+# of the DS9 being used.
+#
```
hamogu, line 201: I don't understand what that comment means. Maybe just delete it?

sherpa/image/DS9.py
```diff
@@ -258,54 +181,34 @@ def xpaset(cmd: str,
...
-    return cnvDict
+# What data types need to be converted before sending to DS9?
+# DS9 8.7 supports
+#
+#     Integer types     Floating-point types
+#     -------------     --------------------
+#     unsigned 8-bit          32 bit
+#              16-bit         64-bit
+#     unsigned 16-bit
+#              32-bit
```
hamogu, line 196: Five lines below, there is also `np.uint32: np.int64`. I suggest to just remove the specific table here. THe code is very readable and then we don't have to worry to keep it in sync.

sherpa/image/DS9.py
```diff
@@ -258,54 +181,34 @@ def xpaset(cmd: str,
...
+# There is no attempt to make this conversion match the capabilities
+# of the DS9 being used.
+#
+_CnvDict = {
+    np.int8: np.int16,
+    np.uint16: np.int32,
+    np.uint32: np.int64,
+}
 
+if hasattr(np, "uint64"):
```
hamogu, line 208: I believe every numpy today has uint64, because it's provided by the C99 standard (https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uint64).

sherpa/image/DS9.py
```diff
@@ -422,30 +303,30 @@ def isOpen(self) -> bool:
...
+
+        Parameters
+        ----------
+        arr
+           An array (expected to be NumPy but need not be) that must
+           be 2-d (axis order is y,x) or 3-d (axis order is z,y,x).
+
+        Notes
+        -----
+        Complex data types cause an invalid. Other data types may be
```
hamogu, line 325:
```suggestion
        Complex data types cause an exception. Other data types may be
```

hamogu (changes requested):
I like this. I think it'll make the code easier to maintain in the future. My suggetions are all minor, and woudl result in cutting even more.

docs/plots/image_backends.rst
DougBurke: Unfortunately this won't provide a usable link since we need ds9+xpa to be installed to be able to build the documentation for it. So this at least provides a visual reason that the "link" isn't active.

sherpa/image/DS9.py
DougBurke: The idea is that we could check the ds9 version and "do the best thing" for that particular version.  That way lies madness (since we assume we have a relatively recent ds9). If we're going to remove the following lines then it definitely doesn't make sense.

sherpa/image/DS9.py
DougBurke: Yeah - this is quite old code I touched, so we can just assume we have it.

docs/plots/image_backends.rst
hamogu: Understood.

hamogu (approved):

PR #2462, opened 2026-05-19
Reset the title field for new image displays

### LINKED ISSUE #2468: sending DS9 WCS info when there is none
In looking through the image code (e.g. #2459) I've seen some issues with recent DS9 (8.7 or so) when we send it an image with no WCS data - such as the `sherpa-test-data/sherpatest/img.fits` dataset. You can see warnings like

```
!! AST: Error in routine wcsSystem at line 9 in file frame/wcsast.C.
!  Invalid Object pointer given (value is zero).
```

This could in part be a DS9 issue, but here we call the XPA `wcs replace` endpoint with the data

```
OBJECT = 'Data'
```

which has no WCS data. The simplest solution is to add in a "fake" identity WCS in this case - such as

```
OBJECT = 'Data'
WCSNAMEP = 'PHYSICAL'
CTYPE1P = 'x'
CTYPE2P = 'y'
CRPIX1P = 1.0
CRPIX2P = 1.0
CDELT1P = 1.0
CDELT2P = 1.0
CRVAL1P = 1.0
CRVAL2P = 1.0
```

I don't think DS9 has a way to set an arbitrary FITS header item via XPA (which would be the alternative approach in this case) but I may be missing something.

DougBurke:
# Summary

Ensure that the name/title of a DS9 image is reset in case subsequent displayed images do not have the OBJECT field set in their header. Ensure that images with no WCS have an identity transform added when displayed in DS9 to fix #2468.

# Details

This is another PR taken from #2458 / #2413.

First we add some tests. The main one for this PR is in `sherpa/astro/ui/tests/test_astro_session.py::test_get_data_image_dataimg_repeated` but there are also some other tests that were added to improve test coverage when developing the larger-scale changes.

The actual bug fix is that we were taking the `name` field for `DataImage` objects from the header, but not clearing it when a new dataset was loaded. The fix is pretty easy.

We then clean up the code used to send the `WCS` information to DS9 (which includes the OBJECT field) to avoid confusing the type checkers. If there is no WCS data then add in an identity transform for the PHYSICAL system. This is to stop DS9 from creating warning messages from the AST library (these can also be seen if the DS9 session is saved and then restored, and are "not a problem" to Sherpa itself). This is to address #2468. Note that since writing this the latest development build of DS9 has been updated to not print the warning message if the `wcs replace` command is just sent the `OBJECT` keyword but I believe this is still a worthwhile clean up.

I have had discussions with the DS9 maintainer about whether there could be an XPA command to just set the OBJECT keyword but

- that would require a minimum ds9 version for CIAO
- it turns out not to be easy and so is not going to be done

### DIFF
--- a/sherpa/astro/io/wcs.py
+++ b/sherpa/astro/io/wcs.py
@@ -49,6 +49,11 @@ except ImportError:
         raise RuntimeError("No WCS support")
 
 
+__doctest_requires__ = {
+    '*': ['sherpa.astro.utils._wcs']
+}
+
+
 class WCS(NoNewAttributesAfterInit):
     """Represent a World Coordinate System transformation.
 
--- a/sherpa/astro/ui/tests/test_astro_session_image.py
+++ b/sherpa/astro/ui/tests/test_astro_session_image.py
@@ -243,7 +243,7 @@ def test_get_data_image_dataimg_repeated():
 
     # What is the name field now?
     obj = s.get_data_image()
-    assert obj.name == 'A_big_blob'  # ideally this would be "Data"
+    assert obj.name == 'Data'
 
 
 @requires_ds9
--- a/sherpa/astro/ui/tests/test_astro_session_image.py
+++ b/sherpa/astro/ui/tests/test_astro_session_image.py
@@ -340,7 +340,18 @@ def test_wcs_with_no_wcs(make_data_path, tmp_path, check_str):
 
     cts = wcsfile.read_text()
 
-    lines = []
+    lines = [
+        "WCSAXES =                    2 / Number of WCS axes",
+        "WCSNAME = 'PHYSICAL'           / Reference name for the coord. frame",
+        "CRPIX1  =                  1.0 / Reference pixel on axis 1",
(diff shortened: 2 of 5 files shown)


DougBurke:
Rebased to address conflict in a test.

DougBurke:
I need to tweak this. I've been talking with @kglotfelty about the `wcs replace` command we use to set the `OBJECT` value and there are "issues" when the file has no WCS. We can "fix" this by sending in a dummy WCS.

**EDITED TO ADD** I have now done this with the "better handle images with no WCS" commit.

DougBurke:
Rebased to pick up recent changes (so the tests should pass, modulo probably the doc build) and to tweak the code slightly.

DougBurke:
Rebased to pick up #2473 and to add more tests - in this case images with only a sky or eqpos transform (and not the case we currently have of either all or neither).

hamogu (approved):
This is mostly added tests, which pass, so it's good. I'm not a DS9 expert, but if the communication between the two works, then testing more can't hurt.

PR #2463, opened 2026-05-19
Image string serialization

DougBurke:
# Summary

Rework the image string code to use the same routine as the plot code. Documentation has been added to the WCS class.

# Details

Requires #2461 (not really for this but for follow-on PRs that depend on both).

As the `display_fields` routine is now used in multiple places it, along with `arr2str` are moved to `sherpa.utils`. This requires a little-bit of code shuffling. This code doesn't quite match that used for a number of the UI code structures (`print_fields`) and I do not believe it is worth combining the two as I try to avoid changing the string representation wherever possible (because I know some people/code can rely on a particular layout).

The WCS code is also converted to use this routine, and as part of it some basic documentation and typing is added to the module (it is not directly related to this PR but I have a lot of related PRs and I don't want to break things up too much). This code could be moved into a separate PR if needed.

### DIFF
--- a/sherpa/astro/io/wcs.py
+++ b/sherpa/astro/io/wcs.py
@@ -189,7 +189,7 @@ class WCS(NoNewAttributesAfterInit):
                x0: float | ArrayType,
                x1: float | ArrayType
                ) -> tuple[np.float64 | np.ndarray, np.float64 | np.ndarray]:
-        """Convert the output coordinates to the inpyt system.
+        """Convert the output coordinates to the input system.
 
         Parameters
         ----------
--- a/sherpa/astro/io/wcs.py
+++ b/sherpa/astro/io/wcs.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2008, 2020, 2021, 2024
+#  Copyright (C) 2008, 2020-2021, 2024, 2026
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -25,10 +25,13 @@ This only supports a limited number of 2D transformations.
 """
 
 import logging
+from typing import overload
 
 import numpy as np
 
-from sherpa.utils import NoNewAttributesAfterInit
+from sherpa.utils import NoNewAttributesAfterInit, arr2str
+from sherpa.utils.types import ArrayType
+
 
 warning = logging.getLogger(__name__).warning
 
@@ -38,16 +41,55 @@ except ImportError:
     warning('WCS support is not available')
 
     def pix2world(*args, **kwargs):
+        "Pixel to World Coordinates"
         raise RuntimeError("No WCS support")
 
     def world2pix(*args, **kwargs):
+        "World to Pixel Coordinates"
         raise RuntimeError("No WCS support")
 
 
 class WCS(NoNewAttributesAfterInit):
+    """Represent a World Coordinate System transformation.
 
-    def __init__(self, name, type, crval, crpix, cdelt,
-                 crota=0.0, epoch=2000.0, equinox=2000.0):
+    This does not support all possible astronomical transforms.
+
+    Parameters
+    ----------
+    name
+       The name of the transformation (expected to be "world" or
+       "physical").
+    type
+       The transform type (expected to be one of "LINEAR", "WCS",
+       or "TAN-P").
+    crval
+       The reference point in the output coordinate system (a
+       two-element sequence).
+    crpix
+       The reference point in the input coordinate system (a
+       two-element sequence).
+    cdelt
+       The width and height of a pixel in the output coordinate
+       system (a two-element sequence).
+    crota
+       The rotation of the transformation.
+    epoch
+       The epoch of the transformation (not always used).
+    equinox
+       The equinox of the transformation (not always used).
+
+    """
+
+    def __init__(self,
+                 name: str,
+                 type: str,
+                 crval: ArrayType,
+                 crpix: ArrayType,
+                 cdelt: ArrayType,
+                 crota: float = 0.0,
+                 epoch: float = 2000.0,
+                 equinox: float = 2000.0
+                 ) -> None:
         self.name = name
         self.type = type
         self.crval = np.asarray(crval, dtype=float)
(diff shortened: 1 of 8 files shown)


sherpa/image/__init__.py
```diff
@@ -205,23 +210,20 @@ def image(self, shape=None, newframe=False, tile=False):
 
 class ModelImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
+    passed through NumPy's array2string.
+    """
+
```
hamogu, line 219: This is not needed, since it's inherited from the parent class `Image`.

sherpa/image/__init__.py
```diff
@@ -252,23 +254,20 @@ def prepare_image(self, data, model):
 
 class RatioImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
+    passed through NumPy's array2string.
+    """
+
```
hamogu, line 263: This is not needed, since it's inherited from the parent class `Image`.

sherpa/image/__init__.py
```diff
@@ -290,23 +289,20 @@ def image(self, shape=None, newframe=False, tile=False):
 
 class ResidImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
+    passed through NumPy's array2string.
+    """
+
```
hamogu, line 298: This is not needed, since it's inherited from the parent class `Image`.

sherpa/astro/io/wcs.py
```diff
@@ -78,14 +116,112 @@ def tostr(vals):
...
+               x0: ArrayType,
+               x1: ArrayType
+               ) -> tuple[np.ndarray, np.ndarray]:
+        ...
+
+    def invert(self,
+               x0: float | ArrayType,
+               x1: float | ArrayType
+               ) -> tuple[np.float64 | np.ndarray, np.float64 | np.ndarray]:
+        """Convert the output coordinates to the inpyt system.
```
hamogu: inpyt -> input

hamogu (changes requested):

sherpa/image/__init__.py
DougBurke: Unfortunately that's not true. With the current design of the `Image` class [*] the "base" class has

```python

    _fields: list[str] = []
    """The fields to include in the string output.

    Names that end in ! are treated as scalars, otherwise they are
    passed through NumPy's array2string.
    """
```

So, while the caption could potentially be dropped (I'm never sure on what happens there), we need to add `_fields = ["name!", ...]` for all the sub-classes, as the `Image` class does not have them.

We **could** change the `display_fields` routine so that it skips fields it does not know about, hence allowing this to only be set in `Image`, but I really don't think that's a good idea.


---

[*] it's also relevant for the `Plot` and `Contour` classes, so it's something we could look at but my PRs [**] for those get quite large - e.g. #1988 and #2076 -  and I'd rather address some simple changes until we come up with an overall agreement on what we want to do

[**] the idea being that we want to separate out the base class, which does the display of the actual data, with the "user" classes where the `plot/contour/image` call sends the data to the base class. So, essentially separate out the two classes, making the user-visible one embed the "base" class rather than extend it. But this is taking us well away from this PR.

sherpa/image/__init__.py
hamogu: OK, let's just leave it as is in this PR.

hamogu (approved):
My concern was due to my error; in fact the code is fine. That leaves one inconsequential typo so I'm approving.

DougBurke:
I'll probably wait for #2461 to land before fixing the typo.

sherpa/image/tests/test_image.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2007, 2015, 2016, 2017, 2018, 2019, 2023
+#  Copyright (C) 2007, 2015-2016, 2017-2019, 2023, 2026
```
wmclaugh:
```suggestion
#  Copyright (C) 2007, 2015-2019, 2023, 2026
```

sherpa/image/__init__.py
```diff
@@ -252,23 +254,20 @@ def prepare_image(self, data, model):
 
 class RatioImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
```
wmclaugh:
```suggestion
    Names that end in ! are displayed directly, otherwise they are
```

sherpa/image/__init__.py
```diff
@@ -205,23 +210,20 @@ def image(self, shape=None, newframe=False, tile=False):
 
 class ModelImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
```
wmclaugh:
```suggestion
    Names that end in ! are displayed directly, otherwise they are
```

sherpa/image/__init__.py
```diff
@@ -290,23 +289,20 @@ def image(self, shape=None, newframe=False, tile=False):
 
 class ResidImage(Image):
 
+    _fields: list[str] = ["name!", "y", "eqpos!", "sky!"]
+    """The fields to include in the string output.
+
+    Names that end in ! are displayed diretly, otherwise they are
```
wmclaugh:
```suggestion
    Names that end in ! are displayed directly, otherwise they are
```

sherpa/astro/io/wcs.py
DougBurke:
```suggestion
        """Convert the output coordinates to the input system.
```

DougBurke:
Ta fo rth ecomments @hamogu and @wmclaugh

PR #2464, opened 2026-05-28
CI: drop manual download of macOS SDK

DougBurke:
# Summary

Simplify the macOS testing on GitHub. There is no functional change.

# Details

The GitHub runners include a numnber of macOS SDKs, so do we need to download a much-older version for the test builds?

This seems to work. I have not made changes to `ci-conda-deployment.yml` as it's less obvious to me what this workflow is doing.

The macOS GitHub runners do include a number of macOS SDKs - e.g. for [macos-15](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#user-content-installed-sdks) - and we could set the `CONDA*blah*` environment variable to point to one of them, but at least for the test/development workflow I'm not sure it's worth it (at least whilst it doesn't seem to matter). I also note that the SDK previously used was much older than those provided in these runners.

### DIFF
--- a/.github/scripts/setup_conda.sh
+++ b/.github/scripts/setup_conda.sh
@@ -13,14 +13,6 @@ if [ "`uname -s`" == "Darwin" ] ; then
     fi
     compilers="clang_osx-${sys} clangxx_osx-${sys} gfortran_osx-${sys}"
 
-    #Download the macOS 11.0 SDK to the CONDA_BUILD_SYSROOT location for the Conda Compilers to work
-    mkdir -p ${GITHUB_WORKSPACE}/../11.0SDK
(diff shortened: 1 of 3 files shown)


DougBurke:
@harlantc - as part of the "clean up" of the code,could you look at this? I have never understood what this is needed for so, and so have no idea if anything needs to be done with the deployment workflow.

harlantc:
For Conda packaging, they still advise building against macOS SDK v11 for backward compatibility reasons. I'll look and see if they're still doing that.

DougBurke:
Thanks - it's just hard to know what SDK is actually being used here. I guess I should also find out how the `workflow` and `deployment` runs interact .....

DougBurke:
So, it's not obvious that this SDK is being picked up, but it's hard to know for sure.

DougBurke:
I'm not sure we gain any benefit from the download but this is not a high priority item, and as it has bit-rotted I'm going to close it.

PR #2465, opened 2026-06-01
Default to the local XPA communicaton protocol when talking to DS9

DougBurke:
# Summary

Communication with DS9 via xpaget and xpaset now defaults to XPA_METHOD=local unless the XPA_METHOD environment variable is set (this is similar to how CIAO calls DS9). This fixes DS9 communication on macOS, allowing for the CI tests to include DS9 on these systems.

# Details

This requires #2462 (well, technically it doesn't but I don't want to worry about merge conflicts, and #2462 has now been merged anyway). It was originally written to go after #2458 and #2459 but has since been re-written.

On macOS the DS9 tests fail with the default XPA_METHOD setting because calls like the following, which Sherpa uses to check if a DS9 session is already running, fail / timeout (note this is purely ds9 behavior and nothing to do with Python or Sherpa):

```
% ds9 -title sherpa &
% sleep 10. $ just to make sure ds9 has started
% xpaget sherpa mode      
XPA$ERROR no 'xpaget' access points match template: sherpa
% echo $?
1
``` 

However, if we switch to `XPA_METHOD=local` then it works:

```
% ds9 -xpa local -title sherpa &
% sleep 10
% XPA_METHOD=local xpaget sherpa mode
none
```

It turns out that this is what CIAO does (it internally sets XPA_METHOD to "local") for **both** linux and macOS, so it should be safe to adopt this approach here too.

There are several ways we could do this, but I've set it up so that whenever an XPA call is made then `-m local` is added **UNLESS** the XPA_METHOD environment variable is set (at least when calling via the `DS9Win` class). This allows the user to change the behavior by manually changing `os.environ["XPA_METHOD"]` if it turns out to be useful. I am not convinced it will be particularly useful, given the way Sherpa loads this (that is, the first time anything from `sherpa.image` is imported it will try to start DS9 and so should need the correct environment), but it is possible to re-import modules.

DS9 is started with the `-xpa local`flag unless the XPA_METHOD environment variable is set.

### DIFF
--- a/.github/workflows/ci-conda-workflow.yml
+++ b/.github/workflows/ci-conda-workflow.yml
@@ -41,6 +41,7 @@ jobs:
             matplotlib-version: 3
             bokeh-version: 3
             xspec-version: 12.14.0i
+            skip-ds9: "true"
 
           - name: MacOS ARM Full Build (Python 3.12)
             os: macos-14
@@ -140,11 +141,12 @@ jobs:
     # been tied to XSPEC, which is why this rule checks for XSPEC
     # support.
     #
-    # The DS9 tests have been shown to be problematic on macOS/GitHub
-    # so we do not install DS9 on this platform.
+    # The DS9 tests are not run on macOS intel as that platform is
+    # slow and depracated. It is easiest to do this with an explicit
+    # check of a matrix field.
     #
     - name: Conda Setup (DS9)
-      if: matrix.xspec-version != '' && runner.os != 'macOS'
+      if: matrix.xspec-version != '' && matrix.skip-ds9 != 'true'
       run: |
         source ${conda_loc}/etc/profile.d/conda.sh
         conda activate build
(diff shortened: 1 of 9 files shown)


DougBurke:
Rebased to address a simple merge conflict from #2484

DougBurke:
@hamogu - this is the next of my "cleanup the ds9" code steps - after this it's 

- #2458 
- #2459

which are mainly code clean up, whereas this one is technically a bug fix (as it allows ds9 testing on macOS CI runs) with minimal code changes. Are you able to review it?

hamogu (approved):

.github/workflows/ci-conda-workflow.yml
```diff
@@ -140,11 +141,12 @@ jobs:
     # been tied to XSPEC, which is why this rule checks for XSPEC
     # support.
     #
-    # The DS9 tests have been shown to be problematic on macOS/GitHub
-    # so we do not install DS9 on this platform.
+    # The DS9 tests are not run on macOS intel as that platform is
+    # slow and depracated. It is easiest to do this with an explicit
```
wmclaugh:
```suggestion
    # slow and deprecated. It is easiest to do this with an explicit
```

PR #2467, opened 2026-06-03
CI: install conda-build into base env and update conda recipe for python 3.13

### LINKED ISSUE #2466: CI builds failing
Somehow merging https://github.com/sherpa/sherpa/pull/2356 has broken the CI builds (at least on conda).

harlantc:
# summary
Fix `conda-build` setup and build for python 3.13

# details
Fix the conda build CI jobs by installing `conda-build` into the base conda env. Also drop activating the `builder` env before runnning `conda-build`

Also fix conda build for python 3.13 by adding pip and numpy 2 (for python 3.13) to conda builds.

fixes #2466

### DIFF
--- a/.github/workflows/ci-conda-deployment.yml
+++ b/.github/workflows/ci-conda-deployment.yml
@@ -86,7 +86,7 @@ jobs:
         curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh -o conda-installer.sh
         bash conda-installer.sh -b -p ${conda_loc}
         source ${conda_loc}/etc/profile.d/conda.sh
-        conda install -yq "conda-build==25.1.2=*_0"
+        conda install -yq conda-build
 
     #Checkout the code with a depth of 0 to grab the tags/branches as well
     - name: Checkout Code
--- a/.github/workflows/ci-conda-deployment.yml
+++ b/.github/workflows/ci-conda-deployment.yml
@@ -1,6 +1,6 @@
 name: Conda Deployments
 
-#Only run when updating main, ciaox, or release/... branches (no PRs).
+#Only run when updating main, ciaox, or release/... branches (no PRs), or on "package test" labels
 #With check-skip job, we skip build on main/ciaox if directly tagged as a release.
 # (To only build releases on the release branch)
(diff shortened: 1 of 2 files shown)


harlantc:
Hmm, I had assumed Sherpa had a mechanism to perform Conda builds in PRs (without uploading the artifacts). I see we don't. In order to test this, I'll see if I can add a mechanism for it (maybe a PR title starting with `PKG:`).

hamogu:
We do have the label "CI arch" that triggers extra jobs. If you think we need it, it could be a label.

DougBurke (approved):
Not that I think I need to approve this, but more to say ta.

harlantc:
Are we OK with me just adding these jobs to the `CI arch` label, or would it be better to put them under a different label?

DougBurke:
I think we probably want a separate "deploy" label rather than overload "CI: arch", as they are different concepts.

harlantc:
@DougBurke could you take another look? I had to update the recipe because numpy 1 doesn't work with python >=3.13. Also made a few more fixes to the Conda deployment CI.

Also, decided to use the label "Packaging" because that's what we're trying to test in that pipeline, but not married to it.

DougBurke:
The `name` field is creating "fun" output (not really a problem, but I thought I'd mention it).

Screenshot 2026-06-03 at 4 51 04 PM

I also note there are a number of warnings about node 20. Do you need to update the "upload artifacts" action version?

The macOS Intel 3.13 run is claiming 

```
Requested labels: macos-13
Job defined at: sherpa/sherpa/.github/workflows/ci-conda-deployment.yml@refs/pull/2467/merge
Waiting for a runner to pick up this job...
Evaluating build-macos.if
Evaluating: (success() && (((github.repository == 'sherpa/sherpa') && (needs.check-skip.outputs.skip_pipeline == 'false'))))
Expanded: (true && (('sherpa/sherpa' == 'sherpa/sherpa') && ('false' == 'false')))
Result: true
```

and not running. Is this an issue?

DougBurke (approved):
Ta

harlantc:
Updated the upload-artifacts due to that warning. Also did fix that image issue you noted.

if that all passes, I'll merge

PR #2469, opened 2026-06-04
utils: internal switch to Path API

DougBurke:
# Summary

The get_datadir/set_datadir routines now internally use the Path API but still accept and return strings. The data directory is now normalized when set.

# Details

As part of this the path is normalized, to remove "un-needed" segments. This should not change the result but may be noticeable in certain cases (e.g "/foo/bar/sherpa/../xxx" becomes "/foo/bar/xxx").

The API could have been changed to use the Path API - e.g. take `Path` objects, or `Path | str` - but it does not seem worth it here.

As part of this clean the module up to remove an unused import (it was need for the `requires_pylab` decorator but that was removed a while ago).

This is not a big change in and of itself but I do plan on using it in  #2447 but it is small enough and completely separate from that code to be reviewed separately.

# Example

Instead of

```
>>> from sherpa.utils.testing import get_datadir
>>> get_datadir()
'/Users/burked/code/sherpa/sherpa-1/sherpa/../sherpa-test-data/sherpatest'
```

we get

```
>>> from sherpa.utils.testing import get_datadir
>>> get_datadir()
'/Users/burked/code/sherpa/sherpa-1/sherpa-test-data/sherpatest'
```

### DIFF
--- a/sherpa/utils/testing.py
+++ b/sherpa/utils/testing.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2017, 2020 - 2024
+#  Copyright (C) 2017, 2020-2024, 2026
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -19,9 +19,7 @@
 #
 
 import importlib
-import os
-
-from sherpa.plot import TemporaryPlottingBackend
+from pathlib import Path
 
 try:
     import pytest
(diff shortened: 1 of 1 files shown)


hamogu (approved):

DougBurke:
Ta for the review @hamogu

PR #2472, opened 2026-06-13
Fix matplotlib 3.11 subplots compatibility

karlhillx:
Closes #2471

matplotlib 3.11 added _raise_if_figure_exists() which throws when
plt.subplots(num=1) is called on an existing figure. The fix is
adding clear=True to destroy and recreate the figure.

There is only one plt.subplots(num=...) call in the codebase
(sherpa/plot/pylab_backend.py:714). The other subplot calls use
plt.subplot (singular) which is unaffected.

PR #2473, opened 2026-06-15
DOCS: point out we can also use bokeh for visualization

### LINKED ISSUE #2479: Read The Docs currently fails
It looks like the RTD environment has upgraded to gcc 15 and so the code fails to build because of #2223

I have a simple work-around for this in #2473 (ensure an older gcc is used) which will let us continue for a while.

DougBurke:
# Summary

Update the Read-The-Docs installation page to mention bokeh as well as matplotlib.

# Details

As well as pointing out bokeh remove the "not-well tested" message for matplotlib as that statement has been outdated for many years. Although it's amusing that I wrote this and then matplotlib 3.11 came along and broke things ...

This was found working on #2432 but is completely separate from that work.

There is also a change needed to allow the RTD code to build (by forcing a gcc version older than 15.0 - as noted in #2223). There are alternative fixes that could be made, but this is by far the simplest.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -6,6 +6,7 @@ channels:
   - nodefaults
 
 dependencies:
+  - gxx=14.2
   - bison
   - python=3.13
   - numpy
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -42,16 +42,13 @@ if installed:
 
 * :term:`Astropy`: for reading and writing files in
   :term:`FITS` format.
-* :term:`matplotlib`: for visualisation of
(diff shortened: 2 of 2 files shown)


DougBurke:
This is obviously not going to pass the tests until #2472 is installed (so not worth looking at until that is done), but this is a documentation-only build.

DougBurke:
@hamogu - this should be safe to review now.

hamogu (approved):

DougBurke:
Ta for the review @hamogu

PR #2474, opened 2026-06-15
Fix links in sphinx documentation

hamogu:
## Summary
Fix inconsistencies leading to briken markup in docstrings. Switch docstring provessing to numpydoc.

## Details
`numpydoc` is more strict that `napoleon`, so I'm using that here to root out inconsistencies in the documentation.

Technically, this could be done without switching to numpydoc, and the output is ver similar, however, we used to use numpydoc for a long time before we switched to napelon to avoid a bunch of warning, and numpydoc is mroe used in the ecosystem (e.g. numpy, scipy, astropy etc. all use numpydoc).

## Open questions
- in some moduels we give plaintext links ot references, in others we mark that up with the RST `name <URL>_` syntax. The latter form is easier to read in rendered docs and doesn't harm much in plain-text docstrings, so I have a slight preference. I don't think it's agood use of our time to go thourgh every single docstringin Sherpa right now, but I could at least turn the ones that I'm editing anyway into links.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -126,8 +126,8 @@ extensions = [
     # Use napoleon over numpydoc for now since it stops a large number
     # of warning messages (about missing links) that I don't have time
     # to investigate.
-    'sphinx.ext.napoleon',
-    # 'numpydoc.numpydoc',
+    # 'sphinx.ext.napoleon',
+    'numpydoc.numpydoc',
     'sphinx.ext.intersphinx',
     'sphinx_astropy.ext.intersphinx_toggle',
     'sphinx_astropy.ext.edit_on_github',
@@ -256,6 +256,7 @@ nbsphinx_prolog = r"""
 napoleon_google_docstring = False
 
 autosummary_generate = True
+numpydoc_class_members_toctree = False
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,10 +123,6 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinx.ext.todo',
     'matplotlib.sphinxext.plot_directive',
-    # Use napoleon over numpydoc for now since it stops a large number
-    # of warning messages (about missing links) that I don't have time
-    # to investigate.
-    # 'sphinx.ext.napoleon',
     'numpydoc.numpydoc',
     'sphinx.ext.intersphinx',
     'sphinx_astropy.ext.intersphinx_toggle',
--- a/sherpa/astro/ui/utils.py
+++ b/sherpa/astro/ui/utils.py
@@ -4372,7 +4372,7 @@ class Session(sherpa.ui.utils.Session):
         References
         ----------
 
-        `K. A. Arnaud, I. M. George & A. F. Tennant, "The OGIP Spectral File Format" <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/ogip_92_007.html>`_
+        .. [1] `K. A. Arnaud, I. M. George & A. F. Tennant, "The OGIP Spectral File Format" <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/ogip_92_007.html>`_
 
         Examples
         --------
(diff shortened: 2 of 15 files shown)


hamogu:
Note that the sphinx build as submitted fails with:
```
/home/docs/checkouts/readthedocs.org/user_builds/sherpa/conda/2474/lib/python3.13/site-packages/numpydoc/docscrape.py:203: UserWarning: potentially wrong underline length... 
[186](https://app.readthedocs.org/projects/sherpa/builds/33149221/#325296041--186)	r(x0,x1)^2 = xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2 
[187](https://app.readthedocs.org/projects/sherpa/builds/33149221/#325296041--187)	------------------------------------------- in 
[188](https://app.readthedocs.org/projects/sherpa/builds/33149221/#325296041--188)	Two-dimensional Hubble-Reynolds model.
```

This needs either #2475, or just take the fixes, but stick with Napoleon, or turn off "warnings as errors".

DougBurke:
I have a commit that "fixes" the warnings/errors from the `sherpa.astro.models` classes so the docs should build, if you want me to push it to this PR. It just reworks the ascii formula into something that `numpydoc` doesn't complain about (and this can then be "improved upon" (or "turn into latex art") in #2475.

hamogu:
Sure, that sounds like a good way forward.

DougBurke:
Ahh, this must have been created before #2473 was merged, which is why the docs build fails.

hamogu:
Yeah, a lot of my stuff is s couple weeks old. Rebased!

sherpa/astro/ui/utils.py
```diff
@@ -4372,7 +4372,7 @@ def get_backscal(self,
         References
         ----------
 
-        `K. A. Arnaud, I. M. George & A. F. Tennant, "The OGIP Spectral File Format" <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/ogip_92_007.html>`_
+        .. [1] `K. A. Arnaud, I. M. George & A. F. Tennant, "The OGIP Spectral File Format" <https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/ogip_92_007.html>`_
```
DougBurke: So - and this holds for a bunch of these - we don't actually refer to this reference, it's more like a general reference (so, at some level, I could see the system complain that there's no corresponding `[1]_`  element). So the things I could suggest are

 1, add such a link somewhere
 2. just drop the reference section from these routines as the context is somewhat self evident

I suggest option 2 but maybe some of them should have option 1, it's a bit hard to see which routines are affected.

Option 1: get_backscal, get_areascal, group, get_grouping, get_quality, ungroup

Option 2: set_grouping, set_quality

For option 2 there's a `Notes` section with a line like `The meaning of the quality column is taken from the OGIP standard, ...` and this can be the place for the `[1]_` (sorry, the GH web page makes it hard to do and I don't have access to this branch at the moment).

DougBurke: I can push a fix for this if that would be useful (after it builds and I check the output okay..)?

hamogu: I can make a suggestion for that, too, probably early next week. I'll let you work on your PRs @DougBurke instead of cleaning up mine. Ultimately, this is about making things more consistent and fixing missing links (which can be annoying when you stumble on them as a super and you don't really know where that was supposed to go). 

I'll have a look. It might well be there all those cases have natural point to put the reference. But if not, I'm also OK with just sticking with napoleon (which doesn't complain about inconsistencies like this) and just fixing what really should be fixed. I don't want this to degrade into busywork of just changing things to make some check happy with no benefit for users; we have better things to spent time on. But I do want to get at least part of this in to fix the things that are potentially confusing (e.g a link with additional information that doesn't work).
I'll take a closer look!

DougBurke: I have two commits I can add here: one to address this particular one about references and one to address more examples of "numpydoc and '-------' get into a fight"

DougBurke: But I can also just not push it here. I won't create a new PR as that would just lead to conflicts.

hamogu: If you have it already, please push. We don't want to do double work! I just didn't want to push (harharhar!) to do it, but if you've already done it, then go for it!

DougBurke:
@hamogu - do you think this is good to go or are there more cases you want to change?

sherpa/models/basic.py
```diff
@@ -1506,17 +1508,17 @@ class Gauss2D(RegriddableModel2D):
 
         f(x0,x1) = ampl * exp(-4 * log(2) * r(x0,x1)^2)
 
-        r(x0,x1)^2 = xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2
-                     -------------------------------------------
-                                fwhm^2 * (1-ellip)^2
+        r(x0,x1)^2 = num(x0,x1) / (fwhm^2 * (1-ellip)^2)
```
hamogu, line 1511: I like this. It's actually easier to read IMO. ?Why didn't I think ion suggesting that?

hamogu:
I think this is good to go, but as the original PR author, I can't "approve" the PR, so @DougBurke needs to do that or @wmclaugh can just merge it.

docs/conf.py
```diff
@@ -126,8 +126,8 @@ def get_xsversion(self):
     # Use napoleon over numpydoc for now since it stops a large number
```
DougBurke: We should fix this comment as it's no-longer valid

DougBurke (approved):
Ta - D.

PR #2475, opened 2026-06-15
Turn equations into a math environment in docstrings.

hamogu:
This is something that I had wanted to explore of a long time, so I just made a PR for a few example so that we can look at them and discuss.

There is a trade-off: I think the rendered math looks much better in the html rendered pages, but of course it's a little harder to read in a plain docstring (e.g. `help(Modelname)` in Python. 
I might be biased, but I'm using the html more and the userbase of Sherpa will also read LaTeX natively. 

Are those equations exposed to the ahelp pages whifh are again rendered in a different way? I don't think so, but maybe I didn't look long enough.

### DIFF
--- a/sherpa/astro/models/__init__.py
+++ b/sherpa/astro/models/__init__.py
@@ -564,7 +564,7 @@ class LineBroad(RegriddableModel1D):
 #           section confuses sphinx (it thinks it is a section title).
 #
 class Lorentz1D(RegriddableModel1D):
-    """One-dimensional normalized Lorentz model function.
+    r"""One-dimensional normalized Lorentz model function.
 
     Attributes
     ----------
@@ -581,11 +581,11 @@ class Lorentz1D(RegriddableModel1D):
 
     Notes
     -----
-    The functional form of the model for points is::
+    The functional form of the model for points is:
 
-        f(x) =              ampl * fwhm
-               --------------------------------------
-               2 * pi * (0.25 * fwhm^2 + (x - pos)^2)
+    .. math::
+
+       f(x) = \frac{ampl * fwhm}{2 * \pi * (0.25 * fwhm^2 + (x - pos)^2)}
 
     and for an integrated grid it is the integral of this over
     the bin.
@@ -957,7 +957,7 @@ class Schechter(RegriddableModel1D):
 
 
 class Beta2D(RegriddableModel2D):
-    """Two-dimensional beta model function.
+    r"""Two-dimensional beta model function.
 
     The beta model is a Lorentz model with a varying power law.
 
@@ -985,17 +985,17 @@ class Beta2D(RegriddableModel2D):
 
     Notes
     -----
-    The functional form of the model for points is::
+    The functional form of the model for points is:
 
-        f(x0,x1) = ampl * (1 + r(x0,x1)^2)^(-alpha)
+    .. math::
 
-        r(x0,x1)^2 = xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2
-                     -------------------------------------------
-                                  r0^2 * (1-ellip)^2
+        f(x_0,x_1) &= ampl * (1 + r(x_0,x_1)^2)^{-\alpha}
 
-        xoff(x0,x1) = (x0 - xpos) * cos(theta) + (x1 - ypos) * sin(theta)
+        r(x_0,x_1)^2 &= \frac{xoff(x_0,x_1)^2 * (1-ellip)^2 + yoff(x_0,x_1)^2}{r_0^2 * (1-ellip)^2}
 
-        yoff(x0,x1) = (x1 - ypos) * cos(theta) - (x0 - xpos) * sin(theta)
+        xoff(x_0,x_1) &= (x_0 - xpos) * \cos(\theta) + (x_1 - ypos) * \sin(\theta)
+
+        yoff(x_0,x_1) &= (x_1 - ypos) * \cos(\theta) - (x_0 - xpos) * \sin(\theta)
 
     The grid version is evaluated by adaptive multidimensional
     integration scheme on hypercubes using cubature rules, based
@@ -1133,7 +1133,7 @@ class DeVaucouleurs2D(RegriddableModel2D):
 
 
 class HubbleReynolds(RegriddableModel2D):
-    """Two-dimensional Hubble-Reynolds model.
+    r"""Two-dimensional Hubble-Reynolds model.
 
     Attributes
     ----------
@@ -1159,13 +1159,13 @@ class HubbleReynolds(RegriddableModel2D):
     -----
     The functional form of the model for points is::
 
-        f(x0,x1) = ampl / (1 + r(x0,x1))^2
+    .. math::
 
-        r(x0,x1)^2 = xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2
-                     -------------------------------------------
-                                  r0^2 * (1-ellip)^2
+       f(x0,x1) = ampl / (1 + r(x0,x1))^2
 
-        xoff(x0,x1) = (x0 - xpos) * cos(theta) + (x1 - ypos) * sin(theta)
+       r(x0,x1)^2 = \frac{xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2}{r0^2 * (1-ellip)^2}
+
+       xoff(x0,x1) = (x0 - xpos) * cos(theta) + (x1 - ypos) * sin(theta)
 
         yoff(x0,x1) = (x1 - ypos) * cos(theta) - (x0 - xpos) * sin(theta)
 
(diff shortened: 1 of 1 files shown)


hamogu (commented):
We dicussed this concept about a decade ago in https://github.com/sherpa/sherpa/pull/337#issuecomment-306363028 where the turned some statistics into math notation.

For now, I've edited jsut a few docstrings (about half that have significant equations) as a proof-of-concept but I won't take long to do the rest, too:

- `sherpa.astro.models.Lorentz1D` is an easy example where the latex is actually shorter than the plain text version.
- `sherpa.astro.models.HubbleReynolds` is longer and the equations look ragged because I didn't alin them in LaTeX. Paraemters veluws are spelled out as strings, which looks odd in LaTeX but introduces less clutter in plain text compared to the alternative `\mathrm{ampl}.
- `sherpa.astro.models.Beta2D` uses an eqnarray as syntax which looks better in html, but adds noise in plain text. Here, I've alos turned parameter names into obvious  latex equivalents (e.g. `x0` into `$x_0` and `theta` into `$\theta$). I think it reads better in the equation and it's obvious how that maps to the parameter.

DougBurke:
Not looked at this yet, but the sherpa ahelp files basically come from the methods of the `Session` object (so from `sherpa.ui.utils.Session` and `sherpa.astro.ui.utils.Session`).

sherpa/astro/models/__init__.py
```diff
@@ -1439,15 +1439,15 @@ class Sersic2D(RegriddableModel2D):
     Notes
     -----
     The functional form of the model for points is can be
-    expressed as the following::
+    expressed as the following:
+
+    ..math::
```
DougBurke:
```suggestion
    .. math::
```

sherpa/astro/models/__init__.py
```diff
@@ -1159,13 +1159,13 @@ class HubbleReynolds(RegriddableModel2D):
     -----
     The functional form of the model for points is::
```
DougBurke:
```suggestion
    The functional form of the model for points is:
```

sherpa/astro/models/__init__.py
```diff
@@ -1159,13 +1159,13 @@ class HubbleReynolds(RegriddableModel2D):
...
 
-        f(x0,x1) = ampl / (1 + r(x0,x1))^2
+    .. math::
 
-        r(x0,x1)^2 = xoff(x0,x1)^2 * (1-ellip)^2 + yoff(x0,x1)^2
-                     -------------------------------------------
-                                  r0^2 * (1-ellip)^2
+       f(x0,x1) = ampl / (1 + r(x0,x1))^2
```
DougBurke:
```suggestion
       f(x0,x1) &= ampl / (1 + r(x0,x1))^2
```

(and for the other lines, which I'm too lazy to suggest), as you did this for `Beta2D` (although it's not a requirement).

DougBurke:
So, it looks lie I can change our ahelp code to support this, so that `ahelp beta2d` would look like (for an 80-column display):

```
    | alpha     | The power-law slope of the profile at large radii.       |
    ------------------------------------------------------------------------

2. Notes

    The functional form of the model for points is:

               f(x_0,x_1) &= ampl * (1 + r(x_0,x_1)^2)^{-\alpha}

    r(x_0,x_1)^2 &= \frac{xoff(x_0,x_1)^2 * (1-ellip)^2 + 
    yoff(x_0,x_1)^2}{r_0^2 * (1-ellip)^2} 

    xoff(x_0,x_1) &= (x_0 - xpos) * \cos(\theta) + (x_1 - ypos) * 
    \sin(\theta) 

    yoff(x_0,x_1) &= (x_1 - ypos) * \cos(\theta) - (x_0 - xpos) * 
    \sin(\theta) 

    The grid version is evaluated by adaptive multidimensional integration 
    scheme on hypercubes using cubature rules, based on code from HIntLib ( 
    [1] ) and GSL ( [2] ).

3. References

     [1] HIntLib - High-dimensional Integration Library 
    [https://github.com/JohannesBuchner/HIntLib] 

     [2] GSL - GNU Scientific Library [https://www.gnu.org/software/gsl/] 

BUGS
```

(I could even replace '&=' with '='). I haven't investigated yet how to take advantage of our mathjax support for the web version - e.g. the CIAO 4.19 version of https://cxc.cfa.harvard.edu/sherpa4.18/ahelp/beta2d.html#Notes but at worst it would just contain the latex-like text.

So, the important things appear to be

1. make sure the section is labelled with a single and not double colon (as most of these changes were)
2. ensure '.. math::'
3. make sure all the statements in the math block all start in the same column (for some reason we have some blocks where one of the lines has an extra space and this seems to cause internal issues to how the text is represented which I'd rather not have to worry about).

The only downside, other than the fact that it's work and conflicts with some of the changes we needed to do for the recent numpydoc rework, is that ideally we'd label all the terms with something like `\rm{term}` to improve the display, but then this is just a lot of extra characters that a reader has to parse when looking at the text, so I am okay with the current approach - e.g.

image

So if you want to work on this for 4.19 I'm happy to review.

hamogu (commented):
There is a trade-off between proper LaTeX form and plain-text readability. One of the most obvious examples is an equations like `(\frac{a}{b})^2`. The correct way to express that in LaTeX would be
`\left(\frac{a}{b}\right)^2`. However, that is significantly longer (by number of characters) and harder to read in plain text. So, I didn't use this and decided to live with () that don't quite have the correct size in LaTeX.

This commit also makes use of symbols in moderation. I;ve replaced "ellip" with "\epsilon" which is almost universally used for "ellipticity". On the other hand, there is no commonly accepted symbol for FWHM so I left that as FWHM which will render a bit off in LaTeX but is readable. 

I've replaced other parameter names with symbols where it's obvious, e.g. x0 with $x_0$, but left others spelled out (e.g. "xpos"). 
There is a sliding scale between 
full LaTeX <-> full ASCII
and I'm trying to find the best middle ground. None of this is set in stone; future commits and build on this or revert some changes. I recognize that there is no "best" solution, but only personal preference of how far to go.

For simular reasons I use table formatting (`&=&` to line up the equal signs) in very rare cases. Ther eare more models that would profit from that in rendered formulas, but in most cases it's OK without and better to read in plan text.

DougBurke:
image

very fancy ;-)

sherpa/astro/models/__init__.py
```diff
@@ -396,27 +404,32 @@ class Dered(RegriddableModel1D):
...
 
-        E(B-V) = nhgal / 58.0
+    where the units of nhgal is :math:`10^{20} \mathrm{cm}^{-2}`. The value of the ratio
+    of total to selective extinction, :math:`R_V`, is initially set to 3.1, the
+    standard value for the diffuse ISM. The final model form is:
 
-    where the units of nhgal is 10^20 cm^-2. The value of the ratio
-    of total to selective extinction, rv, is initially set to 3.1, the
-    standard value for the diffuse ISM. The final model form is::
+    .. math::`
```
DougBurke:
```suggestion
    .. math::
```

This was causing the expression to not be rendered in the HTML and problems for my docstring-to-ahelp conversion code.

sherpa/astro/models/__init__.py
```diff
@@ -330,21 +336,23 @@ class BPL1D(RegriddableModel1D):
...
     Notes
     -----
-    The functional form of the model for points is::
+    The functional form of the model for points is:
 
-        f(x) = ampl * (x / ref)^(-gamma1)   x <= eb
+    .. math::
 
-             = ampl' * (x / ref)^(-gamma2)  otherwise
+        f(x) = A * (\frac{x}{ref})^{-\gamma_1}   \,\mathrm{ for} \, x <= eb
```
DougBurke:
```suggestion
        f(x) = A * (\frac{x}{ref})^{-\gamma_1}   \,\mathrm{for} \, x <= eb
```

sherpa/astro/models/__init__.py
```diff
@@ -330,21 +336,23 @@ class BPL1D(RegriddableModel1D):
...
 
-       ampl' = ampl (eb / ref)^(gamma2 - gamma1)
+             = A' * (\frac{x}{ref})^{-\gamma_2}  \,\mathrm{ otherwise }
```
DougBurke:
```suggestion
             = A' * (\frac{x}{ref})^{-\gamma_2}  \,\mathrm{otherwise}
```

DougBurke: [it makes the special-casing I need for the ahelp generation a bit cleaner, but this and the one in the preceeding line are not hugely important to do]

sherpa/astro/models/__init__.py
hamogu: Typo, sorry. "`" is right next to TAB on the keyboard. TAB accepts the copilot suggestions which are pretty good for mostly technical editing like this one (e.g. the location where I add `.. math::` is pretty obvious, so when I accept the suggestion for auto-complete, I must have hit both keys with my fat fingers).

DougBurke (approved):
As @hamogu mentioned, it's always going to be a choice of exactly how much mark-up to use for these "multi modal" (e.g. plain text and HTML rendered) cases. This is defintelly an improvement and builds on @hamogu's earlier work in #2474

Issue #2471, opened 2026-06-12
matplotlib 3.11.0 breaks our tests


DougBurke:
matplotlib 3.11.0 was released yesterday and has made its way to the pip tests at least. They then break.

I had matplotlib 3.10.9 installed and 

```
pytest sherpa/astro/ui/tests/test_astro_session.py
```

worked. After

```
pip install 'matplotlib==3.11.0'
```

we now get 2 failures

```
============================================================================== FAILURES ===============================================================================
______________________________________________________________ test_plot_fit_resid_get_labels[Session0] _______________________________________________________________

session = <class 'sherpa.ui.utils.Session'>

    @pytest.mark.parametrize("session", [Session, AstroSession])
    def test_plot_fit_resid_get_labels(session):
        """Regression test for plot object handling with plot_fit_xxx
    
        This should check all the residual-style plots, but assume if it
        holds for plot_fit_resid then it will for the others too.
    
        """
    
        s = session()
        s._add_model_types(sherpa.models.basic)
    
        s.load_arrays(1, [1, 2, 3], [4, 5, 7])
        mdl = s.create_model_component("scale1d", "mdl")
        mdl.c0 = 6
        s.set_source(mdl)
    
>       s.plot_fit_resid()

sherpa/astro/ui/tests/test_astro_session.py:4647: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sherpa/ui/utils.py:16163: in plot_fit_resid
    self._jointplot2(plot1obj, plot2obj, overplot=overplot,
sherpa/ui/utils.py:16028: in _jointplot2
    self._jointplot.plottop(plot1, overplot=overplot,
sherpa/plot/__init__.py:1363: in plottop
    backend.set_jointplot(0, 0, self.rows, self.cols,
sherpa/plot/pylab_backend.py:714: in set_jointplot
    fig, axes = plt.subplots(nrows, ncols, sharex=True, num=1,
../../../miniforge3/envs/sherpa-1/lib/python3.14/site-packages/matplotlib/pyplot.py:1884: in subplots
[...]

Issue #1384, opened 2021-12-23
model objects passed into `datastack.set_source` must have a "." in their name


hamogu:
As an example
```
from sherpa.astro import datastack as ds
from sherpa.astro import ui
from sherpa.models.basic import Const1D

# In real live we would make the datastack from files, but as a minimal example
# let's just load some arrays
ui.load_arrays('data1', [1,2,3,4], [1.1, 0.9, 1.1, 1.])
ui.load_arrays('data2', [1,2,3,4], [1.2, 0.9, 1.1, 1.])
stack = ds.DataStack()
stack._add_dataset('data1')
stack._add_dataset('data2')
```
The `stack.set_source` command is very flexible. Just like the usual `ui.set_source` is takes a string, a model expression, or the instance of a model, but with the added twist that model instances must have a "." in their name:
```
stack.set_source('const1d.con1')
stack.set_source(ui.const1d.con1)
stack.set_source(Const1D(name='name.name'))
```
However, a model instance without a "." failes:
```
stack.set_source(Const1D(name='name_name'))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/mambaforge/envs/i386ciao-4.14/lib/python3.9/site-packages/sherpa/astro/datastack/utils.py in _create_stack_model(model, model_id, model_components)
    128             try:
--> 129                 model_type, model_name_ID = model.name.split('.')
    130             except ValueError:

ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/var/folders/r7/f0qh27rn207bwwvt3s7v5gh40000gn/T/ipykernel_29896/694500353.py in <module>
----> 1 stack.set_source(Const1D(name='name_name'))

~/mambaforge/envs/i386ciao-4.14/lib/python3.9/site-packages/sherpa/astro/datastack/utils.py in wrapfunc(self, model)
     53             logger.info('Setting stack model using {0}() for id={1}'.format(
     54                 func.__name__, id_))
---> 55             new_model, model_comps = create_stack_model(model, id_)
     56             func(id_, new_model)
     57             dataset['model_comps'].update(model_comps)

~/mambaforge/envs/i386ciao-4.14/lib/python3.9/site-packages/sherpa/astro/datastack/utils.py in create_stack_model(model, id_)
     88 
     89     """
---> 90     return _create_stack_model(model, id_)
     91 
     92 

[...]

PR #2476, opened 2026-06-17
Ensure the radial setting of PSF models are written out by the save_all command 

### LINKED ISSUE #2470: save_all does not save the radial flag for 1D PSFs
Found when developing #2447 

If you do something like

```python
load_psf("p1", "foo.dat")
set_psf(p1)
p1.radial = 1
```

then the `save_all` output is just

```
######### Set Model Components and Parameters

load_psf("p1", "foo.dat")


######### Associate PSF models with the datasets

set_psf(1, p1)


```

That is, it does not record the change to `radial` (in a real-world case there would be some changes to `p1` but not `radial`).

### LINKED ISSUE #2506: save_all: does not handle load_conv
This is with CIAO 4.18 but the code has been like this for a while;

```
sherpa
Sherpa 4.18.0+2.ga0c5211a

Python 3.12.12 | packaged by conda-forge | (main, Oct 13 2025, 14:47:45) [Clang 19.1.7 ]
IPython 9.13.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: sherpa
Tip: Use `ipython --help-all | less` to view all the IPython configuration options.
Using matplotlib backend: macosx

sherpa In [1]: load_conv("bob", gauss1d.g1)

sherpa In [2]: from io import StringIO

sherpa In [3]: buff = StringIO(); save_all(buff)

sherpa In [4]: clean()

sherpa In [5]: exec(buff.getvalue())
ArgumentErr: 'convolutionkernel' is not a valid model type

sherpa In [6]:
```

This is because the serialization stores something like

```
create_model_component("convolutionkernel", "bob")
load_conv("bob", "gauss1d.g1")
```

where the `create_model_component` line is wrong and can just be dropped.

DougBurke:
# Summary

Ensure that the radial setting of a PSF model and error-estimation options are saved in the save_all output. The method options are now only written out when explicitly set. Fix #2470. Fix #2506.

# Details

This came out of #2447. We have 

- add a test or two to cover some corner cases, including #2470, #2504, #2506
- some minor code clean up
- fix some of the bugs, including #2470 and #2506 (but not #2504)

Unfortunately - for this code at least - the PSF model class 

- doesn't follow other classes and have a `.pars` field (there are reasons for this, it's just why this issue happened)
- has some "parameters" only appear after associated with a dataset

which means we always risk the "missing a field" case (even after this PR; I have thought about trying to make it "generic" so the special-case code would not be needed, but it does not feel like it's worth it at this time given the extra complexity likely required).

A change is made to support properly restoring `load_conv` calls - following how we do it for `load_psf`. This is issue #2506.

Doing a little bit of type checking with `pyright` made me some changes that are either just to make it easier for the type checkers (but probably also make the code a bit cleaner as there is less "boolean blindness") or actually avoid an issue (where the psf or convolution kernel is not set) which is extremely difficult to pull off for users of the UI layer and so no test is added.

# Example

Before this PR, not that the `save_all` output does not mention the `radial` parameter:

```
>>> from io import StringIO
>>> ui.load_psf("bob", ui.gauss1d.g1)
>>> print(bob)
psfmodel.bob
   bob.kernel   frozen   gauss1d.g1
   bob.radial   frozen            0            0            1           
   bob.norm     frozen            1            0            1           
>>> bob.radial = 1
>>> print(bob)
psfmodel.bob
   bob.kernel   frozen   gauss1d.g1
   bob.radial   frozen            1            0            1           
   bob.norm     frozen            1            0            1           
>>> buff = StringIO()
>>> ui.save_all(buff)
>>> print(buff.getvalue())
import numpy
from sherpa.astro.ui import *


######### Set Statistic

set_stat("chi2gehrels")


######### Set Fitting Method

set_method("levmar")

set_method_opt("epsfcn", 1.1920928955078125e-07)
set_method_opt("factor", 100.0)
set_method_opt("ftol", 1.1920928955078125e-07)
set_method_opt("gtol", 1.1920928955078125e-07)
set_method_opt("maxfev", None)
set_method_opt("numcores", 1)
set_method_opt("verbose", 0)
set_method_opt("xtol", 1.1920928955078125e-07)


######### Set Model Components and Parameters

create_model_component("gauss1d", "g1")
g1.integrate = True

g1.fwhm.default_val = 10.0
g1.fwhm.default_min = 1.1754943508222875e-38
g1.fwhm.default_max = 3.4028234663852886e+38
g1.fwhm.val     = 10.0
g1.fwhm.min     = 1.1754943508222875e-38
g1.fwhm.max     = 3.4028234663852886e+38
g1.fwhm.units   = ""
g1.fwhm.frozen  = True

g1.pos.default_val = 0.0
g1.pos.default_min = -3.4028234663852886e+38
g1.pos.default_max = 3.4028234663852886e+38
g1.pos.val     = 0.0
g1.pos.min     = -3.4028234663852886e+38
g1.pos.max     = 3.4028234663852886e+38
g1.pos.units   = ""
g1.pos.frozen  = True

g1.ampl.default_val = 1.0
g1.ampl.default_min = -3.4028234663852886e+38
g1.ampl.default_max = 3.4028234663852886e+38
g1.ampl.val     = 1.0
g1.ampl.min     = -3.4028234663852886e+38
g1.ampl.max     = 3.4028234663852886e+38
g1.ampl.units   = ""
g1.ampl.frozen  = True

load_psf("bob", "gauss1d.g1")


>>> 
```

The new output is

```
>>> >>> print(buff.getvalue())
import numpy
from sherpa.astro.ui import *


######### Set Statistic

set_stat("chi2gehrels")


######### Set Fitting Method

set_method("levmar")

set_method_opt("epsfcn", 1.1920928955078125e-07)
set_method_opt("factor", 100.0)
set_method_opt("ftol", 1.1920928955078125e-07)
set_method_opt("gtol", 1.1920928955078125e-07)
set_method_opt("maxfev", None)
set_method_opt("numcores", 1)
set_method_opt("verbose", 0)
set_method_opt("xtol", 1.1920928955078125e-07)


######### Set Model Components and Parameters

create_model_component("gauss1d", "g1")
g1.integrate = True

g1.fwhm.default_val = 10.0
g1.fwhm.default_min = 1.1754943508222875e-38
g1.fwhm.default_max = 3.4028234663852886e+38
g1.fwhm.val     = 10.0
g1.fwhm.min     = 1.1754943508222875e-38
g1.fwhm.max     = 3.4028234663852886e+38
g1.fwhm.units   = ""
g1.fwhm.frozen  = True

g1.pos.default_val = 0.0
g1.pos.default_min = -3.4028234663852886e+38
g1.pos.default_max = 3.4028234663852886e+38
g1.pos.val     = 0.0
g1.pos.min     = -3.4028234663852886e+38
g1.pos.max     = 3.4028234663852886e+38
g1.pos.units   = ""
g1.pos.frozen  = True

g1.ampl.default_val = 1.0
g1.ampl.default_min = -3.4028234663852886e+38
g1.ampl.default_max = 3.4028234663852886e+38
g1.ampl.val     = 1.0
g1.ampl.min     = -3.4028234663852886e+38
g1.ampl.max     = 3.4028234663852886e+38
g1.ampl.units   = ""
g1.ampl.frozen  = True

load_psf("bob", "gauss1d.g1")
bob.radial = 1.0
bob.norm = 1.0


>>>
```

So, after the `load_psf` call we now have `radial` and `norm` settings.

For the `load_conv` case we have:

```
>>> from io import StringIO
>>> ui.load_conv("c1", ui.gauss1d.g1)
>>> buff = StringIO(); ui.save_all(buff)
>>> ui.clean()
>>> print(c1)
  File "<python-input-5>", line 1, in <module>
    print(c1)
          ^^
NameError: name 'c1' is not defined
```

## before

```
>>> exec(buff.getvalue())
  File "<python-input-6>", line 1, in <module>
    exec(buff.getvalue())
    ~~~~^^^^^^^^^^^^^^^^^
  File "<string>", line 56, in <module>
  File "<string>", line 1, in create_model_component
  File "/home/dburke/sherpa/sherpa-3/sherpa/ui/utils.py", line 6686, in create_model_component
    raise ArgumentErr('badtype', typename)
sherpa.utils.err.ArgumentErr: 'convolutionkernel' is not a valid model type
```

## after

```
>>> exec(buff.getvalue())
>>> print(c1)
Convolution Kernel:
gauss1d.g1
   g1.fwhm      thawed           10  1.17549e-38  3.40282e+38           
   g1.pos       thawed            0 -3.40282e+38  3.40282e+38           
   g1.ampl      thawed            1 -3.40282e+38  3.40282e+38           
```

### DIFF
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -186,7 +186,7 @@ def _save_entries(out: OutType,
     store
        A container with keys. The elements of the container are
        passed to tostatement to create the string that is then
-       written to fh.
+       saved in out.
     tostatement : func
        A function which accepts two arguments, the key and value
        from store, and returns a string. The reason for the name
@@ -374,9 +374,6 @@ def _save_pha_grouping(out: OutType,
     bid
        If not ``None`` then this indicates that the background dataset
        is to be used.
-    fh : None or file-like
-       If ``None``, the information is printed to standard output,
-       otherwise the information is added to the file handle.
     """
 
     _save_pha_array(out, state, pha, "grouping", id, bid=bid)
@@ -1718,10 +1715,10 @@ def save_all(state: SessionType,
 
     _save_id(out, state)
 
-    if fh is None:
-        fh = sys.stdout
+    outfh = sys.stdout if fh is None else fh
 
-    write = lambda msg: fh.write(f"{msg}\n")
+    def write(msg: str) -> None:
+        outfh.write(f"{msg}\n")
 
     # Hard code the required imports.
     #
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -42,8 +42,9 @@ from sherpa.astro import io
 from sherpa.astro.io.wcs import WCS
 
 from sherpa.data import Data, Data1D, Data1DInt, Data2D, Data2DInt
+from sherpa.instrument import ConvolutionKernel, PSFModel
 from sherpa.models import Model
-from sherpa.models.basic import UserModel
+from sherpa.models.basic import TableModel, TableModelBase, UserModel
 from sherpa.models.parameter import Parameter
 from sherpa.utils.types import IdType
 
@@ -1023,21 +1024,21 @@ def _handle_model(out: OutType,
 
     """
 
-    typename = mod.type
     modelname = _get_cpt_name(mod)
     found_xspec = False
-    if typename == "usermodel":
+    if isinstance(mod, UserModel):
         _handle_usermodel(out, mod, modelname)
 
-    elif typename == "psfmodel":
+    elif isinstance(mod, PSFModel):
         cmd = f'load_psf("{mod._name}", "{mod.kernel.name}")'
         _output(out, cmd)
 
-    elif typename in ["fixedtablemodel", "interpolatedtablemodel1d"]:
+    elif isinstance(mod, (TableModel, TableModelBase)):
+        # Use of TableModel is deprecated.
         cmd = f'load_table_model("{modelname}", "{mod.filename}")'
         _output(out, cmd)
 
-    elif typename == "xstablemodel":
+    elif xspec is not None and isinstance(mod, xspec.XSTableModel):
         cmd = f'load_xstable_model("{modelname}", "{mod.filename}"'
         if mod.etable:
             cmd += ', etable=True'
@@ -1048,21 +1049,20 @@ def _handle_model(out: OutType,
 
     else:
         # Normal case:  create an instance of the model.
-        cmd = f'create_model_component("{typename}", "{modelname}")'
+        cmd = f'create_model_component("{mod.type}", "{modelname}")'
         _output(out, cmd)
 
-        # Is this an XSPEC model? We only care if it's the first
-        # one we find.
+        # Is this an XSPEC model?
         #
-        if not found_xspec and xspec is not None:
-            found_xspec = isinstance(mod, xspec.XSModel)
+        if xspec is not None:
+            found_xspec |= isinstance(mod, xspec.XSModel)
 
     # QUS: should this be included in the above checks?
     #      @DougBurke doesn't think so, as the "normal
     #      case" above should probably be run , but there's
     #      no checks to verify this.
     #
-    if typename == "convolutionkernel":
+    if isinstance(mod, ConvolutionKernel):
         # Create general convolution kernel with load_conv
         cmd = f'load_conv("{modelname}", "{mod.kernel.name}")'
         _output(out, cmd)
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -1029,11 +1029,11 @@ def _handle_model(out: OutType,
     if isinstance(mod, UserModel):
         _handle_usermodel(out, mod, modelname)
 
-    elif isinstance(mod, PSFModel):
+    elif isinstance(mod, PSFModel) and mod.kernel is not None:
         cmd = f'load_psf("{mod._name}", "{mod.kernel.name}")'
         _output(out, cmd)
 
-    elif isinstance(mod, ConvolutionKernel):
+    elif isinstance(mod, ConvolutionKernel) and mod.kernel is not None:
         cmd = f'load_conv("{modelname}", "{mod.kernel.name}")'
         _output(out, cmd)
 
@@ -1280,14 +1280,11 @@ def _save_bkg_source(out: OutType,
             except:
                 the_bkg_full_model = None
 
-            have_source = the_bkg_source is not None
-            have_full_model = the_bkg_full_model is not None
-
-            if have_source:
+            if the_bkg_source is not None:
                 # This does not check for the dataset being a DataPHA
                 # object, since (at present) it has to be, as it's the
                 # only one to support backgrounds
-                if have_full_model:
+                if the_bkg_full_model is not None:
                     if repr(the_bkg_source) == repr(the_bkg_full_model):
                         cmd = f"set_bkg_full_model({cmd_id}, {the_bkg_full_model.name}, bkg_id={cmd_bkg_id})"
                     else:
@@ -1295,7 +1292,7 @@ def _save_bkg_source(out: OutType,
                 else:
                     cmd = f"set_bkg_source({cmd_id}, {the_bkg_source.name}, bkg_id={cmd_bkg_id})"
 
-            elif have_full_model:
+            elif the_bkg_full_model is not None:  # have_full_model:
                 cmd = f"set_bkg_full_model({cmd_id}, {the_bkg_full_model.name}, bkg_id={cmd_bkg_id})"
 
             else:
--- a/sherpa/astro/ui/serialize.py
+++ b/sherpa/astro/ui/serialize.py
@@ -801,7 +801,7 @@ def _save_data(out: OutType,
         _handle_filter(out, state, data, idval)
 
 
-def _print_par(par: ParameterType) -> tuple[str, str]:
+def _print_par(par: ParameterType) -> tuple[str, str | None]:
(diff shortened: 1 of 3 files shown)


DougBurke:
Why does conda installing

```
  added / updated specs:
    - astropy
    - gcc_linux-64=14.2
    - gfortran_linux-64
    - gxx_linux-64=14.2
    - numpy=2.3.2
    - pip
    - python=3.13
```

lead to matplotlib also being installed? Looks like it's an astropy dependency, which messes up our CI builds a little bit ....

hamogu:
Matplotlib is an optional astropy dependency (https://github.com/astropy/astropy/blob/main/pyproject.toml#L45), not a required one, but maybe the recipe in conda includes all astropy dependencies that are in the "recommended" group equivalent to `pip install "astropy[recommended]"` (which, as the name implies, is what astropy recommends for all users, so it would make sense to package it in conda like that). We could switch astropy to a pip install instead of conda, then we're guaranteed to get only the strictly required dependencies and since astropy builds wheels it won't take any longer than a conda install.

DougBurke:
Rebased now that #2471 has been fixed. The RTD docs will fail because of #2479

DougBurke:
Rebased as #2473 was merged to fix #2479

DougBurke:
Rebased to remove some clean-up code that is not directly related to the bug fix and may slightly conflict with some other PRs

DougBurke:
Rebased to pick up #2484 and to fix up a few stupid typos in the commit messages. That's not to say there aren't still some hiding away.

sherpa/astro/ui/serialize.py
```diff
@@ -1001,6 +998,125 @@ def _handle_usermodel(out: OutType,
...
+    # If we have a name "aa.bb" then return "bb" otherwise
+    # return the name unchanged.
+    #
+    name = mod.name
+    try:
+        return name.split(".")[1]
+    except IndexError:
+        return name
+
+
```
hamogu, line 1013: Wouldn't it be easier to make this a one-liner and inline it where needed?
```
variable_I_want_to_set = mod.names.split(".")[-1]
```
I always think that the mental overhead for a reader to go and look at what exactly a private function does is higher than reading the one line of standard Python.

hamogu: I also note that heuristics like this can be dangerous, see #1384. Here, the failure mode is the opposite: What if my model is called "xsapec.src1.cold"?
I don't see a way around corner cases like that when you need to rely on heuristics, and we can probably live with it, but it's certainly not ideal.

sherpa/astro/ui/serialize.py
```diff
@@ -1001,6 +998,125 @@ def _handle_usermodel(out: OutType,
...
+        _output(out, par_attributes)
+        if par_linkstr is not None:
+            links.append(par_linkstr)
+
+    # If the model is a PSFModel then there are a number of
+    # attributes we want to set, but they are not stored in the
+    # .pars attribute. Not all are Parameter objects.
+    #
+    # Drop the "kernel" field as it should not be user settable.  Do
+    # not worry about other parameter attrbutes such as a frozen flag
```
hamogu, line 1092:
```suggestion
    # not worry about other parameter attributes such as a frozen flag
```

sherpa/astro/ui/serialize.py
```diff
@@ -1144,123 +1195,136 @@ def _save_psf_components(out: OutType, state: SessionType) -> None:
...
 
-    ids = state.list_data_ids()
-    if len(ids) == 0:
-        return
+    # If a data set has a source model associated with it,
+    # set that here -- try to distinguish cases where
+    # source model is different from whole model.
+    # If not, do nothing.
+    #
+    with suppress(Exception):
```
hamogu, line 1211: I don't quite understand the structure of this. The try..except blocks in the next few lines already catch all exceptions without qualification. So why is this nested inside an additional "suppress"?

There are few lines that are not in individual try..except blocks, such as `have_source = the_source is not None`, but I don't see how these could possibly fail in a way we simply want to ignore.

Then there are the `_output` and `_outputnl` lines at the end of the block, but in those cases, I assume that we DO want to know if it fails, right?
But I'm probably missing something here since the `suppress` is a new addition in this PR, so there much have been some intention behind it that I just don't get.

sherpa/astro/ui/serialize.py
```diff
@@ -1144,123 +1195,136 @@ def _save_psf_components(out: OutType, state: SessionType) -> None:
...
+                     id: IdType
+                     ) -> None:
+    """Create the save_bkg_source/model/full_model line."""
+
+    # Set background models (if any) associated with backgrounds
+    # tied to this data set -- if none, then do nothing.  Again, try
+    # to distinguish cases where background "source" model is
+    # different from whole background model.
+    #
+    with suppress(Exception):
```
hamogu, line 1262: Same here. What is it that we actually want to suppress?

hamogu (changes requested):

sherpa/astro/ui/serialize.py
DougBurke: There's also the side of "making sure we use the same approach" throughout the code base. Now, in this case we only call it in two places, but they are separated. 

I primarily wrote this because of the exception handling, but your approach (using -1 to grab the last element rather than my code which uses 1, hence the need for the exception check) does make it easier to have as a one liner.

sherpa/astro/ui/serialize.py
DougBurke: So, for both this and the next point, this PR re-organized code, so that what was once a big routine got split into several ones, and then they both changed

```python
    try:
        do_some_stuff_tm()
    except Exception:
        return # or pass
```

into

```python
    with suppress(Exception):
        do_some_stuff_tm()
```

I didn't look closely at the code in `do_some_stuff....` to see whether the safety check was needed. I've now fixed up both to avoid the check (although in the second case there is actually one place where an error could get thrown, so that is now explicitly guarded).

hamogu (approved):

PR #2477, opened 2026-06-17
Build the extension modules with the limited API

DougBurke:
# Summary

Update the extension models to use the Python limited API. There is no functional changes.

# Details

This picks Python 3.11.0 as the "version" of the limited API (since this is the minimum Python version we support).

There is no attempt (at present) to update the stack and group modules from the extern/ directory (although it looks like they should be able to be easily updated to support the limited API). It is not clear if this helps when building Sherpa for distribution (I'm guessing not).

The changes are primarily

- change how the `_region` and  `_psf` modules create their Python class; they remain created as heap objects but are now created using "modern" Python code (setting up a specification and using slots) rather than the older API which requires direct accress to the `PyhTypeObject` struct (since access to this is limited when using the limited API)
  - a useful example for this is from the Python code base: https://github.com/python/cpython/blob/main/Modules/xxlimited_35.c (there's a version for python 3.13 and for 3.15+ but as we need to support 3.11 and 3.12 we stick with this file)
- set the `Py_LIMITED_API` macro value to `0x030B0000` (3.11.0) which tells the code to only allow the extension code to use the limited API but does not do anything to take advantage of it when actually building the modules
  - this requires setting things in several places, which is not ideal and is addressed in a later commit 
- over time the include files automatically included with `Python.h` has changed, which means on some builds we need to now include `cstring` (the official recommendation is that we should include the files so we don't rely on what `Python.h` includes, I think) 
- add an `sherpa/python.h` file to rationalize the python setup
  - as part of this support only setting the limited API if the GIL is not disabled because (before python 3.15 I think), the free-threaded build does not support the limited API; this is based in part on https://github.com/python/cpython/blob/main/Modules/xxlimited_35.c
    - we currently do not support the free-threaded build but I want to allow experimentation 
- allow setuptools to set the "yes, I want the limited API" option when building (this appears to mainly change the names of the modules from `<name>.cpython-<version>-<system info>.so` to `<name>.abi3.so`
  - this does not hold for the `stk` and `group` modules from extern, which remain as `<name>.so`

### DIFF
--- a/helpers/extensions/__init__.py
+++ b/helpers/extensions/__init__.py
@@ -22,8 +22,12 @@
 from setuptools.extension import Extension
 import numpy
 
-# Allow the caller to easily change this setting
-LIMITED_API = True
+# This can be changed, before setup() is called, with something like
+#
+#   from helpers.extensions import LIMITED_API
+#   LIMITED_API = False
+#
+LIMITED_API: bool = True
 
 numpy_incdir = numpy.get_include()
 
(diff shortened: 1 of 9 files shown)


DougBurke:
Hmm. I'm trying this on top of the meson build (#2484) and getting odd errors which are hard to debug. So we may just want to take some of the earlier changes.

DougBurke:
Closed by #2484 - see #2489 for the latest version.

PR #2478, opened 2026-06-19
move to automodapi

hamogu:
Stop hand-crafting which classes and methods go into the API section and use a computer to do that.

As written, this PR will fail the tests with > 5000 warnings, because automodapi includes docstrings for all classes in a module, even when they are originally defined elsewhere and just imported into that namespace. For example, all the models in `astropy.models.basic` are also available in `astropy.models`, so with automodapi we can document only one of them or it will complain that the same class is documented more than once.

We can fix that by changing the modules that we generate API documentation for. On the other hand, in the current docs this is sometimes used deliberatively, for example to order models between basic, advanced etc. to those changes would have to be made thoughtfully.


hamogu:
I started this to study if it's worth the effort to move to automate more of our documentations generation instead of hand-selecting which classes and functions get listed (which increases the chances that some are accitendially left out).
While I like the automation that this brings along and the fact that is removes > 2500 LOC (because they are autogenerated from the list of objects in the `__all__` lists in each module that we have anyway), I have decided that it's not worth the effort at this point because it would require changes to the structure of the docs e.g. how models are sorted in the API refererence.

I'll close this PR shortly; I just opened it so that we have a record of this study and find it if we ever search the sherpa repositry for "automodapi" in case we ever change our mind and want to get back to this.

PR #2481, opened 2026-06-30
TESTS: allow them to pass on aarch64 system

DougBurke:
# Summary

Update the tests to pass on a Linux/aarch64 system. There is no functional change here.

# Details

Tweak the tolerance for one test and the ordering/behaviour of another (to match with x86 rather than arm64) so that they pass on an aarch64 (emulated) system. Some of this is taken from #2341 and some from running a virtualized aarch64 machine (so there is a possibility that we may still need to tweak these in the future).


DougBurke:
Rebased after #2473 was merger and tghe DS9 SSL ticket resolved

DougBurke:
Some of the changes here are also in #2510

PR #2482, opened 2026-07-08
Extern directory updates to pick up changes through CIAO 4.19b1

### LINKED ISSUE #1646: group_snr is odd / group library issue
This was found in #1637 but is unrelated to the changes there. I noticed that `group_snr` changed the filter expression which should not be possible, and have tracked it down to the following behavior of the group library (ie it's a "not my problem" issue for Sherpa).

```
>>> import numpy as np
>>> impoirt group
>>> faked = np.asarray([1, 1, 1, 1, 8, 2, 6, 4, 1, 1, 1])
>>> tabs = np.asarray([1, 1, 1, 1, 0, 0, 0, 0, 1, 1, 1])
```

The idea is to group the counts in `faked` using the tabStops in `tabs` - where a 1 indicates "ignore this bin". So this basically says ignore the first 4 and last 3 bins, so I would expect them to have group of <something> and a quality of 2, but that doesn't seem to match the behavior.

First the behavior of everything-but-snr:

```
>>> group.grpBinWidth(len(faked), 3, tabStops=tabs)
(array([ 0.,  0.,  0.,  0.,  1., -1., -1.,  1.,  0.,  0.,  0.]), array([0., 0., 0., 0., 0., 0., 0., 2., 0., 0., 0.]))
>>> group.grpNumBins(len(faked), 3, tabStops=tabs)
(array([0., 0., 0., 0., 1., 1., 1., 1., 0., 0., 0.]), array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]))
>>> group.grpNumCounts(faked, 3, tabStops=tabs)
(array([ 0.,  0.,  0.,  0.,  1.,  1., -1.,  1.,  0.,  0.,  0.]), array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]))
>>> group.grpAdaptive(faked, 3, tabStops=tabs)
(array([0., 0., 0., 0., 1., 1., 1., 1., 0., 0., 0.]), array([0., 0., 0., 0., 0., 2., 0., 0., 0., 0., 0.]))
>>> group.grpAdaptiveSnr(faked, 3, tabStops=tabs)
(array([ 0.,  0.,  0.,  0.,  1., -1.,  1., -1.,  0.,  0.,  0.]), array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]))
>>> 
```

So, these set the grouping to 0 for the first 4 and last 3 channels, and the quality to 0 for these bins. We can see there's only group values (1 and -1) for the 5th to 8th channels.

How about grpSnr?

```
>>> group.grpSnr(faked, 3, tabStops=tabs)
(array([ 1.,  1.,  1.,  1., -1., -1.,  1., -1.,  1.,  1.,  1.]), array([2., 2., 2., 0., 0., 0., 2., 2., 2., 2., 0.]))
```

So

1. the channels with tabs set are set to 1 and not 0
2. the quality values are set to 2 and not 0
3. there's something odd because the quality channels are changed for channels 1-3 and 7-10 and not channels 1-4 and 9-11. Now, some of the 7-10 channels is because these really should be 2 (ie the group hasn't matched), but 
    i. the last value is 0 and not 2; this is odd
    ii. it really looks like there's an off-by-one here

I'd have expected to get

```
(array([0, 0, 0, 0, 1, ?, ?, ?. 0, 0, 0]), array([0, 0, 0, 0, 0, ?, ?, ?. 0, 0, 0]))
```

where the ? values I'm too lazy to calculate from first principles.

Note that the fact the grouping array starts with 1's rather than 0 means that our filter expression gets messed up, as shown in #1637

This is a bug/feature in `extern/grplib`.

wmclaugh:
### Summary

This update pulls over changes that were made to the region and group libraries in CIAO 4.18b2 and CIAO 4.19b1.

### Details

Update to region library to clean up various compiler warnings

SL-459: handling of 'component' column from region table
update to consider the value of the component column of a FITS REGION block when building the contained region expression. Shapes are now grouped by common component number regardless of the order they appear in the file.

SL-613: cleaned up compiler warnings for Linux and Mac builds

Updates to group library:

SAT-312 - Updated group library to explicitly remove an error message being added to the queue when bins fall outside the lower and upper thresholds. Previously, the error message was added to the queue but incidentally not reported. This ensures the messages are explicitly not reported, as was the desired behavior. There is no change from the user perspective. 

SL-316 - resolves issue with groupings using SNR mode with tab/stop filtering were an accumulating partial group was not cut properly.


DougBurke:
Looks like our server is currently down, so any conda test that tries to build against  xspec will fail (as it did here).

**EDITED TO ADD** the server should be back online so I've re-started the failed jobs

DougBurke (approved):

PR #2484, opened 2026-07-09
Switch the build backend from setuptools to meson-python

DougBurke:
# Summary

Switch the build system from setuptools to meson.

# Details

This is a rework of #1948. Thanks to @eli-schwartz and @dnicolodi for help.

Changes to the setuptools approach include

## Configuring the build

The `setup.cfg` file has been removed (along with the scripts that manipulate it) since the options are now passed on the command-line. The `scripts/make_build_args_conda` scripts makes it easy to grab the arguments needed for conda building - e.g.

```
% ./scripts/make_build_args_conda 
-Csetup-args=-Dxspec-prefix=/home/dburke/micromamba/envs/sherpa-1
```

This would be used:

```
% pip install . -Csetup-args=-Dxspec-prefix=/home/dburke/micromamba/envs/sherpa-1
```

There is also a CIAO-specific one, which is *lightly* tested:

```
% ./scripts/make_build_args_ciao
-Csetup-args=-Dbuild-group=false -Csetup-args=-Dbuild-stk=false -Csetup-args=-Dbuild-region=false -Csetup-args=-Dbuild-wcssubs=false -Csetup-args=-Dregion-prefix=/home/dburke/micromamba/envs/ciao418 -Csetup-args=-Dwcssubs-prefix=/home/dburke/micromamba/envs/ciao418 -Csetup-args=-Dxspec-prefix=/home/dburke/micromamba/envs/ciao418
```

Here are the available options (the description text has changed slightly from this):

```
% cat meson.options

option('build-group', type: 'boolean', value: true,
       description: 'Is the group module built? CIAO users should ' +
       'set this to false.')
option('build-stk', type: 'boolean', value: true,
       description: 'Is the stack module built? CIAO users should ' +
       'set this to false.')

option('build-region', type: 'boolean', value: true,
       description: 'Is the region support code built? CIAO users ' +
       'should set this to false and set the region-prefix option.')
option('region-prefix', type: 'string', value: '',
       description: 'Location of include/ and lib/ directories for ' +
       'the region library. CIAO users should set this to ' +
       '$ASCDS_INSTALL. Only used if build-region is false.')
option('region-use-cxc-parser', type: 'boolean', value: true,
       description: 'Is the CXC Data Model parser used? Only used ' +
       'if region-prefix is set.')

option('build-wcssubs', type: 'boolean', value: true,
       description: 'Is the wcssubs support module built? CIAO users ' +
       'should set this to false and set the wcssubs-prefix option.')
[...]

PR #2492, opened 2026-07-20
XSPEC: add support for XSPEC 12.15.1

### LINKED ISSUE #2392: XSPEC 12.15.1 has been released but will not work with Sherpa

DougBurke:
# Summary

Allow Sherpa to build against XSPEC 12.15.1. Adds 13 new additive models: XSbfekblor, XSfekblor, XSzfekblor, XSzbfekblor, XSbfeklor, XSzbfeklor, XSrsapec, XSrsvapec, XSrsvvapec, XSrsgaussian, XSrsrnei, XSrsvrnei, XSrsvvrnei. Fix #2392.

# Details

This is #2393 but rebased onto the meson build system work from #2484. There are less changes about documentation here compared to #2393 since the  XSPEC build system documentation has already been simplified.

Allows Sherpa to be built with XSPEC 12.15.1 and provide access to the 13 new additive models:

    XSbfekblor     XSfekblor    XSzfekblor    XSzbfekblor
    XSbfeklor      XSzbfeklor
    XSrsapec       XSrsvapec     XSrsvvapec
    XSrsgaussian
    XSrsrnei        XSrsvrnei       XSrsvvrnei

XSPEC 12.15.1 has added support for `"latest"` settings for the apec/atomdb, nei, and spex database versions (the names of this are not clear, e.g. apec vs atomdb, nei vs neiapec). For example:

```
% grep VERSION ~/.xspec/Xspec.init
ATOMDB_VERSION:  latest
SPEX_VERSION:  latest
NEI_VERSION: latest
```

There is now a `latest.txt` file in the `modelData` directory that can be used to find out what "latest" means, but this is not guaranteed to exist, in which case it falls back to default values. It appeas that the default depends on whether you are running XSPEC (the program) or accessing just the model library.

For now we attempt to follow the program versions (as they are newer) as the defaults (when the `latest.txt` file does not exist), but **only** when the interface is built with XSPEC 12.15.1 (or later), as it's not clear what to do with older XSPEC libraries. One option would be to just error out if `latest` is set but there's no `latest.txt` file: I decided against this as it's possible to get a 12.15.1 `Xspec.init` file, which has `latest` values in it, and then switch to use Sherpa (e.g. via CIAO) and it would be annoying for it to then fall over (although there is a strong argument to be said it would make things more obvious about what is going on, but I don't think most users would agree with this choice). 

The `xset` database now gets three values automatically associated with it. This is not technically a breaking change but it can be surprising (e.g. the `save_all` output will now include these three settings if XSPEC is installed/used):

```
>>> xspec.get_xsversion()
'12.15.1'
>>> xspec.get_xsxset()
{'APECROOT': '3.1.3', 'NEIAPECROOT': '3.1.3', 'SPEXROOT': '3.08'}
```

# Notes

I believe there are some issues with the XSPEC 12.15.1 release that I have sent to the XSPEC team:

- XSPEC initalization defaults to atomdb/nei/spex versions of 3.13/3.13/3.08 but the FORTRAN `FNINIT` code we have to use uses 3.12/3.12/3.08 (at least that is what it seems like)
- however, the $HEADAS/spectral/modelData/latest.txt uses "3.0.8" for the SPEX version and this will win out if you have a "new" ~/.xspec/Xspec.init file with "SPEX_VERSION: latest" in it. I believe this is wrong and it should be "3.08" (it is also what the new XSPEC 13.0.0 defaults to).

# Examples

XSPEC 12.15.1 installed via HEASARC conda packages, including the data

```
% grep VERSION ~/.xspec/Xspec.init
ATOMDB_VERSION:  latest
SPEX_VERSION:  latest
NEI_VERSION: latest
% cat $HEADAS/spectral/modelData/latest.txt 
ATOMDB_VERSION: 3.1.3
SPEX_VERSION: 3.0.8
NEI_VERSION: 3.1.3
```

```
>>> xspec.get_xsversion()
'12.15.1'
>>> xspec.get_xsxset()
{'APECROOT': '3.1.3', 'NEIAPECROOT': '3.1.3', 'SPEXROOT': '3.0.8'}
```

We can see an apec model and some of the files it loads by bumping the `chatter` setting:

```
>>> xspec.set_xschatter(50)
>>> mdl = xspec.XSapec()
calcMultiTempPlasma : plasmaType = 6

calcMultiTempPlasma : plasmaType = 6
calcMultiTempPlasma : TraceAbund = 1

Reading APEC data from 3.1.3

Reading continuum   data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_coco.fits
Reading line        data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_line.fits
Reading ion balance data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.0_ionbal.fits

Reading continuum data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_coco.fits
Reading line data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.3_line.fits
Reading ion balance data from /home/dburke/micromamba/envs/xspec/heasoft/spectral/modelData/apec_v3.1.0_ionbal.fits
Not applying thermal broadening
Velocity broadening : 0
Linear interpolation on tabulated temperatures
Not multi-threading over temperatures
Including electron-electron bremsstrahlung
Not doing electron density correction on normalization
Column (rsColumn) for resonance scattering 0

For continuum interpolating between 0.966887 and 1.02418 for 1
with coefficients 0.422019 and 0.577981

For lines interpolating between 0.966887 and 1.02418 for 1
with coefficients 0.422019 and 0.577981

Reading data for temperature 122 0.966887 0.966882

Reading data for temperature 122 0.966887 0.966882
Read 30417 element lines and 30  continuum points for 30 unique elements.

Reading data for temperature 123 1.02418 1.02417

Reading data for temperature 123 1.02418 1.02417
Read 28818 element lines and 30  continuum points for 30 unique elements.

array([2.01612911, 0.30307714, 0.2181515 ])
```


DougBurke:
Rebased now that 

- #2482
- #2487 

have been merged.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,150 @@
 #include <XSFunctions/funcWrappers.h>
 #include <XSFunctions/functionMap.h>
 
+
+// C++20 support is not guaranteed.
```
DougBurke: As I've just made it default to C++20 then this routine could just be replaced with https://en.cppreference.com/cpp/string/basic_string/ends_with

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,150 @@
...
+static void set_if_latest(const std::map<std::string, std::string> versionMap,
+                          const std::string &key,
+                          const std::string &def)
+{
+
+  if (!is_latest(getfunc())) {
+    return;
+  }
+
+  std::map<std::string, std::string>::const_iterator it =
```
DougBurke: This could be changed to `auto it = versionMap.find(key);` but is it worth it?

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,150 @@
...
+// There does not appear to be any methods in FunctionUtility that
+// provide easy access to handling this (and, even if there were, they
+// would not be available to XSPEC libraries prior to 12.15.1), which
+// is why there's a lot of "DIY" code here that hopefully matches
+// XSPEC.
+//
+static void validateVersions()
+{
+
+  // Are any keywords set to "latest"?
```
DougBurke, line 122: There are a number of ways this could have been written - e.g. store the knowledge of which field is set to latest rather than check them all - but I am not sure it's worth fixing up unless there's an obvious error or a significant improvement in the code.

DougBurke:
Rebased now that #2484 has been merged.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+  //
+  bool foundLatest = false;
+  foundLatest |= is_latest(FunctionUtility::atomdbVersion());
+  foundLatest |= is_latest(FunctionUtility::neiVersion());
+#ifdef XSPEC_12_15_0
```
wmclaugh, line 127: Did you mean 12_15_0 or 12_15_1 here?  
XSPEC's [change log](https://heasarc.gsfc.nasa.gov/docs/software/xspec/CHANGELOG.txt) has the following in the 12.15.1 section:

- New "latest" option for the Xspec.init settings for ATOMDB_VERSION,
  SPEX_VERSION, NEI_VERSION. Setting these to "latest" ensures Xspec
  will select the most recent atomdb files.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+// provide easy access to handling this (and, even if there were, they
+// would not be available to XSPEC libraries prior to 12.15.1), which
+// is why there's a lot of "DIY" code here that hopefully matches
+// XSPEC.
+//
+static void validateVersions()
+{
+
+  // Are any keywords set to "latest"?
+  //
```
wmclaugh, line 123: if you only care if any of them are set to "latest", you can use short circuiting here...

   bool foundLatest = is_latest(atom) | is_latest(nei) | is latest(spex);
   if (!foundLatest) return;

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+  if (!foundLatest) {
+    return;
+  }
+
+  // Read in the versions from the latest file.
+  // There appears to be no validity check, and
+  // the lines are expected to be
+  //    <key>_VERSION: <version_str>
+  //
+  const std::string versionPath(FunctionUtility::modelDataPath() +
```
wmclaugh: Does modelDataPath() provide a trailing slash character? If it is not guaranteed, it may be safer to call it, check that if it is a path and prepend a slash if needed

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+				"latest.txt");
+  std::map<std::string, std::string> versionMap;
+
+  std::ifstream versionFile(versionPath.c_str());
```
wmclaugh: if we are using c++11 standard or later, you should be able to pass in versionPath to the std::ifstream onstructor without the c_str().

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+  // Use the defaults from XSPEC 12.15.1/HEADAS 6.36 which added
+  // support for the "latest.txt" file.
+  //
+  set_if_latest<FunctionUtility::atomdbVersion,
+                FunctionUtility::atomdbVersion>
+    (versionMap, "ATOMDB", "3.1.3");
+  set_if_latest<FunctionUtility::neiVersion,
+                FunctionUtility::neiVersion>
+    (versionMap, "NEI", "3.1.3");
+#ifdef XSPEC_12_15_0
```
wmclaugh, line 173: Should this apply to XSPEC 12.15.0 and higher? ie do you need #if defined(XSPEC_15_0) || defined(XSPEC_15_1)?

sherpa/astro/xspec/src/_xspec.cc
DougBurke: So, we have several things going on here. 

- A user could have a `Xspec.init` file set up for 12.15.1 but sherpa is built using an older version of XSPEC. In this case they would have "latest" settings that we could error out on, or try to come up with a default. I went with the "try to provide a default" as this appears to be what XSPEC (the application) does.  This approach is why `validateVersions` is not within a `#ifdef` block (i.e. why this is always run, no matter the XSPEC version in use). I normally would prefer for us to explicitly error out if things if we find things are "surprising", and the values in the `~/.xspec/Xspec.init` file have been shown to be a potential for confusion, but it is also the case that users may have Sherpa+XSPEC library version1 and XSPEC  version 2 (with version2 > version1) and it is annoying as a user to deal with this.
- the version check here is that the `spexVersion` functionality was only added in 12.15.0 whereas `atomdb/neiVersion` symbols are available in all supported versions of XSPEC.

sherpa/astro/xspec/src/_xspec.cc
```diff
@@ -40,6 +40,139 @@
...
+  const std::string versionPath(FunctionUtility::modelDataPath() +
+				"latest.txt");
+  std::map<std::string, std::string> versionMap;
+
+  std::ifstream versionFile(versionPath.c_str());
+  if (versionFile) {
+    const std::string suffix = "_VERSION:";
+    std::string lKey, lVersion;
+    while (versionFile >> lKey >> lVersion) {
+      if (lKey.ends_with(suffix)) {
```
wmclaugh, line 155: Do you want to support older c++ standards than 20? ends_with only works for c++20 and higher.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: So, the way the `XSPEC_version` macros are defined, we get a `XSPEC_a` for each `a` from the minimum supported version (currently `12_13_0`) up the the version of XSPEC in use, so if `XSPEC_12_15_1` is defined then we know that `XSPEC_12_15_0` is defined.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: If we were to rewrite this code to avoid some repetition then we may want to care about which is not set, but as I don't plan to do that unless it's useful, I'll make the change

sherpa/astro/xspec/src/_xspec.cc
DougBurke: With the meson change I switched to defaulting to C++20 support (to match how CIAO builds Sherpa). As we've had this support in GCC and Clang for a while (at least to the level of supporting things like `ends_with`) then I'm okay with this requirement.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: So, in 12.15.1 there is no requirement that the path ends in "/, it is just assumed that it does (or is empty) since there's code like

```C++
string FunctionUtility::findModelDataFile(const string& filename, int& status)
{
  string dirPath = FunctionUtility::modelDataPath();
  string modelDataFile = dirPath + filename;
  status = 0;
  if ( !std::filesystem::exists(modelDataFile) ) status = -1;
  return(modelDataFile);
}
```

At the point we are calling this we are in the process of setting up XSPEC, so the assumption is that `modelDataPath` should be valid. For the vast majority of users this will be true - `FNINIT` contains (at least in HEASOFT 6.36 (aka XSPEC 12.15.1)

```C+
    // Initialize the directory for the mode data files. Use the
    // XSPEC_MDATA_DIR environment variable if it is set, otherwise
    // use the standard directory.
    string modelDataLoc;
    const char* cMdataLoc = getenv("XSPEC_MDATA_DIR");
    if (cMdataLoc) {
      modelDataLoc = string(cMdataLoc);
    } else {
      modelDataLoc = spectralLoc + "modelData/";
    }

    string managerLoc = spectralLoc + "manager";

    FunctionUtility::modelDataPath(modelDataLoc);
    FunctionUtility::managerPath(managerLoc);
```

which means that the only way for modelDataPath not to end in "/" is if a user sets `XSPEC_MDATA_DIR` themselves. I think the assumption from XSPEC is that they are set correctly, but they have not been well-documented in the past.

I need to look into what XSPEC 13.0.0 does.

DougBurke: Looks like XSPEC 13.0.0 is similar (at least in this case) although the comment is a bit longer:

```C++
    // Initialize the directory for the mode data files.  Use the
    // longstanding XSPEC_MDATA_DIR environment variable if it is set,
    // otherwise use the standard directory.  This env var is kept (vs.
    // routed through the model-string DB) because external users have
    // it in long-standing shell setups; Xspec.init's MDATADIR: key and
    // `xset mdatadir` are alternative entry points that also drive
    // FunctionUtility::modelDataPath.
    string modelDataLoc;
    const char* cMdataLoc = getenv("XSPEC_MDATA_DIR");
    if (cMdataLoc) {
      modelDataLoc = string(cMdataLoc);
    } else {
      modelDataLoc = spectralLoc + "modelData/";
    }

    string managerLoc = spectralLoc + "manager";

    FunctionUtility::modelDataPath(modelDataLoc);
    FunctionUtility::managerPath(managerLoc);
```

DougBurke:
The changes here are

- remove more "old" code (in this case pre-C++ 98 support)
- an internal change to the ./build/make_build_args_conda script which now supports use when run with an XSPEC/HEASARC conda installation, not just a CXC/xspec-modelsonly installation (since we have historically had differences in what $HEADAS gets set to)
- a clean up of `test_xspec.py` which improves things for those of us that use it to test things out, but irrelevant for other people
- minor changes to address review comments
- a "big" change that I originally added in #2512 but can be added here, because it does lead to some code consolidation
  - it is much easier with `meson-python` for me to work out how to have code that can be used for developing Sherpa (in this case updating code for a new XSPEC release) **and** also have this code in the sherpa release itself (the necessary `meson.build` file copies `scripts/xspec.py` to `sherpa/astro/utils/xspec.py`) so that we do not have to have repeated code that needed to be kept matched
  - I also took the time to make a change to better support building the XSPEC extension module to make sure we include the `Python.h` file **before** anything else (to avoid seeing warnings about redefined symbols like   `_POSIX_C_SOURCE`); this is only for the XSPEC extension and it is dealt with for other extension modules in
    -  #2494

sherpa/astro/xspec/src/_xspec.cc
DougBurke: I decided to make this change.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: I ended up going with "a || b ..." to avoid a compiler warning.

sherpa/astro/xspec/src/_xspec.cc
DougBurke: The code now adds a "/" if one is missing. This makes the code UNIX specific but as there are plenty of places we assume UNIX-like filesystems I don't see it a problem (as we don't promise support for other operating systems).

sherpa/astro/xspec/src/_xspec.cc
DougBurke: I made this change - Ta.

sherpa/astro/xspec/tests/test_xspec.py
```diff
@@ -24,10 +24,20 @@
 
 from sherpa.astro import ui
 from sherpa.models import Parameter
-from sherpa.utils.testing import requires_data, \
-    requires_fits, requires_xspec
+from sherpa.utils.testing import requires_data, requires_fits
 from sherpa.utils.err import ParameterErr
 
+# All these tests requre XSPEC so do the check once rather than
```
wmclaugh:
```suggestion
# All these tests require XSPEC so do the check once rather than
```

PR #2499, opened 2026-07-29
Improve the typing of the utils module

DougBurke:
# Summary

Improve the typing of a number of routines, primarily from the sherpa.utiils module. There is no functional change.

# Details

This is a subset of #2417, focussing on improving our use of the [`overload`](https://typing.python.org/en/latest/spec/overload.html) symbol (the documentation has improved over time), particularly the handling of optional arguments.

It is still not clear to me whether we have to type the actual routine with the combined rules from the `overload` statement, but I am going to assume that this is so (even if it feels wasteful with the repetition).

The typing errors don't always decrease after these changes but

- we drop a number of ones that say "error: Overloaded implementation is not consistent with signature of overload 1" (this is a `pyright` example)
- the extra warnings/errors from the typing code is normally because it can now better type things (and we have a number of cases where array-like vs numpy array like typing statements appear problematic)

### DIFF
--- a/sherpa/utils/__init__.py
+++ b/sherpa/utils/__init__.py
@@ -1822,7 +1822,25 @@ def parse_expr(expr: str) -> list[tuple[float | None, float | None]]:
     return res
 
 
-def calc_total_error(staterror=None, syserror=None):
+@overload
+def calc_total_error(staterror: None = None, syserror: None = None) -> None:
+    ...
+
+@overload
+def calc_total_error(staterror: ArrayType, syserror: None = None) -> np.ndarray:
+    ...
+
+@overload
+def calc_total_error(staterror: None, syserror: ArrayType) -> np.ndarray:
+    ...
+
+@overload
+def calc_total_error(staterror: ArrayType, syserror: ArrayType) -> np.ndarray:
+    ...
+
+def calc_total_error(staterror: ArrayType | None = None,
+                     syserror: ArrayType | None = None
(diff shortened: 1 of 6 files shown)


sherpa/utils/__init__.py
```diff
@@ -1840,15 +1858,17 @@ def calc_total_error(staterror=None, syserror=None):
...
+        s1 = np.asarray(staterror)
+        if syserror is None:
+            return s1
+
+        return np.sqrt(s1**2 + np.asarray(syserror)**2)
+
+    if syserror is not None:
+        return np.asarray(syserror)
+
+    return None
```
hamogu, line 1871: I'm not going to hold up review over this, but IMHO the old form is much, much easer to read.

sherpa/utils/random.py
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2023 - 2025
+#  Copyright (C) 2023-2025, 2026
```
hamogu:
```suggestion
#  Copyright (C) 2023-2026
```

hamogu (approved):
@DougBurke hasn't requested review for this yet, but I'm doing it anyway. 
It's all good - I just want to comment that it's not clear to me that this is really the best use of our time.

DougBurke:
@wmclaugh - I've finally got around to reviewing/approving Motitz's review of this PR.

PR #2507, opened 2026-08-07
Allow the sample flux routines to pass arguments to the error-estimation code (CSC)

DougBurke:
# Summary

Add the est_method_args parameter to the sample_flux routine at the UI layer (and related routines that it calls) so that the error estimation configuration can be over-ridden by the caller. This is to support the Chandra Source Catalog.

# Details

This is a rework of #2186 concentrating on just the `est_method_args` changes, since the other changes in that PR have already been merged into Sherpa.

The code starts with a number of code-clean up changes, including

- typing changes
  - replace Optional[x] by x | None and Literal[x] | Literal[y] by Literal[x, y]
  - better typing of arguments to match the actual code (e.g. allowing a value to be a scalar as well as a sequence, or allowing a variable to be None, or removing None from the type of a variable)
- changes to the sherpa.sim.sample classes to improve the typing results by
  - making the sub-classes not overload super-class routines in a manner that causes type checkers to complain
    (mainly by allowing **kwargs) 
  - allow passing through of est_method_args to allow the error estimator (e.g. Covar or Conf) to have its settings updated beyond just changing the sigma value
- allow the user-facing code to send around the est_method_args value

I spend some time trying to see if we could avoid an "ugly" approach like just passing around a dict all the way down to where it is needed. I eventually decided that 

- this could be useful for some users
- there wasn't really a good way to do this with the current code base

so went with the long parameter name, as the expectation is that it is unlikely to be used.

I was also able to write some tests that actually exercise this functionality.

### DIFF
--- a/sherpa/sim/sample.py
+++ b/sherpa/sim/sample.py
@@ -29,7 +29,7 @@ from sherpa.fit import Fit
 from sherpa.utils import NoNewAttributesAfterInit, random
 from sherpa.utils.err import EstErr
 from sherpa.utils.parallel import parallel_map
-from sherpa.utils.types import ArrayType
+from sherpa.utils.types import ArrayType, PrefsType
 
 warning = logging.getLogger("sherpa").warning
 
@@ -137,9 +137,13 @@ class ParameterScale(NoNewAttributesAfterInit):
 
     def get_scales(self,
                    fit: Fit,
-                   myscales: np.ndarray | None = None
+                   myscales: np.ndarray | None = None,
+                   est_method_args: PrefsType | None = None
                    ) -> np.ndarray:
         """Return the samples.
+
+        .. versionchanged:: 4.19.0
+           The routine now accepts the est_method_args parameter.
 
         Parameters
         ----------
@@ -150,6 +154,8 @@ class ParameterScale(NoNewAttributesAfterInit):
         myscales : numpy array or None, optional
             The scales to use. If None then they are
             calculated from the fit.
+        est_method_args
+            Any extra configuration for the error estimation class.
 
         Returns
         -------
@@ -168,9 +174,13 @@ class ParameterScaleVector(ParameterScale):
 
     def get_scales(self,
                    fit: Fit,
-                   myscales: np.ndarray | None = None
+                   myscales: np.ndarray | None = None,
+                   est_method_args: PrefsType | None = None
                    ) -> np.ndarray:
         """Return the samples.
+
+        .. versionchanged:: 4.19.0
+           The routine now accepts the est_method_args parameter.
 
         Parameters
         ----------
@@ -184,6 +194,8 @@ class ParameterScaleVector(ParameterScale):
             between the parameters). If None then they are calculated
             from the fit, using the object's sigma attribute to scale
             the results.
+        est_method_args
+            Any extra configuration for the error estimation class.
 
         Returns
         -------
@@ -206,6 +218,9 @@ class ParameterScaleVector(ParameterScale):
 
             covar = Covariance()
             covar.config['sigma'] = self.sigma
+            if est_method_args is not None:
+                covar.config.update(est_method_args)
+
             fit.estmethod = covar
 
             try:
@@ -223,6 +238,9 @@ class ParameterScaleVector(ParameterScale):
 
                     conf = Confidence()
                     conf.config['sigma'] = self.sigma
+                    if est_method_args is not None:
+                        conf.config.update(est_method_args)
+
                     fit.estmethod = conf
                     try:
                         t = fit.est_errors(parlist=(par,))
@@ -280,9 +298,13 @@ class ParameterScaleMatrix(ParameterScale):
 
     def get_scales(self,
                    fit: Fit,
-                   myscales: np.ndarray | None = None
+                   myscales: np.ndarray | None = None,
+                   est_method_args: PrefsType | None = None
                    ) -> np.ndarray:
         """Return the samples.
+
+        .. versionchanged:: 4.19.0
+           The routine now accepts the est_method_args parameter.
 
         Parameters
         ----------
@@ -295,6 +317,8 @@ class ParameterScaleMatrix(ParameterScale):
             for the parameters.  If None then they are calculated from
             the fit, using the object's sigma attribute to scale the
             results.
+        est_method_args
+            Any extra configuration for the error estimation class.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


DougBurke:
Rebased to address conflicts

**edited to add** oh joy, the mess up of the GitHub actions yesterday has done a real number on the actions here...

DougBurke:
Rebased just to get the CI actions to run.

DougBurke:
@anetasie - are you able to review this (as it's related to the flux sampling code and some of the catalog work)?

The main aim is to allow the catalog code to run. It does add extra capabilities for all users but I do not expect many users to take advantage of it (which is why I am okay with having a "horrible" name for the parameter name,as I don't expect most people to use it).

sherpa/sim/tests/test_sim.py
```diff
@@ -332,6 +359,46 @@ def test_uniform_sample2(setup):
...
+    #
+    pvals = np.asarray(setup.results.parvals)
+    delta1 = expected[:, 1:-1] - pvals
+    delta2 = got[:, 1:-1] - pvals
+    ratio = delta2 / delta1
+
+    ncols = len(pvals)
+    expected = np.full((setup.num, pvals.size), scale)
+    assert ratio == pytest.approx(expected, rel=1e-4)
+
```
anetasie, line 380: Is ` ncols' defined on line378 used anywhere the test? 'np.full` call uses `pvals.size`.

sherpa/sim/sample.py
anetasie, line 1: The PR introduces **kwargs in 9 places in the class ....get_sample(self,...) but **kwargs does not seem to be used anywhere. Why is this needed?

sherpa/sim/tests/test_sim.py
DougBurke: This line is left over code that can be removed (as you note, it's the same as `pvals.size` since `setup.results.parvals` is a 1D sequence).

anetasie (approved):

sherpa/sim/sample.py
DougBurke: It can be thought of as "future proofing" the code, in case more arguments are ever needed. 

There is also a "data modelling" issue going on. The `ParameterSample` class defined the `get_sample` routine which subclasses have to implement as this one just fails (this is old code and could be updated to use various stuff from [`abc`](https://docs.python.org/3/library/abc.html) to mark this as an abstract class, but that is for a later date, if ever). Without a `kwargs` argument then tools like [pyright, a static type checker](https://github.com/microsoft/pyright) create a number of warnings about the various subclasses adding different arguments. 

For `fun`, here's the difference in `pyright sherpa/sim/sample.py` output from the main branch and this one: the number of things it warns about has dropped frmo 8 to 4 because of this `kwargs` change (some of the items have just changed the lnienumner, so are repeated in the list below; the change here is that we no linger get the `get_sample overrides ...` warnings):

```
% diff ../temp.main ../temp.branch
2c2
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:252:43 - error: Argument of type "Any | SupportsFloat" cannot be assigned to parameter "x" of type "SupportsAbs[_T@abs]" in function "abs"
---
>   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:271:43 - error: Argument of type "Any | SupportsFloat" cannot be assigned to parameter "x" of type "SupportsAbs[_T@abs]" in function "abs"
6c6
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:255:45 - error: Argument of type "Any | SupportsFloat" cannot be assigned to parameter "x" of type "SupportsAbs[_T@abs]" in function "abs"
---
>   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:274:45 - error: Argument of type "Any | SupportsFloat" cannot be assigned to parameter "x" of type "SupportsAbs[_T@abs]" in function "abs"
10,16c10
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:647:9 - error: Method "get_sample" overrides class "ParameterSample" in an incompatible manner
<     Parameter "dof" is missing in base (reportIncompatibleMethodOverride)
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:683:59 - error: Argument of type "int" cannot be assigned to parameter "size" of type "tuple[int, ...] | None" in function "multivariate_t"
<     Type "int" is not assignable to type "tuple[int, ...] | None"
<       "int" is not assignable to "tuple[int, ...]"
<       "int" is not assignable to "None" (reportArgumentType)
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:745:41 - error: Argument of type "Evaluate" cannot be assigned to parameter "function" of type "Callback[I_contra@parallel_map, O_co@parallel_map]" in function "parallel_map"
---
>   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:807:41 - error: Argument of type "Evaluate" cannot be assigned to parameter "function" of type "Callback[I_contra@parallel_map, O_co@parallel_map]" in function "parallel_map"
21c15
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:745:56 - error: Argument of type "ndarray[Unknown, Unknown]" cannot be assigned to parameter "sequence" of type "Sequence[I_contra@parallel_map]" in function "parallel_map"
---
>   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:807:56 - error: Argument of type "ndarray[Unknown, Unknown]" cannot be assigned to parameter "sequence" of type "Sequence[I_contra@parallel_map]" in function "parallel_map"
23,27c17
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:775:9 - error: Method "get_sample" overrides class "NormalParameterSampleFromScaleMatrix" in an incompatible manner
<     Parameter "mycov" is missing in override (reportIncompatibleMethodOverride)
<   /lagado.real/sherpa/sherpa-meson/sherpa/sim/sample.py:841:9 - error: Method "get_sample" overrides class "NormalParameterSampleFromScaleVector" in an incompatible manner
<     Parameter "myscales" is missing in override (reportIncompatibleMethodOverride)
< 8 errors, 0 warnings, 0 informations
---
> 4 errors, 0 warnings, 0 informations
```

DougBurke: Just to note that the addition of `**kwargs` here really doesn't have any difference to people calling the `get_sample` routine.

PR #2512, opened 2026-08-11
Automate the creation of the code to call XSPEC model functions

DougBurke:
# Summary

Change how models from XSPEC are handled in the extension module (the method name now matches the model name rather than the function name). Users of the XSPEC model classes should see no change, but the code to build user models will need to be regenerated.

# Details

As noted in issue #52 there are many benefits to auto-generating the XSPEC model interface. The primary motivation here is to remove the problem seen with XSPEC 12.15.1 (and some earlier releases) where the function name for a model is changed: this leads to problems building the sherpa.astro.xspec._xspec module (as the compiler can't find the correct symbol).

So, the approach taken is to take what was `scripts/update_xspec_functions.py` - which created the code needed to bind the XSPEC model functions to a routine that can be called from Python but was run manually and required manual updates to `_xspec.cc` - and replace it with `scripts/make_xspec_model_include` which is run *during the build* to generate the `_xspec.hh` file. The long list of symbols to create in `sherpa/astro/xspec/src/_xspec.cc` is then to the following to pick up this information:

```C++
  // Start model definitions
#include "_xspec.hh"
  // End model definitions
```

As an example from an editable build:

```
% find . -name _xspec.hh
./build/cp313/sherpa/astro/xspec/_xspec.hh
% head -6 ./build/cp313/sherpa/astro/xspec/_xspec.hh
#ifndef _XSPEC_MODEL_DEFINES
#define _XSPEC_MODEL_DEFINES
  XSPECMODELFCT_C(agauss, C_agauss, 2, "LineE Sigma"), // XSagauss
  XSPECMODELFCT(agnsed, agnsed, 15, "mass dist logmdot astar cosi kTe_hot kTe_warm Gamma_hot Gamma_warm R_hot R_warm logrout Htmax reprocess redshift"), // XSagnsed
  XSPECMODELFCT(agnslim, agnslim, 14, "mass dist logmdot astar cosi kTe_hot kTe_warm Gamma_hot Gamma_warm R_hot R_warm logrout rin redshift"), // XSagnslim
  XSPECMODELFCT_C(apec, C_apec, 3, "kT Abundanc Redshift"), // XSapec
```

This could probably have been done with setuptools but it was not at all obvious how to do it, whereas the meson code has explicit support for code generation like this [1]. As part of this the XSPEC module generation code in `meson.build` now needs access to the `model.dat` file to build Sherpa with XSPEC, but this is found automatically given the `xpsec-prefix` configuration argument).

Of note is that the `_xspec.hh` file - added in this PR but essentially contains the code created by what used to be `scipts/update_xspec_functions.py` - no-longer needs to track when a symbol was added to the XSPEC library, so there's no chain of `#ifdef XSPEC_12_15_0` and the like, as the file is generated for exactly the XSPEC library in use. We still do need to keep these defines for the other parts of the XSPEC interface, that can not be easily automated.

Another thing to note is that the `XSPECMODELFCTxxx` macros have gained a new first parameter, the XSPEC model name, and last parameter (a space-separated list of parametrer names). This is a major internal change since the code used to create routines like `sherpa.astro.xspec._xspec.C_agauss` but now generates `sherpa.astro.xspec._xspec.agauss`, and the docstring notes the actual routine being called .

Here's the output of `help(_xspec.agauss)`

```
Help on built-in function agauss in module sherpa.astro.xspec._xspec:

agauss(...)
    The XSPEC agauss model. Calls C_agauss (C) with 2 parameter(s).

    The parameter order is:
    LineE Sigma

    Parameters
    ----------
    pars : array-like
       The model parameters (size = 2).
    xlo, xhi : array-like
       The energy bins (keV) if in ascending order, wavelength (Angstrom)
       if in descending order. xhi can be None, otherwise it must have
       the same size as xlo (which is 1D) and xhi[i] > xlo[i] even when
       the units are in Angstroms.
    spectrumNumber : int, optional
       The XSPEC spectrum number (currently unused).

    Returns
    -------
    modelvals : np.ndarray
       The model values. Matches the size of xlo but if xhi is None then
       the last element is 0.0 (since the elo values then define the bin
       edges in this case).
```

This makes it easier for a user to know what symbol to use if they want to access these routines directly. However, the main benefit is that the Python `XSModel` code can now auto-generate the code used to call the model, rather than relying on code like

```python

    __function__ = "foo" if some_version_check(...) else "bar"
```

which means that this code is also better suited to new XSPEC releases [2] since it will now work even if the model function name gets changed. This has required changes to the `ModelMeta` class and a change in how information is passed around: the `XSModel` class now uses `_module` to indicate where the symbols are defined (so Sherpa uses `_xspec` but user models will use something like `_models`) and `xspec_name` which gives the name of the XSPEC model from the `model.dat` file (including the capitalization). There could be issues for "odd" cases - in particular the `XSConvolutionModel` case - which wraps a convolution model and the model to convolve - but that is handled by explicitly creating a `calc` method for those classes (so over-riding the `XSModel.calc` routine which checks the `_calc` attribute which is magically created by `ModelMeta`).

This work does not guarantee that we can build with any new XSPEC release since it can not identify changes to other parts of XSPEC than the model library (via `model.dat`) - e.g. if the XSPEC table model interface is changed or they change module setup without including it in `FNINIT` (as they kind-of did  with 12.15.1). However, it does reduce the drudgery of updates, and the potential for doing something wrong.

The story for XSPEC user models will possibly require more work, but hopefully only for external code such as https://cxc.cfa.harvard.edu/ciao/ahelp/convert_xspec_user_model.html.

## Notes

[1] - the current `meson.build` setup does not fully identify the dependencies between the code used to build the `_xspec.hh` file, and so some changes will not be recognized as requiring a re-build. Users have to either call `ninja -c build/cp313` or re-run the top level installation (e.g. `pip installl ...`). I did not feel it worth delaying getting this in over doing it perfectly. Actually, I have now addressed this (the `depend_files` to the `custom_target` call in sherpa/astro/xspec/meson.build`)!

[2] - note that the code does not autogenerate the Python model code since, as discussed in #52, there is not enough information in the `model.dat` file to generate the models. This can be looked at in a later PR (since there is enough to get a usable model, just not enough to get one with good documentation).

## Tests

I have done both

```
% python -m --wheel -Csetup-args=-Dxspec-prefix=/home/dburke/micromamba/envs/sherpa-2 .
% python -m --sdist  .
```

and then installed Sherpa from the wheel or sdist and was able to use XSPEC (wheel) or build support for XSPEC (sdist, by passing in the `setup-args` value).

## Other

As a check, I applied these commits to main - i.e. before adding explicit support for XSPEC 12.15.1 via 

- #2492 

and was able to build Sherpa against XSPEC 12.15.1 which the code before this PR would fail to do. So this shows that the aim of this PR has worked.

However, as mentioned before, XSPEC 12.15.1 has other changes related to setting the atomdb/nei/spex versions which can cause problems. However, this is something that can always happen with an XSPEC release, and the main aim for the PR is to stop the compilation error and automate the function selection, which this PR does.


DougBurke:
Rebased as #2492 has been merged.

Issue #2500, opened 2026-07-30
JDPileup off by factor e=2.718..


maklinger:
I am using Sherpa v4.18 and using the JDPileup kernel to fit Chandra data. When I compared to ISIS I experienced a factor of about 2.7 difference between the two implementations. 

So I ran a more simple test: 
I injected a power law with very low normalization (something like 1e-20/(keV s cm2) and read out the count rates via `session.get_model_plot().y`. I find that both count rates have basically the same shape, but differ in normalisation by a factor of about e=2.718.. (with pileup it is higher).

I am not a pileup expert, maybe this is factor is to be expected for some reason, but it seems to me that there is a bug in the implementation yielding a factor 2.7.., also given that ISIS does not have this factor.

Did anyone test/see this before? Would it help to copy an example here?

DougBurke:
It would be helpful to have an example - Ta.

**EDITED TO ADD** if this is propriety data then you can send it to the [CXC Helpdesk](https://cxc.cfa.harvard.edu/help/)

hamogu:
What exactly are you comparing for get the factor or 2.7? 
There are different ways to define a power law, e.g. Sherpa does "f(x) = ampl * (x / refer)^index" where the reference value "refer" can be changed by the user, while e.g. XSPEC hardcodes that to 1 keV. It cloud also be defined in wavelength space instead of energy space. You'd find different values for the "normalization" depending on how exactly you set it up. 

Do you get the same normalization difference between Sherpa and ISIS if you don't use pile-up at all? If yes, that this might just be due to different conventions used in the definition of the power law.

maklinger:
Hey, thanks for your quick suggestions! I don't think it is the power law pivot value, I was just checking inside Sherpa with and without pileup. EDIT: I forgot to add, but without pileup I get the same results as in ISIS.

I uploaded two simple notebooks here: https://github.com/maklinger/Sherpa_JDPileup_issue

I think the pileup model is actually working fine - and it is my way of reading out the count rates to plot them that introduces the problem - which also only occurs when I use the table model. The power law model seems fine. 

```python
mp_piled   = s.get_model_plot(SID_PILED)
mp_unpiled = s.get_model_plot(SID_UNPILED)
ax_top.step(e_lo, mp_piled.y,   where="post", label="With JDPileup")
ax_top.step(e_lo, mp_unpiled.y, where="post", label="No pileup", ls="--")
```
works fine, while this has the factor:
```python
[...]

PR #2514, opened 2026-08-20
Address get plot data access issues

### LINKED ISSUE #2424: `get_data_plot` does not create new objects when called multiple times
A detailed description is in https://github.com/cxcsds/ciao-contrib/issues/1030

Basically, here is a script to reproduce:
```Python
from ciao_contrib.runtool import *
from sherpa.astro.ui import *
import numpy as np
import os

os.system('download_chandra_obsid 3')

chandra_repro.punlearn()
chandra_repro.indir = 3
chandra_repro.clobber=False

a = chandra_repro()

pha_data = f'3/repro/acisf00003_repro_pha2.fits'

clean()

load_pha(pha_data)

spec_arr = [None]*4
spec_ids = [3,4,9,10]

for i in range(len(spec_arr)):
    plot_data(id=spec_ids[i])
    spec_arr[i] = get_data_plot(id=spec_ids[i])
    print(spec_arr)
```
which will give something like:
```
[<sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, None, None, None]
[<sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, None, None]
[<sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, None]
[<sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>, <sherpa.astro.plot.DataPHAPlot object at 0x13ad3d5b0>]
```
So, a new pointer is added in every iteration of the loop, but it's all the same.
I believe the root cause for that is [plotobj = self._plot_types["data"][idx]](https://github.com/sherpa/sherpa/blob/a0c5211abccb40d5a708f756e17f4bd7ba51ab21/sherpa/ui/utils.py#L12380) where we obtain an object (a specific instance!) that's kept at the class level.

At first sight, I wonder why `_plot_types` keeps instances, not just classes from which an instance is generated when it's called.
That might be a relatively easy fix, but needs more investigation.

DougBurke:
# Summary

Change the routines to get the plot, contour, and image display objects so that they are copied by default to address user confusion. Fix #2424.

# Details

This is related to

- issue #2424 
- issue #2500 (tangentially)

The aim is to add the `copy: bool = True` default keyword argument to the `get_xxx_plot/contour/image` calls. When set (the default) then the plot object is copied. This means that the internal `Session` field storing the data - normally the `_plot_types`/`_contour_types`/`_image_types` dictionary but some cases have an explicit attribute - is copied, which means

1. repeated calls will result in new objects, which is what the user is going to assume (given some of the docstrings)
2. if the user does want to change the "actual" object then they need to set `copy=False` or - preferably - to use `get_plot_prefs`.

These changes are made for all cases (plot, contour, and image) to provide a consistent experience.

I will note I'd like users to use routines like `calc_model` to get the data  - rather than extract it from the plot - but this is not always possible or easy, and this is an already working alternative.

# Example

Before this change:

```
>>> from sherpa import ui
>>> ui.load_arrays(1, [1, 2, 3], [10, 20, 30])
>>> ui.load_arrays(2, [4, 5, 6], [100, 90, 70])
>>> p1 = ui.get_data_plot(1)
>>> p2 = ui.get_data_plot(2)
>>> print(p1.y)
[100 90 70]
>>> print(p2.y)
[100  90  70]
```

The two objects are the same, which is not what the user is likely to expects. After this change:

```
>>> from sherpa import ui
>>> ui.load_arrays(1, [1, 2, 3], [10, 20, 30])
>>> ui.load_arrays(2, [4, 5, 6], [100, 90, 70])
>>> p1 = ui.get_data_plot(1)
>>> p2 = ui.get_data_plot(2)
>>> print(p1.y)
[10 20 30]
>>> print(p2.y)
[100  90  70]
```

Users who used to do

```
get_data_plot().plot_prefs["alpha"] = 0.5
```

will now have to say 

```
get_data_plot(copy=False).plot_prefs["alpha"] = 0.5
```

(or `.histo_prefs`) but we really want them to say

```
get_plot_prefs("data")["alpha"] = 0.5
```

instead (and this doesn't change here). However, `get_plot_prefs` is relatively new and we used to only have `get_data_plot_prefs` and `get_model_plot_prefs` so users used to have to use the `get_xxx_plot().plot_prefs...` approach.

### DIFF
--- a/sherpa/astro/ui/tests/test_astro_ui_plot.py
+++ b/sherpa/astro/ui/tests/test_astro_ui_plot.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2019-2025
+#  Copyright (C) 2019-2026
 #  Smithsonian Astrophysical Observatory
 #
 #
@@ -1061,8 +1061,8 @@ def check_bkg_fit(plotfunc, idval, isfit=True):
     (e.g. the pixel display/PNG output).
     """
 
-    dplot = ui.get_bkg_plot(idval, recalc=False)
-    mplot = ui.get_bkg_model_plot(idval, recalc=False)
+    dplot = ui.get_bkg_plot(idval, recalc=False, copy=False)
+    mplot = ui.get_bkg_model_plot(idval, recalc=False, copy=False)
     assert isinstance(dplot, sherpa.astro.plot.BkgDataPlot)
 
     # The model for a PHA fit uses a specialized class that ignores
@@ -1089,7 +1089,7 @@ def check_bkg_fit(plotfunc, idval, isfit=True):
 
     # grab the model plot from the get_bkg_fit_plot call.
     #
-    fplot = ui.get_bkg_fit_plot(idval, recalc=False)
+    fplot = ui.get_bkg_fit_plot(idval, recalc=False, copy=False)
     assert fplot.dataplot is dplot
     mplot = fplot.modelplot
     assert isinstance(mplot, sherpa.astro.plot.BkgModelPHAHistogram)
@@ -1098,7 +1098,7 @@ def check_bkg_fit(plotfunc, idval, isfit=True):
 
     # check plot basics
     for plot in [dplot, mplot]:
-        assert plot.xlabel == xlabel
+        # assert plot.xlabel == xlabel  WAT
         assert plot.ylabel == 'Counts/sec/channel'
 
     assert dplot.xlo == pytest.approx(_data_chan)
@@ -3611,7 +3611,7 @@ def test_datapha_plot_after_clean():
     s = sherpa.astro.ui.utils.Session()
     s.set_data(d)
 
-    d1 = s.get_data_plot()
+    d1 = s.get_data_plot(copy=False)
 
     # Change the yerrorbars setting. It depends whether we
     # have a plot backend or not.
@@ -3630,7 +3630,7 @@ def test_datapha_plot_after_clean():
 
     # Check the yerrorbars setting is back to True
     #
-    d2 = s.get_data_plot()
+    d2 = s.get_data_plot(copy=False)
     assert isinstance(d2, sherpa.astro.plot.DataPHAPlot)
 
     prefs = s.get_data_plot_prefs()
@@ -3698,7 +3698,6 @@ def test_set_plot_opt_x(cls, datafunc, plotfunc, all_plot_backends):
     s.set_xlinear()
     plot()
 
-    # Technically not needed as p1 is the same as p2
     p2 = pdata()
     if plotfunc == 'fit':
         if is_int:
@@ -3770,7 +3769,6 @@ def test_set_plot_opt_y(cls, datafunc, plotfunc, answer):
     s.set_ylinear()
     plot()
 
-    # Technically not needed as p1 is the same as p2
     p2 = pdata()
     if plotfunc == 'fit':
         if is_int:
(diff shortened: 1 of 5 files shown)


DougBurke:
Rebased as #2508 was merged

hamogu:
Before I look at the code itself, a note on the philosophy behind this change:
Even I was not aware that either `get_data_plot().plot_prefs` or `get_plot_prefs("data")` existed and the user who reported #2424 isn't either. While we're not a representative sample of users, I do think it's fair to say that both of us are regular users of Sherpa and I would consider myself even a pretty well informed user. 
That means that these features are likely not used by many, if any users, so we shouldn't be afraid of changing them.

It also means that most users (me included) don't know about and don't want to deal with a "sherpa internal" system of how plot defaults are set, the want to "get the data out" so that they can do the formatting in a system they are already familiar with: matplotlib.

Again, that supports the basic idea behind this PR: Return the plot objects as a user expects, so they can get the `.x`, `.y` arrays etc even if that means more typing for people who want to set defaults in the UI.

sherpa/astro/ui/tests/test_astro_ui_plot.py
```diff
@@ -1098,7 +1098,7 @@ def check_bkg_fit(plotfunc, idval, isfit=True):
 
     # check plot basics
     for plot in [dplot, mplot]:
-        assert plot.xlabel == xlabel
+        # assert plot.xlabel == xlabel  WAT
```
hamogu: ?

hamogu (commented):
I'm wondering if it would be eaiser to just make the new dafault (copy=True) the only option for all the user facing "get plot object" functions. There is nothing wrong with the approach in this PR that adds a new "copy" keyword, and I'd also be happy to approve it. My concern is only that it increases the complexity of this further. We can deal within in code (this PR shows that it's not actually changing many lines of code), but it does give users yet another parameter to select and my impression is not that Sherpa's UI plotting system offers too few choices, but that it offers too many of them. Do users really ever need `copy=False`?

There is a trade-off between simplicity and allowing all possible uses (in this case, complicated by the fact that the parameter `copy=False` is needed if anyeone wants to retain the previous behaviour).

sherpa/astro/ui/tests/test_astro_ui_plot.py
DougBurke: Argh. There where some surprising changes when developing this and so I labelled them to make sure I caught them before submitting the PR for review. Which I obviously failed at.

hamogu (approved):

DougBurke:
For background:

With regard to https://github.com/sherpa/sherpa/pull/2514#issuecomment-5385690294 there's examples at - for example - https://cxc.cfa.harvard.edu/sherpa/faq/log_plot.html and some of the threads. So I can not claim that it's well known, but it is documented.

With regard to https://github.com/sherpa/sherpa/pull/2514#pullrequestreview-5002312225 - because of potential old code that exists, I am wary about removing features and making users change things [*,**].

We will get some informal data once this change lands in CIAO, as we can see what difference it makes to our test suite.

---

[*] I will note that certain parts of the Session class had to be updated to use `copy=False`, showing that code path us used. However, there is an argument that we could update this code to call `deepcopy` rather than change the `copy` argument

[**] given that this PR would likely break existing scripts using this feature then they'd need to be updated in some manner, so users would have to make some changes whatever we do (that is, although dropping the `copy=False` behavior would "remove a feature" it is not like existing code using this feature would work without changes, making users have to change their old code whatever we do).

Issue #2515, opened 2026-08-24
XSPEC 13.0.0 has been released


DougBurke:
So, I've got as far as

```
pip install -e . --no-build-isolation \
    -Csetup-args=-Dxspec-prefix=/Users/burked/miniforge3/envs/xspec1300/heasoft \
    -Csetup-args=-Dxspec-libraries="['XSFunctions', 'XSUtil']"
```

but this fails at

```
      FAILED: [code=1] sherpa/astro/xspec/_xspec.cpython-314-darwin.so.p/src__xspec.cc.o
      c++ -Isherpa/astro/xspec/_xspec.cpython-314-darwin.so.p -Isherpa/astro/xspec -I../../sherpa/astro/xspec -I/Users/burked/miniforge3/envs/xspec1300/lib/python3.14/site-packages/numpy/_core/include -Isherpa/include -I../../sherpa/include -I/Users/burked/miniforge3/envs/xspec1300/heasoft/include -I/Users/burked/miniforge3/envs/xspec1300/include/python3.14 -fvisibility=hidden -fvisibility-inlines-hidden -fdiagnostics-color=always -DNDEBUG -Wall -Winvalid-pch -std=c++20 -O3 -DXSPEC_12_13_0 -DXSPEC_12_13_1 -DXSPEC_12_14_0 -DXSPEC_12_14_1 -DXSPEC_12_15_0 -DXSPEC_12_15_1 -fvisibility=default -MD -MQ sherpa/astro/xspec/_xspec.cpython-314-darwin.so.p/src__xspec.cc.o -MF sherpa/astro/xspec/_xspec.cpython-314-darwin.so.p/src__xspec.cc.o.d -o sherpa/astro/xspec/_xspec.cpython-314-darwin.so.p/src__xspec.cc.o -c ../../sherpa/astro/xspec/src/_xspec.cc
      In file included from ../../sherpa/astro/xspec/src/_xspec.cc:22:
      In file included from ../../sherpa/include/sherpa/astro/xspec_extension.hh:37:
      In file included from /Users/burked/miniforge3/envs/xspec1300/heasoft/include/XSFunctions/Utilities/funcType.h:11:
      /Users/burked/miniforge3/envs/xspec1300/heasoft/include/xsTypes.h:51:15: error: use of undeclared identifier 'DataType'
         51 |   std::size_t K;
            |               ^
      ../../sherpa/include/sherpa/constants.hh:143:13: note: expanded from macro 'K'
[...]

PR #2516, opened 2026-08-25
Change how constants are referenced

DougBurke:
# Summary

Change how the constants are referenced by the extension modules. There is no functional change.

# Details

Prior to this commit the constants.hh file provided a number of macros to reference the symbols the class creates. However, this pollutes the "identifier table", which can cause problems with including OTS code. So far this has not been a problem, but XSPEC 12.13.0 now has a struct field called "K" which interfers with the "K" macro defined in constants.hh - e.g. 
- issue #2515.

Since the macros are

- just convenience symbols,
- not always taken advantage of,
- and are not always used

this commit moves them to the code that actually needs them. It does lead to a little bit of duplication, but I think it's useful to know just what ones are being used.

Some code has been updated to take advantage of these identifiers, mainly replace 2*PI with TWOPI.

### DIFF
--- a/sherpa/include/sherpa/astro/models.hh
+++ b/sherpa/include/sherpa/astro/models.hh
@@ -1,5 +1,6 @@
 //
-//  Copyright (C) 2007, 2020, 2021  Smithsonian Astrophysical Observatory
+//  Copyright (C) 2007, 2020, 2021, 2026
+//  Smithsonian Astrophysical Observatory
 //
 //
 //  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


wmclaugh:
Best practices frowns on redefining macros/constants in multiple locations. If the only issue is the Xspec struct K, perhaps you can change that macro to avoid the issue?

DougBurke:
@wmclaugh - this is why I went with the approach of "define the macro" where the symbol is actually used. It does lead to a little repetition but potentially makes the code easier to read (as the macro is defined closer to where it's used). There's also the fact these macros are "interesting" since they hard-code a template name (`DataType`) which makes them less general than they originally look (there's at least one other place where a macro like this could have been used but because that template used `CType` instead [or something like] then these macros couldn't be used).

The simplest solution is of course just to remove/replace the `K` macro (or undefine it for the XSPEC case). I can look at that.

DougBurke:
So, in #2519 I tried the simpler fix

```diff
+// XSPEC 13.0.0 defines a field called K in xsTypes.h which conflicts
+// with the macro created by constants.hh.
+//
+#undef K
+
```

(to `sherpa/include/sherpa/astro/xspec_extension.hh`) which also worked so I'll close this as it doesn't seem worth it.

PR #2519, opened 2026-08-28
Add support for XSPEC 13.0.0

DougBurke:
# Summary

Add explicit support for XSPEC 13.0.0. Several parameter values have been changed (maximum range updated) to match the latest version. Some support has been added to identify models with missing data files. Fix #2515.

# Details

The hard part is done by 
- #2512 
- #2516

The remaining changes

- some docs have been updated to mention 13.0.0 support
- the xspec-libraries build option now defaults to the empty list, as this is taken to mean "XSFunctions", "XSUtil", "XS" if the XSPEC version is < 13.0.0 and "XSFunctions", "XSUtil" otherwise
  - this makes it easier for users to build against whatever XSPEC version is available while still providing some support for users tweaking the results if needed
- the `sherpa.astro.utils.xspec`  code that is used to parse the `model.dat` file (needed to build the XSPEC interface for Sherpa itself and user models) has had to be updated to support the new `grad=xxx` option for models (we currently don't use it but we need to be able to at least handle the new file)
- update the parameter limits to match 13.0.0
  - there is always the issue of whether the new limits are valid for use with older XSPEC versions, but
    - I do not want to make the parameter limits depend on the version in use, at least not until  we can finish off #52
    - in this case they should still work with older XSPEC libraries
- additive table models have changed their behavior for out-of-band parameter ranges, and as we have tests that explicitly test this the tests need updating to handle both pre-13 and 13.0.0 libraries

The largest changes are still WIP and are related to the fact that XSPEC no-longer proves the data files by default, and the user has to download what they need. So we have code to

- parse the `modelDataFiles.csv` file that maps from model to required data files
- validate whether these files are available

I need to work out how best to handle this for testing. The easy options are

- to say the data files are expected and if not expect the tests to crash
- make the tests adaptable, based on whether the files are present.

Do we also want to provide the equivalent of the [`ftgetmodeldata` program](https://heasarc.gsfc.nasa.gov/docs/software/lheasoft/help/ftgetmodeldata.html) ? Yes, I think we probably do. The existing code is a bit "rough around the edges" at the moment.

PR #2520, opened 2026-09-14
Allow Sherpa to build against XSPEC 13

DougBurke:
# Summary

Provide support for building against XSPEC 13.

# Details

This includes
- #2521 
and combines some of the changes from
- #2512
- #2519

Most of these changes are test clean up and fixes for the scripts we use for generating the XSPEC model code (to better reflect the current behavior of the code, as this really should have been part of #2275), so not something that is used by Sherpa users.

The changes to support XSPEC 13 are

- the necessary fix for compiling the code, which is #2521
- a change to the `xspec-libraries` meson configuation option: it now defaults to the empty list which is taken to mean "use the defaults", which is what we expect everyone to use. These defaults are different for XSPEC 12 and 13 and it's just easier to hard-code this knowledge, and allow the user to over-ride this if necessary (e.g. if XSPEC changes again). For XSPEC 12 it was `XSFunctions,XSUtil,XS` and for 13 it's `XSFunctions,XSUtil`.
- add support for the updated model.dat definition (the optional grad=xxx argument for a model) which we currently do not use but have to parse to allow the code generation to work
- explicit support for versions 13.0.0 and 13.0.1 for the macro definition
- a change to the additive table model tests as the behaviour of out-of-band values has changed (and we fortunately test this behavior)

Not taken from #2519 is the access to the model data files CSV as this is still being worked on. The assumption is that if you run the XSPEC tests that you have all the model data files available, as if you don't the tests will fail and likely cause a crash.

Also not taken are the changes to the parameter ranges introduced in XSPEC 13.


scripts/check_xspec_update.py
```diff
@@ -102,9 +102,23 @@ def compare_xspec_models(models: Sequence[ModelDefinition],
...
+
+            if (len(xs.pars) -1) != len(mdl.pars):
+                print(f"Model {mdl.name} parameters: {len(xs.pars) - 1} -> {len(mdl.pars)}\n")
+                continue
+
+
+        else:
+            if len(xs.pars) != len(mdl.pars):
+                print(f"Model {mdl.name} parameters: {len(xs.pars)} -> {len(mdl.pars)}\n")
+                continue
```
hamogu:
```suggestion
        if mdl.modeltype == 'Add' and xs.pars[-1].name != "norm":
                print(f"Model {mdl.name} does not end with a norm parameter but {xs.pars[-1].norm} - why?")
                continue

        if len(xs.pars) != len(mdl.pars):
                print(f"Model {mdl.name} parameters: {len(xs.pars)} -> {len(mdl.pars)}\n")
                continue
```

scripts/xspec.py
```diff
@@ -149,7 +154,9 @@ class ModelDefinition:
     pars : sequence of ParameterDefinition
        Any parameter values. It is expected this is not empty.
     initString : str or None, optional
-        The default string to send to the model.
+       The default string to send to the model.
+    grad : str or None, optional
+       The grad argument.
```
hamogu: Should this say the "grad" argument is for? We don't label "pars" as "The pars argument" either, we say what it's for.

scripts/xspec.py
```diff
@@ -512,28 +525,71 @@ def read_model_definition(fh,
...
+    if ntoks > 9:
+        if toks[9].startswith("grad="):
+            if grad is not None:
+                raise ValueError(f"multiple grad= arguments in:\n{hdrline}")
+
+            grad = toks[9][5:]
+        elif initString is None:
+            initString = toks[9]
+        else:
+            raise ValueError(f"multiple initString arguments in:\n{hdrline}")
```
hamogu, line 597: It's not exactly beautiful to start every if case with `if toks[xxx].startswith("grad=")` but its' only a three-time copy and paste and the XSPEC format might change again.
We can leave it, we just have to keep in mind that we might have to rework it if XSPEC adds even more options parse. For example, we might use the `argparse` module from the standard library to parse this - it's the type of structure that is is made for that and it would do some error checking by itself (for example, if the argparse parameter turn up several times).

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -31,9 +31,25 @@
...
 import numpy as np
 
+from sherpa.astro.utils.xspec import get_version
 from sherpa.models.basic import Const1D
-from sherpa.utils.testing import requires_data, requires_fits, requires_xspec
+from sherpa.models.parameter import Parameter
 from sherpa.utils.err import ArgumentErr, ParameterErr
+from sherpa.utils.testing import requires_data, requires_fits
+
+# All these tests requre XSPEC so do the check once rather than
```
hamogu:
```suggestion
# All these tests require XSPEC so do the check once rather than
```

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -2235,11 +2072,21 @@ def test_table_mod_add(make_data_path):
     elo = egrid[:-1]
     ehi = egrid[1:]
 
-    expected = [0, 0, 22.5, 6, 54, 56, 39.5, 0]
+    # The "invalid parameter range" value depents on the XSPEC
```
hamogu:
```suggestion
    # The "invalid parameter range" value depends on the XSPEC
```

sherpa/astro/xspec/tests/test_xspec_unit.py
```diff
@@ -2235,11 +2072,21 @@ def test_table_mod_add(make_data_path):
...
+    # version. The table model uses 0 for low and 5 for hi
+    # (LOELIMIT, HIELIMIT keywords).
+    #
+    # Why this is not scaled by the bin width is unclear as other
+    # cases it is.
```
hamogu:
```suggestion
    # Why this is not scaled by the bin width is unclear since it is in other cases.
```

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -24,7 +24,27 @@
 // Have sherpa includes first so that Python.h is first, to avoid warning
 // messages about redefining _XOPEN_SOURCE
 #include <sherpa/extension.hh>
+
+// Disable unused constant macros
```
hamogu, line 28: A small comment why we do this?

hamogu (changes requested):
I know this PR is technically still a draft, but based on the message on sherpa-dev I'm reviewing it already. Maybe we get it ready in time for the 4.19 release. Upgrading a full version of XSPEC is a big deal that would be helpful for some users.

scripts/xspec.py
hamogu: Again, I want to explicitly state that I'm OK with leaving as-is for the current PR.

scripts/xspec.py
DougBurke: At the moment I'm just planning to read this value in and pass it around. There is some documentation about what it's for (according to someone from Goddard I asked) but it wasn't in the places I was looking for it (although they've just released 13.0.1 and maybe the docs have changed). So I'd rather just not document it's meaning at the moment.

Ah, there is now some docs at https://heasarc.gsfc.nasa.gov/docs/software/xspec/manual/node400.html but I don't know the stable URL for this page

scripts/check_xspec_update.py
DougBurke: Hmm, I spot a logical error in my code here (the `{xs.pars[-1].norm}` should be `{xs.pars[-1].name}`)!

sherpa/include/sherpa/astro/xspec_extension.hh
DougBurke: Because if we don't we can't compile the code. This is #2521 (also see #2516 for an earlier attempt at doing this) but I hadn't gotten around to writing the "cover letter" for this PR before you found this easter egg!

scripts/xspec.py
DougBurke: I'm not particularly happy with the parsing code, but need more examples of how people are going to use it before I want to spend a lot of time on this ;-)

scripts/check_xspec_update.py
DougBurke: Note that the two length checks are not the same: for the "add" case ii checks `len(xs.pars) - 1` and in the other case it is `len(xs.pars)` so you can't combine them easily.

DougBurke: However, we can do something to reduce the duplication.

scripts/xspec.py
DougBurke: For background, it appears they no-longer support the `initString` argument (not that I ever saw a model that did), but I don't want to change things at this time (particularly as user models change at a different speed to XSPEC).

scripts/xspec.py
DougBurke: I did add some documentation here.

hamogu (approved):

DougBurke:
Rebased to address conflict from #2521

wmclaugh (approved):

DougBurke:
Thanks to all who looked at this one!

PR #2521, opened 2026-09-15
XSPEC: allow building against XSPEC 13.0.0

DougBurke:
# Summary

Fix an error seen when compiling the code against XSPEC version 13.

# Details

The XSPEC 13.0.0 code base defines a K symbol that conflicts with the K macro defined by sherpa/constants.hh. It turns out that XSPEC just needs the constat module and not any of the macros it defines, so remove them.

This is a version of 

- #2516

but focussed just on the XSPEC exstention.


DougBurke:
This PR is also included in

- #2520

but it has been separated out to make it easier to review.

sherpa/include/sherpa/astro/xspec_extension.hh
```diff
@@ -24,7 +24,27 @@
 // Have sherpa includes first so that Python.h is first, to avoid warning
 // messages about redefining _XOPEN_SOURCE
 #include <sherpa/extension.hh>
+
+// Disable unused constant macros
```
wmclaugh: The comment here would be better under the #include of <sherpa/constants.hh>  (ie. switch order of lines 28/29) since the code is only removing the macros from constants.hh

wmclaugh:
This will remove the 17 macros (such as K and SQRT_PI). The function templates for k and sqrt_pi (and the others) will still be there. Do you want to remove all the macros from sherpa/constants.hh or just the conflict (K)?

DougBurke:
Since the XSPEC extension code doesn't use any of the macros I believe it's better to remove them all, to avoid any future problem (e.g. if the next release of XSPEC uses another symbol).

If we were creating a library that downstream users were using then it might make sense to have a policy about the use of the macros, but in this case we are injecting code (constants.hh and the like) into someone else's code base (XSPEC) so I claim it's better to avoid polluting the namespace as much as we can here. If that makes sense.

PR #2524, opened 2026-09-16
Improve developer docs after recent meson migration.

hamogu:
## Summary
Add aditional instruction for developer builds in the new meson world.

## Details
(Thanks to Claude Code which helped me figure out what's going on in my environment faster than I would have otherwise.)


docs/developer/index.rst
```diff
@@ -224,18 +224,31 @@ Unfortunately each option has to be supplied separately using the
...
 The ``scripts/make_build_args_ciao`` script can simplify this:
 
-    args=`scripts/make_build_args_ciao`
-    pip install . $args
+    pip install . $(scripts/make_build_args_ciao)
+
+However, as a developer (after all, this is the "Contributing to Sherpa development"
+section) you probably want to install in editable mode.
+Without editable mode, you might run into `ImportError while loading conftest`
+because `conftest.py` will exisit twice - in the source tree and in the installed code.
```
wmclaugh:
```suggestion
because `conftest.py` will exist twice - in the source tree and in the installed code.
```

hamogu:
The questions is if we need all these changes, because some of it duplicates what's already written in the general docs for making a PR.

docs/developer/index.rst
```diff
@@ -177,7 +177,7 @@ Conda can be used to install all the dependencies for Sherpa, including
 ::
 
     conda create -n sherpaciao -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge ds9 ciao
-    conda install -n sherpaciao --only-deps -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge sherpa
+    conda install -n sherpaciao --only-deps -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge sherpa ninja meson meson-python
```
DougBurke: So, I believe that `meson-python` should bring in `meson` so we could simplify the line, but as it's likely to by used via cut-n-paste it's doesn't seem worth it.

These are only needed if you do an editable install though,  but as this is in the developer section and below you say "do an editable install" I guess it's okay.

docs/developer/index.rst
```diff
@@ -224,18 +224,31 @@ Unfortunately each option has to be supplied separately using the
...
 line of::
 
-    pip install . -Csetup-args=build-group=false \
-      -Csetup-args=build-stk=false \
-      -Csetup-args=build-region=false \
-      -Csetup-args=build-wcssubs=false \
-      -Csetup-args=region-prefix=$CONDA_PREFIX \
-      -Csetup-args=wcssubs-prefix=$CONDA_PREFIX \
-      -Csetup-args=xspec-prefix=$CONDA_PREFIX
+    pip install . -Csetup-args=-Dbuild-group=false \
```
DougBurke, line 227: Argh - I'm not sure how I messed this up

docs/developer/index.rst
```diff
@@ -224,18 +224,31 @@ Unfortunately each option has to be supplied separately using the
...
+      -Csetup-args=-Dbuild-stk=false \
+      -Csetup-args=-Dbuild-region=false \
+      -Csetup-args=-Dbuild-wcssubs=false \
+      -Csetup-args=-Dregion-prefix=$CONDA_PREFIX \
+      -Csetup-args=-Dwcssubs-prefix=$CONDA_PREFIX \
+      -Csetup-args=-Dxspec-prefix=$CONDA_PREFIX
 
 The ``scripts/make_build_args_ciao`` script can simplify this:
```
DougBurke:
```suggestion
The ``scripts/make_build_args_ciao`` script can simplify this::
```

At least I hope this would indicate this section should be highlighted, rather than the current appearance (which is a 'bug' in my original version, not this PR):

Image

docs/developer/index.rst
```diff
@@ -224,18 +224,31 @@ Unfortunately each option has to be supplied separately using the
...
+section) you probably want to install in editable mode.
+Without editable mode, you might run into `ImportError while loading conftest`
+because `conftest.py` will exist twice - in the source tree and in the installed code.
+The meson build system will automatically re-build the extension modules when they
+are imported. That means two things:
+
+1. You can edit the source code and then re-run the tests without having to re-install Sherpa manually.
+2. By default `pip` creates an isolated build environment and installs any missing dependencies
+   (e.g. `ninja` which is used my meson) there. When re-building it will look for a
+   temporary directory that no longer exists. Instead, install mason/ninja in the conda environment and then use the `--no-build-isolation` option to `pip`:
```
DougBurke:
```suggestion
   temporary directory that no longer exists. Instead, install mason/ninja in the conda environment and then use the `--no-build-isolation` option to `pip`::
```

(again for the "yellow background" in the HTML output).

docs/developer/index.rst
wmclaugh: Running "conda search meson-python --info" shows that meson-python in conda-forge has dependencies so both ninja and meson.

hamogu:
```suggestion
    conda install -n sherpaciao --only-deps -c https://cxc.cfa.harvard.edu/conda/ciao -c conda-forge sherpa meson-python
```

DougBurke (approved):

Issue #2526, opened 2026-09-16
get_xspec_version doesn't find xspec through xspec-prefix


harlantc:
Ran into an issue when trying to sherpa against CIAO's 12.15.1 Xspec. I supply a prefix with `pip install ... -Csetup-args=-Dxspec-prefix=$CONDA_PREFIX`, but `get_xspec_version` fails to load the library with dll. I can work around it by setting `(DY)LD_LIBRARY_PATH` before calling the build.

I'm not sure why this is happening with 12.15.1, it is possible I dropped LD paths from the xspec-modelsonly package env (we shouldn't be setting that anyway).

Looking at `get_xspec_version`, I don't see how `xspec_libdir` from meson is passed to it, so I suspect that's the problem.

other minor note, also I think this line in `get_xspec_version`
```python
sys.stderr.write("Usage: {sys.argv[0]}\n")
```

should be a `f"Usage: {sys.argv[0]}\n"`

DougBurke:
I have seen some strange issues when using an in-place build (`pip install -e . --no-build-isolation -Csetup-args=-Dxspec-prefix...`) when certain changes cause a rebuild and then the script fails, but if I run it manually it does. So maybe in some cases it's being run in an environment with / without the necessary library path. Will need to do a little sleuthing. Maybe we just need to set `env` when calling https://mesonbuild.com/Reference-manual_functions_run_command.html ?

Good spot with the error message

DougBurke:
@harlantc I'm hoping that #2528 fixes this, but it is a bit of a guess. unfortunately we can't add the path to the get_xspec_version script as it needs to be set before the script is run, hence the use of the env option for `run_command`. Hand wave. Hand wave. Profit.

harlantc:
So I tried updating `get_xspec_version` to load the lib from a full path if supplied, built off the one we have in Meson, but that fails as we're missing all the other libraries XSFunctions needs without a load path.

while I wait for your change, I can patch this in our sherpa package build with a patch. This is only impacting my macOS build so I'm keeping the change to that for now

```diff
diff --git a/meson.build b/meson.build
index 9c8325a0..d9ecd73c 100644
--- a/meson.build
+++ b/meson.build
@@ -164,6 +164,8 @@ if xspec_prefix != ''
   #
   xspec_version = run_command(py,
     ['scripts/get_xspec_version'],
+    env: {'DYLD_LIBRARY_PATH': xspec_libdir},
     check: true
   ).stdout().strip()
```

[...]

PR #2528, opened 2026-09-16
Improve the XSPEC version check when building Sherpa

DougBurke:
# Summary

Ensure the library path is set when calling get_xspec_version during a build.

# Details

Hopefully this addresses #2526


harlantc (approved):

meson.build
```diff
@@ -162,9 +162,18 @@ if xspec_prefix != ''
   # wants to specify the version, rather than pick it up directly,
   # but for now do not support this usage.
   #
+  # Do we need to set up (DY)LD_LIBRARY_PATH? Using append is a bit
+  # excessive as the environment hasn't been set. Would it be better
+  # to do this with meson.add_devenv?
+  #
+  xspec_env = environment()
+  xspec_env.append('LD_LIBRARY_PATH', xspec_libdir, separator: ':')
+  xspec_env.append('DYLD_LIBRARY_PATH', xspec_libdir, separator: ':')
```
wmclaugh: Rather than adding/editing both LD and DYLD library env vars perhaps check the operating system and use DYLD for mac and LD otherwise. Alternatively, I believe add_devenv should handle this automatically.

DougBurke: I was trying to find information on how to get the os from meson but couldn't obviously fine anything , hence just applying both (as they are essentially distinct). Do you or @harlantc know the correct way to do this (I think AI hallucinated me a `get_os` call, or maybe it's https://mesonbuild.com/Reference-manual_builtin_host_machine.html#build_machinesystem and https://mesonbuild.com/Reference-tables.html#operating-system-names is what I'm looking for )

One advantage of this approach over using `add_devenv` (or, as least I imagine `add_devenv` works) is that it only changes the environment for this call; there's a small possibility it could mess something up (maybe the linking), but that's a complete guess on my end.

DougBurke: Let's see if `host_machine.system()` works!

DougBurke: Yay: the `macOS/ARM` build displayed

```
Message: Appending to DYLD_LIBRARY_PATH for get_xspec_version
Message: Found XSPEC version: 12.14.0k
```

so it looks like it worked.

wmclaugh (approved):

PR #2531, opened 2026-09-17
docs: tweak the layout of a table

DougBurke:
# Summary

Tweak the docstring for load_table_model to make it easier to create an ahelp version of it.

# Details

This change does not lool like much, but it makes my life easier when generating the ahelp XML files. It's the difference between handling

```
  <definition_list_item>
    <term>ncols</term>
    <classifier>int, optional</classifier>
    <definition>
      <paragraph>The number of columns to read in (the first <literal>ncols</literal> columns
      ...
```

and

```
  <definition_list_item>
    <term>ncols: int, optional</term>
    <definition>
      <paragraph>The number ...
```

and you really don't want to know this ...


hamogu (approved):
Note: If a change doesn't contain code lines, only docs, you can put `[skip ci]` into the commit message, which will skip the CI runs, but still run RTD. 
That way, we don't have to wait for CI to finish here before we lear about the next PR, since GH restricts how many parallel CI instances can be run for any one project. IT usually doesn't matter, but when we get a lot of small PRs before release time, it's helpful.

